PR #5050, opened 2018-07-04
update mailservers from DO to use Floating IPs

jakubgs:
I've replaced the servers yesterday to upgrade them to Ubuntu 18.04 and this resulted in mailservers from Digital Oceans becoming unavailable due to status-react configuration using the Droplet IPs rather than the Floating IPs.

This server replacement did not affect Google Cloud mailservers or for example bootnodes from both providers because those were properly using Floating IPs.

>__NOTE:__ Redoing because needs testing. Merge was undone.


jeluard (approved):

flexsurfer (approved):

jakubgs:
Change merged with the `./scripts/merge-pr.sh` script.

PR #5054, opened 2018-07-04
[#4301] Move message timestamp to the left for right-to-left languages

### LINKED ISSUE #4301: Arabic language aligned to the left in message bubble overlaps timestamp
***Type***: Bug

***Summary***: Right to left text messages aligned to the right in the message bubble and it overlaps message timestamp. We need to take into account, and when the "align text to the right" event is triggered in message bubble - consider a margin of the message timestamp. 

#### Expected behavior
Right to left text messages does not overlap timestamp in the message bubble

#### Actual behavior
Right to left text messages aligned to the right in the message bubble and does not overlap message timestamp.

screen shot 2018-05-16 at 16 59 25

### Reproduction

- Open Status
- Navigate to any type of chat
- Type arabic text and send (e.g. `أنت محظوظ اليوم!`)

### Additional Information
* Status version: Develop/Release 0.9.18

#### Logs

TF session: https://app.testfairy.com/projects/4803622-status/builds/8304711/sessions/55/?accessToken=gbz/I01Osj4ZslRk28TSQFBW4x8

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803622-status/builds/8304711/sessions/55/?accessToken=gbz/I01Osj4ZslRk28TSQFBW4x8">TestFairy: Status 0.9.16-169-gaa108cda+ (2596) @55</a></strong></div></blockquote>

dmitryn:
fixes #4301

### Summary:

Move message timestamp to the left for right-to-left languages to prevent message text and timestamp overlapping.

### Review notes (optional):
After researching React Native and Android API i haven't found better way to detect RTL text rather than using regexp.

status: ready

screenshot 2018-07-04 18 23 06

### DIFF
--- a/src/status_im/chat/views/message/message.cljs
+++ b/src/status_im/chat/views/message/message.cljs
@@ -144,15 +144,22 @@
       [message-content-command-send message]
       [message-content-command-with-markup message])))
 
-(defview message-timestamp [t justify-timestamp? outgoing command?]
+(def rtl-characters-regex #"[^\u0591-\u06EF\u06FA-\u07FF\u200F\u202B\u202E\uFB1D-\uFDFD\uFE70-\uFEFC]*?[\u0591-\u06EF\u06FA-\u07FF\u200F\u202B\u202E\uFB1D-\uFDFD\uFE70-\uFEFC]")
+
+(defn right-to-left-text? [content]
+  (let [char (first content)]
+    (re-matches rtl-characters-regex char)))
+
+(defview message-timestamp [t justify-timestamp? outgoing command? content]
   (when-not command?
-    [react/text {:style (style/message-timestamp-text justify-timestamp? outgoing)} t]))
+    (let [rtl? (right-to-left-text? content)]
+      [react/text {:style (style/message-timestamp-text justify-timestamp? outgoing rtl?)} t])))
 
 (defn message-view
-  [{:keys [timestamp-str outgoing] :as message} content {:keys [justify-timestamp?]}]
+  [{:keys [timestamp-str outgoing content] :as message} message-content {:keys [justify-timestamp?]}]
   [react/view (style/message-view message)
-   content
-   [message-timestamp timestamp-str justify-timestamp? outgoing (get-in message [:content :command])]])
+   message-content
+   [message-timestamp timestamp-str justify-timestamp? outgoing (get-in message [:content :command]) content]])
 
 (def replacements
   {"\\*[^*]+\\*" {:font-weight :bold}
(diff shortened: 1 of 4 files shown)


src/status_im/chat/views/message/message.cljs
```diff
@@ -143,15 +143,22 @@
       [message-content-command-send message]
       [message-content-command-with-markup message])))
 
-(defview message-timestamp [t justify-timestamp? outgoing command?]
+(def rtl-characters-regex #"^[^\u0591-\u07FF\u200F\u202B\u202E\uFB1D-\uFDFD\uFE70-\uFEFC]*?[\u0591-\u07FF\u200F\u202B\u202E\uFB1D-\uFDFD\uFE70-\uFEFC]")
```
jeluard: Maybe add a link pointing to some relevant documentation ?

jeluard (approved):

goranjovic (approved):

annadanchenko:
@dmitryn I'm still testing but there is already 1 issue:

### issue 1
If type numbers in Arabic keyboard then text overlaps timestamp:
screenshot_20180710-140455

Can reproduce both on iOS (real iphone 7+, iOS 11.4 and on real Pixel XL, Android 8.0.1)
Steps:
1. Add Arabic keyboard
2. join public chat
3. switch to Arabic
4. switch to "numeric keyboard"
5. type  chars from first line shown on keyboard and send them to chat
6. check where timestamp is shown in the chat

dmitryn:
@annadanchenko i've fixed case with arabic numbers. Please keep in mind that there could be other cases with broken layout since i didn't find a way to detect right-to-left text layout for react-native and use regular expression to detect this which is frugal and error-prone.

I will create separate issue to find proper way to detect RTL text direction, but for now this PR should be covering most use-cases.

Serhy:
There are still issue parsing particular scenarios, but as a quick fix (until #5124 fix it) I think it's good to go.
When text everlaps timestamp:
1) LTR text preceeds RTL text (e.g. `Hello رحبا`)
2) Wtispace as first symbol followed by RTL text (e.g. `   رحبا`)
3) Arabic numbers scenario Anna described in https://github.com/status-im/status-react/pull/5054#issuecomment-403790772

Above are not something which is used often and taking into account that
* RTL script does not overlaps timestamp  - OK
* RTL followed by smiles does not overlaps timestamp - OK
* LTR text + RTL text  + LTR text (e.g. `Status ذلانًنته status`)
and the separate issue to find a proper way to detect RTL text direction created (#5124) I think we are good to merge current fix as a solution.

PR #5067, opened 2018-07-05
bug #5057 - automatically converts recovery phrase to lowecase during…

### LINKED ISSUE #5057: Intentionally decapitalize words in the recovery phrase field
### User Story

As a user, I don't want to create a new account while recovering it by mistake.

### Description

*Type*: Bug

*Summary*: Process of recovering account is a case sensitive.  If you use your recovery phrase but with the capitalized first letters, our app will create a different account from that. We need to prevent this by decapitalize words in the recovery phrase field

#### Expected behavior
img_5718

#### Actual behavior
img_5717

### Reproduction

- Open Status
- Recover account

goranjovic:
fixes #5057

### Summary:

Changes recovery screen to automatically convert any typed uppercase letters to lowercase.

### Steps to test:
- Recover account
- Try to recover with uppercase letters in recovery phrase

status: ready

### DIFF
--- a/src/status_im/ui/screens/accounts/recover/views.cljs
+++ b/src/status_im/ui/screens/accounts/recover/views.cljs
@@ -32,7 +32,7 @@
       :multiline           true
       :default-value       passphrase
       :auto-correct        false
-      :on-change-text      #(re-frame/dispatch [:set-in [:accounts/recover :passphrase] %])
+      :on-change-text      #(re-frame/dispatch [:set-in [:accounts/recover :passphrase] (string/lower-case %)])
       :error               error}]))
 
 (defview password-input [password]


src/status_im/ui/screens/accounts/recover/views.cljs
```diff
@@ -32,7 +32,7 @@
       :multiline           true
       :default-value       passphrase
       :auto-correct        false
-      :on-change-text      #(re-frame/dispatch [:set-in [:accounts/recover :passphrase] %])
+      :on-change-text      #(re-frame/dispatch [:set-in [:accounts/recover :passphrase] (string/lower-case %)])
```
jeluard, line 35: How about preventing upper case directly? Does this convert the display too?

dmitryn: Good suggestion to force lowercase in RN component. Maybe CSS `textTransform: lowercase` would work here.

goranjovic: @jeluard If you type say `D` it immediately turns into `d` before you even type the next letter. So, yes - this is both value and display.

goranjovic: @dmitryn Not sure that would work as expected in a text input.

dmitryn (approved):

src/status_im/ui/screens/accounts/recover/views.cljs
dmitryn: @goranjovic yes, just checked and it doesn't work. And your solution is working fine for me, auto corrects text in place.

jeluard (approved):

src/status_im/ui/screens/accounts/recover/views.cljs
goranjovic: TBH, I would've preferred something like a lowercase-only keyboard. But it wouldn't really solve anything since a lot of people would probably copy paste it, not actually type.

dmitryn:
kapture 2018-07-05 at 16 34 55

@jeluard

PR #5068, opened 2018-07-05
Update transactions when navigate to tx details from chat

### LINKED ISSUE #4851: Automatically update Transaction Details screen if new data is available
### Description

*Type*: Feature

*Summary*: A user may expect that Transaction Details screen will update itself automatically when needed (it was also my expectation). But, it does not. The only way to refresh it is to go back and open the screen again. It's not very intuitive.

#### Expected behavior
Transaction Details screen refreshes itself automatically.

#### Actual behavior
The only way to refesh Transaction Details screen is to close and open it again.

### Reproduction

- Open Status
- Send a transaction
- Go to Transaction History and open the transaction details

### Additional Information
* Status version: 0.9.20

dmitryn:
Addresses #4851

### Summary:

Update transactions when navigate to tx details from chat message generated with /send command.

### Review notes (optional):

Data isn't getting updated if user stays on that screen for a while. But that's a part of another issue https://github.com/status-im/status-react/issues/4851

status: ready

### DIFF
--- a/src/status_im/chat/views/message/message.cljs
+++ b/src/status_im/chat/views/message/message.cljs
@@ -62,6 +62,7 @@
   (letsubs [confirmed? [:transaction-confirmed? tx-hash]
             tx-exists? [:wallet-transaction-exists? tx-hash]]
     [react/touchable-highlight {:on-press #(when tx-exists?
+                                             (re-frame/dispatch [:update-transactions])
                                              (re-frame/dispatch [:show-transaction-details tx-hash]))}
      [react/view style/command-send-status-container
       [vector-icons/icon (if confirmed? :icons/check :icons/dots)


janherich:
I noticed that we manually refresh transactions in many places (actions), which is tedious and error prone. On top of that, we don't even guarantee consistent views (like number of confirmations for particular tx) when the user is just looking at the screen and doesn't do any actions.
Should the transaction feed be actually "live", so we are always up-to date with no need for manual fetching tied with particular action or navigation ?
In shorter term, this could be achieved by periodically (each ~15s for example) polling and calling `:update-transactions` as result of polling callback, even better would be some sort of "push" mechanism from `status-go`, so you can "subscribe" to set of transactions ids and each time something new related to txs in set arrives,  signal is triggered.
What do you think @dmitryn @goranjovic ?

dmitryn:
@janherich yes, i would prefer subscription model instead of polling if it's possible (needs researching). We could make polling updates when new block is mined (if `web3.eth.filter('latest')` supported).

If subscription/filter doesn't work for us, regular polling might be good enough solution. We can adjust refresh time parameter to average block mining time.

p.s. I agree that current PR just adds ad hoc fixing, but doesn't solve problem in general.

goranjovic (approved):

flexsurfer (approved):

lukaszfryc:
I sent "/send" ETH in 1:1 chat and then clicked on the transaction message and stayed there for ~10 min. The screen was not updated, I saw 0 confirmations. @dmitryn should the PR support this scenario?

dmitryn:
@lukaszfryc this PR only adds fetching transactions data when you go to tx details from chat message. If data is not there and wallet shows 0 confirmations, then nothing happens. You should go back to chat and tap on message to go back again to start fetching tx data. Or you could go to wallet to start the update from there.

lukaszfryc:
@dmitryn so if I'm on the transaction details screen it will not refresh itself automatically but only when I go back to chat and click and open it again? At least it works like this on my side. If this is how it works, I'd say that this PR does not close https://github.com/status-im/status-react/issues/4851 but refreshes transaction details screen **only when it's being opened**. Of course the polling is done in the background but users still need to reopen the transaction details screen to see updates.

dmitryn:
@lukaszfryc 
>if I'm on the transaction details screen it will not refresh itself automatically but only when I go back to chat and click and open it again?

Correct.

>refreshes transaction details screen only when it's being opened

Yes.

>Of course the polling is done in the background but users still need to reopen the transaction details screen to see updates

It depends on how much time passed between tx sent and user tapped on message. If enough time is passed (tx propagated to etherscan so we could fetch it from there), user will see updated tx details screen almost immediately after he taps on message and arrive to tx details screen (we do fetching in background after he taps and it usually take less than a second to fetch and update the UI)

dmitryn:
And i agree that there are some cases which are not covered by this PR, but at least it covers some of them. We need to think of better way of updating transactions data rather than polling based on events.

lukaszfryc:
@dmitryn cool. I suggest to merge this PR but do not close https://github.com/status-im/status-react/issues/4851. I can add a comment there of what was improved by this PR.

dmitryn:
@lukaszfryc sure, thanks. I will change PR body and remove "fixes #4851" reference from it.

PR #5109, opened 2018-07-10
use https instead of ssh in package.json

yenda:
We shouldn't require ssh to use public repos

status: ready


jakubgs (approved):
Nice.

jeluard:
@alwx Did work on that recently. He probably can provide feedback.

cammellos (approved):
looks good to me, pending @alwx judgement

flexsurfer (approved):
yeah did the same locally for development

alwx:
I don't know why but in this case `npm 9` cannot resolve dependencies on my machine, but everything is fine when I use `yarn` instead. Because we're going to switch to yarn, it can be merged.

But to be honest I don't really understand why this change is needed.

alwx (approved):

cammellos:
@alwx having https helps on CI systems for example, as `ssh` requires to have a key registered with github

alwx:
ok, merge it then

PR #5110, opened 2018-07-10
fix make release-android command

rasom:
status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ prepare-ios: prepare ##@prepare Install iOS specific dependencies
 release: release-android release-ios ##@build build release for Android and iOS
 
 release-android: prod-build-android ##@build build release for Android
+	cd android; ./gradlew react-native-android:installArchives
 	react-native run-android --variant=release
 
 release-ios: prod-build-ios ##@build build release for iOS release


yenda (approved):

Makefile
```diff
@@ -46,6 +46,7 @@ prepare-ios: prepare ##@prepare Install iOS specific dependencies
 release: release-android release-ios ##@build build release for Android and iOS
 
 release-android: prod-build-android ##@build build release for Android
+	cd android; ./gradlew react-native-android:installArchives
```
jeluard, line 49: Does this need to be done each time? How about using the setup script?

flexsurfer: yeah had a same question here

rasom: @jeluard i'm not familiar with makefile's syntax/structure and have no idea where that setup script lives, besides i'm not using `make`. So if you know the way to make it better please add commit to this PR

jeluard: I am not familiar with make either but probably you don't want to install this frequently.
Setup script is here https://github.com/status-im/status-react/blob/develop/scripts/setup
Please update it accordingly

jeluard (changes requested):
This change is better suited in the setup script

PR #5126, opened 2018-07-10
Install gradle archives during setup.

pedropombeiro:
Fixes broken `make release-android` step on new machines.

Addresses the same issue as https://github.com/status-im/status-react/pull/5110, but on the setup script.
Tested the build on a new machine where it was previously failing, and build now passes on the first attempt.

status: ready

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -31,6 +31,7 @@ install_node
 install_watchman
 install_react_native_cli
 install_android_sdk
+install_android_ndk
 install_maven
 install_cocoapods
 
@@ -49,6 +50,6 @@ dependency_setup ./re-natal enable-source-maps
 dependency_setup \
   "mvn -f modules/react-native-status/ios/RCTStatus/pom.xml dependency:unpack"
 
-using_cocoapods && dependency_setup "cd ios && pod install && cd .."
+using_cocoapods && dependency_setup "pushd ios && pod install && popd"
 
 setup_complete
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -156,6 +156,8 @@ function install_android_sdk() {
   elif is_linux; then
     install_android_sdk_linux
   fi
+
+  use_android_sdk
 }
 
 function install_android_sdk_linux() {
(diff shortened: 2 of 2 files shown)


rasom (approved):

jeluard (approved):

jeluard:
Does this depends on the ndk installation?

pedropombeiro:
I believe so @jeluard. Should we show a message to the user if the NDK is not detected?

pedropombeiro:
I hadn't seen yet that the build failed. So I guess that the problem is that the CI server doesn't actually need this in order to do its build. Maybe we should show a warning when the NDK is not detected, but not actually fail?

scripts/setup
```diff
@@ -46,9 +46,16 @@ dependency_setup \
 dependency_setup ./re-natal deps
 dependency_setup ./re-natal enable-source-maps
 
+if [[ -z "${ANDROID_NDK1}" ]]; then
+  dependency_setup \
+    "pushd android && ./gradlew react-native-android:installArchives && popd"
+else
+    cecho "@yellow[[Android NDK installation not detected. If you plan to build for Android, \
+please follow the instructions at https://docs.status.im/docs/build_status_linux.html to install the Android NDK:]]"
```
jeluard: How about if you build android on OSX?

jeluard (commented):
Do you think it would be possible to have the script install the NDK ?

scripts/setup
pedropombeiro: will add a check and display appropriate link

pedropombeiro:
@jeluard @rasom need a new review as the approach has changed slightly: after discussing with Dmitry N who's on OSX, I've changed approach to rely on `local.properties` file content instead of environment variables. This is more reliable and affects only status-react. If `ndk.dir` isn't defined, it will download the NDK to `~/Android/Sdk` and update the `ndk.dir` value to point to it.

jeluard (approved):

rasom (approved):

pedropombeiro:
Added a minor improvement so that we refer the user to the correct documentation for the host platform in case the Android SDK is not present on the machine, and due to that, moved the script which was getting unwieldy to `scripts/lib/setup/installers.sh` so that `scripts/setup` stays as clean as possible.

Will merge to develop once build passes if there's nothing against this (changes are not significant relative to what was already reviewed).

PR #5133, opened 2018-07-11
realm and status-go generated files location

maxhora:
### Summary:

Specify data folder as location for status-go's keychain and realm db files.

### Testing notes:
- No files should be generated in directory from where StatusIm is launched
- As bonus, on macOS it should be possible now to launch bundle directly from mounted .dmg ( Copying to /Applications is not necessary )
- Since keystore and db files location is moved from local working directory to system wide data directory, all old accounts should be not found by app on start

status: ready

### DIFF
--- a/desktop/main.cpp
+++ b/desktop/main.cpp
@@ -8,16 +8,16 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  */
 
- // #define BUILD_FOR_BUNDLE
+// #define BUILD_FOR_BUNDLE
 
 #include <QCommandLineParser>
 #include <QFile>
 #include <QGuiApplication>
 #include <QProcess>
 #include <QQuickView>
+#include <QStandardPaths>
 #include <QTimer>
 #include <QUrl>
-#include <QStandardPaths>
 
 #include "attachedproperties.h"
 #include "reactitem.h"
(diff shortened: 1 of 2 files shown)


vitvly (approved):

vkjr (approved):

cammellos (approved):

churik:
@MaxRis 
Thank you for PR!
Everything looks great, except uninstalling process - it is unclear at all.

Steps:
- stop app
- eject Status mounted disk
- remove folder `StatusIm` from `Application Support` directory
- open `.dmg` again
- start app

Nothing happens, in `Activity monitor` StatusIm process consumes 99,9 % CPU.
Also tried to delete StatusIm using special uninstall software (`Clean My Mac` for me). Results are the same - application cannot start,  StatusIm process consumes 99,9 % CPU.
So please provide instructions how to uninstall app.

maxhora:
@churik, have you tried to copy the app to /Applications folder and start from there to make it work? I don't think that any special uninstall step is required, will try to check running issue on my machine, but I think it's not related to current changes.

churik:
Yes, I did. Here is  [video](http://take.ms/M3NBi).
Regardless of the place from where it was launched (Application Folder, separate folder or mounted disk),  it won't work and only consumes CPU

maxhora:
@churik I can repro the issue with not starting app. To make it work again you can manually create back directory "~/Library/Application Support/StatusIm". Not sure at the moment if that the general issue of app that it can't create this directory on its own or what exactly.

churik:
OK, if fix doesn't take a lot of time - it makes to fix it in P0, but IMO RN issues have higher prio right now.

maxhora:
Hey @churik , finally, it seems the issue you have found is related to current changes. I need to resolve it in this PR to not break first run process completely. Good catch! Thank you

maxhora:
@churik have resolved the issue with new commit in this PR, thank you

churik:
Tested:
- fresh installation
- uninstall by dropping App to Trash
- install App to OS with existing data (with data in `Application Support`) - all accounts and data is preserved;

maxhora:
@churik , thank you!

PR #5149, opened 2018-07-12
Disable auto backup on IOS and android

### LINKED ISSUE #5127: Remove our app from automatic backups

### User Story

As a user, I want my app data to stay locally on my device and not get synced to Google or Apple so that I stay in control of my data.

### Description

*Type*: Feature

*Summary*: The app data currently resides on directories which get automatically picked up by the OS and synced online, if the user has that functionality enabled in the OS.

#### Expected behavior

TODO

#### Actual behavior

TODO

### Solution
*Summary*: We should either disable the sync functionality for the app or move the data to another folder which doesn't get uploaded to the cloud.

cammellos:
fixes #5127 

### Summary:

Move all the realm files from:

`files/` -> `no_backup` on android
`Documents/` -> `Library/` on IOS 

As those directories are not backed up by default.

It also changes the structure of the filesystem for easier management:

`realm/default.realm`
`realm/accounts/{sha3-of-address}`

It takes the `sha3` of the address so that we don't show which addresses the user has on the device in case someone gets access to the filesystem.

### Testing notes:
Upgrade from previous versions, the change should be transparent to the user.
Newly installed devices.
Tested on android simulator/real device, ios only simulator.

### Steps to test:

A bit difficult to test on android as you can't see what's being backed up. If you have access to the filesystem you can see that files that before were under `files` are not under `no_backup` using the hierarchy described above.
Similar for IOS (not sure if you can inspect icloud drive)

status: ready

### DIFF
--- a/src/status_im/data_store/realm/core.cljs
+++ b/src/status_im/data_store/realm/core.cljs
@@ -7,6 +7,8 @@
             [taoensso.timbre :as log]
             [status-im.utils.fs :as fs]
             [status-im.utils.async :as utils.async]
+            [status-im.utils.platform :as utils.platform]
+            [status-im.utils.ethereum.core :as utils.ethereum]
             [cognitect.transit :as transit]
             [status-im.react-native.js-dependencies :as rn-dependencies]
             [status-im.utils.utils :as utils]))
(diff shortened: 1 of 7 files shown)


janherich (approved):

src/status_im/data_store/realm/core.cljs
```diff
@@ -34,26 +36,67 @@
     (when (exists? js/window)
       (rn-dependencies/realm. options-js))))
 
-(defn- delete-realm
-  [file-name]
-  (.deleteFile rn-dependencies/realm (clj->js {:path file-name})))
-
-(defn- default-realm-dir [path]
-  (string/replace path #"default\.realm$" ""))
+(defn- is-account-file? [n]
```
jeluard, line 39: Are those utilities worth testing?

jeluard (approved):

src/status_im/data_store/realm/core.cljs
cammellos: probably won't change much, but worth adding some tests just in case

cammellos: added!

jeluard (approved):

lukaszfryc:
@cammellos I tested different upgrade scenarios on Android and iOS. I also looked at the files structure in Android Studio Device File Explorer. Looks good 👍

PR #5160, opened 2018-07-13
Round robin mailserver on failure

### LINKED ISSUE #5130: Fix mailserver connection problem UX
# Problems

When a mailserver is down and was randomly selected by the app at startup, one need to restart the app or select another mailserver manually in the settings.
A server can be down because it has been blocked, we stopped it, it doesn't have any free slot for connection.

# Implementation

The app should try to connect to another mailserver if the one that was picked doesn't work, unless the user has explicitly selected one.

We should not change make changes to more than half of the mailservers per release cycle in order to avoid having more than half of mailserver known by the app down (unless security reasons requires it)

# Acceptance Criteria

When a mailserver is down user doesn't notice it because the app automaticaly connect to another one.

cammellos:
fixes #5130 

### Summary:

When a connection with a mailserver fails, the system will:

1) If no selection was explicitly set, pick the next mailserver and try to connect to it (will move from `a` to `b`). 
2) If a selection was already made, it will show the usual error.

The timeout has also been shortened to 15 seconds instead of 60.
Mailserver is still picked at random on each login, unless explicitly set.
Initially I thought about excluding custom mailservers from the rotation, but settled on not doing that as I can't see a good reason why.

### Testing notes:

You can test using this build:

apk uploaded to https://i.diawi.com/1QqK3M
ipa uploaded to https://i.diawi.com/TDXkin

All the mailservers but `f` are not working, it should eventually settle on `f` and you should be able to receive old messages.

status: ready

### DIFF
--- a/src/status_im/models/mailserver.cljs
+++ b/src/status_im/models/mailserver.cljs
@@ -59,15 +59,39 @@
 (defn fetch-current [{:keys [db] :as cofx}]
   (fetch (:inbox/current-id db) cofx))
 
+(defn preferred-mailserver-id [{:keys [db] :as cofx}]
+  (let [chain (models.network/get-chain cofx)]
+    (get-in db [:account/account :settings :wnode chain])))
+
+(defn- round-robin
+  "Find the choice and pick the next one, default to first if not found"
+  [choices current-id]
+  (let [next-index (reduce
+                    (fn [index choice]
+                      (if (= current-id choice)
+                        (reduced (inc index))
+                        (inc index)))
+                    0
+                    choices)]
+    (nth choices
+         (mod
+          next-index
+          (count choices)))))
+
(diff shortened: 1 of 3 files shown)


dmitryn (approved):

goranjovic (approved):

Serhy:
This one looks good! Round-robin applied to the Mailserver rotation if not connection established to particular mailserver. Custom mailservers are included in the rotation as well. 

Once contact set particular Maiserver - account sticks to this mailserver. And if it's doesn't work - 'Coul not connect to Mailserver. Tap to reconnect' is shown as before.

Offline messaging works fine once contact connected the "working" mailserver.

Ready for merge, thanks @cammellos !

Serhy:
Lets just revert the changes made for testing purpose before the merge ;)

PR #5161, opened 2018-07-13
bug #5155 - fixed the bug where every token is unknown in send command

### LINKED ISSUE #5155: 'Unknown token' on attempt to send/request tokens in 1-1 chat (which are enabled in wallet)
***Type***: Bug

***Summary***: On attempt to send or request any token (enabled in Wallet) in 1-1 chat there an `Invalid asset`+`Unknown token - [token]` error displays giving no way to send/request tokens in 1-1 chat.

#### Expected behavior
Able to send or request tokens in 1-1 chat

#### Actual behavior
`Unknown token - [token]` error on attempt to send or request token in 1-1 chat (including SNT, STT as default tokens)
screen shot 2018-07-12 at 20 36 39

### Reproduction

- Open Status and create an account
- Navigate to Wallet and set it up (this is to confirm SNT is enabled and available for interaction)
- Add a contact and navigate to 1-1 chat (e.g. of public key to add `0x042e30278db16d56bb646f3167050e6d61a2b92d924cd3e7ecdf9fb51bb1b8549b9de489ee3801fa07d96fd3249b0b4de4679e972e75a854b49f81980e026643ed`)
- Tap `[/]` command button -> `/request` -> tap `SNT`
- Enter any valid amout and tap 'Send' button
**Actual result:** `Unknown token - SNT` error displays
- Try to send `/send SNT 0` message
**Actual result:** `Unknown token - SNT` error displays

### Additional Information
* Status version: Develop 0.9.23 (6197) (494th Jenking nighlty build)
* Operating System:  Android and iOS 
**NOTE:** It's OK on 493rd nightly build, 

#### Logs
TF session: https://app.testfairy.com/projects/4803622-status/builds/8522488/sessions/4396235329/?accessToken=QImiFOl-AEC3p1Pi1gUSDUcRaxs

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803622-status/builds/8522488/sessions/4396235329/?accessToken=QImiFOl-AEC3p1Pi1gUSDUcRaxs">TestFairy: Status 0.9.23 (6197) @2018</a></strong></div></blockquote>

goranjovic:
fixes #5155

### Summary:

Recently format of `chain` was changed from keyword to string, which broke the token validation in `/send` command. This PR amends the issue.

### Steps to test:
- Try sending any token in chat

status: ready

### DIFF
--- a/src/status_im/chat/events/commands.cljs
+++ b/src/status_im/chat/events/commands.cljs
@@ -15,8 +15,8 @@
 ;; because balances are only fetched for them. Revisit this decision with regard to battery/network consequences
 ;; if we were to update all balances.
 (defn- allowed-assets [chain account]
-  (let [visible-token-symbols (get-in account [:settings :wallet :visible-tokens chain])]
-    (->> (tokens/tokens-for chain)
+  (let [visible-token-symbols (get-in account [:settings :wallet :visible-tokens (keyword chain)])]
+    (->> (tokens/tokens-for (keyword chain))
          (filter #(not (:nft? %)))
          (filter #(contains? visible-token-symbols (:symbol %)))
          (map #(vector (-> % :symbol clojure.core/name)


flexsurfer (approved):

cammellos (dismissed):
Please add some tests to avoid regressions

janherich (approved):

janherich:
@cammellos The whole thing will be extensively unit tested in my PR implementing cljs commands. And I mean really everything for every command, validation, parameter suggestions, sending, receiving, rendering preview, etc. :)

@goranjovic One thing I noticed during the recent work is that there are many places where we convert symbols from outer world entered by user or defined by rules out of our control (eq strings, like `"ETH"`, `"mainnet"`...) to clojure keywords and then back and forth multiple times.
This makes totally no sense, it's error prone and there is no inherent advantage of having it converted to clojure keyword.
I thin it's good to use keywords as keys for our internal data, like model attributes (`:current-chat`, `chats`, etc.) but in above cases, everything would be much simpler and less suspicious to errors when we would just leave such external data as they are, eq strings.

goranjovic:
@cammellos This is literally just a quickfix to avoid having a broken develop until Jan's changes are ready. A huge part of the current command code will be deleted very soon, so I don't want to invest too much time changing it now.

@janherich I agree in general, however the thing is how do we separate values that are fully internal vs values which are entered by the user? Currency symbol was a completely internal value until we got this command. On the other hand, chain is never actually entered by user manually...

dmitryn (approved):

cammellos:
@goranjovic @janherich it's more about getting in the habit of adding tests to on each PR/bugfix (where it makes sense of course), so we can make the codebase easier to work with and less threatening, especially for non-core members, otherwise the code risks becoming legacy very quickly, especially if we loose context.

But I'll stop complaining about it, and lift the review :)

goranjovic:
@cammellos Definitely agree with that in general. In fact, one of the major goals in the current refactoring is to make tests possible at all

lukaszfryc:
@goranjovic looks good. I tested sending and requesting STT, ETH in messages and wallet.

PR #5163, opened 2018-07-13
Use `brew_install android-ndk` in OSX

pedropombeiro:
This PR updates `installers.sh` so that the Android NDK is installed with brew on OSX.

Note: this script has not been tested on OSX, only on Linux.

status: ready

### DIFF
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -319,15 +319,22 @@ function install_android_ndk() {
     dependency_setup \
       "pushd android && ./gradlew react-native-android:installArchives && popd"
   else
-    local _ndkParentDir=~/Android/Sdk
-    mkdir -p $_ndkParentDir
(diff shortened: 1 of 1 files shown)


dmitryn (commented):
Seems like it should be `brew cask install android-ndk`, not `brew install android-ndk`
https://github.com/Homebrew/homebrew-core/commit/69c948752356aa56e7f1653366accb1262d124ea

Also, it installs version 17 not 10 https://github.com/Homebrew/homebrew-cask/blob/master/Casks/android-ndk.rb

dmitryn:
Also, on OSX i've installed ndk linux way (downloaded and then extracted to folder). Not sure using homebrew adds something to user xp as script does all automatically anyway.

pedropombeiro:
The suggestion to use `brew` came from @alwx, so I'll let him speak to that.

Issue #3752, opened 2018-04-03
Can't message in Rinkeby


asemiankevich:
### Description

*Type*: Bug

*Summary*: It appears that i can't see messages under Rinkeby / Rinkeby with RPC network. After i switch back to Ropsten - I can see all the messages coming

#### Expected behavior
I can exchange messages under Rinkeby network

#### Actual behavior
Messages are not delivered

### Reproduction

- Open Status
- Having 2 accounts under Rinkeby try to exchange messages

### Additional Information
* Status version: nightly 03/04/2018
* Operating System: Android, iOS

adambabik:
I will investigate first the state of our Rinkeby cluster.

adambabik:
One node was down, I fixed it. There should be three Rinkeby nodes available now. However, there are no mailservers for Rinkeby so only online messaging works.

Serhy:
This is not reproduced anymore. 
Having 2 accounts under Rinkeby network - exchange messages fine. Offline inboxing between two also works fine. 

Issue now only that Rinkeby user can't receive any message from Mainnet or Ropsten (while Mainnet <-> Ropsten message exchange works fine).
Issue for message exchange between Rinkeby <-> Mainnet/Ropsten created in #5188 separately. 

Closing this one.

PR #5167, opened 2018-07-13
update testend enodes to use floating IPs too

jakubgs:
Looks like when the fix was done in https://github.com/status-im/status-react/commit/96628caa9ede2f9eca8405b84c792f3134245833 I missed that `testnet` is actually `ropsten` and needs to be updated the same way too.

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -88,15 +88,15 @@
   {:testnet {"mailserver-a" {:id      "mailserver-a" ;mail-01.do-ams3.eth.beta
                              :name    "Status mailserver A"
                              :password inbox-password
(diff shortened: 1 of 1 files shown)


jeluard (approved):

pedropombeiro (approved):

flexsurfer (approved):

Serhy:
Looks good for me!
Each mailserver under Mainnet and Ropsten (from A to F) - works and offline messages are received under any of mailesevers (from 1-1 and public chats).
Autotest has been rechecked and confirmed as false failed (already under the fixing)
 
Also, Rinkeby mailservers has been checked as well and confirmed that #3752 is not an issue anymore (i could message under Rinkeby and offline inboxing works for Rinkeby as well). There is a separate issue though, is that message sent in Rinkeby network could be received by the contact who is only under Rinkeby (unlike to that user under Mainnet could exchange messages fine with the user who is under Ropsten). 
This one is ready for merge. Thanks a lot @jakubgs

PR #5189, opened 2018-07-16
[#4848] ui, i18n: Replace currency indicator with ~

### LINKED ISSUE #4848: Remove currency indicator from wallet main screen (top left) and replace it with [~] sign
### Description

*Type*: Feature

*Summary*: Since we have currency indicator in the assets section, we dont need to have both currency and currency symbol shown at the top ( for ex $ and USD)

#### Expected behavior

- Currency symbol is removed from the top left so only USD is shown (if USD is selected as main currency)
- Currency symbol is replaced with `~` sign

image

#### Actual behavior

image

### Reproduction
- Open Status
- Tap Wallet

### Additional Information
* Status version: develop

gnl:
This fixes #4848 – remove currency indicator from wallet main screen (top left) and replace it with [~] sign

The `format-currency` utility function in the `i18n` namespace is changed from a single-arity to a multi-arity function with an optional third-parameter – `currency-symbol?` – which is true by default and determines whether the currency symbol ends up in the formatted result.

When it's set to false, the NumberFormatSymbols map is modified to remove the currency symbol placeholder in the format pattern. The `~` is then added in the subscription.

The view remains unchanged.


flexsurfer:
hey @gnl thanks for contribution. could you provide some description, what problem this PR solves?

goranjovic:
Also, especially if this is a bounty PR, make sure to include `fixes #bounty-issue-number` - otherwise, the bounty won't be marked as fulfilled even if we merge it.

gnl:
@flexsurfer Will do and I'll also clean up those nested `def`s after I've made sure they aren't used anywhere else.

@goranjovic Do you mean in the PR title or the commit message?

jeluard:
@gnl In the PR comment itself

jeluard (approved):

gnl:
Guys, can you maybe trigger another CI check+build? I doubt that the recent failure is due to an issue with the code – the difference between the previous commit and this one is the content of a single regex and the whole thing builds and runs just fine on my machine.

goranjovic (approved):

gnl:
Squashed.

PR #5203, opened 2018-07-17
Merge desktop code into develop branch

vkjr:
fixes #5083

### Summary:

This idea behind this PR is to merge mobile and desktop code and have it in the same branch. `desktop_develop_merge` branch contains code from `desktop` rebased on `develop`.

In most files code difference handled by `(if platform/desktop? ...)` statements. 

`vector_icons.cljs` works quite differently for desktop. It doesn't use `react-native-svg` library and able to display svg directly in `Image` component (thanks to underlying Qt).

`js_dependencies.cljs` also has a lot of changes between desktop and mobile versions. Mostly because not all 3rdparty dependencies are implemented and needed by desktop build.

The most problematic are files: `VERSION`, `package.json` and `package-lock.json`. They are different for mobile and desktop and `(if desktop?)` doesn't work there...
I saved different versions of these files in `mobile_files` and `desktop_files` folders. 
If you want to build code of mobile app, you should invoke `scripts/prepare-for-platform.sh mobile`. `scripts/prepare-for-platform.sh desktop` for desktop build.
That script will create in root folder links pointing to `package.json`, `VERSION`, `package-lock.json` in `mobile_files` or `desktop_files` directory.

### Review notes (optional):
Here are steps to run `ios` app from the code:

```
scripts/prepare-for-platform.sh mobile
npm install
mvn -f modules/react-native-status/ios/RCTStatus dependency:unpack
ln -sf './node_modules/re-natal/index.js' './re-natal'
(In terminal 1:) make startdev-ios-simulator
(In terminal 2:) react-native run-ios
```

Here are steps to run `desktop` app from the code:

```
scripts/prepare-for-platform.sh desktop
npm install;
ln -sf './node_modules/re-natal/index.js' './re-natal'
./re-natal use-figwheel;
./re-natal enable-source-maps;
(In terminal 1:) lein figwheel-repl desktop (wait until compilation finished)
(In terminal 2:) npm start
(In terminal 3:) ./ubuntu-server.js
(In terminal 3:) react-native run-desktop
```

### Steps to test:
- Open mobile or desktop Status
- Check that it works correctly
- Step 3, etc.

status: ready

PR #5209, opened 2018-07-17
🚀 re-natal completely removed from the project and replaced with clj-rn

alwx:
In #core-developer we recently decided to get rid of `re-natal`, because we now have an amazing `clj-rn` that can easily be used instead. 

This PR replaces `re-natal` with `clj-rn`.

Status: ready.


scripts/run-osx
```diff
@@ -60,13 +60,13 @@ device=$(/Applications/Genymotion\ Shell.app/Contents/MacOS/genyshell -c "device
 open -a /Applications/Genymotion.app/Contents/MacOS/player.app --args --vm-name "${device##*| }"
 fi
 
-# Install deps, prepare for genymotion and figwheel
-lein deps && npm install && ./re-natal use-android-device "${device_type}" && ./re-natal use-figwheel && npm install
+# Install deps
+lein deps && npm install
```
jeluard: Is the lein bit necessary?

alwx: it's likely that we can remove it. I will check it

dmitryn (approved):

dmitryn:
I wonder how this will play with https://github.com/status-im/status-react/pull/5203 where `re-natal` is still a dependency.

alwx:
@dmitryn interesting question :) From what I see now, it can easily be removed there, so I don't think it's a real problem. I wrote a message there anyway.

vitvly (approved):

vitvly:
Please correct me if I'm wrong: temporarily, taking into account #5203, clj-rn and re-natal can coexist safely, right? The `prepare-for-platform.sh` script from #5203 would just need to restore `.re-natal` for desktop builds to work, not interfering at all with mobile builds.

jeluard:
@siphiuel Goal is to simplify our builds. What do you use to build desktop? Keeping it for desktop only is definitely an option!

vkjr:
@vlasov, @jeluard, I was thinking about the same. It looks like this PR is ready to be merged. But `clj-rn` doesn't support `desktop` platform yet. So possible way to go is to still build desktop with re-natal while mobile platforms built with `clj-rn` until latest extended.

PR #5215, opened 2018-07-18
bug #5214 - allows empty recipient only in dapp transactions (for con…

### LINKED ISSUE #5214: Can proceed to "Sign in" screen without specifying recipient
### Description

*Type*: Bug
*Summary*: you can proceed to "Sign In" screen if you are specifying any asset and amount only, without a recipient.
If asset is ETH -> intrinsic gas too low
If asset is not ETH -> null is not an object (evaluating 'e.substring') error

#### Expected behavior
cannot proceed when "Choose Recipient" is empty
#### Actual behavior
Scenario 1: 
intrinsic gas too low
intris

Scenario 2:
image

### Reproduction

- Open Status
- Sign in with user who has some ETH and STT
- Open Wallet
- Tap on Send transaction
*Scenario1:* 
- specify amount of ETH
*Scenario2:* 
- select another asset (i.e., STT)
- enter valid amount
*For both scenarios:*
- tap on Sign Transaction screen
- enter password 
- tap on Sign Transaction

### Additional Information
* Status version: release 61 (0.9.23 (6212))
* TF session (scenario1, then scenario 2): https://app.testfairy.com/projects/4803622-status/builds/8528545/sessions/4396589731/?accessToken=MD4-n/AieO6SESdEgQJX8OMoY6o

goranjovic:
fixes #5214 

### Summary:

Recipient field in Send is supposed to be optional only for dapp transactions. This is to allow contract creation, in which case recipient doesn't exist. This PR fixes an error in validation conditions that allowed it for regular Wallet Send as well.

### Steps to test:
- see #5214
-also try creating a contract from Simple Dapp

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/send/views.cljs
+++ b/src/status_im/ui/screens/wallet/send/views.cljs
@@ -76,7 +76,7 @@
 (defn- sign-enabled? [amount-error to amount modal?]
   (and
    (nil? amount-error)
-   (or modal? (not (nil? to)) (not= to "")) ;;NOTE(goranjovic) - contract creation will have empty `to`
+   (or modal? (not (empty? to))) ;;NOTE(goranjovic) - contract creation will have empty `to`
    (not (nil? amount))))
 
 ;; "Sign Later" and "Sign Transaction >" buttons


src/status_im/ui/screens/wallet/send/views.cljs
```diff
@@ -73,10 +73,13 @@
       (i18n/label sign-label)
       [vector-icons/icon :icons/forward {:color :white}]]]))
 
+(defn- empty-to? [to]
+  (and (not (nil? to)) (not= to "")))
```
dmitryn: `(and (not (nil? to)) (not= to ""))` same as `(not (empty? to))` ?

dmitryn (approved):

flexsurfer (approved):

src/status_im/ui/screens/wallet/send/views.cljs
flexsurfer: "+"

goranjovic: Hm... almost. `(not (empty? to))` would also pass if `to` is say a vector. Still, I agree it's a better readibility and that's what we need here more.

yenda: also isn't it a `not-empty-to?` function rather than `empty-to?`

dmitryn: @goranjovic right, for empty vector results are not the same. But this function is not going to take vector as i understand. Otherwise, it doesn't make sense to compare it to empty string.

goranjovic: Yeah, that makes sense. Good catch btw

goranjovic:
changed as per comments

mandrigin (approved):

yenda (approved):

janherich (approved):

lukaszfryc:
@goranjovic works as expected now. I tested transactions (ETH, STT), deploying a contract and signing message in the test dapp.

PR #5256, opened 2018-07-22
Modal wallet

flexsurfer:
wallet-modal

image

### DIFF
--- a/src/status_im/ui/screens/wallet/transactions/views.cljs
+++ b/src/status_im/ui/screens/wallet/transactions/views.cljs
@@ -43,20 +43,20 @@
 
 (defn- transaction-type->icon [k]
   (case k
-    :inbound                (transaction-icon :icons/arrow-left (colors/alpha colors/green 0.2) colors/green)
-    :outbound               (transaction-icon :icons/arrow-right (colors/alpha colors/blue 0.1) colors/blue)
-    :failed                 (transaction-icon :icons/exclamation-mark colors/gray-light colors/red)
-    (:postponed :pending)   (transaction-icon :icons/arrow-right colors/gray-light colors/gray)
+    :inbound (transaction-icon :icons/arrow-left (colors/alpha colors/green 0.2) colors/green)
+    :outbound (transaction-icon :icons/arrow-right (colors/alpha colors/blue 0.1) colors/blue)
+    :failed (transaction-icon :icons/exclamation-mark colors/gray-light colors/red)
+    (:postponed :pending) (transaction-icon :icons/arrow-right colors/gray-light colors/gray)
     (throw (str "Unknown transaction type: " k))))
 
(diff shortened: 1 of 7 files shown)


rachelhamlin:
@flexsurfer This is really cool. I'm wondering if we should consider a redesign of the browser so it's more conventional. Navigation & refresh buttons at the top. wdyt about wallet icon in upper right @denis-sharypin?

denis-sharypin (approved):

goranjovic (approved):

src/status_im/ui/screens/wallet/transactions/views.cljs
```diff
@@ -194,9 +195,9 @@
   ([label props-value]
    (details-list-row label props-value nil))
   ([label props-value extra-props-value]
-   (let [[props value]             (if (string? props-value)
-                                     [nil props-value]
-                                     props-value)
+   (let [[props value] (if (string? props-value)
```
yenda, line 198: we don't align anymore ?

flexsurfer: i think we never align in `let`?

yenda: when did we stop ? afaik we've always done it (not that I like it tbh).

asemiankevich:
@flexsurfer a question - if I enable collectibles they are show in modal , however they are not clickable ( can't open details and see all the cats i have). Is it expected?

UPD: actually there is a bug

- open status on mainnet with collectibles
- enable them in wallet
- open dapp
- open wallet modal
- tap any collectible (for ex CryptoStrikers)
- note that collectibles screen is not shown
- tap cross
- note that now you can see collectibles
- tap cross 
- error appears

image


```
TypeError: re-find must match against a string.

This error is located at:
    in browser
    in reagent50
    in RCTView
    in main
    in root
    in reagent1
    in RCTView
    in RCTView
    in t

https://app.testfairy.com/projects/4803590-status/builds/8536969/sessions/4397150579/?accessToken=-1-YGOGfrkP40Q1gf47sYz6mQcU

flexsurfer:
good catch @asemiankevich , thanks, fixed!

asemiankevich:
Looks good to me.


1. can access wallet in public chat. 
2. can access wallet  modal in dapps
3. can copy/share wallet address from wallet modal
4. can access history from wallet modal
5. can see newly added assets in wallet modal
6. can not access collectibles from wallet modal (collectibles are not clickable)
7. txn screen closes and main screen shown when closing modal wallet and open again
8. balance is updated when open wallet modal (without pull-to-refresh)
9. transaction list is updated when open wallet modal (without pull-to-refresh)
10. pull-to-refresh gesture is present both wallet modal and transactions screen
11. can switch between history and wallet screen in wallet modal
12. back device button closes the wallet modal (android)
13. cross button closes the wallet modal (both ios and android)
14. wallet onboarding is not shown when accessing wallet modal directly (after recover or create account)

Issue #566, opened 2016-12-20
Phone number registration adds UK country code before other country codes


richbodo:
Steps to Reproduce:

    I installed the alpha, and when it got to the phone number setup, enter a phone number.
    My phone number was in the format 1 234 567 8901
    (Design Note: All 1s for format choices is a super hard thing to read)

Result:

    I never got an SMS.
    It did say "Thanks, we've sent you a text message..."
    So I assume the format was incorrect or the SMS doesn't actually go out.
    When I look at my profile, I initially saw no phone number.
    30 minutes later, on second look, I see a phone number of the form
    44 1 234 567 8901
    So it added a UK country code to my number, before the US code!

Environment:

    im.status.ethereum-49c604.apk
    I'm using a nexus 6 running Android 6.0.1 at security patch level: 5 Nov 2016
    And I'm using google project fi as my provider.

carlbennetts:
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/110
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/108
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/107
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/102
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/97
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/92
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/86

jarradh:
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/120

Jas0n-F:
https://instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/183

Jas0n-F:
I realise this is closed - but this user is still reporting not receiving the code. 

https://dashboard.instabug.com/applications/status-10bb8cbc-15f6-44f0-8c7a-db478dc34d7d/beta/bugs/231?status=New

jarradh:
[...]

PR #5264, opened 2018-07-23
Delete obsolete code

janherich:

Removes obsolete js bot code & API, not needed anymore, since we completely migrated to the cljs implemented commands. 
I figured that in the event we would want to expose commands functionality to 3rd parties, we would do it via extensions and in complete different way then previously, so it has no point to keep this obsolete code in the codebase, as it can only confuse new developers (and increase app size).

### Testing notes (optional):
Only very superficial testing (app starts and you are able to send/receive message) is needed, I almost added `ship-manual-qa` label here, but since there are some minor changes in the startup sequence, super basic quick testing is probably safer.

status: ready


goranjovic (approved):
:fire:

yenda (approved):

flexsurfer (approved):

rasom (approved):

PR #5268, opened 2018-07-23
Webview loading indicator

flexsurfer:
Webview loading indicator


goranjovic (approved):

jeluard (approved):

yenda (approved):

lukaszfryc:
@flexsurfer works fine 👍I tested loading, refreshing and offline scenarios.

One thing I noticed is that spinner does not hide when loading error apeared:


And, here is a case when the page loading has just started and I went offline. The screen is white but I'd expect to see either the loading error or spinner (if we still try to load the page). But, this case happens on develop as well:

flexsurfer:
good catch with error @lukaszfryc , thanks

PR #5272, opened 2018-07-24
Fix fastlane builds

yenda:
the file with the version number has been moved to subdirectory and a symlink
to it is created, but after the version number is fetched from it where it used
to be

status: ready


jeluard (approved):

maxhora (approved):
@yenda nice catch! thank you!

jakubgs (approved):

vkjr (approved):

lukaszfryc:
@yenda do you need QA help here? If not, please add `skip-manual-qa` label and merge

PR #5290, opened 2018-07-25
Allow only one wallet transactions sync running

dmitryn:
### Summary:

Periodic wallet transactions sync was added in https://github.com/status-im/status-react/pull/5221 which is started each time user signs that lead to multiple syncs running in parallel.

Add flag to control that only one sync is running.

status: ready

### DIFF
--- a/src/status_im/ui/screens/events.cljs
+++ b/src/status_im/ui/screens/events.cljs
@@ -332,7 +332,7 @@
                          [:update-wallet]
                          [:update-transactions]
                          (when platform/mobile? [:get-fcm-token])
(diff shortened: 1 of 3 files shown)


goranjovic (approved):

flexsurfer (approved):

jeluard (approved):

Serhy:
Looks good! As before, transactions sync happens and if there is a list of unconfirmed transactions in chat - they are sync with wallet fine.
Autotests confirmed as failed due to known issues (test_request_and_receive_tokens_in_1_1_chat manually re-tested and confirmed as OK too)

PR #5305, opened 2018-07-26
bug #5265 - transaction history populated only with own transactions

### LINKED ISSUE #5265: Transactions from other account on the same app may appear in transaction history
### User Story

As a user I want correct transaction history shown in all my accounts

### Description

If user has two accounts on the same phone, it is possible that transaction history update is requested, then in the mean time user logs out and logs back in as a different account. Received transactions should not be added to the history of the other account.

*Type*: Bug

#### Expected behavior

Only transactions from the current account are shown

#### Actual behavior

Transactions from the other accounts on the same phone may be shown

### Reproduction

1. Have two accounts A and B
2. Log in to A and create some transactions
3. Open Transaction History and pull to refresh
4. Quickly log out and log in as B
5. Open Transaction history and wait to see A's transactions

Note: it may be easier to reproduce with slow/throttled network

### Solution

Filter out any transaction history responses not related to the current account

goranjovic:
fixes #5265

### Summary:

It was possible for transaction history entries to appear in other accounts on the same device if user logs out and back in (to another account) quickly.

This PR fixes the issue by double checking the account in the callback.

### Steps to test:
see #5265

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/events.cljs
+++ b/src/status_im/ui/screens/wallet/events.cljs
@@ -73,7 +73,7 @@
  (fn [{:keys [web3 chain account-id token-addresses success-event error-event]}]
    (transactions/get-transactions chain
                                   account-id
-                                  #(re-frame/dispatch [success-event %])
+                                  #(re-frame/dispatch [success-event % account-id])
                                   #(re-frame/dispatch [error-event %]))
    (doseq [direction [:inbound :outbound]]
      (erc20/get-token-transactions web3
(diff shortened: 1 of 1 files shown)


yenda (approved):

mandrigin (approved):

flexsurfer:
maybe better to "kill" all http handlers on logout, because it can be not only for transactions

jeluard:
Right, as @flexsurfer mentioned, can this be addressed more systemically?

goranjovic:
@flexsurfer @jeluard Ok, I like the idea. It's maybe too defensive, but I'd still prefer to keep the check even in that case.

goranjovic:
@jeluard @flexsurfer Created a separate issue for the general solution

jeluard (approved):

asemiankevich:
i can't repro https://github.com/status-im/status-react/issues/5265 in this PR

PR #5312, opened 2018-07-26
[TESTS] Added DApp public key, transactions filtering, deep link tests

yevh-berdnyk:
status: ready


test/appium/tests/atomic/account_management/test_wallet_management.py
```diff
@@ -146,3 +146,49 @@ def test_deny_camera_access_scanning_wallet_adders(self):
...
+
+        transaction_history.filters_button.click()
+        for filter_name in 'Outgoing', 'Incoming':
+            transaction_history.filter_checkbox(filter_name).click()
+        wallet_view.done_button.click()
+        if not outgoing_transaction.is_element_displayed():
+            self.errors.append('Outgoing transaction is not shown after selecting outgoing filter')
+        if incoming_transaction.is_element_displayed():
+            self.errors.append('Incoming transaction is shown after selecting outgoing filter')
+        self.verify_no_errors()
```
churik: I would also like to add here test for "Failed" transaction. To simulate failed transaction with "Out of gas" error I can give you recipient address (expired contract, transaction always will end up with same error)

test/appium/tests/atomic/account_management/test_wallet_management.py
```diff
@@ -146,3 +146,49 @@ def test_deny_camera_access_scanning_wallet_adders(self):
...
+        sign_in_view = SignInView(self.driver)
+        sign_in_view.create_user()
+        wallet_view = sign_in_view.wallet_button.click()
+        wallet_view.set_up_wallet()
+        address = wallet_view.get_wallet_address()
+
+        # get incoming transaction
+        self.network_api.get_donate(address=address)
+
+        # get outgoing transaction
```
churik: Is it reusable somewhere? Just wondering - maybe it makes sense to wrap it as separate method (i.e. `send_transaction_from_wallet` with parameters `amount`, `recipient`, `asset`) - action before `send_transaction.sign_transaction(common_password)`

test/appium/tests/atomic/account_management/test_wallet_management.py
```diff
@@ -146,3 +146,49 @@ def test_deny_camera_access_scanning_wallet_adders(self):
         send_transaction.chose_recipient_button.click()
         send_transaction.scan_qr_code_button.click()
         send_transaction.deny_button.wait_for_visibility_of_element(2)
+
+    @marks.testrail_id(3730)
+    def test_filter_transactions_history(self):
```
Serhy, line 151: I thought, why not use some existing account here for this test? With the pre-set transactions. 
It's fine to have like it is now, but i'd say it might be faster to recover account and navigate right to Transaction History screen for checks.

yevh-berdnyk: makes sense, I'll redo it

test/appium/tests/atomic/account_management/test_wallet_management.py
yevh-berdnyk: we use it in different places, but with different steps or sequence of steps - like sending transaction to a contact or with entering an address, selecting recipient before or after setting an amount etc. that's why it's not gathered in a separate method

Serhy (approved):

churik (approved):

Issue #4515, opened 2018-05-29
Can't select assets i have values for


asemiankevich:
### User Story

As a user I want to have possibility to select tokens I have a positive balance for. Like there is a list of 100 tokens and it is difficult for me to remember how many tokens I bought and which of them especially when I am playing with the app. So if I want to see tokens with positive balance the only way to do this is to go to wallet - send - tap asset.  But i want to have a possibility to sort this out with Wallet - manage assets. 

#### Expected behavior
When I want to manage assets I want a feature that allows me to display only assets I have a value > than 0 so I can sort it out quickly.

#### Actual behavior
If I want to see tokens with + balance the only way to do this is to go to wallet - send -asset. 

### Additional Information
* Status version: nightly may 29
* Operating System: Android, iOS

PR #5355, opened 2018-08-01
feature #4842 - initially autoconfiguring visible tokens based on bal…

### LINKED ISSUE #4842: Scan for a user's existing tokens
### User Story

As a user, I want my Status wallet to display all the tokens I own by default, so that I don't have to go looking for them. 

### Description

*Type*: Feature

*Summary*: The wallet supports a long list of tokens but only displays ETH and SNT by default. All other assets must be manually toggled on in the `Manage Assets` list.

This can be confusing for users who make an exchange inside Status, or recover another account with additional assets.

It also reduces the visibility of collectible support. 

#### Expected behavior
Status detects the tokens held by an Ethereum address.

All of the tokens I hold appear in my wallet without any action on my part. 

#### Actual behavior
I need to open the Assets list and toggle on each asset to see them in my wallet.

goranjovic:
fixes #4842

### Summary:

If the user recovers account where they have lots of various collectibles, they'd see only ETH and SNT after the recovery and until they manually enable all the other tokens. This is both tedious and may be a bit scary to see all the tokens seemingly lost after recovery.

This PR checks the balance of all tokens supported by Status during the Wallet Onboarding and automatically enables those with non-zero balance.

For starters, we'll do this only initially - as a setup helper. Eventually, it may be nice to have a special option to autoconfig tokens on-demand.

### Review notes:

This PR also contains very limited `events -> models` refactoring, restricted specifically to `wallet.settings` package. The reason to include this refactoring in this PR was to remove circular dependencies. The reason not to carry out the same kind of refactoring in other touched namespaces was to limit the scope of the PR.

### Steps to test:
- Recover an account with some tokens already on balance (ideally both tokens and collectibles)
- See if they are displayed in main wallet screen

status: ready

### DIFF
--- a/src/status_im/ui/screens/accounts/events.cljs
+++ b/src/status_im/ui/screens/accounts/events.cljs
@@ -4,7 +4,6 @@
             [status-im.native-module.core :as status]
             [status-im.utils.types :refer [json->clj]]
             [status-im.utils.identicon :refer [identicon]]
-            [status-im.utils.random :as random]
             [clojure.string :as str]
             [status-im.i18n :as i18n]
             [status-im.utils.config :as config]
@@ -17,12 +16,11 @@
             [status-im.utils.gfycat.core :refer [generate-gfy]]
             [status-im.utils.hex :as utils.hex]
             [status-im.constants :as constants]
-            [status-im.transport.message.core :as transport]
             status-im.ui.screens.accounts.create.navigation
-            [status-im.chat.models :as chat.models]
             [status-im.ui.screens.accounts.utils :as accounts.utils]
             [status-im.data-store.accounts :as accounts-store]
-            [status-im.ui.screens.navigation :as navigation]))
+            [status-im.ui.screens.navigation :as navigation]
+            [status-im.ui.screens.wallet.settings.models :as wallet.settings.models]))
 
 ;;;; COFX
 
@@ -173,4 +171,5 @@
    (handlers-macro/merge-fx
     cofx
     (wallet-set-up-passed db modal?)
+    (wallet.settings.models/wallet-autoconfig-tokens)
     (accounts.utils/account-update {:wallet-set-up-passed? true}))))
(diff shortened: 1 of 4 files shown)


jeluard (approved):

flexsurfer (approved):

janherich (approved):

asemiankevich:
@goranjovic i think then i can close this one? https://github.com/status-im/status-react/issues/4515 @rachelhamlin was faster than me in this question (even my issue is older! 😅 )

asemiankevich:
@goranjovic  do I understand correctly that the only scenario should work in this PR:

- [x] recover account with assets and collectibles -> complete onboarding on wallet tab -> check if assets and collectibles (those that user owns) are shown

those scenarios **won't work** because the trigger is based on onboarding completion event, right?  (In short words - automatic adding / removing asset when it has + 0 / 0 balance )

- [ ] recover account with assets -> complete onboarding on wallet tab -> sell some asset (full amount) and check if it is removed from portfolio after txn is mined
- [ ] recover account with assets -> complete onboarding on wallet tab -> buy new asset and check if it is removed from portfolio after txn is mined
- [ ] recover with collectibles -> complete onboarding on wallet tab -> gift any collectibles (full amount) and check if it is removed from portfolio after txn is mined
- [ ] recover account without collectibles -> complete onboarding on wallet tab -> get any new collectibles (buy or get as a gift from another account) and check if it is removed from portfolio after txn is mined
- [ ] same for new accounts

if it is so ( above mentioned is not supposed to work in this branch) - is it technically possible to implement them (in a separate branch)?

asemiankevich:
UPD talked to Goran during sync call last night and confirmed the only working scenario here is the top one described above. everything else is out of scope for this PR.

goranjovic:
Yep, this one was specifically to address the scenario where people recover their accounts and then freak out when they don't see any tokens except eth and snt.

rachelhamlin:
@asemiankevich close it (#4515)! 👍

PR #5380, opened 2018-08-04
refactor - extracted all model functions from profile events

goranjovic:
Light Profile refactoring

### Summary:

Extracted model functions from profile events. Proposing naming convention where model functions with side effects end with `!` (the functions used for re-frame effects, as opposed to handlers).

### Testing notes (optional):

No behavior changes - everything should work the same. Only light testing of Profile tab and chat profile is needed. 

status: ready

### DIFF
--- a/src/status_im/ui/screens/profile/events.cljs
+++ b/src/status_im/ui/screens/profile/events.cljs
@@ -1,125 +1,69 @@
 (ns status-im.ui.screens.profile.events
-  (:require [clojure.spec.alpha :as spec]
-            [re-frame.core :as re-frame]
-            [status-im.ui.components.react :as react]
-            [status-im.chat.constants :as chat-const]
+  (:require [re-frame.core :as re-frame]
             [status-im.ui.screens.profile.navigation]
-            [status-im.ui.screens.accounts.utils :as accounts.utils]
-            [status-im.chat.events :as chat-events]
-            [status-im.chat.commands.core :as commands]
             [status-im.utils.handlers :as handlers]
-            [status-im.utils.handlers-macro :as handlers-macro]
-            [status-im.utils.image-processing :refer [img->base64]]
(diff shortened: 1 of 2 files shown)


janherich:
Looks good, I also like the `!` naming convention, only I don't think such functions actually belongs to the model namespace, they should be rather defined in event namespaces (or utils when they are shared).

goranjovic:
@janherich I agree that they don't quite belong with models, but not sure about keeping them with events either, exactly because of the case when they need to be shared, as you mentioned. In fact, all the side effect functions from this PR are pretty generic and probably should be in some shared namespace. 

I'm going through accounts package next (in another PR), let's see how things look there.

src/status_im/ui/screens/profile/events.cljs
```diff
@@ -1,125 +1,69 @@
...
+  (:require [re-frame.core :as re-frame]
             [status-im.ui.screens.profile.navigation]
-            [status-im.ui.screens.accounts.utils :as accounts.utils]
-            [status-im.chat.events :as chat-events]
-            [status-im.chat.commands.core :as commands]
             [status-im.utils.handlers :as handlers]
-            [status-im.utils.handlers-macro :as handlers-macro]
-            [status-im.utils.image-processing :refer [img->base64]]
-            [taoensso.timbre :as log]))
+            [status-im.ui.screens.profile.models :as models]))
```
cammellos: I would call this `models.profile` or at least add `profile` in the name, and also I would not add `-profile` to the methods in the namespace, so in the namespace you have `(defn save ...)` and when using it outside you have: `(models.profile/save ...)`.

Basically I subscribe to these guidlines https://leanpub.com/elementsofclojure

But just a minor point

cammellos (approved):

janherich (approved):

PR #5408, opened 2018-08-07
fix mailserver config #3913

### LINKED ISSUE #3913: Fix mailserver configuration change
When the mailserver configuration is changed:
- make sure that  the previous mailserver is disconnected from peers (will involve json_rpc API)
- the new one is added (should already be done by calling initialize-offline-inbox)
- do not logout  when saving mailserver config

huanghaoyun:
fix #3913

fix mailserver config


yenda:
@huanghaoyun thanks for contributing ! It looks like there is a conflict with your PR, could you resolve it ?
Also the issue also includes disconnecting from the previous mailserver, could you add that to the PR ? Thanks

751895405:
I will handle conflict issue and also I will check the previous mailserver connection state

huanghaoyun:
@yenda there is no connection between mobile client and mailserver ,I  think i should check status-go code to see if there is a method for remove peers .

PR #5427, opened 2018-08-08
fix git tag push credentials, needs to use token instead of pass

jakubgs:
For some reason GitHub stopped accepting the account password when pushing via https, so we need to use the token. `status-im-auto` is a new automation account I've created for Jenkins because I couldn't get accee to `jenkins-status-im`.

### DIFF
--- a/ci/Jenkinsfile.upload_release_android
+++ b/ci/Jenkinsfile.upload_release_android
@@ -55,7 +55,7 @@ timeout(90) {
         stage('Tag Build') {
           withCredentials([[
             $class: 'UsernamePasswordMultiBinding',
-            credentialsId: 'jenkins-status-im',
+            credentialsId: 'status-im-auto',
             usernameVariable: 'GIT_USER',
             passwordVariable: 'GIT_PASS'
           ]]) {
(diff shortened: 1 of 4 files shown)


mandrigin (approved):

ci/Jenkinsfile.upload_release_android
```diff
@@ -55,7 +55,7 @@ timeout(90) {
         stage('Tag Build') {
           withCredentials([[
             $class: 'UsernamePasswordMultiBinding',
-            credentialsId: 'jenkins-status-im',
```
mandrigin, line 58: I believe there is one more place in this file to be changed

mandrigin (changes requested):

ci/Jenkinsfile.upload_release_android
jakubgs: actually, that step shouldn't be there, the `build_no.sh` script does that.

jakubgs: removed it

mandrigin: then this step is also in `upload_ios`

jakubgs: oh right, fixed

maxhora (approved):

mandrigin (approved):

janherich (approved):

PR #5434, opened 2018-08-09
bug #5356 - added snackbar for wallet errors, removed ATT from defaul…

### LINKED ISSUE #5356: 0  fiat amount despite having tokens in Wallet
### Description

*Type*: Bug

*Summary*: Wallet showed $0.00 due to 504 cryptocompare api error. Maybe we can retry the call? Or at least call the API whenever Wallet tab is opened.

### Reproduction

- 

### Additional Information
* Status version: 0.9.24

goranjovic:
fixes #5356 

### Summary:

A number of our users have (silent) problems fetching prices from cryptocompare api - the effect is that total portfolio value in Wallet is shown as zero even if there is some balance in the account. This PR adds the error message, so that in the case of error users would at least be aware of it, rather than to think that they really have no valuable assets.

This PR also removes ATT from the list of default tokens on ropsten (replaced with HND) and adds a migration to remove it from any persisted user settings. This is to remove the false positive error (unable to get token balance) that would appear in the error area on every upgraded app account.

### Testing notes (optional):

- To properly test the main problem, we'll need to block cryptocompare API and then try to use the app.
- There should *not* be an error saying that token balances are not updated (this was triggered by some ATT leftovers)

### Steps to test:
- Block cryptocompare API
- Open wallet and pull to refresh
- There should be an error like in the screenshot

Also, to check balance update failure, e.g. because of connectivity loss:
- Disable wifi/internet
- Open wallet and pull to refresh
- There should be a different error saying balance couldn't be updated

status: ready


jeluard:
It would be nice to have a generic way to handle errors.

goranjovic:
@lukaszfryc https://status-im.slack.com/archives/CA7R1RK0C/p1533819548000308

<blockquote><div><strong><a href="https://status-im.slack.com/archives/CA7R1RK0C/p1533819548000308">Slack</a></strong></div></blockquote>

lukaszfryc:
@goranjovic:

1. The error that appears when cryptocompare api is blocked works great 👍I checked different scenarios with blocking or mocking API responses, etc. Great thing is that the balance refreshes automatically when wallet tab is opened.

2. I don't see error about connectivity loss. When I kill the app, turn off my wifi and log in I see "..." but no error:


Btw, do you know how I can get some ATT (https://etherscan.io/token/0x887834D3b8D450B6bAB109c252Df3DA286d73CE4) so I could test the migration?

<blockquote><div><strong><a href="https://etherscan.io/token/0x887834D3b8D450B6bAB109c252Df3DA286d73CE4">Atmatrix (ATT) Token Tracker</a></strong></div><div>The Ethereum BlockChain Explorer, API and Analytics Platform</div></blockquote>

goranjovic:
@lukaszfryc 

1. awesome

2. weird, when I tried it I'd get the error that balance couldn't be updated. The three dots are only supposed to be there while balances and prices are being updated, so it's possible that the request needs to time out before the error is shown. Can you also try disabling the network while you use the app and then either pull to refresh or switch tabs?

3. You can't - we removed ATT a while ago (but didn't clean up). That's why every version between then and now had errors about ATT in logs. If you don't see an error message, it means it works. If you can access logs, you can double check by searching for ATT.

goranjovic:
@jeluard @rasom @flexsurfer Please review :)

lukaszfryc:
> weird, when I tried it I'd get the error that balance couldn't be updated. The three dots are only supposed to be there while balances and prices are being updated, so it's possible that the request needs to time out before the error is shown. Can you also try disabling the network while you use the app and then either pull to refresh or switch tabs?

I did that and in this case it shows my latest balance but still, there is no error. Does it work on your side?

lukaszfryc:
@goranjovic I found one more issue. When you go offline, log out and create new account, then go to wallet, the app freezes forever after you pass the wallet onboarding. Maybe the cryptocompare request we sent blocks it.

goranjovic:
@lukaszfryc 

> I did that and in this case it shows my latest balance but still, there is no error. Does it work on your side?

Yes, I get the error for whichever service is the first to fail. 

Re other issue: also getting the error - nothing blocked

lukaszfryc:
Let me check it again

goranjovic:
@lukaszfryc Don't bother. I rechecked it later and did see the same behavior you did. Apparently it works ok on emulator, not on real device. The problem seems to be in detection of offline state. 

Since it's also like that on develop and PR fixes the main problem with cryptocompare api being unavailable while really online, I vote that we continue with this one and fix the offline bug separately.

lukaszfryc:
@goranjovic yes, I can also replicate them on develop. I'll create separate issues for these things. Please mention this snackbar in the changelog and free to merge 👍

rasom (approved):

src/status_im/data_store/realm/schemas/base/v8/account.cljs
```diff
@@ -0,0 +1,28 @@
...
+  #{:ATT})
+
+(def reader (transit/reader :json))
+(def writer (transit/writer :json))
+
+(defn serialize [o] (transit/write writer o))
+(defn deserialize [o] (try (transit/read reader o) (catch :default e nil)))
+
+(defn migration [old-realm new-realm]
+  (log/debug "migrating accounts schema v4")
```
cammellos: should not be `v8`?

cammellos (approved):

src/status_im/data_store/realm/schemas/base/v8/account.cljs
goranjovic: gotcha

goranjovic:
Updated changelog, fixed copy paste typos in migration comments

flexsurfer (approved):

PR #5454, opened 2018-08-10
Update desktop dependencies

dmitryn:
### Summary:

Update desktop dependencies introduced by https://github.com/status-im/status-react/pull/4997

status: ready

### DIFF
--- a/src/status_im/utils/notifications.cljs
+++ b/src/status_im/utils/notifications.cljs
@@ -4,7 +4,6 @@
             [status-im.utils.handlers-macro :as handlers-macro]
             [status-im.utils.handlers :as handlers]
             [status-im.react-native.js-dependencies :as rn]
-            [status-im.utils.platform :as platform]
             [status-im.ui.components.react :refer [copy-to-clipboard]]
             [taoensso.timbre :as log]
             [status-im.utils.platform :as platform]))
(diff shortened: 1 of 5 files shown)


maxhora (changes requested):
just referencing `react-native-firebase` for Desktop allows to pass JS package bundling step, but on runtime app fails on startup with error:
```
Debug: ubuntu-server err:  "RNFirebase core module was not found natively on Android, ensure you have correctly added the RNFirebase and Firebase gradle dependencies to your `android/app/build.gradle` file.\r\n\r\n See http://invertase.link/android for the android setup guide."

Debug: ubuntu-server err:  "Module AppRegistry is not a registered callable module (calling runApplication)"
```
This happens, because `react-native-firebase` module doesn't support Desktop platform. As quick workaround, possibly, will work adding of conditional check that `react-native-firebase` module related source code is executing on Mobile platform currently.

maxhora:
Have pushed commit which fixes desktop builds by dirty workaround to frame notifications related code with `when-not platform/desktop?` condition.

yenda:
What would be a more long term fix @MaxRis ?

maxhora:
@yenda 
Probably, more long term workaround can be moving of `utils.notifications` namespace under `mobile` namespace and redefining of `utils.notifications` under `desktop` namespace. (similar like we do with `js-dependencies`)

For Desktop we currently also working on desktop notifications feature ( in-progress PR is there https://github.com/status-im/status-react/pull/5330 ), but I'm not sure what common implementation details are in both cases. On desktop we have defined our own `react-native-desktop-notifications` react-native native module to handle notifications background logic.

yenda:
@MaxRis if you are satisfied with the fix can you remove your change request and squash your commit so that we can merge ?

src/status_im/utils/notifications.cljs
```diff
@@ -4,183 +4,185 @@
...
+   (fn [_ _]
+     (re-frame/dispatch [:show-mainnet-is-default-alert])))
+
+  (handlers/register-handler-fx
+   :request-notifications-denied
+   (fn [_ _]
+     (re-frame/dispatch [:show-mainnet-is-default-alert])))
+
+  (def firebase (object/get rn/react-native-firebase "default"))
+  ;(def firebase "")
```
yenda: comment can be removed

yenda (approved):

src/status_im/utils/notifications.cljs
maxhora: @yenda oh, overlooked it, thanks!

maxhora (approved):

maxhora:
@yenda done! thank you

PR #5480, opened 2018-08-13
iOS 💛 fastlane

mandrigin:
This PR switches all iOS builds to using [fastlane](https://fastlane.tools).

### 1. Code signing via `match`
Instead of manual signing setup, everything is changed to [codesigning guide](https://codesigning.guide) using [fastlane match](https://docs.fastlane.tools/actions/match/).

New certificates are created and stored in a private git repository.
The repo is: https://github.com/status-im/ios-certificates/
Admins are (as of now): @jakubgs, @rasom, @mandrigin

That makes provisioning of the new CI machines much simpler.

### 2. Move all the iOS build steps to `fastlane`

3 lanes are created for iOS:
`ios adhoc` — for the builds to be distributed via diawi. It creates an ad-hoc signed app that only devices with registered UDIDs can install. It is used for PRs and Release Candidate builds.

`ios release` — creates a "proper" release build and uploads it to TestFlight. This build is to be distributed along external testing and this lane is only used in `upload_release_ios` Jenkins job.

`ios nightly` — creates & uploads a testflight build and also builds a second .ipa to be distributed via diawi.

### 3. Keychain unlock

Keychain unlock is moved to fastlane too. See [this line](https://github.com/status-im/status-react/pull/5480/files#diff-56c5021ef06ac6cea807ae53269b6f4fR15)

### 4. Misc

- `Jenkinsfile.nightly` was removed because it wasn't used. `Jenkinsfile.nightly_fastlane` is still here

- Xcode project is changed to use manual signing and match certificates

- builds for mobile are now running on the Jenkins machines with tag `fastlane` instead of `macos<whatever>`

status: ready


rasom (approved):

goranjovic (approved):

mandrigin:
@annadanchenko @lukaszfryc I won't wait for the autotests, because no code changes are here

fastlane/Fastfile
```diff
@@ -6,25 +6,112 @@
...
-# Uncomment the line if you want fastlane to automatically update itself
-# update_fastlane
+# There are a few env variables defined in the .env file in
+# this directory (fastlane/.env)
+
+
+# unlocks keychain if KEYCHAIN_PASSWORD variable is present
+# (to be used on CI machines)
+def unlock_keychain_if_needed
+    if ENV["KEYCHAIN_PASSWORD"]
```
nemoeslovo, line 16: @mandrigin you might want to handle the case when env variable isn't setup, and send some meaningful message as `UI.error("...")`, since build will fail anyway with locked keychain.

I really like the way artsy treat open source contributors when they give suggestions on how to build app without being part of organization as part of the flow https://github.com/artsy/eidolon/blob/e90a1ffd38d73179a6fe2174587dcd98988efc48/fastlane/Fastfile#L37

mandrigin: @nemoeslovo yeah, good point!
theoretically the keychain shouldn't be locked for non ssh builds, but it'd be nice to handle this case too

mandrigin: they also support this magic value in their code for OSS contributors, right?

fastlane/Fastfile
```diff
@@ -6,25 +6,112 @@
...
       ipa: "status_appstore/StatusIm.ipa"
     )
-    slack(message: "New release build uploaded to iTunes Connect")
+    slack(message: "New release build uploaded to TestFlight")
   end
-  desc "Remove inactive TestFlight users"
+
+  desc "`fastlane ios clean` - remove inactive TestFlight users"
+  desc "uses custom plugin, installed via"
+  desc "`sudo get install fastlane-plugin-clean_testflight_testers`"
```
chadyj, line 114: Plugin [install docs](https://github.com/KrauseFx/fastlane-plugin-clean_testflight_testers/blob/master/README.md) mentions `fastlane add_plugin clean_testflight_testers`

chadyj:
Very cool @mandrigin.  What's next for Fastlane?

fastlane/Fastfile
mandrigin: it is the same but probably a better syntax
feel free to fix it an submit a PR :-P

mandrigin:
@chadyj diawi, I guess and github comments later. android after that.

fastlane/Fastfile
chadyj: I tried `sudo get install fastlane-plugin-clean_testflight_testers` but got `command not found` :(

PR #5494, opened 2018-08-14
feature #5310 - improved validation in account recovery screen

### LINKED ISSUE #5310: Validate words when recovery phrase is entered and show warning if they do not match expected words
### Description

*Type*: Feature

*Summary*: Other account will be recovered if user makes a typo when entering a word in Recovery phrase. We can introduce validation of the entered words to see if they are found in https://github.com/status-im/status-go/blob/52cdcf8f0f1127f5ee694f39086a9e4df54174e6/extkeys/mnemonic.go#L97 If not then we can show an extra warning to the user or highlight "wrong" words but give user a final choice to decide if it's ok to restore account with "unexpected" words or not

Several users reported "lost funds" as wrong account was recovered. We can help to prevent it with such validation (even minimalistic UI would be better then seeing 0 instead of 5 ETH on recovered account)

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-go">status-im/status-go</a></strong></div><div>status-go - The Status module that consumes go-ethereum</div></blockquote>

### LINKED ISSUE #5353: Adding a Recovery phrase and password on a fresh new install throws Error and freezes app with a blank screen
### Description

*Type*: Bug

*Summary*: When installing for the first time the application on an Android device and selecting the "I already have an account" option from the first welcome screen, if we insert a random Recovery phrase and password and hit "Sign in", the application throws:
 "Error: No item 1 in vector of length 1 
   This error is located at: 
   in home
   in reagent24
   in RCTView
   in RCTView
..."

whatsapp image 2018-08-01 at 12 21 38

Then, the welcome dialog is shown and when hitting ok, the dialog disappears and what is left is a blank screen that does nothing. To continue using the app, it is necessary to hit the back button to exit the app, and relaunch it. 

This issue was reproducible with three different Android devices. Not been able to check on iOS devices.

The issue was reproduced 3 times/3 attempts.

#### Expected behavior
The expected behavior was to either have a new account created from the random data introduced or a dialog saying that this account does not exist.

#### Actual behavior
Error is thrown and an unresponsive blank screen is shown.

### Reproduction
To reproduce the bug it is necessary to install the application on a device that has never had the application installed before. Once installed, open the application and select "I ALREADY HAVE AN ACCOUNT". On the "Sign in to another account" screen, insert a random recovery phrase and password. In this case it was used: "t t t t t t t t t t t t" "popopopopo". Hit "Sign in", and then the Error Dialog will pop up. Here is a video with all the sequence:
https://drive.google.com/file/d/1FS1QxLrJxWNaXtsrAW8wruTE5YNRVmZ2/view?usp=sharing

### Additional Information
* Status version: 0.9.23 July 23
* Operating System: Android
* Devices: Nexus 5 with Android 6.0.1, Zebra TC200J with Android 7.1.2, Moto X Play with Android 7.1.1
* Might be related to bug https://github.com/status-im/status-react/issues/5232

<blockquote><div>Google Drive</div><div><strong><a href="https://drive.google.com/file/d/1FS1QxLrJxWNaXtsrAW8wruTE5YNRVmZ2/view?usp=sharing&usp=embed_facebook">error_status_app.mp4</a></strong></div></blockquote>

goranjovic:
fixes #5310 
also fixes #5353 

### Summary:

Adds better validation for recovery phrase and password, with error tooltips that depend on error type.

If user enters a syntactically valid recovery phrase that contains words not present in our dictionary (e.g. if the passphrase was generated in a different client app), we show a non-blocking warning and double check if that's all right with a dialog confirmation.

### Review notes (optional):
Also lightly refactored accounts.recover screen and added a bunch of unit tests

### Steps to test:
- Try to recover account with a valid phrase
- Try some edge cases (empty fields, invalid phrases with numbers, symbols, too many/few words)
- Try using words outside our dictionary 

status: ready


goranjovic:
screenshot-2018-08-14_09 42 45 145
screenshot-2018-08-14_12 01 49 393
screenshot-2018-08-14_12 02 03 196

goranjovic:
@obi2020 @denis-sharypin Please double check the copies. My own suggestion is to repeat the reason (words might be misspelled) in the dialog box, because "you could create a new account" is a bit scary without context. What do you think?

src/status_im/ui/screens/accounts/recover/models.cljs
```diff
@@ -29,6 +47,27 @@
 
 ;;;; Handlers
 
+(defn set-phrase [recovery-phrase {:keys [db]}]
+  {:db (-> db
+           (assoc-in [:accounts/recover :passphrase] (string/lower-case recovery-phrase))
```
flexsurfer: i think in such cases better to use `update`
```clojure
(update :accounts/recover assoc :passphrase (string/lower-case recovery-phrase)
                                :passphrase-valid? (not (check-phrase-errors recovery-phrase)
```

src/status_im/ui/screens/accounts/recover/models.cljs
```diff
@@ -55,6 +94,16 @@
    :dispatch-n [[:navigate-to-clean :home]
                 [:request-notifications]]})
 
-(defn recover-account [masked-passphrase password {:keys [db]}]
-  {:db (assoc-in db [:accounts/recover :processing] true)
-   :recover-account-fx [masked-passphrase password]})
+(defn recover-account [{:keys [db]}]
+  (let [{:keys [password passphrase]} (:accounts/recover db)]
+    {:db                 (assoc-in db [:accounts/recover :processing] true)
```
flexsurfer: `:processing?`

flexsurfer (approved):

denis-sharypin:
@goranjovic looks good to me! I'm only curious how do you trigger this checks if button Sign in is not active
image

goranjovic:
@denis-sharypin Right now they trigger when field loses focus.

Another alternative is to trigger while user types, but that's a bit annoying if you say get the error tooltip just because you haven't finished typing it.

src/status_im/ui/screens/accounts/recover/models.cljs
goranjovic: I don't really have a preference there, so sure why not.

gravityblast:
@goranjovic we also have the same list of words in `status-go` + lists for other languages, which are not commonly used but are in the specs. 
If it makes sense we can open an issue for the future to call `status-go` from `status-react` for the validation maybe.

goranjovic:
@pilu Good idea. I certainly don't like the duplication (even if it's never supposed to change)

denis-sharypin:
@goranjovic and what are the circumstances to show this? I'm just documenting this in a design files
screenshot 2018-08-14 12 56 15

goranjovic:
@denis-sharypin 

The exact steps were as follows:

1. tapped the phrase field
2. types that text in
3. tapped the password field. since phrase field lost focus, it displayed the tooltip

denis-sharypin:
@goranjovic  alright, but when do you show `Recovery phrase is invalid` or `Some words might be misspelled` ?

goranjovic:
@denis-sharypin ah ok, it's like this:

1.) if field is empty or has only whitespace chars - "Field is required"
2.) if it's not empty, but has wrong number of words or invalid characters (numbers, symbols, etc.) then "Recovery phrase is invalid"
3.) if word count and chars are ok, and it contains some words not on our list, then it's "Some words might be misspelled"

The last one is just a warning, not an error. It will not block Sign in, just pop out the dialog. Maybe we should have a special style for warning, e.g. yellow?

denis-sharypin:
@goranjovic amazing thanks!

obi2020:
Sorry, I guess this has already been committed, but we can update it in the next content update. I think "misspelled" is not quite the right way to talk about this. Here's how I might edit the confirmation dialogue:

> **Is the phrase correct?**
> 
> If you enter the wrong words, you might create a new account instead of recovering an old one.
> 
> [No, review] [Yes, confirm]
> or
> [Review phrase] [Confirm phrase]

In your version and mine we say a new account "may" or "might" be created. Is it definite that an account is created? Could someone do this deliberately in order to choose their own recovery phrase? @goranjovic

goranjovic:
@obi2020 That's a good point - "may" or "might" are not good here. If you enter different words, you "will" create a new account.

I guess "create" is also a bit awkward here. You never actually create accounts, you merely access them with a unique key combination only you have. 

Also, it's possible that someone may do this on purpose, but I assume in that case they'd know what they're doing and confirm. The real problem is when you do this by accident.

chadyj:
@goranjovic Should we also duplicate the password field here to help users? It is possible to do a typo that goes unnoticed since we don't validate which would then lock the user out of their account.  WDYT?

goranjovic:
Addressed review comments, rebased against develop and updated copies as per suggestions:

screenshot-2018-08-16_12 13 20 527

janherich (approved):

janherich (approved):

asemiankevich:
@goranjovic maybe a silly one - why do we still allow user to recover account with the words out of our dictionary? And if i am wrong and we do not - the error pop up is confusing then 😮

goranjovic:
@asemiankevich Because it could be a valid passphrase, but generated in a different wallet that maybe doesn't use the same dictionary

asemiankevich:
issue 1: seems https://github.com/status-im/status-react/issues/4105 is here again 
```02:17 I/ReactNativeJS: NAME: "TypeError"
02:17 I/ReactNativeJS: MESSAGE: null is not an object (evaluating 'u.cljs$core$IFn$_invoke$arity$1')
02:17 I/ReactNativeJS: 
02:17 I/ReactNativeJS: index.android.bundle:12:6870293
02:17 I/ReactNativeJS: index.android.bundle:12:2446822
02:17 I/ReactNativeJS: index.android.bundle:12:2446614
02:17 I/ReactNativeJS: invoke_interceptor_fn@index.android.bundle:12:2372120
02:17 I/ReactNativeJS: invoke_interceptors@index.android.bundle:12:2372421
02:17 I/ReactNativeJS: execute@index.android.bundle:12:2374870
02:17 I/ReactNativeJS: handle@index.android.bundle:12:2383283
02:17 I/ReactNativeJS: re_frame$router$IEventQueue$_process_1st_event_in_queue$arity$1@index.android.bundle:12:2417832
02:17 I/ReactNativeJS: re_frame$router$IEventQueue$_run_queue$arity$1@index.android.bundle:12:2400171
02:17 I/ReactNativeJS: index.android.bundle:12:2405923
02:17 I/ReactNativeJS: re_frame$router$IEventQueue$_fsm_trigger$arity$3@index.android.bundle:12:2410134
02:17 I/ReactNativeJS: index.android.bundle:12:2402689
02:17 I/ReactNativeJS: index.android.bundle:118:1369
02:17 I/ReactNativeJS: x@index.android.bundle:118:643
02:17 I/ReactNativeJS: callTimers@index.android.bundle:118:2788
02:17 I/ReactNativeJS: value@index.android.bundle:113:3582
02:17 I/ReactNativeJS: index.android.bundle:113:1067
02:17 I/ReactNativeJS: value@index.android.bundle:113:3009
02:17 I/ReactNativeJS: value@index.android.bundle:113:1039
02:17 I/ReactNativeJS: [native code]
02:17 I/ReactNativeJS: 
02:17 I/ReactNativeJS: ***

```

https://app.testfairy.com/projects/4803622-status/builds/8563578/sessions/4398962905/?accessToken=lVSBwZCk8-egKKbjnRFfPju5knY

issue 2: https://github.com/status-im/status-react/issues/5395 when i was logging out (known issue)

issue 3: when i repro issue 1 , and i have words for recovery out of my dictionary - confirm dialog is shown as many times as i multitap sign in. Then as a result - i hardly can log out. Please see TF 
https://app.testfairy.com/projects/4803622-status/builds/8563578/sessions/4398963338/?accessToken=QTD9JgiNJE9MtYUHheUcRddXy40

issue 4: for some reason after installing the build and recover account my network is Ropsten

goranjovic:
@asemiankevich  updated to address issues 1 and 3 - also incidentally fixed #5353 with this

asemiankevich:
Devices under test: 
- Huawei P10 Lite with Android 8 on board
- iPhone 7 running iOS 11.4

Tests performed: 

- [x] recovered known account with valid phrase on both devices
- [x] recovered metamask / mycrypto account with valid phrase on both devices
- [x] empty (or spaced only) `Recovery phrase` and `Password` fields are highlighted with a red tooltip `Required fied` when i tap out of the field

image

- [x] too short / too long (less than 12 or more than 12 words) `Recovery phrase` field results in red tooltip saying `Recovery phrase is invalid`

image


image

- [x] words with numbers/special characters are treated as `Recovery phrase is invalid` red tooltip

- [x] words out of the dictionary (https://github.com/status-im/status-go/blob/52cdcf8f0f1127f5ee694f39086a9e4df54174e6/extkeys/mnemonic.go#L97) result in `Some words might be misspelled`. This does not prevent me from recovering account so when i tap `Sign in` next tooltip is shown ` Is the phrase correct?` with Cancel and Confirm button

image


Issues: 

- multutapping reproted above is fixed , as well as https://github.com/status-im/status-react/issues/5353 (at least i cant repro it now)
- **NEW** whitespaces are not trimmed (or new passphrase is not taken into account when trimmed). So if i  enter `[space] word1..last_word[space]' or inside the phrase - it recovers the account i dont expect to recover
- not an issue but a note that we dont use any other language for recovery other than english however we have several in the dictionary. That could be revisited later.


<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-go">status-im/status-go</a></strong></div><div>status-go - The Status module that consumes go-ethereum</div></blockquote>

goranjovic:
@asemiankevich pushed the fix for trimming - great catch

asemiankevich:
spaces trimming issue is fixed now, looks good to me!

PR #5497, opened 2018-08-14
[#5260] Implement web3 provider Opt-in access

### LINKED ISSUE #5260: Implement web3 provider Opt-in access
### User Story

As a user, I want to make sure my privacy is not at risk when using a DApp.
As a developer, I want to follow community standards.

### Description

*Type*: Feature

Implement web3 provider Opt-in access as defined by [EIP1102](https://eips.ethereum.org/EIPS/eip-1102). This pattern follows the pattern used in our recently merged status web DApp API support.

UX will be augmented so that a user can decide if web3 access should be granted to the DApp (cc @denis-sharypin)

As a first step old behavior will be kept for backwards compatibility: all currently working DApps should still work fine.
Status user can enable privacy aware web3 access in it's profile (cc @denis-sharypin). Once enabled `web3` object won't be injected by default (and old style DApps will be broken).

As this change forces DApps devs to update their DApps we want to have clear documentation detailing steps required to move to this pattern.

### UX considerations

DApps will probably mostly request access at startup. Some access during the lifecycle of a DApp might be expected too.
Permissions are granted on a per DApp level and persistent (i.e. once granted it won't be re-asked). No global permission handling are considered at this stage.
A panel accessible from user profile allows to quickly browse and modify permissions granted per DApp.

flexsurfer:
fixes #5260

we'll remove web3 object after 2 November, in current PR we just introduce transition phase for  [EIP1102](https://eips.ethereum.org/EIPS/eip-1102) 

aug-14-2018 15-15-05


src/status_im/ui/screens/browser/events.cljs
```diff
@@ -23,22 +22,11 @@
 (re-frame/reg-fx
  :send-to-bridge-fx
  (fn [[permissions-allowed webview]]
+   (println :send-to-bridge-fx permissions-allowed (keys permissions-allowed))
```
jeluard: Leftover?

src/status_im/ui/screens/browser/permissions/views.cljs
```diff
@@ -0,0 +1,82 @@
...
+         [react/view {:margin-left 3 :margin-right 3}
+          [react/view styles/dot]]
+         [react/view {:margin-right 3}
+          [react/view styles/dot]]
+         [react/view styles/permissions-panel-ok-icon-container
+          [icons/icon :icons/ok styles/permissions-panel-ok-ico]]
+         [react/view {:margin-left 3 :margin-right 3}
+          [react/view styles/dot]]
+         [react/view {:margin-right 3}
+          [react/view styles/dot]]
```
jeluard, line 49: Maybe some function could be added here for more visibility.
And move styles in `styles/` ns?

src/status_im/ui/screens/browser/permissions/views.cljs
```diff
@@ -0,0 +1,82 @@
...
+                                                                   (:permissions-data params)])
+                                    :label    (i18n/label :t/allow)}]]
+        ;; TODO (andrey) will be in next PR
+        #_[react/view {:flex-direction :row :margin-top 19}
+           [icons/icon :icons/settings {:color colors/blue}]
+           [react/text {:style styles/permissions-panel-permissions-label}
+            (i18n/label :t/manage-permissions)]]]])))
+
+(defn permissions-anim-panel [browser show-permission]
+  (let [timeout (atom nil)
```
jeluard: Can you add a comment to the function?
Also it might be something we want to abstract at some point.

src/status_im/ui/screens/browser/views.cljs
```diff
@@ -74,31 +90,38 @@
   (let [domain-name (nth (re-find #"^\w+://(www\.)?([^/:]+)" url) 2)]
     (get (:inject-js browser-config) domain-name)))
 
+(defn navigation [webview browser can-go-back? can-go-forward?]
+  [react/view styles/toolbar
+   [react/touchable-highlight {:on-press            #(re-frame/dispatch [:browser-nav-back browser])
+                               :disabled            (not can-go-back?)
+                               :style               (when-not can-go-back? styles/disabled-button)
+                               :accessibility-label :previou-page-button}
```
jeluard, line 98: Typo: `previous`

src/status_im/ui/screens/browser/permissions/views.cljs
flexsurfer: tbh i agree here with desktop team , i don't see any reason to move all maps to styles, we can keep "small" ones and move only "big" maps, tbh it takes a lot of time to generate style names and move them to styles ns

src/status_im/ui/screens/browser/events.cljs
flexsurfer: thanks

src/status_im/ui/screens/browser/views.cljs
flexsurfer: thanks

src/status_im/ui/screens/browser/permissions/views.cljs
flexsurfer: yes sure

src/status_im/ui/screens/browser/permissions/views.cljs
jeluard: Downside is that it prevents reuse. With proper components it shouldn't be a pb. We have way too much similar styles everywhere.

But why not? Should we discuss that with others?

yenda:
Correct me if I'm wrong but the web3 object is already initialized in web3-init.js so this is more of a transition phase where you can ask for the ETHEREUM-PROVIDER even though you don't need to ?

flexsurfer:
@yenda yes, right,  we'll remove web3 object after 2 November, in current PR we just provide an ability to request it, transition phase yes

src/status_im/ui/screens/browser/permissions/views.cljs
flexsurfer: let's discuss for sure , but i believe @vkjr discussed couple of times, and tbh in 99% styles used only once

src/status_im/ui/screens/browser/views.cljs
flexsurfer: btw this is not my change, we have this long time @antdanchenko ?

goranjovic (approved):

src/status_im/ui/screens/browser/views.cljs
antdanchenko: well, we are using the same typo https://github.com/status-im/status-react/blob/746d580e3406cceb0707ddb30652203789801607/test/appium/views/web_views/base_web_view.py#L31 

please, change it in both places if you are going to change it in this PR
otherwise I can change it in one of my PRs

asemiankevich:
i have tested this with Test Dapp both options:
- request public key
- request web3 access

i also have checked that access to profile is requested when accessing ENS Usernames dapp (ropsten)

image

I can't check on mainnet as there are no suitable dapps .

I checked as a regression that i can send txn to dapp (fails now for known bug). Waiting for e2e results.

yenda (approved):

asemiankevich (approved):

PR #5515, opened 2018-08-15
Remove superfluous override of `DataDir` at startup for Android

pedropombeiro:
This tiny PR removes setting `DataDir` on Android, since that data is already present in the JSON (iOS already does the same, not setting/overriding `DataDir` a second time).

status: ready


goranjovic (approved):

vitvly:
@jeluard @alwx @rasom Are we good to go here?

jeluard:
@siphiuel Feel free to review ! :)

jeluard (approved):

vitvly (approved):

pedropombeiro:
Apparently this section will be refactored soon, so no point in changing it now.

PR #5528, opened 2018-08-16
[CI] Renamed apk for e2e tests

antdanchenko:
probot is triggering e2e tests with parameters, one of them is apk e.g. (0000.apk, where 0000 is a PR number)
builds which were named in the old way won't be found by tests in case if we will change the setting in porbot to use new names

so it worth to be renamed back


mandrigin (approved):
lgtm

yevh-berdnyk (approved):

PR #5531, opened 2018-08-16
Copy on stack status-go signal data

### LINKED ISSUE #5483: Update status-go to 0.10.0 breaks compatibility with Desktop
According to https://github.com/status-im/status-react/pull/5319

Update to status-go release 0.10.0 breaks possibility to login in Status Desktop or create new account.
It errors out with following log:

```

Debug: call of RCTStatus::emitSignalEvent ... signal:  {"type":"node.crashed","event":{"error":"node is already running"}}

Debug: RCTStatus::startNode StartNode result:  {"error":""}
```
GIT_TAG for status-go 0.10.0 is a6d69eba

maxhora:
fixes #5483

### Summary:

Resolves issues appeared with status-go 0.10.0 upgrade during login

Additional info:
App hanging after user log in, because geth node is crashed, because there was attempt to start it twice, because after first try to start geth node "is-node-started" variable was not set in CLJS, because the signal with event about successful "node.started" was not propagated at first time from RCTStatus.cpp native module, because signal event data of type `char *` was corrupted after passed to main app thread from status-go thread, because something was changed internally in status-go (before it worked fine). Current solution is to copy signal event data by keeping it withing QString type by value on stack.

### Steps to test:
- Open Status
- Login into account

status: ready

### DIFF
--- a/modules/react-native-status/desktop/rctstatus.h
+++ b/modules/react-native-status/desktop/rctstatus.h
@@ -56,15 +56,15 @@ public:
     Q_INVOKABLE void getDeviceUUID(double callbackId);
 
     Q_INVOKABLE static bool JSCEnabled();
-    Q_INVOKABLE static void jailSignalEventCallback(const char* signal);
+    Q_INVOKABLE static void statusGoEventCallback(const char* event);
 
-    void emitSignalEvent(const char* signal);
+    void emitStatusGoEvent(QString event);
 
 Q_SIGNALS:
-    void jailSignalEvent(const char* signal);
+    void statusGoEvent(QString event);
 
 private Q_SLOTS:
-    void onJailSignalEvent(const char* signal);
+    void onStatusGoEvent(QString event);
 
 private:
     QScopedPointer<RCTStatusPrivate> d_ptr;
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

modules/react-native-status/desktop/rctstatus.h
```diff
@@ -58,13 +58,13 @@ class RCTStatus : public QObject, public ModuleInterface {
     Q_INVOKABLE static bool JSCEnabled();
     Q_INVOKABLE static void jailSignalEventCallback(const char* signal);
 
-    void emitSignalEvent(const char* signal);
+    void emitSignalEvent(QString signal);
 
 Q_SIGNALS:
-    void jailSignalEvent(const char* signal);
+    void jailSignalEvent(QString signal);
```
pedropombeiro: Unrelated: I believe jails have been removed recently, so this shouldn't be necessary anymore.

pedropombeiro (approved):

modules/react-native-status/desktop/rctstatus.h
maxhora: Thanks, will update methods names

vkjr (approved):

churik:
Tested (desktop - Mac Osx
- account is created 
- can log in with existing account
- messages are fetched
- can send message in 1-1 and public chat

Tested (IOS, Android):
- account is created 
- messages are fetched

Issue #55, opened 2016-05-05
if the message begins with ! it should never send as plaintext


0xc1c4da:
https://assembly-vc.slack.com/files/jarradhope/F16BMEP0E/phonecommandtyping.mp4

PR #5536, opened 2018-08-16
Merge desktop and mobile builds into a single job

jakubgs:
This PR does several things:

* Creates `Jenkinsfile.combined` which runs major 4 types of builds in parallel
* Creates `Jenkinsfile.{ios,android,linux,macos}` to run each build as a separate job
* Extracts Jenkins stages from `Jenkinsfile`s into `common.groovy`, `mobile.groovy`, and `desktop.groovy`
* Adds a `cleanup` lane to the iOS Fastlane setup to clean up XCode DerivedData folder
* Adds missing Fastlane dependencies to `Gemfile`

The Jenkins builds can be seen here:
https://ci.status.im/job/status-react/job/combined/
Here's an example build:
TODO
Which in turn triggered those:
TODO
This is not done yet, as I still haven't finished uploading artifacts nor making separate builds for e2e tests in parallel.

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/">All [status-react » combined] [Jenkins]</a></strong></div></blockquote>

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -1,0 +1,72 @@
+pipeline {
+  agent { label 'macos' }
+
+  options {
+    buildDiscarder(logRotator(
+      numToKeepStr: '10',
+      daysToKeepStr: '30',
+      artifactNumToKeepStr: '4',
+    ))
+  }
+  
+  environment {
+    LANG = 'en_US.UTF-8'
+    LANGUAGE = 'en_US.UTF-8'
+    LC_ALL = 'en_US.UTF-8'
+    FASTLANE_DISABLE_COLORS = 1
+    REALM_DISABLE_ANALYTICS = 1
+    ANDROID_HOME = '/usr/local/share/android-sdk'
+    ANDROID_SDK_ROOT = '/usr/local/share/android-sdk'
+    ANDROID_NDK = '/Users/jenkins/android-ndk-r10e'
+    ANDROID_NDK_HOME = '/Users/jenkins/android-ndk-r10e'
+  }
+  
+  stages {
+    stage('Prep') {
+      steps {
+        script {
+          /* Necessary to load methods */
+          mobile = load 'ci/mobile.groovy'
+          mobile.prepDeps()
+        }
+      }
+    }
+    stage('Tests') {
+      steps {
+        script { mobile.runTests() }
+      }
+    }
+    stage('Build') {
+      steps {
+        script { mobile.leinBuild() }
+      }
+    }
+    stage('Compile') {
+      steps {
+        script { apk = mobile.compileAndroid(e2e: params.RUN_E2E) }
+      }
+    }
+    stage('Bundle') {
+      when { expression { !params.RUN_E2E } }
+      steps {
+        script { mobile.bundleAndroid() }
+      }
+    }
+    stage('Archive') {
+      when { expression { !params.RUN_E2E } }
+      steps {
+        script { archiveArtifacts apk }
+      }
+    }
+    stage('Run e2e') {
+      when { expression { params.RUN_E2E } }
+      steps { script {
+        apk = mobile.uploadSauceLabs() 
+        build(
+          job: 'end-to-end-tests/status-app-nightly', wait: false,
+          parameters: [string(name: 'apk', value: "--apk=${apk}")]
+        )
+      } }
+    }
+  }
+}
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -1,0 +1,115 @@
+pipeline {
+  agent { label 'master' }
+
+  options {
+    disableConcurrentBuilds()
+    buildDiscarder(logRotator(
+      numToKeepStr: '10',
+      daysToKeepStr: '30',
+      artifactNumToKeepStr: '10',
+    ))
+  }
+  
+  stages {
+    stage('Tag') {
+      steps { script {
+        common = load('ci/common.groovy')
+        /* to avoid race conditions in parallel builds */
+        print "Build Number: ${common.tagBuild()}"
+      } }
+    }
+    stage('Build') {
+      parallel {
+        stage('MacOS') {
+          steps { script {
+            osx = build('status-react/combined/desktop-macos')
+          } }
+        }
+        stage('Linux') {
+          steps { script {
+            nix = build('status-react/combined/desktop-linux')
+          } }
+        }
+        stage('iOS') {
+          steps { script {
+            ios = build('status-react/combined/mobile-ios')
+          } }
+        }
+        stage('Android') {
+          steps { script {
+            dro = build('status-react/combined/mobile-android')
+          } }
+        }
+        stage('Android e2e') {
+          steps {
+            build('status-react/combined/mobile-android-e2e')
+          }
+        }
+      }
+    }
+    stage('Archive') {
+      steps {
+        sh('rm -f pkg/*')
+        copyArtifacts(
+          projectName: 'status-react/combined/mobile-ios', target: 'pkg',
+          flatten: true, selector: specific("${ios.number}")
+        )
+        copyArtifacts(
+          projectName: 'status-react/combined/mobile-android', target: 'pkg',
+          flatten: true, selector: specific("${dro.number}")
+        )
+        copyArtifacts(
+          projectName: 'status-react/combined/desktop-macos', target: 'pkg',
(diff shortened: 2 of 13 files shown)


Gemfile
```diff
@@ -1,6 +1,8 @@
 source "https://rubygems.org"
 
 gem "fastlane"
+gem "fastlane-plugin-diawi"
```
mandrigin: I think those aren't necessary because later in the file it reads from `Pluginfile` where they exist.

ci/Jenkinsfile.combined
```diff
@@ -0,0 +1,62 @@
...
+    }
+    stage('Notify') {
+      steps {
+        slackSend(
+          message: "Combined Nightly build success! (develop)",
+          color: 'good'
+        )
+      }
+    }
+    //stage('Upload') {
```
mandrigin: emm, why commented out?

ci/mobile.groovy
```diff
@@ -0,0 +1,76 @@
...
+    sh 'pod install'
+  }
+}
+
+def bundleAndroid() {
+  withCredentials([
+    string(credentialsId: "SUPPLY_JSON_KEY_DATA", variable: 'GOOGLE_PLAY_JSON_KEY'),
+    string(credentialsId: "SLACK_URL", variable: 'SLACK_URL')
+  ]) {
+    sh 'bundle exec fastlane android nightly'
```
mandrigin: this one uploads to the play store...

any place where we upload to diawi?

fastlane/Fastfile
```diff
@@ -238,6 +237,10 @@ platform :ios do
       )    
   end
 
+  desc "This fastlane step cleans up XCode DerivedData folder"
+  lane :cleanup do
+    clear_derived_data
```
mandrigin, line 242: 👍

mandrigin (changes requested):
awesome job!
a few questions...

ci/Jenkinsfile.combined
jakubgs: For tests, this pr is wip

maxhora (approved):

Gemfile
jakubgs: I see, thanks

ci/mobile.groovy
jakubgs: yeah, in the `uploadiOS()` method

flexsurfer:
@jakubgs could you please squash commits

jakubgs:
@flexsurfer I will when I merge, right now it helps me review changes.

ci/Jenkinsfile.android
```diff
@@ -0,0 +1,72 @@
...
+    ANDROID_NDK = '/Users/jenkins/android-ndk-r10e'
+    ANDROID_NDK_HOME = '/Users/jenkins/android-ndk-r10e'
+  }
+  
+  stages {
+    stage('Prep') {
+      steps {
+        script {
+          /* Necessary to load methods */
+          mobile = load 'ci/mobile.groovy'
```
adambabik, line 29: Can we do that earlier in the code? We use `mobile` in other stages but declaration and loading happens here.

ci/Jenkinsfile.combined
```diff
@@ -0,0 +1,115 @@
...
+          steps { script {
+            ios = build('status-react/combined/mobile-ios')
+          } }
+        }
+        stage('Android') {
+          steps { script {
+            dro = build('status-react/combined/mobile-android')
+          } }
+        }
+        stage('Android e2e') {
```
adambabik, line 43: It looks like a mix of build stages and one stage running e2e tests. Why is it so?

ci/common.groovy
```diff
@@ -0,0 +1,60 @@
+def installJSDeps(platform) {
+  def attempt = 1
+  def maxAttempts = 10
+  def installed = false
+  /* prepare environment for specific platform build */
+  sh "scripts/prepare-for-platform.sh ${platform}"
+  while (!installed && attempt <= maxAttempts) {
+    println "#${attempt} attempt to install npm deps"
+    sh 'npm install'
+    installed = fileExists('node_modules/web3/index.js')
```
adambabik, line 10: This is fragile. Usual case with npm is that it fails to install **all** dependencies. I would just check `npm install` exit code.

ci/mobile.groovy
```diff
@@ -0,0 +1,102 @@
...
+  def filename = "im.status.ethereum-${GIT_COMMIT.take(6)}-n-fl.apk"
+  def newArtifact = (artifact_dir + filename)
+  sh "cp ${artifact} ${newArtifact}"
+  def uploadSpec = '{ "files": [ { "pattern": "*apk/release/' + filename + '", "target": "nightlies-local" }]}'
+  def buildInfo = server.upload(uploadSpec)
+  return 'http://artifacts.status.im:8081/artifactory/nightlies-local/' + filename
+}
+
+def prepDeps() {
+  sh 'rm -rf node_modules'
```
adambabik, line 17: This can extend the build significantly. I would try to use `npm install` and `npm ls` commands to verify that installed packages have correct versions.

scripts/build_no.sh
```diff
@@ -21,36 +18,53 @@ set -e
...
-# even if the current commit has a tag already, it is normal that the same commit
-# is built multiple times (with different build configurations, for instance),
-# so we increment the build number every time.
+# check if current commit has a build tag
+# since we are building in separate jobs we have to check for a tag
+BUILD_TAG=$(git tag --points-at HEAD | grep -e "$REGEX")
+
+# chech for multiple lines
+if [ 1 -lt $(echo "$BUILD_TAG" | grep -c -) ]; then
+    echo "Commit marked with one than one build tag!" >&2
```
adambabik: `with more than one`

ci/common.groovy
adambabik: I mentioned it in the other comment but `npm ls` also return non-zero exit code if the dependencies differ between what is installed and `package-lock.json`.

ci/Jenkinsfile.android
jakubgs: What do you mean? This is the first reference to `mobile` in that file.

ci/Jenkinsfile.combined
jakubgs: Because Android e2e build is different than the regular stage?

ci/common.groovy
jakubgs: I didn't write it, and did not want to change it here. This PR aleady changes a fuckton of things.

ci/mobile.groovy
jakubgs: That's true, but the purpose of this PR is NOT to fix the build process, it is to fix the Jenkinsfiles and extract steps into re-usable methods. If you want to fix those things, we can do it in a separate PR.

scripts/build_no.sh
jakubgs: yeah

jakubgs:
Seems to have worked fine:
https://ci.status.im/job/status-react/job/combined/job/nightly/55/

I think we can consider this a complete change, since it works. I don't want to make this PR any bigger than it already is. I'll make a separate one once this is merged to:
* Change name of DigitalOcean space(`status-im-desktop.ams3.digitaloceanspaces.com`) to be more generic and reference desktop
* Possibly create a separate Jenkinsfile for PRs, OR use this one as is(the `PUBLISH` variable is an attempt at this
* Possibly improve the `npm install` steps with what @adambabik suggested.

As I said, this PR is about extracting steps into re-usable methods and making it all parallel. I don't want to spin more plates than that here.

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/job/nightly/55/">status-react » combined » nightly #55 [Jenkins]</a></strong></div></blockquote>

ci/Jenkinsfile.android
adambabik: It was a bit misleading to me that `mobile` is used other stages while it looks like it is loaded only for `stage('Prep')` but if it works this way than fine.

ci/Jenkinsfile.combined
adambabik: Ah ok. I did not know that.

adambabik (approved):

ci/Jenkinsfile.android
jakubgs: The trick is that I'm not using `def` for `mobile` variable which makes it global variable available in other stages.

mandrigin (approved):

PR #3889, opened 2018-04-16
issue 3690: Use thread pool instead of new threads on Android

pacamara:
Fixes #3690 : "Use thread pool instead of new threads on Android"

### Summary:
Replaces use of `new Thread()` with a global `StatusThreadPoolExecutor` class per the requirement in issue #3690.

### Steps to test:
General smoke test on the status client, especially account lifecycle operations.

status: ready

PR #5544, opened 2018-08-17
Fixes #5293: Messaging/transacting take too long to be processed if there is a 'Pending...' tx message

### LINKED ISSUE #5293: Messaging/transacting take too long to be processed if there is a 'Pending...' tx message
***Type***: Bug

***Summary***: On Mainnet, once there is a transaction sent in 1-1 chat that will take too much time to be mined (or put into a block) which is also visible as `Pending...` status of `/send` message - further messages or transactions start to take way too long time for processing: 
- spinner below sent message takes > 1m and actual message indeed not received by recipient
- transaction after a tap on 'Sign transaction' button processing up to 2-3 minutes

App UI does not lag, but it's communication lags.

**Note 1:** it might be that https://github.com/status-im/status-react/pull/5221 has an impact on this.
**Note 2:**  Maybe transcations sent from Ropsten also bring this issue, but I could not reproduce it:  even with lowest gas price (0.00000001) transaction mined in <1-2 minutes.

#### Expected behavior
Messages and transactions are sent instantly doesn't matter if there are pending transactions that contact has sent
 
#### Actual behavior
Message and transaction processing takes up to 1-2 minutes if there is a transaction in `Pending...` state that contact sent

### Reproduction

- Open Status on two devices (Android 6 - is a good example for one of device) 
- Recover an account with > 0 ETH balance on Mainnet on Android (Contact1)
- Create a new account on Device 2 (Contact2)
- Add both contacts in the contacts list of each other
- Contact1 sends a transaction to Contact 2 with a small gas price (e.g. gas price = `0.0000001`)
**Actual result:** A `/send` message in 1-1 chat about sent transaction pushed and `Pending...` status is displayed (as Expected!)
- Wait > 2 minutes and confirm that sent transaction is still in `Pending...` state in transction history. I.e. it's has been mined yet. 
- Contact1 (who sent the transaction) sends 10 more messages to Contact2 and pays attention whether spinners below messages disappear in less than 2-3 seconds
- Contact1 sends one more transaction to Contact2 from 1-1 chat (tap `/` -> `send ETH` -> `0.000001` amount -> on Wallet screen enter valid password and tap `Sign transaction`).

### Additional Information
* Status version: Develop nightly build `0.9.23 (6249) a339d7e` (25th of July Jenkins build)
* Operating System: Android 6.0.1, Android 7.0

#### Logs
TF session for Contact1 above: from 02:00 https://app.testfairy.com/projects/4803622-status/builds/8535945/sessions/4397065312/?accessToken=U3EQnCAOggTbqIOmwW5dJr4oGBM

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803622-status/builds/8535945/sessions/4397065312/?accessToken=U3EQnCAOggTbqIOmwW5dJr4oGBM">TestFairy: Status 0.9.23 (6249) @2018</a></strong></div></blockquote>

pacamara:
Fixes #5293: Messaging/transacting take too long to be processed if there is a 'Pending...' tx message

### Summary:

Fixes poor choice of thread pool parameters in #3889 .
Still use an unbounded queue, but increase `corePoolSize` to a generous `100*NUMBER_OF_CORES`.
Allow it to drain by setting `allowCoreThreadTimeOut(true)`. Gives this behaviour:

* Pool initially has zero threads in it.
* New tasks add threads to the pool up to `corePoolSize`.
* Once pool is at `corePoolSize` threads, tasks get queued.
* After 1 second of inactivity, threads are killed. Pool can return to zero size if no tasks.

### Testing notes:

Have written ` StatusThreadPoolExecutorTest` to verify above behaviour. Unfortunately, integrating with react-native build is knarly (there are currently no native android tests), so posted as gist for now: https://gist.github.com/pacamara/9dbbd0a2780a107a37c197a9ec26bd43 . To use just create a dummy android project and add both files.

### Steps to test:

Existing reproduce steps for issue #5293 .

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusThreadPoolExecutor.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusThreadPoolExecutor.java
@@ -2,9 +2,13 @@ package im.status.ethereum.module;
 
 import java.util.concurrent.*;
 
+/** Uses an unbounded queue, but allows timeout of core threads
+ * (modified case 2 in
+ * https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ThreadPoolExecutor.html ) */
 public class StatusThreadPoolExecutor {
     private static final int NUMBER_OF_CORES =
             Runtime.getRuntime().availableProcessors();
+    private static final int THREADS_TO_CORES_RATIO = 100;
     private static final int KEEP_ALIVE_TIME = 1;
     private static final TimeUnit KEEP_ALIVE_TIME_UNIT = TimeUnit.SECONDS;
 
@@ -15,11 +19,14 @@ public class StatusThreadPoolExecutor {
         mQueue = new LinkedBlockingQueue<>();
 
         mThreadPool = new ThreadPoolExecutor(
-            NUMBER_OF_CORES,
-            NUMBER_OF_CORES,
+            THREADS_TO_CORES_RATIO * NUMBER_OF_CORES,
+            THREADS_TO_CORES_RATIO * NUMBER_OF_CORES,
             KEEP_ALIVE_TIME,
             KEEP_ALIVE_TIME_UNIT,
             mQueue);
+
+        // Allow pool to drain
+        mThreadPool.allowCoreThreadTimeOut(true);
     }
 
     /** Pugh singleton */


dmitryn (approved):

pacamara:
`StatusThreadPoolExecutorTest` at https://gist.github.com/pacamara/9dbbd0a2780a107a37c197a9ec26bd43 

<blockquote><div> Gist</div><div><strong><a href="https://gist.github.com/pacamara/9dbbd0a2780a107a37c197a9ec26bd43">StatusThreadPoolExecutorTest</a></strong></div><div>StatusThreadPoolExecutorTest</div></blockquote>

mandrigin (approved):
lgtm

rasom (approved):

janherich (approved):

yenda (approved):

pedropombeiro (approved):
Although I agree that the max number of threads in the pool was needlessly low, I think it's important to understand what has started causing threads to block the threadpool, since it seems things were working until recently even with something as low as 4 threads in the pool, but this work can be done in a separate PR.

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusThreadPoolExecutor.java
```diff
@@ -15,11 +18,14 @@ private StatusThreadPoolExecutor() {
         mQueue = new LinkedBlockingQueue<>();
 
         mThreadPool = new ThreadPoolExecutor(
-            NUMBER_OF_CORES,
-            NUMBER_OF_CORES,
+            100 * NUMBER_OF_CORES,
```
reutsharabani: Shouldn't there be a constant `THREADS_TO_CORES_RATIO` to make it clearer?

pedropombeiro: I guess this could even be set to `Integer.MAX_VALUE`, as mentioned in the docs:
> By setting maximumPoolSize to an essentially unbounded value such as Integer.MAX_VALUE, you allow the pool to accommodate an arbitrary number of concurrent tasks.

reutsharabani: > ... you allow the pool to accommodate an arbitrary number of concurrent tasks.

I'm not sure this is a good idea, but I don't have a good view of the flows using this piece of code and the dangers in allocating so many threads (java threads are pretty heavy).

I like the idea of binding the thread count to the core count.

goranjovic (approved):

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusThreadPoolExecutor.java
pacamara: @reutsharabani @PombeirP  I'll update with `THREADS_TO_CORES_RATIO` later today. My instinct about `Integer.MAX_VALUE` is it would allow a runaway/misbehaving client to have a worse impact on the rest of the OS than if threads are constrained, don't you reckon? If the thread count gets really big the phone is spending all its time context switching and the threads just make things worse...

reutsharabani: @pacamara I completely agree. The thread count should correspond with the system resources. No hardware I know can run `Integer.MAX_VALUE` threads without locking the CPU(s) forever and exhausting the memory (for the per-thread stacks - especially in Java).

I think the core count is a good indicator of the system's capabilities and therefore so the thread count can be reasonably determined by it.

pedropombeiro: Yeah, I agree that `Integer.MAX_VALUE` is not the right solution (that would be negating the advantages of having moved from creating threads manually to using a thread pool), but I'm also not sure that having 800 threads in a 8-core phone is feasible. Ideally we'd know when we screw up for some reason and create some new code that blocks threads for a long time, causing the thread pool to be fully utilized for a long time.

Serhy:
https://i.diawi.com/wCBDVy and https://i.diawi.com/xtGVAg builds has been tested on Android 6, Android 7th, iOS 11.4.1 
and confirmed as fixied the issue. For the account that constantly been affected with the issue - all fine now. 
Having, for instance, 5 pending transaction messages in chat does not have an impact of signing latest transactions, neither on messaging. 
It's also visible by the number of autotests that are passing now (3 failed autotest reviewed, retested, confirmed false-positive issues which are already addressed for a fix).
@PombeirP , @pacamara are we going to make more changes here? Sinse if not we can merge, else, - waiting for update.


<blockquote><div><strong><a href="https://i.diawi.com/wCBDVy">Install - Diawi - Development and In-house Apps Wireless Installation</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://i.diawi.com/xtGVAg">Install - Diawi - Development and In-house Apps Wireless Installation</a></strong></div></blockquote>

pedropombeiro:
@Serhy we're just waiting for the `THREADS_TO_CORES_RATIO` change suggested above.

pacamara:
>we're just waiting for the THREADS_TO_CORES_RATIO change suggested above.

@Serhy @PombeirP @reutsharabani Done! :beers:

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusThreadPoolExecutor.java
pacamara: @PombeirP Agree, going to 100 x cores is possibly too far in the opposite direction now :)  Anyhow it's easily adjustable! Note a very large pool can still have one advantage over unpooled threads which is re-use if you have a lot of short lived threads getting rapidly created.

> Ideally we'd know when we screw up for some reason and create some new code that blocks threads for a long time, causing the thread pool to be fully utilized for a long time.

Yes, maybe needs some thread profiling done? eg. https://software.intel.com/en-us/articles/load-balancing-between-threads

pacamara:
(Repushed with a slight change to a comment only.)

Serhy:
Latest builds tested on iOS 11.4.1, Android 6, Android 7- all good!

PR #5579, opened 2018-08-21
Fix/use groovy methods [skip ci]

jakubgs:
Intoduces a `BUILD_TYPE` variable which can have 3 values:
* `debug` - The type we use for PRs and nightlies
* `e2e` - The type used for end-to-end tests(currently only affects Android)
* `release` - The type that gets uploaded to iTunes and Play Store

Still working on it, but this is the gist of it.


mandrigin (approved):
Great job!

flexsurfer (approved):

PR #5503, opened 2018-08-14
Set `FLEET` in env variable

adambabik:
Allow to specify a fleet in the `env` file and use it in the configuration.

Todo:
- [x] provide Mail Servers for `env.staging`.

Closes #5213

PR #5600, opened 2018-08-22
Fixed desktop api for using fleet

vkjr:
Fixes broken desktop build that uses not the same `status-go` as mobile.

### Summary:

This PR complements #5503. 
`react-native-status` api changed to be used with new `status-go`

### Steps to test:
- Open Status
- Try to login, make sure you don' have startNode error

status: ready

### DIFF
--- a/modules/react-native-status/desktop/CMakeLists.txt
+++ b/modules/react-native-status/desktop/CMakeLists.txt
@@ -37,7 +37,7 @@ ExternalProject_Add(StatusGo_ep
   PREFIX ${StatusGo_PREFIX}
   SOURCE_DIR ${StatusGo_SOURCE_DIR}
   GIT_REPOSITORY https://github.com/status-im/status-go.git
-  GIT_TAG v0.11.0
+  GIT_TAG develop-g19b53030
   BUILD_BYPRODUCTS ${StatusGo_STATIC_LIB}
   CONFIGURE_COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/${CONFIGURE_SCRIPT} ${GO_ROOT_PATH} ${StatusGo_ROOT} ${StatusGo_SOURCE_DIR}
   BUILD_COMMAND ""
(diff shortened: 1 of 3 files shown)


modules/react-native-status/desktop/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ ExternalProject_Add(StatusGo_ep
   PREFIX ${StatusGo_PREFIX}
   SOURCE_DIR ${StatusGo_SOURCE_DIR}
   GIT_REPOSITORY https://github.com/status-im/status-go.git
-  GIT_TAG v0.11.0
+  GIT_TAG develop-g19b53030
```
flexsurfer, line 40: why is it develop again? @adambabik

flexsurfer (approved):

maxhora (approved):

modules/react-native-status/desktop/CMakeLists.txt
maxhora: such tag format, seems, is supported by CMake fine and we can use such unified format both for Mobile and Desktop. But can we avoid in future strange tags like "tags-v0.11.0^0-g4afd9e6c-302"? CMake can't handle them.

adambabik: As far as I know, Status Desktop builds status-go from the source. It means that it should be a git tag or git commit SHA. In this case, `19b53030`.

`develop-g19b53030` is the name of the artifact build for ObjC or Android.

@flexsurfer because there was no release `v0.11.1` yet. I will do that asap.

adambabik: https://github.com/status-im/status-go/releases/tag/v0.12.0 and its `v0.12.0` because changing the binding is a breaking change.

churik:
doesn't affect mobile builds;
desktop is working as expected with new status-go version.

modules/react-native-status/desktop/CMakeLists.txt
vkjr: @adambabik, looks like cmake can understand `develop-g19b53030` as well as sha.

adambabik: I have just checked that and indeed, git can handle this just fine. Surprising but it's good to know :)

PR #5602, opened 2018-08-22
[fix #5599] add semaphore for recurring events

### LINKED ISSUE #5599: Concurrency issue with dispatch-later during account login
### Problem

`:sync-wallet-transaction` and `:check-sync` (only on non-rpc nodes) are self-firing events that `dispatch-later` themselves and form an event-loop.
When the user logs out these events keep firing and they add up because a new event loop starts every time user logs in until the app is killed.
A guard on the event that checks if the event is fired is not a complete solution because if the user logs in fast enough, the event loop is not terminated and a new one will start.
A solution that would keep track of the timeouts that need to be cancelled is not easy to implement because you need to be able to flush the timeouts that have be fired.

yenda:
fix #5599

### Summary:

:sync-wallet-transaction and :sync-state are self firing events that
dispatch themselves later. this commit introduces semaphores to prevent
these event-loops from starting multiple times when the user logs out and
logs in again.

### Testing notes (optional):
 The event at stake Fetches updated data for any unconfirmed transactions or incoming chat transactions missing in wallet and schedule new recurring sync request

### Steps to test:
- Open Status
- Check that unconfirmed transactions or incoming chat transactions are still being shown in the wallet within 15 seconds

status: ready

### DIFF
--- a/src/status_im/protocol/models.cljs
+++ b/src/status_im/protocol/models.cljs
@@ -1,0 +1,40 @@
+(ns status-im.protocol.models
+  (:require [status-im.constants :as constants]
+            [status-im.utils.semaphores :as semaphores]
+            [status-im.utils.ethereum.core :as ethereum]
+            [status-im.utils.handlers-macro :as handlers-macro]))
+
+(defn update-sync-state
+  [{:keys [sync-state sync-data] :as db} error sync]
+  (let [{:keys [highestBlock currentBlock] :as state}
+        (js->clj sync :keywordize-keys true)
+        syncing?  (> (- highestBlock currentBlock) constants/blocks-per-hour)
+        new-state (cond
+                    error :offline
+                    syncing? (if (= sync-state :done)
+                               :pending
+                               :in-progress)
+                    :else (if (or (= sync-state :done)
+                                  (= sync-state :pending))
+                            :done
+                            :synced))]
+    (cond-> db
+      (and (not= sync-data state) (= :in-progress new-state))
+      (assoc :sync-data state)
+      (not= sync-state new-state)
+      (assoc :sync-state new-state))))
+
+(defn check-sync-state
+  [{{:keys [web3] :as db} :db :as cofx}]
+  (if (:account/account db)
+    {::web3-get-syncing web3
+     :dispatch-later    [{:ms 10000 :dispatch [:check-sync-state]}]}
+    (semaphores/free :check-sync-state? cofx)))
+
+(defn start-check-sync-state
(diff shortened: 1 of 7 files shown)


src/status_im/wallet/transactions.cljs
```diff
@@ -88,21 +88,27 @@
         (< sync-interval-ms
            (- (time/timestamp) last-updated-at)))))
 
-(defn set-sync-started [{:keys [db]}]
```
yenda: wallet is flushed upon logout so this is not useful

dmitryn (approved):

goranjovic (approved):

janherich (approved):
Some unit tests exercising the functionality would probably be nice.

yenda:
@janherich @goranjovic @dmitryn I made more changes to fix the issue completely, I introduced a semaphore ns in utils that can be reused for similar types of events

src/status_im/protocol/models.cljs
```diff
@@ -0,0 +1,40 @@
...
+                    error :offline
+                    syncing? (if (= sync-state :done)
+                               :pending
+                               :in-progress)
+                    :else (if (or (= sync-state :done)
+                                  (= sync-state :pending))
+                            :done
+                            :synced))]
+    (cond-> db
+      (when (and (not= sync-data state) (= :in-progress new-state)))
```
dmitryn: no need for `when` in `cond->` conditions

goranjovic (approved):

lukaszfryc:
@yenda on Android, a TX I received in 1-1 chat is in "TX not found state" even after 10 min. What's more, I don't see any transactions in transaction history. Not even the faucet transaction. Wallet screen shows correct amount though. 

iOS works fine so far.

yenda:
@lukaszfryc not related, etherescan API was failing during your test
```
18:31 D/ReactNativeJS: DEBUG [status-im.utils.transactions:60] - HTTP GET https://api-ropsten.etherscan.io/api?module=account&action=txlist&address=0xc205fd0e4f69b868654dfd312014d4d36c46f8cc&startblock=0&endblock=99999999&sort=desc&apikey=DMSI4UAAKUBVGCDMVP3H2STAMSAUV7BYFI?q=json
18:32 D/ReactNativeJS: DEBUG [status-im.utils.handlers:37] - Handling re-frame event:  :update-transactions-fail
18:32 D/ReactNativeJS: DEBUG [status-im.ui.screens.wallet.events:189] - Unable to get transactions:  TypeError: Network request failed
```

lukaszfryc:
@yenda other than that it works fine. Please merge

PR #5617, opened 2018-08-23
Implement "Save Password" functionality for iOS

### LINKED ISSUE #2869: User is frequently logged out

### User Story

As a user, I want to stay logged into the app so that I can easily use it without logging on.

### Description

*Type*: Bug

*Summary*: When using Status it can be very frustrating to be logged out of the app on a regular basis. For example, I can receive a push notification, tap it, and go to read the message but I'm blocked with a login screen. 

#### Expected behavior
As a user expect that the app is always accessible without logging in.
#### Actual behavior
As a user I am frequently logged out. 

### Reproduction
Restart the app and the user is logged out.

### Solution
*Summary*: 

### Additional Information
* Status version: Latest build
* Operating System: Android or iOS

#### Logs

[comment]: # (if on Android please replicate bug whilst running adb logcat)
```
...
```

mandrigin:
Store the application password in iOS Keychain using `react-native-keychain` library that we already use.

The password is stored if you tick the checkbox.
It is removed on logout.
Password is tied to the account address.

### What's left?
- [x] upgrade `react-native-keychain` to a release version (3.0.0)
- [x] fix UI layout for checkbox
- [x] set correct properties for the Keychain (should not be transferable)
- [x] fix unit tests

### What won't be addressed here?
- Android support
   - [ ] create an issue for Android 9.0
   - [ ] create an issue for Android 6.0+
- Encryption with Secure Enclave
   - [ ] create an issue for that (fork the RN keychain module)

partially fixes #2869 

simulator screen shot - iphone 8 plus - 2018-08-23 at 16 18 43

status: ready

### DIFF
--- a/src/status_im/ui/screens/accounts/login/models.cljs
+++ b/src/status_im/ui/screens/accounts/login/models.cljs
@@ -7,6 +7,7 @@
             [status-im.utils.config :as config]
             [status-im.utils.keychain.core :as keychain]
             [status-im.utils.notifications :as notifications]
+            [taoensso.timbre :as log]
             [status-im.utils.platform :as platform]
             [status-im.utils.universal-links.core :as universal-links]))
 
@@ -14,8 +15,10 @@
 
 (defn stop-node! [] (status/stop-node))
 
-(defn login! [address password]
-  (status/login address password #(re-frame/dispatch [:login-handler % address])))
+(defn login! [address password save-password]
+  (status/login address
+                password
+                #(re-frame/dispatch [:login-handler % address password save-password])))
 
 (defn clear-web-data! []
   (status/clear-web-data))
@@ -34,22 +37,27 @@
 
 ;;;; Handlers
 
-(defn open-login [address photo-path name {db :db}]
+(defn navigate-to-login [address photo-path name password {db :db}]
   {:db       (update db
                      :accounts/login assoc
                      :address address
                      :photo-path photo-path
+                     :password password
                      :name name)
+   :can-save-user-password [#(re-frame/dispatch [:set-in [:accounts/login :can-save-password] %])]
    :dispatch [:navigate-to :login]})
 
-(defn wrap-with-login-account-fx [db address password]
+(defn wrap-with-login-account-fx [db address password save-password]
   {:db    db
-   :login [address password]})
+   :login [address password save-password]})
 
-(defn login-account-internal [address password {db :db}]
+(defn open-login [address photo-path name cofx]
+  {:get-user-password [address #(re-frame/dispatch [:do-login address photo-path name %])]})
+
+(defn login-account-internal [address password save-password {db :db}]
   (wrap-with-login-account-fx
    (assoc db :node/after-start nil)
-   address password))
+   address password save-password))
 
 (defn- add-custom-bootnodes [config network all-bootnodes]
   (let [bootnodes (as-> all-bootnodes $
@@ -77,17 +85,17 @@
      :network              network
      :config               config}))
 
-(defn- wrap-with-initialize-geth-fx [db address password]
+(defn- wrap-with-initialize-geth-fx [db address password save-password]
   (let [{:keys [network config]} (get-network-by-address db address)]
     {:initialize-geth-fx config
      :db                 (assoc db
                                 :network network
-                                :node/after-start [:login-account-internal address password])}))
+                                :node/after-start [:login-account-internal address password save-password])}))
 
-(defn start-node [address password {db :db}]
+(defn start-node [address password save-password {db :db}]
   (wrap-with-initialize-geth-fx
    (assoc db :node/after-stop nil)
-   address password))
+   address password save-password))
 
 (defn- wrap-with-stop-node-fx [db address password]
   {:db        (assoc db :node/after-stop [:start-node address password])
@@ -98,7 +106,7 @@
       (and config/bootnodes-settings-enabled?
            use-custom-bootnodes)))
 
-(defn login-account [address password {{:keys [network status-node-started?] :as db} :db}]
+(defn login-account [address password save-password {{:keys [network status-node-started?] :as db} :db}]
   (let [{use-custom-bootnodes :use-custom-bootnodes
          account-network :network} (get-network-by-address db address)
         db'     (-> db
@@ -113,17 +121,21 @@
 
                       :else
                       wrap-with-stop-node-fx)]
-    (wrap-fn db' address password)))
+    (when-not (empty? password)
+      (wrap-fn db' address password save-password))))
 
-(defn login-handler [login-result address {db :db}]
+(defn login-handler [login-result address password save-password {db :db}]
   (let [data    (types/json->clj login-result)
         error   (:error data)
         success (zero? (count error))
         db'     (assoc-in db [:accounts/login :processing] false)]
     (if success
-      {:db             db
-       :clear-web-data nil
-       :change-account [address]}
+      (merge
+       {:db             db
+        :clear-web-data nil
+        :change-account [address]}
+       (when save-password
+         {:save-user-password [address password]}))
       {:db (assoc-in db' [:accounts/login :error] error)})))
 
 (defn change-account-handler [address {{:keys [view-id] :as db} :db :as cofx}]
--- a/src/status_im/ui/screens/events.cljs
+++ b/src/status_im/ui/screens/events.cljs
@@ -5,6 +5,7 @@
             status-im.protocol.handlers
             [status-im.ui.screens.accounts.models :as accounts.models]
             status-im.ui.screens.accounts.login.events
+            [status-im.ui.screens.accounts.login.models :as login]
             status-im.ui.screens.accounts.recover.events
             [status-im.ui.screens.contacts.events :as contacts]
             status-im.ui.screens.add-new.new-chat.events
@@ -214,18 +215,17 @@
                            (re-frame/dispatch [:initialize-app encryption-key]))
    :on-accept           handle-reset-data})
 
-(defn initialize-views [{{:accounts/keys [accounts] :as db} :db}]
-  {:db                                  (if (empty? accounts)
-                                          (assoc db :view-id :intro :navigation-stack (list :intro))
-                                          (let [{:keys [address photo-path name]} (first (sort-by :last-sign-in > (vals accounts)))]
-                                            (-> db
-                                                (assoc :view-id :login
-                                                       :navigation-stack (list :login))
-                                                (update :accounts/login assoc
-                                                        :address address
-                                                        :photo-path photo-path
-                                                        :name name))))
-   :handle-initial-push-notification-fx db})
+(defn initialize-views [cofx]
+  (let [{{:accounts/keys [accounts] :as db} :db} cofx
+        {:keys [address photo-path name]} (first (sort-by :last-sign-in > (vals accounts)))
+        default-fx {:handle-initial-push-notification-fx db}]
+    (if (nil? address)
+      (handlers-macro/merge-fx cofx
+                               default-fx
+                               (navigation/navigate-to-clean :intro))
+      (handlers-macro/merge-fx cofx
+                               default-fx
+                               (login/open-login address photo-path name)))))
 
 (defn initialize-db
   "Initialize db to the initial state"
@@ -289,7 +289,8 @@
  (fn [{:keys [db] :as cofx} _]
    (let [{:transport/keys [chats]} db]
      (handlers-macro/merge-fx cofx
-                              {:dispatch [:initialize-keychain]}
+                              {:dispatch [:initialize-keychain]
+                               :clear-user-password [(get-in db [:account/account :address])]}
                               (navigation/navigate-to-clean nil)
                               (transport/stop-whisper)))))
 
--- a/src/status_im/utils/keychain/core.cljs
+++ b/src/status_im/utils/keychain/core.cljs
@@ -13,6 +13,81 @@
 (defn- string->js-array [s]
   (.parse js/JSON (.-password s)))
 
+;; ******************************************************************************** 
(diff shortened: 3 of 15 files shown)


mandrigin:
It is wip, but I'd be glad to hear some suggestions on how to simplify our sign-in flow...

src/status_im/utils/keychain/core.cljs
```diff
@@ -13,6 +13,45 @@
...
+  (if android?
+    (callback "") ;; no-op on Androids (for now)
+    (-> (.getInternetCredentials rn/keychain address)
+        (.then (partial handle-callback callback)))))
+
+;; Clears the password for a specified address from the Keychain
+;; (example of usage is logout or signing in w/o "save-password")
+(defn clear-user-password [address]
+  (if (not android?)
+    (.resetInternetCredentials rn/keychain address)))
```
janherich: `when` should be used instead of `if` in this case (no 2 way branching)

src/status_im/ui/screens/events.cljs
```diff
@@ -215,17 +215,14 @@
...
-                                                        :name name))))
-   :handle-initial-push-notification-fx db})
+  (let [{:keys [address photo-path name]} (first (sort-by :last-sign-in > (vals accounts)))]
+    (cond-> {:handle-initial-push-notification-fx db}
+      ;; if there is nothing, let's just show the intro
+      (empty? address)       (assoc :db (assoc db :view-id :intro :navigation-stack (list :intro)))
+                                 ;; otherwise, we pre-fill the nav stack
+      (not (empty? address)) (-> (assoc :db (assoc db :navigation-stack (list)))
+                                 ;; and trying to open login
+                                 (assoc :dispatch [:open-login address photo-path name])))))
```
janherich: We should not use unnecessary dispatch (`:open-login`), but correctly call model function instead

src/status_im/ui/screens/events.cljs
```diff
@@ -288,6 +285,7 @@
  :logout
  (fn [{:keys [db] :as cofx} _]
    (let [{:transport/keys [chats]} db]
+     (keychain/clear-user-password (get-in db [:account/account :address]))
```
janherich: This call makes `initialize-db` function impure and hard-to test.
Instead of calling side effect directly, it should be hooked into re-frame effect and this function should just indicate the effect in the return value.

src/status_im/ui/screens/accounts/login/models.cljs
```diff
@@ -113,19 +126,31 @@
...
     (if success
-      {:db             db
-       :clear-web-data nil
-       :change-account [address]}
+      (merge
+       {:db             db
+        :clear-web-data nil
+        :change-account [address]}
+       (when save-password
+         {:dispatch [:save-password address password]}))
```
janherich: Unnecessary dispatch

janherich (changes requested):

src/status_im/ui/screens/accounts/login/models.cljs
mandrigin: so it can be just `{:save-password [address password]}` here?

mandrigin:
@janherich I'm also very unsure about `:open-login` and `:open-login-internal` dance. do you think there might be a better way to achieve the same?

src/status_im/ui/screens/events.cljs
mandrigin: ah, I just have seen this pattern in a few places (`src/status_im/utils/notifications.cljs`, `src/status_im/ui/screens/accounts/views.cljs`), so I wanted to keep them consistent.

janherich: That pattern is unfortunately very wrong (unnecessary chaining of events) and originally was everywhere in the app.
It's now much better and recently refactored code (like for example chat input/sending/receiving) doesn't suffer from it at all.
General very simple rule would be - only dispatch as consequence of external action, be it some user input (on-click/on-change callback in view), signal in asynchronous process (callback from http call), etc. 
Do not do it where no asynchronous behavior is needed and when simple function call is sufficient.
By doing it, we are unnecessary bringing asynchronous concerns and potential race condition into the game, because not only are we writing side-effect code (event dispatch is side-effect, impure action) where it could be pure functional, we are also writing asynchronous code for no good reason other then "writing it that way is convenient and we don't have to think about return values", which is very poor excuse, given that we choosed  functional language as our tool :).
Another bad thing is that events are global and so you can quickly create circular dependencies that way, with no sane hierarchy of namespaces and their dependencies (totally intermingled quagmire) which would otherwise be forbidden by the cljs compiler.

To provide some useful advice and not only mad ranting, here is my suggestion how it could implemented in functional way:
```clj
(let [default-effects {:handle-initial-push-notification-fx db}]
  (if (empty? address)
    (handlers-macro/merge-fx cofx 
                             default-effects
                             (navigation/navigate-to-clean :intro))
    (handlers-macro/merge-fx cofx
                             default-effects
                             (model/open-login last-account))))
```

src/status_im/ui/screens/accounts/login/models.cljs
janherich: Exactly, when the `:save-password` event does nothing else other then produces `::save-password` effect, there is no reason for it to be an event (unless some outside action in app requires just the `::save-password` effect to happen and nothing more, which is not the case).
It would be better to implement only `:save-password` effect (global and not namespaces keyword as `::save-password`) and no event, then this effect could be included in return value of any handler function where it's needed, with no unnecessary event chain.

janherich (approved):
Great work !

src/status_im/ui/screens/accounts/login/models.cljs
```diff
@@ -34,22 +37,26 @@
...
 
-(defn open-login [address photo-path name {db :db}]
+(defn navigate-to-login [address photo-path name password {db :db}]
   {:db       (update db
                      :accounts/login assoc
                      :address address
                      :photo-path photo-path
+                     :password password
                      :name name)
    :dispatch [:navigate-to :login]})
```
janherich: Minor point, just advice for the future.
We already have `navigate-to` exposed as pure function in navigation ns, so the above could be written like that without any additional dispatch (given that you already required navigation ns as `navigation`):
```clj
{:db (-> db
        (update :accounts/login merge {:address      address
                                       :photo-path photo-path
                                       :password   password
                                       :name         name})
        (navigation/navigate-to :login))}
```

mandrigin:
Okay, so it's not WIP anymore 🍾

errorists:
savepassword

I'd replace the checkbox with a switch, also ping @obi2020 is the copy ok ?

src/status_im/utils/keychain/core.cljs
```diff
@@ -77,3 +143,25 @@
...
+
+(re-frame/reg-fx
+ :get-user-password
+ (fn [[address callback]]
+   (get-user-password address callback)))
+
+(re-frame/reg-fx
+ :clear-user-password
+ (fn [[address]]
+   (clear-user-password address #(if (not %)
```
rasom: please use `when-not` here

src/status_im/utils/keychain/core.cljs
```diff
@@ -77,3 +143,25 @@
...
 (defn set-username []
   (when platform/desktop? (.setUsername rn/keychain username)))
+
+;;;; Effects
+
+(re-frame/reg-fx
+ :save-user-password
+ (fn [[address password]]
+   (save-user-password address password
+                       #(if (not %)
```
rasom: the same comment regarding `when-not`, also if you put `address ` param to the next line it will make following lines shorter, i hope we still want to have <= 80 chars in line

src/status_im/utils/keychain/core.cljs
```diff
@@ -13,6 +13,72 @@
...
+        (.then callback))))
+
+(defn handle-callback [callback result]
+  (if result
+    (callback (.-password result))
+    (callback "")))
+
+;; Gets the password for a specified address from the Keychain
+(defn get-user-password [address callback]
+  (if (not platform/ios?)
```
rasom: https://github.com/bbatsov/clojure-style-guide#if-not

src/status_im/ui/screens/events.cljs
```diff
@@ -214,18 +215,17 @@
...
-                                                (update :accounts/login assoc
-                                                        :address address
-                                                        :photo-path photo-path
-                                                        :name name))))
-   :handle-initial-push-notification-fx db})
+(defn initialize-views [cofx]
+  (let [{{:accounts/keys [accounts] :as db} :db} cofx
+        {:keys [address photo-path name]} (first (sort-by :last-sign-in > (vals accounts)))
+        default-fx {:handle-initial-push-notification-fx db}]
+    (if (empty? address)
```
rasom: probably it should be `nil?` pred? or it might be a blank string?

jpbowen:
@obi2020 RE: the copy. Some options to shorten: 
Remember me? [ ]
Stay logged in? [ ] 
Keep me logged in? [ ]

... but defer to you

rasom (approved):

goranjovic (approved):

mandrigin:
@errorists we don't have this component in RN at the moment for some reason 😢 
if you don't mind, we can address it as a follow-up issue, because the functionality is important, IMO

errorists:
@mandrigin I don't mind at all, just if it's not too late yet, please make sure the checkbox is wrapped in a touchable with some padding inside so users don't have to be super precise with hitting that small checkbox.

mandrigin:
@errorists it is still in review... so not late at all! :)

src/status_im/utils/keychain/core.cljs
mandrigin: will read :)

src/status_im/ui/screens/events.cljs
mandrigin: theoretically it should only be `nil` here, good point

mandrigin:
@errorists @yenda helped me to find the right UI component :)

simulator screen shot - iphone 8 plus - 2018-08-24 at 13 10 57

src/status_im/utils/keychain/core.cljs
```diff
@@ -20,11 +20,39 @@
...
+   ;; > the item can be restored to the same device that created a backup, 
+   ;; > but it isn’t migrated when restoring another device’s backup data.
+   ;; > ...
+   ;; > For extremely sensitive data 
+   ;; > THAT YOU NEVER WANT STORED IN iCloud, 
+   ;; > you might choose kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly.
+   ;; That is exactly what we use there.
+   ;; Note that the password won't be stored if the device isn't locked by a passcode.
+   :accessible    (enum-val "ACCESSIBLE"     "WHEN_PASSCODE_SET_THIS_DEVICE_ONLY")})
+
```
corpetty: great work and commenting.

corpetty (approved):
great job, looks good from my end.

lukaszfryc:
@mandrigin I noticed few issues:

1) Touch ID prompt appears on the white background. If the cancel button is pressed the prompt is dimissed and the app shows white screen instead of the login screen. I suggest showing the Touch ID prompt (or passcode screen) on top of the login screen. So that if user canceled it she can always login the standard way

2) When I remove Touch ID from my iOS settings and open the app, the iOS passcode screen appears which is what I expect. But, the passcode screen does not appear next time I kill the app and open it again. You can reproduce it by removing Touch ID from iOS settings and then enabling password save and killing the app. When the app is opened again, I'd expect the passcode screen to appear

3) Improve copy in Touch ID prompt. Now it says “Touch ID for Status”. If you can, I'd change it to “Use Touch ID to login to Status”. @jpbowen wdyt?

Other than that it looks really nice 👍 I also did not notice any regressions on Android.

mandrigin:
actually, TouchID shouldn't appear there, it was unintentional :-P
I'll think on how to make it more usable then...

lukaszfryc:
It's even better if we can have a seamless experience (being always logged in unless I logged out deliberately) without Touch ID :)

mandrigin:
yeah, that was my idea, but I think I made the keychain flags waaay too strict

obi2020:
@errorists @jpbowen thanks for the ping. I like `Stay signed in` since we use `Sign in` elsewhere in the UI.

mandrigin:
@lukaszfryc now it shouldn't ask for TouchID or device password to sign in. I will introduce this option later on, in the follow-up PRs.

lukaszfryc:
@mandrigin now when I kill the app and reopen it the user is logged out. The app opens on login screen. I'm testing on iPhone 8, 11.4.1

mandrigin:
@lukasz try with a new  account and make sure that you have a device password and Touch ID or Face ID activated on the device.
I just changed the access rights, so maybe it just can’t access the old entry. So we might try with the account with a different address.

lukaszfryc:
@mandrigin you're right. It works when I created a new account. Also, as you said, it does not work when nor password neither touch / face id are set. I suppose we can't make it work in this case, right? If yes we may consider hiding this feature or disabling the toggle with a hint when we know it won't work.

mandrigin:
Hmm, I think I can do that as a follow up if you don’t mind...

lukaszfryc:
@mandrigin I am not sure. If someone do not have passcode or other form verification configured, the feature will simply not work for him. I'd rather not see this toggle in such case.

mandrigin:
agree. I will take a look

mandrigin:
@lukaszfryc did a basic support of that, can you take a look? I have some problems compiling the app for iOS12, so I can't test on a real device now :(

chadyj:
Unfortunately this PR broke desktop login https://github.com/status-im/status-react/pull/5663

chadyj:
Android issue > https://github.com/status-im/status-react/issues/5793

PR #5652, opened 2018-08-27
LES with discovered nodes for all networks (main, ropsten, rinkeby)

### LINKED ISSUE #5643: Bring back options to select Mainnet and Ropsten with LES
### User Story

As a developer, I'd like to test LES synchronization directly on a smartphone.

Blocked by 
https://github.com/status-im/status-react/issues/5891

### Description

*Type*: Feature

*Summary*: In the previous week, we successfully run tests and merged a few PRs into status-go that allow to sync the chain using LES. The next step is to bring back options to select Mainnet and Ropsten with LES in the app settings.

#### Expected behavior
I can open the Status app, navigate to Profile and change the network to Mainnet or Ropsten with LES. This action should restart the app and after re-login, the chain should start syncing via LES.

#### Actual behavior
I can only select Mainnet or Ropsten with upstream server.

### Solution
*Summary*: 

- [ ] "Mainnet with LES" is available as a network option,
- [ ] "Ropsten with LES" is available as a network option.
- [ ] These options are available only in nightlies with proper information that they are WIP.

### Commnets

- status-go version: https://github.com/status-im/status-go/releases/tag/v0.13.0
- artifacts: `tags-v0.13.0^0-g3521c2ac-327` (I will work on changing it as it should be `v0.13.0`)

dshulyak:
fixes: #5643


.env
```diff
@@ -1,7 +1,6 @@
 TESTFAIRY_ENABLED=0
 ETHEREUM_DEV_CLUSTER=1
 OFFLINE_INBOX_ENABLED=1
-RPC_NETWORKS_ONLY=1
```
mandrigin, line 4: I would rather explicitly set it to `0` for readability

mandrigin (approved):
Looks good. Is there anything else that needs to be fixed?

modules/react-native-status/android/build.gradle
```diff
@@ -16,7 +16,7 @@ dependencies {
     implementation 'com.instabug.library:instabug:3+'
     implementation 'status-im:function:0.0.1'
 
-    String statusGoVersion = 'develop-g19b53030'
+    String statusGoVersion = 'develop-g5ef26596'
```
mandrigin, line 19: maybe we should use `0.13.0` release of status-go?
see: https://github.com/status-im/status-react/issues/5643

mandrigin:
I think it should close https://github.com/status-im/status-react/issues/5643

dshulyak:
Current version will work only with eth.staging fleet. Another difference is that this version has new CHT's for all networks. And enables sync from CHT on rinkeby.
We need to make one more release with rendezvous nodes for eth.beta.

mandrigin:
@dshulyak I think we should only use staging fleet for nightly if the network is non-RPC. I would add another flag to the config file like `FLEET_LES` and used it if `upstream` of the node config isn't set.

dshulyak:
> @dshulyak I think we should only use staging fleet for nightly if the network is non-RPC. I would add another flag to the config file like FLEET_LES and used it if upstream of the node config isn't set.

What are the reasons for doing that?
Staging environment provides unique whisper and rendezvous nodes. LES nodes are the same for either staging or beta fleet. They are discovered from broader network.
That's why this looks unnecessary. If staging is not useful for nightly then we should deploy rendezvous nodes for eth.beta.

dshulyak:
I will close this PR in favor of https://github.com/status-im/status-react/pull/5664 . We still need to decide what we do with eth.staging/nightly situation, but there is no reason to keep this PR open

Issue #4690, opened 2018-06-08
Chat badges


andmironov:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story

As a user, I want to see these badges:
- Errors (Offline)
- Add to Contacts
- Unread messages above the screen

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Feature

*Summary*: 

**Unread messages**
The `N Unread messages` badge is shown if user opens a chat, and there are a lot of unread messages and they do not fit in the screen. We show this badge with a counter which basically means “scroll to the beginning of unread messages”

Tap on Close icon (X) in `N Unread messages` badge :  
- mark all unread messages as read.
- unread messages badge disappears

Tap on `N Unread messages` : 
- scroll to the beginning of unread messages 
- marks all unread messages as read
- unread messages badge disappears

**Add to contacts**
We show it if the user has a chat with somebody who is not in their contacts. 

Tap on `Add to contacts` :
- Badge disappears
- User added to contacts

Tap on Close icon (X) in `Add to contacts`:
- Badge disappears

**Offline. Trying to reconnect**
Shown when there's a network error

#### Expected behavior
Only one badge is visible at a time. Priority:
1. Erros
2. Unread messages
3. Add to contacts

![badges](https://user-images.githubusercontent.com/3115606/41168361-64cebaf0-6b4e-11e8-8eec-f25f5d7c728b.png)

Zeplin:
zpl.io/aN0ZKWQ
zpl://screen?sid=5b1a79ae98f9b848590fb930&pid=5a55d587ba304eed4d2c942f

EugeOrtiz:
Just a couple of questions on the "Offline. Trying to reconnect": is it effectively like that? The app tries to reconnect? How many attempts? What happens after those attempts?

chadyj:
Just curious about this unread message implementation as it seems to be the inverse of the originally proposed [issue](https://github.com/status-im/status-react/issues/4607).  In that example, the scroll position is maintained through sessions so that when a user returns to a chat they are in the same position/history as they were previously. The user can either scroll down and catch up, or simply tap the link and jump to the latest message. 

However in this design, it looks like the behavior is to always take the user to the latest message, thereby losing their scroll position. A user wanting to catch up will always have to scroll up and find their position or tap the button. Is this right?

We have had a lot of feedback that users do NOT want their chat/scroll position changed so they can easily follow along with threads without losing their place. 

[...]

PR #5665, opened 2018-08-28
[WIP] Maintain scroll position in chat

### LINKED ISSUE #5185: Scroll position jumps around as new messages arrive in chat
### Description

*Type*: Bug

*Summary*: As I scroll up to catch up on the conversation, the scroll position jumps around as new messages arrive. This is pretty annoying, especially on busy public chats.

#### Expected behavior
Scroll position does not jump when new message arrives in a chat.

#### Actual behavior
Scroll position jumps as new message arrives in a chat.

### Reproduction

- Join public chat with lot of messages
- Scroll the public chat
- Receive new messages in the chat and observe the scroll position

### Additional Information
* Status version: 0.9.22

dmitryn:
fixes #5185
addresses #4690

### Summary:

* Persist scroll position in chat using AsyncStorage
* Add "scroll to bottom" button

Please notice there is a delay happens before chat gets rendered. I set opacity to 0 then scroll to last offset and then set opacity back to 1. That's the only solution i've found to overcome lack of proper initial scroll offset support in RN https://github.com/facebook/react-native/issues/6849

screenshot 2018-08-28 12 28 41
screenshot 2018-08-28 12 14 29

status: ready


janherich:
@dmitryn out of curiosity why another dependency (async-storage) instead of realm ?

dmitryn:
@janherich i had troubles with getting Realm to save data on app close / stop.

dmitryn:
kapture 2018-08-29 at 12 54 10

There is ~1 second delay before chat gets rendered. Not ideal perfomance, but might be acceptable regarding lack of proper RN API for this.

pedropombeiro:
@dmitryn I just tried the build and seems to be working well. One thing I noticed is that when I started the app I started receiving lots of messages on #status channel, so I was expecting that when I went to that channel, it would position the scroll at the first unread message. Instead, it opened the view at the very bottom (most recent messages). Is that expected?

pedropombeiro:
I went to #blah channel and was positioned at the bottom. From another client, I sent a couple of messages. I could see the unread button getting updated each time, but if I ignored it and started scrolling with my finger, it wouldn't show the new messages (they weren't in view).

UPDATE: the view ends up getting updated if I scroll a bit more up in history and then back.

src/status_im/models/chat.cljs
```diff
@@ -94,3 +96,18 @@
...
+  ([state {{:keys [chat-ui-props current-public-key]} :db}]
+   (when (#{"background" "inactive"} state)
+     (->> chat-ui-props
+          (reduce-kv
+           (fn [acc k v] (assoc acc k (dissoc v :input-ref)))
+           {})
+          clj->js
+          js/JSON.stringify
+          (.setItem js-dependencies/async-storage
+                    (str "@StatusIm:" current-public-key ":chat-ui-props")))
```
cammellos, line 112: could you use the `sha3` of the public key here? 
In that way we won't disclose the user key in case of compromise.

Also we are fine storing `chat-ui-props` unencrypted? no confidential info there?

(we had a discussion already about async-storage, so I take you made the decision to use it instead of realm for good reasons)

resources/icons/dropdown_down.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
+  <path transform="rotate(180 12 12)" fill="" d="M12.1872807,7.00692544 C12.4141917,7.02726144 12.6351377,7.12378244 12.8065087,7.29515444 L19.8740787,14.3627244 C20.2655697,14.7542144 20.2690747,15.3854394 19.8758277,15.7786864 C19.4853037,16.1692114 18.8464497,16.1635214 18.4598657,15.7769374 L12.0831877,9.40026044 L5.70651067,15.7769374 C5.31502067,16.1684274 4.68379467,16.1719334 4.29054767,15.7786864 C3.90002367,15.3881624 3.90571267,14.7493084 4.29229667,14.3627244 L11.3598667,7.29515444 C11.5862877,7.06873444 11.8928977,6.97209244 12.1872807,7.00692544 L12.1872807,7.00692544 Z"/>
```
flexsurfer, line 2: you can use dropdown_up and use transform for container https://github.com/status-im/status-react/blob/d834b6e37d1ecc52a2367676cee0b338c383087f/src/status_im/ui/components/tooltip/views.cljs#L29

flexsurfer:
` i had troubles with getting Realm to save data on app close / stop.` why do you need this data on app close?

chadyj:
Hey great to see this!  This is a really nice touch.

I ran into 2 main issues:

- Scrolling performance with lots of messages was glitchy. When I switched back to nightly I didn't have the problem.

- The [1 second delay](https://github.com/status-im/status-react/pull/5665#issuecomment-416896185) didn't feel good at all. For me, it made the app feel very slow and unresponsive, especially when jumping between screens/profiles/chats.l

I also had 1 small wish... that tapping the "scroll to bottom" icon would have a quick scroll animation similar to WhatsApp. It felt a bit jittery how it jumped.

Using iOS12 iPhone X.

dmitryn:
>why do you need this data on app close?

@flexsurfer to persist offsets. It's being stored in app-db until you logout or close the app.

src/status_im/models/chat.cljs
dmitryn: @cammellos good idea about `sha3`.

`chat-ui-props` doesn't contain any sensitive info, just some flags and data related to UI.
As long as nothing sensitive is stored in AsyncStorage, i don't see issues with it.

dmitryn:
@chadyj thanks for testing it. I also don't like that delay, but i haven't found better solution yet.

dmitryn:
>One thing I noticed is that when I started the app I started receiving lots of messages on #status channel, so I was expecting that when I went to that channel, it would position the scroll at the first unread message. Instead, it opened the view at the very bottom (most recent messages). Is that expected?

@PombeirP on first run it should start at the bottom as we don't know offset for that chat yet.

pedropombeiro:
> @PombeirP on first run it should start at the bottom as we don't know offset for that chat yet.

@dmitryn I guess this is only an noticeable because this is the first time I run the new version of the app, which hadn't yet stored the offsets. If this had been my first use of the app I wouldn't have noticed that shortcoming because the app would have stored the offsets as I started opening chats.

I can see only one particular scenario where this would be noticeable even if this was my first use, which is if someone starts a 1:1 chat with me and sends a huge amount of messages. When I go to the chat the first time, I'll start at the bottom, instead of the first unread. Maybe that's OK, but I was under the impression that we'd start from the first unread message.

Also, do you think it would be easy to add a red divider line marking the first unread message (similar to Slack)? I feel that having that visual anchor makes it easier to navigate history.

image

dmitryn:
Tried couple of new solutions to scrolling issue, none of them look good on Android. I'm out of ideas for now.

rasom:
@dmitryn do we have a fix for ios? can we merge it and work on android separately?

chadyj:
Here is another proposal that works around maintaining the scroll position.
Instead it always defaults scroll position to the last unread message. Then the user can read all, or jump to the latest and clear unread.

https://github.com/status-im/status-react/issues/6375

dmitryn:
@rasom on ios we have built-in `contentOffset` property that we can use https://facebook.github.io/react-native/docs/scrollview#contentoffset
On Android there is no such thing.

Current implementation use delay to make it working which is not good UX. 

<blockquote><div><strong><a href="https://facebook.github.io/react-native/index.html">ScrollView · React Native</a></strong></div><div>Component that wraps platform ScrollView while providing integration with touch locking "responder" system.</div></blockquote>

dmitryn:
Closing this one as another PR addressing this is ready https://github.com/status-im/status-react/issues/6375

PR #5667, opened 2018-08-28
Show a progress indicator while logging in account using saved password.

mandrigin:
On slower devices the white screen is shown for a few seconds after the
status logo. It looks like that something is wrong with the app.

This commit adds a simple progress screen (full-screen with an animated
activity indicator) that is shown before the main screen is loaded.

### Steps to test:
- Open Status on iOS (better use a slower device)
- Login and save your password
- Kill the app and restart it
- The progress indicator is shown after the status logo and until the main UI appears.

status: ready

### DIFF
--- a/src/status_im/ui/screens/progress/styles.cljs
+++ b/src/status_im/ui/screens/progress/styles.cljs
@@ -1,0 +1,10 @@
+(ns status-im.ui.screens.progress.styles
+  (:require [status-im.ui.components.colors :as colors]
+            [status-im.ui.components.toolbar.styles :as toolbar.styles]))
+
+(def container
+  {:flex             1
+   :align-items      :center
+   :justify-content  :center
+   :background-color colors/white})
+
--- a/src/status_im/ui/screens/accounts/login/models.cljs
+++ b/src/status_im/ui/screens/accounts/login/models.cljs
@@ -77,5 +77,6 @@
   (if password
     (handlers-macro/merge-fx cofx
                              {:db (assoc-in db [:accounts/login :password] password)}
+                             (navigation/navigate-to-cofx :progress nil)
                              (user-login))
     (navigation/navigate-to-cofx :login nil cofx)))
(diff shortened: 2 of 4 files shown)


src/status_im/ui/screens/progress/styles.cljs
```diff
@@ -0,0 +1,10 @@
+(ns status-im.ui.screens.progress.styles
+  (:require [status-im.ui.components.colors :as colors]
+            [status-im.ui.components.toolbar.styles :as toolbar.styles]))
+
+(def container
+  {:flex             1
+   :align-items     :center
+   :justify-content :center
+   :background-color :white})
```
goranjovic: `colors/white`, also alignment

goranjovic (approved):

src/status_im/ui/screens/progress/styles.cljs
```diff
@@ -0,0 +1,10 @@
+(ns status-im.ui.screens.progress.styles
+  (:require [status-im.ui.components.colors :as colors]
+            [status-im.ui.components.toolbar.styles :as toolbar.styles]))
+
+(def container
+  {:flex             1
+   :align-items     :center
+   :justify-content :center
```
flexsurfer: indentation

flexsurfer (approved):

yenda (approved):

flexsurfer:
@mandrigin what if you'll press hardware back button on android ? will it return you to progress screen?

src/status_im/ui/screens/progress/styles.cljs
mandrigin: fixed

mandrigin:
@flexsurfer it shouldn't `(navigation/navigate-to-clean :home)`, but this screen is only shown when there is a saved password and there is no Android support for it now.

chadyj:
Related? https://github.com/status-im/status-react/issues/5681

flexsurfer (approved):

mandrigin:
will fix #5681

mandrigin:
@chadyj I think it is supposed to be shown at every launch for nightlies. I checked, the app UI is rendered behind it just fine.
I think QA are using this dialog to check if the build has TestFairy too.
And our users should be aware of that as well.

chadyj:
> @chadyj I think it is supposed to be shown at every launch for nightlies.

Thanks for looking into it.

Odd, I don't recall seeing it before 🤷‍♂️

mandrigin:
@chadyj I think it was just less annoying because you had to enter your password anyway...

PR #5580, opened 2018-08-21
Feature/qtbreakpad integration

maxhora:
### Summary:

Integrates google's breakpad library to intercept unhandled exceptions and launch reportApp to ask user report crash dump related files. reportApp opens folder with files prepared for uploading and at the same time navigates to dedicated Google Web Form in default web browser. On Google Web Form user submit new GitHub issue is created.

Related PR on which current depends - https://github.com/status-im/react-native-desktop/pull/318

### Review notes (optional):
Google Breakpad lib sources and wrapper to use lib in our main.cpp is moved to https://github.com/status-im/google-breakpad

### Testing notes (optional):
App forced to crash just after the start.

### Steps to test:
- Open Status
- Status should crash and report app window should be opened
- Press report button
- Google Web form should be opened in browser and directory with files to upload should be opened on desktop
- Fill Google Web form and upload the files, submit.
- New issue with provided information should appear at https://github.com/MaxRis/StatusImDesktopReports/issues

status: ready

Issue #5423, opened 2018-08-08
Refine macos app bundle installation experience


maxhora:
**Issue**

It was reported that just replacement of app bundle in Applications directory is not working as expected (old app should be removed to trash manually in advance).
Installation flow is not obvious for users as well.

**Type:** Bug.

**Solution**: Background image for mounted .dmg or at least symlink to Applications folder will enhance app installation user experience. Replacement of available app bundle in Application with new one should work.

**Tasks**:

- [x] add  symlink to Applications folder for mounted .dmg 
- [x] make possible to replace app in Application folder for  reinstalling

chadyj:
Not sure if its related, but I also see this problem `We detected a problem with the encryption key` when upgrading.

I hit apply and then get a blank screen.

I see this after deleting the old version of status, and then dragging a new version to Applications.

<img width="1021" alt="screen shot 2018-08-09 at 7 12 04 am" src="https://user-images.githubusercontent.com/116099/43879284-973465fa-9ba3-11e8-99d1-8a9ed3151125.png">

chadyj:
My log file.
```
Debug: starting ubuntu server...

Debug: wait for started...

Debug: ubuntu-server std:  "outerRealmConstructor: undefined"

Debug: ubuntu-server std:  "Loaded realmConstructor: function Realm() { [native code] }"

Debug: ubuntu-server err:  "-- Server starting"

Debug: waiting finished

Warning: WebEngineContext used before QtWebEngine::initialize() or OpenGL context creation failed.

Debug: ubuntu-server err:  "-- Connection from RN client"

Debug: ubuntu-server std:  "outerRealmConstructor: function"

Debug: ubuntu-server std:  "Running application \"StatusIm\" with appParams: {\"initialProps\":{},\"rootTag\":1}. __DEV__ === false, development-level warning are OFF, performance optimizations are ON"

Debug: ubuntu-server std:  "DEBUG [status-im.utils.handlers:37] - Handling re-frame event:  :initialize-keychain"

Debug: ubuntu-server std:  "DEBUG [status-im.utils.keychain.core:57] - initializing realm encryption key..."

Debug: invoked RNKeychainManager::setUsername with username =  "status-im.encryptionkey"

Debug: invoked RNKeychainManager::getGenericPasswordForOptions

Debug: RNKeychainManager::getGenericPasswordForOptions failed:  Password not found

[...]

Issue #5421, opened 2018-08-08
Sign Jenkins build artifacts to avoid security warnings


maxhora:
**Issue**
* macos security warnings are shown on start of application. 
* Keychain asks to re-enter password on every binary update.

**Type:** Bug.

**Solution**: Above might be solved with signing of .dmg and .app according to https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG17

Makes sense to sign AppImage for Linux as well. Image creation tool supports signing via cli options.

churik:
Set `high-severity` here, because it can affect possibility to install and use Mac OSx builds for automation tests.

chadyj:
<img width="420" alt="screen shot 2018-08-15 at 12 24 22 pm" src="https://user-images.githubusercontent.com/116099/44144036-4e3b4aca-a086-11e8-8f03-624f407d5cd4.png">

> "StatusIM" can't be opened because it is from an unidentified developer.
> 
> Your security preferences allow installation of only apps from the App Store and identified developers.

chadyj:
<img width="434" alt="screen shot 2018-08-15 at 12 27 01 pm" src="https://user-images.githubusercontent.com/116099/44144122-91221a3a-a086-11e8-933f-037ac4173eb4.png">

> StatusIM wants to use your confidential information stored in "org.status.status-im.keychain" in your keychain.
> 
> The authenticity of "StatusIM" cannot be verified. To allow this, enter the "login keychain password.

PR #5668, opened 2018-08-28
Make macOS bundle code-signable

gnl:
This is step 1 towards #5421 and #5423 . It updates the CI configuration to support the new codesign-compatible macOS bundle structure: 
https://github.com/status-im/StatusAppFiles/pull/1

The following changes were made to the bundle in order to support code signing:

- remove .env from the bundle root (violates strict macOS bundle structure equirements and we don't need it)
- rename `MacOs` folder to `MacOS` - even though macOS is case-insensitive, code
signing is not
- move helper binaries into Resources - `ubuntu-server` cannot be signed
and all binaries in the `MacOS` folder have to be signed
- remove absolute RPATH from `reportApp` (see #5580) during building/deployment – macOS Gatekeeper doesn't allow those. Fragile temporary solution - look into fixing compiler flags in the future, so the RPATH doesn't end up in the binary in the first place.
See https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling

In addition to these, there are a number of additional violations of Apple's requirements for the bundle structure but they seem common for QT-based apps and don't currently break code-signing, although they might in the future:

- There shouldn't be any `.dylib`s in the `PlugIns` directory, those belong under `Frameworks`
- The framework versions should be consecutive letters starting with `A`, rather than the numbers commonly used in QT apps – `5`

See: https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG300

Status: This currently points to gnl/StatusAppFiles. As soon as it's merged this can be merged as well without breaking anything:
https://github.com/status-im/StatusAppFiles/pull/1
A separate PR can then be created to point to a specific revision in status-im/StatusAppFiles rather than master, so pointing to a separate repo will no longer be necessary in the future.

EDIT: Done – https://github.com/status-im/status-react/pull/5702/files#diff-8cff5e6aadd8182c6dd56f9b32b9045eR182


ci/desktop.groovy
```diff
@@ -172,11 +172,12 @@ def compileMacOS() {
 def bundleMacOS(type = 'nightly') {
   def pkg = common.pkgFilename(type, 'dmg')
   dir(packageFolder) {
-    sh 'git clone https://github.com/vkjr/StatusAppFiles.git'
-    sh 'unzip StatusAppFiles/StatusIm.app.zip'
-    sh 'cp -r assets/share/assets StatusIm.app/Contents/MacOs'
-    sh 'chmod +x StatusIm.app/Contents/MacOs/ubuntu-server'
-    sh 'cp ../desktop/bin/StatusIm StatusIm.app/Contents/MacOs'
+    sh 'curl -L -O "https://github.com/gnl/StatusAppFiles/raw/master/StatusIm.app.zip"'
```
jakubgs: I really dislike this step. We have to fix this at some point.
Also, why are we not using the repo anymore?

gnl: Not a fan of it either. I moved away from cloning the repo because if there are multiple commits, as soon there will be, cloning it will take up a lot more space than necessary. This way we only get the one file.

jakubgs (approved):

maxhora (approved):
Thank you for the investigation of what is applicable for Apple store submitting. There really might be an issues with new added Plugins (have desktop notifications support in mind, especially).

vitvly (approved):

vkjr (approved):

churik:
@gnl can you please leave some testing notes (what exactly should be changed for end user)?
I didn't notice any changes in last [MacOS](https://status-im.ams3.digitaloceanspaces.com/StatusIm-180830-230023-8784e5-pr.dmg) build, 
1) no symlink is added to Applications folder for mounted .dmg
2) security warning is still shown
3) keychain access is required.
Replacing is working (tested on nightly 31/08/2018)

churik:
Ok, after a small talk with @gnl turned out, that these changes shouldn't affect current builds and the only thing that should be checked that everything is working as before.
Tested on [MacOS](https://status-im.ams3.digitaloceanspaces.com/StatusIm-180830-230023-8784e5-pr.dmg) build and didn't notice any regression, so confirm that PR is ready to merge.

maxhora:
Adding my comment here as discussed: "Lets address RPATH issue with reportApp in separate PR (I will look into it to set it from CMake script).
I have added special install_name_tool step directly to Jenkinsfile for reportApp previously, since RPATH was not adjusted for reportApp automatically by deployment tools, but for StatusIm binary it was updated without any such special additions in CMake directly.
But it makes sense to address this in CMake script as you described, thanks"

PR #5675, opened 2018-08-28
Feature/remove old flags

yenda:
Removed: `add-custom-mailservers-enabled?`, `universal-links-enabled?`, `offline-inbox-enabled?`
Enabled on all builds and could be removed: `SPAM_BUTTON_DETECTION_ENABLED`
Doesn't exist as a flag but is still in build configs: `USE_SYM_KEY`
clean up env files from dead flags


dmitryn (approved):

rasom (approved):

annadanchenko (approved):

mandrigin (changes requested):
Should we remove these flags from .env files as well? Otherwise, it’ll be confusing.

yenda:
@mandrigin I removed them as well it was in the todo just wanted to make sure everybody agrees with the removed flags.

mandrigin (approved):
:+1:

PR #5679, opened 2018-08-28
Copy Updates Aug 2018

obi2020:
### Summary:

Several copy updates, include refinements to Home screen prompts and signing phrase explanation.

### Steps to test:
- Open new or unused Status account
- Review Home screen
- Step thru wallet setup
- Start a new chat

status: ready


janherich (approved):

src/status_im/translations/en.cljs
```diff
@@ -438,7 +438,7 @@
    :cost-fee                             "Cost/Fee"
    :nonce                                "Nonce"
    :confirmations                        "Confirmations"
-   :confirmations-helper-text            "Consider transactions complete after 12 confirmations"
+   :confirmations-helper-text            "Transactions with 12 confirmations can be considered irreversible"
```
chadyj: Is it right that transaction with less than 12 confirmation are reversible? The wording seems to imply that.

src/status_im/translations/en.cljs
```diff
@@ -254,10 +254,10 @@
    :edit-chats                           "Edit chats"
    :search-chats                         "Search chats"
    :empty-topic                          "Empty topic"
-   :topic-format                         "Wrong format [a-z0-9\\-]+"
-   :topic-name-error                     "A topic name can contain only following characters: a-z -"
+   :topic-format                         "Topic names contain letters and numbers only"
+   :topic-name-error                     "Topic names contain letters and numbers only"
```
chadyj: How does it look in the app when the message is repeated?

src/status_im/translations/en.cljs
```diff
@@ -472,11 +472,11 @@
...
    :wallet-manage-assets                 "Manage Assets"
    :sign-message                         "Sign Message"
-   :signing-message-phrase-description   "Sign the message by entering your password. Make sure that the words above match your secret signing phrase"
-   :signing-phrase-description           "Sign the transaction by entering your password. Make sure that the words above match your secret signing phrase"
+   :signing-message-phrase-description   "If you recognize these words, enter your login password to sign the message"
+   :signing-phrase-description           "If you recognize these words, enter your login password to sign the transaction"
    :wallet-insufficient-funds            "Insufficient funds"
    :wallet-insufficient-gas              "Not enough ETH for gas"
-   :error-unable-to-get-prices           "Currency conversion error. Try to refresh to update later"
+   :error-unable-to-get-prices           "Currency conversion error. Refresh your screen to try again."
```
chadyj, line 483: How to refresh?

src/status_im/translations/en.cljs
```diff
@@ -602,7 +602,7 @@
    :transactions-filter-type             "Type"
    :transactions-filter-select-all       "Select all"
    :view-transaction-details             "View transaction details"
-   :transaction-description              "Please wait for at least 12 confirmations to make sure your transaction is processed securely"
+   :transaction-description              "Transactions with 12 confirmations can be considered irreversible"
```
chadyj, line 609: Seems odd to repeat the same wording in different contexts?

src/status_im/translations/en.cljs
```diff
@@ -696,7 +696,7 @@
...
+   :mainnet-is-default-alert-text        "Thanks for choosing Status! We are now in Beta, which means new installs are automatically connected to the Mainnet. If you upgraded, you may still need to switch networks, which you can do easily by navigating to your Profile -> Advanced Mode -> Switch Developer Mode on -> Choose your network. Now just have fun! Browse DApps, exchange assets, start your own #chats, and message your friends, all from the world's first decentralized, mobile client for Ethereum.\n\nPlease note that this is a beta release and we advise you that using this app should be done for testing purposes only and you assume the full responsibility for all risks concerning your data and funds."
```
chadyj, line 705: I can't see what changed here?

> We are now in Beta, which means new installs are automatically connected to the Mainnet. If you upgraded, you may still need to switch networks, which you can do easily by navigating to your Profile -> Advanced Mode -> Switch Developer Mode on -> Choose your network.

This wording was included as we changed behaviour going from alpha to beta, but we probably can omit this now, or at least update it?

chadyj (commented):
Looking good Obi!

Made a few comments but otherwise is good to go 👍

src/status_im/translations/en.cljs
obi2020: Maybe I don't understand where these are used. When would they appear together?

src/status_im/translations/en.cljs
obi2020: I think that depends on the OS? e.g. in Android you swipe down to refresh a screen

src/status_im/translations/en.cljs
obi2020: Please help me understand the use case here. In previous, my understanding was that it appeared when users have just sent a tx. When does this appear?

src/status_im/translations/en.cljs
obi2020: Hmm. I believe I wanted to shorten it drastically then link to the wiki. I may have erased my changes due to time constraints (wiki entry not ready). But you're right it should be removed or edited way down because people won't read it.

src/status_im/translations/en.cljs
obi2020: You're right and that's my understanding per this thread. https://ethereum.stackexchange.com/questions/183/how-should-i-handle-blockchain-forks-in-my-dapp/203#203 Though maybe it's not a good idea to let users think there's a way to stop the tx after they've sent it. 

The 12 confirmations comes across as rather specific and we've seen that users think it has something to do with their choices. I'm trying to express that it's related to how Ethereum transactions work (without being too wordy). 

let me try again...

src/status_im/translations/en.cljs
chadyj: Do you have a screenshot of these messages in-app? Would be helpful to see both in-context.

Otherwise we have the same message `topic-format` and `topic-name-error` repeated in different places.

src/status_im/translations/en.cljs
chadyj: Both platforms have pull to refresh AFAIK. Since this is a hidden UI element perhaps we should let users know explicitly how to refresh?

src/status_im/translations/en.cljs
chadyj: The same copy "Transactions with 12 confirmations can be considered irreversible" is repeated in the two labels `confirmations-helper-text` and `transaction-description`.  

We should either clean up the strings to just have 1, or disambiguate the text depending on context. Either way screenshots would help bring clarity and context.

src/status_im/translations/en.cljs
obi2020: I've only found `topic-name-error` in the app. I don't know when the other is triggered. 
topic name error

src/status_im/translations/en.cljs
obi2020: I see what you mean, though we have to assume a user knows how to their device. Otherwise UI language will get stuffed with extra instruction and info

src/status_im/translations/en.cljs
obi2020: My reasoning for making them the same, is that any time that idea is expressed, it should be done in a consistent manner. 

You're right, however, that context might call for makign these difference. As is, I don't see a reason. e.g. if the user can take some action or another to affect the transaction

src/status_im/translations/en.cljs
obi2020: Thinking of pulling in a reference to how Ethereum works. WDYT?

`On the Ethereum network, when a transaction has 12 confirmations you can consider it settled.`

src/status_im/translations/en.cljs
chadyj: Hmm I'm not so sure. How would a user discover it on this particular screen? If you are telling the user to refresh shouldn't you let them know how?

chadyj (approved):

src/status_im/translations/en.cljs
annadanchenko: If we say "Topic names contain letters and numbers only" then upper case letters looks like ok. Same for non-english letters they are not allowed. We also can't create a topic with numbers (even it's said 0-9 are allowed). So, only lowercase english letters, and "-" are valid here. Can we change the copy, please? E.g. "Topic names can contain only lowercase English letters and - "

annadanchenko (changes requested):
please address https://github.com/status-im/status-react/pull/5679#discussion_r213963216

src/status_im/translations/en.cljs
```diff
@@ -472,11 +472,11 @@
    :wallet-settings                      "Wallet settings"
    :wallet-manage-assets                 "Manage Assets"
    :sign-message                         "Sign Message"
-   :signing-message-phrase-description   "Sign the message by entering your password. Make sure that the words above match your secret signing phrase"
-   :signing-phrase-description           "Sign the transaction by entering your password. Make sure that the words above match your secret signing phrase"
+   :signing-message-phrase-description   "If you recognize these words, enter your login password to sign the message"
+   :signing-phrase-description           "If you recognize these words, enter your login password to sign the transaction"
```
annadanchenko, line 480: "login password" sounds strange to me. We have no other password in the app, so I suggest to remove word "login"

obi2020: In testing we've observed that users get confused about what to enter here. They think they need to enter the three words or sometimes even the seed phrase. I'm attempting to make it more obvious this it's the password they use to open the app that's needed.

src/status_im/translations/en.cljs
obi2020: I'm reluctant to say "English" - how about `Topic names contain only lowercase a through z and dash (-)`

obi2020: or `Topic names use only lowercase letters (a to z) & dashes (-)`

annadanchenko: ok with Topic names use only lowercase letters (a to z) & dashes (-)

annadanchenko (approved):

src/status_im/translations/en.cljs
obi2020: This might also be legacy from when there were 2-3 different items in the app with some variation of "pass" in their terminology. Though we've updated those, I believe it's still important to give extra support here so the user know what to enter cc @PhilipandWu

yenda:
@annadanchenko rebased and squashed does it require testing or can it be merged as is ?

churik:
Tested on IPhone 7 and Samsung Galaxy S7.

jakubgs:
Merged by @yenda at https://github.com/status-im/status-react/commit/990aa20d8b591872515d4657c762d216a68b6c7e

Issue #5406, opened 2018-08-06
Unify Desktop build process


maxhora:
Inspired by @jakubgs 

To provide identical application build experience makes sense to create single Makefile to build app for Linux and/or MacOS platforms.

AppImage or .dmg build should happen after single command run, like `make desktop-linux`

It should work with Jenkins as well

PR #5702, opened 2018-08-29
Prettify macOS installation .dmg bundle

### LINKED ISSUE #5423: Refine macos app bundle installation experience

gnl:
Closes #5423. 

## Summary

* Add https://github.com/LinusU/node-appdmg as an optional, macOS specific NPM dependency to create the nicer bundle without affecting the build on non-macOS platforms.

* Rename `StatusIm` to `Status`

* Get bundle skeleton from updated `status-im/StatusAppFiles` repo and point to a specific commit insted of `master` to minimise the risk of future breakage.

## ToDo

Will need a lossless TIFF/PNG Retina and non-Retina background image for this...:

screen shot 2018-08-29 at 22 58 16

...so that it looks somewhat like this...:

https://github.com/LinusU/node-appdmg#json-input

But more status-y. :)

## To Test

Installation, launch, login.


EugeOrtiz:
Designs and specs can be found here >> zpl.io/VknWQKL (Zeplin, permission required)

group 2

maxhora (approved):

vkjr (approved):

jakubgs (approved):
I'm going to approve this but the desktop team really needs to work on automating their build process using something that can by run by a user on their dev machine. If it's just in a `groovy` script/`Jenkisfile` it cannot be.

Also, since when is a shell script file extension a `.bash` instead of `.sh`?

gnl:
@jakubgs 
> I'm going to approve this but the desktop team really needs to work on automating their build process using something that can by run by a user on their dev machine. If it's just in a groovy script/Jenkisfile it cannot be.

Agreed – #5406.

> Also, since when is a shell script file extension a .bash instead of .sh?

Since never, that's just my own thing in order to distinguish between `/bin/sh` and `/bin/bash`. I don't use Bash and rarely write scripts for it so it doesn't quite jump out at me as hard, but I do know it's weird and noone does it, so – fixed.

churik:
@gnl is it expected to have `.background` folder inside mounted disk?
upgr

churik:
As it turned out, I have option "To show hidden files" enabled by default.
Tested: 
- installation
- launching
- log in.
Doesn't affect mobile builds.

PR #5708, opened 2018-08-30
[#4772] Resolve ENS domains in browser

### LINKED ISSUE #4772: [Draft] Resolve ENS domains in browser
### User Story

As a user, I want to browse the fully decentralized web and load content from p2p storage platforms like IPFS and Swarm in the Status browser. 

### Description

*Type*: Feature

*Summary*: Resolve ENS domains to IPFS/Swarm hashes and redirect to content.

#### Expected behavior
Entering a .eth address into the Status URL bar will redirect to the associated content.

#### Actual behavior
The browser can not resolve .eth addresses.

flexsurfer:
fixes #4772


deps.edn
```diff
@@ -10,7 +10,8 @@
          com.andrewmcveigh/cljs-time {:mvn/version "0.5.2"}
          com.taoensso/timbre         {:mvn/version "4.10.0"}
          hickory                     {:mvn/version "0.7.1"}
-         com.cognitect/transit-cljs  {:mvn/version "0.8.248"}}
+         com.cognitect/transit-cljs  {:mvn/version "0.8.248"}
+         mvxcvi/multihash            {:mvn/version "2.0.3"}}
```
goranjovic: why do we need a new dependency?

flexsurfer: to work with multihashes

src/status_im/utils/ethereum/resolver.cljs
```diff
@@ -0,0 +1,16 @@
...
+(def default-address "0x0000000000000000000000000000000000000000")
+(def default-hash "0x0000000000000000000000000000000000000000000000000000000000000000")
+
```
jeluard: Why this? Can't we do the same way as it's done for others field in the `ens` namespace?

src/status_im/ui/screens/browser/events.cljs
```diff
@@ -39,19 +40,42 @@
...
+ (fn [{:keys [web3 registry ens-name cb]}]
+   (resolver/content web3 registry ens-name cb)))
+
 (handlers/register-handler-fx
  :browse-link-from-message
  (fn [_ [_ link]]
    {:browse link}))
 
+(handlers/register-handler-fx
+ :ens-multihash-resolved
```
jeluard, line 54: So we are supporting multihash?

flexsurfer: yes but it's stored in content field

src/status_im/utils/ethereum/resolver.cljs
flexsurfer: it's temporary solutions while https://github.com/status-im/status-react/pull/5459 not merged

flexsurfer: oh right if we're going with `content` field we can use same way

goranjovic (approved):

deps.edn
goranjovic: So I take it there is no way to inline that functionality? Every dep increases our attack surface

jeluard (approved):

src/status_im/ui/screens/browser/events.cljs
```diff
@@ -39,19 +40,42 @@
...
+ [re-frame/trim-v]
```
yenda: can you clean up the `trim-v` co-effect, we are removing them to have a more consistent codebase

src/status_im/ui/screens/browser/events.cljs
```diff
@@ -39,19 +40,42 @@
...
+ (fn [{:keys [db] :as cofx} [hash]]
```
yenda: ideally events should be one-liners that call a function from the models

deps.edn
flexsurfer: anyway it can't be compiled
```
Compiling ["index.desktop.js"] from ["components/src" "react-native/src/cljsjs" "react-native/src/desktop" "src" "env/prod"]...
[31mWARNING[0m: Extending an existing JavaScript type - use a different symbol name instead of js/String e.g string at line 246 target/desktop-prod/multihash/core.cljc
Error encountered performing task 'cljsbuild' with profile(s): 'prod'
```

asemiankevich:
I have tried to resolve some of those addresses:

http://phyrextsai.eth
http://christophershen.eth
http://hozt.portalnetworkweb.eth
http://teamnull.bluzellehackathon.eth
http://cosmos.bluzellehackathon.eth
http://dturker.bluzellehackathon.eth

for now, current implementation looks like this (infura link is shown in the browser)

image

image


https://ipfs.infura.io/ipfs/QmUpAghbXaytas97onhD48h1FN73JWTL5PBoBoaQkXxL9e

PR #5712, opened 2018-08-30
bug #5542 - fixed errors when no currency selected - using usd as def…

### LINKED ISSUE #5542: `Error: Does not support name` sending a transaction while no currency is set for account
***Type***: Bug

***Summary***: It was noticed that currency set in the account could be not set at all. In that case there, apart from that total fiat value on Wallet screen could be converted, all transactions end up with the `Error: Does not support name`. 

#### Expected behavior
Currency always present in the account. The default value is USD.

#### Actual behavior
screen shot 2018-08-17 at 12 48 12

### Reproduction
- Open Status and log in to account with no currency set (for testing we may simulate it in the code probably with set account to fake currency)
- Navigate to Account and check there is USD set as currency value
- Navigate to Wallet and verify USD is set
- Send a transaction to contact from 1-1 chat and from `Wallet` -> `Send transaction`

### Solution
We need to default currency to USD (same like it's set with new account creation/recover) if the currency for the account is not set.

### Additional Information
* Status version: Develop 0.9.26 (6401), relase 0.9.25

#### Logs

TF session: 
- `Error: Does not support name` at 06:08 
- No currency set for account could be seen at 10:25 
https://app.testfairy.com/projects/4803590-status/builds/8558753/sessions/4398751862/?accessToken=9DWD42R828MoKQwn7asGhFXY5kY

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803590-status/builds/8558753/sessions/4398751862/?accessToken=9DWD42R828MoKQwn7asGhFXY5kY">TestFairy: Status 0.9.26 (6396) @2018</a></strong></div></blockquote>

goranjovic:
fixes #5542 

### Summary:

Uses only one function to get the currently selected fiat currency and properly returns USD when either unset or nil.

### Steps to test:
- Open Status in dev mode
- In  REPL use following code to unset the currency:

```
(require '[re-frame.core :as re-frame])
(re-frame/dispatch [:wallet.settings/set-currency nil])
```

- Now navigate to Profile and verify that the currency is USD
- Send some transaction to someone 

status: ready

### DIFF
--- a/src/status_im/chat/commands/impl/transactions.cljs
+++ b/src/status_im/chat/commands/impl/transactions.cljs
@@ -21,6 +21,7 @@
             [status-im.utils.money :as money]
             [status-im.ui.screens.wallet.db :as wallet.db]
             [status-im.ui.screens.wallet.choose-recipient.events :as choose-recipient.events]
+            [status-im.ui.screens.currency-settings.subs :as currency-settings.subs]
             [status-im.models.transactions :as wallet.transactions]
             [status-im.ui.screens.navigation :as navigation]))
 
@@ -228,7 +229,7 @@
 (defn- inject-network-price-info [{:keys [amount asset] :as parameters} {:keys [db]}]
   (let [{:keys [chain prices]} db
         currency               (-> db
-                                   (get-in [:account/account :settings :wallet :currency] :usd)
+                                   currency-settings.subs/get-currency
                                    name
                                    string/upper-case)]
     (assoc parameters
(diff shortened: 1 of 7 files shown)


janherich (approved):

dmitryn (approved):

yevh-berdnyk:
@goranjovic all e2e tests failed due to the alert 'Hello. Thanks for choosing Status...' which is not expected to be there

asemiankevich:
@yevh-berdnyk its me, i gave wrong build to the job

asemiankevich:
i cant repro the issue in this PR.
also , i would like to confirm this @pablanopete . Please check on your side https://status-im.ams3.digitaloceanspaces.com/StatusIm-180830-123408-7e71a9-pr.apk

asemiankevich:
looks good to me

Issue #5563, opened 2018-08-20
java.lang.NoSuchMethodError: No static method SendTransaction(Ljava/lang/String;Ljava/lang/String;) and crash if send ETH from Wallet 


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: Android only. Sent ETH from Wallet and got crash after signing transaction:
``` 
Error java.lang.NoSuchMethodError: No static method SendTransaction(Ljava/lang/String;Ljava/lang/String;)Ljava/lang/String; in class Lcom/github/status_im/status_go/Statusgo; or its super classes (declaration of 'com.github.status_im.status_go.Statusgo' appears in base.apk!classes84.dex)
	at im.status.ethereum.module.StatusModule$8.run(StatusModule.java:517)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
	at java.lang.Thread.run(Thread.java:764) 
```
Note that same scenario for wallet refactoring PR https://github.com/status-im/status-react/pull/5455#issuecomment-413874062 on Android works fine 
#### Expected behavior
[comment]: # (Describe what you expected to happen.)
Tx is sent
#### Actual behavior
[comment]: # (Describe what actually happened.)
Crash
### Reproduction
[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status (Ropsten)
- Wallet -> Send transaction 0.0001 to the contact
- Sign transaction by entering valid passowrd

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: nightly build Aug 20
* Operating System: Android 9, Pixel XL; Android 6.0.1 Galaxy S6, also on Android 7 and 8
on iOS 11.0.3 and iOS 11.4 - all is fine, tx is sent as expected
#### Logs

https://app.testfairy.com/projects/4803622-status/builds/8562817/sessions/4398948981/?accessToken=kAy3uOekgBIzGbYeSRzSfg5Q-s4

[geth.log](https://github.com/status-im/status-react/files/2302015/geth.log)

[...]

PR #5714, opened 2018-08-30
Add `clean` makefile target to ensure build leftover state is removed

pedropombeiro:
Part of #5563 

This PR adds a `make clean` command to status-react, and calls it from Jenkins as well as `scripts/bundle-status-go.sh` (as we've come across old bindings being used in that scenario). There are some changes to how `desktop.groovy` works, so I'd appreciate @MaxRis's opinion on whether we're good with that.

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,13 +26,16 @@ HELP_FUN = \
 
 # Main targets
 
+clean: ##@prepare Remove all output folders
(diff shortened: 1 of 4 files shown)


jakubgs (approved):
NICE

maxhora (approved):
@PombeirP looks great!

PR #5718, opened 2018-08-30
Simplify setup script and fix Android build when setting up from scratch

pedropombeiro:
This PR fixes a problem that still persisted when running `make setup && make release-android` after having cleaned up the repo state.

- gradle was being called without `npm install` having had a chance to execute;
- `./android` folder still contained gradle state which caused gradle to believe that the NDK build step of react-native was up-to-date, leading to errors downstream;
- simplified `scripts/setup`, since it was doing the same as the make file targets.

One thing that needs to be confirmed in the review is the assumption that `mvn -f modules/react-native-status/ios/RCTStatus/pom.xml dependency:unpack` in the setup script can be replaced with `mvn -f modules/react-native-status/ios/RCTStatus dependency:unpack` from `make prepare-ios`. I assume so, but would prefer to have confirmation from someone who's more knowledgeable about maven.

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@
 help: ##@other Show this help
 	@perl -e '$(HELP_FUN)' $(MAKEFILE_LIST)
 
+OS := $(shell uname)
+
 # This is a code for automatic help generator.
 # It supports ANSI colors and categories.
 # To add new item into help output, simply add comments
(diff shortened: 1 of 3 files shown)


jakubgs (approved):
Not sure about your `mvn` question but other than that looks good.

yenda (approved):

PR #5728, opened 2018-09-03
View & preview implemented

janherich:

PR adding demo bits for the extensions `/collectible`

### Steps to test:
- Nothing to test, should not affect normal app flow at all

status: ready


flexsurfer (approved):
why did you move styles?

janherich:
@flexsurfer because they should not be in host code at all, ideally all those views where styles are used (nft suggestions, token suggestions) will be implemented in the extension itself, unfortunately state of the pluto project won't allow it now (but it will in the future).

jeluard (approved):

PR #5734, opened 2018-09-03
[fix #5729] ask for PN permissions when account recovered

### LINKED ISSUE #5729: No push notifications on recovered account on iOS
### Description

*Type*: Bug

*Summary*: on iOS: there is no push notifications if user recovered account. Most likely because we do not ask for push notifications permissions. There was same issue in the past fixed in https://github.com/status-im/status-react/pull/4349

#### Expected behavior
After user recovered account there is same popup asking about push notification permissions as it's shown for new account. If user allows it then for new message received in 1:1 chat with a contact there is PN shown if app is on background or closed.
#### Actual behavior
No PNs, no question about givving permissions
### Reproduction

- Uninstall app
- Install app
- Recover account
- When recovered - check if app asks to allow notifications
- Allow notifications
- check you get PN if app is closed or in background when new msg is received in 1:1 chat

### Additional Information
* Status version:  0.9.27
* Operating System: iOS

dmitryn:
fixes #5729

### Summary:

Ask for PN permissions when account recovered

status: ready

### DIFF
--- a/src/status_im/data_store/realm/schemas/base/core.cljs
+++ b/src/status_im/data_store/realm/schemas/base/core.cljs
@@ -7,7 +7,8 @@
             [status-im.data-store.realm.schemas.base.v6.core :as v6]
             [status-im.data-store.realm.schemas.base.v7.core :as v7]
             [status-im.data-store.realm.schemas.base.v8.core :as v8]
-            [status-im.data-store.realm.schemas.base.v9.core :as v9]))
+            [status-im.data-store.realm.schemas.base.v9.core :as v9]
+            [status-im.data-store.realm.schemas.base.v10.core :as v10]))
 
 ;; put schemas ordered by version
 (def schemas [{:schema        v1/schema
@@ -36,4 +37,7 @@
                :migration     v8/migration}
               {:schema        v9/schema
                :schemaVersion 9
-               :migration     v9/migration}])
+               :migration     v9/migration}
+              {:schema        v10/schema
+               :schemaVersion 10
+               :migration     v10/migration}])
--- a/src/status_im/data_store/realm/schemas/base/v10/account.cljs
+++ b/src/status_im/data_store/realm/schemas/base/v10/account.cljs
@@ -1,0 +1,30 @@
+(ns status-im.data-store.realm.schemas.base.v10.account)
+
+(def schema {:name       :account
+             :primaryKey :address
+             :properties {:address                :string
+                          :public-key             :string
+                          :name                   {:type :string :optional true}
+                          :email                  {:type :string :optional true}
+                          :status                 {:type :string :optional true}
+                          :debug?                 {:type :bool :default false}
+                          :photo-path             :string
+                          :signing-phrase         {:type :string}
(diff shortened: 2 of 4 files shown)


janherich (approved):

yenda (approved):

rasom (approved):

Serhy:
Looks good @dmitryn and original issue desribed in #5729 has been fixed!
There is one case that would be good to take into account as well: 
Since the issue was in 0.9.26 Release (and in 0.9.27 Release) version, some users may recover their accounts from these apps versions and have no ability to manage PN notifications. After upgrade, such accounts still won't have control to enable PN (because at the moment we ask for permissions only on account recover/creation process). 

Could we, in scope of this fix, check whether application on iOS has 'Notifications' option present in app settings, and if not - to raise `Status would like to send you Notifications`: `Allow` / `Don't Allow` pop-up (that should create an ability to manage notifications on iOS in that case? 

**To reproduce:**
1) Remove status app from iOS device
2) Install 0.9.26 / 0.9.27 Release app version
3) Recover an account (Note: we don't ask for notification permissions)
4) Upgrade to app version with current fix ( app version from https://github.com/status-im/status-react/pull/5734#issuecomment-418145638)
5) Log in to account recovered in 0.9.26/0.9.27 above (3rd step)

dmitryn:
@Serhy i've moved PN request to sign in step so users with 0.9.26/27 who were affected would be asked about permissions. Please take a look.

Serhy:
Yes, @dmitryn, it fixes all issues with PN Notification permissions on account login/recover, but now we display `Thanks for choosing Status. We are now in Beta ....` pop-up each time user logs in to account (should be displayed only one time after account Recover/Creation).
TF session for how it looks: https://app.testfairy.com/projects/4803590-status/builds/8583452/sessions/4400760036/?accessToken=sUIAoAU-KrSNoagjjWCnBZEo6UM

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803590-status/builds/8583452/sessions/4400760036/?accessToken=sUIAoAU-KrSNoagjjWCnBZEo6UM">TestFairy: Status 0.9.28 (7796) @2018</a></strong></div></blockquote>

dmitryn:
@Serhy hm, that's weird bug. After warning is shown we set a flag in db so it shouldn't be shown again.

Just checked the code and i verified that `:update-mainnet-warning-shown` event is dispatched after user closes warning. @janherich do you have an idea why realm write might be not working here? https://github.com/status-im/status-react/blob/develop/src/status_im/ui/screens/accounts/events.cljs#L52 I checked Realm account record after write, and flag is not updated there for some reason.

yenda:
@dmitryn I think that is the bug @cammellos spotted yesterday in the DB migration, v8 was skipped because of a bad import

dmitryn:
Fixing this bug in another PR https://github.com/status-im/status-react/pull/5779

dmitryn:
@Serhy I've added fix for mainnet warning, please take a look.

PR #5736, opened 2018-09-04
Enhance command hook with `on-send` and `on-receive`

janherich:

Enhances command `AppHook` with possibility to pass `on-send`/`on-receive` functions which hook-in to the sending/receiving process, like for example here - https://gist.github.com/janherich/92747e730b2e115bcbe145114d024e66.

Requires code in https://github.com/status-im/pluto/pull/50 pluto PR.

### Testing notes (optional):
No QA testing necessary

status: ready

<blockquote><div> Gist</div><div><strong><a href="https://gist.github.com/janherich/92747e730b2e115bcbe145114d024e66">extension.edn</a></strong></div><div>GitHub Gist: instantly share code, notes, and snippets.</div></blockquote>

### DIFF
--- a/src/status_im/chat/commands/core.cljs
+++ b/src/status_im/chat/commands/core.cljs
@@ -139,19 +139,23 @@
                      :description   :string
                      :short-preview :view
                      :preview       :view
+                     :on-send?      :function
+                     :on-receive?   :function
                      :parameters    [{:id           :keyword
                                       :type         {:one-of #{:text :phone :password :number}}
                                       :placeholder  :string
                                       :suggestions? :component}]})
-    (hook-in [_ id {:keys [description scope parameters preview short-preview]} cofx]
+    (hook-in [_ id {:keys [description scope parameters preview short-preview on-send on-receive]} cofx]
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

jeluard:
We probably want to hold this one back while we have more comprehensive model for those type of `on-` events.

janherich:
@jeluard I understand and I agree with no rush, but just for the clarification, `on-send` is something totally different then `:on-click`/`:on-change`... view defined event handler.
While the former defines just hook-in point to the message sending process (could be named differently, like `send-hook`), the latter is external trigger point which starts some UI action, something totally different.

jeluard:
From end user perspective they are not. If that does not fit the implementation then it would make sense to amend it.
We have new concepts here and don’t want to be tied by our current state.

This PR does not have any end user effect  I would suggest to close it for now.

vitvly:
@janherich @jeluard Did you reach a consensus? Do we want to merge it or suspend for now?

jeluard:
Let's close it for now as the design for this is still WIP

PR #5737, opened 2018-09-04
Update status-go to 0.14.0 (Keys Derivation Padding Fix)

adambabik:
The most important change here is fixing a bug with child keys derivation. 

In this release, we would like to allow users to verify if their accounts were affected and we have identified a few scenarios:

## Scenario 1: accounts is unaffected
1. User has already Status installed with an account created,
2. She upgrades and instead of signing in, she recovers the existing account,
3. If the account is unaffected, a message should appear saying that such an account already exists and should be signed in instead of creating a new account.

**Issues:**
1. What will happen if a different password will be used?

## Scenario 2: account is affected

In order to test this scenario, use [this seed phrase](https://github.com/status-im/status-go/pull/1139/files#diff-898dc5d11da1fde8b22d8a9949456000R660).

1. User has already Status installed with an account created,
2. He upgrades and instead of signing in, he recovers the existing account,
3. If the account is affected, a new account should be created,
4. He should see two account after signing out and that indicates that his account is affected and his founds should be transferred.

**Issues:**
1. How can we avoid a situation when a user misspelled a word and recovered a wrong account? Should we only emphasize this in the release notes instruction?
2. Is there any indicator which account was created earlier? Do we make sure the order of created accounts is always kept?
3. Is it possible to delete an account?

### DIFF
--- a/modules/react-native-status/android/build.gradle
+++ b/modules/react-native-status/android/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     implementation 'com.instabug.library:instabug:3+'
     implementation 'status-im:function:0.0.1'
 
-    String statusGoVersion = 'v0.13.1'
+    String statusGoVersion = 'v0.14.1'
     final String statusGoGroup = 'status-im', statusGoName = 'status-go'
 
     // Check if the local status-go jar exists, and compile against that if it does
--- a/modules/react-native-status/desktop/CMakeLists.txt
+++ b/modules/react-native-status/desktop/CMakeLists.txt
@@ -37,7 +37,7 @@ ExternalProject_Add(StatusGo_ep
   PREFIX ${StatusGo_PREFIX}
   SOURCE_DIR ${StatusGo_SOURCE_DIR}
   GIT_REPOSITORY https://github.com/status-im/status-go.git
-  GIT_TAG v0.13.1
+  GIT_TAG v0.14.1
   BUILD_BYPRODUCTS ${StatusGo_STATIC_LIB}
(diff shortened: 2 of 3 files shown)


gravityblast:
> If the account is unaffected, a message should appear saying that such an account already exists

I'm not sure we can do this now because internally status-go logs you in automatically if the account already exists and the password is the same. So I think status-go can't show a message in this situation unless I'm missing something

yenda (approved):

gravityblast (approved):

adambabik:
> I'm not sure we can do this now because internally status-go logs you in automatically if the account already exists and the password is the same. So I think status-go can't show a message in this situation unless I'm missing something

I am talking here about a situation when I recover an account. `RecoverAccount` binding is used in that case. I tried it now and it logged me in using the existing account. However, when I passed a different password, I am stuck on the recover screen.

adambabik:
@annadanchenko can we test this PR on both `eth.staging` and `eth.beta`? After it is released, I'd like to update `eth.beta`.

gravityblast:
> I am talking here about a situation when I recover an account. RecoverAccount binding is used in that case. I tried it now and it logged me in using the existing account. However, when I passed a different password, I am stuck on the recover screen.

yes you are right!

churik:
Tested on desktop build (OS: Mac OSx High Sierra 10.13.6):

- installation
- account creating/restoring
- login/logout
- fetching 24 h history for public channels
- adding user to contact
- 1-1 chat
- mailservers switching
- navigating to user profile
- navigating to user profile from public chat
- clear history / delete chat and relogin

*NOTE:* existing accounts in `StatusIm` folder can be affected, so if you see error `We were not able to decrypt your data`  - only way to workaround it is:
- close app
- delete `StatusIm` folder at all 
- recover your account or create new one .

PR #5741, opened 2018-09-04
Ignore Desktop builds for release.

mandrigin:
Currently, we don't release desktop betas and it isn't given that the
release schedule will be synchronized. So, for simplicity of support,
let's not build desktop against release branches.

status: ready

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -24,10 +24,14 @@ pipeline {
     }
     stage('Build') {
       parallel {
-        stage('MacOS') { steps { script {
+        stage('MacOS') {
+        when { expression { cmn.getBuildType() != 'release' } }
+        steps { script {
           osx = cmn.buildBranch('status-react/combined/desktop-macos', cmn.getBuildType())
         } } }
-        stage('Linux') { steps { script {
+        stage('Linux') {
+        when { expression { cmn.getBuildType() != 'release' } }
+        steps { script {
           nix = cmn.buildBranch('status-react/combined/desktop-linux', cmn.getBuildType())
         } } }
         stage('iOS') { steps { script {
@@ -44,8 +48,11 @@ pipeline {
     stage('Archive') {
       steps { script {
         sh('rm -f pkg/*')
-        cmn.copyArts('status-react/combined/desktop-macos', osx.number)
-        cmn.copyArts('status-react/combined/desktop-linux', nix.number)
+
+        if (cmn.getBuildType() != 'release') {
+          cmn.copyArts('status-react/combined/desktop-macos', osx.number)
+          cmn.copyArts('status-react/combined/desktop-linux', nix.number)
+        }
         cmn.copyArts('status-react/combined/mobile-android', dro.number)
         cmn.copyArts('status-react/combined/mobile-android', e2e.number)
         archiveArtifacts('pkg/*')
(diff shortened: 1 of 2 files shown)


ci/Jenkinsfile.combined
```diff
@@ -25,10 +25,14 @@ pipeline {
     stage('Build') {
       parallel {
         stage('MacOS') { steps { script {
-          osx = cmn.buildBranch('status-react/combined/desktop-macos', cmn.getBuildType())
```
jakubgs: This is not how your use decarative pipelines:
https://jenkins.io/doc/book/pipeline/syntax/#when
So in this case it would be something like:
```
        stage('MacOS') {
          when { expression { cmn.getBuildType() != 'release' } }
          steps { script {
            osx = cmn.buildBranch('status-react/combined/desktop-macos', cmn.getBuildType())
        } } }
```

maxhora (approved):

ci/Jenkinsfile.combined
```diff
@@ -55,8 +62,14 @@ pipeline {
...
         apkUrl = cmn.uploadArtifact(findFiles(glob: "pkg/*${cmn.getBuildType()}.apk")[0].path)
-        dmgUrl = cmn.uploadArtifact(findFiles(glob: 'pkg/*.dmg')[0].path)
-        appUrl = cmn.uploadArtifact(findFiles(glob: 'pkg/*.AppImage')[0].path)
+
+        if (cmn.getBuildType() != 'release') {
+          dmgUrl = cmn.uploadArtifact(findFiles(glob: 'pkg/*.dmg')[0].path)
+          appUrl = cmn.uploadArtifact(findFiles(glob: 'pkg/*.AppImage')[0].path)
+        } else {
+          dmgUrl = '-'
+          appUrl = '-'
```
jakubgs: If you set those as `-` then the `misc/status-im.github.io-update_env` job will run with that as the parameter values and set those to be the links on the https://status-im.github.io/nightly/ page.

jakubgs (changes requested):

ci/Jenkinsfile.combined
jakubgs: the values should be `null` or empty string. Then the values inside of https://github.com/status-im/status-im.github.io/blob/develop/env.sh will not be changed.

mandrigin: you can't really get them there because we don't update the website for release builds, but point taken, I'll tweak it.

vkjr (approved):

ci/Jenkinsfile.combined
jakubgs: Oh, good point, forgot Publish runs only for nightlies.

ci/Jenkinsfile.combined
```diff
@@ -25,9 +25,11 @@ pipeline {
     stage('Build') {
       parallel {
         stage('MacOS') { steps { script {
+          when { expression { cmn.getBuildType() != 'release' } }
```
jakubgs: The `when` block should be on the same level as the `steps` block, see example:
https://jenkins.io/doc/book/pipeline/syntax/#when-example

jakubgs (changes requested):

jakubgs (approved):
Thanks!
This makes the jenkinsfile quite ugly, but hopefully soon we'll have desktop releases too so these kind of exceptions won't be necessary.

PR #5742, opened 2018-09-04
feature #4747 - ens usernames in wallet send

### LINKED ISSUE #4747: ENS support in wallet and profile
### User Story

As a user, I want to use and see ENS names so that I can more easily interact with wallet, accounts, and profiles.

### Description

*Type*: Feature

*Summary*: ENS usernames give users a human-readable way of interacting with Ethereum, and can improve Status usability and utility.

#### Expected behavior
Send a transaction by entering an ENS name.

Recipient address should still be supported, but ENS names are more user-friendly. Let's call out support for them by re-labeling the menu button. 

image

#### Actual behavior

Wallet supports only recipient address. Users are required to copy and paste these addresses to share them with each other.

goranjovic:
fixes #4747 

### Summary:

Add ENS name resolution in Wallet Send

### Steps to test:
- Wallet > Send
- Choose recipient by address
- enter a registered ENS username
- Send the tx

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/choose_recipient/events.cljs
+++ b/src/status_im/ui/screens/wallet/choose_recipient/events.cljs
@@ -4,7 +4,9 @@
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.ethereum.eip681 :as eip681]
             [status-im.utils.handlers :as handlers]
-            [status-im.utils.money :as money]))
+            [status-im.utils.money :as money]
+            [status-im.utils.ethereum.ens :as ens]
+            [re-frame.core :as re-frame]))
 
 (handlers/register-handler-db
  :wallet/toggle-flashlight
@@ -56,6 +58,27 @@
 (defn use-default-eth-gas [fx]
   (assoc-in fx [:db :wallet :send-transaction :gas]
             ethereum/default-transaction-gas))
+
+(re-frame/reg-fx
+ :resolve-address
+ (fn [{:keys [web3 registry ens-name cb]}]
+   (ens/get-addr web3 registry ens-name cb)))
+
+(handlers/register-handler-fx
+ :wallet.send/set-recipient
+ (fn [{:keys [db]} [_ recipient]]
+   (let [{:keys [web3 network]} db
+         network-info (get-in db [:account/account :networks network])
+         chain (ethereum/network->chain-keyword network-info)]
+     (if (ens/is-valid-eth-name? recipient)
+       {:resolve-address {:web3     web3
+                          :registry (get ens/ens-registries chain)
+                          :ens-name recipient
+                          :cb       #(re-frame/dispatch [:wallet.send/set-recipient %])}}
+       (if (ethereum/address? recipient)
+         {:db       (assoc-in db [:wallet :send-transaction :to] recipient)
+          :dispatch [:navigate-back]}
+         {:show-error (i18n/label :t/wallet-invalid-address {:data recipient})})))))
 
 (handlers/register-handler-fx
  :wallet/fill-request-from-url
--- a/src/status_im/utils/ethereum/stateofus.cljs
+++ b/src/status_im/utils/ethereum/stateofus.cljs
@@ -1,18 +1,11 @@
 (ns status-im.utils.ethereum.stateofus
   (:refer-clojure :exclude [name])
   (:require [clojure.string :as string]
-            [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.ethereum.ens :as ens]))
 
 (defn is-valid-name? [ens-name]
-  (string/ends-with? ens-name ".stateofus.eth"))
-
-(defn addr [web3 registry ens-name cb]
-  {:pre [(is-valid-name? ens-name)]}
-  (ens/resolver web3
-                registry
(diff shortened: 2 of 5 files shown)


src/status_im/ui/screens/wallet/choose_recipient/events.cljs
```diff
@@ -57,9 +61,31 @@
...
+                          :cb       #(re-frame/dispatch [:wallet.send/set-recipient %])}}
+       (if (ethereum/address? recipient)
+         {:db       (assoc-in db [:wallet :send-transaction :to] recipient)
+          :dispatch [:navigate-back]}
+         {:show-error (i18n/label :t/wallet-invalid-address {:data recipient})})))))
+
 (handlers/register-handler-fx
  :wallet/fill-request-from-url
  (fn [{{:keys [network] :as db} :db} [_ data origin]]
+   (js/alert data)
```
flexsurfer: leftover

goranjovic: :pig:

src/status_im/utils/ethereum/stateofus.cljs
```diff
@@ -4,11 +4,16 @@
...
+       (string/ends-with? ens-name ".eth")))
+
 (defn is-valid-name? [ens-name]
-  (string/ends-with? ens-name ".stateofus.eth"))
+  (and ens-name
+       (string/ends-with? ens-name ".stateofus.eth")))
 
 (defn addr [web3 registry ens-name cb]
-  {:pre [(is-valid-name? ens-name)]}
+  {:pre [(is-valid-eth-name? ens-name)]}
```
dmitryn: Keep in mind that `:pre` assertions are removed in production builds since we have https://cljs.github.io/api/compiler-options/elide-asserts set to true

src/status_im/ui/screens/wallet/choose_recipient/events.cljs
```diff
@@ -57,6 +61,27 @@
   (assoc-in fx [:db :wallet :send-transaction :gas]
             ethereum/default-transaction-gas))
 
+(re-frame/reg-fx
+ :resolve-address
+ (fn [{:keys [web3 registry ens-name cb]}]
+   (stateofus/addr web3 registry ens-name cb)))
```
flexsurfer: this doesn't look right, why stateofus care about .eth names

dmitryn (approved):
I'd move `wallet.send/set-recipient` to separate fn for easier testability, but it's not required since you don't have tests here anyway

src/status_im/utils/ethereum/stateofus.cljs
```diff
@@ -4,11 +4,16 @@
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.ethereum.ens :as ens]))
 
+(defn is-valid-eth-name? [ens-name]
```
flexsurfer: this as well

jeluard:
Is the idea to resolve any eth address? Or just stateofus?

src/status_im/ui/screens/wallet/choose_recipient/events.cljs
goranjovic: you mean like why is it in this namespace at all?

goranjovic: btw, `.eth` names can also add status whisper pubkey, so following that logic, there should be no `stateofus` ns - only ens

goranjovic:
@jeluard any eth address

jeluard:
@goranjovic Why the `stateofus` ns then?

goranjovic:
@jeluard because I continued to build on code used by chat which is `stateofus` only, for now at least. Will move to main ens ns as per your and @flexsurfer 's comments

jeluard:
Ah ok got it.

src/status_im/utils/ethereum/stateofus.cljs
yenda: also the resolver is for stateofus names so I'm not sure it will work with eth names

yenda: nevermind you are probably right about stateofus namespace

goranjovic: @yenda it works for eth names, no idea how

goranjovic:
@flexsurfer @jeluard @yenda addressed the comments, please re-review

yenda (approved):

asemiankevich:
@dmitryn weird tooltip is shown when i bypass the onboarding

image


UPD seems not related to PR cause i see similar in release build
image

asemiankevich:
Apart the tooltip above, i was able to:
- send ETH from wallet to `nastya3.stateofus.eth` and see balance updated as well as transaction history
- send tokens from wallet to `nastya3.stateofus.eth` and see balance updated. I dont see incoming txns in history but this could happen due https://github.com/status-im/status-react/issues/5749

Correct address resolves by ENS username which is shown then in Recipient field. We have a disussion about details here in [slack](https://status-im.slack.com/archives/CA7R1RK0C/p1536060538000100) but for now this is the only functionality implemented.

PR #5743, opened 2018-09-04
[fix #5063] calculate fiat amount on recipient side

### LINKED ISSUE #5063: Calculate fiat value of /send and /request on the recipient side
### Description

*Type*: Bug

*Summary*: We have a PR https://github.com/status-im/status-react/pull/5044 that introduces localization of `/send` and `/request` messages based on user preferences. So if I set up my wallet in EURO all the transaction messages will be shown in the currency I have selected.

The issue it the following: when i send a request to a person in EUR ( i have my wallet in EUR) - the recipient of the request will see the request incoming in EUR even he has configured his wallet to be shown in AED.
- this is confusing from a user point of view, for example, I could not be aware of what RUR currency is for
- this is a potential security breach because it allows understanding where the request is coming from

#### Expected behavior

- Fiat value of send and request are being calculated on recipient side so they always will be shown in the currency user selected for himself
- Sender still see his requests and sends in his preferred currency as well 

#### Actual behavior

- requests are coming in the currency they were sent from no matter which currency the recipient of the request is using
- respond to such requests are shown in currency of the sender, so I have wallet in EUR and respong to AED request - the respond (actual send) will be shown for me in EUR as well as for recipient who wants AEDs. 

### Reproduction

- Open wallet and configure it to use any non default currency on device 1
- Open wallet and configure it to use any non default currency on device 2
- Send a /request from device 1 to device 2
- Respond to this request from device 2 to device 1

image

dmitryn:
fixes #5063 

### Summary:

Calculate fiat amount on recipient side for /send /receive commands depending on account currency.

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -97,11 +97,28 @@
 (defn add-outgoing-status [{:keys [from] :as message} {:keys [db]}]
   (assoc message :outgoing (= from (:current-public-key db))))
 
+(defn set-fiat-amount [{:keys [content-type content] :as message} cofx]
+  (if (= constants/content-type-command content-type)
+    (let [account-currency (-> cofx
+                               (get-in [:db :account/account :settings :wallet :currency] :usd)
+                               name
+                               string/upper-case)
+          {:keys [amount asset]} (:params content)
+          prices (get-in cofx [:db :prices])]
+      (-> message
+          (assoc-in [:content :params :fiat-amount] (money/fiat-amount-value amount
+                                                                             (keyword asset)
+                                                                             (keyword account-currency)
+                                                                             prices))
+          (assoc-in [:content :params :currency] account-currency)))
+    message))
+
 (defn- add-message
   [batch? {:keys [chat-id message-id clock-value content] :as message} current-chat? {:keys [db] :as cofx}]
   (let [prepared-message (-> message
                              (prepare-message chat-id current-chat?)
-                             (add-outgoing-status cofx))]
+                             (add-outgoing-status cofx)
+                             (set-fiat-amount cofx))]
     (let [fx {:db            (cond->
                               (-> db
                                   (update-in [:chats chat-id :messages] assoc message-id prepared-message)
--- a/test/cljs/status_im/test/chat/models/message.cljs
+++ b/test/cljs/status_im/test/chat/models/message.cljs
@@ -169,3 +169,20 @@
              (get-in fx2 [:db :chats "chat-id" :messages])))
       (is (= {}
              (get-in fx2 [:db :chats "chat-id" :message-groups]))))))
+
+(deftest set-fiat-amount
+  (let [message {:content-type "text/plain"}
+        command-message {:content-type "command"
+                         :content      {:params {:amount "2"
+                                                 :asset  "ETH"}}}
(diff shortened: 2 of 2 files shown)


goranjovic (approved):

src/status_im/chat/models/message.cljs
```diff
@@ -97,11 +97,28 @@
...
+          {:keys [amount asset]} (:params content)
+          prices (get-in cofx [:db :prices])]
+      (-> message
+          (assoc-in [:content :params :fiat-amount] (money/fiat-amount-value amount
+                                                                             (keyword asset)
+                                                                             (keyword account-currency)
+                                                                             prices))
+          (assoc-in [:content :params :currency] account-currency)))
+    message))
+
```
janherich, line 115: This is not a good place for the logic - we shouldn't pollute generic message handling with code specific to particular command, right place for it is `on-receive` protocol method of the `PersonalRequestCommand`.

janherich (changes requested):

src/status_im/chat/models/message.cljs
dmitryn: @janherich i thought about `on-receive`, but that call happens after message got saved to realm https://github.com/status-im/status-react/blob/c35a651f1599c76c0131e4af46eb9504b57ac10f/src/status_im/chat/models/message.cljs#L173 Do you think we can safely perform `on-receive` before `add-message-fn` ?

dmitryn: @janherich or i could produce `data-store/tx` effect in `on-receive` and it will be then merged into all effects by `merge-fx`

janherich: You are right, sorry for jumping the gun.
Let's do it the "simple way" now and in the meantime, I will think of something better, how generic receive flow can leverage `on-receive` protocol method in a more flexible way (I already have an idea).

janherich (approved):

src/status_im/chat/models/message.cljs
dmitryn: Sounds good. Also, i'd like to mention that my changes applied to both /receive and /send command, i.e. we calculate fiat amount on recipient side for both of them.

asemiankevich:
@dmitryn 
```NAME: "Error"
MESSAGE: Error: Doesn't support name: 

This error is located at:
    in asset-selector
    in reagent90
    in RCTView
    in RCTScrollContentView
    in RCTScrollView
    in ScrollView
    in RCTView
    in RCTView
    in KeyboardAvoidingView
    in reagent4
    in reagent86
    in reagent85
    in send-transaction
    in reagent84
    in RCTView
    in main
    in root
    in reagent1
    in RCTView
    in RCTView
    in t
```

STR:
- open status
- switch to ropsten and requst some ETH
- go to 1x1 chat and send 0.001 ETH in chat

Error appears when i tap `Sign transaction -> Got it`

https://app.testfairy.com/projects/4803590-status/builds/8582851/sessions/4400732362/?accessToken=MbTV9ylPNHLG58SbZMmw70i0WdQ

image

dmitryn:
@asemiankevich issue fixed. Please take a look

asemiankevich:
looks good to me now. When i send or request txn in my custom currency - it is shown on recipient's device is his selected currency. Same when responding. Works in chat and wallet as well. 

image

image

PR #5752, opened 2018-09-04
Set .envfile path based on build type

maxhora:
- .env file for prod builds is moved to `deployment/.env` and copied to the distribution packages;
- Customize path to .env file based on build type.

Depends on changes in https://github.com/status-im/react-native-config/pull/2

Note: If to run app with .env file from the repo's root, app doesn't start correctly in this case (because of some settings).


desktop_files/package.json
```diff
@@ -55,7 +55,7 @@
     "react-native": "git+https://github.com/status-im/react-native-desktop.git",
     "react-native-background-timer": "2.0.0",
     "react-native-camera": "0.10.0",
-    "react-native-config": "git+https://github.com/status-im/react-native-config.git",
+    "react-native-config": "git+https://github.com/MaxRis/react-native-config.git",
```
maxhora: This change will be removed before merge

ci/desktop.groovy
```diff
@@ -187,6 +187,8 @@ def bundleMacOS(type = 'nightly') {
     sh 'chmod +x Status.app/Contents/Resources/ubuntu-server'
     sh 'cp ../desktop/bin/StatusIm Status.app/Contents/MacOS/Status'
     sh 'cp ../desktop/reportApp/reportApp Status.app/Contents/MacOS'
+    sh "cp ../deployment/.env  Status.app/Contents/Resources"
+    sh 'ln -sf ../Resources/.env Status.app/Contents/MacOS/.env'
```
gnl: Personally I'd prefer that there are no hidden files under `deployment` and `Resources`, so I'd go with:
```
sh 'cp ../deployment/env  Status.app/Contents/Resources'
sh 'ln -sf ../Resources/env Status.app/Contents/MacOS/.env'
```
...and of course everywhere else `.env` is referenced.

gnl (approved):

vitvly (approved):

churik:
Tested:
- creating new account
- restoring existing account

ci/desktop.groovy
maxhora: @gnl , makes sense, thanks!

maxhora: But why not just name the file "env" and set the appropriate file path to be loaded on app start.

vkjr (approved):

PR #5757, opened 2018-09-05
[#5180] Encrypt account's db with password

### LINKED ISSUE #5180: Encrypt account realm with user password
### User Story

As a user
I want to have each user account encrypted with a different key based on the user password
So that an attacker breaking the keychain will not be able to decrypt my data

### Description

At the moment we encrypt the database on disk using only the keychain. 
User data can be also encrypted with both the keychain and the user password (generating a key from both), so that compromising the keychain does not result in the compromise of all the messages.

#### Expected behavior

Realm account data is encrypted with their password as well.

#### Actual behavior
Realm account data is encrypted only with the keychain password.

### Solution
* Realm files should be correctly migrated to the new encryption key
* Additionally sensitive information should be moved from `base` to `account`, but can be done in a separate issue
* Nice to have: It would be good to take into consideration the fact that a user will be able to change password at some point, so encrypting the database with a new key might be a day-to-day operation.

rasom:
fix #5180

`keccak512-array` from `js-sha3` lib was used for deriving 64 bytes key from concatenated random key and password
`text-encoding` lib was added for utf8 to uintArray encoding (password should be encoded before concatenation)

status: ready

### DIFF
--- a/ci/mobile.groovy
+++ b/ci/mobile.groovy
@@ -48,6 +48,7 @@ def prep(type = 'nightly') {
       sh 'cp .env.jenkins .env'; break
   }
   common.installJSDeps('mobile')
+  sh 'cp patches/js_realm.hpp node_modules/realm/src/js_realm.hpp'
   /* install ruby dependencies */
   sh 'bundle install --quiet'
   /* install Maven dependencies */
--- a/clj-rn.conf.edn
+++ b/clj-rn.conf.edn
@@ -36,7 +36,9 @@
               "react-native/Libraries/vendor/emitter/EventEmitter"
               "react-native-background-timer"
               "react-native-fetch-polyfill"
-              "react-native-testfairy"]
+              "react-native-testfairy"
+              "text-encoding"
+              "js-sha3"]
 
  ;; Resoures
  :resource-dirs ["resources/images"]
--- a/desktop_files/package-lock.json
+++ b/desktop_files/package-lock.json
@@ -662,43 +662,6 @@
         "xtend": "4.0.1"
       }
     },
-    "accepts": {
-      "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.5.tgz",
-      "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
-      "requires": {
-        "mime-types": "2.1.18",
-        "negotiator": "0.6.1"
-      },
-      "dependencies": {
-        "mime-db": {
-          "version": "1.33.0",
-          "resolved": "https://registry.npmjs.org/mime-db/-/mime-db-1.33.0.tgz",
-          "integrity": "sha512-BHJ/EKruNIqJf/QahvxwQZXKygOQ256myeN/Ew+THcAa5q+PjyTTMMeNQC4DZw5AwfvelsUrA6B67NKMqXDbzQ=="
-        },
-        "mime-types": {
-          "version": "2.1.18",
-          "resolved": "https://registry.npmjs.org/mime-types/-/mime-types-2.1.18.tgz",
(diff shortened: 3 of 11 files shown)


janherich (approved):

yenda:
@rasom since keccak512 is sha3 why not use the function available in web3 instead of a library ?

rasom:
@yenda web3's `sha3` is Keccak-256 which returns 32-byte hash, but we need 64-byte key for realm's encryption. Correct me if i'm wrong. 

https://github.com/ethereum/wiki/wiki/JavaScript-API#web3sha3
https://realm.io/docs/javascript/latest/#encryption

yenda:
@rasom correct, keccak-512 is also in eth-lib which will be included in the project as a library of web3-utils. But it is currently blocked by desktop which needs to upgrade to react-native 0.56

rasom:
It will be easy to replace the function then, so i don't think this particular PR should be blocked.

alwx (approved):

lukaszfryc:
@rasom when I upgrade from 0.9.27 and log in, the app says that it could not decrypt my data. I tried to log in again, but I can't because this error appears again (both Android and iOS).




@rasom apart from smoke and upgrade tests, is there anything else I should focus on?

rasom:
@lukaszfryc the issue was most likely related to schema versions, please test the next build (i just rebased this brange onto develop).

> @rasom apart from smoke and upgrade tests, is there anything else I should focus on?

nothing more

lukaszfryc:
@rasom same thing happens on the latest build.

rasom:
it seems that `Realm.writeCopyTo` is broken

rasom:
I spent quite a lot of time trying to figure what is wrong here, and seems that `Realm.writeCopyTo` is broken. Try to reproduce running the next code in REPL on iOS simiulator:

```clojure
(require '[status-im.data-store.realm.core :as realm.core])
(let [schema      {:schema        [{:name       :chat
                                    :primaryKey :chat-id
                                    :properties {:chat-id :string}}]
                   :schemaVersion 1}
      ;; path to your simulator
      path        "...~/Library/Developer/CoreSimulator/Devices/A5C7447B-DD23-47CE-87B5-1F1CCE9FFC70/data/Containers/Data/Application/454FE884-03C4-41B8-8123-465BA4500932/Documents/databases/"
      db          (str path "db.realm")
      new-db      (str path "new_db.realm")
      ;; 01000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
      key         (let [arr (js/Int8Array. 64)]
                    (aset arr 0 1)
                    arr)
      original-db (realm.core/open-realm schema db key)]

  (.write
   original-db
   (fn []
     (.create original-db "chat" (clj->js {:chat-id "id"}))))

  (.writeCopyTo original-db new-db key)

  (let [copied-db (realm.core/open-realm schema new-db key)]
    (realm.core/close copied-db))
  (realm.core/close original-db))
```

result is

```
#object[Error Error: Unable to open a realm at path '/Users/romanvolosovskyi/Library/Developer/CoreSimulator/Devices/A5C7447B-DD23-47CE-87B5-1F1CCE9FFC70/data/Containers/Data/Application/454FE884-03C4-41B8-8123-465BA4500932/Documents/databases/new_db.realm': Invalid mnemonic. top_ref[0]: 4F1C1E1597D35B7C, top_ref[1]: 2CEE9640FC573440, mnemonic: 7D 29 4B 81, fmt[0]: 228, fmt[1]: 240, flags: B5.]
[native code]
status_im$data_store$realm$core$open_realm



eval code
eval@[native code]
figwheel$client$utils$eval_helper
```

There is a test for such case in realm js repo, but it it's commented out https://github.com/realm/realm-js/blob/v2.15.3/tests/js/realm-tests.js#L1292 :|

rasom:
`Realm.writeCopyTo` is fixed here https://github.com/realm/realm-js/pull/2020

lukaszfryc:
All works fine now

mandrigin:
@rasom can we have a follow-up issue to update `realm-js` and get rid of the patch? I hope to get the fix merged today...

PR #5761, opened 2018-09-05
Enhanced receive params

janherich:

Lately, it has been discovered that there is no easy way to plug-in custom command parameters enrichment on the recipient side.

This is resulting again in bloated code with command checking/branching where it should not be at all, eq in the generic `chat/models/message` ns which should be only concerned with generic message handling and nothing command specific - https://github.com/status-im/status-react/pull/5743.

The solution is easy, we already have `EnhancedParams` protocol with method `enhance-parameters` for commands which wish to enrich their parameters during sending (like adding sender network there), we just add the the receive equivalent and rename it accordingly, eq `enhance-send-parameters` and `enhance-receive-parameters` are now two methods of the protocol.

### Testing notes (optional):
Only very superficial regression testing is necessary (ensuring that `/send` and `/request` commands are correctly sent/received just like before)

status: ready

### DIFF
--- a/src/status_im/chat/commands/impl/transactions.cljs
+++ b/src/status_im/chat/commands/impl/transactions.cljs
@@ -306,8 +306,9 @@
                           :success-event :wallet/update-gas-price-success
                           :edit?         false}}))
   protocol/EnhancedParameters
-  (enhance-parameters [_ parameters cofx]
-    (inject-network-price-info parameters cofx)))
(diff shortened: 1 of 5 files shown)


rasom (approved):

goranjovic (approved):

Serhy:
Looks good, @janherich ! 
Autotests reviewed and retested.

PR #5788, opened 2018-09-08
Automate code-signing of macOS bundle

### LINKED ISSUE #5421: Sign Jenkins build artifacts to avoid security warnings

gnl:
Fixes #5421 

## Review notes

The `encrypt-keychain.sh` and `gen-random-passphrase.sh` scripts are only used manually on a trusted host for the original encryption of the keychain file.

## Testing notes

* Application should open and launch without security warnings except for `App was downloaded from the internet`
* If the application was previously installed, there should be a Keychain access prompt on the first launch, but not on any subsequent launches or after future upgrades
* If the application was not previously installed, there should be no Keychain access prompts whatsoever.


deployment/macos/encrypt-keychain.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
vkjr: When the script is called? I didn't find its mentioning among commited files.

vkjr (approved):

deployment/macos/encrypt-keychain.sh
gnl: This isn't called anywhere, it's only meant to be used manually client-side on a trusted host where the original GPG-encryption of the keychain file is performed.

vkjr: Thanks for clarification :)

maxhora (approved):

corpetty (approved):

jakubgs:
Looks like it works, nice work!

gnl:
Squashed and rebased, ready to merge after QA confirmation.

chadyj:
Worked for me! Yay!

Let's merge :)

PR #5797, opened 2018-09-10
[#5745] validate account recover result

### LINKED ISSUE #5745: Recovering the same account with a different password is stuck
### Description

*Type*: Bug

*Summary*: When I recover an account that already exists in the app but I use a different password, I am stuck on the recover screen.

#### Expected behavior
I should get a notification that such an account already exists and the password does not match.

If I forgot the password, what should I do? Can I delete the previous account and recover it with a different password?

#### Actual behavior
I am stuck on the recover screen without any message and I need to kill the app (remove it from the background) to try to recover a different account or recover it with a correct password.

### Reproduction

- Open Status,
- If you don't have any accounts, create one,
- Logout from the current account and go to the recover screen,
- Recover the existing account but use a different password this time.

### Additional Information
* Status version: 0.9.27, 0.9.28 nightly
* Operating System: Android | iOS

dmitryn:
fixes #5745

### Summary:

Validate account recover result and show error to the user in case of unsuccessful recovery (password is not valid).

status: ready

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [0.9.28 - Unreleased]
 ### Added
 ### Fixed
+Fix app freeze on recover with wrong password
 ### Changed
 
 ## [0.9.27]
--- a/src/status_im/accounts/recover/core.cljs
+++ b/src/status_im/accounts/recover/core.cljs
@@ -58,18 +58,20 @@
   (let [password (get-in db [:accounts/recover :password])]
     {:db (assoc-in db [:accounts/recover :password-error] (check-password-errors password))}))
 
-(defn on-account-recovered [result password cofx]
-  (let [db         (:db cofx)
-        data       (types/json->clj result)
-        pubkey     (:pubkey data)
-        account    {:pubkey     pubkey
-                    :address    (:address data)
-                    :photo-path (identicon/identicon pubkey)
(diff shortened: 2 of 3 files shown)


yenda (approved):

rasom (approved):

src/status_im/translations/en.cljs
```diff
@@ -401,6 +401,7 @@
    :twelve-words-in-correct-order        "12 words in correct order"
    :enter-12-words                       "Enter the 12 words of your recovery phrase, separated by single spaces"
    :recover-password-too-short           "Password is too short"
+   :recover-password-invalid             "Password is not valid"
```
rasom: please change it in `translations/en.json`

mandrigin (approved):

src/status_im/translations/en.cljs
dmitryn: @rasom thanks, migrated to json.

lukaszfryc:
@dmitryn 2 things I'd change:

1. Make sure the tooltip is not overlapped, see:



2. Change the copy to "This account already exists but passwords do not match"

flexsurfer (approved):

dmitryn:
@lukaszfryc 
1. Was fixed by https://github.com/status-im/status-react/commit/8fa0522bada0e2923bc5883553423e206a30034b (merged it here and to develop)
2. Changed.

Please take a look.

lukaszfryc:
@dmitryn please update the changelog before merging.

PR #5801, opened 2018-09-10
[#5583]: Stronger security checks in browser; Metamask phishing detector

### LINKED ISSUE #5583: Stronger security checks in browser
### User Story

As a user, I want my browser to check pages for suspected threats so that I'm protected against malicious behavior on web3.

### Description

*Type*: Feature

*Summary*: The goal is to do more than simply verify whether an SSL cert is valid. We'll use Google Safe Browsing APIs to check for content that matches unsafe webpages. 

Google's black list includes phishing sites and sites that host malware/unwanted software.

In the future, we may consider employing MetaMask's [phishing detector](https://github.com/MetaMask/eth-phishing-detect), as well.

#### Expected behavior
When I visit an unsafe website (e.g. ethpaperwallet.net) in Status, I'm blocked from entering with a warning message.

_Google Safe Browsing warning_
image

#### Actual behavior
I can visit certain unsafe sites in Status.

_ethpaperwallet.net_

#### Additional info
From the wall of shame:

> Browser only validates whether SSL certificates are valid or not. But phishing sites can have valid certificates. By comparison, Google Safe Browsing checks suspicious terms from a site against Google servers. 

> We're in a transitionary phase, meaning we support Web 2.0 sites. While dependant on the implicit trust model of a server, we don't follow best practices in this environment. Checking SSL certificates gives us stronger guarantee's that the server we are talking to is the intended one.
> 
> As an extension to this, we could look at DNS at the same time, perhaps via ENS' DNS Layer or DNSCrypt could be another issue, if we don't trust OS's DNS lookups.
> 
> This was an issue with myetherwallet had a DNS poisoning attack made against it (beyond the usual spoofing) 
> 
> Another extension to that is to look at a layer 2 safebrowsing system, ENS is/was looking to build this system out.

alwx:
Fixes #5583 

However, the design for it is not final (basically, there is no design for it yet, so right now it looks almost like Chrome phishing alert). It now uses MetaMask's phishing detector.

An output from @rachelhamlin required :)

image


mobile_files/package.json
```diff
@@ -21,6 +21,7 @@
     "create-react-class": "15.6.2",
     "dns.js": "1.0.1",
     "emojilib": "2.2.9",
+    "eth-phishing-detect": "^1.1.13",
```
jeluard: Why `^` ? Looks different from other deps.

src/status_im/ui/screens/browser/deceptive_site/styles.cljs
```diff
@@ -0,0 +1,38 @@
+(ns status-im.ui.screens.browser.deceptive-site.styles
+  (:require-macros [status-im.utils.styles :refer [defstyle defnstyle]])
+  (:require [status-im.ui.components.colors :as colors]))
+
+(def container
+  {:justify-content  :center
+   :flex             1
+   :background-color "#B14A39"})
```
jeluard: Can we use color ns?

src/status_im/ui/screens/browser/deceptive_site/views.cljs
```diff
@@ -0,0 +1,50 @@
...
+            [re-frame.core :as re-frame]
+            [status-im.ui.components.colors :as colors]
+            [status-im.ui.components.react :as react]
+            [status-im.ui.components.button.view :as button]
+            [status-im.ui.screens.browser.deceptive-site.styles :as styles]
+            [status-im.i18n :as i18n]))
+
+(defn view [{:keys [can-go-back? site]}]
+  (let [more-details-open? (reagent/atom false)]
+    (fn []
```
jeluard: Do we need this fn?

jeluard: Or maybe just use defview?

src/status_im/ui/screens/browser/deceptive_site/views.cljs
```diff
@@ -0,0 +1,50 @@
...
+                         :style               styles/button-container
+                         :accessibility-label :back-to-safety-button}
+          (i18n/label :t/browsing-back-to-safety)
+          nil]
+         [button/button {:on-press            (fn []
+                                                (swap! more-details-open? not))
+                         :text-style          styles/button-text
+                         :accessibility-label :more-details-button}
+          (if @more-details-open?
+            (i18n/label :t/browsing-less-details)
```
jeluard: if could be around key only

jeluard (dismissed):
Can you share details about how `eth-phishing-detect` works exactly? Does it rely on 3rd party HTTP API?

alwx:
@jeluard no, it's basically just a list of phishing resources: https://github.com/MetaMask/eth-phishing-detect/blob/master/src/config.json

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/MetaMask/eth-phishing-detect">MetaMask/eth-phishing-detect</a></strong></div><div>Utility for detecting phishing domains targeting Ethereum users - MetaMask/eth-phishing-detect</div></blockquote>

jeluard:
@alwx Still WIP ?

mobile_files/package.json
```diff
@@ -20,6 +20,7 @@
     "create-react-class": "15.6.2",
     "dns.js": "1.0.1",
     "emojilib": "2.2.9",
+    "eth-phishing-detect": "~1.1.13",
```
jeluard: Why the version with ~ ?

alwx:
@jeluard the technical part of this story is finished; but I have some answers and still didn't get the answer :( See here: https://status-im.slack.com/archives/C9P233D54/p1536655569000100

<blockquote><div><strong><a href="https://status-im.slack.com/archives/C9P233D54/p1536655569000100">Slack</a></strong></div></blockquote>

mobile_files/package.json
alwx: They update eth-phishing-detect database quite frequently, and it's better to have this updates in time. I use `~` here to have only patch versions fetched and used in Status.

alwx:
@jeluard new design added. Can be re-reviewed

screen shot 2018-09-17 at 20 12 32

mobile_files/package.json
jeluard: We want to have reproducible builds so we must fix the version. It's fine if we have to update this frequently.

flexsurfer (approved):

src/status_im/models/browser.cljs
```diff
@@ -31,8 +32,14 @@
       (assoc browser :dapp? true :name (:name dapp))
       (assoc browser :dapp? false :name (i18n/label :t/browser)))))
 
+(defn check-if-phishing-url [{:keys [history history-index] :as browser}]
+  (let [history-host (http/url-host (try (nth history history-index) (catch js/Error _)))]
+    (assoc browser :unsafe? (dependencies/phishing-detect history-host))))
```
yenda, line 37: is it a pure function or a side-effect ?

alwx: it updates browser, so it's not pure.
and this function is used here the same way as the one above (`check-if-dapp-in-list`)

jeluard (approved):

src/status_im/models/browser.cljs
yenda: Why merge if it's not pure, you put that in an event, events are suppose to be pure...

yenda:
It was not even tested

PR #5824, opened 2018-09-13
Feature/emojis signing

yenda:
### Summary:

Replace words in signing phrases by emojis according to designs:
https://www.figma.com/file/Xg0hHWg7Ngqrly2syl4MAM/Transaction-phrase-UXR?node-id=187%3A36

Fix bug where user isn't redirected to wallet onboarding when trying to sign a message when his wallet is not setup, asking him for signing phrase he has never seen before.

Thanks @flexsurfer for the bug fixes on iOS

### Review notes
I voluntarily didn't moved some of the styling from `views` to `styles` namespace because it is totally impractical when working with flexboxes to have to switch inbetween namespaces to understand what is going on.
I also slightly modified some components but I think we should really be more careful with these abstractions, some are only accepting a few standard css keys and I wasted quite some time before figuring out that the `tooltip` component was dropping all the styling keys it was passed.

### Testing notes
- need to test if onboarding is necessary again when updating status with an existing account
- 3 flows to check, sending transaction from wallet, sending transaction from somewhere else (dapp and chat), signing a message (from status test dapp for instance). If wallet isn't set up, user should be redirected to wallet setup first, then to his signing screen.

status: ready

screenshot_1537022776
screenshot_1537022786
screenshot_1537022818
screenshot_1537022823
screenshot_1537022826
screenshot_1537023135


chadyj:
Hey Eric, I see this PR is a WIP so apologies if I'm chiming in too early.
There are some major signing and emoji UX issues surfaced in UXR, and ongoing work by the Core Wallet team on a wallet redesign. Did you get a chance to collaborate  @goranjovic @hesterbruikman @denis-sharypin @PhilipandWu on this? They have some great insight here.

From a product point of view this is an important feature for security and UX and I'm not sure it makes sense to rush into it now before it was validated with UXR. We also have more pressing needs that could use some help with.

hesterbruikman:
cc @lukaszfryc @asemiankevich. Just in case although I believe you are aware. In UX testing we haven't seen the emoji's resolve the signing phrase issue we've had when using the 3-word solution. Design is still iterating on a better solution.

oskarth:
Agree, while nice this seems like duplicate effort. It'd be useful to check with existing work going on in #core-wallet and UXR validation.

denis-sharypin:
The last study that Philip did showed us that people prefer emojis over 3 words. But they still have a small confusion with the description of it. We will make another attempt to change copy but from the app logic perspective, this issue is ready to be implemented.

hesterbruikman:
It wouldn't hurt to implement, as at least the phrase is not mistaken for seedphrase. Meaning it will not keep people from looking for and be open to notifications to back up their seedphrase. 

However, as far as I understand we don't know if the emoji solution is more effective as a phishing guard or effective at all as a phishing guard. People might prefer it more (emojis are fun...), but they still perceive it as 2-factor auth and from the results I heard describe, it's unclear whether people will refrain from a transaction if another emoji combination is shown.

If we implement this, I just want to make sure it doesn't drop of the radar and we assume we mitigate risk of phishing. False security.

denis-sharypin:
Hey @yenda  I just wanted to play with your implementation but it doesn't work on my device. I use iPhone6 
img_7298

denis-sharypin:
@yenda  Still can't see emojis and password field with the button
img_7300

src/status_im/ui/screens/wallet/onboarding/setup/styles.cljs
```diff
@@ -1,38 +1,72 @@
...
-  {:width  151
-   :height 77})
-
 (def signing-phrase
   {:background-color colors/white
    :border-radius    8
-   :margin-left      16
-   :margin-right     16
+   :height          68
+   :margin-horizontal 28
```
dmitryn: this line and many lines below seems not formatted

src/status_im/ui/screens/wallet/onboarding/setup/views.cljs
```diff
@@ -1,56 +1,59 @@
...
-           ^{:key (str "signing-word-" word)}
-           [signing-word word])]
-        [react/text {:style styles/description}
-         (i18n/label :t/wallet-set-up-signing-phrase)]
+        (i18n/label :t/wallet-setup-title)]
+       [react/view {:style {:flex 1}}
+        [react/view {:style {:flex 1
+                             :flex-direction :column
+                             :align-items :center
+                             :justify-content :center}}
```
dmitryn, line 40: move to styles?

src/status_im/ui/screens/wallet/send/styles.cljs
```diff
@@ -30,26 +16,46 @@
...
 
 (def signing-phrase-container
-  {:border-radius    8
+  {:flex 1
+   :border-radius    200
+   :border-width     1
    :height           36
+   :margin-vertical 24
+   :padding-vertical 25
+   :padding-horizontal 25
```
dmitryn: formatting here and below

src/status_im/ui/screens/wallet/onboarding/setup/views.cljs
yenda: as mentionned in the review notes:
```
I voluntarily didn't moved some of the styling from views to styles namespace because it is totally impractical when working with flexboxes to have to switch inbetween namespaces to understand what is going on.
```

dmitryn (approved):

denis-sharypin:
@yenda  Overall looks great. There are some style fixes:

1. Delete an extra line break in the info section
img_7295

2. Not sure if need to make `Got it` in red.  Let's keep it in blue color
img_7296

3. Please make a background of this popup a longer. I want to a avoid this gap between the keyboard and popup during transition. 
screenshot 2018-09-17 18 10 00


Also, @PhilipandWu  what would be our next steps here? Changing copy? We can test this build with the users if you think it's needed.

janherich (approved):

PhilipandWu:
Sorry I’m late to the "party" everyone, GitHub was being finicky while I was abroad.

Echoing earlier comments I don’t think it would hurt to implement either, but I don’t think we should rush as it is important security feature to “get right”. As mentioned earlier from our most recent tests last month with lay users + more tech ones at ethberlin many still perceive it as a 2FA or some variation. People only get that's its to prevent phishing if their bank uses something similar atm. Results have been mixed on whether users are confident that they should not proceed if different emojis are shown as they think they might be sending it to the wrong person or in the wrong wallet.

Latest copy might help resolve this issue in part. We’ve made it much more explicit to the user that’s we’re protecting them, but as you can see from the Figma comments it’s a WIP:
* [see 3rd comment on super-safe transaction design (3rd design from the left)](https://www.figma.com/file/Xg0hHWg7Ngqrly2syl4MAM/Transaction-phrase-UXR?node-id=0%3A1
)

For reference here’s the full prelim TX study findings from last month and GH issue linked:

* [preliminary TX study details](https://docs.google.com/document/d/1vyuyUqyn0g2sTQSN1SW263qFqH9QCxlMK8kz_Fi0w0Y/edit?ouid=105740245751601859065&usp=docs_home&ths=true)
* [GH issue](https://github.com/status-im/design/issues/27
)

yenda:
@PhilipandWu can't we improve the copy in a future PR once design is settle ? This PR is not just about design, switching to emojis require a database change and the password drawer is quite different from the current popup. If it is already integrated subsequent changes will be much faster

lukaszfryc:
I tested the app and from the functional point of view all works good @yenda. 

But, from UX/UI perspective I have few things that we may consider:

1) @hesterbruikman @PhilipandWu @denis-sharypin Have we thought about suggesting users to take a screenshot of the emojis to save them? I bet 90% of people will not remember it after few hours. Saving the words we had previously was much more easier than saving / remembering emojis now.

2) @denis-sharypin Gray text ("You should see.." and "If you see..") on iOS is hardly visible. If we want encourage people to read it we should make it more visible :) I’d suggest to make it white.



3) @denis-sharypin @yenda is there a reason that emojis are grayed out (check screenshot below from Android)? 



4) @denis-sharypin is that blue tooltip ("Only send transactions..") supposed to disappear after few successful transactions? If not, have you consider showing this info as a standard text instead of a tooltip that fades in (current solution). For me, tooltips are more about temporary/context info.

denis-sharypin:
Thanks @lukaszfryc 

1. We can't suggest to do it since it is not secure to keep the screenshot on your phone or cloud. 
2. Agree, good catch. @yenda can you increase the opacity of this text by 20% for all transparent text on blue. 
3. It is should be a bug
4. We need to bring an extra user attention to emojis, that's why we are showing this information with the 1s delay. Standard text on the popup didn't work out: users simply didn't notice it

yenda:
3. What is grayed out I don't see it ?

flexsurfer:
Also @denis-sharypin for most dapps it will be "Send 0 ETH" , can we show something like "Sign contract transaction"?

yenda:
screenshot_1537269582

yenda:
I rebased and fixed the suggestions, will see how the iOS version works with the new react-navigation, android is fine.

lukaszfryc:
> We can't suggest to do it since it is not secure to keep the screenshot on your phone or cloud.

I agree it's not secure to store it in form of a screenshot. Anyway, I still think people will have hard time remembering it. @hesterbruikman would be great to measure it. What's the point of showing emojis when people do not remember them and then they are stressed when they need to decide on emojis while signing transaction.

PhilipandWu:
@yenda  of course, that makes total sense.  

@lukaszfryc  In regards to the screenshot as denis says it's not ideal and we're not sure if we want to encourage this behaviour more while it is the lowest friction option. As a quick anecdote, it's effectively what half of users of the tested users I run done anyways. 

I would say just recognition is more important than exact recall. We could do a follow up test with users a week later to see what their recognition memory is like.

yenda:
The recognition problem is why is suggested to ask the user to choose his emojis within 3 different combinations  when the user logs in with time-repetition like Anki does based on success or failure to recognize them

lukaszfryc:
OK. Let me do the final testing (with react navigation).

lukaszfryc:
@yenda 

I noticed few issues in this PR while testing on Android 8 (Galaxy S8). Note that they do not happen on in latest nightly that includes react navigation PR.

1) Error when going back from emojis modal in send transaction.

Steps:
1. Switch to ropsten and create 1-1 chat 
2. Send ETH in the chat ("/send")
3. Click send transaction button in send transaction screen
4. Once emojis popover appeared, press device back button to dismiss it
5. Press the back button again to close send transaction screen and return to the chat. If nothing happens, click "X" button.
6. The error appears:



2) Similar issue as 1) happens in sign message screen in the test dapp. When you navigate back, the sign message button does nothing when clicked

Steps:
1. Install app and switch to ropsten
2. Open test dapp and go Transactions -> Sign message
3. In sign message, press device back button to go back to the dapp
4. Try to click sign message button again. Nothing happens

hesterbruikman:
Sorry to have been absent here!

Few thoughts in addition to the above. 

- I have one reason for which I would want to block releasing what we have now. Which is that as a phising guard, it should protect the user from 2 things:
 A) prevent from signing a transaction to an other than intended recipient/wallet address
 B) enter your password
The current approach does not protect from B and I'm concerned we'll be called out for this at some point. cc @chadyj Even though it's no different from what we have now. I'm also not much supported in A as I can't verify the recipient at the most crucial step of signing the transaction.
- All of this is no different from our current solution so in a sense any change is good, but I would definitely not want to close this issue.
- I added some copy comments in Figma, most can wait until after this release as I understand @yenda point that in this round we want to get the implementation right and can dot the i's after. Most relate to inconsistency in using `Sign`, `Send`, `Confirm` transaction.
- **I am in much support of people selecting their own emojis.** This doesn't have to be in full control. A middle ground; as people might end up selecting the equivalent of 0000, is to have people feel like they're in control by simply adding a `Generate` your personal emoji combo to secure your transactions. This is not just fun; an active interaction with the emojis will aid recognition. If it's up to me  we even add a trial transaction to it to demonstrate desired behavior when different emoji's come up.

hesterbruikman:
Some additional insights from UX testing to make sure they are captured with this issue. https://docs.google.com/presentation/d/10qb5jRXwRT9lYkxrn8Ui5Bov_JRuf1WBaPqSXd0ZAtY/edit#slide=id.g3f411e60e9_0_197

<blockquote><div>Google Docs</div><div><strong><a href="https://docs.google.com/presentation/d/10qb5jRXwRT9lYkxrn8Ui5Bov_JRuf1WBaPqSXd0ZAtY/edit?usp=embed_facebook">Report RRR8 Chat & Command</a></strong></div><div>Hester Status 17 / 09 / 2018 UXR findings Chat & Command RRR8 Sept 6th 2018</div></blockquote>

yenda:
The initial purpose of this PR is just too replace words by emojis all the rest can be done later. Regarding letting the user choose his own emoji or regenerate it could be done but it will seriously lower the effectiveness of the emoji guard

hesterbruikman:
> The initial purpose of this PR is just too replace words by emojis all the rest can be done later. 

Makes sense

> Regarding letting the user choose his own emoji or regenerate it could be done but it will seriously lower the effectiveness of the emoji guard

To clarify; with `Generate` I mean a button that triggers a slot machine that still provides you with system generated emojis. Agreed that user-generated would be less secure and unnecessary to improve recognition

PhilipandWu:
Quick follow up question @yenda , sorry if i missed this earlier. How will current status users who have already setup their wallet be informed that their words have switched to emojis? Would they be prompted in the next transaction?

ghost:
### Pull Request Checklist

yenda:
@PhilipandWu Wallet onboarding is reset so they will see it the first time they use wallet after the update

corpetty:
I'll word vomit some thoughts around this conversation w.r.t. security:

1. I strongly like the idea of moving away from words as a phishing check due to the confusion behind them being at all related to the recovery phrase.  This should never enter the user's mind, and ideas around what the recover phrase is / how to handle it / and where it gets used should be isolated from any other ideas in the app.
2. Does there exist any study on what it would take a malicious actor to _actually change_ this phishing guard?  This would inform the importance of this feature and its respective changes,  i.e. is this an actual security feature or more of a user trust in status feature.  Some banks do this, and I'm sure that's where the initial implementation started, but that's because it's being served from a central location and you're checking that you're connected to a proper server.  Does that same reasoning exist here within Status as well?
3. @yenda can you elaborate more about the back end changes and how implementing this now makes it easier on iteration in the future?  That wasn't immediately clear to me.

yenda:
@corpetty 3. First it requires to change the word list with the emoji list and update the app-db to reset the  signing-phrase so that user upgrading get emojis and have their wallet set up reset. Also the new pop-up drawer is quite a big change compared to the previous popup in the middle of the screen, as the number of commits and still ongoing work to fix it after a few changes to wallet and react-navigation shows.

denis-sharypin (approved):

hesterbruikman (changes requested):
https://discuss.status.im/t/what-are-our-criteria-for-releasing-security-critical-features/451

1) The copy used does not indicate what one should do when not recognizing the emojis. 
REVIEW PREVIOUS COPY SUGGESTIONS @obi2020 

2) None of the participants described intended behavior with regard to the phishing guard. 
VALIDATE WITH END-USERS IN EXPERIMENTAL SETUP @PhilipandWu 

3) No evidence about the effectiveness of the signingphrase 
VALIDATE WITH END-USERS IN EXPERIMENTAL SETUP

4) The pattern does not prevent phishing for password.
REVIEW RISK @goranjovic AND DESIGN @denis-sharypin 

5) [Sign contract transaction] without introduction might create confusion for lay users. Nowhere in the app is there any mention of ‘Contracts’ yet.
REVERT 'SIGN CONTRACT TRANSACTION' TO TESTED COPY 

6) The critical moment does not show the Wallet address and amount for verification.
REVIEW DESIGN TO SHOW RECIPIENT ADDRESS

oskarth:
@yenda In the future, could you label PR with security so it's clear extra care should be taken before merging it?

lukaszfryc:
@yenda 

I briefly tested latest build on Android (Galaxy S8, Android 8) and all worked fine for me besides:

1) "Sign" and "Send" buttons should grow (to some max width) to accomodate space for the text. 




2) As previously mentioned, the emojis in onboarding screen are grayed out for some reason. They look fine later on. Have a look:





Let me know if I should continue with my testing (iOS) or you want to address [Hester's comment first](https://github.com/status-im/status-react/pull/5824#pullrequestreview-158053850).

corpetty:
w.r.t. security:
- This PR should be pushed forward.  If we have a signing phrase for phishing purposes, it should be differentiated from using something that is similar _in any way_ to the recovery phrase.  
- The argument of whether it is useful at all is independent of its implementation details.  @oskarth has informed me this is more of a guarantee that DApps are not attempting to render a message signing screen that looks like Status but isn't, which should be exacerbated by extensions.  This point answers my previous question on whether this is needed, it is. maybe @jeluard can confirm this for me.
- @hesterbruikman brought up great points, specifically 1,4,5, and 6.

PhilipandWu:
"The argument of whether it is useful at all is independent of its implementation details. @oskarth has informed me this is more of a guarantee that DApps are not attempting to render a message signing screen that looks like Status but isn't, which should be exacerbated by extensions"

Good to know!

For 1, I've included a direct link to the reworked copy @obi2020  recently did. Also, see comments 2 and 3 for more context. https://www.figma.com/file/Xg0hHWg7Ngqrly2syl4MAM/Transaction-phrase-UXR?node-id=0%3A1

<blockquote><div> Figma</div><div><strong><a href="https://www.figma.com/file/Xg0hHWg7Ngqrly2syl4MAM/Transaction-phrase-UXR">Figma</a></strong></div><div>Created with Figma</div></blockquote>

There still might be some pending changes given the discussion, but wanted to add this for frame of reference.

jeluard:
Something to keep in mind while we are still in the process of finding a proper solution: `extensions` will allow 3rd parties to inject arbitrary UI matching our native interface. 
How big and complex those UIs could be will depend on what `hooks` will be offered later on.

lukaszfryc:
@yenda let me know if it makes sense for me to retest this PR.

oskarth:
More than two months old PR. Closing as inactive. Feel free to reopen if this is in error / when it is ready.

yenda:
still waiting for UXR input on that one @hesterbruikman

denis-sharypin:
cc @hesterbruikman

hesterbruikman:
Realistically, this will not happen until Januari. However, a study is still planned. Blocking for the moment is time and a setup that can mimic an actual case of phising as close as possible. Here's the draft test plan: https://drive.google.com/open?id=10egEEdXSRy2_7o3JUdgIjXueMURm5sT_XQ3WWt4RJjk

<blockquote><div><strong><a href="https://drive.google.com/open?id=10egEEdXSRy2_7o3JUdgIjXueMURm5sT_XQ3WWt4RJjk">Meet Google Drive – One place for all your files</a></strong></div><div>Google Drive is a free way to keep your files backed up and easy to reach from any phone, tablet, or computer. Start with 15GB of Google storage – free.</div></blockquote>

PR #5826, opened 2018-09-13
Fix `make setup` on macOS.

mandrigin:
Before, it always downloaded linux NDK no matter which platform you are
using. This way, `make setup` was actually able to break Android build
on macOS.

fixes my devenv

status: ready

### DIFF
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -325,7 +325,13 @@ function install_android_ndk() {
     local _ndkParentDir=~/Android/Sdk
     mkdir -p $_ndkParentDir
     cecho "@cyan[[Downloading Android NDK.]]"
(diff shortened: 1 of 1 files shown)


jeluard (approved):

yenda (approved):

scripts/lib/setup/installers.sh
```diff
@@ -325,7 +325,13 @@ function install_android_ndk() {
     local _ndkParentDir=~/Android/Sdk
     mkdir -p $_ndkParentDir
     cecho "@cyan[[Downloading Android NDK.]]"
-    wget --output-document=android-ndk.zip https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip && \
+
+    PLATFORM="linux"
+    if [ "$(uname)" == "Darwin" ]; then # we run osx
```
yglukhov, line 330: This will not work. Comparison is single `=`

EDIT: Correction: works in sh and bash, not in zsh. `man test` doesn't mention this syntax as valid.

PR #5833, opened 2018-09-13
translations: Add PT translation for Fastlane

pedropombeiro:
This PR adds European Portuguese Google Play translations.

status: ready


chadyj (approved):
Nice one! I assume the language code is correct so we should look out for it on the next deploy.  (I also want to make a Fastlane action that just updates metadata)

pedropombeiro:
Yeah, we could use `pt-PT`, but normally European Portuguese acts as the baseline.

lukaszfryc (approved):

PR #5888, opened 2018-09-18
drop use of artifactory maven repo

jakubgs:
We really need to get rid of Artifactory. I wonder fi this would be enough to drop dependency on it...


vitvly (approved):

pedropombeiro:
@jakubgs I'm actually working on a related issue (remove the need to build status-go every time on Desktop builds), and I'm starting by generating the required static library from the Go side: https://github.com/status-im/status-go/issues/1190

ci/mobile.groovy
```diff
@@ -51,6 +51,12 @@ def prep(type = 'nightly') {
   sh 'cp patches/js_realm.hpp node_modules/realm/src/js_realm.hpp'
   /* install ruby dependencies */
   sh 'bundle install --quiet'
+  /* download status-go for mvn */
+  def pomDir = 'modules/react-native-status/ios/RCTStatus'
+  def repo = "${pwd}/${pomDir}/repo/status-im/status-go-ios/1.15.0"
```
pedropombeiro: `1.15.0`: you mean `0.15.0` right?

also, ideally there would be a single line needing to be updated with this by the `scripts/update-status-go`.

Makefile
```diff
@@ -3,6 +3,8 @@
 help: ##@other Show this help
 	@perl -e '$(HELP_FUN)' $(MAKEFILE_LIST)
 
+GO_VERSION = 0.15.0
```
pedropombeiro: Maybe we can read this from the platform-specific file (.pom file for iOS, `modules/react-native-status/android/build.gradle` for Android)?

pedropombeiro (changes requested):
Looking good! Need to account for remaining platforms, but once I have static libraries being built for Desktop, it's going to be golden and we'll have shorter CI builds.

Makefile
jakubgs: Yeah, I was thinking about that. Kinda flimsy, but better than hardcoding it.

ci/mobile.groovy
jakubgs: That's true, consider this a work in progress. I still have a long way to go to make this work with `pom.xml`. It's so painful.

modules/react-native-status/android/build.gradle
```diff
@@ -16,7 +16,9 @@ dependencies {
     implementation 'com.instabug.library:instabug:3+'
     implementation 'status-im:function:0.0.1'
```
pedropombeiro, line 22: @jakubgs Where would we get this `status-im:function` module from? It is currently in artifactory.

jakubgs:
Turns out the `pom.xml` file is useless beyond downloading the iOS status-go file and extracting it from a Maven respository. Since we are dropping Artifactry we can also drop our dependency on Maven, so I removed `pom.xml` and I just do the same in the `Makefile`.

modules/react-native-status/android/build.gradle
jakubgs: If you look closer you can see that I've added `android/app/libs/function-0.0.1.jar` to the repo itself. It's a tiny file.

Makefile
jakubgs: I've dropped `xml.pom`, so now it's only in two places.

mandrigin (approved):
lgtm

pedropombeiro (approved):

cammellos (approved):

janherich (approved):

rasom (approved):

lukaszfryc:
@jakubgs I cannot run macOS build (I tried fresh install and upgrade scenarios). Could you check the screenshot I'm attaching? Other than that the PR looks good.

image

jakubgs:
Thanks @lukaszfryc but that makes no sense to me. This PR changes nothing about desktop builds. It only affects mobile builds. Can you link to me which build artifact you used?

lukaszfryc:
I used https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-181003-191815-a04d8e-pr.dmg from https://github.com/status-im/status-react/pull/5888#issuecomment-426771813

<blockquote></blockquote>

lukaszfryc:
@jakubgs now all platforms work fine!

PR #5913, opened 2018-09-20
Enable webview debugging in release builds

flexsurfer:


jeluard:
As a next step we could completely remove the flag?

jeluard (approved):

dmitryn (approved):

flexsurfer:
Yeah, we definitely can remove it later, if we'll decide to enable debugging in all builds

yenda (changes requested):
What are the security implications of doing this ? Wouldn't it make it possible for an other app to listen to the webview the way chrome does ?

PR #5915, opened 2018-09-20
Add pfs flag, installation id and nobackup data directory

cammellos:
This PR adds a set of option so that `status-go` and `status-react` are compatible when https://github.com/status-im/status-go/tree/features/x3dh gets merged, so we don't fall in the situation where `HEAD` status-go is not compatible with `HEAD` status-react.

### Summary:
Add a flag for PFS
Adds a bunch of options used by `status-go` so that when we merge https://github.com/status-im/status-go/tree/features/x3dh the two projects are compatible
Add `installation-id` used to identify an installation, used by pfs

### Testing notes (optional):
Everything should be working as before, it just needs to verify that upgrades cleanly

status: wip

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-go">status-im/status-go</a></strong></div><div>The Status module that consumes go-ethereum. Contribute to status-im/status-go development by creating an account on GitHub.</div></blockquote>


yenda (approved):

dmitryn (approved):

pedropombeiro:
@cammellos Why not add the `PFS_ENCRYPTION_ENABLED` flag to the .env files?

src/status_im/accounts/create/core.cljs
```diff
@@ -51,6 +52,7 @@
 (defn on-account-created [{:keys [pubkey address mnemonic]} password seed-backed-up {:keys [signing-phrase status db] :as cofx}]
   (let [normalized-address (utils.hex/normalize-hex address)
         account            {:public-key      pubkey
+                            :installation-id (random/guid)
```
janherich: Out of interest, why not just use `:random-id` injected by standard coeffect ? Now the event it not pure so harder to test.

janherich (changes requested):

src/status_im/accounts/create/core.cljs
cammellos: The format is not desirable as gives out some metadata (it's prefixed with the timestamp). 
Also the way we use that `cofx` is dangerous imo as fns that are composed using it might re-use the same random-id, which is worse than not having functional purity.
I could create a cofx `random-guid` or similar and use that one, but since this patter has been already abused, I personally feel at this point it would encourage misuse, but don't mind doing  so after all.

cammellos:
@PombeirP added!

src/status_im/accounts/create/core.cljs
yenda: @cammellos Very good observation it is indeed problematic!

the solution would be to fix the current cofx so that it returns the random-id function instead of an actual random-id and call it in the functions that receive it instead of using `random-id` directly. You could do that in a separate second commit in this PR that we won't squash to keep it separated.

yenda: or in a simpler step since you don't want the particular format of random-id in that case make indeed a random-guid cofx that returns a function and call it when needed. In that case please create a issue with security and priority label so that we do a separate quick refactoring of the random-id cofx

janherich: Yes, very good points, I would just create `:random-guuid-generator` cofx and used that in the model function.
We can then move `random-id` to `random-id-generator` pattern as well, previously there was quite vocal opposition against passing functions in cofx, which is totally unfounded I think and it's very helpful in cases like this.

cammellos: @yenda I can add a random guid fn to cofx in the scope of this PR (unless is ready to be merged before I come back from lunch break , but given jenkins state unlikely :) ) , otherwise I'll add it to the next, and I might make a separate for the `random-id` issue.

yenda: I would suggest to do the `random-guuid-generator` as mentionned by @janherich, we try to not add impure events so we don't have to fight for refactoring afterward

yenda: The `random-id-geneator` definitely doesn't belong to this PR

janherich (approved):

src/status_im/accounts/create/core.cljs
cammellos: @yenda using a `random-guuid-generator` function passed through cofx would still make the function not pure :) , it's just a neater way to group side-effects and possibly stubbing them in testing, a la dependency injection, but it would still happen at runtime when the function is called.

if we really want to make `on-account-created` pure we need to pass the guuid (not a function generating it) and perform the side-effect outside the function, but needs to be clear from the code that it is only to be used in that function.

yenda: Whether the function is pure or not is debatable, I would argue that if you pass it the random-id function it is still pure, because it's the random-id function that isn't pure and depends on external state.

janherich: @cammellos I disagree, it would still be a pure function, the same way how `map` HOF is a pure function. 
In runtime every model function taking coeffects is impure (because when event-handler is called, it's composed of interceptors as well and one of those calls the `:now` coeffect...), but it doesn't matter, what matters is that the function definition itself is pure and can be used as such in isolated cases like unit-testing.

pedropombeiro (approved):

cammellos:
that s probabily better for slack,  but for example in haskell ( which
enforce purity) this will not work, and you can guarantee functional purity
of map ( or any function taking a fn and calling it as args), as it will
prevent to pass a fn with side effects.

in any language that does not enforce purity of fns technically these kind
of fns cannot be considered pure.

practically it does not matter, and i agree that for our intent and purpose
the solution proposed is better :)

src/status_im/accounts/create/core.cljs
cammellos: that s probabily better for slack,  but for example in haskell ( which
enforce purity) this will not work, and you can guarantee functional purity
of map ( or any function taking a fn and calling it as args), as it will
prevent to pass a fn with side effects.

in any language that does not enforce purity of fns technically these kind
of fns cannot be considered pure.

practically it does not matter, and i agree that for our intent and purpose
the solution proposed is better :)

lukaszfryc:
@cammellos so for now, you only want to test `PFS_ENCRYPTION_ENABLED=0` case and leave `PFS_ENCRYPTION_ENABLED=1` testing when status-go changes will be ready (another PR), right?

cammellos:
correct, thanks!

lukaszfryc:
All works fine, I tested chats after upgrading from 0.9.27 to the PR build.

lukaszfryc:
Let me check desktop build

lukaszfryc:
I checked desktop and did not find any issues.

PR #5917, opened 2018-09-20
explicitly add react-native-safe-area-view as dependency

yenda:
status: ready


dmitryn (approved):

flexsurfer:
@yenda could you please provide some details about this change?

yenda:
@flexsurfer it is a dependency of react-navigation, when building for android I had an error in the metro bundler that this dependency was missing. So I just explicitly add it to the package.json

janherich (approved):

flexsurfer (approved):

PR #5946, opened 2018-09-22
[#5934] persist text in message input field after navigating

### LINKED ISSUE #5656: Text in input field is not erased when switching to other chat
### Description

*Type*: Bug
*Summary*: when message field is not empty text should be preserved for selected chat only, so if you are switching to other chat, field should become empty and if you come back again - text should be preserved.
Currently, if text is added in input field, then after switching to another chat - text was preserved. At attempt to send message all text is erased, message is not sent.

#### Expected behavior
the input field is empty

#### Actual behavior
text for previous chat is preserved; at attempt to send message text is erased.

### Reproduction

- Open Status Desktop
- Log in
- Join to any 2 chats (i.e. #status and #status-desktop)
- Start typing text in  #status (i.e., `1234`)
- Switch to other  #status-desktop

### Additional Information
* Status version: nightly 27/08/2018
* Operating System: Mac OSx
* Video: http://take.ms/PDj1U

<blockquote><div>Monosnap screenshot tool</div><div><strong><a href="https://monosnap.com/file/ugVqpEq2OCBoSmLPm4x0ArIzJC6RZ4">File "screencast 2018-08-27 18-22-22.mp4"</a></strong></div><div>Monosnap — the best tool to take & share your screenshots.</div></blockquote>

### LINKED ISSUE #5934: Text is not persisted in message input field after navigating to user profile and back
### Description

*Type*: Bug
*Summary*: if you type part of message and navigate to user profile in public chat or to own profile, i.e., for copying public key, after coming back message will be erased.

#### Expected behavior
text is persisted

#### Actual behavior
text is erased

### Reproduction

- Open Status
- Open public chat 
- Type any message 
- Click on profile
- Go back to Home

### Additional Information
* Status version: nightly 21/09/18
* Operating System: Mac OS x
* Video: http://take.ms/0MB8y

<blockquote><div>Monosnap screenshot tool</div><div><strong><a href="https://monosnap.com/file/4dw3euMu7V01qHTmr2KqFQth4hv8tv">File "screencast 2018-09-21 10-56-16.mp4"</a></strong></div><div>Monosnap — the best tool to take & share your screenshots.</div></blockquote>

rasom:
fix #5934
fix #5656 
status: ready

### DIFF
--- a/src/status_im/ui/screens/desktop/main/chat/views.cljs
+++ b/src/status_im/ui/screens/desktop/main/chat/views.cljs
@@ -180,35 +180,40 @@
                     :current-public-key current-public-key)]))]]
        [connectivity/error-view]])))
 
-(views/defview chat-text-input []
+(views/defview chat-text-input [chat-id input-text]
   (views/letsubs [inp-ref (atom nil)]
-    (let [component              (reagent/current-component)
+    {:should-component-update
+     (fn [_ [_ old-chat-id] [_ new-chat-id]]
+       ;; update component only when switch to another chat
+       (not= old-chat-id new-chat-id))}
+    (let [component               (reagent/current-component)
           set-container-height-fn #(reagent/set-state component {:container-height %})
           {:keys [container-height empty?] :or {empty? true}} (reagent/state component)]
       [react/view {:style (styles/chat-box container-height)}
-       [react/text-input {:placeholder    (i18n/label :t/type-a-message)
-                          :auto-focus     true
-                          :multiline      true
-                          :blur-on-submit true
-                          :style          (styles/chat-text-input container-height)
-                          :font           :default
-                          :ref            #(reset! inp-ref %)
+       [react/text-input {:placeholder            (i18n/label :t/type-a-message)
+                          :auto-focus             true
+                          :multiline              true
+                          :blur-on-submit         true
+                          :style                  (styles/chat-text-input container-height)
+                          :font                   :default
+                          :ref                    #(reset! inp-ref %)
+                          :default-value          input-text
                           :on-content-size-change #(set-container-height-fn (.-height (.-contentSize (.-nativeEvent %))))
-                          :on-key-press   (fn [e]
-                                            (let [native-event (.-nativeEvent e)
-                                                  key (.-key native-event)
-                                                  modifiers (js->clj (.-modifiers native-event))
-                                                  should-send (and (= key "Enter") (not (contains? (set modifiers) "shift")))]
-                                              (when should-send
-                                                (.clear @inp-ref)
-                                                (.focus @inp-ref)
-                                                (re-frame/dispatch [:send-current-message]))))
-                          :on-change      (fn [e]
-                                            (let [native-event (.-nativeEvent e)
-                                                  text (.-text native-event)]
-                                              (reagent/set-state component {:empty? (= "" text)})
-                                              (re-frame/dispatch [:set-chat-input-text text])))}]
-       [react/touchable-highlight {:style styles/send-button
+                          :on-key-press           (fn [e]
+                                                    (let [native-event (.-nativeEvent e)
+                                                          key          (.-key native-event)
+                                                          modifiers    (js->clj (.-modifiers native-event))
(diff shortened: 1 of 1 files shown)


flexsurfer:
hm this issues says https://github.com/status-im/react-native-desktop/issues/332 `defaultValue` is missed, @rasom did you test on desktop and it works?

janherich:
I also think it would be much, much better to simply fix desktop `text-input` element to treat `defaultValue` properly instead of such hack.
Alternatively, we could try managed approach where `value` is directly mapped via subscription and we `.preventDefault` in on-change handler.
Last time I tried it, it wasn't performing good, but that was at the time where we had the perf regression because of the RN screwed `text-input` completely, maybe now it would be OK.

rasom:
@flexsurfer, i checked only on macos, yes it works.

@janherich, i don't see how it's better, that's replacing one hack with another. 

Let's figure if  `defaultValue` will be fixed anytime soon on `react-native-desktop` side, if not we can live with this hack until this will happen. Otherwise, i will close PR and wait for fix.

janherich:
@rasom That's certainly not a hack, it's a cleaner way more in line with React principles, in fact controlled components are default way to handle UI in React - https://reactjs.org/docs/forms.html#controlled-components.
The only reason why not to do it are potential performance implications (and we should check if it's still the case), otherwise it's simpler and easier to reason about as you can be sure that UI state always map 100% to what's in app state (react props).

<blockquote><div><strong><a href="https://reactjs.org/docs/forms.html">Forms – React</a></strong></div><div>A JavaScript library for building user interfaces</div></blockquote>

rasom:
@janherich yep, it's not a hack, it just doesn't work ¯\\_(ツ)_/¯

i tried this 
```clojure
(views/defview chat-text-input []
  (views/letsubs [inp-ref (atom nil)
                  input-text [:get-current-chat-input-text]]
    (let [component              (reagent/current-component)
          set-container-height-fn #(reagent/set-state component {:container-height %})
          {:keys [container-height empty?] :or {empty? true}} (reagent/state component)]
      [react/view {:style (styles/chat-box container-height)}
       [react/text-input {:placeholder            (i18n/label :t/type-a-message)
                          :auto-focus             true
                          :multiline              true
                          :blur-on-submit         true
                          :style                  (styles/chat-text-input container-height)
                          :font                   :default
                          :ref                    #(reset! inp-ref %)
                          :value                  input-text
                          :on-content-size-change #(set-container-height-fn (.-height (.-contentSize (.-nativeEvent %))))
                          :on-key-press           (fn [e]
                                                    (let [native-event (.-nativeEvent e)
                                                          key          (.-key native-event)
                                                          modifiers    (js->clj (.-modifiers native-event))
                                                          should-send  (and (= key "Enter") (not (contains? (set modifiers) "shift")))]
                                                      (when should-send
                                                        (.clear @inp-ref)
                                                        (.focus @inp-ref)
                                                        (re-frame/dispatch [:send-current-message])))
                                                    (.preventDefault e))
                          :on-change              (fn [e]
                                                    (let [native-event (.-nativeEvent e)
                                                          text         (.-text native-event)]
                                                      (reagent/set-state component {:empty? (= "" text)})
                                                      (re-frame/dispatch [:set-chat-input-text text]))
                                                    (.preventDefault e))}]
       [react/touchable-highlight {:style styles/send-button
                                   :on-press (fn []
                                               (.clear @inp-ref)
                                               (.focus @inp-ref)
                                               (re-frame/dispatch [:send-current-message]))}
        [react/view {:style (styles/send-icon empty?)}
         [icons/icon :icons/arrow-left {:style (styles/send-icon-arrow empty?)}]]]])))
```
is that what you proposed?

inp

flexsurfer (approved):

janherich (approved):

vkjr (approved):

vitvly (approved):

churik:
Tested that text is persisted for particular (1-1 or public) chat:
- when switching between chats
- when navigating between Profile - Home
- when navigating to New Chat and start new chat
- when used "View Profile", "Clear history", "Delete chats"

Actually, this PR fixed two issues, so changed the description.

PR #5999, opened 2018-09-26
[#5554]: UX updates on Recovery flows

### LINKED ISSUE #5554: UX updates on Recovery flows
Align recovery enter mnemonic with backup seedphrase flow

User Story
As a user, I want to have a right perception on what kind of password do I need to provide during the recovery flow that I'm not trying to recall my old password.

(Needed to ETHBerlin)

### Description
_Type_: Feature

_Summary_: Current recovery flow confuses users, makes them think they need to provide their old password which is not correct. I propose to add separate 2 steps for password creation and confirmation. Also, I renamed (Hester proposed this many times) action on the main screen from `I already have an account` to `Access account`.

#### Expected behavior
slice

alwx:
Updates recovery flow, fixes #5554.
Also simplifies and removes a lot of (weird) code in `status-im.accounts.access.core`.

Status: ready.


jeluard:
@alwx Looks like you have some conflicts 😅

cammellos:
@alwx could you please rebase with the latest develop?
Thanks!

yenda (commented):
can you use a local state for on-change ? We should generalize that practice an not dispatch event on every keystroke. If there is a need for a validation function, it should be passed in a subscription

alwx:
@yenda it's a code improvement that needs be applied to all other login-related screens, and that's why I think it's better to do it separately. I can do it, but I see no point in increasing the complexity of this specific PR.

alwx:
@jeluard not anymore! :)

src/status_im/accounts/access/core.cljs
```diff
@@ -0,0 +1,105 @@
...
+    (when-not processing?
+      (if (mnemonic/status-generated-phrase? passphrase)
+        (access-account cofx)
+        {:ui/show-confirmation
+         {:title               (i18n/label :recovery-typo-dialog-title)
+          :content             (i18n/label :recovery-typo-dialog-description)
+          :confirm-button-text (i18n/label :recovery-confirm-phrase)
+          :on-accept           #(re-frame/dispatch [:accounts.access.ui/access-account-confirmed])}}))))
+
+(fx/defn next-step [{:keys [db] :as cofx} step]
```
yenda, line 75: the logic of accounts.create is pretty much the same, did you try to factorize ?

alwx: Not yet, will be done later.
Basically, the answer is the same as the previous one — there is absolutely no need to do it here in this PR. There are a lot of things that can be improved on these screens in general — we need to make everything more generic, use parts of the code, but when you start diving into it, you understand that it's huge and it's not something that needs to be done for a story marked as "critical". 
However, I will start doing it after I finish with current changes.

jeluard (approved):

src/status_im/accounts/access/core.cljs
```diff
@@ -0,0 +1,105 @@
...
+                   :address    address
+                   :photo-path (identicon/identicon pubkey)
+                   :mnemonic   ""}]
+      (accounts.create/on-account-created cofx account password true))
+    {:db (assoc-in db [:accounts/access :error] (i18n/label :t/recover-password-invalid))}))
+
+(fx/defn on-account-accessed [cofx result password]
+  (let [data (types/json->clj result)]
+    (fx/merge cofx
+              (validate-access-result data password))))
```
janherich: There is no need to use `fx/merge` with only one function, could be just `(validate-access-result cofx data password)`

janherich (approved):

src/status_im/accounts/access/core.cljs
alwx: agree, will update in a couple of mins

yevh-berdnyk:
hey @alwx , could you please add accesibility ids for `Recovery phrase`, `Password`, `Confirm password` input fields, `Next` and `Access` buttons here?

alwx:
@yevh-berdnyk done!

yevh-berdnyk:
@alwx I've aligned e2e tests, so please update you local branch

lukaszfryc:
@alwx 

1) When the same account is recovered for the 2nd time with a different password, the error tooltip is not fully visible



2) When seed phrase is misspelled we show a [warning](https://github.com/status-im/status-react/pull/5494). In the new flow, the warning is show at the very end when user already lost the context of the seed phrase. Can we validate the phrase and show the warning on the first step with the recovery phrase?


3) When I read "You'll need this password..", it's not clear for me if I should use my old password here or what. I'd much rather prefer something like "**Set a password** to open the app...". But, I think we can create a separate issue for it. @obi2020 wdyt?

yevh-berdnyk:
@alwx can the same accessibility ids be added for the same elements in create account flow?
password input
confirm password input
next button
P.S sorry, I thought they are inherited

obi2020:
@lukaszfryc I think we should address it in the header. i.e. "Create a new password" Would that require add'l logic to be created. I assume now the same screen is being presented as during initial account creation.

lukaszfryc:
> I think we should address it in the header. i.e. "Create a new password" Would that require add'l logic to be created. I assume now the same screen is being presented as during initial account creation.

@obi2020 currently the screen with account creation does not have any title:



But, I agree with you, let's change "Access account" to "Create a new password" in the password and the confirm password screens:



----

@alwx could you change it and fix https://github.com/status-im/status-react/pull/5999#issuecomment-425900249? Thanks!

alwx:
> When the same account is recovered for the 2nd time with a different password, the error tooltip is not fully visible

That can be a very tricky thing because our toolbar cannot be overlapped by tooltips. However, checking how to fix.

lukaszfryc:
@alwx any update on it? Could you fix rest of things from https://github.com/status-im/status-react/pull/5999#issuecomment-426196820?

oskarth:
Closing PR as inactive. Feel free to reopen if this is in error / when it is ready.

PR #6006, opened 2018-09-26
Fix/chat loading time

### LINKED ISSUE #5112: Cannot quickly switch between chats (TouchableHighlight is working slowly)

### User Story

As a user, I want to switch between chats without lags or hangs

### Description

*Type*: Bug

*Summary*: when user has several chats, especially with a long history (like #status), during switching application can hang

#### Expected behavior
I can switch between chats

#### Actual behavior
I cannot switch between chats or switching is working extremely slowly

### Reproduction

1. Open status-desktop
2. Sign Up
3. Join to several public channels
4. Switching between them

### Additional Information

MacBook Pro (Retina, 15-inch, Mid 2015)
*OS:* Mac OSx High Sierra 10.13.5
*Status version:* Status version: desktop build 10/07/18 (after merge https://github.com/status-im/status-react/pull/4752)
*NOTE:* according to @siphiuel comments, it is RN-related issue

### Logs
* Video: http://take.ms/g1qNC
* Video when you have several chats with 40+ unread messages: http://take.ms/PMEST
* Log: [StatusIm.txt](https://github.com/status-im/react-native-desktop/files/2179254/StatusIm.txt)

vkjr:
fixes #5112

### Summary:
Loading a chat with a lot of messages takes huge time because ScrollView renders all elements at once. Optimizing `Text` loading time in `react-native-desktop` didn't make significant improvement in speed. But ScrollView has a good performance on adding new elements (it doesn't recreate old ones). So I tried "partial loading".  
Lets say we have 100 messages in a chat. At first render only 5 of them will be displayed. When rendering finished, 5 more will be requested until all messages processed. With this approach js doesn't require native side to render all elements immediately, so UI operates better.

### Steps to test:
- Open Status
- Open few chats with a lot of messages and try switching between them

status: ready

### DIFF
--- a/src/status_im/ui/screens/desktop/main/chat/views.cljs
+++ b/src/status_im/ui/screens/desktop/main/chat/views.cljs
@@ -157,15 +157,26 @@
            [react/view {:style (message.style/delivery-status outgoing)}
             [message/message-delivery-status message]]])}))))
 
+(def load-step 5)
+
+(defn load-more [all-messages-count messages-to-load]
+  (let [next-count (min all-messages-count (+ @messages-to-load load-step))]
+    (reset! messages-to-load next-count)))
+
 (views/defview messages-view [{:keys [chat-id group-chat]}]
   (views/letsubs [messages [:get-current-chat-messages-stream]
-                  current-public-key [:get-current-public-key]]
-    (let [chat-id* (atom nil)
-          scroll-ref (atom nil)
+                  current-public-key [:get-current-public-key]
+                  messages-to-load (reagent/atom load-step)
+                  chat-id* (reagent/atom nil)]
+    {:component-did-update #(load-more (count messages) messages-to-load)
+     :component-did-mount  #(load-more (count messages) messages-to-load)}
+    (let [scroll-ref (atom nil)
           scroll-timer (atom nil)
           scroll-height (atom nil)
           _ (when (or (not @chat-id*) (not= @chat-id* chat-id))
-              (reset! chat-id* chat-id))]
+              (do
+                (reset! messages-to-load load-step)
+                (reset! chat-id* chat-id)))]
       [react/view {:style styles/messages-view}
        [react/scroll-view {:scrollEventThrottle    16
                            :headerHeight styles/messages-list-vertical-padding
@@ -182,7 +193,7 @@
                            :ref                    #(reset! scroll-ref %)}
         [react/view
          (doall
-          (for [[index {:keys [from content message-id type value] :as message-obj}] (map-indexed vector messages)]
+          (for [[index {:keys [from content message-id type value] :as message-obj}] (map-indexed vector (take @messages-to-load messages))]
             ^{:key message-obj}
             [message (:text content) (= from current-public-key)
              (assoc message-obj :group-chat group-chat


vitvly (approved):

vitvly:
@vkjr Nice idea!

flexsurfer (approved):

src/status_im/ui/screens/desktop/main/chat/views.cljs
```diff
@@ -148,15 +148,27 @@
...
+    (reset! messages-to-load next-count)))
+
 (views/defview messages-view [{:keys [chat-id group-chat]}]
   (views/letsubs [messages [:get-current-chat-messages-stream]
-                  current-public-key [:get-current-public-key]]
-    (let [chat-id* (atom nil)
-          scroll-ref (atom nil)
+                  current-public-key [:get-current-public-key]
+                  messages-to-load (reagent/atom 5)
+                  chat-id* (reagent/atom nil)]
```
yenda, line 170: messages-count could be binded there instead of recalculating it every time in load-more

src/status_im/ui/screens/desktop/main/chat/views.cljs
```diff
@@ -148,15 +148,27 @@
...
+    {:component-did-update #(load-more messages messages-to-load)
+     :component-did-mount  #(load-more messages messages-to-load)}
+    (let [scroll-ref (atom nil)
           scroll-timer (atom nil)
           scroll-height (atom nil)
           _ (when (or (not @chat-id*) (not= @chat-id* chat-id))
-              (reset! chat-id* chat-id))]
+              (do
```
yenda, line 177: why is this here in the first place ?

src/status_im/ui/screens/desktop/main/chat/views.cljs
vkjr: Do you mean creating atom outside defview?

src/status_im/ui/screens/desktop/main/chat/views.cljs
vkjr: Will putting it on a second place  make any difference?

src/status_im/ui/screens/desktop/main/chat/views.cljs
yenda: no creating it here `message-count (count @messages)`

src/status_im/ui/screens/desktop/main/chat/views.cljs
yenda: no I am asking for the purpose of this code ? Normally you shouldn't have any logic in the view description. So I'm wondering why it has to be there. I think the messages-to-load can be updated in :component-did-update ? What is chat-id* for ? If that is because you can switch chats and the component will still be mounted, I think that means the component logic and parameters are flawed.

src/status_im/ui/screens/desktop/main/chat/views.cljs
vkjr: Got it, agree, will fix

src/status_im/ui/screens/desktop/main/chat/views.cljs
vkjr: Yes, this is because changing chat-id means that we switched to new chat and that totally new messages will be loaded. So we need to reset `messages-to-load` counter and again load messages gradually.

yenda: but if chat-id changes, isn't the component updated again ? at least this logic should also be in component-update ?

vkjr: The problem here is that `:component-did-update` calls after component rendered.
In case when chat-id changed, we need `messages-to-load` value to be updated before first render or all optimization will fail. And looking at react component lifecyle seems that `render()` is appropriate place for this.

maxhora (approved):

src/status_im/ui/screens/desktop/main/chat/views.cljs
vkjr: Per discussion with @yenda, this problem moved to separate issue to be fixed later - #6033

yenda (approved):

churik:
@vkjr 
Chat history area is empty for both public and 1-1 chats

hist

Video: http://take.ms/slBdd

<blockquote><div>Monosnap screenshot tool</div><div><strong><a href="https://monosnap.com/file/WhXNwH8YyqaPPHbxEYVBIQcAcJqFCM">File "screencast 2018-09-28 12-47-49.mp4"</a></strong></div><div>Monosnap — the best tool to take & share your screenshots.</div></blockquote>

vkjr:
Thanks, checking it...

churik:
Tested chat switching between 1+ week history, account creation and restoring.

PR #6014, opened 2018-09-26
[fix 6009] do not allow public chat named with valid contact key

### LINKED ISSUE #6009: Public and 1-1 chat cannot be used if contact code for 1-1 chat and public chat name are matched
### Description

*Type*: Bug
*Summary*: if you create any public chat with contact code name (i.e. `0x040fb3b30e25e8ba6576e1f5092bc1a210e8b43c56521a677c2a2c3f6a32ebda1a1971f396e5cbb4f6a108b788ff4f929f94d896589b6e3529b8f5da6ce710c214`), you cannot use 1-1 chat with user with same contact code.

#### Expected behavior
redirected to 1-1 chat; have both 1-1 and public chat in list

#### Actual behavior
redirected to Public chat; 
1-1 chat was replaced by Public chat, which contains history of 1-1 chat.

### Reproduction

- Open Status
- Join any public channel
- Tap on user profile picture 
- Send him a message
- Copy his contact code
- Go back to Home
- Tap on + > Join Public chat 

You also will be redirected to public chat if you select user from contact list.

### Additional Information

* Status version: nightly 26/09/2018
* Operating System: Android, iOS, Mac OS x

#### Logs

* TF session: https://app.testfairy.com/projects/4803590-status/builds/8606370/sessions/4402358100/?accessToken=DzcGTxiCA-6-awX4aMOkUAwrCCU

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803590-status/builds/8606370/sessions/4402358100/?accessToken=DzcGTxiCA-6-awX4aMOkUAwrCCU">TestFairy: Status 0.9.29 (9168) @2018</a></strong></div></blockquote>

yenda:
fixes #6009 

### Review notes (optional):
Should we update other translations for the error message when creating public chat ? Or should we delete the current one to make it obvious they should be updated ? @chadyj 

screenshot_1537964878

status: ready

### DIFF
--- a/src/status_im/ui/screens/add_new/new_public_chat/db.cljs
+++ b/src/status_im/ui/screens/add_new/new_public_chat/db.cljs
@@ -8,3 +8,8 @@
 
 (spec/def ::topic (spec/and :global/not-empty-string
                             (partial re-matches #"[a-z0-9\-]+")))
+
+(defn valid-topic? [topic]
+  (and topic
+       (spec/valid? ::topic topic)
+       (not (spec/valid? :global/public-key topic))))
--- a/src/status_im/ui/screens/add_new/new_public_chat/subs.cljs
+++ b/src/status_im/ui/screens/add_new/new_public_chat/subs.cljs
@@ -5,8 +5,9 @@
             [cljs.spec.alpha :as spec]))
 
 (re-frame/reg-sub
- :new-topic-error-message
+ :public-chat.new/topic-error-message
  :<- [:get :public-group-topic]
  (fn [topic]
-   (when-not (spec/valid? ::db/topic topic)
+   (when-not (or (empty? topic)
+                 (db/valid-topic? topic))
      (i18n/label :topic-name-error))))
--- a/src/status_im/ui/screens/add_new/new_public_chat/view.cljs
+++ b/src/status_im/ui/screens/add_new/new_public_chat/view.cljs
@@ -9,7 +9,7 @@
             [status-im.ui.components.text-input.view :as text-input.view]
             [status-im.ui.components.toolbar.view :as toolbar]
             [status-im.ui.components.tooltip.views :as tooltip]
-            [status-im.ui.screens.add-new.new-public-chat.db :as v]
+            [status-im.ui.screens.add-new.new-public-chat.db :as db]
             [status-im.ui.screens.add-new.new-public-chat.styles :as styles]
             [status-im.ui.screens.add-new.styles :as add-new.styles]
             status-im.utils.db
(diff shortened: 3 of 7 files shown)


janherich (approved):

chadyj:
> Should we update other translations for the error message when creating public chat ? Or should we delete the current one to make it obvious they should be updated ?

Thanks for spotting. I'll update in Lokalise cc @obi2020

obi2020:
Are numbers now allowed in topic names?

yenda:
they are now that is why this problem surfaced recently since keys contain numbers

obi2020 (changes requested):
Please edit the alert copy as follows: 

"Use only lowercase letters (a-z), numbers, and dashes (-). Do not use contact codes or addresses"

Note my edit in the second half especially. This might be a wider net than is necessary, but I think it's more approachable. Does it work?

yenda:
So should I also ban addresses and not only contact codes ?

obi2020:
I don't know the tech implications, of course, but I chose the wording because people conflate code/address/key. In any case, maybe it's better not to have addresses either, which could be spammy donation requests

obi2020 (approved):

obi2020 (changes requested):
sorry, see previous edit request

yenda:
@obi2020 I changed the error to `Use only lowercase letters (a to z), numbers & dashes (-). Do not use contact codes`. I don't see how topics using addresses can be a problem here, the only problem was that creating a public chat with the a public key as a name was conflicting with the 1-1 chat with a contact that has this public key. Addresses are a different thing that is not used in chat. There is no way to spam simply by joining a chat with an address as a name, why would anyone join such a chat anyway ?

lukaszfryc:
@yenda [desktop build](https://ci.status.im/job/status-react/job/combined/job/desktop-linux/1729/) failed. Could you check it?

```
[desktop-linux] Running shell script
+ cp -r ./deployment/linux/usr ./StatusImPackage/AppDir
[Pipeline] sh
[desktop-linux] Running shell script
+ cp ./deployment/env ./StatusImPackage/AppDir/usr/bin
cp: cannot stat './deployment/env': No such file or directory
```

obi2020 (approved):
alert edits lgtm

yenda:
@lukaszfryc it was just a CI hiccup, next build should be good

flexsurfer:
from user pov it looks strange, why user can't use any codes he wants for topics?

yenda:
@flexsurfer good point, a user might want to have a public chat named after his identity. The question is do we want to this change to make sure people don't accidentally overwrite their chat, or do we want to support this case which implies important changes to app-db structure and a lot of refactoring to be done ? Because the name of the public chat is its key in the chat map so it will conflict with the key of a 1-1 chat which is the identity of the contact.

SlizkovaGanna:
@yenda 
For desktop the fix did not apply, and commit-hash is different from what it is supposed to be for a PR build, so perhaps those changes are not included?

image

Tested for new user, existing user, recovered used; for current's user contact code and different's user contact code on mobile - no issues found.
Error message is: **Use only lowercase letters (a to z), numbers & dashes (-). Do not use contact codes** as expected for mobile

Tested on:
mobile: iOS 12.0.1 and Android 7.0 
desktop: MacOS 10.13.4

rasom:
Why can't we allow contact key as pubchat name? can't we just add some prefix to pubchats' chat-id so that both could coexist in the database?

yenda:
@rasom we can in the future but that probably requires some UX improvements first, because there is only cases were people use a pubkey as new public-chat by mistake so far. So even if we allow it in the future we should have some kind of warning that this is a public chat the user is creating, and ask if he's sure he doesn't want to simply add a contact.

rasom (approved):

SlizkovaGanna:
There are no issues with this PR
Tested for manual input and copy/paste, different users.
Failed tests are known (№1 - https://github.com/status-im/status-react/issues/5899; №3 - https://github.com/status-im/status-react/issues/5975) and false failures.

A desktop issue with error messages was reported separately in https://github.com/status-im/status-react/issues/6395

Tested on:
iOS 12.0.1
 Android 7.0 
MacOS 10.13.4

Issue #4608, opened 2018-06-03
Allow replies to specific messages in public chats


rachelhamlin:
### User Story

As a user, I want to reply to specific messages in **public, 1:1 chats or group chats** so that conversations are easier to follow.

### Description

*Type*: Feature

*Summary*: Double tap on messages in busy chats to reference the message in a reply.

#### Expected behavior
Double tapping on a message allows me to reply with a snippet of the original message included in my own.

Tapping on my reply scrolls back to the original message.

<img src="https://user-images.githubusercontent.com/15470435/40890198-8a2274de-6772-11e8-8c22-444274d2a0b0.png" width = 250 />

#### Actual behavior
No response when I tap on a message.

lukaszfryc:
I think this may require more advanced UI/UX. Simply citing a message may create even more mess in the chat. We may want to research messages threads, like for instance Slack does it.

cc @errorists

lukaszfryc:
Btw, Telegram has really a nice interaction for replying to a message. User can reply to any message by swiping it to the left:

<img src="https://user-images.githubusercontent.com/7532782/42754124-9098235c-88f3-11e8-9b4d-4ce6c86cd3cd.png" width="200">

The reply is quoted and generates a notification. 

We can start with this if we don't want to implement threads. Such quoting is definitely good enough for 1-1 chats.

errorists:
Here's a prototype of the design. I've decided to go with a long press rather than swipes since 

- I find it more discoverable than swipes, 
- we already use it,
- the swipes we have right now are rather janky,
- we create a unified behaviour for all message actions. 

👉 https://framer.cloud/XvSUR/

<blockquote><img src="https://framer.cloud/XvSUR/5/framer/social-1200x630.png" width="48" align="right"><div><strong><a href="https://framer.cloud/XvSUR/">Reply to Messages</a></strong></div><div>Long press on me​ssages to see possible actions</div></blockquote>

rachelhamlin:
This looks great @errorists. 

When I click on a reply, it scrolls that to the top of the window, right? Is there any sort of highlight or animation to emphasize it? 

And if I'm the person being replied to, is there any visual distinction or would the copy simply say, "You"?

errorists:
@rachelhamlin good catch on both, thanks! 

> When I click on a reply, it scrolls that to the top of the window, right? Is there any sort of highlight or animation to emphasize it?

[...]

Issue #20, opened 2016-04-25
Depend on syng-lib using Lein Voom


0xc1c4da:
https://github.com/LonoCloud/lein-voom

PR #6019, opened 2018-09-26
Chat replies + refactoring

### LINKED ISSUE #6017: Fix random-id cofx
# Problem

The `:random-id` cofx is currently a random-id that was generated with our RNG, but since it is a single value, composing fx producing functions that need a random number could lead to reuse of that number

# Solution

Replace the current `:random-id` cofx by one that returns the RNG function instead.

# Acceptance criteria

the `:random-id` cofx has been replaced

janherich:

Implements #4608 + finished refactoring of the chat module (it's nice to implement a new feature and still end up with overall amount of LOC removed and that's with ~50LOC for migration code :) )

There are one thing not implemented for the feature yet, I'm opening PR anyway for review and merging, as it has potential to cause a lot of merge conflicts (that's why high priority):
* The functionality of scrolling back upon tapping on quoted message is not yet implemented, will require more resources/thought

### Testing notes (optional):
Light regression testing of chat + feature testing
fixes #6017 (same tests)

status: ready


src/status_im/utils/universal_links/core.cljs
```diff
@@ -42,13 +42,13 @@
...
   (log/info "universal-links: handling public chat " public-chat)
-  (chat.events/create-new-public-chat cofx public-chat false))
+  (chat/start-public-chat cofx public-chat false))
 
 (fx/defn handle-view-profile [{:keys [db] :as cofx} profile-id]
   (log/info "universal-links: handling view profile" profile-id)
   (if (new-chat.db/own-whisper-identity? db profile-id)
     (navigation/navigate-to-cofx cofx :my-profile nil)
-    (chat.events/show-profile cofx profile-id)))
+    (navigation/navigate-to-cofx (assoc-in cofx [:db :contacts/identity] profile-id) :profile nil)))
```
yenda, line 51: would be cleaner to do that through fx/merge

src/status_im/chat/models.cljs
```diff
@@ -1,13 +1,18 @@
 (ns status-im.chat.models
```
yenda, line 1: this should be named core

src/status_im/events.cljs
```diff
@@ -441,18 +446,138 @@
...
+
+(handlers/register-handler-fx
+ :chat.ui/delete-message
+ (fn [cofx [_ chat-id message-id]]
+   (chat.message/delete-message cofx chat-id message-id)))
+
+(handlers/register-handler-fx
+ :chat.ui/show-profile
+ (fn [cofx [_ identity]]
+   (navigation/navigate-to-cofx
```
yenda, line 530: should be a navigate-to-profile function that does this in a fx/merge

src/status_im/utils/universal_links/core.cljs
yenda: actually it's the same thing here, should be navigate-to-profile fn

src/status_im/events.cljs
janherich: Good point, will change that.

yenda:
Could you give some pointers on how you implemented the support for replies ?

janherich:
@yenda sure, I changed `:content` field to always be a map (it was already case for commands, now it's for all message types, so more regular), and this content can contain `:response-to` reference besides `:text` value.

Backwards compatibility is done through `update-legacy-data` function in message receiving flow (was previously `update-legacy-commands`) which turns string `:content` field into `{:text string-content}` map, clean upgrade path is assured through migrations. 
Other options was to introduce `:response-to` directly to the top level message, so it would be encoded in the transit message key in protocol layer, but I didn't like it for multiple reasons:
* It wouldn't ensure compatibility of the old app versions with messages from new versions anyway
* `response-to` is optional and so would be mostly empty
* It would continue the irregular pattern of sometimes having string, sometimes map (command messages) in content
* More work for no real gain

dmitryn (approved):

src/status_im/utils/universal_links/core.cljs
janherich: Will be done when the profile module is refactored into agreed pattern.

src/status_im/events.cljs
janherich: Will be done when the profile module is refactored into agreed pattern.

yenda (approved):
Looks good. As mentioned we should see with @chadyj if we do a one commit release that will ensure we have a transition release for this breaking change

flexsurfer (approved):

test/cljs/status_im/test/mailserver/core.cljs
```diff
@@ -182,7 +182,7 @@
 
 (deftest delete-mailserver
   (testing "the user is not connected to the mailserver"
-    (let [cofx {:random-id "random-id"
+    (let [cofx {:random-id-generator #(identity "random-id")
```
yenda: can you replace identity by constantly ?

Serhy:
Hey @janherich, something wrong with the latest changes (or builds): https://github.com/status-im/status-react/pull/6019#issuecomment-425003757

No messages at all are received neither in public not in 1-1 chats (both iOS and Android builds)
We were able to fetch offline inbox from default public chats (#status, #dapps...) but sending the message there (again in `#status` public) is not visible from release/develop builds.

Quote of my own messages still works, so continue testing with that limitation for now

yenda:
@Serhy That is the expected behavior, that's why we need a transition version. Did you test the previous version without latest changes ? Our expectation is that it would have crashed older clients

Serhy:
The interaction between `Release 0.9.28` <-> https://github.com/status-im/status-react/pull/6019#issuecomment-425003757 doesn't crash the apps, but `Release 0.9.28` doesn't see any messages (and quoted replies) sent in public/1-1 with this changes.

So no crashes, but 
- anything our future "Release 0.9.29" (with current PR changes) will send in 1-1/public/(group!?) chats -> won't be received by "Release 0.9.28"(and older builds)
- anything current "Release 0.9.28" users send in chats -> received fine by future "Release 0.9.29"

Serhy:
Actually, I realised that https://github.com/status-im/status-react/pull/6019#issuecomment-425048709 may not be true, because of interaction, as I mentioned earlier, - doesn't work between the same builds https://github.com/status-im/status-react/pull/6019#issuecomment-425003757 at all.
I.e.:
1) Install https://status-im.ams3.digitaloceanspaces.com/StatusIm-180927-075820-13dec2-pr.apk on Device1 and Device2
2) Create new accounts on both devices
3) Join the same public chat on both devices
4) Send messages in this public by both devices and check whether any of contacts received a message.

<blockquote></blockquote>

churik:
Same for desktop -> messages, sent to public, 1-1 chats are not received on other devices.
Several more desktop-specific issues:

#### Last message in chat list is shown as json
json

#### No replies are implemented in desktop
@chadyj I believe we should create separate issue for this, WDYT?

janherich:
@churik Yes, that's to be expected, desktop support for replies will be implemented in separate PR.
Regarding the other issues, I'm looking at that, although it's very strange, when I was testing it with 2 accounts on the same device (this PR), 1-1 chats and public chats were working without problems.
Also this PR should receive any messages from the older builds just fine, the opposite (receiving message on older build from this PR build) is unfortunately not supported out of the box, but we solve that with the special "compatibility" release @yenda mentioned.

janherich:
@Serhy I just tested interaction between two devices (iOS simulator & android real device) with the latest build above, and everything works fine, including receiving messages from older devices, please perform the testing on latest build.

asemiankevich:
@janherich thanks Jan, checking!

asemiankevich:
So @janherich messaging works fine, thanks! Some things i have noticed so far:

1. Cross button is hardly accessible on Android
image

2. Tapping on my reply does not actually scroll to the original message i have quoted (i assume should be , based on ticket itself)

3. Can't quote comands (request / send)

4. It seems that this bottom line is a bit different color than expected (i think should be less bright) but i am not sure :D 

image

5. UPD: also when i quote my messages it should be indicated as `You` at the top (as on mocks). It does not behave like that now.

image

could you pelase ellaborate on this? cc @errorists  (points 2 , 3, 4, 5)

errorists:
@asemiankevich Thanks for the screenshots, I couldn't get this build running on my iPhone before. 

> Cross button is hardly accessible on Android

the 'x' button should float above and not be clipped by the parent layer. Afraid I can't give other directions that pointing to this design spec with the correct position outlined [zpl.io/aRYAgnv](http://zpl.io/aRYAgnv)

> Tapping on my reply does not actually scroll to the original message i have quoted (i assume should be , based on ticket itself)

I believe Jan mentioned that scrolling to original messages doesn't work yet, not sure

> Can't quote comands (request / send)

You should not be able to quote transaction messages, it's by design.

> It seems that this bottom line is a bit different color than expected (i think should be less bright) but i am not sure :D

You're right, the correct colour is #EEF2F5 for left messages and #cad1ed for right. The lines should be 2px in height with a 1px rounded corner. Also the reply icon on messages is too low, the top margin is 8px -- if it's my eyes playing tricks and it's in the right position then the text is too high by 2px. Either way the little reply icon and the original sender text label should be aligned. Pretty please, can we have all this fixed ? :) 

> UPD: also when i quote my messages it should be indicated as You at the top (as on mocks). It does not behave like that now.

Absolutely right

janherich:
Thanks for the valuable input @asemiankevich @errorists !
1. Scrolling to message is unfortunately very hard to do, and won't be supported for now - maybe in the future, when we figure out reliable way to do it.
2. Quoting command messages is not allowed by design

I will fix rest of the issues (wrong colours, wrong margins, placement of the cancel element on Android, `you` instead of your actual name) ASAP :)

Serhy:
@janherich Few more points to add:
1) Default image icon is shown in the quote snippet when I quote my own message, while I have set non-default user profile:
Please see TF session: https://app.testfairy.com/projects/4803622-status/builds/8608874/sessions/4402477549/?accessToken=utGpftmT8DmGlyO6dJSYgSheGXA
2) There is no ability to reply to a previous reply (no `Reply` option on a long tap of the replied message). In that case, only the ‘reply’ part and not the quote snippet should be replied to.
3) No ability to reply to big messages (those messages with ‘Show more…’ action in message bubble): quote snippet extends message input so much down it’s out of a tappable area. 
Correct behaviour nicely reflected in design: “The quote snippet should allow for max 5.5 lines of text (140pt in height) after which it’s clipped. The entire text area should scroll.”
**To reproduce:**
* Open Status and join any public chat
* Send a 30-line long message
* Long tap on the message -> `Reply`
TF session: https://app.testfairy.com/projects/4803622-status/builds/8608874/sessions/4402478075/?accessToken=9i/bN5AtERazTOb1WQUgRnmOUm0

screen shot 2018-09-27 at 18 33 23

janherich:
@Serhy Thanks for the additional input, re point 2.) I believe that's correct by design, eq one is not allowed to quote already quoted messages, am I correct @errorists ?
Point 1) I used the same element used to render user icon in public/group-chat messages, will investigate why it's not working for custom photos.

errorists:
>  I believe that's correct by design, eq one is not allowed to quote already quoted messages…

Sorry if the documentation wasn't clear enough on this, but Serhy is right here, you should be able to reply to replies, when doing so you're only quoting the reply part of it and not the quote (so you're only quoting the black text below the line in your reply). It's better illustrated in this prototype, scroll up and try replying to the message with the quote https://framer.cloud/GzkTq/ 


<blockquote><div><strong><a href="https://framer.cloud/GzkTq/">UXR Reply Messages Android</a></strong></div><div>This project was made with Framer, a design tool for turning original concepts into working prototypes.</div></blockquote>

janherich:
@errorists @lukaszfryc @asemiankevich 
I the new build (soon to be posted) I fixed all the styling/color issues, name issue (it's `You` for own messages), fixed the quoted message height to 140px (in quote input) and enabled quoting of already quoted messages.
Please re-test those things when possible, I'm still working on:
* Truncated quoted message to be scrollable while typing response to the message
* Android z-index issue for cancel button
* Showing custom image icon when set

janherich:
@errorists @asemiankevich @Serhy All of the Issue above should be fixed in the latest (#20) build, except two I'm still working on:

* Close reply icon being displayed correctly on Android
* Quoted message text in reply input scrollable if it's cropped at 140px max height

Please test that all remaining issues are fixed and/or report any new issues, thanks :)

errorists:
Hey @janherich fantastic work. I'm definitely still seeing bugs when the quoted snippet is long, attached. 
Also we should truncate the quote in the chat message bubble if it's too long, from the design 'The maximum height for the quote snippet is 5 lines, that’s about 90px. After that we truncate
the snippet.' so if the text overflows we add a '…' at the end.


img_0595

janherich:
Great, thanks for the input @errorists, I will work on that :)

errorists:
Cool! Also, could the avatar be aligned with the bottom of the snippet like in the design and not centred? This way it looks similar in style to our messages. I'm sure we could do some flexbox magic to make that happen :) The distance from the avatar to bottom should be 12px if it helps
I'm not sure on this one so correct me if I'm wrong, but the size of the text in the quote snippet (the one on the screenshot attached above) should be 14px, I think it's 15 now.

janherich:
Build with correctly displayed close reply button even on Android + correctly truncated long messages (both in input and message bubbles) coming, I will work on rest of the things (like last @errorists points) in the afternoon :)

src/status_im/transport/message/transit.cljs
```diff
@@ -92,6 +92,8 @@
                               "c3" (fn [[name profile-image address fcm-token]]
                                      (v1.contact/ContactRequestConfirmed. name profile-image address fcm-token))
                               "c4" (fn [[content content-type message-type clock-value timestamp]]
+                                     (v1.protocol/Message. (if (string? content) {:text content} content) content-type message-type clock-value timestamp))
```
mandrigin: The condition should be inverse to handle malformed messages. If we are sure that it’s a dictionary with “text” key that lead to a string, pass it. Otherwise, convert the value to string and create a dict.

mandrigin (changes requested):

src/status_im/transport/message/transit.cljs
janherich: To be honest, it's not so easy, because for example we already have messages (command content type) where content is a map (and doesn't contain text).

asemiankevich:
@janherich what about issue on desktop? 
image

janherich:
@asemiankevich As I said, there is no way around that except releasing special "transition" version between current release and release including this PR.
Of course desktop build of this PR should display the text properly, is the screenshot from desktop build of this PR ?

asemiankevich:
@janherich yes, latest build. Sorry if it was already discussed, it is a long thread already(

janherich:
@asemiankevich I understand, then I will fix the desktop

mandrigin (approved):

janherich:
@errorists @asemiankevich @Serhy :
There are just couple of styling bugs remaining and the message logic was changed that the new build messages don't kill the older builds (they are just not displayed in older builds).
Given that the PR has very big potential for merge conflicts, and it's very tedious and time consuming to keep it up to date with current rate of refactoring and other PRs landing in (basically fully occupying me and I can't do much else), I would propose to merge it **if** there are no other bugs then minor styling issues and solve them later in the develop, what do you think ?

errorists:
@janherich sure, this is all great work here. Be sure to reach out on Status/Slack if you need help on solving any of those minor styling issues.

asemiankevich:
Migrations does not work

image

STR:

have 2 devices. device 1 is running Google play build, device 2 is running PR build
- on Device 1 open status and create 1x1 chat with device 2
- on Device 1 send text, link and emoji in this chat
- on Device 1 open #test public chat
- on Device 1 upgrade the app to current PR
- on Device 1 open 1x1 chat

janherich:
@asemiankevich I forgot to correctly migrate emoji messages :/ 
Already fixed, build is on the way, together with last styling fixes (avatar placement) for @errorists

asemiankevich:
So here is the assumption:

- all issues outlined including design are fixed
- app is not crashing when upgrading
- newly replied messages **wont be visible** for older versions
- desktop issue is fixed as well

Since the PR is huge and it is rather difficult to maintain and rebase i would like to merge it right now with the assumption some small bugs could pop up in develop ( but no crashes, blockers , etc).

cc @annadanchenko @Serhy

PR #6020, opened 2018-09-26
Remove unnecessary fonts from our app bundle.

### LINKED ISSUE #5998: Remove the San Francisco font files from the 'ios' directory.

### User Story

As a developer, I don't need to bundle the San Francisco font files inside the app as they are already provided by iOS.

### Description

*Type*: Bug

*Summary*: Currently we include San Francisco font files in our builds. It's not necessary, as they are already provided to us by iOS and can be reached via `fontFamily: 'System'` . Please remove the fonts and make sure they're not bundled and distributed with the app. 

Here's a list of font files

`status-react/ios` folder
- `SF-UI-Display-Medium.otf `
- `SF-UI-Display-Regular.otf`
- `SF-UI-Display-Semibold.otf`
- `SF-UI-Display-Thin.otf`
- `SF-UI-Text-Bold.otf`
- `SF-UI-Text-Light.otf`
- `SF-UI-Text-Medium.otf`
- `SF-UI-Text-Regular.otf`
- `SF-UI-Text-Semibold.otf`

Thanks!

mandrigin:
Removed unnecessary bundled fonts.

fixes #5998 

status: ready


yenda (approved):

flexsurfer (approved):

lukaszfryc:
Fonts in Android and iOS app have changed in this PR a bit. But, in my mind they look better now on both platforms. E.g. bold style is stronger now, see screen titles. @errorists could you take a look and let us know if you like them?


PR build on the left (**new fonts**), nightly on the right (fonts we've had so far):

1) Android




2) iOS 



----

Other screens in this PR:

screen shot 2018-09-28 at 11 17 54
screen shot 2018-09-28 at 11 17 40
screen shot 2018-09-28 at 11 17 28

mandrigin:
To me they also look better because now they use a correct system font instead of a wrong (and outdated) variation of SF and Roboto.
They should be more in line in how other apps on these platforms look like.

mandrigin:
I can tweak boldness if needed, right now it uses a plain “bold” variation instead of in-betweenish thing we had before.

lukaszfryc:
TBH, I like the boldness :)

errorists:
Wow, looks great. Thanks, everyone! Yeah, the differences come from updating to the actual pre-installed fonts shipped in newer versions of both OS, so we're closer to the native experience now.

However, there's a 'but':  we shouldn't use the `letter-spacing` property anymore because we're overriding the correct values set by the OS. That's why some of the fonts are too tight and close to each other and some like the 'Topic' and 'Selected' appear too far apart.

mandrigin:
I believe there is only one place where I tweaked letter spacing: “topic” and “selected”. Everything else is the default unless it is overridden somewhere. I can follow-up with another PR to figuring that out if you feel like this one is okay to merge.

errorists:
Totally, It's ok, this is good work just needs a bit of follow-up :)

errorists:
About the boldness, I wouldn't worry too much, we're due for a font change anyway soon. The font weights we'll be using are 400, 500 and 700 if that helps.

mandrigin:
Agree. And we need to do the same for Mac OS Desktop too

mandrigin:
@lukaszfryc I guess we can move it to “merge” then?

lukaszfryc:
@mandrigin yes!

PR #6025, opened 2018-09-26
[DO NOT MERGE] [fix 6017] fix random-id cofx

yenda:
replace random-id cofx by random-id-generator

this chance avoids the risk of using the same random-id
twice when it shouldn't be the case when merging many
fx producing functions

added to https://github.com/status-im/status-react/pull/6019 because requires same qa tests

status: ready

### DIFF
--- a/test/cljs/status_im/test/models/network.cljs
+++ b/test/cljs/status_im/test/models/network.cljs
@@ -96,8 +96,9 @@
 
 (deftest save
   (testing "it does not save a network with an invalid url"
-    (is (nil? (model/save {:random-id "random"
+    (is (nil? (model/save {:random-id-generator #(identity "random")
                            :db {:networks/manage {:url {:value "wrong"}
                                                   :chain {:value "1"}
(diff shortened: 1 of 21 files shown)


test/cljs/status_im/test/models/network.cljs
```diff
@@ -96,8 +96,9 @@
 
 (deftest save
   (testing "it does not save a network with an invalid url"
-    (is (nil? (model/save {:random-id "random"
+    (is (nil? (model/save {:random-id-generator #(identity "random")
```
cammellos, line 99: `(constantly "random")` is probably more idiomatic

cammellos (approved):

test/cljs/status_im/test/models/network.cljs
yenda: thanks I forgot about that one, I will replace it in #6019 once @janherich will have fixed conflicts with your recent merge

PR #6038, opened 2018-09-27
Enabled group chats & propagate signed membership information

cammellos:
### Changes
1) The format of a group-chat id is changed to include the initial admin pk, in order to prevent a member to propagate messages with identical id but different admin.
2) Signed group information are exchanged with each message so that even if the initial message is not received, any user will be able to create a chat if subsequent messages are received.
3) Validate `chat-id` on received messages.
4) Remove `receive` so that all messages go through `receive-many`

3 is necessary in some cases as it is possible to exploit wrapped messages.

I used the same strategy as before, to wrap a message into another message and allow to specify `chat-id`, for reference there is a good explanation here:
 
https://github.com/status-im/status-react/blob/4bd5d986b0669247b631ef1e4ee17b2a96aaec6a/src/status_im/transport/handlers.cljs#L133
https://github.com/status-im/status-react/blob/4bd5d986b0669247b631ef1e4ee17b2a96aaec6a/src/status_im/transport/message/v1/group_chat.cljs#L9

In the case of group chats, where the chat-id != sender-signature, we need to include the chat-id in the wrapper and then call `receive` on the wrapped message, with the `chat-id` taken from the wrapper.
This pattern is exploitable potentially by wrapping a `:user-message` with your `pk` as `chat-id` for example, so in order to avoid this care needs to be taken when wrapping messages & chat-id can be manipulated.

I added a `valid-chat-id?` function to defensively handle this cases.

### Next steps

* Handle group leave
* Handle new members
* rationalize the various group-chats files

### Testing notes

This enables group chats in nightlies, the features available:

1) Create a group chat
2) Exchange messages
3) Delete chat
4) Clean history

### Desktop known issues

* UI is not polished
* Missing checkboxes on contacts list, but clicking works

status: ready


dmitryn (approved):

alwx (approved):

src/status_im/native_module/core.cljs
```diff
@@ -59,3 +59,7 @@
 
 (defn is24Hour []
   (native-module/is24Hour))
+
+(def verify-group-membership-signatures native-module/verify-group-membership-signatures)
```
janherich, line 63: Why is this aliasing necessary ?

janherich (approved):

src/status_im/native_module/core.cljs
cammellos: Not sure, I kept the same pattern as other native-calls, basically we require `native-module.core` instead of `native-module.impl.module`, so any function needs to be aliased

lukaszfryc:
@cammellos I tested all platforms and noticed that:

1) iOS crashes when group chat invitation message is received
On Android, scan QR code of iOS user and then create group chat with the user. The iOS crashes after few seconds. [Stacktrace](https://app.testfairy.com/projects/4803590-status/builds/8611780/sessions/4402790359/?accessToken=Iy5pKXuIl0EJdXkQhI4TGjdWK50):
```
*** Terminating app due to uncaught exception 'RCTFatalException: Exception '*** +[NSString stringWithUTF8String:]: NULL cString' was thrown while invoking verifyGroupMembershipSignatures on target Status with params (
    "[[\"0x04d5091d15caf1ad58004f4cf5698a1891eaf398aed29d68229d3580f07e89eb5ad231d469a737c97841d8ce98f553783820c67ba2a94fdaf96be01358b00af0a80x04e5f4718000cdfebe64c87fe14e7362d2c3590909f412321a90a4fb91fd6999802ca8e4c801e5efcde84ed9ac05cfe3d675f1cf14f77e2310104091a8644086a30x04d5091d15caf1ad58004f4cf5698a1891eaf398aed29d68229d3580f07e89eb5ad231d469a737c97841d8ce98f553783820c67ba2a94fdaf96be01358b00af0a807f057ef-31b5-513d-9142-39e1649903940x04d5091d15caf1ad58004f4cf5698a1891eaf398aed29d68229d3580f07e89eb5ad231d469a737c97841d8ce98f553783820c67ba2a94fdaf96be01358b00af0a8\",\"b1ad0389eaf62a9b703656ae1f1578b71dd97f810e350a0e3a0a81cb53da75853c30ee08fcad60af309f5f9d81337d3b28bbf7bb587d222f7035cea498fd946100\",\"04d5091d15caf1ad58004f4cf5698a1891eaf398aed29d68229d3580f07e89eb5ad231d469a737c97841d8ce98f553783820c67ba2a94fdaf96be01358b00af0a8\"]]",
    1185
)
```

2) Group chat is not deleted when delete button is pressed in the group info screen (options->Group Info->Delete chat). All works fine when delete button from options menu is pressed.
 

3) There is also empty state missing when user does not have any contacts yet. But, this can be added later.


 

---

I tested on iOS 12, Android 8 and macOS 10.14.

cammellos:
@lukaszfryc thanks for reporting, the issues mentioned above (bar the empty state, that's for another PR probably), have been resolved, would you mind testing it again? 
Thanks!

lukaszfryc:
@cammellos https://github.com/status-im/status-react/pull/6038#issuecomment-425857934 are fixed but there are few more:

1) Error when chat name is cleared completely. Instead, user should be able to clear it but not save it.


2) When group chat is created, pressing back button navigates back to “Group chat” screen instead of list of chats in Home tab. To reproduce: on iOS, create a group chat and then swipe back.  

3) There are no checkboxes on iOS and desktop while selecting chat participants.


4) Every participant can change name of a group chat. The name changes only locally. I think it’s expected.

cammellos:
Thanks @lukaszfryc 
I will fix 1 & 2, and if you don't mind leave 3 for later.
4 currently is expected, but will change in next PR to allow only admins to change name.

would that be ok?

lukaszfryc:
sounds good @cammellos

cammellos:
@lukaszfryc 1 and 2 have been addressed, thanks!

lukaszfryc:
@cammellos I think it's ready to merge now. I will report those other bugs as separate issues.

lukaszfryc:
@cammellos I tested release builds with disabled group chats and interacted with the PR build (with enabled group chats). All worked fine. I did the same for 0.9.28 and did not notice any issues apart from backward compatibility issues related to other PR (chat replies).

PR #6041, opened 2018-09-27
[#4728] make [Enter] button default to proceed during account creation

### LINKED ISSUE #4728: Make [Enter] button default to proceed
### User Story

As a user I want to user `Enter` key in the application so i dont need to click Confirm / Proceed with a mouse

### Description

*Type*: Bug

*Summary*: Hitting enter does not confirm user's choice , for example when i set up account and want to confirm my selection i need to click `Continue` with a mouse

#### Actual behavior

Can't use enter key to confirm my choice where necessary 

image

rasom:
fix #4728

status: ready

### DIFF
--- a/src/status_im/ui/screens/accounts/create/views.cljs
+++ b/src/status_im/ui/screens/accounts/create/views.cljs
@@ -27,12 +27,14 @@
                       :input-placeholder (i18n/label :t/name-placeholder)
                       :input-description (i18n/label :t/name-description)}})
 
-(defview input [step error]
+(defview input [{:keys [step error password password-confirm]}]
   [text-input/text-input-with-label
    {:label             (get-in steps [step :input-label])
     :placeholder       (get-in steps [step :input-placeholder])
     :on-change-text    #(re-frame/dispatch [:accounts.create.ui/input-text-changed (get-in steps [step :input-key]) %])
     :secure-text-entry (boolean (#{:enter-password :confirm-password} step))
+    :auto-focus        true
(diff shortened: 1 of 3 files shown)


vitvly (approved):

flexsurfer (approved):

src/status_im/utils/config.cljs
```diff
@@ -33,7 +33,7 @@
 
 ;; CONFIG VALUES
 (def log-level
-  (-> (get-config :LOG_LEVEL "error")
+  (-> (get-config :LOG_LEVEL "debug")
```
yenda: default log-level in the absence of env file should stay error IMO why debug ?

rasom: yep sorry, that wasn't intended

rasom: fixed

churik:
@rasom thanks for PR, it works awesome for IOS, Android and Desktop on account creating.

The same would be great to have for "Add existing account" (mobile, desktop) and for "Start new chat", "Join public chat" (desktop, in mobile is already implemented) to have consistent behavior.

Can it be implemented in this PR as well?

rasom:
ok, done, for "Add existing account" it was added only to password input

churik:
*Tested for IOS, Android:*
- account creation, restoring (can proceed with "Go" key) without tapping on Sign in
- message sending wasn't changed
- can create public chat or enter contact code with "Go" key

*Tested for Desktop:*
- account creation, restoring (can proceed with "Enter" key) without clicking on Sign in
- message sending wasn't changed
- can create public chat or enter contact code with "Enter" key

PR #6049, opened 2018-09-28
change realm schemas directory layout

yenda:
the purpose of this change is to avoid the kind of mistakes that
happened in the past were copy/pasting imports led to wrong
migrations that went unnoticed
it will also make it easier to do migrations in the future as the
different versions of the same schemas are grouped together in the
same file and past migrations are grouped in the same file as well
so previous helper functions can be reused

### Steps to test:
- Install older release
- update to this one, should be able to create account

status: ready


dmitryn (approved):
👍

cammellos (approved):

rasom (approved):

goranjovic (approved):

lukaszfryc:
Works fine. I tested upgrade from Android and iOS. For iOS I used [release/0.9.27 build.](https://ci.status.im/job/status-react/job/release/job/release%252F0.9.27/14/)

PR #6057, opened 2018-09-28
fix store-event cofx position

yenda:
omg please let it be the last one

status: ready


janherich:
@yenda I already discovered and solved that in my reply-message PR which is undergoing last tests before being merged, so I propose to only merge this in case reply-message PR won't be merged in 2-3h max, what do you think ?

yenda:
@janherich it's better to have them separate especially if we release a version without your PR in it. Is there other cofx positions you fixed in your PR ?

janherich:
No, only this one, you are right, we should merge this because of the cherry-picking for release.

janherich (approved):

cammellos (approved):

PR #6064, opened 2018-09-28
Validate incoming message

yenda:
First draft, will go back to it on Monday

- add validate method to StatusMessage protocol
- spec all message types for use in validate method
- use valid method after transit/decode step to reject
invalid messages

status: ready

### DIFF
--- a/src/status_im/transport/db.cljs
+++ b/src/status_im/transport/db.cljs
@@ -1,7 +1,9 @@
 (ns ^{:doc "DB spec and utils for the transport layer"}
  status-im.transport.db
   (:require-macros [status-im.utils.db :refer [allowed-keys]])
-  (:require [cljs.spec.alpha :as spec]))
+  (:require [cljs.spec.alpha :as spec]
+            status-im.ui.screens.contacts.db
+            [clojure.string :as s]))
 
 ;; required
 (spec/def ::ack (spec/coll-of string? :kind vector?))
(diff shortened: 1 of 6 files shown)


src/status_im/transport/db.cljs
```diff
@@ -35,3 +37,41 @@
...
+
+(spec/def :group-chat/admin :global/public-key)
+(spec/def :group-chat/participants (spec/coll-of :global/public-key :kind set?))
+(spec/def :group-chat/signature string?)
+
+(spec/def ::content (spec/conformer (fn [content]
+                                      (cond
+                                        (string? content) {:text content}
+                                        (map? content) content
+                                        :else :clojure.spec/invalid))))
```
janherich, line 58: We should probably validate map content more throughly, dependent on whether it's `command` or `text/plain`/`emoji` content type, certain keys has to be there in order to not crash the rendering. 
For `text/plain` it's just `:text` which has to be string (can by simply conformed to string by calling `(update content :text str)`, for command, it's little bit more complicated and I will provide you with map of required keys/types-of-keys.

janherich (changes requested):

mandrigin (changes requested):
Can we also add unit tests to this functionality? Looks like it can have a good ratio of effort vs result.

src/status_im/transport/db.cljs
janherich: So the spec:

`content-type`:`text/plain`
It has to be a map with required key `:text` of type string, optional key `:response-to` of type string as well.

`content-type`:`command`
It has to be a map with required keys `:command-path` of type `[string set(keywords or strings)]`  and `:params` of type `map(keyword->anything)`

lukaszfryc:
@yenda is there a way for me to send custom messages using a bot or something? I'd like to check different payloads to avoid recent [issue that was crashing the app](https://status-im.slack.com/archives/C88MZP11P/p1538472928000100).

cc @mandrigin

janherich:
@lukaszfryc I have some Clojure repl scripts to do that if you have dev environment running, or @cammellos maybe even has something shell based.

src/status_im/transport/db.cljs
yenda: done

yenda:
@mandrigin waiting for @gnl generative tests for that

janherich (approved):

mandrigin:
@yenda can we just have a list of JSONs to test for now?

gnl:
@mandrigin Proper gen-testing imminent, gimme a few minutes.

gnl (approved):

mandrigin (approved):

PR #6079, opened 2018-10-01
add dependency on libcairo2-dev

jakubgs:
Necessary for desktop builds.
Of course I will also have to build a new image and push it.


yenda (approved):

jakubgs:
Turns out this is not necessary after all.

PR #6089, opened 2018-10-02
Ensure version compatibility

janherich:

Fixes recently introduced incompatibility between newer and older status version.

To test this, just create a new chat and write some messages there from this build. Join the chat from other device (older release) and test that message is received just fine (and vice versa), then check that it's the case for current nightly as well.

status: ready

### DIFF
--- a/src/status_im/transport/message/transit.cljs
+++ b/src/status_im/transport/message/transit.cljs
@@ -46,9 +46,12 @@
 
 (deftype MessageHandler []
   Object
-  (tag [this v] "c7")
+  (tag [this v] "c4")
   (rep [this {:keys [content content-type message-type clock-value timestamp]}]
-    #js [content content-type message-type clock-value timestamp]))
+    (if (= content-type constants/text-content-type)
+      ;; append new content add the end, still pass content the old way at the old index
+      #js [(:text content) content-type message-type clock-value timestamp content]
+      #js [content content-type message-type clock-value timestamp content])))
 
 (deftype MessagesSeenHandler []
   Object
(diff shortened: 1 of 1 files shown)


src/status_im/transport/message/transit.cljs
```diff
@@ -102,8 +105,8 @@
                                      (v1.contact/ContactRequest. name profile-image address fcm-token))
                               "c3" (fn [[name profile-image address fcm-token]]
                                      (v1.contact/ContactRequestConfirmed. name profile-image address fcm-token))
-                              "c4" (fn [[content content-type message-type clock-value timestamp]]
-                                     (v1.protocol/Message. (safe-content-parse content-type content) content-type message-type clock-value timestamp))
+                              "c4" (fn [[legacy-content content-type message-type clock-value timestamp content]]
+                                     (v1.protocol/Message. (safe-content-parse content-type (or content legacy-content)) content-type message-type clock-value timestamp))
```
mandrigin, line 109: I would pass both `content` and `legacy content` into `safe-content-parse` and move the logic of selecting the right one there, so it is all in one place.

src/status_im/transport/message/transit.cljs
```diff
@@ -46,9 +46,12 @@
 
 (deftype MessageHandler []
   Object
-  (tag [this v] "c7")
+  (tag [this v] "c4")
   (rep [this {:keys [content content-type message-type clock-value timestamp]}]
-    #js [content content-type message-type clock-value timestamp]))
+    (if (= content-type constants/text-content-type)
+      ;; append new content add the end, still pass content the old way at the old index
+      #js [(:text content) content-type message-type clock-value timestamp content]
```
cammellos, line 53: should we not always maybe append `content` so we can more easily switch to `content` for other message types as well? (otherwise we need to care for those in never version), or that is not the plan and we plan to keep them separate by message-type?

mandrigin (approved):
one note, looks good otherwise

src/status_im/transport/message/transit.cljs
janherich: Other message types (only commands really) already have `content` as a map, so there is no need to append it twice there I think.

cammellos: It makes it a bit harder to upgrade now and use accretion,
if we add another field you need to remember to add an empty entry or whatever at the end of the array.
Say we add "username".
Now you would have to add a 2 fields on one message (one dummy ) and 1 to the first one, do you see what I mean?

janherich: I can see what do you mean, only all the additional fields will be in 99% added as additional key inside `content` map (`opts` style) and not as an additional field in the js array.
I will change it as you proposed anyway.

janherich: @cammellos Done, if that's what you meant (always appending content at the end of the array)

cammellos (approved):

goranjovic (approved):

asemiankevich:
so me and @churik have checked that messaging works between release, nightly and this PR both mobile and desktop (both sides).

PR #6091, opened 2018-10-02
Disable group chats in nightlies

cammellos:
In light of today discussion about compatibility it seems like nightlies are also used by end-users ( I was not aware of that), so care needs to be taken with breaking changes.

I was hoping to get some early feedback so I have enabled them in nightlies, but I want to make some protocol changes still, and at this point is less effort to just toggle them off again until next PR rather than turning them on and make those changes backward compatible.

skips QA!

status: ready


mandrigin (approved):
lgtm

janherich (approved):

PR #6093, opened 2018-10-02
[JS API] Request user action from Dapp

flexsurfer:
This PR is not about QR code, I only fixed QR code cancellation flow, and QR Code logic should be refactored for sure later.

About this PR:
currently, Dapp can request only data, but this PR introduces functionality to request user action, for example, scan QR, or choose chats, or contacts etc. 

There is no status API yet for this, and scan QR is only one we have, but it's different because it doesn't require permissions, that's why it has different data in `supported-permissions` 
```
{:yield-control? true
 :allowed?       true}
```
there is no data for permission, but `:allowed?       true` instead,
 for example for contacts it may look like
```
{
  :title          (i18n/label :t/wants-to-access-contacts)
  :description     (i18n/label :t/choose-contacts)	
  :icon            :icons/contacts
  :yield-control? true
}
```
and there will be two phases, 1 - request permission 2 - show control for getting data (for example list of contacts) , and only after user will select item, response will be sent to dapp

### DIFF
--- a/resources/js/web3_init.js
+++ b/resources/js/web3_init.js
@@ -6,8 +6,9 @@ function bridgeSend(data){
     WebViewBridge.send(JSON.stringify(data));
 }
 
-function sendAPIrequest(permission) {
+function sendAPIrequest(permission, params) {
     var messageId = callbackId++;
+    var params = params || {};
 
     bridgeSend({
         type: 'api-request',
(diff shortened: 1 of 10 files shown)


jeluard:
@flexsurfer Can you detail what this is about?
Also, is this still WIP?

yenda:
could you explain what this is about ?

goranjovic (approved):

flexsurfer:
sorry guys. @jeluard @yenda updated description

jeluard (approved):

PR #6098, opened 2018-10-02
upload PR build artifacts to a separate bucket

jakubgs:
All of our builds end up in the same bucket: `status-im`
https://status-im.ams3.digitaloceanspaces.com/index.html
Which is already ~300 GB in size and really hard to browse through.
I've created a separate bucket named `status-im-prs` so we can keep them separate.
This will also make it easier to clean up old ones we don't care about anymore.

<blockquote></blockquote>

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -17,7 +17,8 @@ pipeline {
         println "Current JOB: ${env.JOB_NAME}"
         /* load common lib */
         cmn = load('ci/common.groovy')
-
+        /* just for a shorter access */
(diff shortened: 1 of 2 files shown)


jakubgs:
It works:
https://status-im-prs.ams3.digitaloceanspaces.com/index.html

<blockquote></blockquote>

mandrigin (approved):

jakubgs:
I've also done some code cleanup in the Jenkinsfiles and added a function that sets build descriptions with DigitalOcean spaces public links for download.

janherich (approved):

PR #6099, opened 2018-10-02
Fixup command validation

yenda:
fix validation of commands

status: ready

### DIFF
--- a/src/status_im/transport/db.cljs
+++ b/src/status_im/transport/db.cljs
@@ -51,12 +51,7 @@
 (spec/def :message.content/command-path (spec/tuple string? (spec/coll-of (spec/or :scope keyword? :chat-id string?) :kind set? :min-count 1)))
 (spec/def :message.content/params (spec/map-of keyword? any?))
 
-(spec/def ::content (spec/conformer (fn [content]
-                                      (cond
-                                        (string? content) {:text content}
-                                        (map? content) content
-                                        :else :clojure.spec/invalid))))
-(spec/def ::content-type #{"text/plain" "command"})
+(spec/def ::content-type #{"text/plain" "command" "command-request"})
 (spec/def ::message-type #{:group-user-message :public-group-user-message :user-message})
 (spec/def ::clock-value (spec/nilable pos-int?))
 (spec/def ::timestamp (spec/nilable pos-int?))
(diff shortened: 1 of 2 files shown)


janherich (approved):

src/status_im/transport/db.cljs
```diff
@@ -51,12 +51,7 @@
 (spec/def :message.content/command-path (spec/tuple string? (spec/coll-of (spec/or :scope keyword? :chat-id string?) :kind set? :min-count 1)))
 (spec/def :message.content/params (spec/map-of keyword? any?))
 
-(spec/def ::content (spec/conformer (fn [content]
```
yenda, line 54: this is unused

src/status_im/transport/db.cljs
```diff
@@ -51,12 +51,7 @@
...
-                                      (cond
-                                        (string? content) {:text content}
-                                        (map? content) content
-                                        :else :clojure.spec/invalid))))
-(spec/def ::content-type #{"text/plain" "command"})
```
yenda, line 59: legacy "command-request" was missing, so validation was filtering incoming request commands

src/status_im/transport/db.cljs
```diff
@@ -91,6 +86,10 @@
   (spec/merge :message/message-common
               (spec/keys :req-un [:message.command/content])))
 
+(defmethod content-type "command-request" [_]
```
yenda, line 89: when we receive a legacy content-type "command-request", we validate it the same as content-type "command"

src/status_im/transport/message/v1/protocol.cljs
```diff
@@ -111,7 +111,7 @@
              :js-obj     (:js-obj cofx))]})
   (validate [this]
     (when (spec/valid? :message/message this)
-      (spec/conform :message/message this))))
+      this)))
```
yenda, line 114: do not conform the message, it is already valid and conforming would introduce weird spec specific constructs.

jakubgs (approved):
I'm don't fully understand what is happening here, but I get that it's supposed to undo some of the changes done in https://github.com/status-im/status-react/commit/81b2afebc127deb3a6bef27091f13ea0fdcca6e8 that introduced a bug. So I'm approving it. Worst case it won't fix what is already broken.

PR #6141, opened 2018-10-04
[#5985] Add animation for sending and receiving messages in a chat scene

### LINKED ISSUE #5985: Add animation to sent messages

### User Story

As a user, I want to see sent messages animate so that I can distinguish incoming messages from old messages at a glance and draw my attention to new messages.

### Description

*Type*: Feature

*Summary*: Messages in chat should gently animate in after sending, to give a more pleasant and conversation-like chat experience than the harsh binary way they're introduced now.

#### Expected behavior
Every new message in chat is animated in, growing in scale and opacity before appearing fully.

#### Actual behavior
Messages appear without any animation, which is a noticeable UI downgrade from other chat apps.

### Solution

screen recording framer

• Once sent, a message begins at 0 opacity and 0.25 scale
• The message then animates to 1 opacity and 1 scale, from its lower right corner - X axis origin: 1, y axis: 1 (blue messages sent by me) or lower left corner - X axis origin: 0, Y axis: 1 (white messages, sent by others)
• Animation should follow an easeOutQuart bezier curve with values (0.165, 0.84, 0.44, 1), and a duration of 400ms

A prototype to play with the animation https://framer.cloud/xAIfg

thanks!

<blockquote><div><strong><a href="https://framer.cloud/xAIfg">Status Send Messages</a></strong></div><div>This project was made with Framer, a design tool for turning original concepts into working prototypes.</div></blockquote>

wolfgangshilei:
fixes #5985 

### Summary:

The design specification requires that the message body's animation begins from the bottom left/right corner.
The challenge is that scaling in RN is always anchored at the center([0.5, 0.5]) of a view, thus in order to achieve the desired effect, simultaneous translations along both the X and Y axes are required. To be more specific, alongside scaling, we need an initial translation of the view, and as the animation progresses, transform the translations along both axes to zero.

The equations to calculate the initial translations are easy to deduce with a little math. Take the outgoing message's translateX (animate from bottom right) for example, the equation is: 

> translateX = (width - width * initial_scale) / 2  + padding_right * initial_scale - padding_right

in which
width = the layout width of the animated view, 
padding_right = the padding between the text and the animated view,  
initial_scale = the initial scale of the animated view.

Similar equations can be deduced for translateY and incoming message's translateX.

### Review notes (optional):

1. In order to obtain the width/height of the view in the above recipe, I use RN's onLayout handler on the animated view and reset its parent component's state in the handler. This approach can obtain the accurate layout of the animated view in a robust way, however it pays the cost of an extra rendering of the animated view, and as a result incurs a short delay before the animation starts. 
In spite of the trade-offs, this is the best solution I can think of given my current knowledge. But I would glad to be informed should you have better or different solutions; and we could make discussion should there be any alternative implementation.

### Steps to test:
- Open Status
- Step 2: Login and click + in home page, choose a chat with new messages.
- Step 3: Verify that all the messages displayed without animation.
- Step 4: Send a message to this chat, and verify that the message is animated and the animation begins at the bottom right corner.
- Step 5: Send a message to this chat from another device, and verify that the most recent incoming message is animated and the animation begins at the bottom left corner.
- Step 6: Navigate back to Home page, choose "Join public chat", and repeat Step 4 - Step 5.
- Step 7: Navigate back to Home page, click + and choose "Start new chat", enter a contact code and repeat Step 4 - Step 5.

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -31,8 +31,8 @@
 (defn- prepare-message
   [{:keys [content] :as message} chat-id current-chat?]
   ;; TODO janherich: enable the animations again once we can do them more efficiently
-  (cond-> (assoc message :appearing? true)
-    (not current-chat?) (assoc :appearing? false)
+  (cond-> (assoc message :appearing? true :animate? true)
+    (not current-chat?) (assoc :appearing? false :animate? false)
     (emoji-only-content? content) (assoc :content-type constants/content-type-emoji)))
 
 (fx/defn re-index-message-groups
@@ -303,6 +303,10 @@
                                                        :show?       true)
                                                 (add-message-type chat))]
     (upsert-and-send cofx message-data)))
+
+(fx/defn animation-finished
+  [{:keys [db]} chat-id message-id]
+  {:db (assoc-in db [:chats chat-id :messages message-id :animate?] false)})
 
 ;; effects
 
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -347,7 +347,31 @@
    [react/view (style/delivery-status outgoing)
     [message-delivery-status message]]])
 
-(defn chat-message [{:keys [message-id outgoing group-chat modal? current-public-key content-type content] :as message}]
+(defview animated-message-body [{:keys [chat-id message-id] :as message} content]
+  (letsubs [anim-range   (vector 0 1)
+            anim         (animation/create-value (first anim-range))
+            view-size    (reagent/atom nil)]
+    {:component-did-update (fn [_]
+                             (when @view-size
+                               (animation/start
+                                (animation/timing
+                                 anim
+                                 {:toValue         (second anim-range)
+                                  :easing          (.bezier react/easing 0.165 0.84 0.44 1)
+                                  :duration        400
+                                  :useNativeDriver true})
+                                #(re-frame/dispatch [:message/animation-finished-callback chat-id message-id]))))}
+    (let [on-layout-handler (fn [event]
+                              (when-not @view-size
+                                (let [layout (-> event .-nativeEvent .-layout)
+                                      height (.-height layout)
+                                      width  (.-width layout)]
+                                  (reset! view-size {:height height :width width}))))]
+      [react/animated-view {:on-layout on-layout-handler
+                            :style (style/animated-message-body anim anim-range message @view-size)}
+       [message-body message content]])))
+
+(defn chat-message [{:keys [message-id outgoing group-chat modal? current-public-key content-type content animate?] :as message}]
   [react/view
    [react/touchable-highlight {:on-press      (fn [_]
                                                 (re-frame/dispatch [:chat.ui/set-chat-ui-props {:messages-focused? true}])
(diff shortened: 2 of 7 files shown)


src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -356,17 +356,42 @@
...
                                                  (list-selection/chat-message message-id (:text content) (i18n/label :t/message)))}
     [react/view {:accessibility-label :chat-item}
-     (let [incoming-group (and group-chat (not outgoing))]
-       [message-content message-body (merge message
-                                            {:current-public-key current-public-key
-                                             :group-chat         group-chat
-                                             :modal?             modal?
-                                             :incoming-group     incoming-group})])]]])
+     (let [incoming-group (and group-chat (not outgoing))
+           animated?      (and last? (> timestamp current-chat-start-time))]
```
janherich: Is not a good idea to rely on comparison of relative time-stamps in distributed systems, could be out of sync due to network time drift, offline inboxed messages, etc.
Better then maintaining last time chat was opened and comparing it with timestamp of message sent to determine it's a new last message, it's better to just set some flag in message data indicating if it was already animated or not.
Upon completing the animation for particular message, the flag would be simply set by event like `:chat.callback/appear-animation-finished`.

janherich (changes requested):

src/status_im/ui/screens/chat/message/message.cljs
wolfgangshilei: Thanks for the comment. I will look into the technical solution you suggested.

wolfgangshilei: Hi, I have made the changes based on your technical proposal. Please let me know if there is any problem concerning the changes. Thanks.

src/status_im/chat/models/message.cljs
```diff
@@ -31,8 +31,8 @@
 (defn- prepare-message
   [{:keys [content] :as message} chat-id current-chat?]
   ;; TODO janherich: enable the animations again once we can do them more efficiently
-  (cond-> (assoc message :appearing? true)
-    (not current-chat?) (assoc :appearing? false)
+  (cond-> (assoc message :appearing? true :animate? true)
+    (not current-chat?) (assoc :appearing? false :animate? false)
```
janherich, line 35: Minor point, `:appearing?` and `:animate?` are actually the same things, we should use one or another. But that's a minor point can be done later.

janherich (approved):

src/status_im/chat/models/message.cljs
wolfgangshilei: I considered to utilize the flag `:appearing?`, however in the animation case, we need to switch the flag once the animation is done, which seems to be not the case for `:appearing?` today. 
Actually, I am not quite sure about the original purpose of `:appearing?`, so I decided to get around that flag by introducing a new one.

janherich: `:appearing?` was used for exactly the same purpose as `:animate?` is now.
Previously, we did have an message appearing animation, but we decided to remove it (but we left the flag) because of the performance problems.

janherich:
Note for the test team - please test the PR with regard to performance as well, especially in situations like opening a busy chat and getting many message from offline inbox.

src/status_im/chat/models/message.cljs
mandrigin: this animation should be much less performance penalizing, because of the native driver, but worth checking that on an old Android anyways

janherich: @mandrigin That's true, but still, in situation where you open a busy chat, and you get many messages in span of very short time (offline inboxing), it could be problematic when all of them are animated -> previously we had a problem that the application (already far stretched by processing the messages) became totally unresponsive for extended periods of time (15s+), so that's something we definitely have to check. 
In case it's still a problem we need to implement more advanced time buffering, where only the last received message in span of some time (to be tuned) buffer is actually animated.

oskarth:
@wolfgangshilei Sorry about the lack of response!

This PR has been left in CONTRIBUTOR mode. Based on the last comment it seems like this should have been handed over to the QA team but it never was.

@mandrigin care to change your review to either approve or reject with so we can decide what to do with this PR?

ghost:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

mandrigin:
@oskarth we need to test how it works and have some mechanics to not schedule too many animations at once. Sorry about delays.

PR #6142, opened 2018-10-04
Add leave & modify group members

cammellos:
### Summary:

 
Adds:
1) Leaving group chats
2) Modify group chat membership

For a rationale on the decision for the protocol and an explanation: 

https://discuss.status.im/t/group-chats-high-level-protocol/524/13

#### What is still not implemented

1) System messages

### Review notes:

I have removed the various `group-chats` and moved the logic in `group-chats.core`.
We only have plural commands for addition, as only those are useful in case we want to compact events down the line.

### Testing notes:

Desktop views are not polished but should work in the same way as mobile.
Mobile will not be enabled in nightlies, but could you please test it as it was, I will enable it soon in nightlies only in dev-mode.

status: ready


src/status_im/group_chats/core.cljs
```diff
@@ -39,103 +76,66 @@
   (and (string/ends-with? chat-id admin)
        (not= chat-id admin)))
 
+(defn valid-event?
+  "Check if event can be applied to current group"
+  [{:keys [admins contacts]} {:keys [chat-id from member members] :as new-event}]
+  (when from
+    (case (:type new-event)
```
janherich: Minor point, but when comparing value to compile time constants (like hardcoded strings here), it's faster and more efficient to use `condp` instead of `cond`, so it would be:
```clj
(condp = (:type new-event)
  "chat-created" ...)
```

janherich (approved):

src/status_im/group_chats/core.cljs
cammellos: that's interesting, I would have expected the opposite, I will change it as they are both very readable, thanks!

dmitryn (approved):

lukaszfryc:
@cammellos I noticed few issues starting from high sev:

1) When couple of messages are send quickly one after another, only some of them are delivered. This happen only for group chats, 1-1 and public chats work fine. I'm not sure if it's related to this PR or group chats in general (https://github.com/status-im/status-react/pull/6038). See messages sent from first user (top of the screenshot):



2) There is an empty screen when I click "Start group chat" but I do not have any contact:



3) On desktop, there is an empty screen when clicked on myself in chat member list. On Android I go to a contact profile screen with “Add to contacts” option. I suggest to disable clicking on yourself.

4) On desktop, I can't edit chat name

5) On desktop, there are no checkboxes while creating group chat. I don't know if a contact was selected if I clicked on it:



6) As a chat member, I don't see an option to leave the chat and never receive messages from it. Is it supposed to be added in a separte PR?

7) When new member is added or removed, app should return to the group info screen instead of the chat screen

cammellos:
@lukaszfryc thanks for reporting:
1) I will definitely fix
2, 4, 5, 7 I will possibly leave for a later PR if that's ok with you, depending on how long it takes to fix the other issue
3) I will do as suggested
6) Sorry, forgot to mention, if you delete a chat you leave the group, probably we should make it clearer to the user

Thanks!

cammellos:
@lukaszfryc ready to be tested again, 1 3 and 4 should be addressed. 
Name is editable (changes are only local).
Removing chat also make the member leave unless they are the admin.
If the admin removes the chat, the next message sent to them will pop up again.
Thanks!

lukaszfryc:
1) Fixed
3) When you click on yourself, contact screen is opened with the default name. Instead, I'd disable clicking on yourself or redirect to profile (worse option in my mind). I'll create sparate GHI for it.
4) Fixed

@cammellos please merge. I will create a spearate GHI for all other issues.

cammellos:
@lukaszfryc merged, thanks!

Issue #5899, opened 2018-09-19
No 'Intrinsic gas too low' popup when sending TX with Gas Limit < 21000


Serhy:
***Type***: Bug

***Summary***: We don't show 'Intrinsic gas too low' pop-up warning when a contact sends a transaction (from 1-1 chat / Dapp / directly from Wallet) with Gas Limit < 21000. A transaction is not sent as expected, but we need to show up the pop-up (that we have in 0.9.27 release) notifying user for what's happened. 

#### Expected behavior
'Intrinsic gas too low' popup appears when sending TX with Gas Limit < 21000

#### Actual behavior
No 'Intrinsic gas too low' popup appears when sending TX with Gas Limit < 21000

### Reproduction

1) Open Status and create an account
2) Switch to Ropsten network and request test ETH (from Simple Test Dapp)
3) Add a contact to initiate 1-1 chat with it
4) Navigate to 1-1 chat and send a `/send ETH 0` message
5) Once Wallet screen is opened, change the Gas Limit value to `20999` in transaction fee screen
6) Sign this transaction with valid password
**Actual result:** No pop-up with the 'Intrinsic gas too low' message.

7) Navigate to `Wallet` -> `Send tranaction`:
8) Specify recipient, set 0 value in `Amount` field
9) Change the Gas Limit value to `20999` in transaction fee screen
10) Sign this transaction with valid password
**Actual result:** No pop-up with the 'Intrinsic gas too low' message.

11) Open `Simple Test Dapp` -> `Assets` -> tap `Request STT`
12) Once Wallet screen is opened, change the Gas Limit value to `20999` in transaction fee screen
13) Sign this transaction with valid password
**Actual result:** No pop-up with the 'Intrinsic gas too low' message.

### Additional Information
* Status version: Develop 0.9.28 (8561); node 0.14.1
* Operating System: Android and iOS

#### Logs

TF session with logs: https://app.testfairy.com/projects/4803622-status/builds/8597460/sessions/4401830477/?accessToken=FrUZwjmQiDPtcfCCrE5P99qGPP8

<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803622-status/builds/8597460/sessions/4401830477/?accessToken=FrUZwjmQiDPtcfCCrE5P99qGPP8">TestFairy: Status 0.9.28 (8561) @2018</a></strong></div></blockquote>

lukaszfryc:
[...]

PR #6143, opened 2018-10-04
[Fix] python 35 string formatting 

hetvart:
### Summary:
Just a fix to support python v.3.5 string formatting

### Review notes:
`f'{var}'` style became available since python 3.6 version. so it will fail on evns where the version is lower.


antdanchenko (approved):

yevh-berdnyk (approved):

Serhy:
Ready for merge. From autotests: test_public_chat_messaging - known issue #5899, test_public_chat_messaging changed feature (username formats now include three default words).

Issue #1762, opened 2017-09-06
Wallet - Use web3 to fetch wallet transactions


yenda:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story
As a developper, I want to to use web3 to fetch transactions so that Status does not depends on a centralized api (etherscan) for wallet transactions history.

### Description
[comment]: # (Feature or Bug? i.e Type: Bug)
_Type_: Feature

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
_Summary_: 
To reach MVP of the new wallet we used etherscan to fetch the transaction history of the wallet. This task will consist of implementing a new version of the transaction fetching logic using web3.

#### Expected behavior
[comment]: # (Describe what you expected to happen.)
Transactions are fetched using web3

#### Actual behavior
[comment]: # (Describe what actually happened.)
Transactions are fetched using etherscan and are limited to the 20 last transactions

### Solution
[comment]: # (Please summarise the solution and provide a task list on what needs to be fixed.)
_Summary_: 

* [ ] implement and adapt this script in clojurescript https://ethereum.stackexchange.com/questions/2531/common-useful-javascript-snippets-for-geth/3478#3478 to get transactions for a given address from block X to Z in utils/transactions.cljs
* [ ] create an event that recursively calls the function for block n to n-step until the block when the wallet has been created is reached in ui.screens.wallet.events

Depending on the performance cost of this solution it might be worth it to keep the etherscan api call for the 20 first transactions, but use web3 as fallback in case it fails and fetch the rest of the transactions with web3.

status-open-bounty:
![Contract deploying](https://openbounty.status.im/img/deploying_contract.png)
Transaction: [0x4773f73f230fd3134bb2dc2be9d59063b8dad85a98cdb177d4531a8b1ec37bf8](https://etherscan.io/tx/0x4773f73f230fd3134bb2dc2be9d59063b8dad85a98cdb177d4531a8b1ec37bf8)

asemiankevich:
does this impact ERC tokens history? @goranjovic 
also @yenda why we want to have it only 20 txns?

yenda:
@asemiankevich the 20 first transaction would be fetched from etherscan if performances are not good enough in order to speed things up, while still being able to show them if etherscan is down.

debragail:
Are perf tools a thing yet?  Last I checked no one had a toolkit they were using.

[...]

PR #6144, opened 2018-10-04
[WIP] experiment getTransaction from web3 (#1762)

tbenr:
This is a PR to show some experiments I did around #1762 

it currently:
- get current block and recursively search for 50 blocks any transaction related to the current account
- prints some stats on stdout. In my iOS simulator take 7.6 seconds to finish, Rinkeby testnet.

as suspected it is very slow. I don't think even a multithreaded solution gathering blocks in parallel can reach an acceptable performance to search for a meaningful time frame in the past (days?).

The hybrid solution to still get transactions from etherscan could be enhanced by adding a lookup on web3 to check them, but txs not returned by etherscan will never be found.

Hope it helps.
Any suggestion on if\how to procede is welcome!

### DIFF
--- a/src/status_im/ui/screens/wallet/events.cljs
+++ b/src/status_im/ui/screens/wallet/events.cljs
@@ -64,7 +64,8 @@
 (re-frame/reg-fx
  :get-transactions
  (fn [{:keys [web3 chain account-id token-addresses success-event error-event]}]
-   (transactions/get-transactions chain
+   (transactions/get-transactions web3
+                                  chain
                                   account-id
                                   #(re-frame/dispatch [success-event % account-id])
                                   #(re-frame/dispatch [error-event %]))
--- a/src/status_im/utils/ethereum/core.cljs
+++ b/src/status_im/utils/ethereum/core.cljs
@@ -139,6 +139,12 @@
                        (cb result)
                        (handle-error error)))))
 
+(defn get-block-info-with-txs [web3 number cb]
+  (.getBlock (.-eth web3) number :true (fn [error result]
+                                         (if-not error
+                                           (cb (js->clj result :keywordize-keys true))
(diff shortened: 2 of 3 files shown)


oskarth:
Sorry about no one getting back to you earlier @tbenr! Is it still WIP?

Requested review based on suggestions.

jeluard:
Interesting experiment! I guess we don't want to merge it as is given the performance cost?
Could be something that we can activate based on user settings. This way users could choose not to rely on etherscan.

tbenr:
@oskarth this is currently on-hold but my intention is to continue experimenting as suggested by @jeluard

debragail:
still curious if there are any perftools to check performance for this?

StatusSceptre:
Automation has closed this but it seems like a worthy venture still for a user option. Just seeing if experimentation still pursued here.

Swader:
I would classify security / reliability of balance checking in the following order from least to most secure:

- Etherscan API
- RPC to a node (hardcoded)
- RPC to a node (user provided)
- LES
- Reading and verifying the chain

My suggestion to avoid the unsafe and centralized Etherscan is to add an option to connect Web3-style to a running, synced node. Status can provide a hardcoded whitelist (infura, Status' own node, I'll donate one at Bitfalls HQ and another at Audithor.io HQ) and users can select from those, but user should also be able to input their own if they're running a node. This is until we solve the light client problem.

In other words, let Status' web3 instance connect to custom providers.

oskarth:
Closed PR as inactive. Please reopen if this is incorrect. (not you StatusSceptre, whoever you are)

StatusSceptre:
@oskart see comment here 

https://github.com/status-im/status-react/issues/1762#issuecomment-458851543

Will let @tbenr reopen when he’s got more work done on it.

PR #6148, opened 2018-10-04
Move all the legacy handling into transit

janherich:

This PR moves all the legacy message handling into `transit` namespace, so the outermost "edge" of `status-react`.
Doing so has several benefits:
* All the legacy handling (sending&receiving) is in one place and easy to inspect
* Inside logic of the application can be now bit simpler and easier to reason about
* Less potential "traps" for anybody working on the message/command-message logic
* We can finally almost completely delete the legacy `content-type-command-request` from the project, keeping its use just for the legacy code in the `transit` namespace.

Please take your patience to review the code changes and do it with scrutiny.

### Testing notes (optional):
*  Please do all your testing with special test public/1-1 channels first, then proceed to make sure that known public channel like `#status` can be displayed correctly on the PR build and only post messages from the PR build there as a last thing if everything worked fine previously. Doing so will minimise the danger of breaking public communications with protocol changes. It should be the case for all protocol touching PRs from now and will be better described and documented in `Protocol compatibility practises ADR` (coming soon)
* Test that account with previous `/send` and `/request` message is OK upon upgrading to this PR build and all command messages are displayed correctly
* Test that communication between this PR and any other builds (including the oldest build you can get your hands on, providing it's newer then `0.9.22`) is fine, with special regard to command messages

status: ready

### DIFF
--- a/src/status_im/transport/message/transit.cljs
+++ b/src/status_im/transport/message/transit.cljs
@@ -44,14 +44,36 @@
   (rep [this {:keys [name profile-image address fcm-token]}]
     #js [name profile-image address fcm-token]))
 
+;; It's necessary to support old clients understanding only older, verbose command content (`release/0.9.25` and older)
+(defn- new->legacy-command-data [{:keys [command-path params] :as content}]
+  (get {["send" #{:personal-chats}]    [{:command-ref ["transactor" :command 83 "send"]
+                                         :command "send"
+                                         :bot "transactor"
+                                         :command-scope-bitmask 83}
+                                        constants/content-type-command]
+        ["request" #{:personal-chats}] [{:command-ref ["transactor" :command 83 "request"]
+                                         :request-command-ref ["transactor" :command 83 "send"]
+                                         :command "request"
+                                         :request-command "send"
+                                         :bot "transactor"
+                                         :command-scope-bitmask 83
+                                         :prefill [(get params :asset)
+                                                   (get params :amount)]}
+                                        constants/content-type-command-request]}
+       command-path))
+
 (deftype MessageHandler []
   Object
   (tag [this v] "c4")
   (rep [this {:keys [content content-type message-type clock-value timestamp]}]
-    (if (= content-type constants/text-content-type)
-      ;; append new content add the end, still pass content the old way at the old index
+    (condp = content-type
+      constants/content-type-text ;; append new content add the end, still pass content the old way at the old index
       #js [(:text content) content-type message-type clock-value timestamp content]
-      #js [content content-type message-type clock-value timestamp content])))
+      constants/content-type-command ;; handle command compatibility issues
+      (let [[legacy-content legacy-content-type] (new->legacy-command-data content)]
+        #js [(merge content legacy-content) (or legacy-content-type content-type) message-type clock-value timestamp])
+      ;; no need for legacy conversions for rest of the content types
+      #js [content content-type message-type clock-value timestamp])))
 
 (deftype MessagesSeenHandler []
   Object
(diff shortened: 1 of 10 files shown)


ghost:
### Pull Request Checklist

src/status_im/transport/message/transit.cljs
```diff
@@ -44,14 +44,36 @@
...
 (deftype MessageHandler []
   Object
   (tag [this v] "c4")
   (rep [this {:keys [content content-type message-type clock-value timestamp]}]
-    (if (= content-type constants/text-content-type)
-      ;; append new content add the end, still pass content the old way at the old index
+    (condp = content-type
+      constants/text-content-type ;; append new content add the end, still pass content the old way at the old index
       #js [(:text content) content-type message-type clock-value timestamp content]
-      #js [content content-type message-type clock-value timestamp content])))
```
janherich, line 54: @cammellos I know we had a discussion about this last time (always appending content at the end, even when not strictly necessary in command type messages), but I decided to remove it for the following reason, after discussion with @yenda :
1. As I said, most if not all the modification will be done in the content map itself, so no positional shuffling anymore (or very unlikely)
2. We want to avoid repeated content in message record whenever possible, because we fought hard to limit our bandwidth consumption and we don't want to reverse it for reasons like developer convenience - of course this is highly subjective (which aspect is more important) and I'm happy to follow team decision in this regard, currently it's 2:1 for no duplicate payload (@janherich & @yenda vs @cammellos).

src/status_im/transport/message/transit.cljs
```diff
@@ -44,14 +44,36 @@
   (rep [this {:keys [name profile-image address fcm-token]}]
     #js [name profile-image address fcm-token]))
 
+;; It's necessary to support old clients understanding only older, needlesy complex and verbose command content
```
cammellos: Thanks for letting us know you didn't like the previous implementation, but it would be more useful to add which version was the last one to run that code, so we can ditch the code once we are sure that no clients are using it :)

src/status_im/transport/message/transit.cljs
cammellos: @janherich As mentioned in the dm, I don't mind, I don't fully agree, but it's the team decision and I respect that

cammellos (approved):

src/status_im/transport/message/transit.cljs
janherich: Sure, I will add it to the comment, but I don't think it's particularly important as it's version newer then `0.9.20`, so we have to support it (indefinitely ?) according to last team discussion about version compatibility.
Neither is my rant, but you would have to see the old payload to understand my frustration :))

cammellos: I'd rather not see it if it prompted such a bad reaction :) 
I would still add the version number, as it is also for external devs, who might not care about supporting previous versions, and also you never know if we are going to revisit those policies at some point ;)

janherich: True, I removed the unnecessary emotions and added facts (exact version where command `content` was upgraded to newer, sparser version)

src/status_im/transport/message/transit.cljs
yenda: @janherich @cammellos I was in favor of no duplication for the fix but now there is already a few versions that expect the text to be in content. If they are not able to properly read the messages received that would violate the forward compatibility policy

src/status_im/transport/message/transit.cljs
```diff
@@ -44,14 +44,36 @@
...
+       command-path))
+
 (deftype MessageHandler []
   Object
   (tag [this v] "c4")
   (rep [this {:keys [content content-type message-type clock-value timestamp]}]
-    (if (= content-type constants/text-content-type)
-      ;; append new content add the end, still pass content the old way at the old index
+    (condp = content-type
+      constants/text-content-type ;; append new content add the end, still pass content the old way at the old index
```
yenda: please make a choice here either text-content-type or content-type-text to match content-type-command or command-content-type. Anyway personally I would prefer namespaced keywords but that would be for another PR

src/status_im/transport/message/transit.cljs
janherich: @yenda The question is not about the messages of content type `text` (duplication necessary there at this time-point) but about `command` content-type, where we don't have to append content as a last item of array (as the "old" content at zero index was always a map a not interpretation change is needed there), but making it so makes the messages more regular because we can be sure that from now on, valid message content is always at the last index, opposed to:
* old style (string) content as a first and new style (map) content as a last items for the text messages
* map style content as a first item for the command messages

src/status_im/transport/message/transit.cljs
janherich: Ok, I will update it.

janherich:
@yenda Please review again, I already changed constant `text-content-type` to `content-type-text` so it follows the same patter as rest of the content types

yenda (approved):
I am ok with it for now but I think it would be better if compatibility was in it's own namespace in the future, we could take inspiration from realm migration layout and be able to easily figure out for which version each compatibility is for.

yenda:
Things to test: compatibility with older version of status is still working, especially for commands

rasom (approved):

lukaszfryc:
> Please do all your testing with special test public/1-1 channels first

@janherich Should there be some specific messages inside those chats? Like empty messages that caused issues in https://github.com/status-im/status-react/issues/6087. Or it's only about testing the PR on some non-visited chats?

janherich:
@lukaszfryc yes, ideally, such messages should be tested as well. However the PR is simply erasing tech debt, so only regressions are relevant.

lukaszfryc:
@janherich I tested 1-1 and public chats (chatting and transacting with 0.9.23, .24 and .29) and I have not found any issues. I also did upgrade tests. Please merge 👍

PR #6158, opened 2018-10-05
Don't start a node while the node is already being started.

mandrigin:
Restart it instead, so that we are sure which config is there.

Hopefully, that will fix the annoying "Connecting to peers" issues.

status: ready

### DIFF
--- a/src/status_im/node/core.cljs
+++ b/src/status_im/node/core.cljs
@@ -127,18 +127,24 @@
                       (get-node-config db network))
         node-config-json (types/clj->json node-config)]
     (log/info "Node config: " node-config-json)
-    {:db         (assoc db :network network)
+    {:db        (assoc db
+                       :network network
+                       :node/status :starting)
      :node/start node-config-json}))
 
-(defn restart
-  []
-  {:node/stop nil})
+(defn stop
+  [{:keys [db]}]
+  {:db        (assoc db :node/status :stopping)
+   :node/stop nil})
 
 (fx/defn initialize
-  [{{:keys [status-node-started?] :as db} :db :as cofx} address]
-  (if (not status-node-started?)
-    (start cofx address)
-    (restart)))
+  [{{:node/keys [status] :as db} :db :as cofx} address]
+  (let [restart {:db (assoc db :node/restart? true :node/address address)}]
+    (case status
+      :started (stop cofx)
+      :starting restart
+      :stopping restart
+      (start cofx address))))
 
 (re-frame/reg-fx
  :node/start
--- a/src/status_im/signals/core.cljs
+++ b/src/status_im/signals/core.cljs
@@ -11,15 +11,25 @@
 
 (fx/defn status-node-started
   [{db :db :as cofx}]
-  (fx/merge cofx
-            {:db (assoc db :status-node-started? true)}
-            #(when (:password (accounts.db/credentials cofx))
-               (accounts.login/login %))))
+  (let [{:node/keys [restart? address]} db
+        can-login? (and (not restart?)
+                        (:password (accounts.db/credentials cofx)))]
+    (fx/merge cofx
+              {:db (-> db
+                       (assoc :node/status :started)
+                       (dissoc :node/restart? :node/address))}
+
+              (when restart?
+                (node/initialize address))
+              (when can-login?
+                (accounts.login/login)))))
 
 (fx/defn status-node-stopped
-  [cofx]
+  [{db :db :as cofx}]
   (let [{:keys [address]} (accounts.db/credentials cofx)]
-    (node/start cofx address)))
+    (fx/merge cofx
+              {:db (assoc db :node/status :stopped)}
+              (node/start address))))
 
 (fx/defn status-module-initialized [{:keys [db]}]
   {:db                             (assoc db :status-module-initialized? true)
--- a/src/status_im/ui/screens/db.cljs
+++ b/src/status_im/ui/screens/db.cljs
@@ -81,7 +81,10 @@
 ;;object?
 (spec/def ::webview-bridge (spec/nilable any?))
 (spec/def ::status-module-initialized? (spec/nilable boolean?))
-(spec/def ::status-node-started? (spec/nilable boolean?))
+(spec/def :node/status (spec/nilable #{:stopped :starting :started :stopping}))
+(spec/def :node/node-restart? (spec/nilable boolean?))
+(spec/def :node/address (spec/nilable string?))
+
 ;;height of native keyboard if shown
 (spec/def ::keyboard-height (spec/nilable number?))
 (spec/def ::keyboard-max-height (spec/nilable number?))
@@ -218,6 +221,9 @@
                  :inbox/wnodes
                  :inbox/current-id
                  :inbox/fetching?
+                 :node/status
+                 :node/restart?
+                 :node/address
                  :universal-links/url
                  :push-notifications/stored
                  :browser/browsers
@@ -247,7 +253,6 @@
                  ::web3-node-version
                  ::webview-bridge
                  ::status-module-initialized?
-                 ::status-node-started?
                  ::keyboard-height
                  ::keyboard-max-height
                  ::tab-bar-visible?
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -8,6 +8,7 @@
             [status-im.accounts.update.core :as accounts.update]
             [status-im.bootnodes.core :as bootnodes]
             [status-im.browser.core :as browser]
+            [status-im.node.core :as node]
             [status-im.browser.permissions :as browser.permissions]
             [status-im.chat.commands.input :as commands.input]
             [status-im.chat.models :as chat]
--- a/src/status_im/init/core.cljs
+++ b/src/status_im/init/core.cljs
@@ -61,8 +61,9 @@
 
 (fx/defn initialize-app-db
   "Initialize db to initial state"
-  [{{:keys [status-module-initialized? status-node-started? view-id
+  [{{:keys [status-module-initialized? view-id
             network-status network peers-count peers-summary device-UUID]
+     :node/keys [status]
      :or   {network (get app-db :network)}} :db}]
   {:db (assoc app-db
               :contacts/contacts {}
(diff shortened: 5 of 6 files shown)


ghost:
### Pull Request Checklist

src/status_im/node/core.cljs
```diff
@@ -124,18 +124,20 @@
                       (get-node-config db network))
         node-config-json (types/clj->json node-config)]
     (log/info "Node config: " node-config-json)
-    {:db         (assoc db :network network)
+    {:db         (assoc db :network           network
+                        :status-node-busy? true)
```
yenda: could we keep the semantics of signals ? If I understand correctly we receive `node.ready` when the node is done starting ? So it would be better to us `status-node-started?` in the code when the node has been started and `status-node-ready?` when the node is now ready instead of busy? while it is started.
wdyt ?

mandrigin: one thing here is that `status-node-started?` in that case won't match `node.started` signal, because we want to set one of the booleans to be a guard showing that the node is int the process of starting (set it to `true` with calling `start-node`).
`node.started` from status-go comes much later.

maybe the best option is to have a keyword enum `:stopped` / `:busy` / `:started` and keep one variable...

annadanchenko:
@mandrigin I've used iOS app from build 1 to reproduce issue with "Connecting to peers" by logging out and in quickly. Currently I can't reproduce it with this build but for  less than a sec can see "unknown status-go error" message above password field when automatic login is in progress. Here is log (contains all attempts to log in-log out) 
[geth.log](https://github.com/status-im/status-react/files/2450201/geth.log)

mandrigin:
@annadanchenko thanks! I will take a look.
Is account creation works fine both after logout and on a fresh start? (that might be an area of regressions)

src/status_im/node/core.cljs
yenda: I'm ok with an enum, having a state machine is less error prone and gives a clear understanding of what transitions are acceptable or not.

As I understand we have `:stopped` `:starting` (I prefer that over `:busy`) `:started`
From there it will be easier to determine which transition are acceptable or not

mandrigin: okay, so it can't be `:stopping`, right? because that's why I chose `:busy` because I don't care if it is stopping or starting. but if `stopNode` is instantaneous, then `:starting` it's ok.

yenda: if there is a :stopping state better to acknowledge it by setting it when stopping and changing it when it's stopped

annadanchenko:
@mandrigin 
1. can't see "unknown status-go error" anymore as it was in build 1
2. Got error about encryption problem https://s3.amazonaws.com/testfairy/screenshots/4803590/8619221/108433255/3_57a95918ed38791c00c95dde7bd50b05fdb93d01_o.jpg 


Steps:
1. create first account, log in, close app, launch app for several times with turned on "Saved password" toggle
2. create second account, log in and close app. Launch app, turn on option to Save password and log in. 
3. Close app, launch app

``` 
 Realm file decryption failed. 
```
in https://app.testfairy.com/projects/4803590-status/builds/8619221/sessions/4403400551/?accessToken=dxrqaT6O7HGPe93nygO8n-Rl0LU

[geth.log](https://github.com/status-im/status-react/files/2454890/geth.log)


<blockquote></blockquote>
<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803590-status/builds/8619221/sessions/4403400551/?accessToken=dxrqaT6O7HGPe93nygO8n-Rl0LU">TestFairy: Status 0.9.29 (9982) @2018</a></strong></div></blockquote>

mandrigin:
@annadanchenko thanks! can you check if that could be reproduced on develop? It doesn’t look related to my changes from the first glance...

annadanchenko:
@mandrigin you are right I can reproduce issue with encryption key on develop (last nightly build), so not related to this PR

mandrigin:
Great! So it helps then. @yenda I’ll make the proposed state names refactoring, but it’s ready to really be reviewed now.

src/status_im/node/core.cljs
mandrigin: done

src/status_im/node/core.cljs
```diff
@@ -124,18 +124,24 @@
...
+   :node/stop nil})
 
 (fx/defn initialize
-  [{{:keys [status-node-started?] :as db} :db :as cofx} address]
-  (if (not status-node-started?)
-    (start cofx address)
-    (restart)))
+  [{{:keys [status-node-state] :as db} :db :as cofx} address]
+  (let [reinit {:db (assoc db :dispatch-after-start [[:node.needs-reinit address]])}]
+    (cond
```
rasom:
```clojure
(case status-node-state
      :started (restart cofx)
      (:stopping :starting) reinit
      (start cofx address))
```
might be a bit shorter

src/status_im/signals/core.cljs
```diff
@@ -12,14 +12,23 @@
...
   (fx/merge cofx
-            {:db (assoc db :status-node-started? true)}
-            #(when (:password (accounts.db/credentials cofx))
+            (merge
+             {:db (assoc db
+                         :status-node-state    :started
+                         :dispatch-after-start nil)}
+             (when (:dispatch-after-start db)
+               {:dispatch-n (:dispatch-after-start db)}))
+            #(when (and (nil? (:dispatch-after-start db)) ;; node is not going
```
rasom: at some point in time we agreed to not add comments to the end of the line, but place them above form, can't find a link...

rasom (approved):

mandrigin:
@flexsurfer @yenda can you take a look too?

src/status_im/node/core.cljs
```diff
@@ -124,18 +124,24 @@
...
+      (= status-node-state :started) (restart cofx)
```
yenda: it's called `restart` but according to what the function does it is just `stop`

src/status_im/ui/screens/db.cljs
```diff
@@ -241,7 +243,9 @@
                  ::web3-node-version
                  ::webview-bridge
                  ::status-module-initialized?
-                 ::status-node-started?
+                 ::status-node-state
+                 ::status-node-busy?
```
yenda: can be removed

src/status_im/ui/screens/db.cljs
```diff
@@ -241,7 +243,9 @@
                  ::web3-node-version
                  ::webview-bridge
                  ::status-module-initialized?
-                 ::status-node-started?
+                 ::status-node-state
```
yenda: better use a namespaced keyword as it is clearly a key for the `node` module so `:node/state`

src/status_im/ui/screens/db.cljs
```diff
@@ -78,7 +78,9 @@
 ;;object?
 (spec/def ::webview-bridge (spec/nilable any?))
 (spec/def ::status-module-initialized? (spec/nilable boolean?))
-(spec/def ::status-node-started? (spec/nilable boolean?))
+(spec/def ::status-node-state (spec/nilable #{:stopped :starting :started :stopping}))
+(spec/def ::dispatch-after-start (spec/nilable vector?))
```
yenda: I would simply have `:node/restart?` and that would be nilable boolean
and `:node/address` to store the node address if it can't be retrieve somewhere else

src/status_im/node/core.cljs
```diff
@@ -124,18 +124,24 @@
...
+  {:db        (assoc db :status-node-state :stopping)
+   :node/stop nil})
 
 (fx/defn initialize
-  [{{:keys [status-node-started?] :as db} :db :as cofx} address]
-  (if (not status-node-started?)
-    (start cofx address)
-    (restart)))
+  [{{:keys [status-node-state] :as db} :db :as cofx} address]
+  (let [reinit {:db (assoc db :dispatch-after-start [[:node.needs-reinit address]])}]
```
yenda: `assoc :node/restart? true :node/address address`

src/status_im/signals/core.cljs
```diff
@@ -12,14 +12,23 @@
 (fx/defn status-node-started
   [{db :db :as cofx}]
   (fx/merge cofx
-            {:db (assoc db :status-node-started? true)}
-            #(when (:password (accounts.db/credentials cofx))
+            (merge
+             {:db (assoc db
+                         :status-node-state    :started
+                         :dispatch-after-start nil)}
```
yenda: `(-> db
      (assoc :node/state :started)
      (dissoc :node/restart? :node/address))`

src/status_im/signals/core.cljs
```diff
@@ -12,14 +12,23 @@
 (fx/defn status-node-started
   [{db :db :as cofx}]
   (fx/merge cofx
-            {:db (assoc db :status-node-started? true)}
-            #(when (:password (accounts.db/credentials cofx))
+            (merge
```
yenda: `(let [{:node/keys [restart? address]} db
        can-login? (and (not restart? (:password (accounts.db/credentials cofx)))] ...`

src/status_im/signals/core.cljs
```diff
@@ -12,14 +12,23 @@
...
+             {:db (assoc db
+                         :status-node-state    :started
+                         :dispatch-after-start nil)}
+             (when (:dispatch-after-start db)
```
yenda: `(when restart?
 (node/initialize address))
(when can-login?
 ((accounts.login/login))`

yenda (changes requested):

src/status_im/node/core.cljs
mandrigin: yeah, but we never really stop the node, as soon as it is stopped it is auto-restarted. probably the function should be renamed.

src/status_im/node/core.cljs
```diff
@@ -124,18 +124,24 @@
...
 
 (fx/defn initialize
-  [{{:keys [status-node-started?] :as db} :db :as cofx} address]
-  (if (not status-node-started?)
-    (start cofx address)
-    (restart)))
+  [{{:node/keys [node-state] :as db} :db :as cofx} address]
+  (let [reinit {:db (assoc db :node/restart? true :node/address address)}]
+    (cond
+      (= node-state :started) (restart cofx)
```
yenda: still looks like a stop to me rather than a restart

yenda (approved):

mandrigin:
After these refactorings, it would be nice to re-test everything again.

PR #6181, opened 2018-10-08
Include platforms and areas to test in PR template

lukaszfryc:
### Summary:

Update the pull request template to include platforms and areas to test. This is both for devs to think of impact of the changes and for testers to do the testing more efficiently.

### Review notes (optional):
1) Should all feature or bug PR be required to include platforms and areas to test?   
2) Should platforms and areas to test be commented by default?  

status: ready

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,8 +16,35 @@ fixes #...
 <!-- (Specify if something in particular should be looked at, or ignored, during review) -->
 
 ### Testing notes (optional):
-<!-- (Specify if something specific has to be tested, for example upgrade paths) -->
+<!-- (Specify which platforms should be tested) -->
+#### Platforms (optional)
+- Android
+- iOS
+- macOS
+- Linux
 
+<!-- (Specify if some specific areas has to be tested, for example 1-1 chats) -->
+#### Areas that maybe impacted (optional)
+**Functional**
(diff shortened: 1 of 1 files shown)


cammellos (approved):

lukaszfryc:
TBH, I'm not convinced that including this long list of suggestions in PR description is a good idea. I'd rather see a longer description of what are the risks and what should be tested creted by a developer of the PR. Check this PR as a great example: https://github.com/status-im/status-react/pull/6148#issue-220535117

@annadanchenko @cammellos what are your thoughts?

annadanchenko (approved):

lukaszfryc:
@annadanchenko @mandrigin  I'd like to know your thoughts about https://github.com/status-im/status-react/pull/6181#issuecomment-427804939 before merging this PR.

cammellos:
@lukaszfryc Maybe the list is a bit redundant, and also it might change as we add more parts of the app, but it gives in my opinion an idea of what it should be in that section, so you could just leave a couple of points and `etc` for example, either ways I don't mind it.

lukaszfryc:
@annadanchenko what do you think?

annadanchenko:
@lukaszfryc I think having the list as it's proposed now (OS, areas) is fine. It's not supposed to replace PR description but to facilitate and remind of areas in each PR. So, I'm fine with the structure and propose just to start using it and see what should be adjusted from how it's being used.

PR #6197, opened 2018-10-09
Ignore messages with too high clock-value

### LINKED ISSUE #6196: Some messages are pinned in the chat
### User Story

As a user
I don't want to see "pinned" messages
as they prevent me to communicate effectively 

### Description

A malicious user can send messages with an high `clock-value`, effectively pinning messages in the chat.

#### Expected behavior

The message is not pinned

#### Actual behavior
The message is pinned

### Reproduction
A maliciously crafted message needs to be sent on the chat.

cammellos:
fixes #6196

### Summary:

We never made sure `clock-value` is a reasonable value before storing the message in the database, therefore effectively allowing `pinning` of the messages in a chat.

Although this problem is not completely preventable (`clock-value` is theoretically boundless), we can make checks to detect malicious usage, assuming that if the initial part of the clock value (a `timestamp` in ms), is too high (currently set to `now` + `1 month`), the message is to be considered malicious.

Previously this check was only made when setting our own clock value, but the message was stored in the database.

The consequences is that if your clock is skewed +/- 1 month you either won't be receiving other people messages (-1 month), or peers won't be receiving yours (+1 month).

### Testing notes :

1) Go to status chat with a new account
2) There should be no pinned `Test` message from yesterday, nor `Does anyone here...`

For more malformed messages please let me know and we can send some to a chat.

status: ready

### DIFF
--- a/src/status_im/data_store/chats.cljs
+++ b/src/status_im/data_store/chats.cljs
@@ -3,6 +3,7 @@
             [cljs.core.async :as async]
             [re-frame.core :as re-frame]
             [status-im.utils.ethereum.core :as utils.ethereum]
+            [status-im.utils.clocks :as utils.clocks]
             [status-im.data-store.realm.core :as core]))
 
 (defn remove-empty-vals
@@ -49,15 +50,16 @@
                          :message :chat-id chat-id)
       (core/sorted :clock-value :desc)
       (core/single-clj :message)
-      :clock-value))
+      :clock-value
+      (utils.clocks/safe-timestamp)))
 
 (defn- normalize-chat [{:keys [chat-id] :as chat}]
-  (let [last-clock-value (get-last-clock-value chat-id)]
-    (-> chat
-        (update :admins   #(into #{} %))
-        (update :contacts #(into #{} %))
-        (update :membership-updates  (partial unmarshal-membership-updates chat-id))
-        (assoc :last-clock-value (or last-clock-value 0)))))
+  (-> chat
+      (update :admins   #(into #{} %))
+      (update :contacts #(into #{} %))
+      (update :membership-updates  (partial unmarshal-membership-updates chat-id))
+      ;; We cap the clock value to a safe value in case the db has been polluted
+      (assoc :last-clock-value (get-last-clock-value chat-id))))
 
 (re-frame/reg-cofx
  :data-store/all-chats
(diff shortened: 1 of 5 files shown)


janherich (approved):

src/status_im/data_store/chats.cljs
```diff
@@ -2,8 +2,18 @@
   (:require [goog.object :as object]
             [cljs.core.async :as async]
             [re-frame.core :as re-frame]
+            [status-im.utils.clocks :as utils.clocks]
             [status-im.data-store.realm.core :as core]))
 
+(defn- get-last-clock-value [chat-id]
```
flexsurfer: just wondering, is this a new function? can't see where it used

cammellos: mmh, you are right, I think something went funny with the rebase, let me check

cammellos: yes, good catch, I had extracted the fn but not used it :)

cammellos: updated!

flexsurfer (approved):

mandrigin:
@cammellos I didn't make up my mind about it, but could it be better to just replace this "unsafe" clock value to just a timestamp when you received the message? then you will still see the message, potentially out of order, but it will be out of order anyway.

cammellos:
> but could it be better to just replace this "unsafe" clock value to just a timestamp when you received the message?

I have the same feeling, the only thing is that currently message-id is calculated using all the fields of the message, which works fine because the message is immutable, if we start mutating the field then potentially the same message will have different id, which might lead to a few unintended consequences (if we receive the same message multiple times, reply to a message).

To be on the safe side I would start by ignoring those messages, at least in nightlies, as I would like to get some feedback on this, and then we can re-evaluate and see if it is worth deriving a timestamp for those messages (probably we can use the whisper timestamp ) and explore a bit more the consequences. What do you reckon?

mandrigin:
@cammellos I see... maybe we can add another field there that won't participate in message-id generation. aka, leave this field alone, but have another `safe-clock` field that has the original clock value if it is safe or the timestamp of receiving.

cammellos:
@mandrigin yes, that's an option, I would still leave this PR as it is and build on it incrementally, but I can work on that in the next, basically we can explore 2 options:

1) Replace with the whisper timestamp->clock-value, which should work as it is immutable. The consequences might be that people won't be able to reply to malformed messages if sent from a previous version.
2) Use an extra field as you described

I would lean towards 1 as it's a bit lower footprint but don't mind either

PR #6221, opened 2018-10-09
Download latest nightly in desktop

flexsurfer:
image

### DIFF
--- a/src/status_im/ui/components/desktop/events.cljs
+++ b/src/status_im/ui/components/desktop/events.cljs
@@ -4,7 +4,14 @@
 (handlers/register-handler-fx
  :show-desktop-tab
  (fn [{:keys [db] :as cofx} [_ tab-name]]
-   {:db (assoc-in db [:desktop/desktop :tab-view-id] tab-name)
-    :dispatch [:navigate-to (if (and (= tab-name :home) (:current-chat-id db))
-                              :chat
-                              :home)]}))
+   (merge {:db (assoc-in db [:desktop/desktop :tab-view-id] tab-name)
+           :dispatch [:navigate-to (if (and (= tab-name :home) (:current-chat-id db))
+                                     :chat
+                                     :home)]}
+          (when (= tab-name :profile)
+            {:http-get
+             {:url
+              "https://raw.githubusercontent.com/status-im/status-im.github.io/develop/env.sh"
+              :success-event-creator
+              (fn [o]
+                [:fetch-desktop-version-success o])}}))))
--- a/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs
+++ b/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs
@@ -105,8 +105,10 @@
 
 (views/defview profile [{:keys [seed-backed-up? mnemonic] :as user}]
   (views/letsubs [current-view-id [:get :view-id]
+                  nightly-version [:get-in [:desktop/desktop :nightly-version]]
                   editing?        [:get :my-profile/editing?]] ;; TODO janherich: refactor my-profile, unnecessary complicated structure in db (could be just `:staged-name`/`:editing?` fields in account map) and horrible way to access it woth `:get`/`:set` subs/events
-    (let [adv-settings-open?           (= current-view-id :advanced-settings)
+    (let [{:keys [url commit]}         nightly-version
+          adv-settings-open?           (= current-view-id :advanced-settings)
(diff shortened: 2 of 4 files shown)


chadyj:
Love it!

vitvly (approved):
Cool!

vkjr (approved):

yenda:
The idea is great but I don't link the fact that it is within the app and points to github. I think it could simply be a script and it could do more, like kill if a new version exists, download, rename, chmod, and open again

yenda:
This is the linux one https://gist.github.com/yenda/bd0b011c05ffd94e606f8ac0667906da could you do one for mac? 

<blockquote><div> Gist</div><div><strong><a href="https://gist.github.com/yenda/bd0b011c05ffd94e606f8ac0667906da">Script to download and start lastest status-desktop on Linux : WILL CLOSE THE APP IF UPGRADE AVAILABLE!!!!</a></strong></div><div>Script to download and start lastest status-desktop on Linux : WILL CLOSE THE APP IF UPGRADE AVAILABLE!!!! - status</div></blockquote>

churik:
@flexsurfer 
can you please provide steps how it can be tested?
When I install build from PR, obviously I cannot see button because it is last build:)

Also as I understand from https://github.com/status-im/status-react/pull/6221#issuecomment-429090167 that it is done only for Linux, is it right?

cc @SlizkovaGanna

yenda:
@churik no my comment is standalone script that downloads and runs latest nightly if an upgrade is available. it's a proposal to replace this PR as I don't like the idea of having the app pointing at github since it is a step in the wrong direction in terms of decentralization. A script is even more convenient as it is able to close/reopen app after update

churik:
@yenda thanks for explanation!
@chadyj WDYT about https://github.com/status-im/status-react/pull/6221#issuecomment-429301342?

chadyj:
> WDYT about #6221 (comment)?

Should open up the discussion cc @oskarth @arnetheduck @jarradh @mandrigin

flexsurfer:
@yenda could you elaborate what do you mean app pointing at github ? this was just a quick solution for nightlies, it's simple, just check the latest version and download it, instead of opening status-im page manually, let's merge it and use, while researching scripts

@churik you should see the button, because this PR is not rebased onto latest develop, are you trying on mac or linux?

yenda:
@flexsurfer Script for Linux already works fine as it is it's what I'm using

churik:
@flexsurfer yes, I can see it now (cannot say what was wrong last time, sorry).

That's what I see for Linux:
ub
Should it be like this?

 @chadyj 
But actually it is just button with link to latest nightly, so you still need to stop current application manually, unzip dmg file for Mac (and make it executable for Linux), and launch new nightly.
So it is not like update, it is quick reference to latest nightly and it is not replacing existing app, it is downloading newest build. 

let me know what you decided.

flexsurfer:
@churik i need to fix the small thing for Linux, there should be commit not `d-nigh`

flexsurfer:
Yes my idea is just to make it easy to understand that there is new nightly version, and instead opening url in browser and download app, you can click this button, that's it, just quick solution for nightlies , we can use both, script from Eric and this button, i don't think it's bad to use this button temporary in the app , in the future we need some decentralized automatically upgrades for sure
btw that's why i named it download, not update

flexsurfer:
@churik fixed

arnetheduck:
@yenda's script is something I would tend to use on linux if it was available (ie copy-paste a curl command that downloads a script that takes care of everything, akin to how many dev environments get set up these days), but it fails the important UX test of downloading when it's a good time to do so - at the start is not it because that's when you're anxious to start chatting. there's a few other considerations that go in there as well, linked to repro builds and coercion and the like that would require a deeper discussion than I think is appropriate for this PR.

that said, it's a bit of a PITA for most users I think to have to open up a terminal - the in-app link is more standard and adequate for now - the bigger risk of adding this feature is introducing an average / good enough solution that will discourage a dedicated effort to work on a deliberate download and upgrade strategy across the board

churik:
@chadyj  feel free to move it to  "Merge" if you decide to have it.

chadyj:
This is an interesting first step in notifying users about upgrades and looks at helping solve a problem we have with users not upgrading. We will definitely need to update it for the first desktop alpha release, and as as @arnetheduck suggested look into better long term solutions.

But in the meantime lets merge and treat this as an experiment to get feedback on the user-side of updating as well as taking the opportunity to investigate broader upgrade options.

PR #6226, opened 2018-10-09
Always load messages referenced in responses

janherich:

There was a problem with partial loading of messages (pagination), when some message referenced reply not yet loaded, resulting in blank "response to" section.
This PR fixes it in a reasonable effective way, however we should seriously think about redoing our pagination altogether and getting rid of the fields like `:not-loaded-message-ids` and (added in this PR) `:referenced-messages` in favour of always holding all messages in the memory (should be feasible, needs proof) and modelling the rendered range by by `:range [from to]` properties in `app-db`.

## Testing notes
* Test that when app is open and not all messages are loaded from db, you can still see the the referenced messages in replies. The only exception is when those referenced messages are older then 24h, eq they are never fetched from mailserver.

status: ready

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -67,10 +67,19 @@
              {:db db}
              (:chats db)))
 
+(defn- get-referenced-ids
+  "Takes map of message-id->messages and returns set of message ids which are referenced by the original messages,
+  excluding any message id, which is already in the original map"
+  [message-id->messages]
+  (into #{}
+        (comp (keep (comp :response-to :content))
+              (filter #(not (contains? message-id->messages %))))
+        (vals message-id->messages)))
+
 (fx/defn initialize-chats
   "Initialize all persisted chats on startup"
   [{:keys [db default-dapps all-stored-chats get-stored-messages get-stored-user-statuses
-           get-stored-unviewed-messages stored-message-ids] :as cofx}]
+           get-stored-unviewed-messages get-referenced-messages stored-message-ids] :as cofx}]
   (let [stored-unviewed-messages (get-stored-unviewed-messages (:current-public-key db))
         chats (reduce (fn [acc {:keys [chat-id] :as chat}]
                         (let [chat-messages (index-messages (get-stored-messages chat-id))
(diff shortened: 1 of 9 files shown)


flexsurfer:
https://github.com/status-im/status-react/issues/6234

cammellos (approved):

flexsurfer (approved):

churik:
@janherich 
For me https://github.com/status-im/status-react/issues/6234 (scenario 1) is still reproducible.
I created new user, joined to channel `#tery` which contains replies to messages older than 24h.
Still can see Unknown.
1unknown
Also some in #status channel.
status

[StatusIm.txt](https://github.com/status-im/status-react/files/2464686/StatusIm.txt)

janherich:
@churik My bad, I actually misunderstood the issue - what this PR is solving is the more immediate issue where when you open chat with many messages (in last 24h hours) and initial 20 messages are loaded, if those contains reply to older messages, they are displayed blank, till you scroll up to load all the older messages.
The issue you described (referencing messages older then 24h, where there is no easy way to get them) is harder to crack, but ultimately less severe, because it's only relevant for new users. 
I propose to merge this PR anyway, as it's solving the more pressing issue and keep #6234 open, wdyt ?

churik:
OK, will test regression on mobile and edit the issue.

janherich:
@churik I just spoke with @yenda and very soon we will fetch last 7 days worth of messages instead of 24h, which will mostly solve the original issue, as almost nobody will reply to message more then 7 days old.

churik:
@janherich yes, agree. Do you know the PR?

churik:
`Unknown` for now is displayed in 2 cases:
1) if initial message is sent > 24 h ago for new user
2) if user quotes own message from "Not sent" status (possible only in desktop)
hjsdfsdf
Will add this info to #6234.

Tested that message which was sent ~8 days ago and has reply in chat is still displayed after reopening application.
As regression tested chatting and message replies in 1-1 and public chats, sending messages from offline mode (desktop).
Tested on IOS 11.4, Mac OSx 10.13.6

janherich:
@churik It's in progress (https://github.com/status-im/status-react/pull/6223)

janherich:
@churik Awesome, thanks. The second case is tricky, it would be probably wise to just not allow quoting messages unless they are actually sent, will add it to the PR.

churik:
@janherich for now I'll wait for autotests results and send PR to merge, if everything is OK

janherich:
@churik Second case is already handled in new build (coming soon), so you are not able to quote not send messages on desktop

janherich:
@churik Anything new ? Can it be merged into develop ? I'm wary about keeping this open for unnecessary long time because of the potential merge conflict issues.

churik:
@janherich  ready to merge

PR #6264, opened 2018-10-11
[#6198]: Add QR code scanner support in extensions panel

### LINKED ISSUE #6198: Add QR code scanner support in extensions panel
### User Story

As a user, I want an easy way to scan extensions QR code in the `extensions` panel

### Description

*Type*: Feature

alwx:
Fixes #6198

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -5,7 +5,9 @@
             [pluto.storages :as storages]
             [status-im.chat.commands.core :as commands]
             [status-im.chat.commands.impl.transactions :as transactions]
-            [status-im.ui.components.react :as react]))
+            [status-im.ui.components.react :as react]
+            [status-im.ui.screens.navigation :as navigation]
+            [status-im.utils.fx :as fx]))
 
 (def components
   {'view           {:value react/view}
@@ -43,3 +45,8 @@
 (defn load-from [url f]
   (when-let [uri (url->uri url)]
     (storages/fetch uri f)))
+
+(fx/defn set-extension-url-from-qr
+  [cofx url]
+  (fx/merge (assoc-in cofx [:db :extension-url] url)
+            (navigation/navigate-back)))
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -16,6 +16,7 @@
             [status-im.chat.models.message :as chat.message]
             [status-im.contact.core :as contact]
             [status-im.data-store.core :as data-store]
+            [status-im.extensions.core :as extensions]
             [status-im.fleet.core :as fleet]
             [status-im.group-chats.core :as group-chats]
             [status-im.hardwallet.core :as hardwallet]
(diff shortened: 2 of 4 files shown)


ghost:
### Pull Request Checklist

jeluard (approved):

src/status_im/extensions/core.cljs
```diff
@@ -1,7 +1,13 @@
...
 
 (defn url->storage-details [s]
   (when s
     (let [[_ type id] (string/split s #".*[:/]([a-z]*)@(.*)")]
       [(keyword type) id])))
+
+(fx/defn set-extension-url-from-qr
+  [{:keys [db]} url]
+  {:db       (assoc db :extension-url url)
+   :dispatch [:navigate-back]})
```
janherich: Unnecessary dispatch, should be just `navigate-back` function from navigation ns.

janherich (dismissed):

src/status_im/extensions/core.cljs
alwx: I don't agree.

`:navigate-back` is the only handler we use in this case, and it also has a check for a modal dialog. The function you're talking about is a side-effect, it's a strange idea to just call it like that. 

Maybe a good way of solving this is to add an fx and do something like 
```
{:db db
 :navigation/navigate-back nil}
```
But I don't think it's something related to this issue.

flexsurfer (approved):

jeluard:
@alwx Please resolve the conflict ! :)

src/status_im/extensions/core.cljs
yenda: it does the same as :navigate-back does without an unnecessary dispatch there is no side-effect it updates the db and adds the ::navigate-back nil effect

yenda:
```
(fx/defn navigate-back
  [{{:keys [navigation-stack view-id] :as db} :db}]
  (assoc
   {::navigate-back nil}
   :db (let [[previous-view-id :as navigation-stack'] (pop navigation-stack)
             first-in-stack (first navigation-stack)]
         (if (= view-id first-in-stack)
           (-> db
               (assoc :view-id previous-view-id)
               (assoc :navigation-stack navigation-stack'))
           (assoc db :view-id first-in-stack)))))

(handlers/register-handler-fx
 :navigate-back
 (re-frame/enrich -preload-data!)
 (fn [cofx _]
   (navigate-back cofx)))
```

alwx: Then sorry, I simply looked at the wrong function

yenda (approved):

asemiankevich:
@alwx @jeluard 

what should be shown for user when i scan extension URL qr code within qr scanner from Home screen? (Home - plus button - scan QR)? 
If i do this now in this PR - nothing is shown for user, however would be nice to say something (we have some error dialog for this). WDYT? 

image

asemiankevich:
so this PR implements scanning universal links for extensions only. No validation is added (future PRs per @jeluard )

PR #6270, opened 2018-10-11
Implementation of text content parsing

janherich:

Implementation of text parsing in models, enriching text content with important actions (clickable links, tags, mentions) and styling (bold, italic) metadata.

It's not yet plugged into message receiving (`enrich-content` function which will be called every-time message is added to app-db/realm) or rendering (`build-render-tree` which will be called in chat stream subscription), but that will be very straightforward affair, for now it should have to impact to app at all, that's why `skip-manual-qa` flag.

Unit tests are fairly basic and serve more as an additional example in understanding which kind of data those 2 new public functions create, in the future, we should unleash property based testing on it, as I think it's very well suited to it (cc @gnl ).

status: ready


flexsurfer (approved):

yenda:
Have you run some benchmarks ? I still have mixed feelings about this over a vector of tuples. I assume rendering each messages means building the tree and then converting the tree to elements ? Does it brings much improvement ? Also could you provide an example on how to turn the tree into elements ? You can fake elements with tuples for that :stuck_out_tongue:

janherich:
@yenda Yes I did, it's quite performant, running everything (`enrich-content` + `build-render-tree`) on 1000 messages, where each is very complicated containing multiple tags + link where both of them are embedded inside stylings (bold and italic) is less then 500ms.
Needles to say, such scenario is not very common and when just 10% of the messages are specially styled/contain actions, it's around 50ms for 1000 messages.
It's actually faster then current regex parsing for mobile messages (and less memory consuming, because we don't have to memoize anything anymore, except standard subscription caching, which was happening anyway).

The problem with tokenising along whitespaces is that it's actually wrong and could be best demonstrated by following example:
Try to type message `Hi, there is one great resource (https://www.link.com) I have been talking about` in both mobile and desktop. One mobile the link is rendered as clickable action, on desktop it's not.

That being said, the complicated and slow mobile parsing is far from correct as well, just type the following `People *this link(https://www.link.com) is very important !*` - bold styling is taken into account, but the link is suddenly not a clickable action anymore.

When approach from this PR will be implemented, it will bring consistent and much less buggy/restricted behaviour among mobile/desktop and will be good building block from more advanced features like aggregating all mentions from messages one one device and displaying them in some sort of widget, same for tags (trending, etc.). That's why I did go for the two step approach where first representation (result of the `enrich-content`) is normalised to the highest degree (making processing and querying a breeze), while the second denormalised result (`build-render-tree`) is great for logic free rendering (just logic-less depth-first-traversal with simple translation of text + type into `[:text]` element with appropriate styling/on-click handler. 

And yes, for sure I can provide some example of it, you will see how simple is that :)

jeluard:
Have you considered using [instaparse](https://github.com/Engelberg/instaparse) for this?
Also wondering if we have a precise grammar of the syntax we intend to support.

janherich:
@yenda I just ran some additional tests and time to run `build-render-tree` 1000 times with fairly complicated metadata:
```clj
{:text
 "Typical message with maybe one #hashtag, some resources(https://link.com) and *important #channel*",
 :metadata {:link ([56 72]), 
            :tag ([31 39] [89 97]), 
            :bold ([78 98])}}
```
Is 71ms while running [your tokenizer](https://github.com/status-im/status-react/blob/develop/src/status_im/ui/screens/desktop/main/chat/views.cljs#L115-L122) on the same input 1000 times is 78ms. Needless to say, the actual rendering is faster in my case (because I don't need to iterate over every word) and your tokeniser is slower in common case of longer message with many words but no actions/links.

yenda:
@janherich ofc we shouldn't tokenize by words but for instance #6282 came with a solution that doesn't require that. Also with the tokenizer you are pretty much done and only need to build element, with your tree you still need to parse the tree again to render the elements

janherich:
@jeluard Good question, I though about it, unfortunately Instaparse is very, very slow (because it's so powerful and generic), so I'm afraid not really feasible for us. 
Regarding the grammar, its just italic & bold styling + tags, hyperlinks and mentions actions, with simple rule that maximal one level nesting is allowed and style can be nested in style + actions can be nested in styling but not vice versa. 
In the future, we should thing about it and formalise it more, this is good first step, current state of code is that it's absolutely ad hoc and inconsistent between desktop and mobile.
@yenda With tokenizer you get vector as result and you need to map it to view tags. With tree it's the same you just map it to elements with dfs walk, it's the same only you need to traverse the children which is at the same time much more powerful (you can embed actions inside styling, styling inside styling, so for example italic and bold is possible at the same time)

yenda:
@janherich you can do that with vectors of tuples as well it just need to be recursive. I would like this PR to at least include an exemple of how you build the elements from the tree.

vitvly (approved):

yenda (changes requested):
please make a test that demo how to turn the tree into elements

lukaszfryc:
@janherich are you sure this PR will not impact app itself or maybe it's worth to do some quick tests?

mandrigin:
I was thinking if it is possible to use platform native features for that if supported. I know all iOS versions that we support have these APIs and most of the modern Androids too. They can figure out URLs, phone numbers, addresses, times & locations, durations, etc (you can filter that). and for most of the supported languages/alphabets too.

janherich:
@yenda I re-evaluated the decision to return tree as a render recipe, so now it's sequence instead, please see the updated code, I think it's very clear now how it can be rendered.

yenda:
@janherich If I'm not mistaken building the tree will be done at render time. After doing some benchmarking a good chunk of the time is still spent building the tree (roughly 25%). I still prefer a collection of vector.

The following solution for instance is slightly faster and the result can be used almost as is for rendering:

```
(defn links [text]
  (reduce (fn [acc [link & _]]
            (conj acc [link (str "\" [:link \"" link "\"] \"")]))
          #{}
          (re-seq constants/regx-url text)))

(defn tags [text]
  (reduce (fn [acc tag]
            (conj acc [tag (str "\" [:tag \"" tag "\"] \"")]))
          #{}
          (re-seq constants/regx-tag text)))

(defn mentions [text]
  (reduce (fn [acc mention]
            (conj acc [mention (str "\" [:mention \"" mention "\"] \"")]))
          #{}
          (re-seq constants/regx-mention text)))

(defn actions [text]
  (clojure.set/union (links text) (tags text) (mentions text)))

(defn italic [text]
  (mapv (fn [italic]
          [italic (str "\" [:italic " (string/replace italic "~" "\"") "] \"")])
        (re-seq constants/regx-italic text)))

(defn bold [text]
  (mapv (fn [bold]
          [bold (str "\" [:bold " (string/replace bold "*" "\"") "] \"")])
        (re-seq constants/regx-bold text)))

(defn replace-elements [text replacement-fn]
  (reduce (fn [text [match replacement]]
            (clojure.string/replace text match replacement))
          text
          (replacement-fn text)))

(defn wrap-in-vec [text]
  (str "[ \"" text "\" ]"))

(defn format-content [text]
  (clojure.walk/postwalk #(if (string? %)
                           (-> %
                               (string/replace "[:quote]" "\"")
                               (string/replace "[:escape]" "\\"))
                           %)
                         (-> text
                             (string/replace "\"" "[:quote]")
                             (string/replace "\\" "[:escape]")
                             (replace-elements actions)
                             (replace-elements italic)
                             (replace-elements bold)
                             wrap-in-vec
                             edn/read-string)))

```
it could be tested further by checking how well it roundtrips on random strings though

```
status-im.chat.models.message-content> (time (dotimes [x 1000] (format-content  (str"\"www.google" x".com *~hello\"\"~* lol links.com Test #status one three  https://link1.com *#core-chat (@developer)!* By https://link.com the way, ~nice link(https://link.com)~) www.google.com"))))
"Elapsed time: 1713.000000 msecs"
nil
status-im.chat.models.message-content> (time (dotimes [x 1000] (build-render-recipe (enrich-content {:text (str"\"www.google" x".com *~hello\"\"~* lol links.com Test #status one three  https://link1.com *#core-chat (@developer)!* By https://link.com the way, ~nice link(https://link.com)~) www.google.com")}))))
"Elapsed time: 1776.000000 msecs"
nil
status-im.chat.models.message-content> (time (dotimes [x 1000] (format-content (str"\"www.google" x".com *~hello\"\"~* lol links.com Test #status one three  https://link1.com *#core-chat (@developer)!* By https://link.com the way, ~nice link(https://link.com)~) www.google.com"))))
"Elapsed time: 1653.000000 msecs"
nil
status-im.chat.models.message-content> (time (dotimes [x 1000] (build-render-recipe (enrich-content {:text (str"\"www.google" x".com *~hello\"\"~* lol links.com Test #status one three  https://link1.com *#core-chat (@developer)!* By https://link.com the way, ~nice link(https://link.com)~) www.google.com")}))))
"Elapsed time: 1830.000000 msecs"
nil
```

janherich:
@yenda Of course I can make it faster by not being able to process nested styling (that's the slowest part), but the current solution is 2x faster then `parse-text` function in mobile message rendering (not even talking about the fact that it's naively memoized in view, instead of properly via subscriptions - and it's much more then 2x in common case of intermediate long message without any styling/actions).
I also definitely don't want to just take a string and convert it to render-vector in one step (like you did), it's better to have intermediate result (`enrich-content`), which is normalised data structure ideal for querying (necessary for mentions).
Given that the current solution is faster & more memory efficient (not to mention much more powerful) then what we have now (`parse-text`), I wouldn't dwell too much on perf implications, of course when we will need to make it super efficient, we can rework `build-render-recipe` to not perform tree walking and it will speed things up :).
However I would really like to lock the intermediate format (`enrich-content` result) as that's something I want to build upon when implementing mentions.

janherich:
@yenda By the way, your solution is **massively** slower in common case of message with no special formatting or links (even if it's just average length):
screen shot 2018-10-16 at 11 52 06

janherich:
@yenda I'm not sure what's the problem actually, as the `build-render-recipe` already returns flat collection of tuples.

yenda (approved):

PR #6279, opened 2018-10-12
support for all .eth username resolution

yenda:
ACTION BEFORE MERGE: replace "thisisme" by "stateofus" for name resolution without domain name

- resolve all eth usernames
- resolve stateofus.eth without domain

### Steps to test:
- barry should resolve as a username and open a chat with barry, the Wellit Impossible Crocodile

status: ready


src/status_im/ui/screens/add_new/new_chat/events.cljs
```diff
@@ -3,26 +3,29 @@
...
-                                                    chain)
-                                     :ens-name new-identity
-                                     :cb #(re-frame/dispatch [:new-chat/set-new-identity %])}})
-       {:db (assoc db
-                   :contacts/new-identity       new-identity
-                   :contacts/new-identity-error new-identity-error)}))))
+                                     :registry (get ens/ens-registries chain)
+                                     :ens-name (if (ens/is-valid-eth-name? new-identity)
+                                                 new-identity
+                                                 (str new-identity ".thisisme.eth"))
```
jeluard: Should be stateofus?

jeluard (changes requested):

jeluard (approved):

flexsurfer (approved):

src/status_im/ui/screens/add_new/new_chat/events.cljs
jeluard: Will be changed after test

asemiankevich:
@yenda so basically what should be expected in this PR? i dont get how `barry` could be resolved to some username.
So to do this i shoud register barry.stateofus.eth and then when create a private chat i can type only `barry` into this username field right? What about wallet send?

yenda:
@asemiankevich barry is already on mainnet. It's `barry.thisisme.eth` but you can just type `barry` and it will be resolved. Before merging I will switch `thisisme.eth` by `stateofus.eth` to resolve these names instead. It can also resolve any .eth name that has a pubkey registered if you type the full domain.

What should be tested is that you can resolve a `.thisisme.eth` name without the domain (which means it will work for `stateofus.eth` once it is deployed) and any name with the domain

asemiankevich:
can resolve `barry`, `Barry`, `barry.thisisme.eth` on mainnet and open chat with Wellit Impossible Crocodile

jeluard:
@yenda Can you replace `thisisme` ?

PR #6301, opened 2018-10-12
[#6094]: Persist extensions

### LINKED ISSUE #6094: Persist extensions
### User Story

As a user, I want my extensions to persist across restarts.

### Description

*Type*: Feature

After an extension is installed (part of the app-db) it should be persisted and reactivated upon restarts, based on its state.

TBD: do all extensions need to be persisted? What is the default behavior then?

alwx:
Fixes #6094

Also includes some very minor refactoring of extensions events

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -1,14 +1,17 @@
 (ns status-im.extensions.core
   (:require [clojure.string :as string]
-            [re-frame.core :as re-frame]
             [pluto.reader :as reader]
+            [pluto.registry :as registry]
             [pluto.storages :as storages]
+            [re-frame.core :as re-frame]
+            [status-im.accounts.update.core :as accounts.update]
             [status-im.chat.commands.core :as commands]
             [status-im.chat.commands.impl.transactions :as transactions]
             [status-im.ui.components.react :as react]
             [status-im.ui.components.button.view :as button]
-            [status-im.utils.handlers :as handlers]
+            [status-im.i18n :as i18n]
             [status-im.ui.screens.navigation :as navigation]
+            [status-im.utils.handlers :as handlers]
             [status-im.utils.fx :as fx]))
 
 (re-frame/reg-fx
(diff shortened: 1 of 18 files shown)


src/status_im/extensions/core.cljs
```diff
@@ -43,3 +47,64 @@
...
+  {:db (update db :extensions/manage assoc input-key {:value value})})
+
+(fx/defn fetch [cofx id]
+  (get-in cofx [:db :account/account :extensions id]))
+
+(fx/defn edit
+  [cofx id]
+  (let [{:keys [url]} (fetch cofx id)]
+    (-> (set-input cofx :url (str url))
+        (assoc :dispatch [:navigate-to :edit-extension]))))
```
janherich: Should be `navigation/navigate-to` call instead of unnecessary dispatch

janherich (changes requested):

alwx:
Rebasing is done!

jeluard (approved):

src/status_im/extensions/core.cljs
```diff
@@ -183,3 +186,63 @@
...
+  [{:keys [db]} input-key value]
+  {:db (update db :extensions/manage assoc input-key {:value value})})
+
+(fx/defn fetch [cofx id]
+  (get-in cofx [:db :account/account :extensions id]))
+
+(fx/defn edit
+  [cofx id]
+  (let [{:keys [url]} (fetch cofx id)]
+    (fx/merge (set-input cofx :url (str url))
```
janherich, line 200: I'm afraid you forgot to pul `cofx` as first argument to `fx/merge`

janherich (changes requested):

src/status_im/extensions/core.cljs
alwx: `(set-input cofx :url (str url))` doesn't count? At least it works as expected.

janherich (approved):

asemiankevich:
when installing extension
image

asemiankevich:
okay it is fixed now. lets merge

PR #6312, opened 2018-10-15
WIP feature #6297 - wallet on desktop

### LINKED ISSUE #6297: [EPIC] Port Wallet to Desktop
### User Story

As a user I want Wallet on Desktop, because... Wallet on Desktop.

### Description

As the very first version, we'll just try to port the mobile Wallet directly as it is to desktop and only solve the most severe responsiveness issues. This is merely an experiment to see if it would work at all.

- [ ] port 2 panel designs to desktop
- [ ] Implement chat commands on desktop #6945
- [ ] transaction history (http calls don't work on desktop, at least not on linux) - status-im/react-native-desktop#373
- [ ] set recipient by scanning qr (camera on desktop not supported)
- [ ] offline detection - always says it's offline on linux, so it's being ignored by wallet when on desktop

base

goranjovic:
fixes #6297

### Summary:

Wallet on Desktop. Added a tab, which displays main wallet screen in main area. Sidebar is empty.
Basic wallet functionality is also there, with some noted exceptions.

What works:
- signing phrase onboarding
- display of balance and tokens in main wallet screen
- adding/removing enabled tokens
- send transaction to contact or by address
- show own address qr in receive transaction
- set custom gas settings

What does not work:
- send a transaction request (we have no chat commands on desktop yet)
- transaction history (http calls don't work on desktop, at least not on linux) - status-im/react-native-desktop#373
- set recipient by scanning qr (camera on desktop not supported)
- generally this PR does only main Wallet tab, not chat commands and not dapp transactions.
- network switching, only mainnet
- offline detection - always says it's offline on linux, so it's being ignored by wallet when on desktop

Notable responsiveness issues:
- missing title in Assets selection screen (the one where you enable/disable tokens)
- blue area below token list in main wallet screen
- missing component that displays menu popups - using dialog as a short term solution
- checkbox component is weird on desktop

## Testing notes (optional):

All platforms should be tested - shared code was affected.

#### Platforms (optional)
- Android
- iOS
- macOS
- Linux

**Functional**
- wallet / transactions

status: ready

### DIFF
--- a/src/status_im/ui/components/checkbox/styles.cljs
+++ b/src/status_im/ui/components/checkbox/styles.cljs
@@ -1,4 +1,5 @@
 (ns status-im.ui.components.checkbox.styles
+  (:require-macros [status-im.utils.styles :refer [defstyle]])
   (:require [status-im.ui.components.colors :as colors]))
 
 (def wrapper
@@ -18,9 +19,11 @@
           :width         26
           :height        26}))
 
-(def check-icon
-  {:width  12
-   :height 12})
+(defstyle check-icon
+  {:width   12
+   :height  12
+   ;;TODO(goranjovic) - this is a quickfix because checkboxes are invisible on desktop
+   :desktop {:background-color colors/blue}})
 
 (def plain-check-icon
   (merge check-icon
--- a/src/status_im/ui/components/toolbar/styles.cljs
+++ b/src/status_im/ui/components/toolbar/styles.cljs
@@ -16,6 +16,7 @@
    :justify-content  :space-between
    :background-color (or background-color toolbar-background)
    :elevation        (if flat? 0 2)
+   :desktop          {:height 55}
    :android          {:height 55}
    :ios              {:height 56}})
 
@@ -68,7 +69,8 @@
 (defstyle item
   {:ios     {:padding-horizontal 12
              :padding-vertical   16}
-   :android {:padding 16}})
+   :android {:padding 16}
+   :desktop {:padding 16}})
 
 (def item-text
   {:color     colors/blue
@@ -85,9 +87,20 @@
 
 (def item-text-white-background {:color colors/blue})
 
-;;TODO(goranjovic) - Breaks the toolbar title into new line on smaller screens
-;;e.g. see Discover > Popular hashtags on iPhone 5s
-(def ios-content-item {:position :absolute :right 40 :left 40})
+;;TODO(goranjovic) - This doesn't work at all on desktop. Review why we even need absolute position hacks
+;; hiding it as a temporary fix.
+(defstyle ios-content-item
+  {:desktop {:position :absolute
+             :right    40
+             :left     40
+             :width    0
+             :height   0}
+   :android {:position :absolute
+             :right    40
+             :left     40}
+   :ios     {:position :absolute
+             :right    40
+             :left     40}})
 
 (def counter-container
   {:top 3})
--- a/src/status_im/ui/screens/wallet/main/styles.cljs
+++ b/src/status_im/ui/screens/wallet/main/styles.cljs
@@ -7,7 +7,8 @@
 (defstyle main-section
   {:flex    1
    :android {:background-color colors/white}
-   :ios     {:background-color colors/blue}})
+   :ios     {:background-color colors/blue}
+   :desktop {:background-color colors/blue}})
 
 (defstyle scroll-bottom
   {:background-color colors/white
@@ -15,6 +16,7 @@
    :position         :absolute
    :left             0
    :right            0
+   :desktop          {}
    :android          {:height 0}
    :ios              {:height 9999}})
 
@@ -127,7 +129,8 @@
    :color     colors/black})
 
 (def asset-item-currency
-  {:font-size   16
+  {:flex 1
+   :font-size   16
    :color       colors/gray
    :margin-left 6})
 
--- a/src/status_im/ui/screens/wallet/send/views.cljs
+++ b/src/status_im/ui/screens/wallet/send/views.cljs
@@ -29,6 +29,7 @@
             [taoensso.timbre :as log]
             [reagent.core :as reagent]
             [status-im.ui.components.colors :as colors]
+            [status-im.utils.platform :as platform]
             [status-im.ui.screens.wallet.utils :as wallet.utils]))
 
 (defn- toolbar [modal? title]
@@ -144,7 +145,8 @@
         chain                        (ethereum/network->chain-keyword network)
         native-currency              (tokens/native-currency chain)
         {:keys [decimals] :as token} (tokens/asset-for all-tokens chain symbol)
-        online? (= :online network-status)]
+        ;; TODO(goranjovic): offline status detection doesn't work on desktop yet
+        online? (or platform/desktop? (= :online network-status))]
     [wallet.components/simple-screen {:avoid-keyboard? (not modal?)
                                       :status-bar-type (if modal? :modal-wallet :wallet)}
      [toolbar modal? (i18n/label :t/send-transaction)]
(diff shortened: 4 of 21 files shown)


goranjovic:
screenshot from 2018-10-15 11-36-20


screenshot from 2018-10-15 12-21-02

screenshot from 2018-10-15 12-21-16

screenshot from 2018-10-15 13-04-48

screenshot from 2018-10-15 12-46-14

screenshot from 2018-10-15 12-46-31

chadyj:
Woah great one Goran!

Take a look @EugeOrtiz @denis-sharypin @errorists @andmironov 

Should we redesign wallet to work with 2 panes, or perhaps look at a desktop layout that supports 1 pane? Would be great to 1) get this looking decent as a first step, and 2) look at best design solutions for the longer term.

goranjovic:
Also keep in mind that we'll be implementing Wallet Redesign soon enough, so it makes little sense to micro-optimize the existing layout for desktop.  On the other hand, we do want the new screens to look good when big.

chadyj:
> Also keep in mind that we'll be implementing Wallet Redesign soon enough, so it makes little sense to micro-optimize the existing layout for desktop

In that case what about making the left panel blue so the contrast isn't so jarring?

goranjovic:
Good idea, but to be fair, what to put in the sidebar is a good question in old and new designs alike.

dmitryn (approved):

yenda:
Did you try getting rid of the conditional in `init.core`?
```
            #(when-not platform/desktop?
               (initialize-wallet %))
```

goranjovic:
@yenda Haven't even noticed it :D Thanks!

chadyj:
@goranjovic I added you to the Desktop sync tomorrow so we can discuss Wallet and getting this merged. Also @corpetty is joining to discuss desktop security.

chadyj:
We touched on wallet in the desktop sync, but didn't get a chance to hear from @MaxRis @siphiuel @vkjr about whether we should include Wallet in Desktop, most likely behind a "dev mode" flag.  

The reason Wallet is important for Desktop is that it allows our primary use case of sending assets to contacts, but more importantly, it will allow SNT use cases like moderation, tribute to talk, and more. Status without SNT isn't really Status, so I'd advocate that we include it for visibility, and transparency.  Like everything else in Desktop it is a work in progress, but having it will focus attention in much the same way Desktop has done for chat.

maxhora:
I agree that Wallet is essential part. Especially, if to compare with competitors. Can I help somehow to speedup things for merging this in without breaking others functionality?

vitvly:
Agree with a proposal to include it under a dev flag in Desktop.

goranjovic:
Created an issue to add dev mode and network switching to desktop #6477

vkjr:
Looks like I was thinking about this but forget to say in voice on the meeting :)
I think that work on wallet will distract focus point from polishing chat, since it takes not only dev, but also QA and design resources. But if we have enough of them, I think adding it under DEV is a good solution.

src/status_im/ui/components/checkbox/styles.cljs
```diff
@@ -18,9 +19,11 @@
           :width         26
           :height        26}))
 
-(def check-icon
-  {:width  12
-   :height 12})
+(defstyle check-icon
+  {:width   12
+   :height  12
+   ;;TODO(goranjovic) - this is a quickfix because checkboxes are invisible on desktop
```
vkjr, line 25: @goranjovic, is it a bug in `react-native-desktop`? Would you mind adding corresponding issue to rnd repo?

src/status_im/ui/components/toolbar/styles.cljs
```diff
@@ -85,9 +87,20 @@
 
 (def item-text-white-background {:color colors/blue})
 
-;;TODO(goranjovic) - Breaks the toolbar title into new line on smaller screens
-;;e.g. see Discover > Popular hashtags on iPhone 5s
-(def ios-content-item {:position :absolute :right 40 :left 40})
+;;TODO(goranjovic) - This doesn't work at all on desktop. Review why we even need absolute position hacks
```
vkjr, line 90: Creating `react-native-desktop` issue with a small js example representing a problem would be very helpful.

src/status_im/ui/screens/wallet/main/styles.cljs
```diff
@@ -120,7 +122,8 @@
    :color     colors/black})
 
 (def asset-item-currency
-  {:font-size   16
+  {:flex 1
```
vkjr, line 132: Incorrect indentation

src/status_im/ui/screens/wallet/send/views.cljs
```diff
@@ -141,7 +142,8 @@
   (let [{:keys [amount amount-text amount-error asset-error show-password-input? to to-name sufficient-funds?
                 sufficient-gas? in-progress? from-chat? symbol]} transaction
         {:keys [decimals] :as token} (tokens/asset-for (ethereum/network->chain-keyword network) symbol)
-        online? (= :online network-status)]
+        ;; TODO(goranjovic): offline status detection doesn't work on desktop yet
```
vkjr, line 148: @siphiuel, weren't we fixing this?

vkjr (approved):

src/status_im/ui/components/toolbar/styles.cljs
goranjovic: @vkjr This one is entirely `status-react` issue. At some point we used an absolute positioning hack to fix centering on some iphones. This is that issue coming back at us with vengeance.

src/status_im/ui/components/checkbox/styles.cljs
goranjovic: @vkjr yep, adding them all

denis-sharypin:
Guys, @goranjovic , @chadyj  I would prefer to have such layout for desktop's wallet. Wdyt?

base

goranjovic:
@denis-sharypin Just wondering if this idea is a reaction to the given fact that desktop ui already has a sidebar or if there are another reason to go with it?

@siphiuel I think I remember you or someone else mentioning that the reason sidebar was there in the first place is some issue with modals?

chadyj:
That UI looks great and seems to be a way we can bridge mobile and desktop through the 2 pane approach (and is conceptually similar to chat) with the least amount of effort with great results.

What would the default screen be on the right pane?

vitvly:
@goranjovic Yes, but this issue has been recently resolved: https://github.com/status-im/react-native-desktop/issues/326

goranjovic:
@chadyj What I'm trying to figure out is if we even want to stick with the "two pane" solution, given that the original reason for it no longer exists.

chadyj:
Pro of 2 pane is it's compatible with mobile implementation and doesn't require a dedicated redesign.

Con? No designs. Double implementation?

goranjovic:
Con is parallel implementation of a part of navigation. Because there are two levels of navigation on desktop, the outer one for the left pane and the inner one for the right one.

On mobile there is only one screen area, and thus only one navigation.

If we go with two panes, we will need separate navigation logic on both platforms.

denis-sharypin:
But this job was already done for the desktop chat?

goranjovic:
The only thing I want to avoid is us implementing the two panels in Wallet only to double back and remove two panels altogether. That's why I'm raising the question.

chadyj:
@denis-sharypin do you have plans to redesign wallet to 2 panels?
Or are you committed to this approach? https://github.com/status-im/status-react/pull/6312#issuecomment-442043419

Big decisions!

denis-sharypin:
As was discussed on our wallet sync call we are implementing two panels design here https://github.com/status-im/status-react/pull/6312#issuecomment-442043419

chadyj:
Goran shall we close this and revisit once the redesigns are in mobile?
We can create a new EPIC and scope out todo's.

PR #6313, opened 2018-10-15
[#6307] fix each message sent before upgrade is shown twice after 0.9.29

### LINKED ISSUE #6307: Each message sent before upgrade is shown twice after upgrade from 0.9.29
### Description

*Type*: Bug
*Summary*: Each message sent before upgrade is shown twice after upgrade from 0.9.29
https://s3.amazonaws.com/testfairy/screenshots/4803622/8626941/109020185/14_643b03d7310e82ad5fd0f3960e8ff3224fc6032b_o.jpg
#### Expected behavior
Message is shown once
#### Actual behavior
Message sent once is shown twice after upgrade
### Reproduction

- Install 0.9.29 Status
- Send 3 messages to public chat, e.g. 1, 2 and 3
- Upgrade to nightly build
- open the public chat and wait for messages to be fetched.
- check how sent 1,2,3 are shown now (in my case it's 1,1,2,2,3,3)

after upgrade: https://app.testfairy.com/projects/4803622-status/builds/8626941/sessions/4403987481/?accessToken=sBz/h9vUxgG03QEz9C1nwqnEpZE

### Additional Information
* Status version: upgrade from 0.9.29 from playstore to nightly from Oct 14
[comment]: # (Android or iOS.)
* Operating System: <Android|iOS>

<blockquote></blockquote>
<blockquote><div><strong><a href="https://app.testfairy.com/projects/4803622-status/builds/8626941/sessions/4403987481/?accessToken=sBz/h9vUxgG03QEz9C1nwqnEpZE">TestFairy: Status 0.9.29 (10419) @2018</a></strong></div></blockquote>

rasom:
fix #6307

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -145,10 +145,13 @@
             (chat-loading/group-chat-messages chat-id (get chat->message chat-id))))
 
 (defn- add-to-chat?
-  [{:keys [db]} {:keys [chat-id clock-value message-id] :as message}]
+  [{:keys [db]} {:keys [chat-id clock-value message-id message-id-old-format] :as message}]
   (let [{:keys [deleted-at-clock-value messages not-loaded-message-ids]}
         (get-in db [:chats chat-id])]
     (not (or (get messages message-id)
+             ;; TODO(rasom): remove this condition
+             ;; on when 0.9.29 will not be available for users
+             (and message-id-old-format (get messages message-id-old-format))
              (get not-loaded-message-ids message-id)
              (>= deleted-at-clock-value clock-value)))))
 
--- a/src/status_im/transport/message/protocol.cljs
+++ b/src/status_im/transport/message/protocol.cljs
@@ -105,13 +105,18 @@
            this)
           (send-with-pubkey cofx params)))))
   (receive [this chat-id signature _ cofx]
-    {:chat-received-message/add-fx
-     [(assoc (into {} this)
-             :message-id (transport.utils/message-id this)
-             :show?      true
-             :chat-id    chat-id
-             :from       signature
-             :js-obj     (:js-obj cofx))]})
+    (let [old-message (Message. (:text content) content-type message-type
+                                clock-value timestamp)]
+      {:chat-received-message/add-fx
+       [(assoc (into {} this)
+               :message-id (transport.utils/message-id this)
+               ;; TODO(rasom): remove this condition
+               ;; on when 0.9.29 will not be available for users
+               :message-id-old-format (transport.utils/message-id-old-format old-message)
+               :show? true
+               :chat-id chat-id
+               :from signature
+               :js-obj (:js-obj cofx))]}))
   (validate [this]
     (if (spec/valid? :message/message this)
       this
--- a/src/status_im/transport/utils.cljs
+++ b/src/status_im/transport/utils.cljs
@@ -22,6 +22,15 @@
   [message]
   (sha3 (pr-str message)))
 
+(defn message-id-old-format
+  "Get an old format message-id.
+   To be removed on 8th day after 0.9.30"
+  [message]
+  (-> message
+      pr-str
+      (clojure.string/replace "message.protocol" "message.v1.protocol")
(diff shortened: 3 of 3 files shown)


janherich (approved):

src/status_im/chat/models/message.cljs
```diff
@@ -148,10 +148,13 @@
             (chat-loading/group-chat-messages chat-id (get chat->message chat-id))))
 
 (defn- add-to-chat?
-  [{:keys [db]} {:keys [chat-id clock-value message-id] :as message}]
+  [{:keys [db]} {:keys [chat-id clock-value message-id message-id-old-format] :as message}]
   (let [{:keys [deleted-at-clock-value messages not-loaded-message-ids]}
         (get-in db [:chats chat-id])]
     (not (or (get messages message-id)
+             ;; TODO(rasom): remove this condition
```
cammellos, line 152: Not sure it is accurate, I am guessing your are thinking of nightlies, but if we don't include this in the next release, any user that goes to 0.9.29 -> 0.9.31 will still experience the issue (and so on) or I am missing something? have we made a decision that that's ok?
(genuinely asking not expressing my opinion on the issue).

rasom: Yeah, that's true, that user will encounter this problem. 

No decisions were made, my opinion is that it is ok. But we also can keep this condition for all period while 0.9.29 will be available, for TestFlight it is 77 days from now and idk how to check this for Play Market.

Another option might be a migration for ids, but it might cause even more bugs in the result.

cammellos: I would agree that is not a big deal, and seems ok to just leave the code for a bit longer.
If we do go for a migration though I think we should re-think the id generation, as the current method has the same drawbacks as the previous (namespace included / relies on consistent ordering of map keys). (maybe even worth doing now, but I understand there some time pressure).

rasom: yep, there is definitely some pressure, i suppose we will just return to this again later...

rasom: @cammellos, i changed comment a bit

cammellos (approved):

Serhy:
@rasom still can reproduce the issue (tested on Android 7). I need just to wait long enough for messages finished to be fetched. 

1) Install 0.9.29 release build
2) Open Status and create a new account
3) Join any unique public chat (e.g. `sendasand`)
4) Send 5 text messages 
5) Close the app
6) Install https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-181016-075924-befc26-pr.apk build
7) Log in to you contact previously created and wait till `Fetching messages...` disappear and check  public chat

<blockquote></blockquote>

rasom:
@yenda @cammellos any ideas what can be wrong with this PR? When i test it locally it works fine, but when upgrading from 0.9.29 on Android it is still reproducible...

rasom:
oh well https://github.com/status-im/status-react/blob/c017951c14ba22b12a039b57cb3e6619b3edf6f8/src/status_im/transport/message/transit.cljs#L104 `(:text content)` might be the reason

yenda:
@rasom yes namespaces are probably not the only potential cause of problems, if the content passed to the message changes as well between version, best to replace content by :text value I guess

rasom:
well, it seems that message id was broken before namespace renaming

rasom:
@Serhy ready for testing again

Serhy:
@rasom yes, looks good to me! 
Upgrade on Android 6.0.1, Android 7 (from release 0.9.29 app to latest builds here) and MacOS upgrade has no duplicated messages. 
For iOS, because of issue https://github.com/status-im/status-react/issues/6339 I wasn't able to upgrade from 0.9.29 TestFligh Rlease build to iOS build from https://github.com/status-im/status-react/pull/6313#issuecomment-430374487. So I created a build based on release/0.9.29 branch and tested based on it, which is OK.
Failed autotests reviewed and confirmed as not related to this PR (false Fail or known issues)

This one ready for merge. Thanks a lot, @rasom !

PR #6242, opened 2018-10-10
Pairing & multi device & group-chats

cammellos:
### Pairing

Add a new section (under advanced) where you can pair new devices.
I have included a commit from a different PR as both rely on a status-go update and probably best to be tested together.

The code is behind a feature flag but enabled in nighties if development mode is on.
The UI is not polished and this is only the first step, but I decided to enable it in nightlies in dev-mode as it can already provide value to whoever wants to use it.

It is only used to sync up your devices contacts. Group chats will also receive your own messages once devices are paired correctly.

### What's left

1) Polish UI
2) Let the user manually confirm the devices (currently they are enabled without user confirmation)
3) Add information about each device (is this desktop/mobile/last-seen)
4) Sync other kind of info (your name etc)
5) Keep data in sync after initial pairing (contacts, 1-to-1 messages)

### Testing

Both group chats & pairing should only be enabled in nightlies/dev if dev-mode is on.

To pair your devices:

1) Install this build on both desktop & mobile
2) On android/ios go to settings->advanced->devices, on desktop go to Profile->Devices->Pair
3) Pressing `+` will advertise your device. You should now see a new ID in all the `devices` sections.
4) Do the same on the other devices until all your devices shows up in the list.
5) Press on any device ID to share your contacts with all the rest of the devices.
6) Group chats should receive your own messages now, if you start a new group chat with someone, their message should be received by all the paired devices.

status: ready

PR #6362, opened 2018-10-17
Synchronize your own messages in 1-to-1

cammellos:
Once two devices are paired, you should see your own messages in 1-to-1 chats on different devices.

Depends on: #6242 

If you have a device with a previous version it will see those messages as coming from you (a chat opened with yourself will receive those messages).

I don't think is a big deal as the user has control over their devices (they just have to upgrade the version), but let me know if you think otherwise.

Enabled only in nightlies & when dev-mode is on, same as pairing.

Only the last commit is relevant.

We allow now the user to set an arbitrary chat-id in the message, so this needs to be properly validated.
The logic for the validation is here https://github.com/status-im/status-react/blob/62db799d60fe5a852ad6c5285658092488ad7beb/src/status_im/chat/models/message.cljs#L171 and probably needs to be looked at carefully. 

### Step to tests

1) You need to have enabled development mode
2) Go to settings -> Devices
3) On android/ios click on `+`, desktop click on `pair` to advertise your device
4) Any other device should see the new device with a 3-words random name + device-type under `devices`
5) If you enable the device, it will send 1-to-1 & group messages to that device, make sure that devices are enabled both ways
6) New group chats will also be synchronized

The UI is still rough and pending designs.

It will only sync max 4 extra devices for now, we will work on the UI to make sure the user don't enable more.

Enables in nightlis in development mode

status: ready


ghost:
### Pull Request Checklist

yenda:
Could you explain why we need to send the chat-id in the message content ?

cammellos:
@yenda sure, basically device A1 when sending a message to device B1 will also send a message to A2 to keep it in sync.

Currently we use signatures to understand which chat-id the message is for.

In this case, the message will have a signature of A but the chat-id it should be used is actually B (as this is a message sent from A to B).

There's a more in depth discussion in #status-protocol as well (and the reason why it was decided to put it in `content`), but I can go through the reasoning as well here if you'd like.

yenda:
@cammellos I forgot to rejoin that channel after I nuked my db so I can't see the discussion, could you sum it up here ? 
Also I don't see any problem with that for syncing own messages but how about messages between A and B ? chat-id should be ignored right ?

cammellos:
@yenda chat-id if specified is  checked against the signature and if they
don't match the message is discarded ( if not present chat-id is thr
signature)

for the discussion, basically we wanted to move away from whisper
dependency and contextual data ( chat-id not being passed through rhe wire
and pulled by the filter callback).

the two options were to have a special sync message or enrich the current
with chat id.

enriching was seen as the preferred strategy.


in terms of how, 2 options, extend transit with an extra field or add r
to content map.

adding to the content map was preferred as does not require to remove chat
id in the calculation of message id in new client ( as it would be ignored
in old clients if we added a field to transit)


let me know what you think and if something is unclear

yenda:
@cammellos Yeah I'm find with enriching content I think everything should be in a map in the end for extensibility anyway.

But regarding chat-id shouldn't we ignore chat-id when the message is not from yourself as long as we are using whisper ?

cammellos:
@yenda sure,

basically the implications is that we will allow messages that have a `chat-id` malformed to be stored in the database (`chat-id` != `signature` , we are only talking about messages sent from other peers, not ourselves), so it's a bit more permissive.

I don't think it's a problem but at this point we need to be careful knowing that in the database they might be some messages that needs the same check when pulling them out (in case we re-use the `chat-id` field, I don't think it's a concern now, but worth keeping this in mind).

We can also replace the `chat-id` field if it's not equal to the signature, at that point messages are mutable and does not play well with message-id generation, but does not seem to be a concern here as it would clearly be a malformed message.

just a bit of random thoughts, what do you prefer?

yenda:
@cammellos I think that in the database we should remove the chat-id field anyway as the message schema already have a chat-id field, so there is no point storing a second one

cammellos:
@yenda ok, so we are no concerned in being able to restore the raw data if necessary? (once stored in the database we don't know anymore if it had a `chat-id field or no, so we loose the ability to replay messages for example, re-calculate message-id, etc).

I myself don't have a strong opinion on the matter, I would err on the side of caution here if this is the change that will drop this property (do we even have it by now?), as it might end up being useful in the future, and the space requirements for chat-id seems fairly modest.

But it all sounds good, what do other thinks? @janherich @mandrigin

janherich:
I agree with @cammellos that I would rather err on the cautious side and store the messages as they are received/sent, even if it means (sometimes) duplicate chat it in top level message map & content field as well.
The upside is that we will not lose the information that some outgoing message was received as sync from other device, better to have it then not, we can always simplify it in the future when it's proved that everything works flawless and there is no real point retaining that piece of info.

janherich (approved):

src/status_im/events.cljs
```diff
@@ -1101,10 +1101,34 @@
...
 (handlers/register-handler-fx
  :pairing.ui/synchronize-installation-pressed
  []
  (fn [cofx _]
    (pairing/send-installation-message cofx)))
+
+(handlers/register-handler-fx
+ :pairing.ui/enable-installation-pressed
+ []
+ (fn [cofx [_ installation-id]]
```
dmitryn, line 1117: is `[]` as argument is really needed here?

dmitryn (approved):

src/status_im/events.cljs
cammellos: not needed, thanks :)

churik:
@cammellos AFAIK sync is not enabled in current nightlies or it should be after merge https://github.com/status-im/status-react/pull/6532?

cammellos:
@churik sync of contacts is disabled in both, sync of 1-to-1 messages/group chats is enabled here to only paired devices and both group-chats & pairing will be enabled in nightlies if merged.
https://github.com/status-im/status-react/pull/6532 works a bit differently (group chats are kept in sync automatically if a device with the same identity is detected, but that has been changed in this version, which uses a more recent version of status-go).

 If you want you can test only this PR as it includes https://github.com/status-im/status-react/pull/6532.

Thanks for picking this up!

cammellos:
@churik just to make sure

On IOS you should see a device only after you click on "Pair" on desktop
On Desktop you should see a device only after you click on "+" on ios

Did you click on "Pair devices" on desktop? (sorry if the instructions weren't clear)

After a message has been sent, it will detect the device even if you haven't clicked on pair, which seems is what you described, which is the expected behavior

churik:
@cammellos yes, I just didn't realize that for discovering new device for others you need to click on + (it was working a bit different when I tested it last time).
I'm a bit concerned that to add and recognize new device you need to add + (or `Pair device`) in all your devices. It is not clear at all.
Should it be improved in future?

cammellos:
@churik yes, you will always have to click on `+` on all your device, but with proper UX it will be much clearer,  @errorists is working on UX/Designs  and looks really good so far, https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=204%3A33 , he just sent me the link this morning.

<blockquote><div> Figma</div><div><strong><a href="https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile">Figma</a></strong></div><div>Created with Figma</div></blockquote>

churik:
@cammellos several issues:
### 1. App is crashed after log in when you disable sync when app is closed (desktop)
Prerequisites:
User A has app on IOS and desktop, sync is enabled
Steps:
- close app on desktop
- open app on IOS
- open profile > advanced > devices
- disable sync with desktop
- open app on desktop and try to log in

Expected result:
can log in
Actual result:
crash
Crash log: [Crash.log](https://github.com/status-im/status-react/files/2548538/Crash.log)
Status log: [Status.log](https://github.com/status-im/status-react/files/2548541/Status.log)

### 2. Name is too long, and sometimes the button doesn't fit screen size
As suggestion make button starting from new row  until new design is not implemented
ios

### 3. Q - if sync was off, then it was turned on -> all messages sent during period when sync was off weren't synchronized. Is it expected?

 ### 4. Q - cannot find out how to sync contacts with new interface, can you explain please?

 ### 5. App is crashing when you send 15+ messages quickly in a row (desktop)

Prerequisites:
User A has app on IOS and desktop, sync is enabled

**Scenario 1:**
- send 15+ messages quickly from desktop in group or 1-1 chat
App is crashed.

**Scenario 2:**
- close desktop app
- send from mobile 20+ messages in 1-1 chat
- open desktop and log in, send several messages in 1-1 chat
App is crashed.
Video: http://take.ms/2U5Of
Crash log: [Crash.log](https://github.com/status-im/status-react/files/2548718/Crash.log)
Status log: [Status.log](https://github.com/status-im/status-react/files/2548721/Status.log)
Geth log: [geth.log](https://github.com/status-im/status-react/files/2548872/geth.log)

cammellos:
@churik thanks for the feedback
3 is expected, only messages when sync was on will be sent to other devices
4 I have disabled the feature as I didn't have time to address the issues, will enable in a separate PR once this has been merged

I will look at the crashes and let you when fixed, thanks!

cammellos:
@churik do you also have `geth.log` by any chance?

churik:
@cammellos added

cammellos:
thanks, last thing, could you try to replicate on latest nightly (looking at the crash it seems like it's in qt space, I can't replicate on Linux)? (you can just send messages to someone else)
Also how many devices you had paired?

churik:
@cammellos I did, not reproducible

cammellos:
@churik Could you please try out the last build? 
Thanks!

churik:
@cammellos 
Crashes are not reproducible.
Will create remaining issues from https://github.com/status-im/status-react/pull/6242, restoring sync contacts and issue 2 from https://github.com/status-im/status-react/pull/6362#issuecomment-435893574 with long name separately.
I think it shouldn't be enabled in nightly because feature is still WIP.

cammellos:
thanks! I will merge the PR after the fix for the logs, as it's the same
issue, provided it fixes the problem.

On Nov 6, 2018 8:00 PM, "Tetiana Churikova" <notifications@github.com>
wrote:

@cammellos <https://github.com/cammellos>
Crashes are not reproducible.
Will create remaining issues from #6242
<https://github.com/status-im/status-react/pull/6242>, restoring sync
contacts and issue 2 from #6362 (comment)
<https://github.com/status-im/status-react/pull/6362#issuecomment-435893574>
with long name separately.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/status-im/status-react/pull/6362#issuecomment-436368958>,
or mute the thread

churik:
@cammellos I can see Devices on last nightly on Desktop.
Supposed that it will be turned off.
Am I wrong?

cammellos:
@churik I didn't see your edit where you suggested to turn it off (github don't send notifications for edits, and I replied directly to the email).

In general if the problems are only cosmetic seems to me it's a good use case for nightlies (on mobile where the problem is , it is only enabled in development mode), or there are other issues part from cosmetic?

I can probably just as easily fix the cosmetic issues rather than disabling, as designs are almost ready.

churik:
@cammellos 
Thank you, amazing work.
I retested all remainig issues from https://github.com/status-im/status-react/pull/6242 and from here and think, that only two issues are left:
#6638
#6639

PR #6363, opened 2018-10-17
[TESTS] Adapt normal txs tests to work with both Ropsten and Rinkeby test networks

hetvart:
## Summary:
Due to a recent issue with Constantinople Ropsten hard fork update some normal txs tests were failing.
This PR proposes a fix which enables integration with Rinkeby in case of Ropsten unavailability.

### Review notes:
- made network api object to work with Rinkeby
- implemented faucet depending under which network you run tests (i.e. --network flag)
- fixture for defining under which network you run directly in test

IMPORTANT NOTE: 
To control which network/cryptocurrency/api to use there is the set_network fixture defined.

```python
@pytest.fixture(autouse=True)
def set_network(request):
    network = request.config.getoption('network')
    request.instance.network_api = NetworkApi(network=network)
    if network == 'ropsten':
        request.instance.network = 'Ropsten with upstream RPC'
        request.instance.crypto_currency = 'ETHro'
    elif network == 'rinkeby':
        request.instance.network = 'Rinkeby with upstream RPC'
        request.instance.crypto_currency = 'ETHri'
    else:
        request.instance.network = 'Mainnet with upstream RPC'
        request.instance.crypto_currency = 'ETH'
```


yevh-berdnyk (approved):

antdanchenko (changes requested):
Please, redesign the PR in the way when `request.config.getoption('network')` defines which network should be set in tests so tests marked with `switch_network` will run on the chosen network

antdanchenko:
@hetvart great, thanks for the PR, but let's merge the PR after the release

antdanchenko (approved):

PR #6370, opened 2018-10-17
[#2884] added name lookup for recipient after qr code scanning

### LINKED ISSUE #2884: Recipient name is not shown when scanning QR code to send transaction from wallet
### Description

*Type*: Bug

*Summary*: Only recipient’s address shown in the Recipient field when scanning QR code of the request from wallet

#### Expected behavior
Recipient name is shown in the Recipient field

#### Actual behavior
Recipient address is not shown in the Recipient field

### Reproduction
As user A:

1. Open Status app
2. Add user B as a contact
3. Click Wallet button
4. Click Request button
5. Enter valid number to amount input

As user B:

1. Open Status app
2. Click Wallet button
3. Click Send button
4. Click Choose Recipient
5. Scan QR code from user A screen

### Additional Information
Status version: 0.9.12+
Operating System user A: Emulator Google Nexus 5, Android 6.0.0
Operating System user B: Real device OnePlus A5000, Android 7.1.1

kfirmanty:
Fixes #2884.

Should now display contact name beside address when QR code is scanned.

I have tried to make scope of changes as small as possible, but address comparison function was private in contact/subs so I moved it to utils.contacts without changing any logic.

It seems that the same problem is present when entering address manually (name is not displayed even when contact address is entered) and I think that it could be solved in a similar manner.

### DIFF
--- a/src/status_im/contact/subs.cljs
+++ b/src/status_im/contact/subs.cljs
@@ -1,7 +1,6 @@
 (ns status-im.contact.subs
   (:require [re-frame.core :refer [reg-sub subscribe]]
             [status-im.utils.contacts :as utils.contacts]
-            [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.identicon :as identicon]))
 
 (reg-sub :get-current-contact-identity :contacts/identity)
@@ -149,19 +148,10 @@
                :else
                (identicon/identicon chat-id)))))
 
-(defn- address= [{:keys [address] :as contact} s]
-  (when (and address (= (ethereum/normalized-address s)
-                        (ethereum/normalized-address address)))
-    contact))
-
-(defn- contact-by-address [[_ contact] s]
-  (when (address= contact s)
-    contact))
-
 (reg-sub :get-contact-by-address
          :<- [:get-contacts]
          (fn [contacts [_ address]]
-           (some #(contact-by-address % address) contacts)))
+           (utils.contacts/find-contact-by-address contacts address)))
 
 (reg-sub :get-contacts-by-address
          :<- [:get-contacts]
--- a/src/status_im/utils/contacts.cljs
+++ b/src/status_im/utils/contacts.cljs
@@ -1,7 +1,8 @@
 (ns status-im.utils.contacts
   (:require [status-im.js-dependencies :as js-dependencies]
             [status-im.utils.identicon :as identicon]
-            [status-im.utils.gfycat.core :as gfycat]))
+            [status-im.utils.gfycat.core :as gfycat]
+            [status-im.utils.ethereum.core :as ethereum]))
 
 (defn whisper-id->new-contact [whisper-id]
   {:name             (gfycat/generate-gfy whisper-id)
@@ -17,3 +18,10 @@
                          nil)]
     (when normalized-key
       (subs (.sha3 js-dependencies/Web3.prototype normalized-key #js {:encoding "hex"}) 26))))
+
+(defn- contact-by-address [[_ contact] address]
+  (when (ethereum/address= (:address contact) address)
+    contact))
+
+(defn find-contact-by-address [contacts address]
+  (some #(contact-by-address % address) contacts))
--- a/src/status_im/ui/screens/wallet/choose_recipient/events.cljs
+++ b/src/status_im/ui/screens/wallet/choose_recipient/events.cljs
@@ -6,7 +6,8 @@
             [status-im.utils.handlers :as handlers]
             [status-im.utils.money :as money]
             [status-im.utils.ethereum.ens :as ens]
-            [re-frame.core :as re-frame]))
+            [re-frame.core :as re-frame]
+            [status-im.utils.contacts :as utils.contacts]))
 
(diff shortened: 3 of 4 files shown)


ghost:
### Pull Request Checklist

src/status_im/contact/subs.cljs
```diff
@@ -147,19 +146,10 @@
                :else
                (identicon/identicon chat-id)))))
 
-(defn- address= [{:keys [address] :as contact} s]
-  (when (and address (= (ethereum/normalized-address s)
```
rasom, line 153: If we check the address for being not nil here, let's check `s` as well.

src/status_im/contact/subs.cljs
```diff
@@ -147,19 +146,10 @@
                :else
                (identicon/identicon chat-id)))))
 
-(defn- address= [{:keys [address] :as contact} s]
```
rasom: IMO function's name implies that we are going to compare addresses, so I would rather avoid passing the whole contact as a parameter here. `[address1 address2]` or something `[x y]` would fit better.

src/status_im/utils/contacts.cljs
```diff
@@ -17,3 +18,16 @@
                          nil)]
     (when normalized-key
       (subs (.sha3 js-dependencies/Web3.prototype normalized-key #js {:encoding "hex"}) 26))))
+
+(defn- address= [{:keys [address] :as contact} s]
```
rasom: let's put these functions to some common namespace in order to avoid duplication

rasom (dismissed):
see comments

src/status_im/contact/subs.cljs
kfirmanty: No problem, I will clean up this code I just assumed that because I am using code that is already there I will change as little as possible and then maybe clean up code as a part of other task, but I agree that it definitely could be cleaner.

src/status_im/utils/contacts.cljs
kfirmanty: I already moved it from src/status_im/contact/subs.cljs to src/status_im/utils/contacts.cljs so it will be still in contacts related namespaces but available to other functions, but because it is related to address I think that maybe status-im.utils.ethereum.core would be more fitting (it already has functions related to address).

kfirmanty:
I have cleaned up code due to @rasom suggestions. I left find-contact-by-address and contact-by-address in utils.contacts because they are working on contacts but moved address= to utils.ethereum namespace.

src/status_im/contact/subs.cljs
rasom: wow sorry @kfirmanty, it seems that I hadn't woke up fully and started to comment removed code...

rasom (approved):

src/status_im/contact/subs.cljs
kfirmanty: no problem ;) I think that your concerns were valid so I commited cleaned up code as a part of this pull request.

src/status_im/utils/contacts.cljs
```diff
@@ -1,7 +1,8 @@
 (ns status-im.utils.contacts
```
yenda, line 1: I wonder if this namespace can be merged easily into contacts.core without circular dependencies. but that's not really the point of this PR so it's ok for me. This namespace should disappear in the future though

yenda (approved):

src/status_im/utils/contacts.cljs
kfirmanty: It is probably just a matter of convention but if there will be more of functions like
find-contact-by-address, find-contact-by-name etc. then maybe I would think about creating separate contacts.queries namespace, because for me utils are mostly created in case when you can't change source code of some namespace/class in library that you use and you want to extend it (like for example often is the case with String class in Java when classes like StringUtils are created), but again this is a matter of convention and I did not want to change too much because I don't have enough knowledge about whole project.

yenda: yes the convention for that kind of function is to put it in a db namespace

asemiankevich:
@kfirmanty please rebase the branch. Sorry for delay!

asemiankevich:
@kfirmanty would you mind to check why the recipient is not yet shown? 


image


STEPS:
- open status on 1 device and go to wallet -> receive transaction
- open status on 2 device and go to wallet -> send transaction
- on device 2 scan the qr code from device 1 and pay attention to the recipient field

kfirmanty:
@asemiankevich are accounts added as contacts? Because my solution only searches in known contacts for name (as this was a test case in #2884). I tested and it works on my device in such case. When accounts are not added as contacts then recipient name will be empty.

asemiankevich:
true, this is shown when devices are in contacts
image

okay lets merge it

goranjovic:
@kfirmanty Can you please squash the commits into one, so that I can merge the PR

StatusSceptre:
@kfirmanty Looks like this one got looked by! Thanks for the patience and work; can you re-apply for the bounty on gitcoin so I can  pay this one out?

PR #6383, opened 2018-10-18
[TESTS] fix false failed test

hetvart:
### Review notes:
~~- added scrolling to about button in `test_privacy_policy_is_accessible`~~ 
- removed critical mark from `test_can_see_all_transactions_in_history` - fix for this test requires more time, so it won't run against PRs for a while.

### DIFF
--- a/test/appium/tests/atomic/transactions/test_wallet.py
+++ b/test/appium/tests/atomic/transactions/test_wallet.py
@@ -365,7 +365,6 @@ class TestTransactionWalletSingleDevice(SingleDeviceTestCase):
         self.network_api.find_transaction_by_unique_amount(sender['address'], amount)
 
     @marks.testrail_id(5314)
-    @marks.critical
     def test_can_see_all_transactions_in_history(self):
         address = transaction_senders['W']['address']
         passphrase = transaction_senders['W']['passphrase']


test/appium/tests/test_general_verifications.py
```diff
@@ -24,6 +24,7 @@ def test_privacy_policy_is_accessible(self):
         signin_view = SignInView(self.driver)
         home_view = signin_view.create_user()
         profile = home_view.profile_button.click()
+        profile.about_button.scroll_to_element()
```
yevh-berdnyk: Scroll is already added here: https://github.com/status-im/status-react/blob/develop/test/appium/views/profile_view.py#L399

hetvart: @yevh-berdnyk 
@antdanchenko 
to avoid such cases when need to scroll to an element we can move scrolling to base button element level (under click() method).

see `base_element.py -> BaseButton -> click()`

```python
def click(self):
    self.scroll_to_element().click()
    self.driver.info('Tap on %s' % self.name)
    return self.navigate()
``` 

This will allow us not to override click() method for every button which requires scrolling.
WDYT?

yevh-berdnyk:
We can try but I’m afraid of increasing tests run time because of
increasing time for searching elements

test/appium/tests/test_general_verifications.py
antdanchenko: I don't think that we need the scroll for each element since we can miss UX issues

antdanchenko (approved):

test/appium/tests/test_general_verifications.py
hetvart: so, I am removing duplicated scroll in test and merging this PR if no objections.

yevh-berdnyk (approved):

PR #6389, opened 2018-10-18
Makes extensions dev-mode aware

### LINKED ISSUE #6382: Makes extensions dev-mode aware
What is expected when installing an extension:
• install an extension, activated by default, command present
• swith off dev mode, command not present
• restart status, command not present
• enable dev mode, command present

Also make sure that an extension cannot be added twice.

alwx:
Fixes #6382 
Status: ready

### DIFF
--- a/src/status_im/accounts/core.cljs
+++ b/src/status_im/accounts/core.cljs
@@ -41,13 +41,26 @@
             (wallet.settings.models/wallet-autoconfig-tokens)
             (accounts.update/account-update {:wallet-set-up-passed? true} {})))
 
-(fx/defn switch-dev-mode [cofx dev-mode?]
-  (merge (accounts.update/account-update cofx
-                                         {:dev-mode? dev-mode?}
-                                         {})
-         (if dev-mode?
-           {:dev-server/start nil}
-           {:dev-server/stop nil})))
+(fx/defn update-dev-server-state
+  [_ dev-mode?]
+  (if dev-mode?
+    {:dev-server/start nil}
+    {:dev-server/stop nil}))
+
+(fx/defn update-extensions-state
+  [{{:account/keys [account]} :db} dev-mode?]
+  (let [extensions (-> account :extensions vals)]
+    (if dev-mode?
+      {:extensions/load {:extensions extensions
+                         :follow-up  :extensions/add}}
+      {:dispatch [:extensions/deactivate-all (filter :active? extensions)]})))
+
+(fx/defn switch-dev-mode [{{:account/keys [account]} :db :as cofx} dev-mode?]
+  (fx/merge cofx
+            (update-dev-server-state dev-mode?)
+            (update-extensions-state dev-mode?)
+            (accounts.update/account-update {:dev-mode? dev-mode?}
+                                            {})))
 
 (fx/defn switch-web3-opt-in-mode [{:keys [db] :as cofx} opt-in]
   (let [settings (get-in db [:account/account :settings])]
--- a/src/status_im/ui/screens/extensions/add/events.cljs
+++ b/src/status_im/ui/screens/extensions/add/events.cljs
@@ -1,23 +1,32 @@
 (ns status-im.ui.screens.extensions.add.events
-  (:require [re-frame.core :as re-frame]
+  (:require [pluto.registry :as registry]
+            [re-frame.core :as re-frame]
             [status-im.extensions.core :as extensions]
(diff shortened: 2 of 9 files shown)


jeluard (approved):

jeluard:
@alwx Is that still WIP?

flexsurfer (approved):

src/status_im/accounts/core.cljs
```diff
@@ -41,13 +41,16 @@
...
+(fx/defn switch-dev-mode [{{:account/keys [account]} :db :as cofx} dev-mode?]
+  (let [extensions (-> account :extensions vals)]
+    (merge (accounts.update/account-update cofx
+                                           {:dev-mode? dev-mode?}
+                                           {})
+           (if dev-mode?
+             {:dev-server/start        nil
+              :extensions/activate-all extensions}
+             {:dev-server/stop           nil
+              :extensions/deactivate-all (filter :active? extensions)}))))
```
janherich: We shouldn't use plain merge but `fx/merge` to merge static effects map/effects producing functions as `fx/merge` contains important checks (no silent overwriting of effects, accumulation of certain effects, etc.)

src/status_im/ui/screens/extensions/add/events.cljs
```diff
@@ -1,23 +1,41 @@
...
  (fn [cofx [_ data active?]]
-   (extensions/add cofx data active?)))
\ No newline at end of file
+   (extensions/add cofx data active?)))
+
+(re-frame/reg-fx
+ :extensions/activate-all
+ (fn [extensions]
+   (doseq [{:keys [url active?]} extensions]
+     (extensions/load-from url #(re-frame/dispatch [:extensions/add (-> % extensions/read-extension extensions/parse :data) active?])))))
```
janherich: Ideally, we should make `extensions/load` effects plural (accepting collection of `[url follow-up-event]` tuples) so the whole `:extensions/activate-all` would not be necessary (just producing the effect in `activate-extensions` fn will be enough)

src/status_im/ui/screens/extensions/add/events.cljs
```diff
@@ -1,23 +1,41 @@
...
+(handlers/register-handler-fx
+ :extensions/deactivate
+ (fn [cofx [_ name]]
+   (fx/merge cofx #(registry/deactivate name %))))
+
+(re-frame/reg-fx
+ :extensions/deactivate-all
+ (fn [extensions]
+   (doseq [{:keys [name]} extensions]
+     (re-frame/dispatch [:extensions/deactivate name]))))
```
janherich: The combination of this event and fx is unnecessary, the same can we achieved by just:
```clj
(apply fx/merge cofx (map (fn [{:keys [name]}] 
                            (partial registry/deactivate name)) 
                          extensions))
```
Probably worth extracting it to public fx producing function and any other model function can just call it and merge it into its own effects when deactivation of `extensions` collection is necessary.

janherich (changes requested):

src/status_im/ui/screens/extensions/add/events.cljs
```diff
@@ -1,23 +1,32 @@
...
+ :extensions/add
+ (fn [cofx [_ {:keys [data]} active?]]
+   (extensions/add cofx data active?)))
+
+(handlers/register-handler-fx
+ :extensions/deactivate-all
+ (fn [cofx [_ extensions]]
+   (apply fx/merge cofx (map (fn [{:keys [name]}]
+                               (partial registry/deactivate name))
+                             extensions))))
```
janherich, line 32: Great work, in incorporating the feedback, just one small last thing, we don't need event `:extensions/deactival-all` at all -> instead of making it event, please make it a named model function in `extensions.core` ns and in the single place where it's dispatched (function `update-extensions-state` in accounts), call the function to produce map of effects instead of dispatching event.
If the above can't be done due to circular dependencies, then ignore it for now and I will refactor ns structure and take care of that later.

janherich (approved):

Serhy:
Looks good to me. Latest builds from https://github.com/status-im/status-react/pull/6389#issuecomment-431348974 were tested on iOS 12.01, Android 7, Android 6.0.1
• install an extension, activated by default, command present PASS
• swith off dev mode, command not present PASS
• restart status, command not present PASS
• enable dev mode, command present PASS
• extension cannot be added twice PASS
• two extensions can be added and executing the same items above PASS
Ready for merge

PR #6391, opened 2018-10-18
Fixes set-extension-url-from-qr: updates it for the new manage object

alwx:
Makes it possible to set extension url from QR code. 
It doesn't fix any problems with universal links — only makes it possible to use a QR scanner on add extension screen.


jeluard (approved):

flexsurfer (approved):

asemiankevich:
looks good. 
- i can scan extension unversal link 
- other universal links (chat, profile, dapp) are working as expected

PR #6405, opened 2018-10-19
[extensions] ethereum events

flexsurfer:
`ethereum/send-transaction` and `ethereum/call` events support

```
[button {:label "send value transaction" :on-click [ethereum/send-transaction {:to "0xdfb0c49664b4e5e14d9c54d563d8f8c9add7d4a3" :value val :on-result [tx-result]}]}]
    [button {:label "send data transaction" :on-click [ethereum/send-transaction {:to "0xc032537f648a7359076aad96b115c232530c2ff6" :method "set(uint256)" :params [20] :on-result [tx-result]}]}]
    [button {:label "call get" :on-click [ethereum/call {:to "0xc032537f648a7359076aad96b115c232530c2ff6" :method "get()" :on-result [tx-result]}]}]])
```

Example extension: https://get.status.im/extension/ipfs@QmVkdXbeXaidjc15mhT36vj6DD1qDkghATcy3QtGnxZGQ2

<blockquote></blockquote>

For QA: transactions from dapps should be tested

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -16,13 +16,14 @@
             [status-im.ui.components.colors :as colors]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.handlers :as handlers]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            status-im.extensions.ethereum))
 
 (re-frame/reg-fx
  ::alert
  (fn [value] (js/alert value)))
 
-(re-frame/reg-event-fx
+(handlers/register-handler-fx
  :alert
  (fn [_ [_ {:keys [value]}]]
    {::alert value}))
@@ -31,7 +32,7 @@
  ::log
  (fn [value] (js/console.log value)))
 
-(re-frame/reg-event-fx
+(handlers/register-handler-fx
  :log
  (fn [_ [_ {:keys [value]}]]
    {::log value}))
@@ -148,7 +149,7 @@
                 'transaction-status {:value transactions/transaction-status :properties {:outgoing :string :tx-hash :string}}
                 'asset-selector     {:value transactions/choose-nft-asset-suggestion}
                 'token-selector     {:value transactions/choose-nft-token-suggestion}}
-   :queries    {'store/get {:value :store/get :arguments {:key :string}}
+   :queries    {'store/get           {:value :store/get :arguments {:key :string}}
                 'wallet/collectibles {:value :get-collectible-token :arguments {:token :string :symbol :string}}}
    :events     {'alert
                 {:permissions [:read]
@@ -191,30 +192,25 @@
                  :arguments   {:hash        :string
                                :on-success  :event
                                :on-failure? :event}}
-                'ethereum/sign
-                {:arguments
-                 {:account   :string
-                  :message   :string
-                  :on-result :event}}
                 'ethereum/send-transaction
-                {:arguments
-                 {:from       :string
-                  :to         :string
-                  :gas?       :string
-                  :gas-price? :string
-                  :value?     :string
-                  :data?      :string
-                  :nonce?     :string}}
+                {:permissions [:read]
+                 :value       :extensions/ethereum-send-transaction
+                 :arguments   {:to         :string
+                               :gas?       :string
+                               :gas-price? :string
+                               :value?     :string
+                               :method?    :string
+                               :params?    :vector
+                               :nonce?     :string
+                               :on-result? :event}}
                 'ethereum/call
-                {:arguments
-                 {:from?      :string
-                  :to         :string
-                  :gas?       :string
-                  :gas-price? :string
-                  :value?     :string
-                  :data?      :string
-                  :block      :string}}}
-   :hooks {:commands commands/command-hook}})
+                {:permissions [:read]
+                 :value       :extensions/ethereum-call
+                 :arguments   {:to         :string
+                               :method     :string
+                               :params?    :vector
+                               :on-result? :event}}}
+   :hooks      {:commands commands/command-hook}})
 
 (defn read-extension [{:keys [value]}]
   (when (seq value)
--- a/src/status_im/extensions/ethereum.cljs
+++ b/src/status_im/extensions/ethereum.cljs
@@ -1,0 +1,39 @@
+(ns status-im.extensions.ethereum
+  (:require [status-im.utils.handlers :as handlers]
+            [re-frame.core :as re-frame]
+            [status-im.models.wallet :as models.wallet]
+            [status-im.utils.ethereum.abi-spec :as abi-spec]
+            [status-im.utils.fx :as fx]
+            [status-im.ui.screens.navigation :as navigation]))
+
+(handlers/register-handler-fx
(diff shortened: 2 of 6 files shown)


src/status_im/models/wallet.cljs
```diff
@@ -216,3 +242,17 @@
                                (clear-error-message :balance-update)
                                (assoc-in [:wallet :balance-loading?] true)
                                (assoc :prices-loading? true))})))
+
+(defn open-modal-wallet-for-trasnaction [db transaction tx-object]
```
jeluard: Typo trasnaction

src/status_im/models/wallet.cljs
```diff
@@ -157,24 +183,24 @@
...
+                             (assoc-in [:wallet :transactions-queue] nil)
+                             (assoc-in [:wallet :send-transaction] {}))
+                         :wallet/show-transaction-error
+                         message}]
+               (navigation/navigate-back cofx))
 
-        dapp-transaction
-        (web3-error-callback db dapp-transaction message)))))
+             (when on-error
+               {:dispatch (conj on-error "discarded")})))))
```
jeluard: What's "discarded"?

src/status_im/ui/screens/wallet/send/events.cljs
```diff
@@ -126,19 +135,8 @@
...
-                         (when-not gas
-                           [:wallet/update-estimated-gas (first params)])
-                         (when-not gas-price
-                           [:wallet/update-gas-price])
-                         [:navigate-to
-                          (if wallet-set-up-passed?
-                            :wallet-send-modal-stack
-                            :wallet-send-modal-stack-with-onboarding)]]})
+         (let [transaction (models.wallet/prepare-dapp-transaction queued-transaction (:contacts/contacts db))]
+           (models.wallet/open-modal-wallet-for-trasnaction db' transaction (first params)))
```
jeluard: Typo

jeluard (approved):

src/status_im/extensions/ethereum.cljs
```diff
@@ -0,0 +1,37 @@
...
+            [status-im.models.wallet :as models.wallet]
+            [status-im.utils.ethereum.abi-spec :as abi-spec]))
+
+(handlers/register-handler-fx
+ :extensions/transaction-on-result
+ (fn [{db :db} [_ on-result id result method]]
+   {:dispatch-n (concat
+                 [[:navigate-to-clean :wallet-transaction-sent]]
+                 (when on-result
+                   [(on-result {:error nil :result result})]))}))
```
janherich: I would much prefer to use:
```clj
(fx/merge cofx
          (when on-result
            {:dispatch (on-result {:error nil :result result}))}
          (navigation/navigate-to-clean :wallet-transaction-sent))
```

Using `:dispatch-n` is generally dangerous, especially when used to encode successions of steps, because it relies on implementation detail of `re-frame` and not interface, interface says there is no particular order in which the events will be dispatched.

Using some unnecessary events is unfortunately always necessary with extensions, because it was decided that event is the only logic building block exposed to the extension writer, but we should still strive to keep it minimal.

src/status_im/models/wallet.cljs
```diff
@@ -216,3 +242,17 @@
...
+    {:db         (assoc-in db [:wallet :send-transaction] transaction)
+     :dispatch-n [[:update-wallet]
+                  (when-not gas
+                    [:wallet/update-estimated-gas tx-object])
+                  (when-not gas-price
+                    [:wallet/update-gas-price])
+                  [:navigate-to
+                   (if wallet-set-up-passed?
+                     :wallet-send-modal-stack
+                     :wallet-send-modal-stack-with-onboarding)]]}))
```
janherich, line 258: Please encode logic with `fx/merge` macro and model functions instead of `:dispatch-n`, there is a `status-im.models.wallet/update-wallet` fn already for example.
In worst case, when we don't want to do any refactoring in the PR and some functionality is only exposed via events, we can at least do:
```clj
(fx/merge cofx
          {:dispatch-n collection-of-events-which-have-no-model-functions-yet}
          model-fn-1
          model-fn-2
          ...)
```

By doing that, it will be clear that we are trying to rid of the `:dispatch-n` pattern + it will be super easy to actually erase it completely once those events are refactored into model functions.

janherich (changes requested):

src/status_im/models/wallet.cljs
flexsurfer: we need some error message when a user cancelled a transaction, `"transaction was cancelled by user"` ?

src/status_im/extensions/ethereum.cljs
flexsurfer: thanks, i'll fix that, but you could suggest change next time, using new github feature :)

src/status_im/models/wallet.cljs
flexsurfer: yeah, i expected such comment , you're right, but i don't want to made such change in this PR, let me create separate PR for this

flexsurfer:
@jeluard @janherich done

src/status_im/models/wallet.cljs
janherich: Sure, makes sense !

src/status_im/extensions/ethereum.cljs
janherich: Sorry, didn't know about that really, will try it :)

src/status_im/extensions/core.cljs
```diff
@@ -140,15 +141,15 @@
                 'touchable-opacity  {:value touchable-opacity :properties {:on-press :event}}
                 'image              {:value image :properties {:uri :string}}
                 'input              {:value input :properties {:on-change :event :placeholder :string}}
-                'button             {:value button :properties {:on-click :event}}
+                'button             {:value button :properties {:on-click :event :label :string}}
```
jeluard: Actually the label should be a child, not an attribute.

jeluard (changes requested):

src/status_im/extensions/ethereum.cljs
```diff
@@ -0,0 +1,39 @@
...
+            [status-im.utils.fx :as fx]
+            [status-im.ui.screens.navigation :as navigation]))
+
+(handlers/register-handler-fx
+ :extensions/transaction-on-result
+ (fn [cofx [_ on-result id result method]]
+   (fx/merge cofx
+             (when on-result
+               {:dispatch (on-result {:error nil :result result})})
+             (navigation/navigate-to-clean :wallet-transaction-sent))))
```
alwx: are you sure that the number of args provided to this function is correct? it expects 3 afaik, but receives only 2 in this case

alwx (changes requested):

jeluard (approved):

alwx (approved):

janherich (approved):

asemiankevich:
- all transactions to test dapp are fine
- mainnet dapp transactions are fine (cryptokitties look fine)

https://get.status.im/extension/ipfs@QmUk8SD8K3xHsXD5CG6pauxUMRuTCtN14dYpn2o4Xao2mg gives error
image


<blockquote></blockquote>

PR #6448, opened 2018-10-22
Fix old command message id calculation

### LINKED ISSUE #6438: Duplicate request messages in 1:1 chat after upgrade from 0.9.29
### Description

*Type*: Bug

*Summary*: 
Before upgrade there is one request for ETH and one for SNT. After upgrade both requests are shown twice. Before upgrade user switched from mainnet (where requests are sent) to ropsten.

After upgrade:
after

#### Expected behavior
After upgrade same messages are shown. There are no extra messages
#### Actual behavior
each received request message is shown twice 
### Reproduction

- On Android: install 0.9.29 from PlayStore or from apk https://drive.google.com/open?id=1PHnITeXCp3DZGhmhq0SMhjTgcOoUDBe8 (it's same apk as on PlayStore)
- On iOS install 0.9.29 from TestFlight
- both are on mainnet
- iOS: add Android to contacts by scanning QR code and send message "hi from ios"
- Android: open 1:1 chat with iOS, add it to the contacts and send message "hi from android"
- android: send request of 1 ETH in 1:1 chat
- iOs: send request of 2 ETH in 1:1 chat
- iOS: send request of 333 SNT in 1:1 chat
- iOs: send text message and emoji in the same line
- Android: send text message "Thanks"
- Android: switch to Ropsten
- Android: check that you can login and messages look ok (no duplicates)
- Android: upgrade to release build 
- Android: login, wait for messages to be fetched and check messages in 1:1 chat. In my case there are 2 new messages shown for chat, if open chat then each of request messages from iOS is shown twice

### Additional Information
* Status version: upgrade from 0.9.29 to release build 15
* Operating System: Android

#### Logs

[logcat.txt](https://github.com/status-im/status-react/files/2500485/logcat.txt)

janherich:

fixes #6438

### Steps to test:
Make sure that #6438 is not happening anymore

status: ready

### DIFF
--- a/src/status_im/transport/message/protocol.cljs
+++ b/src/status_im/transport/message/protocol.cljs
@@ -105,8 +105,15 @@
            this)
           (send-with-pubkey cofx params)))))
   (receive [this chat-id signature _ cofx]
-    (let [old-message (Message. (:text content) content-type message-type
-                                clock-value timestamp)]
+    (let [old-content      (if (or (= content-type constants/content-type-command)
+                                   (= content-type constants/content-type-command-request))
+                             (dissoc content :command-path)
+                             (:text content))
+          old-content-type (if (= ["request" #{:personal-chats}] (:command-path content))
+                             constants/content-type-command-request
+                             content-type)
+          old-message      (Message. old-content old-content-type message-type
+                                     clock-value timestamp)]
       {:chat-received-message/add-fx
        [(assoc (into {} this)
                :message-id (transport.utils/message-id this)


dmitryn (approved):

src/status_im/transport/message/protocol.cljs
```diff
@@ -105,8 +105,15 @@
            this)
           (send-with-pubkey cofx params)))))
   (receive [this chat-id signature _ cofx]
-    (let [old-message (Message. (:text content) content-type message-type
-                                clock-value timestamp)]
+    (let [old-content      (if (or (= content-type constants/content-type-command)
+                                   (= content-type constants/content-type-command-request))
+                             (dissoc content :command-path)
+                             (:text content))
+          old-content-type (if (= ["request" #{:personal-chats}] (:command-path content))
```
jeluard, line 112: Could you add a comment about what this is about?

janherich: It's about reconstructing the old content-type for message-id calculation, which is consistent with the old format.
It's necessary, because at the system boundary, we convert message with the old format to new format, therefore changing their content (and so id which is derived from sha of serialised message).
Unfortunately, calculation of that id happens **after** those coercions, that's why even the same messages can have different calculated id depending on version of app it's run on, leading to duplicated messages.
Best solution (besides changing the format of the message-id altogether) would be to calculate message id **before** any coercions take place, and we wouldn't need to do such shenanigans, but that would require much more extensive code changes

goranjovic (approved):

rasom (approved):

ghost:
### Pull Request Checklist

annadanchenko:
@janherich unfortunately, both issues with duplicated messages (https://github.com/status-im/status-react/issues/6438 and https://github.com/status-im/status-react/issues/6460) can be reproduced on this PR's build 2.

[logcat.txt](https://github.com/status-im/status-react/files/2505361/logcat.txt)

rasom:
is this PR still relevant? another one was opened

janherich:
@rasom not really, closing it for now

PR #6466, opened 2018-10-22
[#6460] attempt to fix message duplication

### LINKED ISSUE #6460: Received messages duplicated in public chat after upgrade 0.9.29 -> 0.9.30 if contact was in Ropsten prior upgrade
***Type***: Bug
***Summary***: Received messages in public channel duplicated after contact upgrade 0.9.29 Release app version to 0.9.30 when:
- user joins "active" public channel in Mainnet (e.g. `#status`)
- user switched to Ropsten prior upgrade
- user scrolls to the top of public channel after upgrade (the very latest messages in chat are not duplicated - but the deeper in history user scrolls more duplicates for particular message is created.)

#### Expected behavior
No duplicates for messages in public channel after the upgrade on 0.9.30 app version.

#### Actual behavior
Duplicated received messages in public chat after upgrade 0.9.29 -> 0.9.30 if contact was in Ropsten prior upgrade

screen shot 2018-10-22 at 17 32 36

### Reproduction

- Install Status 0.9.29 Release version
- Create new account
- Join `#status` public channel
- Switch to Ropsten RPC network
- Upgrade to Status 0.9.30 version (latest nightly/release)
- Log in to account where `#status` public channel created
- Navigate to public chat and scroll to the very top of chat. Pay attention received message in public chat
 

### Additional Information
* Status version: upgrade from 0.9.29 -> 0.9.30

rasom:
fix #6460

status: ready

### DIFF
--- a/src/status_im/transport/message/protocol.cljs
+++ b/src/status_im/transport/message/protocol.cljs
@@ -77,7 +77,7 @@
                                         :topic   (transport.utils/get-topic constants/contact-discovery)}
                                        whisper-opts)}]}))
 
-(defrecord Message [content content-type message-type clock-value timestamp]
+(defrecord Message [content content-type message-type clock-value timestamp legacy-data]
   StatusMessage
   (send [this chat-id cofx]
     (let [params     {:chat-id       chat-id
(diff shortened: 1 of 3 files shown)


janherich (approved):

rasom:
@yenda @janherich @cammellos  this PR doesn't fix anything... I will be traveling today, so it would be great if someone will continue with this task

rasom:
probably because i added `legacy-data` prop and now ids are broken again

janherich:
@rasom I think it still won't work, because when you dissoc something from clojure record, you get plain map back and I think plain map/records are serialised differently.
I already have quite clean solution almost ready, so if your won't work correctly, I will fix it today.

lukaszfryc:
@rasom @janherich would it be helpful for you guys if I start testing it now? We need to finish this PR today so we could include it in the hackaton release.

janherich:
@lukaszfryc yes, totally

lukaszfryc:
OK, I'll test it now while you continue working on it

rasom:
@janherich i actually checked in repl and the only difference between amp and record was namespace prefix 
```
#status-im.transport.message.protocol.Message{:content nil, :content-type nil, :message-type nil, :clock-value nil, :timestamp nil}
```
vs
```
{:content nil, :content-type nil, :message-type nil, :clock-value nil, :timestamp nil}
```

but anyway, it doesn't work

rasom:
we also can just hash the map with some specific keys for a new id, then namespace will not matter in future

lukaszfryc:
@rasom yes, it does not work. The issuse is still present.

lukaszfryc:
@rasom any update?

lukaszfryc:
@annadanchenko could you test it when it's ready? I'll be afk for next couple of hours.

PR #6469, opened 2018-10-22
[6454] dont clear chat history on deletion

### LINKED ISSUE #6454: Public chat that was re-opened should not be empty
### Description

*Type*: Bug

*Summary*: 

If a public chat is removed from the chat list and then reopened again, its content is empty. There is "Fetching messages" but the old messages (before removing the chat) do not appear. I think it's counter-intuitive and the app should display old messages when public chat is re-opened. 
I'd say old messages should be displayed even though the user has cleared the chat before removing it. So basically joining/re-opening a public chat should fetch all messages.   

### Reproduction

- Join public chat with some messages
- Remove it
- Join the same public chat again
- User should see historical messages (currently 7 days of history) which is not the case

### Additional Information
* Status version: Oct 22, nightly

bhauman:
fixes #6454  

### Summary:

Summary:

If a public chat is removed from the chat list and then reopened again, its content is empty. There is "Fetching messages" but the old messages (before removing the chat) do not appear. I think it's counter-intuitive and the app should display old messages when public chat is re-opened.
I'd say old messages should be displayed even though the user has cleared the chat before removing it. So basically joining/re-opening a public chat should fetch all messages.

Reproduction
Join public chat with some messages
Remove it
Join the same public chat again
User should see historical messages (currently 7 days of history) which is not the case

### Review notes (optional):
This is my first PR on a issue on status-react. So I'm not familiar with all the ramifications here but this appears to solve it. 

It seems like `clear-history` is miss-used here.

status: ready

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -122,7 +122,6 @@
             #(when (public-chat? % chat-id)
                (transport.chat/unsubscribe-from-chat % chat-id))
             (deactivate-chat chat-id)
-            (clear-history chat-id)
             (navigation/navigate-to-cofx :home {})))
 
 (fx/defn send-messages-seen


bhauman:
The real problem is that clear-history changes the `:deleted-at-clock-value` and that is not what is called for in this case.  However it does not seem necessary to actually delete the messages at all so, just skip it. 

If clean up is actually needed then it seems that we just need to call `messages-store/delete-messages-tx` on the chat id not move the `:deleted-at-clock-value`

chadyj:
Congrats on your first PR!  

I left a comment over in the issue https://github.com/status-im/status-react/issues/6454#issuecomment-432121014 that might be worth discussing.

bhauman:
According to the discussion over on https://github.com/status-im/status-react/issues/6454#issuecomment-432121014

Not clearing the history is not the solution here I do think that simply deleting the messages and not moving the `deleted-at-clock-value` meets both the security needs and UX needs.

bhauman:
So I'm going to close this PR and go back to the drawing board. And I'll fix the tests on that one ;)

PR #6470, opened 2018-10-22
Support of java 8+ versions

maxhora:
Running `lein` build commands with openjdk `10.0.2` installed failures with output similar to:
```
Exception in thread "main" java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter, compiling:(org/httpkit/server.clj:1:1)
...
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.DatatypeConverter
``` 
Updating of figwheel-sidecar to 0.5.17 resolves above issue

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -23,7 +23,7 @@
          ;; Figwheel ClojureScript REPL
          cider/piggieback        {:mvn/version "0.3.9"
                                   :exclusions  [com.google.javascript/closure-compiler]}
-         figwheel-sidecar        {:mvn/version "0.5.16"
+         figwheel-sidecar        {:mvn/version "0.5.17"
                                   :exclusions  [com.google.javascript/closure-compiler]}
          re-frisk-remote         {:mvn/version "0.5.5"}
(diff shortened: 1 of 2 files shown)


rasom (approved):

jeluard (changes requested):
Do we need this dependency at all? It's probably coming from transitive dependency.
Also project.clj should always be updated alongside deps.edn.

jeluard:
@MaxRis Where is this dep coming from? Do we use it? My guess is that it should be removed

maxhora:
@jeluard seems like `figwheel-sidecar-0.5.16` has dependency on it:

```
<dependency>
      <groupId>http-kit</groupId>
      <artifactId>http-kit</artifactId>
      <version>2.2.0</version>
    </dependency>
```
I guess, need to check for the newer version of `figwheel-sidecar`

maxhora:
@jeluard updating of figwheel-sidecar to 0.5.17 helped to resolve the issue, thanks

maxhora:
@jeluard could you please take a look again, thanks

PR #6471, opened 2018-10-23
WIP - Move coercion to place after id computation

janherich:

Alternative attempt at fixing #6466 - works by moving the receive coercion (compatibility changes needed to convert older message data to newest message format which app can understand) to place after message-id calculation, so the message-id is always calculated from raw, original received data. 
The only reason why `:message-id-old-format` is necessary anymore is the fact that record namespaces were changed.

status: WIP

### DIFF
--- a/src/status_im/transport/message/protocol.cljs
+++ b/src/status_im/transport/message/protocol.cljs
@@ -77,6 +77,32 @@
                                         :topic   (transport.utils/get-topic constants/contact-discovery)}
                                        whisper-opts)}]}))
 
+(def ^:private legacy-ref->new-path
+  {["transactor" :command 83 "send"]    ["send" #{:personal-chats}]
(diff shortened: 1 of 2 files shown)


annadanchenko:
@janherich I'm going to test https://github.com/status-im/status-react/pull/6448 Please, let me know if this PR overwrites or modifies logic for #6448 and if it makes sense to have it tested

janherich:
@annadanchenko This PR is just another attempt, currently not finished/compiling, so please proceed with #6448, if it fixes the problem, then I will close this PR.

annadanchenko:
#6448 didn't fix the problem @janherich

annadanchenko:
@janherich build for this PR failed, are you fixing it?

janherich:
Will be solved in different way

PR #6499, opened 2018-10-23
dont delete public chat history in app-db

### LINKED ISSUE #6454: Public chat that was re-opened should not be empty

bhauman:
fixes #6454 

### Summary:

**From issue**
If a public chat is removed from the chat list and then reopened again, its content is empty. There is "Fetching messages" but the old messages (before removing the chat) do not appear. I think it's counter-intuitive and the app should display old messages when public chat is re-opened.
I'd say old messages should be displayed even though the user has cleared the chat before removing it. So basically joining/re-opening a public chat should fetch all messages.

**This PR stops treating public and non public chats the same**, when you delete a public chat it doesn't clear the history.  It does delete the public chat messages from the local persistent store as an added layer of security.

The result is a balance of security and you get the needed behavior that if you delete a public chat and rejoin you are not presented with an empty chat window.

**Functional**
- 1-1 chats
- public chats
- group chats

### Steps to test:
- Open Status

Public chats
* Join public chat with some messages
* hit delete chat
* Join the same public chat again
* User should see historical messages (currently 7 days of history) which is not the case

1-1 chats
* initiate a 1-1 chat
* type a few messages
* hit delete chat
* initiate the same 1-1 chat with the same person
* the previous messages should not be seen

status: ready

### DIFF
--- a/src/status_im/data_store/chats.cljs
+++ b/src/status_im/data_store/chats.cljs
@@ -81,12 +81,11 @@
      (update chat :membership-updates marshal-membership-updates)
      true)))
 
-;; Only used in debug mode
 (defn delete-chat-tx
   "Returns tx function for hard deleting the chat"
   [chat-id]
   (fn [realm]
-    (core/delete realm (core/get-by-field realm :chat :chat chat-id))))
+    (core/delete realm (core/get-by-field realm :chat :chat-id chat-id))))
 
 (defn- get-chat-by-id [chat-id realm]
   (core/single (core/get-by-field realm :chat :chat-id chat-id)))
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -122,7 +122,9 @@
             #(when (public-chat? % chat-id)
                (transport.chat/unsubscribe-from-chat % chat-id))
             (deactivate-chat chat-id)
-            (clear-history chat-id)
+            #(if-not (public-chat? % chat-id)
+               (clear-history % chat-id)
+               {:data-store/tx [(messages-store/delete-messages-tx chat-id)]})
             (navigation/navigate-to-cofx :home {})))
 
 (fx/defn send-messages-seen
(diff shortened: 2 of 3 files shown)


bhauman:
This PR may not be needed or a priority.  But I noticed this behavior myself and I thought this would be a good issue to start me on my way into the react-status codebase. As such it has been super informative.

I just started looking that status-react codebase yesterday so keep that in mind if you are reviewing this. I will appreciate any knowledge you can drop on me.

src/status_im/data_store/chats.cljs
```diff
@@ -81,12 +81,11 @@
      (update chat :membership-updates marshal-membership-updates)
      true)))
 
-;; Only used in debug mode
 (defn delete-chat-tx
   "Returns tx function for hard deleting the chat"
   [chat-id]
   (fn [realm]
-    (core/delete realm (core/get-by-field realm :chat :chat chat-id))))
+    (core/delete realm (core/get-by-field realm :chat :chat-id chat-id))))
```
janherich, line 88: Good catch :)

janherich (approved):

rasom (approved):

lukaszfryc:
@bhauman I talked with @chadyj about it and we think it's better to figure out chat archive, clear, leave  and delete behaviour first before merging this PR. When we merge this PR, the delete chat option will differ for 1-1 and public chats. This may create even more confusion.

So, I suggest to close this PR for now. I will create a separate issue to discuss what options a chat should have (e.g. leave, clear, delete, archive) and how exactly they should work.

bhauman:
Sounds good

PR #6502, opened 2018-10-24
Auto-update provisioning profiles when building PRs and nightlies on Jenkins

mandrigin:
Background: we had an option that ignored upgrading provisioning profiles with new devices from AppStore connect for a while. That made it impossible to install nightlies on the new devices registered after we switched to Fastlane. This commit fixes it.

status: ready


rasom (approved):

fastlane/Fastfile
```diff
@@ -53,11 +53,11 @@ end
 
 # builds an ios app with ad-hoc configuration and put it
 # to "status-adhoc" output folder
-def build_ios_adhoc
+def build_ios_adhoc(readonly)
```
jakubgs, line 59: Maybe we can add a comment explaining what this actually does?
```suggestion
# readonly: only fetch existing certificates and profiles, don't generate new ones from AppStore
def build_ios_adhoc(readonly)
```

fastlane/Matchfile
```diff
@@ -3,7 +3,7 @@ git_url("git@github.com:status-im/ios-certificates.git")
 type("development") # The default type, can be: appstore, adhoc, enterprise or development
 
 # app_identifier(["tools.fastlane.app", "tools.fastlane.app2"])
-username("support@status.im") # Your Apple Developer Portal username
```
mandrigin: @rasom @jakubgs which address is the correct one now? Maybe this change should be reverted?

jakubgs: As far as i know `support@status.im` is the correct Apple ID email.

mandrigin: okay, I'll update this PR and I think we can merge it finally :-P

mandrigin:
@jakubgs @rasom I upgraded the PR, addressed the feedback and rebased it to the latest master. It would be nice to merge it if everything is okay :)

jakubgs (approved):
Awesome possum.

mandrigin:
@rasom can you update the password on Jenkins to actually match to whatever `support@status.im` has?

mandrigin:
`@status-github-bot` stop it! :)

asemiankevich:
@mandrigin @jakubgs @rasom anything here? seems it is in review for a while

mandrigin:
@asemiankevich I didn't want to (potentially) interfere with the release, so I put it on hold. I will continue next week.

PR #6503, opened 2018-10-24
[TESTS] User can remove profile picture

hetvart:
### Summary:
Verify that user can remove the profile avatar.

### Review notes:
modified a bit `edit_profile_picture()` method signature to perform two types of actions: deleting and editing

### Steps to test:

1. Open Status app
2. Go to profile and remove profile picture
3. Set profile picture from gallery
4. Refresh screen 
5. Remove profile picture
6. Refresh screen and check the profile picture

**Expected result:**
it is removed

### DIFF
--- a/test/appium/views/profile_view.py
+++ b/test/appium/views/profile_view.py
@@ -394,6 +394,12 @@ class AboutButton(BaseButton):
         return self.navigate()
 
 
+class RemovePictureButton(BaseButton):
+    def __init__(self, driver):
+        super().__init__(driver)
+        self.locator = self.Locator.text_selector('Remove current photo')
+
+
 class ProfileView(BaseView):
 
     def __init__(self, driver):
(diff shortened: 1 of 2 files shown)


yevh-berdnyk (approved):

test/appium/views/profile_view.py
```diff
@@ -502,21 +509,24 @@ def backup_recovery_phrase(self):
         self.ok_got_it_button.click()
         return recovery_phrase
 
-    def edit_profile_picture(self, file_name: str):
+    def edit_profile_picture(self, file_name: str = '', remove: bool = False):
```
antdanchenko: would be great to have add_profile_picture, and remove_profile_picture methods instead of this one

antdanchenko (changes requested):

test/appium/views/profile_view.py
hetvart: don't you think it's redundant?

antdanchenko (approved):

hetvart:
splitted the profile picture editing/deleting onto two separate methods

PR #6532, opened 2018-10-26
Add system messages to group chats

cammellos:
Add system messages to group chats when:

1) Chat is created
2) Members are added
3) Members are removed
4) name is changed

https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=227%3A320 for the designs.

I have only used text-styling (* * for bold) when styling the message.

The reason being is that styling is language-sensitive and depends on the order of subject-object-verb, which is not the same across languages.

Fixes also leaving group chat in desktop for non-admin members, and on mobile when swiping the chat.

Only admin can change the name of the chat.

It also removes `show?` as not used anymore, the migration will be done in a separate PR to avoid conflicts and rebasing.

Enables group chats in nightlies

Group chats are only enabled in development mode.

status: ready


janherich (approved):

src/status_im/group_chats/core.cljs
```diff
@@ -245,26 +243,44 @@
   "Save chat from edited profile"
   [{:keys [db] :as cofx}]
   (let [current-chat-id (get-in cofx [:db :current-chat-id])
-        new-name (get-in cofx [:db :group-chat-profile/profile :name])]
+        my-public-key (get-in db [:account/account :public-key])
+        last-clock-value (get-last-clock-value cofx current-chat-id)
+        new-name (get-in cofx [:db :group-chat-profile/profile :name])
```
goranjovic: alignment in let

src/status_im/group_chats/core.cljs
```diff
@@ -277,24 +293,78 @@
...
+                                                   creator
+                                                   members-added
+                                                   name-changed?
+                                                   members-removed]}]
+  (let [get-contact         (partial models.contact/build-contact cofx)
+        format-message      (fn [contact text clock-value]
+                              {:chat-id chat-id
+                               :content {:text text}
+                               :clock-value clock-value
+                               :from (:whisper-identity contact)})
```
goranjovic: alignment

src/status_im/group_chats/core.cljs
```diff
@@ -245,26 +243,44 @@
...
+                            :clock-value (utils.clocks/send last-clock-value)}]
     (when (valid-name? new-name)
       (fx/merge cofx
-                {:db (assoc db :group-chat-profile/editing? false)}
-                (models.chat/upsert-chat {:chat-id current-chat-id
-                                          :name new-name})))))
+                {:db (assoc db :group-chat-profile/editing? false)
+                 :group-chats/sign-membership {:chat-id current-chat-id
+                                               :from my-public-key
+                                               :events [name-changed-event]}}))))
```
goranjovic: same here, etc

goranjovic (approved):

src/status_im/group_chats/core.cljs
```diff
@@ -306,10 +376,21 @@
    sender-signature]
   (let [dev-mode? (get-in cofx [:db :account/account :dev-mode?])]
     (when (and (config/group-chats-enabled? dev-mode?)
-               (valid-chat-id? chat-id (-> membership-updates first :from)))
-      (let [previous-chat (get-in cofx [:db :chats chat-id])]
+               (valid-chat-id? chat-id (extract-creator membership-update)))
+      (let [previous-chat (get-in cofx [:db :chats chat-id])
+            all-updates (clojure.set/union (into #{} (:membership-updates previous-chat))
```
rasom: might be just `(set ...)` instead of `(into #{} ...)`

rasom (approved):

lukaszfryc:
Link to iOS build: https://ci.status.im/job/status-react/job/pull%20requests/job/PR-6532/16/artifact/pkg/StatusIm-181101-141829-403224-pr.ipa

<blockquote></blockquote>

lukaszfryc:
@cammellos 

1) When user is re-added to a group chat, the chat should be immediately shown on the user side with “{username} joined the group”. Currently, it’s only shown when a new message is received.

2. Text styling does not work. E.g. user name or group name has the same font and color as other text (“created the group”).

3. “\*{username}\* joined the group” shows in chat preview in home tab. Should be without “*” according to the design.

4. I noticed that username is not updated in system messages when user changes his name. On the other hand, the avatar icon is always being updated. I think we should update the username as well. It makes sense to do it since standard chat messages are always updated. 

5. Should we differentiate between leaving a chat vs being removed from a chat? Currently when admin removes a user, there is “{username} left the group”. It’s shown also for the user that was removed. I think there should be “{username} was removed from the group”. But, I think we can handle it in a separate GHI.

6. If admin removes the group chat, it will re-appear again for him when new message is sent. I think it's perfectly OK to leave it like that until we implement group chat admin switching.

cammellos:
thanks for testing @lukaszfryc 
2 & 3) should be displaying ok in mobile, desktop we don't currently have text-styling, but once they are styled it should display correctly, could you confirm?

4) is a similar issue, I think the best approach is to wait until @ mentions are implemented, at that point we can just swap out those names for mentions and they will update

5) Yes, currently we make no distinction, but we can address

6) Is expected, would you like to change the behavior or is ok?

If 2/3/4 we are ok with the styling for desktop and wait for mentions, I would address 1 & 5 in separate PRs, what do you reckon?

lukaszfryc:
@cammellos 

2 & 3) I tested iOS and there is no styling as described. Could you check it on your side?
4) It's not critical, but anyway I think it would be great to fix it in scope of this PR as all other messages update when usernames are changed (including avatar). Up to you :) 
5) Let's fix it in a separate GHI. I will create it.
6) I think it is OK for now. But, in the future I imagine it will not be possible to delete the chat without assigning other admin.

cammellos:
@lukaszfryc 2 & 3 you should see the text in bold, can you confirm? Until we add colors etc to text styling that's the only styling allowed. We are a bit limited to use text-styling as the styling is language-dependent and needs to go through the translation system (subject-object-verb languages have different styling etc). We could make it work for english but that would be a dead-end, as it would not work on a different language.

4 you should see the picture/avatar changing, but not the text. Fixing this would probably be quite long, and might just have to be removed once we implement mention (I believe we should have them fairly soon).

lukaszfryc:
@cammellos I see bold text only on Android. iOS shows all words in standard text:

image

Also, both Android and iOS show last message as “*{username}* joined the group” in chat preview in home tab.

Should we create a separate GHIs for those as well?

cammellos:
@lukaszfryc that's interesting does bold text work at all on IOS? in any case if you don't mind I'd rather handle this in separate PRs, keen to get this in and they don't seem show-stoppers, but I defer to you if you think otherwise.
Thanks!

lukaszfryc:
@cammellos it turns out that it's a [general issue](https://github.com/status-im/status-react/issues?utf8=✓&q=is%3Aissue+is%3Aopen+bold+ios). Feel free to merge this PR. I will create a separate GHI and link them here. Thanks!

cammellos:
@lukaszfryc thanks, will do!

lukaszfryc:
I created following issues as a follow-up:
https://github.com/status-im/status-react/issues/6598
https://github.com/status-im/status-react/issues/6600
https://github.com/status-im/status-react/issues/6603

PR #6551, opened 2018-10-29
[#6517] Fix bugged quantity on chat command preview (iOS)

### LINKED ISSUE #6517: Value divided by 1000 (1 istead of 1000) is shown in Home for last messages for /send and / request command  if value is greater than 100
### Description

*Type*: Bug
*Summary*: if last messages in chat are /send or /request command with value > 100 - it shows wrong value (divided by 1000). For example if last message is `/request 10000 STT`, in Home you can see 10 STT.

#### Expected behavior
the correct value is shown (`/request 10000 STT`)

#### Actual behavior
wr_kl

### Reproduction

- Open status
- Log in
- Send or request any amount of tokens or ETH, which is larger 100.

### Additional Information

* Status version: release  build 24 for 0.9.30
* Operating System: iOS 11.4

kwingram25:
Fixes #6517 

### Summary:
Fixes bug causing values over 1000 to be truncated on chat command short preview on iOS

### Steps to test:
- Start new chat, send /request (token) 10000
- Return to home screen
- Short preview correctly display 10,000 instead of 10

status: ready

### DIFF
--- a/src/status_im/i18n.cljs
+++ b/src/status_im/i18n.cljs
@@ -260,8 +260,7 @@
   ([path] (label path {}))
   ([path options]
    (if (exists? rn-dependencies/i18n.t)
-     (let [options (update options :amount label-number)]
-       (.t rn-dependencies/i18n (name path) (clj->js (label-options options))))
+     (.t rn-dependencies/i18n (name path) (clj->js (label-options options)))
      (name path))))
 
 (defn label-pluralize [count path & options]


goranjovic:
We probably needed `label-number` for something. Can you remember @alwx ?

src/status_im/i18n.cljs
```diff
@@ -260,8 +260,7 @@
   ([path] (label path {}))
   ([path options]
    (if (exists? rn-dependencies/i18n.t)
-     (let [options (update options :amount label-number)]
```
goranjovic, line 263: What specifically was broken in `label-number`? I'm assuming there was a good reason we had this function, we don't want another bug to pop out elsewhere after this fix. 

Does it make sense to fix `label-number` instead?

goranjovic (changes requested):

PR #6552, opened 2018-10-29
Update status-go

cammellos:
yesterday nightly -> today nightly on desktop does not upgrade cleanly because of downgrade of status-go, adding a commit to fix it, will have to rebuild nightly after merge.

status: ready


janherich (approved):

yenda (approved):

churik:
@cammellos is it any specific reason why this PR skipped manual QA?

cammellos:
@churik I had introduced a regression with a previous PR , the build system included a commit that should not have been( the issue we are having was also present in mobile), reverting should have not caused any problem, but seems like desktop & mobile are not running the same status-go version, investigating

PR #6578, opened 2018-11-02
stop using build-* git tags for counting build numbers

jakubgs:
Context: We currently use `build-[0-9]+` tags in status-react in order to globally track build numbers for iOS and Android, this is bad because it also creates release point in GitHub(5630 right now).

Issue: https://github.com/status-im/status-react/issues/6522

Options:

* Get rid of the tags and start using a different way of having unique and incrementable build numbers.
  - Problem: The size of the build number cannot be bigger than signed int, which is 2147483647
* We could use seconds since epoch
  - Problem: That would work only until around 2038, considering the INT size
* Divide the seconds since epoch by 2 and get enouch space to reach till 2106 but with 2 second granurality

I was also thinking about something like `181102048201`, which is just a short timestmap format, but int is not long enough to include seconds.

Status: DONE


jakubgs:
OK, so as far as I can tell the conflicts will only show up if we are trying to upload the release to Apple Store or Play Store, so nightlies could use clashing build numbers and it should work just fine. So I think using just days + hours would be fine for releases.

So Just using dates should be fine:
* Released on 2018/11/15 at 12:10 you get `2018111512`
* Released on 2018/12/01 at 18:05 you get `2018120118`

This should be enough for us.

flexsurfer (approved):

rasom (approved):

mandrigin (changes requested):
Great! it is long due.

Still a few questions to the implementation though:
1) it is still not clear what way exactly was chosen, I don't see it clearly stated in the docs

2) are we sure that we aren't calling this script multiple times during the build process having different build numbers in different places of the app?

jakubgs:
>it is still not clear what way exactly was chosen, I don't see it clearly stated in the docs

I thought the code was pretty self explanatory:
```bash
getNumber() {
    # Year + Month + Day + Hour
    date '+%y%m%d%H'
}
```
But if you tell me where would be a good place to document this I'd be happy to.

>are we sure that we aren't calling this script multiple times during the build process having different build numbers in different places of the app?

That's a fair point. The timestamp granularity includes hours, but it could be the case that the builds builds starts at the end of an hour and the script might be called twice to generate the timestamp for two different hours.
The only place in our codebase that makes me think that could happen is this:
https://github.com/status-im/status-react/blob/develop/src/status_im/utils/build.clj#L25
I'm not sure when it's called but it might be called in addition to the call in `ios/scripts/set_xcode_version.sh` or `android/app/build.gradle`.

I see three solutions to this:
* We drop the hours and just do days in timestamp, though that still has a possibility of happening at the end of the day.
* Store the generated build number somewhere(`./BUILD_NUMBER`?) and using that if it exists.
* Possibly use Jenkins build start timestamp, though that might be tying us too much to Jenkins.

I guess the second option is more sensible.

mandrigin:
> Store the generated build number somewhere(./BUILD_NUMBER?) and using that if it exists.

I think that is the best solution.

> `# Year + Month + Day + Hour`

Yeah, but it is nice to have some explanation in the PR description to have more context.
Btw, this will block us to update TestFlight builds more frequently than once an hour. And might give some troubles with the parallel Jenkins jobs.

I would suggest doing like that:
1) Create a separate script that will generate a build number and store it to some file
2) In this script, it should only return the contents of this file or some pre-defined value (if built locally on a dev machine)
3) Upgrade Jenkinsfiles to call the script from (1) in the beginning
4) Increase the timestamp resolution, so we don't get the same build numbers for different builds every hour.

jakubgs:
>Btw, this will block us to update TestFlight builds more frequently than once an hour. And might give some troubles with the parallel Jenkins jobs.

Oh, does TestFlight also check the build number? I was not aware of this, that changes this quite a bit.

>Create a separate script that will generate a build number and store it to some file

I see no reason for the separate script, why can't the same script just return the contents of the file if it exists and create it when it doesn't? And just return a pre-defined value if it can't detect it's on Jenkins.

>Increase the timestamp resolution, so we don't get the same build numbers for different builds every hour.

What we could do is drop the `20` from the year and end up with timestamps that include minutes, like:

* Released on 2018/11/15 at 12:10 you get 1811151210
* Released on 2018/12/01 at 18:05 you get 1812011805

That would get us to a highest date of one minute before year 2100 I guess. And by that time I'd expect the size of the build number to be increased... or humanity to be extinct.

mandrigin:
> I see no reason for the separate script

The main reason is making it harder to screw things up. If this script is only doing a readonly operation, then we don't care about how/where is it called and with which parameters. The new "mutator" script, on the other hand is new and has no references, so we will know exactly where it is called.

jakubgs:
I see, that makes sense. Okay, I'll adjust the PR, thanks man!

mandrigin (approved):
lgtm

PR #6582, opened 2018-11-04
fix to #5699 : 'Reset to default' from transaction fee screen resets gas limit to 21000 always sending tx from Dapp

### LINKED ISSUE #5699: 'Reset to default' from transaction fee screen resets gas limit to 21000 always sending tx from Dapp
***Type***: Bug

***Summary***:  `Reset to Default` button on Transaction Fee screen when sending TX in Dapp - always defaults Gas Limit value to 21000. 

#### Expected behavior
'Reset to default' from Transaction Fee screen sets Gas Limit to default value (same as when Send Transaction screen is opened) when sending tx from Dapp 

#### Actual behavior
'Reset to default' from transaction fee screen resets gas limit to 21000 always sending tx from Dapp 

### Reproduction

- Open Status and create an account and switch to Ropsten network
- Open Simple Dapp 
- `Assets` tab -> tap `Request STT` button
- Open `Tranaction Fee` screen
- Tap `Reset to default` button

### Additional Information
* Status version: Release 0.9.26, Develop 0.9.26 (7032)

### Additional Information. Issue confirmed as reproduced in:
* Status version: Release 0.9.28, Develop 0.9.29
#### Logs

TD session: https://app.testfairy.com/projects/4803622-status/builds/8573826/sessions/4399769595/?accessToken=NRQUjBaZAtpxVUZPUPpsfH6Lj9U

tbenr:
fixes #5699

Assuming that the dapp can calculate gas by itself (not sure about it), the idea is to store the original gas in the db (a new field `:original-gas` in `:send-transaction`).
When "Reset to default" is pressed, the handler checks if there is an origina-gas. if present it is used to reset the gas, if not present, gas estimation is used.

### DIFF
--- a/src/status_im/models/wallet.cljs
+++ b/src/status_im/models/wallet.cljs
@@ -223,7 +223,9 @@
 (defn open-modal-wallet-for-transaction [db transaction tx-object]
   (let [{:keys [gas gas-price]} transaction
         {:keys [wallet-set-up-passed?]} (:account/account db)]
-    {:db         (assoc-in db [:wallet :send-transaction] transaction)
+    {:db         (-> db
+                     (assoc-in [:wallet :send-transaction] transaction)
+                     (assoc-in [:wallet :send-transaction :original-gas] gas))
      :dispatch-n [[:update-wallet]
                   (when-not gas
                     [:wallet/update-estimated-gas tx-object])
--- a/src/status_im/ui/screens/wallet/events.cljs
+++ b/src/status_im/ui/screens/wallet/events.cljs
@@ -238,7 +238,11 @@
  :wallet/update-estimated-gas-success
  (fn [{:keys [db]} [_ gas]]
    (when gas
-     {:db (assoc-in db [:wallet :send-transaction :gas] (money/bignumber (int (* gas 1.2))))})))
+     (let [adjusted-gas (money/bignumber (int (* gas 1.2)))
+           db-with-adjusted-gas (assoc-in db [:wallet :send-transaction :gas] adjusted-gas)]
+       {:db (if (some? (-> db :wallet :send-transaction :original-gas))
+              db-with-adjusted-gas
+              (assoc-in db-with-adjusted-gas [:wallet :send-transaction :original-gas] adjusted-gas))}))))
 
 (handlers/register-handler-fx
  :wallet-setup-navigate-back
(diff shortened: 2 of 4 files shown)


flexsurfer (approved):

goranjovic (approved):

lukaszfryc:
@tbenr could you check why gas limit reverts to 21000 while deploying the "simple contract" in the test dapp? The original gas is 130627. But it changes to 21000 after the reset button is clicked. I wonder why it's not the original value (130627). To reproduce: switch to Ropsten, open "Status Test Dapp", go to transactions tab and then to "deploy simple contract".

All works fine in other places like Airswap dapp or when requesting STT 👍

tbenr:
@lukaszfryc - this is the case when the gas is estimated by status-app and not by the dapp itself.
I proposed this change to set `original-gas` when the gas estimation is done on status-app side and `original-gas` is not already set (supposedly by dapp).

@goranjovic @flexsurfer I have a question: do we want to run a gas estimation every time "reset to Default" is pressed? or it is ok to "cache" it when the estimation is done the first time at transaction setup?

btw there is room to cleanup the estimation functions to align them to gas-price retrieval, that seems to use always web3 (always async).

lukaszfryc:
> this is the case when the gas is estimated by status-app and not by the dapp itself.
I proposed this change to set original-gas when the gas estimation is done on status-app side and original-gas is not already set (supposedly by dapp).

@tbenr ok, makes sense. 

> @goranjovic @flexsurfer I have a question: do we want to run a gas estimation every time "reset to Default" is pressed? or it is ok to "cache" it when the estimation is done the first time at transaction setup?

Let's wait for an answer before merging this PR.

flexsurfer:
i'm fine to merge as is

goranjovic:
> or it is ok to "cache" it when the estimation is done the first time at transaction setup?

what would be the use case behind it? e.g. user spamming the Estimate button?

tbenr:
@goranjovic 
> what would be the use case behind it? e.g. user spamming the Estimate button?

currently it is just a development convenience. To have a correct gas-estimation on "reset to default" requires additional work: it currently calculates an hardcoded value. Before improving that part we should decide how we want this to work.

goranjovic:
@tbenr It makes sense to cache the gas (limit) value. Not the gasPrice, though.

tbenr:
@goranjovic ok, I was thinking the same: gas estimation should be deterministic for a given transaction, gas price is externally determined.

thus, current PR should be ok :)

goranjovic:
@tbenr could you please rebase and resolve the conflicts so that I can merge it in?

tbenr:
@goranjovic done!

Issue #5793, opened 2018-09-10
[Epic] Implement "Save password" functionality for Android


lukaszfryc:
### Description

*Type*: Feature

*Summary*: Recently "Save password until logout" (https://github.com/status-im/status-react/pull/5617) was implemented for iOS. This issue is about doing the same for Android. More details can be found in https://github.com/status-im/status-react/pull/5617.

### TODO

- [x] Discuss the password storage options // @corpetty @rasom @mandrigin 
- [x] Implement minimum security guarantee in react-native-keychain
- [x] Enable storing password with `TEE` guarantee on Android
- [ ] Hash password before using it and implement the migration

mandrigin:
We use `react-native-keychain` library for saving passwords.

Next steps:
1. fork this library and switch to our fork;
2. use the same library to implement this on Android (essentially, enable Android code-paths);
3. disable or remove whatsoever their "unsafe" option of storing in the facebook DB

Some gotchas
1. passwords shouldn't be accessible to any other app on the OS
2. this feature should not be enabled if the device is rooted
3. there might be an OS version limitation, I would say let's enable it for Android 8.0+

When that is implemented, please ping me and @corpetty for additional security review.

chadyj:
@rasom Interested in taking this one? 

We have iOS and Android parity except for this feature, and about half our users are on Android and they would love this.

cheatfate:
I'm just curious, because i'm actually very interested in this feature.
You said you are using `react-native-keychain`, and i see this library is actually supports `Keychain.BIOMETRY_TYPE::FINGERPRINT`. Is it tough task to implement fingerprint usage on Android?

mandrigin:
@cheatfate for biometrics there are https://github.com/status-im/status-react/issues/5556 and https://github.com/status-im/status-react/issues/5557.

It is not so hard to just use some feature of a lib, but it is much trickier to figure out if this library is actually as safe as it claims. The last thing you want in crypto is to have an adversary to be able to decrypt your private key. We need to make sure that it is safe first, on different devices with different OS versions and flavours from different vendors.

mandrigin:
## [Research] Use an Android Backround Service to hold password in RAM

As a first step, I tried to use a background service to keep password in RAM only.
That wouldn't have solved the whole issue, but it would have a much less of the attack surface than storing it somewhere on the device. (Afaik, there is an idea to do a similar thing with Desktop and Keycard).

[...]

PR #6616, opened 2018-11-06
[ANDROID #5793] enable saving password on android

rasom:
partially solves #5793

As was suggested by @mandrigin this feature is enabled only for Android 23+ because of security concerns https://github.com/oblador/react-native-keychain#security

uses our fork of react-native-keychain with minimum security guarantees for some operations

NOT hashes password (will be a follow-up PR)

status: ready

### DIFF
--- a/src/status_im/native_module/impl/module.cljs
+++ b/src/status_im/native_module/impl/module.cljs
@@ -8,7 +8,8 @@
             [status-im.utils.platform :as p]
             [status-im.utils.async :as async-util]
             [status-im.react-native.js-dependencies :as rn-dependencies]
-            [clojure.string :as string]))
+            [clojure.string :as string]
+            [status-im.utils.platform :as platform]))
 
 ;; if StatusModule is not initialized better to store
 ;; calls and make them only when StatusModule is ready
@@ -171,3 +172,24 @@
 (defn update-mailservers [enodes on-result]
   (when status
     (call-module #(.updateMailservers status enodes on-result))))
+
+(defn rooted-device? [callback]
+  (cond
+    ;; we assume that iOS is safe by default
+    platform/ios?
+    (callback false)
+
+    ;; we assume that Desktop is unsafe by default
+    ;; (theoretically, Desktop is always "rooted", by design
+    platform/desktop?
+    (callback true)
+
+    ;; we check root on android
+    platform/android?
+    (if status
+      (call-module #(.isDeviceRooted status callback))
+      ;; if module isn't initialized we return true to avoid degrading security
+      (callback true))
+
+    ;; in unknown scenarios we also consider the device rooted to avoid degrading security
+    :else (callback true)))
--- a/src/status_im/utils/keychain/core.cljs
+++ b/src/status_im/utils/keychain/core.cljs
@@ -3,10 +3,12 @@
             [taoensso.timbre :as log]
             [status-im.react-native.js-dependencies :as rn]
             [status-im.utils.platform :as platform]
-            [status-im.utils.security :as security]))
+            [status-im.utils.security :as security]
+            [status-im.native-module.core :as status]))
 
 (def key-bytes 64)
 (def username "status-im.encryptionkey")
+(def android-keystore-min-version 23)
 
 (defn- bytes->js-array [b]
   (.from js/Array b))
@@ -14,13 +16,25 @@
 (defn- string->js-array [s]
   (.parse js/JSON (.-password s)))
 
+(defn- check-conditions [callback & checks]
+  (if (= (count checks) 0)
+    (callback true)
+    (let [current-check-fn (first checks)
+          process-check-result (fn [callback-success callback-fail]
+                                 (fn [current-check-passed?]
+                                   (if current-check-passed?
+                                     (callback-success)
+                                     (callback-fail))))]
+      (current-check-fn (process-check-result
+                         #(apply (partial check-conditions callback) (rest checks))
+                         #(callback false))))))
+
 ;; ********************************************************************************
 ;; Storing / Retrieving a user password to/from Keychain
 ;; ********************************************************************************
 ;;
 ;; We are using set/get/reset internet credentials there because they are bound
 ;; to an address (`server`) property.
-
 
 (defn enum-val [enum-name value-name]
   (get-in (js->clj rn/keychain) [enum-name value-name]))
@@ -43,15 +57,37 @@
   ;; > you might choose kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly.
   ;; That is exactly what we use there.
   ;; Note that the password won't be stored if the device isn't locked by a passcode.
-  {:accessible    (enum-val "ACCESSIBLE"     "WHEN_PASSCODE_SET_THIS_DEVICE_ONLY")})
+  {:accessible (enum-val "ACCESSIBLE" "WHEN_PASSCODE_SET_THIS_DEVICE_ONLY")})
+
+(def keychain-secure-hardware
+  ;; (Android) Requires storing the encryption key for the entry in secure hardware
+  ;; or StrongBox (see https://developer.android.com/training/articles/keystore#ExtractionPrevention)
+  "SECURE_HARDWARE")
+
+;; These helpers check if the device is okay to use for password storage
+;; They resolve callback with `true` if the check is passed, with `false` otherwise.
+;; Android only
+(defn- device-not-rooted? [callback]
+  (status/rooted-device? (fn [rooted?] (callback (not rooted?)))))
+
+;; Android only
+(defn- secure-hardware-available? [callback]
+  (-> (.getSecurityLevel rn/keychain)
+      (.then (fn [level] (callback (= level keychain-secure-hardware))))))
+
+;; iOS only
+(defn- device-encrypted? [callback]
+  (-> (.canImplyAuthentication
+       rn/keychain
+       (clj->js
+        {:authenticationType
+         (enum-val "ACCESS_CONTROL" "BIOMETRY_ANY_OR_DEVICE_PASSCODE")}))
+      (.then callback)))
 
 ;; Stores the password for the address to the Keychain
 (defn save-user-password [address password callback]
-  (if-not platform/ios?
-    (callback true) ;; no-op on Androids (for now)
-    (-> (.setInternetCredentials rn/keychain address address password
-                                 (clj->js keychain-restricted-availability))
-        (.then callback))))
+  (-> (.setInternetCredentials rn/keychain address address password keychain-secure-hardware (clj->js keychain-restricted-availability))
+      (.then callback)))
 
 (defn handle-callback [callback result]
   (if result
@@ -60,29 +96,30 @@
 
 ;; Gets the password for a specified address from the Keychain
 (defn get-user-password [address callback]
-  (if-not platform/ios?
-    (callback) ;; no-op on Androids (for now)
+  (if (or platform/ios? platform/android?)
     (-> (.getInternetCredentials rn/keychain address)
-        (.then (partial handle-callback callback)))))
+        (.then (partial handle-callback callback)))
+    (callback))) ;; no-op for Desktop
 
 ;; Clears the password for a specified address from the Keychain
 ;; (example of usage is logout or signing in w/o "save-password")
 (defn clear-user-password [address callback]
-  (if-not platform/ios?
-    (callback true)
+  (if (or platform/ios? platform/android?)
     (-> (.resetInternetCredentials rn/keychain address)
-        (.then callback))))
+        (.then callback))
+    (callback true))) ;; no-op for Desktop
 
 ;; Resolves to `false` if the device doesn't have neither a passcode nor a biometry auth.
 (defn can-save-user-password? [callback]
-  (if-not platform/ios?
-    (callback false)
-    (-> (.canImplyAuthentication
-         rn/keychain
-         (clj->js
-          {:authenticationType
-           (enum-val "ACCESS_CONTROL" "BIOMETRY_ANY_OR_DEVICE_PASSCODE")}))
-        (.then callback))))
+  (cond
+    platform/ios? (device-encrypted? callback)
+
+    platform/android?  (check-conditions
+                        callback
+                        secure-hardware-available?
+                        device-not-rooted?)
+
+    :else (callback false)))
 
 ;; ********************************************************************************
 ;; Storing / Retrieving the realm encryption key to/from the Keychain
--- a/android/app/src/main/java/im/status/ethereum/MainApplication.java
+++ b/android/app/src/main/java/im/status/ethereum/MainApplication.java
@@ -52,7 +52,7 @@ public class MainApplication extends MultiDexApplication implements ReactApplica
                 webViewDebugEnabled = true;
             }
 
-            StatusPackage statusPackage = new StatusPackage(BuildConfig.DEBUG, devCluster);
+            StatusPackage statusPackage = new StatusPackage(BuildConfig.DEBUG, devCluster, RootUtil.isDeviceRooted());
             Function<String, String> callRPC = statusPackage.getCallRPC();
             List<ReactPackage> packages = new ArrayList<ReactPackage>(Arrays.asList(
                     new MainReactPackage(),
--- a/env/dev/env/config.cljs
+++ b/env/dev/env/config.cljs
@@ -1,5 +1,6 @@
 (ns env.config)
 
-(def figwheel-urls {:android "ws://192.168.10.203:3449/figwheel-ws",
-                    :ios     "ws://localhost:3449/figwheel-ws",
-                    :desktop "ws://localhost:3449/figwheel-ws"})
+(def figwheel-urls {:android "ws://localhost:3449/figwheel-ws",
+ :ios "ws://192.168.0.9:3449/figwheel-ws",
+ :desktop "ws://localhost:3449/figwheel-ws"}
+)
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -68,7 +68,7 @@ PODS:
     - React
   - React/Core (0.56.0):
     - yoga (= 0.56.0.React)
-  - RNKeychain (3.0.0):
+  - RNKeychain (3.0.0-rc.3):
     - React
   - yoga (0.56.0.React)
 
@@ -123,4 +123,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 7636f960a0dbec2dd55b8b20e244befa3fdb4438
 
-COCOAPODS: 1.5.2
+COCOAPODS: 1.5.3
--- a/mobile_files/package.json.orig
+++ b/mobile_files/package.json.orig
@@ -47,7 +47,7 @@
     "react-native-image-crop-picker": "0.18.1",
     "react-native-image-resizer": "https://github.com/status-im/react-native-image-resizer.git#1.0.0-1",
     "react-native-invertible-scroll-view": "1.1.0",
-    "react-native-keychain": "3.0.0",
+    "react-native-keychain": "https://github.com/status-im/react-native-keychain#v.3.0.0-status",
     "react-native-level-fs": "3.0.1",
     "react-native-os": "https://github.com/status-im/react-native-os.git#1.1.0-1",
     "react-native-qrcode": "0.2.7",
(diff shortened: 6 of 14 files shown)


janherich (approved):

cammellos:
Not to be a party pooper, but reading the docs:

> Follow best practices and do not store user credentials on a device. Instead use tokens or other forms of authentication and re-ask for user credentials before performing sensitive operations.

Which I believe is what we were doing (for IOS ) and this does the same for android.

The docs recommend to use a temporary token instead of storing the password (which is tricky, but probably can be done).

Is this a concern, should we at least aim for removing the password from the keychain eventually? @corpetty

rasom:
> Which I believe is what we were doing (for IOS ) and this does the same for android.

yep correct, right now the password is stored in keychain on ios

This password is used for decryption of the private key. So if requirements are 
1) not store the password on the device and 
2) be able to decrypt the private key without entering the password 

we can derive/generate some token and encrypt PK with that token but it still will be stored in the keychain. And as we don't have here any session timeout or anything like this (if adversary decrypted PK it can't be undone :) ), it doesn't look more secure to me.
On the other hand, if possible revealing of the password itself is the concern, then we definitely need to use the token. 

> The docs recommend to use a temporary token instead of storing the password (which is tricky, but probably can be done).

well, they do not mention **temporary** tokens, just tokens :). It would make sense, but again in our case if you gained access once any expiration will not help. The solution might be pretty straightforward: create another PK file encrypted with token, store token into keychain, then login using token. Both token and PK file might be regenerated/reencrypted after each loginning (kind of temporary, not sure).
@cammellos let me know if it makes any sense to you  

@corpetty's input on this will be helpful

cammellos:
> create another PK file encrypted with token, store token into keychain, then login using token. Both token and PK file might be regenerated/reencrypted after each loginning

sounds reasonable to me, basically:

- an attacker gets temporary access to the keychain
1) it won't disclose the user password 
2) It will be able to decrypt the pk until re-login (re-keying)

- an attacker gets permanent access to the keychain
1) it won't disclose the user password 
2) It will be able to decrypt the pk

It would be good to have it time-based but not sure how to do it, I would have to research a bit, but login sounds definitely like a step in the right direction.

Also we need to check where status-go stores the encryption key, as it might be in the folder backed up automatically (Documents/ on ios , android anything not under no_backup, although automatic backup are disable on android.

There is also the SQL database with encryption key that uses the user password for encryption, and that need to be re-keyed (should be easy), there's also an argument for storing the PK there, instead of a file.

rasom:
It's a bit more complicated with the database, the solution I mentioned implies having two separate PK files, one encrypted with a password another with a token. If we renew token after logging in we will need to re-encrypt DB before using it, and this operation may be indeterminately long depending on db size. Logging in may become unacceptably slow

**upd**: ok it doesn't matter how many PK files we have, the problem is rather that we can decrypt that file asynchronously after logging in, and this can't be done with DB. It should be re-encrypted during logging in.

cammellos:
not sure, performance is a concern true, but to rekey you can run the command in the background ( during re keying you won't be able to access the db ) , but you can do after login so the user might not notice.

for the 2 files, not sure i understand correctly, is temporary or permanent? but I would avoid if possible as it increases the surface of attack.

if the concern is losing access to the token, we can make it derivable from the password ( one way), so we can cather for that scenario.

rasom:
> for the 2 files, not sure i understand correctly, is temporary or permanent? but I would avoid if possible as it increases the surface of attack.

well yeah it doesn't metter here, my bad

> not sure, performance is a concern true, but to rekey you can run the command in the background ( during re keying you won't be able to access the db ) , but you can do after login so the user might not notice.

again db will not be writable for an undetermined period of time, which makes the app not fully functional and might complicate logic and UX a lot. But better just to test how fast it is for, let's say, 100MB file on the average android device and then make conclusions.

cammellos:
i agree, also i was talking about a different db, the one we use for pfs in
status go ( which it s ok if it s locked for a short as you just will wait
for it to be unlocked and then you d be able to decrypt messages).

i did not consider realm, i forgot was also encrypted with the user
password, that s more problematic true

rasom:
Options i see atm are:
1) disable this functionality on iOS until we implement a more secure solution 
2) merge it as it is for android and introduce a better solution later (assuming we are in beta and we already have this on iOS)

@corpetty @cammellos

cammellos:
i have to defer judgement on this one, my gut feeling is that extending the
feature to android might not be a step in the right direction

@mandrigin what do you think?

rasom:
also, we can explicitly disable this functionality for rooted Android devices

corpetty:
I don't feel we should ever store the password on the device.  If we want to implement a feature like staying signed in, then we could do that via a "logged_in=(True/False)" state which changes over a period of inactivity/logout/screen-lock/etc and is an option to the user.

A type of time-based session token also works in this scenario, which could be wiped at logout/screen-lock/etc 

This also mitigates the rooted/non-rooted device issue. 

Is there something that blocks that based on how we (en/de)crypt data?

yenda:
The private key is much more important than the password and has to be on the device

corpetty:
yes, but how that private key is (en/de)crypted shouldn't _also_ be on the same device, unless stored in a separate key-manager app like lastpass (outside the realm of status development)

yenda:
on Android it can be solved by having status run as a background service.
that way you only have to login again when the service is killed

corpetty:
If a user decides to use a convenience feature like "staying signed in" then they are making a personal decision to trade-off security for usability.  Our goal is to make that trade-off as asymmetric as possible on the side of favoring usability.  

By storing the password on the device (even temporarily), we increase the ways an attacker can get to the private key even outside the time they're taking advantage of the usability feature.

rasom:
@corpetty 
In order to log in, we need to decrypt the private key (which is stored in a file on the device), this always requires the password. We also need to open realm database and it is encrypted using key which is derived from the password.

It's hard to imagine how we can use time-based session token for realm DB, as long as we need to know the last used token when we want to open the DB again (and it might be in a year after the previous launch of the app on the device, completely arbitrary).

> If we want to implement a feature like staying signed in, then we could do that via a "logged_in=(True/False)" state which changes over a period of inactivity/logout/screen-lock/etc and is an option to the user.

well, that's what we have here

rasom:
@yenda 
> on Android it can be solved by having status run as a background service.
that way you only have to login again when the service is killed

i don't understand what exactly can be solved with background service

corpetty:
The real problem with this is the fact that the whisper key and funded account key are the same.  The longer a user stays logged in the longer the account key is exposed. 

These two things should not be linked, as optimizing for a messenger app will have direct consequences on the security profile of the wallet.

rasom:
> The real problem with this is the fact that the whisper key and funded account key are the same. The longer a user stays logged in the longer the account key is exposed.

How exactly is it exposed while the user is logged in? Do you mean that we have a plain key in memory or what exactly? 

> These two things should not be linked, as optimizing for a messenger app will have direct consequences on the security profile of the wallet.

i agree, though if we want to have two separate passwords here (one for whisper, another for funded account) that might be a bit more complicated from UX pov.
We wouldn't need the key for the funded account for logging in, but most of the issues which we have now will remain the same with that separate whisper key if we don't want to store its password on the device.

cammellos:
I think @yenda might be on the right track.

We could store the user password in memory (an atom) and then at that point we have the option to disable access to it (remove from memory) after a set amount of time has expired, on screen lock etc.

Signal app uses a similar strategy (it consider the keychain not secure enough), although uses a background service, not sure exactly why, but most likely is to expire keys after a certain amount of time has elapsed.

On further investigation seems like memory can be dumped on file if the process is killed, so something we need to take into consideration.

Furthermore we should use the hash of the password (not sure if we do that already), so in case we store it (whether in memory or storage), we don't compromise the actual password.

cammellos:
https://github.com/signalapp/Signal-Android/blob/8ec3cf6a435917a404180b0bae2afd9eac44ea03/src/org/thoughtcrime/securesms/service/KeyCachingService.java for reference

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/signalapp/Signal-Android">signalapp/Signal-Android</a></strong></div><div>A private messenger for Android. Contribute to signalapp/Signal-Android development by creating an account on GitHub.</div></blockquote>

rasom:
Oh ok, thanks @cammellos . Then it's not about running the app as a background service, but just a service for keeping the password in memory. 

> Furthermore we should use the hash of the password (not sure if we do that already), so in case we store it (whether in memory or storage), we don't compromise the actual password.

we definitely store a plain password in the keychain atm. We actually could hash it before passing to status-go side (at account creation and recovery).

rasom:
some points 
1) right now in release/nightly/develop versions of the app we store a plain password in the keychain (iOS)
2) originally we decided store password on the device because app was killed too often and user had to enter their password quite frequently if used app actively. So we need this functionality not specifically for staying logged in, but to able to re-login without asking the password. 
3) we don't want to store the password (or its hash) on the device because 
   > how that private key is (en/de)crypted shouldn't also be on the same device, unless stored in a separate key-manager app like lastpass (outside the realm of status development)
4) we might want to derive some token from the password (or just generate it?) on each login and use it for PK and databases (en/de)cryption. This might solve possible password revealing but doesn't solve 2) as the token will be stored on the device. It should be persistent in case if we are going to use this token for realm db, because we will need it in order to open the DB on the next login. Sounds a bit useless though. 
5) on Android we can temporarily store password (rather its hash) in memory using a service in the background (with no guaranty of persistence, service might be killed at any point in time), which would allow as to reuse password if the app was closed but service hasn't been killed. Storing time-based or some randomly generated in the service token will not work at least for realm db, because we need to guarantee that token will be available on the next run.
6) we might want to have a separate PK for whisper and for the funded account. In this case, we can have separate passwords for them and even if whisper key will have all the same problems which a single key has, the password for the funded account will not be required for logging in, thus no need to store it in any form.

rasom:
I would suggest
1) closing this PR
2) optionally disable this functionality on iOS
3) adding changes so that password will be hashed right after account creation/recovery (this will at least solve storing a plain password on iOS) and all migrations necessary for this
4) move further discussion related to storing password to https://discuss.status.im/ 

Also, might be a silly suggestion, as the root reason for adding this functionality is an inconvenience which is caused by the fact that app is frequently killed by OS, we might consider looking for solutions related to performance. We definitely can't guaranty that OS will not kill app (and it definitely will be killed), but probably we can reduce a frequency of this event so that it will not be noticeably inconvenient from UX pov. It might be fine for the user to log in once per day, for instance. 

@corpetty @cammellos @mandrigin @yenda 

<blockquote><div> Status.im</div><div><strong><a href="https://discuss.status.im/">Status.im</a></strong></div><div>A place for Status.im community.</div></blockquote>

cammellos:
sounds reasonable to me, thanks @rasom

corpetty:
I'm for killing it and moving convo to discuss until we come to a better conclusion.

mandrigin:
Okay, let's get back to this issue.

@cammellos I don't see anywhere, where signal's service help them to keep password alive more than the app, as I pointed here: https://github.com/status-im/status-react/issues/5793#issuecomment-446270791 the service is killed as soon as the app is killed.
As I can see, their service is used essentially for expiring the password/locking your account earlier than when the app is killed. Also, if we want any kind of biometric login, we have to store the password somewhere on the device. The most paranoid we can get is (needs PoC though):
- to use a key, randomly generated on some secure chip (that Pixels have, for instance)
- protect this key with biometric data and then use it to encrypt/decrypt password.

Then, when user signs in, it will:
- get the encrypted password from somewhere;
- request access to the decryption key via biometry;
- decrypt the key;
- use it to decrypt the private key.

Franky, taking into account that it is relatively trivial to write a keylogger for Android (see [this](https://spyadvice.com/android-keylogger/)), I'm not sure if entering your password every time is more secure than storing the password in the Key Storage. I will research a bit more details, but let's keep this PR open, because the Clojure APIs will probably be the same.

(On a side note, it is not safe to export this info to React Native at all, in the future we'll need to get rid of `react-native-keychain`).

mandrigin:
@cammellos @corpetty to me, using Android Keystore System looks good enough, especially if we can enable it only when it is hardware supported (TEE/SE). See more info here ("Extraction prevention" is what you want): https://developer.android.com/training/articles/keystore

<blockquote><div>Android Developers</div><div><strong><a href="https://developer.android.com/training/articles/keystore">Android keystore system  |  Android Developers</a></strong></div></blockquote>

mandrigin:
@corpetty @yenda the implementation with security guarantees (from https://github.com/status-im/react-native-keychain/pull/6). It **doesn't** hash the password right now (I think it should be a separate PR).

What's left?
- [x] review/merge security guarantees API to our fork or react-native-keychain.
- [x] test this PR
- [x] add a bit more comments on these security guarantees in Clojure.

src/status_im/native_module/impl/module.cljs
```diff
@@ -167,3 +168,8 @@
 (defn update-mailservers [enodes on-result]
   (when status
     (call-module #(.updateMailservers status enodes on-result))))
+
+(defn rooted-device? [callback]
+  (if (and platform/ios? status)
+    (call-module #(.isDeviceRooted callback))
```
dmitryn: should it be `(call-module #(.isDeviceRooted status callback))` ?

mandrigin: great point, also I see this logic is off a bit

mandrigin: fixed it

src/status_im/utils/keychain/core.cljs
```diff
@@ -43,15 +45,17 @@
   ;; > you might choose kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly.
   ;; That is exactly what we use there.
   ;; Note that the password won't be stored if the device isn't locked by a passcode.
-  {:accessible    (enum-val "ACCESSIBLE"     "WHEN_PASSCODE_SET_THIS_DEVICE_ONLY")})
+  {:accessible (enum-val "ACCESSIBLE" "WHEN_PASSCODE_SET_THIS_DEVICE_ONLY")})
+
+(def keychain-secure-hardware
+  ;; (Android) Requires to store the encryption key for the entry in secure hardware
```
pedropombeiro:
```suggestion
  ;; (Android) Requires storing the encryption key for the entry in secure hardware
```

src/status_im/utils/keychain/core.cljs
```diff
@@ -60,29 +64,36 @@
...
+        (.then callback))
+
+    platform/android?
+    (-> (.getSecurityLevel rn/keychain)
+        (.then (fn [level]
+                 ;; we only enable password saving 
+                 ;; if secure hardware is available on Android
+                 (if (not= level keychain-secure-hardware)
+                   (callback false)
+                   (status/rooted-device? (comp callback not))))))
```
pedropombeiro: Wouldn't this be more readable as:
```clojure
;; we only enable password saving 
;; if secure hardware is available on Android and the device is not rooted
(callback (and (= level keychain-secure-hardware)
               (not status/rooted-device?)))
```

mandrigin: @PombeirP I think I made it more readable in a slightly different style ;)

pedropombeiro (approved):

corpetty (approved):

Serhy:
@mandrigin builds #13 from above Jenkins job tested on Android 6 and iOS 12.0.1:
* Account created and ‘Save password until logout’ for it. App closed - opened -> autologin triggered fine.
* Recovered account and ‘Save password until logout’ for it. App closed - opened -> autologin triggered fine.
* For autologin passwords `-/:;()$&@“.,?!‘[]{}#%^*+=_\|~<>€£¥•.,?!’` , `aaaaa123`, `AaàáâãäåæāăąA` were tested.


1) In the beginning, I was trying to use it on Android (6, 7, 8) emulators. Had no luck in order to enable this feature, - I realised that emulators were rooted. This is a likely scenario for users who with ‘rooted’ access + Device screen passcode set will be puzzled with locked toggle button and message ’Set device passcode to save password’. 
Could we detect that if *device is rooted*, - the ‘Disable root access to save password’ followed by locked and disabled toggle button displayed?
So if the user has rooted Android and no passcode enabled we show (1) ‘Set device passcode to save password’
If the user has rooted Android and passcode enabled we show (2) ‘Unroot your device to save password’
If the user has unrooted Android and passcode enabled we allow to save password with the (3) ‘Save password until logout’


2) I couldn’t enable autologin feature on Samsung J7 (Android 7) real device. It’s not clear yet what am I missing because I tried to set passcode for device screen as PIN, password (six chars), draw pattern, fingerprint, - and in all cases, after I cleared data for Status app and started from scratch (account creation -> log out -> try to ‘Save password until logout’ ) it persist with `Set device passcode to save password` next to the password field. The device is not rooted.
What do I need to check more in order to figure out the reason of feature is not working? 

Additionally while testing builds here - the https://github.com/status-im/status-react/issues/7192 issue discovered with both Android and iOS (No way to login for account after an attempt to save the wrong password with autologin)

mandrigin:
@Serhy for (2) and, maybe (1). If a device doesn't have any hardware key storage support, then this feature won't be available. Obviously, simulators won't have it.

Serhy:
@mandrigin Latest build for Android 6.0.1, Android 7, Android 8, Android 9 tested (iOS checked too). Everything looks good, but not with MacOS: after upgrade from 6th January nightly build (and confirmed for desktop Release candidate) on to https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190107-144730-e0235b-pr.dmg (latest build) i'm getting the `i.getInternetCredentialsForServer is not a function` error and no way to proceed to use app after hit OK button. 

screenshot 2019-01-08 at 14 46 22

Status.log tail for error
```
14:47:46.562 - D - [jsserver] - [status-im.utils.handlers:34] - Handling re-frame event:  :set-initial-props
14:47:46.571 - D - [jsserver] - [status-im.utils.handlers:34] - Handling re-frame event:  :init.callback/get-encryption-key-success
14:47:46.621 - I - [jsserver] - [status-im.data-store.realm.core:124] - realm: moving all realms
14:47:46.625 - W - [jsserver] - [status-im.data-store.core:25] - Could not move realms TypeError: re-matches must match against a string.
14:47:46.625 - D - [jsserver] - [status-im.data-store.realm.core:199] - Opening base realm... (first run)
14:47:46.626 - I - [jsserver] - [status-im.data-store.realm.core:177] - migrate-realm
14:47:46.642 - I - [jsserver] - [status-im.data-store.realm.core:148] - migrate schemas 18
14:47:46.643 - D - [jsserver] - [status-im.data-store.realm.core:47] - Opening realm at  /Users/sergiichumak/Library/Application Support/Status/default.realm/default.realm ...
14:47:46.644 - D - [jsserver] - [status-im.data-store.realm.core:49] - Using encryption key...
14:47:46.646 - D - [jsserver] - [status-im.data-store.realm.core:203] - Created @base-realm
14:47:46.647 - D - [jsserver] - [status-im.utils.handlers:34] - Handling re-frame event:  :init.callback/init-store-success
14:47:46.749 - D - [jsserver] - PRETTY PRINTED EXCEPTION

NAME: "TypeError"
MESSAGE: i.getInternetCredentialsForServer is not a function

TypeError: i.getInternetCredentialsForServer is not a function
    at Object.n.getInternetCredentials (evalmachine.<anonymous>:312:480)
    at Object.status_im.utils.keychain.core.get_user_password (evalmachine.<anonymous>:12:6789339)
    at G__32005_32011 (evalmachine.<anonymous>:12:6797782)
    at evalmachine.<anonymous>:12:1604937
    at Object.re_frame.interceptor.invoke_interceptor_fn (evalmachine.<anonymous>:12:1560178)
    at Object.re_frame.interceptor.invoke_interceptors (evalmachine.<anonymous>:12:1560462)
    at Object.re_frame.interceptor.execute (evalmachine.<anonymous>:12:1562834)
    at Object.re_frame.events.handle (evalmachine.<anonymous>:12:1571338)
    at re_frame.router.EventQueue.re_frame$router$IEventQueue$_process_1st_event_in_queue$arity$1 (evalmachine.<anonymous>:12:1598424)
    at re_frame.router.EventQueue.re_frame$router$IEventQueue$_run_queue$arity$1 (evalmachine.<anonymous>:12:1583352)
```

That was not the case with previous build though (i.e. it's all fine with 13th Jenkins build for MacOS https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-181219-163018-3f0d86-pr.dmg). Also it's all fine upgrading to latest nightly

<blockquote></blockquote>
<blockquote></blockquote>

mandrigin:
fixed in the latest commit :)

Serhy:
Everything looks good now, @mandrigin let's merge!

PR #6617, opened 2018-11-06
[#6294] add desktop alpha release warning

### LINKED ISSUE #6294: Desktop alpha release warning

### User Story

As a user, I want to be informed of the risks of using beta software so that I can take necessary safeguards.

### Description

Desktop doesn't have wallet functionality, but the recovery phrase used can belong to a wallet with assets that need to be protected. As such as need to let users know they are using beta software and should use it for testing.

#### Expected behavior

Warning on signup and recovery

### Solution

Show a notification on signup and recovery.
cc @EugeOrtiz @obi2020 

### Additional Information
Platform: Desktop

cc @corpetty

dmitryn:
fixes #6294

### Summary:

* Show alpha release warning on desktop after login (show it only once)
* Update react-native-desktop to most recent version where Alert dialog layout is fixed.

status: ready

screenshot 2018-11-06 16 19 08

### DIFF
--- a/desktop_files/package.json
+++ b/desktop_files/package.json
@@ -55,7 +55,7 @@
     "querystring-es3": "0.2.1",
     "react": "16.4.1",
     "react-dom": "16.4.2",
-    "react-native": "git+https://github.com/status-im/react-native-desktop.git#master",
+    "react-native": "git+https://github.com/status-im/react-native-desktop.git#v0.56.1",
     "react-native-background-timer": "2.0.0",
     "react-native-camera": "0.10.0",
     "react-native-config": "git+https://github.com/status-im/react-native-config.git",
--- a/desktop_files/package-lock.json
+++ b/desktop_files/package-lock.json
@@ -11065,7 +11065,7 @@
       }
     },
     "react-native": {
-      "version": "git+https://github.com/status-im/react-native-desktop.git#master",
+      "version": "git+https://github.com/status-im/react-native-desktop.git#567ca5e9b53a91d55a2bc4e915bf4f5d2b6304c4",
       "requires": {
         "absolute-path": "0.0.0",
         "art": "0.10.3",
--- a/src/status_im/accounts/core.cljs
+++ b/src/status_im/accounts/core.cljs
@@ -21,6 +21,17 @@
        (i18n/label :mainnet-is-default-alert-text)
        #(re-frame/dispatch [:accounts.ui/mainnet-warning-shown])))))
 
+(fx/defn show-desktop-alpha-release-warning [{:keys [db]}]
+  (let [enter-name-screen? (= :enter-name (get-in db [:accounts/create :step]))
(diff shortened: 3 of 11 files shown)


desktop_files/package.json
```diff
@@ -55,7 +55,7 @@
     "querystring-es3": "0.2.1",
     "react": "16.4.1",
     "react-dom": "16.4.2",
-    "react-native": "git+https://github.com/status-im/react-native-desktop.git#master",
+    "react-native": "git+https://github.com/status-im/react-native-desktop.git#420798a31ea89f8d2de007f7321b6db64a444685",
```
vkjr: Hi!
What branch it points to?
It should be still `master` I believe.

vitvly (approved):

desktop_files/package.json
dmitryn: @vkjr hi! it points to latest commit in master https://github.com/status-im/react-native-desktop/commit/420798a31ea89f8d2de007f7321b6db64a444685
I changed it to be refererring exact commit because reference to master always change, but reference to exact commit is more explicit way to say what version you're using.

vkjr: I think we should discuss this change more thoroughly in a status-react channel.
At least i'd be glad to see some descriptive tag here, instead of commit number, to have better orientation.

dmitryn: @vkjr makes sense to use descriptive tags. For example, we could use git tags to specify version like `v0.56.1`.

@MaxRis @siphiuel what do you guys think about using git tags for this?

rasom (approved):

desktop_files/package.json
vitvly: Good idea.

dmitryn: Changed to git tag `v0.56.1`  https://github.com/status-im/react-native-desktop/tree/v0.56.1

vkjr (approved):

desktop_files/package.json
vkjr: Thanks for changing!
Just to note, `0.56.1` is a version of original `react-native`we are basing on. But within this version we have a lot of commits in `react-native-desktop`, so next tags will be more granular, like `v0.56.1-1`, `v0.56.1-2`, etc..

SlizkovaGanna:
All the test failures were checked, no actual issues found.

Tested on MacOs

**Test cases for new installed build and for replacement:**
- message is correct
- message appears for a new created user
- message does not appear for re-login
- message appears for existing user
- message appears for a new user created when trying to recover an account
- message appears for a recovered user

@dmitryn is that correct that the message should appear not only for the new installation, but for an update too?
I guess my assumption is that we've already shown this warning to a user once (upon the first installation), are we going to show it every time he updates?

dmitryn:
@SlizkovaGanna yes, it should appear once for both new or updated account.

PR #6622, opened 2018-11-06
Automated token settings internal validation 6509

### LINKED ISSUE #6340: Incorrect names for MKR, DGD, ETHOS and DATA tokens
### Description

*Type*: Bug

*Summary*: Tokens MKR, DGD, ETHOS, DATA have inconsistencies in names or/and symbols after comparing data from [tokens.cljs](https://github.com/status-im/status-react/blob/develop/src/status_im/utils/ethereum/tokens.cljs) with data from http://api.ethplorer.io/. 

#### Incorrect names and symbols:
- address 0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2: 
token name 'MKR', expected 'Maker'
- address 0xe0b7927c4af23765cb51314a0e0521a9645f0e2a: 
token name 'Digix DAO', expected 'Digix'
- address 0x5af2be193a6abca9c8817001f45744777db30756: 
symbol 'BQX', expected 'ETHOS', 
token name 'Bitquence', expected 'Ethos'
- address 0x0cf0ee63788a0849fe5297f3407f701e122cc023:
 token name 'DATAcoin', expected 'Streamr DATAcoin'

### Reproduction

- Get token info by address in https://api.ethplorer.io
e.g. `https://api.ethplorer.io/getTokenInfo/0x0cf0ee63788a0849fe5297f3407f701e122cc023?apiKey=freekey`

### Additional Information
* Status version: develop

<blockquote></blockquote>
<blockquote></blockquote>

### LINKED ISSUE #6509: Get token config from the contracts instead of using hardcoded values
### User Story

As a developer I want an easier way to add correct and verified token config to the supported tokens list.

### Description

*Type*: Refactor

*Summary*: Right now when we add new ERC-20 tokens we have to supply the following info: contract address, symbol, name, number of decimals and icon in png format with symbol in the file name. 

Supplying correct information there is extremely important - setting a wrong token name can completely confuse the users and setting a wrong number of decimals can cause money loss. Verifying this data is a big part of manual effort on our end when we add new tokens.

Instead of doing that manually, we should only add the contract address, icon resource files and the symbol and extract full name and decimals from the contract. We could even get the symbol like that, but it is useful to keep it in the code as a reference as contract addresses on their own aren't very legible. We can still query the contract for symbol and then assert it is the same as in code, just in case.

The one complication with this is that tokens list is currently hardcoded, so it's available even at compile time. After this change, the same list would need to be populated in run time, so some changes throughout the code may be needed.

This issue is a pre-requisite for #6225

goranjovic:
fixes #6509

### Summary:

Improves how built in list of tokens is handled. Specifically:

- The list of built in tokens is now accessed as an app-db backed subscription, not a hardcoded constant. This will us to handle custom tokens (added by users, stored in realm) using the same subscription interface (tokens added by user will be enabled in a separate PR)

- Configuration for built-in tokens (name, symbol, decimals) is being checked against the respective token's smart contract. If there is a discrepancy, an error message is shown. If `name` or  `symbol` fields are not present in the contract or if they are empty, this check is ignored. If `decimals` field is missing, its check is only ignored if there is a flag `skip-decimals-check?` in the built in token config - this is because we are getting back 0 when the field is missing, so can't tell it apart from actual 0.
(NOTE: This check happens only if `ERC20_CONTRACT_WARNINGS` flag is enabled which is in all environments except release builds. Should be useful in development, testing and e2e tests).

- Any tokens that had the discrepancies were fixed, i.e. fixes #6340. Specifically:
       - BQX and DATA were renamed to ETHOS and Streamr DATAcoin, resp.
       - MKR and DGD were left as they were since they don't implement `name` and `symbol` constants and hence don't really have a canonical name version.

- Veros (VRS) contract was upgraded. Reflected the change in the token config #6646
- Migration was added to allow BQX -> ETHOS symbol rename in the case when user had this token enabled during the upgrade.

- Read only ERC20 operations were switched from direct low level JSON-RPC to ABI based web3 method calls.

Note: The original intention of this task was to automate population of all token config based on what the contract says. This is not universally possible since not all tokens implement the required fields. It is also not very practical since getting all the info from the contracts would involve a huge number of asynchronous tasks, which does not make sense to put on the init critical path, especially not for a limited number of hardcoded tokens. So, this PR presents a middle way which relies on the hardcoded info (for the build in tokens at least) but dynamically validates those values against the contracts.

### Testing notes (optional):

This PR currently has one Ropsten token intentionally broken to allow the validation to be tested. The steps are:

1. Log in and switch to Ropsten
2. There should be an error saying that for WGN token decimals settings doesn't match what's in the smart contract. This is ok, because I intentionally broke the WGN settings to trigger this error for testing. If the error doesn't appear, there is a problem.
3. Please let me know when this is tested so that I can "unbreak" WGN again.

--------------------

The other thing to test is migration from a previous version. Specifically the edge case is when user has BQX enabled in Wallet and then upgrades the app to this new version where the same token is called ETHOS. There are two scenarios:

Scenario 2:

1. User is on mainnet and doesn't enable BQX
2. Upgrades app to this PR
3. Wallet still doesn't show BQX or ETHOS under enabled tokens.
4. Manage assets screen shows Ethos / ETHOS and does not show BQX

Scenario 3:

1. User is on mainnet and does enable BQX
2. Upgrades app to this PR
3. Instead of BQX Wallet screen shows ETHOS under listed assets
4. Manage assets screen shows Ethos / ETHOS and does not show BQX

--------------------

Veros (VRS) token contract was changed to reflect the token upgrade. It should be unaffected by an app upgrade, whether or not it was enabled in the wallet.

**Functional**
- 1-1 chats - specifically token/eth transfers in chat
- wallet / transactions
- dapps / app browsing - specifically token minting in Andrey's dapp or token exchange transactions
- account recovery - specifically if tokens show up ok in recovered accounts
- networks - specifically if all networks are without token validation errors (except that one intentional error with WGN on Ropsten)

status: ready


goranjovic:
@dmitryn @jeluard @rasom @flexsurfer Guys, please review. I know it's a big one, but that's exactly why I don't want it to get stale and catch conflicts.

jeluard:
@goranjovic Can you detail when the dynamic validation happens and how we should react to discrepancies ?

goranjovic:
@jeluard 

The whole flow goes like this:

1. We have a hardcoded list of default tokens 
2. Right after login this list is populated into a field in app-db (from where it is accessible through a sub) 

{{only when ERC20_CONTRACT_WARNINGS}}
3. Then we asynchronously query `name`, `symbol` and `decimals` against contract of each token in the current network.
4. The actual checks if the info we have on token matches what the smart contract says happen in the callback function of those async calls.
5. The checks will pass if fields are equal OR field is not set in the smart contract. We validate tokens that do not define `name` and `symbol` as correct, since those fields are optional in ERC-20. Even though `decimals` field is also optional, we can't rely on its absence in the contract as web3 doesn't distinguish unset integer from a zero and zero is a valid number of decimals. So, in order to let tokens without `decimals` to validate as correct, we need to add them a `skip-decimals-check?` flag.
6. Finally, if any token on the current network fails to validate, we show a popup with an error (see screenshot below). The idea, at least in this PR, is to automate the validation of token config by cross checking two sources of truth - our hardcoded list and the contracts. Currently the same thing is done by @churik and @yevh-berdnyk in a semi-manual way by cross checking against a trusted exchange web site. 
{{/when}}

We can handle discrepancies in a number of ways and I'm open for any suggestions.

goranjovic:
screenshot-2018-11-09_13 03 18 57

jeluard:
I don't quite see the value of doing that at each status startup. Could be something we run periodically to make sure our list is accurate and uptodate.

goranjovic:
@jeluard What are the reasons not to do that at startup?

jeluard:
Complexity? Extra performance cost?
And then users get eventually a popup with some error but can't act upon it.

goranjovic:
@jeluard So you're thinking of the scenario where a token has some name which passes the validation, but then gets changed by the contract owner? Which is both outside our control and outside user's control.

But in that case, isn't what we're doing now just as bad. We would go on and keep calling that token under the old name.

jeluard:
What scenario do you want to address then ? :)

goranjovic:
Well there is that scenario, and the one where we can immediately see a discrepancy when we add/test a new token. But for that second scenario it really does not make sense to do this on startup.

Unless of course we offer the users something actionable when the discrepancy gets detected, e.g. to save the updated settings.

goranjovic:
Ok, what I'm going to do here is put the token validation behind a flag, so that we can run it locally when we want this (sort of like debug mode). It may make sense to keep it enabled in e2e tests. How does this sound @yevh-berdnyk @antdanchenko ?

goranjovic:
@jeluard I've made the checks optional - only when a flag is enabled. Should let us use it without disturbing anything for the end user.

yevh-berdnyk:
@goranjovic e2e apk is built with Ropsten as a default network and we have only 6 test tokens there. Will the validation be performed for all other tokens?

jeluard (approved):

goranjovic:
@yevh-berdnyk No, validation is only for the tokens on the current network. For it to trigger it is sufficient to switch the network to say mainnet and wait a few seconds. Is there any test suite that checks network switching?

yevh-berdnyk:
@goranjovic yes, we have couple of tests with switching to Mainnet

goranjovic:
@yevh-berdnyk Perfect, if there is a mismatch for any token, they'd be able to catch it.

goranjovic:
@dmitryn @rasom @flexsurfer Please review :panda_face:

dmitryn (approved):

src/status_im/data_store/realm/schemas/base/core.cljs
```diff
@@ -64,6 +64,8 @@
           extension/v12
           account/v15])
 
+(def v17 v15)
```
flexsurfer: could you elaborate?

src/status_im/ui/screens/wallet/events.cljs
```diff
@@ -96,6 +98,50 @@
...
+(defn- validate-token-name! [web3 {:keys [address symbol name]}]
+  (erc20/name web3 address #(when (and (seq %2) ;;NOTE(goranjovic): skipping check if field not set in contract
+                                       (not= name %2))
+                              (let [message (i18n/label :t/token-auto-validate-name-error
+                                                        {:symbol   symbol
+                                                         :expected name
+                                                         :actual   %2
+                                                         :address  address})]
+                                (log/warn message)
+                                (js/alert message)))))
```
flexsurfer: can we dispatch here `:alert` event?

src/status_im/ui/screens/wallet/events.cljs
```diff
@@ -96,6 +98,50 @@
...
+(defn- validate-token-symbol! [web3 {:keys [address symbol]}]
+  (erc20/symbol web3 address #(when (and (seq %2) ;;NOTE(goranjovic): skipping check if field not set in contract
+                                         (not= (clojure.core/name symbol) %2))
+                                (let [message (i18n/label :t/token-auto-validate-symbol-error
+                                                          {:symbol   symbol
+                                                           :expected (clojure.core/name symbol)
+                                                           :actual   %2
+                                                           :address  address})]
+                                  (log/warn message)
+                                  (js/alert message)))))
```
flexsurfer: and here

src/status_im/ui/screens/wallet/events.cljs
```diff
@@ -96,6 +98,50 @@
...
+  (when-not skip-decimals-check?
+    (erc20/decimals web3 address #(when (and (not nft?)
+                                             (not= decimals (int %2)))
+                                    (let [message (i18n/label :t/token-auto-validate-decimals-error
+                                                              {:symbol   symbol
+                                                               :expected decimals
+                                                               :actual   %2
+                                                               :address  address})]
+                                      (log/warn message)
+                                      (js/alert message))))))
```
flexsurfer: and here

src/status_im/utils/ethereum/erc20.cljs
```diff
@@ -22,27 +22,99 @@
...
+
+(defn get-instance* [web3 contract]
+  (.at (.contract (.-eth web3) abi) contract))
+
+(def get-instance
+  (memoize get-instance*))
+
 (defn name [web3 contract cb]
-  (ethereum/call web3 (ethereum/call-params contract "name()") cb))
+  (.name (get-instance web3 contract) cb))
```
flexsurfer, line 101: could you elaborate on this decision?

src/status_im/ui/screens/wallet/events.cljs
```diff
@@ -96,6 +98,50 @@
  (fn [{:keys [web3 obj success-event]}]
    (ethereum/estimate-gas-web3 web3 (clj->js obj) #(re-frame/dispatch [success-event %2]))))
 
+(defn- validate-token-name! [web3 {:keys [address symbol name]}]
```
flexsurfer, line 85: do we have tests for these functions?

flexsurfer (dismissed):

src/status_im/utils/ethereum/erc20.cljs
goranjovic: Yes. The previous solution returned a hex string that had to be parsed manually. That would have involved too much boilerplate and it looked unstable - the same string would be decoded fine or broken, not sure why. In the new one `web3` does it automatically.
If that's the decision you asked about.

src/status_im/data_store/realm/schemas/base/core.cljs
goranjovic: Kept the schema same. Migration only affected the data in the old realm, not its structure.

src/status_im/ui/screens/wallet/events.cljs
goranjovic: Why? This is an effect function, it's perfectly fine to have a side effect in it. Why pollute the environment with another dispatch?

src/status_im/ui/screens/wallet/events.cljs
goranjovic: same thing - why?

src/status_im/ui/screens/wallet/events.cljs
goranjovic: also

src/status_im/ui/screens/wallet/events.cljs
goranjovic: Would they be useful? If I mock up web3 the only thing that remains to be tested is equality check

goranjovic:
@yevh-berdnyk It seems that e2e tests already catch misconfigured tokens perfectly as it is. WGN is intentionally misconfigured in this PR and it's all over the test results :) 

cc @asemiankevich

asemiankevich:
seems autotest are stealing my job here! :D will play with this tomorrow and let you know @goranjovic

src/status_im/utils/ethereum/erc20.cljs
flexsurfer: I thought @yenda and @jeluard  introduced `ethereum/call-params` to replace web3 usage, and we want to get rid of web3

src/status_im/ui/screens/wallet/events.cljs
flexsurfer: oh ok, right, it's effect, then fine, thanks

src/status_im/ui/screens/wallet/events.cljs
flexsurfer: ok

src/status_im/utils/ethereum/erc20.cljs
goranjovic: @yenda @jeluard Any context guys?

src/status_im/data_store/realm/schemas/base/core.cljs
flexsurfer: thanks

yenda: ok but why go back to v15 and not v16 ?

src/status_im/ui/screens/wallet/events.cljs
yenda: it would make sense to at least call show-popup instead of raw js/alert

src/status_im/utils/ethereum/erc20.cljs
yenda: in that case it's still using the old ways call-params doesn't really replace web3 usage, @flexsurfer is probably refering to abi-spec/encode and abi-spec/decode functions

src/status_im/ui/screens/wallet/events.cljs
goranjovic: Sure, why not

src/status_im/data_store/realm/schemas/base/core.cljs
goranjovic: Oh that, good catch, merge mistake.

src/status_im/utils/ethereum/erc20.cljs
goranjovic: Ok so we're good to go then?

yenda: well you can replace the current thing by abi-spec/encode and abi-spec decode

asemiankevich:
image


@goranjovic could you unbreak the WGN token please?

src/status_im/utils/ethereum/erc20.cljs
goranjovic: that didn't work, I'd get garbled text after decode.

goranjovic:
@asemiankevich WGN is now great again, its spirit unbroken

asemiankevich:
@goranjovic tests number 1 and 5 are failed due this validation, do we really have those misconfigs?

goranjovic:
@asemiankevich No, all config is ok now. Let me check

goranjovic:
@asemiankevich the problem was happening when offline. Should be fixed now

PR #6627, opened 2018-11-07
Fixed CK

flexsurfer:
added `isMetaMask` to the provider, because CK requires it

### DIFF
--- a/src/status_im/utils/browser_config.edn
+++ b/src/status_im/utils/browser_config.edn
@@ -1,4 +1,4 @@
 ;; Browser (webview) configuration
 ;; :inject-js is a map of {<domain> <js-code>}, JS code will be injected if the domain name matches (www. is stripped away)
 
-{:inject-js {"cryptokitties.co" "; if (!window.chrome) { window.chrome = { webstore: true }; }"}}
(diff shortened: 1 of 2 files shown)


jeluard:
Maybe let's discuss things with CK first to understand the need to do that.

jeluard:
Oh I missed it is only for CK DApp.. Sounds good then

jeluard (approved):

src/status_im/utils/browser_config.edn
```diff
@@ -1,4 +1,4 @@
 ;; Browser (webview) configuration
 ;; :inject-js is a map of {<domain> <js-code>}, JS code will be injected if the domain name matches (www. is stripped away)
 
-{:inject-js {"cryptokitties.co" "; if (!window.chrome) { window.chrome = { webstore: true }; }"}}
+{:inject-js {"cryptokitties.co" "; if (!window.chrome) { window.chrome = { webstore: true }; window.ethereum.isMetaMask = web3.currentProvider.isMetaMask = true; }"}}
```
goranjovic, line 4: Just to double check, this will only get injected in the ck web page? i.e. other sites won't mistakenly think that we are metamask?

goranjovic: Never mind, just saw Julien's comment above

goranjovic (approved):

goranjovic:
But it makes sense to ping ck anyway. I see no reason to restrict it to metamask

PR #6629, opened 2018-11-07
add pip3, it's necessary for cross-compiling for windows

jakubgs:
This is necessary to use Conan within the image for: https://github.com/status-im/status-react/pull/6465


pedropombeiro:
@jakubgs We'll also need `python3-wheel` here, right?

pedropombeiro (approved):

yenda (approved):

Issue #5749, opened 2018-09-04
No incoming transactions for tokens on Mainnet are shown in Transaction History 


Serhy:
***Type***: Bug
***Summary***: 
1)There are no incoming transactions for tokens on Mainnet are displayed in the Transaction History. However, they are visible as mined (with 12confirations) in Etherscan. **Note**: amount for tokens in Wallet main screen displayed correct (as expected).
2) Outgoing token transactions are shown as `0 ETH` instead of token and amount

#### Expected behavior
Incoming transactions for tokens on Mainnet are shown in Transaction History 
#### Actual behavior
No incoming transactions for tokens on Mainnet are shown in Transaction History 

### Reproduction

- Open Status on Device1 and Device2
- Recover account with ETH and tokens in Mainnet on Device1 (Contact1)
- Create new account on Device2 (Contact2)
- Contact1 sends from 1-1 chat SNT to Contact2 (Note: use default gas limit and gas price so that transaction would could be mined properly) 
- Once transaction gets 12+ confirmations, check 'Transaction History' on Contact2 side whether there is 

### Additional Information
* Status version: Release 0.9.27, Develop 0.9.27, Release 0.2.26
* Operating System: Android and iOS

#### Logs

TF  session with logs for account that received tokens but they are not displayed in Transaction History and `0 ETH` instead tokens (see from the beginning of the session) https://app.testfairy.com/projects/4803622-status/builds/8579390/sessions/4400497345/?accessToken=f3c-LdDfNNQB8wrk5O3sXNAujlE

annadanchenko:
1. note that on Ropsten for same account all is working fine: can see incoming token transactions, can see outgoing token transactions with token and amount instead of 0 ETH. So, issue is only on mainnet
2. It was working fine in 0.9.26 and 0.9.25 - now when we check same scenario issue can be reproduced on these older releases too. So, it looks like there is some API change on etherscan or similar integration issue

Serhy:
Request to Etherscan API did not retrieve latest ERC20 transaction received by address:
For the address `https://api.etherscan.io/api?module=account&action=txlist&address=0xb9f914fe1c6edae2351fb42276868470083a3cd2&startblock=0&endblock=99999999&sort=desc&apikey=DMddddddddDMVP3H2STAMSAUV7BYFw&nocache=true`  this what was retrieved (by the 04 Sep 2018 16:24:25 GMT):
[...]

PR #6634, opened 2018-11-07
bug #5749 (part 1) - limited block range for incoming tokens detection

goranjovic:
partially addresses #5749 

### Summary:

Restricted the range of blocks where we get the event logs for token transfers to max 100000 in order to get around the infura timeouts.

What this means is that we're only scanning for about last two weeks of transactions, but at least they will appear in the history.

## Possible follow up from here

To get the older token transfers we can recursively call the same query with a moving 100k blocks large window all the way down to the block where the oldest token was created (possibly dropping tokens as we go beyond their own creation block). This would be a network intensive operation, so I don't want to do that every time user refreshes the list. Instead this should be done on account recovery. Will be provided in a separate PR.

### Testing notes (optional):

Only token transfers from the last 100k blocks (about two weeks) will appear.

**Functional**
- wallet / transactions

### Steps to test:
- Open Wallet on Mainnet
- Send some tokens to this phone
- Check the transaction history

status: ready

### DIFF
--- a/src/status_im/utils/ethereum/erc20.cljs
+++ b/src/status_im/utils/ethereum/erc20.cljs
@@ -145,6 +145,9 @@
        (error-fn error)
        (add-block-info web3 current-block-number chain direction result success-fn)))))
 
+;; we want to query over a limited number of blocks in order to avoid having timeouts
+(def block-query-limit 100000)
+
 ;;
 ;; Here we are querying event logs for Transfer events.
 ;;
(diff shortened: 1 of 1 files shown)


dmitryn (approved):

yenda (approved):

flexsurfer (approved):

asemiankevich:
@goranjovic token transactions are not shown on Ropsten for this PR.
1. Someday nightly screen
image

2. This Pr screen
image

Etherscan https://ropsten.etherscan.io/address/0x58d8c3D70ce4FA4b9fb10a665C8712238746F2ff#tokentxns

Could you please check?


<blockquote><div><strong><a href="https://ropsten.etherscan.io/address/0x58d8c3D70ce4FA4b9fb10a665C8712238746F2ff#tokentxns">Ropsten Accounts, Address And Contracts</a></strong></div><div>The Ethereum BlockChain Explorer, API and Analytics Platform</div></blockquote>

goranjovic:
@asemiankevich Good catch, my bad forgot about ropsten. :(

goranjovic:
Closing this PR as it's superseded by @bhauman 's changes

PR #6637, opened 2018-11-07
[TESTS] Removed skip mark for a test

### LINKED ISSUE #4829: Text input is broken for Korean, Chinese, Japanese characters
### Description

*Type*: Bug

*Summary*: With recent [react-native upgrade](https://github.com/status-im/status-react/commit/a2762eb19e45810184671b6c47f11faff0270124#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) we introduced a bug with [Chinese characters not being displayed correctly in chat input](https://github.com/facebook/react-native/issues/19339). The bug occurs only on iOS. 

#### Expected behavior
Users can use Korean, Chinese, Japanese keyboards in Status

#### Actual behavior
Users can't use Korean, Chinese, Japanese keyboards in Status due to https://github.com/facebook/react-native/issues/19339

### Reproduction

- Switch to a Chinese keyboard on iOS
- Open Status
- Send messages in Chinese in a chat

### Additional Information
* Status version: 0.9.20
* Worked in 0.9.19
* Operating System: iOS

hetvart:
### Summary 
the test under the PR had been marked with 'skip' flag since June. 
The related bug was fixed https://github.com/status-im/status-react/issues/4829, so I removed the mark.

### DIFF
--- a/test/appium/tests/atomic/chats/test_public.py
+++ b/test/appium/tests/atomic/chats/test_public.py
@@ -110,7 +110,6 @@ class TestPublicChatMultipleDevice(MultipleDeviceTestCase):
 @marks.chat
 class TestPublicChatSingleDevice(SingleDeviceTestCase):
 
-    @marks.skip
     @marks.testrail_id(5392)
     @marks.high
     def test_send_korean_characters(self):


Serhy (approved):

yevh-berdnyk (changes requested):
The test was skipped due to another reason: complete verification was not implemented.
We need a way to type characters one by one here, move cursor to the beginning of the input field, add more characters and then check the formatting.
So the current imlementation of the test isn't full and correct.

PR #6652, opened 2018-11-08
[Feature] Block user [Fix] current chat in bold on Linux [Fix] Username is bold on Linux

yenda:
fix #6033 remove logic from view
fix #1660 block user

- usernames in bold font on Linux
- prevent use of rtl char in username to impersonate
other other user random name
- current chat in bold font
- remove defview and subscriptions to use pure functions with params
- allow user to block/unblock a contact
- this will remove blocked user from chat list and
hid any message sent on any channel by blocked user

# Step to test

Plateform: desktop

sanity check
- test that chats, public chats and private chats are working correctly, responses are displayed properly
- test that users can be blocked in their profile

status: ready

PR #6655, opened 2018-11-08
Fixed styles for generated names on desktop

flexsurfer:
Fixed styles for generated names on desktop

image

### DIFF
--- a/src/status_im/ui/screens/desktop/main/chat/styles.cljs
+++ b/src/status_im/ui/screens/desktop/main/chat/styles.cljs
@@ -37,9 +37,13 @@
    :width              230})
 
 (def author
-  {:color         colors/black
-   :font-size     14
-   :font-weight   :bold})
+  {:color       colors/black
+   :font-weight "500"
+   :font-size   14})
+
+(def author-generated
+  {:color       colors/gray
+   :font-size   14})
 
 (defn chat-box [height]
   {:height            (+ height (* 2 chat-vertical-padding))
--- a/src/status_im/ui/screens/chat/utils.cljs
+++ b/src/status_im/ui/screens/chat/utils.cljs
@@ -8,7 +8,7 @@
             [status-im.utils.http :as http]))
 
 (defn format-author [from username]
-  (str (when username (str username " :: "))
+  (str (when username (str username " • "))
        (gfycat/generate-gfy from))) ; TODO: We defensively generate the name for now, to be revisited when new protocol is defined
 
 (defn format-reply-author [from username current-public-key]
--- a/src/status_im/ui/screens/desktop/main/chat/views.cljs
+++ b/src/status_im/ui/screens/desktop/main/chat/views.cljs
@@ -60,9 +60,11 @@
 
 (views/defview message-author-name [{:keys [from]}]
   (views/letsubs [incoming-name   [:contacts/contact-name-by-identity from]]
-    (let [name (chat-utils/format-author from incoming-name)]
-      [react/touchable-highlight {:on-press #(re-frame/dispatch [:show-contact-dialog from name (boolean incoming-name)])}
-       [react/text {:style styles/author :font :medium} name]])))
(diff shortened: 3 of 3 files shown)


yenda:
This will conflict with https://github.com/status-im/status-react/pull/6652 I suggest you rebase on it. Doing things the way #6652 does will prevent someone abusing rtl chars

flexsurfer:
@EugeOrtiz could you please take a look at the screenshot in the description? tbh I like it more when all 3-w names are grey, do no distract from messages @errorists what do you think also?

flexsurfer:
image

rcullito (approved):
nice :+1: seems to increase readability imho

errorists:
@flexsurfer 👏 Huge improvement! Ideally, we should use the medium font weight for usernames as bold is a bit fat, font-weight: 500 should do the trick.

flexsurfer:
@errorists agree, bold looks fat on desktop

vitvly (approved):

EugeOrtiz:
@flexsurfer @errorists  looking good! But the name is not bold, here are the design specs:

**Username**: Semibold 14 (weight 600) /  Line-height: 16 / Black
**Anonymous name**: Regular 14 (weight 400) / Line-height: 16 / Dark grey
**Dot**: Dark grey
**Paragraph/Message**: Regular 14 (weight 400) / Line-height: 20 / Black
**Timestamp**: Regular 10 (weight 400) /  Line-height: 12 / Dark Grey

[Designs on Figma](https://www.figma.com/file/MRYBI5axcTwL4Ft8xAYhFOfy/Desktop-Chat?node-id=134%3A1887) (select an element to see the specs)

Screenshot:
captura de pantalla 2018-11-26 a la s 9 08 30 a m

SlizkovaGanna:
Tested on Mac, Linux (emulated) - looks good.
1 of the fails I could not reproduce, the other is expected.

**Perhaps we should also consider changing the font in the reply bubble?**
For example, on a screenshot below the first quote is from a user added to contacts (the first name is displayed in bold font in chats), and the second quote is from a user not in contacts (name is displayed in grey)
@errorists @flexsurfer Maybe we should consider making the name in quote bubble also bold for future implementation?

image

EugeOrtiz:
+1 @SlizkovaGanna should we create a new issue or @flexsurfer can you apply it on this issue?
I added the specs to [Figma](https://www.figma.com/file/MRYBI5axcTwL4Ft8xAYhFOfy/Desktop-Chat?node-id=134%3A1887):

captura de pantalla 2018-11-26 a la s 10 29 12 a m

flexsurfer:
Bold
image



600
image




500
image

flexsurfer:
@EugeOrtiz @SlizkovaGanna  i changed from bold to 600, lets create new issues for quotes

src/status_im/ui/screens/desktop/main/chat/styles.cljs
```diff
@@ -37,9 +37,13 @@
    :width              230})
 
 (def author
-  {:color         colors/black
-   :font-size     14
-   :font-weight   :bold})
+  {:color       colors/black
+   :font-weight 600
```
yenda: font-weight has to be a string

flexsurfer: good catch, interesting why there is a difference between 600 and 500 then

yenda: Maybe it's only a problem on Linux

errorists: With font weight 400 is regular 500 is medium, 600 semi bold and 700 bold :)

flexsurfer:
@EugeOrtiz @SlizkovaGanna so semibold doesn't work on desktop, so i used medium, let's merge this PR because it took too much time for this change because current bold names look awful

PR #6706, opened 2018-11-13
[#5749] transaction history fixes

bhauman:
partially addresses #5749 

### Summary:

Restricted the range of blocks where we get the event logs for token transfers to max 100000 in order to get around the infura timeouts.

What this means is that we're only scanning for about last two weeks of transactions, but at least they will appear in the history.

This PR also fixes several bugs:
* the transaction polling process could easily die and not restart
* the confirmations were not being merged correctly leading to strange cases where a transaction is displayed as not confirmed when in fact it was
* prevented latent transaction queries from a different account from interfering with a newly changed account
* stopped continuous polling when transactions are present from other network in chat
* fixed miss-formatted url

## Possible follow up from here

To get the older token transfers we can recursively call the same query with a moving 100k blocks large window all the way down to the block where the oldest token was created (possibly dropping tokens as we go beyond their own creation block). This would be a network intensive operation, so I don't want to do that every time user refreshes the list. Instead this should be done on account recovery. Will be provided in a separate PR.

### Testing notes (optional):

Only token transfers from the last 100k blocks (about two weeks) will appear.

**Functional**
- wallet / transactions

### Steps to test:
- Open Wallet on Mainnet or Ropsten
- Send some tokens to this phone
- Check the transaction history

status: ready


src/status_im/models/transactions.cljs
```diff
@@ -1,116 +1,440 @@
...
+          ;; sanity timeout for work-fn
+          (async/alts! [finished-chan (async-util/timeout timeout-ms)])
+          (recur))))
+    do-now-chan))
+
+;; -----------------------------------------------------------------------------
+;; Helpers functions that help determine if a background sync should execute
+;; -----------------------------------------------------------------------------
+
+(defn- keyed-memoize
```
goranjovic, line 287: maybe put this one to `status-im.utils.core` it looks generally useful.

src/status_im/utils/ethereum/tokens.cljs
```diff
@@ -492,15 +492,18 @@
...
+                    :name     "KudosToken"
+                    :address  "0x93bb0afbd0627bbd3a6c72bc318341d3a22e254a"}])
 
    :custom  []})
 
-(defn tokens-for [chain]
-  (get all chain))
+(defn tokens-for
+  "makes sure all addresses are lower-case
+   TODO: token list should be speced and not accept non-lower-cased addresses"
```
goranjovic, line 504: Some of them were in mixed case intentionally, so that we'd be able to detect if everything still works with [checksummed](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md) addresses. I agree we should do that consistently though and only store normalized version.

goranjovic (approved):

src/status_im/models/transactions.cljs
bhauman: Yes this makes sense, I was waiting until the next round of work on transactions, to see if its actually needed.

src/status_im/utils/ethereum/tokens.cljs
bhauman: Oooooh, this makes sense. Hmmmm it would really be nice to know how and where checksummed contract addresses are supposed to flow through the System.

Makefile
```diff
@@ -51,14 +51,14 @@ _prepare-mobile: ##@prepare Install mobile platform dependencies and prepare wor
 	npm install
 
 $(STATUS_GO_IOS_ARCH):
-	cd $(RCTSTATUS_DIR) && curl -OL "$(GITHUB_URL)/$(STATUS_GO_VER)/status-go-ios.zip"
+	cd $(RCTSTATUS_DIR) && curl -o status-go-ios-$(STATUS_GO_VER).zip -OL "$(GITHUB_URL)/$(STATUS_GO_VER)/status-go-ios.zip"
```
yenda: can you put this change in a separate commit it is unrelated

yenda (approved):

Makefile
bhauman: In a separate PR?

yenda: No it can be in the same just don't squash them

bhauman: But yes I'll pull it out.

bhauman:
I have rebuilt this thing dozens of times and there is always some build that doesn't work.

asemiankevich:
so i have tested it both mainnet and rospten. I am able to see incoming tokens in transaction history. This is limited up to 2 weeks approximately so i dont see older transactions. 

Mainnet
image

Ropsten
image

I am fine to merge this PR with known limitations that will be done in future PRs.

PR #6721, opened 2018-11-14
Show chat name, source contact and time in PN body

pedropombeiro:

### Summary:

This PR adds some needed context to desktop PNs:

image

It is a simple change with no implications on other components.

#### Platforms
- macOS
- Linux
- Windows

**Functional**
- 1-1 chats
- public chats
- group chats

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -7,6 +7,7 @@
             [status-im.i18n :as i18n]
             [status-im.utils.core :as utils]
             [status-im.utils.config :as config]
+            [status-im.utils.contacts :as utils.contacts]
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.datetime :as time]
             [status-im.transport.message.group-chat :as message.group-chat]
@@ -14,6 +15,7 @@
             [status-im.chat.models.loading :as chat-loading]
             [status-im.chat.models.message-content :as message-content]
             [status-im.chat.commands.receiving :as commands-receiving]
+            [status-im.chat.subs :as chat-subs]
             [status-im.utils.clocks :as utils.clocks]
             [status-im.utils.money :as money]
             [status-im.utils.types :as types]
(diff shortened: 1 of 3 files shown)


EugeOrtiz:
@PombeirP Can you provide examples of how they will look for 1:1 chat messages and replies in a public channel? 
I wouldn't worry about general public channel PNs, since [we want to limit them](https://github.com/status-im/status-react/issues/6445).
In any case, you can see the proposed [PNs format here.](https://github.com/status-im/status-react/issues/6621)

pedropombeiro:
There are a few considerations done in the logic:
- Timestamp is only shown if message is older than 15 seconds
- Contact name is only shown if different than chat name (i.e. multi-user chats)

src/status_im/chat/models/message.cljs
```diff
@@ -100,7 +100,11 @@
                (not= from current-public-key)
                (get-in db [:account/account :desktop-notifications?])
                (< (time/seconds-ago (time/to-date timestamp)) constants/one-earth-day))
-      (.sendNotification react/desktop-notification (:text content)))
+      (let [chat-name     (get-in db [:chats chat-id :name])
```
pedropombeiro: @yenda I guess I should use `:get-chat-name` here?

maxhora (approved):

vkjr (approved):

churik:
Amazing work @PombeirP!

Some issues:
1) Cannot see three-random-name of person who is not added to your contact list (Mac Os, Linux)
pns 1

2) On Linux - time is missing on PNs
linux_time

3) As for me would be nice to have the same format in both platform.
For now:
linux_2
I would prefer to have the same view for Linux as it is for Mac OSx.

and separate thank you for nice icon on linux!

pedropombeiro:
Still waiting for input from @yenda about that. Time is not missing, see the 15 sec rule.

churik:
@PombeirP I saw it, but it works differently on Linux and Mac OSx.
On last screen, you can see that the same message for same account contains a timestamp for Mac OSx and doesn't for Linux (I had opened app for Mac OSx and for linux VM at the same time to compare it)
For Mac OSx as Ia can observe timestamps are shown regardless rule of 15 sec.

pedropombeiro:
> and separate thank you for nice icon on linux!

My pleasure, I'm a Linux user and it had been bothering me for a long time!

> I would prefer to have the same view for Linux as it is for Mac OSx. 

I don't understand, there is absolutely nothing in the code that could differentiate between platforms, all 3 desktop platforms run exactly the same code. Since I run Linux, I know that it is showing for me. Unless there is some platform-specific bug in `(time/seconds-ago (time/to-date timestamp))`. Any clues @yenda?

@churik please do one test, sign off the account, send multiple messages from another device to it, wait 15 secs, then sign in and minimize the app. All the notifications should have the time on them.

pedropombeiro:
@churik From what I can understand from the code, it looks like we'll need to do a small refactoring to make it possible to show the anonymous name in the notification, so we have two options: we leave this open until someone from the Clojure team can do that refactoring, or we merge and file a bug against this behavior. Either way is fine by me.

pedropombeiro:
@yenda tells me that his branch already contains the change required for this, so maybe he could fix this there after this is merged to `develop`.

pedropombeiro:
@churik a new build is underway which fixes the missing contact name.

pedropombeiro:
- Refactored the building of the desktop notification out of the `add-message` method for more clarity/separation of concerns. Would appreciate a second review.
- Added support for notification priority. 1:1 messages will have higher priority than others.

pedropombeiro:
OK, @churik build #10 should be OK to test. I had to remove markup since some platforms showed the HTML tags

churik:
@pablodip 
now it is working exactly as you described.
Thank you again.
Ready to merge.

Issue #5975, opened 2018-09-24
Sending a transaction from 1-1 chat with wallet set up ends up on Wallet 'Set up your screen' 


yevh-berdnyk:
[comment]: # (Please replace ... with your information. Remove < and >)

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
***Type***: Bug

***Summary***: User is navigated to Wallet view instead of 1-1 chat if sending funds from 1-1 chat and performing wallet set up after clicking Send message button

#### Expected behavior
User is returned back to the 1-1 chat

#### Actual behavior
User is landed on the Wallet view with 'Set up your Wallet' screen (while Wallet set up already while sending TX in 1-1 chat!)

### Reproduction

- Open Status
- Recover existing account with funds
- Start new 1-1 chat
- Click Send command, select an asset, input valid amount and click Send message button
- Set up your wallet
- Sign transaction and click Got it! button

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: 0.9.29 (8943); node 0.15.0
* Operating System: OnePlus 5, Android 8.1

### Additional Information **(Updated)**
* Status version: 0.11.0 Develop
* Operating System: Android and iOS

goranjovic:
The previous fix for this caused another bug - #6283 reopened this one

antdanchenko:
@goranjovic please take a look at this issue, this one is the last issue in develop branch which should be fixed in order to get 100% passed tests for any PR

goranjovic:
@antdanchenko on it :+1:

oskarth:
@goranjovic What's the latest state of this?

I saw:
> https://github.com/status-im/status-react/pull/6878#issuecomment-445156497

but not quite clear what this means or implies.

We want automated tests to block as soon as possible. IMO we have two options to get there:
1) Fix this now before end of year (issue high prio)
2) Remove this test from automated test for now (issue not high prio due to redesign, automated tests cutting scope)

@antdanchenko

antdanchenko:
@oskarth, @goranjovic 
I am not a big fan of adding workarounds to an automated tests, or disabling tests which fail due to product issues.
Using the approach might lead to the situation when half of the tests are disabled, we always get `100% passed`, but half of the functionality is not tested.

However for this specific case I will add workaround, to make wallet setup before sending a transaction.
Also new test will be added for this specific behavior but will get lower priority which means that the test will run only against nightly builds

oskarth:
[...]

Issue #3960, opened 2018-04-23
Attempt to read from field 'android.animation.Animator android.animation.AnimatorSet$Node.mAnimation' on a null object reference


asemiankevich:
### Description

*Type*: Bug

*Summary*: cant reproduce it but i had a crash with `Attempt to read from field 'android.animation.Animator android.animation.AnimatorSet$Node.mAnimation' on a null object reference` when attempting to send funds in a group chat (android)

#### Expected behavior
no crashes

#### Actual behavior

`android.content.res.Resources$NotFoundException: Drawable android:drawable/seekbar_thumb_material_anim with resource ID #0x1080658
Caused by: java.lang.NullPointerException: Attempt to read from field 'android.animation.Animator android.animation.AnimatorSet$Node.mAnimation' on a null object reference
	at android.animation.AnimatorSet.clone(AnimatorSet.java:725)
	at android.animation.AnimatorSet.clone(AnimatorSet.java:682)
	at android.animation.Animator$AnimatorConstantState.newInstance(Animator.java:589)
	at android.animation.Animator$AnimatorConstantState.newInstance(Animator.java:588)
	at android.content.res.ConstantState.newInstance(ConstantState.java:52)
	at android.content.res.ConstantState.newInstance(ConstantState.java:60)
	at android.content.res.ConfigurationBoundResourceCache.getInstance(ConfigurationBoundResourceCache.java:40)
	at android.animation.AnimatorInflater.loadAnimator(AnimatorInflater.java:116)
	at android.graphics.drawable.AnimatedVectorDrawable$AnimatedVectorDrawableState$PendingAnimator.newInstance(AnimatedVectorDrawable.java:666)
	at android.graphics.drawable.AnimatedVectorDrawable$AnimatedVectorDrawableState.inflatePendingAnimators(AnimatedVectorDrawable.java:644)
	at android.graphics.drawable.AnimatedVectorDrawable.applyTheme(AnimatedVectorDrawable.java:454)
	at android.graphics.drawable.DrawableContainer$DrawableContainerState.applyTheme(DrawableContainer.java:960)
	at android.graphics.drawable.DrawableContainer.applyTheme(DrawableContainer.java:611)
	at android.graphics.drawable.StateListDrawable.applyTheme(StateListDrawable.java:371)
	at android.graphics.drawable.AnimatedStateListDrawable.applyTheme(AnimatedStateListDrawable.java:392)
[...]

PR #3861, opened 2018-04-13
[#3850] Add Aragon Test (ATT) to testnet wallet

kwingram25:
Fixes #3850

### Summary:
Adds [Aragon Test Token](https://ropsten.etherscan.io/token/0xfff28fb56b2a53597c55e1a767c906f339335025) to default asset list on Ropsten/Rinkeby

### Steps to test:
- Open Status
- Switch to testnet
- Visit Wallet tab
- ATT should appear in list

Status: Ready

PR #6722, opened 2018-11-14
[slow sign in] fix slow sign in (first iteration)

rasom:
### Summary:

Signing in is slow for account with tons of messages. This is caused by many reasons ([see this note](https://notes.status.im/VxmgW_rJQkqHsPoqGiMGxQ)), this PR fixes some of them:
> - Remove necessity in the usage of `deduplication-ids` set (part of **A**). In order to tackle deduplication problem proper `message-id` (which will be calculated the same way regardless by which version of the app it has been received) can be calculated ([this function was suggested](https://github.com/status-im/status-react/blob/26ec3f8cd7906a0a80cd75a937e8c8028a358fa9/src/status_im/data_store/messages.cljs#L59)) and updated in all existing messages via db migration. This is done in https://github.com/status-im/status-react/pull/6722
> - Remove necessity in the loading of all `message-ids` (another part of **A**). We can check if the message has been already received by making a direct query to db. While query which is done using [`.objects()`](https://realm.io/docs/javascript/latest/api/Realm.html#objects) with `message-id=...` condition is quite slow, [`objectForPrimaryKey`](https://realm.io/docs/javascript/latest/api/Realm.html#objectForPrimaryKey) works acceptably faster (~50ms fro 1000 queries). Implemented in https://github.com/status-im/status-react/pull/6722. Other suggestions on this are welcome.
> - Load part of data necessary for `:home` screen asynchronously. In this case we can go to **7** without doing **8**. That's the only way to guaranty that `sign in` will have some likely constant regardless of db size. Done in https://github.com/status-im/status-react/pull/6722 (probably touches **E** as no need to calculate it before rendering screen)

### Review notes (optional):

- `message-id` is calculated as `sha3(from + chat-id + clock-value)` for each message
- [`objectForPrimaryKey`](https://realm.io/docs/javascript/latest/api/Realm.html#objectForPrimaryKey) is used to check if `message-id` already exists in DB
- `:init-chats` event is dispatched 100ms after `:home`'s `componentDidMount` on mobile. The reason for this is that otherwise on Android event will handled before `:home` is actually rendered and it will be shown only after `:init-chats`. 
- migration resets `messages-id` in `message` (`:messages-id`, `:reference-id` in `:content`), `messages-status` (`:status-id`, `:messages-id`) entities, let me now if any other reference was missed
- during transition period we will lost all refernces to messages which are done on older clients (as there is no way to calculate new `message-id` from the old one directly, without the message)

#### Platforms (optional)
- All platforms

**Functional**
- signing in 
- 1-1 chats
- public chats
- group chats

### Steps to test:
- Open Status
- create a n account
- join several pub chats
- send ~10000 messages in these pubchats (also it is slower if all these messages are unread)
- logout
- sign in back and check how fast it happens

status: ready

### DIFF
--- a/src/status_im/accounts/login/core.cljs
+++ b/src/status_im/accounts/login/core.cljs
@@ -33,7 +33,7 @@
       (catch (fn [error]
                (log/warn "Could not change account" error)
                ;; If all else fails we fallback to showing initial error
-               (re-frame/dispatch [:init.callback/account-change-error])))))
+               (re-frame/dispatch [:init.callback/account-change-error (str error)])))))
 
 ;;;; Handlers
 (fx/defn login [cofx]
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -27,7 +27,8 @@
             [status-im.utils.platform :as platform]
             [status-im.ui.components.react :as react]
             [status-im.utils.fx :as fx]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [status-im.data-store.messages :as messages-store]))
 
 (defn- wrap-group-message
   "Wrap a group message in a membership update"
@@ -198,14 +199,13 @@
             (chat-loading/group-chat-messages chat-id (get chat->message chat-id))))
 
 (defn- add-to-chat?
-  [{:keys [db]} {:keys [chat-id clock-value message-id from] :as message}]
-  (let [deduplication-id (messages-store/deduplication-id from chat-id clock-value)
-        {:keys [deleted-at-clock-value messages not-loaded-message-ids deduplication-ids]}
+  [{:keys [db]} {:keys [chat-id clock-value message-id from]}]
+  (let [{:keys [deleted-at-clock-value messages not-loaded-message-ids]}
         (get-in db [:chats chat-id])]
     (not (or (get messages message-id)
              (get not-loaded-message-ids message-id)
-             (get deduplication-ids deduplication-id)
-             (>= deleted-at-clock-value clock-value)))))
+             (>= deleted-at-clock-value clock-value)
+             (messages-store/message-exists? message-id)))))
 
 (defn- filter-messages [cofx messages]
   (:accumulated (reduce (fn [{:keys [seen-ids] :as acc}
@@ -290,8 +290,9 @@
 
 (fx/defn upsert-and-send [{:keys [now] :as cofx} {:keys [chat-id] :as message}]
   (let [send-record     (protocol/map->Message (select-keys message transport-keys))
-        message-id      (transport.utils/message-id send-record)
+        message-id      (transport.utils/message-id message)
         message-with-id (assoc message :message-id message-id)]
+
     (fx/merge cofx
               (chat-model/upsert-chat {:chat-id chat-id
                                        :timestamp now})
--- a/src/status_im/chat/specs.cljs
+++ b/src/status_im/chat/specs.cljs
@@ -9,7 +9,7 @@
 (s/def :chat/animations (s/nilable map?))                         ; {id (string) props (map)}
 (s/def :chat/chat-ui-props (s/nilable map?))                      ; {id (string) props (map)}
 (s/def :chat/chat-list-ui-props (s/nilable map?))
-(s/def :chat/layout-height (s/nilable number?))                   ; height of chat's view layout 
+(s/def :chat/layout-height (s/nilable number?))                   ; height of chat's view layout
 (s/def :chat/selected-participants (s/nilable set?))
 (s/def :chat/public-group-topic (s/nilable string?))
 (s/def :chat/public-group-topic-error (s/nilable string?))
@@ -18,7 +18,6 @@
 (s/def :chat/message-statuses (s/nilable map?))                   ; message/user statuses indexed by two level index
 (s/def :chat/not-loaded-message-ids (s/nilable set?))             ; set of message-ids not yet fully loaded from persisted state
 (s/def :chat/referenced-messages (s/nilable map?))                ; map of messages indexed by message-id which are not displayed directly, but referenced by other messages
-(s/def :chat/deduplication-ids (s/nilable set?))                  ; set of helper deduplication ids
 (s/def :chat/last-clock-value (s/nilable number?))                ; last logical clock value of messages in chat
 (s/def :chat/loaded-chats (s/nilable seq?))
 (s/def :chat/bot-db (s/nilable map?))
@@ -26,3 +25,4 @@
 (s/def :chat/cooldown-enabled? (s/nilable boolean?))
 (s/def :chat/last-outgoing-message-sent-at (s/nilable number?))
 (s/def :chat/spam-messages-frequency (s/nilable number?))         ; number of consecutive spam messages sent
+(s/def :chats/loading? (s/nilable boolean?))
--- a/src/status_im/group_chats/core.cljs
+++ b/src/status_im/group_chats/core.cljs
@@ -115,7 +115,11 @@
                                 :dsts          members
                                 :success-event [:transport/message-sent
                                                 chat-id
-                                                (transport.utils/message-id (:message payload))
+                                                (transport.utils/message-id
+                                                 ;; NOTE: There is no clock-value here.
+                                                 ;; Will we have collision?
+                                                 {:from current-public-key
+                                                  :chat-id chat-id})
                                                 :group-user-message]
                                 :payload       payload}}))))
 
--- a/src/status_im/transport/utils.cljs
+++ b/src/status_im/transport/utils.cljs
@@ -19,8 +19,10 @@
 
 (defn message-id
   "Get a message-id"
-  [message]
-  (sha3 (pr-str message)))
+  [{:keys [from chat-id clock-value] :as m}]
+  {:pre [(not (nil? from))
+         (not (nil? chat-id))]}
+  (sha3 (str from chat-id clock-value)))
 
 (defn get-topic
   "Get the topic of a group chat or public chat from the chat-id"
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -52,35 +52,48 @@
               (filter #(not (contains? message-id->messages %))))
         (vals message-id->messages)))
 
+(fx/defn load-chats-messages
+  [{:keys [db get-stored-messages get-stored-user-statuses
+           get-referenced-messages get-stored-unviewed-messages]
+    :as cofx}]
+  (let [chats (:chats db)]
+    (fx/merge
+     cofx
+     {:db (assoc
+           db :chats
+           (reduce
+            (fn [chats chat-id]
+              (let [stored-unviewed-messages (get-stored-unviewed-messages (accounts.db/current-public-key cofx))
+                    chat-messages (index-messages (get-stored-messages chat-id))
+                    message-ids   (keys chat-messages)]
+                (update
+                 chats
+                 chat-id
+                 assoc
+
+                 :messages chat-messages
+                 :message-statuses (get-stored-user-statuses chat-id message-ids)
+                 :unviewed-messages (get stored-unviewed-messages chat-id)
+                 :referenced-messages (into {}
+                                            (map (juxt :message-id identity)
+                                                 (get-referenced-messages
+                                                  (get-referenced-ids chat-messages)))))))
+            chats
+            (keys chats)))}
+     (group-messages))))
+
 (fx/defn initialize-chats
   "Initialize all persisted chats on startup"
-  [{:keys [db default-dapps all-stored-chats get-stored-messages get-stored-user-statuses
-           get-stored-unviewed-messages get-referenced-messages stored-message-ids
-           stored-deduplication-ids] :as cofx}]
-  (let [stored-unviewed-messages (get-stored-unviewed-messages (accounts.db/current-public-key cofx))
-        chats (reduce (fn [acc {:keys [chat-id] :as chat}]
-                        (let [chat-messages (index-messages (get-stored-messages chat-id))
-                              message-ids   (keys chat-messages)
-                              unviewed-ids  (get stored-unviewed-messages chat-id)]
-                          (assoc acc chat-id
-                                 (assoc chat
-                                        :unviewed-messages unviewed-ids
-                                        :messages chat-messages
-                                        :message-statuses (get-stored-user-statuses chat-id message-ids)
-                                        :deduplication-ids (get stored-deduplication-ids chat-id)
-                                        :not-loaded-message-ids (set/difference (get stored-message-ids chat-id)
-                                                                                (set message-ids))
-                                        :referenced-messages (into {}
-                                                                   (map (juxt :message-id identity)
-                                                                        (get-referenced-messages
-                                                                         (get-referenced-ids chat-messages))))))))
+  [{:keys [db default-dapps all-stored-chats] :as cofx}]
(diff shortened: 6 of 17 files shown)


chadyj:
What clients are slow to sign in?

My iOS is almost instant, but Mac desktop takes a long time.

rasom:
@chadyj well, any of them can be slow right now because duration of sign in depends on iteration over ALL messages in db. So assuming you have enough messages it might be 20-30 seconds or even longer. At least @jarradh and @PombeirP are experiencing this problem, and it can be reproduced. 

@PombeirP:
> I'm seeing this long wait time both on my laptop (ThinkPad X1C6) and mobile phone (Huawei P20 Pro), so not slow devices. It only happens if I unlock an account that has tons of messages. If I unlock an account which has lots of channels open but no messages, it opens instantly (on same device and app instance), so it really feels like it is something tied with the handling of messages (maybe sorting?)

so the reason for the recent preformance regression is this code https://github.com/status-im/status-react/blob/26ec3f8cd7906a0a80cd75a937e8c8028a358fa9/src/status_im/data_store/messages.cljs#L65 , but also we iterate all messages here https://github.com/status-im/status-react/blob/26ec3f8cd7906a0a80cd75a937e8c8028a358fa9/src/status_im/data_store/messages.cljs#L43 and this happens during sign in as well 

this PR already contains the fix for the first iteration mentioned, but i'm still working on the second.

Besides this problem, even on empty account sign in takes 3.2-3.5s on galaxy s9 (so it might be slower on many other android devices) which is not perfect, as i can say. There are ways to improve this duration as well, but i'll probably focus on this in another PR.

pedropombeiro:
I downloaded the build to try it, but after 7 seconds I got the warning that it couldn't decrypt the db so I'll wait until it is rebased on latest develop.

rasom:
thanks for trying, @PombeirP!
The first run will be slow because of migrations.
I will rebase it soon, but be careful: until we merge it to develop it might introduce some breaking changes so that you will not be able to upgrade to the next nightly build. If you want to test it make sure you have a backup of the database.

dmitryn (approved):

pedropombeiro (approved):
Can't speak for the code aspect, but I tested on my local db and it migrated fine and signing in was much faster in subsequent times.

lukaszfryc:
> send ~10000 messages in these pubchats (also it is slower if all these messages are unread)

@rasom is there any way I can generate such amount of messages? I can always use a UI script but it will take a long time to send it via UI.

If we have a CLI script that would do it (bot or something), I could write a e2e test to measure sign in time for every nightly/PR.

mandrigin:
oh, I wanted to tell about the bot, but it seems not to be working again :(

src/status_im/accounts/login/core.cljs
```diff
@@ -9,7 +9,8 @@
             [status-im.utils.keychain.core :as keychain]
             [status-im.utils.types :as types]
             [taoensso.timbre :as log]
-            [status-im.utils.security :as security]))
+            [status-im.utils.security :as security]
+            [status-im.utils.datetime :as time]))
```
yenda: is it imported but not used ?

src/status_im/chat/specs.cljs
```diff
@@ -9,20 +9,20 @@
...
 (s/def :chat/selected-participants (s/nilable set?))
 (s/def :chat/public-group-topic (s/nilable string?))
 (s/def :chat/public-group-topic-error (s/nilable string?))
 (s/def :chat/messages (s/nilable map?))                           ; messages indexed by message-id
 (s/def :chat/message-groups (s/nilable map?))                     ; grouped/sorted messages
 (s/def :chat/message-statuses (s/nilable map?))                   ; message/user statuses indexed by two level index
 (s/def :chat/not-loaded-message-ids (s/nilable set?))             ; set of message-ids not yet fully loaded from persisted state
-(s/def :chat/referenced-messages (s/nilable map?))                ; map of messages indexed by message-id which are not displayed directly, but referenced by other messages
-(s/def :chat/deduplication-ids (s/nilable set?))                  ; set of helper deduplication ids
+(s/def :chat/referenced-messages (s/nilable map?))                ; set of helper deduplication ids
```
yenda: it's the wrong comment

src/status_im/chat/models/message.cljs
```diff
@@ -198,14 +199,13 @@
...
-        {:keys [deleted-at-clock-value messages not-loaded-message-ids deduplication-ids]}
+  [{:keys [db]} {:keys [chat-id clock-value message-id from]}]
+  (let [{:keys [deleted-at-clock-value messages not-loaded-message-ids]}
         (get-in db [:chats chat-id])]
     (not (or (get messages message-id)
              (get not-loaded-message-ids message-id)
-             (get deduplication-ids deduplication-id)
-             (>= deleted-at-clock-value clock-value)))))
+             (>= deleted-at-clock-value clock-value)
+             (messages-store/message-exists? chat-id)))))
```
yenda: shouldn't it be message-id? also does it mean we trade initial loading time for a query for each message received ?

src/status_im/group_chats/core.cljs
```diff
@@ -115,7 +115,11 @@
                                 :dsts          members
                                 :success-event [:transport/message-sent
                                                 chat-id
-                                                (transport.utils/message-id (:message payload))
+                                                (transport.utils/message-id
+                                                 ;; NOTE: There is no clock-value here.
+                                                 ;; Will we have collision?
+                                                 {:from current-public-key
```
yenda, line 121: there will definitely be collisions with these params what would be the consequences?

src/status_im/transport/utils.cljs
```diff
@@ -19,8 +19,10 @@
 
 (defn message-id
   "Get a message-id"
-  [message]
-  (sha3 (pr-str message)))
+  [{:keys [from chat-id clock-value] :as m}]
```
yenda, line 22: I don't think we should change message-ids like this considering the consequences we had to deal with the last time we did it. If we do the last time it was discussed it seemed that a form of UUID was the most reasonable choice over any kind of computed id

yenda (changes requested):

yenda:
> :init-chats event is dispatched 100ms after :home's componentDidMount on mobile. The reason for this is that otherwise on Android event will handled before :home is actually rendered and it will be shown only after :init-chats.

what is the problem with that?

src/status_im/chat/models/message.cljs
rasom: oh right

src/status_im/transport/utils.cljs
rasom: Well, that's nice but if it's not deterministic we will have to maintain deduplication stuff. So if that's an issue we still can add UUID later and maintain `old-message-id` or something for a transition period. If introduced way to compute `message-id` is not worse than the one we had before this PR, it should be good enough. Other stuff is a bit out of the scope.

src/status_im/group_chats/core.cljs
rasom: Idk, it seems that we are not storing those messages in DB or i am wrong?

rasom:
@lukaszfryc yes sure i can generate as many as you need

rasom:
@lukaszfryc check pub chats #tchat1000 #tchat2000 #tchat4000 #tchat8000

src/status_im/accounts/login/core.cljs
rasom: fixed

src/status_im/chat/specs.cljs
rasom: fixed

src/status_im/chat/models/message.cljs
rasom: fixed.
Yes, we do a query for each message and that's totally fine comparing to iterating all messages in realm db. Also `objectForPrimaryKey` is quite fast.

rasom:
@yenda 
> what is the problem with that?

Do you mean what is the problem with blocked UI and no visual response for the user or what is the problem with having a timeout there?

rasom:
@yenda fixes according to some comments added, see responses for the rest of them

yenda:
> @yenda
> 
> > what is the problem with that?
> 
> Do you mean what is the problem with blocked UI and no visual response for the user or what is the problem with having a timeout there?

I mean it seems odd that some data processing has to be delayed because of UI, I don't understand what is happening exactly here.

rasom:
> I mean it seems odd that some data processing has to be delayed because of UI, I don't understand what is happening exactly here.

again, odd because it shouldn't happen or odd because here we prefer UI over data processing?
This `:init-chats` thing right now iterates all chats we have in realm db, it means that it can last forever assuming we have enough chats. So the purpose of this timeout is to give the user feedback that account is actually unlocked but chats haven't been loaded yet.

mandrigin:
> I mean it seems odd that some data processing has to be delayed because of UI, I don't understand what is happening exactly here.

I would say that is is not odd at all. We are building a product for a user who interact with UI. Data processing is not some self-serving holy goal, it is just a way for us to bring some value to a user. So we should by no mean make user suffer just because we want our internal structure to be more straightforward or simple or "done right."

rasom:
@lukaszfryc most likely duration of processing of received message has degraded because we are making a query to realm db for each message, it would be great to figure what's the difference comparing to nightlies (when receiving 1000 messages in a row for instance).

rasom:
@lukaszfryc windows build is failing, 
latest versions
android: https://ci.status.im/job/status-react/job/combined/job/mobile-android/7854/
linux: https://ci.status.im/job/status-react/job/combined/job/desktop-linux/3861/
macos: https://ci.status.im/job/status-react/job/combined/job/desktop-macos/3875/
ios: https://ci.status.im/job/status-react/job/combined/job/mobile-ios/3960/


<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/job/mobile-android/7854/">status-react » combined » mobile-android #7854 [Jenkins]</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/job/desktop-linux/3861/">status-react » combined » desktop-linux #3861 [Jenkins]</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/job/desktop-macos/3875/">status-react » combined » desktop-macos #3875 [Jenkins]</a></strong></div></blockquote>
<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/combined/job/mobile-ios/3960/">status-react » combined » mobile-ios #3960 [Jenkins]</a></strong></div></blockquote>

lukaszfryc:
@rasom on iPhone 8, when I have all 4 chats (tchat1000, tchat2000, tchat4000, tchat8000 so 15k messages in total) logging in takes ~6s on nightly vs ~1s on this build. So definitely a big improvement. I'm checking on Android device and macOS now.

lukaszfryc:
@rasom same scenario for desktop (macbook pro); took 5s now it takes ~1s. For Android (Galaxy S8) it took ~20s to log in, now it takes ~5s to log in and show list of chats, **but app is not responsive** for another 5s. Maybe would be better to keep showing spinner until we know that app is responsive?

rasom:
In which state it is not responsive? When chats are shown but with "No messages"? If so we have next options
1) show a spinner while chats are loading and show them only when messages are loaded too
2) add a separate spinner for each chat when the list is ready but the rest of the data related to chats is not 
3) might be another option you will suggest :)

lukaszfryc:
@rasom right after logging there is a spinner for a moment, then list of chats appears but it's not clickable. The app becomes responsive when messages are loaded.

> show a spinner while chats are loading and show them only when messages are loaded too

This would be fine I think. Can you implement it in scope of this PR?

rasom:
@lukaszfryc yes i can, will add it soon

rasom:
@lukaszfryc, it should be ready now
@yenda, can you please proceed with review?

yenda:
I would like to have @cammellos input on this one can it wait until he's back ?

rasom:
@yenda, no that's a bad idea, we will have a week delay in this case. Can you elaborate please what exactly is the biggest concern for you here?

yenda:
@rasom the biggest painpoint for me is the message-id generation, I don't think they need to be deterministic and we could send them along with the messages. For group chats I would like to clarify whether the collisions is a problem or not, it's not just a risk it will inevitably happen, and afaiu it means previous messages with  the same id are going to be overwritten ?

lukaszfryc:
@rasom spinner works good now. But, please hide that debug info that is shown iOS:

rasom:
@yenda 
>  I don't think they need to be deterministic and we could send them along with the messages

well i agree and that's how it was implemented in the first place. The problem is that we can't just stop using them right now without some extra effort, because users will end up with different ids for the same messages (old messages which are store in realm) and `reference-to` thing will be broken.

As a next separate PR it looks pretty straightforward to change `message-id` fn to something like
```clojure
(defn message-id
  "Get a message-id"
  [{:keys [from chat-id clock-value message-id] :as m}]
  {:pre [(not (nil? from))
         (not (nil? chat-id))]}
  (or message-id
      (sha3 (str from chat-id clock-value))))
```
and make sure `message-id` is generated for each message before sending, i just don't want to have it in the scope of this PR because it has consequences for testing. Better to work on this separately. 
Does this PR make the way how ids are calculated worse then we have in `develop`? If no, it's good enough to merge it, because this PR solves different problem.

> For group chats I would like to clarify whether the collisions is a problem or not

the place where collision might happen is `send-membership-update` and this message is not stored in realm db. So shouldn't be a problem.

rasom:
@lukaszfryc this can be reproduced in `develop` as well, at least i saw this in some discussions in chats. So i would suggest fixing this a separate thing because it has nothing to do with this PR.

rasom:
@lukaszfryc see #test-team chat

asemiankevich:
@mandrigin seems you can assist with https://github.com/status-im/status-react/pull/6722#issuecomment-440592523 (please address in future PR, not this one)

lukaszfryc:
> @lukaszfryc see #test-team chat

OK, so feel free to merge once review is done.

yenda:
> the place where collision might happen is send-membership-update and this message is not stored in realm db. So shouldn't be a problem.

I don't understand on the recipient end this message isn't this message stored? If that's the case is there a clock-value there to differentiate messages from the same sender and avoid overwritting the previous ones?

yenda:
Also I agree about the rest it seems like a reasonable change and a broader fix for message-ids can be done later. If the group chat thing is a non-issue I'll remove by request for changes

rasom:
> if that's the case is there a clock-value there to differentiate messages from the same sender and avoid overwritting the previous ones?

the way to test it is to create a group chat, make sure that update was sent and check if there are messages with `clock-value = 0` in db. Will check this later today.

cammellos:
definitely do not wait for me, the message id in that case is used to show
the sent status in the callback, just send a group chat message and check
that the spinner goes away, collision might be a problem if multiple
messages are sent in quick succession, but why can't we just leave the code
as it was?

rasom:
> but why can't we just leave the code as it was?

do you mean the way how that id was generated for membership update or which part of code?

yenda (approved):

src/status_im/group_chats/core.cljs
cammellos: i mean here, why can't we pass the message as before, instead of selecting keys?

rasom:
@lukaszfryc did you test migration from the older version of app?

antdanchenko:
there are 2 tests failed in latest test run and both of them are known issues:
#6825
#5975

lukaszfryc:
@rasom yes, and I just re-tested it on Android. Works fine.

rasom:
thanks @lukaszfryc

PR #6746, opened 2018-11-15
[WIP] Added geolocation event

### LINKED ISSUE #6602: Add geolocation event
### User Story

As a developer, I want an event to access user current location. This will require proper permissions.

See https://github.com/status-im/status-react/blob/develop/src/status_im/extensions/core.cljs

### Description

*Type*: Feature

aashutoshrathi:
If you submit PR for issue with bounty then write here Fixes #6602 where NN is issue number

### Summary:

Imported Geolocation API and added event


cr2fx:
> I understand that this is an application and I should wait for an approval from the funder before starting work.

I guess you're not supposed to start work on the bounty before approval?

PR #6752, opened 2018-11-16
Disable send button when no peers are connected.

cammellos:
Previously we were only checking whether `network-status` is equal to
`:offline`, therefore allowing the user to send messages even when no
peers are connected (i.e. the app is displaying Connecting to peers..).

Another potential issue (need to verify still though), is that we always check `network-status` against `:offline`, but it is initialized to `nil`. So if you start the app when you are offline, it will not report the network status correctly, until there's a network change. need to verify this though.

### Testing

When you see Offline or Connecting to peers, the send button in chats should be disabled.

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/input/send_button.cljs
+++ b/src/status_im/ui/screens/chat/input/send_button.cljs
@@ -14,20 +14,20 @@
        (animation/timing spin-value {:toValue  to-spin-value
                                      :duration 300})))))
 
-(defn sendable? [input-text network-status]
+(defn sendable? [input-text offline?]
   (let [trimmed (string/trim input-text)]
     (not (or (string/blank? trimmed)
              (= trimmed "/")
-             (= :offline network-status)))))
+             offline?))))
 
 (defview send-button-view []
   (letsubs [{:keys [command-completion]}            [:chats/selected-chat-command]
             {:keys [input-text seq-arg-input-text]} [:chats/current-chat]
-            network-status                          [:network-status]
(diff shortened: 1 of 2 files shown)


rasom (approved):

mandrigin (approved):

dmitryn (approved):

lukaszfryc:
@cammellos it looks fine on mobile but I could not see "Connecting to peers.." status while I tested it. I saw "Offline" in each case. Is it expected?

Also, I still can sent message on desktop while being offline. E.g. I go offline and immediately send a message. It's marked as sent but not delivered. Should this PR fix it or it's a different case (https://github.com/status-im/status-react/issues/6396)?

cammellos:
> @cammellos it looks fine on mobile but I could not see "Connecting to peers.." status while I tested it. I saw "Offline" in each case. Is it expected?

It's not easy to get a `connecting to peers...` banner, you need to be online but with 0 peers, not sure how to replicate (this PR does not change any of the banners).

> Also, I still can sent message on desktop while being offline. 
is a different issue, you should not be able to send messages when `Offline` is displayed, but if the app does not know it's offline then you could still be able to send messages, that need to be addressed separately

lukaszfryc:
> is a different issue, you should not be able to send messages when Offline is displayed, but if the app does not know it's offline then you could still be able to send messages, that need to be addressed separately

@cammellos yes, if Offline is displayed, I can't send messages. So, it's a saparate issue as you said.

PR #6754, opened 2018-11-16
[fix] align min chat input container height with tab menu

yenda:
#### Platforms (optional)
- macOS
- Linux
- Windows

status: ready

### DIFF
--- a/src/status_im/ui/screens/desktop/main/chat/styles.cljs
+++ b/src/status_im/ui/screens/desktop/main/chat/styles.cljs
@@ -1,7 +1,10 @@
 (ns status-im.ui.screens.desktop.main.chat.styles
   (:require [status-im.ui.components.colors :as colors]))
 
-(def min-input-container-height 68)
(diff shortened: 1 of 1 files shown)


vitvly (approved):

rcullito (approved):
🚢

rasom (approved):
thank you, Eric!

rasom:
general question: can these values somehow be dependent on each other so that we don't have such problem next time (i mean input container height & tab menu height). Or this doesn't make sense?

yenda:
@rasom the odd thing is the tabs-height is 68 as well. I'll consider this a quickfix and keep in mind that this should be revisited at some point

rasom:
let's add some comment then?

flexsurfer (approved):

SlizkovaGanna:
Failed tests are unrelated/infrastructure fails

Resized and checked that the tab field and input fields do match

image

Confirmed on MacOS, Linux (emulated)

PR #6766, opened 2018-11-18
[#6736] support multi-extensions store

### LINKED ISSUE #6736: Fix store events/queries so that they support multi-extensions
Current multi-extension support broke `store/*` events and queries. They currently hardcode the [extension id](https://github.com/status-im/status-react/blob/develop/src/status_im/extensions/core.cljs#L40) to `:collectible`. 

It should be changed to the real extension id.

flexsurfer:
fixes #6736

Requires https://github.com/status-im/pluto/pull/93

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -11,7 +11,7 @@
          com.taoensso/timbre         {:mvn/version "4.10.0"}
          hickory                     {:mvn/version "0.7.1"}
          com.cognitect/transit-cljs  {:mvn/version "0.8.248"}
-         status-im/pluto             {:mvn/version "iteration-4-1"}
+         status-im/pluto             {:mvn/version "iteration-4-2"}
          mvxcvi/alphabase            {:mvn/version "1.0.0"}
          rasom/cljs-react-navigation {:mvn/version "0.1.4"}}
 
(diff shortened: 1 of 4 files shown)


deps.edn
```diff
@@ -11,7 +11,7 @@
          com.taoensso/timbre         {:mvn/version "4.10.0"}
          hickory                     {:mvn/version "0.7.1"}
          com.cognitect/transit-cljs  {:mvn/version "0.8.248"}
-         status-im/pluto             {:mvn/version "iteration-4-1"}
+         status-im/pluto             {:mvn/version "iteration-4-2-SNAPSHOT"}
```
jeluard: Can you make it `iteration-4-2`? I will publish this version.

jeluard (approved):

deps.edn
flexsurfer: ok, could you elaborate please how versioning works?

jeluard: We don't use SNAPSHOT in status-react develop. SNAPSHOT is just for local tests.
Then it's just 2 step of iteration-4. Will be cleaner once we do proper planning.

flexsurfer: thanks, done

goranjovic (approved):

PR #6770, opened 2018-11-19
introduce priority map

yenda:
- in future PRs we want to reduce the expensive sort operations
on list of messages while keeping the possibility to get a message
by its ID
- priority map allow to keep a map sorted by it's value which
is what we want to do here. we want to keep the messages ordered
by clock-value

status: ready


vitvly (approved):
Cool. Definitely needed.

rasom (approved):

PR #6792, opened 2018-11-20
[slow sign in] faster unviewed messages

rasom:
fixes
> B: all messages-statuses for unviewed messages are fetched/iterated. This data is used for “unviewed messages” counters.

> Denormalize unviewed message counter, store it along with chat. B

from https://notes.status.im/VxmgW_rJQkqHsPoqGiMGxQ

# What's the difference?
- How it worked.
  We fetched ALL `user-statuses` with `status=received` (which mean that message hasn't been seen), iterated them, grouped by chat and then stored `message-ids` of these `user-statuses` in chat's `:unviewed-messages` key.
  
 - How it works now.
 `:unviewed-messages-count` (probably should be `counter`) was added to chat entity. That means that we don't need to iterate over `user-statuses` in order to get this data. In the rest of it the difference is only that we need to update chat's db record each time when uviewed messages are seen.

status: ready

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -67,12 +67,14 @@
   "Adds new public group chat to db & realm"
   [cofx topic]
   (upsert-chat cofx
-               {:chat-id          topic
-                :is-active        true
-                :name             topic
-                :group-chat       true
-                :contacts         #{}
-                :public?          true}))
+               {:chat-id                      topic
+                :is-active                    true
+                :name                         topic
+                :group-chat                   true
+                :contacts                     #{}
+                :public?                      true
+                :unviewed-messages-count      0
+                :loaded-unviewed-messages-ids #{}}))
 
 (fx/defn add-group-chat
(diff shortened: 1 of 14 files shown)


vitvly (approved):

vitvly:
Great PR, also tried locally, haven't noticed any issues.

rasom:
Signing in on slow Android device (Samsung A500H Galaxy A5) with 7000 messages in db (all unread):


| App Version | Signig in duration (~seconds)| 
| -------- | -------- | 
| release/0.9.31| 34.25|
| nightly (22.11.2018)| 22.99|
| this PR| 15.35|
| account without messages| 14.06|

flexsurfer (approved):

rasom:
For anyone who will test this PR, there are four pub chats with 1k, 2k, 4k, and 8k messages: #tchat1000, #tchat2000, #tchat4000, #tchat8000.

asemiankevich:
looks good to me @rasom !

PR #6813, opened 2018-11-20
[slow sign in] faster get-unviewed-messages

rasom:
```clojure
;; db with 7000 unread messages, galaxy s9

;; old version
(time (let [_ (get-unviewed-messages account-id)
        :done))
;; "Elapsed time: 697.000000 msecs"

;;new version
(time (let [_ (get-unviewed-messages2 account-id)]
        :done))
;; "Elapsed time: 299.000000 msecs"

(=
  (get-unviewed-messages account-id)
  (get-unviewed-messages2 account-id))
;; true
```
status: ready

### DIFF
--- a/src/status_im/data_store/messages.cljs
+++ b/src/status_im/data_store/messages.cljs
@@ -82,16 +82,16 @@
 
 (defn- get-unviewed-messages
   [public-key]
-  (into {}
-        (map (fn [[chat-id user-statuses]]
-               [chat-id (into #{} (map :message-id) user-statuses)]))
-        (group-by :chat-id
-                  (-> @core/account-realm
-                      (core/get-by-fields
-                       :user-status
-                       :and {:public-key public-key
-                             :status     "received"})
-                      (core/all-clj :user-status)))))
+  (-> @core/account-realm
+      (core/get-by-fields
+       :user-status
+       :and {:public-key public-key
+             :status     "received"})
+      (.reduce (fn [acc msg _ _]
+                 (let [chat-id (aget msg "chat-id")
+                       message-id (aget msg "message-id")]
+                   (update acc chat-id (fnil conj #{}) message-id)))
+               {})))
 
 (re-frame/reg-cofx
  :data-store/get-unviewed-messages


src/status_im/data_store/messages.cljs
```diff
@@ -82,16 +82,18 @@
...
-                      (core/all-clj :user-status)))))
+  (-> @core/account-realm
+      (core/get-by-fields
+       :user-status
+       :and {:public-key public-key
+             :status     "received"})
+      (.reduce (fn [acc msg _ _]
+                 (let [chat-id (aget msg "chat-id")
+                       message-id (aget msg "message-id")]
+                   (if (contains? acc chat-id)
```
bhauman: Perhaps use `(update acc chat-id (fnil conj #{}) message-id)` and get rid of the explicit conditional.

rasom: nice suggestion, on it

bhauman: and maybe do `(when (and chat-id message-id) ...)` just in case ;)

rasom: well, we didn't have it and, idk, from code where these statuses are created, it seems that both are always not nil. Or you have encountered such statuses? 
Otherwise, we can put these nil checks literally everywhere. :)

rasom: `fnil` thing has been applied, thanks

bhauman: good point :)

bhauman (approved):

bhauman:
Actually that nil check really isn't a bod idea because its fetching from persistent data, and that's at the source.  Chances of bad data / corruption getting in there at some point seems fairly high given enough time.

rasom:
@bhauman i completely forgot about this, but this is checked on db level:
this is our schema for statuses https://github.com/status-im/status-react/blob/develop/src/status_im/data_store/realm/schemas/account/user_status.cljs#L14 and it doesn't allow nil values for `message-id` and `chat-id` fields, otherwise, it would be optional ([example](https://github.com/status-im/status-react/blob/develop/src/status_im/data_store/realm/schemas/account/message.cljs#L22))

flexsurfer (approved):

bhauman:
@rasom oh great, that is good to know

PR #6833, opened 2018-11-21
Several small improvements to the build scripts

pedropombeiro:

This PR serves to capture some commits that I needed to do while working on another longer-lived branch.

### Summary:

This PR does the following:
- Installs the Node/npm versions which are actually recommended in docs.status.im (I don't see the logic in installing Node 9/npm 6.x when our documentation asks for a lower version);
- Improves local build performance by allowing gradle to use a daemon and only disabling that where it supposedly causes issues (Jenkins);
- Avoids repeated downloads of Android NDK by avoiding resetting of `android/local.properties` file;
- Takes proactive measure of warning user if NDK is not configured at the beginning of the build instead of letting the user deal with cryptic message at the end of build;
- Updates `scripts/update-status-go.sh` so that the right files are updated, and any `v` prefix is ignored. 
- Adds the convenience of offering to terminate a process that is keeping port 8081 open and preventing Metro from starting (`make react-native`)

status: ready

### DIFF
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -216,19 +216,25 @@ function install_node_via_nvm() {
   cd "$(repo_path)"
 
   if [ ! -e "$nvmrc" ]; then
-    cecho "@b@blue[[+ Installing Node 9 (Node 10 is not supported by Realm)]]"
+    cecho "@b@blue[[+ Installing Node 8 (Node 10 is not supported by Realm and Node 9 doesn't support npm 5.5.1)]]"
 
-    nvm install 9
-    nvm alias status-im 9
+    nvm install 8.9.4
+    nvm alias status-im 8.9.4
     echo status-im > "$nvmrc"
 
     nvm use status-im
+    npm install -g npm@5.5.1 # Explicitly downgrade to v5.5.1 of npm, since Status Desktop builds require it (npm install hangs with version higher than 5.5.1). We could maintain two versions of npm (5.5.1 for Desktop and a more recent on for mobile) but that can be confusing and lead to inconsistent environments
   else
     local version_alias=$(cat "$nvmrc")
     nvm use $version_alias
 
     local version=$(node -v)
     cecho "+ Node already installed ($version_alias $version via NVM)... skipping."
+  fi
+
+  local npm_version=$(npm -v)
+  if [[ $npm_version != "5.5.1" ]]; then
+    cecho "@b@red[[+ npm version $npm_version is installed. npm version 5.5.1 is recommended.]]"
   fi
 }
 
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -354,7 +354,7 @@ function install_android_ndk() {
       cecho "@cyan[[Extracting Android NDK to $_ndkParentDir.]]" && \
       unzip -q -o ./android-ndk.zip -d "$_ndkParentDir" && \
       rm -f ./android-ndk.zip && \
-      _ndkTargetDir="$_ndkParentDir/$(ls $_ndkParentDir | head -n 1)" && \
+      _ndkTargetDir="$_ndkParentDir/$(ls $_ndkParentDir | grep ndk)" && \
       echo "ndk.dir=$_ndkTargetDir" | tee -a $_localPropertiesPath && \
       cecho "@blue[[Android NDK installation completed in $_ndkTargetDir.]]"
   fi
--- a/scripts/prepare-for-platform.sh
+++ b/scripts/prepare-for-platform.sh
@@ -9,19 +9,25 @@ PLATFORM_FOLDER=""
 
 #if no arguments passed, inform user about possible ones
 
-if [ $# -eq 0 ]
-  then
-    echo -e "${GREEN}This script should be invoked with platform argument: 'mobile' or 'desktop'${NC}"
-    echo "When called it links"
-    # echo "If invoked with 'mobile' argument it will make a copying: "
-    # echo "package.json.mobile -> package.json"
-    # echo "etc.."
-    exit 1
-  else
-    PLATFORM=$1
-    PLATFORM_FOLDER="${PLATFORM}_files"
+if [ $# -eq 0 ]; then
+  echo -e "${GREEN}This script should be invoked with platform argument: 'android', 'ios' or 'desktop'${NC}"
+  echo "If invoked with 'mobile' argument it will link: "
+  echo "package.json.mobile -> package.json"
+  echo "etc.."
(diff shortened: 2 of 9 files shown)


scripts/prepare-for-platform.sh
```diff
@@ -22,6 +22,13 @@ if [ $# -eq 0 ]
     PLATFORM_FOLDER="${PLATFORM}_files"
 fi
 
+if [[ $PLATFORM == 'mobile' ]]; then
+  _localPropertiesPath=./android/local.properties
```
jakubgs: These are set in Jenkins via env variables:
https://github.com/status-im/status-react/blob/develop/ci/Jenkinsfile.android#L25
```
    ANDROID_HOME = '/usr/lib/android-sdk'
    ANDROID_SDK_ROOT = '/usr/lib/android-sdk'
    ANDROID_NDK = '/usr/lib/android-ndk'
    ANDROID_NDK_HOME = '/usr/lib/android-ndk'
```
That should work instead of the `local.properties` file right?

jakubgs (approved):

scripts/prepare-for-platform.sh
pedropombeiro: Yes, I assume that's what's working on the CI server, since there we don't create `local.properties`

pedropombeiro:
Ping @mandrigin

scripts/lib/setup/installers.sh
```diff
@@ -216,20 +216,26 @@ function install_node_via_nvm() {
   cd "$(repo_path)"
 
   if [ ! -e "$nvmrc" ]; then
-    cecho "@b@blue[[+ Installing Node 9 (Node 10 is not supported by Realm)]]"
+    cecho "@b@blue[[+ Installing Node 8 (Node 10 is not supported by Realm and Node 9 doesn't support npm 5.5.1)]]"
 
-    nvm install 9
-    nvm alias status-im 9
+    nvm install 8.9.4
```
mandrigin, line 221: I use “nvm use 9.3.0”

scripts/lib/setup/installers.sh
```diff
@@ -216,20 +216,26 @@ function install_node_via_nvm() {
...
+    nvm alias status-im 8.9.4
     echo status-im > "$nvmrc"
 
     nvm use status-im
+    npm install -g npm@5.5.1
```
mandrigin: nvm should install npm too, the correct revision for the specific node version. Why override?

mandrigin (changes requested):
A few node-related questions.

Otherwise, a great job!

scripts/lib/setup/installers.sh
pedropombeiro: Thanks @mandrigin. The question is whether we should strive to have an environment which works both for building desktop and mobile. If so, desktop is currently [asking](https://status.im/build_status/desktop.html) to install Node 8/npm 5.5.1. @siphiuel does desktop really require Node 8 and npm 5.5.1, or can it work with Node 9?

scripts/lib/setup/installers.sh
pedropombeiro: I remember there used to be problems with 5.6.x which node installed, and the documentation asked to downgrade to 5.5.1. Again, this is based on instructions from `Build Desktop` page.

scripts/lib/setup/installers.sh
mandrigin: I would say we should have different code paths then for desktop and mobile and a task to try to migrate Desktop to a more modern node version :)

scripts/lib/setup/installers.sh
pedropombeiro: In the end I don't mind removing this from the script, if we can update/remove the requirements from the documentation. I just don't like the duality of having the documention requesting something and then our scripts installing another.

scripts/lib/setup/installers.sh
mandrigin: also, you can use an `.nvmrc` file to simplify that logic

vitvly: @PombeirP there are issues with Node versions higher than 8.9.4/npm versions higher than 5.5.1 due to hanging `npm install`. Node 10 is not supported as Realm did not support Node 10 until very recently (https://github.com/realm/realm-js/releases/tag/v2.19.0). Most likely upgrading Desktop to higher Node versions will not be necessary as @MaxRis is working on substituting Node.js with Qt's built-in WebEngine.

scripts/lib/setup/installers.sh
mandrigin: at least, it is worth a comment in the script :)

jakubgs: I agree that strictly sticking to `5.5.1` is the way to avoid madness. I've wasted far too much time debugging issues just because the npm version was off a bit.
We really need yarn...

mandrigin: fair point.
let's add a comment there and be done with that.

scripts/lib/setup/installers.sh
mandrigin: okay, so let's document it in the code as a comment so it won't confuse someone else :)

scripts/lib/setup/installers.sh
pedropombeiro: Good point about the comment. Added it, and also added an additional commit to the PR to help diagnose when Metro can't start due to port 8081 being taken. Something like this will be shown in that case:

image

pedropombeiro:
Changed the `make clean` target to be more aggressive. Spent a lot of time trying to figure out why the build was hanging when doing `make release-android` until I figured out it was due to hidden cache folders inside the repo. Hopefully this type of situation will be less common in the future.

pedropombeiro:
@mandrigin is there anything missing in this PR for you to be able to approve it?

scripts/lib/setup/installers.sh
mandrigin: @PombeirP (replying to your comment here: https://github.com/status-im/status-react/pull/6833#issuecomment-441596885) one thing is to add a comment with the explanation on why we are installing that version of node. And, possibly, a warning message to `startdev-<blah>` commands if the version isn't matching.

pedropombeiro: @mandrigin I'd already added a comment to https://github.com/status-im/status-react/pull/6833/files#diff-0bf2011909118cdaef211364576e068dR226, and will add a check to those commands to ensure we have the right npm version, that's a good thing to do to ensure sanity of the build.

pedropombeiro: Done.

mandrigin (approved):

PR #6838, opened 2018-11-22
[#6642] Feature/chat command extra key

### LINKED ISSUE #6642: Allow adding extra key/value pair to chat command
As a developer, I want to add extra key/value pair data to the chat command message payload accessible to all recipients.

### Description

Chat command parameters are sent to all message recipients. Sometimes you need to add extra parameters. Those should be accessible to local and recipients messages, using the same syntax.
A new event hook property could be responsible for generating those.

This could leverage `command/yield-control`

*Type*: Feature

flexsurfer:
fixes #6642

Introduced two new events

```
[chat.command/set-parameter-with-custom-params {:value "Yohoho" :params {:param1 "value1"}}]

[chat.command/set-custom-parameter {:key :param1 :value "value1"}]

```

usage

```
views/suggetsion
 [view {:style {:flex-direction :row
                :align-items    :flex-start}}
  [button {:on-click [chat.command/set-parameter-with-custom-params {:value "Yohoho" :params {:param1 "Test"}}]}]]
 
```

### DIFF
--- a/src/status_im/chat/commands/impl/transactions.cljs
+++ b/src/status_im/chat/commands/impl/transactions.cljs
@@ -91,9 +91,6 @@
   [{:id          :asset
     :type        :text
     :placeholder (i18n/label :t/send-request-currency)
-    ;; Suggestion components should be structured in such way that they will just take
-    ;; one argument, event-creator fn used to construct event to fire whenever something
-    ;; is selected.
     :suggestions choose-asset-suggestion}
    {:id          :amount
     :type        :number
--- a/src/status_im/chat/commands/sending.cljs
+++ b/src/status_im/chat/commands/sending.cljs
@@ -21,12 +21,12 @@
 
 (fx/defn validate-and-send
   "Validates and sends command in current chat"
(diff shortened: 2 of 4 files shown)


ghost:
### Pull Request Checklist

jeluard (approved):

asemiankevich:
@flexsurfer @jeluard would e2e run be sufficient for sanity checks here?

flexsurfer:
@asemiankevich i think so, thanks

goranjovic:
How does this handle communication between different versions?
What happens when old version receives a messages with the new extra key?

jeluard:
@goranjovic  Versions of what? Extensions? As far as messaging is concerned this is just a serialized payload.

goranjovic:
Versions of messages/Status. My question is what happens when Status before this PR receives a messages from Status after this PR

jeluard:
@goranjovic It's just a map as far as messaging is concerned. So this is not related to message versioning.

jeluard:
@goranjovic This PR does not add any feature in this regards, this is handled by the chat command protocol.

goranjovic (approved):

PR #6845, opened 2018-11-22
[TESTS] redirect to public chat upon tap on the #tag in message

hetvart:
### Steps to test:

1. Open any public chat with a tag message navigating a user to another chat `#another-chat-name` 
2. Follow the tag. **Expected result:** user is being redirected to the chat
3. Back to home screen. **Expected result:** user can see the chat added to the chat list

status: ready

### DIFF
--- a/test/appium/tests/atomic/chats/test_public.py
+++ b/test/appium/tests/atomic/chats/test_public.py
@@ -163,3 +163,21 @@ class TestPublicChatSingleDevice(SingleDeviceTestCase):
 
         self.verify_no_errors()
 
+    @marks.testrail_id(5675)
+    @marks.high
+    def test_redirect_to_public_chat_tapping_tag_message(self):
+        signin = SignInView(self.driver)
+        home_view = signin.create_user()
+        chat = home_view.join_public_chat('montagne-angerufen')
+        tag_message = '#spectentur'
+        chat.send_message(tag_message)
+        chat.element_starts_with_text(tag_message).click()
+        chat.empty_public_chat_message.wait_for_visibility_of_element()
+
+        if not chat.user_name_text.text == tag_message:
+            self.driver.fail('Could not redirect a user to a public chat tapping the tag message.')
+
+        home = chat.get_back_to_home_view()
+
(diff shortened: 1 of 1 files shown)


test/appium/views/chat_view.py
```diff
@@ -266,6 +267,7 @@ class ChatView(BaseView):
     def __init__(self, driver):
         super(ChatView, self).__init__(driver)
 
+        self.chat_name = ChatNameText(self.driver)
```
yevh-berdnyk: 'UserNameText' in the chat_view.py has the same accessibility id, maybe we can use it or combine these two elements into one?

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -155,3 +155,20 @@ def test_user_can_interact_with_public_chat(self):
 
         self.verify_no_errors()
 
+    @marks.testrail_id(5675)
+    @marks.high
+    def test_redirect_to_public_chat_tapping_tag_message(self):
+        signin = SignInView(self.driver)
+        home_view = signin.create_user()
+        chat = home_view.join_public_chat('montagne-angerufen')
+        tag_message = '#spectentur'
```
yevh-berdnyk, line 172: Chat history is fetched only for 24 hours, the test will not get the message each time. Could you add a step with sending the message each time?

yevh-berdnyk (changes requested):

test/appium/views/chat_view.py
hetvart: Well, since these elements exist in two different views I propose to leave both as-is.
The only thing I would change in the line is set `UserNameText` instead of `ChatNameText `

test/appium/tests/atomic/chats/test_public.py
hetvart: > Chat history is fetched only for 24 hours, the test will not get the message each time. Could you add a step with sending the message each time?

please check it out

Serhy (approved):

yevh-berdnyk (approved):

PR #6864, opened 2018-11-23
erase only account's db when it can't be encrypted

rasom:
https://discuss.status.im/t/problem-with-encryption-key/809

Currently, if account's db can't be encrypted the whole `realm` dir is removed and keychain is purged. This causes situation when other perfectly working accounts are purged without reason (we state that only current account will be removed in confirmation window), along with broken UI state when a user can't log into any account after such error.

That's what we show user in such situation:
> To protect yourself, you need to create new account and erase your old data by tapping “Apply”. If you have an existing account and would like to save your recovery phrase then choose “Cancel”, back it up, and restart the app. We strongly recommend creating new account because the old one is stored unencrypted.

With this change, only account's db will be removed in case if it can't be encrypted. 
In case of migration failure different message is shown to the user.

status: ready

### DIFF
--- a/src/status_im/data_store/realm/core.cljs
+++ b/src/status_im/data_store/realm/core.cljs
@@ -30,6 +30,17 @@
     (.schemaVersion rn-dependencies/realm file-name (to-buffer encryption-key))
     (.schemaVersion rn-dependencies/realm file-name)))
 
+(defn encrypted-realm-version-promise
+  [file-name encryption-key]
+  (js/Promise.
+   (fn [on-success on-error]
+     (try
+       (encrypted-realm-version file-name encryption-key)
+       (on-success)
+       (catch :default e
+         (on-error {:message (str e)
+                    :error   :decryption-failed}))))))
+
 (defn open-realm
   [options file-name encryption-key]
   (log/debug "Opening realm at " file-name "...")
@@ -73,6 +84,15 @@
   (log/warn "realm: deleting all realms")
   (fs/unlink realm-dir))
 
+(defn delete-account-realm
+  [address]
+  (log/warn "realm: deleting account db " (utils.ethereum/sha3 address))
+  (let [file (str accounts-realm-dir (utils.ethereum/sha3 address))]
+    (.. (fs/unlink file)
+        (then #(fs/unlink (str file ".lock")))
+        (then #(fs/unlink (str file ".management")))
+        (then #(fs/unlink (str file ".note"))))))
+
 (defn ensure-directories []
   (..
    (fs/mkdir realm-dir)
@@ -104,11 +124,12 @@
 (defn- migrate-schemas
   "Apply migrations in sequence and open database with the last schema"
   [file-name schemas encryption-key current-version]
-  (log/info "migrate schemas")
-  (doseq [schema schemas
-          :when (> (:schemaVersion schema) current-version)
-          :let [migrated-realm (open-realm schema file-name encryption-key)]]
-    (close migrated-realm))
+  (log/info "migrate schemas" current-version)
+  (when (pos? current-version)
+    (doseq [schema schemas
+            :when (> (:schemaVersion schema) current-version)
+            :let [migrated-realm (open-realm schema file-name encryption-key)]]
+      (close migrated-realm)))
   (open-realm (last schemas) file-name encryption-key))
 
 (defn keccak512-array [key]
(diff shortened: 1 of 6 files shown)


src/status_im/data_store/realm/core.cljs
```diff
@@ -104,11 +124,12 @@
 (defn- migrate-schemas
   "Apply migrations in sequence and open database with the last schema"
   [file-name schemas encryption-key current-version]
-  (log/info "migrate schemas")
-  (doseq [schema schemas
-          :when (> (:schemaVersion schema) current-version)
-          :let [migrated-realm (open-realm schema file-name encryption-key)]]
-    (close migrated-realm))
+  (log/info "migrate schemas" current-version)
+  (when (pos? current-version)
```
cammellos, line 128: Is it ok not to migrate on newly created accounts? I take yes, but I would double check

cammellos (approved):

yenda (approved):

src/status_im/data_store/realm/core.cljs
rasom: Yes, i tested and it works fine. No need to iterate migrations in order to create a new database.

churik:
@rasom can I emulate `Problem with encryption key` in this PR build somehow or I should just check for possible regression?

rasom:
@churik if you have rooted android device you can you can add some random chars to the end of db file in order to corrupt it. If you want to reproduce migrations failure i can add commit with broken migration

rcullito (approved):

rasom:
@churik added commit with broken migration:
1) when creating a new account it should work without any problems
2) if you upgrade the app from the previous version you should see confirmation dialog related to migrations failure

churik:
@rasom 
Tests:
1) installed nightly, created account (joined public chat), then upgraded to build without broken migration (https://github.com/status-im/status-react/pull/6864#issuecomment-441639830) -> existing account is working, can create new account

2) **Messages for previous chats are not fetched after erasing db**
**Steps:**
  - installed nightly
 -  created account (joined public chat #status)
 -  upgraded to build with broken migration (https://github.com/status-im/status-react/pull/6864#issuecomment-442220906) 
 - got a migration error at attempt to login to existing account
- tapped "Apply"
As a result existed chat was erased, but if I join #status again history for 24h isn't fetched again, so I can see only new messages.
Video: http://take.ms/cqj9Y
Geth.log:[geth.log](https://github.com/status-im/status-react/files/2624055/geth.log)


3)  installed nightly, created 1 account (joined public chat) and edited db-file, then created 2 account, upgraded to build without broken migration (https://github.com/status-im/status-react/pull/6864#issuecomment-441639830)  -> at attempt to login to account with corrupted db got an error, after tapping apply can login, existed chat was erased; can login to 2 existing account without errors and create new account.

4)  **Unable to login after upgrade with "We were not able to decrypt your data"**
**Steps:**
 - installed nightly
 -  created 1 account (joined public chat) 
 - edited manually db-file for 1 account to corrupt db
 - created 2 account 
 -  tried to login to 1 account 
 -  tap on "Apply"
 -  upgraded to build without broken migration (https://github.com/status-im/status-react/pull/6864#issuecomment-441639830)  
As a result cannot login to any account because of error "We were not able to decrypt your data"
monosnap 2018-11-28 10-46-06
Geth Log: [geth.log](https://github.com/status-im/status-react/files/2624008/geth.log)
Log: [Status.log](https://github.com/status-im/status-react/files/2623998/Status.log)


Can you please clarify what is expected results in case 2 and 4?
I would expect that in case 2 history for 24 h will be fetched, in case 4 I can login to 2 account with non-corrupted db, so consider them as issues.

rasom:
2. yes, that's expected. We have some cache on `status-go` side and it has to be reset. Most likely it is worth having a separate issue for this. If you test similar flow in `develop` atm (corrupt db, then erase it and restore account) it most likely will end up with the same problem. 

4. well yes, if you `tap on "Apply"` in nightly atm you remove whole realm db dir, so this PR will not restore it. That's a dead end.

churik:
@rasom got it, thanks!

**Please, be sure that you removed commit with broken migration before merge.**

Just in case besides mentioned cases in https://github.com/status-im/status-react/pull/6864#issuecomment-442405016, I've checked upgrade on IOS to build without broken migration - no issues found.

PR #6865, opened 2018-11-23
Connection status info

### LINKED ISSUE #6567: Show connection status within app
### Description
It would be nice to have an indicator somewhere in the app that would show which mailserver the app is connected to, and whether the connection is active. This should work in conjunction with the current network status notifications.

In case when there is a connection failure, click on the indicator should display a window with a more detailed error text, that can be useful during error report submission.

rcullito:
fixes #6567 

## Summary

This is a basic port of the criteria laid out by @adambabik during our time in Prague and detailed in the above-linked issue "Show connection status within the app". 

My first question for the clj folks is whether the various states of `:node/status`, which is already in app-db, can be a reasonable proxy for `discovery.started` and `discovery.stopped`. 

If not, I'd be curious about current examples within re-frame around how we can use json-rpc to get these values? I'm still ramping on the inner workings of *chat* but the principle form of message communication seems to be waiting for gethEvents via the signal-received event handler, and I wasn't sure how to go about querying values from the status-go side without merely waiting for these events. 

One other question had to do with whether the whisper min and max here should both be set to 2, and what that means currently? https://github.com/status-im/status-react/blob/develop/src/status_im/node/core.cljs#L50 While this PR is pretty simple, we talked about a visual indicator something akin to a traffic light, and I'm thinking the whisper min and max will come into play here for "yellow", though of course it would be good to discuss this with those who have been around the part of the code base for a while already.

I chose the "advanced settings" tab for this PR because we are already displaying mailserver information there. It seems like a more visual indicator could be planned as a follow up to this in collaboration with @EugeOrtiz for design input. (as has been mentioned on the issue)

## Testing instructions
1. navigate to *advanced settings* of desktop app
2. look for section near the top of the chat screen with connection info. (screenshot attached on #6567)
3. disconnect app from internet to verify "disconnected" state and then watch as peers are added once the network is re-enabled.

### DIFF
--- a/src/status_im/ui/screens/desktop/main/tabs/profile/styles.cljs
+++ b/src/status_im/ui/screens/desktop/main/tabs/profile/styles.cljs
@@ -163,6 +163,10 @@
   {:margin        24
    :font-size     20})
 
+(def connection-message-text
+  {:margin-left   24
+   :margin-bottom 10})
+
 (def title-separator
   {:height            1
    :background-color  colors/gray-light})
--- a/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs
+++ b/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs
@@ -84,7 +84,6 @@
    [pairing.views/sync-devices]
    [pairing.views/installations-list installations]])
 
-
 (defn connection-status
   "generates a composite message of the current connection state given peer and mailserver statuses"
   [peers-count node-status mailserver-state peers-disconnected?]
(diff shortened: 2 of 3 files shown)


chadyj:
Thanks Rob! Can you include some screenshots in the issue?

rcullito:
@chadyj hey chad, sure thing! just added the screenshot of the text on the issue https://github.com/status-im/status-react/issues/6567. I'll follow up with design about assets for the "traffic signal" component.

vitvly (approved):
Thanks @rcullito ! Tested this locally, and turned WiFi on/off. It went from `Connected with 5 peers, including mailserver` to `Disconnected` and then `Connected with 2,4,5 peers`. Looks good to me :)

rcullito:
Thanks for the thorough local testing @siphiuel ! 🙏

rasom (approved):

churik:
@rcullito awesome work!
Doesn't affect mobile.
Ready to merge, tested on Mac OSx, Linux Ubuntu 18.04, Windows 10 Home.

PR #6869, opened 2018-11-24
[TESTS] Block PR

yevh-berdnyk:
status: wip


hetvart (approved):

Serhy (approved):

antdanchenko (changes requested):
we should use https://developer.github.com/v3/checks/runs/ instead of changing status of a commit

antdanchenko (approved):

PR #6879, opened 2018-11-26
add fallback to downloading from DigitalOcean spaces

jakubgs:
Sometimes developers want to use a non-released version of `status-go`. For that reason we should fallback to downloading from our DigitalOcean Space bucket where all builds of `status-go` are uploaded:

https://status-go.ams3.digitaloceanspaces.com/index.html

So if someone wants to use a specific version that was not released they will have to modify the `STATUS_GO_VERSION` file to specify the specific build. For example: `def3af-181126-095340`

<blockquote></blockquote>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 help: ##@other Show this help
 	@perl -e '$(HELP_FUN)' $(MAKEFILE_LIST)
 
+DO_SPACE_URL = https://status-go.ams3.digitaloceanspaces.com/
 GITHUB_URL = https://github.com/status-im/status-go/releases/download
 RCTSTATUS_DIR = modules/react-native-status/ios/RCTStatus
 ANDROID_LIBS_DIR = android/app/libs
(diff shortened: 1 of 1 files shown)


pedropombeiro (approved):

dshulyak (approved):

Makefile
```diff
@@ -51,15 +52,29 @@ _prepare-mobile: ##@prepare Install mobile platform dependencies and prepare wor
...
 
 $(STATUS_GO_IOS_ARCH):
-	curl -s -L \
+	curl --fail --silent --location \
 		"$(GITHUB_URL)/$(STATUS_GO_VER)/status-go-ios.zip" \
-		-o "$(STATUS_GO_IOS_ARCH)"
+		--output "$(STATUS_GO_IOS_ARCH)"; \
+	if [ $$? -ne 0 ]; then \
+		echo "Failed to download from GitHub, checking DigitalOcean Bucket..."; \
+    	curl --fail --silent --location \
```
adambabik: In this line, `||` is used instead of `if [ $$? -ne 0 ]; then` which might be seen as inconsistent. Also, if the second `curl` fails, does the whole target fail?

jakubgs: You're right. Fixed.

Makefile
```diff
@@ -51,15 +52,33 @@ _prepare-mobile: ##@prepare Install mobile platform dependencies and prepare wor
...
 		"$(GITHUB_URL)/$(STATUS_GO_VER)/status-go-ios.zip" \
-		-o "$(STATUS_GO_IOS_ARCH)"
+		--output "$(STATUS_GO_IOS_ARCH)"; \
+	if [ $$? -ne 0 ]; then \
+		echo "Failed to download from GitHub, checking DigitalOcean Bucket..."; \
+		curl --fail --silent --location \
+			"${DO_SPACE_URL}/status-go-ios-$(STATUS_GO_VER).zip" \
+			--output "$(STATUS_GO_IOS_ARCH)"; \
+		if [ $$? -ne 0 ]; then \
+			echo "Failed to download from DigitalOcean Spaces!"; \
```
adambabik, line 60: Maybe we can add `exit 1` in order to not exit with code `0` in case of failure?

rasom (approved):

flexsurfer (approved):

Issue #6415, opened 2018-10-19
Messages sent when "Connecting to peers" is displayed will be lost


churik:
### Description

*Type*: Bug
*Summary*: "Connecting to peers" error doesn't lead to disappearing "Send" button, so you can type and send message but in fact it will not be sent. The same for desktop - you can send message, it will be shown in "Sent" state, but in fact will be never received.

#### Expected behavior
**Mobile**
button is hidden on mobile. If for some reason user sent a message before the send button was hidden, there should be a spinner when the message is trying to be sent. It it's not sent after a timeout, the "Not sent. Tap for options" status with resend option. Refer to https://github.com/status-im/status-react/pull/3964 if you need details of how resend should work.

<img width="200" alt="offl" src="https://user-images.githubusercontent.com/7532782/47304501-2190b380-d627-11e8-9bec-295ad5311984.png"> 

**Desktop**
"Not sent" status on desktop (like for Offline mode).

#### Actual behavior
message status is Sent, but in fact messages are lost
<img width="437" alt="genymotion for personal use - samsung galaxy s7 - 7 1 0 - api 25 - 1440x2560 1440x2560 640dpi - 192 168 58 101 2018-10-19 16-10-30" src="https://user-images.githubusercontent.com/4557972/47220363-40d6d900-d3ba-11e8-8e31-b197df58c959.png">

### Reproduction

- Open Status
- Trigger "Connecting to peers" (for exapmle like in issue https://github.com/status-im/status-react/issues/6396 or https://github.com/status-im/status-react/issues/6414)
- Send message

### Additional Information

* Status version: nightly 19/10/2018
* Operating System: Android, iOS, Mac OSx

#### Logs

* StatusIm.log (desktop): [StatusIm_time.log](https://github.com/status-im/status-react/files/2495860/StatusIm_time.log)
* Geth.log  (desktop): [geth_time.log](https://github.com/status-im/status-react/files/2495854/geth_time.log)

Consequence of  https://github.com/status-im/status-react/issues/6414 or https://github.com/status-im/status-react/issues/6396, but I believe should be fixed separately and has higher priority IMO.

lukaszfryc:
I also replicated it today on release/0.9.30. 

cc @chadyj @mandrigin

lukaszfryc:
[...]

PR #6880, opened 2018-11-26
Envelope marked as sent even if it was expired

dshulyak:
At some point it definitely worked right, i noticed this problem while testing confirmations. And app clearly shows every message as sent, even if i send it into the void.

### DIFF
--- a/src/status_im/signals/core.cljs
+++ b/src/status_im/signals/core.cljs
@@ -55,7 +55,7 @@
       "node.stopped"       (status-node-stopped cofx)
       "module.initialized" (status-module-initialized cofx)
       "envelope.sent"      (transport.message/update-envelope-status cofx (:hash event) :sent)
-      "envelope.expired"   (transport.message/update-envelope-status cofx (:hash event) :sent)
+      "envelope.expired"   (transport.message/update-envelope-status cofx (:hash event) :not-sent)
       "bundles.added"      (pairing/handle-bundles-added cofx event)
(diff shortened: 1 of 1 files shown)


mandrigin (approved):

yenda (approved):

rasom (approved):

lukaszfryc:
@dshulyak do you know steps to reproduce the original issue? Is it related to going offline->online and sending messages?

dshulyak:
> @dshulyak do you know steps to reproduce the original issue? Is it related to going offline->online and sending messages?

@lukaszfryc for example, turn off wi-fi and send a message. state indicator will start spinning and after some time it will be reported as sent. with this pr it will be correctly reported as not sent.

lukaszfryc:
Thanks! Btw, I think this issue should be already fixed as part of https://github.com/status-im/status-react/issues/6415. Also check my comment https://github.com/status-im/status-react/issues/6415#issuecomment-431837807

dshulyak:
> Thanks! Btw, I think this issue should be already fixed as part of #6415. Also check my comment #6415 (comment)

unfortunately no, app will think that connection is open for 30 seconds after actual connectivity is lost.

lukaszfryc:
Works fine 👍 I tested on mobile (iOS) and desktop (macOS). On mobile, I got an option to resend. On desktop, "Not sent" label as resend is not supported there yet.

dshulyak:
Can someone please merge this change?

PR #6887, opened 2018-11-26
[#6789] Allow ENS names to be used for ethereum events

### LINKED ISSUE #6789: Allow ENS names to be used for ethereum events
### User Story

As a developer, I want to use ENS name instead of addresses in ethereum events.

### Description

*Type*: Feature

`ethereum/call` and `ethereum/send-transaction` should be able to resolve contracts from ENS name. An extra `ens` argument could be introduced.

tbenr:
Fixes #6789

### DIFF
--- a/src/status_im/extensions/ethereum.cljs
+++ b/src/status_im/extensions/ethereum.cljs
@@ -8,6 +8,8 @@
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.ethereum.abi-spec :as abi-spec]
             [status-im.utils.fx :as fx]
+            [status-im.utils.ethereum.ens :as ens]
+            [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.handlers :as handlers]
             [status-im.utils.money :as money]
             [clojure.string :as string]
@@ -26,6 +28,21 @@
  :extensions/transaction-on-error
  (fn [{db :db} [_ on-result message]]
    (when on-result {:dispatch (on-result {:error message :result nil})})))
+
+(defn- wrap-with-resolution [db arguments address-keyword f]
+  "funtction responsible to resolve ens taken from argument
+ and call the specified function with resolved address"
+  (let [address (get arguments address-keyword)
+        first-address (if (vector? address)  ;; currently we only support one ens for address
+                        (first address)
+                        address)]
+    (if (ens/is-valid-eth-name? first-address)
+      (let [{:keys [web3 network]} db
+            network-info (get-in db [:account/account :networks network])
+            chain (ethereum/network->chain-keyword network-info)
+            registry (get ens/ens-registries chain)]
+        (ens/get-addr web3 registry first-address #(f db (assoc arguments address-keyword %))))
+      (f db arguments))))
 
 ;; EXTENSION TRANSACTION -> SEND TRANSACTION
 (defn  prepare-extension-transaction [params contacts on-result]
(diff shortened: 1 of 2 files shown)


src/status_im/extensions/ethereum.cljs
```diff
@@ -116,16 +118,24 @@
...
\ No newline at end of file
+ (fn [{db :db} [_ _ {:keys [fromBlock toBlock address topics blockhash on-result] :as arguments}]]
+   (if (ens/is-valid-eth-name? address)
+     (let [{:keys [web3 network]} db
+           network-info (get-in db [:account/account :networks network])
+           chain (ethereum/network->chain-keyword network-info)
+           registry (get ens/ens-registries chain)]
+       (do
+         (print "network: " network "\nchain: " chain)
+         (ens/resolver web3 registry address #(re-frame/dispatch [:extensions/ethereum-logs nil [nil nil (assoc arguments :address %)]]))))
```
jeluard: Can a avoid an extra dispatch and just provide a reference to a function?  The block in the else clause would be this function.

tbenr: Yes, you're right. Btw the PR was just a placeholder for the gitcoin.

tbenr:
what i'd like to do is a general method to extract addresses from the argument\s, resolve all addresses found, and then call to web3 using resolved addresse(s).
So it can be applied to all eth events dealing with addresses.
I need to study how to chain multiple address resolution (asynchronous calls with cb)

jeluard:
@tbenr That's probably doable with a higher order fn (having the original fn as argument). This function could detect that `address` is passed as an argument then resolve it via ENS.

tbenr:
@jeluard yes, but the complex part for me is to resolve multiple ens and then, when all done, call the fn.
for instance, for the `ethereum-logs` I can have a vector of addresses in `:address` parameter and also multiple addresses in the topic definition. 

let's say I add a `multiresolver` in ens that accepts a set of ens `#{"test.eth"}` and returns a map `{:test.eth "0x0000"}` 

so the events are responsible to:

1. extract ens from their arguments and generate the set of ens
2. call the `multiresolver`
3. recreate arguments by substituing ens with resolved addresses using the received map

jeluard:
@tbenr Oh right.. Maybe as a first step ENS names could be resolved for the `address` field only?

tbenr:
@jeluard that will be quite straight forward :-)
ok i'll do the simplified version than I can think on a more complete version on a separate PR

tbenr:
@jeluard this is just a test commit, not tested. But, instead of implementing resolution for every event interface, what about exposing a dedicated event for ens resolution?
i see a couple of benefits be:
- simpler implementation (you don't need to care about where addresses are spreaded along the parameters)
- extension developer is "forced" to resolve  store resolved addresses in the state. (so they suppose to resolve addresses once)

what do you think?

jeluard:
@tbenr  Actually I think both are useful ! I am fine with your current approach (makes it trivial to switch to ENS for those existing calls).

Now it would be useful too to expose ENS resolution as an event.

tbenr:
@jeluard ok! are there some sample extensions to test the other ethereum events?

jeluard:
@tbenr Not on the shelf but kyber has a contract registered using ENS so the following shoudl work:

```clojure
[ethereum/call {:to        "kybernetwork.eth" 
                   :method    "getExpectedRate(address,address,uint256)"
                   :params    ["0x744d70fdbe2ba4cf95131626614a1763df805b9e" "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee" 1000000000000000] 
                   :outputs   ["uint256" "uint256"] 
                   :on-result [on-src-picker-change-result {:selection selection}]}]
```

tbenr:
@jeluard i added also a new event for the resolution (check if you like the name ;))

btw, tried to create a macro to simplify the ens resolution for every event.. something like:

```
(defmacro resolve-ens-and-call
 [ens-holder call]
 `(fn [{db :db} [_ _ {:keys [~ens-holder] :as arguments}]]
   (if (ens/is-valid-eth-name? ~ens-holder)
     (let [{:keys [web3 network]} db
           network-info (get-in db [:account/account :networks network])
           chain (ethereum/network->chain-keyword network-info)
           registry (get ens/ens-registries chain)]
       (do
         (print "\n\nnetwork: " network "\nchain: " chain)
         (ens/resolver web3 registry ens-holder #(~call (assoc arguments ~ens-holder %)))))
    (~call arguments))))

(handlers/register-handler-fx
 :extensions/ethereum-logs
 (resolve-ens-and-call :address call-get-logs))
```

but there's something basically wrong: i'm not able to manage macros yet :)

jeluard:
@tbenr Yes the ENS event name sounds right :)

Did you try to create a function to wrap the call?

Something like:

```clojure
(defn wrap-with-resolution [f ...]
   (if (ens/is-valid-eth-name? to)
                          :data (when (and method params) (abi-spec/encode method params)))	     (let [{:keys [web3 network]} db
         transaction (prepare-extension-transaction tx-object (:contacts/contacts db) on-result)]	           network-info (get-in db [:account/account :networks network])
     (models.wallet/open-modal-wallet-for-transaction db transaction tx-object))))	           chain (ethereum/network->chain-keyword network-info)
           registry (get ens/ens-registries chain)]
       (ens/get-addr web3 registry to #(execute-send-transaction db (assoc db arguments :to %))))
     (f arguments))))

;; Then call it later on
(wrap-with-resolution #(execute-send-transaction db %) ...)
```

tbenr:
I didn't realize macro was not needed to do that.
PR is finished.

jeluard (approved):

goranjovic (approved):

src/status_im/extensions/ethereum.cljs
```diff
@@ -27,6 +29,21 @@
  (fn [{db :db} [_ on-result message]]
    (when on-result {:dispatch (on-result {:error message :result nil})})))
 
+(defn- wrap-with-resolution [db arguments address-keyword f]
```
flexsurfer, line 32: what is `f` ?

flexsurfer (approved):

src/status_im/extensions/ethereum.cljs
jeluard: It's the function wrapped that will be called. Any better name?

jeluard:
@tbenr Actually do you mind change `ethereum/resolve-ens` to `ethereum.ens/resolve` ? Sorry for the last minute change.

src/status_im/extensions/ethereum.cljs
tbenr: `cb`? `callback`? `eventfunction`?

PR #6890, opened 2018-11-27
[#6867] Add on-send-sync to chat.command extension

### LINKED ISSUE #6867: Add on-send-sync to chat.command extension
As a developer I want to be able to decide if a chat.command results in a message being send or not. `on-send-sync` will provide this feature (by leveraging Command/yield-control).
A new event `chat.command/send` provides control on when the command message will be sent. A map of extra parameters can also be provided.

```clojure
events/on-http-result
(let [{} properties]
 [chat.command/send {:key :value}])

events/on-send-sync
(let [{} properties]
  [http/get {:on-result [on-http-result]}])

...
:on-send-sync [on-send-sync]
```

_Note_ This deprecates `chat.command/set-custom-parameter`

flexsurfer:
fixes #6867

Requires https://github.com/status-im/pluto/pull/98

introduced `on-send-sync` property for  chat.command extension, and two new events 
`[chat.command/send {:params {}}]` and `[chat.command/send-plain-text-message {:value "message text"}]`

usage
https://status-im.github.io/pluto/try.html?hash=QmcfsYnFvKXApcFTCNttpQvQKYxiMCqVx5MvsocFyrr2KA

```
events/on-send-sync
 (let [{title :title text :text} properties]
   [chat.command/send {:params {:text text}}}])
 

 hooks/chat.command.principles
 {:description   "Status Principles"
  :on-send-sync  [on-send-sync]
``` 

<blockquote><div><strong><a href="https://status-im.github.io/pluto/index.html">Pluto · A grammar for data manipulation</a></strong></div><div>A grammar for data manipulation</div></blockquote>

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -11,7 +11,7 @@
          com.taoensso/timbre         {:mvn/version "4.10.0"}
          hickory                     {:mvn/version "0.7.1"}
          com.cognitect/transit-cljs  {:mvn/version "0.8.248"}
-         status-im/pluto             {:mvn/version "iteration-4-3"}
+         status-im/pluto             {:mvn/version "iteration-4-4"}
          mvxcvi/alphabase            {:mvn/version "1.0.0"}
          rasom/cljs-react-navigation {:mvn/version "0.1.4"}}
 
--- a/src/status_im/chat/commands/core.cljs
+++ b/src/status_im/chat/commands/core.cljs
@@ -120,23 +120,37 @@
     :preview       :view
     :on-send?      :event
     :on-receive?   :event
+    :on-send-sync? :event
     :parameters?    [{:id           :keyword
                       :type         {:one-of #{:text :phone :password :number}}
                       :placeholder  :string
                       :suggestions? :view}]}
    :hook
    (reify hooks/Hook
-     (hook-in [_ id {:keys [description scope parameters preview short-preview on-send on-receive]} cofx]
-       (let [new-command (reify protocol/Command
-                           (id [_] (name id))
-                           (scope [_] scope)
-                           (description [_] description)
-                           (parameters [_] (or parameters []))
-                           (validate [_ _ _])
-                           (on-send [_ command-message _] (when on-send {:dispatch (on-send command-message)}))
-                           (on-receive [_ command-message _] (when on-receive {:dispatch (on-receive command-message)}))
-                           (short-preview [_ props] (short-preview props))
-                           (preview [_ props] (preview props)))]
+     (hook-in [_ id {:keys [description scope parameters preview short-preview on-send on-receive on-send-sync]} cofx]
+       (let [new-command (if on-send-sync
+                           (reify protocol/Command
+                             (id [_] (name id))
+                             (scope [_] scope)
+                             (description [_] description)
+                             (parameters [_] (or parameters []))
+                             (validate [_ _ _])
+                             (on-send [_ command-message _] (when on-send {:dispatch (on-send command-message)}))
+                             (on-receive [_ command-message _] (when on-receive {:dispatch (on-receive command-message)}))
+                             (short-preview [_ props] (short-preview props))
+                             (preview [_ props] (preview props))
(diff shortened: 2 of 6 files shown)


src/status_im/chat/commands/core.cljs
```diff
@@ -120,23 +120,37 @@
...
+                             (description [_] description)
+                             (parameters [_] (or parameters []))
+                             (validate [_ _ _])
+                             (on-send [_ command-message _] (when on-send {:dispatch (on-send command-message)}))
+                             (on-receive [_ command-message _] (when on-receive {:dispatch (on-receive command-message)}))
+                             (short-preview [_ props] (short-preview props))
+                             (preview [_ props] (preview props))
+                             protocol/Yielding
+                             (yield-control [_ props _] {:dispatch (on-send-sync props)}))
+                           (reify protocol/Command
```
flexsurfer, line 144: there should be a better way for sure! to do not duplicate code

src/status_im/extensions/core.cljs
```diff
@@ -272,6 +285,14 @@
                 {:permissions [:read]
                  :value       :extensions.chat.command/set-parameter-with-custom-params
                  :arguments   {:value :string :params :map}}
+                'chat.command/send-plain-text-message
+                {:permissions [:read]
+                 :value       :extensions.chat.command/send-plain-text-message
+                 :arguments   {:value :string}}
+                'chat.command/send
+                {:permissions [:read]
+                 :value       :extensions.chat.command/send
```
jeluard: Maybe `send-message` then?

deps.edn
```diff
@@ -11,7 +11,7 @@
          com.taoensso/timbre         {:mvn/version "4.10.0"}
          hickory                     {:mvn/version "0.7.1"}
          com.cognitect/transit-cljs  {:mvn/version "0.8.248"}
-         status-im/pluto             {:mvn/version "iteration-4-3"}
+         status-im/pluto             {:mvn/version "iteration-4-5"}
```
jeluard: Why not 4.4?

src/status_im/chat/commands/core.cljs
jeluard: Hmm no idea for now

jeluard:
@flexsurfer Do you have an example of using `send` ? Can I add extra parameters there?

src/status_im/extensions/core.cljs
flexsurfer: why message? it sends command )imo send is fine

deps.edn
flexsurfer: still confusing with the version, can change to 4.4

flexsurfer:
@jeluard example of `send` in the description, yes you can provide any parameters
```
events/on-send-sync
 (let [{title :title text :text} properties]
   [chat.command/send {:params {:text text :custom "custom"}}}])
```

src/status_im/extensions/core.cljs
jeluard: Well it really send a message no? I find it a bit confusing with `send-plain-text-message`

jeluard:
Do we need the extra `params` level? Could be just the map directly.
What happens if one of the default parameter is not provided?

flexsurfer:
@jeluard if i understand right how arguments work i need to describe all arguments and types, but params is dynamic, you can provide any params

What happens if one of the default parameter is not provided?
something bad for sure

jeluard:
> @jeluard if i understand right how arguments work i need to describe all arguments and types, but params is dynamic, you can provide any params

So this is for extra params? Do you distinguish between the default ones and extra ones?

flexsurfer:
it's for both for default and for extra, so default params are mandatory

src/status_im/extensions/core.cljs
flexsurfer: hm, yes it's , i'll change it

jeluard (approved):

rcullito (approved):

PR #6908, opened 2018-11-27
[TESTS] Do not save device stats in InfluxDB if database not provided

lukaszfryc:
Follow up on https://github.com/status-im/status-react/pull/6692. Do not send results to InfluxDB if db credentials not provided.

status: ready

### DIFF
--- a/test/appium/tests/conftest.py
+++ b/test/appium/tests/conftest.py
@@ -190,6 +190,12 @@ def pytest_unconfigure(config):
             pull.create_issue_comment(github_report.build_html_report(testrail_report.run_id))
 
 
+def should_save_device_stats(config):
+    db_args = [config.getoption(option) for option in
+               ('stats_db_host', 'stats_db_port', 'stats_db_username', 'stats_db_password', 'stats_db_database')]
+    return all(db_args)
+
+
 @pytest.mark.hookwrapper
 def pytest_runtest_makereport(item, call):
     outcome = yield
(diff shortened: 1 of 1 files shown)


yevh-berdnyk (approved):

test/appium/tests/conftest.py
```diff
@@ -190,6 +190,14 @@ def pytest_unconfigure(config):
             pull.create_issue_comment(github_report.build_html_report(testrail_report.run_id))
 
 
+def should_save_device_stats(config):
+    return config.getoption('stats_db_host') and \
+           config.getoption('stats_db_port') and \
+           config.getoption('stats_db_username') and \
+           config.getoption('stats_db_username') and \
```
hetvart: twice `stats_db_username` option

test/appium/tests/conftest.py
```diff
@@ -190,6 +190,14 @@ def pytest_unconfigure(config):
             pull.create_issue_comment(github_report.build_html_report(testrail_report.run_id))
 
 
+def should_save_device_stats(config):
+    return config.getoption('stats_db_host') and \
```
hetvart: do not overdrive the method with multi `and` verification.
you can use python `all()` built-in function for such checking:

```python
def should_save_device_stats(config):
    db_args = [config.getoption(option) for option in ('stats_db_host', 'stats_db_port', 'stats_db_username', 'stats_db_password')]
    return all(db_args)
```
`all()` returns True when all elements in the given iterable are true. If not, it returns False.

lukaszfryc: sure, we can sue `all()`. But, I find simple conjunction (`and`) much more readable in this case. There is no need to create an array (`db_args`). But, I can change it if you think we should use `all()`

hetvart: > sure, we can sue `all()`. But, I find simple conjunction (`and`) much more readable in this case. There is no need to create an array (`db_args`). But, I can change it if you think we should use `all()`

I proposed just more pythonic way to do so - gold rule is when you have more than 2 `and/or` conditions. `all()` is written in C and does this verification faster then `and/or` syntax.

Regarding readability, well it depends mostly on developer preferences/style though.

hetvart (approved):

test/appium/tests/conftest.py
lukaszfryc: thanks! I'll change it. I think I still need to learn how to code in a more pythonic way :)

PR #6915, opened 2018-11-28
Check if package-lock.json was changed and fail the build (Mobile only)

### LINKED ISSUE #6906: Fail build if `package-lock.json` changes
In an attempt to start monitoring our dependencies, we should fail the build if the `package-lock.json` file changes as it checksums all dependencies.  This gives us an initial guarantee that whatever changes that are happening are not within the scope of dependencies, and informs us that we need to check whatever changes happen within a dependency when they change.

This will suffice as an initial attempt to start monitoring these things, until we can implement a more robust methodology in the build process such as https://snyk.io

<blockquote><div>Snyk</div><div><strong><a href="https://snyk.io/">Open Source Security Platform | Snyk</a></strong></div><div>Snyk helps you use open source and stay secure. Continuously find and fix vulnerabilities for npm, Maven, NuGet, RubyGems, PyPI and much more.</div></blockquote>

mandrigin:
Due to vulnerabilities, like the one found in [event-stream](https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream), we can't trust npm even when locking the version numbers.
Hence, we are trying to mitigate similar issues by checking if `package-lock.json` was updated after `npm install` and failing the build if it was.
The lock file contains a checksum of the minified package, so if someone updates it but keeps the same version, this script should detect it.

- [x] Write the script
- [x] Update the `package-lock.json` content
- [x] Plug in the new script instead of `npm install`
- [x] Ensure the deps versions in the `package.json` are locked
- [x] (Temporarily) Disable Desktop and add a follow-up PR. (https://github.com/status-im/status-react/pull/6918)

fixes #6906

status: ready

### DIFF
--- a/ci/common.groovy
+++ b/ci/common.groovy
@@ -53,19 +53,27 @@ def installJSDeps(platform) {
   def attempt = 1
   def maxAttempts = 10
   def installed = false
-  def errroCode = 0
+  def errorCode = 0
+  def errorLog = ""
   /* prepare environment for specific platform build */
   sh "scripts/prepare-for-platform.sh ${platform}"
   while (!installed && attempt <= maxAttempts) {
     println("#${attempt} attempt to install npm deps")
 
-    errorCode = sh('scripts/locked-npm-install.sh', returnStatus: true)
+    try {
+      errorLog = sh(
+        returnStdout: true, 
+        script: 'scripts/locked-npm-install.sh'
+      )
+    } catch (exc) {
+        errorCode = 1
+    }
     installed = fileExists('node_modules/web3/index.js')
     attemp = attempt + 1
   }
 
   if(!installed || errorCode != 0) {
-    error "node dependencies installation failed (installed: ${installed}, errorCode: ${errorCode})"
+    error "node dependencies installation failed (installed: ${installed}, errorCode: ${errorCode}, errorLog: ${errorLog})"
   }
 }
 
--- a/ci/common.groovy
+++ b/ci/common.groovy
@@ -53,13 +53,19 @@ def installJSDeps(platform) {
   def attempt = 1
   def maxAttempts = 10
   def installed = false
+  def errroCode = 0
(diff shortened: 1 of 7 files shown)


ci/common.groovy
```diff
@@ -53,14 +53,20 @@ def installJSDeps(platform) {
   def attempt = 1
   def maxAttempts = 10
   def installed = false
+  def errroCode = 0
```
pedropombeiro: typo

pedropombeiro:
Output for `make prepare-desktop`: [output.log](https://github.com/status-im/status-react/files/2623913/output.log)

ci/common.groovy
mandrigin: fixed

mobile_files/package-lock.json
```diff
@@ -8791,7 +8791,9 @@
       }
     },
     "react-native-http-bridge": {
-      "version": "git+https://github.com/status-im/react-native-http-bridge.git#214301a806743d0ce04f4559c3d55b3e8ff95f5d"
```
mandrigin, line 8794: @flexsurfer @yenda so we have switched from our own fork of the `react-native-http-bridge` back to the npm version, right?

pedropombeiro (approved):

rasom (approved):

mandrigin:
Continuing my adventures. `npm` is a truly amazing piece of... technology: https://github.com/npm/npm/issues/18712

pedropombeiro:
I know we tried moving to yarn before and abandoned that for some reason, but haven't come across a write-up of what happened. It does seem to be something we'd want to do sooner rather than later, especially before Reproducible Builds.

cammellos:
@PombeirP description is in the issue https://github.com/status-im/status-react/issues/3838

mandrigin:
Okay, I will try to switch us to Yarn, let's close this PR for now.

PR #6925, opened 2018-11-28
use GIT_REPOSITORY instead of URL to get status-go code

jakubgs:
Why are we even downloading the zip of the repo if we can just clone it.
And if that's not available we can use the DigitalOcean bucket to download the zip.

### DIFF
--- a/modules/react-native-status/desktop/CMakeLists.txt
+++ b/modules/react-native-status/desktop/CMakeLists.txt
@@ -26,12 +26,22 @@ include_directories(${StatusGo_INCLUDE_DIR})
 
 set(CONFIGURE_SCRIPT build-status-go.sh)
 
-file (STRINGS "../../../STATUS_GO_VERSION" STATUS_GO_VERSION)
+file(STRINGS "../../../STATUS_GO_VERSION" STATUS_GO_VERSION)
+
+# custom status-go builds are identified by '{date}-{time}-{commit}'
+if ("${STATUS_GO_VERSION}" MATCHES "[0-9][0-9][0-9][0-9][0-9][0-9]-[0-9][0-9][0-9][0-9][0-9][0-9]-([0-9a-zA-Z]+)")
+  set(STATUS_GO_TAG ${CMAKE_MATCH_1})
+else()
+  set(STATUS_GO_TAG ${STATUS_GO_VERSION})
+endif()
 
 ExternalProject_Add(StatusGo_ep
   PREFIX ${StatusGo_PREFIX}
   SOURCE_DIR ${StatusGo_SOURCE_DIR}
-  URL https://github.com/status-im/status-go/releases/download/${STATUS_GO_VERSION}/status-go-desktop.zip
+  GIT_REPOSITORY https://github.com/status-im/status-go.git
+  GIT_TAG ${STATUS_GO_TAG}
+  URL https://status-go.ams3.digitaloceanspaces.com/status-go-desktop-${STATUS_GO_VERSION}.zip
+      https://github.com/status-im/status-go/archive/${STATUS_GO_VERSION}.zip
   BUILD_BYPRODUCTS ${StatusGo_STATIC_LIB}
   CONFIGURE_COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/${CONFIGURE_SCRIPT} ${CMAKE_SYSTEM_NAME} ${GO_ROOT_PATH} ${StatusGo_ROOT} ${StatusGo_SOURCE_DIR} ${CMAKE_C_COMPILER} ${CMAKE_CXX_COMPILER}
   BUILD_COMMAND ""


cammellos:
Not so sure about this, potentially you might run in the issue of having different code in `archive` and uploaded to `digital ocean` or `android`/`ios` (say someone pushes a commit while the release is building), not worth the risk in my opinion, otherwise we should make the build system create the release before pulling the code, as building might very well take 10/20 minutes

cammellos:
oh actually, if we do pin commit hash & branch should be safe, need to check

jakubgs:
>oh actually, if we do pin commit hash & branch should be safe, need to check

Yeah. that's the idea, either we use the normal contents of `STATUS_GO_VERSION`, which is a git tag, like `v0.16.3-9-g27700aa2` right now, OR we use the timestamp+commit that references a specific build from: https://ci.status.im/job/status-go/job/parallel/
For example: `181129-081705-e76919`, that would use `e76919` to fetch that specific commit from git.

I think that's pretty exact.

<blockquote><div><strong><a href="https://ci.status.im/job/status-go/job/parallel/">parallel [status-go] [Jenkins]</a></strong></div></blockquote>

pedropombeiro (approved):

cammellos (approved):

vkjr (approved):

modules/react-native-status/desktop/CMakeLists.txt
```diff
@@ -26,12 +26,22 @@ include_directories(${StatusGo_INCLUDE_DIR})
...
+  set(STATUS_GO_TAG ${STATUS_GO_VERSION})
+endif()
 
 ExternalProject_Add(StatusGo_ep
   PREFIX ${StatusGo_PREFIX}
   SOURCE_DIR ${StatusGo_SOURCE_DIR}
-  URL https://github.com/status-im/status-go/releases/download/${STATUS_GO_VERSION}/status-go-desktop.zip
+  GIT_REPOSITORY https://github.com/status-im/status-go.git
+  GIT_TAG ${STATUS_GO_TAG}
+  URL https://status-go.ams3.digitaloceanspaces.com/status-go-desktop-${STATUS_GO_VERSION}.zip
```
vitvly, line 43: Didn't know that multiple URLs can be given in sequence :)

vitvly (approved):

modules/react-native-status/desktop/CMakeLists.txt
```diff
@@ -26,12 +26,22 @@ include_directories(${StatusGo_INCLUDE_DIR})
 
 set(CONFIGURE_SCRIPT build-status-go.sh)
 
-file (STRINGS "../../../STATUS_GO_VERSION" STATUS_GO_VERSION)
+file(STRINGS "../../../STATUS_GO_VERSION" STATUS_GO_VERSION)
+
+# custom status-go builds are identified by '{date}-{time}-{commit}'
+if ("${STATUS_GO_VERSION}" MATCHES "[0-9][0-9][0-9][0-9][0-9][0-9]-[0-9][0-9][0-9][0-9][0-9][0-9]-([0-9a-zA-Z]+)")
```
adambabik, line 32: `[0-9]{6}-[0-9][0-9][0-9][0-9][0-9]{6}` is too difficult for whatever executes this file?

adambabik (approved):

modules/react-native-status/desktop/CMakeLists.txt
pedropombeiro: Yup, I thought the same thing, but looking at the CMake documentation it is not supported.

modules/react-native-status/desktop/CMakeLists.txt
jakubgs: Yeah, it's pretty cool:
https://cmake.org/cmake/help/latest/module/ExternalProject.html#examples

modules/react-native-status/desktop/CMakeLists.txt
jakubgs: Yeah, it's really stupid:
https://cmake.org/cmake/help/v3.0/command/string.html
```
^         Matches at beginning of input
$         Matches at end of input
.         Matches any single character
[ ]       Matches any character(s) inside the brackets
[^ ]      Matches any character(s) not inside the brackets
 -        Inside brackets, specifies an inclusive range between
          characters on either side e.g. [a-f] is [abcdef]
          To match a literal - using brackets, make it the first
          or the last character e.g. [+*/-] matches basic
          mathematical operators.
*         Matches preceding pattern zero or more times
+         Matches preceding pattern one or more times
?         Matches preceding pattern zero or once only
|         Matches a pattern on either side of the |
()        Saves a matched subexpression, which can be referenced
          in the REGEX REPLACE operation. Additionally it is saved
          by all regular expression-related commands, including
          e.g. if( MATCHES ), in the variables CMAKE_MATCH_(0..9).
```

PR #6927, opened 2018-11-28
[#6644] Chat command recipients should be able to install an extension

### LINKED ISSUE #6644: Chat command recipients should be able to install an extension
### User Story

As a user, I want the ability to easily install the extension needed to show the chat command message. The extensions URL could be sent as part of the message payload.

### Description

*Type*: Feature

When receiving a message from an unknown command, a placeholder message is displayed.

It should be easy to install the required command from the chat screen.

flexsurfer:
fixes #6644

Requires : https://github.com/status-im/pluto/pull/101

install-extension

### DIFF
--- a/src/status_im/chat/commands/protocol.cljs
+++ b/src/status_im/chat/commands/protocol.cljs
@@ -68,3 +68,7 @@
     "Function which takes original parameters + cofx map and returns new map of parameters")
   (enhance-receive-parameters [this parameters cofx]
     "Function which takes original parameters + cofx map and returns new map of parameters"))
+
+(defprotocol Extension
+  "Protocol for defining extension"
+  (extension-id [this]))
--- a/src/status_im/ui/screens/extensions/add/views.cljs
+++ b/src/status_im/ui/screens/extensions/add/views.cljs
@@ -30,13 +30,13 @@
                  values))
           hooks))
 
-(views/defview show-extension []
+(views/defview show-extension-base [modal?]
   (views/letsubs [{:keys [data errors]} [:get-staged-extension]]
     (if data
       [react/view styles/screen
        [status-bar/status-bar]
        [react/keyboard-avoiding-view components.styles/flex
-        [toolbar/simple-toolbar (i18n/label :t/extension)]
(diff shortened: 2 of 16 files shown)


src/status_im/chat/commands/protocol.cljs
```diff
@@ -68,3 +68,7 @@
     "Function which takes original parameters + cofx map and returns new map of parameters")
   (enhance-receive-parameters [this parameters cofx]
     "Function which takes original parameters + cofx map and returns new map of parameters"))
+
+(defprotocol Extension
```
jeluard, line 72: How about having some concept a bit more generic? Like `Context` that would allow to access a map, eventually containing `::extension/id` ?

src/status_im/ui/screens/extensions/add/views.cljs
```diff
@@ -80,6 +80,12 @@
          [react/text (i18n/label :t/invalid-extension)]
          [react/text (str errors)]]]])))
 
+(views/defview show-extension []
+  [show-extension-base false])
```
jeluard, line 84: Maybe use a map here? `{:modal? false}` reads a bit better

src/status_im/chat/commands/protocol.cljs
flexsurfer: i'm not sure we need this , also we need to check a type if it's extension or not

src/status_im/ui/screens/extensions/add/views.cljs
flexsurfer: it's just general approach, we use it the same in other modules

jeluard: Ok!

src/status_im/chat/commands/protocol.cljs
jeluard: Ok maybe let's keep this for now. If we need more we can introduce something more generic later.

jeluard (approved):

goranjovic (approved):

dmitryn (approved):

PR #6928, opened 2018-11-28
Status 💛 Yarn. Lock dependencies.

### LINKED ISSUE #3838: Move to yarn instead of npm

### User Story

As a developer,
I would like to use yarn instead of npm,
because it's faster (in our case) and does not get stuck installing packages

### Description

We agreed to move to yarn instead of npm.

We tried to do so but failed as we were not able to build a release in a reasonable amount of time.

Moving to yarn is a trivial code change, but after:
`lein prod-build`

and

`cd android && ./gradlew assembleRelease`

it would take a very long time to create the artifact, around 6 hrs on my local machine ( compared to 6 minutes when installing packages through npm). 

This has been replicated locally, on docker and on the macos build server.

The steps that fails is `app:bundleReleaseJsAndAssets`.

We see `jest-worker` hogging the CPU at 100%, memory consumption is also quite high, once it actually failed because the node engine could not allocate memory.

### A few scattered thoughts that might help

There are a few issues (some closed) https://github.com/facebook/react-native/issues/11696 which seems to indicate that it's bundling files that it should not be (realm files for example).

Some people mentioned `.babelrc` was the culprit, but did not elaborate any further.

### What we tried, but did not work

Increase node memory from 4GB to 8GB
Increase gradlew memory from 1GB to 4GB
Add `___tests__` to `inputExcludes` in `android/app/react.gradle`
Decrease the number of `jest-workers`

Currently we don't have the bandwidth to look into this any further, but if anyone wants to have a go at this it would be helpful.

### LINKED ISSUE #6906: Fail build if `package-lock.json` changes

mandrigin:
One more step towards reproducible builds.

yarn (as opposed of npm) promises integrity checks and using lock files to download dependencies.
That way, we can avoid security issues with dependencies, just like [this one](https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream).

Notable change: I had to rename `package-lock.json` in **desktop_files** and **mobile_files** to `package-lock.json.orig`. That is done to avoid duplicate symbol error when bundling JS.

🚨 🚨 🚨  **We use `yarn install --frozen-lockfile` command. It will fail if the lockfile will change. Use `yarn upgrade` to upgrade the lockfile**

- [x] Android
- [x] iOS
- [x] Setup scripts
- [x] Docs →→→ https://github.com/status-im/status.im/pull/172
- [x] Desktop
- [x] force immutable lock file
- [x] Jenkins
   - [x] ~add custom cache folders per runner (so 2 parallel jobs can't conflict)~ not needed as soon as the cache is downloaded
   - [x] macos-1 doesn't have `yarn` installed
   - [x] outdated `yarn` everywhere else
   - [x] desktop fails with `react-native-webview-bridge` error
   - [x] some fails with yarn cache on macos-2

--

fixes #6906
fixes #3838

---

status: ready


corpetty:
In order to keep determinism in line, you have to use `yarn upgrade <package>` when dealing with dependencies.  We should also fail builds for dependency changes so they can be properly audited.

corpetty:
also, great work @mandrigin!!

mandrigin:
> We should also fail builds for dependency changes so they can be properly audited.

Yeah, I was thinking about it. we need to block PRs when the yarn.lock is changed.

mandrigin:
Okay, it looks like 2 parallel yarn commands can conflict if running on the same machine. Custom cache path will help with that, even though it will 2x increase a disk space per machine... That is only required for Jenkins.

jakubgs:
We have plenty of space on jenkins machines, no worries.

jakubgs:
But this also means I will have to upgrade node on all linux and macos hosts for Jenkins to 9.3.0.

mandrigin:
@jakubgs 

> But this also means I will have to upgrade node on all linux and macos hosts for Jenkins to 9.3.0.

why is that? it looks like yarn works just fine with 8.x that is installed. I need to double-check that Desktop is buildable on 9.3.0.

jakubgs:
I guess if it works we can leave it as it is.

jakubgs (approved):
Thanks for doing this! Great work.

Serhy:
@mandrigin, each build https://github.com/status-im/status-react/pull/6928#issuecomment-442810484 tested.
I could install, create account and join public channel. Offline messages received.
I could upgrade from 0.9.32 Release build (iOS, Android and MacOS)
Release 0.9.32 can exchange messages with contact using above builds.
Single failed autotest is known GH issue.

andytudhope:
Please be sure to update this docs issue once this is merged 🎉 

https://github.com/status-im/status.im/issues/51

mandrigin:
@andytudhope https://github.com/status-im/status.im/pull/172

oskarth:
Wow, awesome work, let's hope this one goes through! We've had a bunch of attempts at getting yarn into the code base, so really impressive to see this one progressing so quickly and smoothly (and judging by PR checklist, with all the required changed). Fingers crossed :)

corpetty:
I would like to check determinism of the resulting builds.  I'm happy to see this done quickly and smoothly, but we need to see if we're getting the same resulting hash across multiple platform builds.

mandrigin:
@corpetty it doesn't guarantee determinism in the whole build, because we use other package managers in the build chain too (like CocoaPods), they can affect that.

corpetty:
@mandrigin we could at least hash the part we've changed to see if that is deterministic, then we know we can work forward from there.

corpetty (approved):
We can worry about pushing determinism forward from here.  For now, these changes are great and help multiple things.  Great job @mandrigin

mandrigin:
@corpetty you mean the dependencies? hashes of all of them are in the lock file, and `--frozen-lockfile` that I use everywhere ensures that everything in the **node_modules** is verified against that. Hence, we can ensure that the dependencies always have the same hash.

Regarding the other parts of the toolchain, they are out of what `yarn` is doing.

vitvly (approved):

vkjr:
@mandrigin, what kind of issues you had with `react-native-webview-bridge`? In latest Jenkins build I see that it finished successfully and see no output about this module in console log.

mandrigin:
@vkjr it looks like they were cause by the old `yarn` version, it failed on machine that had yarn 1.8.x. Upgrading yarn to 1.12.3 on the Jenkins machines solved that issue.

vkjr:
Great!
Cool that you renamed package.json. I'm attempted to switch desktop to react-native 0.57 but 2 package.json files caused errors. So waiting for this PR to be commited :)

vkjr (approved):

scripts/run-pre-build-check.sh
```diff
@@ -15,9 +15,15 @@ else
   PLATFORM=$1
 fi
 
-npm_version=$(npm -v)
-if [[ $npm_version != "5.5.1" ]]; then
-  echo -e "${YELLOW}+ npm version $npm_version is installed. npm version 5.5.1 is recommended.${NC}"
+node_version=$(node -v)
+
+if [[ $node_version != "v9.3.0" ]]; then
```
pedropombeiro: we could extract the expected version numbers as variables at the head of the script

pedropombeiro (approved):
Thanks a bunch for doing this! Great job!

scripts/run-pre-build-check.sh
mandrigin: fixed it

mandrigin: sorry for force-push, but I needed to prepare it for merging :)

yenda (approved):

PR #6929, opened 2018-11-28
[#6856] Add ethereum/sign event

### LINKED ISSUE #6856: Add ethereum/sign event
### User Story

As a developer, I want an event to sign data. See https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign

```clojure
[ethereum/sign {:address "" :value "" :on-result [on-sign-result]}] ;; address should be optional
```

### Description

*Type*: Feature

tbenr:
fixes #6856

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -444,6 +444,12 @@
                  :value       :extensions/ethereum-transaction-receipt
                  :arguments   {:value     :string
                                :on-result :event}}
+                'ethereum/sign
+                {:permissions [:read]
+                 :value       :extensions/ethereum-sign
+                 :arguments   {:message?   :string
+                               :data?      :string
+                               :on-result  :event}}
                 'ethereum/send-transaction
                 {:permissions [:read]
                  :value       :extensions/ethereum-send-transaction
--- a/src/status_im/extensions/ethereum.cljs
+++ b/src/status_im/extensions/ethereum.cljs
@@ -17,15 +17,17 @@
             [status-im.native-module.core :as status]))
 
 (handlers/register-handler-fx
- :extensions/transaction-on-result
+ :extensions/wallet-ui-on-result
  (fn [cofx [_ on-result id result method]]
    (fx/merge cofx
              (when on-result
                {:dispatch (on-result {:error nil :result result})})
-             (navigation/navigate-to-clean :wallet-transaction-sent nil))))
+             (cond
+               (= method constants/web3-send-transaction) (navigation/navigate-to-clean :wallet-transaction-sent nil)
+               (= method constants/web3-personal-sign) (navigation/navigate-back)))))
 
 (handlers/register-handler-fx
- :extensions/transaction-on-error
+ :extensions/wallet-ui-on-error
  (fn [{db :db} [_ on-result message]]
    (when on-result {:dispatch (on-result {:error message :result nil})})))
 
@@ -64,8 +66,8 @@
              :gas-price        (when gasPrice
                                  (money/bignumber gasPrice))
              :data             data
-             :on-result        [:extensions/transaction-on-result on-result]
-             :on-error         [:extensions/transaction-on-error on-result]}
+             :on-result        [:extensions/wallet-ui-on-result on-result]
+             :on-error         [:extensions/wallet-ui-on-error on-result]}
       nonce
       (assoc :nonce nonce))))
 
@@ -204,3 +206,19 @@
            registry (get ens/ens-registries chain)]
        (ens/get-addr web3 registry name #(re-frame/dispatch (on-result {:result %}))))
      (re-frame/dispatch (on-result {:error (str "'" name "' is not a valid name")})))))
+
+;; EXTENSION SIGN -> SIGN MESSAGE
+(handlers/register-handler-fx
+ :extensions/ethereum-sign
+ (fn [{db :db :as cofx} [_ _ {:keys [message data id on-result]}]]
+   (if (= (nil? message) (nil? data))
+     {:dispatch (on-result {:error "only one of :message and :data can be used"})}
+     (fx/merge cofx
+               {:db (assoc-in db [:wallet :send-transaction]
+                              {:id                id
+                               :from             (ethereum/normalized-address  (get-in db [:account/account :address]))
+                               :data             (or data (str "0x" (abi-spec/from-utf8 message)))
+                               :on-result        [:extensions/wallet-ui-on-result on-result]
+                               :on-error         [:extensions/wallet-ui-on-error on-result]
+                               :method           constants/web3-personal-sign})}
+               (navigation/navigate-to-cofx :wallet-sign-message-modal nil)))))


jeluard:
@tbenr Looks great! Why is that a WIP ? :)

jeluard (approved):

tbenr:
@jeluard because I haven't tested yet.. :)

tbenr:
i get this error:

```
:code -32601
:message "The method eth_sign does not exist/is not available"
```

calling:

```
{"jsonrpc":"2.0","method":"eth_sign","params":["0x5541e3be81b76d76cdbf968516caa5a5b773763b","0xdeadbeaf"]}
```

I need to investigate more.. (wallet needs to be unlocked?)

src/status_im/constants.cljs
```diff
@@ -200,6 +200,7 @@
 (def ^:const web3-send-transaction "eth_sendTransaction")
 (def ^:const web3-personal-sign "personal_sign")
 (def ^:const web3-get-logs "eth_getLogs")
+(def ^:const web3-sign "eth_sign")
```
flexsurfer: eth_sign is deprecated, but anyway this issue should be solved in different way, we need to open wallet screen directly, added a comment

flexsurfer (changes requested):

flexsurfer:
signing process should be done same as send transaction, so we need to open signing screen (modal wallet) , https://github.com/status-im/status-react/blob/develop/src/status_im/extensions/ethereum.cljs#L61 , should be something like `open-modal-wallet-for-signing-message` , here you can find current implementation https://github.com/status-im/status-react/blob/72e7ae2fffe65639b6e2c13e8929dd42a643c41f/src/status_im/ui/screens/wallet/send/events.cljs#L154

src/status_im/constants.cljs
tbenr: now I get 

```
:code 32700
:message "method is unsupported by RPC interface"
```

flexsurfer: sorry, i edited my comment

jeluard:
@tbenr Sorry for the confusion! Does it make sense now?

tbenr:
@jeluard yes, going to do that!

tbenr:
@jeluard @flexsurfer it works to me...

jeluard:
@tbenr Do you have an example of its usage and the kind of data it returns?

tbenr:
@jeluard yes, but i realize i don't see any message in the signing dialog.. let me verify

tbenr:
ok now it is ok.
that's the idea:

usage 1: plain text message
```
[ethereum/sign {:message "test message"
                  :on-result [store-result]}]
```

usage 2: raw data

```
[ethereum/sign {:data "0x617364"
                  :on-result [store-result]}]
```
you get result in `:result` or error message in `:error`.

it is possible to specify `:address` too (ens resolution is applied) but i don't know if makes sense in our context (can we have multiple account unlocked?)

src/status_im/extensions/ethereum.cljs
```diff
@@ -204,3 +206,25 @@
            registry (get ens/ens-registries chain)]
        (ens/get-addr web3 registry name #(re-frame/dispatch (on-result {:result %}))))
      (re-frame/dispatch (on-result {:error (str "'" name "' is not a valid name")})))))
+
+;; EXTENSION SIGN -> SIGN MESSAGE
+(defn  execute-extension-sign [db {:keys [address message data id on-result]}]
```
jeluard: What happens if the address is not the current wallet one?

src/status_im/extensions/core.cljs
```diff
@@ -444,6 +444,13 @@
                  :value       :extensions/ethereum-transaction-receipt
                  :arguments   {:value     :string
                                :on-result :event}}
+                'ethereum/sign
+                {:permissions [:read]
+                 :value       :extensions/ethereum-sign
+                 :arguments   {:address?   :string
+                               :message?   :string
```
jeluard: I guess this implies we would need some way to express either this argument or this one :)

tbenr: true... when i was playing around with that i see there is a way to do something similar:

```
:arguments  {
                     :value  {:one-of #{
                                  {:message :string}
                                  {:data :string}}}
```

then we have to call

```
[ethereum/sign {:value {:data "0x617364"}}
                  :on-result [store-result]}]

[ethereum/sign {:value {:message "text"}}
                  :on-result [store-result]}]
```

are there other ways?

src/status_im/extensions/ethereum.cljs
tbenr: schermata 2018-11-30 alle 08 02 27

I think makes sense to remove address parameter, unless in some cases we have more then one account unlocked

jeluard:
Right let’s remove it for now.

jeluard:
Nahh it’s perfect as it is now !

tbenr:
> Nahh it’s perfect as it is now !

I added a runtime check ;-) and removed :address

src/status_im/extensions/ethereum.cljs
```diff
@@ -204,3 +206,22 @@
            registry (get ens/ens-registries chain)]
        (ens/get-addr web3 registry name #(re-frame/dispatch (on-result {:result %}))))
      (re-frame/dispatch (on-result {:error (str "'" name "' is not a valid name")})))))
+
+;; EXTENSION SIGN -> SIGN MESSAGE
+(handlers/register-handler-fx
```
flexsurfer, line 211: reformatted little bit according to our code guides 


```
(handlers/register-handler-fx
 :extensions/ethereum-sign
 (fn [{db :db :as cofx} [_ _ {:keys [message data id on-result]}]]
   (if (= (nil? address) (nil? data)) 
      {:dispatch (on-result {:error "only one of :message and :data can be used"}))
     (fx/merge cofx
      {:db (assoc-in db [:wallet :send-transaction]
                       {:id                id
                        :from             (ethereum/normalized-address  (get-in db [:account/account :address])))
                        :data             (or data (str "0x" (abi-spec/from-utf8 message)))
                        :on-result        [:extensions/wallet-ui-on-result on-result]
                        :on-error         [:extensions/wallet-ui-on-error on-result]
                        :method           constants/web3-personal-sign})}
      (navigation/navigate-to-cofx :wallet-sign-message-modal nil)))))
```

flexsurfer (approved):
thanks @tbenr  great job!

asemiankevich:
@jakubgs what is the build generated here for e2e? https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-181130-092726-0169dc-pr.sha256 cc @antdanchenko @hetvart 

<blockquote></blockquote>

jakubgs:
>@jakubgs what is the build generated here for e2e?

Not sure what you are asking. Did you mean to ask "why" instead of "what"? Because if you wanted to ask "what", it's clearly and android end-to-end build.

asemiankevich:
@jakubgs i wanted to ask why the e2e build has `pr.sha256` ending instead of `apk`. Sorry for my broken english

jakubgs:
Oooooh, I didn't notice that, the link is indeed incorrect. Let me check that.

jakubgs:
Thanks for reporting this @asemiankevich, there was indeed a small bug in the change that introduced SHA256 checksums. Fix is here: https://github.com/status-im/status-react/pull/6968
Sorry for misunderstanding, I didn't know you were referring to the github comment with the links.

asemiankevich:
@jakubgs no worries Jacob, thanks for fixing this so fast!

PR #6935, opened 2018-11-29
Use whole message to calculate message-id

cammellos:
Use the whole message so that the confirmation from status-go is picked up

status: ready

### DIFF
--- a/src/status_im/group_chats/core.cljs
+++ b/src/status_im/group_chats/core.cljs
@@ -115,11 +115,9 @@
                                 :dsts          members
                                 :success-event [:transport/message-sent
                                                 chat-id
-                                                (transport.utils/message-id
-                                                 ;; NOTE: There is no clock-value here.
-                                                 ;; Will we have collision?
-                                                 {:from current-public-key
-                                                  :chat-id chat-id})
+                                                (transport.utils/message-id (assoc (:message payload)
+                                                                                   :chat-id chat-id
+                                                                                   :from current-public-key))
                                                 :group-user-message]
                                 :payload       payload}}))))
 


lukaszfryc:
@cammellos I tested it on mobile. Works fine 👍

lukaszfryc:
I started another build to trigger e2e tests.

src/status_im/group_chats/core.cljs
```diff
@@ -115,11 +115,9 @@
                                 :dsts          members
                                 :success-event [:transport/message-sent
                                                 chat-id
-                                                (transport.utils/message-id
-                                                 ;; NOTE: There is no clock-value here.
-                                                 ;; Will we have collision?
-                                                 {:from current-public-key
-                                                  :chat-id chat-id})
+                                                (transport.utils/message-id (assoc (:message payload)
```
yenda, line 118: what is it using from the payload exactly? The `message-id` function is using destructuring and only takes `chat-id` `from` and `clock-value` from the map so I assume the only thing that it's using here from the message payload is `clock-value`?

cammellos: never checked actually,  but looks right, yes

yenda: could you just pass the `clock-value` to the function then?

cammellos: I'd rather not to be honest, I think at this layer we should not care which fields we pass (we pass a well formed `message`), the interface of the function just takes a message and returns a `message-id` (say we change to include an extra field or whatever, this code should ideally not change)

yenda (approved):

src/status_im/group_chats/core.cljs
yenda: ok

rasom (approved):

lukaszfryc:
@cammellos could you merge and cherry pick into release/0.9.32?

cammellos:
@lukaszfryc done

PR #6947, opened 2018-11-29
Fix check for yarn

pedropombeiro:
I've run into a funny issue when switching to `develop`. Even after fixing this typo, it would install `yarn` but not know about it after completing `make setup` because it would be on a bad selection in `nvm` (still pointing to 8.x.x). I had to manually do `nvm use status-im` for things to start working.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,8 @@ clean: ##@prepare Remove all output folders
 	git clean -dxf -f -e android/local.properties
 
 setup: ##@prepare Install all the requirements for status-react
-	./scripts/setup
+	@./scripts/setup
+	@./scripts/run-pre-build-check.sh
 
 prepare-desktop: ##@prepare Install desktop platform dependencies and prepare workspace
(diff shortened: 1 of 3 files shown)


mandrigin (approved):
awesome!

pedropombeiro:
Looks like the issue I had that I needed to close the console where I ran `make setup` and run follow up commands in a new one. Ideally we'd have a warning at the end of the script if stuff like node/yarn is installed.

yenda:
@PombeirP could you also update `run-pre-build-check.sh` so that if $yarn_version is empty it says you have to install it instead of `yarn version is installed...` and also abort instead of keep going if that is possible

jakubgs (approved):
nice

pedropombeiro:
Added better instructions on what to do after `make setup` finishes and node/yarn are still not available on the current console.

jakubgs (approved):
Always good to be clean on what's happening.

PR #6949, opened 2018-11-29
update realm to 2.20.1 and node to 10.14.0

yenda:
Update realm to 2.20.1 to get rid of blocker for update to node 10.14.0

I have tested on desktop and it even worked while concurrently running on the same db as current develop so 2.20.1 is safe to update and that will allow us to finally use latest node version

There is more changes to the lock file for desktop because it's coming from a much older version.

Regarding question for pushing desktop patch on realm to upstream I think it's wiser to wait for the solution that wil get rid of ubuntu-server and see if such a patch is still needed

edit: I needed to patch v2.20.1 with a revert for mobile as well because someone made a commit to support n package manager (sic) and it broke the build for iOS. I was about to propose a PR upstream to revert the commit there but there is already a PR with a cleaner fix https://github.com/realm/realm-js/pull/2109/files

status: ready

### DIFF
--- a/ci/Jenkinsfile.ios
+++ b/ci/Jenkinsfile.ios
@@ -14,6 +14,7 @@ pipeline {
   }
 
   environment {
+    NODE_VERSION = 'v10.14.1'
     BUILD_PLATFORM = 'ios'
     LANG = 'en_US.UTF-8'
     LANGUAGE = 'en_US.UTF-8'
@@ -26,7 +27,7 @@ pipeline {
   stages {
     stage('Prep') {
       steps {
-        script {
+        script { nvm(env.NODE_VERSION) {
           print "Running ${params.BUILD_TYPE} build!"
           /* Necessary to load methods */
           mobile = load 'ci/mobile.groovy'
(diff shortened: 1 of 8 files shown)


ghost:
### Pull Request Checklist

yenda:
waitng for CI to install nvm on iOS slaves @jakubgs

rasom (approved):

pedropombeiro (approved):

yenda:
@mandrigin let's merge?

vkjr:
@yenda, we need to update docs, they still say node shouldn't be of v10 - https://dev.status.im/build_status/

<blockquote><div><strong><a href="https://dev.status.im/build_status/">Status, a Mobile Client for Ethereum 2.0</a></strong></div></blockquote>

pedropombeiro:
Right, they still say:

> Note 2: If you don’t have nvm, node@8 will be installed from Homebrew.
If you don’t have nvm AND already have node installed on your machine then nothing will happen.
Type node -v and make sure you don’t use Node.js v10 because it’s not supported by Realm.js (see Troubleshooting section for additional details).

That's why the PR checklist item is there...

PR #6966, opened 2018-11-30
[TESTS] Added new desktop e2e tests

yevh-berdnyk:
status: ready


test/desktop_sikuli/tests/test_create_account.py
```diff
@@ -64,3 +64,10 @@ def test_switch_accounts(self):
         sign_in.press_enter()
         sign_in.profile_button.click()
         profile.find_text('user_2')
+
+    @pytest.mark.testrail_id(5650)
+    def test_status_log(self):
+        sign_in = SignInView()
+        sign_in.create_account()
+        with open('/root/.local/share/Status/Status.log') as f:
+            assert 'mnemonic' not in f.read()
```
hetvart: shall we use `pytest.fail` here as well?

test/desktop_sikuli/tests/test_recover_access.py
```diff
@@ -26,9 +31,56 @@ def test_recover_access_proceed_with_enter(self):
...
+        sign_in.sign_in_button.click()
+        if not sign_in.element_by_text('Required field').is_visible():
+            self.errors.append("Error 'Required field' is not shown for empty password input")
+        self.verify_no_errors()
+
+    @pytest.mark.testrail_id(5651)
+    def test_recover_account_with_spaces(self):
+        sign_in = SignInView()
+        passphrase = ''.join(
+            list(itertools.chain.from_iterable(zip(base_user['passphrase'].split(), [' ' * i for i in range(1, 13)]))))
```
hetvart: you could use a function like in appium project `fill_string_with_char` -> `test/appium/support/utilities.py`
in my opinion, in test it increases readability

test/desktop_sikuli/views/base_element.py
```diff
@@ -63,14 +76,40 @@ def verify_is_focused(self):
...
+        except Failed:
+            return False
+
+    def verify_element_is_not_present(self):
+        from _pytest.runner import Failed
+        try:
+            self.find_element()
+        except Failed:
+            return
+        pytest.fail("'%s text is displayed but not expected'" % self.text)
```
hetvart: we're looking for an element and if found the method would fail with the message `...text is displayed but not expected`. otherwise, it would catch the exception and nothing return. is it correct?

yevh-berdnyk:
@hetvart I've updated the PR according to your comments, please review again

test/desktop_sikuli/tests/test_recover_access.py
```diff
@@ -16,6 +21,7 @@ def test_recover_access(self):
         profile.share_my_code_button.click()
         profile.find_text(base_user['public_key'])
 
+    @pytest.mark.testrail_id(5571)
```
churik, line 22: This TC includes navigation using "Tab" key, but it is not implemented yet on desktop side.
Should I move it to separate TC?

test/desktop_sikuli/tests/test_profile.py
```diff
@@ -6,13 +6,15 @@
 
 class TestProfile(BaseTestCase):
 
+    @pytest.mark.testrail_id(5590)
```
churik, line 9: I would expect that check for QR code will be added and test will be failed until fix for https://github.com/status-im/status-react/issues/6124

hetvart (approved):

test/desktop_sikuli/tests/test_recover_access.py
yevh-berdnyk: I'd leave it as it is, the test will fail

test/desktop_sikuli/tests/test_profile.py
yevh-berdnyk: This verification will be added after including of pytesseract library

churik (approved):

PR #6967, opened 2018-11-30
[#6957] Wallet shouldn't allows redirect to Home after a transaction …

### LINKED ISSUE #6957: Wallet shouldn't allows redirect to Home after a transaction completion
After a transaction is completed wallet should redirect to the screen that initiated the transaction. Currently it does redirect to the Home page.

This is especially important for the `wallet.settings` hook as current behavior prevents doing anything after the first transaction is completed.

flexsurfer:
fixes #6957

### DIFF
--- a/src/status_im/ui/screens/views.cljs
+++ b/src/status_im/ui/screens/views.cljs
@@ -243,16 +243,43 @@
            :unsigned-transactions        wallet-transactions/transactions
            :transactions-history         wallet-transactions/transactions
            :wallet-transaction-details   wallet-transactions/transaction-details
-           :login                        login})
+           :login                        login
+           :wallet-settings-hook         wallet-settings/settings-hook})
          {:headerMode       "none"
           :initialRouteName "wallet"})}
+
+       :wallet-send-modal-stack
+       {:screen
+        (nav-reagent/stack-navigator (stack-screens {:wallet-send-transaction-modal
+                                                     [:modal send-transaction-modal]
+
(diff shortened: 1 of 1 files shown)


goranjovic:
@flexsurfer @jeluard I'm trying to visualize what is this PR fixing. Can you explain?

src/status_im/ui/screens/views.cljs
```diff
@@ -243,17 +243,44 @@
...
            :wallet-transaction-details   wallet-transactions/transaction-details
-           :login                        login})
+           :login                        login
+           :wallet-settings-hook         wallet-settings/settings-hook})
          {:headerMode       "none"
           :initialRouteName "wallet"})}
+
+       :wallet-send-modal-stack
+       {:screen
+        (nav-reagent/stack-navigator (stack-screens {:wallet-send-transaction-modal
```
jeluard, line 253: What are those?

jeluard (approved):

flexsurfer:
@goranjovic we want to open modal wallet from the wallet O_o we need this for extensions

goranjovic:
@flexsurfer Ok, I understand that. I don't understand what specifically didn't work with the navigation and how this change fixes it

flexsurfer:
@goranjovic before we had this wallet modal stack only for chat, but if you open it from wallet, it moves you to another stack and when you go back you will return to the wrong screen, but all this navigation thing is confusing for sure

goranjovic (approved):

PR #6969, opened 2018-11-30
[#6958] Only render command message when relevant

### LINKED ISSUE #6958: Only render command message when relevant
As a user I don't want to have any extension having register a named command hook be able to access my message details.
Only the right extension (identified by the extension id) should be able to render the command message.

Command message having the correct command name but incorrect extension id will then propose to install the right extension.

flexsurfer:
fixes #6958

### DIFF
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -15,7 +15,8 @@
             [status-im.utils.platform :as platform]
             [status-im.i18n :as i18n]
             [status-im.ui.components.colors :as colors]
-            [status-im.ui.components.icons.vector-icons :as icons]))
+            [status-im.ui.components.icons.vector-icons :as icons]
+            [status-im.chat.commands.protocol :as protocol]))
 
 (defn install-extension-message [extension-id outgoing]
   [react/touchable-highlight {:on-press #(re-frame/dispatch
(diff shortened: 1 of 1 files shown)


ghost:
### Pull Request Checklist

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -30,11 +31,14 @@
...
-      (if-let [extension-id (get-in command-message [:content :params :extension-id])]
+    (let [{:keys [type] :as command} (commands-receiving/lookup-command-by-ref command-message id->command)
+          extension-id (get-in command-message [:content :params :extension-id])]
+      (if (and extension-id (or (not type) (and type (satisfies? protocol/Extension type)
+                                                (not= extension-id (protocol/extension-id type)))))
         [install-extension-message extension-id (:outgoing command-message)]
-        [react/text (str "Unhandled command: " (-> command-message :content :command-path first))]))))
+        (if command
+          (commands/generate-preview command command-message)
+          [react/text (str "Unhandled command: " (-> command-message :content :command-path first))])))))
```
jeluard, line 43: Is this for backward compatibility? Older messages w/o extension id embedded? If so do we get the nice upgrade message if the id is provided but not the right one?

flexsurfer: no, this is for all nonsupported commands doesn't matter if it an extension or not.  We don't have a nice upgrade message, because we don't know if it's an upgrade or different extension, there will be standard install extension message

jeluard: So when do we render the nice installation message exactly? Could you add a comment in the code to clarify?

dmitryn (approved):

flexsurfer:
@jeluard done

src/status_im/ui/screens/chat/message/message.cljs
flexsurfer: done

jeluard (approved):

PR #6972, opened 2018-11-30
Features/pairing prompt

cammellos:
### Description

If a new device is detected, it will prompt the user to pair the devices through a pop up. This pop up will only displayed once for each new `device`, and only if you are not already in the `devices` section.

Currently there are two ways to make a device discoverable:

1) Explicitly, by clicking on `Pair this device`, which reveals information to other devices with the same pk about the `device-type`, eventually the name
2) Implicitly, when pairing is enabled and sending a 1-to-1 message/contact updates. This does not reveal any information to other device with the same pk other than a random id

Also moves `devices` outside `advanced` in desktop to it's own section.

For some of the devices (those implicitly discovered), we don't have info on whether is `desktop` or `mobile`, so probably we should add an icon for this?

 @andmironov @chadyj could you have a quick look if you are ok with ux/copy ?

### Testing

1) Install on device A, create a new account, enable development mode.
2) Install on device B, create a new account, enable development mode.
3) Send a 1-to-1 message to anyone
4) A pop up should be displayed on device B, click on `Go to devices section`
5) You should see a new entry, click on `Enable` and `Pair this device`
6) You should see the same popup on A, go to device section and click on Enable
7) Devices are not paired

status:wip
pairing


goranjovic (approved):

src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs
```diff
@@ -169,6 +178,13 @@
                        :font  (if adv-settings-open? :medium :default)}
            (i18n/label :t/advanced-settings)]
           [vector-icons/icon :icons/forward {:style {:tint-color colors/gray}}]]]
+        [react/touchable-highlight {:style  (styles/profile-row installations-open?)
+                                    :on-press #(re-frame/dispatch [:navigate-to (if installations-open? :home :installations)])}
+         [react/view {:style styles/adv-settings}
+          [react/text {:style (styles/profile-row-text colors/black)
+                       :font  (if adv-settings-open? :medium :default)}
```
flexsurfer: i believe we don't have this anymore, `:font-weight` should be used in styles

flexsurfer: :medium - "500"

flexsurfer: :default - no :font-weight value or "400"

flexsurfer (approved):

chadyj:
Ping @hesterbruikman @PhilipandWu @jpbowen any thoughts on this copy to help users?

hesterbruikman:
> Ping @hesterbruikman @PhilipandWu @jpbowen any thoughts on this copy to help users?

Generally think the copy is clearly writte. I would shorten it to improve readability at glance and at the effect of pairing. Please check for accuracy as this really is just a drive by copy drop:)

```
**New device detected**
Pair and enable the device to sync massages and contacts.

Profile > Settings > Devices

Cancel
Go to Pairing Settings 
```

[Button and instructions need to be an exact reflection. I don't know what the latest Settings menu includes]

Issue #6956, opened 2018-11-30
Upgradable message-ids


yenda:
# Problem

Whenever we make changes to the way message-ids are generated it creates upgradability issues. It happens again in latest release with message responses https://github.com/status-im/status-react/issues/6903#issuecomment-443127013

# Implementation

We need to pass the message-ids with the message to avoid that in the future. These are the problems I think should be taken into consideration:
- pk of the sender and chat-id might be prefixed to the message-id locally to avoid forged message-id overwriting previous messages
- chat-id needs to be consistent so the previous point might just move the problem there

# Acceptance Criteria

- message-id is speced and easily reproducible in other implementations of the communication protocol
- message-id is passed in messages with reasonable guarantees that it cannot be used to overwrite previous messages (ignoring existing ids can be considered but this will open up to MIT attacks if we don't prefix pk, an attacker could silent someone in a public chat by overwritting all their messages for a broad part of the network if they have better connectivity)

mandrigin:
@yenda 

> pk of the sender and chat-id might be prefixed to the message-id locally to avoid forged message-id overwriting previous messages

I don't see how it helps if I myself want to overwrite my own messages, which can be of use for a malicious actor.

mandrigin:
I don't know how bad this idea is, but we can just denormalize and include the whole source message in a reply. Then, we can assign IDs to messages client-side on receiving, so no one will be able to tamper with them.

cammellos:
@yenda @mandrigin we should not overwrite messages (messages should be immutable), it prevents a different scenario, namely a malicious user sending a message with identical `id` as soon as they see it, in which case a race condition is trigger where first-one (not last one) wins.

cammellos:
> we can just denormalize and include the whole source message in a reply


Would need to be signed though, otherwise how can we make sure is not being tampered? 
If we use ids we are back to square one (a user who does not have the original message can't verify that is not tampered), or I am missing something?

mandrigin:
@cammellos in a distributed system you can't really guarantee which message comes first though.
that is why I think that the root of the all issues is allowing ID generation on the sender side, because then it opens the whole can of worms of tampering/races/etc.

actually, the only place where we use these IDs externally is in replies. there, we need to guarantee that the user replied on what he has seen and that everyone will see it. hence, the idea to include the whole message content in the reply.

mandrigin:
one thing that we absolutely have to guarantee about replies is that:
[...]

PR #6976, opened 2018-12-01
fix replies compatibility

### LINKED ISSUE #6903: Replies are shown as ordinary messages if they are sent from current nightly and quoted on release build or when reply to old message sent before upgrade and both contacts are on release build
# Case 1: Replies are shown as ordinary messages if they are sent from current nightly and quoted on release build
### Description

*Type*: Bug
*Summary*: when user has current nightly installed and replies to message, sent from release build, user with release build will see an ordinary message and vice versa.

#### Expected behavior
can see reply message on both A and B

#### Actual behavior
ordinary message on device A and reply message from device B
deblkdjsf

### Reproduction

**Prerequisites:** 
1. device A has installed  0.9.31 version (released) 
2. device B has installed  latest nightly 26/11/2018

- Open Status on device A
- Log in
- Join any channel (i.e. #test)
- Open Status on device A
- Log in
- Join same channel (i.e. #test)
- Send a message from device A
- Reply to message from device B
- Check public chat on device A and B

### Additional Information
* Status version: nightly 26/11/2018
* Operating System: all

# Case 2:  No quotation is shown in the reply message on recipient side if reply to message sent before upgrade

### Description

*Type*: Bug

*Summary*: Recipient gets no quotation in the reply message if reply to message sent before upgrade from 0.9.31 to 0.9.32. On sender side quotation is shown as expected. Both recipient and sender are on 0.9.32. Note that if reply to message sent after upgrade is done on both sides then quotation is shown as expected for both sender and recipient. So, issue is only on recipient side for replies to "old" messages sent while being on 0.9.31.

Recipient (no quotation is shown):

Sender (shows quotation as expected)
screenshot_20181129-160441

#### Expected behavior
Quotation is shown to recipient, sender and other chat participants for replies to messages sent on 0.9.31 and on 0.9.32 
#### Actual behavior
Quotation is not shown to recipient if reply to message sent before upgrade
### Reproduction

1. sender and recipient are on 0.9.31 and they are in contacts of each other
2. both send text message in 1:1 chat
3. both upgrade to 0.9.32
4. sender replies to message from recipient in 1:1 chat
5. check how reply is shown on recipient side

### Additional Information
* Status version: from 0.9.31 release to build 7 of 0.9.32 release

rasom:
fix #6903 
1. `old-message-id` field (indexed) was introduced in `message` entity and is calculated as `message-id` was calculated in `0.9.31`
```clojure
(defn old-message-id
  [message]
  (sha3 (pr-str message)))
```
2. When a reply message is sent from the PR version of app both `response-to` and `response-to-new` (probably better to call it `v2/response-to`) fields are sent as a part of `message`'s `content` field, so that it can be recognized by `0.9.31`.
3. When PR version of app receives reply from `0.9.31` we check whether message's `content` contains `response-to` but doesn't contain `response-to-new`, and if so we check whether DB contains message with `old-message-id=response-to`. If such message has been found we assoc `response-to-new` to content.
4. If message from DB contains only `response-to` but not `response-to-new` attempt to fetch the message by `old-message-id` is done (can be also handled by migration)

status: ready

### DIFF
--- a/src/status_im/chat/models/input.cljs
+++ b/src/status_im/chat/models/input.cljs
@@ -87,10 +87,12 @@
 
 (fx/defn reply-to-message
   "Sets reference to previous chat message and focuses on input"
-  [{:keys [db] :as cofx} message-id]
+  [{:keys [db] :as cofx} message-id old-message-id]
   (let [current-chat-id (:current-chat-id db)]
     (fx/merge cofx
-              {:db (assoc-in db [:chats current-chat-id :metadata :responding-to-message] message-id)}
+              {:db (assoc-in db [:chats current-chat-id :metadata :responding-to-message]
+                             {:message-id     message-id
+                              :old-message-id old-message-id})}
               (chat-input-focus :input-ref))))
 
 (fx/defn cancel-message-reply
@@ -121,15 +123,17 @@
   "no command detected, when not empty, proceed by sending text message without command processing"
   [input-text current-chat-id {:keys [db] :as cofx}]
   (when-not (string/blank? input-text)
-    (let [reply-to-message (get-in db [:chats current-chat-id :metadata :responding-to-message])]
+    (let [{:keys [message-id old-message-id]}
+          (get-in db [:chats current-chat-id :metadata :responding-to-message])]
       (fx/merge cofx
                 {:db (assoc-in db [:chats current-chat-id :metadata :responding-to-message] nil)}
                 (chat.message/send-message {:chat-id      current-chat-id
                                             :content-type constants/content-type-text
                                             :content      (cond-> {:chat-id current-chat-id
                                                                    :text    input-text}
-                                                            reply-to-message
-                                                            (assoc :response-to reply-to-message))})
+                                                            message-id
+                                                            (assoc :response-to old-message-id
+                                                                   :response-to-v2 message-id))})
                 (commands.input/set-command-reference nil)
                 (set-chat-input-text nil)
                 (process-cooldown)))))
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -163,10 +163,20 @@
                                               :to    chat-id
                                               :from  from}}))))
 
+(defn check-response-to
+  [{{:keys [response-to response-to-v2]} :content :as message}
+   old-id->message]
+  (if (and response-to (not response-to-v2))
+    (let [response-to-v2
+          (or (get-in old-id->message [response-to :message-id])
+              (messages-store/get-message-id-by-old response-to))]
+      (assoc-in message [:content :response-to-v2] response-to-v2))
+    message))
+
 (fx/defn add-received-message
-  [{:keys [db now] :as cofx}
-   batch?
-   {:keys [from message-id chat-id js-obj] :as raw-message}]
+  [{:keys [db] :as cofx}
+   old-id->message
+   {:keys [from message-id chat-id js-obj content] :as raw-message}]
   (let [{:keys [web3 current-chat-id view-id]} db
         current-public-key            (accounts.db/current-public-key cofx)
         current-chat?                 (and (or (= :chat view-id)
@@ -176,12 +186,13 @@
         message                       (-> raw-message
                                           (commands-receiving/enhance-receive-parameters cofx)
                                           (ensure-clock-value chat)
+                                          (check-response-to old-id->message)
                                           ;; TODO (cammellos): Refactor so it's not computed twice
                                           (add-outgoing-status current-public-key))]
     (fx/merge cofx
               {:transport/confirm-messages-processed [{:web3   web3
                                                        :js-obj js-obj}]}
-              (add-message batch? message current-chat?)
+              (add-message true message current-chat?)
               ;; Checking :outgoing here only works for now as we don't have a :seen
               ;; status for public chats, if we add processing of our own messages
               ;; for 1-to-1 care needs to be taken not to override the :seen status
@@ -209,17 +220,24 @@
              (messages-store/message-exists? message-id)))))
 
 (defn- filter-messages [cofx messages]
-  (:accumulated (reduce (fn [{:keys [seen-ids] :as acc}
-                             {:keys [message-id] :as message}]
-                          (if (and (add-to-chat? cofx message)
-                                   (not (seen-ids message-id)))
-                            (-> acc
-                                (update :seen-ids conj message-id)
-                                (update :accumulated conj message))
-                            acc))
-                        {:seen-ids    #{}
-                         :accumulated []}
-                        messages)))
+  (:accumulated
+   (reduce (fn [{:keys [seen-ids] :as acc}
+                {:keys [message-id old-message-id] :as message}]
+             (if (and (add-to-chat? cofx message)
+                      (not (seen-ids message-id)))
+               (-> acc
+                   (update :seen-ids conj message-id)
+                   (update :accumulated
+                           (fn [acc]
+                             (-> acc
+                                 (update :messages conj message)
+                                 (assoc-in [:by-old-message-id old-message-id]
+                                           message)))))
+               acc))
+           {:seen-ids    #{}
+            :accumulated {:messages     []
+                          :by-old-message-id {}}}
+           messages)))
 
 (defn extract-chat-id [cofx {:keys [chat-id from message-type]}]
   "Validate and return a valid chat-id"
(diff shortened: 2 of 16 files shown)


asemiankevich:
@rasom would you mind to rebase? e2e is ending again with `sha256` but i remember @jakubgs has fixed that on Friday

asemiankevich:
@rasom case 2 from https://github.com/status-im/status-react/issues/6903 is still actual and i can reproduce it the way it is described:

1. sender and recipient are on 0.9.31 and they are in contacts of each other
2. both send text message in 1:1 chat
3. both upgrade to 0.9.32
4. sender replies to message from recipient in 1:1 chat
5. check how reply is shown on recipient side

NOTE: case 1 from the issue above is fixed.

cc @churik @annadanchenko

rasom:
@asemiankevich `case 2` should be fixed now, branch was rebased

asemiankevich:
@rasom thanks :) would be nice to have some rewies there :)

asemiankevich:
so I dont see none of issues described in this PR build. @churik please confirm too (just to double check)

cammellos:
@rasom could you add a description on the PR of what the logic is please?

Also, just to make sure, this does not change the id as described in https://github.com/status-im/status-react/issues/6956 , so if we push this out without, we probably want to make sure there's a clean upgrade path?

rasom:
yes sure @cammellos, working on the description

> this does not change the id as described in #6956 (comment)

this pr doesn't change the way how `message-id` is calculated, right

rasom:
@cammellos see description, idk if that's enough

cammellos:
@rasom thanks looks good.
In terms of upgrade path, this is non breaking, if we decided to go for https://github.com/status-im/status-react/issues/6956#issuecomment-443622618 , is there a clean way not-to break compatibility? or we are ok breaking when/if we implement that? I think it would be good to figure it out before we push this out

rasom:
>  I think it would be good to figure it out before we push this out

This is a separate thing, but yes, i would prefer to fix `message-id` before 0.9.32 release. Otherwise, we will have compatibility issues next time as well. I don't think we can guaranty anything in nightlies, it's just too much overhead imo.

churik:
Both cases are not reproducible for me.
Case 1 checked on IOS 12.1, Iphone XS (release build) and Mac OS Mojave (PR build).
Case 2 checked on IOS 11.4, Iphone 7 and Android 8.0, Google Pixel (Genymotion) for public and 1-1 chat.

asemiankevich:
@yenda @mandrigin review this please. @cammellos @rasom are you planning to create separate PR to fix the remaining stuff? we are planning this release schedule

cammellos:
On my side, before including this I would like to answer this questions:

1) If we include it, and then change id scheme, is there a clear path to upgrade id without breaking?
2) If the answer of point 1 is `no`, are we ok with breaking compatibility for replies/seen etc? Otherwise cherry-pick
3) If the answer of point 2 is `no`, should we then implement the `id` scheme that we want before pushing this out?

src/status_im/chat/subs.cljs
```diff
@@ -263,4 +263,6 @@
  :chats/reply-message
  :<- [:chats/current-chat]
  (fn [{:keys [metadata messages]}]
-   (get messages (:responding-to-message metadata))))
+   (get messages (get-in metadata [:responding-to-message :message-id]))))
+;; "0x89957b18854848acb520c35561725ae457ada9c5b7f89fedb3b4b1d84fc33a86"
```
mandrigin: is this a leftover?

src/status_im/chat/models/message.cljs
```diff
@@ -176,12 +186,13 @@
...
                                           (commands-receiving/enhance-receive-parameters cofx)
                                           (ensure-clock-value chat)
+                                          (check-response-to old-id->message)
                                           ;; TODO (cammellos): Refactor so it's not computed twice
                                           (add-outgoing-status current-public-key))]
     (fx/merge cofx
               {:transport/confirm-messages-processed [{:web3   web3
                                                        :js-obj js-obj}]}
-              (add-message batch? message current-chat?)
+              (add-message true message current-chat?)
```
mandrigin, line 195: do we call it with `false` anywhere?

src/status_im/data_store/realm/schemas/account/migrations.cljs
```diff
@@ -240,3 +243,34 @@
...
+  (string/replace-first
+   str-message
+   "status-im.data-store.realm.schemas.account.migrations"
+   "status-im.transport.message.protocol"))
+
+(defn old-message-id
+  [message]
+  (sha3 (replace-ns (pr-str message))))
+
+;; TODO (rasom): add migration for response-to before merging
```
mandrigin: was this done?

mandrigin (changes requested):
a few questions after a first glance

src/status_im/chat/models/message.cljs
rasom: yes, we do https://github.com/status-im/status-react/blob/812a2454aaa7d59f128b699016fdbb5c7aa103e7/src/status_im/chat/models/message.cljs#L346

src/status_im/data_store/realm/schemas/account/migrations.cljs
rasom: no need to do this, the comment is removed

src/status_im/chat/subs.cljs
rasom: fixed

rasom:
@cammellos 

> 1. If we include it, and then change id scheme, is there a clear path to upgrade id without breaking?

there is always a way to upgrade it without breaking, the question is how complicated it will be. If we change id scheme after merging this PR and before releasing `0.9.32` for those who upgrade from `0.9.31` upgrading will work the same way as without changing id schema (because it already has been changed after 0.9.31). Idk if that answers your question.

cammellos:
> there is always a way to upgrade it without breaking, the question is how complicated it will be.

yes, seems like having an `old-old-message-id` (or something on those lines) might not be desirable, but necessary if we include this PR as it is, are we ok to pull the trigger?

rasom:
@cammellos in overall:
there is no way to introduce change to message id schema and make functionality like our current implementation of replies (which relies on message id) compatible with the previous version of the app without adding some complexity like `very-very-old-message-id` and without adding necessary migrations for this.

cammellos (approved):

src/status_im/chat/models/input.cljs
```diff
@@ -87,10 +87,12 @@
 
 (fx/defn reply-to-message
   "Sets reference to previous chat message and focuses on input"
-  [{:keys [db] :as cofx} message-id]
+  [{:keys [db] :as cofx} message-id old-message-id]
```
mandrigin, line 90: I would call them `message-id-v1` and `message-id-v2` here and everywhere.

mandrigin (changes requested):

src/status_im/chat/models/input.cljs
rasom: This PR will not be reviewable with such change and it will touch literally everything in the project with hundreds of possible regression. I really don't want to do this now.

mandrigin: @rasom can you make a follow-up PR with just that? So we avoid this old-old thing?

rasom: Do we really need this? I can make it, i just don't see how it will make things better.

mandrigin: If we upgrade ID one more time, having no real naming convention will make a huge mess.

mandrigin (approved):

src/status_im/chat/models/input.cljs
rasom: So that's what we need to do, introduce proper versioning mechanism with naming conventions, etc. We will not solve this right now by renaming `message-id` but will have huge chances to introduce a lot of other problems.

rasom:
I will merge this PR a bit later if it's test and ready for merging @asemiankevich

src/status_im/chat/models/input.cljs
mandrigin: yeah, I agree that this renaming/convention shouldn't be in the release

asemiankevich:
@rasom it was tested before the last commit was done :)

rasom:
@asemiankevich there was no "Tested OK" label, that's why i wanted to check

Serhy:
Please note that autotests https://github.com/status-im/status-react/pull/6976/#issuecomment-443808447 were run not agains latest commit (build from https://github.com/status-im/status-react/pull/6976/#issuecomment-443717773 ), not the  https://github.com/status-im/status-react/pull/6976/commits/cb12e7e629157428dad74ffb7c3a4c3d2586975f

rasom:
will be merged in #6987

PR #6986, opened 2018-12-03
Tell user to run `nvm use` in terminal if needed

pedropombeiro:
This PR addresses the fact that calling `nvm use` in the setup script only works inside that script (the `NVM_DIR` env var is not exported in the parent process, so if the user didn't have `status-im` active going in, it's not going to be active anyway after running `make setup`, therefore we need to instruct him to do so manually).

status: ready

### DIFF
--- a/scripts/run-environment-check.sh
+++ b/scripts/run-environment-check.sh
@@ -25,7 +25,16 @@ fi
 
 if ! program_exists node || ! program_exists yarn; then
   echo -e "${YELLOW}********************************************************************************************"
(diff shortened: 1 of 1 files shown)


ghost:
### Pull Request Checklist

mandrigin (approved):

pedropombeiro:
Ping @jakubgs

jakubgs (approved):
Sorry, must have missed the notification. Looks good.

Issue #6080, opened 2018-10-01
"Not sent" message is shown as "Sent" if it was sent during 2 sec after coming online


churik:
### Description

*Type*: Bug
*Summary*: if you send message immediately after switching off offline mode - it is shown as Sent, but in fact it wasn't sent.

#### Expected behavior
status is "Not sent. Tap for options" with resend option.

 <img width="200" alt="offl" src="https://user-images.githubusercontent.com/7532782/47304308-81d32580-d626-11e8-8ab0-f98061d5e182.png">

#### Actual behavior
status is "sent"
<img width="491" alt="offl" src="https://user-images.githubusercontent.com/4557972/46295006-bdd31780-c59f-11e8-8007-d9d55778167f.png">

### Reproduction

- Open Status
- Type a message
- Turn off WI-FI
- Wait for 10 sec
- Turn on WI-FI and immediately (during 2 sec) send a message

### Additional Information
* Status version: nightly 01/10/2018
* Operating System: Android, IOS, desktop
* TF session: https://app.testfairy.com/projects/4803590-status/builds/8611567/sessions/4402803746/?accessToken=tykYFpL1X6sUO6Njg9IAkpeAVAQ

In status-desktop it is reproducible in another way as well in #5919  (look at https://github.com/status-im/status-react/pull/5919#issuecomment-425896202)

lukaszfryc:
I also experienced it recently. 

I think we should increase sev and priority of this issue as it breaks confidence in chats.

lukaszfryc:
This can be also reproduced in 0.9.30

dshulyak:
The root cause is probably the same as in https://github.com/status-im/status-go/issues/1276

Issue #6961, opened 2018-11-30
Use peer count as a definition for Online/Offline within status desktop


vitvly:
## Description

Currently, we rely on React Native's NetInfo API to check whether there is internet connectivity. On desktop, for instance, this presents a problem as by default react-native-desktop's NetInfo module uses `google.com` in order to check network connectivity.

## Solution
Although it is now possible to invoke `setConnectionCheckUrl()` on Desktop in order to set a custom URL, it would be better to instead rely on active peers count as a definition of being online.

vkjr:
@siphiuel, I narrowed a bit scope for this issue, to fit desktop release.
@cammellos made a #7012 pr that fix all we need for desktop but it doesn't touch mobile wallet that still uses `NetInfo`. 

If we want to get rid from `NetInfo` in wallet, new issue can be added. (cc) @goranjovic

vkjr:
#7012 merged, so closing this one

Issue #6449, opened 2018-10-22
Desktop messages can't be resent


oskarth:
# Problem

As a desktop user I want to be able to resend messages by clicking on them. Right now it is just red text.

# Implementation

Similar to mobile, click and get asked to resend.

# Acceptance Criteria

Can obviously resend messages.

# Notes

N/A

# Future Steps

See if we can share more code with mobile to make the same thing possible on both platforms. When that isn't possible, consider exposing it in Desktop but marking it as a TODO by default, with a link to new issue (build with users).

asemiankevich:
@oskarth as far as i know `resend` feature is disabled on mobile . @janherich may clarify something more

janherich:
@asemiankevich I just checked, and on mobile, it's not even possible to send messages while user is offline, so hard to tell. I will check in the code, but I'm not aware of and was not really involved in any discussion about disabling message re-sending, better to ask @dmitryn since he originally developed the feature.

asemiankevich:
@janherich oh jeez i am sorry i thought you designed everything in Status :D indeed @dmitryn could clarify a bit

oskarth:
Huh, why did we disable a useful feature?

lukaszfryc:
> it's not even possible to send messages while user is offline, so hard to tell.

Actually, it is possible. The app does not detect you are offline immediately. So, if you sent a message before seeing "Offline" status, that message will be marked as "sent" but lost (https://github.com/status-im/status-react/issues/6415)

dmitryn:
> as far as i know resend feature is disabled on mobile

I didn't know it was disabled. @asemiankevich can you point to issue / PR where it has happened? 

As for desktop, resending might be done in similar way using mobile client code. Desktop just needs its own UI part.

asemiankevich:
@dmitryn @oskarth seems i am wrong. I thought it was disabled. sorry for confusion

churik:
@asemiankevich @oskarth  but there is no more "Not sent" status (AFAIK) and I couldn't see this feature for a while.
That's why https://github.com/status-im/status-react/issues/6415 and https://github.com/status-im/status-react/issues/6080 were reported (that's where resend options should be used IMO).

From offline mode it is not possible to send a message (button is disabled). For desktop - it is possible (send button is enabled).
I'm not sure where and how for now and if it is even possible to trigger message in "Not sent" status for mobile.

So probably we need just make it consistent between mobile and desktop and find out where "Not sent" status should be used for both platforms?

[...]

Issue #6396, opened 2018-10-18
Got "Connecting to peers" or "Offline" and couldn't send new messages after going back from sleep mode 


churik:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug
*Summary*: can reproduce it constantly on Mac OSx VM 10.13.6, but cannot find exact action on the local computer.
When VM came back from sleep mode, app doesn't detect online mode and all messages, sent from this state will be never received.

#### Expected behavior
can send\receive messages

#### Actual behavior
"Connecting to peers", messages are displayed as sent, but they are not sent.

### Reproduction
[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Log in
- Join public channel
- Go to sleep mode for ~5 mins
- Go back from sleep mode

NOTE: issue is gone after app restarting

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: desktop nightly 18/10/18
* Operating System: Mac OSx 10.13.6 (used VM in Parallels)

#### Logs

 https://drive.google.com/open?id=1R9nhxD83qYycpUYNeUHdnYJR57WU4HLy

<blockquote><div><strong><a href="https://drive.google.com/open?id=1R9nhxD83qYycpUYNeUHdnYJR57WU4HLy">Meet Google Drive – One place for all your files</a></strong></div><div>Google Drive is a free way to keep your files backed up and easy to reach from any phone, tablet, or computer. Start with 15GB of Google storage – free.</div></blockquote>

chadyj:
@goranjovic Did you have this issue on Linux too?

goranjovic:
@chadyj No, but I do constantly have it on Mac

churik:
Can be reproducible on mobile as well, only restarting app helps (faced with this on IOS and on Android 7.0 (Genymotion emulator, Samsung Galaxy S7))

goranjovic:
Also two hackaton participants mentioned that they had this problem on genymotion run within osx. Restarting the app helped.

churik:
See: https://github.com/status-im/status-react/issues/6284#issuecomment-438699307 for more info

cammellos:
@churik can you still replicate? I can't replicate on my local machine or android. If you do, it would be helpful to get the `Status.log`.

So far what I have:

[...]

PR #7012, opened 2018-12-05
Use peers count

### LINKED ISSUE #6889: Update chosen mail servers with new library binding UpdateMailservers
# Problem

We need to know which mail servers are selected by the user. Discussion is in https://github.com/status-im/status-go/issues/1285. Binding accepts json-encoded list of enodes. And returns error if any enode cannot be unmarshalled into enode.Node data objects.

https://github.com/status-im/status-go/blob/8b310db28d0a535804f61a274900ebca20a40821/lib/library.go#L467

# Implementation

Update must be made before doing any requests. For now status-react can preserve all other logic, status-go part will use mail server that was added by status-react. If mail server changes status-go must be notified of this change.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/status-im/status-go">status-im/status-go</a></strong></div><div>The Status module that consumes go-ethereum. Contribute to status-im/status-go development by creating an account on GitHub.</div></blockquote>

cammellos:
This PR changes the behavior of `Offline` for sending messages and adds mailserver confirmations.

### Offline/Online behavior

We now check that we are only connected to some `peers` instead of using `NetInfo` from `react-native`. 

This is because it has been reported to be quite flaky at times, not reporting online status after sleeping, and for privacy concerns (on ios it pings `apple.com`, on desktop `google.com`).

Adds a new banner `Wallet Offline` and change `Connecting to peers` to  `Chat offline`.

### Mailserver confirmations

A message will be marked as `Sent` only if it made it to the mailserver you are connected to, which will increase the guarantees that we can make about a message (if you see it as sent, it has reached at least a mailserver), this has the consequence that:

- If you are not connected to any mailserver or the mailserver is non responsive/down, and you send a message, it will be marked as `Not sent`, although it might have been actually made it in the network.

Probably this is something that we would like to communicate to the user through UX (i.e. tick if made it to at least a peer, double tick if it made to a mailserver )

Currently I have only enabled this feature in nightlies & devs, I would give it a run and see how we feel about it.

fixes: #6889, #6080  #6396, #6889, #6449

status: ready


chadyj:
Thanks @cammellos!
My 2 cents: `offline` is there to help safeguard users so they don't send a message and have it lost, or start a transaction and have it not go through. It would be better for the app to err on the side of caution to prevent messages or transactions failing.  Showing a user `offline` and then allowing them to send a message creates uncertainty and can erode trust, so we should fix that.  What UX change are you thinking about?

cammellos:
> to help safeguard users so they don't send a message and have it lost

This will make a bit more probable that the user will send a message while the app is actually offline, but:
- One major thing that was fixed yesterday is that all messages were marked as `sent` even if they were not sent https://github.com/status-im/status-react/pull/6880
- Confirmation will be asked from the mailserver, which means a message will only be marked as sent if it reached one mailserver, so will give more guarantees to the user.

So overall I would be inclined to see how reliable it feels after this changes.
On top of that it would solve the issue of the user being stuck in Offline mode and able to receive messages but not send.

In terms of UX:

Basically we could have different banners now:

1) Offline -> Both wallet & chat are offline
2) Wallet Offline -> wallet offline & chat online
3) Chat offline -> (currently that's actually "connecting to peers", but I think Chat offline makes more sense, as the user does not really know what peers are)
4) Nothing -> Online

Essentially this means just adding a banner "Wallet offline" and renaming "Connecting to peers.." to "Chat offline"

chadyj:
> Essentially this means just adding a banner "Wallet offline" and renaming "Connecting to peers.." to "Chat offline"

👍  cc @errorists

yenda (approved):

yenda:
@cammellos seems like the netInfo api is still used for regular offline boolean? Is it planned to remove it in another PR?

cammellos:
@yenda kind of, I think before we can remove it we need to sort out how to signal the app is offline to wallet.
Using peers count couples wallet to whisper which might not be what we want.
 
We had a brief discussion with @goranjovic and he mentioned that the equivalent would be pinging infura to check we are online, but wasn't discussed further.

yenda:
@cammellos ok but until it's done the privacy problem isn't solved then. Also instead of pinging we could use the timeout of http requests. Infura isn't the only website contacted by the wallet etherscan is as well.

cammellos:
> use the timeout of http requests

You mean for wallet? So you mean make a "real" request when the user wants, if it fails set offline, and then retry periodically the real request and mark it as online when succeeds? or you mean something different?

yenda:
@cammellos yes I mean set to offline if it fails, possibly better solution is to set "X is offline" until you get 2 different offline domain then you can say "offline", otherwise it might just be a timeout on one service.

cammellos:
I don't mind doing that, it's a slow feedback but depends on what's needed for wallet to function properly, probably something for the wallet team to think about

yenda:
@cammellos I think the first request is done when you login so at least there is already an initial feedback. Also while in app it's not very different from most mobile apps there is usually a delay when you lose connectivity before you are notified that something wrong is going on.

rasom (approved):

mandrigin:
@cammellos so this will fix https://github.com/status-im/status-react/issues/6889 too, right?

mandrigin (approved):
lgtm but should be merged after 0.17.1 is merged (in testing now): https://github.com/status-im/status-react/pull/6877

mandrigin:
@cammellos can you take a look at the merge conflict here?

cammellos:
@mandrigin done!

churik:
@cammellos we should test it only after https://github.com/status-im/status-react/pull/6877  merge, is it right?

cammellos:
@churik it can be tested at the same time, this version contains a more up-to-date version of `status-go`, then https://github.com/status-im/status-react/pull/6877 (includes all commits + 1 extra)

churik:
@cammellos 
1) On desktop one time per 1 minute can see "Wallet offline" and then "Fetching messages" when the connection is stable
Steps:
- install build from http://take.ms/6GFOy
- log in
- join any public chat 
- wait 1 minute

Not reproducible for mobile.
Video: http://take.ms/6GFOy
Geth.log: [geth.log](https://github.com/status-im/status-react/files/2663094/geth.log)

2) Q - Wallet Offline should be shown on wallet screen? On mobile could't see this banner.

3) as a bonus it fixes https://github.com/status-im/status-react/issues/6080 and https://github.com/status-im/status-react/issues/6396 (will update issue description)


<blockquote><div>Monosnap screenshot tool</div><div><strong><a href="https://monosnap.com/file/MUGxnPXk1HWKlD1Z3Om7U6wFkLLF4X">File "screencast 2018-12-10 13-16-44.mp4"</a></strong></div><div>Monosnap — the best tool to take & share your screenshots.</div></blockquote>

cammellos:
@churik 
`Wallet offline` is expected, it means that we have been reported as offline by `react-native` (which is flaky on desktop), which was the original issue for https://github.com/status-im/status-react/issues/6396 .

On mobile you could also see it, but it's a bit harder as `react-native` is more reliable there.

Given that there's no support for wallet on desktop, should I just hide it there?

churik:
@cammellos yes, please.
Also what should be in STATUS_GO_VERSION, newest one?
Should we retest go version in terms of this PR as well?

cammellos:
@churik will do that, thanks

In terms of `status-go`, this is what has been tested in #6877 + the release hot fix, so it should not have a big impact.

churik:
@cammellos one more issue - if you sent several messages in a row (~20-30) one of them have a spinner for a while and then marked as not sent.
Can reproduce for both mobile and desktop.
Steps:
- join public channel
- send several (~20-30) messages in a row
For me delivered were 53 messages from 57, 4 messages are marked as not sent, but actually, they were sent so I can see them from another device


Log: [geth.log](https://github.com/status-im/status-react/files/2663730/geth.log)

cammellos:
cc: @dshulyak 

I will have a look at status-react code to make sure we miss some confirmations, @dshulyak if you have some spare cycles, could you have a look at `status-go`?

dshulyak:
> I will have a look at status-react code to make sure we miss some confirmations, @dshulyak if you have some spare cycles, could you have a look at status-go?

@cammellos the only thing that comes to my mind is that connection to a chosen mail server was flaky. so we "sent" a message to a regular peer, but it was marked as not sent. because confirmations were restricted to a mail server.

cammellos:
@churik that's likely because the connection with the mailserver was dropped so the message had been marked as `not-sent`.

I would change `not-sent` to `not-confirmed`, so that should be clearer to the user, and then we can investigate on mailserver connection problems, what do you think?

churik:
@cammellos as for me, `not-confirmed` looks for user even more confusing that `not-sent`.
It is happening quite often, steps described in https://github.com/status-im/status-react/pull/7012#issuecomment-445858820 are just shortest way to reproduce it.

So basically question is - can we merge PR which fixes #6080, #6396, #6889 and introduces new issue when sometimes message will be marked as `not-confirmed`?
WDYT @annadanchenko @asemiankevich ?

cammellos:
@churik also, this behavior is not enabled in relases, only in nightlies/dev (specifically for this reason), as I wanted to get a bit more feedback before pulling the trigger

mandrigin:
@churik @cammellos if the mailservers are flaky, this PR uncovers an issue that potentially might lead to messages silently not being delivered. But, of course, more tests are needed. Maybe we can create a separate GHI?

asemiankevich:
> @cammellos as for me, `not-confirmed` looks for user even more confusing that `not-sent`.
> It is happening quite often, steps described in [#7012 (comment)](https://github.com/status-im/status-react/pull/7012#issuecomment-445858820) are just shortest way to reproduce it.
> 
> So basically question is - can we merge PR which fixes #6080, #6396, #6889 and introduces new issue when sometimes message will be marked as `not-confirmed`?
> WDYT @annadanchenko @asemiankevich ?

yes please create an issue

churik:
@cammellos please, replace it to `not-confirmed` and I'll retest all changes and I'll create separate GHI for https://github.com/status-im/status-react/pull/7012#issuecomment-445858820

cammellos:
@churik ready to be tested, I have changed `not-sent` to `not-confirmed`, thanks!

vkjr:
@cammellos, If I understand correctly this PR also fixes #6961.

cammellos:
@vkjr correct, although I haven't disabled `NetInfo` it is not used anymore on desktop (on mobile we still use it for connectivity checks for wallet)

vkjr:
#6961 is important for desktop release (#6926), so I will narrow the scope of that issue :)

churik:
@cammellos latest builds are failed constantly (tried several times, last https://ci.status.im/job/status-react/job/pull%20requests/view/change-requests/job/PR-7012/22/ )
Can you help or I should ping smbd from #core-infra?

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/pull%20requests/view/change-requests/job/PR-7012/22/">status-react » pull requests » PR-7012 #22 [Jenkins]</a></strong></div></blockquote>

cammellos:
@churik not sure, let me try once more

annadanchenko:
@cammellos As I understood from https://github.com/status-im/status-react/pull/7012#issuecomment-444448914 there should be new labels "Wallet offline" and "Chat offline", however if mobile is in flight mode then 
1. in Home tab and inside public chat only "Offline" label is shown (would expect "Chat offline")
2. On main Wallet screen no labels about Offline is shown

img_2552
img_2551

cammellos:
@annadanchenko it works slightly different:

1) If both chat & wallet are offline (airplane mode for example), you will see "Offline"
2) If chat is online but wallet offline you will see "Wallet offline"
3) If wallet is online but chat is offline you will see "Chat offline"

Both "Wallet offline" & "Chat offline" are difficult to replicate as it happens mainly on desktop, where the notifications of online/offline is flakier (but happens sometimes on mobile too).

As for the Wallet without the notification, I haven't changed that I don't think we ever displayed the "Offline" notification there.

churik:
Autotests failed tests are false failures.

Tested on https://github.com/status-im/status-react/pull/7012#issuecomment-446534113.
On Mac OS and Windows:
- fetching history for 24 hours
- relogin while offline
- offline for ~5 mins in background and then go back online 
- switching between WI-FI networks 
- slow connection
- switching mailsevers
- resend messages 
- close app -> switch off WI-FI -> open app ->switch on WI-FI

 On IOS 11.4 (IPhone7):
- fetching history for 24 hours
- сan switch to custom mailserver, login and receive/send messages
- offline for ~5 mins in background and then go back online 
- switching between WI-FI networks 
- resend messages 

From me nothing to prevent this PR from merge.
cc @annadanchenko

annadanchenko:
I'll provide my feedback in 1-2 hours, was investigating other high-prio issue, now back to testing for this PR

annadanchenko:
ok to merge from my side too

PR #7014, opened 2018-12-05
store :raw-payload-hash in message

rasom:
Currently, we calculate `message-id` as `sha3(from + raw_payload)`, but we do not store `raw_payload` and it might be problematic to restore it from DB because:
1) `content` field might be changed and so `Message` record will differ from the original one
2) it is even more problematic for `GroupMembershipUpdate` message because we don't save it in DB

In order to handle this, we can store `sha3(raw_payload)` as `raw-payload-hash` prop of `message` entity and use it in case of emergency :) 

`message-id` will be calculated as `sha3(from + sha3(raw_payload))`

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -307,7 +307,8 @@
                  :content-type constants/content-type-status}]
     (assoc message
            :message-id (transport.utils/system-message-id message)
-           :old-message-id "system")))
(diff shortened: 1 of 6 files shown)


mandrigin (approved):
lgtm

cammellos (approved):
future proofing, thanks!

PR #7018, opened 2018-12-05
[#6866] Add ipfs/add extension event

### LINKED ISSUE #6866: Add a ipfs/add extension event
Add a `ipfs/add` extension event allowing to upload a data structure to IPFS. It must also support adding blobs (like images).
Rely on Infura [API](https://infura.io/docs/ipfs/post/add)  for its implementation.

An example can be found [here](https://github.com/status-im/pluto/blob/master/website/pages/try.html#L37)

```clojure
[ipfs/add {:value ... :on-result [some-event]}]
```

tbenr:
fixes #6866

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -13,6 +13,7 @@
             [status-im.ui.screens.wallet.settings.views :as settings]
             [status-im.i18n :as i18n]
             [status-im.utils.money :as money]
+            [status-im.constants :as constants]
             [status-im.ui.components.colors :as colors]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.handlers :as handlers]
@@ -196,7 +197,7 @@
 (re-frame/reg-event-fx
  :ipfs/cat
  (fn [_ [_ _ {:keys [hash on-success on-failure]}]]
-   {:http-raw-get (merge {:url (str "https://ipfs.infura.io/ipfs/" hash)
+   {:http-raw-get (merge {:url (str constants/ipfs-cat-url hash)
                           :success-event-creator
                           (fn [{:keys [status body]}]
                             (if (= 200 status)
@@ -206,6 +207,30 @@
                          (when on-failure
                            {:failure-event-creator on-failure})
                          {:timeout-ms 5000})}))
+
+(defn- parse-ipfs-add-response [res]
+  (let [{:keys [Name Hash Size]} (parse-json res)]
+    {:name Name
+     :hash Hash
+     :size Size}))
+
+(re-frame/reg-event-fx
+ :ipfs/add
+ (fn [_ [_ _ {:keys [value on-success on-failure]}]]
+   (let [formdata (doto
+                   (js/FormData.)
+                    (.append constants/ipfs-add-param-name value))]
+     {:http-raw-post (merge {:url  constants/ipfs-add-url
+                             :body formdata
+                             :success-event-creator
+                             (fn [{:keys [status body]}]
+                               (if (= 200 status)
+                                 (on-success {:value (parse-ipfs-add-response body)})
+                                 (when on-failure
+                                   (on-failure {:value status}))))}
+                            (when on-failure
+                              {:failure-event-creator on-failure})
+                            {:timeout-ms 5000})})))
 
 (re-frame/reg-event-fx
  :http/post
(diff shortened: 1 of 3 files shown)


tbenr:
the usage is:

```
[ipfs/add {:value "pippo"
             :on-success [store-result]}]
```

the on-success event gets:

```
:value
 :Name "QmXiafTBk6asBoRPxks3idfNpw5DuBSxGuYmPkeBQePY5o"
 :Hash "QmXiafTBk6asBoRPxks3idfNpw5DuBSxGuYmPkeBQePY5o"
 :Size "13"
```

@jeluard  the implementation should support binary data but currently i don't have a clear idea on how to test it.

jeluard:
Looks great! Can we make sure the keywords returned in the event are lowercased?

I guess the binary support can’t be easily checked for now. Sounds good!

tbenr:
@jeluard what about this solution? :)
i applied the lowercase transformation to all result parsing. Don't know if it is too invasive...

src/status_im/extensions/core.cljs
```diff
@@ -207,6 +209,23 @@
                            {:failure-event-creator on-failure})
                          {:timeout-ms 5000})}))
 
+(re-frame/reg-event-fx
+ :ipfs/add
+ (fn [_ [_ _ {:keys [value on-success on-failure]}]]
+   (let [formdata (doto
+                   (js/FormData.)
+                    (.append "extension.event.edn" value))]
```
jeluard: I am wondering if this is worth making this configurable? Is it used at all?

src/status_im/extensions/core.cljs
```diff
@@ -157,7 +158,8 @@
 
 (defn- parse-json [o]
   (when o
-    (js->clj (js/JSON.parse o) :keywordize-keys true)))
+    (utils/transform-keys-to-lower-case
```
jeluard: An alternative is to just manually recreate the map using lower-case keys. The upside is that you control what ends up in the resulting map (no surprises).  And no need to introduce the new function (less code).
Are we talking about a lot of keys here?

src/status_im/extensions/core.cljs
tbenr: as per my testing, it doesn't make any difference. I can move it to costants.

src/status_im/extensions/core.cljs
tbenr: ok I agree, let's avoid surprises and manage keys directly.

src/status_im/extensions/core.cljs
tbenr: dont know if your intention was to make it an optional parameter (defaulting value to a costant)

jeluard: Ok not important for now I guess :)

src/status_im/extensions/core.cljs
```diff
@@ -207,6 +209,28 @@
...
+      (select-keys [:Name :Hash :Size])
+      (set/rename-keys {:Name :name :Hash :hash :Size :size})))
+
+(re-frame/reg-event-fx
+ :ipfs/add
+ (fn [_ [_ _ {:keys [value on-success on-failure]}]]
+   (let [formdata (doto
+                   (js/FormData.)
+                    (.append constants/ipfs-add-param-name value))]
+     {:http-post (merge {:url  constants/ipfs-add-url
```
jeluard: Does this work fine with this one? Don't we need to introduce a raw version?

tbenr: no, it works nicely with the non-raw version! :)

tbenr: btw, I didn't add `:http-raw-post`, so `http/post` event still doesn't work..

src/status_im/extensions/core.cljs
```diff
@@ -207,6 +209,28 @@
                            {:failure-event-creator on-failure})
                          {:timeout-ms 5000})}))
 
+(defn- parse-ipfs-add-response [res]
+  (-> (parse-json res)
+      (select-keys [:Name :Hash :Size])
+      (set/rename-keys {:Name :name :Hash :hash :Size :size})))
```
jeluard: It would probably be simpler / more efficient to just destructure key values and reconstruct inline a map wth the right key names.

e.g.
````clojure
(let {{:keys [Name]} (parse-json res)}
  {:name Name})
````

src/status_im/extensions/core.cljs
jeluard: Do you mind adding it? `http/raw-post` already exist so it should be trivial.
Also it does less thing and should provide a leaner answer.

src/status_im/extensions/core.cljs
tbenr: 👍

src/status_im/extensions/core.cljs
tbenr: 👍

jeluard (approved):

flexsurfer (approved):

jeluard:
@flexsurfer Could you merge this one?

PR #7021, opened 2018-12-06
[#6956] delete duplicates after iteration over messages/statuses

rasom:
if an entity is deleted during iteration "index out of range"-like error happens
> I get `Cannot read property 'message-id' of undefined`

status: ready


cammellos (approved):

pedropombeiro (approved):
Works for me finally.

PR #7032, opened 2018-12-06
[#6952] fix odd app behavior after quick logout

### LINKED ISSUE #6952: Odd app behavior after quick logout
# Problem

When user logs out quickly after login it seems like a lot of things that should be happening after login keep happening after logout.
The most visible result of that is that if the user types his password in the login screen oftentimes he gets logged in without tapping `log in`.
I have also experience blank screen, crashes and error displayed related to `app-db` being `nil`

# Implementation

Make sure delayed events after login aren't affecting the app after quick logout

# Acceptance Criteria

Successions of quick logins and logout shouldn't trigger any of the problems mentioned above

# Notes

The cause of the auto-login seems to be the `node.ready` signal from status-go that callback `accounts.login/login`. The condition being that a password is present in the db (hence the need for the user to type his password but not tapping login). The guard here is insufficient.

# Future Steps

This issue raises again the problem of delayed events that shouldn't happen off-session. If the fix includes a general solution to avoid this class of bug that would be great.

rasom:
fix #6952 

## Node instance management ##

Currently the node instance with "dummy" configs is used to allow 
`CreateAccount` and `RecoverAccount` calls, but there is no point in doing this 
as node can be started right before these calls. Also we restart node on logging
out which causes weird behaviour like described here #6952.

**How it worked in `develop` before this PR**:
1. `node` is started on application startup.
   `PFSEnabled=false,NoDiscovery=true` are used, so it is started only to make
   possible `CreateAccount` and `RecoverAccount` calls, because they require
   running node atm.
2. When the user **presses "Sign in" button**, the node is stopped, then started
   with user specific configs (`InstallationID`, custom bootnodes, etc), and
   only after that `Login` call is performed.
3. When the user **creates** a new account, at first `CreateAccount` call happens,
   and after that node is stopped and started as in **1.**.
4. When the user **restores** their account, at first `RecoverAccount` call happens,
   and after that node is stopped and started as in **1.**.
5. When the user **logs out** the node is stopped and started again as in **1.** to
   make possible CreateAccount`,`RecoverAccount` calls.

**How it works in this PR**:
1. `node` is not started on the app startup
2. When the user **presses "Sign in" button** the node is started
   with user specific configs (`InstallationID`, custom bootnodes, etc), and
   only after that `Login` call is performed.
3. When the user **creates** a new account, at first the node is started with
   default params (the same as would be used when user signs into the app after
   account creation whithout changing any setting), then `CreateAccount` call
   happens, then `Login`.
4. When the user **restores** their account, the flow is the same as **3.** but
   with `RecoverAccount` instead of `CreateAccount`
5. When the user **logs out** the node is stopped. That's it.

status: ready

### DIFF
--- a/src/status_im/accounts/create/core.cljs
+++ b/src/status_im/accounts/create/core.cljs
@@ -19,7 +19,8 @@
             [status-im.utils.signing-phrase.core :as signing-phrase]
             [status-im.utils.types :as types]
             [taoensso.timbre :as log]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.node.core :as node]))
 
 (defn get-signing-phrase [cofx]
   (assoc cofx :signing-phrase (signing-phrase/generate)))
@@ -33,10 +34,17 @@
    #(re-frame/dispatch [:accounts.create.callback/create-account-success (types/json->clj %) password])))
 
 ;;;; Handlers
-
-(defn create-account [{{:accounts/keys [create] :as db} :db}]
-  {:db (update db :accounts/create assoc :step :account-creating :error nil)
-   :accounts.create/create-account (:password create)})
+(defn create-account
+  [{:keys [db random-guid-generator] :as   cofx}]
+  (fx/merge
+   cofx
+   {:db (-> db
+            (update :accounts/create assoc
+                    :step :account-creating
+                    :error nil)
+            (assoc :node/on-ready :create-account
+                   :accounts/new-installation-id (random-guid-generator)))}
+   (node/initialize nil)))
 
 (fx/defn add-account
   "Takes db and new account, creates map of effects describing adding account to database and realm"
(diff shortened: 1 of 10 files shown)


goranjovic (approved):

mandrigin (approved):

yenda (approved):
Thanks!

flexsurfer (approved):

asemiankevich:
@rasom `null is not an object (evaluating 'e.objects')` shown when creating account

image

rasom:
@asemiankevich should be fixed now, rebase mess

asemiankevich:
@rasom `Log out` button is disabled for a while when changing accounts ( i have 1 new and 1 recovered) , so every time i am switching between them - this button seems to be disabled for several seconds (2-3 seconds visually). Is it expected?

image

rasom:
yes @asemiankevich, it is expected, atm the best we can do

asemiankevich:
> yes @asemiankevich, it is expected, atm the best we can do

okay thanks!

PR #7033, opened 2018-12-06
[#6788] Implement ERC20/ERC721 Events

### LINKED ISSUE #6788: Implement ERC20/ERC721 events
### User Story

As a developer, I want to reuse existing extensions [events](https://status.im/extensions/reference_events.html) for ERC20 / ERC721

### Description

*Type*: Feature

Most common ERC20/ERC721 contracts methods should be exposed are generic and reusable extensions [events](https://status.im/extensions/reference_events.html):

ERC20:
* totalSupply
* balanceOf
* transfer
* transferFrom
* approve
* allowance

ERC721:
* ownerOf
* safeTransferFrom
* setApprovalForAll
* getApproved
* isApprovedForAll

tbenr:
fixes #6788

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -556,6 +556,81 @@
                 {:permissions [:read]
                  :value       :extensions/ethereum-cancel-filter
                  :arguments   {:id  :string}}
+                'ethereum.erc20/total-supply
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-total-supply
+                 :arguments   {:contract     :string
+                               :on-result    :event}}
+                'ethereum.erc20/balance-of
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-balance-of
+                 :arguments   {:contract     :string
+                               :token-owner  :string
+                               :on-result    :event}}
+                'ethereum.erc20/transfer
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-transfer
+                 :arguments   {:contract  :string
+                               :to        :string
+                               :value     :number
+                               :on-result :event}}
+                'ethereum.erc20/transfer-from
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-transfer-from
+                 :arguments   {:contract   :string
+                               :from      :string
+                               :to        :string
+                               :value     :number
+                               :on-result :event}}
+                'ethereum.erc20/approve
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-approve
+                 :arguments   {:contract    :string
+                               :spender     :string
+                               :value       :number
+                               :on-result   :event}}
+                'ethereum.erc20/allowance
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc20-allowance
+                 :arguments   {:contract     :string
+                               :token-owner  :string
+                               :spender      :string
+                               :on-result    :event}}
+                'ethereum.erc721/owner-of
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc721-owner-of
+                 :arguments   {:contract    :string
+                               :token-id    :string
+                               :on-result   :event}}
+                'ethereum.erc721/is-approved-for-all
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc721-is-approved-for-all
+                 :arguments   {:contract   :string
+                               :owner      :string
+                               :operator   :string
+                               :on-result  :event}}
+                'ethereum.erc721/get-approved
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc721-get-approved
+                 :arguments   {:contract   :string
+                               :token-id   :string
+                               :on-result  :event}}
+                'ethereum.erc721/set-approval-for-all
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc721-set-approval-for-all
+                 :arguments   {:contract   :string
+                               :operator   :string
+                               :approved   :boolean
+                               :on-result  :event}}
+                'ethereum.erc721/safe-transfer-from
+                {:permissions [:read]
+                 :value       :extensions/ethereum-erc721-safe-transfer-from
+                 :arguments   {:contract   :string
+                               :from       :string
+                               :to         :string
+                               :token-id   :string
+                               :data?      :string
+                               :on-result  :event}}
                 'ethereum/call
                 {:permissions [:read]
                  :value       :extensions/ethereum-call
--- a/src/status_im/extensions/ethereum.cljs
+++ b/src/status_im/extensions/ethereum.cljs
@@ -101,6 +101,114 @@
  (fn [{db :db} [_ _ {:keys [to] :as arguments}]]
    (wrap-with-resolution db arguments :to execute-ethcall)))
 
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-total-supply
+ (fn [{db :db} [_ _ {:keys [contract on-result]}]]
+   (let [json-rpc-args {:to contract
+                        :method "totalSupply()"
+                        :outputs ["uint256"]
+                        :on-result on-result}]
+     (wrap-with-resolution db json-rpc-args :to execute-ethcall))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-balance-of
+ (fn [{db :db} [_ _ {:keys [contract token-owner on-result]}]]
+   (let [json-rpc-args {:to contract
+                        :method "balanceOf(address)"
+                        :params [token-owner]
+                        :outputs ["uint256"]
+                        :on-result on-result}]
+     (wrap-with-resolution db json-rpc-args :to execute-ethcall))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-allowance
+ (fn [{db :db} [_ _ {:keys [contract token-owner spender on-result]}]]
+   (let [json-rpc-args {:to contract
+                        :method "allowance(address,address)"
+                        :params [token-owner spender]
+                        :outputs ["uint256"]
+                        :on-result on-result}]
+     (wrap-with-resolution db json-rpc-args :to execute-ethcall))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-transfer
+ (fn [{db :db} [_ _ {:keys [contract to value on-result]}]]
+   (let [json-rpc-args {:to contract
+                        :method "transfer(address,uint256)"
+                        :params [to value]
+                        :on-result on-result}]
+     (wrap-with-resolution db json-rpc-args :to execute-send-transaction))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-transfer-from
+ (fn [{db :db} [_ _ {:keys [contract from to value on-result]}]]
+   (let [json-rpc-args {:to contract
+                        :method "transferFrom(address,address,uint256)"
+                        :params [from to value]
+                        :on-result on-result}]
+     (wrap-with-resolution db json-rpc-args :to execute-send-transaction))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-approve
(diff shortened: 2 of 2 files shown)


ghost:
### Pull Request Checklist

src/status_im/extensions/core.cljs
```diff
@@ -493,6 +493,47 @@
                  :value       :extensions/ethereum-resolve-ens
                  :arguments   {:name      :string
                                :on-result :event}}
+                'ethereum/ERC20-totalSupply
+                {:permissions [:read]
```
jeluard, line 560: How about `ethereum.erc20/totql-supply` ? The same pattern could be applied to others.

src/status_im/extensions/core.cljs
```diff
@@ -493,6 +493,47 @@
...
+                 :value       :extensions/ethereum-ERC20-totalSupply
+                 :arguments   {:address        :string
+                               :on-result :event}}
+                'ethereum/ERC20-balanceOf
+                {:permissions [:read]
+                 :value       :extensions/ethereum-ERC20-balanceOf
+                 :arguments   {:address         :string
+                               :tokenOwner :string
```
jeluard: Or `token-owner` ? In general clojure style naming is preferred.

src/status_im/extensions/core.cljs
```diff
@@ -493,6 +493,47 @@
...
+                               :tokens    :number
+                               :on-result :event}}
+
+                'ethereum/ERC20-transferFrom
+                {:permissions [:read]
+                 :value       :extensions/ethereum-ERC20-transferFrom
+                 :arguments   {:address   :string
+                               :from      :string
+                               :to        :string
+                               :tokens    :number
```
jeluard: Or `value`?

src/status_im/extensions/core.cljs
```diff
@@ -493,6 +493,47 @@
...
+                 :value       :extensions/ethereum-ERC20-balanceOf
+                 :arguments   {:address         :string
+                               :tokenOwner :string
+                               :on-result  :event}}
+                'ethereum/ERC20-transfer
+                {:permissions [:read]
+                 :value       :extensions/ethereum-ERC20-transfer
+                 :arguments   {:address   :string
+                               :to        :string
+                               :tokens    :number
```
jeluard: Or `value` ?

src/status_im/extensions/core.cljs
tbenr: much better!

tbenr:
@jeluard i have a couple of questions:
1. what about changing `ethereum/call` signature renaming `to` parameter to `address` (contract address) so there is coherence with erc20 (in this context `to` is a parameter to the smartcontract method)
2. when dealing with values (tokens quantity), do you think we should expose the `ufixed` version of it (we have to read contract behind the scene to get decimals and convert uint256 accordingly). This impacts input (parameters) and output (return values)

jeluard:
@tbenr 

1) Yes I already done that in a local branch. So looks good!!
2) Everything wei seems like the simple approach? What is ufixed exactly?

tbenr:
@jeluard 

1. ok, done :)
2. I mean this: when calling

```
[ethereum.erc20/balance-of {:address "0x103262f243E6f67d12D6a4EA0d45302C1FA4BB0a"
                              :token-owner "0xe21af62e34855df858992f93ac1df3de37c30c85"
                              :on-result [store-result]}]
```
[etherscan](https://ropsten.etherscan.io/token/0x103262f243e6f67d12d6a4ea0d45302c1fa4bb0a?a=0xe21af62e34855df858992f93ac1df3de37c30c85)

I should get `197435.5933946745849944` but i'm actually getting `1.9743559339467458e+23` because we are not considering the fixed point (this token has 18 decimals). `ufixed` is the type in abi-spec that do the job (as far as i can see).

jeluard:
@tbenr  Let's keep it as is, the decimals are token dependent thus require per token operation.

So this is not WIP anymore, right ? :)

jeluard:
@tbenr Do you have a basic extension we could use to test?

tbenr:
@jeluard OK! - still WIP, i'm currently adding erc721 methods. when finish I'll provide some examples

tbenr:
@jeluard now I:
1. included erc721 events.
2. correctly send transaction when needed.

BTW I think there is something to review on the UI side: the user is not really aware of what he's signing.

I wasn't able to test all events by the way. in particular for the ERC721.

here some examples:

````
[ethereum.erc20/balance-of {:address "0x103262f243E6f67d12D6a4EA0d45302C1FA4BB0a"
                          :token-owner "0xe21af62e34855df858992f93ac1df3de37c30c85"
                          :on-result [store-result]}]


[ethereum.erc20/transfer-from {:address "0x2e7cd05f437eb256f363417fd8f920e2efa77540"
                             :from "0x8443b668be105143466bcdb8d8fce1f697762ed3"
                             :to "0x5adf1b9e1fa4bd4889fecd598b45079045d98f0e"
                             :value 10000000000
                             :on-result [store-result]}]

[ethereum.erc20/transfer {:address "0x2e7cd05f437eb256f363417fd8f920e2efa77540"
                        :to "0x5adf1b9e1fa4bd4889fecd598b45079045d98f0e"
                        :value 10000000000
                        :on-result [store-result]}]

[ethereum.erc721/owner-of {:address "0xa2c8b4b3116f05b7250cd007b2314a5e2dc9cdda"
                         :token-id "2731"
                         :on-result [store-result]}]

[ethereum.erc721/owner-of {:address "0xdda364c2c8ead6e8d7a87a56f97d51f64fb5f4ff"
                         :token-id "1761480239665022021014448466483078144295760728322270224953624097718321"
                         :on-result [store-result]}]

[ethereum.erc721/owner-of {:address "0xa2c8b4b3116f05b7250cd007b2314a5e2dc9cdda"
                         :token-id "0xaab"
                         :on-result [store-result]}]

[ethereum.erc721/safe-transfer-from {:address "0xd776dcef9ba3da8e317f5bc1d4d37f6bceb51dcd"
                                   :from "0x263e13a485ae463ad6679dac6e1d359a05423cd8"
                                   :to "0x5adf1b9e1fa4bd4889fecd598b45079045d98f0e"
                                   :token-id "34"
                                   :on-result [store-result]}]
````

**Comment by jeluard** [comment]
> I wasn't able to test all events by the way. in particular for the ERC721.

@tbenr Do you have more details about that?

**Comment by tbenr** [comment]
@jeluard i didn't find a faucet in testnet to get an NFT and then run transaction. Do you have any suggestion to get a erc721 "sandbox"?

**Comment by jeluard** [comment]
@tbenr Cool I guess that's fine for now! More interested in your other bounties :)

**Comment by tbenr** [comment]
@jeluard BTW this [transaction](https://ropsten.etherscan.io/tx/0x9938c6ad0e34c2c1bac89b090b8b768714578a329336fefd22760954abcbf5b5) was made by me using the `safe-transfer-from` event and looks good to me (i was not the owner of the NFT so that's why it was supposed to fail). But the gas estimation was not working correctly (it gave me `nil` don't know why, I put gas manually). Gas estimation was ok for ERC20 transfers.

**Review by jeluard — APPROVED** [review]

File: src/status_im/extensions/ethereum.cljs [hunk 5]
```diff
@@ -98,8 +98,119 @@
...
+     (wrap-with-resolution db arguments' :to execute-send-transaction))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-transfer-from
+ (fn [{db :db} [_ _ {:keys [address from to value] :as arguments}]]
+   (let [arguments' (merge arguments
+                           {:to address
+                            :method "transferFrom(address,address,uint256)"
+                            :params [from to value]
+                            :value nil})]
```
jeluard: Do we need those `value nil` ?

src/status_im/extensions/ethereum.cljs
```diff
@@ -98,8 +98,119 @@
...
+                            :method "getApproved(uint256)"
+                            :params [token-id]
+                            :outputs ["address"]})]
+     (wrap-with-resolution db arguments' :address execute-ethcall))))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc721-set-approval-for-all
+ (fn [{db :db} [_ _ {:keys [address to approved] :as arguments}]]
+   (let [arguments' (merge arguments
+                           {:to address
```
jeluard: What is the rational of `to` vs `address` ?

src/status_im/extensions/ethereum.cljs
```diff
@@ -98,8 +98,119 @@
...
- (fn [{db :db} [_ _ {:keys [to] :as arguments}]]
-   (wrap-with-resolution db arguments :to execute-ethcall)))
+ (fn [{db :db} [_ _ {:keys [address] :as arguments}]]
+   (wrap-with-resolution db arguments :address execute-ethcall)))
+
+(handlers/register-handler-fx
+ :extensions/ethereum-erc20-total-supply
+ (fn [{db :db} [_ _ {:keys [address] :as arguments}]]
+   (let [arguments' (merge arguments
+                           {:address address
```
flexsurfer: why do we need to merge the address field to itself?

flexsurfer (approved):

src/status_im/extensions/ethereum.cljs
tbenr: In `execute-send-transaction` didn't changed the `:to` into `:address` like I did with `eth_call`. if it is ok for you i'll change all, so `ethereum/send-transaction` parameters will change too, ok?

jeluard: Why do we want one vs the other?
Ultimately if it's the same info, should be the same property name?
Also we might want to stick with what JSON-RPC uses

src/status_im/extensions/ethereum.cljs
tbenr: you're right. but to avoid confusion and clearly decouple event arguments from internal web3 call arguments, if you agree I create a new map for the internal web3 explicitly. or maybe using `select-keys`.

src/status_im/extensions/ethereum.cljs
tbenr: yes, because `value` for `send-transaction` for an ETH transfer, here we use it as a parameter in a smartcontract method call. as said @flexsurfer to avoid confusion i make a new map here.

src/status_im/extensions/ethereum.cljs
tbenr: in my opinion the confusion here is that we are "projecting" two parameters map (JSON-RPC and smartcontract mehod params) in one single parameters map (extension event params).

JSON-RPC has a `:to` param as the smarcontract address you want to interact with. but if you want to call erc20 `transfer` method you have another `:to` parameter which is the destination address of the tokens. So we changed the first `:to` in `:address` to indicate the smart contract address. similarly we have such collision with :value (wei on JSON-RCP vs erc20 token tranfser amount)

tbenr: so my suggestion was to clearly build a new map for JSON-RCP to avoid inadvertently to pass smartcontracr method params to JSON-RPC interface.

tbenr:
```
(handlers/register-handler-fx
 :extensions/ethereum-erc721-set-approval-for-all
 (fn [{db :db} [_ _ {:keys [address to approved] :as arguments}]]
   (let [json-rpc-args {:to address
                                     :method "setApprovalForAll(address,bool)"
                                     :params [to approved]}]
     (wrap-with-resolution db json-rpc-args :to execute-send-transaction))))
```

src/status_im/extensions/ethereum.cljs
jeluard: Not following 😅  So just omitting value would not work?

src/status_im/extensions/ethereum.cljs
jeluard: Makes sense. I would suggest to stick with contract parameter names. And find a unique name for the contract address. Could be `contract` ? or anything unique we could reuse for similar use cases.

tbenr:
@jeluard @flexsurfer I updated as discussed. Thus, `ethereum-send-transaction` and `ethereum-call` events are tied to json-rpc args while erc20\erc721 are based on contract method args.

PR #7058, opened 2018-12-10
Show schema version details on migrations failure

rasom:
Simplifies debugging of migration failures, specifically in cases when
more than one migration is applied.

status: ready


mandrigin (approved):
useful!

goranjovic (approved):

PR #7071, opened 2018-12-11
set BUILD_TYPE to pr as default

jakubgs:
This is in order to make possible running separate platforms as separate PR builds for GitHub.
Otherwise we get:
```
java.lang.IllegalArgumentException: Null value not allowed as an environment variable: BUILD_TYPE
```


mandrigin (approved):
lgtm

flexsurfer (approved):

PR #7072, opened 2018-12-11
move upload step top platform specific builds, cleanup

jakubgs:
This is in preparation to switch to building different platforms separately.

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -3,6 +3,7 @@ pipeline {
 
   options {
     timestamps()
+    disableConcurrentBuilds()
     /* Prevent Jenkins jobs from running forever */
     timeout(time: 35, unit: 'MINUTES')
     /* Limit builds retained */
@@ -14,9 +15,9 @@ pipeline {
   }
 
   parameters {
-    booleanParam(
+    string(
       name: 'BUILD_TYPE',
-      description: 'Specify build type. Values: pr / nightly / release',
+      description: 'Specify build type. Values: pr / e2e / nightly / release',
       defaultValue: 'pr',
     )
   }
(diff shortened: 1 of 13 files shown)


ghost:
### Pull Request Checklist

mandrigin (approved):
lgtm

ci/Jenkinsfile.android
```diff
@@ -90,4 +92,7 @@ pipeline {
       }
     }
   }
+  post {
+    failure { script { cmn.gitHubNotifyPRFail() } }
```
pedropombeiro: Does this ever get called when we're not building for a PR branch (i.e. `env.CHANGE_ID` doesn't have a value)?

pedropombeiro (approved):
LGTM, added a question for clarification

ci/Jenkinsfile.android
jakubgs: Good catch, yeah, I just caught that, so I added it as an argument the `buildBranch` method. Check it out.

rasom (approved):

ci/Jenkinsfile.linux
```diff
@@ -24,6 +24,7 @@ pipeline {
 
   options {
     timestamps()
+    disableConcurrentBuilds()
```
mandrigin, line 27: why is that needed?

ci/Jenkinsfile.ios
```diff
@@ -69,12 +70,13 @@ pipeline {
...
     stage('Archive') {
       steps {
-        script { archiveArtifacts api }
+        archiveArtifacts api
       }
     }
     stage('Upload') {
       steps {
         script {
+          env.PKG_URL = cmn.uploadArtifact(api)
```
mandrigin: we should upload to diawi instead

jakubgs: I can drop it. Sure.

ci/Jenkinsfile.linux
mandrigin: okay, it is only blocking jobs from the same PR, resolving

jakubgs: The idea is that it affects each PR separately(since each PR has it's own build generated), which means there are no concurrent builds for the same platforms for the same pr.

The idea is to not let someone hog all the builders by pushing a lot of commits at once to a PR.

mandrigin (approved):

oskarth:
I see now why we are the most active ERC20 project :D

jakubgs:
>I see now why we are the most active ERC20 project :D

Yeah, trying to polish the new build setup for GitHub, generates a lot of spam.
Later we can optimize this by writing a more complex script to post just one comment and keep updating it.

PR #7073, opened 2018-12-12
Fetch history action

cammellos:
Adds a entry in the menu to fetch the last 24 hours (default) for a given topic (discovery for public chats, sym key for public).

Also uses cursor to fetch from mailservers, which will partially fix the current issues we are facing.

A cursor will ask for much less data (currently 200 messages) and will paginate over it until the end is reached.

What is still do to (not in the scope of this PR):
- Currently if the request fails it will again start from the first 200 messages, it will not pick up from where it has left
- Possibly we would want to give up and move on if we can't fetch some data (try max N times), rather than starting from scratch next time

### Testing

The behaviour should be similar to what is currently happening (long "Fetching messages  banner"), but unlike develop, eventually it will go away in this PR and history will be fetched.

status: ready


src/status_im/mailserver/core.cljs
```diff
@@ -446,23 +453,32 @@
...
+                    (process-next-messages-request)))))))
 
 (fx/defn retry-next-messages-request
   [{:keys [db] :as cofx}]
   (fx/merge cofx
             {:db (dissoc db :mailserver/request-error)}
             (process-next-messages-request)))
 
+;; At some point we should update `last-request`, as eventually we want to move
+;; on, rather then keep asking for the same data, say after n amounts of attempts
```
mandrigin, line 454: I'm not sure if that commit is the right place, but there is this issue based on the previous postmortem: https://github.com/status-im/status-react/issues/7083

mandrigin (approved):
👍

src/status_im/mailserver/core.cljs
cammellos: Yes, I have added cursor to this PR as adding a request last  24 hours makes it easier to test, the commit partially solves the issue (we fetch much less data with each request , max 200 messages), but it's still a all-or-nothing approach (if for whatever reason a request fails it will start from scratch), that I can take care separately, I want to validate cursors are working as expected first.

rasom (approved):

churik:
@cammellos some of messages are definitely lost when you get back from offline mode.
In this particular case I closed app and sent messages to public channel from another device to check offline inboxing.
Two messages, sent at 15:44 are lost (they are successfully fetched on another devices)
Others are fetched.
Couldn't find exact steps to reproduce, but have geth.log.
untr
[geth 15.48.17.log](https://github.com/status-im/status-react/files/2676523/geth.15.48.17.log)
[Status 15.48.22.log](https://github.com/status-im/status-react/files/2676694/Status.15.48.22.log)

Also I tried install this build and use personal account - can't get all messages in 1-1 and public chats, some of them are still missing.

UPDATE:
as it turned out, my local device time was changed to ~25 sec (to check one issue) a while ago.
So I set automatic time, relaunch app and try to Fetch history for chats with missing messages.
Nothing new was fetched, so I still  have missing messages on desktop.

UPDATE 2:
I changed mail-02.ac-cn-hongkong-c.eth.beta server to another and all messages were fetched. checking with @cammellos what can be wrong here.

churik:
@cammellos Can you please rebase it to avoid error with `verify` module error?

churik:
@cammellos 
1) now I constantly can reproduce  an issue when messages are not fetched from offline mode on desktop when user is on mailserver `mail-02.ac-cn-hongkong-c.eth.beta`
*Prerequisites:* user has build installed on desktop platform (in my case it is Mac OSx) and mail server is `mail-02.ac-cn-hongkong-c.eth.beta`
*Steps:*
- join any public chat (i.e. #12345test)
- log out 
- sent some messages to the same chat from another device
- log in on desktop
- check this chat - no new history fetched
- click on "Fetch history for 24 hours" - no new history fetched
- go to settings and change mailserver (in my case I changed to `mail-01.do-ams3.eth.beta`)

After this history is finally fetched.
**Reproducible only on desktop.**
**Reproducible on nightly as well.**

Video: http://take.ms/ERSKV
[geth.log](https://github.com/status-im/status-react/files/2679490/geth.log)
[Status.log](https://github.com/status-im/status-react/files/2679459/Status.log)

2) Cannot  check fuctionality with fetching history, because it is fetched automatically and for case 1 it doesn't help to fetch history, so I don't know it is working as expected or not.

3) Photo/username sync is not working on my setup (group and 1-1 chats are syncing correctly) - also on nightly, so can be logged as separate issue.


So to sum up:
- 1 and 3 should be created separately?
- do you know some steps to test "Fetching messages" to see the result?

I didn't find any regression issues, and constant "Fetching messages" disappeared yesterday, so I assume that PR is ready to merge after autotests are passed.

churik:
Tested (IOS(IPhone 7), Mac OSx, Android 8.1 (Genymotion)):
- fetching messages for last 24h when automatilcally cjoina chat
- fetching messages for last 24h manually (when change date on device)
- offline inboxing (separate #7098 created)
- messaging in 1-1, group, public chat
- message sync after device pairing.

PR #7081, opened 2018-12-12
Adjust from value to accomodate for late-comers

### LINKED ISSUE #6814: Public chats are missing messages on different devices
# Problem

Public chats are missing messages, that are not recovered even after multiple history requests.

I suspect that the problem is in the way we query history from mail servers. As I understood from the discussion in Status.im we do it in a following way: If at first I query envelopes from 13:10:11 till 13:15:17 and query was successful, next time i will start my query from 13:15:17 till some timestamp. The problem is that mail server can receive a message with timestamp lower than 13:15:17 after we sent back a history for 1st query.

# Implementation

Mail server uses (Expiry timestamp - TTL) as a timestamp for the message. Additionally there is small time window that is tolerated by any whisper node, afair it is 10s. I would suggest to adjust query with a common TTL and mentioned whisper window. 

It will increase rate of duplicates, but we will stop missing messages from mail server.

# Acceptance Criteria

Messages that are received by mail server after history was queried should be sent to the user on a subsequent queries.

link: https://github.com/status-im/status-react/issues/6757

cammellos:
We add a bunch of seconds to the `from` field when requesting from the mailserver to accomodate for late-comers. We still cap to one day max to avoid errors from mailserver.

same as https://github.com/status-im/status-react/pull/6988 but had to re-create as I forced pushed :(

fixes: #6814

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -11,6 +11,7 @@
             [status-im.utils.utils :as utils]
             [taoensso.timbre :as log]
             [status-im.transport.db :as transport.db]
+            [status-im.transport.message.protocol :as protocol]
             [clojure.string :as string]
             [status-im.data-store.mailservers :as data-store.mailservers]
             [status-im.i18n :as i18n]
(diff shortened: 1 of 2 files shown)


mandrigin (approved):
yay! awesome!

pedropombeiro (approved):
lgtm

yenda (approved):

cammellos:
@churik  same as https://github.com/status-im/status-react/pull/6988 , ready to be tested, basically if you join a new public chat you won't see any difference, but next requests should all have 20s more than current develop.

churik:
Can't login with existing account.

upload
Status.log: [Status.log](https://github.com/status-im/status-react/files/2675617/Status.log)
Geth.log: [geth.log](https://github.com/status-im/status-react/files/2675618/geth.log)

But I can log in if I log in to another accout -> log out -> login again.
Cannot reproduce constantly, but definitely saw this several times in this exact PR.

Other cases checked:
- offline/online and fetching messages after back online
- fetching messages for 24 h

If you can, please take a look at it, and I'll continue testing when issues with mailservers will be resolved.

cammellos:
@churik that's an issue already in development, #status-desktop , Vitaliy is taking a look at it, I will rebase once the fix is ready.

rasom (approved):

cammellos:
@churik this is ready to be tested again, thanks!

churik:
Tested on Mac OSx and IOS:
- fetching 24h history 
- offline inboxing for 1-1, group, public chats
- chatting with contact from release build
- chatting with user from different timezone
- chatting with user with custom time
- resend message
- devices sync

PR #7085, opened 2018-12-13
Dont choke on wrongly serialized messages

cammellos:
Discard messages that cannot be deserialized correctly, we should also not store those messages, but this is definitely needed at the moment.

I can look at why it does not serialize correctly, but I think I will move to higher priority for now as this fixes the immediate issue.
 
status: ready

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -3,7 +3,6 @@
             [status-im.accounts.db :as accounts.db]
             [status-im.chat.commands.core :as commands]
             [status-im.chat.models :as chat-model]
-            [status-im.constants :as constants]
             [status-im.data-store.user-statuses :as user-statuses-store]
             [status-im.utils.datetime :as time]
             [status-im.utils.fx :as fx]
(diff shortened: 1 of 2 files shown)


mandrigin (approved):
lgtm!

annadanchenko:
On upgraded app: first message in 1:1 chat is duplicated each time new message arrives in the chat. 



Can't reproduce same issue if upgrade 0.9.32 with latest nightly (ios) and send normal chars
Steps:
1. install 0.9.32 and add a contact. Send "hi" and receive "hello" in 1:1 chat
2. upgrade to PR build
3. launch app, send normal text message, e.g. "after upgrade". Check that messages on the top of the chat are shown as usual
4. kill the app and launch again
5. open same 1:1 chat and send problem char (see https://github.com/status-im/security-reports/issues/1 for details)
6. kill the app and launch again. Check how messages are shown in 1:1 chat
7. send new normal text message in 1:1 chat Check messages on top of the chat, now there is issue that same message is shown twice and duplicated again if send more messages

flexsurfer (approved):

annadanchenko:
I've done tests on macos, win, android and ios for upgrade to reproduce the issue found in previous PR. looks like it's fixed. Will continue testing on fresh install and upgrade with cases to check problem char in
1. 1:1, group, public chats messages (received and sent)
2. received when app is not running (PN in 1:1 chat)
3.  in Reply
4. in password
5. in username
6. in Group name
7. Recover acc who got problem char in 1:1 chat
8. Sync on devices when got problem char in one of them

churik:
On Linux Ubuntu tested:
- joining to public message where message with problem char is posted
- sending/receiving messages
- joining to other several chats
- log in/log out

annadanchenko:
@cammellos let's merge and probably make a hotfix for 0.9.32 mobile release with this fix only cc @mandrigin If we will have release build this weekend - I'll test it

PR #7099, opened 2018-12-14
[slow sign in] adding symkeys and filters in batch 

rasom:
Chat's symmetric keys should be restored after signing in in order to be
used during filters creation. Currently, if there are **N** chats we restore
**N** keys one by one:
1. `ssh.addSymKey` is called
2. `:status-im.transport.core/sym-key-added` event is dispatched
   on success
3. filter is created via `ssh.newMessageFilter`
4. `:shh.callback/filter-added` event is dispatched if filter was
   successfully created

In result we have **2N** `re-frame` events which is expensive in case if we have
a lot of chats.

This PR changes the way keys are restored:
1. batch of `ssh.addSymKey` requests for all sym keys is sent at once
2. `:status-im.transport.core/sym-keys-added` event is dispatched with results
   of all successful `ssh.addSymKey` calls
3. filter is created via `ssh.newMessageFilter`
4. `:shh.callback/filters-added` event is dispatched with all added filters
   as a parameter
5. profit

In ideal case only 2 `re-frame` events are dispatched.

**Possible cons:**
Batch size is not limited atm, probably better to split it into smaller 
chunks if there are, let's say, 1000+ chats. 

**Results**
adding filters on galaxy s9(`develop vs PR`):
100 chats: 3.3s vs 0.5s
200 chats: 6s vs 0.6s

on a slower android device:
50 chats: 5.62s vs 1.07s

status: ready

### DIFF
--- a/src/status_im/transport/filters.cljs
+++ b/src/status_im/transport/filters.cljs
@@ -18,7 +18,7 @@
   (log/debug :stop-watching filter))
 
 (defn add-filter!
-  [web3 {:keys [topics to] :as options} callback chat-id]
+  [web3 {:keys [topics] :as options} callback chat-id]
   (let [options  (assoc options :allowP2P true)]
     (log/debug :add-filter options)
     (when-let [filter (.newMessageFilter (utils/shh web3)
@@ -26,6 +26,24 @@
                                          callback
                                          #(log/warn :add-filter-error (.stringify js/JSON (clj->js options)) %))]
       (re-frame/dispatch [:shh.callback/filter-added (first topics) chat-id filter]))))
+
+(defn add-filters!
+  [web3 filters]
+  (log/debug "PERF" :add-filters (first filters))
+  (re-frame/dispatch
+   [:shh.callback/filters-added
+    (keep
+     (fn [{:keys [options callback chat-id]}]
+       (when-let [filter (.newMessageFilter
+                          (utils/shh web3)
+                          (clj->js (assoc options :allowP2P true))
+                          callback
+                          #(log/warn :add-filter-error
+                                     (.stringify js/JSON (clj->js options)) %))]
+         {:topic   (first (:topics options))
+          :chat-id chat-id
+          :filter  filter}))
+     filters)]))
 
 (re-frame/reg-fx
  :shh/add-filter
@@ -35,6 +53,24 @@
          callback (fn [js-error js-message]
                     (re-frame/dispatch [:transport/messages-received js-error js-message chat-id]))]
      (add-filter! web3 params callback chat-id))))
+
+(re-frame/reg-fx
+ :shh/add-filters
+ (fn [{:keys [web3 filters]}]
+   (log/debug "PERF" :shh/add-filters)
+   (let [filters
+         (reduce
+          (fn [acc {:keys [sym-key-id topic chat-id]}]
+            (conj acc
+                  {:options  {:topics   [topic]
+                              :symKeyID sym-key-id}
+                   :callback (fn [js-error js-message]
+                               (re-frame/dispatch [:transport/messages-received
+                                                   js-error js-message chat-id]))
+                   :chat-id  chat-id}))
+          []
+          filters)]
+     (add-filters! web3 filters))))
 
 (re-frame/reg-fx
  :shh/add-discovery-filter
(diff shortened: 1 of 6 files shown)


yenda:
@rasom I wonder if that wouldn't be better now that there is a DB on go side to persist the keys there and avoiding these interactions altogether. @cammellos

rasom:
@yenda it might be better, though these changes don't prevent from moving this part of logic to go side later

src/status_im/transport/filters.cljs
```diff
@@ -53,13 +89,35 @@
 (handlers/register-handler-fx
  :shh.callback/filter-added
  (fn [{:keys [db] :as cofx} [_ topic chat-id filter]]
+   (log/debug :shh.callback/filter-added (.now js/Date))
```
mandrigin: looks like a leftover

src/status_im/transport/filters.cljs
```diff
@@ -53,13 +89,35 @@
...
              (mailserver/process-next-messages-request))))
 
+(fx/defn add-filter
+  [{:keys [db]} chat-id filter]
+  {:db (assoc-in db [:transport/filters chat-id] filter)})
+
+(handlers/register-handler-fx
+ :shh.callback/filters-added
+ (fn [cofx [_ filters]]
+   (log/debug :shh.callback/filters-added (.now js/Date))
```
mandrigin: looks like a leftover

src/status_im/transport/filters.cljs
```diff
@@ -34,6 +52,24 @@
                     (re-frame/dispatch [:transport/messages-received js-error js-message chat-id]))]
      (add-filter! web3 params callback chat-id))))
 
+(re-frame/reg-fx
+ :shh/add-filters
+ (fn [{:keys [web3 filters]}]
+   (log/debug :shh/add-filters (first filters))
```
mandrigin: looks like a leftover

src/status_im/transport/shh.cljs
```diff
@@ -151,6 +151,32 @@
...
+                   (swap! counter inc)
+                   (if err
+                     (on-error err)
+                     (swap! results conj {:chat-id    chat-id
+                                          :sym-key    sym-key
+                                          :sym-key-id resp}))
+                   (when (= @counter total)
+                     (log/debug :success :add-sym-key-batch)
+                     (on-success @results)))]
+    (log/debug :add-sym-key-batch total)
```
mandrigin: leftover

mandrigin (changes requested):
please, clean up debug logging, otherwise, lgtm

src/status_im/transport/shh.cljs
rasom: it is a valid debug statement, i would prefer to keep it

src/status_im/transport/filters.cljs
rasom: the problem is that we do not log effects and it's hard to figure when it is executed. `log/debug` will not produce logs in release build, so it should be safe to use

rasom:
@mandrigin i would prefer to keep that debug logging

cammellos (approved):
Looks ok, probably next time let's have a discussion before jumping on this, a simpler implementation would just have been to add a method to status-go that takes a batch, so we don't have to deal with aggregating multiple callbacks in status-react.

mandrigin:
@rasom then can you add some tag to them, so one can filter by them? maybe `PERF` or something like that?

pedropombeiro (approved):

rasom:
@mandrigin fixed

mandrigin (changes requested):
hmm, I still see a few log/debug statements added there with no tag

rasom:
@mandrigin should be fixed now

mandrigin (approved):

churik:
Tested on Mac OSx:
1) fresh install
- restoring account - password and seed phrase valigation;
- restoring account - first login - quick logout;
- creating several accounts (quickly and slowly on all steps);
- password and username validation during creating account;
- disabling\enabling logging quickly after log in;
- mailservers switching;
- reopening app when you logged out and when you logged in
2) upgrade (from [release 13](https://ci.status.im/job/status-react/job/release/job/release%2Fdesktop%2F0.9.0/13/) )
- log in with existing account (~ 50 different chats)
- quick log out
- disabling\enabling logs
- can't login with wrong password

asemiankevich:
looks good to me on mobile

Issue #7097, opened 2018-12-14
Messages not loaded from the database if out-of-order messages are received


cammellos:
Some messages are not loaded from the database if messages with an older timestamp are received (from the mailserver for example).

### Step to replicate
The steps to  replicate are a bit artificial, this is the simplest way, but fetching old messages from the mailserver might also result in this behavior, or receiving a message from a peer with the clock set to the wrong time.

1) Post 100 messages (1,2,3,4,5...) in a public chat
2) Make sure all those messages are on device A
3) Change your time to -1 day
4) Login/Logout
5) Go to that chat, do not scroll up
6) Post a few messages (5)
7) Scroll up, there should be some gaps in messages

Login/logout should fix the issue and those messages should appear again.

### Details

This is due to the fact that we use the count of messages loaded to paginate, but the collection underneath (and our count as well) change with the receiving of messages, which leads to unstable pagination.

A possible solution is to use a cursor `clock-value-message-id` and use that to paginate (no skipping, you always fetch the last x messages that have a cursor value > last-clock-value fetched). This might lead to empty pages (pages full of duplicates) if old messages are received and the request will have to be retried, but at least it won't be resulting in skipped messages (which has an impact on the user, unlike empty pages that can be retried).

churik:
@cammellos yeah, this is way to reproduce the issue what @rasom told recently.
Basically on this case I got:
-  messages are shown as sent several times  after scrolling down;
- these messages are shown at the end of other chats when I'm switching between them;
- after  setting automatic date and login\logout issue messages are shown as sent yesterday you can switch between chats normally.
Logs:
[Status_r.log](https://github.com/status-im/status-react/files/2680759/Status_r.log)
[geth.log](https://github.com/status-im/status-react/files/2680761/geth.log)

<img width="568" alt="messages_old" src="https://user-images.githubusercontent.com/4557972/50013206-d65ca680-ffc0-11e8-9a4d-55ca2f81d9af.png">
<img width="568" alt="older" src="https://user-images.githubusercontent.com/4557972/50013453-76b2cb00-ffc1-11e8-877d-d3c8b06b3e87.png">

PR #7111, opened 2018-12-17
Paginate using clock-value & message-id instead of skip/limit

cammellos:
Fixes: #7097 

Paginating using the count of loaded messages might result in some
messages being skipped and not being loaded in the database, in case of
out-of-order messages received.

This commit changes the behavior to sort by `clock-value` and
`message-id`, which gives a consistent sorting.

The initial idea was to use a cursor `clock-value-message-id` and
iterate on that, but realm does not support filtering on string (</>),
so instead we keep track of messages with identical clock-value and
exclude those in the next page query.

The change might result in pages that have duplicates (so messages needs
to be deduped), but won't result in skipped messages.

For a good description of the issue: https://coderwall.com/p/lkcaag/pagination-you-re-probably-doing-it-wrong

This is also using `clock-value` instead of `timestamp`, as that's the one true order, although messages gets grouped by timestamp and then sorted by clock-value.

### Testing

1) Create 0..100 messages in chat A
2) Logout
3) Set you phone date to yesterday
4) Login again, open that chat, don't scroll
5) Publish a few messages
6) Scroll up, check 0..100 are all there

In develop you will see some gaps
In this PR there should not be any 

status: ready

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -111,9 +111,10 @@
     get-stored-user-statuses :get-stored-user-statuses
     get-referenced-messages :get-referenced-messages :as cofx}]
   (when-not (get-in db [:chats current-chat-id :all-loaded?])
-    (let [loaded-count               (count (get-in db [:chats current-chat-id :messages]))
+    (let [previous-pagination-info   (get-in db [:chats current-chat-id :pagination-info])
           {:keys [messages
-                  all-loaded?]}      (get-stored-messages current-chat-id loaded-count)
+                  pagination-info
+                  all-loaded?]}      (get-stored-messages current-chat-id previous-pagination-info)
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

mandrigin:
linter is angry at you 😛
otherwise, looks good!

mandrigin:
@cammellos merge conflicts ☹️

rasom (approved):

churik:
@cammellos can you please rebase it?

cammellos:
@churik rebased, thanks!

churik:
Tested on IOS and on Mac OS - issue is not reproduced (I could reproduce it on `develop`)
@cammellos can you please resolve the conflicts?

cammellos:
@churik rebased, thanks!

PR #7112, opened 2018-12-17
Limit number of participants in group chat to 10

cammellos:
A few things:

- Limit the number of participants to 10 in group chats.
- Add invites, so that users will only receive messages if accepted the invitation.- 
- Move group chats to a separate topic, falling back to discovery for older clients.
- Handle multi-envelopes confirmations

### Testing

When creating a new group chats, users will be presented with a Join chat button, and won't be receiving messages until the click on Join chat.

If running a previous version (0.9.32) group chats will work just the same (there's no option to join, but they will receive messages).

On upgrade from 0.9.32 existing group chats will have to be "joined" before receiving messages, giving the stage of development of group chats (experimental), I feel this is acceptable, but  I could fix otherwise.

status: ready


src/status_im/ui/components/checkbox/view.cljs
```diff
@@ -3,13 +3,16 @@
...
-    (if platform/android?
+    (if (or platform/android?
+            platform/desktop?)
       [react/view (merge styles/wrapper style)
        [react/check-box {:on-value-change     on-value-change
                          :value               checked?
+                         :disabled            (and (not checked?)
+                                                   disabled?)
                          :accessibility-label accessibility-label}]]
       [react/touchable-highlight (merge {:style               (merge styles/wrapper style)
```
mandrigin, line 17: hmm, shouldn't we add this `disabled?` flag to iOS too?

mandrigin (approved):

src/status_im/ui/components/checkbox/view.cljs
cammellos: I will have a look actually :)

src/status_im/group_chats/core.cljs
```diff
@@ -399,6 +465,27 @@
      (map #(assoc % :from from) events))
    all-updates))
 
+(defn get-inviter-pk [my-public-key {:keys [membership-updates] :as chat}]
```
rasom: `chat` is not used

src/status_im/transport/message/core.cljs
```diff
@@ -57,6 +57,19 @@
     {:db            (assoc-in db [:transport/chats chat-id :resend?] nil)
      :data-store/tx [(transport-store/save-transport-tx {:chat-id chat-id
                                                          :chat    updated-chat})]}))
+(fx/defn check-confirmations [{:keys [db] :as cofx} status chat-id message-id]
+  (when-let [{:keys [pending-confirmations not-sent]} (get-in db [:transport/message-ids->confirmations message-id])]
+    (if (= 0 (dec pending-confirmations))
```
rasom: `zero?`

src/status_im/transport/message/core.cljs
```diff
@@ -57,6 +57,19 @@
     {:db            (assoc-in db [:transport/chats chat-id :resend?] nil)
      :data-store/tx [(transport-store/save-transport-tx {:chat-id chat-id
                                                          :chat    updated-chat})]}))
+(fx/defn check-confirmations [{:keys [db] :as cofx} status chat-id message-id]
+  (when-let [{:keys [pending-confirmations not-sent]} (get-in db [:transport/message-ids->confirmations message-id])]
```
rasom: 80 chars limit

rasom (approved):

goranjovic (approved):

asemiankevich:
@cammellos for some reason i dont see 7112 branch in https://ci.status.im/job/status-react/job/prs/job/android/view/change-requests/ (android), could you please check?

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/prs/job/android/view/change-requests/">Pull Requests (17) [status-react » prs » android] [Jenkins]</a></strong></div></blockquote>

cammellos:
@asemiankevich should be building now, thanks!

asemiankevich:
@errorists maybe you could take a look at this from design point of view? I suggest move this `You can select 9 more participants` to this nice placeholder in the middle of the screen as a minimum^ also apply the fonts and colors that we are using across the app. 

Anyway your input here is highly appreciated!

image

there are some more screens down here

image

image

errorists:
@asemiankevich sure, I wasn't aware we're limiting the max number of users but on the design side I'd like us to follow WhatsApp here with a design like this 

screenshot 2019-01-07 at 14 13 45

The designs are in [this Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=231%3A0)

cammellos:
@asemiankevich @errorists thanks, I am happy to implement it, but in a separate PR if that's ok

asemiankevich:
> @asemiankevich @errorists thanks, I am happy to implement it, but in a separate PR if that's ok

@cammellos absolutely fine! thanks Andrea

asemiankevich:
1. Designs will be implemented later in separate PR. Designs are [here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=231%3A0)
2. Questions I have so far:
- no option to decline the invite. Only Join button. @errorists maybe we can think about something here? Not needed? Any user action if I dont want to join this group i was invited to?
- no group chat shown if user was re-invited to the chat after he has deleted it on his device. Steps:

- [ ] create group chat by device A (admin) with device B (non-admin)
- [ ] as device B accept the invite
- [ ] as device B delete the chat. Chat is removed from Home screen. System message in a group chat states that user B left the group (can be checked on device A).
- [ ] as device A re-add user B to the same group chat
- [ ] as device B check that nothing appears on home screen (no group chat shown). **This is same as in develop.**
- [ ] as device A send some messages to this group chat
- [ ] as device B check that you still dont see anything. **This seems to be wrong and in develop user B will see group chat and messages again.**

- some warning shown when i am on Home screen, not sure if this is related to this PR though. 
image
- as a user that was invited to the group chat i am already listed as a member even i did not touch the  `Join group` button. Also, admin could make me admin and then i can add / remove members even i did not join the group yet. Do we want to limit any user functionality prior he taps the `Join` button? @errorists also question to you.

image

- as admin of chat i create with release version  i see `Join chat` button when upgraded from released version to this PR. Also some random inviter is indicated (this is a member of the chat though). When i tap Join -> i see this system message about joining.

image

image

- not clear what is expected when i upgrade a chat which has more than 10 members? No errors for now. 
- `add members` is not hidden if i upgrade the chat which has more than 10 participants already. I can't add more though. @errorists any input?

cammellos:
> no group chat shown if user was re-invited to the chat after he has deleted it on his device

that I will fix

>  some warning shown when i am on Home screen, not sure if this is related to this PR though.

I think was already present, it won't be present on production, but I will take a look at some point

> add members is not hidden if i upgrade the chat which has more than 10 participants already.

I will fix

The rest seems to be mainly ux issues, and if it's ok, I would address in separate prs.

Thanks!

cammellos:
@asemiankevich I have addressed the 2 points (warning is still present). Thanks!

asemiankevich:
@cammellos thanks checking now! meanwhile what about this ?

> as admin of chat i create with release version i see Join chat button when upgraded from released version to this PR. Also some random inviter is indicated (this is a member of the chat though). When i tap Join -> i see this system message about joining.

Quite confusing to me to see this Join chat with random inviter when i am already a chat creator (after upgrade).

Swader:
Is there a specific reason for this limitation? Why a limit at all?

cammellos:
@Swader currently performance, group chats use a stronger encryption and it's more expensive as more users join. 
Currently the focus is to make sure it works correctly, if we don't limit the number of users it's hard to understand which problems are due to performance degradation and which are due to the mechanics.
As we get more feedback we can then work on improving performance and increasing the limit (or removing it if it becomes feasible of course).

errorists:
And what about the 'Join Chat' gateway, is there a specific reason we're adding this now?

cammellos:
@errorists yes, partially based on user feedback (people complained about being added to chats by people that they have not added in cotacts), partially is something we discussed with grame previously (the model was wechat), and also there are some technical reasons (having a confirmation that the user wants to receive messages from this group chats helps with some performance optimization).

Swader:
@cammellos I see, thanks. In that case a message clarifying this limit at the invite screen would be useful from a UX perspective (@hesterbruikman), but it needs to be worded correctly. People's train of thought will be: "Wtf, Telegram is encrypted and I can have dozens of thousands of members in a group". Perhaps the message could contain a link which leads to an explanatory blog post.

cammellos:
@Swader yes fair point, happy to add something on those lines.
For the specific example, telegram group chats are not end-to-end encrypted (only client-server, if the server is compromised all the conversation can be decrypted), and that's why they can have a large number of users.

the encryption we use is similar to signal's, where the group is unlimited, but performance degradation happens on large groups. What's app uses a similar method and limits to 256 (which is clearly much more than 10, but also we are not whatsapp yet ;) )

Swader:
I see. This is all something the average user won't know, and makes perfect content for the blog post explainer.

errorists:
@cammellos thanks for explanation. I went ahead and did a design for the 'Join' prompt, I've also added an option to decline the invitation in which case the group chat should be removed from your Home Screen. I'm not sure if we should disclose the number of members inside the group until you've joined, I figure maybe not. [Figma link](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=1741%3A0) Please keep me in the loop if you're doing this and the design for capping max group members as a separate PR

screenshot 2019-01-08 at 11 24 54

On the subject of informing user about the decisions we made and limitations set on them. I honestly don't believe anyone would care enough to read through a blog post :) It is a fluid limitation, one that we can stress test, I'm totally fine with not telling it's due to performance limitations of the protocol.

asemiankevich:
@errorists thanks !! Exactly what i was asking for!

cammellos:
@errorists thanks, it will be in a separate PR, I will keep you in the loop!

Swader:
@errorists you're probably right IRT majority of users. I'm just a curious minority :)

errorists:
Also to answer @asemiankevich  previous questions. 
I would leave the Add members button visible, but once inside use the same mechanism to inform of the total user number and if it's reached the limitation, exactly the same as when you're creating a new group.
Users should not be visible on the group member list until they've joined and similarly should not be counted against the total users limit.

cammellos:
> I would leave the Add members button visible, but once inside use the same mechanism to inform of the total user number and if it's reached the limitation, exactly the same as when you're creating a new group.

I can revert the change if @asemiankevich is ok (possibly in a different pr, if this tests ok soon)

> Users should not be visible on the group member list until they've joined
 
That we can do, although an admin should see all the invitations I take?

>  and similarly should not be counted against the total users limit.

This is tricky, essentially because if I understand correctly, it means I could "over-invite" a chat (send 15 invitations, and only when 10 have joined prevent to send invitations).

If that is the desired behavior it adds quite a few technical issues (I can go into details, but basically it might end up with having a chat over-joined, and having to exclude members, and we need to make sure all the participants have a consistent view of the chat etc), so I'd rather keep it as it is, unless we feel strongly about this point.

errorists:
> I can revert the change if @asemiankevich is ok (possibly in a different pr, if this tests ok soon)

different PR along with the change in[ how the total number of users is shown](https://github.com/status-im/status-react/pull/7112#issuecomment-451932992) would be great, together these will make sense :) All of those designs are updated and reflected in the [Group Chat Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=227%3A320)

> That we can do, although an admin should see all the invitations I take?

Good point, yes admin should see all invitations.

> This is tricky, essentially because if I understand correctly, it means I could "over-invite" a chat (send 15 invitations, and only when 10 have joined prevent to send invitations).

I thought it would be much simpler and overlook the details, it can be left as is then

asemiankevich:
@cammellos yess the suggestion from @errorists works to me

hesterbruikman:
Such rapid progress here! Great work!

PR #7136, opened 2018-12-18
[TESTS] Workaround on wrong transition to home view instead of chat

antdanchenko:
This workaround is added in order to bypass the behavior from #5975 and to merge #6869 asap


oskarth (approved):
Not familiar enough with test setup but lgtm

goranjovic (approved):

PR #7137, opened 2018-12-18
Add send logs command

### LINKED ISSUE #6710: Standardized debug artifact generation

### User Story

As a developer, I want to work with standardized debug/logging artifacts so that I don't miss any vital information when debugging an issue, and can trust that all required information is present in this standardized artifact.

As a user, I want to have a 1-click solution to provide debugging information to the app developer so that I'm not required to go hunt for logs on the file system and know that the developer won't be missing any important information.

### Description

*Type*: Feature

*Summary*: Currently gathering logs to report a bug is a non-trivial task even for Core contributors. The issues with it are:
- The two relevant logs, status-react log (`Status.log`) and the status-go log (`geth.log`) are located in separate folders, which differ per platform (we currently support 5 platforms). `Status.log` is only generated on desktop platforms, for mobile we currently rely on developer tools which won't be available to end users;
- The user needs to remember to fetch the two of them and not just one;
- The user needs to remember to set the appropriate log level before collecting the logs;
- Having a safe export of the app db could be useful to debug certain issues (e.g. it could tell us how many peers a device is connected to, which mailserver it is connected to, etc.);
- The user probably wants to zip the files so he can easily attach them on an email, which is something that can/should be automated.

We can do much better by automating most of these steps for the end-user.

#### Expected behavior

1. User notices buggy behavior in app and decides he wants to report it;
1. User enables logging in the app (ideally we have a single setting which would take care of that, but that's for a later GHI).
1. When user is finished reproducing bug, he/she clicks on `Send logs` button and the app will:
  1. Fetch logs/system information/P2P status and zip the files;
  1. Share this file using the OS share dialog on mobile (which allows sharing to Gmail, etc), or by opening a file explorer on the folder location (on Desktop).

pedropombeiro:

fixes #6710 

### Summary:

This PR adds one-click log export for Android, Windows, Linux, and MacOS. iOS is not implemented as I don't have the hardware required for it, but it will be just a question of implementing the `sendLogs` React method.

Sample zip file: [Status-debug-logs.zip](https://github.com/status-im/status-react/files/2702608/Status-debug-logs.zip)

Android UI:
screenshot_20181221-123931

Exported `db.json`:

image

An error message is shown in case free space is below 20 MB:

image

### Review notes (optional):

I've tested Android, Windows and Linux, so it's particularly important to test MacOS. Let me know if there's more info you'd like to be added to the zip file that would be useful for future debugging.

#### Platforms
- Android
- iOS (the new Send logs button should simply not show in iOS)
- macOS
- Linux
- Windows

### Steps to test:
- Open Status
- Sign in
- Go to Profile/Advanced
- Enable Development Mode
- Click Send Logs
- A Share operation should be started on mobile. A explorer window should be opened with the zip on Desktop.

### Next steps:

Remove https://status.im/build_status/status_go.html#Android-debugging section, as it won't be possible to access geth.log from `adb` anymore.

status: ready


vitvly (approved):

pedropombeiro:
@siphiuel that was (too?) quick! Just want to make sure you noticed the changes to Qt/Desktop part, e.g. replacing `RCTStatus.getRootDirPath()` with `extern QString getDataStoragePath()`?

vitvly:
@PombeirP yes, recall reading this. getDataStoragePath() got there because it had to rely on STATUS_DATA_DIR env var. When reading also checked whether `geth.log` path was correct. Could there be any issue?

pedropombeiro:
@siphiuel I don't think there's an issue, since the `getDataStoragePath()` also relies on `STATUS_DATA_DIR`, so it seemed like we didn't need another method that calculated the same path (`RCTStatus.getRootDirPath()`) but wanted to confirm that I wasn't missing something.

goranjovic (approved):

Serhy:
Hey @PombeirP there are two questions so far appeared, while we testing this with @asemiankevich:

1) To keep geth.log on Mobile (Android) we had to enable Storage permission for app. Otherwise no geth.log log is kept. I’m not sure exactly where from, but `Send logs` option gathers geth.log log even if Storage permission was not enabled. That’s not the case on MacOS, though: if `Logging enabled?` = False -> no geth.log stored -> no geth.log to send using `Send logs` Doesn’t it violates security in such case of Android option?
To reproduce on Android simply install fresh Status build -> create new account -> use ‘Send logs’ to verify whether geth.log is present in .zip file.

2) With mobile app (Android so far) we re-write Status.log file each time the app starts. In a case of app crashes, ‘Send logs’ is not very useful in that case since we re-write the data. Logcat always availabel actually, but I was wondering whether we can think in scope of this PR to cover such case: 
- for example, Desktop app, when app crashes - offers an option to share logs before app is shut down. Something similar would be nice to have for Mobile app. 
- Or if that’s possible - to set some flag for Status.log to `append` logs next time the app starts (instead of re-write), if app crashed before.

pedropombeiro:
> Hey @PombeirP there are two questions so far appeared, while we testing this with @asemiankevich:

Hey @Serhy. Regarding 1., I think the two options are (cc @annadanchenko @mandrigin):
- 1.1. Leaving things as they were in terms of paths and therefore requiring user to give storage permissions. I initially changed things because it was failing for me and I realize I had forgotten about giving Storage permissions to `Status Debug` app :-)
  - Cons: This has the inconvenient that `geth.log` will not be available to us if user didn't initially give the permission and then wants to send logs to us for some reason.
- 1.2. Moving `geth.log` to the app directory (as is currently implemented in PR)
  - Pros: No permissions needed
  - Cons: Devs won't be able to connect to log using `adb shell tail -f sdcard/Download/geth.log`

I guess your point with `Logging enabled?` setting (which isn't available on mobile) is that if we follow 1.2 approach, we'll be collecting logs without user consent. In this scenario, we could make `Logging enabled?` setting also available for mobile, or just choose approach 1.1.

Regarding 2. right now I'm filtering `adb logcat` with the current process pid. It would be very easy to filter `adb logcat` output on facility name (e.g. `StatusModule`, `ReactNativeJS`) instead of pid.

pedropombeiro:
@Serhy the next build will implement 1.1 approach and filter logcat by facility so you can test and see what you think. Should be available in ~15 minutes.

asemiankevich:
> @Serhy the next build will implement 1.1 approach and filter logcat by facility so you can test and see what you think. Should be available in ~15 minutes.

@PombeirP no changes in jenkins for android, is it okay? https://ci.status.im/job/status-react/job/prs/job/android/view/change-requests/job/PR-7137/

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/prs/job/android/view/change-requests/job/PR-7137/">PR-7137 [status-react » prs » android] [Jenkins]</a></strong></div></blockquote>

asemiankevich:
@PombeirP @Serhy so i am now not able to send logs (none of them) if storage permission is not enabled on Android. The following error dialog is shown when storage permission is not enabled:
image

Also some more thoughts here:

- i am forced to restart the app to be able to send logs if i enable storage permission after the error dialog. Can we have forced log out here then (since it is required)?
- can we send Status.im and db logs if storage permission is NOT enabled? So zip file will contain only 2 files (no geth log if permissions are not granted)
- can we modify the error message and put something more meaningful to user, because now it is saying about insufficient disk space which is confusing and not true?
- regarding iOS - since we dont have this functionality for iOS , can we hide this `send logs` button then? Until we implement it

pedropombeiro:
@asemiankevich I just saw where that error is coming from. I was looking at the free space on the same directory as `geth.log`, instead of the output zip file. Since in the latest commit, `geth.log` is in Downloads, it would be dependent on the permissions. I've fixed that.

> * can we send Status.im and db logs if storage permission is NOT enabled? So zip file will contain only 2 files (no geth log if permissions are not granted)

We already do that.

> * regarding iOS - since we dont have this functionality for iOS , can we hide this `send logs` button then? Until we implement it

Are you sure it is visible on iOS? I don't have iOS device to test with, but I had added code to hide it: https://github.com/status-im/status-react/blob/feature/%236710-send-logs/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs#L147

asemiankevich:
@PombeirP 
> Are you sure it is visible on iOS? I don't have iOS device to test with, but I had added code to hide it: https://github.com/status-im/status-react/blob/feature/%236710-send-logs/src/status_im/ui/screens/desktop/main/tabs/profile/views.cljs#L147

yeah seems so, installed latest iOS build from this PR.

image


pedropombeiro:
@asemiankevich OK, this build should fix it on iOS.

asemiankevich:
> @asemiankevich OK, this build should fix it on iOS.

@PombeirP `send logs` button is now hidden on iOS.

however the error i stated for insufficient space is still in place on android (storage permissions are not enabled)

Serhy:
Just to add to previous comment https://github.com/status-im/status-react/pull/7137#issuecomment-451435342 that it's confirmed on my side with Android 6.0.1 and Android 7 with and without Storage permission disregard the app upgraded from Release0.9.32/Delelop0.9.33 or fresh install (https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190104-085801-81f8c4-pr.apk)

With Created/Recovered account in-app and enabled Storage permissions -> `Send Logs` responds with `Insufficient space` error.

<blockquote></blockquote>

pedropombeiro:
@Serhy the new build should fix that behavior. Turns out that if you don't pre-create the file, `GetUsableSpace` will return 0.

asemiankevich:
> @Serhy the new build should fix that behavior. Turns out that if you don't pre-create the file, `GetUsableSpace` will return 0.

I can confirm it is fixed.
- if i dont enable permissions for storage - sending logs archive contains Status.log and db output
- if permissions are granted - then geth.log is included in archive

PR #7146, opened 2018-12-19
Fix generation of AppImage to point to Status instead of reportApp

### LINKED ISSUE #7128: Linux build can't start
# Problem

After mark build as executable I still can't open file.
Last nightly build, that I can start  is [nightly 15/01/18](https://ci.status.im/job/status-react/job/nightly/853/)
Now I don't see any new processes are launching when I tried to start build.

Video: http://take.ms/0Ummm

<blockquote><div>Monosnap screenshot tool</div><div><strong><a href="https://monosnap.com/file/Rh3bYmyV2luTQjZsgbQ5La9qVFc66G">File "screencast 2018-12-18 11-14-37.mp4"</a></strong></div><div>Monosnap — the best tool to take & share your screenshots.</div></blockquote>

pedropombeiro:
Fixes #7128

### Summary:

A recent change to `linuxdeployqt-continuous-x86_64.AppImage` changed the way that the tool generates the artifacts. This is complicated by the fact that we were calling the tool once for `reportApp` and another for `Status`, so the symlink inside the AppImage was pointing to `reportApp` instead of `Status`. This PR changes the script to use our own uploads of the tools so we can have reproducible builds, and simplifies the script so that a single call to linuxdeployqt is required to handle both executables (resorting to the `-executable` flag).

#### Platforms
- Linux

status: ready

### DIFF
--- a/scripts/build-desktop.sh
+++ b/scripts/build-desktop.sh
@@ -78,8 +78,10 @@ function joinExistingPath() {
 
 STATUSREACTPATH="$(cd "$SCRIPTPATH" && cd '..' && pwd)"
 WORKFOLDER="$(joinExistingPath "$STATUSREACTPATH" 'StatusImPackage')"
-DEPLOYQT="$(joinPath . 'linuxdeployqt-continuous-x86_64.AppImage')"
-APPIMAGETOOL="$(joinPath . 'appimagetool-x86_64.AppImage')"
+DEPLOYQTFNAME='linuxdeployqt-continuous-x86_64_20181215.AppImage'
+APPIMAGETOOLFNAME='appimagetool-x86_64_20181109.AppImage'
+DEPLOYQT=$(joinPath . "$DEPLOYQTFNAME")
+APPIMAGETOOL=$(joinPath . "$APPIMAGETOOLFNAME")
 STATUSIM_APPIMAGE_ARCHIVE="StatusImAppImage_20181208.zip"
 
 function init() {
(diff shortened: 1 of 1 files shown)


ghost:
### Pull Request Checklist

cammellos (approved):

vitvly (approved):

rasom (approved):

pedropombeiro:
Tried logging out from this new build and got a crash:
```
qrc:/main.qml:13:1: module "QtQuick.Layouts" is not installed
```

Will need to check if this is related to the script changes, but it shouldn't be as I compared the AppImage generated by old code and new code and only change was the symlink.

vitvly:
I think that the error message `module "QtQuick.Layouts" is not installed` is a consequence of a crash, because this module is imported by `reportApp` which itself is launched after the crash. So agree that it's not because of the script changes. Could be that it can be fixed by https://github.com/status-im/status-go/pull/1329

jakubgs (approved):

pedropombeiro:
Fixed the deployment issue mentioned above, but still see a segmentation fault error when logging out of the app. I have the impression this is due to new behavior because if I build a commit from last Friday with the same script, it doesn't crash. So most likely not related with this issue.

[crash.zip](https://github.com/status-im/status-react/files/2696819/crash.zip)

rasom:
@PombeirP crash on logout is fixed here https://github.com/status-im/status-react/pull/7144, not related to this PR

churik:
thanks @PombeirP !
ready to merge to develop branch and to `release/desktop/0.9.0`

vitvly:
Merged to both.

PR #7148, opened 2018-12-19
[#6594] Add local chats contacts data in chat command payloads

### LINKED ISSUE #6594: Add local chats contacts data in chat command payloads
### User Story

As a developer, I want to have access to a contacts available in current chat. This data will be part of the previews and on-send/on-receive events.
In 1-1 chat, get access to the recipient details via `:contact`
In private group-chats, get access to all group members via `:contacts`
In public chat, does nothing.

### Description

*Type*: Feature

tbenr:
fixes #6594

### DIFF
--- a/src/status_im/chat/commands/core.cljs
+++ b/src/status_im/chat/commands/core.cljs
@@ -34,6 +34,20 @@
   [type]
   (keyword (str (protocol/id type) "-button")))
 
+(defn add-chat-contacts
+  "Enrich command-message by adding contact list of the current private or group chat"
+  [contacts {:keys [public? group-chat] :as command-message}]
+  (cond
+    public? command-message
+    group-chat (assoc command-message :contacts (map status-im.contact.db/public-key->address contacts))
+    :else (assoc command-message :contact (status-im.contact.db/public-key->address (first contacts)))))
+
+(defn enrich-command-message-for-events
+  "adds new pairs to command-message to be consumed by extension events"
+  [db {:keys [chat-id] :as command-message}]
+  (let [{:keys [contacts public? group-chat]} (get-in db [:chats chat-id])]
+    (add-chat-contacts contacts (assoc command-message :public? public? :group-chat group-chat))))
+
 (defn generate-short-preview
   "Returns short preview for command"
   [{:keys [type]} command-message]
--- a/src/status_im/chat/commands/receiving.cljs
+++ b/src/status_im/chat/commands/receiving.cljs
@@ -1,5 +1,6 @@
 (ns status-im.chat.commands.receiving
   (:require [status-im.chat.commands.protocol :as protocol]
+            [status-im.chat.commands.core :as commands]
             [status-im.utils.fx :as fx]))
 
 (defn lookup-command-by-ref
@@ -14,7 +15,7 @@
   [{:keys [db] :as cofx} message]
   (let [id->command (:id->command db)]
     (when-let [{:keys [type]} (lookup-command-by-ref message id->command)]
-      (protocol/on-receive type message cofx))))
+      (protocol/on-receive type (commands/enrich-command-message-for-events db message) cofx))))
 
 (defn enhance-receive-parameters
   "Enhances parameters for the received command message.
--- a/src/status_im/chat/commands/sending.cljs
+++ b/src/status_im/chat/commands/sending.cljs
@@ -37,15 +37,15 @@
         ;; no yield control, proceed with sending the command message
         (let [command-message (create-command-message chat-id type parameter-map cofx)]
           (fx/merge cofx
-                    #(protocol/on-send type command-message %)
+                    #(protocol/on-send type (commands/enrich-command-message-for-events db command-message) %)
                     (commands.input/set-command-reference nil)
                     (chat.message/send-message command-message)))))))
 
 (fx/defn send
   "Sends command with given parameters in particular chat"
-  [cofx chat-id {:keys [type]} parameter-map]
+  [{:keys [db] :as cofx} chat-id {:keys [type]} parameter-map]
   (let [command-message (create-command-message chat-id type parameter-map cofx)]
     (fx/merge cofx
-              #(protocol/on-send type command-message %)
+              #(protocol/on-send type (commands/enrich-command-message-for-events db command-message) %)
               (commands.input/set-command-reference nil)
               (chat.message/send-message command-message))))
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -32,7 +32,8 @@
 
 (defview message-content-command
   [command-message]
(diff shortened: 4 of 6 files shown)


tbenr:
@jeluard this works but I'm not happy with the solution. Too many places where I added `add-chat-contacts`, but I haven't found a more elegant solution so far.
Btw this adds `{:contact <pkey>}` to 1-1 chats and `{:contacts (<pkey1> <pkey2> ...)}` to group chats for `on-send`, `on-receive`, `preview` and `short-preview`

ops, I left `:group-chats` in the scope for command hook spec. Do I need to remove it?

jeluard:
@tbenr Actually I think a better approach would be to modify the payload where it is generated in the first place:

* for [on-send / on-receive](https://github.com/status-im/status-react/blob/develop/src/status_im/chat/commands/sending.cljs#L10) 
* for [short-preview](https://github.com/status-im/status-react/blob/develop/src/status_im/ui/screens/home/views/inner_item.cljs#L24)
* for [preview](https://github.com/status-im/status-react/blob/develop/src/status_im/ui/screens/chat/message/message.cljs#L46)

Does that make sense?

tbenr:
@jeluard yep, I got confused going around the code ;-)

tbenr:
@jeluard I have a doubt, is it better to add it here [on-send](https://github.com/status-im/status-react/blob/e40e495e118171459db9520bef6ff0382cf9fa1d/src/status_im/chat/commands/sending.cljs#L40)? adding [here](https://github.com/status-im/status-react/blob/e40e495e118171459db9520bef6ff0382cf9fa1d/src/status_im/chat/commands/sending.cljs#L10) we will send it out on the network adding bandwidth overhead?

jeluard:
Good catch! We definitely don't want to send that over the network.

tbenr:
@jeluard I left [:group-chats](https://github.com/status-im/status-react/blob/f45e18550e104ec58f04fb8d9a8ed0557a58c404/src/status_im/chat/commands/core.cljs#L131) scope. Is it OK?

btw: some examples:

```
 views/preview
 (let [{group-chat-contacts :contacts one-on-one-contact :contact} properties]
   [view {:style {:padding 5}}
    [text "contacts ${group-chat-contacts} contact: ${one-on-one-contact}"]])

 events/on-receive
 (let [{one-on-one-contact :contact group-chat-contacts :contacts} properties]
   [store/put {:key "testk" :value "text ${one-on-one-contact} ${group-chat-contacts}"}])

```

jeluard:
> I left :group-chats scope. Is it OK?
Sure no worries!

Sorry I didn't make that clear but using `@(re-frame.core/subscribe)` here is problematic. The good news is that now you can directly manipulate the full app-db and access the relevant data (so same params as the `get-in` ones).

tbenr:
@jeluard you mean like

```
(get-in @re-frame.db/app-db [:chats chat-id :contacts])
```

?

jeluard:
No, directly relying on some var pointing to the db. e.g. you have it [here](https://github.com/status-im/status-react/pull/7148/files#diff-39c7abe084dcc7d5fc35ae2c0839f73cR15)

tbenr:
@jeluard I'm struggling to find a way to get db in case of preview\short-preview. 
My impression is that we need to add a subscription on contacts up to [here](https://github.com/tbenr/status-react/blob/f45e18550e104ec58f04fb8d9a8ed0557a58c404/src/status_im/ui/screens/chat/views.cljs#L118) but i'm not that confident...

jeluard:
@tbenr Yes definitely this is the way to go! :)

tbenr:
@jeluard still wokring... almost there.

jeluard:
Nice @tbenr ! Thanks again for the hard work..

tbenr:
@jeluard :) can't believe in how many stupid things i'm struggling with!

jeluard:
That's the process of digging new projects I guess :)

StatusSceptre:
@tbenr this one is worth a tip when you're done! And extra kudos ;)

tbenr:
@jeluard finally I got it.

but I'd like to ask you one thing: I added `letsub` to subscribe against current `contacts` in the inner views **short-preview** and **preview**, then i enrich the message (getting the pubkeys too). So here I see a performance hit in case of many contacts or a lot of messages (as far as I can see, `commands/add-chat-contacts` is called for every command view, every time you hit a key in the keyboard (why?) ).

I tried the solution to subscribe at top `messages-view` or `messages-view-wrapper` but I didn't find a working way to generate the pubkey seq only once and pass along the views maps. Another solution is to add a new pubkey seq in DB and subscribe it directly.. could be better?

jeluard:
@tbenr Doesn't look like your changes should impact performance much ?

tbenr:
@jeluard I suspect it adds an overhead but I can't say how relevant could be...

jeluard (approved):

src/status_im/chat/commands/core.cljs
```diff
@@ -34,6 +34,22 @@
   [type]
   (keyword (str (protocol/id type) "-button")))
 
+(defn add-chat-contacts
+  "Enrich command-message by adding contact list of the current private or group chat"
+  [contacts {:keys [public? group-chat] :as command-message}]
+  (if public?
```
cammellos: I would probably use `cond` here and `assoc` instead of `merge` (`merge` I would use only when the set second argument is dynamic)

```
(cond 
     public? command-message
     group-chat (assoc command-message :contacts ....)
     :else          (assoc ....))
```
`cond->` could also be used I guess

src/status_im/chat/commands/core.cljs
```diff
@@ -34,6 +34,22 @@
...
+    (merge command-message
+           (if group-chat
+             {:contacts (map status-im.contact.db/public-key->address contacts)}
+             {:contact (status-im.contact.db/public-key->address (first contacts))}))))
+
+(defn enrich-command-message-for-events
+  "adds new pairs to command-message to be consumed by extension events"
+  [db {:keys [chat-id] :as command-message}]
+  (let [{:keys [contacts public? group-chat]} (get-in db [:chats chat-id])]
+    (add-chat-contacts contacts (merge command-message {:public? public? :group-chat group-chat}))))
```
cammellos: `(assoc command-message :public? public? :group-chat group-chat)`

cammellos (changes requested):
Looks good to me, just a few minor style points, could you please add some tests describing the behavior of `enrich-command-message-for-events`?

Thanks for the contribution!

src/status_im/chat/commands/core.cljs
tbenr: 👍 thanks @cammellos for the feedbacks! I prefer `cond` because of the short circuit.

tbenr:
@cammellos I applied your suggestions and added test cases as requested

cammellos (approved):
@tbenr thanks!

goranjovic (approved):

PR #7169, opened 2018-12-21
Simplify Linux bundle building

pedropombeiro:

### Summary:

After examining the [source code](https://github.com/probonopd/linuxdeployqt/blob/master/tools/linuxdeployqt/main.cpp) for `linuxdeployqt`, I realized that our usage is a bit naive and can be simplified. This PR reduces the number of calls to `linuxdeployqt` from 3 to a single call in the bundle phase of the Linux build, making bundling faster (prev: 5m30s, now: 2m30s) and easier to reason about.

I confirmed that the bundle contents are the same before and after, so no QA is needed. The only exception is that we now really don't include the Qt translations, as the stated intention in the script:

image

status: ready

### DIFF
--- a/ci/Jenkinsfile.linux
+++ b/ci/Jenkinsfile.linux
@@ -49,6 +49,7 @@ pipeline {
     LEIN_HOME = '/var/tmp/lein'
     QT_PATH = '/opt/qt'
     STATUSIM_APPIMAGE_DIR = '/opt/desktop-files'
+    VERBOSE_LEVEL = '3'
   }
 
   stages {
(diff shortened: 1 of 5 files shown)


mandrigin (approved):

pedropombeiro:
Looks like the Linux build is now 9 minutes faster (but still the longest one of the batch).

vitvly (approved):

pedropombeiro:
Sorry @mandrigin I thought I had pushed my commit amendment which simplifies the script further, so you ended up reviewing the initial commit.

pedropombeiro:
Build now takes ~17 mins, which is similar to the other target platforms.

mandrigin (approved):

PR #7173, opened 2018-12-24
[#7113] Enable EIP1102 by default

### LINKED ISSUE #7113: Enable EIP1102 by default
### User Story

As a user, I want `privacy mode` to be enabled by default

### Description

*Type*: Feature

flexsurfer:
fixes #7113

Because we'll remove this option soon, i implemented such "temporary" solution, so if an existent user has never changed this option, it will be enabled for them, and also for all new accounts

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -101,11 +101,12 @@
                    (merge testnet-networks mainnet-networks sidechain-networks))))
 
 (defn default-account-settings []
-  {:wallet {:visible-tokens {:testnet #{:STT :HND}
-                             :mainnet #{:SNT}
-                             :rinkeby #{:MOKSHA :KDO}
-                             :xdai    #{}
-                             :poa     #{}}}})
+  {:web3-opt-in? true
+   :wallet       {:visible-tokens {:testnet #{:STT :HND}
+                                   :mainnet #{:SNT}
+                                   :rinkeby #{:MOKSHA :KDO}
+                                   :xdai    #{}
+                                   :poa     #{}}}})
 
 (def currencies
   {:aed {:id :aed :code "AED" :display-name (i18n/label :t/currency-display-name-aed) :symbol "د.إ"}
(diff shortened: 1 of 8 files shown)


dmitryn (approved):

jeluard (approved):
As it is temporary, please add a TODO to simplify later.

pablanopete:
Shouldn’t we make sure the ENS Status name registration dapp works in this mode before its enabled by default? Hasn’t worked since the standard came out. 

2145d47a-4461-41b5-b273-ae6b3a173c12

antdanchenko:
@flexsurfer 
I am going to rework e2e tests in scope of the PR, changes will be added as separate commit,
please, squash my commit before merge, or I'll do that by myself when the PR will pass testing

asemiankevich:
@flexsurfer 
- `names.statusnet.eth` can't be used
- voting dapp works a bit incorrect - web3 permissions pop-up is flashing on android (so i can accept or decline it). It works a bit better on iOS after i reload page i can accept the web3 permissions.
- airswap does not support web3 , says `no wallets detected`
image
- kyber is also not detecting address
image

I havent checked all dapps but seems a lot of them do not support web3 now and this PR does make them unusable with Status. CryptoKitties work fine :D .

Still concerned about own dapps (voting dapp and ens names).  @rachelhamlin @richard-ramos FYI

asemiankevich:
@rachelhamlin please check my comment above, what do you think?

jeluard:
@asemiankevich Those DApps won't be unusable, they will force users to disable web3 privacy.
Now all browsers will make this move soon so DApps will have a big incentive to migrate.

Our DApp should be migrated in time too for sure.

asemiankevich:
> @asemiankevich Those DApps won't be unusable, they will force users to disable web3 privacy.
> Now all browsers will make this move soon so DApps will have a big incentive to migrate.
> 
> Our DApp should be migrated in time too for sure.

@jeluard ok, just outlined the concerns.

rachelhamlin:
@asemiankevich yeah, I agree that browsers will have to provide the incentive for DApps to switch. It's not overly complicated for them to do so.

As EIP1102 is a definite go and the ENS DApp now adheres to it, it makes sense that we do enable `privacy mode` by default for our next release. 

Voting DApp may have been updated along with the ENS DApp, I haven't checked.

PR #7180, opened 2018-12-25
[slow sign in] Fetch generic password only once

rasom:
Previously it was fetched each time when any realm db was opened,
but it should happen only once. Saves up to 300ms on sign in.

status: ready

### DIFF
--- a/src/status_im/utils/keychain/core.cljs
+++ b/src/status_im/utils/keychain/core.cljs
@@ -129,14 +129,23 @@
   (comp validate
         string->js-array))
 
+(defonce generic-password (atom nil))
+
 (defn get-encryption-key []
-  (log/debug "initializing realm encryption key...")
-  (.. (.getGenericPassword rn/keychain)
-      (then
-       (fn [res]
-         (if res
-           (handle-found res)
-           (handle-not-found))))))
+  (log/debug "PERF" "initializing realm encryption key..." (.now js/Date))
+  (if @generic-password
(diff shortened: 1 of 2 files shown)


src/status_im/utils/keychain/core.cljs
```diff
@@ -129,14 +129,22 @@
...
+    (js/Promise.
+     (fn [on-success _]
+       (on-success @generic-password)))
+    (.. (.getGenericPassword rn/keychain)
+        (then
+         (fn [res]
+           (if res
+             (let [res' (handle-found res)]
+               (reset! generic-password res')
+               res')
```
vitvly: `reset!` will already return `res'`, so extra expression is not needed

vitvly (approved):

src/status_im/utils/keychain/core.cljs
vitvly: So the `let` expression might be simplified to just `(reset! generic-password (handle-found res`))`

rasom: done

flexsurfer (approved):

rasom:
@siphiuel @flexsurfer, 
I added one small change to ensure that value will not be printed accidentally.

churik:
Tested on  IOS (fresh install and upgrade), Android and Mac OSx on restored and new accounts - log in and logout (including  changing fleet, bootnodes, changing network).

PR #7211, opened 2019-01-07
fix names of stages to fit what they do better

jakubgs:
Names of stages in Jenkins do not reflect what they actually do.


mandrigin (approved):
lgtm

pedropombeiro (approved):

PR #7221, opened 2019-01-08
Use better label for system browser

jeluard:
### Summary:

Update label used when opening a system browser. It will now reflect the current platform.

status: ready


src/status_im/ui/components/list_selection.cljs
```diff
@@ -30,17 +30,20 @@
          :options     (message-options message-id old-message-id text)
          :cancel-text (i18n/label :t/message-options-cancel)}))
 
+(defn- platform-web-browser []
+  (if platform/ios? :t/browsing-open-in-ios-web-browser :t/browsing-open-in-android-web-browser))
```
cammellos, line 34: no desktop?

cammellos (approved):

src/status_im/ui/components/list_selection.cljs
jeluard: No browser on desktop for now? And not sure there would be a distinction between browsers as on mobile. But good point.

goranjovic (approved):

churik:
@jeluard 
Tested on IOS and Android in 1-1, public and group chats.
Noticed that for all languages except English labels are the same.
Will they be translated separately?

jeluard:
@churik Yes the old label is used for both platforms for most languages. No plan to translate them for now.

PR #7233, opened 2019-01-09
Add Docker image for building for Android

jakubgs:
Resolves: https://github.com/status-im/status-react/issues/7054

Changes:
- Moves docker building files to the `docker` folder.
- Creates a `base` image that will be used by all platforms
- Adds a Docker image for building android app
- Provides credentials necessary to sign Android release via Jenkins env variables

The image is a bit big right now - 4.2 GB - and I intend to do another separate PR that makes it smaller, and also makes other two images use the base image.


docker/android/Dockerfile
```diff
@@ -0,0 +1,65 @@
+# This image is only for extracting and cleanup NDK and SDK
```
PombeirP:
```suggestion
# This image is only for extracting and cleaning up NDK and SDK
```

docker/android/Makefile
```diff
@@ -0,0 +1,57 @@
+GIT_COMMIT = $(shell git rev-parse --short HEAD)
+
+# Software Versions, URLs, and Checksums
+NVM_VERSION = v0.33.11
+NVM_URL = https://raw.githubusercontent.com/creationix/nvm/$(NVM_VERSION)/install.sh
+NVM_INSTALL_SCRIPT = nvm_$(NVM_VERSION)_install.sh
+NVM_NODE_VERSION = v10.14.1
```
PombeirP: We're not using the same Node version as we're installing in `installer.sh` (v10.14.0), we should pick one and stick with it.

docker/android/Makefile
```diff
@@ -0,0 +1,57 @@
...
+	wget -q "$(ANDROID_NDK_URL)" -O "$(ANDROID_NDK_ARCHIVE)"
+	echo "$(ANDROID_NDK_CHECKSUM)  $(ANDROID_NDK_ARCHIVE)" | md5sum --check
+
+$(ANDROID_SDK_ARCHIVE):
+	wget -q "$(ANDROID_SDK_URL)" -O "$(ANDROID_SDK_ARCHIVE)"
+	echo "$(ANDROID_SDK_CHECKSUM)  $(ANDROID_SDK_ARCHIVE)" | md5sum --check
+
+test: ## Run build inside the image as a test
+	docker run -u 1001:1002 \
+        --name android-test --rm \
```
PombeirP: Indent the argument `--name` relative to the command?

docker/android/Makefile
```diff
@@ -0,0 +1,57 @@
...
+		-v /var/tmp/npm:/var/tmp/npm:rw \
+		-v /var/tmp/lein:/var/tmp/lein:rw \
+		-v /var/tmp/yarn:/var/tmp/yarn:rw \
+        -v /var/tmp/gradle:/home/jenkins/.gradle/caches:rw \
```
PombeirP: Same here

docker/android/build.sh
```diff
@@ -0,0 +1,24 @@
...
+#make prepare-android
+## Lint
+#lein cljfmt check
+## Test
+#lein test-cljs
+## Build
+#lein prod-build-android
+# Compile
+id
+cd android
```
PombeirP: Maybe we should run `gradlew` only if `cd android` is successful?

docker/android/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e -x
+
+export FASTLANE_DISABLE_COLORS=1
+export REALM_DISABLE_ANALYTICS=1
+export STATUS_RELEASE_STORE_PASSWORD=1TteVCyHykCa
```
PombeirP: What credentials are these?

mobile_files/yarn.lock
```diff
@@ -59,12 +59,12 @@
     source-map "^0.5.0"
     trim-right "^1.0.1"
 
-"@babel/generator@^7.1.6":
```
PombeirP: Curious as to why this particular PR requires updating `yarn.lock`. If not strictly related, should it be done in a separate PR?

docker/android/Makefile
jakubgs: Good point. Will change that, thanks!

docker/android/build.sh
jakubgs: That's already covered by `set -e -x`.

docker/android/build.sh
jakubgs: Ah, that's from some of my tests. Let me drop that.

mobile_files/yarn.lock
jakubgs: That's because I was having yarn throw errors at me about unpacking, and the suggestion was to try to refresh `yarn.lock`, but I can try rebasing with latest `develop` and see if that issue is gone.

docker/android/Dockerfile
jakubgs: Done.

docker/android/build.sh
jakubgs: Besides, this was just for my own tests while making the image.

PR #7246, opened 2019-01-10
[WIP] Feature/EIP-1193 subscriptions

flexsurfer:
waiting for changes in status-go, currently, there is an error  "error" "The method eth_logs does not exist/is not available"

status: wip


Swader:
Unknown if still needed

PR #7251, opened 2019-01-10
Fixed wallet settings hook screen on iphone x

flexsurfer:
Fixed wallet settings hook screen on iphone x


jeluard (approved):

goranjovic (approved):

Serhy:
Blue background now for Wallet settings hook screen on iPhoneX (and also looks good with Android 6.0.1 and 7)

PR #7254, opened 2019-01-10
Report messages as "processed" to status-go only after they were stored in Realm

mandrigin:
fixes: https://www.pivotaltracker.com/story/show/162661405

### Summary

This suppose to fix the potential issue I was able to reproduce with chats being different on Desktop and mobile. The working theory was that these messages were reported as "confirmed" to status-go side (deduplication) but were never added to the DB.

In this PR, messages actually reported as "confirmed" if and only if they were successfully stored in Realm. This adds an additional event to `re-frame`, but this makes the code to handle errors correctly.

#### Platforms (optional)
- Android
- iOS
- macOS
- Linux
- Windows

**Functional**
- 1-1 chats
- public chats
- group chats

status: ready 

<blockquote><div><strong><a href="https://www.pivotaltracker.com/story/show/162661405">As a user, I want consistent message history between desktop and mobile.  - Pivotal Tracker</a></strong></div></blockquote>

### DIFF
--- a/src/status_im/chat/core.cljs
+++ b/src/status_im/chat/core.cljs
@@ -5,21 +5,20 @@
 ;; Seen messages
 (fx/defn receive-seen
   [{:keys [db js-obj]} chat-id sender {:keys [message-ids]}]
-  (merge {:transport/confirm-messages-processed [{:web3   (:web3 db)
-                                                  :js-obj js-obj}]}
-         (when-let [seen-messages-ids (-> (get-in db [:chats chat-id :messages])
-                                          (select-keys message-ids)
-                                          keys)]
-           (let [statuses (map (fn [message-id]
-                                 {:chat-id          chat-id
-                                  :message-id       message-id
-                                  :public-key       sender
-                                  :status           :seen})
-                               seen-messages-ids)]
-             {:db            (reduce (fn [acc {:keys [message-id] :as status}]
-                                       (assoc-in acc [:chats chat-id :message-statuses
-                                                      message-id sender]
-                                                 status))
-                                     db
-                                     statuses)
-              :data-store/tx [(user-statuses-store/save-statuses-tx statuses)]}))))
+  (when-let [seen-messages-ids (-> (get-in db [:chats chat-id :messages])
+                                   (select-keys message-ids)
+                                   keys)]
+    (let [statuses (map (fn [message-id]
+                          {:chat-id          chat-id
+                           :message-id       message-id
+                           :public-key       sender
+                           :status           :seen})
+                        seen-messages-ids)]
+      {:db            (reduce (fn [acc {:keys [message-id] :as status}]
+                                (assoc-in acc [:chats chat-id :message-statuses
+                                               message-id sender]
+                                          status))
+                              db
+                              statuses)
+       :data-store/tx [{:transaction (user-statuses-store/save-statuses-tx statuses)
+                        :success-event [:message/message-persisted js-obj]}]})))
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -111,7 +111,7 @@
 (fx/defn add-message
   [{:keys [db] :as cofx}
    {{:keys [chat-id message-id clock-value timestamp from] :as message} :message
-    :keys [current-chat? batch? last-clock-value]}]
+    :keys [current-chat? batch? last-clock-value raw-message]}]
   (let [current-public-key (accounts.db/current-public-key cofx)
         prepared-message (-> message
                              (prepare-message chat-id current-chat?)
@@ -136,7 +136,10 @@
                                      (not= from current-public-key))
                                 (update-in [:chats chat-id :loaded-unviewed-messages-ids]
                                            (fnil conj #{}) message-id))
-               :data-store/tx [(messages-store/save-message-tx prepared-message)]}
+               :data-store/tx [(merge
+                                {:transaction (messages-store/save-message-tx prepared-message)}
+                                (when raw-message
+                                  {:success-event [:message/message-persisted raw-message]}))]}
               (when (and platform/desktop?
                          (not batch?)
                          (not (system-message? prepared-message)))
@@ -196,11 +199,10 @@
                                           ;; TODO (cammellos): Refactor so it's not computed twice
                                           (add-outgoing-status current-public-key))]
     (fx/merge cofx
-              {:transport/confirm-messages-processed [{:web3   web3
-                                                       :js-obj js-obj}]}
               (add-message {:batch?       true
                             :message      message
-                            :current-chat current-chat?})
+                            :current-chat current-chat?
+                            :raw-message  js-obj})
               ;; Checking :outgoing here only works for now as we don't have a :seen
               ;; status for public chats, if we add processing of our own messages
               ;; for 1-to-1 care needs to be taken not to override the :seen status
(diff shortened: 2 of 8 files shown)


src/status_im/chat/core.cljs
```diff
@@ -1,12 +1,12 @@
...
   (:require [status-im.data-store.user-statuses :as user-statuses-store]
+            [re-frame.core :as re-frame]
             [status-im.utils.fx :as fx]))
 
 ;; Seen messages
 (fx/defn receive-seen
   [{:keys [db js-obj]} chat-id sender {:keys [message-ids]}]
-  (merge {:transport/confirm-messages-processed [{:web3   (:web3 db)
-                                                  :js-obj js-obj}]}
+  (merge {}
```
rasom: no need to merge with an empty map

src/status_im/chat/models/message.cljs
```diff
@@ -136,7 +136,9 @@
                                      (not= from current-public-key))
                                 (update-in [:chats chat-id :loaded-unviewed-messages-ids]
                                            (fnil conj #{}) message-id))
-               :data-store/tx [(messages-store/save-message-tx prepared-message)]}
+               :data-store/tx [(messages-store/save-message-tx
+                                prepared-message
+                                #(re-frame/dispatch [:message/message-persisted raw-message]))]}
```
rasom: Dispatching an event per each received message may have a negative effect on performance if a batch of them was received. This can be optimized later.

mandrigin: I don’t see how it is much worse than it was, taking into account that we open a tx per message and always were opening it. But sure, the room for optimization is there.

rasom: We do not open tx per message: all `:data-store/tx` effects which are produced during a single event handling are accumulated and executed in a single transaction. See the list of effects which are "merged" this way here https://github.com/status-im/status-react/blob/1104becfbab11506b98028ce81ad6ff869381f51/src/status_im/utils/fx.cljs#L13

src/status_im/data_store/realm/core.cljs
```diff
@@ -285,16 +285,25 @@
...
    (create realm schema-name obj false))
   ([realm schema-name obj update?]
+   (create realm schema-name obj update? nil))
+  ([realm schema-name obj update? on-success]
    (let [obj-to-save (select-keys obj (keys (get-in entity->schemas
-                                                    [schema-name :properties])))]
-     (.create realm (name schema-name) (clj->js obj-to-save) update?))))
+                                                    [schema-name :properties])))
+         obj-in-realm (.create realm (name schema-name) (clj->js obj-to-save) update?)]
+     (when (and (not (nil? on-success))
```
rasom: `(not (nil?` can be replaced with `fn?` here

src/status_im/data_store/user_statuses.cljs
```diff
@@ -40,7 +40,14 @@
...
   "Returns tx function for saving message user statuses"
-  [user-statuses]
-  (fn [realm]
-    (doseq [user-status user-statuses]
-      ((save-status-tx user-status) realm))))
+  ([user-statuses]
+   (save-statuses-tx user-statuses nil))
+  ([user-statuses on-success]
+   (fn [realm]
+     (let [success (reduce *
```
rasom: 1. [`doseq`](https://clojuredocs.org/clojure.core/doseq) always returns nil, not a collection. It is meant for side effects. So here it could be replaced with `for`. It is literally `(reduce * [])` now which always returns `1` as `(*)` returns `1` 
2. `(reduce * [true false true ...])` even if collection contains only one `false` value the whole collection of messages will be treated as not persisted, do we really want this? 
3. this code is not going to be run in the app currently as `seen` messages are not working. How soon are we going to "ressurect" them? If we don't have a plan to do this better just to remove dead code.

rasom (changes requested):
see messages above

src/status_im/data_store/realm/core.cljs
cammellos: mmh, not sure this is what we want exactly, create is run in a transaction, so `on-success` should be only run if all the transactions are successful, otherwise they are all reverted, and in that case we are in the same situation as now, here it looks like it is called straight after calling `create`. I might be missing something though.

I think we have a similar mechanism already https://github.com/status-im/status-react/blob/a4efc661e18b353b84de242e8c44018d21856307/src/status_im/data_store/core.cljs#L49 
so probably you can just add a `success-event` to the message transaction, and it will be fired only after the transaction is successful.

src/status_im/data_store/messages.cljs
```diff
@@ -95,14 +95,17 @@
...
-  (fn [realm]
-    (core/create realm
-                 :message
-                 (prepare-message (merge default-values
-                                         message
-                                         {:from (or from "anonymous")}))
-                 true)))
+  ([message]
+   (save-message-tx message nil))
+  ([{:keys [message-id from] :as message} on-success]
```
cammellos: here we could try something like:

```
{:success-event [:message-confirmed raw-message] ; can be passed as a parameter 
  :transaction (fn [realm] .....)}
```
and theoretically it might just be the only change we need

src/status_im/data_store/user_statuses.cljs
cammellos: same here, probably we need to wait for the transaction to be committed before dispatching

src/status_im/data_store/realm/core.cljs
mandrigin: oh, thanks! I didn't know about this one...

src/status_im/data_store/user_statuses.cljs
```diff
@@ -40,7 +40,8 @@
 
 (defn save-statuses-tx
   "Returns tx function for saving message user statuses"
-  [user-statuses]
-  (fn [realm]
-    (doseq [user-status user-statuses]
-      ((save-status-tx user-status) realm))))
+  ([user-statuses]
```
cammellos: Parentheses are not necessary here (only needed if multiple signatures are provided)

src/status_im/data_store/core.cljs
```diff
@@ -46,14 +46,28 @@
    (then
     #(data-source/open-account address password encryption-key))))
 
+(defn- merge-events-of-type [success-events event-type]
```
cammellos: I would probably write a few tests documenting the behavior (never mind that is not public, but better have it covered),

src/status_im/data_store/user_statuses.cljs
mandrigin: whoops, a leftover :)

src/status_im/data_store/core.cljs
mandrigin: good point, this one should be easily tested :)

src/status_im/data_store/user_statuses.cljs
mandrigin: fixed

mandrigin:
@rasom I fixed the way we do things there

rasom (approved):

src/status_im/data_store/core.cljs
mandrigin: @cammellos I added unit-tests to this function.

cammellos:
@mandrigin looks good, thanks!

cammellos (approved):

Serhy:
Lgtm! No regressions found, no after-upgrade bugs, no performance degradation (tested chat with 100 text simple messages, 5 huge (3k chars) messages).

PR #7258, opened 2019-01-11
Fix issues found in dependency audit v1

pedropombeiro:

Pivotal stories fixed:
https://www.pivotaltracker.com/story/show/163088518
https://www.pivotaltracker.com/story/show/163088531
https://www.pivotaltracker.com/story/show/163088536

### Summary:

This PR addresses the first 3 points of the discuss post at https://discuss.status.im/t/dependency-audit-v1/926. It doesn't address locking down tool versions in Linux (only on MacOS) since that will be solved through a Docker image that can be used to build and debug on Linux.

It adds yarn to the repository so that we always keep control of the version used without imposing that on the user.

### Review notes (optional):

I've tested the scripts on Linux, but would be nice to test how it works on Mac (maybe some changes to the scripts are not fully compatible with Mac).

#### Platforms (optional)
- macOS
- Linux

status: ready


mobile_files/package.json.orig
```diff
@@ -15,7 +15,7 @@
     "assert": "1.4.1",
     "asyncstorage-down": "4.0.1",
     "babel-preset-react-native": "5.0.2",
-    "bignumber.js": "https://github.com/status-im/bignumber.js.git",
+    "bignumber.js": "github:status-im/bignumber.js#master",
```
mandrigin: I would prefer making a tag/release in our fork and stick to it instead of having `master` branch there.

scripts/lib/setup/installers.sh
```diff
@@ -2,6 +2,16 @@
 
 _localPropertiesPath=./android/local.properties
 
+function getRequiredVersion() {
```
mandrigin: I would extract this function to a separate file, maybe. Code deduplication and you can stll use it inside as `$(./get-tool-version.sh conan)`.

mandrigin (changes requested):
I think you did a great job! A few minor points to marginally improve it :)

mobile_files/package.json.orig
jakubgs: Yeah, also, please don't drop the `.git` from it. I've seen before both `npm` and `yarn` issues caused by this.

desktop/docker/linux/Makefile
```diff
@@ -6,6 +6,9 @@ QT_VER_MJR = 5.11
 QT_ARCHIVE = qt-opensource-linux-x64-$(QT_VERSION).run
 QT_MD5SUM  = 974fda61267cfb6e45984ee5f0a285f8
 QT_URL = https://download.qt.io/archive/qt
+QT_CI_COMMIT_SHA = 55ffd9f225708b3aa1443851cfa8dead2c1f9959
+REACT_NATIVE_DESKTOP_COMMIT_SHA = 2a76c435a27f0bf6c089f74222212b3a58e10385
+__read__toolversion__ = $(shell grep $(1) ../../../.TOOLVERSIONS | cut -d'=' -f2-)
```
jakubgs, line 11: Cool trick.

jakubgs (approved):
Nice work.

mobile_files/package.json.orig
pedropombeiro: > Yeah, also, please don't drop the `.git` from it. I've seen before both `npm` and `yarn` issues caused by this.

Ah, I just copied what was on desktop to make them the same, didn't notice that. I'll add it back.

pedropombeiro: Moving this specific commit to a future PR since it involves changes in dependencies.

pedropombeiro: > I would prefer making a tag/release in our fork and stick to it instead of having master branch there.

@mandrigin I have the impression that most of our customized npm dependencies don't follow that (they either point to a branch or to a sha). Is it really a problem for our purposes? AFAICS if someone calls `yarn install` it will retrieve the last resolved sha even if the branch moved. It is only when you call `yarn upgrade` that it will be problematic because it will now be pointing to a newer commit. Am I missing something?

mandrigin: it won't help from history rewrites if we still want to build older versions. otherwise, it is okay, yes.

mandrigin (approved):

scripts/lib/setup/installers.sh
pedropombeiro: @mandrigin Updated 07cf219a6153bb1d1a830f25e4ea2145f9246003 to move the method to `packages.sh` and reuse it and `program_exists` from other scripts.

pedropombeiro:
Looking at docs.status.im to see if there's something in need of update.

pedropombeiro:
Any reason to not install `nvm` if we see that the user doesn't have neither `node` nor `nvm` on his machine? Right now it will install node through the package manager, which seems sub-optimal.

mandrigin:
@PombeirP I think we should always handle everything through nvm

PR #7272, opened 2019-01-15
ignore keys from /private/var/folders, delete-keychain after use

jakubgs:
Resolves issue with temporary keychains showing up after `security list-keychains` when signing desktop app:
```
### Adding code-signing keychain to search list...
    "/Users/jenkins/Library/Keychains/login.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.yxhfQkBR/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.D8l0Y0dZ/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.KgxajoRV/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.u3uZ5NqH/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.42KaB6hF/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.elctVcZE/macos-developer-id.keychain-db"
    "/private/var/folders/lg/klkwp62n51v5y38020d8x4xc0000h8/T/tmp.ii5IHVYZ/macos-developer-id.keychain-db"
    "/Library/Keychains/System.keychain"
```
Resolves: https://github.com/status-im/status-react/issues/6352

### DIFF
--- a/scripts/sign-macos-pkg.sh
+++ b/scripts/sign-macos-pkg.sh
@@ -38,7 +38,9 @@ set -e
 
 echo -e "\n### Storing original keychain search list..."
 ORIG_KEYCHAIN_LIST="$(security list-keychains \
-  | grep -v "/Library/Keychains/System.keychain" | xargs)"
+  | grep -v "/Library/Keychains/System.keychain" \
+  | grep -v "/private/var/folders" \
+  | xargs)"
 
 echo -e "\n### Creating ramdisk..."
 RAMDISK="$(hdiutil attach -nomount ram://20480 | tr -d '[:blank:]')"
@@ -65,8 +67,8 @@ function clean_up {
   security list-keychains -s $ORIG_KEYCHAIN_LIST
   security list-keychains
 
-  echo -e "\n### Wiping keychain file..."
-  rm -P "$KEYCHAIN"
+  echo -e "\n### Delete keychain file..."
+  security delete-keychain "$KEYCHAIN"
 
   echo -e "\n### Destroying ramdisk..."
   diskutil umount force "$RAMDISK"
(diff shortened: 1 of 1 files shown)


vkjr (approved):

scripts/sign-macos-pkg.sh
```diff
@@ -65,8 +67,8 @@ function clean_up {
   security list-keychains -s $ORIG_KEYCHAIN_LIST
   security list-keychains
 
-  echo -e "\n### Wiping keychain file..."
-  rm -P "$KEYCHAIN"
+  echo -e "\n### Delete keychain file..."
+  security delete-keychain "$KEYCHAIN"
```
mandrigin, line 71: this should also happen on errors, I think that is what caused initial issues

mandrigin (changes requested):
one small thing, looks good otherwise

vitvly (approved):

scripts/sign-macos-pkg.sh
jakubgs: It does happen on errors. See:
https://github.com/status-im/status-react/pull/7272/files#diff-7020266fb517002c883a982991d1b155R80
```
trap clean_up EXIT
```

mandrigin (approved):

scripts/sign-macos-pkg.sh
jakubgs: Although, maybe we should make it to be:
```
trap clean_up EXIT ERR
```
Just in case.

mandrigin:
thanks!

pedropombeiro (approved):

PR #7277, opened 2019-01-15
Ensure all dependencies are tagged

pedropombeiro:

### Summary:

This PR is a step towards reproducible builds, by ensuring that we only depend on packages which are tagged. This way if the branch of a dependency moves, we can still build the same app without unintended consequences, and all updates must be explicit from now on. Hopefully this will become a PR check from Probot (that we don't add a forked repo as a dependency, without reference to a tag/release).

It also standardizes on `git+https` URL style, since `github:status-im/*` style URLs seem to break Android builds, and having different styles for same package between mobile and desktop creates issues in `yarn check`.

#### Platforms (optional)

Some desktop package dependencies were updated so they would match what we're using on mobile. I've only done this when there seemed to be no changes that would affect desktop, but in any case, it would be good to do a cursory check (especially account creation, recovery and webview).

status: ready


mandrigin (approved):
oh, that looks much better!

one small point is that I would like to have this rule about using `git+https` documented and enforced somehow...

vkjr (approved):

mandrigin:
@PombeirP it would be cool to merge this one as soon as possible, its an easy place of conflicts and it doesn't need manual testing

PR #7278, opened 2019-01-15
use nvm() wrapper in jenkins for android builds

jakubgs:
Android builds seem to be the only ones not using this `nvm()` wrapper for correct NodeJS version.

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -44,32 +44,36 @@ pipeline {
           mobile = load 'ci/mobile.groovy'
           cmn    = load 'ci/common.groovy'
           print "Running ${cmn.getBuildType()} build!"
-          /* Cleanup and Prep */
-          mobile.prep(cmn.getBuildType())
           /* Run at start to void mismatched numbers */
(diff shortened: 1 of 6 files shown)


ci/Jenkinsfile.android
```diff
@@ -24,6 +24,7 @@ pipeline {
 
   environment {
     BUILD_PLATFORM = 'android'
+    NODE_VERSION = 'v10.14.1'
```
pedropombeiro: This should be `10.12.0`, right?

pedropombeiro (changes requested):

ci/Jenkinsfile.android
pedropombeiro: would be cool if we could read the version from `.TOOLVERSIONS` :-)

jakubgs: Oh, is that change already merged? Then i can try that.

pedropombeiro (approved):
Awesome, thanks Jakub!

mandrigin (approved):

PR #7279, opened 2019-01-15
[TESTS] Added simple upgrade test

antdanchenko:
* added simple upgrade test
* changed apk name of nightly builds on sauce to reflect common name pattern (StatusIm-190115-125934-dd31ee-e2e.apk instead of im.status.ethereum-e2e-dd31ee.apk)


antdanchenko:
@jakubgs please review changes in  [ci/android.groovy](https://github.com/status-im/status-react/pull/7279/files#diff-921ddb8a284f72367049b92eff26b9b6)

ci/android.groovy
```diff
@@ -28,7 +28,7 @@ def uploadToSauceLabs() {
   if (changeId != null) {
     env.SAUCE_LABS_NAME = "${changeId}.apk"
   } else {
-    env.SAUCE_LABS_NAME = "im.status.ethereum-e2e-${GIT_COMMIT.take(6)}.apk"
+    env.SAUCE_LABS_NAME = "${pkg}"
```
jakubgs, line 44: Where does `pkg` come from in that context? I don't see it's source.

jakubgs (changes requested):

ci/android.groovy
antdanchenko: from https://github.com/status-im/status-react/pull/7279/files/8eedc83cc57ef5ffd826dc3abab5c5a6b07655e4#diff-921ddb8a284f72367049b92eff26b9b6R12
Or I can't use this in uploadToSauceLabs?

jakubgs: Nope, it's a context of a different method, so I doubt that will work.

antdanchenko: It seems that I should call `cmn.pkgFilename(type, 'apk')`, right?

jakubgs: Yeah, that should work. I don't think passing that from the original place would make for nice code, that should be just fine.

antdanchenko: i've added the definition of pkg to the method

jakubgs (approved):
lgtm

annadanchenko (approved):

Serhy (approved):

PR #7280, opened 2019-01-15
Install nvm if no system nodejs is found

pedropombeiro:
### Summary:

This PR changes `make setup` behavior from:

1. nvm is not found
1. nodejs is not found
1. nodejs is installed using package manager

to:

1. nvm is not found
1. nodejs is not found
1. nvm is downloaded and installed
1. nodejs is installed using nvm

status: ready

### DIFF
--- a/scripts/lib/setup/installers.sh
+++ b/scripts/lib/setup/installers.sh
@@ -16,6 +16,10 @@ function install_nsis() {
 }
 
 function install_node() {
+  if ! nvm_installed && ! program_exists 'node'; then
+    install_nvm
+  fi
+
   if nvm_installed; then
     install_node_via_nvm
   else
@@ -230,6 +234,22 @@ function install_yarn() {
       cecho "@b@yellow[[+ yarn version $yarn_version is installed. Downloading yarn version $required_version in the local repo.]]"
       yarn policies set-version $required_version
     fi
+  fi
+}
+
+function install_nvm() {
+  local required_version=$(get_tool_version nvm)
+
+  if ! program_version_exists 'nvm' "$required_version"; then
+    cecho "@b@blue[[+ Installing nvm $required_version]]"
+
+    sudo apt install -y build-essential libssl-dev
+    source scripts/3rd-party/nvm/${required_version}/install.sh
+    export NVM_DIR="$HOME/.nvm"
+    [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
+    [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
+  else
+    cecho "+ nvm already installed... skipping."
   fi
 }
 
--- a/.gitignore
+++ b/.gitignore
@@ -147,6 +147,7 @@ Status-Windows-x86_64.exe
 /desktop/CMakeFiles/
 /desktop/reportApp/Makefile
 /deployment/windows/Status.rc
+/docker/*/scripts/
 *_autogen/
 CompleteBundleWin.cmake
 logger_settings.ini
--- a/docker/linux/Dockerfile
+++ b/docker/linux/Dockerfile
@@ -83,9 +83,10 @@ RUN mkdir -p $LEIN_INSTALL \
 ARG NVM_VERSION
 ARG NODE_VERSION
 ARG YARN_VERSION
-RUN curl -s -o- https://raw.githubusercontent.com/creationix/nvm/v${NVM_VERSION}/install.sh | bash
+ADD scripts/3rd-party/nvm/${NVM_VERSION}/install.sh /tmp/nvm/${NVM_VERSION}/
 
-RUN mkdir -p /tmp/react-native-desktop && cd /tmp/react-native-desktop && \
+RUN /tmp/nvm/${NVM_VERSION}/install.sh && \
+    mkdir -p /tmp/react-native-desktop && cd /tmp/react-native-desktop && \
       git init && git remote add origin https://github.com/status-im/react-native-desktop.git && \
       git fetch --depth=1 origin ${REACT_NATIVE_DESKTOP_COMMIT_SHA} && \
       git reset --hard FETCH_HEAD && \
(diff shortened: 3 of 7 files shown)


scripts/lib/setup/installers.sh
```diff
@@ -233,6 +237,22 @@ function install_yarn() {
...
 }
 
+function install_nvm() {
+  local required_version=$(get_tool_version nvm)
+
+  if ! program_version_exists 'nvm' "$required_version"; then
+    cecho "@b@blue[[+ Installing nvm $required_version]]"
+
+    sudo apt install -y build-essential libssl-dev
+    curl -o- https://raw.githubusercontent.com/creationix/nvm/v${required_version}/install.sh | bash
```
jakubgs: One day God will punish us for our hubris.

jakubgs (approved):

scripts/lib/setup/installers.sh
mandrigin: Agree with Jakub I'd rather vendor this file both for security and for stability reasons. One less network request.

mandrigin (changes requested):
let's avoid running shell scripts from GitHub

pedropombeiro:
> let's avoid running shell scripts from GitHub

We already do it in Dockerfiles, but I guess I could host the script in `scripts/3rd-party`. Would that be OK?

jakubgs:
>I guess I could host the script in scripts/3rd-party

If you put it there then we could also use it in Docker creation(in a separate PR).

mandrigin:
@PombeirP that sounds good, yes.

pedropombeiro:
Ping @mandrigin

desktop/docker/windows/Dockerfile
```diff
@@ -49,9 +49,10 @@ RUN mkdir -p $LEIN_INSTALL \
 
 ARG NVM_VERSION
 ARG NODE_VERSION
-RUN curl -s -o- https://raw.githubusercontent.com/creationix/nvm/v${NVM_VERSION}/install.sh | bash
+ADD scripts/3rd-party/nvm/${NVM_VERSION}/install.sh /tmp/nvm/${NVM_VERSION}
```
jakubgs: I really don't think that's going to work. Dockerfile paths are relative to the location of the Dockerfile and I don't think we can include a file from outside of that folder.
Maybe we can symlink the script inside of the directory?

pedropombeiro: Oh, that's right, good catch! :man_facepalming: I should have tested it before pushing the change...

pedropombeiro: Added symlink

jakubgs: Why do you do the whole `scripts/3rd-party/nvm/${NVM_VERSION}/` path? We don't intend to install multiple versions, so why don't you just symlink the script itself in the directory?

pedropombeiro: This is to continue having the version selection being done in the `Dockerfile`, and also to avoid having to regenerate the symlink every time we update a tool version. Makes sense?

jakubgs: Oh, I see, you symlinked the whole `3rd-party` folder so the selection would work. I guess that's fine.

Did you check that this works?

pedropombeiro: I ended up switching to an alternative solution, I'm copying the `scripts/3rd-party` tree to the docker context just for the duration of the docker build.

scripts/lib/setup/installers.sh
pedropombeiro: Done.

mandrigin (approved):

antdanchenko:
There is [an issue](https://assets.saucelabs.com/jobs/2ff44cb7999b4d3fa6e119afef0394a0/0019screenshot.png?auth=0806a4ab20d7d1b2bba55a67312fb744) in product which definitely is not introduced by this PR

However the issue is absent in todays nightly build

Looks like it was introduced in https://github.com/status-im/status-react/pull/7283 
since only #7283  was merged without any testing

I am going to rerun e2e tests against #7283 and if it's so, we may consider reverting the PR

mandrigin:
@antdanchenko yes, please, ping me and @jeluard if that is the case

jeluard:
@antdanchenko Haven't e2e tests run for #7283? If no, when are they triggered?
Are we still considering blocking PR while e2e haven't run?

mandrigin:
@jeluard afaik e2e tests are triggered when the task is moved to "TO TEST" on the "Pipeline for QA".

antdanchenko:
@jeluard we are blocking PR's on start of the e2e test job

A PR needs to be in `TO TEST` section on https://github.com/status-im/status-react/projects/7 with available builds for getting e2e tests running against it
Also you may need to wait for ~20 minutes in queue since tests might be running against other PR in that moment

Next step will be to set blocked status for all PRs by default

mandrigin:
@antdanchenko it maybe would be nice to tell everyone on the next town hall about it, because not everyone is following these changes.

jeluard:
@antdanchenko Could we automate the move to `TO TEST` once 2 reviews have been performed? While it might make sense to skip manual tests, it probably does not to skip automated tests.

antdanchenko:
@jeluard it's already automated (should be Autobot), @PombeirP correct me if I am wrong

antdanchenko:
@jeluard 
Whenever we are making any changes in product there is no sense to skip manual testing, even if we are 99% sure that those changes bring no issues

antdanchenko:
@mandrigin, @jeluard
The issue from this PR is introduced in #7283 
https://github.com/status-im/status-react/pull/7283#issuecomment-455169741

mandrigin:
@jeluard I can reproduce this issue with #7283, but when I revert it the issue disappears. So I will go ahead and revert it on develop so you have more time to debug and fix it.

How to reproduce: try sending any amount to any contact, the sending screen will be broken.

antdanchenko:
@PombeirP please rebase this PR on latest develop when #7291 will be merged

After that I will re-run only failed tests from https://github.com/status-im/status-react/pull/7280#issuecomment-455159614 and your PR will be unblocked

mandrigin:
@PombeirP @antdanchenko the revert was just merged

pedropombeiro:
@antdanchenko @jeluard That is correct: `bot_scripts/assign-approved-pr-to-test.js` will move to `TO TEST` if 2 reviewers have approved and everything is green (maybe that's what was missing).

jeluard:
@PombeirP No you can't merge w/o 2 reviews.
In my experience there is never an automated move to `TO TEST`. Unless this is checked every couple hours.

pedropombeiro:
> @PombeirP No you can't merge w/o 2 reviews.
> In my experience there is never an automated move to `TO TEST`. Unless this is checked every couple hours.

That's not what I said. What I said is that on top of needing two reviews, the GitHub status for the PR needs to be all green (`All checks have passed`).
I believe the bot checks all open PRs every 15 minutes.

pedropombeiro:
As an example that the bot is working, here's a recent PR that the bot picked up and moved to test as everything was green: https://github.com/status-im/status-react/pull/7277

Issue #27, opened 2016-04-25
Confirmation Code should be number input


0xc1c4da:
https://assembly-vc.slack.com/files/jarradhope/F135PNBJ9/confirmationcodeshouldbenumberinput.png

PR #7287, opened 2019-01-16
Change App ID for PR builds of Android app

jakubgs:
We don't want to make the PR builds replace the release when installed, or have access to the accounts.

For that reason this PR changes the ID of Android app from `im.status.ethereum` to `im.status.ethereum.pr` and name from `Status` to `Status PR` for PR builds.

Resolves: https://github.com/status-im/status-react/issues/7286
Status: READY


jakubgs:
It doesn't seem like `gradle` likes my changes:
```
Execution failed for task ':app:processReleaseGoogleServices'.
> No matching client found for package name 'im.status.ethereum.pr'
```

mandrigin:
@jakubgs you maybe can use something like what we use for running dev versions: `react-native run-android --appIdSuffix debug`

jakubgs:
Thanks @mandrigin I'll look into that. But I've been thinking that maybe the proper way of doing this is adding another `buildType` for android:
https://github.com/status-im/status-react/blob/develop/android/app/build.gradle#L195
There's a `release` and `debug`, maybe there should be also a `pr`. But I'm not yet sure how that is selected.
https://developer.android.com/studio/build/build-variants

<blockquote><div>Android Developers</div><div><strong><a href="https://developer.android.com/studio/build/build-variants">Configure build variants  |  Android Developers</a></strong></div></blockquote>

jakubgs:
__OR__ I could just do the `.debug` suffix and call it a day. Not sure if we care about the name referencing PRs. We just want the ID to be different from release ones, right?

mandrigin:
@jakubgs I guess we might need to update package name in `android/app/google-services.json` too...

jakubgs:
>@jakubgs I guess we might need to update package name in android/app/google-services.json too...

I don't think that's correct. I think the ones we need to update are in `android/app/src/`, and add another type of build which is PR, and then run `gradlew assemblePr` instead of `gradlew assembleRelease`.

android/app/src/pr/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">Status Debug</string>
```
mandrigin: Status PR Build?

jakubgs: Right.

ci/ios.groovy
```diff
@@ -19,6 +19,10 @@ def bundle(type) {
   plutil('CFBundleShortVersionString', cmn.version())
   plutil('CFBundleVersion', cmn.genBuildNumber())
   plutil('CFBundleBuildUrl', currentBuild.absoluteUrl)
+  if (type == 'pr') { /* PR builds shouldn't replace normal releases */
+    env.BUNDLE_ID_SUFFIX = '.pr'
+    //plutil('CFBundleIdentifier', 'im.status.ethereum.pr')
```
mandrigin: commented-out line... is that expected?

jakubgs: Considering PR has `Status: WIP` in description, yes :D.

flexsurfer (approved):

pedropombeiro (approved):
lgtm

jakubgs:
I've installed the [Android Build](https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190209-131905-7ebdbd-pr.apk) and it showed up as a new application and did not have access to old accounts.
Can someone confirm for iOS?

Also, Both Status apps show up as "Status". Maybe the name should be also different. I'll look into it.

mandrigin:
@jakubgs iOS has the same bundle ID (I tried build `#27`).

jakubgs:
I've decided that doing the same change for iOS is going to be very messy so I'm limiting this PR just to Android changes so I can merge it. I'll do iOS separately.

pedropombeiro:
lgtm

PR #7298, opened 2019-01-18
Fix merging of :message/message-persisted event

rasom:
Previously `:message/message-persisted` was dispatched for each incoming
message because its argument was not a vector. The event was renamed to
`:message/messages-persisted`.

status: ready


mandrigin (approved):
good catch!

yenda (approved):

PR #7301, opened 2019-01-18
[fix] remove mnemonic from current-account map

yenda:
current-account is duplicated in `:qr-modal` key when tapping on `share my profile` button in the Profile screen

this removes the mnemonic from the duplicated data so that it does persist there after user backed up his mnemonic

this was only happening until logging out/restarting the app

status: ready

### DIFF
--- a/src/status_im/ui/screens/profile/user/views.cljs
+++ b/src/status_im/ui/screens/profile/user/views.cljs
@@ -267,7 +267,7 @@
         [react/view (merge action-button.styles/actions-list
                            styles/share-contact-code-container)
          [button/secondary-button {:on-press            #(re-frame/dispatch [:navigate-to :profile-qr-viewer
-                                                                             {:contact current-account
+                                                                             {:contact (dissoc current-account :mnemonic)
                                                                               :source  :public-key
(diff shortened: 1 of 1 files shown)


corpetty (approved):
Great catch @yenda!

vitvly (approved):

churik:
@yenda would be nice if you can share steps how to verify it. Thanks!

asemiankevich:
> @yenda would be nice if you can share steps how to verify it. Thanks!

i believe we can use http://localhost:4567/ re-frisk tool for that , right @flexsurfer ?

asemiankevich:
@yenda please rebase, builds are 6 days old

yenda:
@asemiankevich re-frisk would work but only in development. I don't think this can be tested except maybe by logging it which would make the issue worst. I would say that as long as it doesn't break the app it's ok?

mandrigin:
@yenda firstly, thanks for noticing and fixing it!
I think yeah, if the app isn't broken, it should be fine, the fix is very trivial.

asemiankevich:
alright lets merge it then

PR #7312, opened 2019-01-21
Disable sending message

cammellos:
Seems like the current version of status-go does not contain some of the
code that was in the previous, so disabling this for now until we upgrade.

status: ready


mandrigin (approved):
look good!

rasom (approved):

PR #7317, opened 2019-01-21
Add logic to cancel older running builds in Jenkins

pedropombeiro:

### Summary:

This PR addresses https://www.pivotaltracker.com/story/show/163367849.

### Review notes (optional):

Test build canceled by newer build: https://ci.status.im/job/status-react/job/prs/job/windows/job/PR-7317/18/

status: ready

### DIFF
--- a/ci/common.groovy
+++ b/ci/common.groovy
@@ -1,4 +1,7 @@
 import groovy.json.JsonBuilder
+import hudson.model.Result
+import hudson.model.Run
+import jenkins.model.CauseOfInterruption.UserInterruption
 
 def version() {
   return readFile("${env.WORKSPACE}/VERSION").trim()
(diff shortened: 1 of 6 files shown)


mandrigin (approved):
Looks good! But the PR title is misleading a bit.

jakubgs (approved):
That's a lot of whitespace around that call.
Great work.

pedropombeiro:
> That's a lot of whitespace around that call.

I thought it would be good to have a visual marker around the point where builds are canceled, but you're right, it's not really needed.

jakubgs:
Just a nitpick, no worries.

pedropombeiro:
Should the child also try to cancel the upstream job (so that that one cancels the children)?

jakubgs:
Wouldn't that be handled by the previous job? Or are you worrying about some kind of race condition?

ci/common.groovy
```diff
@@ -29,6 +32,25 @@ def getBuildType() {
   return params.BUILD_TYPE
 }
 
+@NonCPS
+def abortPreviousRunningBuilds() {
+  Run previousBuild = currentBuild.rawBuild.getPreviousBuildInProgress()
+
+  while (previousBuild != null) {
```
jakubgs, line 39: I guess there's no harm in iterating and making sure all are killed.

pedropombeiro:
Yeah, ideally it would be handled by the upstream job directly, but I don't see where I'd do that.

jakubgs:
>but I don't see where I'd do that.

But it would have already done that when it started. Meaning, if the newest job is trying to kill the still running previous job, that would mean to me that the job before the previous job(if there ever was one) was already killed by the previous job.

Unless some kind of race condition happened I guess.

pedropombeiro:
Nevermind, I thought this job was similar to the `Jenkinsfile.combined` one, where the child jobs are started in a `parallel` entry, but apparently not. So right now the only delay that exists is if the build queue is too large for the new builds to start executing and start killing running builds, but I guess that can be improved in the future.

jakubgs:
>So right now the only delay that exists is if the build queue is too large for the new builds to start executing and start killing running builds, but I guess that can be improved in the future.

Good point, but this is already a big improvement that should shorten the queues, so this scenario is less likely.

PR #7321, opened 2019-01-22
Fix reproducible builds in Docker images

pedropombeiro:

### Summary:

This PR ensures that Docker image tags are unique and can be referenced in the future as images are updated. It achieves this by calculating a hash based on the tool versions that the Docker image contains (new script `scripts/gen-deps-hash.sh`).

It also ensures that node and yarn versions are exactly what we expect. This already allowed us to notice misconfiguration in Jenkins/Docker images.

### Review notes (optional):

The changes to .env files are just a alphabetical sorting, which allows finding duplicate entries.

status: ready


pedropombeiro:
@mandrigin @jakubgs please approve this PR to test if the move to TO TEST will do the right thing (overwrite the initial status of `Mobile e2e tests`)

jakubgs (dismissed):

mandrigin (dismissed):
great job!

jakubgs:
>great job!

A truly TITANIC effort.

pedropombeiro:
> > great job!
> 
> A truly TITANIC effort.

Haha, yeah man, you wouldn't believe it!

pedropombeiro:
Why the hell does it show my face on the status??

jakubgs:
>Why the hell does it show my face on the status??

It became sentient and fell in love with you. Obviously.

pedropombeiro:
Instead of creating a new bot, I'll move this inside the `status-github-bot` so that the logic stays in the same place as the rest of the script, and not spread around. The splitting of `status-github-bot` is another story and will need to be done in its own time.

pedropombeiro:
OK, looks like the `Mobile e2e tests` status got replaced by Adan's script as expected, but it should be blocking the merge and it isn't. Looking into it.

pedropombeiro:
@mandrigin Looks like the `Require status checks to pass before merging` option is not enabled for `status-react`, so people can still merge even while status still display `pending`. Maybe this is an escape hatch so that we don't have to wait for builds to finish in order to merge tiny things, wdyt? If that's the case, then we should probably add a way to disable running tests from the git commit message, and then enable the option for the repo.
Or maybe we decide to live with it as it is and see if it's enough of a deterrent for people to not merge inadvertently.

pedropombeiro:
@mandrigin @jakubgs after merging the branch with `develop` there were two apps which didn't report back new status (`GPG` and `Mobile e2e tests`). This might not be an issue if we can still use the command line merge script, but I imagine this is protected by a git hook on the GH side.

jakubgs:
I can make those not required.

pedropombeiro:
Yup, confirmed:

```
remote: error: GH006: Protected branch update failed for refs/heads/develop.
remote: error: 15 of 19 required status checks have not succeeded: 10 expected and 4 pending.
To github.com:status-im/status-react.git
 ! [remote rejected]     develop -> develop (protected branch hook declined)
error: failed to push some refs to 'git@github.com:status-im/status-react.git'
```

We have two choices here. I could have the bot trigger an e2e build every time the branch is pushed and it is in the IN TEST column, as well as fix the GPG plugin, or we can make the checks not required (less safe alternative). cc @mandrigin

mandrigin:
I would make it less safe but more convenient right now. If there will be more incidents — we can reassess.

There aren’t too many incidents tbh, so let’s not go draconian :fire:

pedropombeiro:
> I would make it less safe but more convenient right now. If there will be more incidents — we can reassess.
> 
> There aren’t too many incidents tbh, so let’s not go draconian fire

Sounds good. A separate question is whether we should trigger e2e automated tests every time a PR is force pushed (e.g. after being rebased) if it is in the TO TEST column. It sounds like the right thing to do, to find out if rebasing introduced any issues, but I guess we need to monitor if we start getting too big a build queue on e2e tests.

pedropombeiro:
@mandrigin completely off topic, but as a test in this PR I tried to ensure that we have the correct Node/yarn versions in `make prepare` (instead of just checking for the presence of the tools). This uncovered the fact that for Windows and Linux builds we are using a hardcoded Docker image tag, which won't play well with reproducible builds. I suggest converting this PR to doing the required work to make those builds reproducible (move the tag to `.TOOLVERSIONS` so that as we build a historic commit, it uses the right Docker image).

pedropombeiro:
@jakubgs I don't suppose we can read the value from the file to use here? https://github.com/status-im/status-react/blob/feature%2Ftest-e2e-block-bot/ci/Jenkinsfile.linux#L6

jakubgs:
Okay, I've made the `GPG` and `Mobile e2e tests` not mandatory.

pedropombeiro:
Ping @mandrigin and @jakubgs for new review

docker/android/Makefile
```diff
@@ -17,11 +17,14 @@ SDK_PLATFORM_VERSION=$(call  __toolversion, android-sdk-platform)
...
 # WARNING: Remember to change the tag when updating the image
-IMAGE_TAG = 1.1.0
+BASE_IMAGE_TAG = $(shell cd ../base && make get-image-tag)
+DEPS_HASH = $(shell $(GIT_ROOT)/scripts/gen-deps-hash.sh -b $(BASE_IMAGE_TAG) -d android-ndk -d android-sdk -d android-sdk-platform -d android-sdk-build-tools)
+IMAGE_TAG = 1.1.0-$(DEPS_HASH)
 IMAGE_NAME = statusteam/status-build-android:$(IMAGE_TAG)
 
 build: $(ANDROID_NDK_ARCHIVE) $(ANDROID_SDK_ARCHIVE)
 	docker build \
+	    --build-arg="BASE_IMAGE_TAG=$(BASE_IMAGE_TAG)" \
```
jakubgs: Wrong indent.

ci/Jenkinsfile.android
```diff
@@ -2,7 +2,7 @@ pipeline {
   agent { 
     docker {
       label 'linux'
-      image 'statusteam/status-build-android:1.1.0'
+      image 'statusteam/status-build-android:1.1.0-16a42e06' /* remember to keep this up-to-date with the value in docker/android/Makefile */
```
jakubgs: You canmove this comment to before the line instead of in the line, easer to read.  Maybe make it a `WARNING:`?

docker/android/Makefile
```diff
@@ -17,11 +17,14 @@ SDK_PLATFORM_VERSION=$(call  __toolversion, android-sdk-platform)
 SDK_BUILD_TOOLS_VERSION=$(call  __toolversion, android-sdk-build-tools)
 
 # WARNING: Remember to change the tag when updating the image
-IMAGE_TAG = 1.1.0
+BASE_IMAGE_TAG = $(shell cd ../base && make get-image-tag)
```
jakubgs, line 20: Oh, I see. So the hash depends on the specific versions we get for that particular image. Interesting solution.

docker/base/Makefile
```diff
@@ -4,9 +4,13 @@ GIT_COMMIT = $(shell git rev-parse --short HEAD)
 GIT_ROOT = $(shell git rev-parse --show-toplevel)
 
 # WARNING: Remember to change the tag when updating the image
-IMAGE_TAG = 1.1.0
+DEPS_HASH = $(shell $(GIT_ROOT)/scripts/gen-deps-hash.sh -d leiningen -d node -d yarn)
```
jakubgs, line 7: I'm wondering if there could be an easy way to track this via `toolversion` script. Something like:
```bash
scripts/getn-deps-hash.sh --start
scripts/toolversion leiningen
scripts/toolversion node
scripts/toolversion yarn
scripts/getn-deps-hash.sh --end --gethash
```
So we don't have to maintain the list of packages to get hash for manually. But that could be a future improvement.

jakubgs (approved):
Really cool idea to improve how deterministic our builds are.

rasom (approved):

cammellos (approved):

PR #7322, opened 2019-01-22
use separate jobs for e2e builds, disable aborting for non-pr builds

jakubgs:
Our most recent Nightly builds started failing due to a recent change that aborts previously running jobs.
https://github.com/status-im/status-react/pull/7317
Example: https://ci.status.im/job/status-react/job/combined/job/mobile-ios/5615/console
```
Aborted by unknown
00:00:20.416 Sending interrupt signal to process
```
This was intended only for PRs, but nightly builds use the same Jenkins jobs for regular and e2e builds, which causes one to abort the other.

This is a two-fold fix:
* Added separate Jobs for nightlies with `-e2e` suffix to make it clearer which one is which
* Added an `if` statement to avoid abording when build is not a PR build.

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -40,13 +40,13 @@ pipeline {
           ios = cmn.buildBranch('status-react/combined/mobile-ios')
         } } }
         stage('iOS e2e') { steps { script {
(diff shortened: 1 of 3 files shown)


mandrigin (approved):
awesome! towards a more usable CI system! 🚀

cammellos (approved):

pedropombeiro (approved):

PR #7335, opened 2019-01-24
Stop sending seen messages

rasom:
Currently, these messages are not used and should be properly revisited
later. It only generates extra traffic on discovery topic.

status: ready


mandrigin (approved):
lgtm

yenda (approved):

PR #7385, opened 2019-01-31
remove wallet address auto generation

gravityblast:
In this PR we are removing the automatic derivation of the wallet address from the whisper public key, so that a user can send a different wallet address that is not derived from the same key used for chat.
This change is needed to interact with keycard accounts and in the future for accounts with chat and wallet keys decoupled. 

In the current app we create contacts in 2 ways:

1 - When Bob adds Alice, Alice receive Bob's contacts info with name, wallet address, etc.. and creates a new entry in her contact list.

2 - When Bob adds Alice, Bob creates a new entry in the contact list with data derived from Alice's public key.

If Alice is using a keycard account, Bob will end up generating a new contact entry with the wallet address derived from Alice's public key, which would be the wrong address. 
Alice is instead using a whisper key derived at `m/43'/60'/ 1581'/0/0` as described in (EIP1581)[https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1581.md], and a wallet key using the standard BIP44 path.

For this reasons, we are removing the automatic derivation of the wallet address and instead using only the address received from the other contact.

After this changes, transactions will be available only to contact that contains a wallet address.

**Functional**
- in-chat transactions 1-1 chats
- wallet / transactions

### DIFF
--- a/src/status_im/chat/commands/core.cljs
+++ b/src/status_im/chat/commands/core.cljs
@@ -35,22 +35,23 @@
   [type]
   (keyword (str (protocol/id type) "-button")))
 
-(defn- contact->address [contact]
-  (str "0x" (db/public-key->address contact)))
+(defn- contact->address [all-contacts pub-key]
+  (get-in all-contacts [pub-key :address]))
 
 (defn add-chat-contacts
   "Enrich command-message by adding contact list of the current private or group chat"
-  [contacts {:keys [public? group-chat] :as command-message}]
+  [all-contacts contacts {:keys [public? group-chat] :as command-message}]
   (cond
     public? command-message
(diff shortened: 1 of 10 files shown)


dmitryn (approved):

goranjovic (approved):

antdanchenko:
@gravityblast why this was merged without testing?

mandrigin:
will be reverted in https://github.com/status-im/status-react/pull/7413

gravityblast:
sorry @antdanchenko @mandrigin ,  I only check the "Jenkins Builds" tests were passing, I'll ask nowfor the right process and follow it in the next PR that I'll re-open

Issue #3451, opened 2018-02-26
Fetch messages in background when Push Notification is received


yenda:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story

- As a user, I expect to see the unread messages in the when I click the push notification. Currently, I have to wait for sync to kick in and retrieve the messages from the network.
- As a Status contributor, I want to have a better sense of what the Sent/Received ratio is. If the user doesn't open the app for 24h after receiving a message, that message will never be counted as received, which is fake news!

### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Features

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*:

The application could run in the background upon reception of a notification to fetch recent messages, decrypt them and consequently send the signal to Mixpanel.

#### Expected behavior
[comment]: # (Describe what you expected to happen.)
1. User receives message and PN;
1. User loses internet connectivity;
1. After a minute, user clicks on PN;
1. App opens and message is ready to be read. 

#### Actual behavior
[comment]: # (Describe what actually happened.)
1. User receives message and PN;
1. User loses internet connectivity;
1. After a minute, user clicks on PN;
1. App opens and message doesn't appear until foreground sync has completed. 

### Solution
[comment]: # (Please summarise the solution and provide a task list on what needs to be fixed.)
*Summary*: 
It could be done using `background app refresh` feature on iOS which allows an app to wake up for 30 sec in the background upon reception of a notification

> When a background update notification is delivered to the user's device, iOS wakes up your app in the background and gives it up to 30 seconds to run. In iOS, the system delivers background update notifications by calling the application:didReceiveRemoteNotification:fetchCompletionHandler: method of your app delegate. Use that method to initiate any download operations needed to fetch new data. Processing remote notifications in the background requires that you add the appropriate background modes to your app.

Similar feature should exist on Android.

1. The notification sent by the client should be a [data-only notification](https://firebase.google.com/docs/cloud-messaging/concept-options#data_messages), otherwise, it will only call app-specific code when it is tapped by the receiver or the app is in the foreground.
1. Upon getting called `onMessageReceived()` (Android)/`didReceiveRemoteNotification()` (iOS), the app should trigger a network sync to retrieve messages, in a platform-specific way that allows for long-running tasks, and create a [local notification](https://github.com/evollu/react-native-fcm#build-custom-push-notification-for-android).

[...]

PR #7389, opened 2019-01-31
Fetch messages from mailserver upon PN if db is unlocked. Part of #3451

pedropombeiro:

### Summary:

This PR improves delivery of messages by triggering a request to the mailserver every time a push notification arrives (therefore there is less risk that a message won't arrive because the user didn't open the app in the last 24 hours). It does so for the target chat ID and only for the last 15 minutes (this can be observed on the logs by filtering for `fetch-history`).

#### Platforms (optional)
- Android
- iOS

**Functional**
- 1-1 chats

### Steps to test:
- Open Status
- Put the app in the background
- Send a message from another client to this account
- Wait for the PN to arrive
- Put the phone in airplane mode
- Tap the notification
- Check that the message has arrived even though the phone is in airplane mode

status: ready

### DIFF
--- a/src/status_im/notifications/core.cljs
+++ b/src/status_im/notifications/core.cljs
@@ -18,6 +18,7 @@
 (def ^:private pn-message-id-hash-length 10)
 (def ^:private pn-pubkey-hash-length 10)
 (def ^:private pn-pubkey-length 132)
+(def ^:private pull-recent-messages-window (* 15 60))
 
 (when-not platform/desktop?
 
@@ -217,7 +218,7 @@
                (vals (:chats db)))))
 
   (fx/defn handle-on-message
-    [{:keys [db] :as cofx} decoded-payload {:keys [force]}]
+    [{:keys [db now] :as cofx} decoded-payload {:keys [force]}]
     (let [view-id            (:view-id db)
           current-chat-id    (:current-chat-id db)
           app-state          (:app-state db)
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

src/status_im/notifications/core.cljs
```diff
@@ -216,17 +217,21 @@
...
-         (assoc-in db [:push-notifications/stored (:to rehydrated-payload)]
-                   (js/JSON.stringify (clj->js rehydrated-payload)))
-         :notifications/display-notification
-         {:title           (i18n/label :notifications-new-message-title)
-          :body            (i18n/label :notifications-new-message-body)
-          :decoded-payload rehydrated-payload}})))
+      (merge
+       (when (and (= (.-length from) pn-pubkey-length)
+                  (show-notification? cofx rehydrated-payload))
+         {:dispatch [:mailserver/fetch-history from (- (quot now 1000) pull-recent-messages-window)]})
```
cammellos, line 233: We should be a bit careful here, you want to make sure that we are not overriding an earlier value of from.

Say we just logged in, and `from` is -1 day, and we haven't yet fetched stuff from the mailserver, we want to make sure that `from` is `(min your-from current-from)`, do you see what I mean? (sorry if it's a bit confusing)

pedropombeiro: If I understand correctly, this is what you're suggesting:

``` clojure
(let [fetch-pn-message-window (- (quot now 1000) pull-recent-messages-window)
      last-updated            (max 1 (get-in db [:account/account :last-updated]))]
  {:dispatch [:mailserver/fetch-history from (min last-updated fetch-pn-message-window)]}))
```

pedropombeiro: OK, fixed.

cammellos (approved):

pedropombeiro:
@churik I noticed this PR went back to TO TEST last week. Is there anything missing on it, or was it just that there was higher priority stuff to test?

Serhy:
All works as described, no regressions founds.

PR #7392, opened 2019-01-31
Add golang to `.TOOLVERSIONS` and fix md5 calculation

pedropombeiro:

### Summary:

This PR adds golang to `.TOOLVERSIONS` and fixes the md5 calculations, such as `qt5_md5`.

status: ready


pedropombeiro:
Ping for reviews

jakubgs (approved):
I think you are right, it might be a good idea to do a separate `desktop` image to move the go installation to a shared image. But this is fine as it is for now.
Although, if go doesn't take too much space, maybe it would be just simpler to add it to the base image.

mandrigin (approved):

PR #7398, opened 2019-02-01
Fix null is not an object error for invalid URLs

### LINKED ISSUE #6291: Null is not an object error is shown when "http:/" as url in Status browser
### Description

*Type*: Bug

*Summary*: 

When a user made a mistake and typed "http:/" instead of "http://", the app can't parse the url and error is shown. When closed, everything works fine.

#### Expected behavior
I'd suggest showing the default error page instead:

#### Actual behavior
Error when "http:/" or "http:///", etc. url: 

### Reproduction

- Type "http:/google.com" in Status browser

### Additional Information
* Status version: nightly, 12 Oct
* Operating System: all

dhamaniasad:
fixes #6291 

### Summary:
The null is not an object error occurred in the `check-if-phishing-url` function that's called whenever a new URL is opened, because the `http/url-host` function would return null when the URL entered was invalid. This null (history-host) was then passed to `dependencies/phishing-detect`, which is a JS library, which has a `str.split` to split a URL into its constituent parts. This `.split` being called on `null` caused the application to error out. I fixed it by adding a check to see if `history-host` is not null. 

### Review notes (optional):
My previous PR failed due to linting errors and a missing GPG signature. So I'm making another PR fixing those issues.

#### Platforms (optional)
- Android
- iOS

### Steps to test:
- Open Status
- Open Status Browser, for example from Profile -> ENS Names page
- Type in invalid URL like http:/ojkdkllkjasjd.com
- You should see ERR_NAME_NOT_RESOLVED instead of the previous null is not an object error

status: ready

### DIFF
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -81,7 +81,7 @@
 
 (defn check-if-phishing-url [{:keys [history history-index] :as browser}]
   (let [history-host (http/url-host (try (nth history history-index) (catch js/Error _)))]
-    (assoc browser :unsafe? (dependencies/phishing-detect history-host))))
+    (cond-> browser history-host (assoc :unsafe? (dependencies/phishing-detect history-host)))))
 
 (def ipfs-proto-code "e3")
 (def swarm-proto-code "e4")


jeluard (approved):

jeluard:
@dhamaniasad Thanks for your PR!

src/status_im/browser/core.cljs
```diff
@@ -80,7 +80,9 @@
 
 (defn check-if-phishing-url [{:keys [history history-index] :as browser}]
   (let [history-host (http/url-host (try (nth history history-index) (catch js/Error _)))]
-    (assoc browser :unsafe? (dependencies/phishing-detect history-host))))
+    (if history-host
+      (assoc browser :unsafe? (dependencies/phishing-detect history-host))
+      (assoc browser :web-view-error? true))))
```
flexsurfer: because we don't use `:web-view-error?` anywhere, and there is no spec for it, better to return just browser object

```
(cond-> browser
 history-host (assoc :unsafe? (dependencies/phishing-detect history-host))
```

flexsurfer (changes requested):

src/status_im/browser/core.cljs
dhamaniasad: This image that was mentioned in the issue seems to be displayed when `:web-view-error?` is true, `unsafe?` would display the phishing warning.

flexsurfer: i can't find the usage of `:web-view-error?` in the codebase, could you point where :web-view-error? is used

flexsurfer: i just suggest to do not assign :web-view-error?, in my code example phishing-detect will not be called if history-host is null

dhamaniasad: If you check [this link](https://github.com/status-im/status-react/blob/develop/test/appium/tests/atomic/dapps_and_browsing/test_browsing.py#L23), `Unable to load page` is supposed to be the error displayed for invalid URLs. And [here](https://github.com/status-im/status-react/blob/develop/translations/en.json#L590) is where that message is defined. 

Also what ends up happening with your example is that `history-host` is undefined on first browser load, because `history` is empty at that point. So it doesn't return the browser and that crashes the page. 

But let me know if I got you wrong, I'm very new to Clojure but I'll get things sorted. 

Do you want me to move the invalid URLs check to another function?

dhamaniasad:
```clojure
(defn check-if-phishing-url [{:keys [history history-index] :as browser}]
  (let [history-host (http/url-host (try (nth history history-index) (catch js/Error _)))]
    (if history-host

    (assoc browser :unsafe? (dependencies/phishing-detect history-host))
    (assoc browser :unsafe? false))))
```
This code has the same effect without using `web-view-error`.

flexsurfer: to avoid null error we just need to do not invoke `phishing-detect`, and also we don't know if it's safe or not, so better to do nothing, something like

```
(if history-host
 (assoc browser :unsafe? (dependencies/phishing-detect history-host)) 
  browser))))
```

but a better way to write this code is 

```
(cond-> browser
 history-host (assoc :unsafe? (dependencies/phishing-detect history-host))
```
thanks for your effort btw ;) clojure ftw :)

dhamaniasad: Got it, thanks. I've pushed a commit to fix this issue.

flexsurfer (approved):

src/status_im/browser/core.cljs
flexsurfer: thanks!

dhamaniasad: Seems like I forgot to sign the commit. Should I do a force push?

jeluard: @dhamaniasad No worries you don't need to sign yourself

jeluard: @dhamaniasad We will handle it from there. In the meantime you might want to look at some other bounties? :)

jeluard: @dhamaniasad Just make sure you squash your commits

dhamaniasad: Do you want me to squash my commits in this PR?

jeluard: @dhamaniasad Please yes :)

dhamaniasad: @jeluard Done.

jeluard:
@dhamaniasad Could you rebase with latest develop? An issue on our side preventing the build has been fixed.
Sorry for the inconvenience.

dhamaniasad:
@jeluard Done

goranjovic (approved):

PR #7404, opened 2019-02-04
[#6597] Add a map component

### LINKED ISSUE #6597: Add a map component
### User Story

As a developer, I want a status [extensions](https://status.im/extensions/) component allowing to show map details.

 This must leverage a webview that will be exposed as a standalone component.
How data is passed back and forth is open to discussion.
Special care must be taken to security and privacy.

### Description

*Type*: Feature

tbenr:
fixes #6597

[current test extension](https://status.im/extensions/play.html?hash=QmXX29Q92NrV3tSQkcwwrFuybtocDbse338geNDWxnZjmj)

[current demo](https://youtu.be/DeEkz7EX4cQ)

In my opinion, device geolocalization should be managed outside the component: a new extension query which gives the device geolocation.


jeluard:
@tbenr Please consider that `webview` will be deprecated from ReactNative and that we will have to rely on https://github.com/react-native-community/react-native-webview

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/react-native-community/react-native-webview">react-native-community/react-native-webview</a></strong></div><div>React Native Cross-Platform WebView. Contribute to react-native-community/react-native-webview development by creating an account on GitHub.</div></blockquote>

tbenr:
@jeluard hum.. i'm actual relying on `react-native-webview-bridge` (on which all dapp-browser is based as far as I can see) is it going to be migrated to `react-native-webview` soon?

jeluard:
@tbenr Right you are good then :D

jeluard:
@tbenr Actually would be better to use directly `webview` as we plan to deprecate the bridge soon.

tbenr:
@jeluard i get prepared being compatible to the standard `react-native-webview`, just avoid using `sendToBridge` and `onBridgeMessage`.

jeluard:
@tbenr That seems about right. Do you mind going ahead and use `react-native-webview` directly?

tbenr:
@jeluard I could but I have to add new react dependency 'cose i don't see `react-native-webview` in package.json .. I'm a bit warried to start struggling with package management and dependencies...

jeluard:
@tbenr You have to try one day :)

tbenr:
@jeluard I will! let me leave it at the end of this PR :)
BTW what do you think about extension<->webview communication?
make sense to you to subscribe the mapview to the key store to get info from extension?

jeluard:
@tbenr Do you mean what you did in the test extension? If so, yes!

tbenr:
@jeluard I think I need official `react-native-webview`.
I did `yarn add react-native-webview` and that's ok but when I run `react-native link react-native-webview` I gotthis error:

image

is there someone who can help? any hint on how to do manual link?

jeluard:
@tbenr I had the same issue and ended up [manually](https://engineering.brigad.co/demystifying-react-native-modules-linking-964399ec731b) linking it.

tbenr:
@jeluard i managed ios build via cocoapods. but on the android side is a mess, have to switch to buildtools version 28, but now:
image

even if I fix it I doubts it will ever build in CI\CD pipeline :-)

jeluard:
@tbenr I had the same but it was fixed when updating some sdk version

tbenr:
@jeluard building now 🥇  :)

tbenr:
@jeluard still working on MapComponent->Extension communication.
btw I updated my local commit tp react-native-webview 5.2.1, the same you are using in your PR #7515

tbenr:
MapComponent->Extension communication is done but it is not working on my physical android 8 🤦‍♂️ 

OK the issue was: https://github.com/react-native-community/react-native-webview/issues/323

@jeluard final PR soon.

jeluard:
Looks like some comments/println could be removed :)
Can you detail what you mean with `MapComponent->Extension communication` ?
Also, any consideration with regards to security?

tbenr:
@jeluard I pushed the latest version:
- its cleaner 😄 (no prints..)
- now resources are loaded locally (js css as assets).
  - iOS: :white_check_mark: (in the **project.pbxproj**  added a folder in Resources directory named `mapview` pointing to _resources/mapview_)
  - android: :white_check_mark: (modified **build.gradle** adding a new asset directory pointing to _resources/mapview_)

the `webview->extension` communication is based on `postMessage`.

`webview_javascript::postMessage(JSON.stringify(message))` **->** `reagent_map_component::on-message(message)` **->** `extension::on-change(json->clj(message))`.

so the app state only changes based on that flow. The message is json `message` containing geo coordinates.

the `extension->webview` is done via `injectJavaScript`: it runs `update(...)` inside webview context.

this is the only way for webview to get info from the app.
~~_Maybe I have to add checks to avoid js injection there to avoid extension to be able to run js code within the webview_~~ :white_check_mark:
**geolocationEnabled** is not enabled so user position cannot be detected.

in the last push I added a `on-should-start-load-with-request` which enforce all urls to be local `file://..`. Everything else is opened to external browser (ie openstreetmap\leaflet links).

assuming react-native-webview doesn't have any security issue, webview runs local html\js\css (assumed trusted). But even if webview runs compromised js, it could send out only user selected geo coordinates... i'm trying to find other consideration to go through...ideas?

tbenr:
@jeluard do we need to consider credits\license topics for openstreetmap\leafletjs?

src/status_im/chat/commands/impl/map.cljs
```diff
@@ -0,0 +1,93 @@
+(ns status-im.chat.commands.impl.map
```
jeluard: Could this be a sub ns of the extension one?

jeluard (commented):
Is the user location accessed in some way?

Goof question about leaflet license, worth checking if we can embed the code directly as is.

src/status_im/chat/commands/impl/map.cljs
tbenr: ok!

tbenr:
@jeluard 
> Is the user location accessed in some way?

I'm not using any device-location related function on leaflet, so user position should be never asked by the map view.

From underlining webview prespective, as per documentation, to be able to get user geoloc you need to set [`:geolocation-enabled`](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#geolocationenabled) explicitly to `true`, and I'm not. But this is for android only. I need to check what happens in iOS if you try to run some JS trying to get user loc.

In the current Status embedded browser is there any additional protection to avoid user position leakage?

> Goof question about leaflet license, worth checking if we can embed the code directly as is.

the license is quite easy: https://github.com/Leaflet/Leaflet/blob/master/LICENSE
we can embed binaries if we report their copyright, who we should ping from Status who can check?

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/Leaflet/Leaflet">Leaflet/Leaflet</a></strong></div><div> :leaves: JavaScript library for mobile-friendly interactive maps - Leaflet/Leaflet</div></blockquote>

src/status_im/chat/commands/impl/map.cljs
tbenr: what about making it a a common status component under `status-im.ui.components`?

jeluard:
I’d rather have it isolated in extensions for now.

tbenr:
@jeluard update on geolocalization `navigator.geolocation.getCurrentPosition();`:
- android: it is always denied when not specifying `:geolocation-enabled true`.
- IOs: it asks permission to access position, no way to block it AFAIK.

I notice a strange behavior on a physical android (galaxy S9 on android 9.0): it seems that many webviews loading concurrently (ie: opening a chat with many maps) some times leafletjs javascript doesn't run correctly, resulting in a map not zoomed on the marker.

I double checked: the javascript is called always in the same way but sometime the result is different. 

image

image

I'm trying to understand what's happening to find a solution\workaround

tbenr:
update: it seems there are similar issues on leaflet related to `fitBounds()`, the method I use. searching for a workaround.

jeluard:
Are you adding location access features? I'd rather keep this component lean and expect users to provide the location to be displayed somehow.

tbenr:
> Are you adding location access features? I'd rather keep this component lean and expect users to provide the location to be displayed somehow.

@jeluard that is my intention. I was trying a way to block webview location access _a priori_. Not possible on IOS.

jeluard:
@tbenr I feel it's better to add this as an extra event, if needed

tbenr:
that's ok. the PR is complete btw. I need only to workaround the leaflet's `fitBounds()` issue (should be fixed inside `resources/mapview/mapview.html` only)

jeluard:
Nice!

jeluard (approved):

flexsurfer:
Wow! Amazing and huge work! Maybe was discussed already, just wondering why can't we use webview-bridge?

jeluard:
@flexsurfer We could but we are trying to get rid of it. Also it's a first step towards webview integration. WDYT?

flexsurfer (approved):

jeluard:
@tbenr Build fails. Can you run tests locally?

tbenr:
@jeluard ahh i know what's going on.. let me fixit.

tbenr:
@jeluard @flexsurfer ~HELP! i'm getting crazy on this stupid thing.~

~only when compiling for prod, it seems that it is not finding the namespace definition (from the new file map.cljs). Am I Missing something to be done when adding new cljs\namespace?~

image

**SOLVED**: I just realize that namespaces are like java packages and must follow directory structure!
thanks Julien for point me to the right direction.

tbenr:
fixing desktop build too... hiding react-native-webview module in js dependencies... this should be the final one 😅

jeluard:
@tbenr Looks like desktop is still failing?

tbenr:
@jeluard ..seems only macos has failed with strange error.. trying to build it locally..

flexsurfer:
@tbenr maybe try to rebase onto latest develop

tbenr:
🎆

jeluard:
@tbenr Thanks!

asemiankevich:
can't access all search results, search results list is not scrollable. For example if i check on web for /moscow, i see a bunch of results and See More option

image

in Status app i can't access See More option, as well as scroll the list of results.
image

@jeluard @tbenr @flexsurfer what do you think - could this be fixed in this PR or requires separate issue?

also, white location link is hardly visible for incoming location messages

image

jeluard:
@asemiankevich The scope of the issue is only the map, the rest is extension dependent.

tbenr:
@asemiankevich @jeluard 

the hardly-visible link maybe is something that can be fixed.. I used the same style already used in `link` component, that is:

```
(defn link [{:keys [uri]}]
  [react/text
   {:style    {:color                colors/white
               :text-decoration-line :underline}
    :on-press #(re-frame/dispatch [:browser.ui/message-link-pressed uri])}
   uri])
```

my `map-link` is:
```
(defn map-link
  "create a link-view which opens native map/navigation app with marker and label"
  [{:keys [text lat lng]}]
  (let [uri (cond
              platform/ios? (str "http://maps.apple.com/?q=" (js/encodeURIComponent text) "&ll=" lat "," lng)
              platform/android? (str "geo:0,0?q=" lat "," lng "(" (js/encodeURIComponent text) ")")
              :else (str "http://www.openstreetmap.org/?mlat=" lat "&mlon=" lng))]
    [react/text
     {:style    {:color                colors/white
                 :text-decoration-line :underline}
      :on-press #(.openURL react/linking uri)}
     text]))
```

**:color                colors/white** is the issue :)

jeluard:
@tbenr Oh nice so maybe this one could be sneaked in?

tbenr:
@jeluard yeah.. i'd suggest to support style coming from the extension, so it will be an extension decision to choose the appropriate style. Agree?

jeluard:
@tbenr Makes sense! Still it would be nice to have good default.
BTW can't `map-link` just reuse link?

tbenr:
@jeluard The idea was to avoid the dialog "open in status or android\ios". The links are designed to open native apps. Maybe `link` can support the choice to avoid the dialog and then i can reuse it...

jeluard:
Ok makes, for another PR :D

tbenr:
@jeluard yeah.. but easy stuff.. I'll do another PR once this is merged... so to keep things clean :)

PR #7409, opened 2019-02-04
Dont count 1-to-1 topics if not enabled

cammellos:
We count 1-to-1 filters only if partitoned topic is enabled, otherwise no message is fetched when one-to-one are present, also fixes an issue with error handling, but does not fix the root cause (`Filter is nil`).

Current nightly does not fetch messages if a one-to-one chat is present.

status: ready


rasom (approved):

mandrigin (approved):

yenda (approved):

cammellos:
closing as already tested and included

PR #7416, opened 2019-02-04
remove wallet address auto generation

gravityblast:
In this PR we are removing the automatic derivation of the wallet address from the whisper public key, so that a user can send a different wallet address that is not derived from the same key used for chat.
This change is needed to interact with keycard accounts and in the future for accounts with chat and wallet keys decoupled.

In the current app we create contacts in 2 ways:

1 - When Bob adds Alice, Alice receive Bob's contacts info with name, wallet address, etc.. and creates a new entry in her contact list.

2 - When Bob adds Alice, Bob creates a new entry in the contact list with data derived from Alice's public key.

If Alice is using a keycard account, Bob will end up generating a new contact entry with the wallet address derived from Alice's public key, which would be the wrong address.
Alice is instead using a whisper key derived at m/43'/60'/ 1581'/0/0 as described in (EIP1581)[https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1581.md], and a wallet key using the standard BIP44 path.

For this reasons, we are removing the automatic derivation of the wallet address and instead using only the address received from the other contact.

After this changes, transactions will be available only to contact that contains a wallet address.

Areas that maybe impacted (optional)
Functional

in-chat transactions 1-1 chats
wallet / transactions

### DIFF
--- a/src/status_im/chat/commands/core.cljs
+++ b/src/status_im/chat/commands/core.cljs
@@ -35,22 +35,23 @@
   [type]
   (keyword (str (protocol/id type) "-button")))
 
-(defn- contact->address [contact]
-  (str "0x" (db/public-key->address contact)))
+(defn- contact->address [all-contacts pub-key]
+  (get-in all-contacts [pub-key :address]))
 
 (defn add-chat-contacts
   "Enrich command-message by adding contact list of the current private or group chat"
-  [contacts {:keys [public? group-chat] :as command-message}]
+  [all-contacts contacts {:keys [public? group-chat] :as command-message}]
   (cond
     public? command-message
-    group-chat (assoc command-message :contacts (map contact->address contacts))
-    :else (assoc command-message :contact (contact->address (first contacts)))))
+    group-chat (assoc command-message :contacts (map #(contact->address all-contacts %) contacts))
+    :else (assoc command-message :contact (contact->address all-contacts (first contacts)))))
 
 (defn enrich-command-message-for-events
   "adds new pairs to command-message to be consumed by extension events"
   [db {:keys [chat-id] :as command-message}]
-  (let [{:keys [contacts public? group-chat]} (get-in db [:chats chat-id])]
-    (add-chat-contacts contacts (assoc command-message :public? public? :group-chat group-chat))))
+  (let [{:keys [contacts public? group-chat]} (get-in db [:chats chat-id])
+        all-contacts (get db :contacts/contacts)]
+    (add-chat-contacts all-contacts contacts (assoc command-message :public? public? :group-chat group-chat))))
 
 (defn generate-short-preview
   "Returns short preview for command"
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -34,8 +34,7 @@
 
 (defn build-contact [{{:keys [chats] :account/keys [account]
                        :contacts/keys [contacts]} :db} public-key]
-  (cond-> (assoc (contact.db/public-key->contact contacts public-key)
-                 :address (contact.db/public-key->address public-key))
+  (cond-> (contact.db/public-key->contact contacts public-key)
     (= public-key (:public-key account))
     (assoc :name (:name account))))
 
@@ -233,9 +232,7 @@
                               {:public-key   public-key
                                :photo-path   profile-image
                                :name         name
-                               :address      (or address
-                                                 (:address contact)
-                                                 (contact.db/public-key->address public-key))
+                               :address      (or address (:address contact))
                                :last-updated timestamp-ms
                                   ;;NOTE (yenda) in case of concurrent contact request
                                :pending?     (get contact :pending? true)}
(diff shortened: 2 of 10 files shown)


dmitryn (approved):

goranjovic (approved):

asemiankevich:
@antdanchenko ^

antdanchenko:
@asemiankevich `intrinsic gas too low` and `'null is not an object (evaluating 'e.substring')'` looks like bugs introduced by this PR

antdanchenko:
at least those issues are reproducible on apk from this PR, and not reproducible on latest nightly build

asemiankevich:
thanks @antdanchenko , please check fail 1 - we need to adjust our tests this is feature change
@gravityblast i also was able to repro those 2 remaining errors in pr and in nightly it is all fine. 
`intrinsic gas too low` - maybe we dont fetch gas limit / gas price from infura?
`null is not an object (evaluating 'e.substring')` - i think is caused by ^

gravityblast:
I think that `Recipient address is not shown` in `test_send_transaction_details_in_1_1_chat` should fail anyway, because we are changing that feature. shouldn't we change the test? @asemiankevich @antdanchenko

asemiankevich:
@Serhy @antdanchenko i am stuck on this PR to make e2e pass. I can;t repro the issues manually at all in this PR. Apart 1 fail everything should pass but it is not. Maybe you can assist?

asemiankevich:
so:
- `intrinsic gas too low` happens when i send txn to non-contact
- `null is not an object (evaluating 'e.substring')` happens when sending tokens to non-contacts

for some reason i missed that when re-checking. Or added user to contacts automatically so the issues were not reproducible.

How do we want to handle those cases:
- saying user that he can't transact with non-contact?
- how to transact with a dapp? UPD: checked with Test Dapp - it is fine for some reason..
- why we force users to add contacts? Metamask does not do it

@denis-sharypin ?  @hesterbruikman @guylouis

gravityblast:
we also already have this behaviour in the release:

1 - open a public chat where some one wrote something, and he's not in your contacts
2 - click on his profile
3 - click on send transaction
4 - try sending the transaction

at this point the transaction address is empty so it's not possible to send the transaction.

> * why we force users to add contacts? Metamask does not do it

This is specific to status chats contacts, in metamask you only send transactions to addresses right? 
In our case it's the same, you can still send a transaction to an address if you know it. 
But in this case we wait for the other user to automatically send it when pressing "add to contacts".

@asemiankevich @denis-sharypin  @hesterbruikman @guylouis

hesterbruikman:
Echoing @gravityblast here.

Two changes seem related:
1 contacts were added automatically without the user tapping [add to contacts], but when tapping on [send message]. This was changed this week: https://github.com/status-im/status-react/pull/7421
@asemiankevich this could by why it didn't show before.

2 For me to send you a transaction from your profile view in my account will only be possible if I am in your contact list. This deviates from the current situation but I believe this is acceptable as:
- Increases privacy. Non-contacts will need to actively derive your wallet address. We don't provide it in Status unless explicitly done so by the user.
- It creates a way forward to the use of multiple wallet addresses of which in the future one can be connected to the WhisperID allowing transactions between non-contacts with a wallet of choice.
- From a UX perspective it would be familiar to new users as it's similar to having a chat with someone without necesarily sharing your banking details.

gravityblast:
thank you @hesterbruikman !

> 1 contacts were added automatically without the user tapping [add to contacts], but when tapping on [send message]. This was changed this week: #7421
@asemiankevich this could by why it didn't show before.

it shouldn't be related to that PR, but actually to this PR I worked on. 

Past behaviour: 

1 - I open a chat with you 
2 - the app automatically create a contact object in my own contact list with data derived from your public key, including the address.

Behaviour with this PR: 

1 - I open a chat with you 
2 - the app automatically create a contact object in my own contact list with data derived from your public key, without including the address. So only the random name is automatically derived.
In this case, I have an "anomymous" contact in my contact list, with an empty wallet address, that I will receive when you added me, as you said.

In both cases my contact info is not sent to you, thanks to the PR you mentioned

asemiankevich:
https://notes.status.im/m0R19NTFScayoLIBly78fg?view

<blockquote><div><strong><a href="https://notes.status.im/m0R19NTFScayoLIBly78fg?view">CodiMD - Collaborative markdown notes</a></strong></div></blockquote>

asemiankevich:
@antdanchenko is there any news about e2e update for this PR? would nice to have it merged some day soon

gravityblast:
@asemiankevich @antdanchenko I'm going to close this PR so we don't need any change in the automated tests, I'm going to write another comment with the reasons why we don't need it anymore.

gravityblast:
We decided to manage the issue described in this PR in a different way. We are going to support multiple wallets, so a keycard account can still receive funds in the address derived from the whisper key. 
The wallet will be able to show both "whisper address" and wallet address, so no changes are needed.

Issue #7401, opened 2019-02-01
privacy leak - sharing contact info without user action


gravityblast:
# Problem

Initially it was possible to have a 1:1 chat with a user without sharing personal contact information like name and profile picture.
It looks like it's not possible anymore, because contact informations are sent automatically to the other user when opening a 1:1 chat.

# Example

1 - Alice writes a messages in a public chat
2 - Bob sends a message to the same public chat
3 - Alice cannot see Bob's real name; she can see the 3 random words.
4 - Bob clicks on Alice's profile and then on "Send message". 1:1 chat opens (without actually sending messages).
5 - At this point Alice can see Bob's real name, because profile informations have been sent at point 4, without any action from Bob. Bob didn't have to click on "Add to contacts"

As we discussed in Status, we changed this behaviour some months ago, but I'm not sure if it has been decided from a UX point of view, and if we want to keep it.

Is there any other reason for this? Maybe related to push notification?

hesterbruikman:
From a UX perspective this is undesireable behavior as regardless of whether we want to expose identities in advance of sending a message (don't think we do), we currently don't communicate this system state well.

cammellos:
I believe it was done during the move to the new protocol, as a mean to exchange a key (without the contact request the two users would not be able to communicate). That mechanism has been now removed, and this simply looks like an artifact of that choice, so there's no harm in removing. Push notifications will be affected in the sense that if you don't add someone in your contacts, when they send you a message, you wont receive push notifications, but that's expected.

hesterbruikman:
@corpetty what's your view on this?

corpetty:
Yes, please remove.  you should be able to contact a public key (which will be better w.r.t. spam after TtT) but the profile shouldn't be exposed until after the contact has been added.

cammellos:
@hesterbruikman I take the same goes for when I copy and paste the public key in `Start new chat` or use an ENS name? (currently I believe they are also added as contacts)

hesterbruikman:
IYes, thanks for bringing that up. IMO only an explicit 'Add to contacts' should lead to adding people to a contact list. There might be an exception in Tribute to Talk where I believe people are added to eachothers contact list upon release funds from escrow. cc @rachelhamlin

PR #7421, opened 2019-02-05
Remove automatic sending of contact request

cammellos:
Fixes: #7401 
Fixes: #7500 
Removes sending automatically a contact request on "Send message" or "Start new chat".

Also allows to send transactions to non-contacts.
cc @goranjovic , the change is pretty small (we just build the contact if it's not in our list), but maybe you would like to have a quick look at it.

### Testing
- Send message should not add a user a contact
- Start new chat -> QR CODE / Public Key should not add a user as a contact

status: ready

### DIFF
--- a/src/status_im/chat/commands/impl/transactions.cljs
+++ b/src/status_im/chat/commands/impl/transactions.cljs
@@ -14,6 +14,7 @@
             [status-im.ui.components.animation :as animation]
             [status-im.ui.components.svgimage :as svgimage]
             [status-im.i18n :as i18n]
+            [status-im.contact.db :as db.contact]
             [status-im.constants :as constants]
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.ethereum.tokens :as tokens]
@@ -296,7 +297,10 @@
   protocol/Yielding
   (yield-control [_ {{{amount :amount asset :asset} :params} :content} {:keys [db] :as cofx}]
     ;; Prefill wallet and navigate there
-    (let [recipient-contact     (get-in db [:contacts/contacts (:current-chat-id db)])
+    (let [recipient-contact     (or
+                                 (get-in db [:contacts/contacts (:current-chat-id db)])
+                                 (db.contact/public-key->new-contact (:current-chat-id db)))
+
           sender-account        (:account/account db)
           chain                 (keyword (:chain db))
           symbol-param          (keyword asset)
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -260,10 +260,9 @@
 (def receive-contact-request-confirmation handle-contact-update)
 (def receive-contact-update handle-contact-update)
 
-(fx/defn add-contact-and-open-chat
+(fx/defn open-chat
   [cofx public-key]
   (fx/merge cofx
-            (add-contact public-key)
             (chat.models/start-chat public-key {:navigation-reset? true})))
 
 (fx/defn hide-contact
@@ -283,8 +282,8 @@
       (fx/merge cofx
                 fx
                 (if config/partitioned-topic-enabled?
-                  (add-contacts-filter contact-identity :add-contact-and-open-chat)
-                  (add-contact-and-open-chat contact-identity))))))
+                  (add-contacts-filter contact-identity :open-chat)
+                  (open-chat contact-identity))))))
 
 (fx/defn open-contact-toggle-list
   [{:keys [db :as cofx]}]
@@ -297,4 +296,4 @@
 (fx/defn add-new-identity-to-contacts
   [{{:contacts/keys [new-identity]} :db :as cofx}]
   (when (seq new-identity)
-    (add-contact-and-open-chat cofx new-identity)))
+    (open-chat cofx new-identity)))
--- a/src/status_im/contact/db.cljs
+++ b/src/status_im/contact/db.cljs
@@ -63,17 +63,6 @@
 (spec/def :contact/new-tag string?)
 (spec/def :ui/contact (spec/keys :opt [:contact/new-tag]))
 
-(defn public-key->new-contact [public-key]
-  {:name       (gfycat/generate-gfy public-key)
-   :photo-path (identicon/identicon public-key)
-   :public-key public-key})
-
-(defn public-key->contact
-  [contacts public-key]
-  (when public-key
-    (get contacts public-key
-         (public-key->new-contact public-key))))
-
 (defn public-key->address [public-key]
   (let [length (count public-key)
         normalized-key (case length
@@ -83,6 +72,18 @@
                          nil)]
     (when normalized-key
       (subs (.sha3 js-dependencies/Web3.prototype normalized-key #js {:encoding "hex"}) 26))))
+
+(defn public-key->new-contact [public-key]
+  {:name       (gfycat/generate-gfy public-key)
+   :address    (public-key->address public-key)
+   :photo-path (identicon/identicon public-key)
+   :public-key public-key})
+
+(defn public-key->contact
+  [contacts public-key]
+  (when public-key
+    (get contacts public-key
+         (public-key->new-contact public-key))))
 
 (defn- contact-by-address [[_ contact] address]
   (when (ethereum/address= (:address contact) address)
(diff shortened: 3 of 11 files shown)


hesterbruikman (approved):

gravityblast (approved):

rasom (approved):

mandrigin (approved):

antdanchenko:
@cammellos please wait for e2e test updates before merging this PR, I will add them in separate commit to the PR

cammellos:
@antdanchenko for sure, thanks for looking into it!

asemiankevich:
@antdanchenko ^

asemiankevich:
@cammellos what is expected in this PR if:
1. user 1: add user 2 to contacts, send him some messages
2. as user 2: send some messages to user 1, but do not add to contacts 
3. recover user 1 on another device

Should i see [add to contacts] as user 1 (recovered) in 1x1 chat ? @churik has logged bug in develop https://github.com/status-im/status-react/issues/7500 however wondering what is expected here in PR

cammellos:
@asemiankevich you should not see it (unless devices were paired, but does not seem to be the case), so looks like the same issue as develop.

antdanchenko:
@cammellos, @hesterbruikman please see the [video](https://app.saucelabs.com/tests/28c30e19fcaf426c8d38c5f8b7c104ac?auth=954522963184f895e7622e7f90f0fde0) starting from 5:15

I would suggest to block send/request commands in chat for users which are not in your contacts (and some message, e.g. `Add a user to contacts before making a transaction) otherwise the flow looks completely broken

gravityblast:
> I would suggest to block send/request commands in chat for users which are not in your contacts

That was what we wanted to do in another PR but we agreed that we want to be able to send transactions to anyone, to enable other features like kudos etc.

antdanchenko:
@gravityblast but we are not, since a user is not added to contacts, we definitely need to let him know whats wrong, `intrinsic gas too low` is a misleading error

cammellos:
@hesterbruikman @gravityblast @antdanchenko I can possibly change that in the scope of this PR, basically we want to allow people to send transaction to non-contacts, is that correct?

antdanchenko:
@cammellos Let's wait for a decision from design or UX before making any new changes

gravityblast:
> basically we want to allow people to send transaction to non-contacts, is that correct?

sorry I confused the problem, I don't want to ask for changes :)
I thought we were talking about sending a tx to someone who didn't add me. but that's another thing.

asemiankevich:
@hesterbruikman @rachelhamlin @denis-sharypin we are running in the exactly same dialog as in previous PR. Can we make some UI hint in this case for user? I personally dont want to "make" automation ignore this error because it is indicator of some problem (the exact aim of autotests). Please advise

hesterbruikman:
@cammellos thanks for pinging. Apologies for the incoming wall of text. I'm not sure we've carefully phrased the _question_ (a) and _decision_ (b) to allow people to send transactions to non-contacts. 

Overall I would feel much more comfortable with a Send Transaction Request (i.s.o Send Transaction), if we want to allow sending kudo rewards to non-contacts, _until_ we can offer all users control over what transactions can be mapped to their name and what they say (i.e. multi-wallet support).

From the usability perspective this PR was really about giving users _control_ and _transparency_. If I add someone as a contact they become a contact. If I don't, they don't. As long as copy and controls in the UI accurately represent what happens, this really is more a _privacy_ and _security decision._ cc @corpetty @jarradh 

1. Regarding phrasing of the _question_: IMO this is not so much about do we want to allow sending transactions to non-contacts, but more about do we want to allow non-contacts access to a given address. 

2. To me the decision depends on what other transactions and exposure are connected to this address at the moment of release. i.e. Will the contact gaining access to the address see a bunch of meaningless transactions low value chat transactions as it is only one of the addresses in Status used for transactions and the user can control to limit transactions that can be connected to their behaviour in chat.
OR 
Will they see your history of Spankchain transactions and when they were made because at the time of release there is only one address you can use for all transactions and this is the same address anyone (non-contacts) can access. 
_Also, taking into consideration the delta to the current implementation and again how transparent we are about the information you give up._

With the recent proposals regarding the wallet structure I'm not sure what address will be used to send transactions to non-contacts and what other transactions and exposure are connected to this. Hence, why I would prefer a Send Transaction Request alternative at this stage.

cammellos:
My turn for a wall of text :)

I understand that some effort will be made in separating wallet keys, but I think that's a separate discussion and I'd like to keep the discussion more focused on the immediate problem we are solving.

Our objective is to remove the automatic sending of the contact request.

There is a mechanism which we were not aware of that relies on this automatic sending of request:

I can only send transactions to users who either added me in their contacts (which makes sense) or (this is the important point, and does not make much sense) **I have added to my contacts**. 

Mind that adding someone to your contacts (say B) is completely transparent to B, and does not require any interaction on B part. No disclosure of data on B's side by having A adding B to their contacts.

Removing the automatic sending of the contact request exposed this issue, as an error is shown if you try to send a transaction to someone not in your contacts.

Furthermore the issue can be replicated outside this PR (see details below @antdanchenko  ):

So just to achieve our initial goal (don't send the automatic contact request), there are a few options:

1) Allow sending transactions to non-contacts (again, this has no impact on privacy, as currently I can just send a transaction to x just by adding them to my contact, without explicit permission or x's info disclosure)

2) Force A to add B to their contacts before sending a transaction (does not make sense to me, as you are not asking permission, you are just ticking a box really, and we would just be bending to a technical constrain for no good reason)

3) Force B to have added A to their contacts before sending a transaction 

4) Have a Send Transaction Request flow for non-contacts (similar to 3 I guess but more explicit)

1 or 2 are low effort, and allows us to achieve the goal of this task in the short term, effectively making the product better now (probably a day worth of effort say, unless there are some complications).

3 & 4 require more effort and will need design/ux input as currently the contact request flow is opaque (in the case of 3), and I understand we haven't started on 4.

So the question is simply, is it better to go for 1 or 2 now, or wait for 3 or 4? 

In my opinion there's no reason to wait for 3 & 4 as the current behavior is clearly an artifact of a technical decision (sending the contact request automatically), and I doubt it was a deliberate choice, and we can always improve and move to 3 or 4 when ready.

What do you think? 

### Replicate issue on release

1) Go to a profile of a user not in your contacts
2) Click on send-transaction, you will be redirected to the chat (this flow does not automatically add the user to contacts)
3) Try to send a transaction

The recipient field is then empty and you can't send a transaction (you have already all the information to send one i.e their public key).

hesterbruikman:
Thanks for bringing focus and options @cammellos . Agreed, we shouldn't derail this PR to accomodate a new requirement. 

> 1. Allow sending transactions to non-contacts (again, this has no impact on privacy, as currently I can just send a transaction to x just by adding them to my contact, without explicit permission or x's info disclosure)

I believe this has impact on privacy, it's just no different from the current situation and out of scope of this PR. So seems like the best way forward for now, next to creating a new issue for 4.

cammellos:
Thanks for the reply,

When I said it has no impact on privacy, I meant that it does not make any concession compared to the current behavior (of course it would improve things as we would not send a contact request automatically).

I will then close this PR and removing the automatic contact request will be done in the scope of 4 (as we can't get this through otherwise it would make things worse), is that correct?

cammellos:
@hesterbruikman as discussed we'd go for 1 so we can get this through and then we will work on 4 once the keys are decouple.

hesterbruikman:
Adding rationale: 2 would reinforce the perception of Contacts being mutual and that by adding a contact you will always have the other persons wallet address. This is the case at the moment, but might/should (?) change.

cammellos:
@antdanchenko I have fixed the issue, could you run e2e tests again against this? (I have moved it to TEST and back but not sure they have been triggered)

asemiankevich:
> @antdanchenko I have fixed the issue, could you run e2e tests again against this? (I have moved it to TEST and back but not sure they have been triggered)

let me start them when build arrives

cammellos:
@asemiankevich thanks!

asemiankevich:
@cammellos so what is left here? @antdanchenko to fix the rest of tests? or they are fixed?

cammellos:
@asemiankevich I will try to fix them myself today, on monday if I have still have made no progress, I will ask @antdanchenko to assist

cammellos:
@antdanchenko I have updated the e2e tests, change is minimal and new flow is tested. @asemiankevich if you would like to test it again, it's ready.

asemiankevich:
Those scenarios work as expected. 

- [x] Send message should not add a user a contact
- [x] Start new chat -> QR CODE / Public Key should not add a user as a contact

No regressions found so far.

PR #7425, opened 2019-02-05
[#7422] Show feedback on mailserver error

### LINKED ISSUE #7422: No user feedback on mailserver error
# Problem

When a user selects a mailserver and the connection to this mailserver is unsuccessful (state goes from `:connecting` to `:error`), we don't handle that at all.

So, essentially, after that, no error message is shown, there are no re-tries and we don't even try to request historic messages. App is essentially unusable.

# Acceptance Criteria

When the mailserver connection goes to `:error`, show an error message to a user ("error connecting to the mailserver `{mailserver_name}`") and 2 buttons ("retry" and "pick another"). When user presses "pick another", navigate him to the mailserver choosing screen.

rasom:
fix #7422

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -361,8 +361,22 @@
   "mark mailserver status as `:error` if custom mailserver is used
   otherwise try to reconnect to another mailserver"
   [{:keys [db] :as cofx}]
-  (if (preferred-mailserver-id cofx)
-    {:db (update-mailserver-state db :error)}
+  (if-let [preferred-mailserver (preferred-mailserver-id cofx)]
+    (let [current-fleet (fleet/current-fleet db)]
+      {:db
+       (update-mailserver-state db :error)
+       :ui/show-confirmation
+       {:title               (i18n/label :t/mailserver-error-title)
+        :content             (i18n/label :t/mailserver-error-content)
+        :confirm-button-text (i18n/label :t/mailserver-pick-another)
+        :on-accept           #(re-frame/dispatch
+                               [:navigate-to :offline-messaging-settings])
+        :extra-options       [{:text    (i18n/label :t/mailserver-retry)
+                               :onPress #(re-frame/dispatch
+                                          [:mailserver.ui/connect-confirmed
+                                           current-fleet
+                                           preferred-mailserver])
+                               :style   "default"}]}})
     (let [{:keys [address]} (fetch-current cofx)]
       (fx/merge cofx
                 {:mailserver/remove-peer address}
@@ -384,7 +398,7 @@
         (fx/merge cofx
                   {:db (dissoc db :mailserver/connection-checks)}
                   (when (= :connecting (:mailserver/state db))
-                    (change-mailserver cofx)))
+                    (change-mailserver)))
         {:db (update db :mailserver/connection-checks dec)}))))
 
 (fx/defn reset-request-to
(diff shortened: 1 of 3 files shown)


mandrigin (approved):
great!

src/status_im/mailserver/core.cljs
```diff
@@ -384,7 +397,7 @@
         (fx/merge cofx
                   {:db (dissoc db :mailserver/connection-checks)}
                   (when (= :connecting (:mailserver/state db))
-                    (change-mailserver cofx)))
+                    (change-mailserver)))
```
flexsurfer, line 401: interesting i thought it should be #(when (change %))

flexsurfer (approved):

src/status_im/mailserver/core.cljs
rasom: it can be `#(when (change %))`, literally the same code

Serhy:
@rasom At this moment all our mailsevers are working fine on both iOS and Android, so I tested with 'fake' mailserver record which always will fail to connect. 

**1)** If Device if offline - we still do retries to connect to mailsever and 'Error connecting to mailserver' pop-up is shown. No need to check mailsever connectivity if device is 'Offline'. To reproduce:
* Open Status and create new account
* Switch device to Airplane mode and leave Status on foreground

**2)** The new 'Error connecting to mailserver' pop-up is persistent and that has two options only: Retry and 'Pick another mailserver.'.If user due to natural connectivity issues could not connect to any of mail servers - we fire up the pop-up again and again, making the app almost unusable. There is a hook, however, tap 'Pick another mailserver' and return back to different screen, but not obvious. UX may suffer in that case. 
@errorists  Do we need 'Cancel' option here as well? If 'Cancel' is tapped - it will to stop retrying process and giving an ability for user to keep using an app with no connection established to mailsever? To reproduce:
* Open Status and create new account
* Create your own mailsever record that won't work (e.g. 
enode://531e252ec966b7e83f5538c19bf1cde7381cc7949026a6e499b6e998e695751aadf26d4c98d5a4eabfb7cefd31c3c88d600a775f14ed5781520a88ecd25da3c6:status-offline-inbox@35.225.227.79:3333) **OR** enable Airplane mode on device.
* Once 'Error connecting to mailserver' is shown tap 'Retry' or 'Pick another mailsever' -> tap your own 'broken' mailsever entity.

mandrigin:
@rasom one question: when we didn't select the mailserver "by hand", it looks like it re-tries random ones forever...
didn't we suppose to have had a limit there? maybe it was also broken?

Serhy:
I could not connect to `mail-02.gc-us-central1-a.eth.beta` on Android (while this mailsever is OK for iOS) so tested pop-up in a "natural way". As well as with the fake mailserver. 

Scenario with the device appearing Offline still relevant. So if user appears offline, we do retry and show 'Error connecting to mailserver' pop-up. It can be cancelled now (with 'Cancel' button), but anyway, @mandrigin, do you think we don't need retry in that case or it fine? 

1) Open Status -> create new account
2) Switch airplane mode ON
**Actual result:** 'Offline' badge appears on the screen. In ~10 seconds 'Error connecting to mailsever' pop-up appears
screenshot 2019-02-05 at 18 40 30

**Expected result:** 'Offline' badge appears on the screen.

mandrigin:
@Serhy I think it is fine for now

rasom:
> didn't we suppose to have had a limit there? maybe it was also broken?

@mandrigin I don't know, can take a look but that's a bit different thing I suppose 

> When a user selects a mailserver and the connection to this mailserver is unsuccessful ...

Serhy:
Ready for merge.

mandrigin:
@rasom please squash, merge and cherry-pick to `release/0.9.33`

Issue #7368, opened 2019-01-29
Push notification is shown only on one device even if there are several paired devices with same account


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: Push notification is shown only on one device while same account can be recovered on several devices and these devices can be paired, so user would expect to have same behaviour (getting PN, messages, contacts, chats) on all devices with same account.

#### Expected behavior
[comment]: # (Describe what you expected to happen.)
PN is shown on all paired devices where same account exists
#### Actual behavior
[comment]: # (Describe what actually happened.)
PN is shown only on one of such devices
### Reproduction
[comment]: # (Describe how we can replicate the bug step by step.)

- Recover same account on 2 devices and pair them
- Add a contact 
- get new message in 1:1 chat from added contact while app is on background or closed on both devices
- check if push notification is shown on both devices that just received a message

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: 0.9.33
* Operating System: Android and iOS

cammellos:
@churik this should now work as expected (limited to 3 devices, if you have more hanging around, then you might get inconsistent behavior), should we close it?

churik:
Yes, for sure! Don't know why `fixes` keyword didn't make it in #7426

PR #7426, opened 2019-02-05
Allow multiple push notifications

cammellos:
We keep tokens synchronized across devices, so that the user can notify
us on any paired device.
Currently we record the installation id associated to the fcm-token even
though is not necessary, but it will be once we send device-to-device
messages, in which case we want to notify only those devices.

## How does it work?

A device advertise their own token on contact request/updates etc (same as before)
In addition to advertising their own token, they advertise the token of each enabled paired device.

When sending a message, a peer will take the last 3 fcm-token (3 is the maximum number of paired devices), and send a push notifications to those devices. It will be forgiving meaning that it will not use pairing info to exclude devices.

### Normal flow

1) Create device A1 & A2 and B1
2) Pair device A1 & A2
3) Add B1 to contacts
4) Send a message from B to A, both A1 & A2 should get a notification

### Pairing flow

1) Follow normal flow
2) Create device B2 and pair B1
3) Sync all from B1 to B2
4) Send a message from B2 to A, both A1 & A2 should get a notification

### Non-paired flow

1) Create device A1 & A2 and B1
2) Add B1 to contacts from A1
3) Add B1 to contacts from A2
4) Send a message from B to A, both A1 & A2 should get notifications

### Too many devices

1) Create device A1 & A2 & A3 and B1
2) Pair device A1 & A2 & A3
3) Add B1 to contacts
4) Send a message from B to A, both A1 & A2 & A3 should get notifications
5) create device A4
6) Add B to contacts on A4
7) Send a message from B to A, only 3 devices should get notifications (A4 included, one of A1/A2/A3 should be missing)

status: ready

### DIFF
--- a/src/status_im/accounts/update/core.cljs
+++ b/src/status_im/accounts/update/core.cljs
@@ -1,15 +1,23 @@
 (ns status-im.accounts.update.core
-  (:require [status-im.data-store.accounts :as accounts-store]
-            [status-im.transport.message.protocol :as protocol]
-            [status-im.data-store.transport :as transport-store]
-            [status-im.transport.message.contact :as message.contact]
-            [status-im.utils.fx :as fx]))
+  (:require
+   [status-im.contact.device-info :as device-info]
+   [status-im.data-store.accounts :as accounts-store]
+   [status-im.data-store.transport :as transport-store]
+   [status-im.transport.message.protocol :as protocol]
+   [status-im.transport.message.contact :as message.contact]
+   [status-im.utils.fx :as fx]))
 
-(fx/defn account-update-message [{:keys [db]}]
+(fx/defn account-update-message [{:keys [db] :as cofx}]
   (let [account (:account/account db)
         fcm-token (get-in db [:notifications :fcm-token])
         {:keys [name photo-path address]} account]
-    (message.contact/ContactUpdate. name photo-path address fcm-token)))
+    (message.contact/ContactUpdate. name photo-path address fcm-token (device-info/all cofx))))
+
+(fx/defn send-account-update [cofx]
+  (protocol/send
+   (account-update-message cofx)
+   nil
+   cofx))
 
 (fx/defn send-contact-update-fx
   [{:keys [db] :as cofx} chat-id payload]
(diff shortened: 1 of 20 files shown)


src/status_im/accounts/update/core.cljs
```diff
@@ -1,24 +1,30 @@
...
+   [status-im.contact.device-info :as device-info]
+   [status-im.data-store.accounts :as accounts-store]
+   [status-im.transport.message.protocol :as protocol]
+   [status-im.transport.message.contact :as message.contact]
+   [status-im.utils.fx :as fx]))
+
+(fx/defn send-account-update [cofx]
+  (let [fcm-token       (get-in cofx [:db :notifications :fcm-token])
+        {:keys [name photo-path address]} (get-in cofx [:db :account/account])]
+    (protocol/send (message.contact/ContactUpdate. name photo-path address fcm-token (device-info/all cofx)) nil cofx)))
```
rasom: is it possible to split this line into two? it's much longer than 80 chars

rasom (approved):
5m

src/status_im/accounts/update/core.cljs
cammellos: it is indeed possible

mandrigin (approved):

asemiankevich:
@antdanchenko ^

cammellos:
This should not break any tests, possibly something went wrong with rebasing

churik:
@cammellos thanks for the PR, PNs are working much better now!
scenario 'Too many devices' on step 4 says that if you pair 3 devices (A1 & A2 & A3) - all should get notifications.
So steps: 
* Create device A1 (IOS) & A2 (Android1) & A3(Android2) and B1 (created completely new account)
* Pair device A1 & A2 & A3
* Add B1 to contacts (on IOS)
* Send a message from B to A
**Expected result:** A1 & A2 & A3 should get notifications
**Actual result:** A2 & A3 only get notifications.

IOS can't get PNs. Only three installations exist, followed steps one by one with new account.

Test user data:
Log: [geth.log](https://github.com/status-im/status-react/files/2864740/geth.log)
DB: [im.status.ethereum.zip](https://github.com/status-im/status-react/files/2864741/im.status.ethereum.zip)
Password: `qwerty`

churik:
Ok, retested it three times (IOS-Android-Android and IOS-IOS-Android) and couldn't replicate https://github.com/status-im/status-react/pull/7426#issuecomment-463592093

Retested all scenarios above and as a regression checked PNs on desktop and compatibility with 0.9.33.

Thanks for your patience @cammellos and for your help @Serhy and @asemiankevich!
It fixes #7368

PR #7433, opened 2019-02-06
Loading history on mobile network

### LINKED ISSUE #7288: Request chat history on demand while on cellular connection
## Pivotal link
https://www.pivotaltracker.com/story/show/163186553

# Problem

Currently, chats' history is requested from the mailservers regardless of what kind of internet connection is used on the device. For cellular connection, it might cause unacceptable traffic consumption. 

# Implementation

When network connection type changes to cellular a dialog with the next options can be shown to the user:
1. stop chats' syncing, continue on demand/when switch back to a different type of connection
2. continue syncing
3. always stop syncing when such event occurs
4. always continue syncing when such even occurs

On `1.` and `2.` a chosen action is applied and the same dialog is shown on next switching to the cellular network.

If the user chooses `3.` or `4.` a chosen action is applied and on the next occurrence the dialog "stop syncing while on cellular"/"continue syncing while on cellular" with two options can be shown:
- ok
- go to settings

In order to not annoy the user with all these dialogs, it should be shown once per session (or once per some period of time). If we have some connection specific badge it might be shown to the user only once at all. But we still have to make settings easily accessible from almost any screen then (as well as that force-requesting button).

# Acceptance Criteria

- when a device switches to cellular network the user can stop syncing
- when a device switches to wifi network syncing is resumed
- when syncing is stopped the user can demand history by tapping some button
- network connection settings are accessible in profile settings

# Notes

From UI/UX side we need:
1. "force-requesting" button
2. connection state badges

rasom:
fix #7288

notes:
1. A bottom sheet is shown when switch from wifi to cellular network
2. The options, selected in the bottom sheet can be changed in profile settings
3. If "remember my choice" was selected next time bottom sheet will not be shown on switching to cellular
4. If syncing on mobile wasn't enabled history messages from mail servers will not be requested. Connection to mail server can still be established.
5. When devices switches back to wifi and app is running, history is requested
6. History is also requested when user decides to enable syncing while app is on cellular network
7. A separate button for force requesting of history (without enabling syncing on mobile network) will be implemented as a separate PR

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -18,7 +18,8 @@
             [status-im.utils.handlers :as handlers]
             [status-im.accounts.update.core :as accounts.update]
             [status-im.ui.screens.navigation :as navigation]
-            [status-im.transport.partitioned-topic :as transport.topic]))
+            [status-im.transport.partitioned-topic :as transport.topic]
+            [status-im.ui.screens.mobile-network-settings.utils :as mobile-network-utils]))
 
 ;; How do mailserver work ?
 ;;
@@ -332,21 +333,23 @@
 
 (fx/defn process-next-messages-request
   [{:keys [db now] :as cofx}]
-  (when (and (transport.db/all-filters-added? cofx)
-             (not (:mailserver/current-request db)))
+  (when (and
+         (mobile-network-utils/syncing-allowed? cofx)
+         (transport.db/all-filters-added? cofx)
+         (not (:mailserver/current-request db)))
     (when-let [mailserver (get-mailserver-when-ready cofx)]
       (let [request-to (or (:mailserver/request-to db)
                            (quot now 1000))
-            requests (prepare-messages-requests cofx request-to)
-            web3 (:web3 db)]
+            requests   (prepare-messages-requests cofx request-to)
+            web3       (:web3 db)]
         (if-let [request (first requests)]
-          {:db (assoc db
-                      :mailserver/pending-requests (count requests)
-                      :mailserver/current-request request
-                      :mailserver/request-to request-to)
-           :mailserver/request-messages {:web3     web3
-                                         :mailserver    mailserver
-                                         :request request}}
+          {:db                          (assoc db
+                                               :mailserver/pending-requests (count requests)
+                                               :mailserver/current-request request
+                                               :mailserver/request-to request-to)
+           :mailserver/request-messages {:web3       web3
+                                         :mailserver mailserver
+                                         :request    request}}
           {:db (dissoc db
                        :mailserver/pending-requests
                        :mailserver/request-to)})))))
@@ -475,7 +478,6 @@
         ;; If a cursor is returned, add cursor and fire request again
         (if (seq cursor)
           (when-let [mailserver (get-mailserver-when-ready cofx)]
-
             (let [request-with-cursor (assoc request :cursor cursor)]
               {:db (assoc db :mailserver/current-request request-with-cursor)
                :mailserver/request-messages {:web3 (:web3 db)
--- a/src/status_im/ui/components/connectivity/subs.cljs
+++ b/src/status_im/ui/components/connectivity/subs.cljs
@@ -1,7 +1,8 @@
 (ns status-im.ui.components.connectivity.subs
-  (:require [re-frame.core :as re-frame]
-            [status-im.utils.platform :as utils.platform]
-            [status-im.i18n :as i18n]))
+  (:require
+   [re-frame.core :as re-frame]
+   [status-im.utils.platform :as utils.platform]
+   [status-im.ui.screens.mobile-network-settings.utils :as mobile-network-utils]))
 
 (re-frame/reg-sub
  :connectivity/status-properties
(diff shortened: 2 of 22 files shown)


resources/icons/cancel.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
yenda: this is already in develop :main-icons/cancel

resources/icons/network24.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
yenda: this is already in develop `main-icons/network`

resources/icons/cancel.svg
rasom: that's part of another PR, fixed here https://github.com/status-im/status-react/pull/7396/files

src/status_im/ui/screens/mobile_network_settings/events.cljs
```diff
@@ -0,0 +1,101 @@
...
+            (not= :create-account (:view-id db)))
+
+       [(bottom-sheet/show-bottom-sheet
+         {:view :mobile-network})
+        (sheet-defaults)]
+
+       logged-in?
+       [(bottom-sheet/hide-bottom-sheet)]))))
+
+(defn apply-settings
```
cammellos, line 37: Probably it would be good to add some tests for this namespace, also I generally prefer not to use `trooleans` (nil, true, false), at which point I'd rather use strings to capture state

cammellos (approved):

src/status_im/ui/screens/mobile_network_settings/events.cljs
rasom: do you mean for `syncing-on-mobile-network?` and `remember-syncing-choice?`? These fields are not nillable

cammellos: `remember?` and there's another example I think in a different file `new-show?` where we explicitly check that is `false?`, although in that  case looks we have less control as coming from props

src/status_im/mailserver/core.cljs
```diff
@@ -332,24 +332,29 @@
...
-           :mailserver/request-messages {:web3     web3
-                                         :mailserver    mailserver
-                                         :request request}}
-          {:db (dissoc db
-                       :mailserver/pending-requests
-                       :mailserver/request-to)})))))
+  (let [network (:network/type db)
+        {:keys [syncing-on-mobile-network?]} (:account/account db)]
+    (when (and
+           (not (and (= network "cellular")
```
mandrigin: You've defined a function called `cellular?` Maybe we can re-use it  instead of keeping the hardcoded values?

src/status_im/ui/components/connectivity/subs.cljs
```diff
@@ -36,9 +37,13 @@
                        mailserver-request-error?
                        :t/mailserver-request-error-status
 
+                       (and (= network-type "cellular")
```
mandrigin: You've defined a function called `cellular?` Maybe we can re-use it  instead of keeping the hardcoded values?

mandrigin (changes requested):
a few tiny changes

src/status_im/ui/screens/mobile_network_settings/events.cljs
rasom: re `remember?`, is it better now or you want it to be something like `:default`, `:true`, `:false`?

src/status_im/mailserver/core.cljs
rasom: fixed

src/status_im/ui/screens/mobile_network_settings/events.cljs
cammellos: I'd rather have it as the same type (`:not-answered :true :false`), rather than :default true false

mandrigin (approved):

src/status_im/ui/screens/mobile_network_settings/events.cljs
rasom: well, it means that in [some places](https://github.com/status-im/status-react/blob/54d26a723d989341b0d84ce65d7feffbd7f318c3/src/status_im/ui/screens/mobile_network_settings/events.cljs#L75 ) we need to convert boolean to this :true/:false thing, I realize the concern around troolean thing, but I have doubts it is worth doing here

cammellos: oh I see, never mind then, probably not worth

asemiankevich:
@rasom would you mind to correct a small typo? 

image

rasom:
@asemiankevich done

hesterbruikman:
@rasom @asemiankevich just to be on the safe side. There was some copy in the designs that needed an update: "Until connected to wi-fi?" was this copy updated? Should be "Until connected to Wi-Fi"

rasom:
nope it wasn't, I will file a separate PR for this. Thanks @hesterbruikman

Issue #7436, opened 2019-02-06
Endless fetching and hot phone on recovered account with 1:1, group and public chats


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: Recovered account with 9 1:1 chats, 2 groups. Created one new group, added 9 members there and join couple of public chats, installed gfycat extension. At some point I've noticed that fetching indicator constantly moving and phone gets hot. 

I've paired this device with another to check if can repro on another phone (I can't) and switched to other mailserver (1st and 2nd in the list) but got popup that "can't connect to mailserver, select another"...

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: release build 19 for 0.9.33  
* Operating System: Android 9, real google pixel xl

#### Logs

[Status.log](https://github.com/status-im/status-react/files/2836411/Status.log) (7 Mb)

[geth.log](https://github.com/status-im/status-react/files/2836415/geth.log)

annadanchenko:
@cammellos I could reproduce it on fresh account with group chat of 9 members that send some messages inside it. It also joined 1 public chat and one 1:1 chat

cammellos:
@annadanchenko do you have logs for that? it should not be related to group chats/1-to-1 etc, it might happen randomly on any fresh account (judging from the previous logs)

annadanchenko:
> @annadanchenko do you have logs for that? it should not be related to group chats/1-to-1 etc, it might happen randomly on any fresh account (judging from the previous logs)

sorry, I don't (had disabled logs on ios)

Serhy:
I've also got this issue (with similar logs and it's size attached in description).
Could not reproduce on 0.9.32 Release. 
Can reproduce on 0.9.33 Release.
Below scenario looks good for me on iOS Release 0.9.33, but 4 of 5 times on Android 7 I've got mailserver round-robin'ing (that 5th "lucky" time got successful connection on 5-6th mailsever request).
To keep the steps:
1) Install fresh Status app.
2) Recover an account seedphrase `z z z z z z z z z z z z z`. It should receive 1 contact request and 2 group chats from offline inbox.
3) Add the account in contacts and join those group chats
4) Create two more new group chats with the contact you've added
5) Close app
6) Open app, login to account. Wait until you successfully connected to mailserver.

[...]

PR #7439, opened 2019-02-06
Feature/adjust mailserver limit

cammellos:
reduce limit on failures, increase on successes.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -37,9 +37,13 @@
 (def seven-days (* 7 one-day))
 (def maximum-number-of-attempts 2)
 (def request-timeout 30)
+(def min-limit 200)
+(def max-limit 2000)
+(def default-limit max-limit)
 (def connection-timeout
   "Time after which mailserver connection is considered to have failed"
   10000)
+(def limit (atom default-limit))
 
 (defn connected? [{:keys [db]} id]
   (= (:mailserver/current-id db) id))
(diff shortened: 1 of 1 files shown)


mandrigin (approved):
lgtm

annadanchenko:
@cammellos @Serhy I can't repro #7436 on this PR yet

Serhy:
@annadanchenko @cammellos me neither can reproduce issue with this PR. iOS and Android - both looks good for the accounts I had an issue.

flexsurfer (approved):

src/status_im/mailserver/core.cljs
```diff
@@ -37,9 +37,13 @@
 (def seven-days (* 7 one-day))
 (def maximum-number-of-attempts 2)
 (def request-timeout 30)
+(def min-limit 200)
+(def max-limit 2000)
+(def default-limit max-limit)
 (def connection-timeout
   "Time after which mailserver connection is considered to have failed"
   10000)
+(def limit (atom default-limit))
```
flexsurfer, line 46: just wondering, why do you use atom and not app-db?

Serhy:
@mandrigin, @cammellos that's good for merge!

src/status_im/mailserver/core.cljs
cammellos: I didn't think much of it, I instinctively went for an atom as it's very much local state, not used for subscriptions etc, rather than global (no-one should be accessing this from anywhere), but I don't mind putting in re-frame db, it's all good with me

PR #7444, opened 2019-02-07
Stickers polishing 2

flexsurfer:
• the clock handle in the Recents icon is black
• When the Recents view is opened, tapping the 'plus' icon for Sticker Market doesn't open the Store but triggers the 'recents' icon nearby instead
• The horizontal scroll in the sticker selection toolbar should also span the recents icon, so it begins it the middle between `+` and the Recents icon
• upon sending a sticker, we shouldn't dismiss the sticker menu
• there is a gap between the toolbar and scrollable area, pls remove it :)

### DIFF
--- a/resources/icons/stickers/recent.svg
+++ b/resources/icons/stickers/recent.svg
@@ -1,3 +1,4 @@
-<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-    <path fill="black" fill-opacity="0" d="M11 5V11.1716C11 11.702 11.2107 12.2107 11.5858 12.5858L16 17" stroke="black" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
+<svg width="28" height="28" viewBox="0 0 28 28" fill="none" xmlns="http://www.w3.org/2000/svg">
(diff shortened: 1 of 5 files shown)


jeluard (approved):

errorists:
Looking good! One last thing which crossed my mind, a way to dismiss the menu, if you look at the input field you can dismiss it by tapping anywhere inside chat, the same would be great to have for the sticker menu so users wouldn't have to go to input first to be able to close it. 
Also if possible, it would be great if we didn't animate the sticker menu opening when the input is already focused and keyboard visible, the view should only animate when you don't have the keyboard visible and the input is unfocused. This is to prevent the jumping of chat when switching between input and stickers 
That's all 🙏

dmitryn (approved):

flexsurfer:
@errorists done

flexsurfer:
@jeluard i thought we decided to wait approve from designer @errorists ?

errorists:
@flexsurfer you included the things I asked you about, the PR is good ❤️

jeluard:
@flexsurfer @errorists Ah sorry I thought that was done as part of moving to `MERGE` column..
But I knew your PR would be perfect!

PR #7446, opened 2019-02-07
WIP

cammellos:
status: wip


Serhy:
@cammellos With builds here I have no issues with syncing after app restarts with scenarios described in the https://github.com/status-im/status-react/issues/7436
Neither with my own account.
App may spend some to sync but only right after recover, but the same is on 0.9.32 Release.

iOS and Android tested. Upgrade from 0.9.32 Release to this PR build also looks good to me: account used in 0.9.32 with GroupChats, Public, 1-1 chats -> upgraded to this build, I did not experience sync issues. 
Also I switched accounts with Group, Public, 1-1 chats to each mailserver and restarted an app, and no sync issues were at all.

PR #7448, opened 2019-02-07
Dont setup topic as it resets the discovery filter

cammellos:
disabling for now as it causes to re-fetch the discovery topic on login when a group chat is created.
status: ready

### DIFF
--- a/src/status_im/group_chats/core.cljs
+++ b/src/status_im/group_chats/core.cljs
@@ -522,7 +522,9 @@
                                             :members-joined           (:members-joined new-group)
                                             :contacts                 (:contacts new-group)})
                   (add-system-messages chat-id previous-chat new-group)
-                  (set-up-topic chat-id previous-chat)
+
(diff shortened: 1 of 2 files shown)


rasom (approved):

mandrigin (approved):

cammellos:
this looks like a skip-qa, any objection?

mandrigin:
I looks like it has already been/being tested as a part of 0.9.33 release. @annadanchenko @Serhy?
I personally think we can merge that.

Serhy:
@cammellos @mandrigin correct, yes. In the scope of release/0.9.33 it's confirmed that upgrading Android (6.0.1 in my case) from 0.9.33 release on to this PR build has no issues, neither issues with group chats usage between 0.9.33. Restarting an app -> connected to mailserver without delays (< 7 sec). 
Autotests gives us green light too, - ready for merge from me.

PR #7464, opened 2019-02-11
enable sending to partitioned topic

rasom:
Let's merge it after the day after release.

status: ready


cammellos:
@rasom this will make 0.9.34 incompatible with 0.9.33, (as 0.9.33 is not listening to partitioned topic)
is this what we want?

rasom:
@cammellos you are right, I completely forgot that it's not included to 0.9.33. Closing PR for now

PR #7475, opened 2019-02-12
Use dedup-id instead of raw-object

cammellos:
This is a backward/forward compatible change with status-go.
We are changing the way messages are confirmed from passing the
raw-object to status-go to a dedup-id instead, which needs to be sent
back.
Based on the response from statu-go we detect whether they are ids or
object and act accordingly.

status: ready

### DIFF
--- a/src/status_im/transport/message/core.cljs
+++ b/src/status_im/transport/message/core.cljs
@@ -1,6 +1,7 @@
 (ns ^{:doc "Definition of the StatusMessage protocol"}
  status-im.transport.message.core
   (:require [re-frame.core :as re-frame]
+            [goog.object :as o]
             [status-im.chat.models.message :as models.message]
             [status-im.utils.config :as config]
             [status-im.data-store.transport :as transport-store]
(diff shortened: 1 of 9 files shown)


yenda (approved):

src/status_im/transport/message/core.cljs
```diff
@@ -12,10 +13,24 @@
...
+                    (o/get js-message "message")
+                    js-message)
+     :id id}))
+
 (fx/defn receive-message
   [cofx now-in-s filter-chat-id js-message]
   (let [blocked-contacts (get-in cofx [:db :contacts/blocked] #{})
-        {:keys [payload sig timestamp ttl]} (js->clj js-message :keywordize-keys true)
+        {{:keys [payload sig timestamp ttl]} :message
+         dedup-id :id
```
yenda, line 36: Could you add a docstring to explain in the code what the dedup-id is for and where it is coming from, not sure where is the best place to do that maybe here.

yenda (approved):

src/status_im/transport/message/core.cljs
cammellos: sure, good idea

rasom:
typo `statu-go` in commit message

rasom (approved):

PR #7517, opened 2019-02-15
Custom network and seen status autotests fix

Serhy:
* When creating custom network updated network chain locator
* No 'Seen' message status under messages. Updated test to check that 'Sent' status is displayed only
* We hardcoded ETH value for boundary case in the past. Updated test to calculate ETH value depending of the transaction fee amount.


churik (approved):
Thanks Sergey! nice work!

antdanchenko (approved):

PR #7523, opened 2019-02-19
Update bundle script

cammellos:
With the change in status-go mobile the name of the artifact as changed
and currently the script is looking at the wrong file.
I have also added an option to not clean the repo (only if specified),
as I rarely want to clean the repo myself, and I have always to comment
it out.

I have only updated it for linux/android, not sure about IOS.

status: ready


mandrigin (approved):
👍

yenda (approved):

rasom (approved):

PR #7526, opened 2019-02-19
Confirm pairing messages

cammellos:
Partially addresses https://github.com/status-im/status-react/issues/7513
This commit adds confirmation to pairing messages, which previously were
not confirmed.

status: ready

### DIFF
--- a/src/status_im/pairing/core.cljs
+++ b/src/status_im/pairing/core.cljs
@@ -35,6 +35,11 @@
         device-type     utils.platform/os]
     (protocol/send (transport.pairing/PairInstallation. installation-id device-type installation-name fcm-token) nil cofx)))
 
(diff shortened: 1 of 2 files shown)


rasom (approved):

STATUS_GO_VERSION
```diff
@@ -1 +1 @@
-0.21.0g-beta.2
+0.23.0-beta.0
```
mandrigin: can we use a `beta.1` instead? it has a few more fixes...

mandrigin (approved):

STATUS_GO_VERSION
cammellos: sure

mandrigin:
merging since `skip-manual-qa` label

PR #7528, opened 2019-02-19
[FIX] Revert "Allow multiple push notifications"

yenda:
Reproduce in develop:

- add user as contact
- block user -> `Error: contact.device-info must be of type 'contact-device-info'` appears and user is only blocked during current session (db error, blocking is not persisted)

status: ready


cammellos:
I'd rather fix the issue then revert the commit, first it includes a db migration, also it seems a bit uncalled for, my bad the bug was introduced, but reverting in this case seems unnecessary

yenda:
@cammellos ok with me. I thought we were reverting all commits introducing bugs from now on so I don't think it was uncalled for

cammellos:
@yenda no worries, I understand, probably is good to have a discussion, as it might be that I misunderstood and it is what we agreed on, in this case it's probably easier just to fix because of the db migration, I will take a look at it (and I will take a look at pn at the same time).

yenda:
@cammellos are you working on it right now? because otherwise I'll do it now

cammellos:
@yenda working on it, I'll keep you updated.

mandrigin:
@cammellos we usually revert so we don't panic and don't introduce even more regressions while trying to fix the code. It is not a big deal to re-introduce afterwards.

mandrigin (approved):

cammellos:
@mandrigin not sure I agree, in this case it's obviously not the right thing to do to revert, as it will introduce more problems (the db has been migrated already). also it should be a judgement call, sometimes reverting is the right strategy, sometimes it is not. In this case it's clearly not, and should have been checked before

mandrigin:
@cammellos okay, if there is a DB migration, then revert really can introduce more issues than it should, especially since it have been in a nightly.

PR #7529, opened 2019-02-19
Revert "Fetch messages from mailserver upon PN if db is unlocked. Par…

### LINKED ISSUE #7521: Null is not an object (evaluating 'u.length') when message from blocked user is coming
### Description

*Type*: Bug
*Summary*: it was fixed in #7170, but it is again reproducible in current nightly. When message from blocked user is coming, user is seeing `Null is not an object (evaluating 'u.length')`

#### Expected behavior
no error

#### Actual behavior
img_8f7deebbb6e1-1

### Reproduction

- Open Status
- Create account and add User2 to contacts
- Start chat with User2, exchange several messages
- block User2
- as User2: send several messge to 1-1 chat

### Additional Information
* Status version:  nightly 18/02/2019

yenda:
…t of #3451"

This reverts commit 40b3100cfc30289be70b98e4c2797e29ffcc07a3.

fixes #7521

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -404,11 +404,6 @@
    (mailserver/check-connection cofx)))
 
 (handlers/register-handler-fx
- :mailserver/fetch-history
- (fn [cofx [_ chat-id from-timestamp]]
-   (mailserver/fetch-history cofx chat-id from-timestamp)))
-
-(handlers/register-handler-fx
  :mailserver.callback/generate-mailserver-symkey-success
  (fn [cofx [_ mailserver sym-key-id]]
    (mailserver/add-mailserver-sym-key cofx mailserver sym-key-id)))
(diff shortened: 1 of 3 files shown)


cammellos:
same here, not sure reverting should be our default strategy, it has definitely a time and place, but most of the times I think we should fix the issue if possible, if too hard/time consuming reverting should be considered.

yenda:
@cammellos why is breaking an existing feature more acceptable than reverting one? Also what is hard about a revert? It only requires to reopen the original PR and patch it

cammellos:
@yenda that's not what I am saying, breaking an existing feature is not acceptable nor more acceptable than reverting.
This is a different scenario: Given that a feature has been broken (unavoidable at this point), it's better in my view to fix then to revert (if easy enough), as I prefer to move forward rather than backward. There are of course exception (the code has just been merged, the fix is difficult and likely to introduce new bugs etc).

yenda:
@cammellos I don't have a strong opinion about this, I also like to move forward rather than backward, but if we choose to have QA test nightlies only, we should have no hesitation to revert commits with identified bugs in the future. I fixed this one now anyway

Issue #7474, opened 2019-02-12
Messages disappear on restarting the app 


cammellos:
A user reported that messages disappear after restarting the app, both messages that they received and they sent.

They had conversations in #status channel.

After restart, the messages from the previous conversation disappeared (until October 2018, likely the last time it was in the chat before upgrading to 0.9.33).

We asked to replicate, and it was partially replicated (some of the newer messages were persisted, but older were not).

Fetch 24 hours did not help.

We posted a reply to a message that disappeared, and the original message was not visible on the user device (which seems to indicate that the original message was not saved in the db).

Device is running 0.9.33, IOS 12.1.2, Iphone 5s.

Two extensions installed, giphy and an example extension made by status.

We asked to uninstall the extensions and it could still be replicated (all history gone).

Serhy:
To reproduce behavior of disappearing messages after relogin could be done with the device date change to the past:

**Reproduction:**
1) Install Status
2) Manually change the date to yesterday (at this moment it's 19th of February)
3) Open Status and create new account
4) Join `#status` public chat
5) Wait till message history is fetched (Note: it fetches last 24h, so, depends of time when you execute this scenario, yesterdays messages - 18th of Feb at the current moment - should be present too)
6) Close Status app
7) Update the date to current date (20th of February at this moment)
8) Open Status, log in to account
9) Verify that: a) message history fetched in 5th step above is kept. b) message history is fetched and displays for the last 24h
10) Relogin to account
11) Verify that: a) message history fetched in 5th step above is kept. b) message history is fetched and displays for the last 24h
**Actual result:** Latest message displayed in `#status` public chat is the last message posted at 18th of Feb (next messages, starting from 19th of Feb, are not visible. However, on Home screen the 'real' message is shown. 

<img width="970" alt="screenshot 2019-02-20 at 15 06 20" src="https://user-images.githubusercontent.com/8749671/53093971-8dc1ec00-3521-11e9-97b4-292e7c4dd711.png">

Confirmed with iOS, MacOS, Android - develop build (20th of February nightly build)

PR #7535, opened 2019-02-20
Don't override denormalized last-clock-value on messages

cammellos:
Fixes: #7474 

The denormalized-clock-value was erroneously set to the one of the last
message received. This meant that on chats were the clock-value raced
ahead of the timestamp (#status), a message from the mailserver or a
message from someone with an old clock-value would basically make those
messages be sorted in the past.

The correct behavior is that last-clock-value for a given chat should be
the maximum last clock value ever seen for that chat.

@rasom I have also removed some other code that sets the last clock value to the last message in the sorted group, I believe is not necessary, also sorting of the messages in chat groups is by timestamp (for performance optimization, but might result in the wrong order). Does that look ok?

status: ready

### DIFF
--- a/src/status_im/utils/clocks.cljs
+++ b/src/status_im/utils/clocks.cljs
@@ -94,6 +94,5 @@
   (inc (max local-clock (->timestamp-bid))))
 
 (defn receive [message-clock local-clock]
-  (-> (+ 1000 (max (or message-clock 0) (or local-clock 0)))
-      safe-timestamp
-      inc))
+  (-> (max (or message-clock 0) (or local-clock 0))
+      safe-timestamp))
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -111,7 +111,7 @@
 (fx/defn add-message
   [{:keys [db] :as cofx}
    {{:keys [chat-id message-id clock-value timestamp from] :as message} :message
-    :keys [current-chat? batch? last-clock-value dedup-id raw-message]}]
+    :keys [current-chat? batch? dedup-id raw-message]}]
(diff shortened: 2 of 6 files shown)


mandrigin (approved):

src/status_im/utils/clocks.cljs
```diff
@@ -94,6 +94,5 @@
   (inc (max local-clock (->timestamp-bid))))
 
 (defn receive [message-clock local-clock]
-  (-> (+ 1000 (max (or message-clock 0) (or local-clock 0)))
-      safe-timestamp
-      inc))
+  (-> (max (or message-clock 0) (or local-clock 0))
+      safe-timestamp))
```
mandrigin, line 98: I noticed that we don't increment it here anymore, is that expected?

mandrigin (approved):

src/status_im/utils/clocks.cljs
cammellos: Yes, it's not needed really, it's in the standard algorithm https://en.wikipedia.org/wiki/Lamport_timestamps so that's why it was there in the first place, but we are not interested in "receive" events, so it can be skipped (as long as we increment it on send we are good).

mandrigin: then sounds good!

rasom (approved):
looks good, 7m

Serhy:
No regressions found with iOS, Android 8.1, MacOS builds.
Build upgrade, message ordering, offline inboxed history. And can't reproduce https://github.com/status-im/status-react/issues/7474#issuecomment-465567246 anymore (with 20th (yesterday) and 21st of Feb (today) dates)

cammellos:
Thanks @Serhy!

PR #7536, opened 2019-02-20
Miscellaneous unrelated improvements from feature/nix branch

pedropombeiro:

### Summary:

This PR contains changes that currently live in the `feature/nix` branch, but isn't dependent and can be merged to develop immediately, so as to simplify the `feature/nix` branch as much as possible.

status: ready


pedropombeiro:
I was going to mark it as `skip-manual-qa`, but since this touches with communication with status-go, it's not a bad idea to have a second pair of eyes ensure that things work as expected.

mandrigin (approved):

pedropombeiro:
Hmm, looks like std::async doesn't work on Jenkins as it has `gcc 5.5.0-1` installed and I have `gcc 7.3.0` :-(

Moved it back to the `feature/nix` branch where we have guarantees of the gcc version being used.

pedropombeiro:
Ping @cammellos.

The reason I removed the directives in `desktop/reportApp/CMakeLists.txt` is that they are already inherited from the parent project and are then duplicated (confirmed).

cammellos (approved):

jakubgs (approved):

PR #7543, opened 2019-02-21
Retry mailserver request when initial request fails

cammellos:
If the initial request fails we immediately show the error pop-up. This
PR changes the behavior so that it is retried just like any other
request. If 3 requests in a row fail, we show the error pop up if the
user has specifically set a mailserver, otherwise is changed
automatically.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -40,6 +40,7 @@
 (def request-timeout 30)
 (def min-limit 200)
 (def max-limit 2000)
+(def backoff-interval-ms 3000)
 (def default-limit max-limit)
 (def connection-timeout
   "Time after which mailserver connection is considered to have failed"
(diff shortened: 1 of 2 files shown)


src/status_im/mailserver/core.cljs
```diff
@@ -290,7 +291,8 @@
                           (log/info "mailserver: messages request success for topic " topics "from" from "to" to)
                           (do
                             (log/error "mailserver: messages request error for topic " topics ": " error)
-                            (re-frame/dispatch [:mailserver.callback/request-error (i18n/label :t/mailserver-request-error-title)])))))))
+                            (js/setTimeout #(re-frame/dispatch [:mailserver.callback/resend-request {:request-id "failed-request"}])
```
yenda: why not use the set-timeout in utils here?

cammellos: forgot about it :)

mandrigin (approved):

rasom (approved):

Serhy:
Looks good, no regressions found.

PR #7546, opened 2019-02-21
[WIP] slower bottom nav animation

rasom:


rasom:
@churik does it look better? :)

churik:
well, it looks as animation now but I don't know what is the reason to put bottom nav above keyboard and then hide. Probably we need another opinion about this?
I would prefer to remove this animation if we have such a possibility to do this.

rasom:
@churik , you can read starting from this comment https://github.com/status-im/status-react/pull/7483#issuecomment-463920754

churik:
thanks, OK, I'll close an issue then.
No need for PR I believe, no much difference - probably after a while we will get more feedback.
Thanks for a try!

PR #7548, opened 2019-02-21
remove mounting of /dev/fuse for linux builds

jakubgs:
We can't use `--privileged` because:

1. It's horrible practice security wise.
2. It stops use from using [user namespaces](https://docs.docker.com/engine/security/userns-remap/) with docker.

I'm updating `appimagetool` because the current one doesn't support `--appimage-extract-and-run`.

State: READY

### DIFF
--- a/scripts/build-desktop.sh
+++ b/scripts/build-desktop.sh
@@ -79,10 +79,16 @@ function joinExistingPath() {
 STATUSREACTPATH="$(cd "$SCRIPTPATH" && cd '..' && pwd)"
 WORKFOLDER="$(joinExistingPath "$STATUSREACTPATH" 'StatusImPackage')"
 DEPLOYQTFNAME='linuxdeployqt-continuous-x86_64_20181215.AppImage'
-APPIMAGETOOLFNAME='appimagetool-x86_64_20181109.AppImage'
+APPIMAGETOOLFNAME='appimagetool-x86_64_20190221.AppImage'
 DEPLOYQT=$(joinPath . "$DEPLOYQTFNAME")
 APPIMAGETOOL=$(joinPath . "$APPIMAGETOOLFNAME")
 STATUSIM_APPIMAGE_ARCHIVE="StatusImAppImage_20181208.zip"
+
+APPIMAGE_OPTIONS=""
+if [[ ! -c /dev/fuse ]]; then # doesn't exist when run under docker
+    # We extract it to avoid having to use fuse and privileged mode in docker
+    APPIMAGE_OPTIONS="--appimage-extract-and-run"
+fi
 
 function init() {
   if [ -z $STATUSREACTPATH ]; then
@@ -311,7 +317,7 @@ function bundleLinux() {
     rm -f $usrBinPath/Status.AppImage
   popd
 
-  $DEPLOYQT \
+  $DEPLOYQT $APPIMAGE_OPTIONS \
     $desktopFilePath \
     -verbose=$VERBOSE_LEVEL -always-overwrite -no-strip \
     -no-translations -bundle-non-qt-libs \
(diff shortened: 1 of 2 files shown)


mandrigin (approved):

jakubgs:
Aaaaand it still fails:
```
[2019-02-21T16:01:20.022Z] + ./scripts/build-desktop.sh bundle
[2019-02-21T16:01:20.022Z] realpath: /opt/qt/gcc_64/bin: No such file or directory
[2019-02-21T16:01:20.022Z] Creating AppImage...
...(omitted)...
[2019-02-21T16:01:22.690Z] ~/workspace/status-react_prs_linux_PR-7548
[2019-02-21T16:01:22.690Z] fuse: device not found, try 'modprobe fuse' first
[2019-02-21T16:01:23.647Z] open dir error: No such file or directory
```
Makes no sense why we would need `fuse` for this.

jakubgs:
Okay, here's the real culprit:
```
/home/jenkins/workspace/status-react_prs_linux_PR-7548/linuxdeployqt-continuous-x86_64_20181215.AppImage \
  /home/jenkins/workspace/status-react_prs_linux_PR-7548/StatusImPackage/AppDir/usr/share/applications/Status.desktop \
  -verbose=3 -always-overwrite -no-strip -no-translations -bundle-non-qt-libs \
  -qmake=/opt/qt/bin/qmake \
  -executable=/home/jenkins/workspace/status-react_prs_linux_PR-7548/StatusImPackage/AppDir/usr/bin/reportApp \
  -qmldir=/home/jenkins/workspace/status-react_prs_linux_PR-7548/node_modules/react-native \
  -qmldir=/home/jenkins/workspace/status-react_prs_linux_PR-7548/desktop/reportApp \
  -extra-plugins=imageformats/libqsvg.so -appimage
```

jakubgs:
Interesting: https://github.com/probonopd/linuxdeployqt/issues/326#issuecomment-431413468

>Using `--privileged` is a security nightmare.

Yep.

>Your problem is not that you cannot build AppImages in FUSE. Your problem is that you failed to run the linuxdeployqt AppImage in a standard Docker container, due to lack of FUSE there.

>You can simply [extract any AppImage](https://docs.appimage.org/user-guide/run-appimages.html#extract-the-contents-of-an-appimage), and run the AppImage's contents by calling `squashfs-root/AppDir`.
>A shortcut for that has been implemented recently, i.e., in more recent AppImages, you can automate this using `--appimage-extract-and-run`. Please beware of the drawbacks, and don't advertise that option for anything but usage in Docker.

This might help us.

jakubgs:
Okay, so `fuse` isn't necessary to build, it's necessary to use the `linuxdeployqt` AppImage bundle we use. But there is a way around:
```
 $ ./linuxdeployqt-continuous-x86_64_20181215.AppImage --appimage-help
AppImage options:

  --appimage-extract [<pattern>]  Extract content from embedded filesystem image
                                  If pattern is passed, only extract matching files
```

jakubgs:
Well I'll be damned. It works!

vkjr (approved):

scripts/build-desktop.sh
```diff
@@ -79,11 +79,17 @@ function joinExistingPath() {
...
-APPIMAGETOOLFNAME='appimagetool-x86_64_20181109.AppImage'
+APPIMAGETOOLFNAME='appimagetool-x86_64_20190221.AppImage'
 DEPLOYQT=$(joinPath . "$DEPLOYQTFNAME")
 APPIMAGETOOL=$(joinPath . "$APPIMAGETOOLFNAME")
 STATUSIM_APPIMAGE_ARCHIVE="StatusImAppImage_20181208.zip"
 
+APPIMAGE_OPTIONS=""
+if [[ ! -c /dev/fuse ]]; then # doesn't exist when run under docker
+    # We extract it to avoid having to use fuse and privileged mode in docker
+    APPIMAGE_OPTIONS="--appimage-extract-and-run"
```
pedropombeiro, line 90: Isn't `--appimage-extract` enough here? I'm not familiar with `--appimage-extract-and-run`

jakubgs: It's exactly the same as `--appimage-extract` except it uses a temporary directory and cleans up after itself.

pedropombeiro: oh, TYL

pedropombeiro (approved):
Great job!

churik:
@jakubgs 
I installed Linux build, seems fine to me.
Anything else should be checked here?

jakubgs:
@churik don't think so, thanks!

PR #7560, opened 2019-02-22
Created a new event which allows extensions to send chat messages.

### LINKED ISSUE #7354: Allow to send a chat message from an extension
### User Story

As a developer, I want to be able to send a chat message (private chat, public channel) from an [extension](https://status.im/extensions/) event.

### Description

*Type*: Feature

krisc:
Fixes #7354 

You can use this event from an extension like this:

```
[chat/send-message {:to "0xSomeProfileAddress or NameOfAPublicChannelWithout#" :value "Chat Message"}]
```

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -35,6 +35,11 @@
  (fn [_ [_ _ m]]
    {::identity-event m}))
 
+(re-frame/reg-event-fx
+ :chat/send-message
+ (fn [cofx [_ message-text {:keys [to value] :as arguments}]]
+   (chat.input/send-plain-text-message-fx (assoc cofx :now (datetime/timestamp)) value to)))
+
 (re-frame/reg-fx
  ::alert
  (fn [value] (js/alert value)))
@@ -474,6 +479,10 @@
                 {:permissions [:read]
                  :value       :extensions.chat.command/send-message
                  :arguments   {:params :map}}
+                'chat/send-message
+                {:permissions [:read]
+                 :value       :chat/send-message
+                 :arguments   {:to :string :value :string}}
                 'log
                 {:permissions [:read]
                  :value       :log
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -37,7 +37,7 @@
 
 (re-frame/reg-event-fx
  :chat/send-message
- (fn [cofx [_ message-text {:keys [to value] :as arguments}]]
+ (fn [cofx [_ _ {:keys [to value]}]]
    (chat.input/send-plain-text-message-fx (assoc cofx :now (datetime/timestamp)) value to)))
 
 (re-frame/reg-fx
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -479,10 +479,12 @@
                 {:permissions [:read]
                  :value       :extensions.chat.command/send-message
                  :arguments   {:params :map}}
-                'chat/send-message
-                {:permissions [:read]
-                 :value       :chat/send-message
-                 :arguments   {:to :string :value :string}}
+                
+                ;; TODO: figure out better security model
+;;                 'chat/send-message
+;;                 {:permissions [:read]
+;;                  :value       :chat/send-message
+;;                  :arguments   {:to :string :value :string}}
                 'log
                 {:permissions [:read]
                  :value       :log


src/status_im/extensions/core.cljs
```diff
@@ -35,6 +35,11 @@
  (fn [_ [_ _ m]]
    {::identity-event m}))
 
+(re-frame/reg-event-fx
+ :chat/send-message
+ (fn [cofx [_ message-text {:keys [to value] :as arguments}]]
```
jeluard: message-text not used? same for arguments

src/status_im/extensions/core.cljs
```diff
@@ -474,6 +479,10 @@
                 {:permissions [:read]
                  :value       :extensions.chat.command/send-message
                  :arguments   {:params :map}}
+                'chat/send-message
+                {:permissions [:read]
+                 :value       :chat/send-message
+                 :arguments   {:to :string :value :string}}
```
jeluard: What's to? Can it be used for all chat types?

jeluard (changes requested):
Can you provide an example extension so that we can test?

flexsurfer:
just wondering if it was intended by issue, in this PR message will be sent on behalf of a user, it doesn't look secure and idk how it can be used by extension, but anyway currently it's not possible to send any message not from the user account, if we really want to send a message on behalf of a user we must show a warning with message "Extension wants to publish a message in #chat-name "Give me your money!"" "Publish" "Cancel"

src/status_im/extensions/core.cljs
krisc: oh yeah I was using those for debug logs. I can remove them.

src/status_im/extensions/core.cljs
krisc: ':to' can be a user's address or a public channel

krisc:
> Can you provide an example extension so that we can test?

@jeluard 
https://status.im/extensions/play.html?hash=QmPBEETBEXRFZLut5ykhwn1cWTR9PEoDxvPmUVjozLLLj6

The `:to` field is my address, but you can replace it with another account you have access to or with a public channel (without the '#')

<blockquote><div><strong><a href="https://status.im/extensions/play.html?hash=QmPBEETBEXRFZLut5ykhwn1cWTR9PEoDxvPmUVjozLLLj6">Status, a Mobile Client for Ethereum 2.0</a></strong></div></blockquote>

krisc:
> just wondering if it was intended by issue, in this PR message will be sent on behalf of a user, it doesn't look secure and idk how it can be used by extension, but anyway currently it's not possible to send any message not from the user account, if we really want to send a message on behalf of a user we must show a warning with message "Extension wants to publish a message in #chat-name "Give me your money!"" "Publish" "Cancel"

@flexsurfer  Indeed, this sends a message on behalf of the user without warning. @jeluard 
 in https://github.com/status-im/status-react/issues/7354 you did not specify this, but will a warning be necessary before the bounty is considered fulfilled?

krisc:
@jeluard care to comment on this?

jeluard:
@flexsurfer It sends a message as the currently logged user right? if so, why do you think this deserves extra security consideration on top of the regular extensions security features (permissions) ?

flexsurfer:
it's just IMHO, we can discuss this later for sure, but as a user, I would like to control every message from an extension on behalf of my name, and also maybe we could add some small text in every message, something like "sent from an extension"

jeluard:
@flexsurfer Ok we should discuss that point together then. The underlying rationale was that every security/privacy aspect is governed by `permission` concepts.
How is that worst that sending a transaction for instance? Or accessing private info?

flexsurfer:
@jeluard for transactions we confirm every transaction initiated by an extension with the password, so for me, it's like the same if an extension could send transactions without confirmation, imagine you use extension somehow, and after some time you find messages sent by you in every public chat but you never sent these messages, so my concern is only about that user must be aware when extension wants to publish the message on behalf of user , but again it's only my opinion, transaction is a good example btw, user gives permissions to initiate a transaction but he makes a decision send it or not

krisc:
@jeluard @flexsurfer Have we come to a resolution? I'm just wondering if there is more for me to do or not.

jeluard (approved):

jeluard:
@flexsurfer WDYT about merging as is and figure out a better security model while we tackle permissions?

flexsurfer:
@jeluard looks dangerous, can we disable it for now and merge? because if we merge it as is someone could just create simple extension send it in public chat and anyone who installs it may become a spammer

jeluard:
@flexsurfer Makes sense.

@krisc Can you comment the event definition from `capacities` so that we can merge but decide later what to do?

jeluard:
@krisc Also please squash your commits :D

jeluard:
@krisc Also the build is failing

krisc:
@jeluard I think recent complications with updating to the latest dev branch is causing this. This will be easiest for me to resolve by making a new PR. Sorry about that. Submitting one soon...

flexsurfer (approved):

krisc:
@StatusSceptre @jeluard please see https://github.com/status-im/status-react/pull/7694

PR #7562, opened 2019-02-23
[WIP] remove some unnecessary backgrounds

rasom:
attempt to fix redrawing https://discuss.status.im/t/before-i-jump-on-the-react-native-is-slow-bandwagon/1064

before/after

intro

welcome

wallet

chat_list

my_profile

chat

### DIFF
--- a/mobile_files/package.json.orig
+++ b/mobile_files/package.json.orig
@@ -61,6 +61,7 @@
     "react-native-udp": "git+https://github.com/status-im/react-native-udp.git#2.3.1-1",
     "react-native-webview-bridge": "git+https://github.com/status-im/react-native-webview-bridge.git#0.33.16-4",
     "react-navigation": "^2.12.1",
+    "react-navigation-stack": "git+https://github.com/status-im/react-navigation-stack#6d1133b25c4d6ba059cffe7ccf69c92da628a448",
     "realm": "2.21.0",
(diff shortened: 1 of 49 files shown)


mobile_files/package.json.orig
```diff
@@ -61,6 +61,7 @@
     "react-native-udp": "git+https://github.com/status-im/react-native-udp.git#2.3.1-1",
     "react-native-webview-bridge": "git+https://github.com/status-im/react-native-webview-bridge.git#0.33.16-4",
     "react-navigation": "^2.12.1",
+    "react-navigation-stack": "git+https://github.com/status-im/react-navigation-stack#6d1133b25c4d6ba059cffe7ccf69c92da628a448",
```
yenda, line 64: what is this for?

rasom: https://github.com/status-im/react-navigation-stack/commit/6d1133b25c4d6ba059cffe7ccf69c92da628a448

yenda: ok I see it is a transitive dependency of react-navigation
probably worth pushing upstream if we stick with it

yenda: nice catch!

rasom: we need to upgrade react-navigation in out project first

rasom:
This PR will be split into several small PRs in order to minimize possible regressions.

PR #7586, opened 2019-02-26
Dont change mailserver if no network

cammellos:
When we are offline, we don't try to change mailserver, and we don't
show a pop up to the user, as it is not that the mailserver is not
working, we are just offline.

### Testing

1) Login, wait for all the mailserver stuff to be finished
2) Select one mailserver in settings
3) Turn on airplane mode, wait until Offline is displayed
4) Open a chat
5) Fetch 24 hours
6) Wait 20s
7) Turn off airplane mode

On develop it will show a popup on 6
On this PR it should not

Everything else should work the same after 7

I would add to release if works.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -383,27 +383,28 @@
   "mark mailserver status as `:error` if custom mailserver is used
   otherwise try to reconnect to another mailserver"
   [{:keys [db] :as cofx}]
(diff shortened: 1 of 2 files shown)


yenda (approved):

mandrigin (approved):

mandrigin:
added a release label, because the old behaviour is very annoying

churik:
checked on nightly and on PR build on Mac OSx 10 - behavior is same as described by @cammellos.
LGTM

PR #7597, opened 2019-02-27
Fix separate topic

cammellos:
It was using still the separate topic, worked locally as clients were still listening to it.
status: ready


mandrigin (approved):

mandrigin:
reverted on the release branch as well

yenda (approved):

mandrigin (approved):

PR #7601, opened 2019-02-27
Fix concurrent retry-request

cammellos:
Sometimes it happens that the expired signal is received while the
there's a new request in flight.
This happens in cases such as:

1) We send a request (A)
2) We get disconnected from the mailserver
3) We connect to a new mailserver
4) We send a request (B)
5) We receive an expired signal for A

In such cases the request should not be retried or counted as a failure.

### Testing

Testing is difficult, can only be seen through the logs, but in general regression mailserver testing should be enough.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -262,6 +262,10 @@
     (log/debug "Adjusting mailserver request" "from:" from "adjusted-from:" adjusted-from)
     adjusted-from))
 
+(fx/defn handle-request-success [{:keys [db]} request-id]
+  (when (:mailserver/current-request db)
+    {:db (assoc-in db [:mailserver/current-request :request-id] request-id)}))
+
 (defn request-messages! [web3 {:keys [sym-key-id address]} {:keys [topics cursor to from] :as request}]
   ;; Add some room to from, unless we break day boundaries so that messages that have
   ;; been received after the last request are also fetched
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

rasom (approved):

src/status_im/mailserver/core.cljs
```diff
@@ -288,10 +292,12 @@
...
                       (fn [error request-id]
                         (if-not error
-                          (log/info "mailserver: messages request success for topic " topics "from" from "to" to)
+                          (do
+                            (log/info "mailserver: messages request success for topic " topics "from" from "to" to)
+                            (re-frame/dispatch [:mailserver.callback/request-success {:request-id request-id}]))
                           (do
                             (log/error "mailserver: messages request error for topic " topics ": " error)
-                            (utils/set-timeout #(re-frame/dispatch [:mailserver.callback/resend-request {:request-id "failed-request"}])
+                            (utils/set-timeout #(re-frame/dispatch [:mailserver.callback/resend-request {:request-id nil}])
```
yenda, line 300: just pass nothing

Serhy:
iOS and Android builds were tested. New regressions found.
Mailservers round-robin if not pre-set still on place (and once connected successfully - offline messages are received), manually set mailserver also works (and messages are received - checked each mailserver in the list), custom mailserver can be added and messages are fetched.
I didn't try to reproduce scanario in description (no sense). 
Thanks, @cammellos, let's merge!

PR #7607, opened 2019-02-27
revert Replace react-navigation-stack with fork without default background color

rasom:
It seems that our fork introduced some regression and nav transitions on iOS become slow.


rasom:
@errorists confirmed that this fixes slow nav transitions on iOS

rasom:
@mandrigin so is that our hero? It seems that this change was merged into release branch during building

errorists:
@rasom thanks a bunch! 🙏

yenda (approved):
So just removing background colors slows down iOS for some reason?

rasom:
@yenda I'm not 100% sure yet what's the reason here

errorists (approved):

mandrigin (approved):

mandrigin:
@rasom yes, that was the case

PR #7608, opened 2019-02-27
fix arrow alignment

yenda:
Tested on Android
Design only, no tests required (check design on iOS)
Removes the animation for the send button, now the arrow always points up

screenshot_1551294985

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/styles/input/send_button.cljs
+++ b/src/status_im/ui/screens/chat/styles/input/send_button.cljs
@@ -1,17 +1,12 @@
 (ns status-im.ui.screens.chat.styles.input.send-button
   (:require [status-im.ui.components.colors :as colors]))
 
-(defn send-message-container [rotation]
+(def send-message-container
   {:background-color colors/blue
    :width            30
    :height           30
+   :padding          3
    :border-radius    15
    :margin           10
-   :padding          4
    :margin-left      8
-   :margin-bottom    11
-   :transform        [{:rotate rotation}]})
-
-(def send-message-icon
-  {:height 22
-   :width  22})
+   :margin-bottom    11})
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/chat/styles/input/send_button.cljs
```diff
@@ -3,14 +3,15 @@
...
+   :padding-vertical   3
+   :padding-horizontal 4
+   :border-radius      15
+   :margin             10
+   :margin-left        8
+   :margin-bottom      11
+   :transform          [{:rotate rotation}]})
 
 (def send-message-icon
   {:height 22
```
errorists: hey @yenda shouldn't all our icons be 24px in size, does 22 here mean we scale down this one?

flexsurfer (approved):

src/status_im/ui/screens/chat/styles/input/send_button.cljs
yenda: this style is not used anymore I forgot to remove it

errorists: I've noticed it because the 30 width, height are correct, padding should be a 3 on all sides, here it's 4 horizontal.

errorists:
@yenda dumb question, why do we rotate the arrow? shouldn't it be always pointing up? does it get excited after the second character? xD

yenda:
I have no idea I think you can't send a single char?

yenda:
Well you can so should I remove the whole animation?

errorists:
You can, or at least should. For users an emoji is a single character even tho unicode, it's a couple chars. And if you backspace it to a single character, the arrow remains pointing up. Hell yeah, let's remove it.

errorists:
It should always be a static arrow pointing up,[ correct icon in Figma if needed](https://www.figma.com/file/cb4p8AxLtTF3q1L6JYDnKN15/Index?node-id=1105%3A52)

yenda:
@errorists done, I removed the animation, always points up now

errorists:
Thanks Eric, good way to start your day, remove something useless :)

vitvly (approved):

PR #7615, opened 2019-02-28
[temporary workaround] ci: don't check certificates in AppStoreConnect when building iOS

mandrigin:
⚠️⚠️ ⚠️ it will break adding new devices ⚠️⚠️⚠️
but will help workaround the isssue that we don't have 2FA installed on our Apple ID
(essentially, our iOS apps will build again)

to be reverted when we setup a proper 2FA on our main AppStoreConnect account

status: ready


pedropombeiro (approved):

jakubgs (approved):
Good enough for now.

PR #7618, opened 2019-02-28
[reduce overdraw] Remove stack card background on Android

rasom:
It was a bit simpler than that stupid attempt to make a fork :D

But at least it helped to reveal an issue with release build.


yenda (approved):
Nice catch!

flexsurfer (approved):

PR #7623, opened 2019-02-28
Bottom tab animation on leaving/entering main tabs

rasom:
status: ready

### DIFF
--- a/src/status_im/android/core.cljs
+++ b/src/status_im/android/core.cljs
@@ -38,7 +38,7 @@
                       "keyboardDidHide"
                       (fn [_]
                         (dispatch [:show-tab-bar])
-                        (when (zero? @keyboard-height)
+                        (when-not (zero? @keyboard-height)
                           (dispatch [:set :keyboard-height 0]))))
         (.hide react/splash-screen)
         (.addEventListener react/app-state "change" app-state-change-handler)
--- a/src/status_im/chat/subs.cljs
+++ b/src/status_im/chat/subs.cljs
@@ -7,7 +7,8 @@
             [status-im.chat.db :as chat.db]
             [status-im.models.transactions :as transactions]
             [status-im.utils.platform :as platform]
-            [status-im.ui.components.bottom-bar.styles :as tabs-styles]))
+            [status-im.ui.components.bottom-bar.styles :as tabs-styles]
+            [status-im.ui.components.bottom-bar.styles :as tabs.styles]))
 
 (re-frame/reg-sub ::chats :chats)
 (re-frame/reg-sub ::access-scope->command-id :access-scope->command-id)
@@ -77,10 +78,13 @@
  (fn [kb-height]
    (cond
      (and platform/iphone-x? (> kb-height 0))
-     (- kb-height 34 tabs-styles/tabs-height)
-     platform/ios? (- kb-height (if (> kb-height 0)
-                                  tabs-styles/tabs-height
-                                  0))
+     (- kb-height (* 2 tabs.styles/minimized-tabs-height))
+
+     platform/ios?
+     (+ kb-height (- (if (> kb-height 0)
+                       tabs.styles/minimized-tabs-height
+                       0)))
+
      :default 0)))
 
 (re-frame/reg-sub
--- a/src/status_im/ui/components/animation.cljs
+++ b/src/status_im/ui/components/animation.cljs
@@ -57,3 +57,5 @@
   (js->clj (.getLayout value-xy)))
 
 (defn easing [] js/ReactNative.Easing)
+
+(defn cubic [] (.-cubic (easing)))
--- a/src/status_im/ui/components/bottom_bar/core.cljs
+++ b/src/status_im/ui/components/bottom_bar/core.cljs
@@ -10,16 +10,22 @@
    [status-im.i18n :as i18n]
    [re-frame.core :as re-frame]))
 
+(defonce visible? (animation/create-value 1))
+(defonce last-to-value (atom 1))
+
 (defn animate
   ([visible duration to]
    (animate visible duration to nil))
   ([visible duration to callback]
-   (animation/start
-    (animation/timing visible
-                      {:toValue         to
-                       :duration        duration
-                       :useNativeDriver true})
-    callback)))
+   (when (not= to @last-to-value)
+     (reset! last-to-value to)
+     (animation/start
+      (animation/timing visible
+                        {:toValue         to
+                         :duration        duration
+                         :easing          (animation/cubic)
+                         :useNativeDriver true})
+      callback))))
 
 (def tabs-list-data
   [{:nav-stack           :chat-stack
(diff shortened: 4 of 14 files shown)


errorists:
thanks for working on it, just chiming in to make sure, the minimised state should be included in all subviews and not only chats. We show the full tab bar on each of the root views, Home, Wallet, Profile, but once you navigate inside any of their subviews we should minimise it. sorry if I didn't make it clear enough :)

rasom:
@errorists thanks

yep it wasn't clear, though it probably makes it even a bit simpler :)

errorists:
@rasom coool, so yes it's ok but I'd like us to improve if possible. How many callbacks do we get from the transition, because I have a feeling that the animation is too delayed, both on entry and especially exit. It should happen ms away from when we request the transition. 
The swipe gesture also has a tolerance, if you drag and let go the screen less than halfway, it doesn't fire the transition, so we don't animate the tab bar. But if you drag it to at least middle of screen, the back event is fired and then we could immediately start the tab bar animation? Is that out of reach?

flexsurfer (approved):

errorists:
Also, correct me if I'm wrong but it looks like on exit transition we use different timings for the tab bar background and icons/labels, the tab background animates much faster than the icons. It should use the same value for both to look good. I couldn't find the easing value in code, could we try [circular](https://facebook.github.io/react-native/docs/easing#circle) function to make the tab bar motion more lifelike? 🙏

rasom:
@errorists added one change, so that we react on starting transition. It's probably a bit faster than before, but I doubt it is really noticeable. I also doubt we can do it faster on gestures, we react on the same event but still animation happens with delay (when navigate from chat to chats using gesture). In this commit only chats->chat->chats (except input focusing, it is broken) will work properly, just to test if this way to initiate animation is better.

rasom:
@errorists can you please take a look at iOS version https://i.diawi.com/RW2cLm ?

errorists:
@rasom thanks, tried it and there is a difference actually quite a huge one 🚀 

In the previous build, this is where the animation starts on a video recording, you can see it's more than halfway during the transition
screenshot 2019-03-07 at 08 49 40
In the new one it starts exactly on the same frame as transition, samesies for exit transition
that's cool, awesome work you're doing here.

image

here's one bug I managed to find, on transitioning to a minimised tab state, it looks like we leave a blank area under where the tab bar was. I take it because it's where scroll view ends, if so it's ok to leave it like that, I don't want us animating the view height or stuff

two things I'd like us to improve further if possible:
swipe back gesture, is there anything we can do to trigger it when the navigate transition starts, it looks like we animate the tab bar only after the transition ends.
also it would be cool if we'd have a playground to test the different animation curves and timings - hello fiddle :) cc @flexsurfer - , I think there's still room for improvement there, I'd love to see a build with the mentioned circular easing as it should make it accelerate and decelerate faster, nicer.

rasom:
@errorists  tha bug should be fixed https://i.diawi.com/SXKA38

rasom:
> Also, correct me if I'm wrong but it looks like on exit transition we use different timings for the tab bar background and icons/labels, the tab background animates much faster than the icons.

@errorists I'm not sure that I understand what you mean here. In this PR the wall bar is animated, there is no separate animation for icons

errorists:
> @errorists I'm not sure that I understand what you mean here. In this PR the wall bar is animated, there is no separate animation for icons

That's fixed in the current build and no longer an issue, what happened before was the tab bar container animated faster than its content 👍

rasom:
@errorists a new build with circular easing https://i.diawi.com/voSq2s

errorists:
@rasom thanks for giving it a shot, shame it's worse, too stiff, hard to prescribe the right one with only https://easings.net to help you ;) 
Please change it to Cubic and I'll shut up about easing 🙌  
If we can do something about the swipe back gesture, that would make it perfect, if not I'm giving it a green ✅ review, we can go to testing.

rasom:
@errorists https://i.diawi.com/Y5pT44

errorists:
@rasom Cubic is the winner. Thanks! How is it on Android, everything ok?

rasom:
@errorists looks fine to me https://youtu.be/Zke5TgpmP_g

rasom:
@flexsurfer @yenda, pr should be ready finally, please review

cammellos (approved):

flexsurfer (approved):

asemiankevich:
@errorists could you please explain what exactly is expected here?  Thanks!

errorists:
@asemiankevich Sure, the expected behaviour is as follows: whenever you transition from any of the root views so : Home, Wallet, Profile to any of their subviews (chats, send tx flow, settings etc.) the tab bar should shrink and remain this way until you transition back to the root view or change tabs to another root view. 
You may ask why bother? We do this do open up some extra space to show as much content as possible and reduce the visual noise of having too much navigation placed on the bottom, so the footprint of tab bar is much smaller when you're navigating deeper inside one of the key areas. We leave the labels when you're at ground level to better communicate what are the features of the app and where you can go to, I'm afraid leaving icons totally on their own would not be clear enough for some, at least not yet.

👉 What to test? I'd say if you can find a reproducible way for the tab bar to not animate when it's required to :) try navigating around the app, changing tabs, you'll notice a pattern when it shrinks and when expands, I could only fool it once into not animating but it's not reproducible on my end.

asemiankevich:
@rasom can't access Done button on iOS (iphone XR) , chat tab is autofocused. Then i can swich to wallet tab and procceed but looks buggy  because this onboarding screen present twice : on chat tab and wallet tab

image

rasom:
@asemiankevich can you please specify steps to reproduce?

asemiankevich:
> @asemiankevich can you please specify steps to reproduce?

- open status
- create new account
- switch to ropsten
- open Test Dapp
- open Transaction tab
- tap deploy contract
- see Onboarding screen present however you are on chat tab
- switch to Wallet tab
- complete onboarding on wallet tab
- see that one more onboarding is shown

asemiankevich:
@rasom also , when i am under mobile network and try to create account - i can't switch between tabs. + Mobile Settings pop up is not shown. it's like something invisible appears on foreground when 3G enabled which locks the screen for any action. 

i was able to reproduce as well as @Serhy 

Steps:
- turn off Wifi
- turn on mobile newtork
- open status and proceed with account creation
- switch between tabs

rasom:
@asemiankevich is the last issue specific to this PR or can be reproduced in develop?

asemiankevich:
> @asemiankevich is the last issue specific to this PR or can be reproduced in develop?

develop looks good however i dont see the Syncing pop up upon launch here too.. But it is not frozen, i can tap everywhere.

rasom:
did you test it on android? I can't reproduce the issue on iOS

asemiankevich:
@rasom right, can't reproduce the issues on android

rasom:
@asemiankevich i have an impression that our messages contain opposite statements. Can you please clarify: is it iOS or Android bug?

asemiankevich:
> @asemiankevich i have an impression that our messages contain opposite statements. Can you please clarify: is it iOS or Android bug?

as i stated above, all issues i described, i saw on iOS (iPhone XR) device and can't reproduce on Android device.

rasom:
testing last ipa https://webapp.diawi.com/install/Xdt4fM , can't reproduce the issue on iPhone X

asemiankevich:
@rasom would you mind to check video i captured? https://drive.google.com/open?id=1xmf9_uTmUcdgmbwukbYdGjCMhAbGd2w8

rasom:
requested access

asemiankevich:
> requested access

yeah granted. Not sure why it was not done automatically, sorry Oo

rasom:
I was asking about the last issue, 
> when i am under mobile network and try to create account - i can't switch between tabs. 

you def can switch between tabs on video, am I missing something?

asemiankevich:
> I was asking about the last issue,
> 
> > when i am under mobile network and try to create account - i can't switch between tabs.
> 
> you def can switch between tabs on video, am I missing something?

yeah that one i could not reproduce anymore, I probably misread your comment, multitasking. Sorry for confusion (

@Serhy would you mind to check https://webapp.diawi.com/install/Xdt4fM too? You were able to repro the issue with frozen tabs too :D

rasom:
huh, ok, I'm switching to the first problem then

Serhy:
Ok, so with https://webapp.diawi.com/install/Xdt4fM (and with Develop build) and strong 3G mobile connection I don't have an issue with freezes. 
What happens though, is when I have super limited connection (set to 2G (or limited access to network with 3G established as when I checked in https://github.com/status-im/status-react/pull/7623#issuecomment-470910466 ) - in that case I have freezes in app. 
It's on iPhoneX but all fine on Android. And it's in Develop too.

We'll create a separate issue for this, @rasom 

That what may happen to you @asemiankevich, when you've got frozen iOS on account creation: mobile network connected but become really limited at some point... and app freeze. And next time it was all fine.

rasom:
@Serhy is it specific to this PR or can be reproduced in develop?

rasom:
@asemiankevich on-boarding and modals related issue should be fixed now

PR #7627, opened 2019-03-01
use legacy build system to avoid xcode 10.1 issues

jakubgs:
Since I've upgraded all MacOS hosts to XCode 10.1 which uses a new build system various errors have showed up:
```
error: Build input file cannot be found: ... bignum.cc
```
Or:
```
error: Multiple commands produce
```

I'm forcing use of the legacy build system as the simplest and temporary fix to build issues.

For more details read:
* https://github.com/facebook/react-native/issues/20492
* https://github.com/facebook/react-native/issues/21168

### DIFF
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -39,9 +39,9 @@ pipeline {
         stage('iOS') { steps { script {
           ios = cmn.buildBranch('status-react/combined/mobile-ios')
         } } }
-        stage('iOS e2e') { steps { script {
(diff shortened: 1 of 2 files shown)


cammellos (approved):

mandrigin (approved):

pedropombeiro (approved):

jakubgs:
It seems like my fix doesn't fix e2e builds:
```
❌  ld: symbol(s) not found for architecture i386
❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
According to @mandrigin we can disable those for now since they are not that useful anyway.

PR #7628, opened 2019-03-01
fix/#4158 new messages counter being shown for already seen messages

### LINKED ISSUE #4158: New messages counter is shown for already seen messages

### User Story

As a user I don't want to see new messages counter when all messages are already seen

### Description

*Type*: Bug

*Summary*: New messages counter doesn't disappear until navigating to home view if a new message was received when viewing another user profile or group chat info. Even if a user already read all new messages and sent new one on the chat the counter is still present

#### Expected behavior
New messages counter disappears when a user navigates to a chat with new messages

#### Actual behavior
New messages counter is still shown when user returns to a chat with new messages from profile or group info view

### Reproduction

One-to-one chat:
- Open Status on 2 devices
- Create 2 users and start a chat between them
- As user A tap `Chat options` -> `View profile`
- As user B send a message to user A
- As user A return to the chat view and send a message to user B

Group chat:
- Open Status on 2 devices
- Create 2 users
- Create a group chat
- As user A click `Chat options` -> `Group info`
- As user B send a message in the group chat
- As user A return to the chat view and send a message in the group chat

Public chat:
- Open Status on 2 devices
- Create 2 users and join the same public chat
- As user B send a message in the public chat
- As user A tap user's B picture and open his profile
- As user B send one more message in the chat
- As user A return to the chat view and send a message

### Additional Information
* Status version: `6fc03fc2`, nightly build from May 8

bitsikka:
Fixes #4158

status: WIP

### DIFF
--- a/src/status_im/ui/screens/navigation.cljs
+++ b/src/status_im/ui/screens/navigation.cljs
@@ -115,16 +115,27 @@
    {:db (assoc db :modal modal-view)}))
 
 (fx/defn navigate-back
-  [{{:keys [navigation-stack view-id] :as db} :db}]
-  (assoc
-   {::navigate-back nil}
-   :db (let [[previous-view-id :as navigation-stack'] (pop navigation-stack)
-             first-in-stack (first navigation-stack)]
-         (if (= view-id first-in-stack)
-           (-> db
-               (assoc :view-id previous-view-id)
-               (assoc :navigation-stack navigation-stack'))
-           (assoc db :view-id first-in-stack)))))
+  [{:keys [db] :as cofx}]
+  (let [{:keys [navigation-stack view-id current-chat-id]}
+        db
+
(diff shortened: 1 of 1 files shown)


bitsikka:
@yevh-berdnyk, @StatusSceptre, I see Jenkins builds are failing, but can I please get a review and some feedback?

bitsikka:
Jenkins builds failing at least because of the use of `status-im.chat.models/mark-messages-seen`. I cannot :require `status-im.chat.models` in `status-im.ui.screens.navigation` because of circular dependency.

Other then that it builds fine in development builds for me.

It seems one possible way would be to move `mark-messages-seen` and it's dependencies to `navigation` and update it's usage in other places. Does not seem like a good idea. 🤔 

 I would appreciate any direction/ideas.

bitsikka:
closing this PR for a better one

PR #7629, opened 2019-03-01
Disable font scaling

### LINKED ISSUE #7215: Buttons offscreen for "App" would like to connect to your wallet
# Problem
When approving a new site on an iPhone XS Max, the buttons to approve/deny are pushed off the bottom of the screen.

You can see just the top of them which is barely enough to tap, but you can't read the text so you have to guess what the buttons say.

img!UNITO-UNDERSCORE!7777

rasom:
fix #7215

Let's disable font scaling till we handle it properly. It looks off at the moment:

screen shot 2019-03-01 at 09 16 51
screen shot 2019-03-01 at 09 10 09

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -101,16 +101,19 @@
 
 (defn text
   ([t]
-   [text-class t])
+   [text-class {:allowFontScaling false} t])
   ([opts t & ts]
    (->> (conj ts t)
         (transform-to-uppercase opts)
-        (concat [text-class (add-font-style :style opts)])
+        (concat [text-class (add-font-style
+                             :style
+                             (assoc opts :allowFontScaling false))])
         (vec))))
 
 (defn text-input [{:keys [style] :as opts} text]
   [text-input-class (merge
                      {:underline-color-android :transparent
+                      :allowFontScaling        false
                       :placeholder-text-color  colors/text-gray
                       :placeholder             (i18n/label :t/type-a-message)
                       :value                   text}


errorists:
hey @rasom , just curious, is there anything that needs to be done to have it working properly? It is a big factor in accessibility for visually impaired.

flexsurfer (approved):

rasom:
@errorists it's hard for me to say at the moment. We probably need to specify which text resources should be scalable, which not, and carefully enable scalable resources and test them so that it doesn't look as weird as now. Not sure that there is a single thing which can be done in order to handle the issue, but I don't have experience in this.

errorists:
@rasom thanks, I'll make a note to research into it and either see if we can support OS level font scaling like Dynamic Type on iOS, or in future implement a font-scale slider inside our settings. I think allowing to resize text in the range of .75 - 1.25x multiplier will be plenty sufficient and won't cause weirdness like on your screenshot, but again will have to look into it.

rasom:
@errorists there is `maxFontSizeMultiplier` prop https://facebook.github.io/react-native/docs/text#maxfontsizemultiplier which probably can help handle scaling properly

<blockquote><div><strong><a href="https://facebook.github.io/react-native/index.html">Text · React Native</a></strong></div><div>A React component for displaying text.</div></blockquote>

rasom:
We can set 1.5x limit and see if it looks fine, as alternative to this PR

mandrigin:
@rasom let's fix it instead of disabling, because according to some devs a lot of people use it: https://twitter.com/marcoarment/status/1099295178651783169

<blockquote><div>Twitter</div><div><strong><a href="https://twitter.com/marcoarment/status/1099295178651783169">Marco Arment on Twitter</a></strong></div><div>“I recently added accessibility analytics to Overcast, and my numbers are even more dramatic: 

40% of Overcast users use a non-default system font size.

Dynamic Type support isn’t optional and can’t be a low priority for any iOS developer these days. https://t.co/NjZVRsdoUg”</div></blockquote>

rasom:
@mandrigin well, sure, let's fix it if we can do it by just setting a limit

Otherwise, if it requires a set of changes, it would be better to make it at least usable, even if not scaled (because it's not usable for some ppl at the moment), and then apply changes one by one.

errorists:
I'm inclined to agree to both options, I think doing it properly will require setting max scale value for each text object separately, like if you look at the screenshot for example, the circle with initial letter of each chat, that shouldn't scale at all, so it might require more work and I'm not sure how well our components respond to layout changes due to text scale, might get ugly :) I was thinking of maybe doing an accessibility audit before the public release and addressing it nearer that.

mandrigin:
@rasom I'm all up for for setting limits. I'm against disabling it completely :-D

mandrigin:
@errorists I think having it in the "ugly but works" state is okay for now (for beta). Way better than not having that feature at all.
Also, I agree that we need to have some kind of accessibility audit of our application before release.

errorists:
@mandrigin I'm fine with ugly, it's the cutoff text not fitting inside containers I worry about since that you can't read at all regardless of size :)
I'd start with a 1.25 scale limit and see how that's like

rasom:
@mandrigin @errorists will file PR with 1.5x limit in 10 mins (or just update this)

rasom:
nope, I will not file PR :D

`maxFontSizeMultiplier` is available only in RN 0.58+

So, we can fix #7215 by just disabling scaling for that particular view. And then return to this issue after upgrading RN.

mandrigin:
it's time to upgrade react-native... and we don't need any custom patches anymore, btw.

mandrigin:
so we can just use upstream version

rasom:
@mandrigin ok

then I'm closing this PR and we will get back to this problem after upgrading RN. Hopefully before next release.

Issue #23, opened 2016-04-25
Grey Area between staging and text input


0xc1c4da:
Maybe we should remove this staging area entirely? Let's have discussion in #status

https://assembly-vc.slack.com/files/jarradhope/F135P6S3F/greybetweenstagingareandtextinput.png

PR #7637, opened 2019-03-01
Add unpinning of mailservers

cammellos:
### Pinned mailservers

Currently once a user select a mailserver, that mailserver will be "pinned". This means that we won't be automatically selecting a different mailserver in case there are some connection problems. We will instead show a pop-up asking the user to retry or change mailserver.

The issue currenty is that once you pin a mailserver, there's no way back, so you will be presented with the pop up everytime.

This PR adds a "Unpin" feature, that allows user to forget their preference, so that the pop up won't be displayed, and we'll rotate the changes.

### Testing

1) Disable automatic selection
2) Replicate failed requests to the mailserver (trying going offline/Fetch 24 hours, non existent mailserver)
3) You should see a pop up
4) Enable automatic selection
5) It should rotate automatically without asking you.

New accounts should have automatic selection enabled.

For now only implemented on mobile, will implement on desktop in a separate PR.

Known issues:
- When enabling automatic selection, it will change mailserver.

@errorists it looks ugly, I have added a couple of screenshots, i i's under advanced, but for sure can be improved upon, what do you think, is it ok for a rough cut or there's some magic to apply?

disabled
enabled

status: ready


errorists:
@cammellos hey, could you take a look at [this discussion](https://github.com/status-im/status-react/issues/6393) and see if the designs proposed there are relevant to what you want to achieve?

mandrigin:
Yeah, I agree with @errorists if this checkbox isn't hard to implement, it is much more obvious. Otherwise, it is a super important thing to have.

mandrigin (approved):

cammellos:
@errorists is ok, I will probably use a checkbox, as those toggles currently have weird issues that results in it being ON while it's actually OFF sometimes https://github.com/status-im/status-react/issues/7619 , in order to minimize the risk. Would that be ok?

errorists:
@cammellos yeah, a checkbox would be ok given the issues, just please make sure to hide the remaining mail servers when automated selection is toggled on to make it more clear.

cammellos:
@errorists the only issue with not showing the other mailservers when is automatic, is custom mailservers.

Essentially I can create/edit a mailserver. If automatic is selected, after I have created it, I won't be seeing it in the list, which is a bit odd, and might give the impression to the user that it was not created.

Maybe we should only hide normal mailserver, and still display custom ones? (most of the people don't have custom mailservers)

errorists:
Yeah I see what you mean, could you in this case decrease opacity instead of hiding, a value of 0.4 would be enough, this way you'd see anything you add but still get the correct impression that manual selection is disabled and we're doing it for you.

rasom (approved):

cammellos:
@errorists updated screenshots

errorists:
thanks @cammellos no further comments on my side

annadanchenko:
@cammellos can see err="very old message" in geth.log and it took ~4 minutes to retrieve a history for "testanna" public chat. Clock on the phone is set to automatic, so should be ok. Re-installing to see if can reproduce.

[geth.log](https://github.com/status-im/status-react/files/2919902/geth.log)
[Status.log](https://github.com/status-im/status-react/files/2919903/Status.log)

annadanchenko:
@cammellos checked same on release 0.10.0 with same device:

### install 1, join "testanna" public chat, wait for history
it took ~2 minutes to get history of the same chat. there are no  err="very old message" in geth.log
[geth.log](https://github.com/status-im/status-react/files/2919985/geth.log)
[Status.log](https://github.com/status-im/status-react/files/2919986/Status.log)

### install 2, join "testanna" public chat, wait for history
it took ~5 minutes to get history and geth.log contains  err="very old message"
[geth.log](https://github.com/status-im/status-react/files/2920033/geth.log)
[Status.log](https://github.com/status-im/status-react/files/2920035/Status.log)

cammellos:
@dshulyak @adambabik this "very old message" it's expected every now and then, is that correct? just to make sure (it means they send us an envelope that is +-20s?)

mandrigin:
@annadanchenko but does it happen on `develop` branch? It seems unlikely this PR will affect such functionality...

cammellos:
@mandrigin yes, happens on release as well, not this PR specific (the "very old message")

annadanchenko:
> but does it happen on `develop` branch? It seems unlikely this PR will affect such functionality...

most likely it's on "develop" too.  Because this PR also upgrades status-go (so similar to https://github.com/status-im/status-react/pull/7620) I was just sharing my findings as I test.

annadanchenko:
@cammellos I'm still testing , so far there are 2 issues:

### Issue 1 When tap on Automatic selection text nothing happens. 
Would expect it will change the state of checkbox but it doesn't. User needs to tap exactly on checkbox now to change its state.

### Issue 2 Can select mailserver when Automatic selection is ON. 

Steps:
- Sign in, open Mailservers list with Automatic selection turned on. 
- tap on some mailserver in the list (that is greyed out). As a result, mailserver is "highlighted"  with normal black text and Warning popup is shown asking if you want to connect to this mailserver. If tap on Confirm then it's selected in the list with black text that looks confusing - when both automatic selection and "pinned mailserver" are present


Can be 2 possible solutions:
1. if automatic selection is ON then do nothing if tap on any mailserver. So, user needs to set it off to choose a mailserver
2. if tap on mailserver while automatic selection is on then switch checkbox to OFF and select mailserver.

adambabik:
@cammellos yes, that's possible if the local clock is skewed. It happens when the absolute difference is >20s from the server time.

But we sync time on mobile using NTP unless it was not enabled on mobile or NTP servers failed to respond.

annadanchenko:
ok, issues 1-2 are fixed (from https://github.com/status-im/status-react/pull/7637#issuecomment-468738491)

churik:
@cammellos I see that it will be implemented in desktop in separate PR, but for now there is no way to select mailserver manually (autoselection is enabled by default).
Not sure that it is right, I would prefer to have possibility to change mailserver.
WDYT?
maildf

cammellos:
@churik I will make them selectable on desktop for now, is that ok?

churik:
@cammellos yes, I believe that's OK for now

cammellos:
@churik I have added the checkbox to desktop, it's not very visible, but given it's on by default it should be ok for now, I can address in a later PR, what do you think?

annadanchenko:
@cammellos new build 5 is crashing on iOS upon app 1st app launch. ON Android it doesn't... colllecting logs


[StatusIm  04-03-2019, 12-27.crash.txt](https://github.com/status-im/status-react/files/2926151/StatusIm.04-03-2019.12-27.crash.txt)

cammellos:
thanks, weird as no code changes were made to the ios codebase, possibly something infrastructure related, thanks for the logs.

mandrigin:
@annadanchenko can you double-check if a release branch's latest iOS build have the same problem?

annadanchenko:
@mandrigin I can't install ipa file from https://ci.status.im/job/status-react/job/release/job/release%252F0.10.x/23/ via xCode, it says

The executable was signed with invalid entitlements.
The entitlements specified in your application’s Code Signing Entitlements file are invalid, not permitted, or do not match those specified in your provisioning profile. (0xE8008016).

<blockquote><div><strong><a href="https://ci.status.im/job/status-react/job/release/job/release%252F0.10.x/23/">status-react » release » release/0.10.x #23 [Jenkins]</a></strong></div></blockquote>

If use TestFlight ok, build 0.10.1 (2019030411) can be installed and launched fine - no crash

annadanchenko:
on build 6 iOS app can be launched fine

churik:
Tested on Mac OS, Windows, Linux:
- 1-1, group and public chats;
- pairing;
- resend messages;
- fetching messages from offline mode;
- compatibility with 0.10.0
- updating profile info;
- changing mailserver / disabling-enabling autoselect mode;
- logout;
- restoring app from seed phrase;
- wrong password on login;
- disabling/enabling logging.

No issues from my side.

asemiankevich:
i've checked pairing functionality - looks good to me

Serhy:
Tested group chat on iOS <-> Android and communication between PR build (Android 8.1) and Release 0.10.0 (iOS 12.1.2) in 1-1/public/Group chats
Will add an update to e2e test here (false failed on an attempt to switch mailserver - because we disable ability to manually select mailserver if 'Autoselect mailserver' enabled)

mandrigin:
@cammellos I need your help with cherry-picking to the release branch, it has some merge conflicts.

cammellos:
it relies on a previous pr for mailservers, safe to include as throughly tested, I will do that, which one is the release branch?

mandrigin:
the branch is still `release/0.10.x`

cammellos:
@mandrigin cherry picked and pushed

PR #7644, opened 2019-03-01
Remove :wallet-modal screen as it is not used anymore

rasom:
status: ready


flexsurfer:
Why is it not used anymore?
Edit: ah because we don't lose the state anymore, so we can switch to wallet tab and back to the browser for example without losing state, ok

flexsurfer (approved):

mandrigin (approved):

rasom:
The only thing which might have regressions here is Wallet screen.

churik:
Tested (IOS, Android):
- no wallet modal in DApp and in chat
- /send , /request command from chat
- send asstets from wallet
- send transaction to DApp

Some issues are found, but they are not introduced by particular PR and will be reported separately, so it is ready to merge.

PR #7677, opened 2019-03-05
Update ko.json

jinhojang6:
Korean translations update


rasom (approved):

rasom:
already tested by @jinhojang6

jakubgs (commented):
This PR would benefit from running this file through `jq --sort-keys`.

goranjovic (approved):

rasom:
@jakubgs I'm not sure it actually matters because usually these files are generated by Lokalise and it might be just changed again in the next PR, not very important

jakubgs:
Yeah, which is why I didn't block the PR on this.
Maybe we'd like a pre-commit check that would sort those in the future.

churik:
Korean translations are updated in the build, checked on IOS 11.4.1 and Android 8

Issue #5281, opened 2018-07-24
Days in chat doesn't take into account user timezone


churik:
### User Story

As a user, I want to see in chat history split by days according to my timezone (always started from 00:00 and to 00:00 next day)

### Description

*Type*: Bug
*Summary*: in public, 1-1 chats day for every user starts at 00:00 UTC. this means that i.e. "Today" for the user from Tokyo starts at 10 AM, all messages sent before 10 AM were placed in "Yesterday" history.

#### Expected behavior
Every day starts from 00:00
(screen from Telegram, UTC +3)
![right_beh](https://user-images.githubusercontent.com/4557972/43140308-5d08ed86-8f5c-11e8-9408-c360eed1311b.png)

#### Actual behavior
Every day starts from UTC 00 (in my case from 3 AM)
(screen from Status, UTC +3)
![img_d639ad88272e-1 jpeg 2018-07-24 15-44-56](https://user-images.githubusercontent.com/4557972/43140416-9e195504-8f5c-11e8-93d6-97b31e0c9267.png)

### Reproduction
*Prerequisite: you should be located not in UTC 00*
- Open Status
- Join any public channel
- Send message in 00:01

OR
- join #status channel
- check messages for 23-24 Jul

### Additional Information

* Status version: status nightly 23/07/18 

* Operating System: Android, iOS

churik:
Still reproducible

PR #7678, opened 2019-03-05
Fix datemark calculation not taking into account local time

cammellos:
Fixes: #5281

status: ready


yenda (approved):

rasom (approved):

churik:
@cammellos 
see the same issue (checked on IOS and Mac OS build).

I'm located in GMT+1, but yesterday and today are split by GMT 00 (checked in #status channel)
kjlsdfd

cammellos:
@churik should be fixed in the next build, sorry for the inconvenience.

churik:
Thanks @cammellos!
Works perfectly on IOS , Android and Mac OS.
Also checked case when manually select timezone.

PR #7691, opened 2019-03-06
[#7527] Extension is not shown in chat if Development mode is switche…

### LINKED ISSUE #7527: Extension is not shown in chat if Development mode is switched off
### Description
_Type_: Bug

_Summary_: Extension command and its chat message is not shown after user installed the extension and relaunched the app if development mode is OFF. If user tries to install extension again by tapping on its placeholder message then popup is shown that extension is already installed.

So, in order to have extension working user needs to turn ON development mode. Because user can install extension by scanning QR code or by tapping on extension's message in the chat (e.g. for Gfycat) and have no idea about development mode we need to warn user about setting development mode on or find other solution.

We also might add some info on the message placeholder "To see this message install the extension" to inform about turning on development mode.

So, there should be solution that explains to the user why extension is not shown even if it's installed for 2 cases:

1. user never turned on development mode
1. user turned off development mode after installing an extension

after user relaunch the app:
img!UNITO-UNDERSCORE!2814

### Reproduction

* Install Status and create an account
* Install Gfycat extension from https://status-im.github.io/pluto/try.html?hash=Qmb1B3jXNdc9WZCcWFzpkTXtrurnKxYQFvUDhp2J4SVCCX
* send some Gfycat image to the public chat
* Close the app
* open app again and check if you can see the image and Gfycat command in the chat
Note: if you will turn on development mode then extension image will be shown again as expected. If you will turn it off and relaunch the app then instead of the image there is placeholder with text "to see this message Install the extension"

### Additional Information

* Status version: 0.9.33 and 0.9.34

flexsurfer:
fixes #7527

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -594,7 +594,7 @@
 (handlers/register-handler-fx
  :extensions.ui/install-extension-button-pressed
  (fn [cofx [_ url]]
-   (extensions.registry/load cofx url true)))
+   (extensions.registry/install-from-message cofx url true)))
 
 (handlers/register-handler-fx
  :extensions.ui/install-button-pressed
--- a/src/status_im/extensions/registry.cljs
+++ b/src/status_im/extensions/registry.cljs
@@ -89,6 +89,17 @@
                                      :active? true}]
                        :follow-up  (if modal? :extensions/stage-modal :extensions/stage)}}))
 
+(fx/defn install-from-message
+  [cofx url modal?]
+  (if (get-in cofx [:db :account/account :dev-mode?])
+    (load cofx url modal?)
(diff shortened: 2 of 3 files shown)


jeluard (approved):

dmitryn (approved):

churik:
Checked on Gfycat example (IOS 11.4.1, Android 8): 
- user never turned on development mode - can install extension, message turning on Dev mode is shown, development mode is turned on;
- user  turned on development mode - can install extension, no message;
- user switch off development mode and tap on message - message about turning on Dev mode  is shown, development mode is turned on, gif is shown;
- user relogin with switched on dev-mode -  gif is shown normally

PR #7666, opened 2019-03-05
[iOS, Android] Upgrade RN to 0.59.2

rasom:
status: ready

PR #7698, opened 2019-03-08
upgrade android ndk to r17c

jakubgs:
Required by #7666
I've pushed the image and it's already up:
https://cloud.docker.com/u/statusteam/repository/docker/statusteam/status-build-android/tags

### DIFF
--- a/.TOOLVERSIONS
+++ b/.TOOLVERSIONS
@@ -1,4 +1,4 @@
-android-ndk;r10e;070be287539e3e7706f8dabfb6bf9879
+android-ndk;r17c;a4b6b8281e7d101efd994d31e64af746
 android-sdk-build-tools;28.0.1;
 android-sdk-platform;android-27;
 android-sdk;4333796;aa190cfd7299cd6a1c687355bb2764e4
--- a/.TOOLVERSIONS
+++ b/.TOOLVERSIONS
@@ -1,6 +1,6 @@
 android-ndk;r17c;a4b6b8281e7d101efd994d31e64af746
-android-sdk-build-tools;28.0.1;
-android-sdk-platform;android-27;
+android-sdk-build-tools;28.0.3;
+android-sdk-platform;android-28;
 android-sdk;4333796;aa190cfd7299cd6a1c687355bb2764e4
 clojure_cli;1.9.0.381;
 cmake;3.12.2;
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -3,7 +3,7 @@ pipeline {
     docker {
       label 'linux'
       /* WARNING: remember to keep this up-to-date with the value in docker/android/Makefile */
-      image 'statusteam/status-build-android:1.1.1-d3611a31'
+      image 'statusteam/status-build-android:1.1.1-481a7564'
       args (
         "-v /home/jenkins/tmp:/var/tmp:rw "+
         "-v /home/jenkins/status-im.keystore:/tmp/status-im.keystore:ro"
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -3,7 +3,7 @@ pipeline {
     docker {
       label 'linux'
       /* WARNING: remember to keep this up-to-date with the value in docker/android/Makefile */
-      image 'statusteam/status-build-android:1.1.1-cd1596b3'
+      image 'statusteam/status-build-android:1.1.1-d3611a31'
       args (
         "-v /home/jenkins/tmp:/var/tmp:rw "+
         "-v /home/jenkins/status-im.keystore:/tmp/status-im.keystore:ro"
--- a/docker/android/Makefile
+++ b/docker/android/Makefile
@@ -17,7 +17,7 @@ SDK_PLATFORM_VERSION=$(call  __toolversion, android-sdk-platform)
 SDK_BUILD_TOOLS_VERSION=$(call  __toolversion, android-sdk-build-tools)
 
 # WARNING: Remember to change the tag when updating the image
-BASE_IMAGE_TAG = $(shell cd $(GIT_ROOT)/docker/base && make get-image-tag)
+BASE_IMAGE_TAG = $(shell cd $(GIT_ROOT)/docker/base && $(MAKE) get-image-tag --no-print-directory)
 DEPS_HASH = $(shell $(GIT_ROOT)/scripts/gen-deps-hash.sh -b $(BASE_IMAGE_TAG) \
 					-d android-ndk \
 					-d android-sdk \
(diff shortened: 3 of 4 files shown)


rasom (approved):

jakubgs:
Doesn't seem the codebase likes the upgrade:
```
:react-native-android:buildReactNdkLib
A problem was found with the configuration of task ':react-native-android:buildReactNdkLib'. Registering invalid inputs and outputs via TaskInputs and TaskOutputs methods has been deprecated and is scheduled to be removed in Gradle 5.0.
 - File '/home/jenkins/workspace/status-react_prs_android_PR-7698/node_modules/react-native/ReactAndroid/src/main/jni/react' specified for property '$1' is not a file.
/usr/lib/android-ndk/build/core/setup-app-platform.mk:135: *** Android NDK: Aborting    .  Stop.
Android NDK: APP_PLATFORM set to unknown platform: android-16.    
make[1]: Entering directory `/home/jenkins/workspace/status-react_prs_android_PR-7698/node_modules/react-native/ReactAndroid/src/main/jni/react/jni'
make[1]: Leaving directory `/home/jenkins/workspace/status-react_prs_android_PR-7698/node_modules/react-native/ReactAndroid/src/main/jni/react/jni'
:react-native-android:buildReactNdkLib FAILED

* What went wrong:
Execution failed for task ':react-native-android:buildReactNdkLib'.
> Process 'command '/usr/lib/android-ndk/ndk-build'' finished with non-zero exit value 2
```
Not sure what `APP_PLATFORM` does but we set it to `android-18`:
https://github.com/status-im/status-react/blob/38aca129cd65216078381d10abb7fb7347fce427/android/app/jni/Application.mk#L1-L4

pedropombeiro:
Wondering if we should do this after we've merged the `feature/nix` branch

rasom:
@PombeirP why?

.TOOLVERSIONS
```diff
@@ -1,5 +1,5 @@
-android-ndk;r10e;070be287539e3e7706f8dabfb6bf9879
-android-sdk-build-tools;28.0.1;
+android-ndk;r17c;a4b6b8281e7d101efd994d31e64af746
+android-sdk-build-tools;28.0.3;
 android-sdk-platform;android-27;
```
rasom: @PombeirP we also need to use android-28, can we change this as well? Or this config doesn't have effect on docker image?

pedropombeiro: OK, I'll push a new image with that change too. In your #status-core message you mentioned

> we need to upgrade android-sdk-build-tools to 28.0.3 and android-sdk-platform to android-27

that's why I didn't change it

pedropombeiro:
> @PombeirP why?

Because all these dockerfiles disappear in that branch, so we'll need to redo this work in Nix before merging to develop.

.TOOLVERSIONS
rasom: oh I see, that was a mistake, sorry

rasom:
> so we'll need to redo this work in Nix before merging to develop

ok, I see. The problem here is that it probably will be a bit hard to maintain this PR because it's quite big, so if we switch to Nix soon then no problem to merge it after. Otherwise it might be a bit tricky with conflicts, already rebased it twice (though without huge difficulties).

rasom:
> to maintain this PR because it's quite big

I mean another PR https://github.com/status-im/status-react/pull/7666 :)

.TOOLVERSIONS
```diff
@@ -1,6 +1,6 @@
-android-ndk;r10e;070be287539e3e7706f8dabfb6bf9879
-android-sdk-build-tools;28.0.1;
-android-sdk-platform;android-27;
+android-ndk;r17c;a4b6b8281e7d101efd994d31e64af746
+android-sdk-build-tools;28.0.3;
+android-sdk-platform;android-28;
 android-sdk;4333796;aa190cfd7299cd6a1c687355bb2764e4
```
rasom, line 4: should this line be changed? how do we manage this file btw? is this some docker related stuff?

.TOOLVERSIONS
```diff
@@ -1,6 +1,6 @@
-android-ndk;r10e;070be287539e3e7706f8dabfb6bf9879
-android-sdk-build-tools;28.0.1;
-android-sdk-platform;android-27;
+android-ndk;r17c;a4b6b8281e7d101efd994d31e64af746
+android-sdk-build-tools;28.0.3;
+android-sdk-platform;android-28;
```
rasom, line 3: btw it seems that `android-26` is hardcoded here https://github.com/status-im/status-react/blob/6d3211de46f302ab1a92eb3bb711733dd9cb8d9f/docker/android/Dockerfile#L13

pedropombeiro: I'm not too familiar with the Android SDK dependencies, I assume that we need this baseline 26 platform for some reason.

.TOOLVERSIONS
pedropombeiro: This file is read when generating the Dockerfile to figure out which files to incorporate in it. It is also used to tag the Docker image with a different hash, depending on the values we put here.

rasom: ok, understood

.TOOLVERSIONS
jakubgs: Yeah, it is, that's kinda how `sdkmanager` works, the packages included are named after the version of sdk. I guess it could be parametrized.

jakubgs: Ah wait, it actually already is parametrized right below it:
https://github.com/status-im/status-react/blob/6d3211de46f302ab1a92eb3bb711733dd9cb8d9f/docker/android/Dockerfile#L14
So my guess would be that when I was building this it wouldn't work if also `platform:android-26` was not installed.

pedropombeiro:
When I build the app locally, I do see this:

```
Configuration 'compile' in project ':react-native-image-crop-picker' is deprecated. Use 'implementation' instead.
WARNING: The specified Android SDK Build Tools version (26.0.1) is ignored, as it is below the minimum supported version (26.0.2) for Android Gradle Plugin 3.0.1.
Android SDK Build Tools 26.0.2 will be used.
To suppress this warning, remove "buildToolsVersion '26.0.1'" from your build.gradle file, as each version of the Android Gradle Plugin now has a default version of the build tools.
```

The `buildToolsVersion "26.0.1"` directive is in `node_modules/react-native-image-crop-picker/android./build.gradle`

Maybe we should just update that module. cc @jeluard

jakubgs:
Considering we've merged the Nix branch and the way to manage NDK is this file now:
https://github.com/status-im/status-react/blob/develop/scripts/lib/setup/nix/mobile/android-ndk/default.nix
This PR is no obsolete.

PR #7701, opened 2019-03-08
support style in link and map-link components in extensions

tbenr:
@jeluard @flexsurfer 

as discussed here: https://github.com/status-im/status-react/pull/7404#issuecomment-470143806

I implemented the ability to override the standard style for links.
I also added some new features to the link component:

`link  {:value link :properties {:uri :string :text? :string :open-in? {:one-of #{:device :status}}}}`

link definition now supports:
- displaying a `text` different from the `uri`
- with `open-in` you can choose to open the link on the device or in status browser. if you don't specify, dialog is shown to the user who choose how to open it.

extension example is here: https://get.status.im/extension/ipfs@QmXv1ZzHbJYiA7n8TM8aUaAQsW5Cm8jHj3MBmshVNP59qW

image

### DIFF
--- a/src/status_im/extensions/core.cljs
+++ b/src/status_im/extensions/core.cljs
@@ -25,7 +25,8 @@
             [status-im.utils.ethereum.tokens :as tokens]
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.chat.commands.sending :as commands-sending]
-            [status-im.browser.core :as browser]))
+            [status-im.browser.core :as browser]
+            [status-im.utils.platform :as platform]))
 
 (re-frame/reg-fx
  ::identity-event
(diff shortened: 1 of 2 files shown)


tbenr:
I added `[status-im.utils.http :as http]` but at the end i didn't used... 👎 
that's because i removed `http/normalize-url` from `#(.openURL react/linking uri)` because it doesn't support `geo:` android native URL protocol.

dmitryn (approved):

jeluard:
@tbenr Can you remove the `require` then? :)

jeluard (approved):

tbenr:
@jeluard done ;)

PR #7702, opened 2019-03-08
Add contact request message

cammellos:
When pfs is enabled, if a contact code has not been already fetched, the
user will be presented with a different input, and it will be asked to
sent ad contact request message instead.

This message will be displayed differently and it will give the user an
option to accept the contact request, or block the contact.

Currently this PR does not limit the number of contact requests that can
be sent, it will be done in a later PR.
All the code is behind a the pfs toggle feature flag.

Sorry for the largish pr, but I had to code it through before settling on an implementation.

### Testing

All the code is behind a feature flag, so only regression testing on:
- Sending 1-to-1/group-chats/public-chat messages
- Commands
- Contact requests

status: ready
accepted
received
sent


src/status_im/chat/models.cljs
```diff
@@ -228,6 +229,8 @@
     (fx/merge cofx
               {:db (-> (assoc db :current-chat-id chat-id)
                        (set-chat-ui-props {:validation-messages nil}))}
+              #(when-not (get-in db [:chats chat-id :group-chat])
```
cammellos, line 232: test this

src/status_im/chat/models/input.cljs
```diff
@@ -138,6 +139,20 @@
                 (set-chat-input-text nil)
                 (process-cooldown)))))
 
+(defn contact-request-message-fx
```
cammellos, line 142: test this

src/status_im/chat/models/input.cljs
```diff
@@ -161,15 +176,24 @@
   "Sends message from current chat input"
   [{{:keys [current-chat-id id->command access-scope->command-id] :as db} :db :as cofx}]
   (let [{:keys [input-text custom-params]} (get-in db [:chats current-chat-id])
+        contact-request? (and
```
cammellos, line 179: test this

src/status_im/chat/models/message.cljs
```diff
@@ -351,12 +352,19 @@
 
 (def ^:private transport-keys [:content :content-type :message-type :clock-value :timestamp])
 
-(fx/defn upsert-and-send [{:keys [now] :as cofx} {:keys [chat-id from] :as message}]
-  (let [send-record     (protocol/map->Message (select-keys message transport-keys))
-        old-message-id  (transport.utils/old-message-id send-record)
-        wrapped-record  (if (= (:message-type send-record) :group-user-message)
-                          (wrap-group-message cofx chat-id send-record)
-                          send-record)
+(fx/defn upsert-and-send [{:keys [now] :as cofx} {:keys [chat-id
```
cammellos, line 355: test this

src/status_im/contact_code/core.cljs
```diff
@@ -92,3 +95,38 @@
 
 (fx/defn publishing-failed [cofx]
   (log/warn "failed to publish contact-code"))
+
+(defn- parse-response [response-js]
+  (-> response-js
+      js/JSON.parse
+      (js->clj :keywordize-keys true)))
+
+(fx/defn add-contact-code [{:keys [db]} chat-id contact-code]
```
cammellos, line 104: test this

yenda:
Could you provide some screenshots for this PR as it seems there is a UI part to it?

cammellos:
@yenda attached screenshots, although the feature is not to be tested yet, as not completed (it's behind a feature flag)

rasom (approved):

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -1013,6 +1013,26 @@ public void run() {
     StatusThreadPoolExecutor.getInstance().execute(r);
   }
 
+  @ReactMethod
+  public void getContactCode(final String identity, final Callback callback) {
```
mandrigin, line 1017: hmm, why is it only for Android?

cammellos: good, point, totally forgot about desktop and ios, thanks!

cammellos:
@mandrigin ready to be reviewed again

modules/react-native-status/ios/RCTStatus/RCTStatus.m
```diff
@@ -254,6 +254,16 @@ - (void)handleSignal:(NSString *)signal
 #endif
 }
 
+//////////////////////////////////////////////////////////////////// getContactCode
+RCT_EXPORT_METHOD(getContactCode:(NSString *)identity
+                  callback:(RCTResponseSenderBlock)callback) {
+  NSString* result = StatusgoGetContactCode(identity);
```
mandrigin:
```suggestion
  NSString *result = StatusgoGetContactCode(identity);
```

mandrigin (approved):
one nitpick, looks good otherwise!

src/status_im/accounts/update/core.cljs
```diff
@@ -25,7 +25,7 @@
     (let [updated-chat  (assoc chat :resend? "contact-update")
           tx            [(transport-store/save-transport-tx {:chat-id chat-id
                                                              :chat    updated-chat})]
-          success-event [:transport/contact-message-sent chat-id]]
+          success-event [:transport/contact-message-sent chat-id nil]]
```
yenda, line 28: why do you need to add nil here? Isn't it going to be nil anyway if you don't provide it?

src/status_im/transport/message/contact.cljs
```diff
@@ -21,11 +21,3 @@
   (validate [this]
     (when (spec/valid? :message/contact-update this)
       this)))
-
-(fx/defn remove-chat-filter
```
yenda, line 25: when did this function became useless? shouldn't we be removing chat-filter when we delete and/or block a contact?

cammellos: I believe this function was used when we used a symmetric key for each contact ( we don't anymore), so it's not used currently

src/status_im/accounts/update/core.cljs
cammellos: not sure, I could check, but i'd rather be explicit, there's no harm and it's clearer imo

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -111,6 +111,52 @@
...
 
+(defview contact-request-message
+  [{:keys [chat-id message-id content from timestamp-str group-chat outgoing current-public-key expanded?] :as message}]
+  (letsubs [{:keys [last-updated
+                    pending?]} [:contacts/contact-by-identity chat-id]]
+    (let [sent-by-us? (= current-public-key from)
+          pending-state? (if sent-by-us?
+                           ;; last-updated is only updated if we have been added to their contacts
+                           (not (pos? last-updated))
+                           pending?)]
```
yenda, line 122: could it be anything but true here? How could it not be in a pending state if the contact-request-message is not from us?

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -111,6 +111,52 @@
...
    {:justify-timestamp? true}])
 
+(defview contact-request-message
+  [{:keys [chat-id message-id content from timestamp-str group-chat outgoing current-public-key expanded?] :as message}]
+  (letsubs [{:keys [last-updated
+                    pending?]} [:contacts/contact-by-identity chat-id]]
+    (let [sent-by-us? (= current-public-key from)
+          pending-state? (if sent-by-us?
+                           ;; last-updated is only updated if we have been added to their contacts
+                           (not (pos? last-updated))
```
yenda, line 121: `not pos?` as in `zero?` or could it be `nil`?

src/status_im/ui/screens/chat/message/message.cljs
cammellos: ok, first case: 

if it's sent by us, we check whether the user has replied to it (`last-updated`) -> we show it as accepted, if we don't have a `last-updated` -> still pending

second case, not sent by us:

if pending? -> true we never added the user to contacts -> we show it as pending
if pending is false -> we already added the user to contacts -> we show it as accepted

or I am missing something?

src/status_im/ui/screens/chat/message/message.cljs
cammellos: it could be `nil` for now, but it is coherced to int from realm when loading from it, but I could just change that

src/status_im/ui/screens/chat/message/message.cljs
yenda: ok so with system-tag this can all be replaced by `pending?`

mandrigin:
@cammellos are you planning to do anything with this PR or should we just close it?

cammellos:
@mandrigin close it for now, I have my plate full so can't spend much time on it, but I will keep the branch. Thanks

PR #7708, opened 2019-03-11
Set maximum number of attempts in config

cammellos:
Messages will be retried for up to a minute, after which we should receive the same signal, and it will be marked as not confirmed.

### Testing

A bit difficult, but you could disconnect to a mailsever (try to connect to a fake one). Send a message, wait 20 seconds, connect to a different mailserver, if the connection is fast enough it should be marked as Sent.

I have tested this locally, and it seems to me the user experience is quite good, if it's the case, I think it might be worth enabling confirmations on release builds.

status: ready


rasom (approved):

yenda (approved):

dshulyak (approved):

Serhy:
Looks good to me!

Android, iOS, MacOS latest builds were tested with public, 1-1 and group chats.
With develop build I could reproduce 'Not confirmed' message status when creating new account -> join to chat right away -> send a message. And likelihood to get 'Not confirmed' were high. It's indeed now retrying for 60 seconds and if no luck for confirmation to be received - 'Not confirmed. Tap for options' status appears. 
The same scenario happens when trying to connect to fake mailserver (@cammellos scenario advice from description) -> send text message -> wait ~50 seconds -> connect to hongkong mailserver -> message "Confirmed".

Really better UX for message senders!

cammellos:
@Serhy thanks for testing, should we try to enable in  the next release? What do you think?

Serhy:
@cammellos Well, I don't any cons for not enabling it in next release. 
Pros are better UX with messaging, better message reliability (in terms that user sees less "fake" 'Not confirmed' message status)
By the way, it's improves https://github.com/status-im/status-react/issues/7559 issue - but not solves - it's just harder to face that issue :)

PR #7710, opened 2019-03-11
Add e2e for declining invitation to group chat

churik:
Automate [test case](https://ethstatus.testrail.net/index.php?/cases/view/5756).
Please, review carefully - the first contribution to e2e mobile tests.
Checked on SauceLabs, tests are passed


test/appium/views/chat_view.py
```diff
@@ -205,6 +205,18 @@ def __init__(self, driver):
         super(JoinChatButton, self).__init__(driver)
         self.locator = self.Locator.text_part_selector('JOIN GROUP')
 
+class DeclineChatButton(BaseButton):
+    def __init__(self, driver):
+        super(DeclineChatButton, self).__init__(driver)
+        self.locator = self.Locator.text_part_selector('Decline invitation')
+
+
+class ChatElementByTextPart(BaseText):
```
antdanchenko: Can we use ChatElementByText instead of adding ChatElementByTextPart?

test/appium/views/home_view.py
```diff
@@ -174,3 +179,7 @@ def open_status_test_dapp(self, allow_all=True):
             else:
                 status_test_dapp.deny_button.click()
         return status_test_dapp
+
+    def check_empty_home_view(self):
```
antdanchenko: no need to add the method, you should search that chat name is not display directly on [the line](https://github.com/status-im/status-react/pull/7710/files#diff-f6abb94023b68e6c7cd70e556ca21b17R193), since empty home screen label might change

test/appium/views/chat_view.py
```diff
@@ -447,6 +460,10 @@ def chat_element_by_text(self, text):
         self.driver.info("Looking for a message by text: '%s'" % text)
         return ChatElementByText(self.driver, text)
 
+    def chat_element_by_text_part(self, text):
```
Serhy: This method is redundant because of two things:
1) It searches for text part (text = 'left group chat') which could be done with generic find_text_part() method
2) It does not confirm whether *user_name* (username of who left the chat) is present.

I'd go with the way as test_group_chat_system_messages test does for system message

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -170,3 +176,32 @@ def test_add_new_group_chat_member(self):
...
+            chat_name = device_1_home.get_public_chat_name()
+            device_1_chat, device_2_chat = create_new_group_chat(device_1_home, device_2_home, chat_name)
+            device_2_chat.decline_invitation_button.click()
+
+            # device 2: check that chat is deleted and user is redirected to Home view
+            home_view = HomeView(self.drivers[1])
+            home_view.check_empty_home_view()
+
+            # device 1: check system message about leaving a group chat
+            if device_1_chat.chat_element_by_text_part(leave_message).is_element_displayed() is False:
```
Serhy: 'if not blah.is_element_displayed()' is better construction (note: is_element_displayed() already returns 'False'). It's more readable (and also it's PEP8 standard:) )

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -170,3 +176,32 @@ def test_add_new_group_chat_member(self):
...
+            leave_message = 'left the group'
+            message_for_device_2 = 'This message should not be visible for device 2'
+
+            device_1_home, device_2_home = create_users(self.drivers[0], self.drivers[1])
+            chat_name = device_1_home.get_public_chat_name()
+            device_1_chat, device_2_chat = create_new_group_chat(device_1_home, device_2_home, chat_name)
+            device_2_chat.decline_invitation_button.click()
+
+            # device 2: check that chat is deleted and user is redirected to Home view
+            home_view = HomeView(self.drivers[1])
```
Serhy: Unnecessary assignment. device_1_home already has HomeView (there is a create_user() which has get_home_view that has `return HomeView`)

Serhy (changes requested):

test/appium/views/chat_view.py
churik: Well, I thought about it - but I don't think that we need especially grab a username for this particular check. For this we need to go to Profile -> Grab username especial for this part. For leaving group chat IMO it is redundant.


According to find_text_part()  - seems I missed this method, thanks!

test/appium/views/home_view.py
churik: In this case, I guess we need a separate small test to check the empty view, because it can break from time to time (I faced with this twice, when this message is not presented). 
That's why I added separate method here - it can be helpful IMO. If you think that it is unnecessary I can delete it.

antdanchenko: We should rely on values which are created by us or cannot be changed in future when using search by text (e.g. timestamps in amounts for transaction, public keys, `3 words user names` etc.)

Searching product captions by text may lead to endless changing of expected value and flaky test

Better ways are:

1. Search for no ChatElements are present at all (if home.chat_name_text.find_elements: fail 'Home screen is not empty').
2. Add accessibility label for empty chat caption in product and get element by label instead of text

churik:
@antdanchenko @Serhy second review, please. Thanks a lot for comments, everything should be fixed.

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -3,20 +3,31 @@
...
+    return device_default_username
 
-def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
+def create_new_group_chat(device_1_home, device_2_home, chat_name):
     # device 2: get public key and default username
     device_2_public_key = device_2_home.get_public_key()
-    device_2_profile_view = device_2_home.profile_button.click()
-    device_2_default_username = device_2_profile_view.default_username_text.text
-    device_2_home.home_button.click()
+    device_2_default_username = get_default_username(device_2_home)
```
antdanchenko: we definitely don't need to do this action in all tests where are we going to create a group chat

churik:
@Serhy @antdanchenko last review hopefully :)

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -47,28 +66,26 @@ def test_create_new_group_chat(self):
...
-        device_2_chat.send_message(message_from_device_2)
+        for i in range(len(device_chats)):
+            device_chats[i].send_message(messages[i])
 
-        if not device_2_chat.chat_element_by_text(message_from_device_1).is_element_displayed():
-            self.errors.append("Message with test '%s' was not received" % message_from_device_1)
-        if not device_1_chat.chat_element_by_text(message_from_device_2).is_element_displayed():
-            self.errors.append("Message with test '%s' was not received" % message_from_device_2)
+        visible_messages = list(reversed(messages))
+        for i in range(len(device_chats)):
```
antdanchenko: `for chat in device_chat` so you can call dict by chat key which should be created in https://github.com/status-im/status-react/pull/7710/files#r265949069

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -27,10 +38,18 @@ def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
...
     # device 2: open group chat
     device_2_chat = device_2_home.get_chat_with_user(chat_name).click()
-    device_2_chat.join_chat_button.click()
 
     return device_1_chat, device_2_chat
 
+def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
+    device_1_chat, device_2_chat = create_new_group_chat(device_1_home, device_2_home, chat_name)
+    device_2_chat.join_chat_button.click()
+    return [device_1_chat, device_2_chat]
```
antdanchenko: instead of list, you can use tuple `device_1_chat, device_2_chat`

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -47,28 +66,26 @@ def test_create_new_group_chat(self):
...
         self.create_drivers(2)
 
         device_1_home, device_2_home = create_users(self.drivers[0], self.drivers[1])
         chat_name = device_1_home.get_public_chat_name()
-        device_1_chat, device_2_chat = create_and_join_group_chat(device_1_home, device_2_home, chat_name)
+        device_chats = create_and_join_group_chat(device_1_home, device_2_home, chat_name)
 
-        device_1_chat.send_message(message_from_device_1)
-        device_2_chat.send_message(message_from_device_2)
+        for i in range(len(device_chats)):
```
antdanchenko: `for chat in device chats:` or `for i, chat in enumerate(device_chats)`

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -47,28 +66,26 @@ def test_create_new_group_chat(self):
...
                 self.errors.append('Oops! Chat screen does not match the entered chat name %s' % chat_name)
-
         self.verify_no_errors()
 
     @marks.testrail_id(3993)
     @marks.critical
     def test_send_message_in_group_chat(self):
-        message_from_device_1 = 'Hello from device 1'
-        message_from_device_2 = 'Hi there! Sent from device 2'
+        messages = ["Hello from device 1", "Hi there! Sent from device 2"]
```
antdanchenko: you can map messages to chats in dict() instead of creating this list, chat as key and message as value

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -27,10 +38,18 @@ def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
...
-    device_2_chat.join_chat_button.click()
 
     return device_1_chat, device_2_chat
 
+def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
+    device_1_chat, device_2_chat = create_new_group_chat(device_1_home, device_2_home, chat_name)
+    device_2_chat.join_chat_button.click()
+    return [device_1_chat, device_2_chat]
+
+def check_group_chat_is_deleted(self, device_home, chat_name):
```
antdanchenko: please do verification directly in test instead of separate verification method

test/appium/tests/atomic/chats/test_group_chat.py
antdanchenko: as result we can use method in test like:
device_1_chat, device_2_chat = create_and_join_group_chat()

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -47,28 +66,26 @@ def test_create_new_group_chat(self):
...
+            device_chats[i].send_message(messages[i])
 
-        if not device_2_chat.chat_element_by_text(message_from_device_1).is_element_displayed():
-            self.errors.append("Message with test '%s' was not received" % message_from_device_1)
-        if not device_1_chat.chat_element_by_text(message_from_device_2).is_element_displayed():
-            self.errors.append("Message with test '%s' was not received" % message_from_device_2)
+        visible_messages = list(reversed(messages))
+        for i in range(len(device_chats)):
+            if not device_chats[i].chat_element_by_text(visible_messages[i]).is_element_displayed():
+                self.errors.append("Message with test '%s' was not received" % visible_messages[i])
```
antdanchenko: typo `test`

test/appium/tests/atomic/chats/test_group_chat.py
```diff
@@ -27,10 +38,18 @@ def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
...
     return device_1_chat, device_2_chat
 
+def create_and_join_group_chat(device_1_home, device_2_home, chat_name):
+    device_1_chat, device_2_chat = create_new_group_chat(device_1_home, device_2_home, chat_name)
+    device_2_chat.join_chat_button.click()
+    return [device_1_chat, device_2_chat]
+
+def check_group_chat_is_deleted(self, device_home, chat_name):
+    if device_home.element_by_text(chat_name).is_element_displayed():
+        self.errors.append("Public chat '%s' is shown, but the chat has been deleted" % chat_name)
```
antdanchenko: group chat instead of public

antdanchenko (approved):

Serhy (approved):

churik:
Will make another PR with required changes

PR #7721, opened 2019-03-12
Don't specify `to` when making the request to mailserver.

cammellos:
In some cases (say the clock of the device has drifted) specifying a `to` parameter
is decrimental and might result in missing messages.

We have changed the mailservers to default `to` if not present to `now`,
which should gives us better guarantees.

We also removed the 24 hours limit as now all the requests will be
paginated according to the `limit` parameter.

### Testing

Currently it only works on `staging` and `test` fleet

1) Create a new account
2) Connect to `staging` or `test` fleet
3) Set your clock to -1 hour
4) Join an empty public chat
5) Logout
6) From a different device send a few messages
7) Login and Fetch 24 hours on that chat 

On develop those offline messages won't be fetched, while they should be fetched in this build.

Please do no merge before we update `beta` fleet, I will only do that once tested successfully.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -283,18 +283,14 @@
 (defn request-messages! [web3 {:keys [sym-key-id address]} {:keys [topics cursor to from] :as request}]
   ;; Add some room to from, unless we break day boundaries so that messages that have
   ;; been received after the last request are also fetched
-  (let [adjusted-from (adjust-request-for-transit-time from)
+  (let [actual-from   (adjust-request-for-transit-time from)
         actual-limit  (or (:limit request)
-                          @limit)
-        actual-from   (if (> (- to adjusted-from) one-day)
-                        from
-                        adjusted-from)]
+                          @limit)]
     (log/info "mailserver: request-messages for: "
               " topics " topics
               " from " actual-from
               " cursor " cursor
-              " limit " actual-limit
-              " to   " to)
+              " limit " actual-limit)
     (.requestMessages (transport.utils/shh web3)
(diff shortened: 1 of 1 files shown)


mandrigin (approved):

rasom (approved):

cammellos:
just rebased against develop, might help with tests

asemiankevich:
@cammellos would you mind to rebase please?

cammellos:
@asemiankevich , done !

asemiankevich:
so the descirbed scenario works , i can see messages on device which is 1h behind. 
It results in error on beta fleet but seems to be expected, right @cammellos ? 

I want to do more testing here though

cammellos:
@asemiankevich correct, beta fleet has not been updated, so it will throw an error.

asemiankevich:
@cammellos https://ethstatus.testrail.net/index.php?/tests/view/538902 will fail because of 
image

however messaging on test / staging fleet looks good to me. Is there anything else i need to check in this PR ? Above mentioned scenario is also working okay

cammellos:
could you smoke test release app with staging fleet? just to make sure it s
all ok, and we can then upgrade the fleet. Thanks!

asemiankevich:
@cammellos i briefly checked release app, i can chat and transact on staging , looks good to me

cammellos:
@asemiankevich I have deployed the mailservers, could you re-run auto tests please?
Thanks!

asemiankevich:
> @asemiankevich I have deployed the mailservers, could you re-run auto tests please?
> Thanks!

yes sure

PR #7728, opened 2019-03-13
bug #7724 - correctly showing password errors for sign message

### LINKED ISSUE #7724: No "wrong password" error if enter wrong password on Sign Message screen
### Description

*Type*: Bug

*Summary*: On Sign message screen there is no error and nothing happens if user entered wrong password and tap Sign. Would expect error message same as when Signing transaction and entering wrong password.

Note that on 0.10.1 release issue does not happen and expected error message is shown

### Reproduction

- Open Status and switch to Ropsten 
- Open Status Test Dapp
- On Transactions tap on Sign message button
- Enter wrong password and  tao on Sign. Check if there is any error message "Wrong password" shown 

### Additional Information
* Status version: nightly March 12. 

#### Logs
see March 12, 16:22
[Status.log](https://github.com/status-im/status-react/files/2957574/Status.log)
[geth.log](https://github.com/status-im/status-react/files/2957575/geth.log)

goranjovic:
fixes #7724

### Summary:

Fixed bug with missing error popup for wrong password

### Steps to test:
see #7724

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/sign_message/views.cljs
+++ b/src/status_im/ui/screens/wallet/sign_message/views.cljs
@@ -56,7 +56,7 @@
             opacity-value   (animation/create-value 0)]
     {:component-did-mount #(send.animations/animate-sign-panel opacity-value bottom-value)}
     [react/animated-view {:style (styles/animated-sign-panel bottom-value)}
-     (when (:wrong-password? value-atom)
+     (when (:wrong-password? @value-atom)
        [tooltip/tooltip (i18n/label :t/wrong-password) styles/password-error-tooltip])
      [react/animated-view {:style (styles/sign-panel opacity-value)}
       [react/view styles/spinner-container
@@ -104,7 +104,9 @@
           nil]]]]
       [enter-password-buttons value-atom false
        #(re-frame/dispatch [:wallet/discard-transaction-navigate-back])
-       #(re-frame/dispatch [:wallet/sign-message typed? (merge screen-params @value-atom)])
+       #(re-frame/dispatch [:wallet/sign-message typed? (merge screen-params @value-atom)
+                            (fn []
+                              (swap! value-atom assoc :wrong-password? true))])
        :t/transactions-sign]
       [password-input-panel value-atom :t/signing-message-phrase-description false]
       (when in-progress?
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/wallet/sign_message/views.cljs
```diff
@@ -103,7 +103,7 @@
           nil]]]]
       [enter-password-buttons value-atom false
        #(re-frame/dispatch [:wallet/discard-transaction-navigate-back])
-       #(re-frame/dispatch [:wallet/sign-message typed? (merge screen-params @value-atom)])
+       #(re-frame/dispatch [:wallet/sign-message typed? (merge screen-params @value-atom) value-atom])
```
cammellos: Interesting, I am all for local state when actually local and well encapsulated, but in this case looks like it's not local anymore, and relies on a state modification happening in a separate component/file (state is being passed around, but modified directly i.e `swap`), at this point I would just consider using the re-frame database, as there's no encapsulation effectively anymore.

another option is might be  a matter of refactoring (`views.cljs` never access state directly, but always access it through `events.cljs`, i.e `(events/wrong-password?)` `(events/sign-message! typed? screen-params)` so that `value-atom` never leaves `events.cljs` directly, but always through an API.

Not much of a big deal to be honest, but it's good to agree on a direction, what do you think?

goranjovic: I agree with your point, although I'd rather fix it in a different way - pass a callback from the view code e.g. `(show-wrong-password)`. That callback can alter the state and then it's local i.e. encapsulated.

Whether or not there's a password error really is UI state and nothing more. It doesn't belong in the events code and it doesn't belong in re-frame database.

cammellos: callback sounds like a good approach, best to keep the atom in the view I agree :+1:

goranjovic:
@cammellos @dmitryn guys, please re-review - I replaced the direct atom access with a callback.

cammellos (approved):

flexsurfer (approved):

asemiankevich:
image

Issue #7579, opened 2019-02-25
App in "Offline" state is launched if open app via Push notification (with saved password)


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: Offline state is shown for long time (10 minutes at least) if open app by tapping on push notification. password should be saved, so 1:1 chat is opened. Found on Android, can't test on iOS because no PNs are received on iOS when app is closed atm https://github.com/status-im/status-react/issues/7438

#### Expected behavior
[comment]: # (Describe what you expected to happen.)
1:1 chat is opened, app connects to mailserver and massage that triggered PN is shown
#### Actual behavior
[comment]: # (Describe what actually happened.)
1:1 chat is opened with Offline status, no new messages shown. Log contains
```
02-25 17:01:56.108 24541 24577 D StatusModule: ConnectionChange: wifi, is expensive false
02-25 17:01:56.113 24541 24576 E ReactNativeJS: ERROR [status-im.mailserver.core:118] - mailserver: update-mailservers error there is no running node
02-25 17:01:56.120 24541 24576 D ReactNativeJS: DEBUG [status-im.utils.handlers:34] - Handling re-frame event:  :init-rest-of-chats
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: console.error: "ERROR [status-im.mailserver.core:118] - mailserver: update-mailservers error there is no running node", stack:
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: cljs$core$IFn$_invoke$arity$3@12:346554
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: cljs$core$IFn$_invoke$arity$2@12:346351
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: cljs$core$IFn$_invoke$arity$2@12:351623
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: c@12:7236495
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: r@12:7236677
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: <unknown>@12:2635905
02-25 17:01:56.121 24541 24577 E unknown:ReactNative: <unknown>@12:2655168
[...]

PR #7737, opened 2019-03-14
Wait until activity is ready

cammellos:
Fixes: #7579 

When resuming from a push notification sometimes `ReactInstanceManager.attachRootViewToInstance()` would be called after the db is initialized, resulting in the node not being started as `getCurrentActivity` would return `null` and we give up trying.

This PR changes the behavior so that we wait very inelegantly for it to return a value, and we give up after (roughly) 10 seconds.

### Testing

- Login/logout
- Create a new account
- Resuming from background
- Push notifications 

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -90,12 +90,25 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
     }
 
     private boolean checkAvailability() {
+      // We wait at least 10s for getCurrentActivity to return a value,
+      // otherwise we give up
+      for (int attempts = 0; attempts < 100; attempts++) {
         if (getCurrentActivity() != null) {
-            return true;
+          return true;
         }
+        try {
+          Thread.sleep(100);
+        } catch (InterruptedException ex) {
+          if (getCurrentActivity() != null) {
+            return true;
+          }
+          Log.d(TAG, "Activity doesn't exist");
+          return false;
+        }
+      }
 
-        Log.d(TAG, "Activity doesn't exist");
-        return false;
+      Log.d(TAG, "Activity doesn't exist");
+      return false;
 
     }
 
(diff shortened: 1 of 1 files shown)


modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -90,12 +90,23 @@ public void onHostDestroy() {
     }
 
     private boolean checkAvailability() {
+      for(int attempts = 0; attempts < 10; attempts++) {
```
mandrigin:
```suggestion
      for (int attempts = 0; attempts < 10; attempts++) {
```

mandrigin (approved):

rasom (approved):

mandrigin (approved):

annadanchenko:
@cammellos 
issue 1 Offline state if app is opened from second push notification (with saved password)

Steps:
1. Add someone to contact
2. log out, turn on "Save password until logout", log in
3. close Status app
4. get push notification from your contact who send text message in 1:1 chat with you. Note, username is shown as some hash (not real user name) it's expected
5. get second push notification from same contact. Note that now for some reason username in Push notification is shown as a name of the user (in my case it was Pixel)
6. Tap on the second push notification 
7. Check if you get history and if there is no Offline state. In my case - no history is retrieved, Offline is shown

Note: if use first Pn or get only 1 Pn and open the app - it works as expected. So issue seems to be with next pns

logcat: 
[new3.log](https://github.com/status-im/status-react/files/2971811/new3.log)
no geth.log was created 

PNs (second with real user name):

Screenshot_20190315-162933

Offline if open from second PN:
Screenshot_20190315-163523

annadanchenko:
@PombeirP not sure why it's moved back to TO Test by bot. It has issue that should be fixed, so I moved it away from IN TESTING

cammellos:
@annadanchenko ready to be tested again:

Easier repro step for the issue you had before:
1) Send push notification
2) Wait more than 10s 
3) Open app -> Offline is displayed

New build should not have that issue.

Could you please test logging in/out background/foreground app and starting stopping? (only on android, ios/desktop code is left untouched)

Thanks!

antdanchenko:
@cammellos please rebase the PR on develop

cammellos:
Done!

Issue #5287, opened 2018-07-25
Do not redirect to home tab when closing error dialog after scanning invalid QR contact code


asemiankevich:
### Description

*Type*: Feature

*Summary*: when i scan an invalid QR to start new chat - error dialog appears saying the content is not valid (expected). But when i click OK - dialog closes so I need to perform actions again to scan new code. I expect i stay on the same screen to have the possibility to scan any other QR again

#### Expected behavior

- [ ] do not redirect a user to Home tab after the error dialog for scanning invalid contact code is closed
- [ ] have the camera opened to the user can scan any other code when the error dialog is closed

#### Actual behavior
User is redirected to Home tab after he closes the error dialog

### Reproduction

- Open Status
- Tap plus button -> start new chat
- Tap scan button
- Scan invalid QR (that does not contain valid contact code)
- Pay attention `Unable to read this code` dialog appears
- Tap ok to close the dialog

ghost:
This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 40.0 DAI (40.0 USD @ $1.0/DAI)  attached to it as part of the Status-im fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/status-im/status-react/5287/2226).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/status-im/status-react/5287/2226).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $54,670.48 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #7745, opened 2019-03-15
After dismissing error, no redirect to home and new qr can be scanned.

krisc:
WIP PR for https://github.com/status-im/status-react/issues/5287

@yenda So this maybe too hacky, and perhaps there is a more elegant solution, but I merge a navigate-clean and the navigate-to right back to qr-scanner. So after dismissing the error message when scanning an invalid qr code, navigating back works and the user can also scan another qr code. Currently, however, the navigation-stack is cluttered and was attempting to reset it to what it was before merge.

Let me know if merging navigate-clean then navigate-to qr-scanner is a step in the wrong direction, and I'll figure out another solution.

### DIFF
--- a/src/status_im/qr_scanner/core.cljs
+++ b/src/status_im/qr_scanner/core.cljs
@@ -1,6 +1,7 @@
 (ns status-im.qr-scanner.core
   (:require [re-frame.core :as re-frame]
             [status-im.i18n :as i18n]
+            [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.utils :as utils]
             [status-im.utils.fx :as fx]))
 
@@ -18,10 +19,16 @@
                                               50))
                                            #(re-frame/dispatch [deny-handler qr-codes]))}})
 
+(fx/defn scan-qr-code-after-error-dismiss
+  [{:keys [db]}]
+  (let [view-id (:view-id db)]
+    {:db (assoc-in db [:navigation/screen-params view-id :barcode-read?] false)}))
+
 (fx/defn set-qr-code
   [{:keys [db]} context data]
   (merge {:db (-> db
                   (update :qr-codes dissoc context)
+                  (update-in [:navigation/screen-params :qr-scanner] assoc :barcode-read? true)
                   (dissoc :current-qr-context))}
          (when-let [qr-codes (:qr-codes db)]
            {:dispatch [(:handler qr-codes) context data (dissoc qr-codes :handler)]})))
--- a/src/status_im/ui/screens/add_new/models.cljs
+++ b/src/status_im/ui/screens/add_new/models.cljs
@@ -6,17 +6,11 @@
             [status-im.utils.universal-links.core :as universal-links]
             [status-im.utils.fx :as fx]))
 
-(fx/defn process-qr-code
+(fx/defn handle-qr-code
   [cofx data]
   (if (spec/valid? :global/public-key data)
     (universal-links/handle-view-profile cofx data)
     (or (universal-links/handle-url cofx data)
-        {:utils/show-popup {:title      (i18n/label :t/unable-to-read-this-code)
-                            :content    (i18n/label :t/use-valid-qr-code {:data data})
-                            :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
-
-(fx/defn handle-qr-code
-  [cofx data]
-  (fx/merge cofx
-            (navigation/navigate-to-clean :home {})
-            (process-qr-code data)))
+        {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
+                            :content (i18n/label :t/use-valid-qr-code {:data data})
+                            :on-dismiss #(re-frame/dispatch [:qr-scanner.ui/qr-code-error-dismissed])}})))
--- a/src/status_im/ui/screens/qr_scanner/views.cljs
+++ b/src/status_im/ui/screens/qr_scanner/views.cljs
@@ -24,26 +24,26 @@
   (re-frame/dispatch [:qr-scanner.callback/scan-qr-code-success identifier (camera/get-qr-code-data data)]))
 
 (defview qr-scanner []
-  (letsubs [{identifier :current-qr-context} [:get-screen-params]
-            camera-initialized? (reagent/atom false)
-            barcode-read? (reagent/atom false)]
-    [react/view styles/barcode-scanner-container
-     [qr-scanner-toolbar (or (:toolbar-title identifier) (i18n/label :t/scan-qr)) identifier]
-     [camera/camera {:onBarCodeRead #(if (:multiple? identifier)
-                                       (on-barcode-read identifier %)
-                                       (when-not @barcode-read?
-                                         (do (reset! barcode-read? true)
-                                             (on-barcode-read identifier %))))
-                     :ref           #(reset! camera-initialized? true)
-                     :captureAudio  false
-                     :style         styles/barcode-scanner}]
-     [react/view styles/rectangle-container
-      [react/view styles/rectangle
-       [react/image {:source {:uri :corner_left_top}
-                     :style  styles/corner-left-top}]
-       [react/image {:source {:uri :corner_right_top}
-                     :style  styles/corner-right-top}]
-       [react/image {:source {:uri :corner_right_bottom}
-                     :style  styles/corner-right-bottom}]
-       [react/image {:source {:uri :corner_left_bottom}
-                     :style  styles/corner-left-bottom}]]]]))
+  (letsubs [{identifier :current-qr-context
+             barcode-read-sub? :barcode-read?} [:get-screen-params]
+            camera-initialized? (reagent/atom false)]
+    (let [barcode-read? barcode-read-sub?]
+      [react/view styles/barcode-scanner-container
+       [qr-scanner-toolbar (or (:toolbar-title identifier) (i18n/label :t/scan-qr)) identifier]
+       [camera/camera {:onBarCodeRead #(if (:multiple? identifier)
+                                         (on-barcode-read identifier %)
+                                         (when-not barcode-read?
+                                           (on-barcode-read identifier %)))
+                       :ref           #(reset! camera-initialized? true)
+                       :captureAudio  false
+                       :style         styles/barcode-scanner}]
+       [react/view styles/rectangle-container
+        [react/view styles/rectangle
+         [react/image {:source {:uri :corner_left_top}
+                       :style  styles/corner-left-top}]
+         [react/image {:source {:uri :corner_right_top}
+                       :style  styles/corner-right-top}]
+         [react/image {:source {:uri :corner_right_bottom}
+                       :style  styles/corner-right-bottom}]
+         [react/image {:source {:uri :corner_left_bottom}
+                       :style  styles/corner-left-bottom}]]]])))
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -271,7 +271,7 @@
     (if (some? validation-result)
       {:utils/show-popup {:title (i18n/label :t/unable-to-read-this-code)
                           :content validation-result
-                          :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}
+                          :on-dismiss #(re-frame/dispatch [:qr-scanner.ui/qr-code-error-dismissed])}}
       (fx/merge cofx
                 fx
                 (if config/partitioned-topic-enabled?
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -680,6 +680,11 @@
    (qr-scanner/scan-qr-code cofx identifier (merge {:handler handler} opts))))
 
 (handlers/register-handler-fx
+ :qr-scanner.ui/qr-code-error-dismissed
+ (fn [cofx [_ _]]
+   (qr-scanner/scan-qr-code-after-error-dismiss cofx)))
+
+(handlers/register-handler-fx
  :qr-scanner.callback/scan-qr-code-success
  (fn [cofx [_ context data]]
    (qr-scanner/set-qr-code cofx context data)))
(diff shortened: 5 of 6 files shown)


krisc:
@yenda  Actually, something like this seems more straightforward, but I may be using merge incorrectly as this isn't working as I expect:

```
(fx/defn scan-qr-code-after-error-dismiss
  [{:keys [db] :as cofx}]
  (let [navigation-stack (vec (:navigation-stack db))]
     (fx/merge cofx     		   
               (navigation/navigate-to-clean :qr-scanner {})
               {:db (assoc db :navigation-stack navigation-stack)})))
```

src/status_im/ui/screens/add_new/models.cljs
```diff
@@ -6,17 +6,11 @@
...
-                            :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
-
-(fx/defn handle-qr-code
-  [cofx data]
-  (fx/merge cofx
-            (navigation/navigate-to-clean :home {})
-            (process-qr-code data)))
+        {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
+                            :content (i18n/label :t/use-valid-qr-code {:data data})
+                            :on-dismiss #(re-frame/dispatch [:qr-scanner.ui/qr-code-error-dismissed])}})))
```
yenda, line 16: there shouldn't be any kind of navigation on dismiss

yenda (changes requested):
I have looked at the code and the problem comes from `status-im.ui.screens.qr-scanner.views`, you'll see that there is a `barcode-read?` atom that is set tot true once the barcode is read and no new barcode can be read after that. I think the reason is that without this atom there would be many event per second dispatched by the camera for the same qr code. The reason it works with your navigation is that this atom is then reset when reopening the view but that is not the correct way to proceed.

yenda:
I gave it a quick try here, there is probably a cleaner way to do it https://github.com/status-im/status-react/commit/409a83a4ca838b28018bdd2ff0bfd5173b8a213b

krisc:
@yenda Great! that works as expected. If I understand correctly, the change you made to `status-im.ui.screens.qr-scanner.views` was you made the `barcode-read?` atom retrievable as a screen-param, right? Because I didn't notice that as a screen-param in re-frisk before.

yenda:
@krisc yes I added it as a screen param so that it can be reverted back from an event.

krisc:
@yenda Is this PR ready for merging or does it still require cleanup? Also, is the new function `scan-qr-code-after-error-dismiss` I created in the qr-scanner module fine, like, naming-wise?

yenda (approved):

src/status_im/qr_scanner/core.cljs
```diff
@@ -1,6 +1,8 @@
 (ns status-im.qr-scanner.core
   (:require [re-frame.core :as re-frame]
             [status-im.i18n :as i18n]
+            [status-im.ui.screens.navigation :as navigation]
+            [taoensso.timbre :as log]
```
jeluard: Import not used?

src/status_im/ui/screens/qr_scanner/views.cljs
```diff
@@ -24,26 +24,27 @@
...
-      [react/view styles/rectangle
-       [react/image {:source {:uri :corner_left_top}
-                     :style  styles/corner-left-top}]
-       [react/image {:source {:uri :corner_right_top}
-                     :style  styles/corner-right-top}]
-       [react/image {:source {:uri :corner_right_bottom}
-                     :style  styles/corner-right-bottom}]
-       [react/image {:source {:uri :corner_left_bottom}
-                     :style  styles/corner-left-bottom}]]]]))
+  (letsubs [{identifier :current-qr-context
```
jeluard, line 27: What is this change about?

yenda: > I have looked at the code and the problem comes from `status-im.ui.screens.qr-scanner.views`, you'll see that there is a `barcode-read?` atom that is set tot true once the barcode is read and no new barcode can be read after that. I think the reason is that without this atom there would be many event per second dispatched by the camera for the same qr code. The reason it works with your navigation is that this atom is then reset when reopening the view but that is not the correct way to proceed.

src/status_im/qr_scanner/core.cljs
krisc: Oh yeah, I was using logs to debug. I'll remove this.

krisc:
@flexsurfer do you have a chance to review these changes?

env/dev/env/config.cljs
```diff
@@ -1,6 +1,6 @@
 (ns env.config)
 
 (def figwheel-urls {:android "ws://localhost:3449/figwheel-ws",
- :ios "ws://192.168.0.9:3449/figwheel-ws",
+ :ios "ws://192.168.1.84:3449/figwheel-ws",
```
flexsurfer: could you please revert this change from the commit?

flexsurfer (approved):

asemiankevich:
@krisc @yenda i noticed that the scanner throws an error fallback only 2 times in a row no matter how many times i scan the code. When i tap back -> i see the pop up then, please check the screen recording 

https://drive.google.com/open?id=1pJCV1J0LhFJhmf-ur1OBm_wDaI75GnBU

Steps:

1. open status
2. tap Plus button
3. tap Start new chat
4. tap Scan QR button
5. scan invalid QR 3 times in a row
6. notice that only 2 error pop-ups are shown. Third one is shown only if i tap back button and return on Chats screen

krisc:
> @krisc @yenda i noticed that the scanner throws an error fallback only 2 times in a row no matter how many times i scan the code. When i tap back -> i see the pop up then, please check the screen recording
> 
> https://drive.google.com/open?id=1pJCV1J0LhFJhmf-ur1OBm_wDaI75GnBU
> 
> Steps:
> 
> 1. open status
> 2. tap Plus button
> 3. tap Start new chat
> 4. tap Scan QR button
> 5. scan invalid QR 3 times in a row
> 6. notice that only 2 error pop-ups are shown. Third one is shown only if i tap back button and return on Chats screen

I need permsission to view that file ... but in any case I followed those steps and the same happens for me too.

@yenda upon scanning a qr-code, I don't see the event `scan-qr-code-success` changing `barcode-read?` to true in re-frisk. But I do notice it changing to false with the event `scan-qr-code-after-error-dismiss`. Perhaps I don't fully understand the changes you made, but is this as expected?

krisc:
@yenda oh nvm, I do see using `adb log cat` that `barcode-read?` is indeed changing as expected. I did, however, notice that at first it's `nil` and then `scan-qr-code-after-error-dismiss` changes it to `false`. I tried initializing it to `false` and vice versa with `scan-qr-code-after-error-dismiss` changing it to `nil` but it doesn't seem to make a difference.

I'm currently at a loss, but I can look at this again tomorrow. If someone can explain to me how `letsubs` works, that would help me.

krisc:
@yenda I put in additional `log` messages to pinpoint where the `barcode-read?` is changed. Here is what `scan-qr-code-after-error-dismiss` looks like:
```
(fx/defn scan-qr-code-after-error-dismiss
  [{:keys [db]}]
  (let [view-id (:view-id db)]
  	(log/debug "barcode-read? = " (get-in db [:navigation/screen-params view-id :barcode-read?]))
  	(log/debug "setting barcode-read? to false")
    {:db (assoc-in db [:navigation/screen-params view-id :barcode-read?] false)}))
```
Upon scanning an invalid qr-code and the dismissing the error for the first time, the first log statement says:
```
barcode-read? =  nil
```
Then upon scanning an invalid qr-code and dismissing the error a second time, the first log statemtn says:
```
barcode-read? =  false
```
I would expect on both times, that it should read `true`. Because in `defview qr-scanner` the atom `barcode-read?` indeed changes to `true`. So my guess is that the atom in `defview qr-scanner` doesn't actually reference the `screen-param` in the `db`.

So if my guess is correct, does this mean that the resetting of the atom `defview qr-scanner` actually has to make a change to `db` directly?

annadanchenko:
@yenda does it make sense to continue testing the build while there are unanswered questions?

yenda:
@annadanchenko no

yenda:
@krisc sorry for not reaching out earlier

In your log statement you are not checking for the value of `@barcode-read?` from the view but the value of the `:barcode-read` key in `@re-frame.db/app-db`.

In the view the symbol `barcode-read` refers to an atom, when you dereference that atom (`@` is shorthand for dereferencing), you get either the value of the key `:barcode-read?` in the subscription `:get-screen-params` (which is computed from `@re-frame.db/app-db`, with a simple `get-in`) or true if the value was reset (when you scan a qr code).

- you start the camera:  `@barcode-read?` is nil
- you scan a wrong QR code ` @barcode-read?` is briefly reset! to true (note that this is what you don't see in the logs, because it is not the key in app-db that is changed but the value referenced by the atom), in the events that follow, the value of  `:barcode-read?` is set to `false`, which triggers a re-rendering and  `@barcode-read?` becomes false

what happens:
- you scan a wrong QR code again, ` @barcode-read?` is briefly set to true, in the events that follow, the value of  `:barcode-read?` is already set to `false`, which triggers doesn't triggers a re-rendering and  `@barcode-read?` stays true, the scanner isn't dispatching an event anymore when scanning is successful.

possible solution:
- make sure the barcode-read? value coming from the subscription is updated everytime it is needed
- check if the reset! of barcode-read? in the view itself is necessary

yenda:
@krisc if you are still planning to fix this that's awesome, I'll try to be more reactive

krisc:
@yenda Thanks, that is very helpful!

> @krisc if you are still planning to fix this that's awesome, I'll try to be more reactive

I'll take another look in the next couple days and see what I can do.

krisc:
> * check if the reset! of barcode-read? in the view itself is necessary

@yenda  I think I will try not using this atom in the view at all. I might try setting `:barcode-read?` to true on the `scan-qr-code-success` event. That way, I think it will trigger a re-render after `qr-code-error-dismissed` sets it to false. Let me know if I am mistaken.

yenda:
@krisc I would expect this as well, my only worry is that depending on how often onBarCodeRead is called, it my result in multiple events being dispatch for a single barcode, so please keep that it mind and verify that it doesn't happen

krisc:
@asemiankevich  @yenda  Okay, I think I got it to work as expected! The code still needs clean up though. I had the `barcode-read?` set back and forth to `true` and `false` as needed, but I still have the atom there. I'll remove it and do other clean up tomorrow. Let me know if I overlooked anything.

yenda:
@krisc does it work when you scan 3 wrong qr codes in a row?

krisc:
> @krisc does it work when you scan 3 wrong qr codes in a row?

@yenda  Yes, it does. But I've done minimal testing only. I'll clean up the code and do more manual testing today.

krisc:
@asemiankevich @yenda This should be it. Any number of invalid qr-codes can be scanned in a row.

It works without the use of an atom in views.

I also found a case where if a valid contact code is scanned and then the user navigates back to the qr-scanner, another code can't be scanned because `:barcode-read?` is still true. So I changed the `:screens/on-will-focus` event to set `:bardcode-read?` to false if focusing on qr-scanner.

krisc:
@Serhy have you any idea why this test is failing? https://github.com/status-im/status-react/pull/7745#issuecomment-480322384

churik:
@krisc that's because of faucet error (user didn't receive ETH on Ropsten), now all tests are passed.

Serhy:
Thanks @churik! Yes, that was a thing.
Issue #5287  itself no more reproduced! Confirmed with Android 6.0.1, Android 8.1, iOS 12.1.4
Camera screen stays when scanning wrong QRs until appropriate QR code scanned with `(+) -> Scan QR code` /  `(+) -> Start new chat -> Scan QR code` 

Thanks a lot, @krisc !

jeluard:
@krisc It appears this PR introduces this regression https://github.com/status-im/status-react/issues/8044

I noticed you removed an atom check that ensured no multiple events were fired for the same QR code. Can you provide more details about this point?

krisc:
> @krisc It appears this PR introduces this regression #8044
> 
> I noticed you removed an atom check that ensured no multiple events were fired for the same QR code. Can you provide more details about this point?

Does https://github.com/status-im/status-react/pull/8048 resolve this? But to answer your question, I removed the atom check because we are now using the `:barcode-read?` screen-param in `app-db`

jeluard:
#8048 is a revert of this PR so it solves all issues introduced by it.

The atom check is required AFAIK to make sure the same code is not read several times in a row. Your solution doesn't prevent that.

krisc:
Sorry, I've been busy with Ethereal Hackathon these past few weeks. I'll have more time to get back on this after the weekend.

krisc:
@jeluard so sorry for the delay, I haven't had much time to work on Status the last few weeks and when I did, I was wrestling with nix issues. But I put the atom check back into the code. What's the best way to deal with this problem? Submitting a brand new PR would be easiest for me. Shall I do that?

jeluard:
Not sure what has to be fixed here. @asemiankevich @Serhy can you comment on that?

Serhy:
@krisc new PR is perfectly fine!
Lets apply the fix at one place as originally was in https://github.com/status-im/status-react/issues/5287
`Home` -> `(+)` -> `Start new chat` -> `Scan QR` next to 'Enter contact code field'

So **reproduction**:
1) Open Status
2) Tap plus button -> `Start new chat`
3) Tap `Scan QR code` button
4) Scan invalid QR (that does not contain valid contact code) e.g. 
Screenshot 2019-05-21 at 12 28 38

5) Pay attention to 'Unable to read this code' pop-up 
6) Tap ok to close the dialog

**Actual behavior**
User is redirected to Home tab after he closes the error dialog
**Expected behavior**
User keep staying on camera opened screen. User can scan QR code again within this screen

krisc:
@Serhy @asemiankevich @jeluard  WIP PR will be ready soon! I just realized that `src/status_im/contact/core.cljs` has changed a lot since my original merged PR https://github.com/status-im/status-react/pull/7745 . Namely, `handle-qr-code` function is no longer there. Can someone fill me in on what happened here?

PR #7756, opened 2019-03-18
Disable Etherscan and Cryptocompare requests in chaos mode

rasom:
Chaos mode option is added in profile settings

Screen Shot 2019-03-19 at 09 42 08

status: ready

### DIFF
--- a/src/status_im/data_store/realm/schemas/base/account.cljs
+++ b/src/status_im/data_store/realm/schemas/base/account.cljs
@@ -238,4 +238,3 @@
 (def v21 (update v20 :properties merge
                  {:syncing-on-mobile-network? {:type :bool :default false}
                   :remember-syncing-choice?   {:type :bool :default false}}))
-
--- a/src/status_im/models/transactions.cljs
+++ b/src/status_im/models/transactions.cljs
@@ -12,7 +12,8 @@
             [taoensso.timbre :as log]
             [status-im.utils.fx :as fx]
             [re-frame.core :as re-frame]
-            [re-frame.db])
+            [re-frame.db]
+            [status-im.utils.config :as config])
   (:require-macros
    [cljs.core.async.macros :refer [go-loop go]]))
 
@@ -167,7 +168,7 @@
     (when (etherscan-supported? chain)
       (let [network-subdomain (when-let [subdomain (network->subdomain chain)]
                                 (str subdomain "."))]
-        (str "https://" network-subdomain  "etherscan.io/tx/" hash)))))
+        (str "https://" network-subdomain "etherscan.io/tx/" hash)))))
 
 (def etherscan-api-key "DMSI4UAAKUBVGCDMVP3H2STAMSAUV7BYFI")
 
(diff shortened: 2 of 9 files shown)


rasom:
@mandrigin @annadanchenko 
should we hide "Transaction history" on wallet screen or add a warning on top of transaction list?

.env
```diff
@@ -19,3 +19,4 @@ RN_BRIDGE_THRESHOLD_WARNINGS=0
 RPC_NETWORKS_ONLY=0
 STICKERS_ENABLED=1
 PARTITIONED_TOPIC=1
+CHAOS_MODE=1
```
mandrigin: I think it should be a dynamic option (in app-db, but we can default it from here)

mandrigin:
@rasom no, let it be as broken as it is. the idea is not to fix all the issues, but to find them.

.env
rasom: let's figure how we handle Etherscan disabling in this PR and I will file another one with a new profile setting which is stored in app-db

src/status_im/models/transactions.cljs
```diff
@@ -218,7 +219,7 @@
             result))))
 
 (defn- etherscan-transactions [chain account on-success on-error]
-  (if (etherscan-supported? chain)
+  (if (and (etherscan-supported? chain) (not config/chaos-mode?))
     (let [url (get-transaction-url chain account)]
       (log/debug "HTTP GET" url)
       (http/get url
```
mandrigin: I would go as far, as tweak this in `http/get` actually, to make it as similar to the real "etherescan is down" scenario as possible.

rasom: I just replaced urls, is it enough? :)

.env
mandrigin: sure!

rasom: well, added in this PR :)

src/status_im/data_store/realm/schemas/base/account.cljs
```diff
@@ -239,3 +239,6 @@
                  {:syncing-on-mobile-network? {:type :bool :default false}
                   :remember-syncing-choice?   {:type :bool :default false}}))
 
+(def v22 (update v21 :properties merge
```
cammellos: You could stick it into `settings` and it would not require a migration, but either places are fine with me

src/status_im/models/transactions.cljs
```diff
@@ -177,12 +178,17 @@
...
-         account "&startblock=0&endblock=99999999&sort=desc&apikey=" etherscan-api-key "&q=json")))
+(defn- get-transaction-url
+  ([chain account] (get-transaction-url chain account false))
+  ([chain account chaos-mode?]
+   {:pre  [(keyword? chain) (string? account)]
+    :post [(string? %)]}
+   (let [network-subdomain (get-api-network-subdomain chain)]
+     (str "https://" network-subdomain
+          (if chaos-mode?
+            "some-broken-url-af435c85ff48c2b9bdefd1295678fc24.io/api?module=account&action=txlist&address=0x"
```
cammellos: you could also use `http://httpstat.us/500` so it's more realistic (this will throw a dns error likely), but minor thing

src/status_im/utils/prices.cljs
```diff
@@ -10,14 +11,17 @@
 ;; Example usage:
 ;; (get-prices "ETH" "USD" println print)
 
-(def api-url "https://min-api.cryptocompare.com/data")
+(defn api-url [chaos-mode?]
+  (if chaos-mode?
+    "https://min-api.thisurldoesntexist5b60647d927da46.com/data"
```
cammellos: same in case we want it to be up but return a given status code `http://httpstat.us/500`

cammellos (approved):

src/status_im/data_store/realm/schemas/base/account.cljs
rasom: oh ok, I didn't know we have such field

src/status_im/models/transactions.cljs
rasom: ok

src/status_im/data_store/realm/schemas/base/account.cljs
rasom: fixed

src/status_im/models/transactions.cljs
rasom: fixed

mandrigin (approved):

goranjovic (approved):

annadanchenko:
@rasom after enabling chaos mode via UI I can see USD value of the ETH/SNT if request them in 1:1 chat. Only after app is relaunched the USD value in the chat is shown as 0 USD. Is it expected that after enabling Chaos mode app should be restarted?

Screenshot_20190319-164535

rasom:
If prices were already loaded before enabling this mode they are in `app-db`. So this change will have effect only after re-login (but also prices will not be updated during the session). That's the same situation as if cryptocompare/ethersan would stop working when the app is already started.

Issue #7779, opened 2019-03-20
Error "fx/defn expects a map of cofx as first argument got :url..." if scan QR code for bootnode


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug
[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: If scan qr code for bootnode from https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d then error is shown and bootnode address is not inserted. Error:
```
03-20 18:07:33.628 11632 11672 I ReactNativeJS: MESSAGE: fx/defn expects a map of cofx as first argument got :url in function function (e) {switch(arguments.length){case 2:return status_im.bootnodes.core.set_input.cljs$core$IFn$_invoke$arity$2(arguments[0],arguments[1]);case 3:return status_im.bootnodes.core.set_input.cljs$core$IFn$_invoke$arity$3(arguments[0],arguments[1],arguments[2]);default:throw Error(["Invalid arity: ",cljs.core.str.cljs$core$IFn$_invoke$arity$1(arguments.length)].join(""))}}
03-20 18:07:33.628 11632 11672 I ReactNativeJS: 
03-20 18:07:33.628 11632 11672 I ReactNativeJS: cljs$core$IFn$_invoke$arity$3@index.android.bundle:12:8282981
03-20 18:07:33.628 11632 11672 I ReactNativeJS: cljs$core$IFn$_invoke$arity$2@index.android.bundle:12:8295606
03-20 18:07:33.628 11632 11672 I ReactNativeJS: index.android.bundle:12:8406483
03-20 18:07:33.628 11632 11672 I ReactNativeJS: index.android.bundle:12:1794765
03-20 18:07:33.628 11632 11672 I ReactNativeJS: invoke_interceptor_fn@index.android.bundle:12:1725002
03-20 18:07:33.628 11632 11672 I ReactNativeJS: invoke_interceptors@index.android.bundle:12:1725303
03-20 18:07:33.628 11632 11672 I ReactNativeJS: execute@index.android.bundle:12:1727752
03-20 18:07:33.628 11632 11672 I ReactNativeJS: handle@index.android.bundle:12:1736165
[...]

PR #7782, opened 2019-03-21
Fix qr code and bootnode

cammellos:
Fixes: #7779

Just ordering of parameters.

status: ready

### DIFF
--- a/src/status_im/bootnodes/core.cljs
+++ b/src/status_im/bootnodes/core.cljs
@@ -3,6 +3,7 @@
             [re-frame.core :as re-frame]
             [status-im.accounts.update.core :as accounts.update]
             [status-im.i18n :as i18n]
+
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.fx :as fx]))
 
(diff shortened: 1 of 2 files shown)


rasom (approved):

jeluard (approved):

churik:
@cammellos there is no error when scanning anymore, however, "Save" button is inactive when scanning QR from https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d.
I found that it is because of two whitespaces are placed after the bootnode address.
Should it be fixed in QR code on https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d or we should trim the input, WDYT?

cammellos:
@churik I will strip trailing whitespaces when scanning from qr, but for now only when from QR, is that ok?

churik:
@cammellos I believe it is enough for now, agree

cammellos:
@churik should be fixed now

churik:
Can save bootnode when scan valid QR, paste valid bootnode from https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d when app is offline / online and the bootnodes are disabled / enabled.
Tested on IOS 11.4.1 and Android 8.0

PR #7777, opened 2019-03-20
upgrade ruby to 2.5

jakubgs:
This might fix these errors:
```
 bundler: failed to load command: fastlane (/var/tmp/bundle-1/bin/fastlane)
 MultiJson::AdapterError: [!] Did not recognize your adapter specification (libruby.so.2.5: cannot open shared object file: No such file or directory - /var/tmp/bundle-1/gems/json-2.2.0/lib/json/ext/parser.so).
   /var/tmp/bundle-1/gems/json-2.2.0/lib/json/ext.rb:7:in `require'
   /var/tmp/bundle-1/gems/json-2.2.0/lib/json/ext.rb:7:in `<module:Ext>'
   /var/tmp/bundle-1/gems/json-2.2.0/lib/json/ext.rb:6:in `<module:JSON>'
   /var/tmp/bundle-1/gems/json-2.2.0/lib/json/ext.rb:3:in `<top (required)>'
...
```
Which show up in Android builds when running in `release` mode when Fastlane is used for uploading to PlayStore.

PR #7506, opened 2019-02-14
Replace nvm, apt, Homebrew, external Qt install and ad-hoc downloads with Nix package manager

pedropombeiro:
[comment]: # (Please replace ... with your information. Remove < and >)
[comment]: # (To auto-close issue on merge, please insert the related issue number after # i.e fixes #566)

### Summary:

[comment]: # (Summarise the problem and how the pull request solves it)

This PR introduces the Nix package manager and in the process:
- removes dependencies on nvm, Homebrew and apt-get;
- removes the Dockerfiles in `/docker` and `scripts/3rd-party/nvm/0.33.11/install.sh`;
- removes the requirement to have Qt/macdeployqt pre-installed and uses its own build;
- removes the need to install Android NDK and uses its own build;
- removes yarn 1.13.0 local install from the repo;
- moves from downloading linuxdeployqt (Linux) and NSIS (Windows) to using its own build in Nix.

Sorry for the size of the PR, but it is a big undertaking at the same time. I've tried to fixup commits as much as possible to keep them as focused as possible and get rid of noise/failed attempts.

Documentation updates are done in https://github.com/status-im/status.im/pull/244.

The Nix scripts are structured as follows:

```
- /default.nix
  - /scripts/lib/setup/nix
    - desktop/default.nix
      - appimagetool/default.nix
      - linuxdeployqt/default.nix
      - windows/default.nix
        - conan/default.nix
        - nsis/default.nix
    - global-node-packages/generate-nix.sh
      - output/default.nix
    - mobile/default.nix
      - android-ndk/default.nix
```

### Done

- [x] Replace existing tools with Nix equivalents (Conan, Node, NVM, yarn, Leiningen, OpenJDK, Maven, Python, etc.);
- [x] Install react-native using Nix instead of npm;
- [x] Integrate Qt5 Nix package (Linux and macOS);
- [x] Fix up desktop app libraries on both Linux and macOS;
- [x] Remove yarn 1.13.0 release from repo;
- [x] Replace NSIS apt install with custom Nix recipe;
- [x] Replace Android NDK, `appimagetool` and `linuxdeployqt` downloads with custom Nix recipes;
- [x] Move deployment helper tools used for generating Linux Desktop app to Nix recipes;
- [x] Split `/default.nix` to task-based Nix expressions. E.g. we only need to install Conan and NSIS when running `make release-windows-desktop`;

### TODO list (as of 28.02.2019)

- Blockers:
[...]

PR #7784, opened 2019-03-21
re-add custom cache folders in /var/tmp

jakubgs:
Removed in #7777, but apparently this is still an issue for Gradle.

I'm also putting `bundler` cache path in `/tmp/bundle` because otherwise it fails due to clashing ruby versions:
```
jenkins@de7ef417c83c:mobile-android$ export BUNDLE_PATH=/var/tmp/bundle-0
jenkins@de7ef417c83c:mobile-android$ bundle exec fastlane android nightly
/usr/bin/env: ‘ruby2.3’: No such file or directory
```
My best guess is that this was caused by the `nix` branch(#7506) which is already using Ruby `2.5`, while `develop` is still using `2.3`.

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -44,6 +44,12 @@ pipeline {
     ANDROID_SDK_ROOT  = '/usr/lib/android-sdk'
     ANDROID_NDK       = '/usr/lib/android-ndk'
     ANDROID_NDK_HOME  = '/usr/lib/android-ndk'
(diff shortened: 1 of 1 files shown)


mandrigin (approved):

jakubgs:
Successful manual release job for Android:
https://ci.status.im/job/status-react/job/combined/job/mobile-android/10576/

rasom (approved):

PR #7803, opened 2019-03-21
remove white line on android wallet screen

yenda:
Since the PR to remove overdraw was merged, the lack of blue background in the wallet screen introduced a white line between the asset total and wallet actions.
This PR covers it up on Android

Tested on android emulator

Screenshot_1553185982

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/main/views.cljs
+++ b/src/status_im/ui/screens/wallet/main/views.cljs
@@ -1,28 +1,28 @@
-  (ns status-im.ui.screens.wallet.main.views
-    (:require-macros [status-im.utils.views :as views])
-    (:require [reagent.core :as reagent]
-              [re-frame.core :as re-frame]
-              [status-im.i18n :as i18n]
-              [status-im.ui.components.colors :as colors]
-              [status-im.ui.components.list.views :as list]
-              [status-im.ui.components.react :as react]
-              [status-im.ui.components.toolbar.view :as toolbar]
-              [status-im.ui.components.icons.vector-icons :as vector-icons]
-              [status-im.ui.screens.wallet.onboarding.views :as onboarding.views]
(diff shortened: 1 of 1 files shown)


yenda:
If the other PR is already in release branch this fix should be included in the release as well

vitvly (approved):

rasom (approved):

rasom:
What exactly is fixed? That's nightly

photo_2019-03-21_21-38-38

asemiankevich:
@yenda could you please be more specific what is done here ? agree with @rasom , dont see any difference at first sight

yenda:
@rasom @asemiankevich weird. That's the emulator on develop without this fix for me:
Screenshot_1553199112

asemiankevich:
@yenda this is develop on iPhone7 simulator

image

errorists:
so just to chime in, it looks like a half pixel, this kind of stuff usually happens on scaling when an object can't be positioned exactly on a pixel grid, so say the 'Send' button has a Y coordinate of 320.5. Does this happen on a real device, cause it might only be the case of scaling the simulator window?

annadanchenko:
I can confirm that it happens on one of my real devices (Pixel XL). Other device doesn't have it (Galaxy S6)

errorists:
Ok so it makes sense now, the gap is created only on some [screen densities](https://material.io/design/layout/density-resolution.html#pixel-density), so yeah I'd +1 fixing that 👍

errorists (approved):

annadanchenko:
@yenda what's wrong with security checks on this PR? I see "Failed to read manifest file" at least on several of them
Screen Shot 2019-03-22 at 10 12 14

yenda:
@annadanchenko I think it is because it's not rebased on latest develop, doing it now

PR #7804, opened 2019-03-21
Add jenkins docker image

### LINKED ISSUE #7793: Share Android SDK as a Docker volume in Android CI builds
# Problem

Currently the Android SDK is being downloaded on every single build.

Carry over from https://github.com/status-im/status-react/pull/7506

pedropombeiro:

fixes #7793

### Summary:

This PR improves the performance of CI builds by including the Android SDK in the Docker image. The Dockerfile directory was moved from infra-utils repo. It also creates 3 Nix packages for the macOS, Windows, and Linux base images so that `build-desktop.sh` doesn't have to download them every time (both on CI and dev machine).

status: ready


ci/nix/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+ADD nix-install-${NIX_VERSION}.sh /tmp/nix-install.sh
+
+RUN groupadd --gid ${JENKINS_GID} jenkins \
+ && useradd --shell /bin/bash --create-home --password "" \
+      --comment "jenkins" --groups root \
+      --uid ${JENKINS_UID} --gid ${JENKINS_GID} jenkins
+
+# this special wrapper script runs sh inside of nix-shell
+#ADD bash.sh /bin/bash
```
jakubgs: This is not being used, we can drop it.

ci/nix/Makefile
```diff
@@ -0,0 +1,53 @@
...
+JENKINS_UID = 1001
+JENKINS_GID = 1001
+
+NIX_VERSION = 2.2.1
+NIX_INSTALL = nix-install-$(NIX_VERSION).sh
+NIX_SHA = 0039df605d2324f721f0c99ab76fc58fa60f791d84b30196eed805e8e65b1e51
+NIX_URL = https://nixos.org/releases/nix/nix-$(NIX_VERSION)/install
+
+# WARNING: Remember to change the tag when updating the image
+DEPS_HASH = $(shell $(GIT_ROOT)/scripts/gen-deps-hash.sh -b $(NIX_SHA) \
```
jakubgs: This is missing `nix` version in the hash.

ci/nix/bash.sh
```diff
@@ -0,0 +1,2 @@
+#!/nix/var/nix/profiles/default/bin/nix-shell
+exec bash "$@"
```
jakubgs: We can drop this file.

ci/nix/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+ARG SDK_BUILD_TOOLS_VERSION
+
+# names of packages to install with sdkmanager
+ENV SDK_PACKAGES emulator patcher;v4 \
+                 platforms;android-26 \ 
+                 platforms;${SDK_PLATFORM_VERSION} \
+                 build-tools;${SDK_BUILD_TOOLS_VERSION} \
+                 extras;android;m2repository extras;google;m2repository
+
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
```
jakubgs: The `PATH` here was for the sake of Jenkins, if this is an intermediate image for preparing NDK and SDK this shouldn't be necessary.

ci/nix/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+ && rm -fr extras emulator ndk-bundle/toolchains/{aarch64-linux-android-4.9,x86-4.9,x86_64-4.9,llvm}
+
+################################################################################
+FROM ubuntu:18.04
+ENV LANG=en_US.UTF-8 \
+    LC_ALL=en_US.UTF-8 \
+    LANGUAGE=en_US.UTF-8
+
+ARG NIX_VERSION
+ARG DEPS_HASH
```
jakubgs: Do we do this just so it's part of image env for documentation sake? Because we're not using it for anything here.

jakubgs (commented):
All in all cool, but why are you putting it under `ci/nix` rather than `docker/nix`?

ci/nix/Makefile
pedropombeiro: good catch!

ci/nix/Dockerfile
pedropombeiro: I should add a comment. The goal is for certain dependencies which are not really directly used in the building of the image to still cause a new image to be generated (think StatusImWindowsBaseImage.zip which is now downloaded by Nix)

jakubgs: Ah, makes sense, clever. A commend would be good.

pedropombeiro:
> All in all cool, but why are you putting it under `ci/nix` rather than `docker/nix`?

Because this is only used in CI? I can move it to /docker/nix if you feel that's best.

jakubgs:
Nah, it's fine I guess. I just thought it's not that descriptive. Maybe `ci/docker` would make more sense?

jakubgs (approved):
good_work

yenda (approved):

pedropombeiro:
Merging without waiting for e2e tests since this doesn't impact app functionality (it's only CI infra stuff)

PR #7805, opened 2019-03-22
introduce system-tags

### LINKED ISSUE #7673: User automatically added to contact list after blocking / unblocking
### Description

*Type*: Bug
*Summary*: if you block and unblock a user, it will be added automatically to your contact list on app, but technically user is not your contact, because he can't see your custom name. So after blocking and unlocking user, you can't add it to contacts normally to share your profile data.

#### Expected behavior
user B  is not in your contact list; can see "Add to contact" in 1-1 with user B

#### Actual behavior
user B is in your contact list, but in fact he is not (he can't see your custom name) 

### Reproduction

- User A: create 1-1 chat with User B, don't add him to your contacts
- User A: exchange several  messages with user B
- User A: block user B
- User A: unblock User B
- User B: send several message to user A
- User A: check contact list and 1-1 chat with user B

### Additional Information
* Status version: release 0.10.0
* Operating System: Android, IOS

yenda:
fixes synchronization of blocked users
fix #7673 User automatically added to contact list after blocking / unblocking

required for tribute to talk

### Summary:

`system-tags` is a set of namespaced keywords that allows us to define properties of the contact. In this PR we introduce `:contact/added` and `:contact/blocked`. In the future additional keywords for tribute to talk and contact group and permissions management will be added.

- remove unused status and description property from contact
- replace pending? and blocked? property from contact by system-tags

#### Platforms
- Android
- iOS
- macOS
- Linux
- Windows

#### Areas that are impacted
- 1-1 chats
- account recovery
- pairing/contact synchronization

status: ready


src/status_im/ui/screens/desktop/main/chat/views.cljs
```diff
@@ -375,12 +375,12 @@
...
                   maybe-contact [:contacts/current-contact]]
     (let [contact (or maybe-contact (contact.db/public-key->new-contact identity))
-          {:keys [pending? public-key]} contact]
+          {:keys [added? public-key]} contact]
       [react/view {:style styles/chat-profile-body}
        [profile.views/profile-badge contact]
        ;; for private chat, public key will be chat-id
        [react/view
-        (if (or (nil? pending?) pending?)
+        (if added?
```
cammellos, line 384: maybe I got it the other way around, but should this not be the opposite?
if `added?` we don't want to show the `add-to-contacts` bar?

cammellos (approved):
much better! 
I would add a description of the possible tags (and states), as I understand:

I have added the user to my contacts  -> `:contact/added`
I have blocked the user -> `:contact/blocked`
The user has added my but I have not added them -> ? (we don't have a system tag for this, we check the presence only, might be worth adding on an update received)

src/status_im/ui/screens/desktop/main/chat/views.cljs
yenda: indeed, I tested on android so I missed this as it is desktop

yenda:
@cammellos there is a description of the tags in contact.db
I also had `:contact/request-received` but I realized it is not used at all unlike `:contact/added` which is the previsous `:pending?` to `false` (but not `nil`) so I removed it (see second commit).

cammellos:
thanks @yenda , what do you think about re-introducing it? it is true that is not used currently, but otherwise we might loose information (say block->unblock , has this user added us or it go in through blocking?, although there are other ways to check `last-updated`)

Either ways I am ok though, but I have a slight preference to add it (if it does not require much extra work)

yenda:
@cammellos I don't mind adding it especially considering it has no use at that point it would only be conj it to system-tags on contact updates. But it has no use so I don't see why we should do it. 

In the case of blocking then unblocking there is still the extra information the user shared in his contact-request. In any case you don't see the contact anywhere until he contacts you again since the chat was deleted when blocking, unless you added it in which case he's in your contact list.

cammellos:
@yenda I do have a bit of secret agenda :) 
https://github.com/status-im/status-react/pull/7702/files#diff-a298d80ef69d5e8c89e03d1232a56dd4R121

I would use it if I had it there.

I will give priority to this PR in any case as it's more important to get this in and rebase once that's done, but if it's not in this PR I will probably add it there, and would save me a migration, but don't mind adding it there though, up to you.

yenda:
@cammellos ok then I guess it would make sense, so is
`(true? pending?) (conj ":contact/request-received")` enough for migrations?

vitvly (approved):

cammellos:
yes, looks good, thanks @yenda !

src/status_im/contact/core.cljs
```diff
@@ -31,12 +31,6 @@
...
 
-(defn can-add-to-contacts? [{:keys [pending? dapp?]}]
-  (and (not dapp?)
-       (or pending?
-           ;; it's not in the contact list at all
-           (nil? pending?))))
-
 (defn build-contact [{{:keys [chats] :account/keys [account]
                        :contacts/keys [contacts]} :db} public-key]
   (cond-> (assoc (contact.db/public-key->contact contacts public-key)
```
yenda: we can probably remove the address thingy as it is computed already in public-key->contact function

churik:
@yenda 
#### 1) After first sync contacts are not synced (only blocked users are synced)

**Steps:**
- on device 1: create the account, add one contact and block one user
- on device 2: restore same account
- pair devices
- on device 1: tap on "Sync all devices"

**Expected result:** 1 contact and 1 blocked user are synced
**Actual result:** only 1 blocked user is synced
Log from device 2:
[geth (1).log](https://github.com/status-im/status-react/files/2996992/geth.1.log)

#### 2) After initial sync if you block user, it doesn't affect to other devices (so there is no sync of blocked users)
**Steps:**
- on device 1: create the account, add one contact and block one user
- on device 2: restore same account
- pair devices
- on device 1: tap on "Sync all devices"
- on device 1:  block another user (i.e. User C)

**Expected result:**  `User C` is blocked on device 2
**Actual result:** no update in `Blocked user list` on device 2

Checked on https://github.com/status-im/status-react/commit/0064d7e1e4a94225c33dbb9421c788bd2527d399 build

yenda:
@churik 1 and 2 should be fixed. However for now it does sync blocked/unblocked but doesn't act on it during the session and doesn't delete the chat. Could you create an issue for that, it will require more refactoring of the sync code and I'd prefer to merge this PR to move ttt forward

churik:
@yenda 
1. is not fixed. Tried on build with https://github.com/status-im/status-react/commit/4c22534aacb89137d8b0195562b613c14f6e48ca commit, on both IOS and Android.

2. is fixed partially.
After I block user on device A > it will shown in device B in "Blocked users", but in fact I will continue to receive messages from this user on device B.
So it brings more confusion than it wasn't synced at all.

Ping me please if you need more info.

yenda:
@churik
1. I am not able to reproduce, are you sure you are adding as contact and not just opening a chat? I can see both blocked user and contact avec pressing `sync all devices`
2. so I would suggest we disable sync blocking user and enable it once syncing is refactored to actually block users ok?

churik:
@yenda, in this case, we needed to disable sync of blocked users at all, because of these who added after initial sync to "Blocked users" are not blocked, because you can receive messages from them.
Ping me when you disable it fully - and I'll recheck sync again, OK?

yenda:
@churik yes, what about 1?

churik:
@yenda I can reproduce 1 with same steps as I mentioned in https://github.com/status-im/status-react/pull/7805#issuecomment-475652891 - so I have 1 blocked user and one ordinary contact. After sync on device B I can see only blocked user, contact list is empty.
Probably disable of sync blocked users can help with this issue.

churik:
@yenda (still testing, so far):
Found issue in desktop:
**3) "Add to contact" is shown on public and group chat header**

Steps:
- create account
- join public chat
sdfsdfvcbcvb


If you tap on "Add to contact", then on + , error is shown
error
Tested on Mac OSx

yenda:
@churik should be fixed

churik:
Tested (IOS 11.4.1, Android 8, Mac OSx:
- https://github.com/status-im/status-react/issues/7673 is fixed
- restoring account with 1-1 accounts, adding to contacts
- sending messages in 1-1, chat, group, public chat
- add contact from public chat
- sync after pairing devices
- blocking user with or without adding him to contacts
- check sending contact request from release build
- PNs

churik:
@yenda hold on with merging please

PR #7807, opened 2019-03-22
fix message on empty search results

### LINKED ISSUE #7774: There is no text "Searching message history coming..." inside Messages section for search results
# Problem

For search results: It might look like search inside Messages does not work while we didn't implement it and should show  "Searching message history coming..." inside Messages section. This text was shown in 0.10.1 but is not shown in 0.11.0:

Screenshot_20190320-114419

Expected:

# Steps:
1. Open status and search on Chat screen

# Platfroms
Android and ios

yenda:
fix #7774 

status: ready

### DIFF
--- a/src/status_im/ui/screens/home/filter/views.cljs
+++ b/src/status_im/ui/screens/home/filter/views.cljs
@@ -117,23 +117,25 @@
     :key-fn                      first
     ;; true by default on iOS
     :stickySectionHeadersEnabled false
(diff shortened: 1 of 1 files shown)


vitvly (approved):

flexsurfer (approved):

Serhy:
Android 8, iOS, MacOS and confirmed #7774 is fixed.
Failed e2e test manually could not reproduce (it should be fixed already)

PR #7809, opened 2019-03-22
implement a custom retry for curl uploads

jakubgs:
Since the `curl` retry options don't seem to handle both `Operation timed out` and `Connection timed out` errors we have to write our own retry code.


mandrigin (approved):
looks good but make sure to actually fix the URL

rasom (approved):

pedropombeiro (approved):

PR #7820, opened 2019-03-22
[#7819] fix - avoid unnecessary effects when joining/syncing already active public chat

### LINKED ISSUE #7819: avoid series of effects involved in start-public-chat if re-adding already active chat
# Problem

If a user re-joins a public chat that she already joined, currently the process of series of effects involved in starting a fresh public chat is triggered unnecessarily. Instead the app should avoid it by simply navigating to the already active chat.

While, on the surface, this does not change anything for users in general, once PR #7454 is merged, not having this will cause visual inconsistency of showing loading intro header meaninglessly.

This should also make re-addition and navigation to of already active chat tad bit faster.

# Implementation

when adding/joining/re-adding public chat, check if the chat topic is already active and navigate to it instead of going through the whole creating the chat process.

# Notes

As a general optimization, this implementation approach is also good to make use of in other points in the app where new chat creation is triggered (e.g. from DApp).

# Future Steps

Implement the same approach to per DApp public chat, and any other areas that creates new chat.

bitsikka:
Fixes #7819 - which was created first when all repercussions were not fully known; Later #7952, #7974, #7992 were created as they were discovered.

status: ready

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -41,6 +41,12 @@
   ([cofx chat-id]
    (public-chat? (get-chat cofx chat-id))))
 
+(defn active-chat?
+  ([chat]
+   (:is-active chat))
+  ([cofx chat-id]
+   (active-chat? (get-chat cofx chat-id))))
+
 (defn group-chat?
   ([chat]
    (and (multi-user-chat? chat)
@@ -289,13 +295,16 @@
 (fx/defn start-public-chat
   "Starts a new public chat"
   [cofx topic {:keys [dont-navigate?] :as opts}]
-  (fx/merge cofx
-            (add-public-chat topic)
-            #(when-not dont-navigate?
-               (navigate-to-chat % topic opts))
-            (public-chat/join-public-chat topic)
-            (when platform/desktop?
-              (desktop.events/change-tab :home))))
+  (if (active-chat? cofx topic)
+    (when-not dont-navigate?
+      (navigate-to-chat cofx topic opts))
+    (fx/merge cofx
+              (add-public-chat topic)
+              #(when-not dont-navigate?
+                 (navigate-to-chat % topic opts))
+              (public-chat/join-public-chat topic)
+              #(when platform/desktop?
+                 (desktop.events/change-tab % :home)))))
 
 (fx/defn disable-chat-cooldown
   "Turns off chat cooldown (protection against message spamming)"
--- a/src/status_im/pairing/core.cljs
+++ b/src/status_im/pairing/core.cljs
@@ -155,7 +155,6 @@
   (let [contacts (:contacts/contacts db)
         contact-batches (partition-all contact-batch-n (vals contacts))]
     (concat (mapv contact-batch->sync-installation-message contact-batches)
-
             [(sync-installation-account-message cofx)]
             (chats->sync-installation-messages cofx))))
 
--- a/src/status_im/utils/universal_links/core.cljs
+++ b/src/status_im/utils/universal_links/core.cljs
@@ -5,6 +5,7 @@
             [re-frame.core :as re-frame]
             [status-im.accounts.db :as accounts.db]
             [status-im.chat.models :as chat]
+            [status-im.pairing.core :as pairing]
             [status-im.extensions.registry :as extensions.registry]
             [status-im.ui.components.list-selection :as list-selection]
             [status-im.ui.components.react :as react]
@@ -65,7 +66,10 @@
 
 (fx/defn handle-public-chat [cofx public-chat]
   (log/info "universal-links: handling public chat" public-chat)
-  (chat/start-public-chat cofx public-chat {}))
+  (fx/merge
+   cofx
+   (chat/start-public-chat public-chat {})
+   (pairing/sync-public-chat public-chat)))
 
 (fx/defn handle-view-profile [{:keys [db] :as cofx} public-key]
   (log/info "universal-links: handling view profile" public-key)


yenda (changes requested):
Nice finding, here is how I would recommend to fix it according to guidelines https://github.com/status-im/status-react/blob/develop/doc/codebase-structure-and-guidelines.md

The event `:chat.ui/start-public-chat` doesn't have a good name for an event, it should say what happened and not what will happen, so something like "chat.ui/public-chat-topic-pressed" for instance.
That way the event will still allow for the inclusion of the semantics that you propose, which is to navigate to the chat if it is already open.

The number one argument for doing this is that we want to keep the logic in the views as small as possible to make things more testable. You can easily write a test that shows that `chat.ui/public-chat-topic-pressed` returns the effects to navigate to the right chat when it already exists or create the chat and navigate to it when it doesn't, but it is harder to do so if the logic is in the view in the first place.
Also keep in mind that reagent component rerender when their argument are changing, `active-chat` subscription aggregates a lot of things (contacts, chats, messages..) so it will trigger a lot of unnecessary re-renders (probably not a big deal though in that case)

Keep up the good work and thank you very much for your contributions and interest in learning Clojure!

bitsikka:
> Nice finding, here is how I would recommend to fix it according to guidelines https://github.com/status-im/status-react/blob/develop/doc/codebase-structure-and-guidelines.md
> 
> The event `:chat.ui/start-public-chat` doesn't have a good name for an event, it should say what happened and not what will happen, so something like "chat.ui/public-chat-topic-pressed" for instance.
> That way the event will still allow for the inclusion of the semantics that you propose, which is to navigate to the chat if it is already open.
> 
> The number one argument for doing this is that we want to keep the logic in the views as small as possible to make things more testable. You can easily write a test that shows that `chat.ui/public-chat-topic-pressed` returns the effects to navigate to the right chat when it already exists or create the chat and navigate to it when it doesn't, but it is harder to do so if the logic is in the view in the first place.
> Also keep in mind that reagent component rerender when their argument are changing, `active-chat` subscription aggregates a lot of things (contacts, chats, messages..) so it will trigger a lot of unnecessary re-renders (probably not a big deal though in that case)
> 
> Keep up the good work and thank you very much for your contributions and interest in learning Clojure!

Thanks @yenda !

Totally make sense! I see what you mean.

Thanks for a detailed feedback. It did feel a bit off, implementing this PR the way it is.

I shall go about refactoring as you suggested.

yenda:
@bitsikka yes please, also note that there is other occurrences of `start-public-chat`  event dispatch, for instance in browser

bitsikka:
> @bitsikka yes please, also note that there is other occurrences of `start-public-chat` event dispatch, for instance in browser

Understood. This refactoring will touch a bit wider area.

bitsikka:
@yenda in the changes I just pushed I kept the event name same `:chat.ui/start-public-chat` for the shake of touching less number of files than necessary. The event with the name and the handler with the same name was used in quite a lot of different places. In some cases the "pressed" part did not make sense.

Maybe this thing is complecting too many concerns? 🤔 

Apart from that, now it should handle all cases of checking-chat-exists(only if not, then creating it) and navigating to it.

Should I still change the name?

bitsikka:
Not sure why the builds are failing. Can't see the *Details*. I get **Access Denied** when trying to see the details.

bitsikka:
> Not sure why the builds are failing. Can't see the _Details_. I get **Access Denied** when trying to see the details.

nvm.. circular dependency

src/status_im/chat/models.cljs
```diff
@@ -265,13 +265,21 @@
...
-            (when platform/desktop?
-              (desktop.events/change-tab :home))))
+  (let [chats           (into []
+                              (comp (filter :is-active)
+                                    (filter :public?)
+                                    (filter #(= topic (:chat-id %))))
+                              (vals (get-in cofx [:db :chats])))
+        already-active? (seq chats)]
+    (if already-active?
+      (if (not dont-navigate?) (navigate-to-chat cofx topic opts) cofx)
```
yenda: you never want to return the cofx as effects 
the if here could be a when

yenda (changes requested):

src/status_im/chat/models.cljs
bitsikka: got it

bitsikka:
@yenda fixed the `if to when` as well as including back an desirable(i think) improvement that was causing circular dependency before

bitsikka:
example of "improvement"

an inconsistency bug
there would've been no channel sync between devices if a public chat was created as a result of clicking a Universal link

yenda:
@bitsikka why did you use a negative predicate `dont-sync?` instead of `sync?`, I think even `dont-navigate` would be better as `navigate?` but doesn't need to be fixed in this PR. It's weird to have negative predicates and lead to double negations in the code which can be confusing

src/status_im/chat/models.cljs
bitsikka: > @bitsikka why did you use a negative predicate `dont-sync?` instead of `sync?`, I think even `dont-navigate` would be better as `navigate?` but doesn't need to be fixed in this PR. It's weird to have negative predicates and lead to double negations in the code which can be confusing

It is weird indeed - I was just following `dont-navigate?` which existed before, and it kind-of made sense when coming from dispatch site. `dont-sync?` is only ever set in `status-im.pairing.core/sync-public-chat`, everywhere else it is ignored hence nil/false implicitly by-default.

bitsikka:
> @bitsikka why did you use a negative predicate `dont-sync?` instead of `sync?`, I think even `dont-navigate` would be better as `navigate?` but doesn't need to be fixed in this PR. It's weird to have negative predicates and lead to double negations in the code which can be confusing

It is weird indeed - I was just following `dont-navigate?` which existed before, and it kind-of made sense when coming from dispatch site. `dont-sync?` is only ever set (to true) in `status-im.pairing.core/sync-public-chat`, everywhere else it is ignored hence nil/false implicitly by-default.

yenda:
@bitsikka then you can give it a default value of true 

`{:keys [dont-navigate? sync?] :or {sync? true} :as opts }`

bitsikka:
> @bitsikka then you can give it a default value of true
> 
> `{:keys [dont-navigate? sync?] :or {sync? true} :as opts }`

nice 😄 I like that very much 👍 

fix coming up

bitsikka:
done!

so far so good

gained a bit! thanks!

annadanchenko:
@bitsikka is it ready for review again?

bitsikka:
> @bitsikka is it ready for review again?

@annadanchenko yes

bitsikka:
this will probably conflict while merging with #7963, so I'll rebase and update it once again after #7963 is merged

bitsikka:
rebased and ready for review

bitsikka:
> example of "improvement"
> 
> an inconsistency bug
> there would've been no channel sync between devices if a public chat was created as a result of clicking a Universal link

fwiw, just captured this in a new issue #7992

oskarth:
@bitsikka Thanks for your patience and contributions!

@flexsurfer @yenda @rasom This PR has been open for a while, and it's quite tricky for non-core contributors to know exactly what magic button to push to move things forward.

Please lets try to follow what we agreed upon here https://github.com/status-im/status-react/blob/develop/doc/decisions/0011-tweak-pr-process.md and make sure reviews happen in a timely manner.

oskarth:
ping @flexsurfer @yenda @rasom

flexsurfer (approved):
sorry for the delay! lgtm, thanks!

rasom (approved):

mandrigin:
@yenda is your "Request Changes" still actual? If no, can you dismiss or re-review this?

src/status_im/chat/models.cljs
```diff
@@ -282,14 +288,19 @@
...
+      (navigate-to-chat cofx topic opts))
+    (fx/merge cofx
+              (add-public-chat topic)
+              #(when-not dont-navigate?
+                 (navigate-to-chat % topic opts))
+              (public-chat/join-public-chat topic)
+              #(when platform/desktop?
+                 (desktop.events/change-tab % :home))
+              #(when sync?
+                 {:dispatch [:chat.ui/sync-public-chat topic]}))))
```
yenda: this shouldn't dispatch an event but simply call `(pairing/sync-public-chat topic))` directly

yenda (changes requested):

yenda:
one change required and good to go

src/status_im/chat/models.cljs
yenda: as a rule of thumb if your event name sounds like an action, it probably doesn't need an event. Events are about what happened not what will happen

bitsikka: It causes a circular dependency though, that's the only reason I did it like that

bitsikka:
> one change required and good to go

@yenda

The Change you requested causes circular dependency, and I don't know yet any other way to avoid that other than the way I approached it. Do you know how else to avoid it?

src/status_im/chat/models.cljs
yenda: circular dependencies means that some namespaces might be doing more that what they should but to avoid having to fix that in that PR, I would say it is ok to have a fx/merge in events.cljs since it was already the case.

bitsikka: but then it would not fix https://github.com/status-im/status-react/issues/7992 case

bitsikka: It has been a while since I worked on this and I have a bit of a broken/slow build system hindering work at the moment, so off the top of my head I'm now wondering if moving `sync-public-chat` from pairing to `chat.models` is a good idea to resolve this :thinking:

yenda: I just had a look and the circular dependency comes from handle-sync-installation which call `start-public-chat`, wouldn't this create an endless loop of syncing messages between two devices?

In any case I would add pairing to handle-public-chat as well with a merge fx there.

bitsikka: > wouldn't this create an endless loop of syncing messages between two devices?

I thought not if we set :sync? to false in `handle-public-chat` which we do there and only there

bitsikka: > In any case I would add pairing to handle-public-chat as well with a merge fx there.

not sure I get what you mean by that 😕

bitsikka: but:

how about if, in here
https://github.com/status-im/status-react/blob/285df85d920bd3fd1fa76ef5193fd2c02e53bf57/src/status_im/pairing/core.cljs#L293

we dispatch `:chat.ui/start-public-chat` instead of calling `models.chat/start-public-chat`?

because that is the only line, as you have noticed, that is requiring `chat.models` in pairing namespace

yenda: @bitsikka to solve #7992, instead of sync in start-public-chat with sync? parameter you move sync outside as it was before and add it for universal link using fx/merge

The long-term fix will probably involve separating sending pairing messages and receiving them

bitsikka: I think I see. Ok, interpreting that in update PR pronto. Will ping ya once done

bitsikka:
@yenda done!

yenda (approved):

yenda:
@bitsikka Looks good thanks!

churik:
@bitsikka would appreciate if you can rebase it in latest develop for testing.
Also I didn't quite understand, issues  #7952, #7974, #7992  should be addressed in separate PRs, right?
BTW, thanks a lot for uncovering those!

bitsikka:
@churik rebased

All the issues are handled in this PR

I'm sorry about terse descriptions and lack of tests - will do better next time

**Joining already joined public chat causes never ending "Loading..." #7952**

1) In device A add public chat like #status
2) In same device go to + button and add the same public chat #status

The "Never ending Loading.." part is not happening in my brief test anymore for some reason, However, still the underlaying not necessary and `non-visual` side-effects to app-db and realm are happening which this PR would prevent

I should've added tests to make this more apparent - sorry about that

**upon re-syncing unseen message badges disappear(marked seen) without seeing the new messages #7974**

1) Have a couple of public chats in device A
2) Pair and sync A with device B
3) be in Home/Chat list screen on both devices
4) Wait and let both devices receive messages in any number of public chats (so message badges appear)
5) In Device A - go to Profile>Devices press "Sync all devices" again
6) Notice in device B - unread badges disappear even though you have not read the messages in any of the devices yet

**new public chat started as a result of clicking a universal link is not synced across paired devices #7992**

1) Have two devices A and B paired and synced
2) Open a Universal Link for a NEW public chat in device A
3) Check device B - it does not have the new public chat

Initially, the implementation was such that this had to be in this PR. Implementation changed towards the end with @yenda's guidance, and this should then have been a separate PR, but it is not - it is in this PR as well. Sorry about that.

churik:
Tested on IOS / Android.
All issues are fixed.
As regression checked devices pairing and creating different types of chats.

Also briefly checked that can install \ join the public chat from Mac OSx build.
@bitsikka thanks for a great job!

PR #7831, opened 2019-03-26
testing cud mode with test DNS record

jakubgs:
This is only for testing purposes, no need for reviews, __DO NOT MERGE__.

### DIFF
--- a/src/status_im/chaos_mode/core.cljs
+++ b/src/status_im/chaos_mode/core.cljs
@@ -7,7 +7,7 @@
 
 (def interval-5m (* 5 60 1000))
 (defonce interval-id (atom nil))
-(def url "https://cloudflare-dns.com/dns-query?name=chaos-unicorn-day.status.im&type=TXT")
+(def url "https://cloudflare-dns.com/dns-query?name=chaos-unicorn-day-test.status.im&type=TXT")
 
 (defn chaos-mode-switched [chaos-mode?]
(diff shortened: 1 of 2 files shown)


mandrigin:
@annadanchenko I'll rebase it, give me 1 minute

annadanchenko:
testing is done 
1. when dns entry is on and staging cluster is off: User gets popup with a unicorns and link to details only once when chaos mode is turned on.After this can't sent transaction (500 error), can't see history of transactions, can't see usd value in wallet/chat, can't get chat history but can send messages. Unfortunately, whisper nodes are re-used from other fleets, so peers were available. Can add custom bootnode and mailserver after this can get history (used QRs from https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d ). 
2.  when dns entry is switched off and when staging cluster is back: Tx and messages can be sent without restarting app. Chaos mode is switched off automatically

jakubgs:
Wait, shouldn't the changes done by @mandrigin and @cammellos be merged? Or were those just for tests?

annadanchenko:
> Wait, shouldn't the changes done by @mandrigin and @cammellos be merged? Or were those just for tests?

there was separate pr from @cammellos that is merged: https://github.com/status-im/status-react/pull/7839

jakubgs:
Cool.

PR #7836, opened 2019-03-26
[TESTS] Added test sign_typed_message

antdanchenko:
test case is https://ethstatus.testrail.net/index.php?/cases/view/5784


Serhy (approved):
I'd use 'create new user' for this test instead recover, but doesn't really matter.

churik (approved):

antdanchenko:
We need to recover the account in order to [check](https://github.com/status-im/status-react/pull/7836/files#diff-2b2286014877157679a2e0a6e00a5bf8R79) expected hash, the hash is always the same for the account
Expected result can't be hardcoded in case of creating new account

PR #7839, opened 2019-03-26
Set chaos mode every time, dont show the popup everytime

cammellos:
We prevented setting chaos mode when the flag did not change, in order not to show the pop up, this changes the behavior so that we don't show the pop up, but we do set chaos mode in status-go etc.

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -180,17 +180,17 @@
  :accounts.ui/chaos-mode-switched
  (fn [{:keys [db] :as cofx} [_ chaos-mode?]]
    (let [old-chaos-mode? (get-in db [:account/account :settings :chaos-mode?])]
-     (when (not= old-chaos-mode? chaos-mode?)
-       (fx/merge
-        cofx
-        (when chaos-mode?
-          {:ui/show-confirmation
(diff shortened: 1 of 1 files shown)


mandrigin (approved):

rasom (approved):

annadanchenko:
to test it I'll need it to check test dns entry, so @mandrigin cherry-picked this to https://github.com/status-im/status-react/pull/7831 but ios build is failing there now... maybe it would be faster to update test url here, so in case more changes are needed then I will not wait for updates/builds in #7831

mandrigin:
@annadanchenko but it needs to be rebuilt here anyway, so I don't see why is it better here

annadanchenko:
> @annadanchenko but it needs to be rebuilt here anyway, so I don't see why is it better here

in case more changes will be needed for this PR - it will need retesting on its new builds

annadanchenko:
@cammellos looks ok now. I've triggered test in https://ci.status.im/job/end-to-end-tests/job/status-app-end-to-end-tests/2612/ but it's ok to merge to develop and cherry pick to release as the chaos mode is not covered in tests anyway. if there will be regressions I'll ask to revert

mandrigin:
will merge it now then

PR #7840, opened 2019-03-26
Replace usage of stdenvNoCC with default stdenv

pedropombeiro:

### Summary:

While working on migrating the whole project to Nix, I encountered some issues that were fixed by moving from using `stdenv` to `stdenvNoCC`. This PR focuses on finding the root causes and going back to using the standard `stdenv`.

### Testing notes (optional):

This PR affects mostly the Desktop app, but also potentially Android which is built on Linux. No need for in depth testing, but at least check that the app still works on the indicated platforms.

#### Platforms
- Android
- macOS
- Linux

status: ready


mandrigin (approved):

scripts/build-desktop.sh
```diff
@@ -79,6 +79,9 @@ if is_windows_target; then
   CMAKE_EXTRA_FLAGS="$CMAKE_EXTRA_FLAGS -DCMAKE_C_COMPILER='x86_64-w64-mingw32.shared-gcc'"
   CMAKE_EXTRA_FLAGS="$CMAKE_EXTRA_FLAGS -DCMAKE_CXX_COMPILER='x86_64-w64-mingw32.shared-g++'"
   CMAKE_EXTRA_FLAGS="$CMAKE_EXTRA_FLAGS -DCMAKE_RC_COMPILER='x86_64-w64-mingw32.shared-windres'"
+elif is_macos; then
+  CMAKE_EXTRA_FLAGS="$CMAKE_EXTRA_FLAGS -DCMAKE_C_COMPILER='gcc'"
+  CMAKE_EXTRA_FLAGS="$CMAKE_EXTRA_FLAGS -DCMAKE_CXX_COMPILER='g++'"
```
jakubgs, line 84: So because the previous env that you used had no C compiler it was using system `gcc` and `g++` correct? And by adding this you make it you the ones that's from Nix env?

pedropombeiro: Correct, otherwise it would use Nix's clang and we get a CMake error when using that.

jakubgs: Thanks for explanation!

jakubgs (approved):

pedropombeiro:
Dang, iOS build seems to not like `stdenv` (it is passing a gcc `ld` flag that the Apple `ld` executable doesn't know about, so we should use the gcc `ld`). The fact that I can't easily build it locally doesn't help.

churik:
Tested fresh installation on Windows 10, Mac OSx, upgrade from latest nightly on Mac OS and Linux Ubuntu 18.04, IOS 11.4.1 and Android 8.
In all cases checked account creation and joining public chat.

PR #7849, opened 2019-03-28
use call-private-rpc in ethereum/call

yenda:
- remove web3 parameter for ethereum/call
- remove first parameter for callback of ethereum/call which was always nil

status: ready

### DIFF
--- a/src/status_im/utils/ethereum/core.cljs
+++ b/src/status_im/utils/ethereum/core.cljs
@@ -1,6 +1,7 @@
 (ns status-im.utils.ethereum.core
   (:require [clojure.string :as string]
             [status-im.js-dependencies :as dependencies]
+            [status-im.native-module.core :as status]
             [status-im.utils.ethereum.tokens :as tokens]
             [status-im.utils.money :as money]
             [taoensso.timbre :as log]))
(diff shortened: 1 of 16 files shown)


src/status_im/utils/ethereum/core.cljs
```diff
@@ -124,8 +125,16 @@
 (defn- sig->method-id [signature]
   (apply str (take 10 (sha3 signature))))
 
-(defn call [web3 params cb]
-  (.call (.-eth web3) (clj->js params) cb))
+(defn call [params callback]
+  (status/call-private-rpc
+   (.stringify js/JSON (clj->js {:jsonprc "2.0"
+                                 :id      1
+                                 :method  "eth_call"
```
flexsurfer, line 132: could you move it in constants

src/status_im/utils/ethereum/core.cljs
```diff
@@ -124,8 +125,16 @@
...
+                                 :params  [params "latest"]}))
+   (fn [response]
+     (if (= "" response)
+       (log/warn :web3-response-error)
+       (callback (get (js->clj (.parse js/JSON response)) "result"))))))
```
flexsurfer, line 137: we have `types/json->clj`, or you don't need to keywordize it here?

src/status_im/utils/ethereum/ens.cljs
```diff
@@ -18,98 +18,114 @@
...
-  (ethereum/normalized-address (if (string/blank? s)
-                                 default-namehash
-                                 (let [[label remainder] (-> s
-                                                             string/lower-case
-                                                             (string/split #"\." 2))]
-                                   (ethereum/sha3 (+ (namehash remainder)
-                                                     (subs (ethereum/sha3 label) 2))
-                                                  {:encoding "hex"})))))
+(defn namehash
+  [s]
```
flexsurfer, line 22: why do you move args down? i return it back every time :) we need a consensus

flexsurfer (approved):

src/status_im/utils/ethereum/core.cljs
yenda: no there is a circular dependency

src/status_im/utils/ethereum/core.cljs
yenda: no I don't need to because I use only result here

src/status_im/utils/ethereum/ens.cljs
yenda: we should always put they down, easier to add a docstring, easier to keep the line length limit

vitvly (approved):

PR #7868, opened 2019-03-29
[feature] contracts namespace

yenda:
-  introduce `open-sign-transaction-flow`, effect version of `open-modal-wallet-for-transaction`
-  The contract namespace provides the `call` effect:
    - takes contract identifier from contracts, method, params and callback
    - select the correct contract address based on current network
    - encode params, decodes results
    - start wallet transaction flow for write calls

The goal is to unify and facilitate the use of contracts, by describing them as data

status: ready


vitvly (approved):

yenda:
@goranjovic @flexsurfer can you review?

flexsurfer (approved):

PR #7869, opened 2019-03-29
Add reading nodes from contract

cammellos:
This PR is part of network incentivisation. It adds a way for a client
to pull nodes from a contract.

This is done by selecting the `eth.contract` fleet. If that is selected
on login it will fetch nodes from a contract and pass them to status-go.
If these can't be fetched, it will default to `eth.beta`.

Currently contract information are hard-coded, but eventually the user
will be able to add their own (probably).

Toggled off in release.

There's no discovery, only static nodes for nodes for now.

### Testing

1) Go to fleets
2) Select `eth.contract`
3) Login
4) Everything should work as expected (roughly)

If login is when offline it might throw an exception `unable to find peer with id`, that's something it will be fixed at later stages.

Upgrades status-go, small changes but good to do some sanity checks.

status: ready


src/status_im/accounts/login/core.cljs
```diff
@@ -14,12 +16,33 @@
...
             [status-im.i18n :as i18n]
             [status-im.node.core :as node]
             [status-im.ui.screens.mobile-network-settings.events :as mobile-network]
             [status-im.chaos-mode.core :as chaos-mode]))
 
+(def rpc-endpoint "https://goerli.infura.io/v3/f315575765b14720b32382a61a89341a")
+(def contract-address "0xfbf4c8e2B41fAfF8c616a0E49Fb4365a5355Ffaf")
+(def contract-fleet? #{:eth.contract})
+
+(defn fetch-nodes [current-fleet resolve reject]
```
cammellos, line 30: Test this

src/status_im/fleet/core.cljs
```diff
@@ -23,10 +23,15 @@
...
   (not (nil? (some #(= fleet %) fleets-with-les))))
 
-(def fleets
-  (reduce merge (map #(:fleets (types/json->clj %))
-                     [(slurp "resources/config/fleets.json")
-                      (slurp "resources/config/fleets-les.json")])))
+(def default-fleets (slurp "resources/config/fleets.json"))
+(def default-les-fleets (slurp "resources/config/fleets-les.json"))
+
+(defn fleets [{:keys [custom-fleets]}]
```
cammellos, line 29: Test this

src/status_im/fleet/core.cljs
```diff
@@ -42,11 +47,11 @@
           {}
           mailservers))
 
-(def default-mailservers
+(defn default-mailservers [db]
```
cammellos, line 50: Test this

src/status_im/fleet/core.cljs
```diff
@@ -57,6 +62,30 @@
...
+                      (assoc acc
+                             (keyword n)
+                             n))
+                    {}
+                    nodes))]
+    {:boot (format-nodes nodes)
+     :mail (format-nodes nodes)
+     :whisper (format-nodes nodes)}))
+
+(fx/defn set-nodes [{:keys [db]} fleet nodes]
```
cammellos, line 77: Test this

yenda:
is the method to read nodes hardcoded in status-go?
isn't there a chicken/egg problem when fetching bootnodes from contract?

cammellos:
@yenda the method to read nodes from the contract is in status go https://github.com/status-im/status-go/pull/1429, the ABI is in there as well, but there's nothing "hardcoded", it can only be used with that particular contract of course, but different deployment/networks, is that what you mean?

In terms of chicken/egg the node is started first without passing any nodes and does not require any connectivity (it is started in order to verify user credentials), once credentials are verified, we fetch the nodes from the contract and re-start the node with the correct `ClusterConfig` (this is the same as current login flow, that hasn't changed, other than in-between we fetch nodes from the contract if the user has chosen a "contract" fleet)

yenda:
@cammellos I mean that normally we call contracts from status-react using call-private-rpc, I was wondering why this particular contract call requires its own binding, is it to reduce back and forth between go and react side?

Regarding the chicken/egg I guess it doesn't apply as long as we use rpc? but I meant that you won't be able to interrogate a contract if you don't have access to rpc or valid bootnodes in the first place

cammellos:
Yes, partially for that reason, as currently we make a few calls in sequence to the contract, as we want to return an array of strings but solidity does not allow that, so we need to get the count and fetch each element separately. Also status-go will have to interact directly with the contract for registering and other actions, so it seemed apt to keep the code in status-go and just provide an endpoint.

In terms of RPC, currently we  hit infura (or any other RPC endpoint that the user passes), so yes, as long as RPC are available we will be able to query it.

yenda:
@cammellos why not store the stuff in ipfs and have the contract references contenthashes as we do with tribute to talk?

cammellos:
it's an option yes, we might go for it in the future, also we would not have to worry about pinning as we can ask status-nodes to do that as part of the service. 
Currently though it's just simpler and less committal to simply store strings in an array, and we can probably revisit the decision once we are a bit further in the implementation.

yenda:
I wouldn't say it is simpler, all the stuff from tribute to talk can be reused to support that, also it is less expensive to store a hash on chain than a string

cammellos:
It is less expensive true, but adding a dependency (ipfs), it's clearly adding complexity, surely not reducing it, and it this stage I don't see any reason to do so, as we still have to flesh out major parts of the implementation and things are likely to change, but as I said, we might just end up doing that in the near future, it's just not the right time to make that decision in my opinion.

rasom:
> it is started in order to verify user credentials

I believe `Statusgo.verify` can be called without running node, not sure if this changes anything in the context of conversation though.

cammellos:
Thanks @rasom even better, currently querying nodes does not also need the node to be running

mandrigin (approved):

rasom (approved):

Serhy:
@cammellos two things related to `eth.contract` fleet 
1) Once I connected to eth.contract fleet there is a list of four mailservers - all have the same name `enode`. Are those mailserver in static contract indeed has the same name value (or do not have name value at all so we default to `enode:` name)?

Screenshot 2019-04-03 at 15 55 39

2) If login to account with eth.contract fleet set while device is offline - there are several more mailservers are listed on Mailserver screen, not sure where do they come from (didn't see `unable to find peer with id` error neither in logcat nor in app though)

Screenshot 2019-04-03 at 15 57 35


iOS build job continue failing, - Times Out on Prep step: https://ci.status.im/job/status-react/job/prs/job/ios/view/change-requests/job/PR-7869/4/ @jakubgs could you help to understand what happens?

jakubgs:
Our Nix cache for MacOS is broken right now, so builds for iOS and MacOS will fail. Me and Pedro are working to fix it.

cammellos:
@Serhy currently they have a very long name, so probably it does not display correctly, you see more when offline as we default to the beta fleet, but we still use long names, probably I won't be fixing this in the scope of this PR if you are ok with it

jakubgs:
Builds should be fixed now, at least from Nix cache perspective.

Serhy:
Nix cache fixed, but iOS build keep failing something with https://ci.status.im/job/status-react/job/prs/job/ios/view/change-requests/job/PR-7869/7/console

```
12:07:25  ❌  /Users/jenkins/workspace/status-react_prs_ios_PR-7869/modules/react-native-status/ios/RCTStatus/RCTStatus.m:259:19: expected method body
12:07:25  
12:07:25  (NSString *) address
12:07:25  ~~~~~~~~~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
12:07:25  
12:07:25  
12:07:25  
12:07:25  ❌  /Users/jenkins/workspace/status-react_prs_ios_PR-7869/modules/react-native-status/ios/RCTStatus/RCTStatus.m:260:62: expected identifier or '('
12:07:25  
12:07:25  callback:(RCTResponseSenderBlock)callback) {
12:07:25  ^
12:07:25  
12:07:25  

```
@cammellos it's something with the new RCT_EXPORT_METHOD. Could you help with that "missing method body", please? (should it probably be `address:(NSString *) address`?

cammellos:
@Serhy I think you are right, thanks, well spotted!

Serhy:
Latest builds tested on iOS 12.1.4 and Android 8.1 and confirmed having the same results as before (in https://github.com/status-im/status-react/pull/7869#issuecomment-479482788)

In order to merge this PR we need to test new status-go-beta.10 with release 0.11 which we are doing now... We'll let know when we are good to merge

churik:
**2) Contacts are not synced between 0.11.0 mobile builds <-> and this PR builds**

Steps:
- Device A: create account on 0.11.0 (tried on IOS), add contact
- Device B: restore the same account on PR build (tried Mac OS and IOS builds)
- Pair devices A and B
- Device A: tap on sync all

Expected: public chats and contact are shown on Device B.
Also group chats and 1-1 chats are synced correctly.
Actual result: contacts are not synced.

Logs for device B:
[geth.log](https://github.com/status-im/status-react/files/3044118/geth.log)
[Status.log](https://github.com/status-im/status-react/files/3044121/Status.log)

Note, that it is not reproducible with current desktop build (I'm using 0.9.0, status-go v 0.23.0-beta8.chaos), but reproducible with mobile (0.11.0, status-go v 0.23.0-beta8.chaos)

churik:
@Serhy 
Tested on Linux Ubuntu 18.04, Windows 10 Home, Mac OS:
- fresh install;
- messaging;
- upgrade from 0.11.0 desktop release with set of data;
- compatibility with 0.11.0 release;
- switching mailservers;
- enable \ disable logging;
- pairing.

No regression is found.

annadanchenko:
tested ok Android/iOS builds for regressions (using staging fleet)

annadanchenko:
sorry started tests using this PR number but other apk - just ignore autotest results below

Serhy:
Looks OK for me Release 0.11.0 on staging fleet. 
Lets merge, @cammellos !

PR #7873, opened 2019-04-02
set maxFontSizeMultiplier for text and text-input components on iOS

rasom:
according to https://github.com/status-im/status-react/pull/7629#issuecomment-468605192

status: ready

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -87,17 +87,24 @@
 
 (def slider (get-class "Slider"))
 
+(def max-font-size-multiplier 1.25)
+
+(defn prepare-text-props [props]
+  (-> props
+      (update :style typography/get-style)
+      (assoc :max-font-size-multiplier max-font-size-multiplier)))
+
 ;; Accessor methods for React Components
 (defn text
   "For nested text elements, use nested-text instead"
   ([text-element]
    [text {} text-element])
   ([options text-element]
-   [text-class (update options :style typography/get-style) text-element]))
+   [text-class (prepare-text-props options) text-element]))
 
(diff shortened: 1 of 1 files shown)


mandrigin (approved):
lgtm

flexsurfer (approved):

churik:
@rasom 
Font is still too large for DApps permissions popup:

I used max font (on 5 size I still able to see the buttons, on 6 and more (size font is the same starting from 6) - buttons are completely overlapped by footer.

Tested on IOS 11.4.1, IPhone 7.

mandrigin:
@churik maybe we can make it a separate issue and issue a bounty on it?

churik:
@mandrigin 
there is separate issue for this https://github.com/status-im/status-react/issues/7215 - issue is exactly about these screens, but in discussion turned out that problem is more general (text size).

So I don't mind actually - we can test and merge this PR, but the initial issue is still not fixed.

rasom:
@churik @mandrigin ok then, let's fix that specific case separately

errorists (approved):

churik:
Tested on IOS 11.4.1 with the largest font (from accessibility settings).
Tested:
- send, request transactions from chat and wallet
- another user profile view
- pairing (contact list  with 30+ contacts)
- meilserver, bootnodes, scan qr code views
- add custom network screen
- transaction history \ transaction view.
Except of small cutting the bottom part (pretty minor):
photo_2019-04-04 11 01 40 jpeg 2019-04-04 11-02-17 

and  https://github.com/status-im/status-react/issues/7215 (which should be separate bounty) - looks good to me.
Also checked that I can use Android version with largest font (LG V20, Android 8)

PR #7874, opened 2019-04-02
add repo name to ghcmgr url

jakubgs:
Small fix to use the new URl for [`ghcmgr`](https://github.com/status-im/github-comment-manager) API which includes repo name, since now it supports multiple repos.


mandrigin (approved):

rasom (approved):

jakubgs:
There is no need to test this.

churik:
I see, only waiting for e2e. 
Well, it shouldn't be merged without e2e tests.

jakubgs:
The change touches only and exclusively our CI setup, and specifically something that posts comments on GitHub about builds. There is not a single thing that has been changed about the app or how it's being built in this PR.

PR #7879, opened 2019-04-02
fix nix cache build process

jakubgs:
This is a fix for nix-cache builds failing due to me running `nix-collect-garbage`. It's a bit cleaner.
Though we should rework our nix setup so we can just do `nix build`.

This change doesn't in any way modify App build process, just nix cache.


pedropombeiro (approved):
Great job!

mandrigin (approved):
👍

PR #7880, opened 2019-04-03
Scan url with QR scanner

krisc:
This PR is for this bounty https://github.com/status-im/status-react/issues/5776

It fixes the issue as described, but I left it WIP because I'm not sure if additional checks have to be made or not. For example, if the qr-code returns data `30032621` should this show an error message or should it just navigate to whatever that data is in the browser? 

If it should show an error, do we have a function already in status that checks for valid urls?

### DIFF
--- a/src/status_im/ui/screens/add_new/models.cljs
+++ b/src/status_im/ui/screens/add_new/models.cljs
@@ -4,13 +4,21 @@
             [status-im.i18n :as i18n]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.universal-links.core :as universal-links]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.browser.core :as browser]
+            [status-im.constants :as constants]
+            [status-im.extensions.core :as extensions]))
+
+(defn- valid-url? [url]
+  (boolean (re-matches constants/regx-url url)))
 
 (fx/defn handle-qr-code
   [cofx data]
   (if (spec/valid? :global/public-key data)
     (universal-links/handle-view-profile cofx data)
     (or (universal-links/handle-url cofx data)
-        {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
-                            :content (i18n/label :t/use-valid-qr-code {:data data})
-                            :on-dismiss #(re-frame/dispatch [:qr-scanner.ui/qr-code-error-dismissed])}})))
+        (if (valid-url? data)
+          (browser/handle-message-link cofx data)
+          {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
+                              :content (i18n/label :t/use-valid-qr-code {:data data})
+                              :on-dismiss #(re-frame/dispatch [:qr-scanner.ui/qr-code-error-dismissed])}}))))
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -199,6 +199,12 @@
     (utils.universal-links/handle-url cofx link)
     {:browser/show-browser-selection link}))
 
+(fx/defn browser-selection-cancel
+  [{:keys [db]}]
+  (let [view-id (:view-id db)]
+    (if (= view-id :qr-scanner)
+      {:db (assoc-in db [:navigation/screen-params view-id :barcode-read?] false)})))
+
 (fx/defn update-browser-on-nav-change
   [cofx url error?]
   (let [browser (get-current-browser (:db cofx))
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -1412,6 +1412,11 @@
    (browser/handle-message-link cofx link)))
 
 (handlers/register-handler-fx
+ :browser.ui/browser-selection-cancel
+ (fn [cofx [_ _]]
+   (browser/browser-selection-cancel cofx)))
+
+(handlers/register-handler-fx
  :browser.ui/remove-browser-pressed
  (fn [cofx [_ browser-id]]
    (browser/remove-browser cofx browser-id)))
--- a/src/status_im/ui/components/list_selection.cljs
+++ b/src/status_im/ui/components/list_selection.cljs
@@ -39,7 +39,8 @@
                         :action #(re-frame/dispatch [:browser.ui/open-in-status-option-selected link])}
                        {:label  (i18n/label (platform-web-browser))
                         :action #(.openURL react/linking (http/normalize-url link))}]
-         :cancel-text (i18n/label :t/browsing-cancel)}))
+         :cancel-text (i18n/label :t/browsing-cancel)
+         :on-cancel   #(re-frame/dispatch [:browser.ui/browser-selection-cancel])}))
 
 (defn browse-in-web-browser [link]
   (show {:title       (i18n/label :t/browsing-title)
--- a/translations/en.json
+++ b/translations/en.json
@@ -381,7 +381,7 @@
     "currency-display-name-npr": "Nepal Rupee",
     "backup-recovery-phrase": "Backup recovery phrase",
     "send-request": "Send request",
-    "use-valid-qr-code": "This QR code doesn't contain a valid universal link, contact code or username: {{data}}",
+    "use-valid-qr-code": "This QR code doesn't contain a valid universal link, url, contact code or username: {{data}}",
     "paste-json": "Paste JSON",
     "browsing-site-blocked-title": "This site is blocked",
     "browsing-site-blocked-description1": "We detected potential malicious activity from this address. To protect you and your wallet, we're preventing further navigation.\n\nIf you think this is an error, let us know in the ",


mandrigin:
@krisc thanks for your contribution! maybe @yenda or @flexsurfer can help you with the error function. But I would check if what is returned is similar to a URL, and show error otherwise.

env/dev/env/config.cljs
```diff
@@ -1,6 +1,6 @@
 (ns env.config)
 
 (def figwheel-urls {:android "ws://localhost:3449/figwheel-ws",
- :ios "ws://192.168.0.9:3449/figwheel-ws",
+ :ios "ws://192.168.1.84:3449/figwheel-ws",
```
mandrigin: this change should be reverted

mandrigin (changes requested):

mandrigin:
@krisc also, the formatting should be fixed (that's why CI builds are failing). I recommend running `lein cljfmt fix` to fix formatting.

krisc:
> @krisc also, the formatting should be fixed (that's why CI builds are failing). I recommend running `lein cljfmt fix` to fix formatting.

Oh, thanks! wasn't aware of that command. I get that error every time and didn't know why.

krisc:
> @krisc thanks for your contribution! maybe @yenda or @flexsurfer can help you with the error function. But I would check if what is returned is similar to a URL, and show error otherwise.

@yenda or @flexsurfer , is there a function somewhere in status that checks for vaild URLs? Perhaps someone could direct me to the code where URLs in chat messages are rendered as blue and underlined?

For now, I'm just checking if the string has a "." in it.

krisc:
@mandrigin This should be it unless I should be error checking a different way. And FYI, the case where browser selection is canceled and then being able to scan another qr-code depends on code that came from this PR https://github.com/status-im/status-react/pull/7745

This PR pretty much has the aforementioned PR merged since it shares so much code.  The other PR should probably be merged to upstream first.

mandrigin:
@krisc sounds good, then let's first land #7745 and then this one. I will take a look at the code today.

src/status_im/ui/screens/add_new/models.cljs
```diff
@@ -4,19 +4,16 @@
...
-        {:utils/show-popup {:title      (i18n/label :t/unable-to-read-this-code)
-                            :content    (i18n/label :t/use-valid-qr-code {:data data})
-                            :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
-
-(fx/defn handle-qr-code
-  [cofx data]
-  (fx/merge cofx
-            (navigation/navigate-to-clean :home {})
-            (process-qr-code data)))
+        (if (clojure.string/includes? data ".")
```
mandrigin: I think you can use something like that to check for URLs:
```clojure
(defn is-url [url] 
  (try 
    (and (js/URL. url) true) 
  (catch js/Error e false)))
```

mandrigin (changes requested):
the condition needs to be changed, I think, looks good otherwise!

src/status_im/ui/screens/add_new/models.cljs
krisc: hmm this seems to always return true

mandrigin: hmm, I tried it in a repl and for me it returns `false` for, say, “aaa” and `true` for “http://google.com"...

krisc: hmm I'm using figwheel for status-im, and returns true for anything. I did try in a web repl, and it does return `false` for something like "aaa". which repl did you use?

krisc: I think this should be handled the same way a URL is handled when tapped on by a user in chat so I would like to see where in the code that renders a chat message as a link. for example, if a message contains "aaaa google.com" then the 'google.com' becomes underlined. I'm having trouble trying to find where this happens :/

mandrigin: @yenda maybe you know?

krisc:
@mandrigin @yenda I ended up using `extensions/valid-uri?` but was not yet able to test it because I'm having issues with `make shell`

mandrigin (approved):
I think that's good enough for now.

krisc:
Derp, I should have read the code https://github.com/status-im/status-react/blob/64b63d5593cebd7382b6894a71848589f9628e67/src/status_im/extensions/core.cljs#L865

It will always return `false` unless it's a status extension.

> I think this should be handled the same way a URL is handled when tapped on by a user in chat so I would like to see where in the code that renders a chat message as a link. for example, if a message contains "aaaa google.com" then the 'google.com' becomes underlined. I'm having trouble trying to find where this happens :/

If someone could point me to where this happens, that would be great!

src/status_im/ui/screens/add_new/models.cljs
yenda: we have a `regx-url` in `constants.cljs` to check for urls

yenda:
@krisc this is done in `status-im.chat.models.message-content`, the message string is parsed and urls are replaced by `[:link "the url"]`, to find urls the regexp used is in `constants/regex-url`

krisc:
> @krisc this is done in `status-im.chat.models.message-content`, the message string is parsed and urls are replaced by `[:link "the url"]`, to find urls the regexp used is in `constants/regex-url`

Exactly what I was looking for. Thanks!

krisc:
Ok, this should be it!

src/status_im/chat/models/message_content.cljs
```diff
@@ -87,12 +88,15 @@
...
   [{:keys [text] :as content}]
+  ; (log/debug "enrich-content text = " text)
+  ; (log/debug "enrich-content content = " content)
   (let [[_ metadata] (reduce (fn [[text metadata] [type regex]]
                                (if-let [matches (query-regex regex text)]
                                  [(clear-ranges matches text) (assoc metadata type matches)]
                                  [text metadata]))
                              [text {}]
                              (into stylings actions))]
+    ; (log/debug "enrich-content metadata = " metadata)
```
mandrigin: remove commented-out code

src/status_im/chat/models/message_content.cljs
```diff
@@ -1,5 +1,6 @@
 (ns status-im.chat.models.message-content
   (:require [clojure.string :as string]
+            [taoensso.timbre :as log]
```
mandrigin: unused import

mandrigin (changes requested):
a few tiny changes, looks good otherwise!

src/status_im/chat/models/message_content.cljs
krisc: oh yeah forgot to remove my debug messages

mandrigin (approved):
looks good now! just don't forget to squash the commits ;-)

krisc:
@mandrigin not sure why it's saying "jenkins/prs/android-e2e — This commit cannot be built". I ran `lein cljfmt fix` before committing

mandrigin:
@krisc looks like a random thing, restarted it

krisc:
@mandrigin BTW is there a way I can get access to the jenkins build logs? When I click on "Details" is says `
Access Denied
krisc is missing the Overall/Read permission`

mandrigin:
@krisc I think the access is limited until we audit our build logs for sensitive info... We will get there, it isn't just done yet.

rasom (approved):

churik:
@krisc 
#### 1) Part of header is not white (Android)
dsfdfsdfdsfdf
#### 2) After opening URL in native browser can't scan valid URL
Steps: 
- tap on + 
- scan QR code that contains link
- tap on open on IOS (or Android)
- go back to Status
- try to scan same QR with link
Expected result: can scan QR
Actual result: QR is not recognizable again

#### 3) Several times I saw crash on Android after scanning valid QR
Unfortunately didn't notice any specific steps to reproduce, during testing I scanned ~30 times different URLs and three times I saw this error anter tapping on "Open to Status" when scanning QR with link. Logcat: 
[crashlog.log](https://github.com/status-im/status-react/files/3084277/crashlog.log)

photo_2019-04-16 12 12 07

#### 4) Ethereum address is considered as a valid link.
I didn't expect that wallet address can be scanned with "Scan QR code" in chat.
In this case I would prefer to see error.
Example of QR:
photo_2019-04-16 12 14 49

krisc:
@churik thanks for uncovering these issues. I'll have more time to look into them next week.

krisc:
Sorry, I've been busy with Ethereal Hackathon these past few weeks. I'll have more time to get back on this after the weekend.

But these issues may be related to some code I wrote for this issue https://github.com/status-im/status-react/pull/7745#issuecomment-487472440

flexsurfer:
hey @krisc any updates on this? :)

krisc:
> hey @krisc any updates on this? :)

I was gonna try to resolve issues with https://github.com/status-im/status-react/pull/7745#issuecomment-487472440 because it might be related to the issues here.

Sorry I'm taking so long. I've been have some issues building and running since we switched to nix. So testing has been really slow for me.

krisc:
I think I got my nix issues taken care of, but the easiest thing for me to move forward with this issue is to start a new PR. hope that's okay.

flexsurfer:
yeah that's fine, thanks!

PR #7885, opened 2019-04-03
[fix] contract call effect should destructure keys not vector

yenda:
- fix destructuring in `::call` fx
- return params decoding should return nil and not "0x"

# testing

- only the stickers are using this for now

status: ready

### DIFF
--- a/src/status_im/utils/ethereum/contracts.cljs
+++ b/src/status_im/utils/ethereum/contracts.cljs
@@ -34,7 +34,7 @@
 
 (re-frame/reg-fx
  ::call
- (fn [[address data callback]]
+ (fn [{:keys [address data callback]}]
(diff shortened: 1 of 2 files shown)


src/status_im/utils/ethereum/contracts.cljs
```diff
@@ -60,4 +60,6 @@
             :on-result on-result})
           {::call {:address  contract-address
                    :data     data
-                   :callback #(callback (abi-spec/decode % return-params))}})))))
+                   :callback #(callback (when (empty? return-params)
```
vitvly: Should be an `if`?

vitvly (approved):

flexsurfer (approved):

PR #7888, opened 2019-04-03
use auto@status.im Apple account for signing app

jakubgs:
Changing to a different Apple account to avoid the Two-Factor Auth mess that Apple made mandatory for all "Holder" accounts like our support@status.im.

### DIFF
--- a/ci/Jenkinsfile.fastlane.clean
+++ b/ci/Jenkinsfile.fastlane.clean
@@ -1,33 +1,39 @@
-env.LANG="en_US.UTF-8"
-env.LANGUAGE="en_US.UTF-8"
-env.LC_ALL="en_US.UTF-8"
-env.FASTLANE_DISABLE_COLORS=1
-env.REALM_DISABLE_ANALYTICS=1
+pipeline {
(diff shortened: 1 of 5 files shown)


mandrigin (approved):
lgtm!

jakubgs:
Holy shit it works!
amazing
I think...

pedropombeiro (approved):

churik:
@jakubgs anything specific to test here, except IOS build?
Already checked:
- can install build
- can create an account
- can restore existing account

Also checked Mac OSx build

jakubgs:
I don't think so. As long as iOS installs and runs then we know the signing process was successful. Thakns for testing!

jakubgs:
I added changes in `ci/Jenkinsfile.fastlane.clean` to fix the [Clean TestFlight Users](https://ci.status.im/job/status-tools/job/Clean%20Testflight%20Users/) Jenkins job. Also modified it to use the new pipeline format.

churik:
Can install latest IOS build, create account - considering PR as ready to merge

PR #7901, opened 2019-04-03
[feature] implement erc20 get-transaction for SNT and STT

yenda:
- change `ethereum/get-transaction` to use call-private-rpc instead
of web3 object
- implement a version for erc20 tokens SNT and STT that provides enriches
result of get-transaction with `:recipient` and `:snt-amount`

### DIFF
--- a/src/status_im/utils/ethereum/core.cljs
+++ b/src/status_im/utils/ethereum/core.cljs
@@ -3,6 +3,7 @@
             [status-im.js-dependencies :as dependencies]
             [status-im.native-module.core :as status]
             [status-im.utils.ethereum.tokens :as tokens]
+            [status-im.utils.ethereum.abi-spec :as abi-spec]
             [status-im.utils.money :as money]
             [taoensso.timbre :as log]))
 
@@ -132,7 +133,7 @@
                                  :method  "eth_call"
                                  :params  [params "latest"]}))
    (fn [response]
-     (if (= "" response)
+     (if (string/blank? response)
        (log/warn :web3-response-error)
        (callback (get (js->clj (.parse js/JSON response)) "result"))))))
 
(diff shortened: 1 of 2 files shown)


src/status_im/utils/ethereum/abi_spec.cljs
```diff
@@ -41,7 +41,7 @@
 
 (defn hex-to-number [x]
   (when x
-    (.toNumber (dependencies/Web3.prototype.toBigNumber (str "0x" x) 16))))
+    (dependencies/Web3.prototype.toBigNumber (str "0x" x) 16)))
```
flexsurfer: `.toBigNumber` has only one parameter, so 16 is just confusing and better to remove it
how does it work without `.toNumber`?

src/status_im/utils/ethereum/core.cljs
```diff
@@ -175,11 +176,31 @@
                                      (cb (js->clj result :keywordize-keys true))
                                      (handle-error error)))))
 
-(defn get-transaction [web3 number cb]
-  (.getTransaction (.-eth web3) number (fn [error result]
-                                         (if-not error
-                                           (cb (js->clj result :keywordize-keys true))
-                                           (handle-error error)))))
+#_(defn get-transaction [web3 number cb]
```
flexsurfer: remove?

flexsurfer (approved):

src/status_im/utils/ethereum/abi_spec.cljs
flexsurfer: `.toBigNumber` has only one parameter, so better to remove 16 to do not confuse

src/status_im/utils/ethereum/core.cljs
```diff
@@ -175,11 +176,25 @@
...
+  (first (abi-spec/decode hex ["uint"])))
+
+(defn get-transaction [transaction-hash callback]
+  (status/call-private-rpc
+   (.stringify js/JSON (clj->js {:jsonprc "2.0"
+                                 :id      1
+                                 :method  "eth_getTransactionByHash"
+                                 :params  [transaction-hash]}))
+   (fn [response]
+     (if (= "" response)
```
vitvly: Can we use string/blank? instead?

vitvly (changes requested):

vitvly (approved):

oskarth:
@yenda what's the state of this PR?

yenda:
It doesn't need to be tested just e2e tests as there is no changes to previous code

churik:
@yenda e2e cannot be built:
```
The triggering condition was at timestamp=1556005729: inotify-add-watch(/home/jenkins/workspace/us-react_prs_android-e2e_PR-7551/node_modules/unset-value/node_modules/has-value) -> The user limit on the total number of inotify watches was reached; increase the fs.inotify.max_user_watches sysctl
09:48:54  All requests will continue to fail with this message until you resolve
09:48:54  the underlying problem.  You will find more information on fixing this at
09:48:54  https://facebook.github.io/watchman/docs/troubleshooting.html#poison-inotify-add-watch. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.html.
```
Infrastructure error?

PR #7919, opened 2019-04-05
Fetch missed range of messages

rasom:
The goal of this PR is to show to a user that some messages might be out of range of the last request to mailserver. In this case user can request to "fill that gap" and receive missed messages (if such exist).
- only the last "gap" is shown in UI, the user will not be informed about previous gaps in requests
- the range of the gap is not limited at the moment, i.e. if user missed 5 days of history they will be proposed to request the whole range of messages 

testing notes:
- only 60s of history is fetched by default in the latest build of this PR; it will allow to test functionality without waiting 24h
- if 24h of history is necessary for testing it can be loaded via "Fetch last 24 hours" option in chat's menu

status: ready

### DIFF
--- a/src/status_im/chat/db.cljs
+++ b/src/status_im/chat/db.cljs
@@ -5,7 +5,8 @@
             [status-im.chat.commands.core :as commands]
             [status-im.chat.commands.input :as commands.input]
             [status-im.group-chats.db :as group-chats.db]
-            [status-im.utils.gfycat.core :as gfycat]))
+            [status-im.utils.gfycat.core :as gfycat]
+            [status-im.transport.partitioned-topic :as topic]))
 
 (defn group-chat-name
   [{:keys [public? name]}]
@@ -49,11 +50,28 @@
           {}
           chats))
 
+(defn topic-by-current-chat
+  [{:keys [current-chat-id chats] :as db}]
+  (let [{:keys [public?]} (get chats current-chat-id)
+        public-key (get-in db [:account/account :public-key])]
+    (if public?
+      (get-in db [:transport/chats current-chat-id :topic])
+      (topic/public-key->discovery-topic-hash public-key))))
+
+(defn messages-gap
+  [mailserver-topics topic]
+  (let [{:keys [gap-from gap-to]}
+        (get mailserver-topics topic)]
+    {:from    gap-from
+     :to      gap-to
+     :exists? (and gap-from gap-to
+                   (> gap-to gap-from))}))
+
 (defn sort-message-groups
   "Sorts message groups according to timestamp of first message in group"
   [message-groups messages]
   (sort-by
-   (comp unchecked-negate :timestamp (partial get messages) :message-id first second)
+   (comp :timestamp (partial get messages) :message-id first second)
    message-groups))
 
 (defn quoted-message-data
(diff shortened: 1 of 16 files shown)


src/status_im/mailserver/core.cljs
```diff
@@ -37,6 +35,7 @@
 
 (def one-day (* 24 3600))
 (def seven-days (* 7 one-day))
+(def max-request-range 60 #_one-day)
```
rasom: for testing purposes; must be fixed before merging

rasom:
@errorists 
iOS:


Android:

src/status_im/mailserver/core.cljs
```diff
@@ -485,22 +510,101 @@
...
 
-(defn get-updated-mailserver-topics [db topics last-request]
-  (reduce (fn [acc topic]
-            (if-let [mailserver-topic (some-> (get-in db [:mailserver/topics topic])
-                                              (assoc :last-request last-request))]
-              (assoc acc topic mailserver-topic)
-              acc))
-          {}
-          topics))
+(defn recalculate-gap
```
cammellos: Probably it would be good to add some tests for this function, although +1 for the drawings :)

src/status_im/chat/db.cljs
```diff
@@ -64,27 +73,86 @@
...
-                           (cond-> (-> message
-                                       (update :content dissoc :response-to :response-to-v2)
-                                       (assoc :datemark      datemark
-                                              :timestamp-str timestamp-str
-                                              :user-statuses (get message-statuses message-id)))
-                             quote ;; quoted message reference
-                             (assoc-in [:content :response-to] quote)))))
-                  message-references))
-          message-groups))
+  [message-groups messages message-statuses referenced-messages messages-gap]
```
cammellos, line 137: I believe we have some tests already for this, they should probably be updated to reflect the changes and test the new functionalities

cammellos (changes requested):
Code looks ok, but probably best to add some tests to check the most complex logic (calculating the gap & inserting the gaps in the message stream)

errorists (approved):

src/status_im/mailserver/core.cljs
rasom: sure, on it

src/status_im/chat/db.cljs
rasom: ok

rasom: there are no tests btw

src/status_im/mailserver/core.cljs
rasom: done

src/status_im/chat/db.cljs
rasom: added

rasom:
@cammellos tests are added

cammellos (approved):

churik:
#### 1) "Fetch messages" is shown even there are no messages to fetch - if you close app more than for default fetching time (in case of real app it will be 24 h)

Don't know if it is possible but it would be nice to show "Fetch messages" if there are messages to fetch.

**Steps:**
- close app for > 1 min
- reopen app and check all public chat

Even if no new messages are sent, "Fetch messages" is shown.

#### 2) "Fetch messages" is always placed according to timestamp  regardless of message ordering

If several messages are received and several are included to "gap" in 1 minute - "Fetch messages" will be always displayed on the top of messages with current timestamp regardless of order.

**Steps:**
- send message "new 3" to public chat from another device 
- close app
- send several more messages in the same minute
- wait 1 min
- open app

**Expected result:** "Fetch messages" below "new3"
**Actual result:**  "Fetch messages" is under "new3"  
dfsdfs

#### 3) "Fetch messages" is not working for 1-1 and group chats

**Steps:**
- close app 
- send some messages to 1-1, group chat from other device
- wait for 1 min
- open app 

**Expected result:** "Fetch messages" in 1-1, group chat
**Actual result:** messages are lost

#### 4) Endless spinner "Fetch messages"  if user has unstable connection while fetching

**Steps:**
- close app 
- send some messages to public chat from other device
- wait for 1 min
- open app 
- go to airplane mode
- tap on  "Fetch messages" 
- go to online

**Expected result:** Messages are fetched
**Actual result:** endless spinner 
photo_2019-04-10 16 29 09

#### 5) No "Fetch messages" for desktop app

Is it supposed only as mobile feature for now?
It is extremely useful for desktop app IMO.
#### 6) Question: there is no way somehow emulate messages fetching for example for #status channel for ~1 week or month?

I tried to send messages to past (i.e. to 4.04, https://github.com/status-im/status-react/issues/6946) - but there is no "gap" in this case.
I can, for example, generate several thousands of messages, but it is not the same if I will do this within one hour versus fetching for several days.

rasom:
1) there is no way to know this, that's by design. For instance if you open app 2 days after previous request and there were no messages during last 24h there is no way to determine if messages were sent in the 24h-48h range. That's what this "Fetch messages" button is for.
2) Yes, because we request history by timestamp. Not an issue, specifically for bigger range like 24h, it will be not that noticeable if appear at all.
3) checking
4) checking
5) well it is useful, but I prefer to add a separate PR for this. It wasn't my objective from the beginning and PR is quite big already.
6) there is no way currently. If you want I can add an option (another button as "Fetch last 24 hours") to load 48h-72h range or something, just specify what exactly you need

rasom:
3 should be fixed now

rasom:
4 should be fixed too

rasom:
@churik 2 is probably partially fixed, 3, 4 are fixed, also added fetching of 48h-72h range as mentioned before. Default interval was changed back to 24h. Let me know if anything should be changed/added for testing

churik:
Thank you! @rasom can you please do default interval again 60s?

churik:
@rasom 
#### 2) Now "Fetch messages" always is shown in the bottom

**Steps:**
- send message to public chat from another device ("A")
- close app
- send several more messages ("B", "C")
- wait 1 min
- open app
- send message to public chat from another device ("D")

**Expected result:**  "Fetch messages" between "A", "D"
**Actual result:** "Fetch messages" in the bottom
Monosnap 2019-04-11 15-59-40

#### 4) Still can reproduce with same steps (tried on IOS)
dfsfdsdf

#### 7) Should we show "Fetch messages" before joining group chat? (minor one)
Now it is shown even if the user didn't join group chat yet
sdfasdf

rasom:
2) @churik the same for 24h?

rasom:
`7` I'll take a look, but that's mostly because of 60s interval

churik:
@rasom 2. For 24h - If I manually fetch history for 48-72h, "Fetch messages" is shown below "Today". In fact, messages are fetched for yesterday and part of day before yesterday (for 09-10.04), so it should be under today.
Monosnap 2019-04-11 18-13-30

rasom:
I tested `4` on android, and it is fixed there... checking iOS version

rasom:
re 2:
what I mean is that
> Now "Fetch messages" always is shown in the bottom

is not exactly what happens if we have 24h default range. But with 60s range it easily can be the case.

churik:
@rasom yes, it is not exactly the case for 24h, agree

rasom:
2 (24h case) is fixed, 4 fixed on iOS too

annadanchenko:
> only the last "gap" is shown in UI, the user will not be informed about previous gaps in requests

@rasom is there any short-term plan to fix this? User will have history gaps that are not identified in UI (thus assuming there is issue with delivering messages) if opening the app but not fetching messages for the gap and then opening the app again later... At least, this how I understood expected behaviour in this PR.

rasom:
> is there any short-term plan to fix this?

well, I'm still looking into this. It's definitely doable but will require a bit more coding than this PR (which is quite big already).

mandrigin:
maybe we can create a separate issue for a separate PR and merge this one?

rasom:
@mandrigin regardless of issue I planned to work on it during the weekend.

1. We can actually create a separate entity and store all `gaps` for a specific `chat-id`-`topic` pair (the `topic` might be the same for different chats and requests which were executed in the past without listening to specific chat are irrelevant, so we can't just save them by topic). Then we can show multiple gaps in the chat if necessary
2. We can indefinitely request history from the past (like an endless "gap") and this functionality also requires some separate efforts separate from these "multiple gaps"

So yes, I would prefer to merge this PR. It only helps with the latest gap, which is probably the most annoying one for the user. Another PR which allows to fetch 7d history for pub chats will probably solve this issue for pub chats, so it will be mostly relevant to 1-1 and private groups.

churik:
@rasom 
there is still a way to reproduce endless spinner for "Fetch messages"
**Steps:**
- when "Fetch messages" is shown, tap on it
- go to offline
- go to online
**Video:** https://take.ms/7mXPk

WDYT, better to fix here or separately?
All other issues are not reproducible for me.

And I think that 'fetch 48-72h' should be removed before merge, right?

rasom:
> WDYT, better to fix here or separately?

probably it will be a bit simpler to file separate PR, yes

> And I think that 'fetch 48-72h' should be removed before merge, right?

yes, on it

rasom:
@churik updated

churik:
Tested (on builds with default interval = 60 sec):
- gaps in 1-1/public/group chats > 10 mins;
- retrieving chat history for 10 intervals with > 500 messages (including messages with > 10000 chars, emojis, /gfycat);
- "fetch messages" is shown after app was offline or closed > 1 interval;
- sending messages after "fetch messages" is shown;
- "fetch messages" is blocked while offline;
- "fetch messages" at the end of chat history (https://github.com/status-im/status-react/pull/7919#issuecomment-482173266);

Tested (on builds with default interval = 24h):
- "fetch messages" inside of chat history;
- "fetch messages" when offline -> online;
- "fetch messages" on custom mailserver (from https://gist.github.com/jacqueswww/3ddfc5ecfa09ecea702744ab15a2200d);

https://github.com/status-im/status-react/pull/7919#issuecomment-482526381 will be reported separately.

Checked that  'fetch 48-72h' is removed in last build.

@annadanchenko @rasom did I miss something?

rasom:
@churik nope

annadanchenko:
> @annadanchenko @rasom did I miss something?

@churik thank you for summary! I don't see any extra tests to be done at the moment.

PR #7923, opened 2019-04-05
Implicit make shell

pedropombeiro:

### Summary:

This PR aims to simplify things for developers by getting rid of the pre-requisite of starting a Nix shell. The makefile is now smart enough to detect if a Nix shell is needed and start one if required.

This PR also adds a `desktop-server` target to make it easier to start a ubuntu-server in the context of Nix, along with the default variables needed to run a debug Desktop app alongside with the release one.

The PR depends on the nix-status-go PR, so it currently includes its commits (will rebase once that one is merged).

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean setup react-native test release _ensure-in-nix-shell _list
+.PHONY: clean clean-nix react-native test release _list
 
 help: ##@other Show this help
 	@perl -e '$(HELP_FUN)' $(MAKEFILE_LIST)
@@ -28,35 +28,30 @@ HELP_FUN = \
 
 export NIX_CONF_DIR = $(PWD)/nix
 
+export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
+export STATUS_NODE_PORT ?= 12345 # no need to specify this if just running dev instance alongside release build
+export STATUS_DATA_DIR ?= ~/status-files/data1 # this is where Realm data files, Geth node data, and logs will reside; also not strictly needed for dev alongside release
+
+# WARNING: This has to be located right before the targets
+ifdef IN_NIX_SHELL
+SHELL := env bash
+else
+SHELL := ./nix/shell.sh
+endif
+
 # Main targets
 
+clean: SHELL := /bin/sh
 clean: ##@prepare Remove all output folders
 	git clean -dxf -f
 
+clean-nix: SHELL := /bin/sh
 clean-nix: ##@prepare Remove complete nix setup
 	sudo rm -rf /nix ~/.cache/nix
 
-setup: ##@prepare Install all the requirements for status-react
-	@./scripts/setup
-
-_ensure-in-nix-shell:
-ifndef IN_NIX_SHELL
-	$(error '$(RED)Please run '$(BOLD)$(MAKE) shell$(RESET)$(RED)' first$(RESET)')
-endif
-
 shell: ##@prepare Enter into a pre-configured shell
 ifndef IN_NIX_SHELL
-	@if command -v "nix" >/dev/null 2>&1 || [ -f ~/.nix-profile/etc/profile.d/nix.sh ]; then \
-		echo "Configuring Nix shell..."; \
-		if ! command -v "nix" >/dev/null 2>&1; then \
-			. ~/.nix-profile/etc/profile.d/nix.sh; \
-				nix-shell; \
-		else \
-			nix-shell; \
-		fi \
-	else \
-		echo "Please run 'make setup' first"; \
-	fi
+	@ENTER_NIX_SHELL
 else
 	@echo "Nix shell is already active"
 endif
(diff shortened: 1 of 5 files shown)


jakubgs:
This repeating of `_run-in-nix-shell` is absolutely abysmal. Here's a better solution:

In the `Makefile` right before the first target you add:
```Makefile
# WARNING: This has to be located right before the targets
ifdef IN_NIX_SHELL
SHELL := env bash
else
SHELL := ./nix/shell
endif
```
You change the `shell` target to be:
```Makefile
shell: ##@prepare Enter into a pre-configured shell
ifndef IN_NIX_SHELL
	@ENTER_NIX_SHELL
else
	@echo "Nix shell is already active"
endif
```
And add a `nix/shell` file:
```bash
#!/usr/bin/env bash
export TERM=xterm # fix for colors
shift # we remove the first -c from arguments
. ~/.nix-profile/etc/profile.d/nix.sh
if [[ $@ == "ENTER_NIX_SHELL" ]]; then
    exec nix-shell
else
    exec nix-shell --run "$@"
fi
```
And that's it.
This way we can drop all the other `ifndef IN_NIX_SHELL` as well as the `_run-in-nix-shell` target completely.

jakubgs:
Actually, the `setup` target might also require this before it:
```Makefile
setup: SHELL := /bin/sh
```
So it doesn't use `nix/shell`.

pedropombeiro:
This PR now goes one step further and does away with `make setup` entirely. Once you clone the repo, you can do something like `make release-android` directly and it will install Nix, download all dependencies and build/deploy the Android app without any intervention.

ci/ios.groovy
```diff
@@ -1,7 +1,9 @@
 utils = load('ci/utils.groovy')
 
 def plutil(name, value) {
-  utils.nix_sh "plutil -replace ${name} -string ${value} ios/StatusIm/Info.plist"
+  return """
+  plutil -replace ${name} -string ${value} ios/StatusIm/Info.plist
```
jakubgs: I think you need a semilcolon for this to work.

nix/shell.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+export TERM=xterm # fix for colors
```
jakubgs, line 6: I'm not entirely sure hardcoding this is a good idea, but we'll see in CI output if it is.

Also, I kinda liked `nix/shell`, adding the `.sh` makes it clunky :D.

Makefile
```diff
@@ -26,36 +26,30 @@ HELP_FUN = \
...
-__toolversion = $(shell $(GIT_ROOT)/scripts/toolversion $(1))
+export NIX_CONF_DIR = $(PWD)/nix
+
+export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
+export STATUS_NODE_PORT ?= 12345 # no need to specify this if just running dev instance alongside release build
+export STATUS_DATA_DIR ?= ~/status-files/data1 # this is where Realm data files, Geth node data, and logs will reside; also not strictly needed for dev alongside release
+
+# WARNING: This has to be located right before the targets
+ifdef IN_NIX_SHELL
+SHELL := env bash
```
jakubgs, line 37: By default `make` uses `/bin/sh`, here we make it use `bash` from Nix. I wonder if it makes a difference anywhere. I guess it's fine.

Makefile
```diff
@@ -197,14 +178,14 @@ _unknown-startdev-target-%:
...
 	$(eval SYSTEM := $(word 2, $(subst -, , $@)))
 	$(eval DEVICE := $(word 3, $(subst -, , $@)))
 	scripts/prepare-for-platform.sh ${SYSTEM} || $(MAKE) _unknown-startdev-target-$@
 	@ if [ -z "$(DEVICE)" ]; then \
-		$(MAKE) watch-$(SYSTEM) || $(MAKE) _unknown-startdev-target-$@; \
-	else \
-		$(MAKE) watch-$(SYSTEM)-$(DEVICE) || $(MAKE) _unknown-startdev-target-$@; \
+			$(MAKE) watch-$(SYSTEM) || $(MAKE) _unknown-startdev-target-$@; \
+		else \
+			$(MAKE) watch-$(SYSTEM)-$(DEVICE) || $(MAKE) _unknown-startdev-target-$@; \
```
jakubgs: This indentation looks wrong.

fastlane/Fastfile
```diff
@@ -130,7 +130,8 @@ def build_ios_e2e
         destination: "generic/platform=iOS Simulator",
         # 2. fixing compilations issues as stated in https://stackoverflow.com/a/20505258
         #    it looks like i386 isn't supported by React Native
-        xcargs: "ARCHS=\"x86_64\" ONLY_ACTIVE_ARCH=NO",
+        # -UseModernBuildSystem=N is a temporary fix for Xcode 10.1
+        xcargs: "ARCHS=\"x86_64\" ONLY_ACTIVE_ARCH=NO -UseModernBuildSystem=N",
```
jakubgs: Actually, we could move `UseModernBuildSystem=N` to `export_options` like here:
https://github.com/status-im/status-react/blob/22bd2f1761deba99e7c5b2d47b4345474bcde1e4/fastlane/Fastfile#L216-L222

jakubgs (approved):
Very nice.

ci/ios.groovy
pedropombeiro: Indeed, thanks for spotting it!

nix/shell.sh
pedropombeiro: TBH I only added it because it makes VS Code recognize the file type and do the right syntax highlighting :-)

jakubgs: Ah, weird it doesn't recognize it from the shebang.

fastlane/Fastfile
pedropombeiro: That section of the file actually has it in both places, should I do the same?

jakubgs: Yes. It's redundant. And `xcargs` is more of a hack.

cammellos (approved):

Makefile
```diff
@@ -66,51 +61,47 @@ endif
 #----------------
 release: release-android release-ios ##@build build release for Android and iOS
 
-release-android: prod-build-android ##@build build release for Android
-ifdef IN_NIX_SHELL
+release-android: ##@build build release for Android
+	@$(MAKE) prod-build-android && \
```
jakubgs, line 65: I don't get the point of this `@$(MAKE)`. Why are you adding this?

pedropombeiro: oh, this should have been reverted to the old way. Thanks for catching it.

pedropombeiro: I recall why it's done this way. If it is done the old way, it is going to instantiate the nix-shell twice, taking much more time. So by doing this we're optimizing so that a single nix-shell is opened. Make sense?

jakubgs: No, the reverse is happening. You are running `make ${target}`, then the target runs another `make` which creates two nested nix shells.

pedropombeiro: the second make will know it is already inside a nix-shell though, right?

jakubgs: Ah yes, that's true, but also redundant.

pedropombeiro: Why is it redundant? It's what makes this approach work, and not the previous one.

jakubgs: Because we are already changing the shell in the Makefile to be `nix-shell` using `nix/shell`, why would we need to run `make` again?

pedropombeiro:
Will merge this PR without e2e tests since this does not touch anything in the app itself.

Makefile
pedropombeiro: I'm probably missing the alternative. To me, the alternative would be to duplicate the `prod-build-android` target inside the `release-android` target, which is not desireable.

jakubgs: But what is wrong with adding it as a dependency of that target?

PR #7925, opened 2019-04-06
Fix 'nix build' command by separating out into a shell.nix file

pedropombeiro:

### Summary:

This PR separates out shell-specific dependencies into a shell.nix file, and makes default.nix call a derivation.nix which contains all our project dependencies. This allows running `nix build` instead of nix-shell for building our Nix cache.

### Review notes (optional):

This PR is dependent on `feature/nix-status-go`, so it includes its commits. Please review only the last commit.

### Testing notes (optional):

This is a build-only esthetic change, shouldn't impact app functionality, so the automated tests should be enough to test.

status: ready

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,103 +1,19 @@
-# target-os = [ 'windows' 'linux' 'macos' 'android' 'ios' ]
-{ pkgs ? import ((import <nixpkgs> { }).fetchFromGitHub {
+# target-os = [ 'windows' 'linux' 'macos' 'android' 'ios' 'all' ]
+{ config ? { android_sdk.accept_license = true; },
+  pkgs ? import ((import <nixpkgs> { }).fetchFromGitHub {
     owner = "status-im";
     repo = "nixpkgs";
     rev = "db492b61572251c2866f6b5e6e94e9d70e7d3021";
     sha256 = "188r7gbcrxi20nj6xh9bmdf3lbjwb94v9s0wpacl7q39g1fca66h";
-  }) { config = { android_sdk.accept_license = true; }; },
-  target-os ? "" }:
+  }) { inherit config; },
+  stdenv ? pkgs.stdenv,
+  target-os ? "all" }:
 
-with pkgs;
-  let
-    targetDesktop = {
-      "linux" = true;
-      "windows" = true;
-      "macos" = true;
-      "" = true;
-    }.${target-os} or false;
-    targetMobile = {
-      "android" = true;
-      "ios" = true;
-      "" = true;
-    }.${target-os} or false;
-    # TODO: Try to use stdenv for iOS. The problem is with building iOS as the build is trying to pass parameters to Apple's ld that are meant for GNU's ld (e.g. -dynamiclib)
-    _stdenv = if target-os == "ios" || target-os == "" then stdenvNoCC else stdenv;
-    statusDesktop = callPackage ./nix/desktop { inherit target-os; stdenv = _stdenv; };
-    statusMobile = callPackage ./nix/mobile { inherit target-os status-go; androidPkgs = androidComposition; stdenv = _stdenv; };
-    status-go = callPackage ./nix/status-go { inherit (xcodeenv) composeXcodeWrapper; inherit xcodewrapperArgs; androidPkgs = androidComposition; };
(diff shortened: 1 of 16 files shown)


derivation.nix
```diff
@@ -0,0 +1,83 @@
+# target-os = [ 'windows' 'linux' 'macos' 'android' 'ios' ]
+{ system ? builtins.currentSystem
+, config ? {}, overlays ? []
+, pkgs ? (import <nixpkgs> { inherit system config overlays; })
+, target-os ? "" }:
+
+with pkgs;
+  let
+    targetDesktop = {
```
jakubgs, line 13: Wouldn't it make sense to add `desktop` as a valid value here?

shell.nix
```diff
@@ -0,0 +1,41 @@
+{ pkgs ? import <nixpkgs> {},
+  target-os ? "" }:
+with pkgs;
+
+let
+  projectDeps = import ./default.nix { inherit target-os; };
+  targetMobile = {
```
jakubgs, line 7: Same here, we should add `mobile` as an option. Maybe even `all` to both too for explicit building of everything?

jakubgs (approved):

derivation.nix
pedropombeiro: Not sure what would we gain by adding that extra option?

shell.nix
pedropombeiro: We could use all instead of "", yeah.

derivation.nix
jakubgs: We would gain the ability to specify `target-os` as `mobile` instead of any specific target if in that context it makes sense. I'm trying to cover a potential use, but it's also just aesthetics. If the flag is called `targetDesktop`, it should be triggerable via `desktop` option, and so on. It's just intuitive in my opinion. But this is just a tiny nitpick.

shell.nix
jakubgs: Yeah, explicitly instead of implicitly. But we should keep `""` too.

jakubgs:
Lel:
```
[2019-04-09T14:17:21.366Z] \033[0;31mConan package manager not found. Exiting...\033[0m
```

shell.nix
pedropombeiro: We don't really need `""`, since that's a fallback value we control.

default.nix
```diff
@@ -1,103 +1,16 @@
...
-      ''
-        if [ -n "$ANDROID_SDK_ROOT" ] && [ ! -d "$ANDROID_SDK_ROOT" ]; then
-          ./scripts/setup # we assume that if the Android SDK dir does not exist, setup script needs to be run
-        fi
-        set +e
-      '';
-    hardeningDisable = status-go.hardeningDisable;
-  }
+(pkgs.callPackage ./derivation.nix { }).overrideAttrs(_: { src = null;
+                                                           # TODO: Figure out if there's a better way to do this
```
jakubgs: Why is this part indented to weirdly?

mandrigin (approved):
lgtm

nix/desktop/default.nix
```diff
@@ -6,15 +6,15 @@ with stdenv;
...
   targetDarwin = {
     "macos" = true;
-    "" = isDarwin;
+    "all" = isDarwin;
   }.${target-os} or false;
   targetWindows = {
     "windows" = true;
-    "" = isLinux;
+    "all" = isLinux;
   }.${target-os} or false;
```
jakubgs, line 18: Are we sure we don't want to keep the implicit `""` empty value to build everything?

pedropombeiro: Yeah, there's no benefit in that. We control all values that are passed, so we can decide to be explicit. What we could add is logic to throw an exception if the value is unknown.

jakubgs: Oh, that would be better!

pedropombeiro: OK, added logic so that we get an error if an unknown value is passed in.

jakubgs (approved):
Perfection.

default.nix
pedropombeiro: haha, just had just fixed it like 5 minutes ago.

PR #7957, opened 2019-04-11
Copy Android SDK from nix store instead of downloading

pedropombeiro:

### Summary

This PR gets rid of the `install_android_sdk`/`use_android_sdk` functions in installers.sh and replaces them with logic in `android.nix` to copy the nixpkgs cached Android SDK to the usual location, meaning that we now have a single source of truth for the Android SDK component versions we use, and a single download.
It also adds logic to remove the temp Android SDK folder to the `clean-nix` make target. The next time that a nix command is executed, this folder will be populated from the nix store again.

### Testing notes

Manual testing not needed, the automated will be enough to verify that the Android app still runs.

#### Platforms

- Android

status: ready


ci/Jenkinsfile.android
```diff
@@ -2,7 +2,7 @@ pipeline {
   agent { 
     docker {
       label 'linux'
-      image 'statusteam/nix:jenkins-1.0.0-150d5721'
+      image 'statusteam/nix:jenkins-1.0.0-e20e5e15'
```
mandrigin: these are a bit hard to track... can we maybe bump versions, like `1.0.1`, etc?

mandrigin (approved):

ci/Jenkinsfile.android
jakubgs: Hopefully soon this won't be necessary. I want to make a docker image that will have part of its store as a volume on the host. So we start with fresh image very time, but don't have to re-download from cache every time.

pedropombeiro: OK, we can start doing that from now on. For this PR it won't be beneficial since the image is already out there.

mandrigin: yeah, I agree

mandrigin:
@jakubgs macos-02 `[2019-04-12T06:46:14.189Z] error: opening file '/nix/store/pl98rh8chfsc5a5bl23v6vnj8ks0jrfp-gomobile-20190319-167ebed.drv': No such file or directory` :(

Makefile
```diff
@@ -47,7 +47,7 @@ clean: ##@prepare Remove all output folders
 
 clean-nix: SHELL := /bin/sh
 clean-nix: ##@prepare Remove complete nix setup
-	sudo rm -rf /nix ~/.cache/nix
```
jakubgs, line 51: If we are purging nix completely, should't we also remove `~/.nix-profile`, `~/.nix-defexpr`, and `~/.nix-channels`
?

scripts/setup
```diff
@@ -46,8 +46,7 @@ fi
 ####
 setup_header "Installing requirements..."
 
-install_nix && \
```
jakubgs, line 49: At this point we might as well get rid of `scripts/lib/setup/installers.sh` and just put the contents of its `install_nix` method here. Make it simpler. Less files is better.

nix/mobile/android.nix
```diff
@@ -0,0 +1,59 @@
...
+      export JAVA_HOME="${openjdk}"
+      export ANDROID_HOME=~/.status/Android/Sdk
+      export ANDROID_SDK_ROOT="$ANDROID_HOME"
+      export ANDROID_NDK_ROOT="${androidComposition.androidsdk}/libexec/android-sdk/ndk-bundle"
+      export ANDROID_NDK_HOME="$ANDROID_NDK_ROOT"
+      export ANDROID_NDK="$ANDROID_NDK_ROOT"
+      export PATH="$ANDROID_HOME/bin/$ANDROID_HOME/tools:$ANDROID_HOME/tools/bin:$ANDROID_HOME/platform-tools:$ANDROID_HOME/build-tools:$PATH"
+
+      if ! [ -d $ANDROID_HOME ]; then
+        echo "=> pulling the Android SDK out of the nix store and into a writeable directory"
```
jakubgs, line 46: Maybe a comment or message explaining why we do this? It's not clear to someone who doesn't know the history as to why we need the directory to be writeable.

jakubgs (approved):
The diff of `nix/status-go/default.nix` is really fucked up. GitHub can't seem to be able to properly diff that.

scripts/setup
pedropombeiro: Done.

Makefile
pedropombeiro: yes!

pedropombeiro:
> The diff of `nix/status-go/default.nix` is really fucked up. GitHub can't seem to be able to properly diff that.

You need to use better tools for the job. I normally check diffs with GitExtensions and BeyondCompare:

image

So actually not much changed there.

jakubgs:
>So actually not much changed there.

Yeah, I know. Poor GitHub can't handle `.nix` files right.

PR #7958, opened 2019-04-11
modify .env settings using Jenkins params

jakubgs:
Now we can specify ANY setting that exists in the `.env` file and modify it before build. It just has to be added as parameter in Jenkins job configuration.

This was done mostly so testers can make manual builds with changed `FLEET`.

Successful manual build with `eth.staging` passed as `FLEET` parameter:
https://ci.status.im/job/status-react/job/manual/269/

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -80,7 +80,7 @@ pipeline {
     }
     stage('Bundle') {
       steps {
-        script { apk = mobile.android.bundle(btype) }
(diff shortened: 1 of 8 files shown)


mandrigin (approved):

pedropombeiro (approved):

mandrigin:
@jakubgs does it need to be tested or successful build is enough?

jakubgs:
I don't think it needs to be tested. It's only changed to CI setup.

PR #7961, opened 2019-04-12
Load 7 days history by default from non-discovery topics

rasom:
In most cases it is acceptable to load 7 days (instead of 1) of public chat history.
Discovery topic is still a bit "overloaded", that's why only 1 day is requested.  

status: ready

### DIFF
--- a/src/status_im/data_store/realm/schemas/account/mailserver_requests_gap.cljs
+++ b/src/status_im/data_store/realm/schemas/account/mailserver_requests_gap.cljs
@@ -1,0 +1,8 @@
+(ns status-im.data-store.realm.schemas.account.mailserver-requests-gap)
+
+(def v1 {:name       :mailserver-requests-gap
+         :properties {:chat-id {:type    :bool
+                                :indexed true}
+                      :from    {:type    :integer
+                                :indexed true}
+                      :to      :integer}})
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -724,7 +724,7 @@
  :chat.ui/fetch-history-pressed
  (fn [{:keys [now] :as cofx} [_ chat-id]]
    (mailserver/fetch-history cofx chat-id
-                             {:from (- (quot now 1000) (* 24 3600))})))
+                             {:from (- (quot now 1000) mailserver/one-day)})))
 
 (handlers/register-handler-fx
  :chat.ui/fill-the-gap
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -39,7 +39,7 @@
 (def max-discovery-request-range one-day)
 (def maximum-number-of-attempts 2)
 (def request-timeout 30)
-(def min-limit 100)
+(def min-limit 50)
 (def max-limit 2000)
 (def backoff-interval-ms 3000)
 (def default-limit max-limit)
@@ -364,10 +364,11 @@
           {:topic     topic
            :from      from
            :to        to
+           :limit     (when-not discovery? 100)
            :force-to? (not (nil? force-request-to))})))))
 
 (defn aggregate-requests
-  [acc {:keys [topic from to force-to?]}]
+  [acc {:keys [topic from to force-to? limit]}]
   (update acc [from to force-to?]
           (fn [{:keys [topics]}]
             {:topics    ((fnil conj #{}) topics topic)
@@ -375,7 +376,8 @@
              :to        to
              ;; To is sent to the mailserver only when force-to? is true,
              ;; also we use to calculate when the last-request was sent.
-             :force-to? force-to?})))
+             :force-to? force-to?
+             :limit     limit})))
 
 (defn prepare-messages-requests
   [{{:keys [:mailserver/requests-from
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -35,7 +35,8 @@
 
 (def one-day (* 24 3600))
 (def seven-days (* 7 one-day))
-(def max-request-range one-day)
+(def max-request-range seven-days)
+(def max-discovery-request-range one-day)
 (def maximum-number-of-attempts 2)
 (def request-timeout 30)
 (def min-limit 100)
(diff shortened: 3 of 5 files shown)


mandrigin (approved):

cammellos (approved):

jakubgs:
@annadanchenko just noticed this spike in Mailserver response latency:
mail01gc_proc_time_spike
Is `mail-01.gc-us-central1-a.eth.beta` the one you're using?

churik:
@rasom 
1) Question: for now you say that history is loading for 7 days only from non-discovery topics.
In my world it is for 1-to-1 and group chat messages (https://hackmd.io/HqkkDjccS3KnD65lvX_tXQ - where we use single topics)
But in PR build I can see that history is loaded for all types of chats.
Sorry if I missed something, but can you elaborate what it is for end user?

2) When I restore existing account with chats (eventually restored three 1-1 chats, so account is not overloaded), fetching messages lasts at least several minutes. The results are here:

  | Current nightly | PR | Mailserver
-- | -- | -- | --
Fetching messages | ~30-40 sec | ~6 mins | automatic selection
Device is discoverable (New device is detected is shown on other devices) | ~10-20 sec | ~3 min | automatic selection
Fetching messages for #status channel | ~30-40 sec | ~3 min | mail-01.gc-us

When I recover account and connect to the same mailserver (in my case  `mail-01.gc-us`) on nightly and PR builds (made this check to compare behavior regardless of mailserver) - on nightly recovery lasts <1 min without any errors, on PR builds it takes > 5 mins and should several times tap on "Error: no connection to mailserver". If I recover account with "automatic selection" enabled - during fetching 4-5 maiservers are switched.

Geth.log from device  when restoring account with "automatic selection"  [geth (1).log](https://github.com/status-im/status-react/files/3079657/geth.1.log)
IMO it is too long from user POV.
WDYT?
Please let me know if you need any additional info / logs.

rasom:
@jakubgs it is not related to this PR, rather to load on discovery topic or something else

rasom:
@churik 1-1 and group chat communication is based on discovery topic at the moment

You should see the difference in pub chats, in 1-1 only last 24h should be loaded. So if you have difficulties with loading 1-1s history it is unrelated.

rasom:
@churik @jakubgs how do you determine that specifically using of this PR causes spike?

churik:
@rasom 1-1 chat is restored with last message > 48h, and it is restored constantly in PR build and it is not restored on nightly.
So I think it is related.

rasom:
@churik please don't add new points to the same comment. I don't mind editing existing ones, but it might be difficult to spot changes otherwise

churik:
@rasom - about spike - I tested loading history for #status channel on this particular mailserver (https://github.com/status-im/status-react/pull/7961#issuecomment-483200057) - so I suppose that can be the reaason. I can do more tests to investigate current behavior of different mailservers using this PR build.

mandrigin:
@rasom can you also update the copy on this screen, otherwise it tells a lie now... https://github.com/status-im/status-react/issues/7454

rasom:
> 1-1 chat is restored with last message > 48h, and it is restored constantly in PR build and it is not restored on nightly.

just restored my own account and history is definitely not loaded for > 48h (1-1 chats). If it is... it could probably be related to device synchronization, but it doesn't happen by default. Used previous build https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190413-082525-c745b6-pr.apk

rasom:
@mandrigin fixed

rasom:
@churik 
> When I restore existing account with chats (eventually restored three 1-1 chats, so account is not overloaded), fetching messages lasts at least several minutes. The results are here:

can you please retest both versions and provide `adb logcat "*:S" ReactNative:V ReactNativeJS:V` for both?

churik:
@rasom 

I compared restoring the same account on the current nightly and last PR build.

Mailserver | Nightly (12.04.19) | Log | PR | Log
-- | -- | -- | -- | --
mail02-ac-cn-hongkong | ~1,5 min | [HC-nightly.log](https://github.com/status-im/status-react/files/3080486/hongkongNightly.log)  | ~10 min |  [HC-PR.log](https://github.com/status-im/status-react/files/3080491/hongkong.log)
mail-03.gc-us-central1 | ~1 min | [US-nightly.log](https://github.com/status-im/status-react/files/3080502/usNightly.log)  | ~ 4 min | [US-PR.log](https://github.com/status-im/status-react/files/3080500/03-gc-us-PR.log)
mail-03.do-ams-3 | ~30 sec | [AMS-nightly.log](https://github.com/status-im/status-react/files/3080518/centralNightly.log)  | ~1 min | [AMS-PR.log](https://github.com/status-im/status-react/files/3080522/mail03.do-ams3.log)

rasom:
@churik was that nightly or release? from the `HC-nightly.log ` log:
> `04-15 15:52:03.736 19668 19713 I ReactNativeJS: INFO [status-im.mailserver.core:307] - mailserver: request-messages for:   topics  ("0x703bd2d4" "0xf8946aac" "0xe0dc2e7e")  from  1555249884  cursor  5cb46e19b0e5d6e7a6311acc95565b83b64081c43625287db73637906348c5feaf2ef40e  limit  2000`

currently this line also contains `force-to?` entry https://github.com/status-im/status-react/blob/a4d8f57b09ed1a48a17f5971e9556520fbb19e9a/src/status_im/mailserver/core.cljs#L306

churik:
@rasom log https://github.com/status-im/status-react/files/3080486/hongkongNightly.log from nightly 12/04/19

rasom:
ok, old nightly then

rasom:
It seems that we can't merge this PR, it will significantly worsen UX.

- fetching messages from mailservers which are not located in the same region as client is very slow (for instance, fetching from hk/us mailservers when client is located in Europe)
- fetching 7d of topic with no messages has higher chances to fail regardless of specified limit, and that's what happen when we attempt to open a blank pubchat or add a new contact (because 7d of "-contact-code" topic are fetched) 
- we already have > 15k messages in 24h discovery topic, fetching is not too slow but it could be faster  with less messages

mandrigin:
> fetching 7d of topic with no messages has higher chances to fail regardless of specified limit, and that's what happen when we attempt to open a blank pubchat or add a new contact (because 7d of "-contact-code" topic are fetched)

that is definitely something we need to dig into, that behaviour looks very suspicious

PR #7964, opened 2019-04-12
Improve makefile to only make Nix install the bare minimum for intended platform

pedropombeiro:

### Summary

This PR allows the Makefile to leverage the target-os argument to only install the packages that are required for what we're building. E.g. a new user that clones the repo on iOS and runs `make release-ios` won't see the Android SDK, NDK, gradle, etc being downloaded.

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ HELP_FUN = \
 			   }; \
 			   print "\n"; \
 		   }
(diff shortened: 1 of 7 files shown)


jakubgs (approved):
Just lovely.
ooooh

mandrigin (approved):

PR #7971, opened 2019-04-12
Build Linux with --pure

pedropombeiro:

### Summary
This PR adds support for building on Linux with `--pure` flag (more platforms to follow, in baby steps), so that we have more guarantees that we're not using something from the host environment and contaminating our build.
The nix cache builds are also run with `--pure` (Linux and macOS).

### Review notes
I've consolidated 3 environment variables which served the same purpose into `TARGET_OS`. This variable is now defined in the Makefile and also [made into an environment variable](https://github.com/status-im/status-react/pull/7971/files#diff-5204cbe23b218cae9de0b1aedd4edcd2R30
) in shell.nix. Therefore there's no longer a need to define `TARGET_SYSTEM_NAME=Windows` before invoking `build-desktop.sh`.

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ release-desktop: ##@build build release for desktop release
 release-windows-desktop: export TARGET_OS ?= windows
 release-windows-desktop: ##@build build release for desktop release
 	@$(MAKE) prod-build-desktop && \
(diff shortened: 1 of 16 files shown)


jakubgs (approved):
Less variables is always nice.

mandrigin (approved):
oh, this is huge!

Issue #7454, opened 2019-02-08
Add basic copy to public chat empty screen state


hesterbruikman:
# Problem
As a user I want to see what's going on in the Status app and what I can use it for. As soon as I join a public chat that hasn't been active in the last 24h, all I see is an empty screen. I want to understand wy this is and where I can find more active channels.

We can offer a quick solution while we wait for the epics '[Onboarding](https://www.pivotaltracker.com/epic/show/4182848)' and '[Empty state screens](https://www.pivotaltracker.com/epic/show/4182832)' to progress.

# Implementation
Add basic copy to empty public chat views
This ideally includes an update of the screen design, according to below mockup.

![public chat empty!UNITO-UNDERSCORE!bounty](https://user-images.githubusercontent.com/35233054/52554719-f9041380-2de7-11e9-8679-c80409bc624b.png)

# Acceptance Criteria
Open new topic or public chat that does not retrieve any messages from the last 24h, text message is shown on screen.

# Notes

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 40.0 DAI (40.0 USD @ $1.0/DAI)  attached to it as part of the Status-im fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/status-im/status-react/7454/2305).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/status-im/status-react/7454/2305).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $56,866.62 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [krisc](https://gitcoin.co/profile/krisc) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/status-im/status-react/7454/2305?mutate_worker_action=approve&worker=krisc) | [reject worker](https://gitcoin.co/issue/status-im/status-react/7454/2305?mutate_worker_action=reject&worker=krisc))_.**

[...]

PR #7973, opened 2019-04-13
Fix endless spinner on Fetch messages

### LINKED ISSUE #7969: Tap on "Fetch messages" leads to endless spinner if connection is lost during fetching
### Description
_Type_: Bug
_Summary_: introduced in https://github.com/status-im/status-react/pull/7919 - decision was to fix it separately. "Fetch messages" still can lead to the endless spinner when the app goes offline-online during fetching. 

#### Expected behavior
messages fetched normally

#### Actual behavior
https://user-images.githubusercontent.com/4557972/55971883-9f528500-5c82-11e9-9bd6-b3e399175304.png

### Reproduction
**Prerequisites:** "Fetch messages" should be shown (app offline for > 24h)

* Open chat with "Fetch messages"
* tap on it
* go to offline 
* go back online

### Additional Information

* Status version: nightly 13/04/2019

rasom:
fix #7969 

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -227,6 +227,13 @@
               (when-not (or sym-key-id generating-sym-key?)
                 (generate-mailserver-symkey mailserver)))))
 
+(defn executing-gap-request?
+  [{:mailserver/keys [current-request fetching-gaps-in-progress]}]
+  (= (get fetching-gaps-in-progress (:chat-id current-request))
+     (select-keys
+      current-request
+      [:from :to :force-to? :topics :chat-id])))
+
 (fx/defn connect-to-mailserver
   "Add mailserver as a peer using `::add-peer` cofx and generate sym-key when
    it doesn't exists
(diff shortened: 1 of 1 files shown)


bitsikka:
Cool!

Leaving this comment here for whatever it's worth, please don't mind me

This looks like a better approach to handle abstraction for :mailserver/fetching lifecycle in general for all different scenarios of mailserver/fetching 🤔 

I ran into a different scenario in #7454 and surely there are others

In https://github.com/status-im/status-react/pull/7663#issuecomment-473954395 @yenda said
> @bitsikka good job on spotting mailserver/fetching? inconsistencies. I didn't realize that. After discussing with @cammellos I think the best solution would be to have it rely on a value separated from pending-requests. However this shouldn't be part of this PR.

> @StatusSceptre @hesterbruikman could we make a bounty for this?
Fix mailserver/fetching? inconsistencies There should be a 1 second delay between the moment the last request gets completed and the mailserver/fetching? subscription becomes false.

In this comment for **Application Performance and UX** https://discuss.status.im/t/application-performance-and-ux/1168/3 @hesterbruikman said
> This is some painful stuff to watch. For reference here, some factors I think we should investigate:
> ...
> **Design**
> - GUI performance (loading states, empty states, user controlled fetching)
>
> ...

rasom:
@bitsikka this specific bug is not related to general fetching from mail servers, it's about fetching messages by demand when "gap" is detected. It was a leftover after #7919, and this issues appears because some state (which was added in #7919) is not cleared properly on failure.

bitsikka:
> @bitsikka this specific bug is not related to general fetching from mail servers, it's about fetching messages by demand when "gap" is detected. It was a leftover after #7919, and this issues appears because some state (which was added in #7919) is not cleared properly on failure.

understood

I'm just noting a pattern/lifecycle I think I'm noticing surrounding mailserver fetching in different scenarios and getting inspired by how you are handling it in this scenario - to perhaps use the same pattern in other scenarios

mandrigin (approved):

cammellos (approved):

churik:
@rasom can you please make builds with 60s default interval for testing? Or just say me where can I change it in code, and I'll do it by myself.

rasom:
@churik this line https://github.com/status-im/status-react/blob/9a2dfe31baaaa22e171864628368ab16bd9c6b7d/src/status_im/mailserver/core.cljs#L38

should be `(def max-request-range 60) `

churik:
Tested on IOS 11.4 and Android 8 :
- can't reproduce https://github.com/status-im/status-react/issues/7969;
- fetching messages is working as before;
- "Fetching messages" is blocked while offline;

@rasom delete or squash my commits, and thanks for clue!

Failed test is due to faucet error, so it is not introduced by PR

PR #7979, opened 2019-04-15
[#7371]Add an event to display a view fullscreen

### LINKED ISSUE #7371: Add an event to display a view fullscreen
### User Story

As a developer, I want an event allowing to display a view fullscreen.

### Description

*Type*: Feature

tbenr:
fixes #7371

test extension:
https://extensions.status.im/?hash=QmRCUoBXf8Sacw7ccK5g76wKxqzYKecfw5Y9SBkp2o8eMd

extension opens a splashscreen `:on-activation` and also on command from chat.
it tests also the new alerts "installed"\\"uninstalled" coming from the extension lifecycle itself (alerts have been removed from the installation\deinstallation process)

STATUS: done.


tbenr:
@jeluard I can't access jenkins logs, can you see why is failing?

I was thinking if `:on-open` is better than `:on-success`...

jeluard:
`on-open` is great!

Can you try to rebase your PR? Doesn't look related to your code

tbenr:
@jeluard updated and rebased. I updated extension example accordingly.

src/status_im/extensions/core.cljs
```diff
@@ -609,6 +629,16 @@
                  :arguments   {:chain-id    :number
                                :on-success? :event
                                :on-failure? :event}}
+                'show-fullscreen
```
jeluard: How about `screen/open` and `screen/close` ?

jeluard (approved):

src/status_im/extensions/core.cljs
tbenr: yeah..

tbenr:
there is a bad interaction between installation alert and `:on-activation` screen display.

image

when click OK the splashscreen is closed.

i'll address it on a dedicated PR?

jeluard:
Oh weird. Do you know why?
Actually I am wondering if the alert is needed at all.

tbenr:
> Oh weird. Do you know why? Actually I am wondering if the alert is needed at all.
> […](#)
> Le 17 avr. 2019 à 12:33, Enrico Del Fante ***@***.***> a écrit : there is a bad interaction between installation alert and :on-activation screen display. https://get.status.im/extension/ipfs@QmQWhDDCBWWJxZv9Zr6BaikuxgFxMwG8u97ZaEXuZYWZ11/ when click OK the splashscreen is closed. i'll address it on a dedicated PR? — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or mute the thread.

yes the easiest solution is to remove popup completely

jeluard:
@tbenr I think it's fine to do that

tbenr:
@jeluard i'm doing some tests but it's more complex that it seems: the challenge is to navigate back to profile **before** actually fire the on-activation event (which potentially can navigate to the extension screen).

tbenr:
@jeluard I think I found a way.. take a look :)

tbenr:
@jeluard moved alerts in the extension itself

https://extensions.status.im/?hash=QmRCUoBXf8Sacw7ccK5g76wKxqzYKecfw5Y9SBkp2o8eMd

works :-)

src/status_im/extensions/registry.cljs
```diff
@@ -143,22 +143,18 @@
...
-                                    :content    (i18n/label :t/extension-installed)
-                                    :on-dismiss #(re-frame/dispatch (if modal?
-                                                                      [:navigate-back]
-                                                                      [:navigate-to-clean :my-profile]))}}
-                (when hooks (accounts.update/account-update {:extensions new-extensions} {}))
-                (when hooks (add-to-registry url extension-data true))
-                (when on-installation (on-installation))
-                (when on-activation (on-activation))))))
+    (do
+      (re-frame/dispatch (if modal?
```
jeluard: Can't we still have regular fx usage here?

src/status_im/extensions/registry.cljs
```diff
@@ -143,22 +143,18 @@
...
+                           [:navigate-back]
+                           [:navigate-to-clean :my-profile]))
+      (if ephemeral?
+        (fx/merge cofx
```
jeluard: Can the if be moved inside the merge?

src/status_im/extensions/registry.cljs
tbenr: I tried first but the sequence was not correct: i get navigated to extension screen and then the other navigation event arrives moving back to profile. In this way the it seems to work (go back to profile and thennavigate to extension screen arrives).. i dont know if there is a better way to achive the same result

src/status_im/extensions/registry.cljs
tbenr: not really sure we can get compact version with the condition inside.. please guide me if have a good idea :)

src/status_im/extensions/registry.cljs
jeluard: Hmm then it means we rely on timing (even with your option) which is not good. Depending on phone speed you can get weird results. 
Do you think we could have some kind of callback mechanism? e.g. the first event fires itself the second event, thus controlling the flow?

src/status_im/extensions/registry.cljs
jeluard: Right probably simpler the way you did it.

src/status_im/extensions/registry.cljs
tbenr: I don't know really how it works behind the scene here, but I was thinking there is a queue of events and in this way i'm sure I put the events in the correct order.. but if it is not the case, totally agree with you in case it actually depends on timings..

jeluard: You can find some details here: https://github.com/Day8/re-frame/blob/master/docs/Effects.md#order-of-effects

tbenr: ok, right :-)
this evening i'll come with a a better solution!

tbenr: need to study carefully this too https://github.com/Day8/re-frame/blob/master/docs/EventHandlingInfographic.md

bitsikka: don't know too much context.. although I'm working on a bounty that requires full screen overlay so was looking into this for inspiration.. fwiw thought I'd chime in and share that:

in re-frame there is also `:dispatch-later` and `:dispatch-n`

https://github.com/Day8/re-frame/blob/master/docs/API.md

tbenr: yep.. have to do some tests with `:dispatch-n` which should preserve event ordering...but i still have to study\understand if event order guarantee is enough here to have a correct behaviour

tbenr: really need to put event\fx processing in debug mode to see what happens :)

tbenr: Going off for vacation... I'll continue next week!

translations/en.json
```diff
@@ -548,7 +548,6 @@
     "unknown-realm-error": "Something went wrong",
     "notifications": "Notifications",
     "currency-display-name-czk": "Czech Koruna",
-    "extension-installed": "You installed an extension",
```
flexsurfer, line 553: just wondering how is this change related to this PR?

src/status_im/ui/screens/routing/profile_stack.cljs
```diff
@@ -17,6 +17,7 @@
                      :dapps-permissions
                      :manage-dapps-permissions
                      :extensions-settings
+                     :extension-screen-holder
```
flexsurfer, line 20: don't we want to have it in wallet stack as well?

src/status_im/extensions/capacities/views.cljs
```diff
@@ -0,0 +1,13 @@
...
+            [status-im.ui.components.status-bar.view :as status-bar]
+            [status-im.ui.screens.profile.photo-capture.styles :as styles]))
+
+(defview screen-holder []
+  (letsubs [{:keys [view]} [:get-screen-params]]
+    (when view
+      (let [[_ _ reactview] (view)]
+        [react/view styles/container
+         [status-bar/status-bar]
+         reactview]))))
```
flexsurfer, line 13: no toolbar?

src/status_im/ui/screens/routing/profile_stack.cljs
tbenr: right

src/status_im/extensions/capacities/views.cljs
tbenr: we agreed with @jeluard to leave extension developer a full screen and the freedom to place the close button where he prefers...

translations/en.json
tbenr: 'cose in the PR I removed the alerts related to the installation\uninstallation which overlap with extension screen in case the extension uses lifecycle events to show a welcome\splash (see https://github.com/status-im/status-react/pull/7979#issuecomment-484029746). maybe alerts in lifecycle events could become a default bestpractice in extension developement? (we loose translation)

src/status_im/extensions/capacities/views.cljs
jeluard: Right this is fullscreen accessible to users, so nothing. It's up to end dev to decide what's there.

src/status_im/extensions/registry.cljs
yenda: event ordering is not guaranteed with dispatch-n. If you want ordering either use merge-fx (changes to app-db will apply in order, but not side effects), or dispatch the next event in the current one if you want sequential side effects

tbenr: @yenda so you're saying that's ok for you the current version where i'm dispatching the event before returning effects? moreover it seems to me that `(on-*)` lifecycle are returning `nil` so i think they are directly dispatching events instead of generating effects.

So `:navigate-back` event is guaranteed to run before the event dispatched by `(on-installation)` \ `(on-activation)`, right?

jeluard:
@tbenr Do you mind rebasing? Thanks!

tbenr:
@jeluard done!

jeluard:
Thanks Enrico! You are the best!

src/status_im/extensions/registry.cljs
tbenr: i was playing with a solution like:

```
(fx/defn identity-fx
  [_ cb]
  (when cb {::identity-event {:cb cb}}))

(fx/defn install
  [{:keys [db] :as cofx} url {:keys [hooks] :as extension-data} modal?]
  (let [{:account/keys [account]} db
        on-installation (get-in extension-data [:lifecycle :on-installation])
        on-activation   (get-in extension-data [:lifecycle :on-activation])
        ephemeral?      (get-in extension-data [:lifecycle :ephemeral])
        extension       {:id        url
                         :name      (or (get-in extension-data [:meta :name]) "Unnamed")
                         :url       url
                         :lifecycle {:on-installation   on-installation
                                     :on-deinstallation (get-in extension-data [:lifecycle :on-deinstallation])
                                     :on-activation     on-activation
                                     :on-deactivation   (get-in extension-data [:lifecycle :on-deactivation])
                                     :ephemeral?        ephemeral?}
                         :active?   true}
        new-extensions  (assoc (:extensions account) url extension)]
    (if ephemeral?
      (fx/merge cofx
                (when on-activation (on-activation)))
      (fx/merge cofx
                 {:dispatch (if modal?
                              [:navigate-back]
                              [:navigate-to-clean :my-profile])}
                (identity-fx on-installation)
                (identity-fx on-activation)
                (when hooks (accounts.update/account-update {:extensions new-extensions} {}))
                (when hooks (add-to-registry url extension-data true))))))
```
to be adjusted but it gives the idea.. don't know if it is better than the current solution.. i'm actually postponing the on-* execution to fx processing

jeluard: Ping @flexsurfer for ideas

flexsurfer: sorry for the delay, @tbenr  have you looked to https://github.com/status-im/status-react/blob/6b99225e0680c176ad424a168fa534864d4d68df/src/status_im/ui/screens/navigation.cljs#L117 and https://github.com/status-im/status-react/blob/6b99225e0680c176ad424a168fa534864d4d68df/src/status_im/ui/screens/navigation.cljs#L17

src/status_im/extensions/registry.cljs
tbenr: @jeluard i merged the merges 😄

tbenr:
@jeluard @flexsurfer I think now it is OK. thanks @flexsurfer !

tbenr:
(I closed \ repoened PR by mistake)

flexsurfer (approved):
thanks @tbenr !

Serhy:
Tested on iOS 12.2, Andriod8.1, Android6.0.1
As described, with the https://extensions.status.im/?hash=QmRCUoBXf8Sacw7ccK5g76wKxqzYKecfw5Y9SBkp2o8eMd extension splashscreen appers when extension: used in chat, activated, installed (extension activation is part of install)
installation\deinstallation activities are showing Alert pop-up with `[Extention_name] installed/deinstalled!`

No pop-up upon installing / deinstalling existing extension we have (https://status.im/extensions/examples.html) which as I understand is now a part of an extension to pro  provide in its lifecycle.

@tbenr , @jeluard unlike to Develop/Release0.12.1 builds (all good there) in this PR builds I have the next error when navigating to Wallet Kuber extension **[on Android only]**:

Screenshot 2019-05-10 at 17 30 23

And all logs of above error [KuberExtensionWithSplashScreenExt.log](https://github.com/status-im/status-react/files/3166904/KuberExtensionWithSplashScreenExt.log)

I know `Kuber` in WIP but worth to note it so any regression is not introduced.

**To reproduce:**
2) From https://status.im/extensions/examples.html -> `Kyber exchange (WIP)` -> `Install` (https://get.status.im/extension/ipfs@QmeqDrXwwwesgb6Dj9UurtgX1VoS4h6tApxLTLU1BSD7Vo)
3) After extension installed navigate to `Wallet` -> top-right `...` button -> `Exchange Assets`

Note: it's all fine with "simpler" Wallet 'Hello World' extension https://extensions.status.im/?hash=QmTgi12UgbAdQgxwCaSJgWcHvtwtaTpoazi5SWRhChKQhK

tbenr:
thanks @Serhy I confirm what you said about installation\uninstallation popups. Regarding the Kyber exchange, the issue is in the extension code itself (there is a reference to `:icons/forward` which probably as been removed in a recent PR), it doesn't depend by this PR. I think you'll get the same error using the latest develop version. @jeluard can you confirm?

jeluard:
@tbenr Agree with you. (btw looks like you have to rebase :()
@Serhy Can you double check on develop?

Serhy:
Re-tested Kuber instance with latest nightly. It turns out error is present there as well, however there is no it's UI red-screen representation - but I see that error logcat.
Also, tested the same scenario with different PR and see the same red-error about icon as well. 

Sorry for false alarm, @tbenr, @jeluard 

@tbenr could you rebase develop please, so we merge then?

tbenr:
rebased!

PR #7980, opened 2019-04-16
Request to squash commits if the PR has multiple

mandrigin:
Allow our script to do interactive rebase while merging the PR, should streamline the PR process.

PS: if you know a better way to squash stuff non-interactively, feedback appreciated

status: ready

### DIFF
--- a/scripts/merge-pr.sh
+++ b/scripts/merge-pr.sh
@@ -104,6 +104,13 @@ check_is_pr_single_commit() {
   fi
 }
 
+request_to_squash() {
+  if [[ $(git rev-list $BRANCH..$PR_LOCAL_BRANCH | wc -l) -gt 1 ]] ;then
+    confirm "PR has multiple commits, do interactive rebase?"
+    git rebase -i $BRANCH
+  fi
+}
+
 confirm_pr() {
   git log -p $BRANCH..$PR_LOCAL_BRANCH
   confirm "Do you like this PR?"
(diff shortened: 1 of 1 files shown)


rasom (approved):

scripts/merge-pr.sh
```diff
@@ -104,6 +104,13 @@ check_is_pr_single_commit() {
   fi
 }
 
+request_to_squash() {
+  if [[ $(git rev-list $BRANCH..$PR_LOCAL_BRANCH | wc -l) -gt 1 ]] ;then
+    confirm "PR has multiple commits, do interactive rebase?"
+    git rebase -i $BRANCH
```
jakubgs, line 110: A simpler way would be something like:
```bash
MSG=$(git log --format=%B -n 1 $PR_LOCAL_BRANC)
git reset --soft $BRANCH
git commit -m "$MSG"
```

jakubgs: Hm, no, but that would take last commit message, not first commit message...

mandrigin: my thought process was that having an interactive rebase is better than nothing :-D later we can improve the flow even more

jakubgs: That's fair.

jakubgs (approved):

scripts/merge-pr.sh
jakubgs: Though maybe a message or a link to how to use `rebase -i` would be nice.

mandrigin:
it doesn't affect anything except merging multi-commit PRs, so I will merge it with no separate QA. tested it locally and it _works on my machine_ 😆

jakubgs:
but_i_understand

PR #7981, opened 2019-04-16
Update partitioned topic derivation algorithm

rasom:
@cammellos :
> so xPK % number-of-partitions,  instead of random-seeded-number(pk) % number-of-partitions

status: ready

### DIFF
--- a/src/status_im/transport/partitioned_topic.cljs
+++ b/src/status_im/transport/partitioned_topic.cljs
@@ -1,6 +1,5 @@
 (ns status-im.transport.partitioned-topic
-  (:require [status-im.utils.random :as random]
-            [status-im.transport.utils :as utils]
+  (:require [status-im.transport.utils :as utils]
             [status-im.constants :as constants]
             [status-im.utils.config :as config]))
 
@@ -26,9 +25,11 @@
 
 (defn- partitioned-topic
   [public-key]
-  (let [gen (random/rand-gen public-key)]
-    (-> (random/seeded-rand-int gen n-partitions)
-        (str "-discovery"))))
+  (-> public-key
+      (subs 3 35)
+      (js/parseInt 16)
+      (mod n-partitions)
+      (str "-discovery")))
 
 (defn partitioned-topic-hash
   "Given a public key return a partitioned topic between 0 and n"
--- a/test/cljs/status_im/test/transport/core.cljs
+++ b/test/cljs/status_im/test/transport/core.cljs
@@ -17,7 +17,7 @@
     (testing "it adds the discover filters"
       (is (= {:web3 nil :private-key-id "1" :topics '({:topic   "topic-4"
                                                        :chat-id "4"}
-                                                      {:topic   "0x2af2e6e7"
+                                                      {:topic   "0xd2a5f1a3"
                                                        :chat-id :discovery-topic}
                                                       {:topic   "0xf8946aac"
                                                        :chat-id :discovery-topic})}


mandrigin (changes requested):
https://github.com/status-im/specs/pull/3 should be updated with this algorithm cc @adambabik @oskarth

cammellos (approved):

oskarth:
Indeed @mandrigin, preferably separate proposal with compatibility and clear upgrade path etc. See previous dev calls for legwork that was identified as needed before putting this in production https://github.com/status-im/pm/issues

rasom:
@mandrigin @oskarth so what's the plan then? We need to create proposal, discuss it, approve and only then implement, right? 

There is a probability that the whole algorithm (I mean not only topic derivation) will be changed (isn't it?), and on the other hand it might be accepted just as it is in this PR. We are already listening to the topic which we are not going to use, so including this commit to the next release will not make any harm even if we will change the whole thing.

I don't mind to close this PR if we need to wait till discussion will be finished.

adambabik:
Let's just create another PR to `status-protocol` branch in https://github.com/status-im/specs repo. It should be a self-contained section somewhere below `Topic` header, for example as another header `### Partitioned Topic`.

adambabik:
Oh and one concern I have is the number of topics an individual would need to subscribe to. For example, if I have `N` private contacts, I need to provide `N` topics to the Whisper filter. Has anyone verified what's the maximum `N` for Whisper?

oskarth:
What @adambabik said. Spoke with @mandrigin about this, but it should be a separate PR to specs repo with a proposal that takes into account upgradability, etc. It doesn't have to build on the current spec WIP PR, but it can refer to it.

Relevant here would also be to mention:
- bandwidth usage and how it might change with launch and users
- how this would impact trade-off of potentially not having compatibility
- what a 100% or 99.9% compatibilty would look like (multicast vs device abstraction) and what it requires
- entertain option of what keep listening to discovery topic and send less over it through some form of toggle would look like, i.e. 99.9% compatibility case (similar to device sync, but perhaps corner case re recovered account way in future)
- upgrade path, both from old to new as well as from proposal to further (say we change topic partition again in future)
- summary of options with trade-offs
- possibly options for sacrificing compatibility in short term and regaining it later, e.g. with device data sync 
- evaluation should have a big premium/bias towards keeping compatibility, for reasons mentioned in several previous core dev calls

I realize this sounds like a lot, and it'd be easier to just hack it and implement it. But we are trying to create a protocol and spec that exists on its own, and can't be changed ad hoc. This will only increase in importance going forward, especially as we get more clients. By doing this properly now we are changing the culture of what spec changes look like, and ensuring we know how to do it now rather than later. It is similar to Chaos Unicorn Day in that sense. It is cheap to learn this now, expensive to learn it later.

As a role model here I just want to point to @arnetheduck, who has a lot of experience on working on p2p protocols, including getting 70m installs for a compatible reverse-engineered p2p app 10 years ago. @cammellos rightfully points out there are some differences in terms of protocol negotiation etc, and this is the kind of consideration we want to see in the spec proposal. Similarly, @mandrigin and others have pointed out that we do have real-world constraints, such as launching an app with non-sucky BW usage, which might lead to different trade-offs. This type of thing can too be reasoned about thoughtfully in a proposal which clearly outlines different options and trade-offs with rationale.

Once that proposal is in a better shape, let's bring it up in Core Dev call.

rasom:
@adambabik @oskarth @mandrigin ok

PR #7982, opened 2019-04-16
don't use docker containers on linux, nix rules all

jakubgs:
Since we use Nix for all dependencies, using Docker seems to be redundant and unnecessarily complicate things.

I did try to create a docker container that could share a `/nix/store` with the host system, but we can't set `NIX_STORE_DIR` to something like `/opt/nix/store`, because then out https://nix-cache.status.im doesn't work since it expects `/nix/store` to be the path. And if we map `/nix/store` as a volume to the host from the container then it would need to contain at least the minimal Nix installation(`bash`, `gcc`, `ld`, `curl`) in advance, which makes the container redundant and useless.


mandrigin (approved):
**-275** 🥇🥇🥇🥇

pedropombeiro (approved):

PR #7997, opened 2019-04-18
Multiple messages gaps per chat

rasom:
- gaps are shown whenever some range of time has not been covered by mailserver request
- chat may have multiple gaps
- if there are no messages between two gaps, they are merged and shown as one in the chat. When user tries to load messages merged gaps are requested one by one
- gaps are stored per chat, nor per topic, i.e. if a new chat is created and it has the same topic as another already existing chat, a new one will have a separate list of gaps.

todo:
- [x] delete gaps older than N (30-45?) days
- [x] make sure gaps can be fetched "in parallel"
- [x] make sure gaps fetching might be started during regular syncing
- [x] limit to 24h
- [x] remove ranges/gaps when chat is removed
- [x] db migration (add private groups/1-1 to `:discovery` mail server topic)
- [x] tests
- [x] switch back to 24h after testing
status: ready

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -19,7 +19,9 @@
             [status-im.utils.gfycat.core :as gfycat]
             [status-im.utils.platform :as platform]
             [status-im.utils.priority-map :refer [empty-message-map]]
-            [status-im.utils.utils :as utils]))
+            [status-im.utils.utils :as utils]
+            [status-im.mailserver.core :as mailserver]
+            [status-im.transport.partitioned-topic :as transport.topic]))
 
 (defn- get-chat [cofx chat-id]
   (get-in cofx [:db :chats chat-id]))
@@ -163,6 +165,10 @@
   (fx/merge cofx
             #(when (public-chat? % chat-id)
                (transport.chat/unsubscribe-from-chat % chat-id))
+            #(when (group-chat? % chat-id)
+               (mailserver/remove-chat-from-mailserver-topic % chat-id))
+            (mailserver/remove-gaps chat-id)
+            (mailserver/remove-range chat-id)
             (deactivate-chat chat-id)
             (clear-history chat-id)
             #(when (one-to-one-chat? % chat-id)
@@ -242,13 +248,12 @@
 (fx/defn preload-chat-data
   "Takes chat-id and coeffects map, returns effects necessary when navigating to chat"
   [{:keys [db] :as cofx} chat-id]
-  (let [chat (get-in db [:chats chat-id])]
-    (fx/merge cofx
-              {:db (-> (assoc db :current-chat-id chat-id)
-                       (set-chat-ui-props {:validation-messages nil}))}
-              (contact-code/listen-to-chat chat-id)
-              (when platform/desktop?
-                (mark-messages-seen chat-id)))))
+  (fx/merge cofx
+            {:db (-> (assoc db :current-chat-id chat-id)
+                     (set-chat-ui-props {:validation-messages nil}))}
+            (contact-code/listen-to-chat chat-id)
+            (when platform/desktop?
+              (mark-messages-seen chat-id))))
 
 (fx/defn navigate-to-chat
   "Takes coeffects map and chat-id, returns effects necessary for navigation and preloading data"
@@ -277,7 +282,7 @@
   ;; don't allow to open chat with yourself
   (when (not= (accounts.db/current-public-key cofx) chat-id)
     (fx/merge cofx
-              (upsert-chat {:chat-id chat-id
+              (upsert-chat {:chat-id   chat-id
                             :is-active true})
               (navigate-to-chat chat-id opts))))
 
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -21,7 +21,8 @@
             [status-im.transport.partitioned-topic :as transport.topic]
             [status-im.utils.config :as config]
             [status-im.chat.models.loading :as chat.models.loading]
-            [status-im.chat.models.message :as chat.models.message]))
+            [status-im.chat.models.message :as chat.models.message]
+            [status-im.mailserver.core :as mailserver]))
 
 (fx/defn load-contacts
   [{:keys [db all-contacts]}]
@@ -74,7 +75,12 @@
       (fx/merge cofx
                 {:db (assoc-in db [:contacts/new-identity] "")}
                 (upsert-contact contact)
-                (send-contact-request contact)))))
+                (mailserver/upsert-mailserver-topic
+                 {:chat-ids [public-key]
+                  :topic    transport.topic/discovery-topic-hash
+                  :fetch?   false})
+                (send-contact-request contact)
+                (mailserver/process-next-messages-request)))))
 
 (fx/defn add-contacts-filter [{:keys [db]} public-key action]
   (when (not= (get-in db [:account/account :public-key]) public-key)
--- a/src/status_im/group_chats/core.cljs
+++ b/src/status_im/group_chats/core.cljs
@@ -12,14 +12,13 @@
             [status-im.group-chats.db :as group-chats.db]
             [status-im.i18n :as i18n]
             [status-im.native-module.core :as native-module]
-            [status-im.transport.chat.core :as transport.chat]
             [status-im.transport.message.group-chat :as message.group-chat]
             [status-im.transport.message.protocol :as protocol]
-            [status-im.transport.message.public-chat :as transport.public-chat]
             [status-im.transport.partitioned-topic :as transport.topic]
             [status-im.utils.clocks :as utils.clocks]
-            [status-im.utils.config :as config]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.mailserver.core :as mailserver]
+            [taoensso.timbre :as log]))
 
 ;; Description of the flow:
 ;; the flow is complicated a bit by 2 asynchronous call to status-go, which might make the logic a bit more opaque.
@@ -213,10 +212,17 @@
         events            [create-event
                            (members-added-event clock-value selected-contacts)]]
 
-    {:group-chats/sign-membership {:chat-id chat-id
-                                   :from   my-public-key
-                                   :events events}
-     :db (assoc db :group/selected-contacts #{})}))
+    (fx/merge
+     cofx
+     {:group-chats/sign-membership {:chat-id chat-id
+                                    :from    my-public-key
+                                    :events  events}
+      :db                          (assoc db :group/selected-contacts #{})}
+     (mailserver/upsert-mailserver-topic
+      {:chat-ids [chat-id]
+       :topic    transport.topic/discovery-topic-hash
+       :fetch?   false})
+     (mailserver/process-next-messages-request))))
 
 (fx/defn remove-member
   "Format group update message and sign membership"
@@ -236,7 +242,7 @@
 
 (fx/defn join-chat
   "Format group update message and sign membership"
-  [{:keys [db] :as cofx} chat-id]
+  [cofx chat-id]
   (let [my-public-key     (accounts.db/current-public-key cofx)
         last-clock-value  (get-last-clock-value cofx chat-id)
         chat              (get-in cofx [:db :chats chat-id])
@@ -244,9 +250,16 @@
     (when (valid-event? chat (assoc event
                                     :from
                                     my-public-key))
-      {:group-chats/sign-membership {:chat-id chat-id
-                                     :from    my-public-key
-                                     :events  [event]}})))
+      (fx/merge
+       cofx
+       {:group-chats/sign-membership {:chat-id chat-id
+                                      :from    my-public-key
+                                      :events  [event]}}
+       (mailserver/upsert-mailserver-topic
+        {:chat-ids [chat-id]
+         :topic    transport.topic/discovery-topic-hash
+         :fetch    false})
+       (mailserver/process-next-messages-request)))))
 
 (fx/defn make-admin
   "Format group update with make admin message and sign membership"
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -18,7 +18,8 @@
             [status-im.accounts.update.core :as accounts.update]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.transport.partitioned-topic :as transport.topic]
-            [status-im.ui.screens.mobile-network-settings.utils :as mobile-network-utils]))
+            [status-im.ui.screens.mobile-network-settings.utils :as mobile-network-utils]
+            [status-im.utils.random :as rand]))
 
 ;; How do mailserver work ?
 ;;
@@ -35,6 +36,7 @@
 
 (def one-day (* 24 3600))
 (def seven-days (* 7 one-day))
+(def max-gaps-range (* 30 one-day))
 (def max-request-range one-day)
 (def maximum-number-of-attempts 2)
 (def request-timeout 30)
@@ -249,7 +251,8 @@
     (fx/merge cofx
               {:db (cond-> (dissoc db :mailserver/current-request)
                      gap-request?
-                     (assoc :mailserver/fetching-gaps-in-progress {}))}
+                     (-> (assoc :mailserver/fetching-gaps-in-progress {})
+                         (dissoc :mailserver/planned-gap-requests)))}
               (if added?
                 (mark-trusted-peer)
                 (add-peer)))))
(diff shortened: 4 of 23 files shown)


src/status_im/chat/models.cljs
```diff
@@ -276,10 +281,11 @@
   [{:keys [db] :as cofx} chat-id opts]
   ;; don't allow to open chat with yourself
   (when (not= (accounts.db/current-public-key cofx) chat-id)
-    (fx/merge cofx
-              (upsert-chat {:chat-id chat-id
-                            :is-active true})
-              (navigate-to-chat chat-id opts))))
+    (let []
```
mandrigin: why this `let`?

src/status_im/contact/core.cljs
```diff
@@ -74,7 +75,12 @@
       (fx/merge cofx
                 {:db (assoc-in db [:contacts/new-identity] "")}
                 (upsert-contact contact)
-                (send-contact-request contact)))))
+                (mailserver/upsert-mailserver-topic
```
mandrigin, line 78: how is this related to the message gaps?

mandrigin (approved):
a few notes, but generally looks good

churik:
@rasom so far:

#### 1) "Fetch message" can be located under last message, if message should be fetched in next interval after it

Steps:
- send message in particular minute(i.e. 'aaaa' in 12:54)
- go to offline
- send message from another device 1 minute later (i.e. 12:55. also reproducible for me with 2, 3 mins - didn't try more)
- wait for 1 minute
- go back and check where is "Fetch messages"
Expected result: "Fetch message" after 'aaa'
Actual result: "Fetch message" before 'aaa'
It can be just due to 60s interval, so let me know if we can check this for 24 h.
ererwewrerw

####  2) I guess, there are always redundant gaps, so you need to tap on "Fetch message" more times than number of missing gaps;
I made a test:
- send message in 13:14
**1 Gap (2 min):**
- go to offline
- send message in 13:16
- go back to 13:18 (so I suppose that messages from 13:17 are fetched too)
- send message
**2 Gap (1 min):**
- go back to offline
- send message to 13:19
- go back to offline in 13:21 (so I suppose that messages from 13:20 are fetched too)

So I expect 2 times for` 1 Gap` and 1 time for `2 gap` will fetch the whole history.
In fact for 1 gap I should tap 4 times.
Video: https://take.ms/3xppF
Logs: [Status-debug-logs (1).zip](https://github.com/status-im/status-react/files/3107290/Status-debug-logs.1.zip)

####  3) For now it is not obvious that you need to tap on "Fetch message" several times if you miss a week for example (so 7 times should be tapped to fetch missing history)

I found it very confusing when you need to tap on "Fetch messages" several times - so I'm wondering it would be much clearer if I can see smth like  "Fetch messages for 24.03.19" 
And fact that it can fetch nothing (so history for interval can be empty) makes it even more confusing.

Let's ask @hesterbruikman how we can make it more clear for users.

src/status_im/contact/core.cljs
rasom: we need to add `public-key` to the set of `chat-ids` of `discovery`'s `mailserver-topic` in order to track requests per chat

src/status_im/chat/models.cljs
rasom: just leftover, will remove

rasom:
1. @asemiankevich as I already mentioned in previous PR that's specific 60s range issue, it will be hard to notice if we have 24h range. When you log into account at time `T`, request which loads history for last 60s is executed for a range `(T-60s) -- T`. Then you send a message at `T+10`, go offline, and let's say open app 100s later. The next request will be for a range `(T+50)--(T+110)`, which means that a new gap for a range `(T)--(T+50)` will be created. And this gap will be shown before the message which is sent at `T+10`.
Again, I wouldn't give too much attention to this as it's unlikely a big problem for 24h range.

2. If gaps are merged and we press "fetch messages" on merged gaps, 60s of each gap are fetched. i.e. if we have two gaps `(T-180)--(T)` and `(T+80)--(T+100)`, when fetching will be initiated the first time only one gap will still be present: `(T-180)--(T-60)`. In order to fetch all messages from this gap we will need to press it two more times. Before merging 60s will be switched to 24h. 

3. We can expand gap and show missed ranges when user clicks some question mark aside of "Fetch messages" panel, or something. Also take in count that this situation is much less likely to appear (gap which needs several iterations) with 24h range.

src/status_im/contact/core.cljs
mandrigin: okay, cool, was curious, because it looks a bit weird... maybe a comment here will help, but I'll resolve this comment chain.

hesterbruikman:
My initial thought is that we should not offload this to the user. If the user taps "Fetch message" it should mean we fetch all ranges within the gap. The fetch messages button should represent all these ranges. 

@rasom can you elaborate on why we can't fetch mutiple ranges triggered by one tap? Apologies if you've mentioned it before, I don't think I fully understand this yet.

rasom:
@hesterbruikman as I mentioned before - we fetch all gaps, but we fetch only 24h of each gap. If we fetch more than 24h there is a chance that request will be "never ending". It's not a problem to request more than 24h but it might not work.

hesterbruikman:
So currently we say 'Fetch messages'. Would it be more accurate to say 'Fetch last 24h'? At least that way we don't imply that we fetch more messages than that. This would probably require somehow marking all gaps of being 24h batches.

Please tell me if this is a correct understanding:
- Based on date, last message fetched and last prior message we know that there is a gap
- If there is one or more 24h gap we show 'fetch messages' in the gap(s)
- When the user taps on it we only fetch 24h. 
_The range of the gaps might be bigger. We know how big the range is._

If above is true, ideally we expand with:
- gap is filled with fetched 24h of messages
- if more gaps (of again 24h) exist 'Fetch last 24h' is shown again, now in a different position
- user taps again
- repeat
- if more messages exist

rasom:
>  Would it be more accurate to say 'Fetch last 24h'? 

no I doubt. If we don't state from which point in time we fetch these "last 24h" and thus it doesn't make sense. Also gap might be 1h, 123s, etc - not necessary 24h and more.

- based on ranges of previous requests to mailservers we know were we might have gaps (no guaranty that gaps actually exist, app could be online all the time).
- if we requested history three times for intervals `(T-24h)--(T)`, `(T+12h)--(T+36h)`, `(T+72h)--(T+96h)` two gaps exist `T--(T+12h)`, `(T+36h)--(T+72h)`
- if there are no received messages in  `(T+12h)--(T+36h)` interval these two gaps are shown as one
- when we press `Fetch messages` not more than 24h of each gap is requested, assuming that gaps are merged we proceed with two requests: `T--(T+12h)`, `(T+48h)--(T+72h)` and then only is left `(T+36h)--(T+48h)`

hesterbruikman:
I think I'm out of ideas:/

- Does it make sense to start fetching the oldest messages first? If possible, maybe it's a way to load all older messages on top of, rather than below the button. That way we can give it more of a familiar 'load more' experience, sometimes used on web design. Making multiple taps more natural.
- at this stage, understanding the work that has gone into this PR, do you think the concept in this prototype could be a better option overall? There, under the [...] a bottom sheet comes up that let's the user specify cor how many days they want to fetch history, giving control per chat.
https://framer.cloud/VjTBM/

mandrigin:
@hesterbruikman 
> Does it make sense to start fetching the oldest messages first?

I'm not sure if that is a good idea. If I'm scrolling up in the chatroom, I will probably want to gradually show more and more history from down to up.

hesterbruikman:
"I'm not sure if that is a good idea. If I'm scrolling up in the chatroom, I will probably want to gradually show more and more history from down to up."

Normally, yes. But in this case it means:
 - Tapping to fetch >messages appear below
- Noticing they are only recent messages, tapping again >new messages appear in between the message I just fetched and the button
- Repeat

An alternative in this order would be to introduce an animation and push the already fetched message down. Keeping focus on the last loaded message, which is always right below the button, therefore not requiring scroll to see what messages were added.

mandrigin:
@hesterbruikman ah, when the "Fetch" button is pressed, scroll shoudn't move, I think.

So basically if you see 

```
[ fetch ]
Alice: howdy?
---------------------
[                      ]  [ ↑ ]
```

then after pressing fetch it should be
```
Bob: oh, Alice, you here!
Alice: howdy?
---------------------
[                      ]  [ ↑ ]
```

hesterbruikman:
> So basically if you see
> 
> ```
> [ fetch ]
> Alice: howdy?
> ---------------------
> [                      ]  [ ↑ ]
> ```
> 
> then after pressing fetch it should be
> 
> ```
> Bob: oh, Alice, you here!
> Alice: howdy?
> ---------------------
> [                      ]  [ ↑ ]
> ```

Scroll remaining by the button definitely helps. 

If we can fix that, it would leave 2 issues IMO that strengthen eachothers effect:
- Having to tap Fetch messages multiple times
- Messages appearing without defined transition

If there's no way around the user tapping [Fetch messages] mutiple times I would suggest the following to mitigiate these issues:
- Maintain scroll position on button
- Change copy to: 'Fetch more messages'
- Create a new issue to animate pushing newer messages down when older messages are added to the top

@rasom @churik @mandrigin  wdyt?

src/status_im/chat/commands/impl/transactions.cljs
```diff
@@ -27,7 +27,8 @@
             [status-im.models.transactions :as wallet.transactions]
             [status-im.ui.screens.navigation :as navigation]
             status-im.chat.commands.impl.transactions.subs
-            [status-im.ui.screens.wallet.utils :as wallet.utils]))
+            [status-im.ui.screens.wallet.utils :as wallet.utils]
+            [taoensso.timbre :as log]))
```
cammellos: looks like log might not be needed? I can't see any other change in this file

src/status_im/group_chats/core.cljs
```diff
@@ -504,6 +517,7 @@
                               ;; don't allow anything but group messages
                               (instance? protocol/Message message)
                               (= :group-user-message (:message-type message)))
+                     (log/debug "Receiving " (prn-str message))
```
cammellos: Possibly leftover?

src/status_im/mailserver/core.cljs
```diff
@@ -249,7 +251,10 @@
     (fx/merge cofx
               {:db (cond-> (dissoc db :mailserver/current-request)
                      gap-request?
-                     (assoc :mailserver/fetching-gaps-in-progress {}))}
+                     (as-> $
```
cammellos: This looks a bit odd, some leftover? you probably can thread first directly without `as->` or I am missing something?

src/status_im/mailserver/core.cljs
```diff
@@ -764,28 +912,35 @@
...
+                                   :gap       id}]))
+                              gaps)
+        first-request   (val (first requests))
+        current-request (:mailserver/current-request db)]
+    (cond-> {:db (-> db
+                     (assoc :mailserver/planned-gap-requests requests)
+                     (update :mailserver/fetching-gaps-in-progress
+                             assoc chat-id requests))}
+      (not current-request)
+      (as-> $
```
cammellos: same here, not sure why `as->` is needed, but might be missing something

src/status_im/transport/message/core.cljs
```diff
@@ -44,14 +44,16 @@
...
-                    #(protocol/receive valid-message
-                                       (or
-                                        filter-chat-id
-                                        (get-in valid-message [:content :chat-id])
-                                        sig)
-                                       sig
-                                       timestamp
-                                       %)))
+                    #(do
+                       (log/debug "Receiving " (prn-str valid-message))
```
cammellos: probably left over?

cammellos (approved):

rasom:
@cammellos changes according to comments are added

cammellos:
@rasom thanks!

rasom:
> Scroll remaining by the button definitely helps.

@hesterbruikman, did you try it? That's how it works.  

The only case when it can be a bit scrolled atm is when we have messages with incorrect timestamps, so that message from timespan which we requested is added in-between already rendered messages. It would be quite hard to keep position in such case at the moment (but still it could be fixed if we handle messages with such timestamps differently). In all other cases, the position should be maintained as expected.

hesterbruikman:
Thanks for clarifying @rasom!

I tried the build, but couldn't get the button to show up on iOS. Tried the connect/disconnect tricks by @churik, but couldn't mimick the behavior so I'm relying on the screen recording.

I'm fine with the behavior deviating if it happens as a result of timestamps. Seems like that's a different problem we shouldn't address here and we have several issues describing it.

In that case I'd still prefer a copy update to 'Fetch more messages' to deal with the multiple tap requirement. And I'll create a new issue to look into animating messages pushed down.

Sounds good?

Could you give a rough estimate on effort for this new issue? Would like to see if we can put a bounty on it and actively invite people who are into animations in RN, we have similar issues that would be worthwhile looking into. (E.g. search bar behavior).

rasom:
@hesterbruikman just figured that another case (and probably the one which happens more frequently) is when we have some responses already rendered, but messages to which responses were sent are loaded by last request and so previously received messages are re-rendered and position is lost.

hesterbruikman:
Do you mean the scroll position is lost in that case? If so, do you see a way to prevent that?

Overall it seems to me you've carefully investigated the fetching of messages. If we hit the limit of ordering and sequence of display that feels natural there, I think the best way forward is to document the cases where natural chronological order is lost for whatever reason. E.g. rerendering of a reply as an original message comes in or timestamps inconsistency.

We can then move forward with the PR and follow up to improve on the rendering behavior to make it more gradual so that at least there's a visual transition to show what changed. Instant changes are simply more difficult to grasp than seeing the progress of something moving up/down/re-render.

rasom:
> Do you mean the scroll position is lost in that case?

yes, because before we load missed messages we render response without quoted message, and then we re-render response message with it.

> We can then move forward with the PR and follow up to improve on the rendering behavior to make it more gradual so that at least there's a visual transition to show what changed. Instant changes are simply more difficult to grasp than seeing the progress of something moving up/down/re-render.

agree, why can try to address rendering issues in a separate PR, there is no guaranty that we can fix all of them. Current PR is big enough already.

churik:
@rasom 
#### 4) No gaps are shown for 1-1 chats (that are not added to contacts)
Steps: 
- create 1-1 chat
- go to offline
- wait for 1 minute+ (checked on builds with default interval 1 min)
- go back and check where is "Fetch messages"

All fine with public and group chats.
Platform: IOS, Android

rasom:
@churik was that 1-1 chat added to contacts?

churik:
@rasom no, it wasn't. So we don't allow to fetch messages for users that are not added as contacts?

rasom:
@churik currently yes

churik:
@rasom so it will be addressed separately I guess?

rasom:
@churik yes

churik:
**Builds for IOS, Android (60s):**
IOS: https://i.diawi.com/rvEGaS
Android: https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190423-071929-fac16d-pr.apk

**Builds for IOS, Android (24h, with the possibility to fetch  48-60, 84-96 ):**
IOS: https://i.diawi.com/4BNmLw
Android: https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190424-054508-b97f0c-pr.apk

**Tested on build with interval 1 min:**
- several gaps with one interval and several messages between them (https://github.com/status-im/status-react/pull/7997#issuecomment-485766889, some issues that are introduced by such short default interval) in 1-1, group, public chats;
Issue https://github.com/status-im/status-react/pull/7997#issuecomment-486141385 will be addressed separately;
- several gaps with 1+ intervals and several messages between and after them in 1-1, group, public chats;
- "fetch messages" is blocked while offline; 
- the gap at the end of history;
- the gap is not shown before joining the public chat (when the invite is received);
- scroll position is persisted when you fetch messages (so you are able to see last message below).
Issue https://github.com/status-im/status-react/pull/7997#issuecomment-486083811 will be addressed separately;
- try to tap on several "fetch messages" simultaneously;
- app in background/foreground doesn't affect "fetch messages"
- upgrade app (gaps are shown only for intervals missed in the upgraded app) - IOS only, can't upgrade Android app to PR build.

**Tested on build with interval 24h:**
- several gaps with 1+ intervals and several messages between and after them in public chats;
- "fetch messages" is blocked while offline; 
- scroll position is persisted when you fetch messages (so you are able to see last message below).

**Also checked:**
- pairing;
- "Fetch 24h"
- checked compatibility with current release 0.11.0;
- block user, can't receive messages to 1-1 from him;
- doesn't affect desktop builds

@rasom I believe that that `fetch 48-60, 84-96` can be removed now, if you can't see more cases to be checked.

@annadanchenko let me know if you see more cases / areas

rasom:
> I believe that that fetch 48-60, 84-96 can be removed now, if you can't see more cases to be checked.

done

annadanchenko:
> @annadanchenko let me know if you see more cases / areas

Thank you @churik I don't have other suggestions, ok to merge. We'll need to try upgrade with nightly after the merge

PR #8003, opened 2019-04-19
Block user from 1-1 chat header

churik:
Tests a bit complicated, but I don't see any reason to split it, because we need to check that messages from blocked user are deleted from public and 1-1 chat and are not reappered. 

I decided that deleting user messages after blocking from group chat can be separate test with high priority, so modified https://ethstatus.testrail.net/index.php?/cases/view/5763  (to avoid overcomplicating this test, which is already too big)

### DIFF
--- a/test/appium/tests/atomic/chats/test_chats_management.py
+++ b/test/appium/tests/atomic/chats/test_chats_management.py
@@ -1,7 +1,7 @@
 import pytest
 import time
 
-from tests import marks, camera_access_error_text
+from tests import marks, camera_access_error_text, get_current_time
 from tests.users import basic_user
 from tests.base_test_case import SingleDeviceTestCase, MultipleDeviceTestCase
 from views.sign_in_view import SignInView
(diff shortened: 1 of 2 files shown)


Serhy:
@churik, I agree that test too complicated for PR level for now. 
Block user is the main feature we test here and so we may test it in the scope of 1-1 chat only on critical level of the test case.
It will speed up the test execution time and make it less false-error prone.
Block user in public chat part and group chat could be lower priority.

Could you split it, please, by chat type?

churik:
Well, I don't agree that 1-1 chat is critical.
Both public and 1-1 chat are critical, and I can even say that mostly spammer are using public channels (agree that group chats are not critical at the moment)
Even if we split it by chat type:
1) it didn't cover the case when you block user from 1-1 chat -> his messages are hidden from public chat
2) two cases will take even more time.
So I don't see any specific reason here.
We can discuss it on our sync test-team meeting btw

Serhy (approved):

asemiankevich (approved):

PR #8005, opened 2019-04-19
[fix] fetch wallet token transactions

yenda:
- previously wallet transaction fetching was quite hazardous
- most of the time the `:wallet/all-tokens` map was nil so no token
was recognized
- the query for transfers was made accross 100000 blocks, now it
only checks that the first time, then from latest block checked minus
12 to follow progress of unconfirmed transactions

blocking tribute to talk

### Steps to test

send snt/stt to the account (no with another account which sends a message), unlike before the transaction should appear in transaction list

status: ready

### DIFF
--- a/src/status_im/models/transactions.cljs
+++ b/src/status_im/models/transactions.cljs
@@ -114,11 +114,13 @@
        (error-fn error)
        (add-block-info web3 current-block-number chain-tokens direction result success-fn)))))
 
-(defn- limited-from-block [current-block-number]
+(defn- limited-from-block [current-block-number latest-block-checked]
   {:pre [(integer? current-block-number)]
    ;; needs to be a positive etherium hex
    :post [(string? %) (string/starts-with? % "0x")]}
-  (-> current-block-number (- block-query-limit) (max 0) ethereum/int->hex))
+  (if latest-block-checked
+    (-> latest-block-checked (- confirmations-count-threshold) ethereum/int->hex)
+    (-> current-block-number (- block-query-limit) (max 0) ethereum/int->hex)))
 
 ;; Here we are querying event logs for Transfer events.
 ;;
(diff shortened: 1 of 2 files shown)


src/status_im/models/transactions.cljs
```diff
@@ -377,12 +382,14 @@
             transaction-map (:transactions wallet)
             transaction-ids (set (keys transaction-map))
             chaos-mode? (get-in account [:settings :chaos-mode?])]
-        (if-not (or (have-unconfirmed-transactions? (vals transaction-map))
+        (if-not (or @latest-block-checked
+                    (have-unconfirmed-transactions? (vals transaction-map))
                     (not-empty (set/difference chat-transaction-ids transaction-ids)))
-          (done-fn)
-          (transactions-query-helper web3 all-tokens account-address chain done-fn chaos-mode?))))))
+          (done-fn))
```
yenda: need to fix this parens

goranjovic (approved):

vitvly (approved):

churik:
@yenda amazing work!
It fixes https://github.com/status-im/status-react/issues/5749.
All that I noticed https://github.com/status-im/status-react/issues/7329 is reproduced, but if you see 0 ETH - only relogin will help to see  normal transactions for tokens.
However, it is the same on nightly. And I faced with behavior that sometimes after relogin not all transactions are loaded and only relogin helps to see all token and ETH transactions (video https://take.ms/JLEYs) 
These issues are not constantly reproduced (IMO all related to all related to https://github.com/status-im/status-react/issues/7329)

Should be fixed separately later, WDYT @goranjovic @yenda ?

churik:
Tested (IOS 11.4.1 - IPhone7, Android 8 - LG V20: on both mainnet and Ropsten):
- /send, /request command in chats (tokens, ETH)
- sending transactions to DApp (status-im.github.io/dapp)
- sending tokens and ETH from wallet 
- incoming and outgoing transactions from DApps

It is definitely better than in nightly - in most cases fresh token transactions are shown correctly.
Hewever, still reproducible https://github.com/status-im/status-react/issues/6835#issuecomment-485442348 and https://github.com/status-im/status-react/issues/7329#issuecomment-485433471

PR #8009, opened 2019-04-22
[7947] Feature: Blank out app preview screen when switching apps on m…

### LINKED ISSUE #7947: Feature: Blank out app preview screen when switching apps on mobile
# Feature Request

As a user, I'd like the app preview screen to be blanked out when switching apps.  This would increase my privacy.

As a bonus, I would like this to be a user option, but on by default.

bitsikka:
Fixes #7947 

Notes: https://github.com/status-im/status-react/pull/8009#issuecomment-490671670

status: ready

### DIFF
--- a/android/app/src/main/java/im/status/ethereum/MainActivity.java
+++ b/android/app/src/main/java/im/status/ethereum/MainActivity.java
@@ -12,8 +12,10 @@ import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Looper;
+import android.preference.PreferenceManager;
 import android.support.v4.app.ActivityCompat;
 import android.util.Log;
+import android.view.WindowManager;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.res.Configuration;
@@ -109,7 +111,7 @@ public class MainActivity extends ReactActivity
         final ActivityManager activityManager = getActivityManager();
         Log.v("RNBootstrap", "Available system memory "+getAvailableMemory(activityManager).availMem + ", maximum usable application memory " + activityManager.getLargeMemoryClass()+"M");
 
-
+        setSecureFlag();
         SplashScreen.show(this, true);
         super.onCreate(savedInstanceState);
 
@@ -200,6 +202,16 @@ public class MainActivity extends ReactActivity
         SharedPreferences.Editor editor = preferences.edit();
         editor.putBoolean(REJECTED_ROOTED_NOTIFICATION, true);
         editor.commit();
+    }
+
+    private void setSecureFlag() {
+        final SharedPreferences sharedPrefs = PreferenceManager.getDefaultSharedPreferences(this);
+        final boolean setSecure = sharedPrefs.getBoolean("BLANK_PREVIEW", false);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.ICE_CREAM_SANDWICH && setSecure) {
+            getWindow().addFlags(WindowManager.LayoutParams.FLAG_SECURE);
+        } else {
+            getWindow().clearFlags(WindowManager.LayoutParams.FLAG_SECURE);
+        }
     }
 
     @TargetApi(Build.VERSION_CODES.M)
--- a/ios/StatusIm/AppDelegate.m
+++ b/ios/StatusIm/AppDelegate.m
@@ -20,7 +20,10 @@
 #import "RNFirebaseNotifications.h"
 #import "RNFirebaseMessaging.h"
 
-@implementation AppDelegate
+@implementation AppDelegate 
+{
+    UIView *_blankView;
+}
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
@@ -47,6 +50,10 @@
     RCTSetLogThreshold(RCTLogLevelTrace);
   }
 
+  NSDictionary *appDefaults = [NSDictionary
+      dictionaryWithObject:[NSNumber numberWithBool:NO] forKey:@"BLANK_PREVIEW"];
+  [[NSUserDefaults standardUserDefaults] registerDefaults:appDefaults];
+
   RCTBridge *bridge = [[RCTBridge alloc] initWithDelegate:self launchOptions:launchOptions];
   RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge
                                                       moduleName:@"StatusIm"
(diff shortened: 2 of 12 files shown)


bitsikka:
blanked out screen looks like this

image

bitsikka:
screen with Share panel looks like this

image

bitsikka:
Screen that has an Alert box active looks like this

image

bitsikka:
Tested only in iOS simulator so far

src/status_im/ui/screens/views.cljs
```diff
@@ -35,6 +37,19 @@
...
       [bottom-sheet/bottom-sheet opts])))
 
+(views/defview inactive-overlay []
+  (letsubs [app-state [:get :app-state]]
+    {:component-will-mount
+     (fn []
+       (react/dismiss-keyboard!))
+     :component-will-update
+     (fn []
+       (react/dismiss-keyboard!))}
```
flexsurfer: just wondering why do we need to dismiss keyboard on an update?

flexsurfer (approved):

src/status_im/ui/screens/views.cljs
bitsikka: oh :) I was looking for all the cases where some UI elements would render above the blanked out component and see if I could deal with to not make it so. Keyboard was the only one I could deal with like that :) in that hacky way. Couldn't deal with `alert` and `share` as you can see

should I omit it?

I should probably also not use :get subscription considering `re-frame subscriptions optimisation`

yenda:
Have we considered this approach? https://stackoverflow.com/questions/9822076/how-do-i-prevent-android-taking-a-screenshot-when-my-app-goes-to-the-background

bitsikka:
> Have we considered this approach? https://stackoverflow.com/questions/9822076/how-do-i-prevent-android-taking-a-screenshot-when-my-app-goes-to-the-background

didn't know about this.. sounds more robust
will check it out

rasom:
yep seems that on android we need to try `FLAG_SECURE`

https://developer.android.com/reference/android/view/Display.html#FLAG_SECURE
https://github.com/kristiansorens/react-native-flag-secure-android

bitsikka:
> yep seems that on android we need to try `FLAG_SECURE`
> 
> https://developer.android.com/reference/android/view/Display.html#FLAG_SECURE
> https://github.com/kristiansorens/react-native-flag-secure-android

Thanks @rasom @yenda for the pointers

I barely got my development machine in order after having it being broken more than a week ago.

I am hardly familiar with lower React-Native / Android development / Java areas, but this will give me a chance to get to know it

Going to do the due diligence and make a proper implementation soon

Please bare with me

bitsikka:
@flexsurfer @yenda @rasom 

Changed the implementation as guided using FLAG_SECURE in android. Additionally, used SharedPreferences/NSUserDefaults to persist flag setting - additions are made to Status native-module.

@corpetty includes bonus (user setting) as requested

Tested in iOS Simulator and Android-avd:

- tested preview is blank by default
- tested preview is blank/not-blank according to user setting
- tested preview setting persists through app re-starts

### Notes:
- in Android (when preview privacy mode is on) prevents screen capture in non preview screen as well
- in iOS does not prevent screen capture
- in iOS if there is keyboard showing at the time of putting app to background, keyboard renders in preview above blank screen

flexsurfer:
@rasom @yenda ping

flexsurfer:
@mandrigin @rasom @yenda ping

ios/StatusIm/AppDelegate.m
```diff
@@ -101,4 +105,36 @@ - (NSURL *)sourceURLForBridge:(RCTBridge *)bridge
 #endif
 }
 
+- (void)applicationWillResignActive:(UIApplication *)application {
+  if ([[NSUserDefaults standardUserDefaults] boolForKey:@"BLANK_PREVIEW"]) {
+    // fill screen with our own colour
+    UIView *colourView = [[UIView alloc]initWithFrame:self.window.frame];
```
mandrigin: I think it is better to just save it as a local variable than a view with a tag.

android/app/src/main/java/im/status/ethereum/MainActivity.java
```diff
@@ -202,6 +203,18 @@ private void rejectRootedNotification() {
         editor.commit();
     }
 
+    private static final String BLANK_PREVIEW = "BLANK_PREVIEW";
```
mandrigin: hmm, the same constant declared twice kind of defeat the purpose of having a constant in the first place. Maybe we can just make a static class like `Constants` or `Preferences.BLANK_PREVIEW` and use it in both places.

mandrigin (changes requested):
a few small changes, good job otherwise!

StatusSceptre:
@flexsurfer @mandrigin - can we consider this PR complete so I can release the funds for the bounty?

android/app/src/main/java/im/status/ethereum/MainActivity.java
bitsikka: I think we can do without the constant. Changed the implementation a little bit in update that is coming shortly.

ios/StatusIm/AppDelegate.m
bitsikka: done

bitsikka:
@mandrigin updated PR with suggested changes.

Please note that at this point I am barely getting acquainted with Java and Objective-c and the implementation might still be sloppy. Hope the work is acceptable. If not, with feedback, still going to do what it takes.

Those (Java/Obj-c) parts have mostly been an adaptation(with prior knowledge of languages other than those) of code found in the interwebs; with finally reading documentations in the past 24 hrs :p.  Sorry about that and thanks for baring with me.

Also definitely going to continue to spend some time acquiring knowledge on them for knowledge sake.

mandrigin (approved):

bitsikka:
> An unknown server-side error occurred while processing the command. Original error: io.appium.uiautomator2.common.exceptions.TakeScreenshotException: Failed to capture a screenshot. Does the current view have 'secure' flag set? at io.appium.uiautoma

yes 'secure' flag is set

@antdanchenko @churik help! looks like e2e tests are going to have to be updated for this one as well

or the implementation is going to have to change if significant number of tests depend on capturing screenshot 🤔

mandrigin:
@bitsikka maybe we can have this option behind a build flag somehow? so it is disabled for e2e?

mandrigin:
@bitsikka or, easier, it should be `false` by default (it is `true` for Android in your code).

bitsikka:
> @bitsikka or, easier, it should be `false` by default (it is `true` for Android in your code).

yes, it is `true` by default for both Android and iOS(for iOS, just the blanking out in background state - not preventing screencap). @corpetty OP of the issue, requested it that it be so, and it made sense to do that, but we had no idea of this consequence

I'll make it `false` cause now, that also makes sense :)

churik:
Thanks, @bitsikka!

Tested IOS 11.4.1 (IPhone 7), Android 8 (LG V20):
- going to background with wallet, chat screens, with popup, with opened keyboard, on login screen - preview is blanked out;
- reopening app;

churik (approved):

PR #1444, opened 2017-07-13
chat: fix input length styling after commands

janherich:
[comment]: # (Please replace ... with your information. Remove < and >)
[comment]: # (To auto-close issue on merge, please insert the related issue number after # i.e fixes #566)
fixes #1354 

### Summary:
[comment]: # (Summarise the problem and how the pull request solves it)
This PR fixes overflowing characters after  any `/command` input if the text is too long and doesn't fit on one line together with `/command` prefix. 
The reason why it was happening was that minimal width of the input element for commands was set to 200, which together with `/command` width was often more then width of the wrapping container element.
Now the minimal width for the input element is computed by subtracting `/command` width from enclosing container-width.

### Steps to test:
- Open Status
- Open console
- Select `/confirmation-code` command and write some long number there, notice that it doesn't overflow over right edge (over `x` button) as before.

[comment]: # (PRs will only be accepted if squashed into single commit.)
status: ready

PR #8018, opened 2019-04-23
Update status-go to v0.24.0-beta.1

dshulyak:
Feature:

* Expose vanilla ECDSA signing (#1444)

Other fixes:

* If whisper peers are communicating and their time is not synced they will not disconnect from one another
* signal that history request was completed is sent only after all relevant envelopes were added to the whisper filters

Link to status-go release: https://github.com/status-im/status-go/releases/tag/v0.24.0-beta.1


mandrigin (approved):

cammellos (approved):

dshulyak:
Changed release version to v0.24.0-beta.1 . It includes one more important fix.

dshulyak:
I restarted tests. Doesn't seem like failure is caused by changes in this PR.
What else needs to be done before it will be merged?
@annadanchenko do you think that update needs to be tested separately before merge?

mandrigin:
@annadanchenko I marked it as a release blocker because it contains the fix for never-ending progressbar for private messages.

annadanchenko:
@dshulyak yes it should be tested before merge

annadanchenko:
@mandrigin @dshulyak when tested ok - let's included it in the next nearest release after 0.12.0

pedropombeiro:
`STATUS_GO_SHA256` also needs to be updated. FWIW, the official way to update status go is to use `scripts/update-status-go.sh` which takes care of everything, regardless of the implementation details.

Serhy:
No regressions found with Develop build.
@mandrigin could you merge it and cherrypick in release branch please?

PR #8021, opened 2019-04-23
[fix] wallet on-error dispatch

yenda:
the code to dispatch an event on transaction error was not working
this fixes it

status: ready

### DIFF
--- a/src/status_im/models/wallet.cljs
+++ b/src/status_im/models/wallet.cljs
@@ -160,22 +160,23 @@
           (update :gas-price str)
           (dissoc :message-id :id :gas)))))
 
-(defn handle-transaction-error [{:keys [db]} {:keys [code message]}]
+(fx/defn handle-transaction-error
+  [{:keys [db] :as cofx} {:keys [code message]}]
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

mandrigin (approved):

annadanchenko:
@yenda 
1. Windows build failed
2. To prevent side effects on release 0.12.0 can this PR be merged after we do code freeze (eod Wednesday) please? It's "skip qa" so we will not catch any side effects

goranjovic (approved):

yenda:
@annadanchenko windows build failure is just so infra error I'll restart it, regarding merge we can wait but it won't matter it is fixing something that was just not working but isn't used in the current app anyway

annadanchenko:
@yenda can you provide some details on how you've tested it, please? we might need to include such test into automated test suite - would be cool to get details for this

yenda:
@annadanchenko this is not something that can be tested as there is no code using it yet, and such code would have not worked as there was a bug that this PR fixes which is why this is skip-qa labelled.

PR #8024, opened 2019-04-23
fix token transactions

### LINKED ISSUE #7329: Outgoing token transactions in transaction history are shown as 0 ETH  after relogin
### Description

*Type*: Bug
*Summary*: after relogin all outgoing transactions in tokens are shown as 0 ETH, and only after several  minutes they are replaced with token transactions. This relates only to transactions made after restoring account. All others outgoing token transactions made before restoring account from seed phrase are shown as 0 ETH.
Comparing history from etherscan.io and from Status:
comp
The issue is similar to https://github.com/status-im/status-react/issues/6835 and it is reproducible on Ropsten and Mainnet

#### Expected behavior
can see transactions in tokens (similar to etherscan.io > ERC 20 token Txns)

#### Actual behavior
0 ETH is shown

### Reproduction

- Open Status
- Restore from seed phrase any account, that contains outgoing token transaction
- Send some tokens from the wallet 
- Relogin and check transaction history (check outgoing transaction made before and after restore)

### Additional Information

* Status version: nightly 16/01/2019

* Operating System: Android, IOS

yenda:
Fixes #7329 I would include this in release as not having token transactions shown at least for the last 14 days reliably is ridiculous

In #8005, token transactions were fetched more reliably but due to a bug, only inbound or outbound historical transactions would be fetched, and sometime some recent transactions would be skipped because the last block checked pointer was shared between inbound and outbound transaction

This fix makes sure that both inbound and outbound transactions are made from the same block, and removes an unecessary extra call to get-current-block

# Testing

check that both inbound and outbound historical (last 14 days) transactions are shown, check that new token transactions are shown as well

status: ready

### DIFF
--- a/src/status_im/models/transactions.cljs
+++ b/src/status_im/models/transactions.cljs
@@ -137,12 +137,11 @@
 (defn- get-token-transfer-logs
   ;; NOTE(goranjovic): here we use direct JSON-RPC calls to get event logs because of web3 event issues with infura
   ;; we still use web3 to get other data, such as block info
-  [web3 current-block-number chain-tokens direction address cb]
+  [web3 current-block-number from-block chain-tokens direction address cb]
   {:pre [web3 (integer? current-block-number) (map? chain-tokens) (keyword? direction) (string? address) (fn? cb)]}
   (let [[from to] (if (= :inbound direction)
                     [nil (add-padding (ethereum/normalized-address address))]
                     [(add-padding (ethereum/normalized-address address)) nil])
-        from-block (limited-from-block current-block-number @latest-block-checked)
         args {:jsonrpc "2.0"
               :id      2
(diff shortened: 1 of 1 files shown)


goranjovic (approved):

mandrigin (approved):

annadanchenko:
@yenda ios and android-e2e builds are failing

churik:
@yenda thank you!
It works.
Checked (on Mainnet and Ropsten):
- restore the wallet with token and ETH transactions (all token transactions during last 14 days are shown normally and match Etherscan)
- send several transactions in tokens \ ETH and check incoming and outgoing. Check that transaction are shown normally after relogin.
- failed tests was checked manually - it works.
Failed due to infrastracture error.

PR #8025, opened 2019-04-24
Maintain scroll position after fetching gap

rasom:
1. previously the gap was shown before the first message which has `timestamp` higher than gap's `from`; in this PR the gap is shown after all messages which `timestamp` is lower than gap, also after messages which are inside the gap and before the first message which `timestamp` is higher than gap's `to`. 
    before:
    ```
    message A (T=10)
    gap (20-30)
    message B (25)
    message C (35)
    ```

    after:
    ```
    message A (T=10)
    message B (25)
    gap (20-30)
    message C (35)
    ```

    This minimizes chances that requested messages will be rendered between messages B and C instead of rendering between gap and C. In result, we can scroll to message C before loading and bottom side of the chat will maintain the same position after fetching messages (no guaranty but higher chances).

2. Before fetching the gap, the list is scrolled so that the next message after the gap (message C in the example above) is placed in the center of the screen. In this case, after rendering of the newly fetched messages, the user will still be able to see older, already seen messages along with new rendered messages which will replace gap above that "message C"

In some cases the app will behave properly without these changes, it depends on a number of referenced messages which are fetched and have to be rendered in responses. Also, messages like `B` might not be present in the chat's history.

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/views.cljs
+++ b/src/status_im/ui/screens/chat/views.cljs
@@ -70,36 +70,45 @@
   [{{:keys [value]} :row}]
   [message-datemark/chat-datemark-mobile value])
 
-(defview gap [{:keys [ids]}]
+(defview gap
+  [ids idx list-ref in-progress? connected?]
+  [react/view {:align-self          :stretch
+               :margin-top          24
+               :margin-bottom       24
+               :height              48
+               :align-items         :center
+               :justify-content     :center
+               :border-color        colors/gray-light
+               :border-top-width    1
+               :border-bottom-width 1
+               :background-color    :white}
+   [react/touchable-highlight
+    {:on-press (when (and connected? (not in-progress?))
+                 #(do
+                    (when @list-ref
+                      (.scrollToIndex @list-ref #js {:index        (max 0 (dec idx))
+                                                     :viewOffset   20
+                                                     :viewPosition 0.5}))
+                    (re-frame/dispatch [:chat.ui/fill-gaps ids])))}
+    [react/view {:flex            1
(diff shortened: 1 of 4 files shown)


cammellos (approved):

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -70,36 +70,45 @@
   [{{:keys [value]} :row}]
   [message-datemark/chat-datemark-mobile value])
 
-(defview gap [{:keys [ids]}]
+(defview gap
+  [ids idx list-ref in-progress? connected?]
+  [react/view {:align-self          :stretch
```
flexsurfer, line 75: move to styles?

flexsurfer (approved):

src/status_im/ui/screens/chat/views.cljs
rasom: ehm, let me finish with desktop first, and other related PRs. I'm sure it will be changed 2-3 times during next PRs. It's just simplier to keep everything in one place for now

rasom:
@hesterbruikman that's the best I can propose without spending too much time on this task

churik:
Tested when "Fetching messages" (IOS 11.4.1, IPhone 7, Android 8, LG V20) is close to the top and to the bottom of the screen, and somewhere in the middle.
Looks much reliable with this small automatic scroll next message after the gap is always visible to me, so when fetching messages always can see the last message after a gap on the middle of the screen.

Ready to merge from my after e2e

churik:
failed e2e is due to an infrastructure issue, not related to PR

Issue #6946, opened 2018-11-29
You can send message to the past if you set custom date


churik:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story

As a user I want to see messages in order that they were sent

### Description
*Type*: Bug
*Summary*: if you set date on sender side to yesterday manually and send a message, receiver will see it below "Yesterday", so order is wrong.

#### Expected behavior
Message is last

#### Actual behavior
Message is shown as sent yesterday (so below "Yesterday")
![kkk](https://user-images.githubusercontent.com/4557972/49233127-9fe62f80-f3f5-11e8-9a24-26c9a1928b14.png)

### Reproduction

- Open Status on device A
- Join any chat (i.e. #test)
- Send some message to it
- Open Status on device B
- Set device B date to yesterday
- Send some message 

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: nightly 29/11/2018
* Operating System: all

lukaszfryc:
duplicate of https://github.com/status-im/status-react/issues/5605?

churik:
Yes, it is closed, so couldn't find it. So reopen #5605 or leave this one?

lukaszfryc:
We can leave this one

Swader:
Still a thing

churik:
Everyday test "Can join test public chat and send/recieve text messages" https://ethstatus.testrail.net/index.php?/tests/view/705822 is failed in nightly because of this issue.
The reason is that device and runner are located in different timezones - so test is seeking message below "Today", in fact it is located under yesterday.
Logs/video: https://app.saucelabs.com/tests/ddf863e56f544cf5896136a714fd771d?auth=c26b016cb28c842c3e03e815e7b0c6e4

churik:
cc @rachelhamlin 
From my POV it is not acceptable to have the possibility to send a message in the past and change chat history.

rachelhamlin:
What are the reasons a user would change the date to the past on their device? Is this a real possibility?

churik:
@rachelhamlin  just to fake chat history.
Normal user shouldn't do this.
I'm not sure about timezones - but can investigate it.

rachelhamlin:
Mmm good point. Thanks @churik. Let's see if anything comes up about this during the audit.

rachelhamlin:
Requesting @cammellos or @adambabik to take a look into this one to see if it's a) a go fix and b) a straightforward one 🤞

[...]

Issue #6419, opened 2018-10-19
Timestamps for messages doesn't match real time if user send message from device with custom time


churik:
### Description

*Type*: Bug
*Summary*:  when anybody send a message from device with custom time (not another timezone, but custom time, i.e. when there is 10:00 UTC, and user set 09:45) - timestamps for such messages are shown with difference in minutes from nearest timezone, so in chat it looks like wrong message ordering (but it is not, wrong are message timestamps)

#### Expected behavior
timestamps for messages are set according to real message ordering and receiver time settings;

#### Actual behavior
timestamps for messages sent from device with custom time don't match real time
<img width="568" alt="us_t" src="https://user-images.githubusercontent.com/4557972/47223441-9494f280-d3b9-11e8-88d6-98bb1bd9a4db.png">

### Reproduction

*Prerequisites:* for `User A` timezone and time is set automatically; for `User B` time is custom and set 10 minutes earlier)

- `User A`: open Status, join any public channel (i.e. #chuone), send message 
- `User B`: open Status, join any public channel (i.e. #chuone), send message 
- `User A`: check chat history

### Additional Information
[comment]: # (Please do your best to fill this out.)
* Status version: nightly 19/10/2018
* Operating System: Android, iOS, desktop

SlizkovaGanna:
Reproduces for me also

<img width="668" alt="image" src="https://user-images.githubusercontent.com/43638377/49367680-6ebb7700-f6f5-11e8-96ca-4e14ddbe2321.png">

0xc1c4da:
➤ Hester Bruikman commented:

Data sync might improve situation; still inherent to peer-to-peer network.

Can give impression that connectivity is lost even though messages was received, but message order got displaced. How to make sure it's clear that a dated messages is received.

Compare timestamps (local + Whisper timestamp). Define behavior for message based on threshold while preventing 'faking' of historic messages.

churik:
Not reproducible anymore

PR #8033, opened 2019-04-24
Allow to fetch more history in public chats (up to 30 days)

rasom:
Please review only the last commit, merge after #8025

status: ready

### DIFF
--- a/src/status_im/chat/db.cljs
+++ b/src/status_im/chat/db.cljs
@@ -6,7 +6,8 @@
             [status-im.chat.commands.input :as commands.input]
             [status-im.group-chats.db :as group-chats.db]
             [status-im.utils.gfycat.core :as gfycat]
-            [status-im.transport.partitioned-topic :as topic]))
+            [status-im.transport.partitioned-topic :as topic]
+            [status-im.mailserver.core :as mailserver]))
 
 (defn group-chat-name
   [{:keys [public? name]}]
@@ -86,10 +87,14 @@
 (defn datemark? [{:keys [type]}]
   (= type :datemark))
 
+(defn gap? [{:keys [type]}]
+  (= type :gap))
+
 (defn transform-message
   [messages message-statuses referenced-messages]
   (fn [{:keys [message-id timestamp-str] :as reference}]
-    (if (datemark? reference)
+    (if (or (datemark? reference)
+            (gap? reference))
       reference
       (let [{:keys [content] :as message} (get messages message-id)
             {:keys [response-to response-to-v2]} content
@@ -141,18 +146,31 @@
 (defn messages-with-datemarks-and-statuses
   "Converts message groups into sequence of messages interspersed with datemarks,
   with correct user statuses associated into message"
-  [message-groups messages message-statuses referenced-messages messages-gaps]
+  [message-groups messages message-statuses referenced-messages messages-gaps
+   {:keys [highest-request-to lowest-request-from]} all-loaded? public?]
   (transduce
    (comp
     (mapcat add-datemark)
     (map (transform-message messages message-statuses referenced-messages)))
-   (completing
-    (fn [{:keys [messages datemark-reference previous-message gaps]}
-         message]
+   (fn
+     ([]
+      (let [acc {:messages         (list)
+                 :previous-message nil
+                 :gaps             messages-gaps}]
+        (if (and
+             public?
+             all-loaded?
+             (< (- highest-request-to lowest-request-from)
+                mailserver/max-gaps-range))
+          (update acc :messages conj {:type       :gap
+                                      :value      (str :first-gap)
+                                      :first-gap? true})
+          acc)))
+     ([{:keys [messages datemark-reference previous-message gaps]} message]
       (let [new-datemark? (datemark? message)
             {:keys [gaps-number gap]}
             (check-gap gaps previous-message message)
-            add-gap? (pos? gaps-number)]
+            add-gap?      (pos? gaps-number)]
         {:messages           (cond-> messages
 
                                add-gap?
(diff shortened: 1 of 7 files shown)


hesterbruikman (approved):
- I feel like I have my memory back being able to go into history! Great work! 
- I'll create that issue to polish and animate new messages being loaded. 
- I believe we decided to keep 'Fetch more messages' rather than load to maintain consistency. But I prefer load as it's more conventional. I'll check other places we need to update to Load for consistency.

goranjovic (approved):

rasom:
@hesterbruikman it is still "Fetch more messages" for gaps which appear between history requests. I just wanted to emphasize that this one is rather about **loading/fetching of older messages** than about **fetching of missed messages**. It looks and works similarly but that's two different cases imo. Let me know if it should be changed.

hesterbruikman:
Thanks @rasom. Let's stick with "Fetch more messages" for both, for now. It is the most accurate capture of what happens.  The more I think about it 'Load' might set the wrong expectations.

In time, a more suitable solution for *older messages* would be to fetch on scroll, which I realize comes with it's own challenges. It would however require less effort from the user to mentally process as it's a response to existing behavior  rather than demand for something extra ('tap this button'). Cc @errorist

rasom:
> a more suitable solution for older messages would be to fetch on scroll,

I would add fetching by scroll separately and we will still need to have current ui, for the case when fetching history is disabled by default on mobile connection. Because this fetching comes with some traffic cost

yenda:
This features relies on mailservers being able to store 30 days of all whisper traffic doesn't it? How much is it right now? How does it grows with increased user base?

cammellos:
@yenda it's 30 days currently I believe, I don't know how much it grows with current user base as I don't know how large it is, but at the moment 30 days worth of data are around 2.2/2.5 GB

jakubgs:
Correct, around `2.6 GB`:
```
 $ ansible mail -o -a 'du -hs /docker/statusd-mail/data/wnode'
mail-03.do-ams3.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
mail-02.do-ams3.eth.beta | CHANGED | rc=0 | (stdout) 2.4G	/docker/statusd-mail/data/wnode
mail-01.do-ams3.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
mail-02.gc-us-central1-a.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
mail-03.ac-cn-hongkong-c.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
mail-02.ac-cn-hongkong-c.eth.beta | CHANGED | rc=0 | (stdout) 2.3G	/docker/statusd-mail/data/wnode
mail-01.gc-us-central1-a.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
mail-01.ac-cn-hongkong-c.eth.beta | CHANGED | rc=0 | (stdout) 2.4G	/docker/statusd-mail/data/wnode
mail-03.gc-us-central1-a.eth.beta | CHANGED | rc=0 | (stdout) 2.6G	/docker/statusd-mail/data/wnode
```
And it is indeed 30 days:
https://github.com/status-im/infra-eth-cluster/blob/c6a847fefa7267a731ac1e0eee638fdad2c6be00/ansible/roles/statusd-mailsrv/defaults/main.yml#L57

But considering the amount of space left we could easily double it. Though of course the more users we get the bigger it will grow over time.

churik:
@rasom thanks for additional  "Fetch 48-60h", "Fetch 84-96h", you can remove it now.

1) Just FYI: https://github.com/status-im/status-react/issues/6946 and https://github.com/status-im/status-react/issues/6419 are very visible in this PR - because if you have a message out of order after fetching the previous 24h you will be redirected to last message according to chat order.
I would prefer to see then fixed if it is possible along with this PR or at least put high-priority after this PR.
What do you think?
Video (example with wrong timestamps): https://take.ms/zggUk

2) On month history also more visible problems with scrolling, i.e. :
- 2.1 blinking of the message that starts history when scrolling up
ezgif com-crop

- 2.2 that messages are not rendering quickly enough when scrolling fast
ezgif com-video-to-gif (1)

- 2.3 that there is no way to be scrolled fast to the last message from the middle of the chat.

I know, that it is not about this particular PR right now, just want to emphasize that these problems will be more visible if we decide to merge this PR now.

Tested on IOS and Android:
- history can be fetched during the last month, day by day on `ams3-do` and `mail-03-hongkong` mailservers. Didn't notice any spikes.
- history matches history on desktop (checked for several days)
- "Fetch messages" (gaps in history) and feature from PR are working fine together
- can't  fetch messages when offline
- messages fetching normally while offline->online during fetching
- the feature is available only for public chats.
- the desktop build is not affected (checked Mac OSx)

@rachelhamlin @annadanchenko let me know if you think we should include this PR in 0.12.0.
IMO we can (but taking into account issues 1 and 2)

rasom:
> I know, that it is not about this particular PR right now, just want to emphasize that these problems will be more visible if we decide to merge this PR now.

both problems are out of the scope of this PR. That behavior on video doesn't seem to be a problem imo. Regarding slow scroll - it's a problem for anyone who used app regularly for more than month. I would say that ability to fetch history gives more advantages than slow scrolling disadvantages (assuming that's only a revealing of an existing issue).

mandrigin:
> I would say that ability to fetch history gives more advantages than slow scrolling disadvantages 

I agree with that too

rachelhamlin:
I don't think the time stamp issues (#6946, #6419) or slow scroll should block this PR—user has to manually change or have custom time on their device? Sounds far out. But whether to include this PR in the current release is test team's decision. Very nice to have, imo, not a need. Nice work @rasom.

PR #8034, opened 2019-04-25
updating dev dependencies

yenda:
update cider and figwheel (more importantly use release of figwheel instead of snapshot)

status: ready

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -21,16 +21,16 @@
                  :sha "144eeecfb389edd9b5d4d94507acf828c5265b97"}
 
          ;; Figwheel ClojureScript REPL
-         cider/piggieback        {:mvn/version "0.3.9"
+         cider/piggieback        {:mvn/version "0.4.0"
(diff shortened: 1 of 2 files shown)


flexsurfer:
why do we need project.clj ? and why don't we use `figwheel-main` instead of sidecar?

jeluard (approved):

yenda:
@flexsurfer I think desktop is still relying on project.clj? @vkjr 
regarding figwheel I don't know, I just want to update the deps because the nrepl version is incompatible with latest cider

rasom (approved):

vkjr:
@yenda, do I get correctly that `project.clj` was needed only for building project with `lein`? 
`clj-rn` already supports desktop platform, so desktop can be built without project.clj.

But I see in a Makefile that lein is still mentioned in `prod-build-ios`, `prod-build-android`, `prod-build-desktop` commands.

PR #8022, opened 2019-04-23
Feature fix/ttt settings

yenda:
Splitted the final touches to the settings part of ttt to limit scope of the final ttt PR

- settings are stored in a manifest that is pointed at by the contract
- during login the contract is checked to fetch the settings

### Steps to test

On Ropsten on an account with some test eth, go to settings and set up tribute to talk, try recovering the account on another device (should fetch the settings), try changing network (shouldn't keep settings).

status: ready

PR #8035, opened 2019-04-25
Refactor/move block contact

yenda:
this will avoid circular dependencies in future Tribute to Talk commits

# testing

Block user feature can be affected, though no code is changed it is just moved

status: ready


annadanchenko:
@yenda if it's related to refactoring block user functionality then I would recommend not to skip manual qa

vitvly (approved):

yenda:
@annadanchenko this is just moving some code from one ns to another, it looks big because #8022 needs to be merged first

yenda:
@flexsurfer can you have a look please?

flexsurfer (approved):

PR #8039, opened 2019-04-25
workaround issue with building iOS release builds

mandrigin:
no app code is affected, and it solves "[Symbols tool failed](https://stackoverflow.com/questions/26882037/symbols-tool-failed-error-while-exporting-iphone-application-with-app-store-pr)" issue.

status: ready

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -217,6 +217,7 @@ platform :ios do
         clean: true,
         export_method: "app-store",
         output_directory: "status_appstore",
+        include_symbols: false,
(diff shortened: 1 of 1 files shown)


mandrigin:
I think we can skill **all** QA, not only manual for this one.

jakubgs (approved):
NICE

jakubgs:
Looks like you forgot to sign the commit.

mandrigin:
whatever, it will be re-signed on merge anyway

rasom (approved):

PR #8042, opened 2019-04-25
Fixed broken EIP1577 support

jeluard:
### Summary

Fixes broken EIP1577 support.

`Note to Testers` Browse `pac-txt.eth` , it should load fine (but the page itself has an error)

status: ready

### DIFF
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -12,6 +12,7 @@
             [status-im.utils.ethereum.core :as ethereum]
             [status-im.utils.ethereum.ens :as ens]
             [status-im.utils.ethereum.resolver :as resolver]
+            [status-im.utils.contenthash :as contenthash]
             [status-im.utils.fx :as fx]
             [status-im.utils.http :as http]
             [status-im.utils.multihash :as multihash]
@@ -66,17 +67,21 @@
   (let [history-host (http/url-host (try (nth history history-index) (catch js/Error _)))]
     (cond-> browser history-host (assoc :unsafe? (dependencies/phishing-detect history-host)))))
 
+(defn- content->hash [hex]
+  (when (and hex (not= hex "0x"))
+    ;; TODO(julien) Remove once our ENS DApp are migrated
+    (multihash/base58 (multihash/create :sha2-256 (subs hex 2)))))
+
 (defn resolve-ens-content-callback [hex]
-  (let [hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 2))))]
+  (let [hash (content->hash hex)]
     (if (and hash (not= hash resolver/default-hash))
-      (re-frame/dispatch [:browser.callback/resolve-ens-multihash-success constants/ipfs-proto-code hash])
+      (re-frame/dispatch [:browser.callback/resolve-ens-multihash-success {:namespace :ipfs :hash hash}])
       (re-frame/dispatch [:browser.callback/resolve-ens-contenthash]))))
 
 (defn resolve-ens-contenthash-callback [hex]
-  (let [proto-code (when hex (subs hex 2 4))
-        hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 12))))]
-    (if (and proto-code (#{constants/swarm-proto-code constants/ipfs-proto-code} proto-code) hash (not= hash resolver/default-hash))
-      (re-frame/dispatch [:browser.callback/resolve-ens-multihash-success proto-code hash])
+  (let [{:keys [hash] :as m} (contenthash/decode hex)]
+    (if (and hash (not= hash resolver/default-hash))
+      (re-frame/dispatch [:browser.callback/resolve-ens-multihash-success m])
       (re-frame/dispatch [:browser.callback/resolve-ens-multihash-error]))))
 
 (fx/defn resolve-url
@@ -150,21 +155,29 @@
                              :history new-history
                              :history-index new-index)))))
 
+(defmulti storage-gateway :namespace)
+
+(defmethod storage-gateway :ipfs
+  [{:keys [hash]}]
+  (let [base32hash (-> (.encode js-dependencies/hi-base32 (alphabase.base58/decode hash))
+                       (string/replace #"=" "")
+                       (string/lower-case))]
+    (str "https://" base32hash ".infura.status.im")))
+
+(defmethod storage-gateway :swarm
+  [{:keys [hash]}]
+  (str "https://swarm-gateways.net/bzz:/" hash))
+
 (fx/defn resolve-ens-multihash-success
-  [{:keys [db] :as cofx} proto-code hash]
+  [{:keys [db] :as cofx} m]
   (let [current-url (get-current-url (get-current-browser db))
-        host (http/url-host current-url)
-        path  (subs current-url (+ (.indexOf current-url host) (count host)))
-        gateway (if (= constants/ipfs-proto-code proto-code)
-                  (let [base32hash (-> (.encode js-dependencies/hi-base32 (alphabase.base58/decode hash))
-                                       (string/replace #"=" "")
-                                       (string/lower-case))]
-                    (str base32hash ".infura.status.im"))
-                  (str "swarm-gateways.net/bzz:/" hash))]
+        host        (http/url-host current-url)
+        path        (subs current-url (+ (.indexOf current-url host) (count host)))
+        gateway     (storage-gateway m)]
     (fx/merge cofx
               {:db (-> (update db :browser/options
                                assoc
-                               :url (str "https://" gateway path)
+                               :url (str gateway path)
                                :resolving? false)
                        (assoc-in [:browser/options :resolved-ens host] gateway))})))
 
--- a/src/status_im/utils/contenthash.cljs
+++ b/src/status_im/utils/contenthash.cljs
@@ -15,15 +15,15 @@
              (nil? ipld))
     (str "0xe301" (hex/encode (b58/decode hash)))))
 
-(defn decode [contenthash]
-  (when (and contenthash
-             (string/starts-with? contenthash "0xe3011220")
-             (= 74 (count contenthash)))
+(defn decode [hex]
+  (when (and hex (not= hex "0x")
+             (string/starts-with? hex "0xe30101")
+             (= 78 (count hex)))
     {:namespace :ipfs
-     :hash (-> contenthash
-               (subs 6)
-               hex/decode
-               b58/encode)}))
+     :hash  (-> hex
+                (subs 10)
+                hex/decode
+                b58/encode)}))
 
 (fx/defn cat
   [cofx {:keys [contenthash on-success on-failure]}]
--- a/test/cljs/status_im/test/utils/contenthash.cljs
+++ b/test/cljs/status_im/test/utils/contenthash.cljs
@@ -7,7 +7,7 @@
 
 (deftest contenthash-decode
   (testing "decoding a valid ipfs hash"
-    (is (= (contenthash/decode "0xe301122029f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f")
+    (is (= (contenthash/decode "0xe3010170122029f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f")
            {:namespace :ipfs
             :hash "QmRAQB6YaCyidP37UdDnjFY5vQuiBrcqdyoW1CuDgwxkD4"})))
   (testing "decoding an invalid ipfs hash"
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -1502,8 +1502,8 @@
 
 (handlers/register-handler-fx
  :browser.callback/resolve-ens-multihash-success
- (fn [cofx [_ proto-code hash]]
-   (browser/resolve-ens-multihash-success cofx proto-code hash)))
+ (fn [cofx [_ m]]
+   (browser/resolve-ens-multihash-success cofx m)))
 
 (handlers/register-handler-fx
  :browser.callback/resolve-ens-multihash-error
--- a/src/status_im/utils/ethereum/ens.cljs
+++ b/src/status_im/utils/ethereum/ens.cljs
@@ -6,7 +6,8 @@
   "
   (:refer-clojure :exclude [name])
   (:require [clojure.string :as string]
-            [status-im.utils.ethereum.core :as ethereum]))
+            [status-im.utils.ethereum.core :as ethereum]
+            [status-im.utils.ethereum.abi-spec :as abi-spec]))
 
 ;; this is the addresses of ens registries for the different networks
 (def ens-registries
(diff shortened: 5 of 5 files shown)


src/status_im/browser/core.cljs
```diff
@@ -67,14 +67,14 @@
...
-  (let [hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 2))))]
+  (let [hash (when (and hex (not= hex "0x")) (multihash/base58 (multihash/create :sha2-256 (subs hex 2))))]
     (if (and hash (not= hash resolver/default-hash))
       (re-frame/dispatch [:browser.callback/resolve-ens-multihash-success constants/ipfs-proto-code hash])
       (re-frame/dispatch [:browser.callback/resolve-ens-contenthash]))))
 
 (defn resolve-ens-contenthash-callback [hex]
   (let [proto-code (when hex (subs hex 2 4))
-        hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 12))))]
+        hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 14))))]
```
goranjovic: out of curiosity, what are the extra two chars? is it '0x'?

goranjovic (approved):

src/status_im/browser/core.cljs
jeluard: No just data used to identify various algo used, that we ignore here.

yenda (changes requested):
support for EIP1577 should be handled in `utils/contenthash` and not only browser ens resolution. There is already basic support for ens and adding swarm should be trivial as the only difference is the protocol code `e4` instead of `e3`

jeluard:
@yenda @flexsurfer I might switch to https://github.com/pldespaigne/content-hash , they handle the parsing more cleanly

yenda:
> @yenda @flexsurfer I might switch to https://github.com/pldespaigne/content-hash , they handle the parsing more cleanly

Not sure it does it more cleanly but for sure it does it more completely as the current version of contenthash namespace only has support for basic ipfs sources. The idea was to improve it over time as we support more codecs like we did for abi-spec, and fork it as a library at some point in the future.
If you really don't feel like extending it and prefer strongly to use the js lib then go for it

src/status_im/utils/contenthash.cljs
```diff
@@ -15,21 +13,12 @@
...
+  (when (and hex (not= hex "0x")
+             (string/starts-with? hex "0xe30101701220")
+             (= 78 (count hex)))
     {:namespace :ipfs
-     :hash (-> contenthash
-               (subs 6)
-               hex/decode
-               b58/encode)}))
-
-(fx/defn cat
```
yenda: why did you remove this? it is needed for tribute to talk

yenda (changes requested):

src/status_im/browser/core.cljs
```diff
@@ -66,17 +67,21 @@
...
     (cond-> browser history-host (assoc :unsafe? (dependencies/phishing-detect history-host)))))
 
+(defn- content->hash [hex]
+  (when (and hex (not= hex "0x"))
+    ;; TODO(julien) Remove once our ENS DApp are migrated
+    (multihash/base58 (multihash/create :sha2-256 (subs hex 2)))))
+
 (defn resolve-ens-content-callback [hex]
-  (let [hash (when hex (multihash/base58 (multihash/create :sha2-256 (subs hex 2))))]
+  (let [hash (content->hash hex)]
```
yenda, line 76: why not use contenthash/encode here?

src/status_im/browser/core.cljs
```diff
@@ -150,21 +155,29 @@
...
                              :history-index new-index)))))
 
+(defmulti storage-gateway :namespace)
+
+(defmethod storage-gateway :ipfs
+  [{:keys [hash]}]
+  (let [base32hash (-> (.encode js-dependencies/hi-base32 (alphabase.base58/decode hash))
+                       (string/replace #"=" "")
+                       (string/lower-case))]
+    (str "https://" base32hash ".infura.status.im")))
```
yenda, line 165: what is this gateway should it be used for all content? there is already a gateway defined in ipfs.core

src/status_im/browser/core.cljs
jeluard: It's not the same format, no header

src/status_im/browser/core.cljs
jeluard: It's necessary for websites. It introduces subdomains to ensure no cookies/localStorage are shared.

src/status_im/utils/contenthash.cljs
```diff
@@ -15,21 +13,12 @@
              (nil? ipld))
     (str "0xe301" (hex/encode (b58/decode hash)))))
 
-(defn decode [contenthash]
-  (when (and contenthash
-             (string/starts-with? contenthash "0xe3011220")
-             (= 74 (count contenthash)))
+(defn decode [hex]
+  (when (and hex (not= hex "0x")
+             (string/starts-with? hex "0xe30101701220")
```
yenda: I doubt that all of the contenthash we are decoding have content type `dag-pb`

src/status_im/utils/contenthash.cljs
jeluard: I have not seen it used. Will restore.

test/cljs/status_im/test/utils/contenthash.cljs
```diff
@@ -7,7 +7,7 @@
 
 (deftest contenthash-decode
   (testing "decoding a valid ipfs hash"
-    (is (= (contenthash/decode "0xe301122029f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f")
+    (is (= (contenthash/decode "0xe3010170122029f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f")
```
yenda, line 10: previous test was fine, the extra 01 is a typo in EIP and 70 is certainly not the content type we want for all our decode files

src/status_im/browser/core.cljs
```diff
@@ -150,21 +155,29 @@
...
+
+(defmethod storage-gateway :swarm
```
yenda, line 167: it seems like you didn't add support for swarm in contenthash ns so is it necessary?

jeluard: Not harmful! I am sure you are excited to add swarm support for TtT :)

src/status_im/utils/contenthash.cljs
jeluard: That's the only one described in the spec. So far only one website follows the spec, `pac-txt.eth`

We can revisite later

(besides you implemented that part)

src/status_im/browser/core.cljs
yenda: yes but definitely not like this from the `browser.core` namespace

test/cljs/status_im/test/utils/contenthash.cljs
jeluard: Not it's not I double checked with spec writer.

src/status_im/utils/contenthash.cljs
yenda: no it's not, the spec is https://github.com/multiformats/multicodec/blob/master/table.csv
the EIP is just giving an exemple and it has typos

test/cljs/status_im/test/utils/contenthash.cljs
yenda: did he provide an explanation for the extra 01 then?

src/status_im/utils/contenthash.cljs
yenda: I can't accept the PR with this change 70 is wrong for sure

yenda (dismissed):
https://github.com/multiformats/multicodec/blob/master/table.csv#L38

src/status_im/utils/contenthash.cljs
yenda: @decanus could you clarify please?

jeluard:
Passing it to @yenda (aka SuperDupont) as he really doesn't enjoy those `01` :slightly_smiling_face:

test/cljs/status_im/test/utils/contenthash.cljs
decanus: The second 01 is variable length @yenda

yenda: @decanus could you point me to the spec that specify that I can't find it. Is it specific to EIP1577? It makes the contenthash unreadable with a tool like ipfs CID inspector https://cid.ipfs.io/#f017010122029f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f

yenda: providing variable length isn't necessary because multicodecs codes are distributed so that they don't overlap, but introducing a variable length could introduce ambiguity

decanus: I'm not 100% certain, let me get @arachnid in

Arachnid: Per multiformats/multicodec:

> multicodec is a self-describing multiformat, it wraps other formats with a tiny bit of self-description. A multicodec identifier is a varint.

The varint encoding for 0xe3 (IPFS) is 0xe301 - 'e30101' in the EIP1577 example is multicodec value 0xe3, followed by CID version 0x01.

yenda: @Arachnid why is the the varint encoding done like this? Also it is not consistent between ipfs and swarm examples:

`0x e30101 (ipfs) 70  (should be 7001 with that encoding?) 12 20 (hashing algo and length) 29f2d17be6139079dc48696d1f582a8530eb9805b561eda517e22a892c7e3f1f (hash)`

`0x e40101 (swarm) fa01 (swarm manifest) 1b 20 (hashing algo and length) d1de9994b4d039f6548d191eb26786769f580809256b4685ef316805265ea162 (hash)`

This way of encoding is not compatible with the cid inspector mentionned in the spec https://cid.ipfs.io/#f01701b20d1de9994b4d039f6548d191eb26786769f580809256b4685ef316805265ea162 and does not correspond to the exemples I could find in implementations of multiformat. 

For instance:
- https://github.com/multiformats/js-cid has
`zdj7WkRPAX9o9nb9zPbXzwG7JEs78uyhwbUs8JSUayB98DWWY`
which is
`0x01701220DEBD7ADB3CE56544D22A6F6B93396F6980A8067C2CC134F0F7801B6331092B95`

Arachnid: The use of varints is described in the multicodec spec: https://github.com/multiformats/multicodec. Any compliant multicodec encoder/decoder should handle it.

Multicodec values are not CIDs, but these particular ones contain a CID. If you strip off the varint prefix, they should decode as a CID.

The CID you link to appears to be valid. I'm not sure what you're trying to demonstrate with that link?

I'm also not sure what's inconsistent about the IPFS and Swarm examples; can you clarify?

yenda: @Arachnid Ok so after experimenting more with cid.ipfs.io and the cid-js library I can confirm that I was wrong and the EIP examples are correct and compatible with the different implementations. 

I confused myself because the varint encoding of `70` is `70` and after looking at the protobuf doc https://developers.google.com/protocol-buffers/docs/encoding#varints I understand how it is done and why the encoding of `e3` is `e301`. Thank you for your patience.

yenda:
@flexsurfer @jeluard so if you are interested the protocodes for the codecs in https://github.com/multiformats/multicodec/blob/master/table.csv are in hex but need to be converted to varint before being used in a contenthash in hex again. 

This means `70` stays `70` but `e3` becomes `e301` for instance. 

The reference for varints is https://developers.google.com/protocol-buffers/docs/encoding#varints and an implementation is https://github.com/chrisdickinson/varint#readme

test/cljs/status_im/test/utils/contenthash.cljs
jeluard: @Arachnid Thanks for the clarification! Is the intention to allow support for various content-types and hash functions? Or shall we treat the examples as normative?

Arachnid: @yenda No worries!

@jeluard Certainly - as many as you care to support. The examples are just the ones we know of so far.

flexsurfer (approved):

jeluard:
@churik You can test using http://condescending-jones-85ad1f.netlify.com
Click on the link on a regular mobile browser, it should open status wallet screen.

churik:
@jeluard some changes are requested, please move it to `TO TEST` once all will be resolved.
Also please share how to test this PR and platforms \ screenshots where you tested it.
Thank you.

dmitryn (approved):

churik:
@jeluard rebase to current nightly to fix e2e build please, thanks.

jeluard:
@churik Done

churik:
Tested on IOS 11.4.1 (IPhone 7), Android 8 (LG V20) as mentioned in the description.

test/cljs/status_im/test/utils/contenthash.cljs
yenda: @Arachnid in CIDv1 it seems like the content-type isn't optional, however the current codecs for content-type are quite limited, for instance nothings really fits for a dapp in html5. Isn't it going to be a problem for adoption of the EIP?

yenda: actually it doesn't seem to be a goal of multiformat to support these kind of content-types:

Q. What formats (multicodec codes) does CID support?

`We are figuring this out at this time. It will likely be a table of formats for secure distributed systems. So far, we want to address: IPFS's original protobuf format, the new IPLD CBOR format, git, bitcoin, and ethereum object`

Arachnid: To be clear, the content-type field here specifies the type of the hash - in the case of IPFS it's `dag-pb` - not the type of the content it references.

yenda: ok so if I understand correctly for IPFS contenthashes when we use the infura gateway with the simple `add` method instead of `dag_put` we have `raw 0x55` content-type? And same applies for swarm when uploading a file directly without a manifest?

edit: ok I figured there is a cid-version option for the `add` post method on infura and it returns for instance "zb2rhZfjRh2FHHB2RkHVEvL2vJnCTcu7kwRqgVsf9gpkLgteo" which has content type `raw 0x55` so I understand how it works now. Thanks!

Arachnid: That sounds right - but since IPFS returns CIDs, you should simply be able to use the CID it gives you, and let it worry about the content type.

yenda: yes that's what I do now, I added the `cid-version` option to our call to `add` so that it returns a b58 encoded CID, then I just have to convert it to hex and prepend "0xe301" to get the contenthash

PR #8055, opened 2019-04-26
[fix 8054] network in app-db is not always correct

yenda:
- when login in after a different account using a different network
the new account inherited the network from the previous one
- this fixes it by using the account network when initializing app-db
instead of reusing current network

status: ready

### DIFF
--- a/src/status_im/init/core.cljs
+++ b/src/status_im/init/core.cljs
@@ -188,7 +188,7 @@
                         :accounts/accounts accounts
                         :mailserver/mailservers mailservers
                         :network-status network-status
-                        :network network
+                        :network account-network
                         :network/type (:network/type db)
                         :chain (ethereum/network->chain-name account-network)
                         :universal-links/url url
@@ -270,4 +270,3 @@
 (re-frame/reg-fx
  :init/reset-account-data
  reset-account-data!)
-


vitvly (approved):

cammellos (approved):

rasom:
@yenda it seems that `develop` is broken by this PR

yenda:
@rasom what is broken?

rasom:
when I try to log in or create a new acc

rasom:
```
re-frame db spec check failed.

------
SUMMARY
------

After event id: :init.callback/account-change-success
Number of problems: 2

Failing root db keys:
:network

------
PROBLEMS
------

Key path: [:network]
Val: {:id "mainnet_rpc", :name "Mainnet with upstream RPC", :config {:NetworkId 1, :DataDir "/ethereum/mainnet_rpc", :UpstreamConfig {:Enabled true, :URL "https://mainnet.infura.io/v3/f315575765b14720b32382a61a89341a"}}, :rpc-url nil}

Number of problems: 2

Spec: :status-im.ui.screens.db/network
Predicate: cljs.core/string?

Spec: :status-im.ui.screens.db/network
Predicate: nil?
```

src/status_im/init/core.cljs
```diff
@@ -188,7 +188,7 @@
                         :accounts/accounts accounts
                         :mailserver/mailservers mailservers
                         :network-status network-status
-                        :network network
+                        :network account-network
```
rasom, line 191: it probably should be `account-network-id` instead of `account-network`

yenda: Not sure, I think it lacks a default value for new account, can you revert the commit?

rasom: just checked previous commit 
Screen Shot 2019-04-27 at 18 54 47


I will file a fix in a moment

rasom:
https://github.com/status-im/status-react/pull/8062

asemiankevich:
@yenda do not merge without e2e at least, it was discussed many many times.

yenda:
@asemiankevich I tested manually on Android and it worked fine

yenda:
@asemiankevich but I concede, I should have waited for e2e tests, they didn't triggered and I thought manual tests were enough, but apparently too light. The results of the e2e tests are funny, because they capture the problem (no valid network so missing data for some tokens), but they didn't capture the previous problem that after login in to an account on another network, the previous network was used. Maybe we could create a test for that though I don't know if it is worth doing it in an end to end test.

asemiankevich:
absolutely , @antdanchenko @Serhy @churik please take a look - can we add this to e2e framework? at least to nightly checks. Thanks!

Issue #8054, opened 2019-04-26
`:network` in app-db isn't always correct


yenda:
# Problem

If I create an account, switch to ropsten, then create another account, it will be on mainnet but the `:network` key on app db will still be `testnet_rpc` which is going to lead to a lot of undesirable effects

# Implementation

Set the field with the correct value

yenda:
solved

PR #8062, opened 2019-04-27
update :network with proper value

rasom:
fixed error in dev mode (and probably bugs caused broken db state), caused by #8055 (I bet e2e tests would fail there)


yenda (approved):

cammellos (approved):

yenda:
Have you checked if it does fix the previous but as well?

rasom:
@yenda if you mean #8054, then yes

PR #8036, opened 2019-04-25
Tribute to Talk adr

yenda:
Discussion around tribute to talk doc

PR #8065, opened 2019-04-29
Change e2e kickoff column to "E2E Tests"

pedropombeiro:

### Summary

Changes Probot config so that e2e tests are kicked off when moving card to `E2E tests` column.

status: ready

### DIFF
--- a/.github/github-bot.yml
+++ b/.github/github-bot.yml
@@ -10,7 +10,7 @@ project-board:
 automated-tests:
   repo-full-name: 'status-im/status-react'
   job-full-name: 'end-to-end-tests/status-app-end-to-end-tests'
-  kickoff-column-name: 'REVIEW'
+  kickoff-column-name: 'E2E Tests'
 
 github-team:
   slug: 'clojure'


annadanchenko (approved):

pedropombeiro:
Maybe we can use this PR to test why e2e tests are not getting triggered. Not sure if config is taken from `develop` or from PR branch.

pedropombeiro:
For e2e tests to be triggered, we'll need Jakub to address https://github.com/status-im/infra-misc/issues/16 first. Maybe it's the only thing that needs to be done. Unfortunately he just started his vacations today.

annadanchenko:
@PombeirP now tests are triggered 2-3 times for PRs in Review column, for example: #8009  and #8036 Would expect it's done once and after there are at least 2 approving reviews

mandrigin (approved):

pedropombeiro:
> @PombeirP now tests are triggered 2-3 times for PRs in Review column, for example: #8009 and #8036 Would expect it's done once and after there are at least 2 approving reviews

My guess is that they're not being removed from the bot queue for some reason, will look into that, should be an easy fix.

PR #8083, opened 2019-04-30
Create androidenv package with accepted license so build can run with pure environment

pedropombeiro:

### Summary

This PR consists of 2 smaller PRs:
- The first commit enables pure builds for Android:
  - cleans up some leftovers from other Nix PRs
  - creates a Nix expression that builds a utility Android SDK that has its licenses accepted. Instead of copying stuff to `~/.status/Android/Sdk` we create symlinks to nixpkgs' androidenv package and add some files that contain the license file hashes symbolizing the acceptance. It's possible that a future change to `sdkmanager` might break this scheme, but there are hundreds of other projects using the same workaround (outside of nix community), so we're not really doing bleeding edge stuff here. We always have the ability to revert to the previous way.
  - does away with looking at whether an `ANDROID_SDK_ROOT` folder exists to determine whether to run `scripts/setup` and instead creates a hidden/ignored file in the repo as a marker.
  - moves to building Android artifacts in a pure environment (passing `--pure` to `nix-shell` so that the host environment has less chances of contaminating the build).
- The second commit moves to using a nix package for fastlane on Android, removing the last barrier to pure builds.

### Review notes

Looks like we'll need to reimport the project in snyk so that it can learn the new location for Gemfile.lock.

### Testing notes

These changes should not affect the functionality, so the automated build should be enough of a test.

status: ready


mandrigin (approved):

rasom (approved):

pedropombeiro:
@jakubgs This PR is almost ready to merge, we just need to reimport the project in snyk so that it can learn the new location for Gemfile.lock.

jakubgs (approved):

PR #8084, opened 2019-05-01
[#7859] Allow to select chat item below plus button

### LINKED ISSUE #7859: Can't select listitem below plus button on iOS
## Problem
When in chat list view I cannot select the listitem below the plusbutton. When there are more than 8 items in the list, the last item can never be selected.

### Expected behavior
I can tap next to the plus button to select a list item

### Current behavior
Space around the list item is not tappable.

# (Feature or Bug? i.e Type: Bug)
Bug

# Implementation
Either the area around the plus button should be tappable, or the last listitem or equal space should always be empty.

rasom:
fix #7859

status: ready

### DIFF
--- a/src/status_im/ui/screens/home/styles.cljs
+++ b/src/status_im/ui/screens/home/styles.cljs
@@ -179,12 +179,11 @@
   {:position    :absolute
    :align-items :center
    :bottom      (+ tabs.styles/tabs-diff 6)
-   :right       0
-   :left        0})
+   :width       40
+   :height      40})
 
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

yenda (approved):

Serhy:
Looks good to me! (iOS 12.2, Android 8.1, MacOS builds were tested)
Can navigate to any chat on Chats screen including the one listed below plus button fine (same with Dapps view)

PR #8095, opened 2019-05-01
small changes in subscriptions

flexsurfer:
[do not merge]
use `re-frame` alias
removed `subs-keys` because they look redundant

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -1,5 +1,5 @@
 (ns status-im.subs
-  (:require [re-frame.core :refer [reg-sub subscribe] :as re-frame]
+  (:require [re-frame.core :as re-frame]
             status-im.tribute-to-talk.subs
             status-im.ui.screens.hardwallet.connect.subs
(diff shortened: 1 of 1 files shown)


src/status_im/subs.cljs
```diff
@@ -47,153 +47,144 @@
 
 ;; TOP LEVEL ===========================================================================================================
 
-(def subs-keys (atom {}))
-
-(defn reg-root-key-sub [sub-name db-key keys-key]
-  (swap! subs-keys update keys-key conj db-key)
-  (reg-sub sub-name :<- [keys-key] (fn [db] (get db db-key))))
+(defn reg-root-key-sub [sub-name db-key]
```
yenda, line 50: you can add a doc string that explains why (get db db-key) is safer than the shorthand (reg-sub :bla :bla) which can cause very odd errors which is why reg-root-key-sub is better for root keys that the existing shorthand.

rasom (approved):

yenda (approved):

PR #8127, opened 2019-05-04
[#6573]Have the ability to use fingerprint log in to the app (including Touch ID)

### LINKED ISSUE #6573: [Onboarding] Have the ability to use fingerprint log in to the app (including Touch ID)
### User Story

As a user it would be nice to use fingerprint log in to the app. In many environments (such as conferences) it feels awkward to enter passwords on the phone while people/cameras are potentially watching and that password unlocks access to digital assets.

---

To the bounty hunter:

(1) Add a setting to require touch/face-id to sign-in (just like "Browser privacy mode")
(2) Implement biometric protection to the saved password entry for iOS and for Android (you might use a 3rd party library for that).
(3) This should be disabled by default.

tbenr:
fixes #6573

biometric authentication is done: it supports touch-id and face-id for iOS and fingerprint for Android.

* I added a new settings

image

* when enabled it activates, biometric auth is engaged:

1. at login when password has been previously saved.
1. when app comes in foreground after being in background for more than 5 seconds.

* implemented black list for android based on deviceinfo (i updated react native module status to pull up device model info)

status: ready


src/status_im/accounts/login/core.cljs
```diff
@@ -292,13 +293,22 @@
...
+  (if (and password
+           (or bioauth-success bioauth-notrequired))
     (fx/merge cofx
               {:db (assoc-in db [:accounts/login :password] password)}
               (navigation/navigate-to-cofx :progress nil)
               (user-login false))
-    (navigation/navigate-to-cofx cofx :login nil)))
+    (fx/merge cofx
+              (when bioauth-message
+                {:utils/show-popup {:title "biometric auth" :content bioauth-message}})
```
mandrigin: this needs to be localized

src/status_im/biometric_auth/core.cljs
```diff
@@ -0,0 +1,62 @@
...
+    (-> (.isSupported rn/touchid)
+        ;; currently, for android, react-native-touch-id
+        ;; is not returning supported biometric type
+        ;; defaulting to :fingerprint 
+        (.then #(callback (or (keyword %) :fingerprint)))
+        (.catch #(callback false)))))
+
+(defn- authenticate
+  ([cb]
+   (authenticate cb "Login in Status"))
```
mandrigin: should be localized also

src/status_im/biometric_auth/core.cljs
```diff
@@ -0,0 +1,62 @@
...
+      (= code "USER_CANCELED") nil
+      (= code "USER_FALLBACK") nil
+      ;; add here more specific errors if needed
+      ;; https://github.com/naoufal/react-native-touch-id#unified-errors
+      :else (i18n/label :biometric-auth-error {:code code}))))
+
+(defn get-supported [callback]
+  (if platform/desktop?
+    (callback false)
+    (-> (.isSupported rn/touchid)
```
mandrigin: I'd love to see a black/whitelist of devices because generally speaking biometric auth on some android phones is garbage: https://www.zdnet.com/article/nokia-9-buggy-update-lets-anyone-bypass-fingerprint-scanner-with-a-pack-of-gum/

mandrigin (changes requested):
thank you! tremendous job!

there are a few notes I left and one bigger one: 
- biometric login should be optional and should be manually enabled in preferences, so it would be nice to have one more setting
- there should be a time out (can be hardcoded as of now), to require biometric login even when the app was just minimized, otherwise, the usefulness of this feature might be very low (the app can stay in bg for hours). let's put like 5 minutes after the app was minimized.

src/status_im/accounts/login/core.cljs
tbenr: yeah, i'll do a localization update once we have a finalized version

src/status_im/biometric_auth/core.cljs
tbenr: very good point

tbenr:
> * biometric login should be optional and should be manually enabled in preferences, so it would be nice to have one more setting

@mandrigin I started in that direction but then i decided to publish this wip without, just to start discussing..
did you saw the draft ui from @flexsurfer  https://github.com/status-im/status-react/issues/6573#issuecomment-489567302 ? could be good to have a clear idea of what will be the final flow so I can continue working having that in mind...

> * there should be a time out (can be hardcoded as of now), to require biometric login even when the app was just minimized, otherwise, the usefulness of this feature might be very low (the app can stay in bg for hours). let's put like 5 minutes after the app was minimized.

yeah right.. it is needed..

i'll continue on blacklist\white list and bg timeout first..

tbenr:
@mandrigin implemented reauth on resume from background with time threshold check (5 seconds hardcoded)
implemented black list for android based on deviceinfo (i updated react native module status to pull up device model info)

ghost:
`mobile_files/yarn.lock` changed. Pinging @jakubgs, @mandrigin, and @corpetty

ghost:
`mobile_files/yarn.lock` changed. Pinging @jakubgs, @mandrigin, and @corpetty

jakubgs:
I see we added:
```yarn.lock
react-native-touch-id@^4.4.1:
  version "4.4.1"
  resolved "https://registry.yarnpkg.com/react-native-touch-id/-/react-native-touch-id-4.4.1.tgz#8b1bb2d04c30bac36bb9696d2d723e719c4a8b08"
  integrity sha512-1jTl8fC+0fxvqegy/XXTyo6vMvPhjzkoDdaqoYZx0OH8AT250NuXnNPyKktvigIcys3+2acciqOeaCall7lrvg==
```
The codebase is here: https://github.com/naoufal/react-native-touch-id
That's quite a complex one, so either we get a Java dev to audit this, or we outsource auditing it.

tbenr:
@jakubgs I understand, consider also that it seems quite popular. It is used also by ledger live mobile app https://github.com/LedgerHQ/ledger-live-mobile/blob/c22f2e93c968df73587537f7834c6d8e1f678fa9/package.json#L116  but maybe it is right to not trust anyone :)

jakubgs:
@tbenr that's a fair point, but just because a big project uses doesn't mean it doesn't have security holes. @corpetty what do you think? It's actually not that big code-wise:
```
 caspair > find ./ -iname '*.java' | xargs wc -l
   45 ./examples/BiometricAuthExample/android/app/src/main/java/com/biometricauthexample/MainApplication.java
   15 ./examples/BiometricAuthExample/android/app/src/main/java/com/biometricauthexample/MainActivity.java
   42 ./examples/TouchIDExample/android/app/src/main/java/com/touchidexample/MainApplication.java
   15 ./examples/TouchIDExample/android/app/src/main/java/com/touchidexample/MainActivity.java
   30 ./android/src/main/java/com/rnfingerprint/DialogResultHandler.java
   33 ./android/src/main/java/com/rnfingerprint/FingerprintAuthPackage.java
  187 ./android/src/main/java/com/rnfingerprint/FingerprintDialog.java
   57 ./android/src/main/java/com/rnfingerprint/FingerprintCipher.java
   13 ./android/src/main/java/com/rnfingerprint/FingerprintAuthConstants.java
   67 ./android/src/main/java/com/rnfingerprint/FingerprintHandler.java
  158 ./android/src/main/java/com/rnfingerprint/FingerprintAuthModule.java
  662 total
```
I'm no Java or Android dev so most of that stuff is lost on me.

tbenr:
@jakubgs I am a java/android dev but maybe it is better a third party looks into it

jakubgs:
Well, feel free to check the code out and see if you see any red flags. I will feel better if we at least have someone read it to see if there isn't anything glaringly obvious there, and I'm sure Corey will too.

Though it would be best to have someone with security experience in this area check it out, which is why outsourcing this kind of audits makes sense. We'll see what Corey has to say.

tbenr:
@corpetty @jakubgs i look at it more closely and i find out that the implementation is quite inline with what is reported [here](https://dev.to/adammc331/android-fingerprint-authentication-tutorial-2835) which is by the end based on an official google [example](https://github.com/googlesamples/android-FingerprintDialog/). Didn't notice any meaningful differences... another couple of eyes would be useful 👀

modules/react-native-status/ios/RCTStatus/RCTStatus.m
```diff
@@ -536,16 +536,73 @@ - (void)handleSignal:(NSString *)signal
...
     return ([format rangeOfString:@"a"].location == NSNotFound);
 }
 
+- (NSString *)getBuildId {
+    #if TARGET_OS_TV
+        return @"not available";
+    #else
+        size_t bufferSize = 64;
+        NSMutableData *buffer = [[NSMutableData alloc] initWithLength:bufferSize];
+        int status = sysctlbyname("kern.osversion", buffer.mutableBytes, &bufferSize, NULL, 0);
```
mandrigin: that might be an issue when apple reviews our code... I think for iOS the device model (iPhone8,2; etc) is enough to see if we should enable biometrics.

mandrigin (changes requested):
Thanks a lot!

A few small changes are still required probably.

also regarding enabling/disabling biometrics, I think it is okay to just add another checkbox into our `dev` menu, disabled by default. later on we can add a real UI in a separate PR.

we consider our developer menu a place for experiments, so let's use it :)

tbenr:
@mandrigin ok, i'll add a new setting and it should be almost ready...

src/status_im/biometric_auth/core.cljs
```diff
@@ -0,0 +1,93 @@
...
+               :cancelText (i18n/label :cancel)}))))
+
+(defn- get-error-message
+  "must return an error message for the user"
+  [touchid-error-code]
+  (cond
+    ;; no message if user canceled or falled back to password
+    (= touchid-error-code "USER_CANCELED") nil
+    (= touchid-error-code "USER_FALLBACK") nil
+    ;; add here more specific errors if needed
```
corpetty, line 53: I'd recommend adding a few more, not sure this will catch most situations.  In particular at least: 
- `NOT_SUPPORTED`
- `NOT_PRESENT`
- `NOT_ENROLLED`

src/status_im/constants.cljs
```diff
@@ -6,6 +6,8 @@
 
 (def ethereum-rpc-url "http://localhost:8545")
 
+(def ms-in-bg-for-require-bioauth 5000)
```
corpetty, line 8: does this mean if app is in background for more than 5 seconds, then we ask for re-auth via biometrics?

src/status_im/biometric_auth/core.cljs
```diff
@@ -0,0 +1,93 @@
...
+  (cond
+    (= (:brand deviceinfo) "bannedbrand") true
+    :else false))
+
+;; biometric auth config
+;; https://github.com/naoufal/react-native-touch-id#authenticatereason-config
+(def authenticate-options
+  (clj->js (merge
+            {:unifiedErrors true
+             :passcodeFallback false}
```
corpetty: I think `passcodeFallback` is iOS only.

corpetty (commented):
Thanks for the contribution!  Great work.

## Potential Issues:
- I see you're accepting `USER_FALLBACK` error message, but don't see where handling the fallback passcode (seems this is only iOS specific and refers to OS PIN passcode, not Status) entry is done.  It is also set to false in the options.  Am I missing something?
- What happens if a user fails to authenticate?
- From what I can tell, the package included follows best practices on implementing biometrics.  There are some issues regarding a few bugs on the the compatibility with older devices (seeming to be introduced in 4.2 - see recent issues in the dependency repo).  
- What happens if a user has an incompatible device (including compatibility bugs mentioned below)?
- What happens if a user doesn't have any biometrics stored but is supported?


## Thoughts
In any event, I'm fine with moving forward (assuming the above points are addressed and it gets added in the manner @mandrigin offers up (dev mode)).  Authentication mechanisms and dependencies are a part of the final audit scope already, so this will definitely be included. 

This is a great start to bio authentication, but I would like to eventually see lockout options after failures, and password-fallback (not implemented in added dependency, so extra dev work there).  This is probably out of the scope of the original bounty.

src/status_im/constants.cljs
tbenr: @corpetty correct

tbenr:
> Thanks for the contribution! Great work.
> 
> ## Potential Issues:
> * I see you're accepting `USER_FALLBACK` error message, but don't see where handling the fallback passcode (seems this is only iOS specific and refers to OS PIN passcode, not Status) entry is done.  It is also set to false in the options.  Am I missing something?

what I implemented in general here is that if user fails bioauth he has to relogin using status account password (current login method). That's why I explicitly turn off fallback.

> * What happens if a user fails to authenticate?

when coming back from background: 
if user cancelled he gets:
image

if bio auth error he gets:
image


and user can retry or logoff and authenticate with normal status login.

if user is on login with saved password (app restarted), if user fails to authenticate (or cancel) he is automatically logged off and he has to reauthenticate with password:

image


> * From what I can tell, the package included follows best practices on implementing biometrics.  There are some issues regarding a few bugs on the the compatibility with older devices (seeming to be introduced in 4.2 - see recent issues in the dependency repo).

> * What happens if a user has an incompatible device (including compatibility bugs mentioned below)?

the bioauth checkbox in settings is disabled (user cannot enable it), and currently can continue saving password.

> * What happens if a user doesn't have any biometrics stored but is supported?

he gets an error when tries to enable it.

image


> 
> ## Thoughts
> In any event, I'm fine with moving forward (assuming the above points are addressed and it gets added in the manner @mandrigin offers up (dev mode)). Authentication mechanisms and dependencies are a part of the final audit scope already, so this will definitely be included.
> 
> This is a great start to bio authentication, but I would like to eventually see lockout options after failures, and password-fallback (not implemented in added dependency, so extra dev work there). This is probably out of the scope of the original bounty.

src/status_im/biometric_auth/core.cljs
tbenr: @corpetty these cases are managed but there is no specific error message defined (see screenshot in the comment below). What is missing is a specific error messages for these cases, but they can be added if needed. The important point here is to return `nil` when we don't want to show a message to user (ie. if user cancelled doesn't make sense to notify that he pressed cancel, he already knows 😄)

src/status_im/biometric_auth/core.cljs
tbenr: yeah.. moving in ios section...

tbenr:
just improved a message text

tbenr:
rebased to resolve conflicts

mandrigin (approved):
I think it is good to have this option under dev mode. Everything else we can address in separate PRs. Thanks a lot @tbenr for your job!

tbenr:
@mandrigin ok i'll move it under dev mode

mandrigin:
@tbenr isn't it already there? :)

tbenr:
@mandrigin well.. not really, it is under advanced settings but appears regardless dev-mode

mandrigin:
@tbenr ah, okay, let's then move it under dev mode and I think it it should be okay to move to testing.

tbenr:
@mandrigin done!

mandrigin:
@flexsurfer can you take a look at this review too?

src/status_im/ui/screens/profile/user/views.cljs
```diff
@@ -176,65 +177,82 @@
                                           :action-fn           #(re-frame/dispatch [:accounts.logout.ui/logout-pressed])}]]]]))
 
 (defview advanced-settings [{:keys [network networks dev-mode? settings]} on-show]
-  {:component-did-mount on-show}
-  [react/view
-   (when (and config/extensions-enabled? dev-mode?)
+  (letsubs [supported-biometric-auth [:get-supported-biometric-auth]]
```
yenda: looks like there is some changes here that are not related to the PR? Maybe a merge accident?

src/status_im/ui/screens/profile/user/views.cljs
```diff
@@ -176,65 +177,82 @@
...
+      {:label-kw  :t/chaos-mode
+       :value     (:chaos-mode? settings)
+       :action-fn #(re-frame/dispatch [:accounts.ui/chaos-mode-switched %])}]
+     (when dev-mode?
+       [profile.components/settings-item-separator]
+       [profile.components/settings-switch-item
+        {:label-kw  :t/biometric-auth-setting-label
+         :value     (:biometric-auth? settings)
+         :active?   (some? supported-biometric-auth)
+         :action-fn (fn [biometric-auth?]
```
flexsurfer: this fn should be moved to event

src/status_im/ui/screens/profile/user/views.cljs
```diff
@@ -176,65 +177,82 @@
...
+                      (if biometric-auth?
+                        (biometric-auth/authenticate
```
flexsurfer: this should be moved to fx

src/status_im/ui/screens/profile/user/views.cljs
tbenr: @yenda are you referring to deleted lines here?

tbenr: btw that's very strange.. let me check and fix

tbenr: @yenda deleted lines are actually there (after the sub), it is a diff glitch

image

src/status_im/ui/screens/profile/user/views.cljs
tbenr: @flexsurfer done!

src/status_im/ui/screens/profile/user/views.cljs
tbenr: @flexsurfer done!

src/status_im/accounts/login/core.cljs
```diff
@@ -273,13 +274,14 @@
                 (navigation/navigate-to-cofx :enter-pin-login nil)))))
 
 (fx/defn get-user-password
-  [_ address]
+  [_ address biometric-auth?]
```
flexsurfer: there is no need to pass it here, you can retrieve it from db in `:accounts.login.callback/get-user-password-success` event

src/status_im/subs.cljs
```diff
@@ -279,6 +279,11 @@
    (-> selected-participants
        (contains? element))))
 
+(re-frame/reg-sub
```
flexsurfer: this should be registered with `reg-root-key-sub`

flexsurfer (approved):

src/status_im/accounts/login/core.cljs
tbenr: ah yes, right, was like that before introducing the profile setting...

tbenr: in any case i have to pass address to `:accounts.login.callback/get-user-password-success` to be able to lookup in db.

tbenr:
@flexsurfer @mandrigin 
I notice there is a problem in the login flow when the password is saved (and biometric-auth is not enabled).
In my ios simulator i get this and I replicated in current develop branch too, so it seems not be related to this PR (instead it seems to work if bioauth is enabled).

image

I can investigate, if someone has an idea please share so i can fix :)

flexsurfer:
yeah sometimes this happens in dev, i don't know the reason, but i believe it's not related to this PR

tbenr:
@flexsurfer good to know, maybe is related to https://github.com/react-navigation/react-navigation/pull/5203

tbenr:
fixed a mistake while rebasing..

asemiankevich:
@tbenr 

1. password is used to sign transactions too, is it planned to be implemented here or separately?
2. even i enabled password saving and was able to login with fingerprint -> when i change network it requires password
3. @hesterbruikman or @andmironov please review from design standpoint, i think we need some designs here first
4. how about face id?

cc @annadanchenko

mandrigin:
@asemiankevich 
> password is used to sign transactions too, is it planned to be implemented here or separately?

separately

> @hesterbruikman or @andmironov please review from design standpoint, i think we need some designs here first

IMO that's unnecessary, since it is an experimental feature for now, there should be one more iteration to improve the UX

> how about face id?

does it not work? the librarly looks like it just uses a biometric login, not just a fingerprint

asemiankevich:
what i have noticed so far: i can lock the prompts for valid input 

1. open status -> create account -> logout and enable save password -> login
2. profile -> dev mode -> biomterics toggle -> tap wrong finger several times -> receive the error message -> close it
3. toggle the biometrics toggle again and repeat step 2 -> close the error message
4. toggle the biometrics toggle again

As a result error message is shown -> no more attempts left for valid input. 

image

@mandrigin is it expected?

churik:
From my POV:
### 5. From description:

> when enabled it activates, biometric auth is engaged:

> 1. at login when password has been previously saved.
> 2. when app comes in foreground after being in background for more than 5 seconds.

So as for me it doesn't resolve initial issue #6573  that user should each time after manual logout enter password, but adds security is app usage. Not sure what is the goal. 

### 6. Issue: when app is in the background for more than 5 seconds - when trying use wrong finger to unblock phone several times - can get several popups "You must authenticate"

**Steps:**
1) create account
2) enable development mode > biometric authentication
3) put app to background > 5s
4) go back to app and several times use wrong finger to login. Expected here, that I can enter password and login. In fact When tap on "Enter password" > again redirected to popup "You must aithenticate"
5) after several attempts instead of 1 popup "You must aithenticate" you will see 2 popups

If here tap on logout > and one more time on logout -> app is crashed
Logs: [status_logs-1.zip](https://github.com/status-im/status-react/files/3217414/status_logs-1.zip)

ezgif com-optimize

tbenr:
@churik thanks, it is not clear to me when the second poppup comes in.

are you saying that, after a wrong fingerprint, you press "Enter Password" here:

image

and you get another popup instead of just landing to login page?

tbenr:
@churik ah yes.. I replicated it.. let me fix it...

tbenr:
@churik i fixed the "you must authenticate" popup when you press "Enter Password" on failed fingerprint.
now when you press "Enter Password" you go directly to login.

but I'm not able to replicate 2 popup overlap and app crash on second logout:

> 5. after several attempts instead of 1 popup "You must aithenticate" you will see 2 popups

it would be very useful if you could post another video to clarify how to get there.

thanks!

tbenr:
@churik 

> So as for me it doesn't resolve initial issue #6573 that user should each time after manual logout enter password, but adds security is app usage. Not sure what is the goal.

manual logout requires the user to enter password again but if you don't do that (ie close\kill the app) you can login with fingerprint\faceid only (as far as I understood this is the main flow we want to cover)

tbenr:
@asemiankevich

> > how about face id?
> 
> does it not work? the librarly looks like it just uses a biometric login, not just a fingerprint

it should work with face-id as well, but I didn't tested

tbenr:
> what i have noticed so far: i can lock the prompts for valid input
> 
> 1. open status -> create account -> logout and enable save password -> login
> 2. profile -> dev mode -> biomterics toggle -> tap wrong finger several times -> receive the error message -> close it
> 3. toggle the biometrics toggle again and repeat step 2 -> close the error message
> 4. toggle the biometrics toggle again
> 
> As a result error message is shown -> no more attempts left for valid input.
> 
> image
> 

@asemiankevich to activate bioauth I'm verifying the user is really able to authenticate with bioauth (to double check it works on fancy android phones). Here we are really performing an authentication and if the user continue failing, standard phone security rules applies (ie authentication lockout occurs, the same thing happens when you enter wrong pin several times). maybe when the user locks biometric have to lock the phone and unlock it again (on iOS using pin), but I don't know, it may be different from device to device...

churik:
@tbenr I noticed new behavior, suppose that it is an issue:
#### 7) At attempt to set up biometric authentication with wrong finger and tap on "Enter password" - popup closed

**Steps:**
- create account
- when enabling biometric protection, use the wrong finger
- tap on "Enter password"

**Expected result:** redirected to the login. Actually, in this case, I would prefer don't see this option (I mean when I enter the password, it doesn't enable biometric  authentication anyway)
**Actual result:** popup closed

ezgif com-optimize (1)

tbenr:
@churik yes, it is possible to hide it in that case, let me fix it...

tbenr:
@churik i rebased and fixed the "enter password" label.

- now it appears only on login (after kill\restart app).
- in resume from background and on setting (enabilng biometric) it doesn't appear anymore.

IMHO in resume form background it is better to let user explicitly press "logout" to reenter password,
but now is more configurable and can be easily reintroduced (even with a different label) where needed.

I also changed "Logout" to "Relogin".

p.s.
are you "chu" in meritocracy app? :)

react-native/src/mobile/status_im/react_native/js_dependencies.cljs
```diff
@@ -10,6 +10,7 @@
 (def status-keycard         (js-require/js-require "react-native-status-keycard"))
 (def realm                  (js/require "realm"))
 (def webview-bridge         (js-require/js-require "react-native-webview-bridge"))
+(defn touchid []            (.-default (js/require "react-native-touch-id")))
```
yenda, line 13: @rasom is this still ok after the ram-bundle PR merge?

src/status_im/events.cljs
```diff
@@ -60,6 +60,7 @@
             [status-im.wallet.core :as wallet]
             [status-im.wallet.db :as wallet.db]
             [status-im.web3.core :as web3]
+            [status-im.biometric-auth.core :as biomentric-auth]
```
yenda, line 63: Please don't register new events like this, events are now registered by using the attribute map of fx/defn

src/status_im/events.cljs
```diff
@@ -95,6 +96,11 @@
  (fn [cofx [_ device-uuid]]
    (init/set-device-uuid cofx device-uuid)))
 
+(handlers/register-handler-fx
+ :init.callback/get-supported-biometric-auth-success
+ (fn [cofx [_ supported-biometric-auth]]
+   (init/set-supported-biometric-auth cofx supported-biometric-auth)))
```
yenda: for instance here in  `set-supported-biometric-auth` you would have {:events [:init.callback/get-supported-biometric-auth-success]} before the parameters map. It helps getting rid of boilerplate, exposing circular dependencies and prevents registration of events that do more than calling one function as required in the guidelines

src/status_im/ui/screens/profile/user/views.cljs
yenda: sorry for the late reply, even though this seems compliant with how the defview macro works I think it would be better to turn this view into a simple defn and subscribe to the subscription in the parent, passing only the value as parameter of this component, there is no compelling reason for this view to use the defview macro

src/status_im/ui/screens/db.cljs
```diff
@@ -176,6 +177,10 @@
 
 (spec/def ::device-UUID (spec/nilable string?))
 
+;;;; Supported Biometric authentication types
+
+(spec/def ::supported-biometric-auth (spec/nilable keyword?))
```
yenda: is it just one single keyword? how does it work? is the list open or could we have a set of possible keywords here? I don't find any exemple of what I should expect here

src/status_im/subs.cljs
```diff
@@ -89,6 +89,7 @@
 (reg-root-key-sub :get-pairing-installations :pairing/installations)
 (reg-root-key-sub :network/type :network/type)
 (reg-root-key-sub :tooltips :tooltips)
+(reg-root-key-sub :get-supported-biometric-auth :supported-biometric-auth)
```
yenda: don't add the `get-` prefix here it's superfluous, the other subs that use that are wrong

yenda:
@tbenr I just have some minor comments but otherwise lgtm, sorry I was at a conference that's why I didn't reply earlier

churik:
@tbenr No, I'm not) 
keep me updated about your decisions according to https://github.com/status-im/status-react/pull/8127#pullrequestreview-242395162, I need to check the final version. Thanks!

tbenr:
@yenda thanks for feedback, i'll go through them

@churik i'll ping you when final version will be ready

src/status_im/ui/screens/db.cljs
tbenr: @yenda yes it is a keyword only, inherited by touch-id module.
`:TouchID` (ios)
`:FaceID` (ios)
`:fingerprint` (android)
`nil` (not supported\desktop)

I can make it a strict set of possible keywords.

src/status_im/ui/screens/profile/user/views.cljs
tbenr: @yenda I did a test and seems to me that if I make it defn when pushing "Advanced ^" the view doesn't scroll down anymore... I leave it a defview but I move subscription in parent.. makes sense?

src/status_im/ui/screens/db.cljs
tbenr: @yenda done

tbenr:
@yenda i think i covered all the points.

yenda (approved):
there is still events created in src/status_im/events.cljs that could be declared within fx/defn but otherwise lgtm

tbenr:
right..let me check them.. need to get familiar with the new fx/defn..

tbenr:
@churik this should be a good one

churik:
@tbenr 
one more issue: 
### 8. When biometric authentication is locked because of number of attempts, popup "Biometric authentication error (UNKNOWN_ERROR) appears after each relogin

**Steps:**
- enable biometric authentication
- enable "Save password"
- close app
- try to login with the wrong finger until "Biometric authentication error (UNKNOWN_ERROR)" popup
- close popup
- login 
- logout

**Expected result:** no popup is shown because  "Save password" is not enabled anymore
**Actual  result:** 
photo_2019-05-30 17 04 10
**Logs:**
[status_logs.zip](https://github.com/status-im/status-react/files/3237572/status_logs.zip)

tbenr:
@churik uhmm, the reason is that, even if you see checkbox disabled there, the password is still saved (it is deleted only on manual logout).

i can fix it by, when such authentication error occurs, deleting the saved password or disable biometric auth in user preferences.. maybe the latter is better, what do you think?

churik:
@tbenr I logged out manually, and still see popup. That's why I logged it.
So I close popup -> login -> logout -> popup is shown.

tbenr:
@churik can't replicate here 👀.. are you using a physical device?
does biometric auth popup appear also without locking biometric auth?
login(savepassword) -> enable bio -> logout -> login

churik:
@tbenr I can't reproduce it either today.
Anyway, it is not a blocker.
So good to merge from my side.
Thanks for your work @tbenr !

flexsurfer:
congrats @tbenr , awesome job!

tbenr:
@churik \ @flexsurfer thanks!

ghost:

PR #8155, opened 2019-05-08
Fix send tx one by one in dapp test

Serhy:
* fix send tx one by one in dapp test
+ added `faucet.ropsten.be` as one more faucet for some tests


churik (approved):
Thanks @Serhy !

asemiankevich (approved):

PR #8164, opened 2019-05-09
Optimized ScrollView instead of FlatList for desktop chat

### LINKED ISSUE #8163: Bad chat performance in mobile UI for desktop
# Problem

`FlatList` implementation in `react-native-desktop` works slow and makes chat in mobile ui for desktop unusable

## Implementation

As a workaround, until `FlatList` implementation is buggy we can use `ScrollView` component that has optimization on Qt side of `react-native-desktop`

## Acceptance Criteria

User able to scroll chat back and forth without serious lags

vkjr:
fixes #8163

### Summary

FlatList implementation in react-native-desktop works slow and makes chat in mobile ui for desktop unusable

As a workaround, until FlatList implementation is buggy we can use ScrollView component that has optimization on Qt side of react-native-desktop. 

### Review notes
`ScrollView` that used instead of `FlatList` taken from current desktop UI and works the same way

### Testing notes
Currently in new mobile UI you can't open user profile by clicking on his avatar. This is a known issue. I have a solution for it in `react-native-desktop` but this solution causes bug in old desktop UI which is default now (only part of avatar image responds to click.)
My suggestion is to leave this bug for now and fix it before we fully switch to new UI

#### Platforms
- Android
- iOS
- macOS
- Linux
- Windows

Only chat area should be impacted

##### Functional

- 1-1 chats
- public chats
- group chats

### Steps to test
- enable mobile UI by setting `MOBILE_UI_FOR_DESKTOP=1` in `.env` file
- make sure that chat has better performance and behaves the same way as a current desktop UI.

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/views.cljs
+++ b/src/status_im/ui/screens/chat/views.cljs
@@ -1,5 +1,6 @@
 (ns status-im.ui.screens.chat.views
   (:require [re-frame.core :as re-frame]
+            [reagent.core :as reagent]
             [status-im.chat.models :as models.chat]
             [status-im.contact.db :as contact.db]
             [status-im.group-chats.db :as group-chats.db]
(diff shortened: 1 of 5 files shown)


src/status_im/ui/screens/chat/views.cljs
```diff
@@ -300,7 +352,9 @@
                                  (re-frame/dispatch [:set :layout-height (-> e .-nativeEvent .-layout .-height)]))}
         [chat-toolbar current-chat public? modal?]
         [messages-view-animation
-         [messages-view current-chat modal?]]
+         (if platform/desktop?
+           [messages-view-desktop current-chat modal?]
```
mandrigin, line 357: probably a TODO comment here would be nice

mandrigin (approved):
looks good overall

src/status_im/ui/screens/chat/views.cljs
vkjr: added one

mandrigin: Thanks!

flexsurfer: what about other places where flatlist is used? tbh i thought that we make changes in flatlist component itself

flexsurfer:
@vkjr what about other places where flatlist is used? tbh i thought that we make changes in flatlist component itself?

flexsurfer:
i mean something like
```
(defn flat-list
  "A wrapper for FlatList.
   See https://facebook.github.io/react-native/docs/flatlist.html"
  [{:keys [data] :as props}]
  {:pre [(or (nil? data)
             (sequential? data))]}
(if desktop?
  [flat-list-scroll-view props]
  [flat-list-class
   (merge (base-list-props props)
          props
          {:data (wrap-data data)})])
```

vkjr:
@flexsurfer, flatlist shows bad performance when there are a lot of elements and a lot of dynamic loads\unloads needed (lot of scrolling). So far it looks like in other places except chat it doesn't affect user experience. If you think that it would be a more clean solution - I can substitute it everywhere but I also think that it would be good to reduce the surface where mobile and desktop code differs as much as possible.

mandrigin:
@flexsurfer @vkjr in this discussion I would accept @vkjr's side, taking into accoun that that is a temporary measure and we dont want to spend too much time fixing potential regressions by replacing `FlatList` with `ScrollView` in the other places.

flexsurfer:
> @flexsurfer, flatlist shows bad performance when there are a lot of elements and a lot of dynamic loads\unloads needed (lot of scrolling). So far it looks like in other places except chat it doesn't affect user experience. If you think that it would be a more clean solution - I can substitute it everywhere but I also think that it would be good to reduce the surface where mobile and desktop code differs as much as possible.

ok then :)

flexsurfer (approved):

churik:
@vkjr 
overall experience: scroll is definitely is better, but some new issues are found:
#### 1) Input is not erased after sending messages
**Steps:**
- create account
- join any chat
- send a message
**Expected result:** input is erased after sending
**Actual result:** input is preseved after sending
**Platform:** mobile UI (Mac OSx)

#### 2)Can't fetch the previous history
**Steps:**
- create account
- join any public chat
- tap on "Fetch more messages"
**Expected result:** messages are fetched
**Actual result:** error
**Platform:** mobile UI (Mac OSx)
asdsadsad

#### 3)Share my code is not visible
**Steps:**
- create account
- go to profile
- tap on "Share my contact code"
**Expected result:** input is erased after sending
**Actual result:** input is preseved after sending
**Platform:** mobile UI (Mac OSx)
fdfsdf

Didn't notice issues with mobile / regular desktop builds.

vkjr:
@churik 
1) Fixed
2) Fixed
3) I noticed that QR-code shows up if you use a scroll on a "share my profile page". So I suspect there can be some issue with `react-native-desktop`. I would suggest to put it aside until mobile UI is not default and 2-pane UI is not there because that can take time and I think implementing 2-pane UI has bigger priority. wdyt?

Also in mobile UI you can notice that text input doesn't reset to blank if message sent with send icon. It is expected behavior for now. In main branch of `react-native-desktop` this problem fixed already but we can't switch to it until fully switch to mobile UI because old desktop UI built with incorrect workaround of that problem. So I suggested also to save some time and ignore issue until mobile UI become default.

churik:
@vkjr Can't get build `android e2e` build, can you take a look? 
https://ci.status.im/job/status-react/job/prs/job/android-e2e/job/PR-8164/7/

vkjr:
Some problem on last `upload` stage, so restarted it

vkjr:
@jakubgs, could you please take a look? I see `Invalid or unsupported command "saucelabs"
` error while building `android-e2e`. Have no idea what it is about.

churik:
@vkjr  can you rebase please?

jakubgs:
Please rabase on `develop` so you include this commit: https://github.com/status-im/status-react/commit/cc009a3c2cf41a6a99878d2f9eab7325473b6d3a
Your branch still is using `FASTLANE_SCRIPT` which has a bug that caused it to be empty.

vkjr:
Done

churik:
Tested briefly (creating accounts, chatting) on Linux Ubuntu 18.04, Windows 10 Home, Mac OSx (normal desktop builds) and on IOS 11.4.1 (IPhone 7), Android 8 (LG V20) - no regression is found.

PR #8165, opened 2019-05-09
remove unecessary check in update-status-go script

yenda:
Remove a check preventing updates of status-go from branches

status: ready

### DIFF
--- a/scripts/update-status-go.sh
+++ b/scripts/update-status-go.sh
@@ -24,12 +24,6 @@ if [ $# -eq 0 ]; then
 fi
 
 STATUS_GO_OWNER="$(cat ${GIT_ROOT}/STATUS_GO_OWNER)"
-STATUS_GO_VERSION="v${1#"v"}"
-if [ "$STATUS_GO_OWNER" == 'status-im' ] && [ "$STATUS_GO_VERSION" != "$1" ]; then
-  echo "status-go release branches should include the v prefix!"
-  echo "Please create a new branch called $STATUS_GO_VERSION"
-  exit 1
-fi
 STATUS_GO_VERSION=$1
 STATUS_GO_SHA256=$(nix-prefetch-url --unpack https://github.com/${STATUS_GO_OWNER}/status-go/archive/${STATUS_GO_VERSION}.zip)
 


mandrigin (approved):

jakubgs:
What a mess this is. Some releases in `status-go` have `v` some don't:
https://github.com/status-im/status-go/releases
And the one we use currently does:
https://github.com/status-im/status-react/blob/3469c4a9b1c34f13329794abaf47dba70e25aca2/STATUS_GO_VERSION#L3
We have to pick one convention and stick with and and stop switching this. Personally I think adding a prefix is better, because it makes it easier to identify release tags in git.
Last time I talked about this with Adam he said we should have `v` prefix, but then I still see releases with out it, despite the `release` target in the `Makefile` adding the prefix:
https://github.com/status-im/status-go/blob/7002311f96a9ba91431e50f7046e3c2166153b7b/Makefile#L215-L221
what the fuck is this shit - treize

jakubgs:
If the `release` target in `Makefile` from `status-go` enforces `v` prefix then I think we should not be removing this `if` check.

pedropombeiro (approved):

pedropombeiro:
> If the `release` target in `Makefile` from `status-go` enforces `v` prefix then I think we should not be removing this `if` check.

I think the issue is with dev builds, where a dev wants to test a non-release tag (which doesn't with `v`)

jakubgs:
>I think the issue is with dev builds, where a dev wants to test a non-release tag (which doesn't with v)

Why not though? We have plenty of non-release releases like [`v0.24.0-beta.0`](https://github.com/status-im/status-go/releases/tag/v0.24.0-beta.0). All release tags(even if they are just a draft or pre-release) should have `v` prefix. Or none should. It's either or.

yenda:
@jakubgs  no it's for git branches, when we want to try a PR
This test is useless, for releases it should be checked somewhere else

jakubgs:
>no it's for git branches, when we want to try a PR

Okay, I see, just because someone is testing a branch with `status-react` doesn't mean it itself is a release candidate.
This version stored in `STATUS_GO_VERSION` file is used by nix to fetch the right `status-go` branch/tag/commit:
https://github.com/status-im/status-react/blob/3469c4a9b1c34f13329794abaf47dba70e25aca2/nix/status-go/default.nix#L12-L17
I guess adding a check for whether the thing specified by the user is a branch or a tag is a bit of a pain. Then why not just leave it a warning?
>You are using a status-go version name not prefixed with `v`, which means it isn't a release. Make sure this is what you want.

Or something.

PR #8175, opened 2019-05-10
[fix] ipfs CID

yenda:
# summary
the `add` post method on infura gateway allows a cid-version argument that I added which returns a CID instead of a simple hash. That way it is clear which content-id the ipfs object has. 

should fix tribute to talk once the ttt PR is rebased

# testing
I added a unit test:
this is the base58 encoded CID from ipfs
https://cid.ipfs.io/#zb2rhZfjRh2FHHB2RkHVEvL2vJnCTcu7kwRqgVsf9gpkLgteo
this is the conversion to hex
https://cid.ipfs.io/#f015512202cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824 (removed contenthash prefix and added f which means hex in multibase format)

both have the same BASE32 CIDV1

isn't used anywhere except in tribute-to-talk, so doesn't require testing yet

status: ready

### DIFF
--- a/src/status_im/ipfs/core.cljs
+++ b/src/status_im/ipfs/core.cljs
@@ -4,7 +4,7 @@
 
 ;; we currently use an ipfs gateway but this detail is not relevant
 ;; outside of this namespace
-(def ^:const ipfs-add-url "https://ipfs.infura.io:5001/api/v0/add")
+(def ^:const ipfs-add-url "https://ipfs.infura.io:5001/api/v0/add?cid-version=1")
 (def ^:const ipfs-cat-url "https://ipfs.infura.io/ipfs/")
 
 (fx/defn cat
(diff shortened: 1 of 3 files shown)


yenda:
@flexsurfer @jeluard please review

src/status_im/ipfs/core.cljs
```diff
@@ -4,8 +4,8 @@
 
 ;; we currently use an ipfs gateway but this detail is not relevant
 ;; outside of this namespace
-(def ^:const ipfs-add-url "https://ipfs.infura.io:5001/api/v0/add")
-(def ^:const ipfs-cat-url "https://ipfs.infura.io/ipfs/")
+(def ipfs-add-url "https://ipfs.infura.io:5001/api/v0/add?cid-version=1")
```
flexsurfer: just wondering why don't you like const?

flexsurfer (approved):

src/status_im/ipfs/core.cljs
yenda: I put them back it was a pita when I had to experiment with urls

jeluard (approved):

PR #8179, opened 2019-05-10
remove duplicate entries in en.json

tbenr:
remove duplicate entries in localization file en.json


translations/en.json
```diff
@@ -281,7 +280,6 @@
     "currency-display-name-mzn": "Mozambique Metical",
     "block": "Block",
     "wallet-set-up-title": "Set up your wallet",
-    "loading": "Loading... ",
```
bitsikka, line 284: btw this "Loading... " was intentionally made to have "space" at the end to separate it out visually from the spinner. The spacing will probably need to be handled differently(:margin/:padding) where it is used.

tbenr: yes, I think it is better to manage it via styles instead of relying on a space.. but if you want i can leave the entry with the space and delete the other one...

flexsurfer (approved):

jeluard (approved):

churik:
Failed e2e https://ethstatus.testrail.net/index.php?/tests/view/649213 is not related to PR, everything is fine if checking manually.
Checked "Learn more" in Pairing section and "Loading..." for empty public chat state on IOs and Android

PR #8183, opened 2019-05-10
refactor nix_sh to give it optional Map arguments

jakubgs:
Refactoring `nix_sh` to take optional arguments:

* `pure` - Descides if `--pure` is used.
* `keep` - List of env variables to make available in pure mode
* `args` - Map of arguments and values to pass to nix

I've also done a bunch of other cleanup here and there.


ci/android.groovy
```diff
@@ -42,7 +49,10 @@ def uploadToPlayStore(type = 'nightly') {
   withCredentials([
     string(credentialsId: "SUPPLY_JSON_KEY_DATA", variable: 'GOOGLE_PLAY_JSON_KEY'),
   ]) {
-    utils.nix_fastlane_sh "android ${type}"
+    nix.shell(
+      "fastlane android ${type}",
+      keep:['FASTLANE_DISABLE_COLORS', 'GOOGLE_PLAY_JSON_KEY']
```
pedropombeiro:
```suggestion
      keep: ['FASTLANE_DISABLE_COLORS', 'GOOGLE_PLAY_JSON_KEY']
```

pedropombeiro (approved):
Love it!

jakubgs:
@PombeirP what do you think about moving the `LANG`, `LANGUAGE`, and `LC_ALL` env variables to `shell.nix`? Probably in `shellHook` but maybe there's a better place to set it.

jakubgs:
I also need to figure out which stage uses the `REALM_DISABLE_ANALYTICS`. I can't recall if it was `Build` or `Bundle`.

jakubgs:
@PombeirP What do you think about it now? I think putting the locale settings in `shell.nix` makes sense, but maybe you have a better place.

Also, doesn't this mean that all the places where we define env variables in `shellHook` could be just there as variables without having to wrap them in `shellHook`, or am I missing something?

pedropombeiro:
> @PombeirP What do you think about it now? I think putting the locale settings in `shell.nix` makes sense, but maybe you have a better place.

I'm not sure if we want to restrain the locale settings for international users, I don't have much experience there as I have my whole system in English anyway.

> Also, doesn't this mean that all the places where we define env variables in `shellHook` could be just there as variables without having to wrap them in `shellHook`, or am I missing something?

Putting them outside of shellHook is just a shortcut for putting them inside shellHook, because that's what Nix will do just before creating a shell. The problem with having them as attributes is that they're not easily accessible for upper-level Nix scripts like shellHook is. If we want to have a hierarchical structure to the Nix scripts, it makes sense keeping the env var definitions inside the script that defines them, and shellHook is an easy way to bubble them up to the top level shellHook. There might be a different way to do it, like using the `//` operator to merge maps though. Could be worth looking into.

pedropombeiro:
Some env vars also don't lend themselves well to being defined outside shellHook. Consider the following cases:
``` bash
    export FASTLANE_PLUGINFILE_PATH=$PWD/fastlane/Pluginfile
    STATUS_REACT_HOME=$(git rev-parse --show-toplevel)
```

jakubgs:
>I'm not sure if we want to restrain the locale settings for international users, I don't have much experience there as I have my whole system in English anyway.

That's a fair point. It is kind of an interference with users working environment, so maybe it's not the best idea. I've moved it back to the Jenkinsfiles for now.

 >Putting them outside of shellHook is just a shortcut for putting them inside shellHook, because that's what Nix will do just before creating a shell. The problem with having them as attributes is that they're not easily accessible for upper-level Nix scripts like shellHook is. 

Ah, I wasn't aware of that, good to know.

>There might be a different way to do it, like using the // operator to merge maps though. Could be worth looking into.

Well, that would just be a stylistic change. I actually like how `shellHooks` is one place where we set env variables we clearly want the user to have in the shell, if we made them just regular arguments in the nix expression the only think making them stand out would be being uppercase, and maybe some comments. It's not important I guess, I just thought we were doing something kinda backwards.

cammellos (approved):

PR #8191, opened 2019-05-12
[experiment] limit number of rendered messages when open a chat

rasom:
- Currently we render 20 last messages when chat is opened first time, if we have more messages loaded we render all of them. Having much more than 20 messages doesn't affect performance of flat list too drastically because most of items which are not viewable are not drawn. At the same time ~20 message currently produce 2200-2400 calls to UIManager through bridge. These calls are not too expensive per se (drawing is more expensive) but still cost some time on js thread and we don't limit them (having 200 loaded messages means much more such calls -> js thread loaded with unnecessary work).
- In this pr we limit number of rendered messages to 5, if it is not enough to 11 and then 20. After that we add 20 more messages to stream when necessary. This 5-11-20 limit is re-applied each time the chat is opened, which means that even if we already loaded 500 messages they will not contribute to total number of UIManager calls during navigation, only when user will scroll messages. 

____

Results:

### navigation from chat list to chat (#status) w/o opacity wrapper, galaxy s9
- 20 messages without limit: **0.43-0.6s** 
- 20 messages with limit: **0.24s; 45% faster**

### navigation from chat list to chat with opacity wrapper, galaxy s9
- 20 messages without limit: **0.53s** (0.16s <- for switching screens)
- 20 messages with limit: **0.27s** (0.16s <- for switching screens) **49% faster**

---
### navigation from chat list to chat (#status) w/o opacity wrapper, galaxy a500d (slower device)
- 20 messages without limit: **1.8s**
- 20 messages with limit: **1.234s; 31% faster** 

### navigation from chat list to chat with opacity wrapper, galaxy a500d (slower device)
- 20 messages without limit: **2.1s** (0.54s <- for switching screens) 
- 20 messages with limit: **1.34s** (0.54s <- for switching screens) **36% faster**

----

Navigation from chats list to chat is **30-45%** faster.

### DIFF
--- a/src/status_im/ui/screens/chat/views.cljs
+++ b/src/status_im/ui/screens/chat/views.cljs
@@ -29,7 +29,8 @@
             [status-im.utils.platform :as platform]
             [status-im.utils.utils :as utils]
             [status-im.utils.datetime :as datetime]
-            [status-im.ui.screens.chat.message.gap :as gap])
+            [status-im.ui.screens.chat.message.gap :as gap]
+            [reagent.core :as reagent])
   (:require-macros [status-im.utils.views :refer [defview letsubs]]))
 
 (defn add-contact-bar [public-key]
@@ -153,7 +154,8 @@
   (letsubs [intro-status [:chats/current-chat-intro-status]
             height       [:chats/content-layout-height]
             input-height [:chats/current-chat-ui-prop :input-height]
-            {:keys [:lowest-request-from :highest-request-to]} [:chats/range]]
+            {:keys [:lowest-request-from :highest-request-to]} [:chats/range]
+            all-loaded?  [:chats/all-loaded?]]
     (let [icon-text  (if public? chat-id name)
           intro-name (if public? chat-name name)]
       ;; TODO This when check ought to be unnecessary but for now it prevents
@@ -237,12 +239,32 @@
 
 (defonce messages-list-ref (atom nil))
 
+(def ^:const initial-limit 5)
+(def ^:const second-initial-limit-step 6)
+(def second-limit (+ initial-limit second-initial-limit-step))
+(def ^:const third-initial-limit-step 11)
+(def ^:const default-limit-step 20)
+(def ^:const initial-threshold 0.5)
+(def ^:const default-threshold 2)
+
+(defonce messages-limit (reagent/atom initial-limit))
+
+(defn increment-limit [lim]
+  (+ lim
+     (case lim
+       initial-limit second-initial-limit-step
+       second-limit third-initial-limit-step
+       default-limit-step)))
+
 (defview messages-view
   [{:keys [group-chat chat-id pending-invite-inviter-name] :as chat}
    modal?]
   (letsubs [messages           [:chats/current-chat-messages-stream]
             current-public-key [:account/public-key]]
-    {:component-did-mount
+    {:component-will-mount
+     (fn []
+       (reset! messages-limit initial-limit))
+     :component-did-mount
      (fn [args]
        (when-not (:messages-initialized? (second (.-argv (.-props args))))
          (re-frame/dispatch [:chat.ui/load-more-messages]))
@@ -250,8 +272,12 @@
                            {:messages-focused? true
                             :input-focused?    false}]))}
     (let [no-messages (empty? messages)
+          m-limit     @messages-limit
+          threshold   (if (= m-limit initial-limit)
+                        initial-threshold
+                        default-threshold)
           flat-list-conf
-          {:data                      messages
+          {:data                      (take m-limit messages)
            :ref                       #(reset! messages-list-ref %)
            :footer                    [chat-intro-header-container chat no-messages]
            :key-fn                    #(or (:message-id %) (:value %))
@@ -264,9 +290,12 @@
                                           :idx                idx
                                           :list-ref           messages-list-ref}])
            :inverted                  true
-           :onEndReached              #(re-frame/dispatch
-                                        [:chat.ui/load-more-messages])
-           :enableEmptySections       true
+           :onEndReachedThreshold     threshold
+           :onEndReached              (fn []
+                                        (swap! messages-limit increment-limit)
+                                        (when (> @messages-limit (count messages))
+                                          (re-frame/dispatch
+                                           [:chat.ui/load-more-messages])))
            :keyboardShouldPersistTaps :handled}
           group-header {:header [group-chat-footer chat-id]}]
       (if pending-invite-inviter-name


flexsurfer:
just wondering why is it so ` ~20 message currently produce 2200-2400 calls ` ? maybe we could optimize message itself?

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -299,6 +344,8 @@
                     :on-layout (fn [e]
                                  (re-frame/dispatch [:set :layout-height (-> e .-nativeEvent .-layout .-height)]))}
         [chat-toolbar current-chat public? modal?]
+        [react/text {:style {:height 14}}
```
flexsurfer: leftover?

rasom:
> just wondering why is it so ~20 message currently produce 2200-2400 calls ? maybe we could optimize message itself?

yep I have a branch in progress with 18-20% improvement on this number. We definitely have some redundant components there. But still impact of such changes will be much less noticeable

src/status_im/ui/screens/chat/views.cljs
rasom: in current build when you open chat screen first time `limit enabled: true` is shown which means it is rendered with new limits, the next time `limit enabled: false` is shown -  that's an old version. Just for testing

flexsurfer: i mean, do you plan to remove this after testing?

rasom: Sure

flexsurfer (approved):

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -236,21 +242,54 @@
...
+(def initial-threshold 0.5)
+(def default-threshold 2)
+
+(defonce messages-limit (reagent/atom initial-limit))
+
+(defonce limit-enabled? (reagent/atom true))
+
+(defn increment-limit [lim]
+  (+ lim
+     (cond
```
yenda: maybe better to use case here and put the numerical values directly

rasom: how is it better?

yenda: case is an order of magnitude faster, and that not even considering that you are doing some additions here in the conditions that are not sure can be optimize by the compiler

```
(case lim
     ;; first step
     5 11
     ;; second step
     11 16
     ;; next steps
     (+ lim 20)
```

```
status-im.wallet.custom-tokens.core>  (time (dotimes [x 10000000] (cond (= x 1) 2 (= x 3) 4 :else 5)))
"Elapsed time: 5275.000000 msecs"
nil
status-im.wallet.custom-tokens.core> (time (dotimes [x 10000000] (case x 1 2 3 4 5)))
"Elapsed time: 24.000000 msecs"
nil
```

rasom: 1. relying on magic numbers is going to be an issue, I prefer to avoid doing this. If values are co-dependent, that should be expressed in the code.
2. The difference in efficiency in negligible compering to the whole rendering thing as long as this operation is not going to be executed repeatedly, repl on slow android device https://gist.github.com/rasom/b6a154bef04894c3d5ecff5fca6fbfcf

rasom: Eric, 10000000 times, really? :)

yenda: This isn't magic numbers, they are only used once in the code, what's the point of having 4 extra symbols in the registry for something so simple. Anyway it's nitpicking, but I think encapsulating the increment-limit within one function, in a simple case statement, without extra references dangling around is neater here.

yenda (approved):

src/status_im/ui/screens/chat/views.cljs
yenda: A docstring explaining why you use theses steps would be better than renaming the numbers 5 11 16 and 20

yenda: It's only magic numbers if there is no explanation

rasom: I disagree, because numbers are co-dependent. And as optimization this change improves nothing.

mandrigin (approved):

src/status_im/ui/screens/chat/views.cljs
yenda: Would that work for you then?

```
(def ^:const first-step 5)
(def ^:const second-step 11) 
(def ^:const third-step 16)
(def ^:const step-increment 20)

(defn increment-limit [lim]
(case lim
     first-step second-step
     second-step third-step
     (+ lim step-increment)))
```
Works because the steps are defined as constants and the compiler will replace them by the int

I would still like to understand why the increments are +6 then +4 then +20, looks like the real magic numbers here

rasom: > I would still like to understand why the increments are +6 then +4 then +20, looks like the real magic numbers here

At first, it is 5, then +6, then +11. Not `+6 then +4 then +20`.

And... well that's true, they are magic numbers (but not in terms of programming), and as I'm 99% sure we are going to change these numbers I prefer to code the relations between them. 

The way how I "inferred" those numbers: on a slower device (samsung a500d, 720 x 1280 pixels, 16:9 ratio (~294 ppi density)) I opened `#status` and `#status-core` chats during several days, and almost always not more than 5 messages were shown initially. 11 is the max number of messages which fits on that screen. Both number are device specific, and probably can be derived more precisely in some better way, but 11 is kind of a safe number for almost all devices, chances that more than 11 messages will be shown on the initial screen in a regular chat are quite low at the moment.

Serhy:
Great, @rasom,  no regressions and improved performance when navigating to chat with messages.
Improvement perfectly visible especially when open chat screen with 20+ messages required "heavy" 
text parsing:
Results for 'Galaxy Note 4' when navigating from "Chats" to the chat that has 20 messages with content  ``` *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* ```:
- with limit 9.5s
- without limit 5.3s -> **55% faster**

@rasom could you please remove code required for testing, we'll do smoke test of new builds and then merge.

rasom:
@Serhy "without limit" (check it when transition is ended) is actually the old version. So if it is faster, then there is no improvement.

Serhy:
> @Serhy "without limit" (check it when transition is ended) is actually the old version. So if it is faster, then there is no improvement.

Sorry, misplaced the values above:
The time it takes from tap on Chats screen upon chat with 20 messages till the messages appear 
- when no limit applied: 9.5s
- with limit (when 5 messages rendered firstly): 5.3s

rasom:
@Serhy that sounds better, I hope it is so :D

mandrigin:
Wait, that is the whole payload that can make the app choke? `*bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold* *bold*`

mandrigin:
And that's 9.5 **seconds** for **20** messages?

rasom:
> And that's 9.5 seconds for 20 messages?

that's not exactly a problem which this pr targeted, but that's definitely something we must fix

yenda:
@mandrigin @rasom we should store the messages as a tree of element based to avoid having to do any parsing at runtime, that is why I suggested Jan to do but he insisted on storing the rendering info separately which can be quite expensive in some cases. And since we already store the raw payload anyway we can always use it in case we do some migrations on rendering info

mandrigin:
I will create a separate issue for that

rasom:
@yenda do you mean we parse that data each time messages are rendered? If not, the way how we store data doesn't matter.

yenda:
@rasom afaik we parse them once to build up this kind of index that says where to cut the file to build a tree of elements when rendering, I think we should store this tree of elements in edn instead so that we only have to render

rasom:
>  could you please remove code required for testing, we'll do smoke test of new builds and then merge.

@Serhy done

Serhy:
I don't see any regressions with Android 6.0.1, Android 8.1, iOS 12.2, MacOS builds.
When chat view with 5+ messages in history opened - the first 5 messages is rendered, then next bunch of messages appear.

PR #8202, opened 2019-05-14
[experiment] lighter entry point

rasom:
10,781,645B -> 987,387B

galaxy s9:
131 frame (4.36667s) -> 61 frame (2.03333s)


src/status_im/ui/screens/browser/open_dapp/views.cljs
```diff
@@ -32,7 +32,7 @@
                 [vector-icons/icon :main-icons/browser {:color colors/gray}]]}]])
 
 (def dapp-image-data {:image (:dapp-store resources/ui) :width 768 :height 333})
-(def dapp-image (components.common/image-contain nil dapp-image-data))
+(def dapp-image [components.common/image-contain nil dapp-image-data])
```
flexsurfer: it was intentional because this is a constant we don't need to call it each time

flexsurfer: btw maybe we could consider this optimization in other places in the app

flexsurfer: or it will take more time on startup? hm

rasom: well that's just for testing: if it is called right there subscription is also initialized right away which causes exception at the moment

rasom:
> RN "hello world" example in js (no clojurescript) can be loaded in 0.48s on galaxy s9 (14% of current 3.5s) and in 0.73s on slow samsung a500d (6.6% of current 11.5s !!!). That's what we need to achieve in our app (or at least approach). So we definitely can't blame RN itself in that part.

1.23 galaxy s9 with cljs "hello world"

rasom:
js hw build on android №20

rasom:
samsung galaxy a500d

|  | js initialization | app startup  |
| -------- | -------- | -------- |
| `develop` | 11.3s| 13.9s|
| cljs "Hello world"| 2.7s | 3.73s|
| js "Hello world"  | 0.85s| 2.5s |

PR #8204, opened 2019-05-14
Ubuntu-server rejects connections not from localhost

vkjr:
Related to https://github.com/status-im/security-reports/issues/6

### Summary
This is a proposed fix to security issue in ubuntu-server.js.
We can check the address of socket connection and destroy it if it is not local.

**NOTE:** Commiting this file won't solve issue immediately, new binaries should be generated from updated `ubuntu-server.js` and used in all newly created packages

### Review notes
@corpetty, could you please take a look at proposed changes. If you think they will fix the problem, I'll commit the `ubuntu-server.js` in this PR and than generate new binaries for platforms to use them in installers.
I've checked with other laptop that when modified ubuntu-server.js runs it rejects incoming non-local connections and saves ability to accept local ones.

### Testing notes
Doesn't require testing until binaries updated

#### Platforms

* macOS
* Linux
* Windows

status: ready

### DIFF
--- a/ubuntu-server.js
+++ b/ubuntu-server.js
@@ -114,6 +114,18 @@ function rnUbuntuServer(readable, writable) {
   });
 }
 
+var closeDangerousConnection = function(sock) {
+  var remoteAddress = sock.remoteAddress;
+  if(remoteAddress.indexOf("127.0.0.1") == -1) {
+    console.log("WARN: connection not from localhost, will be closed: ", remoteAddress);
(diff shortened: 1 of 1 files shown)


vkjr:
@annadanchenko, FYI

corpetty (approved):
This should get the job done.  Thanks for getting on this quickly.

yenda (approved):

jakubgs (approved):

PR #8212, opened 2019-05-15
Fetch new version of base images with fixed ubuntu-server

vkjr:
Related to status-im/security-reports#6

### Summary

New binaries of `ubuntu-server` with security fix were generated for all platforms and uploaded to digitaloceanspaces
Nix files were updated to use fixed binaries

### Review notes
ubuntu-server binaries uploaded, nix files changed to download new versions

### Testing notes
Replacing ubuntu-server won't affect app logic but can affect app startup. So if it starts and able to login then `ubuntu-server` works fine with `react-native-desktop` and `realm.

#### Platforms
- macOS
- Linux
- Windows

##### Functional
- new account
- user profile updates

### Steps to test

- Open Status
- Ensure it runs
- Ensure you are able to login

status: ready

### DIFF
--- a/nix/desktop/linux/base-image/default.nix
+++ b/nix/desktop/linux/base-image/default.nix
@@ -5,13 +5,13 @@ with pkgs;
 let
   package = stdenv.mkDerivation rec {
     name = "StatusImAppImage";
-    version = "20181208";
+    version = "20190515";
 
     src =
       if stdenv.hostPlatform.system == "x86_64-linux" then
         fetchurl {
           url = "https://desktop-app-files.ams3.digitaloceanspaces.com/${name}_${version}.zip";
-          sha256 = "15c6p5v6325kj2whc298dn1dyigi0yzk2nzh1y10d03aqr4j8mp5";
+          sha256 = "0g7qa97cr0f9807sfd3khxiqz575i9kxi6lfda350ilaw8lnnfv2";
         }
       else throw "${name} is not supported on ${stdenv.hostPlatform.system}";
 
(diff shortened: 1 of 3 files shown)


mandrigin (approved):

pedropombeiro (approved):

jakubgs (approved):

annadanchenko:
@vkjr is it to include the fix into nightlies?

asemiankevich:
@annadanchenko @vkjr is 0.13 version expected here? Also, do we need to verify checksummes here again?

I've checked MacOS - looks okay with upgrade and fresh install, can recover and create account and receive some messages.

vkjr:
@annadanchenko, yes, this one fixes the same thing in `develop`, so will affect nightlies.

@asemiankevich, yes, version 0.13 expected, it is bumped in `develop` [3 days ago](https://github.com/status-im/status-react/commit/713a461415a479277658a66036917830436eb0f6)
I don't think there is a need to check hashes. You already checked that packages on our servers contain correct versions of `ubuntu-server`. This commit uses the same packages, so your check still valid.

annadanchenko:
Will do smoke test on Windows

Serhy:
Linux smoke test looks good to me. 
Can create new account, recover account, receive messages in 1-1 chat, offline history of public chats. 
shasum of ubuntu-server = `2b1219f249e71b6f2cffd6ed2f77c1a5747194c8879b9af5b83da0db61c16263` as expected.

annadanchenko:
same tests on windows are ok (create new account, recover account, receive messages in 1-1 chat, offline history of public chats, upgrade from previous stable version)

vkjr:
Thanks! Merged

PR #8215, opened 2019-05-15
[wip]RAM bundle

rasom:


yenda:
Since you have to fiddle with js imports, do you think you'll have time to give a shot to shadow-cljs? Among other things it also provides deterministic build order which should help with @PombeirP work on deterministic builds

rasom:
@yenda  can you please clarify how this would improve performance related stuff? Will usage of ` shadow-cljs` make some difference in performance, at least do we expect it to make? If that's about making the code better or something, but has nothing to do with performance,  i would skip it now. I would prefer to focus on things which have a direct effect on performance (or at least we expect them to have).

PR #8218, opened 2019-05-15
check also branch name when skipping rebase

jakubgs:
To help with https://github.com/status-im/status-react/pull/8214 failing due to rebasing.

### DIFF
--- a/ci/utils.groovy
+++ b/ci/utils.groovy
@@ -47,7 +47,7 @@ def doGitRebase() {
     println 'Skipping rebase for canary build...'
     return
   }
-  if (params.BUILD_TYPE == 'release') {
+  if (params.BUILD_TYPE == 'release' || branchName().startsWith('release/')) {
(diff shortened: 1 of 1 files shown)


pedropombeiro (approved):

ci/utils.groovy
```diff
@@ -47,7 +47,7 @@ def doGitRebase() {
     println 'Skipping rebase for canary build...'
     return
   }
-  if (params.BUILD_TYPE == 'release') {
+  if (params.BUILD_TYPE == 'release' || branchName().startsWith('release')) {
```
pedropombeiro: Not sure if we should make it
```suggestion
  if (params.BUILD_TYPE == 'release' || branchName().startsWith('release/')) {
```

vkjr (approved):

ci/utils.groovy
jakubgs: Agreed.

PR #8220, opened 2019-05-15
use CHANGE_TARGET to rebase on actual PR target

jakubgs:
Here's a better fix for rebase issues with https://github.com/status-im/status-react/pull/8214.

This way we rebase against the real target of PR, not `develop` always.

### DIFF
--- a/ci/utils.groovy
+++ b/ci/utils.groovy
@@ -51,10 +51,14 @@ def doGitRebase() {
     println 'Skipping rebase due to release build...'
     return
   }
+  def rebaseBranch = 'develop'
+  if (CHANGE_TARGET) { /* This is available for PR builds */
+    rebaseBranch = CHANGE_TARGET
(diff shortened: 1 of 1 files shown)


mandrigin (commented):
smart!

mandrigin (approved):

jakubgs:
Looks correct in the logs:
```
[2019-05-15T15:44:18.741Z] + git rebase develop
[2019-05-15T15:44:18.741Z] Current branch PR-8220 is up to date.
```

jakubgs:
Adjusted so it uses `develop` if `CHANGE_TARGET` is not set.

jakubgs:
Actually, now that I think about it, maybe rebasing makes only sense for PRs? Then we could throw out the other two `if`s.

vkjr (approved):

PR #8221, opened 2019-05-15
Feature/transaction signals

### LINKED ISSUE #8151: Use status-go subscriptions for wallet
# Problem
As described in https://discuss.status.im/t/proposal-fix-the-transactions-list/1190 the current wallet implementation relies on a loop that pull transaction data from a third party service to fetch the transaction list.

This causes: dependence to a third party service, delays and performance issues. 

An overview of the background required to understand the problem.
A problem description.

## Implementation
As requested in https://discuss.status.im/t/performances-before-moving-communication-protocol-to-status-go/1199 status-go now exposes an api to subscribe to `eth_newFilter`, `eth_newBlockFilter` signals.

The documentation is available here:
https://github.com/status-im/status-go/blob/e541f00e5845f45f471e8c1d39740c8b3a61d6c1/services/subscriptions/README.md

## Acceptance Criteria
The current loop is replaced by these subscription and signals.
Status no longer relies on etherscan for fetching the transaction list

yenda:
fixes #8151 

- remove the transaction fetching loop entirely to rely only on subscriptions
for live transactions and token transfer updates
- fetch token transfers history via etherscan API to lift the 100000 blocks
limit on token transfers history

- inbound token transfers are caught via a filter on ethlogs
- outbound token transfers and other transactions are caught by filtering
transaction in the current block that have the wallet address as `to` or `from` field

# Tests

This affects the transaction history. It should also affect the transaction messages in chat (they should update more frequently, not test yet).

Tested on Android:
- send and receive eth and token transactions, they should show up in the transactions list as soon as they are added to the blockchain

status: ready

### DIFF
--- a/src/status_im/accounts/login/core.cljs
+++ b/src/status_im/accounts/login/core.cljs
@@ -85,9 +85,9 @@
 (fx/defn initialize-wallet [cofx]
   (fx/merge cofx
             (models.wallet/initialize-tokens)
+            (transactions/initialize)
             (ethereum.subscriptions/initialize)
-            (models.wallet/update-wallet)
-            (transactions/start-sync)))
+            (models.wallet/update-wallet)))
 
 (fx/defn user-login [{:keys [db] :as cofx} create-database?]
   (let [{:keys [address password]} (accounts.db/credentials cofx)]
--- a/src/status_im/accounts/logout/core.cljs
+++ b/src/status_im/accounts/logout/core.cljs
@@ -1,7 +1,6 @@
 (ns status-im.accounts.logout.core
   (:require [re-frame.core :as re-frame]
             [status-im.chaos-mode.core :as chaos-mode]
-            [status-im.ethereum.transactions.core :as transactions]
             [status-im.i18n :as i18n]
             [status-im.init.core :as init]
             [status-im.node.core :as node]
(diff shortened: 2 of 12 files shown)


yenda:
@siphiuel @flexsurfer @jeluard please review

vitvly (approved):

flexsurfer (approved):

churik:
@yenda 
I have several questions:
- PR is related to performance.
1) Can you elaborate please on what screens / states are improvements to have some base for comparison and what exactly was improved?
2) Would be nice to have some clues about how it is possible to measure.
If not - I'll try to use smth similar to https://github.com/status-im/status-react/pull/8195#issuecomment-492614211
Please, tell me if you consider that this approach is not useful here.

And also I by mistake tested #8224 instead of this PR, so issues are here: https://github.com/status-im/status-react/pull/8224#issuecomment-493414051
Because this PR contains #8224 the same is here.

yenda:
@churik testing #8224 is fine it's the next after this one, thanks for the feedback I'll take care of it today.

1. in the transaction history screen, new transactions should appear sometimes more than 10 seconds before current develop. 
2. Besides new transactions appearing faster, it could be interesting to measure data consumption (without joining any chat to reduce noise)

PR #8224, opened 2019-05-17
Refactor/transaction details

yenda:
removing computations from transaction details view and simplifying related subscriptions

status: ready

### DIFF
--- a/src/status_im/ui/components/bottom_bar/core.cljs
+++ b/src/status_im/ui/components/bottom_bar/core.cljs
@@ -43,7 +43,6 @@
       {:nav-stack           :wallet-stack
        :content             {:title (i18n/label :t/wallet)
                              :icon  :main-icons/wallet}
-       :count-subscription  :get-wallet-unread-messages-number
        :accessibility-label :wallet-tab-button})
     {:nav-stack           :profile-stack
      :content             {:title (i18n/label :t/profile)
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -279,6 +279,18 @@
    (-> selected-participants
        (contains? element))))
 
+(re-frame/reg-sub
+ :ethereum/chain-keyword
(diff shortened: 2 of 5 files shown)


src/status_im/ui/components/bottom_bar/core.cljs
```diff
@@ -43,7 +43,6 @@
       {:nav-stack           :wallet-stack
        :content             {:title (i18n/label :t/wallet)
                              :icon  :main-icons/wallet}
-       :count-subscription  :get-wallet-unread-messages-number
```
flexsurfer, line 46: don't we want to use it for new transactions?

flexsurfer (approved):

src/status_im/ui/components/bottom_bar/core.cljs
yenda: We don't do anything with it yet, it's a subscription that returns 0. We can always add it back once it will be used

churik:
@yenda 

**1) New outgoing transactions are shown with "-{amount} confirmations" and token transactions are displayed as `0 ETH`**
On right side - current nightly, on left - PR.
I just send transaction in tokens to my contact and open transaction details to compare.
ezgif com-video-to-gif (2)
After relogin I can see right number of confirmations, but transactions in tokens are still shown as `O ETHro`

**2) Transactions in tokens which are not added to `Assets` are shown without amount and name in transaction history**
thkjsdf

**3) When tap on + in chat screen while offline "Network request failed" error**
Caught by e2e: https://github.com/status-im/status-react/pull/8224#issuecomment-493396661
Steps:
- go to offline
- login
- tap on "+"

churik:
well, I tested this PR instead of #8221 - by mistake.
It contains issues mentioned in https://github.com/status-im/status-react/pull/8224#issuecomment-493414051 - probably it would be easier for you to fix it.

PR #8227, opened 2019-05-17
Fix deep links tests

Serhy:
- for web views (app browsing and android chromium browsing) added method to swipe page so that element is visible to user
- sign_in method updated to accept all agreements (it was failing sometimes on pop-up about ROOTED device)


churik (approved):
separate methods for swiping on web elements!
Nice work @Serhy 👍

asemiankevich (approved):

PR #8231, opened 2019-05-18
[performance] wallet update

yenda:
### Summary
_a few files are moved around because of circular dependencies, this continues migration of wallet and ethereum code into their own module (previous in `ui.screens.wallet` and `utils.ethereums` respectively)_

`wallet-autoconfig-token` is a very expensive call on mainnet
because it checks the balance of every known token.

it is called:

* when wallet is refreshed by pulling
* when user goes on any wallet screen

this PR changes that by:

* calling it only when the wallet is initialized and there is no
visible-token configuration

it only calls update-wallet when a new transaction arrives

## Testing
I made the following experiment on Android:

* use mobile data and disable sync to reduce the noise from whisper
* recover `satoshi document engage inflict goddess auction rule unfair bid next buddy shy`
* go on ropsten network
* clear history in GlassWire (https://play.google.com/store/apps/details?id=com.glasswire.android&hl=en&rdid=com.glasswire.android&pli=1)
* restart app
* go to wallet history, check transaction details of first transaction, go back to wallet, pull to refresh, go to chat, kill app
* reopen app, go to transaction history, wait 2 minutes, pull to refresh, kill app
* switch 5 times between wallet and chat and wallet and transaction history

develop: 2.9mb
PR: 1.9 mb

The difference will be even bigger once persistence is added, the difference is already nice considering the PR pulls both transaction and token transfer history, and does it again everytime it misses a block. With transaction history this number will be very low.
status: ready


yenda:
@hesterbruikman could you check out this PR it should improve your wallet experience problems

src/status_im/constants.cljs
```diff
@@ -102,11 +102,7 @@
 
 (defn default-account-settings []
   {:web3-opt-in? true
-   :wallet       {:visible-tokens {:testnet #{:STT :HND}
-                                   :mainnet #{:SNT}
-                                   :rinkeby #{:MOKSHA :KDO}
-                                   :xdai    #{}
-                                   :poa     #{}}}})
+   :wallet       {:visible-tokens {}}})
```
flexsurfer, line 105: could you elaborate?

flexsurfer (approved):

src/status_im/constants.cljs
yenda: @flexsurfer I don't understand why we used to force these default tokens on the user, when `visible-tokens` is empty the wallet autoconfigures and adds the tokens for which the user has a balance.

dmitryn (approved):
What's the reason to have re-frame events prepended with `:TODO.remove` like `:TODO.remove/update-estimated-gas` and `TODO.remove/update-wallet`?

Not sure is it used somewhere or not.

UPD: nevermind, i see it's being removed in another PR https://github.com/status-im/status-react/pull/8232/files

yenda:
@dmitryn it is used in some places that are doing things wrong, but that I don't want to touch yet as the scope of these PRs is already big enough. It's to make sure it is obvious that this is not the way to do things

PR #8232, opened 2019-05-19
[refactoring] remove web3, clean up wallet effects

yenda:
- introduce json-rpc namespace, which provides `call` and `eth-call`,
a generic way of calling a json-rpc method taking care of conversions
and error handling
- remove web3 usage from wallet
- clean up effects, reducing the amount of computations when login in

status: ready


src/status_im/ethereum/json_rpc.cljs
```diff
@@ -0,0 +1,74 @@
...
+  (when-let [method-options (json-rpc-api method)]
+    (let [{:keys [id on-result subscription?]
+           :or {on-result identity
+                id        1
+                params    []}} method-options
+          on-error (or on-error
+                       #(log/error :json-rpc/error method :params params :error %))]
+      (if (nil? method)
+        (log/error :json-rpc/method-not-found method)
+        (status/call-private-rpc
```
flexsurfer, line 38: it would be good to have something like `(status/call-private-rpc "name" params)`

flexsurfer (approved):

src/status_im/ethereum/json_rpc.cljs
yenda: That is what I made initially but it turns out you always have different options so I figured it is better to have this generic function taking a map.

dmitryn (approved):

PR #8233, opened 2019-05-20
Refactor/call params

yenda:
- use `json-rpc/eth-call` and `wallet/eth-transaction-call` everywhere
- move all conversions to abi-spec

The goal of this PR is to unify the way we interact with contracts, so that future improvements can be made and impact all the callees. `json-rpc/eth-call` is for read only calls and `wallet/eth-transaction-call` is for transnational calls.

status: ready


flexsurfer (approved):

jeluard (approved):

asemiankevich:
@yenda all builds are failed

PR #8235, opened 2019-05-20
[#7983] Basic user added ERC-20 support

### LINKED ISSUE #7983: Basic user added ERC-20 support
# Problem
Currently Status only supports a limited fixed list which is manually maintained by core contributors.
If a user own some tokens which are not on that list, they would not appear in any of the Wallet screens.

The issue is particularly irritating when the user buys unsupported tokens on one of the exchange dapps listed in Status but then never sees them appear under assets.

## Implementation
Screen designs in figma - https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet_Source-of-Truth?node-id=160%3A3954

* new screens for custom token entry by address and token details (see designs)
* new database operations to store, retrieve and delete custom tokens. Store address and all values retrieved from the contract (name, symbol, decimals).
* addition of custom tokens loaded from database to all-tokens in app-db upon login

Validation:

* check if the address corresponds to an ERC-20 token contract
* check if the address is already in the token list (either built in or another custom token)
* check that contract defines name, symbol and decimals fields
* check that name or symbol of the added token isn't a duplicate of another token (either built in or custom)

Notes:

* use the same logo icon for all custom tokens, the default (i.e. unset) icon is ok

## Acceptance Criteria

* User can add new tokens by address (subject to some constraints). 
* After the addition, user can see the current balance of new tokens (if any) and send/receive them in wallet and chat.
* Transactions involving user added tokens should appear in Transaction history, in the same way as transactions involving tokens from the default list

## Notes
parent issue - #6939

flexsurfer:
fixes #7983

Tested: 
1) adding new custom token
2) select / deselect
3) send from wallet
4) send from chat

Not in the scope of this PR: remove custom token, will be implemented in separate issue https://github.com/status-im/status-react/issues/8236

https://www.youtube.com/watch?v=4U5wGHtGW-s

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -286,7 +286,6 @@
                                        :wallet-sign-message-modal
                                        :contact-code
                                        :wallet-onboarding-setup
-                                       :wallet-settings-assets
                                        :wallet-modal
                                        :wallet-onboarding-setup-modal
                                        :wallet-settings-hook)
@@ -301,7 +300,6 @@
                                      :recent-recipients
                                      :wallet-send-assets
                                      :wallet-request-assets
-                                     :wallet-settings-assets
                                      :wallet-modal} current-view)
                               [view {:background-color colors/white
                                      :position         :absolute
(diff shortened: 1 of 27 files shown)


dmitryn (approved):

jeluard (approved):

asemiankevich:
looks good to me:
- can add custom token
- can see its name , symbol and decimals are populated from contract
- can see added token in list of assets, also when sending / receiving 
- can enable / disable custom token
- can't add same token twice
- can't enter invalid contract address - validation error is in place

Unfortunately i was not able to buy APPC token so can't check transaction history with custom token added.

PR #8240, opened 2019-05-20
nix: Avoid bad practice of inheriting pkgs

pedropombeiro:

### Summary

Normally we should avoid passing nixpkgs to other recipes and instead pass the actual expressions that the called recipe needs, so that it becomes straightforward to override any of those values if needed.

status: ready


ios/Podfile.lock
```diff
@@ -139,6 +139,6 @@ SPEC CHECKSUMS:
   SSZipArchive: 41455d4b8d2b6ab93990820b50dc697c2554a322
   yoga: 128daf064cacaede0c3bb27424b6b4c71052e6cd
 
-PODFILE CHECKSUM: 6a6c6a3aad3b4979df3974dc03b0bf305426ce72
+PODFILE CHECKSUM: 1b73a7ab29d939e99e86434864c837afaf55851c
```
mandrigin, line 142: hmm, I wonder why did that change

mandrigin (approved):

rasom (approved):

PR #8245, opened 2019-05-21
[WIP] sesamum

flexsurfer:
Special mode for devs and QA
https://www.youtube.com/watch?v=LNkVbT6Rxsg

What's cool:
It's possible to mock screens and test UI
It's possible to test UI with long labels
It's possible to test UI on the smallest device screen without running app on this device separately

What's not cool:
code of tests incorporated into app code
i haven't found a way how to redef `i18b/label`, temporarily added atom, not cool
enabling mode incorporated into recovery code


yenda:
Very cool idea!

PR #8248, opened 2019-05-21
[Fixes #8066] Added native ENS registration

### LINKED ISSUE #8066: [ENS Names]  ENS username and name verification screens
Implement ENS username registration and verification screens

image

image

image

crossed elements are not in the scope of this PR

Acceptance criteria:

1. pixel perfect
2. it should be possible to register a new name in stateofus.eth domain and register a public key for this name
3. it should be possible to verify any ens name and register a public key for this name
4. verified name should persist in status
5. it's possible to remove this name from status

Figma: https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=510%3A3116

jeluard:
fixes #8066

### Summary

Implement ENS registration according to latest Figma designs.

_Note to testers_  Local names persistency is not part of this PR. Thus initial welcome screen is always the same even if you already registered some names using native registration.

Address and public key will be connected to names, and stateofus names can be be registered.

status: WIP


errorists:
👏 Great 👏 work 👏 @jeluard 

All of the issues I could find, if you run into trouble with using fontWeight, try asking @yenda, he's the typography master 👑 

Group

---

Group

---

Group 2

---

Artboard

---

Group 5

---

Group 4

src/status_im/utils/ethereum/stateofus.cljs
```diff
@@ -0,0 +1,20 @@
+(ns status-im.utils.ethereum.stateofus
+  (:require [clojure.string :as string]))
+
+(def domain "stateofus.eth")
+
+(defn- subdomain [username]
+  (str username "." domain))
+
+(def registrars
```
yenda, line 9: do we actually need this?

```
(ens/owner "0x112234455c3a32fd11230c42e7bccd4a84e02010" "stateofus.eth" println)
0x11d9f481effd20d76cee832559bd9aca25405841
```

jeluard: Good point, will drop it

src/status_im/utils/ethereum/ens.cljs
```diff
@@ -129,4 +129,11 @@
             ens-name
             #(addr % ens-name cb)))
 
+(defn get-pubkey
```
yenda: this already exists in resolver.cljs

yenda (changes requested):

flexsurfer:
@jeluard , @yenda introduced typography some time ago, so we don't need to set font styles manually anymore,  typography key should be used instead, you can see typography name in figma cc @errorists

src/status_im/utils/ethereum/stateofus.cljs
jeluard: Actually it's something we need to check statically and pretty often so doesn't appear to be worth the hassle and perf cost.

flexsurfer:
image

https://github.com/status-im/status-react/blob/496f3f1cc78703c3d629f9f458a0e966b04e5c61/src/status_im/ui/components/typography.cljs#L45

jeluard:
@flexsurfer @yenda I made some changes

src/status_im/subs.cljs
```diff
@@ -1633,3 +1634,27 @@
  :<- [:search/filter]
  (fn [[chats search-filter]]
    (apply-filter search-filter chats extract-chat-attributes)))
+
+;;ENS ==================================================================================================================
+
+(re-frame/reg-sub
+ :ens/state
+ (fn [db]
```
flexsurfer: for root keys we need to use `reg-root-key-sub`

src/status_im/subs.cljs
```diff
@@ -1633,3 +1634,27 @@
...
+   (get-in db [:ens :state])))
+
+(re-frame/reg-sub
+ :ens/username
+ (fn [db]
+   (get-in db [:ens :username])))
```
flexsurfer: `:ens` looks like too general, maybe we could use something like `:ens-usernames`

src/status_im/ui/screens/ens/events.cljs
```diff
@@ -0,0 +1,123 @@
+(ns status-im.ui.screens.ens.events
```
flexsurfer, line 1: i believe we don't use events in ui namespace, handlers should be in root events file, fx and functions in core.cljs

ghost:
### Pull Request Checklist

src/status_im/ui/screens/ens/views.cljs
```diff
@@ -0,0 +1,367 @@
...
+       [react/text {:style {:margin-top 32 :margin-bottom 8 :color colors/black :typography :header}}
+        (i18n/label :t/ens-get-name)]
+       [react/text {:style {:margin-top 8 :margin-bottom 24 :color colors/gray :font-size 15 :margin-horizontal 16 :text-align :center}}
+        (i18n/label :t/ens-welcome-hints)]
+       [welcome-item {:icon-label "1" :title (i18n/label :t/ens-welcome-point-1-title)}
+        [react/view {:flex-direction :row}
+         [react/nested-text
+          {:style {:color colors/gray}}
+          (i18n/label :t/ens-welcome-point-1)
+          [{:style {:color colors/black :text-decoration-line :underline}}
```
yenda: black doesn't need to be specified it's the default

src/status_im/utils/ethereum/stateofus.cljs
yenda: what do you mean we need to check it statically? what if we change the contract address for stateofus.eth?

src/status_im/ui/screens/ens/events.cljs
```diff
@@ -0,0 +1,123 @@
...
+     :data            data
+     :on-result       [:ens/on-registration-success]
+     :on-error        [:ens/on-registration-failure]}))
+
+(handlers/register-handler-fx
+ :ens/register
+ (fn [{db :db} [_ data]]
+   (let [transaction (prepare-extension-transaction data)]
+     ;; Last argument is nil: no need to estimate gas
+     (wallet/open-modal-wallet-for-transaction db transaction nil))))
```
yenda: use `wallet/eth-transaction-call` here

src/status_im/utils/ethereum/core.cljs
```diff
@@ -108,6 +110,12 @@
   (when (= 66 (count s))
     (normalized-address (subs s 26))))
 
+(defn coordinates [public-key]
```
yenda, line 122: it is better to always put the param list of the second line, a good way to convince yourself is to write a docstring for this function

src/status_im/ui/screens/ens/views.cljs
```diff
@@ -0,0 +1,367 @@
...
+   (when (= state :registering)
+     [registration-bottom-bar props])])
+
+(defn- toolbar []
+  [toolbar/toolbar nil
+   [toolbar/nav-button (actions/back #(re-frame/dispatch [:ens/navigate-back]))]
+   [toolbar/content-title (i18n/label :t/ens-your-username)]])
+
+(views/defview register []
+  (views/letsubs [{:keys [registrar]}          [:get-screen-params :ens-register]
```
yenda: make just one subscription for the register screen with all the data you need

src/status_im/subs.cljs
yenda: and clearly here  there should be only one root key for ens and the rest is derived subscription. but looking at how they are used it should even rather be screen subscriptions that aggregate from an :ens root key and a few others

src/status_im/subs.cljs
jeluard: I think :ens is fine

src/status_im/ui/screens/ens/events.cljs
jeluard: There are plenty events files.

src/status_im/subs.cljs
jeluard: Not clear what you mean. Do you have an exemple?

src/status_im/utils/ethereum/stateofus.cljs
jeluard: Well same as if we change any contract address?

src/status_im/ui/screens/ens/events.cljs
jeluard: Can't find this function.

src/status_im/utils/ethereum/ens.cljs
yenda: It still already exists?

src/status_im/ui/screens/ens/events.cljs
yenda: Which is wrong there shouldnt be any

src/status_im/ui/screens/ens/events.cljs
yenda: It's in my PR which should get merged today

src/status_im/utils/ethereum/stateofus.cljs
yenda: @decanus do you have any recommendation here? Should we rather rely on hardcoded contract addresses for the registrar or solve it with the ens resolver contract?

src/status_im/ui/screens/ens/events.cljs
```diff
@@ -0,0 +1,123 @@
...
+ ::ens/resolve
+ (fn [[registry name cb]]
+   (ens/get-addr registry name cb)))
+
+(defn- on-resolve [registry custom-domain? name address public-key s]
+  (if (and (seq address) (= address (ethereum/normalized-address s)))
+    (resolver/pubkey registry name
+                     (fn [ss]
+                       (if (= ss public-key)
+                         (re-frame/dispatch [:ens/set-state :connected])
```
yenda, line 27: don't do double dispatch, do one callback event that does these two things

src/status_im/ui/screens/ens/events.cljs
```diff
@@ -0,0 +1,123 @@
...
+                         (re-frame/dispatch [:ens/set-state :registrable]))))
+    (if (and (nil? s) (not custom-domain?)) ;; No address for a stateofus subdomain: it can be registered
+      (re-frame/dispatch [:ens/set-state :registrable])
+      (re-frame/dispatch [:ens/set-state :unregistrable]))))
+
+(defn- chain [{:keys [network] :as db}]
+  (let [network (get-in db [:account/account :networks network])]
+    (ethereum/network->chain-keyword network)))
```
yenda: this is already done by ethereum/chain-keyword which takes db, so you don't actually need this functions

src/status_im/ui/screens/events.cljs
```diff
@@ -9,6 +9,7 @@
             status-im.web3.events
             status-im.ui.screens.add-new.new-chat.navigation
             status-im.ui.screens.profile.events
+            status-im.ui.screens.ens.events
```
yenda, line 12: these are all wrong and should be refactored at some point, don't add a new one

src/status_im/subs.cljs
yenda: I agree :ens is fine but doing a get-in from the root app-db is not fine we are trying to optimize subscriptions now, this is going to get triggered for every change in app-db

src/status_im/subs.cljs
yenda: https://github.com/status-im/status-react/blob/develop/src/status_im/subs.cljs#L1199-1212 is a good exemple

Basically you need to think about each subscription you make as a cache, in your case you want to cache the :ens key of app-db as a root key, so that the subscriptions you use it in are only recomputed when the something in :ens is changed.

yenda (changes requested):

flexsurfer:
image

errorists:
Thanks Andrey for the Android demo. Some things I've noticed: 

👉 On Android, whenever we bold a font, it's no longer using the Inter font but rather the system default Roboto, not an issue contained to Julien's work, rather making a general note. cc @yenda because he loves fonts :P

Group5

- [ ] layout here needs some adjustments, especially considering smaller screens. The empty spaces are left too large, content could use additional vertical centring

Group

- [ ] the text and button should be vertically centred

Group2

- [ ] a small one, but both the top and bottom margin should have the same value here

Group3

- [ ] is there an underline here? if yes pls remove it 🙏

flexsurfer:
image

src/status_im/ui/screens/ens/events.cljs
jeluard: Which 2 things?

yenda: I missed the if

yenda:
could you fixup all these commit it can't be reviewed in that state

yenda (changes requested):
- there is still an ens events namespace, the best solution is to move to ens.core (outside of ui) and use the fx/defn attribute map like this: `{:events [:this-in-an-event]}` to declare the events.
- duplicated ens subscription
- internal transactions should use eth-transaction-call

jeluard:
@yenda Not done, please don't review.

PR #8250, opened 2019-05-21
[#8249] "" when grant access to ens dapp shown, it prevents the name …

### LINKED ISSUE #8249: "" when grant access to ens dapp shown, it prevents the name to be registered
# Problem
Can't register ens username . 

## Steps

1. open status on mainnet
1. open names.statusnet.eth dapp
1. select any free username and tap Grant access -> note that "" shown instead of dapp name
image
1. tap Allow -> Register -> Sent SNT

flexsurfer:
fixes #8249

### DIFF
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -162,11 +162,11 @@
   (let [base32hash (-> (.encode js-dependencies/hi-base32 (alphabase.base58/decode hash))
                        (string/replace #"=" "")
                        (string/lower-case))]
-    (str "https://" base32hash ".infura.status.im")))
+    (str base32hash ".infura.status.im")))
 
 (defmethod storage-gateway :swarm
   [{:keys [hash]}]
-  (str "https://swarm-gateways.net/bzz:/" hash))
+  (str "swarm-gateways.net/bzz:/" hash))
 
 (fx/defn resolve-ens-multihash-success
   [{:keys [db] :as cofx} m]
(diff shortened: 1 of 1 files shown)


src/status_im/browser/core.cljs
```diff
@@ -179,7 +179,7 @@
                                assoc
                                :url (str gateway path)
                                :resolving? false)
-                       (assoc-in [:browser/options :resolved-ens host] gateway))})))
+                       (assoc-in [:browser/options :resolved-ens host] (subs gateway 8)))})))
```
jeluard: Magic number? Can we use a separate function for that?

jeluard (changes requested):

yenda (approved):

flexsurfer:
@jeluard done

src/status_im/browser/core.cljs
yenda: it is not a real fix anyway

jeluard:
@flexsurfer Thanks!

jeluard (approved):

src/status_im/browser/core.cljs
```diff
@@ -179,7 +179,7 @@
                                assoc
                                :url (str gateway path)
                                :resolving? false)
-                       (assoc-in [:browser/options :resolved-ens host] gateway))})))
+                       (assoc-in [:browser/options :resolved-ens host] (string/replace gateway #"https://" "")))})))
```
yenda: nah if you are going to remove the https prefix just don't put it there in the first place

`(str "https://" gateway path)`

yenda (changes requested):
don't add https:// in the first place, also it is ok for a hotfix but I think there is more to it that this

flexsurfer:
@yenda done

yenda (approved):

churik:
Checked on OS 11.4.1 (IPhone 7), Android 8 (LG V20):
- can grant access to `names .statusnet.eth`
- can grant access to other apps (on Mainnet and Ropsten)
- can revoke and grant access again

PR #8252, opened 2019-05-22
RAM bundle

rasom:
1. I haven't found the way to create a release bundle with cljs compiled with advanced optimizations. It just fails with meaningless exception. The reason why I didn't realize that from the beginning is that our PR builds are not bundled as releases. I was under impression that they are. 
2. Thus there is no impressive boost because of this change. The big part of it was minimized cljs build. So now it is only **14%** of js loading, and even less of the whole startup. 
3. I will try to figure out how to integrate cljs with advanced optimization in a separate task. It works when js is bundled for debugging, so it still might be possible with release version.
4. Some modules still might be stripped out of initial bundle, but this doesn't really matter at the moment because bundle is **11,697,115B** where **11,097,753B** is cljs. So the first thing to minimize here is cljs. (release is **13,956,525B**)

## What's done in this PR
- almost all modules (except "react-native", "realm" and few more) are required only by request, not on app initialisation 
- RAM bundle is enabled on android

**testing notes**:
App is bundled as a release, this should be reverted before merging as we build PR differently at the moment.
In order to compare startup time:
- install the last release version and measure time necessary for loading intro screen
- remove app
- install this PR's version
- measure startup time again

status: ready


src/status_im/utils/js_require.clj
```diff
@@ -0,0 +1,9 @@
+(ns status-im.utils.js-require)
```
yenda, line 1: is it a lazy-require macro? maybe name it that way to make it more explicit?

yenda (commented):
Does it have an impact on runtime? Might load faster but the modules still have to be required at some point right? and then it needs to dereferenced everytime it is used?

src/status_im/utils/ethereum/macros.clj
```diff
@@ -21,10 +29,19 @@
   (mapv #(assoc-in % [:icon :source] (token->icon network %)) tokens))
 
 (defn network->icon [network]
-  (let [s (str "./resources/images/tokens/" (name network) "/0-native.png")]
+  (let [s     (str "./resources/images/tokens/" (name network) "/0-native.png")
```
yenda: can you explain this change?

src/status_im/utils/ethereum/macros.clj
```diff
@@ -1,13 +1,21 @@
 (ns status-im.utils.ethereum.macros
-  (:require [clojure.string :as string]
-            [clojure.java.io :as io]))
+  (:require [clojure.java.io :as io]))
 
 (defn token-icon-path
   [network symbol]
-  (let [s (str "./resources/images/tokens/" (name network) "/" (name symbol) ".png")]
+  (let [s     (str "./resources/images/tokens/" (name network) "/" (name symbol) ".png")
```
yenda: isn't it going to make the performance of the manage token asset list even worst?

ci/android.groovy
```diff
@@ -26,7 +26,7 @@ def bundle() {
       )
     ]) {
       nix.shell(
-        "./gradlew assemble${target.capitalize()} ${gradleOpt}",
+        "./gradlew assembleRelease ${gradleOpt}",
```
mandrigin: so we always do releases now?

mandrigin (dismissed):

ci/android.groovy
rasom: that's for testing

src/status_im/utils/ethereum/macros.clj
rasom: we can load this data right after startup if it will, though we definitely don't need these images on startup

src/status_im/utils/ethereum/macros.clj
rasom: Do you explain why we need to load messages only be request and not right away or what exactly?

src/status_im/utils/js_require.clj
rasom: kind of, ok

ci/android.groovy
rasom: see "testing notes:" in PR~ description

yenda:
Aren't lazy loading and ram-bundle on Android two different things? What is the impact of each change individually?

rasom:
@yenda https://facebook.github.io/react-native/docs/performance#configure-preloading-and-inline-requires

yenda (approved):
Kudos for the work I guess it must have been a lot of pain

flexsurfer (approved):

rasom:
> Does it have an impact on runtime? Might load faster but the modules still have to be required at some point right? and then it needs to dereferenced everytime it is used?

Not very noticeable, I can't say that I feel any extra slowness because of this, probably because app is slow as it is 

Dereferencing is definitely extra cheap, the only thing which might have impact is requiring of modules.

asemiankevich:
e2e fails should not be related to this PR

Serhy:
The performance improvement part, - looks good to me (tested iOS, Android).
With Galaxy Note 4 cold boot (from tap on Status icon till login screen is shown) with this changes it takes ~6.7 seconds VS ~7.35 seconds in Release 0.13.0 build (~10% faster than in release)
 
There is one regression found with recently implemented 'Add custom token' feature (not reproduced with nightly 27th of May develop build)
App crashes when tap 'Add custom token' option from 'Wallet' -> 'Assets' view:

Screenshot 2019-05-27 at 15 35 51

[Status.log](https://github.com/status-im/status-react/files/3223755/Status.log)

@rasom, could you take a look, please, at what is missing with `Add custom token` and also revert back the change for Android app packages made of PRs are different than release/nightly at the same time?

rasom:
@Serhy 10% is exactly what is expected.
Just pushed a new commit with fix.

Serhy:
@rasom could you please revert the change for Android PR-builds and let's merge!
(failed autotest was due to Android build is not `im.status.ethereum.pr` app package)

rasom:
@Serhy sure, on it

PR #8265, opened 2019-05-22
[#8256] do not call getApplicationInfo every time card connected

### LINKED ISSUE #8256: Error java.lang.IllegalArgumentException: 16 > 0
### Description
During keycard setup the error happens some times:

`Error java.lang.IllegalArgumentException: 16 > 0`

### Reproduction
No exact steps to reproduce but it may happen during pairing or generating keys step.

### Device
This user is using Galaxy S10+

### Solution
I expect it might happen because of unnecessary call to `getApplicationInfo()` method on "card connected" event.

dmitryn:
fixes #8256 

### Summary

Call `getApplicationInfo` explicitly when it's needed rather than calling it automatically on every card connected event.

Should fix `java.lang.IllegalArgumentException: 16 > 0` sporadic error during setup.

status: ready

### DIFF
--- a/src/status_im/hardwallet/core.cljs
+++ b/src/status_im/hardwallet/core.cljs
@@ -301,6 +301,10 @@
   [{:keys [db]}]
   {:db (assoc-in db [:hardwallet :on-card-read] nil)})
 
+(fx/defn clear-on-card-connected
+  [{:keys [db]}]
+  {:db (assoc-in db [:hardwallet :on-card-connected] nil)})
+
 (fx/defn dispatch-event
   [_ event]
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

yenda (approved):

asemiankevich:
well i was not able to reproduce the error in this PR, I've seen this one https://github.com/status-im/status-react/issues/8189 but not the mentioned error. I am not sure for 100% that the bug is fixed, but i did not find any regression with this fix so lets deliver this in new build and see how it goes.

PR #8272, opened 2019-05-23
save keycard keyUID and use it for login

dmitryn:
### Summary
Based on keycard integration doc https://notes.status.im/jZudDaprS2mSqQLuJ_1_-g?view#SELECT-checks

* add `keycard-key-uid` to Account to store keyUID
* use it for login
* add temporary support to login by keycard-instance-uid (for old accounts created with keycard)

### Testing notes
Setup and login flow should be tested.

status: ready

### DIFF
--- a/src/status_im/data_store/realm/schemas/base/account.cljs
+++ b/src/status_im/data_store/realm/schemas/base/account.cljs
@@ -242,3 +242,7 @@
 (def v22 (update (update v21 :properties dissoc :mainnet-warning-shown?)
                  :properties merge
                  {:mainnet-warning-shown-version {:type :string :optional true}}))
+
+(def v23 (assoc-in v22
+                   [:properties :keycard-key-uid]
+                   {:type :string :optional true}))
--- a/src/status_im/hardwallet/core.cljs
+++ b/src/status_im/hardwallet/core.cljs
@@ -30,6 +30,14 @@
          (filter #(= keycard-instance-uid (:keycard-instance-uid %)))
          first)))
 
+(defn- find-account-by-keycard-key-uid
+  [db keycard-key-uid]
+  (when keycard-key-uid
+    (->> (:accounts/accounts db)
+         vals
+         (filter #(= keycard-key-uid (:keycard-key-uid %)))
+         first)))
+
 (defn get-pairing
   ([db]
    (get-pairing db (get-in db [:hardwallet :application-info :instance-uid])))
@@ -257,20 +265,35 @@
        :hardwallet/get-keys {:pairing pairing
                              :pin     pin}})))
 
+(defn- get-account-for-login
+  "Finds account by key-uid.
+  Temporary, it also searches by instance-uid which was used before key-uid was introduced -
+  workaround should be removed before release."
+  [db key-uid instance-uid auto-login?]
+  (if auto-login?
+    (or
+     (find-account-by-keycard-key-uid db key-uid)
+     ; TODO: remove before release
+     (let [acc (find-account-by-keycard-instance-uid db instance-uid)]
+       (when-not (:keycard-key-uid acc)
+         acc)))
+    (get-in db [:accounts/accounts (get-in db [:accounts/login :address])])))
+
 (fx/defn login-with-keycard
   [{:keys [db] :as cofx} auto-login?]
-  (let [account-login-address (get-in db [:accounts/login :address])
-        account-was-manually-selected? account-login-address
-        account-instance-uid (get-in db [:accounts/accounts account-login-address :keycard-instance-uid])
+  (let [keycard-key-uid (get-in db [:hardwallet :application-info :key-uid])
         keycard-instance-uid (get-in db [:hardwallet :application-info :instance-uid])
-        account (find-account-by-keycard-instance-uid db keycard-instance-uid)
-        account-mismatch? (if account-was-manually-selected?
-                            (not= account-instance-uid keycard-instance-uid)
-                            (nil? account))
+        account (get-account-for-login db keycard-key-uid keycard-instance-uid auto-login?)
+        account-key-uid (get account :keycard-key-uid)
+        account-instance-uid (get account :keycard-instance-uid)
+        account-mismatch? (if auto-login?
+                            (nil? account)
+                            (if (:keycard-key-uid account)
+                              (not= account-key-uid keycard-key-uid)
+                              (not= account-instance-uid keycard-instance-uid)))
         pairing (:keycard-pairing account)]
-
     (cond
-      (empty? keycard-instance-uid)
+      (empty? keycard-key-uid)
       (fx/merge cofx
                 {:utils/show-popup {:title   (i18n/label :t/no-account-on-card)
                                     :content (i18n/label :t/no-account-on-card-text)}}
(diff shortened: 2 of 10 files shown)


flexsurfer (approved):

dmitryn:
@jeluard @rasom please review

src/status_im/data_store/realm/schemas/base/account.cljs
```diff
@@ -242,3 +242,7 @@
 (def v22 (update (update v21 :properties dissoc :mainnet-warning-shown?)
                  :properties merge
                  {:mainnet-warning-shown-version {:type :string :optional true}}))
+
+(def v23 (assoc-in v22
+                   [:properties :keycard-key-uid]
```
yenda, line 247: could you explain the role of this uid from a security pov? this part of the db is only encrypted at the app level and not the account level

src/status_im/events.cljs
```diff
@@ -1015,6 +1015,11 @@
  (fn [cofx _]
    (hardwallet/get-application-info cofx nil nil)))
 
+(handlers/register-handler-fx
+ :hardwallet/save-key-uid-to-account
```
yenda: you can now declare this event by adding `{:events  :hardwallet/save-key-uid-to-account]}` in the attribute map of `hardwallet/save-key-uid-to-account` (and use an event name that reflects what happened rather than the action you want to take

src/status_im/hardwallet/core.cljs
```diff
@@ -257,20 +265,32 @@
        :hardwallet/get-keys {:pairing pairing
                              :pin     pin}})))
 
+(defn- get-account-for-login
+  [db key-uid instance-uid auto-login?]
+  (if auto-login?
+    (or
+     (find-account-by-keycard-key-uid db key-uid)
+     ; NOTE: temporary fix to allow accounts without key-uid to login using instance-uid
```
yenda: why is it a temporary fix? what is need to make it permanent?

src/status_im/hardwallet/core.cljs
```diff
@@ -1315,6 +1353,8 @@
                        (assoc-in [:hardwallet :wallet-address] wallet-address)
                        (assoc-in [:hardwallet :encryption-public-key] encryption-public-key)
                        (assoc-in [:hardwallet :keycard-instance-uid] (or instance-uid' instance-uid))
+                       (assoc-in [:hardwallet :keycard-key-uid] key-uid)
+                       (assoc-in [:hardwallet :application-info :key-uid] key-uid)
```
yenda: why does the information need to be present in two different places?

src/status_im/data_store/realm/schemas/base/account.cljs
dmitryn: @yenda key-uid is uid of master key on keycard so we can link it to status account. 

More info on key UID from keycard docs https://status.im/keycard_api/sdk_create_wallet.html
> The UID is tied to the key itself (is derived from the public key) so it will change if the wallet on card is replaced. The key UID is also part of the response of the applet selection command, so the wallet can be identified immediately upon selection.

Since it's derived from public key i assume it's safe to store it on client unencrypted.

src/status_im/hardwallet/core.cljs
dmitryn: @yenda To remove this workaround we need all accounts created with keycard before this fix to have key-uid. 
I've added `hardwallet/save-key-uid-to-account` which is called when keycard account does login to ensure old accounts will eventually have key-uid stored.

src/status_im/events.cljs
dmitryn: @yenda right, i'm aware that fx/defn now have this feature. I can start using it here.

I'm trying to follow event name conventions, but for this particular event i'm not sure i can come up with good name of what's happened because chain of events should happen in order to trigger this like: `on-get-keys-success -> on-get-application-info-success -> save-key-uid-to-account`

src/status_im/hardwallet/core.cljs
dmitryn: @yenda 
`[:hardwallet :keycard-key-uid]` is used in account creation and will be cleared. 
`[:hardwallet :application-info :key-uid]` is be used for login. Also, it updates on every card tap when we get app-info from keycard.

src/status_im/hardwallet/core.cljs
yenda: @dmitryn it is not possible to currently create an account with a keycard in the releases right? So the intention is to remove this for the next release or something?

dmitryn: @yenda yes, it should be removed before official release. Now apk with keycard functionality was only released unofficially in #status-keycard channel. So this fix is only for some of the very pro-active users who might actually started playing with keycard before the release.

src/status_im/hardwallet/core.cljs
```diff
@@ -689,6 +709,21 @@
     {:hardwallet/get-application-info {:pairing    pairing'
                                        :on-success on-card-read}}))
 
+; NOTE: temporary fn to migrate old accounts without keycard-key-uid saved in account,
```
yenda: defn supports docstrings

dmitryn: moved to docstring

src/status_im/hardwallet/core.cljs
yenda: could you add this information in the docstring just in case it gets forgotten so that someone reading that code knows what'sup

src/status_im/events.cljs
yenda: if this is a chain of event without any user action you might as well not use an event for the intermediate events that don't need up to date cofx and don't produce effects themselves, I'm doing this for swarm for instance, where only errors in the chain dispatch events.

yenda: it might be too many changes that you don't want to make in this PR that's ok, but think about it for the future, if you have these sequences of event just encapsulate them with multiple levels of callback

src/status_im/hardwallet/core.cljs
yenda: ok I see so they are the same only until next login. What about the one in `:account/account`? Is it `[:hardwallet :keycard-key-uid]` that is getting cleared and placed there afterwards?

dmitryn: @yenda yes, `[:hardwallet :keycard-key-uid]` gets stored into `account/account` and gets cleared after

yenda: could you add a comment there, also why is there so many assoc-in wouldn't it be better at this point to prepare the hardwallet map or merge it with a map containing these changes?

src/status_im/hardwallet/core.cljs
dmitryn: added

src/status_im/hardwallet/core.cljs
```diff
@@ -1326,6 +1367,8 @@
                        (assoc-in [:hardwallet :wallet-address] wallet-address)
                        (assoc-in [:hardwallet :encryption-public-key] encryption-public-key)
                        (assoc-in [:hardwallet :keycard-instance-uid] (or instance-uid' instance-uid))
```
yenda: this one is actually an `(update-in [:hardwallet :keycard-instance-uid] #(or % instance-uid))`

yenda (approved):

.env.jenkins
```diff
@@ -6,7 +6,7 @@ ETHEREUM_DEV_CLUSTER=1
 EXTENSIONS=1
 FLEET=eth.beta
 GROUP_CHATS_ENABLED=1
-HARDWALLET_ENABLED=0
+HARDWALLET_ENABLED=1
```
yenda: does this mean that it is enabled on releases now?

src/status_im/hardwallet/core.cljs
dmitryn: good idea, i'm going to put all the data into map and `assoc` it once

.env.jenkins
dmitryn: @yenda  it's only for QA testing, will be disabled before merge

src/status_im/hardwallet/core.cljs
dmitryn: yes, i've changed it to `update`

src/status_im/events.cljs
dmitryn: Good suggestions in general, but for this particular case i'd prefer to leave it as is because this code will removed before release.

src/status_im/hardwallet/core.cljs
```diff
@@ -1310,22 +1352,14 @@
...
               {:db (-> db
-                       (assoc-in [:hardwallet :whisper-public-key] whisper-public-key')
-                       (assoc-in [:hardwallet :whisper-private-key] whisper-private-key)
-                       (assoc-in [:hardwallet :whisper-address] whisper-address)
-                       (assoc-in [:hardwallet :wallet-address] wallet-address)
-                       (assoc-in [:hardwallet :encryption-public-key] encryption-public-key)
-                       (assoc-in [:hardwallet :keycard-instance-uid] (or instance-uid' instance-uid))
+                       (assoc-in [:hardwallet :account] (-> account-data
+                                                            (update :whisper-public-key #(str "0x" %))
+                                                            (update :instance-uid #(or (get-in db [:hardwallet :account :instance-uid])
```
yenda: get-in already has an arity for default value so you can just write it
`(get-in db [:hardwallet :account :instance-uid] %)`

yenda:
@dmitryn looks like it is ready to merge are you waiting for something?

PR #8274, opened 2019-05-24
Update the defn macro

yenda:
This macro should make the events.cljs ns obsolete over time.

the defn macro knows register the function as an events under the
keywords provided in the :events key of the attributes map. It also
adds the interceptors provided in the :interceptors map

exemple:

```clojure
(fx/defn hello4
  {:doc "this function is useless as well"
   :events [:test/valid1 :test/valid2]}
  [{:keys [db]} b]
  {:db (assoc db :a b) :b (:a db)})
```

## Testing

Tested in Android emulator.

This PR is more of a dev experience improvement, but since the macro is basically behind everything happening in status-react, it is still worth testing. 
In this PR it is used in the login events, so to test it:
- create an account, 
- logout play around a bit. 

I don't think the app would even be able to login if there was a bug in this PR

status: ready

### DIFF
--- a/src/status_im/utils/fx.clj
+++ b/src/status_im/utils/fx.clj
@@ -1,6 +1,15 @@
 (ns status-im.utils.fx
   (:refer-clojure :exclude [defn]))
 
+(defn- register-events
+  [events interceptors name args]
+  (mapv (fn [event]
+          `(status-im.utils.handlers/register-handler-fx
+            ~event
+            ~interceptors
+            (fn [cofx# [_# ~@args]] (~name cofx# ~@args))))
+        events))
+
 (defmacro defn
   "Defines an fx producing function
   Takes the same arguments as the defn macro
@@ -12,7 +21,8 @@
   Notes:
   - destructuring of cofx is possible
   - supports docstring
-  - supports attr-map
+  - supports attr-map with optional :events key which needs to be a vector of
+  event keywords under which the function will be registered
   - TODO: add suport for `prepost-map?` (don't forget to add it to arglist)
   - TODO: add validation of macro parameters"
   {:arglists '([name doc-string? attr-map? [params*] body])}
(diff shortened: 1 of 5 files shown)


test/cljs/status_im/test/utils/fx.cljs
```diff
@@ -17,6 +17,18 @@
   [{:keys [db]} b]
   (identity nil))
 
+(fx/defn hello4
+  {:doc "this function is useless as well"
+   :events [:test/valid1 :test/valid2]}
```
dmitryn, line 22: @yenda  If i set multiple events like in this case, does it mean `hello4` would be called when `test/valid1`, `:test/valid2` gets dispatched? What is the use case for having multiple events handled by same defn?
Also, Is it possible to know what event triggered the call inside `hello4` body? Seems like event-name is not passed to the function.

yenda: sure I will add that. 

The reason for supporting multiple events is to encourage the use of names that describe what happens for events and not actions. So the function can be an action and multiple events can trigger it. For instance navigate-back can be triggered by wallet/back-button-pressed and profile/back-button-pressed

dmitryn (approved):

test/cljs/status_im/test/utils/fx.cljs
yenda: So I've looked into it and actually I think for the rare events that need to be self aware it is better to rely on an interceptor

flexsurfer (approved):

test/cljs/status_im/test/utils/fx.cljs
dmitryn: Ok, that makes sense not to add event-name to args as it's rarely used. It can be added to cofx by interceptor when needed as you said.

yenda:
I added a test commit because this is hardcore macro chemistry and I wanted to make sure it worked so it is intended to be removed (or not) but will help with testing

src/status_im/utils/fx.clj
```diff
@@ -26,17 +36,27 @@
...
+         (clojure.core/defn ~(with-meta name m)
+           ([~@argsyms] (fn [cofx#] (~(with-meta name m) cofx# ~@argsyms)))
+           ([cofx# ~@args]
+            (if (and (map? cofx#)
+                     (not (nil? (:db cofx#))))
+              (let [~cofx cofx#]
+                ~@fdecl)
+              (throw (js/Error. (str "fx/defn expects a map of cofx as first argument got " cofx# " in function " ~name))))))
+         ~(when (not-empty events)
+            `(when-not (find-ns 'status-im.utils.handlers)
```
rasom: it seems that we don't want to use `find-ns` anymore

rasom (changes requested):

src/status_im/utils/fx.clj
yenda: I don't think it can build without it before your PR is merged

rasom: do you mean it will work only with `shadow-cljs`?

yenda: I mean we will have warnings if I remove it now

churik:
@rasom changes are still requested?

churik:
Tested:
- creating account
- logout
- force logout actions (changing networks / bootnodes / log level)

src/status_im/utils/fx.clj
rasom: Warning during development or in release builds?

rasom: I don't see any warnings in dev mode

rasom: I mean, without `find-ns`

yenda: that is because `[status-im.utils.handlers :as handlers]` is already required in the only namespace (`status-im.accounts.login.core`) where the part that needs it is macro-expended.

yenda: @rasom I removed it can you unblock the PR?

rasom (approved):

yenda:
@churik the windows build failing isn't related, there is no more space on the node that built it @jakubgs

yenda:
@churik is it ok to merge?

churik:
@yenda yes, go ahead

PR #8278, opened 2019-05-24
Prepare to use status-go based protocol: 1-1 and public chat management APIs.

mandrigin:
Connect to stubs of status-go protocol API, behind the flag. Since status-go isn't updated yet, setting this flag will break the app.

The idea is to merge the stub APIs support first, then implement them on status-go side and then make this mode a default.

More PRs will follow:
- group chats;
- message management (sending/retrieving/paging/etc)

## Note to QA

Joining and leaving chats should work as expected. The new functionality is disabled and should NOT be tested.

---

status: ready


cammellos:
@mandrigin not sure I agree with the approach, I appreciate the intention and it's useful to spec out the endpoints we want to use, but I would not actually code them in until we are ready to implement them, otherwise is just extra code that we have to maintain and adds little value.

Instead I would just add some comments, if we already have some plans.

Also the move is likely to be gradual, will take some time and certain parts might well still be in realm (UI data for example), so it's likely not a either or scenario, I think comments or a doc make more sense in this case.

mandrigin:
@cammellos we already have status-console-client that implements most of the functionality, we just need to bring them together in a way, compatible with the current implementation.

I have a branch of `status-go` that actually works with these endpoints, but it uses stub data.

cammellos:
@mandrigin not sure, I still see this commit as unnecessary, why now and not when we are ready to integrate?

mandrigin:
@cammellos I decided to start from status-react before implementing something real in status-go mostly to get first feedback on where to plug the protocol, so right now it is very easy to change stuff (I have spent almost no time making these stub methods), in case our API should look different to be easily integrated with status-react.

Basically, I want to avoid the scenario when we agreed on the API in a doc, but when we started implementing it, it turns out that our current status-react code isn't suited well for it.

Again the timeline as I see it now is:
- I'm doing these stub API and connect them to status-react until mid-next week;
- Then we implement this part of the API and update status-go;
- Then we expand these APIs, etc, repeat.

All this work will happen behind this flag until we have a fully functional chats of certain types there (I guess public + 1-1).

Then we migrate group chats in a simiar manner.

That way all the PRs are relatively small.

yenda:
I'm not opposed to the concept but we already have a json-rpc api namespace, with the advantage of listing the endpoints so that it is not tight to status-go

mandrigin:
@yenda I didn't know that, I'll move the methods declarations there then. This is the kind of feedback I'm looking for before moving forward with expanding and implementing everything.

yenda:
@mandrigin if you are interested I explained the reasoning behind it here https://discuss.status.im/t/wallet-improvement-log/1221/10?u=yenda

mandrigin:
hmm, even more things work than I expected :-D

src/status_im/events.cljs
```diff
@@ -60,6 +60,7 @@
             [status-im.wallet.core :as wallet]
             [status-im.wallet.db :as wallet.db]
             [status-im.web3.core :as web3]
+            [status-im.chat.status-go.core :as chats-go]
```
yenda: there should not be a status-go namespace

yenda: for rpc use json-rpc directly
for direct calls we already have `native-module`

src/status_im/chat/status_go/core.cljs
```diff
@@ -0,0 +1,54 @@
+(ns status-im.chat.status-go.core
+  (:require [re-frame.core :as re-frame]
+            [taoensso.timbre :as log]
+            [status-im.ethereum.json-rpc :as json-rpc]
+            [status-im.utils.handlers :as handlers]
+            [status-im.utils.types :as types]))
+
+(defn- invoke-api
```
yenda: this method is unnecessary just use json-rpc/call directly

mandrigin: agree

src/status_im/events.cljs
mandrigin: yean, naming is bad, but I wanted to collect all the effects related to protocol on status-go side to one namespace. idk how to name it better, `chat-rpc`?

yenda: I think json-rpc is still the place to go you are going to have something similar to `invoke-api` above which won't bring anything and will call json-rpc anyway, after reworking all the rpc calls in the app I've came to the conclusion that it is better to call json-rpc directly it is barely more verbose (most params in the map are optional anyway) and makes need for optimization and new api-calls more visible for instance

https://github.com/status-im/status-react/blob/develop/src/status_im/ethereum/transactions/core.cljs#L111-L139

src/status_im/events.cljs
```diff
@@ -116,6 +117,16 @@
              {:db (assoc db :chats/loading? false)}
              (chat.loading/initialize-chats {:from 10}))))
 
+(handlers/register-handler-fx
```
yenda: these are not needed anymore btw

src/status_im/chat/models/loading.cljs
```diff
@@ -60,10 +61,20 @@
          message-id)))
    statuses))
 
-(fx/defn initialize-chats
-  "Initialize persisted chats on startup"
+(defn update-chats-in-app-db [{:keys [db] :as cofx} chats]
```
yenda: the fx/defn macro now accepts this syntax
```clojure
(fx/defn update-chats-in-app-db 
  {:events [:chats-list/load-success]}
  [{:keys [db] :as cofx} chats] 
  ...)
```

src/status_im/events.cljs
mandrigin: great, thanks for an example, I'll take a looks and change stuff accordingly

mandrigin:
@yenda I think it is ready to be reviewed again

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -72,3 +76,8 @@
            on-success)
          :on-error
          on-error}))
+
+;; effects
+(re-frame/reg-fx
```
yenda, line 81: if you go that path the effect needs to be plural and added to the list of mergeable effects, otherwise it will probably happen that you get this effect twice in the same event

src/status_im/chat/models/loading.cljs
```diff
@@ -4,7 +4,10 @@
             [status-im.chat.models :as chat-model]
             [status-im.utils.datetime :as time]
             [status-im.utils.fx :as fx]
+            [status-im.utils.config :as config]
```
yenda: please order requires alphabetically

src/status_im/ethereum/json_rpc.cljs
mandrigin: done

src/status_im/chat/models/loading.cljs
```diff
@@ -2,9 +2,12 @@
   (:require [status-im.accounts.db :as accounts.db]
             [status-im.chat.commands.core :as commands]
             [status-im.chat.models :as chat-model]
+            [status-im.utils.config :as config]
             [status-im.utils.datetime :as time]
             [status-im.utils.fx :as fx]
             [status-im.utils.priority-map :refer [empty-message-map]]
+            [taoensso.timbre :as log]
+            [re-frame.core :as re-frame]
```
yenda: sort by alphabetical order (M-x cljr-clean-ns in emacs)

src/status_im/signals/core.cljs
```diff
@@ -8,6 +8,7 @@
             [status-im.node.core :as node]
             [status-im.pairing.core :as pairing]
             [status-im.transport.message.core :as transport.message]
+            [status-im.chat.models.loading :as chat.loading]
```
yenda: sort by alphabetical order (M-x cljr-clean-ns in emacs)

src/status_im/chat/models/loading.cljs
```diff
@@ -60,10 +63,38 @@
...
+  [{:keys [db] :as cofx} chats]
+  (fx/merge cofx
+            {:db (assoc db :chats chats)}
+            (commands/load-commands commands/register)))
+
+(defn- unkeywordize-chat-names
+  [chats-response]
+  (assoc chats-response
+         :chats
+         (into {}
```
yenda: sorry I didn't catch that earlier reduce-kv will be more efficient here:
`(reduce-kv (fn [acc chat-name chat] (assoc acc chat-name chat)) {} (:chats chats-response))`

mandrigin: I also found a bug there, so thanks :) fixed.

mandrigin:
@rasom @flexsurfer @cammellos I want more reviews! 😆

yenda (approved):

src/status_im/chat/models.cljs
```diff
@@ -101,34 +103,30 @@
...
-  [cofx chat-id chat-name admin participants]
-  (upsert-chat cofx
-               {:chat-id     chat-id
-                :name        chat-name
-                :is-active   true
-                :group-chat  true
-                :group-admin admin
-                :contacts    participants}))
+  (if config/use-status-go-protocol?
+    {:json-rpc/call [{:method "status_joinPublicChat"
```
flexsurfer, line 113: why json-rpc ?

mandrigin: what do you mean by that? why do we export API as json rpc or why do I use this effect name?

flexsurfer: why do we export API as json rpc . sorry

mandrigin: (1) it is more modular on status-go side
(2) it is more universal if someone wants to create a 3rd party chat UI, they don't need to rely on native code, they can just export this JSON api to `localhost` and make requests.
(3) also easier to debug separately, you only need status-go and `curl` for that.

flexsurfer: ok i think i know the answer :)

rasom (approved):

src/status_im/data_store/realm/schemas/account/chat.cljs
```diff
@@ -251,3 +251,9 @@
   (update v13 :properties assoc
           :group-chat-local-version {:type :int
                                      :optional true}))
+(def v15
```
flexsurfer: could you elaborate on this?

flexsurfer: i don't see it used

mandrigin: the only places these fields are mentioned anywhere in the source code is tests and realm schema. they are always `nil` as far as I'm concerned.

yenda: I think what @flexsurfer means is that if you want these changes to take effect you also need to use the schema in the core namespace

yenda: but  better remove these changes and make a seperate PR for it, because db changes are irreversible and this PR doesn't need them

mandrigin: ah, yeah, forgot to attach them. sure, I will create a separate PR for them, makes sense.

mandrigin: @yenda @flexsurfer https://github.com/status-im/status-react/pull/8315

Serhy:
On latest Android, iOS and MacOS builds I tested no issues with joining and leaving chats. Works same to nightly build.
Upgrade from nightly on to this PR builds with deleted 1-1/group/public chats also no regressions (and can proceed to message in chats after the upgrade fine).

Issue #8286, opened 2019-05-25
Fix find-ns-obj not supported for target react-native error with advanced compilation


yenda:
# Problem

As described in the following issue:
https://github.com/thheller/shadow-cljs/issues/500

We currently need to use a fork a cljs to use advaned compilation with shadow-cljs because of a `find-ns-obj not supported for target react-native error`

## Implementation

Find what is the cause of the error in the code base and fix it

## Acceptance Criteria

The solution should be based on this PR and not use a fork of cljs

https://github.com/status-im/status-react/pull/8285

yenda:
@StatusSceptre can we quickly bounty that to get it solved?

rasom:
https://github.com/status-im/status-react/blob/448e84a0e99a5a14583c82c278a7c5e0ae84b673/src/status_im/utils/views.clj#L55 this line

yenda:
@rasom I remember that it was added because otherwise the macro would complain about missing require in the namespace that don't require it already, is it going to become a problem again if we simply remove it?

StatusSceptre:
@yenda I am a little confused here. Do you want me to create a bounty on this, even though it has already been solved and merged (with multiple authors)?

yenda:
@StatusSceptre it doesn't need to go on gitcoin since it is already done but we should still reward @hlolli

StatusSceptre:
I'll need a receiving address for that then, could not find a gitcoin profile for @hlolli...

PR #8287, opened 2019-05-25
Shadow cljs fix find-ns-obj

hlolli:
### Summary

Admit that I didn't test this (nixos stuff is missing on this branch) but this seems a bit obvious looking at it here https://github.com/clojure/clojurescript/blob/master/src/main/cljs/cljs/core.cljs#L11649

And @thheller was right, that in the case of an advanced compilation, this makes no sense.

cc #8286 @yenda


yenda:
@hlolli thanks, let see if it works. However the PR can't be merged as is, how could we keep this code for dev where it is needed?

rasom:
Hey @hlolli!

It definitely fixes the issue and works fine with both `:none` and `:advance` optimization levels. 

Thanks

rasom (approved):

yenda:
@rasom I think you already merged that one so it can be closed right? we just need to see with @StatusSceptre how we pay up a bounty for it to @hlolli

rasom:
@yenda no I didn't and it is not needed anymore as I decided to not use `shadow-cljs`.

rasom:
can be closed

PR #8289, opened 2019-05-27
Fix custom tokens

flexsurfer:
Fixed crash when trying to add custom token


jeluard (approved):

yenda (approved):

dmitryn (approved):

annadanchenko:
@flexsurfer can you describe cases when crash happens, please? So it's more clear if fix should go to the release

flexsurfer:
@annadanchenko i believe it shouldn't, it's only in develop, and there is only one case, when you add custom token app crashes

PR #8315, opened 2019-05-29
Remove unused chat fields

mandrigin:
Some fields in the `chats` dictionary in both app-db and realm are unused. This commit removed these fields, to simplify implementation of go protocol.

## Note to QA

Migrations from the current release version should work, the list of chats should be migrated successfully.

---

status: ready

### DIFF
--- a/src/status_im/data_store/realm/schemas/account/chat.cljs
+++ b/src/status_im/data_store/realm/schemas/account/chat.cljs
@@ -251,3 +251,9 @@
   (update v13 :properties assoc
           :group-chat-local-version {:type :int
                                      :optional true}))
+(def v15
+  (update v14 :properties dissoc
+          :message-overhead
+          :removed-from-at
+          :added-to-at
+          :removed-at))
--- a/src/status_im/data_store/chats.cljs
+++ b/src/status_im/data_store/chats.cljs
@@ -106,8 +106,7 @@
   (fn [realm]
     (let [chat (get-chat-by-id chat-id realm)]
       (doto chat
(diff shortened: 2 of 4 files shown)


yenda (approved):

mandrigin:
@flexsurfer can you take a look too?

src/status_im/data_store/realm/schemas/account/chat.cljs
```diff
@@ -251,3 +251,9 @@
   (update v13 :properties assoc
           :group-chat-local-version {:type :int
                                      :optional true}))
+(def v15
+  (-> v14
+      (update :properties dissoc :message-overhead)
```
cammellos: `dissoc` is variadic, so you can have just

`(update v14 :properties dissoc :message-overhead :removed-from-at .....)`

cammellos (approved):

src/status_im/data_store/realm/schemas/account/chat.cljs
mandrigin: oh, thanks! will do

flexsurfer (approved):

src/status_im/data_store/realm/schemas/account/chat.cljs
mandrigin: done!

Serhy:
Chats (existing 1-1/group/public, deleted, with cleared history) migrations tested on Android, iOS, MacOS builds. No bugs found.

PR #8319, opened 2019-05-30
[feature] allow internal transaction to specify asset

yenda:
- allows transaction flow started by an internal feature to show
the right asset in the overview and in transaction messages
(required for tribute to talk)

status: ready


vitvly (approved):

yenda:
@flexsurfer do you agree with this change for ttt? I guess it's going to be overwritten with yours later on anyway

flexsurfer (approved):

PR #8320, opened 2019-05-30
[feature] support for new CIDv1

### LINKED ISSUE #8104: bzz://theswarm.eth doesn't work
# Problem
bzz://theswarm.eth doesn't resolve

# Expected behavior
It resolves, probably using Swarm gateway

OR link to issue on "We don't yet have Swarm support, we'd love to, here's the <GHI> you can follow along"

## Notes
Customer feedback https://www.reddit.com/r/ethereum/comments/bjhptq/status_release_0120_introducing_discover_v1_and/

cc @mandrigin @flexsurfer @yenda @annadanchenko fyi

yenda:
fixes #8104 

### Steps to test

Tested on emulator, open theswarm.eth page should open

status: ready

### DIFF
--- a/src/status_im/ethereum/ens.cljs
+++ b/src/status_im/ethereum/ens.cljs
@@ -105,8 +105,12 @@
     :method "contenthash(bytes32)"
     :params [(namehash ens-name)]
     :on-success
-    (fn [hash]
-      (cb hash))}))
+    (fn [raw-hash]
+      ;; NOTE: it would be better if our abi-spec/decode was able to do that
+      (let [hash (subs raw-hash 130)
+            [cid & hash-and-zeros] (string/split hash "1b20")
+            hash (str "0x" cid "1b20" (subs (apply str hash-and-zeros) 0 64))]
+        (cb hash)))}))
 
 (defn content
   [resolver ens-name cb]
--- a/src/status_im/utils/contenthash.cljs
+++ b/src/status_im/utils/contenthash.cljs
@@ -23,17 +23,36 @@
                             hash))]
       (str "0xe301" (hex/encode (b58/decode b58-hash))))))
 
-(defn decode [hex]
-  (when (and hex (not= hex "0x")
-             (string/starts-with? hex "0xe30101")
-             ;; TODO properly decode the CID
-             ;; extract the content-type using varint ns
-             (= 78 (count hex)))
-    {:namespace :ipfs
-     :hash  (-> hex
-                (subs 10)
-                hex/decode
-                b58/encode)}))
+(defn decode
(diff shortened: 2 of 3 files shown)


flexsurfer (approved):

vitvly (approved):

src/status_im/utils/contenthash.cljs
```diff
@@ -23,17 +23,34 @@
...
-     :hash  (-> hex
-                (subs 10)
-                hex/decode
-                b58/encode)}))
+    (cond
+      (string/starts-with? hex "0xe40101")
+      ;; assuming the content type is 2 bytes
+      {:namespace :swarm
+       :hash (-> hex
+                 (subs 10)
```
flexsurfer: are you sure here is 10 ?

flexsurfer:
if you could add a few tests would be great

flexsurfer (changes requested):
swarm

src/status_im/utils/contenthash.cljs
flexsurfer: there is additional `01` for content type

yenda: yes 2 bytes for content type

flexsurfer: so it should be 12 `0xe40101fa01` ?

yenda:
@flexsurfer fixed swarm and added test

yenda:
@flexsurfer actually I decided to split on the hash bytes so that we don't depend on the content-type length

src/status_im/utils/contenthash.cljs
yenda: no fa01 is just one kind of contenthash

flexsurfer (approved):

yenda:
@flexsurfer theswarm.eth should open now

src/status_im/ethereum/ens.cljs
```diff
@@ -105,8 +105,12 @@
     :method "contenthash(bytes32)"
     :params [(namehash ens-name)]
     :on-success
-    (fn [hash]
-      (cb hash))}))
+    (fn [raw-hash]
+      ;; NOTE: it would be better if our abi-spec/decode was able to do that
+      (let [hash (subs raw-hash 130)
```
flexsurfer, line 110: does it work the same for ipfs and swarm?

yenda: yes it's not decoding the contenthash itself at this point, just extracting it from the bytes soup returned by the contract

flexsurfer (approved):

PR #8323, opened 2019-05-30
E2E for sync public, group chats and profile details: username and photo

churik:
status: ready


asemiankevich (approved):

Serhy (approved):
All good! Apart from the scenario itself which is really long. But I don't know how could it be done in a different way :)

PR #8326, opened 2019-05-30
don't exclude arm64-v8a libraries

jakubgs:
For some reason we were already building for `arm64-v8a`, but we were excluding the built libraries from the resulting APK bundle.

Fixes: #8292

Issue #8292, opened 2019-05-28
 Google Play 64-bit requirement to solve before August 1, 2019


annadanchenko:
Warning from PlayStore:

This release is not compliant with the Google Play 64-bit requirement

The following APKs or App Bundles are available to 64-bit devices, but they only have 32-bit native code: 2019052417.

From August 1, 2019 all releases must be compliant with the Google Play 64-bit requirement.

Include 64-bit and 32-bit native code in your app. Use the Android App Bundle publishing format to automatically ensure that each device architecture receives only the native code it needs. This avoids increasing the overall size of your app. [Learn More](https://developer.android.com/distribute/best-practices/develop/64-bit)

Extra info in blog post: https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html
<img width="1580" alt="Screen Shot 2019-05-28 at 14 37 31" src="https://user-images.githubusercontent.com/6584810/58478366-305bbd80-8156-11e9-8501-a3738c8bebf6.png">

jakubgs:
I'm researching the 64 bit support for Android builds and here is what I found:
#### What we include
```
 $ zipinfo -1 StatusIm-190524-171002-7f38c5-release.apk | grep \.so$
lib/armeabi-v7a/libc++_shared.so
lib/armeabi-v7a/libconceal.so
...(omitted)...
lib/x86/libc++_shared.so
lib/x86/libconceal.so
...(omitted)...
```
So we clearly do support `armeabi-v7a`, but we lack `arm64-v8a` libraries.
According to our `build.gradle` we include it:
https://github.com/status-im/status-react/blob/131b423a0acb92618192be17a92a9e51feda592e/android/app/build.gradle#L158
but then we exclude the `arm64-v8a`(as well as some others):
https://github.com/status-im/status-react/blob/131b423a0acb92618192be17a92a9e51feda592e/android/app/build.gradle#L177
Not sure why we build for it and then exclude it(as we do with `x86_64`), but it seems like just removing that `exclude` should fix things.

jakubgs:
According to this:
https://developer.android.com/distribute/best-practices/develop/64-bit#build_your_app_with_64-bit_libraries
>Most Android Studio projects use Gradle as the underlying build system, so this section applies to both cases. Enabling builds for your native code is as simple as adding the arm64-v8a and/or x86_64, depending on the architecture(s) you wish to support, to the ndk.abiFilters setting in your app's 'build.gradle' file.

[...]

PR #8331, opened 2019-05-31
DIVIDE APK SIZE BY 2 IN 2 LINES YOU WON'T BELIEVE THAT ONE TRICK

jakubgs:
Not sure why we build for x86 and x86_64 considering almost no one uses x86 and we exclude the libraries for x86_64 anyway.

This also reduces the APK size from 51 MB to 31 MB.

Related to #8292 and #8326.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -149,8 +149,6 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-
-
     defaultConfig {
         applicationId "im.status.ethereum"
         minSdkVersion 23
@@ -159,7 +157,7 @@ android {
         versionCode getVersionCode()
         versionName getVersionName()
         ndk {
-            abiFilters "armeabi-v7a", "arm64-v8a", "x86", "x86-64"
+            abiFilters getEnvOrConfig('NDK_ABI_FILTERS').split(';')
         }
     }
     /**
@@ -179,7 +177,6 @@ android {
         exclude 'META-INF/rxjava.properties'
         exclude '/lib/mips64/**'
         exclude '/lib/arm64-v8a/**'
-        exclude '/lib/x86_64/**'
         /** Fix for: Execution failed for task ':app:transformNativeLibsWithStripDebugSymbolForDebug'.
         *   with recent version of ndk (17.0.4754217)
         */
@@ -204,7 +201,7 @@ android {
             reset()
             enable enableSeparateBuildPerCPUArchitecture
             universalApk false  // If true, also generate a universal APK
-            include "armeabi-v7a", "x86"
+            include "armeabi-v7a", "arm64-v8a", "x86"
         }
     }
     buildTypes {
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ HELP_FUN = \
 		   }
 HOST_OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 
+# Defines which variables will be kept for Nix pure shell, use semicolon as divider
+export NIX_KEEP ?= BUILD_ENV
 export NIX_CONF_DIR = $(PWD)/nix
 
 export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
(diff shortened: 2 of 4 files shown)


yenda (approved):

jakubgs:
I am in awe of your oh so subtle marketing prowess.

pedropombeiro (approved):

churik:
@jakubgs Can you elaborate what should be tested here?

yenda:
@jakubgs lmao seems like statustestbot is an x86 lover

asemiankevich:
@jakubgs something wrong with e2e build. all tests are failed, please take a look. I was able to install .apk on my real device btw

yenda:
@asemiankevich that was my joke, this PR reduces the size of the apk by removing x86 builds, no phones runs x86 builds it's all arm.
Since I am not sure there is a way to emulate arm I think the best solution could be to build x86 for e2e apk only @jakubgs is that feasible?

yenda:
@jakubgs btw this is true for dev environment as well, tried you PR and

```
Skipping device 'status(AVD) - 9' for 'app:debug': Could not find build of variant which supports density 560 and an ABI in x86

```

I think we should have a conditional here for dev and e2e builds

jakubgs:
Ah, I see, so the simulators use `x86`, which also affects `e2e` tests. Makes sense. I'll take a look at this today.

jakubgs:
Ok, then how about this:

* Remove `x86_64` from default builds
* Use only `armeabi-v7a`, `arm64-v8a` for release builds(which includes nightlies)

jakubgs:
Here's a manual build using `nightly` as `BUILD_TYPE`:
https://ci.status.im/job/status-react/job/combined/job/mobile-android/11033/
The APK should include only `armeabi-v7a` and `arm64-v8a`:
[StatusIm-190604-174951-7d9566-nightly.apk](https://ci.status.im/job/status-react/job/combined/job/mobile-android/11033/artifact/StatusIm-190604-174951-7d9566-nightly.apk)

asemiankevich:
> StatusIm-190604-174951-7d9566-nightly.apk

this build is okay for real device as well, but for autotests we need e2e build, with Rosten network by default, is it feasible?

jakubgs:
Actually, I just checked that build with `zipinfo` and it still contians `x86` libraries, which it shouldn't:
```bash
 zipinfo -1 StatusIm-190604-174951-7d9566-nightly.apk | grep \.so$ | awk -F'/' '{print $2}' | uniq
armeabi-v7a
x86
```
It should contain only `armeabi-v7a`. I'll have to fix this. What I want to achieve is that only release and nightly builds will lack `x86`, but PR builds will still have it.

Though now that I think about it that would cause problems for e2e release builds anyway. I need to think about this.

jakubgs:
Okay, instead I've introduced a `NDK_ABI_FILTERS` env variable that is set to:
```
NDK_ABI_FILTERS=armeabi-v7a;arm64-v8a;x86
```
By default in `gradle.properties`, but is overridden during CI builds for everything except `e2e` ones.
I think that's a bit cleaner.

jakubgs:
The manual nightly build for Android seems to prove this approach works:
https://ci.status.im/job/status-react/job/combined/job/mobile-android/11036/
```bash
 $ zipinfo -1 StatusIm-190604-210633-5dce6a-nightly.apk | grep \.so$ | awk -F'/' '{print $2}' | uniq
armeabi-v7a
```

jakubgs:
I like how this reduced `android-e2e` build time from ~18 minutes to ~13 minutes.

asemiankevich:
alright seems it works now, is there anything you are planning to achieve in this branch or we can merge it? @jakubgs

android/app/build.gradle
```diff
@@ -200,7 +197,7 @@ android {
             reset()
             enable enableSeparateBuildPerCPUArchitecture
             universalApk false  // If true, also generate a universal APK
-            include "armeabi-v7a", "x86"
+            include "armeabi-v7a", "arm64-v8a", "x86"
```
yenda, line 204: so you still include x86 here after all?

jakubgs: The configuration section isn't used right now since `enableSeparateBuildPerCPUArchitecture` is set to `false`:
https://github.com/status-im/status-react/blob/c2b55233225f6e4d59c9ca5174b5261cc1137fec/android/app/build.gradle#L97
And that means that this section, which normally would have caused Gradle to build a separate APK for each platform, does nothing.  I'm just adding `arm64-v8a` to match what we have set in `ndk.abiFilters`.

Now, since it seems to be necessary for dev builds, because devs use the emulator, `x86` should be included by default, which is the `ndk.abiFilters` setting, but for pr, release, and nightly builds(except for e2e ones) it will be removed by setting the `NDK_ABI_FILTERS` env variable in `android.gradlke`.
See:
https://github.com/status-im/status-react/pull/8331/files#diff-921ddb8a284f72367049b92eff26b9b6R10

jakubgs:
Let's see if rebasing on latest develop will fix the remaining 2% on E2E tests...

PR #8332, opened 2019-06-01
Move translations to node_modules in release build

rasom:
Resulting index.*.ios is ~10% smaller, loading of js takes ~10% less time.

Only English and device's language (if translation for this language exists) are loaded during startup, other languages are loaded only if necessary.

status: ready


mandrigin (approved):

Makefile
```diff
@@ -65,23 +65,31 @@ release: release-android release-ios ##@build build release for Android and iOS
 release-android: export TARGET_OS ?= android
 release-android: ##@build build release for Android
 	@$(MAKE) prod-build-android && \
+	cp -R translations status-modules/translations && \
+    cp -R status-modules node_modules/status-modules && \
```
pedropombeiro: weird indent here

pedropombeiro (commented):
Without understanding exactly what is going on in this PR, I'd like to know if it's possible to avoid touching node_modules in this case, since the Nix PR I'm working on is based on making node_modules come from the Nix store, and therefore read-only.

Makefile
rasom: will fix

rasom:
>  I'd like to know if it's possible to avoid touching node_modules in this case

as far as I understand it is not possible. Probably we could have some link instead of copying files, but I'm not sure it will change anything. Probably there is a way to specify modules from outside `node_modules`, but I'm not sure how.

prod/status_im/i18n_resources.cljs
```diff
@@ -0,0 +1,49 @@
+(ns status-im.i18n-resources
```
yenda, line 1: Is there a way to minimize the difference between dev and prod code or is it going to be like this for every mobule?

rasom: for translations there is no way, but this PR doesn't introduce cljs modules. For cljs modules I will create a macro which will encapsulate the difference and we will not need to put any extra code to `prod` and `dev`. But again, this PR is a different case.

yenda:
@PombeirP please approve if that is not conflicting with your work

yenda:
@rasom how can qa test this? the builds don't contain the changes to measure perf improvements do they?

rasom:
I'm not sure what you mean by "changes to measure perf improvements". QAs can start app, measure startup time, that's it.

rasom:
https://notes.status.im/c-ugk2-4S0-WxM0HrYm2Dg points 5-8 are sufficient here

asemiankevich:
@rasom for some reason i see some error when switching device language on android, however i can't attach report , it is not working on my huawei. @churik cant reproduce this on her samsung

image

rasom:
@asemiankevich can you share logs?

asemiankevich:
> @asemiankevich can you share logs?

`however i can't attach report , it is not working on my huawei.` Unfortunately not, also dont see anything special in logcat under android studio.

rasom:
ok, except dialog, does the rest of it work?

rasom:
> also dont see anything special in logcat

good to know, but I would be glad to see these logs anyway

asemiankevich:
> ok, except dialog, does the rest of it work?

yes, works just fine when i close the dialog

rasom:
@asemiankevich did you try to reproduce it in nightly?

asemiankevich:
> > also dont see anything special in logcat
> 
> good to know, but I would be glad to see these logs anyway

i will log that separately, it exists in nightly too

rasom:
@asemiankevich ok, thanks

rasom:
@PombeirP it would be great to get you input on this task. Right now I'm not sure that we have any other option than create this module, also it will be used for cljs modules. We could add link to this folder, idk if that would help.

pedropombeiro (approved):
Having addon stuff in node_modules is not ideal, but I can stuff this in a Nix `shellHook` and move it away from the Makefile later on. Ideally, all the `release-*` targets should eventually be handled by Nix and just invoked from the Makefile so we have better control over the sources that get compiled.

rasom:
ok, thanks

ci/common.groovy
```diff
@@ -71,6 +71,8 @@ def prep(type = 'nightly') {
   }
   /* node deps, pods, and status-go download */
   utils.nix.shell("scripts/prepare-for-platform.sh ${prepareTarget}", pure: false)
+  utils.nix.shell("cp -R translations status-modules/translations")
```
pedropombeiro: i would merge these 2 lines together, so that the build is faster (by not waiting to instantiate a Nix shell each time). Even better, no Nix shell since these are only file copies:
```suggestion
  sh("cp -R translations status-modules/translations && cp -R status-modules node_modules/status-modules")
```

rasom: ok, I will change it now

PR #8336, opened 2019-06-02
[perf] upgrade realm and improve schemas/queries

yenda:
* upgrade to realm 2.28 to benefit from perf improvements
* remove user-statuses and replace by seen and outgoing-status fields
to get rid of a lot of bloat queries and computations and in-queries which don't
work in this version of realm
* remove unused seen message, bottom-infos

## Tests
Tested on Android

Tested:

* send and receive messages from public and private chats

To test:

* start from nightlies with some chats that have unread messages and unsent messages
* install this version and check that the migrations worked

MV5BZjkyYThhNWMtYWRiZi00ZjQyLThiYzYtZjkwYmVkYTRlYTIwXkEyXkFqcGdeQXVyNTk1ODg5ODA@ _V1_

status: wip


mandrigin (approved):

jakubgs:
I wonder if this might help with #8326 .

jakubgs (approved):
I'm approving only the `yarn.lock` changed, the Clojure stuff is black magic.

src/status_im/chat/models/message.cljs
```diff
@@ -39,9 +38,9 @@
   [{:keys [content content-type] :as message} chat-id current-chat?]
   (let [emoji? (message-content/emoji-only-content? content)]
     ;; TODO janherich: enable the animations again once we can do them more efficiently
-    (cond-> (assoc message :appearing? true)
-      (not current-chat?)
-      (assoc :appearing? false)
+    (cond-> message
+      current-chat?
+      (assoc :seen true)
```
flexsurfer, line 43: could you elaborate?

flexsurfer (approved):

src/status_im/chat/models/message.cljs
yenda: `appearing` is useless, not needed.
If the message is received in the current chat it is considered seen. It was already the case before just not done here

asemiankevich:
@yenda is it something that was broken here? 
image

yenda:
@asemiankevich this is a bug I've seen a few times already in other PR, I have never been able to reproduce in the emulator so I don't really now where it is coming from

asemiankevich:
> @asemiankevich this is a bug I've seen a few times already in other PR, I have never been able to reproduce in the emulator so I don't really now where it is coming from

@yenda i've logged it separately, no worries! Apart from this current PR looks good to me, upgrade looks okay-ish.

Issue #8263, opened 2019-05-22
Heavily formatted messages slow down message rendering


Serhy:
**_Type_**: Bug

**_Summary_**: When there is a message which requires formatting in chat, it appears on chat view much longer than plain-text messages. More items in a single message require formatting - exponentially more time for message to be parsed and displayed to user. 
When messages are really large and there are a lot of items to process - app freezes during that time (below figures are app-freezes values)

For example, on Samsung Galaxy Note 4 (Android 6.0.1) - slow device:
1) The time it takes for a single message having 100 links and 1000 chars in total (including whitespaces) when navigating from Chats view to public chat with the message -> ~3.5 seconds
2) Same number of characters but when no any formatting is applied to message content takes <0.3s 
3) 5 messages with 200 chars and 20 links each (so 1st case above but split in 5 messages) - takes ~2 seconds for messages to appear
4) 5 messages with 200 chars each (so 2nd case above but split in 5 messages) - takes <0.3 seconds for messages to appear

#### Expected behavior
When user of the `Samsung Galaxy Note 4`-like device navigates to chat with the single message which has any type of formatted text (bold, italic, backquoted, link, tag) repeated 250 times with 2500 chars in total, the time it takes from chat view (actual chat, excluded navigation between 'Chats' -> 'Chat') and the actual message content to appear - is less than 0.3s

#### Actual behavior
When user of the `Samsung Galaxy Note 4`-like device navigates to chat with the single message which has any type of formatted text (bold, italic, backquoted, link, tag) repeated 250 times with 2500 chars in total, the time it takes from chat view (actual chat, excluded navigation between 'Chats' -> 'Chat') and the actual message content to appear is:

| Text | Links | Tags | Bold | Italic | Backquotes | Mix of everything (without text) |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | 
| <0.3s | 26s | 26s | 16s | 16s | 26s | 23s |

VIdeo for above measurements from Galaxy Note 4: https://drive.google.com/file/d/1-0rUsqj5_OjK1HiDqu-AXYLylvyt_ezD/view
Message content used in Video you can get here: https://docs.google.com/spreadsheets/d/1pXvXuwZ0AR-pJ5BPsiZL1a1V5LajLMQAeBd6FsMUN_o/edit?usp=sharing 

### Reproduction

* Open Status and create new account
* Join unique public chat 
* Send message with text required formatting (you can get data from the above link)
[...]

PR #8342, opened 2019-06-03
[perf] small perf improvement for nested-text

yenda:
A simple optim that seems to bring about 30% speedup improvement to compute `nested-text` elements

## Measurements

384 ms with no optim average 40 runs of 10k calls
269 ms with simple element for non nested text on average of 40 runs of 10k calls

```clojure
(def a (atom 0))
(defn nested-text
  ([options & nested-text-elements]
   (let [start (status-im.utils.datetime/timestamp)
         options-with-style (prepare-text-props options)
         elements
         (reduce (fn [acc text-element]
                   (conj acc
                         (cond
                           (string? text-element)
                           [text-class options-with-style text-element]

                           (vector? text-element)
                           (let [[{:keys [style] :as options} nested-text-elements]
                                 text-element]
                             [(if (string? nested-text-elements)
                                text-class
                                nested-text) (prepare-text-props
                                              (-> options-with-style
                                                  (merge (dissoc options style))
                                                  (update :style merge style)))
                              nested-text-elements]))))
                 [text-class options-with-style]
                 nested-text-elements)
         end  (status-im.utils.datetime/timestamp)]
     (swap! a #(+ % (- end start)))
     elements
     nil)))
```

```clojure
(def b (let [{:keys [content] :as message} (get-in @re-frame.db/app-db [:chats "test" :messages "0xd5c4ff6f2f705560211caee516b5e0e655edb6323e3ad5c9aaf3284d4c6266e2"])] (status-im.ui.screens.chat.utils/render-chunks (:render-recipe content) message)))
(dotimes [y 10] (reset! a 0) (dotimes [x 10000] (apply nested-text {} (conj b (str x)))) (println @a))
```

### Steps to test

- check rendering speed of #test public channel which is full of nested-text messages

status: ready

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -128,9 +128,12 @@
 
                        (vector? text-element)
                        (let [[options nested-text-elements] text-element]
-                         [nested-text (prepare-text-props
-                                       (utils.core/deep-merge options-with-style
-                                                              options))
+                         [(if (string? nested-text-elements)
+                            text-class
+                            nested-text)
+                          (prepare-text-props
+                           (utils.core/deep-merge options-with-style
+                                                  options))
                           nested-text-elements]))))
              [text-class options-with-style]
              nested-text-elements))))


flexsurfer (approved):

mandrigin (approved):

Serhy:
No regressions with the text formatting found in 1-1/public/Group chats. 
With Galaxy Note 4, - 5 messages of bold text (message taken from #8263 issue) in public chat after navigation appears:

**Nightly** develop build: **72 second**
**PR build**: **46 seconds** (~35% faster)

PR #8345, opened 2019-06-03
[fix] externs for Preview Privacy Mode

yenda:
@churik fixes the `Preview privacy mode` switch

## Testing

no need to wait for e2e tests to end before testing, this is an advanced compilation issue, the js call made when tapping the switch was turned into garbage because it wasn't referenced in externs.js
this is the only thing affected by this change

status: ready

### DIFF
--- a/externs.js
+++ b/externs.js
@@ -425,6 +425,7 @@ var TopLevel = {
     "setAutoCancel" : function () {},
     "setBackgroundColor" : function () {},
     "setBarStyle" : function () {},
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

mandrigin (approved):

churik:
Issue https://github.com/status-im/status-react/pull/8304#issuecomment-498231223  is fixed on IOS and Android
Thanks @yenda!

PR #8357, opened 2019-06-04
always checkout to status-react folder

jakubgs:
This is a possible solution for Pedros issues with changing checkout folder for `status-react`.

It's really ugly, but I can't use [`customWorkspace`](https://jenkins.io/doc/book/pipeline/syntax/#common-options) because the context doesn't have access to `EXECUTOR_NUMBER` variable([JENKINS-50907](https://issues.jenkins-ci.org/browse/JENKINS-50907)). Because of that we can only set a static path, and that would cause clashes when more than 1 job of the same type is running on the same host.

Context: https://discourse.nixos.org/t/path-for-src-in-derivation-file-is-created-with-captured-directory-name-as-suffix/3102

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -5,6 +5,8 @@ pipeline {
     timestamps()
     /* Prevent Jenkins jobs from running forever */
     timeout(time: 45, unit: 'MINUTES')
+    /* make project dir the same for consisten nix src */
+    checkoutToSubdirectory('status-react')
     /* Limit builds retained */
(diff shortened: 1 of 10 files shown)


pedropombeiro (approved):
Thanks @jakub!

jakubgs:
Looks like this is not necessary after all thanks to: https://github.com/NixOS/nix/issues/1305
We can use this instead:
```
src = builtins.path { name = "stack2nix"; path = ./.; };
```

PR #8334, opened 2019-06-01
[developerXP] remove unused deps

yenda:
- remove warning for asyncstorage deprecation
- remove unused asyncstorage-down library

### Steps to test

I couldn't find any usage for this library in the codebase so this shouldn't have change anything.
Only requires a basic sanity check
Tested on Android emulator, opening chat, wallet, settings and playing around.

status: ready

PR #8359, opened 2019-06-04
Minor fixes

pedropombeiro:

### Summary

This PR serves to offload some work from the Nix gradle/node2nix PR, to make that one as light as possible.

### Review notes

This PR:
- moves files which are no longer different between mobile and desktop to the root, so we have as few symlinks as possible.
- fixes the `GIT_ROOT` variable so that it works even if the script is invoked from within a folder with a different git project (I needed this for a Nix build).
- updates `react-native-dialogs` to 1.0.4, since 1.0.2 contains a duplicate type which caused a build error.

#### Platforms

- Android
- Linux

Should have no impact on the app, unless `react-native-dialogs` 1.0.4 is somehow broken.

status: ready


scripts/gen-deps-hash.sh
```diff
@@ -6,7 +6,7 @@
 # the combined versions of all the specified tools.
 ################################################################################
 
-GIT_ROOT=$(git rev-parse --show-toplevel)
+GIT_ROOT=$(cd "${BASH_SOURCE%/*}" && git rev-parse --show-toplevel)
```
jakubgs, line 9: I get it, but why exactly is this necessary?

jakubgs: Oh:
>if the script is invoked from within a folder with a different git project 

Sorry, didn't read the description =].

yenda:
@PombeirP some of these changes are already in https://github.com/status-im/status-react/pull/8334 could you have a look at it please and documents how things will work now with nix, do we have to change the nix node modules file as well now when changing deps?

pedropombeiro:
> @PombeirP some of these changes are already in #8334 could you have a look at it please and documents how things will work now with nix, do we have to change the nix node modules file as well now when changing deps?

I've added a post in Discuss to shed some light into what is changing: https://discuss.status.im/t/the-road-to-nix-a-functional-package-manager-to-rule-them-all/1049/114?u=pedro

rasom (approved):

pedropombeiro:
@rasom do you know what happened with the builds? It's mentioning status-modules, but as far as I can tell, my branch isn't rebased on latest develop which contains that.

ci/utils.groovy
```diff
@@ -1,11 +1,7 @@
 nix = load 'ci/nix.groovy'
 
 def getVersion(type = null) {
```
jakubgs: Then I guess we should drop the `type` argument and remove it from all the `getVersion()` calls.

nix/desktop/realm-node/generate-nix.sh
```diff
@@ -20,12 +21,21 @@ cat << EOF > $input
 ]
 EOF
 
+# Specify the package.json file containing the build dependencies to install
+cat << EOF > $supplement_input
+[
+  "node-pre-gyp"
+]
+EOF
```
jakubgs, line 29: What's the need for this multi-line syntax?

jakubgs (approved):
Love how much stuff this removes.

nix/desktop/realm-node/generate-nix.sh
pedropombeiro: this file might change often and include more modules, I just don't want to keep changing syntax as we add and remove modules.

ci/utils.groovy
pedropombeiro: You're right, I missed that, fixing.

nix/desktop/realm-node/generate-nix.sh
jakubgs: fair

PR #8360, opened 2019-06-05
Move goog.i18n deps to a module

rasom:
This PR adds the first cljs module to the project and introduces `defmodule` macro which makes definition of module's api and loading of module a bit simpler.

Moving all `goog/i18n` mentions to a separate module makes `index.android.js` 8.91% smaller (2729KB vs 2996KB) and startup ~4% faster. The loading of the module even on a slow device is quite fast (90ms), assuming its size is ~216KB. Probably it is smaller and faster than it was before this PR because I rewrote it a bit so that compilation warning doesn't appear anymore, and thus `good/i18n` calls are now optimized better during compilation.

This warning is fixed:
```
[2019-06-01T06:57:46.639Z] Jun 01, 2019 6:57:43 AM com.google.javascript.jscomp.LoggerErrorManager println
[2019-06-01T06:57:46.639Z] WARNING: /home/jenkins/workspace/status-react_prs_android_PR-8332/target/android-prod/status_im/utils/datetime.js:40: WARNING - incomplete alias created for namespace goog.i18n, possibly due to await/yield transpilation.
[2019-06-01T06:57:46.639Z] See https://github.com/google/closure-compiler/wiki/FAQ#i-got-an-incomplete-alias-created-for-namespace-error--what-do-i-do for more details.
[2019-06-01T06:57:46.639Z] var loc = (function (){var G__23425 = goog.i18n;
[2019-06-01T06:57:46.639Z]                                       ^^^^^^^^^
[2019-06-01T06:57:46.639Z] 
```

Loading of module on Samsung a500d:
```
06-05 08:25:32.863 30156 30193 D ReactNativeJS: DEBUG [status-im.goog.i18n-module:10] - :load-goog.i18n-module
06-05 08:25:32.953 30156 30193 D ReactNativeJS: DEBUG [status-im.goog.i18n-module:17] - :goog.i18n-module-loaded
```

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ endif
 release: release-android release-ios ##@build build release for Android and iOS
 
 release-android: export TARGET_OS ?= android
+release-android: export BUILD_ENV ?= prod
 release-android: ##@build build release for Android
 	@$(MAKE) prod-build-android && \
 	cp -R translations status-modules/translations && \
@@ -70,6 +71,7 @@ release-android: ##@build build release for Android
 	react-native run-android --variant=release
 
 release-ios: export TARGET_OS ?= ios
+release-ios: export BUILD_ENV ?= prod
 release-ios: ##@build build release for iOS release
 	# Open XCode inside the Nix context
 	@$(MAKE) prod-build-ios && \
@@ -100,16 +102,21 @@ prod-build:
 
 prod-build-android: export TARGET_OS ?= android
 prod-build-android:
-	lein prod-build-android
+	BUILD_ENV=prod lein prod-build-android && \
+	node prepare-modules.js
 
 prod-build-ios: export TARGET_OS ?= ios
+prod-build-ios: export BUILD_ENV = prod
 prod-build-ios:
-	lein prod-build-ios
+	BUILD_ENV=prod lein prod-build-ios && \
+	node prepare-modules.js
 
 prod-build-desktop: export TARGET_OS ?= $(HOST_OS)
+prod-build-desktop: export BUILD_ENV = prod
 prod-build-desktop:
 	git clean -qdxf -f ./index.desktop.js desktop/ && \
-	lein prod-build-desktop
+	BUILD_ENV=prod lein prod-build-desktop && \
+	node prepare-modules.js
 
 #--------------
 # REPL
--- a/.gitignore
+++ b/.gitignore
@@ -162,5 +162,6 @@ conan.cmake
 /.ran-setup
 /.nix-gcroots/
 
-#modules
+# modules
 status-modules/translations
+status-modules/cljs
--- a/externs.js
+++ b/externs.js
@@ -559,5 +559,5 @@ var TopLevel = {
     "decimalPlaces": function () {},
     "_android": function () {},
     "isSupported" : function () {},
-    "authenticate" : function () {},
+    "authenticate" : function () {}
 }
--- a/prepare-modules.js
+++ b/prepare-modules.js
@@ -1,0 +1,20 @@
+var fs = require("fs");
+
+var modules = [
+    "i18n"
+];
+
+modules.forEach(
+    function (moduleName) {
+        fs.readFile(`status-modules/cljs/${moduleName}-raw.js`, "utf8", function (err, data) {
+            if (err) throw err;
+            fs.writeFile(`status-modules/cljs/${moduleName}.js`,
+                ("module.exports=`" + data.replace(/[\\$'"]/g, "\\$&") + "`;"),
+                function (err) {
+                    if (err) {
+                        return console.log(err);
+                    }
+                });
+        });
+    });
+
--- a/project.clj
+++ b/project.clj
@@ -29,7 +29,7 @@
   :githooks {:auto-install true
              :pre-commit   ["lein cljfmt check src/status_im/core.cljs $(git diff --diff-filter=d --cached --name-only src test/cljs)"]}
   :cljfmt {:indents {letsubs [[:inner 0]]}}
-  :clean-targets ["target/" "index.ios.js" "index.android.js"]
+  :clean-targets ["target/" "index.ios.js" "index.android.js" "status-modules/cljs"]
   :aliases {"prod-build"         ^{:doc "Recompile code with prod profile."}
             ["do" "clean"
              ["with-profile" "prod" "cljsbuild" "once" "ios"]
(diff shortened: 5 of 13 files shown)


yenda (commented):
Awesome work does it means that we can expect i18n calls to be slightly more performant ? Computing timestamps is actually one of the bottlenecks so that'd be neat!

rasom:
> does it means that we can expect i18n calls to be slightly more performant

I doubt that it will have any effect in runtime.

yenda:
@rasom ok thanks. btw I've read that type hinting in cljs is only useful for booleans, I wonder if we are using enough of them for it to be worth doing

yenda (approved):

ci/Jenkinsfile.android
```diff
@@ -64,7 +64,11 @@ pipeline {
     }
     stage('Build') {
       steps {
-        script { cmn.nix.shell('lein prod-build-android')}
+        script {
+          cmn.nix.shell('cp params_prod.edn params.edn')
+          cmn.nix.shell('lein prod-build-android')
+          cmn.nix.shell('node prepare-modules.js')
```
jakubgs: Wouldn't it make more sense to just do `make prod-build-android` instead of repeating exactly the same thing you have in the Makefile?

Makefile
```diff
@@ -100,16 +100,22 @@ prod-build:
 
 prod-build-android: export TARGET_OS ?= android
 prod-build-android:
-	lein prod-build-android
+	cp params_prod.edn params.edn && \
```
jakubgs: Should this `cp` be here? Why not make it e separate target? Like:
```
prod-params:
    cp  params_prod.edn params.edn 

prod-build-android: prod-params
```
Or something like that.
Also, can you explain what this file changes exactly? Maybe you can add that as a comment to the Makefile target?

rasom: This file changes the parameter which makes macro behave differently in prod and dev env.

ci/Jenkinsfile.android
rasom: will rewrite later today. That's not my focus usually, I'm satisfied if this part works at all :)

jakubgs:
This PR now has conflicts due to #8360 being merged.
When you rebase you'll have to use the `make` targets as I mentioned in the comment earlier.

ci/Jenkinsfile.android
jakubgs: #8360 was merged and does this already by using the Makefile target.

Makefile
jakubgs: Obviously, but differently how?

jakubgs: Also, who use these files? Is there any reason why this can't be managed by env variables?

rasom: currently `defmodule` macro uses this file (introduced in this PR), another macro will need platform version in that file. This is necessary to provide env vars during cljs compile time. It can be done via env vars though might be a bit implicit. I would merge it as it is and then spend time on this later, because I'm kind of have a bit different objective. It works and I'm happy.

rasom:
@jakubgs rebased

Makefile
```diff
@@ -100,16 +100,22 @@ prod-build:
...
+	lein prod-build-android && \
+	node prepare-modules.js
 
 prod-build-ios: export TARGET_OS ?= ios
 prod-build-ios:
-	lein prod-build-ios
+	cp params_prod.edn params.edn && \
+	lein prod-build-ios && \
+	node prepare-modules.js
```
pedropombeiro, line 112: Can we delete `params.edn` at the end?

Makefile
jakubgs: Sure, just don't want to multiply entities in the repo if it's not necessary. And it doesn't have to be implicit, you can clearly define the env variable above the Makefile target and make it very clear so as to what it does, like so:
```Makefile
prod-build-android: export TARGET_OS ?= android
prod-build-android: export BUILD_ENV ?= prod
prod-build-android:
	lein prod-build-android && \
	node prepare-modules.js
```
We also have `.env*` files in the repo root which contian various settings like that, might be a good place for that too.

Makefile
rasom: nope, it should be changed before cljs compilation

pedropombeiro: and can't we delete that file after cljs is compiled?

rasom: I'm slow. I'll try to move it to env vars now because it seems it's a pain in the ass at the moment.

Makefile
rasom: for some reason `BUILD_ENV` is not seen from lein if defined this way, only if I make it `BUILD_ENV=prod lein prod-build-android && \`

rasom:
@jakubgs @PombeirP does it look better now? `params.edn` files are removed

mandrigin (approved):

cammellos (approved):

Makefile
pedropombeiro: Thanks, it does look easier to integrate with Nix now.

pedropombeiro (approved):

Makefile
rasom: ok, great

jakubgs (approved):
Much nicer.

Makefile
jakubgs: Might be because of `--pure`.
https://github.com/status-im/status-react/blob/cdae3423bcc986d9a390de0a5479cfa4c005c121/nix/shell.sh#L34-L38
For now the simplest thing to do would be to just add a `--keep BUILD_ENV` somewhere in there. What do you think @PombeirP ?
I can do a separate PR that will add an env variables to do that.

jakubgs: How about this: https://github.com/status-im/status-react/pull/8398

jakubgs: I've merged #8398 so try rebasing and adding an `export NIX_KEEP ?= 'BUILD_ENV'` somewhere at the top of the `Makefile`, probably next to this:
https://github.com/status-im/status-react/blob/0e7f73fd11e396c5e5fa898178d3872eb676f234/Makefile#L29
And add a comment saying something like:
```Makefile
# Defined which variables will be kept for Nix pure shell, use semicolon as divider
```
And we should be golden.

Serhy:
No regression found. 
- Can change device localisation and Status picks it up as app localisation too.
- Can change currency and transact in 1-1, Wallet with the preset previously fiat and amount corresponding to fiat
- Request message history and date on 'Fetch more messages' correspond to the date which is about to be requested.

PR #8361, opened 2019-06-05
[#8026] [Wallet] Signing transaction

### LINKED ISSUE #8026: [Wallet] Sign transaction module
Implement a new module, it should 

- [x] get tx data map on input
- [x] show title (1)
- [x] show tx details on the bottom sheet (details should be different for sending eth or calling contract (2))
- [x] it's only possible to change one field -  network fee in this module
- [x] sign with password
- [x] show sent popup

this issue doesn't include:
- USD values (3)  https://github.com/status-im/status-react/issues/8027
- sent popup design [GHI]
- tx fee slider [GHI]
- to field details [GHI]
- approveAndCall detection (showing token amount) [GHI]
- prompts for signing phrase, gas and gas price [GHI]

image

flexsurfer:
fixes #8026

image

The new method is introduced for signing transactions `signing/sign`

https://github.com/status-im/status-react/pull/8361/files#diff-598cee30285d1d51df0def4398019d98R234

```clojure
(fx/defn sign [{:keys [db] :as cofx} tx]
  "Signing transaction or message, shows signing sheet
   tx
   {:tx-obj - transaction object to send https://github.com/ethereum/wiki/wiki/JavaScript-API#parameters-25
    :message {:address  :data  :typed? } - message data to sign
    :on-result - re-frame event vector
    :on-error - re-frame event vector}"
  (fx/merge cofx
            {:db (update db :signing/queue conj tx)}
            (check-queue)))
```

Usage

```clojure
(signing/sign cofx {:tx-obj     {:to  to  :value amount-hex}
                    :on-result [:success-event]})))) 
```

FOR QA and Design teams:
this PR doesn't include:
- USD values (3)  https://github.com/status-im/status-react/issues/8027
- emoji  [GHI]
- sign with passcode  [GHI]
- select wallet [GHI]
- sent popup design [GHI]
- fee slider [GHI]
- to field details [GHI]
- approveAndCall detection (showing token amount) [GHI]
- prompts for signing phrase, gas and gas price [GHI]

FOR QA:
we had default gas 105000 for tokens, in current implementation we rely on gas estimation by the network, but sometimes it fails and we'll show error tooltip for gas in that case, also we need to verify that gas estimation works properly

TESTED:
- sending eth or token from chat
- sending eth or token from dapp
- sending eth or token from wallet
- buying stickers 
- deploy contract from dapp
- sign message from dapp
- sign typed message from dapp
- change gas and gas price
- request stt in simpledapp

NOT TESTED:
-sending real eth or snt

CONCERNS:
- gas estimation may work wrong
- gas estimation may fail, in that case error will be shown and user must enter it manually


src/status_im/chat/commands/impl/transactions.cljs
```diff
@@ -304,39 +300,20 @@
...
-                 ;; we can include the utility helper without running into circ-dep problem
-                 :wallet/update-gas-price
-                 {:success-event :wallet/update-gas-price-success
-                  :edit?         false}}
-                (navigation/navigate-to-cofx next-view-id {}))))
+    (let [{:keys [symbol decimals address]} (tokens/asset-for (:wallet/all-tokens db) (keyword (:chain db)) (keyword asset))
+          {:keys [value]} (wallet.db/parse-amount amount decimals)
+          amount' (str "0x" (abi-spec/number-to-hex (money/formatted->internal value symbol decimals)))
+          to      (ethereum/public-key->address (:current-chat-id db))
+          to'     (ethereum/normalized-address (if (= symbol :ETH) to address))
```
yenda: don't use these `'` find a better name or thread the functions

flexsurfer: yeah yeah i know

src/status_im/browser/core.cljs
```diff
@@ -274,19 +275,58 @@
...
+              :result    {:jsonrpc "2.0"
+                          :id      (int id)
+                          :result  result}}
+    :webview webview-bridge}})
+
+(defn normalize-sign-message-params
+  "NOTE (andrey) we need this function, because params may be mixed up,
+  so we need to figure out which one is address and which message"
+  [params]
+  (let [first_param           (first params)
```
dmitryn: sorry for neatpicking, but why `_` in names, but not `-`?

src/status_im/browser/core.cljs
```diff
@@ -274,19 +275,58 @@
...
+        second_param          (second params)
```
dmitryn: you could use destructuring here:
`[first-param second-param] params`

src/status_im/signing/core.cljs
```diff
@@ -0,0 +1,251 @@
...
+   :interceptors [(re-frame/inject-cofx :random-id-generator)]}
+  [cofx response tx-obj]
+  (let [cofx-in-progress-false (assoc-in cofx [:db :signing/sign :in-progress?] false)
+        {:keys [result error]} (types/json->clj response)]
+    (if error
+      (transaction-error cofx-in-progress-false error)
+      (transaction-result cofx-in-progress-false result tx-obj))))
+
+(fx/defn discard
+  "Discrad transactrion signing"
```
dmitryn: typos

src/status_im/subs.cljs
```diff
@@ -1868,3 +1831,68 @@
...
+ :signing/amount-errors
+ :<- [:signing/tx]
+ :<- [:balance]
+ (fn [[{:keys [amount token gas gasPrice]} balance]]
+   (if (and amount token)
+     (let [amount-bn (money/formatted->internal (money/bignumber amount) (:symbol token) (:decimals token))
+           amount-error (or (get-amount-error amount (:decimals token))
+                            (get-sufficient-funds-error balance (:symbol token) amount-bn))]
+       (if amount-error
+         amount-error
```
dmitryn: could be shorter with `or`

`(or amount-error (get-sufficient-gas-error ...))`

src/status_im/hardwallet/core.cljs
```diff
@@ -62,7 +62,8 @@
     (if navigate-to-browser?
       (fx/merge cofx
                 {:db (assoc-in db [:hardwallet :on-card-connected] nil)}
-                (wallet/discard-transaction)
+                ;;TODO use new signing flow
+                ;;(wallet/discard-transaction)
```
dmitryn, line 66: Could you change it to `signing/discard`  as it seems like doing what `wallet/discard-transaction` did before

dmitryn (approved):

src/status_im/hardwallet/core.cljs
dmitryn: Or maybe put it into comments for later

src/status_im/browser/core.cljs
flexsurfer: this is something old, just moved here, but agree it's weird

src/status_im/browser/core.cljs
yenda: can you make this change since git doesn't understand that it is just moved anyway?

src/status_im/chat/commands/impl/transactions.cljs
```diff
@@ -304,39 +300,21 @@
...
-                 {:success-event :wallet/update-gas-price-success
-                  :edit?         false}}
-                (navigation/navigate-to-cofx next-view-id {}))))
+    (let [{:keys [symbol decimals address]} (tokens/asset-for (:wallet/all-tokens db) (keyword (:chain db)) (keyword asset))
+          {:keys [value]} (wallet.db/parse-amount amount decimals)
+          current-chat-id (:current-chat-id db)
+          amount-hex      (str "0x" (abi-spec/number-to-hex (money/formatted->internal value symbol decimals)))
+          to              (ethereum/public-key->address current-chat-id)
+          to-norm         (ethereum/normalized-address (if (= symbol :ETH) to address))
+          tx-obj          (if (= symbol :ETH)
```
yenda, line 309: I am not a big fan of this tx-obj everywhere, it isn't really defined anywhere what it is about compared to just transaction

src/status_im/extensions/capacities/ethereum.cljs
```diff
@@ -42,35 +37,11 @@
...
-             :on-error         [:extensions/wallet-ui-on-failure on-failure]}
-      nonce
-      (assoc :nonce nonce))))
-
 (defn- execute-send-transaction [db {:keys [method params on-success on-failure] :as arguments}]
-  (let [tx-object (assoc (select-keys arguments [:to :gas :gas-price :value :nonce])
-                         :data (when (and method params) (abi-spec/encode method params)))
-        transaction (prepare-extension-transaction tx-object (:contacts/contacts db) on-success on-failure)]
-    (wallet/open-modal-wallet-for-transaction db transaction tx-object)))
+  (signing/sign {:db db} {:tx-obj    (assoc (select-keys arguments [:to :gas :gas-price :value :nonce])
```
yenda, line 41: why not just transaction? so far we have been pretty strict about not using abbreviations in the code base and I don't see how it is an object

src/status_im/signing/core.cljs
```diff
@@ -0,0 +1,250 @@
+(ns status-im.signing.core
+  (:require [status-im.utils.fx :as fx]
+            [status-im.i18n :as i18n]
+            [status-im.utils.money :as money]
+            [status-im.utils.hex :as utils.hex]
```
yenda, line 5: can you run import clean (remove unused/order alphabetically)

src/status_im/signing/core.cljs
```diff
@@ -0,0 +1,250 @@
...
+      (let [[address value] (abi-spec/decode (str "0x" (subs data 10)) ["address" "uint256"])]
+        (when (and address value)
+          {:to       address
+           :contact  (get-contact db address)
+           :contract to
+           :amount   (money/to-fixed (money/token->unit value decimals))
+           :token    token
+           :symbol   symbol})))))
+
+(defn parse-tx-obj [db {:keys [to value data]}]
```
yenda, line 140: I would prefer we really don't import this from the whole wallet this distinction between transaction and transaction-object was uncalled for and the use of abbreviations stands out with the rest of the codebase

src/status_im/chat/commands/impl/transactions.cljs
yenda: in the end this tx-obj is just (select-keys transaction [:to :from :data])

src/status_im/extensions/capacities/ethereum.cljs
```diff
@@ -4,22 +4,17 @@
...
-            [status-im.i18n :as i18n]
             [status-im.native-module.core :as status]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.fx :as fx]
             [status-im.utils.handlers :as handlers]
-            [status-im.utils.hex :as hex]
-            [status-im.utils.money :as money]
             [status-im.utils.types :as types]
-            [status-im.wallet.core :as wallet]))
+            [status-im.signing.core :as signing]))
```
yenda, line 12: import order

src/status_im/extensions/capacities/ethereum.cljs
flexsurfer: it's object from here https://github.com/ethereum/wiki/wiki/JavaScript-API#parameters-25, Object - The transaction object to send:
but i'm open for suggestions on better naming

src/status_im/subs.cljs
yenda: based on @rasom comments during latest dev call, don't you think these are good candidate for local component state?   while the functions themselves should stay separated from the UI and tested, does the whole app need to know about this state?

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -38,15 +39,26 @@
...
+            [react/view
+             [icons/icon :main-icons/next {:color colors/gray-transparent-40}]]
             (= :check accessory)
-            [icons/icon :main-icons/check {:color colors/gray}]
-            :else accessory)
-          {:key accessory}))]]))
+            [react/view
+             [icons/icon :main-icons/check {:color colors/gray}]]
+            :else
+            [react/view {:padding-right 8 :flex-shrink 1}
```
yenda, line 53: I disagree with this change, it is adding an unecessary wrapper around all the components below, and this has a performance impact. The components below can support this styling themselves

src/status_im/signing/core.cljs
flexsurfer: i agree here, but tbh haven't found a time to figure out how to do it better, so i decided to go with tx for signing screen, and tx-obj, raw transaction object to sign, yes we need to split it somehow and use better naming i agree

src/status_im/ui/screens/signing/sheets.cljs
```diff
@@ -0,0 +1,51 @@
...
+  (:require [status-im.ui.components.react :as react]
+            [re-frame.core :as re-frame]
+            [status-im.ui.components.text-input.view :as text-input]
+            [status-im.i18n :as i18n]
+            [status-im.ui.components.button.view :as button]
+            [status-im.ui.components.colors :as colors]))
+
+(views/defview fee-bottom-sheet [fee-display-symbol]
+  (views/letsubs [{gas-edit :gas gas-price-edit :gasPrice max-fee :max-fee} [:signing/edit-fee]]
+    [react/view
```
yenda, line 12: that level of nesting is really overkill, here it is no big deal because it's not a component in a list though

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -0,0 +1,197 @@
...
+            [status-im.ui.components.text-input.view :as text-input]
+            [status-im.i18n :as i18n]
+            [status-im.utils.security :as security]
+            [status-im.ui.screens.signing.sheets :as sheets]
+            [status-im.ethereum.tokens :as tokens]
+            [clojure.string :as string]
+            [status-im.ui.screens.signing.styles :as styles]))
+
+(defn hide-panel-anim
+  [bottom-anim-value alpha-value window-height]
```
yenda, line 23: could you use a translateY value here instead of height in order to use nativeDrivers?

src/status_im/ui/components/list_item/views.cljs
flexsurfer: this is something new from performance investigation, i can fix this here for sure, but we need general performance guidelines

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -0,0 +1,197 @@
...
+    [(anim/spring bottom-anim-value {:toValue window-height})
+     (anim/timing alpha-value {:toValue  0
+                               :duration 500})])))
+
+(defn show-panel-anim
+  [bottom-anim-value alpha-value]
+  (anim/start
+   (anim/parallel
+    [(anim/spring bottom-anim-value {:toValue -40})
+     (anim/timing alpha-value {:toValue  0.4
```
yenda, line 35: usenativedriver

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -0,0 +1,197 @@
...
+            :accessories [[acc-text fee fee-display-symbol] :chevron]
+            :on-press    #(re-frame/dispatch [:signing.ui/open-fee-sheet
+                                              {:content        (fn [] [sheets/fee-bottom-sheet fee-display-symbol])
+                                               :content-height 280}])}]
+          [react/view {:align-items :center :margin-top 16 :margin-bottom 40}
+           [button/primary-button {:on-press  #(re-frame/dispatch [:set :signing/sign {:type :password}])
+                                   :disabled? (or amount-error gas-error)}
+            (i18n/label :t/sign-with-password)]]])])))
+
+(views/defview signing-view [tx window-height]
```
yenda: use a regular defn here, you don't make any subscription and it will make it more clear what happens

src/status_im/signing/core.cljs
```diff
@@ -0,0 +1,250 @@
...
+                    :type      :pending
+                    :hash      hash
+                    :data      data
+                    :token     token
+                    :symbol    symbol
+                    :value     (money/to-fixed (money/bignumber value))
+                    :gas-price (money/to-fixed (money/bignumber gasPrice))
+                    :gas-limit (money/to-fixed (money/bignumber gas))})}))
+
+(defn get-transfer-token [db to data]
```
yenda, line 124: this should be somewhere in the ethereum namespaces

src/status_im/signing/core.cljs
```diff
@@ -0,0 +1,250 @@
...
+  (let [{:keys [symbol decimals] :as token} (tokens/address->token (:wallet/all-tokens db) (ethereum/chain-keyword db) to)]
+    (when (and token data (string? data) (string/starts-with? data (abi-spec/signature->method-id "transfer(address,uint256)")))
```
yenda: there is already a constant for that

yenda (commented):
I really think tx should be called transaction and tx-obj only extracted from transaction when it is actually needed (which afaik is only when calling geth)

yenda:
Could you cut the old part of your schema and put the schema of how the new wallet works in /docs?

src/status_im/ui/components/list_item/views.cljs
yenda: in this case it's simple that can kind of over wrapping is an extra thing to render on native-side, and since it is in the list it's an extra wrapping for each element. And it is only used for some styling it is unecessary because all elements support that kind of styling

flexsurfer:
thanks @yenda really useful review!

src/status_im/signing/core.cljs
yenda: `event-transfer-hash`

src/status_im/ui/screens/signing/views.cljs
yenda: (don't forget to deref your atoms)

src/status_im/ui/screens/signing/views.cljs
yenda: (you can't do height because native drivers don't support it but translation is similar)

src/status_im/chat/commands/impl/transactions.cljs
flexsurfer: i've made tx-obj separate because in that case I'm sure it won't be changed and will be signed as is, if we mix it with a transaction for signing screen, it might be messed

hesterbruikman:
@flexsurfer Great work @flexsurfer and @andmironov ! I loved the improvement this overview screen brings over the current transaction signing flow already on first use.

Observations
1. In case of an address in the [To field] on the [signing screen] I think we need to make a VD desicion @andmironov :
- Truncating further (and allow users to go back to check the full address), 
- Decrease font size (think we can go to 13 when using Roboto mono)
- Figure out a way to show the full address. Right now the address gets to close to the cancel button IMO. I know we'll have ENS names and random names, but this address scenario will still occur.
IMG_0125

2. @andmironov can you confirm please: Font for address should be Roboto mono correct? I don't see it in the designs as it's a more recent decision.

3. I managed to set a gas limit that resulted in "Intrinsic gas too low" error. Is there a minimum we should specify or would this be resolved by the fee slider?

IMG_0130

flexsurfer:
thanks @hesterbruikman , in 3  did you set gas manually or it was estimated automatically?

hesterbruikman:
> thanks @hesterbruikman , in 3 did you set gas manually or it was estimated automatically?

Manually as I didn't have enough gas for the default. I recognize the issue with automated gas btw, but unfortunately not the details. We had someone raise the issue in #status a while back. I compared some gas settings by other wallets and ours was about 3 times as low if I recall correctly. Goran at the time mentioned this was a problem specific to developer use cases. Sorry, don't recall more.

asemiankevich:
@flexsurfer

1.  can we increase tapping area for cancel button? My fat fingers are struggling

2. can we please remove this Wrong password tooltip when i start re-entering password? I believe it is something we already do have, not against to address it in separate issue
image

3. sometimes we can't receive gas estimation from network (expected for now i believe as i got from your comments)
image

4. is it expected we do not show Asset field when i send / receive transaction in chat? @hesterbruikman 
image

flexsurfer:
thanks @asemiankevich 
1 - yes, will do
2 - yes, will do
3 - it's a major question and we need to decide what to do here, with @hesterbruikman maybe
option use hardcoded values in such case for eth and tokens as before, but it won't work for contracts 
4 - no it should work, i tested with STT and HND and SNT, what token do you try to send ?

asemiankevich:
@flexsurfer i should be more specific! my bad . In 1x1 chat when sending **0 ETH** transaction

flexsurfer:
hm that's an interesting case, all 0ETH transaction treated as contract interaction, i'll take a look as well

src/status_im/browser/core.cljs
```diff
@@ -274,19 +275,57 @@
...
+  ;;TODO check and test id
+  {:browser/send-to-bridge
+   {:message {:type      constants/web3-send-async-callback
+              :messageId message-id
+              :result    {:jsonrpc "2.0"
+                          :id      (int id)
+                          :result  result}}
+    :webview webview-bridge}})
+
+(defn normalize-sign-message-params
```
vitvly, line 299: Could this fn be made simpler with
```
(defn normalize [[first-param second-param]]
  (cond (ethereum/address? (second params))
       [first-param second-param]
       (ethereum/address? second-param)
       [second-param first-param]
       :else nil))
```
Can it happen that neither of [first-param second-param] are addresses? If this is impossible, one of the clauses in `cond` can go away.

src/status_im/ui/screens/wallet/send/events.cljs
```diff
@@ -266,191 +19,25 @@
...
-
-(handlers/register-handler-fx
- :wallet.send/edit-value
- (fn [cofx [_ key value]]
-   (wallet/edit-value key value cofx)))
-
-(handlers/register-handler-fx
- :wallet.send/set-gas-details
- (fn [{:keys [db]} [_ gas gas-price]]
    {:db (-> db
```
vitvly, line 25: Can be replaced with
```
{:db (update-in db [:wallet :send-transaction]
         assoc :symbol symbol
                    :amount nil
                    :amount-text nil
                    :asset-error nil)}
```

src/status_im/ui/screens/wallet/send/events.cljs
```diff
@@ -266,191 +19,25 @@
...
-                                         :account  from}
-                          :on-completed #(re-frame/dispatch [::sign-message-completed
-                                                             screen-params
-                                                             (types/json->clj %)
-                                                             password-error-cb])}})))))
+   (let [{:keys [to symbol amount]} (get-in cofx [:db :wallet :send-transaction])
+         {:keys [symbol address]} (tokens/asset-for (:wallet/all-tokens db) (keyword (:chain db)) symbol)
+         amount-hex (str "0x" (abi-spec/number-to-hex amount))
+         to-norm (ethereum/normalized-address to)]
+     (signing/sign cofx {:tx-obj    (if (= symbol :ETH)
```
vitvly, line 38: Similar code for tx-obj construction is used here (Yielding protocol impl for Send command): https://github.com/status-im/status-react/blob/401969db27f68813ecf6960a0768ab746ac5ba6b/src/status_im/chat/commands/impl/transactions.cljs#L309. Maybe makes sense to extract it into a separate fn.

vitvly (approved):

flexsurfer:
@asemiankevich 1,2,4 fixed

flexsurfer:
@asemiankevich @hesterbruikman i fixed gas estimation should work fine, but we need to decide what to show when gas estimation in progress or failed, maybe we could set default gas as before? but that will be weird if we set to default and the user presses sign and we change gas on estimated

hesterbruikman:
Thanks for looking out for those states. Will need some time to discuss this with design team as I want to make sure we have reusable loading and failure behavior for list items as a component.

Direction would be:
- <8sec Loading spinner on network fee amount to show 'in progress' 
Replaced by gas estimation when available 
- +8sec  Show failed state. 

'Sign transaction with...' should be disabled during loading and failed states.

Especially in the failed state the failed label should be present enough to nudge people to go into Network fee settings. Here a Set default option should be available. 

I would always offer 'Set default gas' as an option the users explicitly needs to tap in order to sign the transaction. @flexsurfer is it possible for this setting to overwrite the fee set by the DApp? Agree it gets confusing if the fee changes last minute.

All in all I think this offers most transparency and control. Hidden magic gets confusing... 

Cc @errorist

asemiankevich:
> @asemiankevich @hesterbruikman i fixed gas estimation should work fine, but we need to decide what to show when gas estimation in progress or failed, maybe we could set default gas as before? but that will be weird if we set to default and the user presses sign and we change gas on estimated

@flexsurfer builds are failing on Tests step, could you please take a look?

antdanchenko:
@flexsurfer I am going to push e2e test update in this PR, please mind my commit before force push

flexsurfer:
@asemiankevich fixed

asemiankevich:
thanks @flexsurfer looks good to me! Can we wait for @antdanchenko to adjust the e2e before the branch is merged?

flexsurfer:
@antdanchenko how long it may take?

asemiankevich:
@annadanchenko anton estimated e2e adjustment +1 day for this feature, should we wait or merge this feature and fix tests in develop later? Concerned about rebasing mostly. @flexsurfer @antdanchenko

annadanchenko:
@flexsurfer When signing message from Dapp (from CryptoKitties or from test dapp) there is overlap with Title in the app  and status bar of the phone (Galaxy S6, Android 6.0.1 and Pixel XL, Android 9)

note that it doesn't happen on Galaxy S10, Pixel 3 and iphone 7 plus

Screenshot_20190612-120757

antdanchenko:
@flexsurfer please check [test_send_two_transactions_one_after_another_in_dapp](https://app.saucelabs.com/tests/9a1ab8e34e6c4111ac2d81282c9c2224?auth=b35ba525af3d62a354b9f78c78522283)
Second transaction screen is not shown after sending two transactions one after another 
in Status Test DApp > Transactions tab > Send two Txs, one after another, 0.00001 and 0.00002 ETH

asemiankevich:
> @flexsurfer please check [test_send_two_transactions_one_after_another_in_dapp](https://app.saucelabs.com/tests/9a1ab8e34e6c4111ac2d81282c9c2224?auth=b35ba525af3d62a354b9f78c78522283)
> Second transaction screen is not shown after sending two transactions one after another
> in Status Test DApp > Transactions tab > Send two Txs, one after another, 0.00001 and 0.00002 ETH

yes it is not working anymore, screen is not responsive after first txn is signed. I bet it was working when i was testing it.

flexsurfer:
@asemiankevich fixed

asemiankevich:
@antdanchenko the problem is fixed as i checked manually but test is failed because it does not tap on button second time. Could you please take a look? https://ethstatus.testrail.net/index.php?/tests/view/715952

Serhy:
All e2e should be fixed now. Single test above failed due to faucet didn't work, but it's not this PR issue (manually all fine)

asemiankevich:
@annadanchenko would you mind if this branch will be merged and the issue u are referring to will be fixed in develop separately?

PR #8364, opened 2019-06-05
[POC] simplify chat components for perf

yenda:
# Testing

this PR is a POC to check if we can improve the speed of message rendering 
don't mind the wrong styling

Steps to test:
- measure the time it takes to open the #test2 public chat in develop and compare it to this PR


rasom:
@yenda you also can use Snoopy to check if there is less calls to UIManager

yenda:
@rasom nice thanks but do you have any notes on how to do it I have no idea

rasom:
@yenda will add notes tomorrow

Issue #8366, opened 2019-06-05
Can't launch mobile app after realm upgrade: Unfortunately, Status has stopped (Android 6.0.1, Galaxy S6 and S7) 


annadanchenko:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
*Summary*: Can't launch app, see "Unfortunately, Status has stopped" popup instead. Installed Android nightly build from https://ci.status.im/job/status-react/job/nightly/1105/

Note that previous nightly https://ci.status.im/job/status-react/job/nightly/1104/ can be installed and launched fine. Assume that the issues is introduced by https://github.com/status-im/status-react/pull/8336
#### Expected behavior

[comment]: # (Describe what you expected to happen.)
app launched
#### Actual behavior
 "Unfortunately, Status has stopped" popup 

logcat contains
```
06-05 15:49:48.421  3059  3059 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
06-05 15:49:48.421  3059  3059 F DEBUG   : Build fingerprint: 'samsung/zerofltexx/zeroflte:6.0.1/MMB29K/G920FXXU5DQA8:user/release-keys'
06-05 15:49:48.421  3059  3059 F DEBUG   : Revision: '11'
06-05 15:49:48.421  3059  3059 F DEBUG   : ABI: 'arm'
06-05 15:49:48.421  3059  3059 F DEBUG   : pid: 8122, tid: 8172, name: mqt_js  >>> im.status.ethereum <<<
06-05 15:49:48.421  3059  3059 F DEBUG   : signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x460eb084
06-05 15:49:48.421  3507  3507 I KnoxTimeoutHandler: SD activityfalse
06-05 15:49:48.421  3050  3050 D libEGL  : eglTerminate EGLDisplay = 0x7ffb963b98
06-05 15:49:48.421  3059  3059 F DEBUG   :     r0 00000000  r1 00000000  r2 f5110ee1  r3 460eb084
06-05 15:49:48.421  3059  3059 F DEBUG   :     r4 d82a2b88  r5 00000000  r6 d406f3ac  r7 d406f300
[...]

PR #8367, opened 2019-06-05
temporary fix for realm issues

jakubgs:
Dropping 64bit platforms as a temporary workaround for Realm.js issues in #8366.

We will have to undo-this using #8326 once Realm.js issue is fixed.

See for context:
* https://github.com/realm/realm-js/issues/2282
* https://github.com/facebook/react-native/issues/25060

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -155,7 +155,7 @@ android {
         versionCode getVersionCode()
         versionName getVersionName()
         ndk {
-            abiFilters "armeabi-v7a", "arm64-v8a", "x86", "x86-64"
+            abiFilters "armeabi-v7a", "x86"
         }
     }
     /**


yenda (approved):

yenda:
Realm makes the bug more obvious but it seems these crashes happen a lot on 64 bits capable devices and are caused by RN, we'll have to follow up on https://github.com/facebook/react-native/issues/24261

flexsurfer (approved):

annadanchenko:
@jakubgs @yenda first launch of the app was fine but if close it and try to launch again then unfortunately, Status can't be launched and popup "Status PR has stopped" is shown.


Steps:
1. Install app
2. Launch app. Do not create any account
3. kill app
4. Launch app again

full logcat: 
[reproduce.txt](https://github.com/status-im/status-react/files/3261099/reproduce.txt)

yenda:
@jakubgs can you confirm that this is the only place where 64bits stuff has to be excluded?

yenda:
@jakubgs don't you also need to exclude arm64-v8a in packagingOptions?

jakubgs:
Yes.

jakubgs:
I've tested this on my OnePlus 6 with a Snapdragon 845 which a 64 bit cpu and it works fine. I tested as @annadanchenko did, with closing it and re-opening and everything worked fine.

But then again I installed the nightly and it works just fine, so maybe this isn't just 64 bit specific, maybe it's CPU specific.

krisc:
This did not work for me. Using Samsung S6 Edge, Android 7.

jakubgs:
Thanks for the info, try #8368.

yenda:
Closing because #8368 is a better fix

PR #8369, opened 2019-06-06
Use `initialNumToRender` property + simplify our messaging FlatList.

mandrigin:
see https://facebook.github.io/react-native/docs/flatlist#initialnumtorender

## Note to QA:
compare opening #status chatroom in this PR to any other PR, it should not be worse (and it might be even better)

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/views.cljs
+++ b/src/status_im/ui/screens/chat/views.cljs
@@ -339,33 +339,13 @@
 
 (defonce messages-list-ref (atom nil))
 
-(def ^:const initial-limit 5)
-(def ^:const second-initial-limit-step 6)
-(def second-limit (+ initial-limit second-initial-limit-step))
-(def ^:const third-initial-limit-step 11)
-(def ^:const default-limit-step 20)
-(def ^:const initial-threshold 0.5)
-(def ^:const default-threshold 2)
-
-(defonce messages-limit (reagent/atom initial-limit))
-
-(defn increment-limit [lim]
-  (+ lim
-     (case lim
-       initial-limit second-initial-limit-step
-       second-limit third-initial-limit-step
-       default-limit-step)))
-
 (defview messages-view
   [{:keys [group-chat chat-id pending-invite-inviter-name contact] :as chat}
    modal?]
   (letsubs [messages           [:chats/current-chat-messages-stream]
             photo-path         [:chats/photo-path chat-id]
             current-public-key [:account/public-key]]
-    {:component-will-mount
-     (fn []
-       (reset! messages-limit initial-limit))
-     :component-did-mount
+    {:component-did-mount
      (fn [args]
        (when-not (:messages-initialized? (second (.-argv (.-props args))))
          (re-frame/dispatch [:chat.ui/load-more-messages]))
@@ -373,12 +353,8 @@
                            {:messages-focused? true
                             :input-focused?    false}]))}
     (let [no-messages (empty? messages)
-          m-limit     @messages-limit
-          threshold   (if (= m-limit initial-limit)
-                        initial-threshold
-                        default-threshold)
           flat-list-conf
-          {:data                      (take m-limit messages)
+          {:data                      messages
            :ref                       #(reset! messages-list-ref %)
            :footer                    [chat-intro-header-container chat no-messages]
            :key-fn                    #(or (:message-id %) (:value %))
@@ -390,13 +366,9 @@
                                           :row                message
                                           :idx                idx
                                           :list-ref           messages-list-ref}])
+           :initialNumToRender        6
            :inverted                  true
-           :onEndReachedThreshold     threshold
-           :onEndReached              (fn []
-                                        (swap! messages-limit increment-limit)
-                                        (when (> @messages-limit (count messages))
-                                          (re-frame/dispatch
-                                           [:chat.ui/load-more-messages])))
+           :onEndReached              #(re-frame/dispatch [:chat.ui/load-more-messages])
            :keyboardShouldPersistTaps :handled}
           group-header {:header [group-chat-footer chat-id]}]
       (if pending-invite-inviter-name


yenda:
@mandrigin I already did that in my POC PR, but it's good to make a small separate one that can be merge already. I think it should also include getting completely rid of the threshold stuff though.

mandrigin:
@yenda okay, I will remove threshold here too

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -390,13 +374,10 @@
                                           :row                message
                                           :idx                idx
                                           :list-ref           messages-list-ref}])
+           :initialNumToRender        initial-limit
```
yenda: 0

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -341,44 +341,28 @@
 
 (def ^:const initial-limit 5)
```
yenda: all of this can be removed

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -341,44 +341,28 @@
...
          (re-frame/dispatch [:chat.ui/load-more-messages]))
        (re-frame/dispatch [:chat.ui/set-chat-ui-props
                            {:messages-focused? true
                             :input-focused?    false}]))}
     (let [no-messages (empty? messages)
-          m-limit     @messages-limit
-          threshold   (if (= m-limit initial-limit)
-                        initial-threshold
-                        default-threshold)
+          threshold   default-threshold
```
yenda: can be removed

src/status_im/ui/screens/chat/views.cljs
mandrigin: I disagree with this one, it looks much better if the first 5-7 messages are loaded already.

yenda: no it will delay opening of the screen on slow devices and design said it's a no no

yenda: they said better have a blank screen than a lag

mandrigin: if 5 messages are causing a lag, there is something wrong with message rendering and not with the container, tbh.

I think with you suggestion we just swipe the real issue under the rug and that's it.

Also, this we can also do later, if we won't be able to make rendering faster.

src/status_im/ui/screens/chat/views.cljs
mandrigin: why is that? this threshold is 2 which is very different from theshold of 0.5 that is default in RN.

yenda: you can put the threshold directly in the parameters of the flat list since it is not computed anymore

mandrigin: oh, sorry, the `default-threshold` is 0.5, will remove.

src/status_im/ui/screens/chat/views.cljs
yenda: yes there is an issue with the rendering, see https://discuss.status.im/t/performances-hunting-inefficiency/1250/8
but that means that for now, 0 is better, you can add a todo/warning that it has to be changed once we fix the bigger issue with the chat components

in my PR here https://github.com/status-im/status-react/pull/8364 I still didn't get them to render faster, the only thing I've left to try on cljs side is to pass the style as well so it's not computed in the component. After that it means we have to investigate lower levels, maybe reagent level of creating components, maybe it should even be done before idk

yenda: well they render a little bit faster but it is still painfully slow on slow devices

mandrigin: I still disagree that 0 is better for now, I think it is better to leave it as-is at the moment. When we will need to release 0.14.0, this code will be slightly faster than before, so that is an improvement to the quality of life of our users already.

If we make it 0, we will introduce blinking every time you open the chatroom, which is annoying. leaving this blinking is a last resort and I'm not convinced that we did try every other way than fixing that.

yenda: no it will improve the situation, currently unless you have a small screen device you see blinking anyway with 5 messages, because it doesn't fill the screen, with 0 you will see the `chat-intro-header-container` and then enough messages to fill the screen will be displayed as soon as loaded. So that's an improvement compared to now where it just looks like you have a chat with 5 messages on high end devices, because you see the header and 5 last messages, and then suddenly more messages appear inbetween.

yenda: I mean lets just ask @errorists what is the least worst situation here:
- show 5 messages and chat intro then fill messages inbetween later
- show chat header only then show messages when they are loaded

mandrigin: it isn't 5 anymore, btw, but yeah, the size should be somehow calculated...
again, the major issue there is that our rendering of chat messages is slow for no reason. if all the possible 1.5 screen messages are loaded in < 0.3 second, you won't have this issue.

mandrigin: @yenda also, you are talking like it is something that is already decided, and that we won't improve upon it. In the end there should be neither blinking nor delays to load < 50 text messages.

yenda: 6 is worst since the screen doesn't navigate until it's rendered so it's just lag from the user of low-end device perspective, and it is still not enough in most cases for the high devices to not show the header on half the screen

yenda: No I'm talking like I already wanted to do what you are doing so I had this conversation and answer from @errorists was `better, open the blank screen as fast as possible, start rendering messages`

So my point is for this iteration, according to design, better render 0 for now. Then as soon as we figure how to render these messages as quick as they should we increase, potentially to the default flat list value to fill the screen and even a bit more

mandrigin: okay, fine, I'm not really convinced, but I want this change to be merged asap :-P

yenda (approved):

yenda:
@mandrigin just tested both:
- with 6 when there isn't big messages in the last 6 messages of the chat, I see the chat intro when navigating to the chat for a split second
- with 0 I always see it and see 1 message then after a split second the rest renders

So it seems like 0 still renders 1 message anyway, in slow-mo I clearly see that it opens with 1 message renders and then all the other renders at once. I still don't have an explanation for that behavior

mandrigin:
@yenda it might be a bug in react-native, though I'm not sure it is worth fixing now.

yenda:
@mandrigin I think a much better (still temporary fix as long as we believe we can render faster) is to show the same chat intro with loading indicator that we show when loading a chat for the first time until the messages are render, and to actually render a full batch this time. On high end devices it will be a split second, and on low end it will be a more honest view

mandrigin:
Feel free to create a separate issue for that, we can fix it later.

flexsurfer (approved):

rasom (approved):

annadanchenko:
got a crash once when was switching between 2 chats (Pixel XL, Android 9). trying to reproduce now

--------- beginning of crash
06-11 10:03:47.342  3001  3042 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x10 in tid 3042 (mqt_js), pid 3001 (tus.ethereum.pr)
06-11 10:03:47.499  5326  5326 I chatty  : uid=10840(im.status.ethereum.pr) expire 2 lines
06-11 10:03:47.504   826   826 I /system/bin/tombstoned: received crash request for pid 3042
06-11 10:03:47.525  5326  5326 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
06-11 10:03:47.526  5326  5326 F DEBUG   : Build fingerprint: 'google/marlin/marlin:9/PQ3A.190505.001/5373320:user/release-keys'
06-11 10:03:47.526  5326  5326 F DEBUG   : Revision: '0'
06-11 10:03:47.526  5326  5326 F DEBUG   : ABI: 'arm'
06-11 10:03:47.526  5326  5326 F DEBUG   : pid: 3001, tid: 3042, name: mqt_js  >>> im.status.ethereum.pr <<<
06-11 10:03:47.526  5326  5326 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x10
06-11 10:03:47.526  5326  5326 F DEBUG   : Cause: null pointer dereference
06-11 10:03:47.526  5326  5326 F DEBUG   :     r0  000000d2  r1  000086d8  r2  cb5c23f8  r3  fffffffb
06-11 10:03:47.526  5326  5326 F DEBUG   :     r4  fffffffe  r5  cb5c23f8  r6  cb5c23f8  r7  cb5c23f8
06-11 10:03:47.526  5326  5326 F DEBUG   :     r8  00000000  r9  fffffffb  r10 69e0de00  r11 b9032378
06-11 10:03:47.526  5326  5326 F DEBUG   :     ip  ccfec351  sp  cb5c2288  lr  ca11d679  pc  ccfec3c0
06-11 10:03:47.531  5326  5326 F DEBUG   : 
06-11 10:03:47.531  5326  5326 F DEBUG   : backtrace:
06-11 10:03:47.531  5326  5326 F DEBUG   :     #00 pc 001de3c0  /data/app/im.status.ethereum.pr-LoKaXtlm0UmWaW4Cuk5MEg==/lib/arm/libjsc.so (offset 0x68000)
06-11 10:03:47.531  5326  5326 F DEBUG   :     #01 pc 0000e677  <anonymous:ca10f000>


full logcat 
[cr.txt](https://github.com/status-im/status-react/files/3275400/cr.txt)

annadanchenko:
@mandrigin should https://github.com/status-im/status-react/issues/8397 be fixed by this PR too? So, messages on entire screen are shown at the same time. Currently, only first message is shown then some delay and then the rest of messages are shown (in android build 3 of this PR)

mandrigin:
@annadanchenko no, this isn't suppose to load all the messages (yet), it should load only on message and only then load everything else. I think we can fix that by the release thing, a bit separately.

annadanchenko:
after rebase it loads last 6 messages + group icon before reloading the rest of messages. Not sure it's normal as one message was expected to be loaded before the rest of them

mandrigin:
@annadanchenko I looked into this animation in slow-mo, it looks like it starts with 0 and while animating it manages to render 6 more messages. So count it as an expected, it is just too quick to render now.

On chats that have a lot of heavy formatting it works as expected (opens with 0 messages).

annadanchenko:
time to show first group of 5 messages and then the rest is relatively the same.

PR #8374, opened 2019-06-07
[WIP] more cljs modules

rasom:


yenda (commented):
Why are the main accounts.db fns moved to account.models?
https://github.com/status-im/status-react/blob/develop/doc/codebase-structure-and-guidelines.md

rasom:
@yenda 
> Why are the main accounts.db fns moved to account.models?

1. because there are "hidden" cyclic deps. I write "hidden" because we kind of bypass them by adding sub namespaces, but in fact our modules have many of them, and these deps will be revealed during modules adoption.
  For instance, `s.accounts.db` was used in `s.utils.handlers` which makes zero sense as `handlers` is meant to be an util ns and `s.accounts.db` even in that "guideline" is meant to be "db logic called by other modules". That's a very short path to cyclic deps, and I got in that trap at the moment it tried to separate `network` module.
2. There is no real reason in having specs and "getter and setter functions used by fx producing functions and subscriptions" in the same place, more than that if we define modules, there is no sense in "db logic called by other modules", because a single ns from a module should be module's entry point, and that's definitely not ".db" ns.

So it was necessary to solve cyclic deps for `network` module and you may look at it as a preparation for `accounts` module.

src/status_im/network/module.cljs
```diff
@@ -0,0 +1,42 @@
...
+   :connect                     'status-im.network.core/connect
+   :delete                      'status-im.network.core/delete
+   :get-network-id-for-chain-id 'status-im.network.core/get-network-id-for-chain-id
+   :get-network                 'status-im.network.core/get-network
+   :remove-network              'status-im.network.core/remove-network
+   :edit-network-view           'status-im.network.ui.edit-network.views/edit-network
+   :network-settings-view       'status-im.network.ui.views/network-settings
+   :network-details-view        'status-im.network.ui.network-details.views/network-details})
+
+(defn save [& args]
```
yenda, line 19: The macro should deal with all that boilerplate

yenda:
1. re-frame creates a lot of hidden dependencies by loosely coupling some parts of the code, sub namespaces don't, the circular dependencies are created when code is moved around. The handlers namespace has a dependency on accounts because some registered events have a check on login.
2. you should document and discuss all these architectural decisions first, they look promising but it's probably not gonna fly head first in a 50 files PRs. Don't make me wrong I'm all for a better structure for the project, the current one was just intending to improve on the previous one and remove most of the circular dependencies behind re-frame cascades of events

by account module you mean the settings stuff once logged in? do you measure the runtime impact of all these changes, it cannot disappear from startup time and not go somewhere else, is a module loaded at once the first time part of it is needed or just the function needed when needed

rasom:
@yenda runtime impact is dispersed as these deps are not loaded all at once and we can configure which calls require loading and which are not. I measured it for 200KB translation module and it was 90ms on a slow device, but, for instance `network` is 20KB, extensions are going to be 60KB and so on, - impact on runtime is negligible and might be completely not noticeable even on slow devices.

> re-frame creates a lot of hidden dependencies by loosely coupling some parts of the code, sub namespaces don't

re-frame stuff isn't an objective here, I'm not going to argue. But sub-namespaces are a problem. We can't say that we have a module if it doesn't have a single entry point and sub-namespaces ruin the whole thing. For example:

we have the next namespaces

`a.db`
`a.core`
`b.db`
`b.core`

and we have next deps:

`a.db` -> `b.db`
`a.core` -> `b.core`
`b.core` -> `a.db`
`b.core` -> `b.db`

This graph will will be compiled succesfully as we kind of "don't have" cyclic deps, but we actually have `a -> b -> a` cycle here. And we have many such cases. This will not work with modules and this is in general quite a slippery slope; we could just as well put the whole project into a single file. The module should have an interface which should be a single entry point, and this makes such `a -> b -> a` harmful. 

> you should document and discuss all these architectural decisions first, they look promising but it's probably not gonna fly head first in a 50 files PRs

this is a good enough place for starting the discussion;  37 files changed in this pr, check only the last commit.

yenda:
@rasom account.login and account.create aren't meant to be considered part of a same account module, they are different namespaces because they are taking care of different concerns. I'm not sure that is actually what you are talking about here but I'm just saying that they are not "sub namespaces" in the sense that they are just there to avoid circular deps, they are there because they serve a different purpose than the account namespace, more specific.

I see what you mean with the circular deps and I'm looking forward to your solution to avoid them, I hope it's not by relying on loose coupling though because I would consider it worst especially if we are talking about saving ms, but I'm all for a solution that would make the code org better as always.

rasom:
> I see what you mean with the circular deps and I'm looking forward to your solution to avoid them

the solution is pretty simple: the module should be accessed only through a single entry point (I would prefer it to be called `module` and separated from `core`) and all common stuff to be put into a... separate module :) There is no magic here

> account.login and account.create aren't meant to be considered part of a same account module, they are different namespaces because they are taking care of different concerns.

that's kind of irrelevant to the issue which appeared in this pr; as I said it was `s.accounts.db` in `s.utils.handlers` (`s.utils.handlers` -> `s.accounts.db` -> `s.newtork.module` -> `s.network.events` -> `s.utils.handlers`) which caused some difficulties. And we will reveal similar things when move module's APIs into a single ns. 

> if we are talking about saving ms

sure, every single module will save us 20-100ms on slow devices. 20 of them may save a second or so, and so on.

yenda:
@rasom ok all good, could you investigate improvements to the defmodule macro though, correct me if I'm wrong but defmodule could take a list of function and a list and view and all the boilerplate and keyword map can be generated. That would ease up readability and adoption

rasom:
@yenda I thought about this, but I would prefer to write few more modules as is just to make sure which cases we have so that it will be handled later in one go. There is no rush on this as it works just fine now, but def can be added.

rasom:
`status-im.extensions.module -> status-im.extensions.core -> status-im.extensions.capacities.events -> status-im.browser.core -> status-im.browser.permissions -> status-im.extensions.module` and here we go... :)

yenda:
@rasom that is a re-frame event loose coupling real circular dependency you got here, the reason why I insisted all events to be declared in the same namespace and only calling one function, the only way to get the compiler to figure them out, but it wasn't enforced hard enough

rasom:
@yenda that might be a solution, but in fact this code just doesn't make sense

https://github.com/status-im/status-react/blob/1ff0a937acbd04f661d87060a913067093abea7b/src/status_im/extensions/capacities/events.cljs#L309-L312

when we have all these handlers

https://github.com/status-im/status-react/blob/62a31ac58906d4ce60f66bb8947b904d81e422ec/src/status_im/events.cljs#L1516-L1530

rasom:
has been split into separate PRs

PR #8378, opened 2019-06-07
rebase only PR builds

jakubgs:
Changes:
* Reversing the logic. Only PR builds should be rebased
* Remove unnecessary `load` calls for `nix.groovy` and other libraries to reduce groovy load time
  - cuts down initial Groovy scripts load from ~25 to 10 seconds for all builds
* Moved `notifyPR` to `github.groovy` to avoid loading all of `common.groovy` at the end of every build
  - cuts down fGroovy load from ~25 to 4 sec of off all builds
  - Thanks for catching that @PombeirP!

### DIFF
--- a/ci/utils.groovy
+++ b/ci/utils.groovy
@@ -1,5 +1,3 @@
-nix = load 'ci/nix.groovy'
-
 def getVersion() {
   def path = "${env.WORKSPACE}/VERSION"
   return readFile(path).trim()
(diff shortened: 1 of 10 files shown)


pedropombeiro:
I got bitten by this just recently in my reproducible builds branch because I wanted to compare the CI build output with my local build and since package.json.orig had changed in develop, this would force me to rebase and fix conflicts locally too. I think that if I'm asking for a manual build, I know what I'm trying to build, it's not like in a long-running PR where you need to be warned that develop moved and you have merge conflicts. So in the end, the logic should be opt-in rather than opt-out.

ci/utils.groovy
```diff
@@ -43,6 +43,10 @@ def doGitRebase() {
     println 'Skipping rebase for canary build...'
     return
   }
+  if (params.BUILD_TYPE == 'manual') {
```
pedropombeiro: When I say it should be opt-in, I mean reversing the logic here:
```
if (params.BUILD_TYPE == 'pr') {
   println 'Rebasing on develop since it is a PR build...'
}
```

jakubgs: Yeah, I was trying to think of any other cases but I guess only PR builds need rebases. Changed.

pedropombeiro (approved):

pedropombeiro (approved):
Wohoo!

rasom (approved):

PR #8379, opened 2019-06-07
[fix #8375] "save password" unavailable for non bioauth devices

### LINKED ISSUE #8375: "Save password" is unavailable for devices without biometric authentication
### Description
*Type*: Bug
*Summary*: on devices without biometric authentication not possible to use "Save password" anymore.
Instead of toggle "Save password is unavailable: your device may be rooted or lacks the necessary security feature"
Most likely was introduced in https://github.com/status-im/status-react/pull/8127

#### Expected behavior
Can use "Save password"

#### Actual behavior
download (3)

### Reproduction

- Open Status
- Create an account
- Log out

### Additional Information

- Status version: nightly 07/06/2018
- Operating System: Android (AFAIK there is no more iPhones without touch ID that Status can support)
- Devices: without biometric authentication (tested on Samsung Galaxy S7, Google Nexus 6)

tbenr:
fixes #8375 

I forgot to remove biometric availability check on `an-save-user-password?` function.

fixed also `:supported-biometric-auth` in `:db` to be always `nil` when bioauth not supported (as per specs)

status: ready to test

### DIFF
--- a/src/status_im/biometric_auth/core.cljs
+++ b/src/status_im/biometric_auth/core.cljs
@@ -71,9 +71,9 @@
 (defn get-supported [callback]
   (cond platform/ios? (do-get-supported callback)
         platform/android? (if android-device-blacklisted?
-                            (callback false)
+                            (callback nil)
                             (do-get-supported callback))
-        :else (callback false)))
+        :else (callback nil)))
 
 (defn authenticate
   ([cb]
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

tbenr:
removed require no more ... required

mandrigin (approved):

Serhy:
Looks good to me, @tbenr ! 
- On Android device where I lost ability to save password - I can save it fine.
- On iOS with bio auth feature everything kept on place as before: can use bio-auth, can disable it and proceed with login in to account via password enter.
- With rooted Android no way to use autologin, bio-auth disabled as well.

PR #8380, opened 2019-06-07
run checks and build steps in parallel

jakubgs:
By running Lint and Test stages in parallel to build stages we lose ~4 min on ever build.

Full manual build in 17 min, compared to usual ~24 of a nightly:
https://ci.status.im/job/status-react/job/manual/354/

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -52,24 +52,32 @@ pipeline {
         }
       }
     }
-    stage('Lint') {
-      steps {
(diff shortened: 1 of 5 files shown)


pedropombeiro (approved):

mandrigin (approved):

jakubgs:
When a build appears longer than others, that's because retarded Jenkins adds the time the builds waits for an executor(host) to the duration time. Makes no sense whatsoever, but ah well.

PR #8392, opened 2019-06-10
[perf] network module

rasom:
please review only the last commit

`status-im.screens.ui.network.*` was moved to `status-im.network.ui.*`
`accounts` module was a bit rearranged in order to fix cyclic deps `s.utils.handlers -> s.accounts.db -> s.newtork.module -> s.network.events -> s.utils.handlers`

status: ready

### DIFF
--- a/src/status_im/accounts/model.cljs
+++ b/src/status_im/accounts/model.cljs
@@ -1,0 +1,20 @@
+(ns status-im.accounts.model
+  (:require [status-im.utils.security :as security]
+            [cljs.spec.alpha :as spec]
+            [clojure.string :as string]))
+
+(defn logged-in? [cofx]
+  (boolean
+   (get-in cofx [:db :account/account])))
+
+(defn credentials [cofx]
+  (select-keys (get-in cofx [:db :accounts/login]) [:address :password :save-password?]))
+
+(defn account-creation-next-enabled? [{:keys [step password password-confirm name]}]
+  (or (and password (= :enter-password step) (spec/valid? :status-im.accounts.db/password (security/safe-unmask-data password)))
+      (and password-confirm (= :confirm-password step) (spec/valid? :status-im.accounts.db/password password-confirm))
+      (and name (= :enter-name step) (not (string/blank? name)))))
+
+(defn current-public-key
+  [cofx]
+  (get-in cofx [:db :account/account :public-key]))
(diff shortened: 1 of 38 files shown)


mandrigin (approved):

src/status_im/accounts/model.cljs
```diff
@@ -0,0 +1,20 @@
+(ns status-im.accounts.model
```
flexsurfer, line 1: i believe usually we use core name

src/status_im/network/events.cljs
```diff
@@ -0,0 +1,58 @@
+(ns status-im.network.events
+  (:require [status-im.network.core :as network]
+            [status-im.utils.handlers :as handlers]
+            [re-frame.core :as re-frame]
+            [status-im.utils.utils :as utils]))
+
+(handlers/register-handler-fx
```
flexsurfer, line 7: with new fx/defn we don't need to separate events anymore, just use {:events []} in fx/defn and you can move all these handlers to core

src/status_im/network/module.cljs
```diff
@@ -0,0 +1,52 @@
...
+(defn network-settings-view []
+  [(get-symbol :network-settings-view)])
+
+(defn network-details-view []
+  [(get-symbol :network-details-view)])
+
+;; Preloaded handlers, subs, functions
+
+(re-frame/reg-sub
+ :get-network-id
```
flexsurfer, line 49: why this sub is here not in subs?

src/status_im/accounts/model.cljs
rasom: not really, the code from `db` ns was moved there in order to solve circular dependency. It is "must contain getter and setter functions used by fx producing functions and subscriptions" from https://github.com/status-im/status-react/blob/develop/doc/codebase-structure-and-guidelines.md , which has no real reason to be in the same ns with specs. I can rename it differently if you have some ideas.

src/status_im/network/events.cljs
rasom: The whole objective was rather to rearrange code so that it can be compiled as a module, not to rewrite it. I know about this but it can be rewritten later if necessary.

src/status_im/network/module.cljs
rasom: because this sub is needed outside of module right after logging in, when the rest of module's namespaces can still wait to be loaded. It could be put into a separate ns, but wouldn't make too much difference.

flexsurfer (approved):

cammellos (approved):

asemiankevich:
@rasom would you mind to rebase please? We have new sign txn screen in develop, this is the reason for autotests to fail. Thank you!

rasom:
@asemiankevich it was rebased before your comment

PR #8393, opened 2019-06-10
[perf] extensions module

rasom:
please review only the last commit

status: ready

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -121,18 +121,38 @@
                                                          :externs            ["externs.js"]
                                                          :language-in        :es-2015
                                                          :language-out       :es-2015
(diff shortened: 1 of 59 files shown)


mandrigin (approved):

jeluard (approved):

asemiankevich:
@rasom could u please advise if we need to test any visible perf improvements here , like:
- installing extension
- open extension 
- switching between extension related screens

Or we just need to check that extensions are not broken here? Probably you did some measurements by yourself too? Would be good to know! Thank you!

rasom:
> Or we just need to check that extensions are not broken here?

that's the only thing to be checked actually. You can also compare startup, it should be a bit faster on slower devices (up to 200ms), but more important to make sure nothing is broken.

Issue #8383, opened 2019-06-07
[PERF] Minify html, js and css


yenda:
# Problem

When opening a webview, we inject a modified version of web3 and 25k of non minified code of ours.
The web3 code is slurped and hence there is a 140ko string in the advanced compiled code (see #8381)

Extensions are also slurping a non minified 4k html file

## Implementation

Minify our html/css/js in the build pipeline
Find a way to inject web3 without slurping it into the bundle

## Acceptance Criteria

Can't find web3 in index.android.js
Can't find our html/css/js in index.android.js
Both are injected minified in the webview

rasom:
this is handled by #8394 except magnification, but assuming it has very love impact on performance there is no real need to do magnification here

pedropombeiro:
Is this issue still relevant?

Issue #8381, opened 2019-06-07
[PERF] Replace use of slurp macro


yenda:
# Problem

Using the slurp macro is bad, whatever you slurp ends up being copied as is as a string in the minified js produced by the compiler, even with advanced optimization.

## Implementation

Each use case of the slurp macro is unique. The easiest case to get rid on is json config files, because they can be used like modules. @rasom can probably describe how to proceed

## Acceptance Criteria

The slurp macro is deleted

rasom:
Implemented in #8394, though impact on performance is kind of limited. On the other hand it makes it a bit simpler to reason about the rest of index.*.js, because besides `status-im.utils.gfycat.*` the rest of it is actually js code, not string resources.

PR #8394, opened 2019-06-10
[perf] prod version of slurp

rasom:
solves #8381
solves #8383

review only the last commit

All resources loaded by slurp are moved to status-modules/resources dir
in release builds and are loaded only by demand instead of being bundled into
index.*.js.

status: ready

### DIFF
--- a/src/status_im/ui/screens/browser/views.cljs
+++ b/src/status_im/ui/screens/browser/views.cljs
@@ -24,8 +24,13 @@
    [status-im.utils.slurp :refer [slurp]]
    [status-im.utils.views :as views]))
 
+(def browser-config-edn
+  (slurp "./src/status_im/utils/browser_config.edn"))
+
 (def browser-config
-  (edn/read-string (slurp "./src/status_im/utils/browser_config.edn")))
+  (memoize
+   (fn []
+     (edn/read-string (browser-config-edn)))))
 
 (defn toolbar-content [url url-original {:keys [secure?]} url-editing?]
   (let [url-text (atom url)]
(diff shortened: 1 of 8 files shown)


src/status_im/ui/screens/browser/views.cljs
```diff
@@ -24,8 +24,14 @@
    [status-im.utils.slurp :refer [slurp]]
    [status-im.utils.views :as views]))
 
+(def browser-config-edn
+  (slurp "./src/status_im/utils/browser_config.edn"))
+
 (def browser-config
-  (edn/read-string (slurp "./src/status_im/utils/browser_config.edn")))
+  (let [config (atom nil)]
```
flexsurfer: could elaborate on atom? what the purpose?

flexsurfer (approved):

src/status_im/ui/screens/browser/views.cljs
rasom: caching of the result, probably simpler to wrap with `memoize`

rasom: does it look better now?

flexsurfer: it's not about look. i just didn't notice that it's `def` not `defn`, but this looks better :)

mandrigin (approved):

Serhy:
No regressions in DApp and browsing on iOS and Android (8th, 6.0.1 tested).
Also, can install and use extensions fine (gfycat, wallet 'hello world' extension), can revoke access to Dapp from Profile

PR #8402, opened 2019-06-11
Minor fixes

pedropombeiro:

### Summary

This PR adds code to remove watchman watches in the CI server after the build completes, and adds some minor fixes.

Documentation change PR (review please): https://github.com/status-im/status.im/pull/362

status: ready


ci/Jenkinsfile.android
```diff
@@ -103,6 +103,7 @@ pipeline {
     }
     stage('Cleanup') {
       steps {
+        cmn.nix.shell 'watchman watch-del $(pwd)'
```
jakubgs: Wouldn't it make more sense to just run this in the `clean` target of our `Makefile`?

jakubgs (approved):

flexsurfer:
no change for xcode version?

ci/Jenkinsfile.android
pedropombeiro: That's a dev-side scenario and outside of the scope of this PR (though we could add it if we're sure of what we want to do there). This is just to make sure we don't leave new watches running in CI.

pedropombeiro:
> no change for xcode version?

@flexsurfer We could switch to require 10.2 if everyone agrees. I think it makes more sense than requiring 10.1, since it's easier to upgrade than to downgrade. Loosening the requirements to 10.x takes more work, since the default script in nixpkgs doesn't support that.

ci/Jenkinsfile.android
jakubgs: Then at least make it a separate target like `watchman-clean` and call it with `sh 'make watchman-clean'`, othewise you have to add a `script {}` block in the Jenkinsfile.

jakubgs (approved):

pedropombeiro:
Please review docs in https://github.com/status-im/status.im/pull/362

cammellos (approved):

PR #8403, opened 2019-06-11
Scan URL with QR Scanner from Chats

krisc:
This PR resolves this bounty https://github.com/status-im/status-react/issues/5776

Previously was this PR https://github.com/status-im/status-react/pull/7880, but issues with nix, it was easier for me to make a new PR.

@churik had some issues in the old PR, but I currently can't recreate them in this PR. Can someone else confirm that these issues are not present?
https://github.com/status-im/status-react/pull/7880#issuecomment-483599614

She also mentioned that Ethereum addresses are also a valid link, and I am not sure what to do about that. I feel like they should indeed be valid. Let me know what I should do about that.

Also just a note, the changes I made to `src/status_im/browser/core.cljs` are relevant to https://github.com/status-im/status-react/pull/8327 that has not been resolved yet. I wasn't sure if I should have left it in or not, but I just left it in. Let me know if I should remove it.

### DIFF
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -201,6 +201,12 @@
     (universal-links/handle-url cofx link)
     {:browser/show-browser-selection link}))
 
+(fx/defn browser-selection-cancel
+  [{:keys [db]}]
+  (let [view-id (:view-id db)]
+    (if (= view-id :qr-scanner)
+      {:db (assoc-in db [:navigation/screen-params view-id :barcode-read?] false)})))
+
 (fx/defn update-browser-on-nav-change
   [cofx url error?]
   (let [browser (get-current-browser (:db cofx))
--- a/src/status_im/ui/screens/add_new/models.cljs
+++ b/src/status_im/ui/screens/add_new/models.cljs
@@ -4,16 +4,23 @@
             [status-im.i18n :as i18n]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.universal-links.core :as universal-links]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.browser.core :as browser]
+            [status-im.constants :as constants]
+            [status-im.extensions.core :as extensions]))
+
+(defn- valid-url? [url]
+  (boolean (re-matches constants/regx-url url)))
 
 (fx/defn process-qr-code
   [cofx data]
-  (if (spec/valid? :global/public-key data)
-    (universal-links/handle-view-profile cofx data)
-    (or (universal-links/handle-url cofx data)
-        {:utils/show-popup {:title      (i18n/label :t/unable-to-read-this-code)
-                            :content    (i18n/label :t/use-valid-qr-code {:data data})
-                            :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
+  (cond (spec/valid? :global/public-key data)  (universal-links/handle-view-profile cofx data)
+        (universal-links/universal-link? data) (universal-links/handle-url cofx data)
+        (universal-links/deep-link? data)      (universal-links/handle-url cofx data)
+        (valid-url? data)                      (browser/handle-message-link cofx data)
+        :else   {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
+                                    :content (i18n/label :t/use-valid-qr-code {:data data})
+                                    :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}))
 
 (fx/defn handle-qr-code
   [cofx data]
--- a/env/dev/env/config.cljs
+++ b/env/dev/env/config.cljs
@@ -1,6 +1,6 @@
 (ns env.config)
 
 (def figwheel-urls {:android "ws://localhost:3449/figwheel-ws",
- :ios "ws://192.168.0.9:3449/figwheel-ws",
+ :ios "ws://192.168.1.88:3449/figwheel-ws",
  :desktop "ws://localhost:3449/figwheel-ws"}
 )
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -1398,6 +1398,11 @@
    (browser/handle-message-link cofx link)))
 
 (handlers/register-handler-fx
+ :browser.ui/browser-selection-cancel
+ (fn [cofx [_ _]]
+   (browser/browser-selection-cancel cofx)))
+
+(handlers/register-handler-fx
  :browser.ui/remove-browser-pressed
  (fn [cofx [_ browser-id]]
    (browser/remove-browser cofx browser-id)))
--- a/src/status_im/ui/components/list_selection.cljs
+++ b/src/status_im/ui/components/list_selection.cljs
@@ -42,7 +42,8 @@
                         :action #(re-frame/dispatch [:browser.ui/open-url link])}
                        {:label  (i18n/label (platform-web-browser))
                         :action #(.openURL (react/linking) (http/normalize-url link))}]
-         :cancel-text (i18n/label :t/browsing-cancel)}))
+         :cancel-text (i18n/label :t/browsing-cancel)
+         :on-cancel   #(re-frame/dispatch [:browser.ui/browser-selection-cancel])}))
 
 (defn browse-in-web-browser [link]
   (show {:title       (i18n/label :t/browsing-title)
--- a/translations/en.json
+++ b/translations/en.json
@@ -383,7 +383,7 @@
     "currency-display-name-npr": "Nepal Rupee",
     "backup-recovery-phrase": "Backup recovery phrase",
     "send-request": "Send request",
-    "use-valid-qr-code": "This QR code doesn't contain a valid universal link, contact code or username: {{data}}",
+    "use-valid-qr-code": "This QR code doesn't contain a valid universal link, url, contact code or username: {{data}}",
     "paste-json": "Paste JSON",
     "browsing-site-blocked-title": "This site is blocked",
     "browsing-site-blocked-description1": "We detected potential malicious activity from this address. To protect you and your wallet, we're preventing further navigation.\n\nIf you think this is an error, let us know in the ",


mandrigin (approved):

src/status_im/browser/core.cljs
```diff
@@ -200,6 +200,12 @@
     (universal-links/handle-url cofx link)
     {:browser/show-browser-selection link}))
 
+(fx/defn browser-selection-cancel
```
flexsurfer, line 204: you can use `{:events [:browser.ui/browser-selection-cancel]}` in that case you don't need to register handler in events ns

flexsurfer (approved):

jeluard:
> She also mentioned that Ethereum addresses are also a valid link, and I am not sure what to do about that. I feel like they should indeed be valid. Let me know what I should do about that.

Probably the best option would be to have all QR scanners support all types of URL?

src/status_im/ui/screens/add_new/models.cljs
```diff
@@ -4,16 +4,24 @@
             [status-im.i18n :as i18n]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.universal-links.core :as universal-links]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.browser.core :as browser]
+            [status-im.constants :as constants]
+            [status-im.extensions.core :as extensions]))
+
+(defn- valid-url? [url]
```
jeluard, line 12: What kind of URL do we validate here?

src/status_im/ui/screens/add_new/models.cljs
```diff
@@ -4,16 +4,24 @@
...
 
 (fx/defn process-qr-code
   [cofx data]
   (if (spec/valid? :global/public-key data)
     (universal-links/handle-view-profile cofx data)
     (or (universal-links/handle-url cofx data)
-        {:utils/show-popup {:title      (i18n/label :t/unable-to-read-this-code)
-                            :content    (i18n/label :t/use-valid-qr-code {:data data})
-                            :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
+        (if (valid-url? data)
```
jeluard: Maybe use some `cond` instead of the mix of or and if? That would require a predicate to detect universal links.

hesterbruikman:
Thanks @krisc! Gave it a test run with 3 cases:

#### General QR code scanning
1. [+-button] > [Scan QR code] to scan
a [x] tap Profile tab > [Share my Profile]
b [x] tap Wallet tab > [Receive]

#### Scanning from wallet
2. Wallet tab > [Send] > [Scan QR code] to scan
a [x] tap Profile tab > [Share my Profile] _Error is correct, but needs new copy_
b [x] tap Wallet tab > [Receive]

#### Scanning from chat
3. [+-button] > [Start new chat] > [Scan icon] to scan
a [x] tap Profile tab > [Share my Profile]
b [x] tap Wallet tab > [Receive] _Error is correct, but needs new copy_

2a and 3b should indeed be _invalid_ as @churik mentioned in point 4 (https://github.com/status-im/status-react/pull/7880#issuecomment-483599614). Thanks for making sure we get this right. The reason they are different stems from the technicalities of keys used inside Status. At the moment your contact code and wallet address are basically the same thing. Therefore yes, you would expect it to be valid to scan either whether you want to chat or transact. However, this is not necessarily the case going forward. The key used for chat and the key used for wallet are theoretically different now and in the near future will actually be different.

### Requested changes
For 2a, scanning a chat key from wallet, could you replace the copy in the error message with:

> **No wallet address**
> The QR code you scanned doesn’t seem to contain a wallet address. Please try another QR code or enter recipient address.
QR scanning in wallet


For 3b, scanning a wallet address when starting a new chat, could you replace the copy in the error message with:

> **No Chat key**
> The QR code you scanned doesn’t seem to contain a Chat key. Please try another QR code or enter the Chat key.

Also, could you try to have the camera close and the error show up on the New chat screen that holds the input field. This gives more context to the error message.
QR scanning in chat

src/status_im/browser/core.cljs
krisc: I'd like to keep it the way it is, unless you think it's preferable otherwise

krisc:
> > She also mentioned that Ethereum addresses are also a valid link, and I am not sure what to do about that. I feel like they should indeed be valid. Let me know what I should do about that.
> 
> Probably the best option would be to have all QR scanners support all types of URL?

I agree with Scan QR code from Chat's tab, but not from Add new contact (should just be chat profiles)

src/status_im/ui/screens/add_new/models.cljs
krisc: it uses `regx-url` defined in `status-im.constants`

src/status_im/ui/screens/add_new/models.cljs
krisc: okay I will clean this up.

krisc:
@hesterbruikman , thanks for the feedback! I'll get to work on these requested changes.

krisc:
@hesterbruikman I'm beginning your request for changes for 2a and 3b, but I'm wondering if these changes should even be part of this PR as they don't really address the original issue https://github.com/status-im/status-react/issues/5776 although they have to do with the QR Scanner. I'm just wondering if your requests should be their own issue with possibly their own bounty. I don't want to clutter this PR with a bunch of commits that will end up being squashed and undocumented.

I'll work on these requests for changes either way :) I'm just not sure how we want to keep this organized from a project management perspective

hesterbruikman:
@krisc I see what you mean. I'll set up seperate bounty in the coming days. There's also some related QR behavior I want to look into. I'll keep you posted once I set up new issue(s)!

krisc:
> @krisc I see what you mean. I'll set up seperate bounty in the coming days. There's also some related QR behavior I want to look into. I'll keep you posted once I set up new issue(s)!

Cool. I seem to be deep into the QR scanner issues lol

krisc:
@jeluard I tried cleaning it up like this:
```
(fx/defn process-qr-code
  [cofx data]
  (if (spec/valid? :global/public-key data)
    (universal-links/handle-view-profile cofx data)
    (cond (universal-links/universal-link? data) (universal-links/handle-url cofx data)
          (valid-url? data) (browser/handle-message-link cofx data)
          :else {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
                              :content (i18n/label :t/use-valid-qr-code {:data data})
                              :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}})))
```
But the behavior was different, eg., wallet address was interpreted as a URL to open in the browser for some reason.

If it's okay, I think I'd prefer to delay cleaning up this area of the code until after @hesterbruikman creates new QR-related issues as was discussed [here](https://github.com/status-im/status-react/pull/8403#issuecomment-501787251) as I'll probably make more changes to this section the code.

I believe the code works as expected right now. But I would like to know why it says `Mobile e2e tests — Failure - e2e tests are failed`. This seems to happen everytime I make a PR.

src/status_im/browser/core.cljs
jeluard: We are moving to this new method so yes it would be preferable

src/status_im/ui/screens/add_new/models.cljs
jeluard: Probably cleaner to have this logic in a dedicated namespace. Do we have a relevant one already?

jeluard:
@krisc I'd rather clean it up now, delayed cleanup rarely happen. Seems trivial to have a single cond?

churik:
@krisc  e2e failure is not related to your PR, this is an infrastructure issue.

krisc:
> @krisc I'd rather clean it up now, delayed cleanup rarely happen. Seems trivial to have a single cond?

fair enough. I'll try get to this when I get a chance today or tomorrow.

src/status_im/ui/screens/add_new/models.cljs
krisc: let me know where this helper function should be placed and I will move it. Otherwise i could just have it's implementation within the `cond` statement itself and not define any helper function.

krisc:
@jeluard **EDIT**: I now have this:

```
(fx/defn process-qr-code
  [cofx data]
  (cond (spec/valid? :global/public-key data)  (universal-links/handle-view-profile cofx data)
        (universal-links/universal-link? data) (universal-links/handle-url cofx data)
        (universal-links/deep-link? data)      (universal-links/handle-url cofx data)
        (valid-url? data)                      (browser/handle-message-link cofx data)
        :else   {:utils/show-popup {:title   (i18n/label :t/unable-to-read-this-code)
                                    :content (i18n/label :t/use-valid-qr-code {:data data})
                                    :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}))
```

I added a check for deep-links to account for ethereum wallet addresses. With that said, I'm not sure how to manually test for universal-links. Other than that, this works as expected. I still have to change the method of handling events and to move the `valid-url?` function somewhere. Let me know what to do about that.

src/status_im/browser/core.cljs
krisc: I don't completely follow, but is this what you mean? In `list_selection.cljs` the `browse` fn will be define as such:

```

(defn browse [link]
  (show {:title       (i18n/label :t/browsing-title)
         :options     [{:label  (i18n/label :t/browsing-open-in-status)
                        :action #(re-frame/dispatch [:browser.ui/open-url link])}
                       {:label  (i18n/label (platform-web-browser))
                        :action #(.openURL (react/linking) (http/normalize-url link))}]
         :cancel-text (i18n/label :t/browsing-cancel)
         :on-cancel   #(re-frame/dispatch {:events [:browser.ui/browser-selection-cancel]})}))
```
And this:
```
(handlers/register-handler-fx
 :browser.ui/browser-selection-cancel
 (fn [cofx [_ _]]
   (browser/browser-selection-cancel cofx)))
```
is no longer needed in the events ns?

jeluard:
Probably out of scope for this PR. I guess we are fine now.

churik:
@krisc please, rebase it to current `develop`.
Thanks!

churik:
@krisc thank you for the contribution!

#### 1) After opening URL in native browser can't scan valid URL
Steps: 
- tap on + 
- scan QR code that contains link
- tap on open on IOS (or Android)
- go back to Status
- try to scan same QR with link
Expected result: can scan QR
Actual result: QR is not recognizable again

#### 2) White screen after scanning ethereum address 
Example of QR:
photo_2019-04-16 12 14 49
When scanning this QR, the application ends up with the white screen.

**Platforms:** IOS, Android
**Logs:** 
[Status-debug-logs.zip](https://github.com/status-im/status-react/files/3310812/Status-debug-logs.zip)
Basically, I don't think it is a good idea to leave the camera opened after scanning valid QR - in other places (i.e. after scanning valid key or address in wallet) - camera is closed after scanning. I guess that is because they used the same view (i.e. chats -> after scanning chats) but anyway will be more consistent to close it.

krisc:
@churik thanks for the feedback. Issue 2 acts as expected for me, ie., Goes to Send Transaction screen. Not sure why you're getting an error. As for issue 1, should it return to the home screen when changing back to the Status app?

churik:
Sorry @krisc for the late feedback.
It's OK if it returns to the home screen when changing back to the Status app.

flexsurfer:
hey @krisc any updates? thanks

krisc:
Been busy with hackathons, and another one is starting next week too. But I'll try to see if I can possibly take a look again this week. But if anyone else wants to help me with this, I'll help them get onboard with the issue.

oskarth:
@krisc any chance you can take a look at this? Otherwise I'd suggest we close this PR

krisc:
> @krisc any chance you can take a look at this? Otherwise I'd suggest we close this PR

Sorry. Didn't have a chance before the Grow Ethereum hackathon. Seems like it's hackathon season lol. I won't be able to work on Status issues until September.

oskarth:
@krisc No worries, completely understandable! Let's close this PR for now and we can re-open it in the future if things change. Hope that's OK for you.

Keep hacking!

PR #8425, opened 2019-06-17
[#8348] [Multi-Account] Account overview + individual wallet UI

### LINKED ISSUE #8348: [Multi-Account] Account overview + individual wallet UI
### User Story
As a user, I want to see both an overview of all my wallets _and_ a detailed view of each specific wallet, so that I can manage my multiple accounts confidently. 

### Description
_Type_: Feature

_Summary_: 

Implementation plan for multi-account from Core UI POV is as follows:

1. @yenda to build the [`Account explorer` and `Wallet` UI](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=1743%3A303) for just one account. 
1. Underlying multi-account creation work in progress in parallel from status-go.
1. `Add account` flows built either for real or dummy accounts, depending on progress of item 2. 

Notes

* What changes must be implemented by design to help clarify the purpose of the user's 2 default accounts (chat and 'main')?
* How are we handling backwards compatibility with user's existing chat account?

#### Expected behavior
The following screens are pixel perfect for one account:

image

* Short tapping on one of the wallets from the overview screen loads the wallet (bottom row).
* Long tapping on one of the wallets from the overview screen pops up the send/receive bottom card (top row, second screen).
* Tx history is shown per wallet but not for the overview.
* `Add account` button opens bottom card with options, but options are not yet active.

flexsurfer:
fixes #8348

Replaced wallet main screen by account screen
Introduced new accounts explorer screen
Moved all events from wallet ui ns to wallet core
Removed unused wallet screens

For QA:
test all wallet screens

and these two
https://github.com/status-im/status-react/issues/8370


flexsurfer:
@errorists  could you check assets management list rendering on ios, thanks

src/status_im/events.cljs
```diff
@@ -1997,4 +2002,16 @@
...
\ No newline at end of file
+   (react/dismiss-keyboard!)))
+
+(handlers/register-handler-fx
+ :wallet-send-request
+ (fn [{:keys [db] :as cofx} [_ public-key amount symbol decimals]]
+   (assert public-key)
+   (let [request-command (get-in db [:id->command ["request" #{:personal-chats}]])]
+     (fx/merge cofx
+               (chat/start-chat public-key nil)
```
jeluard, line 2011: Why that?

src/status_im/subs.cljs
```diff
@@ -1053,6 +1068,30 @@
  (fn [[balance visible-assets]]
    (map #(assoc % :amount (get balance (:symbol %))) visible-assets)))
 
+(defn update-value [balance prices currency]
+  (fn [{:keys [symbol decimals] :as token}]
+    (let [price (get-in prices [symbol (-> currency :code keyword) :price])]
+      (assoc token
+             :price price
+             :value (when (and balance price)
```
jeluard, line 1082: Maybe value shouldn't be in the map at all when nil

jeluard (approved):

src/status_im/events.cljs
flexsurfer: i just moved this code, i think we want to send a message and . if there is no chat we want to start it

asemiankevich:
checks:
- [x] added custom token on mainnet
- [x] breed a cat on mainnet
- [x] managed assets enable / disable
- [x] all transactions in simple dapp
- [x] can check collectibles / assets / switching between
- [x] send / receive from wallet / chat
- [x] can install stickers

@errorists 
@flexsurfer  some issues, please take a look:

- [x] on android when having a lot of decimals, amount value is cut, values are not aligned (looks ok on iOS)
- [x] Add account button is cut on android (no borders)

image

- [x] Set currency is now in wallet, please remove from Profile
- [x] when set currency from new wallet screen -> tapping back does not redirect me to wallet, i have to tap the wallet tap again (some kind of UX)
- [ ] token values (prices) are not shown when account is recovered, should force it by currency switch
(then shown fine)

image

image


- [ ] can't capture screenshot on Android, says that screenshots are not allowed by app / organization
- [x] View signing phrase button is not tappable
- [x] Disable (gray-out) Add account  / Add a watch only account bottom sheet options until developed
- [ ] no incoming chat message when sending eth / tokens from wallet to contact

flexsurfer:
thanks @asemiankevich 

> when set currency from new wallet screen -> tapping back does not redirect me to wallet, i have to tap the wallet tap again (some kind of UX)

this is how react navigation works, it's not only in this case

> can't capture screenshot on Android, says that screenshots are not allowed by app / organization

not related to this PR

> View signing phrase button is not tappable

not in the scope of this PR, will disable it for now

> no incoming chat message when sending eth / tokens from wallet to contact

this is how it works, we should discuss this with UX team, but i think you expect message only when you use command in chat, it's kind of weird that when you send tokens from somewhere not from chat you send a message, maybe we could add a checkbox, send a message or smth, but again not in scope of this PR, i didn't change anything related to send tokens in this PR

hesterbruikman:
@flexsurfer first of all major kudos on getting this done so quickly! Below some observations, 1,2,4,7 need design follow-up. Hope the rest is more actionable.

1. The shadow on Android; Know you're already looking into this:)
2. The number of asset decimals throws off the fiat amount. Wraps over 2 lines instead of one. Width should be set to accommodate default 4 digits, 2 decimals. That conflicts with nr of decimals for assets which should be 18. As well as the use of the longer 3 letter currency denotation. This is currently in the designs and I believe we should follow this as the $ symbol can be used to denote different currencies. Will explore in Figma and ping once there is a more solid solution. cc @errorists @andmironov 
3. Share icon opens native share dialog; should open popover with QR code and option to copy address. https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=2556%3A31102 Not blocking, we can update this through a bounty later.
4. The from Account explorer to Single account view and vice versa shows grey boxed of the Account explorer cards during transition. Also getting a frowny face with the transition and navigation direction. Think we can specify this better in design. Can we have a call on the transitions that are possible? 
5. The collectibles tab should persist with an empty state in case the user does not have collectibles. Added an empty state here: https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=2840%3A55
6. As you followed on Figma, the cards for Account explorer and viewer are updated to accommodate Keycard identifier as well as increasing contrast ratio (now white or 70% white, instead of 40%), and standardizing the wallet address truncation. @andmironov can you confirm the font should be SF Mono and checksummed? (showing lower case and upper case)
7. Account settings not included yet, assuming you were waiting for me. Let me ping you once I've received feedback on the latest version.
8. After setting currency, navigation should go back to the Account explorer. In the PR returns to it's current settings location. Personally I'd be happy if we remove it from settings if it's an issue to specify different navigation behavior for each. All things transactional and financial in Wallet.
9. It took suspiciously long to include the token value in the total asset amount. I didn't time it, so it could just be deceiving, but would appreciate if you can check if there's anything that can cause a significant delay.

flexsurfer:
thanks @hesterbruikman 
1 in progress
2 in progress
3 yes popup component isn't ready, so will be implemented in a separate issue
4 it's a common issue, we've spoken about, on Android we don't have a background
5 to be done in this PR
6 to be done in this PR
7 in the separate issue, disabled for now
8 to be done in this PR
9 separate issue

errorists:
heya @flexsurfer 

Account explorer:
- The correct on scroll behaviour of the navigation bar is not implemented yet. Lmk if you need help visualising it.
- The screen doesn't have the black status bar style applied to it, if you go to Send (white status bar) and then back, the white style stays instead of reverting back to black
- The '~' character shouldn't be used when the total value is 0
- The bottom padding on cards is too small, it should be 12.
- The share icon on cards should be white
- The horizontal scroll container is clipping its content on the left side, to observe scroll to the left, you'll see the cards disappear immediately under some arbitrary white bound, instead when scrolled they should be visible all the way to device's edge
- there is no touchable feedback on the Assets - Collectibles tabs, at least none that I can observe
- the touchable for tabs is too small, I get a lot of missed taps when hitting Assets - Collectibles - History
- we removed the long press gesture right? Imo it shouldn't really be there, I doubt anyone would expect it other than looking for a way to remove an account.

Account details:
- on scroll if scroll position is larger than 0 the top navigation bar should reveal the 1px light grey divider on the bottom, and in reverse, hide it if the position is 0.
- imo this could use a Pinterest style transition on opening-closing accounts instead of the regular React Navigation push, not a bug or anything, just a suggestion for v2.
- I know it's not in Figma but it looks super weird with only a single account. When there's nothing to swipe and no other accounts, the single card should imo have the right padding at 16, so it's the same on both sides.

- We're missing monospaced fonts for account numbers. To get that working try `fontFamily: 'SF Mono' : 'Roboto Mono' : 'monospace'`

thanks buddy 🙏

flexsurfer:
@hesterbruikman @errorists 

before | after | after 2
--- | --- | ---
image | image | image


the layout doesn't work great with big numbers, we need to find the better layout, maybe remove a token name, because there is no sense to have it here, and show fiat value under asset value

hesterbruikman:
@flexsurfer I think that's a good solution. The focus of this PR should be enabling an account explorer and viewing multiple accounts. We can iterate on the other solutions to support big numbers/decimals. wdyt @errorists 

We do need to use the Currency acronym (USD) instead of symbols ($). As multiple currencies are denoted with the $ symbol. Could you add that after the fiat amount?

errorists:
@flexsurfer For the short term sure, I guess we can replace the full token name with the fiat value. Long term what I miss the most here is some indication of how much has the asset's value gone up/down.

flexsurfer:
@hesterbruikman like this?
image

flexsurfer:
@errorists 

> we removed the long press gesture right?

from where?

errorists:
> @errorists
> 
> > we removed the long press gesture right?
> 
> from where?

from account cards in account explorer, I recall it was there to allow sending-receiving, wanted to make sure that was still the case

flexsurfer:
@errorists it works fine on ios simulator, doesn't it work for you?

errorists:
> @errorists it works fine on ios simulator, doesn't it work for you?

on my device it didn't, i kept my finger on the screen for quite a while

flexsurfer:
@asemiankevich, i made changes for all comments 

Functionality not done:
- token values (prices) are not shown when account is recovered, should force it by currency switch
(then shown fine) [separate issue]
- can't capture screenshot on Android, says that screenshots are not allowed by app / organization [separate issue not related to wallet]
- no incoming chat message when sending eth / tokens from wallet to contact
- Share icon opens native share dialog; should open popover with QR code and option to copy address [separate issue]
- It took suspiciously long to include the token value in the total asset amount. [separate performance issue]

Design not done:
- The correct on scroll behaviour of the navigation bar is not implemented yet. [separate issue]
- The horizontal scroll container is clipping its content on the left side
- Pinterest style transition [separate issue]
- the single card should imo have the right padding at 16
- monospace (will be supported after onboarding PR merge)

asemiankevich:
so how it looks like:

1. wallet screen collectibles

image

2. wallet screen assets

image

3. account details
image

4. history
image

src/status_im/subs.cljs
```diff
@@ -1022,6 +1022,21 @@
...
+ (fn [[all-tokens chain]]
+   (tokens/sorted-tokens-for all-tokens chain)))
+
+(re-frame/reg-sub
+ :wallet/grouped-chain-tokens
+ :<- [:wallet/sorted-chain-tokens]
+ :<- [:wallet/visible-tokens-symbols]
+ (fn [[all-tokens visible-tokens]]
+   (let [vt-set (set visible-tokens)]
+     (group-by :custom? (map #(assoc % :checked? (boolean (get vt-set (keyword (:symbol %))))) all-tokens)))))
```
vitvly, line 1038: Maybe can be replaced with `->>` for better readability.

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -27,38 +27,45 @@
...
+              [react/view
+               [icons/icon :main-icons/next {:color colors/gray-transparent-40}]]
+              (= :check accessory)
+              [react/view
+               [icons/icon :main-icons/check {:color colors/gray}]]
+              :else
+              [react/view {:margin-right 8 :flex-shrink 1}
+               (cond
+                 (string? accessory)
+                 [react/text {:style styles/accessory-text :number-of-lines 1}
```
vitvly, line 63: Do we need any ellipsize-mode here?

src/status_im/ui/screens/wallet/accounts/views.cljs
```diff
@@ -0,0 +1,166 @@
...
+
+(views/defview assets-and-collections []
+  (views/letsubs [{:keys [tokens nfts]} [:wallet/visible-assets-with-values]
+                  currency [:wallet/currency]]
+    (let [{:keys [tab]} @state]
+      [react/view {:flex 1}
+       [react/view {:flex-direction :row :margin-bottom 8 :margin-horizontal 4}
+        [tab-title state :assets (i18n/label :t/wallet-assets) (= tab :assets)]
+        [tab-title state :nft (i18n/label :t/wallet-collectibles) (= tab :nft)]]
+       (if (= tab :assets)
```
vitvly, line 108: `cond` could work to avoid nested `if`.

src/status_im/ui/screens/wallet/accounts/views.cljs
```diff
@@ -0,0 +1,166 @@
...
+      (:code currency)]
+     [react/text {:style {:color colors/gray}} (i18n/label :t/wallet-total-value)]]))
+
+(views/defview accounts-options []
+  (views/letsubs [{:keys [seed-backed-up?]} [:account/account]]
+    [react/view {:flex-direction :row :align-items :center}
+     [react/view {:flex 1 :padding-left 16}
+      (when-not seed-backed-up?
+        [react/view {:flex-direction :row :align-items :center}
+         [react/view {:width           14 :height 14 :background-color colors/gray :border-radius 7 :align-items :center
```
vitvly, line 145: Weird formatting.

src/status_im/wallet/core.cljs
```diff
@@ -300,4 +301,54 @@
...
+
+(fx/defn set-and-validate-amount
+  {:events [:wallet.send/set-and-validate-amount]}
+  [{:keys [db]} amount symbol decimals]
+  {:db (set-and-validate-amount-db db amount symbol decimals)})
+
+(fx/defn set-symbol
+  {:events [:wallet.send/set-symbol]}
+  [{:keys [db]} symbol]
+  {:db (-> db
```
vitvly, line 321: Using `update-in` with `assoc` would be more readable.

src/status_im/wallet/core.cljs
```diff
@@ -300,4 +301,54 @@
...
+                                      :value amount-hex}
+                                     {:to   (ethereum/normalized-address address)
+                                      :data (abi-spec/encode "transfer(address,uint256)" [to-norm amount-hex])})
+                        :on-result [:navigate-back]})))
+
+(fx/defn set-and-validate-amount-request
+  {:events [:wallet.request/set-and-validate-amount]}
+  [{:keys [db]} amount symbol decimals]
+  (let [{:keys [value error]} (wallet.db/parse-amount amount decimals)]
+    {:db (-> db
```
vitvly, line 345: Same here

vitvly (approved):

asemiankevich:
@flexsurfer issue found:

- can't manage assets, they are always enabled (Ropsten oO not mainnet)

Checks that are passed:
- can share wallet address
- can switch between collectibles / assets
- can add custom token
- can set new currency
- can send / receive funds from wallet / chat
- can see account details
- can check transaction history, switch between history / assets

**NOTE:** 
- view signing phrase is out of scope for this PR
- add account is out of scope  for this PR

flexsurfer:
@asemiankevich known issue https://github.com/status-im/status-react/issues/8254

hesterbruikman:
Thanks for the screenshots @asemiankevich. 
Current behavior: If fiat currency is smaller than 0,00, only acronym is shown
Expected behavior: If fiat currency is smaller than 0,00, $0.00 and acronym are shown

@flexsurfer Given the conversation on Core UI call just now, I can create a seperate issue for this. What do you think?

flexsurfer:
@hesterbruikman should be fixed already, in the latest build

antdanchenko:
@flexsurfer I am going to fix tests today, please mind my commits before force push

errorists:
> Pinterest style transition [separate issue]

https://www.dropbox.com/s/7ugg3cugo174j4m/andrey.gif?dl=0

> The correct on scroll behaviour of the navigation bar is not implemented yet. [separate issue]

https://www.dropbox.com/s/1iawojw2i68dphf/andrey2.gif?dl=0

…just adding previews for context 👀

hesterbruikman:
@flexsurfer I noticed Account settings is not included. Do you plan on a seperate PR for that?

Next to that these are the todo's I noted for design. Am I missing anything?
- Manage assets icon
- Export account flow

> Current behavior: If fiat currency is smaller than 0,00, only acronym is shown
> Expected behavior: If fiat currency is smaller than 0,00, $0.00 and acronym are shown
> @hesterbruikman should be fixed already, in the latest build
Doesn't show in the latest build for me. @asemiankevich can you check in case I'm using the wrong build?

cc @guylouis

PR #8363, opened 2019-06-05
Partitioned topic and topic negotiation

cammellos:
Adds topic negotiation and partitioned topic for group-chats and pairing.

All the code has been implemented in statusgo: https://github.com/status-im/status-go/pull/1466

Basically all the whisper filter management is done at that level.

## Technical description

* On startup we load all chats and send a list of them to status go:

For a public chat: `{:chatId "status"}`, we create a single filter, based on the name of the chat.

For each contact added by us, each user in a group chat and each one to one chat open, we send:
`{:chatId "0x", :oneToOne true}`. This will create a chats, to listen to their contact code.

Any previously negotiated topic is also returned.

* Once loaded, we create our filters, and upsert the mailserver topics, both of which are solely based on the filters loaded.

In order to remove a chat, we delete/stopwatching first the the filter in status-react and then ask status-go to remove the filter. For a public chat we always remove, for a one-to-one we remove only if the user is not in our contacts, or in a group chat or we have a chat open. Negotiated topics are never removed, as otherwise the other user won't be able to contact us anymore.

On stopping whisper we don't have to ask status-go to remove filters as they are removed automatically.

Some more logic can be pushed in status-go, but that will be in subsequent PRs.

## Testing

This is a fairly large change and affects quite a large part of the codebase.
Specifically:

1) Group chats
2) Pairing
3) Retrieving history from mailserver
4) Creating/restoring accounts
5) Gaps
6) Compatibility with previous versions
7) Deleting/Creating one-to-one/group/public chats
8)Performance just after login (between login and plus icon is visible), it should be slightly faster if anything but roughly there should be no difference

Everything should work just in the same way.

To test the actual feature you will have check `geth.log`, in Debug level, for lines.

`Sending on negotiated topic`

`Sending on partitioned topic`

`Sending on old discovery topic`

And also you need to make sure the public key printed after is the one for the right contact (we also send pairing messages, to your own public key).

## How it works

It's all based on device-to-device communications, so make sure you pair your devices before testing (might work even if you don't pair them), and that you don't have more than 3 devices on each side.

The idea is that given a group of devices A1,A2,A3 you will send on:

`old discovery topic` that happens if 1) if you don't have information about their devices (no communication happened before) or  2) one of those devices is running an older version.

`partitioned topic` if all 3 devices are running this version, but you have not send a message back from *each one* of those devices.

`negotiated topic` if all 3 devices are running this version and you have sent a message back from each one of those devices.

#### 1 device on each side

[...]

PR #8431, opened 2019-06-19
Fix bug with outgoing-status

cammellos:
outgoing-status was not added on received messages, which meant that if
we receceived a message from a different device in a public chat
:outgoing-status would not be saved in the database, resulting in the
message being displayed as not coming from us on logging out/logging in
again.

status: ready


yenda (approved):

cammellos:
closing as included in #8363

PR #8435, opened 2019-06-19
[7968] Feature Information bottom sheet on Wallet > [Receive]

### LINKED ISSUE #7968: Information bottom sheet on Wallet > [Receive]
# Problem
When people participate in a Status chat, other users can access and view their wallet address. 

Wallet address becomes accessible through:

* Profile view
* Command features in chat
* Send transaction > Select recipient
 (regardless of whether the viewing user has been added as a contact.

People are currently not informed that their wallet address is exposed when they participate in a chat. 

While work is in progress to make it such that the wallet address is not exposed to anyone other than trusted users, we want to implement a message to inform people about who can see their wallet address.

## Implementation
A contextual warning in wallet that limits unnecessary exposure, but informs at the moment exposure of the wallet address has potential for adverse effects.

* Bottom sheet is trigger by tapping on [Receive] in the Wallet tab, before copying or sharing the wallet address.
* Bottom sheet component; component is dismissable by pulling it down or tapping outside of it.
* Bottom sheet includes a 'don't show me again' checkbox. When user ticks this box, warning should no longer come up.
* Bottom sheet should come up for all users, irrespective of whether they already have value in their Wallet.

For behavior see: https://www.figma.com/proto/M6BzDsCWdrHZxzy7RUzoCCLz/Warnings-and-alerts?node-id=17%3A313&viewport=8%2C-200%2C0.7&scaling=scale-down

Wallet address warning

## Acceptance Criteria
Use of bottom sheet component with content and behaviour as specified above. 
https://github.com/status-im/status-components

## Future Steps
Copy will need to be updated, removing the first sentence: "People you talk to on Status can see this code"

bitsikka:
Fixes #7968 

### Summary
- Implemented the feature as requested and specified
- bonus: added "Supress wallet receive warning" setting in Profile as well, in case user decides to enable warning for whatever reason - maybe not needed

### Dev Review notes
- implemented new event via new process in fx/defn
- not entirely sure about naming and appropriate namespace

### Testing notes
- Tested only in iOS-simulator
- Tested only in fresh install (as opposed to existing install upgraded)

status: ready


hesterbruikman:
Hi @bitsikka, great work, thank you! We are in an unfortunate situation. 
Apologies for not informing you sooner!

As time past by we are now closer to seperating your chat key from using your wallet address. Last week we also reached the conclusion that the Chat key will not be exposed to receive assets on. Meaning that there will no longer be a need to warn people that they will share their transactions with those who have access to their chat key.

I want to ask you to hold off on merging. Most likely we will not have a release before this split of keys, after which the warning becomes obsolete. 

cc @churik @antdanchenko @corpetty

bitsikka:
> Hi @bitsikka, great work, thank you! We are in an unfortunate situation.
> Apologies for not informing you sooner!
> 
> As time past by we are now closer to seperating your chat key from using your wallet address. Last week we also reached the conclusion that the Chat key will not be exposed to receive assets on. Meaning that there will no longer be a need to warn people that they will share their transactions with those who have access to their chat key.
> 
> I want to ask you to hold off on merging. Most likely we will not have a release before this split of keys, after which the warning becomes obsolete.
> 
> cc @churik @antdanchenko @corpetty

@hesterbruikman got it
I did have some suspicions whether this was still relevant, as I have been watching recent big/sweeping changes happening all around that makes old issues like this shaky to work on.

Will ask on hunches like that next time, specially after my own delays too :p.

If `receive` screen is still going to remain unchanged:

A minor part might still be relevant, in that `receive` screen (without the bottom sheet) was also updated in this PR to match the design - which it is not quite matched, especially verbiage at the top. I could strip everything else out, if needed.

PR #8440, opened 2019-06-19
Improved chat message tapping area

jeluard:
### Summary

Better tapping area for message. Now tapping on user name shows the user profile.

### Testing notes

No change in behaviour. Only better and new tapping area.

status: ready


errorists:
works great, however you want to include some conditional for incoming messages without the user photo like private chats. there it's missing the left padding 8.

rasom (approved):

errorists (approved):

asemiankevich:
I've checked that i can open user's profile in group chat and public chat by tapping user name, also avatar is still tappable, 1x1 chat looks okay to me.

I did not notice any regressions here, looks good to merge.

PR #8453, opened 2019-06-21
[8452] remove scroll-view hack

### LINKED ISSUE #8452: perhaps remove scroll-view hack in chat view with updated solution
# Problem

In the spirit of:

I had noticed this a while ago and it had been in the back of my mind for a while
https://github.com/status-im/status-react/blob/3973ebd607f4765cd8bbc874e784c596b014ad87/src/status_im/ui/screens/chat/views.cljs#L440-L445

So, I revisited the issue - followed the link in the comment and decided to try the solution in the last comment of the issue https://github.com/facebook/react-native/issues/11071#issuecomment-365417656

The newer solution in that comment seems to work. It does what the hack was needed for previously.

Maybe I missed something, which is why I added *Perhaps* in the issue title :)

## Implementation

Coming right up. Please test, verify and merge if proven useful.

bitsikka:
fixes #8452

tested in iOS-simulator and Android-avd

this affects both mobile and desktop, but I have not been able to test in desktop yet

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/styles/main.cljs
+++ b/src/status_im/ui/screens/chat/styles/main.cljs
@@ -1,9 +1,6 @@
 (ns status-im.ui.screens.chat.styles.main
   (:require [status-im.ui.components.colors :as colors]))
 
-(def scroll-root
-  {:flex 1})
-
 (def chat-view
   {:flex             1
    :background-color colors/white})
--- a/src/status_im/ui/screens/chat/views.cljs
+++ b/src/status_im/ui/screens/chat/views.cljs
@@ -438,32 +438,26 @@
     ;; components (e.g. chat-toolbar) there can be a brief visual inconsistancy like showing 'add contact'
     ;; in public chat
     (when (= chat-id current-chat-id)
-      ;; this scroll-view is a hack that allows us to use on-blur and on-focus on Android
-      ;; more details here: https://github.com/facebook/react-native/issues/11071
-      [react/scroll-view {:scroll-enabled               false
-                          :style                        style/scroll-root
-                          :content-container-style      style/scroll-root
-                          :keyboard-should-persist-taps :handled}
+      [react/view {:style     style/chat-view
+                   :on-layout (fn [e]
+                                (re-frame/dispatch [:set :layout-height (-> e .-nativeEvent .-layout .-height)]))}
        ^{:key current-chat-id}
-       [react/view {:style     style/chat-view
-                    :on-layout (fn [e]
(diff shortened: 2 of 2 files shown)


bitsikka:
upon further investigation it seems we don't even need `:on-end-editing` anymore; And `:on-blur` works just fine without the hack or `:on-end-editing`.

bitsikka:
confirmed by testing in android-avd and ios-simulator - hack or alternative solution with `:on-end-editing` not needed anymore.

most recent update by facebook and react-native-bot by marking it "resolved" and adding  **Resolution: Locked**  label also implies no need for hack.

flexsurfer (approved):

dmitryn (approved):

churik:
@bitsikka can you please define affected area (i.e. where navigation should be checked)?

yenda (approved):

bitsikka:
> @bitsikka can you please define affected area (i.e. where navigation should be checked)?

@churik when text input has focus and keyboard is shown, touching outside should hide keyboard and text input should lose focus as expected

Note: I have noted though that on android-avd(Pixel 2), touching/clicking in a small area just above text-field, it does not work as expected(lose focus / hide-keyboard) :(

Which is undesirable

I don't know for sure if it is due to this PR or there's something else causing this. I figure this PR is still better to merge since the area above text-field isn't significantly high

I intend to investigate the problem behind small area above text-field in Pixel-2 later

churik:
@bitsikka 
checked on IOS/Android according to https://github.com/status-im/status-react/pull/8453#issuecomment-507696180 in chat input, password input - looks OK to me.

Also checked on IOS/Android:
- general navigation (wallet, chat, profile)
- commands in chat
- universal links

Checked on Mac OS desktop build:
- can install / login
- can send message
- chat input is preserved when switching between chats.

Did I miss smth?
If it is enough from your POV as well, we can wait for e2e and merge.
Thanks a lot!

bitsikka:
> @bitsikka
> checked on IOS/Android according to [#8453 (comment)](https://github.com/status-im/status-react/pull/8453#issuecomment-507696180) in chat input, password input - looks OK to me.
> 
> Also checked on IOS/Android:
> 
> * general navigation (wallet, chat, profile)
> * commands in chat
> * universal links
> 
> Checked on Mac OS desktop build:
> 
> * can install / login
> * can send message
> * chat input is preserved when switching between chats.
> 
> Did I miss smth?
> If it is enough from your POV as well, we can wait for e2e and merge.
> Thanks a lot!

that's plenty 👍 thanks!

PR #8459, opened 2019-06-21
[Android] default white background on all screens

rasom:
fixes some UI issues

status: ready

### DIFF
--- a/src/status_im/ui/screens/routing/core.cljs
+++ b/src/status_im/ui/screens/routing/core.cljs
@@ -116,7 +116,7 @@
    routes
    (cond->
     (merge {:headerMode        "none"
-            :cardStyle         {:backgroundColor (when (or platform/ios? platform/desktop?) :white)}
+            :cardStyle         {:backgroundColor :white}
             #_:transitionConfig
             #_(fn []
                 #js {:transitionSpec #js{:duration 10}})


errorists (approved):

rachelhamlin (approved):

rachelhamlin:
Thanks @rasom!

mandrigin (approved):

asemiankevich:
@rasom which issues are fixed here, maybe we have some IDs ? Or any other insight? Thank you!

rasom:
@asemiankevich the complain was that background is transparent in some screens and that's why transitions look ugly. I'm sure you can find such cases in nightly, so these should be fixed

hesterbruikman:
@asemiankevich an example is account explorer on nightly Android. 
- Wallet
- Tap on main account
- Go back from main account to account explorer

asemiankevich:
alright thanks!

rasom:
well actually I don't see any difference in that wallet case...

rasom:
It seems that it is already fixed in `develop` after react-navigation upgrade, so @asemiankevich please compare with release (for instance nav from accounts list to a specific account before login).
 We still need to merge this PR as it removes unnecessary condition.

asemiankevich:
> It seems that it is already fixed in `develop` after react-navigation upgrade, so @asemiankevich please compare with release (for instance nav from accounts list to a specific account before login).
> We still need to merge this PR as it removes unnecessary condition.

i am checking with release now but we dont have the case @hesterbruikman mentioned in release yet.

asemiankevich:
i've checked this PR against nightly and release, (screens mostly) i did not notice any regressions.

PR #8468, opened 2019-06-22
Changed placeholder to show "Enter password"

demmojo:
### Summary

Originally both the password label and the placeholder were displaying "Password" on login. I corrected the placeholder to show "Enter password" instead.

#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

### DIFF
--- a/src/status_im/ui/screens/accounts/login/views.cljs
+++ b/src/status_im/ui/screens/accounts/login/views.cljs
@@ -67,7 +67,7 @@
                     :important-for-accessibility :no-hide-descendants}
         [text-input/text-input-with-label
          {:label             (i18n/label :t/password)
-          :placeholder       (i18n/label :t/password)
+          :placeholder       (i18n/label :t/enter-password)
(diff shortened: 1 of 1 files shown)


jeluard:
@demmojo Does this reflect some Figma/design decision?

demmojo:
@jeluard having both the label and placeholder be the same is redundant. It would be better to either differentiate them or even have a blank placeholder (this is done on GitHub for e.g)

jeluard:
@errorists @hesterbruikman Any thoughts about that?

errorists:
thanks for looking in. It's something we corrected in the [ongoing onboarding redesign](https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=1922%3A21041), the password label is removed and placeholder says 'Enter password'

demmojo:
Thanks for the update and sharing the redesign link! @errorists

PR #8472, opened 2019-06-23
tooltip animation with native driver

rasom:
status: ready


mandrigin (approved):

src/status_im/ui/components/list_item/styles.cljs
```diff
@@ -29,6 +29,6 @@
    :height        size})
 
 (def error
-  {:bottom-value 0
+  {:bottom-value -15
```
jeluard: Can you comment or make those defs?

src/status_im/ui/components/tooltip/styles.cljs
```diff
@@ -23,7 +23,7 @@
    :align-items :center
    :left        0
    :right       0
-   :bottom      bottom-value
+   :transform   [{:translateY bottom-value}]
```
jeluard, line 27: For perf?

jeluard (approved):

src/status_im/ui/components/list_item/styles.cljs
rasom: @jeluard I'm not sure what to comment here, I tried to make it look exactly the same way is it was before this change, and seems that these magic numbers come from other magic numbers already existing in styles (hundreds of them). So no comments unfortunately.

src/status_im/ui/components/tooltip/styles.cljs
rasom: Yes, for perf. `:bottom` can't be animated using native driver, thus animation works in js thread, loads it with unnecessary calculations and also looks laggy at some circumstances.

cammellos (approved):

src/status_im/ui/components/tooltip/styles.cljs
rasom: I changed it a bit so that instead of adding new magic number we only change number's sign :D

Serhy:
No regressions found with tooltips on Android (6.0.1, 8.0, 8.1), iOS 12, MacOS.
Thanks @rasom !

PR #8481, opened 2019-06-24
fix margin between plus button and bottom tab bar #8166

### LINKED ISSUE #8166: Fix the margin between the plus button and bottom tab bar
# Problem
With one of recent PRs that aimed to fix the bottom row of chat list not being tappable we introduced the side effect of the chat's plus button almost touching the bottom tab bar, which is bad :) It was OK when first implemented.

## Implementation
Increase the bottom margin between the tappable and bottom tab bar to 16. When in doubt, here's the [Figma file ](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=12%3A128)

## Acceptance Criteria
Margin increased, if appropriate, steps are taken so it will not happen again :pray: 

## Notes
thanks!

## Future Steps
Steps which should be taken after this issue has been resolved.

Rikku-x:
fixes #8166

### Summary

Set margin between plus button and bottom tab bar to 16.

status: ready

### DIFF
--- a/src/status_im/ui/screens/home/styles.cljs
+++ b/src/status_im/ui/screens/home/styles.cljs
@@ -188,7 +188,10 @@
 (def action-button-container
   {:position    :absolute
    :align-items :center
-   :bottom      (+ tabs.styles/tabs-diff 6)
+   :bottom      (+ tabs.styles/tabs-diff (cond
+                                           platform/ios? 16
+                                           platform/android? 0
+                                           platform/desktop? 6))
    :width       40
    :height      40})
 


src/status_im/ui/screens/home/styles.cljs
```diff
@@ -185,7 +185,10 @@
 (def action-button-container
   {:position    :absolute
    :align-items :center
-   :bottom      (+ tabs.styles/tabs-diff 6)
+   :bottom      (+ tabs.styles/tabs-diff (cond
+                                           platform/ios? 16
```
jeluard, line 192: @errorists Does that sound right to you? Different padding based on the platform?

errorists: No, not at all unless it's required to make it appear in the right place. I tested it and on iOS it's correct.

errorists (approved):
Correct on iOS, thanks! Anyone with Android?

src/status_im/ui/screens/home/styles.cljs
Rikku-x: @jeluard even though `:bottom` value is different for each platform, margin is 16.  I tested it on iOS and Android, but not on desktop (couldn't run dev build).

Rikku-x: I made it 16 margin for iOS and Android and left desktop value unchanged.

rasom (approved):

asemiankevich:
image
 
@errorists does not look very different from develop  to me ..

Rikku-x:
@asemiankevich 
I updated the margin for iOS, on Android it was also not 16, so I changed it.

asemiankevich:
e2e fails are not related to this codechange

PR #8488, opened 2019-06-25
[android] Restore compatibility with Android 6.x

mandrigin:
Updates react-native to provide `java.util.Supplier` interface on its
own, instead of relying on it being in the stdlib.

The actual fix is these 4 lines: https://github.com/status-im/react-native/commit/72d47ff50d628f140b724a6b392002da48d63ac9#diff-b9097bc72809ab84057ff8a0b1e0cce0R44

Squashed to keep patches single commits.

## Notes to QA
- install on Android 6 device
- open a chat
- shouldn't crash

status: ready

### DIFF
--- a/mobile_files/yarn.lock
+++ b/mobile_files/yarn.lock
@@ -5748,7 +5748,7 @@ react-native-webview@^5.2.1:
 
 "react-native@git+https://github.com/status-im/react-native.git#status-0.59.9":
   version "0.59.9"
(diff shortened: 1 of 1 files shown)


rasom (approved):

pedropombeiro (approved):

annadanchenko:
tested on Android 6.0.1 only. Ok to merge to unblock testing on Android 6.0.1 but autotests and tests on desktop can bring new issues

churik:
Tested on Mac OSx and Linux Ubuntu 18.04.
No regression caused by PR. Formatting in chats is working as before.

PR #8494, opened 2019-06-25
Move installations to status-go

cammellos:
Move installation management to status-go.

This commits moves all the management of installations to status-go, so that realm is not used anymore, we can't delete the schema as we need to pull them to migrate, but we delete all the records as soon as they are migrated.

### Testing
- Pairing (enabled/disabled devices)
-  Upgrade from previous version

For the upgrade from previous version, we want to make sure all the devices and names are there enabled after the installation.
To test:
1) Create a set of devices and pair them using the old version.
2) Update to this PR
3) Turn on airplane mode before login (we don't want to receive some messages, so we can make sure that the migration went ok)
4) Check that devices contains the same list, with the devices correctly named and enabled/disabled

status: ready


STATUS_GO_SHA256
```diff
@@ -1,3 +1,2 @@
 ## DO NOT EDIT THIS FILE BY HAND. USE `scripts/update-status-go.sh <tag>` instead
-
-1bhwwmypb64kf180kp634jwfppd3h6yda3qyh2icj3lz7y10pgvp
```
jakubgs, line 3: Pretty sure that empty line is necessary. See `scripts/update-status-go.sh`.

cammellos: seems to be working fine without (android builds were built correctly, the rest is because I haven't removed some binding from status-go), but I will add it back just to make sure

externs.js
```diff
@@ -109,6 +109,8 @@ var TopLevel = {
     "ease" : function () {},
     "Easing" : function () {},
     "enableInstallation" : function () {},
+    "getOurInstallation": function() {},
```
yenda: can be remvoed?

src/status_im/native_module/impl/module.cljs
```diff
@@ -145,12 +145,36 @@
     (.signGroupMembership (status) content callback)))
 
 (defn enable-installation [installation-id callback]
-  (when (status)
-    (.enableInstallation (status) installation-id callback)))
+  (let [params {:jsonrpc "2.0"
```
yenda: this could be removed and replaced by `status-im.ethereum.json-rpc/call`

src/status_im/pairing/core.cljs
```diff
@@ -192,8 +247,16 @@
   (native-module/disable-installation installation-id
                                       (partial handle-disable-installation-response installation-id)))
 
+(defn set-installation-metadata! [installation-id metadata]
+  (native-module/set-installation-metadata installation-id
+                                           metadata
+                                           (partial handle-set-installation-metadata-response installation-id metadata)))
+
+(defn get-our-installations []
+  (native-module/get-our-installations handle-get-our-installations-response))
```
yenda: status-im.ethereum.json-rpc/call could be used directly here for instance

cammellos:
@yenda updated to use `json-rpc`, would you mind reviewing again? thanks!

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -25,6 +25,13 @@
    "eth_newFilter" {:subscription? true}
    "eth_syncing" {}
    "net_version" {}
+   "shhext_enableInstallation" {}
```
mandrigin, line 28: I would say that it should be `status_`, but no biggie

mandrigin (approved):

STATUS_GO_SHA256
jakubgs: Fair enough.

rasom (approved):

churik:
Tested on Linux Ubuntu 18.04, Windows 10 Home, Mac OSx:

- upgrade from current nightly and from desktop release with a set of devices in sync. 
All names are preserved after the upgrade, can disable/enable existing devices and add new.
All devices are discoverable.
- compatibility with 0.13.0 (desktop)
- pairing with the release version
- sync of public, group, 1-1 chats, contact, username and user picture.

Serhy:
Tested builds with Android 8.0 (and 8.1) and iOS 12.3.1 
1) Release 0.13 installed
2) Paired one account between three devices
3) Two devices kept OFF for sync
4) Upgraded on to this PR build (release version built from `features/move-installations-to-status-go` branch)
As expected, all device names kept, those devices which were OFF kept OFF for sync, and those which were ON - kept ON for syncing (logged in to account after upgrade in Airplane mode)
Further messaging works and sync between devices

Just to note here (not this PR issue!) there are bugs after upgrade (confirmed with upgrading from 0.13.0 to nightly 3rd of July build):
After upgrade:
1) When I send messages in 1-1 from curtain device (not exactly clear which account affected - but reproduced with each of three devices separately) - it creates another 1-1 chat on another device (instead of sending message in 1-1 chat both are synced to)
2) Fetch 24h history in public chat - duplicates my own messages (it seems duplicated only those messages which were sent previously exactly from the device who sees duplication.) 

This PR ready for merge from me!

PR #8514, opened 2019-07-02
Use a signal for messages

cammellos:
This commits changes the behavior to read from a signal instead of
polling each filter.

We receive a signal from status-go every 0.3 seconds, only if new
messages are received. We receive a single signal for all the chats, and
we don't dispatch anymore on every message.

### Testing

Nothing should be any different. Receiving messages for one-to-one, public/group-chats/pairing

Could also someone testing performance receiving messages? (Joining a chat with many many messages should be more performant in this PR, also general performance could also be improved, as we will be doing less work/polling, but not sure about that).

I have also noticed that replies are not working between my desktop and this develop (this PR included), but they work between this PR & develop, not sure which version of desktop I am running (it might be a month old or so), but does not seem to be introduced in this PR the issue, worth checking this case against latest release.

status: ready


src/status_im/transport/core.cljs
```diff
@@ -47,13 +45,8 @@
...
-  account A messages without this."
-  [{:keys [db] :as cofx} callback]
-  (let [{:transport/keys [filters]} db]
-    (fx/merge
-     cofx
-     (transport.filters/stop-filters callback)
-     (publisher/stop-fx))))
+  "Stops whisper protocol"
+  [cofx]
+  (fx/merge
```
dmitryn: Not sure `fx/merge` is needed here. Could be just `(publisher/stop-fx cofx)` call.

dmitryn (approved):

yenda (approved):

Serhy:
@cammellos, tested with Android8, iOS 12.3, MacOS: looks good to me. No regressions found.
Can upgrade with a set of existing data (1-1/public/group chats) and proceed to message between two iOS and Android 
Receiving Push Notifications and messages from history works like in develop build
Messaging, profile data (ENSusernames in chats and userpics) behaves same to develop
Can pair three devices (contacts and chats are paired as expected, receive messages in each of three devices fine)

In respect to performance, - syncing time indeed improved. Scenario(with Release build made of `features/use-signal-for-messages` branch)  performed with public chat having 200 messages (receiving all of the messages after `Connected` status appears) same to what has been done with previous 0.13 release.
And so on GalaxyNote4 it took ~8.53s to sync, while the same scenario it's ~5s (average number of 3 reproductions - video for receiving 200 messages in public with 0.14 release version build
https://drive.google.com/file/d/1ItFLobr_tOkO1oNteJKfJvlb5yzmjG2h/view?usp=sharing) 

Ready for merge!

PR #8517, opened 2019-07-02
upgrade react-native: 0.59.9 > 0.59.10

jakubgs:
This is due to issues with 64 bit builds, for more details: https://github.com/facebook/react-native/issues/24261
This might make #8326 work and make it possible to have proper 64 bit builds required by Play Store.

### DIFF
--- a/mobile_files/package.json.orig
+++ b/mobile_files/package.json.orig
@@ -33,7 +33,7 @@
     "querystring-es3": "0.2.1",
     "react": "16.8.3",
     "react-dom": "16.4.2",
-    "react-native": "git+https://github.com/status-im/react-native.git#v0.59.9-status",
+    "react-native": "git+https://github.com/status-im/react-native.git#v0.59.10",
     "react-native-background-timer": "2.1.0-alpha.6",
     "react-native-camera": "git+https://github.com/status-im/react-native-camera.git#v1.1.5-1-status",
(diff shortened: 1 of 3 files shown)


mobile_files/package.json.orig
```diff
@@ -33,7 +33,7 @@
     "querystring-es3": "0.2.1",
     "react": "16.8.3",
     "react-dom": "16.4.2",
-    "react-native": "git+https://github.com/status-im/react-native.git#status-0.59.9",
+    "react-native": "git+https://github.com/status-im/react-native.git#status-0.59.10",
```
pedropombeiro: Please make this a tag

pedropombeiro (changes requested):

mobile_files/package.json.orig
jakubgs: Ah, good catch, looks like we missed that when Igor updated it to `0.59.9`.

pedropombeiro (approved):

flexsurfer (approved):

asemiankevich:
i've checked and builds all looks fine on android 7 / 8 and ios 12.1 / 13 beta 
- create account
- exchange messages
- exchange transactions

@jakubgs if this is sufficient we can merge.

ghost:

PR #8520, opened 2019-07-03
[#8507] fixed unknown recipient

### LINKED ISSUE #8507: Can't send transaction from wallet, recipient is shown as unknown
# Problem

Unknown recipient shown when pasting / scanning wallet address or selecting from contacts

## Steps

1. open status
2. go to wallet
3. choose recipient -> scan valid address and send txn

Repeat with copypasting address or select from contacts

Transaction is failed since recipient is shown as Unknown

image

flexsurfer:
fixes #8507

### DIFF
--- a/src/status_im/ui/screens/signing/views.cljs
+++ b/src/status_im/ui/screens/signing/views.cljs
@@ -48,11 +48,16 @@
    txt1 " "
    [{:style {:color colors/gray}} txt2]])
 
+(defn displayed-name [contact]
+  (if (or (:preferred-name contact) (:name contact))
+    (accounts/displayed-name contact)
+    (:address contact)))
+
 (defn contact-item [contact]
   [list-item/list-item {:type        :small
                         :title       (i18n/label :t/to)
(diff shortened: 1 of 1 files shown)


vitvly (approved):

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -48,11 +48,16 @@
    txt1 " "
    [{:style {:color colors/gray}} txt2]])
 
+(defn displayed-name [contact]
+  (if (or (:preferred-name contact) (:name contact))
+    (accounts/displayed-name contact)
+    (:address contact)))
```
jeluard, line 54: Do you think it's worth adding to the main `displayed-name` ?

jeluard (approved):

Serhy:
Those 5 autotests are failing due to custom username which is applied to new users (not this PR issue, the same with nightly build)
https://github.com/status-im/status-react/pull/8523 should "fix" this issue with autotests.

PR #8521, opened 2019-07-03
mark tests under construction with @pytest.mark.skip

jakubgs:
Currently a lot of end-to-end tests fail every time we run tests:
```
======== 33 failed, 108 passed, 15 skipped, 4 error in 2150.93 seconds =========
```
I see no point in running tests that we know are going to fail because they are still under construction.

When we are ready to fix them we can find them by grepping for `@pytest.mark.skip`.

### DIFF
--- a/test/appium/tests/atomic/account_management/test_profile.py
+++ b/test/appium/tests/atomic/account_management/test_profile.py
@@ -66,6 +66,7 @@ class TestProfileSingleDevice(SingleDeviceTestCase):
             self.errors.append("Can't share address")
         self.verify_no_errors()
 
+    @pytest.mark.skip
     @marks.testrail_id(5375)
     @marks.high
     def test_copy_contact_code_and_wallet_address(self):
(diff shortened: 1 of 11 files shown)


Serhy (commented):
Those tests with @marks.critical should work. At least they were passing two days ago ( https://ci.status.im/job/end-to-end-tests/job/status-app-end-to-end-tests/3084/ )
I'm ok to skip the rest though.

jakubgs:
Okay, but so what if they are "critical" if they always fail? They are effectively useless.

A clock that always shows 12:00 am is useless.

antdanchenko:
@jakubgs the issue might be not in the clock, but in a time itself, for example the time has stopped at 12:00 am :)

did you check that those tests are failing cause they are broken before skipping them?
if they are failing due to an issue in the product it makes sense to fix an issue in the product instead of skipping a test

Serhy:
They are not failing always, why. 
Todays run: https://ci.status.im/job/end-to-end-tests/job/status-app-end-to-end-tests/3095/console - all those critical @marks.testrail_id(5334), 5324, 5305, 5338, 5316, 5326, 5362 - all of those tests passed today for PR build.

If they failed in nightly build or some PR build it may be valid fail.

jakubgs:
I'm going to close it as I don't know which can and which cannot be skipped at this time.

PR #8525, opened 2019-07-04
[WIP] Implement api to work with android keychain

dshulyak:
ref https://github.com/status-im/status-go/issues/1463


oskarth:
I believe @dshulyak left so we can probably close this one, and revisit this issue once we have more resources. Feel free to re-open if this is in error.

PR #7479, opened 2019-02-12
[WIP] Change App ID for PR builds of iOS app

jakubgs:
Sister PR to https://github.com/status-im/status-react/pull/7287 for changing App ID of iOS PR builds to `im.status.ethereum.pr`.

I've added the necessary profile in `ios-certificates` repo here: https://github.com/status-im/ios-certificates/commit/6c3104907319a45efc72f05dc98ad1714d4560b3

State: WIP

PR #8528, opened 2019-07-04
[Breaking change] remove realm

yenda:
https://notes.status.im/removing-realm

### Steps to test

- test that all features are working
- there is no persistence so after logging out,  account is lost

status: ready


jakubgs:
Interesting:
```
ld: library not found for -lTcpSockets
```
This error looks very much like what I'm getting in #7479, but there it's `React`:
```
ld: library not found for -lReact
```
I wonder if you have any idea how to solve this. It seems to be related to how Xcode links libraries, but no matter how much I messed around with those settings in Xcode I couldn't fix it, but that's probably because I'm not that familiar with iOS development.

yenda:
@jakubgs I removed some library and didn't clean up the pods because I don't care about iOS for this experiment. In this case it's missing `react-native-tcp` which I removed

jeluard:
This doesn't include the status-go replacement right?

yenda:
@jeluard no it's in-memory, it's a breaking change to iterate on

yenda:
@flexsurfer could you check that PR for iOS?

gravityblast:
@yenda did everyone agree on this decision to move the db from react to go?
It's fine for me as I said, but this requires breaking compatibility for sure and requires more review from the security point of view. So I would ping @corpetty and @oskarth and keep it as a discussion for the V1 scope call. /cc @guylouis

oskarth:
Do we have to Go DB working already? Ie it includes all state writes/reads (just latest db version)? If not, it makes zero sense to me why we’d start by removing Realm before this done.

Write to both, have flag for reading only from Go instead of Realm, and then we can deprecate Realm once we have confidence it is working. It is really straightforward.

 It required very little additional work to do this gracefully in two steps, ensures we don’t break stuff and we can get it into the code base sooner rather than later.

rachelhamlin:
I believe @yenda's main intention was to investigate how much effort this would be. But should not be merged until we have consensus.

oskarth:
Isn't all the effort in using a different database, rather than just deleting all the realm dependencies and interaction points? Sorry if I'm missing something here.

flexsurfer:
@yenda , @PombeirP says if you change deps you have to run `make nix-upgrade-npm`, just in case

yenda:
@flexsurfer only for desktop, error is just caused by the fact that I removed react as a dependency, got cocky here :D

pedropombeiro:
> @yenda , @PombeirP says if you change deps you have to run `make nix-upgrade-npm`, just in case

@flexsurfer, sorry for misleading you, after a week of vacation I forgot that I had made some improvements before merging Reproducible Android builds PR that don't require that for mobile.

PR #8542, opened 2019-07-09
Onboarding sign-in

### LINKED ISSUE #8032: [Onboarding] Update sign-in flow
### User Story
As an existing user, I want to log into my account or choose from multiple accounts, if I have them. 

### Description
_Type_: Feature

_Summary_: Update account sign-in flow according to new [onboarding designs](https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=927%3A14702). 

#### Expected behavior
The two primary sign-in screens have minor UI adjustments.

image

### LINKED ISSUE #8540: [Onboarding] Create new account flow
# Problem

Currently, if you try to create one more new account, it uses old flow

## Implementation

New flow with generating 5 accounts is used, notifications and welcome screens aren't shown

vitvly:
Fixes #8032 
Fixes #8540


vitvly:
Accounts screen:
Simulator Screen Shot - iPhone Xs - 2019-07-09 at 17 58 30

Login screen:
Simulator Screen Shot - iPhone Xs - 2019-07-09 at 22 04 24

Login processing screen:
Simulator Screen Shot - iPhone Xs - 2019-07-09 at 22 04 29

src/status_im/init/core.cljs
```diff
@@ -137,22 +137,25 @@
...
-              #(filter (fn [account]
-                         (= account-with-notification
-                            (:public-key account)))
-                       %)
-              #(sort-by :last-sign-in > %))
-            {:keys [address photo-path name]} (first (selection-fn (vals accounts)))]
-        (accounts.login/open-login cofx address photo-path name)))))
+    (cond (empty? accounts)
+          (navigation/navigate-to-cofx cofx :intro nil)
+          (= 1 (count accounts))
```
flexsurfer: why 1 ? what if 2 ?

flexsurfer (changes requested):

src/status_im/init/core.cljs
vitvly: Then we move to the `:else` clause. If i understand Figma design correctly, from now on we will show account selection in all cases when there is more than one account. But maybe this is wrong - @hesterbruikman can you please clarify? Good point however - i've removed the account selection logic for now.

hesterbruikman: I would say desired behavior is as follows:

On launch...
- Show [Password/pin entry] for last used account
- If Save account was set for the last used account - skip unlock [Password/Pin entry]
- Account selection is shown when: user taps Back from [Password/Pin entry] OR when user selects [Log out] in [Profile > Settings]


cc @andmironov @errorists

maciej: not sure If I can help anyhow

asemiankevich:
@siphiuel 

1. can't navigate back after creating an account, cant create one more (only recover)
2. please add accessability ID for [password] field
3. Save password check box can not be unchecked 
4. Endless spinner if i tap Submit with empty password 2  times in a row

image

src/status_im/init/core.cljs
vitvly: Got it, thanks, fixing.

flexsurfer (approved):

src/status_im/init/core.cljs
flexsurfer: @siphiuel so same as we had before, not changes needed here from UX pov

hesterbruikman: > 
> 
> not sure If I can help anyhow

Sorry about that! I keep making that mistake:'( Have a great day

hesterbruikman:
Here's what my eyes can catch:
@andmironov can you please check if you notice anything else?

5. Save password should be left-aligned with the grey input box
6. Full list item on account selection screen looks centered. @andmironov can you please take a look and provide the specs for this one?
Also looks like the random name is cutoff by the text frame. The ... truncation in the middle works great, but for longer names still it's like there's a white block covering the last letter for a pixel or 2.
7. Identicons should be centered, looks like they are more aligned to the top, both on Account selection and Password/PIN screens
8. Address should be truncated further on both screens: 0x8ueq...8ujj

src/status_im/init/core.cljs
andmironov: >I would say desired behavior is as follows:

looks good to me!

andmironov:
Testing the PR I've spotted a few visual inconsistencies with the mockups:

**Unlock screen**
- "Unlock" title – wrong text style (see Figma);
- Wrong layout (the margin between the title and the list) (see Figma);
- Account list item: wrong margins inside and outside, around chevron (see Figma);
- The address is displayed too long (should be 0x8ueq...8ujj)(see Figma);

**Enter password screen**
- Wrong margin between account name and account key (see Figma);
- Wrong margins around "save password" checkbox and label (see Figma);
- The address is displayed too long (should be 0x8ueq...8ujj)(see Figma);
- Wrong style of error message (see Figma);
- Wrong padding around "Access key" button (see Figma);
- Wrong loading state (when entering a password and tapping "submit" the text input disappears for a second);
- After entering a wrong password the "Submit button" gets a wrong background color forever;
- After entering wrong password for several times, the screen is forever "Processing" with a small text input (see screenshot)

photo_2019-07-10 13 30 52

_WIll add more comments re: account creation flow in a moment_

hesterbruikman:
@andmironov can you please add account creation comments here: https://github.com/status-im/status-react/pull/8234

Still checking what the best process is to update that flow (before/after illustrations) as it has some dependencies with other work. Than at least we have all necessary updates in one place.

flexsurfer:
@asemiankevich @siphiuel can we fix blockers and merge this one? and fix others in separate PR?

vitvly:
Items 1-5 fixed.
  - \# 6 - @hesterbruikman can you please provide an example of cutoff?
  - \# 7 - don't quite understand this. You mean icons are not centered inside the grey circle?
  - `Unlock` label style and margin - fixed (top and bottom set to 24)
  - `Account list item wrong margins` - don't understand which ones. @andmironov  Can you please clarify? Left margin for icon, icon-to-text, and right margin from chevron are all 16. Chevron icon size is 24 by default though, which is different from the one in Figma. This is the reason for the left-right asymmetry. However, if explicit width/height like in Figma are specified (9x16), the chevron becomes too small.
  - `Wrong margin between account name and account key` - it's set to 4, like in Figma. Should it be a different value?
  - `Wrong margins around "save password"` - fixed
  - `Wrong style of error message` - fixed
  - Wrong padding around "Access key" button - this button has the same margins/padding as Submit. There is some in Figma that causes button label to jump (like this https://monosnap.com/file/1XHGrOWvxOsSUPV4kjIZ5XfsCAInvA) , and i cannot see the proper values. Can you please specify what these should be?
  - `Disappearing text input` - cannot reproduce this.
  - `Wrong background for Submit button` - fixed
  - `screen is forever "Processing"` - cannot reproduce.

vitvly:
Changed address display throughout Onboarding so that it uses a shortened version.

asemiankevich:
it looks good to me more or less, i am not 100% sure about the margins and alignment, but anyway. Let @Serhy adjust autotests here.

asemiankevich:
@hesterbruikman @errorists would you mind to review one more time?

asemiankevich:
@siphiuel on Desktop build when i tap [Get Started] button -> I see blank screen (macOS)

image

Could you please take a look? Seems okay in nightly

UPD i was wrong, nightly is the same https://github.com/status-im/status-react/issues/8558

PR #8546, opened 2019-07-09
Upgrade firebase 5.1.1

pedropombeiro:

### Summary

Small PR to remove unrelated stuff from https://github.com/status-im/status-react/pull/8545.

### Testing notes

We should test push notifications, especially check if there's no crash on iOS (since that's what the fork we're replacing was fixing).

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats

status: ready

### DIFF
--- a/mobile_files/package.json.orig
+++ b/mobile_files/package.json.orig
@@ -40,7 +40,7 @@
     "react-native-crypto": "2.1.1",
     "react-native-dialogs": "1.0.4",
     "react-native-fetch-polyfill": "1.1.2",
-    "react-native-firebase": "git+https://github.com/status-im/react-native-firebase.git#v5.0.0-rc1-status",
+    "react-native-firebase": "5.1.1",
     "react-native-fs": "2.11.15",
     "react-native-gesture-handler": "^1.3.0",
     "react-native-http-bridge": "0.6.1",
--- a/mobile_files/yarn.lock
+++ b/mobile_files/yarn.lock
@@ -5587,9 +5587,10 @@ react-native-fetch-polyfill@1.1.2:
   resolved "https://registry.yarnpkg.com/react-native-fetch-polyfill/-/react-native-fetch-polyfill-1.1.2.tgz#256b5a0abd78cc4992f7a7cf82543da2f2124a73"
   integrity sha1-JWtaCr14zEmS96fPglQ9ovISSnM=
 
-"react-native-firebase@git+https://github.com/status-im/react-native-firebase.git#v5.0.0-rc1-status":
-  version "5.0.0-rc1"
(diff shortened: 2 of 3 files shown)


cammellos (approved):

flexsurfer (approved):

yenda (approved):

annadanchenko:
@PombeirP both on nightly and in this PR I have an issue if tap on PN received when android app was closed: https://github.com/status-im/status-react/issues/8550 

I don't think it's a blocker to test this PR as I'm getting the notification but I can't be 100% sure that opening app from PN works fine in this PR because of the mentioned issue.

pedropombeiro:
@annadanchenko is this only happening recently? It looks like the result of a bad refactoring.

annadanchenko:
> @annadanchenko is this only happening recently? It looks like the result of a bad refactoring.

it's working fine in release 0.13.0 but it was out on May 28 that is quite some time ago

annadanchenko:
@PombeirP no PNs are received if messages are sent between 0.13.0 release and this pr. not sure it's expected (in both cases: if app is on background ios/android or closed on android)

pedropombeiro:
@annadanchenko there should be no behavioral changes between current nightly and this PR in terms of notifications. So if you send a message between 0.13.0 release and latest nightly it behaves differently?

annadanchenko:
> @annadanchenko there should be no behavioral changes between current nightly and this PR in terms of notifications. So if you send a message between 0.13.0 release and latest nightly it behaves differently?

yes, it's same with nightly and 0.13.0 so not introduced here

pedropombeiro:
> yes, it's same with nightly and 0.13.0 so not introduced here

That's concerning. If it wasn't introduced by this PR, then we need to check what other changes have been made recently to the PNs code in https://github.com/status-im/status-react/commits/develop/src/status_im/notifications/core.cljs and https://github.com/status-im/status-react/commits/develop/src/status_im/notifications/background.cljs.

pedropombeiro:
Something that might explain not receiving notifications is if the SHA3 calculated by this commit was different that the one in 0.13.0: https://github.com/status-im/status-react/commit/04de022eae5c56b55b99951e65315fabb7e21e78#diff-47fede718382f29c064b23777c288041. In this commit we moved from `dependencies/web3-prototype.sha3` to `dependencies/web3-utils.sha3`. I'd expect it to return the same string, but could be a case difference or something like that. cc @yenda

pedropombeiro:
@annadanchenko if this PR isn't making things worse, I'd suggest merging it so that the Reproducible Android builds PR can become smaller (it is currently based on this branch).

PR #8549, opened 2019-07-10
Reproducible Android builds

### LINKED ISSUE #8203: Reproducible Android builds
# Problem
In order to provide verifiability in the builds we provide to users, we want to make our builds reproducible (Android first). This will also allow us to publish the Android app on [F-Droid](https://f-droid.org/en/docs/Reproducible_Builds/).

## Implementation
The recent effort of migrating to Nix (including pinning our tools, SDKs and ensuring a pure build environment) should have done most of the work required for reproducible builds (Nix takes care for instance of setting build artifacts' timestamps to a known value).

Comparing 2 builds from our CI build servers yielded a single type of differences in index.android.bundle, related to how our logging library and Leiningen interact. Leiningen normally [creates a temp file path](https://github.com/technomancy/leiningen/blob/f7e1adad6ff5137d6ea56bc429c3b620c6f84128/leiningen-core/src/leiningen/core/eval.clj#L238) that ends up in the js code, and this file path is used by Timbre to calculate a [callsite ID](https://github.com/ptaoussanis/timbre/blob/master/src/taoensso/timbre.cljx#L558) (although fixing the path is not enough because Timbre also adds a random number before hashing everything). We'll need to address this by either replacing Leiningen or modifying/replacing the Timbre logging library.

image

## Acceptance Criteria

* [x] Distinct builds from CI servers produce the exact same output binary (bitwise equality).
* [x] Distinct builds from 2 different developer machines (same host and target platform) produce the exact same output binary (bitwise equality).

## Notes
Issue for similar effort @ Briar: https://code.briarproject.org/briar/briar/issues/1273
Reproducible builds post @ F-Droid: https://f-droid.org/en/docs/Reproducible_Builds/
SOURCE_DATE_EPOCH: https://reproducible-builds.org/docs/source-date-epoch/

## Future Steps

* Modify our canary build to compare hashes with the previous build and fail if it ever changes.
* Address other platforms.

pedropombeiro:

fixes #8203

### Summary

This PR enables Reproducible Android builds by running the Android build end-to-end in a sandboxed Nix environment (which minimizes the influence of the host machine configuration on the build). To do so, I've:
- wrapped yarn dependencies in Nix using [yarn2nix](https://github.com/moretea/yarn2nix);
- wrapped Maven dependencies using a custom Nix script, for both clojars and gradle dependencies. This is a script that could be productized into a separate repo, but I haven't done so for lack of time;
- forked the logging library so that it doesn't cause the build path to leak into the JavaScript;
- patched the Go compiler in Nix so that the temp build path that it adds to the executable is controlled by Nix using an environment variable;

Documentation change PR (review please): https://github.com/status-im/status.im/pull/375

### Review notes

The bulk of the work is done in the commit named `nix: Wrap gradle, npm packages, lein prod-build and gradle assembleRelease in Nix expressions to improve reproducible builds with constant build paths`

Notable changes with this PR:

- `make release-android` no longer deploys to connected Android device, it simply generates an .apk (much like `make release-ios` or `make release-desktop`). To deploy to the phone, you can call `make android-install`;
- for mobile dev environment, `node_modules` is now copied from the Nix store. A sentinel file called `node_modules/.copied~` is added to mark the latest time we modified the folder from a Nix script. If something in `node_modules/` changes after that, the directory is refreshed from the Nix store. Anytime `mobile_files/yarn.lock` changes, we need to run `make update-npm-nix` to regenerate the yarn2nix file;
- for Android release builds, the local Maven cache is not used. Instead, we use the artifacts downloaded into the Nix store (they can be recalculated using `make update-gradle-nix` and `make update-lein-nix`); 
- `scripts/prepare-for-platform.sh` is no longer needed for mobile platforms, that logic being moved to Nix scripts;
- `prod-build-android` and `release-android` are now built by Nix;
- added `disable-githooks` makefile target so that the CI server can disable them before the build;
- upgraded com.android.tools.build:gradle to 3.4.1;
- upgraded Gradle from 4.10 to 5.1.1;
- added several trimmed down Nix expressions to give us lighter-weight access to a specialized shell (e.g. for fastlane, watchman, adb, etc).

To avoid filing up the Nix cache, we should:
- not store the release builds nor the prod-builds (avoid uploading to Nix cache, and prune them regularly on CI servers);
- regularly run `nix-store --optimize` on CI servers and Nix cache server.

I've assigned this task to @jakubgs on [Wrike](https://www.wrike.com/open.htm?id=364672126).

### Testing notes

We need to do some general testing to see if the changes haven't broken anything, but normally nothing should have changed, except for the way we build the app.

In addition, we should verify that 2 android builds of the same commit, built on 2 different CI servers produce exactly the same apk file.

#### Platforms

- Android

### Next steps

- Improve `make startdev-android-*` and other debug targets to use the Nix Maven repository, so that we have a single local cache and offline development is made possible;
status: ready


Makefile
```diff
@@ -41,90 +41,113 @@ endif
 
 # Main targets
 
+_fix-perms: SHELL := /bin/sh
+_fix-perms: ##@prepare Fix permissions so that directory can be cleaned
+	$(shell test -d node_modules && chmod -R 744 node_modules)
+	$(shell test -d node_modules.tmp && chmod -R 744 node_modules.tmp)
```
jakubgs, line 47: Why inside `$(shell ...)`?

scripts/copy-translations.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "node_modules directory is missing, aborting!"
+  exit 1
+fi
+
+# 1. copy translations to node_modules
+# 2. touch node_modules/.copied~ to avoid copying node_modules again during build
+
+cp -R "${GIT_ROOT}/translations" "${GIT_ROOT}/status-modules/"
+[ "$use_chmod" == 'chmod' ] && chmod u+w "${GIT_ROOT}/node_modules"
+cp -R "${GIT_ROOT}/status-modules" "${GIT_ROOT}/node_modules/"
```
jakubgs, line 19: Why are we copying them? Can't we use symlinks?

Makefile
```diff
@@ -41,90 +41,113 @@ endif
...
-	cp -R status-modules node_modules/status-modules && \
-	react-native run-android --variant=release
+	scripts/release-$(TARGET_OS).sh
 
 release-ios: export TARGET_OS ?= ios
 release-ios: export BUILD_ENV ?= prod
 release-ios: ##@build build release for iOS release
 	# Open XCode inside the Nix context
-	@$(MAKE) prod-build-ios && \
+	@watchman watch-del $${STATUS_REACT_HOME} && \
```
jakubgs: Wouldn't it make more sense to add `watchman-clean` target to dependency targets instead of running this by hand?

Makefile
```diff
@@ -41,90 +41,113 @@ endif
...
+	watchman watch-del $${STATUS_REACT_HOME}
 
+prod-build-android: SHELL := /bin/sh
 prod-build-android: export TARGET_OS ?= android
 prod-build-android: export BUILD_ENV ?= prod
 prod-build-android:
-	lein prod-build-android && \
-	node prepare-modules.js
+	@git clean -dxf -f ./index.$(TARGET_OS).js && \
+	_NIX_RESULT_PATH=$(shell . ~/.nix-profile/etc/profile.d/nix.sh && nix-build --argstr target-os $(TARGET_OS) --pure --no-out-link --show-trace -A targets.mobile.prod-build) && \
```
jakubgs, line 133: Can you please add some comments here? Nobody sane will know what is happening in lines like this. Also, what is the git clean for?

nix/shell.sh
```diff
@@ -23,13 +23,23 @@ fi
...
-  if [ "$platform" != 'all' ]; then
+  if [ "$platform" != 'all' ] && [ "$platform" != 'android' ] && [ "$platform" != 'ios' ]; then
     # This is a dirty workaround to the fact that 'yarn install' is an impure operation, so we need to call it from an impure shell. Hopefull we'll be able to fix this later on with something like yarn2nix
-    nix-shell --show-trace --argstr target-os ${TARGET_OS} --run "scripts/prepare-for-platform.sh $platform" || exit
+    nix-shell --show-trace --argstr target-os ${TARGET_OS} --run "scripts/prepare-for-desktop-platform.sh" || exit
   fi
+
+  entry_point='shell.nix'
+  attrset_option=''
+  if [ -n "${_NIX_ATTR}" ]; then
```
jakubgs, line 39: Please add a comment explaining what `_NIX_ATTR` is and what it does.

Makefile
```diff
@@ -41,90 +41,113 @@ endif
...
+	node prepare-modules.js && \
+	sed -i'' "s|$${STATUS_REACT_HOME}|.|g" index.$(TARGET_OS).js
 
 prod-build-desktop: export TARGET_OS ?= $(HOST_OS)
 prod-build-desktop: export BUILD_ENV ?= prod
 prod-build-desktop:
 	git clean -qdxf -f ./index.desktop.js desktop/ && \
 	lein prod-build-desktop && \
 	node prepare-modules.js
+	sed -i'' "s|$${STATUS_REACT_HOME}|.|g" index.desktop.js
```
jakubgs: Why? Comments please. Or put it in a separate Makefile target with a descriptive name and a comment and call it using `$(MAKE)`.

ci/Jenkinsfile.android
```diff
@@ -52,24 +50,30 @@ pipeline {
         }
       }
     }
+    stage('Implicit dependencies') {
+      steps {
+        /* Build implicit dependencies if needed (we run `lein deps :tree` but it's not really required, for this purpose) */
```
jakubgs: I don't get what this comment means, at all.

ci/Jenkinsfile.android
```diff
@@ -102,14 +106,12 @@ pipeline {
...
-        sh 'make watchman-clean'
-        sh 'make clean'
-      }
+      steps { sh 'make clean' }
     }
   }
   post {
     success { script { load('ci/github.groovy').notifyPR(true) } }
     failure { script { load('ci/github.groovy').notifyPR(false) } }
+    always { sh 'make _fix-perms' }
```
jakubgs, line 125: I thought you decided to drop read-only `node_modules`.

ci/Jenkinsfile.fastlane.clean
```diff
@@ -23,7 +23,7 @@ pipeline {
     stage('Prep') {
       steps { script {
         nix = load('ci/nix.groovy')
-        nix.shell('bundle install --gemfile=fastlane/Gemfile')
+        nix.shell('bundle install --gemfile=fastlane/Gemfile', attr: 'targets.mobile.fastlane.shell')
```
jakubgs: Please break up long and multi-argument lines like this:
```groovy
nix.shell(
  'bundle install --gemfile=fastlane/Gemfile',
  attr: 'targets.mobile.fastlane.shell'
)
```

ci/Jenkinsfile.android
```diff
@@ -32,9 +32,7 @@ pipeline {
     /* since we are mounting it we need to specify location */
     STATUS_RELEASE_STORE_FILE = '/home/jenkins/status-im.keystore'
     /* We use EXECUTOR_NUMBER to avoid multiple instances clashing */
-    LEIN_HOME         = "/var/tmp/lein-${EXECUTOR_NUMBER}"
-    YARN_CACHE_FOLDER = "/var/tmp/yarn-${EXECUTOR_NUMBER}"
-    GRADLE_USER_HOME  = "/var/tmp/gradle-${EXECUTOR_NUMBER}"
```
jakubgs, line 37: We are removing Yarn cache env variable because we use `yarn2nix` and not Yarn, correct? But why do we remove `GRADLE_USER_HOME`? Because Gradle no longer downloads anything?

ci/Jenkinsfile.nix-cache
```diff
@@ -35,9 +35,16 @@ pipeline {
     stage('Build') {
       steps {
         sh """
+          args="--argstr target-os android --show-trace -A targets.mobile.prod-build"
           . ~/.nix-profile/etc/profile.d/nix.sh && \
-          nix build -v --no-link && \
-          nix-shell --pure --run echo
+          nix build -v --no-link --show-trace && \
+          nix-shell --pure --run echo --show-trace && \
+          nix-build --pure --no-out-link \$args
```
jakubgs: PLEASE add some comments, why are we doing this? Also, couldn't this be a separate stage? Why are you packing up the `Build` stage so much, it's always better to split those things so you know which part fails at a glance.

ci/Jenkinsfile.nix-cache
```diff
@@ -35,9 +35,16 @@ pipeline {
...
+
+          prodBuildDrv=\$(nix-instantiate --quiet \$args) && \
+          prodBuildSrcPath=\$(nix-store -q --binding src \$prodBuildDrv) && \
+          prodBuildOutPath=\$(nix-store -q --outputs \$prodBuildDrv) && \
+          nix-store --delete \$prodBuildDrv \$prodBuildSrcPath \$prodBuildOutPath
         """
```
jakubgs: Also, don't just put everything in one `sh` block, that's awful. Use a `script {}` Groovy block and call commands separately if you can.

ci/Jenkinsfile.nix-cache
```diff
@@ -46,8 +53,14 @@ pipeline {
...
             . ~/.nix-profile/etc/profile.d/nix.sh && \
-            find /nix/store/ -mindepth 1 -maxdepth 1 -not -name '.links' -and -not -name '*.lock' | \
-              xargs nix-copy-closure -v --to ${NIX_CACHE_USER}@${NIX_CACHE_HOST}
+            find /nix/store/ -mindepth 1 -maxdepth 1 \
+              -not -name '.links' -and -not -name '*.lock' \
+              -and -not -name '*-status-react-prod-build-source' \
+              -and -not -name '*-status-react-release-android-source' \
+              -and -not -name '*-prod-build-*' \
+              -and -not -name '*-release-android' | \
+              xargs nix-copy-closure -v --to ${NIX_CACHE_USER}@${NIX_CACHE_HOST} && \
```
jakubgs: Wow, this is really getting out of control. I wonder if there's a better way of doing this.
Also, we are already doing `nix-store --delete` in the stage before, why then do we need those excludes here?

ci/android.groovy
```diff
@@ -31,18 +32,30 @@ def bundle() {
...
+        sbox: [
+          env.STATUS_RELEASE_STORE_FILE
+        ],
+        attr: "targets.mobile.android.release"
       )
     }
   }
-  sh 'find android/app/build/outputs/apk'
-  def outApk = "android/app/build/outputs/apk/${target}/app-${target}.apk"
+  sh 'find android/result/'
```
jakubgs: We can drop this, it's just some debug command.

ci/ios.groovy
```diff
@@ -20,7 +20,8 @@ def bundle() {
   nix.shell(
     plutil('CFBundleShortVersionString', utils.getVersion()) +
     plutil('CFBundleVersion', utils.genBuildNumber()) +
-    plutil('CFBundleBuildUrl', currentBuild.absoluteUrl)
+    plutil('CFBundleBuildUrl', currentBuild.absoluteUrl),
+    attr: 'targets.mobile.ios.shell'
```
jakubgs: If we are doing the `attr` explicitly, we could just make each `plutil` call it's own `nix.shell` call, since it's not that expensive now.

ci/nix.groovy
```diff
@@ -15,23 +15,107 @@ def shell(Map opts = [:], String cmd) {
...
+    nix-shell --run \'${cmd}\' ${_getNixCommandArgs(opts, true)}
   """
 }
 
+/**
+ * Arguments:
+ *  - pure - Use --pure mode with Nix for more deterministic behaviour
+ *  - keep - List of env variables to pass through to Nix build
+ *  - args - Map of arguments to provide to --argstr
+ *  - attr - Name of attribute to use with --attr flag
```
jakubgs, line 35: Update the comments with new arguments and their meaning.

ci/nix.groovy
```diff
@@ -15,23 +15,107 @@ def shell(Map opts = [:], String cmd) {
...
+      nix-build ${_getNixCommandArgs(opts, false)}
+    """
+    ).trim()
+  }
+
+private def _getNixCommandArgs(Map opts = [:], boolean isShell) {
+  def keepFlags = []
+  def entryPoint = "\'${env.WORKSPACE}/shell.nix\'"
+  if (!isShell || opts.attr != null) {
+    entryPoint = "\'${env.WORKSPACE}/default.nix\'"
```
jakubgs, line 91: This does not need to be here at all, it can be added after the `nix-build` or `nix-shell` and the `_getNixCommandArgs` call in the parent method.

ci/nix.groovy
```diff
@@ -15,23 +15,107 @@ def shell(Map opts = [:], String cmd) {
...
+
+  if (isShell) {
+    keepFlags = opts.keep.collect { var -> "--keep ${var} " }
+  } else {
+    /* Export the environment variables we want to keep into
+     * a Nix attribute set we can pass to Nix and source it from the build script */
+    def envVarsList = opts.keep.collect { var -> "${var}=\"${env[var]}\";" }
+    keepFlags = ["--arg env \'{${envVarsList.join("")}}\'"]
+
+    File envFile = File.createTempFile("nix-env", ".tmp")
```
jakubgs: Could we extract this whole thing related to creation of this tempfile into a separate method like `makeNixBuildEnvFile()` or something? This code is hard to read as it is.

ci/utils.groovy
```diff
@@ -57,12 +57,20 @@ def doGitRebase() {
...
   def number = sh(
     returnStdout: true,
-    script: "./scripts/gen_build_no.sh"
+    script: "${env.WORKSPACE}/scripts/gen_build_no.sh"
   ).trim()
   println "Build Number: ${number}"
   return number
 }
 
+def getBuildNumber() {
```
jakubgs: Why is this method defined twice? Keep one or the other.

nix/derivation.nix
```diff
@@ -0,0 +1,66 @@
...
+  prod-build-fn = pkgs.callPackage ./targets/prod-build.nix { inherit stdenv pkgs target-os nodejs localMavenRepoBuilder mkFilter; };
+  nodejs = pkgs.nodejs-10_x;
+  yarn = pkgs.yarn.override { inherit nodejs; };
+  selectedSources =
+    stdenv.lib.optional platform.targetDesktop desktop ++
+    stdenv.lib.optional platform.targetMobile mobile;
+
+  # TARGETS
+  leiningen-shell = pkgs.mkShell (shellBootstraper {
+    buildInputs = with pkgs; [ clojure leiningen maven nodejs ];
```
jakubgs, line 36: Why are you passing `buildInputs` to `shellBootstraper` if it doesn't use it as far as I can tell? Does this syntax somehow passes it to `pkgs.mkShell`?

nix/go/default.nix
```diff
@@ -0,0 +1,20 @@
...
+    # since that directory path ends up in the string table and .gnu.version_d ELF header
+    postPatch = (oldAttrs.postPatch or "") + ''
+      substituteInPlace "src/cmd/go/internal/work/action.go" --replace \
+        'tmp, err := ioutil.TempDir(os.Getenv("GOTMPDIR"), "go-build")' \
+        'var err error
+		tmp := os.Getenv("NIX_GOWORKDIR")
+		if tmp == "" {
+			tmp, err = ioutil.TempDir(os.Getenv("GOTMPDIR"), "go-build")
+		}'
+      # Disable chown tests, they fail with 'invalid argument'
```
jakubgs: You're mixing tabs and spaces my dude.

nix/derivation.nix
```diff
@@ -0,0 +1,66 @@
...
+, target-os }:
+
+let
+  platform = pkgs.callPackage ./platform.nix { inherit target-os; };
+  shellBootstraper = pkgs.callPackage ./shell-bootstrap.nix { };
+  # TODO: Try to use stdenv for iOS. The problem is with building iOS as the build is trying to pass parameters to Apple's ld that are meant for GNU's ld (e.g. -dynamiclib)
+  stdenv = pkgs.stdenvNoCC;
+  maven = pkgs.maven;
+  baseGo = pkgs.go_1_11;
+  go = pkgs.callPackage ./go { inherit baseGo; };
```
jakubgs: Maybe small comment explaining we are patching the Go compiler here?

nix/derivation.nix
```diff
@@ -0,0 +1,66 @@
...
+  buildGoPackage = pkgs.buildGoPackage.override { inherit go; };
+  desktop = pkgs.callPackage ./desktop { inherit target-os stdenv status-go pkgs nodejs; inherit (pkgs) darwin; go = baseGo; };
+  mobile = pkgs.callPackage ./mobile { inherit target-os config stdenv pkgs nodejs yarn status-go maven localMavenRepoBuilder mkFilter prod-build-fn; inherit (pkgs.xcodeenv) composeXcodeWrapper; };
+  status-go = pkgs.callPackage ./status-go { inherit target-os go buildGoPackage; inherit (mobile.ios) xcodeWrapper; androidPkgs = mobile.android.androidComposition; };
+  mkFilter = import ./tools/mkFilter.nix { inherit (stdenv) lib; };
```
jakubgs, line 24: `mkFilter` doesn't mean much by itself, a comment stating that this is a `Filter for repo files when copying to build dir` or something like that.

nix/derivation.nix
```diff
@@ -0,0 +1,66 @@
...
+  localMavenRepoBuilder = pkgs.callPackage ./tools/maven/maven-repo-builder.nix { inherit (pkgs) stdenv; };
+  prod-build-fn = pkgs.callPackage ./targets/prod-build.nix { inherit stdenv pkgs target-os nodejs localMavenRepoBuilder mkFilter; };
```
jakubgs, line 27: I have no idea what this does. Please explain. I can't even find an explanation for what `targets/prod-build.nix` does inside the file.

nix/derivation.nix
```diff
@@ -0,0 +1,66 @@
...
+    stdenv.lib.optional platform.targetMobile mobile;
+
+  # TARGETS
+  leiningen-shell = pkgs.mkShell (shellBootstraper {
+    buildInputs = with pkgs; [ clojure leiningen maven nodejs ];
+    shellHook =
+      if target-os == "android" then mobile.android.shellHook else
+      if target-os == "ios" then mobile.ios.shellHook else "";
+  });
+  watchman-shell = pkgs.mkShell {
```
jakubgs: Why does watchman need a shell? Does this just exist so we can call `make watchman-clean` without having to init a full Nix shell? Of so, please add a comment stating this.
Also, since this is not part of things necessary for building `status-react` as a whole, wouldn't it make sense to move stuff like that to something like `nix/helpers.nix` or something, to separate them from what is actually needed?

nix/mobile/android/default.nix
```diff
@@ -0,0 +1,60 @@
...
+with stdenv;
+
+let
+  platform = callPackage ../../platform.nix { inherit target-os; };
+
+  inherit (callPackage ./android-env.nix { }) androidComposition licensedAndroidEnv;
+
+  mavenAndNpmDeps = callPackage ./maven-and-npm-deps { inherit stdenv stdenvNoCC gradle bash zlib androidEnvShellHook localMavenRepoBuilder mkFilter projectNodePackage status-go; };
+  gradle = pkgs.gradleGen.gradleGen rec {
+    name = "gradle-5.1.1";
```
jakubgs: I wonder if it wouldn't make sense to extract stuff like this(specific commits, urls, versions for a tool) into it's own `nix/mobile/android/gradle.nix` file. Kinda like `android-env.nix` is.

nix/tools/mkFilter_test.nix
```diff
@@ -0,0 +1,37 @@
+{ pkgs ? import <nixpkgs> { },
+  lib ? pkgs.stdenv.lib }:
+
+# This file serves as an example of the syntax required to call a Nix function, and serves to test mkFilter.nix
+# 
+# nix-shell --show-trace nix/tools/mkFilter_test.nix
```
jakubgs: I think my version was cooler:
https://gist.github.com/jakubgs/94d6d823b802691fb7a80afb29b450ac
Because it shows how to use `nix-instantiate` to test stuff without having to call `mkShell`.

nix/mobile/android/default.nix
```diff
@@ -0,0 +1,60 @@
...
+    export ANDROID_NDK_ROOT="${androidComposition.androidsdk}/libexec/android-sdk/ndk-bundle"
+    export ANDROID_NDK_HOME="$ANDROID_NDK_ROOT"
+    export ANDROID_NDK="$ANDROID_NDK_ROOT"
+    export PATH="$ANDROID_HOME/bin:$ANDROID_HOME/tools:$ANDROID_HOME/tools/bin:$ANDROID_HOME/platform-tools:$ANDROID_HOME/build-tools:$PATH"
+  '';
+
+  # TARGETS
+  release = callPackage ./targets/release-android.nix { inherit target-os gradle androidEnvShellHook mavenAndNpmDeps mkFilter nodejs prod-build status-go zlib; };
+  generate-maven-and-npm-deps-shell = callPackage ./maven-and-npm-deps/maven/shell.nix { inherit gradle maven androidEnvShellHook projectNodePackage status-go; };
+  adb-shell = pkgs.mkShell {
```
jakubgs: Same here, another helper we don't use in the actual build, but just in some very specific `Makefile` targets, maybe something like `helpers.nix` would be good to not mix this stuff.

nix/mobile/android/default.nix
```diff
@@ -0,0 +1,60 @@
...
+in {
+  inherit androidComposition;
+
+  buildInputs = assert platform.targetAndroid; [ mavenAndNpmDeps.buildInputs openjdk gradle ];
+  shellHook =
+    androidEnvShellHook + 
+    mavenAndNpmDeps.shellHook + ''
+    $STATUS_REACT_HOME/scripts/generate-keystore.sh
+
+    $STATUS_REACT_HOME/nix/mobile/reset-node_modules.sh "${mavenAndNpmDeps.buildInputs}/project" || exit
```
jakubgs: What does this do? Comment?

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+let
+  mavenLocalRepo = callPackage ./maven { inherit localMavenRepoBuilder; stdenv = if stdenv.isLinux then stdenv else stdenvNoCC; };
+
+  jsc-filename = "jsc-android-236355.1.1";
+  react-native-deps = callPackage ./maven/reactnative-android-native-deps.nix { inherit stdenvNoCC jsc-filename; };
+
+  # fake build to pre-download deps into fixed-output derivation
+  deps = 
+    let
+      projectDir = "$NIX_BUILD_TOP/project";
```
jakubgs, line 28: I assume `NIX_BUILD_TOP` is the root of the build directory nix creates? Can we clarify that with a comment?

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+          mkdir -p ${mavenRepoDir}
+          cp -a ${mavenLocalRepo}/. ${mavenRepoDir}
+          mkdir -p ${reactNativeMavenPackageDir}
+          chmod u+w ${reactNativeMavenPackageDir}/
+          cp -a ${react-native-deps}/deps ${reactNativeDepsDir}
+          for d in `find ${reactNativeDepsDir} -mindepth 1 -maxdepth 1 -type d`; do 
+            chmod -R u+w $d
+          done
+
+          # Copy fresh node_modules
```
jakubgs: Fresh how?

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+          cp -aL --copy-contents ${projectNodePackage}/libexec/${projectNodePackage.passthru.pname}/node_modules/. ${projectDir}/node_modules/
+          chmod -R u+w ${projectDir}/node_modules
+          cp -aL --copy-contents ${projectNodePackage}/libexec/${projectNodePackage.passthru.pname}/deps/${projectNodePackage.passthru.pname}/node_modules/. ${projectDir}/node_modules/
+          chmod u+w ${projectDir}/node_modules
+
+          # Adjust permissions
+          chmod -R u+w ${projectDir}
+
+          cp -R ${projectDir}/status-modules/ ${projectDir}/node_modules/status-modules/
+          cp -R ${projectDir}/translations/ ${projectDir}/node_modules/status-modules/translations/
```
jakubgs, line 84: Why not symlink?

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+
+          runHook postUnpack
+        '';
+        patchPhase = ''
+          runHook prePatch
+
+          prevSet=$-
+          set -e
+
+          patchShebangs ${projectDir}
```
jakubgs, line 100: Comment PLZ.

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+          runHook postPatch
+        '';
+        buildPhase =
+          androidEnvShellHook +
+          status-go.shellHook + ''
+          export HOME=$NIX_BUILD_TOP
+          export REACT_NATIVE_DEPENDENCIES="${reactNativeDepsDir}"
+          export STATUS_REACT_HOME="${projectDir}"
+
+          pushd ${projectDir}/android
```
jakubgs, line 162: Please explain.

nix/mobile/android/maven-and-npm-deps/default.nix
```diff
@@ -0,0 +1,207 @@
...
+
+  shellHook = ''
+    # This avoids RN trying to download dependencies. Maybe we need to wrap this in a special RN environment derivation
+    ln -sf $STATUS_REACT_HOME/mobile_files/package.json.orig $STATUS_REACT_HOME/package.json
+    ln -sf $STATUS_REACT_HOME/mobile_files/metro.config.js $STATUS_REACT_HOME/metro.config.js
+    ln -sf $STATUS_REACT_HOME/mobile_files/yarn.lock $STATUS_REACT_HOME/yarn.lock
+
+    export STATUSREACT_NIX_MAVEN_REPO="${deps}/.m2/repository"
+  '';
+}
```
jakubgs, line 209: After reading this file I feel like we've made things far too complicated for the benefit that we get. I do not look forward to fixing any of this when it breaks.

nix/mobile/android/maven-and-npm-deps/maven/fetch-maven-deps.sh
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env bash
+
+set -Eeu
+
+# This script takes care of generating/updating the maven-inputs.txt file.
```
jakubgs, line 5: I'm not even going to try to review this monster. I assume it works and I don't want to touch it EVER.
If it does start causing trouble, I think a better approach would be to `strace` the download process and instead of hardcoding possible sources of packages just find them in the calls that do the requests(dns resolution in this case I guess).

nix/mobile/android/maven-and-npm-deps/maven/shell.nix
```diff
@@ -0,0 +1,26 @@
...
+  shellHook =
+    androidEnvShellHook +
+    status-go.shellHook +
+  ''
+    if [ -d ./node_modules ]; then
+      chmod -R u+w ./node_modules
+      rm -rf ./node_modules
+    fi
+
+    cp -aL --copy-contents ${projectNodePackage}/libexec/${projectNodePackage.passthru.pname}/node_modules/. ./node_modules/
```
jakubgs: Why not symlink since this is an `mkShell` not `mkDerivation`.

nix/mobile/android/targets/release-android.nix
```diff
@@ -0,0 +1,129 @@
...
+
+    ${if keystore-file != "" then "cp -a --no-preserve=ownership ${keystore-file} ${gradleHome}/; export STATUS_RELEASE_STORE_FILE=${gradleHome}/$(basename ${keystore-file})" else ""}
+
+    # Ensure we have the right .env file
+    cp -f $sourceRoot/${envFileName} $sourceRoot/.env
+
+    # Copy index.*.js input file
+    cp -a --no-preserve=ownership ${prod-build}/index*.js $sourceRoot/
+
+    # Copy android/ directory
```
jakubgs, line 59: I thought `builtins.path` and `mkFIlter` was already taking care of this stuff.

nix/mobile/android/targets/release-android.nix
```diff
@@ -0,0 +1,129 @@
...
+    chmod u+w $sourceRoot/node_modules/realm/android
+  '';
+  patchPhase = ''
+    prevSet=$-
+    set -e
+
+    substituteInPlace android/gradlew \
+      --replace 'exec gradle' "exec gradle -S -Dmaven.repo.local='${localMavenRepo}' --offline ${gradle-opts}"
+    substituteInPlace android/app/build.gradle \
+      --replace 'versionCode 9999' 'versionCode ${build-number}'
```
jakubgs: Oh, so we set it to `9999` earlier to avoid re-building stuff, but now in the release build we change it to the actual value. Messy, but I get it. Maybe a better solution would be to implement something in `build.gradle` to cover this, rather than do this kind of patching?

nix/mobile/android/targets/release-android.nix
```diff
@@ -0,0 +1,129 @@
...
+    set $prevSet
+  '';
+  buildPhase =
+    let
+      exportEnvVars = lib.concatStrings (
+        builtins.attrValues (
+          lib.mapAttrs (name: value:
+            "export ${name}='${value}';"
+            ) env));
+      unsetEnvVars = lib.concatStrings (
```
jakubgs: Can you explain these `lib.concatStrings` here with comment? I'm at a loss on what they do.

shell.nix
```diff
@@ -3,20 +3,15 @@
   target-os ? "all" }:
 
 let
-  projectDeps = import ./default.nix { inherit target-os pkgs nixpkgs-bootstrap; inherit (nixpkgs-bootstrap) config; };
+  project = import ./default.nix { inherit target-os pkgs nixpkgs-bootstrap; inherit (nixpkgs-bootstrap) config; };
+  projectShell = project.shell;
```
jakubgs: This variable seems pointless.

shell.nix
```diff
@@ -3,20 +3,15 @@
...
   mkShell = pkgs.mkShell.override { inherit stdenv; };
-  fastlane = pkgs.callPackage ./fastlane {
-    bundlerEnv = _: pkgs.bundlerEnv { 
-      name = "fastlane-gems";
-      gemdir = ./fastlane;
-    };
-  };
 
-in mkShell {
+in mkShell (shellBootstraper {
```
jakubgs: So `mkShell` inherits here things like `buildInputs` passed to `shellBootstrapper` or am I missing something?

nix/tools/lein/generate-nix.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# This script takes care of generating/updating the nix files in this directory.
+# For this, we start with a clean cache (in ./.m2~/repository/) and call cljsbuild
+#  to cause it to download all the artifacts. At the same time, we note them
+#  in lein-project-deps-maven-inputs.txt so that we can use that as an input
+#  to maven-inputs2nix.sh
```
jakubgs: I REALLY think one thing we need to do is extract the maven stuff into it's own repo, because this is really hard to follow. It needs to have clearly defined inputs and outputs, because I can't fully follow this.

nix/mobile/default.nix
```diff
@@ -9,19 +9,52 @@ let
...
   };
   xcodeWrapper = composeXcodeWrapper xcodewrapperArgs;
-  androidPlatform = callPackage ./android.nix { inherit config gradle; };
-  selectedSources =
-    [ status-go ] ++
-    lib.optional platform.targetAndroid androidPlatform;
+  androidPlatform = callPackage ./android { inherit config pkgs target-os mkFilter nodejs maven localMavenRepoBuilder projectNodePackage prod-build-fn; status-go = status-go.android; };
+  iosPlatform = callPackage ./ios { inherit config pkgs mkFilter xcodeWrapper projectNodePackage; status-go = status-go.ios; };
+  fastlane = callPackage ./fastlane { inherit stdenv target-os; };
+  selectedSources = [
```
jakubgs, line 15: Can you explain this variable to me?

nix/mobile/default.nix
```diff
@@ -9,19 +9,52 @@ let
...
+      fastlane
+      status-go.android
+      status-go.ios
+    ] ++
+    lib.optional platform.targetAndroid androidPlatform ++
+    lib.optional platform.targetIOS iosPlatform;
+
+  yarn2nix = import (fetchFromGitHub {
```
jakubgs: I really think stuff like that should be its own separate file. It makes it much harder to read what is going on here.

nix/mobile/default.nix
```diff
@@ -9,19 +9,52 @@ let
...
+    owner = "moretea";
+    repo = "yarn2nix";
+    rev = "3cc020e384ce2a439813adb7a0cc772a034d90bb";
+    sha256 = "0h2kzdfiw43rbiiffpqq9lkhvdv8mgzz2w29pzrxgv8d39x67vr9";
+    name = "yarn2nix-source";
+  }) { inherit pkgs nodejs yarn; };
+  projectNodePackage = yarn2nix.mkYarnPackage {
```
jakubgs: Same here, I get that this invokes `yarn2nix` on our node dependencies, but all the given parameters seem to be hardcoded aside from `projectNodePackage`. This should be it's own file with clear comments on what it does.

jakubgs (changes requested):
Nix expressions are extremely hard to read if you don't know what is the purpose of specific attributes. It is not a procedural flow that can be read from top to bottom with a logical order of steps, unlike most languages. It is a bunch of attributes(variables?) defined and used in various places.

The structure of Nix expressions is much more like the structure of an endlessly branching out tree, rather than a text with quotes and annotations referencing other texts. Without proper comments for things that are not obvious it's really tough to figure out what is for what purpose. It seems to me like most lines that do something less than trivial require a comment like:
```nix
# A is modified because of X issue and used in C and B
someDamnAttribute = callPackage ./moar/crazy/stuff { inherit all the shit }
```
Otherwise this is just madness.

I'm not finished with this PR, I'm going to re-read it tomorrow.

Makefile
pedropombeiro: because otherwise I get

```
test -d node_modules && chmod -R 744 node_modules
test -d node_modules.tmp && chmod -R 744 node_modules.tmp
Makefile:46: recipe for target '_fix-perms' failed
```

scripts/copy-translations.sh
pedropombeiro: It was like that in Roman's PR. I'll try with symlinks.

Makefile
pedropombeiro: Good point.

Makefile
pedropombeiro: Done.
The git clean already existed for desktop target, if I recall correctly, so I just extended it to the rest of the targets. It's just to make sure that whatever we get is the result of the nix-build.

nix/shell.sh
pedropombeiro: Done.

nix/mobile/default.nix
pedropombeiro: I was hoping the naming/usage would be sufficient :-) There are quite a few expressions from which we need to gather buildInputs and shellHooks. Initially we were doing that manually, but it made more sense to do something like `lib.catAttrs "buildInputs" selectedSources` to avoid code repetition between sources for buildInputs and shellHooks.

nix/mobile/default.nix
pedropombeiro: Agree.

nix/mobile/default.nix
pedropombeiro: Yeah, I'll move yarn2nix and this expression to its own file.

shell.nix
pedropombeiro: Yeah, `shellBootstraper` modifies an attribute set and passes that to `mkShell`.

nix/mobile/android/targets/release-android.nix
pedropombeiro: Well here we're copying from the prod-build-android step (or something like that, i'm on mobile now), not from the user repo.

nix/mobile/android/default.nix
pedropombeiro: Already done that.

nix/go/default.nix
pedropombeiro: I believe that's because the file we're replacing stuff in uses tabs. If we don't care I can just use spaces here.

nix/derivation.nix
pedropombeiro: Some stuff is possible to isolate to a completely separate file (although at that point there's no guarantee you're using the same version as in the full shell), others need some stuff that is parameterized here (think if you need a gradle shell), that's why I decided to leave things inside the tree structure and use attribute sets. Happy to discuss further in a call.

nix/derivation.nix
pedropombeiro: It does.

nix/derivation.nix
pedropombeiro: Maybe better if the comment is inside go/default.nix so we don't duplicate it?

ci/Jenkinsfile.nix-cache
pedropombeiro: For stuff that is already present in the machine (which won't get covered by a new delete). If we gc the build hosts, then I guess we don't need it.

nix/derivation.nix
jakubgs: No need to duplicate, just a short comment saying we patch the Go compiler, the details can be in `go/default.nix`.

Makefile
pedropombeiro: This is to get rid of the build path from the JavaScript bundle. On second thought, this was premature optimization, we can worry this per-platform as we move each platform to being reproducible. I'll remove this from non-Android targets.

ci/Jenkinsfile.android
pedropombeiro: Expanded the comment with:
> Implicit dependencies include building a patched node_modules, fetching maven dependencies, and anything else required.
           We do this before the parallel steps so we have a known starting situation.

ci/Jenkinsfile.android
pedropombeiro: I did, but I still need to copy the read-only version from Nix store into `node_modules.tmp`, which then gets moved into `node_modules`. If the build is stopped while that is happening, we'll end up with read-only `node_modules.tmp`, so better to have this to avoid having to fix things manually every now and then.

ci/Jenkinsfile.android
pedropombeiro: Correct on yarn. For Gradle, we just leave the default Gradle user home which means it'll end up in the Nix build directory. Does that make sense?

ci/Jenkinsfile.nix-cache
pedropombeiro: Good point, thanks for catching it. This is a case where I did this while focused on getting stuff to work. With this week's radical change of focus, I'm kind of rushing to get this PR reviewed so it doesn't end up collecting dust. I'd normally have taken a bit more time going over things to make sure stuff like this hadn't fallen through the cracks.

ci/nix.groovy
pedropombeiro: I thought I had already explained why it makes sense to have it in the function that returned the args, but we can discuss it in a Meet.

ci/utils.groovy
pedropombeiro: It's not defined twice, one is called `genBuildNumber` (which creates the `BUILD_NUMBER` file) and the other is `getBuildNumber` (which reads it). Maybe I can rename it to `readBuildNumber` instead.

nix/derivation.nix
pedropombeiro: It's a smaller duplication, but still a duplication. If in the future we no longer patch the Go compiler but simply build a specific version of it, we'll still need to update both the caller and callee. At this point, I'd probably prefer to call the folder `./patched-go`, because that enforces us to change both places.

nix/derivation.nix
pedropombeiro: Adding a comment
```
# Import a function that can build a prod-build target with specified node dependencies Nix expression
```

ci/Jenkinsfile.android
jakubgs: Makes sense.

ci/utils.groovy
jakubgs: Oh, I guess I'm blind, fair enough.

nix/derivation.nix
jakubgs: That makes sense!

nix/mobile/default.nix
jakubgs: Okay.

Makefile
```diff
@@ -64,58 +61,82 @@ else
...
 
 release-desktop: export TARGET_OS ?= $(HOST_OS)
 release-desktop: ##@build build release for desktop release
 	@$(MAKE) prod-build-desktop && \
-	cp -R translations status-modules/translations && \
-	cp -R status-modules node_modules/status-modules && \
-	scripts/build-desktop.sh
+	scripts/copy-translations.sh && \
+	scripts/build-desktop.sh; \
+	watchman watch-del $${STATUS_REACT_HOME}
```
jakubgs: You can just do `$(MAKE) watchman-clean`.

nix/derivation.nix
jakubgs: Fair enough, but a comment about a purpose would be nice.

nix/mobile/android/maven-and-npm-deps/default.nix
jakubgs: Oh, I see:
```
        # Patch the shebangs of the bundled modules to prevent them from
        # calling executables outside the Nix store as much as possible
```

scripts/copy-translations.sh
pedropombeiro: Hmm, looks like `--link` isn't supported in macOS

ci/Jenkinsfile.nix-cache
pedropombeiro: https://ci.status.im/job/status-react/job/nix-cache/job/macos/

nix/mobile/ios/default.nix
```diff
@@ -0,0 +1,46 @@
+{ config, stdenv, stdenvNoCC, callPackage,
+  pkgs, xcodeWrapper, mkFilter, fetchurl, nodejs, bash, zlib, procps,
+  status-go, projectNodePackage }:
+
+with stdenv;
+
+let
+  src =
+    let path = ./../../..;
```
jakubgs, line 16: If it works it works, but I'm always kinda skittish around hardcoded relative paths, but I assume that we'd have to pass some kind of `repoRoot` variable around to avoid this. Not sure if its worth it, thoughts?

pedropombeiro:
Refactored the mkShell code a bit so that we embed the shell-bootstrapper logic in an overridden mkShell function and pass that everywhere instead of relying on `pkgs.mkShell`. That way we're sure that the behavior is consistent in any shell.

nix/mobile/android/maven-and-npm-deps/default.nix
pedropombeiro: I was only going to comment stuff that is specific to our project, but I can do that, sure.

nix/mobile/android/maven-and-npm-deps/default.nix
pedropombeiro: Is `# Copy mint node_modules from Nix store` more explicit?

nix/mobile/android/maven-and-npm-deps/default.nix
pedropombeiro: Explain what, exactly?

nix/mobile/android/maven-and-npm-deps/default.nix
pedropombeiro: I personally think the risk of this breaking is relatively low, since the way we're interacting with platforms involved (Maven, Gradle) doesn't change that often. The bigger risk is that react-native/ReactAndroid changes its gradle files (I don't expect radical changes in this type of code either), but that'll be part of any future RN upgrade process.

pedropombeiro: I'm not sure what you mean by "for the benefit that we get". Do you mean reproducible builds as a whole?

nix/mobile/android/maven-and-npm-deps/maven/fetch-maven-deps.sh
pedropombeiro: I'm willing to timebox giving `strace` a shot, now that all the other problems are sorted out.

nix/mobile/android/targets/release-android.nix
pedropombeiro: Added a property to gradle.properties:
```
versionCode=9999
```

This property gets replaced in a release-android build.

nix/mobile/android/maven-and-npm-deps/default.nix
pedropombeiro: Gradle doesn't seem to like symlinks:

```
* What went wrong:
Execution failed for task ':app:bundleReleaseJsAndAssets'.
> Could not list contents of '/build/status-react-release-android-source/node_modules/status-modules/translations'. Couldn't follow symbolic link.
```

nix/mobile/android/targets/release-android.nix
pedropombeiro: I'm simplifying this code as follows:

```nix
      # Take the env attribute set and build a couple of scripts
      #  (one to export the environment variables, and another to unset them)
      exportEnvVars = concatStringsSep ";" (mapAttrsToList (name: value: "export ${name}='${value}'") env);
      unsetEnvVars = concatStringsSep ";" (mapAttrsToList (name: value: "unset ${name}") env);
```

nix/tools/mkFilter_test.nix
pedropombeiro: This file's your baby man, whatever you want.

shell.nix
pedropombeiro: Indeed, removing.

Makefile
```diff
@@ -64,58 +61,82 @@ else
...
+add-gcroots: ##@nix Add Nix GC roots to avoid status-react expressions being garbage collected
+	scripts/add-gcroots.sh
+
+clean-nix: SHELL := /bin/sh
+clean-nix: ##@nix Remove complete nix setup
+	sudo rm -rf /nix ~/.nix-profile ~/.nix-defexpr ~/.nix-channels ~/.cache/nix ~/.status .nix-gcroots
+
+update-npm-nix: SHELL := /bin/sh
+update-npm-nix: ##@nix Update node2nix expressions based on current package.json
+	nix/desktop/realm-node/generate-nix.sh
```
jakubgs, line 74: This is not gonna work with `SHELL := /bin/sh` without first doing `. ~/.nix-profile/etc/profile.d/nix.sh`:
```
 $ make update-npm-nix
nix/desktop/realm-node/generate-nix.sh
nix/desktop/realm-node/generate-nix.sh: line 41: node2nix: command not found
```
Same foes for other `update-*` targets.

nix/shell.sh
```diff
@@ -2,6 +2,12 @@
 
 #
 # This script is used by the Makefile to have an implicit nix-shell.
+# The following environment variables modify the script behavior:
+# - TARGET_OS: This attribute is passed as `target-os` to Nix, limiting the scope
+#     of the Nix expressions.
+# - _NIX_ATTR: This attribute can be used to specify an attribute set
+#     from inside the expression in `default.nix`, allowing drilling down into a specific attribute.
+# - NIX_KEEP: This variable allows specifying which env vars to keep for Nix pure shell.
```
jakubgs: Either call both `NIX_ATTR` and `NIX_KEEP` or `_NIX_ATTR` and `_NIX_KEEP`.
Truth be told `TARGET_OS` should probably be `NIX_TARGET_OS` too.

nix/tools/mkFilter_test.nix
```diff
@@ -0,0 +1,37 @@
+{ pkgs ? import <nixpkgs> { },
+  lib ? pkgs.stdenv.lib }:
+
+# This file serves as an example of the syntax required to call a Nix function, and serves to test mkFilter.nix
+# 
+# nix-instantiate --strict --json --eval nix/tools/mkFilter_test.nix
```
jakubgs: This is wrong now. Your version uses `mkShell` unnecessarily.
Look at https://gist.github.com/jakubgs/94d6d823b802691fb7a80afb29b450ac.
You can return anything from this file, you don't need `mkShell` at all.

nix/targets/prod-build.nix
```diff
@@ -0,0 +1,66 @@
+#
+# This Nix expression builds an index.*.js file for the current repository given a node modules Nix expression
+#
+
+{ stdenv, stdenvNoCC, lib, target-os, callPackage, pkgs,
+  mkFilter, clojure, leiningen, maven, nodejs, localMavenRepoBuilder }:
+
+# The Nix expression takes a second argument to specify the node dependencies
+{ projectNodePackage }:
```
jakubgs, line 9: Why is this a separate argument, and not part of the first?

jakubgs (approved):
It kinda annoys me how Nix expressions switch between `lowerCamelCase` and `list-stype-variables`.

nix/shell.sh
pedropombeiro: I'm avoiding a possible name clash with nix's own environment variables, so I'll rename NIX_KEEP

nix/mobile/ios/default.nix
pedropombeiro: Unfortunately, I'm not aware of an easy way to pass around a path like that, or to find it out. `locateDominatingFile` might be what we need, but its API is cumbersome and there are no usages in nixpkgs to draw inspiration from (maybe tellingly).

Makefile
pedropombeiro: Good catch!

nix/tools/mkFilter_test.nix
pedropombeiro: Replaced.

nix/targets/prod-build.nix
pedropombeiro: The first argument I pass when I import the expression, and the second I pass when I actually invoke it.

yenda (approved):

deps.edn
```diff
@@ -1,14 +1,16 @@
 {:paths ["components/src" "src" "react-native/src/cljsjs" "resources"]
  :deps  {org.clojure/clojure         {:mvn/version "1.10.0"} ;; Keep in sync with version from Nix
          org.clojure/clojurescript   {:mvn/version "1.10.520"}
+         com.google.javascript/closure-compiler-unshaded {:mvn/version "v20190325"}
```
flexsurfer: could you please elaborate on these two lines

pedropombeiro:
> It kinda annoys me how Nix expressions switch between `lowerCamelCase` and `list-stype-variables`.

I've been thinking about that too. In terms of readability, I kinda prefer dash-separated, but in terms of editing, I think camel-cased works out better. But I think we'll need to start moving towards camel-cased, as I think that's what the community mostly uses.

flexsurfer (approved):

deps.edn
pedropombeiro: I added these lines to make it the same as in project.clj. If it doesn't make sense I can remove them (at the time I was considering using a plugin to allow us to use the edn file in project.clj).

pedropombeiro: Actually, I'll just get rid of them, since they're not required in this PR. Thanks for pointing them out.

nix/targets/prod-build.nix
jakubgs: Makes sense. Than maybe update the comment? It's not a second argument. It's an argument to function that is returned from the Nix expression that is the file.

asemiankevich:
> In addition, we should verify that 2 android builds of the same commit, built on 2 different CI servers produce exactly the same apk file.

@jakubgs could you please assist here?

jakubgs:
@asemiankevich You can just build the same commit again by clicking "Build with Parameters" here:
https://ci.status.im/job/status-react/job/prs/job/android/job/PR-8549/

Also, this PR is too fucking big. I wish we could extract somethings out of it. Like implementing of `yarn2nix` in a separate PR.

pedropombeiro:
> Also, this PR is too fucking big. I wish we could extract somethings out of it. Like implementing of `yarn2nix` in a separate PR.

We probably could, but since we want to pivot from working on this ASAP, it's not likely that we'd get buy-in to do that.

asemiankevich:
looks good to me, also same commit built twice produces same apk, checked with 39 and 40 job https://ci.status.im/job/status-react/job/prs/job/android/job/PR-8549/

ghost:

ghost:

PR #8552, opened 2019-07-10
Wallet/integrate flat transfer

### LINKED ISSUE #1762: Wallet - Use web3 to fetch wallet transactions

yenda:
Fixes #1762

### Summary

- remove use of etherscan and subscriptions
- replace by status-go

Wallet transaction history

Tested in emulator for  `satoshi document engage inflict goddess auction rule unfair bid next buddy shy` on ropsten

### Steps to test

status: ready


yenda:
@rachelhamlin @asemiankevich for now we aimed at "good enough" transactions list, all the correct data is saved in status-go so we'll be able to settle and what informations we show and how we show it later without having to scrap the transactions already downloaded. This means that some transactions aren't shown as they were in previous versions, that is the case for erc721 token that are going to show up as incoming 0ERC20 transaction instead of previously outgoing ETH transactions, neither of which being a proper representation.

flexsurfer (approved):

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -198,6 +151,23 @@
...
 
+(re-frame/reg-fx
+ ::get-transfers
+ (fn [{:keys [chain-tokens from-block to-block]
+       :or {from-block "0"
+            to-block nil}}]
+   ;; start inbound token transaction subscriptions
+   ;; outbound token transactions are already caught in new blocks filter
+   (json-rpc/call
+    {:method "wallet_getTransfers"
```
flexsurfer, line 162: move to constants?

yenda: I don't think constants are usefull for this kind of things, also the call function only works with registered method

jeluard (approved):

asemiankevich:
@yenda would you mind to squash your commits into 1 ?

yenda:
@asemiankevich is it voluntarily separated in two, the first one starts using go service for live updates, the second one for history

asemiankevich:
@yenda builds failed after last force push ((

yenda:
@asemiankevich yes but don't worry about it, it is failing because of the last nix commit in develop, will get sorted out, but the old builds are still valid

asemiankevich:
i've checked transaction history both on mainnet and ropsten on my devices , it looks good both new and recovered account. the only thing i noticed that transactions are slowly mined (especially tokens).. But anyway lets merge this PR @yenda

yenda:
@asemiankevich yes it is slow for the first time but then it is persisted so should be fast

PR #8574, opened 2019-07-15
nix: Use build.sh for release-android make target

pedropombeiro:
I(@jakubgs) overtook this PR from Pedro and it kinda grew, ah well.
Changes:

* Adds `nix/build.sh` as a generalized way to call `nix-build` for specific targets from `Makefile`
* Adds `nix/clean.sh` as a generic way to clean build artifacts from `/nix/store`
  - All of those things should have in the name `status-react-(shell|build|source)`
* Renames `nix-clean` to `nix-purge` for clarity and adds `nix-clean` to call `nix/clean.sh`
* Rename various build artifacts from Nix to match the `status-react-(shell|build|source)-*` schema
* Refactors `ci/Jenkinsfile.nix-cache` to use `nix.shell()` and `nix.build()` methods from `ci/nix.groovy`
* Adds `link` argument to `nix.build()` to allow adding `--no-out-link` flag to builds
* Sets `TMPDIR` in `Makefile` to `/tmp` to avoid running out of space in `/run/user/${UID}`
* Adds `_NIX_PURE` env variable to control if `nix/shell.sh` creates a pure shell
* Sets name for Nix shells we create to `status-react-shell` to make cleanup easier

If you want me to extract some of these changes to a separate PR I can do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,14 @@ HOST_OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 # Defines which variables will be kept for Nix pure shell, use semicolon as divider
 export _NIX_KEEP ?= BUILD_ENV
 export NIX_CONF_DIR = $(PWD)/nix
+# We don't want to use /run/user/$UID because it runs out of space too easilly
+export TMPDIR = /tmp
 
 export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
+
+#----------------
+# Nix targets
+#----------------
 
 # WARNING: This has to be located right before the targets
 ifdef IN_NIX_SHELL
@@ -46,6 +52,29 @@ ifndef IN_NIX_SHELL
 else
 	@echo "${YELLOW}Nix shell is already active$(RESET)"
 endif
+
+nix-clean: ##@nix Remove all status-react build artifacts from /nix/store
+	nix/clean.sh
+
+nix-purge: SHELL := /bin/sh
+nix-purge: ##@nix Completely remove the complete Nix setup
+	sudo rm -rf /nix ~/.nix-profile ~/.nix-defexpr ~/.nix-channels ~/.cache/nix ~/.status .nix-gcroots
+
+nix-add-gcroots: SHELL := /bin/sh
+nix-add-gcroots: ##@nix Add Nix GC roots to avoid status-react expressions being garbage collected
+	scripts/add-nix-gcroots.sh
+
+nix-update-npm: SHELL := /bin/sh
+nix-update-npm: ##@nix Update node2nix expressions based on current package.json
+	nix/desktop/realm-node/generate-nix.sh
+
+nix-update-gradle: SHELL := /bin/sh
+nix-update-gradle: ##@nix Update maven nix expressions based on current gradle setup
+	nix/mobile/android/maven-and-npm-deps/maven/generate-nix.sh
+
+nix-update-lein: SHELL := /bin/sh
+nix-update-lein: ##@nix Update maven nix expressions based on current lein setup
+	nix/tools/lein/generate-nix.sh nix/lein
 
 #----------------
 # General targets
@@ -73,39 +102,21 @@ disable-githooks: ##@prepare Disables lein githooks
 	rm project.clj~
 
 #----------------
-# Nix targets
-#----------------
-
-nix-clean: SHELL := /bin/sh
-nix-clean: ##@nix Remove complete nix setup
-	sudo rm -rf /nix ~/.nix-profile ~/.nix-defexpr ~/.nix-channels ~/.cache/nix ~/.status .nix-gcroots
-
-nix-add-gcroots: SHELL := /bin/sh
-nix-add-gcroots: ##@nix Add Nix GC roots to avoid status-react expressions being garbage collected
-	scripts/add-nix-gcroots.sh
-
-nix-update-npm: SHELL := /bin/sh
-nix-update-npm: ##@nix Update node2nix expressions based on current package.json
-	nix/desktop/realm-node/generate-nix.sh
-
-nix-update-gradle: SHELL := /bin/sh
-nix-update-gradle: ##@nix Update maven nix expressions based on current gradle setup
-	nix/mobile/android/maven-and-npm-deps/maven/generate-nix.sh
-
-nix-update-lein: SHELL := /bin/sh
-nix-update-lein: ##@nix Update maven nix expressions based on current lein setup
-	nix/tools/lein/generate-nix.sh nix/lein
-
-#----------------
 # Release builds
 #----------------
 release: release-android release-ios ##@build build release for Android and iOS
 
-release-android: SHELL := /bin/sh
 release-android: export TARGET_OS ?= android
 release-android: export BUILD_ENV ?= prod
+release-android: export BUILD_TYPE ?= nightly
+release-android: export NDK_ABI_FILTERS ?= armeabi-v7a;arm64-v8a;x86
+release-android: export STORE_FILE ?= ~/.gradle/status-im.keystore
 release-android: ##@build build release for Android
-	scripts/release-$(TARGET_OS).sh
+	nix/build.sh targets.mobile.$(TARGET_OS).release \
+		--arg env '{NDK_ABI_FILTERS="$(NDK_ABI_FILTERS)";}' \
+		--argstr build-type $(BUILD_TYPE) \
+		--argstr keystore-file $(STORE_FILE) \
+		--option extra-sandbox-paths $(STORE_FILE)
 
 release-ios: export TARGET_OS ?= ios
 release-ios: export BUILD_ENV ?= prod
@@ -139,9 +150,9 @@ jsbundle-android: export TARGET_OS ?= android
 jsbundle-android: export BUILD_ENV ?= prod
 jsbundle-android: ##@jsbundle Compile JavaScript and Clojure into index.android.js 
 	# Call nix-build to build the 'targets.mobile.jsbundle' attribute and copy the index.android.js file to the project root
-	@git clean -dxf -f ./index.$(TARGET_OS).js && \
-	_NIX_RESULT_PATH=$(shell . ~/.nix-profile/etc/profile.d/nix.sh && nix-build --argstr target-os $(TARGET_OS) --pure --no-out-link --show-trace -A targets.mobile.jsbundle) && \
-	[ -n "$${_NIX_RESULT_PATH}" ] && cp -av $${_NIX_RESULT_PATH}/* .
+	@git clean -dxf ./index.$(TARGET_OS).js
+	nix/build.sh targets.mobile.jsbundle && \
+	mv result/index.$(TARGET_OS).js ./
 
 prod-build-ios: jsbundle-ios ##@legacy temporary legacy alias for jsbundle-ios
 	@echo "${YELLOW}This a deprecated target name, use jsbundle-ios.$(RESET)"
@@ -238,14 +249,17 @@ coverage: ##@test Run tests once in NodeJS generating coverage
 # Other
 #--------------
 react-native-desktop: export TARGET_OS ?= $(HOST_OS)
+react-native-desktop: export _NIX_PURE ?= true
 react-native-desktop: ##@other Start react native packager
 	@scripts/start-react-native.sh
 
 react-native-android: export TARGET_OS ?= android
+react-native-android: export _NIX_PURE ?= true
 react-native-android: ##@other Start react native packager for Android client
 	@scripts/start-react-native.sh
 
 react-native-ios: export TARGET_OS ?= ios
+react-native-ios: export _NIX_PURE ?= true
 react-native-ios: ##@other Start react native packager for Android client
 	@scripts/start-react-native.sh
 
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -26,6 +26,7 @@ pipeline {
     LC_ALL   = "en_US.UTF-8"
     LANGUAGE = "en_US.UTF-8"
     TARGET_OS = 'android'
+    BUILD_ENV = 'prod'
     NIX_CONF_DIR = "${env.WORKSPACE}/nix"
     FASTLANE_DISABLE_COLORS = 1
     REALM_DISABLE_ANALYTICS = 1
@@ -83,7 +84,9 @@ pipeline {
         stage('Build') { stages {
           stage('JSBundle') {
             steps {
-              script { cmn.nix.build(attr: 'targets.mobile.jsbundle') }
+              script {
+                cmn.nix.shell('make jsbundle-android', pure: false)
+              }
             }
           }
           stage('Bundle') {
--- a/ci/Jenkinsfile.nix-cache
+++ b/ci/Jenkinsfile.nix-cache
@@ -23,73 +23,81 @@ pipeline {
   }
 
   stages {
+    stage('Prep') {
+      steps { script {
+        nix = load('ci/nix.groovy')
+      } }
+    }
     stage('Setup') {
-      steps {
+      steps { script {
         sh 'scripts/setup'
-        sh '''
-          . ~/.nix-profile/etc/profile.d/nix.sh && \
-          nix-env -i openssh
-        '''
-      }
+        nix.shell('nix-env -i openssh', pure: false)
+      } }
     }
     stage('Build status-go') {
       steps { script {
         ['android', 'desktop', 'ios'].each { os ->
-          sh """
-            . ~/.nix-profile/etc/profile.d/nix.sh && \
-            nix-build --argstr target-os all -A targets.status-go.${os}.buildInputs
-          """
+          nix.build(
+            attr: "targets.status-go.${os}.buildInputs",
+            args: ['target-os': 'all'],
+            link: false
+          )
         }
       } }
     }
-    stage('Build jsbundle-android') {
-      steps {
-        // Run a Nix build to build/fetch everything that is necessary to produce a jsbundle for TARGET_OS=android (e.g. maven and node repos)
-        sh '''
-          args="--argstr target-os android --show-trace -A targets.mobile.jsbundle"
-
-          . ~/.nix-profile/etc/profile.d/nix.sh && \
-          nix-build --pure --no-out-link $args
-
-          prodBuildDrv=$(nix-instantiate --quiet $args) && \
-          prodBuildSrcPath=$(nix-store -q --binding src $prodBuildDrv) && \
-          prodBuildOutPath=$(nix-store -q --outputs $prodBuildDrv) && \
-          nix-store --delete $prodBuildDrv $prodBuildSrcPath $prodBuildOutPath
-        '''
-      }
+    stage('Build jsbundle') {
+      steps { script {
+        /* build/fetch things required to produce a js-bundle for android
+         * (e.g. maven and node repos) */
+        nix.build(
+          attr: 'targets.mobile.jsbundle',
+          args: ['target-os': 'android'],
+          pure: false,
+          link: false
+        )
+      } }
+    }
+    stage('Build android deps') {
+      steps { script {
+        /* build/fetch things required to build jsbundle and android */
+        nix.build(
+          attr: 'targets.mobile.android.buildInputs',
+          args: ['target-os': 'android'],
+          pure: false,
+          link: false
+        )
+      } }
     }
     stage('Build nix shell deps') {
-      steps {
-        // Run a Nix build to build/fetch everything that is necessary to instantiate shell.nix for TARGET_OS=all
-        sh '''
-          . ~/.nix-profile/etc/profile.d/nix.sh && \
-          nix-shell --argstr target-os all --pure --show-trace shell.nix
-        '''
-      }
+      steps { script {
+        /* build/fetch things required to instantiate shell.nix for TARGET_OS=all */
+        nix.shell("nix-build ${args.join(' ')}", pure: false)
+        nix.build(
+          attr: 'shell',
+          args: ['target-os': 'all'],
+          link: false
+        )
+      } }
     }
     stage('Upload') {
-      steps {
+      steps { script {
         sshagent(credentials: ['nix-cache-ssh']) {
-          sh """
-            . ~/.nix-profile/etc/profile.d/nix.sh && \
-            find /nix/store/ -mindepth 1 -maxdepth 1 \
-              -not -name '.links' -and -not -name '*.lock' \
-              -and -not -name '*-status-react-jsbundle-source' \
-              -and -not -name '*-status-react-release-android-source' \
-              -and -not -name '*-jsbundle-*' \
-              -and -not -name '*-release-android' | \
-              xargs nix-copy-closure -v --to ${NIX_CACHE_USER}@${NIX_CACHE_HOST}
-          """
+          nix.shell("""
+              find /nix/store/ -mindepth 1 -maxdepth 1 \
+                -not -name '.links' -and -not -name '*.lock' \
+                -and -not -name '*-status-react-*' \
+                | xargs nix-copy-closure -v --to ${NIX_CACHE_USER}@${NIX_CACHE_HOST}
+            """,
+            pure: false
+          )
         }
-      }
+      } }
     }
   }
   post {
-    always {
-      sh '''
-        . ~/.nix-profile/etc/profile.d/nix.sh && \
-        nix-store --optimize
-      '''
-   }
+    always { script {
+      nix.shell('nix-store --optimize', pure: false)
+      nix.shell('nix/clean.sh', pure: false)
+    } }
   }
 }
--- a/nix/build.sh
+++ b/nix/build.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -e
+
+# cleanup for artifacts created during builds
+function cleanup() {
+  # clear trapped signals
+  trap - EXIT ERR INT QUIT
+  # do the actual cleanup
+  ./nix/clean.sh ${nixOpts[@]}
+}
+
+trap cleanup EXIT ERR INT QUIT
+
+# build output will end up under /nix, we have to extract it
+function extractResults() {
+  local nixResultPath="$1"
+  echo "Saving build result: ${nixResultPath}"
+  mkdir -p result
+  cp -vfr ${nixResultPath}/* result/
+  chmod u+w -R result/
+}
+
+targetAttr="${1}"
+shift
+
+if [[ -z "${targetAttr}" ]]; then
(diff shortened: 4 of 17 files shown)


scripts/release-android.sh
```diff
@@ -33,6 +33,7 @@ function cleanup() {
 trap "cleanup" EXIT ERR INT QUIT
 
 [ -z "$BUILD_TYPE" ] && BUILD_TYPE='nightly'
+[ -z "$_NIX_ATTR" ] && _NIX_ATTR="targets.mobile.${TARGET_OS}.release"
```
jakubgs: `TARGET_OS` isn't always set.

jakubgs (approved):
Nice, more consistent.

jakubgs:
I fixed the `bootstrapped-shell.nix:6:1 called without required argument 'target-os'` by adding the missing `target-so` inheritance for `nix/bootstrapped-shell.nix`.

jakubgs:
That's something new:
```
 $ make release-android
...
scripts/release-android.sh: line 5: git: command not found
make: *** [Makefile:101: release-android] Error 127
```
Looks like there's an unnecessary `git` call in `scripts/release-android.sh`:
```
GIT_ROOT=$(cd "${BASH_SOURCE%/*}" && git rev-parse --show-toplevel)
```
I'll remove it.

jakubgs:
I'm starting to think that we could generalize this more. I'm gonna play with this a bit. I'm thinking of adding a `nix/build.sh` to easily call Nix targets from the `Makefile`, and also include the cleanup steps in it. 

I want to try to completely get rid of `scripts/release-android.sh` in favor of a much simpler `Makefile` target called with `nix/build.sh` as `SHELL`.

jakubgs:
I've added a commit with a `nix/build.sh` script. Tell me what you think about it.

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ -n "${BUILD_TYPE}" ]]; then
+    nixOpts+=("--argstr build-type ${BUILD_TYPE}")
+fi
+
+targetAttr="${1}"
+shift
+
+if [[ -z "${targetAttr}" ]]; then
+    echo "First argument is madatory and has to specify the Nix attribute!"
```
pedropombeiro: There are 2 indents instead of one here.

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+    exit 1
+fi
+
+set -x
+nixResultPath=$(
+    nix-build ${nixOpts[@]} -A "${targetAttr}" ${@} default.nix
```
pedropombeiro: Same here

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+)
+
+# No path is returned if Nix finds now modifications warranting a rebuild
+if [ -z "${nixResultPath}" ]; then
+    exit 0
```
pedropombeiro: And here.

nix/build.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -e
+
+# cleanup for artefacts created during builfs
```
pedropombeiro:
```suggestion
# cleanup for artifacts created during builds
```

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+function cleanup() {
+  trap - EXIT ERR INT QUIT
+
+  if [ -n "$nixResultPath" ]; then
+    echo "Deleting derivations from Nix store..."
+    releaseDrv=$(nix-instantiate --quiet $nixOpts)
```
pedropombeiro: It's going to be challenging to make the deletion logic generic, especially because I think nix-instantiate will kick off a build if necessary. Since in this case we're calling it with the same `$nixOpts` it's fine, but we could rename `releaseDrv` to `buildDrv`.

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+    if [ -n "$releaseDrv" ]; then
+      local releaseSrcPath=$(nix-store -q --binding src $releaseDrv)
+      local releaseOutPath=$(nix-store -q --outputs $releaseDrv)
+      local releaseRefs=( $(nix-store -q --references $releaseDrv) )
+      local prodBuildDrv=$(printf -- '%s\n' "${releaseRefs[@]}" | grep -e "prod-build-android.drv")
```
pedropombeiro: I guess in this area we'll just blindly look for references we don't want to keep (such as prod-build-{android,ios,desktop}), and ignore if none are present.

nix/mobile/android/default.nix
```diff
@@ -35,6 +35,10 @@ in {
     $STATUS_REACT_HOME/nix/mobile/reset-node_modules.sh "${mavenAndNpmDeps.deriv}/project" || exit
   '';
 
+  shell = mkShell {
```
pedropombeiro: I guess we don't need this shell then.

nix/shell.sh
```diff
@@ -5,10 +5,11 @@
 # The following environment variables modify the script behavior:
 # - TARGET_OS: This attribute is passed as `target-os` to Nix, limiting the scope
 #     of the Nix expressions.
-# - _NIX_ATTR: This attribute can be used to specify an attribute set
-#     from inside the expression in `default.nix`, allowing drilling down into a specific attribute.
+# - _NIX_ATTR: Used to specify an attribute set from inside the expression in `default.nix`.
+#     This allows for drilling down into a specific attribute in Nix expressions.
+# - _NIX_PURE: This variable allows for making the shell pure with the use of --pure.
+#     Take not that this makes Nix tools like `nix-build` unavailable in the shell.
```
pedropombeiro:
```suggestion
#     Take note that this makes Nix tools like `nix-build` unavailable in the shell.
```

Makefile
```diff
@@ -222,10 +227,12 @@ coverage: ##@test Run tests once in NodeJS generating coverage
 # Other
 #--------------
 react-native-desktop: export TARGET_OS ?= $(HOST_OS)
+react-native-desktop: export _NIX_PURE ?= true
 react-native-desktop: ##@other Start react native packager
 	@scripts/start-react-native.sh
 
 react-native-android: export TARGET_OS ?= android
+react-native-desktop: export _NIX_PURE ?= true
```
pedropombeiro:
```suggestion
react-native-android: export _NIX_PURE ?= true
```

Makefile
```diff
@@ -222,10 +227,12 @@ coverage: ##@test Run tests once in NodeJS generating coverage
 # Other
 #--------------
 react-native-desktop: export TARGET_OS ?= $(HOST_OS)
+react-native-desktop: export _NIX_PURE ?= true
```
pedropombeiro, line 252: The macOS build isn't currently pure, right?

nix/build.sh
```diff
@@ -0,0 +1,63 @@
...
+nixOpts=(
+  "--pure"
+  "--fallback"
+  "--no-out-link"
+  "--show-trace"
+  "--argstr target-os ${TARGET_OS}"
+)
+
+if [[ -n "${BUILD_TYPE}" ]]; then
+    nixOpts+=("--argstr build-type ${BUILD_TYPE}")
```
pedropombeiro: I think we should make this more generic. `BUILD_TYPE` probably only makes sense for the `release-*` builds.

pedropombeiro:
> I've added a commit with a `nix/build.sh` script. Tell me what you think about it.

Looks good! I still get the `no free disk space` message though. Don't you?

I've just a concern about making it more generic, but it'd be a good approach.

Makefile
pedropombeiro: hmm, this is just for react-native, so I guess it's fine.

jakubgs:
>Looks good! I still get the no free disk space message though. Don't you?

Yes I did, no idea what that's about. Will look into it.

nix/build.sh
jakubgs: k

nix/mobile/android/default.nix
jakubgs: Ah, true, forgot to remove that test.

Makefile
jakubgs: Yep.

nix/build.sh
jakubgs: That's fair.

nix/build.sh
jakubgs: Yeah, I haven't thought about that part that much, I thought you could help me of thinking of a general logic for cleaning stuff we don't want.

jakubgs:
Okay, so my understanding of what we build is that:

0. `nix-build --attr targets.mobile.android.release ${moarFlags[@]}`
1. This creates the derivation: `/nix/store/${HASH}-release-android.drv`
  - I understand that `release-android` comes from the `name` attr. passed to `stdenv.mkDerivation` in:
  https://github.com/status-im/status-react/blob/8afd9c6b03211e533a6ead84598da07dd8c50dde/nix/mobile/android/targets/release-android.nix#L13
  Not sure why the `${target-os}` part is necessary, since we know we are building Android.
2. It also creates the dependent derivation of `/nix/store/${HASH}-prod-build-android.drv`
  - This happens because `nix/mobile/android/targets/release-android.nix` is used by `nix/mobile/android/default.nix`.
3. This causes `/nix/store/${HASH}-status-react-prod-build-source` to be created.
  - This is the result of calling `builtins.path` in:
  https://github.com/status-im/status-react/blob/dc8a6809ec1f9e12e49459b8bbb4f3cf0f354d44/nix/mobile/android/targets/release-android.nix#L28
4. This is then built to create `/nix/store/${HASH}-prod-build-android` to be created.
  - This directory contains the resulting `index.android.js` which will be used to make the APK.
5. Then finally of that the build directory is created: `/nix/store/${HASH}-release-android` which will container the resulting APK file.
  - I can see it also creates `/nix/store/${HASH}-release-android.lock` for something.

Does this sound correct @PombeirP ?

jakubgs:
So the things we would want to delete in `nix/cleanup.sh` would be:

* `/nix/store/${HASH}-status-react-prod-build-source`
* `/nix/store/${HASH}-prod-build-android`
* `/nix/store/${HASH}-release-android`

Not sure if we also want to clean up the `.drv` and `.lock` files as well.

jakubgs:
I have something the kinda works:
```bash
#!/usr/bin/env bash

set -e

function getSources() {
    nix-store --query --binding src "${1}"
}

function getOutputs() {
    nix-store --query --outputs "${1}"
}

function getDrvFiles() {
    nix-store --query --deriver "${1}"
}

function getReferrers() {
    nix-store --query --referrers "${1}"
}

toDelete=(
  "prod-build-android"
  "release-android"
)

for drvName in "${toDelete[@]}"; do
    drvPaths=$(find /nix/store -type d -name "*${drvName}*")
    echo "${drvPaths}" | while read path; do
        echo " ? Checking:   ${path}"
        drv=$(getDrvFiles "${path}")
        # if drv is unknown-deriver then path is a source
        if [[ "${drv}" == "unknown-deriver" ]]; then
            drv=$(getReferrers "${path}")
            src="${path}"
        else
            src=$(getSources "${drv}")
        fi
        echo " - Derivation: ${drv}"
        echo " - Source:     ${src}"

        echo "Deleting..."
        echo nix-store --delete "${path}" "${drv}" "${src}"
    done
done
```
Which gives me:
```
 $ ./nix/cleanup.sh                                                                                        
 ? Checking:   /nix/store/kqci40czsgpqg7azzaiqiaij619qqmvs-prod-build-android
 - Derivation: /nix/store/vyifsanfk7vhniq5zjgrnxyd50jwm8p3-prod-build-android.drv
 - Source:     /nix/store/5dm5a0i89k8540v17lid17zs2cs2vmrk-status-react-prod-build-source
Deleting...
nix-store --delete /nix/store/kqci40czsgpqg7azzaiqiaij619qqmvs-prod-build-android /nix/store/vyifsanfk7vhniq5zjgrnxyd50jwm8p3-prod-build-android.drv /nix/store/5dm5a0i89k8540v17lid17zs2cs2vmrk-status-react-prod-build-source
 ? Checking:   /nix/store/k022fv7ikcn09vk3jrkaxz9baa2v8f0w-status-react-release-android-source
 - Derivation: 
 - Source:     /nix/store/k022fv7ikcn09vk3jrkaxz9baa2v8f0w-status-react-release-android-source
Deleting...
nix-store --delete /nix/store/k022fv7ikcn09vk3jrkaxz9baa2v8f0w-status-react-release-android-source  /nix/store/k022fv7ikcn09vk3jrkaxz9baa2v8f0w-status-react-release-android-source
 ? Checking:   /nix/store/5zwrzf1xy2jsna1vh48bd7nkqibpq0m4-status-react-release-android-source
 - Derivation: /nix/store/igzvdf3lwkjb6sbysf5lxqnr9l63lni3-release-android.drv
 - Source:     /nix/store/5zwrzf1xy2jsna1vh48bd7nkqibpq0m4-status-react-release-android-source
Deleting...
nix-store --delete /nix/store/5zwrzf1xy2jsna1vh48bd7nkqibpq0m4-status-react-release-android-source /nix/store/igzvdf3lwkjb6sbysf5lxqnr9l63lni3-release-android.drv /nix/store/5zwrzf1xy2jsna1vh48bd7nkqibpq0m4-status-react-release-android-source
```
It's a start.

jakubgs:
The cleanup would be even simpler if we prefixed all build with `status-react-` or something.

jakubgs:
I don't get how this one exists without a `drv`:
```
/nix/store/k022fv7ikcn09vk3jrkaxz9baa2v8f0w-status-react-release-android-source
```
I can't find it's derivation.

jakubgs:
At first I thought it would make more sense to do the `find` for `/nix/store/*.drv`, but if a source directory can exist without its derivation than we don't want to miss it.

jakubgs:
I'm not sure if I have the right idea for this clenaup script. Feel free to modify it if you have a better idea.

pedropombeiro:
Yeah, the flow you described sounds correct. Some comments:

> Not sure why the ${target-os} part is necessary, since we know we are building Android.

This was to be generic so that if we later copy that to some other platform, there is less to modify.

> The cleanup would be even simpler if we prefixed all build with status-react- or something.

The same thought crossed my mind. We could prefix all sources with `status-react-source-*` so we can find them later.

> Not sure if we also want to clean up the .drv and .lock files as well.

The `.lock` files are internal to Nix (maybe while something is being built?). I wouldn't worry about the `.drv` files, since they're small and might be useful for diagnosing differences between 2 builds.

> I don't get how this one exists without a drv

I assume that's because it is not really a derivation. You can store anything on the Nix store (e.g. with `nix-store --add <path>`), it doesn't need to have a derivation. I guess `builtins.path` just does that.

nix/cleanup.sh
```diff
@@ -0,0 +1,45 @@
...
+            drv=$(getReferrers "${path}")
+            src="${path}"
+        else
+            src=$(getSources "${drv}")
+        fi
+        echo " - Derivation: ${drv}"
+        echo " - Source:     ${src}"
+
+        echo "Deleting..."
+        echo nix-store --delete "${path}" "${drv}" "${src}"
```
pedropombeiro: We could just build up a list of paths to delete and make a single call to `nix-store --delete` at the end.

nix/build.sh
```diff
@@ -0,0 +1,53 @@
...
+}
+
+trap cleanup EXIT ERR INT QUIT
+
+# build output will end up under /nix, we have to extract it
+function extractResults() {
+  local nixResultPath="$1"
+  echo "Saving build result: ${nixResultPath}"
+  mkdir -p results
+  cp -r "${nixResultPath}/*" results/
```
pedropombeiro: I don't understand how this `results/` folder gets handled later on. Here we seem to be replicating the job of Nix if we were to not pass `--no-out-link`. Who will take care of moving things from `result/` to the expected folder?

nix/cleanup.sh
jakubgs: Yeah, I agree, I was mostly just trying to be explicit while developing.

nix/build.sh
jakubgs: Well, I haven't implemented that yet because I didn't know what you would think of this approach. We will have to change the CI groovy scripts to use the `results/` folder rather than the current output path for the APK.

And I didn't know that `--no-out-link` was used to block this behavior, that might be cleaner, thanks!

jakubgs:
I'm thinking this looks nice:
```log
...
error: build of '/nix/store/zlryys1q6wbcjwwfdj87pdlpwqnlg0sn-release-android.drv' failed
 ? Checking:   /nix/store/jjxfpxa5ns07j4gypx098lp9n74scsih-status-react-source-release-android
 - Derivation: /nix/store/zlryys1q6wbcjwwfdj87pdlpwqnlg0sn-release-android.drv
 - Source:     /nix/store/jjxfpxa5ns07j4gypx098lp9n74scsih-status-react-source-release-android
 ? Checking:   /nix/store/zwfzhkib955jagdblxqn6ndlyfn7dpbx-prod-build-android
 - Derivation: /nix/store/jl190wxnq25b9spvqpgg164nhr1w0crs-prod-build-android.drv
 - Source:     /nix/store/plsjh2p9kl8zamr6jxdiqc6s4nwg4lmf-status-react-source-prod-build
Deleting...
finding garbage collector roots...
removing stale temporary roots file '/nix/var/nix/temproots/16478'
deleting '/nix/store/zlryys1q6wbcjwwfdj87pdlpwqnlg0sn-release-android.drv'
deleting '/nix/store/jjxfpxa5ns07j4gypx098lp9n74scsih-status-react-source-release-android'
deleting '/nix/store/jl190wxnq25b9spvqpgg164nhr1w0crs-prod-build-android.drv'
deleting '/nix/store/plsjh2p9kl8zamr6jxdiqc6s4nwg4lmf-status-react-source-prod-build'
deleting '/nix/store/zwfzhkib955jagdblxqn6ndlyfn7dpbx-prod-build-android'
deleting '/nix/store/trash'
deleting unused links...
note: currently hard linking saves -0.00 MiB
5 store paths deleted, 23.49 MiB freed
make: *** [Makefile:103: release-android] Error 1
```
Not sure if I want to keep it as verbose, but it does work.

pedropombeiro:
It does look nice! However for day to day development it'll probably make the actual build error less visible, so for non-CI builds we should probably hide it.

jakubgs:
In `Jenkinsfile.nix-cache` I don't think we have to build the `jsbundle`(`prod-build`) at all.
We can prepare the `buildInputs` using this:
```bash
 $ nix-build $(nix-instantiate --quiet --strict -A targets.mobile.jsbundle.buildInputs)
/nix/store/6m95wp2y9jwyfm2bi7ip6k6crzlb7c61-clojure-1.10.0.411
/nix/store/4cr5dn1q3r3dpb7751khi8dxp8bxkhnf-leiningen-2.8.3
/nix/store/ba3pm79drryfgj7phy7g49gv1qi25x1c-nodejs-10.15.0
```
Or is there something else that it does?

pedropombeiro:
> In `Jenkinsfile.nix-cache` I don't think we have to build the `jsbundle`(`prod-build`) at all.
> We can prepare the `buildInputs` using this:
> 
> ```shell
>  $ nix-build $(nix-instantiate --quiet --strict -A targets.mobile.jsbundle.buildInputs)
> /nix/store/6m95wp2y9jwyfm2bi7ip6k6crzlb7c61-clojure-1.10.0.411
> /nix/store/4cr5dn1q3r3dpb7751khi8dxp8bxkhnf-leiningen-2.8.3
> /nix/store/ba3pm79drryfgj7phy7g49gv1qi25x1c-nodejs-10.15.0
> ```
> 
> Or is there something else that it does?

The main thing is to build and cache the `gradle-install-android-archives-and-patched-npm-modules` expression (and dependencies).

ci/Jenkinsfile.nix-cache
```diff
@@ -23,61 +23,63 @@ pipeline {
...
-      }
+    stage('Build jsbundle-android') {
+      steps { script {
+        /* build/fetch things required to produce a js-bundle for android
+         * (e.g. maven and node repos) */
+        def args = [
+          '--pure', '--show-trace', '--no-out-link',
+          '--attr targets.mobile.jsbundle-android',
+          '--argstr target-os android',
+          '--argstr target-os android',
```
PombeirP: This argument is repeated.

jakubgs:
>`gradle-install-android-archives-and-patched-npm-modules`

Christ that's a mouthful. Is there no way to split this? Do node modules have to be with gradle archives?

jakubgs:
Also, how exactly does building `targets.mobile.jsbundle`(`targets.mobile.prod-build`) build `gradle-install-android-archives-and-patched-npm-modules`, because I'm not seeing `mavenAndNpmDeps` being passed to it in here:
https://github.com/status-im/status-react/blob/0d78e71ece8a403ef0bb03f6ce652de443d433c7/nix/mobile/android/default.nix#L12-L15
So I'm confused.

jakubgs:
@PombeirP I can achieve the same thing using:
```bash
 $  nix-build \
      $(nix-instantiate --argstr target-os android --quiet -A targets.mobile.android.buildInputs) \
      $(nix-instantiate --quiet -A targets.mobile.jsbundle.buildInputs) 
/nix/store/h1xjr56mm6gjb0pwr4gxxl1d4mbvl3dm-gradle-install-android-archives-and-patched-npm-modules
/nix/store/60zh4z7yxfzf2mbkxaavmq95xylzlypa-openjdk-8u202bga
/nix/store/ddz1kxpwqvryl23ij4km3kp1qrn9ifwz-gradle-5.1.1
/nix/store/6m95wp2y9jwyfm2bi7ip6k6crzlb7c61-clojure-1.10.0.411
/nix/store/4cr5dn1q3r3dpb7751khi8dxp8bxkhnf-leiningen-2.8.3
/nix/store/ba3pm79drryfgj7phy7g49gv1qi25x1c-nodejs-10.15.0
```

jakubgs:
It looks like the `out of space` error is due to Nix using the `/run/user/${UID}` directory for builds.
Since that directory us usually a small fraction of the system RAM, for example `1.6 GB` it runs out of space because copying `node_modules` takes up `1.5 GB` out of that.
More details: https://github.com/NixOS/nix/issues/2957

Looks like this location can be changed by setting `TMPDIR` to something else that exists, like `/tmp/nix`.

jakubgs:
Ugh...
```
copying path '/nix/store/a8innbp9nfvxh4s3r72x4vb2plmlkxxh-status-go-v0.30.0-beta.0-macos' from 'https://nix-cache.status.im'...
hash mismatch importing path '/nix/store/a8innbp9nfvxh4s3r72x4vb2plmlkxxh-status-go-v0.30.0-beta.0-macos';
  wanted: sha256:1kp41rqn6lrjja6snyxhhnb6akdxrrrb66brrkp4l2njgnfrkhwc
  got:    sha256:18pzs22q7bf4x88n0s36n33h7hr44l5hhv9gzl7qz7rylv8f39ms
error: build of '/nix/store/4l2xypl442n87qylnafhm656zxq2ng3n-status-go-v0.30.0-beta.0-macos.drv' failed
```
Fortunately #8602 fixes this.

pedropombeiro:
> > `gradle-install-android-archives-and-patched-npm-modules`
> 
> Christ that's a mouthful. Is there no way to split this? Do node modules have to be with gradle archives?

Since Gradle builds RN native stuff on top of node_modules, when targetting Android we need to have a node_modules that's been run through this.

pedropombeiro:
> Also, how exactly does building `targets.mobile.jsbundle`(`targets.mobile.prod-build`) build `gradle-install-android-archives-and-patched-npm-modules`, because I'm not seeing `mavenAndNpmDeps` being passed to it in here:
> 
> https://github.com/status-im/status-react/blob/0d78e71ece8a403ef0bb03f6ce652de443d433c7/nix/mobile/android/default.nix#L12-L15
> 
> So I'm confused.

Why do you say it builds that expression? From what I see only the main shell, `release-android` and `generate-maven-and-npm-deps-shell` depend on that. It wouldn't make sense for `prod-build` to have dependencies on the `gradle react-native:installArchives` phase, since that is an Android-only operation, and the transformation of cljs to JavaScript is agnostic to that.

dmitryn (approved):

cammellos (approved):

ci/Jenkinsfile.nix-cache
```diff
@@ -23,73 +23,81 @@ pipeline {
...
   }
   post {
-    always {
-      sh '''
-        . ~/.nix-profile/etc/profile.d/nix.sh && \
-        nix-store --optimize
-      '''
-   }
+    always { script {
+      nix.shell('nix-store --optimize', pure: false)
```
pedropombeiro, line 99: @jakubgs shouldn't we pass an attrset here so that it takes less time to instantiate the shell?

jakubgs: I guess? It seems pretty fast already considering we use `target-os` set to `none`.

ci/Jenkinsfile.android
```diff
@@ -83,7 +84,9 @@ pipeline {
         stage('Build') { stages {
           stage('JSBundle') {
             steps {
-              script { cmn.nix.build(attr: 'targets.mobile.jsbundle') }
+              script {
+                cmn.nix.shell('make jsbundle-android', pure: false)
```
pedropombeiro, line 88: @jakubgs Any reason `pure` needs to be false here? We should be able to run it in a pure shell.

ci/Jenkinsfile.nix-cache
```diff
@@ -23,73 +23,81 @@ pipeline {
...
     stage('Build status-go') {
       steps { script {
         ['android', 'desktop', 'ios'].each { os ->
-          sh """
-            . ~/.nix-profile/etc/profile.d/nix.sh && \
-            nix-build --argstr target-os all -A targets.status-go.${os}.buildInputs
-          """
+          nix.build(
+            attr: "targets.status-go.${os}.buildInputs",
+            args: ['target-os': 'all'],
```
PombeirP: @jakubgs not sure I understand the reason for passing `all` for `target-os`.

ci/Jenkinsfile.nix-cache
pedropombeiro: The reasoning is that we should be as explicit as possible about the dependencies we expect for a given shell we start (no more and no less, ideally). I don't have a clear picture of what dependencies are brought in with `target-os` set to `none`, but I imagine it's still a lot more than what is needed here (e.g. git, jq, gnumake, clojure). I see 2 alternatives:
- add a `targets.nix.shell` target with no input dependencies;
- modify `shell.nix` so that there are no  `buildInputs` nor `inputsFrom` if `target-os` is set to `none` (feels less explicit though);

ci/Jenkinsfile.android
jakubgs: Good point, I'll fix it in #8652.

ci/Jenkinsfile.nix-cache
jakubgs: The second one sounds nicer. I'll do it in #8652.

PR #8581, opened 2019-07-16
allow for TARGET_OS to be not set

jakubgs:
This solves the issue that @cammellos found when running `make` in a specific combination:
```bash
 ./scripts/update-status-go.sh v0.30.0-beta.2
# changes the `status-go` version
make startdev-android-avd
# causes the building of `status-go` as expected
make shell
# rebuilds status-go again, but for linux, not android, due to empty `TARGET_OS`
```
In order to avoid this we will be starting the Nix shell using `make shell` without any `TARGET_OS` set by default. If the

### DIFF
--- a/nix/README.md
+++ b/nix/README.md
@@ -14,3 +14,11 @@ In order to access an interactive Nix shell a user should run `make shell`.
 
 The Nix shell is started in this repo via the [`nix/shell.sh`](/nix/shell.sh) script, which is a wrapper around the `nix-shell` command and is intended for use with our main [`Makefile`](/Makefile). This allows for an implicit use of `nix-shell` as the default shell in the `Makefile`.
 
+By default the shell starts without any specific target platform, if you want to change that you should export the `TARGET_OS` env variable with the right value:
+
+```bash
+TARGET_OS=android make shell
+```
+This way your shell and all other nix commands should run in a setup that is tailored towards Android development.
+
+For valid values you can check the [`nix/platform.nix`](/nix/platform.nix) file.
--- a/nix/shell.sh
+++ b/nix/shell.sh
@@ -11,6 +11,7 @@
 #
 
 GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
 NC='\033[0m'
 
 export TERM=xterm # fix for colors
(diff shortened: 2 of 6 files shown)


nix/README.md
```diff
@@ -14,3 +14,12 @@ In order to access an interactive Nix shell a user should run `make shell`.
 
 The Nix shell is started in this repo via the [`nix/shell.sh`](/nix/shell.sh) script, which is a wrapper around the `nix-shell` command and is intended for use with our main [`Makefile`](/Makefile). This allows for an implicit use of `nix-shell` as the default shell in the `Makefile`.
 
+By default the shell starts without any specific target platform, if you want to change that you should export the `TARGET_OS` env varable with the right value:
```
pedropombeiro:
```suggestion
By default the shell starts without any specific target platform, if you want to change that you should export the `TARGET_OS` env variable with the right value:
```

pedropombeiro: Maybe we could show just doing `TARGET_OS=android make shell`?

nix/shell.sh
```diff
@@ -27,38 +28,51 @@ if ! command -v "nix" >/dev/null 2>&1; then
...
-  if [ -n "${_NIX_ATTR}" ]; then
-    entry_point='default.nix'
-    attrset_option="-A ${_NIX_ATTR}"
-  fi
+shellArgs=(
+  "--show-trace"
+)
+
+targetArg=""
+if [[ -n "${TARGET_OS}" ]]; then
```
pedropombeiro, line 40: Why the double square brackets?

pedropombeiro (commented):
I'm not sure if this is the best way to address the problem, tbh. What will be the use of opening a `make shell` without specifying a `TARGET_OS`? I would prefer to force the existence of a `TARGET_OS` value and have specific `make shell-<platform>` targets. I'd probably get rid of the `all` value, since we start having some incompatibilities between targets (e.g. mobile uses yarn2nix while desktop uses plain yarn).

UPDATE: OK, I saw the reasoning in #status-core (to be able to get just the shell tools). If that's useful sure, we can do that.

yenda (approved):

nix/README.md
jakubgs: fixed

nix/shell.sh
jakubgs: http://mywiki.wooledge.org/BashFAQ/031

>To cut a long story short: test implements the old, portable syntax of the command. In almost all shells (the oldest Bourne shells are the exception), [ is a synonym for test (but requires a final argument of ]). Although all modern shells have built-in implementations of [, there usually still is an external executable of that name, e.g. /bin/[. POSIX defines a mandatory feature set for [, but almost every shell offers extensions to it. So, if you want portable code, you should be careful not to use any of those extensions.

>[[ is a new, improved version of it, and it is a keyword rather than a program. This makes it easier to use, as shown below. [[ is understood by KornShell, Zsh and BASH (as of version 2.02), but not by other POSIX shell implementations (e.g. posh, yash, or dash) or the BourneShell.

If we are using bash, `[[ ]]` is the preferred way.

jakubgs:
>I saw the reasoning in #status-core (to be able to get just the shell tools). If that's useful sure, we can do that.

@PombeirP yeah, I think that's the main reason people run `make shell` rather than a specific target for building something for specific platform. To have access to tools like `pod`, `clojure`, or `leiningen`.

nix/shell.sh
pedropombeiro: Wasn't aware of that, I was using `[[` only when absolutely needed.

pedropombeiro (approved):

PR #8582, opened 2019-07-17
Status go/integrate browser api

yenda:
### Summary

This moves the browser history data from realm db in status-react to sqlite in status-go

### Testing notes
- only the browser data are affected
- old browser data is deleted

#### Platforms
Same effects on all platforms

### Steps to test

- Open Status
- login, visit different websites/dapps
- logout
- login again, history should still be there

status: ready


STATUS_GO_VERSION
```diff
@@ -1,3 +1,3 @@
 ## DO NOT EDIT THIS FILE BY HAND. USE `scripts/update-status-go.sh <tag>` instead
 
-v0.30.0-beta.0
+develop
```
jakubgs: You shouldn't be using a branch as a `status-go` version, look above, the usage of the script is described:
```
## DO NOT EDIT THIS FILE BY HAND. USE `scripts/update-status-go.sh <tag>` instead
```
Please use a tag that will not change what it points at. Otherwise the same build will end up with different hashes.

env/dev/env/config.cljs
```diff
@@ -1,6 +1,6 @@
 (ns env.config)
 
-(def figwheel-urls {:android "ws://localhost:3449/figwheel-ws",
- :ios "ws://192.168.0.9:3449/figwheel-ws",
+(def figwheel-urls {:android "ws://10.0.2.2:3449/figwheel-ws",
```
flexsurfer: please remove this from PR

src/status_im/data_store/browser.cljs
```diff
@@ -1,26 +1,49 @@
...
+                         {}
+                         all-stored-browsers)]
+    {:db (assoc db :browser/browsers browsers)}))
+
+(re-frame/reg-fx
+ :data-store/get-browsers
+ (fn []
+   (json-rpc/call
+    {:method "browsers_getBrowsers"
+     :on-success #(re-frame/dispatch [::browsers %])})))
```
flexsurfer: why not `:browsers/initialize-browsers` ?

src/status_im/data_store/browser.cljs
```diff
@@ -1,26 +1,49 @@
...
 
 (defn save-browser-tx
   "Returns tx function for saving browser"
-  [{:keys [browser-id] :as browser}]
+  [{:keys [browser-id timestamp name dapp? history history-index]}]
   (fn [realm]
```
flexsurfer: why does it still have realm?

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -6,7 +6,15 @@
             [status-im.native-module.core :as status]
             [status-im.utils.money :as money]
             [status-im.utils.types :as types]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [cognitect.transit :as transit]))
+
+(def reader (transit/reader :json))
+
+(defn deserialize
```
flexsurfer: isn't used for now?

flexsurfer (approved):

flexsurfer:
great job!

src/status_im/ethereum/json_rpc.cljs
yenda: yeah indeed I was using it at first but it turns out some things don't work anymore

src/status_im/data_store/browser.cljs
yenda: `::initialize-browser` maybe? Since we are using fx/defn now to declare event, I think we should start to use `::` more because then we can do aliased namespaced keywords instead of synthetic ones, and those can be found more easily accross the code base by IDEs

STATUS_GO_VERSION
yenda: that's literally what I did, I used the script and the script updated this file

env/dev/env/config.cljs
yenda: yep it sneaked it somehow

src/status_im/data_store/browser.cljs
flexsurfer: true

src/status_im/data_store/browser.cljs
yenda: Good point, my idea was to keep these save-tx effects because they establish a clean separation between persistence and in memory. So status-go api doesn't use realm but it stays like that until all other save-tx effects work stop needing realm

vitvly (approved):

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -35,7 +35,10 @@
    "status_joinPublicChat" {}
    "status_chats" {}
    "status_startOneOnOneChat" {}
-   "status_removeChat" {}})
+   "status_removeChat" {}
```
jeluard, line 38: Maybe worth a clearer naming convention?

jeluard (approved):

src/status_im/ethereum/json_rpc.cljs
yenda: yes but that's on status-go side though

PR #8602, opened 2019-07-19
nix: Use status-go commit sha1 in Nix expression to allow for moving branches

pedropombeiro:

### Summary

This PR adds a SHA-1 to the status-go Nix expression so we can get rid of the infamous issue in a Nix build:
```
09:12:13    wanted: sha256:1kp41rqn6lrjja6snyxhhnb6akdxrrrb66brrkp4l2njgnfrkhwc
09:12:13    got:    sha256:18pzs22q7bf4x88n0s36n33h7hr44l5hhv9gzl7qz7rylv8f39ms
```
By adding the SHA-1 to the inputs, we ensure that the Nix expression is rebuilt even if we build the same branch after the branch has moved.

### Review notes

The `STATUS_GO_OWNER` and `STATUS_GO_SHA256` files have been deleted and the content moved to `STATUS_GO_VERSION`.

Output:

```
  fix/nix-status-go-name ● ⍟28  TARGET_OS=android make shell                                                                                                                           ✔  7516  5.95G RAM  17:42:05
Configuring default Nix shell for target 'android'...
trace: Matched version: v0.30.0-beta.0
trace: Matched commit-sha1: f2139105bfa9a82ff4f8cb2ec6b98c2149964a71
trace: Matched owner: status-im
trace: Matched src-sha256: 1nm0b1nbz4ly2fx6n2pybrfa1m9x0zzj7kkkqf7fwjxg6hjnp6y5
these derivations will be built:
  /nix/store/g2zdby5b9i4d0q2n03plnh00vadmkkcw-gomobile-20190319-167ebed.drv
  /nix/store/c4nzld4rwfwhawmwxz8fnib36vz5rnd1-status-go-v0.30.0-beta.0-f213910-android.drv
  /nix/store/7p2mg98h8661il7qs910isaaj05q753r-gradle-install-android-archives-and-patched-npm-modules.drv
building '/nix/store/g2zdby5b9i4d0q2n03plnh00vadmkkcw-gomobile-20190319-167ebed.drv'...
unpacking sources
unpacking source archive /nix/store/iwx6i49dpy29g58kk5crxg32fz5dw99w-mobile-167ebed
source root is mobile-167ebed
patching sources
applying patch /nix/store/h55sjq5gdf3hw7wi92s5y2kxdkzgx0vq-ndk-search-path.patch
patching file cmd/gomobile/env.go
applying patch /nix/store/70s2ip1a1wsq3jf5nw49nfgidx0dri5w-resolve-nix-android-sdk.patch
patching file cmd/gomobile/bind_androidapp.go
Creating /nix/store/8875yp19hgbg1dcd3pp2sfy4n9cz57df-gomobile-20190319-167ebed-dev
Copying from /nix/store/iwx6i49dpy29g58kk5crxg32fz5dw99w-mobile-167ebed
```

status: ready


nix/status-go/extract-status-go-config.nix
```diff
@@ -0,0 +1,20 @@
...
+    fieldName:
+      let lines = splitString "\n" content;
+          # Find the first line containing the field name
+          matchedLine = findFirst (line: (builtins.match "${fieldName}=.+" line) != null) null lines;
+      in if matchedLine != null
+         # If line is found, extract the field value
+         then head (builtins.match "${fieldName}=(.+)" matchedLine)
+         else throw "Could not find '${fieldName}' in ${content}";
+
+in extractStatusGoConfig
```
jakubgs: This looks really complex, why use a custom format if you could just make it a `status-go-verison.json` and load it in Nix using `lib.importJSON`?

modules/react-native-status/desktop/CMakeLists.txt
```diff
@@ -16,16 +16,30 @@ if (WIN32)
...
+  if(STATUS_GO_VERSION MATCHES ";version=([^;]+).*")
+    set(version ${CMAKE_MATCH_1})
+  else()
+    message(FATAL_ERROR "Version name missing from STATUS_GO_VERSION")
+  endif()
+  if(STATUS_GO_VERSION MATCHES ";commit-sha1=([^;]+).*")
+    set(commit ${CMAKE_MATCH_1})
+  else()
+    message(FATAL_ERROR "Commit SHA1 missing from STATUS_GO_VERSION")
+  endif()
```
jakubgs, line 35: Isn't this overcomplicated? We could turn `STATUS_GO_VERSION` into `status-go-version.json` and just ues https://github.com/sbellus/json-cmake here to load it. Would be cleaner I think.

nix/desktop/linux/linuxdeployqt/default.nix
```diff
@@ -3,17 +3,18 @@
 with pkgs;
 
 stdenv.mkDerivation rec {
```
jakubgs, line 5: Why `rec` instead of `let`?

pedropombeiro: why would a `let` be better in this case?

modules/react-native-status/desktop/CMakeLists.txt
pedropombeiro: yeah, I guess we could do this with JSON, if bringing an extra dependency to CMake is OK

pedropombeiro:
OK, converted to JSON.

nix/status-go/extract-status-go-config.nix
pedropombeiro: Done

modules/react-native-status/desktop/CMakeLists.txt
jakubgs: It's just a single file, we can even drop it directly into the repo.

scripts/update-status-go.sh
```diff
@@ -23,18 +23,21 @@ if [ $# -eq 0 ]; then
...
-cat << EOF > ${GIT_ROOT}/STATUS_GO_VERSION
-## DO NOT EDIT THIS FILE BY HAND. USE \`scripts/update-status-go.sh <tag>\` instead
-
-$STATUS_GO_VERSION
+STATUS_GO_SHA256=$(nix-prefetch-url --unpack ${repoUrl}/archive/${STATUS_GO_VERSION}.zip)
+STATUS_GO_COMMIT_SHA1=$(git ls-remote ${repoUrl} U ${STATUS_GO_VERSION} | cut -f1)
+
+cat << EOF > ${GIT_ROOT}/status-go-version.json
+{
+    "_comment": "DO NOT EDIT THIS FILE BY HAND. USE `scripts/update-status-go.sh <tag>` instead",
```
jakubgs: Don't use backticks here, you're executing a command instead of adding text.

status-go-version.json
```diff
@@ -0,0 +1,7 @@
+{
+    "_comment": "DO NOT EDIT THIS FILE BY HAND. USE  instead",
```
jakubgs: See, the command is missing due to use of backticks.

jakubgs (approved):
Nice, by using JSON now the version file can be used more easily form any other place.

nix/desktop/linux/linuxdeployqt/default.nix
jakubgs: Not sure, I just thought it was the preferred way of preparing variables for use before the actual call. I have no strong opinion on this, which is why my comment ends with a question mark =].

jakubgs:
After we merge this I will have to purge all the `status-go` entries in `/nix/store` on all CI hosts, including cache.

nix/desktop/linux/linuxdeployqt/default.nix
pedropombeiro: This is exactly why the `rec` keyword exists, so that you can reuse fields from the attribute set recursively without needing to define a `let` section.

yenda (approved):

nix/desktop/linux/linuxdeployqt/default.nix
jakubgs: Fair enough, good PR!

PR #8611, opened 2019-07-22
fix caching od status-go, improve update-status-go.sh script

jakubgs:
Fixes for several things:

* Fixes `scripts/update-status-go.sh` to accept commit SHA1 hashes, not just `revs`
* Modifies `nix/nix.conf` to not use `nix-cache-cn.status.im` by default and keep it as an option
* Exposes `status-go` targets for the 3 platforms in Nix (android, ios, desktop)
* Modifies `ci/Jenkinsfile.nix-cache` to build `status-go` explicitly for all platforms

In addition to that I've purged our Nix caches everywhere in attempt to get rid of clashing packages with mismatched hashes.

Ones of the reasons for a lot of the mismatched hashes were old versions of packages still stored in the Chines Nix cache at `nix-cache-cn.status.im` that was enabled by default and would be used during builds despite having old versions of packages. This happens because the `rsync` process between main cache and Chinese cache happens periodically.

Docs update: https://github.com/status-im/status.im/pull/380


yenda (approved):

dmitryn (approved):

jakubgs:
Docs update: https://github.com/status-im/status.im/pull/380

PR #8619, opened 2019-07-23
rename prod-build to jsbundle so it means something

jakubgs:
>There are only two hard things in Computer Science: cache invalidation and naming things.
> &mdash; <cite>Phil Karlton</cite>

I've been meaning to do this for a while. `prod-build` is far too generic, it means almost nothing. A `Makefile` or a Nix target name should indicate what is actually going to be built. If I do `make prod-build` I have no idea what that means, but if it's `make jsbundle` I have some inkling as to what it does.

It also simplifies `nix/targets/jsbundle.nix`(previously `prod-build.nix`) so it doesn't return a lambda that returns a derivation and just returns a derivation directly.

This is also a prep PR to slim down changes in #8574.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,46 +105,55 @@ release-ios: export BUILD_ENV ?= prod
 release-ios: watchman-clean ##@build build release for iOS release
 	# Open XCode inside the Nix context
 	@git clean -dxf -f target/ios && \
-	$(MAKE) prod-build-ios && \
+	$(MAKE) jsbundle-ios && \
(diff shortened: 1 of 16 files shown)


flexsurfer (approved):

dmitryn (approved):

jakubgs:
I have changed __NOTHING__ about our codebase, just the names of build targets, why the fuck do e2e tests give me 84%? Nobody knows.

jakubgs:
And now its 95%...
cannot compute

PR #8625, opened 2019-07-24
Clear password input value when navigating away from login screen

### LINKED ISSUE #8600: Password is not cleared after navigation between 'Login to account' <-> 'Unlock' views
***Type***: Bug

***Summary***: When user enters password to login to existing account but not submits it - we do not clear password field after user navigate back-and-forth between `Login to account` <-> `Unlock`/`Sign in to another account` views
Please see steps to reproduce below

#### Expected behavior
Password input always empty upon navigation to `Login to account` view
 
#### Actual behavior
Password is not cleared when user navigates between `Login to account` with entered password and `Recover account`/`Unlock` views 

Password_not_empty (1)

### Reproduction

- Log out of this account
- On Login view enter any value in password input
- Tap `Access key` / `<` back buttons
- Navigate to account Login view back and check whether password field is empty
 

### Additional Information
- Status version: Develop nightly 19th of July

vitvly:
Fixes #8600

For some reason TextInput does not properly re-draw itself when navigating away and back again. Trying to erase a symbol or type a new one starts from scratch, as if there were no symbols input or visible before. Thus the visual representation of input's value does not match the actual value stored.

A workaround: use react-navigation's global `onWillBlur` event.

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -144,18 +144,36 @@
             [text-class (dissoc options-with-style :nested?)]
             nested-text-elements)))
 
+(def text-input-refs (atom #{}))
+
 (defn text-input
   [options text]
-  [(text-input-class)
-   (merge
-    {:underline-color-android  :transparent
-     :max-font-size-multiplier max-font-size-multiplier
-     :placeholder-text-color   colors/text-gray
-     :placeholder              (i18n/label :t/type-a-message)
-     :value                    text}
-    (-> options
-        (update :style typography/get-style)
-        (update :style dissoc :line-height)))])
+  (let [input-ref (atom nil)]
+    (reagent/create-class
+      {:component-will-unmount #(when @input-ref
+                                  (swap! text-input-refs disj @input-ref))
+       :reagent-render
+       (fn [options text]
+         [(text-input-class)
+          (merge
+            {:underline-color-android  :transparent
+             :max-font-size-multiplier max-font-size-multiplier
+             :placeholder-text-color   colors/text-gray
+             :placeholder              (i18n/label :t/type-a-message)
+             :ref                      (fn [r] 
+                                         (if r
+                                           (when (nil? @input-ref)
+                                             (swap! text-input-refs conj r))
+                                           (when @input-ref
+                                             (swap! text-input-refs disj @input-ref)))
+                                         (reset! input-ref r)
+                                         (when (:ref options) 
+                                           ((:ref options) r)))
+             :value                    text}
+          (-> options
+              (dissoc :ref)
+              (update :style typography/get-style)
+              (update :style dissoc :line-height)))])})))
 
 (defn i18n-text
   [{:keys [style key]}]
--- a/src/status_im/ui/screens/routing/core.cljs
+++ b/src/status_im/ui/screens/routing/core.cljs
@@ -38,7 +38,10 @@
       (when-not modal?
         (status-bar/set-status-bar current-view-id)))
     :on-will-blur
-    (fn [] (reset! screen-focused? false))}])
+    (fn []
+      (reset! screen-focused? false)
+      (doseq [text-input @react/text-input-refs]
+        (.clear text-input)))}])
 
 (defn wrap
   "Wraps screen with main view and adds navigation-events component"


yenda (approved):

flexsurfer (approved):

asemiankevich:
@siphiuel @flexsurfer i still can reproduce it in android: https://drive.google.com/open?id=1bBhoJaNhLgeoHIdQqJRVly1VvqFn2M_C

Password is not cleared when i get back to login screen, however it is when i tap inside the field.

iOS works as expected. If i switch between Enter password and Access key -> password field is cleared.

oskarth:
What's the state of this PR @siphiuel?

flexsurfer (changes requested):
we need a more general solution, any text input should be cleared, can we clear all inputs on blur?

flexsurfer:
maybe text-input component should register itself on ref, and all registered components should be cleared on screen blur

vitvly:
@flexsurfer made the fix more generic as per your suggestions

components/src/status_im/ui/components/react.cljs
```diff
@@ -144,18 +144,33 @@
...
+          (merge
+            {:underline-color-android  :transparent
+             :max-font-size-multiplier max-font-size-multiplier
+             :placeholder-text-color   colors/text-gray
+             :placeholder              (i18n/label :t/type-a-message)
+             :ref                      (fn [r] 
+                                         (when r
+                                           (swap! text-input-refs conj r)
+                                           (reset! input-ref r)
+                                           (when (:ref options)
```
flexsurfer: maybe better to call ref anyway even if it's nil?

src/status_im/ui/screens/routing/core.cljs
```diff
@@ -38,7 +38,10 @@
       (when-not modal?
         (status-bar/set-status-bar current-view-id)))
     :on-will-blur
-    (fn [] (reset! screen-focused? false))}])
+    (fn []
+      (reset! screen-focused? false)
+      (doseq [text-input @react/text-input-refs]
+        (.clear text-input)))}])
```
flexsurfer, line 44: if ref is nil, is it safe to call`clear` ? a mean maybe we have to remove nil refs from list?

components/src/status_im/ui/components/react.cljs
```diff
@@ -144,18 +144,33 @@
...
-     :placeholder-text-color   colors/text-gray
-     :placeholder              (i18n/label :t/type-a-message)
-     :value                    text}
-    (-> options
-        (update :style typography/get-style)
-        (update :style dissoc :line-height)))])
+  (let [input-ref (atom nil)]
+    (reagent/create-class
+      {:component-will-unmount #(when @input-ref
+                                  (swap! text-input-refs disj @input-ref))
```
flexsurfer, line 154: better to remove on ref nil , because it's not always unmounted

src/status_im/ui/screens/routing/core.cljs
vitvly: text-input makes sure that only non-nil refs are added to `text-input-refs`.

components/src/status_im/ui/components/react.cljs
vitvly: Why would we want that?

components/src/status_im/ui/components/react.cljs
vitvly: You mean both in `component-will-unmount` and when ref is nil?

vitvly: At least for password input I noticed that it is not unmounted when navigating between accounts list/login screen, but it gets unmounted after login, must be due to react-navigation.

components/src/status_im/ui/components/react.cljs
flexsurfer: we don't know, maybe someone handle the nil , and logic will be broken. we just have to keep ref function behaviour

components/src/status_im/ui/components/react.cljs
flexsurfer: i mean remove and add in ref function, if nil remove overwise add 
```
(if r 
  (when (nil? @input-ref ) 
    (swap! text-input-refs conj r))
  (when @input-ref
    (swap! text-input-refs disj @input-ref)))
(reset! input-ref r)
```

components/src/status_im/ui/components/react.cljs
vitvly: Right, i see now, fixed.

components/src/status_im/ui/components/react.cljs
flexsurfer: @siphiuel wdyt about this ^

vitvly: Yes, agree with that, pushed a fix.

flexsurfer: so we don't need :component-will-unmount anymore then? we `disj` when ref is nil

vitvly: It seems that we still do. Here's a log excerpt after login:
```
08-21 15:40:20.975 26177 26244 I ReactNativeJS: INFO [status-im.ui.components.react:155] - # disj on unmount
08-21 15:40:20.977 26177 26244 I ReactNativeJS: INFO [status-im.ui.components.react:171] - # disj on ref nil
08-21 15:40:20.999 26177 26244 I ReactNativeJS: INFO [status-im.ui.components.react:168] - #conj on ref not nil
```
Actually, with this logic we will still have a dead ref in` text-input-refs`, because :ref callback with nil is always invoked before the one with the non-nil value.

vitvly: So maybe it makes sense to only remove on unmount, what do you think?

flexsurfer: the log shows we don't need it, because we have it in ref why do you want to `disj` twice the same object?

vitvly: The problem is that after `disj` (twice ) we `conj` it again, although it's not visible anymore. I am confused as to why we want that to happen if it's about to unmount (or already unmounted) anyway.

vitvly: disj twice is wrong, yes

flexsurfer (approved):

PR #8640, opened 2019-07-25
[fix #8638] issues with wallet transaction list

### LINKED ISSUE #8638: Several transactions are shown for outgoing token transaction
### Description

*Type*: Bug

*Summary*: mentioned in https://github.com/status-im/status-react/pull/8552#issuecomment-511362722.
Several issues are introduced by  https://github.com/status-im/status-react/pull/8552:

### 1. If you send tokens from your account there are two transactions instead of one.
The first transaction is without amount (`... HND`) in "Pending" state, the second transaction is actually what is expected:
hnd
These transactions are not updated for me inside transaction details until manual reopening app. So, in ~20 min I still see the same, and no confirmations if I tap on `1 HND`  transaction.
On receiver side 125 confirmations for this transaction.
**All fine after relogin.**

### 2. Instead of 1 transaction after buying sticker pack 2 are shown, second is "0 ERC20"
**Steps:**
- switch to Ropsten
- make sure that have ETH and STT on Ropsten
- buy sticker pack

Monosnap 2019-07-25 15-19-06

### 3. Transactions in tokens are doubled if you send two transactions (one in tokens, one in ETH) in a row
**Steps:**
- send a transaction in tokens
- send a transaction in ETH
- check transaction history.
Transaction in tokens appears twice in history (in green initial transaction, in red - what appears after sending ETH)
dsda
**All fine after relogin.**

### Additional Information

- Status version: nightly 25/07/2019

yenda:
fix #8638 

does not include the `2.` because it isn't a bug but `improvement` as all transfers are shown to the user now, next step is to improve UX but at least it properly shows all the transfers that occurs within a transaction to the user. ERC20 is shown when the token name is unknown to Status.

### Summary

pending transaction was using hash so mined transaction was using it's id
to have a unique id, the fix adds a rule when the existing transaction by hash
is in :pending state, the mined transaction overwrites it

### Steps to test

- send a transaction
- check history, transaction should be in pending state
- after transaction has been added to a block, transaction turns into sent transaction
- while transaction is pending, if looking at transaction details, confirmations start incrementing as soon as the transaction has been mined

status: ready

### DIFF
--- a/src/status_im/ethereum/transactions/core.cljs
+++ b/src/status_im/ethereum/transactions/core.cljs
@@ -124,10 +124,15 @@
         transfer-by-id   (get-in db [:wallet :transaction id])
         unique-id (when-not (or transfer-by-id
                                 (= transfer transfer-by-hash))
-                    (if transfer-by-hash id hash))]
+                    (if (and transfer-by-hash
+                             (not (= :pending
+                                     (:type transfer-by-hash))))
+                      id
+                      hash))]
     (when unique-id
       (fx/merge cofx
-                {:db (assoc-in db [:wallet :transactions unique-id] transfer)}
+                {:db (assoc-in db [:wallet :transactions unique-id]
+                               (assoc transfer :hash unique-id))}
                 (check-transaction transfer)))))
 
 (fx/defn new-transfers


flexsurfer (approved):

churik:
Tested on Mainnet and on Ropsten with incoming and outgoing transactions (also checked transaction details, open on ethercan.io)
Also checked that transaction history is fetched when recover account with old transactions.

churik (approved):

PR #8644, opened 2019-07-26
Add datasync,v1messages & disable discovery topic

cammellos:
In preparation for v1 this commits adds a few options so we can get
start debugging the protocol for v1.
This options are:

1) Datasync: If enabled it will send datasync messages
2) V1Messages: If enabled it will send v1 messages (just adding a
signature to the message)
3) Disable discovery topic: If enabled it will stop listening/publishing
on the discovery topic. You will be able to receive messages only from
clients who have this enabled as well.

If any of this option is on, it will only be compatitle with builds >=
this one. A logout is required for any change to take effect.

All this options will be removed before v1, they are there just to make
it easier for us to test and find potential issues.

### Testing

No testing of the new features is necessary, only light regression testing of messages.
If you want to try enabling the options and get a "v1 experience":
1) Enable device-to-device, enable datasync, enable v1messages, enable disable-discovery-topic, logout/login. You will only be able to communicate with devices that have same settings, running the same build.
One-to-one/group chat messages that have datasync enabled are not confirmed, as that feature is not implemented yet.

status: ready


yenda (approved):

vkjr (approved):

annadanchenko:
>  You will only be able to communicate with devices that have same settings, running the same build.

@cammellos   is it about communication in public chat? If so, I can see messages from device with all new settings on using device where these new settings are not enabled

rachelhamlin:
@annadanchenko conversation in #test-team leads me to believe these data sync changes should be tested for bandwidth & message reliability as well, even though this is merged. From Oskar:

> I don't know what the exact test plans are called, but things that impact lack of discovery topic and data sync/reliability, so that'd be stuff like:
> 
> - 1:1 for old/new contact and various configurations
> - check bandwidth of normal chatting and for 1:1 specifically
> - general messaging stability
> - possible bonus: use bad mailserver and check if 1:1 messages still arrive after going offline for a bit and both online (maybe wait until confirmations PR)

oskarth:
For the record, it was still right to merge this as it's under a flag and it passed reasonable acceptance criteria. What's missing is a follow up task. How that fits into QA release hardening roadmap seems up to the QA team, but it'd be good to have it (and results) documented somewhere.

annadanchenko:
discussed tests for Bandwidth consumption with @cammellos 
test results: https://docs.google.com/spreadsheets/d/13kffxZaPnvULoy5Qh5sZSI2551KusCLJWUcdhKyobkE/edit#gid=0

PR #8652, opened 2019-07-26
use nix shell for nix-update-* targets, fix nix-cache

jakubgs:
Changes:

* Remove `SHELL := /bin/sh` from definition of Makefile targets that update nix packages
* Check for Nix shell in those scripts and make `make shell` required for them
* Remove unnecessary command from `ci/Jenkinsfile.nix-cache`
* Add installation of `node2nix` if missing in `nix/desktop/realm-node/generate-nix.sh`
* Adjust the `find` for `nix-copy-closure` to use `-type d` which means we can't hit `*.lock`
* Changed `shell.nix` to load only most bare bones shell when `TARGET_OS=none`
* Changed `ci/Jenkinsfile.nix-cache` to use `TARGET_OS=none` by default.
* Made `make jsbundle-android` run in a pure shell in `ci/Jenkinsfile.nix-cache`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,19 +60,19 @@ nix-purge: SHELL := /bin/sh
 nix-purge: ##@nix Completely remove the complete Nix setup
 	sudo rm -rf /nix ~/.nix-profile ~/.nix-defexpr ~/.nix-channels ~/.cache/nix ~/.status .nix-gcroots
 
-nix-add-gcroots: SHELL := /bin/sh
+nix-add-gcroots: export TARGET_OS := none
 nix-add-gcroots: ##@nix Add Nix GC roots to avoid status-react expressions being garbage collected
 	scripts/add-nix-gcroots.sh
 
-nix-update-npm: SHELL := /bin/sh
+nix-update-npm: export TARGET_OS := none
 nix-update-npm: ##@nix Update node2nix expressions based on current package.json
 	nix/desktop/realm-node/generate-nix.sh
 
-nix-update-gradle: SHELL := /bin/sh
+nix-update-gradle: export TARGET_OS := android
 nix-update-gradle: ##@nix Update maven nix expressions based on current gradle setup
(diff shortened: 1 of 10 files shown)


jakubgs:
```
invalid output dir: mkdir /private/var/folders/sn: permission denied
```
Ugh, I need to figure out what this one is about.

Makefile
```diff
@@ -60,19 +60,15 @@ nix-purge: SHELL := /bin/sh
 nix-purge: ##@nix Completely remove the complete Nix setup
 	sudo rm -rf /nix ~/.nix-profile ~/.nix-defexpr ~/.nix-channels ~/.cache/nix ~/.status .nix-gcroots
 
-nix-add-gcroots: SHELL := /bin/sh
 nix-add-gcroots: ##@nix Add Nix GC roots to avoid status-react expressions being garbage collected
 	scripts/add-nix-gcroots.sh
 
-nix-update-npm: SHELL := /bin/sh
```
pedropombeiro, line 67: Should we add a specific Nix shell target for these Make targets so we don't force a ton of stuff to be downloaded/built?

pedropombeiro (approved):
Looks good, just have a minor concern about instantiating a heavy shell environment when the actual required dependencies are tiny.

Makefile
jakubgs: You're right.

jakubgs:
I've included changes from @PombeirP [comments in #8574](ttps://github.com/status-im/status-react/pull/8574#discussion_r308587282):

* Changed `shell.nix` to load only most bare bones shell when `TARGET_OS=none`
* Changed `ci/Jenkinsfile.nix-cache` to use `TARGET_OS=none` by default.
* Made `make jsbundle-android` run in a pure shell in `ci/Jenkinsfile.nix-cache`

pedropombeiro (approved):

flexsurfer (approved):

PR #8668, opened 2019-07-29
nix: use sha1 for status-go output filename

jakubgs:
This fixes issue with branch names containing slashes.

### DIFF
--- a/nix/status-go/default.nix
+++ b/nix/status-go/default.nix
@@ -16,13 +16,14 @@ let
   version = versionJSON.version;
   sha256 = versionJSON.src-sha256;
   rev = versionJSON.commit-sha1;
+  shortRev = strings.substring 0 7 rev;
   goPackagePath = "github.com/${owner}/${repo}";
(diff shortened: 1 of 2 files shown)


cammellos (approved):

nix/status-go/default.nix
```diff
@@ -22,7 +22,7 @@ let
   mobileConfigs = {
     android = {
       name = "android";
-      outputFileName = "status-go-${version}.aar";
+      outputFileName = "status-go-${strings.substring 0 7 rev}.aar";
```
pedropombeiro: We're already calculating a shortened SHA1 above in https://github.com/status-im/status-react/blob/develop/nix/status-go/default.nix#L20, maybe we should create a variable and share the value?

pedropombeiro (approved):

nix/status-go/default.nix
pedropombeiro: Decided to save you time and force-push that change.

Serhy:
All failed tests are due to known issue https://github.com/status-im/status-react/issues/8645

PR #8674, opened 2019-07-30
[#8673] Temporay hide commands from chat and extensions

### LINKED ISSUE #8673: Temporay hide commands from chat and extensions
Temporay hide commands from chat and extensions

flexsurfer:
fixes #8673


churik:
Critical tests for commands are disabled in https://github.com/status-im/status-react/pull/8601/commits/2dafdb4eecc587d6800981eb0e4cd93c5064d1cc

src/status_im/ui/screens/chat/input/input.cljs
```diff
@@ -204,7 +204,7 @@
         (when input-text-empty?
           [stickers/button show-stickers?])
         (if input-text-empty?
-          [commands-button]
+          [react/view];;TODO temporay hide for v1 code freeze [commands-button]
```
yenda: could it use the config/extensions-enabled? flag as well here?

yenda (approved):

src/status_im/ui/screens/chat/input/input.cljs
flexsurfer: no, it's not related to extensions

flexsurfer:
after discussion with @Serhy also hided send transaction from profile, and send to contracts request from contacts in wallet

churik:
@flexsurfer if we plan to hide all places, "Send transaction" in other user profile should also be hidden.

churik:
ah, OK, you did it :)

dmitryn (approved):

churik:
@flexsurfer checked:
- no extensions in profile
- if send extension from 0.13 - unhandled command is shown
- no commands in 1-1 chat
- no "Send transaction" in profile
- no send to contacts or request from contacts in wallet

but:
- I still can type `/request ETH 1` in 1-1 chat and this will be accepted as transaction request.

@flexsurfer is it expected?

flexsurfer:
@churik not expected, fixed

churik:
@flexsurfer builds are failed (not your PR is the reason obviously, but can't test)
Will ping someone in core-infra

churik (approved):

hesterbruikman:
>     * no send to contacts or request from contacts in wallet
@Serhy @flexsurfer can you elaborate on this one? Is it intended to be temporary or permanent?

I ask because @guylouis made a good point that for transactions from Wallet we don't need Proof of address. We do need to make sure you can only select people that have added you, rather than all people you added to _your_ contact list.

flexsurfer:
it's temporary, but why we don't need Proof of address?

hesterbruikman:
> it's temporary, but why we don't need Proof of address?

Not 100% so tell me if I'm wrong. As I understand we need proof of address when sending in chat because the contact is expecting an address tied to the profile of the ongoing chat. When sending a transaction from wallet, the sender address has nothing to do with the contact itself. It's just a way to enter an address completely independent from chat.

We would probably need proof of address if we would show a transaction send to a contact from wallet in chat. So if I select your address from my contact list in Wallet, send a transaction and that transaction opens a chat with you.

guylouis:
I think it means that proof of ownership is established not only when we send "/send" commands in chat - which is what we had mostly discussed -  but also at the initial moment where A adds B as a contact, B provides his wallet adress and needs to prove its his too. 
We had not discussed this explicitly in our call if I remember well, but it makes sense !

PR #8680, opened 2019-07-31
integrate status-go initKeystore native call

yenda:
### Summary

Uses initKeystore native binding when starting the app instead of startNode, as having a random node just to use some keystore features is not necessary anymore.

### Steps to test

This impact startup, account creation, account restoration and account login. If one can use an account after login in it is good.

status: ready


yenda:
@vkjr can you check the desktop integration and see if I haven't done any mistake please

flexsurfer (approved):
Could you please elaborate what this method does on status-do side?

yenda:
@flexsurfer it initialize the keystore so that you can create accounts without having to start a node. You can see it as a lightweight startNode which doesn't  require a whole dummy config just to create an account

PR #8519, opened 2019-07-03
Fixes #8136 [Onboarding] Recovery

pacamara:
Fixes #8136 [Onboarding] Recovery

### Summary
* Needs supporting PR https://github.com/status-im/status-go/pull/1527 , to enable test recovery before user supplies password.
* FIXED ~~Use of `resolve` in `accounts.create.core` is workaround to avoid circular dependency with `accounts.recover.core`, what better solution would you prefer? Refactor?~~
* Change to `goPackagePath` in `nix/status-go/default.nix` seems necessary. Don't see that the update script will also rewrite package names in accordance with `$STATUS_GO_OWNER` or am I missing something?

##### Functional
- account recovery

status: ready

PR #8471, opened 2019-06-23
add back tooltip-triangle icon

rasom:
status: ready

PR #8687, opened 2019-08-01
Recovery v1

dmitryn:
### Summary

Recovery v1 flow 

https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=927%3A14702&viewport=494%2C241%2C0.17914098501205444

status: ready


dmitryn:
I need status-go method that takes mnemonic string and returns [address, pubkey] back. 

To show this screen:

Screenshot 2019-08-01 18 57 07


Similar to https://github.com/status-im/status-go/blob/476240fbd107fe193096350f5181c32c0539eff0/account/accounts.go#L134 but we don't need to provide password at this point.

@gravityblast  do we have such method or it needs to be implemented?

resources/icons/main/keycard-logo.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
flexsurfer: but we don't use svg   anymore

dmitryn: @flexsurfer how to implement Select key storage screen then? There is gray/blue color icon toggle  happen on click. For "This device" option there is an SVG icon used so i did same for "Keycard" option. 

Screenshot 2019-08-01 19 59 40

flexsurfer: i mean we use pngs instead

flexsurfer: we keep svg only for desktop, but i believe we'll remove them soon

dmitryn: ok, i'll change it to png

flexsurfer (approved):

gravityblast:
@dmitryn you can use the account generator, here is the documentation: https://github.com/status-im/status-go/blob/develop/account/generator/README.md

dmitryn:
@gravityblast let me describe flow i should be using, please correct me if i'm wrong as docs are not very clear on methods API:

`MultiAccountImportMnemonic(mnemonic) -> String id`
`MultiAccountDeriveAddresses(id, "m/44'/60'/0'/0/0") -> Map(path, address)`

This will get me the address , how to get public key?

dmitryn:
@gravityblast nevermind, i'm actually seeing public key in method response.

asemiankevich:
@dmitryn is it ready or still WIP? last builds are failed so e2e can't be executed

dmitryn:
@asemiankevich it's ready. Just fixed tests, next build should be ready for e2e.

asemiankevich:
@dmitryn i see this screen which is wrong because NFC is on in settings. I cant create new account since i cant bypass this screen

image

asemiankevich:
@dmitryn also, second issue - no validation for Seed Phrase present in this PR.

asemiankevich:
@dmitryn third issue - keyboard overlaps bottom sheet. Steps:
1. recover basic account
2. tap profile
3. tap Logout
4. tap Access Key

image

dmitryn:
@asemiankevich 
>i see this screen which is wrong because NFC is on in settings. I cant create new account since i cant bypass this screen

Do you try to create keycard account? Keycard is not enabled here, this PR is about basic recovery.

>no validation for Seed Phrase

There is validation, it's activated when you tap Next.

>keyboard overlaps bottom sheet.

It does because there is a focused input under sheet.

asemiankevich:
well options are not disabled :D okay then will skip this keycard ... :(

asemiankevich:
@dmitryn can we please increase tap area and open keyboard right away when this screen is shown? @hesterbruikman kindly asking for your review as well here 

I hardly can guess where i should tap to call the keyboard, seems its only where cursor blinks, not very comfortable.

image

dmitryn:
@asemiankevich it shows automatically for me in android emulator (`auto-focus` property is set to true). Are you using real device android?

asemiankevich:
@dmitryn yeah real device, android 8

dmitryn:
@asemiankevich Seems like we have issue with `auto-focus` property not working as expecting across the project. There is an example from Edit Network screen showing it doesn't always work as expected (notice how it sometimes opens keyboard, but sometimes does not). Same issue happen on Enter Passphrase screen in current PR.



@status-im/status-core-ui have you guys noticed similar issue before? I don't see any issues in React Native github repo about that.

pacamara:
@rachelhamlin @dmitryn @siphiuel @flexsurfer Looks like this is implementing the same functionality as my PR at #8519 ? If so where does it leave that?

asemiankevich:
> @asemiankevich Seems like we have issue with `auto-focus` property not working as expecting across the project. There is an example from Edit Network screen showing it doesn't always work as expected (notice how it sometimes opens keyboard, but sometimes does not). Same issue happen on Enter Passphrase screen in current PR.
> 
> 
> 
> @status-im/status-core-ui have you guys noticed similar issue before? I don't see any issues in React Native github repo about that.

okay, then lets fix this:

- missing validation for pass phrase
- overlap of the screens if i recover account and go through the screens (password set up , confirmation and Got it) too fast

image

hesterbruikman:
Great work Dmitry! Tested on iOS. Can you take a look at the issues below?

1. On iOS after typing the full seed phrase I can't continue as the Next button is hidden under the Keyboard. Return works, but don't believe it's sufficient to rely on the native keyboard only.
2. The blue cursor should have a slightly different styling. 2px x 24px rectangle with a corner radius of 1.
3. Fontsize of the entered seedphrase should be a bit smaller, 16px
4. The [...] used to truncate the key should be an ellipsis […]. Don't believe this is the case atm, can you check @dmitryn? Also please check if you're using Roboto Mono. 
5. There's some cutoff going on on the Password entry field (and Confirm). It resolves after the 5th character. Can you check what's going on there?

dmitryn:
@hesterbruikman thanks for feedback.

1. fixed.
2. I'm using default bottom button component with "next" icon https://github.com/status-im/status-react/blob/develop/resources/icons/main/next.svg - do you mean this icon by blue cursor?
3. fixed.
4. It's using ellipsis […] and monospace font.
5. fixed.

dmitryn:
@asemiankevich validation should be fixed and work realtime. 
I cannot reproduce overlap issue though.

asemiankevich:
@dmitryn thanks.

one question:

there is old Splash Screen with Get started button (it is on background here)

image


and i dont see new one 

image

is it something wrong  here or in develop?

@hesterbruikman (sorry!) @andmironov

dmitryn:
@asemiankevich i'm changing it to new Success screen

dmitryn:
@asemiankevich new splash screen is there.

Serhy:
@dmitryn 

**Issue 1**
WIth iPhoneX (iOS 12) the bottom-sheet panel with `Enter Seed phrase`-`Recover with Keycard` options is hidden behind keyboard when tap `Access key` button while cursor is focused in password field of login to account view.

Here is how it looks when I tap 'Access key':



**To reproduce:**
1) Create new account
2) Close Status
3) Open Status (you should appear right on Login form with auto-focus)
4) Tap `Access key` button


I'm not also sure whether it's OK to make real-time recovery phrase validation. This is how it looks with iPhoneX (and same on Android):

Untitled

@hesterbruikman just to double-check with you, - are we OK with real-time seedphrase validation? 

In any case, the **Issue2**: width of the tooltip with an error about seedphrase depends on the width of the value in the seedphrase input

**Issue3** 
According to design https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=927%3A14702&viewport=494%2C241%2C0.17914098501205444 we seem will drop an ability to recover an accounts with any word outside the valid words in the dictionary. With nightly we just warned user that some words might be misspeled but allowing to recover "misspelled" phrase. 
With new recovery flow we also warn user with the same `Some words might be misspelled` tooltip but nothing happens when I tap `Next >` button
@hesterbruikman are ok with it?
**To reproduce:**
1) Open Status
2) Tap `Access key`
3) Type `asd asd asd asd asd asd asd asd asd asd asd asd` as seedphrase value
4) Tap `Next >`
**Actual result:** Nothing happens
**Expected result: (based on release/develop 0.14)** `Is the phrase correct? Cancel/Confirm Phrase` pop-up appears. And tapping `Confirm Phrase` account is recovered.

** Issue 4 **

keyboard overlaps bottom sheet. Steps:

1. recover basic account
2. tap profile
3. tap Logout
4. tap Access Key

image

hesterbruikman:
>   2. I'm using default bottom button component with "next" icon https://github.com/status-im/status-react/blob/develop/resources/icons/main/next.svg - do you mean this icon by blue cursor?

With Blue cursor I mean the rectangle on the text entry field.
Entry field cursor

dmitryn:
@hesterbruikman i'm not sure it's possible to change that in react native. If i set font-size to 24, text also becomes bigger. Also, i don't see a way to change border radius of caret/cursor.

dmitryn:
@Serhy regarding Issue 1 - that is the issue with bottom sheet component in general, it doesn't expect opened keyboard and gets stuck to the bottom of the screen. Do you think we should lift it up or force closing the keyboard when bottom sheet appears?

hesterbruikman:
Thanks from highlighting @Serhy !

Issue 2 I don't believe we specified real-tme validation in designs, but am very happy to have it. As we do have to be careful with the extend of validation, here are some rules for your review. I'll also ping some others in core-UI to check:

On `space` check for: 

- Typos
Tooltip: "Some words might be misspelled"
- Dictionary
Tooltip: "Some words might be misspelled" _(Opting here for a generic message. Could be something like This word is not in the dictionary, but I'd say we don't want to provide too detailed feedback out of security concerns)_
- Word count
If <12 words Next button **disabled** _(Currently, "Recovery phrase is invalid", I think we can remove this tooltip from real-time validation and reserve for '_Next_' validation)_
If 12 or more Next button **enabled**

On `Next` check for:
- Word count
If 12 15 18 21 24 Recover
If not 12 or 24 show tooltip
Tooltip: "Recovery phrase is invalid"

- Typos (see tooltip above)
- Dictionary (see tooltip above)


@dmitryn here are some more rules discussed in #status-core-ui with mostly the advice to see what code can be reused from the existing validation.

*Additional rules from current validation in release*
- Case is irrelevant for validation. All characters should be displayed and validated as lower case
- Punctuation, numbers and special characters are invalid. Tooltip: "Some words might be misspelled"
- More existing rules here: https://github.com/status-im/status-react/blob/b13864d0521637bca0af2f85ebf133db1f7935a7/src/status_im/ethereum/mnemonic.cljs

hesterbruikman:
> @Serhy regarding Issue 1 - that is the issue with bottom sheet component in general, it doesn't expect opened keyboard and gets stuck to the bottom of the screen. Do you think we should lift it up or force closing the keyboard when bottom sheet appears?

From design POV, force close keyboard. The bottom sheets with little handle at the top never have a text entry field.

dmitryn:
@Serhy i've added keyboard dismiss fix. Now checking what's wrong with the error tooltip component width.

dmitryn:
Passphrase validation now runs on Next button pressed and on space entered.

asemiankevich:
thanks @dmitryn looks great! However i still want to double check the point, when user enters less or more than 12 words - what is expected result? Should we show the error as the figma says?

@hesterbruikman @andmironov 

image

dmitryn:
@asemiankevich if you enter space after 13th word for instance, it would run the validation.

asemiankevich:
> @asemiankevich if you enter space after 13th word for instance, it would run the validation.

Well that is also true, however there is nothing saying about the incorrect words count.. Because if i delete the space manually and still have 13 words in place - i get stuck from user POV: no error messages and Next button is not active.

https://drive.google.com/open?id=10kpqsOEf4y939jjETyu8_AJmoELzTtIx

also, from my Point of view - spaces before 1 word and after last word should be trimmed by the app and not be considered for validation

image


@hesterbruikman  @andmironov @flexsurfer @errorists maybe you can review?

hesterbruikman:
> thanks @dmitryn looks great! However i still want to double check the point, when user enters less or more than 12 words - what is expected result? Should we show the error as the figma says?
> 
> @hesterbruikman @andmironov
> 
> image


I suggest to include a variation to the Figma error message: _**"Please make sure the phrase you enter has 12, 15, 18, 21, or 24 words."**_ When the user tap the **inactive `Next` button**.

I do agree that we somehow need to inform users of what's going on, without showing an error message that is triggered continuously. Which would be the case with the message being dependent on real-time word count validation.

As I understand it, we accept 12, 15, 18, 21, 24 word seed phrases. @dmitryn please correct me if I'm wrong. Ideally, the `Next` button state depends on reaching either of these word counts. This would be the real-time validation.

- +12 words Disabled
- 12 words Enabled
- +12 <15 Disabled
- 15 Enabled
- +15 <18 Disabled
- 18 Enabled
- +18 <21 Disabled
- 21 Enabled
- +21 <24 Disabled
- 24 Enabled
- +24 Disabled

dmitryn:
@hesterbruikman 
>we accept 12, 15, 18, 21, 24 word seed phrases

Correct.

>I suggest to include a variation to the Figma error message: "Please make sure the phrase you enter has 12, 15, 18, 21, or 24 words." When the user tap the inactive Next button.

I can add it. But still doesn't solve issue @asemiankevich described - how does user know that inactive Next button needs to be tapped to display the error message?

flexsurfer:
can we just make next button enabled if the input is not empty,on press trim string and show error messages if there are not enough words?

hesterbruikman:
> I can add it. But still doesn't solve issue @asemiankevich described - how does user know that inactive Next button needs to be tapped to display the error message?

I think that's fine. Convinced people will tap when they believe they are done.

> can we just make next button enabled if the input is not empty,on press trim string and show error messages if there are not enough words?

My concern is that with real time validation on word count, the message will come up uncessesarily for the first 11 words. We will not know if people are simply not done or if they forgot a word. 

Best future alternative I see is to have a show of the word count. However, we need visual design to include it and I'd rather safe it as a future improvement outside of this PR.

src/status_im/multiaccounts/recover/core.cljs
```diff
@@ -127,9 +137,159 @@
...
+    paths
+    (fn [result]
+      (re-frame/dispatch [:multiaccounts.recover/derive-addresses-success result])))))
+
+(fx/defn on-import-mnemonic-success
+  {:events [:multiaccounts.recover/import-mnemonic-success]}
+  [{:keys [db] :as cofx} result]
+  (let [{:keys [id]} (types/json->clj result)]
+    {:multiaccounts.recover/derive-addresses {:account-id id
+                                              :paths      [constants/path-default-wallet]}}))
```
flexsurfer: @dmitryn you have to derive whisper as well, for multiaccounts we have different paths for whisper and for wallet

flexsurfer (changes requested):

dmitryn:
@asemiankevich added words count validation and made Next button available to tap when phrase is not empty.

src/status_im/multiaccounts/recover/core.cljs
dmitryn: @flexsurfer yes, but what key we show on "Successfully recovered" screen? https://github.com/status-im/status-react/pull/8687#issuecomment-517351780
Now we show key derived from wallet path.

flexsurfer: we should show whisper public key, you need to add to paths constant/whisper and get public key from whisper path keyword, not wallet

flexsurfer: it should work same as here https://github.com/status-im/status-react/blob/60de46b6c22aa8266a6739b7c730cb7337af7e05/src/status_im/multiaccounts/create/core.cljs#L290

dmitryn: I mean this derivation is only used to show user what key he recovered

flexsurfer: so this is how it should work
1) import mnemonic -> get id and root address
2) derive whisper path and wallet path using id
3) show whisper public key
4) create new multiaccount with root address, whisper public key and default wallet account (address)

flexsurfer: also you need to change `validate-recover-result`

dmitryn: @flexsurfer i'm using status-go `recoverAccount(passphrase, password)`, should i use `multiAccountStoreDerived` instead?

dmitryn: This one https://github.com/status-im/status-react/blob/60de46b6c22aa8266a6739b7c730cb7337af7e05/src/status_im/multiaccounts/create/core.cljs#L49

flexsurfer: yes you should use `multiAccountStoreDerived`

dmitryn: @flexsurfer requests changes applied, please take a look

src/status_im/ui/screens/multiaccounts/recover/subs.cljs
```diff
@@ -0,0 +1,53 @@
+(ns status-im.ui.screens.multiaccounts.recover.subs
```
flexsurfer: this file looks weird, have you seen this https://github.com/status-im/status-react/blob/develop/doc/codebase-structure-and-guidelines.md#subscriptions ?  you could have only one subscription :multiaccounts/recover and just retrieve keys in a view

flexsurfer (approved):

hesterbruikman:
> @asemiankevich added words count validation and made Next button available to tap when phrase is not empty.

Are you guys ok having a quick call today or tomorrow? Seems like the faster way to get the validation logic right. The latest build shows the behavior I was hoping we could prevent. Meaning it shows an 'error' message on every space.

asemiankevich:
sure, please schedule something for tomorrow @hesterbruikman

src/status_im/ui/screens/multiaccounts/recover/subs.cljs
dmitryn: ok, i've deleted that file and now using root subscription

dmitryn:
1. Add words count
2. Next is disabled until words count criteria met (12,15,etc words entered)
3. Error tooltip (typo error, word not in dictionary error) gets shown when user hits Next button

cc @hesterbruikman

annadanchenko:
> Error tooltip (typo error, word not in dictionary error) gets shown when user hits Next button

not sure here if user can proceed by taping Next in case of typo or not. What I would expect is that user can see a warning about typo but can proceed by tapping on Next (for words not in our dictionary but valid seed words in other wallets)

dmitryn:
@annadanchenko how do we know when to show error tooltip then? We need some action from user to trigger it. I was using real time validation technique where i run validation on each type (with delay like 0.5 seconds) and show error tooltip when something is wrong. I can return it back and only show typo errors this way.

hesterbruikman:
> 
> 
> @annadanchenko how do we know when to show error tooltip then? We need some action from user to trigger it. I was using real time validation technique where i run validation on each type (with delay like 0.5 seconds) and show error tooltip when something is wrong. I can return it back and only show typo errors this way.

Yes exactly, that would be great. Real-time validation for dictionary on space. Didn't work well for word count, but does work for dictionary check.

dmitryn:
@hesterbruikman what about typo errors? When should we run validation for it?

hesterbruikman:
> 
> 
> @hesterbruikman what about typo errors? When should we run validation for it?

Also on space, word count is the only odd one out. And thanks for adding the 0.5 sec delay. That's good practice!

hesterbruikman:
Updated the designs as we discussed. Took some liberty with how to display the word count information to stay in line with the pattern we use for ENS. 

These are the changes for the use cases:
1. Recover first screen
2. Recover with incorrect word count
3. Recover with correct word count
4. Recover with correct word count and typo
5. Recover with incorrect word count and typo

### 1. Recover first screen
- Hint copy adjusted and moved to bottom
Recover_start

### 2. Recover with incorrect word count
- Word count appears above hint copy
- Next is disabled
Recover with phrase (incomplete word count)

### 3. Recover with correct word count
- Hint copy disappears
- ✓ 12 words appears
- Next is enabled
Recover with phrase (complete word count)

### 4. Recover with correct word count and typo
- Hint copy disappears
- ✓ 12 words appears
- Tooltip appears on typo detection
- Next enabled
Wrong recovery phrase (typo)

### 5. Recover with incorrect word count and typo
- Hint copy remains
- Word count appears above hint copy
- Tooltip appears on typo detection
- Next disabled
Wrong recovery phrase (word not in dictionary)

dmitryn:
@jakubgs ios build is failing https://ci.status.im/blue/organizations/jenkins/status-react%2Fprs%2Fios/detail/PR-8687/32/pipeline/106

dmitryn:
Seems like issue is getting solved here https://github.com/status-im/status-react/issues/8745

jakubgs:
"getting solved" is probably a vast exaggeration. I'm mostly trying to avoid suicidal thoughts right now.
There is nothing more infuriating than Apple code signing.

dmitryn:
@asemiankevich New validation logic has been implemented. It can be tested on Android as iOS doesn't build for now.

asemiankevich:
thanks @dmitryn . Regarding points 4 and 5:

4. **Recover with correct word count and typo**

- Hint copy disappears
- ✓ 12 words appears
- Tooltip appears on typo detection
- Next enabled

I assume we still **want** to let user to proceed in this case? Like we have 12 words, some of them [misspelled] which could mean either a typo or a word outside the dictionary we check against. Now next button is enabled but i can't proceed until i fix the typos - expected?

image


5. **Recover with incorrect word count and typo**

- Hint copy remains
- Word count appears above hint copy
- Tooltip appears on typo detection
- Next disabled


No typo detection is done in this case, for example i have 11 words and one of them is not a word. Only check i see is words count. Expected?

image

@hesterbruikman @andmironov @annadanchenko please confirm.

asemiankevich:
Also, @dmitryn and @flexsurfer - i know you have done something with recovery method - i can't recover my old accounts here in this PR. each time i do that - some random wallet key / public key pair is generated, so no assets recovered. This is the reason tests are failing here.

Please take a look

hesterbruikman:
> 4. I assume we still want to let user to proceed in this case? Like we have 12 words, some of them [misspelled] which could mean either a typo or a word outside the dictionary we check against. Now next button is enabled but i can't proceed until i fix the typos - expected?

Running into this as well. Indeed, Next should be enabled and tap should continue. 

> 5. No typo detection is done in this case, for example i have 11 words and one of them is not a 
word. Only check i see is words count. Expected?

@asemiankevich As validation happens on space. Could it be that tooltip for typo is not shown yet? Simply because the validation hasn't been triggered for the 11th word.

### Other
- @dmitryn can you think of any case in which the key recovery is not successful, for example in the case where a validated word is not in our dictionary and contains a typo? Want to make sure we capture was should happen in that situation.
- Not sure how to best trigger validation on the 12th word. If there is a typo in this word, validation will likely not be triggered as the user will not add a space to continue. I'm willing to let this be as I think it's an edge case and don't see a clear solution. 

## Visual UI notes
@dmitryn 
- [ ] Can you please adjust the spacing. Word count 38px from the text entry box, Hint copy 77px from the text entry box.
- [ ] Tooltip is missing shading on Android. Can you check please? I thought this was working fine, but it might be Android specific. I know we have background issues there. It's a commonly used pattern so there might be something resusable.
- [ ] Next button should come up with Keyboard at all times (disabled+enabled), it currently only moves up when it's enabled at the 12th word. I know you fixed it before so seems like either it doesn't work with Android keyboard yet or the disabled state of the Next button is throwing a fit.

flexsurfer:
@asemiankevich @dmitryn wallet should be the same as before, whisper should be different that's correct, but i'm not sure what does mean "each time i do that - some random keys generated" they should be the same each time

asemiankevich:
> @asemiankevich @dmitryn wallet should be the same as before, whisper should be different that's correct, but i'm not sure what does mean "each time i do that - some random keys generated" they should be the same each time

well it means that i have recovery phrase which i want to restore and i expect a particular wallet key . However when i recover account - i see the wallet key is not the one i expected.

asemiankevich:
yeah @flexsurfer right.. I recovered `0x6f2f9f19512c919f1b265ebfe9b2adf44e75d0cd` key , but balance is 0 in status app , however it is not on etherscan https://etherscan.io/address/0x6f2f9f19512c919f1b265ebfe9b2adf44e75d0cd

image

dmitryn:
@hesterbruikman 
> can you think of any case in which the key recovery is not successful, for example in the case where a validated word is not in our dictionary and contains a typo? Want to make sure we capture was should happen in that situation.

No, if correct amount of words is entered, key would be recovered even if words are not in dictionary or has a typo.

>Not sure how to best trigger validation on the 12th word

I also don't see other way than trigger validation on each character (which we decided not to do)

> Running into this as well. Indeed, Next should be enabled and tap should continue.

Yes, now if some error detected we don't allow to proceed. I will change that

dmitryn:
@hesterbruikman 

> Next button should come up with Keyboard at all times (disabled+enabled), it currently only moves up when it's enabled at the 12th word. I know you fixed it before so seems like either it doesn't work with Android keyboard yet or the disabled state of the Next button is throwing a fit.

There is no space for Next button, space is taken by hint+margin.

>Can you please adjust the spacing. Word count 38px from the text entry box, Hint copy 77px from the text entry box.

It's 38px for words count from text entry. And also 38px for hint from words count (margin is not measured for text entry for hint box, but from words counter)

>Tooltip is missing shading on Android. Can you check please? I thought this was working fine, but it might be Android specific. I know we have background issues there. It's a commonly used pattern so there might be something resusable.

I'm using standard error tooltip for text input. Will check what's wrong there.

dmitryn:
>wallet should be the same as before, whisper should be different that's correct, but i'm not sure what does mean "each time i do that - some random keys generated" they should be the same each time

@asemiankevich yes, just checked one more time - wallet is the same as before, but whisper identity is new.

dmitryn:
@hesterbruikman i don't see way to add shadow to tooltip triangle on Android as it doesn't support `shadow-*` style properties, but only `elevation` property which seems not working with SVG (we use svg for tooltip triangle icon). So tooltip error without bottom triangle on Android could look like this:

Screenshot 2019-08-14 15 13 22

hesterbruikman:
> @hesterbruikman i don't see way to add shadow to tooltip triangle on Android as it doesn't support `shadow-*` style properties, but only `elevation` property which seems not working with SVG (we use svg for tooltip triangle icon). So tooltip error without bottom triangle on Android could look like this:

You might find something useful here, Roman looked into the same issue recently: https://github.com/status-im/status-react/pull/8471

flexsurfer:
@dmitryn we don't use svg on mobile

dmitryn:
@hesterbruikman 
>You might find something useful here, Roman looked into the same issue recently: #8471

I don't see shadow on tooltip there.

dmitryn:
@flexsurfer 
>we don't use svg on mobile

SVG icon is used for error tooltip - do you mean we should remove it? 
https://github.com/status-im/status-react/blob/fd2c37df7180e00afe19f89b49f26d3d4c1bee41/src/status_im/ui/components/tooltip/views.cljs#L20

flexsurfer:
i believe we just didn't rename it, it uses pngs

dmitryn:
@flexsurfer right, now i see what you mean, there is an image used instead of SVG. I will try to create PNG with shadow then.

dmitryn:
@hesterbruikman how can i export tooltip triangle png with shadow from Figma?

hesterbruikman:
> 
> @hesterbruikman how can i export tooltip triangle png with shadow from Figma?

Can you check if this works? png in 3 sizes. Find it a bit difficult to tell as the image is white.
[Tip@.zip](https://github.com/status-im/status-react/files/3501517/Tip%40.zip)

hesterbruikman:
Also took a look at the spacing, the text boxes were overlapping in Figma, which probably threw off the spec overlay. With the spacing on the right image below, the Next button should be able to stay above the keybord without conflicting with the hint text.

> > Next button should come up with Keyboard at all times (disabled+enabled), it currently only moves up when it's enabled at the 12th word. I know you fixed it before so seems like either it doesn't work with Android keyboard yet or the disabled state of the Next button is throwing a fit.
> 
> There is no space for Next button, space is taken by hint+margin.
> 
> > Can you please adjust the spacing. Word count 38px from the text entry box, Hint copy 77px from the text entry box.
> 
> It's 38px for words count from text entry. And also 38px for hint from words count (margin is not measured for text entry for hint box, but from words counter)


spacings

dmitryn:
>Can you check if this works? png in 3 sizes. Find it a bit difficult to tell as the image is white.

@hesterbruikman there is no shadow. Also, even having shadow, i don't think triangle could perfectly match upper box with elevation property – it will overlap box bottom border line.

Maybe we can remove triangle from tooltip box?

flexsurfer:
@hesterbruikman @dmitryn @asemiankevich this PR is a blocker for other PRs, could we merge it and fix small issues separately?

hesterbruikman:
> @hesterbruikman there is no shadow. Also, even having shadow, i don't think triangle could perfectly match upper box with elevation property – it will overlap box bottom border line.

Let's merge without; I'll create a bounty to add it later 👍

asemiankevich:
issue with account recovery is already fixed in @flexsurfer PR

flexsurfer:
@dmitryn please resolve conflicts

externs.js
```diff
@@ -568,5 +568,6 @@ var TopLevel = {
     "multiAccountDeriveAddresses" : function () {},
     "multiAccountReset" : function () {},
     "multiAccountLoadAccount" : function () {},
-    "multiAccountStoreAccount" : function () {}
+    "multiAccountStoreAccount" : function () {},
+    "multiAccountImportMnemonic" : function () {},
```
flexsurfer, line 572: @dmitryn  something wrong with methods, here you have 2 , but in java you have `multiAccountDeriveAddresses` in ios only `multiAccountImportMnemonic`

dmitryn: @flexsurfer maybe it was merge conflicts, now i'm only adding `multiAccountImportMnemonic` to both java and objective-c code.

PR #8697, opened 2019-08-05
Remove old-message-id

cammellos:
Old message id was used for compatibility of replies with older clients.
Given that v1 is breaking, this is not needed anymore and simplifies
moving messages to status-go. No protocol/data-store change is made, to minimize
changes.

### Testing

- replies from this build to this build

status: ready

### DIFF
--- a/src/status_im/chat/models/input.cljs
+++ b/src/status_im/chat/models/input.cljs
@@ -87,12 +87,11 @@
 
 (fx/defn reply-to-message
   "Sets reference to previous chat message and focuses on input"
-  [{:keys [db] :as cofx} message-id old-message-id]
+  [{:keys [db] :as cofx} message-id]
   (let [current-chat-id (:current-chat-id db)]
     (fx/merge cofx
               {:db (assoc-in db [:chats current-chat-id :metadata :responding-to-message]
-                             {:message-id     message-id
-                              :old-message-id old-message-id})}
+                             {:message-id     message-id})}
               (chat-input-focus :input-ref))))
 
 (fx/defn cancel-message-reply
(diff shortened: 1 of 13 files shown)


flexsurfer (approved):

src/status_im/chat/models/input.cljs
```diff
@@ -134,8 +133,7 @@
                                             :content      (cond-> {:chat-id current-chat-id
                                                                    :text    input-text}
                                                             message-id
-                                                            (assoc :response-to old-message-id
-                                                                   :response-to-v2 message-id)
+                                                            (assoc :response-to-v2 message-id)
```
pedropombeiro, line 136: Does this `v2` now tie into the status-protocol-go's `v1`? Maybe we should get rid of this `v2` naming at some point?

cammellos: Yes, I decided not to make any protocol change in this PR to keep the change surface small, but given that this is merged, we can just rename  `-v2` to `-v1` or non-namespaced

vkjr (approved):

yenda (approved):

Serhy:
@cammellos still can't verify with iOS on real builds due to https://github.com/status-im/status-react/issues/8745

But there is a regression found with Android builds (looks OK with 14th August nightly develop Android build):
There is no reply (in 1-1/Group/Public chats) when replying to message with emoji

Reply5555


**To reproduce:**
2) Join public chat (e.g. `#replywithemoji` or any other type of chat)
3) Send (or receive) a text (or emoji) message
4) Reply to message with emoji

jakubgs:
iOS signing issue should be fixed.

Serhy:
With iOS build, there are same results and same single issue with replies between builds (PR_build <-> PR_build) as in https://github.com/status-im/status-react/pull/8697#issuecomment-521239879

cammellos:
@Serhy closing this as included in #8763 , could not replicate the issue there, but it might still exist.

PR #8719, opened 2019-08-07
Add datasync confirmations, enable device-to-device by default

cammellos:
This commit adds datasync confirmations and enables device-to-device for
all the communications, as that's what we will go with v1, so this build is compatible with > 0.9.32.

### Testing

- Message confirmations, with datasync on and off

Messages should be marked as `sent` if there is network, and should be marked as not sent (after a while), if sent while offline (turn on airplane mode and quickly send a message).

status: ready


flexsurfer (approved):
do we want to enable pfs?

cammellos:
@flexsurfer yes, this commit enables it (device to device = pfs ) , as we want it to be the default for v1

churik:
Tested device pairing and blocking / unblocking users with 0.13.0, PR <-> PR.
No new issues found. (build with commit https://github.com/status-im/status-react/commit/2d77e38314dc0f4c168e36b55a28b9b46f8203d0)

Serhy:
With the latest builds (2019081212) messaging in 1-1 / Group / Public chat keep working the same to develop build when DataSync (and V1 Messages) settings are enabled in profile and when they disabled. 
Messages statuses are 'Sent' as expected with the network connected (tested with WiFi and 3G connections). 
`Not Sent` status was a bit hard to get for basic messages in a way for turning On/Off the network - but in a case we're lucky to catch blue 'Send' button and tap it - there were or endless spinner (when messages sent but still no network during 3+ minutes) or 'Sent' status when a network connection established. 
However, it's possible to test it with the stickers (by tapping one while offline). In that case 'Not sent. Tap for options' status appears with both working options for the message ('Delete' and 'Resend')
A separate issue for endless spinner created in https://github.com/status-im/status-react/issues/8734

PR #8725, opened 2019-08-08
Status go/integrate accounts persistence

yenda:
### Summary
Currently the regular account creation, login/logout cycle seems to work. Ongoing testing and small fixes on 12/08/19

TODOS

* [x] can't logout after saved password
* [x] empty chat screen after login with saved password (and maybe other conditions)
* [x] check if all settings are migrated
* [x] custom bootnodes 
* [x] stickers (using settings, to be improved afterward)
* [x] save and use whisper and wallet accounts
* [x] write js externs for new native bindings and remove old ones
* [x] recovery flow (waiting for recovery flow v1)
* [x] fix recovery flow stucked on loading screen after password input
* [x] fix adding account in wallet
* [x] fix pending message status (needs rebase of status-go branch)
* [x] desktop native bindings kudos @siphiuel 
* [x] iOS native bindings kudos @siphiuel 
* [x] fix tests

### Testing
This PR basically affects the entire app, however I think there is no point digging too hard during testing considering that:

* a followup PR will remove realm which will smooth up the login flow
* keycard flow will be fixed in a followup PR

#### Knowns bugs that will be fixed in future PR

* [ ] keycard flow (worked on by @dmitryn )
* [ ] there might be a realm error when starting up, it can be ignored as next PR will remove realm anyway

#### Tested by dev:
Android real devices are hanging, looking into it (tested One Plus 5T)
iOS is fine (tested iPhone 6 Plus)

I tested the following actions on Android emulator:

* creating and recovering multiaccount
* logout and login
* join public and 1-1 chat
* send public and 1-1 message
* create new wallet account
* logout and login

status: ready


externs.js
```diff
@@ -563,5 +567,6 @@ var TopLevel = {
     "createAppContainer" : function () {},
     "useScreens" : function () {},
     "multiAccountGenerateAndDeriveAddresses" : function () {},
+    "multiAccountStireAccount" : function () {},
```
flexsurfer: Stire

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -119,10 +119,14 @@
...
 (fx/defn add-transfer
+  "We determine a unique id for the transfer before adding it because some
+   transaction can contain multiple transfers and they would overwrite each other
+   in the transfer map if identified by hash"
   [{:keys [db] :as cofx} {:keys [hash id] :as transfer} address]
-  (let [transfer-by-hash (get-in db [:wallet :accounts address :transactions hash])]
-        ;;transfer-by-id   (get-in db [:wallet :transaction id]) ;; TODO didn't found any usage of this
-    (when-let [unique-id (when (not= transfer transfer-by-hash) ;(or transfer-by-id)
+  (let [transfer-by-hash (get-in db [:wallet :accounts address :transactions hash])
+        transfer-by-id   (get-in db [:wallet :transaction id])]
```
flexsurfer: @siphiuel @yenda could you please point where we set this? i couldn't find

yenda: it's status-go side

flexsurfer: wut? i mean where in status react we `assoc` value to `:wallet :transaction id` ?

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -353,54 +387,46 @@ public void moveToInternalStorage(Callback callback) {
...
-            @Override
-            public void run() {
-                String result = Statusgo.login(json);
-
-                callback.invoke(result);
-            }
-        };
+                @Override
+                public void run() {
+                    String result =Statusgo.openAccounts(rootDir);
```
flexsurfer, line 432: `=S`

src/status_im/ethereum/transactions/core.cljs
yenda: it's right here. the transfer has a hash and id when it arrives from status-go, and in this function we check if there is already a transfer in the map that has this hash or id as key, if the hash is already there we use the id, if the id is already there we already have the transfer in the map

src/status_im/ethereum/core.cljs
```diff
@@ -99,7 +99,7 @@
 (defn chain-keyword
   [db]
   (let [network-id (get-in db [:multiaccount :network])
-        network    (get-in db [:multiaccount :networks network-id])]
+        network    (get-in db [:multiaccount :networks/networks network-id])]
```
flexsurfer, line 102: how did it work before?

flexsurfer (approved):
this is just incredible work! HUGE KUDOS!

src/status_im/ethereum/transactions/core.cljs
flexsurfer: still there should be a place where we put `:transaction` i can't find it, to get `(get-in db [:wallet :transaction id])` we need to `put` it there first

src/status_im/ethereum/core.cljs
yenda: by renaming the key somewhere else

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
yenda: what?

src/status_im/ethereum/transactions/core.cljs
yenda: that is suppose to be the list of transactions

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
flexsurfer: space

dmitryn (approved):

jakubgs (approved):
Approving only the Nix changes.

>−5832 lines

yenda:
@asemiankevich pending build is going to fix mainnet warning reappearing all every login on real device and logout with saved password

asemiankevich:
@yenda Network configs are missing is that okay?
image

asemiankevich:
@yenda error on logout

image

yenda:
@asemiankevich thank you I realize my fix for logout was wrong, the issue is that addresses of multiaccounts listed by status-go are lower-case, but then the address of the multiaccount is multicase. So when user saved password it was saved under lowercase address in keystore, and when logging out it was trying to delete the eip55 version.
fixed now

asemiankevich:
@yenda 

1. network name is missing
image

2. cant resolve `simpledapp.eth` on Ropsten
image

3. transactions  fail with

image

yenda:
@asemiankevich bugs should be fixed, note that now the app logs out when you switch network so that might influence the e2e tests

asemiankevich:
@yenda status is asking for notifications before the setting is shown in onboarding flow. It should show this pop up only when i am on Enable Notifications screen settings and tap Enable notifications

Sorry , i have to mention it earlier.
It does not happen in nightly.


image


also, i see this on iOS

image

asemiankevich:
@yenda one more issue: when sending tokens to address from wallet -> invalid recipient shown and  network fee is 0

Steps:
1. open status and faucet some funds + tokens in simpledapp
2. add one more account in your wallet
3. send some tokens from Status account to new account in wallet

Pay attention : recipient is New Contract, Network fee is empty

image

asemiankevich:
@yenda Eric sorry for delay, would you mind to look into this? Can't sign message to cryptokitties 

image


cant add ethereum address in wallet
image

cant sent stt
image

churik:
at attempt to sell some SNT on AirSwap:
the same error as for 3 from https://github.com/status-im/status-react/pull/8725#issuecomment-522445387 - cannot locate account for address

asemiankevich:
iOS issues are fixed.
issue for adding account for multiaccount when using recovered account will be logged separately

PR #8733, opened 2019-08-12
disable coverage

flexsurfer:
disable coverage


jakubgs (approved):
It would be cleaner if we just used `catchError` to ignore failure, but this is fine too.

PR #8754, opened 2019-08-14
Fixed critical e2e in new profile

churik:
Fixed critical e2e in https://github.com/status-im/status-react/pull/8722


test/appium/views/base_view.py
```diff
@@ -555,12 +555,15 @@ def relogin(self, password=common_password):
         sign_in_view = self.get_sign_in_view()
         sign_in_view.sign_in(password)
 
+    def close_share_chat_key_popup(self):
+        TouchAction(self.driver).tap(None, 255, 104, 1).perform()
+
```
churik, line 560: Probably there is a better way to implement this (it should tap outside popup)

yevh-berdnyk (approved):

test/appium/views/base_view.py
Serhy: We could use just swipe_down for this purpose but this one is fine too.

Serhy (approved):

PR #8755, opened 2019-08-14
Fix back button issues on Android

### LINKED ISSUE #8735: Save password is unavailable: your device may be rooted screen is shown when tapping Back button
# Problem

Back button breaks Accounts screen on Androids. There are several cases here, which will be logged as separate issues

1. Case one: user taps `<` **in-app** button after logout (app is launched after install, new account created) -> then Back device button

Steps:
1. launch status
2. tap Get started
3. tap Generate a key
4. tap Next -> Next 
5. enter password and confirm
6. tap profile -> Log out -> confirm logout
7. tap `<` in-app button at the top left
8. tap Back device button 

image

Video: https://drive.google.com/open?id=1pIrRiA2Vygp8-1LDJy-af9aTn5vaLqGF

### LINKED ISSUE #8736: Welcome screen is shown instead of accounts list if tapping Back from Get yourself a key first screen
# Problem

Back button breaks Accounts screen on Androids. There are several cases here, which will be logged as separate issues:

2. Case two: Welcome screen is shown when user taps Back device button when being on Get yourself a key first screen. Looks like all accounts are gone (which is not true)

Steps:
1. open status and create account
2. logout
3. tap Generate new key
4. tap build-in `<` button or device Back button

Actual result: welcome screen is shown
Expected result: user should be redirected to previous screen (list of accounts)

image

Video: https://drive.google.com/open?id=1pIrRiA2Vygp8-1LDJy-af9aTn5vaLqGF

vitvly:
Fixes #8735 
Fixes #8736

Kudos to @yenda for suggesting `navigate-reset`.

## Description
What happened there is that re-frame's `navigation-stack` and react-navigation's stack begin to differ once hardware Back button is pressed. So unless react-navigation's stack is properly reset, pressing hardware Back may lead to weird behaviors.

## QA notes
Current behavior is this: 
  - pressing hardware Back on multiaccounts screen: exits the app
  - pressing hardware/software Back on `Generate keys` step in the wizard - navigates to multiaccounts screen

### DIFF
--- a/src/status_im/multiaccounts/create/core.cljs
+++ b/src/status_im/multiaccounts/create/core.cljs
@@ -75,7 +75,7 @@
                 (navigation/navigate-to-cofx :intro-wizard nil))
 
       (fx/merge {:db (dissoc db :intro-wizard)}
-                (navigation/navigate-to-clean (if first-time-setup? :intro :multiaccounts) nil)))))
+                (navigation/navigate-back)))))
 
 (fx/defn exit-wizard [{:keys [db] :as cofx}]
   (fx/merge {:db (dissoc db :intro-wizard)}
--- a/src/status_im/multiaccounts/recover/core.cljs
+++ b/src/status_im/multiaccounts/recover/core.cljs
@@ -210,8 +210,11 @@
 
 (fx/defn cancel-pressed
   {:events [:recover.ui/cancel-pressed]}
-  [cofx]
-  (navigation/navigate-back cofx))
+  [{:keys [db] :as cofx}]
+  ;; Workaround for multiple Cancel button clicks
+  ;; that can break navigation tree
+  (when-not (#{:multiaccounts :login} (:view-id db))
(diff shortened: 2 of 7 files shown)


yenda (approved):

asemiankevich:
@siphiuel i found one more case which could be great to fix:

1. open status
2. tap access key -> enter seed phrase
3. tap Cancel button at the top left twice in a row quickly

Result:

**Your PR: blank screen**

image


**Nightly: your device may be rooted screen**

image


Since there is a fix for `your device may be rooted` issue in this branch - how painful to address the issue above here in your PR as well? Or should i log it separately?

vitvly:
@asemiankevich added a workaround for multiple Cancel button clicks in recovery wizard.

vitvly:
Also pushed a fix for a similar issue for Keycard recovery (clicking back there would lead to default intro screen, clicking back twice would break navigation)

PR #8763, opened 2019-08-15
Features/move messages status go

cammellos:

1) Move messages to status-go
2) Use message-id computed from status-go
3) Remove old replies

Old message id was used for compatibility of replies with older clients.
Given that v1 is breaking, this is not needed anymore and simplifies
moving messages to status-go. No protocol/data-store change is made, to minimize
changes.

### Testing
- Sending/receiving messages in all kind of chats
- Replies
- Login/logout messages
- Performance of scrolling (just to make sure there are no major regressions) / changing chat
- Delete chats
- Block contacts

status: ready


src/status_im/chat/models.cljs
```diff
@@ -207,36 +207,30 @@
...
       (fx/merge cofx
                 {:db            (reduce (fn [acc message-id]
                                           (assoc-in acc [:chats chat-id :messages
                                                          message-id :seen]
                                                     true))
                                         db
-                                        loaded-unviewed-ids)
-                 :data-store/tx [(messages-store/mark-messages-seen-tx loaded-unviewed-ids)]}
+                                        loaded-unviewed-ids)}
+                (messages-store/mark-messages-seen loaded-unviewed-ids)
```
flexsurfer, line 233: are `message-store` and `data-store.messages` aliases for the same ns or different?

flexsurfer (approved):

src/status_im/chat/models.cljs
cammellos: same, will fix that

cammellos:
@Serhy I can actually replicate the issue you reported on the other PR, I will put this in testing as it's a much larger change, but in the meantime I'll look at that issue, if everything else will test fine, I might fix in a separate PR.

churik:
#### 1) "Fetch more messages" is not shown 

**Steps:**
- create account
- join any public chat
- scroll to oldest message to see "Fetch more history"

**Expected result:** can fetch history up to 30 days
**Actual result:** no "Fetch more history" button
**Note:** on IOS couldn't reproduce for #status channel, but if I see it - it has an issue on scrolling
screencast 2019-08-19 13-20-11
Always reproducible on empty chats, i.e. #t12345

cammellos:
@churik thanks for reporting the issue, should be fixed in the next build.

churik:
#### 2) 1-1 chat with blocked user reappears after relogin

**Steps:**
- create account
- add contact
- send several messages to 1-1
- as a contact: send several messages to 1-1
- block user
- close and open app

**Expected result:** no chat with blocked contact
**Actual result:** chat with blocked contact is restored (with own messages only)

churik:
#### 3) Number of sent messages in group chats are multiplied to amount of users 

**Steps:**
- create account
- add several contacts
- create a group chat with 2 accounts
- send "1"
- send "2"

**Expected result:** 1 and 2 as outgoing messages
**Actual result:** 
photo_2019-08-19 15 33 46
If from this state you tap on user icon in group chat > send message : last three messages in group chat will be shown in 1-1 chat.

churik:
#### 4) Styling is not applied to emoji replies

**Steps:**
- create account
- join public chat
- reply to any message with emoji

**Expected result:** blue style of reply, like for other outgoing messages
**Actual result:** 
photo_2019-08-19 17 25 18

cammellos:
@churik 2/3/4 should be fixed in the last build ( I could not replicate the issue with `Send message`, but I have checked only post fix).
Please use fresh accounts to test with the new build as I have made some db changes. Thanks!

churik:
All issues are fixed.
Tested  on IOS 11.4.1, Android 8:
- Sending/receiving messages in all kind of chats
- Replies (not compatible with 0.13.0, expected)
- Login/logout
- scrolling on 30 days history, didn't notice major regression
- Delete chats
- Block contacts
- Pairing / Syncing (PR <-> PR)
- adding to contacts
- unblocking
- fetching offline history

Issue #8757, opened 2019-08-15
[nix] Android builds failing due to "still alive" error


jakubgs:
Currently Android builds randomly fail due to some form of race condition resulting in this error:
```
finding garbage collector roots...
removing stale temporary roots file '/nix/var/nix/temproots/29792'
removing stale temporary roots file '/nix/var/nix/temproots/29812'
error: cannot delete path \
  '/nix/store/0w5f4dk7zcsylq426yq8yj32y6s8lbzj-status-react-source-gradle-install' \
  since it is still alive
make: *** [Makefile:158: jsbundle-android] Error 1
```
https://ci.status.im/job/status-react/job/prs/job/android/view/change-requests/job/PR-8713/2/

jakubgs:
Related Nix issue: https://github.com/NixOS/nix/issues/1810
Which would have added more info about why exactly the Nix store entry is "alive", but there hasn't been any activity there in almost a year.

jakubgs:
I checked manual for `nix-build` to see if I could disable garbage collection after build - since we run `nix/clean.sh` afterwards anyway - but couldn't find anything useful in that regard.

jakubgs:
One idea I have is to switch CI hosts to a multi-user Nix installation:
https://nixos.org/nix/manual/#sect-multi-user-installation
Worth a try.

jakubgs:
I attempted to install Nix in multi-user mode on `linux-04` but I've encountered an issue that the `jenkins` user does not get a profile folder. When sourcing `/etc/profile.d/nix.sh` it creates a symlink from `~/.nix-profile` to `/nix/var/nix/profiles/per-user/jenkins/profile` which does not exist.
I've opened an issue: https://github.com/NixOS/nix/issues/3051

jakubgs:
I also upgrade Nix from `2.2.1` to `2.2.2` on all hosts in https://github.com/status-im/infra-ci/commit/4b285d0fb62bc71161f2196ffd89859d53e01216.

jakubgs:
It seems like it is still happening, case in point:
https://ci.status.im/job/status-react/job/prs/job/android-e2e/job/PR-8764/14/
Pedro showed me this from the manual:
>`--no-out-link`
>Do not create a symlink to the output path. __Note that as a result the output does not become a root of the garbage collector, and so might be deleted by nix-store --gc__.

Not sure if this is relevant to our case as we do not use `--gc`:
[...]

PR #8788, opened 2019-08-20
fix nix/clean.sh race condition

jakubgs:
To avoid trying to remove stuff that an ongoing `nix-build` process might be using we call `wait-for.sh`.

Possible fix for #8757.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -135,16 +135,16 @@ SPEC CHECKSUMS:
   GoogleUtilities: 06b66f9567769a7958db20a92f0128b2843e49d5
   nanopb: 5601e6bca2dbf1ed831b519092ec110f66982ca3
   Protobuf: 1eb9700044745f00181c136ef21b8ff3ad5a0fd5
-  React: 1d605e098d69bdf08960787f3446f0a9dc2e2ccf
-  react-native-background-timer: bb7a98c8e97fc7c290de2d423dd09ddb73dcbcbb
-  react-native-camera: 68ad5143d2d0636236d46c7de8d2a6455ca52a36
-  react-native-webview: a42108b827082f8f0333529b0772102031d5960d
-  RNKeychain: 627c6095cef215dd3d9804a9a9cf45ab96aa3997
+  React: 36d0768f9e93be2473b37e7fa64f92c1d5341eef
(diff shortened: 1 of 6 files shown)


pedropombeiro (approved):

flexsurfer (approved):

jakubgs:
>we can simplify and use openjdk instead of pkgs.openjdk here.

~~How would that work? If it's under `pkgs` why would it not fail?~~
EDIT: Oh, I'm a moron, I didn't notice the `with pkgs`.

jakubgs:
I couldn't make the `wait-for.sh` script to find `pod install` properly without the `-f` option so i decided to upgrade CocoaPods from `1.5.3` to `1.7.5` in https://github.com/status-im/infra-ci/commit/4786581c.
Hopefully they fixed the race conditions when running multiple instances of `pod install`, if they didn't I will find a different solution.

jakubgs:
Okay, I changed it so that `wait-for.sh` takes as arguments after the timeout not just process name but whole `pgrep` options. So in `nix/mobile/ios/install-pods-and-status-go.sh` we can pass the `-f` flag.

jakubgs:
I don't like how this delays the `e2e` builds by quite a bit waiting for the android build for finish.
I might still merge this anyway and find a better solution later, just to avoid failed builds.

jakubgs:
I've decided to use a different approach. Now the `nix/clean.sh` script can be run without any arguments and then it uses the regex as before to clean everything, but if a path to a nix store entry is given(like we do now with the result path) then only that and it's related paths are cleaned.

I've checked the builds and even if `android` and `android-e2e` run on the same host, they still show up as different result paths, as they should:
```
Cleanup of /nix/store...
Searching by result: '/nix/store/s17r0vpyy109hxv5dil26gxhss8cwgnb-status-react-build-jsbundle-android'
 ? Checking:   /nix/store/s17r0vpyy109hxv5dil26gxhss8cwgnb-status-react-build-jsbundle-android
 - Derivation: /nix/store/jvjf8f7fd5rxah0vjpxgfrayvx0ygmyg-status-react-build-jsbundle-android.drv
 - Source:     /nix/store/wmzcjws14zxxzcr07whbc79k5is9c101-status-react-source-jsbundle
Deleting...
```
This should be much better than adding any silly blocking waits.

PR #8791, opened 2019-08-20
use react-native net-info community lib

yenda:
core lib is deprecated

### Steps to test

test if the app works properly when switching network on and off and going from mobile net to wifi

status: ready

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -257,6 +257,7 @@ android {
 }
 
 dependencies {
+    implementation project(':@react-native-community_netinfo')
     implementation project(':react-native-background-timer')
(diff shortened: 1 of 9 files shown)


flexsurfer (approved):

pedropombeiro (approved):

pedropombeiro:
The reason this build is failing is because react-native-netinfo depends on a Gradle plugin which is not in the Nix files. However, it's not enough to run `make nix-update-gradle` as this is a special type of Maven dependency which is currently not detected. Will work on that.

flexsurfer:
@PombeirP @yenda is this still relevant?

PR #8793, opened 2019-08-20
add OpenJDK toleiningen shell, necessary for generate-keystore.sh

jakubgs:
Fix for:
```
scripts/generate-keystore.sh: line 22: keytool: command not found
```
Since when using `--pure` then `keytool` is not available in the shell.


jakubgs:
Added change to #8788.

PR #8811, opened 2019-08-22
[fix 8796] big number error when adding owned ens name

### LINKED ISSUE #8796: 'Cannot write BigNumber' at attempt to bind own ENS username
# Description

*Type*: Bug

*Summary*: on nightly 21/08/2019 after commit https://github.com/status-im/status-react/commit/f2cf42715b21518829e7d02687ef9dc6f4392b38 at attempt to bind ENS name, that was bought before changing whisper ids, error is shown 'Cannot write BigNumber'.
Before there was an error, that name was already taken. 
On IOS error: 'Cannot write Function'

#### Expected behavior
can bind my own ENS

### Reproduction

- Restore account with ENS name, that was bought on 0.13.0
- Go to profile > ENS name
- Enter your ENS name
- Tap Next

### Additional Information

- Status version: nightly 21/08/2019
- Logs and video: https://app.saucelabs.com/tests/a7e2b1304e6c43d39cbf16c0464cc29f?auth=93d94767223031d8d7ac7f2a35c1376e#82

yenda:
fix #8796

## Testing

Registering a username and adding a username that you already own from a recover account should work (only if the right public-key is registered)

status: ready

### DIFF
--- a/src/status_im/ens/core.cljs
+++ b/src/status_im/ens/core.cljs
@@ -30,8 +30,12 @@
  (fn [[registry name cb]]
    (resolver/pubkey registry name cb)))
 
+(defn- final-state? [state]
+  (#{:saved :registered :registration-failed} state))
+
 (defn assoc-state-for [db username state]
-  (assoc-in db [:ens/registration :states username] state))
+  (cond-> (assoc-in db [:ens/registration :states username] state)
+    (final-state? state) (update :ens/registration dissoc :registering?)))
 
 (defn assoc-details-for [db username k v]
   (assoc-in db [:ens/names :details username k] v))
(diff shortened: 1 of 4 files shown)


flexsurfer (approved):

churik:
@yenda 
**1) no "Username added" screen after you attach username.**
**Steps:** 
- add name that connected with your whisper id
- tap next

**Expected result:**
sdsdf

**Actual result:**
screencast 2019-08-22 14-50-54

Logs and video of expected result: https://app.saucelabs.com/tests/d002f9ecb8ce4bb281efaa244eeb9657?auth=0d35eb158e630d03631534c65f90c712#99
The same screen is expected after registering new name - it is missing also.

**2) Q - will we handle cases when Username is owned by you, but not connected with your public key separately?**
Currently, we have workaround to change whisper id in old Dapp `names.statusnet.eth` and then bind it in profile.
But it is of course not obvious at all and if users bought ENS names on 0.13.0 - there is no other way to do it.
@hesterbruikman @rachelhamlin need your opinion on this.

yenda:
1) I'll look at it
2) not in this PR, this require quite a few changes

churik:
@yenda one more small issue.
**2) after tapping on "OK, got it" on "Username is added" screen keyborad is opened and closed.** 
It wasn't before.
screencast 2019-08-22 16-41-06
Would you mind to fix it here?

Thank you for working on this one!

churik:
@yenda 
one more issue here, sorry.
**3) If you add first ENS -> it is not shown on your profile**
If you add second username -> it is shown
Expected, that first added ENS will be used in your profile by default.
Logs and video: https://github.com/status-im/status-react/pull/8811#issuecomment-523941745

yenda:
@churik 
 2) looks fixed in the emulator, it doesn't look like it was related to this PR though, we really have a problem with screens and navigation in general. 
The keyboard was appearing because all the registration process is actually one screen that looks different based on some data. 
First we shouldn't do that, and second when the registration process is done and you tap ok, it removes this data, and for a moment the screen goes back to its form where it has an input field. 
It probably doesn't even show in a frame, but that is enough for the keyboard to show up.
 So I reverted the condition so that the default screen isn't the one with an input to avoid that, but it's more like a temporary thing


 3) fixed, that one was my bad, I inverted the condition to set the registered name as preferred name, so it was doing it for when you have more than one name instead of only for the first name.

Thanks for testing!

asemiankevich:
1. can buy username and see it in profile
2. can add previously bought username and see it in profile
3. can switch between 2 usernames and see the profile updated

I think we are good with this PR.

PR #8815, opened 2019-08-22
Add processing indicator when creating multiaccount

### LINKED ISSUE #8743: Fix [Next] button on password confirmation screen (onboarding)
# Problem

There are several issues with Next button when confirming password in new onboarding flow:

1. No visual effect that button is tapped
2. I can tap this button several times (should be disabled once tapped first time)
3. If i tap the button several times, i either close next screen (with Fingerprint settings) or i can run into https://github.com/status-im/status-react/issues/8710
4. Screens are not smooth by tapping this Next button, looks like they are stuck for a second

## Steps

1. open status
2. go though onboarding until password set up and confirmation
3. enter password and tap Next button quickly several times

vitvly:
Fixes #8743

### DIFF
--- a/src/status_im/multiaccounts/create/core.cljs
+++ b/src/status_im/multiaccounts/create/core.cljs
@@ -109,7 +109,7 @@
 (fx/defn intro-step-forward
   {:events [:intro-wizard/step-forward-pressed]}
   [{:keys [db] :as cofx} {:keys [skip?] :as opts}]
-  (let  [{:keys [step first-time-setup? selected-storage-type]} (:intro-wizard db)]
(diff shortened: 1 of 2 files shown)


yenda (approved):

flexsurfer (changes requested):
could you please add processing check into `:intro-wizard/step-forward-pressed` event as well, because it's still possible to press the button twice

vitvly:
@flexsurfer Thanks, good point, fixed.

flexsurfer (approved):

churik:
Tested on Android 8 And IOS 11.4.1 on both recovery sign up flows. 
Work good!

PR #8821, opened 2019-08-22
[Fix 8635] About showed "Version ()" instead of Version 0.14.0 (2019…)

### LINKED ISSUE #8635: About shows "Version ()" instead of Version 0.14.0 (201907...) on Android
### Description

*Type*: Bug

*Summary*: Only on Android: on both today's nightly and release build 12 there is no version shown in Profile -> About. 
Screenshot_20190725-112419

#### Expected behavior

a value like Version 0.14.0 (2019072413)

#### Actual behavior

Version ()

### Reproduction

- Open Status, create an account
- Open Profile -> About and check value shown for Version

### Additional Information

- Status version: nightly from July 25, release build 12 
- Operating System: Android only (real Pixel XL, Android 9)

MajorTomSec:
fixes #8635 

### Summary

This fixes #8635 by adding VERSION and BUILD_NUMBER files in the correct nix template and updating bash script in order to prevent it from failing, due to the git repository being not initialized in the nix environment.

#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

### Steps to test
- Open Status
- Create an account
- Open Profile -> About and check value shown for Version

status: ready


scripts/build_no.sh
```diff
@@ -23,12 +23,10 @@
 #
 #####################################################################
 
-# Fail on first error
 set -e
 
-GIT_ROOT=$(git rev-parse --show-toplevel)
 DEV_BUILD_NUMBER=9999
-BUILD_NUMBER_FILE="${GIT_ROOT}/BUILD_NUMBER"
+BUILD_NUMBER_FILE="BUILD_NUMBER"
```
jakubgs: Wouldn't it make more sense to do something like:
```bash
if [[ -z "${BUILD_NUMBER_FILE}" ]]; then
  BUILD_NUMBER_FILE="${GIT_ROOT}/BUILD_NUMBER"
fi
```
And then provide the path of `BUILD_NUMBER_FILE` in Nix?
What do you think @PombeirP?

jakubgs (changes requested):

scripts/build_no.sh
pedropombeiro: I don't understand in which situation this script gets called and .git isn't there. Normally it is passed to Nix [here](https://github.com/status-im/status-react/blob/develop/ci/android.groovy#L38) where we have a Git repo. Unless it is happening [here](https://github.com/status-im/status-react/blob/develop/src/status_im/utils/build.clj#L25), but if that is the case, then I don't understand the need to add the `VERSION` to the jsbundle Nix expression, since it'll cause the bundle to be invalidated any time the `VERSION` file changes.

But yeah, I agree that passing an env var to the file in this situation is better as it still allows us having the script live anywhere and still have it behave correctly.

nix/targets/jsbundle.nix
```diff
@@ -20,12 +20,12 @@ in stdenv.mkDerivation {
       filter =
         # Keep this filter as restrictive as possible in order to avoid unnecessary rebuilds and limit closure size
         mkFilter {
-          dirRootsToInclude = [ 
+          dirRootsToInclude = [
             "components/src" "react-native/src/cljsjs" "react-native/src/mobile" "src" "env/prod" "prod" # Taken from project.clj :profiles :prod :cljsbuild :builds :android :source-paths
-            "resources" "status-modules/cljs" "status-modules/resources"
+            "resources" "status-modules/cljs" "status-modules/resources" "scripts"
```
pedropombeiro: I'd maybe move the `scripts/gen_build_no.sh` and `scripts/build_no.sh` files to a subfolder (`scripts/version` or something), so that changes in unrelated scripts don't cause this Nix expression to be rebuilt.

pedropombeiro (changes requested):
Great job so far @MajorTomSec! Just added a couple questions/remarks.

MajorTomSec:
Thanks for the reviews.
I took into account all of your remarks and came up with this.

@PombeirP the issue is indeed with https://github.com/status-im/status-react/blob/develop/src/status_im/utils/build.clj#L25 and this is why the VERSION file has to be added to nix as well (see [here](https://github.com/status-im/status-react/blob/develop/src/status_im/utils/build.clj#L36)

I thought instead of using an env variable for specifying BUILD_NUMBER_FILE, I used pwd when git is not available:
`GIT_ROOT=$(git rev-parse --show-toplevel || pwd)`

I believe this way nothing would be broken. As build.clj uses relative path for VERSION and the script path, this would not fail here anyway.

Let me know if you have other remarks.

EDIT: @PombeirP we cannot keep the set -e with the workaround I put there.
But your workaround seems to solve the issue, nice one ! I will update ASAP.

MajorTomSec:
I believe everything should be fine by now. 
Waiting for any other remark anyone might have.

scripts/version/build_no.sh
```diff
@@ -26,12 +26,14 @@
 # Fail on first error
 set -e
 
-GIT_ROOT=$(git rev-parse --show-toplevel)
 DEV_BUILD_NUMBER=9999
-BUILD_NUMBER_FILE="${GIT_ROOT}/BUILD_NUMBER"
+
+SCRIPTPATH="$(cd "$(dirname "$0")" ; pwd -P)"
```
jakubgs, line 31: That's okay, but I kinda like your suggestion of:
```
GIT_ROOT=$(git rev-parse --show-toplevel || pwd)
```
It's a nice trick, but this works too.

jakubgs (approved):

yenda:
@PombeirP seems like your request for changes has been fulfilled?

pedropombeiro (approved):

churik:
Checked on Android 8.0 and IOS 11.4.1.
@MajorTomSec thank you for the contribution!

PR #8836, opened 2019-08-26
Use no backup directory

cammellos:
### Testing
Make sure files are under `no_backup` in android and `~/Library` in `ios`

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -114,6 +114,10 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
 
     }
 
+    public String getNoBackupDirectory() {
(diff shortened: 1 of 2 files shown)


pedropombeiro (approved):

churik:
@cammellos any clue where I can find `no_backup` on Android?
on my Android 8 (LG V20) in  `/sdcard/Android/data/im.status.ethereum.pr/files ` is empty.

On IOS all files are under `~/Library/im.status.ethereum`

cammellos:
@churik unless the device is rooted you probably can't check it, unless you
run it in the emulator

churik:
@cammellos running on emulator, still can't see data.
Can you point me where I should check?

cammellos:
i am not home now, but it should be under
 /data/user/0/im.status.ethereum.pr/no_backup or something similar, not under /files, also you need to have created at least one account, just in case

churik:
Thank you, found it. Can confirm that it exists.

PR #8839, opened 2019-08-26
don't go into android dir, copy results if not linking

jakubgs:
Fix for wrong path of `result` folder when building inside of `android` directory.

Now it just fails when uploading due to lack of 64bit build:
```
Google Api Error: releaseValidationErrorKeyNotCompliantWith64BitRequirementError: Release is not compliant with the Play 64-bit requirement. - Release is not compliant with the Play 64-bit requirement.
```
https://ci.status.im/job/status-react/job/combined/job/mobile-android/11687/

### DIFF
--- a/ci/nix.groovy
+++ b/ci/nix.groovy
@@ -59,7 +59,7 @@ def build(Map opts = [:]) {
   opts.args = defaults.args + opts.args
   opts.keep = (opts.keep + defaults.keep).unique()
 
-  return sh(
+  def resultPath = sh(
     returnStdout: true,
     script: """
       set +x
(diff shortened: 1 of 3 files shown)


ci/nix.groovy
```diff
@@ -68,6 +68,10 @@ def build(Map opts = [:]) {
       nix-build ${_getNixCommandArgs(opts, false)}
     """
   ).trim()
+  if (!opts.link) { /* if not linking copy results */
```
pedropombeiro: nit, I think this reads better:
```suggestion
  if (!opts.link) { /* if not linking, copy results */
```

pedropombeiro (approved):

flexsurfer (approved):

PR #8841, opened 2019-08-27
Don't use webview and remove unused npm packages

flexsurfer:
removed
react-native-svg
react-native-svg-transformer
react-native-webview
react-native-http-bridge


vitvly (approved):

vitvly:
Question: seems that webview-bridge and qrcode packages depend on webview too. Do we want to keep the webview dep in our code?

flexsurfer:
@siphiuel no we don't, webview-bridge doesn't depend on webview , don't know about qrcode , maybe . we could use another qrcode, i don't see the reason to use webview for qrcode that's why it's so slow?

vitvly:
@flexsurfer regarding qrcode - had to make a small change https://github.com/siphiuel/react-native-qrcode/commit/26fc0e4dc31e7e0ea277337e32604867162e53d4 in order to get Hermes build with rn 0.60 running for me

flexsurfer:
hm qrcode depends on webview but it isn't supported anymore, so we can't use qrcode library with latest RN

vitvly:
Actually we can, just need to upgrade deps. Had to use a fork as i don't have permissions to status-im/react-native-webview.

flexsurfer:
> @flexsurfer regarding qrcode - had to make a small change [siphiuel/react-native-qrcode@26fc0e4](https://github.com/siphiuel/react-native-qrcode/commit/26fc0e4dc31e7e0ea277337e32604867162e53d4) in order to get Hermes build with rn 0.60 running for me

yes, so better to find another lib, because using webview for qrcode doesn't look good from performance pov, and also we have this glitch then it takes time to render qrcode and UI looks bad

dmitryn (approved):

flexsurfer:
btw from security pov it's better to have our own qr code generator because it's very critical part

flexsurfer:
ok this one looks better https://github.com/soldair/node-qrcode, we could generate image/png and use react/image component to render should be faster

flexsurfer:
@PombeirP could you also remove `react-native-http-bridge` please

pedropombeiro:
> @PombeirP could you also remove `react-native-http-bridge` please

Done

PR #8847, opened 2019-08-28
[#8321] Change or remove back button from DApp browser screen

### LINKED ISSUE #8321: Change or remove back button from DApp browser screen
# Problem
I was using the Teller Network DApp within the Status browser. As I was tapping specific sellers, and trying to go "back" to the main screen - I kept accidentally pressing the browser back button on the top left. This was very frustrating, because I kept having to re-open the DApp even though I just wanted to go back to the previous screen.

I believe this is a flaw in the placement of the back button, and would recommend we replace (or even remove) the back button.

## Notes
Here is a screenshot of the button. Any thoughts for improvements to this? From a timeframe perspective, considering this is primarily client work, we can do it after the code freeze.
IMG!UNITO-UNDERSCORE!6360

flexsurfer:
fixes #8321

### DIFF
--- a/src/status_im/ui/screens/browser/views.cljs
+++ b/src/status_im/ui/screens/browser/views.cljs
@@ -58,10 +58,10 @@
   [toolbar.view/toolbar
    {:browser? true}
    [toolbar.view/nav-button
-    (actions/back (fn []
-                    (re-frame/dispatch [:navigate-back])
-                    (when error?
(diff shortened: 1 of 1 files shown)


dmitryn (approved):

churik:
Button is changed to "Close" (Android  8 and IOS 11.4.1.)

PR #8858, opened 2019-08-29
Move from componentWillMount to componentDidMount

pedropombeiro:

### Summary

This PR migrates code from using the deprecated `componentWillMount` to using `componentDidMount`. It probably introduces some visual oddities as things are initialized later, but we can either leave that for fixing later or address it in this PR (although help will be needed in that case).

### Review notes

I'd appreciate a sanity check, as the replacements I did were mostly mechanical and might not make the most sense.

#### Platforms

- Android
- iOS

Test navigating across the different screens of the app, namely wallet, connectivity warning, situations where keyboard should show up automatically, etc.

status: ready

### DIFF
--- a/src/status_im/android/core.cljs
+++ b/src/status_im/android/core.cljs
@@ -26,8 +26,8 @@
 (defn app-root [props]
   (let [keyboard-height (subscribe [:keyboard-height])]
     (reagent/create-class
-     {:component-will-mount
-      (fn []
+     {:component-did-mount
+      (fn [this]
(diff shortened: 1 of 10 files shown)


flexsurfer (approved):

yenda (approved):

Serhy:
Tested with Android 8.1, Android 9, iOS 12 different screens of app where keyboard (device and numeric keyboard on Wallet), mobile connectivity, - no regressions found.

Issue #8708, opened 2019-08-06
App crash when open Status after it was put on background with Android hardware 'Back' button


Serhy:
***Type***: Bug

***Summary***: When user taps upon Android hardware 'Back' button being on Chats main view Status is put on the background. However, opening Status again results in an app crash with error `java.lang.illegalArgumentException: No view found for id 0x5ff (unknown) for fragment ScreenFragment{478dd37 #0 id=0x5ff}` 

<img width="352" alt="Screenshot 2019-08-06 at 12 05 50" src="https://user-images.githubusercontent.com/8749671/62527372-aca00a80-b843-11e9-9ed8-b90446eeea95.png">

#### Expected behavior

User end up on Chats view after opening Status if it was put on background with Android hardware 'Back' button before

#### Actual behavior

Status crashes after open if it was put on background with Android hardware 'Back' button before

### Reproduction

- Open Status and create new account
- Staying on Chats main view tap Android hardware `Back` button
- Open Status from the list of running apps (or via Status icon in the list of apps)

### Additional Information

- Status version: Release/Develop 0.14 (
- Operating System: Android

#### Logs

```
08-06 13:23:22.738   354   354 W ResourceType: No package identifier when getting name for resource number 0x00000629
08-06 13:23:22.739   354   354 E FragmentManager: No view found for id 0x629 (unknown) for fragment ScreenFragment{dcd9ea5 #0 id=0x629}
08-06 13:23:22.739   354   354 E FragmentManager: Activity state:
08-06 13:23:22.739   354   354 D FragmentManager:   Local Activity 86d8a26 State:
08-06 13:23:22.740   354   354 D FragmentManager:     mResumed=true mStopped=false mFinished=false
08-06 13:23:22.740   354   354 D FragmentManager:     mChangingConfigurations=false
08-06 13:23:22.740   354   354 D FragmentManager:     mCurrentConfig={0 1.0 themeSeq = 0 showBtnBg = 0 255mcc1mnc [en_US,ru_UA,uk_UA] ldltr sw411dp w411dp h707dp 420dpi nrml long port finger -keyb/v/h -nav/h appBounds=Rect(0, 0 - 1080, 1920) s.10}
[...]

PR #8868, opened 2019-08-29
Comment away react-native-screens

vitvly:
Temporarily comments away react-native-screens for Android.

https://github.com/status-im/status-react/issues/8708 might be caused by https://github.com/kmagiera/react-native-screens/issues/54

### DIFF
--- a/src/status_im/android/core.cljs
+++ b/src/status_im/android/core.cljs
@@ -43,7 +43,8 @@
                         (when-not (zero? @keyboard-height)
                           (dispatch [:set :keyboard-height 0]))))
         (.hide react/splash-screen)
-        (.useScreens rn-dependencies/react-native-screens)
+        ;; TODO Temporarily comment away due to current bug https://github.com/kmagiera/react-native-screens/issues/54
+        ;(.useScreens rn-dependencies/react-native-screens)
         (.addEventListener react/app-state "change" app-state-change-handler)
         (.addEventListener rn-dependencies/react-native-languages "change" on-languages-change)
         (.addEventListener rn-dependencies/react-native-shake


flexsurfer (approved):

yenda:
Looks like it is not up to date, btw the issue you link mentions that moving useScreen to component-did-mount can potentially fix things is that the case now that it has been done?

vitvly:
@yenda haven't tried that approach as it wasn't consistently working for everyone.

yenda:
@siphiuel I just checked and moving it to component-did-mount didn't fix it so it looks like this fix is needed. Please add a TODO before the comment

src/status_im/android/core.cljs
```diff
@@ -43,7 +43,8 @@
                         (when-not (zero? @keyboard-height)
                           (dispatch [:set :keyboard-height 0]))))
         (.hide react/splash-screen)
-        (.useScreens rn-dependencies/react-native-screens)
+        ; TODO Temporarily comment away due to current bug https://github.com/kmagiera/react-native-screens/issues/54
```
yenda: use double `;` for comments

vitvly: Done.

yenda (approved):

jakubgs:
The iOS signing issue has been fixed, so the build is green.

Serhy:
Can't reproduce initial issue #8708 anymore with Android 6.0.1, 8.1, 9 
Everything is ok with the chats and navigation through major screens (creating public chat, messaging, put app background->enable 3G connection -> app foreground = bottom sheet appears, etc)
Failed tests are not this PR issue and retested to confirm they are pass.

PR #8870, opened 2019-08-29
Fix high / critical e2e after profile redesign

churik:
Fixed:
1) Need help section in Profile
2) Backup seed warning shown in Wallet if account has funds
3) Check there's no seed phrase in logcat when making back up of the passphrase
4) Check sync of username, photo, public and group chats

Deleted AboutButton(BaseButton) class - it was declared twice

### DIFF
--- a/test/appium/tests/atomic/account_management/test_profile.py
+++ b/test/appium/tests/atomic/account_management/test_profile.py
@@ -163,6 +163,8 @@ class TestProfileSingleDevice(SingleDeviceTestCase):
         if sign_in_view.profile_button.counter.text != '1':
             self.errors.append('Profile button counter is not shown after re-login')
         sign_in_view.profile_button.click()
+        profile_view.privacy_and_security_button.click()
+        profile_view.backup_recovery_phrase_button.click()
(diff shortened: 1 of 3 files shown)


test/appium/tests/atomic/account_management/test_profile.py
```diff
@@ -223,7 +226,7 @@ def test_need_help_section(self):
         base_web_view.find_full_text('Frequently Asked Questions')
         base_web_view.click_system_back_button()
         profile_view.submit_bug_button.click()
-        profile_view.find_full_text('#status')
+        profile_view.wait_for_element_starts_with_text('Welcome to Gmail')
```
Serhy: Here we need to update not the check but above button (on to 'Request a feature') which redirects to '#status' channel.
For the submit_bug_button (old red 'Report bug' button) we have separate test case (which should be updated as well in TestRail btw)

churik: you are right, fixed

Serhy (approved):

PR #8879, opened 2019-08-30
Fix issue #8846 - Update terminology for consistency

### LINKED ISSUE #8846: Copy updates for consistency
In V1 we transition to some new terminology. On most screens, this new copy is included in the latest develop builds. However, some places that have not been touched for V1 might still include old copy that needs to be updated. This includes:
- Permission dialogs
- Permission settings
- Ohter screens. This issue requires a review of other places where this terminology is not consistent.

#### Changes
- PIN > `6-digit passcode`
- 6-digit code >  `6-digit passcode`
- Contact code > `Chat key`
- Recovery phrase >  `Seed phrase`
- Sign in > `Unlock`
- Recover or Add existing account > `Access key`
- Create new account > `Generate a new key`
- Other accounts > Back key without reference
_Note: Inside wallet we do use `Generate a new account`. We only refer to account in relation to a multiaccount in wallet. Never in relation to the 'thing you unlock' when opening Status._

### Why change the language
We have introduced layers of abstraction over time (Public key and Whisper ID becoming contact code. Seed phrase becoming Recovery phrase). The abstractions were intended to make the app easier to understand by moving away from the technical lingo and giving an indication of functionality, i.e. 'contact' and 'recover'. However, these abstractions:

- Have created inconsistency between code base and application, leading to miscommunication in product development.
- Have not been shown effective in explaining key concepts (e.g. contact code explaining that it is required to connect with contacts).
- Give too little information about how the app actually works.
- Contribute to continued misunderstanding about how the app works which becomes an issue when something goes wrong or new features are introduced.
- Add a layer of complexity to some of our core users who are familiar with the technology and might attempt to reverse engineer how the app works.

## Implementation
- Change in https://github.com/status-im/status-react/blob/develop/translations/en.json
- Reserve time for translations

## Acceptance Criteria
- Copy is updated on permission dialogs
- Copy is updated on permission in Proofile > Settings
- List of reviewed flows to check for remaining inconsistencies on latest develop.
- all occurrences of PIN and 6-digit passcode are changed to 6-digit passcode

MajorTomSec:
Fixes #8846 

### Summary

This PR aims to solve conflicts between terminology usages, as seen in issue #8846.
Changes have been made to `translations/en.json` only.

I took into account @rachelhamlin [comment](https://github.com/status-im/status-react/issues/8846#issuecomment-526645032) on capitalization of nouns. 
### Review notes
Please ensure the following changes are as expected:

* Recover or Add existing account > Access key
* Create new account > Generate a new key

I was not sure if using those terms properly when making the changes.

Obviously, all other changes need to be reviewed but I believe they are correct.
### Testing notes

No conflict in terminology should appear when using Status.

#### Platforms

Any.

status: ready


flexsurfer (approved):

translations/en.json
```diff
@@ -339,9 +339,9 @@
     "cryptokitty-name": "CryptoKitty #{{id}}",
     "network-id": "Network ID",
     "connection-problem": "Messages connection problem",
-    "contact-code": "Contact code",
+    "contact-code": "chat key",
```
churik: Started from capital letter, this shouldn't be changed IMO
so "Chat key"

translations/en.json
```diff
@@ -447,7 +447,7 @@
     "currency-display-name-bnd": "Brunei Darussalam Dollar",
     "remove-from-chat": "Remove from chat",
     "wallet-backup-recovery-description": "This will help you to keep your assets safe",
-    "recovery-phrase": "Recovery phrase",
+    "recovery-phrase": "seed phrase",
```
churik: "Seed phrase"

translations/en.json
```diff
@@ -34,10 +34,10 @@
     "delete-and-leave-group": "Delete and leave group",
     "new-group-chat": "New group chat",
     "edit-chats": "Edit chats",
-    "sign-in": "Sign in",
+    "sign-in": "unlock",
```
churik: "Unlock"

translations/en.json
```diff
@@ -700,11 +700,11 @@
     "soon": "Soon",
     "wallet-asset": "Asset",
     "close-app-content": "The app will stop and close. When you reopen it, the selected network will be used",
-    "logout-app-content": "The multiaccount will be logged out. When you log in again, the selected network will be used",
+    "logout-app-content": "The multiaccount will be logged out. When you unlock again, the selected network will be used",
```
churik: "The multiaccount will be logged out. When you unlock it again, the selected network will be used" - it shows that IT is multiaccount.

translations/en.json
MajorTomSec: I removed the capital letters as requested by @rachelhamlin [here](https://github.com/status-im/status-react/issues/8846#issuecomment-526645032).
I can revert this, but I agreed with her in the first place that this makes it odd in the middle of sentences. 
Can you confirm your request ?

translations/en.json
MajorTomSec: Agreed, but as stated in the original issue, `We only refer to account in relation to a multiaccount in wallet. Never in relation to the 'thing you unlock' when opening Status.`
Doesn't that mean we still refer to the "multiaccount" in adequate places, such as this one ? Or maybe I did not get your comment right.

translations/en.json
churik: I think @rachelhamlin meant inside the sentences. I'll ask her again, thanks!

translations/en.json
churik: I would prefer to replace this even with "The account will be logged out. When you unlock it again, the selected network will be used" - it sounds mostly understandable from my POV. Will confirm and get back to this ASAP.

translations/en.json
rachelhamlin: Yes, thanks guys - I would write it as follows:

`Send me your chat key.`
`Chat keys are great.` :) 

Essentially, don't treat it as a proper noun. If UX disagrees on this, well - they're on vacation so we can hash it out later 😂

@churik @MajorTomSec

bitsikka: btw, fwiw, I remember there being "Chat key" with key "chat-key" used in `share` pop-up in Profile tab which I worked on. Which together with this would be kind of duplicate. Translators are going to be like why do we have to translate the same phrase twice?

It is possible to have only one and handle sentence-casing/small-casing in code, but I don't know what is wise. Also, I don't know what happens with `casing` functions with different languages.

churik:
OK, based on discussion - @MajorTomSec  can you please replace to capital letter in all mentioned places and replace "The multiaccount will be logged out. When you unlock it again, the selected network will be used" to  "The account will be logged out. When you unlock it again, the selected network will be used"?
Thanks!

MajorTomSec:
Just pushed the requested changes by @churik !

EDIT: Oops, pushed some mistakes at the same time, fixing.

EDIT2: Fixed.

PR #8888, opened 2019-09-02
fix subtitle in list items

yenda:
arguments were missing in reagent-render args so component wasn't reacting to changes

## Testing

Tested on Android emulator
By default I was always seeing `Available only on Mainnet and Testnet` on ENS username item in profile, now I see the proper subtitle

status: ready

### DIFF
--- a/src/status_im/ui/components/list_item/views.cljs
+++ b/src/status_im/ui/components/list_item/views.cljs
@@ -134,12 +134,12 @@
    (when title-row-accessory
      [react/view styles/title-row-accessory-container title-row-accessory])])
 
-(defn subtitle-row [{:keys [subtitle subtitle-max-lines subtitle-row-accessory]}
-                    icon? theme]
+(defn subtitle-row
+  [subtitle-row-elements icon? theme]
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

bitsikka:
awesome!!! Thankyou @yenda somewhere in the back of my mind it was nagging me why the chat list subtitle was not re-rendering, when, the same sub components as before was re-used.

This totally explains it. So liberating to have reasons :)

So this means we did not need for subtitle component to subscribe its own subscription 😅 to re-render as needed.

Thankyou also for going easy on me :)

flexsurfer:
@bitsikka could you please revert subscriptions then? :)

bitsikka:
> @bitsikka could you please revert subscriptions then? :)

of course 👍 
on it

yenda:
> Thankyou also for going easy on me :)

No worries that is probably the number one mistake people do with reagent, we've all done it a few times.

churik:
Tested on IOS 11.4.1. and Android 8 - on mainnet and ropsten "Register a universal username....", on other networks "Only available on Mainnet and Ropsten"

PR #8890, opened 2019-09-03
revert unnecessary :chats/last-message-content subscription

bitsikka:
fixes issue discussed here https://github.com/status-im/status-react/pull/8888#issuecomment-527311963

better to merge after #8888 without which this will not work

### Testing notes
- In home screen, with this, last message preview and timestamp should still continue to be updated correctly as expected, as chat progresses with new messages

status: ready

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -716,16 +716,6 @@
    unviewed-messages-count))
 
 (re-frame/reg-sub
- :chats/last-message-content
- (fn [[_ chat-id]]
-   (re-frame/subscribe [:chats/chat chat-id]))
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

churik:
@bitsikka 
no preview for last message is shown.
Probably rebase may help?

bitsikka:
> @bitsikka
> no preview for last message is shown.
> Probably rebase may help?

@churik 
indeed that is probably the reason
rebased 👍

churik:
Tested on Android 8 and IOS 11.4.1 - a preview of all types of chats.

PR #8898, opened 2019-09-03
Use react-native net-info community package

pedropombeiro:

### Summary

This PR is an offshoot of the broader RN 0.60.x upgrade PR. It moves the app to the new community net-info package, which is removed in 0.60.

#### Platforms

- Android
- iOS

Offline<->online transition handling in the app.

status: ready


android/app/build.gradle
```diff
@@ -255,6 +255,7 @@ android {
 }
 
 dependencies {
+    implementation project(':@react-native-community_netinfo')
```
vitvly, line 258: A question - so there was a problem with resolving scoped module paths that requires this underscore symbol replacement?

vitvly (approved):

android/app/build.gradle
flexsurfer: A question - why does it have "@" ?

mobile_files/package.json.orig
```diff
@@ -6,6 +6,7 @@
     "start": "node node_modules/react-native/local-cli/cli.js start"
   },
   "dependencies": {
+    "@react-native-community/netinfo": "^3",
```
yenda: remove the ^

android/app/build.gradle
pedropombeiro: it's part of the package name, isn't it? This was added by `react native link`

pedropombeiro: > A question - so there was a problem with resolving scoped module paths that requires this underscore symbol replacement?

I imagine this is a Maven/Gradle limitation, so while the node package uses `/`, it needs to be replaced with an underscore in the Gradle files.

mobile_files/package.json.orig
pedropombeiro: Having the ^ here allows upgrading to a newer 3.x release on-demand by running `yarn update @react-native-community/netinfo`. Upgrades will not happen otherwise because we normally pass `--frozen-lockfile` to `yarn install`. I'm not saying what you suggest is wrong, but I'd like to understand the scenario we're trying to guard against.

pedropombeiro:
Any ideas why https://ci.status.im/blue/organizations/jenkins/status-react%2Fprs%2Fandroid/detail/PR-8898/3/pipeline is failing accessing `js-dependencies/net-info`?

yenda:
@PompeirP there is a js-dependencies namespace in tests for that don't do it in the regular one

churik:
@flexsurfer @yenda did your comments be addressed?
If yes, I'll start testing

pedropombeiro:
@churik I did address the comments, so if the e2e tests are good, then it's ready for testing.

churik:
All failed e2e are not related to this PR.
Checked on IOS 11.4.1, Android 8:
- сan create/restore account offline
- can logout/login offline
- send a message and sign transaction buttons are disabled while offline
- can resend messages
- fetching messages from offline for 1-1, public, group chats
- receiving invites to group chats from offline
- search chats from offline 
- the browser shows info about no Internet connection if offline
- can receive PN when offline -> online
- connect to a custom mailserver
- can advertise device when offline -> online
- devices sync (offline -> online)

yenda:
@churik does it mean we can merge?

churik:
I'm waiting for someone with 3G connection on test devices who can help to check sync with mobile data. Already pinged @Serhy and @asemiankevich in dm.

Serhy:
With devices set to 3G connection can receive message sent while device was offline in 1-1/Public/Group, can receive PN from offline->connect to 3G, accounts are synced between both devices in 3G mode. When syncing in mobile mode set to 'Stop sync' - no offline history retrieved as expected.
Ready for merge then!

PR #8912, opened 2019-09-04
list-item suppliment/fixes

bitsikka:
In kind of a response to this comment by @errorists https://github.com/status-im/status-react/issues/8070#issuecomment-521203445

> likely something else that I forgot but that there makes up the majority of our UI.

fixes/adds these:

- fix `title` being squished by looong text accessory - it happens in profile>sync settings>mailserver list-item(some of the mailserver names are quite long). I think it is a good idea to limit in general, the length of text-accessory(about 2/3 of device width seems to work best).
  Note: there has always been an option for using component accessory for cases when longer than 2/3 accessory is needed - like transaction signing bottom-pop-up for example
- have `react-key` key - useful for homogeneous list-items like `chat-list/asset-list/list-with-radio-buttons` etc. Default key generated by handy `gensym` CLJS core utility.
- add `:selectable` `:theme` together with `:selected?` config key for list-items with **radio** accessory
- bring `list-item/divider` within `list-item` by having a `type` called `divider`
  - prevents having to require list-item namespace just for that
  + keeps data fed to flat-list with list-items as clean as possible
- docs in code updated accordingly
- screens with instances of list-items updated accordingly

Status: ready


src/status_im/ui/components/list_item/views.cljs
```diff
@@ -340,19 +358,28 @@
...
 ;; boolean - false by default - self explanatory
 
+;; selected?
+;; boolean
+;; when (= :theme :selectable) this switch controls whether the
+;; list-item is in a visually selected state (background-color
+;; of list-item is colors/gray-selected
+;; useful for selectable list-items like list with radio buttons
+
 (defn list-item
```
yenda, line 217: could you move all these comments in the docstring if that seems relevant?

bitsikka: yes indeed - forgot about docstring 👍

bitsikka: done 👍 
followed this convention https://github.com/bbatsov/clojure-style-guide#documentation

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -245,114 +247,144 @@
...
-           title-accessibility-label subtitle subtitle-max-lines
-           subtitle-row-accessory content accessories on-press
-           on-long-press error accessibility-label disabled?]
-    :or   {type                    :default
-           theme                   :default
-           disabled?               false
-           container-margin-top    0
-           container-margin-bottom 0
-           subtitle-max-lines      1}}]
+  "A general purpose Status-React specfic list item component.
```
flexsurfer: status-react

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -245,114 +247,144 @@
...
+  Every key is optional. Use as needed.
+  Combination of around 4 related keys are enough for most cases.
+  Spec: https://www.figma.com/file/cb4p8AxLtTF3q1L6JYDnKN15/Index?node-id=787%3A1108
+
+  `react-key`
```
flexsurfer, line 223: is it used somewhere? i believe we use metadata for that `^{:key ""}`

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -245,114 +247,144 @@
...
+  `type`
+  `:default` (default), `:small`, `:section-header`, or `:divider`
+  `:section-header`
+  Specifying only these is sufficient.
+  {:title <Section title> :type :section-header}
+  Optionally set `container-margin-top/bottom`
+  `:divider`
+  A simple common gray divider seen in various screens.
+  Simply use:
+  {:type :divider} and specify nothing else.
```
flexsurfer, line 239: btw it would be useful to have something like `:top-devider? bool` and `:bottom-devider? bool` options, to render divider on top or bottom of list-item

src/status_im/ui/components/list_item/views.cljs
```diff
@@ -364,26 +396,51 @@
...
+              (merge title-row-elements
+                     {:title               title
+                      :title-prefix        title-prefix
+                      :title-row-accessory title-row-accessory})
+              subtitle-row-elements
+              {:subtitle               subtitle
+               :subtitle-max-lines     subtitle-max-lines
+               :subtitle-row-accessory subtitle-row-accessory}]
+          ^{:key react-key}
+          (case type
```
flexsurfer: `if` ?

flexsurfer (approved):

src/status_im/ui/screens/about_app/views.cljs
```diff
@@ -24,7 +24,7 @@
      {:type                :small
       :title               (i18n/label :t/version)
       :accessibility-label :version
-      :accessories         [(str app-version ";\n" node-version)]}]]])
+      :accessories         [(str app-version "; " node-version)]}]]])
```
flexsurfer: it's important to see full version, because there is no sense to see only part of version, so maybe we should use here custom component to show version , don't think we should tune list-item for that, or i've a better idea to show app-version and node-version in different items, that might work

bitsikka: I like `app-version` and `node-version` being shown using separate list-item 👍 

It was like that in old design. 

The only reason why it is in one at the moment, is because it was specified in figma like that.

bitsikka: @rachelhamlin can we please decide on the verbiage for each? "App version" and "Node version" seem obvious, but just want to make sure.

src/status_im/ui/components/list_item/views.cljs
bitsikka: definitely.. I was very very tempted to add `item-description` case, fairly common in unrealized-yet Profile subscreens :), but decided(I think for better) against it.

src/status_im/ui/components/list_item/views.cljs
bitsikka: Adding this is indeed a nitpick in a way :) Reasoning for having it in list-item component is similar to why `flat-list` component has a `keyExtractor` prop https://facebook.github.io/react-native/docs/flatlist#keyextractor.  Plus with this, we can even ignore it, since there is a sane default with `gensym`. And one less nesting of form in instances where  it is really needed.

src/status_im/ui/components/list_item/views.cljs
bitsikka: I am not sure. That sounds like more of a concern/prop for `flat-list`, `scroll-view`, or a plain `view` which wrap/contain list-items. In case of a `flat-list`, it has `header` and `footer` that could contain a divider.

Furthermore, I was thinking `divider` to serve more as section divider(keeping in mind that some sections in some screen spec does not have a `divider` but only white space. cc @andmironov @errorists 

If we add them, they could serve(maybe abused in semantical sense) as what a `border-top` or `border-bottom` on list-item would serve. So far I only see `transaction popup` implementation needing those. Everywhere else, list-item does not have a border. I figured not having border in list-item spec was intentional. cc @errorists

src/status_im/ui/components/list_item/views.cljs
bitsikka: 👍 will correct!

src/status_im/ui/screens/about_app/views.cljs
bitsikka: Oh and tuning `list-item` is not because of this specific case, but in general. I saw that in `Profile-tab>Sync settings>Mailserver` in iPhone X or Xs, if the chosen mailserver happens to be one of the long named ones, it squishes the `title`. I figured, generally it is good to have some whitespace separating `title` and `long text accessory`. cc @errorists.

Note: (personal observation) In latest `transaction popup` design whitespace between Title/accessories seems to be cutting close. I'm not sure what is going to happen in smaller width devices. cc @andmironov @errorists

andmironov: Great point @bitsikka regarding the list item in transaction overview modals. I'll add a rule for that (a fixed-width minimum margin ~6px between title and accessory + truncate the values if necessary)

src/status_im/ui/components/list_item/views.cljs
andmironov: Yep, I guess the transaction overview modal seems to be the only place at the moment with list items separated by dividers. Due to high information density, dividers are quite necessary there, but it's more of an exception and in the rest of the app we tend not to use dividers, or as @bitsikka said, use them as a way to separate sections (e.g in settings)

src/status_im/ui/screens/about_app/views.cljs
andmironov: Screen Shot 2019-09-05 at 11 32 34

rachelhamlin: @bitsikka to make sure I understand context, are you speaking of the "About" screen within profile? I am happy with "App version" and "Node version" if so. I see in [Figma](https://www.figma.com/file/bPS9GrvMr7LnH7vnkmuveQfd/Settings?node-id=415%3A153) that they are combined into one line. I'm happy either way, but would probably prefer separate items as well.

bitsikka: @rachelhamlin Yes, that is the one. Sounds good! separate items it is 👍

bitsikka: @andmironov I tried to handle the edge case automatically in the component but it got too complicated. To keep things simple, I went with what I had done already(bearing in mind we have alternative provision for edge cases already):

- Long stringified accessory has max-width of 62% of device width. Btw, `Node version` is still too long in a separate line of its own.
- That means `title` is also constrained to not be longer than 30ish%(considering hard right-margin in `title` of 16px when there is accessory)
- This will handle simple cases like the long `Mailserver` name gracefully enough.
- This does not mean edge case like the `transaction modal` has to confirm to that rule.
- For edge cases like that, an implementer can choose to use `component` as `accessory` and/or `title`. In which case there is still maximum freedom outside the bounds of `title`/`accessories` area of list-component.
- That means we are leaving it up to the implementer of edge-case like scenario to conform to edge-case spec you just made.

Note: It will be best also for designers of verbiage that go in list-item and elsewhere to be mindful about smaller width devices.

bitsikka:
- fixed actionably clear outstanding issues
- rebased

flexsurfer:
@bitsikka could you resolve conflicts please, i've made some fixes and changes in list-item, removed blue theme and fixed container styles, please tell if you need any help with that, thanks

bitsikka:
> @bitsikka could you resolve conflicts please, i've made some fixes and changes in list-item, removed blue theme and fixed container styles, please tell if you need any help with that, thanks

@flexsurfer 
I've resolved it locally, but I can't seem to launch dev build(iOS) to check for correctness. It builds fine even launches in emulator but gets stuck on Status logo screen. So I have not been able to progress on this or other issue I have committed to work on.

Also I noted that :blue theme related stuff is not completely removed, which I would clean-up if I could get dev build running on emulator.

So, I'm kind of waiting for flurry of changes in the pipeline 👍 🚀 , hoping that will fix the build problem 😄

bitsikka:
I'm pushing it anyway.. without clean-up to remeaining `:blue` theme stuff
consider it a wip
I think I can cleanup more about the :background-color in `styles/container` as well

bitsikka:
oopsie.. I can see problems in GH `files changed` tab.. fixing..

flexsurfer:
@bitsikka is it ready for testing?

bitsikka:
> @bitsikka is it ready for testing?

@flexsurfer 
Not yet
having this issue https://github.com/status-im/status-react/issues/8447#issuecomment-529375074

flexsurfer:
@bitsikka please rebase, should be fixed now

bitsikka:
> @bitsikka please rebase, should be fixed now

👍 Thanks! life saver :)
I will wrap this up first soon
and move on to the design touchup wallet ui stuff

bitsikka:
@flexsurfer 
- resolved conflict
- brought back `:min-height` in `styles/container` and gave reasons in the comment.
- cleaned up `:blue` theme handelings hanging around
- and other tiny cleanups

One more thing I need a decision on:
- for icon accessory can we just have list-item accept just :main-icons/icon and remove `:check` and `:more` option?
- if we argree on all `:main-icons/icon` accessory to be gray colored we can just keep the 3rd case and remove `:check` and `:more` cc @errorists 
https://github.com/status-im/status-react/blob/1a683a31a8b7e0924ade877425f5a1a78a23b41e/src/status_im/ui/components/list_item/views.cljs#L183-L193

I can think of 1 instance where `:main-icons/share` is being used - in `other people's profile`

this should keep the `:accessories` vector as clean as before, but more flexible.

errorists:
> for icon accessory can we just have list-item accept just :main-icons/icon and remove :check and :more option?

Yes we can! in design it's easy enough to override and on occasions we did just that like you pointed out in profiles. I imagine if possible, it would be nice to default to chevron as it's the vast majority of our use cases.

> if we argree on all :main-icons/icon accessory to be gray colored we can just keep the 3rd case and remove :check and :more cc @errorists

I think we can agree on all of them being 40% dark grey, yes. Regular dark grey is too much for something as secondary.

edit: so to sum up, in the component you can customise which icon is used in the right side accessory, the appearance is unified at 40% dark grey like we use for chevrons currently.

bitsikka:
> > for icon accessory can we just have list-item accept just :main-icons/icon and remove :check and :more option?
> 
> Yes we can! in design it's easy enough to override and on occasions we did just that like you pointed out in profiles. I imagine if possible, it would be nice to default to chevron as it's the vast majority of our use cases.
> 
> > if we argree on all :main-icons/icon accessory to be gray colored we can just keep the 3rd case and remove :check and :more cc @errorists
> 
> I think we can agree on all of them being 40% dark grey, yes. Regular dark grey is too much for something as secondary.

Cool! this will help keep `list-item` implementation towards lean

yes `:chevron` unlike other (24x24)icons is `10x16` and is such a special case, that I did not mention and kept :)

bitsikka:
I'm going to do this https://github.com/status-im/status-react/issues/8955#issuecomment-529562624 for `Mailserver` like `loong` content requiring `:type :small` `list-item` instances, and it should be ready

bitsikka:
> I'm going to do this [#8955 (comment)](https://github.com/status-im/status-react/issues/8955#issuecomment-529562624) for `Mailserver` like `loong` content requiring `:type :small` `list-item` instances, and it should be ready

Did this (works pretty good - in line with updated spec):
```clojure
[list-item/list-item
     {:type                :small
      :accessibility-label :node-version
      :title-prefix        :t/node-version
      :title
      [react/text
       {:number-of-lines 1
        :ellipsize-mode  :middle
        :style
        {:color        colors/gray
         :padding-left 16
         :text-align   :right
         :line-height  22}}
       node-version]}]
```
Kept 62% of device width limit on `Accessory text` - can still hack/by-pass-limit with `component` as accessory or like above

For list-items with long accessories, better to hack `list-item` like this than put it in `accessories` vector because of the implementation rationale described here https://github.com/status-im/status-react/issues/8955#issuecomment-529556821.

fin :)

churik:
@flexsurfer @yenda please let me know when this one will be ready for testing

flexsurfer:
@bitsikka is it ready?

bitsikka:
> @bitsikka is it ready?
@flexsurfer @churik 
Yes, it is ready

bitsikka:
Sorry, didn't post any testing notes

## Testing notes
- The supplementary features mentioned in OP above, are not manifested anywhere visually yet - no need for testing
- Test `Mailserver` title is not squished by `looong` mailserver name
- Test `Node version` title is not squished by `looong` node alpha-numeric version

churik:
awesome work!
Thank you @bitsikka !

PR #8913, opened 2019-09-04
fix profile photo icon - no border in non-identicon photo

bitsikka:
Small nitpick
- while fixing chat-list list-items I noticed in figma there was no border in profile pic with user chosen photo, but only on identicon
- visible in photos with light background
- seemed like it was done for a reason, like maybe allowing for users to get creative with borderless profile photo, maybe to stand out from the rest

- takes advantage of the fact that in current implementation identicons are saved as PNG and user chosen photos are saved as JPEG

Status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/photos.cljs
+++ b/src/status_im/ui/screens/chat/photos.cljs
@@ -2,18 +2,21 @@
   (:require [clojure.string :as string]
             [status-im.ui.components.react :as react]
             [status-im.ui.screens.chat.styles.photos :as style]
+            [status-im.ui.screens.profile.db :as profile.db]
             [status-im.utils.identicon :as identicon]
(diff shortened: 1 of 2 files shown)


flexsurfer:
@andmironov @errorists could you confirm?

errorists:
Yeah I can confirm, thanks for pinging @flexsurfer and @bitsikka for the keen eye!

In an ideal world, it would be based not on filetype but sampling the edge pixels if they are white when picking the profile photo. This is the closest we can get without fiddling with 3rd party libraries.

flexsurfer (approved):

bitsikka:
rebased

asemiankevich:
fails are not related to this PR, i think we can merge

PR #8914, opened 2019-09-04
integrate status-go mailserver and mailserver-topics apis

yenda:
integrate the mailserver api to store custom mailservers on status-go as well as the mailserver-topics api

I modified initialize-protocol so that it waits for mailservers, mailservers-topic and mailservers-ranges to be initialized before initializing whisper. Now only mailservers-ranges is still coming from realm but it will be trivial to integrate it once the service is available

##### Functional

- mailservers

### Steps to test

changes are not platform dependent so it is not a big deal if iOS doesn't build

I tested the following scenario on Android emulator:

- create account
- login. select mailserver manually
- logout, login
- add mailserver manually, connect to it.
- logout, login, selected mailserver should still be selected

Note that when you select your mailserver, the fields appear empty in the details, but that is not the case, it's a bug with text input unrelated to this PR. Additionally, once you selected your own mailserver, it will have a very long id and will hide mailserver label in the settings, also known bug.

you can use this custom mailserver:

frame

enode://531e252ec966b7e83f5538c19bf1cde7381cc7949026a6e499b6e998e695751aadf26d4c98d5a4eabfb7cefd31c3c88d600a775f14ed5781520a88ecd25da3c6:status-offline-inbox@35.225.227.79:30504

It is pretty hard to QA test mailserver topics. I checked in development that mailserver topics are properly retrieved after login by looking into the app-db and that there was no errors logged in adb logcat.

status: ready

### DIFF
--- a/src/status_im/data_store/core.cljs
+++ b/src/status_im/data_store/core.cljs
@@ -201,10 +201,11 @@
 
 (fx/defn handle-change-multiaccount-success
   {:events [::multiaccount-change-success]
-   :interceptors [(re-frame/inject-cofx :data-store/get-all-mailservers)
-                  (re-frame/inject-cofx :data-store/mailserver-topics)]}
-  [cofx]
-  (protocol/initialize-protocol cofx))
+   :interceptors [(re-frame/inject-cofx :data-store/mailserver-topics)
+                  (re-frame/inject-cofx :data-store/all-chat-requests-ranges)]}
+  [{:data-store/keys [mailserver-topics mailserver-ranges] :as cofx}]
+  (protocol/initialize-protocol cofx {:mailserver-topics mailserver-topics
+                                      :mailserver-ranges mailserver-ranges}))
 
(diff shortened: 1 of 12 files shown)


flexsurfer (approved):

Serhy:
Tested with Androids 8 and Android 9 and it looks good. Can set up custom mailserver, exchange with messages, messages are received when going back from offline to online.
Can change mailserver (`Pick another mailserver` option from pop-up when it fails to connect to mailserver) to one of default mailservers and after re-login it still set. Returning to 'Automatic selection' setting of mailservers picks random mailserver as usual.
Failed autotests are not this PR issue and will be fixed separately.

Not great that iOS can't be build (all 5 + manual were failed) but that's known as https://github.com/status-im/infra-ci/issues/7 
I think we are good to merge @yenda, if the changes are not platform dependent :)
We can fix iOS later if there are regressions

PR #8916, opened 2019-09-04
Preserve text-input's default value

### LINKED ISSUE #8910: default value for input fields doesn't work
# Problem

In various input fields across the app, the input fields takes a default value, for instance when adding an account or when editing a custom mailserver. Currently only the placeholder is shown and not the current value if there is any.

vitvly:
Fixes #8910 

This issue was introduced in https://github.com/status-im/status-react/pull/8625. Consider this scenario:
  1. User navigates from Home view to Chat view
  2. Inputs some text into message box, but does not send it
  3. Goes back to Home and then again to Chat
  4. The text entered at step 2 is missing, because it's being set via `default-value` prop, and it is erased in `on-will-blur` navigation events handler.

After adding logging it turns out that text-input's `onRef` handler is invoked before the `onWillBlur` for Home view in step 3. So the default-value gets cleared, as we are not tracking which inputs belong to which views. Restoring `default-value` via `setNativeProps` seems to resolve this issue.

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -126,14 +126,18 @@
             [text-class (dissoc options-with-style :nested?)]
             nested-text-elements)))
 
-(def text-input-refs (atom #{}))
+;; We track all currently mounted text input refs
+;; in a ref-to-defaultValue map
+;; so that we can clear them (restore their default values)
+;; when global react-navigation's onWillBlur event is invoked
+(def text-input-refs (atom {}))
 
 (defn text-input
   [options text]
   (let [input-ref (atom nil)]
     (reagent/create-class
       {:component-will-unmount #(when @input-ref
-                                  (swap! text-input-refs disj @input-ref))
+                                  (swap! text-input-refs dissoc @input-ref))
        :reagent-render
        (fn [options text]
          [text-input-class
@@ -143,11 +147,10 @@
              :placeholder-text-color   colors/text-gray
              :placeholder              (i18n/label :t/type-a-message)
              :ref                      (fn [r] 
-                                         (if r
-                                           (when (nil? @input-ref)
-                                             (swap! text-input-refs conj r))
-                                           (when @input-ref
-                                             (swap! text-input-refs disj @input-ref)))
+                                         ;; Store input and its defaultValue
+                                         ;; one we receive a non-nil ref
+                                         (when (and r (nil? @input-ref))
+                                           (swap! text-input-refs assoc r (:default-value options)))
                                          (reset! input-ref r)
                                          (when (:ref options) 
                                            ((:ref options) r)))
--- a/src/status_im/ui/screens/routing/core.cljs
+++ b/src/status_im/ui/screens/routing/core.cljs
@@ -40,8 +40,10 @@
     :on-will-blur
     (fn []
       (reset! screen-focused? false)
-      (doseq [text-input @react/text-input-refs]
-        (.clear text-input)))}])
+      ;; Reset currently mounted text inputs to their default values
+      ;; on navigating away; this is a privacy measure
+      (doseq [[text-input default-value] @react/text-input-refs]
+        (.setNativeProps text-input (clj->js {:text default-value}))))}])
 
 (defn wrap
   "Wraps screen with main view and adds navigation-events component"
--- a/src/status_im/ui/screens/add_new/new_chat/views.cljs
+++ b/src/status_im/ui/screens/add_new/new_chat/views.cljs
@@ -37,6 +37,9 @@
                                                   (re-frame/dispatch [:contact.ui/contact-code-submitted]))
                           :placeholder         (i18n/label :t/enter-contact-code)
                           :style               (add-new.styles/input @tw)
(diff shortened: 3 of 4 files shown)


yenda (approved):
Awesome thanks!

flexsurfer:
@siphiuel could please add a description, what the problem was and how it was fixed, thx

vitvly:
@flexsurfer yes sure, added a description.

yenda:
@siphiuel can you rebase?

vitvly:
@yenda done

churik:
@siphiuel would you mind to rebase it to check IOS as well?
Can't install previous due to signing problem

churik:
@siphiuel can you please confirm that  it should work only for chat input field?
Consider following scenario:
- tap on + > Join public chat
- enter topic
- switch to wallet 
- switch to home

I would prefer to see input in this case.
It is reproducible for all fields except chat input here.
On 0.13.2 input is preserved.
So I suppose that is an issue.

components/src/status_im/ui/components/react.cljs
```diff
@@ -126,14 +126,14 @@
             [text-class (dissoc options-with-style :nested?)]
             nested-text-elements)))
 
-(def text-input-refs (atom #{}))
+(def text-input-refs (atom {}))
```
yenda, line 133: please add comment explaining what this is for

src/status_im/ui/screens/routing/core.cljs
```diff
@@ -40,8 +40,8 @@
     :on-will-blur
     (fn []
       (reset! screen-focused? false)
-      (doseq [text-input @react/text-input-refs]
-        (.clear text-input)))}])
+      (doseq [[text-input default-value] @react/text-input-refs]
+        (.setNativeProps text-input (clj->js {:text default-value}))))}])
```
yenda, line 46: explain

components/src/status_im/ui/components/react.cljs
```diff
@@ -143,11 +143,8 @@
              :placeholder-text-color   colors/text-gray
              :placeholder              (i18n/label :t/type-a-message)
              :ref                      (fn [r] 
-                                         (if r
-                                           (when (nil? @input-ref)
-                                             (swap! text-input-refs conj r))
-                                           (when @input-ref
-                                             (swap! text-input-refs disj @input-ref)))
+                                         (when (and r (nil? @input-ref))
+                                           (swap! text-input-refs assoc r (:default-value options)))
```
yenda, line 153: explain

components/src/status_im/ui/components/react.cljs
```diff
@@ -126,14 +126,14 @@
...
             nested-text-elements)))
 
-(def text-input-refs (atom #{}))
+(def text-input-refs (atom {}))
 
 (defn text-input
   [options text]
   (let [input-ref (atom nil)]
     (reagent/create-class
       {:component-will-unmount #(when @input-ref
```
yenda, line 139: explain

vitvly:
@churik it must be because the chat title value is not stored in defaut-value property, will fix

components/src/status_im/ui/components/react.cljs
```diff
@@ -126,14 +126,18 @@
             [text-class (dissoc options-with-style :nested?)]
             nested-text-elements)))
 
-(def text-input-refs (atom #{}))
+;; We track all currently mounted text input refs
+;; in a ref-to-defaultValue map
```
yenda, line 130: put "`" "`" around `ref-to-defaultValue`

components/src/status_im/ui/components/react.cljs
```diff
@@ -126,14 +126,18 @@
...
+;; so that we can clear them (restore their default values)
+;; when global react-navigation's onWillBlur event is invoked
```
yenda, line 132: `onWillBlur`

yenda: `

components/src/status_im/ui/components/react.cljs
yenda: `

yenda: this symbol `

components/src/status_im/ui/components/react.cljs
```diff
@@ -143,11 +147,10 @@
              :placeholder-text-color   colors/text-gray
              :placeholder              (i18n/label :t/type-a-message)
              :ref                      (fn [r] 
-                                         (if r
-                                           (when (nil? @input-ref)
-                                             (swap! text-input-refs conj r))
-                                           (when @input-ref
-                                             (swap! text-input-refs disj @input-ref)))
+                                         ;; Store input and its defaultValue
```
yenda, line 150: defaultValue `

vitvly: Added.

vitvly:
@churik added defaultValue for inputs in `New Chat`/`New Public Chat` views

churik:
Tested that https://github.com/status-im/status-react/issues/8600 is not reproducible and that input is preserved for  `New Chat`/`New Public Chat` views, Dapps, chat input field

PR #8862, opened 2019-08-29
Remove community webview, replace qrcode

flexsurfer:
Removed community webview, replaced qrcode

fixes #8849

QA:
tested on ios and Android:
collectibles
qrcodes for wallet and profile

PR #8965, opened 2019-09-11
[8957] fix - Wallet address can't be copied by tapping on it

### LINKED ISSUE #8957: Wallet address can't be copied by tapping on it
### Description

*Type*: Bug
*Summary*: during wallet redesign possibility to copy code by tap (as it made for profile > chat key) is lost.

#### Expected behavior
can copy address

#### Actual behavior
nothing happens
xzcx

### Reproduction

- Open Status
- Create account
- Go to wallet > select Account
- Tap on receive
- Tap on wallet address 
### Additional Information

- Status version: nightly 09/09/2019

bitsikka:
fixes #8957 

Having worked on `copyable-text-view` with copied cue animation component during user.profile work, had always meant to fix this as a separate PR. Couldn't help getting this done :)

- made wallet address copyable just like user.profile share popup
- fixed bottom margin of `Share address` button
- fixed(in both wallet and profile share popover) qr code size that got shrinked during move from using webview for qr image to `qrcode` in #8862

Note: In profile, button did not have link icon anymore - haven't fixed that one, as it was mentioned in code-comment as TODO in remainder work for button component.

## Testing notes
- Test wallet tab > share pop-over
   - qr width is snug
   - Share address button bottom margin is correct
   - address is copyable with `Copied` visual cue
- Test User Profile tab > share pop-over
   - similar to above

status: ready

### DIFF
--- a/src/status_im/ui/components/qr_code_viewer/views.cljs
+++ b/src/status_im/ui/components/qr_code_viewer/views.cljs
@@ -1,21 +1,28 @@
 (ns status-im.ui.components.qr-code-viewer.views
-  (:require [reagent.core :as reagent]
+  (:require [cljs-bean.core :as bean]
+            [reagent.core :as reagent]
             [status-im.react-native.js-dependencies :as rn-dependencies]
             [status-im.ui.components.qr-code-viewer.styles :as styles]
             [status-im.ui.components.react :as react]
-            [status-im.ui.components.svg :as svg])
-  (:require-macros [status-im.utils.views :refer [defview letsubs]]))
+            [status-im.ui.components.svg :as svg]))
 
 (defn qr-code [{:keys [size value]}]
   (let [uri (reagent/atom nil)]
-    (.toString rn-dependencies/qr-code value #(reset! uri %2))
+    (.toString
+     rn-dependencies/qr-code
+     value
+     (bean/->js {:margin 0 :width size})
+     #(reset! uri %2))
     (fn []
       (when @uri
         [svg/svgxml {:xml @uri :width size :height size}]))))
 
-(defn qr-code-view [size value]
+(defn qr-code-view
+  "Qr Code view including the frame.
+  Note: `size` includes frame with `styles/qr-code-padding.`"
+  [size value]
   (when (and size value)
     [react/view {:style               (styles/qr-code-container size)
                  :accessibility-label :qr-code-image}
      [qr-code {:value value
-               :size  size}]]))
+               :size  (- size (* styles/qr-code-padding 2))}]]))
--- a/src/status_im/ui/screens/wallet/request/views.cljs
+++ b/src/status_im/ui/screens/wallet/request/views.cljs
@@ -1,20 +1,20 @@
 (ns status-im.ui.screens.wallet.request.views
   (:require [re-frame.core :as re-frame]
+            [reagent.core :as reagent]
             [status-im.ethereum.eip55 :as eip55]
             [status-im.ethereum.eip681 :as eip681]
             [status-im.ethereum.tokens :as tokens]
             [status-im.i18n :as i18n]
+            [status-im.ui.components.button :as button]
+            [status-im.ui.components.copyable-text :as copyable-text]
             [status-im.ui.components.qr-code-viewer.views :as qr-code-viewer]
             [status-im.ui.components.react :as react]
             [status-im.ui.components.styles :as components.styles]
+            [status-im.ui.components.toolbar :as toolbar]
             [status-im.ui.screens.wallet.components.views :as wallet.components]
             [status-im.ui.screens.wallet.request.styles :as styles]
-            [status-im.wallet.utils :as wallet.utils]
             [status-im.utils.utils :as utils]
-            [status-im.ui.components.colors :as colors]
-            [reagent.core :as reagent]
-            [status-im.ui.components.common.common :as components.common]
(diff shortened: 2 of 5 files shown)


src/status_im/ui/screens/wallet/request/views.cljs
```diff
@@ -62,28 +62,33 @@
...
+      [copyable-text/copyable-text-view
+       {:label           :t/ens-wallet-address
+        :container-style {:margin-top 12 :margin-bottom 4}
+        :copied-text     (eip55/address->checksum address)}
+       [react/text {:number-of-lines     1
+                    :ellipsize-mode      :middle
                     :accessibility-label :address-text
-                    :style           {:line-height 22 :font-size 15
-                                      :font-family "monospace"}}
+                    :style               {:line-height 22 :font-size 15
```
vitvly, line 78: I think we need not to provide line-height and font-size explicitly as it's the default typography. Actually, for `font-size` 15 we have `line-height` 21, but we shouldn't deviate from typography defaults. Not introduced in your PR anyways:) But maybe you can remove this explicit setting.

src/status_im/ui/components/qr_code_viewer/views.cljs
```diff
@@ -3,19 +3,25 @@
...
-  (:require-macros [status-im.utils.views :refer [defview letsubs]]))
+            [status-im.ui.components.svg :as svg]))
 
 (defn qr-code [{:keys [size value]}]
   (let [uri (reagent/atom nil)]
-    (.toString rn-dependencies/qr-code value #(reset! uri %2))
+    (.toString
+     rn-dependencies/qr-code
+     value
+     #js {:margin 0 :width size}
```
vitvly: I think we tend to use clj->js instead.

vitvly (approved):

src/status_im/ui/components/qr_code_viewer/views.cljs
flexsurfer: no,  why ? #js is better

src/status_im/ui/screens/wallet/request/views.cljs
bitsikka: I understand there is default typography spec implemented already. And there was an inquiry about overriding it in another bigger PR as well. The reason for the overriding of default typography implementation here is same as before - which is, it is specified in spec for share popover here https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=2796%3A22587.

In this one I did not leave a comment specifying reason for overriding, as in the other, because I'm beginning to suspect maybe there is a misunderstanding of what `typography spec` is for vs `spec for text elements in components`. @errorists please correct me if I'm wrong. As I understand it:
- Typography Spec - is for body text
- Text style spec for textual elements of components can override the Typography Spec if explicitly defined in spec for the component. There must've been a reason for overriding line-height in spec for popover or list-item.

Following those two rules still ensures `vertical rythm` which is the main concern behind having various `line-hight` spec.

Not following, or following only the first one can actually break `vertical rythm`. Because in the former there are padding/margin vertical centering etc to account for.

If this is true, better we update our collective mental-model about it.

Or maybe I'm taking figma spec too literally 😅 , believing everything specified there is not arbitrary, but have a reason like ensuring vertical rythm and such.

src/status_im/ui/components/qr_code_viewer/views.cljs
bitsikka: Understood. There is also `https://github.com/mfikes/cljs-bean` in `deps.edn` committed already. Which is supposed to be faster than `clj->js`. I think we're going to move over to `->js` and `->clj` provided by `cljs-bean` everywhere sometimes perhaps soon. Shall I use it instead?

src/status_im/ui/screens/wallet/request/views.cljs
bitsikka: Vertical rythm to me is how the negative space / white space flows along with visible elements to create kind of a rythm/beat. Also how in the srceenshot(iPhone Xs) below there is no clipping of content at the bottom. Even when it clips in devices of varying height and width it does so in a more or less sensible position.

image

src/status_im/ui/components/qr_code_viewer/views.cljs
flexsurfer: yes please

bitsikka: 👍 going to use `->js` and `->clj` for js interop going forward

vitvly: @flexsurfer i somehow missed `#js` usage across our codebase :( thanks @bitsikka for reminding about `cljs-bean`. Good stuff :)

bitsikka: did it.

Indeed good stuff :)
huge fan of mfikes's work, and every ounce of perf improvement matters.

have to really check doc/source though, maybe not needed for #js <col> 🤔 because cljs transpiler might already automatically keep that as native js data-structure and there myght be no conversion of ds involved.

I have to do some long chores today.. will look into doc/source later.

flexsurfer (approved):

churik:
Tested on IOS 11.4.1 (IPhone 7), Android 8 (LG V20)
Thank you @bitsikka for fixing this!

PR #8976, opened 2019-09-12
[DEBUG] memoize image

cammellos:


ghost:
### Pull Request Checklist

yenda:
Did this improve perfs significantly? I did it some time ago and it wasn't very convincing but maybe there was too much other perf issues back then

cammellos:
@yenda I checked with sehry, it's a 45% performance improvent if all the images are memoized, I don't think memoizing is a good strategy, but we can possibly pre-calculate them and save them

PR #8983, opened 2019-09-12
Fix/nix store delete

pedropombeiro:

### Summary

Don't break build if some Nix paths are still alive and can't be cleaned.

status: ready

### DIFF
--- a/nix/clean.sh
+++ b/nix/clean.sh
@@ -9,6 +9,7 @@ function getSources()   { nix-store --query --binding src "${1}"; }
 function getOutputs()   { nix-store --query --outputs "${1}"; }
 function getDrvFiles()  { nix-store --query --deriver "${1}"; }
 function getReferrers() { nix-store --query --referrers "${1}"; }
+function getRoots()     { nix-store --query --roots ${1} }
 
 function findRelated() {
     path="${1}"
@@ -26,10 +27,12 @@ function findRelated() {
     elif [[ -f "${drv}" ]]; then
         src=$(getSources "${drv}")
     fi
-    log " - Derivation: ${drv}"
-    log " - Source:     ${src}"
-    
-    found+=("${drv}" "${src}")
+    if [ $(getRoots "${drv}" | wc -l) -eq 0 ]; then
+        log " - Derivation: ${drv}"
+        log " - Source:     ${src}"
+        found+=("${drv}" "${src}")
+    fi
+
     printf '%s\n' "${found[@]}"
 }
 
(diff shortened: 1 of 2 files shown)


ghost:
`mobile/js_files/yarn.lock` changed. Pinging @jakubgs, @PombeirP, and @corpetty

ghost:
### Pull Request Checklist

pedropombeiro:
We should probably turn this whole script around and work from a `nix-store --gc --print-dead` point of view.

nix/clean.sh
```diff
@@ -26,10 +26,12 @@ function findRelated() {
     elif [[ -f "${drv}" ]]; then
         src=$(getSources "${drv}")
     fi
-    log " - Derivation: ${drv}"
-    log " - Source:     ${src}"
-    
-    found+=("${drv}" "${src}")
+    if [ $(nix-store --query --roots ${drv} | wc -l) -eq 0 ]; then
```
jakubgs: Can you make this into a method like the ones above? LIke `getGCRoots()`?

nix/clean.sh
```diff
@@ -73,13 +75,12 @@ if [[ -n "${nixResultPath}" ]]; then
...
 
 # remove duplicates and return
 toDelete=$(printf '%s\n' "${toDelete[@]}" | sort | uniq)
 
 log "Deleting..."
-nix-store --ignore-liveness --delete ${toDelete[@]}
-
-log "Deleting dead Android Gradle and NPM derivations..."
-nix-store --gc --print-dead | grep -E 'android-gradle-and-npm-modules(\.drv)?$' | xargs nix-store --delete
+nix-store --ignore-liveness --delete ${toDelete[@]} || exit 0
```
jakubgs: I don't think we need this, we already ignore `nix/clean.sh` failures from `nix/build.sh` for CI:
https://github.com/status-im/status-react/blob/4a4af538760a5d1848b54e7f63309e76f40d4be3/nix/build.sh#L13-L21

pedropombeiro: I thought this is what caused https://ci.status.im/job/status-react/job/combined/job/mobile-android-e2e/991/console to fail

jakubgs: OH! I think I know what is happening. We check for `JENKINS_URL` but its not available because we are running within `nix-shell` which doesn't inherit that env variable. That's why the clause that should ignore the failure is not triggered.

pedropombeiro: Yeah, we use `IN_CI_ENVIRONMENT` for that. I'll change that.

jakubgs: Maybe a better check would be to change it to check for `jenkins` username. The `USER` env variable is not affected by `nix-shell`.

jakubgs: Ah, using `IN_CI_ENVIRONMENT` is even better, forgot about that.

jakubgs (approved):

PR #8996, opened 2019-09-13
fix log level config

yenda:
@cammellos related to bootnodes, I added some docstrings to the important functions

the log level config for status-go wasn't saved in the node-config
as a result even though the setting looked applied in the UI it wasn't
in the node

Waiting for build to test

status: ready

### DIFF
--- a/src/status_im/node/core.cljs
+++ b/src/status_im/node/core.cljs
@@ -169,6 +169,10 @@
   (types/clj->json (get-multiaccount-node-config db)))
 
 (fx/defn save-new-config
+  "Saves a new status-go config for the current account
+   This RPC method is the only way to change the node config of an account.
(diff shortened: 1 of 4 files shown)


src/status_im/node/core.cljs
```diff
@@ -170,13 +170,18 @@
   (types/clj->json (get-multiaccount-node-config db)))
 
 (fx/defn save-new-config
+  "Saves a new status-go config for the current account
+   This RPC method is the only way to change the node config of an account.
+   NOTE: it is better used indirectly through `prepare-new-config`, which will take
```
yenda, line 174:
```suggestion
   NOTE: it has to be used indirectly through `prepare-new-config`, which will take
```

flexsurfer (approved):

PR #9002, opened 2019-09-14
Macros

yenda:
@flexsurfer please try it in dev and let's merge it

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -108,24 +108,16 @@
 (fx/defn add-public-chat
   "Adds new public group chat to db"
   [cofx topic]
-  (if config/use-status-go-protocol?
-    {::json-rpc/call [{:method "status_joinPublicChat"
-                       :params [topic]
-                       :on-success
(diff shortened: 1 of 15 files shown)


yenda:
@flexsurfer @Serhy I added another commit which might bring some nice perf improvements. I rewrote safe-merge so that it is only using reduce and conditionals instead of complex functions and at the function level it is 10 times faster. There is some places like when adding messages where it can be called 100 times in a row which on desktop used to take 180ms and now takes 19ms for simple test values, so the real results on real devices might be even better

flexsurfer:
image

yenda:
@flexsurfer yes it is fixed already in the latest commit, forgot special case for db

flexsurfer:
i use latest commit

yenda:
@jakubgs jenkins didn't start a build again on latest commit

yenda:
@flexsurfer @jakubgs lol I forgot to push

flexsurfer:
idk if its your PR or develop, but app is broken, there is no account in the wallet, and app is crashing on log out

flexsurfer (approved):

PR #9005, opened 2019-09-18
[#8968] custom seed phrase popover

### LINKED ISSUE #8968: User enters seed - add warning if word not in dictionary
# Situation

When a user imports a multiaccount by typing in his 12 words seed phrase:
- We *do* check the words against our dictionary. 
- Even if they don't match our dictionary, we currently allow the user to proceed without warning.
- If there is a mispelling or use of unsupported language, the user will end up with a new, empty account
- If the seed phrase IS valid, intentionally does not match our dictionary, and comes from another wallet, the user will see the funds from that account inside Status.

# Improvement

Upon detection of a word that is not in our dictionnary, we will show a warning pop-up.

This pop-up will thus show if there is a mispelling, or the user types a seed phrase in a language we don't support

# Definition of the popup 
WARNING : exact copy is not ok on figma or image below. Copy should be : 
"This looks like a custom seed phrase and doesn't match the Status dictionary. This could also mean some words are misspelled. If so, you'll end up creating a new account.
Continue
Cancel"

https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=927%3A14702

Onboarding_–_Figma

cc @rachelhamlin @andmironov @hesterbruikman

dmitryn:
fixes #8968

### Summary

Show "custom seed phrase" popover when recover with seed phrase containing words not from the dictionary.

Figma https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=4311%3A0&viewport=1143%2C-2667%2C0.7808234691619873

status: ready

### DIFF
--- a/src/status_im/ui/screens/multiaccounts/recover/views.cljs
+++ b/src/status_im/ui/screens/multiaccounts/recover/views.cljs
@@ -19,7 +19,44 @@
             [status-im.utils.utils :as utils]
             [status-im.constants :as constants]
             [status-im.ui.components.list-item.views :as list-item]
-            [status-im.utils.platform :as platform]))
+            [status-im.utils.platform :as platform]
+            [status-im.react-native.resources :as resources]
+            [status-im.ui.components.icons.vector-icons :as icons]
+            [status-im.ui.components.button :as button]))
+
+(defview custom-seed-phrase []
+  [react/view
+   [react/view {:margin-top 24 :margin-horizontal 24 :align-items :center}
+    [react/view {:width       32 :height 32 :border-radius 16
+                 :align-items :center :justify-content :center}
+     [icons/icon :main-icons/help {:color colors/blue}]]
+    [react/text {:style {:typography    :title-bold
+                         :margin-top    8
+                         :margin-bottom 8}}
+     (i18n/label :t/custom-seed-phrase)]
+    [react/view {:flex-wrap       :wrap
+                 :flex-direction  :row
+                 :justify-content :center
+                 :text-align      :center}
+     [react/nested-text
+      {:style {:color       colors/gray
+               :text-align  :center
+               :line-height 22}}
+      (i18n/label :t/custom-seed-phrase-text-1)
+      [{:style {:color colors/black}}
+       (i18n/label :t/custom-seed-phrase-text-2)]
+      (i18n/label :t/custom-seed-phrase-text-3)
+      [{:style {:color colors/black}}
+       (i18n/label :t/custom-seed-phrase-text-4)]]]
+    [react/view {:margin-vertical 24
+                 :align-items     :center}
+     [button/button {:on-press            #(re-frame/dispatch [::multiaccounts.recover/continue-pressed])
+                     :accessibility-label :continue-custom-seed-phrase
+                     :label               (i18n/label :t/continue)}]
+     [button/button {:on-press            #(re-frame/dispatch [:hide-popover])
+                     :accessibility-label :cancel-custom-seed-phrase
+                     :type                :secondary
+                     :label               (i18n/label :t/cancel)}]]]])
 
 (defn bottom-sheet-view []
   [react/view {:flex 1 :flex-direction :row}
@@ -37,7 +74,16 @@
         :title               :t/recover-with-keycard
         :disabled?           (not config/hardwallet-enabled?)
         :accessibility-label :recover-with-keycard-button
-        :icon                :main-icons/keycard-logo
+        :icon                [react/view {:border-width     1
+                                          :border-radius    20
+                                          :border-color     colors/blue-light
+                                          :background-color colors/blue-light
+                                          :justify-content  :center
+                                          :align-items      :center
+                                          :width            40
+                                          :height           40}
+                              [react/image {:source (resources/get-image :keycard-logo-blue)
+                                            :style  {:width 24 :height 24}}]]
         :on-press            #(re-frame/dispatch [::hardwallet/recover-with-keycard-pressed])}])]])
 
 (def bottom-sheet
@@ -76,7 +122,6 @@
         [text-input/text-input-with-label
          {:on-change-text    #(re-frame/dispatch [::multiaccounts.recover/enter-phrase-input-changed (security/mask-data %)])
           :auto-focus        true
-          :on-submit-editing #(re-frame/dispatch [::multiaccounts.recover/enter-phrase-input-submitted])
           :error             (when passphrase-error (i18n/label passphrase-error))
           :placeholder       nil
           :height            120
--- a/src/status_im/multiaccounts/recover/core.cljs
+++ b/src/status_im/multiaccounts/recover/core.cljs
@@ -8,6 +8,7 @@
             [status-im.multiaccounts.create.core :as multiaccounts.create]
             [status-im.multiaccounts.db :as db]
             [status-im.native-module.core :as status]
+            [status-im.popover.core :as popover]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.fx :as fx]
             [status-im.utils.security :as security]
@@ -150,12 +151,23 @@
             (navigation/navigate-to-cofx :recover-multiaccount-enter-phrase nil)))
 
 (fx/defn proceed-to-import-mnemonic
-  {:events [::enter-phrase-input-submitted ::enter-phrase-next-pressed]}
+  {:events [::enter-phrase-next-pressed]}
   [{:keys [db] :as cofx}]
   (let [{:keys [password passphrase]} (:multiaccounts/recover db)]
-    (when (mnemonic/valid-length? passphrase)
-      {::import-multiaccount {:passphrase passphrase
-                              :password password}})))
+    (if (check-phrase-warnings passphrase)
+      (popover/show-popover cofx {:view :custom-seed-phrase})
+      (when (mnemonic/valid-length? passphrase)
+        {::import-multiaccount {:passphrase passphrase
+                                :password   password}}))))
+
+(fx/defn continue-to-import-mnemonic
+  {:events [::continue-pressed]}
+  [{:keys [db] :as cofx}]
+  (let [{:keys [password passphrase]} (:multiaccounts/recover db)]
+    (fx/merge cofx
+              {::import-multiaccount {:passphrase passphrase
+                                      :password   password}}
+              (popover/hide-popover))))
 
 (fx/defn cancel-pressed
   {:events [::cancel-pressed]}
--- a/src/status_im/ui/screens/popover/views.cljs
+++ b/src/status_im/ui/screens/popover/views.cljs
@@ -6,7 +6,8 @@
             [re-frame.core :as re-frame]
             [status-im.ui.screens.wallet.signing-phrase.views :as signing-phrase]
             [status-im.ui.screens.wallet.request.views :as request]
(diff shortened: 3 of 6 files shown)


flexsurfer (approved):

yenda (approved):

yenda:
Tested on Android

Screenshot_20190919-124431

vitvly (approved):

churik:
@dmitryn please add accessibility-ids to "Continue" and "Cancel" buttons

churik:
@dmitryn please add accessibility-ids to "Continue" and "Cancel" buttons

dmitryn:
@churik added

churik:
@errorists please take a quick look, meanwhile I'll fix tests

errorists:
left comments here, left Figma - right Screenshot

Screenshot 2019-09-23 at 09 06 10

churik:
@dmitryn now due to this popup it is possible to create account with wrong amount of words.

**Steps:**
- enter seed phrase, that contains 13 words and some words outside dictionary
- tap on done

**Expected result:** 
keyboard is closed, "Next" button is disabled
**Actual result:**
popup "some words are misspelled"

churik:
@dmitryn Please, take my commit when squashing

Serhy (approved):

src/status_im/ui/screens/multiaccounts/recover/views.cljs
```diff
@@ -19,7 +19,51 @@
...
+     [react/text {:style {:color       colors/gray
+                          :text-align  :center
+                          :line-height 22}}
+      (i18n/label :t/custom-seed-phrase-text-3)]
+     [react/text {:style {:color       colors/black
+                          :text-align  :center
+                          :line-height 22}}
+      (i18n/label :t/custom-seed-phrase-text-4)]]
+    [react/view {:margin-bottom 24 :margin-horizontal 24 :align-items :center}
+     [components.common/button {:on-press     #(re-frame/dispatch [::multiaccounts.recover/continue-pressed])
```
flexsurfer: please use `button/button` instead, common button is deprecated

yenda:
@dmitryn use `nested-text` element

dmitryn:
@yenda thanks for the tip, i'll use it

dmitryn:
@churik bug fixed by disabling ability to proceed from keyboard unless you tap Next button.
@errorists made layout fixes, please take a look.

yenda:
@errorists please review

iPhone 6 Plus

Screenshot_20190925-080456

One Plus 7 Pro

Image-1

errorists:
Looks good @yenda thanks!

yenda:
ok looks good to merge for me
@churik if you want to do a last round of manual testing I think the following cases are way enough:
- test `a a a a a a a a a a a a` twice and check that it recovers the same account
- test a known seed phrase and check that it recovers the usual account

yenda:
@churik actually I just tested that so I'll merge

3esmit:
Why **allow invalid BIP39**? What's the specific use case of this? Is anyone aware of someone that would need that? 

Do **you trust users** they will use properly this feature?

Auditors and other wallet providers will complain about this, because **app gives option to user create an unsafe wallet**. 

This PR is solving this analogous issue: https://github.com/kvhnuke/etherwallet/issues/485
e.g. User enter seed phrase misspelled, send funds, later never can recover.
 
But not this one https://github.com/kvhnuke/etherwallet/issues/551#issuecomment-306759035  @tayvano said:
> What I failed to realized is that new users, instead of generating a wallet on the homepage like I stupidly assumed they would do, were simply typing in whatever words they felt like into that box and getting the address and calling it a day.

e.g. App gets popular... web2 users: "So I can enter any seed phrase "paswordy thing"? this ones the app give me are too difficult, I will use here a custom seed phrase with something easier to remember..."; Later get's account mined by a low-entropy account bruteforcer. 

From what we learned of Parity seed phrase case is that we should never trust the user with optional security. Instead **app should enforce security**. 

And anyway, if someone gets hurt by this, even if is the user mistake, the ultimate fault is of the app that allowed it, and therefore the social bashing would be over the app, not the random affected user. 

cc @corpetty

3esmit:
However the parity seed phrase problem had a worse checking. Status is looking for 12 words, which could be a 12 letter password separated by spaces in worst case, so perhaps the issue I mentioned above is less problematic, but **we might see people loosing funds by entering famous phrases, e.g. songs lyrics**, instead of real BIP39. 

This weak accounts are usually precalculated and keep being watched, so they are very quickly drained once one put values. Researchers were able to identity millions stolen this way, in both Bitcoin and Ethereum, and probably in any cryptocurrency where wallets allow users to make brain wallets. 

https://www.securityweek.com/most-bitcoin-brain-wallets-drained-attackers https://pdfs.semanticscholar.org/7e89/dabb8717006590b674bfb0ba45ce2bd0c131.pdf
https://www.wired.com/story/blockchain-bandit-ethereum-weak-private-keys/

hesterbruikman:
Thanks for jumping in here @3esmit. I've been a little out of the loop. Here a prior discussion on topic where you can find some of the reasoning: https://github.com/status-im/status-react/pull/8687#issuecomment-519882161 _(this comment onwards)_

As you point out, the 12 word restriction makes it less problematic, however, it's actually 12, 15, 18, 21, 24 words. So the chance of randomly hitting one of those word counts is higher.

Relevant last words summarizing that we do allow words that are not in our dictionary (also if they contain an typo as we won't know)
https://github.com/status-im/status-react/pull/8687#issuecomment-521197939

In the decision to allow words not in our dictionary is the assumption that people now or in the future potentially have used a different dictionary than the one(s) used by Status. The underlying rationale here (which is obviously open to debate) is to mitigate risk rather than block the user.

I have 3 questions to move forward:
1. Do we want to block the user from entering a seed phrase not in our dictionary(s) altogether?
2. If we don't block the user what are all exact risks they are taking?
3. What are way to mitigate these risks if we don't block the user?

1/ Personally I prefer exploring all mitigation options before deciding to block as this is an excellent opportunity to educate. We don't know what we don't know and people will never learn if we don't tell them or allow for low risk mistakes (maybe we give this type of custom account limited options, perhaps even chat & browsing options only). IMO the problem isn't necessarily that you can recover random seed phrases, the problem is what you can do next.

2/ I think you fairly point out that we don't inform users of the risk they're taking, mainly that they will never ever be able to recover this account. What are other risks?

3/ Also, I would agree that it's a little too easy to continue given the risks. We definitely don't want to encourage unsafe behavior! Maybe the explicit 'type CONTINUE' to recover this custom seed phrase is a more suitable pattern. Something only the brave will endure to continue.

yenda:
Since we are talking about multiaccount seed phrase validation (not wallet account), based on @3esmit solid arguments:
1. yes we want to block the user 
2. otherwise he might enter shitty seedphrases and get his account stolen by bots, or mispell something and lose the account forever
3. we block the user

The only valid use case was testing by using an invalid seedphrase. Status is using the standard dictionary anyway, there is no other valid dictionary. The seedphrase point is to have maximal entropy while remaining human and typo friendly, not be pleasing and easy to remember (that would mean it is also easy to crack). 

Also as mentioned we should not only consider the words and their count but also the checksum, because repeating 12 times the same word is not a valid seedphrase either.

yenda:
For education purposes I propose this metaphor:

This is the kind of key you need:

image

This is the kind of key you are trying to use:

image

3esmit:
@hesterbruikman 
1. Yes, because it's not only a potential security issue, but also hurt an important principle: `interoperability` with BIP39.

2. Risk of user misspell word is mitigated by warning, but random users are known for making bad decisions to themselves as ever as possible, even when warned.

3. The best way to mitigate this without blocking custom seed phrases would be to remove it from the default user access, and provide a plugin tool which allow easy development and activation of custom key systems, therefore users that somehow created an invalid BIP39 could take an conscious decision on adding an custom system.

4. I agree that blocking can be bad because it might affect some users, and is in our principles accessibility and inclusion, however it cannot hurt interoperability. Custom seed phrase feature introduces a way of creating this invalid BIP 39 seeds, which would not be supported by other wallets, such as MetaMask, MyCrypto, or by Nano Ledger, etc

5. Two risks: Loss of recovery phrase (in case of misspell word and ignored warning) and loss of funds (in case of use of common phrases or low entropy e.g. `one two three four five ...`) 

6. If supporting invalid BIP39 is an feature we need, and we want to make it difficult, there is a way of doing so that would be making painful to user, 
e.g. 
- [ ] I accept that this account will only work in Status.im; 
- [ ] I accept that this account have no safety guarantee ; 
- [ ] I am probably doing something bad to myself. 
[I am genius] 

Does other wallets support invalid BIP39, or how is called here, custom seed phrase?

See this comment in the original issue: https://github.com/status-im/status-react/issues/8968#issuecomment-531276242

https://our.status.im/our-principles/

**VIII. Inclusivity**
We believe in fair and widespread access to our software, with an emphasis on ease-of-use. This also extends to social inclusivity, permissionless participation, interoperability, and investing in educational efforts.

churik:
@hesterbruikman would you mind to create an issue? 
we need to fix it for `v1 audit` I believe.
cc @rachelhamlin

hesterbruikman:
Thanks for the discussion @3esmit @yenda! Reflecting I don't see a convincing current use case to support 'invalid BIP39', but want to check some other wallets to be sure @churik before I create an issue to fix.

I like the rephrasing of why should we block to why should we support invalid BIP39. From the 'blocking' persective I still stand by mitigation over blocking; but realistically, we don't have the time to do this well beyond a warning:)

Other thoughts in response:
> 2/ Risk of user misspell word is mitigated by warning, but random users are known for making bad decisions to themselves as ever as possible, even when warned.

Agree that warnings are mostly ineffective, depending on their design. That said it's a mitigation, we can consider others like lowering the severity by limiting options.

>3/ The main way to mitigate this would be to remove it from the default user access, and provide a plugin tool which allow easy development and activation of custom key systems, therefore users that somehow created an invalid BIP39 could take an conscious decision on adding an custom system.

Love this idea!

>5/ Two risks: Loss of recovery phrase (in case of misspell word and ignored warning) and loss of funds (in case of use of common phrases or low entropy e.g. one two three four five ...)

Loss off recovery phrase would only occur if people use the Recovery flow to create a new multiaccount. People are weird, but given the primary actions are on creating a new multiaccount I consider this a highly unlikely edge case.

tayvano:
Don't allow custom seed phrases.

Period.

End of conversation.

There is no such thing.

That's called a brain wallets and we don't allow people to use those for obvious humans-cant-entropy-and-will-have-their-funds-stolen-and-they-will-blame-you-and-i-will-blame-you reasons.

The BIP32/39/44 spec has a checksum (last word) as well a limitations by wordlist. Honestly, just use the library and call it a day. You're dangerously close to rolling your own crypto here and reducing interoperability is the *one* place this ecosystem has it. Pretty much literally everything else you do will have a million viable paths you can go down and decisions you will *have* to make.

Standard bip support is not one of them.

---

Okay so now that we've accepted that.....yes bip spec has issues. Yes plausible deniability is only a feature for fucking crazy crypto fucks, not users who are error prone. Yes advanced users, like you mr. developer, want X feature. No, it does not mean you should try to "improve" or mitigate bc you can't bc it would turn into xkcd standards comic. There are 1000 initiatives already to improve it and address some common issues. Work with them if you really want to.die on this hill.

Issue #8819, opened 2019-08-22
Sometimes chat history reappears for deleted chats


churik:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug

*Summary*: caught by autotests several times. It is the reason why sometimes test `test_long_press_to_delete_public_chat` is failed. Sometimes for chat with deleted history it reappeared after relogin and joining the same chat. Tried to reproduce this one ~5 times with new accounts - no luck.

#### Expected behavior
no history for deleted chat

#### Actual behavior
sometimes deleted history is fetched

### Reproduction

- Open Status
- Join public chat
- Send message
- Delete public chat by long-tapping
- Log out
- Login
- Join the same public chat

### Additional Information

[comment]: # (Please do your best to fill this out.)
- Status version: nightly 22/08/2019
- Operating System: Android (probably IOS as well)

#### Logs

Logs and video: https://app.saucelabs.com/tests/fce076e22c214dcb979a612a41c5d231?auth=672b490a949db5200e161f964490af37#95

churik:
Updated logs and video: https://saucelabs.com/jobs/ef23df5e879f45ff90c4005f6078fb0f?auth=120c5c03347eaa054ab252dbcd0ed0b9

churik:
no more failures

PR #9006, opened 2019-09-18
Use identicon / gfycat from status-go

cammellos:
## Use gfycat and identicon from status go

This commit moves gfycat and identicon generation to status-go. Names and picture have changed as we use a different algorithm to generate it (linear feedback shift register instead of Mersienne Twister).

Messages and contacts info are pre-generated on message arrival.
The rest of the calls are blocking (visiting profile for example), but no performance degradation is present (they were blocking before as well effectively).

I was not able to run desktop locally, as I understand it's currently not working and pending being fixed, so I just stubbed the endpoint for now, and will need to be fixed later, not sure what's the best approach for this.

We confirmed the performance improvement with the test-team, roughly 45% improvement.
This PR will go through normal QA given the scope of the changes.

status: ready?


cammellos:
@Serhy I have made some changes, now images are computed only in `status-go`, 3 words name have changed and images look a bit funny, but could you please test performance of this one when you have time? (there might be some bugs here and there)

What interesting to test:
1) Downloading messages
2) Switch chats 

Thanks!

cammellos:
(only on android, ios I haven't finished)

cammellos:
As a side note, to me it looks much more responsive when the chat is initially empty and then we load messages from status-go.
For example, say you have chat #status , you login:
1) Click on #status -> empty chat shows -> messages are loaded . To me this feel fairly responsive.
2) After you did 1, leave and click again , chat is already populated, but UI freeze, feels less responsive, although total time might be less.

Even though 1 is potentially longer, it feels much more responsive, so there might be a case for offloading messages once a user leaves a chat. @flexsurfer @errorists what is preferable?

cammellos:
@Serhy ping in case you haven't seen my previous message while on the offsite, thanks!

Serhy:
@cammellos looking at it now!

Serhy:
@cammellos Something is wrong here with the latest build on Android (6.0.1 tested so far od real device Galaxy note 4).
App crashes once I get a redirect to the screen with 5 users created during new/recovered account creation. 
So when I Open Status fresh install -> Get Started -> Generate a key:

Screenshot 2019-09-20 at 12 36 51

cammellos:
@Serhy last build should fix the issue, let me know in the meantime if you see any perf. improvement in the previous builds, thanks!

errorists:
@cammellos 

> For example, say you have chat #status , you login: Click on #status -> empty chat shows -> messages are loaded . To me this feel fairly responsive.
After you did 1, leave and click again , chat is already populated, but UI freeze, feels less responsive, although total time might be less. (…) what is preferable?

so it's hard to say without looking at the build or at least a video but I'd agree with you that it's better to present the content-less chat view first without any hiccup and only that is done load the messages rather than trying to force everything into a single bite and watch the UI choke on it.

Serhy:
Tested on Galaxy note 4 the latest build based on #status public chat. Because we load last 24h history there are up to ten messages but it's anyway gives an idea of whether we are ok to proceed with loading empty chat and then the messages from status-go.

@errorists This is a video I recorded and slowed it down twice: https://drive.google.com/file/d/1udJtr-sOTcX6O7Qnnox1g3Jx7k0fL8xH/view 
Scenario for above video was that I joined #status, then closed app -> open Status and loging and stay on Chats screen. From that moment:
- Tap on #status chat. 
Time between tap action (visible on video) disappeared till the messages appeared ~1.25 sec: 
(0.25 sec takes to start UI load new screen + 0.25sec for that rendering of empty chat is finished + 0.6 sec for empty chat keep staying (i.e. the chat logo with 'It's been quiet here for the last 24h') + 0.15s for messages appear and images has 0 opacity  
- Return back to Chats. Tap on #status chat again.
Time between tap action (visible on video) disappeared till the messages appeared ~0.8 sec: 
(0.55 sec takes to start UI load new screen + 0.25 for that rendering with new screen with all messages is finished)

So the idea of offloading messages every time user navigates away from chat view is to have **always** behavior that: 
User tap on chat with **any** type of messages -> it will always take about ~0.25 sec to get a redirect to empty chat (aka 'no message history' chat view) - and then while messages are 'retrieving' from status-go side we show empty chat view(currently it will be static 'There is no messages for the last 24h history' view)... and finally display the messages.

Apparently yes, UX in that case better from my POW. We have some responsive app activity happening when tapping on chat view. Instead of heavy lags (when a lot of messages in chat) experience upon tap.

errorists:
@Serhy 
on the video, the part that showed the chat header before messages get loaded is bad, like super bad :D it should either be absolutely empty or show the end state with all messages present. The chat header should be treated the same was as rest of messages, either it's in its correct position or not shown at all.

Serhy:
I've checked perf comparison between nightly develop build and current PR build. It's not very precise because those are different builds. But it's just to give an idea of improvements in time loads that are really visible.
On both builds: Join `#status` public and load messages. Close the app. Open app and login. From Chats view tap `#status`. Measure the time it takes for messages to appear -> navigate back to Chats and tap `#status` again to measure the time it takes for messages to appear.

| Build  | Navigate to chat view first time | Navigate second time |
| ------------- | ------------- | -- |
| 23rd of Sept nightly  | 2.3 sec  | 1.8 sec |
| Current PR build  | 1.3 sec  | 0.8 sec |

@cammellos, as @errorists suggested above https://github.com/status-im/status-react/pull/9006#issuecomment-534102697 - the way of having empty chat upon navigating to chat is not the best approach (even it looks faster). Well, it's not exactly at what he points, but at this stage we don't show chat header upon after-the-first navigation (so after messages are loaded - the more "laggy" second case). Let's keep current behavior I think. At least in the scope of current PR.

@cammellos is the iOS part ready too?

cammellos:
thanks sehry, I agree, I will just include the perf improvement and keep as it is, ios is ready if you want to test performance ( i dont have an ios device myself), there s a few issuea i need to fix, but they are not performance related

flexsurfer:
> 3 words name have changed and images look a bit funny

cc @errorists @andmironov @rachelhamlin @hesterbruikman 


btw why have 3 words changed ? do we use another dictionary?

cammellos:
no, different algorithm to choose the numbers, before we were using
Mersenne Twister, and we were relying on a library to do the initial seed
(chance.js).

That's much more complex then what we need, so we changed it to Fibonacci
linear feedback shift register, which is  much simpler and easier to re
implement in different languages. ( 5 lines of code vs 20/30 for Mersenne
twister as chance.js implemented it)

src/status_im/contact/db.cljs
```diff
@@ -65,7 +65,8 @@
 (defn public-key->new-contact [public-key]
   {:name        (gfycat/generate-gfy public-key)
    :address     (ethereum/public-key->address public-key)
-   :photo-path  (identicon/identicon public-key)
+   :photo-path  ""
```
yenda: why not removing it rather than using empty string?

cammellos: It's up to us, in this case I prefer having an empty string as it makes it easier to interact with status-go and database (strings are not nullable), plus is consistent type (always string, never null), but can change if you feel strongly about it.

src/status_im/constants.cljs
```diff
@@ -21,7 +21,7 @@
 (def max-chat-name-length 20)
 (def max-group-chat-participants 10)
 (def response-suggesstion-resize-duration 100)
-(def default-number-of-messages 20)
+(def default-number-of-messages 10)
```
yenda: that might have had an impact on perfs, though that might not be enough for some big displays

cammellos: oh I actually forgot to change it back, I was testing something

yenda: tested on One Plus Pro 7 there is a clear jump when you open a chat with small messages and more than 10 fit in one screen

yenda: sorry our messages crossed

src/status_im/contact/db.cljs
yenda: in the clojure code having an empty string here is not idiomatic, the photo-path is not there and "" is truthy, for consistency I don't think we should do this. If there is a need for an empty string at the borders of the program it can be added there

src/status_im/multiaccounts/core.cljs
```diff
@@ -10,15 +10,29 @@
...
             [status-im.utils.platform :as platform]
             [status-im.utils.utils :as utils]))
 
-(defn displayed-name [account]
-  (let [name (or (:preferred-name account) (:name account))]
+(defn displayed-name
+  "Use preferred name, name or alias in that order"
+  [{:keys [name preferred-name alias public-key]}]
+  (let [name (cond
+               (seq preferred-name) preferred-name
```
yenda: I feel the same here all these should be either nil or usable values

src/status_im/native_module/core.cljs
```diff
@@ -265,3 +265,12 @@
 
     ;; in unknown scenarios we also consider the device rooted to avoid degrading security
     :else (callback true)))
+
+(defn generate-gfycat
+  [seed]
+  (.generateAlias (status) seed))
+
+(defn identicon
+  [pk]
```
yenda: isn't it `seed` as well? (if we want to keep it generic, otherwise `public-key` rather than pk to keep naming consistent)

src/status_im/subs.cljs
```diff
@@ -722,9 +723,10 @@
  :<- [:contacts/contacts]
  :<- [:multiaccount]
  (fn [[contacts multiaccount] [_ id]]
-   (or (:photo-path (contacts id))
-       (when (= id (:public-key multiaccount))
-         (:photo-path multiaccount)))))
+   (multiaccounts/displayed-photo (or (contacts id)
+                                      (when (= id (:public-key multiaccount))
```
yenda, line 727: this condition will return a photo-path so it will always recompute the photo because displayed-photo expects a map

src/status_im/subs.cljs
```diff
@@ -722,9 +723,10 @@
...
+                                        (:photo-path multiaccount))
```
yenda:
```suggestion
                                        multiaccount)
```

src/status_im/transport/message/protocol.cljs
```diff
@@ -82,6 +82,8 @@
                                        :whisper-timestamp timestamp
                                        :raw-payload-hash (ethereum/sha3
                                                           (.-payload (:js-obj cofx)))
+                                       :alias (get-in cofx [:metadata :author :alias])
```
yenda, line 85: can those be empty strings?

src/status_im/subs.cljs
cammellos: oh yes, good catch

src/status_im/ui/screens/hardwallet/login/views.cljs
```diff
@@ -30,7 +31,7 @@
      [react/view styles/login-view
       [react/view styles/login-badge-container
        [react/view styles/login-badge
-        [photos/photo photo-path {:size styles/login-badge-image-size}]
+        [photos/photo (multiaccounts/displayed-photo account) {:size styles/login-badge-image-size}]
```
yenda, line 35:
```suggestion
        ;;TODO this should be done in a subscription
        [photos/photo (multiaccounts/displayed-photo account) {:size styles/login-badge-image-size}]
```

src/status_im/transport/message/protocol.cljs
cammellos: no, I believe we always fill it in

src/status_im/ui/screens/keycard/views.cljs
```diff
@@ -397,7 +398,7 @@
                       :height          69
                       :justify-content :center
                       :align-items     :center}
-          [photos/photo photo-path {:size 61}]
+          [photos/photo (multiaccounts/displayed-photo account) {:size 61}]
```
yenda, line 402:
```suggestion
          ;;TODO this should be done in a subscription
          [photos/photo (multiaccounts/displayed-photo account) {:size 61}]
```

src/status_im/ui/screens/keycard/views.cljs
```diff
@@ -471,7 +472,7 @@
                         :height          69
                         :justify-content :center
                         :align-items     :center}
-            [photos/photo photo-path {:size 61}]
+            [photos/photo (multiaccounts/displayed-photo account) {:size 61}]
```
yenda, line 477:
```suggestion
            ;;TODO this should be done in a subscription
            [photos/photo (multiaccounts/displayed-photo account) {:size 61}]
```

src/status_im/ui/screens/multiaccounts/login/views.cljs
```diff
@@ -29,19 +30,21 @@
   (.blur password-text-input)
   (re-frame/dispatch [:multiaccounts.login.ui/password-input-submitted]))
 
-(defn multiaccount-login-badge [photo-path name public-key]
+(defn multiaccount-login-badge [{:keys [public-key] :as multiaccount}]
   [react/view styles/login-badge
-   [photos/photo photo-path {:size styles/login-badge-image-size}]
+   [photos/photo
+    (multiaccounts/displayed-photo multiaccount)
```
yenda, line 37:
```suggestion
    ;;TODO this should be done in a subscription
    (multiaccounts/displayed-photo multiaccount)
```

src/status_im/ui/screens/multiaccounts/login/views.cljs
```diff
@@ -29,19 +30,21 @@
...
+    {:size styles/login-badge-image-size}]
    [react/view
     [react/text {:style          styles/login-badge-name
                  :ellipsize-mode :middle
                  :numberOfLines  1}
-     name]
+     (multiaccounts/displayed-name multiaccount)]
```
yenda, line 44:
```suggestion
     ;;TODO this should be done in a subscription
     (multiaccounts/displayed-name multiaccount)]
```

src/status_im/ui/screens/profile/contact/views.cljs
```diff
@@ -91,13 +91,14 @@
...
 
-(defn- header-in-toolbar [{:keys [photo-path] :as account}]
+(defn- header-in-toolbar [account]
   (let [displayed-name (multiaccounts/displayed-name account)]
     [react/view {:flex           1
                  :flex-direction :row
                  :align-items    :center
                  :align-self     :stretch}
-     [photos/photo photo-path {:size 40}]
+     [photos/photo (multiaccounts/displayed-photo account)
```
yenda, line 102:
```suggestion
     ;;TODO this should be done in a subscription
     [photos/photo (multiaccounts/displayed-photo account)
```

src/status_im/ui/screens/profile/user/views.cljs
```diff
@@ -70,19 +70,19 @@
...
 
-(defn- header-in-toolbar [{:keys [photo-path] :as account}]
+(defn- header-in-toolbar [account]
   (let [displayed-name (multiaccounts/displayed-name account)]
     [react/view {:flex           1
                  :flex-direction :row
                  :align-items    :center
                  :align-self     :stretch}
-     [photos/photo photo-path {:size 40}]
+     [photos/photo (multiaccounts/displayed-photo account) {:size 40}]
```
yenda, line 86:
```suggestion
     ;;TODO this should be done in a subscription
     [photos/photo (multiaccounts/displayed-photo account) {:size 40}]
```

src/status_im/native_module/core.cljs
cammellos: we can have seed, although there are some expectations on the format (it needs to be a public key for gfycat actually)

src/status_im/ui/screens/wallet/components/views.cljs
```diff
@@ -186,7 +188,9 @@
 (defn render-contact [contact request?]
   [list/touchable-item #(re-frame/dispatch [:wallet/fill-request-from-contact contact request?])
    [list/item
-    [photos/photo (:photo-path contact) {:size list.styles/image-size}]
+    [photos/photo
+     (multiaccounts/displayed-photo contact)
```
yenda, line 194:
```suggestion
     ;;TODO this should be done in a subscription
     (multiaccounts/displayed-photo contact)
```

src/status_im/ui/screens/wallet/components/views.cljs
```diff
@@ -173,7 +173,9 @@
       [react/view styles/recipient-container
        [react/view styles/recipient-icon
         (when contact
-          [photos/photo (:photo-path contact) {:size list.styles/image-size}])]
+          [photos/photo
+           (multiaccounts/displayed-photo contact)
```
yenda, line 178:
```suggestion
           ;;TODO this should be done in a subscription
           (multiaccounts/displayed-photo contact)
```

src/status_im/native_module/core.cljs
yenda: if there is expectations could you add  a precondition? and also could you add a docstring for both that explains that it is synchronous calls and what they do?

yenda:
@errorists designer opinion needed, looks like the colors are less pastel than before, 

Screenshot_20190925-085706

and I never got such a boring generated picture before

Screenshot_20190925-085732

errorists:
@yenda they definitely are, thanks for the heads up Eric! I don't like these, very Windows 95, what happened?
also on the first screen, the Top Bar overlays the system status bar, why are these of all things so troublesome…

src/status_im/utils/identicon.cljs
```diff
@@ -1,5 +1,7 @@
 (ns status-im.utils.identicon
-  (:require [status-im.js-dependencies :as dependencies]))
+  (:require
+   [status-im.native-module.core :as native-module]
+   [status-im.js-dependencies :as dependencies]))
```
flexsurfer: dependencies isn't used anymore? can we delete identicon-js package as well?

src/status_im/utils/gfycat/core.cljs
```diff
@@ -1,6 +1,7 @@
 (ns status-im.utils.gfycat.core
   (:require [status-im.utils.gfycat.animals :as animals]
             [status-im.utils.gfycat.adjectives :as adjectives]
```
flexsurfer: dictionaries aren't used anymore? can we delete them?

src/status_im/utils/identicon.cljs
```diff
@@ -8,6 +10,6 @@
                                     :margin     0.24
                                     :size       default-size})))
   ([hash options]
-   (str "data:image/png;base64,"
-        (let [identicon-js dependencies/identicon-js]
-          (str (new identicon-js hash options))))))
+   (native-module/identicon hash)))
```
yenda: this arity is useless now since there is no options

src/status_im/ui/screens/contacts_list/views.cljs
```diff
@@ -9,10 +10,10 @@
             [status-im.ui.components.toolbar.view :as toolbar.view])
   (:require-macros [status-im.utils.views :refer [defview letsubs]]))
 
-(defn contacts-list-item [{:keys [public-key name photo-path] :as contact}]
+(defn contacts-list-item [{:keys [public-key] :as contact}]
   [list.views/big-list-item
-   {:text         name
-    :image-source photo-path
+   {:text         (multiaccounts/displayed-name contact)
```
flexsurfer, line 16: formatting

yenda:
@cammellos for the color issue in the identicons, does the status-go lib allow some tweaking? I know that with the identicon.js recent updates that we never did, I had to tweak the options like this because they removed them as default:

`:background [255 255 255]
  :saturation 0.5
  :brightness 0.7
  :margin     0.24`

note: no background transparency, add saturation and brightness setting

cammellos:
looking into it, pretty sure we can

src/status_im/multiaccounts/core.cljs
```diff
@@ -10,15 +10,29 @@
...
+               (seq name) name
+               (seq alias) alias)]
     (if (ens/is-valid-eth-name? name)
```
flexsurfer, line 23: so do we still calculate this for every message?

src/status_im/multiaccounts/core.cljs
```diff
@@ -10,15 +10,29 @@
...
       (let [username (stateofus/username name)]
         (str "@" (or username name)))
-      (or name (gfycat/generate-gfy (:public-key account))))))
+      (or name (gfycat/generate-gfy public-key)))))
+
+(defn displayed-photo
+  "If a photo-path is set use it, otherwise fallback on identicon or generate"
+  [{:keys [photo-path identicon public-key]}]
+  (cond
+    (seq photo-path) photo-path
```
flexsurfer: so do we still calculate this for every message?

src/status_im/multiaccounts/core.cljs
cammellos: we don't calculate it in javascript, calculation is in status-go and it's saved in the database. On new messages we always calculate it, as it's faster than pulling them from db, on loading messages it's already there, same goes from images.

src/status_im/utils/gfycat/core.cljs
cammellos: yes, I was waiting to remove libraries and dictionaries until we made a decision about desktop

cammellos:
@flexsurfer @yenda ready to be reviewed again, I have added the pastel colors back.
screen

src/status_im/chat/models.cljs
```diff
@@ -85,7 +85,7 @@
   [chat-id {:keys [db now]}]
   (let [name (get-in db [:contacts/contacts chat-id :name])]
     {:chat-id            chat-id
-     :name               (or name (gfycat/generate-gfy chat-id))
+     :name               (or name "")
```
yenda, line 88: how can name be falsy?

src/status_im/multiaccounts/core.cljs
```diff
@@ -10,15 +10,29 @@
...
             [status-im.utils.platform :as platform]
             [status-im.utils.utils :as utils]))
 
-(defn displayed-name [account]
-  (let [name (or (:preferred-name account) (:name account))]
+(defn displayed-name
+  "Use preferred name, name or alias in that order"
+  [{:keys [name preferred-name alias public-key]}]
+  (let [name (or
+              preferred-name
```
yenda: first arg should be on same line as or, same for line 27

src/status_im/ui/screens/contacts_list/views.cljs
```diff
@@ -9,11 +10,11 @@
             [status-im.ui.components.toolbar.view :as toolbar.view])
   (:require-macros [status-im.utils.views :refer [defview letsubs]]))
 
-(defn contacts-list-item [{:keys [public-key name photo-path] :as contact}]
+(defn contacts-list-item [{:keys [public-key] :as contact}]
   [list.views/big-list-item
-   {:text         name
-    :image-source photo-path
-    :action-fn      #(re-frame/dispatch [:chat.ui/show-profile public-key])}])
+   {:text         (multiaccounts/displayed-name contact)
```
yenda, line 16:
```suggestion
   ;;TODO this should be done in a subscription
   {:text         (multiaccounts/displayed-name contact)
```

src/status_im/utils/gfycat/core.cljs
```diff
@@ -1,28 +1,14 @@
...
+    (native-module/generate-gfycat public-key)))
 
-(defn generate-gfy
-  ([public-key]
-   (case public-key
-     nil unknown-gfy
-     "0" unknown-gfy
-     (build-gfy public-key)))
-  ([] (generate-gfy (datetime/timestamp))))
+(def generate-gfy (memoize build-gfy))
```
yenda, line 14: should we still memoize?

yenda: well why not

src/status_im/chat/models.cljs
cammellos: This fn is used for all kind of chats, so after I have added a `pre` to generate-gfy chokes on it, as it's passed the `chat-id` of `#status` for example. Also I believe `name` is not used for anything but group-chats.

yenda: if generate-gfy chokes on it that would mean there is a case where name is nil and yet chat-id is not a public-key, which scenario is that?

cammellos: public chat?

yenda: do we know at this point which type of chat it is? if that is the case it could be better handled with cond-> and make :name optional

cammellos: maybe, but should we do it another time? this code hasn't been changed and I want to limit the changes as it's already quite a change

yenda, line 88:
```suggestion
     ;;TODO should be handled with cond-> and make :name optional since public-chats don't have a name
     :name (or name "")
```

yenda:
Screenshot_20190925-155603
Screenshot_20190925-155751
Screenshot_20190925-155801

yenda:
@errorists see screenshots above, looks good now?

@cammellos I think there is a bug with group chat the name is `<no value>` for some reason

cammellos:
@yenda can't replicate, I have tried both on chats I created and I was invited to, checked main screen and inside the chat, any repro steps?

yenda:
@cammellos just reproduced again. On Android I create an account, joins #test public chat, open profile of the user who posted the last message, tap Send Message, Add to Contact banner, go back Create group chat

errorists:
@yenda colors look good now thanks, I love this secret sauce of `:background [255 255 255] :saturation 0.5 :brightness 0.7`, makes perfect sense

cammellos:
@yenda Tried to replicate on multiple devices: Table/phone using the last apk generated https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190925-143937-722052-pr.apk , and on emulator https://status-im.ams3.digitaloceanspaces.com/StatusIm-190925-112646-2b7f7b-e2e.apk and live code . 
I can't replicate on none of those devices, not sure exactly how to progress, could you send me a screenshot of the error, maybe that can help even if I can't replicate?

yenda:
@cammellos same here I tried with multiple accounts even with the same contact than in the screenshots above and could not reproduce.

Tested on One Plus 7 Pro

Joined public, 1-1 and group chats
Visited multiple profiles
Visited user profile
Tested custom user and custom contact picture in all of the above

Good to merge

yenda (approved):

yenda:
@churik e2e tests need to be changed for effect the fact that the algo to generate names has changed

cammellos:
@yenda I am on it, @churik I will ask you to take a look once done

churik:
@cammellos please ping me if you need help.
Should we manually check it (at least happy-path) on IOS?

yenda:
@churik just going to the screen with 5 choices of identity is enough for iOS, any iOS specific error would have triggered by then

churik:
@cammellos tested on IOS 11.4.1:
-  screen with 5 choices of identity 
- creating a user
- navigating to other's user profile from 

Steps to reproduce an issue with *no value*:
- open user profile from public chat
- tap on "Add to contacts"
- Go back to home
- Create group chat with this user 

Reproducible every time for me (tested on IOS 11.4.1)

With these steps https://github.com/status-im/status-react/pull/9006#issuecomment-535042135 can't replicate.

yenda:
@churik @cammellos indeed I can replicated now, so it's when you add to contact from a public chat but don't start a 1-1 chat now and start a group chat directly

cammellos:
@churik I have updated the tests but it seems it's still running the old ones? (There no reference to Cadetblue Popular Monoclo in the code anymore)

churik:
@cammellos for this you need to launch job with branch set to your PR: https://ci.status.im/job/end-to-end-tests/job/status-app-end-to-end-tests/3457/

cammellos:
Thanks @churik !

churik:
Wow, @cammellos you are rock-star! 
Thanks for your initiative.
Only one failure, that is not related to your PR.
When issue with "no value" will be fixed - ready to merge

cammellos:
@churik should be fixed in the next build, could you also run the tests once the apk is ready please? Thanks!

cammellos:
This is ready to be merged, if anyone wants to confirm the bug with group chat has been fixed ( I tested myself and it's ok), should also fix: #8819

yenda:
Tested the bug is gone

cammellos:
@yenda would you mind merging, I will have to be afk for the rest of the day, thanks!

modules/react-native-status/desktop/rctstatus.cpp
```diff
@@ -490,3 +490,17 @@ void RCTStatus::chaosModeUpdate(bool on, double callbackId) {
             d->bridge->invokePromiseCallback(callbackId, QVariantList{result});
         }, on, callbackId);
 }
+
+QString RCTStatus::generateAlias(QString publicKey) {
+    Q_D(RCTStatus);
+    qCDebug(RCTSTATUS) << "::generateAlias call";
+    //return GenerateGfycat(publicKey.toUtf8().data());
+    return "test";
```
vkjr, line 498: @cammellos, was this intentionally left commented?

cammellos: @vkjr Yes, I could not run desktop locally, I was told that is currently not working and adviced to leave it as it is, so I can't really tests this methods, (they need to be synchronous, not async, not sure how desktop handles this), I can take a look at it, but I might need some help from you getting stuff going

vkjr: Ok, I'll take care of them :)

Issue #8485, opened 2019-06-25
[#8026] Update not enough ETH warning


hesterbruikman:
### Description

Type: Feature

Summary: Continuation of the work done in this PR for #8026.

New overview screen warning: "Not enough ETH for gas". This warning provides insufficient information to new users as the what the problem is and how to fix it and refers to gas which is inconsistent with language in the Overview screen (Network fee). Update includes a warning that provides more context.

**Current behavior**
![IMG_0129](https://user-images.githubusercontent.com/35233054/60078077-8ad34400-972b-11e9-8fc3-70f1b54b9926.JPG)

**Expected behavior**
![image](https://user-images.githubusercontent.com/15470435/64772400-73487200-d548-11e9-8b8a-cfab0bb05bec.png)
Expected behavior was updated as of [this comment](https://github.com/status-im/status-react/issues/8485#issuecomment-527892583)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 70.0 DAI (70.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/status-im/status-react/8485/3404).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/status-im/status-react/8485/3404).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $139,587.91 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [majortomsec](https://gitcoin.co/profile/majortomsec) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/status-im/status-react/8485/3404?mutate_worker_action=approve&worker=majortomsec) | [reject worker](https://gitcoin.co/issue/status-im/status-react/8485/3404?mutate_worker_action=reject&worker=majortomsec))_.**

I'd like to work on adding sufficient information on this warning screen.
I've worked with Status before and I enjoyed it, so I'm willing to work more with the team !

[...]

Issue #8691, opened 2019-08-02
"Invalid number" (no gas limit is set by default) instead of insufficient balance error at attempt to register ENS name


churik:
### Description

[comment]: # (Feature or Bug? i.e Type: Bug)
*Type*: Bug
*Summary*: if you don't have enough SNT to register new name - instead of appropriate error application couldn't set Network fee (Gas Limit is always empty)
If you set it Gas Limit manually, you can proceed and sign transaction - "Intrisic gas too low" error.
Theoretically, it can be reproduced in any DApp if token balance is not validated on DApp side (tried in AirSwap and UniSwap, balance is validated.)

#### Expected behavior
"Insufficient balance" error or button Register is disabled when you don't have enough SNT (so you can't open transaction screen)

#### Actual behavior
![IMG_3C3D24D4FE85-1](https://user-images.githubusercontent.com/4557972/62355103-a261d180-b50e-11e9-803d-c0a767a29f5f.jpeg)

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Create new account 
- Try to register new ENS name

### Additional Information

[comment]: # (Please do your best to fill this out.)
- Status version: nightly 01.08.2019
- Operating System: Android, iOS

asemiankevich:
@rachelhamlin

flexsurfer:
@rachelhamlin @hesterbruikman @andmironov @errorists in the current case there is an error in the contract because there is not enough SNT, and contract can't be run, and gas can't be estimated, so we need to show a proper message when gas can't be estimated and there is an error in the contract and we know transaction will be failed even if user provide gas manually

hesterbruikman:
@flexsurfer thanks for tagging!

What can the user do in this situation?

It sounds like:
The issue might resolved if the user somehow buys more ETH
Does this solve the issue or might the contract error also occur when there is enough ETH for gas available?

flexsurfer:
@hesterbruikman this current case not related to ETH it's about a contract execution, contract can't be executed, and we don't know the exact reason, yeah for this current case we could guess that there is not enought SNT and if user buys more SNT it will work, but we need more general message for all cases, something like "gas cannot be estimated because of an error during contract execution"

hesterbruikman:
Thanks for clarifying. Sounds like we're looking for a bandaid:/ In this case the user also can't change the amount as it's specified by the contract.

What causes the contract error?

For a message I suggest:
[...]

PR #9009, opened 2019-09-20
Network fee error changed from tooltip to red text

igormuba:
#8485

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- network consumption

### Steps to test

- Open Status
- Click on your wallet
- Choose "send"
- Fill the wallet field
- On the amount field input the maximum amount you have on your wallet, so that you see the error
- Proceed to sign
- You will see the red message saying that, because you are transfering all you have, you don't have anything left for fees

status: ready

### DIFF
--- a/src/status_im/ui/components/list_item/views.cljs
+++ b/src/status_im/ui/components/list_item/views.cljs
@@ -405,3 +405,74 @@
                  [accessories-column accessories width])]]
              (when error
                [tooltip/tooltip error styles/error])])))})))
+
+
+(defn list-item-with-red-text
+
+[_]
+(let [width (reagent/atom 0)
+      r-key (name (gensym "list-item"))]
+  (reagent/create-class
+    {:reagent-render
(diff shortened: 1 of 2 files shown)


yenda:
@flexsurfer shouldn't it be an option rather than a different element?

flexsurfer:
thanks, @igormuba , but the purpose of `list-item` is to have one component with options, could you please add an option to `'list-item` instead duplication of component, thanks

rachelhamlin:
Let's actually pause work on this PR @igormuba before you make any changes to the list-item. I want to ensure we're meeting the spec, and it looks like we've got some confusion. I'm going to merge #8485 and #8691 into one, I think, and then we can recommission.

PR #9021, opened 2019-09-21
Use modern build system for iOS

pedropombeiro:

### Summary

This PR gets rid of the code to use the legacy build system for iOS to work around an issue with Xcode 10.1 (we now require 10.2.1). It also does a bit of reformatting (using Rubocop) and simplification of the Ruby code.

status: ready


jakubgs (approved):
I see, the main change is dropping `UseModernBuildSystem=N`, but I approve of the formatting fixes and Fastlane upgrade too. Nice.

jakubgs:
Uh oh, but are we sure we want to skip manual QA? This is a big change in build system.

pedropombeiro:
> Uh oh, but are we sure we want to skip manual QA? This is a big change in build system.

True, especially since I don't have an iOS device to test on. Dropping the label. @churik do you have an iOS device you could test this with? It should be enough to just open the app and create an account.

Serhy:
On iOS 12.4.1 (iPhoneX) can install Status, can create new and can recover successfully accounts. 
Message history is loaded, messages exchanged between two contacts in 1-1 chat fine

PR #9026, opened 2019-09-23
[#8516] Cannot delete dapp from "Recent"

### LINKED ISSUE #8516: Cannot delete dapp from "Recent"
### Description

*Type*: Bug

*Summary*: cannot use swipe and delete gesture to remove last visited dapps - when tap on "Delete", nothing happens (Android only). In chat it was replaced with long tap > delete , IMO it will be better to have this behavior for dapps as well.

#### Expected behavior
can delete browser entry

please see comments from @flexsurfer on the issue:
`maybe related to native animation, but anyway we need to change it on long press`

`"Clear all" also should be implemented`

#### Actual behavior
nothing happens

### Reproduction

- Open Status
- Open "DApps"
- Open any dapp and go back to main "DApps" screen
- Below "Recent" swipe to delete visited dapp from history

### Additional Information

- Status version: nightly 02/07/2019
- Operating System:  Android 8

flexsurfer:
fixes #8516

image

tested on ios simulator


flexsurfer:
@Serhy @churik AT blocker

vitvly (approved):

flexsurfer:
@Serhy @churik do we need to add/fix e2e test for this PR ?

src/status_im/browser/core.cljs
```diff
@@ -52,6 +52,15 @@
                      :params [browser-id]
                      :on-success #()}]})
 
+(fx/defn clear-all-browsers
```
yenda, line 55:
```suggestion
(fx/defn clear-all-browsers
"TODO: use rpc-call that deletes all browsers to avoid this"
```

yenda (approved):

churik:
@flexsurfer yes, we need, I'm on it

Serhy (approved):

churik:
@flexsurfer - ready to merge :)
please, take my commit when squashing

flexsurfer:
thanks @churik can we run e2e again just to make sure they are 100% now ?

churik:
@flexsurfer it is different test (high-prio), and it is not enabled in PR now - we are aiming to enable them until next week.
As for failed `test_long_press_to_delete_public_chat` - the reason is https://github.com/status-im/status-react/issues/8819 - so nothing related to this PR

PR #9030, opened 2019-09-24
Update logcat tests and disable upgrade-app test

Serhy:
- we had often false failing test_logcat_backup_recovery_phrase test. It was checking whether any of two words, used as confirmation to back up seed phrase, are present in logcat. If any of two is present test fails. Issue is that in dictionary for seed-phrases there are many words which are used as normal log items: for example "talk" or "list" or "icon" and many more. So we modified the test to check whether both words are in logcat
- test_send_eth_from_wallet_incorrect_address updated so no ERROR in Jenkins. 
- upgrade_app test disabled (no upgrade support for now anyway)


churik (approved):
Nice work!

PR #9032, opened 2019-09-24
throw errors on dangerous behaviours

yenda:
This PR doesn't fix anything but instead will help solve a few pending bugs. (it will change the way these bugs manifest and decrease their severity, accounts shouldn't be toasted anymore and instead the app crashes and the issue is logged)

- if multiaccount settings are saved on top of an empty map or nil,
this means something went wrong, the state of the app is unstable,
and actually saving will result in loss of data. It should never
happen, but if it does, throw and error and abort.
- sometimes two fxs are merged when they shouldn't, this is caused by
bugs and should never happen, but if it does, throw an error with arguments
for both effects to help localize the error

status: ready

### DIFF
--- a/src/status_im/multiaccounts/update/core.cljs
+++ b/src/status_im/multiaccounts/update/core.cljs
@@ -5,7 +5,8 @@
             [status-im.transport.message.contact :as message.contact]
             [status-im.transport.message.protocol :as protocol]
             [status-im.utils.fx :as fx]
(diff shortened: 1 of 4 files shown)


flexsurfer (approved):

yenda:
tested on iPhone, it doesn't crash the app with #8930 but it does prevent account toasting

PR #9033, opened 2019-09-25
biometric

### LINKED ISSUE #8966: Biometric sign-in
I've been working on Face ID/Touch ID/Fingerprint sign-in options in settings and ran into a few problems:

1. Is it really "SIgn-in"? I mean, obviously you can not sign in on another device using your fingerprint, which means we are not offering you a "fingerprint sign-in". Most apps call this "app lock" which is something on top of sign-in/sign-out.

2. "Save password" option. If the user has this turned on, there's no reason to enable biometric sign-in at all. Although this is true only if we assume that the goal of  "biometric sign-in" is to speed up the sign-in process, make it smooth. Because what can be smoother than not asking for anything at all?

In general, I have a feeling that we have not thought this through and it requires more discussion and research. What is the problem we're solving here? I understand that this feature is in high demand by our community, but maybe they are asking for something else actually? What if some of them want an "app lock" – a way to prevent strangers from opening the app when they got their phone. And others want transaction signing with a fingerprint. And others just did not notice that "Save password" option (or use an old Android build) and are just annoyed by entering the password each time they run the app? 

If it was up to me, I'd suggest postponing this feature to have some more time for proper research.

But, anyway, to make it work now as a really simple app-lock, we can do this:

**Settings**
"App lock with Face ID" option in "Privacy and security". 

Screen Shot 2019-09-11 at 12 36 11

**Save password feature**
If the user does not have the "save password" box checked they get asked for the password each time they run the app. If the  goal of "biometric sign-in" aka “app-lock” is making it easy yet secure to sign-in, it basically means that "biometric sign-in" is a security level ontop of "save password". (for those who choose that it’s better for them to instead of being asked for nothing, to be asked for biometric data to run the app) This means that to enable "biometric sign-in" the user should first check the "save password" box on the sign-in screen.

So if the user tries to turn the "App-lock with Face ID" toggle on, they will get this popup:

Screen Shot 2019-09-11 at 12 40 42

If the "save password" box is checked, it just going to turn the app-lock on and inform that everything is okay. It's important to mention here when are we going to check for the face id/fingerprint (each time the user runs the app) since we don't have an option for the user to control that yet like some other apps do.

Screen Shot 2019-09-11 at 12 42 26

**Sign-in screen**
If the app-lock is enabled, it checks for the biometric data on the sign-in screen. If the face/fingerprint is not recognized, the native "Face not recognized" alert is shown. But we still need to provide a re-try option for the users, this is why there's a face id/fingerprint icon in the password field for those who have the app-lock enabled.

Screen Shot 2019-09-11 at 12 48 04

**Save password again**
If the user has the app-lock enabled but tries to un-check the save password box, we need to tell them that by doing that, they disable the app lock.

Screen Shot 2019-09-11 at 12 53 47

**Onboarding screens**
I’d remove the whole thing from onboarding since it has almost nothing to do with sign-in/sign-up. Biometric app-lock is a feature for those who don't want to **enter the password every time** but want an **extra layer of security on top of saving the password**.
Also, we want the user to first check the "save password" box to keep things consistent.

> Note: Copy on all these screens may need some more work/discussion. Also, I'm not sure that "App-lock" is the best way to call this feature, I just think that "sign in with Face/Touch ID” is wrong

flexsurfer:
fixes #8966 

https://miro.com/welcomeonboard/QFujCCoPZdhADBbdCqslx3rA1FnMS1AefhXb4ObUuDaq86R9i5YAWjKQdrQ7S6HC

image

| Bio  | Bio 2 | Bio 3 |
| ------------- | ------------- | ------------- |
| image  | image | image |

Tested on ios simulator with face id and on real Android device with fingerprint

**Not implemented:**
1) popover when deselect "save password"
2) all text on popover has the same style, no bold parts because it's not worth it to split text, it will make translation more complex


yenda:
I don't understand why so much was move, like why logout is in login now?

flexsurfer:
> I don't understand why so much was move, like why logout is in login now?

to solve circle deps

yenda:
@flexsurfer there has to be a better way to solve the circular dependency than having logout in login namespace what was the circular deps error?

flexsurfer:
@yenda we need to logout when login, and on login and logout we need to clear app-db, so there is no sense to split them, there are a few function for logout, why would we want to separate them? yes i could try to separate them but i don't see a strong reason why would we want this, for me it's event better to have login/logout in the same namespace

flexsurfer:
@Serhy @churik i need your help with e2e here, please :)

src/status_im/multiaccounts/login/core.cljs
```diff
@@ -221,26 +289,24 @@
...
+  (let [{:keys [address password save-password? creating?]} (:multiaccounts/login db)
         recovering? (recovering-multiaccount? cofx)
         login-only? (not (or creating?
                              recovering?
                              (keycard-setup? cofx)))
         nodes nil]
     (fx/merge cofx
               {:db (-> db
-                       (dissoc :multiaccounts/login)
+                       (assoc :multiaccounts/login {:save-password? save-password?})
```
yenda:
```suggestion
                       ;; :multiaccount/login MUST be dissoc as this is how the app knows the login is successful
                       (dissoc :multiaccounts/login)
```

yenda (dismissed):

yenda:
What is the point of the touch-id screen during onboarding, saying yes does nothing, this PR should remove it or make it do something

flexsurfer:
@yenda should've been removed here https://github.com/status-im/status-react/pull/8903

src/status_im/multiaccounts/login/core.cljs
flexsurfer: why? `:multiaccounts/login` isn't used for  login is successful decisions

yenda:
@errorists why don't we do touch-id the same way we do save passwords btw? with a switch in login screen? The current flow where you have to look in profile -> security to find the option, get logged out and have to check "save password" feels super clunky

src/status_im/multiaccounts/login/core.cljs
yenda: it is that is why the `+` menu is spinning forever in this PR

flexsurfer:
@yenda,  @andmironov realized that as well, but we decided to implement it separately 
image

andmironov:
Hey! Minor stuff I've spotted while testing on iPhone 10:

- Status asking for permission to use Face ID IMHO should happen in settings when the user enables Face ID. In the current build it happens when I restart the app after enabling Face ID;

- Check for Face ID happens not on top of the sign-in screen with the password field in the foreground but on top of a blank screen which feels a bit off. If biometry fails, the screens move in the background;

- No way to re-run the biometric check if I cancel or fail it. Should be an icon in the password input which triggers the check;

yenda:
@flexsurfer I disagree with implemented separately, the whole reason this PR moves so much code is because this is done in the settings instead of a simple switch

flexsurfer:
@yenda we still need it in settings even if we want to implement it on sign in screen

yenda:
@flexsurfer you would only need it to disable it, not to enable it. but you could also just consider that it is disabled when you logout, like we do for save password. So that you can implement later, because save password should be consistent as well, then logout shouldn't remove saved password/biometrics.

flexsurfer:
> * Status asking for permission to use Face ID IMHO should happen in settings when the user enables Face ID. In the current build it happens when I restart the app after enabling Face ID;

in that case, we need to ask biometric before logout

> * No way to re-run the biometric check if I cancel or fail it. Should be an icon in the password input which triggers the check;

if biometric is failed we need to logout user and ask for the password, it's not possible to use biometric without password, a user has to enter password

flexsurfer:
@yenda we had a long discussion here https://github.com/status-im/status-react/issues/8966, and decided to implement it like this, so we don't have biometric during onboarding, but after onboarding there should be a way how to enable it from settings,  we need it to be merged this week for audit, I'm not sure if we want to delay it, but we can continue discussion for sure in the issue

churik:
@flexsurfer we don't have option "Save password" for accounts, created with Keycard.
So I suppose it should be hidden for such accounts.

yenda:
This PR moves too much code so it is hard to follow how much it changed the login flow, but as far as I can tell, the app already logged in before you authenticate with biometrics, the node is started and dbs are unlocked. Biometrics should be checked before login with password.

flexsurfer:
@churik yeah good point

yenda:
> @yenda we had a long discussion here #8966, and decided to implement it like this, so we don't have biometric during onboarding, but after onboarding there should be a way how to enable it from settings, we need it to be merged this week for audit, I'm not sure if we want to delay it, but we can continue discussion for sure in the issue

yes so to avoid bugs, better just do what I propose with minimal changes to the login flow. Currently if you logout the touchid is silently removed as well just like the saved password anyway, so we could just have a toggle at login, and check biometrics before login so that the dbs are not unlocked and the node not started and the messages not fetched. Right now the biometric check happens in the middle of the login flow, for instance I receive my wallet events in the logs.

flexsurfer:
> This PR moves too much code so it is hard to follow how much it changed the login flow, but as far as I can tell, the app already logged in before you authenticate with biometrics, the node is started and dbs are unlocked. 

yes it doesn't change login flow much, there is just one step between login success signal from node and loading account data from db and views initialization 

> Biometrics should be checked before login with password.

Could you explain why? because it's a significant change.

yenda:
> Could you explain why? because it's a significant change.

no the PR does significant changes, but doing it like this would require far less.

> yes it doesn't change login flow much, there is just one step between login success signal from node and loading account data from db and views initialization

this is not just one step, the login success signals means what it means, the app is already logged in

yenda:
@corpetty if you could chime in

so what I propose for now is that currently the password is stored in the keychain using `.resetInternetCredentials rn/keychain address`, I would store auth-method the same way except key would be `address-auth-method`.

the flow would be:
- get auth-method
- if nil normal login
- if password get password and login
- if biometric authenticate with biometrics, if authentication works get password then login else go to login screen

churik:
@flexsurfer 
so to summarize issues:
**1. Remove biometric auth for accounts, created with Keycard**
**2. No icon is shown on Android 8.1 (LG V20)**
photo_2019-09-26 12 48 44
**3. If disable bio auth in phone settings it remains switched on in app in disabled state**

Steps:

- turn on "Lock app with Touch ID"
- delete all fingerprints
- try to login

Expected result:
bio auth is disabled automatically

Actual result:
error is shown; toggle is locked
sdlfkklsadjf

andmironov:
Just in case, here are the mockups

The sign in flow
https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=4567%3A29128

The settings
https://www.figma.com/file/bPS9GrvMr7LnH7vnkmuveQfd/Settings?node-id=576%3A0

rachelhamlin:
Hmm, I don't think what I'm experiencing is expected.

- On my first run-through, I got the native pop-up to save face ID after logging in again. On subsequent run-throughs, using different random names, I did not see the native pop-up again. Is the setting enabled for every account if I do it once on the device? FWIW, I agree w/ @andmironov that it would be preferable to run that flow -before- the user logs in again. Is that possible? 
- When I enable Face ID and then log out, it does not log me back in with Face ID. Save password is once again unchecked. [VIDEO](https://imgur.com/gT7Xdwi)
- Separate from biometric: `Save password` does not work as expected. If I save password, I expect it to be remembered on logout, no? Every time I log out, my PW & the setting are cleared. I'm unclear as to what `save password` actually ever did. XD @churik do you know?

yenda:
@rachelhamlin @churik @andmironov I don't think it is worth testing right now, as discussed with @flexsurfer ~~we are going for the login setting rather than profile setting first~~ (misunderstood that part, I don't understand the design choice here), and that completely changes (but simplifies a lot) this PR, so you are testing code that didn't pass code review yet and will most likely change radically

rachelhamlin:
Yup. So because @yenda is requesting code changes and @churik, @andmironov and I have issues with the feature from a user perspective, we can't merge this this week.

I want to understand:
1) Do @yenda's suggestions affect the security critical aspect of this code? 
2) From @corpetty, can we continue to work on this feature into next week and still make the cut? I understand status-go goes to audit first, and thus we might have a bit of a buffer here.

Corey is on vacation today/tomorrow, so I would suggest @flexsurfer continue to work on this, under the assumption that it will still make the security audit.

flexsurfer:
> * On my first run-through, I got the native pop-up to save face ID after logging in again. On subsequent run-throughs, using different random names, I did not see the native pop-up again. Is the setting enabled for every account if I do it once on the device? FWIW, I agree w/ @andmironov that it would be preferable to run that flow -before- the user logs in again. Is that possible?

this option isn't for the device but for every multiaccount , so if you enabled it for multiaccount it will work only for this multiaccount, for others you have to enable it separately, but I agree having a few multiaccounts is confusing

> * When I enable Face ID and then log out, it does not log me back in with Face ID. Save password is once again unchecked. [VIDEO](https://imgur.com/gT7Xdwi)

yes thats correct, because you logged out

> * Separate from biometric: `Save password` does not work as expected. If I save password, I expect it to be remembered on logout, no? Every time I log out, my PW & the setting are cleared. I'm unclear as to what `save password` actually ever did. XD @churik do you know?

Save password is needed if you close the app and open again, not log out, log out means - clear password

yenda:
> Save password is needed if you close the app and open again, not log out, log out means - clear password

@andmironov @flexsurfer 
this is another inconsistency:
- first it's called `logout` whereas login is now called `unlock`, so it should be called `lock multiaccount`, and then it would be expected that it locks the multiaccount and remove password/biometrics.
- when you want to switch to a different multiaccount, why do you have to erase the password? there is no technical reason why it needs to be done, right now there is just `logout` which is actually `lock` but there could also be `switch` or something which doesn't delete password/biometrics.

All of the above could be simplified with a first iteration that just has 2 switches in login screen, one for biometrics and one for password
Second phase would be to also allow the user to disable password and biometrics in settings, but not to enable, because you would do that when you login.
Third phase find a name for switch multiaccount (not easy if we want to avoid ever mentioning the word multiaccount) button and rename logout lock or find a better name for logout and ask user if he want to keep the password and biometric settings when he logsout

rachelhamlin:
> yes thats correct, because you logged out

Okay, noted re: logout. It's working when I manually close the app. 

But I'm still not seeing the native flow when I configure biometric from additional multiaccounts. If I enable Face ID, save password, and then quit app, Face ID works even though I haven't opted in on my device.

yenda:
> But I'm still not seeing the native flow when I configure biometric from additional multiaccounts. If I enable Face ID, save password, and then quit app, Face ID works even though I haven't opted in on my device.

Permissions are granted per app, so once you accept it once, it is never asked again. But this screen is pointless anyway, afaiu it is removed in the onboarding touchup PR

rachelhamlin:
> > But I'm still not seeing the native flow when I configure biometric from additional multiaccounts. If I enable Face ID, save password, and then quit app, Face ID works even though I haven't opted in on my device.
> 
> Permissions are granted per app, so once you accept it once, it is never asked again.

Ah, okay. That's what I meant here:

> On my first run-through, I got the native pop-up to save face ID after logging in again. On subsequent run-throughs, using different random names, I did not see the native pop-up again. Is the setting enabled for every account if I do it once on the device? FWIW, I agree w/ @andmironov that it would be preferable to run that flow -before- the user logs in again. Is that possible?

A bit strange, but I don't expect users to have multiple 'multiaccounts', so that's okay. 

What about the missing face ID icon on the login screen? [This screen](https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=4566%3A12). 

Also, is it possible to run the native flow along with the setting, or not? I think this is what @andmironov was also suggesting:
1. From settings screen, I enable biometric login. 

2 or 3. Native flow runs.

2 or 3. I'm told to `save password`.

4. Save password and log in

Serhy:
**Issue 4** (following @churik iteration)
On any real device (Android 8.1, Android 9, iPhoneX iOS12.4) I have endless spinner on `(+)` button on the Chats screen (Doesn't matter device support FaceID or not) after account creation or recover. So there is no way to join any chat. No issue with develop build.
Below are the steps and logs (dafault loglevel was set - I didn't see anything screaming in geth.log):
1) Install fresh PR build ( https://i.diawi.com/ZCvc34 or https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-190926-092035-e66c89-pr.apk for Android)
2) Create new account
3) Stay on Chats screen.

Logs: 
[Status-debug-logs.zip](https://github.com/status-im/status-react/files/3657541/Status-debug-logs.7.zip)

**Issue 5**
On Android 6.0.1 I can't create or recover account. App crashes right away when i tap 'Generate new key' (for account creation) or `Enter Seed Phrase` (to recover). It doesn't happen on nightly develop (26th of Sept version). Device (real Galaxy Note 4) has fingerprint enabled.
**Reproduction**:
1) Install Status PR build on Android 6.0.1
2) From the very first page after Status launch tap any of: `Get started -> Generate a key` or `Get started -> top left Back button` or `Access key -> Enter Seed phrase`
**Actual result:** 
Screenshot 2019-09-26 at 14 39 51

Logcat logs (grepped by Status PID):
[logcat.txt](https://github.com/status-im/status-react/files/3657567/14880prcoesscrashonstart.txt)

**Issue 6:** If to deny (on iPhoneX) system request to `...allow "Status" to use Face ID` after app relaunch -> the `Biometric authentication error. Unable perform biometric authentication (NOT_AVAILABLE)` error appears. Issue is that from that moment there is no way to enable FaceID for Status: each time user re-enables FaceID from app settings again na relaunches Status - the `Biometric authentication error` error appears.
To reproduce:
1) Install Status build
2) Create new account 
3) `Profile -> Privacy and Security -> Lock up with Face ID -> ON -> OK, save password`
4) Login with the valid password with enabled `Save password` option 
5) Close Status
6) Open Status
7) On `Do you want to allow "Status" to use FaceID?` pop-up tap `Don't allow` (Note: `Biometric authentication error` appears -> OK)
8) Now, try to enable FaceID again (repeat steps from 3rd till 6th)
Actual result: `Biometric authentication error` persists all the time. No way to enable FaceID.
Screenshot 2019-09-26 at 14 58 54

yenda:
> On any real device (Android 8.1, Android 9, iPhoneX iOS12.4) I have endless spinner on (+) button on the Chats screen (Doesn't matter device support FaceID or not) after account creation or recover. So there is no way to join any chat.

@Serhy https://github.com/status-im/status-react/pull/9033#discussion_r328522314
> it is that is why the + menu is spinning forever in this PR

this PR hasn't passed code review yet

flexsurfer:
> What about the missing face ID icon on the login screen? This screen.

we could discuss this, but from security POV if you failed faceid you must enter the password this time, next time you still will have faceid first

> Also, is it possible to run the native flow along with the setting, or not? I think this is what @andmironov was also suggesting:

yes i'll fix that, because i remember i asked in discussion but @andmironov said we don't need to ask native flow in settings, maybe i misunderstood

yenda:
@rachelhamlin 

> A bit strange, but I don't expect users to have multiple 'multiaccounts', so that's okay.

until we can have multiple chat accounts under a multiaccount, there is very valid reasons to have multiple multiaccount, and the fact that logout removes the password is actually quite annoying when you just want to switch between them. 

someone might also have a keycard account and a regular account, for when he doesn't have his keycard with him, because I am not sure we really thought that one through. You need the keycard everytime the app is killed.

rachelhamlin:
> > What about the missing face ID icon on the login screen? This screen.
> 
> we could discuss this, but from security POV if you failed faceid you must enter the password this time, next time you still will have faceid first

But that's not standard is it? With other apps, sometimes Face ID simply fails so I trigger it again no problem. I can see how it would be more secure to not allow that, but I think it works differently with some of my other apps. Perhaps I'm wrong. Not sure why any app includes a face ID on log-in, in that case.

> 
> > Also, is it possible to run the native flow along with the setting, or not? I think this is what @andmironov was also suggesting:
> 
> yes i'll fix that, because i remember i asked in discussion but @andmironov said we don't need to ask native flow in settings, maybe i misunderstood

Not sure, but if possible, let's move it to settings. 

@yenda need to understand better what you mean re: "the fact that logout removes the password is actually quite annoying when you just want to switch between them."

yenda:
> @yenda need to understand better what you mean re: "the fact that logout removes the password is actually quite annoying when you just want to switch between them."

I mean that if I have a multiaccount A and B and I want to switch between them, that doesn't mean I want to re-enter the password for both every time I want to switch

flexsurfer:
@yenda @andmironov @churik @Serhy pushed changes, can be re-reviewed and re-tested

churik:
Issues 1-3 are addressed. Waiting for e2e, @Serhy tests with Face ID and @yenda review.

src/status_im/init/core.cljs
```diff
@@ -75,7 +69,7 @@
 
 (fx/defn start-app [cofx]
   (fx/merge cofx
-            {::get-supported-biometric-auth         nil
+            {:get-supported-biometric-auth          nil
```
yenda, line 72: use an alias namespace here

src/status_im/multiaccounts/login/core.cljs
```diff
@@ -156,25 +157,32 @@
...
+        auth-method     (:auth-method db)
+        new-auth-method (if save-password?
+                          (when-not (or (= "biometric" auth-method) (= "password" auth-method))
+                            (if (= auth-method "biometric-prepare") "biometric" "password"))
+                          (when (and auth-method (not= auth-method "none")) "none"))]
     (fx/merge cofx
               {:db (assoc db :chats/loading? true)
                ::json-rpc/call
-               [{:method "mailservers_getMailserverTopics"
+               [{:method     "mailservers_getMailserverTopics"
```
yenda, line 169: why all the spaces here? it is not the same map

src/status_im/biometric_auth/core.cljs
```diff
@@ -1,93 +0,0 @@
-(ns status-im.biometric-auth.core
```
yenda, line 1: this makes sense to move this into auth.biometric.cljs but not in multiaccount considering the future changes coming

Serhy:
I've tested it with iPhoneX. Below is the summary of scenarios performed:
- New account. Create -> Lock up with FaceID -> **Save password** -> FaceID enabled as expected
- New account. Create -> Lock up with FaceID -> **Cancel Save password** offer. Toggle button disabled. Toggle button ON again -> Save password. Can login with FaceID.
- New account. Create -> Lock up with FaceID -> tap Save password -> on login page **disable Save password option** -> FaceID toggle in disabled state, no autologin option applied
- FaceID enabled -> Log out of account -> Login without `Save password` checkmark -> FaceID in OFF state.
- FaceID enabled. App closed. Open app and show face of your cat. **"Face not Recognised"** native dialog appears. Tap `Cancel` login page with FaceID button next to password displayed. From this step:
 * a) Tap FaceID button with your face to login: successful login and FaceID kept on.
 * b) Enter invalid password -> Wrong password tooltip
 * c) Enter valid password with ‘Save password’ checked. Re-open app -> FaceID login triggered
 * d) Enter valid password but Save password un-checked. Re-open app -> Have to login with password as expected.

- FaceID enabled. Toggle OFF FaceID. Re-open app -> have to login with password (expected)

- Fresh build installed. Create new account -> toggle FaceID button. The `Do you want to allow “Status” to user Face ID?”` Dialog appears -> **Tap ‘Don’t allow’** -> The `Biometric authentication error` appears. Now any time FaceID button is toggled the ‘Biometric authentication error’ appears. Until user will navigate to Status app settings and enable ‘Face ID’ option from there.
Screenshot 2019-10-02 at 14 56 21


I'm not sure only with the last item. If we could make more user friendly error message, in a similar way we do for when 'Disallow' camera permission, that would be really good.
So when user tap user icon from Profile screen -> Select from Gallery -> tap 'Dont allow' in 'Grant permission' dialog. Tap usericon again -> Select from Gallery: 
Screenshot 2019-10-02 at 15 01 58

@flexsurfer is it possible to change FaceID error message with the custom one similar to above?

flexsurfer:
@Serhy @rachelhamlin could you help with the copy for this error?

Serhy:
@rachelhamlin @andmironov would the next error message make sense when FaceID disabled in Status app settings and user is trying to enable Face ID from `Profile-> Privacy and Security`: 
Title: `Error`
Description: `To grant the required Face ID permission, please go to your system settings and make sure that Status > Face ID is selected` ?

andmironov:
all the latest sign-in/unlock screens are here https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=4566%3A12

flexsurfer:
done 
image

vitvly (approved):

Serhy:
Retested with iPhoneX (FaceID) and Galaxy S8 (Finger print), looks good to me too! 
Kudos to you @flexsurfer , thanks!

Serhy (approved):

andmironov:
Heya, great job here! I think it's pretty solid and is ready to be merged, but still, here are some minor things I've spotted wanted to ask if I'm missing something:

1. When I **re-enable** biometric lock, why is it asking me to go and sing out and sign in again? I assume that's expected behavior and is happening because of how we handle password saving?

2. Biometric disabled -> I switch on the "Lock app with Face ID" -> I get the native biometric check  -> I tap "Cancel". This means I did not enable the biometric yet, but already for some reason got checked. Feels a bit off. I don't know what we can do about it now but i'll keep that in mind for future iterations of this feature.

flexsurfer:
> 1. When I **re-enable** biometric lock, why is it asking me to go and sing out and sign in again? I assume that's expected behavior and is happening because of how we handle password saving?

it's because we clear password when you disable biometric, for security reasons

rachelhamlin:
Sorry, can you elaborate on this @andmironov? 

> Biometric disabled -> I switch on the "Lock app with Face ID" -> I get the native biometric check -> I tap "Cancel". This means I did not enable the biometric yet, but already for some reason got checked. Feels a bit off. I don't know what we can do about it now but i'll keep that in mind for future iterations of this feature.

I think I see what you mean. But it only happens after you enable on the system level, after it asks, "Do you want to allow "Status" to use Face ID?" 

If you choose not to allow, you don't have this issue. But once enabled natively, the Face ID check happens before you hit the "Cancel" button on subsequent run-throughs.

As for this: 

> When I re-enable biometric lock, why is it asking me to go and sing out and sign in again? I assume that's expected behavior and is happening because of how we handle password saving?

I want to note here for sanity that we will eliminate the need to save password after we work on the rest of EIP1581 & pin code. From @flexsurfer:

> so to finish password task, we need next
> 1) use key from  Andrea's PR instead password
> 2)implement pincode for login and signing transactions
> 3) update biometric login and implement for signing transactions
> 1 can't be released without 2 , so 1 and 2 one task

andmironov:
Here's my banking app as an example of (2), it does not trigger the Face ID check when I turn it on, the face id is checked only when logging in (sorry for the russian language in UI but it's pretty straightforward)

I'm turning the Face ID lock on and off
https://www.dropbox.com/s/b455932djk9floi/IMG_0216.MP4?dl=0

And here's Status
https://www.dropbox.com/s/y9ztli3ajxz9d8h/RPReplay_Final1570186677.MP4?dl=0

rachelhamlin:
> Here's my banking app as an example of (2), it does not trigger the Face ID check when I turn it on, the face id is checked only when logging in (sorry for the russian language in UI but it's pretty straightforward)
> 
> I'm turning the Face ID lock on and off
> https://www.dropbox.com/s/b455932djk9floi/IMG_0216.MP4?dl=0
> 
> And here's Status
> https://www.dropbox.com/s/y9ztli3ajxz9d8h/RPReplay_Final1570186677.MP4?dl=0

@flexsurfer is it possible to fix? We don't have to do so now. I just want to log it.

flexsurfer:
we can log it for sure, but i'm not sure we can fix it

PR #9038, opened 2019-09-25
build multiple APKs for different architectures

jakubgs:
In order to reduce size of APKs people will be downloading from PlayStore this PR changes `release` build to produce multiple APKs for different architectures(`armeabi-v7a` and `arm64-v8a`) as well as one universal APK which includes all of them.

Builds for PRs and nightlies remain unchanged and produce one APK for `armeabi-v7a`, `arm64-v8a`, and `x86`.

Changes:
* Replace `NDK_ABI_FILTERS` env variable with:
  - `ANDROID_ABI_SPLIT` to control if one or multiple APKs are produced
  - `ANDROID_ABI_INCLUDE` to control for which architectures APKs are built
* Replaces `APK_PATH` env variable with `APK_PATHS` in `Fastlfile` to upload multiple APKs
* Adjusts `ci/Jenkinsfile.android` to archive and upload multiple APKs
* Adds `renameAPKs` method in `ci/android.groovy` for dynamically renaming multiple APKs
* Adjusts `nix/mobile/android/targets/release-android.nix` to copy multiple APKs to `$out`

Successful manual `release` build: https://ci.status.im/job/tests/job/android-build-branch-test/325

split_apks

__Note:__ The `universal` APK is bigger(`~73MB`) than our normal ones(`~43MB`) because it includes `x86` and `x86_64` and that cannot be changed.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -93,16 +93,6 @@ project.ext.react = [
 apply from: "../../node_modules/react-native/react.gradle"
 
 /**
- * Set this to true to create two separate APKs instead of one:
- *   - An APK that only works on ARM devices
- *   - An APK that only works on x86 devices
- * The advantage is the size of the APK is reduced by about 4MB.
- * Upload all the APKs to the Play Store and people will download
- * the correct one based on the CPU architecture of their device.
- */
-def enableSeparateBuildPerCPUArchitecture = false
-
-/**
  * Run Proguard to shrink the Java bytecode in release builds.
  */
 def enableProguardInReleaseBuilds = false
@@ -184,10 +174,13 @@ android {
         multiDexEnabled true
         versionCode getVersionCode()
         versionName getVersionName()
-        ndk {
-            abiFilters getEnvOrConfig('NDK_ABI_FILTERS').split(';')
-        }
         missingDimensionStrategy 'react-native-camera', 'general'
+        /* this needs to be empty if we want APKs split by ABIs */
+        if (!getEnvOrConfig('ANDROID_ABI_SPLIT').toBoolean()) {
+            ndk {
+                abiFilters getEnvOrConfig('ANDROID_ABI_INCLUDE').split(";")
+            }
+        }
     }
     /**
      * Arbitrary project metadata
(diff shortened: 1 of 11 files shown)


fastlane/Fastfile
```diff
@@ -258,14 +258,17 @@ end
 
 platform :android do
   # Optional env variables
-  APK_PATH = ENV['APK_PATH'] || 'result/app.apk'
+  APK_PATHS = ["result/app.apk"]
```
pedropombeiro:
```suggestion
  APK_PATHS = ENV["APK_PATHS"]&.split(";") or ["result/app.apk"]
```

android/app/build.gradle
```diff
@@ -242,9 +235,9 @@ android {
     splits {
         abi {
             reset()
-            enable enableSeparateBuildPerCPUArchitecture
-            universalApk false  // If true, also generate a universal APK
-            include "armeabi-v7a", "arm64-v8a", "x86"
+            enable getEnvOrConfig('ANDROID_ABI_SPLIT').toBoolean()
+            include getEnvOrConfig('ANDROID_ABI_INCLUDE').split(";")
+            universalApk true
```
pedropombeiro, line 240: Do we want a universal APK in all scenarios, or just for e2e builds?

pedropombeiro (approved):
Great job! Just a question and a minor improvement.

android/app/build.gradle
jakubgs: I think it's useful for testers, don't you think?

fastlane/Fastfile
jakubgs: Oooh, I never heard about this: http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

android/app/build.gradle
pedropombeiro: cc @churik @Serhy what do you think? Is it worth producing and storing an APK that targets all platforms except for Android e2e tests?

yenda:
where can I find one of these apk? does it need to be merged before it work? currently if I click in the bot links it doesn't work

jakubgs:
@yenda you can find an example of a release build in the PR description:
https://ci.status.im/job/tests/job/android-build-branch-test/325
You can download them from there. As I said in the OP, this only affects `release` type builds.
Do you think we should also build a separate `x86`, or is it fine if `universal` covers that?

yenda:
@jakubgs I suppose you can merge this if it works?

jakubgs:
I was hoping for some manual testing of the universal and per-architecture APKs first.

Serhy:
Tested each of three builds on Android 9, Android 8.1, Android 6.0.1
Can install, create and recover account, join public chat and receive messages for the last 24h. 
Could not install `arm64-v8a` on Galaxy Note 4 and Galaxy J7 due to their cpu architecture (instruction set is `armabi-v7a`)

jakubgs:
@Serhy thanks for that, that is to be expected. When we upload those separate APKs to Play Store they will give the right one to the user automatically.

PR #9042, opened 2019-09-25
replace require-macro for defnstyle and defstyle

yenda:
- renamed the macros def and defn so that they are now used with aliased
namespace `styles/def` and `styles/defn` to force user to use aliased require
instead of require-macro and refer
- this makes sure the cljs file is required which includes the require for
platform ns needed after macroexpension

status: ready


vkjr (approved):

yenda:
@churik is it a known e2e fail?

yenda:
Tested on One Plus 7 Pro creating account and trying out different screens, + fixes compilation and runtime error on desktop

churik:
It is not related to your PR for sure. 
I'll take a look on it

PR #8630, opened 2019-07-24
Desktop two-pane UI with react-navigation v3

vkjr:
fixes #7399

### Summary
This PR contains desktop mobile UI updated to work with react-navigation v3

### Testing notes
Suggestion is to test only mobile platforms.

#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- CPU performance / speed of the app
- network consumption

### Steps to test

- Open Status
- ...
- Step 3, etc.

status: wip

PR #9056, opened 2019-09-27
upgrade status-go with geth 1.9.5

adambabik:


jakubgs:
I did a separate PR just for Go upgrade to not compound the issues: #9065

oskarth:
@adambabik How will this play with the ongoing security audit? That's not quite clear to me. If this gets in later, how does that impact compatibility? Is there a related spec PR?

cc @rachelhamlin FYI

oskarth:
and @corpetty

pedropombeiro:
@adambabik it looks like the build fails on Windows because of `gosigar`:

```
# github.com/status-im/status-go/vendor/github.com/elastic/gosigar/sys/windows
../vendor/github.com/elastic/gosigar/sys/windows/privileges.go:244:9: assignment mismatch: 2 variables but "github.com/status-im/status-go/vendor/golang.org/x/sys/windows".GetCurrentProcess returns 1 values
```

I already had issues with this library once and had to patch and make an upstream PR. Is it required for the upgrade?

pedropombeiro:
@adambabik this commit seems relevant to the failing Windows build: https://github.com/golang/sys/commit/855e68c8590b0d6a9d08863d2982eb8aeddd98d3

It looks like for a while, `GetCurrentProcess` was changed to return only one argument, before being reverted.

pedropombeiro:
That change is reflected here https://github.com/status-im/status-go/commit/9e971a052d0a51949664ba8cf16e96f2ddcfa820#diff-b17bb1df6ac4cec827bbb4813d22c685R181

pedropombeiro:
Fixed the Windows build by upgrading `golang.org/x/sys` and `golang.org/x/crypto` on status-go. Not sure if you agree with the approach for the short term @adambabik. I'm not sure if there's anything we can do in https://github.com/elastic/gosigar, because I don't see a `go.mod` file there.

adambabik:
@oskarth what kind of spec change do you have on mind? Does our audit verify go-ethereum and Whisper libraries as well? This is nothing more than just a library update.

adambabik:
Good catch @PombeirP! I merge the new changes from develop because the build failed due to some ClojureScript errors.

adambabik:
The same :/ 

```
[2019-10-01T19:30:19.636Z] Compiling ["index.desktop.js" "status-modules/cljs/i18n-raw.js" "status-modules/cljs/network-raw.js"] from ["components/src" "react-native/src/cljsjs" "react-native/src/desktop" "src" "env/prod" "prod"]...
[2019-10-01T19:31:15.861Z] WARNING: Use of undeclared Var status-im.ui.screens.desktop.main.chat.views/incoming-name at line 69 /home/jenkins/workspace/status-react_prs_linux_PR-9056/src/status_im/ui/screens/desktop/main/chat/views.cljs
[2019-10-01T19:31:15.861Z] Error encountered performing task 'cljsbuild' with profile(s): 'prod'
```

jakubgs:
Yeah, this issue was introduced by 42961f3a0ea92ddb9c0bfe9fce7129de0b7d883f and is fixed in #8630.

jakubgs:
Since #8630 has been merged rebasing should fix the desktop build failures.

oskarth:
@adambabik Perhaps this has been done already in a previous PR (says `develop` but not clear whether that is on 1.9.x to me), but for geth 1.9 there are major changes. See https://blog.ethereum.org/2019/07/10/geth-v1-9-0/ and specifically this:

> The PoW calculation in Whisper v6 was incompatible with Parity due to not fully adhering to the spec. This was fixed, but it also means that Whisped v6 shipped with Geth v1.9.0 is incompatible with previous versions.

oskarth:
Looking at develop commit hash `40e66e6186164a422e432fd754b8f86b958014f3` I see  https://github.com/status-im/status-go/commit/40e66e6186164a422e432fd754b8f86b958014f3#diff-37aff102a57d3d7b797f152915a6dc16 and https://github.com/status-im/status-go/blob/40e66e6186164a422e432fd754b8f86b958014f3/go.mod#L36 which seems to suggest the previous version was running under 1.8.x. Am I missing something here?

It'd be good to have clarity on this PoW change w.r.t v1 and the security audit, as this was initially raised a month ago (and before that as well), in e.g. protocol call.

pedropombeiro (approved):

corpetty:
I would also like some clarity here w.r.t. PoW change and us addressing it.  I was under the assumption that we fixed this months ago, is that right?

adambabik:
@corpetty no, it's only now with this PR. However, we are using extremely low PoW anyway so we actually do not use PoW protection at all.

oskarth:
@adambabik Regardless of our low PoW, that doesn't answer the underlying issue of compatibility. They call it out explicitly, and we should be explicit about which version of PoW we support and why. It doesn't make any sense to ship an old version in v1 and then do another release an incompatible version. From what I can tell here, this seems like a blocker for security audit and something we have to sort out asap, so there's clarity in (a) code (b) spec (c) commit hash for ToB

oskarth:
^ @PombeirP @andremedeiros @cammellos @rachelhamlin FYI.

oskarth:
Also cc @kdeme as he identified the spec mismatch in the first place here https://github.com/ethereum/go-ethereum/issues/18070 and fixed here https://github.com/ethereum/go-ethereum/pull/19753

kdeme:
Be careful, even though you are using extremely low PoW, this can still cause a lot of messages to be dropped.

PoW calculation is: `Pow  = 2**leading_zeros(pow_hash) / (size * ttl)`

So practically, you can do the worst case calculation easily. Due to a completely different `pow_hash` you will have at worst case 0 leading zeros. 
So e.g. if you set `ttl` to 10 and message size is e.g. 256 (I have no idea how much overhead Status messages have), and you are already a factor lower than your PoW requirement.

So yeah, If you already have to break compatibility with v1, and it is still possible to do so, then I suggest you try to get this in there still. 
Else the options are either PoW at practically 0 or implement both calculations at receiving end. Neither are nice.

andremedeiros:
@Serhy @churik could one of you do a sanity check on this build?

adambabik:
@oskarth 

> It doesn't make any sense to ship an old version in v1 and then do another release an incompatible version. From what I can tell here, this seems like a blocker for security audit and something we have to sort out asap, so there's clarity in (a) code (b) spec (c) commit hash for ToB

I agree 100%. We decided on breaking compatibility for V1 so this fix of PoW calculation for Whisper V6 should also be included in V1 release.

Regarding clarity:
* code -- this PR is to test that status-go with geth 1.9.5 works fine with the Android, iOS and Desktop versions. We had some troubles with the CI but all checks are green, e2e tests pass so I guess the only thing left is some manual testing. In status-go, it's https://github.com/status-im/status-go/pull/1617 where I bumped geth to 1.9.5.
* spec -- yes, we need a PR saying that we support Whisper spec and use Whisper V6 from geth 1.9.5. Do we have any PR related to compatibility changes in V1?
* Security Audit -- not sure how flexible they are and whether they will look into go-ethereum and Whisper codebases. https://github.com/status-im/status-go/pull/1617 looks massive due to `vendor/` but other changes are just in test files + geth and Whisper packages bumps.

@kdeme thanks for clarification on this. Yeah, in my opinion, we definitely should try to squeeze this in. We practically have PoW close to 0 (`0.002` to be precise; we need to confirm whether it's both for incoming and outgoing message) but like you explained it might still lead to some messages being dropped. Implementation of any ugly solutions is not an option in my opinion because we haven't released V1 so there won't be a better chance to do it nice.

kdeme:
@adambabik re dropping messages based on PoW @ incoming and outgoing.

In the Nimbus implementation following drops at a node can occur:
1. When an incoming message its PoW is lower than the PoW requirement of the node, the message will be dropped.
2. When an outgoing message its PoW is lower than the PoW requirement of the node. I don't think this is a literal requirement of the EIP but we do it not to leak any information that a certain message originates from that node.
3.  When an outgoing message its PoW is lower than the PoW requirement of the peer it is being send to by the node, this message will not be send (it's not really dropped). A whisper node should know the PoW requirements of all its peers, this is passed at handshake. Practically, between Status nodes, this will be the same PoW requirement I guess.

On top of that you then have filters which can also filter (not drop) based on PoW and other parameters.

It's been a while that I looked at the geth implementation, but iirc it is the same situation there. After all, functionally wise, inspiration was drawn from that implementation.

Serhy:
No regressions found here when tested PR build between iOS 12 and Android 8.1
- Can create/recover account and attach userpic
- Can send/receive (from offline) messages in 1-1/public/group chats in different networks (and same network as well)
- Can send transactions from wallet
- Can block/unblock user (and messaging stop receiving).
- Can transact in Dapp (buy and gift cryptokitty)
- Device syncing works fine between iOS and Android 
Ready for merge, @adambabik

jakubgs:
After [status-go#1617](https://github.com/status-im/status-go/pull/1617) was merged the commit in `develop` that adds `1.9.5` is [`26880b83`](https://github.com/status-im/status-go/commit/26880b83d7cf4e950ac11eb8b14464adf10f3cb8) and not `40e66e61` used in this PR.

I'd recommend tagging the `develop` commit with a `0.34.0-beta.0` version tag, and using the tag with `./scripts/update-status-go.sh` script to update `status-go-version.json` properly.

adambabik:
@jakubgs that's my plan :) Trying to create a release https://ci.status.im/blue/organizations/jenkins/status-go%2Fmanual/detail/manual/475/pipeline first.

jakubgs (approved):

adambabik:
How can I merge it after all checks get green?

jakubgs:
With the `./scripts/merge-pr.sh ${PR_ID}` script.

Issue #9037, opened 2019-09-25
Error states for new Tx overview: contract & network fees


rachelhamlin:
# Problem

We redesigned the transaction overview screen in #8026, but error states have not been updated to match.

There are two key error states we need to cover:
(1) When the user does not have enough ETH to pay for network fees (gas).
(2) Contract-related errors that are special cases arising from an inability of the contract to accurately calculate gas.

Initially we spec'd out these error states in #8026 and #8691, respectively—but as there is overlap between them, this issue merges them into one.

## Implementation

[Figma screens for insufficient network fee](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=3864%3A33253)
![image](https://user-images.githubusercontent.com/15470435/65598568-a4d32b80-df93-11e9-9c49-7076400b24a7.png)

[Figma screens for contract errors](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=3864%3A33253)
![image](https://user-images.githubusercontent.com/15470435/65598544-9b49c380-df93-11e9-854e-ff3065fe5320.png)

## Acceptance Criteria

The terminology to be used in app is now `network fee` rather than `gas`. 

There is a loading state utilizing spinner animation as the network fee is being calculated.

#### For the network fee error
When the **user** is initiating the transaction (from wallet, or in future, chat): 
- They will not make it past [this screen](https://prnt.sc/pardu1) if they have an insufficient number of tokens for the transaction. They will never see both errors at once on the final overview.
- They can, however, bump into the insufficient network fee error, if they are low on ETH.

When a **DApp** initiates the transaction:
- Either or BOTH errors can occur at once:
- The user may have enough tokens, but not enough ETH for gas;
- Or the user may have enough ETH for gas, but not enough tokens;
- Or the user may be lacking in both. 

The error states should accurately reflect this.  

Note that if both error states are displayed, tapping on either `info` bubble should display the `Not enough [token]` (_do not hard code ETH!_) or `Not enough ETH` tool tip. If one is open already and the other is tapped, the first tool tip is minimized, and the second is opened. 

#### For the contract error
The user is gated from signing the Tx unless they manually `Set custom fee`. Then they may continue with an additional warning about the likelihood of failure.

rachelhamlin:
[...]

PR #9063, opened 2019-09-27
Add currency conversions for signing sheet - Fixes #8027

### LINKED ISSUE #8027: [Wallet] Show values in the user's currency
Currently, the transaction overview shows only the token being transacted. It should also include the user's currency. 

[FIGMA LINK](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Mobile-wallet?node-id=3865%3A45)

**Expected behavior**
image

**Actual behavior**
image

acolytec3:
fixes #8027

### Summary

Displays the amount and fee associated with a transaction in the User's wallet's currency 

##### Functional
- wallet / transactions

### Steps to test
- Open Status
- Select Wallet from bottom taskbar
- Select account tile
- Select "Send"
- Fill out required transaction details
- Select "sign transaction"
- Validate that amount/fee values display a converted wallet currency amount next to them

status: ready

### DIFF
--- a/src/status_im/ui/screens/signing/views.cljs
+++ b/src/status_im/ui/screens/signing/views.cljs
@@ -46,11 +46,6 @@
 (defn separator []
   [react/view {:height 1 :background-color colors/gray-lighter}])
 
-(defn acc-text [txt1 txt2]
-  [react/nested-text nil
-   txt1 " "
-   [{:style {:color colors/gray}} txt2]])
-
 (defn displayed-name [contact]
   (if (or (:preferred-name contact) (:name contact))
     (multiaccounts/displayed-name contact)
@@ -218,9 +213,13 @@
                   sign                  [:signing/sign]
                   chain                 [:ethereum/chain-keyword]
                   {:keys [amount-error gas-error]} [:signing/amount-errors (:address from)]
-                  keycard-multiaccount? [:keycard-multiaccount?]]
+                  keycard-multiaccount? [:keycard-multiaccount?]
+                  prices                [:prices]
+                  wallet-currency       [:wallet/currency]]
     (let [display-symbol     (wallet.utils/display-symbol token)
-          fee-display-symbol (wallet.utils/display-symbol (tokens/native-currency chain))]
+          fee-display-symbol (wallet.utils/display-symbol (tokens/native-currency chain))
+          converted-value (* amount (get-in prices [(keyword display-symbol) (keyword (:code wallet-currency)) :price]))
+          converted-fee-value (* fee (get-in prices [(keyword fee-display-symbol) (keyword (:code wallet-currency)) :price]))]
       [react/view styles/sheet
        [header sign tx display-symbol fee fee-display-symbol]
        [separator]
@@ -239,24 +238,43 @@
               {:type  :small
                :title :t/send-request-amount
                :error amount-error
-               :accessories [[acc-text (if amount (str amount) "0")
-                              (or display-symbol fee-display-symbol)]]}]
-             [separator]])
-          [list-item/list-item
-           {:type        :small
-            :title       :t/network-fee
-            :error       gas-error
-            :accessories [[acc-text fee fee-display-symbol] :chevron]
-            :on-press    #(re-frame/dispatch
-                           [:signing.ui/open-fee-sheet
-                            {:content        (fn [] [sheets/fee-bottom-sheet fee-display-symbol])
-                             :content-height 270}])}]
-          [react/view {:align-items :center :margin-top 16 :margin-bottom 40}
-           (if keycard-multiaccount?
-             [sign-with-keycard-button amount-error gas-error]
-             [button/button {:on-press  #(re-frame/dispatch [:set :signing/sign {:type :password}])
-                             :disabled? (or amount-error gas-error)
-                             :label     :t/sign-with-password}])]])])))
+               :accessories [[react/nested-text {:style {:color colors/gray}}
+                              [{:style {:color colors/black}} (if amount (str amount) "0")]
+                              " "
+                              (or display-symbol fee-display-symbol)
+                              " • "
+                              [{:style {:color colors/black}} (if converted-value (str "~" (i18n/format-currency converted-value (:code wallet-currency)))
+                                                                  [react/activity-indicator {:color   :colors/gray
+                                                                                             :ios   {:size  :small}
+                                                                                             :android {:size :16}}])]
+                              " "
+                              (str (:code wallet-currency))]]}]
+             [separator]
+             [list-item/list-item
+              {:type        :small
+               :title       :t/network-fee
+               :error       gas-error
+               :accessories [[react/nested-text {:style {:color colors/gray}}
+                              [{:style {:color colors/black}} fee]
+                              " "
+                              fee-display-symbol
+                              " • "
+                              [{:style {:color colors/black}} (if converted-fee-value (str "~" (i18n/format-currency converted-fee-value (:code wallet-currency)))
+                                                                  [react/activity-indicator {:color   :colors/gray
+                                                                                             :ios   {:size  :small}
+                                                                                             :android {:size :16}}])]
+                              " "
+                              (str (:code wallet-currency))] :chevron]
+               :on-press    #(re-frame/dispatch
+                              [:signing.ui/open-fee-sheet
+                               {:content        (fn [] [sheets/fee-bottom-sheet fee-display-symbol])
+                                :content-height 270}])}]
+             [react/view {:align-items :center :margin-top 16 :margin-bottom 40}
+              (if keycard-multiaccount?
+                [sign-with-keycard-button amount-error gas-error]
+                [button/button {:on-press  #(re-frame/dispatch [:set :signing/sign {:type :password}])
+                                :disabled? (or amount-error gas-error)
+                                :label     :t/sign-with-password}])]])])])))
 
 (defn signing-view [tx window-height]
   (let [bottom-anim-value (anim/create-value window-height)


acolytec3:
@yenda How do I get access to the Jenkins logs to see what it doesn't like?  I've run the tests and linter and both pass at this point.

cammellos:
@acolytec3 does not seems related to the pr, tests fails because we run out of space on jenkins (maybe just make sure they pass locally in the meantime), @jakubgs could you have a look please?

jakubgs:
I've cleaned up the hosts, should be good now:
```
 $ ansible ci-slave-linux -m shell -a 'df -h /'

linux-05.do-ams3.ci.misc | CHANGED | rc=0 >>
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       310G   90G  221G  29% /

linux-04.do-ams3.ci.misc | CHANGED | rc=0 >>
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       310G   53G  258G  17% /

linux-01.do-ams3.ci.misc | CHANGED | rc=0 >>
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       310G   98G  213G  32% /

linux-03.do-ams3.ci.misc | CHANGED | rc=0 >>
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       310G   71G  240G  23% /

linux-02.do-ams3.ci.misc | CHANGED | rc=0 >>
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       310G   94G  217G  31% /
```
Please push again to rebuild everything.

acolytec3:
Done, squashed and updated.

Below is a screengrab of how the currency conversion looks.

Currency Conversion  

2 questions:
1) The original render in the issue shows a spacing dot of some sort between the crypto and wallet currency amounts and I wasn't sure if there was a specific character I should use as that spacer or if there is some prebuilt component to place there.  I just have a blank space currently.
2) Do you want me to insert the "~" for either or both currency fields?  Here again, the render has it displayed for the fee but not the value section.

yenda:
@errorists could you answer those questions?

@acolytec3 before a designer comes by to answer your question I think you should use an interpunct for now https://en.wikipedia.org/wiki/Interpunct and also be mindful of the font-color, the value is black. You can check the nested-text component and how it is used to achieve that within one text element. I'd use the tilde (~) for both

acolytec3:
@yenda I've incorporated a nested-text element into the presentation so now the text of the wallet-currency is now black.  I think the only thing left is now I have an odd spacing issue where there's a gap between the value and the currency symbol as shown below.  Does nested-text take a padding or margin style element that would shrink this gap?  I looked at the prop functions for nested text and it looks like all the styling options are font related and not spacing.

Currency Conv 2

Also, since I already had the code basically written, I went ahead and added the loading spinner as well.  The bounty can still be paid out to @alexjg since I know he was planning to work on it.  Below is a screen-grab.

spinner

errorists:
yes, Eric is right, both should use `~`. What I couldn't help but notice is the huuge space between value and currency, it's not right, should be a single space character. Also the space around the `•` bullet character (is that a bullet there? looks super small) should be a single space.

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -240,13 +245,19 @@
...
+                                                                              [react/activity-indicator {:color   :colors/gray
+                                                                                                         :ios   {:size  :small}
+                                                                                                         :android {:size :16}}]) (:code wallet-currency)]}]
              [separator]])
           [list-item/list-item
            {:type        :small
             :title       :t/network-fee
             :error       gas-error
-            :accessories [[acc-text fee fee-display-symbol] :chevron]
+            :accessories [[acc-text fee fee-display-symbol] "·" (if converted-fee-value [react/nested-text {:styles {:color :colors/black}} (str "~" (i18n/format-currency converted-fee-value (:code wallet-currency)))]
```
yenda: don't use `acc-text` here, if you look at the definition above in this namespace it is just a nested-text with 2 parts and here you have more. You also misused nested-text here, you should use nested-text instead of acc-text. here is an example of how to use it:

https://github.com/status-im/status-react/blob/496f3f1cc78703c3d629f9f458a0e966b04e5c61/src/status_im/ui/screens/privacy_policy/views.cljs

as you can see you give a global style and then you can have as many arguments as you want, strings or vector, the strings get the global style and the vectors are tuples of style + string, the style is merged with the global style so you only have to put what changes there

andmironov:
Hey! Spotted there are error tooltips on the screens, just wanted to make sure the error states stay consistent, here's an issue about those https://github.com/status-im/status-react/issues/9037#issuecomment-534986779

cc @rachelhamlin

hesterbruikman:
> Hey! Spotted there are error tooltips on the screens, just wanted to make sure the error states stay consistent, here's an issue about those [#9037 (comment)](https://github.com/status-im/status-react/issues/9037#issuecomment-534986779)

On the tooltip. There is a component here: https://github.com/status-im/status-react/blob/bf6b328996581c77fff437efcbf9d411e00d2ef9/src/status_im/ui/components/tooltip/styles.cljs

@yenda do you know if it's up to date, following these designs?: https://www.figma.com/file/cb4p8AxLtTF3q1L6JYDnKN15/Index?node-id=4290%3A1769

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -239,14 +243,26 @@
...
+                             (or display-symbol fee-display-symbol)
+                             " · " 
+                             [{:style {:color colors/black}} (if converted-value (str "~" (i18n/format-currency converted-value (:code wallet-currency)))
+                                                                   [react/activity-indicator {:color   :colors/gray
+                                                                                              :ios   {:size  :small}
+                                                                                              :android {:size :16}}])]
+                             " "
+                             (str (:code wallet-currency))]}
+             [separator]]
+              [list-item/list-item
```
yenda: careful with parens you slurped one too many here, list-item is not part of the accessories

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -239,14 +243,26 @@
...
-               :accessories [[acc-text (if amount (str amount) "0")
-                              (or display-symbol fee-display-symbol)]]}]
-             [separator]])
-          [list-item/list-item
-           {:type        :small
-            :title       :t/network-fee
-            :error       gas-error
-            :accessories [[acc-text fee fee-display-symbol] :chevron]
+               :accessories [react/nested-text {:style {:color colors/gray}} 
+                            [{:style {:color colors/black}} (if amount (str amount) "0")]
```
yenda: CI won't accept this indent, run:
`make shell` then `lein cljfmt fix' to fix it

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -46,16 +46,16 @@
...
-  [react/nested-text nil
-   txt1 " "
-   [{:style {:color colors/gray}} txt2]])
-
 (defn displayed-name [contact]
   (if (or (:preferred-name contact) (:name contact))
     (multiaccounts/displayed-name contact)
     (:address contact)))
 
+(defn acc-text [txt1 txt2]
```
yenda: no point moving this

yenda (changes requested):

src/status_im/ui/screens/signing/views.cljs
acolytec3: Thanks for catching that.  I meant to change this back to WIP as I'm struggling to get the style props to cooperate for the vectors in my nested-text element where they override the global style though maybe that misplaced parens is the issue.  I'll check back in on it later tonight and see if that resolves it.

src/status_im/ui/screens/signing/views.cljs
acolytec3: Will do!

src/status_im/ui/screens/signing/views.cljs
acolytec3: Sorry, human error here.  Was actually just going to suggest removing this since I was replacing the acc-text component with a more generic nested-text element in the sheets function.  From a style perspective, is it preferred to extract these kind of subcomponents out into their own function definition?  If so, I can just repurpose the existing acc-text function to incorporate the additional elements for the currency conversion.

acolytec3: I've gone ahead and removed the acc-text function definition since I'm not using it in my PR (which I think is now basically complete) and it's not used anywhere else..  Let me know if you want me to revise things further on this one.

acolytec3:
@yenda @errorists I've updated the PR with the fixes for the nested-text element and seems to have resolved the spacing elements as well.  I also changed out the bullet I was using to the larger one from @errorists's comment so let me know if there are any more visual things I need to work on.

Currency Conversion

errorists:
Looks good now, thanks @acolytec3

yenda (approved):

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -158,7 +153,7 @@
 
 (defn- signing-phrase-view [phrase]
   [react/view {:align-items :center}
-   [react/text {:style {:color colors/gray :padding-bottom 8}} (i18n/label :t/signing-phrase)]
+   [react/text {:style {:color colors/gray :t-bottom 8}} (i18n/label :t/signing-phrase)]
```
flexsurfer: what is `:t-bottom` ?

acolytec3: t for typo.  I've reverted this change.

andmironov (commented):
looks good to me as well!

churik:
two questions:
1) there is no value on transaction fee:
no_value
Should it be added separately @flexsurfer ?

2) for some currecncies (e.g MDL) - we don't have separate symbol, so screen looks like:
mdl_mad
Is it OK @andmironov ?

From my perspective PR is ready to merge.

rachelhamlin:
Thanks for calling these out @churik.

- Lack of symbol is okay from my POV. 
- Network fee conversion to local currency would be good for when we update that screen - [backlogged issue](https://github.com/status-im/status-react/issues/9039). Okay for now in my opinion.

Well done @acolytec3! Thanks for your contribution.

PR #9073, opened 2019-09-30
[#8945] "null is not an object" error when tap on sticker sent in chat

### LINKED ISSUE #8945: "null is not an object" error when tap on sticker sent in chat
***Type***: Bug

***Summary***: Tapping upon sticker sent in chat result in `null is not an object (evaluating 'n.lastIndexOf')` error. 
If to install sticker pack -> send sticker in chat and tap it - everything is OK and I redirected to the sticker pack with list of stickers, but the error appears after re-login

#### Expected behavior
Redirect to sticker pack when tap on sticker in chat

#### Actual behavior
`null is not an object (evaluating 'n.lastIndexOf')`  when tap on sticker in chat

Screenshot 2019-09-06 at 19 27 24

### Reproduction

- Join public chat
- Tap stickers icon and install `Status Cat` sticker pack
- Send new sticker in chat
- Relogin
- Navigate to public chat
- Tap on sticker in the chat view

### Additional Information

- Status version: Develop 0.14.0 (2019090603) 

flexsurfer:
fixes #8945

tested on ios simulator

### DIFF
--- a/src/status_im/stickers/core.cljs
+++ b/src/status_im/stickers/core.cljs
@@ -120,8 +120,8 @@
                  (navigation/navigate-to-cofx % :stickers-pack-modal pack)))))
 
 (fx/defn open-sticker-pack
-  [{{:stickers/keys [packs packs-installed] :keys [network] :as db} :db :as cofx} id]
-  (when (and id (string/starts-with? network "mainnet"))
+  [{{:stickers/keys [packs packs-installed] :networks/keys [current-network] :as db} :db :as cofx} id]
+  (when (and id (string/starts-with? current-network "mainnet"))
     (let [pack    (or (get packs-installed id)
                       (get packs id))
           contract-address (contracts/get-address db :status/stickers)]


vitvly (approved):

dmitryn (approved):

yenda:
@churik what happened with e2e tests here? The failures don't seem to be related to the changes

flexsurfer:
@yenda @churik @Serhy it seems we have broken e2e in develop, because i can see same results for all PRs

Serhy:
I reviewed above tests:
Those which failed with `Transaction with amount **** is not found in list of transactions...` failed because transaction were not mined in the next 6mins after it sent. This is etherscan info for above two addresses https://ropsten.etherscan.io/address/0x9164f86170290ba3080e37d19d1953d3c5945913 and https://ropsten.etherscan.io/address/0xf184747445c3B85CEb147DfB136067CB93d95F1D 
I am not sure yet why there is a bunch of `pending` transactions since I just sent tx between my accounts (in Ropsten) with 2Gwei price and it got confirmation within next 30 seconds... The same gas price I see as the latest tx for above addresses but why it is not picked up by miners (and so all the rest txs are pending as well)...

test_modify_transaction_fee_values - failed because signing of the respective tx were last > 10 sec. I'd say this is more app failure (common fail not this PR fail) rather than test fail. So this test clicks 'Sign' transaction button and what happened is that the process of signing tx (with the spinning spinner) lasts > 10 sec. 
But I'll increase the time to wait for transaction sign in process.

test_logcat_backup_recovery_phrase test: false failure (rare case here) where two words from sign in phrase are in logcat but they were used as part of valid logs (currently it was 'expand' and 'index' words). I'll update this test too separately.

yenda:
was this PR tested? does it fix the bug?

Serhy:
Tested with iOS 12.3, Android 8.1, Android 6.0.1 - can't reproduce the bug.

PR #9075, opened 2019-09-30
Fix ENS input when starting a new chat

### LINKED ISSUE #9046: Can't type ENS name in "New chat" screen (when start typing "0X04" appears)
### Description

*Type*: Bug
*Summary*: when start to type ENS name, can't finish because "0X04" appears on the start of the string, and it is not possible to delete it.
Also when you paste valid ENS - it is replaced by public key 

#### Expected behavior
can type ENS

#### Actual behavior
0X04 appears
lop

### Reproduction

- Open Status
- Tap on + > Start new chat
- Start typing ENS name

### Additional Information

- Status version: nightly 25/09/2019

vitvly:
Fixes #9046 

For some text inputs it's fine to preserve content when navigating away and back again. So the flag is introduced to text-input component implementation.

Context: when entering an ENS name, `set-new-identity` with a hex address argument is invoked every time a resolution is performed. Hence new chat name input becomes broken as it dynamically replaces a human-readable name with a hex string. In this context it is wrong to set default-value to the value of new-identity sub as it then contains a computed value, not the one entered into the input field.

### DIFF
--- a/src/status_im/ui/screens/add_new/new_chat/views.cljs
+++ b/src/status_im/ui/screens/add_new/new_chat/views.cljs
@@ -37,9 +37,10 @@
                                                   (re-frame/dispatch [:contact.ui/contact-code-submitted]))
                           :placeholder         (i18n/label :t/enter-contact-code)
                           :style               (add-new.styles/input @tw)
-                          ;; Set default-value as otherwise it will
-                          ;; be erased in global `onWillBlur` handler
-                          :default-value       new-identity
+                          ;; This input is fine to preserve inputs
+                          ;; so its contents will not be erased 
+                          ;; in onWillBlur navigation event handler
+                          :preserve-input?     true
                           :accessibility-label :enter-contact-code-input
                           :return-key-type     :go}]]
       (when-not platform/desktop?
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/add_new/new_chat/views.cljs
```diff
@@ -2,6 +2,7 @@
   (:require-macros [status-im.utils.views :as views])
   (:require [re-frame.core :as re-frame]
             [status-im.i18n :as i18n]
+            [taoensso.timbre :as log]
```
flexsurfer: leftover?

vitvly:
@flexsurfer @yenda can you please have a look

flexsurfer (approved):
this text-input hack getting more a more complex, maybe we're moving in the wrong direction

churik:
@siphiuel 
can't join to public chat in this PR (tested on IOS) - when click "Go" on keyboard, keyboard is closed.

churik:
Tested:
- issues is not reproducible
- input fields: password, public chat name, 1-1 chat name, message input, amount field
- that chat input is preserved when navigating between chats
- that password is erased when navigationg between accounts
Looks good to me

PR #9087, opened 2019-10-01
[#8657] Keyboard overlaps sign button if you tap on sign after changing "Recipient" or "Asset" field on Send Transaction screen

### LINKED ISSUE #8657: Keyboard overlaps sign button if you tap on sign after changing "Recipient" or "Asset" field on Send Transaction screen
### Description

*Type*: Bug
*Summary*: when you change  "Recipient" or "Asset"  after setting an amount and tapping Back, at attempt to sign transaction keyboard remains opened and overlaps "sign" button.

#### Expected behavior
no overlapping

#### Actual behavior
photo_2019-07-29 11 09 11

### Reproduction

- Open Status
- Go to wallet -> send
- Set recepient, asset, amount
- Change recepient or asset
- Try to sign transaction

### Additional Information

- Status version: nightly 29/07/2019, release 0.14.0 (build 16)
- Operating System: iOS

tbenr:
fixes #8657

I did several tests and the only way I found so far is to set `:keyboard-should-persist-taps` to `:never`.

btw the root cause seems to be that, if you navigate to asset or recipient when the amount input text has focus, when you navigate back the amount input automatically receives the focus again and keyboard pop-ups. This "focus restoration" generates a state where keyboard remains on top.

I don't know yet why\how the focus is auto restored when navigating back, but `:keyboard-should-persist-taps` to `:never` force input to always loose focus before navigating away.

can you check if this is an acceptable solution? The side-effect is that user, when amount text input has focus, has to tap twice to navigate to recipient or asset. To mitigate I added `:keyboard-dismiss-mode  :on-drag`.

### DIFF
--- a/src/status_im/ui/screens/wallet/send/views.cljs
+++ b/src/status_im/ui/screens/wallet/send/views.cljs
@@ -27,7 +27,8 @@
      [topbar/simple-toolbar (i18n/label :t/send-transaction)]
      [react/view components.styles/flex
       [wallet.components/network-info]
-      [react/scroll-view {:keyboard-should-persist-taps :always
+      [react/scroll-view {:keyboard-should-persist-taps :never
+                          :keyboard-dismiss-mode        :on-drag
                           :ref                          #(reset! scroll %)
                           :on-content-size-change       #(when (and scroll @scroll)
                                                            (.scrollToEnd @scroll))}


flexsurfer:
@siphiuel could you take a look into description? is it related to changes in the input field ?

vitvly:
@flexsurfer i've tried to comment the default-value setting in onBlur event handler, but this does not help fix the problem. So it should not be related to text-input clearing we do in onBlur.

vitvly (approved):

tbenr:
@siphiuel I did the same test during my investigations.

I also found this: https://stackoverflow.com/questions/29685421/hide-keyboard-in-react-native

I'd like to do some tests this evening.

tbenr:
@siphiuel @flexsurfer playing with on-press keyboard-dismiss but cant find a viable alternative solution.

churik:
@tbenr thanks for amazing work!
Tested on IOS and Android - setting focus outside field close keyboard, changing asset throw off amount value.
Failed e2e - infrastructure issue.

Ready to merge.

PR #9089, opened 2019-10-02
Fix api request to etherscan

Serhy:
In some test for transactions we check eth/token balances per address in etherescan via their api. Recently api request start failing, but same GET requests were working fine in browser. So I added user-agent in api call as additional header. 
Transaction relates tests should work now.


churik (approved):
🥇

PR #9096, opened 2019-10-02
[Fixes #9088] Store eip1581 path

### LINKED ISSUE #9088: Store eip1581 extended key
# Problem

After a discussion triggered by https://github.com/status-im/status-react/pull/9033 it was decided to store the extended key at path `m/43'/60'/1581'` so that in the future if and when we would like to derive an encryption key from this path it will be available to the user without having to input their seed phrase again.
This has no impact on the user experience, so it's just a technical task.

## Acceptance Criteria

- The key at path `m/43'/60'/1581'` should be stored encrypted on disk, similarly as we do with the chat key (`m/43'/60'/1581/0'/0'`) and wallet key (`m/44'/60'/0'/0/0`)
- Stop storing the master key
- Store instead of the master key the key at `m/44'/60'/0'/0/` and use that to derive wallets
- Ensure that this key can be accessed by the application
- Update https://notes.status.im/QxNC1H2cQMG7t8GvEb_Ctw with the relevant key informations

cammellos:
When creating the account we store as well the path specified in eip1581
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1581.md ,
`m / 43' / 60' / 1581'`.

The reason for doing so is that eventually we might want to derive an
encryption key from it, which would require the user to re-enter their
seed phrase if we would not store this.

This commit changes also the behavior **not to** store the master key, and instead store the key at 
`m /44'/60' /0'/0` from which wallets can be derived.

I could not test on keycard as I do not own one.

status: ready

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -237,12 +237,19 @@
 
 (def ^:const status-create-address "status_createaddress")
 
-(def ^:const path-root "m/44'/60'/0'/0")
-(def ^:const path-default-wallet "m/44'/60'/0'/0/0")
-(def ^:const path-whisper "m/43'/60'/1581'/0'/0")
+; BIP44 Wallet Root Key, the extended key from which any wallet can be derived
+(def ^:const path-wallet-root "m/44'/60'/0'/0")
+; EIP1581 Root Key, the extended key from which any whisper key/encryption key can be derived
+(def ^:const path-eip1581 "m/43'/60'/1581'")
+; BIP44-0 Wallet key, the default wallet key
+(def ^:const path-default-wallet (str path-wallet-root "/0"))
+; EIP1581 Chat Key 0, the default whisper key
+(def ^:const path-whisper (str path-eip1581 "/0'/0"))
 
 (def ^:const path-default-wallet-keyword (keyword path-default-wallet))
 (def ^:const path-whisper-keyword (keyword path-whisper))
+(def ^:const path-wallet-root-keyword (keyword path-wallet-root))
+(def ^:const path-eip1581-keyword (keyword path-eip1581))
 
 ;; (ethereum/sha3 "Transfer(address,address,uint256)")
 (def ^:const event-transfer-hash "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef")
--- a/src/status_im/wallet/accounts/core.cljs
+++ b/src/status_im/wallet/accounts/core.cljs
@@ -21,15 +21,15 @@
 
 (re-frame/reg-fx
  ::generate-account
- (fn [{:keys [address hashed-password path-num]}]
-   (status/multiaccount-load-account
-    address
-    hashed-password
-    (fn [value]
-      (let [{:keys [id error]} (types/json->clj value)]
-        (if error
-          (re-frame/dispatch [::generate-new-account-error])
-          (let [path (str constants/path-root "/" path-num)]
+ (fn [{:keys [derivation-info hashed-password path-num]}]
+   (let [{:keys [address path]} derivation-info]
+     (status/multiaccount-load-account
+      address
+      hashed-password
+      (fn [value]
+        (let [{:keys [id error]} (types/json->clj value)]
+          (if error
+            (re-frame/dispatch [::generate-new-account-error])
             (status/multiaccount-derive-addresses
              id
              [path]
@@ -45,7 +45,7 @@
                                         {:name (str "Account " path-num)
                                          :address address
                                          :public-key public-key
-                                         :path path
+                                         :path (str constants/path-wallet-root "/" path-num)
                                          :color (rand-nth colors/account-colors)}])))))))))))))
 
 (fx/defn set-symbol-request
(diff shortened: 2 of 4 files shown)


gravityblast (approved):
LGTM

cammellos:
@guylouis @dmitryn I am not able to test this change on keycard accounts as I don't own one.
The relevant changes is that the master key is not stored anymore on the filesystem for normal accounts (as I understand with a keycard this has never been the case, the master key is indeed stored, but only on the keycard).
For non-keycard accounts we store `m /44'/60' /0'/0 ` and use that to create new wallet accounts (which the keycard is not storing), and if that has not been created, we use the address of the master key instead (which will fail for non keycard accounts, but should work for keycards, as my understanding is that we do store the master key there).

Is that correct or I misunderstand something?

@churik are you able to test creating/recovering an account and generating new wallets (in both cases) on this?

churik:
OFC, on it!

corpetty:
It seems as though this was prematurely approved without testing out keycard.  Am I misreading that?  If that is the case, maybe a label like `needs testing` would be helpful as `status: ready` signals that nothing needs to be done even though it's intent was more along the lines of "the code is done."

churik:
Tested on Android 8:
- creating ordinary account (passcode with special chars) 
- restoring ordinary account (passcode with special chars) 
- creating account with keycard
- restoring account with keycard (create account, reinstall, restore)

All cases work fine for me.

churik:
@corpetty  we have such label `request-manual-qa`

cammellos:
@corpetty I think there's some confusion on the workflow:

`status: ready` means that the PR is ready to be reviewed (as opposed to status: wip which means that we haven't finished writing the code yet), not that is ready to be merged or that it has been tested (we add a label/move in the to merge review).

In terms of testing, up until last week all PRs went through manual QA, and I believe we add a label when manual QA is required. This one does not really need manual QA, other than someone needs to validate that it works with keycard, as specified in the comment (any developer can do).

@churik thanks, I'll wait for either @yenda or @flexsurfer to approve then is ready to merge.

yenda:
Does it means that a second account created with keycard will be different than one created with the same multiaccount without keycard?

cammellos:
@yenda No it should not, the path is identical,  `m /44'/60' /0'/0/x`, the only difference is that in the case of keycard it will be generated from the master key, while if it's not a keycard account, it will be generated from the `BIP44 Wallet Root key`.

But best to check that is the case:

@churik would you mind doing a test please if you have some time?

1) Import the same account once with keycard once without
2) Create an extra wallet on both, they should have identical addresses (the initial one as well)

Thanks!

cammellos:
mmh turns out that they do seem to differ (I tried with a pre this commit and this commit, not on keycard), it might be that the path is relative to the key, let me check

cammellos:
@yenda turns out we need to use the relative path, so if deriving from `m/a/b/c` you derive `m/0` for example, but if deriving from the master key, you'd derive `m/a/b/c/0`, which makes sense.

@churik could you have a look at the test with keycard when you have some time? It should show identical wallet addresses now.
Thanks!

churik:
@cammellos 

> Import the same account once with keycard once without

It shouldn't be possible according to https://github.com/status-im/status-react/issues/9062#issuecomment-537965163

> Create an extra wallet on both, they should have identical addresses (the initial one as well)

Due to https://github.com/status-im/status-react/issues/9019 it is not possible for now

So can't help here.

guylouis:
@cammellos 
just want to make sure: are we blocked here by something ?
can keycard team help in anyway ? cc @dmitryn

cammellos:
@guylouis No, I think it's fine, as one functionality is not supported (restore same account with keycard), and the other is not implemented yet (add multiple wallets with keycard).
Just waiting for a review from a clojure dev, otherwise is ready to merge. cc @flexsurfer @yenda

src/status_im/constants.cljs
```diff
@@ -237,12 +237,19 @@
...
 
-(def ^:const path-root "m/44'/60'/0'/0")
-(def ^:const path-default-wallet "m/44'/60'/0'/0/0")
-(def ^:const path-whisper "m/43'/60'/1581'/0'/0")
+; BIP44 Wallet Root Key, the extended key from which any wallet can be derived
+(def ^:const path-wallet-root "m/44'/60'/0'/0")
+; EIP1581 Root Key, the extended key from which any whisper key/encryption key can be derived
+(def ^:const path-eip1581 "m/43'/60'/1581'")
+; BIP44-0 Wallet key, the default wallet key
+(def ^:const path-default-wallet (str path-wallet-root "/0"))
```
flexsurfer, line 245: before "m/44'/60'/0'/0/0" after "m/44'/60'/0'/0"/0" is it the same path ?

flexsurfer: for some reason before it was "0/"

flexsurfer: was it mistake?

cammellos: wait, not sure i understand, path looks the same to me:

before:

m/44'/60'/0'/0
m/44'/60'/0'/0/0

after:

m/44'/60'/0'/0
(str "m/44'/60'/0'/0" "/0")

looks the same to me, or i am missing something?

flexsurfer: oh my, confused by "

flexsurfer:
@cammellos could you summarize why so many changes? I thought we just want to store an encryption key in this PR?

src/status_im/wallet/accounts/core.cljs
```diff
@@ -21,15 +21,17 @@
...
- (fn [{:keys [address hashed-password path-num]}]
-   (status/multiaccount-load-account
-    address
-    hashed-password
-    (fn [value]
-      (let [{:keys [id error]} (types/json->clj value)]
-        (if error
-          (re-frame/dispatch [::generate-new-account-error])
-          (let [path (str constants/path-root "/" path-num)]
+ (fn [{:keys [derivation-info
```
flexsurfer: just wondering why keys in a column and not in a row?

cammellos: i like columns :) but i can put them on a row, np

cammellos:
@flexsurfer that was he initial idea, but turns out we were also storing the master key on disk, which we should not have ( after a discussion with @corpetty), so I have changed the behaviour to store the root wallet key instead. It is explained as well in the issue

flexsurfer (approved):

PR #9103, opened 2019-10-03
Fix for 2 failed e2e

churik:
* Can request contact public key in Status test Dapp - failed because too fast tried to tap "status_api_button"

* test_send_funds_between_accounts_in_multiaccount_instance - failed because of 2 reasons:
1) account balance was fetched too fast (added waiting for confirmations of the last transaction)
2) adding two floats in Python wasn't equal to the actual balance on the wallet.
So special method `get_rounded_balance` was added to handle this case.
* reworked tests for blocked users to use new accounts instead of predefined.

### DIFF
--- a/test/appium/support/api/network_api.py
+++ b/test/appium/support/api/network_api.py
@@ -5,6 +5,7 @@ import pytest
 import requests
 import time
 from json import JSONDecodeError
+from decimal import Decimal
 
 
 class NetworkApi(object):
@@ -130,3 +131,10 @@ class NetworkApi(object):
         url = '%s/ping/%s?count=%s&interval=%s' % (self.chat_bot_url, chat_name, messages_number, interval)
         text = requests.request('GET', url).text
         return [i.split(maxsplit=5)[-1].strip('*') for i in text.splitlines()]
+
+    def get_rounded_balance(self, fetched_balance, actual_balance):
+        fetched_balance, actual_balance = str(fetched_balance), str(actual_balance)
+        # get actual number of decimals on account balance
+        decimals = abs(Decimal(fetched_balance).as_tuple().exponent)
+        rounded_balance = round(float(actual_balance), decimals)
+        return rounded_balance
--- a/test/appium/tests/atomic/transactions/test_wallet.py
+++ b/test/appium/tests/atomic/transactions/test_wallet.py
@@ -1,5 +1,4 @@
 import random
-
 import pytest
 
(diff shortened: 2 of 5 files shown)


test/appium/tests/atomic/transactions/test_wallet.py
```diff
@@ -487,9 +489,12 @@ def test_send_funds_between_accounts_in_multiaccount_instance(self):
         send_transaction.back_button.click()
         balance_of_sub_account = float(self.network_api.get_balance(sub_account_address)) / 1000000000000000000
         balance_of_status_account = float(self.network_api.get_balance(status_account_address)) / 1000000000000000000
-        expected_balance = str(float(balance_after_receiving_tx) - transaction_amount_1 - float(total_fee))
+        self.driver.implicitly_wait(60)
```
Serhy: No need in implicit wait if we confirm (by api_request) that status_account_address received transaction from sub_account. For this we need fix above two lines on to:
        self.network_api.wait_for_confirmation_of_transaction(status_account_address, transaction_amount_1)
        self.network_api.verify_balance_is_updated((updated_balance+transaction_amount_1), status_account_address)

test/appium/support/api/network_api.py
```diff
@@ -5,6 +5,7 @@
 import requests
 import time
 from json import JSONDecodeError
+from decimal import *
```
Serhy: No need to import everything from decimal, only Decimal is enough

test/appium/tests/atomic/transactions/test_wallet.py
```diff
@@ -461,7 +461,9 @@ def test_send_funds_between_accounts_in_multiaccount_instance(self):
         wallet_view.send_transaction_button.click()
         wallet_view.back_button.click()
         balance_after_receiving_tx = float(wallet_view.eth_asset_value.text)
-        if balance_after_receiving_tx != float(transaction_amount):
+        expected_balance = self.network_api.get_rounded_balance(str(balance_after_receiving_tx),
```
Serhy: We already have str value for this above, lets just remove float conversion for balance_after_receiving_tx, same with transaction_amount which is str already.

test/appium/tests/atomic/transactions/test_wallet.py
churik: Added! Thx a lot.

Serhy (approved):

PR #9132, opened 2019-10-07
Fixed e2e: return boolean instead of string

churik:
Sometimes test `test_mobile_data_usage_settings` failed because of wrong comparison of grabbed attributes.
Replaced with explicit comparison - checked 5 times locally, test is passed.

UPD: added retirn of boolean instead of string for cases 'true', 'True', 'false', 'False'

### DIFF
--- a/test/appium/tests/atomic/account_management/test_profile.py
+++ b/test/appium/tests/atomic/account_management/test_profile.py
@@ -109,7 +109,7 @@ class TestProfileSingleDevice(SingleDeviceTestCase):
         self.verify_no_errors()
 
 
-    @marks.testrail_id(6229)
+    @marks.testrail_id(5454)
     @marks.critical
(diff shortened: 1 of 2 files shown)


Serhy (changes requested):
Test passed 5 of 5 runs because with "is True"/"is False" it will always pass :)
attribute_value method returns 'true'/'false' as str type. Thus, 
`if "false" is True` ->  `False`
`if "false" is False` ->  `False`
`if "false" == "false"` -> `True`

And that's why `if profile_view.use_mobile_data.attribute_value("checked"):` always return `True`

For our needs here lets make is_checked method inside use_mobile_data/ask_me_when_on_mobile_network elements which will return boolean value

churik:
@Serhy ready to be reviewed again

Serhy (approved):

PR #9134, opened 2019-10-07
Feature/dapp accounts

### LINKED ISSUE #8977: [Multi-account] Allow user to choose wallet for DApp transactions
# Problem

**As a multiaccount user, I want to control which account I use to transact with DApps, so that (1) I never need to move assets between accounts to take advantage of DApps; and (2) my privacy can be better protected by not funneling all transactions through a single account.**

In v1, users can initiate transactions from the wallet screen (send/receive buttons) or from within a DApp. Chat commands are temporarily disabled.

Users deserve an option to select which of their accounts they transact with when using DApps. Currently, the 'default' account can not be changed. This severely limits the utility of multi-account. 

Update 2/10/19, after discussion btwn @flexsurfer, @simonricoo, @errorists, @hesterbruikman and @rachelhamlin:  

- Account selection will be global and apply to all previous tabs and DApps. 
- Cookies & data in previous tabs will be cleared.
- User will have to grant permission to DApp to access new account when opening again. 
- Out of scope: selection _within_ a DApp/tab.

A future, more privacy-centric model more in line with the [windows manager mentality](https://discuss.status.im/t/window-based-navigation/1374/7) would allow users to switch accounts or even have a dedicated account per DApp.

An additional future requirement might be to manage DApp permissions at an account level. They are currently global, tied to the current Status wallet. 

## Implementation

We've designed this control in context of the DApp browser.

[Figma screens](https://www.figma.com/file/MYWmd1buvc2AMvUmFP9w42t5/Browser?node-id=2998%3A0)

image

## Acceptance Criteria

- From the browser home screen, a user sees the ['Account name' option](https://www.figma.com/file/MYWmd1buvc2AMvUmFP9w42t5/Browser?node-id=2998%3A153) at the bottom, it hides on scroll. 
- Clicking on 'Status account' opens a bottom sheet with a scrolling list of all the user's accounts.
- The [DApp permissions modal](https://www.figma.com/file/MYWmd1buvc2AMvUmFP9w42t5/Browser?node-id=3000%3A5095) reflects the specific account a user is connecting with, rather than just "wallet" generically. 

**Out of scope for this issue**
- Selection from within a DApp/tab

flexsurfer:
fixes  #8977

https://www.youtube.com/watch?v=nQi784i4A0M

tested on ios simulator
change account
sign message from dapp
sign transaction from dapp


errorists:
hey @flexsurfer is that intentional that once you open a webview the little circle account button doesn't do anything on press? I was under the impression that we're including it in the toolbar to change the global account, currently opened website would get reloaded and its data cleared.

flexsurfer:
@errorists yes it's intentional, i thought that this change is for v1.5 , no ? because it may be confusing, that you may think you change account only for this browser tab

errorists:
@flexsurfer we discussed it in design and we think it won't be confusing since it operates exactly like the main button, it's just physically smaller, the description in bottom sheet also says that you're changing account to use in Dapps and not this single 'tab'. 
so v1 was having only the global switch on the root view
v1.5 was about having the global switch also on the webview
if we're adding the icon on the toolbar it should be able to change the account. If we're to leave it inactive I'd ask you to revert to v1 as in the previous toolbar without the icon.

flexsurfer:
@errorists ok thanks, i'll add it then

churik:
@flexsurfer 
### 1) If I have collectibles in one account, after switch to another in DApp I still can see them
I'm not sure it is an issue (it is DApp view) but I should know expected behavior here.

*Prerequisites:*
- have account with CK
*Steps:*
- create new account in wallet
- sign in both accounts in CK DApp
- switch accounts.

Not sure about expected results, I think preferable is to clear login data and show start page.
Currently if you, for example, on page with list of kitties in your account, after switching to another wallet you still see these kitties (but they are not belonging to it).
Each time you switched accounts Dapp requires permission and signing phrase.
Video is [here](https://drive.google.com/file/d/174CAmUbOtrqN4NYa4DoTup_R4TF1_hGl/view?usp=sharing).

### 2) No accounts in "DApp permissions"
Expected from my side to have ability to select what wallet can be used in Privacy and Security > Set DApp access permissions
wallet account

Testing is still in progress.

yenda (approved):

flexsurfer:
thanks @churik these 2 are expected, i'm not sure if we can fix this in this PR, but i think we can show account name instead wallet

churik:
@flexsurfer at least yes, please do.
Thank you!

flexsurfer:
image
@churik done

churik:
Tested on IOS 11.4.1. and Android 8.
- can sign transaction for DApp from different accounts
- permissions are resetting for all dapps in case of switching accounts 
- "Always allow" in DApps
- account is shown in permissions 
- can revoke access

TC is created (to be automated): https://ethstatus.testrail.net/index.php?/cases/view/6232

PR #9150, opened 2019-10-09
Clean /tmp and $TMPDIR in make clean

vitvly:
Sometimes Metro bundler cache/Yarn cache get broken and this issue manifests itself in weird errors during app startup, e.g. `ReactClassInterface: You are attempting to define 'UNSAFE_componentWillReceiveProps' on your component more than once. This conflict may be due to a mixin`.

The solution is to clean temp directories, on macOS at least Yarn and Metro bundler put their cache files into `/tmp` and `$TMPDIR` folders.

---
Changes:
* Define `TMPDIR` to be `/tmp/tmp-status-react-$(BUILD_TAG)` where `BUILD_TAG` is the commit
  - In case of Jenkins `BUILD_TAG` is a unique build identifier
* Add `TMPDIR` to `_NIX_KEEP` so it's passed to Nix shell and builds
* Add `_tmpdir-mk` target to `Makefile` to be called every time `make` is called
* Add `_tmpdir-rm` target to `Makefile` to be called with `make clean`
* Parallelize the 2 `Archive` and `Upload` stages to shorten the builds a bit
* Undo the use of Nix shell for `s3cmd` upload, since Nix isn't available on master
* Fixed `fastlane ios clean` job by adding missing `NIX_IGNORE_SYMLINK_STORE=1` (#9232)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: nix-add-gcroots clean nix-clean disable-githooks react-native-android react-native-ios react-native-desktop test release _list _fix-node-perms
+.PHONY: nix-add-gcroots clean nix-clean disable-githooks react-native-android react-native-ios react-native-desktop test release _list _fix-node-perms _tmpdir-mk _tmpdir-rm
 
 help: SHELL := /bin/sh
 help: ##@other Show this help
@@ -27,13 +27,18 @@ HELP_FUN = \
 		   }
 HOST_OS := $(shell uname | tr '[:upper:]' '[:lower:]')
 
+# This can come from Jenkins
+ifndef BUILD_TAG
+export BUILD_TAG = $(shell git rev-parse --short HEAD)
+endif
+
 # Defines which variables will be kept for Nix pure shell, use semicolon as divider
-export _NIX_KEEP ?= BUILD_ENV
+export _NIX_KEEP ?= TMPDIR,BUILD_ENV
 export NIX_CONF_DIR = $(PWD)/nix
 # We don't want to use /run/user/$UID because it runs out of space too easilly
-export TMPDIR = /tmp
-
-export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
+export TMPDIR = /tmp/tmp-status-react-$(BUILD_TAG)
+# this has to be specified for both the Node.JS server process and the Qt process
+export REACT_SERVER_PORT ?= 5001
 
 #----------------
 # Nix targets
@@ -89,9 +94,19 @@ _fix-node-perms: ##@prepare Fix permissions so that directory can be cleaned
 	$(shell test -d node_modules && chmod -R 744 node_modules)
 	$(shell test -d node_modules.tmp && chmod -R 744 node_modules.tmp)
 
+_tmpdir-mk: SHELL := /bin/sh
+_tmpdir-mk: ##@prepare Create a TMPDIR for temporary files
+	@mkdir -p "$(TMPDIR)"
+# Make sure TMPDIR exists every time make is called
+-include _tmpdir-mk
+
+_tmpdir-rm: SHELL := /bin/sh
+_tmpdir-rm: ##@prepare Remove TMPDIR
+	rm -fr "$(TMPDIR)"
+
 clean: SHELL := /bin/sh
-clean: _fix-node-perms ##@prepare Remove all output folders
-	git clean -dxf -f
+clean: _fix-node-perms _tmpdir-rm ##@prepare Remove all output folders
+	git clean -dxf
 
 watchman-clean: export _NIX_ATTR := targets.watchman.shell
 watchman-clean: ##@prepare Delete repo directory from watchman 
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -61,7 +61,7 @@ pipeline {
         script { cmn.nix.shell('lein deps :tree', attr: 'targets.leiningen.shell') }
       }
     }
-    stage('Parallel') {
+    stage('Parallel Assemble') {
       parallel {
         stage('Checks') { stages {
           stage('Lint') {
@@ -74,11 +74,6 @@ pipeline {
               script { cmn.nix.shell('lein test-cljs', attr: 'targets.leiningen.shell') }
             }
           }
-          /* stage('Coverage') {
-            steps {
-              script { android.coverage() }
-            }
-          } */
         } }
         stage('Build') { stages {
           stage('JSBundle') {
@@ -96,35 +91,41 @@ pipeline {
         } }
       }
     }
-    stage('Archive') {
-      steps { script {
-        apks.each { archiveArtifacts it }
-      } }
-    }
-    stage('Upload') {
-      steps {
-        script {
-          def urls = apks.collect { cmn.uploadArtifact(it) }
-          /* return only the universal APK */
-          if (urls.size() > 1) { 
-            env.PKG_URL = urls.find { it.contains('universal') }
-          } else { /* if no universal is available pick first */
-            env.PKG_URL = urls.first()
-          }
-          /* build type specific */
-          switch (btype) {
-            case 'release':
-              android.uploadToPlayStore(); break;
-            case 'nightly':
-              env.DIAWI_URL = android.uploadToDiawi(); break;
-            case 'e2e':
-              env.SAUCE_URL = android.uploadToSauceLabs(); break;
+    stage('Parallel Upload') {
+      parallel {
+        stage('Archive') {
+          steps { script {
+            apks.each { archiveArtifacts it }
+          } }
+        }
+        stage('Upload') {
+          steps {
+            script {
+              def urls = apks.collect { cmn.uploadArtifact(it) }
+              /* return only the universal APK */
+              if (urls.size() > 1) { 
+                env.PKG_URL = urls.find { it.contains('universal') }
+              } else { /* if no universal is available pick first */
+                env.PKG_URL = urls.first()
+              }
+              /* build type specific */
+              switch (btype) {
+                case 'release':
+                  android.uploadToPlayStore(); break;
+                case 'nightly':
+                  env.DIAWI_URL = android.uploadToDiawi(); break;
+                case 'e2e':
+                  env.SAUCE_URL = android.uploadToSauceLabs(); break;
+              }
+            }
           }
         }
       }
     }
     stage('Cleanup') {
-      steps { sh 'make clean' }
+      steps {
+        sh 'make clean'
+      }
     }
   }
   post {
--- a/ci/Jenkinsfile.combined
+++ b/ci/Jenkinsfile.combined
@@ -31,7 +31,7 @@ pipeline {
           osx    = cmn.ci.Build('status-react/combined/desktop-macos')
         } } }
         stage('Linux') { steps { script {
-          nix    = cmn.ci.Build('status-react/combined/desktop-linux')
+          tux    = cmn.ci.Build('status-react/combined/desktop-linux')
         } } }
         stage('Windows') { steps { script {
           win    = cmn.ci.Build('status-react/combined/desktop-windows')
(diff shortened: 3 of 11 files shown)


jakubgs (changes requested):
This will not work:
```
Cleaning Yarn/Metro caches in temp dirs ...
/bin/sh: 6: pushd: not found
/bin/sh: 9: popd: not found
/bin/sh: 6: pushd: not found
/bin/sh: 9: popd: not found
Makefile:90: recipe for target 'clean' failed
make: *** [clean] Error 127
```
This logic a bit clunky, there's a simpler way to do this by using a `status-react`-specific `TMPDIR`:
```Makefile
# We don't want to use /run/user/$UID because it runs out of space too easilly
export TMPDIR = /tmp/tmp-status-react
```
Then add a target that will run first on every `make` call to crate it:
```Makefile
#----------------
# Required targets
#----------------

mk-tmpdir: SHELL := /bin/sh
mk-tmpdir: ##@prepare Create a TMPDIR for temporary files
	@mkdir -p "$(TMPDIR)"

# Run these tasks every time make is called
-include mk-tmpdir
```
This should be above the Nix section in Makefile.
Then modify the `clean` target to look like this:
```Makefile
rm-tmpdir: SHELL := /bin/sh
rm-tmpdir: ##@prepare Remove TMPDIR
	rm -r "$(TMPDIR)"

clean: SHELL := /bin/sh
clean: _fix-node-perms rm-tmpdir ##@prepare Remove all output folders
	git clean -dxf -f
```
It would be also good to add both new targets at the end of `.PHONY`.

vitvly:
@jakubgs i tested this on macOS and it worked there, but maybe because my `/bin/sh` points to `/bin/bash`.

vitvly:
@jakubgs thanks for reviewing! A question - you suggest to use a separate $TMPDIR for status-react, but what about `/tmp`? These 2 folders are different, and both contain Yarn/Metro caches.

jakubgs:
>These 2 folders are different, and both contain Yarn/Metro caches.

Well, `TMPDIR` is a canonical unix variable: https://en.wikipedia.org/wiki/TMPDIR
It should affect where all well behaved programs store their temporary files.

But you might be right, `yarn` might not respect `TMPDIR`. Not sure about metro but Yarn has `YARN_CACHE_FOLDER` env variable we could set. It's worth a try.

Makefile
```diff
@@ -31,10 +31,21 @@ HOST_OS := $(shell uname | tr '[:upper:]' '[:lower:]')
...
-export TMPDIR = /tmp
+export TMPDIR = /tmp/tmp-status-react
 
 export REACT_SERVER_PORT ?= 5001 # any value different from default 5000 will work; this has to be specified for both the Node.JS server process and the Qt process
 
+#----------------
+# Required targets
+#----------------
+
+tmpdir-mk: SHELL := /bin/sh
```
pedropombeiro: I'd prefer if this was called `_tmpdir-mk`, to make it clearer that it is not something that's supposed to be used by the end user.

pedropombeiro (approved):

Makefile
jakubgs: That's fair.

jakubgs:
Damn, I added some commands before `make clean` and it doesn't seem like either Yarn or Metro respect the set `TMPDIR`:
```
[Pipeline] { (Cleanup)
[Pipeline] sh
15:03:21  + ls -l /tmp/tmp-status-react
15:03:21  total 0
[Pipeline] sh
15:03:21  + find /tmp/tmp-status-react
15:03:21  /tmp/tmp-status-react
[Pipeline] sh
15:03:22  + make clean
15:03:22  rm -r "/tmp/tmp-status-react"
15:03:22  git clean -dxf -f
```
https://ci.status.im/job/status-react/job/prs/job/android-e2e/job/PR-9150/3/

jakubgs:
@siphiuel are you referring in your description to the following files?
```
admin@linux-01.do-ams3.ci.misc:~ % sudo find /run/user/1001/ -name 'yarn--*' | head -n 3 
/run/user/1001/yarn--1571053553426-0.22263707960114654
/run/user/1001/yarn--1571052679528-0.7234799062122665
/run/user/1001/yarn--1571050818044-0.9092570746829618

admin@linux-01.do-ams3.ci.misc:~ % sudo find /run/user/1001/ -name '*-metro-*' | head -n 3
/run/user/1001/haste-map-metro-4-50919a9011980e131c55b2de66d3b833
/run/user/1001/haste-map-metro-4-d30b8c66ec93963468d6157fa1f0fa31
/run/user/1001/haste-map-metro-4-cbf3b0da2185ea8e534114174fb5d51f
```
Because if so then on CI hosts they don't end up in `/tmp` but in `/run/user/1001`, where `1001` is the UID of the `jenkins` user. I'd love to be able to control where they get created but it appears this is where they end up when run via Nix.

jakubgs:
According to `jest` docs:
>Default: "/tmp/<path>"
>The directory where Jest should store its cached dependency information.
>https://jestjs.io/docs/en/configuration#cachedirectory-string

This is the part of the `jest` package that creates the temporary file called `haste-map-metro-4-${H}`:
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-haste-map/src/index.ts#L303-L318
The directory in which it locates that cache file is defined via `cacheDirectory`, which comes from:
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-types/src/Config.ts#L27-L32
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-cli/src/cli/args.ts#L103-L108
Is created here:
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-runtime/src/index.ts#L198
Via the `createDirectory` function:
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-util/src/createDirectory.ts
There is a `--cacheDirectory` flag which can be set when calling `jest` in CLI, but we don't do that.

I need to find a way to customize it, because we can't just blindly remove the cache folder, since other CI builds on the same host could be using one of the directories we would try to remove.

jakubgs:
Ah, I see, by default if `cacheDirectory` is not specified it uses `os.tmpdir()`:
https://github.com/facebook/jest/blob/f3dab7cd/packages/jest-haste-map/src/index.ts#L248
Which according to NodeJS docs:
>The `os.tmpdir()` method returns a string specifying the operating system's default directory for temporary files.
>https://nodejs.org/api/os.html#os_os_tmpdir

Now, `operating system's default directory for temporary files` should be controlled by `TMPDIR`, which clearly isn't, and my guess is because we do not include it in allowed env variables for Nix via the `--keep` flag. I will test this.

pedropombeiro:
> Now, `operating system's default directory for temporary files` should be controlled by `TMPDIR`, which clearly isn't,

Yeah, it must be something like the `--keep` flag, because looking at the source code of node.js 10.15, it should be fetching `$TMPDIR` https://github.com/nodejs/node/blob/v10.15.0%5E0/lib/os.js#L131-L134.

jakubgs:
I managed to make it work. I added a `find` before cleaning the `TMPDIR` and the contents were what I expected them to be:

## MacOS
```
21:16:28  + make clean
21:16:30  find "/tmp/tmp-status-react-6035"
21:16:30  /tmp/tmp-status-react-6035
21:16:30  /tmp/tmp-status-react-6035/jenkins-state
21:16:30  /tmp/tmp-status-react-6035/jenkins-state/pid
21:16:30  /tmp/tmp-status-react-6035/jenkins-state/sock
21:16:30  /tmp/tmp-status-react-6035/jenkins-state/log
21:16:30  /tmp/tmp-status-react-6035/yarn--1571339785740-0.7473295541979732
21:16:30  /tmp/tmp-status-react-6035/yarn--1571339785740-0.7473295541979732/yarn
21:16:30  /tmp/tmp-status-react-6035/yarn--1571339785740-0.7473295541979732/node
21:16:30  /tmp/tmp-status-react-6035/v8-compile-cache-502
21:16:30  /tmp/tmp-status-react-6035/v8-compile-cache-502/6.8.275.32-node.45
21:16:30  /tmp/tmp-status-react-6035/v8-compile-cache-502/6.8.275.32-node.45/zSoptzSnixzSstorezS2f2ipyd1c27wrpl3prgvsqc82rhhf0rw-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.MAP
21:16:30  /tmp/tmp-status-react-6035/v8-compile-cache-502/6.8.275.32-node.45/zSoptzSnixzSstorezS2f2ipyd1c27wrpl3prgvsqc82rhhf0rw-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.BLOB
21:16:30  /tmp/tmp-status-react-6035/env-vars
21:16:30  rm -fr "/tmp/tmp-status-react-6035"
```
## Linux
```
21:05:27  + make clean
21:05:27  find "/tmp/tmp-status-react-6082"
21:05:27  /tmp/tmp-status-react-6082
21:05:27  /tmp/tmp-status-react-6082/jenkins-state
21:05:27  /tmp/tmp-status-react-6082/jenkins-state/log
21:05:27  /tmp/tmp-status-react-6082/jenkins-state/pid
21:05:27  /tmp/tmp-status-react-6082/jenkins-state/sock
21:05:27  /tmp/tmp-status-react-6082/env-vars
21:05:27  /tmp/tmp-status-react-6082/v8-compile-cache-1001
21:05:27  /tmp/tmp-status-react-6082/v8-compile-cache-1001/6.8.275.32-node.45
21:05:27  /tmp/tmp-status-react-6082/v8-compile-cache-1001/6.8.275.32-node.45/zSnixzSstorezSm6xj5bd7ssjhpnl6sm62i3fqxdd19a5s-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.MAP
21:05:27  /tmp/tmp-status-react-6082/v8-compile-cache-1001/6.8.275.32-node.45/zSnixzSstorezSm6xj5bd7ssjhpnl6sm62i3fqxdd19a5s-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.BLOB
21:05:27  /tmp/tmp-status-react-6082/yarn--1571339124002-0.15680803455239944
21:05:27  /tmp/tmp-status-react-6082/yarn--1571339124002-0.15680803455239944/node
21:05:27  /tmp/tmp-status-react-6082/yarn--1571339124002-0.15680803455239944/yarn
21:05:27  rm -fr "/tmp/tmp-status-react-6082"
```
## Windows
```
21:00:43  find "/tmp/tmp-status-react-2462"
21:00:43  /tmp/tmp-status-react-2462
21:00:43  /tmp/tmp-status-react-2462/jenkins-state
21:00:43  /tmp/tmp-status-react-2462/jenkins-state/sock
21:00:43  /tmp/tmp-status-react-2462/jenkins-state/pid
21:00:43  /tmp/tmp-status-react-2462/jenkins-state/log
21:00:43  /tmp/tmp-status-react-2462/yarn--1571338839511-0.30598898209784386
21:00:43  /tmp/tmp-status-react-2462/yarn--1571338839511-0.30598898209784386/node
21:00:43  /tmp/tmp-status-react-2462/yarn--1571338839511-0.30598898209784386/yarn
21:00:43  /tmp/tmp-status-react-2462/env-vars
21:00:43  /tmp/tmp-status-react-2462/v8-compile-cache-1001
21:00:43  /tmp/tmp-status-react-2462/v8-compile-cache-1001/6.8.275.32-node.45
21:00:43  /tmp/tmp-status-react-2462/v8-compile-cache-1001/6.8.275.32-node.45/zSnixzSstorezSm6xj5bd7ssjhpnl6sm62i3fqxdd19a5s-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.BLOB
21:00:43  /tmp/tmp-status-react-2462/v8-compile-cache-1001/6.8.275.32-node.45/zSnixzSstorezSm6xj5bd7ssjhpnl6sm62i3fqxdd19a5s-yarn-1.13.0zSlibexeczSyarnzSbinzSyarn.js.MAP
21:00:43  rm -fr "/tmp/tmp-status-react-2462"
```
## Android & iOS
It appears mobile platforms do not create any of those `yarn` temporary files. I assume that's due to more extensive use of Nix with tools like `yarn2nix`.

jakubgs (approved):

pedropombeiro (approved):

PR #9151, opened 2019-10-09
Test if recover seedphrase button disabled on Profile

Serhy:
- 1st test to check 'Back up seed phrase' not active after backup seedphrase for new account
- 2nd test to check  'Back up seed phrase' not active in recovered accounts

### DIFF
--- a/test/appium/tests/atomic/account_management/test_recover.py
+++ b/test/appium/tests/atomic/account_management/test_recover.py
@@ -12,33 +12,19 @@ from views.recover_access_view import RecoverAccessView
 @marks.account
 class TestRecoverAccountSingleDevice(SingleDeviceTestCase):
 
-    @marks.testrail_id(5301)
-    @marks.critical
-    @marks.battery_consumption
-    @marks.skip
-    def test_recover_account(self):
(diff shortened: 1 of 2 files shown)


test/appium/tests/atomic/account_management/test_recover.py
```diff
@@ -13,32 +13,18 @@
 class TestRecoverAccountSingleDevice(SingleDeviceTestCase):
 
     @marks.testrail_id(5301)
-    @marks.critical
-    @marks.battery_consumption
-    @marks.skip
+    @marks.medium
     def test_recover_account(self):
```
churik: Let's rename it?
i.e. test_recover_no_seed?

churik: Let's rename it?
i.e. test_recover_no_seed_phrase?

churik (approved):

test/appium/tests/atomic/account_management/test_recover.py
Serhy: Fixed

PR #9172, opened 2019-10-11
Upgrade xcode version to 11.1

pedropombeiro:

status: ready

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -2,19 +2,22 @@
 
 version: 1
 update_configs:
+  # Keep build.gradle files up to date, batching pull requests weekly
   - package_manager: "java:gradle"
     directory: "/android"
     update_schedule: "weekly"
     default_reviewers:
-    - jakubgs
+    - "jakubgs"
     allowed_updates:
       - match:
           update_type: "security"
+  # Keep /mobile/js_files/yarn.lock up to date, batching pull requests weekly
   - package_manager: "javascript"
     directory: "/mobile/js_files"
     update_schedule: "weekly"
     default_reviewers:
     - "jakubgs"
+  # Keep /desktop/js_files/yarn.lock up to date, batching pull requests weekly
   - package_manager: "javascript"
     directory: "/desktop/js_files"
     update_schedule: "weekly"
--- a/ci/Jenkinsfile.ios
+++ b/ci/Jenkinsfile.ios
@@ -1,5 +1,5 @@
 pipeline {
-  agent { label 'macos' }
+  agent { label 'macos-xcode-11.1' }
 
   parameters {
     string(
@@ -27,6 +27,7 @@ pipeline {
     LANGUAGE = "en_US.UTF-8"
     TARGET_OS = 'ios'
     NIX_CONF_DIR = "${env.WORKSPACE}/nix"
+    NIX_IGNORE_SYMLINK_STORE = 1 // https://github.com/NixOS/nix/issues/2925#issuecomment-499544039
     FASTLANE_DISABLE_COLORS = 1
     BUNDLE_PATH = "${HOME}/.bundle"
     /* We use EXECUTOR_NUMBER to avoid multiple instances clashing */
--- a/ci/Jenkinsfile.macos
+++ b/ci/Jenkinsfile.macos
@@ -1,5 +1,5 @@
 pipeline {
-  agent { label 'macos' }
+  agent { label 'macos-xcode-11.1' }
 
   parameters {
     string(
@@ -27,6 +27,7 @@ pipeline {
     LANGUAGE = "en_US.UTF-8"
     TARGET_OS = 'macos'
     NIX_CONF_DIR = "${env.WORKSPACE}/nix"
+    NIX_IGNORE_SYMLINK_STORE = 1 // https://github.com/NixOS/nix/issues/2925#issuecomment-499544039
     VERBOSE_LEVEL = '3'
     /* We use EXECUTOR_NUMBER to avoid multiple instances clashing */
     LEIN_HOME         = "/var/tmp/lein-${EXECUTOR_NUMBER}"
--- a/nix/mobile/default.nix
+++ b/nix/mobile/default.nix
@@ -6,7 +6,7 @@ let
 
   platform = callPackage ../platform.nix { inherit target-os; };
   xcodewrapperArgs = {
-    version = "10.2.1";
+    version = "11.1";
   };
   xcodeWrapper = composeXcodeWrapper xcodewrapperArgs;
   androidPlatform = callPackage ./android { inherit config target-os mkShell mkFilter nodejs maven localMavenRepoBuilder projectNodePackage jsbundle; status-go = status-go.android; };
--- a/scripts/sign-macos-pkg.sh
+++ b/scripts/sign-macos-pkg.sh
@@ -101,9 +101,9 @@ echo -e "\n### Signing object..."
 # If `OBJECT` is a directory, we assume it's an app
 # bundle, otherwise we consider it to be a dmg.
 if [ -d "$OBJECT" ]; then
-  codesign --sign "$DEV_ID" --keychain "$KEYCHAIN" --deep --force --verbose=4 "$OBJECT"
+  codesign --sign "$DEV_ID" --keychain "$KEYCHAIN" --options runtime --deep --force --verbose=4 "$OBJECT"
 else
-  codesign --sign "$DEV_ID" --keychain "$KEYCHAIN" --force --verbose=4 "$OBJECT"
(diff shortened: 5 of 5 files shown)


jakubgs (approved):

jakubgs:
The android tests seem broken:
```bash
[nix-shell:~/workspace/status-react_prs_android_PR-9172]$ lein test-cljs

;; ======================================================================
;; Testing with Node:

internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module 'react'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Module._load (internal/modules/cjs/loader.js:507:25)
    at Function.hookedLoader [as _load] (/home/jenkins/workspace/status-react_prs_android_PR-9172/target/test/test.js:14:10)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at /home/jenkins/workspace/status-react_prs_android_PR-9172/target/test/reagent/impl/util.js:11:139
    at Object.<anonymous> (/home/jenkins/workspace/status-react_prs_android_PR-9172/target/test/reagent/impl/util.js:17:3)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
Error encountered performing task 'doo' with profile(s): 'test'
Subprocess failed
```
Not sure how XCode change could have caused that, must be something unrelated.

jakubgs:
I tested exactly the same thing on my local machine and it works fine:
```
 $ IN_CI_ENVIRONMENT=1 nix-shell --run 'lein test-cljs' --pure --no-out-link --keep LOCALE_ARCHIVE_2_27 --keep IN_CI_ENVIRONMENT --argstr target-os android --attr targets.leiningen.shell default.nix                
Checking for modifications in node_modules...
...(omitted)...
Ran 219 tests containing 981 assertions.
0 failures, 0 errors.
```

jakubgs:
Okay, the only remaining failing build is MacOS one:
```
[2019-10-14T10:08:55.951Z] ### Assessing Gatekeeper validation...
[2019-10-14T10:08:56.631Z] StatusIm-191014-095652-6d1653-pr-universal.dmg: rejected
[2019-10-14T10:08:56.631Z] source=Unnotarized Developer ID
```
Seems like XCode upgrade changed something about the signing.
This is the step that failed:
https://github.com/status-im/status-react/blob/9ff21348f3827debc252e963f4609ec8347a6641/scripts/sign-macos-pkg.sh#L112-L117

jakubgs:
The MacOS signing issue seems to be clearly related to XCode upgrade based on this issue:
https://github.com/electron-userland/electron-builder/issues/3828

>Unfortunately, I also updated to OSX 10.14.5 and have the same problem.
>https://github.com/electron-userland/electron-builder/issues/3828#issuecomment-494444857

And this link seems to indicate it's a matter of adjusting options when running `codesign`:
https://stackoverflow.com/questions/53112078/how-to-upload-dmg-file-for-notarization-in-xcode

>```
>codesign -s "Developer ID Application: Name (ID)" MyApp.dmg --options runtime
>```

jakubgs:
This thread on Apple developer portal might shed some more light on the issue:
https://forums.developer.apple.com/thread/123480

>More specifically, if you want to sign using a Developer ID certificate, whatever you're signing has to be notarized. You might have used Mac Development certificates or some other when it worked.
> - kumowoon1025

And:

>Notarization is not something done to your developer ID.  It is a process specifically applied to your application, after it is code-signed.
> - jerryfrit

It appears something called `notarization` is required for our Developer ID to sign the MacOS app.

jakubgs:
This article seems to explain it:
https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution

>Notarization gives users more confidence that the Developer ID-signed software you distribute has been checked by Apple for malicious components.

>Apple notary service is an automated system that scans your software for malicious content, checks for code-signing issues, and returns the results to you quickly. If there are no issues, the notary service generates a ticket for you to staple to your software; the notary service also publishes that ticket online where Gatekeeper can find it.

The requirements are as follows:

>Notarization requires Xcode 10 or later. Building a new app for notarization requires macOS 10.13.6 or later. Stapling an app requires macOS 10.12 or later.
>
>Apple's notary service requires you to adopt the following protections:
>* Enable code-signing for all of the executables you distribute.
>* Enable the Hardened Runtime capability for your app and command line targets, as described in Enable hardened runtime.
>* Use a “Developer ID” application, kernel extension, or installer certificate for your code-signing signature. (Don't use a Mac Distribution or local development certificate.) For more information, see Create, export, and delete signing certificates.
>* Include a secure timestamp with your code-signing signature. (The Xcode distribution workflow includes a secure timestamp by default. For custom workflows, include the --timestamp option when running the codesign tool.)
>* Don’t include the com.apple.security.get-task-allow entitlement with the value set to any variation of true. If your software hosts third-party plug-ins and needs this entitlement to debug the plug-in in the context of a host executable, see Avoid the Get-Task-Allow Entitlement.
>* Link against the macOS 10.9 or later SDK.

This looks like a decent amount of work.

jakubgs:
I tested the two types of assessments that are done with `spctl` and `execute` one is accepted:
```
jenkins@macos:StatusImPackage % spctl --assess --type execute --verbose=2 Status.app 
Status.app: accepted
source=Developer ID
```
But the `open` type is rejected:
```
jenkins@macos-02:StatusImPackage % spctl --assess --type open --context context:primary-signature --verbose=2 Status.app
Status.app: rejected
source=Unnotarized Developer ID
```
Which is weird, because it's the `execute` one that should be called on the `Status.app` directory:
https://github.com/status-im/status-react/blob/9ff21348f3827debc252e963f4609ec8347a6641/scripts/sign-macos-pkg.sh#L112-L117

jakubgs:
Right, so it appears the DMG check fails, as I expected:
```
### Assessing Gatekeeper validation...
+ '[' -d StatusIm-191014-114512-2469cc-pr-universal.dmg ']'
+ spctl --assess --type open --context context:primary-signature --verbose=2 StatusIm-191014-114512-2469cc-pr-universal.dmg
StatusIm-191014-114512-2469cc-pr-universal.dmg: rejected
source=Unnotarized Developer ID
```

jakubgs:
The `--raw` flag provides a bit more info, but not very helpful:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>assessment:authority</key>
	<dict>
		<key>assessment:authority:flags</key>
		<integer>0</integer>
		<key>assessment:authority:row</key>
		<integer>15</integer>
		<key>assessment:authority:source</key>
		<string>Unnotarized Developer ID</string>
	</dict>
	<key>assessment:remote</key>
	<true/>
	<key>assessment:verdict</key>
	<false/>
</dict>
</plist>
```

jakubgs:
According to Apple article:
>Notarization gives users __more__ confidence that the Developer ID-signed software you distribute has been checked by Apple for malicious components.
>https://developer.apple.com/documentation/security/notarizing_your_app_before_distribution
The key word here being "more". I don't think this is mandatory, it's just nice to have, since otherwise the user installing the `.dmg` will probably see some red flags.

My first idea was to run the `spctl` command with some flags to let it pass despite lack of "notarization" but there might be no such flag.

jakubgs:
Based on [my comment above](https://github.com/status-im/status-react/pull/9172#issuecomment-541625484) which lists how many stars have to align to gain this mysterious "Notarization" status I think it would be prudent to just disable the `dmg` check we do for now, and fix it in a separate PR.

Considering that the MacOS app doesn't even start, it just crashes, if you just ignore the security warning, then fixing the signing verification seems pointless. I'm going to push a change that just disables this check temporarily.

jakubgs:
I've submitted a separate issue about this "notarization": https://github.com/status-im/status-react/issues/9212

Would you be fine with this PR in order to tackle the MacOS issues separately?
@vkjr @flexsurfer @yenda @churik @Serhy 

I don't want to block the XCode 11.1 upgrade (already 2/3 CI hosts are upgraded) just due to this, especially since people report that MacOS build crashes anyway.

Serhy:
> I've submitted a separate issue about this "notarization": #9212
> 
> Would you be fine with this PR in order to tackle the MacOS issues separately?
> @vkjr @flexsurfer @yenda @churik @Serhy
> 
> I don't want to block the XCode 11.1 upgrade (already 2/3 CI hosts are upgraded) just due to this, especially since people report that MacOS build crashes anyway.

I'm OK with that since Desktop is not in V1 list. Plus, it seems this behavior of #9212 I had with previous builds.

flexsurfer (approved):

andremedeiros (approved):

PR #9178, opened 2019-10-11
Verify ENS names

cammellos:
This commits verifies ens names when new messages or contact requests come through.

A batch of ens names is sent to status-go which will then verifying them
and the result will be passed back in a callback to status-react.

Currently this can only be partially tested as https://github.com/status-im/status-go/issues/1641 is unresolved, so very few messages comes through.

status: ready

### DIFF
--- a/src/status_im/ethereum/json_rpc.cljs
+++ b/src/status_im/ethereum/json_rpc.cljs
@@ -55,6 +55,7 @@
    ;;TODO not used anywhere?
    "shhext_deleteChat" {}
    "shhext_saveContact" {}
+   "shhext_verifyENSNames" {}
    "status_chats" {}
    "wallet_getTransfers" {}
    "wallet_getTokensBalances" {}
--- a/test/cljs/status_im/test/transport/core.cljs
+++ b/test/cljs/status_im/test/transport/core.cljs
@@ -1,6 +1,7 @@
 (ns status-im.test.transport.core
   (:require [cljs.test :refer-macros [deftest is testing]]
             [status-im.utils.fx :as fx]
+            status-im.transport.impl.receive
             [status-im.protocol.core :as protocol]
             [status-im.transport.core :as transport]
             [status-im.transport.message.core :as message]))
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -101,11 +101,11 @@
                 {:db (assoc-in db [:contacts/new-identity] "")}
                 (upsert-contact contact)))))
 
-(defn handle-contact-update
-  [public-key
+(fx/defn handle-contact-update
+  [{{:contacts/keys [contacts] :as db} :db :as cofx}
+   public-key
    timestamp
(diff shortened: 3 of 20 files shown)


src/status_im/ethereum/json_rpc.cljs
```diff
@@ -55,6 +55,7 @@
    ;;TODO not used anywhere?
    "shhext_deleteChat" {}
    "shhext_saveContact" {}
+   "shhext_verifyENSNames" {}
```
yenda, line 58: I thought this would entirely be done on go side?

cammellos: No, eventually maybe, but we are not parsing messages yet in status-go (we only do for the console client, but otherwise we would be parsing them twice, as we still have to port some of the handling, so it's disabled in status-react), and the ens name is inside the `content` field (or contact-request).
So status-react will parse the message and send the usernames to be verified to status-go (note that even if we did resolve them in status-go, it would be asynchronous, as resolving username can take up to 1,2 seconds, so we would pass the messages immediately and send another signal once resolved, so from a user perspective it would be asynchronous in both cases)

test/cljs/status_im/test/transport/core.cljs
```diff
@@ -56,15 +56,15 @@
...
-(deftest receive-whisper-messages-test
-  (testing "an error is reported"
-    (is (nil? (:chat-received-message/add-fx (message/receive-whisper-messages {:db {}} "error" #js [] nil)))))
-  (testing "messages is undefined"
-    (is (nil? (:chat-received-message/add-fx (message/receive-whisper-messages {:db {}} nil js/undefined nil)))))
-  (testing "happy path"
-    (let [actual (message/receive-whisper-messages {:db {}} nil messages sig)]
-      (testing "it add an fx for the message"
-        (is (:chat-received-message/add-fx actual))))))
+#_(deftest receive-whisper-messages-test
```
flexsurfer: why commented?

cammellos: oh leftover, forgot to uncomment it, thanks

flexsurfer (approved):

cammellos:
@churik would it be possible to disable this test? I can't merge status-go as we are still waiting to fix the geth 1.9 upgrade, but I would like to merge this code in so I don't have to rebase, i will re-enable the test once we merge status-go (and so we can verify the ens name), would that be ok?

churik:
@cammellos 
so it will be broken until merge  geth 1.9 upgrade?

cammellos:
@churik yes, until we merge 1.9 and include the feature for verifying ens names (current develop shows ens name even if they are not verified, so arguably is already broken :) )

churik:
@cammellos mind my commit when squashing. Thanks!

cammellos:
Thanks!

cammellos:
Merging this in, the feature is disabled currently until 1.9 issues are fixed, but at least I won't have to constantly rebase, I will enable once it sorted.

PR #9176, opened 2019-10-11
drop support of the LES fleet

jakubgs:
I've removed the LES fleet in https://github.com/status-im/infra-les/commit/47ee45872d193bd759f056f60e0c4b018dfe0ce7 since no one seemed to use it and it was a waste of money. In this PR I'm removing the ability to select it in the app.

PR #9204, opened 2019-10-14
drop les

flexsurfer:
upgrade for drop support of the LES fleet #9176


jakubgs (approved):
Nice! Thanks.

oskarth:
I'm not quite sure why we are doing this? LES is still relevant, it just hasn't been picked up and prioritized since Igor left. Bounties were supposed to be created to fix it but I don't think it happened, and since then it hasn't been on critical path for v1. Removing it seems shortsighted to me. See https://github.com/status-im/status-go/issues/1025

oskarth:
If it is about keeping the UI clean for v1 launch (as opposed to just hiding it in dev mode), I'd suggest hiding it behind a flag, as opposed to just removing it and making it harder for people to fix issues with it (once bounty is up).

jakubgs:
What would be the point of that flag? If we want this functionality in some uncertain future then we can just revert his commit. Otherwise we will just be keeping dead code and maintaining it without using it due to it being available via a flag.

flexsurfer:
@oskarth i would agree with @jakubgs , and part of the code looks stale, should be reimplemented anyway

PR #9208, opened 2019-10-14
fix navigation reset

yenda:
currently navigation-reset only works properly if you reset to a simple route

this fix allows us to use navigation reset with more complex routes, by
ensuring that the navigation stack is populated properly so that navigate
back doesn't end up emptying the stack.

this is temporary as the proper way to do navigation in general would be
to get rid of view id and navigation-stack entirely, since it is a duplication
of the state of react-navigation

status: ready

### DIFF
--- a/src/status_im/ui/screens/navigation.cljs
+++ b/src/status_im/ui/screens/navigation.cljs
@@ -49,9 +49,13 @@
 
 (fx/defn navigate-reset
   [{:keys [db]} {:keys [index actions] :as config}]
-  (let [view-id (:routeName (get actions index))]
-    {:db              (assoc db :view-id view-id
-                             :navigation-stack (list view-id))
+  (let [stack (into '() (map :routeName actions))
+        view-id (get stack index)]
+    {:db              (assoc db
+                             :view-id view-id
+                             ;;NOTE: stricly needs to be a list
+                             ;;because navigate-back pops it
+                             :navigation-stack stack)
      ::navigate-reset config}))
 
 (def unload-data-interceptor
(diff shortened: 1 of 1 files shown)


vitvly (approved):
Thanks @yenda! This means that in all places we use navigate-reset currently, we have to modify the params map passed so as not to have issues with navigate-back, is that the case?

yenda:
@siphiuel no it just means that now you can have more than one item in the actions vector, previously you could basically only reset to one of the top routes, there is an example in my ens PR

src/status_im/ui/screens/navigation.cljs
```diff
@@ -49,9 +49,13 @@
 
 (fx/defn navigate-reset
   [{:keys [db]} {:keys [index actions] :as config}]
-  (let [view-id (:routeName (get actions index))]
-    {:db              (assoc db :view-id view-id
-                             :navigation-stack (list view-id))
+  (let [stack (into '() (map :routeName actions))
```
flexsurfer, line 52: why list?

yenda: it is explained in the note below, navigate back uses `pop` which only works on list not lazyseq

yenda:
@churik is this e2e test suppose to pass?

churik:
@yenda for next times: look at Failure screenshot in test_decline_invitation_to_group_chat.
In this case unexpected popup from SauceLab is shown : https://assets.saucelabs.com/jobs/f5fcbf3713d64d95a24ac21c3b35a371/0011screenshot.png?auth=40582f53f961936feecc5448191721c7
So this is infrastructure failure and not related to your PR.

Issue #8997, opened 2019-09-13
Error 'Hzb.checkViodeoAuthorizationStatus in not a function' at attempt to scan QR on IOS


churik:
###  Description

*Type*: Bug

*Summary*: when scanning QR code getting error 'Hzb.checkViodeoAuthorizationStatus is not a function' instead of dialog about granting permission to camera. 
Have a device with IOS 11.4.1. only, so would appreciate if someone can check on iPhones with newer versions.

#### Expected behavior
can scan QR

#### Actual behavior
![error5](https://user-images.githubusercontent.com/4557972/64871768-3b275900-d646-11e9-8633-addef0f977e8.png)

### Reproduction

- Open Status
- Create account
- Tap on + > Scan QR code

### Additional Information

[comment]: # (Please do your best to fill this out.)
- Status version: nightly 13/09/2019
- Operating System: iOS 11.4.1.

Serhy:
The same with iOS 12.4.1 when opening camera in any place of app: 
`+ -> Scan QR code` 
`+ -> Start new chat -> button next to 'Enter chat key' input`
`Wallet -> Status account -> Send -> Specify recipient -> Scan QR code`
`Profile -> Edit upon profile userpic -> Capture`

PR #9209, opened 2019-10-15
Fix #8960

### LINKED ISSUE #8960: Can't scan wallet QR for transaction due to "network doesn't match" error
#### Description

*Type*: Bug

*Summary*: <no value> instead of current network id is shown when scanning QR to set recipient in a transaction screen.

#### Expected behavior
QR is scanned

#### Actual behavior

IMG_5FEC68E8E54B-1

### Reproduction

- Open Status
- Go to wallet, select account
- Tap on "Send"
- scan valid QR (mainnet - mainnet)

### Additional Information

- Status version: nightly 10/09/2019
- Operating System: Android, iOS (actually on android it is crashed now due to https://github.com/status-im/status-react/issues/8772, but I'm sure it is not platform-specific problem)

acolytec3:
fixes #8960 

### Summary

Issue as described in #8960 where using the camera to scan a QR code with a URL link throws an error.  

Solution fixes bug by pointing network to the correct key in the database.  

### Testing/Review notes
Metamask QR codes only work if your network is set to mainnet.  [Metamask doesn't really support EIP681](https://github.com/MetaMask/metamask-extension/issues/5125).  I don't know if it's possible to tell if you're scanning a Metamask QR code specifically but all they do is tack "ethereum" onto the front of the address so Status will throw the error message since it thinks you're reading an EIP681 compliant URI.

I tested using [this link generator](https://brunobar79.github.io/eip681-link-generator/#) and it processed the URI just fine and throws the appropriate error if the app is set to a different network than the URI calls for.

#### Platforms
Only tested on Android.  May be blocked by #8997 for iOS devices.  I can't verify as I don't have a Mac or iOS devices.

- Android
- iOS

##### Functional

- wallet/transfers

### Steps to test

- Open Status
- Go to Wallet -> Account -> Send
- Tap "Choose Recipient" and then "Scan QR Code"
- Scan QR code

status: ready

### DIFF
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -90,8 +90,8 @@
 
 (fx/defn fill-request-from-url
   {:events [:wallet/fill-request-from-url]}
-  [{{:keys [network] :wallet/keys [all-tokens] :as db} :db} data origin]
-  (let [current-chain-id                       (get-in constants/default-networks [network :config :NetworkId])
+  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db} data origin]
+  (let [current-chain-id                       (get-in constants/default-networks [current-network :config :NetworkId])
         {:keys [address chain-id] :as details} (extract-details data current-chain-id all-tokens)
         valid-network?                         (boolean (= current-chain-id chain-id))
         previous-state                         (get-in db [:wallet :send-transaction])


acolytec3:
@yenda @churik @flexsurfer @rachelhamlin  This one is ready to go.  I've only tested on Android since that's all I've got, but it works fine on my Oneplusu 6T with a build from the develop branch from a a day or two ago.

yenda (approved):

churik:
Failed e2e:
- `test_send_transaction_from_daap` - endless spinner after sending transaction
- `test_send_two_transactions_one_after_another_in_dapp` - Gas prise is not set by default

So both are not related to PR

churik:
@acolytec3 
awesome work!
Tested the following scenarios on IOS 11.4.1 and Android 8:
- QR address on the same network
- QR from MM (on the mainnet only)
- QR from wallet when on different network
- scanning QR from home page

PR #9211, opened 2019-10-15
Fixed several case ids and add test for checking log_level and fleet

churik:
status: ready


Serhy (approved):
Great, thanks @churik

Issue #8940, opened 2019-09-06
Redesign Back, swipe-to-the-left and hardware device Back buttons to align the behavior


asemiankevich:
# Problem

We have a lot of discrepancies between the following actions across all the application:

1. tapping Back button which is build-in the app
2. tapping Back button with is device hardware button
3. doing left-to-right swipe 

What we need to achieve:

- define [Back] and [Cancel] behavior (like when tapping Back is actually a cancel of action or tapping back should lead to previous screen)
- define where this swipe from left-to-right gesture is implemented and define what it should do (sending user to previous screen or something else)

This will resolve the following existing issues and prevent new ones:
- https://github.com/status-im/status-react/issues/8462
- https://github.com/status-im/status-react/issues/7437
- https://github.com/status-im/status-react/issues/8851
- a lot of issues with Keycard when i tap Back randomly and the app becomes completely broken

@andmironov @hesterbruikman @errorists i need your input here since it is a design question anyway.

@rachelhamlin

churik:
cc @hesterbruikman

errorists:
whew, ok here goes nothing :)

1. tapping <- back button, takes you to the previous screen in the current navigation stack. It is joined by the screen edge swipe, they are mutual and if the <- is present, you should be able to swipe. Unlike modal views (presented with an X instead of <-) where swipe back is not possible.
You should not be able to swipe when there is no back arrow meaning you can't go back to the previous screen in the navigation stack.
If it doesn't work as you expect it to it means someone botched up the navigation stack. A navigation stack is the linear sequence of screens you follow. it's like a powerpoint slide deck.

It's not a cancel action, but then again it would depend I guess on what you mean by cancel. It's like hitting the back button in your internet browser, if you loose data you had in that page then you might think of it as a cancel.

Now the Android back button, I have no clue really as I don't use an Android phone 😁  By Google's definition it does the same thing as the in-app Back <-/X so it's a duplicate. On newer devices it's replaced with a edge swipe gesture just like iOS. There are areas like the browser where we could point the software button to go backward to the previous page rather than exiting the webview.

In any case, it should act as similar as humanly possible to any natively written app. All of the issues you point to sound like implementation errors when dealing with React Navigation i.e. wrongly linked screens.

errorists:
I'd like to ask, because I might've missed some thread, what's the expectation of us, should we create documentation for what happens when you press back/swipe/hit device back button on each and every screen?

andmironov:
[...]

PR #9215, opened 2019-10-15
[#8940] Redesign Back, swipe-to-the-left and hardware device Back but…

flexsurfer:
partially #8940

the first step of navigation improvement, for v1 we want to avoid broken app states, so i added a list of views there it's possible to navigate back safely

### DIFF
--- a/src/status_im/ui/screens/navigation.cljs
+++ b/src/status_im/ui/screens/navigation.cljs
@@ -1,24 +1,9 @@
 (ns status-im.ui.screens.navigation
   (:require [re-frame.core :as re-frame]
-            [status-im.react-native.js-dependencies :as js-dependencies]
             [status-im.utils.handlers :as handlers]
             [status-im.utils.navigation :as navigation]
-            [status-im.utils.platform :as platform]
             [taoensso.timbre :as log]
             [status-im.utils.fx :as fx]))
-
-;; private helper fns
-
-(defn- push-view [db view-id]
-  (-> db
-      (update :navigation-stack conj view-id)
-      (assoc :view-id view-id)))
-
-;; public fns
-
-(fx/defn navigate-forget
-  [{:keys [db]} view-id]
-  {:db (assoc db :view-id view-id)})
 
 (defmulti unload-data!
   (fn [db] (:view-id db)))
@@ -32,7 +17,7 @@
 
 (defn- -preload-data! [{:keys [was-modal?] :as db} & args]
   (if was-modal?
-    (dissoc db :was-modal?) ;;TODO check how it worked with this bug
+    (dissoc db :was-modal?)
     (apply preload-data! db args)))
 
 (fx/defn navigate-to-cofx
@@ -44,7 +29,9 @@
                             screen-params))]
     {:db           (if (= view-id go-to-view-id)
                      db
-                     (push-view db go-to-view-id))
+                     (-> db
+                         (update :navigation-stack conj go-to-view-id)
+                         (assoc :view-id go-to-view-id)))
      ::navigate-to [go-to-view-id screen-params]}))
 
 (fx/defn navigate-reset
@@ -69,8 +56,6 @@
 (def navigation-interceptors
   [unload-data-interceptor (re-frame/enrich preload-data!)])
 
-;; effects
-
 (re-frame/reg-fx
  ::navigate-to
  (fn [[view-id params]]
@@ -89,8 +74,6 @@
    (log/debug :navigate-reset config)
    (navigation/navigate-reset config)))
 
-;; event handlers
-
 (handlers/register-handler-fx
  :navigate-to
  navigation-interceptors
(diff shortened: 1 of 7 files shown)


src/status_im/ui/screens/routing/back_actions.cljs
```diff
@@ -0,0 +1,29 @@
+(ns status-im.ui.screens.routing.back-actions)
+
+(def back-actions {:chat                              :default
```
yenda, line 3: shouldn't you use `:intro-wizard/navigate-back` for some screens here?

flexsurfer: no for now, this PR is only to prevent broken states of the app, later we could decide what to do with cancel actions

yenda: but what does it mean for this action? it prevents hardware button and 3 finger slide back but if you press the cancel button it works?

flexsurfer: yes

rachelhamlin:
@flexsurfer what should we test in this PR?

flexsurfer:
so now there is a list of screens  where you can use hardware back button on android or swipe on ios, for most onboarding screens and keycard screens hardware button or swipe will be unavailable

flexsurfer:
@siphiuel @dmitryn, please review

src/status_im/ui/screens/routing/back_actions.cljs
vitvly: Hardware back button is not prevented here, it still works.

src/status_im/ui/screens/routing/core.cljs
```diff
@@ -118,6 +118,13 @@
       (utils/update-if-present :initialRouteName name)
       (utils/update-if-present :mode name)))
 
+(defn new-get-state-for-action [default-get-state-for-action]
+  (fn [action state]
+    (let [event (get back-actions/back-actions @view-id)]
+      (when (and (= (.-type action) (.-BACK navigation/navigation-actions)) event (not= :default event))
+        (re-frame/dispatch [event]))
```
vitvly, line 136: This dispatch is never invoked currently. So this means that `create.core/intro-step-back` is never invoked on Android hardware back or iOS swipe, and `recover.core/cancel-pressed` is never invoked for actions included above + clicking on the Back arrow. So wizard state is not updated. This causes e.g. these issues:
 1) If one presses Android's hardware back on Onboarding's `Select account` screen, and then generates keys again, they are redirected to `Select key storage` screen bypassing `Select account` screen
2) On Basic recovery's `Select storage` screen clicking on Cancel is not going to show any confirmation dialog.

vitvly (changes requested):

src/status_im/ui/screens/routing/core.cljs
vitvly: Maybe there are some more, but maybe we can get rid of wizard state being dependent on a specific step

src/status_im/ui/screens/routing/back_actions.cljs
vitvly: I think there's no way to disable hw back button unless resorting to RN's `BackHandler`.

src/status_im/ui/screens/navigation.cljs
```diff
@@ -136,38 +118,4 @@
              {:db (-> db
                       (assoc :prev-tab-view-id (:view-id db))
                       (assoc :prev-view-id (:view-id db)))}
-             (navigate-to-cofx view-id {}))))
-
-;; This atom stores event vector
-;; to be dispatched when a react-navigation's BACK
-;; actions is invoked
-(def wizard-back-event (atom nil))
```
vitvly, line 148: Kudos for removing this atom in favor of `back-actions`, it was whacky.

src/status_im/ui/screens/routing/core.cljs
```diff
@@ -118,6 +118,13 @@
...
+      (default-get-state-for-action action state))))
```
vitvly, line 137: For now we're always popping the item from stack here, disregarding whatever might happen in `event` handler, but when Cancel processing is in place (and thus a possibility to block navigation action), how should this work? Currently `screens.navigation/processing-back-event?` exists, will there be some alternative?

flexsurfer:
@siphiuel thanks, added fix for Android https://github.com/status-im/status-react/pull/9215/files#diff-f65399a61bf325c3844d2b6b43bf1829R53
so my idea in this PR just to allow hardware button on  Android or swipe on iOS only on whitelisted screens

src/status_im/ui/screens/routing/core.cljs
flexsurfer: we could try to find an alternative after v1, for now Cancel screens just ignore hardware button or swipe

src/status_im/ui/screens/routing/core.cljs
flexsurfer: i didn't whitelist onboarding screens, so they will work fine because you can't use hardware buttons on them

flexsurfer: don't want to mess with wizard state in this PR, but we could try to get rid of it for sure

vitvly (approved):

PR #9219, opened 2019-10-16
desktop builds need GnuPG, also s3cmd

jakubgs:
Due to default shell changing to `zsh` in MacOS `10.15` the `PATH` has been affected and is causing failures due to missing tools like `s3cmd` and `gpg`, so I've just pulled those in via Nix.

After this one is merged `macos-01` and `macos-03` and be put back online in Jenkins.

### DIFF
--- a/nix/desktop/macos/default.nix
+++ b/nix/desktop/macos/default.nix
@@ -1,4 +1,4 @@
-{ stdenv, callPackage,
+{ stdenv, callPackage, pkgs,
   darwin, qt5, status-go, baseImageFactory }:
 
 with darwin.apple_sdk.frameworks;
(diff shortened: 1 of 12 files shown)


cammellos (approved):

corpetty (approved):

nix/desktop/macos/default.nix
```diff
@@ -11,6 +11,7 @@ let
 
 in {
   buildInputs = [
+    pkgs.gnupg22
```
pedropombeiro, line 14: @jakubgs why aren't we receiving `gnupg22` instead of `pkgs` in the argument list? We should avoid passing `pkgs` as much as possible, and in this case it doesn't seem to be required.

jakubgs: Good point, I'll fix it in another PR.

PR #9232, opened 2019-10-17
fix failing TestFlight clean job

jakubgs:
This fixes failing job that removes inactive TestFlight users using `fastlane clean` command.

Successful job on this branch: https://ci.status.im/job/status-tools/job/clean-testflight-users/640


jakubgs:
Included this fix in #9150.

PR #9240, opened 2019-10-19
Add ENS name resolution for EIP681 URIs and support in QR code scanner

### LINKED ISSUE #9238: EIP681 does not support ENS_NAME
Currently if you try to use EIP-681 using ENS names it gives an error about invalid link. 

Anywhere an address is expected, it should accept ENS names, for example, this should be valid:
`ethereum:pay-snt.thetoken.eth/transfer?address=unicorn.stateofus.eth&uint256=1e16`

image

The resolution is up to the payer - at any time between receiving the URL and sending the transaction - therefore it should be left to the UI that calls EIP681 to also call ENSRegistry, and the specification should only validate the ens name according EIP137

## [EIP681](https://eips.ethereum.org/EIPS/eip-681) says:

> ethereum_address        = ( "0x" 40*40HEXDIG ) / ENS_NAME

> For the syntax of ENS_NAME, please consult ERC#137 defining Ethereum Name Service.

> If using ENS names instead of hexadecimal addresses, the resolution is up to the payer, at any time between receiving the URL and sending the transaction. Hexadecimal addresses always take precedence over ENS names, i. e. even if there exists a matching ENS name consisting of 40 hexadecimal digits, it should never be resolved. Instead, the hexadecimal address should be used directly.

## [EIP137](https://eips.ethereum.org/EIPS/eip-137) says: 

> ENS names must conform to the following syntax:
> \<domain\> ::= \<label\> | \<domain\> "." \<label\>
> \<label\> ::= any valid string label per [UTS46](https://unicode.org/reports/tr46/)

### LINKED ISSUE #9357: ENS name is not resolved when scanned from QRCode
# Bug Report

After fixing #9238, it was noted that it caused this other bug.

## Problem

QRcode scanner uses EIP681 to read transaction requests. the specification allows ENS names to be used in this format, which was not happening. #9238 fixed it but the transaction screen still not resolving the ENS name, even if displaying it.

#### Expected behavior

This qr code scanned 
`ethereum:pay-snt.thetoken.eth/transfer?address=unicorn.stateofus.eth&uint256=1e16`

image

Allowing the transfer of token `snt.thetoken.eth` to `unicorn.stateofus.eth`

#### Notes

Related to #9240

acolytec3:
### Summary

EIP681 URI parsing does not currently support ENS name verification.  This PR enables the `parse-uri` function to accept properly formatted ENS names as the to address in a URI as well enables the QR scanner to properly resolve URIs with an ENS address.

This fixes #9238 and fixes #9357.

### Review notes
NA

### Testing notes
NA

#### Platforms

- Android
- iOS

Wallet/Send-Transaction/QR-Code

##### Functional

- wallet / transactions

##### Non-functional

N/A

### Steps to test

- Open Status
- Open Wallet
- Select any account
- Select Send
- Select specify-recipient
- Select 'Scan QR code'
- Scan the below QR code.
`ethereum:pay-snt.thetoken.eth/transfer?address=unicorn.stateofus.eth&uint256=1e16`
QR code

- Verify that ENS address is resolved to ethereum address and currency changes to SNT.

status: ready

### DIFF
--- a/src/status_im/ethereum/eip681.cljs
+++ b/src/status_im/ethereum/eip681.cljs
@@ -5,7 +5,9 @@
 
    e.g. ethereum:0x1234@1/transfer?to=0x5678&value=1e18&gas=5000"
   (:require [clojure.string :as string]
+            [re-frame.core :as re-frame]
             [status-im.ethereum.core :as ethereum]
+            [status-im.ethereum.ens :as ens]
             [status-im.ethereum.tokens :as tokens]
             [status-im.utils.money :as money]))
 
@@ -41,21 +43,34 @@
                {:function-arguments (apply dissoc m valid-native-arguments)}))
       arguments)))
 
-;; TODO add ENS support
-
 (defn parse-uri
-  "Parse a EIP 681 URI as a map (keyword / strings). Parsed map will contain at least the key `address`.
-   Note that values are not decoded and you might need to rely on specific methods for some fields (parse-value, parse-number).
+  "Parse a EIP 681 URI as a map (keyword / strings). Parsed map will contain at least the key `address` 
+   which will be either a valid ENS or Ethereum address.
+   Note that values are not decoded and you might need to rely on specific methods for some fields 
+   (parse-value, parse-number).
    Invalid URI will be parsed as `nil`."
   [s]
   (when (string? s)
     (let [[_ authority-path query] (re-find uri-pattern s)]
       (when authority-path
-        (let [[_ address chain-id function-name] (re-find authority-path-pattern authority-path)]
-          (when (ethereum/address? address)
-            (when-let [arguments (parse-arguments function-name query)]
-              (merge {:address address :chain-id (if chain-id (js/parseInt chain-id) (ethereum/chain-keyword->chain-id :mainnet))}
-                     arguments))))))))
+        (let [[_ raw-address chain-id function-name] (re-find authority-path-pattern authority-path)]
+          (when (or (ethereum/address? raw-address)
+                    (if (string/starts-with? raw-address "pay-")
+                      (let [pay-address (string/replace-first raw-address "pay-" "")]
+                        (or (ens/is-valid-eth-name? pay-address)
+                            (ethereum/address? pay-address)))))
+            (let [address (if (string/starts-with? raw-address "pay-")
+                            (string/replace-first raw-address "pay-" "")
+                            raw-address)]
+              (when-let [arguments (parse-arguments function-name query)]
+                (let [contract-address (get-in arguments [:function-arguments :address])]
+                  (if-not (or (not contract-address) (or (ens/is-valid-eth-name? contract-address) (ethereum/address? contract-address)))
+                    nil
+                    (merge {:address address
+                            :chain-id (if chain-id
+                                        (js/parseInt chain-id)
+                                        (ethereum/chain-keyword->chain-id :mainnet))}
+                           arguments)))))))))))
 
 (defn parse-eth-value [s]
   "Takes a map as returned by `parse-uri` and returns value as BigNumber"
--- a/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
+++ b/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
@@ -62,7 +62,7 @@
                        ;:torchMode     (camera/set-torch camera-flashlight)
                        :onBarCodeRead #(when-not @read-once?
                                          (reset! read-once? true)
-                                         (re-frame/dispatch [:wallet.send/resolve-ens-addresses (camera/get-qr-code-data %) :qr]))}]]
+                                         (re-frame/dispatch [:wallet.send/qr-code-request-scanned (camera/get-qr-code-data %) :qr]))}]]
       [viewfinder dimensions (size dimensions)]]
      [toolbar/toolbar
       {:center {:type                :secondary
--- a/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
+++ b/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
@@ -1,0 +1,72 @@
+(ns status-im.ui.screens.wallet.choose-recipient.views
+  (:require-macros [status-im.utils.views :refer [defview letsubs]])
+  (:require [re-frame.core :as re-frame]
+            [status-im.i18n :as i18n]
+            [status-im.ui.components.toolbar :as toolbar]
+            [status-im.ui.components.camera :as camera]
+            [status-im.ui.components.react :as react]
+            [status-im.ui.components.status-bar.view :as status-bar]
+            [status-im.ui.components.toolbar.actions :as actions]
+            [status-im.ui.components.toolbar.view :as topbar]
+            [status-im.ui.screens.wallet.choose-recipient.styles :as styles]
+            [status-im.utils.platform :as platform]
+            [status-im.ethereum.eip681 :as eip681]))
+
+(defn- topbar [camera-flashlight]
+  [topbar/toolbar
+   {:transparent? true}
+   [topbar/nav-button (actions/back-white actions/default-handler)]
+   [topbar/content-title {:color :white}
+    (i18n/label :t/wallet-choose-recipient)]
+   [topbar/actions [{:icon      (if (= :on camera-flashlight)
+                                  :main-icons/flash-active
+                                  :main-icons/flash-inactive)
+                     :icon-opts {:color :white}
+                     :handler   #(re-frame/dispatch [:wallet/toggle-flashlight])}]]])
+
+(defn- viewfinder [{:keys [height width]} size]
+  (let [height (cond-> height
+                 platform/iphone-x? (- 78))]
+    [react/view {:style styles/viewfinder-port}
+     [react/view {:style (styles/viewfinder-translucent height width size :top)}]
+     [react/view {:style (styles/viewfinder-translucent height width size :right)}]
+     [react/view {:style (styles/viewfinder-translucent height width size :bottom)}]
+     [react/view {:style (styles/viewfinder-translucent height width size :left)}]
+     [react/image {:source {:uri :corner_left_top}
+                   :style  (styles/corner-left-top height width size)}]
+     [react/image {:source {:uri :corner_right_top}
+                   :style  (styles/corner-right-top height width size)}]
+     [react/image {:source {:uri :corner_left_bottom}
+                   :style  (styles/corner-left-bottom height width size)}]
+     [react/image {:source {:uri :corner_right_bottom}
+                   :style  (styles/corner-right-bottom height width size)}]]))
+
+(defn- size [{:keys [height width]}]
+  (int (* 2 (/ (min height width) 3))))
+
+(defview choose-recipient []
+  (letsubs [read-once?        (atom false)
+            dimensions        [:dimensions/window]
+            camera-flashlight [:wallet.send/camera-flashlight]]
+    [react/view {:style styles/qr-code}
+     [status-bar/status-bar {:type :transparent}]
+     [topbar camera-flashlight]
+     [react/text {:style               (styles/qr-code-text dimensions)
+                  :accessibility-label :scan-qr-code-with-wallet-address-text}
+      (i18n/label :t/scan-qr-code)]
+     [react/view {:style          styles/qr-container
+                  :pointer-events :none}
+      [react/with-activity-indicator
+       {}
+       [camera/camera {:style         styles/preview
+                       ;:torchMode     (camera/set-torch camera-flashlight)
+                       :onBarCodeRead #(when-not @read-once?
+                                         (reset! read-once? true)
+                                         (re-frame/dispatch [:wallet.send/resolve-ens-addresses (camera/get-qr-code-data %) :qr]))}]]
+      [viewfinder dimensions (size dimensions)]]
+     [toolbar/toolbar
+      {:center {:type                :secondary
+                :disabled?           false
+                :on-press            #(re-frame/dispatch [:navigate-back])
+                :accessibility-label :cancel-button
+                :label               :t/cancel}}]]))
--- a/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
+++ b/src/status_im/ui/screens/wallet/choose_recipient/views.cljs
@@ -62,7 +62,7 @@
                        ;:torchMode     (camera/set-torch camera-flashlight)
                        :onBarCodeRead #(when-not @read-once?
                                          (reset! read-once? true)
-                                         (re-frame/dispatch [:wallet.send/qr-code-request-scanned (camera/get-qr-code-data %) :qr]))}]]
+                                         (re-frame/dispatch [:wallet.send/resolve-ens-addresses (camera/get-qr-code-data %) :qr]))}]]
       [viewfinder dimensions (size dimensions)]]
      [toolbar/toolbar
       {:center {:type                :secondary
--- a/src/status_im/ui/screens/wallet/components/views.cljs
+++ b/src/status_im/ui/screens/wallet/components/views.cljs
@@ -6,10 +6,305 @@
             [status-im.ui.components.react :as react]
             [status-im.ui.components.styles :as components.styles]
             [status-im.ui.components.toolbar.view :as topbar]
+            [status-im.ui.components.tooltip.views :as tooltip]
+            [status-im.ui.screens.chat.photos :as photos]
+            [status-im.ui.screens.wallet.choose-recipient.views
+             :as
+             choose-recipient]
+            [status-im.ethereum.ens :as ens]
             [status-im.ui.screens.wallet.components.styles :as styles]
             [status-im.ui.components.text-input.view :as text-input]
             [status-im.ui.components.colors :as colors])
   (:require-macros [status-im.utils.views :as views]))
+
+;; Wallet tab has a different coloring scheme (dark) that forces color changes (background, text)
+;; It might be replaced by some theme mechanism
+
+(defn text-input [props text]
+  [react/text-input (utils.core/deep-merge {:placeholder-text-color colors/gray
+                                            :selection-color        colors/black
+                                            :style                  {:color     colors/black
+                                                                     :height    52}}
+                                           props)
+   text])
+
+(def default-action (actions/back-white actions/default-handler))
+
+(defn topbar [title]
+  [topbar/simple-toolbar title])
+
+(defn- top-view [avoid-keyboard?]
+  (if avoid-keyboard?
+    react/keyboard-avoiding-view
+    react/view))
+
+(defn simple-screen
+  ([toolbar content] (simple-screen nil toolbar content))
+  ([{:keys [avoid-keyboard?]} toolbar content]
+   [(top-view avoid-keyboard?) {:flex 1}
+    [status-bar/status-bar]
+    toolbar
+    content]))
+
+(defn- cartouche-content [{:keys [disabled?]} content]
+  [react/view {:style (styles/cartouche-content-wrapper disabled?)}
+   [react/view {:flex 1}
+    content]])
+
+(defn cartouche [{:keys [disabled? on-press icon icon-opts] :or {icon :main-icons/next} :as m} header content]
+  [react/view {:style styles/cartouche-container}
+   [react/text
+    header]
+   (if (or disabled? (nil? on-press))
+     [cartouche-content m content]
+     [react/touchable-highlight {:on-press on-press}
+      [react/view
+       [cartouche-content m
+        (if-not (true? disabled?)
+          [react/view styles/cartouche-icon-wrapper
+           [react/view {:flex 1} ;; Let content shrink if needed
+            content]
+           [vector-icons/icon icon icon-opts]]
+          content)]]])])
+
+(defn view-asset [symbol]
+  [react/view
+   [react/i18n-text {:style styles/label :key :wallet-asset}]
+   [react/view styles/asset-container-read-only
+    [react/text {:style styles/asset-text}
+     (name symbol)]]])
+
+(defn- type->handler [k]
+  (case k
+    :send    :wallet.send/set-symbol
+    :request :wallet.request/set-symbol
+    (throw (str "Unknown type: " k))))
+
+(defn- render-token [{:keys [symbol name icon decimals amount color] :as token} type]
+  [list/touchable-item  #(do (re-frame/dispatch [(type->handler type) symbol])
+                             (re-frame/dispatch [:navigate-back]))
+   [react/view
+    [list/item
+     (if icon
+       [list/item-image icon]
+       [chat-icon/custom-icon-view-list name color])
+     [list/item-content
+      [react/view {:flex-direction :row}
+       [react/text {:style styles/text}
+        name]
+       [react/text {:style {:text-transform :uppercase}}
+        (wallet.utils/display-symbol token)]]
+      [list/item-secondary (wallet.utils/format-amount amount decimals)]]]]])
+
+(views/defview assets [type address]
+  (views/letsubs [assets [:wallet/transferrable-assets-with-amount address]]
+    [simple-screen
+     [topbar (i18n/label :t/wallet-assets)]
+     [react/view {:style (assoc components.styles/flex :background-color :white)}
+      [list/flat-list {:default-separator? true
+                       :data               assets
+                       :key-fn             (comp str :symbol)
+                       :render-fn          #(render-token % type)}]]]))
+
+(views/defview send-assets []
+  (views/letsubs [address [:get-screen-params]]
+    [assets :send address]))
+
+(views/defview request-assets []
+  (views/letsubs [address [:get-screen-params]]
+    [assets :request address]))
+
+(defn- type->view [k]
+  (case k
+    :send    :wallet-send-assets
+    :request :wallet-request-assets
+    (throw (str "Unknown type: " k))))
+
+(views/defview asset-selector [{:keys [disabled? type symbol error address]}]
+  (views/letsubs [balance    [:balance address]
+                  chain      [:ethereum/chain-keyword]
+                  all-tokens [:wallet/all-tokens]]
+    (let [{:keys [name icon decimals color] :as token} (tokens/asset-for all-tokens chain symbol)]
+      (when name
+        [react/view
+         [cartouche {:disabled? disabled? :on-press #(re-frame/dispatch [:navigate-to (type->view type) address])}
+          (i18n/label :t/wallet-asset)
+          [react/view {:style               styles/asset-content-container
+                       :accessibility-label :choose-asset-button}
+           (if icon
+             [list/item-image (assoc icon :style styles/asset-icon :image-style {:width 32 :height 32})]
+             [chat-icon/custom-icon-view-list name color 32])
+           [react/view styles/asset-text-content
+            [react/view styles/asset-label-content
+             [react/text {:style (merge styles/text-content styles/asset-label)}
+              name]
+             [react/text {:style styles/text-secondary-content}
+              (wallet.utils/display-symbol token)]]
+            [react/text {:style (merge styles/text-secondary-content styles/asset-label)}
+             (str (wallet.utils/format-amount (get balance symbol) decimals))]]]]
+         (when error
+           [tooltip/tooltip error {}])]))))
+
+(defn- recipient-address [address modal?]
+  [react/text {:style               (merge styles/recipient-address (when-not address styles/recipient-no-address))
+               :accessibility-label :recipient-address-text}
+   (if (ens/is-valid-eth-name? address)
+     address
+     (or (eip55/address->checksum (ethereum/normalized-address address))
+         (if modal?
+           (i18n/label :t/new-contract)
+           (i18n/label :t/specify-recipient))))])
+
+(views/defview recipient-contact [address name request?]
+  (views/letsubs [contact [:contacts/contact-by-address address]]
+    (let [address? (and (not (nil? address)) (not= address ""))]
+      [react/view styles/recipient-container
+       [react/view styles/recipient-icon
+        (when contact
+          [photos/photo
+           ;;TODO this should be done in a subscription
+           (multiaccounts/displayed-photo contact)
+           {:size list.styles/image-size}])]
+       [react/view {:style styles/recipient-name}
+        [react/text {:style               (styles/participant true)
+                     :accessibility-label (if request? :contact-name-text :recipient-name-text)
+                     :number-of-lines     1}
+         name]
+        [react/text {:style               (styles/participant (and (not name) address?))
+                     :accessibility-label (if request? :contact-address-text :recipient-address-text)}
+         (eip55/address->checksum (ethereum/normalized-address address))]]])))
+
+(defn render-contact [contact request?]
+  [list/touchable-item #(re-frame/dispatch [:wallet/fill-request-from-contact contact request?])
+   [list/item
+    [photos/photo
+     ;;TODO this should be done in a subscription
+     (multiaccounts/displayed-photo contact)
+     {:size list.styles/image-size}]
+    [list/item-content
+     [list/item-primary {:accessibility-label :contact-name-text}
+      (multiaccounts/displayed-name contact)]
+     [react/text {:style list.styles/secondary-text
+                  :accessibility-label :contact-address-text}
+      (eip55/address->checksum (ethereum/normalized-address (:address contact)))]]]])
+
+(defn render-account [account]
+  [list/touchable-item #(re-frame/dispatch [:wallet/fill-request-from-contact account false])
+   [list/item
+    [chat-icon/custom-icon-view-list (:name account) (:color account)]
+    [list/item-content
+     [list/item-primary {:accessibility-label :contact-name-text}
+      (:name account)]
+     [react/text {:style list.styles/secondary-text
+                  :accessibility-label :contact-address-text}
+      (eip55/address->checksum (ethereum/normalized-address (:address account)))]]]])
+
+(views/defview recent-recipients []
+  (views/letsubs [contacts           [:contacts/active]
+                  {:keys [request?]} [:get-screen-params :recent-recipients]]
+    [simple-screen
+     [topbar (i18n/label :t/recipient)]
+     [react/view styles/recent-recipients
+      [list/flat-list {:data      contacts
+                       :key-fn    :address
+                       :render-fn #(render-contact % request?)}]]]))
+
+(views/defview accounts []
+  (views/letsubs [{:keys [accounts]} [:multiaccount]]
+    [simple-screen
+     [topbar (i18n/label :t/accounts)]
+     [react/view styles/recent-recipients
+      [list/flat-list {:data      accounts
+                       :key-fn    :address
+                       :render-fn render-account}]]]))
+
+;;TODO workaround for https://github.com/facebook/react-native/issues/23653 (https://github.com/status-im/status-react/issues/8548)
+(def tw (reagent/atom "95%"))
+
+(views/defview contact-code []
+  (views/letsubs [content (reagent/atom nil)]
+    [simple-screen {:avoid-keyboard? true}
+     [topbar (i18n/label :t/recipient)]
+     [react/view components.styles/flex
+      [cartouche {}
+       (i18n/label :t/recipient)
+       [text-input {:ref                 (fn [v] (js/setTimeout #(reset! tw (if v "100%" "95%")) 100))
+                    :multiline           true
+                    :style               (styles/contact-code-text-input @tw)
+                    :placeholder         (i18n/label :t/recipient-code)
+                    :on-change-text      #(reset! content %)
+                    :accessibility-label :recipient-address-input}]]
+      [toolbar/toolbar {:center {:type      :secondary
+                                 :disabled? (string/blank? @content)
+                                 :on-press  #(re-frame/dispatch [:wallet.send/set-recipient @content])
+                                 :label     :t/done}}]]]))
+
+(defn recipient-qr-code []
+  [choose-recipient/choose-recipient])
+
+(defn- request-camera-permissions []
+  (re-frame/dispatch [:request-permissions {:permissions [:camera]
+                                            :on-allowed  #(re-frame/dispatch [:navigate-to :recipient-qr-code])
+                                            :on-denied   #(utils.utils/set-timeout
+                                                           (fn []
+                                                             (utils.utils/show-popup (i18n/label :t/error)
+                                                                                     (i18n/label :t/camera-access-error)))
+                                                           50)}]))
+
+(defn- on-choose-recipient [contact-only? request?]
+  (list-selection/show {:title   (i18n/label :t/wallet-choose-recipient)
+                        ;;TODO temporary hide for v1
+                        #_(concat
+                           [{:label  (i18n/label :t/recent-recipients)
+                             :action #(re-frame/dispatch [:navigate-to :recent-recipients {:request? request?}])}]
+                           (when-not contact-only?))
+                        :options [{:label  (i18n/label :t/accounts)
+                                   :action #(re-frame/dispatch [:navigate-to :select-account])}
+                                  {:label  (i18n/label :t/scan-qr)
+                                   :action request-camera-permissions}
+                                  {:label  (i18n/label :t/recipient-code)
+                                   :action #(re-frame/dispatch [:navigate-to :contact-code])}]}))
+
+(defn recipient-selector [{:keys [name address disabled? contact-only? request? modal?]}]
+  [cartouche {:on-press  #(on-choose-recipient contact-only? request?)
+              :disabled? disabled?
+              :icon      :main-icons/more
+              :icon-opts {:accessibility-label :choose-contact-button}}
+   (i18n/label :t/wallet-choose-recipient)
+   [react/view {:accessibility-label :choose-recipient-button}
+    (if name
+      [recipient-contact address name request?]
+      [recipient-address address modal?])]])
+
+(defn amount-input [{:keys [input-options amount amount-text disabled?]}
+                    {:keys [symbol decimals]}]
+  [react/view {:style               components.styles/flex
+               :accessibility-label :specify-amount-button}
+   [text-input
+    (merge
+     input-options
+     ;; We only auto-correct and prettify user's input when it is valid and positive.
+     ;; Otherwise, user might want to fix his input and autocorrection will give more harm than good.
+     ;; Positive check is because we don't want to replace unfinished 0.000 with just plain 0, that is annoying and
+     ;; potentially dangerous on this screen (e.g. sending 7 ETH instead of 0.0007)
+     {:default-value  (if (empty? amount-text)
+                        (str (money/to-fixed (money/internal->formatted amount symbol decimals)))
+                        amount-text)}
+     (if disabled?
+       {:editable    false
+        :placeholder ""}
+       {:keyboard-type       :numeric
+        :placeholder         (i18n/label :t/amount-placeholder)
+        :style               components.styles/flex
+        :accessibility-label :amount-input}))]])
+
+(defn amount-selector [{:keys [error disabled?] :as m} token]
+  [react/view
+   [cartouche {:disabled? disabled?}
+    (i18n/label :t/amount)
+    [amount-input m token]]
+   (when error
+     [tooltip/tooltip error])])
 
 (defn separator []
   [react/view styles/separator])
--- a/src/status_im/utils/universal_links/core.cljs
+++ b/src/status_im/utils/universal_links/core.cljs
@@ -6,7 +6,6 @@
             [status-im.multiaccounts.model :as multiaccounts.model]
             [status-im.chat.models :as chat]
             [status-im.constants :as constants]
-            [status-im.ethereum.eip681 :as eip681]
             [status-im.pairing.core :as pairing]
             [status-im.utils.security :as security]
             [status-im.ui.components.list-selection :as list-selection]
@@ -86,8 +85,10 @@
       (navigation/navigate-to-cofx (assoc-in cofx [:db :contacts/identity] public-key) :profile nil))))
 
 (fx/defn handle-eip681 [cofx url]
-  (choose-recipient/resolve-ens-addresses url :deep-link)
-  {:dispatch [:navigate-to :wallet-send-transaction]})
+  (fx/merge
+   cofx
+   (choose-recipient/resolve-ens-addresses url :deep-link)
+   (navigation/navigate-to-cofx :wallet-send-transaction nil)))
 
 (defn handle-not-found [full-url]
   (log/info "universal-links: no handler for " full-url))
--- a/src/status_im/utils/universal_links/core.cljs
+++ b/src/status_im/utils/universal_links/core.cljs
@@ -1,5 +1,6 @@
 (ns status-im.utils.universal-links.core
   (:require [cljs.spec.alpha :as spec]
+            [clojure.string :as string]
             [goog.string :as gstring]
             [re-frame.core :as re-frame]
             [status-im.multiaccounts.model :as multiaccounts.model]
@@ -16,7 +17,8 @@
             [status-im.utils.config :as config]
             [status-im.utils.fx :as fx]
             [status-im.utils.platform :as platform]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [status-im.wallet.choose-recipient.core :as choose-recipient]))
 
 ;; TODO(yenda) investigate why `handle-universal-link` event is
 ;; dispatched 7 times for the same link
@@ -42,6 +44,9 @@
   (some->> url
            (re-matches regex)
            peek))
+
+(defn is-request-url? [url]
+  (string/starts-with? url "ethereum:"))
 
 (defn universal-link? [url]
   (boolean
@@ -81,8 +86,8 @@
       (navigation/navigate-to-cofx (assoc-in cofx [:db :contacts/identity] public-key) :profile nil))))
 
 (fx/defn handle-eip681 [cofx url]
-  {:dispatch-n [[:navigate-to :wallet]
-                [:wallet/fill-request-from-url url]]})
+  (choose-recipient/resolve-ens-addresses url :deep-link)
+  {:dispatch [:navigate-to :wallet-send-transaction]})
 
 (defn handle-not-found [full-url]
   (log/info "universal-links: no handler for " full-url))
@@ -107,7 +112,7 @@
     (match-url url browse-regex)
     (handle-browse cofx (match-url url browse-regex))
 
-    (some? (eip681/parse-uri url))
+    (is-request-url? url)
     (handle-eip681 cofx url)
 
     :else (handle-not-found url)))
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -174,7 +174,7 @@
                   [[:address] [:function-arguments :address]])]
       (if (empty? ens-names)
       ;; if there is no ens-names, we dispatch request-uri-parsed immediately
-        (request-uri-parsed message origin)
+        (request-uri-parsed cofx message origin)
         {::resolve-addresses
          {:registry (get ens/ens-registries (ethereum/chain-keyword db))
           :ens-names ens-names
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -157,7 +157,7 @@
 
 (fx/defn resolve-ens-addresses
   {:events [:wallet.send/resolve-ens-addresses :wallet.send/qr-code-request-scanned]}
-  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db}
+  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db :as cofx}
    uri origin]
   (if-let [message (eip681/parse-uri uri)]
   ;; first we get a vector of ens-names to resolve and a vector of paths of
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -49,11 +49,10 @@
 (defn- extract-details
   "First try to parse as EIP681 URI, if not assume this is an address directly.
    Returns a map containing at least the `address` and `chain-id` keys"
-  [s chain-id all-tokens]
-  (or (let [m (eip681/parse-uri s)]
-        (merge m (eip681/extract-request-details m all-tokens)))
-      (when (ethereum/address? s)
-        {:address s :chain-id chain-id})))
+  [m chain-id all-tokens]
+  (or (merge m (eip681/extract-request-details m all-tokens))
+      (when (ethereum/address? m)
+        {:address m :chain-id chain-id})))
 
 ;; NOTE(janherich) - whenever changing assets, we want to clear the previusly set amount/amount-text
 (defn changed-asset [{:keys [db] :as fx} old-symbol new-symbol]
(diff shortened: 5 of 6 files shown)


acolytec3:
@3esmit I believe this solves the basic question of validating ENS addresses within an EIP681 URI, per this specific issue as I have seen events in re-frisk where the ENS address is correctly parsed and set in the wallet/send-transaction/to field in the DB.  That said, the only place I'm aware that these URIs are accepted in the app is in the send-transaction flow where you scan a QR code and that flow doesn't handle ENS addresses at present and will throw an unhandled error if you scan a URI with an ENS address embedded in it since it tries to pass it to `eip55/address->checksum`.  I think I need to add ENS name resolution [here](https://github.com/status-im/status-react/blob/cbe17cbe020189f5d7aaacba2f82da63061e47ac/src/status_im/wallet/choose_recipient/core.cljs#L32) in order to make it work so will try to add that to this PR as well.

3esmit (approved):
Seems good to me.

3esmit:
Thanks for this. This is one step closer to full EIP681 support :)
Regarding the Name Resolution on the UI, I think that it should be another issue/PR because this would be relative to only allowing the ENS names to be used in EIP681.

If the UI is not resolving ENS Names in any `address` field, then we should investigate that separately.

3esmit:
You might also remove the comment "TODO ens support" :)

3esmit:
Ah, I see, you say about "generating ERC681" using ENS. Yeah, that is totally a new feature, it changes the UI because users might want to enable or disable using ENS name in the payment link. 

We certainly would need discussion on that, so I will open another issue "Allow users generate EIP681 with ENS name".

acolytec3:
> Thanks for this. This is one step closer to full EIP681 support :)
> Regarding the Name Resolution on the UI, I think that it should be another issue/PR because this would be relative to only allowing the ENS names to be used in EIP681.
> 
> If the UI is not resolving ENS Names in any `address` field, then we should investigate that separately.

Sorry, I think I misstated this.  ENS name resolution works fine if you enter an address manually in the send flow.  It's only when you scan it from a EIP681 URI in the QR flow that it's failing, due to the above noted explanation.  It's happening because that flow doesn't expect ENS addresses today so assumes any time the QR read succeeds that it has a valid Ethereum address and then tries to run it through the checksum validation method before populating in the to field in the send-transaction screen and that's where the error occurs.  To your above point, I guess that could probably be addressed as a separate issue/PR since it deals with a specific UI flow not resolving ENS addresses.

acolytec3:
> You might also remove the comment "TODO ens support" :)

Done!

3esmit:
Also, seems like that it should only support ENS if prefix `pay-` is used, because of EIP831.  See discussion https://gitter.im/ethereum/EIPs?at=5da9b857ea0d6159fd877aa6

3esmit (changes requested):
To not conflicting with [EIP831](https://eips.ethereum.org/EIPS/eip-831):
>`prefix` is optional and defines the use-case for this URI. If no prefix is given: “pay-“ is assumed to be concise and ensure backward compatibility to ERC-67. When the prefix is omitted, the payload must start with 0x. Also prefixes must not start with 0x. So starting with 0x can be used as a clear signal that there is no prefix.

So this would be the valid semantic:

`ethereum:` + (ETH_ADDRESS | (`pay-` + (ETH_ADDRESS | ENS_NAME)) + payload.

acolytec3:
Got it. I'll take a look at the regex and see if I can make that work.

acolytec3:
@3esmit I've added in additional validation for the `pay-` prefix as requested.  This uses string comparisons/substring methods to parse the address since that seemed more straightforward for parsing this individual field. Does this feel right?

3esmit (approved):

3esmit:
Looks good. Glad you didn't had to use regex for that. 

I checked the logic and it is according to EIP681: checks if address, if not check if there is prefix `pay`, if there is, checks if is ens name valid or address valid.

@yenda can you validate if the PR and merge it? Thanks!

acolytec3:
Cool, just remember we need another issue to deal with the QR code methods
not resolving ens addresses.

churik:
@acolytec3 please, rebase it to current develop and I try to rebuild it

3esmit:
@churik This only fixes the error message saying that the link is invalid, while it was valid. 
There is still another issue that is not fixed in this PR which is the UI is not looking up the ENS name when it came from EIP681/qrcode scanner.

churik:
@acolytec3 thank you for contribution!

Error on scanning address from initial issue (Android 8):
`ethereum:pay-snt.thetoken.eth/transfer?address=unicorn.stateofus.eth&uint256=1e16`
photo_2019-10-30 15 52 29

Logcat: 
[logcat_EIP.log](https://github.com/status-im/status-react/files/3789560/logcat_EIP.log)

If error about invalid address is shown, it should be handled gracefully.

acolytec3:
@churik That's the error that needs to be solved by #9357.  It's generated somewhere in the flow of populating the address that starts with `fill-details-from-url` function in `wallet.choose_recipient.core` but I haven't been able to track it down so far.

churik:
@3esmit @acolytec3 on nightly currently at attempt to scan this code I see "Invalid address error", can close popup and continue working.
In this PR after scanning this error I can't continue working with application (it crashes and I should reopen it), so don't think that merging this without fixing this error is a good idea.

acolytec3:
@churik  @3esmit  I think I finally figured out where it's breaking.  Will work on a fix and push a new commit in the next day or two.

yenda (changes requested):
Thanks for your PR, could you please add tests for the new cases and also make sure the parse-uri function doesn't have lines longer than 80 chars

acolytec3:
@yenda @3esmit  Don't mind the commit I just pushed.  I'm just saving some work to github around this and trying to rebuild my dev environment.  The app stopped building in my local dev environment yesterday and adb is throwing some weird logs so trying a completely fresh rebuild of my entire status repo locally.

acolytec3:
@yenda  @3esmit  I've added additional commits to fix the error that @churik described where the ENS address is correctly parsed but fails to resolve.  The QR scanner will now correctly resolve an ENS address and populate the address field in the DB.  The to-do left  add the tests @yenda requested.  
I will note that the error message generated when invalidly constructed ENS address is unresolvable is pretty generic and doesn't clearly communicate what the issue is.  

Screenshot_20191102-185146

src/status_im/ui/screens/wallet/components/views.cljs
```diff
@@ -162,10 +163,12 @@
 (defn- recipient-address [address modal?]
   [react/text {:style               (merge styles/recipient-address (when-not address styles/recipient-no-address))
                :accessibility-label :recipient-address-text}
-   (or (eip55/address->checksum (ethereum/normalized-address address))
-       (if modal?
-         (i18n/label :t/new-contract)
-         (i18n/label :t/specify-recipient)))])
+   (if (ens/is-valid-eth-name? address)
```
3esmit, line 151: seems like that this fixes #9357

3esmit:
There is an error in linting preventing the CI to build, please fix it so I can test it. See https://ci.status.im/job/status-react/job/prs/job/android/job/PR-9240/9/display/redirect

src/status_im/ui/screens/wallet/components/views.cljs
acolytec3: The overall PR does now fix #9357 as well though I'm not 100% sure if this particular line of code is useful or not.  I had originally added this line to make sure it didn't try to run an ENS name through the EIP55 checksum process since that's what generated the error @churik identified (and I saw as well).  Not sure if it's still useful or not since we're taking care of name resolution in the `fill-request-from-url` method that populates the send transaction now.

acolytec3:
> There is an error in linting preventing the CI to build, please fix it so I can test it. See https://ci.status.im/job/status-react/job/prs/job/android/job/PR-9240/9/display/redirect

This is fixed in my latest commit.  I also added some new tests to address ENS name validation in the EIP681 tests.  Can you review and make sure I didn't miss any scenarios?

acolytec3:
@3esmit I think I need to make another adjustment to the logic in `parse-uri` too since I didn't do any ENS validation on the second address (in case of an ERC20 token transfer) so I will add that as well.

yenda:
@acolytec3 great job so far! are you planning to finish this soon? I'd like to use it for the implementation of send and receive command

acolytec3:
@yenda I've been trying to finish out the piece of resolving the second ENS name if the transaction involves a token transfer but it's proving tricky because of the way the name resolution occurs. The `eip681/extract-details` function expects the token address to be resolved already so been working through how to incorporate that into the overall flow in a way that makes sense. I've got a commit that's make 50% done but still need some more time to work out some bugs in it.

acolytec3:
@3esmit @yenda So...I've made progress.  The PR as it stands works for simple ETH payments supporting ENS  For ERC20 transfers, it mostly works except I'm having to hardcode the EIP681 URI in one point.  I think this isn't related to the core PR as much as my ineptitude with re-frame so will try and figure that piece out next.

qr-resolve-ERC20-ens

src/status_im/ethereum/eip681.cljs
```diff
@@ -41,21 +42,31 @@
...
+        (let [[_ raw-address chain-id function-name] (re-find authority-path-pattern authority-path)]
+          (when (or (ethereum/address? raw-address)
+                    (if (string/starts-with? raw-address "pay-")
+                      (let [pay-address (string/replace-first raw-address "pay-" "")]
+                        (or (ens/is-valid-eth-name? pay-address)
+                            (ethereum/address? pay-address)))))
+            (let [address (if (string/starts-with? raw-address "pay-")
+                            (string/replace-first raw-address "pay-" "")
+                            raw-address)]
+              (when-let [arguments (parse-arguments function-name query)]
```
3esmit, line 65: Seems like `parse-arguments` are still not validating the parameters, I guess this will be left to whatever calls it?

There is needs for input sanitization in this case? if ,yes, they should be done here or by calling method?

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -72,6 +73,13 @@
  (fn [{:keys [registry ens-name cb]}]
    (ens/get-addr registry ens-name cb)))
 
+(fx/defn resolve-token-address-from-qr
+  {:events [:wallet.send/resolve-token-address-from-qr]}
+  [{:keys [db]} address]
+  (let [raw-data "ethereum:pay-snt.thetoken.eth/transfer?address=acolytec3.stateofus.eth&uint256=1e2"
```
3esmit: Is this in the main code? please undo this.

3esmit (changes requested):
There are other bugs preventing me from testing this, but by reading the changes I noticed that you probably submitted by mistake the hardcoded link in the commit, besides that seems fine.

src/status_im/ethereum/eip681.cljs
3esmit: This is not a request to change, just opening discussion. @yenda, wdyt?

yenda: What kind of validation would you expect? is it just `:value` `:gas` and `:gasPrice`?

src/status_im/wallet/choose_recipient/core.cljs
yenda: is it the place you said you had to hardcode @acolytec3 ?

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -93,29 +101,36 @@
...
-      (and address valid-network?) (update :db #(fill-request-details % details false))
-      symbol-changed? (changed-asset old-symbol new-symbol)
-      (and old-amount new-amount (not= old-amount new-amount)) (changed-amount-warning old-amount new-amount)
+  (assoc-in db [:wallet :eip681] nil)
+  (if (and (string/includes? data "transfer")
+           (< (string/index-of data ".eth") (string/index-of data "transfer")))
+    {:resolve-address
+     {:registry  (get ens/ens-registries (ethereum/chain-id->chain-keyword (get-in constants/default-networks [current-network :config :NetworkId])))
+      :ens-name  (get-in (eip681/parse-uri data) [:address])
+      :cb        #(re-frame/dispatch [:wallet.send/resolve-token-address-from-qr %])}}
```
yenda: you just need to pass the data here like this:

```suggestion
      :cb        #(re-frame/dispatch [:wallet.send/resolve-token-address-from-qr data %])}}
```

yenda (changes requested):
So I think you should change your approach here, the `parse-uri` function should be an asynchronous function that returns the result in a callback, so that you can resolve the ens names there.

This has two implications:
- you will only call fill-request-from-url through an event dispatch once you have been able to extract and resolve all the data. That means you dispatch immediately after parsing if there is no ens names, or in the callback of the ens resolution otherwise. I would suggest for simplicity sake that in case of token transactions where both the contract and the recipient are refered to by ens address, you resolve the contract, then in the callback your resolve the recipient address.
@3esmit is there cases where we would have more than 2 ens names to resolve?

- you will need a `is-request-uri?` or something function for `status-im.utils.universal-links.core`, which will be actually better because right now it parses the uri twice, once in `route-url` and then in `fill-request-url` again.

Also you could rename the event of `fill-request-from-url` to the new convention with something like `::request-uri-parsed`. The `::` is a shorthand for the current namespace, so that means the full event name will be `:status-im.wallet.choose-recipient.core/request-uri-parsed`, and you can use refer to it in other namespaces using the alias `::wallet.choose-recipient/request-uri-parsed` (if you required the namespace with this alias like this: `[status-im.wallet.choose-recipient.core :as choose-recipient.core]`)

If you have any question or need help with clojurescript/re-frame mention me and I'll try to help ASAP.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: Yep, was having trouble passing the complete URL string through all the functions and then the callback so had just hardcoded it to see if the general proof of concept worked.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: This fixed it.  I had been trying to pass data after the anonymous parameter and that's what kept messing me up.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: Fixed this one based on the below feedback from @yenda.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: I went ahead and pushed a new commit that resolves the hardcoding issue I was having.  Thanks for pointing out that easy remedy to my problem!

acolytec3:
@yenda To make sure I'm following, you're suggesting do all the parsing of the EIP681 url in the `eip681/parse-uri`, including any function parameters?  That would probably simplify the PR a lot since I've basically had to add an extra layer of parsing in `fill-request-from-url` to check for the contract address since `parse-uri` doesn't currently look for the second address.  Admittedly not the best solution in hindsight, but just a first pass to get everything going.

A reworked solution would look something like:
* Scan QR code and pass resulting URL to a new `is-request-url?` in `universal-links`.
* `is-request-url?` passes the url to a slightly reworked `eip681/parse-uri` function that parses out and validates all the address, parameters, and values in the URI
* `is-request-url?` would resolve each address (if one or more is an ENS address) via a looping callback process (similar to what `set-recipient` or my `resolve-token-address-from-url` 
* Once all ENS addresses are resolved, pass sanitized corresponding data elements to `fill-request-from-url`?

If we go this route, I'm assuming we'd want to also rework `fill-request-from-url` to take all the elements as parameters rather than reparsing the way it does currently, right?

acolytec3:
And, just for fun, at least it works now as currently coded.  :-)

qr-resolve-ERC20-ens

yenda:
@acolytec3 almost, `is-request-uri?` should just to the minimum to verify that the uri is a eip681, so I don't think it should include resolving, especially since resolving would be asyncronous and the way it is used in the universal-link namespace it would make things more complex for no good reason.

Check this:

in `route-url` you have this condition:
```
(some? (eip681/parse-uri url))
    (handle-eip681 cofx url)
```

instead you will have
```
(is-request-uri? url)
(handle-eip681 cofx url)
```

`handle-eip681` does this:

```
(fx/defn handle-eip681 [cofx url]
  {:dispatch-n [[:navigate-to :wallet-send-transaction]
                [:wallet/fill-request-from-url url :deep-link]]})
```
so here instead of dispatching `fill-request-from-url` you could call your modified `parse-uri` directly that dispatches `fill-request-from-url` (renamed as I detailed above) as a callback once parsing and resolving is done

yenda:
@acolytec3 

> And, just for fun, at least it works now as currently coded. :-)

Nice good job, why isn't it picking up the right asset and amount though? seems like you requested 100 unicorn tokens but it shows up with invalid amount of eth

acolytec3:
> @acolytec3
> 
> > And, just for fun, at least it works now as currently coded. :-)
> 
> Nice good job, why isn't it picking up the right asset and amount though? seems like you requested 100 unicorn tokens but it shows up with invalid amount of eth

@yenda, there are still some other bugs in it and that signing sheet shouldn't even be showing. Something is going awry in `changed-asset` related to the `update-gas-price` call.  It works some times and not others.  Once I have time to rework the whole process along the lines outlined above, I'll take a look at this part too.

acolytec3:
> So I think you should change your approach here, the `parse-uri` function should be an asynchronous function that returns the result in a callback, so that you can resolve the ens names there.
> 
> This has two implications:
> 
> * you will only call fill-request-from-url through an event dispatch once you have been able to extract and resolve all the data. That means you dispatch immediately after parsing if there is no ens names, or in the callback of the ens resolution otherwise. I would suggest for simplicity sake that in case of token transactions where both the contract and the recipient are refered to by ens address, you resolve the contract, then in the callback your resolve the recipient address.
>   @3esmit is there cases where we would have more than 2 ens names to resolve?
> * you will need a `is-request-uri?` or something function for `status-im.utils.universal-links.core`, which will be actually better because right now it parses the uri twice, once in `route-url` and then in `fill-request-url` again.
> 
> Also you could rename the event of `fill-request-from-url` to the new convention with something like `::request-uri-parsed`. The `::` is a shorthand for the current namespace, so that means the full event name will be `:status-im.wallet.choose-recipient.core/request-uri-parsed`, and you can use refer to it in other namespaces using the alias `::wallet.choose-recipient/request-uri-parsed` (if you required the namespace with this alias like this: `[status-im.wallet.choose-recipient.core :as choose-recipient.core]`)
> 
> If you have any question or need help with clojurescript/re-frame mention me and I'll try to help ASAP.

@yenda I think I follow this for the most part so will start re-working things.  One question for @3esmit: at one point you mentioned that you wanted the `eip681` module to only validate the addresses conformed to ENS or Ethereum standards and wanted ENS resolution to be handled by the middleware/UI layers.  Is that still accurate? Now that I have a better sense of things, I think I can pretty easily re-work `fill-request-from-url` to resolve any ENS addresses with minimal effort (similarly to what I currently have designed but with a little cleaning up).

src/status_im/ethereum/eip681.cljs
3esmit: validation of parameters and values, for example, if is uint256, than it should be an valid uint256 format, if is address, it should be an valid address format. 
e.g. 1. `ethereum:pay-a.eth/transfer&uimt256=1e10&address=b.eth` (`uimt256` is invalid); 
2. `ethereum:pay-a.eth/transfer&uint256=1e10&address=Null` (`Null` is invalid).

acolytec3: I think at least 1-2 should be pretty straight forward. I'm going to update `parse-uri` in the morning and I think it should be pretty straight forward to at least do validation on all the standard required/optional fields. Catching extra parameters might be trickier. Are you only concerned about that with the transfer function?

3esmit: I see that this is out of scope for this PR, we might open an issue about it after this was merged, as this issue and pr is just about the ens support, not validation of fields on EIP681.

3esmit:
ENS address might be needed to resolve for the target_address and also for parameter value, this is part of EIP681, it explicitly mentions that a value can be ethereum_address, which can be an ENS_NAME, or starts with 0x +40 hex digits.

from https://eips.ethereum.org/EIPS/eip-681

    request                 = erc831_part target_address [ "@" chain_id ] [ "/" function_name ] [ "?" parameters ]
    erc831_part             = schema and optional prefix as defined in #831 - typically "ethereum" ":" [ "pay-" ] in this case
    target_address          = ethereum_address
    chain_id                = 1*DIGIT
    function_name           = STRING
    ethereum_address        = ( "0x" 40*40HEXDIG ) / ENS_NAME
    parameters              = parameter *( "&" parameter )
    parameter               = key "=" value
    key                     = "value" / "gas" / "gasLimit" / "gasPrice" / TYPE
    value                   = number / ethereum_address / STRING
    number                  = [ "-" / "+" ] *DIGIT [ "." 1*DIGIT ] [ ( "e" / "E" ) [ 1*DIGIT ] [ "+" UNIT ]

3esmit (approved):

acolytec3:
@3esmit @yenda  I've added ENS/ethereum address validation (and tests) for a second address where it exists within the function arguments in `eip681/parse-uri` so that should take care of that piece.  Now, going to start on reworking the UI piece to get `fill-request-from-url` properly working again along the lines suggested above.

acolytec3:
@jakubgs @PombeirP @corpetty Disregard above ping from the bot.  Rebased on develop and forgot to push --force so it accidentally grabbed all the develop commits.  Corrected now.

acolytec3:
@3esmit @yenda I think I have basically everything re-worked at this point.  Name resolution is working properly for both the main "to" address as well as any secondary address that's part of the function arguments so ERC20 token transfers now render correctly.  The one thing that's not working and I haven't been able to track down is the value is being rendered as wei when doing an ERC20 transfer.  I've traced the value through the flow and it's getting correctly converted to a `bigNumber` in the `eip681/extract-request-details` and then passed through `request-uri-parsed` to `fill-request-details` but the value still renders like the value is in ETH.  For example, scanning the qr code for this url ethereum:pay-snt.thetoken.eth/transfer?address=gitcoin.eth&uint256=1e2, renders this transaction.
Screenshot_20191111-100848

@yenda, could you review the PR and see if you see anything I'm missing here?  I didn't change anything about the flow of `extract-details` and `fill-request-details` so stumped as to why it's still doing this.  

@3esmit With this code base, I'm not able to recreate #9380 or #9381 so assuming they've both been addressed as well.

3esmit:
@acolytec3  Its working now indeed, but not from the chat tab screen, when using the `+` plus button -> `Scan QR code`.

acolytec3:
Got it. I think I know what the issue is.

src/status_im/ethereum/eip681.cljs
```diff
@@ -41,21 +43,41 @@
...
+                    (merge {:address address
+                            :chain-id (if chain-id
+                                        (js/parseInt chain-id)
+                                        (ethereum/chain-keyword->chain-id :mainnet))}
+                           arguments)))))))))))
+
+(defn parse-uri-handler
+  [s origin]
+  (let [m (parse-uri s)]
+    (if-not (= m nil)
```
yenda: what does m mean here?

also I think you can replace this whole thing by:

```
(when-let [m (parse-uri s)]
  (re-frame/dispatch [:wallet.send/resolve-ens-addresses m origin nil])
```

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -72,6 +69,30 @@
  (fn [{:keys [registry ens-name cb]}]
    (ens/get-addr registry ens-name cb)))
 
+(fx/defn resolve-ens-addresses
+  {:events [:wallet.send/resolve-ens-addresses]}
+  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db} message origin address]
+  (if address
```
yenda: this function needs some clarity, as a first step you could put the following into a separate function:
```
{:resolve-address
         {:registry  (get ens/ens-registries (ethereum/chain-id->chain-keyword (get-in constants/default-networks [current-network :config :NetworkId])))
          :ens-name  (get-in message [:function-arguments :address])
          :cb        #(re-frame/dispatch [:wallet.send/resolve-ens-addresses (merge message {:address address}) origin %])}}
```

yenda: Also not sure how well you know javascript, but I think here it might be good to use promises. It would work like this:
- you make an effect called :resolve-addresses
```clojure
(fn [registry ens-names cb]
;; TODO this is incomplete, for instance you want more than just a list of ens names, probably also their path
         (-> (js/Promise.all 
                   (clj->js (mapv (fn [[path ens-name]] 
                                               (js/Promise. (fn [resolve reject] 
                                                                       (ens/get-addr registry ens-name #(resolve [path %]))) ens-names)
               (.then (fn [results] 
 ;; TODO here we put the results back into the message
                   (cb message)))
```
This snippet is completely wrong just "pseudo-code" to show the idea, if should take probably a vector of tuples, because you want to know the path of each address to replace them in the message, and you returned the resolved message in the then by calling directly  `request-uri-parsed`, avoiding calling resolve-ens-addresses` recursively.
- could also work with reject in case the name is not resolved

wdyt?

acolytec3: It makes sense and I have a basic working knowledge of promises on the JavaScript side. I'd thought about that in an earlier iteration of this but just didn't have enough understanding of how to go back and forth between cljs and js to make it work. I'll see if I can rework `resolve-ens-names` per this flow.

src/status_im/ethereum/eip681.cljs
acolytec3: This was just an abstraction over `parse-uri` so the current tests could continue to be used.  I'll clean up according to the suggestion.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: @yenda I think this is down to my lack of comprehension on how to properly handle calling javascript functions within clojurescript but I'm not able to make something along the lines of what you suggest work.  Below is what I have so far.  Figwheel compiles it without any complaints but I get possible unhandled promise rejection/syntax errors in the ADB logs when trying to run the app with this code in place.  Is it something simple that I'm missing?  Note, this is definitely still complete as it assumes you're getting both 2 addresses in ENS form and doesn't account for parsing path to validate where the ENS addresses are actually found within the URI..  
```(re-frame/reg-fx
 :resolve-addresses
 (fn [registry ens-names message origin]
 (-> js/Promise.all
     (clj->js (mapv (fn [[path ens-name]]
                      js/Promise. (fn [resolve reject]
                                    ens/get-addr registry ens-name #(resolve [path %]))) ens-names)
              .then (fn [[results]]
                      (let [contract-address (get-in results [:address])
                            to-address (get-in results [:function-arguments :address])]
                        (cond-> true
                          true (js/console.log results)
                          contract-address (update-in message [:address] contract-address)
                          to-address (update-in message [:function-arguments :address] to-address)
                          true (re-frame/dispatch [:wallet/request-uri-parsed message origin]))))
              .catch (fn [[error]]
                       (js/console.log error))
              ))))
```

src/status_im/ethereum/eip681.cljs
acolytec3: Fixed with [16164e](https://github.com/status-im/status-react/pull/9240/commits/16164e80ae3fc7aee8a0ca7d634ab65bb9a98277)

src/status_im/wallet/choose_recipient/core.cljs
yenda: hey kudos for keeping up with the good work, don't hesitate to ask questions earlier without spending too much time scratching your head. For starter you have a few missing/missplaced parens:

```clojure
;; this is a threading maco `->`, it means the result of each expression is passed to the next
;; so in this case I want to pass the result of Promise.all to .then and .catch to register my callbacks
(-> (js/Promise.all (clj->js (mapv (fn [[path ens-name]]
                                     (js/Promise.
                                      (fn [resolve reject]
                                        (ens/get-addr registry ens-name #(resolve [path %]))))) ens-names)))
    (.then (fn [[results]]
             (let [contract-address (get-in results [:address])
                   to-address (get-in results [:function-arguments :address])]
;; here you are using another kind of threading macro, a conditional one. What you are doing is
;; saying "here is my expression, at each stage if my condition on the left is true then pass the value of the previous expression as first argument to the expression on the right
               (cond-> true
                 true (js/console.log results) ;;here you are logging `true` and `results` and probably returning nil
                 contract-address (update-in message [:address] contract-address) ;; here you will have an arity error because you are passing the value of the previous expression, nil, as first argument
                 to-address (update-in message [:function-arguments :address] to-address)
                 true (re-frame/dispatch [:wallet/request-uri-parsed message origin])))))
    (.catch (fn [[error]]
              (js/console.log error))))
```
instead of cond you probably want to use a `do`
note that message is an immutable value, `update-in` doesn't change message, it just returns an updated version of message, what you might want to do is return it or do more things with it, but if you just do it in the middle of effectful logging statements, it is not going to have any effect.

yenda: threading:
```clojure
(-> message ;; my first expression is just the symbol message, provided I have defined it or provided it as an argument to the function that runs this code
      (update-in [:address] contract-address) ;; the macro will insert message here
      ;; if the function only takes one argument I can omit parens
      send ;; imagine send being a side-effect function takes a message as first argument
```

acolytec3: @yenda I'm still struggling to make this work.  I've moved parens around and fixed all the bugs I can find but it's still throwing an "undefined is not a function" error when I scan a qr code.
I've traced this code through and verified that it is passing the ens address to the promise and the `get-addr` function within each promise is resolving the address but as far as I can tell, the promises are never actually resolving since the `.then` piece never fires.  Do you see any more misplaced parens?
```
(re-frame/reg-fx
 :resolve-addresses
 (fn [{:keys [registry ens-names message origin]}]
   (-> js/Promise.all
       (clj->js (mapv (fn [[path ens-name]]
                        (js/Promise.
                         (fn [resolve reject]
                           (ens/get-addr registry ens-name #(resolve [path %]))))) ens-names))
       (.then (fn [results]
                (let [contract-address (get-in results [:address])
                      to-address (get-in results [:function-arguments :address])]
                  (-> message
                      contract-address (update-in [:address] contract-address)
                      to-address (update-in  [:function-arguments :address] to-address)
                      (re-frame/dispatch [:wallet/request-uri-parsed origin])))))
       (.catch (fn [error]
                 (js/console.log error))))))
```

acolytec3: @yenda Any thoughts on this latest iteration?  I'd love to get this PR finished up but this approach using interop promises is proving quite challenging to make work.  It seems like the `resolve [path %]` is the likely source of the error about calling an undefined function that I'm seeing but I confess I have no idea how to fix it.  It just seems like the get-addr function isn't calling that resolve callback once it gets the ENS name back from the resolver contract.

yenda: @acolytec3 sorry I missed your previous message, you forgot the parenthesis around (js/Promise.all.. and its argument

yenda: also could you push your changes I wanted to help you but seems like the last commit doesn't include these changes

acolytec3:
I pushed my latest commit incorporating your suggested parens. I haven't
had a chance to test it yet but will do so tonight or tomorrow morning.

yenda:
@acolytec3 I've worked a bit on it in this branch https://github.com/status-im/status-react/tree/acolytek3/promises please have a look, I think you can just cherry-pick commit , I tested it for the QR code in the issue here https://github.com/status-im/status-react/issues/9238 and it seems to work 

I noticed that when I scan with a qr code, `origin` is `deep-link`, I suppose that's a bug?

acolytec3:
> @acolytec3 I've worked a bit on it in this branch https://github.com/status-im/status-react/tree/acolytek3/promises please have a look, I think you can just cherry-pick commit , I tested it for the QR code in the issue here #9238 and it seems to work
> 
> I noticed that when I scan with a qr code, `origin` is `deep-link`, I suppose that's a bug?

I'm sure.  I think I hard-coded "deep-link" somewhere incorrectly.  It should be `:qr` when coming from a QR code and `:deep-link` only when you tap on already formed EIP681 URL in a chat or somewhere similar.
I'll find and correct.

acolytec3:
Thanks to @yenda, I think this PR might be finally ready for review again.  I've tested all the possible ENS combos and they are resolving correctly and translating to ethereum addresses in the `Send-Transaction` view.

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -68,18 +65,66 @@
...
 (re-frame/reg-fx
- :resolve-address
+ ::resolve-address
  (fn [{:keys [registry ens-name cb]}]
    (ens/get-addr registry ens-name cb)))
 
+(re-frame/reg-fx
+ ::resolve-addresses
+ (fn [{:keys [registry ens-names callback]}]
+   ;; resolve all addresses then callt
```
yenda:
```suggestion
   ;; resolve all addresses then call the callback function with the array of addresses as parameter
```

acolytec3: I'll have to fix it tonight.  On mobile right now so wouldn't be able to rebase/squash if I accept the commit now.  That said, if it's ready to be merged sooner, feel free to go ahead and do so.

acolytec3: Fixed in latest commit.

3esmit (approved):
ENS names works in QR Code scanner. @ latest commit (`920a855`).

yenda (approved):

src/status_im/utils/universal_links/core.cljs
```diff
@@ -81,8 +85,8 @@
       (navigation/navigate-to-cofx (assoc-in cofx [:db :contacts/identity] public-key) :profile nil))))
 
 (fx/defn handle-eip681 [cofx url]
-  {:dispatch-n [[:navigate-to :wallet-send-transaction]
-                [:wallet/fill-request-from-url url :deep-link]]})
+  (eip681/parse-uri-handler url :deep-link)
+  {:dispatch [:navigate-to :wallet-send-transaction]})
```
flexsurfer: im not sure i understand this one

flexsurfer (changes requested):

src/status_im/utils/universal_links/core.cljs
flexsurfer: if should look like this
```clojure
(fx/merge cofx
          #(when-let [parsed-message (parse-uri url)]
               (resolve-ens-addresses % parsed-message :deep-link nil])
          (navigation/navigate-to-cofx :wallet-send-transaction nil)
```

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -68,18 +65,67 @@
...
+                    (if (ens/is-valid-eth-name? address)
+                      (-> acc
+                          (update :paths conj path)
+                          (update :ens-names conj address))
+                      acc)))
+                {:paths [] :ens-names []}
+                [[:address] [:function-arguments :address]])]
+    (if (empty? ens-names)
+      ;; if there is no ens-names, we dispatch request-uri-parsed immediately
+      (re-frame/dispatch [:wallet/request-uri-parsed message origin])
```
flexsurfer: we shouldn't use `dispatch` here, call of `(request-uri-parsed cofx ...` should be here

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -68,18 +65,67 @@
...
+ ::resolve-address
  (fn [{:keys [registry ens-name cb]}]
    (ens/get-addr registry ens-name cb)))
 
+(re-frame/reg-fx
+ ::resolve-addresses
+ (fn [{:keys [registry ens-names callback]}]
+   ;; resolve all addresses then call the callback function with the array of 
+   ;;addresses as parameter
+   (-> (js/Promise.all
```
flexsurfer, line 83: could you please elaborate on why js/Promise here?

src/status_im/utils/universal_links/core.cljs
acolytec3: @yenda Thoughts here? Didn't we decide to handle it the way it's currently handled since we're checking for a  EIP681 url earlier up in the universal links stack and didn't want to have to parse it twice at this stage in the code?

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: @yenda Again, please confirm but this section occurs after the `parse-uri-handler` call has occurred so we're just resolving any ENS names and then updating the `send-transaction` details.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: @yenda preferred to use promises here so will have to comment on this approach since this isn't my code. I had a different implementation that recursively called `resolve-addresses` to resolve any ENS addresses but i think the idea was that this was a cleaner, presumably less resource intensive approach.

src/status_im/utils/universal_links/core.cljs
flexsurfer: let's wait for @yenda  but still, if you want to dispatch here it should be at least 
```
{:dispatch-n [[:parse-uri-handler ][:navigate-to :wallet-send-transaction]}
```

src/status_im/wallet/choose_recipient/core.cljs
flexsurfer: again, let's wait for @yenda :)  but if you want to dispatch it should be `{:dispatch []}`

src/status_im/wallet/choose_recipient/core.cljs
yenda: @flexsurfer there can be more than one ens name to resolve so using promises allows to do that in parallel and return once all names have been resolved.

src/status_im/wallet/choose_recipient/core.cljs
yenda: @flexsurfer is correct, dispatch is a side-effect, and it's not actually necessary here since you can call the function directly (you'll have to move the code though because request-uri-parsed is not defined at this point)

src/status_im/ethereum/eip681.cljs
```diff
@@ -41,21 +43,40 @@
...
+                (let [contract-address (get-in arguments [:function-arguments :address])]
+                  (if-not (or (not contract-address) (or (ens/is-valid-eth-name? contract-address) (ethereum/address? contract-address)))
+                    nil
+                    (merge {:address address
+                            :chain-id (if chain-id
+                                        (js/parseInt chain-id)
+                                        (ethereum/chain-keyword->chain-id :mainnet))}
+                           arguments)))))))))))
+
+(defn parse-uri-handler
```
yenda: this function can be deleted

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
```diff
@@ -61,7 +62,7 @@
                        ;:torchMode     (camera/set-torch camera-flashlight)
                        :onBarCodeRead #(when-not @read-once?
                                          (reset! read-once? true)
-                                         (re-frame/dispatch [:wallet/fill-request-from-url (camera/get-qr-code-data %) :qr]))}]]
+                                         (eip681/parse-uri-handler (camera/get-qr-code-data %) :qr))}]]
```
yenda: you should dispatch `::wallet/qr-code-request-read` here

src/status_im/utils/universal_links/core.cljs
yenda: @acolytec3 well since `parse-uri-handler` is parsing anyway calling it here will parse again. I think you need to handle the link and qr code cases because the context is different:
- with the qr code you are in a callback, so it makes sense to dispatch an event, so you can handle it in the re-frame context
- with the deep-link you are already in the re-frame context, so you don't have to dispatch an event

So only the qr code context should dispatch an event and the deep-link context should call `resolve-ens-addresses` directly.

I would recommend using @flexsurfer first proposition except that you should move the when-let inside the resolve-ens-addresses function.

yenda:
@acolytec3 Thanks you so much for the good job so far, we are fixing some rogue side effects and simplifying the code now which is important for the future and this should soon be ready to merge!

churik:
@acolytec3 @3esmit 

### ISSUE 1) Every uint256 resolves as 1e18 (tokens only)

I tested links with different amounts and assets and for me it looks like amount is not passed as it should be.
For example:
`ethereum:0xe41d2489571d322189246dafa5ebde1f4699f498/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1` should be resolved to ZRX > amount 1 (tried in couple of others wallets)
In fact:
photo_2019-11-25 14 22 37

So link like `ethereum:0xe41d2489571d322189246dafa5ebde1f4699f498/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1e-3` resolves amount  to 0.

#### ISSUE 2) [Object object] instead of link when scanning URL with wrong network id and no errors are shown if invalid address is passed

URL: `ethereum:xxx@1/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1e-3`

In nightly:
photo_2019-11-25 14 39 21

In PR this address can't be scanned - hence no error is shown.

Link: `ethereum:0xe41d2489571d322189246dafa5ebde1f4699f498@4/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1e-3`

photo_2019-11-25 14 43 12 jpeg 2019-11-25 14-44-21

####  ISSUE 3) ENS name is not resolved when asset is ETH
`ethereum:tanyatest1.eth?value=1e15` is not working (ETH), but `ethereum:0x744d70fdbe2ba4cf95131626614a1763df805b9e/transfer?address=tanyatest1.eth&uint256=2e1` (SNT, but the same address) is resolved as expected.

@yenda please let me know if smth from the above is acceptable and I should create separate issue for it.

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
acolytec3: qr-code-request-read doesn't appear to be a function.  Did you mean `resolve-ens-addresses`?

yenda: I was talking about dispatching a new event here called like that. And yes that would be an event that calls resolve-ens-address

src/status_im/utils/universal_links/core.cljs
acolytec3: Got it.  Will be resolved along those lines in next commit.

src/status_im/ethereum/eip681.cljs
acolytec3: Resolved in next commit.

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
acolytec3: Why not just dispatch `resolve-ens-addresses` directly?  I've done that approach in my next commit and it seems to work fine.

src/status_im/wallet/choose_recipient/core.cljs
acolytec3: Resolved in latest commit.

src/status_im/utils/universal_links/core.cljs
yenda: this is still wrong it won't work because resolve-ens-address will produce a map of effects, and it won't be used, you need to do something like this to merge the effects:

```
(fx/merge cofx
          (resolve-ens-addresses % url :deep-link nil]
          (navigation/navigate-to-cofx :wallet-send-transaction nil)
```

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
```diff
@@ -61,7 +62,7 @@
                        ;:torchMode     (camera/set-torch camera-flashlight)
                        :onBarCodeRead #(when-not @read-once?
                                          (reset! read-once? true)
-                                         (re-frame/dispatch [:wallet/fill-request-from-url (camera/get-qr-code-data %) :qr]))}]]
+                                         (re-frame/dispatch [:wallet.send/resolve-ens-addresses (camera/get-qr-code-data %) :qr]))}]]
```
yenda, line 65: ideally we want events to be named after what happened not an action, that is why fx/defn can take a vector or events in it metadata, so you can associate multiple events with a particular function. In that case it could be `::qr-code-request-scanned` for instance

acolytec3: I think I follow.  I'm still not sure I understand how the `::event-name` notation would apply here since that appears to only let you auto-resolve names within the current namespace (not for example, `choose-recipients.core` since the `qr-code-request-scanned` event is being called from `choose-recipients.view`.  I've pushed a new commit that adds `wallet.send/qr-code-request-scanned` as another alias for `resolve-ens-addresses`.

src/status_im/utils/universal_links/core.cljs
acolytec3: Sorry for misunderstanding.  Should be resolved along these lines in the latest commit.

acolytec3:
Updates on the issues @churik  raised:
1) I saw this issue before and have not yet found a solution for this.  Does the build produce the expected result?
2) I know the root cause and just need to code a fix.  Hopefully will include in a commit tonight.
3) This issue appears to be resolved.  I am able to resolve a ETH transaction with an ENS address like the ones presented above without issue.

3esmit:
Nice catches @churik, thanks! 

Does issue 1) is caused by this PR? Btw, `1e-3` is an invalid amount, instead it should be `1e3`. Seems like this is the case of opening another issue because the case here is just regarding ENS support.

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
yenda: `:hello` is a regular keyword
`:hi/hello` is a namespaced keyword
`::hello` is a namespaced keyword, if you are in `status-im.hi` namespace it will actually be resolved to `:status-im.hi` namespace
`::hi/hello` is also a namespace keyword, if you are in a namespace where the `status-im.hi` namespace is required like this: `[status-im.hi :as hi]` then it is equivalent to `:status-im.hi/hello` anywhere or `::hello` in the `status-im.hi` namespace

so `::` is shorthand for `use current namespace as namespace for this keyword` or if the keyword as a namespace it is shorthand for `resolve this namespace this is an alias`. If the alias doesn't exist it's a compilation error

yenda: So you register the event as `::event-name` and you use it as `::alias/event-name` in other namespaces. The advantage is that it helps preventing hidden circular dependencies because it forces to import the namespaces where the events are declared, as well as conflict in event names

churik:
@3esmit 
thanks!
> Does issue 1) is caused by this PR? Btw, 1e-3 is an invalid amount, instead it should be 1e3. Seems like this is the case of opening another issue because the case here is just regarding ENS support.

Well, not sure about this - it is because in current nightly (25/11/2019) amount is always resolved to 1 for me regardless of the passed parameter.

As for links - I made it using https://brunobar79.github.io/eip681-link-generator/ - please tell me if smth works wrong here, but in MetamMask generated links with amount worked as expected.
It can be in separate issue if this one is only about ENS.

And while testing discovered one more issue:

#### ISSUE 4) Error when scanning link with empty balance
Steps:
- create an account 
- scan QR with link like `ethereum:0x744d70fdbe2ba4cf95131626614a1763df805b9e/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1e2`
Expected result: can scan (no errors in nightly)
Actual result: 
photo_2019-11-26 19 08 44

acolytec3:
@yenda Got it.  I think I follow.  To make sure I understand.  I should declare
`resolve-ens-addresses` event metadata as something like:
`{:events [::resolve-ens-addresses :wallet.send/resolve-ens-addresses
:wallet.send/request-qr-code-read]}`

This way, we have appropriate aliases for resolve-ens-addresses when we're
directly resolving ENS addresses outside of the `choose-recipient.core`
namespace and then just `:dispatch ::resolve-ens-addresses` within the
event's namespace.  Right?

acolytec3:
Issue 2 should now be resolved with appropriate error messaging in place.
Screenshot_20191126-151251

I think issue 4 is really driven by having 2 ethereum addresses and not whether the count has a balance or not.  If I replace the first address in the URI with an ENS address like below, it resolves just fine.  More to come.
`ethereum:pay-snt.thetoken.eth/transfer?address=0xd21DB0e43048AcB94f428eD61dC244c82f1ff2a8&uint256=1e2`

acolytec3:
@yenda Regarding issue 4 that @churik highlighted, it only appears when you have only ethereum addresses in the URI.  If I change the straight function call to `request-uri-parsed` back to a `dispatch` event, it works.  Any idea why changing this causes that error to appear?

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -122,3 +134,41 @@
...
+                      (if (ens/is-valid-eth-name? address)
+                        (-> acc
+                            (update :paths conj path)
+                            (update :ens-names conj address))
+                        acc)))
+                  {:paths [] :ens-names []}
+                  [[:address] [:function-arguments :address]])]
+      (if (empty? ens-names)
+      ;; if there is no ens-names, we dispatch request-uri-parsed immediately
+        (request-uri-parsed message origin)
```
yenda:
```suggestion
        (request-uri-parsed cofx message origin)
```

acolytec3:
@yenda Inserting cofx into the call to `request-parse-uri` just generates an error saying fx/defn expects a map of cofx as the first parameter (since `cofx` isn't defined in the function).  I tried passing in an empty map as well and same error.  What am I missing since it seems like it shouldn't need any cofx since it's being called within the namespace.

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -122,3 +134,41 @@
   [{db :db} {:keys [address name public-key]} request?]
   {:db         (fill-request-details db {:address address :name name :public-key public-key} request?)
    :dispatch   [:navigate-back]})
+
+(fx/defn resolve-ens-addresses
+  {:events [:wallet.send/resolve-ens-addresses :wallet.send/qr-code-request-scanned]}
+  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db}
```
yenda:
```suggestion
  [{{:networks/keys [current-network] :wallet/keys [all-tokens] :as db} :db :as cofx}
```

flexsurfer:
@acolytec3 if you need any help with resolving conflicts, please ask, i'll help

acolytec3:
@flexsurfer I just pushed my latest commit which should resolve all of the existing issues, aside from that oddity about values not being correctly represented (which isn't part of my changes) and now building a version rebased on `develop`.  I'm going to test it locally and make sure it all still works but if I have issues, will let you know.

churik:
Please tell me when it should be retested.
And thanks for your work and patience!

acolytec3:
@flexsurfer Having trouble building on the latest `develop` branch.  I'm getting `stream error in the HTTP/2 framing layer` during the build process for the androkd-sdk and looks like that was reported in the status-core channel as well.  I've merged all the code and can push a commit but not sure if it will build or not until my local environment figures itself out.  Do you want me to go ahead and push that merged commit now and then you build and verify it's working?

flexsurfer:
@acolytec3  lets push changes anyway, cc @jakubgs

acolytec3:
@flexsurfer Okay, just pushed.  Hopefully this didn't break anything.  I haven't squashed yet and created a clone of the previous branch so we have it in case something goes awry.

flexsurfer:
@acolytec3, unfortunately, conflicts have been resolved incorrectly, some changes from develop were reverted

acolytec3:
@flexsurfer Ugh, sorry, I'll revert to my previous commit and try again once my current build finishes.  I'm starting from a clean nix build to see if that solves the stream error and it's going to take a while.

flexsurfer:
also this commit might be helpful to understand what has happened in develop https://github.com/status-im/status-react/commit/507cc5cf39e88b3898e23d95687af917e00bb29f#diff-09bf352f4017403bdf258a3d90eac06a

acolytec3:
@flexsurfer can you take a look at this commit and see if it's got the merges correct?  445fa32b5  I've taken another pass at merging on top of current `develop`.  I still can't get my local environment to build as the `run-android` make target is still hanging during the build process.  That just started this morning when I pulled down the latest `develop` branch so not sure I have any idea how to remedy it other than possibly doing a `make nix-purge` and starting fresh.  Has anyone else tried that?

jakubgs:
>`stream error in the HTTP/2 framing layer`
@acolytec3 

Could you gist the whole log?

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
```diff
@@ -0,0 +1,72 @@
+(ns status-im.ui.screens.wallet.choose-recipient.views
```
yenda, line 1: this view doesn't exist anymore in develop

src/status_im/ui/screens/wallet/components/views.cljs
```diff
@@ -6,11 +6,306 @@
             [status-im.ui.components.react :as react]
             [status-im.ui.components.styles :as components.styles]
             [status-im.ui.components.toolbar.view :as topbar]
+            [status-im.ui.components.tooltip.views :as tooltip]
```
yenda, line 9: most of the code here was deleted in develop

acolytec3:
> > `stream error in the HTTP/2 framing layer`
> > @acolytec3
> 
> Could you gist the whole log?

@jakubgs I didn't hang onto the console output from the make target.  Is there a build log somewhere it might have been outputted to?  I'm not real familiar with the logging from the build process and haven't been able to track it down.

src/status_im/ui/screens/wallet/choose_recipient/views.cljs
acolytec3: Makes sense now that I look at it again.  I'll try it again later today.

flexsurfer:
probably it might be better to create a new branch from latest develop and move all your changes there and file a new PR, but it's up to you

3esmit:
@flexsurfer thats an horrible thing, it should be a better way of doing this. I have dealt with merging conflicts in past and it was a matter of accepting the correct changes, and manually merging where it messes up. 

I am trying to understand here why it broke so bad.

acolytec3:
@3esmit It's not a big deal.  Once I figure out why my build environment refuses to work, it shouldn't take too long to reconstruct everything.  Just a case a bad timing where multiple PRs touching the same code going on at the same time.  I think I know what all needs to be updated and can hopefully get it done quickly once I get to the underlying environment issue fixed. 

@flexsurfer Whatever that stream error was yesterday has completely hosed my setup so I think I'm going to have to start from scratch.  I've tried `nix-purge` and then rebuilding and no luck.  None of the make commands related to building `startdev` or `android-ports` or `run-android` work and I'm not getting any error messages at all.  It just hangs when I try to to run them.  I'm going to completely blow away my local repos and start again so may be a while before I can do any actual coding.

3esmit:
Some problems Ive noticed in this process:
- there is no rebase merging commit - seems like the history was erased? @acolytec3 did you used any force commit?  Where are all the commits from the Status PR I tested? I would revert to the last working state and do a merge develop again, until it works. 
- @flexsurfer This PR is not aware of the changes done, and perhaps its using components or flows that changed, your help in merging would be appreciated. 
- @acolytec3 did not setup the signature of commits, and its annoying to see the error `x` everytime in your commits. I can help you configure this, but you should be able to do it on your own by reading this https://help.github.com/en/github/authenticating-to-github/managing-commit-signature-verification

acolytec3:
@3esmit I haven't done any force commits in the last few days but I did rebase on the latest develop yesterday and that's when everything got hosed.  Once my PC finishes reinstalling the build environment, I'll get the PR cleaned up and working again as I still have all the working code from the PR and just need to integrate it with the redesigned interface in the qr flow that @flexsurfer introduced in `develop`.

jakubgs:
@acolytec3 what do you mean by:
>I'm not getting any error messages at all.

I find that really hard to believe. What happens when you do `make shell TARGET_OS=android`?

cammellos:
@acolytec3 what do you use as a target platform for android? (geymotion, avd , real ) 
Make sure you are using the same `adb` version (needs to be `1.0.41`), if you are using genymotion or any software that might not be using the `adb` provided with nix.

jakubgs:
@acolytec3 after some talk with @cammellos I see that it's possible that the process hangs. I though what you were describing was exiting without error. But if it hangs than I'd like to see what processes are present. Could you try the `make` target that hangs for you without errors and show me the result of `ps faux`? Specifically the tree of processes started from the `make` command.

acolytec3:
@cammellos @jakubgs After running `make nix-purge`, deleting the entire local copy of repo, cloning a fresh one from github, and then running `make shell TARGET_OS=android` it's now producing this error.  I've attached the entire console output from the time I kicked off the make command. 
I've added the `ps faux` from `make startdev-android-real` as well as the place where it's hanging in the build process.

https://gist.github.com/acolytec3/8f6c9070889ea6e1acbe4999a3317864

acolytec3:
@cammellos @jakubgs Pause on any further worrying about my make issues.  I'm going to let the `startdev-android-real` go for a while and see if it ever fully builds when I'm on `develop`.  I had an old branch that I tried `startdev...` from and it did finally finish and boot to the REPL (though that branch has it's own issues so `run-android` doesn't work but I think that's related to the code in the branch and not the make recipe.  Will circle back if I don't make any progress the next 24 hours.

acolytec3:
@flexsurfer @yenda @3esmit I'm pausing work on this for the moment as I can't get the app to build.  Gist is below with the errors I see when trying to build from a completely clean repo cloned from status-react building off the current `develop` branch. I've seen comments in the status-core channel that others are maybe seeing similar errors but evidently not to the same level as me so not sure if it's just my environment or what though I know I haven't changed anything external. I'll fetch the latest `develop` branch in a day or two and try again but something that got merged in the past 36 hours has broken the build process for me.

https://gist.github.com/acolytec3/54ac7c1d085bccc48e6e65eaea94ba75

acolytec3:
@flexsurfer @yenda @churik @3esmit  My build issues seems to be sorted.  Submitted a clean PR rebased on current develop that imports all of the commits from this PR and seems to be working as expected.  I'm going to close this one.

https://github.com/status-im/status-react/pull/9563

PR #9257, opened 2019-10-21
Local Push Notifications POC

### LINKED ISSUE #7046: Group mobile push notifications by sender
# Problem

Currently, push notifications are not grouped (even though we call `setGroupSummary`), so the user potentially ends up with a ton of push notifications in the notification center.
 
# Implementation

We should:
- maintain a summary PN with a fixed ID
  - we call `setNotificationId` on it so we can update it later without creating a new one. Maybe the sender ID hash (`from`) can be used as the parameter so that we group by sender.
  - we call `setGroupSummary true` to mark it as the summary
- for both the summary and the individual message PNs, we call `setGroup` with a concatenation of `group-id` and the sender hash (e.g. `(str group-id "-" from)`)

Relevant documentation:
- https://developer.android.com/training/notify-user/group
- e.g. `AndroidNotification.md` in https://github.com/invertase/react-native-firebase-docs/tree/master/docs/notifications/reference

# Acceptance Criteria

PNs are grouped by sender in both Android and iOS.

<blockquote><div>Android Developers</div><div><strong><a href="https://developer.android.com/training/notify-user/group">Create a Group of Notifications  |  Android Developers</a></strong></div></blockquote>
<blockquote><div> GitHub</div><div><strong><a href="https://github.com/invertase/react-native-firebase-docs">invertase/react-native-firebase-docs</a></strong></div><div>📖 React Native Firebase documentation source files - please use the website to browse the docs. - invertase/react-native-firebase-docs</div></blockquote>

### LINKED ISSUE #7501: Too many notifications from Status
This is ridiculous:
screenshot_20190214-122203

yenda:
fix #7501
fix #7046 done for android

PNs feature is hidden behind a flag and disabled in releases.


jakubgs:
My guess would be that `make nix-update-gradle` needs to be run to fix the import errors from Gradle.

modules/react-native-status/android/src/main/java/im/status/ethereum/module/NewMessageSignalHandler.java
```diff
@@ -0,0 +1,313 @@
...
+            Log.e(TAG, "JSON conversion failed: " + e.getMessage());
+        }
+    }
+
+    private Person getPerson(String publicKey, String icon, String name) {
+        // TODO: invalidate cache if icon and name are not the same as
+        // the Person returned (in case the user set a different icon or username for instance)
+        // not critical it's just for notifications at the moment
+        // using a HashMap to cache Person because it's immutable
+        Person person = (Person) persons.get(publicKey);
```
rasom: Why do you need to cast it her? It seems that `persons` is defined as `HashMap<String, Person>`, should work without casting here.

modules/react-native-status/android/src/main/java/im/status/ethereum/module/NewMessageSignalHandler.java
```diff
@@ -0,0 +1,313 @@
...
+                ArrayList<StatusMessage> messages = chat.getMessages();
+                // parse the new messages
+                for (int j = 0; j < messagesData.length(); j++) {
+                    StatusMessage message = createMessage(messagesData.getJSONObject(j));
+                    if (message != null) {
+                        messages.add(message);
+                    }
+                }
+
+                if (messages.size() > 0) {
```
rasom: not a big deal but it could be `!messages.isEmpty()`

rasom (approved):

yenda:
@churik the important thing here is that the notification setting in profile is only on Android and that by default there is no notifications

jakubgs (approved):
Approving dependency changes only.

churik:
@yenda 
I tested the following scenarios on Android 7, 8, 9:
- creating account
- restoring account
- PN with the opened app (for 1-1 chat) and switched "Notifications" toggle. I see that functionality itself is in POC state, so suppose that shouldn't be checked(I mean that PNs are just come when app in the foreground, no action on click ect) 
- switching off / on notification setting
- checked that htere is no setting and PNs on IOS.

In all cases on Android notification setting in profile is switched off by default.
Please let me know if I missed smth.
Let's wait for e2e on this build.

yenda:
@churik thanks, what do you mean notifications just come when app is in the foreground? they don't come when app is in the background running as a service?

churik:
@yenda they come when app in the foreground and in the background and didn't come when app is closed. 
Video from LG V20 (Android8):
screencast 2019-11-06 18-18-58 2

PR #9258, opened 2019-10-21
Fix profile pic bugs (#9234)

### LINKED ISSUE #9234: "Remove current photo" is shown for new account and profile info is duplicated in header after any change
# Bug Report

#### Expected behavior
1. Removing photo is not shown when photo is not set
2. The header is not changed after removing/setting custom photo

If you scrolling and switching between profile and other tabs.

#### Actual behavior

removing_photo

### Reproduction

- Open Status
- Create new multiaccount
- Open Profile
- (issue 1) Tap on photo > Edit 
- (issue 2) Remove photo (or set new)

### Additional Information
- Status version: nightly 17/10/2019

acolytec3:
Fixes #9234

### Summary

"Remove current picture" option is displayed on edit profile bottom-sheet even if no profile picture has been set.  This resolves that user experience bug as well as the duplicate minimized header displayed after adding/removing a profile pic.

#### Platforms

- Android
- iOS

Profile/Contacts

##### Functional

- User profile updates - setting profile pic

##### Non-functional

NA

### Steps to test

(Bug 1 - Seeing Remove Current Picture when no profile pic set)
- Open Status
- Create new multiaccount
- Open Profile
- Tap on photo
- Add photo (if not already added)
- Remove photo
- "Remove current picture" should not be shown

(Bug 2 - Seeing profile-pic/public name on both minimized-toolbar and below)
- Same as above
- Should now no longer see minimized toolbar displayed after updating or removing profile pic

status: ready

Now looks like this when no profile picture is set

Screenshot_20191021-193253

This shows the minimized toolbar no longer being visible after adding/removing a profile pic.
mini-toolbar

### DIFF
--- a/src/status_im/ui/components/large_toolbar/view.cljs
+++ b/src/status_im/ui/components/large_toolbar/view.cljs
@@ -67,7 +67,7 @@
    - minimized-toolbar
    - flat-list-with-large-header"
   [header-in-toolbar nav-item toolbar-action-items header content list-ref]
-  (let [to-hide  (reagent/atom false)
+  (let [to-hide  (reagent/atom true)
         anim-opacity (animation/create-value 0)
         scroll-y (animation/create-value 0)]
     (animation/add-listener scroll-y (fn [anim]
@@ -83,3 +83,25 @@
                                           #(reset! to-hide false)))))
     {:minimized-toolbar [minimized-toolbar header-in-toolbar nav-item toolbar-action-items anim-opacity]
      :content-with-header [flat-list-with-large-header header content list-ref scroll-y]}))
+
+(defn add-listener [anim-opacity scroll-y]
+  (let [to-hide  (atom false)]
+    (animation/add-listener
+     scroll-y
+     (fn [anim]
+       (cond
+         (and (>= (.-value anim) 40) (not @to-hide))
+         (animation/start
+          (styles/minimized-toolbar-fade-in anim-opacity)
+          #(reset! to-hide true))
+
+         (and (< (.-value anim) 40) @to-hide)
+         (animation/start
+          (styles/minimized-toolbar-fade-out anim-opacity)
+          #(reset! to-hide false)))))))
+
+(defn minimized-toolbar-handler [header-in-toolbar nav-item toolbar-action-items anim-opacity]
+  [minimized-toolbar header-in-toolbar nav-item toolbar-action-items anim-opacity])
+
+(defn flat-list-with-header-handler [header content list-ref scroll-y]
+  [flat-list-with-large-header header content list-ref scroll-y])
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -28,7 +28,8 @@
             [status-im.wallet.core :as wallet]
             [taoensso.timbre :as log]
             [status-im.ui.screens.db :refer [app-db]]
-            [status-im.multiaccounts.biometric.core :as biometric]))
+            [status-im.multiaccounts.biometric.core :as biometric]
+            [status-im.utils.identicon :as identicon]))
 
 (def rpc-endpoint "https://goerli.infura.io/v3/f315575765b14720b32382a61a89341a")
 (def contract-address "0xfbf4c8e2B41fAfF8c616a0E49Fb4365a5355Ffaf")
@@ -271,6 +272,7 @@
                                :address address
                                :photo-path photo-path
                                :name name)
+                       (assoc :profile/photo-added? (= (identicon/identicon public-key) photo-path))
                        (update :multiaccounts/login dissoc
                                :error
                                :password))}
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -107,6 +107,8 @@
 (reg-root-key-sub :my-profile/advanced? :my-profile/advanced?)
 (reg-root-key-sub :my-profile/editing? :my-profile/editing?)
 (reg-root-key-sub :my-profile/profile :my-profile/profile)
+(reg-root-key-sub :profile/photo-added? :profile/photo-added?)
+
 ;;multiaccount
 (reg-root-key-sub :multiaccounts/multiaccounts :multiaccounts/multiaccounts)
 (reg-root-key-sub :multiaccounts/login :multiaccounts/login)
--- a/src/status_im/ui/screens/profile/contact/views.cljs
+++ b/src/status_im/ui/screens/profile/contact/views.cljs
@@ -115,6 +115,7 @@
     :allow-icon-change?     false
     :include-remove-action? false}])
 
+;;TO-DO Rework generate-view to use 3 functions from large-toolbar
 (views/defview profile []
   (views/letsubs [list-ref (reagent/atom nil)
                   contact  [:contacts/current-contact]]
--- a/src/status_im/ui/screens/profile/db.cljs
+++ b/src/status_im/ui/screens/profile/db.cljs
@@ -22,6 +22,7 @@
 (spec/def :profile/name correct-name?)
 (spec/def :profile/status (spec/nilable string?))
 (spec/def :profile/photo-path (spec/nilable base64-encoded-image-path?))
+(spec/def :profile/photo-added? (spec/nilable boolean?))
 
 (spec/def :my-profile/default-name (spec/nilable string?))
 (spec/def :my-profile/editing? (spec/nilable boolean?))
--- a/src/status_im/ui/screens/profile/events.cljs
+++ b/src/status_im/ui/screens/profile/events.cljs
@@ -38,7 +38,8 @@
              {:db (-> db
                       (assoc-in [:my-profile/profile :photo-path]
                                 (identicon/identicon (multiaccounts.model/current-public-key cofx)))
-                      (assoc :my-profile/editing? true))}
+                      (assoc :my-profile/editing? true
+                             :profile/photo-added? false))}
              (profile.models/save))))
 
 (handlers/register-handler-fx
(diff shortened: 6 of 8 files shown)


acolytec3:
@churik @errorists @yenda @flexsurfer 

Here's the PR.  I've solved issue #1 by checking if the profile image is set as a base64 png or jpeg.  This seems to be in line with how `update-picture` works and testing on my Android device works as expected.  I can take any recognized image regardless of extension and the app recognizes that it's been set as a profile pic and lets me remove it and replaces with the identicon.  Thereafter, ti will not present the Delete current picture option until I set a new profile pic.  

Will see if I can identify the root cause of bug 2 next.

acolytec3:
@churik @yenda @flexsurfer @errorists Both bugs should now be fixed.  The only thing I'm not sure how to solve is briefly seeing the minimized-toolbar flickering after updating/deleting the profile pic.  Any idea if that's a react-native thing?

rasom:
> Any idea if that's a react-native thing?

no it's not, it's almost likely animation we are running here https://github.com/status-im/status-react/blob/f7076016d8a15ea5441aae0f16c9c148c16340ea/src/status_im/ui/components/large_toolbar/view.cljs#L80

acolytec3:
@rasom I've tried adjusting the values in the section you referenced as well as the opacity parameter and it doesn't seem to have any positive impact.  I guess what isn't obvious to me is why it's rendering  at all prior to the animation listener being set since the view depends on the value of the `to-hide` parameter.

rasom:
@acolytec3 

> why it's rendering at all prior to the animation listener being set since the view depends on the value of the to-hide parameter

The problem with that `generate-view` function is that `to-hide` is re-defined each time when `generate-view` is called, and it is called each time when there is a change in `:multiaccount` sub (for instance when you add or remove the picture)... Things like this `to-hide` atom usually are defined outside the renderer function scope, but in this case it's a bit messy and in result it doesn't work as expected...

rasom:
https://github.com/status-im/status-react/commit/4b9617d739661177c4dd5c9884405eaa299c94cc

@acolytec3 that's how it can be fixed, you can adapt this commit to current PR, it should be cleared a bit and also applied to all places where `generate-view` has been called. If you don't want to work on this I could add another PR on top of yours tomorrow, just to fix that flickering issue.

acolytec3:
@rasom Looks like the only place generated view is used is in the contact view so it would only have to be replicated in one other place.  That said, can we not just put the `to-hide` state defaulted to true in the DB and update it as needed or does that have other unintended consequences?  It would be a lot cleaner to handle it within the `minimized-toolbar` function rather than having to put all of these changes in both places.  Thoughts?

rasom:
@acolytec3 I wouldn't put that value into `app-db` as it is used in two places already and that might make everything even worse because of possible state inconsistency.

rasom:
@acolytec3 also in case if we have these three separate functions instead of `generate-view`:
1. `to-hide` should be just `atom`, there is no need in `reagent/atom` there
2. `to-hide`'s default value should be `false`

acolytec3:
@rasom Sorry for not responding sooner.  I made the edits that @yenda requested around how we check for a profile-pic and pushed them to my original PR.  I'll take another look at your commit and see if I can figure out how it fits in.  When I looked at it yesterday, I wasn't following some of the changes but haven't had a change to really look at it closely.

acolytec3:
@rasom I've been trying to incorporate your recommended changes into my PR to resolve the flickering and while it does resolve the issue for the profile page, I can't seem to make it work on the contact page, which is the other page where `generate-view` is called.  [8dbafc1](https://github.com/status-im/status-react/pull/9258/commits/8dbafc1edafd882ee358266b24b1899a45a5d729) attempts to incorporate your recommended changes into the contact view but I'm getting errors when I try to login to the app.  Do you see anything obviously wrong with my attempt?  Modifying the contact view is a little beyond the scope of the original description of this issue so would be happy to hand off the flickering issue to you if you see a quick way to resolve things.  That issue feels like it might deserve it's own issue/PR and you have a better grasp of it than I do at present.

rasom:
@acolytec3 I will check it a bit later today, working on another task at the moment

rasom:
@acolytec3 just pushed a separate commit with fix

acolytec3:
@rasom  Thanks!  I built it locally and looks like the flicker is gone.  This PR should be fully ready for review at this point.  Not sure who's on point for review - @yenda @flexsurfer @rasom  but it should be ready to go at this point.

src/status_im/ui/components/large_toolbar/view.cljs
```diff
@@ -82,4 +82,26 @@
...
+
+         (and (< (.-value anim) 40) @to-hide)
+         (animation/start
+          (styles/minimized-toolbar-fade-out anim-opacity)
+          #(reset! to-hide false)))))))
+
+(defn minimized-tb [header-in-toolbar nav-item toolbar-action-items anim-opacity]
+  [minimized-toolbar header-in-toolbar nav-item toolbar-action-items anim-opacity])
+
+(defn flat-blabla [header content list-ref scroll-y]
```
rasom: @acolytec3 I actually put dummy names to that commit, that was kind of POC, let's change them to something more meaningful. That concerns `minimized-tb`, `flat-blabla`, `my-profile-1` (it can be removed I suppose), `mt`, `chw` and probably something else if I missed

src/status_im/ui/components/large_toolbar/view.cljs
```diff
@@ -82,4 +82,26 @@
                                           (styles/minimized-toolbar-fade-out anim-opacity)
                                           #(reset! to-hide false)))))
     {:minimized-toolbar [minimized-toolbar header-in-toolbar nav-item toolbar-action-items anim-opacity]
-     :content-with-header [flat-list-with-large-header header content list-ref scroll-y]}))
\ No newline at end of file
+     :content-with-header [flat-list-with-large-header header content list-ref scroll-y]}))
+
+(defn add-listener [anim-opacity scroll-y]
+  (let [to-hide  (reagent/atom true)]
```
rasom: `@to-hide` can be `false` by default, no need to run animation on initialisation

rasom: also no need to use `reagent/atom` here, it should be just `atom`

acolytec3: Will fix in next commit.

src/status_im/ui/components/large_toolbar/view.cljs
acolytec3: Will be cleaned up in next commit.

rasom:
@acolytec3 generate-view is not used anywhere it can be removed too https://github.com/status-im/status-react/blob/e0f2adc9e0f2c59f95f7e1b5bfce0223e6a1e30f/src/status_im/ui/components/large_toolbar/view.cljs#L64

acolytec3:
@rasom It's still used in the [contact screen](https://github.com/status-im/status-react/blob/bada7c4dd4f65b73203c49f45c1f8e479d205042/src/status_im/ui/screens/profile/contact/views.cljs#L135) under profile though I'm not sure if the same flickering issue impacts that view or not.

rasom:
@acolytec3 ok then, please add a todo note to `generate-view`, that it should be replaced with these 3 functions later. I will make that change after merging this PR.

rasom (approved):

acolytec3:
@rasom Done.  Thanks!

churik:
@acolytec3 please, rebase it to current develop - no builds for tests.
It helped in https://github.com/status-im/status-react/pull/9240

acolytec3:
@churik Ugh, sorry, I thought I did.  I'll rebase later today once I've got
access to my dev environment again.  On mobile now.

acolytec3:
@churik I rebased on current develop and force pushed.  Looks like it's still not building.  Is there any sort of error logs I could reference to see what the source of the error is?

churik:
As I can understand, related part:
```
2019-10-30T19:10:18.458Z] + IN_CI_ENVIRONMENT=1 nix-shell --run lein test-cljs --pure --no-out-link --keep LOCALE_ARCHIVE_2_27 --keep IN_CI_ENVIRONMENT --argstr target-os android --attr targets.leiningen.shell /home/jenkins/workspace/us-react_prs_android-e2e_PR-9258/default.nix
[2019-10-30T19:10:26.569Z] Checking for modifications in node_modules...
[2019-10-30T19:10:26.569Z] No modifications detected.
[2019-10-30T19:10:26.917Z] WARNING: Wrong number of args (1) passed to status-im.ui.screens.profile.user.views/header at line 223 /build/status-react-source-jsbundle/src/status_im/ui/screens/profile/user/views.cljs
[2019-10-30T19:10:26.917Z] Error encountered performing task 'cljsbuild' with profile(s): 'prod'
[2019-10-30T19:10:26.917Z] Subprocess failed
[2019-10-30T19:10:26.917Z] builder for '/nix/store/q6hir7hgrhrh3a3zimxkvrcrxnhcgfc7-status-react-build-jsbundle-android.drv' failed with exit code 1
[2019-10-30T19:10:26.917Z] error: build of '/nix/store/q6hir7hgrhrh3a3zimxkvrcrxnhcgfc7-status-react-build-jsbundle-android.drv' failed
```
Full log: 
[build_log.log](https://github.com/status-im/status-react/files/3792850/build_log.log)

acolytec3:
@churik Sorry about that.  A regression in my code snuck back in with a missing subscription.  I've added it back, rebased on current develop, and pushed.  Hopefully everything builds this time.  I did reconfirm that it's working on my device locally.

Serhy:
Above failed e2e tests are due to too slow faucet funds receiving. Nnot related to this PR (and apparently Status app in general, because I did not receive faucet from simpledapp ETHro in 10 mins either)
Tested build with Android 8.1, Android 6.0.1 and iOS 12, no regression found, both items fixed.
Thank you, @acolytec3

PR #9259, opened 2019-10-22
fix fleet settings not being saved to node config

yenda:
# Summary

Change of fleet in settings was not persisted to the node config. This should fix it though I am not sure how to test that, @cammellos and @oskarth you could try the PR and see if it solves your issue.

status: ready

### DIFF
--- a/src/status_im/log_level/core.cljs
+++ b/src/status_im/log_level/core.cljs
@@ -13,9 +13,9 @@
                         (assoc settings :log-level log-level)
                         (dissoc settings :log-level))]
     (fx/merge cofx
-              (multiaccounts.update/update-settings new-settings
-                                                    {})
-              (node/prepare-new-config {:on-success #(when (not= (node/get-log-level settings) (node/get-log-level new-settings))
+              (multiaccounts.update/update-settings new-settings {})
+              (node/prepare-new-config {:on-success #(when (not= (node/get-log-level settings)
+                                                                 (node/get-log-level new-settings))
                                                        (re-frame/dispatch [:logout]))}))))
 
 (fx/defn show-change-log-level-confirmation
--- a/src/status_im/node/core.cljs
+++ b/src/status_im/node/core.cljs
@@ -2,12 +2,12 @@
   (:require [re-frame.core :as re-frame]
             [status-im.constants :as constants]
             [status-im.ethereum.json-rpc :as json-rpc]
-            [status-im.fleet.core :as fleet]
             [status-im.native-module.core :as status]
             [status-im.utils.config :as config]
             [status-im.utils.fx :as fx]
             [status-im.utils.platform :as utils.platform]
-            [status-im.utils.types :as types]))
+            [status-im.utils.types :as types])
+  (:require-macros [status-im.utils.slurp :refer [slurp]]))
 
 (defn- add-custom-bootnodes [config network all-bootnodes]
   (let [bootnodes (as-> all-bootnodes $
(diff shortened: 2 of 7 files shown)


oskarth:
Thanks, Core should be owning this issue 100% as part of delivering v1. @adambabik even created this issue https://github.com/status-im/status-react/issues/9216 and core was pinged about it. It'd be good to have clarity and confidence that this will be dealt with from the Core team.

src/status_im/log_level/core.cljs
```diff
@@ -13,9 +13,9 @@
                         (assoc settings :log-level log-level)
                         (dissoc settings :log-level))]
     (fx/merge cofx
-              (multiaccounts.update/update-settings new-settings
-                                                    {})
-              (node/prepare-new-config {:on-success #(when (not= (node/get-log-level settings) (node/get-log-level new-settings))
+              (multiaccounts.update/update-settings new-settings {})
+              (node/prepare-new-config {:on-success #(when (not= (node/get-log-level settings)
+                                                                 (node/get-log-level new-settings))
```
rasom, line 18: Are there any changes at all? Is it just code formatting?

src/status_im/node/core.cljs
```diff
@@ -74,13 +74,30 @@
...
+
+(defn current-fleet-key [db]
+  (keyword (get-in db [:multiaccount :settings :fleet]
+                   config/fleet)))
+
+(defn get-current-fleet
+  [db]
+  (get (fleets db)
+       (current-fleet-key db)))
+
```
rasom, line 94: Why do we need to move this code form `fleet.core` to `node.core`?

yenda: `fleet.core` is mostly about saving a new fleet config and to do that you need to call `node.core`. So if `get-current-fleet` is in `fleet.core` as well you get a circular dependency. I suppose we could make a fleet.db ns otherwsie. Right now it's similar to `log-level.core` and `get-log-level` is in `node.core` too.

src/status_im/log_level/core.cljs
yenda: just formatted to be the same layout as in `fleet.core` and more readable

src/status_im/node/core.cljs
rasom: I would go for that `fleet.something` ns, then we actually don't need to make changes in other namespaces except `fleet.core` and `node.core`.

yenda: ok but shouldn't I change log-level namespace as well for consistency then? Also it will still change other namespaces, but just the ns imported.

rasom: We def can do it without changes in other namespaces, the call can go through fleet.core, not a big deal though. 

Regarding `log-level`, it's not objective of this PR so I wouldn't do this here. Actually if everyone is fine with putting that fleet related code into `node.core` no problem. But it's a bit puzzling then what are `fleet.core` and `node.core` responsible for.

src/status_im/node/core.cljs
```diff
@@ -74,13 +74,30 @@
       (if utils.platform/desktop? ""
           config/log-level-status-go)))
 
+(def default-fleets (slurp "resources/config/fleets.json"))
+
+(defn fleets [{:keys [custom-fleets]}]
+  (as-> [(default-fleets)] $
+    (mapv #(:fleets (types/json->clj %)) $)
```
cammellos, line 82: Should we not rather do the conversion in the `def` above?

cammellos (approved):

cammellos:
@yenda is this good to go?

PR #9262, opened 2019-10-22
include status-go params to set version and commit

### LINKED ISSUE #9270: Status-go version missing from About page
## Problem
Right now the app is not showing the `status-go` version:

This also translates to that version not being shown in peer name when connecting to other peers.

## Notes
Based on my investigation and help from Eric and Adam I found out this:

#### Before
We used to set the `Version` value by using the `-X` flag in `-ldflags` when building `status-go`:
>```Makefile
>BUILD_FLAGS_MOBILE ?= $(shell echo "-ldflags '\
> -X main.buildStamp=`date -u '+%Y-%m-%d.%H:%M:%S'` \
> -X github.com/status-im/status-go/params.Version=$(RELEASE_TAG) \
> -X github.com/status-im/status-go/params.GitCommit=$(GIT_COMMIT)'")
>```
>https://github.com/status-im/status-go/blob/81b0a7b2/Makefile#L25-L28

And this in turn sets the value of `Version` in the `NodeConfig` structure when `NewNodeConfig` is called. But currently it appears that it is not called at all. So setting it this way does nothing.

#### Nix
Currently we are not even using the `-X` flags when building `status-go` for mobile:
https://github.com/status-im/status-react/blob/57b1722863148579e0b6d4413c912e55106a3749/nix/status-go/default.nix#L56-L57

#### Now
Right now the config of `status-go` node is modified when a Status account is created/recovered via the `settings_saveNodeConfig` RPC call that modified the `NodeConfig` struct:
https://github.com/status-im/status-react/blob/57b1722863148579e0b6d4413c912e55106a3749/src/status_im/node/core.cljs#L138-L147
But this config does not include a value for `Version`, only the `Name` set here:
https://github.com/status-im/status-react/blob/57b1722863148579e0b6d4413c912e55106a3749/src/status_im/node/core.cljs#L56-L57

## Solutions
I can see 2 options here, either:

* Modify `status-go` so that setting `Version` at compile time with `-X` works
* Modify `status-react` so it sets the right `Version` value with `settings_saveNodeConfig`

The first option makes most sense to me.

jakubgs:
While working on https://github.com/status-im/status-go/pull/1648 which implements new Prometheus metrics for `status-go` based on peer name I've stumbled upon the lack of `status-go` version in the peer name of the mobile app, which can be also found in the "About" page:
Screenshot_20191022-144004
Where the proper peer name should look like:
```
StatusIM/v0.34.0-beta.3/android-arm64/go1.12.9
```

Adam showed me that the `status-go` version comes from the `params`:
https://github.com/status-im/status-go/blob/81b0a7b2/params/config.go#L248-L252
Which used to be sed via the `-X` flag for `-ldflags` in the `Makefile`:
https://github.com/status-im/status-go/blob/81b0a7b2/Makefile#L22-L23

This doesn't work, because `NewNodeConfig` function is never called, so `config.Version` stays empty. This was fixed in https://github.com/status-im/status-go/pull/1656 and this PR implements that fix in `status-react` by adjusting how `status-go` is built by Nix, also updates its version.

Fixes #9270.

### DIFF
--- a/nix/status-go/build-desktop-status-go.nix
+++ b/nix/status-go/build-desktop-status-go.nix
@@ -14,19 +14,25 @@ let
     inherit buildGoPackage go xcodeWrapper utils;
   };
 
+  # Remove desktop-only arguments from args
   args = removeAttrs args' [
-    "goBuildFlags"
-    "goBuildLdFlags"
-    "outputFileName"
-    "hostSystem"
-  ]; # Remove desktop-only arguments from args
+    "goBuildFlags" "goBuildLdFlags" "outputFileName" "hostSystem"
+  ];
+
   buildStatusGoDesktopLib = buildStatusGo (args // {
     buildMessage = "Building desktop library";
     #GOOS=windows GOARCH=amd64 CGO_ENABLED=1 go build ${goBuildFlags} -buildmode=c-archive -o $out/${outputFileName} ./lib
-    buildPhase = ''
+    buildPhase =
+      let
+        CGO_LDFLAGS = stdenv.lib.concatStringsSep " " goBuildLdFlags;
+      in ''
       pushd "$NIX_BUILD_TOP/go/src/${goPackagePath}" >/dev/null
 
-      go build -o $out/${outputFileName} ${goBuildFlags} -buildmode=c-archive ${goBuildLdFlags} ./lib
+      go build -o $out/${outputFileName} \
+          ${goBuildFlags} \
+          -buildmode=c-archive \
+          -ldflags='${CGO_LDFLAGS}' \
+          ./lib
 
       popd >/dev/null
     '';
--- a/Makefile
+++ b/Makefile
@@ -310,7 +310,7 @@ android-install: export _NIX_ATTR := targets.mobile.android.adb.shell
 android-install: export TARGET_OS ?= android
 android-install: export BUILD_TYPE ?= release
 android-install:
-	adb install result/app.apk
+	adb install result/app-$(BUILD_TYPE).apk
 
 _list: SHELL := /bin/sh
 _list:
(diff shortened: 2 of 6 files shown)


jakubgs:
This fix didn't seem to do it. I installed the build but I still don't see the version in the name:
I guess it's the same issue Pedro had with passing multiple flags to `-ldflags`, as the comment states:
https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/nix/status-go/default.nix#L56-L57

jakubgs:
This fix doesn't work, because `NewNodeConfig` function is never called, so `Version` stays empty.
See #9270 for more details. I'll have to figure out some better fix for this.

jakubgs:
Once https://github.com/status-im/status-go/pull/1656 is merged this should work just fine.

nix/status-go/build-desktop-status-go.nix
```diff
@@ -14,19 +14,25 @@ let
...
   };
 
+  # Remove desktop-only arguments from args
   args = removeAttrs args' [
-    "goBuildFlags"
-    "goBuildLdFlags"
-    "outputFileName"
-    "hostSystem"
-  ]; # Remove desktop-only arguments from args
+    "goBuildFlags" "outputFileName" "hostSystem" "goBuildLdFlags"
```
pedropombeiro: Why not keep both flags-related names together?
```suggestion
    "goBuildFlags" "goBuildLdFlags" "outputFileName" "hostSystem"
```

pedropombeiro (approved):
Awesome work @jakubgs!

nix/status-go/build-desktop-status-go.nix
jakubgs: Sure.

jakubgs:
I'm waiting to tag a new release of `status-go` so we're not using a hash. That's also another thing I'll need to fix somehow. Maybe some kind of validation for what can be used for value of `version` in `status-go-version.json`. Or Nix expressions could use the contents of the `VERSION` file from `status-go` instead. Not sure which is better.

PR #9263, opened 2019-10-22
Receive messages perf improvements

cammellos:
This commit includes a few performance fixes: 

1) Pass a string payload instead of an hex encoded string, to avoid unecessary conversion 

2) Don't js->clj on messages, as that's fairly expensive and we can get away without 

3) Don't use `pr-str` `read-string`, rather convert to json

## Findings

Looking at the performance of the app when receiving messages through the profiler, we can see that the bottleneck is (as we already know) in the JS thread, which is hogged for the whole time preventing any input from the user to be handled, therefore "freezing" the app.

Looking a bit deeper into it, we can split the workload in two sections (events):

The actual receiving of the raw messages from status-go https://github.com/status-im/status-react/blob/792df3276d806eb92089d8c5e7504072c146f0a6/src/status_im/signals/core.cljs#L61 `messages.new` event.
The processing of the actual Text Messages https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/src/status_im/events.cljs#L638

Both events should be handle in 16ms for keeping the ui responsive.
Investigation showed that neither or them is close to that.

With regard of the first event the bottlenecks identified where:

1) Conversion from `js->clj`, this was by far the most expensive operation, and it has been addressed by keeping the payload in `# js`.
2) Calling `hexToUtf8`, this was not necessary as we can just provide the `string` version from `status-go` at no cost.
3) Transit parsing, to a minor extent (not changed)
4) Spec validation is expensive (not changed)

In the second part of the event (currently heavier then the first part), the most time is spent running the following code (roughly in order of performance):

https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/src/status_im/chat/models/message.cljs#L276 

https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/src/status_im/chat/models/message.cljs#L277 

https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/src/status_im/chat/models/message.cljs#L264

https://github.com/status-im/status-react/blob/474ff00c7f0ebab86bda26be6b77faee26aa8f73/src/status_im/chat/models/message.cljs#L282

### How to address

First we should be as already discussed moving most of the code to status-go, that would be the natural first step. Before we do that, it would be though useful to simplify the message handling code and verify that if we ditch some performance improvements (say we don't use message groups anymore, but we get an already sorted list from status-go), we have still an overall positive effect.

Another thing is that we currently naively save messages in the db (including recalculating groups etc), even if the user is not on that chat (or the message is not in the current pagination window), which is wasteful.

As discussed with @yenda there's also a fair amount of regex parsing for each message received (this is more problematic as the longer the message is), my suggestion would be to parse messages on the sender side, and send metadata with each message. This might incur in higher bandwidth cost, but has some benefits:

1) Better perf. on receiving a message
2) Separate the way a user inputs a message from how it is rendered, allowing us to add styling/features more easily

For example:
A message that is composed as: 

```
Hey @nobody **this** is a www.link.com or check this #channel
```
Would be sent across the wire as something like (copying twitter api, which does something similar, also making it verbose, but for some fields the bandwidth can be the same, i.e bold , two integers, for asterisks):

```

{:text "Hey @nobody this is a www.link.com or check this #channel"
  :entities {:mentions [[x,y,pk]], :links [[x,y]], :channels [[x,y,channel]], :bold [[x,y]]}

```

Changing input from markdown to html for example will not change the metadata associated, and there's a bit of a better upgrade path (as markings are stripped, if a client for example can't see bold it won't see `**`, it will just be shown as plain text)
But I guess that's something for later.

Another thing that we should be doing is to split the workload in manageable chunks.
Currently we just process whatever the amount of messages is received in one go, i.e (200, 300, 500 etc).
This gives us the shortest time to process messages, but it's detrimental to the user experience as it's definitely hogging the JS thread and results in the app freezing.
We should probably split off the messages in batches that can be processed in less than 16ms and dispatch after each chunk https://github.com/day8/re-frame/blob/69a3fcd411369fff65404e7d6a924ce1968e2dba/docs/Solve-the-CPU-hog-problem.md , as described.
I have done that but the results were mixed. On a more powerful device, I saw a slow down (but no real freeze) after pulling 700 messages in one go, which was promising, compared to 10 seconds of freezing. It took much longer though to process the messages overall (roughly a minute or so), but there's room for improvement (see points above). From inspection I noticed that the initial event was mostly under 16ms, while the second event would often go above (~100ms). Different batch sizes were tried, 1 is the one that gave the smoothest experience.

On a slower device though, the UI completely choked as events took much longer than 16ms, so this is something that we should revisit.

### TLDR

We should move stuff to status-go, and at the same time simplify and remove computations that can be avoided.
On my side, I will address some of these issues as I work on related code for task in scope for v1.

status: ready


yenda (approved):

cammellos:
@churik is this a false negative?

rasom:
@cammellos 
> Both events should be handle in 16ms for keeping the ui responsive.
Investigation showed that neither or them is close to that.

on any device?

src/status_im/transport/message/core.cljs
```diff
@@ -26,13 +26,19 @@
...
-  [cofx now-in-s filter-chat-id message]
+  [cofx now-in-s filter-chat-id message-js]
   (let [blocked-contacts (get-in cofx [:db :contacts/blocked] #{})
-        {{:keys [payload timestamp]} :message
-         metadata :metadata
-         raw-payload :raw-payload} (add-raw-payload message)
+        payload (.-payload message-js)
+        timestamp (.-timestamp (.-message message-js))
+        metadata-js (.-metadata message-js)
+        metadata {:author {:publicKey (.-publicKey (.-author metadata-js))
```
rasom, line 34: https://github.com/mfikes/cljs-bean probably could be useful here, although if we are moving to go anyway...

rasom:
Do we have any specific numbers how faster conversion became after these changes?

cammellos:
@Serhy benchmarked it here (40/45% improvement, same code when it was benchmarked as this PR https://github.com/status-im/status-react/pull/9230#issuecomment-544466568 , later changes have not been included as I was testing stuff)

src/status_im/transport/message/core.cljs
cammellos: Yes, I think we should be using it eventually, as at some point I think we will pass the parsed JSON messages from status-go, and then it would come handy

cammellos:
@rasom 

> on any device?

I didn't get anywhere close to that on any device ( I don't have top of the range devices, but I have an Huawei mediapad which should have decent performance), for what is worth in desktop running emulator (with the built apk) I also was not able to be under 16ms for the second part of the event, although not sure that's reliable.

rasom (approved):

churik:
@cammellos no, it is not false failure - it is an issue, however, I can't reproduce it manually.
Basically sometimes list of assets is not loaded and instead of all your tokens / collectibles you will see only  ETH.

Issue is pretty serious, but I didn't see it when tested manually

PR #9305, opened 2019-10-28
[WIP] Handle signals separately

yenda:
Handling each type of status-go signal in a separate listener on status-react side facilitates code separation and allow fine grained handling behavior. This PR uses one listener per type of event instead of the previous single listener with case dispatching on clojurescript side.

(could be removed)
Additionally for Android a Javascript object is passed directly to avoid JSON parsing in javascript. This saves a few ms of parsing on very fast devices, probably more on slower devices.
The object is wrapped in a cljs-bean which avoids the expensive js->clj conversion while still allowing cljs idiomatic usages such as accessing field by keywords and destructuring on the js object.

TODO:
- [ ] implement native part for desktop @vkjr 
- [ ] implement native part for iOS
- [ ] separate code path for iOS so that it parses the JSON


rasom:
Does it mean that adding a new signal from now on requires adding code in four places?

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -117,11 +122,113 @@ public String getNoBackupDirectory() {
...
+        }
+        return map;
+    }
+    
+    private static WritableArray convertJsonToArray(JSONArray jsonArray) throws JSONException {
+        WritableArray array = new WritableNativeArray();
+        
+        for (int i = 0; i < jsonArray.length(); i++) {
+            Object value = jsonArray.get(i);
+            if (value instanceof JSONObject) {
```
rasom, line 158: Not like it's very important, but `instanceof` is kind of a code smell and it would probably be better if you define a separate method for each type. Also `instanceof` should be a bit slower than dispatching, though not sure it will have a serious impact here.

yenda: @rasom thanks for the review, maybe I should get rid of this, it's just some code I got from a blog article, I wanted to test passing js objects directly vs parse json on react side. It saves a few ms but I don't know if it is worth it. I'm very rusty in Java how would you replace that?

yenda:
@rasom 

>Does it mean that adding a new signal from now on requires adding code in four places?

it depends, here I choose to not handle unknown signals at all, but we could handle them.

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
rasom: Probably by defining a separate `convert` method for each type and then allow it to choose the right one based on the type. 
```java
private static void convert(WritableArray arr, JSONObject obj) {
    array.pushMap(convertJsonToMap((JSONObject) obj));
}
...
private static void convert(WritableArray arr, Object obj) {
    array.pushString(obj.toString());
}

//and then in convertJsonToArray
...
for (int i = 0; i < jsonArray.length(); i++) {
     convert(array, jsonArray.get(i));
}
...
```
probably something like this will work

rasom:
I mean, if we kind of make cljs code clearer by forcing cljs devs to add changes to java/objc/cpp code each time we create a new signal I'm not sure that it actually is worth doing.

yenda:
@rasom it's just about adding a case on each platform, and we very rarely add new events. 

Also it would encourage to improve the api between status-go and status-react, because ideally you shouldn't need any custom handling of a particular event to remove unnecessary stuff.

Might be the case that ultimately it's just about adding the name of the event to a list of supported events. 

We could also test a default case but I think iOS would warn on unhandled event:
https://callstack.com/blog/sending-events-to-javascript-from-your-native-module-in-react-native/
> One more thing. You will get a warning notice if you emit an event while there are no listeners. The easiest way to optimize your module and avoid this warning is to override startObserving and stopObservingmethods where you set flag hasListeners and send events only when it is YES.

Interestingly we are using the deprecated `sendAppEventWithName` to dispatch the events on iOS

rasom:
>  it's just about adding a case on each platform, and we very rarely add new events.

I don't think that the point that we do this rarely makes situation better. If we can write the code on each platform which will allow to dispatch a separate event for every signal without touching native side each time, that would be nice, otherwise I just don't see why this change is necessary. Not like it provides any perf improvement (we can covert json to js object on native side without separate events) or adds a new feature.

yenda:
@rasom I think it is doable, however since iOS wants a list of supported events, we either still need to maintain that list, or we need to wrap iOS events as it is done right now, so under the hood the `status/add-event-listener` would still be a case that dispatch on event type for iOS.

flexsurfer:
hey @yenda still relevant?

flexsurfer:
hey @yenda still relevant?

flexsurfer:
feel free to reopen

PR #9373, opened 2019-10-31
Wallet update method

churik:
- Reworked `wait_balance_changed_on_wallet_screen` method and replaced asset-specific classes.
- Added check of changing amount in `test_transaction_appears_in_history`
- Added test test_fetching_balance_after_offline
- Temporary changed locator for UnreadMessagesCountText
- Replace mailserver address with staging mailserver

### DIFF
--- a/test/appium/views/wallet_view.py
+++ b/test/appium/views/wallet_view.py
@@ -23,7 +23,7 @@ class TransactionHistoryButton(BaseButton):
 
     def __init__(self, driver):
         super(TransactionHistoryButton, self).__init__(driver)
-        self.locator = self.Locator.xpath_selector("//*[@text='History']")
+        self.locator = self.Locator.text_selector("History")
 
     def navigate(self):
         from views.transactions_view import TransactionsView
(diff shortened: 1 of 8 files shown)


test/appium/views/wallet_view.py
```diff
@@ -391,36 +383,42 @@ def get_usd_total_value(self):
...
+        eth = self.get_asset_amount_by_name('ETHro')
         expected_usd = round(eth * expected_rate, 2)
         percentage_diff = abs((usd - expected_usd) / ((usd + expected_usd) / 2)) * 100
         if percentage_diff > 2:
             errors.append('Difference between current (%s) and expected (%s) USD balance > 2%%!!' % (usd, expected_usd))
         else:
             self.driver.info('Current USD balance %s is ok' % usd)
 
-    def wait_balance_changed_on_wallet_screen(self, expected_balance=0.1, wait_time=300):
+    def wait_balance_changed_on_wallet_screen(self, asset = 'ETHro', check_that_initial_balance_updated=False, expected_balance=0, wait_time=300):
```
Serhy: I have to say this method became overcomplicated for me with new logic for how it works. 

1) Not clear when `check_that_initial_balance_updated` supposed to be used (kind of method dedicated to check that balance_changed_on_wallet screen and plus initial_balance_updated?. I looked then to section where it used and again can't understand.
2) Imo checking initial_balance inside the method is not the best approach. It narrows down area when method could be used. For example I will afraid to use it if I sent TX in Dapp, returned to Dapp main screen - then to wallet multiaccount view - and finally to account view to check balance is changed (because during those navigation TX could be already mined and changed).
3) Two 'elif's do the same job but have different conditions.
4)  Method can not be used if I expect balance = 0 (for example send all tokens I have)

Serhy (approved):

PR #9378, opened 2019-11-02
Set max decimal places at 6 for signing sheet

### LINKED ISSUE #9235: Amount and Network Fee text overlap in 'Sending' screen when there are long values
# Bug Report

## Problem
Transaction amount and network fee often calculated up to eighteenth digit precision  (in case of 18-decimals asset). In that case the 'Amount' and 'Network Fee' text on the 'Sending...' screen overlapped by it's value. 

#### Expected behavior
Long values of amount and network fee does not overlap the text

`I would ask to truncate … the numeric values at the end so that the entire text label of Amount / Network Fee is visible. Both are already displayed in their entirety in the header of the bottom sheet.`

#### Actual behavior
Screenshot 2019-10-18 at 09 48 49

### Reproduction

- Navigate to `Wallet` -> `Status account `-> `Send` and select any Recipient
- Enter in Amount field the `0.0848969596969696969` value -> `Sign transaction`
- Tap `Network Fee` and update gas price to 1.0989898

### Additional Information

- Status version: develop nightly 0.14.0 (2019171002)
- Operating System: <Android and iOS>

acolytec3:
Fixes #9235 

### Summary

Amount and Network fee values overlap labels on signing sheet.  This resolves the issue by setting a maximum length and truncating any characters/digits in the value that exceed that length

### Review notes
As currently designed, it has the possibility of displaying a 0 in the last position of each value since I'm just trimming the string version of each value to fit in the field without overlapping the label.

#### Platforms
- Android
- iOS

##### Functional

- wallet / transactions

### Steps to test

* Open Status and create new account
* Navigate to Wallet -> Status account -> Send and select any Recipient
* Enter in Amount field the 0.0848969596969696969 value -> Sign transaction
* Tap Network Fee and update gas price to 1.0989898
* Verify that amount and network fee amounts are truncated to 6 digits of precision.

status: ready 
Screenshot_20191115-113012

### DIFF
--- a/src/status_im/ui/screens/signing/views.cljs
+++ b/src/status_im/ui/screens/signing/views.cljs
@@ -21,7 +21,8 @@
             [clojure.string :as string]
             [status-im.ui.screens.signing.styles :as styles]
             [status-im.react-native.resources :as resources]
-            [status-im.ui.screens.hardwallet.pin.views :as pin.views]))
+            [status-im.ui.screens.hardwallet.pin.views :as pin.views]
+            [status-im.utils.utils :as utils]))
 
 (defn hide-panel-anim
   [bottom-anim-value alpha-value window-height]
@@ -229,7 +230,7 @@
       :title :t/send-request-amount
       :error amount-error
       :accessories [[react/nested-text {:style {:color colors/gray}}
-                     [{:style {:color colors/black}} (str (or amount 0))]
+                     [{:style {:color colors/black}} (utils/format-decimals amount 6)]
                      " "
                      (or display-symbol fee-display-symbol)
                      " • "
@@ -249,7 +250,7 @@
       :title       :t/network-fee
       :error       gas-error
       :accessories [[react/nested-text {:style {:color colors/gray}}
-                     [{:style {:color colors/black}} fee]
+                     [{:style {:color colors/black}} (utils/format-decimals fee 6)]
                      " "
                      fee-display-symbol
                      " • "
--- a/src/status_im/utils/utils.cljs
+++ b/src/status_im/utils/utils.cljs
@@ -1,5 +1,8 @@
 (ns status-im.utils.utils
-  (:require [status-im.i18n :as i18n]
+  (:require [clojure.string :as string]
+            [goog.string :as gstring]
+            [goog.string.format]
+            [status-im.i18n :as i18n]
             [status-im.react-native.js-dependencies :as rn-dependencies]
             [re-frame.core :as re-frame]
             [status-im.utils.platform :as platform]
@@ -151,3 +154,9 @@
   (if platform/desktop?
     (js/clearInterval id)
     (.clearInterval rn-dependencies/background-timer id)))
+
+(defn format-decimals [amount places]
+  (let [decimal-part (get (string/split (str amount) ".") 1)]
+    (if (> (count decimal-part) places)
+      (gstring/format (str "%." places "f") amount)
+      (or (str amount) 0))))
--- a/test/cljs/status_im/test/utils/utils.cljs
+++ b/test/cljs/status_im/test/utils/utils.cljs
@@ -49,3 +49,9 @@
   (is (= {:a 1 :b 2} (u/deep-merge {:a 1} {:b 2})))
   (is (= {:a {:b 1 :c 2}} (u/deep-merge {:a {:b 1 :c 1}} {:a {:c 2}})))
   (is (= {:a {:b {:c 2}} :d 1} (u/deep-merge {:a {:b {:c 1}} :d 1} {:a {:b {:c 2}}}))))
+
+(deftest format-decimals-test
+  (is (= "1" (uu/format-decimals 1 5)))
+  (is (= "1.1" (uu/format-decimals 1.1 5)))
+  (is (= "1.111111" (uu/format-decimals 1.111111 7)))
+  (is (= "1.1" (uu/format-decimals 1.111 1))))


acolytec3:
@StatusSceptre @rachelhamlin @errorists Submitted an initial PR.  I'm not 100% sure if this approach will work on all screens or not since I'm picking a maximum number of characters for the amount/network fee values based on when the overlap occurs on my Oneplus 6T.  What I don't know is if the size of the window and therefore the number of characters available for that field always scales exactly or if that maximum # characters will change based on the screen size or not.

3esmit (dismissed):
The code looks good, but build fails on lint check.

rasom:
It will not work on all screens, for instance on galaxy s9 "Network fee" is truncated now:


imo preferred solution will be to set `:min-width` for these labels (in order to make sure that at least English version fits there) and rearrange ` :accessories` view of these `list-item`s so that the value will be truncated according to [react native's `:ellipsize-mode `property](https://facebook.github.io/react-native/docs/text#ellipsizemode). In this case calculation of length happens on native side and is flexible.

acolytec3:
I wondered about that. I tried setting a minWidth prop on the labels and
setting an `ellipsizemode tail` on the two value fields but it didn't seem
like it did anything. The value fields still ran over the labels on larger
values and never actually ellipsized. Would redefining the title label as a
component with the minWidth solve that?

rasom:
@acolytec3 agree, tried to play with that `list-item` component, and although the solution mentioned in my previous comment is definitely doable it is really hard to implement it using `list-item` and it is not worth efforts almost likely. I think we can just decrease `field-width` for fee field and go with that for now.

acolytec3:
Sorry for slow follow up here. @rasom where are you suggesting to add that field-width prop. I've tried applying it to the labels and the nested text elements but seems like it was ignored by react.   Is it worth just don't something like setting the length of the calls equivalent to some maximum percentage of the screen width? We should be able to calculate the pixel width of the value strings given the font size and just tweak my formatted value component to truncate accordingly, I think anyway.

flexsurfer:
@acolytec3 ,thanks for your effort, after discussion with @andmironov  we decided to go another way, and cut the number to 6 digits after the comma

acolytec3:
Got it. Do you want me to revise my PR to reflect that or are you going to include in a separate commit?

flexsurfer:
if you could revise would be great, thanks!

acolytec3:
@flexsurfer @andmironov  Done.  Renders like this.  Is this what you have in mind?
Screenshot_20191115-113012

flexsurfer:
great thanks @acolytec3 , only we don't need dots in the end anymore

acolytec3:
@flexsurfer Removed the `...` in the latest commit.

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -242,14 +250,14 @@
                      " "
                      (str (:code wallet-currency))]]}]))
 
-(defn fee-item [prices wallet-currency fee-display-symbol fee gas-error]
+(defn fee-item [prices wallet-currency fee-display-symbol fee gas-error window-width]
```
flexsurfer: is `window-width` still needed ?

Serhy:
@acolytec3 could you also rebase develop please?

src/status_im/ui/screens/signing/views.cljs
acolytec3: Nope, removed in my latest commit.

acolytec3:
@Serhy @churik  Squashed and rebased on `develop`.

churik:
@acolytec3 can you please rebase your branch to current `develop`?

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -274,7 +282,8 @@
                   {:keys [amount-error gas-error]} [:signing/amount-errors (:address from)]
                   keycard-multiaccount? [:keycard-multiaccount?]
                   prices                [:prices]
-                  wallet-currency       [:wallet/currency]]
+                  wallet-currency       [:wallet/currency]
+                  window-width [:dimensions/window-width]]
```
flexsurfer: @acolytec3 and here we need to remove `window-width`

acolytec3: Done.

flexsurfer:
@acolytec3 would be super nice to move this function to utils and write a few unit tests for it, but it's up to you :)

acolytec3:
@flexsurfer So just replace the `formatted-value` component with a function that returns the truncated string set to how ever many units of precision?

flexsurfer:
but `formatted-value-item` function returns a string, so just move it to utils ns and rename probably, also we could parametrize 6, number of digits, smth like `format-decimals [value num]` , `(format-decimals 0.001234552, 6)`

acolytec3:
@flexsurfer Sure, happy to.

acolytec3:
@flexsurfer Done, rebased on `develop` and ready for testing again.

flexsurfer (approved):
Nice! Thanks!

acolytec3:
These don't feel like issues related to the code in this PR, right?

flexsurfer:
@Serhy  on it

Serhy:
Two e2e test failed occasionally (we have issues in GH which should fix both)
Manually tested with iPhoneX, Galaxy Note 4 (Android 6.0.1), Samsung J7 (Android 8.1)б - Network fee and Amount items on signing transaction views from DApp or Wallet not overlapped by value anymore!
Well there is exceptional case of huge unrealistic amounts (integer value) which could be entered manually only which I consider as very edge case we may skip:

Screenshot 2019-11-21 at 18 12 49


Thanks @acolytec3 !

acolytec3:
@Serhy Yep, this fix will only truncate values when the decimal places extend beyond 6 places.  My original PR would have taken the whole amount/fee into account when truncating but then you could have weird cases where the actual value is something like 123456789123.456 getting truncated to 1234567... which wouldn't be a good representation of the value.  

Glad it's working!

PR #9379, opened 2019-11-02
Use staging fleet and upgrade status-go

cammellos:
We need to use staging fleet as 1.9 is only running there.


rasom (approved):

cammellos:
@churik @Serhy are these tests failing on develop? I have verified 2,3,4 happen there as well, but not sure about 1. Thanks!

status-go-version.json
```diff
@@ -2,7 +2,7 @@
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "5f2544f1e8f2ee832bbab95c8d74e26ba0ab2481",
-    "commit-sha1": "5f2544f1e8f2ee832bbab95c8d74e26ba0ab2481",
-    "src-sha256": "0dlkr6056mhyny9k895zs5012ywiz75qm6y7a6hypb3pfxr35931"
+    "version": "develop",
```
jakubgs: Please don't use `develop` as a version for `status-go`. That will cause the version to appear in the App (and for the p2p peer) as `develop`, which is meaningless. It's fine for development, but it's not fine for merging into `develop` in `status-react`.

I will be releasing `0.34.0-beta.4` today, I'll ping you when I do, please use that.

jakubgs (changes requested):

Serhy:
> @churik @Serhy are these tests failing on develop? I have verified 2,3,4 happen there as well, but not sure about 1. Thanks!

Failed tests not related to this PR changes. 

1st. Failed due to too slow Network Fee data fetching (1 min was not enough for Network fee to be updated in Sign TX form). https://github.com/status-im/status-react/issues/9217 exists for that
2nd. Known issue https://github.com/status-im/status-react/issues/9372
3rd. Could not reproduce with real device with beta/staging fleets. I was about to disable this test btw, because it was already covered by other tests. 
4th. Failed as expected in this PR due to changed Fleet to non-beta one.

So only 3rd test I could say a false positive for this PR (though I doubt because it seem wallet balance was not updated according to the timing when transaction received)

status-go-version.json
cammellos: cool, let me know when ready

jakubgs:
There you go: [`v0.34.0-beta.4`](https://github.com/status-im/status-go/releases/tag/v0.34.0-beta.4)

cammellos:
@jakubgs updated

jakubgs (approved):

PR #9392, opened 2019-11-04
Dont use network info but only rely on peers count for mailservers

cammellos:
This code was still relaying on the OS network offline/online detection, the correct behavior is to solely rely on peers count for chat related activies.
status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -472,16 +472,6 @@
   [{:keys [db]}]
   {:db (dissoc db :mailserver/request-to)})
 
-(fx/defn network-connection-status-changed
-  "when host reconnects, reset request-to and
-  reconnect to mailserver"
-  [{:keys [db] :as cofx} is-connected?]
-  (when (and (multiaccounts.model/logged-in? cofx)
-             is-connected?)
-    (fx/merge cofx
-              (reset-request-to)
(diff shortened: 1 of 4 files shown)


cammellos:
@Serhy I was looking at the failed test, but looks like the message has been received on the second device, but still the test failed?  Am I missing something?

cammellos:
and "one more message" was never sent

churik:
@cammellos in this case https://app.saucelabs.com/tests/6d5f1368db544af6b08a098aafda9a14?auth=611754ba139cba944fbbd9cd7924510e#84 message was not sent from device1, so it wasn't received on device2.
Appium clicks on "Send message" button, but message wasn't sent.
The reason why it is not sent doesn't seem obvious to me, I tried to catch it on real device and on the emulator with builds from your PR - no success.
I relaunched failed tests, let's see.

yenda (approved):

cammellos:
@Serhy is this good to go?

Serhy:
Send message button could delay sometimes in emulator to appear when text message inserted in input. Or, as I see here and sometimes it may happen with another app parts, - when value was inserted in input field then click action upon actionable element does nothing and only second click may work (apparently the first click UNfocuses from input). 
So I added a 'try to click again' action here.

Ready for merge. 
Thanks a lot, @cammellos

PR #9395, opened 2019-11-05
don't use SHA1s as status-go version

jakubgs:
In order to avoid metrics that will contain commit SHA1s I'm enforcing use of `develop` as version when a commit SHA is used. This will lower the metric cardinality in Prometheus.

I also added a fix for CI to avoid appending `-universal` to all builds. It makes sense only for Android.

### DIFF
--- a/nix/status-go/default.nix
+++ b/nix/status-go/default.nix
@@ -12,8 +12,8 @@ let
   gomobile = callPackage ./gomobile { inherit (androidPkgs) platform-tools; inherit target-os xcodeWrapper utils buildGoPackage; };
   buildStatusGoDesktopLib = callPackage ./build-desktop-status-go.nix { inherit buildGoPackage go xcodeWrapper utils; };
   buildStatusGoMobileLib = callPackage ./build-mobile-status-go.nix { inherit buildGoPackage go gomobile xcodeWrapper utils; };
-  extractStatusGoConfig = callPackage ./extract-status-go-config.nix { inherit (stdenv) lib; };
   versionJSON = importJSON ../../status-go-version.json; # TODO: Simplify this path search with lib.locateDominatingFile
+  versionRegex = "^v?[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+(-[[:alnum:].]+)$";
   owner = versionJSON.owner;
   repo = versionJSON.repo;
   version = versionJSON.version;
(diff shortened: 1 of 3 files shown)


nix/status-go/default.nix
```diff
@@ -58,8 +58,10 @@ let
   goBuildFlags = "-v";
   # status-go params to be set at build time, important for About section and metrics
   goBuildParams = {
-    Version = removePrefix "v" version; # Geth forces a 'v' prefix
     GitCommit = rev;
+    Version = if builtins.match versionRegex version != null
```
adambabik: LGTM but does anyone ever need the actual value here? For example, to verify that an app is actually build with a proper status-go? Or it can be verified in some other way?

jakubgs: You mean if a dev wants to check the `status-go` version in the app, and all they will see is `develop`, instead of a commit? A fair point, but sounds pretty niche. Lets see what core devs and testers think about it.

jakubgs:
@yenda @flexsurfer @rasom @cammellos @Serhy @churik 
What do you think about `status-go` version showing up as `develop` in the app development builds. Specifically if the `status-go` version is specified as a commit SHA or a PR branch name. Would that be a problem?

The purpose of this is to not show commit SHAs in metrics we collect about connected peers, because that really increases cardinality.

nix/status-go/default.nix
pedropombeiro: Maybe the syntax would be clearer if we can add some parenthesis around the function call?

pedropombeiro (approved):

yenda:
@jakubgs so it concerns the status-go version right? are we still going to be able to distinguish between different nightly versions?

jakubgs:
@yenda yes, only `status-go` version, the one you see under "Node version" in the "About" page.
It does not affect anything related to the "App version".
Screenshot_20191105-110437
Ignore the `v` prefix before `develop`, Geth is adding that, it's retarded.

nix/status-go/default.nix
jakubgs: Added.

yenda (approved):

Serhy:
It's totally fine with me. I can't recall time when I needed hash value in node version (the `goX.Y.Z` in the tail was enough). 
Also OK to not have `-universal` suffix in build name values.

PR #9406, opened 2019-11-07
[#9350] fix back button on keycard login screen

### LINKED ISSUE #9350: Replace "Cancel" with "<" and remove extra screen if only one key

There is currently a `Cancel` button in the sign in flow. This should be replaced with a `<`. 

When there's only one key, the `<` button should be present from step 2 onward, so users re-enter the passcode or just come back for whatever reason.

Image:

back

Figma:

rasom:
fix #9350 

status: ready

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -482,6 +482,12 @@
  (fn [[macc acc]]
    (some #(when (= (:address %) (:address acc)) %) (:accounts macc))))
 
+(re-frame/reg-sub
+ :multiple-multiaccounts?
+ :<- [:multiaccounts/multiaccounts]
+ (fn [multiaccounts]
+   (> (count multiaccounts) 1)))
+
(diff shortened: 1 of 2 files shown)


src/status_im/subs.cljs
```diff
@@ -482,6 +482,12 @@
  (fn [[macc acc]]
    (some #(when (= (:address %) (:address acc)) %) (:accounts macc))))
 
+(re-frame/reg-sub
+ :multiaccounts-count
```
yenda: maybe it could be a predicate `multiple-multiaccounts?` here since that is the only usage of this sub

yenda (approved):

src/status_im/subs.cljs
rasom: ok, sounds reasonable

rasom: done

Issue #9429, opened 2019-11-11
Message counter reappears after relogin in chats 


Serhy:
# Bug Report

## Problem
After message received and read - message counter disappears as expected. However, after re-login to account, each message counts as "unread" and so unread message counter reappears each relogin in any type of chats.

#### Expected behavior
All read messages keep the 'read' status and no 'Unread' message counter in chat after contact relogin.

#### Actual behavior
<img width="754" alt="Screenshot 2019-11-11 at 15 43 44" src="https://user-images.githubusercontent.com/8749671/68591735-114daf80-049a-11ea-864d-b5b0e536aa65.png">

### Reproduction

- Create two new users: UserA and UserB
- UserA adds UserB and vice versa
- Both users join the same public chat
- UserA sends text message in 1-1 chat with UserB
- UserA sends text message in public chat
- UserB navigates to both chats to read messages
- UserB reopens Status and logins to account

### Additional Information
- Status version: develop `0.14.0 (2019111102)`
- Operating System: Android and iOS

cammellos:
Close via https://github.com/status-im/status-react/commit/13b01ce8795c8890cc5b733bef21dba65435ca41

PR #9434, opened 2019-11-12
Fix replies, clear-history, logout bug

### LINKED ISSUE #9427: Error "... function n(a0) {[bytecode]}" appears when message received in Group chat user not joined to yet
# Bug Report

## Problem
When user receives a message in Group chat he was invited to but not joined yet, - the `fx/defn expects a map of cofx as first argument got ' ' inn function function n(a0) { [bytecode] }` error appears. Canceling this error allows to proceed with app using but each new message raises error again

#### Expected behavior
No errors when when message received in Group chat user not joined to yet

#### Actual behavior

Screenshot 2019-11-11 at 15 09 36

#### Notes

Likely instroduced with https://github.com/status-im/status-react/pull/9410#issuecomment-551496370

### Reproduction

- Open Status and create new  accounts UserA and UserB
- UserA adds UserB in conntacts
- UserA creates Group Chat with UserB (UserB doe NOT accepts invite in this new group chat)
-  UserA sends text message(s)

### Additional Information

- Status version: develop `0.14.0 (201911112)`

### LINKED ISSUE #9430: Received messages disappeared in chats after re-login
# Bug Report

## Problem
Received messages in public/1-1/group chats disappear from respective chat view after re-login to account.
If I received message while I was online - those messages are kept upon next login BUT issue # https://github.com/status-im/status-react/issues/9429 takes place. And it needs 2-3 relogin attempts so those received messages are gone
But in a case of public chat like `#status` or `#status-core` it's enough to join chat, read messages (or not even read them so unread message counters are in place), - and upon next login there are no messages in chats.
Those messages user SENT - always present, but not those user received.
 
**NOTE**: the last message always kept displayed on main `Chats` view per chat if it was received by user NOT from offline inbox but while user was online 

#### Expected behavior
All messages received in chat kept in chats view after relogin

#### Actual behavior
Receive messages in 1-1/public/group chat -> reopen Status and login -> no messages are present in chat view

### Reproduction

- Join `#status` and/or `#status` public chat 
- Read messages received from offline inbox
- Re-login to account
- Open chats again and verify messages are present

### Additional Information
- Status version: develop 0.14.0 (2019111102)

### LINKED ISSUE #9431: Clearing history doesn't clear day swimlines in chat and messages reappear after log out
# Bug Report

## Problem

"Clear history" leaves days in chat and if you log out through Profile > log out and login again - messages will reappear in chat. 

The issue is not reproducible if you reopen app, only with relogin through profile.

#### Expected behavior
Chat is empty

#### Actual behavior
Just after clearing:
photo_2019-11-11 16 11 25
After relogin messages reappear.

### Reproduction

- Open Status
- Join any public chat with messages
- Tap on ... > Clear history
- Go to profile
- Log out
- Login again and open the same chat

### Additional Information

- Status version: nightly 11/11/2019

#### Logs
Full logcat and video: https://app.saucelabs.com/tests/cce1309ff3e943c4b8f75ed79083493f?auth=a0a9f840f1e7e22af5d2a83204b94b33#87

### LINKED ISSUE #9433: No replies on message receiver chat view
# Bug Report

## Description
There are no replies on message receiver side in 1-1/public/group chats

#### Expected behavior
Replies are present for message receiver
#### Actual behavior

Screenshot 2019-11-11 at 18 55 58

### Reproduction

- Open Status and create UserA and UserB
- UserA and UserB join the same public chat
- UserA sennds text message in public chat
- UserB long-tap on message -> `Reply`
- UserB sends text message with quote
- UserA verifies there is a reply present

### Additional Information
- Status version: develop `0.14.0 (2019111102)`

cammellos:
Fixes #9433 : this was due to the fact that `.-response-to` was
returning nil, because of the dash in the name, instead in this cases
`(aget .. "response-to")` should be used.

Fixes #9431 : This was a left-over from the move from message-groups to
message-list, and the code was not updated.

Fixes #9430 #9429 Both of these were due to the same issue, cofx were
wrongly passed to the function resulting in the db being updated but the
fxs being discarded.

There's still a separate issue that might result in messages not being
saved on logout, because of a race condition (if you logout while is
fetching messages, some of the message might not be saved). I will
address that separately as we might be able to just save messages as
they come in status-go, rather then having to pass them to status-react
and back to status-go for saving.

Might fix #9427 as well

status: ready

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -151,7 +151,7 @@
      cofx
      {:db            (update-in db [:chats chat-id] merge
                                 {:messages                  {}
-                                 :message-groups            {}
+                                 :message-list              nil
                                  :last-message-content      nil
                                  :last-message-content-type nil
                                  :last-message-timestamp    nil
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -292,27 +292,18 @@
               (send chat-id message wrapped-record)
               (update-message-status chat-id message-id :sending))))
 
-(fx/defn remove-message-from-group
-  [{:keys [db]} chat-id {:keys [timestamp message-id]}]
-  (let [datemark (time/day-relative timestamp)]
-    {:db (update-in db [:chats chat-id :message-groups]
-                    (fn [groups]
-                      (let [message-references (get groups datemark)]
-                        (if (= 1 (count message-references))
-                          ;; message removed is the only one in group, remove whole group
-                          (dissoc groups datemark)
-                          ;; remove message from `message-references` list
-                          (assoc groups datemark
-                                 (remove (comp (partial = message-id) :message-id)
-                                         message-references))))))}))
+(fx/defn rebuild-message-list
+  [{:keys [db]} chat-id]
+  {:db (assoc-in db [:chats chat-id :message-list]
+                 (message-list/add-many nil (vals (get-in db [:chats chat-id :messages]))))})
 
 (fx/defn delete-message
-  "Deletes chat message, along its occurence in all references, like `:message-groups`"
+  "Deletes chat message, rebuild message-list"
   [{:keys [db] :as cofx} chat-id message-id]
   (fx/merge cofx
             {:db            (update-in db [:chats chat-id :messages] dissoc message-id)}
             (messages-store/delete-message message-id)
-            (remove-message-from-group chat-id (get-in db [:chats chat-id :messages message-id]))))
+            (rebuild-message-list chat-id)))
 
 (fx/defn add-system-messages [cofx messages]
   (let [messages-fx (map #(add-message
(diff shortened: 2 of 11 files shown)


cammellos:
@Serhy @churik would you mind giving it a spin to check the bugs have been fixed?
Also the failed ENS test seems like it actually worked, likely the issue is that names are resolved async. for performance reason (so you will see initially the 3 random words name, once it resolves it it will show autotester, as in screenshot), so maybe is worth checking the username with a timeout, and fail if it cross the threshold.
Thanks!

churik:
OFC! on it

churik:
Tested all issues mentioned in description + basic group chat testing as regression.
So fixed issues: #9433 #9431 #9430 #9429  #9427

> There's still a separate issue that might result in messages not being
saved on logout, because of a race condition (if you logout while is
fetching messages, some of the message might not be saved). I will
address that separately as we might be able to just save messages as
they come in status-go, rather then having to pass them to status-react
and back to status-go for saving.

couldn't catch it manually.

> Also the failed ENS test seems like it actually worked, likely the issue is that names are resolved async. for performance reason (so you will see initially the 3 random words name, once it resolves it it will show autotester, as in screenshot), so maybe is worth checking the username with a timeout, and fail if it cross the threshold.

Yes, this is the reason, I'll push a fix separately.
Thank you for such amazing work!

rasom (approved):

yenda (approved):

flexsurfer (approved):

PR #9438, opened 2019-11-12
Calls to gfycat and identicon made async in few places

### LINKED ISSUE #9097: Missed avatar and name on account recover, desktop
# Problem

Avatar and name not shown when recover account 

Screen Shot 2019-10-02 at 7 03 06 PM

## Implementation

TBD

## Acceptance Criteria

Avatar and name should be shown

vkjr:
fixes #9097

### Summary

Existing problem: functions that generate random name and identicon are implemented in `status-go` and made synchronous. And `react-native-desktop` doesn't support sync calls (I spent some time on implementing it and can't do it in a reasonable time).
So in this PR I changed synchronous name and identicon generation in the few places:
- on account recover
- on new chat adding
- in public chats 

During implementation I spent a lot of time on understanding how Status works internally and wasn't ready to remove all synchronous calls at once, so this is proof-of-concept and I'd like to hear feedback on it.

### Review notes
I noticed a bit strange behavior in current implementation. When you starting chat with a new person it is added to `:contacts/contacts` but after restart info about this contact disappears unless you intentionally pressed "add contact" on it. At the same time, if you open a profile of this person, `create-contact` function calls and since then it persistently lives in `:contacts/contacts`. As far as I understand this happens because tribute-to-talk introduced changes into how `:contacts/contacts` treated.
So to overcome contacts removing I added a call of `create-contact` to `preload-chat-data` function that calls on new contact adding. In this way I prevent contacts regeneration and need of sync calls there. I'm not sure if this is correct so would be glad to hear your feedback.

#### Platforms
- Android
- iOS
- macOS
- Linux
- Windows

Displaying of avatars and names

##### Functional

- 1-1 chats
- public chats
- group chats
- account recovery
- new account
- user profile updates

### Tested on mac and ios:
- contact can be recovered
- 1-to-1 chat can be created
- can join to public chat
- message can be sent successfully

status: ready


vkjr:
@flexsurfer, @siphiuel, could you please take a look at this PR?

flexsurfer:
@vkjr i'm not sure why we do this for mobile app as well? @cammellos could you take a look?

vkjr:
@flexsurfer, I just thought that maintaining two different logics in the code would be a bit tedious and it would be nicer to have a single one for both mobile and desktop :)

yenda:
One advantage of removing these two calls is that they are not working with chrome remote debugging tools. So I agree on principle but I haven't analyzed the consequences.

cammellos:
Yes, if we could get async for mobile probably best, as we'd rather have that

src/status_im/contact/db.cljs
```diff
@@ -120,6 +120,10 @@
                  (assoc % :admin? true)
                  %)))))
 
+(defn contact-exists?
+  [db public-key]
+  (some? (get-in db [:contacts/contacts public-key])))
```
cammellos: I don't think you need to use `some?` here, would work the same without if used for truthy, falsey

src/status_im/utils/gfycat/core.cljs
```diff
@@ -12,3 +13,14 @@
     (native-module/generate-gfycat public-key)))
 
 (def generate-gfy (memoize build-gfy))
+
+(re-frame/reg-fx
+ :insert-gfycat-fx
+ (fn [[public-key path]]
+   (native-module/generate-gfycat-async public-key #(re-frame/dispatch [:set-in path %]))))
```
cammellos: Unless necessary, we generally avoid using `set-in` , but rather have a well defined event for this (i.e `gfycat-generated`), same for identicon

cammellos (changes requested):
I would probably use async methods for both mobile/desktop, so that we are consistent, otherwise looks ok

churik:
@vkjr 
Not able to join group chat  - the reason for failure of the most of the test cases.
After receiving invite to group chat "Join to the chat" button is not shown.
Logs and video: https://app.saucelabs.com/tests/adf0d4cf17c044d0aba48e7212270354?auth=47473cab59523e476d0b38453b05ac2f#66

vkjr:
@churik, @cammelos thanks, will fix

src/status_im/native_module/core.cljs
```diff
@@ -269,7 +269,26 @@
   {:pre [(utils.db/valid-public-key? public-key)]}
   (.generateAlias (status) public-key))
 
+(defn generate-gfycat-async
+  "Generate a 3 words random name based on the user public-key, asynchronously.
+  It is really asynchronous only on desktop because it doesn't support sync calls"
```
yenda: seems like it is not true anymore it is also async on mobile?

src/status_im/utils/gfycat/core.cljs
```diff
@@ -12,3 +13,14 @@
     (native-module/generate-gfycat public-key)))
 
 (def generate-gfy (memoize build-gfy))
+
+(re-frame/reg-fx
+ :insert-gfycat-fx
```
yenda: no need for -fx prefix, should be just called generate-gfycat

src/status_im/utils/gfycat/core.cljs
```diff
@@ -12,3 +13,14 @@
...
+ (fn [[public-key path]]
+   (native-module/generate-gfycat-async public-key #(re-frame/dispatch [:set-generated-gfycat path %]))))
```
yenda: call the event something like "gfycat-generated" or "generate-gfycat-callback", something that happened not an action name.

src/status_im/utils/identicon.cljs
```diff
@@ -1,5 +1,13 @@
 (ns status-im.utils.identicon
   (:require
+   [re-frame.core :as re-frame]
    [status-im.native-module.core :as native-module]))
 
 (def identicon (memoize native-module/identicon))
+
+(def identicon-async (memoize native-module/identicon-async))
```
yenda: I don't think it makes sense to memoize an async call

churik:
@vkjr please, remove `Tested - Issues` label and ping me when all comments and e2e failures will be addressed.

vkjr:
@yenda, @cammellos, your review notes fixed, please approve if there are no new ones :)

@churik, e2e tests pass except those that already fixed in `develop`. PR rebased.

yenda:
@vkjr can you remove the sync calls?

vkjr:
@yenda, I won't be able to remove all sync calls in this PR. It was started to remove those that break desktop functionality. But now desktop is postponed and I'm switching to more priority UI issues. But still, it would be nice to merge this PR to save progress and other sync calls can be removed later.

vkjr:
@yenda, @flexsurfer, @siphiuel, @cammellos summoning you again :)

src/status_im/multiaccounts/recover/core.cljs
```diff
@@ -101,9 +101,15 @@
           constants/path-whisper
           constants/path-default-wallet]
          (fn [result]
-           (let [derived-data (types/json->clj result)]
-             (re-frame/dispatch [::import-multiaccount-success
-                                 root-data derived-data])))))))))
+           (let [derived-data (types/json->clj result)
+                 publicKey (get-in derived-data [constants/path-whisper-keyword :publicKey])]
```
cammellos: we use `kebab-case` rather than `camelCase`

vkjr: Fixed

vkjr:
@yenda, @flexsurfer, could you please approve if you don't have other comments.

cammellos (approved):

churik:
@vkjr does it require manual testing?

vkjr:
@churik, yes, it would be nice to test it.

vkjr:
Tested on mac and ios:
- contact can be recovered
- 1-to-1 chat can be created
- can join to public chat
- message can be sent successfully
- i can see messages of others
- profile can be opened, avatar and name shown correctly

Issue #9229, opened 2019-10-16
Clean keycard and remove multi-account


guylouis:
**Situation** 

Currently, after a keycard has been set-up after onboarding the keycard has 
- setup a PIN, Pairing code, PUK
- secrets loaded (which correspond to status multiaccount seed)

**Problem** 

There is no possibility for the user to remove the secrets loaded physically from the card

A user might want to delete the secrets from his card because
- despite the high level of hardware security of his card, he wants to remove them physically from there to prevent any attack to succeed now or later 
- he wants to change the multiaccount on his card

**Solution** 

A simple approach would be 
- user is logged in his keycard multiaccount
- user goes to 'settings/keycard'
- user choses 'reset keycard' (to be added, does not exist today)
- user is prompted 
"You're about to delete the keys on this card, your pairing code/PUK & PIN remain will unchanged.
You must have your seed if you want to recover any funds from this key in the future. You can now restore another key on this card. Your Pairing code remains unchanged and will be needed if you use a new phone. 
Are you sure you want to proceed ? yes/no"
- if the user clicks 'yes' then 
            - we leaves PIN, PUK, Pairing uncode untouched on the card 
            - the secret on the card is deleted
            - the multiaccount is removed from the phone

3esmit:
If user forgets Pairing code, PUK and/or PIN, it should be able to reset?

guylouis:
given the low value ($) of the card it self we've not designed it this way ...
the user must remember his PIN (or PUK) to access his card, and he must remember his pairing if he has no phone paired

If we implement https://github.com/status-im/status-react/issues/9064 after v1, the user will be able to recover any keycard if he has his seed (and if PUK/Pairing hasn't been changed form the derived version of the seed by another app...)

rasom:
[Keycard's side of this task is already implemented](https://github.com/status-im/status-react/blob/develop/src/status_im/ui/screens/hardwallet/settings/views.cljs#L127), so we need to make sure that there is an API for multiacc removal.

guylouis:
Ok
Decided with @Rachel Hamlin <rachel@status.im> that resetting a keycard and
removing its account is post v1.0

cammellos:
@errorists @hesterbruikman could you please have a look at the UI for this?
Thanks

guylouis:
would be great to progress on this 
 @johnlea-quiup 
wdyt about the UX ?

hesterbruikman:
[...]

PR #9439, opened 2019-11-13
[#9229] make sure reset card functionality works on cljs side

rasom:
status: wip


scripts/update-status-go.sh
```diff
@@ -28,7 +28,7 @@ if [ $# -eq 0 ]; then
 fi
 
 STATUS_GO_REPO="${STATUS_GO_REPO:=status-go}"
-STATUS_GO_OWNER="${STATUS_GO_OWNER:=status-im}"
+STATUS_GO_OWNER="${STATUS_GO_OWNER:=rasom}"
```
jakubgs, line 31: Please don't merge this PR with this change.

status-go-version.json
```diff
@@ -1,8 +1,8 @@
 {
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
-    "owner": "status-im",
+    "owner": "rasom",
```
jakubgs, line 3: Or this one either.

scripts/update-status-go.sh
rasom: I was not going to merge it that way, also it can't be built at the moment

PR #9440, opened 2019-11-13
upgrade multiple desktop deps

jakubgs:
Combines package upgrade for Desktop from PRs #9415 to #9424.


vkjr (approved):
Works correctly with latest desktop PR - #9438

PR #9441, opened 2019-11-13
[#9345] fix some issues on keycard login screen

### LINKED ISSUE #9345: Keycard sign in – "More" button behavior is wrong + breadcrumbs must be dropped

1. On the sign in screen, the **"..."** in the top right corner should open a drop-down with one item only "Get a new key"

Image:
get-a-new-key

Figma:

2. The breadcrumbs that read "Step 1 of 2" should be removed to match the non-Keycard flow.

rasom:
fix #9345 

status: ready

### DIFF
--- a/src/status_im/hardwallet/core.cljs
+++ b/src/status_im/hardwallet/core.cljs
@@ -20,7 +20,8 @@
             status-im.hardwallet.fx
             [status-im.ui.components.react :as react]
             [status-im.constants :as constants]
-            [status-im.multiaccounts.update.core :as multiaccounts.update]))
+            [status-im.multiaccounts.update.core :as multiaccounts.update]
+            [status-im.ui.components.bottom-sheet.core :as bottom-sheet]))
 
 (def default-pin "000000")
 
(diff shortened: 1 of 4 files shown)


yenda (approved):

churik:
@rasom thank you for tackling it.

**1) "Create new key" instead of "Get a new key"** 
In https://github.com/status-im/status-react/issues/9345 I see that in the sign in screen, the "..." in the top right corner should open a drop-down with one item only "Get a new key".
Currently, it opens a screen with "Create new key"
photo_2019-11-15 15 17 40

**2) "Create new key" opens "Store your key on keycard" screen.**
But it may happen that user wants to generate account without keycard, so in this case I expect the same behavior as for "Generate new key":

photo_2019-11-15 15 08 42

**3) Tapping back after tapping on "Create new key" redirects to carousel view instead of login screen**

Steps: 
- create an account with Keycard
- log out
- tap on options > "Create new key"
- tap back

Expected result: login screen
Actual result: 
photo_2019-11-15 15 18 08


@andmironov please fix if smth from above is not expected, thanks!

rasom:
@churik issues should be fixed now

rasom:
@churik hold on, it seems that a new account still can't be created and we probably need to open a different screen when go to "Access key", because there is now way to restore an acc from mnemonic there at the moment.

rasom:
@churik should be fixed now

churik:
Retested failed e2e - all failures are related to balance fetching problems, so they are reproducible from time to time.

PR #9443, opened 2019-11-13
Allow use of local status-go repository in Nix build/shell

pedropombeiro:

### Summary

This PR adds support for a `STATUS_GO_SRC_OVERRIDE` environment variable that if non-empty will point to a local status-go repository, and be used instead of the version specified in `status-go-version.json`. This makes for much faster iteration cycles when testing status-go changes.

Usage:

```shell
make release-android STATUS_GO_SRC_OVERRIDE=$GOPATH/src/github.com/status-im/status-go
```

or

```
export STATUS_GO_SRC_OVERRIDE=$GOPATH/src/github.com/status-im/status-go
make release-android
```

Documentation change PR (review please): https://github.com/status-im/status.im/pull/422

### Testing notes

This doesn't need to be tested since it only impacts the dev environment if the `STATUS_GO_SRC_OVERRIDE` env var is defined.

status: ready


nix/status-go/default.nix
```diff
@@ -1,34 +1,50 @@
...
+  srcData =
+    # If config.status_go.src_override is defined, instruct Nix to use that path to build status-go
+    if (attrByPath ["status_go" "src_override"] "" config) != "" then rec {
+        owner = "status-im";
+        repo = "status-go";
+        version = "local";
+        goPackagePath = "github.com/${owner}/${repo}";
+        rev = "unknown";
+        shortRev = "unknown";
+        versionName = "local";
```
jakubgs: Why use `unknown` or `local`? We are already using `develop` for development versions, use that.

jakubgs (approved):
Nice work. But maybe add a sub-section in `nix/README.md` about this?

nix/status-go/default.nix
pedropombeiro: What if the user is on a different branch? It felt weird using `develop` if my branch is named otherwise, that's why I chose a completely different hardcoded name. WDYT?

pedropombeiro:
> But maybe add a sub-section in nix/README.md about this?

Done

nix/status-go/default.nix
jakubgs: Well, that just means in metrics we'll have two versions: `vunknown` and `vdevelop`.

In both cases the exact commit or branch is unknown, but they appear separately, is that a good thing?

pedropombeiro: why would the `v` prefix be there? I think it might actually be a good thing to know that local test builds are not taken into account in `develop`. Or not?

jakubgs: Because Geth code adds the `v` because it's retarded.

But `develop` also includes local builds, just ones that were built using commit hash instead of a custom `status-go` path, both are local, both are development builds. What's the difference?

pedropombeiro: > Because Geth code adds the v because it's retarded.

Isn't that what you're removing with `then removePrefix "v" version # Geth forces a 'v' prefix`? So it looks like whatever we assign to `versionName` will be used as-is.

In any case, I don't care much either way, so I'll just change `versionName` to `develop`.

jakubgs: No, that part removes `v` in case version contains `v`, so we don't get `vvSomething`.

PR #9449, opened 2019-11-15
erc20 validation in different builds

flexsurfer:
ERC 20 tokens validation implemented in status-react, and executed each time after sign in, this is a huge task which makes UI hanging, so I keep it only in e2e builds,

@churik @Serhy maybe it's better to write a script and test it separately in e2e , we don't want to do it in the app for sure


jakubgs (approved):

Serhy:
@flexsurfer for me it looks like more unit test than e2e: we validate each token from list of our default token list. 
Btw, when it may even happen validation failed?

Serhy:
We seem used it in order to validate tokens on a time when we didn't have 'Add custom token' feature and each tokens required separate issue to be added in Status app. 
Yes it's definitely what we could do outside the app.

Serhy (approved):

PR #9450, opened 2019-11-15
[#9446] Resume logging when connection to card is restored

### LINKED ISSUE #9446: Connection with card is not restored if it was lost during login
# Bug Report

## Problem
After interrupting connection with the card during log in it is not restoring after tapping card again, so you need to enter PIN again.
#### Expected behavior
connection is restored

#### Actual behavior
endless spinner 
photo_2019-11-14 11 01 30

### Reproduction

- Open Status
- Create account with KeyCard
- Log out
- enter PIN to login and put card off from the phone to interrupt connection
- tap card back again

### Additional Information

- Status version: nightly 14/11/2019
- Operating System: Android

rasom:
fix #9446 

Currently there are two ways to initiate logging in with keycard:
- enter PIN code and after that connect card to the phone
- connect card to the phone and enter PIN after that

Before this commit in both cases when connection to the keyacrd was lost
and then restored, logging in didn't resume. In result a user saw a
pop-up with endless spinner.

The reason of this bug was that `:on-card-connected` and `:on-card-read`
actions were not restored in app-db after losing connection to the card.

This commit introduces helper functions for both `:on-card-connected` and
`:on-card-read` which allow to reset these values and stash them until
particular flow of calls to keycard will be finished. In case if
connection was lost before the flow is finished the valueas are restored
so that it can be succesfully resumed on th next connection.

Also a banch of log entries were added to simplify debugging of
interactions with keycard and native module.

status: ready

### DIFF
--- a/src/status_im/native_module/core.cljs
+++ b/src/status_im/native_module/core.cljs
@@ -5,7 +5,8 @@
             [status-im.react-native.js-dependencies :as rn-dependencies]
             [status-im.ui.components.react :as react]
             [status-im.utils.platform :as platform]
-            [status-im.utils.types :as types]))
+            [status-im.utils.types :as types]
+            [taoensso.timbre :as log]))
 
 (defn status []
   (when (exists? (.-NativeModules rn-dependencies/react-native))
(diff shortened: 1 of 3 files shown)


vitvly (approved):

churik:
Tested lost connection to the card during:
- multiaccount creation
- enter PIN code and after that connect card to the phone
- connect card to the phone and enter PIN after that
- canceling login when the connection is lost
Looks good to me.


Device: LG V20
OS: Android 8

src/status_im/native_module/core.cljs
```diff
@@ -14,46 +15,56 @@
 (def adjust-resize 16)
 
 (defn clear-web-data []
+  (log/debug "[native-module] clear-web-data")
```
flexsurfer, line 18: do you want to merge all these logs ?

PR #9454, opened 2019-11-15
nix: Remove need to pass default values in `--arg config`

pedropombeiro:
This PR refactors the Nix bootstrapping so that invokations don't need to pass default `config` fields (such as `android_sdk.accept_license = true;`), but can still override them if needed.


jakubgs (approved):
NICE

PR #9455, opened 2019-11-16
[WIP] MVP Images

flexsurfer:
MVP Images [WIP] Take a picture is not implemented yet

Upload an image to infura and send a hash

DEMO https://www.youtube.com/watch?v=6Vhv-j6n8Po

| Choose | Upload | Send |
| ------------- | ------------- | ------------- |
| image | image | image |

### DIFF
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -289,5 +289,8 @@
                 [react/image {:style  {:margin 10 :width 140 :height 140}
                               ;;TODO (perf) move to event
                               :source {:uri (contenthash/url (-> content :sticker :hash))}}]
-                [message-bubble-wrapper message
-                 [react/text (str "Unhandled content-type " content-type)]]))))]])))
+                (if (= content-type constants/content-type-image)
+                  [react/image {:style {:margin-vertical 10 :width 140 :height 140 :border-radius 8}
+                                :source {:uri (str "https://ipfs.infura.io/ipfs/" (:hash content))}}]
+                  [message-bubble-wrapper message
+                   [react/text (str "Unhandled content-type " content-type)]])))))]])))
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -20,7 +20,9 @@
             [status-im.utils.gfycat.core :as gfycat]
             [status-im.utils.platform :as platform]
             [status-im.utils.utils :as utils]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [status-im.utils.image-processing :as image-processing]
+            [status-im.ipfs.core :as ipfs]))
 
 (defn- get-chat [cofx chat-id]
   (get-in cofx [:db :chats chat-id]))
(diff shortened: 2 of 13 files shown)


errorists:
👏A👏W👏E👏S👏O👏M👏E👏

I treat this as a proof of concept for the purpose of UI correctness, @flexsurfer please let me know if you'd like me to point out any UI styling issues, there's only a handful.

However, one huge UX request : please allow images to be previewed and send with their correct aspect ratio preserved and not cropped in a square. I think this needs to be addressed before I feel comfortable shipping this MVP.

this is as good illustration as any of what I mean
Photos -_ Camera access

I guess it can be left outside the scope for now but the first immediate thing to address would be to add full screen image previews once tapped on an image in chat

Photos -_ Camera Preview

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -195,6 +195,12 @@
    [react/image {:style {:margin 10 :width 140 :height 140}
                  :source {:uri (contenthash/url (:hash content))}}]])
 
+(defmethod message-content constants/content-type-image
+  [wrapper {:keys [content] :as message}]
+  [wrapper message
+   [react/image {:style {:margin-vertical 10 :width 140 :height 140 :border-radius 8}
+                 :source {:uri (str "https://ipfs.infura.io/ipfs/" (:hash content))}}]])
```
cammellos: I would not impose IPFS in the URL, we can still validate the link client side if we want to have only IPFS domains, but otherwise changing this would be breaking. Rather I would pass the full url, and then we can only show if `starts-with` ipfs or something.
Also markdown supports images, so it's possible to embed images in text content if we'd like.

errorists: @cammellos you mean unfurl a link that leads to an image? or place an inline image next to regular text? I would say it's not something we want, images should always be placed separately as their own message bubble in chat.

cammellos: Yes, that's what I meant (regarding markdown, not the URL, that's a protocol thing), thanks for the reply.

flexsurfer: this is a temporary implementation, there will be a multihash instead ipfs hash, we don't want to send any web2 links or images,  multihash might be used for ipfs or swarm or our own storage later

rachelhamlin:
Blocking us from shipping:

- Image doesn't appear right now because a status-go update is required
- Are images maintaining their aspect ratio now?
- If I add text after choosing an image, the image is then removed and only the text is sent. Is it possible to send both at once? 1. Select an image; 2. Add text; 3. Send both at once. 
- Take image flow is not complete. Tapping on the camera icon to reverse the camera will lead to 1) `Appearance` settings, and 2) take a photo and make it your profile picture. 😂 Nice workaround for custom photos! Now I can't delete it.

errorists:
> If I add text after choosing an image, the image is then removed and only the text is sent. Is it possible to send both at once? 1. Select an image; 2. Add text; 3. Send both at once.

yes, that's something made possible with the 'Final' (needs better naming) design. I - maybe - could make it work with MVP but it wouldn't be pretty

rachelhamlin:
@errorists is it better to disable the text field when an image is inserted? User can still 'x' it out. But if they type, they remove their image. It wasn't 100% clear to me that the image weren't still being sent along w/ my text, until I sent it.

Just a thought, not a deal breaker for me though.

errorists:
Probably not, I mean when you focus the input, you cancel out of the image sending dialog. to bring it back you need to tap the image icon again, so it's either one or the other. 

It's the MVP flavour of this design for which I apologise but in my excuse there's a better, fully formed one waiting there

errorists:
since it doesn't hurt to ask, @Ferossgp would you care to estimate how much longer it would take you to add the UI features in the [full design](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=4175%3A23237)? I mean the image composition menu that has a horizontally scrollable gallery built in and a camera viewfinder? and the image queue above the input that allows for queuing multiple images, pasting images and adding text / composing replies to an image?

Ferossgp:
@errorists UI components are not complex, but given the fact that we should interact with some native library like [Camera Roll](https://github.com/react-native-community/react-native-cameraroll) and camera preview, I would expect complexity and possible pitfalls on that. It also depends on how multiple images would be handled. If there will be no limitation on one image in MVP then it should not be hard (Meaning that message with image structure will not limit only one URL). I'm not sure at this time how to better implement paste image cause I've never worked with this in RN. I would allocate 2/3 days without paste image

rachelhamlin:
@errorists no need to apologize, I think it works just fine for now. 2/3 days ain't bad for the bulk of the remainder, but for the MVP let's just ensure these user requirements are met:

- Aspect ratio is maintained for images in chat
- Take image flow is working

cc @flexsurfer

hesterbruikman:
@flexsurfer should we reopen this in light of https://github.com/status-im/specs/issues/69#issuecomment-621749551

PR #9468, opened 2019-11-18
[WIP] Wallet next level

flexsurfer:
This PR moves the wallet to the next level

3 new feature

- [x] adding a watch-only account
- [ ] importing a seed account
- [ ] importing a private key account

1 change for generating accounts, new screen with derivation path

User can change here only last part of the path, or use the default. Before it was hidden from the user, and was kinda magic, and it's not clear how to recover this account later in status or other wallets, now user can choose which one he wants and understand clearly what wallet with what path will be generated 

@andmironov @errorists would be great to have a design for this screen, and also we need some description

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -185,7 +185,7 @@
 (reg-root-key-sub :intro-wizard-state :intro-wizard)
 
 (reg-root-key-sub :popover/popover :popover/popover)
-(reg-root-key-sub :generate-account :generate-account)
+(reg-root-key-sub :add-account :add-account)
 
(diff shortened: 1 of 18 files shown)


flexsurfer:
Watch-only accounts and new derivation screens can be reviewed in PR builds

andmironov:
If I understand correctly, you want to replace the flow below with this one screen which I can't approve :-) 

I totally get the intention of adding this feature to v1 but I'm concerned about the fact that every user would see this mysterious "Derivation path" screen which I believe is confusing

Screen Shot 2019-11-18 at 13 51 59

PR #9475, opened 2019-11-18
fix 9394 select mailserver based on ping

### LINKED ISSUE #9394: Select mailservers based on ping
### User Story

Currently the mailservers from the fleet are selected randomly, which results in me getting connected to the Hong Kong mailservers almost every time when I'm in Europe and should be getting the european ones.

### Description

This results in a really bad experience for users with slow sync times and bad latency.

### Acceptance Criteria

The simplest solution would be to just do an ICMP ping to a few mailserver, group them into response time buckets, and pick a random one from the fastest bucket.

### Notes

Example ping output:
```
 $ for DC in "do-ams3" "gc-us-central1-a" "ac-cn-hongkong-c"; do \
     ping -c1 mail-01.${DC}.eth.beta.statusim.net | grep 'time='; \
   done 
64 bytes from 206.189.243.162 (206.189.243.162): icmp_seq=1 ttl=55 time=29.1 ms
64 bytes from 210.19.188.35.bc.googleusercontent.com (35.188.19.210): icmp_seq=1 ttl=50 time=127 ms
64 bytes from 47.91.156.93 (47.91.156.93): icmp_seq=1 ttl=47 time=279 ms
```
As you can see for me from Poland the ranking of access time is simple:

1. Amsterdam - `29.1 ms`
2. United States - `127 ms`
3. Hong Kong - `279 ms`

yenda:
fix #9394 

- if not mailserver was actively selected by user,
use rpc call to get latency for known mailservers
and use the best one
- this happens when `set-current-mailserver` is called which happens
in `change-mailserver` when user unpins his preferred mailserver and when
there's been too many failed attemps to fetch messages or to connect to
then current mailserverm as well as when user logs in.

## To test

- better performance of mailservers (HongKong should never be randomly picked from Europe)
- messages are fetched


jakubgs (commented):
I can't really read Clojure but three things:

* Are you picking only the fastest mailsever? That might lead to some dogpiling. I was thinking we should pick the N fastest and pick randomly from those to avoid dogpiling.
* I can't find you checking the value of `error` field, do you ignore it?
* Have you seen the [design done by Maciej](https://github.com/status-im/status-react/issues/9394#issuecomment-555375407)? No rush though, could be done in separate PR.

And thanks for taking a look at this!

jakubgs:
Oh, and one more note, this call currently does not work on Windows. Not sure what will happen there.

yenda:
@jakubgs 
- ok I can pick between the 3 fastest?
- yes I ignore it but actually there is a problem because the error ping is `-1` could you just not return it or return `null` so I don't have to exclude the `-1` when sorting
- yes seen the design but it should be a separate PR for after v1 we don't have time for that now

jakubgs:
@yenda selecting randomly from 3 fastest sounds reasonable.

And regarding returning `null` instead of `-1`, sure, here's the PR: https://github.com/status-im/status-go/pull/1701
Also renamed `rtt_ms` to `rttMs` to not mix CamelCase with sneak_case.

yenda:
@jakubgs the issue with selecting one of the best 3 is that on some fleet there is only 3
mailservers

yenda:
@jakubgs ok I used a pool of 1/4 of the fleet size with lowest latency mailservers to pick from and updated based on your casing change

status-go-version.json
```diff
@@ -2,7 +2,7 @@
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.34.0-beta.8",
-    "commit-sha1": "9d7c570593b1f88e9688204d8e1041d57b98da1f",
-    "src-sha256": "1kwxf0h80vdj493c7s3lpmdjpbc72plbll0rj1vv7kzf8a84ff2k"
+    "version": "develop",
+    "commit-sha1": "464c30a43019fe9b774036f685f99c54fc22b882",
+    "src-sha256": "0p6qryhszs48qxkb3h2j1p0wip3ii9928r34yi0z1vfakwk4056f"
```
jakubgs: Please use the `v0.35.1` tag instead of `develop`.

jakubgs (approved):
NICE!

status-go-version.json
yenda: you don't need to update the version file is status-go?

cammellos (approved):

status-go-version.json
jakubgs: Well, technically, yes, in practice it has no effect, but you're right I should do it.
Try now, I updated the tag `v0.35.1` tag, it should point at https://github.com/status-im/status-go/commit/4769fb91c3247c72c21013bf36194843d8d72d18.

jakubgs:
@yenda I have merged #9531 which upgrades `nixpkgs` and in the process fixes iOS builds.
Please rebase and make sure you use the `v0.35.1` tag for `status-go`.

churik:
@yenda it should be tested or you will test it by yourself?

Also can I ask you to take a look at failed e2e `test_fetching_balance_after_offline`?

The very strange error that I see for the first time, I don't think it is related to your PR, but at least worth to take a look.

'Index out of bounds' after the account was restored.
Full logcat: 
[logcat.log](https://github.com/status-im/status-react/files/3900974/logcat.log)
Steps+video: https://saucelabs.com/jobs/9cec83b7db5246838041a8dc90755f59?auth=70694c27140aba2b2127e1b832ac1502

Couldn't reproduce issue manually.

yenda:
@churik yes I tested, I think there was a bug that I fixed after the e2e test that could have cause that error under some wierd circumstances. Can you confirm that the current failing test are unrelated?

churik:
@yenda yes, current  failing tests are unrelated.
Fix is in progress in https://github.com/status-im/status-react/pull/9544

PR #9476, opened 2019-11-18
Fix qr navigation bug in qr scanning

### LINKED ISSUE #9381: Chat's QRCode scanner dont exits after scan
# Bug Report

## Problem

When a qrcode is scanned it brings to result screen, but it is kept open in background, when tapping back it shows the camera open, however it wont work for scanning qr codes anymore.

#### Expected behavior

After scan it should close camera and scanner. 

#### Actual behavior

After scan camera is kept open in background 

### Reproduction

1. Open Status
2. Open QR code scanner from Chat Tab
3. a) scan a contact code:
image
3. b) scan an EIP 681 link
image

4. Tap back (in top left corner)

### Additional Information

- Status version: nightly, stable, all
- Operating System: Android

acolytec3:
fixes #9381 

### Summary

When the Scan QR code option is selected from the Chat home navigation and a QR code is scanned, later navigating back to the Chat tab returns to the QR scanner rather than the Chat Home and is clunky to the user.  This fixes this issue and navigates the user to the chat home page upon next tap on the Chat menu item.

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats
- public chats
- group chats

### Steps to test

Open Status
(From the Chats Tab) tap on the + plus button
Tap on Scan QR code
Scan some code
Tap Chat to navigate to chat screen
Chat home view should display

status: ready

fix-qr-scan-navigation

### DIFF
--- a/src/status_im/qr_scanner/core.cljs
+++ b/src/status_im/qr_scanner/core.cljs
@@ -2,7 +2,8 @@
   (:require [re-frame.core :as re-frame]
             [status-im.i18n :as i18n]
             [status-im.utils.utils :as utils]
-            [status-im.utils.fx :as fx]))
+            [status-im.utils.fx :as fx]
+            [status-im.utils.navigation :as navigation]))
 
 (fx/defn scan-qr-code
   [{:keys [db]} {:keys [deny-handler] :as identifier} qr-codes]
@@ -21,11 +22,16 @@
 
 (fx/defn set-qr-code
   [{:keys [db]} context data]
-  (merge {:db (-> db
-                  (update :qr-codes dissoc context)
-                  (dissoc :current-qr-context))}
-         (when-let [qr-codes (get-in db [:qr-codes context])]
-           {:dispatch [(:handler qr-codes) context data (dissoc qr-codes :handler)]})))
+  (let [navigation-stack {:chat-stack :home}]
+    (navigation/navigate-reset {:index   0
+                                :key     :chat-stack
+                                :actions [{:routeName :home}]})
+    (merge {:db (-> db
+                    (update :qr-codes dissoc context)
+                    (dissoc :current-qr-context)
+                    (update :navigation-stack navigation-stack))}
+           (when-let [qr-codes (get-in db [:qr-codes context])]
+             {:dispatch [(:handler qr-codes) context data (dissoc qr-codes :handler)]}))))
 
 (fx/defn set-qr-code-cancel
   [{:keys [db]} context]


src/status_im/qr_scanner/core.cljs
```diff
@@ -21,11 +22,16 @@
 
 (fx/defn set-qr-code
   [{:keys [db]} context data]
-  (merge {:db (-> db
-                  (update :qr-codes dissoc context)
-                  (dissoc :current-qr-context))}
-         (when-let [qr-codes (get-in db [:qr-codes context])]
-           {:dispatch [(:handler qr-codes) context data (dissoc qr-codes :handler)]})))
+  (let [[navigation-stack'] `({:chat-stack :home})]
```
yenda: On the left side of the let binding, you have the assignment, which can use destructuring, which means you can assign only parts of the structure on the right to symbols on the left.

In this particular case you have on the right a list ``()` that contains a map {:chat-stack :home} and on the left part of the assignement you destructure that list to only name the first element `navigation-stack'`.
This is unecessary so you can just do 

```suggestion
  (let [navigation-stack {:chat-stack :home}]
```
note that I also remove the `'` at the end, which means nothing, we used to use it sometimes to avoid overwriting a name in the scope but it's a better practice to find another name instead. In that case there is no other `navigation-stack` symbol in the scope.

acolytec3:
Just rebased on `develop` but looks like the builds are still failing.  Can someone tell me what the logs are saying?

src/status_im/qr_scanner/core.cljs
```diff
@@ -21,11 +22,16 @@
...
-         (when-let [qr-codes (get-in db [:qr-codes context])]
-           {:dispatch [(:handler qr-codes) context data (dissoc qr-codes :handler)]})))
+  (let [navigation-stack {:chat-stack :home}]
+    (navigation/navigate-reset {:index   0
+                                :key     :chat-stack
+                                :actions [{:routeName :home}]})
+    (merge {:db (-> db
+                    (update :qr-codes dissoc context)
+                    (dissoc :current-qr-context)
+                    (update :navigation-stack navigation-stack'))}
```
yenda:
```suggestion
                    (update :navigation-stack navigation-stack))}
```
this is why the build fails

flexsurfer:
sorry @acolytec3 i've reworked qr scanner here https://github.com/status-im/status-react/pull/9499 , now it's modal ,so this issues will be addressed in this PR

PR #9479, opened 2019-11-19
Weak-password disallows confirm code step

### LINKED ISSUE #9453: Can open multiaccount with empty password if tap on "Next" quickly 2 times in a row
# Bug Report

## Problem
If tap quickly several times on the "Next" button on the "Select key storage" screen, one "Create your password" screen will be dismissed and you will be redirected directly to "Confirm your password" screen and can proceed with an empty screen.

#### Expected behavior
will be redirected to "Create your password" screen

#### Actual behavior
redirected "Confirm your password" screen and can proceed with an empty screen.

### Reproduction

- Open Status
- Tap on "Generate new key"
- Proceed to "Select key storage" screen
- tap "Next" quickly 2 times

### Additional Information

- Status version: nightly 15/11/2019

acolytec3:
fixes #9453 

### Summary

When creating a new account in intro-wizard, on "Select Key Storage" screen, there is a race condition where you can double tap "Next" and skip Create Password Screen and get to "Confirm Password" screen with a blank password.

This PR fixes this bug and doesn't allow user to proceed to "Confirm Password" screen until a password is entered and passes the weak password checks.

### Testing notes
Run ADB logs to see console error message that is printed when user double taps "Next" button.

#### Platforms

- Android
- iOS

##### Functional

-new account

##### Non-functional
NA

### Steps to test

* Open Status
* Tap on "Generate new key"
* Proceed to "Select key storage" screen
* tap "Next" quickly 2 times
* Verify that you can't navigate to "Confirm Password" screen and that "Still weak password so not proceeding" appears in ADB logs

status: ready

### DIFF
--- a/src/status_im/multiaccounts/create/core.cljs
+++ b/src/status_im/multiaccounts/create/core.cljs
@@ -138,12 +138,16 @@
 (fx/defn intro-step-forward
   {:events [:intro-wizard/step-forward-pressed]}
   [{:keys [db] :as cofx} {:keys [skip?] :as opts}]
-  (let  [{:keys [step first-time-setup? selected-storage-type processing?]} (:intro-wizard db)]
+  (let  [{:keys [step selected-storage-type processing? weak-password?]} (:intro-wizard db)]
     (cond (confirm-failure? db)
           (on-confirm-failure cofx)
 
           (= step :generate-key)
           (init-key-generation cofx)
+
+          (and  (= step :create-code)
+                (= weak-password? true))
+          nil
 
           (and (= step :confirm-code)
                (not processing?))


acolytec3:
@churik Took a stab at this and seems to have been resolved in my PR.  @yenda @flexsurfer Any thoughts on the best way to throw an error in the logs?  It seems like this shouldn't be something the user is prompted about since it should only be done inadvertently.  Just leave the user on the Create Password screen until they enter something that satisfies the "weak password" check.

src/status_im/multiaccounts/create/core.cljs
```diff
@@ -138,13 +138,17 @@
...
-  (let  [{:keys [step first-time-setup? selected-storage-type processing?]} (:intro-wizard db)]
+  (let  [{:keys [step first-time-setup? selected-storage-type processing? weak-password?]} (:intro-wizard db)]
     (cond (confirm-failure? db)
           (on-confirm-failure cofx)
 
           (= step :generate-key)
           (init-key-generation cofx)
 
+          (and  (= step :create-code)
+                (= (get-in db [:intro-wizard :weak-password?]) true))
```
flexsurfer: i can see you destructured weak-password? in let why don't you use it here? and i'm not sure if we need a log so you can do smth like `(and  (= step :create-code) weak-password?) nil`

acolytec3: @flexsurfer Just coding on not enough sleep.  Fixed in my latest commit.  I also replaced the logging message with a `nil` so should be ready for final review.

flexsurfer (approved):

vitvly (approved):

yenda (approved):

yenda:
@acolytec3 thanks for another PR! You are killing it. Please make sure you sign your commit with GPG

churik:
@acolytec3 
Couldn't find in logs "Still weak password so not proceeding" (fulladb logcat with `debug` level [logcat.log](https://github.com/status-im/status-react/files/3868903/logcat.log)) - but issue is not reproducible for sure.

Thanks a lot for your effort!

acolytec3:
@churik Per the recommendation of @flexsurfer , I removed the logging in my latest commit.  It just returns a `nil` now when the navigation logic identifies this condition.  In my initial build, it was definitely showing the error log I'd put in there every time I tried to double tap the "Next" button though.  Glad it's working right now!

Issue #9492, opened 2019-11-19
Userpic (avatar) in group chat is not shown if sender of the message matches with last system message


churik:
# Bug Report

## Problem
If user sends a message just after he was joined to public chat - his userpic is not shown for both messages (system and sent).
So if sender of the message is the same that in last system message - userpic is not shown.

#### Expected behavior
Avatar is shown

#### Actual behavior
<img width="226" alt="photo_2019-11-19 16 08 54 jpeg 2019-11-19 16-09-16" src="https://user-images.githubusercontent.com/4557972/69158601-fd7ffa00-0ae6-11ea-8753-8223e5b06ab5.png">

### Reproduction

- User B: Open Status
- User B: Create group chat with user A
- User A: join the group chat and send a message

### Additional Information

[comment]: # (Please do your best to fill this out.)
- Status version: nightly 18/11/2019
- Operating System: Android, iOS

PR #9498, opened 2019-11-20
[Fixes: #9492] Consider system messages always in their own group

cammellos:
System messages should always display the user picture, and should never
be grouped.

status: ready


flexsurfer:
@cammellos could you please elaborate what is system message ?

cammellos:
https://github.com/status-im/status-react/issues/9492 , when in a group chat you have events we display: `blah has joined ....`

flexsurfer (approved):

vitvly (approved):

PR #9501, opened 2019-11-20
Added waiting for some elements to avoid failures

churik:
Added waiting for the absence or presence of some elements in the screens to avoid failures in:
- test_modify_transaction_fee_values
- test_send_stt_from_wallet
- test_request_public_key_status_test_daap


test/appium/views/home_view.py
```diff
@@ -205,11 +205,10 @@ def open_status_test_dapp(self, allow_all=True):
         dapp_view = self.dapp_tab_button.click()
         dapp_view.open_url(test_dapp_url)
         status_test_dapp = dapp_view.get_status_test_dapp_view()
-        for _ in range(2):
-            if allow_all:
-                status_test_dapp.allow_button.click()
-            else:
-                status_test_dapp.deny_button.click()
+        if allow_all:
```
Serhy, line 208: Why removed 'Allow'/'Deny' click twice here? In simpledapp we allow access to wallet and then with the second window, access to public key

Serhy: Ah I see click_until_absense_of_element does the thing, ok.

Serhy (approved):

PR #9503, opened 2019-11-20
Add troubleshooting notes for Yarn/Git and adb server/client issues

vitvly:
This PR addes 2 troubleshooting notes for Git issues during `yarn install` and some adb info.

And also a new Makefile command: `make android-devices`.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -42,3 +42,60 @@ _callImmediatesPass@http://localhost:8081/index.bundle?pla<…>
 ## Solution
 
 go through known faulty commit looking for deleted requires
+
+
+# Git "unable to access" errors during `yarn install`
+
+## Description
+Developer updates `package.json` file with a new dependency using a GitHub URL. So it looks like this:
+```
+  "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#feature/exportKeyWithPath",
+```
+Afterwards, when running e.g. `make react-native-android`, they might see the following confusing error:
+```
+# macOS
+fatal: unable to access 'https://github.com/siphiuel/react-native-status-keycard.git/': SSL certificate problem: unable to get local issuer certificate
+# Linux
+fatal: unable to access 'https://github.com/status-im/react-native-status-keycard.git/': Could not resolve host: github.com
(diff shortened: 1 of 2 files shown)


TROUBLESHOOTING.md
```diff
@@ -42,3 +42,59 @@ _callImmediatesPass@http://localhost:8081/index.bundle?pla<…>
...
+## Cause
+`yarn.lock` is not updated to be in sync with `package.json`.
+
+## Solution
+Update yarn.lock file. In order to do this, perform the following steps on a clean `status-react` repo:
+```
+cd status-react
+cp mobile/js_files/package.json .
+yarn install
+cp yarn.lock mobile/js_files/
```
jakubgs: You could just use `ln` to symlink the files.

jakubgs (approved):
Nice.

TROUBLESHOOTING.md
vitvly: Yes, thanks, fixed.

TROUBLESHOOTING.md
```diff
@@ -42,3 +42,60 @@ _callImmediatesPass@http://localhost:8081/index.bundle?pla<…>
...
+
+## Cause
+`yarn.lock` is not updated to be in sync with `package.json`.
+
+## Solution
+Update yarn.lock file. In order to do this, perform the following steps on a clean `status-react` repo:
+```
+cd status-react
+ln -sf mobile/js_files/package.json .
+ln -sf mobile/js_files/yarn.lock .
```
yenda, line 71: you don't want to symlink this one since the point of `yarn install` is to recreate it. However maybe @PombeirP has a better solution, maybe `yarn add "git-url#tag"` works but I haven't tested it. That would avoid having all the unrelated changes to yarn.lock

jakubgs: Hmmm, going point actually.

vitvly: If symlinked, it's still updated correctly. And with a symlink one does not have to copy `yarn.lock` over again to `mobile/js_files`.

jakubgs: Yes, unless you want to re-create `yarn.lock` from scratch, but that's an extreme edge case, I doubt we'd want to do that, so whatever.

yenda (approved):

PR #9508, opened 2019-11-21
Add SNT/STT to default account settings

### LINKED ISSUE #9493: No SNT by default after multiaccount creation / restoring
# Problem

acolytec3:
fixes #9493 

### Summary

SNT is not shown on Wallet home page as a default.  This was previously the case but probably lost with the change from account to multiaccount.  This adds them back to default account settings so they appear when a new multiaccount is retrieved/created.

### Testing notes
STT will show as a default if testing on Ropsten network

#### Platforms

- Android
- iOS

##### Functional

- wallet / transactions

### Steps to test

-Open Status
- Create/restore a multiaccount
- Tap Wallet home page
- Verify that SNT is displayed beneath ETH by default

status: ready 

Below is from a newly created multiaccount.

Screenshot_20191121-145806

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -82,7 +82,8 @@
 (def default-multiaccount-settings
   {:web3-opt-in?     true
    :preview-privacy? false
-   :wallet           {:visible-tokens {}}})
+   :wallet           {:visible-tokens {}
+                      :first-login true}})
 
 (def currencies
   {:aed {:id :aed :code "AED" :display-name (i18n/label :t/currency-display-name-aed) :symbol "د.إ"}
--- a/src/status_im/wallet/core.cljs
+++ b/src/status_im/wallet/core.cljs
@@ -220,6 +220,8 @@
   (let [addresses (or addresses (map (comp string/lower-case :address) accounts))
         chain     (ethereum/chain-keyword db)
         assets    (get-in settings [:wallet :visible-tokens chain])
+        first-login?   (get-in settings [:wallet :first-login])
+        SNT            (if first-login? :SNT nil)
         tokens    (->> (tokens/tokens-for all-tokens chain)
                        (remove #(or (:hidden? %)))
                        (reduce (fn [acc {:keys [address symbol]}]
@@ -238,7 +240,7 @@
          (multiaccounts.update/update-settings
           (assoc-in settings
                     [:wallet :visible-tokens chain]
-                    #{})
+                    (merge #{} SNT))
           {}))))))
 
 (fx/defn update-prices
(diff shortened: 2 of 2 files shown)


rachelhamlin (approved):

yenda:
Needs to be tested with a multiaccount that has other tokens as well, because this might break the token discovery

flexsurfer (approved):

acolytec3:
> Needs to be tested with a multiaccount that has other tokens as well, because this might break the token discovery

@yenda Added 3 more tokens, logged out, and logged back in and it seems to work.  Is there another scenario to look at?
Screenshot_20191122-100935

yenda:
@acolytec3 I was talking about restoring a multiaccount that owns different tokens. The correct behavior is that it should find these 3 tokens and display them

acolytec3:
> @acolytec3 I was talking about restoring a multiaccount that owns different tokens. The correct behavior is that it should find these 3 tokens and display them

Got it.  I tried restoring one of my accounts that has other tokens and it did not show them so agree that this does break that.  I'll take a look at the restore account code and see what changes are needed.

acolytec3:
@yenda @flexsurfer I think what's causing it not to pull up the other accounts is the `clean-up-results` method that's part of the get-token-balances flow.  Conceptually, we could do something like inject SNT into that method so it always adds SNT to the set of visible-tokens though that would mean that you could never permanently remove SNT from the Multiaccount view.  Is the idea to only add it as a visible token if there are no others during account creation?

acolytec3:
As I think about it, we could add a "first login" flag that gets set during account recovery and it could be checked by the process that cleans up balances (either before or after the `clean-up-balances` function is called and inject SNT back into the list of visible tokens if not already there on first login.

churik:
@acolytec3 thank you for the contribution!
The issue mentioned by @yenda:
>  was talking about restoring a multiaccount that owns different tokens. The correct behavior is that it should find these 3 tokens and display them

should be fixed.
As for question:
> is the idea to only add it as a visible token if there are no others during account creation?

The idea is to get SNT enabled by default, so:
- if user restores multiaccount with ETH + ZRX + ADX => should be enabled ETH + ZRX + ADX + SNT (0.0) 
- if user creates multiaccount => should be enabled ETH + SNT
- if user restores multiaccount with ETH =>  should be enabled ETH + SNT.

If user wants he should be able to switch off SNT using options > Manage Assests.

Tell me if smth is unclear.

acolytec3:
That makes sense. I think I have an idea how to make it work so will try to get a revised PR submitted this week.

acolytec3:
Apologies for the slow follow-up here.  I will hopefully have an updated PR done in the next day or so.  Worked out most of the bugs in my code this morning and just have to verify that it's working the way it's supposed to.

acolytec3:
@yenda @flexsurfer @churik  I've submitted an updated PR that I believe fixes the issue.  I've tested it with a newly created multiaccount where it  displays SNT .  I don't have any accounts on mainnet that have a balance in other tokens but not SNT so need someone else to test and verify if that works correctly to.

Also...not sure what happened but I think I fat-fingered and closed this PR.  I'll submit a new one since I'm not sure how to re-open this one.

PR #9512, opened 2019-11-22
Change 'chats' to 'chat'

rachelhamlin:
- Uncapitalizing "Public Chats" on empty chat screen
- Changing "Chats" to "Chat" in bottom nav


flexsurfer:
will be addressed here https://github.com/status-im/status-react/pull/9499

PR #9513, opened 2019-11-22
Fix/slow checkbox in assets

yenda:
There was two issues with the "Manage assets" screen when pressing a checkbox:
- it was taking extra time before subscriptions were recomputed
- it was taking extra time after as well

The first problem was apparently due to nested `touchable-highlight` components and fixed by making the checkbox a simple view
The second was caused by the re-rendering of every element of the list, sometimes multitple times, everytime the `:checked?` value of only one item was changed. To fix it I overwrote the `should-component-update` method.

### DIFF
--- a/src/status_im/ui/components/checkbox/view.cljs
+++ b/src/status_im/ui/components/checkbox/view.cljs
@@ -10,7 +10,7 @@
   [{:keys [on-value-change checked? accessibility-label
            disabled? style icon-style]
     :or   {accessibility-label :checkbox}}]
-  [react/touchable-highlight
+  [(if on-value-change react/touchable-highlight react/view)
    (merge {:style               (merge
                                  styles/wrapper
                                  style)
--- a/src/status_im/ui/components/list/views.cljs
+++ b/src/status_im/ui/components/list/views.cljs
@@ -103,7 +103,7 @@
                                       {:on-long-press on-long-press}))
    (conj item
          [react/view {:style (merge style styles/item-checkbox)}
-          [checkbox/checkbox props]])])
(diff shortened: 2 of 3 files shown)


src/status_im/ui/components/checkbox/view.cljs
```diff
@@ -8,9 +8,9 @@
   "react/check-box is currently not available on iOS,
   use it once it is available on all platforms"
   [{:keys [on-value-change checked? accessibility-label
-           disabled? style icon-style]
+           disabled? style icon-style not-touchable?]
     :or   {accessibility-label :checkbox}}]
-  [react/touchable-highlight
+  [(if on-value-change react/touchable-highlight react/view)
```
flexsurfer, line 13: `not-touchable?` isn't used ?

flexsurfer (approved):

rasom:
On my slow device it still takes 1.2-1.3s to check/uncheck token

rasom:
@yenda it seem that after `:wallet.settings/toggle-visible-token` it takes 800ms till rendering of the particular checkbox begins, so I suppose we need to rearrange subs so that each item is subscribed separately or something like this.

yenda:
@rasom but how long did it take before that? And are you sure that this is because of subscriptions? Did you check that they are ran immediately? 

Also are you talking about the numbers for this PR or develop?

rasom:
@yenda this PR. Will check how much it takes to start subs calculations. Then will check develop

rasom:
@yenda my mistake was that I checked dev env, in prod build it is faster:
 0.67-0.93s vs  1.7-2.3s in `develop`

rasom (approved):

PR #9526, opened 2019-11-25
[UI REVIEW] Touchables review

flexsurfer:
DO NOT MERGE!

you can find things like this, as you can see "cancel" and "learn more" are text, so it will be difficult to press, also "next" is not a component and has less touchable area

image


errorists:
I'm compiling a list of issues and fixes in [this Figma here](https://www.figma.com/file/Q1oiRp70EPqC3Eh6PgJABu/touchables?node-id=0%3A1), anyone can edit it

once done, I'll think how to best integrate it with the Component Library.

PR #9527, opened 2019-11-25
nix: Use mkFilter in local status-go source

pedropombeiro:
This PR improves the filtering of status-go Nix closure so that it doesn't capture folders such as `.ethereumtest`, `protocol` or `eth-node`.

status: ready


jakubgs (approved):

churik:
Not intentionally moved it to `IN TESTING`, don't think it requires manual testing

PR #9531, opened 2019-11-26
Nixpkgs upgrade

### LINKED ISSUE #9500: Upgrade nixpkgs to include Go 1.13
I was requested by @cammellos to upgrade the current Go version we are using:
https://github.com/status-im/status-react/blob/5ecc7590d98fafd834a065dad82c01649b47c973/nix/derivation.nix#L20
But in order to do that I might have to do a full upgrade of our [`nixpkgs`](https://github.com/status-im/nixpkgs) fork.

Previous Go upgrade for reference: https://github.com/status-im/nixpkgs/commit/35e0ad9f6d83f431fe5a156793e44f095b8039ab

jakubgs:
This upgrades a LOT of the tools and libraries that Nix provides by upgrading our fork of `nixpkgs`.

Changes:
* Upgrades `nixpkgs`
* Upgrades Go: `1.12` > `1.13`
* Upgrades Android SDK: `28.0.3` > `29.0.2`
* Makes `uploadToSauceLabs()` not use a pure Nix shell to avoid issues with certs
* Narrows down target for `scripts/copy-translations.sh` to `targets.mobile.jsbundle`
* Removes Desktop builds from CI since work on them is on-hold and new `nixpkgs` break them
* Fixes an issue with our patched `mkShell` in `shell.nix` by inheriting instead of setting `shellHook`
* Fixes gomobile status-go build by setting `GO111MODULE=off`
* Fixes gradle accepting empty `STATUS_GO_SRC_OVERRIDE` value

Resolves: #9500


nix/status-go/build-status-go.nix
```diff
@@ -53,11 +53,8 @@ let
     '';
 
     # replace hardcoded paths to go package in /nix/store, otherwise Nix will fail the build
-    preFixup = ''
-      ${preFixup}
-
+    preFixup = preFixup + ''
       find $out -type f -exec ${removeExpr removeReferences} '{}' + || true
-      return
```
pedropombeiro, line 60: Why did you remove this `return` here? It is required, otherwise the host `preFixup` script will continue executing, which we don't want. Maybe we can add a comment to make it explicit.

nix/derivation.nix
```diff
@@ -60,9 +62,9 @@ in {
 
   shell = {
     buildInputs = unique ([
+      yarn
```
pedropombeiro, line 65: Why did yarn need to come first in the list?

nix/mobile/android/android-env.nix
```diff
@@ -10,18 +10,18 @@ let
 
   androidComposition = androidenv.composeAndroidPackages {
     toolsVersion = "26.1.1";
-    platformToolsVersion = "28.0.2";
-    buildToolsVersions = [ "28.0.3" ];
+    platformToolsVersion = "29.0.5";
+    buildToolsVersions = [ "29.0.2" ];
```
pedropombeiro, line 14: I wonder if we should try to read these values from `android/gradle.properties`

nix/status-go/build-mobile-status-go.nix
```diff
@@ -31,10 +31,11 @@ let
       in with targetConfig; ''
       mkdir ${NIX_GOWORKDIR}
 
-      GOPATH=${gomobile.dev}:$GOPATH \
-      PATH=${makeBinPath [ gomobile.bin ]}:$PATH \
-      NIX_GOWORKDIR=${NIX_GOWORKDIR} \
-      ${concatStringsSep " " envVars} \
+      export GO111MODULE=off
```
pedropombeiro, line 34: Was there an error that you saw that caused us to need to export the variables?

pedropombeiro (approved):
Great job! Left a couple of comments.

nix/derivation.nix
jakubgs: OCD

nix/status-go/build-mobile-status-go.nix
jakubgs:
```
Building mobile library for android

type-checking package "github.com/status-im/status-go/mobile" failed (/build/go/src/github.com/status-im/status-go/mobile/multiaccount.go:4:2: could not import encoding/json (go/build: importGo encoding/json: exit status 2
go: cannot find GOROOT directory: /nix/store/eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee-go-1.13.4/share/go
```

vitvly (approved):

nix/mobile/android/android-env.nix
jakubgs: Not sure how difficult that would be in Nix.

jakubgs: I'll do it in another PR, since this fixes iOS fastlane build errors and I want to get this merged.

flexsurfer (approved):

PR #9541, opened 2019-11-28
[#9354] Unlock keycard account without card

### LINKED ISSUE #9354: Keycard sign in – Don't ask for the Keycard all the time to sign in
_Replacing https://github.com/status-im/status-react/issues/8180 with this new issue_

The user should be able to choose not to be asked for the Keycard to sign in all the time. This should be done using the same way currently "Save password" feature works, for non-keycard keys in Status.

- The should be a checkbox during sign in. I suggest it to be called. `Don’t ask for the card all the time` with a description below `If you don't want to be asked to tap the card each time you open Status`;
- If the user checks that checkbox, we should ask to enable biometric sign in for additional security;
- On iOS "biometric check" might be Face ID or Touch ID depending on the device. On Android afaik it's only possible to use fingerprint;
- This `Don’t ask for the card all the time`  checkbox should be disabled during device-to-card interaction process;
- There should be a way to disable and re-enable biometric sign in in the settings;
- If the `Don’t ask for the card all the time` feature is enabled as well as the biometric sign in, the sign in should happen with a biometric check. If it fails, the user should be offered to sign in using the passcode + Keycard tap;
- If the user unchecks the checkbox, when the biometric has been enabled before, we should show a popup and warn that it will be disabled now;
- The num pad should have an additional key (Face ID or fingerprint icon)
- This should work on small screens as well, see https://github.com/status-im/status-react/issues/9353 for additional specifications;

**Checkbox + Enable biometric**

rm1

**SIgn in when biometric is enabled**

rm2

**Disbaling `Don’t ask for the card all the time`**

rm3

Figma:
All the flows are described here, see flows number 6 and 7

rasom:
fix #9354 

To make it work `encryption-public-key` and `whisper-private-key` are 
stored on the devices when a user chooses this option. The former key is
used for multiaccount's database encryption, the latter is needed for a 
messaging. In case if a user wants to sign a transaction the card is 
still needed, we don't store wallet's keys on the device.

Other things were fixed/added:
- A user can enable biometric auth for a regular account when chooses
  to save the password on the device (if biometric auth is available). 
  This is done for feature parity between keycard and "on device" 
  accounts.
- The option to create/restore an account on a keycard is not shown on
  the devices which do not support NFC. Currently, the app just crashes
  if the user continues a flow which is not supported by the device.

status: ready


src/status_im/multiaccounts/biometric/core.cljs
```diff
@@ -81,6 +82,7 @@
       (.catch #(callback nil))))
 
 (defn get-supported [callback]
+  (log/debug "[biometric] get-supported")
```
cammellos, line 85: do we need this logs? (I don't mind keeping them, just wondering if they were leftovers)

cammellos (approved):

src/status_im/multiaccounts/biometric/core.cljs
rasom: I can remove this one, but they are actually useful during debugging, sometimes it's a bit hard to figure what's hidden behind the chain of re-frame's events.

cammellos: no that's fine,  I was just wondering if they were leftovers, by all means keep them if useful

andmironov (approved):

flexsurfer (approved):

flexsurfer:
one minor case:
select "save passord", cancel biometric, deselect "save password" , select "save password" - no popup with biometric, idk if it's expected

churik:
Small one, can be fixed later, but still the issue:

### Can't login if delete biometric data
Steps:
- create multiaccount
- enable biometric authentication inside of status app
- relogin
- remove fingerprints in phone settings (in case of touch id)
- reopen app

**Expected result:**
- logged out, "Don't ask for card to sign in" is disabled, can login with card

**Actual result:**
Biometric authentication error, if disable manually "Don't ask for card to sign in"  - can't proceed to 2 step
Can login after I add new fingerprint.
WDYT @rasom worth to be added separately?

Check-list
========================================================================
- "Save password" - enabling from profile, login screen
- "Don't ask for card to sign in" - enabling from profile, login screen
- "Save password" on device without biometric authentication
- "Save password" and "Don't ask for card to sign in"  when all biometric data is removed from device
- Signing messages in Dapps, signing transactions require keycard


Only thing that I would suggest to edit - after enabling "Don't ask for card to sign in" we need to relogin with card one more time.
So we have a controversial situation until first relogin - when "Don't ask for card to sign in" is enabled, but we still are obliged to use keycard.
@andmironov take a look:
telegram-cloud-document-4-5994797008227927564
WDYT?


Anyway, I don't think that any of these are blockers for PRs, and for sure can be addressed separately.

rasom:
@churik lets fix these issues in separate PRs

PR #9546, opened 2019-11-28
[#9213] ENS addr is checked instead of owner

### LINKED ISSUE #9213: ENS addr is checked instead of owner
# Bug Report

## Problem

the native ens registration flow checks the addr and not the owner of the ens name

#### Expected behavior

it should check if the owner is the user wallet, otherwise he can't make a transaction to change the pubkey for instance because he doesn't own the name, even if the addr is his.

this means that we probably need a few extra flows for the following cases:
- user is owner, addr is not his: need to set addr
- user is owner, addr is not his and pubkey is not his: need to set addr and pubkey
- user is not owner but addr and pubkey are his: allow to add username but warn user that if he doesn't have control over owning address, he will get his identity and funds stolen

#### Actual behavior

owner is ignored and only addr is checked

#### Notes

not a priority because you need to manage names outside of status to get these situations

### LINKED ISSUE #9473: Fix ENS setting glitches + disallow setting before Tx confirmed
# Bug Report

## Problem

As I was reclaiming my ENS name `rachel.stateofus.eth` on the 18/11/19 nightly, I ran into a couple of small snags:

1. The '->' button to trigger the transaction is very, very slow to respond. I tapped on it many times before the bottom sheet finally appeared.

2. When the bottom sheet did appear, it loaded with this `Invalid number` error.

IMG_5953

3. After I signed and saw the confirmation screen, the contract interaction bottom sheet remained open and the `Invalid number` appeared again. I also had to tap the `Cancel` button multiple times to make it go away.

IMG_5956

4. ([separate issue](https://github.com/status-im/status-react/issues/8837)) The system `Save password` option came up while I was trying to close the bottom sheet again. 

#### Expected behavior

- Do not allow user to "Show my ENS username in chats" unless the setting Tx is mined. It should be grayed out if the Tx is not yet confirmed.
- Buttons respond on first tap.
- No 'invalid number' error.
- Bottom sheet disappears once Tx has been signed. This should be fixed by preventing user from tapping button and triggering multiple Txs.

#### Acceptance Criteria

See above.

### Reproduction

- In ENS flow, enter a name you already own and follow instructions.

### Additional Information

- Status version: Nightly
- Operating System: iOS 13.1.3

flexsurfer:
fixes #9213
fixes #9473

### DIFF
--- a/src/status_im/utils/debounce.cljs
+++ b/src/status_im/utils/debounce.cljs
@@ -1,0 +1,11 @@
+(ns status-im.utils.debounce
+  (:require [re-frame.core :as re-frame]))
+
+(def timeout (atom {}))
+
+(defn debounce [event time]
+  (when @timeout (js/clearTimeout @timeout))
+  (reset! timeout (js/setTimeout #(re-frame/dispatch event) time)))
+
+(defn clear []
+  (when @timeout (js/clearTimeout @timeout)))
--- a/src/status_im/utils/debounce.cljs
+++ b/src/status_im/utils/debounce.cljs
@@ -1,11 +1,24 @@
 (ns status-im.utils.debounce
   (:require [re-frame.core :as re-frame]))
 
-(def timeout (atom {}))
+(def timeout (atom nil))
 
-(defn debounce [event time]
+(defn debounce-and-dispatch
+  "Dispatches event only if there were no calls of this function in period of *time* ms"
+  [event time]
   (when @timeout (js/clearTimeout @timeout))
   (reset! timeout (js/setTimeout #(re-frame/dispatch event) time)))
 
 (defn clear []
   (when @timeout (js/clearTimeout @timeout)))
+
+(def chill? (atom false))
+
+(defn dispatch-and-chill
+  "Dispateches event and ignores next calls in period of *time* ms"
+  [event time]
+  (when-not @chill?
+    (reset! chill? true)
+    (js/setTimeout #(reset! chill? false) time)
+    (re-frame/dispatch event)))
+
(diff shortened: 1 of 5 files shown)


vitvly (approved):

src/status_im/utils/debounce.cljs
```diff
@@ -0,0 +1,11 @@
+(ns status-im.utils.debounce
+  (:require [re-frame.core :as re-frame]))
+
+(def timeout (atom {}))
```
yenda: why does it start with a map?

src/status_im/utils/debounce.cljs
```diff
@@ -0,0 +1,11 @@
...
+
+(defn debounce [event time]
```
yenda: could you add a docstring to explain the expected behavior?

churik:
Checked on IOS 13.1 and Android 9.1:
- owner and ENS Ethereum Address  is set to wallet address when buy new ENS name stateofus.eth, can restore multiaccount and set it in chat

- owner only is set, ENS Ethereum Address  is not set (for `.eth` names) - can add it in Profile and  set visible in chat, can start chat with this ENS

- set primary ens, adding several ens names

- checked that glitches that mentioned as fixed in https://github.com/status-im/status-react/issues/9473 are fixed.

Waiting for e2e to complete.

Thanks @flexsurfer, really nice work!

Also noticed small issue (multitapping on "Register" button is also allowed, will create separate issue) - but it is in nightly and not related to this change.

flexsurfer:
@churik it would be really great to have an issue for v1 with all buttons where you can press several time, because they might break the app or be dangerous like when you sign tx twice or smth

churik:
@flexsurfer I believe we checked most of the places, but I'll check during next regression

PR #9548, opened 2019-11-28
Allow slashes in status-go branch names

pedropombeiro:
This PR renames the version used in the Nix path so that it doesn't contain forward slashes.

### DIFF
--- a/nix/status-go/default.nix
+++ b/nix/status-go/default.nix
@@ -43,13 +43,17 @@ let
         versionJSON = importJSON ../../status-go-version.json; # TODO: Simplify this path search with lib.locateDominatingFile
         versionRegex = "^v?[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+(-[[:alnum:].]+)$";
         sha256 = versionJSON.src-sha256;
+        sanitizeVersion = builtins.replaceStrings
+          [ "/" ]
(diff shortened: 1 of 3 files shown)


nix/status-go/default.nix
```diff
@@ -43,8 +43,12 @@ let
         versionJSON = importJSON ../../status-go-version.json; # TODO: Simplify this path search with lib.locateDominatingFile
         versionRegex = "^v?[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+(-[[:alnum:].]+)$";
         sha256 = versionJSON.src-sha256;
+        sanitizeVersion = builtins.replaceStrings
+          [ "/" ]
+          [ "_" ];
       in rec {
-        inherit (versionJSON) owner repo version;
+        inherit (versionJSON) owner repo;
+        version = sanitizeVersion versionJSON.version;
```
jakubgs: What's the difference between `version` and `versionName`?

jakubgs (approved):

nix/status-go/default.nix
pedropombeiro: Actually `version` should move to the `let` area since we don't use it elsewhere. I've refactored the PR, please re-review.

jakubgs (approved):
LGTM

PR #9553, opened 2019-11-29
[#9543] Signing phrase popup minor fix

### LINKED ISSUE #9543: Signing phrase popup minor fix
1. Let's treat the tap outside the popup (on the overlay) as "Maybe later" (if it's not treated this way already);
2. Let's show the signing phrase popup next time only when the app is opened again (not each time the wallet tab is opened)

flexsurfer:
fixes #9543

### DIFF
--- a/src/status_im/ui/screens/db.cljs
+++ b/src/status_im/ui/screens/db.cljs
@@ -255,7 +255,8 @@
                                 :signing/sign
                                 :signing/tx
                                 :signing/edit-fee
-                                :popover/popover]
+                                :popover/popover
+                                :wallet/sign-phrase-showed?]
                           :opt-un [::modal
                                    ::was-modal?
                                    ::rpc-url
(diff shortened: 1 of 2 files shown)


andmironov:
Tried this build on ANdroid, got an error when opened the app: 
```
Error. Vector's key for assoc must be a number
```

andmironov (approved):

rasom (approved):

churik:
> Error. Vector's key for assoc must be a number

@andmironov do you still see this issue? Can't reproduce it on Android 8 and Android 9

churik:
Tested on Android 8, Android 9 and IOS 13.
Looks good to me.
Failed e2e are unrelated.

flexsurfer:
> > Error. Vector's key for assoc must be a number
> 
> @andmironov do you still see this issue? Can't reproduce it on Android 8 and Android 9

@churik it's only possible to reproduce when you upgrade specific version, so that's fine

PR #9554, opened 2019-11-29
[#8585] Add watch only account

### LINKED ISSUE #8585: Add watch only account front-end
Placeholder to fill out later

Requires @gravityblast work on login API to enable

# Problem

An overview of the background required to understand the problem.
A problem description.

## Implementation

Known steps towards feature implementation.
What needs further specifying and investigating.

## Acceptance Criteria

Rules for the future PR to be accepted.

## Notes

Random notes to keep in mind while implementing it. Mostly about related issues and future plans and thoughts.

## Future Steps

Steps which should be taken after this issue has been resolved.

flexsurfer:
DEMO: https://www.youtube.com/watch?v=mw28EFE7mXs

fixes #8585

for QA: it's possible to delete watch only account in the settings

### DIFF
--- a/src/status_im/ui/screens/intro/views.cljs
+++ b/src/status_im/ui/screens/intro/views.cljs
@@ -549,7 +549,8 @@
       (toolbar/nav-button
        (actions/back #(re-frame/dispatch [:intro-wizard/navigate-back])))
       nil]
-     [react/view {:style {:flex 1 :justify-content :space-between}}
+     [react/view {:style {:flex            1
+                          :justify-content :space-between}}
       [top-bar {:step :enter-phrase}]
       [enter-phrase wizard-state]
       [bottom-bar (merge {:step :enter-phrase
--- a/src/status_im/ui/screens/intro/views.cljs
+++ b/src/status_im/ui/screens/intro/views.cljs
@@ -545,14 +545,11 @@
 (defview wizard-enter-phrase []
   (letsubs [wizard-state [:intro-wizard/enter-phrase]]
     [react/keyboard-avoiding-view {:style {:flex 1}}
-     [toolbar/toolbar
-      {:style {:border-bottom-width 0
-               :margin-top 16}}
+     [toolbar/toolbar {:style {:border-bottom-width 0}}
       (toolbar/nav-button
        (actions/back #(re-frame/dispatch [:intro-wizard/navigate-back])))
       nil]
-     [react/view {:style {:flex 1
-                          :justify-content :space-between}}
+     [react/view {:style {:flex 1 :justify-content :space-between}}
       [top-bar {:step :enter-phrase}]
       [enter-phrase wizard-state]
       [bottom-bar (merge {:step :enter-phrase
--- a/src/status_im/ui/screens/wallet/account/views.cljs
+++ b/src/status_im/ui/screens/wallet/account/views.cljs
@@ -39,7 +39,7 @@
      [icons/icon icon {:color colors/white}]
      [react/text {:style {:margin-left 8 :color colors/white}} label]]]])
 
-(views/defview account-card [{:keys [address color] :as account}]
+(views/defview account-card [{:keys [address color type] :as account}]
   (views/letsubs [currency        [:wallet/currency]
                   portfolio-value [:account-portfolio-value address]
                   window-width    [:dimensions/window-width]]
@@ -62,10 +62,13 @@
                                       :accessibility-label :share-wallet-address-icon}]]]
      [react/view {:height                     52 :background-color colors/black-transparent-20
                   :border-bottom-right-radius 8 :border-bottom-left-radius 8 :flex-direction :row}
-      [button
-       (i18n/label :t/wallet-send)
-       :main-icons/send
-       #(re-frame/dispatch [:wallet/prepare-transaction-from-wallet account])]
+      (if (= type :watch)
+        [react/view {:flex 1 :align-items :center :justify-content :center}
+         [react/text {:style {:margin-left 8 :color colors/white}} "Watch-only"]]
+        [button
+         (i18n/label :t/wallet-send)
+         :main-icons/send
+         #(re-frame/dispatch [:wallet/prepare-transaction-from-wallet account])])
       [react/view {:style styles/divider}]
       [button
        (i18n/label :t/receive)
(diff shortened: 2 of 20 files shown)


src/status_im/ui/screens/wallet/account/views.cljs
```diff
@@ -62,10 +62,13 @@
                                       :accessibility-label :share-wallet-address-icon}]]]
      [react/view {:height                     52 :background-color colors/black-transparent-20
                   :border-bottom-right-radius 8 :border-bottom-left-radius 8 :flex-direction :row}
-      [button
-       (i18n/label :t/wallet-send)
-       :main-icons/send
-       #(re-frame/dispatch [:wallet/prepare-transaction-from-wallet account])]
+      (if (= type :watch)
+        [react/view {:flex 1 :align-items :center :justify-content :center}
+         [react/text {:style {:margin-left 8 :color colors/white}} "Watch-only"]]
```
vitvly: Can "Watch-only" be moved to translations?

src/status_im/ui/screens/wallet/add_new/views.cljs
```diff
@@ -29,40 +31,83 @@
...
-      :on-press
-      #(re-frame/dispatch
-        [:navigate-to :add-new-account-password])}]]])
+      :on-press    #(re-frame/dispatch [:wallet.accounts/start-adding-new-account {:type :generate}])}]
+    [list-item/list-item
+     {:type                 :section-header
+      :container-margin-top 24
+      :title                "Advanced"}]
+    [list-item/list-item
+     {:title       "Enter a seed phrase"
```
vitvly, line 40: Also translation missing here.

src/status_im/ui/screens/wallet/add_new/views.cljs
```diff
@@ -29,40 +31,83 @@
...
+                  :align-items     :center :margin-horizontal 16}
+      [react/view
+       [react/text {:style {:typography :header :margin-top 16}} "Add a watch-only address"]
+       [react/text {:style {:color colors/gray :text-align :center :margin-vertical 16}}
+        "Enter the address to watch"]]
+      [react/view {:align-items :center :flex 1 :flex-direction :row}
+       [react/text-input {:auto-focus        true
+                          :multiline         true
+                          :text-align        :center
+                          :placeholder       "Enter address"
```
vitvly, line 78: And here.

vitvly (approved):

src/status_im/ui/screens/intro/views.cljs
```diff
@@ -545,14 +545,11 @@
...
-     [toolbar/toolbar
-      {:style {:border-bottom-width 0
-               :margin-top 16}}
+     [toolbar/toolbar {:style {:border-bottom-width 0}}
       (toolbar/nav-button
        (actions/back #(re-frame/dispatch [:intro-wizard/navigate-back])))
       nil]
-     [react/view {:style {:flex 1
-                          :justify-content :space-between}}
+     [react/view {:style {:flex 1 :justify-content :space-between}}
```
rasom: What's the point of putting of all these properties to one line? It makes it harder to read imo

rachelhamlin (approved):
Wow, okay! Way to go @flexsurfer, this looks great.

churik:
### Issue 1: No name for account by default

If you add new account (not watch-only) in wallet - 'Account 1' is set by default. In case of watch-only account no name by default and focus is not set by defaul in Name field.
**Steps:**
-  create an account
- add watch-only account

**Expected result:**
Name is set by default and Finish button is active

**Actual result:**
photo_2019-12-03 17 16 25


### Issue 2: Watch-only wallets is counted in overall balance

I.e. on your accounts 20 USD, on watch-only accounts 200 USD, overall balance is 220 USD.
User can add any accounts and it doesn't mean that he is the owner.
In this case I would prefer to see separate settings in the wallet, but by default, it should be disabled I suppose.

Let's ask @hesterbruikman and @andmironov about it.

hesterbruikman:
> Issue 2: Watch-only wallets is counted in overall balance

Thanks for pinging @churik. I checked the specification doc of May this year. In [there ](https://docs.google.com/document/d/1-aLBJ1gemMfGJDPRqBDffpCOyNDDAYBfNlWoRIAJVkM/edit?usp=sharing) we concluded to include watch-only in the overall balance for P0. As in, we planned to change this eventually, but didn't have the bandwidth to come up with a better proposal at the time.

As time passed, maybe there are better ways of handling this @guylouis @andmironov wdyt? If there are no immediate alternatives I think it's ok to include the amount following the same assumption as back then that the most likely use case is adding your own cold storage address.

guylouis:
Yes we had discussed this and considered that the base case is that the watch only accounts are some of the user's own cold wallets or multisigs contracts, and that the base line scenario should be to include them in the sum. 

Of course, if the user is using this feature to watch some accounts that don't belong to him or that he don't want to add for some reason then it would be nice to have the option to exclude it for instance by going to its properties. 

Probably the one thing to decide is if any account can be removed from the sum or only 'watch only' accounts. My 2 cents is that it could be any to give maximum flexibility to users and be as generic as possible.

hesterbruikman:
> Probably the one thing to decide is if any account can be removed from the sum or only 'watch only' accounts. My 2 cents is that it could be any to give maximum flexibility to users and be as generic as possible.

Good point! If that's the case, it could be another multi-select item in the context menu, similar to `Manage assets`; All accounts selected by default. cc @andmironov 

I would still leave it out of V1 for now to focus on more pressing issues.

andmironov:
@hesterbruikman @churik  Hey, I really like the idea, provides great flexibility, but I'd discuss (or just add) that later. My opinion is that adding a watch-only account to your Status wallet means that you want to have the balance of it to be summed up with your overall balance. At least I see it as the main use-case

andmironov (approved):

flexsurfer:
@churik are we ok to merge it then?

churik:
@flexsurfer issue 1 is fixed?

flexsurfer:
@churik it's expected, as designed, we don't have a default value for watch accounts names

churik:
Ok, than let's set focus on this field. 
"Account is added" with disabled "Finish"  button seems to me unexpected. But I can be wrong ofc, let's ask another opinions.
@andmironov your opinion on issue 1?

flexsurfer:
@churik Andrei gave the approve, if you delete name in the generated account, finish button will be disabled as well

PR #9555, opened 2019-11-29
Make tooltips visible in page xml

Serhy:
Tooltips are not visible in page xml when we navigate between screen. They are visible but not searchable. This PR will fix it.

Status: ready


flexsurfer:
we could add a config for e2e tests , wdyt ?

Serhy:
Yeaa, config for e2e build is a great and more than enough idea, @flexsurfer !

flexsurfer (approved):

Serhy:
Works for me now. Thanks @flexsurfer for help!

PR #9557, opened 2019-11-30
fiddle code snippets

flexsurfer:
image


fiddle/src/status_im/ui/components/react.cljs
```diff
@@ -51,6 +51,7 @@
 (def text-input-class (get-web-class "TextInput"))
 (def image-class (get-web-class "Image"))
 (def picker-obj nil)
+(def animated-flat-list-class #())
```
vitvly, line 54: Why does is it declared differently from e.g. `picker-class` below?

vitvly (approved):

Issue #9470, opened 2019-11-18
Ignore Dark Mode setting on iOS13


errorists:
# Problem

At present, whenever dark mode is enabled on iOS, it only introduces a bunch of bugs without any advantages. 

- Status bars are presented white making them invisible against our white backgrounds, 

- photo picker dialogues are presented with incorrect styling all over, making them difficult to use
- dark keyboards are an unwelcome sight without a Dark Mode of our own.

## Implementation

To quote the maestro @flexsurfer 

> it seems like simple one to ignore dark mode
add this to your Info.plist file:
    <key>UIUserInterfaceStyle</key>
    <string>Light</string>

## Acceptance Criteria

There are no visible changes to our UI when dark mode is enabled.

thank you kindly! 🙏

rasom:
All mentioned issues except Status Bar are fixed in #9564. 
Status Bar should be fixed with RN upgrade https://github.com/facebook/react-native/issues/26619#issuecomment-541143915

PR #9564, opened 2019-12-02
[#9470] ignore dark mode on iOS

rasom:
fixes (partially)  #9470

### DIFF
--- a/ios/StatusIm/Info.plist
+++ b/ios/StatusIm/Info.plist
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UIUserInterfaceStyle</key>
+	<string>Light</string>
 	<key>CFBundleBuildUrl</key>
 	<string>????</string>
 	<key>NSFaceIDUsageDescription</key>


ios/StatusIm/AppDelegate.m
```diff
@@ -50,6 +50,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:bridge
                                                       moduleName:@"StatusIm"
                                                initialProperties:nil];
+  if (@available(iOS 13.0, *)) {
+      rootView.overrideUserInterfaceStyle = UIUserInterfaceStyleLight;
+  }
```
jakubgs: Maybe a comment explaining this hack would be a good idea?

rasom: It looks kind of self explanatory to me, what exactly would you like to see there? I will add it.

flexsurfer (approved):

errorists:
I don't see a difference, Status Bar is still using the white appearance, making it invisible. Keyboards are using the dark appearance, image picker remains all bugged up 😿

ios/StatusIm/AppDelegate.m
jakubgs: Comments aren't supposed to explain what code does, they are supposed to explain why we want what it does. Why do we want this hack? It's not clear from the code.

rasom:
@errorists the issue with amount on transaction sending is fixed, but it seems that the stuff which is outside application's view is not affected by this change

errorists:
@rasom can you go a layer upward, more global, more root? ;)

ios/StatusIm/AppDelegate.m
rasom: We will have a link to the original GH issue in the commit, should be sufficient enough imo, but well I can put that link to the code as well if necessary.

jakubgs: Yes, usually `git blame` is all you need, but if something is a "temporary solution" as you state in the PR description than I think it's only fair to indicate that with a comment. It's too easy to forget about stuff like that and leave it there without a clear indication that it's a temporary hack. But that's just my preference.

rasom:
it seems that we need to upgrade RN in order to fix status bar https://github.com/facebook/react-native/issues/26619#issuecomment-541143915

rasom:
keyboard should be fixed now, not sure if we can fix image picker

rasom:
I read a bit more about `UIUserInterfaceStyle` and seems that it is a problem only in case if app is built using Xcode 10.*, it should work in our case. 

@errorists please take a look when build will be ready. Keyboard, image picker, transaction amount are fixed, status bar is not. We still need to upgrade RN for status bar, I will try to do this tomorrow.

rasom:
I will merge this one and upgrade RN in a separate PR, it might take some time.

PR #9568, opened 2019-12-02
[#9288]Improve 'snackbar' animation behavior

### LINKED ISSUE #9288: Improve 'snackbar' animation behavior
# Problem
We use a 'snackbar' (i.e. notification bar) to indicate the client's connectivity status. On spotty networks the snackbar changes states every few milliseconds and makes the app appear glitchy (i.e. bar is visually perceived as 'flicker').

The bar has the following states:

- Connecting
- Offline
- Connected

The snackbar is used in combination with the blue 'nightrider' when connected that indicates 'fetching messages'. When the green connection bar appears, it is sometimes shown at the same time as the fetching messages bar.

Click each button in this prototype to view the behavior: https://framer.cloud/MsLgV/

## Implementation
The best solution to remove the glitchy appearance is still to be specified, but it [mostly has to do with](https://github.com/status-im/status-react/issues/9288#issuecomment-557576337): 

• adding delays when we consider a change of state permanent,
• being more lax and forgiving
• smoothing the animation so it doesn't drop frames,
• forcing the state to be persistent for at least ~1000ms so it doesn't blink,
• only saying that you're Online after being Offline for a prolonged period, not every time we wake the app.

#### Other ideas
- Seperate the logic of when the client is considered 'connected' from the immediate display of the snackbar. E.g. if the connection remains for ca. 1000ms show snackbar
- Introduce an 'Unstable network' state or stay offline if connection is lost more than x times within a given period.
- Introduce dependency between blue nightrider and snackbar to ensure they do not appear at the same time. 
- We may want to change the definition of connectivity. 

## Acceptance Criteria
The basic requirement of this issue is UI smoothing.

When on a spotty network, the snackbar showing connectivity status shouldn't change as frequently as it does now, but still inform users of whether or not they are able to send and receive messages.

Please force an uninterrupted display of a state for 1000ms.

Snackbar and nightrider never appear at the same time. Snackbar takes precedence over nightrider. User will only see the nightrider if definitively _online_ and collecting messages.

Please also prevent the snackbar from displaying each time the user wakes the app from background. It's gratuitous, unless the app is having trouble connecting and remains _offline_ for some reasonable amount of time.

## Notes
Feel free to propose additional solutions, we're not sure on the best approach here.

## Future Steps
Test on spotty network
Consider reintroducing the snackbar to the Wallet depending on latest loading state indicators (https://github.com/status-im/status-react/issues/8443#issuecomment-539045418)

tbenr:
fixes #9288

~~I started moving the logic from `subscription` to an `fx.` The idea is that, from `fx,` I set a new connectivity state in `db` (by "absorbing"\"smoothing" the unwanted fast state changes). The connectivity view subscribes directly to this new `:connectivity/status`.
**a feedback is needed on that**
I don't like a lot the "intrusiveness" in `mailserver` but I didn't find an alternative way to completely decouple state logic from subscription\ui.~~

still need to do:
- [x] "absorbing" \ "smoothing" logic
    - transition from `connected` or `connecting` to `offline` are immediate.
    - all other transitions (including activation of 'nightrider') will be applied no more than 1 per second.
- [x] wallet state\chat state separation
    - states are offline-connecting-connected.
    - I left the three error messages currently supported, showed here:
       https://github.com/status-im/status-react/issues/9288#issuecomment-560474381
       https://github.com/status-im/status-react/issues/9288#issuecomment-560496633 
    - wallet offline state can be based on `:offline?` subscription.
- [x] avoid simulaneous snackbar\nightrider
   - when snackbar goes away, if still loading, nightrider appears.
- [x] avoid "connecting" when coming back from background
   - coming back from background, in the first 5 seconds, all state transitions are delayed by 5s. Thus if the app is able to reconnect within this timeframe, no 'snackbar' is displayed. The same applies to 'nightrider'.
- [x] code cleanup

status: ready

### DIFF
--- a/src/status_im/init/core.cljs
+++ b/src/status_im/init/core.cljs
@@ -21,13 +21,14 @@
 (fx/defn initialize-app-db
   "Initialize db to initial state"
   [{{:keys [view-id hardwallet initial-props desktop/desktop
-            supported-biometric-auth network/type]} :db}]
+            supported-biometric-auth network/type app-active-since]} :db now :now}]
   {:db (assoc app-db
               :initial-props initial-props
               :desktop/desktop (merge desktop (:desktop/desktop app-db))
               :network/type type
               :hardwallet (dissoc hardwallet :secrets)
               :supported-biometric-auth supported-biometric-auth
+              :app-active-since (or app-active-since now)
               :view-id view-id)})
 
 (fx/defn initialize-views
--- a/src/status_im/ui/components/connectivity/view.cljs
+++ b/src/status_im/ui/components/connectivity/view.cljs
@@ -1,15 +1,17 @@
 (ns status-im.ui.components.connectivity.view
   (:require-macros [status-im.utils.views :refer [defview letsubs] :as views])
   (:require [reagent.core :as reagent]
+            [re-frame.core :as re-frame]
             [status-im.ui.components.react :as react]
             [status-im.ui.components.connectivity.styles :as styles]
             [status-im.i18n :as i18n]
             [status-im.ui.components.colors :as colors]
             [status-im.ui.components.animation :as animation]
             [status-im.utils.utils :as utils]
+            [status-im.utils.datetime :as datetime]
             [status-im.utils.platform :as platform]))
 
-(def connectivity-bar-height 35)
+(def connectivity-bar-height 36)
 (def neg-connectivity-bar-height (- connectivity-bar-height))
 
 (defn easing [direction n]
@@ -68,54 +70,76 @@
 
 (def to-hide? (reagent/atom false))
 
-(defn manage-visibility [connected? anim-opacity anim-y]
+(defn manage-visibility [connected? animate? anim-opacity anim-y status-hidden]
+  "status-hidden is a per-view state, while to-hide? is a global state common to
+all connectivity views (we have at least one view in home and one in chat)"
   (if connected?
-    (when @to-hide?
-      (animation/start
-       (animation/parallel
-        [(animation/timing anim-opacity
-                           {:toValue 0
-                            :delay 800
-                            :duration 150
-                            :easing (.-ease (animation/easing))
-                            :useNativeDriver true})
-         (animation/timing anim-y
-                           {:toValue         (if platform/desktop? 0 neg-connectivity-bar-height)
-                            :delay           800
-                            :duration        150
-                            :easing          (.-ease (animation/easing))
-                            :useNativeDriver true})])
-       ;; second param of start() - a callback that fires when animation stops
-       #(reset! to-hide? false)))
+    (if animate?
+      (when (and @to-hide? (not @status-hidden))
+        (animation/start
+         (animation/parallel
+          [(animation/timing anim-opacity
+                             {:toValue 0
+                              :delay 800
+                              :duration 150
+                              :easing (.-ease (animation/easing))
+                              :useNativeDriver true})
+           (animation/timing anim-y
+                             {:toValue         (if platform/desktop? 0 neg-connectivity-bar-height)
+                              :delay           800
+                              :duration        150
+                              :easing          (.-ease (animation/easing))
+                              :useNativeDriver true})])
+         ;; second param of start() - a callback that fires when animation stops
+         #(do (reset! to-hide? false) (reset! status-hidden true))))
+      (do
+        (animation/set-value anim-opacity 0)
+        (animation/set-value anim-y (if platform/desktop? 0 neg-connectivity-bar-height))
+        (reset! to-hide? false)
+        (reset! status-hidden true)))
     ;; else
-    (animation/start
-     (animation/parallel
-      [(animation/timing anim-opacity
-                         {:toValue 1
-                          :duration 150
-                          :easing (.-ease (animation/easing))
-                          :useNativeDriver true})
-       (animation/timing anim-y
-                         {:toValue (if platform/desktop? connectivity-bar-height 0)
-                          :duration 150
-                          :easing (.-ease (animation/easing))
-                          :useNativeDriver true})])
-     ;; second param of start() - a callback that fires when animation stops
-     #(reset! to-hide? true))))
+    (if animate?
+      (when (and (not @to-hide?) @status-hidden)
+        (animation/start
+         (animation/parallel
+          [(animation/timing anim-opacity
+                             {:toValue 1
+                              :duration 150
+                              :easing (.-ease (animation/easing))
+                              :useNativeDriver true})
+           (animation/timing anim-y
+                             {:toValue (if platform/desktop? connectivity-bar-height 0)
+                              :duration 150
+                              :easing (.-ease (animation/easing))
+                              :useNativeDriver true})])
+         ;; second param of start() - a callback that fires when animation stops
+         #(do (reset! to-hide? true) (reset! status-hidden false))))
+      (do
+        (animation/set-value anim-opacity 1)
+        (animation/set-value anim-y (if platform/desktop? connectivity-bar-height 0))
+        (reset! to-hide? true)
+        (reset! status-hidden false)))))
 
 (defn connectivity-status
-  [{:keys [connected?]} anim-translate-y]
+  [{:keys [connected?]} anim-translate-y status-hidden]
   (let [anim-translate-y (or anim-translate-y (animation/create-value 0))
         anim-opacity     (animation/create-value 0)]
-    (manage-visibility connected?
-                       anim-opacity anim-translate-y)
     (reagent/create-class
-     {:component-did-update
+     {:component-did-mount
+      (fn []
+        (manage-visibility connected? false
+                           anim-opacity anim-translate-y status-hidden))
+      :should-component-update
+      ;; ignore :loading-indicator?
+      (fn [_ [_ old_p] [_ new_p]]
+        (not= (dissoc old_p :loading-indicator?)
+              (dissoc new_p :loading-indicator?)))
+      :component-did-update
       (fn [comp]
-        (manage-visibility (:connected? (reagent/props comp))
-                           anim-opacity anim-translate-y))
+        (manage-visibility (:connected? (reagent/props comp)) true
+                           anim-opacity anim-translate-y status-hidden))
       :reagent-render
-      (fn [{:keys [view-id message on-press-fn connected? connecting?] :as opts}]
+      (fn [{:keys [view-id message on-press-event connected? connecting?] :as opts}]
         [react/animated-view {:style               (styles/text-wrapper
                                                     (assoc opts
                                                            :height (if platform/desktop?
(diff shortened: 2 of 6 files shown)


tbenr:
@errorists @hesterbruikman please have a look if makes sense to you. The only missing part is the wallet status. A quick feedback from @cammellos to https://github.com/status-im/status-react/issues/9288#issuecomment-560506209 would help :-)

tbenr:
@jakubgs `make release-android` and `make test` are ok on my laptop.. can you have a look why android failed? thanks!

jakubgs:
```
internal/modules/cjs/loader.js:638
    throw err;
    ^

Error: Cannot find module 'react'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Module._load (internal/modules/cjs/loader.js:562:25)
    at Function.hookedLoader [as _load] (/home/jenkins/workspace/status-react_prs_android_PR-9568/target/test/test.js:14:10)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at /home/jenkins/workspace/status-react_prs_android_PR-9568/target/test/reagent/impl/util.js:11:139
    at Object.<anonymous> (/home/jenkins/workspace/status-react_prs_android_PR-9568/target/test/reagent/impl/util.js:17:3)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
Error encountered performing task 'doo' with profile(s): 'test'
Subprocess failed
```
I've seen this failure before, but I never knew what the reason is. It seems random.

hesterbruikman:
Having trouble testing properly. On iOS I appear to be offline on chat entirely, even though my device is online.

Tried public chat and 1:1 message.

If you think the cause is unrelated, could you record your screen with the search bar behavior @tbenr ?

IMG_0308

tbenr:
@hesterbruikman I had the same issue in android too. I reinstalled the app and created a new account. I think something changed in one of the latest PR.

tbenr:
@hesterbruikman and i think due to the same PR, now all channels are empty. seems that we cant see messages coming from the old app.

hesterbruikman:
@tbenr thanks for your patience on this issue! Tried the Android build; Here are some things I notice:

- Feels a lot better than the manic blue nightrider and snackbar toggling we had before!
- (Unrelated to PR?) Newly added Public chats are not fetching messages even though I'm positive they contain messages. A public chat that I already joined with the account I retrieved did fetch messages. Is this the same thing you came across with _"all channels are empty"_?

- I seem to only get a `Connecting...` state when switching to Wifi and not when switching to Mobile data. This seems like a bug as the 3 specified states for both Wallet and Chat are: Connecting - Connected - Offline; Implemented seperately. Result is that I'm clearly on 4G, sync over mobile data is set, but I appear offline. See image below.

Connecting state

- The copy of `History syncing offline, waiting for Wi-Fi. Change` can be shortened. See this design by @errorists 
Waiting for wifi




I will leave this install for ca. 24h with some high traffic channels so I can see how the `fetching messages blue nightrider` behaves in combination with the snackbar.

cammellos:
@hesterbruikman current PR/nightly builds are not compatible with earlier versions, so in order to check whether this build is receiving messages it should be tried in a chat where we know that there are messages from recent builds (use this PR for example, that's the safest)

hesterbruikman:
Thanks @cammellos that resolves the 2nd bullet point 👍 
> (Unrelated to PR?) Newly added Public chats are not fetching messages even though I'm positive they contain messages. A public chat that I already joined with the account I retrieved did fetch messages. Is this the same thing you came across with "all channels are empty"?

Sent message from one account in public channel using PR build, shows up when checking the same public channel with another account using PR build.

tbenr:
@hesterbruikman thanks for the feedback! A couple of questions:
1. I'm changing `History syncing offline, waiting for Wi-Fi. Change` to `Offline waiting for Wi-Fi. Change`. Since it is a translated label, whats the best way to "re-trigger" translation process in the other language?

2. regarding bullet 3 related to the missing `Connecting...` sounds a bit strange since i didn't change anything which could justify a different behaviour between mobile and wifi connectivity. As per my experience sometime you have to wait a bit more until it tries to reconnect. Can you try again waiting a bit more?

Notice also that sometimes you can pass from `Offline` directly to `Connected` because I impose, as requested, the state change to be effective after 1s, and if in the meantime you get connected, the final state `Connected` will override the intermediate `Connecting...` not yet "activated"

errorists:
the iOS builds all expired, can I ask for some links, I want to test this one out :)

hesterbruikman:
>     1. I'm changing `History syncing offline, waiting for Wi-Fi. Change` to `Offline waiting for Wi-Fi. Change`. Since it is a translated label, whats the best way to "re-trigger" translation process in the other language?

@tbenr checked with @jinhojang6 Automated, changes will appear in Localise

> 2. regarding bullet 3 related to the missing Connecting... sounds a bit strange since i didn't change anything which could justify a different behaviour between mobile and wifi connectivity. As per my experience sometime you have to wait a bit more until it tries to reconnect. Can you try again waiting a bit more?

Could be an existing bug on spotty networks. I have **not** been able to reproduce 👍. Please bear with me for an hour or so while I walk around the house trying to mimic spotty network behavior 🏃‍♀

hesterbruikman:
Here's a [recording](https://file.globalupload.io/BLROMtUC7q.gif) that shows issue nr. 2. It could be that it's simply caused by my phone showing I have 4G connection, while I actually don't have internet connect. If that's the case my phone is fooling me, Status is actually `Offline` and not `Connecting...`

To reproduce:
1. Start one multiaccount
2. Set Sync settings to Use mobile data
3. Sign out
4. Switch Wi-Fi off
5. On same build, start another multiaccount
6. Join same public channel
7. Send bunch of messages 
8. Switch to first multiaccount

Result: Snackbar remains set to `Offline` for ca. 30sec, while phone seems to have 4G connection. Then switches instantly to Connected.

9. Toggle Airplane mode on
10. Toggle Airplane mode off

Result: Snackbar remains set to `Offline` for ca. 15sec, while phone seems to have 4G connection. Then switches instantly to Connected.

tbenr:
@hesterbruikman Thanks for the feedback! I think I found the issue

tbenr:
@hesterbruikman yes now it is definitely better. As always I introduce this with the very last change before committing :)

changed `History syncing offline, waiting for Wi-Fi. Change` to `Offline waiting for Wi-Fi. Change`

@errorists you have a fresh new iOS build to test.

errorists:
thanks a bunch! In the limited testing I was able to do on rather ok networks, it seems good, much improved and not such a nuisance anymore 🎉 I only managed once to get the 'Connected' state stuck right after opening the app, changing tabs helped to make it disappear. I'm curious for QA to stress test this

tbenr:
@errorists I'd like to catch that 'connected' stuck. Are you on android or ios?

errorists:
@tbenr latest iOS. I tested the build yesterday in the evening, this happened today in the morning after waking the app and signing in. Can reproduce each time I wake the app from being killed.

edit: Ok, this time I can't make it go away, changing tabs didn't help.

errorists:
@tbenr recorded a video [here](https://www.dropbox.com/s/v0thlfvdisfzjpa/tbenr.mp4?dl=0)

tbenr:
@errorists great! This evening I'll take a look and fix

tbenr:
@errorists I replicated the issue and now should be fixed.

errorists:
hey, so are we ready to merge this please? :) @tbenr if you don't mind adding one more small fix in the meantime, to increase the height of the connectivity snackbar from 35 to 36 to comply with the design? 🙏

tbenr:
@errorists done!

hesterbruikman:
@tbenr Happy New Year!

One issue here, can you see if you can identify the cause in the code? If need be we can separate it from this issue to not delay merging these improvements.

- Let phone go to sleep mode 
- Wake up phone and Open Status and then go back to app (to Status) 
- it shows "Offline" state ~5-10 sec.

Will try to replicate in parallel.

tbenr:
@hesterbruikman happy new year you too!
I'll double check it, if you could take a video of that it would help

hesterbruikman:
Thanks @tbenr!

Can't reproduce myself, but @churik is looking into it. Could be an issue in nightly that is coincidentally resolved (not introduced) by your PR.

churik:
I think that it is separate issue, steps in my case are:
- login to Status
- join any chat
- lock phone
- wait ~3 min (in my case sleep in "Lock screen" is set to "After 1 min of inactivity")
- unlock phone

Status came back online after 5-30 sec.
In the case of nightly "Offline" is shown during this time.
https://drive.google.com/open?id=1fS21FKalXFpSkr6da3oBk28DTPGhgnpl (33 sec in general)

In PR nothing is shown, then "Connecting":
https://drive.google.com/file/d/1vzI3r9CPkzjddncuf7jFKf1zqe272gDd (26 sec in general)

hesterbruikman:
To me it seems the behavior of the PR is as desired, but there are definitely some improvements we can make in a new issue. I'd like to hash out the details of this new issue here.

Some observations, please correct me if any of these are wrong:
- The app is not considered `Offline` while it is booting and doesn't show `Offline` snackbar. I'm fine with that. It would be strange to see the app Offline while you are clearly connected to mobile data or wifi.
- `Connecting...` being shown for 5-30 seconds is one issue. The fact that I'm not sure what's going on during that time is another. `Connecting...` seems to capture two states in one label which is confusing. From the UI I don't know if the app has issues connecting with a Mailserrver or if the Mailserver is simply receiving a lot of traffic and therefore fetching messages is taking a long time.

@tbenr do you see ways to differentiate between `Connecting...` and `Fetching...` This is what the blue nightrider was responding to, but I think it's behavior was inaccurate / gave conflicting and too much information together with the snackbar.

So for Chat you would get:
- Offline (no Internet connection)
- Connecting (find and connect to Mailserver)
- Fetching (collecting messages from the Mailserver)

tbenr:
@hesterbruikman 

> @tbenr do you see ways to differentiate between `Connecting...` and `Fetching...` This is what the blue nightrider was responding to, but I think it's behavior was inaccurate / gave conflicting and too much information together with the snackbar.
> 
> So for Chat you would get:
> 
> * Offline (no Internet connection)
> * Connecting (find and connect to Mailserver)
> * Fetching (collecting messages from the Mailserver)

it is already differentiated but, as per my experience, most of he time is spent in "connecting" state and "connecting"+"fetching" state (don't know if\when the latter happens). Because of requirements by @errorists we don't want to see nightrider together with snackbar, so the only chance you have to see nightrider is when the fetching takes more than 1s ("fetching" state alone for more than 1s).

hesterbruikman:
> it is already differentiated but, as per my experience, most of he time is spent in "connecting" state and "connecting"+"fetching" state (don't know if\when the latter happens). Because of requirements by @errorists we don't want to see nightrider together with snackbar, so the only chance you have to see nightrider is when the fetching takes more than 1s ("fetching" state alone for more than 1s).

I see, so does the fact that the nightrider doesn't come up means that the fetching alone takes less than 1s?

tbenr:
@hesterbruikman exactly :)

src/status_im/init/core.cljs
```diff
@@ -21,13 +22,14 @@
...
   [{{:keys [view-id hardwallet initial-props desktop/desktop
-            supported-biometric-auth network/type]} :db}]
+            supported-biometric-auth network/type app-active-since]} :db}]
   {:db (assoc app-db
               :initial-props initial-props
               :desktop/desktop (merge desktop (:desktop/desktop app-db))
               :network/type type
               :hardwallet (dissoc hardwallet :secrets)
               :supported-biometric-auth supported-biometric-auth
+              :app-active-since (or app-active-since (datetime/timestamp))
```
flexsurfer: we have :now cofx

src/status_im/init/core.cljs
```diff
@@ -21,13 +22,14 @@
 (fx/defn initialize-app-db
   "Initialize db to initial state"
   [{{:keys [view-id hardwallet initial-props desktop/desktop
-            supported-biometric-auth network/type]} :db}]
+            supported-biometric-auth network/type app-active-since]} :db}]
```
flexsurfer: you can use `now :now`

tbenr: right.

src/status_im/init/core.cljs
```diff
@@ -21,13 +21,14 @@
 (fx/defn initialize-app-db
   "Initialize db to initial state"
   [{{:keys [view-id hardwallet initial-props desktop/desktop
-            supported-biometric-auth network/type]} :db}]
+            supported-biometric-auth network/type app-active-since now]} :db}]
```
flexsurfer: its not  in the db its in cofx , so you need to destructure it  `{{} :db now :now}`

tbenr: 😢

churik:
@tbenr  can I ask you to rebase your branch to current develop?

tbenr:
@churik yes, give me few minutes

tbenr:
@churik done, let's see how e2e goes

src/status_im/ui/components/connectivity/view.cljs
```diff
@@ -133,29 +157,84 @@
...
             window-width (reagent/atom 0)]
-    {:component-did-mount
-     (fn []
-       (when anim-translate-y
-         (if (:connected? status-properties)
-           (animation/set-value anim-translate-y neg-connectivity-bar-height)
-           (animation/set-value anim-translate-y 0))))}
-    (let [{:keys [loading-indicator?]} status-properties]
+    (let [loading-indicator? (:loading-indicator? current-ui-status-properties)
+          _ (propagate-status status-properties app-active-since current-ui-status-properties)]
```
flexsurfer: you must use react  lifecycle methods like `component-did-update`

flexsurfer (changes requested):

tbenr:
@flexsurfer i'd like to do some test around your review but latest rebase to dev always remains like that:
image

in logs i see:
`DEBUG [status-im.events:602] - :event-str {"type":"mailserver.request.completed","event":{"requestID":"0x088301241e6d472480e6b843687f054372f343ae65ec8e08be3c748a1e293b42","lastEnvelopeHash":"0x0000000000000000000000000000000000000000000000000000000000000000","cursor":"","errorMessage":"request is invalid: query range must be smaller or equal to 24 hours"}}`

errorists:
@tbenr you need to change your fleet settings from prod to eth.staging, learned that the hard way ;)

tbenr:
@errorists mercy :)

src/status_im/ui/components/connectivity/view.cljs
tbenr: @flexsurfer doesn't work very well: i need to run that code every time one of the letsub changes.
lifecycle fires only if view's props change, right?

flexsurfer: in theory it should work , but if not you could create new component and pass parameters

flexsurfer (approved):

churik:
@tbenr awesome work!
Checked:
- offline/online transition
- changing network state / reopening app
- background / foreground
- "retry to connect to mailserver" state
- @Serhy also checked mobile data popup and transition from offline to 3g network.

Thanks a lot!

PR #9572, opened 2019-12-03
Fix keycard account unlocking on disabled bioauth

rasom:
This commit fixes the case when biometric auth was disabled on the
device after being enabled for a keycard account.

fixes the case described here:
https://github.com/status-im/status-react/pull/9541#issuecomment-560472173
Steps:
- create multiaccount
- enable biometric authentication inside of status app
- relogin
- remove fingerprints in phone settings (in case of touch id)
- reopen app

**Expected result:**
- logged out, "Don't ask for card to sign in" is disabled, can login with card

status: ready

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -325,10 +325,13 @@
                        :save-password? true)}
        (navigation/navigate-to-cofx :progress nil)
        login)
-      (navigation/navigate-to-cofx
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

vitvly (approved):

churik:
Tested with ordinary multiaccount and multiaccount, created with keycard.
Can login if delete all biometric data and can't enable biometric auth until add them again.

PR #9574, opened 2019-12-04
Fix recovery with seed after visiting keycard recovery

rasom:
Steps to reproduce in `develop`:

- open "Recover with Keycard" screen
- go back
- open "Enter see phrase" screen and recover account
- see the next screen

status: ready


flexsurfer:
i've also seen this screen randomly when creating new regular account instead of welcome screen, dunno how to reproduce, probably when click next twice

flexsurfer (approved):

cammellos (approved):

PR #9579, opened 2019-12-04
Upgrade Xcode to 11.2.1

jakubgs:
One of the CI hosts - `macos-03` - has been upgraded to OSX 10.15.1 and Xcode to `11.2.1`.

To accommodate this upgrade this PR updates the host labels used by CI jobs, and also adds `fastlane` binary to the Nix shell for building the iOS app.

Once this is merged I will upgrade the other two MacOS hosts to Xcode `11.2.1`.

### DIFF
--- a/nix/mobile/ios/default.nix
+++ b/nix/mobile/ios/default.nix
@@ -1,4 +1,4 @@
-{ config, stdenv, stdenvNoCC, callPackage, mkShell,
+{ config, stdenv, stdenvNoCC, callPackage, mkShell, fastlane, cocoapods,
   xcodeWrapper, mkFilter, fetchurl, flock, nodejs, bash, zlib, procps,
   status-go, projectNodePackage }:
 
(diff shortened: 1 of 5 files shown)


flexsurfer (approved):

nix/mobile/ios/default.nix
```diff
@@ -27,23 +27,19 @@ let
...
-    buildInputs = [ xcodeWrapper ];
-  };
-
-in {
-  inherit xcodeWrapper;
+  selectedSources = [ fastlane status-go ];
 
   buildInputs = unique ([
+    cocoapods
+    fastlane.drv
```
pedropombeiro, line 34: Can't say I like including `fastlane` here when it should already be taken care of by sitting on `selectedSources`. One way around this would be to have `fastlane` inherit `buildInputs` from `fastlane.drv` so as to expose that here implicitly, but that's not perfect either.

nix/mobile/ios/default.nix
```diff
@@ -27,23 +27,19 @@ let
...
     xcodeWrapper
     flock # used in reset-node_modules.sh
     procps
   ] ++ catAttrs "buildInputs" selectedSources);
+
   shellHook = ''
     ${status-go.shellHook}
+    ${fastlane.shellHook}
```
pedropombeiro, line 42: This sounds like we're basically manually doing `concatString (catAttrs "shellHook" selectedSources)`

pedropombeiro (approved):

nix/mobile/ios/default.nix
jakubgs: Addressed in `nix-refactor` branch: 6a00c9a40c87c2325c9c0e900aa504d54695131f

nix/mobile/ios/default.nix
jakubgs: But inheriting `buildInputs` from `fastlane.drv` doesn't add `fastlane` to the shell.

PR #9588, opened 2019-12-05
Replace address with key-uid in multiaccs db

rasom:
Account's address was used as a primary key in accounts db and as a
deterministic id of an account in some API calls. Also it was used as a
part of the name of the account specific database. This revealed some
extra information about the account and wasn't necessary.
At first the hash of the address was planned to be used as a
deterministic id, but we already have a keyUid which is calculated as
sha256 hash of account's public key and has similar properties:
- it is deterministic
- doesn't reveal accounts public key or address in plain

status: ready


flexsurfer (approved):

yenda (approved):

rasom:
app crashes on recovery, checking what's wrong

vitvly (approved):

rasom:
tests passed on both `status-react` and `status-go`, merging

PR #9605, opened 2019-12-06
Major Nix refactor

jakubgs:
This is a __major__ nix refactor that changes mostly how we move stuff around and pass in Nix expressions, but the resulting shells should behave mostly the same.

Changes:
* Adds a new `nix-gc` Makefile target for removing old packages
* Moves all `nix/*.sh` files to `nix/scripts/*.sh` to make things more tidy
* Renames `TARGET_OS` into `TARGET` and makes it effective only with `nix/scripts/shell.sh`
* Renames `target-os` Nix argument to just `target` and makes it effective only with `shell.nix`
* Drops `IN_CI_ENVIRONMENT` env variable which was useless
* Drops use of `target-os` argument outside of `shell.nix` (with few exceptions, but just in naming)
* `nix/platform.nix` has been made obsolete and removed
* Moves the definition of all major targets to `nix/targets.nix`
* Moves the definition of all major shells to `nix/shells.nix`
* Makes `default.nix` and `shell.nix` just thin wrappers around `nix/default.nix`
* `nix/nixpkgs-bootstrap.nix` has been moved to `nix/pkgs.nix`
* All package and tool overrides have been moved to `nix/pkgs.nix`
* Explicit passing of contents of `pkgs` has been removed in favor of `callPackage` doing it for us
* `nix/bootstrapped-shell.nix` has been moved to `nix/tools/mkShell.nix`
* A new `mergeSh` tool has been added to `pkgs` from `nix/tools/mergeSh.nix`
  - This tool is used to merge shells created using `mkShell`
* `mobile/targets/jsbundle.nix` has been moved to `mobile/android/jsbundle/default.nix`
* Moves `status-go` version sanitization to `nix/status-go/utils.nix`, also fixes regex
* Renames `version` to `rawVersion` and `versionName` to `cleanVersion` in `status-go` derivation
* Ports `nix/mobile/ios/install-pods-and-status-go.sh` to Nix sub-shells
* Moves adjustment of `inotify/max_user_watches` out into `scripts/inotify_fix.sh`
* Makes iOS builds use the Nix version of Fastlane

And others... I may have went a bit overboard.

I don't expect everyone to read it, but at least test the main `make` targets you use.

Corresponding documentation update: https://github.com/status-im/status.im/pull/430

Status: READY


jakubgs:
Also, me creating this PR for over a week:
GOD DAMNIT

ci/common.groovy
```diff
@@ -43,28 +43,26 @@ def prep(type = 'nightly') {
...
+  if (['macos', 'linux', 'windows'].contains(env.TARGET)) {
     /* node deps, pods, and status-go download */
     nix.shell('scripts/prepare-for-desktop-platform.sh', pure: false)
-    sh('scripts/copy-translations.sh')
-  } else if (env.TARGET_OS != 'android') {
-    // run script in the nix shell so that node_modules gets instantiated before attempting the copies
-    nix.shell('scripts/copy-translations.sh chmod')
   }
+  /* run script in the nix shell so that node_modules gets instantiated before attempting the copies */
+  nix.shell('scripts/copy-translations.sh chmod', attr: "shells.${env.TARGET}")
```
pedropombeiro, line 65: Previously this was not running for Android (I assume because it was already handled in the end-to-end build), do we want to run it for Android now?

scripts/inotify_fix.sh
```diff
@@ -0,0 +1,23 @@
...
+  echo "inotify fix not applicable on non-linux OS"
+  return
+fi
+
+watches=$(cat /proc/sys/fs/inotify/max_user_watches)
+required_watches=524288
+
+if [ $watches -lt $required_watches ]; then
+  cecho "@b@cyan[[fs.inotify.max_user_watches limit is too low ($watches), increasing it]]"
+  echo fs.inotify.max_user_watches=$required_watches | sudo tee -a /etc/sysctl.conf
```
pedropombeiro, line 21: Are we OK with doing this on CI machines? I assumed we wanted to have all the configuration in Ansible.

nix/scripts/build.sh
```diff
@@ -2,17 +2,17 @@
...
 # cleanup for artifacts created during builds
 function cleanup() {
   # clear trapped signals
   trap - EXIT ERR INT QUIT
   # do the actual cleanup, ignore failure
-  if ./nix/clean.sh "${nixResultPath}"; then
+  if ${GIT_ROOT}/nix/scripts/clean.sh "${nixResultPath}"; then
     echo "Successful cleanup!"
-  elif [[ -n "${IN_CI_ENVIRONMENT}" ]]; then
+  elif [[ -n "${JENKINS_URL}" ]]; then
```
pedropombeiro, line 15: The idea of `IN_CI_ENVIRONMENT` was to avoid tying us with a specific CI provider, but I guess it's easy enough to adapt this if we ever move to a different CI provider.

nix/shells.nix
```diff
@@ -0,0 +1,65 @@
...
+  watchman = watchman-sh;
+  fastlane = fastlane-sh;
+  adb = targets.mobile.android.adb.shell;
+  # helpers for use with target argument
+  linux = targets.desktop.linux.shell;
+  macos = targets.desktop.macos.shell;
+  windows = targets.desktop.windows.shell;
+  android = targets.mobile.android.shell;
+  ios = targets.mobile.ios.shell;
+  # all shells together dependning on host OS
```
pedropombeiro:
```suggestion
  # all shells together depending on host OS
```

nix/tools/mkShell.nix
```diff
@@ -0,0 +1,40 @@
...
+      LANGUAGE="en_US.UTF-8";
+
+      shellSetup = ''
+        set -e
+
+        if [ -z "$STATUS_REACT_HOME" ]; then
+          export STATUS_REACT_HOME=$(git rev-parse --show-toplevel)
+        fi
+
+        export SHELL_SETUP=done
```
pedropombeiro, line 27: The way it was done previously, setup would not be run again unless you ran `make clean`. It seems that it is now run every time you run `make shell`. Am I missing something?

Is the idea that `scripts/setup` has shrunk so much that it's no longer an issue to run it every time?

scripts/add-nix-gcroots.sh
```diff
@@ -12,6 +12,6 @@ GIT_ROOT=$(cd "${BASH_SOURCE%/*}" && git rev-parse --show-toplevel)
 rm -rf .nix-gcroots
 mkdir .nix-gcroots
 
-drv=$(nix-instantiate --argstr target-os all --add-root ${GIT_ROOT}/shell.nix)
+drv=$(nix-instantiate --add-root ${GIT_ROOT}/shell.nix)
```
pedropombeiro: Why do we remove the `target-os all` here? At that point this seems pointless since we're not pinning anything meaningful.

nix/desktop/macos/default.nix
```diff
@@ -1,15 +1,14 @@
...
-
 let
-  inherit (stdenv.lib) concatStrings catAttrs;
+  inherit (lib) concatStrings catAttrs;
   baseImage = baseImageFactory "macos";
 
-in {
+in rec {
   buildInputs = [
     pkgs.gnupg22
```
pedropombeiro: I think you forgot to remove `pkgs.` here:
```suggestion
    gnupg22
```

nix/mobile/android/maven-and-npm-deps/maven/reactnative-android-native-deps.nix
```diff
@@ -1,4 +1,4 @@
-{ stdenvNoCC, fetchurl }:
+{ stdenv, fetchurl }:
```
pedropombeiro: What's the reason for switching from `stdenvNoCC` to `stdenv`? We don't really need an environment with a compiler for this derivation. I know that `stdenv` is mapped to `stdenvNoCC` at the top-level, but that's just circumstantial. By using `stdenvNoCC`, we're explicitly stating we don't need a compiler.

pedropombeiro (commented):
Can totally relate with that picture!

Noticed this when starting from scrash on macOS:

```
unpacking source archive /private/tmp/tmp-status-react-9c44c0f60/nix-build-status-go-fd49b01-source.drv-0/fd49b0140ebafdcec35b4da84685bcd8559a7dd9.tar.gz
Warning: Failed to set locale category LC_NUMERIC to en_CH.
Warning: Failed to set locale category LC_TIME to en_CH.
Warning: Failed to set locale category LC_COLLATE to en_CH.
Warning: Failed to set locale category LC_MONETARY to en_CH.
Warning: Failed to set locale category LC_MESSAGES to en_CH.
```

Likely not harmful, but wondering if we should lock the locales.

nix/mobile/default.nix
```diff
@@ -1,38 +1,42 @@
...
-      fastlane
-      status-go.android
-      status-go.ios
-    ] ++
-    optional platform.targetAndroid androidPlatform ++
-    optional platform.targetIOS iosPlatform;
+    fastlane
+    status-go.android
+    status-go.ios
+    androidPlatform
```
pedropombeiro, line 24: If I start a `make shell TARGET=ios` on a clean system, will that download Android stuff (it wasn't previously)?

nix/pkgs.nix
```diff
@@ -0,0 +1,38 @@
...
+      clojure = pkgs.clojure.overrideAttrs (old: rec { version = "1.10.0.411"; });
+      go = pkgs.callPackage ./patched-go { baseGo = pkgs.go_1_13; };
+
+      # custom builders
+      buildGoPackage = pkgs.buildGoPackage.override { inherit go; };
+    };
+  };
+in {
+  config ? { },
+  pkgs ? (import (fetchFromGitHub nixpkgsSrc)) { config = defaultConfig // config; },
```
pedropombeiro: I guess the top-level `let` actually belongs here. This looks a bit weird, not sure if putting the `let` here wouldn't be worse though. Not sure if there's a cleaner way to write this.

nix/status-go/default.nix
```diff
@@ -1,28 +1,24 @@
...
+  utils = callPackage ./utils.nix { inherit xcodeWrapper; };
+  gomobile = callPackage ./gomobile { inherit (androidPkgs) platform-tools; inherit xcodeWrapper utils buildGoPackage; };
   buildStatusGoDesktopLib = callPackage ./build-desktop-status-go.nix { inherit buildGoPackage go xcodeWrapper utils; };
   buildStatusGoMobileLib = callPackage ./build-mobile-status-go.nix { inherit buildGoPackage go gomobile xcodeWrapper utils; };
   srcData =
     # If config.status_go.src_override is defined, instruct Nix to use that path to build status-go
     if (attrByPath ["status_go" "src_override"] "" config) != "" then rec {
         owner = "status-im";
         repo = "status-go";
-        version = "develop";
```
pedropombeiro, line 21: No more `version` for local builds? I'm getting this when using the Go override path:

```
Running: nix-build --pure --fallback --no-out-link --show-trace --attr targets.mobile.android.release --arg config {status_go.src_override="/home/pedro/src/github.com/status-im/status-go";} --arg env {BUILD_ENV="prod";ANDROID_ABI_SPLIT="false";ANDROID_ABI_INCLUDE="armeabi-v7a;arm64-v8a;x86";} --argstr build-type nightly --argstr build-number 9999 --argstr keystore-file /home/pedro/.gradle/status-im.keystore --option extra-sandbox-paths /home/pedro/.gradle/status-im.keystore default.nix --keep TMPDIR,BUILD_ENV,STATUS_GO_SRC_OVERRIDE
error: while evaluating the attribute 'buildPhase' of the derivation 'status-react-build-release-android' at /home/pedro/src/github.com/status-im/status-react/nix/mobile/android/targets/release-android.nix:35:10:
while evaluating the derivation attribute 'name' at /nix/store/id4y2z7gvdb17nhmalqaf4ib11x34nfm-nixpkgs-source/pkgs/stdenv/generic/make-derivation.nix:191:11:
while evaluating the attribute 'version' at /home/pedro/src/github.com/status-im/status-react/nix/status-go/build-status-go.nix:22:5:
attribute 'version' missing, at /home/pedro/src/github.com/status-im/status-react/nix/status-go/default.nix:97:19
```

scripts/add-nix-gcroots.sh
jakubgs: But I did...

ci/common.groovy
jakubgs: I don't see where it is handled for android.

scripts/inotify_fix.sh
jakubgs: It is done in ansible:
https://github.com/status-im/infra-ci/tree/master/ansible/roles/linux-limits
I see no harm in doing it again.

nix/scripts/build.sh
jakubgs: It seemed like noise to me, and yes, it is trivial to switch.

nix/shells.nix
jakubgs: k

nix/tools/mkShell.nix
jakubgs: Where do you see it being run?

nix/desktop/macos/default.nix
jakubgs: true

pedropombeiro:
@jakubgs just added a commit of automatic changes done by the system on macOS. I'm also getting a build error with missing `Yoga-prefix.pch` on `make release-ios`. Will try `develop` next.

nix/mobile/android/maven-and-npm-deps/maven/reactnative-android-native-deps.nix
jakubgs: I guess I can switch it back to `a`, but I prefer to use what's standard, and there's not many cases of use of `stdenvNoCC` in `nixpkgs`.
Also if you look at the `nix/pkgs.nix` you'll see that we use `stdenvNoCC` for all shells:
https://github.com/status-im/status-react/blob/9c44c0f60b157334717e8b83b82668dc5937363e/nix/pkgs.nix#L21

nix/mobile/default.nix
jakubgs: Hmmm, it might, can you test that?

nix/pkgs.nix
jakubgs: Oh, you mean move the `config` as an argument to the top and make the `in` return just the `pkgs`? Makes sense.

scripts/add-nix-gcroots.sh
pedropombeiro: No sure I understand your answer.

nix/tools/mkShell.nix
pedropombeiro: Just add an `echo` here and you'll notice it gets output every single time you call `make shell`: https://github.com/status-im/status-react/blob/nix-refactor/nix/tools/mkShell.nix#L34

I guess currently it's not an issue since `setup` is lightweight, but if it were to become heavy again we'd have to fix this.

scripts/inotify_fix.sh
pedropombeiro: K

ci/common.groovy
pedropombeiro: Here: https://github.com/status-im/status-react/blob/dcae15ee0b5b5ffc318d1ea94f84c7796e6f471c/nix/mobile/android/maven-and-npm-deps/default.nix#L82-L84

nix/mobile/default.nix
pedropombeiro: I've tested it and only see `/nix/store/0nvv75l3zpw421l4v2yzwdy47cnngmp7-jsc_android___jsc_android_245459.0.0.tgz`. It's not too bad. There are 2 downsides with not knowing what we're targeting: unnecessary bandwidth(+ time) and unnecessary dependencies (e.g. if an Android dep becomes broken on macOS we break the iOS env too). We might just decide we prefer living with that than with the complexity of `target-os`.

pedropombeiro: At least I don't see the NDK being pulled, so I guess we're fine here. That would be a huge dep to pull unnecessarily.

nix/tools/mkShell.nix
jakubgs: But `nix/tools/mkShell.nix` is not calling `scripts/setup` at all.

ci/common.groovy
jakubgs: Oh, fair enough.

scripts/add-nix-gcroots.sh
jakubgs: But `target-os`(or now `target`) is irrelevant unless used with `/shell.nix`.

nix/tools/mkShell.nix
pedropombeiro: Ah I see, we're only calling it from `shell.sh` if the `nix` command isn't available. I guess that's fine then. This command needs updating then: https://github.com/status-im/status-react/blob/dcae15ee0b5b5ffc318d1ea94f84c7796e6f471c/nix/tools/mkShell.nix#L2

ci/common.groovy
pedropombeiro: I guess we can add a comment in the Groovy file explaining that

scripts/add-nix-gcroots.sh
pedropombeiro: And we are instantiating `${GIT_ROOT}/shell.nix`, so it would make sense to still pass `target all` to ensure we're pinning as many derivations as possible, so they don't get deleted by a garbage collection. If we pin `target none`, there'll be very few derivations pinned.

nix/tools/mkShell.nix
jakubgs: done

scripts/add-nix-gcroots.sh
jakubgs: Okay, fair enough, fixed.

nix/status-go/default.nix
jakubgs: I'm confused as to the purpose of both `version` and `versionName` it's kinda messy. Shouldn't those be combined into one? Or is there any reason not to?

nix/mobile/default.nix
jakubgs: I will have to think about this one tomorrow.

nix/status-go/default.nix
pedropombeiro: `version`: literally what is read from the json file
`sanitizedVersion`: same as `version` but replacing illegal path characters such as `/`
`versionName`: processed version name (`v` prefix removed, changed to `develop` if not a valid version)

I think it's good to capture the unprocessed `version` into the `passthru` map so that we know exactly what the input for the derivation was. Other than that we can use `versionName` (maybe with a better name).

yenda:
On Android I get this with `make run-android`:

```
error Failed to install the app. Make sure you have the Android development environment set up: https://facebook.github.io/react-native/docs/getting-started.html#android-development-environment. Run CLI with --verbose flag for more details.
Error: Command failed: ./gradlew app:installDebug -PreactNativeDevServerPort=8081

* Where:
Build file '/home/yenda/status-react/android/build.gradle' line: 48

* What went wrong:
A problem occurred configuring project ':@react-native-community_netinfo'.
> STATUS_GO_ANDROID_LIBDIR environment variable is not valid!

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3s

    at checkExecSyncError (child_process.js:629:11)
    at execFileSync (child_process.js:647:13)
    at runOnAllDevices (/home/yenda/status-react/node_modules/@react-native-community/cli-platform-android/build/commands/runAndroid/runOnAllDevices.js:94:39)
    at buildAndRun (/home/yenda/status-react/node_modules/@react-native-community/cli-platform-android/build/commands/runAndroid/index.js:158:41)
    at then.result (/home/yenda/status-react/node_modules/@react-native-community/cli-platform-android/build/commands/runAndroid/index.js:125:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Makefile:251: recipe for target 'run-android' failed
make: *** [run-android] Error 1
```

jakubgs:
@yenda was this on MacOS?

nix/status-go/default.nix
jakubgs: I've addressed this, please check.

nix/status-go/build-status-go.nix
```diff
@@ -19,7 +19,7 @@ let
   args = removeAttrs args' [ "buildMessage" ]; # Remove our arguments from args before passing them on to buildGoPackage
   buildStatusGo = buildGoPackage (args // {
     pname = repo;
-    version = "${version}-${strings.substring 0 7 rev}-${host}";
+    version = "${rawVersion}-${strings.substring 0 7 rev}-${host}";
```
pedropombeiro: I think you need to use the sanitized version here @jakubgs otherwise Nix will fail when creating the drv file.

nix/status-go/default.nix
```diff
@@ -42,20 +39,14 @@ let
...
       in rec {
         inherit (versionJSON) owner repo version;
-        sanitizedVersion = sanitizeVersion versionJSON.version;
-        goPackagePath = "github.com/${owner}/${repo}";
         rev = versionJSON.commit-sha1;
         shortRev = strings.substring 0 7 rev;
-        versionName = if (builtins.match versionRegex sanitizedVersion) != null
-                      then removePrefix "v" sanitizedVersion # Geth forces a 'v' prefix
-                      else "develop"; # to reduce metrics cardinality in Prometheus
+        rawVersion = versionJSON.versionm;
```
pedropombeiro:
```suggestion
        rawVersion = versionJSON.version;
```

nix/status-go/build-status-go.nix
jakubgs: right, fixed

yenda:
@jakubgs no Linux

nix/mobile/default.nix
jakubgs: I think it looks correct:
```
jenkins@macos-01.ms-eu-dublin.ci.misc:~/repos/status-react % make shell TARGET=ios
Configuring default Nix shell for target 'ios'...
Checking for modifications in node_modules...
No modifications detected.
Checking for modifications in ios/Pods...
No modifications detected.

[nix-shell:~/repos/status-react]$ echo $buildInputs | tr ' ' '\n'
/nix/store/w6l0vw6yzss9lc8cvxzyb8z2ajhhwi9g-xcode-wrapper-11.2.1
/nix/store/d6dg4pl5laprgchzhzhx58z0bj91pbbj-watchman-4.9.0
/nix/store/1lnv10pvii0s8snmfw2snb4dv4rnj5xr-bundler-1.17.3
/nix/store/x6hwh3h408k4c28sir7askm3fqpl4hb0-procps-1003.1-2008
/nix/store/bjpxsdfjqd51mglpsl9pz1gdw7a4fpz5-flock-0.2.3
/nix/store/x5pzk8c516g6k46gxq3v8jxipgg6vvlf-status-go-0.36.1-fd49b01-ios
/nix/store/vp6r9vqbf5pig2avhacwz7hiw4xwcd39-ruby-2.6.5
/nix/store/1lnv10pvii0s8snmfw2snb4dv4rnj5xr-bundler-1.17.3
/nix/store/zf6hzp1xn20fmf23s6k1vc6ig6s65xqa-fastlane-2.133.0
/nix/store/8627xzqkrf5cp23a8g4q9s87aa8h2idb-curl-7.67.0-bin
/nix/store/iyv7b8781cqh0r4v5fxzgnyfvqbjmdcw-cocoapods-1.8.4
/nix/store/vp6r9vqbf5pig2avhacwz7hiw4xwcd39-ruby-2.6.5
/nix/store/1lnv10pvii0s8snmfw2snb4dv4rnj5xr-bundler-1.17.3
/nix/store/zf6hzp1xn20fmf23s6k1vc6ig6s65xqa-fastlane-2.133.0
/nix/store/09jcdiv7fqj2jx4r4vpxbr5nrr8rfzaa-curl-7.67.0-dev
/nix/store/iyv7b8781cqh0r4v5fxzgnyfvqbjmdcw-cocoapods-1.8.4
/nix/store/iyv7b8781cqh0r4v5fxzgnyfvqbjmdcw-cocoapods-1.8.4
/nix/store/5pp98fn44axh3gdl68naba8ryak0bh3p-bash-4.4-p23-dev
/nix/store/09jcdiv7fqj2jx4r4vpxbr5nrr8rfzaa-curl-7.67.0-dev
/nix/store/9a5a676m264kbbd0hpwgl3bkxs2b56mx-wget-1.20.3
/nix/store/0lzhngrf5dsa4xrrz6ssz7qganb6cm02-file-5.37
/nix/store/15wng30fs9nbzfd2xcsdypi0qip8m1s8-unzip-6.0
/nix/store/bjpxsdfjqd51mglpsl9pz1gdw7a4fpz5-flock-0.2.3
/nix/store/90g2i8jcalvngb7hj1a1vk57vjbmdzb4-git-2.24.0
/nix/store/6ndn9v9id1j2gdbndi7x0bh53s0h7l7m-gnumake-4.2.1
/nix/store/2865ahxs2sj43vlsmzw7zcndm1fcxjbd-jq-1.6-dev
/nix/store/25fjxm795r5fn4pby15lhibh3gb4mf6b-ncurses-6.1-20190112-dev
/nix/store/0dj2sd7c6h3kn8ik9i3kcm0x8i4zarjb-clojure-1.10.0.411
/nix/store/1k2xnwi4nfn0q7j518aq80zk1vf8zys6-leiningen-2.9.1
/nix/store/y2vd5zk2pn3j36kkjdfsllxw10z2rccb-apache-maven-3.6.2
/nix/store/d6dg4pl5laprgchzhzhx58z0bj91pbbj-watchman-4.9.0
/nix/store/a7az4bx32yy968nw8c3ppsqdv0kv3mfd-yarn-1.19.1
/nix/store/dfjphxv0jbzi7fg9h6l5q6ri15hz2skz-nodejs-10.17.0
/nix/store/ylvwym6krs4wz6lxdpr458bpjpr6s08d-python-2.7.17
/nix/store/iyv7b8781cqh0r4v5fxzgnyfvqbjmdcw-cocoapods-1.8.4
/nix/store/60hjqrb8dn47jrffgcdzn85bz3bzbp24-clang-wrapper-7.1.0
```

jakubgs:
@yenda @PombeirP I think I fixed all the issues, please check it out.

pedropombeiro:
I tested both release and debugging targets for iOS and it worked fine for me.

pedropombeiro (approved):

nix/status-go/default.nix
jakubgs: I think you might be looking at an old verison of the diff, this has already been fixed.

jakubgs:
Successful `nix-cache` jobs on this branch:

* https://ci.status.im/job/status-react/job/nix-cache/job/linux/637/
* https://ci.status.im/job/status-react/job/nix-cache/job/macos/606/

jakubgs:
And a documentation update: https://github.com/status-im/status.im/pull/430

PR #9629, opened 2019-12-10
[#9575] Make sure status supports EIP 1102 and EIP 1193

### LINKED ISSUE #9575: Make sure status supports EIP 1102 and EIP 1193
https://medium.com/metamask/breaking-changes-to-the-metamask-inpage-provider-b4dde069dd0a

1) In Metamask, the provider by default has an empty array of accounts, which will be populated if the user gives access to the dapp, while Status' provider returns null. This causes Status to raise this error: "The requested account has not been authorized by the user." in this code block 

It is causing problems in Embark because it uses getAccounts a lot to determine if a dapp is connected or not.

2) web3.eth.getCoinbase() returns default account , should return selected account for dapps

flexsurfer:
fixes #9575

tested : CK, airswap, uniswap

for QA: test all most popular dapps from dap.ps


churik:
Tested (mostly just refused transactions, testing is too expensive in these apps):
- Total swap
- Dragonereum
- Maker Dao
- Ethereum name service
- Yang Dao
- dap.ps (upvoting)
- all possible actions in Status Test Dapp (on Ropsten)
- revoking access for dapps manually
- switching accounts in dapps

@flexsurfer please, take my commit while squashing

rasom (approved):

PR #9631, opened 2019-12-10
disable dependabot for desktop js modules

jakubgs:
Since desktop work is on-hold and the bot is noisy.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -20,17 +20,3 @@ update_configs:
         update_type: security
     default_reviewers:
     - "jakubgs"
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

pedropombeiro (approved):

pedropombeiro:
Maybe we can just change the update schedule to `never`? Need to check if that's a valid value.

pedropombeiro:
Hmm, doesn't seem so: https://dependabot.com/docs/config-file/#update_schedule-required

Approved.

pedropombeiro (approved):

PR #9635, opened 2019-12-10
Add SNT/STT to defaults

### LINKED ISSUE #9493: No SNT by default after multiaccount creation / restoring

acolytec3:
fixes #9493

Summary
SNT is not shown on Wallet home page as a default. This was previously the case but probably lost with the change from account to multiaccount. This adds them back to default account settings so they appear when a new multiaccount is retrieved/created.

Platforms
Android
iOS
Functional
wallet / transactions
Steps to test
-Open Status

Create/restore a multiaccount
Tap Wallet home page
Verify that SNT is displayed beneath ETH by default
status: ready

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -82,7 +82,8 @@
 (def default-multiaccount-settings
   {:web3-opt-in?     true
    :preview-privacy? false
-   :wallet           {:visible-tokens {}}})
+   :wallet           {:visible-tokens {}
+                      :first-login true}})
 
 (def currencies
   {:aed {:id :aed :code "AED" :display-name (i18n/label :t/currency-display-name-aed) :symbol "د.إ"}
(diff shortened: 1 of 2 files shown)


acolytec3:
@churik @flexsurfer @yenda Here's the reboot of #9508.  Should be ready for validation.  One scenario I can't test is the multiaccount on mainnet with balances in tokens other than SNT to verify that SNT is added by default on first login.

acolytec3:
Do any of these failed tests appear to reflect issues with the code? They don't seem to but I'm not 100% sure with the second one.

churik:
@acolytec3 thank you for your contribution!
issues are not related to this PR.
Anyway, error that assets are not shown after restoring multiaccount with tokens are still in place.

Probably it is not worth to continue with it for now, WDYT @flexsurfer?

flexsurfer:
yeah, thanks for your effort and kudos @acolytec3 let's postpone this for now

PR #9640, opened 2019-12-11
move custom tokens to a separate table

yenda:
As part of rfc-002, `making migrations more robus`, this PR moves the custom tokens to a separate DB.

status: ready


src/status_im/multiaccounts/login/core.cljs
```diff
@@ -249,7 +253,7 @@
                                              :default-mailserver true})
               (chaos-mode/check-chaos-mode)
               (when-not platform/desktop?
-                (initialize-wallet)))))
+                (initialize-wallet nil)))))
```
rasom, line 256: we need to extend that macro for multi rarity support :)

rasom (approved):

PR #9656, opened 2019-12-12
[#9649] Add "Send" button and a QR icon on the wallet screen

### LINKED ISSUE #9649: Add "Send" button and a QR icon on the wallet screen
Problems:
1. It was hard to send a transaction by scanning a QR code. The user journey was too long, even though the use case is quite common;
2. It was not that long of a user journey to send a transaction, but still too long and it feels like adding this "send" floating button is easy and will improve the low a lot

Screen Shot 2019-12-12 at 12 36 42

Figma https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=1743%3A303

flexsurfer:
Gamechanger

fixes #9649

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -443,11 +443,10 @@
    public-key))
 
 (re-frame/reg-sub
- :multiaccount/default-address
+ :multiaccount/default-account
  :<- [:multiaccount]
  (fn [{:keys [accounts]}]
-   (ethereum/normalized-hex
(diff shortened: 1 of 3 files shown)


churik:
Tested on EIP links with ENS names, tokens, ETH.
Also checked error handling in case of invalid address / network / link.

Thx @flexsurfer!

rasom (approved):

PR #9673, opened 2019-12-17
do not store mnemonic on recovered accounts

yenda:
mnemonic was saved in the db for recovered account and always present in the app-db without possibility to delete it

- remove seed-backed-up? boolean and use mnemonic presence instead, so that this bug cannot be reintroduced without consequences in the UI
- saving mnemonic is now a param that has to be set to true by default mnemonic is not saved when creating an account to reduce the risk of reintroducing that bug

## Tested on Android

- created account, seed phrase backed up 
- recovered account


cammellos (approved):
we could add some tests  to verify the behavior

yenda:
@cammellos I changed the PR so that now the UI asking to back up the seedphrase is directly related to the presence of the mnemonic. Also by default on-account-created doesn't save the mnemonic anymore so that saving mnemonic is a setting that has to be set as true.

flexsurfer (approved):

rasom (approved):

yenda:
@churik this test failure isn't related either?

churik:
no, it is known issue

PR #9731, opened 2019-12-27
[#9724] Status is not usable after relogin if add watch-only account …

### LINKED ISSUE #9724: Status is not usable after relogin if add watch-only account that matches address of one of the created accounts
# Bug Report

## Problem
If add watch-only account that matches address of Status account (or any account that is already created in your multiaccounts) it is not possible to login again or to create new multiaccount or login to existing one

#### Expected behavior
Can't add watch-only account with same address

#### Actual behavior
photo_2019-12-23 16 37 21

### Reproduction

- Open Status
- Create multiaccount
- Open wallet -> copy address of status account
- Add watch-only account that matches the address of status account

### Additional Information

- Status version: nightly 23/12/2019

#### Logs

[no-rows-in-result-set.log](https://github.com/status-im/status-react/files/3995455/no-rows-in-result-set.log)

flexsurfer:
fixes #9724

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -526,6 +526,14 @@
  (fn [accounts]
    (filter #(not= (:type %) :watch) accounts)))
 
+(re-frame/reg-sub
+ :add-account-disabled?
+ :<- [:multiaccount/accounts]
+ :<- [:add-account]
+ (fn [[accounts {:keys [address]}]]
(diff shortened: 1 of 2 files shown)


vkjr (approved):

churik:
Validation is added - can't add an account if it was already added (checked that it is not case-sensitive).

PR #9733, opened 2019-12-27
[#9493] No SNT by default after multiaccount creation / restoring

### LINKED ISSUE #9493: No SNT by default after multiaccount creation / restoring

flexsurfer:
fixes #9493

Show SNT by default, and fetch only balances for visible tokens after tokens initialization


churik:
Tested on ordinary and KK (fresh and recovered) accounts with tokens and with empty balances (IOS 13, Android 8)

Thanks @flexsurfer, amazing work!

churik (approved):

PR #9734, opened 2019-12-27
Test reply to messages in chats

Serhy:
Test replies are received in public and 1-1 chats. Check username in reply text snippet in message box corresponds to username message being quoted ("You" in a case you reply to yourself)


churik (approved):
👍

PR #9739, opened 2020-01-02
Fix tabbar notification badge position

### LINKED ISSUE #9436: Message notification badges are aligned incorrectly in tab bar
unknown
as you can see on the image above, the badge is offset horizontally to the right, appearing detached from the icon it belongs to. I believe this was introduced here https://github.com/status-im/status-react/pull/9340

#### Expected behavior

Badge is placed 16px to the right of the left icon edge and 6px up from the bottom edge or alternatively directly with its top edge.

Screenshot 2019-11-12 at 11 49 09

#### Actual behavior

Badge is placed somewhere else

#### Acceptance Criteria

Badge is placed correctly and grows to the right with two digit numbers, so it's X position stays the same.

[Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=16%3A357)

Ferossgp:
Fix: #9436 
Screenshot 2020-01-02 at 14 55 30
Screenshot 2020-01-02 at 14 56 17
Screenshot 2020-01-02 at 14 56 01

### DIFF
--- a/src/status_im/ui/components/badge.cljs
+++ b/src/status_im/ui/components/badge.cljs
@@ -9,9 +9,12 @@
                  {:height 18 :border-radius 9 :min-width 18 :padding-horizontal 6}
                  {:height 22 :border-radius 11 :min-width 22 :padding-horizontal 8})
                {:background-color colors/blue
-                :justify-content :center
-                :align-items :center})
-   [react/text {:style {:typography :caption :color colors/white}} label]])
+                :justify-content  :center
+                :align-items      :center})
+   [react/text {:style {:typography  :caption
+                        :font-weight "500"
+                        :color       colors/white}}
+    label]])
 
 (defn message-counter [value & [small?]]
   [badge
--- a/src/status_im/ui/components/tabbar/styles.cljs
+++ b/src/status_im/ui/components/tabbar/styles.cljs
@@ -46,9 +46,13 @@
    :top      0
    :position :absolute})
 
+;; NOTE: Extra padding to allow badge width to be up to 42 (in case of 99+)
+;; 42 Max allowed width, 24 icon width as per spec, 16 left pos as per spec.
+(def ^:private message-counter-left (+ (/ (- 42 24) 2) 16))
+
 (def message-counter
-  {:right    -10
-   :top      0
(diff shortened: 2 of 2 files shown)


src/status_im/ui/components/badge.cljs
```diff
@@ -9,9 +9,12 @@
                  {:height 18 :border-radius 9 :min-width 18 :padding-horizontal 6}
                  {:height 22 :border-radius 11 :min-width 22 :padding-horizontal 8})
                {:background-color colors/blue
-                :justify-content :center
-                :align-items :center})
-   [react/text {:style {:typography :caption :color colors/white}} label]])
+                :justify-content  :center
+                :align-items      :center})
+   [react/text {:style {:typography  :caption
+                        :font-weight "500"
```
flexsurfer, line 15: `font-weight` is set in typography, so no need to set it here

flexsurfer (approved):

src/status_im/ui/components/badge.cljs
Ferossgp: Yes, you are right, but in typography `:caption` is merged with `default-style` which has `font-weight` 400. Because caption is used in other places, I decided to make change locally to not break other specs. Maybe I can create another typography style for that?

flexsurfer: @errorists wdyt?

flexsurfer: but if it should be by design to use :caption and "500" then its fine

churik:
Looks good to me.
@errorists can you please take a look?

src/status_im/ui/components/badge.cljs
errorists: I think it's up to you guys, how you prefer to implement text styles. tbh these aren't gospel and on occasion we have to use local overrides that don't have a counterpart in the typography stylesheet. the result is correct as in `font-size: 12` and `font-weight: 500`, I certainly appreciate looking up correct values in Figma, thanks!

flexsurfer:
@Ferossgp pls, rebase, squash and sign commits

errorists (approved):

PR #9741, opened 2020-01-02
Add request address command

cammellos:
This PR adds the request-address command as described in the specs.
The flow is:

1) User A prepares a transaction to send to user B, requesting user B their address
2) User B should see a message with the transaction details, and they can either accept or decline the request. If they accept they will select an account.
3) User A will receive the response and send back a transaction, which will need to be signed
4) User B will then validate the transaction, and only once validated it will be shown to the user.

Currently the validation is done on Ropsten for testing purposes, so make sure both accounts are using the ropsten network before testing.

Known issues:
1) Select account sheet animates once the account is selected, after clicking on accept request address
2) Token transfer have not been tested (It would be handy if someone could test those on ropsten)
3) Declining a transaction does not hide the previous message

Likely there are other issues here and there, if testing, please let me know.

status: ready


cammellos:
@rachelhamlin requesting a transaction has also been implemented if you'd like to give it a go

src/status_im/commands/core.cljs
```diff
@@ -0,0 +1,38 @@
+(ns status-im.commands.core
+  (:require
+   [re-frame.core :as re-frame]
+   [status-im.ethereum.json-rpc :as json-rpc]
+   [status-im.utils.fx :as fx]))
+
+(fx/defn handle-prepare-accept-request-address-for-transaction
+  {:events [:commands/prepare-accept-request-address-for-transaction]}
```
yenda: use `::` and then alias the events `::commands` where they are used so that you have explicit usage of that namespace and there is no need to add the require in `events` which could get accidentally refactored away

yenda:
UX bug turned out to be quite a rabbit hole, while implementing link to transaction I realize that the lack of pending transaction in history was actually due to a bug, which I fixed, then while testing it it turned out that there was also a bug in send transaction drawer when trying to select one of your own account as recipient.

yenda:
@churik @Serhy this is ready for testing

Serhy:
With the 14th builds in jenkins on both iOS and Android for the accounts switched to Ropsten network and created 1-1 chat between each other:

### Issue #1
Similar to above Issue1: 
Send `Send transaction` command to user UserB -> UserB taps `Accept and share address` (Select account bottom sheet appears) -> tap Status account and tap any of Status account / another account in the list. 
**Actual result:** bottom sheet does not collapse giving no ability to select-and-confirm anything apart Status account (I can pull down and proceed with default Status account selected but that's anyway not the case)

### Issue #2
Can't Cancel bottom-sheet which appears after `Accept and share address` tapped by tapping the `Cancel` button.

Serhy:
### Issue 3
There are issues with keyboards appearance when user taps `Sign and send`
**Preconditions**:
1) UserA (iOS) and UserB (Android) create/recover and account with positive ETH/token balance and switch to Ropsten
2) UserA and UserB join the same 1-1 chat
3) Both UserA and UserB send `Send transaction` (ETH or token) to each other
4) Both UserA and UserB confirm address
5) Both UserA and UserB navigate back to Chats view

#### Scenario #1 (When dialpad on Android appears)
6) UserB (Android) opens 1-1 chat
7) UserB taps `Sign and send`
**Actual result:**  
Screenshot 2020-01-09 at 12 39 47

#### Scenario #2 (When keyboard on Android appears)
6) UserB (Android) opens 1-1 chat
7) UserB taps inside message input, so the keyboard appears
8) UserB taps `Sign and Send`
**Actual result:**  
Screenshot 2020-01-09 at 12 42 17


#### Scenario #3 (When keyboard on iOS keep staying overlapping Send tranasaction bottom-sheet)
6) UserA (iOS) opens 1-1 chat
7) UserA taps inside message input, so the keyboard appears
8) UserA taps `Sign and Send`
**Actual result:**  
Screenshot 2020-01-09 at 12 45 27
9) However, after I put Status on iOS on background and return back it to foreground it becomes like in Scenario2 above with Android (so Send transaction with keyboard below)


#### **Expected results:**
I think expect results for above would be: when `Sign and send` button is pressed we a) hide/close any keyboard/dialpad and then call for `Send transaction` bottom sheet. So `Send transaction` always appears without keyboard/dialpad

yenda:
@Serhy build 17 should contain fixes for all your issues, also I switched to ropsten instead of mainnet so you can do all the tests you want. build 16 is the same but mainnet only

src/status_im/node/core.cljs
```diff
@@ -134,6 +134,8 @@
               :InstallationID installation-id
               :MaxMessageDeliveryAttempts config/max-message-delivery-attempts
               :MailServerConfirmations  config/mailserver-confirmations-enabled?
+              :VerifyTransactionURL "https://ropsten.infura.io/v3/f315575765b14720b32382a61a89341a" #_"https://mainnet.infura.io/v3/f315575765b14720b32382a61a89341a"
+              :VerifyTransactionChainID 3 #_1
```
flexsurfer: whats this?

src/status_im/signing/core.cljs
```diff
@@ -202,13 +206,47 @@
     (when (and (not in-progress?) (seq queue))
       (show-sign cofx))))
 
+(fx/defn send-transaction-message
+  {:events [::send-transaction-message]}
+  [cofx chat-id transaction-hash signature]
+  {::json-rpc/call [{:method "shhext_sendTransaction"
+                     :params [chat-id transaction-hash (:result (types/json->clj signature))]
+                     :on-success #(re-frame/dispatch [:transport/message-sent % 1])}]})
```
flexsurfer, line 213: what does 1 mean ?

churik:
### ISSUE 4: Can't chat when users are on different networks (i.e. Ropsten and mainnet)

**Steps:**
1) device A: install status
2) device A: create new multiaccount
3) device A: join public chat (on mainnet), send some messages
4) device B: repeat steps 1-3 on Ropsten network

**Expected result:**  can see whole history
**Actual result:**  no messages sent from Ropsten on Mainnet and vise versa

src/status_im/subs.cljs
```diff
@@ -2038,6 +2062,33 @@
...
+(re-frame/reg-sub
+ :wallet.request/prepare-transaction-with-balance
+ :<- [:wallet/prepare-transaction]
+ :<- [:wallet]
+ :<- [:offline?]
+ :<- [:wallet/all-tokens]
+ :<- [:ethereum/chain-keyword]
+ (fn [[{:keys [symbol from to amount-text] :as transaction}
+       wallet offline? all-tokens chain]]
+   (println "TRANSACTION" transaction)
```
flexsurfer: leftover?

src/status_im/ui/screens/chat/extensions/views.cljs
```diff
@@ -37,7 +37,7 @@
                                   :height    panel-height
                                   :transform [{:translateY bottom-anim-value}]
                                   :opacity   alpha-value}}
-     [react/view
+     [react/view {:style {:flex-direction :row}}
```
flexsurfer, line 40: why row ? @errorists

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -201,6 +203,233 @@
...
                  :source {:uri (contenthash/url (-> content :sticker :hash))}}]])
 
+(defn- final-status? [command-state]
+  (or (= command-state constants/command-state-request-address-for-transaction-declined)
+      (= command-state constants/command-state-request-transaction-declined)
+      (= command-state constants/command-state-transaction-sent)))
+
+(defn- command-pending-status
+  [command-state direction to transaction-type]
+  [react/view {:style {:flex-direction :row
```
flexsurfer, line 216: could be moved to styles ns

src/status_im/node/core.cljs
cammellos: we use that to validate transaction, it should always point to mainnet as we don't validate ropsten transactions, so it's not the same as the network the user is in

src/status_im/signing/core.cljs
cammellos: 1 means that only 1 confirmation is needed (as opposed to `n` for group chats), it's an artifact of when sending was mostly in status-react, we will remove it at some point but still some work to do

src/status_im/node/core.cljs
yenda: though once we support multiple networks it can probably go away

src/status_im/ui/screens/chat/extensions/views.cljs
yenda: @errorists can tell if that's a good idea or not, basically I kept finding it super weird to have only 2 commands that are aligned vertically instead of horizontally, in the designs they are on top of each other but there is also more commands. So I thought we should align them like this until there is more

Serhy:
### Issue #4 
Status crashes when navigating to 1-1 chat with sent or received `Send transaction`/`Receive transaction` command message with STT. Apparently it happens with tokens having no custom icon. Well not all tokens with default icon behaves like that, but it's the only pattern I see when sending request command with ~10 different tokens. Appears for STT and all fine for all other tokens in Ropsten. Also if I'm on Mainnet then tokens with default icon sent in `Request` command (ADX and TGT) does not crash app but VRS does (and several other tokens with non-default icons behave fine)

**To reproduce:**
1) Create a new account and enable STT token on Ropsten (or recover account having STT already)
2) In 1-1 chat send the `Send transaction` / `Request transaction` with STT token selected.
**Actual result:**  App crashes 
Screenshot 2020-01-09 at 15 09 44

Logcat with exception stacktrace: 
[Status.log](https://github.com/status-im/status-react/files/4040500/Status.log)

churik:
### ISSUE 5: Messages from device with custom time are not sent immediately

I think it may be important because issue is not reproducible on current nightly. 
So messages from devices with custom time will be sent later (if time is behind).
So if you have device time -5 mins and send message from PR to PR build - it will be received only after 5 minutes.

**Steps:**
- Device A: install status
- Device A: set custom time
- Device A: create new multiaccount
- Device A: join public chat 
- Device B (with automatic time): create multiaccount, join the same public chat
- Device A: send message and check it on other device

**Expected result:**  message is received on Device B
**Actual result:**  message is not received on Device B (it will appear on Device B after 5 mins)
cc @cammellos

cammellos:
@churik can you replicate consistently? are messages on device A marked as "Sent"? 
thanks

churik:
@cammellos, yes, consistently from PR build.
Yes, they are marked as "Sent"

yenda:
@Serhy @churik 4 should be fixed

cammellos:
@churik @Serhy issue 5 should also be fixed, I haven't had time to test the build, but if you manage to test it let me know, thanks!

Serhy:
All above issues are fixed but **Issue 3:** Scenario1 and Scenario2 on Android 8.1 (Samsung J7) (can't reproduce with Android 9 though)

### Issue 6
Custom token displayed as `ERC20` in Send/Request if I just added it to the list of tokens for the first time. However if I relogin, then I see token recognised in the command message. Can't Sign the transaction though, because of 'Insufficient funds' error (even my address has enough funds) - that's might be due to #9792 which might be the root cause of this 'Issue6'

**Update**: in latest build here difference between develop for #9792 issue is that here custom token kept on `Wallet` -> `Manage Assets` -> `Assets` list  

**To reproduce:**
2) Add custom token (for Ropsten it may be `0x101848D5C5bBca18E6b4431eEdF6B95E9ADF82FA` - WEENUS token) 
3) Send `Send transaction` or `Request transaction` with custom token
**Actual result**: ERC20 is displayed in command message

Screenshot 2020-01-10 at 15 42 14

### Issue 7 
We need to reflect testnet ETH as we do in Wallet: i.e. ETHro/ETHri. It's now displayed as `ETH` which may have consequences which is described in next Issue8

### Issue 8 
Commands view in different network. Not sure if it's part of this PR but what happens is that commands I sent from Ropsten looks on Mainnet as `ERC20` for all tokens and `ETH` recognised as mainnet ETH (fiat conversion applied to it)

Screenshot 2020-01-10 at 15 45 32

Serhy:
Tested https://i.diawi.com/3KL5uB, 
and http://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200110-154443-c4e801-pr-universal.apk
Agreed to skip Issue8 for now. 
Issue9 and Issue 3: Scenario1 and Scenario2 to have a separate GHI.

1) Can send tx with ETH (0 , 1e-18 ETH, max I have)
2) Can send tx with tokens including custom token
3) Can request with token and reply to it
4) Can request with ETH and reply to it
5) Sign phrase appears when no wallet onboarding set up done 
6) Fiat values calculated accordingly to amount
7) Check status (Shared -> Pending -> Confirmed) on both recipient and sender
9) Changed name of shared account reflected in previous messages - like for that!
10) Messaging in different networks and with different custom time works

Lets merge, @cammellos @yenda (just to fix conflict). Kudos to you! 
Rest minor issues and known issues we may fix separately.

yenda (approved):

PR #9743, opened 2020-01-03
New searchbar design

### LINKED ISSUE #7697: Search bar animation behavior
# Problem
The Status Home tab (soon to be Chat) has a searchbar implemented. This searchbar allows to find chats in the chat list and will in the near future also enable finding messages. 

The searchbar was originally intended to behave like: https://www.dropbox.com/s/xyf77rl9jvesh1y/ScreenFlow%20copy%203.mp4?dl=0

For the most part this behavior has been implemented. There are a few issues that remain:

* Search bar does not appear second (or more) times after pull Home screen down gesture been done while Search bar is present on screen: https://github.com/status-im/status-react/issues/7676
* It's cumbersome to hide the searchbar during inactivity, after it's been pulled down in view: https://github.com/status-im/status-react/issues/7643
* More detailes on expected behavior https://github.com/status-im/status-react/issues/7643#issuecomment-468648407

To fix these issues the implementation needs to be adapted and potentially the searchbar needs to be reimplemented in full. 

# Implementation
[@yenda please edit]

# Acceptance Criteria
Implementation solves issues above and behaves like example video when it comes to appear/disappear behavior.

# Notes
# Future Steps

vkjr:
fixes #7697

### Summary

New searchbar design. Searchbar placed abover the first list element and become visible when user scrolls up from the top of the list. If user scrolls to first list element, searchbar hides.

https://youtu.be/SYSaMEk8OmI

### Review notes
Per discussion with @errorists search behavior changed to show full list of chats when search input is empty. Test fort that fixed.
`pre-binded-header` and `pre-binded-footer` props added to base-list-proprs to avoid components remounting.  
Chat list and filtered items list merged into one, so `filter/views.cljs` removed.

#### Platforms
- Android
- iOS

Chats list, search functionality.

status: ready


errorists:
Screenshot 2020-01-08 at 14 43 35
sad panda

vkjr:
@errorists, you are too quick :))) Last version just pushed.

This version of searchbar hides itself when user drags list around first item. But unfortunately that works only when there are enough chats in the list to fill all screen. Otherwise searchbar is visible. I wasn't able to make it work with only few added chats. I'd suggest to commit changes anyway and move this particular case to a separate issue, wdyt?

errorists:
sorry, wanted to test this one badly! :) On the video it works quite well, cool! are you interpolating the position of the searchbar based on the scroll value? The reason I'm asking, the intention was to reverse engineer the native behaviour since that was unavailable out of the box. To this simple designer here, it always seemed like positioning the searchbar at the first index in the list and then applying negative padding would do the trick. 
I'll test it out properly when it's ready, but based on that I think we could smooth out the rough edges a bit in a later issue like adding animation to the 'Cancel' button state and input width. But yeah I feel like if there's nothing glaring during testing we should definitely merge this one now, it's the good stuff.

vkjr:
>To this simple designer here, it always seemed like positioning the searchbar at the first index in the list and then applying negative padding would do the trick

This implementation is quite close. Searchbar placed not as a first element but as a `header`. To be able to hide it has negative absolute positioning. And to be able to see it, `contentInset` property applied to list.

errorists:
beautiful, thanks for the explanation!

errorists (dismissed):
tested on iOS, works perfectly! 🙏  just one minor note if that's not too much, can we remove the 'Chats' header that's at the start of the search results list? there's no other possibility for search results than chats so it's kinda no-shit sherlock useless :)

src/status_im/ui/screens/home/styles.cljs
```diff
@@ -29,7 +29,10 @@
    :padding-horizontal 16
    :background-color   colors/white
    :align-items        :center
-   :justify-content    :center})
+   :justify-content    :center
```
flexsurfer: map formatting

vkjr:
@errorists, sure will remove it!

churik:
@vkjr 
### ISSUE 1: can't open search on Android.
It can be seen from failed test ` test_search_chat_on_home` and also reproducible on real devices (Android 10, Xiaomi Mi Note 9 Pro)

churik:
### ISSUE 2: Search is shown after creating fresh account on empty chat list
**Steps:**
- install status
- create new multiaccount
Expected, that there will be no search when list is empty
IMG_30C135A7A4D5-1
**Device:**  IPhone7
**OS:**  IOS 13

churik:
### ISSUE 3: Search is opened when switching to new tab or joining new chat
**Steps:**
- install status
- create new multiaccount
- join several chats (if you are doing that trough + > Join Public chat, search will be opening after go back to home)
- close app
- open app > Switch to wallet
- go back to home view - Search is opened again

Did you face this behavior @errorists ? It is extremely noticeable, so I'm wondering, is it problem of my device or IOS 13.
**Device:**  IPhone7
**OS:**  IOS 13

errorists:
@churik ouch, no I did not notice it on my end and can't seem to reproduce, what do you mean that search is open, that it's focused or just visible on top? if it's visible then it's not an issue, it should remain at its previous position as it's anchored to the top of the list. if it's focused then, that's an issue.

churik:
@errorists 
OK, in that case, it looks strange to me - when you justy open chat list, no search is shown on the top.
Then if you switch to wallet or Join to new chat through + > Join new chat - it appears on the top (not focused).
Is it expected behavior? 

And also I didn't manage to close it at all, can you elaborate how to do it? :)

errorists:
@churik I can't reproduce it, the search stays exactly in the same position when changing tabs or pulling up any of the Star New… screens, this is correct and expected. can you record a video of what you mean?

errorists:
But you are right that when there are no chats present on the list, it should be removed / hidden and unavailable.

churik:
Ofc. When I reopen app - no search is shown
Switching to browser and go back - Search appears (?)

flexsurfer:
probably it works different on ios and android

errorists:
@churik thanks! you're right, it shouldn't behave like that, on my iPhone X/iOS13 it remains in the same place so here it would be hidden after changing tabs.

churik:
@flexsurfer on Android it doesn't work at all (at least I and e2e are not able to open it)

errorists:
Ok I found the cause, it's buggy like you show it when I dismiss the 'Invite friends/public chats' banner. it works as it should when it's there, even with many chats on the list.

edit: without the banner I can't seem to hide it, like at all. so yeah, banner bug :)

vkjr:
@churik, thanks for testing!

@churik, @errorists - as I described in a very first message, current implementation doesn't allow to hide searchbar when there are small amount of chats. This is a case for ISSUE 2 and ISSUE 3. ISSUE 3 shouldn't be reproducible when there are plenty of chats.
For me that looks like a react-native bug, because on first render searchbar is hiddent but after scroll you can't get it back. So far I don't know how to fix it. And as I suggested before, we probably can fix this later in a separate PR.

Will check ISSUE 1.

vkjr:
@churik, I'm wrong about ISSUE 2, you meant the case when there are no chats at all.. of course searchbar shouldn't be available at all at this moment.

vkjr:
@errorists, can you reproduce ISSUE 3 even when amount of chats exceeds screen lengths?

errorists:
No, you're right when the list overflows the screen height it works as it should. It's ok to have it like that for the purpose of this PR as it's much better than what we had before, but this needs a fix. 
I'll cc @Ferossgp here, he may have ideas/experience with this issue

churik:
@vkjr please, rebase your PR to current develop, thanks.

churik:
@vkjr ISSUE 1 and ISSUE 2 are fixed, thanks.
For me it is not possible to hide search regardless of amount of chats 
Am I doing something wrong?
no-hide

@errorists are you ok with that behavior?

Also issue with the banner (https://github.com/status-im/status-react/pull/9743#issuecomment-572490863) is still here: 
so if I create account, don't close "Chat and transact privately with freiends" and tap on + > Join public chat > and get back to chat list - Search field is opened (not focused).
So it appears without pull-to-refresh gesture.

errorists:
it needs improvement for sure but if @vkjr doesn't know how to address that, the only thing that comes to my mind is to merge this PR and pass it over to @Ferossgp  
this work here is still much better than what we had before so thanks for that! 🙏 
it should work exactly like it does in Signal. there's a tolerance there, if I scroll the searchbar halfway it will go back, but If I scroll it more than halfway it correctly appears. regardless of how many chats I have opened.

vkjr:
@churik, can see issue that you described, thanks! I would suggest a separate PR to fix those related to searchbar hiding. Have some ideas on them.

As for inability to reproduce hiding - it hides when you stop scrollview between top of the first item and the middle of a searchbar (see video in description). 

Will rebase pr now!

vkjr:
@flexsurfer, could you please review this one?

flexsurfer (approved):

churik:
Let's address issues separately

vkjr:
Created issue #9855

PR #9753, opened 2020-01-06
Change safe-area from position to padding

### LINKED ISSUE #9695: Add white background to the bottom of iOS Safe Area
# Problem

The bottom iOS safe area has a transparent background, making all view transitions visible underneath

IMG_1214

## Implementation

the background of the iOS strip on the bottom of safe area should be white. it was like that some time before.

## Acceptance Criteria

Bottom Safe Area is white, nothing shows underneath it

## Notes

thank you 🙏

Ferossgp:
Makes the view stretch to the size of the safe area instead of moving it from the bottom. Fixes transparent background on navigation animation on iPhone X. On android do not set background as it is drawn above the tabbar thus over the screens because of the absolute position. Needs manual QA cause it may affect the drawing of minimized tabbar over the UI, for example, the android with BG was drawn above the browser navigation. It could be refactored using `safe-area-view` but will take more time with the same result. Fixes #9695

### DIFF
--- a/src/status_im/ui/components/tabbar/core.cljs
+++ b/src/status_im/ui/components/tabbar/core.cljs
@@ -104,9 +104,8 @@
 
 (defn tabs-animation-wrapper-ios
   [content]
-  [react/view
+  [react/view {:style tabs.styles/title-cover-wrapper-ios}
    [react/view
-    {:style tabs.styles/title-cover-wrapper}
     content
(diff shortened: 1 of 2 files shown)


errorists:
tested on iOS, bug is squashed💥 , thanks!

yenda (approved):

Issue #9469, opened 2019-11-18
Update chat message appearance


errorists:
The purpose of this task is to align any chat message styling with the intended design, some of the values outlines in the story may already be correct and present in code, nevertheless I ask to double check that's the case. The purpose here is to make sure the chat message UI is bug free and a joy to look at while providing a solid foundation to build chat based features like message extensions, images etc.

### User Story

As a user, I want to the chat message appearance to be consistent, easily scannable and aesthetically appealing so that I can enjoy using chat more. 

### Description

**1. Update the rounded corners on all messages,** 

incoming messages, top left: `16`, top right: `16`, bottom right: `16`, bottom left: `4`

<img width="141" alt="Group 19" src="https://user-images.githubusercontent.com/4932774/69034954-b8729f80-09e2-11ea-9fd3-a807490a6966.png">

outgoing messages, top left: `16`, top right: `16`, bottom right: `4`, bottom left: `4`

<img width="141" alt="Group 20" src="https://user-images.githubusercontent.com/4932774/69034987-cfb18d00-09e2-11ea-98b9-767098883e01.png">

**2. Make sure inner paddings match design**

Horizontal: `12` , Vertical: `6`

<img width="118" alt="Group 21" src="https://user-images.githubusercontent.com/4932774/69035198-5f573b80-09e3-11ea-8881-0f375ef59fb3.png">

**3. Make sure the minimum space in a line between a timestamp is 12**

<img width="121" alt="Group 22" src="https://user-images.githubusercontent.com/4932774/69035316-aba27b80-09e3-11ea-921b-a7815b251352.png">

**4. Make sure the text is styled correctly**

<img width="351" alt="Group 23" src="https://user-images.githubusercontent.com/4932774/69035372-ca087700-09e3-11ea-94fa-d4e0f1372aca.png">

**5. Make sure that avatars and usernames are wrapped inside touchable and styled correctly**

<img width="362" alt="Group 39" src="https://user-images.githubusercontent.com/4932774/69747889-79adb980-1147-11ea-83e1-63f23d5fd44c.png">

**6. Make sure that the date headers and status indicators match the spec**

<img width="549" alt="Group 27" src="https://user-images.githubusercontent.com/4932774/69035758-a85bbf80-09e4-11ea-9fb4-2a55df2aa182.png">

<img width="257" alt="Group 28" src="https://user-images.githubusercontent.com/4932774/69035787-bc9fbc80-09e4-11ea-8211-68bbeb482de2.png">

[...]

PR #9757, opened 2020-01-07
Change ui of chat message

Ferossgp:
Adds some UI changes as a part of #9469 further improvements require bigger refactor on message components.


src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -93,6 +93,10 @@
     {:align-self    :flex-start
      :padding-left  (if platform/desktop? 24 8)}))
 
+(def message-author-touchable
+  {:margin-left        12
+   :padding-horizontal 2})
```
errorists: @Ferossgp hey, my mistake. it should be vertical here if needed at all (text line height is 18, touchable is 22 so 2px padding would center it inside the touchable)

src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -136,23 +140,33 @@
 
 (styles/defn emoji-message
   [{:keys [incoming-group]}]
-  {:font-size 40
-   :desktop   {:line-height 46}
-   :margin-top (if incoming-group 4 0)})
+  {:font-size    40
```
errorists: is that emoji only font-size? 40 is hardcore, can we lower it to 28

Ferossgp: Yes, it is only emoji size - I can do it.

flexsurfer (approved):

errorists:
hey, can we do something about those guys? :)

Screenshot 2020-01-09 at 09 04 25

errorists:
• Rounded corners ✅ 
• Styling on authors, looks off. Touchable container is 22px high, inside is vertically centred text label with the author name, line height 18px. 8px margin between the touchable and the message bubble preceding it.
• Timestamp, almost got it, it's 1px too high ;) it should appear like the [baseline](https://en.wikipedia.org/wiki/Baseline_(typography)#/media/File:Typography_Line_Terms.svg) of the text and timestamps are positioned at the same level.
• with Emoji messages it's hard to tell since it's placed on a white background, the goal is to have the font-size larger which is correct now but placed on regular message bubbles like other messages. but it's ok for the WIP PR ✅
• Reply icon, did you use the glyph that's in the font? I replaced the icon which was always giving us headaches with positioning with the '↪' arrow that's in Inter, our font, followed by a regular space. it appears like it's using a graphic asset as the margin between it and username is too large.

errorists:
Reply icons are correct now! ✅ 
Timestamp Y position is correct ✅ 
Author values are still incorrect

Screenshot 2020-01-10 at 12 15 57


Areas of the #9469 issue which are not fixed yet:

• 3. Make sure the minimum space in a line between a timestamp is 12 (is 6 now, I kinda like it but it's probably wiser to have a larger margin)
• 6. Make sure that the date headers and status indicators match the spec
• 7. Update the appearance of Emoji only messages
• 8. Update the appearance of Sticker messages
_• 9. Make sure messages grow horizontally to fill the available space correctly_ (fixed in next commit)
• 10. Update the appearance of long messages

errorists:
Message author values are correct ✅ 
Update the appearance of Emoji only messages ✅ 
Make sure messages grow horizontally to fill the available space correctly ✅ 
I came to like the current space between the timestamp so no further changes are necessary there ✅ 

left in the issue, can be divided into separate PRs.
• 6. Make sure that the date headers and status indicators match the spec
• 8. Update the appearance of Sticker messages
• 10. Update the appearance of long messages

src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -46,11 +39,12 @@
...
-         {:color (if (and outgoing (not emoji?))
-                   colors/white-transparent-70
-                   colors/gray)}
+         {:line-height 10
+          :color       (if outgoing
+                         colors/white-transparent-70
+                         colors/gray)}
          (when justify-timestamp? {:position              :absolute
-                                   :bottom                7
+                                   :bottom                (+ 6 3) ; 6 Bubble bottom, 3 message baseline
```
flexsurfer: why don't just use `9` ? and left (+ 6 3) in comment?

flexsurfer (commented):
do we really need all these fonts? and BETA fonts ?

Ferossgp:
@flexsurfer I haven't looked if they are used, I updated only the fonts which were already present. I would suggest to cleanup them in a sperate PR in case it may affect some platforms or code.

errorists:
> do we really need all these fonts?

@flexsurfer that's a great question. to be honest, we don't. we use the following font weights: 
400 (Regular), 500 (Medium), 700 (Bold), with matching italics.

BETA, they're not really 'beta' like an app :) for those open source fonts, it means that some contributors haven't provided languages like Cyrillic, Farsi etc They're safe to use, but yea, we don't use them yet in the app.

flexsurfer (approved):

Serhy:
Tested with iOS and Android (v8.1 and v9), ready for merge!
@Ferossgp please mind my commit fixing e2e test when squashing

errorists:
when merging pls don't close the original issue as there are some items left unresolved there 🙏

PR #9772, opened 2020-01-08
updated images

andmironov:
removed the unnecessary border from the ETH logo


flexsurfer (approved):

flexsurfer:
great job!

Serhy:
@andmironov where is it supposed to be reflected?
I just checked iOS and Android builds and don't see a difference between nightly and PR build on Wallet screen: ETH logo has the same image with border in both builds. 

Also not sure yet where  do we use `default-native.png`, but I found there is https://github.com/status-im/status-react/blob/233d7400275e040e6116233e147f75713ba01610/resources/images/tokens/mainnet/0-native.png which is ETH logo too (and probably it is the one used on Wallet)

andmironov:
Screen Shot 2020-01-13 at 15 19 04

I was trying to remove this double border from the ETH image which it has because the ETH image was wrongly exported with that border

Serhy:
Updated more images (so they are updated now in wallet) 
One thing to mention:

Screenshot 2020-01-13 at 17 48 16

@flexsurfer could you help to include border for the already selected asset in Send transaction bottom sheet (or if it's OK now, @andmironov for above screenshot - I can merge)

andremedeiros:
@errorists can you 👀 this to make sure you're happy with it?

errorists:
sure. while it's not exactly my choice for the eth logo, I would probably go for the white on black to increase contrast against our dominant white colour scheme

IMG_1262

I'm just wondering, on the screenshot Serhy posted it looks wrong, like it's clipped without the border. other than that, good job ✅

andmironov:
hey! so, the problem appears to be a bit more complex than just replacing the image.

1. Semi-transparent border across the app, everywhere there's an asset image https://github.com/status-im/status-react/issues/9832

2. Update the ETH logo to have dark background https://github.com/status-im/status-react/issues/9833

Please feel free to close this pull request and work on the issues above. Thanks for your help @Serhy  @errorists

PR #9773, opened 2020-01-08
UI corrections in stickers panel

### LINKED ISSUE #9699: UI corrections in the stickers panel, scroll / animation
# Problem

Current implementation of the sticker panel has a couple of issues that combined make it appear buggy.

• Whenever it's presented, it visibly scrolls to the sticker pack you had previously selected. Instead it should start at the correct index right away without scrolling to it.
• When switching between the text input and stickers or commands, the entire view blinks and jumps, visible on the attached video. Instead the input and icons toolbar should simply animate to the new position, like it does when switching from stickers to text input.
• another issue spotted by @Ferossgp `if you open stickers in chat then go back and enter again the same chat the stickers panel opens. IMO it should not. It is minor, but I think it can be added in backlog`, needs a fix

[Attached video here](https://www.dropbox.com/s/rpsqyj9truf9cse/stickers%20bugs.mp4?dl=0)

## Implementation

• Stickers start at the correct index without scrolling to it each time
• switching between text input and stickers doesn't blink the view, instead smoothly animating the toolbar to its new position.

## Acceptance Criteria

Both issues were corrected

## Notes

thanks a bunch! 🙏

Ferossgp:
Fixes only ios for #9699

### DIFF
--- a/src/status_im/ui/screens/chat/extensions/views.cljs
+++ b/src/status_im/ui/screens/chat/extensions/views.cljs
@@ -8,8 +8,6 @@
             [status-im.ui.components.animation :as anim]
             [status-im.i18n :as i18n]))
 
-(def panel-height 264)
-
 (defn button [showing?]
   [react/touchable-highlight
    {:on-press            (fn [_]
@@ -30,7 +28,9 @@
                                :useNativeDriver true})])))
 
 (views/defview extensions-view []
-  (views/letsubs [bottom-anim-value (anim/create-value panel-height)
+  (views/letsubs [panel-height [:chats/chat-panel-height]
+
+                  bottom-anim-value (anim/create-value @panel-height)
                   alpha-value       (anim/create-value 0)]
     {:component-did-mount #(show-panel-anim bottom-anim-value alpha-value)}
     [react/animated-view {:style {:background-color :white
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -84,6 +84,7 @@
 (reg-root-key-sub :ui/search :ui/search)
 (reg-root-key-sub :web3-node-version :web3-node-version)
 (reg-root-key-sub :keyboard-height :keyboard-height)
+(reg-root-key-sub :keyboard-max-height :keyboard-max-height)
 (reg-root-key-sub :sync-data :sync-data)
 (reg-root-key-sub :layout-height :layout-height)
 (reg-root-key-sub :mobile-network/remember-choice? :mobile-network/remember-choice?)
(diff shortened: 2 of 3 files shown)


errorists:
sweet, the changes together make a difference, especially the unified height of keyboard and all panels, no more jumping toolbars! 😍 

one small thing to add on css side of things, I noticed the pack selection menu has some wrong values

Screenshot 2020-01-09 at 12 13 44

on the left is build, on right design. the correct height should be 44, icons should centre inside vertically and the tab indicator should sit directly at the bottom of the toolbar without the margin.

thanks a bunch 🙏

Ferossgp:
@flexsurfer There is still some bouncing on android between the transition of an `input-bottom-sheet` to the keyboard. This happens because of the time of processing dispatch and triggering hide of the bottom sheet. For a frame, the keyboard and panel are visible at the same time thus creating a jump. I don't know a fast solution on that, maybe do you have some ideas? Also, there is a bit of jumping I think because the tabbar is set to absolute during the keyboard open and elements are moved down instead of tabbar. When the keyboard closes all elements move up with the height of the tab bar and again create a bad jumping UI. If we do not have a good solution now, I propose to merge it now for iOs and separately looking for android solutions.

errorists (approved):
stickers menu is perfect now :heart_eyes:, thanks!!

src/status_im/ui/screens/chat/extensions/views.cljs
```diff
@@ -30,11 +28,13 @@
...
 (views/defview extensions-view []
-  (views/letsubs [bottom-anim-value (anim/create-value panel-height)
+  (views/letsubs [stickers-panel-height [:chats/stickers-panel-height]
+
+                  bottom-anim-value (anim/create-value @stickers-panel-height)
                   alpha-value       (anim/create-value 0)]
     {:component-did-mount #(show-panel-anim bottom-anim-value alpha-value)}
     [react/animated-view {:style {:background-color :white
-                                  :height    panel-height
+                                  :height    stickers-panel-height
```
flexsurfer: why do we use stickers-panel-height here? probably it should be `chat-panel-height ?`

flexsurfer (approved):

src/status_im/ui/screens/chat/extensions/views.cljs
errorists: I can only add that the design for all panels: commands, stickers, images, anything else we might have in the future - shares the same height.

Ferossgp: I will rename into `chat-panel-height` :+1:, initially, it was created only for stickers and now it needs a rename cause it is reused here.

Serhy:
Tested latest builds ( c89404f) iOS 13.3 and Android 8.1. Lgtm with iOS!

PR #9791, opened 2020-01-10
fix #9758

### LINKED ISSUE #9758: Log level (INFO) in settings in Profile doesn't match real settings
# Bug Report

## Problem
To get geth.log now you need to set custom log level in Settings, because by default it is set "ERROR" level and LogsEnabled= false but in Profile > advanced setting we show "INFO" instead of "ERROR"
Most likely was introduced by  https://github.com/status-im/status-react/pull/9708 

#### Expected behavior
settings in profile match real one

#### Actual behavior
- "INFO" in Profile
- "ERROR" level and LogsEnabled= false  are set (what log settings should be set by default @rachelhamlin ?)

### Reproduction

- Open Status
- Create multiaccount
- Check Profile > settings > Advanced > Log level

### Additional Information

- Status version: nightly 07/01/2020

cc @yenda @rachelhamlin

yenda:
tested on Android
status: ready

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -82,7 +82,8 @@
 (def default-multiaccount
   {:preview-privacy? config/blank-preview?
    :wallet/visible-tokens {:mainnet #{:SNT}}
-   :currency :usd})
+   :currency :usd
+   :log-level config/log-level-status-go})
 
 (defn default-visible-tokens [chain]
   (get-in default-multiaccount [:wallet/visible-tokens chain]))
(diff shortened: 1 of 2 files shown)


cammellos (approved):

rasom (approved):

yenda:
@churik I suppose none of the failing tests are related?

adambabik (commented):
What about this config https://github.com/status-im/status-react/blob/dd52eb176644843ef31d71fb3f02d336bdb2e860/src/status_im/node/core.cljs#L95?

yenda:
@adambabik that is were the default value for log-level in multiaccount is used, same config

PR #9795, opened 2020-01-10
Add glossary view

### LINKED ISSUE #9778: Add glossary of terms to profile page
# Problem

Status introduces a lot of new and unfamiliar terminology. 

## Implementation

Create a basic `Glossary` page under the Profile `Need help?` section.

- `Frequently asked questions`
- `Glossary`
- `Submit a bug`
- `Request a feature`

## Acceptance Criteria

Implement a static text page with the following copy: 

**Account** - Your Status account, accessed by the seed phrase that you create or import during onboarding. A Status account can hold more than one Ethereum address, in addition to the one created during onboarding. We refer to these as additional accounts within the wallet.
**Seed Phrase** - A set of friendly-to-read words, randomly selected from the BIP39 standard list and used to recover or access your Ethereum account on other wallets and devices. Also referred to as a “mnemonic phrase,” “recovery phrase” or “wallet backup” across the crypto ecosystem. Most crypto apps use this same standard to generate accounts.
**Chat Key** - Messages on the Status chat protocol are sent and received using encryption keys. The public chat key is a string of characters you share with others so they can send you messages in Status.
**Chat Name** - Three random words, derived algorithmically from your chat key and used as your default alias in chat. Chat names are completely unique; no other user can have the same three words.
**ENS Name** - Custom alias for your chat key that you can register using the Ethereum Name Service. ENS names are decentralized usernames. 
**Wallet Key** - A 64 character hex address based on the Ethereum standard and beginning with 0x. Public-facing, your wallet key is shared with others when you want to receive funds. Also referred to as an “Ethereum address” or “wallet address.”
**Peer** - A device connected to the Status chat network. Each user can represent one or more peers, depending on their number of devices.
**Mailserver** - A node in the Status network that routes and stores messages, for up to 30 days.

## Notes

Formatting is up to you. We don't have any precedence for a page like this. I'd probably suggest words are bolded, and that we use slightly higher paragraph spacing than normal to separate the entries. cc @errorists

Ferossgp:
Fixes #9778


errorists:
@Ferossgp first of all, nice touch on the sticky indexes!

four bugs I found:
• the list should scroll all the way to the top bar, looks like it's clipped at around 10px below it? probably some content inset gone wrong.
IMG_1253
 
 • there is a typo at the end in 'Walllet', probably requires a thorough spellcheck as I didn't read the whole thing, this one caught my attention

• I can't scroll all the way to the end of the list, 
IMG_1254


• edit: fourth issue, the scrollbar appears at around 16px from the right screen edge

errorists:
ok, fifth issue, the `Glossary` list item should be second on the list in `Need help?`, not last

errorists (approved):

rachelhamlin (approved):
Yay, thanks Gheorghe!

yenda (approved):

PR #9810, opened 2020-01-13
Fix limited Cryptokitties display and order

### LINKED ISSUE #6839: A limit of 20 CryptoKitties is displayed in the collectibles list in the wallet with incorrect order
Type: **Bug**

Summary: As a user, I would like to see the correct number of my collectibles in the wallet, in the right order.
Currently, only 20 CK are shown, without any option to show more of them.
As a user, I would be very confused

**Expected behavior**

The exact number of collectibles should be displayed in the correct order

**Actual behavior**

Only 20 CK are displayed in a different order from the App

**Reproduction**

1. Recover/open an account that has more than 20 collectibles.
2. Navigate to the wallet and click '>' button next to CryptoKitties collectibles in the list.

image

3. List through them to make sure that the valid number of collectibles is displayed
4. Go to the CK App to check that the order that collectibles are displayed is the same

**Collectibles list, top of the list:**

image

**App list, top of the list:**

image

**Additional Information**

Status version: 0.9.31 (1811211215); node 27700aa2 (Nov 22 nightly build)
Operating System: iOS, Android

deoidh:
fixes #6839 

### Summary

Beforehand a limited amount of 20 kitties would appear in the status wallet, I investigated to find that the Cryptokitties API is only limited to 20 collectables per call, which explains why the limited amount was appearing in one's wallet. A parameter of the collectables by wallet call is providing an `offset` to index items for an individual wallet, so I simply created a `range` that extended from zero to the number of cats in a said wallet, incrementing by 20. This was then mapped so that it could be fed into `http-get-n`. Where each individual response was pushed to a sorted array, to finally be loaded simultaneously with one call to `load-kitties` when all of the ids were obtained.  

Unfortunately maintaining an array's order whilst leveraging `mapv` is not possible, I did find that `filterv` work effectively in this department but not when applied to `http-get-n`, it resulted in the following error: 

```
TypeError: cb is not a function. (In 'cb(value)', 'cb' is "TypeError: Network request failed")
```

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- network consumption

### Steps to test
* Open status application
* Navigate to the wallet tab
* Expand a wallet with more than 20 Cryptokitties
* Expand the collectables tab and Select Cryptokitties
* Enjoy the sight of accurate balances

status: WIP

### DIFF
--- a/src/status_im/wallet/collectibles/core.cljs
+++ b/src/status_im/wallet/collectibles/core.cljs
@@ -51,30 +51,35 @@
 ;; Crypto Kitties
 (def ck :CK)
 
+;; TODO Can't maintain the order (ascending) of id's array, mapv randomises indexes and filterv results in a error
 (handlers/register-handler-fx
  :load-kitties
  (fn [{db :db} [_ ids]]
    {:db db
     :http-get-n (mapv (fn [id]
-                        {:url (str "https://api.cryptokitties.co/kitties/" id)
+                {:url (str "https://api.cryptokitties.co/kitties/" id)
                          :success-event-creator (fn [o]
                                                   [:load-collectible-success ck {id (http/parse-payload o)}])
                          :failure-event-creator (fn [o]
-                                                  [:load-collectible-failure ck {id (http/parse-payload o)}])})
-                      ids)}))
+                                                  [:load-collectible-failure ck {id (http/parse-payload o)}])
+                         :timeout-ms            10000})
+                                              ids)}))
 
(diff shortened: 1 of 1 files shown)


flexsurfer:
https://github.com/status-im/status-react/pull/9968

PR #9563, opened 2019-12-02
Add ENS name resolution to EIP681 support

acolytec3:
Summary
EIP681 URI parsing does not currently support ENS name verification. This PR enables the parse-uri function to accept properly formatted ENS names as the to address in a URI as well enables the QR scanner to properly resolve URIs with an ENS address.

This fixes #9238 and fixes #9357.

Review notes
NA

Testing notes
NA

Platforms
Android
iOS
Areas that maybe impacted
Wallet/Send-Transaction/QR-Code

Functional
wallet / transactions
Non-functional
N/A

Steps to test
Open Status

Open Wallet

Select any account

Select Send

Select specify-recipient

Select 'Scan QR code'

Scan the below QR code.
ethereum:pay-snt.thetoken.eth/transfer?address=unicorn.stateofus.eth&uint256=1e16
qr code

Verify that ENS address is resolved to ethereum address and currency changes to SNT.

status: ready

PR #9811, opened 2020-01-13
Open app by ens name

### LINKED ISSUE #7651: Deeplinking issue (universal link to ENS username)
### User Story
As a user, I expect that when I am sent a status link like https://get.status.im/user/foo.stateofus.eth, a chat is directly opened with the user `foo.stateofus.eth`. This is currently not the case..

#### Expected behavior

* Opens a chat directly with the user.

#### Actual behavior

* Status is opened

### Reproduction

* Simply share a link and click on it

### Additional Information

* Status version: v0.10.0
* Operating System: iOS

Ferossgp:
Allow Universal Link for ens name. fixes #7651

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -1205,12 +1205,23 @@
  :contact/qr-code-scanned
  [(re-frame/inject-cofx :random-id-generator)]
  (fn [{:keys [db] :as cofx} [_ contact-identity _]]
-   (let [validation-result (new-chat.db/validate-pub-key db contact-identity)]
-     (if (some? validation-result)
-       {:utils/show-popup {:title (i18n/label :t/unable-to-read-this-code)
-                           :content validation-result
-                           :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}
-       (chat/start-chat cofx contact-identity {:navigation-reset? true})))))
+   (let [public-key?       (and (string? contact-identity)
+                                (string/starts-with? contact-identity "0x"))
+         validation-result (new-chat.db/validate-pub-key db contact-identity)]
+     (cond
+       (and public-key? (not (some? validation-result)))
+       (chat/start-chat cofx contact-identity {:navigation-reset? true})
+
+       (and (not public-key?) (string? contact-identity))
+       (let [chain (ethereum/chain-keyword db)]
+         {:resolve-public-key {:chain            chain
+                               :contact-identity contact-identity
+                               :cb               #(re-frame/dispatch [:contact/qr-code-scanned %])}})
+
+       :else
+       {:utils/show-popup {:title      (i18n/label :t/unable-to-read-this-code)
+                           :content    validation-result
+                           :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}))))
 
 (handlers/register-handler-fx
  :contact.ui/start-group-chat-pressed
--- a/src/status_im/utils/universal_links/core.cljs
+++ b/src/status_im/utils/universal_links/core.cljs
@@ -6,16 +6,15 @@
             [status-im.multiaccounts.model :as multiaccounts.model]
             [status-im.chat.models :as chat]
             [status-im.constants :as constants]
-            [status-im.ethereum.eip681 :as eip681]
+            [status-im.ethereum.ens :as ens]
+            [status-im.ethereum.core :as ethereum]
             [status-im.pairing.core :as pairing]
             [status-im.utils.security :as security]
             [status-im.ui.components.list-selection :as list-selection]
             [status-im.ui.components.react :as react]
             [status-im.ui.screens.add-new.new-chat.db :as new-chat.db]
             [status-im.ui.screens.navigation :as navigation]
-            [status-im.utils.config :as config]
             [status-im.utils.fx :as fx]
-            [status-im.utils.platform :as platform]
             [taoensso.timbre :as log]
             [status-im.wallet.choose-recipient.core :as choose-recipient]))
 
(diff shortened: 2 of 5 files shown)


hesterbruikman:
Leaving GH review for @flexsurfer but looks good to me!

flexsurfer:
is there an issue ?

hesterbruikman:
> is there an issue ?

Created here as it's where most of the work will happen: https://github.com/status-im/universal-links-handler/issues/17

src/status_im/events.cljs
```diff
@@ -1208,12 +1208,23 @@
...
-                           :on-dismiss #(re-frame/dispatch [:navigate-to-clean :home])}}
-       (chat/start-chat cofx contact-identity {:navigation-reset? true})))))
+   (let [public-key?       (and (string? contact-identity)
+                                (string/starts-with? contact-identity "0x"))
+         validation-result (new-chat.db/validate-pub-key db contact-identity)]
+     (cond
+       (and public-key? (not (some? validation-result)))
+       (chat/start-chat cofx contact-identity {:navigation-reset? true})
+
+       (string? contact-identity)
```
Ferossgp:
```suggestion
       (and (not public-key?) (string? contact-identity))
```

src/status_im/utils/universal_links/core.cljs
```diff
@@ -76,11 +75,21 @@
    (chat/start-public-chat public-chat {})
    (pairing/sync-public-chat public-chat)))
 
-(fx/defn handle-view-profile [{:keys [db] :as cofx} public-key]
-  (log/info "universal-links: handling view profile" public-key)
-  (if (new-chat.db/own-public-key? db public-key)
+(fx/defn handle-view-profile
```
Ferossgp, line 75: If the public-key is wrong it navigates to a bad screen. Noticed that but havent't fixed it in current PR, not sure if there is a task for that @hesterbruikman

src/status_im/utils/universal_links/core.cljs
```diff
@@ -105,7 +114,11 @@
     (handle-public-chat cofx (match-url url public-chat-regex))
 
     (spec/valid? :global/public-key (match-url url profile-regex))
```
Ferossgp, line 122: In QR reader there is a different check which uses etherium resolver, but here and in other places it is just regexp. Not sure if it should be unified somehow

Ferossgp: This is the root of the problem described here https://github.com/status-im/status-react/pull/9811#discussion_r367270363

src/status_im/utils/universal_links/core.cljs
hesterbruikman: @Ferossgp what do you mean with 'public key' is wrong? And what screen do you get? Don't recall an issue for this.

src/status_im/utils/universal_links/core.cljs
hesterbruikman: Not sure if it helps, but there are different versions of QR code reader. QR code reader in Wallet checks for accurate Ethereum account address for transactions. QR code reader in Chat checks for valid chat key (public key) or username.

src/status_im/utils/universal_links/core.cljs
Ferossgp: I mean if a user does not exist with that pub-key. You get this screen with an empty name, then when you try to send a message it shows for a second an error.
Screenshot 2020-01-16 at 12 50 53

flexsurfer:
i thought ens names has been implemented here https://github.com/status-im/status-react/pull/9563 ?

hesterbruikman:
> i thought ens names has been implemented here #9563 ?

Could it be that this was implemented for wallet only? Looking at the issue it adds ENS name to a transaction request. The goal here is to add ENS name to the Chat key QR.

flexsurfer:
would be helpful to have a detailed description

Ferossgp:
@flexsurfer it doesn't include ens-name for chat - thus opening https://get.status.im/user/statuse2e is not working. Also, the QR scanner in start chat cant read this QR code.

src/status_im/utils/universal_links/core.cljs
hesterbruikman: Let me try to reproduce. Can you share the steps that resulted in a non-existent public key? Puzzles me because public key is the user identifier. I can't imagine a way to get it if it doesn't exist.

Ferossgp: The experiment was with a broken link, for example, you can remove the last symbol from the link. This mimics a user error when the link is shared.

flexsurfer (approved):

churik:
In https://github.com/status-im/status-react/pull/9563 it was implemented only for wallet, that's right.
@Ferossgp 
### Issue 1:  Redirect to unknown profile if ENS name is not existing
Expected, that there will be no redirect or error.
Currently:
photo_2020-01-23 11 35 03
 It is not reproducible in nightly.


Didn't notice other issues.

Serhy:
LGTM! Tested with iOS13 and Android 9:
- non-existing ENS name in universal links opens nothing and no errors 
- existing ENS names opens profile page of contact
- opening my own ENS name opens main Profile tab 
Thanks @Ferossgp !

hesterbruikman:
@jakubgs looks like we probably want to include in release branch after https://github.com/status-im/status-react/issues/10017 is fixed

Else people who have Status installed will scan the QR code displayed for ENS name on https://join.status.im/user/hester and they wouldn't be able to start a chat. (`unable to read this code`) error.

PR #9812, opened 2020-01-13
Use default keyboard for username input

Ferossgp:
Seems like there is no reason for custom keyboard type as the input is used to input a username.

### DIFF
--- a/src/status_im/ui/screens/ens/views.cljs
+++ b/src/status_im/ui/screens/ens/views.cljs
@@ -204,7 +204,6 @@
          :auto-complete-type     "off"
          :auto-focus             true
          :auto-correct           false
-         :keyboard-type          :visible-password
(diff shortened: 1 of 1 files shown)


yenda:
Looks like it was added in this commit https://github.com/status-im/status-react/commit/7b6dfad70288e032ded6e5b93da6bde902c7d1b5 there might be a reason but I am not sure

flexsurfer:
we use this prop when we don't want to save entered text for keyboard suggestions, probably if this field is only for ens names that's fine to remove it

flexsurfer (approved):

churik:
Keyboard during seed phrase entering and back up doesn't contain suggestions, as expected.
Set ENS name, got suggestions - don't see that it can affect somehow to security. 
Looks good to me.

PR #9813, opened 2020-01-13
Nix: Expose .env feature flags to Nix as attrset

pedropombeiro:

### Summary

This PR allows Nix code to read feature flags present in the correct .env file.

Currently only tested for Android builds (since this is where we'll use this for Nimbus).

status: ready

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,5 @@
 {
-  config ? { }, # for passing status_go.src_override
+  config ? { status-im = { build-type = ""; }; }, # for passing build options, see nix/README.md
 }:
 
 let
--- a/ci/android.groovy
+++ b/ci/android.groovy
@@ -36,10 +36,11 @@ def bundle() {
     /* Nix target which produces the final APKs */
     nix.build(
       attr: 'targets.mobile.android.release',
-      args: [
-        'gradle-opts': gradleOpt,
-        'build-number': utils.readBuildNumber(),
-        'build-type': btype,
+      conf: [
+        'status-im.ci': '1',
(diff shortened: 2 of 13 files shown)


nix/env.nix
```diff
@@ -0,0 +1,31 @@
+# This Nix expression takes care of reading/parsing the correct .env configuration file and return it as an attr set
```
jakubgs: Move this to `nix/tools/envParser.nix`.

default.nix
```diff
@@ -1,9 +1,10 @@
 {
   config ? { }, # for passing status_go.src_override
+  build-type ? "nightly",
 }:
 
 let
-  main = import ./nix/default.nix { inherit config; };
+  main = import ./nix/default.nix { inherit config build-type; };
```
jakubgs: Please move `build-type` into `config`. No need to make it special.

jakubgs:
If changing contents of `config` causes unnecessary rebuilds lets just ditch it and put our own stuff into `params`, including the `status-go` path override thing. It should be passed around implicitly by `callPackage` if it is required as an argument by called Nix expression, so the code should remain clean except for cases when we call `import` directly.

pedropombeiro:
I did some tests and changing values inside config doesn't cause unnecessary rebuilds. I'll redo this PR using `config` sub-values.

jakubgs (approved):
LGTM. But maybe you could add some comments explaining our use of `config`. And possible a section in `nix/README.md` explaining that in more detail.

jakubgs (approved):
Very nice, thanks for the README update.

PR #9825, opened 2020-01-14
Replace static onboarding to animated

### LINKED ISSUE #9774: Update onboarding illustrations to animated versions
# Problem
The static illustrations in onboarding can be replaced with animated illustrations we didn't have before.

## Implementation
@errorists can you add the animations here

## Acceptance Criteria
Exact same dimensions as current illustrations, but replaced with animated sourcefiles.

Ferossgp:
Closes #9774 

I tried an implementation with WebView and meet some complications. First which is easier to fix is autoplay on iOs, for that a small change can be added in the fork of the webview-bridge lib. The second is that react-native resources are served by a local server and inside the webview we do not have access to it. It can be fixed by adding custom js which will be used in webview to require that files. Because of complications, I added `react-native-video` library.

Didn't tried the desktop version.


flexsurfer (dismissed):

errorists (approved):
These look awesome! the perceived production value of our app just gone through the roof. thanks @Ferossgp 👏

jakubgs:
This PR will need a run of `make nix-update-gradle` but right now that fails. I'll have to make a separate PR with a fix for Nix. I'll update when that's ready.

jakubgs:
I took the liberty of adding a fix for `make nix-update-gradle`, adding a missing dependency, and rebasing the whole branch on `develop`.

flexsurfer:
when i press "Get started" i can see black square on android

flexsurfer:
and also these animations so distracting, so i barely can read and understand the text

flexsurfer:
https://drive.google.com/file/d/1I1WYoE-W1-MoZpfhH-kMfbgfCTJY_RTQ/view?usp=sharing

flexsurfer:
also key doesn't look centred 
image

yenda:
I tested yesterday on Android OP7Pro and the videos are not centered vertically and horizontally so they are kind of "in the corner"

errorists:
> when i press "Get started" i can see black square on android. 
>the videos are not centered vertically and horizontally so they are kind of "in the corner"

imo, this is a deal breaker on Android and if we are to ship those animation on that platform it should be fixed. even a delay in starting the animation would be better than the black rectangle, and centring sounds like a no brainer. 

> and also these animations so distracting, so i barely can read and understand the text

this I disagree with. I tested it thoroughly on iOS, gave it to a couple friends and asked them to summarise what they saw later. all of them could recall what the slides were about. I mean sure, the animation catches your attention first rather than the text but that's to be expected. 

Nevertheless, I appreciate your feedback guys! 🤗  @Ferossgp any ideas on how to fix it on Android?

yenda:
@errorists @Ferossgp react-native-video allows to use a placeholder, one could simply use the image as placeholder to avoid any black rectangle.

Ferossgp:
Looking onto it. About centering, this is right - I wanted to center them but then I realized that images were not centered, and I reversed that with another commit, now I see the point to center them horizontally - this is needed cause video is not scaled. About centering them vertically. IMO it is better to add some space at the top. If we center them and other sizes like the text on each slide are different then the center of video will be different on every slide and videos will have different positions. I will test now on other devices and will push a fix soon. Image placeholder can create blinks and sensation of frame drop when the image is replaced with the video, maybe it is possible to export the first frame from video for that 🤔

errorists:
I'll export the first frames of video for image placeholders, thanks for the tip!

I also noticed a bug on iOS, it will freeze the videos from playing if I start minimising the app (probably because it's taking a screenshot for app previews and doesn't wake up the animation after resuming?) here's a [video](https://www.dropbox.com/s/ab4h179mcdgnpy7/iosrecording.mp4?dl=0)

flexsurfer:
this is needed cause video is not scaled. - does that mean we have issues on small screens ?

Ferossgp:
@flexsurfer No - it uses the size of the block in which it is rendered in that case. This is because it uses the min of height or width to draw a square and if the height is smaller, the width of the video is smaller than the width of the screen thus making block not centered.

errorists:
@Ferossgp [first frames of videos for placeholders ](https://www.dropbox.com/sh/d7s5jfx608u0jew/AAD1rX3lt6kCqsYOl3tUDStAa?dl=0)

flexsurfer:
image

Ferossgp:
@flexsurfer because the black background happens during the animation of the keys video, I replaced it with the image on android - other videos are working correctly so I decided to keep them as it is. Tha scale problem is also committed.

churik:
Tested on:
- Motorola X 2nd Gen, Android 6 (English, russian, chinese)
- Nexus 6, Android 6 (English, russian)
- OnePlus7, Android 9 (English, russian)
- Google Pixel 4, Android 10 (English, russian)
- Samsung Galaxy Tab S3, Android 7 (English, russian)
- Samsung Galaxy S10, Android 10
- Iphone XS, IOS 12
- IPhone XS Max, IOS 13
 - IPhone 6S, IOS 11
-  IPhone 7, IOS 13

Looks great!
@flexsurfer would you mend to review PR once more?

flexsurfer (approved):

PR #9828, opened 2020-01-15
Leverage Nimbus Android API

pedropombeiro:

### Summary

This PR integrates https://github.com/status-im/status-go/pull/1794 to allow building an Android app that leverages Nimbus instead of Geth (as per https://github.com/status-im/status-go/issues/1677). The choice of node is done through the new `STATUS_GO_ENABLE_NIMBUS` feature flag (disabled by default).

### Testing notes

We should test that the app's behavior hasn't changed (the automated tests should be sufficient for this).

#### Platforms

- Android

status: ready


nix/status-go/default.nix
```diff
@@ -51,21 +53,52 @@ let
...
-        "ANDROID_HOME=${androidPkgs.androidsdk}/libexec/android-sdk"
-        "ANDROID_NDK_HOME=${androidPkgs.ndk-bundle}/libexec/android-sdk/ndk-bundle"
-        "PATH=${makeBinPath [ openjdk ]}:$PATH"
-      ];
-      gomobileExtraFlags = [ "-androidapi 18" ];
-    };
-    ios = {
+    android =
+      let
+        androidLevel = if enableNimbus then "23" else "18"; # Target Android API level 23 when linking with Nimbus to avoid undefined stderr/stdout symbols when linking
```
jakubgs: Wouldn't it make more sense to just bump everything to 23? 6.0(23) is more than 4 years old.

jakubgs: Besides, we already use `23` as a minimum in Gradle:
https://github.com/status-im/status-react/blob/f34a7bc3f132b61b5c5cbd27d4aa773b392885f4/android/gradle.properties#L24

nix/status-go/default.nix
```diff
@@ -51,21 +53,52 @@ let
...
+      };
+      # platforms = {
+      #   x86 = {
+      #     nimbus = assert enableNimbus; nimbus.wrappers-ios.x86;
+      #     gomobileTarget = "${name}/386";
+      #   };
+      #   arm64 = {
+      #     nimbus = assert enableNimbus; nimbus.wrappers-ios.arm64;
+      #     gomobileTarget = "${name}/arm64";
+      # };
```
jakubgs: Why are those commented?

pedropombeiro: This is what the future code will look like for iOS once we provide support for it. It's probably not valuable enough to keep it, since it can be easily adapted from the Android side of things, I'll remove it.

nix/targets.nix
```diff
@@ -12,6 +12,7 @@ let
   status-go = callPackage ./status-go {
     inherit (mobile) xcodeWrapper;
     androidPkgs = mobile.android.androidComposition;
+    envFlags = callPackage ./tools/envParser.nix { };
```
jakubgs: If this is used only by `status-go` why is it loaded here and passed via `callPackage`? Why not load it directly in `nix/status-go/default.nix`?

status-go-version.json
```diff
@@ -2,7 +2,7 @@
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.39.6",
-    "commit-sha1": "25d46c6d82fbf3e4b6659cb96cf1789bac87f12a",
-    "src-sha256": "1m6d07hlxfp8yr94ld73m6dflrgdyjhmpr2ajnjkhjlv9ffvicgb"
+    "version": "feature/nimbus-backend",
```
jakubgs: I assume this is temporary.

nix/status-go/default.nix
jakubgs: I'm fine with leaving it IF you add a comment explaining what it is and what it will be for in the future.

status-go-version.json
pedropombeiro: Yeah, it's been updated since

nix/targets.nix
pedropombeiro: I'm still not comfortable with the idea of referencing paths with `../` in Nix, it feels like I'm referencing paths that the current file isn't supposed to know about. But it does look better. Changed.

nix/status-go/default.nix
```diff
@@ -3,14 +3,17 @@
 
 let
   inherit (stdenv.lib)
-    catAttrs concatStrings fileContents importJSON makeBinPath
+    catAttrs concatStrings concatStringsSep fileContents importJSON makeBinPath
     optional optionalString strings attrValues mapAttrs attrByPath
     traceValFn;
 
+  envFlags = callPackage ../tools/envParser.nix { };
```
jakubgs, line 10: Would be nicer if it worked as a method that takes the `.env` file path as an argument to parse it. But it's fine.

jakubgs (approved):

PR #9841, opened 2020-01-16
Update e2e and fleet in e2e env

Serhy:
Update e2e and fleet in e2e env


.env.e2e
```diff
@@ -3,7 +3,7 @@ DEFAULT_NETWORK=testnet_rpc
 ERC20_CONTRACT_WARNINGS=1
 ETHEREUM_DEV_CLUSTER=1
 EXTENSIONS=0
-FLEET=eth.staging
+FLEET=eth.prod
```
jakubgs, line 6: I actually though it would be better to use a separate fleet for tests. Or do you insist on testing on prod?

churik: @jakubgs `eth.staging` now and `eth.prod` have the same settings and same status software?

Serhy: As long as the separate fleet corresponds to prod version - that's fine (and even has more sense) to stay with eth.test/eth.stage. 
It would be handy to have it customisable (like in manual build) at least we are able to have e2e build with `eth.prod` before we release (in and in further subsequent releases)

jakubgs: >same settings

Yes.
>same software

No.

>It would be handy to have it customisable (like in manual build)

That was removed in #9398 because people complained it caused some confusion.
You want it back now?

churik: > same software

> No.

so, in that case, I would prefer to see how e2e will go on real `eth.prod` with newest status software.
right @Serhy ?

Not sure about taking back fleet settings, we are not changing them so often.

Serhy: Right, at least before V1 goes live let's keep it on `eth.prod`. Then after V1 we will revert back to `staging` env which will have more sense.

jakubgs: Fair enough.

jakubgs (approved):

PR #9842, opened 2020-01-17
Remove outdated assets

Ferossgp:
Haven't looked at icon assets cause there are roo many for manual lookup. Waits for #9869


src/status_im/android/core.cljs
```diff
@@ -32,14 +32,12 @@
                       "keyboardDidShow"
                       (fn [e]
                         (let [h (.. e -endCoordinates -height)]
-                          (dispatch [:hide-tab-bar])
```
flexsurfer: whos this related to assets ?

errorists:
I glanced at the removed assets, I believe all of them are no longer used ✅  
with Inter i think we probably could remove the SemiBold weight too, but honestly i'm not sure if any of the other designers didn't use it somewhere so it's good too

flexsurfer (approved):

src/status_im/android/core.cljs
Ferossgp: Not related, but I found that these events are not used because the handling was moved into atom on tabbar component. Should I move it into a separate PR?

Ferossgp: https://github.com/status-im/status-react/pull/9843

ios/StatusIm.xcodeproj/project.pbxproj
```diff
@@ -187,7 +186,6 @@
 			isa = PBXGroup;
 			children = (
 				2028E0111D4275BD00227DCD /* SF */,
-				B23B48FE1E76917B006D4535 /* RobotoMono-Medium.ttf */,
```
errorists, line 210: @Ferossgp are you sure we don't use this anymore? we use monospaced fonts in places like account addresses, I guess all of them use the default 'monospace' declaration which defaults to 'Menlo' on iOS. Not ideal, Roboto Mono would be better since it looks more similar to Inter. Even better would be SF Mono, the font Apple uses on iOS but afaik it's private and unavailable?

flexsurfer: good point

Ferossgp: Actually, the font does not exist, it was removed before that PR, I've removed the link to the inexistent font. I also found now that there is also a link to an inexisting SF font. We can replace the font to the one you prefer in a separate PR.

errorists: Ok. I'll follow up on the monospace font declaration then in a separate issue. As for SF, thanks! For a minute there I was worried we still have it. It's licensed by Apple so we actually can't embed it our app assets, a day before the App Store launch we could get into some shit by doing that :)

Serhy:
Don't see any difference between develop in iOS and Android builds

PR #9847, opened 2020-01-17
peg clock value to whisper timestamp

cammellos:
peg clock value to whisper timestamp, so that messages can't be ordered in the past/future , up to 120s, all should work just the same

### DIFF
--- a/status-go-version.json
+++ b/status-go-version.json
@@ -3,6 +3,6 @@
     "owner": "status-im",
     "repo": "status-go",
     "version": "v0.39.6",
-    "commit-sha1": "25d46c6d82fbf3e4b6659cb96cf1789bac87f12a",
-    "src-sha256": "1m6d07hlxfp8yr94ld73m6dflrgdyjhmpr2ajnjkhjlv9ffvicgb"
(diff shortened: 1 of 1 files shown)


churik:
@cammellos 
may it fix issue in pairing that we discussed few days ago? when 1 of the paired devices has custom time.

cammellos:
no, this is slightly different, but actually that might already be fixed in develop, if you could try next time you have a chance (on new accounts only), thanks!

churik:
should be tested manually @cammellos ?

cammellos:
@churik it will probably be included in @yenda PR, I will ask you to test maybe develop once is merged, basically by changing the clock of the device and checking what happens (everything should work as normal)

churik:
@cammellos into https://github.com/status-im/status-react/pull/9814?

cammellos:
@churik yes

jakubgs (approved):
Are we sure we don't want to bump the `version`?

cammellos:
@jakubgs this PR will be closed (but we always point to latest develop, only after testing is successful), this commit is included in v0.39.8

cammellos:
Closing this as will be included in a different PR

PR #9859, opened 2020-01-21
ci: fix GitHub release by using glob for all files

jakubgs:
This should be much simpler than specifying filenames by using `pkgFilename()` generator.

This was done because we wanted to release just mobile versions on GitHub, but now we don't even build them so that's not necessary.

Also, adjusted `ghcmgr` POST to use `ipa` file if Diawi link is not available.

### DIFF
--- a/ci/github.groovy
+++ b/ci/github.groovy
@@ -181,15 +181,16 @@ def publishRelease(Map args) {
   }
 }
 
-def publishReleaseMobile() {
+def publishReleaseMobile(path='pkg') {
+  def found = findFiles(glob: "${path}/*")
+  if (found.size() == 0) {
+    sh "ls ${path}"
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

ci/github.groovy
```diff
@@ -181,15 +181,16 @@ def publishRelease(Map args) {
   }
 }
 
-def publishReleaseMobile() {
+def publishReleaseMobile(path='pkg') {
+  def found = findFiles(glob: "${path}/*")
+  if (found.size() == 0) {
+    sh "ls ${path}"
+    error("No file to  release in ${path}")
```
pedropombeiro:
```suggestion
    error("No file to release in ${path}")
```

pedropombeiro (approved):

PR #9870, opened 2020-01-21
fix: remove ipad support

andremedeiros:
We don't design for iPad specifically, so we can't claim the app supports those screens.

cc @flexsurfer @errorists 

status: ready

### DIFF
--- a/ios/StatusIm.xcodeproj/project.pbxproj
+++ b/ios/StatusIm.xcodeproj/project.pbxproj
@@ -24,7 +24,7 @@
 		81C6E6AE0AA739BE9D87C1D0 /* libPods-StatusImTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = FC1CBCFE6C906043D6CCEEE1 /* libPods-StatusImTests.a */; };
 		8391E8E0E93C41A98AAA6631 /* Inter-SemiBoldItalic.otf in Resources */ = {isa = PBXBuildFile; fileRef = A4F2BBE8D4DD4140A6CCAC39 /* Inter-SemiBoldItalic.otf */; };
 		8E55E6877F950B81C8D711C5 /* libPods-StatusIm.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 101A4045637A2ADF57D28EF5 /* libPods-StatusIm.a */; };
-		925C1F4C1F7B73B20063DFA0 /* BuildFile in Frameworks */ = {isa = PBXBuildFile; };
+		925C1F4C1F7B73B20063DFA0 /* (null) in Frameworks */ = {isa = PBXBuildFile; };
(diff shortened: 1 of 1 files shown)


ios/StatusIm.xcodeproj/project.pbxproj
```diff
@@ -24,7 +24,7 @@
 		81C6E6AE0AA739BE9D87C1D0 /* libPods-StatusImTests.a in Frameworks */ = {isa = PBXBuildFile; fileRef = FC1CBCFE6C906043D6CCEEE1 /* libPods-StatusImTests.a */; };
 		8391E8E0E93C41A98AAA6631 /* Inter-SemiBoldItalic.otf in Resources */ = {isa = PBXBuildFile; fileRef = A4F2BBE8D4DD4140A6CCAC39 /* Inter-SemiBoldItalic.otf */; };
 		8E55E6877F950B81C8D711C5 /* libPods-StatusIm.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 101A4045637A2ADF57D28EF5 /* libPods-StatusIm.a */; };
-		925C1F4C1F7B73B20063DFA0 /* BuildFile in Frameworks */ = {isa = PBXBuildFile; };
+		925C1F4C1F7B73B20063DFA0 /* (null) in Frameworks */ = {isa = PBXBuildFile; };
```
andremedeiros, line 27: Not 100% sure why this changed.

ios/StatusIm.xcodeproj/project.pbxproj
```diff
@@ -161,7 +161,7 @@
 				B24FC7FD1DE7195700D694FF /* Social.framework in Frameworks */,
 				CE4E31B31D8695250033ED64 /* Statusgo.framework in Frameworks */,
 				20AB9EC61D47CC0300E7FD9C /* libRCTStatus.a in Frameworks */,
-				925C1F4C1F7B73B20063DFA0 /* BuildFile in Frameworks */,
+				925C1F4C1F7B73B20063DFA0 /* (null) in Frameworks */,
```
andremedeiros, line 164: Not 100% sure why this changed either.

errorists (approved):
works as expected, thanks a bunch for sparing us the one star reviews 🙌 

IMG_0005

flexsurfer:
dunno, i would keep ipad support , UI is not great but at least browser works great  https://twitter.com/izqui9/status/1063047953353465856

vkjr:
+1 to @flexsurfer, I personally feel iphone-only apps on ipad a bit irritating. And we have 2-pane UI support on iPad (if still works) so it should be a bit more convenient in chats.

errorists:
The fact that we have it is because some responsive layouts in React allowed us to, it never was given much thought and consideration. The UI there is a train wreck riddled with bugs, I don't want to invite any more excuses for people to leave us one star reviews.

PR #9878, opened 2020-01-22
status-go: upgrade 0.39.8 > 0.39.10

jakubgs:
This should fix incompatibility issues with the `eth.prod` fleet by upgrading `status-go` to `0.39.10`.

Also included a bump of `fastlane-plugin-diawi` to `1.4.0` which includes my fix for setting delay.


adambabik (approved):

cammellos (approved):

adambabik:
This bump includes two major PRs from status-go which potentially can affect the whole `shhext_*` namespace. Please test all parts using `shhext_*` JSON-RPC commands.

yenda (approved):

jakubgs:
I'm trying the PR build and no messages seem to be loading.
I also cannot select a mailserver by hand because I cannot unclick the `Auto` button in this screen:
Screenshot_20200122-111535
So I'm not sure this is working. Any thoughts @adambabik ?

jakubgs:
According to @adambabik this selection issue might be to this change: https://github.com/status-im/status-go/commit/79b8112f89e218e5d24b6c4400eeaf6d4e8ec6db

So `0.39.8` might be the most stable version we have right now.

PR #9881, opened 2020-01-22
remove chaos mode checking

flexsurfer:
remove chaos mode checking


vitvly (approved):

churik:
Failed e2e are related to `eth.prod` fleet, can create account, join to chat and fetch history in `eth.staging` fleet

I believe it is enough - feel free to ping me @jakubgs  if think that it is not

PR #9888, opened 2020-01-22
Fix ens names & contacts

### LINKED ISSUE #9887: Contact of ENS users removed from 'Contact list' after relogin
# Bug Report

## Problem

It may appear at some point after re-login that contact which was added in 'Contact list' - removed from contact list and often contact name of this user dropped to default three-name username. 
Can be reproduced with ENS username when ENS-name user adds ens after he was added in contacts 

#### Expected behavior
Contact always kept in 'Contact list' on any account disregard when (and how many) ENS name has been added (or changed, if there are multiple ENS names) after relogin

#### Actual behavior
Contact removed from 'Contact list'  after relogin if ENS name added after this user been added in contacts and user relogins (please see steps to reproduce)

https://drive.google.com/file/d/1T9RP8xMapTpVGj_vDas2mOjyqjKuehRo/view?usp=sharing

### Reproduction

- Open Status on device 1 and create new account (`UserA`)
- Recover user which has ENS name (`ENS-user`) (you can use your own or this one: `fog pair manual warfare rug combine junk talk fold found position over` - it's already with ENS name `@chicken`) 
- `UserA` adds `ENS-user` by searching it in from `Start new chat` -> enter `chicken` -> hit `Done` 
- `UserA` sends several messages to `ENS-user`
- `ENS-user` adds `UserA` in contacts and sends several messages
- `ENS-user` adds ens-name from `Profile` -> `ENS usernames` (the `chicken` in our case)
- `UserA` re-logins to account and checks if `@chicken` contact still in the `Start new chat`

Even if `UserA` adds `ENS-user` in contact again, each time he receives message from `ENS-user` and re-logins then ENS-user disappears from UserA contact list

### Additional Information
- Status version: develop and RC 

cammellos:
fixes: #9887
status: ready


flexsurfer (approved):

Serhy:
Tested iOS and Android 9,  #9887 can not be reproduced anymore (I used `eth.prod` fine btw). Changing ENS (when multiple of them in account) also changes username it on recipient side, re-login to accounts keep the contact in 'Contact list'. E2E failed tests rechecked (there was one when app crashed at the point not related to this PR which I could not reproduce (but will keep logcat with crash stacktrace and log separately), another test had not enough time to wait updated balance and third one required some ETHro to fill in it's balance (updated already)
Thanks a lot @cammellos !

PR #9889, opened 2020-01-23
Remove react/modal view from android navigation

### LINKED ISSUE #9884: Stickerpage overlaps with "Send transaction" screen
# Bug Report

## Problem

At attempt to buy 'TOZEMOON' stickers 'Send transaction' screen doesn't cover stickers layer, so screens are overlapping and it is possible to tap 20 SNT several times

#### Expected behavior
no overlapping

#### Actual behavior
photo_2020-01-22 14 27 18

### Reproduction

- Open Status
- Join to #status
-  Tap on `TOZEMOON' stickerpack to but it

### Additional Information

- Status version: RC [build 10 ](https://ci.status.im/job/status-react/job/release/job/release%252F1.0.x/10/)

- Operating System: Android (checked on Android 10 and Android 8)

Ferossgp:
Fixes #9884 

React modal is drawn above the app (maximum z-index) while the bottom sheet and popover are part of the app with position absolute. https://facebook.github.io/react-native/docs/modal.html documentation also suggest to not use modal in that case cause we lose some control over navigation. Now modal are handled by navigation and we can draw our popover and bottom sheet on bigger z-index and  better control what should be on the top.

Further changes will come in another PR which upgrades react-navigation and adds some more control over it.

An extra benefit of that is unifying the handling on iOs and Android which should make easier maintenance in future.

### DIFF
--- a/components/src/status_im/ui/components/react.cljs
+++ b/components/src/status_im/ui/components/react.cljs
@@ -311,10 +311,7 @@
      children)))
 
 (defn main-screen-modal-view [current-view & components]
-  ;; NOTE on Android we use Modal component and it manages statusbar area by itself
-  [(if platform/ios?
-     (create-main-screen-view current-view)
-     view)
+  [(create-main-screen-view current-view)
    styles/flex
    [(if (= current-view :chat-modal)
       view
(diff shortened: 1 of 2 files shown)


errorists (approved):
works ok on Android

churik:
Tested on IOS 13 and Android 10, looks good to me, issue is gone

flexsurfer (approved):

PR #9890, opened 2020-01-23
[DEBUG]add db debug in dev/e2e

cammellos:
add database logging similar to re-frisk, in e2e and dev builds, helps to debug some state issues.
large logs might be cut off in logcat, but should help in most cases.

status:ready


rasom (approved):

Serhy:
We log even seedphrase and passwords now. Fine for e2e, not sure it's OK for develop (means the same will be in nightlies?)

cammellos:
@Serhy I have changed it so only e2e will log everything, seems safe to me as we don't have any confidential information I believe, and can help us debug some issues (for example mailserver issue).
I have also changed default loglevel to `DEBUG` for e2e/develop , update tests.

Not sure what to do about the tests looking for the seed phrase/password, they are useful for checking regressions, but unfortunately with this change they'll break. Don't know what's best, disable logging or keep the tests. What do you think?

yenda:
@Serhy @cammellos nightlies are built with .env or .env.nightly?

cammellos:
@yenda not sure, I'd expect `.env.nightly`

PR #9898, opened 2020-01-23
add watch only address using qrcode

tbenr:
when adding a watch-only address I added:

- ability to add via QR code scan
~~avoid adding if address has been already added~~
~~i18n on some hardcoded strings~~ (fixed by latest pr)

status: ready

image


flexsurfer:
> avoid adding if address has been already added

i believe we have this already, next button is disabled if address has been already added

tbenr:
@flexsurfer oh right it's in the `:add-account-disabled?` sub.
my qr implementation bypass that so i implemented it also in the fx running after

andmironov (approved):

tbenr:
@andmironov i changed a bit the message and how it is displayed

flexsurfer:
@tbenr could we just insert scanned address into input field? in that case we don't need to duplicate logic and validations

tbenr:
@flexsurfer it was one option but then, as a user, I thought it was better to go straight to next screen. But I can change it, no prob (I should have time this evening).

flexsurfer:
@tbenr yeah i think better to change that, because it would be great if you can review scanned address before adding it, and code will be simpler, thanks!

tbenr:
@flexsurfer done!

flexsurfer (approved):

Serhy:
Looks great to me on iOS and Android when adding QR code from Status. 
However, I guess at the moment it's not often case when user will need to scan QR codes from Status app account addresses.

Mostly it will be pure address var encoded in QR code (while in Status app has encoded in `ethereum:0x.........@XX` format (where XX is network id). And in that case it fails to add address saying 'Scanned QR code doesn't contain a valid address'

So when scanning QR code (value `ethereum:0xDdc867bAB69c9488E65f610E071752f621C7F504`)
Screenshot 2020-02-13 at 16 02 16
OR 
with the content value (`ethereum:0xDdc867bAB69c9488E65f610E071752f621C7F504@3`)
Screenshot 2020-02-13 at 16 02 32
then it's parses and inserted fine 

But when scanning address (normalised one, - value is` 0xDdc867bAB69c9488E65f610E071752f621C7F504`)
Screenshot 2020-02-13 at 15 52 03
or same but non-normalised 
Screenshot 2020-02-13 at 15 52 36
then 'Scanned QR code doesn't contain a valid address' :(

@tbenr, could you update it please so it could eat pure ETH address as well?

tbenr:
@Serhy done!

churik:
Retested with QRs from the description, with QR `ethereum:0x.........@XX` format on IOS and Android. 
Also checked error handling when scanning invalid QR.
Thank you @tbenr !
Looks good from my side.
It contains also `design-review` label.
@errorists would you mind to take a look?

errorists (approved):

hesterbruikman:
Not sure if the issue @Serhy raised was resolved in the PR build I just tried. For a 'pure' ETH address without `Ethereum:` prefix I get the error "Scanned QR code doesn't contain valid address"

churik:
@hesterbruikman what build and platform you tried?

hesterbruikman:
Platform: Android
App version: 0.14.0 (20200205205612)
StatusIM-200205-205455-290d14-pr-universal.apk

churik:
@hesterbruikman try https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200214-073916-bbb549-pr-universal.apk - this is last one which I tested

hesterbruikman:
Works! Thanks for checking. Sorry I must have scrolled to the wrong date build on mobile 😬

PR #9907, opened 2020-01-24
Adds a screen for debugging blocks lag

rasom:

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -718,6 +718,7 @@
 	"network": "Network",
 	"network-chain": "Network chain",
 	"network-details": "Network details",
+	"network-info": "Network info",
 	"network-fee": "Network fee",
 	"network-id": "Network ID",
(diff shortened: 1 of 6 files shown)


translations/en.json
```diff
@@ -718,6 +718,7 @@
 	"network": "Network",
 	"network-chain": "Network chain",
 	"network-details": "Network details",
+  "network-info": "Network info",
```
vitvly: I've noticed that en.json uses tabs instead of spaces. But does it even matter?

vitvly (approved):

translations/en.json
yenda: yeah it does though if we could switch the whole file to what @rasom is using I'd prefer cause it's fucked up now I have to enable whitespace mode and copy paste whitespaces to keep it even

rasom: fixed

PR #9908, opened 2020-01-24
nix: throw error if 'make nix-purge' is called on NixOS

jakubgs:
In case someone runs NixOS and tries to use `make nix-purge`.


corpetty (approved):
lgtm

Issue #9280, opened 2019-10-25
Assets (tokens) are not loaded due to infura error


churik:
# Bug Report

## Problem
Sometimes due to "context deadline exceeded" from `infura.io` balance can't be fetched. Can't reproduce, but potentially it can be an issue on mainnet as well.
Caught by e2e. 

#### Expected behavior
Balance assets are loaded

#### Actual behavior
No assets (tokens) are shown.
Full video / log: https://app.saucelabs.com/tests/aaa060e1b584497db876f67e5687a319?auth=f2c8623188b3a2a34e16a0e655e57fcd#54

### Reproduction

- Open Status
- Restore account with funds (ETH + tokens)

### Additional Information

- Status version: nightly 23/10
- Operating System: Android, iOS

#### Logs

```
10-23 04:58:37.739  4296  5214 E GoLog   : ERROR[10-23|11:58:37.739] failed to get transfer in between two bloks from=4200000 to=4300000 error="Post https://ropsten.infura.io/v3/f315575765b14720b32382a61a89341a: context deadline exceeded"
10-23 04:58:37.739  4296  5214 E GoLog   : ERROR[10-23|11:58:37.739] failed to get next batch                 error="Post https://ropsten.infura.io/v3/f315575765b14720b32382a61a89341a: context deadline exceeded"
```

churik:
@adambabik @cammellos would be nice if you can comment issue to understand it better.
Thanks a lot!

cammellos:
@churik it means it's taking too long the request, and it's timing out, we can probably extend it (at least for tests)

ryanschneider:
Can someone familiar with the status code point me to where this is happening?  Based on the logs it sounds like a eth_getLogs query is being slow, but we’ve done a lot of optimizations on our end to make those performant so I’d like to see if I can reproduce a consistently slow query.

StatusSceptre:
@yenda @cammellos can one of you point @ryanschneider to the relevant code?

yenda:
@ryanschneider I think this is happening here: https://github.com/status-im/status-go/blob/develop/services/wallet/downloader.go#L242

churik:
@rachelhamlin it is more visible now, I faced it several times per day.
It is pretty dangerous from user perspective, user may think that he lost assets.
So should be added to v1 from my POV.
Assets don't appear at all in some cases, sometimes they appear after several re-logins

ryanschneider:
[...]

PR #9910, opened 2020-01-24
[#9280] Add retries on error when fetching balances

rasom:
Naïve implementation, but it might fix most of cases. A better fix with some additional ui can be added later.

status: ready

### DIFF
--- a/src/status_im/wallet/core.cljs
+++ b/src/status_im/wallet/core.cljs
@@ -27,24 +27,31 @@
             [status-im.ethereum.stateofus :as stateofus]
             [status-im.ui.components.bottom-sheet.core :as bottom-sheet]))
 
-(re-frame/reg-fx
- :wallet/get-balance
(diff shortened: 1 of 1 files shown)


src/status_im/wallet/core.cljs
```diff
@@ -27,24 +27,31 @@
...
- (fn [{:keys [account-address on-success on-error]}]
-   (json-rpc/call
-    {:method     "eth_getBalance"
-     :params     [account-address "latest"]
-     :on-success on-success
-     :on-error   on-error})))
+(defn get-balance
+  [{:keys [address on-success on-error number-of-retries]
+    :as params
+    :or {number-of-retries 2}}]
```
flexsurfer: 4 maybe ? :)

flexsurfer (approved):

src/status_im/wallet/core.cljs
rasom: well, ok

PR #9913, opened 2020-01-24
Fire request once network status changes & periodically check connection

### LINKED ISSUE #9372: Message sent to user who is offline not received when he appears back online
# Bug Report

## Problem
No messages sent in 1-1 chat are received by recipient who appears online after being offline when messages were sent. 
I **can't reproduce it with real devices** but it started to constantly fail with e2e using emulators. 

#### Expected behavior
Message sent to contact who is offline is received when contact appears online
#### Actual behavior
Message sent to contact who is offline is not received when contact appears online
Video: https://app.saucelabs.com/tests/a65c1b55bad1430db08505714b514e70

#### Notes
This appeared after https://github.com/status-im/status-react/pull/9157/commits/33af176246b8a6b3261590253c9f17b370bf4301 commit

I attached logcat below.

#### Acceptance Criteria
e2e test related to offline message receiving (test_offline_messaging_1_1_chat) passes

### Reproduction (for 1-1 chat)
- Open Status on two emulators and create two new accounts on each device (UserA and UserB)
- UserB, staying on 'Chats' view enables airplane mode 
- UserA adds UserB in contacts
- UserA sends several message to UserB (who is still offline)
- UserB disables airplane mode
- UserB verifies new chat with messages sent from UserA appears on Chats view 

### Reproduction (for  public chat)
- Enable Airplane mode
- Join public chat with messages (e.g. `#status`)
- Disabled Airplane mode
- Verify messages are received in next ~60 sec

 
### Additional Information
- Status version: Develop StatusIm-191029-025900-dfe259-nightly-universal.apk
- Operating System: Android 8.0 (

#### Logs

Logcat for scenario of joining public chat in offline mode when messages are not received after appearing online: 
[logcat_where_no_messages_in_public.log](https://github.com/status-im/status-react/files/3793663/logcat_where_no_messages_in_public.log)
Video for this logcat: https://app.saucelabs.com/tests/a65c1b55bad1430db08505714b514e70

Logcat for the same scenario but with working build: 
[logcat_where_everything_is_ok.log](https://github.com/status-im/status-react/files/3793665/logcat_where_everything_is_ok.log)
Corresponding video with this logs: https://app.saucelabs.com/tests/aa9ebb7788b34f71bd1c5fa61035bc64#61

cammellos:
Fixes: #9372 

This commit fixes a couple of issues.

The first one is that when we go from offline->cellular-network->wifi
historical requests were not triggered (only when going from
offline->online, but if they are disabled in cellular network they won't
fire).

The second is due to the fact that it is possible that the connection
status gets into a bad state, (this happens for example if get-latency
returns all errors, but we are online, and it won't be retried), and no
connection attempt is made anymore.

I have changed the logic to periodically check the connection state, and
try to connect if too much time has passed since last error (10s),
similarly to what we did before, but in a tick, so we are less reliant
on having the right state in the db.

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -193,21 +193,19 @@
                 (generate-mailserver-symkey mailserver)))))
 
 (fx/defn add-peer
-  [{:keys [db] :as cofx}]
+  [{:keys [db now] :as cofx}]
   (let [{:keys [address sym-key-id generating-sym-key?] :as mailserver}
         (fetch-current db)]
     (fx/merge
      cofx
-     {:db (-> db
-              (update-mailserver-state :connecting)
-              (update :mailserver/connection-checks inc))
+     {:db (assoc
+           (update-mailserver-state db :connecting)
+           :mailserver/last-connection-attempt now)
       :mailserver/add-peer address
       ;; Any message sent before this takes effect will not be marked as sent
       ;; probably we should improve the UX so that is more transparent to the
       ;; user
-      :mailserver/update-mailservers [address]
-      :utils/dispatch-later [{:ms constants/connection-timeout
-                              :dispatch [:mailserver/check-connection-timeout]}]}
+      :mailserver/update-mailservers [address]}
      (when-not (or sym-key-id generating-sym-key?)
        (generate-mailserver-symkey mailserver)))))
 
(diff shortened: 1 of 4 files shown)


rasom (approved):

Serhy:
Can't say that https://github.com/status-im/status-react/issues/9269 is fixed 
`Error connecting to mailserver` appears as soon as I tap on the same mailserver, and after I tap "Retry" in 10 sec `Error connecting to mailserver` appears again, but almost all the time it connected successfully after that. Same behavior actually with Release candidate.

But I could not reproduce https://github.com/status-im/status-react/issues/9269, neither with real devices nor when e2e tests running.
And could not reproduce endless `Connecting...` status with PR.
Messages are fetched from offline fine as well.

src/status_im/mailserver/core.cljs
```diff
@@ -521,23 +518,24 @@
...
       decrement the connection check counter
    else
       change mailserver if mailserver is connected"
-  [{:keys [db] :as cofx}]
+  [{:keys [db now] :as cofx}]
   ;; check if logged into multiaccount
   (when (contains? db :multiaccount)
-    (let [connection-checks (dec (:mailserver/connection-checks db))]
-      (if (>= 0 connection-checks)
+    (let [last-connection-attempt (:mailserver/last-connection-attempt db)]
```
yenda, line 531: is the docstring still relevant after these changes?

yenda:
you mean connection-state of mailservers correct? if get-latency returns only errors/nothing, couldn't we dispatch-later a retry instead of using the same loop as publish-updates, isn't going to publish updates every 10 sec instead of 2 minutes now?

yenda (changes requested):
(just requesting changes because it's tested so it doesn't get merged immediatly)
you mean connection-state of mailservers correct? if get-latency returns only errors/nothing, couldn't we dispatch-later a retry instead of using the same loop as publish-updates, isn't it going to publish updates every 10 sec instead of 2 minutes now?

src/status_im/mailserver/core.cljs
cammellos: Nope, I need to update it thanks

cammellos:
@yenda, thanks for the feedback, It would not publish updates, basically `utils.publisher` is just a ticker (we could rename it, before it was publishing updates/contact codes, so made more sense), that run every 10s, it's up to each separate job/fn to make sure they don't act too often.

In case of account-updates we use `last-updated` (we publish updates every 48 hours), while we use `last-connection-attempt` for mailserver (10s same as the ticker). 

I would not use `dispatch-later` and self scheduling, mainly because I think it's a bit more prone to error, as that code changes more often, if there's any error it will stop, etc. + having the responsibility of scheduling to `utils.publisher` lift one responsibility from it, so less responsibility to single components (Both though work fine, I'd sleep a bit better knowing that a different component is taking care of it though :) )

@Serhy I might take a look at that separately then, as this might be worth including in v1

yenda (approved):

Serhy:
> @Serhy I might take a look at that separately then, as this might be worth including in v1

@cammellos sure, I'm OK with that!

PR #9915, opened 2020-01-24
Make bottom sheet height dynamical

### LINKED ISSUE #9848: Wrap bottom sheet component around the total height of its nested elements
# Problem

Our bottom sheet component requires us to specify its height. Which is not optimal and forces us to create scrollable layouts where we can't exactly tell how large the nested layout is, see Account Selection in Browser.

## Implementation

Let's figure out a way to calculate the height of the bottom sheet based on the size of its nested content. So regardless of what we place, it will always wrap its size based on that. 

fig1

And if the content grows too large - and by that i mean more than the available screen height - only then we overflow the content inside so it scrolls. 

fig 2

## Acceptance Criteria

Bottom sheet calculates its height correctly based on the height of nested elements inside it. All currently used bottom sheets start to use this behaviour, hardcoded scroll is removed where used.

## Notes

This can work in tandem with this issue https://github.com/status-im/status-react/issues/9846
Thanks!

Ferossgp:
To test max height one can add multiple accounts and see how it changes inside browser account select.
Fixes #9848

### DIFF
--- a/src/status_im/ui/components/bottom_sheet/view.cljs
+++ b/src/status_im/ui/components/bottom_sheet/view.cljs
@@ -5,13 +5,13 @@
             [reagent.core :as reagent]
             [re-frame.core :as re-frame]))
 
-(def initial-animation-duration 300)
+(def initial-animation-duration 400)
 (def release-animation-duration 150)
 (def cancellation-animation-duration 100)
 (def swipe-opacity-range 100)
-(def cancellation-height 180)
+(def cancellation-coefficient 0.3)
 (def min-opacity 0.05)
-(def min-velocity 0.1)
+(def min-velocity 0.4)
 
 (defn- animate
   [{:keys [opacity new-opacity-value
@@ -32,13 +32,6 @@
                         :friction        6
                         :useNativeDriver true})])))
 
-(defn animate-panel-open [opacity-value bottom-value]
-  (animate {:bottom            bottom-value
-            :new-bottom-value  0
-            :opacity           opacity-value
-            :new-opacity-value 1
-            :duration          initial-animation-duration}))
-
 (defn- on-move
   [{:keys [height bottom-value opacity-value]}]
   (fn [_ state]
@@ -48,12 +41,12 @@
               (animation/set-value bottom-value dy)
               (animation/set-value opacity-value opacity))
             (neg? dy)
-            (animation/set-value bottom-value (/ dy 2))))))
+            (animation/set-value bottom-value dy)))))
 
-(defn cancelled? [height dy vy]
+(defn- cancelled? [height dy vy]
   (or
    (<= min-velocity vy)
-   (> cancellation-height (- height dy))))
+   (> (* cancellation-coefficient height) (- height dy))))
 
 (defn- cancel
   ([opts] (cancel opts nil))
@@ -68,18 +61,18 @@
                                      (when (fn? callback) (callback)))})))
 
 (defn- on-release
-  [{:keys [height bottom-value opacity-value on-cancel] :as opts}]
+  [{:keys [height bottom-value close-sheet opacity-value]}]
   (fn [_ state]
     (let [{:strs [dy vy]} (js->clj state)]
       (if (cancelled? height dy vy)
-        (cancel opts on-cancel)
+        (close-sheet)
         (animate {:bottom            bottom-value
                   :new-bottom-value  0
                   :opacity           opacity-value
                   :new-opacity-value 1
                   :duration          release-animation-duration})))))
 
-(defn swipe-pan-responder [opts]
+(defn- swipe-pan-responder [opts]
   (.create
    react/pan-responder
    (clj->js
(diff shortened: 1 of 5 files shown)


flexsurfer:
still wip?

Ferossgp:
@flexsurfer yes, there is one small bug with animation during pan responder. I'm thinking on how to fix it now.

src/status_im/ui/components/bottom_sheet/view.cljs
```diff
@@ -90,75 +83,116 @@
...
+                                   (when (fn? on-cancel)
+                                     (animation/set-value bottom-value height)
+                                     (animation/set-value opacity-value 0)
+                                     (reset! internal false)
+                                     (on-cancel)))})))
+
+(defn bottom-sheet-view [{:keys [window-height]}]
+  (let [opacity-value    (animation/create-value 0)
+        bottom-value     (animation/create-value window-height)
+        content-height   (reagent/atom window-height)
```
Ferossgp:
```suggestion
        content-height   (reagent/atom (* 0.4 window-height))
```
For a smoother animation on first time open

errorists (approved):

src/status_im/ui/components/bottom_sheet/view.cljs
```diff
@@ -90,75 +83,116 @@
...
-    :reagent-render
-    (fn [{:keys [opacity-value bottom-value height
-                 content on-cancel disable-drag?]
-          :or   {on-cancel #(re-frame/dispatch [:bottom-sheet/hide])}
-          :as   opts}]
-      [react/keyboard-avoiding-view {:style styles/container}
-       [react/touchable-highlight
-        {:on-press #(cancel opts on-cancel)
-         :style    styles/container}
+(defn- on-open [{:keys [bottom-value internal opacity-value]}]
```
flexsurfer: probably its better to use asterisks for atoms , like `*internal*` ?

src/status_im/ui/components/bottom_sheet/view.cljs
```diff
@@ -90,75 +83,116 @@
...
-                                      styles/border-radius
-                                      styles/bottom-padding)]
-          (when @show-sheet?
-            [bottom-sheet-view (assoc opts
-                                      :content content
-                                      :height total-content-height)])))})))
+(defn bottom-sheet [props]
+  [react/safe-area-consumer
+   (fn [insets]
+     (reagent/as-element
```
flexsurfer, line 194: why do we need `as-element` here ?

src/status_im/ui/components/bottom_sheet/view.cljs
```diff
@@ -90,75 +83,116 @@
...
+      [bottom-sheet-view (assoc props
+                                :window-height @(re-frame/subscribe [:dimensions/window-height])
```
flexsurfer, line 196: better to use `letsubs`

src/status_im/ui/screens/views.cljs
```diff
@@ -31,47 +31,33 @@
...
-                 (= view :learn-more)
-                 (merge about-app/learn-more)
-
-                 (= view :recover-sheet)
-                 (merge (recover.views/bottom-sheet)))
-          height-atom (reagent/atom (if (:content-height opts) (:content-height opts) nil))]
-      [bottom-sheet-comp opts height-atom])))
+(defn bottom-sheet []
+  (fn []
+    (let [{:keys [show? view]} @(re-frame/subscribe [:bottom-sheet])
```
flexsurfer: `letsusb`

src/status_im/ui/components/bottom_sheet/view.cljs
flexsurfer: not only here

flexsurfer (changes requested):

src/status_im/ui/components/bottom_sheet/view.cljs
Ferossgp: Because it is a react context consumer https://github.com/reagent-project/reagent/blob/ff4c9a6031712994ae92ac60fdd7df8ec77e92f8/doc/ReactFeatures.md#context

src/status_im/ui/components/bottom_sheet/view.cljs
Ferossgp: With the upgrade of reagent and re-frame I think this macro doesn't make anymore sens for components. If we would like to make it just shorter we can use the largely adopted by the community composition of deref and subscribe `<sub` [lamdaisland](https://lambdaisland.com/blog/2017-02-11-re-frame-form-1-subscriptions) which is just pure function. While keeping the defview for screens where we can add other screen-related meta information like `on-focus` etc.

flexsurfer: but i still think its better to move subscriptions outside renderer even if we don't want to use letsubs, `@(re-frame/subscribe` in renderer function doesn't look right

src/status_im/ui/components/bottom_sheet/view.cljs
yenda: Aren't' muffles traditionally used only for dynamic vars and not atoms?

src/status_im/ui/components/bottom_sheet/view.cljs
yenda: Isn't it possible to use the safe-area-view and avoid that consumer?

Ferossgp: In this case no, cause I need the value to use in calculations https://github.com/status-im/status-react/pull/9915/files/96f371faf9a254f0b0fe7821ebee23819e37cbbd#diff-04c3dd84b9db26e4b7433a9e25932009R127 but yes, for cases where the only padding is needed better to use `safe-area-view`

src/status_im/ui/components/bottom_sheet/view.cljs
flexsurfer: @yenda yeah probably, do you have any better idea? because it would be better if we could recognize atoms in code

yenda: Maybe an "-atom" suffixe, most of the time this doesn't happen as the atoms aren't passed by reference but by value instead so that could be a solution

src/status_im/ui/components/bottom_sheet/view.cljs
rasom: @flexsurfer it is fine to have subs in renderer function, we don't really need `letsubs` for quite a while already https://github.com/day8/re-frame/issues/218#issuecomment-264691621

flexsurfer: i know , but personaly i prefer letsubs, more clean and idiomatic , but its up to you guys how you want to use it

flexsurfer (approved):

rasom:
@Ferossgp give me please 1h for review, do not merge it now

rasom:
If I slowly slide bottom sheet down it suddenly disappears with no animation at all:

rasom (changes requested):
issue with animation on sliding sheet down should be fixed

Ferossgp:
@rasom this happens because the velocity is big enough to close, I could empirically increase it to 0.4, wdyt?
To be more precise this condition evals to `true` : https://github.com/status-im/status-react/pull/9915/files#diff-04c3dd84b9db26e4b7433a9e25932009R48

rasom:
The only problem (or at least difference comparing to what we have in `develop`) I see now is that it is possible to drag sheet up so that it leaves screen

`develop` (that's max top position I could drag sheet up to):


current branch:

rasom (approved):

churik:
Tested on IPhone 7(IOS 13) and Xiaomi Mi Note 9 (Android 10)

Agreed with https://github.com/status-im/status-react/pull/9915#issuecomment-591944244, don't think that's the blocker.

Tested with 11 accounts in mutiaccounts, 7 names in ENS names and with options in chat / browser.

errorists:
Agree, I don't think the overdrag issue is that big of a deal

PR #9917, opened 2020-01-24
componentizied input-search and search in set currency and token management

tbenr:
- added search feature when selecting default currency
- added search feature in asset management
- moved search-input to a dedicated ui component
- removed auto-correction and auto-capitalization in search input

image

image

**note for testing:** non-regression tests should be done on chat search:
image

status: ready

### DIFF
--- a/src/status_im/search/core.cljs
+++ b/src/status_im/search/core.cljs
@@ -1,5 +1,17 @@
 (ns status-im.search.core
   (:require [status-im.utils.fx :as fx]))
 
-(fx/defn filter-changed [cofx search-filter]
-  {:db (assoc-in (:db cofx) [:ui/search :filter] search-filter)})
+(fx/defn home-filter-changed
+  {:events [:search/home-filter-changed]}
+  [cofx search-filter]
+  {:db (assoc-in (:db cofx) [:ui/search :home-filter] search-filter)})
+
+(fx/defn currency-filter-changed
+  {:events [:search/currency-filter-changed]}
+  [cofx search-filter]
+  {:db (assoc-in (:db cofx) [:ui/search :currency-filter] search-filter)})
+
+(fx/defn token-filter-changed
+  {:events [:search/token-filter-changed]}
+  [cofx search-filter]
+  {:db (assoc-in (:db cofx) [:ui/search :token-filter] search-filter)})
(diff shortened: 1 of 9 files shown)


tbenr:
@rachelhamlin to set default currency as per device settings we need to move from `react-native-languages` to `react-native-localize` dependency, which takes more time. For now I implemented better way to set it :)

flexsurfer:
thanks @tbenr ! that's great! do you think  we could add search for assets management in this PR or in a separate one  ?

tbenr:
@flexsurfer this morning I realize it was missing there too :) let me add it in this pr

tbenr:
@flexsurfer done!

src/status_im/search/core.cljs
```diff
@@ -1,5 +1,17 @@
 (ns status-im.search.core
   (:require [status-im.utils.fx :as fx]))
 
-(fx/defn filter-changed [cofx search-filter]
+(fx/defn filter-changed
```
tbenr: @yenda maybe here and in other places could make sense to rename `filter-changed` and all other related events\keys to `chat-filtered-changed`. So we have consistency across all search filter stuff. WDYT?

errorists:
@tbenr thanks man! I wouldn't have it focused on entering the view tho. It's not critical but if possible I'd ask to change that

errorists (approved):

tbenr:
@errorists oky no prob!

src/status_im/search/core.cljs
tbenr: @yenda i did that :)

tbenr:
@errorists done!

flexsurfer (approved):

tbenr:
rebased

churik:
Thank you for contribution and apologies for the delay in testing.
Noticed one small issue: value in the search field in asset or currency management is erased while switching tabs.
It is not reproducible with the search field in chat.
To reproduce type some value in search in wallet > settings > manage assets, switch tab and go back to wallet.
Expected, that value will be preserved.
In fact, value is empty, but search results are filtered according to the previous query.
manage-asset

tbenr:
@churik ok yes i recreated it. give me some time to fix it, i'm pretty busy these days.

tbenr:
@churik should be fixed now, I rebased too.

churik:
Retested all search fields on IOS13 and Android8.
Looks good to me! 
Test case is added https://ethstatus.testrail.net/index.php?/cases/view/6269

PR #9922, opened 2020-01-27
Fix ens names in chat

cammellos:
We now store the clock value of the message and check against it, to make sure we don't verify old usernames.
status: ready

### DIFF
--- a/src/status_im/ens/core.cljs
+++ b/src/status_im/ens/core.cljs
@@ -278,19 +278,25 @@
  (fn [names]
    (verify-names (distinct names))))
 
-(defn should-be-verified? [cofx ens-name signature]
-  (and ens-name
-       (not (get-in cofx [:contacts/contacts signature :ens-verified]))
-       (not= signature (multiaccounts.model/current-public-key cofx))
-       (or (valid-custom-domain? ens-name)
-           (stateofus/valid-username? ens-name))))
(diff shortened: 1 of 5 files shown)


yenda (approved):

src/status_im/ens/core.cljs
```diff
@@ -278,19 +278,25 @@
...
+         (> clock-value ens-verified-at)
+         ; it's not us
+         (not= signature (multiaccounts.model/current-public-key cofx))
+         ; and it's a valid name
+         (or (valid-custom-domain? ens-name)
+             (stateofus/valid-username? ens-name)))))
+
+(fx/defn verify-names-from-message [cofx {:keys [content clock-value]} signature]
+  (when (should-be-verified? cofx (:ens-name content) signature clock-value)
     {::verify-names [{:name (:ens-name content)
```
yenda, line 299: it only ever checks one name at a time though?

cammellos: it's accumulated, through `fx/merge`, and then we aggregate them and send a single request.
Currently though, we process one message at a time, so it would only send a request at a time

flexsurfer:
@cammellos is this still relevant?

cammellos:
@flexsurfer no, forgot about this one, thanks!

PR #9924, opened 2020-01-27
Clarify wallet 'send' button

rachelhamlin:
The 'Send' button in the wallet generated Tx flow leads to the Tx overview, where you can edit network fee and check details.

I changed the button that leads to this screen from 'Send' to 'Next', so users know there is another step in the process.

Simon's screenshot:
image

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -1106,7 +1106,7 @@
     "wallet-manage-assets": "Manage assets",
     "wallet-offline": "Wallet offline",
     "wallet-request": "Request",
-    "wallet-send": "Send",
+    "wallet-send": "Next",
     "wallet-send-min-wei": "Min 1 wei",
     "wallet-settings": "Wallet settings",
     "wallet-total-value": "Total value",


translations/en.json
```diff
@@ -1106,7 +1106,7 @@
     "wallet-manage-assets": "Manage assets",
     "wallet-offline": "Wallet offline",
     "wallet-request": "Request",
-    "wallet-send": "Send",
+    "wallet-send": "Next",
```
flexsurfer, line 1109: we can't just change it in translation because we have same UI for sending in chat, but we send a message there

flexsurfer (changes requested):

translations/en.json
yenda, line 1109: this is not only going to affect the signing flow, this is also going to affect the send button in the wallet UI. The correct way to change this would rather be to use the "next" label in the bottom sheet instead of changing the meaning of this one

simonricoo (approved):

andmironov (approved):

flexsurfer:
@rachelhamlin do we have an issue?

rachelhamlin:
We do now @flexsurfer :)  https://github.com/status-im/status-react/issues/9974

PR #9935, opened 2020-01-28
Correctly dispatch on updates & disable them

cammellos:
On moving updates to status-go we removed the dispatch on the rpc request, and therefore updates were published too often. should have listen to @yenda more carefully ;) 
this commit fixes the issue and disables them for now, as we don't really need them (ens is not resolved in contact updates).

status: ready

### DIFF
--- a/src/status_im/multiaccounts/update/publisher.cljs
+++ b/src/status_im/multiaccounts/update/publisher.cljs
@@ -1,5 +1,6 @@
 (ns status-im.multiaccounts.update.publisher
   (:require [taoensso.timbre :as log]
+            [re-frame.core :as re-frame]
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

Serhy:
I'll restart failed e2e (just about to merge the fixes for them it's not this PR related) 
@cammellos what exactly should keep working here (if I need to check manually on real devices)?

cammellos:
@Serhy nothing, it's just disabling a feature, so should work as before (there was a bug but it's difficult to replicate, you have to trigger using RC2 -> upgrade to RC3  and check the logs, so not worth it, as the feature is now disabled)

PR #9939, opened 2020-01-29
Fixes cryptokitties loading/ordering

### LINKED ISSUE #6839: A limit of 20 CryptoKitties is displayed in the collectibles list in the wallet with incorrect order

krzyzowiec:

Fixes #6839 

### Summary

This is a completion of/improvement upon the following [work](https://github.com/status-im/status-react/pull/9810). The kitty ids are grabbed in groups of 20 and sent to :load-kitties to be loaded individually. The problem of ordering is not due to any strange behavior on the part of mapv, but due to the perfectly normal behavior of a plain hash map. It doesn't care about ordering.

This was a bit tricky to diagnose because the :CK collectible hash doesn't exist at db initialization, it is created when :load-collectible-success decides to merge the first kitty into the db under the :CK symbol and sees that it doesn't exist, after which merge creates a hash-map for the non-existent key. (this is the default behavior of merge)

To fix this, I update db in the :load-kitties handler to ensure that the :CK key is initialized to an empty ordered hash map, which I set to descending order. Here I am following the cryptokitties API, which returns the most recently created kitties first. (a custom function can be supplied to change the ordering if necessary, as I have no idea what order they use to display their cats on their site)

### Testing notes
Testing this in the app was a bit tricky. I loaded some kitties manually in the repl, but it would be good for someone to check this with actual kitties on their wallet account.

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- network consumption

### Steps to test

Open status application
Navigate to the wallet tab
Expand a wallet with more than 20 kitties
Expand the collectibles tab and select kitties
Ideally see all kitties in the proper order

status: ready


src/status_im/wallet/collectibles/core.cljs
```diff
@@ -54,7 +54,7 @@
 (handlers/register-handler-fx
  :load-kitties
  (fn [{db :db} [_ ids]]
-   {:db db
+   {:db (update-in db [:collectibles] merge {ck (sorted-map-by >)})
```
flexsurfer, line 57: could you elaborate on why we need this? thanks

flexsurfer (approved):

src/status_im/wallet/collectibles/core.cljs
krzyzowiec: The original bug report mentioned that the displayed order of the cats is different than on the website. That's because they are unsorted.

If you mean why that code in that precise spot, it's because you must have the sorted map there first, before you add any cats, if you want to maintain a sorted order after the merges. Putting it there seemed like the least obtrusive way to do it. (I don't have to touch the initialization, and every other function stays the same)

Did you try it and compare?

Serhy:
Hey @krzyzowiec , builds failed due to incorrect formatting (https://ci.status.im/blue/organizations/jenkins/status-react%2Fprs%2Fandroid/detail/PR-9939/1/pipeline )
Could you look at it, please?

PR #9940, opened 2020-01-29
Pass extra argument to handle multiaccounts on logout

### LINKED ISSUE #9902: Make Sign out go to Multiaccount overview 'Your keys' screen
# Problem
When the user has set `Save password`, going to the `unlock view` can result in an endless loop of sign out>unlock, making it impossible to switch accounts. This scenario is currently counteracted (does not occur), because auth-methods are erased on `Sign out`. However, erasing auth-methods on sign out causes other issues:
- Requiring Keycard to open a Keycard multiaccount, in spite of having set `Don't ask for Keycard`, see #9763
- The user having to re-enable biometric unlock (fingerprint or faceID) without being informed that they are required to do so when they sign out

## Implementation
To resolve these issues, allow maintaining the stored auth-method, `Sign out` should go to the `Your keys` (multiaccount overview) screen.

See future steps, current key encryption and storage needs to be investigated. If keys are encrypted with the password, rather than a key generated by status-react (as proposed [here](https://docs.google.com/document/d/1IrjvvAuAQbYsr3xqZvXfgAeeyDZRl_x9KRxWnLcwVuI/edit#bookmark=id.h09ezqb1viei)), credentials (password?) might be required. Possibly auth-method is erased to prompt password.

#### Desired flow
Sign out > Select keys to unlock > Unlock screen
Frame 26

## Acceptance Criteria
`Sign out` opens `Your keys`

## Future Steps
- Depending on how keys are encrypted and stored, create issue to investigate key encryption according to [proposal](https://docs.google.com/document/d/1IrjvvAuAQbYsr3xqZvXfgAeeyDZRl_x9KRxWnLcwVuI/edit#bookmark=id.h09ezqb1viei)
- Depending on whether keys can be encrypted and stored according to proposal, investigate maintaining stored auth-method

Ferossgp:
Fixes #9902

On the first time opening an account will show an animation as navigating back. This happens because the stack is initialized as a login screen as the first element and multi-account screen as the second. I would suggest approaching this the way making always multiaccount screen main, and when the app is open from background to navigate to the login screen of latests account.

### DIFF
--- a/src/status_im/init/core.cljs
+++ b/src/status_im/init/core.cljs
@@ -32,18 +32,24 @@
               :view-id view-id)})
 
 (fx/defn initialize-views
-  [cofx]
+  [cofx {:keys [logout?]}]
   (let [{{:multiaccounts/keys [multiaccounts] :as db} :db} cofx]
-    (if (empty? multiaccounts)
+    (cond
+      (empty? multiaccounts)
       (navigation/navigate-to-cofx cofx :intro nil)
+
+      logout?
+      (navigation/navigate-to-cofx cofx :multiaccounts nil)
+
+      :else
       (let [{:keys [key-uid public-key photo-path name]} (first (#(sort-by :last-sign-in > %) (vals multiaccounts)))]
         (multiaccounts.login/open-login cofx key-uid photo-path name public-key)))))
 
 (fx/defn initialize-multiaccounts
   {:events [::initialize-multiaccounts]}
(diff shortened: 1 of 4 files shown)


flexsurfer (approved):

hesterbruikman (approved):

Serhy:
looks good to me with iOS and Android. e2e fixed

PR #9942, opened 2020-01-29
[#9927] Fast blocks sync after being offline

### LINKED ISSUE #9927: Transaction history is delayed in status
# Bug Report

## Problem

When sending transaction in status there is no confirmed transaction in history , but on etherescan it has  a few confirmations already

#### Expected behavior

Confirmed transaction appears in transactions history

rasom:
fix #9927

Front end changes:
 As there is no guarantee that `newblock` event will be dispatched 
consequently (e.g. if there was a delay after block#1 the next event
might be dispatched for a block#2000, no guarantee that block#2 will be
the next one), `newTransactions` field was added with a map of accounts
to the number of new transactions received to this block. In result if
there are new transactions we request them all of db instead of fetching 
them on for the block specified in `newblock` event, as it was done 
previously.  

Back end changes:
- In order to avoid handling of the reorganized blocks we use an offset 
from the latest known block when start listening to new blocks. Before 
this commit the offset was 15 blocks for all networks. This offset is 
too big for mainnet and causes noticeable delay of marking a transfer as 
confirmed in Status (comparing to etherscan). So it was changed to be 5
blocks on mainnet and is still 15 blocks on other networks.
- Also before this commit all new blocks were handled one by one with 
network specific interval (10s for mainnet), which means that in case of
lost internet connection or application suspension (happens on iOS) 
receiving of new blocks would be paused and then resumed with the same 
"speed" - 1 blocks per 10s. In case if that pause is big enough the 
application would never catch up with the latest block in the network, 
and this also causes the state of transfers to be delayed in the 
application. In this commit in case if there was more than 40s delay 
after receiving of the previous block the whole history in range between 
the previous received block and ("latest"-reorgeSafetyDepth) block is 
checked at once and app catches up with a recent state of the chain.

status: ready


yenda:
@rasom could you describe what was changed and how it fixed the issue?

 looks like it works better for me now, I tried on a recover account and sent funds and the transaction was updated shortly after etherescan shows the update which was not the case before. To verify I also went back to RC4 and sent a transaction after starting the app (killed for a while, no background) and it still shows as pending  when etherscan already confirmed 5 blocks)

rasom:
@yenda 
changes on go side:
- there is a [`reorgSafetyDepth`](https://github.com/status-im/status-go/blob/a92a95cf83568cca01353f5daa1f29228f5018bd/services/wallet/reactor.go#L33) constant, which is used as an offset from the latest block, so that we start listen to the new blocks starting from `"latets"-reorgSafetyDepth`. It is added to avoid fetching of blocks which will be reorganised, as I understand. It was 15, which is probably fine for testnet networks, but is a bit too big for mainnet. So I changed it to 5, and that's the reason why transactions appear faster right after sending.
- another change is that in case if the next block is received after a noticeable delay (it may happen if app is suspended on iOS or when you loose internet connection), instead of fetching next blocks one by one we scan the range `last-recieved-block=>("latest"-reorgSafetyDepth)` using the same mechanism we have for tx history. For instance, if you are 200 blocks behind "latest", currently in `develop` it will take not less than 2000s to fetch these blocks, and during this time new blocks will be added. In result you are likely to stay quite far from "latest" block unless your app is always online and in foreground. I set that delay threshold to 60s but it can be 30-40s.

yenda:
can you add the explanation in the commit message?
is it still a wip?

flexsurfer (approved):

rasom:
@yenda yes it is
of course I will add explanation to commit message

rasom:
@yenda @flexsurfer ready for review and testing, message is added

flexsurfer (approved):

yenda (approved):

yenda:
when did we stop supporting reorg of blocks and have this delay instead?

rasom:
@yenda we never stopped to support reorgs, what do you mean?

rasom:
@yenda as far as I understand the idea of that safetyDepth thing was rather to minimise the occurrences of reorgs in the app to avoid "flickering" state there, that's the best I can suggest (it seems that I have no one to ask about details). But we still handle reorgs if necessary.

Serhy:
Works great for me on Android 8, iOS 13: sent tokens and ETH from 1-1 chat / Wallet appear in Wallet history according to the confirmations I compared in Etherscan. No delay, (well maybe +- 1 block which is absolutely fine). 
Balance is updated, 1-1 chat outgoing and incoming TX status picked up fast enough (right when TX appears in history).
Amazing update!

PR #9945, opened 2020-01-29
Rework and update react-navigation

Ferossgp:
Uses react-navigation 5 and removes the legacy navigation code.

- [x] Status bar color in QR scanner
- [x] Insets colors in QR scanner
- [x] Insets on screens without the bottom tab-bar.
- [x] Keyboard avoiding view iOS
- [x] Back handler to be revised, now it is too easy to forget adding back handler
- [x] Android tab bar elevation is not shown on the open state (because it is translated outside of the container, android does not allow rendering outside of parent view)
- [x] Debug screen interaction right after navigation event. Seems like the screen does not handle touches while the animation is outside of the screen.
- [x] navigation-stack to be removed from re-frame

The separate tasks to be added:
- [ ] Back handler special cases https://github.com/status-im/status-react/issues/9992
- [ ] Fix two-pane navigation
- [ ] Connect animation value for screen transition and bottom tab bar
- [ ] Animate top margin on the bottom tab bar


vkjr:
@Ferossgp, probably you know, we have custom two-pane navigator that should be used for desktop and for tablets. I'm curious if its functionality survived upgrade to react-navigation 4? :)

Ferossgp:
@vkjr UI wise it works, but it should be updated cause navigation object structure was changed, so it throws error during navigation events.

deps.edn
```diff
@@ -8,12 +8,12 @@
                                       :sha           "b0d3b0f016f217c6a397a54e5eba9486471fb172"
                                       :deps/manifest :deps}
          com.andrewmcveigh/cljs-time {:mvn/version "0.5.2"}
+         binaryage/oops              {:mvn/version "0.7.0"}
```
flexsurfer: whats this ? why do you need it ?

react-native/src/mobile/status_im/react_native/js_dependencies.cljs
```diff
@@ -4,6 +4,7 @@
 (def fs                     (js/require "react-native-fs"))
 (def keychain               (js/require "react-native-keychain"))
 (def qr-code                (js/require "qrcode"))
+(def react                  (js/require "react"))
```
flexsurfer, line 7: why do you need react ?

src/status_im/ios/core.cljs
```diff
@@ -38,7 +39,7 @@
                       #(when-not (= 0 @keyboard-height)
                          (dispatch [:set :keyboard-height 0])))
         (.hide react/splash-screen)
-        (.useScreens rn-dependencies/react-native-screens)
+        (ocall rn-dependencies/react-native-screens "enableScreens")
```
flexsurfer: are you sure we need this with rn5 ?

src/status_im/android/core.cljs
```diff
@@ -41,8 +42,7 @@
                         (when-not (zero? @keyboard-height)
                           (dispatch [:set :keyboard-height 0]))))
         (.hide react/splash-screen)
-        ;; TODO Temporarily comment away due to current bug https://github.com/kmagiera/react-native-screens/issues/54
-                                        ;(.useScreens rn-dependencies/react-native-screens)
+        ;; (ocall rn-dependencies/react-native-screens "enableScreens")
```
flexsurfer: can we remove this ?

src/status_im/chat/models/loading.cljs
```diff
@@ -38,6 +38,7 @@
               (filters/load-filters))))
 
 (fx/defn offload-all-messages
+  {:events [::offload-all-messages]}
```
flexsurfer, line 42: how is this related to PR?

src/status_im/ui/screens/routing/main.cljs
```diff
@@ -0,0 +1,87 @@
...
+    {:name      :qr-scanner
+     :no-wrap   true
+     :component qr-scanner/qr-scanner}]])
+
+(defn get-main-component [_]
+  (fn []
+    [main-stack {:header-mode :none
+                 ;; :mode        :modal
+}
+     [(if @(re-frame/subscribe [:multiaccount/logged-in?])
```
flexsurfer: not sure about this, could you elaborate

flexsurfer (approved):
AWESOME JOB!!

src/status_im/chat/models/loading.cljs
Ferossgp: This was called during `screens/on-will-focus` because this event is not called anymore I moved this into a separate event. It is called during the focus on home view.

src/status_im/ui/screens/routing/main.cljs
Ferossgp: Because navigation now can be initialized dynamically, we can render different stacks conditionally. This if is equivalent to switch navigator from older react-navigation.

src/status_im/ios/core.cljs
Ferossgp: We can remove it, an artifact from rn4 :+1:

errorists:
@Ferossgp seeing as not all checkmarks are checked, should I review it now or wait till it's finished

Ferossgp:
@errorists you could check it now, I updated the PR description

src/status_im/ui/components/status_bar/view.cljs
```diff
@@ -3,8 +3,11 @@
             [status-im.ui.components.status-bar.styles :as styles]
             [status-im.utils.platform :as platform]))
 
+;; TODO: Integrate into navigation
 (defn get-config [view-id]
-  (or (get {:qr-scanner {:type :black}}
+  (or (get {:qr-scanner      {:type :black}
+            :new-chat        {:type :black}
```
Ferossgp: Should be black only on ios because it uses presentation modal.

churik:
## ISSUE 1: Can't Start app on IOS
**Steps:**
- open status

**Expected result:** app is started

**Actual result:** photo_2020-02-27 13 04 53

flexsurfer:
please do not merge before v.1.1 is cut

src/status_im/hardwallet/core.cljs
```diff
@@ -19,6 +19,7 @@
             [status-im.multiaccounts.update.core :as multiaccounts.update]
             [status-im.multiaccounts.recover.core :as multiaccounts.recover]))
 
+;; FIXME(Ferossgp): Separate into multiple components with EXPLICIT back action
```
Ferossgp: Fixed here: https://github.com/status-im/status-react/pull/10015/commits/09fcc1d44bcca31ac6fc9b1d41002c8b30f1e9ad

errorists (approved):

src/status_im/ui/screens/qr_scanner/views.cljs
```diff
@@ -44,15 +44,16 @@
...
-    [react/view {:style {:flex 1 :background-color colors/black}}
+    [react/safe-area-view {:style {:flex             1
+                                   :background-color colors/black}}
      [topbar camera-flashlight opts]
      [react/with-activity-indicator
       {}
       [camera/camera
        {:style         {:flex 1}
-        ;:torchMode     (camera/set-torch camera-flashlight)
+                                        ;:torchMode     (camera/set-torch camera-flashlight)
```
rasom: this one should probably babe removed

rasom (approved):

Serhy:
**Issue 1**
Any option from the list in `Wallet` -> `Add account` (Generate /  Watch only / Seedphrase/  Private key) opens weird half-empty screen:
Screenshot 2020-03-10 at 13 32 37
Nothing happens if to enter valid values and tap `Add account`

**Issue 2**
Back handler special case from #9992 still reproduced. 
Hardware back button on Android basically does not close any QR screen (QR code pop-up). List of view where QR code pop-up appears:
2a) Contact profile view: `Open 1-1 chat` ->  tap top-right `...` menu -> tap `View profile` -> tap contact element under `Profile details`
2b) Tap `Profile` tab -> top-right `Share` button
2c) `Wallet` -> `Share` button on account (`Status account`)
2d) `Wallet` -> `Status account` -> `Receive`
2e) `Wallet` -> `Status account` -> `Share` button in the top-right card corner

**Issue3**
**On iOS only** After the account creation or recover if to Log Out then I appear on "Welcome to Status Let's go" view. Not reproducible if I close the app -> login to acc -> Log out, - but only in a case user went through onboarding screens in the the same session he tries to logout. 

To reproduce:
1) Open Status
2) Create new account (or Recover account with seed phrase)
3) Log out via `Profile` -> `Sign out`  -> `Log out`
**Actual result:** User winds up on `Welcome to Status` view and nothing happens when tap 'Let's go'
LogoutLetsGO (1)


**Issue4**
Also like in issue4, - maybe with the same root cause, - it's on iOS only and only after account creation/recover (so when when user went through onboarding screens): weird lag when navigating out of `Start new chat` / `Join public chat` views (which are now bottom-sheet views)
LAgsBotmNav

**Issue5**
There is an 'Undefined' account screen appears for a short time when login in or login out of account:
Please see video https://drive.google.com/file/d/1nKRuOA_Gl1uK6qmVdFi1PYT3LmqfT4mQ/view?usp=sharing (3rd and 9th seconds). 
iOS and Android affected, though depends of device ( on iPhoneX it's almost blinking but visible)

**Issue6**
Im not sure about preconditions with this issue, but I've got it several times when the app crash right after opening `Status`. Crash with the error message `Cannot read property 'options' of undefined`
Logs are here: 
[Status.log](https://github.com/status-im/status-react/files/4312497/Status.log)
Simply opened Status (not from background but when app is closed) by tapping it's icon.
I can't guarantee it's not in develop because in latest nightly with several attempts I could not reproduce, but if there are clues in logs why it may happen that would be helpful to know. Otherwise will test after above 5 issues fixed (maybe this crash will be fixed along the way)

Similar to that 6th issue  (maybe even the reason of Issue6) i've got 
Screenshot 2020-03-10 at 16 17 47
It appeared twice after I cleared app data and opened Status.
To try to reproduce:
1) Install Status on Android (v8.1 in my case)
2) Open Status app Info settings from list of app on your device
3) Then `Storage`-> `Clear Data`
4) Open Status

Serhy:
@Ferossgp there is a typo in formatting (`1 file(s) formatted incorrectly`)

mobile/js_files/package.json
```diff
@@ -27,23 +31,23 @@
...
-    "react-native-gesture-handler": "^1.3.0",
+    "react-native-gesture-handler": "^1.5.3",
     "react-native-image-crop-picker": "^0.25.0",
     "react-native-image-resizer": "git+https://github.com/status-im/react-native-image-resizer.git#1.0.0-1-status",
     "react-native-keychain": "git+https://github.com/status-im/react-native-keychain.git#v.3.0.0-status",
     "react-native-languages": "^3.0.2",
     "react-native-mail": "git+https://github.com/status-im/react-native-mail.git#v4.0.0-status",
     "react-native-navigation-twopane": "git+https://github.com/status-im/react-native-navigation-twopane.git#v0.0.2-status",
+    "react-native-reanimated": "^1.7.0",
     "react-native-safe-area-context": "^0.6.0",
```
yenda: There is a 0.7.3 that fixes the bug with android

Ferossgp:
TODO: Add on-blur for screens to handle this case https://github.com/status-im/status-react/pull/10168

Serhy:
Everything is good with Android and iOS 13 to me! All issues fixed reported above but:
Hardware back button special cases will go with https://github.com/status-im/status-react/pull/10173 separately
Group chat screens will be updated in https://github.com/status-im/status-react/pull/10099

PR #9946, opened 2020-01-29
Explicitly start messenger

cammellos:
We should only start receiving messages/processing topics once all the
initializiation is completed, so an endpoint that start the messenger
has been added.
Before we would initialize the topics/receiving messages at the same time,
which resulted in a race condition where a topic was signaled from
status-go and was considered new just because was not loaded yet.

status: ready


yenda (approved):

src/status_im/transport/core.cljs
```diff
@@ -34,10 +35,19 @@
 (fx/defn fetch-node-info-fx [cofx]
   {::fetch-node-info []})
 
+(fx/defn init-messenger [cofx]
```
yenda: would be worth adding here as a NOTE why this is all necessary

cammellos: done

flexsurfer (approved):

PR #9949, opened 2020-01-30
Add ENV option to enable waku-mode

cammellos:
Adds option to enable waku-mode through ENV variable.

status: ready


flexsurfer (approved):

status-go-version.json
```diff
@@ -2,7 +2,7 @@
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.42.0",
-    "commit-sha1": "9cf640842b4d0266bcc1b30f6d776c250fdb5c04",
-    "src-sha256": "0h8da2w3fbgrv6rmd4pib46r97g5q2wwg4a790f91dadsl67pmcl"
+    "version": "v0.41.1",
```
flexsurfer: in develop it's 42, should it be 42.1 ?

cammellos: It's a completely different version, not associated to the tag. (PR in status-go is not merged yet) https://github.com/status-im/status-go/pull/1832 , I will update to the correct version once it gets in

src/status_im/commands/core.cljs
```diff
@@ -23,20 +24,26 @@
   {:events [::accept-request-address-for-transaction]}
   [{:keys [db]} message-id address]
   {:db (dissoc db :commands/select-account)
-   ::json-rpc/call [{:method "shhext_acceptRequestAddressForTransaction"
+   ::json-rpc/call [{:method (if config/waku-enabled?
+                               "wakuext_acceptRequestAddressForTransaction"
+                               "shhext_acceptRequestAddressForTransaction")
```
jakubgs, line 29: I'm no Clojure wizard but it seems to me like all of thse `if config/waku-enabled?` could be replaced with something like:
```clojure
get-rpc-call-prefix + "acceptRequestAddressForTransaction"
```
Not sure how concatenation works in Clojure.
or
```clojure
get-rpc-call "acceptRequestAddressForTransaction"
```

PR #9950, opened 2020-01-30
attempt to fix #9942

rasom:
Storing absolute path for different configs breaks compatibility on iOS as app's dir is changed after upgrade. The solution is to store relative paths and to concatenate it with backend.rootDataDir. The only exception is LogFile as it is stored outside backend.rootDataDir on Android. LogDir config was added to allow adding of custom dir for log file.
Configs concerned:
DataDir
LogDir
LogFile
KeystoreDir
BackupDisabledDataDir

status: ready


yenda (approved):

flexsurfer (approved):

Serhy:
Not sure if it's ready yet, but I've tested upgrade on iOS and Android build from 01st February nightly on to nightly build made off current PR branch (https://ci.status.im/job/status-react/job/manual/597/)

After iOS upgrade I can login to account fine, but can not send any transaction: Status displays `Transaction failed` pop-up

Screenshot 2020-02-03 at 12 54 43

After Android (8.1) upgrade I could not login to account :( 

Screenshot 2020-02-03 at 13 06 00

flexsurfer:
@Serhy you should test only upgrades on top of this branch, for example , upgrade 11 build on top of 10

flexsurfer:
tested on ios, upgrade 11 build on top of 10, sign in works, transactions work

Serhy:
Aha, I got it now, thanks @flexsurfer 
Yes, iOS and Android, - upgrade from 10th Jenkins to 11th Jenkins build works, can login, exchange messages and send transactions

rasom:
it works fine but I still haven't fixed tests on go side, working on this...

PR #9957, opened 2020-01-31
[DONT MERGE]Temporary hide spinners for v1

flexsurfer:
Temporary fix for https://github.com/status-im/status-react/issues/9956 and https://github.com/status-im/status-react/issues/9928

Temporary hide spinners for v1

### DIFF
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -478,7 +478,7 @@
            content message-type] :as message}]
   (when (not= constants/message-type-private-group-system-message message-type)
     (case outgoing-status
-      :sending  [message-activity-indicator]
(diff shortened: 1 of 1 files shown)


cammellos (approved):

rachelhamlin:
I'm not a fan of this idea because:

1) In the majority of cases spinners will probably work as they should. 
2) It uses a shortcut to solve an occasional problem with a global change—better wait on a solution to the underlying issue.
3) Creates a new UI chore for us to track and remember later on.
4) Design doesn't like it :P and we should indeed make a practice of thinking through UI decisions rather than rushing them.

errorists:
> Design doesn't like it :P

just to elaborate on that, for the same reasons you do :) I don't want the imminent release date to force our hand and remove something that serves a purpose. feels like throwing the baby out with the bathwater, imo

yenda (approved):

PR #9958, opened 2020-01-31
[#9954] Add 'view profile' to message long tap

### LINKED ISSUE #9954: Add 'view profile' to message long tap
# Problem

This is being done largely as a counter spam mechanism, so that users can easily load up a person's profile and subsequently block them if they are flooding a chat. 

Also, you can not currently long tap a sticker, because the `reply` and `copy` UI is not built yet. The `view profile` option is valid for stickers now though.

Note: labeled v1 release, but not a v1 hard blocker.

## Implementation

Use the same menu option that we do in overflow menus:

image

## Acceptance Criteria

1. New option to `view profile` is available when you long tap on a message
2. When you long tap on a sticker, you get the `view profile` option only

flexsurfer:
fixes #9954

Stickers
image

Text
image

### DIFF
--- a/src/status_im/ui/components/list_selection.cljs
+++ b/src/status_im/ui/components/list_selection.cljs
@@ -13,14 +13,19 @@
             (:url content))
     (.share react/sharing (clj->js content))))
 
-(defn- message-options [message-id text]
-  [{:label  (i18n/label :t/message-reply)
-    :action #(re-frame/dispatch [:chat.ui/reply-to-message message-id])}
-   {:label  (i18n/label :t/sharing-copy-to-clipboard)
-    :action #(react/copy-to-clipboard text)}
-   (when-not platform/desktop?
+(defn- message-options [message-id text from outgoing]
+  (into
+   []
+   (concat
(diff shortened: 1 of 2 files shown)


errorists (approved):

flexsurfer:
@Serhy could you take a quick look ?

Serhy:
@flexsurfer can we not show `View profile` option if I long-tap own message? OR as an option to show but redirect to `Profile` main tab in that case (like we do if to open universal link with own ENS/public key)
Because now I can open profile screen where Send Message/Add to contacts has no effect and can block myself in chat :)

Everything else looks good to me on iOS and Android 8!

errorists:
oh absolutely. long press your own message, you don't get to see 'View profile'. thanks for spotting @Serhy

rachelhamlin:
Agreed with @Serhy. Otherwise great work! 👏

flexsurfer:
thanks, @Serhy fixed

Serhy:
Lgtm, no issues, e2e rechecked manually (one of it already updated in develop)

PR #9959, opened 2020-01-31
Git ignore trace.edn and config.cljs

Ferossgp:
They are auto generated and should not be present in git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -183,3 +183,5 @@ status-modules/resources
 /fiddle/node_modules/
 /fiddle/target/
 /fiddle/resources/public/images/
+
+trace.edn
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -183,3 +183,5 @@ status-modules/resources
 /fiddle/node_modules/
 /fiddle/target/
 /fiddle/resources/public/images/
+
+trace.edn
```
flexsurfer, line 187: what is trace.edn?also we need to add config.cljs as well?

Ferossgp: It was added, unfortunately the file was not deleted and git kept tracking it. https://github.com/status-im/status-react/commit/591e568e373dd05125c49ca4e4d856f650c7041a

Ferossgp: trace.edn is also generated on some crashes with trace of error.

flexsurfer (approved):

PR #9964, opened 2020-02-03
remove event duplication

flexsurfer:
remove event duplication

### DIFF
--- a/src/status_im/wallet/core.cljs
+++ b/src/status_im/wallet/core.cljs
@@ -505,47 +505,26 @@
                                         [to-norm amount-hex])})})))))
 
 (fx/defn sign-transaction-button-clicked
-  {:events  [:wallet.ui/sign-transaction-button-clicked]}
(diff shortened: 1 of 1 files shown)


src/status_im/wallet/core.cljs
```diff
@@ -542,10 +519,12 @@
                                  {:to    to-norm
                                   :value amount-hex}
                                  {:to       (ethereum/normalized-hex address)
-                                  :from     from-address
                                   :data     (abi-spec/encode
                                              "transfer(address,uint256)"
-                                             [to-norm amount-hex])}))}))))
+                                             [to-norm amount-hex])
+                                  ;;Note: data from qr (eip681)
```
yenda: these two should be in the first map that is merged because they should be added in both cases

flexsurfer: mhm, probably you're right, thanks

flexsurfer: @yenda  done

yenda (approved):

PR #9995, opened 2020-02-06
Offload chat messages

cammellos:
1) Messages are offloaded from any chat once we go back from the home.
    This allows us to ignore any message that is coming in from a chat we
    are not currently focused.
2) After 5 seconds of not-scrolling activity, any received message that
    is not currently visible will be offloaded to the database.
3) Similarly received messages that are not visible will be offloaded to
    the database directly

This should improve performance in the following areas:
1) Receiving messages when on the home should be much faster and less slow downs on the app
2) Joining a chat with a lot of messages (i.e spam) should be much faster, as long as the user does not keep scrolling up as messages become visible

The drawback is that when you go from home -> to a chat, you won't see the messages immediately (it will always be the same as it is now when you first go to a chat after login, as messages will be loaded from the database), but this is likely necessary (or will be changed) anyway once we implement scroll to last not seen message.

status: ready


Serhy:
@cammellos 
I tested on Galaxy Note 4 (Android 6.0.1)
Yes, as you @cammellos stated receiving messages when on the home is faster (just don't see whether it slowing less downs on the app because nightly is responsive as well when navigating between tabs while messages are receiving):
Receiving 100 of stickers in public chat from time the unread counter is 1 till it's 99+ takes:
Nightly build:  ~14s (video: https://drive.google.com/file/d/125S5alyBtIzNcvL4o5S4DMQ1rXmyxr24/view?usp=sharing)
PR build: ~11.5s (video: https://drive.google.com/file/d/15pnqb3gIK5J5hbm_Sc1oN7qhxZOHggQN/view?usp=sharing)

The same is on navigating to chat - redirect happens faster, though there is a delay on message appearing (not noticeable actually so fine)


This item is not really clear to me:
* After 5 seconds of not-scrolling activity, any received message that
is not currently visible will be offloaded to the database.
So I opened a chat with 100 messages. Scrolled up to ~60th message. Waited ~15-20 sec. Then from another device I send text messages to this public chat. Staying in public chat it's visible messages are shifted up (the "dancing messages" bug when receiving message while view is in the middle of chat).
Is that what 5-seconds not scrolling activity suppose to do?

cammellos:
@Serhy thanks for checking
> After 5 seconds of not-scrolling activity, any received message that
Is a bit different, basically you open a chat with 100 messages for example, scroll up, then scroll down to the latest message, wait 5 seconds, scroll up again. You should notice that those messages are re-fetched from the database. This helps with memory consumption and rendering performance when receiving messages while you are on the chat.

Hope is clear, thanks for taking the time!

Serhy:
> Hope is clear, thanks for taking the time!

Aha, I've got it, @cammellos !
Yes, actually this is the best part for me of this improvement :) 
It's really noticeable when having 100+ messages (like join chat with 100+ new messages - scroll up like you read all of them and then when messaging in this same chat - there is a lag upon sending a message (due to we re-render all the message list)
But with 5-second non activity we almost always have to re-render last 20 messages and that's much better UX

errorists:
@cammellos tested on Galaxy S8+, loaded both `#status` and `#chitchat` started scrolling as they were downloading messages. Opening chats works as expected, scroll is buttery smooth, didn't notice any lag or hiccup or other visual artefacts. Receiving messages on home looks ok, hard for me to compare if it's better or worse. anything else I should test?

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -334,6 +344,7 @@
                                           :idx                idx
                                           :list-ref           messages-list-ref}])
            :inverted                  true
+           :onViewableItemsChanged    on-viewable-items-changed
```
yenda, line 347: are the default good enough?
there is a bunch of visibility config for this for instance

```
:viewability-config (clj->js {:waitForInteraction false
                                                           :minimumViewTime 100
                                                           :viewAreaCoveragePercentThreshold 60})
```

src/status_im/ui/screens/chat/state.cljs
```diff
@@ -0,0 +1,3 @@
+(ns status-im.ui.screens.chat.state)
+
+(defonce viewable-item (atom nil))
```
yenda, line 3: shouldn't this be cleaned after logout?

cammellos:
@errorists that should be enough, thanks!

src/status_im/ui/screens/chat/state.cljs
cammellos: I can probably clean it up as we land on a chat, it's probably safer, but should get reset as soon as we load messages

src/status_im/ui/screens/chat/views.cljs
cammellos: They should be ok

src/status_im/chat/models/loading.cljs
```diff
@@ -31,29 +36,81 @@
...
+            chat-id (:chat-id (.-item last-element))]
+        (when (and last-element-clock-value
+                   (get-in db [:chats chat-id :messages-initialized?]))
+          (let [new-messages (reduce-kv (fn [acc message-id {:keys [clock-value] :as v}]
+                                          (if (<= last-element-clock-value clock-value)
+                                            (assoc acc message-id v)
+                                            acc))
+                                        {}
+                                        (get-in db [:chats chat-id :messages]))]
+            (fx/merge cofx
```
flexsurfer: `fx/merge` and `-> db` looks redundant here

flexsurfer (approved):

PR #9997, opened 2020-02-07
Check for empty public key

cammellos:
We check that ENS does not return an empty public key. In addition to
that validation is done on status-go in order to avoid saving the chat.

status: ready


rasom (approved):

src/status_im/ui/screens/add_new/new_chat/db.cljs
```diff
@@ -9,9 +9,12 @@
   [{:keys [multiaccount]} public-key]
   (= (:public-key multiaccount) public-key))
 
+(def empty-public-key "0x0400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")
```
flexsurfer: you can use this one https://github.com/status-im/status-react/blob/67f64f38168bd989b811a8f0e76d14dde79da6dd/src/status_im/ethereum/ens.cljs#L20

flexsurfer (changes requested):

src/status_im/ui/screens/add_new/new_chat/db.cljs
cammellos: Updated, thanks!

flexsurfer (approved):

yenda (approved):

PR #10019, opened 2020-02-10
temporary fix for slow chats with long ens names

flexsurfer:
temporary fix for slow chats with long ens names

### DIFF
--- a/src/status_im/ui/screens/chat/utils.cljs
+++ b/src/status_im/ui/screens/chat/utils.cljs
@@ -13,14 +13,13 @@
 
 (defn format-author [alias style name]
   (let [additional-styles (style false)]
-    (if name
-      (let [name (subs name 0 40)]
-        [react/text {:number-of-lines 1
-                     :style (merge {:color       colors/blue
-                                    :font-size   13
-                                    :line-height 18
-                                    :font-weight "500"} additional-styles)}
-         (str "@" (or (stateofus/username name) name))])
+    (if (< (count name) 40)
+      [react/text {:number-of-lines 1
+                   :style (merge {:color       colors/blue
+                                  :font-size   13
+                                  :line-height 18
+                                  :font-weight "500"} additional-styles)}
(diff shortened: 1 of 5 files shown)


yenda (changes requested):
I think the fix should also prevent registration of usernames more than 40 char long, because otherwise this opens the door to impersonation by registering a longer username. I understand it is temporary, the actual fix should be in status-go and in the contract.

src/status_im/ui/screens/chat/utils.cljs
```diff
@@ -36,7 +37,7 @@
     (or (and (= from current-public-key)
              [react/text {:style (style true)}
               (str reply-symbol (i18n/label :t/You))])
-        (format-author reply-name style false))))
+        (format-author (subs reply-name 0 40) style false))))
```
yenda: this is going to shorten some long random names?

flexsurfer:
this PR doesn't solve anything

flexsurfer:
or it does

flexsurfer:
probably we just need to remove number-of-lines and show full name if  it less than 40

src/status_im/ui/screens/chat/utils.cljs
```diff
@@ -13,9 +13,9 @@
 
 (defn format-author [alias style name]
   (let [additional-styles (style false)]
-    (if name
-      [react/text {:number-of-lines 1
-                   :style (merge {:color       colors/blue
+    (if (and name
+             (< (count name) 40))
```
flexsurfer, line 17: @yenda  here is a full name with domain, so it can be like `myname.domain1.domain2.eth` also we allow to register 40 symbols for status name, but here we don't count stateofus.eth domain length, so if user will register ens name contains 39 symbols we wont show it here

flexsurfer: so hiding names not a solution for sure, because we should show such names for sure `myname.domain1.domain2.eth`

PR #10022, opened 2020-02-11
temporary fix for app crash if chat has long ens names

flexsurfer:
temporary fix for app crash if chat has long ens names

### DIFF
--- a/src/status_im/ui/screens/chat/utils.cljs
+++ b/src/status_im/ui/screens/chat/utils.cljs
@@ -14,12 +14,13 @@
 (defn format-author [alias style name]
   (let [additional-styles (style false)]
     (if name
-      [react/text {:number-of-lines 1
-                   :style (merge {:color       colors/blue
-                                  :font-size   13
-                                  :line-height 18
-                                  :font-weight "500"} additional-styles)}
-       (str "@" (or (stateofus/username name) name))]
+      (let [name (subs name 0 80)]
+        [react/text {:number-of-lines 2
+                     :style (merge {:color       colors/blue
+                                    :font-size   13
+                                    :line-height 18
+                                    :font-weight "500"} additional-styles)}
+         (str "@" (or (stateofus/username name) name))])
       [react/text {:style (merge {:color       colors/gray
                                   :font-size   12
                                   :line-height 18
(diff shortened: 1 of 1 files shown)


cammellos:
What works slow here? have you checked the performance of `stateofus/username`? 

```
(defn username [name]
  (when (and name (string/ends-with? name domain))
    (first (string/split name "."))))
```
If not I would first remove that, and try the ellipsis (as I understand that's preferrable if performance are ok) https://github.com/status-im/status-react/pull/10021 . If performance are acceptable then the issue is in `username` function, which looks fairly inefficient.

Split is not lazy, so it will go through the whole string regardless, if we really wants to keep exactly the same behavior we can use either `loop/recur` or `reduce/reduced`, which both short circuit at the first `.`. Otherwise if we are not concerned about names such as `foo.bar.stateofus.eth` we an just take a substring `(subs name 0 (- (count name) state-of-us-length))` which should also be more performant.
`ends-with?` is likely performant enough, but worth checking.

flexsurfer:
@cammellos because it works slow only on android, i believe problem in text component on Android, probably it ellipsize every component and its slow, and still its just a simple and quick fix for v1, we must prepare proper fix in v1.1.

cammellos (approved):

Serhy:
App doesn't crash for me on Samsung J7 (Android 8.1) and Galaxy Note 4 (Android 6.0.1) in status chat with loong ENS (fetched messages and scrolled up to 7th of February morning time history).
iOS is OK too (but I had no problems with release build there as well)

flexsurfer:
@Serhy try to join #status-chinese and fetch history for 3-4days

flexsurfer:
increased length up to 80

Serhy:
@flexsurfer with 40 char-long-ens build ```#status-chunese``` works even better for me for the last 5 days than ```#status``` (messages loads quicker, less chat-message-dancing behavior)

flexsurfer:
@Serhy and what about #status-chunese in latest nightly or RC ?

flexsurfer:
chat still pretty fast with 80 so lets keep 80

Serhy:
In RC, can't fetch and scroll to the top of 4 days history in ```#status-chinese```.  I gave up after 5 appearing `Status isn't responding` pop-ups -> tap `Wait`

PR #10025, opened 2020-02-11
Predefined multiaccounts for development

rasom:
This commit allows to add predefined multiaccs to `.dev-multiaccounts.json` to avoid manual entering password/seed during account creation/restoring/unlocking.

```json
{
"multiaccounts":
{
"account-name1": "seed-word#1 seed-word#2 ...",
"account-name2": "seed-word#1 seed-word#2 ..."
},
"password": "123123"
}
```

status: ready

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -185,3 +185,4 @@ status-modules/resources
 /fiddle/resources/public/images/
 
 trace.edn
+.dev-multiaccounts.json
--- a/src/status_im/multiaccounts/dev.cljs
+++ b/src/status_im/multiaccounts/dev.cljs
@@ -1,0 +1,28 @@
+(ns status-im.multiaccounts.dev
+  (:require [status-im.utils.types :as types])
+  (:require-macros [status-im.utils.slurp :as slurp]))
+
+;; Predefined accounts can be added to .dev-multiaccounts.json to simplify
+;; multiacc creation, restoring, and unlocking during development.
+;; Example JSON:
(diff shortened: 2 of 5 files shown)


cammellos:
@rasom mind that 1-to-1 encryption is device-to-device, creating multiple accounts restoring the same seed is bound to cause problems with receiving 1-to-1 messages (contact updates/group chat messages) etc

rasom:
@cammellos I'm not sure that it will be useful for all cases, but sometimes you need to restore account frequently or login/logout. For me it was the case when I was working on tx history. I literally entered the same seed thousands of times. :) 

And well, if you don't have that accounts file locally you won't even see this option.

cammellos:
@rasom I see, I didn't see that they were in `gitignore`, that's actually pretty handy!

yenda:
@rasom on android I've been using `adb shell input text 'seed%sphrase'` for a while, it types the seed phrase for you

rasom:
@yenda that's a nice option, but I would prefer to not switch to keyboard specifically when I'm testing app on a real device. Just pressing some button on screen is faster anyway.

yenda:
@rasom yes sure, I'm not typing anything just copy pasting the command though, and it types itself on the real device

rasom:
@yenda I understand that you don't have to type it again each time, but you have to switch to keyboard from the device (to run that command). Also it wouldn't work on iOS.

PR #10027, opened 2020-02-11
rename get.status.im to join.status.im

jakubgs:
I've renamed the https://get.status.im/ site to https://join.status.im/ in https://github.com/status-im/infra-misc/commit/d4784edd.

The old links should work due to a CloudFlare HTTP redirect I deployed, but I think a URL like:
https://join.status.im/user/jakubgs
Reads much better as a sentence, which literally reads as "Join Status.im user Jakubgs".


yenda (approved):
would need marketing approval but I think it's a good idea

errorists (approved):

Serhy:
test_open_public_chat_using_deep_link test fails because https://join.status.im/  (and https://get.status.im/ before) does not determine whether Status is already installed on device, thus showing 'Download Status'
If app is opened then I navigate to the deeplink appropriate content fine. 

We'll need to update https://join.status.im/  page itself

jakubgs:
But there's no such thing as determining if app is installed from the browser. That would be a breach of privacy. It's impossible to detect from a browser if an app is installed.

There is something different available on Android and I opened an issue about that: #10029

PR #10043, opened 2020-02-14
[#10041] Request token price after adding

### LINKED ISSUE #10041: Fiat value is '0' in added custom tokens
# Bug Report

## Problem

When adding a custom token with total amount in fiat > 0, we don't calculate fiat value (only token amount is displayed)

#### Expected behavior

Fiat value calculated in added token with valid > 0 amount (until app restart)

#### Actual behavior

Fiat value is '0' in added token with amount until app restart
Screenshot 2020-02-14 at 11 37 39

### Reproduction
- Open Status and recover an account with custom tokens available (or new account - and you may add watch-only address then)
- In `Wallet` -> `...` -> `Manage assets` -> `Add custom token` paste address of custom token (e.g. 0x016ee7373248a80BDe1fD6bAA001311d233b3CFa  which is `3X Short Bitcoin`)
- Add watch-only address having that token (e.g. `0x3f5ce5fbfe3e9af3971dd833d26ba9b5c936f0be`) 

### Additional Information

- Status version: `1.0.1`

rasom:
fix #10041 

status: ready

### DIFF
--- a/src/status_im/utils/prices.cljs
+++ b/src/status_im/utils/prices.cljs
@@ -1,7 +1,8 @@
 (ns status-im.utils.prices
-  (:require [status-im.utils.http :as http]
+  (:require [clojure.string :as string]
+            [status-im.utils.http :as http]
             [status-im.utils.types :as types]
-            [status-im.utils.config :as config]))
+            [taoensso.timbre :as log]))
 
 ;; Responsible for interacting with Cryptocompare API to get current prices for
 ;; currencies and tokens.
@@ -15,7 +16,7 @@
 (def status-identifier "extraParams=Status.im")
 
 (defn- ->url-param-syms [syms]
-  ((comp (partial clojure.string/join ",") (partial map name)) syms))
+  ((comp (partial string/join ",") (partial map name)) syms))
 
 (defn- gen-price-url [fsyms tsyms chaos-mode?]
   (if chaos-mode?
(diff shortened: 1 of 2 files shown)


src/status_im/utils/prices.cljs
```diff
@@ -40,6 +41,11 @@
                                      :last-day (:OPEN24HOUR entry)}}))}))))
 
 (defn get-prices [from to mainnet? on-success on-error chaos-mode?]
+  (log/debug "[prices] get-prices"
+             "from" from
+             "to" to
+             "mainnet?" mainnet?
+             "chaos-mode?" chaos-mode?)
```
yenda, line 48: didn't we remove chaos-mode?

yenda (approved):

src/status_im/utils/prices.cljs
rasom: I have no idea, it is at least in this function

rasom: @yenda as far as I see it is still in the project

flexsurfer:
@rasom related bug, when we remove token we don't recalculate total value in fiat, probably could be done in this PR ?

src/status_im/utils/prices.cljs
flexsurfer: i just disabled it, not removed, but dunno if we want to remove all related code completely

flexsurfer (approved):

PR #10068, opened 2020-02-24
[#10012] Show remaining PIN attemts only after entering wrong PIN

### LINKED ISSUE #10012: Number of remaining keycard PIN attempts is wrong - because it can't be stored in the phone
The number of passcode retries possible is a number held by the card. The card can for instance be used by several phones. 

**Current behaviour**
we show the number of attempt left at different places. Example: when signing transaction, when logging in (see screens below) before card is tapped. 
The problem is that we are not sure what this number is before we tap the card to get this information from the card. 
Moreover we have optmized all flows to have only one tap.  Example when logging in we ask to the user his PIN, and then ask him to tap, we are not asking the user for a first tap to without PIN to know the number of remaining attempts. 

image

image

**Solution**
The number of remaining attempts for the keycard should ONLY be shown when the user has made a wrong pin code. 

As fgor the design it's been specified by @andmironov here https://github.com/status-im/status-react/issues/9347

rasom:
fix #10012 

status: ready

### DIFF
--- a/src/status_im/ui/screens/hardwallet/pin/views.cljs
+++ b/src/status_im/ui/screens/hardwallet/pin/views.cljs
@@ -10,6 +10,8 @@
             [status-im.utils.platform :as platform]
             [status-im.ui.components.topbar :as topbar]))
 
+(def ^:const default-pin-retries-number 3)
+
 (defn numpad-button [n step enabled? small-screen?]
   [react/touchable-highlight
    {:on-press #(when enabled?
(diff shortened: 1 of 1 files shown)


vitvly (approved):

src/status_im/ui/screens/hardwallet/pin/views.cljs
```diff
@@ -10,6 +10,8 @@
             [status-im.utils.platform :as platform]
             [status-im.ui.components.topbar :as topbar]))
 
+(def ^:const default-pin-retries-number 3)
```
flexsurfer, line 13: why not in constants ns ?

rasom: until it is needed anywhere else I see no point in making it "global"

PR #10078, opened 2020-02-25
use Android keystore from Jenkins credentials

jakubgs:
To improve security of Keystore I'm moving it to Jenkins credentials store.
This way we won't have to store the Keystore on individual CI hosts, only the `master-01`.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -244,7 +244,6 @@ android {
         release {
             // Caution! In production, you need to generate your own keystore file.
             // see https://facebook.github.io/react-native/docs/signed-apk-android.
-            signingConfig signingConfigs.debug
             minifyEnabled enableProguardInReleaseBuilds
             proguardFiles getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro"
             signingConfig signingConfigs.release
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -29,8 +29,6 @@ pipeline {
     BUILD_ENV = 'prod'
     NIX_CONF_DIR = "${env.WORKSPACE}/nix"
     FASTLANE_DISABLE_COLORS = 1
-    /* since we are mounting it we need to specify location */
-    STATUS_RELEASE_STORE_FILE = '/home/jenkins/status-im.keystore'
     /* We use EXECUTOR_NUMBER to avoid multiple instances clashing */
     LEIN_HOME = "/var/tmp/lein-${EXECUTOR_NUMBER}"
     /* coverage report identification */
--- a/ci/android.groovy
+++ b/ci/android.groovy
@@ -25,6 +25,10 @@ def bundle() {
 
   /* credentials necessary to open the keystore and sign the APK */
   withCredentials([
+    file(
+      credentialsId: 'status-im.keystore',
+      variable: 'STATUS_RELEASE_STORE_FILE'
+    ),
     string(
       credentialsId: 'android-keystore-pass',
       variable: 'STATUS_RELEASE_STORE_PASSWORD'


jakubgs:
Oh lovely, looks like it doesn't work:
```
java.lang.RuntimeException: java.lang.RuntimeException:
  com.android.ide.common.signing.KeytoolException:
  Failed to read key **** from store "****":
    Keystore was tampered with, or password was incorrect
```
Might be related to how Jenkins changes the permissions of the file.

rasom (approved):

jakubgs:
I tried doing `chmod 640` on the keystore but that didn't work:
```
+ chmod 640 ****
...
java.lang.RuntimeException: java.lang.RuntimeException: com.android.ide.common.signing.KeytoolException: Failed to read key **** from store "****": Keystore was tampered with, or password was incorrect
```

jakubgs:
Whelp. I ran `md5sum` on the Keystore file produced by Jenkins from credentials and got:
```
[Pipeline] sh (hide)
16:51:12  + chmod 640 ****
[Pipeline] sh
16:51:12  + ls -l ****
16:51:12  -rw-r----- 1 jenkins jenkins 2159 Feb 25 15:51 ****
[Pipeline] sh
16:51:12  + md5sum ****
16:51:12  4000bd8b1f8b86f16e8688773e8ee88b  ****
```
But that's not a correct checksum. That means Jenkins modifies the contents when creating the file.

jakubgs:
I crated a `test-file.txt` credential of same type:
https://ci.status.im/credentials/store/system/domain/_/credential/test-file.txt/
Which simply contains text `TEST` and hash an MD5 sum of:
```
2debfdcf79f03e4a65a667d21ef9de14
```
And ran this test CI job to verify this:
```groovy
pipeline {
    agent { label "linux" }
    
    stages {
        stage('Test') {
            steps {
                withCredentials([
                    file(
                        credentialsId: 'test-file.txt',
                        variable: 'TEST_FILE'
                    )]
                ) {
                    // contains text "TEST"
                    sh "ls -l ${env.TEST_FILE}"
                    sh "md5sum ${env.TEST_FILE}"
                    // expected sum
                    sh "echo TEST | md5sum"
                }
            }
        }
    }
}
```
And the result was correct:
```
+ ls -l ****
-r-------- 1 jenkins jenkins 5 Feb 25 16:10 ****
[Pipeline] sh
+ md5sum ****
2debfdcf79f03e4a65a667d21ef9de14  ****
[Pipeline] sh
+ echo TEST
+ md5sum
2debfdcf79f03e4a65a667d21ef9de14  -
```
https://ci.status.im/job/tests/job/secret-file-test/1/

So there must have been something wrong with how I uploaded the Keystore file. I guess.

jakubgs:
Okay, `scp`ing the file again and re-uploading to Jenkins fixed the MD5 sum, but I'm still getting:
```
Failed to read key **** from store "****": Keystore was tampered with, or password was incorrect
```
So it must be something more than that.

jakubgs:
It works! Somehow I managed to upload the wrong keystore twice. 3rd time's the charm.

andremedeiros (approved):

corpetty:
Does this put extra security scrutiny on `master-01`? Is this a point of failure in case something goes wrong.  What could go wrong?

jakubgs:
>Does this put extra security scrutiny on `master-01`?

Not sure what you mean by that. Before the file was directly on the filesystem of 5 Linux CI hosts. Now it's encrypted in the `credentials.xml` file on `master-01` which means it requires both access to the filesystem and admin access to Jenkins UI to extract it. And that's in ADDITION to the two passwords you need to unlock the keystore itself and then the specific key in the keystore.

What more do you want?

jakubgs:
>Is this a point of failure in case something goes wrong.

You mean as in we lose the host? Or what?
I have an encrypted version of the Keystore. If you want we can add it to `infra-pass`.

>What could go wrong?

That's a very broad question. Not sure what to tell you. I explained how the credentials store works. You can check how extracting a value from it works here:
https://stackoverflow.com/a/37683492

jakubgs:
Merging this as a partial fix. We can improve upon it in the future.

PR #10090, opened 2020-02-27
Use community WebView

### LINKED ISSUE #4: Migration to community WebView
# Migration to community WebView

## Introduction

Currently we rely on react-native-webview-bridge library. It is an extension of RN's WebView that injects a special script into each webpage it loads. Its raison d'etre is allowing bidirectional communication between RN and WebView.

## Current status

We maintain our own [fork](https://github.com/status-im/react-native-webview-bridge) of react-native-webview-bridge, adding fixes for web3.js support for dapps, webview caching, permissions fixes, etc. Comprehensive list is in a [comparison sheet](https://docs.google.com/spreadsheets/d/1bskrf5Vrc8cQ65j5pn51JApwRU9TQTgK_9skIZrWgho/edit#gid=0).

`webview-bridge`library  uses `sendToBridge()` method for RN->WebView communication and `send()` method for WebView->RN communication.

In `status-react` codebase, we use them as
  - `:browser/send-to-bridge` fx (`status-im.browser.core` ns). Also used by `:browser.callback/call-rpc` fx) 
  - :on-bridge-message (`status-im.browser.views` ns). Dispatches a `:browser/bridge-message-received` event.

Correspondingly, on WebView side there is a `WebViewBridge` global variable that has a `send()` method and a `onMessage()` method.

## The problem
There are some issues with it, e.g.:
  - upstream version hasn't been maintained in a while, and has some issues, e.g. dependency on UIWebView on iOS that is deprecated (https://github.com/status-im/status-react/issues/8897)

  - RN's WebView that WebView-Bridge is based on is itself deprecated in favor of https://github.com/react-native-community/react-native-webview.

## Why Don't We Just...?
We could go on with maintaining our webview-bridge fork, but then we'd have to roll out our own WKWebView (UIWebView replacement) support, and would skip on community updates to WebView.

## Proposal

We should migrate to community WebView. There would be some slight API changes, e.g.:

| Direction | `webview-bridge` | `community webview` |
| --------- | ------------------ | ---------------------- |
|RN -> WebView| `WebViewBridge.sendToBridge()`| `injectJavaScript()` method or `injectedJavaScript` prop|
|RN -> WebView|`onMessage`|`onMessage`|
|WebView -> RN|`WebViewBridge.send()`|`ReactNativeWebView.postMessage()`|
|WebView -> RN|`onBridgeMessage`|`onMessage`|

However, currently there are unsolved issues with it that prevent using community WebView as an (almost) drop-in replacement. We will maintain our own [fork](https://github.com/status-im/react-native-webview) until these are merged upstream.

The most important is allowing for JS injection before the document load. It is required by `web3.js`, and previously Status team implemented their own fixes for [Android](https://github.com/status-im/react-native-webview-bridge/commit/381467999a6737617eef6d0b6a26b1eed13edc41#diff-0182b1b9ed632c87dec03071732bcbddR186) and [iOS](https://github.com/status-im/react-native-webview-bridge/commit/85c4266b9109dfb150525535b2bd36a1f2969a34).

A proposed plan is:

### Stage 1
JS injection implementation. We might borrow ideas from https://github.com/react-native-community/react-native-webview/pull/229 so that:
  - on Android, override `RNCWebViewClient.shouldInterceptRequest()` and `RNCWebViewClient.interceptRequest()`. Use `OkHttpClient` during request interception to re-send a request.
Another option is overriding `ReactWebViewClient.onPageStarted()` and invoking `evaluateJavascript()` as implemented in https://github.com/status-im/react-native-webview-bridge/commit/381467999a6737617eef6d0b6a26b1eed13edc41#diff-0182b1b9ed632c87dec03071732bcbddR186.
  - on iOS, use `WKUserScript` with `injectionTime:WKUserScriptInjectionTimeAtDocumentStart` param, and then call `WKUserContentController.addUserScript`.

### Stage 2
Caching, cookies persistence.

#### Alternate route
Or, alternatively, we can extend the API that Status offers to Dapps. There is currently a `StatusAPI` object and a `sendAPIRequest()` function in `provider.js`. They provide endpoints for contact code fetching and QR code scanning.

There are some security issues associated with exposing more of Status internal API. These should be solved with a number of approaches, namely (credit goes to @corpetty for elaborating on this):
  - sanitization: making sure calls are properly formed according to a specification
  - permissions: dapps requesting access to chat key, profile attributes, some account, etc. should request specific permissions
  - sandboxing: dapps (and websites) shouldn't be able to talk outside of their context and given permissions

### Stage 3
Permissions fixes + other.

### Stage 4
To be elaborated on. We can expose Status chat functionality as an API to embedded WebViews. A `postMessage`/`onMessage` mechanism could be used to provide a nice, more abstract API to dapps so that they can use what Status already has as part of status-go. This will help to avoid duplicate functionality in Status itself and nested WebViews.

vitvly:
Fixes https://github.com/status-im/team-core/issues/4


jakubgs:
As far as I can tell `yarn2nix` has no responsibility to run `yarn prepare` based on:

>prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.
https://docs.npmjs.com/misc/scripts

Which means that the [`prepare` target](https://github.com/react-native-community/react-native-webview/blob/86c44588f88d87effea87ae996dacad7d8f9504d/package.json#L22) defined in `react-native-webview` is supposed to be run before publishing the package, not during installation.

The simplest solution I can see is to create a commit that includes the result of `yarn run prepare` and tag it for use in our dependencies. Using branches is a horrible idea.

vitvly:
It does work on a test RN project though, which is weird. Docs say it should run on local npm install, and it seems to work in that case.

android/app/src/main/java/im/status/ethereum/MainApplication.java
```diff
@@ -52,5 +52,6 @@ public ReactNativeHost getReactNativeHost() {
     public void onCreate() {
         super.onCreate();
         SoLoader.init(this, /* native exopackage */ false);
+        WebView.setWebContentsDebuggingEnabled(true);
```
flexsurfer: why `BuildConfig.DEBUG_WEBVIEW == "1")` isn't used? don't think we want this enabled in release

android/app/src/main/java/im/status/ethereum/MainActivity.java
```diff
@@ -231,7 +231,7 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
         }
         if (grantResults.length == 1 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
             // Permission has been granted. Start camera preview Activity.
-            com.github.alinz.reactnativewebviewbridge.WebViewBridgeManager.grantAccess(requestCode);
+            //com.github.alinz.reactnativewebviewbridge.WebViewBridgeManager.grantAccess(requestCode);
```
flexsurfer: could you add a comment in code, why it was commented ?

android/gradle.properties
```diff
@@ -21,6 +21,7 @@
 # $keytool -genkey -v -keystore ./status-keystore.jks -keyalg RSA -keysize 2048 -validity 10000 -alias status
 
 # Version requirements used throughout the Gradle scripts
+kotlinVersion=1.3.11
```
flexsurfer, line 24: why is this needed ?

mobile/js_files/DEPENDENCIES.md
```diff
@@ -152,7 +152,7 @@ TODO: updating requires a pod update on iOS
 
 used for touch-id identification
 
-## "react-native-webview-bridge": "git+https://github.com/status-im/react-native-webview-bridge.git#fix/classnames-colision"
+## "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#js_injection_before_load"
```
flexsurfer: why #js_injection_before_load ?

src/status_im/browser/core.cljs
```diff
@@ -420,7 +420,10 @@
  :browser/send-to-bridge
  (fn [{:keys [message webview]}]
    (when (and message webview)
-     (.sendToBridge webview (types/clj->json message)))))
+     (let [msg (str "ReactNativeWebView.onMessage('"
```
flexsurfer: could you elaborate ?  looks weird

flexsurfer (approved):

android/app/src/main/java/im/status/ethereum/MainApplication.java
vitvly: Oh, sorry, did not know about the flag. Will fix.

android/app/src/main/java/im/status/ethereum/MainActivity.java
vitvly: I will simply remove the line, it's not used any more.

android/gradle.properties
vitvly: Without it, android builds were failing (specifically `make run-android` command for me). I'll uncomment and will check if it's still reproducible.

mobile/js_files/DEPENDENCIES.md
vitvly: Wrong, forgot to replace with a tag instead.

src/status_im/browser/core.cljs
vitvly: It's a community-webview way of posting messages in RN->WebView direction. I used a modified version of what was suggested [here](https://github.com/react-native-community/react-native-webview/issues/809#issuecomment-527245591).
We invoke `onMessage` fn that is defined in `provider.js`

android/gradle.properties
vitvly: Yeah, it's still reproducible:
```
A problem occurred configuring project ':react-native-webview'.
> Could not resolve all artifacts for configuration ':react-native-webview:classpath'.
   > Could not find org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50.
     Searched in the following locations:
       - file:/nix/store/2vcnkyzsghmb5l139ndd0kcz2jsjdc5r-status-react-patched-npm-gradle-modules/.m2/repository/org/jetbrains/kotlin/kotlin-gradle-plugin/1.3.50/kotlin-gradle-plugin-1.3.50.pom
       - file:/nix/store/2vcnkyzsghmb5l139ndd0kcz2jsjdc5r-status-react-patched-npm-gradle-modules/.m2/repository/org/jetbrains/kotlin/kotlin-gradle-plugin/1.3.50/kotlin-gradle-plugin-1.3.50.jar
     Required by:
         project :react-native-webview
```

churik:
@siphiuel please rebase to current develop

vitvly:
@churik rebased

Serhy:
@siphiuel I've created separately issue https://github.com/status-im/status-react/issues/10189 because it happens in develop as well.. But with this PR it reveals the crash (on Android9 real device) very well. Kind of happens from the first attempt (unlike to Develop where I need to close-open > 10 time webview so it crashed)

**Issue 1**
We are using `simpledapp.eth` as a test Dapp. But with PR build it does't work on Androids (all fine with iOS thought). 
After allowing permissions to retrieve wallet address and enable web3 it is not picking it up. 
It's all fine with the cryptokitties Dapp, but because all works with Develop nightly build there is some issue with WebView changes.

To reproduce:
2) Navigate to Browser and enter `simpledapp.eth` in URL input and hit GO
3) Allow permissions to wallet and web3 (tap Allow button) and allow permissions to address (hit Allow again) 
4) Check on Account tab there is an address value in defaultAccount / coinbase/ accounts. Try to send any transaction from this Dapp
Actual result: nothing happens when tapping any of transaction/message related buttons. Empty addresses.
Screenshot 2020-03-19 at 15 54 16

These are logs for when simpledapp.eth loads with allowing permissions [PRWebView.txt](https://github.com/status-im/status-react/files/4354917/PRWebView.txt)

Maybe @flexsurfer may give a hint what's wrong with update made here WebView and why it doesn't work with simpledapp.eth. Because I was able to exchange tokens in Airswap and purchase cryptokitty (I mean some other Dapps working)

vitvly:
@Serhy thanks so much for the tip that app crashes https://github.com/status-im/status-react/issues/10189 could have started with https://github.com/status-im/status-react/pull/9945. Actually it seems that https://github.com/status-im/status-react/pull/10176 is the cause.

Just pushed the commit that seems to fix crashes on Android.

It is similar to an older issue we had: https://github.com/status-im/status-react/pull/8868
In general, `react-native-screens` repo has a lot of issues with crashes, like https://github.com/software-mansion/react-native-screens/issues/214, https://github.com/software-mansion/react-native-screens/issues/309. 

If this is confirmed to work after QA, then i suggest we just revert https://github.com/status-im/status-react/pull/10176.

vitvly:
@Serhy simpledapp.eth fixed

Serhy:
No crashes now when closing webview! 
simpledapp.eth also works fine! 
Can loging to google account and cookies are kept in tab.  
But Dapp store now (`https://dap.ps`) doesn't  show any Dapp in the list of daps (View all or any other particular tab). All the lists are empty.   

Screenshot 2020-03-23 at 15 09 28

vitvly:
@Serhy thanks, checking for what's up with dap.ps

vitvly:
@Serhy fixed dap.ps issue

Serhy:
Looks good to me now. Dap.ps and simpledapp.eth are workoing now. Can send transaction in DApps. 
Cookies are supported now fine. 
@siphiuel could you rebase develop please (and fix conflicts) and I'll restart e2e tests..

vitvly:
@Serhy thanks, rebased

Serhy:
That on test failed due to known bug which is in develop https://github.com/status-im/status-react/issues/10203

PR #10100, opened 2020-02-28
[#9749] Support importing private key and seed

### LINKED ISSUE #9749: Support importing private key
By seed (MetaMask) and privatekey.

flexsurfer:
fixes #9749

QA: keycard is not supported https://github.com/status-im/status-react/issues/10101

### DIFF
--- a/src/status_im/wallet/accounts/core.cljs
+++ b/src/status_im/wallet/accounts/core.cljs
@@ -13,78 +13,228 @@
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.fx :as fx]
             [status-im.utils.types :as types]
-            [status-im.wallet.core :as wallet]))
+            [status-im.wallet.core :as wallet]
+            [clojure.string :as string]
+            [status-im.utils.security :as security]
+            [status-im.multiaccounts.recover.core :as recover]
+            [status-im.ethereum.mnemonic :as mnemonic]))
+
+(fx/defn start-adding-new-account
+  {:events [:wallet.accounts/start-adding-new-account]}
+  [{:keys [db] :as cofx} {:keys [type] :as add-account}]
+  (let [{:keys [latest-derived-path]} (:multiaccount db)
+        path-num (inc latest-derived-path)
+        account  (merge
+                  {:color (rand-nth colors/account-colors)}
+                  (when (= type :generate)
+                    {:name (str "Account " path-num)}))]
+    (fx/merge cofx
+              {:db (assoc db :add-account (assoc add-account :account account))}
+              (navigation/navigate-to-cofx :add-new-account nil))))
+
+(fx/defn new-account-error
+  {:events [::new-account-error]}
+  [{:keys [db]} error-key error]
+  {:db (update db :add-account merge {error-key error
+                                      :step nil})})
+
+(defn account-stored [path type]
+  (fn [result]
+    (let [{:keys [error publicKey address]} (types/json->clj result)]
+      (if error
+        (re-frame/dispatch [::new-account-error :account-error error])
+        (re-frame/dispatch [:wallet.accounts/account-stored
+                            {:address    address
+                             :public-key publicKey
+                             :type       type
+                             :path       path}])))))
+
+(def dec-pass-error "could not decrypt key with given password")
+
+(defn normalize-path [path]
+  (if (string/starts-with? path "m/")
+    (str constants/path-wallet-root
+         "/" (last (string/split path "/")))
+    path))
+
+(defn derive-and-store-account [path hashed-password type]
+  (fn [value]
+    (let [{:keys [id error]} (types/json->clj value)]
+      (if error
+        (re-frame/dispatch [::new-account-error :password-error error])
+        (status/multiaccount-derive-addresses
+         id
+         [path]
+         (fn [_]
+           (status/multiaccount-store-derived
+            id
+            [path]
+            hashed-password
+            (fn [result]
+              (let [{:keys [error] :as result} (types/json->clj result)
+                    {:keys [publicKey address]} (get result (keyword path))]
+                (if error
+                  (re-frame/dispatch [::new-account-error :account-error error])
+                  (re-frame/dispatch
+                   [:wallet.accounts/account-stored
+                    {:address    address
+                     :public-key publicKey
+                     :type       type
+                     :path       (normalize-path path)}])))))))))))
+
+(def pass-error "cannot retrieve a valid key for a given account: could not decrypt key with given password")
+
+(defn store-account [path hashed-password type]
+  (fn [value]
+    (let [{:keys [id error]} (types/json->clj value)]
+      (if error
+        (re-frame/dispatch [::new-account-error
+                            (if (= error pass-error) :password-error :account-error)
+                            error])
+        (status/multiaccount-store-account
+         id
+         hashed-password
+         (account-stored path type))))))
 
 (re-frame/reg-fx
- :list.selection/open-share
- (fn [obj]
-   (list-selection/open-share obj)))
+ ::verify-password
+ (fn [{:keys [address hashed-password]}]
+   (status/verify
+    address hashed-password
+    #(re-frame/dispatch [:wallet.accounts/add-new-account-password-verifyied % hashed-password]))))
 
 (re-frame/reg-fx
  ::generate-account
- (fn [{:keys [derivation-info hashed-password path-num]}]
+ (fn [{:keys [derivation-info hashed-password]}]
    (let [{:keys [address path]} derivation-info]
      (status/multiaccount-load-account
       address
       hashed-password
-      (fn [value]
-        (let [{:keys [id error]} (types/json->clj value)]
-          (if error
-            (re-frame/dispatch [::generate-new-account-error])
-            (status/multiaccount-derive-addresses
-             id
-             [path]
-             (fn [result]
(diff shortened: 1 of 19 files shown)


src/status_im/multiaccounts/recover/core.cljs
```diff
@@ -24,7 +24,7 @@
   (contains? multiaccounts key-uid))
 
 (re-frame/reg-fx
- ::validate-mnemonic
+ :validate-mnemonic
```
yenda: why remove the namespace?

src/status_im/wallet/accounts/core.cljs
```diff
@@ -13,78 +13,191 @@
...
+                                             :address wallet-root-address}
+                                            ;; Fallback on the master account for keycards, use the absolute path
+                                            {:path    (str constants/path-wallet-root "/" path-num)
+                                             :address (get-in db [:multiaccount :address])})
+                         :hashed-password hashed-password}}))
+
+(fx/defn import-new-account-seed
+  [{:keys [db]} passphrase hashed-password]
+  {:db                   (assoc-in db [:add-account :step] :generating)
+   :validate-mnemonic   [(security/safe-unmask-data passphrase)
```
yenda: keep the namespaced keywords and import via alias, if there is a circular dep it just means that the effect is not in the right place from a code organization pov

churik:
#### ISSUE 1: field seed phrase is case-sensitive and contains autosuggestion

**Steps:**
- open status
- go to wallet > Add account > Enter seed phrase (try with capital letters)

**Expected result:**
Seed phrase input have the same validation rules as in flow for recovery account

**Actual result:**
Monosnap 2020-02-28 13-12-46

Also, keyboard has turned on autosuggestion when focusing in field (referred to the issue https://github.com/status-im/status-react/issues/9034)

churik:
#### ISSUE 2: Can add account with invalid password
**Steps:**
- open status
- go to wallet > Add account > Enter seed phrase
- type invalid password
- fill other fields

**Expected result:**
Validation error

**Actual result:**
photo_2020-02-28 13 39 27
but account is added

src/status_im/wallet/accounts/core.cljs
flexsurfer: fixed

src/status_im/multiaccounts/recover/core.cljs
flexsurfer: fixed

flexsurfer:
issue1 : fixed

flexsurfer:
@churik  can't reproduce issue 2, btw for importing seed and private key there is no wrong password, because you can use any password you want, yeah i know its confusing

flexsurfer:
summon @hesterbruikman , so when you import accounts with seed or private key, you can use any new password to import them , that might be confusing for sure

hesterbruikman:
> summon @hesterbruikman , so when you import accounts with seed or private key, you can use any new password to import them , that might be confusing for sure

If I recall correctly this should be your existing password. Also as it says, `Enter your password`. cc @guylouis to verify

If it is indeed the existing password, validation is missing

flexsurfer:
@hesterbruikman we just store new account on the device we can use any password, it's not related anyhow to multiaccount, so its up to us how we want to implement it, if we want new password we should ask it twice , if multiaccount, we need to verify it somehow @gravityblast ?

hesterbruikman:
I'm pretty sure the idea was to simply ask for the same password as to not let people juggle different passwords to sign within the same multiaccount in Status. That's why there's no confirm password and we only say `Enter your password`. Exception is the case where keys are encrypted with a password set up elsewhere.

guylouis:
100%

In my mind, from our past discussions, a new account imported from a private key should be protected  the same way an account on the BIP44 tree is, and we should not create a new password to avoid any added complexity. 

If the private key is encrypted in a json file (but I don't think this issue is about this case) of course, that's something else and at the time of importing, this encryption password is being asked.

flexsurfer:
@churik issue 2 fixed, also now we check if password is the same as for multiaccount

churik:
#### ISSUE 3: Keychain suggests to save password

**Steps:**
- open status
- go to wallet > Add account > Enter seed phrase
- fill all fields
- proceed

**Expected result:**
Account is added, no suggestions about storing password

**Actual result:**
photo_2020-02-28 16 47 34

churik:
#### ISSUE 4: Can add 2 accounts with same seed 

**Steps:**
1) open status
2) go to wallet > Add account > Enter seed phrase
3) fill all fields
4) proceed
5) repeat steps 2-5 with same seed phrase

**Expected result:**
can't add same account twice

**Actual result:**
photo_2020-02-28 16 54 52
It will be deleted if you relogin.
Actually you can create new account with same seed phrase via recovery flow as well - not sure what is expected result here.

flexsurfer:
ISSUE 3 - hm is this solved somehow for other password fields ? i remember we had this but not sure how we solved it

churik:
@flexsurfer 
https://github.com/status-im/status-react/issues/8837  - seems it's gone after redesign, but it appeared again here

churik:
#### ISSUE 5: Can't  export private key

**Steps:**
1) open status
2) go to wallet > Add account > Enter private key 
3) fill all fields
4) proceed

**Expected result:**
account is exported

**Actual result:**
photo_2020-02-28 17 26 20
[status_logs_1.zip](https://github.com/status-im/status-react/files/4268001/status_logs_1.zip)

churik:
#### ISSUE 6: Seed phrase field is overlapped on Android

photo_2020-02-28 17 49 43
Device: Xiaomi Mi Note 9 Pro, 1080 x 2280

churik:
#### ISSUE 7: With whitspaces between, after or before seed phrase restores different account

See https://github.com/status-im/status-react/issues/9670

**Steps:**
1) open status
2) go to wallet > Add account > Enter seed phrase
3) on seed phrase put spaces in the end of phrase
4) fill all fields
5) proceed


**Expected result:**
account is restored according to trimmed seed phrase

**Actual result:**
with whitespaces in different places different accounts are recovered

churik:
@flexsurfer here fix for several cases will be required (and ideally 2 new cases).
So please wait for my commit.
Also would be awesome if you can add accessibility-ids to all fields in new screens - it makes e2e easier to support.
Thanks!

src/status_im/wallet/accounts/core.cljs
```diff
@@ -13,78 +13,222 @@
...
+              (let [{:keys [error] :as result} (types/json->clj result)
+                    {:keys [publicKey address]} (get result (keyword path))]
+                (if error
+                  (re-frame/dispatch [::new-account-error :account-error error])
+                  (re-frame/dispatch
+                   [:wallet.accounts/account-stored
+                    {:address    address
+                     :public-key publicKey
+                     :type       type
+                     :path       (if (string/starts-with? path "m/")
```
rasom: can we extract this path derivation part to a separate fn?

src/status_im/wallet/accounts/core.cljs
```diff
@@ -13,78 +13,222 @@
...
+                                   (str constants/path-wallet-root
+                                        "/" (last (string/split path "/")))
+                                   path)}])))))))))))
+
+(def pass-error "cannot retrieve a valid key for a given account: could not decrypt key with given password")
```
rasom, line 85: Is that returned form status-go side? We probably need to start using error codes or something, because comparing messages might not be the best idea and definitely isn't a very reliable way.

src/status_im/wallet/accounts/core.cljs
```diff
@@ -13,78 +13,222 @@
...
 (re-frame/reg-fx
- :list.selection/open-share
- (fn [obj]
-   (list-selection/open-share obj)))
+ ::verify-password
+ (fn [{:keys [address hashed-password]}]
+   (status/verify address hashed-password
+                  #(re-frame/dispatch [:wallet.accounts/add-new-account-password-verifyied
+                                       %
+                                       hashed-password]))))
```
rasom:
```clojure
   (status/verify 
    address hashed-password
    #(re-frame/dispatch
      [:wallet.accounts/add-new-account-password-verifyied % hashed-password]))))
```
not a big deal, but that lonely percent sign looks weird to me

rasom (approved):

flexsurfer:
@churik thanks all issue fixed
ISSUE 3 - can't be fixed

churik:
@corpetty 
is it OK to release this feature with https://github.com/status-im/status-react/pull/10100#issuecomment-592572011 ?
Seems there is no way to fix it shortly.

churik:
Tested:
- fields validation on new screens
- using new types of accounts in DApps + collectibles 
- changing settings on new accounts
- using new accounts in commands
- using ENS from main account
- adding account via seed phrase in wallet and restoring it in user flow
- changing networks
- attempt to add the same account twice (seed phrase + public key)
- search for sensitive info in logs

Apart from https://github.com/status-im/status-react/issues/9895 and https://github.com/status-im/status-react/issues/8837 don't see other issues.
Amazing work!

hesterbruikman:
👋@flexsurfer what's the behavior for Keycard in this current implementation? @rasom mentioned we're missing designs for Add account > Generate keys in case of Keycard. Assuming the same is true here.

flexsurfer:
the behavior is that  Keycard is not supported in this PR , here is the issue  #10101

PR #10106, opened 2020-02-28
[#10084]  Make commands in chat available on Ropsten for e2e builds

### LINKED ISSUE #10084: Make commands in chat available on Ropsten for e2e builds
# Problem

For e2e we need to automate sending ETH and tokens in 1-1 chat.

## Implementation

From @cammellos : we just have to set an env variable (or pass config)

## Acceptance Criteria
All e2e builds (nightly and PRs) have  commands in chat in Ropsten network

flexsurfer:
fixes #10084


cammellos:
@flexsurfer that might not be enough for the confirmation to be checked, you might want to also make `VerifyTransactionURL` in `src/status_im/node/core.cljs`

rasom (approved):

churik:
only STT and ETHro are working in e2e builds, but I believe it is enough @Serhy ?

Serhy:
> only STT and ETHro are working in e2e builds, but I believe it is enough @Serhy ?

Yes. Enough for now. Better to have at least one token with different number of Decimals (like XLS, - 2 decimals max), but for sure it opens for us wide range of additional autotests number having only ETHro and STT :)

jakubgs (approved):

PR #10117, opened 2020-03-02
[#10089] Override the behaviour of public chat' unread messages indic…

### LINKED ISSUE #10089: Override the behaviour of public chat' unread messages indicator
# Problem

Public chats count all new messages for the purpose of incrementing the unread counter. That doesn't perform well in actual use since not all messages are created equal and it's impossible to tell how much of those you should really care about. Often that leaves you with spiralling numbers that you stop caring about, that are hard to keep track of. Instead, public chats should differentiate between any unread message and messages that are directed at you like mentions or direct replies.

## Implementation

Screenshot 2020-02-26 at 16 55 48

For public chats, instead of counting each unread message against the counter, we place a binary 'yes/no' indicator that shows if this chat received any new messages. If any of those messages mention you or are replies to your message, we increase the counter by their total number. 

Opening the chat clears the indicator, at least until we have better way of controlling unread messages, see #6375 

The unread messages indicator is a 12px high and wide blue circle that's placed directly at the center of where an unread counter appears (so if the unread counter is positioned 16px from the right screen edge, and is 22px in size, this means that the smaller indicator is placed 21px from the edge - (22-12) / 2 + 16

The same rules apply for the total counter placed in the Tab Bar, public chat messages which aren't mentions or replies, don't increase the total count. If you only have unread messages in public chats, we similarly place a 12px indicator instead of unread badge. 

Screenshot 2020-02-26 at 17 03 44

It's placed 12px from the bottom and 18px from the left edge of the chat icon and includes a 2px white border to separate it from the icon. When in doubt check[ Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=8499%3A177).

## Acceptance Criteria

Messages in public chats that aren't mentions or direct replies, don't count against the total unread message count, only mention and direct replies. The unread indicator is implemented according to spec.

## Notes

[Designs in Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=8499%3A140)

Thanks!

flexsurfer:
fixes #10089

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -847,13 +847,6 @@
        :messages))))
 
 (re-frame/reg-sub
- :chats/unviewed-messages-count
- (fn [[_ chat-id]]
-   (re-frame/subscribe [:chats/chat chat-id]))
- (fn [{:keys [unviewed-messages-count]}]
-   unviewed-messages-count))
-
-(re-frame/reg-sub
  :chats/photo-path
  :<- [:contacts/contacts]
  :<- [:multiaccount]
@@ -867,7 +860,9 @@
  :chats/unread-messages-number
  :<- [:chats/active-chats]
  (fn [chats _]
-   (apply + (map :unviewed-messages-count (vals chats)))))
+   (let [grouped-chats (group-by :public? (vals chats))]
+     {:public (apply + (map :unviewed-messages-count (get grouped-chats true)))
+      :other (apply + (map :unviewed-messages-count (get grouped-chats false)))})))
 
 (re-frame/reg-sub
  :chats/cooldown-enabled?
--- a/src/status_im/ui/components/tabbar/core.cljs
+++ b/src/status_im/ui/components/tabbar/core.cljs
@@ -6,7 +6,6 @@
    [status-im.ui.components.react :as react]
    [status-im.utils.platform :as platform]
    [status-im.ui.components.icons.vector-icons :as vector-icons]
-   [status-im.ui.components.common.common :as components.common]
    [status-im.ui.components.badge :as badge]
    [status-im.i18n :as i18n]
    [re-frame.core :as re-frame]))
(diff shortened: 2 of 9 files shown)


src/status_im/subs.cljs
```diff
@@ -863,7 +856,9 @@
  :chats/unread-messages-number
  :<- [:chats/active-chats]
  (fn [chats _]
-   (apply + (map :unviewed-messages-count (vals chats)))))
+   (let [grouped-chats (group-by :public? (vals chats))]
+     {:public (apply + (map :unviewed-messages-count (get grouped-chats true)))
```
rasom, line 864: `(reduce + ...`? :)

rasom (approved):

src/status_im/subs.cljs
flexsurfer: what's the difference?

rasom: It just looks more idiomatic to me to use `reduce` here. But I've just found that this is an old debate topic https://stackoverflow.com/questions/3153396/clojure-reduce-vs-apply , not a big deal

rasom: on the second thought 
```clojure
(transduce (map :unviewed-messages-count) (completing +) (get grouped-chats true))
```
could do the same thing in one go, doesn't matter here though

flexsurfer:
@churik @Serhy could you take a look so we can cut v1.1.

flexsurfer:
@errorists could you take a look

errorists:
tested on iOS

📏 perfect score for the list item indicator 10/10! 📐 

however the one on the tab bar has a problem, it has a blue halo and the border is set to inside not outside

Screenshot 2020-03-03 at 09 29 33

I recall we had the same issue with the edit avatar picture button back when we still had it. 
so likely you want to 

`(def counter-public
   {:right            1
    :border-radius   8
    :width            16
    :height           16
`

to account for the larger size. I asked @Ferossgp how to get rid of the halo

also I'm quite sure it's not introduced in this PR but what the hell is with this empty space over here

IMG_1343

src/status_im/ui/components/tabbar/styles.cljs
```diff
@@ -46,6 +46,17 @@
...
+(def counter-public
+  {:right            3
+   :top              0
+   :position         :absolute
+   :background-color colors/blue
+   :border-radius    6
+   :width            12
+   :height           12
+   :border-width     2
+   :border-color     :white})
```
Ferossgp: To avoid background color leak around the border you can use 2 circles:
```suggestion
(def counter-public-container
  {:right            2
   :top              0
   :position         :absolute
   :border-radius    8
   :width            16
   :height           16
   :justify-content  :center
   :align-items      :center
   :background-color :white})

(def counter-public
  {:background-color colors/blue
   :width            12
   :border-radius    6
   :height           12})
```

src/status_im/ui/components/tabbar/core.cljs
```diff
@@ -67,11 +66,15 @@
...
-          [badge/message-counter @count true]])]
+       (when count
+         (cond
+           (or (pos? @count) (pos? (:other @count)))
+           [react/view {:style (if (= nav-stack :chat-stack)
+                                 tabs.styles/message-counter
+                                 tabs.styles/counter)}
+            [badge/message-counter (or (:other @count) @count) true]]
+           (pos? (:public @count))
+           [react/view {:style tabs.styles/counter-public}]))]
```
Ferossgp:
```suggestion
           [react/view {:style tabs.styles/counter-public-container}
            [react/view {:style tabs.styles/counter-public}]]))]
```

errorists (approved):

Serhy:
With iOS 13.3 and Android 8.1 no bugs found. 
@flexsurfer  please mind my commit when merging.

PR #10120, opened 2020-03-02
[#10119] Android app UI is slow after  background

### LINKED ISSUE #10119: Android app UI is slow when recieving new messages
# Bug Report
## Problem
When i open app UI is slow. Sometimes it's slow during using it

#### Expected behavior
UI is smooth

## Investigations 
_Updated 5 Mar 2020_

* [x] 1) `dispatch-later` is paused when app in background, use `:utils/dispatch-later` instead
* [ ] 2) it seems after a long time in background `:utils/dispatch-later` works differently as well (not proved technically, only visually) 
* [x] 3) `dispatch-later` with 20ms leads to subs and views recalculations, if dispatch is used, subs recalculated only once , `dispatch-later` should be replaced by another approach 
* [ ] 4) parametrized subscriptions leads to a lot of computations when app-db is changed, do not use them
* [ ] 5) subscriptions usage is not optimized, shouldn't be used in chat views at all, currently, we recalculate all data for every message, message should be added in event, subscription must only return the result without recalculating data
* [ ] 6) `process-response` signal is to often, probably we should send it not more than once a second
* [ ] 7) move all data modification to status-go, currently, we recalculate timestamps for all messages in subscription when receiving every new message
* [x] 8) `cljs-bean` should be used only when you want to access js object , not when you want to modify this data
* [ ] 9) in `flat-list` new array is created each time data is changed 

## ideal implementation
status-go signals json with new data -> status-react converts json to js, and append data to js array , js array is set to flat-list directly, no cljs data or subscriptions involved

flexsurfer:
partially f i x e s  #10119

### DIFF
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -60,19 +60,9 @@
               (multiaccounts.update/multiaccount-update :photo-path photo-path {:dont-sync? true}))))
 
 (fx/defn ensure-contact
-  [{:keys [db] :as cofx}
+  [{:keys [db]}
    {:keys [public-key] :as contact}]
-  (let [new? (get-in db [:contacts/contacts public-key])
-        us? (= public-key (multiaccounts.model/current-public-key cofx))]
-    (fx/merge cofx
-              {:db (-> db
-                       (update-in [:contacts/contacts public-key] merge contact))}
-
-              (cond
-                us?
-                (handle-update-from-contact-request contact)
-                new?
-                (transport.filters/load-contact contact)))))
+  {:db (update-in db [:contacts/contacts public-key] merge contact)})
 
 (fx/defn upsert-contact
   [{:keys [db] :as cofx}
--- a/src/status_im/transport/message/core.cljs
+++ b/src/status_im/transport/message/core.cljs
@@ -15,10 +15,10 @@
             [status-im.ethereum.core :as ethereum]
             [status-im.native-module.core :as status]
             [status-im.ens.core :as ens]
-            [cljs-bean.core :as clj-bean]
             [status-im.utils.fx :as fx]
             [taoensso.timbre :as log]
-            [status-im.ethereum.json-rpc :as json-rpc]))
+            [status-im.ethereum.json-rpc :as json-rpc]
+            [status-im.utils.types :as types]))
 
 (defn- js-obj->seq [obj]
   ;; Sometimes the filter will return a single object instead of a collection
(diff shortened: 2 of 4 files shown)


src/status_im/contact/core.cljs
```diff
@@ -60,19 +60,9 @@
...
    {:keys [public-key] :as contact}]
-  (let [new? (get-in db [:contacts/contacts public-key])
-        us? (= public-key (multiaccounts.model/current-public-key cofx))]
-    (fx/merge cofx
-              {:db (-> db
-                       (update-in [:contacts/contacts public-key] merge contact))}
-
-              (cond
-                us?
-                (handle-update-from-contact-request contact)
```
rasom, line 73: this code was doing nothing important?

flexsurfer: nothing :) но это не точно :)

rasom: oh nice, that's very comforting! :)

rasom:
If that's because of `:dispatch-later` queuing, can we try `:utils/dispatch-later` instead?

flexsurfer:
@Serhy could you please test this one when joining public chat with thousands of messages

Serhy:
Something is wrong. 
With latest https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200303-110209-40948b-pr-universal.apk / https://i.diawi.com/pEgCzY builds after new account is created/recovered there is endless spinner instead `(+)` button.
Fresh build installed.

After I re-open app and try to loging then a `sql: no rows in result set` error appears.

These are the logs after account created:
[loagcat.log](https://github.com/status-im/status-react/files/4281621/dispach-later-mod.log)

flexsurfer:
@Serhy probably it's latest develop, not this PR :(

churik:
@flexsurfer can confirm https://github.com/status-im/status-react/pull/10120#issuecomment-593972666 - endless spinner with builds from this PR and works normally in latest develop (nightly 04/03/2020)

flexsurfer:
@Serhy @churik thanks, fixed

Serhy:
@flexsurfer I tested this build on Galaxy Note 4 and doesn't look good (it got the similar behavior to what we had before moving processing of messages into status-go)

So with https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200303-110209-40948b-pr-universal.apk build when I join public chat having 1000+ messages during last 24h (stickers in my case), - then app freezes while all messages are coming up to 40-43sec (video https://drive.google.com/file/d/15jLN7ElNjHWicFpo9tubsXDg51jDP76b/view?usp=sharing)

While there are no lags at all when joining the same public chat in nightly (4th March) build.

flexsurfer:
thanks @Serhy , ok so i reverted this changes, and now this PR only fixes performance issue when open app from background

how to repro
1) hide app in background,
2) send lots of message to this user or public chat
3) open from background after  5-10m

rasom (approved):

Serhy:
No bugs found in latest build: opening app from background performance issue not reproduced, https://github.com/status-im/status-react/pull/10120#issuecomment-594455979 (when moving between screens also not reproduced).
With @churik help we did not see any regressions

Lets merge, @flexsurfer

PR #10126, opened 2020-03-04
temporary fix for contacts signal

flexsurfer:
every 30s app receives a signal with new contacts and app is slow if we use dispatch-later, this is a temporary solution for release, should be fixed properly later

### DIFF
--- a/src/status_im/transport/message/core.cljs
+++ b/src/status_im/transport/message/core.cljs
@@ -53,7 +53,10 @@
       (seq contacts)
       (let [contact (.pop contacts)]
         (fx/merge cofx
-                  {:utils/dispatch-later [{:ms 20 :dispatch [::process response-js]}]}
+                  ;;TODO temporary fix for release, we have and issue with contacts updates , UI is really slow
+                  ;;we need to inspect all subsctiptions and views, but for now to temporary make it better
+                  ;; we use dispatch instead dispatch-later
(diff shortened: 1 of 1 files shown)


rasom (approved):

churik:
@flexsurfer 
When "Fetch more messages" is using, chat view is not updated until reopening.
To reproduce:
- create new multiaccount
- join #status 
- tap on "Fetch more messages" 

Chat view is not updated.
History can be seen only if tap Back and reopen chat again.

On current nightly it is the same, and I believe it is likely related to https://github.com/status-im/status-react/pull/10120.

So it is not this particular PR fault, but I believe it should be fixed ASAP.
WDYT?

flexsurfer:
@churik i believe it was introduced here https://github.com/status-im/status-react/pull/9995

churik:
Ok, thanks for checking!
I'll check it and if yes, create separate issue.
It is very visible, so I think it is a blocker for v1.1.0

churik:
@flexsurfer yes, it is separate. logging issue

churik:
@flexsurfer tested that ENS name is updated for fetched messages + if new name is set it is updated after new messages in chat.
Any thoughts what other areas can be affected?

flexsurfer:
@churik nothing can be affected :) its only about performance

PR #10130, opened 2020-03-04
disable playstore upload

jakubgs:
Disabled playstore uploads because the API key has been deleted and no longer works.


rasom (approved):

churik:
Nothing to test here.

PR #10015, opened 2020-02-10
Integrate keycard interaction animation with keycard flows

Ferossgp:
This PR replaces all screens which were related to keycard connection with a bottom sheet. Before when interaction with the card was required we used to navigate to a screen which says to connect card, during errors and lost of the connection we navigate to other screens. Now all that is handled in the same sheet by animating to different states. 
Fixes #9365
Fixes #9704
Fixes #9676
Fixes #9660
Fixes #9664
Fixes #9626
Fixes #9702
Fixes #9334 

To be checked: #9451 #10023 #10032

PR #10135, opened 2020-03-05
[#10011] Listen to nfc state changes

### LINKED ISSUE #10011: NFC status (enable/disabled) is checked only before a keycard flow
Problem: 
Currently  NFC check is done only before navigation to screens which need NFC. If the user enables it using quick access sheet, while screen with "Turn on NFC" the app does not react to that. User should close/move to bg and open app again to see NFC changes.

Solution:
 It should be a listener as other checks like internet conenction

rasom:
fix #10011 

should be merged after #10015
 
status: ready

### DIFF
--- a/src/status_im/hardwallet/card.cljs
+++ b/src/status_im/hardwallet/card.cljs
@@ -51,12 +51,24 @@
              platform/android?)
     (.addListener event-emitter "keyCardOnDisconnected" callback)))
 
+(defn on-nfc-enabled [callback]
+  (when (and config/hardwallet-enabled?
+             platform/android?)
+    (.addListener event-emitter "keyCardOnNFCEnabled" callback)))
+
+(defn on-nfc-disabled [callback]
+  (when (and config/hardwallet-enabled?
+             platform/android?)
+    (.addListener event-emitter "keyCardOnNFCDisabled" callback)))
+
 (defn register-card-events []
   (doseq [listener @active-listeners]
     (remove-event-listener listener))
   (reset! active-listeners
           [(on-card-connected #(re-frame/dispatch [:hardwallet.callback/on-card-connected]))
-           (on-card-disconnected #(re-frame/dispatch [:hardwallet.callback/on-card-disconnected]))]))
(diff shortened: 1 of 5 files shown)


flexsurfer (approved):

src/status_im/hardwallet/card.cljs
```diff
@@ -51,12 +51,24 @@
              platform/android?)
     (.addListener event-emitter "keyCardOnDisconnected" callback)))
 
+(defn on-nfc-enabled [callback]
+  (when (and config/hardwallet-enabled?
```
flexsurfer, line 55: why do we need this when in every method ?

src/status_im/hardwallet/card.cljs
```diff
@@ -51,12 +51,24 @@
...
+    (.addListener event-emitter "keyCardOnNFCDisabled" callback)))
+
 (defn register-card-events []
   (doseq [listener @active-listeners]
     (remove-event-listener listener))
   (reset! active-listeners
           [(on-card-connected #(re-frame/dispatch [:hardwallet.callback/on-card-connected]))
-           (on-card-disconnected #(re-frame/dispatch [:hardwallet.callback/on-card-disconnected]))]))
+           (on-card-disconnected #(re-frame/dispatch [:hardwallet.callback/on-card-disconnected]))
+           (on-nfc-enabled #(re-frame/dispatch [:hardwallet.callback/check-nfc-enabled-success true]))
```
flexsurfer, line 70: register-card-events is only for android , why do we need same conditions in every method? we could simplify and add listeners right here 
`(.addListener event-emitter "keyCardOnNFCEnabled" #(re-frame/dispatch [:hardwallet.callback/check-nfc-enabled-success true]))`

src/status_im/hardwallet/card.cljs
rasom: Let's keep it as is, I don't want to touch stuff unrelated to this PR, we will go for refactoring some time soon

PR #10139, opened 2020-03-06
[wip] install referrer

flexsurfer:
* [x] read referrer
* [ ] call https endpoint

wip


flexsurfer:
@jakubgs why?

jakubgs:
Because I asked like 3 times in `#core` with this list of branches: 
https://gist.github.com/jakubgs/68283ba8983893dd5696578da6769a4c
If anyone objects. And no one did. So I removed them.
They were all +60 days old based on git author date.

jakubgs:
Ah right, because I refreshed the list which added to it these:
```
change-ios-app-id
ens/remove-name
feature/install-referrer
feature/install-referrer-status-module
fix/listen-to-nfc-status#10011
fix/trim-ens-name#10115
group-chat-creation-ui-2
master
snyk-fix-02e410d84143cc400b019dc62c52a6a1
test/waku-default-eth-test-fleet
upgrade-geth-1.9.11
```
This is the script I used: https://github.com/status-im/infra-utils/blob/master/github/list_old_branches.sh
Feel free to reopen it.

Ferossgp:
@flexsurfer will this PR handle also if the referral comes with an Universal Link (join.status.im/0x000...?invite=f79989f679386be7aa56dc1c4&cid=foo) or only if it is from play store? 
The params support was added here:
https://github.com/status-im/universal-links-handler/issues/31

flexsurfer:
this PR only implements android InstallReferrer, im not sure if we still should have it as separate PR, I'm out of context and doesn't work on it anymore

hesterbruikman:
@flexsurfer @Ferossgp I think only Play Store makes sense. If we reward only for new invites that install the app and purchase Stickerpack, the referral is irrelevant if it doesn't come from the Play Store as it implies the user already had the app installed

oskarth:

flexsurfer:
@Ferossgp is this PR still useful at some point? if so feel free to reopen

PR #10162, opened 2020-03-13
Ensure that keycard flows continue when card is already connected

rasom:
This commit ensures that a user can flawlessly interact with a keycard, regardless of the moment when the card was tapped to the device. To make it so:
- in case if the card was tapped before the interaction the user doesn't need to re-connect card to continue 
- in case if the connection with the card was lost during the interaction the application restarts that interaction as soon as the card was connected again (unless the user canceled the flow)

 Keycard related flows fixed in this PR:
- multiaccount creation (didn't work without card re-tapping after entering pin)
- multiaccount restoring via seed (didn't work without card re-tapping after entering pin and other steps)
- multiaccount restoring via pairing (didn't work without card re-tapping after entering pin and other steps)
- multiaccount unlocking (if card was tapped before entering pin app returns weird error popup)

 

status: ready

### DIFF
--- a/src/status_im/hardwallet/common.cljs
+++ b/src/status_im/hardwallet/common.cljs
@@ -104,6 +104,7 @@
 
 (fx/defn set-on-card-connected
   [{:keys [db]} on-connect]
+  (log/debug "[hardwallet] set-on-card-connected" on-connect)
   {:db (-> db
            (assoc-in [:hardwallet :on-card-connected] on-connect)
            (assoc-in [:hardwallet :last-on-card-connected] nil))})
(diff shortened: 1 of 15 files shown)


src/status_im/hardwallet/common.cljs
```diff
@@ -166,26 +174,36 @@
...
-                       :backdrop-dismiss? false
-                       :disable-drag?     true
-                       :content           (keycard-sheet-content on-cancel
-                                                                 connected?)}}))))
-
-(fx/defn hide-pair-sheet
+(fx/defn show-connection-sheet
+  [{:keys [db] :as cofx} {:keys [on-card-connected on-card-read handler]
+                          {:keys [on-cancel]
+                           :or   {on-cancel [::cancel-sheet-confirm]}}
```
Ferossgp, line 180: Maybe this could be on the same map level as other options?

Ferossgp (approved):
LGTM

src/status_im/hardwallet/common.cljs
rasom: of course we can, if there is any particular reason to do so

PR #10168, opened 2020-03-16
Fix messages when receiving on tab

### LINKED ISSUE #10154: No messages are shown in chat view if you receive them while being on different tab
# Bug Report

## Problem
Chat view is not updated with received messages if you received them while chat (1-1/public/group) view opened and you was at different tab (Wallet/Profile/Browser). Messages appear if to navigate back to Chats and again open chat with new messages.

**NOTE**: it's all fine when I put app on background with chat view and receive messages to this chat

Video: https://drive.google.com/file/d/18o7Tu1RH00Ntk4_weqVjqrMcTrxBvjtD/view?usp=sharing

#### Expected behavior
Messages are loaded when I open chat view from different tab

#### Actual behavior
Messages are loaded when I open chat view from different tab

### Reproduction

1) Open status and create new account (UserA)
2) UserA joins public chat (or Group/1-1 chat)
3) UserA switches to Wallet/Profile/Browser tab
4) UserB sends messages in the same public chat (UserA sees indicator of unread messages appeared on 'Chats' tab in that case)
5) UserA switches to Chat view to check what is the message received there? But there is none :(

### Additional Information
- Status version: develop nightly 11th March

cammellos:
While on tabs we want a slightly different behavior:
 Unread counter should increase, message should be loaded in the chat
On moving to the chat tab from one of these tabs, it should mark the messages as seen.

Fixes: #10154 

status: ready

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -41,13 +41,16 @@
 (fx/defn offload-all-messages
   [{:keys [db] :as cofx}]
   (when-let [current-chat-id (:current-chat-id db)]
-    {:db (update-in db [:chats current-chat-id]
+    {:db
+     (-> db
+         (dissoc :loaded-chat-id)
+         (update-in [:chats current-chat-id]
                     assoc
                     :all-loaded? false
                     :cursor nil
                     :messages-initialized? false
                     :messages {}
-                    :message-list nil)}))
+                    :message-list nil))}))
 
 (fx/defn handle-chat-visibility-changed
   {:events [:chat.ui/message-visibility-changed]}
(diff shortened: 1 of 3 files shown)


Ferossgp (approved):

flexsurfer (changes requested):
not sure if we want this, it will be a really bad UX, because the state will be lost, imagine you in  the middle of reading, then you want to check your wallet and back and state will be lost, messages will be reloaded @errorists wdyt ?

errorists:
@flexsurfer you're absolutely right, this reset doesn't sound like a win for our users

cammellos:
will change this

cammellos:
@flexsurfer @Ferossgp ready to be reviewed again

flexsurfer (approved):

PR #10173, opened 2020-03-17
Handle back button special cases, popover and bottom sheet

Ferossgp:


src/status_im/hardwallet/common.cljs
```diff
@@ -188,10 +188,11 @@
...
      {:dismiss-keyboard true}
      (bottom-sheet/show-bottom-sheet
-      {:view {:show-handle?      false
-              :backdrop-dismiss? false
-              :disable-drag?     true
-              :content           (keycard-sheet-content on-cancel connected?)}})
+      {:view {:show-handle?       false
+              :backdrop-dismiss?  false
+              :disable-drag?      true
+              :back-button-cancel true
```
flexsurfer: true by default, so we can omit it here

flexsurfer (approved):

churik:
Tested on IOS (briefly installing / rocovering / input fields / popuops in porfile / wallet)
On Android: 
- most of bottom sheets and popover can be closed via "back" button (i. e "Share profile", signing phrase, bottom menu in chat) - but not "Send transaction" dialog.

@Ferossgp is it intended? 

So far I can trigger and issue due to this:
- open 1-1 chat
- tap on request
- several times tap on Back button - you will be on "Home" instead of 1-1 with opened "Request transaction" dialog

Ferossgp:
@churik Request transaction is using a different component "Bottom panel" (it is similar to the bottom sheet but does not have drag and can't be closed on background press) which should be replaced with the bottom sheet which is used in the keycard. Not sure if I should do it here or in a separate PR

Serhy:
Originally the main issue was tapping back button while QR code pop-up on screen
It's cool that we cover it here + some of bottom sheets. Also it's issue is present in Release1.1
Taking that into account, I'd proceed then with what we have now and create separate issue for tapping hardware back button while Send/Request bottom panes is opened. 
I'll create an issue then.

churik:
Ready to merge.

Serhy:
> @churik Request transaction is using a different component "Bottom panel" (it is similar to the bottom sheet but does not have drag and can't be closed on background press) which should be replaced with the bottom sheet which is used in the keycard. Not sure if I should do it here or in a separate PR

Follow up issue for that is https://github.com/status-im/status-react/issues/10187 (covers back gesture and hardware back button)

PR #10176, opened 2020-03-17
Use react-native-screens on ios and android to increase performance

Ferossgp:
https://reactnavigation.org/docs/react-native-screens


Serhy:
Looks good to me with Android 8 and iOS 13.3. Navigation (including navigate back via hardware back button) kept. No crashes when opening app from background, logout, changing network etc.

After this PR merge we can just close https://github.com/status-im/status-react/pull/10175

Kudos @Ferossgp !

flexsurfer (approved):

PR #10191, opened 2020-03-20
Fix android tabbar animation

### LINKED ISSUE #10211: Screen is jumping when switching between tabs with open tabbar with and without text
# Bug Report

## Problem
Video: https://drive.google.com/file/d/1sKznc9AMSDEAouq_-PLEI4DLFy8paXoy/view?usp=sharing
Added to release, because issue is extremely visible. 

#### Expected behavior
navigation is smooth

#### Actual behavior
screen is jumping

### Reproduction

- Open Status
- Join to any chat (leave in chat screen)
- Switch tab

### Additional Information

- Status version: nightly 24/03/2020
- Operating System: Android

Ferossgp:
Fixes #10211

### DIFF
--- a/src/status_im/ui/components/tabbar/core.cljs
+++ b/src/status_im/ui/components/tabbar/core.cljs
@@ -11,19 +11,17 @@
    [status-im.i18n :as i18n]
    [re-frame.core :as re-frame]))
 
-(defonce visible? (animation/create-value 0))
-
-(defonce minimized-state (reagent/atom nil))
+(defonce visible-native (animation/create-value 0))
 (defonce last-to-value (atom 1))
 
 (defn animate
-  ([visible duration to]
-   (animate visible duration to nil))
-  ([visible duration to callback]
+  ([visible-native duration to]
+   (animate visible-native duration to nil))
+  ([visible-native duration to callback]
    (when (not= to @last-to-value)
      (reset! last-to-value to)
      (animation/start
-      (animation/timing visible
+      (animation/timing visible-native
                         {:toValue         to
                          :duration        duration
                          :easing          (animation/cubic)
@@ -37,12 +35,8 @@
 
 (defn minimize-bar [route-name]
   (if (main-tab? route-name)
-    (do
-      (reset! minimized-state false)
-      (animate visible? 150 0))
-    (do
-      (reset! minimized-state true)
-      (animate visible? 150 1))))
+    (animate visible-native 150 0)
+    (animate visible-native 150 1)))
 
 (def tabs-list-data
   (->>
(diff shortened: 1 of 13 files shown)


src/status_im/ui/components/tabbar/core.cljs
```diff
@@ -12,14 +12,13 @@
    [re-frame.core :as re-frame]))
 
 (defonce visible? (animation/create-value 0))
-
-(defonce minimized-state (reagent/atom nil))
+(defonce visible?' (animation/create-value 0))
```
flexsurfer: could we name it differently? self explained

flexsurfer (approved):

errorists:
what does it do? I can't recall what was wrong with android tab bar previously in my limited use

src/status_im/ui/components/tabbar/styles.cljs
```diff
@@ -115,11 +108,21 @@
...
+                                                       {:inputRange  [0 1]
+                                                        :outputRange [0 tabs-height]})}]}
          (when keyboard
-           {:position :absolute
-            :bottom   (- tabs-height)})))
+           {:position :absolute})))
+
+(defn space-handler [visible inset]
+  {:height     (+ inset minimized-tabs-height)
+   :margin-top (animation/interpolate visible
```
Ferossgp: Because of react-native animation does not support layout changes (margin-top), this animation value should be on js thread. Keeping 2 animation values allows us to animate this invisible block and make space for the native animation which uses translate.

This also solves the problem that touchable inside translate move do not receive touches https://github.com/facebook/react-native/pull/26374

flexsurfer: could you add this comment in the code ? would be useful

Ferossgp:
@errorists this PR fixes the problem on android which makes the text in tabbar opened state not touchable, now press zone is the icon and space around it at a height of the minimised state

errorists:
beautiful

flexsurfer:
@Ferossgp should it also fix not smooth animation of the screen when switching between tabs?

Ferossgp:
@flexsurfer, unfortunately, no, I will do that when I will connect the animation value of the bar with the animation progress of the screen. To do that I will need to use reanimated cause the screen uses it. This will fix the smooth transition cause reanimated allows the usage of native driver for layout. At the moment I integrate that library here: https://github.com/status-im/status-react/pull/10131

flexsurfer:
@Ferossgp @churik but we can't release v.1.2 without the fix, can we find a quick temporary solution?

churik:
Tested tabbar with and in minimized state (in ENS, in chat screen).

Tested on Android 10.

errorists (approved):

churik:
> @Ferossgp @churik but we can't release v.1.2 without the fix, can we find a quick temporary solution?

created separate issue https://github.com/status-im/status-react/issues/10211

churik:
Fixes #10211 
Tested on Android 10, Android 8 with minimized tabbar on Profile and on Chat tabs - no more jumping on switching.
Just in case tested IOS build.

PR #10198, opened 2020-03-23
Add Filipino translation

jinhojang6:
Add Filipino(a.k.a Tagalog) translation (fil.json)

- Worked well in AVD. But iOS doesn't support Filipino so cannot apply the translation

Update: the Filipino translation doesn't work in prod mode

Screen Shot 2020-03-23 at 11 29 16 PM

- removed an unused i18n key from all languages: custom-seed-phrase-text-2


jinhojang6:
@flexsurfer @yenda 
https://github.com/status-im/status-react/blob/2272fafc05c82514d829dc7c1d33a3dcf0f8dd60/src/status_im/ui/screens/multiaccounts/recover/views.cljs#L37

Do we need the `custom-seed-phrase-text-2` i18n key? `en.json` doesn't have a corresponding copy for this. It makes translators confusing as it is translatable on Lokalise as `empty` but it should not be translated.

flexsurfer:
@Serhy ?

Serhy:
Indeed we don't need this key anymore. 
I see this translation updated by purpose https://github.com/status-im/status-react/pull/9765 but the key with some translations left there in PR. In fact it's was used (empty string) and removing only translation gives this:

Screenshot 2020-03-24 at 10 16 59

So I removed this key usage. If it works I can merge.

Serhy (approved):

jinhojang6:
@flexsurfer 

The Filipino(Tagalog, i18n: `fil`) translation doesn't work in prod mode. But worked well on my Android emulator. Can you please help this? I found that Tagalog uses the `tl` language code in some cases.

bluebibi:
unsubscribe.

2020년 3월 25일 (수) 오전 12:27, Jinho Jang <notifications@github.com>님이 작성:

kberey:
Since IOs don't want to support Tagalog version, I will ask my Fellow country man that don't buy IOS instead go to Android lol.

flexsurfer:
for some reason it uses `fi.json` dunno why

flexsurfer:
@jinhojang6 fixed

flexsurfer (approved):

jinhojang6:
@flexsurfer It works like a magic. Thanks!

Issue #10197, opened 2020-03-23
Positive token balances display '0' after re-login in Wallet views


Serhy:
# Bug Report

## Problem

It started often to appear that after re-login (maybe 1 of 3 re-login) token balances display 0, even there are funds (balance > 0). Below are logcat when after login all token balances show 0. 

#### Expected behavior

Appropriate token balance displayed on Wallet  
#### Actual behavior

<img width="344" alt="Screenshot 2020-03-23 at 16 07 09" src="https://user-images.githubusercontent.com/8749671/77325253-a3a37100-6d20-11ea-99ad-f1653a5003f5.png">

### Reproduction

- Open Status and recover an account with any token where balance > 0 
- Relogin to account multiple times each login checking balance on main wallet view (switch network from Ropsten to Mainnet) 

### Additional Information

- Status version: develop nightly 23rd of March
- Operating System: Android and iOS

#### Logs

Logcat when tokens are 0 after login:
[tokens3.log](https://github.com/status-im/status-react/files/4369819/tokens3.log)

churik:
сс @rasom

churik:
@flexsurfer  - can you please take a loook, can it be related to https://github.com/status-im/status-react/commit/2272fafc05c82514d829dc7c1d33a3dcf0f8dd60?

flexsurfer:
no its not related

rasom:
that's what can be found in the attached logs
```
03-23 16:02:43.881 16708 16749 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :status-im.wallet.core/update-token-balance-fail
03-23 16:02:43.888 16708 16749 D ReactNativeJS: DEBUG [status-im.wallet.core:82] - Unable to get tokens balances:  {:code -32000, :message "header not found"}
```

rasom:
Almost likely it is thrown here https://github.com/status-im/status-go/blob/develop/vendor/github.com/ethereum/go-ethereum/eth/api_backend.go#L108

Serhy:
With the latest nightly it reproduced probably 1 of 7 relogins. 
I set up autologin and re-open Status multiple times to catch the bug. 
Yesterday it was easy to see the bug  on network change, but now 5 of 5 network changes (with relogin)  succeeded in relation to 0 token balances)

rasom:
@Serhy @churik 
[...]

PR #10206, opened 2020-03-24
[#10197] More retriers when requesting balances

rasom:
fix (well rather hide better) #10197 

status: ready

### DIFF
--- a/src/status_im/ethereum/json_rpc.cljs
+++ b/src/status_im/ethereum/json_rpc.cljs
@@ -7,7 +7,8 @@
             [status-im.native-module.core :as status]
             [status-im.utils.money :as money]
             [status-im.utils.types :as types]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [status-im.utils.utils :as utils]))
 
 (def json-rpc-api
   {"eth_call" {}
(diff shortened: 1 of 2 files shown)


rasom:
@Serhy 
> With the latest nightly it reproduced probably 1 of 7 relogins.

Can you please check if it is better this way? :)

src/status_im/wallet/core.cljs
```diff
@@ -34,7 +34,7 @@
    {:method            "eth_getBalance"
     :params            [address "latest"]
     :on-success        on-success
-    :number-of-retries 10
+    :number-of-retries 50
```
flexsurfer, line 37: O_o

rasom: yeah, I'm really tired of it

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -171,11 +172,20 @@
    "mailservers_getChatRequestRanges" {}
    "mailservers_deleteChatRequestRange" {}})
 
-(defn on-error-retry [call-method {:keys [method number-of-retries on-error] :as arg}]
+(defn on-error-retry
+  [call-method {:keys [method number-of-retries delay on-error] :as arg}]
```
flexsurfer: do we use delay?

flexsurfer, line 176: `delay` is not used currently ?

flexsurfer (approved):

src/status_im/ethereum/json_rpc.cljs
rasom: see line 179

Serhy:
We with @churik tested and could not reproduce originnal issue now with this PR builds on iOS and Android. 
I could not reproduce bug in Release1.2 build as well actually today (~20 of 20 successful attempts made)... 
Failed autotests are knows GHI (apart   test_refresh_button_browsing_app_webview which is not knnown but exists in develop too)

I don't see any obstacles to not merge it! Thanks @rasom !

PR #10207, opened 2020-03-25
fix #10052 make finding the public key easier in profile

### LINKED ISSUE #10052: Too hard to find and share chat key
# Feature Issue

Currently the only way to find your chat key is the tiny icon in the top right corner of `Profile` view.
Several users have complained that they were unable to find their own chat key and share it.

### User Story

As a user, I want to to be able to actually find my chat key.

### Description

The icon is too hidden, too small, and - as one user in `#support` suggested - looks like a logout button. What we need is a more obvious way - or ways - to get and share your chat key/ENS name.

yenda:
Ready:

* [x] confirmation that having the same effect in other users profile is desirablec: confirmed and done
* [x] figuring out what to do about the random names being gone: random name shown for user with ens name, public key for others
* [x] waiting for android icons: thanks @flexsurfer 

After @jakubgs review:

* [x] identify missing ens name (created issue #10218)
* [x] add share logo to toolbar in user profile


errorists:
iOS build is sad so I didn't test yet. 

> confirmation that having the same effect in other users profile is desirable

absolutely, yes. sorry for not making it clear in my comment

> figuring out what to do about the random names being gone

so this is a temporary quick fix anyway for 1.2 while I'm figuring out how to make multiple chat IDs work in the UI, wdyt about this

if random name -> disclose chat key
else -> disclose random name

it's still an improvement that might be just enough until we have a more thorough fix in place

Screenshot 2020-03-25 at 09 14 59

> waiting for android icons

Andrey kindly provided those in `chat-ui` discord. if it's not too much to ask, it would be cool to replace all instances of share on Android to use the platform specific icon like in accounts, browser, invite friends.

src/status_im/ui/screens/profile/components/views.cljs
```diff
@@ -51,46 +48,55 @@
...
+(defn chat-key-popover [public-key ens-name]
+  (re-frame/dispatch [:show-popover
+                      {:view :share-chat-key
+                       :address public-key
+                       :ens-name ens-name}]))
+
+(defn- profile-header-display
+  [{:keys [name public-key ens-name] :as contact}
+   allow-icon-change? include-remove-action?]
+  [react/touchable-opacity {:on-press #(chat-key-popover public-key ens-name)}
```
yenda: I would have used [:<> but it looks like we are still using a deadass version of reagent

jakubgs:
I checked the build and I noticed 2 things when looking at `@bridge` contact:

1. The contact "header" doesn't show the share icon
2. Despite ENS name being `bridge.stateofus.eth` only chat key is shown

Screenshot:
status_bridge_contact

yenda:
@jakubgs thanks for having a look, @errorists should I add the icon to the user profiles as well and not only the account profile?

flexsurfer (approved):

errorists:
@yenda yes please, let's have them consistent

yenda:
@jakubgs regarding the ens name it doesn't seem to be related to this PR, the ens name is actually there after reopening the app, so I think there is something amiss with ens name validation that should be fixed.

churik:
@yenda 
I see 2 issues here.
#### 1) No name is shown in headline when scrolling
photo_2020-03-31 12 59 48 jpeg 2020-03-31 13-00-28


#### 2) On tap on own user picture  don't see my ENS name (but I see it if I tap on "Share" icon in the heading)
On tap on "Share" icon in the headline:
photo_2020-03-31 13 05 29

On tap on your profile name or profile picture:

photo_2020-03-31 13 05 32

churik:
@yenda on build with  https://github.com/status-im/status-react/commit/900e3468b96cc492125e03b6092319abe299a7b0 issue 1 is still reproducible (tested on IOS)
photo_2020-04-15 13 49 00

churik:
failed e2e are fixed in https://github.com/status-im/status-react/pull/10337 and not related to PR

yenda:
@churik indeed sorry I missed that one, it was caused by the fact that the layout didn't worked with touchable-opacity, so I had to wrap the inner components in a view

churik:
@errorists would you mind to take a look?

churik:
@yenda thank you, all fixed, looks good to me!
1 question - why icons are different for IOS and Android, is it intended?
photo_2020-04-16 10 30 15 jpeg 2020-04-16 10-30-32

errorists:
@churik the different icons are a part of making it more obvious to their respective audience :)

churik:
approved by @errorists

PR #9800, opened 2020-01-12
[#9657] system messages ui changes in group chat

ghmulti:
Fixes #9657

### Summary

Suggest to create separate task for joining together events without separation (chained events) as this pr affects mostly styling 

#### Platforms
Tested on Android 

#### Areas that maybe impacted
Group chats

status: wip (awaiting feedback)

Screenshot from 2020-01-12 20-23-20

Colored block for margin/padding overview
Screenshot from 2020-01-12 20-21-24

Issue #9657, opened 2019-12-12
Correct the appearance of group chat system messages


errorists:
# Problem

Currently system messages in group chats appear like regular messages with mentions inside. It's an improvement over the previous UI, but that's not at all how we want them to look like.

![](https://media.discordapp.net/attachments/626738033701683200/654671507318112276/photo_2019-12-12_14.10.29.jpeg)

## Implementation

<img width="1261" alt="Screenshot 2019-12-12 at 16 12 51" src="https://user-images.githubusercontent.com/4932774/70725612-01292a00-1cfd-11ea-9e02-a89fc3a265fe.png">

---

each system message is just text centred in the chat thread. 8px margin on all sides. Black for username, Grey for the rest.

<img width="375" alt="Group 44" src="https://user-images.githubusercontent.com/4932774/70726377-469a2700-1cfe-11ea-9ef1-1bf8be59396f.png">

---

and that's how they appear in a chat

<img width="375" alt="Group Chat -_ Someone left chat" src="https://user-images.githubusercontent.com/4932774/70726364-439f3680-1cfe-11ea-96d3-1d44abc74b36.png">

Usernames are not links, anonymous random names appear without the `@` character preceding them.

## Acceptance Criteria

Implemented according to design. [Link to Figma here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=480%3A361)

## Notes

Thank you!

cammellos:
This has been merged now @errorists could you check if it can be closed please?
Thanks

errorists:
@cammellos yes, the contributor did an excellent job there, it can be closed. thanks

PR #10208, opened 2020-03-25
[#9657] system messages ui changes in group chat

cammellos:
Duplicate of #9800 , rebased with updated status-go version.

cc @ghmulti


flexsurfer (approved):

cammellos:
@churik @Serhy could you take a look at why tests fail? seems like is not able to pick up the correct text of messages.

I have run the tests from the branch against the build, so it's running the updated tests.

PR #10246, opened 2020-03-30
refactor and simplify mkFilter, use more restrictive rules

### LINKED ISSUE #10234: Nix store eating up too much disk space
## Problem

@flexsurfer reported and issue where running `make run-android` repeatedly eats up more and more disk space. He provided the following log, saying it happens every time he runs that target:

```
these derivations will be built:
  /nix/store/nl4q6658pjyq8nyxlqbr2idsr4yk41b8-status-react-patched-npm-gradle-modules.drv
building '/nix/store/nl4q6658pjyq8nyxlqbr2idsr4yk41b8-status-react-patched-npm-gradle-modules.drv'...
unpacking sources
patching sources
patching script interpreter paths in /private/tmp/tmp-status-react-5fab3268b/nix-build-status-react-patched-npm-gradle-modules.drv-0/project
/private/tmp/tmp-status-react-5fab3268b/nix-build-status-react-patched-npm-gradle-modules.drv-0/project/android/gradlew: interpreter directive changed from "/usr/bin/env sh" to "/nix/store/mg4y1vjgvz508n4qrhcilz52j0iil1f2-bash-4.4-p23/bin/sh"
Jetifier found 1215 file(s) to forward-jetify. Using 8 workers...
installing
Yarn modules changed, copying new version over
Copying node_modules from Nix store (/nix/store/2gyvyxs0ldch8nwijn2xhd27zpv0lz4g-status-react-patched-npm-gradle-modules/project/node_modules)...
```

jakubgs:
This should improve #10234 which was related to the derivation for `status-react-patched-npm-gradle-modules` using too broad filter and including some files which would cause unnecessary rebuilds when running targets like `make run-android`.

The thing is, XML files like `AndroidManifest.xml` need to be part of the `status-react-source-gradle-install` derivation because they are used in the build of `status-react-patched-npm-gradle-modules`.

I also changed the `TARGET` to `default` for `make run-android` to stop rebuilds of unnecessary things.

Status: __Ready__

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ _run-%:
 	npx react-native run-$(SYSTEM)
 
 # TODO: Migrate this to a Nix recipe, much the same way as nix/mobile/android/targets/release-android.nix
-run-android: export TARGET := android
+run-android: export TARGET := default
 run-android: ##@run Run Android build
 	npx react-native run-android --appIdSuffix debug
 
--- a/nix/mobile/android/jsbundle/default.nix
+++ b/nix/mobile/android/jsbundle/default.nix
@@ -20,12 +20,22 @@ in stdenv.mkDerivation {
       filter =
         # Keep this filter as restrictive as possible in order to avoid unnecessary rebuilds and limit closure size
         mkFilter {
-          dirRootsToInclude = [ 
-            "components/src" "react-native/src/cljsjs" "react-native/src/mobile" "src" "env/prod" "prod" # Taken from project.clj :profiles :prod :cljsbuild :builds :android :source-paths
-            "resources" "status-modules/cljs" "status-modules/resources" "scripts/version"
+          include = [ 
+            "src/.*" "prod/.*" "env/prod/.*"
+            "components/src/.*" 
+            "react-native/src" 
+            "react-native/src/cljsjs/.*"
+            "react-native/src/mobile/.*"
+            "status-modules/cljs/.*"
+            "status-modules/resources/.*"
+            "build.clj" "externs.js"
+            "project.clj" "prepare-modules.js"
+            "resources/js/.*"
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -269,7 +269,7 @@ _run-%:
 	npx react-native run-$(SYSTEM)
 
 # TODO: Migrate this to a Nix recipe, much the same way as nix/mobile/android/targets/release-android.nix
-run-android: export TARGET := android
+run-android: export TARGET := default
```
flexsurfer, line 272: could you pls elaborate on how this change works?

jakubgs: Sure, It's pretty simple.
There's a `nix/shells.nix` file that defines shells we use in CI and `Makefile`:
https://github.com/status-im/status-react/blob/afa2a92045ff171e7065864b64b13a693262e500/nix/shells.nix#L51-L71
These shells are defined using the `mkShell` derivation, here's a simple example:
https://github.com/status-im/status-react/blob/afa2a92045ff171e7065864b64b13a693262e500/nix/shells.nix#L25-L28
Which creates a the `keytool-sh` derivation that is then exported as `keytool` which includes `openjdk8` in `PATH` when created using `make shell TARGET=keytool`:
```
 > make shell TARGET=keytool
Configuring default Nix shell for target 'keytool'...

[nix-shell:~/work/status-react]$ which keytool
/nix/store/88zs3fkbwq5vxrynmc54j7s03vbyfbxw-openjdk-8u222-ga/bin/keytool
```
The `default` target shell is just one that includes all the tools we usually use:
https://github.com/status-im/status-react/blob/afa2a92045ff171e7065864b64b13a693262e500/nix/shells.nix#L30-L43

flexsurfer (approved):

PR #10254, opened 2020-03-31
[#10077] Can't accept agreement on ENS registration when app translat…

### LINKED ISSUE #10077: Can't accept agreement on ENS registration when app translated into French
# Bug Report

## Problem

When app translated into French or any other language which makes the `Agree to Terms and Conditions` text too long while registering ENS name, - the checkbox shifted out of screen making impossible it to be checked thus no way to buy ENS name (especially when large fonts are set to use on device but default device font on Samsung S8 already gives and issue)

We need to set margin to the `Agree to Terms of registration. I understand that my wallet address will be publicly connected to my username` 
#### Expected behavior

We need to set margin to the `Agree to Terms of registration. I understand that my wallet address will be publicly connected to my username`
Able to check checkbox when app is translated into French while registering ENS name.

#### Actual behavior

Screenshot 2020-02-25 at 16 06 00

### Reproduction
- Open Status while device is set to Francais and create new account
- Navigate to `Profile` -> `Add ENS name` -> `Add username` -> type any valid non existing text value and tap `->` to proceed to `Your username` view
- Scroll to the bottom to see `Agree to Terms of name registration` to see the checkbox

### Additional Information
- Status version: release 1.0.1

flexsurfer:
fixes #10077

### DIFF
--- a/src/status_im/ui/screens/ens/views.cljs
+++ b/src/status_im/ui/screens/ens/views.cljs
@@ -258,23 +258,22 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn- agreement [checked contract]
-  [react/view {:flex-direction :row
-               :margin-left 26 ;; 10 for checkbox + 16
-               :margin-right 16
-               :margin-top 14
-               :align-items :flex-start
+  [react/view {:flex-direction  :row
+               :margin-left     26                          ;; 10 for checkbox + 16
+               :margin-right    16
+               :margin-top      14
+               :align-items     :flex-start
                :justify-content :center}
    [checkbox/checkbox {:checked?        @checked
                        :style           {:padding 0}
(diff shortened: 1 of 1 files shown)


src/status_im/ui/screens/ens/views.cljs
```diff
@@ -258,23 +258,22 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn- agreement [checked contract]
-  [react/view {:flex-direction :row
-               :margin-left 26 ;; 10 for checkbox + 16
-               :margin-right 16
-               :margin-top 14
-               :align-items :flex-start
+  [react/view {:flex-direction  :row
+               :margin-left     26                          ;; 10 for checkbox + 16
```
rasom, line 262: can we either not touch this line or put the comment above it?

rasom (approved):

Issue #10243, opened 2020-03-30
No error message when a wrong PIN is entered creating a new account on a Keycard


guylouis:
# Bug Report

When a keycard multiaccount user tries to create a new account and by mistake enters a wrong PIN when confirming the creation, then there is no error message. This is both misleading (the user doesn't know what to do next) and dangerous (because the user can try several times and block his keycard because he'll reach the maximum number of tries)

Video https://drive.google.com/open?id=1sq9dd5wJWwQCiNgVxo4xOxpIj5znvSwv

# Solution 

Show to the user that he made a wrong PIN entry

guylouis:
cc @Ferossgp @rasom

PR #10260, opened 2020-04-01
[#10243] Handle wrong pin on keycard account creation

rasom:
fix #10260 

status: ready

### DIFF
--- a/src/status_im/hardwallet/core.cljs
+++ b/src/status_im/hardwallet/core.cljs
@@ -156,10 +156,10 @@
                                                                                           :puk         []})}
                 (common/hide-connection-sheet)))))
 
-(fx/defn dispatch-on-verified-event
-  [{:keys [db]} event]
-  {:dispatch [event]
-   :db       (assoc-in db [:hardwallet :pin :on-verified] nil)})
+(fx/defn clear-on-verify-handlers
+  [{:keys [db]}]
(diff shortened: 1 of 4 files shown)


flexsurfer (approved):

src/status_im/hardwallet/core.cljs
```diff
@@ -181,55 +181,51 @@
...
+                             (not on-verified-failure))
                     (if exporting?
                       (navigation/navigate-back)
                       (navigation/navigate-to-cofx :enter-pin-settings nil)))
-                  (common/get-application-info (common/get-pairing db) nil))
+                  (common/get-application-info (common/get-pairing db) nil)
+                  (when on-verified-failure
+                    (fn [_] {:utils/dispatch-later
+                             [{:dispatch [on-verified-failure]
+                               :ms 200}]}))
```
vitvly, line 222: Why this number - 200? (Edit - i see it wasn't introduced in this PR, but curious still)

vitvly (approved):

src/status_im/hardwallet/core.cljs
rasom: it is pretty random, the problem is that we need to wait till bottom sheet will be hidden. But I haven't even tried to put efforts into this as the whole process of account generation (specifically UI) is going to be rewritten soon

PR #10271, opened 2020-04-03
Fix incorrect loading of messages

cammellos:
In some instances when receiving messages from a mailsever in the chat
you are in, the flag `all-loaded?` would not be reset, meaning that
messages not in the current view would be added to the db, but would not be seen until actually
reloading the chat (go back home, open again).

### Testing

1) Open a chat with a few messages (> 50)
2) Keep scrolling on the latest message (basically scroll in place without moving) wait until all messages are received
3) Scroll up, you should see all messages
you can use the chat `test123` messages are ordered from 1 to 100, on 1.2 there are gaps, while there won't be if using this PR.

status: ready

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -162,7 +162,7 @@
   (when-let [current-chat-id (:current-chat-id db)]
     (if-not (get-in db [:chats current-chat-id :messages-initialized?])
       (do
-       ; reset chat viewable-items state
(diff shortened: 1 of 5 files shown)


flexsurfer (approved):

churik:
Confirm that issue exists on 1.2 and it is fixed in this PR.
Also tested fetching messages for other public chats - no issues are found.
Awesome job @cammellos !

PR #10292, opened 2020-04-07
nix: fix status-go source override

jakubgs:
Fixes custom source for `status-go` provided via `STATUS_GO_SRC_OVERRIDE` env variable.

I did not update the arguments when working on `mkFilter` function in: #10246

Status: __Ready__


rasom (approved):
it works locally, let's merge

PR #10294, opened 2020-04-07
Notifications

yenda:
status: ready


cammellos (approved):
it would be worth having an e2e test covering this, as we don't test this part so often and if the format changes we might not notice, maybe worth mentioning to test-team. Nice work!

Serhy:
**Issue 1**
PushNotification appears from any contact when received in 1-1 chat. 
I think we should not have notification for contacts who not in the user `Contacts` list (that like we had in the past: if message received from contact who not is in the contact list - no PushNotification appears.

**Issue 2**
When tapping on push notifications while in Status a different chat view was opened then an empty chat view is opened.  If I send any message from that view there is no message appears in view. They sent in "real" 1-1 chat but endless spinner will be below each message.

Video of issue: https://drive.google.com/file/d/1z4F-Eb4YXKIGCmCtxUlMPxOLV9e4LJas/view
(@cammellos maybe knows the secret behind 2nd issue)

cammellos:
A few things I was thinking:

It's probably best not to use `messages.new` for notifications, rather, we should probably have a new field that only contains the notifications that java processes.

That way we can customize notifications (only for contacts, only for some chats and not necessarily only for messages, for example we could have mentions).

In terms of enabling them only for contacts, I am not sure it will be clear enough for the user. Currently there's no incentive in adding someone to contacts (it's basically a noop), so I would be more inclined to have a separate menu (settings for a chat, we might have something similar already, which will come handy when we do disappearing messages, mute chat). 

Another thing is that we can always show an updated list of notifications, as status-go can instruct java to display the whole list.
For example instead of having  every single message differently, we could say:
"You received 2 messages from etc"
etc

cc @errorists

yenda:
@Serhy 
1. this was a limitation due to the fact that we just couldn't technically show a notification for messages that were not from contact since notifications required exchanging an id for the push notification service. Now it is not necessary anymore so we can show notifications for pretty much anything. But as @cammellos mentionned we can do some filtering on status-go side in the future based on user preferences.
2. I am not able to reproduce this issue on my android device, do you have so more specific reproducible steps, I tried this:
- open a chat and a add contact, exchange a few message
- open #status chat and put app in background
- send message from other device, open notification
In the video it looks like it's what you did and it opens a different chat but in my case it opens the right one

yenda:
@cammellos @Serhy ok actually I was able to reproduce in the PR build but not in the debug build

yenda:
@Serhy ok looks like it is good after rebase I wasn't able to reproduce. I think  we could go with the current settings of receiving notifications for all 1-1 chat, and next step is to do some work on status-go for filtering what gets a notifications, add settings in status-react and notifications for other chats and part of the app. @errorists @rachelhamlin wdyt?

rachelhamlin:
@yenda @Serhy +1 on 1-1 notifications for all dms. There's so little reason to add someone as a contact on Status right now—I'd rather ensure the PN service is running for as many users as possible and mirror SMS behavior. 

Later, we can add settings. Sounds brilliant.

flexsurfer (approved):

Serhy:
@yenda thank you. 
With the latest builds (`0.14.0 (20200409103022)`) tested in iOS13 and Android8.1 and Android9:
I'm trying to realise what's happening with PN now, because it's randomly (now until I understand) happens any of three:
a) Either this Issue2 from https://github.com/status-im/status-react/pull/10294#issuecomment-611419877 that separate 1-1 chat opens still reproduced OR
b) Everything is absolutely fine and no Issue2 OR
c) Error `Cannnot read property 'lastIndexOf' of null` appears when tap on PN (see below **Issue3** it's description and logs) 

**Issue3** This one is annoying in terms of reproduction because it was appearing all the time, but gone after I restarted Android devices. I anyway report it and trying to understand meanwhile when it appears. 
So there is no redirect to 1-1 chat view after tap on PushNotification but Status opens with error: `Cannot read property 'lastIndexOf' of null`

To reproduce: 
1) Create new accounts on Device1 and Device2
2) Device2 enables 'Notifications' from Profile
3) Device2 puts app on background
4) Device1 sends message Device2 

Screenshot 2020-04-09 at 19 22 43

Logcat:
 ```
04-09 17:23:40.478 31470 31470 E StatusModule: intent received: im.status.ethereum.module.TAP_NOTIFICATION
04-09 17:23:40.507 31470 31544 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :handle-universal-link
04-09 17:23:40.509 31470 31544 D ReactNativeJS: DEBUG [status-im.utils.universal-links.events:9] - universal links: event received for  status-im://chat/private/0x042d34e113f45eae1d01fb125f2cccd715a192d130c92d1712411a72f26ce3a6bd06d5243d155aeddf86450a04ad38894d738dede501039ae41691877eda9c7763
04-09 17:23:40.511 31470 31544 I ReactNativeJS: INFO [status-im.utils.universal-links.core:77] - universal-links: handling private chat 0x042d34e113f45eae1d01fb125f2cccd715a192d130c92d1712411a72f26ce3a6bd06d5243d155aeddf86450a04ad38894d738dede501039ae41691877eda9c7763
04-09 17:23:40.513 31470 31544 I ReactNativeJS: PRETTY PRINTED EXCEPTION
04-09 17:23:40.513 31470 31544 I ReactNativeJS: 
04-09 17:23:40.513 31470 31544 I ReactNativeJS: ***
04-09 17:23:40.513 31470 31544 I ReactNativeJS: NAME: "TypeError"
04-09 17:23:40.513 31470 31544 I ReactNativeJS: MESSAGE: Cannot read property 'lastIndexOf' of null
04-09 17:23:40.513 31470 31544 I ReactNativeJS: 
04-09 17:23:40.513 31470 31544 I ReactNativeJS: TypeError: Cannot read property 'lastIndexOf' of null
04-09 17:23:40.513 31470 31544 I ReactNativeJS:     at y (address at index.android.bundle:1:1763665)
04-09 17:23:40.513 31470 31544 I ReactNativeJS:     at n (address at index.android.bundle:1:3187866)
04-09 17:23:40.513 31470 31544 I ReactNativeJS:     at anonymous (address at index.android.bundle:1:3191568)
04-09 17:23:40.513 31470 31544 I ReactNativeJS:     at anonymous (address at index.android.bundle:1:3191206)
04-09 17:23:40.513 31470 31544 I ReactNativeJS:     at anonymous (address at index.android.bundle:1:3014735)
```

**Issue4** Very often I see PN not destroyed (marked as read) after it was tapped. Instead PN received from contacts accumulates into huge list. And when there are PN from multiple contacts it becomes messy so I ended up on tapping wrong PN and opening 1-1 chat with false impression there were no messages
**To reproduce:**
1) Create new account UserA and enables notifications
2) UserA puts app on background 
3) Another UserB sends text message to UserA
4) UserA tap on PN received and opens Status
5) UserA puts app on background
5) Now third UserC sends text message to UserA
6) UserA checks the list of PN to see if there are no PushNotification at all
**Actual result:** You got 1 messages  in 1 chat (the very last message). Nothing happens when I tap it (I can swipe right/left upon it to remove, but the with the next PN whole list of previous messages re-appears)

Screenshot 2020-04-10 at 11 49 56


**Issue5** Contact names in PN expose bold markdown (<b> tags)

Screenshot 2020-04-10 at 11 57 16

**Comment by Serhy** [comment]
**Issue6** I'd be very thankful if we fix the next (it a bug itself and needed to fixed to allow easier implementation of autotest)
If user puts Status on background by tapping hardware Back button then no navigation to 1-1 chat will happen after tap on PushNotification (user appears on Home view)
**To reproduce:**
1) Create new UserA and enabled notifications
2) UserA taps hardware Android Back button to put Status on background
3) UserB sends a message to UserA
4) UserA taps on appeared PN
**Actual result:** Chats main view opened (not 1-1 chat)

**Comment by rachelhamlin** [comment]
These all sound like potential blockers to me. Nice work @Serhy.

Also going to want to update that copy :) 

Do all the notifications compile into one @yenda? 

"You have a new message" would be my preference for my single message.
"You have 2 new messages" would be my preference for plural.

**Comment by yenda** [comment]
Issue 1 as discussed could be dealt with later, and is actually a feature TM

Issue 2 Issue3 I think those should be fixed now, I fixed two things: 
- Android intent were not being recreated even when the data changed because it would only care about the notificationId, so one problem was that when you received notifications from different chats they would not always reuse the same notificationId and inherit the intent of another chat, opening the wrong chat. So I made sure intents are now refreshed properly. 
- if the chat didn't exist already it wasn't created, hence the error in issue 3.

Issue4 Issue5 These issues were caused by the legacy implementation of groups, I removed it entirely because after testing on Android 7 it does some grouping already (albeit not as good as Android 8 and newer)

Issue6 I am looking into this one I thought it would be fixed along with 2/3 but I reproduced it again in the PR build

**Comment by yenda** [comment]
Ok 6 should be fixed in the next build. Android is actually calling onHostDestroy when you press the back button, which stops the service, so I just made sure the service isn't stopped then. That means even if you kill the app the service will keep running.

edit: well I happened to have fixed another bug but there is still an issue with the universal link being empty when app was killed and opened from notification (which happens with back button press)

**Comment by yenda** [comment]
@Serhy same bug already exists it's not related to PN but to universal links, when the activity is destroyed with the back button, the universal link isn't loaded on resume. So if you open a universal link with after the same steps you get the same result. I would suggest we create an issue for it and fix it separately, as it is not cause by this PR and is a broader issue that needs to be fixed

**Comment by yenda** [comment]
@Serhy I added a fix for universal links

**Comment by Serhy** [comment]
> actually calling onHostDestroy when you press the back button, which stops the service, so I just made sure the service isn't stopped then. That means even if you kill the app the service will keep running.

@yenda Indeed all issues fixed, push notifications are working fine right now!

However, what happens is that even when I want to kill the app on Android with enabled `Notifications` - it resumes after opening without asking me to login to account. Disregard how many times I closed it.
And I keep seeing each 10 seconds in logcat 
```

04-17 14:21:58.271  1596  1640 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :mailserver/check-connection-timeout
```

So `Close Status app` actions (close it from the list of device apps) has no effect. 
I'm sure we need to shut down if user closes Status from the list of device applications.

yenda:
@Serhy you need to logout if you want to kill the app, it's not just an app now but also a service, and killing the app doesn't kill the service, otherwise it will be killed as well when you press the back button

Serhy:
I don't know it's OK from security point of view. 
So having `Notifications` enabled - we always keep service alive and so Status is available to be opened without login any time (even no `Save password` / `FaceID` turned on) after user closed Status from list of apps.
And in that case we would need to align it to `Notifications` = OFF behavior then for consistency (so to update behavior of app NOT to kill the app if it closed)

yenda:
It's different because with notifications you always have a notification opened that says the app is running even after you kill it.

rachelhamlin:
Understood that this is a necessity of serverless notifications. 

What happens if I log out? Are notifications turned off? Do they have to be manually turned on again?

yenda:
@rachelhamlin they are turned off until you login again

Serhy:
Yes, that's true

Screenshot 2020-04-17 at 16 40 17

This notification keep staying if I close Status. 

Video of app behavior when notifications are ON: https://drive.google.com/file/d/1fRwmkYbB0nOBqP-KtTlDxO4gHHgBwG2T/view?usp=sharing

> What happens if I log out? Are notifications turned off? Do they have to be manually turned on again?

Notifications turned off (means you close the service), but once open Status and login -> Notifications enabled again.

andremedeiros:
So, I understand this might be a bit too late to pull a change like this, but I'd like to suggest changing the "Don't kill the app" message to go somewhere else, maybe inside the app itself.

We have a moment, when we offer to enable notifications, to educate the user and to explain why notifications are the way they are, and the shortcomings and advantages. Doing things like this, despite it acting as a reminder, will ultimately, IMHO, be a disservice to users in general.

@hesterbruikman curious to hear your thoughts.

hesterbruikman:
I kind of like having the `don't kill the app` message shown by the OS as it's a system constraint to do with how you manage your apps not how you manage settings within Status. Counting on @errorists wisdom on this though.

errorists:
@hesterbruikman I don't have an opinion yet without having used it, I'm not familiar at all with Android notifications. Once this is out in the open and I and everyone else will get their feet wet, I'll be able to give informed advice. I definitely don't mind releasing it as it is here.

Serhy:
Thank you @andremedeiros @hesterbruikman @errorists !
There were no problems apart that question. Then lets merge @yenda!
It won't be included in upcoming release/v1.3 but enough time to polish if needed in v1.4.

Issue #6375, opened 2018-10-18
[THE ISSUE] Add an unread messages indicator inside chats


errorists:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story
As a user, I want to see the number of unread messages in a chat so that I can continue my chats where I left them and track my progress in catching up with new messages.

### Description
[comment]: # (Feature or Bug? i.e Type: Bug)
_Type_: Feature

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
_Summary_: We need a visible way to indicate and track the number of unread messages left in a chat, and if desired, skip past them and mark all messages as read.

#### Expected behavior
User has an X number of unread messages in a chat. After opening the chat, we present a small badge to the bottom right with a counter of unread messages left. The chat is opened in the position before any new unread messages, where you last left it.
Scrolling downward should update the counter as the user scrolls past the unread messages, marking them as read. Tapping on the counter should scroll down to the last unread message and mark everything as read. When there are no unread messages left, we dismiss the counter.

### Solution
You can follow the intended design in this prototype https://framer.cloud/taJfZ
Mobile designs are here in [Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=568%3A0)
Desktop designs [Figma](https://www.figma.com/file/MRYBI5axcTwL4Ft8xAYhFOfy/Desktop-Chat?node-id=20%3A154)

Some things to be aware of, please enclose the badge in a touchable that's slightly larger than the button itself, adding a 12px padding inside the touchable or a margin outside the button. The text inside the badge should be right aligned. The SVG for the icon can be exported directly from Figma, it's called `scrollBottomIcon`, lmk if you need help or further directions.

<blockquote>![](https://framer.cloud/taJfZ/2/framer/social-1200x630.png)<div><strong>[Scroll - Unread](https://framer.cloud/taJfZ)</strong></div><div>This project was made with Framer, a design tool for turning original concepts into working prototypes.</div></blockquote>

chadyj:
Thanks @errorists 

@dmitryn @yenda @janherich Would love your feedback on this. I recall we had some scroll position issues?

dmitryn:
@chadyj yes, there are issues with maintaining scroll position on Android that i didn't overcome without using delay (not good for UX) https://github.com/status-im/status-react/pull/5665

oskarth:
[...]

PR #10304, opened 2020-04-08
Get replies directly from status-go

### LINKED ISSUE #10274: Chat screen is flickering
# Bug Report

version 1.1.0 iPhone
## Problem

Poor Majestic Icelandgull:
It just jumps, as if it’s refreshing the chat. 
For example, if you scroll up a few messages and then someone sends a new message to the chat, you will notice the chat jump so you know a new message has appeared. 
Now imagine this happening instantly every 4/5 seconds.
It’s not doing it now. But if I came out of status and then came back in half an hour it would probably be doing it again. 
It’s intermittent but when it does happen, it happens a lot. 

**But the other day I typed a reply to someone and when I hit send the message just disappeared due to the flickering.**

#### Expected behavior

no flickering

### LINKED ISSUE #10297: Blinking messages with message replies when joining such chat for the first time
# Bug Report

## Problem

When joining chat with messages for the first time (confirmed in public chat and 1-1 chat) the message replies starting to blinking each 5 seconds and whole list of messages redrawing too  
When re-navigating to that chat, then only username blinks only once in reply snipped. 

**NOTE:**  In a case of above scenario when the message which was replied to is visible (like the reply is the very next message) then no issues

#### Expected behavior

No message blinking when the joinning the chat for the first time

#### Actual behavior

Each 5 seconds after joining the chat with reply this what happens:

### Reproduction
- Join public chat where offline inboxed history exists

### Additional Information

- Status version: develop and release 1.2.0

#### Logs

These are logcat logs for what happens when blinking happens:

```
04-08 15:09:39.940  3150  3227 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :chat.ui/message-visibility-changed
04-08 15:09:40.173  3150  3227 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :chat.ui/load-more-messages
04-08 15:09:40.216  3150  3227 D ReactNativeJS: DEBUG [status-im.utils.handlers:32] - Handling re-frame event:  :status-im.chat.models.loading/messages-loaded


```

cammellos:
Fixes: #10297
Fixes: #10274

Before if a reply was missing (for example if it the message replied to
arrived after the reply) status-react would subscribe to it dynamically
and use that. This caused some flickering issues when offloading the
messages from the database, as the message being replied to would be
offloaded and later loaded back again, triggering scroll and ending up
in a loop.

This commit changes the behavior so that status-go always adds the
reply to the message. In case the message being replied to arrives after
the reply, this will be included in the message update, and status-react
will replace it in the local storage.

Don't merge because will have to prepare status-go release.

### Testing

Test receiving/sending replies, fetching replies from mailserver, also worth checking very quickly performance in fetching messages from the mailserver, to make sure there's no regression, as status-go is doing a bit more work.

status: ready


flexsurfer (approved):

test/cljs/status_im/test/chat/models/message.cljs
```diff
@@ -88,37 +88,38 @@
...
-                               :from "a"
-                               :clock-value 2
-                               :chat-id "a"})))
-  (testing "it returns false when the clock-value is less than the deleted-clock-value in chat"
-    (is (not (message/add-to-chat? {:db {:chats {"a" {:deleted-at-clock-value 1}}}}
-                                   {:message-id "message-id"
-                                    :from "a"
-                                    :clock-value 0
-                                    :chat-id "a"})))))
+(deftest message-loaded?
```
andremedeiros, line 91: 👏

test/cljs/status_im/test/chat/models/message.cljs
```diff
@@ -88,37 +88,38 @@
...
+                                       :from "a"
+                                       :clock-value 1
+                                       :chat-id "a"}))))
+  (testing "it returns true when it's already in the loaded message"
+    (is (message/message-loaded? {:db {:chats {"a" {:messages {"message-id" {}}}}}}
+                                 {:message-id "message-id"
+                                  :from "a"
+                                  :clock-value 1
+                                  :chat-id "a"}))))
+(deftest earlier-than-deleted-at?
```
andremedeiros, line 104: 👏

andremedeiros (approved):

churik:
@cammellos 
there is definitely no regression from performance side: so tested scenario
1) create new user
2) go to offline 
3) join chat with 500 messages
4) go back to online -> measure time from "Connected" state until messages are fetched (15 last, so 485-500)

attempt | 1 | 2 | 3 | 4 | 5 | Average
-- | -- | -- | -- | -- | -- | --
nightly | 24 | 23 | 21 | 22 | 21 | 22.2
PR | 14 | 19 | 18 | 18 | 18 | 17.4


On scrolling up can see that messages continue fetching, that's why chat is jumping (it is same on PR and nightly)
I didn't notice any issues with replies and https://github.com/status-im/status-react/issues/10297 is fixed.
However didn't see any changes according to https://github.com/status-im/status-react/issues/10274 - chat is still flickering (scrolling) when new messages are coming.
ezgif com-video-to-gif (1)

cammellos:
@churik thanks
I believe the issues described are the same, I have slowed down andrey's video and I can see it was due to replies.

> "However didn't see any changes according to #10274 - chat is still flickering (scrolling) when new messages are coming."


Yes, that's a known behavior, my understanding is that #10274 was describing a different issue (the same as #10297) .

What should we do should we close both and open a new one? I think it's a bit more complicated, because if you are on top of the chat, you want messages to scroll on new one, so you can follow, while if you are looking at old messages, then you probably don't want to scroll.

churik:
Ok, I'll create new one with proper explanation, because I see the issue that it is very uncomfortable to read chats when it is constantly updating with new issues, so basically scrolling position is not saved, you are always in "jumping" mode.

churik:
@cammellos 
should we merge it after cutting release branch and after preparing everything on go side, yes?

cammellos:
@churik thanks for opening the issue, it's actually a blocker for #6375  , which relies on being able to disable "auto-scroll" on new messages

I will prepare a status-go release and merge.
Thanks!

PR #10307, opened 2020-04-08
fix #10039

### LINKED ISSUE #10039: ENS search is case sensitive, or so it seems.
# Bug Report

It seems as though the ENS search is case sensitive. When the starting letter is capital names are not found and I cannot start a chat.

ENS however defines that names should be lowercased automatically.

#### Expected behavior

Lowercase automatically when searching for names.

This seems to only happen on the iPhone 10

yenda:
close #10039 

- wasn't able to reproduce the issue as it seems that all inputs are already
properly turned to lower case
- added :auto-capitalize :none property to the input field though to ensure
the keyboard doesn't force the user to start with an uppercase later

status: ready

### DIFF
--- a/src/status_im/ui/screens/add_new/new_chat/views.cljs
+++ b/src/status_im/ui/screens/add_new/new_chat/views.cljs
@@ -88,6 +88,7 @@
          ;; in onWillBlur navigation event handler
          :preserve-input?     true
          :accessibility-label :enter-contact-code-input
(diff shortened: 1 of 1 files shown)


cammellos (approved):

andremedeiros (approved):

churik:
failed e2e is not related to PR, ready to merge if nothing for QA

PR #10308, opened 2020-04-08
fix #10058 wrong domain name shown when connecting .eth

### LINKED ISSUE #10058: Wrong domain name is shown if .eth name is connected
# Bug Report

## Problem

The wrong ENS name is shown when linking a `.eth` name.
issue

#### Expected behavior

blah.eth should be shown

#### Actual behavior

blah.eth.stateofus.eth is shown

yenda:
close #10058 

add a check on the function that adds the stateofus domain to
usernames. if the username already contains a domain, don't
concat stateofus domain to it

# Testing

- add a .eth domain, on the confirmation screen the correct name should be shown instead of currently "domain.eth.stateofus.eth"

status: ready

### DIFF
--- a/src/status_im/ethereum/stateofus.cljs
+++ b/src/status_im/ethereum/stateofus.cljs
@@ -4,8 +4,21 @@
 
 (def domain "stateofus.eth")
 
-(defn subdomain [username]
+(defn subdomain
+  [username]
   (str username "." domain))
+
+(defn username-with-domain
+  "checks if the username is a status username or a ens name
+  for that we check if there is a dot in the username, which
+  would indicate that there is already a domain name so we don't
+  concatenated stateofus domain to it"
+  [username]
+  (when (and (string? username)
+             (seq username))
+    (if (string/includes? username ".")
+      username
+      (subdomain username))))
 
 (defn username [name]
   (when (and name (string/ends-with? name domain))
--- a/src/status_im/ui/screens/ens/views.cljs
+++ b/src/status_im/ui/screens/ens/views.cljs
@@ -361,10 +361,10 @@
   (case state
     :available
     (i18n/label :t/ens-username-registration-confirmation
-                {:username (stateofus/subdomain username)})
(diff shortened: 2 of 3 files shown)


src/status_im/ethereum/stateofus.cljs
```diff
@@ -4,8 +4,16 @@
...
 
-(defn subdomain [username]
-  (str username "." domain))
+(defn subdomain
+  "checks if the username is a status username or a ens name
+  for that we check if there is a dot in the username, which
+  would indicate that there is already a domain name so we don't
+  concatenated stateofus domain to it"
+  [username]
+  (if (and username
```
cammellos: What happens if it's nil? is it a valid value?
looks like if it's nil we would do:
`(str nil ".domain")`
Is that a valid case?

Otherwise I would have the check on top
`(when (seq username)...)`

cammellos (approved):

src/status_im/ethereum/stateofus.cljs
```diff
@@ -4,8 +4,16 @@
 
 (def domain "stateofus.eth")
 
-(defn subdomain [username]
-  (str username "." domain))
+(defn subdomain
```
flexsurfer, line 7: i would do it outside this function, we shouldn't ask for subdomain for stateofus names

Serhy:
@yenda I can still reproduce the issue with https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200409-151243-437a77-pr-universal.apk (tested on Android 8.1) and https://i.diawi.com/L2cJLW (tested on iOS13.3) 

1) Recover `fashion quality know robust copy neck stand embody entry task orient suggest` account
2) Add ENS name `statuse2e.eth`
**Actual result:** 
Screenshot 2020-04-10 at 15 19 32

yenda:
@Serhy indeed it was another screen I thought the issue was just about attaching a name. Fixed

Serhy:
Tested with iOS and Android8 and can confirm it's fixed

PR #10317, opened 2020-04-10
signals processing performance improvements [#10287]

### LINKED ISSUE #10287: Android app UI is slow when recieving new contacts every 30s
# Bug Report
## Problem
Every 30s app freezes for a few seconds

#### Expected behavior
no freezes

### Additional Information
currently, we add all ens names from public chats to contacts object, that means there are can be thousands of contacts, also there are "dead" names that mean they never will be resolved, but we are still trying to resolve them each 30s, and send false result to status-react, this leads to heavy computations and UI is slow

* [x] do not react on contact's field not used in status-react #10317
* [x] process all contacts in one event #10317
* [x] do not resolve same names each 30s #10313

Below still critical not solved issue

* [ ] implement `ens-name-async` similiar to `identicon-async` do not add to contacts every user from chat (it takes ~10s to get contacts from app-db)
* [ ] optimize subscriptions #10381

https://drive.google.com/file/d/16F2oCn9zvVTUrsd68tFlJ95kaGb1fQlV/view?usp=sharing

flexsurfer:
partially addresses #10287

we split all items from a signal by `dispatch-later [{:ms 20` this leads to performance issues because each item cause all subscriptions and views recomputation and re-rendrering, we need this only for messages, but still, we don't want to dispatch every message, we could split them by small batches 

tested on top on release app on my Android device, app much faster and responsive 

QA: test ens names resolution in chats, new messages, sync device with new chats

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -63,6 +63,11 @@
   [{:keys [current-chat-id] :as db} ui-element]
   (update-in db [:chat-ui-props current-chat-id ui-element] not))
 
+(defn dissoc-join-time-fields [db chat-id]
+  (update-in db [:chats chat-id] dissoc
+             :join-time-mail-request-id
+             :might-have-join-time-messages?))
+
 (fx/defn join-time-messages-checked
   "The key :might-have-join-time-messages? in public chats signals that
   the public chat is freshly (re)created and requests for messages to the
@@ -74,11 +79,15 @@
   public chat is not fresh anymore."
   [{:keys [db] :as cofx} chat-id]
   (when (:might-have-join-time-messages? (get-chat cofx chat-id))
-    {:db (update-in db
-                    [:chats chat-id]
-                    dissoc
-                    :join-time-mail-request-id
-                    :might-have-join-time-messages?)}))
+    {:db (dissoc-join-time-fields db chat-id)}))
+
+(fx/defn join-time-messages-checked-for-chats
+  [{:keys [db]} chat-ids]
+  {:db (reduce #(if (:might-have-join-time-messages? (get-chat {:db %1} %2))
+                  (dissoc-join-time-fields %1 %2)
+                  %1)
+               db
+               chat-ids)})
 
 (defn- create-new-chat
   [chat-id {:keys [db now]}]
@@ -106,6 +115,30 @@
               {:db (update-in db [:chats chat-id] merge chat)}
               (when (and public? new?)
                 (transport.filters/load-chat chat-id)))))
+
+(defn map-chats [{:keys [db] :as cofx}]
+  (fn [val]
+    (merge
+     (or (get (:chats db) (:chat-id val))
+         (create-new-chat (:chat-id val) cofx))
+     val)))
+
+(defn filter-chats [db]
+  (fn [val]
+    (and (not (get-in db [:chats (:chat-id val)])) (:public? val))))
+
+(fx/defn ensure-chats
+  "Add chats to db and update"
+  [{:keys [db] :as cofx} chats]
+  (let [chats (map (map-chats cofx) chats)
+        filtered-chats (filter (filter-chats db) chats)]
+    (fx/merge cofx
+              {:db (update db :chats #(reduce
+                                       (fn [acc {:keys [chat-id] :as chat}]
+                                         (update acc chat-id merge chat))
+                                       %
+                                       chats))}
+              (transport.filters/load-chats filtered-chats))))
 
 (fx/defn upsert-chat
   "Upsert chat when not deleted"
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -2,7 +2,6 @@
   (:require [re-frame.core :as re-frame]
             [status-im.multiaccounts.model :as multiaccounts.model]
             [status-im.ethereum.json-rpc :as json-rpc]
-            [status-im.utils.config :as config]
             [status-im.chat.db :as chat.db]
             [status-im.waku.core :as waku]
             [status-im.chat.models :as chat-model]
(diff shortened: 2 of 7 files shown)


src/status_im/chat/models/message.cljs
```diff
@@ -91,18 +82,16 @@
...
-                                      ;; if the chat has not been initialized, but run into
-                                      ;; some troubles disabling it, so next time
-                                     (update-in [:chats chat-id :messages] assoc message-id prepared-message)
-                                     (update-in [:chats chat-id :message-list] message-list/add prepared-message))
-
-                                  (and (not seen-by-user?)
-                                       (not= from current-public-key))
-                                  (update-in [:chats chat-id :loaded-unviewed-messages-ids]
-                                             (fnil conj #{}) message-id))}))))
+                {:db (cond-> (-> db
```
flexsurfer, line 85: no changes here, just formatting

cammellos:
Overall looks ok, if we don't need to batch better not batching, but @Serhy you will want to probably test performance on this one.

On this PR if I join #testmanymessages the UI is a bit choppy, it freezes while receiving messages (not for too long, but if I jump across tabs while I receive messages it's very visible, moto g5), while release is completely smooth. 

I think we need to find the right batch size, 50 on my phone seems to be too much to give a smooth UI, but likely lower values will be better.

To test perfomance you will want to create a chat with messages from many different users, and as soon as you open chat tabs (profile/wallet not home as that would offload messages). Once we have enough messages to fill up the screen it starts again being smooth as they are actually not rendered.

Also receiving many one to one messages from different users:
1) Go offline
2) Send many one to one messages from different users
3) Go online

src/status_im/transport/message/core.cljs
```diff
@@ -45,31 +34,34 @@
...
                   {:utils/dispatch-later [{:ms 20 :dispatch [::process response-js]}]}
-                  (handle-chat (-> chat (types/js->clj) (data-store.chats/<-rpc)))))
+                  (handle-chats (map #(-> %
+                                          (data-store.chats/<-rpc)
+                                          (dissoc :unviewed-messages-count))
+                                     chats-clj))))
 
       (seq messages)
-      (let [message (.pop messages)]
+      (let [messages-clj (types/js->clj (.splice messages 0 50))]
```
cammellos: I would probably have this `50` somewhere more visible, some constants in `config` or top of the file

cammellos (approved):
Provided performance are good enough looks good to me

src/status_im/chat/models.cljs
```diff
@@ -107,6 +107,25 @@
               (when (and public? new?)
                 (transport.filters/load-chat chat-id)))))
 
+(fx/defn ensure-chats
+  "Add chats to db and update"
+  [{:keys [db] :as cofx} chats]
+  (let [chats (map #(merge
```
rasom: can we use anonymous function instead of fn literal? It is a bit more readable when body contains more than one form

flexsurfer:
probably that  means (apply fx/merge is slow, then we need to implement fx for chats instead apply fx/merge, will do that, and probably we could also decrease the number of messages

cammellos:
@flexsurfer most likely the issue is with processing each message, any event needs to run in less than 16ms to give smooth UI, so batching too many is going to be well above that, previously even processing a batch of 2 messages was above 16ms on my phone, that's  the main reason for having it processing separately, but things might be different now , so we might be able to increase the batch.

I quickly looked at `apply`, and last time I checked there's not much difference, but haven't spent much time on it

flexsurfer:
ok I've made optimizations, should be faster now, but actually if you don't open chat  with lots of messages, app should be super fast and responsive

flexsurfer:
@Serhy do you have a time to try it ?

Serhy:
@flexsurfer @cammellos 
I tested Android and iOS build:
- can see ENS names (both stateofus and .eth) in public chats
- messaging in 1-1 and public works (including offline messaging)
- can sync devices (contact iOS13 and Android6 and Android 8) and contacts, public chats are synced. 

For the contact syncing the only difference between develop build is that right after the sync, - public chats are synced too but there were like no call for offline message history (so last message preview keep staying in `No messages` and non unread badge indication were too). History fetched right away after I navigated to public chat OR if to relogin to the account.

Performance checked too and there are app freezes  when receiving hundreds of messages. Video:
https://drive.google.com/file/d/1nX-55OehWtlqDAtoF7HuT_18UK6CnmKx/view
So receiving in public chat 500 messages each from different user freezes app for ~2-3sec on GalaxyNote4. On Galaxy S8 it's about 0.3-0.5sec

A bit worse when also 500 messages received from 5 different users in 1-1 chats:
Video: https://drive.google.com/file/d/1-VEys1br-wUv-H2AI3reB-O2IEL3uGJV/view?usp=sharing

On Release1.2.1 the same user who receives ~500 messages from 5 contacts also struggling with lags: https://drive.google.com/file/d/1v1McBo43-md4u5WcsBperpXBrfKGC4wJ/view?usp=sharing but a bit less than in current PR build.
 

So there is a performance degradation when receiving messages from online inbox comparing to Release1.2/Develop builds. 
+ no history fetched in Public chats after device syncing right away (which is issue but not a big deal because it's all fine upon navigation to such public chat or after relogin)

Taking into account that https://github.com/status-im/status-react/issues/10287 freezes app each 30 seconds i'd say this PR brings more benefits than regression. However I don't want to back in to the times of "App freezes when receiving many messages" :)

Not sure if we can go with it. I'd say yes, since 2-3 lag on the lowest device when receiving 500 messages is like an edge case and on for example on Galaxy S8 I have no performance regression at all...

cammellos:
@Serhy @flexsurfer  I am not sure I fully agree, I think we should split off what actually fixed the issue and introduce those changes rather than the whole lot.

There are at least 3 distinct changes in this PR:

1) Contacts are not batched anymore
2) Chats are not batched anymore
3) Messages are now in batch of 30 instead of a single

I believe what fixes @flexsurfer issue is mostly one fix that is already in develop (worth checking if that solves the issue) and likely 1, while I am not sure 3 has an impact.

The original issue it's not connected to receiving messages seems to me, as happens every 30 seconds, when contacts with ENS names signal is received, which partly has been fixed in develop, as much more contacts than necessary were sent.

On my device for example, even receiving a single message cause some slow downs, although not that much noticeable, 30/50 gives even more. 

We need to be able to have most of the events below 16ms to give a smooth UX, if we allow PR that have performance degration in , it's like playing whack-a-mole, I think we should have a no-freeze policy when it comes to these things, as they have an overall impact on the whole app, not just in edge cases.

We should optimize so that receiving a single message is below 16ms on a slower device, and then start batching once we are confident (maybe it is already the time, 30 might be too many, but a lower number might be ok, even 2 or 4, as we get better, that can become 8 16 and go up as the code gets more perfomant.).

What do you think?

flexsurfer:
@cammellos @Serhy i've reverted changes for messages, so in this PR only changes for contacts, chats and installations, if they are fine please approve and lets merge then, we want to cut 1.3 with this change today, thanks!

cammellos (approved):

Serhy:
We tested with @churik and it's ready to merge! 

* we can see ENS names (both stateofus and .eth) in public chats
* messaging in 1-1 and group and public works (including offline messaging)
* can sync devices (contact iOS13 and Android 8): contacts and public chats are synced.

No issues with performance (when switching between views) that were before when receiving 500 messages in public or from different individual contacts.

Lets merge, @flexsurfer , thanks!

PR #10328, opened 2020-04-14
Show correct error message when signing typed data

### LINKED ISSUE #8275: Show proper error when signing typed data
# Problem

Currently, if signed typed data has an error we show wrong password warning

## Implementation

Show popup with the error text

Initial issue: https://github.com/status-im/status-react/issues/8241

0x-r4bbit:
Prior to this commit `sign-message-completed` effect would always use
the `wrong-password` label for error handling. We now ensure that `wrong-password`
is only used in case the error in question has a code `5`, which is
associated to unmatching passwords.

In any other case we'll use the `message` attached to the `error`.

Fixes #8275

### DIFF
--- a/src/status_im/signing/core.cljs
+++ b/src/status_im/signing/core.cljs
@@ -301,7 +301,9 @@
   (let [{:keys [result error]} (types/json->clj result)
         on-result (get-in db [:signing/tx :on-result])]
     (if error
-      {:db (update db :signing/sign assoc :error (i18n/label :t/wrong-password) :in-progress? false)}
+      {:db (update db :signing/sign assoc
+                   :error (if (= 5 (:code error)) (i18n/label :t/wrong-password) (:message error))
+                   :in-progress? false)}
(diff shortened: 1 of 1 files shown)


0x-r4bbit:
Since this is a fix, I assume it'd be good to add an e2e test as well? 
If this is desired, I'd love any pointers to get me started on that.

cammellos:
@PascalPrecht We don't add e2e for every fix (they are fairly slow to run as they run on a different platform on emulators), if you feel this needs testing you can add some tests in clojure (they are unit/integration tests). 

In case the feature needs e2e tests, generally is the test team that writes them and maintains them.

cammellos (approved):

0x-r4bbit:
> In case the feature needs e2e tests, generally is the test team that writes them and maintains them.

@cammellos fair enough! Thanks for letting me know! 🙏

flexsurfer (approved):

PR #10332, opened 2020-04-14
fix: strip leading and trailing whitespace from QR Code data

### LINKED ISSUE #10306: enode QR scanner result includes whitespace/newline
## Problem

Currently if I generate a QR code with a trailing newline the input field will reject it:
status_invorrect_enode

#### Expected behavior

Newlines and whitespace characters around enode should be trimmed automatically.

#### Actual behavior

Fails enode regex test due to newline/whitespace.

#### Notes

This might also apply to scanning of ethereum addresses.

michaelsbradleyjr:
fixes #10306.

see also https://github.com/status-im/status-react/pull/10280#issuecomment-611374756.

### Summary

All leading and trailing whitespace is now removed from strings derived from QR codes, both in the case of `get-qr-code-data` (real camera) and `qr-test-view` (mock camera).

#### Platforms

- Android
- iOS

##### Functional

- mailservers

### Steps to test

- Use a QR code generator that will preserve leading and trailing whitespace in the text to be encoded, and generate a QR code that has e.g. leading and/or trailing spaces and new lines. On macOS you can use `qrencode` (`brew install qrencode`).
- Open Status
- Navigate to Profile -> Sync settings -> Mailserver
- Click the `+` in the top-right of the screen.
- Click the QR code button-icon in the second field that's titled *Mailserver address*.
- Depending on the `QR_READ_TEST_MENU` flag in your development setup, either enter some text that has leading/trailing whitespace, or take a picture of the QR code you generated in the first step.
- When you are returned to the *Add mailserver* screen notice that the leading/trailing whitespace has been stripped away.

status: ready

### DIFF
--- a/src/status_im/ui/components/camera.cljs
+++ b/src/status_im/ui/components/camera.cljs
@@ -1,6 +1,7 @@
 (ns status-im.ui.components.camera
   (:require [goog.object :as object]
             [reagent.core :as reagent]
+            [clojure.string :as string]
             [clojure.walk :as walk]
             [status-im.react-native.js-dependencies :as js-dependencies]
             [status-im.utils.platform :as platform]))
@@ -34,4 +35,5 @@
   (reagent/create-element default-camera (clj->js (merge {:inverted true} props))))
 
 (defn get-qr-code-data [code]
-  (.-data code))
+  (when-let [data (.-data code)]
+    (string/trim data)))
--- a/src/status_im/ui/screens/qr_scanner/views.cljs
+++ b/src/status_im/ui/screens/qr_scanner/views.cljs
@@ -1,6 +1,7 @@
 (ns status-im.ui.screens.qr-scanner.views
   (:require-macros [status-im.utils.views :refer [defview letsubs]])
   (:require [re-frame.core :as re-frame]
+            [clojure.string :as string]
             [status-im.i18n :as i18n]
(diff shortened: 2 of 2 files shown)


flexsurfer (approved):

src/status_im/ui/components/camera.cljs
```diff
@@ -34,4 +35,4 @@
   (reagent/create-element default-camera (clj->js (merge {:inverted true} props))))
 
 (defn get-qr-code-data [code]
-  (.-data code))
+  (string/trim (.-data code)))
```
flexsurfer:
```suggestion
  (when (.-data code) (string/trim (.-data code))))
```

flexsurfer (changes requested):
string functions are not null safe

src/status_im/ui/components/camera.cljs
michaelsbradleyjr: I updated in both places to use `when-let`, which carries an assumption (I think) that `nil` is okay to pass on even if `(.-data code)` or `@text-value` was `false`. Is it acceptable?

flexsurfer (approved):

churik:
Checked from:
- scanning QR from Start new chat > QR for chat key
- adding custom mailserver / bootnode
- EIP-681 links from the wallet.

In all places, whitespaces are trimmed.
Thanks!

PR #10333, opened 2020-04-15
Show redeem text for Keycard redeem txs

### LINKED ISSUE #10223: Update the Keycard request screen to Reedem transactions
# Feature Issue

The request screen already implemented in Status is enough to make the preloaded card scenario work. From an UX perspective however it would appropriate to show a specialized screen. The fields to detect in the typed data are:

1. receiver (address)
1. code (bytes32)

The screen should show something like “You are redeeming the preloaded assets to the address 0x…” so that the user can confirm that they are really redeeming to the correct address. Showing the code is probably not needed especially if it is not an alphanumeric value.

### Acceptance Criteria

Given a Redeem transaction, the specialized request screen UI is shown instead of the generic one.

### Notes
The request screen is implemented in the `feature/9275-keycard-signTypedData` branch.

vitvly:
Fixes #10223 

Additionally, this PR changes signature request screen so that it uses proper animations, and removes obsolete atom that was previously required in order for request screen to work.

QA notes: in order to test it, please run `keycard-cli info` and fetch the address from the `Cash Applet` section. This can be then by @gravityblast  used to generate a redeem link.


hesterbruikman:
@siphiuel do you have any test recommendations? DO I need a link to the redeem dapp to trigger the transaction?

vitvly:
@hester yes you'd need one to properly test it. In order to generate it, can you please run `keycard-cli info` and fetch the address inside the `Cash Applet` section?

@gravityblast - can you please generate a redeem link for @hesterbruikman ?

vitvly:
Fixed an issue with tx sheet hiding.
@rasom @Ferossgp @flexsurfer can you please review?

Ferossgp (approved):
LGTM! Maybe you could add the English strings also in this PR? Also would be interested from the QA if it fits on smaller screens, because the animation block together with all the content seems to take a lot of height.

churik:
@gravityblast can you please generate link for me?
```
Cash Applet:
  Installed: true
  PublicKey: 0x04375c36ea66de59322494a895fd11e3c4c6508118aeb81a7fdc29796ef087e408e3cee1752c710cbe7c7cd9b3842b94fa7db109241b1cdfdc619
```

vitvly:
@Ferossgp thanks for review!
I'll check how it looks on smaller screens, you're right - it does take lots of screen space.

hesterbruikman:
@gravityblast can you generate a link for me as well? 
`Cash Applet:`
`Installed: true`
`PublicKey: 0x0446fdaefce8c5772264ec3743eb2572cf6d178e6288cdb52ebec85b1489aa6e7d5efc440c5750f0eab185f391395649a1e2b39d54d1b03303763b5df6db8fc5b5`

vitvly:
@churik rebased against latest develop

churik:
@siphiuel 
### ISSUE 1: username in recipient instead of account name
according to design:
image
In fact:
photo_2020-05-05 16 27 16
 - Username instead of address is shown
 - "Cancel" instead of "Reject" (doesn't really matter IMO cc @hesterbruikman )

###  2 -  question: now I see sum \ amount in Dapp only 
photo_2020-05-05 16 32 03
@hesterbruikman 
from this design https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=9051%3A46835 it is not clear for me where user can see the assets / amount to redeem, and so it is in this PR.
Probably for some tokens I don't want to redeem them and pay the gas, how can I see that?

hesterbruikman:
Thanks for tagging @churik and apologies for the wait @siphiuel 

1. ISSUE 1: username in recipient instead of account name
Expected is:
_Use case A_: If user provides dapp with recipient address that corresponds to an account used in Status (any account in wallet) > **show username**
_Use case B_: If user provides dapp with recipient address that does not correspond to an account used in Status >> **show account address**

What is shown to me seems like use case A. I still need to check myself if use case B works. This depends on how the dapp allows setting the recipient address

2. question: now I see sum \ amount in Dapp only
Yes, you're right @churik it's an option to include this in the confirmation screen. This wasn't accounted for in this issue and at this point I think it's ok to handle this in the dapp. Something like `Tap to check gift` before actually redeeming. Not sure if it's the best solution, but in terms of process I'd rather iterate and improve later than hold up the PR now if aside from this point it works as expected.

churik:
Thanks for clarification @hesterbruikman !
I'll check both A and B cases.

vitvly:
@churik added handling for issue 1: now if will show name and image if receiver's account is present in Status, otherwise it will show numeric address.

churik:
Link is generated for Keycard A (user A)

### Case 1:
User A, Keycard A, Status account (main) in Dapp: no username / userpic, which is expected in this case (like in https://github.com/status-im/status-react/pull/10333#issuecomment-624096328 (issue 1 turned out is expected result)
photo_2020-05-06 15 10 21 jpeg 2020-05-06 15-11-41




### Case 2:
User B, Keycard B, Status account (main) in Dapp : error


### Case 3:
User C, ordinary multiaccount, account of user A a is set in Dapp: request keycard for signing message, so can't proceed

### Case 4:
User A, Keycard A, another account C (created) in Dapp : user will receive funds to account C



### Case 5:
User A, no keycard,  Status account (main) in Dapp (in progress, waiting for link): request keycard for signing message, so can't proceed


cc @hesterbruikman @gravityblast 
would like to have your opinions should we fix smth in this or leave it as it is.

hesterbruikman:
- *Case 1*, yes should be username

- *Case 2*, I'm not sure what's happening here. Is the dapp or Status not getting the signature from the expected Keycard? How would the dapp know what Keycard to expect?
_@gravityblast to verify: is the `Recipient` the Keycard address and the `Receiver` the address set in the dapp (currently Status account used with dapp?)_

- *Case 3*, this seems like the case in which User C has funds associated with Keycard, is using one of User C's accounts with the dapp, but wants to send them to user A, is that right @churik? If so, yes, makes sense you can't proceed. 
_As this requires an address input field in the dapp: @gravityblast I have something in mind for that based on Rocket. Let me look into that early next week._

- *Case 4*, looks good

- *Case 5*, That makes sense as in the scenario the user doesn't have a Keycard to proof ownership of associated funds. The alternative would be to add a field to manually enter the redeem code if it's not (only) provided through Keycard. Not sure if the dapp already has this. In this case, the first confirmation request wouldn't be required either, right?

_With case 4 I notice that the combination of `From`, `To` and `Amount` is confusing in a contract interaction. If it's always about `Signing a message` with 0 token value I'm not sure why we show `Amount`. Not an issue for this PR, but something we might need to fix because it makes this flow confusing_

churik:
@hesterbruikman for cases 3 and 5 I used ordinary multiaccounts, not created with Keycard. 
Dapp on pressing "Redeem" button requires signing message with keycard, so it is not possible to proceed and I guess it is fine for now.

gravityblast:
> Case 2, I'm not sure what's happening here. Is the dapp or Status not getting the signature from the expected Keycard? How would the dapp know what Keycard to expect?
@gravityblast to verify: is the Recipient the Keycard address and the Receiver the address set in the dapp (currently Status account used with dapp?)

@hesterbruikman @churik Yes the recipient of the "Redeemable" is the keycard address, which is the only key that can sign the message to redeem the tokens. If you use the wrong card, you cannot redeem them. The "receiver" is the address that will receive the tokens, in this simple dapp I set the current web3 account of your browser, but it could be specified in a text fields. So in the end it's just an address that the dapp specifies and the native screen shows. 
Is it clear that "account" is the actual "receiver" address? I'm just asking because for a moment I thought it was the account I was using to sign.

guylouis:
1. After some discussions, I suggest  we change the field name from "Account" to "Redeem to"

2. VItalyi the field "Message" is hardcoded as "Redeem assets" right ? If yes, let's remove this field totally since it's not usefull at all isn't it ?

@gravityblast @siphiuel @hesterbruikman

vitvly:
@churik @guylouis @hesterbruikman 
1. Changed the field title from `Account` to `Redeem to`.
2. Removed `Message field` altogether
3. The account image and name is shown only when `receiver` in the tx data matches one of current account's wallet addresses.

churik:
@siphiuel @gravityblast 

Redeem link is generated for Keycard A (user A) - basically as it is not possible to use it with real accounts, I think we can narrow down to 3 main cases

### Case 1 - success flow:
User A, Keycard A, Status account (main) in Dapp
(keycard A is associated with status account): 
photo_2020-05-18 17 48 13 jpeg 2020-05-18 17-50-08


### Case 2 - additional account added and set in dapp:
User A, Keycard A, another account C (created) in Dapp
(keycard A is associated with status account, which is not set in DApp): 
user will receive funds to account C
photo_2020-05-18 17 48 13 jpeg 2020-05-18 17-48-24
User will receive asset to account C.

### Case 3 - keycard of another user, invalid flow:
User B, Keycard B, Status account (main) in Dapp
(another keycard and account, not related to redeem link): 



For me it looks now that there is no case when address is shown - always I see profile name regardless of user, address and keycard.
So I think it is wrong currently for cases 2 and 3?
Sorry for pinging @hesterbruikman and @gravityblast again, but it is still unclear to me.

hesterbruikman:
For each case could you please add if the Keycard that is used, is associated to any account within the unlocked multiaccount? If yes, the behavior is expected.

guylouis:
As a note **these flows can be tested with a non-keycard multiaccount.** This makes it easier to understand what's going on.

a basic way to see the flow is 
i. dapp asks for a keycard to be tapped (any keycard)
ii.  if the keycard tapped is listed on the smart contract and has some gifts then the dapp transfers the metatx that was signed in i. to the smart contract. This meta tx is allowing a transfer of the gifts to the wallet adress used by the browser (because this is how  @gravityblast 's dapp has implemented it, he could have also added an option for the user to specify any another wallet adress like 0x1234...)

In a nutshell, I think with @gravityblast dApp, the account 'field' will always be the 3 word name of the account used by Status browser. Never a 0x1234... adress

this being said, I don't understand the error message in case C... @gravityblast wdyt ?

guylouis:
**also, as a postmortem (eventhough we're not done here :) )** 
for the future we have to document with more details (context, technical flow) our spec, this will help everyone - from implementer to tester - to understand how things work and what to expect of them 
redeem screen spec is https://github.com/status-im/status-react/issues/10223

vitvly:
@churik @guylouis @hester adjusted account display

churik:
### Issue 1: (probably unrelated to this PR, can be related to test DApp, needs clarification) Redeem button is not tappable after allowing access to wallet

When open redeem link (i.e. http://test-pn.keycard.cash/redeem//#/buckets/0xFE76E0bA86DE51804f17169Ed904B4b0704A0Fef/redeemables/0xe52938df98b1e4ccf6a37d37ffb96ec73d54bf9e  for 1 time) and Tap "Allow" button (or after changing wallet in DApp)- "Redeem" button is not tappable
photo_2020-05-21 11 03 45 jpeg 2020-05-21 11-03-56
Not reproducible with other DApps

### Issue 2: Default Gas limit is not set when Redeen fund not to Status account
#### Steps:
- add additional account A in wallet
- open redeem link 
- Allow access to wallet (reopen it to avoid Issue 1)
- tap with Keycard associated with redeem link

#### Expected result:
can sign transaction

#### Actual result:
default Gas limit is not set 
photo_2020-05-21 11 10 28

Not reproducible with other Dapps.


Apart from this: I think I can't emulate situation "Receiver address is not in Status multiaccount", because I can't specify it manually (is it right, @gravityblast ?)

gravityblast:
@churik 

> Tap "Allow" button (or after changing wallet in DApp)- "Redeem" button is not tappable

that might be a bug in the dapp. what device are you using?

> Default Gas limit is not set when Redeen fund not to Status account

so do you mean the gas is not correct when you create a new account and use that for dapps?
that should be related to status more than the dapp or the redeem screen, it's strange it's not reproducible with other dapps. 

> I think I can't emulate situation "Receiver address is not in Status multiaccount", because I can't specify it manually (is it right, @gravityblast ?)

not yet but I'll add an option to do it in the dapp

churik:
@gravityblast 
> that might be a bug in the dapp. what device are you using?

Xiaomi mi Note 3 Pro, Android 10

> so do you mean the gas is not correct when you create a new account and use that for dapps?

that should be related to status more than the dapp or the redeem screen, it's strange it's not reproducible with other dapps.

Agree, it is strange. Gas Limit is not set at all in that case cc @siphiuel

guylouis:
just for reference, and because browsing discord is sometimes hard, here's the overview of the logic of the redeem screen when called by a dApp (thanks @hesterbruikman )
https://miro.com/app/board/o9J_ksNjxAw=/

gravityblast:
@churik I deployed the fix for the dapp

churik:
Redeem link is generated for Keycard A (user A)

### Case 1 - success flow:
User A (account created with keycard A), Keycard A, Status account (main) in Dapp
(keycard A is associated with status account): 
User will receive funds to Status account

### Case 2 - additional account added and set in dapp:
User A (account created with keycard A), Keycard A, `account 1` (created) in Dapp
Gas error
photo_2020-05-26 17 02 59




### Case 3 - keycard of another user, invalid flow:
User B, Keycard B, Status account (main) in Dapp
(another keycard, not related to redeem link): 
Wrong Signer error
photo_2020-05-26 16 56 48

### Case 4 - regular account, created with password, keycard A:
User C (account created with password), Keycard A, Status account (main) in Dapp:
User will receive funds to Status account

### Case 5 - regular account, created with password, keycard A, additional account added and set in dapp::
User C (account created with password), Keycard A, `account 2` is added with private key
Gas error
photo_2020-05-26 17 02 59


Redeem screen view for all cases:
photo_2020-05-26 17 24 16

Are we OK for Gas error and redeem screen?
cc @hesterbruikman @guylouis @gravityblast

gravityblast:
@churik what happens after the gas error if you click redeem again?
I tried multiple times, sometimes it works, sometimes the redeem screen stays there forever, sometimes it close successfully but it doesn't open the tx screen.
@siphiuel if you try and you have the same problems, can you check maybe the logs to understand what's happening? It might also be a bug in the dapp but I can't see errors there.

churik:
@gravityblast 
on my Android10 if I tap \ cancel "Redeem" button I always see bottom "Send transaction" screen (tried around 20 times)
On what device are you trying?

gravityblast:
@churik pixel 2 xl

gravityblast:
@churik @siphiuel 
I still see the app is often blocking at the call to `web3.eth.estimateGas`.
I tried re-installing the apk from scratch and generating a new account, then:
1 - open the redeem url
2 - click on redeem and the redeem screen opens 
3 - I tap the card and the screen closes
4 - the transaction screen opens correctly
5 - I click on cancel so that the transaction screen closes

after that if I refresh the page or I close and re-open the browser, sometimes it blocks at point 3, so the tx screen doesn't open. if it opens, I try again and then many times it continues to stop at point 3.
I added a small wrench icon at the botton of the page, clicking that you can see some debug messages. When it stops, I see the last thing is `calling estimateGas`, but it doesn't go after that.
Does it happen to you too @churik @siphiuel ?

gravityblast:
@siphiuel @churik I found out that the problem is not related to the screen but to an error thrown from estimateGas, the usual wrong gas error. so I can catch it in another way.

churik:
@gravityblast 
I tried on build with commit https://github.com/status-im/status-react/commit/f159d84aa33de8ddaa41e606485df9e9650ad44d and yes, it blocks on point 3 for me with (I see "Redeeming" on the DApp, but transaction screen have never appeared)
Video is [here](https://drive.google.com/file/d/1vqUko2P5l6-_b4ofT8FaFxn0YQ2vfH4s/view?usp=sharing)

churik:
after last deployment fixes for dapp https://github.com/status-im/status-react/pull/10333#issuecomment-636097815 is not reproducible, so this https://github.com/status-im/status-react/pull/10333#issuecomment-634096153 is current behavior and it looks good to me.

guylouis:
@gravityblast do you think we can merge ?

gravityblast:
@guylouis yes we can merge it :+1:

PR #10338, opened 2020-04-15
Add component library skeleton

Ferossgp:
This PR adds structure of the lib, a basic README and implements:
```
• color variables
• text variables
• spacing values
• text component
• universal header component
• animated header component for scroll animations
```
The reanimated, gesture-handler, react, react-native wrappers were added directly into the lib.

All components previews are added into the **previews** namespaces and will be built as a separate app when shadow-cljs will be integrated.

For review, in-ns `status-im.ui.screens.views` navigation container can be replaced with `preview-screens` from `quo.previews.main`, which will render implemented views.


src/quo/design_system/typography.cljs
```diff
@@ -0,0 +1,30 @@
...
+
+(def font-regular {:font-family "Inter-Regular"}) ; 400
+
+(def font-medium {:font-family "Inter-Medium"}) ; 500 ff
+
+(def font-semi-bold {:font-family "Inter-SemiBold"}) ; 600
+
+(def font-bold {:font-family "Inter-Bold"}) ; 700
+
+(def monospace (if platform/ios? "Menlo-Regular" "monospace"))
```
errorists: @Ferossgp we should do something about this one, it's better if we try to use tabular digits and font features available in Inter than a separate monospace design https://github.com/status-im/status-react/issues/10201

src/quo/design_system/colors.cljs
```diff
@@ -0,0 +1,50 @@
+(ns quo.design-system.colors
+  (:require [reagent.core :as reagent]))
+
+(def white "#FFFFFF")
+(def black "#000000")
+
+(def light-theme
+  {:positive-01    "rgba(68,208,88,1)"     ; Primary Positive, text, icons color
```
flexsurfer, line 23: im not sure i understand these names and values it kinda hard to match with figma

Ferossgp: Added here some mapping of colors from figma with the code https://github.com/status-im/status-react/pull/10338/files#diff-27ee3b73723c60dff177c9746948ca4dR7 wdyt?

flexsurfer (approved):

PR #10339, opened 2020-04-15
fix Gradle permission errors during build

jakubgs:
This is a fix for the failure of local builds due to:
```
Could not create service of type ScriptPluginFactory using BuildScopeServices.createScriptPluginFactory().
Could not create service of type FileHasher using BuildSessionScopeServices.createFileSnapshotter().
```
By adding `--stracktrace` to `gradle` in Nix you can see the actual error:
```
Caused by: org.gradle.api.UncheckedIOException:
  Failed to create parent directory '/build/project/android/.gradle/5.6.4'
  when creating directory '/build/project/android/.gradle/5.6.4/fileHashes'
```
Which is caused by lack of write permissions on `.gradle` folder in the build diretory.


Ferossgp (approved):
💪

flexsurfer (approved):

PR #10342, opened 2020-04-15
fix: fix mailserver popup showing after canceling it

### LINKED ISSUE #10065: Popup error reappears if tap on "Cancel" in "Error connecting to mailserver" popup
# Bug Report

## Problem
Before if you dismiss "Error connecting to mailserver" popup (by tapping "Cancel") - you can see popup again on tapping banner in the header ("Tap to reconnect")
Currently, popup appears again in several seconds, so can't dismiss it

#### Expected behavior
button "Cancel" closes popup (con see it again if tap on "Tap to reconnect" in case of unsuccessful connection)

#### Actual behavior
poup reappears

### Reproduction

- Open Status
- Go to Profile > Sync settings > Mailserver
- Turn off automatic selection
- Add mailserver, that matches required format, but in fact invalid (i.e. enode://ee2b53b0ace9692167a510515bca3025695dbf0e1a68e1dff9716da620efb195f05a5b9e873fb9b75ac85de801106c565b8e2b6c5f0d93b8759d1578bfcaf03e:password@105.197.238.155:553 )
- Connect to it
- In "Error connecting to mailserver" tap "Cancel"

### Additional Information

- Status version: nightly 21/02/2020

jrainville:
### Summary

Fixes #10065 by setting a `dismissed` flag on Cancel, so the popup will not appear again.
Selecting a new mail server sets `dismissed` as false, so the error can appear again

#### Platforms

Tested Android simulator

### Steps to test

See issue above to see reproduction steps

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -61,6 +61,9 @@
 (defn preferred-mailserver-id [db]
   (get-in db [:multiaccount :pinned-mailservers (node/current-fleet-key db)]))
 
+(defn connection-error-dismissed [db]
+  (get-in db [:mailserver/connection-error-dismissed]))
+
 (defn mailserver-address->id [db address]
   (let [current-fleet (node/current-fleet-key db)]
     (:id (some #(when (= address (:address %))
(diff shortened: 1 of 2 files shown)


flexsurfer:
@jrainville thanks for the PR! great job! but this issue much deeper than it looks, Mailserver management is completely outdated and we need to reintroduce it with proper design and flows, but even if we want to fix the popup we should fix the cause - adding mailserver process should be stopped, not the consequences - multi popup

src/status_im/mailserver/core.cljs
```diff
@@ -504,22 +507,27 @@
...
-                                                 :advanced-settings
-                                                 :offline-messaging-settings)])
-          :extra-options       [{:text    (i18n/label :t/mailserver-retry)
-                                 :onPress #(re-frame/dispatch
-                                            [:mailserver.ui/connect-confirmed
-                                             current-fleet
-                                             preferred-mailserver])
-                                 :style   "default"}]}})
+         (update-mailserver-state db :error)}
+        (let [error-dismissed (connection-error-dismissed db)]
```
0x-r4bbit: Shall we call this `error-dismissed?` instead? It seems to be common practice to suffix boolean values with `?`. Or does this only apply to certain booleans?

src/status_im/mailserver/core.cljs
```diff
@@ -504,22 +507,27 @@
...
+             {:title               (i18n/label :t/mailserver-error-title)
+              :content             (i18n/label :t/mailserver-error-content)
+              :confirm-button-text (i18n/label :t/mailserver-pick-another)
+              :on-cancel           #(re-frame/dispatch [:mailserver.ui/dismiss-connection-error true])
+              :on-accept           #(re-frame/dispatch
+                                      [:navigate-to (if platform/desktop?
+                                                      :advanced-settings
+                                                      :offline-messaging-settings)])
+              :extra-options       [{:text    (i18n/label :t/mailserver-retry)
+                                     :onPress #(re-frame/dispatch
```
0x-r4bbit, line 526: This is unrelated to the changes in this PR, but I'm wondering if there's a reason `onPress` is written in camelCase here while most other events seem to be using kebab-case

0x-r4bbit (commented):
Nice @jrainville !

Learning a lot from your changes. I've left some inline comments about code-style related things.

src/status_im/mailserver/core.cljs
flexsurfer: kebab-case should be used

src/status_im/mailserver/core.cljs
flexsurfer: yes its better to use `?`

src/status_im/mailserver/core.cljs
jrainville: This was code that was already there, I just changed the indentation. I can change it to kebab-case if it's an available prop though.

src/status_im/mailserver/core.cljs
jrainville: Thanks for letting me know, I'll change it.

src/status_im/mailserver/core.cljs
flexsurfer: yeah, as i said its a really old code, should be re-implemented from scratch

PR #10349, opened 2020-04-16
fix: display value decimals instead of using exp notation

### LINKED ISSUE #9389: Transaction's amount is shown in exponential notation
# Bug Report

#### Expected behavior

Transaction's amount and fee are shown as regular numbers

[comment]: # (Describe what you expected to happen. Please include designs if possible.)

#### Actual behavior
Transaction's amount is shown in exponential notation.

[comment]: # (Describe what actually happened. Please include screenshots if possible.)

### Reproduction

Try to send 0.00000000000001 ETH and see values amount value in transaction sheet.

richard-ramos:
fixes #9389

### Summary
Displays the amounts in wallet and chat with decimal values instead of using exponential notation

##### Functional

- 1-1 chats
- wallet / transactions

status: ready

### DIFF
--- a/src/status_im/signing/core.cljs
+++ b/src/status_im/signing/core.cljs
@@ -157,7 +157,7 @@
          (if (nil? to)
            {:contact {:name (i18n/label :t/new-contract)}}
            (let [eth-value  (when value (money/bignumber value))
-                 eth-amount (when eth-value (money/to-number (money/wei->ether eth-value)))
+                 eth-amount (when eth-value (money/to-fixed (money/wei->ether eth-value)))
                  token      (get-transfer-token db to data)]
              (cond
                (and eth-amount (or (not (zero? eth-amount)) (nil? data)))
--- a/src/status_im/ui/screens/chat/message/command.cljs
+++ b/src/status_im/ui/screens/chat/message/command.cljs
@@ -174,7 +174,7 @@
       [react/text {:style {:margin-bottom 2
                            :font-size 20
                            :line-height 24}}
-       (str amount " " (name symbol))]
+       (str (money/to-fixed amount) " " (name symbol))]
       [react/text {:style {:font-size 12
                            :line-height 16
                            :color colors/gray}}


jrainville (commented):
Looks good. Does it make it display 18 decimals or something though?

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -182,7 +182,7 @@
       :error amount-error
       :accessories [[react/view {:style {:flex-direction :row}}
                      [react/nested-text {:style {:color colors/gray}}
-                      [{:style {:color colors/black}} (utils/format-decimals amount 6)]
+                      [{:style {:color colors/black}} (str amount)]
```
flexsurfer: we need to keep this

flexsurfer: this is by design

richard-ramos:
i.e.: `10200000` wei will be displayed as `0.0000000000102`, which is the value in decimals that the user enter in the send transaction screen. 

Since the amount of decimals that a transaction will have is specified by the user, I think it's okay to show the same decimal precision, otherwise we'll end up with transactions sending 0.00000 eth which is not the real amount. 

Contract interactions will be displayed as usual with fees being limited to 6 decimals (I did not make changes on that regard)

src/status_im/ui/screens/signing/views.cljs
richard-ramos: so, it's okay if it shows `0.00000` if the number is very small?

flexsurfer: mhmh, thats not expected for sure :) @errorists any ideas? we did this with Andrey, but not sure why we missed this case

richard-ramos: Screenshot from 2020-04-16 12-03-41
For reference, this is how it currently looks in this PR,   notice how the amount at the top contains all the decimals but the amount+fee below only shows the first 6

flexsurfer (approved):

Serhy:
@richard-ramos could you please rebase develop here, - iOS fails to build by some reason

Serhy:
I've rebased develop.

Serhy:
Looks good to me! (tested iPhoneX iOS13.3 and Samsung J7 Android 8.1)
Thanks you @richard-ramos

flexsurfer:
@richard-ramos could you pls squash commits

richard-ramos:
@flexsurfer, Done!

PR #10354, opened 2020-04-17
Refactor delete account algorithm to properly remove deleted account

0x-r4bbit:
Prior to this commit, deleted accounts would stay in the database but
set to `nil`. With this change accounts are properly removed from the database
when scheduled for deletion.

### DIFF
--- a/src/status_im/wallet/accounts/core.cljs
+++ b/src/status_im/wallet/accounts/core.cljs
@@ -274,7 +274,7 @@
                                  :on-success #()}]
                :db (-> db
                        (assoc :multiaccount/accounts new-accounts)
-                       (assoc-in [:wallet :accounts deleted-address] nil))}
+                       (update-in [:wallet :accounts] dissoc deleted-address))}
               (navigation/navigate-to-cofx :wallet nil))))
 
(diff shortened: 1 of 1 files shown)


src/status_im/wallet/accounts/core.cljs
```diff
@@ -265,7 +265,7 @@
                                  :on-success #()}]
                :db (-> db
                        (assoc :multiaccount/accounts new-accounts)
-                       (assoc-in [:wallet :accounts deleted-address] nil))}
+                        (update-in [:wallet :accounts] dissoc deleted-address))}
```
rasom: please remove an extra whitespace before `(update`, that's the reason why builds fail

0x-r4bbit:
@rasom Updated the PR accordingly :) 

Thanks for the review!

flexsurfer (approved):

Serhy:
LGTM! (Tested on iOS13.3 and Android 9)
Can add any account in multiaccount, can delete Watch only and then add the same account by via Seed phrase. E2E tests failed not because this PR i'm looking at them separately.
Thanks @PascalPrecht ! Let's merge!

PR #10360, opened 2020-04-17
feat: add i18n strings for the missing strings

### LINKED ISSUE #9678: Fix hardcoded items
# Bug Report

## Problem

Items that are hardcoded should be prepared for translation.

#### Expected behavior

ELI5 define these terms so they can be translated. 

#### Actual behavior

  - Add a watch only address, Enter the address to watch, and Enter address ... (https://github.com/status-im/status-react/blob/e7f01c42a7c4c3465daeef5aa70c434c6f5e1d9e/src/status_im/ui/screens/wallet/add_new/views.cljs)
- "fingerprint"
image

image

Thanks to @jinhojang6 for reporting.

jrainville:
Fixes #9678

I used this regex to find the strings in views in the `ui` directory:

```regex
\"[a-zA-Z]+[a-z0-9 A-Z\-\.\_\:]*\"
```

status: ready


translations/en.json
```diff
@@ -692,13 +696,19 @@
     "network-invalid-status-code": "Invalid status code: {{code}}",
     "network-invalid-url": "Network URL is invalid",
     "network-settings": "Network settings",
+    "network-info-press-to-refresh": "PRESS TO REFRESH",
```
flexsurfer, line 699: this is probably for debugging, we don't want to translate this

src/status_im/ui/screens/signing/views.cljs
```diff
@@ -220,7 +220,7 @@
                                  :padding-horizontal 16
                                  :padding-vertical   10
                                  :margin-vertical    14}}
-      [transaction-data-item {:label "Label"
+      [transaction-data-item {:label (i18n/label :t/label)
```
flexsurfer, line 223: i'm not sure if its something temporary? why we want to show "Label" ?

src/status_im/ui/screens/qr_scanner/views.cljs
```diff
@@ -39,10 +39,10 @@
                          :on-change-text #(reset! text-value %)}]
       [react/view {:flex-direction :row}
        [button/button
-        {:label    "Cancel"
+        {:label    (i18n/label :t/cancel)
```
flexsurfer, line 42: this is only for debugging, we don't need to translate it

flexsurfer (changes requested):
probably we should keep only biometric translations

flexsurfer:
feel free to reopen if needed

PR #10367, opened 2020-04-20
 re-frisk 1.0

flexsurfer:
re-frisk 1.0


src/status_im/ios/core.cljs
```diff
@@ -23,34 +23,32 @@
...
-                           "ShakeEvent"
-                           on-shake)
-        (dispatch [:set-initial-props (reagent/props this)]))
-      :component-will-unmount
-      (fn []
-        (.removeEventListener react/app-state "change" app-state-change-handler)
-        (.removeEventListener rn-dependencies/react-native-languages "change" on-languages-change))
-      :display-name "root"
-      :reagent-render views/main})))
+  ;;TODO fix that for all platforms, because this sub is always running
```
Ferossgp: Can we have one app-root and init for all platforms and call platform-specific under `when platform/*`? Seems like mostly it is the same code with minor differences.

src/status_im/ui/components/bottom_sheet/view.cljs
```diff
@@ -208,5 +208,6 @@
    (fn [insets]
      (reagent/as-element
       [bottom-sheet-view (assoc props
-                                :window-height @(re-frame/subscribe [:dimensions/window-height])
+                           ;;TODO fix that, because this sub is always running
+                                ;:window-height @(re-frame/subscribe [:dimensions/window-height])
```
Ferossgp: Seems like this is already fixed in reagent 8 because of subscription cache

cammellos (approved):
Let's get this bad boy in!!

PR #10372, opened 2020-04-20
feat: persistent send/receive action at bottom of asset list

### LINKED ISSUE #10302: Show persistant Send/Receive action at bottom of asset list
# Problem
`Send`/`Receive` are primary actions that can be scrolled out of view. Design shows these at the bottom of the asset list when the account card that includes the `Send`/`Receive` buttons is scrolled out of view

## Implementation
Frame 3
Design on [Figma](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=2607%3A768)

## Notes
To test go to 
- Wallet
- Tap on `...` > `Manage assets`
- Select a series of assets
- Tap on single account
- Scroll down

## Future Steps

michaelsbradleyjr:
fixes #10302.

### Summary

When the account card is scrolled out of view a send/receive button group appears at the bottom of the asset list; uses native animation to raise and lower the button group into and out of view.

### Review notes

I'm not entirely certain if the way it looks in dark mode is correct.

#### Platforms

- Android
- iOS

##### Functional

- wallet / transactions

##### Non-functional

- battery performance

Since native driver is used for the raise/lower animation there is probably not much impact. However `:scrollEventThrottle 1` was used with the `:on-scroll` handler, and maybe there is a more optimal value for that setting with respect to performance.

### Steps to test

- Open Status
- Navigate to Wallet -> Status account
- If there aren't enough assets listed to overflow the screen, then go back to Wallet -> Manage assets and select more assets. Then return to Status account.
- Scroll down the list until the account card begins to disappear from view; note how a send/receive button group slides up from the bottom. Scroll back up until the account card begins to reappear; note how the button group at the bottom disappears.

status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/account/views.cljs
+++ b/src/status_im/ui/screens/wallet/account/views.cljs
@@ -15,7 +15,8 @@
             [status-im.ui.components.list-item.views :as list-item]
             [status-im.utils.money :as money]
             [status-im.wallet.utils :as wallet.utils]
-            [status-im.ui.components.topbar :as topbar]))
+            [status-im.ui.components.topbar :as topbar]
+            [status-im.ui.components.animation :as animation]))
 
 (def state (reagent/atom {:tab :assets}))
 
@@ -28,12 +29,14 @@
                                     {:content        sheets/account-settings
                                      :content-height 60}])}]}])
 
-(defn button [label icon handler]
+(defn button [label icon color handler]
   [react/touchable-highlight {:on-press handler :style {:flex 1}}
    [react/view {:flex 1 :align-items :center :justify-content :center}
     [react/view {:flex-direction :row :align-items :center}
-     [icons/icon icon {:color colors/white-persist}]
-     [react/text {:style {:margin-left 8 :color colors/white-persist}} label]]]])
+     [icons/icon icon {:color color}]
+     [react/text {:style {:margin-left 8 :color color}} label]]]])
+
+(def button-group-height 52)
 
 (views/defview account-card [{:keys [address color type] :as account}]
   (views/letsubs [currency        [:wallet/currency]
@@ -57,7 +60,7 @@
       [react/touchable-highlight {:on-press #(re-frame/dispatch [:show-popover {:view :share-account :address address}])}
        [icons/icon :main-icons/share {:color colors/white-persist
                                       :accessibility-label :share-wallet-address-icon}]]]
-     [react/view {:height                     52 :background-color colors/black-transparent-20
+     [react/view {:height                     button-group-height :background-color colors/black-transparent-20
                   :border-bottom-right-radius 8 :border-bottom-left-radius 8 :flex-direction :row}
       (if (= type :watch)
         [react/view {:flex 1 :align-items :center :justify-content :center}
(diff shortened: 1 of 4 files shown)


Ferossgp:
@michaelsbradleyjr inside test we mock js code with empty objects. To mock the ScrollView just add `:ScrollView #js {}` at line 21 https://github.com/status-im/status-react/blob/develop/test/cljs/status_im/react_native/js_dependencies.cljs#L21

Ferossgp (approved):
:+1:

michaelsbradleyjr:
@Ferossgp done, thanks for the tip!

Ferossgp:
@michaelsbradleyjr e2e tests are failing most probably because of missing externs. You can add extern in `externs.js` file or use `oops.core/oget` to get the ScrollView

churik:
@michaelsbradleyjr error is shown on tapping on "status account" from wallet.
You can find logs / steps in any failed test in https://github.com/status-im/status-react/pull/10372#issuecomment-617592594

Ping me when error will be fixed.
Thanks!

michaelsbradleyjr:
@churik I have updated `externs.js` to include `"ScrollView" : function () {}`, per @Ferossgp's suggestion. I think that will fix the problem encountered in the e2e tests.

churik:
@michaelsbradleyjr 
"Send" and "Receive" buttons overlap the last asset in the list.
Monosnap 2020-04-23 13-06-53

michaelsbradleyjr:
@Ferossgp and @churik, I've added a WIP commit 85f564d615e0a59aa93229f1c45a70fbbaf011b8 that uses `padding-bottom` style to prevent overlap of the last item in the list by the bottom send/receive buttons.

In order to not disrupt the expected vertical spacing re: the *Load more* control, I needed to spread the `padding-bottom` logic across `status-im.ui.screens.wallet.account.views` and `status-im.ui.screens.wallet.transactions.views`. I don't think that can be helped without a more involved refactoring.

I tested locally that padding provides the desired visual affect for all three account tabs — *Assets*, *Collectibles*, and *History*, so I think it's ready for another round of testing on @churik's end, pending squash of the WIP commit.

@Ferossgp if you can let me know whether or not the solution is acceptable (maybe there's something simpler/better?), I'll squash the commit.

src/status_im/ui/screens/wallet/account/views.cljs
```diff
@@ -119,13 +125,64 @@
...
+    (let [anim-y (animation/create-value button-group-height)
+          scroll-y (animation/create-value 0)]
+      (anim-listener anim-y scroll-y)
+      [react/view {:flex 1 :background-color colors/white}
+       [toolbar-view name]
+       [react/animated-scroll-view
+        {:on-scroll (animation/event
+                     [{:nativeEvent {:contentOffset {:y scroll-y}}}]
+                     {:useNativeDriver true})
+         :scrollEventThrottle 1}
```
Ferossgp:
```suggestion
		 :contentContainerStyle {:padding-bottom button-group-height}
         :scrollEventThrottle 1}
```

This change can be added here to solve the problem of covering. Better to use here, because it is directly connected with this scroll view which adds the bottom bars. Also, it will add the correct spacing to any new element that can be to this scroll-view, and will be controlled by the same variable as height of the bar.

See [this](https://reactnative.dev/docs/scrollview.html#contentcontainerstyle)

michaelsbradleyjr: Thanks @Ferossgp!  I dropped my WIP commit and applied the change you suggested, it works great.

michaelsbradleyjr:
I believe this PR is ready to proceed back to review and testing.

churik:
Tested on Android 10 and IOS 13.
Works great!

flexsurfer:
ping @yenda  :)

PR #10388, opened 2020-04-21
fix(signing/views): allow discarding signing popup when tx in progress

### LINKED ISSUE #9977: Can't leave signing screen in case of losing internet connection
# Problem

When signing transaction initiated and internet disconnected user can't exit signing screen. `Cancel` button is available but not functioning. After some time we will get timeout error but until this not grayed out `Cancel` is confusing.

Screenshot 2020-01-31 at 12 24 53

## Steps to reproduce
- initiate stickerpack purchasing
- sign with the password and quickly turn off the internet
You will stay in signing screen with spinner running and non-functional `Cancel` button.

## Implementation

TBD

## Acceptance Criteria

TBD

0x-r4bbit:
### Summary

As discussed in #9977, there's an issue that the signing view popup cannot
be cancelled/discarded when a transaction is in progress.
This is because the `cancel-is-pressed` event is only conditionally applied to
the cancel button, rendering it non-functional for the time-being.

This commit changes that behaviour to always attach the event handler to
the cancel button, so that the popup can be closed even when a transaction
has been sent.

Fixes #9977

### Review notes

One thing to keep in mind:
Once the transaction has been signed and is in progress, there's no way to "cancel" it, so hitting the "Cancel" button in the UI really just "discards" it. One could consider changing the label to "Discard", but it probably won't work well because the UI is used in a context where cancelling makes sense.

### Testing notes

Best way to test this is to turn off the `signing/transaction-completed` event so the app thinks the transaction is in progress forever.

#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

### DIFF
--- a/src/status_im/ui/screens/signing/views.cljs
+++ b/src/status_im/ui/screens/signing/views.cljs
@@ -84,10 +84,10 @@
        [{:style {:color colors/black}} (displayed-name contact)]]
       [react/text {:style {:margin-top 6 :color colors/gray}}
        (str fee " " fee-display-symbol " " (string/lower-case (i18n/label :t/network-fee)))])]
-   [button/button (merge {:type :secondary
-                          :container-style {:padding-horizontal 24}
-                          :label (i18n/label :t/cancel)}
-                         (when-not in-progress? {:on-press #(re-frame/dispatch [:signing.ui/cancel-is-pressed])}))]])
+   [button/button {:type :secondary
+                   :container-style {:padding-horizontal 24}
+                   :label (i18n/label :t/cancel)
+                   :on-press #(re-frame/dispatch [:signing.ui/cancel-is-pressed])}]])
 
 (views/defview keycard-pin-view []
   (views/letsubs [pin [:hardwallet/pin]


src/status_im/ui/screens/signing/views.cljs
```diff
@@ -84,10 +84,10 @@
...
       [react/text {:style {:margin-top 6 :color colors/gray}}
        (str fee " " fee-display-symbol " " (string/lower-case (i18n/label :t/network-fee)))])]
-   [button/button (merge {:type :secondary
-                          :container-style {:padding-horizontal 24}
-                          :label (i18n/label :t/cancel)}
-                         (when-not in-progress? {:on-press #(re-frame/dispatch [:signing.ui/cancel-is-pressed])}))]])
+   [button/button {:type :secondary
+                   :container-style {:padding-horizontal 24}
+                   :label (i18n/label :t/cancel)
+                   :on-press #(re-frame/dispatch [:signing.ui/cancel-is-pressed])}]])
```
0x-r4bbit, line 90: Notice that transactions can't actually be cancelled, so hitting `Cancel` while the tx is in-flight really just discards the popup. Might be worth considering a different label depending on context.

flexsurfer:
ping @hesterbruikman @errorists

hesterbruikman:
Let me preface with, I don't have a deep understanding of the issue. 

It seems to me that indeed `Cancel` should be tappable when visible (fix 1). `Cancel` should then trigger a tx refresh (fix 2, if needed). Aside from these 2 fixes I do consider this a (nevertheless important) edge case and think these fixes are sufficient fine for now.

I expect two things will happen:
- The user remains in the Stickermarket where there should be a `pending` state before the stickerpack can be installed (@errorists I believe this is the case, right?)
- The user goes to wallet to check if the transaction happened. If they are still disconnected, they should see the loading state there, that will result in an `Error on loading` message. This should sufficiently inform the user that something is off in their last transaction and to check their history. 

In the future I expect an aggregated tx history and inapp notifications (include a tx pending state @errorists) to make it easier to recover from this glitch.

wdyt @errorists ?

errorists:
oh I'm not familiar with the issue either, I've read the `9977` description and the steps taken to fix that and it all seems correct

> The user remains in the Stickermarket where there should be a pending state before the stickerpack can be installed

I think so, yes

errorists (approved):

flexsurfer:
i don't think its correct, because you can't cancel tx when its in progress, so let's say user entered password and pressed send, now he is waiting and looking at the loading indicator, and he decides oh I want to cancel this tx, and make a new one instead, but its not possible to cancel tx because its already sent to the blockchain, so probably cancel should be replaced by close

hesterbruikman:
100% agree it's weird. I've had someone in a shop hit Cancel on my crypto payment asking me to pay again 😠, because it seemed Cancel was an option. Changing the copy on the button however doesn't seem right either. A button is not something that should change without user action.

An alternative would be to replace the loading state with an error state. Something like `Error processing, please verify tx history`; similar to the error state when the wallet can't load fiat values or history.

Again, though, feels like an edge case. And I'm happy with this as intermediary solution because the pending state is visible on the stickerpack and at least the user is no longer stuck having to close the app. I suggest creating another issue to describe desired behavior and put some more thought into that.

0x-r4bbit:
Thanks for all the feedback @flexsurfer @hesterbruikman and @errorists !

> An alternative would be to replace the loading state with an error state. Something like Error processing, please verify tx history; similar to the error state when the wallet can't load fiat values or history.

So just to clarify, we're not necessarily even talking about an error state but rather a "pending". I think the issue can be broken down to:

When a transaction is signed, it's in-flight (can't be undone) and **until** it completes (which might take a while) the `Cancel` button doesn't work. 

I wonder always renaming that `Cancel` label to `Close` is the way to go...

hesterbruikman:
> So just to clarify, we're not necessarily even talking about an error state but rather a "pending". 
Pending seems tricky because if connection is lost we wouldn't know when to change the state. The user might sit and wait for something to happen.

> I wonder always renaming that Cancel label to Close is the way to go...
That makes a lot of sense 😅

0x-r4bbit:
@flexsurfer I can update the label from `Cancel` to `Close`. Are we sure this can be safely applied to all cases where `Cancel` is used right now? Given that the label is i18n'd, it's probably used in many places. Also, how do we handle translations in general? 

I assume this PR should update labels for `t/cancel` in all languages?

PR #10391, opened 2020-04-21
fix: prevent multiple taps on Register button from opening several tx during ENS registration

### LINKED ISSUE #9566: Multitapping on "Register" button opens several transaction signing screens when register new ENS
# Bug Report

## Problem
Several screens with transaction signing are shown when tapping on "Register" button several times.
Found during testing https://github.com/status-im/status-react/pull/9546

#### Expected behavior
only one screen is opened 

#### Actual behavior

multitapping

### Reproduction

- Open Status
- Go to ENS
- Type new name
- Proceed to "Your username" screen
- Agree to the terms
- Tap on Register several times

### Additional Information

- Status version: nightly 02/12/2019

michaelsbradleyjr:
fixes #9566

### Summary

To prevent multiple taps from opening several transactions, the `::ens/register-name-pressed` event is debounced (via `status-im.utils.debounce/dispatch-and-chill`).

### Review notes

I set the "chill" time to 2 seconds — too long or short? I'm not sure.

### Testing notes

If you tap on the Cancel button quickly to do a repeat test of the `dispatch-and-chill` effect, keep in mind the 2 second chill-time must expire before tapping Register again will have an effect.

#### Platforms

- Android
- iOS

##### Functional

- user profile updates

### Steps to test

- Open Status
- Navigate through Profile -> ENS usernames -> Your username *(requires some input)*
- Check the *"Agree..."* box
- Tap the Register button multiple times quickly; note that only one transaction is opened.

status: ready

### DIFF
--- a/src/status_im/ui/screens/ens/views.cljs
+++ b/src/status_im/ui/screens/ens/views.cljs
@@ -295,7 +295,7 @@
        (i18n/label :t/ens-deposit)]]]
     [button {:disabled?    (not @checked?)
              :label-style  (when (not @checked?) {:color colors/gray})
-             :on-press     #(re-frame/dispatch [::ens/register-name-pressed])}
+             :on-press     #(debounce/dispatch-and-chill [::ens/register-name-pressed] 2000)}
      (i18n/label :t/ens-register)]]])
 
 (defn- registration


andremedeiros:
I don't even think the button should be re-enabled until we get a response back.

Is there a visual indication that there is work being done at all?

michaelsbradleyjr:
> Is there a visual indication that there is work being done at all?

I'm not 100% sure I know what you mean.  When you tap the Register button, a transaction card appears very quickly. I think most users would then proceed to go ahead with the transaction. To get a behavior where the Register button might seem stuck, a user would need to deliberately be quick in tapping the Cancel button in the transaction card and then tap Register again (a Status tester might do that, which is why I mentioned it in *Testing notes*).

> I don't even think the button should be re-enabled until we get a response back.

My initial idea was to make it a little state machine with a ClojureScript `atom`, but control doesn't get passed back to [`registration-bottom-bar`'s on-click handler](https://github.com/status-im/status-react/blob/361eeaa16158d2066ef840d19fab69da3ea6869f/src/status_im/ui/screens/ens/views.cljs#L298); so I'd need to pass the atom along and I wasn't sure if that's how this kind of problem is typically solved with re-frame. On a hunch I guessed that maybe this kind of problem is solved with debouncing and a code search in status-react turned up a couple of cases that seem similar and that use `dispatch-and-chill`, so that's the direction I went. I'm happy to go a different route if `dispatch-and-chill` is not the correct approach.

flexsurfer (approved):

flexsurfer (approved):

churik:
@michaelsbradleyjr can you please rebase it to current develop to have builds in PR?

michaelsbradleyjr:
@churik it's been rebased.

rachelhamlin:
Nice work @michaelsbradleyjr! I'll be very happy to see this one fixed.

flexsurfer (approved):

Serhy:
Looks good to me! Can't reproduce initial issue on iOS and Android (v8). I'm not sure yet why test_can_add_existing_ens e2e test failed but it's false failure. 
Thanks a lot  @michaelsbradleyjr ! Let's merge in develop!

michaelsbradleyjr:
@Serhy I just now rebased against `develop` and force-pushed, thinking maybe the failures were related to being behind `develop`, sorry if it was bad timing to force-push.

michaelsbradleyjr:
Also, I forgot to sign the commit as part of the rebase... fixing now.

Serhy:
One min, something wrong (likely develop)

Right after start in emulators there 
Screenshot 2020-04-24 at 17 47 49

PR #10400, opened 2020-04-22
start loading messages earlier

flexsurfer:
start loading messages earlier

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -10,17 +10,15 @@
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.i18n :as i18n]
             [status-im.mailserver.core :as mailserver]
-            [status-im.transport.message.protocol :as transport.protocol]
             [status-im.ui.components.colors :as colors]
             [status-im.ui.components.react :as react]
             [status-im.ui.screens.navigation :as navigation]
             [status-im.utils.clocks :as utils.clocks]
-            [status-im.utils.config :as config]
             [status-im.utils.fx :as fx]
-            [status-im.utils.gfycat.core :as gfycat]
             [status-im.utils.platform :as platform]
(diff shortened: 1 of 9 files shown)


src/status_im/chat/models.cljs
```diff
@@ -295,16 +245,17 @@
               (when-not (group-chat? cofx chat-id)
                 (transport.filters/load-chat chat-id))
               (when platform/desktop?
-                (mark-messages-seen chat-id))
+                (message-seen/mark-messages-seen chat-id))
               (when (and (one-to-one-chat? cofx chat-id) (not (contact.db/contact-exists? db chat-id)))
-                (contact.core/create-contact chat-id)))))
+                (contact.core/create-contact chat-id))
+              (loading/load-messages))))
```
flexsurfer, line 251: only one change is here, all other because of cycle deps :(

flexsurfer:
@cammellos tested  on real device, i can't see chat logo anymore, so much better

src/status_im/chat/models.cljs
```diff
@@ -295,16 +245,17 @@
...
 
 (fx/defn navigate-to-chat
   "Takes coeffects map and chat-id, returns effects necessary for navigation and preloading data"
   [cofx chat-id]
   (fx/merge cofx
-            (navigation/navigate-to-cofx :chat {})
-            (preload-chat-data chat-id)))
+            (preload-chat-data chat-id)
```
cammellos, line 257: Is the order important here? Maybe we should leave a comment if so.

cammellos (approved):
@flexsurfer nice!

src/status_im/chat/models.cljs
flexsurfer: re-frame doesn't guaranty order of fx handlers, so I did it just in case :)

PR #10402, opened 2020-04-22
Revert "gradle: upgrade plugin to 3.6.0"

rasom:
This reverts commit 938f3df8431bd3d16f9954ecdb84bea8a68b4532.

fixes blank screen on startup on Android


jakubgs:
Maybe instead we can try downgrading to `3.5.3` ?
According to the docs we are severely outdated in regard to Gradle:
https://developer.android.com/studio/releases/gradle-plugin#updating-gradle
Maybe we should try disabling the new APK creator:
https://developer.android.com/studio/releases/gradle-plugin#zipflinger
```
android.useNewApkCreator=false
```

jakubgs:
Testing that in: https://ci.status.im/job/status-react/job/combined/job/mobile-android/12571/

flexsurfer (approved):

rasom:
@jakubgs I just want to unblock `develop`

jakubgs (approved):

PR #10403, opened 2020-04-22
bump re-frame and remove 16ms delay for reagent tick

### LINKED ISSUE #10378: [perfromance] Investigate reagent tick
Reagent doesn't support react native , so it uses fake function for next tick `(js/setTimeout 16)` and in combination with `fx/merge` when we have one big event , it doesn't work good, probably we should fork reagent library and use  `(js/setTimeout 0)`because currently we just waste 16ms

image

flexsurfer:
fixes #10378

we used our own re-frame fork only because of next-tick, so I bumped re-frame to latest version

removed reagent 16ms delay

BEFORE:
image
AFTER:
image


cammellos (commented):
Does anyone know why we were doing this? maybe @rasom remembers?

cammellos (approved):

flexsurfer:
if you're asking why we forked re-frame and used setInterval instead nextTick , @yenda did this, but I believe its a right solution because setInterval in RN is the quickest way to run the new task in JS thread that's what we need, and nextTick works differently in RN

flexsurfer:
as for reagent 16ms , we just didn't know about this

rasom (approved):

PR #10405, opened 2020-04-22
Handle when message confirmations arrive out of order.

### LINKED ISSUE #10389: Sometimes messages-replies don't change state to "Sent" (endless spinner)
# Bug Report

## Problem

The issue is mostly visible from restored accounts (however sometimes can be reproduced and from freshly created account).
If send first message as reply - it may freeze with endless spinner (reproducibility ~90% with restored accounts) - and ~30-40% of the next messages-replies don't change state.
 
#### Expected behavior
messages are sent

#### Actual behavior
photo_2020-04-21 16 18 41 jpeg 2020-04-21 16-19-02

### Reproduction

The easiest way to reproduce for me:
- Open Status
- Restore account from seed phrase
- Join any public channel
- Select any message > Reply > Send reply

But it can be also reproduced from any account at attempt to send several reply-messages.

### Additional Information

- Status version: release v1.2 and v1.3

#### Logs

[status_logs-1.zip](https://github.com/status-im/status-react/files/4510011/status_logs-1.zip)

cammellos:
Fixes: #10389 

In some cases message confirmations might arrive out of order, before
status-go calls the callback for a sent message.

This commit changes the behavior so that confirmations out of order are
not ignored and they are checked when the callback for sending message
is called.

Please test with replies and non reply messages. It might fix the issue we had with confirmations in general (it might also be partially due to this).

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -813,15 +813,12 @@
 (handlers/register-handler-fx
  :transport/message-sent
  (fn [cofx [_ response messages-count]]
-   (let [{:keys [localChatId id messageType]} (-> response :messages first)]
-     (fx/merge cofx
-               (handle-update response)
-               (transport.message/set-message-envelope-hash localChatId id messageType messages-count)))))
-
-(handlers/register-handler-fx
- :transport/contact-message-sent
- (fn [cofx [_ chat-id envelope-hash]]
-   (transport.message/set-contact-message-envelope-hash cofx chat-id envelope-hash)))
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

Serhy:
Can not reproduce https://github.com/status-im/status-react/issues/10389 issue with replies / basic text messages / emoji messages/ replies to emoji in 1-1/public/group chat

Waiting e2e tests...

PR #10410, opened 2020-04-22
fix Gradle plugin by downgrading to 3.5.3

jakubgs:
The upgrade to `3.6.0` in #10387 caused issues which is why it was reverted in #10402.

Here I'm going one version lower to `3.5.3` and it seems to work fine. This confirms that the breaking change is in `3.6.0` specifically and I can work on that separately.


cammellos (approved):
we should try out the builds first , just in case

jakubgs:
I ran a manual release build here and tested it myself:
https://ci.status.im/job/status-react/job/combined/job/mobile-android/12575/
But we can definitely have a tester verify this in this PR too.

churik:
Can install, recover account and create fresh account on IOS13 and Android10.
Suppose it is enough for this PR.

PR #10413, opened 2020-04-22
fix: display the value to sign as a bytes32 hex string if the data is 66 chars long

### LINKED ISSUE #8590: Hexadecimal values not shown on "Sign Message" screen
# Problem

Hexadecimal values aren't being shown in the sign screen. It either displays as an empty box, or attempts to convert these to a string value.

## Steps to reproduce:
1. Browse https://status-im.github.io/dapp
2. Select the tab [Transactions]
3. Replace `Test message` by a random hex string,  i.e. `0x1234` or a bytes 32: `0xf38e51f05d0e9620018bf3281449f72804aafed68c19ae089380d6b25ae305a8`
4. Press [Sign message]
<p align="center">

</p>

## Expected behavior:
Hexadecimal value should be displayed.

## Additional Information
- Status version: 0.13.0
- Operating System: Android

richard-ramos:
fixes #8590

### Summary

If the value to sign is a bytes32 hexadecimal string, like `0x1122334455667788112233445566778811223344556677881122334455667788`, it will be displayed as such in the Signature screen, otherwise, it will do the transformation to a UTF8 String as usual.

Screenshot from 2020-04-22 13-54-06

Screenshot from 2020-04-22 13-53-23

In the issue I had mentioned before that any hexadecimal string should be displayed as such, but after reviewing this behavior in other web3 browsers, t**he hexadecimal string is shown only for bytes32 values.**

##### Functional

- dapps / app browsing

### Steps to test
- Open Status
- Open Dapp Browser
- Go to https://status-im.github.io/dapp
- Select the tab "Transaction"
- Change the message to sign, either use a normal string or a bytes32 value.
- Click on "Sign message"

status: ready

### DIFF
--- a/src/status_im/ethereum/core.cljs
+++ b/src/status_im/ethereum/core.cljs
@@ -138,3 +138,13 @@
                          nil)]
     (when normalized-key
       (subs (sha3 normalized-key) 26))))
+
+(def ^:const bytes32-length 66) ; length of '0x' + 64 hex values. (a 32bytes value has 64 nibbles)
+
+(defn hex->text
+  "Converts a hexstring to UTF8 text. If the data received is 32 bytes long, 
+   return the value unconverted"
+  [data]
+  (if (= bytes32-length (count (normalized-hex data)))
+    data ; Assume it's a bytes32
+    (hex-to-utf8 data)))
--- a/src/status_im/signing/core.cljs
+++ b/src/status_im/signing/core.cljs
@@ -195,7 +195,7 @@
                    :signing/sign {:type           (cond pinless? :pinless
                                                         keycard-multiaccount? :keycard
                                                         :else :password)
-                                  :formatted-data (if typed? (types/json->clj data) (ethereum/hex-to-utf8 data))
+                                  :formatted-data (if typed? (types/json->clj data) (ethereum/hex->text data))
                                   :keycard-step (when pinless? :connect)})}
        (when pinless?
          (signing.keycard/hash-message {:data data


src/status_im/signing/core.cljs
```diff
@@ -195,7 +195,9 @@
                    :signing/sign {:type           (cond pinless? :pinless
                                                         keycard-multiaccount? :keycard
                                                         :else :password)
-                                  :formatted-data (if typed? (types/json->clj data) (ethereum/hex-to-utf8 data))
+                                  :formatted-data (if typed? (types/json->clj data) (if (= 66 (count data))
```
rasom: @richard-ramos can you please extract `66` to some const and give it a name? We could also extract the whole `if` form to a separate fn

flexsurfer: and please write some comments in the code why 66

richard-ramos: Done!

src/status_im/ethereum/core.cljs
```diff
@@ -138,3 +138,13 @@
                          nil)]
     (when normalized-key
       (subs (sha3 normalized-key) 26))))
+
+(def ^:const bytes32-length 66) ; length of '0x' + 64 hex values. (a 32bytes value has 64 nibbles)
```
flexsurfer, line 142: but what if text message has a 66 length?

richard-ramos: It will be shown as a hex string. I don't think it's possible to identify if a hex value should be shown as a string or as a hexadecimal string (EIP712 fixes that with typed data). This assumes that it's more common that a 32 bytes hex string will be represented as such and not as a string value. 

For comparison with Metamask, that implements this same behavior, I tried executing this instruction - Signing a string with 32 characters - and it's represented as a hex value:

```
web3.eth.personal.sign("ABCDEFGHABCDEFGHABCDEFGHABCDEFGH", web3.eth.defaultAccount);
```

Screenshot from 2020-04-23 10-37-36

However, just removing/adding one character will make the string be displayed as text

Screenshot from 2020-04-23 10-37-17
Screenshot from 2020-04-23 10-39-02

flexsurfer: that's weird, mhm dunno what to do

flexsurfer (approved):

PR #10419, opened 2020-04-23
Add animated headers to profiles

Ferossgp:


errorists:
Tested on iOS, works great dude! 🙏  if possible I would ask to remove the translateY animation from the smaller top bar state: it's a nice touch but happens too slow and the animation pulls my attention away from the scrolled content. let's just stick to opacity only

src/quo/components/animated_header.cljs
```diff
@@ -10,7 +10,7 @@
 
 (defn header-wrapper-style [{:keys [value offset]}]
   (merge
-   {:background-color :white}
+   {:background-color (:ui-background @colors/theme)}
```
flexsurfer, line 13: could you elaborate on why you're using deref here?

flexsurfer (approved):

src/quo/components/animated_header.cljs
Ferossgp: I've made colors which are dynamic in Quo as an atom, so it would be as a reminder that this value is dynamic and should be used only in functions (called during render)

churik:
Works great on IOS 13.2.3. and Android 10.

churik:
@Ferossgp 
in this PR builds on the screen (when other user profile is opened):
Sauce Labs | Test Details: 2020-05-04 15-21-53
back button can't be found by accessibility-id 'back-button'

Can you please add back back button? :)

Ferossgp:
@churik rebased and added accessibility

churik:
Good to go!

PR #10420, opened 2020-04-23
Nix refactor of Android SDK derivation

jakubgs:
Refactoring Android SDK Nix derivation in preparation for a refactor of `status-go` derivations.

Since the SDK is used both in Android app build process and `status-go` build process I wanted to avoid importing it and passing it around via args explicitly. To do that I added our custom Android SDK+NDK to the `pkgs` which are passed around via `callPackage` implicitly.

I also did the same for `xcodeWrapper` by defining it as part of `pkgs` in `nix/pkgs.nix`.
And I removed some useless configuration for [Buck](https://buck.build/) which we don't use anymore.

And upgraded Nix from `2.3.2` to `2.3.4`.


cammellos (approved):

Serhy:
@jakubgs Android builds gives me the crash right after app launch (confirmed with Android 8 real device and Android 10 (emulator) with this PR build and now with develop:

Screenshot 2020-04-24 at 18 01 41

jakubgs:
Damn.

PR #10422, opened 2020-04-23
Show ENS name only when confirmed

### LINKED ISSUE #9912: Do not show ENS name in profile unless Tx is confirmed
# Problem

If my ENS Tx fails, it seems to get stuck in a pending state on my profile, at least for a given session. It would be better that we don't display it on the profile until the Tx is mined.

Another user can register the name while I'm waiting because my Tx has, in fact, failed. 

### Current state
**Success??**

**Wait a minute... (What I think is happening based on ENS screen)**

**What actually happened**

### Preferred implementation

- No ENS name on my profile and I still see `ENS usernames` as my entry point
- Ideally, we 1) add a pending state for the list usernames. It should indicate that a name is pending somewhere on this page. 
- 2) We add a state for a failed transaction. 

**This screen can stay like this until names are confirmed** 

**This screen should have a pending and failed Tx state, ideally**

jrainville:
### Summary

Fixes #9912

The problem was that the ENS name was added to the list straight when the TX was sent, but the TX could then error and you'd have no idea, plus the name would still display.

Instead, we register the name in a list with the TX hash and watch for changes in transactions, when the TX completes (error or success), we show it in the ENS page.

The look of it is not super pretty, I did with what I know, but feel free to tell me how to improve the look and the code style.

In progress:
image

Success:
image

### Review notes

One thing that needs to be improved, and I don't know how, is that the state is only updated when we go on the page itself.

Also, I don't think it's related to my PR, but sometimes, the `address` that is gotten from `:multiaccount/current-account` is not correct and it is not even part of my mulriaccount. I think it's also what causes the "unknown address" issue when sending a TX.

#### Platforms

I only tested in the Android Simulator and in Ropsten.

The profile is impacted. When registering the name, you need to go back to the ENS page (or wait on it) for the name on the profile to finally appear.

### Steps to test
Register a new ENS name

status: ready


src/status_im/ens/core.cljs
```diff
@@ -29,6 +31,26 @@
...
+ (fn [{:keys [db]} [_ new-state username custom-domain? tx-hash]]
+   {:db (assoc-in db [:ens/registrations tx-hash] {:state new-state
+                                                   :username username
+                                                   :custom-domain? custom-domain?})}))
+
+(handlers/register-handler-fx
+ :update-ens-tx-state-and-redirect
+ (fn [cofx [_ new-state username custom-domain? tx-hash]]
+   (fx/merge cofx
+             (re-frame/dispatch [:update-ens-tx-state new-state username custom-domain? tx-hash])
```
cammellos: This is dispatching straight away, `re-frame/dispatch` would immediately dispatch as called (is a side-effect).
We use `fx` to run side effects, so in your case it would simply be:
```
{:dispatch-n  [[:update-ens-tx-state new-state username custom-domain? tx-hash]
                     [::redirect-to-ens-summary]]}
```

src/status_im/ens/core.cljs
```diff
@@ -29,6 +31,26 @@
...
+             (re-frame/dispatch [::redirect-to-ens-summary]))))
+
+(handlers/register-handler-fx
+ :clear-ens-registration
+ (fn [{:keys [db]} [_ tx-hash]]
+   (let [registrations (get db :ens/registrations)]
```
cammellos: you can also use
`(:ens/registrations db)`
it's a bit more idiomatic, but `(get db ...)` is also fine

There's also actually `(db :ens/registrations)` :) which is the fastest but we don't use it so often, both `keyword` and `map` implement `IFn` https://clojure.github.io/clojure/javadoc/clojure/lang/IFn.html

src/status_im/ui/screens/ens/views.cljs
```diff
@@ -626,9 +627,48 @@
              [name-item {:name name :hide-chevron? true :action action}]]
             [radio/radio (= name preferred-name)]]]))]]]])
 
-(defn- registered [names {:keys [preferred-name public-key name] :as account} show?]
+(views/defview in-progress-registrations [registrations address]
+  (views/letsubs [transactions [:wallet/transactions address]]
```
cammellos: I think this needs to be rethinked a bit, we avoid side effects in views

cammellos (changes requested):
Nice work!
There's a bit to change I think, the view with `doseq` and `dispatches` probably we want to refactor, maybe we can go through it and find a way to move the side effects outside the view?

src/status_im/ens/core.cljs
cammellos: you can also do the whole thing in one go
`{:db (update db :ens/registrations dissoc tx-hash)}`

https://clojuredocs.org/clojure.core/update

src/status_im/ens/core.cljs
```diff
@@ -29,6 +31,26 @@
     username
     (stateofus/subdomain username)))
 
+(handlers/register-handler-fx
```
flexsurfer: handlers/register-handler-fx is deprecated, please use `fx/defn`

jrainville:
I fixed and updated. Thanks @cammellos and @flexsurfer

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -237,6 +236,30 @@
...
+    (doseq [[hash {:keys [state username custom-domain?]}] registrations]
+      (when (or (= state :dismissed) (= state :submitted))
+        (doseq [transaction transfers]
+          (let [transaction-hash (get transaction :hash)
+                type (get transaction :type)
+                transaction-success (get transaction :transfer)]
+            (when (= hash transaction-hash)
+              ; TODO Return from the loop once we find a match
+              (when (= transaction-success true)
+                (re-frame/dispatch [:update-ens-tx-state :success username custom-domain? hash])
```
flexsurfer: here are two things, events handlers MUST be pure, `re-frame/dispatch` is not pure, so you should use `:dispatch` fx, but we have internal rule to do not use `:dispath` , so you should use fx/merge here and instead `re-frame/dispatch` use `(update-ens-tx-state cofx ...)`

flexsurfer (changes requested):
events

rachelhamlin:
Awesome job @jrainville! This is a really important fix.

src/status_im/ethereum/transactions/core.cljs
cammellos: We are getting there, nice work! This is not the easiest of the tasks to start with :)

So, now we basically have an effectful handler (btw any time you see `doseq` in clojure you know that is used for side effects).

We want to turn this into a pure function, as per what Mr Surfer said.

The way we do this in re-frame is instead of actually run the side-effects in the handler (i.e `(re-frame/dispatch [])`, we return a map with the side effects to run.

So for example if you want to make an HTTP request, you would return from the event-handler some like:

`{:db {....} :make-http-request [:param]}`

And then register an handler:

```
(re-frame/reg-fx :make-http-request
  (fn [params] ...))
```

https://github.com/day8/re-frame/blob/master/docs/Effects.md

In this case you are running side-effects only in order to `dispatch` other events, and run their event handlers.
As a general rule, we avoid having having event-handlers dispatching if they don't have to, and we instead accumulate the results of other event-handlers. To give you an example:

Say you have:

```
(fx/defn handle-a
  {:events [::event-a]}
  [cofx]
  (do-stuff-a))
```

and 

```
(fx/defn handle-b
 {:events [::event-b]}
  [cofx]
 (do-stuff-b))
```

Which reacts to`::event-a` and `::event-b`.

Then you have a new event `::event-c` which needs to run the same logic as both `::event-a` and `::event-b`.

One way you could do this (and this is what we did before) is to have something like this:


```
(fx/defn handle-c
  {:events [::event-c]}
 [cofx]
   {:db (modify-db-for-c cofx)
     :dispatch-n [[::event-a] [::event-b]]})
```

So essentially the handler of `::event-c` is further dispatching two events, similarly to the function you have above (the difference is that here is a pure function, and we run the side-effects in an handler which is provided by re-frame).

Eventually we moved to a different pattern, instead of further dispatching, we call the handlers directly and cumulate the changes, so we have something like:

```
(fx/defn handle-c
  {:events [::event-c]}
 [cofx]
  (fx/merge cofx
                  {:db (modify-db-for-c cofx)
                  (handle-a)
                  (handle-b)))
```

Which is equivalent (ish) in terms of code. This has some benefits, is more explict, easier to test, easier to follow the logic, and some drawbacks, handlers take longer to run (so perfomance might be an issue, as they don't yeild control for rendering),  they couple logic more heavily and adds complexity in accumlating them (`fx/merge` is our own macro to merge effects).
Regadless, most time for now we use this strategy.

So in your case, I would start by pulling all the objects you are interested in:
instead of:
```
 (let [registrations (get db :ens/registrations)]
    (doseq [[hash {:keys [state username custom-domain?]}] registrations]
      (when (or (= state :dismissed) (= state :submitted))
```

you can filter them, and already get those that match the transaction-hash in transfers:

```
 (let [set-of-transactions-hash (reduce (fn [acc {:keys [hash]}] (conj acc hash)) #{} transfers) 
         registrations (filter #(This function should match against the set, submitted and dismissed) (get db :ens/registrations)]
```
Which should give the "actionable" transactions (i.e those that match `transfers` and are `dismissed` or `submitted`.

The it's the tricky bit, you want to comulate the fxs.
How we do this is to use the form:

`(apply fx/merge cofx [vector of fxs])`

`apply` in clojure works similarly to `apply` in javascripts, minus the binding as that's not a thing in clojure, but they are effectively the same.

so what you want to do now, is to have a function that handles a single transaction and instead of dispatching calls the right handler:
```
(fx/defn whatever [cofx transaction] ;; This is an actionable transaction
  (cond 
     (:transaction-success transactions) ;; Is this two conditions mutually exclusive or a transaction (:transaction-success transactions) can be true an :failed ? 
     (fx/merge cofx (ens/update-ens-transaction-state..) (ens/save-username))
     (= :failed (:type transaction)
          (ens/update-ens-transacation-state ...))))
```

And finally you can comulate those and apply them, so roughly your code woudl be something like:


``` 
(let [set-of-transactions-hash (build-a-set-of-transactions-for-fast-loopkup) 
         registrations (filter-only-those-you-are-interested-in)
         fxs (map wathever registrations)]
        (apply fx/merge cofx fxs))
```

And that should do the tricky, there are other issues that might crop up, depending on how the code is structured. It would also be useful to add a few tests for this function, as it would be pure so you can easily make sure what you get in the db is what you expect.

This is very tricky stuff actually, very brave choice to start with! and sorry if I am too verbose or there's stuff you already know, let me know if you have any question or something does not make sense.

flexsurfer: Yeah i couldn't do it better than Mr. Cammellos, thanks

jrainville: @cammellos WOW! What a great explanation. Don't worry about being too verbose everything you said there is useful to me.

Also, it's not exactly my first issue tackled in react-status, but it's definitely the first that is that big haha.

Thanks again. I'm gonna fix it up!

jrainville:
FYI, I just rebased on top of develop, but I wasn't able to test the changes yet, since my testing environment had to be reinstalled.

errorists:
@jrainville cool! regarding the UI, you can take a browse[ through these designs here](https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=1828%3A20099), the pending state for a name is indicated by changing the subtitle prop of the list item to say 'Transaction pending…' 

Screenshot 2020-05-08 at 09 44 38

adjacent there you'll find all the states necessary

churik:
@jrainville can you please rebase it one more time?
@cammellos @flexsurfer can you please re-review changes?

jrainville:
Rebased.

Should I do the design changes as @errorists commented in this PR or another one?

hesterbruikman:
I'd say if you can manage the design changes by Friday, please include it in this PR, else let me know and I'll create a separate issue with your name on it;)

jrainville:
I'll start working on it in another branch, but if I finish before Friday, I'll put it back in this one. I'll let you know.

jrainville:
Alright, I already finished implementing the new design. It looks just like Figma/the screenshot.
Clicking on it does nothing.
When it ends in a success, it is removed and goes at the bottom of the list like before.
When it errors, it stays at the top, but says that there was an error. Clicking on it then dismisses the message.

flexsurfer:
@jrainville great job, could you please fix linting
```
src/status_im/ens/core.cljs:5:14: warning: namespace status-im.utils.handlers is required but never used

[2020-05-20T20:34:04.737Z] src/status_im/ens/core.cljs:20:14: error: unresolved symbol status-im.utils.types

[2020-05-20T20:34:04.737Z] src/status_im/ui/screens/ens/views.cljs:629:57: warning: unused binding address

[2020-05-20T20:34:04.737Z] src/status_im/ui/screens/ens/views.cljs:631:39: warning: unused binding custom-domain?

[2020-05-20T20:34:04.737Z] src/status_im/ui/screens/ens/views.cljs:642:68: warning: unused binding name
```

jrainville:
Linting errors fixed and squashed

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -234,6 +234,39 @@
...
+                       nil)))
+              registrations)
+        ]
+    (apply fx/merge cofx fxs)))
+
+(fx/defn new-transfers
+  {:events [::new-transfers]}
+  [cofx transfers params]
+  (fx/merge cofx
+            (handle-mew-transfer transfers params)
```
flexsurfer:
```suggestion
            (handle-new-transfer transfers params)
```

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -205,8 +206,7 @@
   [{:keys [db]} address]
   {:db (assoc-in db [:wallet :fetching address :all-fetched?] true)})
 
-(fx/defn new-transfers
-  {:events [::new-transfers]}
+(fx/defn handle-mew-transfer
```
flexsurfer:
```suggestion
(fx/defn handle-new-transfer
```

src/status_im/ens/core.cljs
```diff
@@ -22,6 +22,24 @@
...
+  {:events [:update-ens-tx-state]}
+  [{:keys [db]} new-state username custom-domain? tx-hash]
+  {:db (assoc-in db [:ens/registrations tx-hash] {:state          new-state
+                                                  :username       username
+                                                  :custom-domain? custom-domain?})})
+
+(fx/defn update-ens-tx-state-and-redirect
+  {:events [:update-ens-tx-state-and-redirect]}
+  [_ new-state username custom-domain? tx-hash]
+  {:dispatch-n [[:update-ens-tx-state new-state username custom-domain? tx-hash]
```
flexsurfer: here handlers should be used instead events 
```clojure
(fx/merge cofx
   update-ens-tx-state
redirect-to-ens-summary
```

src/status_im/ui/screens/ens/views.cljs
```diff
@@ -624,7 +626,20 @@
              [name-item {:name name :hide-chevron? true :action action}]]
             [radio/radio (= name preferred-name)]]]))]]]])
 
-(defn- registered [names {:keys [preferred-name public-key] :as account} _]
+(views/defview in-progress-registrations [registrations]
+  [react/view {:style {:margin-top 8}}
+   (for [[hash {:keys [state username]}] registrations]
+     (when-not (or (= state :dismissed) (= state :success))
```
flexsurfer:
```suggestion
   (for [[hash {:keys [state username]}] registrations
            :when (or (= state :dismissed) (= state :success))]
```

src/status_im/ui/screens/ens/views.cljs
```diff
@@ -624,7 +626,20 @@
...
+       ^{:key hash}
+       [name-item {:name username
+                   :action (when-not (= state :submitted)
+                             #(re-frame/dispatch [:clear-ens-registration hash]))
+                   :subtitle (case state
+                                   :submitted (i18n/label :t/ens-registration-in-progress)
+                                   :failure (i18n/label :t/ens-registration-failure)
+                                   nil)}]))])
```
flexsurfer:
```suggestion
                                   nil)}])])
```

src/status_im/ethereum/transactions/core.cljs
```diff
@@ -234,6 +234,39 @@
...
+                            (contains? set-of-transactions-hash hash)))
+                        (get db :ens/registrations))
+        fxs (map (fn [[hash {:keys [username custom-domain?]}]]
+                   (let [transfer (first (filter (fn [transfer] (let [transfer-hash (get transfer :hash)] (= transfer-hash hash))) transfers))
+                         type (get transfer :type)
+                         transaction-success (get transfer :transfer)]
+                     (cond
+                       (= transaction-success true)
+                       (fx/merge cofx
+                         (status-im.ens.core/clear-ens-registration hash)
```
flexsurfer: you can use an alias here, for example `ens.core`

flexsurfer (changes requested):

jrainville:
@flexsurfer fixed in https://github.com/status-im/status-react/pull/10422/commits/c1dd922e5d105b14dccaebff45804e7791f2c5c0
I'll squash once I have your approval

flexsurfer (approved):

Serhy:
@jrainville thanks for fix! 
There is a regression appeared: on both iOS and Androids I can't assign existing name - nothing happens when I click `->` (Next) button when trying to bind ENS name which already assigned to address:
It's all fine in latest nightly develop for the 1st of June 

**To reproduce:** 
1) Open Status and recover the account with ENS name (e.g. you can use this one `fashion quality know robust copy neck stand embody entry task orient suggest`)
2) From Profile tap `ENS names` -> `Get Started`
3) Type `statuse2e`
4) Tap `->` blue button to proceed
**Actual result:** Button displayed as active but nothing happens when I press it. No errors in logcat too. 

ezgif com-video-to-gif


When I want to register new ENS then it works fine (and `->` blue button is actionable in that case)

jrainville:
@Serhy I just tested the latest develop and the bug happens there too. So I think the bug is in my PR only because I rebased on top of the regression.
My branch is based on top off c3d14f16645e6736ee9380cf2e7894df01759866, so I assume the regression happens there or before

Serhy:
@jrainville hmm, that's possible, yes. I was tested develop based on where it was all fine 9a114128bc74cd1fc9701eb8e3d8f562d6f735b9 and just tested latest develop https://github.com/status-im/status-react/commit/9d7691312be1f2427c2513cb1e9d680e68013880 which is also looks good to me. 

Could you rebase the very latest develop (and fix conflict) please?

jrainville:
@Serhy you were right, the issue was indeed in my PR. Somehow, while I tested develop, I probably didn't reset the app correctly.

Anyway, I rebased and fixed the issue. You can test again to confirm.

Serhy:
Looks good to me, @jrainville thanks! Tested on iOS13 and Android 9.
While transaction is pending no ENS assigned by default in chat and can't be selected from 'Primary username' list. Same after transaction fails. 
Can assign existing ENS and can add new ENS as well.

cammellos (approved):

PR #10436, opened 2020-04-27
fix: make token icons consistent

### LINKED ISSUE #9832: Make sure that everywhere the asset image is done properly
When discussing this pull request https://github.com/status-im/status-react/pull/9772 we noticed that the asset image is not done properly and consistently across the app. 

For instance, on the `Send transaction` bottom sheet the asset image has no border, although on the `Wallet` main screen it has.

Any asset image across the app should follow these rules:
- It should be a circle with a square .png image inside;
- It should have a #000000 1px **inner border** (important, not an outer border, most likely means that there should be nested objects) with 10% opacity;

Here's a more visual explanation of why that matters:

Screen Shot 2020-01-15 at 15 43 01

emizzle:
Fixes #9832

### Summary
Ensure token icons are consistent across the app, by adding a 1px 10% opacity #000000 inner border to all token icons.

#### Platforms
- all

- Wallet screens

status: ready

### DIFF
--- a/src/status_im/ui/components/list/styles.cljs
+++ b/src/status_im/ui/components/list/styles.cljs
@@ -39,6 +39,11 @@
   {:width         image-size
    :height        image-size})
 
+(def item-token-icon
+  {:border-radius 50
+   :border-color  "rgba(0,0,0,0.1)"
+   :border-width  1})
+
 (def big-item-image
   {:width         image-size
    :height        image-size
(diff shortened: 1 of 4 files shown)


simonricoo:
@emizzle I'm working on a new set of icons and those will not need an inner boarder. Sorry, it's our fault we should have removed this issue when we decided to work on the new set. I will follow up with a new PR with new icons. Can you please revoke this fix as no border will be needed? CC @errorists

emizzle:
Closed due to new icons (see comment above).

PR #10438, opened 2020-04-27
fix(multiaccounts/login): ensure log-level is updated correctly

### LINKED ISSUE #10127: Log level is not applied
# Bug Report

## Problem

By default we set log level in DISABLED state in profile setting. However changing it to DEBUG/TRACE does not make any effect. In logcat (Status.log for iOS) there are no DEBUG/TRACE records anyway. 

It's fine with develop (nightly/PR) builds.

#### Expected behavior

Log level setting applies and reflected in logs for release builds

#### Actual behavior

Log level setting change from Profile is not reflected in logs for release builds

### Reproduction

- Instal release build
- Create new account
- Navigate to `Profile` ->  `Advanced` -> `Log level`
- Set TRACE value
- Re-login to account, navigate between different screens and check logcat has DEBUG/TRACE/INFO records (also sending bug report from `Profile` -> `Need help?` -> `Submit bug`  to your email should has Status.log in package with the similar results: DEBUG/TRACE logs should be there) 

### Additional Information
- Status version: Release 1.1 (and Release 1.0.2)

#### Logs

Logcat with TRACE log level setting in 1.1.x release build:

[logcatTRACErelease.log](https://github.com/status-im/status-react/files/4288656/logcatTRACErelease.log)

0x-r4bbit:
As reported in #10127 there's an issue that the log level isn't properly
applied when changed through log settings UI.

Turned out that we're setting the log-level only once in the entire
lifetime of the application and that is during bootstrap.
When the log level is changed through the UI, we never actually update
the log level in the logger module.

To fix this, we just need to ensure that, whenever the log-level is changed,
we also call `log/set-level!` with the new level.

It should also be noted that, every time the log level is changed through
the UI, the user is logged out, so one place to update the logger is
after very successful login.

Fixes #10127

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -25,6 +25,7 @@
             [status-im.utils.handlers :as handlers]
             [status-im.utils.identicon :as identicon]
             [status-im.utils.keychain.core :as keychain]
+            [status-im.utils.logging.core :as logging]
             [status-im.utils.platform :as platform]
             [status-im.utils.security :as security]
             [status-im.utils.types :as types]
@@ -210,6 +211,7 @@
               (contact/initialize-contacts)
               (stickers/init-stickers-packs)
               (mobile-network/on-network-status-change)
+              (logging/set-log-level multiaccount)
               (multiaccounts/switch-preview-privacy-mode-flag))))
 
 (defn get-new-auth-method [auth-method save-password?]
@@ -270,6 +272,7 @@
                                              :mailserver-topics {}
                                              :default-mailserver true})
               (multiaccounts/switch-preview-privacy-mode-flag)
+              (logging/set-log-level multiaccount)
               (when-not platform/desktop?
                 (initialize-wallet accounts nil)))))
 
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -487,8 +487,7 @@
  :log-level/current-log-level
  :<- [:multiaccount]
  (fn [multiaccount]
-   (or (get multiaccount :log-level)
-       config/log-level-status-go)))
+   (get multiaccount :log-level)))
 
 (re-frame/reg-sub
  :waku/enabled
--- a/src/status_im/utils/logging/core.cljs
+++ b/src/status_im/utils/logging/core.cljs
@@ -42,6 +42,21 @@
     (get-js-logs)
     #(re-frame/dispatch [callback-handler %]))))
 
+(re-frame/reg-fx
+ :logs/set-level
+ (fn [log-level]
+   (log/set-level! log-level)))
+
+(fx/defn set-log-level
+  [{:keys [db]}  multiaccount]
+  (let [log-level (if-let [level (get multiaccount :log-level)]
+                    (if (clojure.string/blank? level) "ERROR" level)
+                    config/log-level-status-go)]
+    {:db (assoc-in db [:multiaccount :log-level] log-level)
+     :logs/set-level (-> log-level
+                         clojure.string/lower-case
+                         keyword)}))
+
 (fx/defn send-logs
   [{:keys [db]}]
   ;; TODO: Add message explaining db export


src/status_im/multiaccounts/login/core.cljs
```diff
@@ -190,6 +191,11 @@
         (data-store.settings/rpc->settings settings)
         multiaccount (dissoc settings :networks/current-network :networks/networks)
         network-id (str (get-in networks [current-network :config :NetworkId]))]
+    (log/set-level! (if-let [log-level (get multiaccount :log-level)]
+                      (-> (if (string/blank? log-level) "ERROR" log-level)
+                          string/lower-case
+                          keyword)
+                      config/log-level))
```
0x-r4bbit: Alright, couple things to note here:

- When logging in, there's a possibility that the account in question doesn't have an actual `log-level` yet. In that case, we're going with `config/log-level` which is also what's being set during app bootstrap
- If there is a log-level, there's a possibility that it's an empty string. This is the case when logging was set to "DISABLED". I couldn't find a way to actually disable the logger so I'm defaulting to "ERROR" here

src/status_im/subs.cljs
```diff
@@ -484,7 +484,9 @@
  :<- [:multiaccount]
  (fn [multiaccount]
    (or (get multiaccount :log-level)
-       config/log-level-status-go)))
+       (-> config/log-level
+           name
+           string/upper-case))))
```
0x-r4bbit: The log-settings view has a subscription to this function here and previously tried to read from `:multiaccount :log-level`. If there's no log-level in the current multi account, it would fallback to`status-go-log-level`.

^ This is an issue because it would always result in the UI showing a different log level than what is actually configured. That's why I decided to fallback to `config/log-level` instead because that's the same value that's set during app bootstrap.

0x-r4bbit (commented):
Just tested this again and there's still an issue (would like some feedback on that):

- Generally it tends to be hard to keep `multiaccount :log-level `and `log/log-level` in sync
- When logging in for the first time, the log-level is set to `INFO` in the app state, that's why we're hooking into `get-settings-callback` to update `log/log-level`
- That callback is only called when logging in from an existing account though. When creating a brand new account, it just happened that the log-level in app state is also set to `INFO` but I can't find the place where that happens.
- The latter, again, results in `log/log-level` and app state log-level to be out of sync.

I find that managing state of log-level is quite inconsistent and was wondering if there's a way to subscribe to changes in `multiaccount :log-level` and set `log/set-log-level!` based on that as a reaction.

@yenda said that subscriptions are only possible within views, but maybe there's another reframe-y way to do this?

src/status_im/multiaccounts/login/core.cljs
yenda: Since this is a side-effect, you should create a set-level! effect in logging.core with reg-event-fx and use that effect here.
Also as a default you should probably use `config/log-level-status-go` and assoc the log-level in the db here so that you don't need to change the subscription. You simplify the subscription by simply returning `(get multiaccount :log-level)` since it will then always have a log-level.
Would also be nice to put a default log-level for newly created accounts since it seems it is not the case otherwise you would always have a log-level here.

src/status_im/subs.cljs
```diff
@@ -483,8 +483,7 @@
  :log-level/current-log-level
  :<- [:multiaccount]
  (fn [multiaccount]
-   (or (get multiaccount :log-level)
-       config/log-level-status-go)))
+   (get multiaccount :log-level)))
```
0x-r4bbit, line 490: `multiaccount :log-level` is now always set because it'll always be initialised during login

src/status_im/utils/logging/core.cljs
```diff
@@ -41,6 +42,21 @@
...
+(re-frame/reg-fx
+ :logs/set-level
+ (fn [log-level]
+   (log/set-level! log-level)))
+
+(fx/defn set-log-level
+  [{:keys [db]}  multiaccount]
+  (let [log-level (if-let [level (get multiaccount :log-level)]
+                    (if (clojure.string/blank? level) "ERROR" level)
+                    config/log-level-status-go)]
```
0x-r4bbit, line 54: I'm pretty sure there's a "nicer" way of doing this.

0x-r4bbit (commented):
@yenda thanks for all the feedback and help. This should now do the trick!

src/status_im/utils/logging/core.cljs
flexsurfer: @PascalPrecht could you please elaborate why if its nil we set log-level-status-go?

flexsurfer:
@PascalPrecht could you please resolve conflicts, thanks

src/status_im/utils/logging/core.cljs
0x-r4bbit: @flexsurfer Yes sir! This was requested here: https://github.com/status-im/status-react/pull/10438#discussion_r415635013

flexsurfer: i knew that, thanks :)

0x-r4bbit:
@flexsurfer done!

flexsurfer (approved):

churik:
Checked on Android and IOS:
- log level by default
- setting to DEBUG, TRACE, disabling.

Looks good to me!

PR #10441, opened 2020-04-27
[#10418] Can login to account just via tap upon it from `Your keys` view

### LINKED ISSUE #10418: Can login to account just via tap upon it from `Your keys` view
# Bug Report

## Problem

Issue came from user report: 
At one point it may happen that user able to login to some of accounts without entering the password 
but just tapping upon multiaccount item on 'Your keys' view. 
It may be connected to navigation or the way we sort accounts on 'Your keys' screen after curtain (9 items in user case, 10 in my case) number of accounts created.
See steps to reproduce

#### Expected behavior

If I am on `Your keys` view - there is no way I can loginn to any account without just entering password (but the actual root cause of this particular issue might be different - like account sorting on "Your keys" view after 9+ accounts created in Status) 
#### Actual behavior

Can login to acccount via just tap it's item from the list of 'Your keys' view.

123213d

### Reproduction

1)  Open Status
2) Create new account -> Close Status -> Tap `<` back at top-left button
3) Repeat 2nd item 7 times more
**Actual result:** You should end up having 8 accounts. Notice that each time when you open Status after account creating accounts sorted in order they were created (latest create is the first in list on `Your keys` and the very first acc is at the bottom)

4) Create one more account.
**Actual result:** You should end up having 9 accounts. However, that last account created is not the  
first one in `You keys` 
5) Login to the second account from the top of the `Your keys` screen with `Save password` option
6) Close Status
7) Open Status
**Actual result != Expected result:** You appear on the login view of the account which was (and keep staying) very first in the list we had in 4th step
8) Tap `<` back button to navigate to list of accounts (`Your keys` screen)
9) Tap upon the account you logged in in the 5th item
**Actual result:** You logged into this account right away. Appear on Chats view.
10)_(Optional to show issue further)_ If you login with `Save password` option (or FaceID or TouchID) into any but first account (basically repeat 5th-8th steps for third-fourth.... accs) you will be able to log in to this account just via tap. If you login to the first account then as expected (until forced LogOut) you will autologin to this account after Open-Close Status)

### Additional Information
- Status version: Release1.3/Develop

flexsurfer:
fixes #10418

### DIFF
--- a/src/status_im/init/core.cljs
+++ b/src/status_im/init/core.cljs
@@ -35,9 +35,10 @@
 (fx/defn initialize-views
   {:events [::initialize-view]}
   [cofx {:keys [logout?]}]
-  (let [{{:multiaccounts/keys [multiaccounts] :as db} :db} cofx]
+  (let [{{:multiaccounts/keys [multiaccounts]} :db} cofx]
     (when (and (seq multiaccounts) (not logout?))
-      (let [{:keys [key-uid public-key photo-path name]} (first (#(sort-by :last-sign-in > %) (vals multiaccounts)))]
+      (let [{:keys [key-uid public-key photo-path name]}
+            (first (sort-by :timestamp > (vals multiaccounts)))]
         (multiaccounts.login/open-login cofx key-uid photo-path name public-key)))))
 
 (fx/defn initialize-multiaccounts


vitvly (approved):

Serhy:
Can not reproduce #10418 on Android and iOS (for iOS I upgraded build where bug reproduced). 
Thanks @flexsurfer !

PR #10442, opened 2020-04-27
fix: carousel image height on narrow devices

### LINKED ISSUE #10427: First carrousel image onboarding is cut off
# Problem
In dark mode it becomes visible that the walkie-talkie is cut off at the bottom.

Screenshot_20200424-112146_Status PR

## Implementation
Requires investigation as to why the image is cut off.
Specification of image size, position and padding in [Figma](https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=4540%3A11631)

## Acceptance Criteria
Full image is shown

## Notes
cc @errorists for review

emizzle:
Fixes #10427

### Summary
Narrow devices were causing text in the intro carousel to grow to two lines on *the first* slide but not others, causing the height to be recalculated incorrectly on the last slide (that only had one line of text).

Create a vector of atoms that can individually track the heights required for each slide independently.

#### Screenshots during testing
iPhone SE (2nd gen)
iPhone SE (2nd gen)
iPhone 11
iPhone 11
Nexus One
Nexus One
Pixel 2
Pixel 2

- Intro screen

status: ready

### DIFF
--- a/src/status_im/ui/screens/intro/views.cljs
+++ b/src/status_im/ui/screens/intro/views.cljs
@@ -34,6 +34,9 @@
         scroll-view-ref (atom nil)
         width (r/atom 0)
         height (r/atom 0)
+        text-height (r/atom 0)
+        text-temp-height (atom 0)
+        text-temp-timer (atom nil)
         bottom-margin (if (> window-height 600) 32 16)]
     (fn []
       [react/view {:style {:align-items :center
@@ -50,8 +53,7 @@
                            :pinch-gesture-enabled false
                            :on-scroll #(let [^js x (.-nativeEvent.contentOffset.x ^js %)]
                                          (reset! scroll-x x))
-                           :style {;:width @width
-                                   :margin-bottom bottom-margin}}
+                           :style {:margin-bottom bottom-margin}}
         (doall
          (for [s slides]
            ^{:key (:title s)}
@@ -60,17 +62,28 @@
                                 :justify-content :flex-end
                                 :align-items :center
                                 :padding-horizontal 32}}
-            (let [size (min @width @height) #_(- (min @width @height) #_(* 2 margin))]
+            (let [size (min @width @height)]
               [react/view {:style {:flex 1}
                            :on-layout (fn [^js e]
-                                        (reset! height (-> e .-nativeEvent .-layout .-height)))}
+                                        (let [new-height (-> e .-nativeEvent .-layout .-height)]
+                                          (swap! height #(if (pos? %) (min % new-height) new-height))))}
                [react/image {:source (:image s)
                              :resize-mode :contain
                              :style {:width size
                                      :height size}}]])
             [react/i18n-text {:style styles/wizard-title :key (:title s)}]
-            [react/i18n-text {:style styles/wizard-text
-                              :key   (:text s)}]]))]
+            [react/text {:style (styles/wizard-text @text-height)
+                         :on-layout
+                         (fn [^js e]
+                           (let [new-height (-> e .-nativeEvent .-layout .-height)]
+                             (when (and (not= new-height @text-temp-height)
+                                        (not (zero? new-height))
+                                        (< new-height 200))
+                               (swap! text-temp-height #(if (pos? %) (max % new-height) new-height))
+                               (when @text-temp-timer (js/clearTimeout @text-temp-timer))
+                               (reset! text-temp-timer
+                                       (js/setTimeout #(reset! text-height @text-temp-height) 500)))))}
+             (i18n/label (:text s))]]))]
        (let [selected (hash-set (quot (int @scroll-x) (int @width)))]
          [dots-selector {:selected selected :n (count slides)
                          :color colors/blue}])])))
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/intro/views.cljs
```diff
@@ -62,7 +62,8 @@
                                 :justify-content :flex-end
                                 :align-items :center
                                 :padding-horizontal 32}}
-            (let [margin 32
+            (let [height (:atom (first (filterv #(= (:key %) (:title s)) heights)))
```
flexsurfer: im not sure if we want to set different sizes for images @errorists  wdyt? my suggestion is just to use lower height for all . something like
```clojure
(let [new-height (-> e .-nativeEvent .-layout .-height)]
     (swap! height #(if (pos? %) (min % new-height) new-height))
```
so you don't need `heights` ,only one atom as before

flexsurfer (changes requested):

src/status_im/ui/screens/intro/views.cljs
errorists: different sizes for images? like you mean, in a single carousel on a single device, they could have different dimensions? oh for sure **not**!!

emizzle: @flexsurfer @errorists, so I had this initially and it means that there is a large gap in those slides with only one line of text. The issue is that the first slide has multiple lines of title text and slides 2 and 3 *do not*. So what ends up happening is the minimum image height is calculated for the image in slide 1 and is used for the images in slides 2 and 3, which means there ends up being a large gap between the image and the title in slides 2 and 3. Are you sure this is what you want?

flexsurfer: ping @errorists

errorists: there's already a larger gap between the image and title in the design. 
☝️ image aligns to the top, 
👇 Text floats to the bottom, 
with the space between them having at least 16pt in height, otherwise we start to scale the image down

Screenshot 2020-04-28 at 10 35 42

flexsurfer: > Are you sure this is what you want?

looks like yes :)

emizzle: OK, all updated. Here's what it looks like on a narrow device:

flexsurfer (approved):

Serhy:
test_long_press_delete_clear_all_dapps test already fixed and in develop but by some reason is not picked up. 
Not this PR fault! Looks good to me on Galaxy Note 4, iPhoneX, Samsung J7

flexsurfer:
@errorists could you take a look, please?

flexsurfer:
or @hesterbruikman

hesterbruikman:
Sorry to be a pain on this @emizzle 😕❤️  They're the first screens and deserve the attention. Would like to ask @errorists to take another look on Tuesday. Wondering if we want to have the following:
- Image scaled and float to the top (as now)
- Header float to the bottom
- Body float to the top

Getting this effect now on a small screen, where headers seem out of sync across screens

Screenshot_20200501-161922_Status PR

errorists:
I believe that's okay! 
I mean would it be even possible to position the header at the same exact location without some absolute positioning, which I don't really want us to do here?
I think the ideal here is :
🖼️ Container with image aligned to the top
🔠 Container with header, all share the same height, text is aligned to the bottom of container
🔡 Container with subtitle, all share the same height, text is aligned to the top

Android - 2

flexsurfer:
@errorists done

flexsurfer:
@Serhy @hesterbruikman @errorists could someone take a look and approve ?

errorists:
I tested and can only confirm the layout looks good on a regular phone, don't have anything small enough to test. @hesterbruikman 👀  ?

src/status_im/ui/screens/intro/views.cljs
```diff
@@ -62,18 +64,28 @@
...
+                         :on-layout
+                         (fn [^js e]
+                           (let [new-height (-> e .-nativeEvent .-layout .-height)]
+                             (when (and (not= new-height @text-temp-height)
+                                        (not (zero? new-height))
+                                        (< new-height 200))
+                               (swap! text-temp-height #(if (pos? %) (max % new-height) new-height))
+                               (when @text-temp-timer (js/clearTimeout @text-temp-timer))
+                               (reset! text-temp-timer
+                                       (js/setTimeout #(reset! text-height @text-temp-height) 500)))))}
```
emizzle, line 85: Why the need to set a timeout here?

flexsurfer: because on-layout is called several times , we want to use the last one, so we use "debounce" with timer

churik:
Caurousel looks fine on IPhone SE, IPhone 7, iPhone 11.
Failed e2e are not related to this PR.

flexsurfer:
can't merge because of `PR does not allow 'edits from maintainers', so it will be kept open`

flexsurfer:
merged

PR #10463, opened 2020-04-29
Removed check spec

flexsurfer:
removed check spec, and moved db, events and navigation ns from ui to root


cammellos (approved):

src/status_im/navigation.cljs
```diff
@@ -75,34 +34,56 @@
    (log/debug :navigate-replace view-id params)
    (navigation/navigate-replace (name view-id) params)))
 
-(handlers/register-handler-fx
- :navigate-to
- navigation-interceptors
- (fn [cofx [_ & [go-to-view-id screen-params]]]
-   (navigate-to-cofx cofx go-to-view-id screen-params)))
+(defmulti unload-data!
```
Ferossgp: I think this interceptor could be removed as it is not used.

rasom (approved):

.gitignore
```diff
@@ -191,3 +191,4 @@ status-modules/resources
 trace.edn
 
 app
+project.clj
```
dijonkitchen, line 194: Should this be gitignored?

PR #10498, opened 2020-05-01
React native 62

Ferossgp:
`NOTE:` Flipper is disabled by default on iOs to not create problems during development


flexsurfer (approved):

ios/Podfile
```diff
@@ -1,14 +1,51 @@
-# Uncomment the next line to define a global platform for your project
-# platform :ios, '9.0'
+platform :ios, '9.0'
```
andremedeiros, line 1: Is this intended?

Ferossgp: I used to follow the rn template, as the next version will come to drop ios 9 and less https://github.com/facebook/react-native/releases/tag/v0.63.0-rc.1 and I expect changes here

mobile/js_files/package.json
```diff
@@ -22,13 +22,12 @@
     "emojilib": "^2.4.0",
     "eth-phishing-detect": "^1.1.13",
     "functional-red-black-tree": "^1.0.1",
-    "hermes-engine": "0.2.1",
```
jakubgs, line 25: Is hermes now pulled in by `react-native` or what?

Ferossgp: Yes, they are dependent on each other and if they are not in sync the android build does not work. It was hardcoded here because when react-native 60 was using a version with a bug and it was overridden to fix it. More info there: https://github.com/status-im/status-react/pull/9001

Ferossgp: Also new version available, and will come with next react-native version https://github.com/facebook/hermes/releases/tag/v0.5.0

jakubgs: Sweet, thanks for explaining.

android/gradle.properties
```diff
@@ -46,3 +46,6 @@ ANDROID_ABI_INCLUDE=armeabi-v7a;arm64-v8a;x86
 org.gradle.jvmargs=-Xmx8704M
 
 versionCode=9999
+
+# Flipper
+FLIPPER_VERSION=0.33.1
```
jakubgs, line 51: Can we maybe explain what flipper is for in the comment?

churik:
### Issue 1: Public key is not copied from user profile 
### Issue 2: Confirming from keyboard on emulator doesn't redirect to chat

Test results (with command highlighted): 
https://app.saucelabs.com/tests/d613ade299614766ac7a1415d7fff3f0?auth=d1f3829aa958fe79e4ef35f20fafbb0e#77
POST appium/device/press_keycode
{ "keycode": 66 }
On nightly on this action user is redirected to 1-1 chat, on this PR nothing happens

Ferossgp:
@churik  I just tested manually, and both work. Also tested the keycode with:
`adb shell input keyevent 66`. 
I have some assumptions about why the test may fail.
1. The event 66 makes sense only if the focus is on the input, maybe that after paste the focus is lost?
2. Text is not copied cause there is a long press on it instead of a simple press. In code, it awaits just for on-press.

I can try to check my assumptions and fix e2e but I do not know how to run just these failed tests.

churik:
Test-run: https://ethstatus.testrail.net/index.php?/runs/view/5336&group_by=cases:section_id&group_order=asc&group_id=736 
Failures are the same as in nightly - so no pr-related failures

churik:
@Ferossgp 
### Issue1: Cannot call a class as a function when scrolling on account view
photo_2020-05-06 16 01 34
Steps:
- create new multiaccount
- open wallet > manage assets > add several different assets (>5)
- open Status account
- scroll

[Status-debug-logs-5.zip](https://github.com/status-im/status-react/files/4587213/Status-debug-logs-5.zip)

churik:
Tested on Android 10, IOS 11, IOS 12, IOS 13 - exploratory testing with testing values in input fields, context menus, scrolling, fetching old history, also keycard brief testing.
Found https://github.com/status-im/status-react/issues/10591 which is already in develop.

@Ferossgp please mind my commit when squashing.

PR #10561, opened 2020-05-04
fastlane: disable uploading of Play Store metadata

jakubgs:
Apparently we want to let marketing to edit the app description by hand and not to override that.


andremedeiros (approved):

jakubgs:
Superseded by #10564.

PR #10573, opened 2020-05-05
[#10414] Implement support for latest version of eip-1193

### LINKED ISSUE #10414: Implement support for latest version of eip-1193
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md#request

flexsurfer:
fixes #10414


resources/js/provider.js
```diff
@@ -1,242 +1,249 @@
...
-}
-
-function qrCodeResponse(data, callback){
-    var result = data.data;
-    var regex = new RegExp(callback.regex);
-    if (!result) {
-        if (callback.reject) {
-            callback.reject(new Error("Cancelled"));
-        }
+(function(){
```
vitvly, line 1: @flexsurfer a timely move! :)

I have a question though - why do we need to wrap in `(function() {...})()`. I understand that this is because we don't want the variables to pollute global namespace, correct?

vitvly (approved):

resources/js/provider.js
flexsurfer: correct

Serhy:
@flexsurfer could you please check what's wrong with `simpledapp.eth` under Ropsten? It does not see the wallet after permissions allowed. 

1) Install Status and create new account
2) Switch to Ropsten network
3) Open simpledapp.eth in Browser and check Accounts are with values

resources/js/provider.js
flexsurfer: @siphiuel i also have a question, why does it work locally and doesn't in PR build ? :)

flexsurfer: ok , its because of this `msg (str "ReactNativeWebView.onMessage('"` but why does it work locally, have no idea

vitvly: Ahh, right, as you're now in function scope, you have to prepend `window.`.

flexsurfer:
@Serhy should be fixed now

Serhy:
I've tested with iOS13, Anndroid 8.1 and Android 9 several Daps (Ddex.io, cryptokitties, dragonereum, defyzap, zerion, sablier) and seen no bugs.
Also updated simpledapp in e2e test.
Thanks you @flexsurfer !!

flexsurfer:
@siphiuel so you can change the spec, eip1193 fully supported and request method as well

PR #10589, opened 2020-05-06
Move icons paths

Ferossgp:
Follows the same approach as tokens and images.
There are some icons which are not used anywhere in the app, I have not removed them, but most probably we should review that. Now that we have a map with all icons we could statically analyze which icons are needed and remove others.


jakubgs:
Okay, I merged https://github.com/status-im/status-react-jenkins/pull/5 and tagged the commit in `master` as `v1.1.4`, please update all the `Jenkinsfile`s in `ci` folder in your PR that use the `library` directive.

flexsurfer:
@yenda @Ferossgp so now bundle will be even bigger and will crash on all phones? :)

Ferossgp:
@flexsurfer no, it will be smaller cause all DPI icons that are not used are removed, so the number of image files in the bundle decreased by ±400

flexsurfer (approved):

Serhy:
Looks good to me with iOS13 and Android 8.1 (i've looked through main views to confirm no crashes and icons are on place)
Thank you @Ferossgp !

PR #10590, opened 2020-05-06
Align top recipient multiline input

Ferossgp:
Android by default align it in the middle


flexsurfer:
could you elaborate on why middle is bad?

errorists:
@flexsurfer this is following a chat we had in design discord, this looks like a bug and it's wrong, multiline inputs should only be allowed to align text to top

image

flexsurfer (approved):

PR #10595, opened 2020-05-06
Status is no longer "Alpha" software

cyanlemons:
Simple README fix.


hesterbruikman (approved):
Good catch! Thanks @cyanlemons

PR #10603, opened 2020-05-07
Require translations

yenda:
require translations inline so that they aren't  part of the bundle


jakubgs (approved):
I like it. It should make JS bundle smaller and also stop the awful huge dumps during Gradle builds.

Serhy:
Tested language change (Russian, Chinese, Korean, English) with iOS13 and Android8.1 without issues.

PR #10609, opened 2020-05-08
IMAGES  [DO NOT MERGE]

### LINKED ISSUE #10474: Send images > 4. Implement UI image message sending and display
# Problem
As part of epic Send images, discussed in https://github.com/status-im/specs/issues/69:

* [ ] 1. Write spec doc based on [this thread](https://github.com/status-im/specs/issues/69) #10417
* [ ] 2. Implementation image content type for chat messages https://github.com/status-im/specs/pull/107 
* [ ] 3. Implement image compression for message type (status-go) - OPTIONAL
* [ ] **4. Message sending and display :point_left:**
* [ ] 5. UI for capture and image selection https://github.com/status-im/status-react/pull/9455
* [ ] 6. Message actions and reply #10475 
* [ ] 7. View images in user profile for 1:1 chats

## Implementation
Image display follows [Figma ](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=4190%3A24181)
> Separate message, each image container aka 'message bubble' can only have a single image inside, we shouldn't mix text and images, those would be split into two separate messages on the list : one for image, another for text

Frame 52

# Acceptance Criteria

* Sending/Sent/Error states of image sending are displayed
* Image can be viewed full screen on tap
* Implementation follows Figma design

## Notes
Testing along with #9455 

## Future Steps

* [ ] 6. Message actions and reply

### LINKED ISSUE #10475: Send images > 6. Message actions and reply
# Problem
As part of epic Send images, discussed in status-im/specs#69:

* [ ] 1. Write spec doc based on [this thread](https://github.com/status-im/specs/issues/69) #10417
* [ ] 2. Implementation image content type for chat messages status-im/specs#107
* [ ] 3. Implement image compression for message type (status-go) - OPTIONAL
* [ ] 4. Message sending and display #10474
* [ ] 5. UI for capture and image selection #9455
* [ ] **6. Message actions and reply :point_left:**
* [ ] 7. View images in user profile for 1:1 chats

## Implementation
Follow [Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=4190%3A25193) design

Frame 53

- - -
Frame 54

## Acceptance Criteria

* Implementation follows Figma
* User can reply to image
* User can copy image and paste in text input field

## Future Steps

1. View images in user profile for 1:1 chats

flexsurfer:
Original PR: https://github.com/status-im/status-react/pull/9455

fixes: #10474
fixes: #10475

* [x] Send images > 4. Implement UI image message sending and display #10474
* [x] Send images > 6. Message actions and reply #10475

BEFORE MERGE

* [x] disable images for public chats
* [x] update status-go version
* [x] test compatibility on staging fleet


errorists:
👏 Amazing job!! 👏 

Issues I found so far, will edit if I find more

- [x] images seem to have bottom and top? margins, they aren't aligned the sam way regular chat messages do
IMG_1503


- [x] the images menu should share it's height with stickers panel (we want to get the keyboard height and match that, we do that with stickers), for calculating layout in image grid:
- Each image is width/height is: Total panel height - 8 (margin between images) / 2
- Photos button is constant height, and camera preview is flex remaining space, width = image width
- [x] would it be possible to increase the amount of images we show before the menu ends? Currently it's 7, I'm thinking going to 10-12 would be an improvement. Clean / recycle them from memory as you scroll or smth, designer here, don't know this stuff ;)
- [x] increase the touchable size for the (x) icon, should be a 44 by 44.

Screenshot 2020-05-21 at 17 17 18

- [x] font-weight: 500 on the 'Photos' and 'Close' buttons

- [ ] permission alerts don't appear above the chat view like you'd expect, instead they have a dedicated view with white background as if the app was suspended
- [ ] the camera view lacks focusing controls and lens switch, is that available in the package we use or should we find another one? I would rank this a must have
- [ ] if I copy an image it pastes a text link inside the input, instead of pasting the image

Future improvements to be followed in a separate issue / PR
• animate the camera and detailed image view so the thumbnail scales up to fill the full screen view
• image pinch to zoom/pan, flick to close
• queueing multiple images, multi selection

flexsurfer:
@errorists fixed all except 

- [ ] permission alerts don't appear above the chat view like you'd expect, instead they have a dedicated view with white background as if the app was suspended
this one because of screenshots options 

- [ ] the camera view lacks focusing controls and lens switch, is that available in the package we use or should we find another one? I would rank this a must have
not sure about  lens switch, but focusing controls  could be implemented

- [ ] if I copy an image it pastes a text link inside the input, instead of pasting the image
yeah because actually its a text :) but we could do something about that

errorists:
@flexsurfer cool, thanks! 💪   
so what do you think about the focus control on camera module (tap to focus)? The camera lens control, that's a nice to have and only available on some phones. Do you want to do focus in a separate PR?

> yeah because actually its a text :) but we could do something about that

would be sweet, it's a much better UX for sending images copied from browser / other chats

edit: also tap to focus helps with QR code recognition when for some reason your phone can't focus on the code

flexsurfer:
for some reason focus on touch doesn't work, so i would address it in separate PR

flexsurfer:
because we have only base64 string for image, disabled copy and share for now

flexsurfer:
so we need to create an issue for copy and share

flexsurfer:
added focus on tap, for qr codes as well

errorists:
@flexsurfer I tested latest available build (#14) and see no tap to focus, also there seems to be no flashlight control available? 
Another thing I just noticed, we ask for camera and photos permissions consecutively, which should not be the case. Only gallery permission should be asked upon opening the image menu the first time. 
For camera permission you need to tap the camera button first
Screenshot 2020-05-25 at 14 31 32 
and only once you allowed them, can you proceed to the camera viewfinder.

---

IMG_1509

Also here it should just say `Image` instead of `Update to the…`, [Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=7723%3A87)

---

The close icon still appears to be positioned incorrectly inside the touchable, should be centred both vertically and horizontally

Screenshot 2020-05-25 at 14 39 39

flexsurfer:
just tested on ios and android phones and tap to focus is working, no flashlight yes, could be implemented in separate PR

errorists:
@flexsurfer weird, which build are you testing? I just retried the latest one and no, tap to focus doesn't work on my phone: when I tap on any portion of the camera viewfinder nothing happens: I have no visual feedback that camera refocused and focus region didn't change. It's easiest to see on macro shots, bring your camera close to a foreground object and tap on the background.

flexsurfer:
@errorists there is no visual feedback yes, only focus is changing, so it might be two things , 1 - it might work differently on different phones, for example on android its more noticeable, 2 - probably it needs some tuning in numbers, because how it works, when you tap we get coordinates and pass them to focus function and phone tries to focus on that area

errorists:
@flexsurfer thanks. so that's not acceptable, it either works as it normally does elsewhere or not. I tried to get it working like crazy on a bunch of differently composed photos: it doesn't work at all. It works as expected in Discord tho and any other camera module I tried.

errorists:
Tried it on a Samsung S8+ Android phone, tap to focus doesn't work there too

flexsurfer:
> Another thing I just noticed, we ask for camera and photos permissions consecutively

yes that's true, on ios if you use camera it asks permissions by itself, we can change this not showing camera in image panel, honestly its kinda useless feature or we need to implement permissions natively and check if user granted permissions if not do not show camera

flexsurfer:
so we have two leftovers
copy and share
do not ask permissions 
all other should be fixed now

flexsurfer:
reworked buttons, so we have only one leftover: copy and share

errorists:
adding one more to list of bugs, if I tap on the white empty area to the left/right of image it still registers as pressing the image

flexsurfer:
fixed

errorists (approved):

src/status_im/chat/models/images.cljs
```diff
@@ -0,0 +1,120 @@
...
+            [status-im.utils.image-processing :as image-processing]
+            [taoensso.timbre :as log]
+            [clojure.string :as string]
+            [status-im.utils.platform :as platform]))
+
+(defn- resize-and-call [uri cb]
+  (react/image-get-size
+   uri
+   (fn [width height]
+     (let [resize? (> (max width height) 2000)]
```
cammellos: maybe a `def` of the `2000` value?

src/status_im/chat/models/images.cljs
```diff
@@ -0,0 +1,120 @@
...
+       (image-processing/resize
+        uri
+        (if resize? 2000 width)
+        (if resize? 2000 height)
+        60
+        (fn [resized-image]
+          (let [path (aget resized-image "path")
+                path (if (string/starts-with? path "file") path (str "file://" path))]
```
cammellos, line 26: We could live a comment explaining why it behaves unpredictably at times

src/status_im/chat/models/input.cljs
```diff
@@ -112,8 +115,19 @@
...
-  [{:keys [db] :as cofx} {:keys [hash pack]} current-chat-id]
+(fx/defn send-image
+  [{{:keys [current-chat-id] :as db} :db :as cofx}]
+  (let [image-path (get-in db [:chats current-chat-id :metadata :sending-image :uri])]
+    (fx/merge cofx
+              {:db (update-in db [:chats current-chat-id :metadata] dissoc :sending-image)}
+              (when-not (string/blank? image-path)
+                (chat.message/send-message {:chat-id      current-chat-id
+                                            :content-type constants/content-type-image
+                                            :image-path   (string/replace image-path #"file://" "")
```
cammellos, line 126: Do we need the `file://` at all? otherwise we could strip it as soon as it's created.

src/status_im/ipfs/core.cljs
```diff
@@ -31,18 +31,18 @@
 
 (fx/defn add
   "Add `value` on ipfs, and returns its b58 encoded CID"
```
cammellos, line 33: This changes are not needed probably, we can keep them if useful, but they might have been left over of the ipfs implementation.

src/status_im/ui/screens/chat/sheets.cljs
```diff
@@ -210,3 +210,49 @@
...
+                                   (hide-sheet-and-dispatch  [:chat.ui/show-profile from]))}])
+       [list-item/list-item
+        {:theme    :action
+         :title    :t/message-reply
+         :icon     :main-icons/reply
+         :on-press #(do
+                      (when from-preview?
+                        (re-frame/dispatch [:navigate-back]))
+                      (hide-sheet-and-dispatch [:chat.ui/reply-to-message message]))}]
+       #_[list-item/list-item
```
cammellos, line 240: We could leave a few comments on why they are commented out for now

cammellos (approved):
Left a few comments, overall looks good to me, we need to do some compatibility testing on staging, ideally before merging, so ping me when you think it's stable enough to be tested and I will add some description on what to test and deploy on staging the fleet.

churik:
TESTING IS STILL IN PROGRESS

## ISSUE 1: If video is processed, some random picture is attached

As we don't support attaching video files, better to disable them for selecting.
Currently, after "Processing" is finished, some random photo can be attached (not reproducible all the times, sometimes input field remaining empty) - but anyway it takes time of the user so better just show some warning (i.e. Not supported format) if it is not possible to disable processing.
cc @errorists 

## ISSUE 2: Sometimes message description was sent before image, sometimes - after
IMO better to have descriptions after image is sent, because in this case on home screen on the preview we have this description instead of just `Image` word
Monosnap 2020-05-27 16-26-21

## ISSUE 3: Images and stickers message don't have timestamps
I think it can be fixed later, but need to agree with @hesterbruikman @errorists

errorists:
@churik as always, thanks for testing 🙏  

re: issue1, it would be preferable if we didn't allow selecting videos at all. If that's not possible, then after tapping it we should return a tooltip error saying that videos aren't supported. Let me know please if that's something we need to do, I'll follow up with a design.

re: issue2 you're right, description should always appear below the image if you composed a message with both an image and text

re: issue3, I'm curious do people really find this an issue? I think timestamps are an anachronism and they're largely irrelevant with how people use chat in 2020, so I'm definitely cool with not showing them on visual messages.

edit: also, nice pictures!

flexsurfer:
not sure if #2 can be fixed in this PR, but will take a look at #1 videos should be available for selection for sure

flexsurfer:
fixed 1
2 and 3 won't be addressed in this PR

errorists:
thanks, 3 is not an issue

Serhy:
Tested on Android 9, Android 8.1, iOS 13 on `eth.test` fleet.

Can not reproduce Issue 1 from above https://github.com/status-im/status-react/pull/10609#issuecomment-634717654 (can't add video file). Also there is no way to record video from camera mod. 

Can add JPG, JPEG, PNG, BMP images.

Can not add (nothing appear in snippet to send) corrupted image file or different file renamed to use supported image extension (like to rename .csv to .jpeg and send it). 

Save, Reply to image message works as well. 

Looking why that single test fails in a case to fix here in the branch (manually can still copy and paster message content). 

Otherwise it looks good to me!

flexsurfer:
@Serhy I've changed touchable area for message bubbles , might be a reason of failed e2e

cammellos:
@Serhy could you also test communications between release app and this PR? Both using the test fleet.

The behavior is that images should not really be received.
 
It's dependent on size, so small images might be received by release and will be shown as Unknown content type. Larger images won't be received at all.

I would try to send long messages from release, start group chats with many members (you just have to invite many, they don't need to accept the invitation).

Thanks!

Serhy:
@cammellos thank you a lot, I indeed forgot compatibility between release. Will test it.

Serhy:
With the `eth.test` fleet set in Release 1.3 and PR build I tested:
- big message (with 50k chars) could be sent in group chat with 10 members (and received by one participant). Works fine between Release 1.3 and this PR build
- images are sent to user with release 1.3 build are all displayed as `Unhandled content-type-7`. Same for image with 3KB size and 19MB size - both displayed like that:

Screenshot 2020-05-28 at 18 48 45

However on message preview (on Chats view) it has `Update to latest version to see a nice image`.

@flexsurfer Is it possible to have the same explanation on images in chat messages instead `Unhandled content-type-7`?

cammellos:
@Serhy thanks, regarding Unhandled content-type that's not possible, we fixed this in development and it will be released with 1.4, but 1.3 will still see that (1.4 won't come with images though)

Serhy:
Sounds good then (much less people will upgrade from 1.3 to 1.5) 
I didn't see messages that were not sent or lost messages between Release <-> PR in test fleet

flexsurfer:
@cammellos so sounds like we can finish this PR ? do you have PR in status-go we need to merge first to have new version ?

cammellos:
@flexsurfer yes, PR is ready, I can take over from here if this is done. We need to deploy the fleet to eth.prod. @Serhy would you like to run more tests against the fleet? should I deploy on staging first for some more tests?

Serhy:
On staging first please, @cammellos 
And could you please set default fleet to `eth.staging` (for .env.e2e is enough)

cammellos:
@Serhy done, staging deployed and PR is using staging fleet by default for e2e

Serhy:
@cammellos thanks, tested Release build with Staging fleet and e2e tests ran against PR build with staging fleet. 
Let's revert default staging fleet back here. And before merge, as I understand, - deploy fleet to eth.prod?

cammellos:
> Let's revert default staging fleet back here. And before merge, as I understand, - deploy fleet to eth.prod?

Yes, I will do that on monday, thanks!

PR #10612, opened 2020-05-08
Incoming outgoing bug

### LINKED ISSUE #10608: Transaction command status in 1-1 chat flipped after Accept/Decline transaction
# Bug Report

## Problem

When 1-1 command confirms or declines status then command indication about Incoming or Outgoing transaction changed to the it's opposite state

#### Expected behavior

Transaction command keep the same state after Accept/Decline
### Reproduction

1) UserA (with ETH) adds new UserB and they join 1-1 chat
2) UserA sends a `Send` command (`Outgoing transaction` on UserA side)
3) UserB receives it as `Incoming transaction`
4) UserB taps -> any of `Access and share address` OR `Decline`
**Actual result:** On UserA side and UserB side the command message changed status to the opposite one (so UserA who supposed to send transaction gets `Incoming transaction` with the `Sign and send` button and UserB sees on his side `Outgoing transaction`). And how it looks on UserB side you may see in above actual result gif

### Additional Information

- Status version: develop 8th of May

yenda:
fix  #10608

case only works with litterals or const, but const isn't consistent with clojure and should be avoided.

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -369,13 +369,16 @@
                "keycard-multiacc?" keycard-multiaccount?)
     (fx/merge cofx
               {:db (assoc db :auth-method auth-method)}
-              #(case auth-method
-                 keychain/auth-method-biometric
+              #(cond
+                 (= auth-method
+                    keychain/auth-method-biometric)
                  (biometric/biometric-auth %)
-                 keychain/auth-method-password
+                 (= auth-method
+                    keychain/auth-method-password)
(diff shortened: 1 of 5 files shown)


flexsurfer (approved):

rasom (approved):

yenda:
@Serhy ok to merge?

Serhy:
Yes, looks good to me. 
Transactions and FaceID (and autologin without FaceID) working on Android 8, Android 6, iOS13. Failed autotest manually looks good too (because I used higher gas price in the transaction to deploy contract)

PR #10624, opened 2020-05-11
fix android crash

flexsurfer:
fix android crash


jakubgs (approved):
Interesting, what caused this to change?

PR #10629, opened 2020-05-12
[#10434] Disable touch events during autologin

### LINKED ISSUE #10434: Account is freezing with endless spinner if tap on account from list  during sign in when autologin is enabled
# Bug Report

## Problem
During autologin before home view is shown, app redirects to list of accounts for several moments.
If during this time user click on this account from list - in case of keycard account is freezing (public chats history is not loaded, instead of + endless spinner is shown)

#### Expected behavior
Redirect to home view, no errors

#### Actual behavior
**Ordinary account:**
`Cannot Read property 'ToLowerCase' error` error, can close it and continue work with app
**Keycard account:**
- history of public chats is not loaded
- endless spinner is shown instead of +

#### Notes

Issue is not only about keycard, it can be reproduced with the ordinary account as well, but in case of ordinary account
1) It is difficult to reproduce (autologin is faster)
2) error is shown, but after that account is working as expected
In the case of keycard account history for public chats are not shown and the endless spinner is displayed instead of + button on Home view
ezgif com-video-to-gif (4)

### Reproduction

The easiest way to reproduce:
- Open Status
- Create account with keycard
- Enable "Don't use keycard to sign in"
- Close app
- Open app and when account list is shown, tap on keycard account

### Additional Information

- Status version: nightly 24/04/2020
- Operating System: Android

#### Logs

Ordinary account: 
[Status-debug-logs.zip](https://github.com/status-im/status-react/files/4529821/Status-debug-logs.zip)

Keycard account:
[double_login.log](https://github.com/status-im/status-react/files/4529832/double_login.log)

Would love to have feedback from @hesterbruikman - how better to improve autologin process (i.e. spinner) to avoid this issue

rasom:
fix #10434 

status: ready

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -375,11 +375,8 @@
                  (biometric/biometric-auth %)
                  (= auth-method
                     keychain/auth-method-password)
-                 (get-credentials % key-uid)
-
-                 ;;nil or "none" or "biometric-prepare"
-                 :else
-                 (open-login-callback % nil)))))
+                 (get-credentials % key-uid))
+              (open-login-callback nil))))
 
 (fx/defn biometric-auth-done
   {:events [:biometric-auth-done]}


src/status_im/ui/screens/views.cljs
```diff
@@ -97,26 +97,29 @@
...
-          [reloader/reload-view @reloader/cnt])
-        (when config/keycard-test-menu-enabled?
-          [hardwallet.test-menu/test-menu])]])}))
+      (let [disbale-touch-events?
+            @(re-frame/subscribe [:disable-touch-events?])]
+        [react/safe-area-provider
+         ^{:key (str @colors/theme @reloader/cnt)}
+         [react/view {:flex             1
+                      :background-color colors/black-persist
+                      :pointer-events disbale-touch-events?}
```
flexsurfer: shouldn't it be something like this `:pointer-events (if disbale-touch-events? 'none' 'auto')` ?

rasom: youre right, though for some reason it works with just true/false

rasom: weird

flexsurfer: interesting, but it might stop working in next RN versions

rasom: sure, fixed

flexsurfer (approved):

src/status_im/ui/screens/views.cljs
```diff
@@ -93,30 +93,36 @@
...
-        [wallet/request-transaction]
-        [wallet/select-account]
-        [signing/signing]
-        [bottom-sheet]
-        [popover/popover]
-        (when debug?
-          [reloader/reload-view @reloader/cnt])
-        (when config/keycard-test-menu-enabled?
-          [hardwallet.test-menu/test-menu])]])}))
+      (let [disbale-touch-events?
```
Ferossgp:
```suggestion
      (let [disable-touch-events?
```

src/status_im/ui/screens/views.cljs
```diff
@@ -93,30 +93,36 @@
...
+            @(re-frame/subscribe [:disable-touch-events?])]
+        [react/safe-area-provider
+         ^{:key (str @colors/theme @reloader/cnt)}
+         [react/view
+          {:flex             1
+           :background-color colors/black-persist
+           :pointer-events   (if disbale-touch-events?
```
Ferossgp:
```suggestion
           :pointer-events   (if disable-touch-events?
```

Ferossgp (commented):
This approach seems risky to me, it is quite easy to forget about this field. Also, there are many places that change it, this may lead to race conditions and disable touch for the whole app. Maybe it could be solved by changing optimistic render to something more pessimistic. Like showing a loader until we receive the auth method from the status-go? It will lead also to better UI, cause the user will not see the flicking multi-accounts screen for a moment. 
Or for the simplicity of this PR and fixing the issue, bio-auth could be fixed using always the right account. Because of the bio-auth approach to get the multi-account from DB, the bio-auth event is started from one account, and on press on another account in the multi-accounts screen changes the value for multiaccount in the app DB. So multi-account in DB does not relate to the one which has requested bio. Sending the multi-account as a parameter to the event instead of getting it from DB, will ensure that the event will be always executed on the right account.

churik:
@rasom 
the issue is still reproducible with the same steps on build https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200513-115755-b0e3dc-pr-universal.apk - the difference is that I tap several times -> and again Authentication required screen is shown for a user who is already logged in.

rasom:
@churik unfortunately I can't reproduce it so it's kind of hard to test, will check again

rasom:
@Ferossgp 
> This approach seems risky to me

Yeah I know, I'm limited in time at the moment, so either can do some simple stupid fix here or defer this task to someone else, mostly because I'm still partially blocked after shadow-cljs update and can't even reproduce problem properly.

churik:
Failed e2e tests are related to Ropsten load, so not related to this PR.

PR #10640, opened 2020-05-14
nix: refactor loading of node.js modules

jakubgs:
This is the long awaited refactor of how we include Gradle and Node.js dependencies for Android.
The main issue with how it was done before was in the __horrific__ `nix/mobile/android/maven-and-npm-deps/default.nix` file, which for some incomprehensible reason combined both Gradle and Node dependencies into one derivation - that also pulls files from repo - by copying them every time. This is not necessary and makes everything slower.

My change essentially makes the Android shell and release build use Gradle dependencies __directly__ from `nix/deps/gradle`, and loads the Node.js dependencies ALMOST directly from `nix/deps/gralde`, but puts them through a patching process here:
https://github.com/status-im/status-react/blob/5550d701e2c4acea5c84ea451e593ec41c7c9f19/nix/mobile/android/default.nix#L14-L16
Because multiple `react-native-*` modules have `build.gradle` files that need to be modified to not try to use remote repos, and instead use `localMaven()` repo, which we provide from Nix store.

__TODO__: There's one last thing that I might have to modify which is the `nix/mobile/reset-node_modules.sh` script which copies the patches node modules to repo root.

Changes:
- Drop `nix/mobile/android/maven-and-npm-deps/default.nix`
- Replace it with much simpler `nix/mobile/android/patchNodeModules`
- Move Gradle patching tool to `nix/pkgs/patch-maven-srcs`
- Move `nix/mobile/android/targets/release-android.nix` to `nix/mobile/android/release.nix`
- Simplify it by using `gradle.deps` and patched node modules separately
- Change `TARGET` for `release-android` to `default`
- Add support for more build input types in `mergeSh` tool
- Add a `status-go` shell in `nix/shells.nix`

Status: __READY__


jakubgs:
The only thing that I consider unfinished is this line:
https://github.com/status-im/status-react/blob/0781e205edf6c84f1ca5b74c28bf610ceff59b45/nix/scripts/node_modules.sh#L61
And where it should include an `-L` flag to follow symlinks and copy their contents.
I need to test this with at least one developer to see if it works with their usual flow.

flexsurfer (approved):

PR #10645, opened 2020-05-14
nix: store deps lists in json

jakubgs:
Data should be data.

@iurimatias made me aware of this, which is quite false:
status_react_lang_stats
So I fixed it.


flexsurfer (approved):

jakubgs:
@churik should I worry about `test_can_add_existing_ens`? Some weird text in the screenshot.

jakubgs:
Note to self to rebase after https://github.com/status-im/status-react/pull/10635 is merged.

PR #10674, opened 2020-05-18
Do not use signer arg for keycard_signTypedData

### LINKED ISSUE #10245: keycard_signTypedData(signer, message) should not take the signer argument
# Problem

The keycard_signTypedData(signer, message) has been modelled after eth_signTypedData(signer, message). In eth_signTypedData the signer is the account used to sign the message. However in keycard_signTypedData a Keycard is used for this purpose so it unclear what the signer argument is used for. Ideally, we would remove it, changing the function to keycard_signTypedData(message)

If the signer argument is actually used for something it should be clearly documented so that dApps know what they have to pass and why.

## Implementation

 The function should become keycard_signTypedData(message) and all references to signer should be removed.

vitvly:
Fixes #10245 

The signature for `keycard_signTypedData` is changed from
```
ethereum.send("keycard_signTypedData", [account, JSON.stringify(params(chainId))]);
```
to
```
ethereum.send("keycard_signTypedData", JSON.stringify(params(chainId)));
```

### DIFF
--- a/src/status_im/ui/screens/signing/views.cljs
+++ b/src/status_im/ui/screens/signing/views.cljs
@@ -148,7 +148,7 @@
                    :signing :t/try-keeping-the-card-still
                    :error :t/tap-card-again
                    :success :t/transaction-signed)]
-    [react/view (assoc styles/message :padding-vertical 16 :align-items :center)
+    [react/view (assoc (styles/message) :padding-vertical 16 :align-items :center)
      [react/view {:style {:align-self :flex-start :padding-left 16 :margin-bottom 24}}
       [react/text {:style {:font-size (if small-screen? 15 17) :font-weight "700"}}
        (i18n/label :t/confirmation-request)]]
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/signing/views.cljs
```diff
@@ -148,7 +148,7 @@
                    :signing :t/try-keeping-the-card-still
                    :error :t/tap-card-again
                    :success :t/transaction-signed)]
-    [react/view (assoc styles/message :padding-vertical 16 :align-items :center)
+    [react/view (assoc (styles/message) :padding-vertical 16 :align-items :center)
```
flexsurfer, line 151: was broken?

flexsurfer (approved):

src/status_im/ui/screens/signing/views.cljs
vitvly: Yes.

Serhy:
Can send transaction via keycard and without keycard bind to account, can sign message in Dapp too.

PR #10680, opened 2020-05-18
Quo Animated Button

Ferossgp:
After review will remove the previews

### DIFF
--- a/src/quo/components/text_input.cljs
+++ b/src/quo/components/text_input.cljs
@@ -92,6 +92,7 @@
           :margin              0
           :text-align-vertical :center
           :flex                1
+          :color               (:text-01 @colors/theme)
           :height              height}
          (when-not before
            {:padding-left (:base spacing/spacing)})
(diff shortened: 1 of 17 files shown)


src/quo/components/text_input.cljs
```diff
@@ -164,8 +165,7 @@
             (merge {:style                   (text-input-style multiline input-style before after)
                     :ref                     ref
                     :placeholder-text-color  (:text-02 @colors/theme)
-                    :color                   (:text-01 @colors/theme)
-                    :underline-color-android :transparent
+                    :underline-color-android "transparent"
```
flexsurfer: why string?

flexsurfer (approved):

src/quo/components/text_input.cljs
Ferossgp: The app was crashing on android, thought that problem may be here - but it was in another file where I was using `:true` instead of `true`. Will discard this line

PR #10686, opened 2020-05-19
Use interactive components from Quo

Ferossgp:
This PR introduces Quo components to the Status App. All components are introduced in one PR to enable animations to all components at once because the Gesture Handler was changed. Code that was not using components was refactored, thus UI regressions may be introduced. 

Components updated and refactored:
1. Toolbar
2. Buttons
3. Icon Buttons
4. Bottom sheet


.env.release
```diff
@@ -18,3 +18,4 @@ RPC_NETWORKS_ONLY=1
 PARTITIONED_TOPIC=0
 MOBILE_UI_FOR_DESKTOP=1
 HARDWALLET_ENABLED=1
+DISABLE_ROOT_ALERT=0
```
flexsurfer: probably better to name it ENABLE_ROOT_ALERT in that case it won't be enabled in release by mistake

src/status_im/ui/screens/browser/accounts.cljs
```diff
@@ -2,16 +2,16 @@
...
             [status-im.utils.utils :as utils]
             [re-frame.core :as re-frame]))
 
 (defn render-account [dapps-account]
   (fn [account]
-    [list-item/list-item
-     {:theme     :selectable
-      :selected? (= (:address dapps-account) (:address account))
+    [quo/list-item
+     {:accessory :radio
```
flexsurfer: whole item should be touchable

flexsurfer (approved):

errorists:
it's glorious but I'm unsure about haptic feedback on Android, on the Samsung Galaxy phone I tried it's absolutely godawful, like if being jolted with electricity when touching the screen. If that's the standard there's no way we should leave that enabled on Android.

oskarth:

Serhy:
@Ferossgp thanks a lot for huge work done here and for fixing regressions
Looks good to me now for merge based on tests i've made on Android 8/9 and iOS13.

PR #10689, opened 2020-05-19
[Fixes: #5124] Fix rtl flag

### LINKED ISSUE #5124: Detect right-to-left text direction
### User Story
As a developer I want to know when text is going to be displayed right-to-left so that i could adjust views and styles to display that properly.

### Description
For Arabic, Hebrew and other right-to-left languages React Native aligns text to the right. That is breaking layout for chat messages.

As a quickfix we're using regexp to detect most of right-to-left languages, but this method is error-prone and don't cover all the cases https://github.com/status-im/status-react/pull/5054

### Solution
I haven't found the API for Text or View components to tell if text is going to be displayed and aligned as RTL text. Probably there is some API for this on native level (android/ios), needs to be researched further.

cammellos:
rtl flag was using the wrong field name.

status: ready

fixes #5124

### DIFF
--- a/src/status_im/data_store/messages.cljs
+++ b/src/status_im/data_store/messages.cljs
@@ -35,7 +35,7 @@
                        :ens-name (:ensName message)
                        :line-count (:lineCount message)
                        :parsed-text (:parsedText message)
-                       :rtl (:rtl message)
+                       :rtl? (:rtl message)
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

churik:
Issue is fixed, failed e2e are not related to PR.

PR #10693, opened 2020-05-19
mentions

### LINKED ISSUE #10509: Mentions
* [ ] Parse `@0xpk` messages in Status Go
* [ ] Implement message styling in status-react
* [ ] Implement auto completion in status-react to allow ENS name and random name entry
* [ ] Show notification for mention

### References
https://github.com/status-im/specs/issues/67
https://github.com/status-im/status-react/issues/7117

yenda:
fix #10509

current implementation looks good enough performance wise, can be further improved by using more local state as well as react on-key-press in most cases, only changing the "completion?" value instead of recomputing everything


errorists:
issues

- [ ] input is broken, inserting a character deletes it first and places it again, this happens in a split second but is visible and super annoying. Concerns all text, not only mentions
- [ ] Input has a different height when there's no text inside (looks like it's lower by 1px) and we only show the placeholder than when there's text inside. this is accompanied by blinking of the screen when text is first inserted. 
- [ ] text inside input is offset a bit to the bottom, it doesn't look like it's aligned properly
- [ ] when I choose someone from the autocomplete list, the autocomplete persists, I need to choose the username _again_ to dismiss it, placing the username again in the input
- [ ] selecting someone from the autocomplete dropdown places the cursor after the `@` character and not at the end of username
- [ ] placing cursor inside a username, removes the mention text colouring for all text after the cursor
- [ ] autocomplete doesn't always work for me, example: I type `@b` and I still see the complete list of usernames beginning with all letters, not only `b`
- [ ] the autocomplete list is blinking annoyingly whenever it's summoned / dismissed / changes height. It should animate, please look at Discord it's also RN.
- [ ] messages with a mention are blocking me from long pressing the message: I need to long press it on the exterior parts of message bubble without text to fire up the event
- [ ] mentioning a username that's not in the chat (`@here`) leads to weird message formatting

IMG_1498 2

- [ ] replying to a message with a mention, in the reply menu the entire chat key is visible instead of the username

IMG_1501


---
things missing

- [ ] no `@` character in the system keyboard
- [ ] tapping a username inside a mentioned message should display that user profile

tested on iOS, 
@churik please test this one diligently, I'm sure there are many more issues hidden here 🙏

src/status_im/chat/models/input.cljs
```diff
@@ -23,11 +23,217 @@
...
+(defn tag-text-parts
+  "Takes a string, a map of users and a regexp of known mentions
+  and splits the string into parts each parts being a vector
+  with a part type:
+  - `text` for regular text
+  - `mention` for mentions, which will contain the user alias and public key
+  This is used to quickly render the input field with the proper styling"
+  [text users mention-regexp]
+  (if mention-regexp
+    (loop [[text mention & rest] (string/split text mention-regexp)
```
rasom, line 49: Why not just `reduce`?

src/status_im/chat/models/input.cljs
```diff
@@ -23,11 +23,217 @@
...
+                [:text (if previous-mention?
+                         (subs text 1)
+                         text)])
+          computed-text)))
+    [[:text text]]))
+
+(def end-mention-regexp
+  #"(@(?:(?:[A-Z][a-z]*(?:\s[A-Z]?[a-z]*)?(?:\s[A-Z]?[a-z]*)?)|(?:[a-zA-Z0-9\.]*))$)")
+
+(defn compute-all-text-parts
```
rasom, line 72: we need some tests for this function, it is huge and hard to reading

src/status_im/chat/models/input.cljs
```diff
@@ -23,11 +23,217 @@
...
+                                   0
+                                   (- (count text-before-cursor)
+                                      (count completing?))))
+        text-after-cursor (str (when completed? " ") text-after-cursor)
+        mentioned (cond-> mentioned
+                    completed? (conj mentioned completed?))
+        ;; prepare regexp for currently mentioned users
+        mention-regexp
+        (when (seq mentioned)
+          (re-pattern (str "(@" (string/replace
```
rasom, line 113: Can we extract magic values to vars?

Issue #10471, opened 2020-04-30
Markdown  formatting characters are visible on message preview on home view


churik:
# Bug Report

## Problem
On home:
![photo_2020-04-30 09 54 17](https://user-images.githubusercontent.com/4557972/80686189-a2f0be00-8ac8-11ea-9410-67488d01d563.jpeg)

On chat view:
![photo_2020-04-30 09 54 56](https://user-images.githubusercontent.com/4557972/80686212-aa17cc00-8ac8-11ea-985f-e29c9ce97289.jpeg)

#### Expected behavior
no markdown formatting characters on home view

#### Actual behavior
markdown formatting characters are visible

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Join to public chat
- Send any message with formatting 

### Additional Information
- Status version: nightly 30/04/2020
- Operating System: Android, iOS

PR #10700, opened 2020-05-22
[Fixes: #10471] Parse markdown in chat subheader.

cammellos:
This commit adds parsing of markdown in the chat subheader.

Because of the added complexity performance are impacted but I have
also noticed that on each loading of the chat screen we calculate
alias & identicon through status-go, so that has been changed so that
they are returned from status-go.

Overall performance is now roughly identical, tested loading 150
one-to-one chats.

Another issue that I've spotted is that some `subs` are unnecessarely
recalculated (`active-chats`), when unrelated fields changes. I will
address this in a separate PR that should improve performance.

status: ready (I need to test on a real device, but ready for review)


src/status_im/chat/db.cljs
```diff
@@ -29,15 +32,25 @@
         (assoc :inviter-name inviter-name)
         :always
         (assoc :chat-name (group-chat-name chat))))
-    (let [{contact-name :name :as contact}
+    (let [photo (if (seq identicon)
```
flexsurfer, line 35: why seq ?

flexsurfer (approved):

src/status_im/chat/db.cljs
cammellos: We could use `blank?` but we just want to check if it's not `""` or `nil`(i.e empty)

Serhy:
@cammellos 
Tested on Android 8.1 and iOS 13

**Issue 1** Both Android and iOS Empty space on chat preview when URL sent in the message.
Screenshot 2020-05-22 at 17 19 22

**Issue 2** On iOS only. The code block (the one which is tripple-backquotes): makes some gap between chats. 
To reproduce: send in chat the ` ``` codeblock ``` `  message and check message preview on Chats screen:

Screenshot 2020-05-22 at 17 19 34

cammellos:
@Serhy I believe I have addressed both issues, I don't own an IOS device so could not test issue 2 specifically, if you would not mind testing them. Thanks!

cammellos:
(in the new build, building now)

Serhy:
Right, @cammellos that fixed both issues, can't reproduce with iOS and Android anymore.

PR #10711, opened 2020-05-25
Chat view performance improvements

cammellos:
Improve chat loading performance
    
##    Move collections top level
    
Move `messages`,`message-lists`,`pagination-info` from nested in
    `chats` to top level at the db.
    The reason for this change is that if any of the `messages` fields
    change, any `sub` that relies on `chat` will be recomputed, which is
    unnecessary.
    
## Move chat-name to events
    
`chat-name` was computed dynamically, while it is now only calculated
    when loading chat the first time around.
    
## Remove `enrich-chats`
    
Enrich chats was doing a lot of work, and many subscriptions were
    relying on it.
    Not all the computations were necessary, for example it would always
    calculate the name of who invited the user to a group chat, regardless
    of whether it was actually used in the view.
    This commit changes that behavior so that we use smaller subscriptions
    to calculate such fields.
    In general we should move computations to events, if that's not
    desirable (there are some cases where we might not want to do that), we
    should have "bottom/leaf heavy" subscriptions as opposed to "top heavy",
    especially if they are to be shared, so only when (and if) we load that
    particular view, the subscription is triggered, while others can be
    re-used.
    
I have compared performance with current release, and there's a
    noticeable difference. Opening a chat is faster (messages are loaded
    faster), and clicking on the home view on a chat is more responsing
    (the animation on-press is much quicker).

Sorry for the big PR but changing one thing lead to another. I can split it off for easier reviewing if necessary.

I have left subscriptions names as they are, using `raw` for those that do not do any calculation or don't fallback on defaults, probably it would be good to revisit to have a consistent naming, I would also separate `subs.cljs` in multiple files (under a single directory), `subs/chats.cljs` `subs/contacts.cljs` etc , as sometimes is difficult to find what can be re-used (or any other scheme is good), but that's for a separate PR pending discussion.

## Testing

No functional changes, just performance improvement. Performance can be tested if necessary, I have included a video that shows the difference, mind as well that current develop parses markdown so it does a bit more, but pr seems to be faster despite of that.
Areas that are impacted is main chat view, chat view mostly.
Receiving messages is also faster it looks (significantly), although haven't looked much into it as wasn't optimizing that particular part.

release
pr

status: ready

### DIFF
--- a/src/status_im/ui/components/chat_icon/screen.cljs
+++ b/src/status_im/ui/components/chat_icon/screen.cljs
@@ -8,8 +8,6 @@
             [status-im.ui.screens.chat.photos :as photos]
             [status-im.utils.platform :as platform]))
 
-;;TODO REWORK THIS NAMESPACE
-
 (defn default-chat-icon [name styles]
   (when-not (string/blank? name)
     [react/view (:default-chat-icon styles)
@@ -32,15 +30,16 @@
      [react/view online-dot-right]]]])
 
 (defn chat-icon-view
-  [contact group-chat name _online styles]
+  [chat-id group-chat name styles]
   [react/view (:container styles)
-   (if-not group-chat
-     [photos/photo (multiaccounts/displayed-photo contact) styles]
-     [default-chat-icon name styles])])
+   (if group-chat
+     [default-chat-icon name styles]
+     (let [photo-path @(re-frame.core/subscribe [:chats/photo-path chat-id])]
+       [photos/photo photo-path styles]))])
 
 (defn chat-icon-view-toolbar
-  [contact group-chat name color online]
-  [chat-icon-view contact group-chat name online
+  [chat-id group-chat name color]
+  [chat-icon-view chat-id group-chat name
    {:container              styles/container-chat-toolbar
     :online-view-wrapper    styles/online-view-wrapper
     :online-view            styles/online-view
@@ -52,8 +51,8 @@
     :default-chat-icon-text (styles/default-chat-icon-text 36)}])
 
 (defn chat-icon-view-chat-list
-  [contact group-chat name color online]
-  [chat-icon-view contact group-chat name online
+  [chat-id group-chat name color]
+  [chat-icon-view chat-id group-chat name
    {:container              styles/container-chat-list
     :online-view-wrapper    styles/online-view-wrapper
     :online-view            styles/online-view
@@ -65,8 +64,8 @@
     :default-chat-icon-text (styles/default-chat-icon-text 40)}])
 
 (defn chat-icon-view-chat-sheet
-  [contact group-chat name color online]
-  [chat-icon-view contact group-chat name online
+  [chat-id group-chat name color]
+  [chat-icon-view chat-id group-chat name
    {:container              styles/container-chat-list
     :online-view-wrapper    styles/online-view-wrapper
     :online-view            styles/online-view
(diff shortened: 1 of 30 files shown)


src/status_im/ui/components/chat_icon/screen.cljs
```diff
@@ -8,8 +8,6 @@
             [status-im.ui.screens.chat.photos :as photos]
             [status-im.utils.platform :as platform]))
 
-;;TODO REWORK THIS NAMESPACE
```
flexsurfer: this still need to rework it :)

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -38,13 +38,12 @@
 
 (defview quoted-message
   [_ {:keys [from text]} outgoing current-public-key]
-  (letsubs [{:keys [ens-name alias]} [:contacts/contact-name-by-identity from]]
+  (letsubs [contact-name [:contacts/contact-name-by-identity from]]
```
flexsurfer, line 42: probably not in this PR but we should add this data to the message itself in the event, because of https://github.com/status-im/status-react/issues/10380

flexsurfer: I would say it's even better to do not update the name when you on the chat view, and update it only next time you open chat, because 1- its rare case when this happens, 2 - we have 30s delay for verifications, so from performance and from ux it will be better to show it next time you open chat

src/status_im/ui/screens/chat/sheets.cljs
```diff
@@ -27,40 +26,42 @@
...
-     :on-press #(hide-sheet-and-dispatch [:chat.ui/fetch-history-pressed chat-id])}]
-   [list-item/list-item
-    {:theme    :action-destructive
-     :title    :t/delete-chat
-     :accessibility-label :delete-chat-button
-     :icon     :main-icons/delete
-     :on-press #(hide-sheet-and-dispatch [:chat.ui/remove-chat-pressed chat-id])}]])
+(defn one-to-one-chat-actions [{:keys [chat-id]}]
+  (let [photo @(re-frame/subscribe [:chats/photo-path chat-id])
+        contact-name @(re-frame/subscribe [:contacts/contact-name-by-identity chat-id])]
```
flexsurfer, line 32: we know for sure these fields shouldn't change during this view is opened, so instead of creating and disposing of subscriptions we could use direct access to app-db probably in such cases , yeah i think we should consider this idea in general because in that case we will have less "noise" subscriptions, i mean such subscriptions which are run each time we change something but we know for sure result will be the same

flexsurfer (approved):

src/status_im/ui/screens/home/views/inner_item.cljs
```diff
@@ -103,20 +103,20 @@
...
-                                  contact group-chat truncated-chat-name color online false]
+                                  chat-id group-chat chat-name color online false]
       :title-prefix              (cond
                                    private-group? :main-icons/tiny-group
                                    public-group? :main-icons/tiny-public
                                    :else nil)
-      :title                     truncated-chat-name
+      :title                     (if group-chat
+                                   (utils/truncate-str chat-name 30)
+                                   @(re-frame/subscribe [:contacts/contact-name-by-identity chat-id]))
```
flexsurfer, line 127: image

src/status_im/ui/components/chat_icon/screen.cljs
cammellos: oh sorry, that was removed by mistake, generally all caps comment are the things I leave before commiting :)

src/status_im/ui/screens/home/views/inner_item.cljs
cammellos: :D , this is tricky, I know it looks a bit funny, but I don't want to run the `subscribe` if it's not a one-to-one. If I pass a component it will be not applying the header style, so I am not sure what's another option.

src/status_im/ui/screens/chat/message/message.cljs
cammellos: Most of the times though it does happen when you are in the chat view, think of the simplest case:
1) Open #status
2) Receive messages, no names, status-go verifies them
3) Names are received

In such case I'd want to see a name as soon as it's available, not on reloading the chat?

We can though look at performance, my guess is that this is not the bottleneck, more likely is the actually building up the list (datemarks,timestamps etc)

src/status_im/ui/screens/chat/sheets.cljs
cammellos: some stuff is actually already denormalized, for example `chat` now has the `photo` (which never changes), I haven't yet used the denormalized version yet as I am not 100% sure it always has it, but I can change that in a separate PR (we can subscribe only if it's nil for example).

churik:
Tested on different types of chats: 1-1, public, group chats - big messages with markdown, links, emoji, stickers, Arabic / Chinese text
Rendering messages in public chats is working faster (5 sec vs 2.5 for history 01/06/20 in Status chat).
LOOKS AWESOME!

Issue #10719, opened 2020-05-27
Change copy of seed backup warning message in Keycard onboarding


guylouis:
in Keycard onboarding, the following message is shown to the user after he is been shown is seed phrase. 
![image](https://user-images.githubusercontent.com/11461437/83050102-527b6a80-a04c-11ea-9d1f-31e970148126.png)

The copy is actually incorrect, since if the user has lost his device, if he still has his pairing code then he will be able to access his funds from an other phone. 

The copy should be changed to 
"
Written the seed phrase down! 
You won't have a second chance
If you lose access, for example by losing your keycard, you can only access your keys with your seed phrase. No one, but you has your seed phrase. Write it down. Keep it safe
"

cc @hesterbruikman

hesterbruikman:
For clarity, these are the current dialogs for regular multiaccount and keycard multiaccount. The issue is to change the keycard multiaccount message only as the keys can be accessed on another device with the pairingcode (not requiring the seed phrase)

![Frame 2](https://user-images.githubusercontent.com/35233054/83503064-d311ef00-a4c2-11ea-9ea9-cf3c288ee9ec.png)

PR #10733, opened 2020-06-02
[#10719] Change seed backup warning message

rasom:
fix #10733


hesterbruikman (approved):
Looks good to me! For other reviewers, please note that the difference is subtle. Instead of 'If you lose access, for example by losing your _device_' is changed to 'If you lose access, for example by losing your _keycard_' 

Screenshot_20200602-123937_Status PR 1

guylouis (approved):

PR #10752, opened 2020-06-03
Fix searching name of ens contacts

cammellos:
We stopped searching for ENS names in home view. This commit restores the old behavior and simplify the `home-item` subscription.

status: ready

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -969,16 +969,11 @@
 
 (re-frame/reg-sub
  :home-items
- :<- [:chats/active-chats]
  :<- [:search/home-filter]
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

churik:
@cammellos 
before it was possible to search for both values - ENS-name and 3-random name.
I.e. user with name @sergey and 3-random name `Light Bla One` can be found using `sergey` or `Light Bla One`
Can you restore this as well, if it is not too complicated?

cammellos:
@churik thanks for spotting the issue, you should be able to search for both now.

churik:
Thank you! 
Ready to be merged and cherry-picked to release 1.4

PR #10757, opened 2020-06-04
Upgrade status-go to improve indexing performance.

cammellos:
Create different index for filtering
    
The index for message was fairly inefficient as it was only using the
    cursor, as it was referring to the old `chat_id` field.
    
This meant that newer messages would be fetched much faster then older
    messages.
    
The index has been changed so that now it includes `local_chat_id`
    (which is currently used for filtering), and not using `hide`.
    
The reason being is that `hide` is a low cardinality index, so there's
    no performance benefit to have it in, also it's mostly ignored by the
    query planner.
    

status: ready


flexsurfer (approved):

cammellos:
@churik @Serhy could you have a quick look at the tests? are this genuine failures or false positives?

PR #10765, opened 2020-06-05
Set status bar style with theme

Ferossgp:
Instead of making the app full screen then changing it on js start, we can use the same theme as in the app for the Status Bar. Fixes the resize of the screen, and missing color for status bar on the app start without required login.


flexsurfer (approved):

hesterbruikman:
@Ferossgp can't install `android-e2e` build, only `android` which Jenkins doesn't seem to provide

Ferossgp:
@hesterbruikman restarted the build

hesterbruikman (approved):
Sorry for the delay! Looks good, thanks @Ferossgp

PR #10774, opened 2020-06-08
Use text inputs from Quo

### LINKED ISSUE #10787: 'Enter seed phrase input' is not centered
# Bug Report

## Problem

photo_2020-06-09 13 01 18 jpeg 2020-06-09 13-02-10

#### Expected behavior
on IOS and Android input is centered

### Reproduction

- Open Status
- Tap on Access key

### Additional Information

- Status version: release 1.4
- Operating System: iOS

Ferossgp:
Fixes #10787


src/status_im/network/ui/edit_network/views.cljs
```diff
@@ -33,31 +33,31 @@
...
-          [react/view styles/network-type
+          [react/view {:padding-vertical 8}
+           [quo/text-input
+            {:label          (i18n/label :t/name)
+             :placeholder    (i18n/label :t/specify-name)
+             :default-value  (get-in manage-network [:name :value])
+             :on-change-text #(re-frame/dispatch [::network/input-changed :name %])
+             :auto-focus     true}]]
+          [react/view {:padding-vertical 8}
+           [quo/text-input
```
flexsurfer, line 44: we should always try to have as many views as possible , why don't add a property to text-input to be able to provide styles, so you have set vertical margin for it ?

flexsurfer (approved):

src/status_im/network/ui/edit_network/views.cljs
Ferossgp: Ideally, we will have components like that: https://github.com/mobily/stacks
And we will not have hardcoded style with paddings but grids, rows, columns which will be centralized. So this concrete element should be in a `Stack` (as per library above naming), and the stack will provide vertical paddings for all its children. I do not use it for now, as I'm not convinced by its API, and wrapping it in the current version of reagent is not pretty.

flexsurfer: i mean thats cool, but i didn't find anything about performance, stack probably is cool for designers, but if we're talking about performance fewer views always better, and always better to align object itself but not by placing it into a container

Ferossgp: I do not understand what is the problems with performance, views are at almost 0 costs. It takes only bridge time, which is so minor. I haven't seen any problems, and I could not find any descriptions about the performance cost of using extra views. Do you have some benchmarks for that?

Ferossgp: Also behind the reason to wrap everything in extra view with paddings, instead of aligning object itself:
https://visly.app/blog/layout-isolated-components
https://mxstbr.com/thoughts/margin

It will become even more noticeable when we will want to support tablets.

PR #10797, opened 2020-06-10
Use text inputs from Quo

Ferossgp:
Same as #10774 but fixes the problem with clearing inputs on navigation blur which kept the old reference to the input


flexsurfer (approved):

Serhy:
test_decline_transactions_in_1_1_chat is a false fail (test expected too soon for Decline transaction be updated on receiver side - will look separately)
Manaally can create new use r nad send message in 1-1/public chats on iOS /Android 8/ Android 9

PR #10804, opened 2020-06-11
[keycard] Remove unnecessary parameter from generate-and-load-key

rasom:
`multiaccount` was used only by simulated keycard, so it is removed
and will be retreived by accessing app-db directly. This is done because
simulated kk implementation should be compatible with real keycard and
thus can't require extra parameters.

status: ready

### DIFF
--- a/src/status_im/hardwallet/simulated_keycard.cljs
+++ b/src/status_im/hardwallet/simulated_keycard.cljs
@@ -112,11 +112,18 @@
     (later #(on-success kk1-pair))))
 
 (defn generate-and-load-key
-  [{:keys [pin pairing on-success multiaccount]}]
+  [{:keys [pin pairing on-success]}]
   (when (and (= pin (get @state :pin))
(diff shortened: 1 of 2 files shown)


src/status_im/hardwallet/simulated_keycard.cljs
```diff
@@ -112,11 +112,18 @@
...
 
 (defn generate-and-load-key
-  [{:keys [pin pairing on-success multiaccount]}]
+  [{:keys [pin pairing on-success]}]
   (when (and (= pin (get @state :pin))
              (= pairing kk1-pair))
-    (let [{:keys [id address public-key derived key-uid]}
-          multiaccount
+    (let [{:keys [selected-id multiaccounts root-key derived]}
+          (:intro-wizard @re-frame.db/app-db)
```
flexsurfer, line 119: why direct access to ratom?

flexsurfer (approved):

PR #10814, opened 2020-06-15
e2e: fix for nightly 16/06

churik:
- added two possible URLs for `test_share_user_profile_url_public_chat` (HTTP and HTTPS)
- added separate waiting for transaction status change in `test_decline_transactions_in_1_1_chat `
- added clearing of text input in `test_can_change_account_settings `
- replaced click to click_until_absense_of_element in `test_can_change_account_settings` (sometimes doesn't work from 1 time)

### DIFF
--- a/test/appium/tests/atomic/chats/test_commands.py
+++ b/test/appium/tests/atomic/chats/test_commands.py
@@ -239,12 +239,9 @@ class TestCommandsMultipleDevices(MultipleDeviceTestCase):
         chat_2 = home_2.get_chat(sender['username']).click()
         chat_2_receiver_message = chat_2.chat_element_by_text('↓ Incoming transaction')
         chat_2_receiver_message.decline_transaction.click()
-        chat_1.element_by_text_part('Transaction declined').wait_for_element(20)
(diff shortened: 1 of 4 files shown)


churik:
Also confirmed with USDC

test/appium/tests/atomic/chats/test_commands.py
```diff
@@ -239,7 +239,8 @@ def test_decline_transactions_in_1_1_chat(self):
         chat_2 = home_2.get_chat(sender['username']).click()
         chat_2_receiver_message = chat_2.chat_element_by_text('↓ Incoming transaction')
         chat_2_receiver_message.decline_transaction.click()
-        chat_1.element_by_text_part('Transaction declined').wait_for_element(20)
+        for chat in chat_1, chat_2:
```
Serhy: It won't wait for 'Transaction declined'  in necessary message because there is a message in both chats with 'Transaction declined' status too (so it will take it)
Let's make it with:
`if not chat_2_sender_message.contains_text('Transaction declined') and chat_1_request_message.contains_text('Transaction declined'):`

Serhy (approved):
Great!

PR #10816, opened 2020-06-15
Override padding in search text input

Ferossgp:


yenda (approved):

Serhy:
Text in the `Search` input aligned not on iOS and Androids (8th and 9th tested) 
It fixes a part of https://github.com/status-im/status-react/issues/10807

Issue #10766, opened 2020-06-05
Add Notification Action on Android


hesterbruikman:
# Problem

Status' notification service on Android requires the app to remain in the foreground. Therefore a Foreground notification is shown to users as soon as they enable notifications.

In order the receive notifications the remains open and the keys unlocked. This poses a security risk for messaging when a phone is stolen while notifications are enabled and the service is running.

Currently the only way to stop the notification service from running is by 'force quiting' the app. There is no easy way to do this. The users has to go to App info using longpress on the app icon or by going to OS settings > Status.

- As a user I want to be able to close the app from the persistent notification banner.

## Acceptance Criteria
Banner includes `Stop Status` button
cc @errorists for other thoughts on copy

## Notes
Banner could include Open button as well to make opening the app easier, not having to find app icon

## Future Steps
Investigate how to distinguish locking app view from keys

hesterbruikman:
@yenda please add any details you think are necessary for anyone to work on this

Issue #10762, opened 2020-06-05
Make the Android Status bar notification icon same in appearance as its neighbours


errorists:
# Bug Report

## Problem

The Status notification icon in the status bar is using a blue colour fill, making it stand out unnecessarily from the rest. 

#### Expected behavior

It should follow the same system appearance, usually that's a light grey colour and only use the alpha channel found in the asset. 

Icons to be used can be found [here](https://www.dropbox.com/home/Exports/yenda%20icon)

#### Actual behavior

The icon is blue
<img width="424" alt="Screenshot 2020-06-05 at 12 18 42" src="https://user-images.githubusercontent.com/8749671/83859601-d3aabf80-a726-11ea-99f5-193245fd5fc1.png">

#### Acceptance Criteria

Icon's Status Bar appearance is fixed

### Reproduction

Turn on notifications on Android, gaze into the Status Bar :)

Issue #4187, opened 2018-05-10
Show mobile unread message badge notification


chadyj:
[comment]: # (Please replace ... with your information. Remove < and >)

### User Story
As a user, I want to see system badge notifications so that I know if there are new messages.

### Description
[comment]: # (Feature or Bug? i.e Type: Bug)
_Type_: Feature

[comment]: # (Describe the feature you would like, or briefly summarise the bug and what you did, what you expected to happen, and what actually happens. Sections below)
_Summary_: Badge notifications are a handy system pattern to let users know if there are unread notifications. Adding this to Status would help users know when there are new messages so they can check the app and respond.

#### Expected behavior

* iOS: Show a red badge without the unread count
* Android: Show the notification badge (ie. notification dot)

#### Actual behavior
No badge

### Solution
![](https://user-images.githubusercontent.com/116099/39864644-929866d6-544a-11e8-93a8-5acef2aa794a.png)

lukaszfryc:
@chadyj should it be only for messages or all notifications like passphrase backup? I think it should only apply to unread messages.

chadyj:
Yep unread messages

pedropombeiro:
If we wanted to show unread message count, this could tie in with https://github.com/status-im/status-react/issues/3451

chadyj:
Sorry @status-github-bot we want this issue.

lukaszfryc:
I think this feature should be implemented after [the mentions](https://github.com/status-im/design/issues/25). And, by default I'd show the unread counter only for mentions and other important notifications. See https://github.com/status-im/design/issues/25#issuecomment-420587747

chadyj:
Hey @dmitryn by any chance is your [PR for desktop badge notifications](https://github.com/status-im/status-react/pull/6146) extendable for mobile badge notifications?

dmitryn:
@chadyj yes, i will add support for iphone and android badges.

ghost:
This issue has been automatically closed. Please re-open if this issue is important to you.

Swader:
Still not resolved.

churik:
Still relevant in Statusv1.0.1

flexsurfer:
Still relevant in Statusv1.14.0 but now it always shows (1) if pn is enabled
![image](https://user-images.githubusercontent.com/11790366/83857880-ecfa3e80-a71b-11ea-983c-85968f3fc5bb.png)

churik:
Obsolete and will be re-considered in redesigned app.

PR #10817, opened 2020-06-15
Improve notifications

yenda:

*otherwise*

fixes #...

### Summary

...

Documentation change PR (review please): https://github.com/status-im/status.im/pull/xxx

#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- network consumption

### Steps to test

- Open Status
- ...
- Step 3, etc.

status: ready


hesterbruikman:
Great work @yenda !

Tried on Android 7 and Android 10 

- #10762 seems fixed to me. Would wait on @errorists eyes to verify though
- #10766 seems fixed to me. 🙌
- `Label translation`; Don't believe there's a GH issue for this. How would this work? Are there specific keys that need translation, Lokalise process, or is a manual effort required (e.g. like is done for play store content)
- `Remove notification badge for persistent notification` With this, do you mean #4187 ? If so, lgtm. It is a feature to see the notification badge, but if there's no way to get rid of the badge always showing 1, I don't think it's worth having

cc @Serhy

hesterbruikman:
Out of scope for this PR. @yenda I realized we're showing random name and message content in the notification. Do you foresee a feature where users can set whether they want to show this in the notification? If so, it would be good to think through if/how to show a potential nickname as well.  @errorists @0kok0 curious to hear your thoughts on this as well

Please let me know if you already have future plans so I can plan for it ☺️

yenda:
@hesterbruikman 

- regarding label we could do it with the other translation files and use some script in the build to get the strings into the android specific files

- yeah I still don't know what launcher @flexsurfer has to see a number in the badge, but stock Android and most launcher cannot show more than a dot. This PR should make sure the dot isn't shown for the service notification which is up all the time.

- for showing name and message content I suppose you are talking about hiding it when the screen is locked? if that is the case then Android already offers that option in the settings.

flexsurfer (approved):

churik:
Tested on Android 10:
- #10762 fixed 
- #10766 fixed
- retested PNs with stickers, images, long PNs (several issues will be added separately)
- several options to close service
- biometric authentication
- enable notification service on several multiaccounts
- open PNs from different places

Didn't find new issues related to this PR.

flexsurfer:
@yenda could you squash commits pls ?

PR #10824, opened 2020-06-16
Disable reload button on unsafe sites

vitvly:


flexsurfer (approved):

Serhy:
Checked with Android 8,9 and iOS, - no refresh button when navigating to blocked sites.

PR #10826, opened 2020-06-16
[#10634] Image of 'Get started' view changes height/width if navigate…

### LINKED ISSUE #10634: Image of 'Get started' view changes height/width if navigate back from "Enter seed phrase"
# Bug Report

## Problem

If to navigate back to main welcome view from "Enter your seed phrase" view then images become really small aligned to the top.
It happens only when navigating back from "Enter you seed phrase" view, but all **fine** when navigating back from `Generate keys`  or `Recover with Keycard` view.

No issues on iOS

#### Expected behavior

Images keep staying on the same place

#### Actual behavior

Screenshot 2020-05-13 at 13 56 33

### Reproduction

- Install Status and open the application
- Tap `Access key` -> `Enter seed phrase`
- Tap `<` top-left button OR hardware Back button

### Additional Information
- Status version: develop 0.14.0 (20200513025927)
- Operating System: Android only

flexsurfer:
fixes #10634

### DIFF
--- a/src/status_im/ui/screens/intro/views.cljs
+++ b/src/status_im/ui/screens/intro/views.cljs
@@ -29,7 +29,7 @@
       ^{:key i}
       [react/view {:style (styles/dot color (selected i))}]))])
 
-(defn intro-viewer [slides window-height]
+(defn intro-viewer [slides window-height _]
   (let [scroll-x (r/atom 0)
         scroll-view-ref (atom nil)
         width (r/atom 0)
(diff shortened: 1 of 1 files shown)


Serhy:
The #10634  is fixed. **But**: the regression is that when app just launches there are no images at all (any iOS and Androids).
On Android thought, when reproducing the steps from #10634 images re-appear and with the correct width-height.

Screenshot_20200616-141417_Status PR

flexsurfer:
@Serhy fixed

vitvly (approved):

Serhy:
Thanks, @flexsurfer 
With develop it was 100% reproduced. And now I can't reproduce with Galaxy j7, Galaxy S8 but 1 of 3 times reproduced on low-level device Galaxy Note 4 with Android6.0.1. That's an edge case taking into account the specific device, likelihood and severity of issue.  
Lets merge!

PR #10831, opened 2020-06-17
Show search cancel on android

### LINKED ISSUE #10807: Input is search fields on Android is not aligned and  cancel is not shown
# Bug Report

## Problem
Text is cut, cancel doesn't appear after input some text.
All fine with IOS.
photo_2020-06-12 16 01 08

#### Expected behavior
Search field is the same

#### Actual behavior
Text is cut; no Cancle button

### Reproduction

- Open Status
- Create multiaccount
- Join several chats
- Open chat search

### Additional Information

- Status version: nightly 12/06/2020
- Operating System: Android

Ferossgp:
Fixes #10807

### DIFF
--- a/src/quo/components/text_input.cljs
+++ b/src/quo/components/text_input.cljs
@@ -66,9 +66,9 @@
       false)))
 
 ;; TODO(Ferossgp): Check performance for android layout animations
-;; (when (and platform/android?
-;;            (aget rn/ui-manager "setLayoutAnimationEnabledExperimental"))
-;;   (ocall rn/ui-manager "setLayoutAnimationEnabledExperimental" true))
+(when (and platform/android?
+           (aget rn/ui-manager "setLayoutAnimationEnabledExperimental"))
+  (ocall rn/ui-manager "setLayoutAnimationEnabledExperimental" true))
 
 (def height 44)                         ; 22 line-height + 11*2 vertical padding
 (def multiline-height 88)               ; 3 * 22 three line-height + 11* vertical padding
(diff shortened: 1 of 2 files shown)


src/quo/components/text_input.cljs
```diff
@@ -66,9 +66,9 @@
       false)))
 
 ;; TODO(Ferossgp): Check performance for android layout animations
-;; (when (and platform/android?
-;;            (aget rn/ui-manager "setLayoutAnimationEnabledExperimental"))
-;;   (ocall rn/ui-manager "setLayoutAnimationEnabledExperimental" true))
+(when (and platform/android?
```
flexsurfer, line 69: hows this related to this PR ?

src/quo/components/text_input.cljs
```diff
@@ -152,11 +152,14 @@
...
                  monospace]
-          :or   {cancel-label "Cancel"
-                 show-cancel  true}
+          :or   {cancel-label "Cancel"}
           :as   props}]
       {:pre [(check-spec ::text-input props)]}
-      (let [after (cond
+      (let [show-cancel (if (nil? show-cancel)
+                          ;; Enabled by default on iOs and disabled on Android
+                          (boolean platform/ios?)
```
flexsurfer: isn't it already boolean ?

flexsurfer (approved):

src/quo/components/text_input.cljs
Ferossgp: Here I used Layout animation to animate the cancel button on iOs, to use the same animation on android, it should be enabled like that. I commented that code because at some point I thought that we will never show cancel on android (forgot about the search input)

flexsurfer (approved):

churik:
Tested on IOS and Android - cancel is shown and working as expected.
There are several more issues with the search input, but they are reproducible with nightly as well, so will be reported separately

PR #10839, opened 2020-06-17
make go-maven-resolver fail when any resolution fails

jakubgs:
Before errors like this:
```
finder.go:91: error: 'no pom data' for: <Dep ID=com.afollestad.material-dialogs:commons:0.9.6.0 O=false S= >
```
Would be printed to `stderr` but exit code would not be non-zero.

Now the whole `make nix-update-gradle` target should fail if any
dependency resolution fails.


flexsurfer (approved):

jakubgs:
For reference: https://github.com/status-im/go-maven-resolver/pull/4

PR #10845, opened 2020-06-18
[#10312] Blocking user takes ~5-10s, and chats are emtpy during that …

### LINKED ISSUE #10312: Blocking user takes ~5-10s, and chats are emtpy during that time
# Bug Report

## Problem

When blocking user, all chats are empty for ~5-10s

#### Expected behavior

There are messages in the chat then blocking user

https://drive.google.com/file/d/16AK_y1vAKWHArxCnQ87rjsC-wYckNEjV/view

flexsurfer:
fixes #10312

### DIFF
--- a/src/quo/components/button/view.cljs
+++ b/src/quo/components/button/view.cljs
@@ -6,7 +6,8 @@
             [quo.design-system.spacing :as spacing]
             [quo.components.text :as text]
             ;; FIXME:
-            [status-im.ui.components.icons.vector-icons :as icons]))
+            [status-im.ui.components.icons.vector-icons :as icons]
+            [status-im.utils.label :as utils.label]))
 
 (defn style-container [type]
   (merge {:height          44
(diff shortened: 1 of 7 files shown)


src/quo/components/button/view.cljs
```diff
@@ -81,16 +82,21 @@
...
          (= type :icon)
          [icons/icon icon {:color icon-color}]
 
-         (string? children)
+         (or (keyword? children) (string? children))
          [text/text {:weight          :medium
                      :number-of-lines 1
                      :style           {:color text-color}}
-          children]
+          (utils.label/stringify children)]
```
Ferossgp, line 99: We want to have Quo as an independent library, but this couples the component hardly to the current status implementation of translations. If at some time we will want to separate the lib, or change the implementation of i18n then we will need to refactor Quo for that.

vitvly (approved):

Serhy:
Tested on Android v8.1, v6.0.1, v9 and iOS13:
- Joined 10 public chats with 500 messages from 5 users
Blocking any users runs an animation an until it finished taps has no effect. Messages are blocked as expected, notifications (and messages) are not received from blocked contact.

Fixed e2e test, another one with send_transaction_with_custom_token known issue in github.
Thanks @flexsurfer , lets merge!

PR #10851, opened 2020-06-19
Handle encoded urls

cammellos:
We didn't check for encoded characters in the URL.
Instead now first we decode the URL components, and then match against
the URLs.
We also make sure that there's no HTML in the link passed.

status: ready


flexsurfer (approved):

Serhy:
Tested with Android 8, 9, 6.0.1 and iOS 13 webview, can't reproduce issue anymore. 
Will add necessary check in e2e test separately.
Failed e2e above is not related to this PR and has existing issue in GH

0kok0 (approved):
Looks good, lets merge this

cammellos:
cherry picked on 1.4 branch

PR #10844, opened 2020-06-18
WIP: Debug android TLS issue

Ferossgp:
USE THIS COMMIT ONLY TO HAVE MORE CRASHES ON DEVICE

PR #10852, opened 2020-06-19
Use RN's OkHttp in webview lib

vitvly:
Test using RN's OkHttpClient instance in react-native-webview.

The fix is in this commit: https://github.com/status-im/react-native-webview/commit/3bef57fcf8243c2b52b5ab668630d8c92d400de4

Instead of initialising `httpClient` instance in `RNCWebViewManager`'s constructor, we do it on-demand in `shouldInterceptRequest()`.

Related to #10844

### DIFF
--- a/mobile/js_files/package.json
+++ b/mobile/js_files/package.json
@@ -56,7 +56,7 @@
     "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#v2.5.23",
     "react-native-svg": "^9.8.4",
     "react-native-touch-id": "^4.4.1",
-    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.2.3_1",
+    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.2.3_5",
     "web3-utils": "^1.2.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


flexsurfer:
@siphiuel could you remove not related code?

vitvly:
@flexsurfer will squash and drop `TLS issue reproducibility` commit, but maybe later, after QA does a round of tests?

flexsurfer (approved):

Serhy:
I've started one e2e, the same test against https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200623-105722-a917ec-pr-x86.apk build which ran 100 times with: installing Status and then re-open app twice, - no craches. 
Manually on my Android 8.1 I had no crashes from 30+ runs too.

Seems that's all... no more TLS issue!

Should anything specific be verified in in webview, @siphiuel ?

vitvly:
@Serhy thanks! The way this works, is that we postpone the http client initialisation to the moment it is actually used in the browser. So it seems that HttpClient constructor is throwing exceptions during app startup, but should not during actual usage.

So maybe it would make sense to test opening browser pages? I tested it myself a couple of times, but just not sure if it was definite.

flexsurfer:
@siphiuel do you think this fix should be in community repo as well? i mean its not status specific bug , right ?

vitvly:
@flexsurfer it's Status-specific, in community repo they do not override `shouldInterceptRequest`. And we use `OkHttpClient` function there.

Serhy:
Tested webview on iOS, Android 9, Android 8, signing transactions with keycard account (and no-keycard account), images in webview, non-latin text

Also, as a good side effect of postponed initialise of OkHttpClient,  - Status now takes **~1.5-2.5 times** to **launch faster** (depends of device - I compared this PR build and two different PR builds: Galaxy Note 4 (Android 6) : 6.5sec with PR/nightly_develop build and 4.5 sec this build, while on Android 8 Galaxy J7: different_PR/nightly_develop = 7 sec, current PR = 2.9 sec)

PR #10857, opened 2020-06-22
[#9037] Error states for new Tx overview: contract & network fees

### LINKED ISSUE #9037: Error states for new Tx overview: contract & network fees

flexsurfer:
fixes #9037

### DIFF
--- a/src/status_im/signing/gas.cljs
+++ b/src/status_im/signing/gas.cljs
@@ -80,8 +80,10 @@
 
 (fx/defn update-estimated-gas-error
   {:events [:signing/update-estimated-gas-error]}
-  [{db :db}]
-  {:db (assoc-in db [:signing/edit-fee :gas-loading?] false)})
+  [{db :db} {:keys [message]}]
+  {:db (-> db
+           (assoc-in [:signing/edit-fee :gas-loading?] false)
+           (assoc-in [:signing/tx :gas-error-message] message))})
 
 (fx/defn update-gas-price-error
   {:events [:signing/update-gas-price-error]}
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -2091,29 +2091,31 @@
     {:amount-error (i18n/label :t/wallet-insufficient-funds)}))
 
 (defn get-sufficient-gas-error
-  [balance symbol amount ^js gas ^js gasPrice]
+  [gas-error-message balance symbol amount ^js gas ^js gasPrice]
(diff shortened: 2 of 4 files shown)


vitvly (approved):

churik:
Checked cases (IOS 13 and Android 10, dark and white mode):
- not enough ETH to pay gas for wallet and DApp (authorize screen) 
- not enough SNT, but enough ETH for gas ("Sign anyway" flow, authorize screen) for DApps
- not enough ETH + not enough ETH to pay gas for DApp
- not enough ETH, amount is too precise on Send Transaction screen
image will be implemented separately

PR #10858, opened 2020-06-22
Merge webview.js into provider.js

### LINKED ISSUE #10660: Navigation Back and Forward doesn't work on dap.ps website
# Bug Report

## Problem
Arrows Back and Forward are always grayed out on dap.ps store.
Issue is not reproducible on v.1.3. and on IOS.

#### Expected behavior
Can go "Back"

#### Actual behavior
Monosnap 2020-05-15 14-36-55

### Reproduction

- Open Status
- Open browser tab
- Tap on https://dapp.ps
- Tap on any category

### Additional Information

- Status version: nightly 15/05/2020
- Operating System: Android

vitvly:
Fixes #10660 

Functions in `webview.js` did not see `bridgeSend` defined inside an anonymous function in `provider.js`.


resources/js/provider.js
```diff
@@ -3,7 +3,11 @@
     var callbackId = 0;
     var callbacks = {};
 
-    var bridgeSend = function (data) {
+    // TODO var is not necessary here
+    // if no var is prepended, this will be treated 
+    // as a global window.bridgeSend variable
+    // which is referenced from webview.js
+    bridgeSend = function (data) {
```
flexsurfer: good catch! but probably we should move provider js code here instead of making this method global ? im not sure why provider.js is separate file

vitvly: @flexsurfer an even better catch! :) Merged.

flexsurfer (approved):

Serhy:
Tested with Android 8, 9 and iOS 13, - can navigate in dap.ps and out of it. History record kept in respective dap.ps screen after closing dapview.

PR #10861, opened 2020-06-23
Disable Register button if not enough SNT for ENS name

flexsurfer:
Disable Register button if not enough SNT for ENS name

And also fix SNT token icon


vitvly (approved):

oskarth:

Serhy:
Tested on iOS13, Android9, Android 8.1, thanks @flexsurfer all fixed!
E2E test not this PR issue (will update it separately - because frequent recovers of the same account message sent to it not received right away)

Issue #10659, opened 2020-05-15
Update framework for keycard to support pairing and commands in 1-1 chats


churik:
# Problem

Currently there is no option to add automation tests for:
- pairing card flows (Recover from keycard)
- sending assets in one-to-one chat 
At attempt to send ETH in 1-1 chat (Ropsten) - "Sign and send" label on message is not changed

## Acceptance Criteria
- can recover predefined card using "Recovery card" flow
- can send ETH in 1-1 chat to ENS and using "Sign and send" label on message

churik:
cc @rasom @guylouis

guylouis:
do we agree that this is not blocking for release, right?

churik:
yes

PR #10871, opened 2020-06-25
[simulated kk] Add proper signing for send/request transaction messages

rasom:
part of #10659

status: ready


flexsurfer (approved):

oskarth:

PR #10877, opened 2020-06-25
[#10105] Add set max to Send transaction

### LINKED ISSUE #10105: Add set max to Send transaction
# Feature Issue
### User Story
As a user I want to be able to quickly move all my funds, because, for example I lost my seed phrase

### Description
Designs on [Figma](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=5849%3A42299)

* [ ] Adds available amount in wallet for selected asset
* [ ] In case ETH is selected, the user will need to manually edit the total to reserve some funds for gas. This design does not resolve this issue yet
* [ ] When user taps on `Send all`
* [ ] Amount is updated to available assets in account

1!UNITO-UNDERSCORE!1 Chat

### Acceptance Criteria
Implementation follows design

### Notes
Best implemented after or paired with https://github.com/status-im/status-react/issues/10103

flexsurfer:
fixes #10105

Also added fiat value

### DIFF
--- a/src/status_im/ui/screens/wallet/send/views.cljs
+++ b/src/status_im/ui/screens/wallet/send/views.cljs
@@ -18,37 +18,34 @@
             [status-im.ui.screens.wallet.send.sheets :as sheets]
             [status-im.ui.screens.wallet.components.views :as components]
             [status-im.utils.utils :as utils]
-            [status-im.ui.components.button :as button]
-            [status-im.ethereum.core :as ethereum]))
+            [status-im.ethereum.core :as ethereum]
+            [status-im.utils.money :as money]))
 
-(defn header [{:keys [label small-screen? on-cancel]}]
+(defn header [{:keys [label small-screen?]}]
   [react/view (styles/header small-screen?)
    [react/view {:flex 1}
     [react/text {:style (merge {:typography :title-bold} (when small-screen? {:font-size 15}))}
-     (i18n/label (or label :t/send-transaction))]]
-   [button/button {:type            :secondary
-                   :container-style {:padding-horizontal 24}
-                   :label           (i18n/label :t/cancel)
-                   :on-press        on-cancel}]])
+     (i18n/label (or label :t/send-transaction))]]])
 
-(defn asset-selector [{:keys [request? token from]}]
+(defn asset-selector [{:keys [request? token from]} window-width]
   (let [{:keys [name icon color]} token]
     [react/touchable-highlight
-     {:style    {:flex 1}
-      :on-press (when-not request? #(do
-                                      (re-frame/dispatch [:dismiss-keyboard])
-                                      (re-frame/dispatch [:bottom-sheet/show-sheet
-                                                          {:content        (fn [] [sheets/assets (:address from)])
-                                                           :content-height 300}])))}
+     {:on-press (when-not request?
+                  #(do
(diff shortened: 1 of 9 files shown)


src/status_im/ui/screens/wallet/send/views.cljs
```diff
@@ -70,40 +67,56 @@
...
+                                                  {:content        sheets/choose-recipient
+                                                   :content-height 200}]))
+      :accessories         [:chevron]}]))
+
+(defn set-max [token]
+  [react/touchable-highlight
+   {:on-press
+    #(when token
+       (re-frame/dispatch [:wallet.send/set-max-amount token]))}
+   [react/view {:height            35 :border-radius 40 :background-color colors/blue-light
```
rasom: This looks unreadable to me, do you mind to add comas here or keep one pair per line?

flexsurfer: done

Serhy:
Tested with iOS13, Android8, Android 9 3rd build (`62871f0`)
1) On iOS there is a comma decimal separator. When asset is not integer, - fiat value is `NaN`. It's NaN also when amount is invalid input (like `1.234.56`). Not sure what is better to display as fiat value when invalid input entered. @errorists? 
2) In the 1-1 chat -> Send TX there is no `Set max` option (but fiat value is present ok)
For above two screenshot:

Screenshot 2020-06-26 at 14 37 53


3) I guess we may add fiat value in `Request TX` option in 1-1 chat too. No fiat value at the moment there.

flexsurfer:
thanks @Serhy 
2. it's intentional because when we set max we decrease on fee, so it doesn't make sense if you send in chat because you need to request address, and later fee will be different
3. yes we can

flexsurfer:
actually this set max is tricky for eth, if you set max and then change fee, it won't be max anymore

errorists:
How do I get the NaN again? I couldn't seem to replicate it, anyway that shouldn't be displayed under any circumstances, should be a 0.0
wrt to max and network fees, imo it's okay, it's just if after the change in fee, the total value is greater than available amount we need to crap out with an error and disable proceeding to next step.

Serhy:
> actually this set max is tricky for eth, if you set max and then change fee, it won't be max anymore

The way Trust Wallet works is: Set Max = sets all available ETH -> Next -> and on the next screen it becomes (Max - Fee). Further changes of fee deducts/increases this `Max` value thus making an ability to send everything.

If I'm correct we have a fixate ETH amount on the `Sending... ...... ETH` screen after specifying Asset-From-To on previous screen.

flexsurfer:
thanks @Serhy  2 and 3 fixed

oskarth:

Serhy:
Tested on iOS13, Android8.1, Android 9 - all above issues fixed, can set max value for ETH/token.

@flexsurfer I just noticed the currency in newly introduced fiat value always set to `$` disregard the fiat currency what is set when sending transaction. 
Screenshot 2020-07-01 at 17 41 42

flexsurfer:
@Serhy fixed, hided it for all cases

Serhy:
Looks good now! Meaningful currency code everywhere.
E2e test will be fixed separately (not this PR issue)

PR #10886, opened 2020-06-29
[#10865] Load only the current language json on app init 

### LINKED ISSUE #10865: Load only the current language json on app init
# Feature Issue

Currently, the Status app loads all the languages in the app bundle. We could postpone and load the languages as needed. Use the react-native inline require to load only the current language. This will increase the speed on app initialization as of a smaller loaded payload on start.

flexsurfer:
fixes #10865


jinhojang6 (approved):

Ferossgp (approved):

oskarth:

Serhy:
Tested with Android 6.0.1, Android 8, Android 9 and iOS13.
Depends of device, but on Galaxy J7 it's 2.2 times faster than another PR or nightly/release build (10 seconds vs 4.5 seconds), Galaxy S8 is 1.5 time faster (3sec vs 2sec).
Language is used according to device language settings as before.
Thanks @flexsurfer, nice improvement!

PR #10887, opened 2020-06-29
[#10868] ENS name is pending when tx has failed

### LINKED ISSUE #10868: ENS name is pending when tx has failed
# Bug Report

## Problem

Some users report that they have pending ens names with failed tx, and it's not possible to remove or register this name again

#### Expected behavior

there is no ens name in the list if tx has failed

flexsurfer:
fixes #10868


vitvly (approved):

oskarth:

churik:
In case of failed transaction ENS name is released and dropped from ENS names.
Thank you for work, @flexsurfer !

PR #10904, opened 2020-07-02
disable different ID for PR builds on iOS

jakubgs:
This was introduced in d91675116015abd0d168d8602ad24b0599c4290f and seems broken on some iOS devices. Fails with: `Unable to install "Status PR"`
photo_2020-07-02_13 41 03
So I'm temporarily making PR builds use the same ID as release builds, as it was before.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -197,7 +197,8 @@ platform :ios do
   desc 'This lane builds a new adhoc build and leaves an .ipa that is ad-hoc signed (can be uploaded to diawi)'
   lane :pr do
     unlock_keychain_if_needed
-    build_ios_adhoc(pr_build: true)
+    # Temporarily switched to not build PR builds since they don't start
+    build_ios_adhoc(pr_build: false)
   end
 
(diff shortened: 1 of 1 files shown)


churik (approved):

jakubgs:
@churik did it work for you?

jakubgs:
I actually installed this iOS build and it doesn't work for me. The PR ID change might be a red herring.

Ferossgp:
@jakubgs does this device had PR app already installed or it was the first time?

Ferossgp:
> I actually installed this iOS build and it doesn't work for me. The PR ID change might be a red herring.

Was this device ID added to the provisioning profile?

jakubgs:
It already had it. Now they are both installed and greyed out and I can't launch them at all.

jakubgs:
Gheorghe is right, my other iPhone didn't have its UDID added to the provisioning profile It does work.

jakubgs:
I think I might have fixed it in #10907.

jakubgs:
Closing in favor of #10907 which fixes the crashes.

PR #10910, opened 2020-07-03
Chat api

bgits:
### Summary
Implements DApp Chat api as per: https://github.com/status-im/specs/issues/117

### Review notes
This is my first PR on this code base, wondering how these changes should be organized and if these state mutation are idiomatic? 

There are 3 functions in `permissions.cljs` that dispatch re-frame events, I'm wondering if they should live elsewhere? Are they inline with convention to dispatch re-frame events if the action is coming from the browser API?

#### Outstanding bugs

Looking for guidance on how to resolve these issues:
- The first time a permission is invoked to send a message or join a room it causes a crash upon approval. 
- When grabbing messages and sending over the bridge on iOS the json payload can not be parsed if the message contains the character `'`. I have not found a way to escape it so for now it is removed.

### Testing notes
There is a test dapp which makes calls to the api, it can be run locally and use in the webview. It can be found here: https://github.com/status-im/status-api-tester.git
#### Platforms

- Android
- iOS

Security? Two of the calls have the permission `:yield-control? true` I don't know the full impact of this.

##### Functional

- public chats
- dapps / app browsing

### Steps to test

After running npm install in the `status-api-test` directory one can start the dapp using `embark run`. It should be accessible from `http://localhost:8000/`. Open it in the dapp browser and use the Status api tab to test calls. Keep the localhost console open as some function currently just log to console. 

status: wip

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -257,7 +257,7 @@ PODS:
     - React
   - react-native-splash-screen (3.2.0):
     - React
-  - react-native-webview (9.4.0):
+  - react-native-webview (10.3.1):
     - React
   - React-RCTActionSheet (0.62.2):
     - React-Core/RCTActionSheetHeaders (= 0.62.2)
@@ -588,7 +588,7 @@ SPEC CHECKSUMS:
   react-native-safe-area-context: 60f654e00b6cc416573f6d5dbfce3839958eb57a
   react-native-shake: de052eaa3eadc4a326b8ddd7ac80c06e8d84528c
   react-native-splash-screen: 200d11d188e2e78cea3ad319964f6142b6384865
-  react-native-webview: cf5527893252b3b036eea024a1da6996f7344c74
+  react-native-webview: 40bbeb6d011226f34cb83f845aeb0fdf515cfc5f
   React-RCTActionSheet: f41ea8a811aac770e0cc6e0ad6b270c644ea8b7c
   React-RCTAnimation: 49ab98b1c1ff4445148b72a3d61554138565bad0
   React-RCTBlob: a332773f0ebc413a0ce85942a55b064471587a71
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -32,7 +32,7 @@ in {
   gradlePropParser = callPackage ./tools/gradlePropParser.nix { };
 
   # Package version adjustments
-  xcodeWrapper = super.xcodeenv.composeXcodeWrapper { version = "11.4.1"; };
+  xcodeWrapper = super.xcodeenv.composeXcodeWrapper { version = "11.5"; };
   openjdk = super.pkgs.openjdk8_headless;
   nodejs = super.pkgs.nodejs-12_x;
 
--- a/resources/js/provider.js
+++ b/resources/js/provider.js
@@ -151,6 +151,10 @@
         return sendAPIrequest('send-to-public-chat', params);
     }
 
+    StatusAPI.prototype.getChatMessages = function() {
+      return sendAPIrequest('get-chat-messages');
+    }
+
     var EthereumProvider = function () {};
 
     EthereumProvider.prototype.isStatus = true;
--- a/resources/js/provider.js
+++ b/resources/js/provider.js
@@ -146,6 +146,11 @@
         return sendAPIrequest('contact-code');
     };
 
+    StatusAPI.prototype.sendToPublicChat = function(topic, message) {
+        const params = { topic, message };
+        return sendAPIrequest('send-to-public-chat', params);
+    }
+
     var EthereumProvider = function () {};
 
     EthereumProvider.prototype.isStatus = true;
--- a/resources/js/provider.js
+++ b/resources/js/provider.js
@@ -151,6 +151,11 @@
         return sendAPIrequest('send-to-public-chat', params);
     }
 
+    StatusAPI.prototype.gotoPublicChat = function(topic) {
+      const params = { topic }
+      return sendAPIrequest('goto-public-chat', params);
+    }
+
     StatusAPI.prototype.getChatMessages = function() {
       return sendAPIrequest('get-chat-messages');
     }
--- a/src/status_im/browser/permissions.cljs
+++ b/src/status_im/browser/permissions.cljs
@@ -5,6 +5,7 @@
             [status-im.navigation :as navigation]
             [status-im.qr-scanner.core :as qr-scanner]
             [re-frame.core :as re-frame]
+            [clojure.string :as string]
             [status-im.utils.fx :as fx]))
 
 (declare process-next-permission)
@@ -13,12 +14,15 @@
 (def supported-permissions
   {constants/dapp-permission-qr-code           {:yield-control? true
                                                 :allowed?       true}
-   ;;TODO require this to be invoked each time, otherwise dapps could exploit this.
    constants/dapp-permission-send-to-public-chat {:type       :chat
                                                   :yield-control? true
                                                   :title       "Wants to send a message"
                                                   :description "To public chat"
                                                   :icon        :main-icons/public-chat}
+   constants/dapp-permission-get-chat-messages {:type        :chat
+                                                :title       "Wants to get messages"
+                                                :description "In the current chat"
+                                                :icon        :main-icons/public-chat}
    constants/dapp-permission-contact-code      {:type        :profile
                                                 :title       (i18n/label :t/wants-to-access-profile)
                                                 :description (i18n/label :t/your-contact-code)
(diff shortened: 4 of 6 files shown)


hesterbruikman:
@bgits Thanks for kicking this off. I understand from Marketing that a Chat API, in the wider realm of extensions, has been requested by dapps. If the functionality can drive existing dapp audiences to use Status to benefit from messaging features, it makes a strong case to work on this now. 

#### Suggestion/questions
- Call with Marketing (cc @j-zerah ) to get requirements for a concrete use case of a dapp using the chat API 
- Can you think of any cases that do not require the user to share their public chat key? 
- How would notifications fit in?

cc @0kok0 for security eyes

flexsurfer:
thanks @bgits , permissions will be asked only ones, so if user gives permission to dapp, then dapp will be doing next calls silently, sending message on behalf of user is really dangerous, also current UX for permissions isn't great, if we'll have more permissions and dapp will ask all permisions on the first run, they all will be asked one after another 
current chat means dapp's public chat ?

bgits:
> * Can you think of any cases that do not require the user to share their public chat key?

There are some as such gas abstraction or gasless voting where the dapp just needs to submit a signed ethereum transaction to a channel.

bgits:
> thanks @bgits , permissions will be asked only ones, so if user gives permission to dapp, then dapp will be doing next calls silently, sending message on behalf of user is really dangerous, also current UX for permissions isn't great, if we'll have more permissions and dapp will ask all permisions on the first run, they all will be asked one after another
> current chat means dapp's public chat ?

I understand the security concerns. Perhaps for now it makes sense to put this behind either a "Security" or "Advanced" flag that is by default turned off until we think of a better approach? 

Another approach that I think requires less security concern is allowing dapps to send messages from a random key. This is still useful for certain applications such a gasless voting & gas abstraction but removes the concern of dapps impersonating a users to do malicious things.

j-zerah:
@bgits @hesterbruikman - Some very interesting ideas have come up in partner outreach. Id love to discuss further and perhaps even loop those teams in to collaborate on these. 

Two high level ideas from partners:
1.) Notifications from a dapp indicating someone that their position in a defi protocol has changed/is at risk
2.) Sending messages to friends inviting them to partake in a prediction market via the chat interface

bgits:
> 1.) Notifications from a dapp indicating someone that their position in a defi protocol has changed/is at risk
For this I think we can use the notification service @richard-ramos created and extend it to use waku.

Sending messages to friends assumes they are already on Status?

j-zerah:
> Sending messages to friends assumes they are already on Status?

Correct. Imagine some friends have a private group chat and one of the members shares a prediction market. The other friends can take a position in the market directly from the chat interface with a simple tap

hesterbruikman:
cc @corpetty for planning purposes. We'll need security requirements for a chat API to be used by dapps. I suggest limiting to use case 1 described by @j-zerah [described](https://github.com/status-im/status-react/pull/10910#issuecomment-654389318):

> 1.) Notifications from a dapp indicating someone that their position in a defi protocol has changed/is at risk

(most generalizable, least privacy sensitive)

I'm thinking of a timeframe of around 2 months, so we don't postpone this indefinitely. I understand it could draw dapps and bring utility, and more functioning dapps, to Status. We'll likely need UI work on permissions etc after we have a security model as a reference point. Does that sound feasible @corpetty?

cc @iurimatias @andremedeiros

0kok0:
I assume this will start with messages being sent on the users behalf, later exposing wallet functionality, accessing images etc. 

### Iteration 0: 

#### Dapp requirements
Example threat: Poorly implemented collectible dapps lure users into compromising their and contact's privacy by pushing/pulling messages in the clear.

- Clear data usage policy (GDPR?)
- https, potentially self-signed included in the permission 
- Can we require/enforce waku usage for the dapp to facilitate e2e encryption? 
- ethereum inherited goodies: let's require every API call to be signed by a dapp pubkey 

#### API gateway: 
Example threat: Malicious Dapp successfully circumvents API endpoint safety measures and can execute code on the user's device.

API gateway process for all API calls that can be strictly isolated. Implements least authority. Performs the following tasks for all API calls:
- logging and throttling: User can see what Dapps did on its behalf, can revoke access. Malicious dapps/Endpoints will try to authenticate using a variety of credential combinations/tokens. Must fail after too many authentication attempts in a certain amount of time.
- Dapp input sensitization 
- Prevent excessive usage and data exposure due to the central point for data and object-specific policies, considering their individual sensitivity, relying on clients to perform the data filtering before displaying it to the user.
- limits the damage of e.g., injections
- enforces https (minimum) 

####  Permission/Token: 
- use OS provided permissions: https://developer.android.com/guide/topics/permissions/overview
- State Dapp Data usage policy and granted permissions
- API tokens exchanged after handshake (minimum https/https), dapps get assigned a token, that: 
  - is revocable 
  - TTL, expiration, must not be longer than one year, so we implement dynamic least authority without requiring user reviewing permissions manually (like browser, FB,..)

bgits:
> I assume this will start with messages being sent on the users behalf, later exposing wallet functionality, accessing images etc.
> 
> ### Iteration 0:
> #### Dapp requirements
> Example threat: Poorly implemented collectible dapps lure users into compromising their and contact's privacy by pushing/pulling messages in the clear.
> 
> * Clear data usage policy (GDPR?)
> * https, potentially self-signed included in the permission
> * Can we require/enforce waku usage for the dapp to facilitate e2e encryption?
> * ethereum inherited goodies: let's require every API call to be signed by a dapp pubkey
> 
> #### API gateway:
> Example threat: Malicious Dapp successfully circumvents API endpoint safety measures and can execute code on the user's device.
> 
> API gateway process for all API calls that can be strictly isolated. Implements least authority. Performs the following tasks for all API calls:
> 
> * logging and throttling: User can see what Dapps did on its behalf, can revoke access. Malicious dapps/Endpoints will try to authenticate using a variety of credential combinations/tokens. Must fail after too many authentication attempts in a certain amount of time.
> * Dapp input sensitization
> * Prevent excessive usage and data exposure due to the central point for data and object-specific policies, considering their individual sensitivity, relying on clients to perform the data filtering before displaying it to the user.
> * limits the damage of e.g., injections
> * enforces https (minimum)
> 
> #### Permission/Token:
> * use OS provided permissions: https://developer.android.com/guide/topics/permissions/overview
> * State Dapp Data usage policy and granted permissions
> * API tokens exchanged after handshake (minimum https/https), dapps get assigned a token, that:
>   
>   * is revocable
>   * TTL, expiration, must not be longer than one year, so we implement dynamic least authority without requiring user reviewing permissions manually (like browser, FB,..)

The existing API already exposes wallet functionality using the EIP-1102 standard. It also exposes some chat functionality: https://status.im/developer_tools/status_extras/status_dapp_api.html

How to make these API enhancements secure is something we want to do and think specific exploit cases will crystalize during testing. It would be helpful to have specific exploits that use API so they can be tested against and designed to mitigate.

In the case of: "Example threat: Malicious Dapp successfully circumvents API endpoint safety measures and can execute code on the user's device.", How can we create this scenario using the chat-api?

0kok0:
>  In the case of: "Example threat: Malicious Dapp successfully circumvents API endpoint safety measures and can execute code on the user's device.", How can we create this scenario using the chat-api?

Please, if you ever possess this type of information, never disclose it publicly to prove a point. Send me a DM if you require a more concrete/testable threat model. I agree it's very generic. 

However, I am afraid I have to disagree that we should do exploit driven security development; that's the definition of being reactive. 
Also, I think having an API Gateway will payoff due to enhanced maintainability in the future and not only improve security. Building an API will allow an attacker to scale any potential exploit quickly, and in the current maturity and decentralized nature of the ecosystem, we should assume Dapps are coming and going, with little to no reasonable security assumptions.  

Let me ask a question: would you still like to use Linux if there was no kernel isolation, no package manager, or an hour to create a package?

bgits:
> > In the case of: "Example threat: Malicious Dapp successfully circumvents API endpoint safety measures and can execute code on the user's device.", How can we create this scenario using the chat-api?
> 
> Please, if you ever possess this type of information, never disclose it publicly to prove a point. Send me a DM if you require a more concrete/testable threat model. I agree it's very generic.
> 
> However, I am afraid I have to disagree that we should do exploit driven security development; that's the definition of being reactive.
> Also, I think having an API Gateway will payoff due to enhanced maintainability in the future and not only improve security. Building an API will allow an attacker to scale any potential exploit quickly, and in the current maturity and decentralized nature of the ecosystem, we should assume Dapps are coming and going, with little to no reasonable security assumptions.
> 
> Let me ask a question: would you still like to use Linux if there was no kernel isolation, no package manager, or an hour to create a package?

Not disagreeing with you. I think having an actual example makes the threat more tangible and less prone to repeating.

0kok0:
You're right, I'll make sure to provide detailed examples during the next cycle

flexsurfer:
feel free to reopen

PR #10912, opened 2020-07-05
Add reactions picker UI

Ferossgp:


src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -298,28 +298,45 @@
...
-                [react/image {:style  {:margin 10 :width 140 :height 140}
-                              ;;TODO (perf) move to event
-                              :source {:uri (contenthash/url (-> content :sticker :hash))}}]]
-               (if (and (= content-type constants/content-type-image)
-                        ;; Disabling images for public-chats
-                        (not public?))
-                 [react/touchable-highlight (image-message-press-handlers message)
-                  [message-content-image message]]
-                 [unknown-content-type message])))))])))
+(defmulti ->message :content-type)
```
flexsurfer, line 222: this one is really critical from performance pov, so here we should write code for better performance not readability

Ferossgp: I will benchmark it

Ferossgp: Percentile to milliseconds, profiled using https://reactjs.org/docs/profiler.html and https://github.com/welch/tdigest

Multimethod in this branch(removed the reactions modal to lower the code difference):
```
[Sun Jul 05 2020 17:22:19.205]  INFO     INFO [status-im.utils.profiler:23] - [profile: Messsage ( mount )]: 
 approximating 685 samples using 52 centroids
0 = 0
0.25  = 7.368131868131869
0.5  = 9.883040935672515
0.75  = 13.682692307692308
1.0 = 241
```

Develop:
```
[Sun Jul 05 2020 17:24:32.855]  INFO     INFO [status-im.utils.profiler:23] - [profile: Messsage ( mount )]: 
 approximating 654 samples using 46 centroids
0 = 0
0.25  = 7.326086956521739
0.5  = 10.205882352941176
0.75  = 14.872727272727273
1.0 = 82
```

The set of samples messages is quite small but we can already see that this branch has quite better performance then develop.

flexsurfer: not sure about comparing branches, but all benchmark say multimethods are slow https://github.com/alexkehayias/chocolatier#performance
`Don't use multimethod, use cond or condp`

flexsurfer: https://insideclojure.org/2015/04/27/poly-perf/
```

Expression | 1.6.0 | 1.7.0-beta2
-- | -- | --
(value-case 1) | 21 | 17
(value-case 5) | 26 | 18
(value-cond 1) | 4 | 7
(value-cond 5) | 90 | 74
(value-multi 1) | 41 | 44
(value-multi 5) | 47 | 40


```

Ferossgp: The worst-case with multimethod in JVM benchmark is 47ns. We have 16ms to render which is equal to 16000000. 16000000/47 so we can safely dispatch 340425 messages at once if we assume that it is the slowest part of the message render. We could definitely go with a `case` here which is transpiled into a switch, and we could gain constant time dispatching. But performance problems do not come from this part of code.

hesterbruikman:
Love these @Ferossgp @errorists!

Tested on:
Samsung s10e
Android 10

Fairphone 2
Android 7

#### Findings
- image 1 - User A reaction on message user B shows the username for user B. Against the backdrop of the chat the text feels a little out of place. I doubt it adds value. How do you feel about removing the name?
- image 2 - For both user A reacting to user B and user B reacting to user A, the picker and message move to the top when tapping while the Keyboard is open. Is this because there's no overlay on Keyboard? Not a blocker, but the behavior felt unexpected. I think I expected a transition such as `close keyboard` > `scroll down` > `show picker below message`

Frame 1

errorists:
whoa, everything on image 1 went wrong :D the menu should be aligned with the bottom left corner of the bubble, and the text is positioned in an incorrect location. We can leave the usernames and avatars where they are and don't include them in the long press menus, sure.
image 2, you are 100% correct :) thanks!

Ferossgp:
yeah, image 1 should be fixed. But the image 2 uses the same behaviour as Apple Messages, I can try to not move the list from the background

errorists:
re: image 2, not exactly, with Messages they dismiss the keyboard/toolbar for the time when the message is active and leave the scroll position intact. Here the scroll position of the chat is moved downward with the keyboard and the message moves out of existing context. It would be preferable if we either don't move the chat scroll position (Apple) or animate the Y position of the message so that it's as close to its intended position in the scrolled chat as possible.

src/status_im/chat/models/input.cljs
```diff
@@ -135,6 +135,24 @@
...
+                                   :text           "Update to latest version to see a nice emoji reacion!"}))
+
+(fx/defn send-retract-emoji-reaction-message
+  {:events [::send-emoji-reaction-retraction]}
+  [{{:keys [current-chat-id] :as db} :db :as cofx} {:keys [message-id emoji-id]}]
+  (chat.message/send-message cofx {:chat-id        current-chat-id
+                                   :content-type   constants/content-type-emoji-reaction-retraction
+                                   :emoji-reaction {:messageId message-id
+                                                    :emojiId   emoji-id}
+                                   :text           "Update to latest version to see a nice emoji reacion!"}))
```
Samyoul: This has the `content-type-emoji-reaction-retraction` content-type, but doesn't use the `emoji-reaction-retraction` object. It uses the `emoji-reaction` object

src/status_im/chat/models/input.cljs
```diff
@@ -135,6 +135,24 @@
...
                                      :text    "Update to latest version to see a nice sticker here!"})))
 
+(fx/defn send-emoji-reaction-message
+  {:events [::send-emoji-reaction]}
+  [{{:keys [current-chat-id] :as db} :db :as cofx} {:keys [message-id emoji-id]}]
+  (chat.message/send-message cofx {:chat-id        current-chat-id
+                                   :content-type   constants/content-type-emoji-reaction
+                                   :emoji-reaction {:messageId message-id
+                                                    :emojiId   emoji-id}
+                                   :text           "Update to latest version to see a nice emoji reacion!"}))
```
Samyoul: Small tpyo

```suggestion
                                   :text           "Update to latest version to see a nice emoji reaction!"}))
```

src/status_im/chat/models/input.cljs
Samyoul: Tpyo

```suggestion
                                   :text           "Update to latest version to see a nice emoji reaction!"}))
```

Serhy:
Tested last builds https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200727-111252-518894-pr-universal.apk and https://i.diawi.com/rtMZZJ  

Pair of bugs so far found:

**Issue 1** 
On Android no actions (neither Reply to message, Copy or emoji reactions)are working at the moment. However, iOS works fine.

**Issue 2** When emoji reaction pad bar is opened on image - if to tap image it opens image in a full view, but emoji pad keep staying making some subsequent further issues(in a case keyboard were active before long tap). 
To reproduce:
1) Send image in 1-1 chat.
2) Put cursor in message box so keyboard appears.
3) Long tap image to call for emoji reactions pad. It opens image thumbnail with emoji pad.
4) Tap on thumbnail. 
**Actual result**: Image opened in full mobile view + thumbnail + emoji reactions pad
**Expected result**:  Image opened in full mobile view but NO thumbnail + emoji reactions
5) Tap outside of thumbnail/emoji reactions
Actual result: Keyboard howerovers main image

Screen Recording 2020-07-27 at 19 24 06 (online-video-cutter com)


**Issue3**
'View Details'  when long tap on Sticker opens Sticker pack view, while it should open User profile view (sticker market view opens when we just tap upon sticker)

Serhy:
Two bugs discovered so far:

**Issue 4**
Regression related to 1-1 chat: for synced accounts messages are not visible when sending in 1-1 chat. On receiver side they are visible, but nothing for message sender(s) (both devices).
To reproduce:
1) Device1: Create new account. Back up seed phrase so you are able to recover this account (UserA)
2) Device2: Recover account created in step(1) 
3) Sync both account via Profile -> Sync Settings -> Devices
4) Add on any of Device1/Device2 new contact (doesn't matter from public chat or via (+)->Start new chat): 'UserB' 
**Actual result = Expected result:** Contact is in 'Contacts list' on both sides
5) In Device1/Device2 send any message in 1-1 chat to UserB
**Actual result:** Message is sent but it's not present in 1-1 chat (empty chat). Re-login/Fetch24history doesn't help. UserB receives this message, though. If UserB will send the message then UserA receives and sees it in 1-1 chat on both devices.

Note:  there is no such issue in Group and Public chats for UserA
Note2: there is no such issue in nightly build

**Issue 5**
Emoji reactions are not received in Group Chats by all participants. They are visible on Emoji-reaction-sender side but not by all other in the group.
1) Create UserA and UserB. UserA adds UserB in contacts. 
2) UserA creates Group chat with UserB
3) UserA sends a text/emoji/image message
4) Any of users react to message with emoji
Actual result: no one in Group sees the emoji reaction. Counter not increases if everyone in the Group react with the same emoji upon the same message

cammellos:
@Serhy thanks, taking a look at those

Serhy:
Tested latest build which are https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200728-143640-274fff-pr-universal.apk and https://i.diawi.com/hRfwPQ on Android 8.1, Android 9 and iOS 13

**Precondition**:
On iOS and Android 9 there is a same user used and accounts were synced between devices. On Android 8.1 there is a recovered from seed phrase different user created.

Scenarios verified:
1) All Emoji reactions in 1-1 chat Text, emoji, link, tag, code, and  sticker messages
2) Emoji reactions in Group chat (Text, emoji, link, tag, code, sticker)
3) Emoji reactions in Public chat (Text, emoji, link, tag, code, image, sticker, command(no emoji raction appears) )
4) Emoji reactions are fetched from offline 
5) Reactions from different network received
6) Clear history but emoji reaction made for message which is not in history
7) Emoji reaction not received in 1-1 / Group / Public chat from the blocked user
8) Each emoji reaction increases counter upon different users

There are few more new issues:
**Issue 6**
On Android, there is an empty space instead of keyboard comes up on longpress to call reactions picker. 
**Issue 7**
There is a weird behavior that message I long tap on Android is disappearing if first send a message and then do longtap on any message on chat. Also, there is no emoji reactions pad in that case.

Both issues above are visible below:



**Issue 8** random name visible on background when call reactions on sticker. 
Screenshot 2020-07-29 at 00 15 13

Ferossgp:
> Issue 8

It is the name of the author of the sticker, will change the color of the text to make it more accentuate. See the comparison with a simple text message

Simulator Screen Shot - iPhone 11 - 2020-07-29 at 10 24 27
Simulator Screen Shot - iPhone 11 - 2020-07-29 at 10 24 32

Ferossgp:
>Issue 6

Fixed

> Issue 7

Could not replicate, do you have more context for that issue?

churik:
@Ferossgp 
More context on **ISSUE 7**
1) reproducible only on Android < 10 (8, 8.1, 9)
2) reproducible with opened keyboard after sending message to any type of chat. Pretty visible.
Updated video [attached](https://drive.google.com/file/d/1RxHeTrJoMybfPOu8yXkmO1bFl5QXYsCa/view?usp=sharing)

UPD: found one more way to reproduce the issue (also only on Android 8, 8.1, 9) - long tap on image with opened keyboard (reproducibility ~50%), so every second time.
Updated video [attached](https://drive.google.com/file/d/1C4i4RF4a4pH9Jy_mmX0o528pOfIEYDDZ/view?usp=sharing)

churik:
Tested om builds with https://github.com/status-im/status-react/commit/e9bcae4e8c469788799f56f34b3307f255a6059e commit 


> Issue3
> 'View Details' when long tap on Sticker opens Sticker pack view, while it should open User profile view (sticker market view opens when we just tap upon sticker)

Currently, it is possible to tap on own message sticker and you will be redirected to own profile with non-working options Send message and Add to contacts. In these case I believe no options should be available.

> Issue6 On Android, there is an empty space instead of keyboard comes up on longpress to call reactions picker.

Still reproducible to me on Android 10, 8, IOS13
photo_2020-07-29 15 05 54
Or maybe I misunderstood an issue.

#### ISSUE 9 Error when tap on profile picture or name on message sticker in reactions panel
**Steps:**
- as User A: send sticker message to any chat
- as User B: long tap on this message
- as User B: tap on Username or profile picture
photo_2020-07-29 15 18 09 jpeg 2020-07-29 15-18-35

Ferossgp:
@churik 
>Currently, it is possible to tap on own message sticker and you will be redirected to own profile with non-working options Send message and Add to contacts. In these case I believe no options should be available.

Fixed

>ISSUE 9 Error when tap on profile picture or name on message sticker in reactions panel

Fixed

> Issue6 On Android, there is an empty space instead of keyboard comes up on longpress to call reactions picker.

The space on the bottom is added to keep the list position, if we close keyboard then all messages are moved to the bottom. But I disabled it on android because of our approach with inputs it has some problems.

Still, cannot reproduce on my devices. But I have a theory about the latest sent message. The message is shown in the list when it is sent, but then when the confirmation comes, the message from the list is removed and replaced with the new one. If you long press on the message before that it will be removed and thus picker will not appear. I disabled now long-press if the message isn't in sent status, if that the problem then it is a synthetic case.

 For the image, I have no idea why it does not appear, will have a look tomorrow

hesterbruikman:
fyi, haven't been able to reproduce issue 7 on Galaxy S10e (Android 10), Fairphone 2 (Android 7), iPhone 8 (iOS 13) and Fairphone 3 (e/os/Android 9)

churik:
It is fixed on older Android (8, 8.1, 9) as well with closing keyboard.
Thanks @Ferossgp 

I'll take a look into e2e today and will do acceptance testing.

package.json
```diff
@@ -59,6 +59,7 @@
     "react-native-svg": "^9.8.4",
     "react-native-touch-id": "^4.4.1",
     "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.2.3_5",
+    "tdigest": "^0.1.1",
```
flexsurfer, line 62: could you please elaborate why we need this new dep ?

src/status_im/chat/models/reactions.cljs
```diff
@@ -0,0 +1,99 @@
...
+  (message.protocol/send-retract-reaction cofx
+                                          (assoc reaction :chat-id current-chat-id)))
+
+(fx/defn receive-one
+  {:events [::receive-one]}
+  [{:keys [db]} reaction]
+  {:db (update db :reactions process-reactions [reaction])})
+
+
+;;
```
flexsurfer: ?

src/status_im/core.cljs
```diff
@@ -62,7 +62,7 @@
       (.addListener ^js react/keyboard
                     (if platform/ios?
                       "keyboardWillHide"
-                      "keyboardWDidHide")
+                      "keyboardDidHide")
```
flexsurfer, line 65: so it didn't work before ?

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -8,7 +8,7 @@
             [status-im.utils.types :as types]
             [status-im.utils.utils :as utils]
             [taoensso.timbre :as log]))
-
+;; TODO: Add reactions when go is ready
```
flexsurfer: go isn't ready?

churik:
Test suite for full nightly set: https://ethstatus.testrail.net/index.php?/runs/view/5833&group_by=cases:section_id&group_order=asc
Several cases should be updated, working on it.

src/status_im/ui/screens/chat/message/reactions_row.cljs
```diff
@@ -0,0 +1,22 @@
...
+                                           :own      own})}
+   [rn/image {:source (get constants/reactions emoji-id)
+              :style  {:width        16
+                       :height       16
+                       :margin-right 4}}]
+   [quo/text {:style (styles/reaction-quantity-style {:own own})}
+    quantity]])
+
+(defn message-reactions [message reactions]
+  (when (pos? (count reactions))
```
flexsurfer: could be just `(when (seq reactions))` ?

flexsurfer (commented):
great job :)

src/status_im/core.cljs
Ferossgp: No, on android after first open it was using always height as "open"

package.json
Ferossgp: This is used for statisitcs inside profiller, here https://github.com/status-im/status-react/pull/10912/files#diff-1a5bd8a31497455c8d7ffb1082688594R19

flexsurfer: yeah, i saw that, i mean could we avoid using it? we should try to avoid adding new dependencies because of security reasons at first

Ferossgp: Without it, it will be not possible to see live updating of render statistics. The only way will be to accumulate all data and calculate it at the end. But to calculate we will again want to use a lib cause for math. This library should not go into release as the namespace is not required and compiler will remove it.

flexsurfer (approved):

churik:
@Ferossgp testing in progress, one more small issue.
#### Issue 10 Profile name is hardly visible on dark mode
Steps:

- as User A: send message to any chat
- as User B: long tap on this message
- as User B: tap on Username or profile picture
photo_2020-07-30 13 36 52 jpeg 2020-07-30 13-37-31
Monosnap 2020-07-30 13-39-14

#### Issue 11 Cannot convert null value to object on tap on long-tap on own sent sticker

- as User A: send sticker
- as User A: long-tap on it
photo_2020-07-30 13 46 42 jpeg 2020-07-30 13-46-51

Ferossgp:
@churik Fixed 10 and 11

churik:
@hesterbruikman 
noticed that reactions are also available in Offline, so if you react somehow to message when you are offline, therefore you'll see wrong state until you re-add the same reaction.
Example:
- user A: send message to User B
- user A: go to airplane mode
- user A: long tap on the message, add thumb up(wait for ~5 mins)
- user A: go back to online
- user A: check reactions on message (have thumb up)
- user B: check reactions on message (doesn't have thumb up)

So in this case to reset counter User A should delete thumbnail reaction and re-add it again.
Is it acceptable behavior for now?

hesterbruikman:
Sounds like there is no handling of offline state for reactions. Indeed, I'd say acceptable for now.

Two directions, I'm not sure which is best at this point:
- either not allow to place react when offline (similar to how you can't send a message)
- check for new reactions when client comes online and send

churik:
test fixes in #11018, waiting for results

churik:
all failed e2e are fixed, results are in https://github.com/status-im/status-react/pull/11018#issuecomment-666427258

churik:
One more issue (thanks @qoqobolo for discovering): if some message had reactions from `User B` and then  `User A` blocks `User B` - changing reactions from `User B` on messages for `User A` has no effect

**Steps:**
- User A, User B: join public chat
- User A: send message A
- User B: like message A
- User A: blocks User B
- User B: dislike message A
- User A: unblocks User B

**Expected result:** message reactions are synchronized after unblocking 
**Actual result:** message reactions are still the same - to sync it you need to re-add reactions.
Basically situation is similar to offline.
Can be fixed separately IMO.


@Ferossgp I think last issue to fix
## Issue 12 'No messages' on preview if joining public channel with last message with reaction


**Steps:**
- User A: join public chat, send message with reaction
- User B: join same public chat, navigate back

**Expected result:**  can see message
**Actual result:** 
photo_2020-07-30 17 59 39 jpeg 2020-07-30 17-59-52

cammellos:
@churik the last issue seems to be in status-go, I will fix it and merge the PR. 
cc @Ferossgp @hesterbruikman

churik:
I don't see other issues except mentioned above, which are not blockers.

Apart from https://github.com/status-im/status-react/pull/10912#issuecomment-665289315 tested:
- all redirects to user profile  from any type of messages from reaction panel ;
- 10+ of each reaction on messages
- compatibility with 1.4.1 (messages, replies, stickers)
- PNs
- reactions in offline/online (issue described here https://github.com/status-im/status-react/pull/10912#issuecomment-666379724)
- reactions after blocking/unblocking users (issue described https://github.com/status-im/status-react/pull/10912#issuecomment-666487363)
- preview of messages with reaction
- the app is in the background with reactions panel opened
- reaction panel in all type of messages in dark mode
- reactions from keycard accounts
- devices pairing 

Waiting for e2e test results from ( e2e are triggered from https://github.com/status-im/status-react/pull/11018 branch)

churik:
failed tests are relaunched in https://github.com/status-im/status-react/pull/11018#issuecomment-667067396 with https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200731-065744-e11103-pr-x86.apk - no issues related to PR.

Ready to merge. 

Also rechecked manually different scenarios of logout to check status-go error - not reproducible.

PR #10915, opened 2020-07-06
Merge password screens

Ferossgp:


flexsurfer (approved):

errorists:
I requested one small change to where the `Just a moment…` spinner is presented, otherwise it's all good👌!

hesterbruikman:
Great work @Ferossgp! Also autoscroll looks smooth!

I'm guessing validation on type for password match takes significantly more effort. What do you think @errorists @Ferossgp? Happy to create another issue to address this in the future. PR as it is already an improvement.

cc @churik @Serhy tested:
- Create > Confirm > Next (Letters, Numbers, Special characters)
- Create > Error > Correct > Next
- Go back

Tested on:
- Samsung s10e
Android 10

- iPhone 8
iOS 13.0

- Fairphone
Android 7

Screen_Recording_20200709-115121_1_1

Ferossgp:
> I'm guessing validation on type for password match takes significantly more effort. 

Actually that's intentionally implemented like that. I found it intruding that input shows error while I just start typing, and it will show all the time that my password is wrong until I write it correctly. But it is actually not wrong, I just haven't finished typing it.

hesterbruikman:
> Actually that's intentionally implemented like that. I found it intruding that input shows error while I just start typing, and it will show all the time that my password is wrong until I write it correctly. But it is actually not wrong, I just haven't finished typing it.

That makes sense. What I was thinking was a validation on the confirmation, character by character. 

sta
stu

would give an error on the 3rd character. Next would only be enabled if also 6 characters are entered in the Confirm field.

Either way, these behaviors tend to be finicky to get right and can easily be annoying as you point out. So let's park it for now

Ferossgp:
> would give an error on the 3rd character. Next would only be enabled if also 6 characters are entered in the Confirm field.

LGTM! Added that behavior

Serhy:
@Ferossgp after rebase there is an error when tapping Next with valid passwords entered:
Screenshot 2020-07-13 at 12 18 05

Serhy:
Looks good to me! Tested on Android 8, 9 and iOS13
Lets merge (after conflicts are resolved), thanks @Ferossgp

PR #10923, opened 2020-07-07
fix tokens validation

### LINKED ISSUE #10922: Two tokens (Flexacoin and FUTURAX) is checked and unchecked at the same time in "Wallet" -> "Manage assets"
# Bug Report

## Problem

When trying to check or uncheck one of two tokens (Flexacoin or FUTURAX) in "Wallet" -> "Manage assets"  they both are checked and unchecked at the same time by one tap.

#### Expected behavior

When user checks/unchecks one of the tokens, only one token is checked/unchecked.

#### Actual behavior

When user checks/unchecks one of tokens (Flexacoin or FUTURAX), two tokens - both Flexacoin and FUTURAX - are checked/unchecked.

ezgif-4-8e83a4ed3d5c

### Reproduction

- Open Wallet
- Tap 3 dots at the top right
- Select "Manage assets"
- Find in list Flexacoin and FUTURAX
- Check Flexacoin
- Uncheck Flexacoin or FUTURAX
- Check FUTURAX
- Uncheck Flexacoin or FUTURAX

### Additional Information

- Status version: release 1.4.1

flexsurfer:
fix tokens validation
fixes: #10922


churik:
@flexsurfer can't reproduce `Wrong name for token FUEL at address 0xea38eaa3c86c8f9b751533ba2e562deb9acded40 - set to Etherparty FUEL but detected as Fuel Token` manually, but it is failed constantly on e2e builds when switching from Ropsten (default in e2e builds) to Mainnet (https://github.com/status-im/status-react/pull/10923#issuecomment-655380176)

Can you please take a look?
0034screenshot

flexsurfer:
@churik i just fixed validation in e2e but i didn't fix warnings, so currently there are warnings in e2e

churik:
@flexsurfer so what we need here? to tap OK to dismiss the warning?
It blocks now 10 e2e, so I'm not sure what is expected here.
For me sounds better to have some kind of button somewhere in e2e which will show this warning on tapping.

Also not clear what can I test to check PR itself.

flexsurfer:
we need a decision what to do with these warnings, my suggestion we just need to fix tokens data in status-react according to contracts @Serhy wdyt ?

Serhy:
> we need a decision what to do with these warnings, my suggestion we just need to fix tokens data in status-react according to contracts @Serhy wdyt ?

Yes, correct. To set token names according to their `name` in contract. Do you want me to gather that data, @flexsurfer ?

churik:
@flexsurfer just found:
```
:symbol  :SUPR
 :nft?    true
 :name    "SupeRare"
 :address "0x41a322b28d0ff354040e2cbc676f0320d8c8850d" 
```
https://etherscan.io/token/0xb932a70a57673d89f4acffbe830e8ed7f75fb9e0
https://etherscan.io/token/0x41a322b28d0ff354040e2cbc676f0320d8c8850d

We use https://etherscan.io/token/0x41a322b28d0ff354040e2cbc676f0320d8c8850d

But when I made transaction in 
https://superrare.co/ I've got transaction from https://etherscan.io/token/0xb932a70a57673d89f4acffbe830e8ed7f75fb9e0 (so seems this is right one)

Tr hash: 0x888aa55af3e6e495af94fe6a61b6659d12237b3bc27b08561862b622815bb9ad

churik:
Another one:
https://ethermon.io/
https://etherscan.io/token/0x5d00d312e171be5342067c09bae883f9bcb2003b
In Status:
```
                   {:symbol  :EMONA
                    :nft?    true
                    :name    "EtheremonAsset"
                    :address "0xb2c0782ae4a299f7358758b2d15da9bf29e1dd99"}
```

flexsurfer:
@Serhy @churik fixed all tokens according to warnings , please take a look on changes

Serhy:
1) FXC token left unchanged (duplicated and with the same icons)
2) On iOS Decentraland token lost icon after update: nothing is shown on iOS and on Android there is a default token icon
Screenshot 2020-07-10 at 14 20 32

churik:
@flexsurfer https://github.com/status-im/status-react/pull/10923#issuecomment-656624357
can you please check the issues when you've got a chance to?

flexsurfer:
i'm not sure what to do with these two cases @hesterbruikman 
1. tokens have the same symbol, but symbol must be unique 
2. token has strange name `^` 
we can remove them from list, but the problem users can add them as custom tokens and problem is here

churik:
according to FXC

https://etherscan.io/address/0xc92D6E3E64302C59d734f3292E2A13A13D7E1817 - last transaction was 92 days ago; not sure it is used now.
https://etherscan.io/address/0x4a57E687b9126435a9B19E4A802113e266AdeBde - contract is active.

flexsurfer:
ok so for 1 we can remove "old" token and prevent creating tokens with the same symbol , not sure what to do with 2

hesterbruikman:
@flexsurfer is the issue that we don't have an icon for MANA (decentraland)?

flexsurfer:
@hesterbruikman no, the issue that it has a symbol name `^` that can't be used as a path for the image

Serhy:
I thought that we may need to have some balance between token names and symbols used in the contracts and, like, in more centralised world. 
For example when the one search for Dentacoin, - token symbol in coinmarketcap is DCN. And centralised exchanges having this token also uses DCN. So here better even to use DCN than `٨`

Similar thing with DAI and SAI tokens:
DAI: https://etherscan.io/address/0x6b175474e89094c44da98b954eedeac495271d0f#readContract -> symbol in contract DAI. We use DAI.
SAI: https://etherscan.io/address/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359#readContract -> symbol in contract DAI (hex `0x4441490000000000000000000000000000000000000000000000000000000000`). We use at the moment in develop `SAI` which is good. 

There are some other token with similar story when contract symbol is not the same (or even similar) to what in coinmarketcap.

So what I suggest at this point is to use DCN for Dentacoin which should fix (2) in https://github.com/status-im/status-react/pull/10923#issuecomment-667063314 and so we able to push this in v1.6

@flexsurfer @hesterbruikman what do you think?

hesterbruikman:
That makes sense to me @Serhy `DCN` for Dentacoin.  

@flexsurfer any thoughts on the best approach to update in general? As I understand, it's best to get token name from token tracker on Etherscan (is that right @Serhy?)

What I recall from the most recent round was that Simon provided icons for a range of tokens. @Ferossgp then wrote a script to get info for those tokens. From what I gather it's probably more durable for us to contribute to https://github.com/ethereum-lists/tokens providing icon design there and using an open source list. Wdyt?

flexsurfer:
@Serhy i would agree with you, but still think we should keep validation and show this error in e2e and handle it , also we need to prevent creating custom token with an existing symbol

flexsurfer:
@Serhy so i've fixed all issues, removed this FXC ```https://etherscan.io/address/0xc92D6E3E64302C59d734f3292E2A13A13D7E1817 - last transaction was 92 days ago; not sure it is used now.```
and reverted DCN back and added it to ignore on symbol verification

flexsurfer:
hm actually i added one more verification and found that we have token with the same name but with different symbol and contract address @churik @Serhy 
```
{:address  "0x08d32b0da63e2C3bcF8019c9c5d849d7a9d791e6"
                    :decimals 0
                    :symbol   :DCN
                    :name     "Dentacoin"}

{:symbol   :٨
                    :name     "Dentacoin"
                    :address  "0x08d32b0da63e2c3bcf8019c9c5d849d7a9d791e6"
                    :decimals 0}
```

flexsurfer:
added two more validation, symbols and addresses duplication, and found 1 item with the duplicated symbol and one with duplicated address

Serhy (approved):

Serhy:
Tested on Android 8.1, Android 9 and iOS 13 
All changed tokens display to what they should according to address and token name. 
Original issue https://github.com/status-im/status-react/issues/10922 fixed too (because one token removed)

hesterbruikman:
> 
> 
> @flexsurfer just found:
> 
> ```
> :symbol  :SUPR
>  :nft?    true
>  :name    "SupeRare"
>  :address "0x41a322b28d0ff354040e2cbc676f0320d8c8850d" 
> ```
> 
> https://etherscan.io/token/0xb932a70a57673d89f4acffbe830e8ed7f75fb9e0
> https://etherscan.io/token/0x41a322b28d0ff354040e2cbc676f0320d8c8850d
> 
> We use https://etherscan.io/token/0x41a322b28d0ff354040e2cbc676f0320d8c8850d
> 
> But when I made transaction in
> https://superrare.co/ I've got transaction from https://etherscan.io/token/0xb932a70a57673d89f4acffbe830e8ed7f75fb9e0 (so seems this is right one)
> 
> Tr hash: 0x888aa55af3e6e495af94fe6a61b6659d12237b3bc27b08561862b622815bb9ad

Indeed looks like we use the [ERC20 token address](https://etherscan.io/address/0x41a322b28d0ff354040e2cbc676f0320d8c8850d) where we should be using the [ERC721 token address](https://etherscan.io/token/0xb932a70a57673d89f4acffbe830e8ed7f75fb9e0)

Symbol **SUPR**
Name **SuperRare**

@flexsurfer can you please take a look to see if that works. Let me know if I need to create a separate issue

flexsurfer:
it needs a separate issue for sure

PR #10929, opened 2020-07-08
Quo left over components in ens screen

Ferossgp:
Ens name list items update to new quo component

### DIFF
--- a/src/quo/animated.cljs
+++ b/src/quo/animated.cljs
@@ -103,13 +103,13 @@
             else-node))))
 
 (defn block [opts]
-  (.block ^js animated (clj->js opts)))
+  (.block ^js animated (to-array opts)))
 
 (defn interpolate [anim-value config]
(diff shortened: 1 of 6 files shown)


Serhy:
Updated tests accordingly. Thanks a lot @Ferossgp for this!

src/quo/animated.cljs
```diff
@@ -103,13 +103,13 @@
             else-node))))
 
 (defn block [opts]
-  (.block ^js animated (clj->js opts)))
+  (.block ^js animated (to-array opts)))
```
flexsurfer, line 106: could you elaborate please?

flexsurfer (approved):

src/quo/animated.cljs
Ferossgp: the block will always receive an array of animated nodes, all nodes are already JS object. So this is just a micro-optimization because to-array iterate faster then clj->js.

PR #10936, opened 2020-07-10
[DO NOT MERGE] Nim-Status: iOS/Android support

vitvly:
Depends on  https://github.com/status-im/nim-status/pull/44 (only because it's behind master for now)

## Overview

`nim-status` is a wrapper around `status-go`. Currently `status-go` is transformed into something that can be consumed by Android/iOS with help of `gomobile` (https://github.com/golang/go/wiki/Mobile). In this PR, `gomobile` outputs are replaced by 2 libs (static/shared) - `libstatus` and `libnim_status`, compiled from `status-go` and `nim-status` respectively.

To that end, 2 new derivations are added to `nix/status-go` - `shared` (for status-go builds) and `nim-status` (for `nim-status` builds).

Huge thanks to @jakubgs for hand-holding during Nix travails.


nix/status-go/default.nix
```diff
@@ -34,11 +34,15 @@ let
   goBuildFlags = [ "-v" ];
 
 in rec {
-  mobile = callPackage ./mobile {
-    inherit meta source goBuildFlags goBuildLdFlags;
+  shared = callPackage ./shared {
```
jakubgs: I don't get the name `shared`. It's generic as hell. Can we change it to something that actually means something?

android/build.gradle
```diff
@@ -62,7 +56,9 @@ allprojects {
         jcenter()
 
         // for geth, function, and status-go
-        flatDir { dirs "libs", System.env.STATUS_GO_ANDROID_LIBDIR }
+        flatDir { dirs "libs", System.env.STATUS_GO_ANDROID_LIBDIR, 
+                  System.env.STATUS_GO_ANDROID_ALL_LIBDIR,
```
jakubgs: Do we really need two env variables for `status-go`? What's the point?
Also, the `all` keyword seems silly to me. I'd just drop it entirely as it's just noise. Since we use the combined derivation as the main input it should just be called `status-go`, not `status-go-all`.

nix/status-go/nim-status/default.nix
```diff
@@ -0,0 +1,56 @@
...
+, stdenv
+, lib }:
+let
+  inherit (lib) getAttr;
+  callPackage = newScope { inherit meta source shared; };
+  androidAbiMap = {
+    "386" = "x86";
+    "arm" = "armeabi-v7a";
+    "arm64" = "arm64-v8a";
+  };
```
jakubgs: Define stuff like `androidAbiMap` closest to where it's used. So do it more like:
```nix
  android-all = stdenv.mkDerivation {
    name = "status-go.nim-status.android-all";
    phases = [ "symlinkPhase" ];
    symlinkPhase = let
      androidAbiMap = {
        "386" = "x86";
        "arm" = "armeabi-v7a";
        "arm64" = "arm64-v8a";
      };
    in ''
      mkdir -p $out
      ln -s ${android.x86} $out/${getAttr "386" androidAbiMap}
      ln -s ${android.arm} $out/${getAttr "arm" androidAbiMap}
      ln -s ${android.arm64} $out/${getAttr "arm64" androidAbiMap}
    '';
  }
```
But at that point you can either change the `ln` calls into an iteration/map, or just drop the dict entirely:
```nix
    symlinkPhase = ''
      mkdir -p $out
      ln -s ${android.x86} $out/x86
      ln -s ${android.arm} $out/armeabi-v7a
      ln -s ${android.arm64} $out/arm64-v8a
    '';
```

nix/status-go/nim-status/default.nix
```diff
@@ -0,0 +1,56 @@
...
+
+  ios-all = stdenv.mkDerivation {
+    name = "status-go.nim-status.ios-all";
+    phases = [ "symlinkPhase" ];
+    inherit xcodeWrapper;
+    symlinkPhase = ''
+      mkdir -p $out
+      export PATH=${xcodeWrapper}/bin:$PATH 
+      lipo -create ${ios.x86}/libnim_status.a ${ios.arm}/libnim_status.a ${ios.arm64}/libnim_status.a -output $out/libnim_status.a
```
jakubgs: What is this `lipo` call? Maybe a comment?

nix/status-go/default.nix
```diff
@@ -34,11 +34,15 @@ let
...
+    inherit meta source;
+  };
+
+  nim-status = callPackage ./nim-status {
```
jakubgs: Why are we building `nim-status` inside of `status-go` folder? Why not just move it to under `nix/`?

nix/shells.nix
```diff
@@ -46,7 +46,6 @@ let
       buildInputs = with pkgs; [ gradle maven goMavenResolver ];
       inputsFrom = [ node-sh ];
       shellHook = ''
-        export STATUS_GO_ANDROID_LIBDIR="DUMMY"
```
jakubgs: This variable is here for a reason.

nix/mobile/ios/default.nix
```diff
@@ -1,11 +1,12 @@
...
-, status-go, fastlane }:
+, status-go-shared, status-go-nim-status, fastlane }:
 
 let
   inherit (lib) catAttrs unique;
 
   pod-shell = callPackage ./pod-shell.nix { };
-  status-go-shell = callPackage ./status-go-shell.nix { inherit status-go; };
+  status-go-shell = callPackage ./status-go-shell.nix { 
+    inherit status-go-shared status-go-nim-status; };
```
jakubgs: Please move the closing `}` to next line.

nix/status-go/nim-status/build.nix
```diff
@@ -0,0 +1,129 @@
...
+  inherit (lib) attrNames getAttr strings concatStringsSep concatMapStrings;
+
+  osId = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 2;
+  osArch = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 0;
+
+  ANDROID_HOME = androidPkgs;
+  ANDROID_NDK_HOME = "${androidPkgs}/ndk-bundle";
+
+  nimIosPatch = pkgs.fetchurl { url = "https://patch-diff.githubusercontent.com/raw/nim-lang/Nim/pull/15268.patch"; sha256 = "04jg0ngqlyzwrc19rafx4s257my6bhd27myv4p78hmsyzhp36za6"; };
+  patchedNim = pkgs.nim.overrideAttrs (_: { version = "pr-15268"; patches = [ nimIosPatch ]; });
```
jakubgs: One, don't create long lines like this, the enter key is your friend.
Second, if we want to use a patched nim compiler lets just put it in the `nix/overlay.nix`.

nix/status-go/nim-status/build.nix
```diff
@@ -0,0 +1,129 @@
...
+            CXX=$(xcrun --sdk ${iosSdk} --find clang++) ";
+
+    arPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ar"}";
+    ranlibPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ranlib "}";
+ 
+
+in stdenv.mkDerivation rec {
+  name = "nim-status"; # TODO: use pname and version
+  inherit compilerFlags linkerFlags osId compilerVars 
+    nimCpu nimPlatform arPath ranlibPath;
```
jakubgs: This whole `inherit` line is useless and confusing. Just use the variables where you use them:
```nix
  preBuildPhase = ''
    echo '${compilerFlags}' >> config.nims
    echo '${linkerFlags}' >> config.nims
  '';
```
So it's clear where they are used. I tried searching for `compilerFlags` and where they are used for a while only to realize that my `/\<compilerFlags\>` vim search couldn't match `$compilerFlags`

nix/status-go/nim-status/build.nix
```diff
@@ -0,0 +1,129 @@
...
+  buildInputs = with pkgs; [ git coreutils findutils gnugrep gnumake gcc patchedNim ];
+  src = fetchFromGitHub {
+    owner = "status-im";
+    repo = "nim-status";
+    name = "nim-status";
+    rev = "113ab223795fa44f9f6d3ecb9a0da7e033022ea9";
+    sha256 = "19nvn8b0n1r4nqjiq14057iwyr2m4a37i33yvf3ngjik26jyyzxk";
+    #fetchSubmodules = true;
```
jakubgs: Remove stuff you don't use.

nix/status-go/nim-status/build.nix
```diff
@@ -0,0 +1,129 @@
...
+
+    arPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ar"}";
+    ranlibPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ranlib "}";
+ 
+
+in stdenv.mkDerivation rec {
+  name = "nim-status"; # TODO: use pname and version
+  inherit compilerFlags linkerFlags osId compilerVars 
+    nimCpu nimPlatform arPath ranlibPath;
+  buildInputs = with pkgs; [ git coreutils findutils gnugrep gnumake gcc patchedNim ];
```
jakubgs: The only thing that actually makes sense here is `patchedNim`. Everything else is either redundant due to `stdenv` or not used.

nix/status-go/nim-status/build.nix
```diff
@@ -0,0 +1,129 @@
...
+  nimPlatform = "--os:${(if platform == "ios" then "ios" else "android")} ";
+  iosSdk = if arch == "386" then "iphonesimulator" else "iphoneos";
+  iosArch = getAttr arch iosArchMap;
+
+  compilerFlags = "switch(\"passC\", \""  +(concatStringsSep " " 
+  (if platform == "android" || platform == "androideabi"then
+    [("-isysroot " + ANDROID_NDK_HOME + "/sysroot") ("-target ${androidTarget}" + api)]
+    else
+    ["-isysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}"]
+    )) +"\")";
```
jakubgs: GOD that's ugly. I'll look into making this simpler.

nix/mobile/default.nix
```diff
@@ -6,17 +6,17 @@ let
...
-    status-go = status-go.mobile.android;
+    status-go-android-all = status-go.shared.android-all;
+    nim-status-android-all = status-go.nim-status.android-all;
   };
 
   ios = callPackage ./ios {
     inherit fastlane;
-    status-go = status-go.mobile.ios;
+    status-go-shared = status-go.shared.ios-all;
+    status-go-nim-status = status-go.nim-status.ios-all;
```
jakubgs: This naming is horrid. Why can't se just call them `status-go` and `nim-status`?

nix/mobile/android/release.nix
```diff
@@ -81,7 +83,8 @@ in stdenv.mkDerivation rec {
   ANDROID_NDK_ROOT = "${androidPkgs}/ndk-bundle";
 
   # Used by the Android Gradle build script in android/build.gradle
-  STATUS_GO_ANDROID_LIBDIR = "${status-go}";
+  STATUS_GO_ANDROID_ALL_LIBDIR = "${status-go-android-all}";
```
jakubgs: So we get rid of `STATUS_GO_ANDROID_LIBDIR` but not remove reference to it from Gradle config?

android/build.gradle
```diff
@@ -41,12 +41,6 @@ subprojects {
 }
 
 allprojects {
-    beforeEvaluate {
-      if (System.env.STATUS_GO_ANDROID_LIBDIR == null || System.env.STATUS_GO_ANDROID_LIBDIR == "") {
-        throw new GradleException('STATUS_GO_ANDROID_LIBDIR environment variable is not valid!')
-      }
-    }
```
jakubgs, line 48: Why did you remove this?

nix/status-go/shared/build.nix
```diff
@@ -0,0 +1,141 @@
...
+  goTags = if platform == "android" || platform == "androideabi" then
+  ""
+  else " -tags ios ";
+
+in buildGo114Package rec {
+  pname = source.repo;
+  version = "${source.cleanVersion}-${source.shortRev}-${platform}-${arch}";
+
+  inherit meta compilerVars goOs goArch goTags targetArch platform ldArch buildMode 
+  compilerFlags linkerFlags libraryFileName;
```
jakubgs: You're inheriting stuff that you don't need to inherit. Maybe you're confused as to what `inherit` does.
It __puts__ the inherited variables as keys into the __map__ within which you call `inherit`. That means that the map that you are passing to `buildGo114Package` as argument will contain the key `ldArch` with it's value. For what purpose? You are already using `ldArch` in the `buildPhase` explicitly, so what's the point of inheriting it?

nix/status-go/shared/build.nix
```diff
@@ -0,0 +1,141 @@
...
+    "386" = "x86_64";
+    "arm" = "armv7";
+    "arm64" = "arm64";
+  };
+  iosSdk = if arch == "386" then "iphonesimulator" else "iphoneos";
+  iosArch = getAttr arch iosArchMap;
+  compilerFlags = if platform == "android" || platform == "androideabi" then
+  "\"-isysroot $ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/sysroot -target ${androidTarget}${api}\""
+  else
+  "\"-isysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -miphonesimulator-version-min=7.0 -fembed-bitcode -arch ${iosArch}\"";
```
jakubgs: There has to be a cleaner way of doing this. I'll look into it.

nix/status-go/shared/build.nix
```diff
@@ -0,0 +1,141 @@
...
+  goOs = if platform == "android" || platform == "androideabi" then
+    "android"
+    else "darwin";
+
+  goArch = if platform == "android" || platform == "androideabi" then
+  if arch == "386" then "386" 
+  else if platform == "androideabi" then "arm "else "arm64"
+    else
+    if arch == "386" then "amd64" else 
+    if arch == "arm" then "arm" else "arm64";
```
jakubgs: Christ...

nix/status-go/shared/build.nix
```diff
@@ -0,0 +1,141 @@
...
+  '';
+
+  buildPhase = ''
+    runHook preBuildPhase
+    echo "Building shared library..."
+
+    export GOOS=${goOs} GOARCH=${goArch} API=${api}
+    export TARGET=${targetArch}-linux-${platform}
+
+    #export LIBRARY_PATH=$ANDROID_NDK_HOME/platforms/android-$API/arch-${ldArch}/usr/lib
```
jakubgs: if unused, remove.

nix/status-go/default.nix
vitvly: I'll dump it so there will be just `status-go` and `nim-status` instead of `status-go.shared` and `status-go.nim-status`.

nix/status-go/nim-status/build.nix
jakubgs: How about something like:
```nix
{
  compilerFlags =
    if platform == "android" || platform == "androideabi"
    then "-isysroot " + ANDROID_NDK_HOME + "/sysroot -target ${androidTarget}${api}")
    else "-isysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}";

  linkerFlags =
    if platform == "android"
    then "--sysroot " + ANDROID_NDK_HOME + "/platforms/android-${api}/arch-${ldArch} -target ${androidTarget}"
    else "--sysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}";
}
```
and
```nix
{
  preBuildPhase = ''
    echo 'switch("passC", "${compilerFlags}")' >> config.nims
    echo 'switch("passL", "${linkerFlags}")' >> config.nims
  '';
}
```

nix/status-go/shared/build.nix
```diff
@@ -0,0 +1,141 @@
...
+  androidTarget = targetArch + "-linux-" + platform;
+
+  iosArchMap = {
+    "386" = "x86_64";
+    "arm" = "armv7";
+    "arm64" = "arm64";
+  };
+  iosSdk = if arch == "386" then "iphonesimulator" else "iphoneos";
+  iosArch = getAttr arch iosArchMap;
+  compilerFlags = if platform == "android" || platform == "androideabi" then
```
jakubgs: You can use [`hasPrefix`](https://github.com/status-im/nixpkgs/blob/3355743cfd8d89cfa8d3b8e6d40c48e62e72d36a/lib/strings.nix#L182-L196) to do this:
```nix
lib.hasPrefix "android" platform
```

nix/status-go/default.nix
jakubgs: Sounds good.

nix/status-go/nim-status/default.nix
vitvly: `lipo` merges arch-specific binaries into one fat iOS binary. Will add a comment.

nix/status-go/default.nix
vitvly: Will move so that they are siblings.

nix/status-go/shared/build.nix
vitvly: These are remnants of `nix-shell` debugging. `inherit`ed variables can be `echo $var`ed in `nix-shell`.

nix/status-go/nim-status/default.nix
jakubgs: Also, making this multi-line would make it easier to read.

nix/status-go/shared/build.nix
jakubgs: That's true, but so can be the phase in which they are used.

nix/status-go/nim-status/build.nix
jakubgs: Aven nicer:
```nix
{
  isAndroid = lib.hasPrefix "android" platform;
  isIOS = platform == "android";

  compilerFlags =
    if isAndroid then "-isysroot ${ANDROID_NDK_HOME}/sysroot -target ${androidTarget}${api}"
    else if isIOS then "-isysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}"
    else throw "Unsupported platform!";

  linkerFlags =
    if isAndroid then "--sysroot ${ANDROID_NDK_HOME}/platforms/android-${api}/arch-${ldArch} -target ${androidTarget}"
    else if isIOS then "--sysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}"
    else throw "Unsupported platform!";
}
```

nix/status-go/shared/build.nix
jakubgs: I was thinking more like:
```nix
{
  goArch =
    if isAndroid then (
      if arch == "386" then "386" else
      if platform == "androideabi" then "arm"
      else "arm64"
    ) else if isIOS then (
      if arch == "386" then "amd64" else 
      if arch == "arm" then "arm"
      else "arm64"
    ) else throw "Uknown Platform!";
}
```

jakubgs: The `if platform == "androideabi" then "arm"` part makes me worried tho. Every other if is checking `arch`, but this one is checking `platform`. Some weird edge case?

jakubgs: The `else "arm64"` is kinda ugly too. I'd prefer a clear case rather than `else`.

nix/status-go/default.nix
vitvly: Done.

android/build.gradle
vitvly: Dropped `all` suffix.

nix/status-go/nim-status/default.nix
vitvly: Removed `androidAbiMap`.

nix/status-go/nim-status/default.nix
vitvly: Added a comment + multiline invocation.

nix/status-go/default.nix
vitvly: Done.

nix/shells.nix
vitvly: Restored.

nix/mobile/ios/default.nix
vitvly: Done.

nix/status-go/nim-status/build.nix
vitvly: Fixed long-line issue, but haven't moved it to `overlay.nix` yet.

nix/status-go/nim-status/build.nix
vitvly: Agree, thanks, removed all of these `inherit` entries.

cammellos:
@siphiuel could you rebase against develop and re-run e2e tests please once the build is ready? (go to projects -> Pipelien for Qa -> Move issue from e2e column to another column and back to e2e will trigger them) 

https://github.com/status-im/status-react/projects/7 

Thank you!

nix/mobile/ios/status-go-shell.nix
```diff
@@ -1,30 +1,15 @@
...
 
-{ lib, mkShell, status-go }:
+{ lib, mkShell, status-go, nim-status }:
 
 mkShell {
   shellHook = ''
-    export STATUS_GO_IOS_LIBDIR=${status-go}/Statusgo.framework
+    export RCTSTATUS_DIR="$STATUS_REACT_HOME/modules/react-native-status/ios/RCTStatus"
+    cp ${status-go}/* $RCTSTATUS_DIR
+    cp ${nim-status}/* $RCTSTATUS_DIR
```
jakubgs, line 12: You got rid of the scripts that try to avoid re-copying stuff. Is that because it's just not worth to optimize it?

nix/status-go/default.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, callPackage, mkShell, openjdk, androidPkgs }:
```
jakubgs, line 1: We are including `callPackage` but we also make our own with `newScope`. Better to drop this one.
We don't seem to use `openjdk` and `androidPkgs` either.

nix/status-go/default.nix
```diff
@@ -15,30 +15,49 @@ let
...
 
-  goBuildFlags = [ "-v" ];
+  android-x86 = callPackage ./build.nix { platform = "android"; arch = "386"; };
+  android-arm = callPackage ./build.nix { platform = "androideabi"; arch = "arm"; };
+  android-arm64 = callPackage ./build.nix { platform = "android"; arch = "arm64"; };
+
+  android = stdenv.mkDerivation {
+    name = "status-go.android";
+    phases = [ "symlinkPhase" ];
+    symlinkPhase = ''
```
jakubgs: I would add a comment above this derivation explaining what it does.

nix/status-go/default.nix
```diff
@@ -15,30 +15,49 @@ let
...
-  mobile = callPackage ./mobile {
-    inherit meta source goBuildFlags goBuildLdFlags;
+  ios-x86 = callPackage ./build.nix { platform = "ios"; arch = "386"; };
+  ios-arm = callPackage ./build.nix { platform = "ios"; arch = "arm"; };
+  ios-arm64 = callPackage ./build.nix { platform = "ios"; arch = "arm64"; };
+
+
+  ios = stdenv.mkDerivation {
+    name = "status-go.ios";
+    phases = [ "symlinkPhase" ];
```
jakubgs: Same here, I'd add a comment what's happening here. You also forgot to add a comment explaining the `lipo` call.

nix/status-go/default.nix
```diff
@@ -15,30 +15,49 @@ let
...
-  # These are necessary for status-go to show correct version
-  paramsLdFlags = attrValues (mapAttrs (name: value:
-    "-X github.com/status-im/status-go/params.${name}=${value}"
-  ) goBuildParams);
-
-  goBuildLdFlags = paramsLdFlags ++ [
-    "-s" # -s disabled symbol table
-    "-w" # -w disables DWARF debugging information
-  ];
+
```
jakubgs, line 13: Unnecessary newline.

nix/status-go/default.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, callPackage, mkShell, openjdk, androidPkgs }:
+{ lib, stdenv, newScope, xcodeWrapper, callPackage, mkShell, openjdk, androidPkgs }:
 
 let
   inherit (lib)
```
jakubgs: It seems like this `inherit` doesn't inherit anything we actually use in this file and can be dropped.

nix/status-go/default.nix
```diff
@@ -15,30 +15,49 @@ let
...
+  ios-x86 = callPackage ./build.nix { platform = "ios"; arch = "386"; };
+  ios-arm = callPackage ./build.nix { platform = "ios"; arch = "arm"; };
+  ios-arm64 = callPackage ./build.nix { platform = "ios"; arch = "arm64"; };
+
+
+  ios = stdenv.mkDerivation {
+    name = "status-go.ios";
+    phases = [ "symlinkPhase" ];
+    inherit xcodeWrapper;
+    symlinkPhase = ''
```
jakubgs: If we're running just one phase, which is the `symlinkPhase` then we can ditch the whole default builder thing which does the phases and just specify our own shell script as the builder, like here:
https://github.com/status-im/nixpkgs/blob/3355743c/pkgs/os-specific/darwin/native-x11-and-opengl/default.nix

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+    "arm64" = "arm64";
+  };
+
+  iosSdk = if arch == "386" then "iphonesimulator" else "iphoneos";
+  iosArch = getAttr arch iosArchMap;
+
+  compilerFlags = if isAndroid then
+  "\"-isysroot $ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/sysroot -target ${androidTarget}${api}\""
+  else if isIOS then
+  "\"-isysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -miphonesimulator-version-min=7.0 -fembed-bitcode -arch ${iosArch}\""
```
jakubgs: If you add quotes in `buildPhase` like so:
```
export CGO_CFLAGS="${compilerFlags}"
export CGO_LDFLAGS="${linkerFlags}"
```
You can avoid the escaping of quotes (`\"`) you do in here.

Actually, they might even be useless in this case.

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+  else throw "Unsupported platform!";
+
+  linkerFlags = if isAndroid then
+  "\" --sysroot $ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/sysroot \
+  -target ${androidTarget}${api} -v -Wl,-soname,libstatus.so\""
+  else if isIOS then
+  "\"--sysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -miphonesimulator-version-min=7.0 -fembed-bitcode -arch ${iosArch}\""
+  else throw "Unsupported platform!";
+
+  #linkerFlags = linkerFlags1 + (concatStringsSep " " goBuildLdFlags);
```
jakubgs, line 69: Useless.

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+      ''
+        export CC=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/clang
+        export CXX=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/clang++
+        export PATH=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${osId}-${osArch}/bin:$PATH
+      '' 
+      else if isIOS then ''
+        export PATH=${xcodeWrapper}/bin:$PATH 
+        export CC=$(xcrun --sdk ${iosSdk} --find clang) 
+        export CXX=$(xcrun --sdk ${iosSdk} --find clang++)
+        ''
```
jakubgs: Inconsistent formatting of `''` quotes. EIther put them in the `if`/`else` lines, or in their own lines.

nix/mobile/android/default.nix
```diff
@@ -13,7 +15,8 @@ let
 
   # TARGETS
   release = callPackage ./release.nix {
-    inherit keystore jsbundle status-go watchmanFactory;
+    inherit keystore jsbundle watchmanFactory 
+    status-go nim-status;
```
jakubgs: Just put it in the same line.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
+{ pkgs, stdenv, lib, fetchFromGitHub
+# Dependencies
+, xcodeWrapper
+, git, androidPkgs
+, platform ? "android"
+, arch ? "386"
+, api ? "23" }:
+
+let
+  inherit (lib) attrNames getAttr strings concatStringsSep concatMapStrings;
```
jakubgs: The only thing used from here is `getAttr`, which might just as well be called as `lib.getAttr` and this line dropped.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
...
+
+  osId = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 2;
+  osArch = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 0;
+
+  ANDROID_HOME = androidPkgs;
+  ANDROID_NDK_HOME = "${androidPkgs}/ndk-bundle";
```
jakubgs, line 21: This appears to be not used in `mkDerivation` and hence useless.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
...
+
+  nimIosPatch = pkgs.fetchurl { url = "https://patch-diff.githubusercontent.com/raw/nim-lang/Nim/pull/15268.patch"; sha256 = "04jg0ngqlyzwrc19rafx4s257my6bhd27myv4p78hmsyzhp36za6"; };
+  patchedNim = pkgs.nim.overrideAttrs (_: { 
+    version = "pr-15268"; 
+    patches = [ nimIosPatch ]; 
+  });
```
jakubgs: Lets move this to `nix/overlay.nix` and just replace default `nim` with this one.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
...
+      else "PATH=${xcodeWrapper}/bin:$PATH \
+            CC=$(xcrun --sdk ${iosSdk} --find clang) \
+            CXX=$(xcrun --sdk ${iosSdk} --find clang++) ";
+
+    arPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ar"}";
+    ranlibPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ranlib "}";
+ 
+
+in stdenv.mkDerivation rec {
+  name = "nim-status"; # TODO: use pname and version
```
jakubgs: Using `name` means that version is not included in the name of package.
This should look more like:
```nix
let
  src = fetchFromGitHub {
    owner = "status-im";
    repo = "nim-status";
    name = "nim-status";
    rev = "113ab223795fa44f9f6d3ecb9a0da7e033022ea9";
    sha256 = "19nvn8b0n1r4nqjiq14057iwyr2m4a37i33yvf3ngjik26jyyzxk";
  };
in stdenv.mkDerivation {
  pname = "nim-status";
  version = strings.substring 0 7 src.rev;
  inherit src;
}
```
You could keep `src` within `mkDerivation` object, but that would require the `rec` before object, and `rec` is dodgy.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
...
+    "--sysroot ${ANDROID_NDK_HOME}/platforms/android-${api}/arch-${ldArch} -target ${androidTarget}"
+    else if isIOS then
+    "--sysroot $(xcrun --sdk ${iosSdk} --show-sdk-path) -fembed-bitcode -arch ${iosArch}"
+    else throw "Unsupported platform!";
+
+    compilerVars = if platform == "android" || platform == "androideabi" then
+      "PATH=${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin"}:$PATH "
+      else "PATH=${xcodeWrapper}/bin:$PATH \
+            CC=$(xcrun --sdk ${iosSdk} --find clang) \
+            CXX=$(xcrun --sdk ${iosSdk} --find clang++) ";
```
jakubgs: I think this could use `isAndroid` and `isIOS` as well, plus the `else throw` case.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,126 @@
...
+
+    arPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ar"}";
+    ranlibPath = "${ANDROID_NDK_HOME + "/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ranlib "}";
```
jakubgs: That's a really weird way to do that. Why not:
```nix
{
    arPath = "${ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ar";
    ranlibPath = "${ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${osId}-${osArch}/bin/${targetArch}-linux-${platform}-ranlib";
}
```
You're double-templating for no reason.

nix/nim-status/default.nix
```diff
@@ -0,0 +1,51 @@
...
+  android-arm64 = callPackage ./build.nix { platform = "android"; arch = "arm64"; };
+
+  ios-x86 = callPackage ./build.nix { platform = "ios"; arch = "386"; };
+  ios-arm = callPackage ./build.nix { platform = "ios"; arch = "arm"; };
+  ios-arm64 = callPackage ./build.nix { platform = "ios"; arch = "arm64"; };
+
+  android = stdenv.mkDerivation {
+    name = "nim-status.android";
+    phases = [ "symlinkPhase" ];
+    symlinkPhase = ''
```
jakubgs: Same here, you can just use a `builder = writeScript "nim-status-android-builder" ''`.

nix/nim-status/default.nix
```diff
@@ -0,0 +1,51 @@
+{ newScope
+, xcodeWrapper
+, stdenv
+, lib }:
+let
+  inherit (lib) getAttr;
```
jakubgs: Not used.

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+      -o ${libraryFileName} \
+      $BUILD_FLAGS \
+      ./statusgo-lib
+
+    echo "Shared library built:"
+    ls -la ./libstatus.*
+  '';
+
+  fixupPhase = ''
+    set -x
```
jakubgs: Unnecessary `set -x`.

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+
+  # These are necessary for status-go to show correct version
+  paramsLdFlags = attrValues (mapAttrs (name: value:
+    "-X github.com/status-im/status-go/params.${name}=${value}"
+  ) goBuildParams);
+
+  goBuildLdFlags = paramsLdFlags ++ [
+    "-s" # -s disabled symbol table
+    "-w" # -w disables DWARF debugging information
+  ];
```
jakubgs: This appears to be unused as well. Using vim's `*` binding is a great way to find those.

cammellos:
@siphiuel looks like there are some crashes in the e2e tests, I have attached one of the logcat (error starts at 13349 though is not clear to me exactly what it is) `: runtime.cc:2067] Thread[35,tid=5643,Native,Thread*=0xc9205e00,peer=0x14b51308,"mqt_native_modules"] attempting to detach while still running code
`
Some of the transactions etc failed tests might not be related as we might be using a different token, though I should have changed it back.

I will add the manual qa label to this one, as it might require some attention from the QA team
[logcat.log](https://github.com/status-im/status-react/files/5335670/logcat.log)

nix/status-go/build.nix
```diff
@@ -0,0 +1,164 @@
...
+      else throw "Unsupported platform!";
+
+  goOs = if isAndroid then "android" else "darwin";
+
+  goArch = if isAndroid then arch
+  else if isIOS then (
+    if arch == "386" then "amd64" 
+    else arch
+  )
+  else throw "Unsupported platform!";
```
jakubgs: Not the pretties `if` I've seen, how about:
```nix
{
  goArch =
    if isAndroid then arch
    else if isIOS then (if arch == "386" then "amd64" else arch)
    else throw "Unsupported platform!";
}
```

jakubgs (commented):
Almost there.

vitvly:
@jakubgs thanks will fix
@cammellos thanks, yes, i had a look at these too, one that i can reproduce is crash during logout() fn invocation. Tried to debug, but it's happening somewhere inside `status-go`'s logout(), it seems. But this is a first glance, will dig deeper. This is Android-only, on iOS everything works fine.

nix/mobile/ios/status-go-shell.nix
jakubgs: I'm all for it if it doesn't cause issues/slowdowns, I didn't write that.

nix/nim-status/build.nix
jakubgs: Same goes for the inclusion of `androidPkgs` at the top of the file.

churik:
@siphiuel 
### ISSUE 1: Constant crash on logout on Android.
**Steps:**
- create new account
- log out

Logs: [logcat (2).log](https://github.com/status-im/status-react/files/5339603/logcat.2.log)
steps + logs: https://app.saucelabs.com/tests/ff1803bac9244b9da361b459a637698d?auth=d5fdb3b243f15dbbf4c8e74726ea2034#49

nix/mobile/ios/status-go-shell.nix
vitvly: To be honest, i accidentally dropped it, but looking at the history, maybe it doesn't really make sense, it's not that much.

nix/status-go/default.nix
vitvly: Removed.

nix/status-go/default.nix
vitvly: Added.

nix/status-go/default.nix
vitvly: Done.

nix/status-go/default.nix
vitvly: Removed.

nix/status-go/default.nix
vitvly: Done.

nix/status-go/default.nix
vitvly: Thanks, used `writeScript` instead.

nix/status-go/build.nix
vitvly: Done.

nix/status-go/build.nix
vitvly: Removed.

nix/status-go/build.nix
vitvly: Done.

nix/mobile/android/default.nix
vitvly: Fixed.

nix/nim-status/build.nix
vitvly: Replaced with androidToolPathPrefix.

nix/nim-status/default.nix
vitvly: Done.

nix/nim-status/build.nix
```diff
@@ -0,0 +1,122 @@
...
+, git 
+, platform ? "android"
+, arch ? "386"
+, api ? "23" }:
+
+let
+  osId = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 2;
+  osArch = builtins.elemAt (builtins.split "\-" stdenv.hostPlatform.system) 0;
+
+  ANDROID_NDK_HOME = "${androidPkgs}/ndk-bundle";
```
jakubgs, line 21: I'd rename that to `ndkBundle` since it isn't used as an env variable and the way it looks might confuse someone.

jakubgs (approved):
Looking good so far. Nice work. Glad we have another Nix wizard in Status.

nix/mobile/ios/status-go-shell.nix
jakubgs: Fair enough, if it works fine I'm fine with it.

churik:
@siphiuel can you check android build please? Thanks!

vitvly:
@churik sorry, i've just pushed updated changes (using recent `master` of `nim-status`), working on them now, will let you know once done.

cammellos:
@churik is the failed test related to this PR? Otherwise this would be ready to be tested, essentially there should be no feature changes and should be stable, we are looking for testing the app running for a long time, receiving lots of messages etc
Thanks!

churik:
### ISSUE 1 "Choose chat name"  screen with random names is rendering too slow (Android only)
screen is rendered around 4 sec
ezgif com-gif-maker (2)

nix/nim-status/pcre.nix
```diff
@@ -0,0 +1,132 @@
...
+, api ? "23" } :
+
+
+let
+  src = pkgs.fetchurl {
+    name = "pcre-source.tar.gz";
+    url = "https://ftp.pcre.org/pub/pcre/pcre-8.44.tar.gz";
+    sha256 = "0a3pnhzhd8wipmqq9dcn3phb494hkn47pxqsf8skj3xxyd5gvjmf";
+
+  };
```
jakubgs, line 18: Why can't we use `pkgs.pcre.src`?

nix/nim-status/openssl.nix
```diff
@@ -0,0 +1,115 @@
...
+, api ? "23" } :
+
+
+let
+  src = pkgs.fetchurl {
+    name = "openssl-source.tar.gz";
+    url = "https://www.openssl.org/source/openssl-1.1.1h.tar.gz";
+    sha256 = "5c9ca8774bd7b03e5784f26ae9e9e6d749c9da2438545077e6b3d755a06595d9";
+
+  };
```
jakubgs, line 18: Why can't we use `pkgs.openssl.src`?

churik:
### ISSUE 2 Crash on IOS when tapping "Generate keys" on nightly build from PR branch
Made [nightly builds](https://ci.status.im/job/status-react/job/manual/794/) from  'feature/nim-status-ios' 
App is crashed at attempt to create multiaccount
ezgif com-gif-maker (3)

[StatusIm-201109-122725-afdb9d-nightly.ipa_2020_11_9_15_42_49.log](https://github.com/status-im/status-react/files/5511219/StatusIm-201109-122725-afdb9d-nightly.ipa_2020_11_9_15_42_49.log)

### ISSUE 3 Can't build IOS release build
Build: https://ci.status.im/job/status-react/job/combined/job/mobile-ios/8357/console
```
16:11:13  [15:11:12]: Exit status: 
16:11:13  +------------------------------------+-------------------------------------------------------------+
16:11:13  |                                           Lane Context                                           |
16:11:13  +------------------------------------+-------------------------------------------------------------+
16:11:13  | PLATFORM_NAME                      | ios                                                         |
16:11:13  | LANE_NAME                          | ios release                                                 |
16:11:13  | KEYCHAIN_PATH                      | ~/Library/Keychains/adhoc_1604934507.625528.keychain-db     |
16:11:13  | SIGH_PROFILE_TYPE                  | app-store                                                   |
16:11:13  | MATCH_PROVISIONING_PROFILE_MAPPING | {"im.status.ethereum"=>"match AppStore im.status.ethereum"} |
```

vitvly:
On issue 2 - tried running `make release-android BUILD_TYPE=nightly` and testing the resulting build, and it worked fine. Maybe some stale build problem on Jenkins.

churik:
@siphiuel 
IOS PR build is crashed on "Generate keys" (nightly builds are fine)
[StatusIm-201119-110935-545faf-pr.ipa_2020_11_20_13_0_40.log](https://github.com/status-im/status-react/files/5573338/StatusIm-201119-110935-545faf-pr.ipa_2020_11_20_13_0_40.log)

churik:
@siphiuel any updates on this?

vitvly:
@churik i've updated the `nim-status` branch with recent changes in that repo's `master`, and now am integrating it into status-react builds. Once done will push an update here, and check the iOS PR build crash.

flexsurfer:
@siphiuel any updates ?

vitvly:
@flexsurfer will rebase, resolve conflicts, and retest. This is still relevant.

vitvly:
Superseded by https://github.com/status-im/status-react/pull/11793.

PR #10950, opened 2020-07-14
Disable by default list item animations on android

Ferossgp:
Because performance on android is worse, disable till will find a solution to improve

### DIFF
--- a/src/mocks/js_dependencies.cljs
+++ b/src/mocks/js_dependencies.cljs
@@ -164,6 +164,7 @@
                                        :PureNativeButton         #js {}
                                        :TapGestureHandler        #js {}
                                        :PanGestureHandler        #js {}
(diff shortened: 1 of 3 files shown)


hesterbruikman:
Moving conversation from Discord here for reference.

Definitely noticing a difference on Fairphone 2, Android 7. Seems before I could count to 4s, for a bottom sheet to come up, now 2s

Serhy:
On Samsung Galaxy J7 with this build is much better. Scrolling is smooth on lists of Chats and Browser views. 
For Assets view created separate issue https://github.com/status-im/status-react/issues/10957

Serhy:
Light-blue shadowing upon element taps looks much better than grey one we had before Quo integration (and definitely better than having no indication list element taped :)

cammellos (approved):

PR #10958, opened 2020-07-15
[#10928] Always show blocked users

### LINKED ISSUE #10928: Blocked users are not shown in Profile > Contacts in case when contact list is empty
# Bug Report

## Problem
Consequence of https://github.com/status-im/status-react/pull/10896
When user doesn't have contacts but has blocked users, no button "Blocked users" is shown in Contacts
Reason of failed 6213

#### Expected behavior
 "Blocked users" is shown

#### Actual behavior
Monosnap 2020-07-08 16-23-35

### Reproduction

- Open Status
- Create multiaccount
- Block any user from public chat (don't add him to Contacts)

### Additional Information

- Status version: nightly 08/07/20

rasom:
fix #10928

status: ready

### DIFF
--- a/src/status_im/ui/screens/contacts_list/views.cljs
+++ b/src/status_im/ui/screens/contacts_list/views.cljs
@@ -32,33 +32,31 @@
             contacts      [:contacts/active]]
     [react/view {:flex 1}
      [topbar/topbar {:title :t/contacts}]
-     (if (seq contacts)
-       [react/scroll-view {:flex 1}
-        [add-new-contact]
-        (when (pos? blocked-contacts-count)
-          [react/view {:margin-vertical 16}
-           [quo/list-item
-            {:title               (i18n/label :t/blocked-users)
-             :icon                :main-icons/cancel
-             :theme               :negative
-             :accessibility-label :blocked-users-list-button
-             :chevron             true
-             :accessory           :text
-             :accessory-text      blocked-contacts-count
-             :on-press            #(re-frame/dispatch [:navigate-to :blocked-users-list])}]])
+     [react/scroll-view {:flex 1}
+      [add-new-contact]
+      (when (pos? blocked-contacts-count)
+        [react/view {:margin-vertical 16}
+         [quo/list-item
+          {:title               (i18n/label :t/blocked-users)
+           :icon                :main-icons/cancel
+           :theme               :negative
+           :accessibility-label :blocked-users-list-button
+           :chevron             true
+           :accessory           :text
+           :accessory-text      blocked-contacts-count
+           :on-press            #(re-frame/dispatch [:navigate-to :blocked-users-list])}]])
+      (if (seq contacts)
         [list.views/flat-list
          {:data                      contacts
           :key-fn                    :address
-          :render-fn                 contacts-list-item}]]
-       [react/view {:flex 1}
-        [add-new-contact]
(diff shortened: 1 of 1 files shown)


cammellos (approved):

hesterbruikman:
🙌 Thanks @rasom! Found 2 issues that probably need to be addressed outside of this PR:
- Issue 1 - depends on whether a blocked user is technically still a contact. If so, the UI of `Contacts` needs an update which can happen in another issue. If not, the blocked user profile should consistently show `Add to contacts` (currently shows `Remove from contacts`
- Issue 2 - I'm not seeing messages of unblocked users. Is this because they're only shown going forward, not retroactively?

## Tested 
### Devices

- Galaxy S10e Android 10

### Scenarios
✅ Scenario A (block non-contact)
1. Tap contact in public chat
2. Tap `Block this user`
3. View blocked user in `Blocked users`
4. View chat (no messages of blocked user)

Scenario B (block contact)
1. Tap contact in public chat
2. Tap `Add to contacts`
3. Tap `Block this user`
4. View blocked user in `Blocked users`
5. View chat (no messages of blocked user)
6. Tap on blocked user in `Blocked users`
⚠️ 7. View Profile > User is listed as `Blocked user`, not as `Contact`, while Profile still shows `Remove from contacts`. I'd expect either for the user to either be removed from contacts or a better distinction on the Contact screen showing not a separate section of Blocked users, but an icon or label that shows a Contact is blocked.

Scenario C (unblock)
1. Tap on user in `blocked users`
2. Tap `Unblock this user`
⚠️ 3. View chat > Not seeing messages of unblocked user. This happens after unblocking a contact and unblocking a non-contact. When unblocking a non-contact, they disappear altogether as their profile is no longer in Contacts, Blocked users, or in chat. Could be an issue of delay or could be that messages of unblocked contacts are only shown going forward and that's simply not clear from the UI

rasom:
@hesterbruikman this PR only fixes UI so that we show `Blocked users` button even if there are no added contacts. The rest of it, the logic of blocking, etc, is not changed in this pr and should be addressed separately.

hesterbruikman:
Check, I'll create separate issues for those above

cammellos:
@hesterbruikman 
> Issue 2 - I'm not seeing messages of unblocked users. Is this because they're only shown going forward, not retroactively?

This is the expected behavior, when blocking a contact we remove all the messages, if unblocked only newly retrieved messages will be shown. 
The rationale is that messages of blocked users might have some ill effect on the correct functioning of the app (for example they filled the database, so that the app is sluggish), so they are completely removed. If we soft delete them in order to show them to the user once the contact is unblocked, then we are still exposed to this sort of risk. 
In my opinion is not worth the risk, we can warn the user before blocking that any message will be completely deleted and not recovered.

hesterbruikman:
>  In my opinion is not worth the risk, we can warn the user before blocking that any message will be completely deleted and not recovered.

Cool, thanks for explaining @cammellos! There is a dialog on block, can take another look at the copy there as part of this separate issue. Mostly I think the issue is we don't communicate so you're left wondering "did that work? 🤔"

churik:
test_long_press_to_delete_chat is not related - fixed in another PR

Serhy:
Tested with iOS and Android8, Android9. 
test_decline_transactions_in_1_1_chat - not related to this PR, - will update it separately.

PR #10960, opened 2020-07-15
nix: fix conditional that selects .env.e2e for ABI=x86 builds

jakubgs:
This should fix an issue where `nightly` type builds for end-to-end tests don't use `.env.e2e`.
The order of the if's was incorrect. When ABI is set to only `x86` using the `.env.e2e` should override build type.

### DIFF
--- a/nix/mobile/android/release.nix
+++ b/nix/mobile/android/release.nix
@@ -32,9 +32,9 @@ let
   name = "status-react-build-${baseName}";
 
   envFileName = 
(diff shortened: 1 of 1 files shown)


Serhy (approved):

Serhy:
Checked manual e2e build made of this branch and PR build here and testnet is set by default in both. 
Thanks a lot @jakubgs !

rasom (approved):

andremedeiros (commented):
👏

PR #10971, opened 2020-07-17
Add e2e test for PN

Serhy:
Add e2e test to check PN received. 

Test case (and the test script here) covers:
a) PNs are present if Notifications enabled during onboarding
b) PN is present if Notifications enabled from Profile
c) PN for text message
d) PN for emoji message

Negative cases (when PN have not to appear like for blocked users) and PN for commands/image/sticker/reply message will be done later in next PRs.

### DIFF
--- a/test/appium/tests/atomic/chats/test_one_to_one.py
+++ b/test/appium/tests/atomic/chats/test_one_to_one.py
@@ -35,6 +35,64 @@ class TestMessagesOneToOneChatMultiple(MultipleDeviceTestCase):
         device_2_chat = device_2_home.get_chat(default_username_1).click()
         device_2_chat.chat_element_by_text(message).wait_for_visibility_of_element()
 
+    @marks.testrail_id(6283)
+    @marks.high
+    def test_push_notification_1_1_chat(self):
+        self.create_drivers(2)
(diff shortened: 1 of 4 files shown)


test/appium/tests/atomic/chats/test_one_to_one.py
```diff
@@ -35,6 +35,55 @@ def test_text_message_1_1_chat(self):
...
+        device_1.open_notification_bar()
+        if not device_1.element_by_text_part(emoji_unicode).is_element_displayed(10):
+            device_1.driver.fail("Push notification with emoji was not received")
+        chat_view_device_1 = device_1.click_upon_push_notification_by_text(emoji_unicode)
+        device_1.just_fyi("Check Device 1 is actually on chat")
+        if not (chat_view_device_1.element_by_text_part(message).is_element_displayed()
+            and chat_view_device_1.element_by_text_part(emoji_unicode).is_element_displayed()):
+            device_1.driver.fail("Failed to open chat view after tap on PN")
+
+
```
churik, line 95: what about checking notification badge?

Serhy:
Added a check for PN is gone after message was seen. 
We don't have yet that feature of red badge on Status app icon for unread messages.

churik (approved):

PR #10997, opened 2020-07-27
[#10988] Add entry point to start new chat from search results 

### LINKED ISSUE #10988: Add entry point to start new chat from search results
# Problem
We've had reports of people not finding ways to start a 1:1 chat. Presumably a contributing factor to this is expecting to search for a 'user ID' in the searchbar on the chat tab. As is possible in Telegram. This seems likely as we've also had reports from CC's asking to search for ENS names directly.

We want to:
- Direct people entering an ENS name (or other unsupported queries) to where they can find what they're looking for
- Set expectations of the searchbar in line with it's functionality
- Prevent introducing higher expectations, such as searching for public chats

Design on [Figma](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=2485%3A30)
Home Tab -_ Fill Search

## Future Steps
None, after this we're done 😜

Ferossgp:
Fixes #10988


src/status_im/ui/screens/home/views.cljs
```diff
@@ -96,7 +98,29 @@
...
                          (re-frame/dispatch [:search/home-filter-changed ""])))
      :on-change      (fn [text]
-                       (re-frame/dispatch [:search/home-filter-changed text]))}]])
+                       (re-frame/dispatch [:search/home-filter-changed text])
+                       (when-not (seq chats)
+                         (debounce/debounce-and-dispatch [:new-chat/set-new-identity text] 300)))}]])
+
+(defn start-suggestion [search-value]
+  (let [{:keys [state ens-name public-key]} @(re-frame/subscribe [:contacts/new-identity])]
+    [react/view {}
```
flexsurfer: **just** wondering, why an empty map is needed here?

src/status_im/ui/screens/home/views.cljs
```diff
@@ -96,7 +98,29 @@
                        (when-not search-filter
                          (re-frame/dispatch [:search/home-filter-changed ""])))
      :on-change      (fn [text]
-                       (re-frame/dispatch [:search/home-filter-changed text]))}]])
+                       (re-frame/dispatch [:search/home-filter-changed text])
+                       (when-not (seq chats)
+                         (debounce/debounce-and-dispatch [:new-chat/set-new-identity text] 300)))}]])
```
flexsurfer: do we clean it later? for example, if a user presses cancel? i mean **just** to make sure if we open create new chat later input field will be empty

flexsurfer (approved):

src/status_im/ui/screens/home/views.cljs
Ferossgp: Yes, it is not needed here

src/status_im/ui/screens/home/views.cljs
Ferossgp: It is cleared when we open `start a new chat` screen. But I will add now also clear here, so if we will use in other palces too, it will be clean.

churik:
@Ferossgp 
### ISSUE 1:  Error when tap on public chat name under "Do you mean"
*Steps:*
1) join any public chat
2) type 'aaaaa' in search
3) tap on 'aaaa' under Do you mean

**Expected result:** redirect to public chat
**Actual result:** 
IMAGE 2020-07-27 18:15:18

### ISSUE 2:  No chats are suggested if I type existing ENS with domain 

I.e. when typing 'tanyatest1.eth' or 'qbnewzz.stateofus.eth' I would expect that 1-1 chat is suggested

### ISSUE 3: Search input is not cleared after redirecting to chat

*Steps:*
1) join any public chat
2) type 'tanyatest10' in search
3) tap to start 1-1 chat
4) tap back

**Expected result:** redirect to home view, search input is cleared
**Actual result:** redirect to home view, in the search field the same value is set

### ISSUE 4: 1-1 chat is not suggested in case of pasting ENS to search input

**Steps:**
Scenario 1: 
1) join any public chat
2) paste 'tanyatest10' in search

**Expected result:** 1-1 chat will be suggested
**Actual result:**  nothing happened

Ferossgp:
@churik 

> ISSUE 1

Fixed

> ISSUE 2:

Most probably the connection was slow? We do not show spinner for loading but use the same logic as start a new chat. I've tried these ens and it worked.
> ISSUE 3: Search input is not cleared after redirecting to chat

This is the same behavior as searching other chats
> ISSUE 4

Fixed

qoqobolo:
@Ferossgp 

### ISSUE 5: Search input and chat list disappear after deleting public chat if navigating from suggestions

_Steps:_

1. Join any public chat
2. Type random word in search
3. Tap "Join a public chat"
4. In opened chat tap 3 dots on the top right 
5. Delete chat -> Delete

**Expected result:** redirect to home screen, chat list and search input are shown
**Actual result:** redirect to home screen, no chat list and search input 

Video: https://drive.google.com/file/d/1-sJDkNWG5bQnBmX6bBdNavR87_cQBdNP/view?usp=sharing

Ferossgp:
@qoqobolo Because this is also present in current release and it is not a regression of this PR, I would prefer to have it as a separate issue to be solved independently of this task

qoqobolo:
@Ferossgp Yes, you are right about both issues (3 and 5)! Thanks for a great job and sorry for the distraction.

qoqobolo:
Tested latest build which are https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200728-101749-cb4b02-pr-universal.apk and https://i.diawi.com/ox3cbt on Android 8, Android 10 and iOS 13.

**Scenarios verified:**

1) Join 1-1 chats and public chats and search for them (typing value, pasting value)
2) Typing and pasting chat’s name that are not in my chat list and joining them via “Do you mean” links: 
- ens names (no domen,  with .stateofus.eth, with .eth)
- valid values for public chat names, but not valid ENS
3) Invalid values + adding chars in the end of value / deleting chars in the end
4) Navigating back and forth after some actions in chat view (deleting chats, clearing history)

PR #10999, opened 2020-07-27
fix scan qr window

### LINKED ISSUE #10980: Header text and button "Cancel" are overlapped with status bar on "Scan QR code" view
# Bug Report

## Problem

Header text and button "Cancel" are overlapped with the status bar on "Scan QR code" screen.

photo_2020-07-17_16-14-34

### Reproduction

- Open Status
- Home screen > tap "+" 
- Start new chat
- Tap QR code button on the top right

The same situation in all cases when scanning QR code during add new contact with QR code or in the Wallet.

### Additional Information

- Status version: Nightly build 1.4.1

- Operating System: Android 8, 10

flexsurfer:
fix #10980 scan qr window

### DIFF
--- a/src/status_im/ui/components/colors.cljs
+++ b/src/status_im/ui/components/colors.cljs
@@ -40,7 +40,7 @@
 
 ;; BLACK
 (def black (:black light))                                  ;; Used as the default text color
-(def black-persist (:black light))                          ;; this doesn't with theme
+(def black-persist (:white dark))                           ;; this doesn't with theme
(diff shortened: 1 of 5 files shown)


Ferossgp:
The app background is white on light theme, because of that when we give control of inset to the navigation we get this:
photo_2020-07-28 10 37 13

IMO it should control the insets on its own as we do with animated headers. So this way we can control the insets background color.

flexsurfer:
thanks @Ferossgp , fixed

Ferossgp (approved):

PR #11004, opened 2020-07-28
F-Droid build enabling changes

jakubgs:
After a lot of research in #8512 I finally have a working F-Droid build using this branch.

Changes:
- Set GRADLE_USER_HOME to a temp dir
- Enable `build-use-sandbox = true` for Android Nix builds
- Use `org.gradle.daemon=false` for Nix Android builds
- Add `ANDROID_APK_SIGNED` to make unsigned builds for F-Droid

I think we should be able to release `1.5.0` on F-Droid.
It should be also doable to release `1.4.1` with these changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,7 @@ release-android: export BUILD_ENV ?= prod
 release-android: export BUILD_TYPE ?= nightly
 release-android: export BUILD_NUMBER ?= $(TMP_BUILD_NUMBER)
 release-android: export KEYSTORE_PATH ?= $(HOME)/.gradle/status-im.keystore
+release-android: export ANDROID_APK_SIGNED ?= true
 release-android: export ANDROID_ABI_SPLIT ?= false
 release-android: export ANDROID_ABI_INCLUDE ?= armeabi-v7a;arm64-v8a;x86
 release-android: keystore ##@build build release for Android
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -228,6 +228,8 @@ android {
             keyAlias 'androiddebugkey'
             keyPassword 'android'
         }
+        /* Caution! In production, you need to generate your own keystore file.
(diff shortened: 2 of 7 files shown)


cammellos (approved):

flexsurfer (approved):
Awesome!

rasom (approved):

hesterbruikman:
@jakubgs this is fantastic! For my understanding, does this mean packaging is all done and we could submit 1.5 when ready or should we still go through a [request for packaging](https://gitlab.com/fdroid/rfp/-/issues)?

If we can start with 1.4.1, what would be actions needed?

jakubgs:
@hesterbruikman I have not submitted the App to F-Droid yet. And I've been wondering how to do it.

We can either wait till `1.5.0` and release that. Or we have two options:

* Take `release/1.4.x` branch, apply the changes from this PR and make it into `1.4.2`, and submit that
* Fork `release/1.4.x` branch into `release/1.4.x-fdroid` and apply the changes from this PR leaving it as `1.4.1`

I think the first option is more... proper? Kosher?

hesterbruikman:
I'm thinking the 2nd, keeps any view on retention data more pure as we know what was in a release for 1.4.1, followed by 1.5. More build makes impact harder to track. So I think it makes sense to:
- Treat 1.4.1 as trial
- Submit 1.5 when ready and communicate it's availability, e.g. also adding F-droid badge to https://status.im/get/

What makes you consider the first option 1.4.2 more kosher?

jakubgs:
Just a general rule that if you apply changes you should change the version. But to be fair the changes are only related to the build system, and none the that app codebase itself, so in theory the resulting application should be almost identical.

If you think that's better I don't really mind. This way we can get it ready for the `1.5.0` release when we can announce availability though F-Droid catalog.

jakubgs:
@churik should I worry about the one e2e test failing?

churik:
@jakubgs no, that is fixed in develop

Issue #10242, opened 2020-03-30
Add Update Gesture to Wallet


3esmit:
# Feature Issue

It's a common feature in apps the gesture to swipe down on the panel to update it. 

Currently the wallet don't have any way of updating it by user request, relaying on a non transparent (and confusing) update methodology. 

### User Story

As a user, I want to update wallet balance so that I can be sure I have the latest information.

hesterbruikman:
Design for this was originally specified as part of #8443. The gesture was dropped to simplify and unblock a basic loading state [here](https://github.com/status-im/status-react/issues/8443#issuecomment-553798316). PR for basic loading state [here](https://github.com/status-im/status-react/pull/10196).

Original designs (still valid). Find then below and on [Figma](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=4183%3A1914).

Pull-to-refresh gesture should work on:
- Wallet overview screen 
- Individual account screen

Pull-to-refresh gesture should update:
- Currency value
- Tx history

Spinner shown on pull-to-refresh should also be shown on system-initiated refresh. All existing spinners on fiat value should be removed.

<img width="761" alt="Screenshot_2020-03-30_at_15 21 48" src="https://user-images.githubusercontent.com/35233054/77929623-29d43000-72aa-11ea-855b-aa006c06e5cd.png">

> Here's an old GIF with the old design, but the concept is the same https://dribbble.com/shots/3777816-Status-Wallet

hesterbruikman:
Further requirements on `Pull to refresh` behavior based on https://github.com/status-im/status-react/pull/11007#issuecomment-670483323

The user should always be able to trigger `Pull to refresh`, client behavior on this action depends on `mobile data setting`:
- On wifi: 
     - Fetch updates
- On mobile data:
     - `Enabled` - Fetch updates
     - `Disabled` - Prompt settings bottom sheet `Sync using mobile data`
     - `Data saver` - Fetch updates

dotemacs:
Hello @andremedeiros 

I've started on this issue as per Ceri's instruction.

This is the direction I started to take:
https://github.com/status-im/status-react/compare/develop...dotemacs:add-update-gesture-to-wallet-10242?expand=1

To explain, I've added `refresh-control` in a simplest possible way, just show my initial approach, in order to get some feedback.

Is `quo.react-native` the right ns to pull in `RefreshControl`?
Or should I put it in some other ns?

[...]

PR #11007, opened 2020-07-29
[#9986] Allow to disable fetching of tx history on mobile network

### LINKED ISSUE #9986: Allow to disable fetching of tx history on mobile network
# Allow to disable fetching of tx history on mobile network

### User Story

As a user, I want to have an option to disable fetching of tx history on mobile network to save mobile traffic. 

### Description

Currently we fetch history regardless of available network, and this might require some internet traffic (tens of megabytes on initial scan if account contains transactions and hundreds of megabytes if when listening to recent history during the day). This can be expansive for some users.

### Acceptance Criteria

[comment]: # (Rules for the future PR to be accepted.)

### Notes

[comment]: # (For the developer who will implement. Include planning poker estimates, guidance, etc.)

rasom:
The esiest way to test it would be to compare data usage on release version and this one. You would need to switch to mobile network, select "stop syncing", and then keep app in the foreground for 5mins. 

fixes #9986

status: ready


src/status_im/ui/screens/wallet/accounts/views.cljs
```diff
@@ -180,6 +182,39 @@
...
 
+(defn wallet-service-state []
+  (let [status  @(re-frame/subscribe [:network-status])
+        type    @(re-frame/subscribe [:network/type])
+        setting @(re-frame/subscribe [:wallet-service/manual-setting])
+        state   @(re-frame/subscribe [:wallet-service/state])]
+    [react/view
+     {:style {:flex 1
+              :flex-direction :row}}
+     [react/text {:style {:flex 1}} (str "setting: " setting)]
```
flexsurfer: i18n?

flexsurfer: or is this for testing?

src/status_im/wallet/core.cljs
```diff
@@ -601,3 +602,77 @@
...
+                "state"    state
+                "network-type" network-type
+                "network-status" network-status)
+      (when (and config/disable-wallet-on-mobile-network?
+                 (or
+                  (nil? network-status)
+                  (= network-status :online))
+                 (nil? setting))
+        (cond
+          (and (not= network-type "wifi")
```
flexsurfer: constant?

src/status_im/ui/screens/wallet/accounts/views.cljs
rasom: yep, that's temporary

flexsurfer (approved):

hesterbruikman:
@rasom This is a good solution. Didn't learn too much about bandwidth use in target markets unfortunately, but anyway we can reduce and give control makes sense. Does `Force start` enable syncing over mobile data?

@errorists what are your thoughts on Enable/Disable message syncing and tx syncing over mobile data? Should this be a single setting or should both be controlled separately? As this impacts Settings and the Sync over Mobile bottom sheet

errorists:
@hesterbruikman it should be a single setting and unified with the chat dialogues

rasom:
@hesterbruikman @errorists 
Done, fetching of tx history is disabled/enabled on mobile with the same setting as message history.

hesterbruikman:
🙌 great! Thanks @rasom 

I think this needs minor copy changes to accurately represent the mobile data switch. What do you think of below @errorists @rasom This is the fastest route I can think of to include in this PR. You might want to better organize the sync settings though. If so I'll create a separate issue

_Note that white line in the bottom sheet is weird_ 🤔

Frame 1

rasom:
@hesterbruikman
translations are fixed (though EN only)


> You might want to better organize the sync settings though. If so I'll create a separate issue
 
We could have two options there: 

1) keep syncing both messages and transactions
2) keep syncing only messages

As far as I understand currently wallet requires much more traffic than messages (@cammellos @oskarth do you confirm?). There are many users who use the app as a messenger in the first place, not as a wallet. For them it might make sense to keep syncing only messages history. And we can also add a button in wallet which will refresh history on demand without listening to the network (thus without using mobile data constantly).

hesterbruikman:
> And we can also add a button in wallet which will refresh history on demand without listening to the network (thus without using mobile data constantly).

If you're feeling adventurous, this is a long outstanding issue to reduce frequency of fetching and include fetching on demand: https://github.com/status-im/status-react/issues/10242 (pull to refresh). Having this manual fetching, I think we could keep the mobile data setting a catch all `mobile sync` for messaging and tx

Mobile data
- Enable - _auto-sync messages and tx when on mobile network, pull to refresh available in addition_
- Disable - does not sync messages and tx prompt settings bottom sheet when:_
   - _switching to mobile network_
   - _attempting pull to refresh_
- Data saver - _syncs tx on pull to refresh only_


Potential addition to the original design of pull to refresh could be:
If user pulls to refresh while on mobile data, UI could prompt the `Sync using mobile data?` currently used for messaging

rasom:
@hesterbruikman I would merge this PR at first and then take a look at #10242  later

hesterbruikman:
Sounds good to me. Also without #10242 being able to switch off sync over mobile data for messages _and_ tx is a win if you want to spare your data plan. Let's consider `datasaver` as a separate feature

churik:
@rasom 
### ISSUE / QUESTION 1 IOS vs Android - different behavior 
**Steps:** 
1) turn off WI-FI (LTE on IOS, 4G on Android)
2) restore account with transaction history

**On IOS(LTE)**: 
- initial scanning: transaction history is not loaded, usage: 1,8 MB on current PR build vs 1,7 MB on release build
- foreground for 5 mins: 0,4 MB on current PR build vs 11,3 MB on release build

**on Android(4G+)**:
- initial scanning: transaction history is loaded very quickly, usage: 12,9 MB on current PR build VS  13,2 on release build
- foreground for 5 mins: 7.1 MB on current PR build vs 5,3 MB on release build
So not sure that feature is working on Android (note: automatic lock is turned on on Android, so can be that app was on background for a couple of moments)

Logs from PR build, Android: [Status-debug-logs.zip](https://github.com/status-im/status-react/files/5063311/Status-debug-logs.zip)



### ISSUE 2 On IOS when creating account on mobile network after enabling wifi transaction history is not loaded until app restart (also in nightly)

As it is an issue in nightly as well, can be reported separately, so it is in case you can take a look in this PR into this.
**Steps:** 
1) turn off WI-FI (LTE on IOS)
2) restore account with transaction history
3) enable wi-fi

**Expected result:** transaction history is loaded
**Actual result:** transaction history is not  loaded until app restart
**Log:** [status_logs.zip](https://github.com/status-im/status-react/files/5063404/status_logs.zip)

rasom:
> ISSUE 2 On IOS 

is issue 2 iOS specific?

churik:
It is not reproducible on Android, only on IOS. Tried on IOS 13.5.1, however, can try on another IOS versions if it makes sense

rasom:
@churik both issues should be fixed now

churik:
**Android:** 
Initial scanning: 158 KB on current PR build VS 13,2 MB on release build
Foreground 5 min (with coming back to background and back): 409 KB (2 messages in 1-1 chat were received) on current PR build vs 5,3 MB on release build
**IOS:**
Initial scanning: 150 KB on  current PR build VS 1,7 MB on release build
Foreground 5 min (with coming back to background and back):  100 KB vs 11,3 MB on release build

Messaging in 1-1 chats is still working while on 3G network.
Checked, that transactions are downloaded when turning in  WI-Fi or start syncing and disabled again when turning it off.

GREAT WORK!

PR #11019, opened 2020-07-30
nix: allow XCode versions 11.5 and higher

jakubgs:
This is kinda hacky, but is simple enough since the version check uses grep.

For details see:
https://github.com/status-im/nixpkgs/blob/status-mods/pkgs/development/mobile/xcodeenv/compose-xcodewrapper.nix

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -32,9 +32,12 @@ in {
   gradlePropParser = callPackage ./tools/gradlePropParser.nix { };
 
   # Package version adjustments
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

jakubgs:
Ah damn, I can't use a regex like that because it's also included in the package name.

jakubgs:
I updated the `xcodeWrapper` by putting our own version at `nix/pkgs/xcodeenv/compose-xcodewrapper.nix` and adding it to the packages overlay. It now supports the `allowHigher` argument which defaults to `false` but we set it to `true`.

I might try getting this into official Nixpkgs.

PR #11028, opened 2020-08-03
Switch to WebView version that doesn't silently grant camera permission

vkjr:
fixes #... TBD

### Summary

When page requests for permission to use camera on Android `react-native-webview` checks whether Status app has that permission and if yes - allows access. But this is wrong behaviour because if user has granted camera access to Status app that doesn't mean 3rd-party site also should get access.
Fix for such a problem described here - https://github.com/googlesamples/android-PermissionRequest and implemented in latest WebView changes.

This pr switches `status-react` to update WebView version. 

### Review notes
Switching to WebView will result in silent not granting camera\mic access. To speedup fix the UI that explicitly asks user will be added in an additional pr. 

#### Platforms
- Android

##### Functional

- web browsing

### Steps to test
- Open Status
- allow camera access (for example by initiating qr code scanning and allowing permission request)
- send link `https://fatal0.netlify.com/android/webviewvideo.html` to some chat.
- click on that link and open in Status browser
- make sure that image from camera didn't appear on the page

status: ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#v2.5.23",
     "react-native-svg": "^9.8.4",
     "react-native-touch-id": "^4.4.1",
-    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.2.3_5",
+    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.2.3_7",
     "tdigest": "^0.1.1",
     "web3-utils": "^1.2.1"
   },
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

hesterbruikman:
Tested on Galaxy S10e Android 10. Camera did not appear when navigating to https://fatal0.netlify.com/android/webviewvideo.html (expected ✅) 

Note: Browser also did not request permission to use camera. I think this is a feature out of scope of this PR that addresses the security vulnerability

churik:
@vkjr 
tested briefly on IOS13 and Andrid10:
- browser history 
- dapp transacting
- dapp access rights
- issue is fixed.

PR #11034, opened 2020-08-04
Dialog added to ask permissions for webpage

vkjr:
fixes #TBD

### Summary
Followup on #11028
On privacy and security page new toggle added "Webview camera permissions request". 
When it DISABLED and some page wants camera access - user gets notification that access was blocked.
When it ENABLED and some page wants camera access - user asked if he wants to allow access.

Screenshot 2020-09-10 at 08 52 40
Screenshot 2020-09-10 at 08 52 26

#### Platforms

* Android

Web browsing

### Steps to test

* Test 1 
  * Open Status for a new account (`webview camera permissions request` toggle should be disabled)
  * open chat that contains link https://fatal0.netlify.com/android/webviewvideo.html 
  * click the link
  * make sure app notifies you that camera access request blocked

* Test 2
  * Go to `Privacy and Security` page, enable `webview camera permission request` toggle
  * open chat that contains link https://fatal0.netlify.com/android/webviewvideo.html 
  * click the link
  * make sure app asks if you want to deny/allow access request
  * press "Deny", make sure that picture from camera didn't appear on the page

* Test 3
  * open chat that contains link https://fatal0.netlify.com/android/webviewvideo.html 
  * click the link again
  * make sure app asks if you want to deny/allow access request
  * press "Allow", make sure that picture from camera did appear on the page. If camera access wasn't yet provided by system, you will have an additional system permission request.

status: wip


flexsurfer:
@vkjr do we give permissions per domain? or for webview and all domains? I would deny any permissions for webview for now

vkjr:
@flexsurfer, as far as I understood WebView requests for permissions every time when page tries to invoke related js code, for example call `navigator.mediaDevices.getUserMedia`. So even if you gave permission once, after page reload it will ask you again.

vkjr:
@0kok0, what do you think, should we allow camera/mic access at all?

0kok0:
@vkjr if we add camera/mic permission, we have to take care protecting against  permission sniffing/bypass/spoofing/... in webview. 

As a user, i would probably expect it to work for example when using some webrtc app, but I would be positively surprised by a message like: 
"the website you have visited asked for permission to access your camera/microphone. 
By default, the status browser does not grant such permission to any website to protect your privacy. 
You can change the default behavior here (link settings/privacy and security)."

If we, by default, disallow it globally, i'd call it a security feature (tor browser does that).
We can still add it if requested by users/dapps. @hesterbruikman wdyt?

hesterbruikman:
Disable by default makes sense. 

It would add a feature in `Privacy and security` > `Privacy` 
Something like an disable/enable toggle:
`Camera access in browser` > switched to disable by default

cc @errorists

vkjr:
@hesterbruikman, maybe switch like `Allow browser request camera/mic permissions`? I think we shouldn't have a switch that automatically allows camera on ANY site.

errorists:
I believe it should always ask you for a permission without a permanent switch. _Maybe_ whitelisting domains, but even then I'd have a time limit like `Allow from XYZ for the next 24 hours`

Screenshot 2020-08-06 at 08 48 23

0kok0:
> @hesterbruikman, maybe switch like `Allow browser request camera/mic permissions`? I think we shouldn't have a switch that automatically allows camera on ANY site.

There's a misunderstanding. The settings switch is to enable a popup as sketched by @errorists in the first place.

> I believe it should always ask you for a permission without a permanent switch

Definitely. The settings switch is supposed to ask the user for "permission for being asked for permission" so to say. The default would just be to block websites that request camera/mic access to protect user privacy as like in e.g. the Tor Browser.

src/status_im/ui/screens/browser/views.cljs
```diff
@@ -102,6 +105,26 @@
        :accessibility-label :modal-chat-button}
       [icons/icon :main-icons/message]]]))
 
+(defn show-access-request [resources on-allow on-deny]
+  (utils/show-confirmation {:title               (i18n/label :t/permission-request)
```
0kok0: Consider that window.title can be reset, spoofed, might not be unique and can be reused by other pages if the permission is persistent, same goes window.location.

hesterbruikman:
@0kok0 Can you articulate the privacy risk in case camera use is not blocked by default? i.e. the risk of having `always request permission` as a default opposed to `always blocked and not able to request permission, unless granted the right to request permission`

0kok0:
> always request permission as a default opposed to always blocked and not able to request permission, unless granted the right to request permission

Securitywise, disabling permission requests completely yields a different level of security compared to asking for permission. There's no permission-request-bypass, if permissions are disabled. 

In a broader sense, the issue boils down to the idea that asking a permission is not a neutral act, but setting an agenda and attracting/distracting the users attention, not necessarily for a good reason. 

There's an extreme present with websites/services asking for permission to access my location. I, as a user, would like to not be bothered by any pizza delivery if they can access my GPS while I'm visiting their site. It's not that bad with camera access, but similar. This is especially relevant for a privacy browser, which might not be my default to do online shopping, but the browser I use to interact with Dapps. As a privacy affine user, I'd the browser that handles my tokens not to ask me for potentially invasive permissions. 

I don't like the idea of the browser just not reacting and leaving a part of the site empty when the permission request is suppressed. I'd add a small popup, informing users "a request to access your camera by this site has been blocked". 
When I use Firefox/brave or some extensions, this info off stuff being kept away from me by the software I use, is very rewarding at times. 

Also I want to make sure, I'm not arguing for rejecting all camera access requests, I'm asking to give the user the option to make an informed decision about the default behavior and their UX, if their status browser is supposed to act as every other browser or as something that passively keeps requests, popups and the like away from them.

vkjr:
So per additional discussion with @hesterbruikman this PR goes following way:
- Toggle should be added in security settings that enables camera permission requests in a webview (disabled by default)
- When page requests camera access and toggle is ENABLED - permission automatically declines and app shows a notification that it was declined
- When page requests camera access and toggle is DISABLED - app shows dialog allowing user to accept or deny permission request.

src/status_im/multiaccounts/core.cljs
```diff
@@ -103,6 +103,12 @@
              :preview-privacy? (boolean private?)
              {})))
 
+(fx/defn switch-webview-permission-requests?
+  [{:keys [db] :as cofx} enabled?]
+  (multiaccounts.update/multiaccount-update cofx
```
rasom: can we move `cofx` to the next line, that would help to keep it in 80 chars range

src/status_im/ui/screens/browser/views.cljs
```diff
@@ -103,12 +105,65 @@
...
+    [react/text {:style styles/blocked-access-text}
+     (str url " " (i18n/label :t/page-would-like-to-use-camera))]]
+   [react/view styles/blocked-access-buttons-container
+    [react/view styles/blocked-access-button-wrapper
+     [quo/button
+      {:theme    :positive
+       :style    styles/blocked-access-button
+       :on-press (fn []
+                   (components.permissions/request-permissions
+                    {:permissions (map #(get resources-to-permissions-map %) resources)
```
rasom: `resources-to-permissions-map` is never `nil` so it safe to go with  `(map resources-to-permissions-map resources)`

src/status_im/ui/screens/privacy_and_security_settings/views.cljs
```diff
@@ -63,6 +63,15 @@
                       :on-press                #(re-frame/dispatch
                                                  [:multiaccounts.ui/preview-privacy-mode-switched
                                                   ((complement boolean) preview-privacy?)])}]
+      [quo/list-item {:size                    :small
+                      :title                   (i18n/label :t/webview-camera-permission-requests)
+                      :active                  webview-allow-permission-requests?
+                      :accessory               :switch
+                      :subtitle                (i18n/label :t/webview-camera-permission-requests-subtitle)
+                      :subtitle-max-lines 2
```
rasom: map alignment

rasom (approved):

churik:
@vkjr what about IOS?
IOS build also has the same setting in Profile, but it does nothing.
If it will be implemented separately, IMO better to hide it to avoid confusion otherwise to make it similar to Android.

vkjr:
@churik, thanks, you are absolutely right! It won't be implemented separately since on ios there is on permission requests from webview. So we definitely need to hide it from settings!

vkjr:
@churik, it is hidden now for ios

churik:
Tested on Android 10, Android 8:
- setting by default is disabled
- website is asking every time if you enable permission when open chat (from browser history, from chat, by entering URL)
- denying access
- allowing access
- 'Webview camera permission requests' is preserved after relogin
- no 'Webview camera permission requests' on IOS
- camera is disbled on IOS

0kok0:
Awesome work @vkjr

PR #11039, opened 2020-08-05
Show gfycat in new chat search

### LINKED ISSUE #11027: Add the 3w Anonymous Name if inserting a chat key in New Chat dialogue
# Problem

After inserting a chat key in the input, the subtitle hint is only disclosing the chat key you just entered, unlike with ENS names where it's both the name and its chat key. It would be more intuitive for people here if it showed the 3w anonymous name of the contact next to the chat key.

## Implementation

so it's like this

IMG_1599 2

---

… and instead should be like this

Screenshot 2020-08-03 at 10 13 45

## Acceptance Criteria

this small fix is made and tested

thanks!

Ferossgp:
fixes #11027

### DIFF
--- a/src/status_im/ui/screens/add_new/new_chat/views.cljs
+++ b/src/status_im/ui/screens/add_new/new_chat/views.cljs
@@ -6,6 +6,7 @@
             [status-im.ui.components.colors :as colors]
             [status-im.ui.components.icons.vector-icons :as vector-icons]
             [quo.core :as quo]
+            [status-im.utils.gfycat.core :as gfycat]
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

hesterbruikman:
Awesome @Ferossgp  👌 

Tested on Android 10 (Samsung Galaxy s10e)

1. Create account
2. Copy chat key
3. Create new account
4. Plus button > Start new chat
5. Paste chat key > see random name

Screen_Recording_20200805-175331_1

PR #11050, opened 2020-08-06
Use topbars from Quo

Ferossgp:


src/status_im/ui/screens/chat/views.cljs
```diff
@@ -32,20 +32,16 @@
...
-                   #(re-frame/dispatch [:navigate-to :home])}
-    :accessories  [{:icon                :main-icons/more
-                    :accessibility-label :chat-menu-button
-                    :handler
-                    #(re-frame/dispatch [:bottom-sheet/show-sheet
-                                         {:content (fn []
-                                                     [sheets/actions current-chat])
-                                          :height  256}])}]}])
+   {:content           [toolbar-content/toolbar-content-view]
+    :border-bottom     true
```
flexsurfer: do we have any statistics or info from @errorists what is default , with border or without ?

flexsurfer (approved):

src/status_im/ui/screens/chat/views.cljs
errorists: without border is mostly only used in Onboarding, you know the kind of views where we didn't expect the content to scroll. Having the border on is a sensible default 👍

errorists: it's mostly like so, 
Regular Top Bar:
• with border (default)
• without border
Extended Top Bar
• with the minimised state / shadow on scroll

flexsurfer: thanks, so it seems its better to have a border by default, and have a hide-border property

Serhy:
@Ferossgp  on Android there is a buggy behavior after tap on `Enter recipient address` when sending transaction. After that all bottom sheets for sending transaction disappear and I can't put focus in any input in app later, the 'Send' button in Wallet is not responding on taps and few other buttons are not working. 
Exactly like e2e tests failed with: e.g. https://app.saucelabs.com/tests/17a25e22f77a4974aa6f9f53ac644b04?auth=0a3139449e834d46b7172934edc4d9de#2

To reproduce:
1) Create new account
2 From Chats view tap Wallet
3) Tap Wallet -> `Send TX` button
4) Tap `Choose recipient` and tap `Enter recipient button`

It's all fine when tapping from `Choose recipient` the `Accounts`/`Scan QR code`

Ferossgp:
@Serhy fixed

Serhy:
Tested with Andriod 8.1 and iOS, - thanks @Ferossgp let's merged. Please mind my commits for fixed e2e

PR #11055, opened 2020-08-07
Resolve contract from ens name

Ferossgp:
Use ENS name for the acquisition contract. So it can be switched or removed dynamically without store release. Also checks for referrals after login, as the network is not available before and we cannot fetch the ens name

### DIFF
--- a/src/status_im/acquisition/core.cljs
+++ b/src/status_im/acquisition/core.cljs
@@ -1,6 +1,10 @@
 (ns status-im.acquisition.core
   (:require [re-frame.core :as re-frame]
+            [clojure.string :as string]
             [status-im.utils.fx :as fx]
+            [status-im.ethereum.core :as ethereum]
+            [status-im.ethereum.ens :as ens]
+            [status-im.ethereum.contracts :as contracts]
             [status-im.acquisition.chat :as chat]
             [status-im.acquisition.advertiser :as advertiser]
             [status-im.acquisition.persistance :as persistence]
@@ -84,12 +88,23 @@
                                                   (when-not (nil? tx)
                                                     (re-frame/dispatch [::add-tx-watcher tx])))})))))
 
-(fx/defn create
-  {}
-  [_]
-  {::get-referrer nil})
+(re-frame/reg-fx
+ ::resolve-contract
+ (fn [{:keys [chain contract on-success]}]
+   (let [register (get ens/ens-registries chain)]
+     (when contract
+       (if (string/starts-with? contract "0x")
+         (on-success contract)
+         (ens/resolver register contract on-success))))))
 
-(fx/defn login
-  {}
-  [_]
-  {::check-referrer nil})
+(fx/defn create [{:keys [db]}]
+  {::resolve-contract {:chain      (ethereum/chain-keyword db)
+                       :contract   (contracts/get-address db :status/acquisition)
+                       :on-success #(re-frame/dispatch [:set-in [:acquisition :contract] %])}
+   ::get-referrer     nil})
+
+(fx/defn login [{:keys [db]}]
+  {::resolve-contract {:chain      (ethereum/chain-keyword db)
+                       :contract   (contracts/get-address db :status/acquisition)
+                       :on-success #(re-frame/dispatch [:set-in [:acquisition :contract] %])}
+   ::check-referrer   nil})
--- a/src/status_im/ui/components/invite/events.cljs
+++ b/src/status_im/ui/components/invite/events.cljs
@@ -2,8 +2,8 @@
   (:require [re-frame.core :as re-frame]
             [reagent.ratom :refer [make-reaction]]
             [status-im.utils.fx :as fx]
+            [status-im.utils.config :as config]
             [status-im.ethereum.json-rpc :as json-rpc]
-            [status-im.ethereum.contracts :as contracts]
             [status-im.ethereum.core :as ethereum]
             [status-im.ui.components.react :as react]
             [status-im.navigation :as navigation]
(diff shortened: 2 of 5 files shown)


hesterbruikman:
@Ferossgp @Serhy Adding expected behavior here as guidance for testing

### Expected behavior
When ENS name in backend service is empty:
- `Earn crypto by inviting friends` is not visible and tap `Invite friends` shows regular share dialog, no info sheet referral program. This content is hidden on:
     - Chat view info area
     - Plus button `Invite friends`
     - `Contacts` >  `Invite friends`
- When using a friend invite url, regular 1:1 chat with friend opens. No dialog regarding referral program (Accept T&C to release Starter Pack and attribute friend) is visible
- When using a advertiser referral url to onboard, dialog with request to attribute advertiser/influencer is visible right after welcome screen

hesterbruikman:
Check ahead of release 1.6 which is to include advertiser flow for referral program only (based on new contract) plus only high-prio fixes that would otherwise go into a `1.5.1`. Therefore it is likely based of `1.5` instead of `develop`.

Does this PR include any dependencies on `develop` that are not in [1.5 branch](https://github.com/status-im/status-react/commits/release/1.5.x)?

Ferossgp:
> When using a advertiser referral url to onboard, dialog with request to attribute advertiser/influencer is visible right after welcome screen

If we want that behavior then it means the feature should be toggle from backend, ENS can be used as a kill switch to all program cc @andremedeiros

src/status_im/acquisition/core.cljs
```diff
@@ -65,7 +69,8 @@
 (fx/defn has-referrer
   {:events [::has-referrer]}
   [{:keys [db] :as cofx} flow-state referrer]
-  (when referrer
+  (when (and referrer
+             (get-in db [:acquisition :contract]))
```
Ferossgp: Remove this line, we should allow advertiser popup even if the contract is disabled as the advertiser may be without reward

src/status_im/ui/components/invite/events.cljs
```diff
@@ -164,17 +165,23 @@
...
-                           :method     "getDefaultPack()"
-                           :outputs    ["address" "uint256" "address[]" "uint256[]" "uint256[]"]
-                           :on-success #(re-frame/dispatch [::starter-pack-amount (vec %) (prn %)])}]}))
+  [{:keys [db]} contract]
+  {::json-rpc/eth-call [{:contract   contract
+                         :method     "getDefaultPack()"
+                         :outputs    ["address" "uint256" "address[]" "uint256[]" "uint256[]"]
+                         :on-success #(re-frame/dispatch [::starter-pack-amount (vec %) (prn %)])}]})
 
 (re-frame/reg-sub-raw
```
flexsurfer, line 160: could you please elaborate on this? why would you want to have raw subs and dispatch from them

Ferossgp: This is a nice approach to use just subscription for external data, this way we do not need to trigger events on some actions but they will be triggered when the data is requested by the subscription. There is a similar example in the docs https://github.com/day8/re-frame/blob/master/docs/Subscribing-To-External-Data.md#some-code

src/status_im/ui/components/invite/views.cljs
```diff
@@ -248,7 +247,7 @@
             (i18n/label :t/invite-reward {:value  (str (get reward :eth-amount) " ETH")})]])]])))
 
 (defn list-item [{:keys [accessibility-label]}]
-  (if-not config/referrals-invite-enabled?
+  (if-not @(re-frame/subscribe [::events/enabled])
```
flexsurfer: events looks too general, probably `::invite.events` could work better

src/status_im/ui/components/invite/events.cljs
```diff
@@ -164,17 +165,23 @@
...
-   (make-reaction
-    (fn []
-      (get-in @db [:acquisition :starter-pack :pack])))))
+   (let [contract (get-in @db [:acquisition :contract])]
+     (re-frame/dispatch [::starter-pack contract])
+     (make-reaction
+      (fn []
+        (get-in @db [:acquisition :starter-pack :pack]))))))
+
+(re-frame/reg-sub
```
flexsurfer, line 173: we still don't have a decision on moving subs from subs.cljs

src/status_im/ui/components/invite/events.cljs
flexsurfer: i'm not sure i understand the reason why you can't dispatch event in the place where you set value in app-db ?  i mean the cool thing about re-frame its simple, and predictable, this one just looks like can be done in "regular" way but just adds one more case for devs to know an remember , not sure if it worth it

flexsurfer: for me it looks like unpredictable thing, i don't expect that event will be dispatched when i use subscription, instead i'll be looking to component-will-mount

flexsurfer: probably if we really want to use this approach at least we could give a names for such subscriptions, like `::get-starter-pack-and-dispatch`

Ferossgp: Well, that's described in re-frame documentation, so it is correct in terms of the re-frame flow. The problem with the dispatching event is that we show it in 3 places, and we need then to dispatch it every time before showing any of these screens. Also as a user of API, you should no care from where this data is resolved. Mostly you will look at this subscription as "give me the starter pack" and the subscription will give you the value. The underline implementation makes sense only if you want to change it. But as this event doesn't involve data mutation it can be interpreted as a pure pulling of data.

Dispatch actually is not needed here at all, I can call the fx directly, the dispatch call is here because firstly it was implemented with dispatch during navigation to the invite screen, then refactored to raw when it became needed in other places. So I think I can remove the dispatch by replacing it with the fx call directly, and keep the name as it is.

src/status_im/ui/components/invite/events.cljs
Ferossgp: Yes, but I wanted to keep the referral program code fully isolated from the app, as we do not know how it will evolve later, and how much time it will be there.

flexsurfer (approved):

Serhy:
Thanks you, @Ferossgp 
Lets merge it

PR #11070, opened 2020-08-11
[mentions] Replace mentions with public keys before sending message

rasom:
Now you can try to type `"hey @user"` or something in chat's input (in case if user's ens/name was already resolved in the app, which means you either have this user in contacts or received their message in group chats). Mention should be replaced via public key then and appear in the message bubble as expected. No suggestions yet.  

Mentions of unresolved ENS usernames won't be parsed. 

status: ready

### DIFF
--- a/src/status_im/chat/models/mentions.cljs
+++ b/src/status_im/chat/models/mentions.cljs
@@ -1,0 +1,123 @@
+(ns status-im.chat.models.mentions
+  (:require [clojure.string :as string]
+            [status-im.contact.db :as contact.db]))
+
+(defn get-mentionable-users
+  [{{:keys          [current-chat-id messages]
+     :contacts/keys [contacts] :as db} :db}]
+  (let [{:keys [group-chat public?] :as chat}
+        (get-in db [:chats current-chat-id])
+        chat-specific-suggestions
+        (cond
+          (and group-chat (not public?))
+          (let [{:keys [public-key]} (:multiaccount db)
+                all-contacts (:contacts/contacts db)
+                group-contacts
+                (contact.db/get-all-contacts-in-group-chat
+                 (disj (:contacts chat) public-key)
+                 nil
+                 all-contacts
+                 nil)]
+            (reduce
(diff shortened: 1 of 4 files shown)


src/status_im/chat/models/mentions.cljs
```diff
@@ -0,0 +1,121 @@
...
+           (let [new-text (string/join
+                           [(subs text 0 (inc mention-key-idx))
+                            public-key
+                            (subs text (+ (inc mention-key-idx)
+                                          (count alias)))])
+                 mention-end (+ (inc mention-key-idx) (count public-key))]
+             (recur new-text users mention-end))))))))
+
+(defn check-mentions [cofx text]
+  (replace-mentions text (get-mentionable-users cofx)))
```
flexsurfer: probably we shouldn't call `get-mentionable-users` at all if there are no mentions in the message

rasom: There are two ways to know whether there are mentions in the message:
1. Count `@` entries on text changes 
2. Check for `@` after pressing send button

The first one might require a bit complicated logic because of possibility of copy/pasting/cutting and should be actually implemented only if we find out that the second one is not sufficient enough. Currently we check an entire message for `@` and if it is not found we stop checking.

flexsurfer: yeah but i meant `get-mentionable-users` it will be called anyway , if there are no mentions at all, so i suggest something like
```
(defn check-mentions [cofx text]
  (if (string/index-of text "@" 0)
     (replace-mentions text (get-mentionable-users cofx)))
     text)
```
or move `get-mentionable-users ` inside `replace-mentions` 

```
(if-not mention-key-idx
       text
       (let [users (get-mentionable-users cofx)
               {:keys [public-key alias]} (match-mention text users mention-key-idx) ]
```

rasom: ok I see, I'll just wrap it into a function

rasom: done

flexsurfer (approved):

PR #11078, opened 2020-08-12
WIP: Improve list performance

Ferossgp:


Serhy:
With the latest commit (build https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-200824-073832-afe922-pr-universal.apk), performance on Assets view is a little bit better than in Release 1.6 build not not much. 

When scrolling from top to bottom unlike to the Release/develop build, - in the PR build there is a blank space (while in Release build scrolling lags because of loading tokens in the list). 
ууууу2

Selecting (Check/Uncheck) an Asset takes the same time from appearing/disappearing of the checkmark after the tap like in Release build.

PR #11082, opened 2020-08-12
Subheader generation performance improved

vkjr:
fixes # TBD

### Summary

Length calculation in message preview was incorrect and as a result huge preview texts could be generated from huge messages. That wasn't visible in UI due to visual text truncation but affected performance on Android. Now that fixed and short preview components generated.

#### Platforms
- Android
- iOS

Last message preview in a chat list

##### Functional

- 1-1 chats
- public chats
- group chats

##### Non-functional

### Steps to test
- Open Status
- Open chat and post huge message
- Get back to chats preview
- Make sure there is no performance penalty when opening chat 

status: ready

### DIFF
--- a/src/status_im/ui/screens/home/views/inner_item.cljs
+++ b/src/status_im/ui/screens/home/views/inner_item.cljs
@@ -20,53 +20,59 @@
   (letsubs [contact-name [:contacts/contact-name-by-identity from]]
     contact-name))
 
-(defn render-subheader-inline [acc {:keys [type destination literal children]}]
-  (case type
-    "paragraph"
-    (conj acc (reduce
-               (fn [acc e] (render-subheader-inline acc e))
-               [react/text-class " "]
-               children))
+;; if truncated subheader text is too short we won't get ellipsize at the end of text
+(def max-subheader-length 100)
 
-    "blockquote"
-    (conj acc (.substring literal 0 (dec (.-length literal))))
+(defn truncate-literal [literal]
+  (let [size (min max-subheader-length (.-length literal))]
+    {:components (.substring literal 0 size)
+     :length     size}))
 
-    "codeblock"
-    (conj acc (.substring literal 0 (dec (.-length literal))))
+(defn add-parsed-to-subheader [acc {:keys [type destination literal children]}]
+  (let [result (case type
+                 "paragraph"
+                 (reduce
+                  (fn [{:keys [_ length] :as acc-paragraph} parsed-child]
+                    (if (>= length max-subheader-length)
+                      (reduced acc-paragraph)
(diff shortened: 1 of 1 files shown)


src/status_im/ui/screens/home/views/inner_item.cljs
```diff
@@ -13,60 +13,47 @@
...
 
-    "blockquote"
-    (conj acc (.substring literal 0 (dec (.-length literal))))
+                 "status-tag"
+                 (str "#" literal)
 
-    "codeblock"
-    (conj acc (.substring literal 0 (dec (.-length literal))))
+                 "link"
+                 destination
```
0kok0: How does a truncated link look, can it be opened from preview?

vkjr: It looks like this. No, it can't be opened from preview

Screenshot 2020-08-17 at 12 07 41

cammellos (changes requested):
I don't think using `generate-gfy` will cut it, it means that we won't show any ENS username for example.
A subscription should be used instead in this cases, is that causing the issue?

vkjr:
@cammellos, I tried to recursively construct nested components and calculate their length but that looked quite clumsy and calculation should rerun whenever we have update on ens name. Let me update PR and try it again to save correct names.

cammellos:
@vkjr what happens if you just keep the code as it is now, and restore `mention-element` as it was before?

src/status_im/ui/screens/home/views/inner_item.cljs
```diff
@@ -13,60 +13,47 @@
...
 
-    "mention"
-    (conj acc [react/text-class
-               [mention-element literal]])
```
cammellos, line 39: Here, if you just use this instead, and uncomment mention-element, what is the behavior?

qoqobolo:
With build [60c247c](https://github.com/status-im/status-react/commit/60c247caa128bf505c6912710b63e9205e0b80b4) next scenarios have been tested :

1. Changes in time of preview generation compared to the Release build. Tested on the case when several chats (all types) are in Home screen and contain huge messages (>10000 characters);
2. Messages with text formatting;
3. Login to multiaсcount containing several chats with huge unread messages;
4. Navigate to Home screen from other tabs;
5. Put app to background and back;
6. [This](https://docs.google.com/document/d/1XueoOwwBN9cgbahpFJmgtetTKAN7gMkmQW828-qX7GI/edit?usp=sharing) text
7. Preview of chats containing stickers, audio messages, and commands.

cammellos (approved):

vkjr:
@qoqobolo, pr is ready for testing subheader performance... again :)

Issue #10786, opened 2020-06-09
Some assets (ADX in particular case) are not shown on wallet when they are received to multiaccount until manual adding


churik:
# Bug Report

## Problem
When receiving some assets to wallet, sometimes they are not shown until manual enabling them in assets (transaction is confirmed) and reopening app.
Reproducibility 100% with ADX.

#### Expected behavior
ADX amount is shown

#### Actual behavior
ADX is not shown, amount is displayed only after (1) enabling ADX manually in assets and (2) restarting app
Video is [here]( https://drive.google.com/file/d/1z8xgWZB3h1ct6kB5JSVlTXltH4h8l921/view?usp=sharing
)

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Create empty multiaccount A
- Send some ADX from another multiaccount to multiaccount A
- Multiaccount A: wait until transaction is confirmed 

### Additional Information

- Status version: release 1.4
- Operating System: Android, iOS

#### Logs
[status_logs-4.zip](https://github.com/status-im/status-react/files/4751598/status_logs-4.zip)

churik:
Also confirmed with USDC

hesterbruikman:
Wondering if this is a bug or designed behavior. @churik do you know if other assets are added upon being received? 

If I recall correctly we only show SNT and Ether by default and don't do path scanning on importing keys (or on other occasions) to see if you have other assets. Don't know if there's a difference for the situation of receiving assets to 'existing' accounts.

cc @johnlea-quiup

churik:
@hesterbruikman 
it is an issue for sure, tried with AST and BLT - and they are shown after 3d confirmation on wallet without enabling them manually
![photo_2020-06-16 16 21 30](https://user-images.githubusercontent.com/4557972/84786768-739cfe80-afed-11ea-85df-5ec416019ab9.jpeg)

churik:
Confirmed with AMB

hesterbruikman:
Sounds like two problems, we might address separately

1. having to restart after manually enabling certain assets
2. having to add assets manually in the first place

I'd treat this GH issue as a focus on 1. For the second, I'll create a new issue in the `Import accounts` epic to enable scanning for assets on accounts. Once that's introduced in onboarding, I think it'll be more straightforward to add a way to call this method from within the wallet

hesterbruikman:
[...]

PR #11102, opened 2020-08-19
Add update gesture to wallet

dotemacs:
**CONTEXT**: I was in a chat with Ceri, @andremedeiros & @hesterbruikman while I worked on this issue: https://github.com/status-im/status-react/issues/10242

### Summary

Works on a feature as specified in https://github.com/status-im/status-react/issues/10242

Documentation change PR (review please): https://github.com/status-im/status.im/pull/xxx
(❓ **QUESTION:** What do I put above, the URL of this PR?)
### Review notes
I've posted the question below, but here it is for completeness: 

> hi, I have a question about this event :wallet.ui/pull-to-refresh.
>
> For this task, is this a sufficient/correct place to call it?
>
> Because under a single wallet view, there are Collectibles and History.
>
> So for :wallet.transactions.history/screen which populates History and for :wallet/visible-assets-with-values, which would  deal with Assets & Collectibles, should I do any additional work? And if I should, is there a way to add this data, as a stub to pre-populate it, so that it's meaningful?

#### Platforms

- Android
- iOS

##### Functional

- wallet / transactions

### Steps to test

- Go to **Wallet**, pull to refresh

### DIFF
--- a/src/status_im/ui/screens/wallet/accounts/views.cljs
+++ b/src/status_im/ui/screens/wallet/accounts/views.cljs
@@ -176,8 +176,7 @@
        [quo/animated-header
         {:extended-header   total-value
          :use-insets        true
-         :refresh              (rc/refresh-control)
-         :on-scroll-begin-drag rc/refresh-action
+         :refresh           rc/refresh-control
          :right-accessories [{:on-press            #(re-frame/dispatch
                                                      [::qr-scanner/scan-code
                                                       {:handler :wallet.send/qr-scanner-result}])
--- a/src/status_im/ui/screens/wallet/accounts/views.cljs
+++ b/src/status_im/ui/screens/wallet/accounts/views.cljs
@@ -10,6 +10,7 @@
             [status-im.ui.components.react :as react]
             [status-im.ui.screens.wallet.accounts.sheets :as sheets]
             [status-im.ui.screens.wallet.accounts.styles :as styles]
+            [status-im.ui.screens.wallet.refresh-control :as rc]
             [status-im.qr-scanner.core :as qr-scanner]
             [status-im.wallet.utils :as wallet.utils]
             [status-im.keycard.login :as keycard.login])
@@ -17,8 +18,7 @@
 
 (views/defview account-card [{:keys [name color address type] :as account}]
   (views/letsubs [currency        [:wallet/currency]
-                  portfolio-value [:account-portfolio-value address]
-                  prices-loading? [:prices-loading?]]
+                  portfolio-value [:account-portfolio-value address]]
     [react/touchable-highlight
      {:on-press            #(re-frame/dispatch [:navigate-to :wallet-account account])
       :accessibility-label (str "accountcard" name)
@@ -28,10 +28,9 @@
      [react/view {:style (styles/card color)}
       [react/view {:flex-direction :row :align-items :center :justify-content :space-between}
        [react/view {:style {:flex-direction :row}}
-        (if prices-loading?
-          [react/small-loading-indicator :colors/white-persist]
-          [react/text {:style               {:color colors/white-persist :font-weight "500"}
-                       :accessibility-label "account-total-value"} portfolio-value])
+        [react/text {:style
+                     {:color colors/white-persist :font-weight "500"}
+                     :accessibility-label "account-total-value"} portfolio-value]
         [react/text {:style {:color colors/white-transparent-persist :font-weight "500"}} (str " " (:code currency))]]
        [react/touchable-highlight
         {:on-press #(re-frame/dispatch [:show-popover
@@ -177,6 +176,8 @@
        [quo/animated-header
         {:extended-header   total-value
          :use-insets        true
+         :refresh              (rc/refresh-control)
+         :on-scroll-begin-drag rc/refresh-action
          :right-accessories [{:on-press            #(re-frame/dispatch
                                                      [::qr-scanner/scan-code
                                                       {:handler :wallet.send/qr-scanner-result}])
--- a/src/status_im/ui/screens/wallet/refresh_control.cljs
+++ b/src/status_im/ui/screens/wallet/refresh_control.cljs
@@ -1,0 +1,15 @@
+(ns status-im.ui.screens.wallet.refresh-control
+  (:require [re-frame.core :as re-frame]
+            [reagent.core :as reagent]
+            [quo.react-native :as rn]))
+
+(defn refresh-action []
+  (fn []
+    (when (false? @(re-frame/subscribe [:prices-loading?]))
+      (re-frame/dispatch [:wallet.ui/pull-to-refresh]))))
+
+(defn refresh-control
+  []
+  (fn []
+    (reagent/as-element
+     [rn/refresh-control {:refreshing @(re-frame/subscribe [:prices-loading?])}])))
--- a/src/status_im/ui/screens/wallet/refresh_control.cljs
+++ b/src/status_im/ui/screens/wallet/refresh_control.cljs
@@ -4,12 +4,10 @@
             [quo.react-native :as rn]))
 
 (defn refresh-action []
-  (fn []
-    (when (false? @(re-frame/subscribe [:prices-loading?]))
-      (re-frame/dispatch [:wallet.ui/pull-to-refresh]))))
+  (re-frame/dispatch-sync [:wallet.ui/pull-to-refresh])
+  (reagent/flush))
 
-(defn refresh-control
-  []
-  (fn []
-    (reagent/as-element
-     [rn/refresh-control {:refreshing @(re-frame/subscribe [:prices-loading?])}])))
+(defn refresh-control []
+  (reagent/as-element
+   [rn/refresh-control {:refreshing @(re-frame/subscribe [:prices-loading?])
+                        :onRefresh  refresh-action}]))
--- a/src/status_im/wallet/prices.cljs
+++ b/src/status_im/wallet/prices.cljs
@@ -6,6 +6,7 @@
             [status-im.ethereum.tokens :as tokens]
             [status-im.wallet.utils :as wallet.utils]
             [status-im.constants :as constants]
+            [status-im.ui.screens.mobile-network-settings.sheets :as sheets]
             [clojure.set :as set]
             [taoensso.timbre :as log]))
 
@@ -23,12 +24,15 @@
 (re-frame/reg-fx
  :wallet/get-prices
  (fn [{:keys [from to mainnet? success-event error-event chaos-mode?]}]
-   (prices/get-prices from
-                      to
-                      mainnet?
-                      #(re-frame/dispatch [success-event %])
-                      #(re-frame/dispatch [error-event %])
-                      chaos-mode?)))
+   (let [milliseconds 1000
+         get-prices   (fn []
+                        (prices/get-prices from
+                                           to
+                                           mainnet?
+                                           #(re-frame/dispatch [success-event %])
+                                           #(re-frame/dispatch [error-event %])
+                                           chaos-mode?))]
+     (js/setTimeout get-prices milliseconds))))
 
 (fx/defn on-update-prices-success
   {:events [::update-prices-success]}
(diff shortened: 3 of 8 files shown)


dotemacs:
Hello @hesterbruikman

I'm tackling https://github.com/status-im/status-react/issues/10242 in this PR.

I've removed the extra spinners as per our discussion [here](https://github.com/status-im/status-react/issues/10242#issuecomment-677471250).

The spinner now appears only at the very top.

I've put a demo of it here: https://share.getcloudapp.com/GGuR4Nyz 
Is this the correct/desired behaviour? Thanks

hesterbruikman:
@rasom this is a WIP related to #11007 Pull to refresh and `enable` if fetching over mobile data is disabled. Can you please take a look and answer questions @dotemacs may have?

src/status_im/ui/screens/wallet/refresh_control.cljs
```diff
@@ -0,0 +1,11 @@
...
+  (:require [re-frame.core :as re-frame]
+            [reagent.core :as reagent]
+            [quo.react-native :as rn]))
+
+(defn refresh-control
+  []
+  (let [prices-loading? @(re-frame/subscribe [:prices-loading?])]
+    (reagent/as-element
+     [rn/refresh-control {:refreshing prices-loading?
+                          :on-refresh #(re-frame/dispatch [:wallet.ui/pull-to-refresh])}])))
```
dotemacs: @rasom hi, I have a question about this event `:wallet.ui/pull-to-refresh`.

For this task, is this a sufficient/correct place to call it?

Because under a single wallet view, there are **Collectibles** and **History**.

So for `:wallet.transactions.history/screen` which populates **History** and for `:wallet/visible-assets-with-values`, which would deal with **Assets** & **Collectibles**, should I do any additional work? And if I should, is there a way to add this data, as a stub to pre-populate it, so that it's meaningful?

Thanks
/cc @hesterbruikman

dotemacs: For context here is the view I'm talking about:


Screenshot 2020-08-26 at 13 57 08

dotemacs:
@rasom any chance that I can get access to the Jenkins builds to see the output? Thanks

I've already authenticated against Status' via GitHub, but I'm getting the error
image


when I try to access the above **jenkins/prs/ios**. On this URL:
https://ci.status.im/job/status-react/job/prs/job/ios/job/PR-11102/9/display/redirect

Ferossgp:
@dotemacs tests are failing because there is no mock for `RefreshControl` here: https://github.com/status-im/status-react/blob/develop/src/mocks/js_dependencies.cljs#L23

Also, there are some linting problems.
Run
`make lint` - to see lint
`make test` - to run unit test

dotemacs:
Thanks @Ferossgp, I'll get right on it

dotemacs:
> @dotemacs tests are failing because there is no mock for `RefreshControl` here: https://github.com/status-im/status-react/blob/develop/src/mocks/js_dependencies.cljs#L23
> 
> Also, there are some linting problems.
> Run
> `make lint` - to see lint
> `make test` - to run unit test

@Ferossgp I've addressed all the linting issues and added the mock, thanks again.

Ferossgp:
Hey @dotemacs! The refresh is jumping up/down when you drag https://www.dropbox.com/s/z1n268agb4r5kyk/video0.mp4?dl=0

Can you explain why this happens, and how it could be fixed?

dotemacs:
> Hey @dotemacs! The refresh is jumping up/down when you drag https://www.dropbox.com/s/z1n268agb4r5kyk/video0.mp4?dl=0
> 
> Can you explain why this happens, and how it could be fixed?

Hey @Ferossgp 

Thanks for the feedback.

I did look into this and I'm not sure what could be the cause for it.

Is your question in the style of "It would be good to fix this, can you do it?" or was it more "I know what the issue is and would like you see you do it" :)  ?

I looked at the animated components in the hope that the root cause might be there.
I then used a plain `ScrollView` (non animated), but the behaviour persists.
I was hoping that some behaviour that could be handled via `on-scroll` could be helpful, but the steps I've tried, didn't yield much.

If you know what the root cause is, I'd appreciate a helpful nudge in the right direction.

Thanks

Ferossgp:
Hey! @dotemacs, It would be good to fix this, but no need for the fix in this PR.

When drag is at the refreshing position, the spinner is set to `true`, and calls the function which dispatches `:wallet.ui/pull-to-refresh`. The react-native assumes that if you call the function then it means that now the user should handle the refreshing state (the refreshing boolean). But the refreshing flag is `false` because instead of direct call and changing state the dispatch is put into the execution queue, which will be executed at the next animation frame. When the queue is at this event execution, we set the flag to `true` and update the UI. So while the event is in the queue we should keep the flag `true` otherwise the spinner is removed and we see the jump. So pseudo code will look like this:
`(when (or react-native-state our-state) show-spinner)`
1. During drag `react-native-state=true`, `our-state=false` => `show-spinner => true`
2. During the `:on-refresh` callback call `react-native-state=false`, `our-state=false` => `show-spinner => false`
3. After event execution `react-native-state=false`, `our-state=true` => `show-spinner => true`

To avoid this bug we should fix step 2. It could be done in current implementation with a combination of `dispatch-sync` and `reagent/flush`. Or by having a local state for the refreshing spinner.

There is the documentation for similar case on inputs https://day8.github.io/re-frame/FAQs/laggy-input/#question

dotemacs:
Hey @Ferossgp

Thanks for the explanation, I'll try to tackle it today!

dotemacs:
Hey @Ferossgp 

Thanks for the feedback on the bug.

At first I thought "doh", why didn't I think of it, it seems so obvious :)

But then after looking at it, and using the steps you've suggested `dispatch-sync` and `reagent/flush`, `prices-loading?` does not get updated within `on-refresh`. The resetting of `refresh-state` (local state) does get updated, but the refresh-control does not pick it up .

I think that what you've diagnosed in the first instance is the root cause, but I'm unsure on how to synchronously force the refreshing state update.


### local state approach
So if I use use the ratom to pass it to `:refreshing`, the spinner pops up so quickly that it disappears just as quickly. After checking, it looks like `refresh-state` which is passed to `:refreshing` stays `false` throughout. The ratom is updated, but `:refreshing` sees it as `false` only. As if it just stays with its original state within `:refreshing`.

### dispatch-sync approach
Using `dispatch-sync` with `reagent/flush` within `:on-refresh` does not update `prices-loading?`.
The call to this event `:prices-loading?` does occur, you can see the call to it via `re-frisk`, but it just isn't seen within `:on-refresh`.

### extra info

I've added the timeout function just to see if that would allow for the data to be updated, but it still isn't the case.

See the file in the last commit for what I'm talking about: https://github.com/status-im/status-react/blob/f549871f55c6a6a18e0cdff8127203c646681c1c/src/status_im/ui/screens/wallet/refresh_control.cljs#L12

### question

Am I doing something obviously wrong here or ...? 
Any helpful nudges would be appreciated. Thanks

dotemacs:
Hey @flexsurfer 

Since I could not find a clean way of setting `:prices-loading?` within
refresh-control's `:on-refresh`, I'm opting for this workaround, in this last commit.

The drawback is that as soon as the user pulls to refresh, the syncing
will start. Can live with that?

If we can, I'll clean up the commit messages a bit. If not, can you let me know what your thoughts are?
Thanks

dotemacs:
This is how the pull to refresh feature acts now:

https://share.getcloudapp.com/12uJQLQb

dotemacs:
To summarise what I'm asking, as I'm seeing that people are watching my video, but nobody is providing feedback 😄 (I'm talking about this: https://share.getcloudapp.com/12uJQLQb)

The idea is to provide pull to refresh feature for the user for the wallet tab.
I did that.

But there was a jitter, when you did the pulling, it would "bounce". @Ferossgp provided pointers on how to fix it. But those steps, of using `dispatch-sync` within refresh-control's `on-refresh` were suffering from the same issue.

So as a workaround, I implemented the pull to refresh feature to be initiated from scroll-view's `on-scroll-begin-drag`.

Also a note, I've added a `setTimout` of 1 second to `:wallet/get-prices`, so that the spinner would not disappear too quickly, because the remote call would often be under a second.  

### Questions

I'm looking for feedback on the following:

- How do you feel about the pull to refresh feature being initiated the way I've done it, described above?
- How do you feel about the "delay" of 1 second for `:wallet/get-prices`, described above?
- If you're not happy with this workaround, what alternative approach do you suggest?

hesterbruikman:
Thanks @dotemacs We'll need some time to organize a code review. That said you can consider the assignment completed. I'll move the conversation back to PeopleOps. Hopefully we can follow up to merge this solution soon.

dotemacs:
> Thanks @dotemacs We'll need some time to organize a code review. That said you can consider the assignment completed. I'll move the conversation back to PeopleOps. Hopefully we can follow up to merge this solution soon.

Hey @hesterbruikman 

Thanks for the confirmation.

If you're happy with the implementation of the feature, great.

I'll still need to:
- rebase the latest `develop` branch on top of this branch (as there is a conflict with `src/status_im/ui/screens/wallet/accounts/views.cljs`)
- remove the commit before last

But I just want to hear back from one of the devs if my approach is OK first.

Thanks again

3esmit:
@hesterbruikman currently it's frustrating that I can't force the update of the wallet. I received the asset in my wallet and it's not showing, it appears on etherscan but in Status not. This PR seems like it would fix this issue.

src/status_im/ui/screens/wallet/refresh_control.cljs
```diff
@@ -4,12 +4,10 @@
             [quo.react-native :as rn]))
 
 (defn refresh-action []
-  (fn []
-    (when (false? @(re-frame/subscribe [:prices-loading?]))
-      (re-frame/dispatch [:wallet.ui/pull-to-refresh]))))
+  (re-frame/dispatch-sync [:wallet.ui/pull-to-refresh])
+  (reagent/flush))
```
dotemacs, line 8: @Ferossgp 

I've tried adding/using `flush` everywhere, but not within this function.

I take it this resolves the issue when pulling to refresh that a slight "jump" occurs?
(I wouldn't ask, but getting the immutable nix build without a failure is not happening...)

Ferossgp: Sorry that it took so long, I was on vacation. Yes, this resolves the jump, as it forces the synchronizing state to react immediately after setting the new value, without waiting for the next animation frame.

dotemacs: No problem about the time lag. I was told you were on holiday...

Yea, I figured what it does, but what I found annoying is that I put the `flush` in various where I called the pull to refresh and it didn't prevent the jump. What was your reasoning for adding it in this function? Was it trial & error? Thanks

Ferossgp: I've put it there because it is needed to call flush immediately after setting the value into DB. Very strange that it hasn't worked, maybe some other executions were in between thus postponing the call of flush

dotemacs: Possibly. Thanks for the explanation

hesterbruikman (approved):
Works like a charm. No jitter and update of _fiat value_ on pull to refresh. Tested with VEF and USD on Android 10 and iOS. I'll create a follow up issue to capture pulling account balance by watching address changes on validators or through a gateway. Which I understand to be much more complex

churik:
Doesn't help with issues like https://github.com/status-im/status-react/issues/10786 or when some assets amount wasn't fetched when you come from offline - online.
For updating currency conversion rates works absolutely stunning.
So tested:
- update gesture itself on IOS / Android
- update after going from offline
- update gesture when on mobile network (it opens popup with settings for mobile, so looks good)

The only issue that I noticed: endless spinner when trying to use update gesture when offline (on IOS).
Would be nice to fix it.
Great work @dotemacs !
Sorry fore delay in testing this time.

churik:
@hesterbruikman @cammellos 
discussed with @rasom - IMO as we have some problems sometimes with fetching assets after coming back from offline or https://github.com/status-im/status-react/issues/10786 in this PR would be great when updating wallet balance also check for new assets on the wallet.
In that case, we can close  #10786 and it will resolve other weirdness that happens sometimes with balance when it is not fetched.
WDYT?

hesterbruikman:
I agree. Ideally the gesture update includes:
- [ ] Fiat value update
- [ ] Wallet balance update
- [ ] Check address for new assets and add those not yet added manually

That said, I'm not sure if it becomes a problem if this PR remains open for much longer. We could merge this and create separate issues for 2 and 3. They should then be in the Next column and eventually in a single release. In it's current state, the update gesture can be somewhat misleading

churik:
@hesterbruikman @cammellos except of https://github.com/status-im/status-react/pull/11102#issuecomment-700821382 didn't find issues in PR, don't think that it is a blocker.
After merging this will create issue https://github.com/status-im/status-react/pull/11102#issuecomment-701386018 - IMO we don't have issues with 3, but point 3 is needed for sure.

hesterbruikman:
> 
> 
> @hesterbruikman @cammellos except of [#11102 (comment)](https://github.com/status-im/status-react/pull/11102#issuecomment-700821382) didn't find issues in PR, don't think that it is a blocker.
> After merging this will create issue [#11102 (comment)](https://github.com/status-im/status-react/pull/11102#issuecomment-701386018) - IMO we don't have issues with 3, but point 3 is needed for sure.

Sounds good, thanks @churik When you've created the issue I'll move it to Next on the board so we don't loose sight. Upon release we should aim to include all updates in the single gesture

src/status_im/wallet/prices.cljs
```diff
@@ -47,27 +48,32 @@
...
-        :to            [(:code currency)]
-        :mainnet?      mainnet?
-        :success-event ::update-prices-success
-        :error-event   ::update-prices-fail
-        :chaos-mode?   chaos-mode?}
+    (if (and (= type "cellular")
+             (not syncing-on-mobile-network?))
+      (do (re-frame/dispatch [:bottom-sheet/show-sheet
+                              {:content sheets/settings-sheet}])
+          {:db (assoc db :prices-loading? false)})
```
flexsurfer, line 63: why not `:dispatch` ?

src/status_im/wallet/prices.cljs
```diff
@@ -47,27 +48,32 @@
...
-       {:from          (if mainnet?
-                         (conj tokens "ETH")
-                         [(-> (tokens/native-currency chain)
-                              (wallet.utils/exchange-symbol))])
-        :to            [(:code currency)]
-        :mainnet?      mainnet?
-        :success-event ::update-prices-success
-        :error-event   ::update-prices-fail
-        :chaos-mode?   chaos-mode?}
+    (if (and (= type "cellular")
```
flexsurfer, line 59: how's this related to this PR ?

src/status_im/ui/screens/wallet/accounts/views.cljs
```diff
@@ -28,10 +28,9 @@
      [react/view {:style (styles/card color)}
       [react/view {:flex-direction :row :align-items :center :justify-content :space-between}
        [react/view {:style {:flex-direction :row}}
-        (if prices-loading?
-          [react/small-loading-indicator :colors/white-persist]
-          [react/text {:style               {:color colors/white-persist :font-weight "500"}
-                       :accessibility-label "account-total-value"} portfolio-value])
+        [react/text {:style
+                     {:color colors/white-persist :font-weight "500"}
+                     :accessibility-label "account-total-value"} portfolio-value]
```
flexsurfer, line 33: `:account-total-value` keyword

dotemacs:
@flexsurfer 

I'm not sure who the comments are aimed at. 

I'm no longer working on this PR. 

Thanks

PR #11109, opened 2020-08-21
Use different acquisition gateway based on chain

Ferossgp:


flexsurfer (approved):

Serhy:
Looks good to me! With Android 8 and Android 9 tested the Advertiser flow: new account, seedphrase recover, keycard account with default invite_code used for advertiser link as well as custom one - in all cases 'Accept'ing the pop-up gets tracked and reflected by incremented 'attribution' counter. Clicks count also tracked.
Thanks @Ferossgp lets merge it!

Issue #24, opened 2016-04-25
Black 'hit' states on button


0xc1c4da:
When items are tapped the hit state is black?

https://assembly-vc.slack.com/files/jarradhope/F135ZQW2E/uglytappress.png
https://assembly-vc.slack.com/files/jarradhope/F135CPEQ4/moreuglytappress.png

PR #11122, opened 2020-08-31
Mentions suggestions

rasom:
status: ready

### DIFF
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -118,6 +118,8 @@
 (reg-root-key-sub :chat/memberships :chat/memberships)
 (reg-root-key-sub :camera-roll-photos :camera-roll-photos)
 (reg-root-key-sub :group-chat/invitations :group-chat/invitations)
+(reg-root-key-sub :chats/mention-suggestions :chats/mention-suggestions)
+(reg-root-key-sub :chats/cursor :chats/cursor)
 
 ;;browser
 (reg-root-key-sub :browsers :browser/browsers)
@@ -853,6 +855,43 @@
       (-> transaction
           (wallet.db/get-confirmations current-block)
           (>= transactions/confirmations-count-threshold))})))
+
+(re-frame/reg-sub
+ :chats/mentionable-contacts
+ :<- [:contacts/contacts]
+ (fn [contacts]
+   (reduce
+    (fn [acc [key {:keys [alias name identicon]}]]
+      (if (and alias (not= alias ""))
+        (let [name (string/replace name ".stateofus.eth" "")]
+          (assoc acc alias {:alias      alias
+                            :name       (or name alias)
+                            :identicon  identicon
+                            :public-key key}))
+        acc))
+    {}
+    contacts)))
+
+(re-frame/reg-sub
+ :chats/mentionable-users
+ :<- [:chats/current-chat]
+ :<- [:chats/mentionable-contacts]
+ (fn [[{:keys [users]} contacts]]
+   (merge users contacts)))
+
+(re-frame/reg-sub
+ :chat/mention-suggestions
+ :<- [:chats/current-chat-id]
+ :<- [:chats/mention-suggestions]
+ (fn [[chat-id mentions]]
+   (take 15 (get mentions chat-id))))
+
+(re-frame/reg-sub
+ :chat/cursor
+ :<- [:chats/current-chat-id]
+ :<- [:chats/cursor]
+ (fn [[chat-id cursor]]
+   (get cursor chat-id)))
 
 ;;BOOTNODES ============================================================================================================
 
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -13,7 +13,12 @@
             [status-im.utils.config :as config]
             [re-frame.core :as re-frame]
             [status-im.i18n :as i18n]
-            [clojure.string :as string]))
+            [clojure.string :as string]
+            [status-im.chat.models.mentions :as mentions]
+            [status-im.ui.components.list.views :as list]
+            [quo.components.list.item :as list-item]
+            [status-im.ui.screens.chat.styles.photos :as photo-style]
+            [reagent.core :as reagent]))
 
 (def panel->icons {:extensions :main-icons/commands
                    :images     :main-icons/photo})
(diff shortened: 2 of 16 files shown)


rasom:
@hesterbruikman @errorists can you please take a look at suggestions list and let me know what should be changed there

errorists:
@rasom sure, here goes

- [ ] The container for suggestions, the in-out animation of opacity should be faster (how faster? about ⅓ of the current timing) but going a step further a transition along the Y axis whenever it appears or changes height would give a more pleasing effect than opacity. I'd use the same animation curve we have for the bottom Tab Bar animation.
- [ ] The container with suggestions should be able to span the entire available height all the way to the Top Bar as seen in Figma [here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=144%3A3334)
- [ ] The container shouldn't overlap the chat toolbar, so the 1pt top grey border of the toolbar is visible when suggestions are shown
- [x] 8pt vertical padding on the suggestions container
- [x] List items: so those should use a custom local style rather than the component as we need them much more dense, You can view them in [Figma here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=144%3A3502)
to recap it's 8pt padding, 36pt image, 8pt margin between the image and text labels. text labels are 13pt font size, line height 18pt, title font weight 500 black, subtitle font weight 400 dark grey.
- [x] the user avatars should have the same 10% black border as they do elsewhere in chats

rasom:
thank you, @errorists 
> The container for suggestions, the in-out animation of opacity should be faster

I did not add any animation there so it is either inherited somehow or that's just how fast that container is rendered. Anyway I'm not going to work on animation in this PR because it is quite big already, that will be left for the next iteration.  

> The container with suggestions should be able to span the entire available height

It means that if we have enough users to fill that area we do it right away after entering "@", correct?

The rest of it, I'll probably add questions later during implementation.

errorists:
> It means that if we have enough users to fill that area we do it right away after entering "@", correct?

correct, that too. on observation, right now the maximum available area we devote to suggestions is a bit too small. the Y position the suggestions container can start should be right around the end of the top bar, so say on an iPhone X that would be 100 (44 for status bar + 56 top bar), basically covering the entire scrollable area of the chat.

rasom:
@errorists 
> on observation, right now the maximum available area we devote to suggestions is a bit too small

as far as I saw it is that small in other messengers and it usually doesn't cover entire chat right away. Also all that math might be a bit tricky when hiding/dismissing keyboard, as in other messengers by default that area has some fixed max height there is no need to react on keyboard events, but if we always want to fill an entire area between input and top bar...

errorists:
@rasom if you see technical reasons, like you said keyboard hide-dismiss events, then by all means I'm cool with capping it at say `4.5 x list item height` (the 0.5 is to show it's scrollable). Imo, this chat leftover that's not available to show suggestions is wasted space as in the moment you're picking the right person, the tiny window into chat context doesn't concern you.

rasom:
@errorists yep, i agree that chat leftover is useless, but fixed height would really make it a lot simpler technically. And we always can revisit it if necessary.

rasom:
@errorists updated UI on iOS

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -73,27 +77,102 @@
...
+       :on-selection-change
+       (fn [_]
+         (when cursor
+           (re-frame/dispatch [::mentions/clear-cursor])))
+
+       :on-change
+       (fn [args]
+         (let [text (.-text ^js (.-nativeEvent ^js args))]
+           (on-text-change text)
+           (when platform/ios?
```
flexsurfer, line 128: could you pls elaborate why this is only for ios ?

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -73,27 +77,102 @@
...
+              (fn []
+                (re-frame/dispatch [:chat.ui/select-mention user]))}
+
+       subtitle?
+       (assoc :subtitle alias))]))
+
+(defn autocomplete-mentions []
+  (let [suggestions @(re-frame/subscribe [:chat/mention-suggestions])]
+    (when (seq suggestions)
+      (let [height (* 52 (min 4.5 (count suggestions)))]
```
flexsurfer: why 4.5 ? will it work on small screens ? probably we could count this number depending on the screen height

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -73,27 +77,102 @@
...
+        [rn/view
```
flexsurfer, line 182: why we need extra view ?

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -73,27 +77,102 @@
...
+       (fn [_]
+         (when cursor
+           (re-frame/dispatch [::mentions/clear-cursor])))
+
+       :on-change
+       (fn [args]
+         (let [text (.-text ^js (.-nativeEvent ^js args))]
+           (on-text-change text)
+           (when platform/ios?
+             (re-frame/dispatch [::mentions/calculate-suggestions mentionable-users]))))
```
flexsurfer, line 129: i would suggest to use `debounce/debounce-and-dispatch` here, because if we'll dispatch on every symbol, input will be slow , we should calculate suggestions only when the user has stopped typing

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -73,27 +77,102 @@
...
+
+       :on-text-input
+       (fn [args]
+         (let [native-event  (.-nativeEvent ^js args)
+               text          (.-text ^js native-event)
+               previous-text (.-previousText ^js native-event)
+               range         (.-range ^js native-event)
+               start         (.-start ^js range)
+               end           (.-end ^js range)]
+           (re-frame/dispatch
```
flexsurfer, line 139: same here `debounce/debounce-and-dispatch`

src/status_im/ui/screens/chat/components/input.cljs
errorists: that's a good point, can it be percentage based? I know I said 4.5 but this might indeed overflow on smaller screens

src/status_im/ui/screens/chat/components/input.cljs
rasom: Yep, I will add a comment there, the problem is that event object contains different data on Android and iOS.

src/status_im/ui/screens/chat/components/input.cljs
rasom: > probably we could count this number depending on the screen height

we probably can also take in count screen height and keyboard height, but I would address this when it is proven to be an issue. Meanwhile I will just change it to 3.5

for 4.5 see https://github.com/status-im/status-react/pull/11122#issuecomment-685700952

src/status_im/ui/screens/chat/components/input.cljs
rasom: I will check, I actually copied this layout from previous PR and didn't touch places which worked

src/status_im/ui/screens/chat/components/input.cljs
rasom: How can I verify that input will be slow? I mean I tried on average Android and it looks just fine.

rasom: Just to clarify: in case if there is no "@" in text this is about one `if` in handler.

errorists:
thanks @rasom , found two issues

IMG_1646

- [ ] the padding is applied to the parent element clipping the nested scrollable content, it should be applied to the list instead so that it doesn't clip. 
- [ ] For list items there, we want to unify all of the titles to use the same font style regardless if there's a subtitle present or not. this is to have uniformity, otherwise cells with 3w random names 'pop out'

src/status_im/ui/screens/chat/components/input.cljs
rasom: > that's a good point, can it be percentage based?

it can be either fixed number or based on keyboard/screen height, but I'm still not convinced that this is a big problem and would revisit it after this PR. It is big and a bit complicated already.

src/status_im/ui/screens/chat/components/input.cljs
rasom: oops wrong reason, this specific condition is because of order of events. But I will add comments in all places where we have platform based conditions in this PR

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: i mean why would you want to flood the queue with events if you know you need only the latest one , so when you typing fast and when press send, your send event will be in the end of this queue of events, 400ms delay will do the job

rasom: > i mean why would you want to flood the queue with events

because I don't want to have 400ms delay unless it is really necessary? :)

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: 4.5 works fine on small device

src/status_im/ui/screens/chat/components/input.cljs
rasom: added notes

src/status_im/ui/screens/chat/components/input.cljs
rasom: then we are good for now I think, thanks @flexsurfer

flexsurfer:
just tried and not sure i understand how it works
1) opened #status channel, sent message with mention
2) opened #test channel, in mention suggestion were no participants from #test channel, only list of ens names but not from this test channel, in test channel were visible a few messages without ens names 
i would expect to see participants of test channel ?

src/status_im/ui/screens/chat/components/input.cljs
rasom: checked, with one view layout is broken on Android. Why? I have no idea. You can try it locally if you wish

rasom:
@flexsurfer The idea was to merge `:contacts/contacts` and users from chat's context, and that's what you see. The problem is though that `:contacts/contacts` contains not only actually contacts but also all resolved ENS names. If that's not ok we can change it, but doesn't look like an issue to me.

flexsurfer:
@rasom thats ok to have ens names, but im more wondering about the absence of current chat participants (users from chat's context) what does that mean btw "users from chat's context" ?

rasom:
@errorists padding and text size should be fixed in the next build

flexsurfer:
found one more bug if text input more than 1 line it cuts the suggestions list (at least i reproduced with 3 lines )

rasom:
@flexsurfer 

> thats ok to have ens names, but im more wondering about the absence of current chat participants

are you sure that those users are absent? I actually made it so that only first 10 suggestions are shown and you have to continue typing in order to see more precise suggestions. That's how it works in discord, for instance. We can render more items but I would limit that number anyway. Probably we need to add some hint like "continue typing blabla" and work on some prioritisation of suggestions. But I suppose that's out of scope here anyway.

rasom:
@flexsurfer 
> found one more bug if text input more than 1 line it cuts the suggestions list (at least i reproduced with 3 lines )

should be fixed now

rasom:
@flexsurfer @errorists 
issues are fixed, let me know if there's something left

src/status_im/subs.cljs
```diff
@@ -829,6 +830,41 @@
           (wallet.db/get-confirmations current-block)
           (>= transactions/confirmations-count-threshold))})))
 
+(re-frame/reg-sub
+ :chats/mentionable-contacts
+ :<- [:contacts/contacts]
+ (fn [contacts]
+   (reduce
+    (fn [acc [key {:keys [alias name identicon]}]]
+      (let [name (string/replace name ".stateofus.eth" "")]
```
flexsurfer: probably better to use const from `status-im.ethereum.stateofus`

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: from my experience its better to use it , better for performance, but its up to you

flexsurfer (approved):

src/status_im/ui/screens/chat/components/input.cljs
rasom: I will test it more carefully in the next PR which will be about highlighting mention in the text input and then we will decide if this is necessary.

errorists:
@rasom looks good! if I was picky I'd only ask to set the font-weight to `500` on single line list items, the same way it's on titles for double line, to unify those so text looks exactly the same.

rasom:
> font-weight to 500 on single line list items

@errorists fixed

churik:
@rasom  testing is still in progress.
Let me know which issues will be fixed here and which separately.

### ISSUE 1: No suggestions for  users without ENS if join public chat until reopening chat view
**Steps:**
1) create new account
2) join any chat
3) try to mention any user without ENS

**Expected result:**
suggestion in list
**Actual result:**
No suggestion in list (note: after going back to home view and reopening chat all 3 random names appearing in list)
ezgif com-video-to-gif

### ISSUE 2: adjust suggestions list with local usernames

###  ISSUE 3: white borders in Dark theme
photo_2020-09-07 16 51 55

###  ISSUE 4: If start chat with ENS-name for 1 time after login - error at attempt to call mentions

**Steps:**
1) create a new account
2) start chat with ENS name (i.e. with `@sergey`)
3) type @ in the input field

**Expected result:**
no error, your username and ENS name of receiver are in list
**Actual result:**
IMAGE 2020-09-07 17:09:01

###  Question 5: Max length when typing "@" without typing next symbol is 15?
Asking because didn't find any pattern in behavior - seems that list after typing '@' displays some random 15 contacts / resolved ENS / or participants (without any sorting as I see)
In fact, I have 22 contacts.
All fine when I start typing the name, i.e. '@a' - I can see contacts starting from A.

###  ISSUE 6: if navigate to any user profile  - it starts appearing in suggestions in any chat

**Steps:**
1) create a new account
2) join any public chat with message from user without ENS-name (user A)
3) tap on user icon navigate to user A profile
4) join empty public chat
5) start typing `@` + first name of user A

**Expected result:**
 User A not in the list, because he is not added to contacts
**Actua result:**
User A is shown in suggestions

###  ISSUE 7:  If several messages with mentions were sent one after the other in public or group chat,  only the first mention will be shown  in preview in the chat list (home screen).

**Steps:**

- Join public or group chat
- Send text message -> navigate to home view to check that message is shown in preview
- Open chat and send mention -> navigate to home view to check that this mention is shown in preview
- Open chat and send one more mention  -> navigate to home view to check that the previous mention is shown in preview

ezgif com-resize (8)

### ISSUE 8: If copy (or reply into) message with mention by long tap and paste it to text input, chat key of the mentioned user will be pasted instead of 3-random or ENS name

Same with reply
Steps:

- Open any chat
- Send message with mention
- Long tap to copy this message
- Paste message to the text input

ezgif com-resize (9)

rasom:
issue 1: fixed
issue 2: out of scope in this PR. It will require additions to UI and to the way how mentions are filtered and most likely to the way how mentions are shown in messages. 
issue 3: fixed
issue 4: will fix tomorrow
issue 5: yes it is limited to 15 now, if we do not limit it that list can be very long and slow. We can show some hint there, but some messenger (for instance discord) just have such limits and no hints
issue 6: will fix tomorrow. The thing is that when `:chat.ui/show-profile` is dispatched that profile is added to contacts in app-db. Why? I don't know :D
issue 7: out of scope
issue 8: out of scope

rasom:
4 is fixed
6 is also fixed but in #11138 (although this one is also out of scope actually, but wateva)

rasom:
@churik @qoqobolo 
1. fixed
2. out of scope
3. fixed 
4. fixed
5. not fixed (likely won't be fixed)
6. fixed
7. out of scope
8. out of scope

qoqobolo:
> 6 is also fixed but in [#11138](https://github.com/status-im/status-react/pull/11138) (although this one is also out of scope actually, but wateva)

@rasom thank you for fixes! Could you rebase this PR including fix 6, please?

rasom:
@churik @qoqobolo 
fyi: as discussed with @errorists in #core-ui, nicknames will not affect mentions (and mention suggestions), so that's not only out of scope but also doesn't require a separate issue

qoqobolo:
@rasom issue 6 is still reproducible on [#24](https://ci.status.im/job/status-react/job/prs/job/android/job/PR-11122/24/)

ezgif com-resize (10)

rasom:
@qoqobolo ok, at first I only fixed the case when that `:profile` screen was opened on pressing mention in the message. Should be fixed on pressing identicon and username now. Totally unrelated to this PR though, hopefully it won't brake other things :D

rasom:
issue 7 was one line thing, fixed in this PR

rasom:
issue 8 is solved here https://github.com/status-im/status-react/pull/11146

rasom:
@churik @qoqobolo 
so now
1. fixed
2. out of scope, will be added in some form after highlighted mentions in input
3. fixed 
4. fixed
5. not fixed (likely won't be fixed)
6. fixed
7. fixed
8. fixed in #11146

qoqobolo:
@rasom thanks for the fixes not even related to this PR! But unfortunately there is one more :(

#### ISSUE 9: Blank line appears in suggestions if I join a group chat as a member, or if I remove or add another user as admin (after system message). 

#### Case 1:
_Steps:_

- Create group chat
- Open group info
- Remove some member
- Navigate to the chat
- Type @ to check suggestions

ezgif com-resize (11)

#### Case 2:
_Steps:_

- Device 1: create group chat
- Device 2: join this chat and type @ to  check suggestions

ezgif com-resize (12)

rasom:
@qoqobolo issue 9 should be fixed now

churik:
@rasom 
#### ISSUE 10: crash with IndexOutOfBounds exception if suggestion lists are opened for several chats
Reproduced for me on Android.

**Steps:**
- join public chat 1 (in my case `chu`)
- join public chat 2 (in my case `chu1`)
- open `chu`, type @ in input field, go back
- open `chu1`, type @ in input field, go back
- open `chu`  again, select any value from the suggestions list, send message
- type `@`

**Expected result:**
can see list again


**Actual result:**
IMAGE 2020-09-10 14:37:54
Logs: [Status-debug-logs.zip](https://github.com/status-im/status-react/files/5201651/Status-debug-logs.zip)

churik:
#### ISSUE 11: usernames from list are not resolved after reopening chat view if they appears in the beginning of message history

**Steps:**
- join public chat 1 (in my case `chu`)
- scroll to the beginning of history
- tap back and open chat again
- try to send message with mention user from the beginning of the history

**Expected result:**
mention in message

**Actual result:**
Monosnap 2020-09-10 15-37-31

rasom:
@churik both 10 and 11 are fixed. 10 was Android specific

churik:
#### ISSUE 12: own name appears in suggestions list after getting history from public chat and sending messages to 1-1

**Steps:**
- start 1-1 chat
- join public chat 1 (in my case `chu`)
- scroll to the beginning of history (wait until history will be loaded)
- open 1-1 chat
- send message
- reopen 1-1 chat 
- start `@`


**Expected result:**
no your name in suggestions

**Actual result:**
photo_2020-09-10 19 30 57 jpeg 2020-09-10 19-31-14

rasom:
@churik discord also does it, not sure that's a big deal. If there is nothing more critical I would merge it and move to spam related stuff.

churik:
@rasom 
testing is still in progress. 
I'll let you know when it is finished, hopefully, nothing critical will be found

churik:
@rasom 
Thank you for your patience.
All mentioned but not fixed issues will be reported separately

PR #11125, opened 2020-09-01
Update nREPL to the latest stable

dotemacs:
### Summary

With the update of Emacs CIDER 0.26.1, it expects to have the latest
cider-nrepl dependency to be at 0.25.3 otherwise it throws an error
when connecting to the project's REPL.

### Platforms

Not applicable, it can only be seen on developers machines.

### Steps to test

With Emacs, install the latest CIDER mode at the time of this writing, start the project as per documentation and you should not see an error.

To see the error, downgrade `cider-nrepl` to `0.24.0`.

### DIFF
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -14,7 +14,7 @@
                 [mvxcvi/alphabase "1.0.0"]
                 ;; dev dependencies
                 [refactor-nrepl "2.5.0"]
-                [cider/cider-nrepl "0.24.0"]
+                [cider/cider-nrepl "0.25.3"]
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

dotemacs:
For the curious, this is the exact error message in the CIDER REPL, once you connect:

```
WARNING: CIDER 0.26.1 requires cider-nrepl 0.25.3, but you're currently using cider-nrepl 0.24.0. 
The version mismatch might break some functionality! (More information)
```

The "More information" link leads to a 404 page in CIDER's documentation...

PR #11128, opened 2020-09-01
[#10974] ENS resolving for IPNS names.

### LINKED ISSUE #10974: ENS resolving for IPNS names.
# Bug Report

Recently I asked @hesterbruikman why ENS resolving does not work in browser. I was actually wrong on whether or not it works. If you use for example `pactxt.eth` in the browser it resolves. However `uniswap.eth` does not.

## Problem

This is because there was an issue in the contenthash library that did not resolve names correctly when they were IPNS hashes. You can see a more detailed discussion here: https://github.com/ensdomains/ens-app/issues/849. This was recently fixed in ENS which can be viewed here: https://github.com/ensdomains/ui/pull/77.

#### Expected behavior

Names using IPNS names resolve properly in the browser.

#### Actual behavior

They currently do not.

### Reproduction

- Open Status
- Open browser
- Type `uniswap.eth` and hit enter
- Name does not resolve.

flexsurfer:
fixes #10974

IPNS names resolution
change status.im proxy on cloudflare , @jakubgs we can shut down our infura proxy after this PR will be released 
change infura ipfs to cloudflare , so we use only cloudflare for ipfs

for QA:
test stickers
test regular web2 domains
test ethereum.eth
test uniswap.eth
do we know any other eth dapps ?

also fixed 2 issues dunno if we have them registered:
open ethereum.eth -> click Learn more -> 
1) url is changed but should be the same ethereum.eth
2) on android url is shown in 3 lines instead 1


churik:
Tested on IOS 13 and Android 10:
- can install and use free stickerpacks, stickerpacks that were bought earlier
- can open web2 links through chat, QR scanner, deep links
- can open IPNS links through chat, QR scanner, deep links
- browser history is preserved
- collectibles are shown and can navigate to collectible view
- testing spoof-url case - no changes comparing to current nightly.

All issues mentioned in the description are fixed.
Briefly checked on Android 6 and Android 8

Thank you @flexsurfer

cammellos (approved):

PR #11129, opened 2020-09-01
Aligning react versions

dotemacs:
When starting the project the following would be spat out:

```
NPM dependency "react" has installed version "16.11.0"

"16.13.0" was required by jar:file:/Users/alex/.m2/repository/reagent/reagent/0.10.0/reagent-0.10.0.jar!/deps.cljs
```

I've bumped up the version in package.json to to match the one being
pulled in by reagent.

### Summary

Ensuring that the versions of `react` as defined by NPM & shadow-cljs match.

### Review notes

If any known issues exist that require `16.11.0` then this would be an issue.
But since one version is defined and the source of another used...

### Testing notes

Start the app and confirm that there are no issues.

I've tested it in the simulator, no issues seen. 

#### Platforms

- Android
- iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "hi-base32": "^0.5.0",
     "i18n-js": "^3.3.0",
     "qrcode": "^1.4.1",
-    "react": "16.11.0",
+    "react": "16.13.0",
     "react-dom": "^16.4.2",
     "react-native": "0.62.2",
     "react-native-background-timer": "^2.1.1",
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

dotemacs:
Could it be that this is a intermittent failure?

I seriously doubt that React version would be the cause of it.

flexsurfer:
thanks, not related to this PR

PR #11146, opened 2020-09-09
[mentions] Fix mention appearance in reply and quoted message

rasom:
Solves issue 8 from [this comment](https://github.com/status-im/status-react/pull/11122#issuecomment-688407561).

status: ready


cammellos (approved):

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -25,10 +26,24 @@
...
 
+(defn get-quoted-text-with-mentions [parsed-text]
+  (string/join
+   (mapv (fn [{:keys [type literal children]}]
+           (cond
+             (= type "paragraph")
+             (get-quoted-text-with-mentions children)
+
+             (= type "mention")
+             @(re-frame/subscribe [:contacts/contact-name-by-identity literal])
```
flexsurfer, line 37: that is something weird, probably you could pass `contacts` as a parameter to get-quoted-text-with-mentions?

flexsurfer (dismissed):

src/status_im/ui/screens/chat/components/reply.cljs
cammellos: Would it not then re-triggered rendering when anything in contacts is changed? 
This way we can much better use the subs cache and avoid logic in the view (if you pass contacts then you will have to run the code to fetch the contact, get the name etc at each render).
Or I misunderstand what you mean?

rasom: Well it is probably not the best way to solve it for text copying but how exactly is it weird otherwise, can you please elaborate?

flexsurfer: the function is not pure. and subscription kinda hidden

rasom: Does it mean that by inlining this code in all three places where it is used both concerns will not be valid anymore? :D 
Just to clarify this is done to avoid duplication. 

Is there any practical concern...?

flexsurfer: if you want to subscribe it should be a component , not a pure function 
```clojure
[quoted-text-with-mentions parsed-text]
```

```clojure
(defn quoted-text-with-mentions [parsed-text]
  [quo/text {:size            :small}
     (string/join)
```

rasom: Ok, so:
1) it is neither a component nor a pure function, am I ok then? :D
2) why so? This rule is based on what?
3) it won't work for text copying case

flexsurfer: 1. how so ?
2. general rule ? functions must be pure ?

rasom: 1. The function I wrote is not a component. Also it relies on a global state, which makes it impure function. That means it is neither component, nor a pure functions 
Your words:
    > if you want to subscribe it should be a component , not a pure function

    but it is not a pure function. I was rather joking though 
2. > general rule ? functions must be pure ?

    I mean, no offence, our view layer is 99% based on impure functions (components are impure functions). This `get-quoted-text-with-mentions` is an impure function called from another impure function (component) to avoid duplication. It just makes sense in that context. Probably functions must be pure but that makes a little sense when we considering our views.

PR #11148, opened 2020-09-09
Get rid of form-2 subs in defview

rasom:
We don't need form-2 subs and those subs cause bugs in `defview` when they have parameters. This PR solves it. Next step would be removing `defview` from code. 

status: ready

### DIFF
--- a/src/status_im/utils/views.clj
+++ b/src/status_im/utils/views.clj
@@ -2,9 +2,8 @@
   (:require [clojure.walk :as w]))
 
 (defn atom? [sub]
-  (or (vector? sub)
-      (and (seq sub)
-           (#{`reagent.core/atom} (first sub)))))
+  (and (seq sub)
+       (#{`reagent.core/atom} (first sub))))
 
 (defn walk-sub [sub form->sym]
   (if (coll? sub)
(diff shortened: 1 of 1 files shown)


flexsurfer:
i still think form-2 should be used, if you need to pass a parameter from another subscription, that means subscriptions was incorrectly designed

rasom:
@flexsurfer please elaborate why form-2 should be used [(just reminder)](https://github.com/day8/re-frame/issues/218)

rasom:
> that means subscriptions was incorrectly designed

We have a safe way to not care about "incorrect design" (that's kind of arguable) and bugs caused by it, so what's the reason to not use it?

flexsurfer:
yeah i know it will work, and its not critical, it's just my personal preference, however, i don't mind to follow any team's decision, and i know most devs for removing defview, so i'll approve this PR, but suggest to replace `devfiew` by `sub` function which will deref subscription

flexsurfer (approved):

rasom:
> but suggest to replace devfiew by sub function which will deref subscription

that's fine, but this PR is not about removing `defview`, it just fixes all these views with frozen state.

> and i know most devs for removing defview

I would also consider some punishment for the retard who introduced `defview` in the first place

PR #11162, opened 2020-09-14
Always use waku

### LINKED ISSUE #11092: Disable Whisper and don't allow disabling of Waku
# Description

While researching disabling of Whisper on our fleets don in https://github.com/status-im/infra-eth-cluster/issues/32#issuecomment-674950824 I found out that there is a tiny number of nodes that run most recent version(`1.5.0`) but use only Whisper without Waku. I think it's about time that we disable Whisper and remove ability to disable Waku in advanced settings:
status_advanced_settings

# Reasoning

* Supporting of Whisper will increase probability of weird bugs popping up
* Whisper has a negative effect on bandwith used on mobile devices
* Narrowing down settings to only waku will simplify Waku node setup

Ferossgp:
Fixes #11092


Serhy:
Tested with iOS 13, Android 9 and Android 9.
Messaging in 1-1/public/Group chat works, can exchange transactions, can pair devices, change network, ENS names can be set up and are visible to recipients, PN appear on both Android and iOS, no messages (and no PN) from blocked users.
Ready for merge

Serhy:
Tested with iOS 13, Android 9 and Android 9.
Messaging in 1-1/public/Group chat works, can exchange transactions, can pair devices, change network, ENS names can be set up and are visible to recipients, PN appear on both Android and iOS, no messages (and no PN) from blocked users.
In logcat it's visible Waku is set.
Ready for merge

cammellos (approved):

Issue #9484, opened 2019-11-19
Reintroduce back collapsing of long messages in chats


Serhy:
# Bug Report

## Problem

After moving message parsing to status-go (in https://github.com/status-im/status-react/pull/9409) we lost (as known issue) collapsing messages when message content has more than 20 lines. 

#### Expected behavior
Messages with more than X (variable) lines are collapsed in message bubble 

#### Actual behavior
Messages with more than X (variable) lines are collapsed in message bubble 

![123123](https://user-images.githubusercontent.com/8749671/69140816-fb5c7200-0acb-11ea-82db-79864ba00345.gif)

### Reproduction

- Open Status and create new account
- Join any chat (e.g. public chat `#test777`)
- Send message containing 200 lines

### Additional Information
- Status version: develop `0.14.0 (2019181102)`
- Operating System: Android and iOS

Serhy:
There is a https://github.com/status-im/status-react/issues/9469 task dedicated to styling of long messages but this issue is for reintroducing line-limit back.

churik:
cc @rachelhamlin

StatusSceptre:
Potentially fix together with #9485?

churik:
@rachelhamlin 
IMO they are not the same. 
The decision was (fix me if I'm wrong):
1) limit sending messages to 2000 chars (can prevent https://github.com/status-im/status-react/issues/9485 )
2) reintroduce back collapsing of long messages (it was before 20 rows)

rachelhamlin:
Is 2000 characters less than 20 rows? I guess it depends on how many characters it takes before we need to collapse messages. :)

churik:
Collapsing is more about showing the content I believe, sending is about the limitation of symbols for the sender to prevent long messages to appear, and there are several possible ways to do it - i.e. showing validation message at attempt to typing \ pasting > 2000 chars or to split content when it was sent.

So it would be nice to decide what we should do here.

rachelhamlin:
Well, let's definitely fix this issue of messages needing to be collapsed again, and keep the other one separate.

PR #11164, opened 2020-09-15
[#9484] Reintroduce back collapsing of long messages in chats

flexsurfer:
max height is 200px

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "react-native-image-resizer": "^1.2.3",
     "react-native-keychain": "git+https://github.com/status-im/react-native-keychain.git#v.3.0.0-5-status",
     "react-native-languages": "^3.0.2",
+    "react-native-linear-gradient": "^2.5.6",
     "react-native-mail": "git+https://github.com/status-im/react-native-mail.git#v4.0.0-status",
     "react-native-navigation-bar-color": "^2.0.1",
     "react-native-permissions": "^2.1.5",
--- a/src/mocks/js_dependencies.cljs
+++ b/src/mocks/js_dependencies.cljs
@@ -190,6 +190,8 @@
   #js {:localNotification identity
        :requestPermission identity})
 
+(def react-native-gradien #js {:default #js {}})
+
 (def push-notification-ios #js {})
 ;; Update i18n_resources.cljs
 (defn mock [module]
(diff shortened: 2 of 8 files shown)


cammellos (approved):

errorists:
[Needs a gradient on](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=7687%3A3461) top of the last two lines of text, otherwise this UI doesn't really work that well

flexsurfer:
@errorists @Serhy i've added gradient

Serhy:
@flexsurfer there are three issues found:
**Issue 1** 
When large message received in 1-1 or Group chats, - there is no collapse button shown. Message is cut to the respective number of chars (lines), so it collapsed, but no way to reveal whole message. All fine in public chat though. 
Screenshot 2020-09-17 at 12 47 46


**Issue 2**
 When I long-tap on revealed whole message - there is no Reaction pop-up appears. Screen is shadowing as expected but just no pop-up to React-Copy-Reply. All fine with collapsed message.
37hf8y3grf3rf

**Issue 3** 
Weird behavior with what happens with iOS: when I send from **Android** large message by pasting it with native Android `Paste` - it has no `Collapse` button on iOS side in public chat.
Same message received by Android user (posted by another Android user) in public chat has the collapsing button as expected. Or, it's also fine when both users, - sender and recipient, - are using iOS device. Not even sure which platform the issue is Android or iOS. Below are the steps to reproduce:
1) Open status on iOS (UserIOS) and Android (UserAndroid) and create new accounts on both
2) Both users join same public chat
3) UserAndroid with the native Android `Paste` option pastes any long test in message input and sends message
4) UserIOS checks whether there is a collapsing button present.

flexsurfer:
thanks @Serhy all should be fixed now

Serhy:
@flexsurfer Everything looks good now, above issues are fixed. 
Two cases though, when there is no way to Copy or leave an emoji reaction on long message:
1) Your own message
2020_09_21_12_54_30_trim

2) When reveal the received message after long-tap made when it was collapsed
2020_09_21_12_53_20_trim

In both cases focus is made on message head and the Reaction pop-up is somewhere below of screen. 

While second scenario has easy workaround, - copying your own big message is impossible.

flexsurfer:
thanks @Serhy fixed

Serhy:
Tested with Android 9, Android 8, iOS 14 (Beta) 
All issues are fixed in 1-1/Group/Public chats
- Long messages collapsed and could be revealed through 'Reveal' button (except your own message sent in chat which are showed in full)
- Long tap on any message view (collapsed/full) shows Reactions panel
- Can reply to long messages and I receive PN for them

Thank you a lot @flexsurfer !

PR #11174, opened 2020-09-16
 Move some pn settings to advanced

### LINKED ISSUE #11170: Tidy up remote push notification options
### User Story

As a user
I want to have push notification options presented in a clear an concise way
So that it's easy for me to make informed decisions

### Description

Currently notifications options are all in one place,
we should split what is useful to most users:
Show notifications
Notifications from non-contacts
Show @ Mentions

To what is only for advanced users:
- Send push notifications
- Push notification server
- Push notification servers

### Acceptance Criteria

- Notification settings are according to designs

notification-1
notification-2

cammellos:
Fixes: #11170 

status: ready


translations/en.json
```diff
@@ -790,13 +790,15 @@
     "not-keycard-text": "The card you used is not a Keycard. You need to purchase a Keycard to use it",
     "not-keycard-title": "Not a Keycard",
     "notifications": "Notifications",
+    "show-notifications": "Show notifications",
     "notifications-settings": "Notifications",
-    "notification-settings": "Notification settings",
+    "notification-settings": "Notifications",
```
flexsurfer, line 794: do we need two keys with the same copy ?

flexsurfer (approved):

translations/en.json
cammellos: good point :)

PR #11195, opened 2020-09-21
[#10060] Wrong ERC20 decimal rounding in "Set max"

### LINKED ISSUE #10060: Wrong ERC20 decimal rounding in "Set max"
# Problem

My intention was to send entire amount of an ERC20 token to another address.

Context: ERC20 with 18 decimals
Entire Amount in my wallet: all decimal was non-0:  the exact entire amount was using all 18 decimals

ISSUE 1: the displayed amount on Status was ceiled. So the displayed amount was grater then my actual amount (checked later with etherscan)
So when i manually set the amount it was accepted by the UI but then the transaction failed

ISSUE2: missing of "MAX" button which set the exact amount without having ti type it in my my self

replicated in ropsten:

image

it seems that it is ceiled up to 14 decimals

flexsurfer:
fixes #10060


rasom (approved):

hesterbruikman (approved):
Tested on Android and iOS, rounding error is resolved (same entered amount as visible on Etherscan)

There are some other issues I think need to be addressed separately so this critical PR is not held up:
- formatting, progressively decreasing font size by decimals
- tx overview screen showing the amount send minus the gas cost; it currently looks as though I'm sending the full amount, which is not really the case (cc @johnlea-quiup

churik:
issue is fixed - now balance is the same as on etherscan without rounding.
Tested on tokens with different decimals and ETH - set max is set literally the same value as in wallet, without rounding.

Checked on IOS13 and Android10.

PR #11217, opened 2020-09-23
[#10667] Upgrade eth-phishing-detect library

### LINKED ISSUE #10667: Upgrade eth-phishing-detect library
# Bug Report

## Problem

 eth-phishing-detect library is outdated , we need to find a way to use a recent version for every release

flexsurfer:
fixes #10667


cammellos (approved):

Serhy:
For new weblinks from package it shows they are blocked, Dapps from DApp loads fine.

PR #11232, opened 2020-09-25
Fix issue with long path names

cammellos:
This fixes an issue with desktop by shortening the path of the database.
It will migrate the filename to a shorter path.
Effectively it should work as before and maintain accounts once
installed.

status: ready


cammellos:
Already merged, closing

PR #11241, opened 2020-09-28
[#11216] Fix moving cursor after pressing mention suggestion on samsung device

### LINKED ISSUE #11216: Cursor moved in the middle of ENS name when typing message with mentioned user
# Bug Report

## Problem

Mentions can be located by ENS/nickname/three-words name. 
In case user has ENS name set, - in the input field it's mention displays ENS name (disregard it was searched by nickname or three-words name or ENS). But when I continue typing message after selecting user from mention list, - cursor automatically moves in the middle of mention (to the point when I entered several chars before selecting user for mention).
Please see gif below for better understanding. 

Could not reproduced with nicknames/three-word names
Could not reproduce on iOS, but on Android 9 and Android 8 

#### Expected behavior

User continue typing from the place after ENS name in mention used

#### Actual behavior

Note here cursor shifted by itself to the point I stopped typing ENS name

2020_09_23_15_24_03_trim

### Reproduction

- Add new user by ENS name (e.g. 'statuse2e')
- Put cursor in the input field
- Type `@stat`
- Tap upon 'statuse2e' user from suggested mentions list
- Continue typing text

### Additional Information

- Status version: develop 1.7
- Operating System: Android9, Android8

rasom:
fix #11216

based on https://github.com/nikolaiwarner/react-native-text-input-reset from this discussion https://github.com/facebook/react-native/pull/12462

status: ready

### DIFF
--- a/src/status_im/native_module/core.cljs
+++ b/src/status_im/native_module/core.cljs
@@ -380,3 +380,8 @@
 (defn deactivate-keep-awake []
   (log/debug "[native-module] deactivateKeepAwake")
   (.deactivateKeepAwake ^js (status)))
+
+(defn reset-keyboard-input [input selection]
+  (log/debug "[native-module] resetKeyboardInput")
+  (when platform/android?
+    (.resetKeyboardInputCursor ^js (status) input selection)))
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -19,6 +19,11 @@ import android.view.WindowManager;
 import android.webkit.CookieManager;
 import android.webkit.CookieSyncManager;
 import android.webkit.WebStorage;
+import android.content.Context;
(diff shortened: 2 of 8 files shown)


src/status_im/native_module/core.cljs
```diff
@@ -380,3 +380,7 @@
 (defn deactivate-keep-awake []
   (log/debug "[native-module] deactivateKeepAwake")
   (.deactivateKeepAwake ^js (status)))
+
+(defn reset-keyboard-input [input selection]
+  (log/debug "[native-module] resetKeyboardInput")
+  (.resetKeyboardInputCursor ^js (status) input selection))
```
flexsurfer: we don't have the implementation in objc. ? how will it work on ios ?

rasom: It won't be called in objc as it is not needed there. It is not needed even on all Androids. I will add some comments now

rasom: added comment

flexsurfer (approved):

PR #11244, opened 2020-09-29
Feature local notifications

Ferossgp:
Push notifications for transactions


cammellos (approved):

Serhy:
Testing is blocked by https://github.com/status-im/status-react/issues/11299 (transaction history is not loaded)

cammellos:
@Serhy @Ferossgp if you rebuild it should be fixed,  we changed the token back to the infura account

Serhy:
@Ferossgp I've tested in iOS14 and Android 9
- transactions within main account: ETH and token. 
- transactions of tokens which are not in the list yet
- custom token transactions
- transaction in sub accounts

There are two issues noticed:
**Issue 1**
When notification about incoming/outgoing transaction received we also increase "unread message" counter in Status app icon. However it is not cleared until app re-login. Not clear when it should disappear (unlike when counter appears for unread message and after message was read it disappears). I'd suggest in a case of transaction disappears after app is opened OR no counter increased at all. @errorists what's the behavior you'd prefer in that case?
Steps to reproduce:
1) Recover user with funds on one device (UserA) on create new user on another Android device (UserB)
2) UserB enables Notification settings along with 'Wallet transactions' setting
3) UserA sends transaction (ETH or token) to UserB
4) Once transaction received and notification on UserB side appears put Status on background
5) Check Status app icon
Actual result: `(1)` is shown on Status app icon. 
Expected result: No counters displayed on Status app icon. 

sssssss

Each transaction will increase counter keeping until user re-login. 


**Issue 2**:
When I tap on appeared notification for transaction app there is an error appears: `Error. Undefined is not a function`. It's the same on both iOS and Android.
ssdsdsdsdqqqq

I suggest when we open notification for transaction we simply open Status without redirects

Serhy:
@Ferossgp Looks good now! Issues above are fixed. Thanks!
Tapping upon notification about transaction I redirected to Wallet account where transaction received.
Please mind my commit when merging

PR #11253, opened 2020-10-01
Fix issue with migration, release

cammellos:
Upgrade status-go to fix an issue with release migration https://github.com/status-im/status-react/issues/11249


flexsurfer (approved):

status-go-version.json
```diff
@@ -2,7 +2,7 @@
     "_comment": "DO NOT EDIT THIS FILE BY HAND. USE 'scripts/update-status-go.sh <tag>' instead",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.62.3",
-    "commit-sha1": "397ee840c650b7fd4a414b38e200274d2d1b716c",
-    "src-sha256": "1x6cs06qqbncvlbkw2ch3alkrhf24m966c7jkvvz6rb9fhbxgrgj"
+    "version": "v0.62.7",
```
flexsurfer: actually are we sure we can use this version? or should we make a branch from 0.62.3 and add only your changes for hotfix and make hotfix version ?

cammellos: I think is probably best if we use HEAD, it contains a few more commits, but because this touches migrations, we risk having them out-of-order, so it's probably less risky to have HEAD (it's 4 extra commits, already tested).

flexsurfer: already tested in hotfix? will they work with outdated status-react ?

flexsurfer: Migrate db to shorter path name looks scary for hotfix

cammellos: Yes, it does look scary, though has been tested, let's create an hotfix branch :)

jakubgs:
Looks like the build was failing because of a bug in `scripts/update-status-go.sh` that didn't account for the same version matching both a branch and a tag in the repo checked with `git ls-remote`:
```
 > git ls-remote https://github.com/status-im/status-go v0.62.3.hotfix.3
59e6602405bfbcf8446d26aca9b8087e84529f8e    refs/heads/release/v0.62.3.hotfix.3
59e6602405bfbcf8446d26aca9b8087e84529f8e    refs/tags/v0.62.3.hotfix.3
```
Which resulted in the `commit-sha1` field containing the same SHA1 twice with a newline:
```
 > cat status-go-version.json   
{
    "owner": "status-im",
    "repo": "status-go",
    "version": "v0.62.3.hotfix.3",
    "commit-sha1": "59e6602405bfbcf8446d26aca9b8087e84529f8e
59e6602405bfbcf8446d26aca9b8087e84529f8e",
    "src-sha256": "140gyhjc54x7vzv9384yv01l1z8fkw3195i0bhg4gv8259b1jq7s"
}
```
Which resulted in the URL Nix was using to fetch the version being broken:
```
 > nix show-derivation /nix/store/09aggy683mzajs8wg6cw86fb0kji0zdz-status-go-59e6602-source.drv | grep urls
      "showURLs": "",
      "urls": "https://github.com/status-im/status-go/archive/59e6602405bfbcf8446d26aca9b8087e84529f8e\n59e6602405bfbcf8446d26aca9b8087e84529f8e.tar.gz"
```
I'll make a separate PR with a fix for the script.

cammellos:
Closing this as already addressed in different PR

PR #11255, opened 2020-10-01
update-status-go: fix case where two matching refs exist

jakubgs:
Before it was possible to break the format of `status-go-version.json`:
```
 > git ls-remote https://github.com/status-im/status-go v0.62.3.hotfix.3
59e6602405bfbcf8446d26aca9b8087e84529f8e    refs/heads/release/v0.62.3.hotfix.3
59e6602405bfbcf8446d26aca9b8087e84529f8e    refs/tags/v0.62.3.hotfix.3
```
Which would result in `commit-sha1` key being set to two commits insted of one.

Issue originally found in #11253.

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -101,11 +101,13 @@ pipeline {
               } else { /* if no universal is available pick first */
                 env.PKG_URL = urls.first()
               }
-              if (utils.isNightlyBuild()) {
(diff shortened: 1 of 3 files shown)


cammellos (approved):

jakubgs:
I've also snuck in addition of download links to build descriptions for individual platform builds.

PR #11273, opened 2020-10-05
do not show zeros for token balances, and fetch all tokens only when …

flexsurfer:
do not show zeros for token balances, and fetch all tokens only when creating new account


cammellos (approved):

churik:
That's amazing work, @flexsurfer !
Tests related to PR are fixed.

PR #11275, opened 2020-10-05
fix availability of INFURA_TOKEN for Android build

jakubgs:
The env variable `INFURA_TOKEN` is used at build time of JS bundle, not the final APK file.
We never passed the `secretsFile` to the derivation for JS bundle so it never saw the `INFURA_TOKEN`.

Depends on: https://github.com/status-im/status-react-jenkins/pull/16

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -1,4 +1,4 @@
-library 'status-react-jenkins@v1.2.4'
+library 'status-react-jenkins@v1.2.5'
 
 pipeline {
   agent { label 'linux' }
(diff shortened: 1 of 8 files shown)


cammellos (approved):

flexsurfer (approved):

jakubgs:
My `cat "${secretsFile}"` shows that the secret is clearly passed:
```
15:17:13  secretsPhase
15:17:13  export INFURA_TOKEN="****";
```

jakubgs:
It appears my fix doesn't work unless I also include the fix by @cammellos from https://github.com/status-im/status-react/pull/11272.

Issue #11192, opened 2020-09-21
Mentions are not resolved in messages with markdown


churik:
# Bug Report
## Problem
Mentions in messages with markdown looks as `@chat-key`

#### Expected behavior
mentions in message

#### Actual behavior
![@roman 2020-09-21 11-07-46](https://user-images.githubusercontent.com/4557972/93749741-b842c200-fbfa-11ea-8fc6-7d74df91fd19.png)

### Reproduction
[comment]: # (Describe how we can replicate the bug step by step.)

* Open Status
* Join any chat
* Send message with markdown and mention inside formatted text (i.e. `>hey @mention` )

### Additional Information

* Status version: nightly 21.09.20
* Operating System: Android, iOS

rasom:
This is not a markdown issue, more likely we son't want to resolve `@name` as `@0xpk` in case if it is placed inside some formatted text.

rasom:
Ok so, there are two ways to solve this issue:
1. do not replace mention with pubkey in case if that mention is inside some formatted text (telegram, discord)
2. parse `0xpk` as mentions even inside markdown (whatsap)

@hesterbruikman @errorists what do you think would be the best way to go?

errorists:
I'm very much for option 1

rasom:
@errorists then the next question: should we detect this during typing and do not suggest options and do not highlight mentions in input? That might be a bit trickier because mention might be inside some formatting tag and that tag might not be closed and... it adds a lot of complexity actually

errorists:
I think ideally the moment you close a tag it strips of mentions from the input inside the tag, wdyt?

rasom:
It is doable, might be complicated though

churik:
Looks good to me. Tested on Android 10 and IOS 13 in 1-1, public chats with different markdown.
Aldo found that markdown with * and ~ is not applied to messages with mentions but I believe it is small issue which will be marked an `won't fix`, so I'll add it to https://github.com/status-im/status-react/issues/11271

cammellos:
Closing this one then, thank you!

PR #11279, opened 2020-10-06
[#11192] Prevent replacing of mention with pubkey inside markdown

rasom:
fixes (partially) #11192 

This PR doesn't prevent mention from being highlighted in the input even if it won't be replaced with pubkey on sending.

status: ready

### DIFF
--- a/src/status_im/chat/models/mentions.cljs
+++ b/src/status_im/chat/models/mentions.cljs
@@ -12,6 +12,163 @@
 
 (def at-sign "@")
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defn re-pos [re s]
+  (loop [res [] s s last-idx 0]
+    (if-let [m (.exec re s)]
+      (let [new-idx (.-index m)
+            idx     (+ last-idx new-idx)
+            c       (get m 0)]
+        (recur (conj res [idx c]) (subs s (inc new-idx)) (inc idx)))
+      res)))
+
+(defn check-style-tag [text idxs idx]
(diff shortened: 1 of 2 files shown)


src/status_im/chat/models/mentions.cljs
```diff
@@ -12,6 +12,157 @@
 
 (def at-sign "@")
 
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
flexsurfer, line 15: why this part is separated? just wondering

rasom: Because file is big already and i needed some visual separation as i was tired by this code :D I thought about moving it to a separate ns but I didn't find a proper name for this piece of... code and just left it here.

flexsurfer (approved):

rasom:
it is actually extremely slow, I'm trying to improve perf at the moment

rasom:
Now it is much faster in case of long message with a few mentions (message with two mentions 4096 chars long is parsed in 5-8ms vs 80-100ms on a relatively slow android device). It is still slow if a long message consists of only mentions or literally each word wrapped into some markdown, but that's not likely to happen often anyway.

churik:
@rasom 
pubkey indeed is not shown now. but an attempt to mention smbd inside markdown, i.e.
`__text @mention some text__`, I cant't tap on mention after sending, so it is resolved as text part.
Is it expected for now?
IMAGE 2020-10-13 12:48:40

rasom:
@churik yep, if mention is added inside any markdown (quote, code block or some text with styles) it shouldn't be resolved. The problem is that we still highlight such mention in the input, but fixing that is not going to be an easy thing and likely will have noticeable impact on input's performance.

PR #11292, opened 2020-10-13
[Fixes: #11290] Increase group size to 20

### LINKED ISSUE #11290: Increase private group chat size to 20
As a user 
I want to be able to add more than 10 people to a group chat
So I can chat with more people in a single place

### Description
Currently the limit is set to 10, it should be increased to 20, and verified that performance sending messages is acceptable on slow devices.

cammellos:
Fixes: #11290

Increases the maximum size of a group chat to 20.

We should test the performance sending messages on a lower end device when 20 participants are in the group.

status: ready

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -50,7 +50,7 @@
 (def command-state-transaction-sent 7)
 
 (def min-password-length 6)
-(def max-group-chat-participants 10)
+(def max-group-chat-participants 20)
 (def default-number-of-messages 20)
 
 (def mailserver-password "status-offline-inbox")


flexsurfer (approved):

Serhy:
Tested on Android 8, Android 9 and iOS14:
- can add up to 20 group chat members
- can not add 21st chat participant from Group Info as well as via Group invitation link.
 
Tested on Android 6.0.1 (Galaxy Note 4) for performance of sending messages: I really don't see much difference between having 10 members and 20 members. 
When sending a message in a Group chat with 10 users with 30 messages in group, - it takes  0.85-1s for message to appear in chat view after Send button pressed. 
Similar numbers on the same build in the same Group chat but with 10 members: 0.9-1sec. Plus it varies from time to time (does not depend from number of users). 

Below are GIFs for two cases (slowed down 10 times) showing how it looks with:

Group chat with 20 members:
ezgif-2-e0c8fd52ec03

and group chat with 10 members
ezgif-3-83d825fba21d

It looks good to merge for me, @cammellos

PR #11307, opened 2020-10-19
Markdown updates

rasom:
This PR concerns usage of markdown in the messages.

On status-go side:
- new `strong-emph` node was added for strong emphasised text with triple `"*"` and `"_"`
- `del` node got proper marshalling and can be recognised (works with single and double `"~"`)

On status react side:
- styles for `strong-emph` nodes
- styles for `del` (strikethrough) nodes
- unsuccessful attempt to use monospaced fonts for `code`/`codeblock` (we have a font which is used as monospaced but doesn't support that property actually `¯\_(ツ)_/¯`). It seems it is out of scope here

status: ready


rasom:
it appears that we don't have a monospaced font, so code blocks won't be monospaced as it is intended

flexsurfer (approved):

rasom:
～～foo～～

qoqobolo:
Checked:

- single, double and tripple "*" and "_"; single and double "~"
- all types of chats
- combinations of markdowns
- links, tags, mentions (3-random, ENS, nicknames), emoji
- replies
- long messages (collapsed)
- pasting
- line break
- russian, arabic, korean, chineese languages
- Android and iOS

PR #11313, opened 2020-10-20
[#11189] Backward compatible ENS registrar

### LINKED ISSUE #11189: Add backwards compatibility for ENS Usernames
# Feature Issue

Currently status-react registers ENS Usernames by calling a hardcoded smart contract address `0xDB5ac1a559b02E12F29fC0eC0e37Be8E046DEF49`

However this will break compatibility once the ENS Usernames get migrated to a new version as the UI don't knows the new address.

### User Story

As a developer I want that migrations on ENS Usernames happen seamlessly, even if they happen in a future version of Status App. 

### Description

UI reads owner of `stateofus.eth` and use it as the ENS Usernames registrar contract. 

Instead of the hardcoded address, UI (https://github.com/status-im/status-react/blob/develop/src/status_im/ethereum/stateofus.cljs) calls `ensRegistry.owner(namehash("stateofus.eth"))` to find out the ENS Username registrar payment address.

In future version of UsernameRegistrar, a ERC721 NFT would be created to each username, resolvable by `ensRegistry.resolver(namehash("stateofus.eth")).addr(namehash("stateofus.eth"))`. 

### Notes

For security purposes, it's important to keep using `approveAndCall`. The price should also be read from the smart contract and displayed on the UI.

rasom:
fix #11189

status: ready

### DIFF
--- a/src/status_im/ethereum/stateofus.cljs
+++ b/src/status_im/ethereum/stateofus.cljs
@@ -1,6 +1,7 @@
 (ns status-im.ethereum.stateofus
   (:require [clojure.string :as string]
-            [status-im.ethereum.json-rpc :as json-rpc]))
+            [status-im.ethereum.json-rpc :as json-rpc]
+            [status-im.ethereum.ens :as ens]))
 
 (def domain "stateofus.eth")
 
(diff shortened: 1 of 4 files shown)


cammellos (approved):

src/status_im/ethereum/stateofus.cljs
```diff
@@ -24,12 +25,26 @@
...
    ;;:testnet "0x11d9F481effd20D76cEE832559bd9Aca25405841"
    })
 
+(def registrars-cache (atom {}))
+
+(defn get-registrar [chain callback]
+  (if-let [contract (get @registrars-cache chain)]
+    (callback contract)
+    (let [registry (get ens/ens-registries chain)]
+      (ens/get-addr
```
3esmit:
```suggestion
      (ens/get-owner
```

Sorry, I overlooked the ENS Registry migration contract and didn't noticed they actually blocked the required feature (reclaim control) until it's migrated to the new ENS Registry.

Due the ENS bug, and it's migration to new address, currently the get-addr of mainnet cannot be changed on non-migrated domains (case of stateofus.eth), and to migrate we have to change the address of the UsernameRegistrar (defeating the feature). Although, function get-owner can achieve the same intended functionality for this feature. 
image

The use of "get-addr" instead of "get-owner" was chosen as gas optimal for future versions of UsernameRegistrar with https://github.com/status-im/ens-usernames/issues/108 in mind, however the use of "get-owner" won't block this planned feature or be absurdly more gas expansive to implement.

3esmit (changes requested):
Due a limitation on non migrated ENS, we cannot change the resolver of stateofus.eth without migrating to ENS and upgrading the contract. @rasom 

As my attached suggestion, change `get-addr` to `get-owner`. therefore using the call `ensRegistry.owner(namehash("stateofus.eth"))` to get registrar address.

3esmit (approved):

PR #11320, opened 2020-10-22
Use ens for referrals rinkeby contract

Ferossgp:


flexsurfer (approved):

Serhy:
It seem works in terms of ENS for contract. 
However, currently with Rinkeby I have no tokens/ETH discplayed in 1-1 chat after referral link is opened. It looks like this:
Screenshot 2020-10-26 at 11 24 50

And when hit Accept there is an Oops error pop-up appears (no Starter pack received in Rinkeby):
Screenshot 2020-10-26 at 11 58 17

I'd suggest if everything is OK with ENS name and contract it point to then lets merge and look separately on why it does not pick up correct data in Rinkeby.

Serhy:
Works now with PR build on Rinkeby

PR #11321, opened 2020-10-22
Image preview interactions

Ferossgp:
Mind that there is a difference between iOs and Android. On Android, there is no bottom sheet because of https://github.com/software-mansion/react-native-gesture-handler/issues/139 and no swipe to dismiss.

### DIFF
--- a/src/status_im/ui/screens/chat/image/preview/views.cljs
+++ b/src/status_im/ui/screens/chat/image/preview/views.cljs
@@ -1,39 +1,79 @@
 (ns status-im.ui.screens.chat.image.preview.views
-  (:require-macros [status-im.utils.views :refer [defview letsubs]])
   (:require [status-im.ui.components.colors :as colors]
             [status-im.ui.components.react :as react]
+            [reagent.core :as reagent]
             [re-frame.core :as re-frame]
             [quo.core :as quo]
+            [quo.platform :as platform]
             [status-im.i18n :as i18n]
             [status-im.ui.components.icons.vector-icons :as icons]
             [status-im.ui.screens.chat.sheets :as sheets]
-            [quo.components.safe-area :as safe-area]))
+            [quo.components.safe-area :as safe-area]
+            ["react-native-image-viewing" :default image-viewing]))
 
-(defview preview-image []
-  (letsubs [{:keys [content] :as message} [:get-screen-params]
-            {:keys [width height]} [:dimensions/window]]
-    [safe-area/consumer
-     (fn [insets]
-       [react/view {:style {:flex 1 :justify-content :flex-end
-                            :padding-top (:top insets) :padding-bottom (:bottom insets)
-                            :background-color colors/black-persist}}
-        [react/view {:flex 1 :align-items :center :justify-content :center}
-         [react/image {:style       (merge {:width            width
-                                            :height           (- height 200)
-                                            :background-color :black})
-                       :resize-mode :contain
-                       :source      {:uri (:image content)}}]]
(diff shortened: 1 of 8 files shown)


qoqobolo:
@Ferossgp could you rebase it plz

qoqobolo:
### ISSUE 1: Can't open chat from the home screen

If open the home screen and start new 1-1 chat via ENS/chat-key, or join some public chat and tap it in the chat list, it will not be opened. After relogin, just the first chat is opened, and all the following ones do not.

_Steps:_

- Open Home screen
- Tap some public chat to join it
- Tap public chat in the chat list to try to open it
- Relogin
- Tap the same public chat
- Navigate to Home screen
- Try to open it again

Android, iOS

ezgif com-gif-maker (3)

qoqobolo:
### ISSUE 2: Can't close keyboard after entering seed phrase during multiaccount recovering on Android

ezgif com-gif-maker (4)

qoqobolo:
### ISSUE 3: Can't open an image in Timeline (Status tab)

_Steps:_
- Open Status tab
- Tap +
- Send one picture to Timeline
- Tap the picture to open it

Android (currently impossible to check on iOS)

Ferossgp:
@qoqobolo added image preview to that images too

churik:
as it seems that PR can cause some regression, may be better to merge it after 1.9 cut, WDYT @cammellos @Ferossgp ?

Ferossgp:
@churik Issue 1 and 2: I've reverted the update to the navigation library, so there should be no regression (I will update it in a separate PR to check the reasons behind regressions). As per issue 3, timeline was merged after this PR was created and because of that, it wasn't implemented in that PR.

churik:
OK then, thanks for explanation @Ferossgp

src/status_im/ui/screens/chat/image/preview/views.cljs
```diff
@@ -1,39 +1,79 @@
...
-          {:on-press #(re-frame/dispatch [:bottom-sheet/show-sheet
-                                          {:content (sheets/image-long-press message true)
-                                           :height  64}])}
-          [icons/icon :main-icons/more {:container-style {:width  24 :height 24
-                                                          :margin 20}
-                                        :color           colors/white-persist}]]]])]))
+(defn footer-options []
+  (let [show-sheet (reagent/atom false)]
+    (fn [{:keys [message on-close]}]
+      ;; FIXME(Ferossgp): Bottom sheet doesn't work on Android because of https://github.com/software-mansion/react-native-gesture-handler/issues/139
```
flexsurfer, line 17: do we need to fix this in the PR ?

src/status_im/ui/screens/chat/image/preview/views.cljs
```diff
@@ -1,39 +1,79 @@
...
+       [footer-options props]]])])
+
+(defn preview-image [{{:keys [content] :as message} :message
+                      visible                       :visible
+                      on-close                      :on-close}]
+  [:> image-viewing {:images              #js [#js {:uri (:image content)}]
+                     :on-request-close    on-close
+                     :hideHeaderOnZoom    false
+                     :hideFooterOnZoom    false
+                     :swipeToCloseEnabled platform/ios?
```
flexsurfer: mix of cases

src/status_im/ui/screens/chat/image/preview/views.cljs
Ferossgp: We should use the same approach on Android as on iOs when the linked bug is fixed

flexsurfer (approved):

PR #11324, opened 2020-10-22
[#11317] Prevent joining to pubchats with upper case letter in topic

### LINKED ISSUE #11317: It's possible to join public chat with capital letter via tag in message
# Bug Report

## Problem
It should not be possible to join public chat if it has capital letter in name. And it's the case when joining from (+) -> Join public chat menu and deep link. But not the case when in message tag has capital letter.
We need to eliminate markdown of tags in chat if tag has capital letters. 

#### Expected behavior

No way to join public chat via tag in message if the tag has capital letter

#### Actual behavior

Screenshot 2020-10-21 at 12 56 24

### Reproduction

- Join unique public chat
- Send message `#PubLicChaT`
- Verify tap upon `PubLicChaT` text in message
**Expected result:** No tag markdown, no redirect to public chat

### Additional Information
- Status version: develop/release 1.8

rasom:
fix #11317

Technically this still can happen on status-go side, but we prevent showing invalid pub chats in the messages and also validate topic before joining on status-react side.


src/status_im/chat/models.cljs
```diff
@@ -259,17 +259,23 @@
               (transport.filters/load-chat chat-id)
               (navigate-to-chat chat-id))))
 
+(defn valid-pubchat-topic? [topic]
+  (re-matches #"^[0-9a-z-]*$" topic))
```
flexsurfer: we have this in `status-im.ui.screens.add-new.new-public-chat.db`

flexsurfer (approved):

src/status_im/chat/models.cljs
rasom: thanks, fixed

churik:
@rasom I have no idea why, but exactly on this PR e2e build transaction history is not loaded.
All fine with i.e. https://github.com/status-im/status-react/pull/11260 PR, so it is not infura limit as far as I can understand.
May it cause by some changes in go?

rasom:
@churik nope status-go change is tiny and doesn't touch anythings related to the wallet

PR #11327, opened 2020-10-22
[#11323] Turn off "Block screenshots" by default in nightly and PRs b…

### LINKED ISSUE #11323: Turn off "Block screenshots" by default in nightly and PRs build
# Bug Report

## Problem
"Block screenshots" by default totally makes sense in release builds, but it slows down taking screens directly in nightly-PRs builds.
Shouldn't be difficult to disable it by default in these build types, but it will speed up getting useful info for issues and make easier testing work.

#### Expected behavior
"Block screenshots" is turned off by default

#### Actual behavior
"Block screenshots" is enabled

flexsurfer:
fixes #11323


jakubgs:
Are we sure we want to do this for nightlies too? PR builds have a separate App ID on iOS and Android so they appear as a separate installation, but nightlies can be installed over a release build, so this might be a privacy concern for those that install a nightly over a release. Thoughts?

churik:
@jakubgs as @cammellos mentioned:

> https://status.im/nightly/
> the disclaimer is pretty clear that they are testing builds

And by the way, on nightlies 'DEBUG' log level is set by default which is from my POV, in that case, is more important privacy concern.

cc @corpetty

jakubgs:
Fair enough.

flexsurfer:
@churik could you approve ?

churik (approved):

PR #11334, opened 2020-10-26
WIP: Images permission management 

Ferossgp:
Permission handling for images in chat. Requires Xcode 12 as it is the only version that supports iOS 14.

I met with the problem that the https://github.com/status-im/react-native-config doesn't work with xcode12, while the original lib works. Ideally fixing xcode12 should be a separate PR and this one rebased on top of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,3 +165,6 @@ trace.edn
 
 /app/
 project.clj
+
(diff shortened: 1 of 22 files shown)


flexsurfer:
lets close for now, feel free to reopen if you will work on it

PR #11345, opened 2020-10-30
Navigate to the root screen when pressing active tab 

flexsurfer:
Navigate to the root screen when pressing the active tab

### DIFF
--- a/src/status_im/ui/components/tabbar/core.cljs
+++ b/src/status_im/ui/components/tabbar/core.cljs
@@ -75,16 +75,16 @@
 (def tabs
   (reagent/adapt-react-class
    (fn [props]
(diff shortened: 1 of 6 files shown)


errorists:
@flexsurfer yo, does that work for all tabs? because it's part of the native tab bar behaviour, if you tap on the currently active tab it takes you to the root level of the tab, regardless how deep you are. So it should have the same effect when in wallet or browsing settings.

flexsurfer:
yes @errorists  thanks

errorists:
beautiful, thank you

Ferossgp (approved):

Serhy:
Thanks @flexsurfer, tested with iOS14 and Android 9 and Android 8. Also fixed e2e here.

PR #11352, opened 2020-11-02
Exclude google deps based on env

### LINKED ISSUE #11010: Research a way to avoid pulling in Play Services/Firebase
# Intro

During research on releasing the app via F-Droid in #8512 it was found that our dependency on `status-react-camera` pulls in a lot of dependencies we actually don't use, but are not compatible with the F-Droid requirement to not include non-FOSS software.

See: https://f-droid.org/docs/Inclusion_Policy/

>We cannot build apps using proprietary tracking/analytic dependencies like Crashlytics and Firebase. Please talk to upstream about an untainted build flavor

# Dependencies

Here's a list of stuff `status-react-camera` pulls in that should not be there:
```
+--- project :react-native-camera
|    +--- com.google.android.gms:play-services-vision:17.0.2
|    |    +--- com.google.android.gms:play-services-base:16.0.1
|    |    |    +--- com.google.android.gms:play-services-basement:16.0.1 -> 16.1.0
|    |    |    |    \--- androidx.legacy:legacy-support-v4:1.0.0
|    |    |    |         ...OMITTED...
|    |    |    \--- com.google.android.gms:play-services-tasks:16.0.1
|    |    |         \--- com.google.android.gms:play-services-basement:16.0.1 -> 16.1.0 (*)
|    |    +--- com.google.android.gms:play-services-basement:16.1.0 (*)
|    |    \--- com.google.android.gms:play-services-vision-common:[17.0.2] -> 17.0.2
|    |         +--- com.google.android.gms:play-services-base:16.0.1 (*)
|    |         +--- com.google.android.gms:play-services-basement:16.1.0 (*)
|    |         +--- com.google.android.gms:play-services-clearcut:16.0.0
|    |         |    +--- com.google.android.gms:play-services-base:16.0.1 (*)
|    |         |    +--- com.google.android.gms:play-services-basement:16.0.1 -> 16.1.0 (*)
|    |         |    +--- com.google.android.gms:play-services-phenotype:16.0.0
|    |         |    |    +--- com.google.android.gms:play-services-base:16.0.1 (*)
|    |         |    |    +--- com.google.android.gms:play-services-basement:16.0.1 -> 16.1.0 (*)
|    |         |    |    \--- com.google.android.gms:play-services-tasks:16.0.1 (*)
|    |         |    \--- com.google.android.gms:play-services-tasks:16.0.1 (*)
|    |         +--- com.google.android.gms:play-services-flags:16.0.1
|    |         |    +--- com.google.android.gms:play-services-base:16.0.1 (*)
|    |         |    \--- com.google.android.gms:play-services-basement:16.0.1 -> 16.1.0 (*)
|    |         \--- com.google.android.gms:play-services-phenotype:16.0.0 (*)
```
```
+--- com.google.firebase:firebase-ml-vision:19.0.3
|    +--- com.google.firebase:firebase-common:16.0.3
|    \--- com.google.firebase:firebase-ml-common:[17.0.0] -> 17.0.0
|         +--- com.google.firebase:firebase-common:16.0.3 (*)
|         \--- com.google.firebase:firebase-iid:17.0.3
|              +--- com.google.firebase:firebase-common:16.0.3 (*)
|              \--- com.google.firebase:firebase-iid-interop:16.0.1
```

# Details

This was pointed out by @licaon-kter in [fdroiddata#7179](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/7179#note_387866683) that the APK includes `com.google.android.gms.*` dependencies:
https://www.virustotal.com/gui/file/f3232309d534ae3d688dbc9cd089719e353d6eeccf7ca262572060f011454afa/details

### LINKED ISSUE #11029: Make installreferrer dependency optional for F-Droid
# Description

During research done in #8512 and #11010 we found out that the dependency on [`installreferrer`](https://developer.android.com/google/play/installreferrer/library) that we are adding in #10620 is incompatible with the FOSS-only requirement of the F-Droid catalog due to the [license used](https://developer.android.com/studio/terms.html).

There is a relatively recent issue for this: https://gitlab.com/fdroid/fdroidclient/-/issues/1932
But there is no conclusive decision.

# Requirements

In order to make publishing to F-Droid possible we will need to be able to build a version of the release that does not include the dependency on `installreferrer` library. This should be achieved through settings in `.env.*` files or via a Gradle build flavor.

Ferossgp:
@jakubgs can you please check if with `GOOGLE_FREE=1` env is enough? Also let me know if there are nix changes also required.

Resolves #11029 and Resolves #11010


jakubgs:
Nice work dude.

According to the FDroid PR - [fdroiddata#7179](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/7179) - the issue was clearly with `com.google.android` and `com.android.installreferrer`:

>Looking at https://www.virustotal.com/gui/file/f3232309d534ae3d688dbc9cd089719e353d6eeccf7ca262572060f011454afa/details I see a lot of `googleapis.com` URIs and `com.google.android.gms.common.api.GoogleApiActivity`... not looking good :(

https://gitlab.com/fdroid/fdroiddata/-/merge_requests/7179#note_387867379

So think this looks good. Though we probably should change Nix expressions to support this too.

Mind if I commit to your PR? We can squash later.

Ferossgp:
> Mind if I commit to your PR? 

Sure!

jakubgs:
Since this is done via `.env.*` file I don't think it's necessary to add special flags fin `Makefile` or Nix.
I guess we could just edit that, or add `.env.fdroid`.

I'm a bit torn because there's two ways, or actually even three, in which we add configuration like that:

1. The `gradle.properties` file that holds config values like `ANDROID_ABI_SPLIT` and we fetch with `getEnvOrConfig`
2. Env variables like `LOG_LEVEL` which we set in `.env.*` files and normally affect just the Clojure side
3. Env variables which are secrets provided via a secrets file give from outside of Nix sandbox to avoid including it in derivation

The way you're doing it is kind of a mix between 1 and 2, because it affects both Clojure and Gradle.

I wish I could make this a bit less messy, but maybe that's for another time...

jakubgs:
Waaaaait. This line:
```clojure
(def google-free (enabled? (get-config :GOOGLE_FREE "0")))
```
What does it do? It gets `GOOGLE_FREE` form env right? When is that? At Clojure build time?
Because I don't think we source `.env.*` files when making the JSBundle:
https://github.com/status-im/status-react/blob/8d7e3e8690efb9dc1cc41c4a30f92aa220db4539/nix/mobile/android/jsbundle/default.nix#L74-L76

Ferossgp:
> At Clojure build time?

It is done during the app bundle into APK, not in Clojure build. It is transformed into a JS function call during Clojure compilation.

jakubgs:
I see, that makes sense. Thanks for explaining. Then I guess I don't need to change anything. We just need to adjust the `.env` file for FDroid builds with `sed`. Not pretty but it should work.

jakubgs (approved):

Serhy:
@Ferossgp on real device (Samsung with Android 9 / 8) app crashes right after start with:

Screenshot 2020-11-03 at 18 55 52

jakubgs:
Same here, tried the PR build on a OnePlus 6 and it explodes the same way.

Serhy:
Just in case to mention, - after latest commit it's still appears. App crashes on Android (all OK with ipa build on iPhone)

Serhy:
Looks good to me: can install on Android 8, Android 9, iOS 14. Create/recover users. Message between users in 1-1/public chats, notifications appears, can send transaction in 1-1/Wallet/Dapp.

Issue #11355, opened 2020-11-02
Sometimes app crashes on logout events  with "discovery.stopped"


churik:
# Bug Report

## Problem
2-3 cases from 180 are crashed on logout. Sometimes app is crashed on logout on real devices as well - regardless of place (Profile > logout, switching networks, changing fleet etc)

#### Expected behavior
no crash

#### Actual behavior
app is crashed after logout confirmation

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Create multiaccount
- Log out

### Additional Information

- Status version: nightly 02/11/20
- Operating System: Android, iOS

#### Logs

```
11-02 03:42:11.870  6816  6909 D StatusModule: Signal event: {"type":"discovery.stopped","event":null}
11-02 03:42:11.877  6816     0 E Go      : panic: runtime error: invalid memory address or nil pointer dereference
11-02 03:42:11.877  6816     0 E Go      : [signal SIGSEGV: segmentation violation code=0x1 addr=0x128 pc=0xb49a6359]
11-02 03:42:11.877  6816     0 E Go      : 
11-02 03:42:11.877  6816     0 E Go      : goroutine 342 [running]:
11-02 03:42:11.877  6816     0 E Go      : github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/p2p.(*Server).Peers(0x0, 0x0, 0x85591c00, 0xb4c145a5)
11-02 03:42:11.878  6816     0 E Go      : 	/build/go/src/github.com/status-im/status-go/vendor/github.com/ethereum/go-ethereum/p2p/server.go:301 +0x39
11-02 03:42:11.878  6816     0 E Go      : github.com/status-im/status-go/eth-node/bridge/geth.(*gethNodeWrapper).PeersCount(0x838a4d10, 0x83dfd760)
11-02 03:42:11.878  6816     0 E Go      : 	/build/go/src/github.com/status-im/status-go/eth-node/bridge/geth/node.go:104 +0x30
11-02 03:42:11.878  6816     0 E Go      : github.com/status-im/status-go/protocol.(*Messenger).online(...)
11-02 03:42:11.878  6816     0 E Go      : 	/build/go/src/github.com/status-im/status-go/protocol/messenger.go:328
11-02 03:42:11.878  6816     0 E Go      : github.com/status-im/status-go/protocol.(*Messenger).watchConnectionChange.func1(0x838e4d80, 0x83d37208)
[...]

PR #11358, opened 2020-11-03
[Fixes: #11355] Fix crash on logout

cammellos:
On shutdown the app was calling `PeersCount`, but sometimes the server
would be unavailable, resulting in a call to a nil pointer.

Upgrades status-go version with the fix.


flexsurfer (approved):

cammellos:
@churik one test failed, but I don't believe that code has been changed/touched, could you please have a look? 
Also the logout issue should be fixed, I can't be 100% sure as I wasn't able to replicate, but maybe we can keep an eye on it.
Thanks!

churik:
fail is not related to PR for sure.
Thank you for this fix! I'll reopen issue if it doesn't help.

cammellos:
Thank you, merged!

PR #11360, opened 2020-11-03
Tests/new e2e

churik:
- test with default log level and fleet is extended and moved to medium: now it also checks changing fleet and log level
- test with pairing keycard when offline and adding watch -only account (partially blocked due to issue, will be investigated further)
- added check to create group chat without contacts


Serhy (approved):
❤️

Issue #11349, opened 2020-10-30
Members stop receiving messages if one of the members sends multiple (several batches) images to group chat


qoqobolo:
# Bug Report
## Problem
**_This issue is constantly reproduced in develop after #11316 was merged._**

If create a group chat and send 5 images, then 5 more (sometimes for several more), then the other members will not receive all the images from the second batch. After the images are stopped being received by other members of the group chat, text messages (including in 1-1 chat) begin to be received with a delay (3-20 minutes).

### Reproduction
User 1:

* Create a group chat with User 2

User 2:

* Join the group chat

User 1:

* Send 5 images in a batch
* Send 5 more images in a batch
* Continue until message receiving stops

User 2:

* Check how many images were received

### Additional Information

* Status version: develop with #11316 
* Operating System: Android, iOS

#### Logs
[SENDER status_logs.zip](https://github.com/status-im/status-react/files/5465797/SENDER.status_logs.zip) iOS device

[RECEIVER Status-debug-logs.zip](https://github.com/status-im/status-react/files/5465799/RECEIVER.Status-debug-logs.zip) Android device

cc @Ferossgp @cammellos @hesterbruikman

cammellos:
Thanks for reporting the issue!

A few reasons why it might be the case (we'd have to investigate further to make sure):
1) Rate limiting, you might be hitting the node rate limiting
2) Messages are dispatched in batches to a given user (though that should be every 300ms, so maybe too quick for this), but we never check the batch is too large, but we might be hitting the limit

The bug does not seem to be crucial, but it's worth investigating just to understand what's happening, as it might not be necessarily due to images

churik:
IMO as it easy to reproduce now, better to look at it until feature will go to release (so before 1.9)
Not sure though now many users now are using group chat, so likely not many users will be affected.

cammellos:
We are hitting the limit in terms of size of messages on the sender side:
```
t=2020-10-30T10:51:17+0000 lvl=eror msg="error sending message" namespace=mvds error="huge messages are not allowed [fe1b2ec7ba9bf5283c7f3a1e4f26fffd2d6452205e9ed0547b34d6597bff2dd8][1167920][1048576]"
```
Investigating further

cammellos:
It's probably due to batching of messages in MVDS, as we don't limit those. I will push out a potential fix and test it.

PR #11361, opened 2020-11-03
[Fixes: #11349] Split large messages in batches

cammellos:
When sending messages in quick succession, it might be that multiple
messages are batched together in datasync, resulting in a single large
payload.

This commit changes the behavior so that we can pass a max-message-size
and we split the message in batches before sending.

status: ready


cammellos:
@qoqobolo would you mind testing this when you have time to make sure https://github.com/status-im/status-react/issues/11349 is fixed please?
Thank you!

qoqobolo:
@cammellos sure! On it

qoqobolo:
@cammellos can't reproduce #11349, all messages are received. Thanks for the fix!
Should I send it to merge now or work is still in progress? 
> status: wip

cammellos:
@qoqobolo thanks for testing, not yet, I have to merge the status-go PR first, I will take care of it.
Thank you

rasom (approved):

PR #11371, opened 2020-11-04
🙅🏽 [#8955] Mailserver and other text overlap

### LINKED ISSUE #8955: In profile address is shown instead of name and "Mailserver" label is overlapped when connected to custom mailserver
###  Description

*Type*: Bug

*Summary*: when connected to custom mailserver, address is shown  instead of name and no "Mailserver" label in Profile > Sync settings

#### Expected behavior
photo_2019-09-09 14 18 00

#### Actual behavior
dfs

### Reproduction

- Open Status
- Go to profile > Sync settings
- Add custom mailserver 
- Connect to it
- Go back

### Solution

*Summary*:
I'm actually not sure how we can handle mailservers with long names.
For now "Mailserver" label is overlapped, which is not right IMO:

photo_2019-09-09 14 26 34

@andmironov or @errorists please take a look, what is better from your POV.

### Additional Information

- Status version: nightly09/09/2019

shivekkhurana:
fixes #8955 

### Summary
This is a long standing issue that applies to `quo.list.item` component. This component has a `left-item` and a `right-item` and is used throughout the app.

The left and the right item views should appear in the same row. In case the views cannot fit, left item has priority, ie: right item should be truncated first.

The current system achieved this by setting `flex:1` on the left item, which means "hey left item, feel free to grow and occupy all the space available after right-item has rendered". The problem with this approach is that when right item needs more space than available, left item gets truncated (because it cannot grow, because right occupied all space).

This PR fixes the issue by making left and right item shrink and grow differently, making sure that right item is truncated more aggressively, essentially prioritizing display of left item.

### Testing notes
This change applies to most list views of the app. Please get a general feel of how everything looks. I have made sure that it works.

#### Platforms
- Android
- iOS

##### Functional
- public chats
- group chats
- user profile updates
- networks
- mailservers

### Steps to test
Generic: 
Have a look at settings screen, including all sub-menus. Keep an eye for alignment of right chevrons (right pointing arrows).

*Mailservers Scenario*:
- Goto settings > Sync settings
- Select mailservers
- Turn off automatic mailserver selection
- Choose a mailserver with long name (maybe Hong Kong one?)
- Click back
- The name of the mailserver will be truncated with ellipses in between

*About Scenario*:
- Goto settings > About
- You should see the text "Node version" on left and actual version on right
- Without this PR, the left bit was truncated

status: ready

### DIFF
--- a/src/quo/components/list/item.cljs
+++ b/src/quo/components/list/item.cljs
@@ -84,10 +84,13 @@
 
 (defn title-column
   [{:keys [title text-color subtitle subtitle-max-lines
-           title-accessibility-label size text-size title-text-weight]}]
+           title-accessibility-label size text-size title-text-weight
+           right-side-present?]}]
   [rn/view {:style (merge (:tiny spacing/padding-horizontal)
-                          {:justify-content :center
-                           :flex            1})}
+                          ;; make left-side title grow if nothing is present on right-side
+                          (when-not right-side-present?
+                            {:flex            1
+                             :justify-content :center}))}
    (cond
 
      (and title subtitle)
@@ -124,15 +127,25 @@
 
 (defn left-side [props]
   [rn/view {:style {:flex-direction :row
-                    :flex           1
+                    ;; Occupy only content width, never grow, but shrink if need be
+                    :flex-grow      0
+                    :flex-shrink    1
                     :align-items    :center}}
    [icon-column props]
    [title-column props]])
 
 (defn right-side [{:keys [chevron active accessory accessory-text animated-accessory?]}]
   (when (or chevron accessory)
-    [rn/view {:style {:align-items    :center
-                      :flex-direction :row}}
+    [rn/view {:style {:align-items     :center
+                      :justify-content :flex-end
+                      :flex-direction  :row
+                      ;; Grow to occupy full space, shrink when need be, but always maitaining 16px left gutter
+                      :flex-grow       1
+                      :flex-shrink     1
+                      :margin-left     16
+                      ;; When the left-side leaves no room for right-side, the rendered element is pushed out. A flex-basis ensures that there is some room reserved.
+                      ;; The number 80px was determined by trial and error.
+                      :flex-basis      80}}
      [rn/view {:style (:tiny spacing/padding-horizontal)}
       (case accessory
         :radio    [controls/radio {:value active}]
@@ -142,6 +155,7 @@
                    {:value active}]
         :switch   [controls/switch {:value active}]
         :text     [text/text {:color           :secondary
+                              :ellipsize-mode  :middle
                               :number-of-lines 1}
                    accessory-text]
         accessory)]
@@ -204,7 +218,8 @@
                    :size                      size
                    :text-size                 text-size
                    :subtitle                  subtitle
-                   :subtitle-max-lines        subtitle-max-lines}]
+                   :subtitle-max-lines        subtitle-max-lines
+                   :right-side-present?       (or accessory chevron)}]
        [right-side {:chevron             chevron
                     :active              active
                     :on-press            on-press
--- a/src/status_im/ui/screens/sync_settings/views.cljs
+++ b/src/status_im/ui/screens/sync_settings/views.cljs
@@ -9,8 +9,10 @@
 
 (views/defview sync-settings []
   (views/letsubs [{:keys [syncing-on-mobile-network?
-                          use-mailservers?]}           [:multiaccount]
-                  mailserver-id                        [:mailserver/current-id]]
+                          use-mailservers?]}          [:multiaccount]
+                  current-mailserver-id               [:mailserver/current-id]
+                  current-fleet                       [:fleets/current-fleet]
+                  mailservers                         [:mailserver/mailservers]]
     [react/view {:style {:flex 1 :background-color colors/white}}
      [topbar/topbar {:title (i18n/label :t/sync-settings)}]
      [react/scroll-view
@@ -29,7 +31,8 @@
                       :title               (i18n/label :t/offline-messaging)
                       :on-press            #(re-frame/dispatch [:navigate-to :offline-messaging-settings])
                       :accessory           :text
-                      :accessory-text      (when use-mailservers? mailserver-id)
+                      :accessory-text      (when use-mailservers?
+                                             (get-in mailservers [current-fleet current-mailserver-id :name]))
                       :chevron             true}]
       ;; TODO(Ferossgp): Devider componemt
       [react/view {:height           1


shivekkhurana:
|Mailserver Screen | About Screen |
|---|---|
|Simulator Screen Shot - iPhone 11 - 2020-11-04 at 18 05 19 | Simulator Screen Shot - iPhone 11 - 2020-11-04 at 18 05 09|

flexsurfer (dismissed):
@Ferossgp wdyt ?

flexsurfer:
@shivekkhurana we need to make sure in all places we use this component left item have a priority because for example when signing txs we have labels at left and values at right, and values should have a priority

src/quo/components/list/item.cljs
```diff
@@ -124,30 +124,32 @@
...
                               :number-of-lines 1}
                    accessory-text]
         accessory)]
      (when (and chevron platform/ios?)
-       [rn/view {:style {:padding-right (:tiny spacing/spacing)}}
-        [icons/icon :main-icons/next {:container-style {:opacity         0.4
-                                                        :align-items     :center
-                                                        :justify-content :center}
-                                      :resize-mode     :center
+       [rn/view
```
flexsurfer: could you elaborate on  this change ?

Ferossgp:
@shivekkhurana Thank you for the PR! There are some regressions, could you please approach them?
The space on the right of accessory is too small

<details>
<summary>Screenshots</summary>


Simulator Screen Shot - iPhone 11 - 2020-11-05 at 10 47 07

</details>

Every text accessory is truncated even if there is plenty of space:

<details>
<summary>Screenshots</summary>


Simulator Screen Shot - iPhone 11 - 2020-11-05 at 10 46 54
Simulator Screen Shot - iPhone 11 - 2020-11-05 at 10 46 30


</details>


You can use the Quo Previews under Settings tabs in Profile in Dev mode to have better development feedback while modifying Quo components.

src/quo/components/list/item.cljs
shivekkhurana: I had to remove this container style to offset the effect of adding `justify-content :flex-end` (line 139). 

The padding caused the right cheveron icon to be oddly placed because how the flex-box grew and shrank. I had to remove the additional padding so that the flex model can decide the alignment. There was some trial and error involved.

After reading this again, I feel that `:align-items` and/or `:justify-contents` to `:center` should not be truncated. I'll re-check.

shivekkhurana:
@Ferossgp Thanks for pointing out quo views. I'll check it and try to address your questions.

src/quo/components/list/item.cljs
flexsurfer: so then two more points:
1) we need this view anymore it just empty, the rule - fewer views is always better
2) so we lost padding for icon ?

shivekkhurana: 1. View is redundant, I already deleted it locally, will push soon.
2. Yes we did. But with the updated flex-box model, the `right-item` was placed with ample padding. It seemed fine to me visually, but having seen the Quo component, I realize I was wrong. I'll fix it.

shivekkhurana:
@Ferossgp 
> Every text accessory is truncated even if there is plenty of space:

I think it happens on that screen by design choice. The prod. app exhibits the same behavior.
<details><summary>Screenshot</summary>

IMG_1888

</details>

flexsurfer:
> I think it happens on that screen by design choice. The prod. app exhibits the same behavio

yes addresses truncated by design 0000...0000

Ferossgp:
@shivekkhurana yes, by design it should have 3 dots but in current PR there are 6 dots, it doubled them. Also, in the Quo Preview text acessory is trucated, while it should be visible

shivekkhurana:
I have changed how the Flexbox works. The current system allowed right-side to occupy as much space as need be, and filled the remaining with left (`x` means content, and `-` means empty space):
|xxx-------------- |xxxxxx|

This caused a problem when right-side occupied too much space, not leaving enough room for left content:
|-| xxxxxxxxxxxxxxxxxxxxx|

I have reversed it, because at most places, left label has a priority. Now, the system lets left-side take as much space it needs, and fills the rest with right:
|xxx|----------------------x|

This makes sure that left item is rendered on priority, and right is forced to shrink:
|xxxxx|xxxxx...xxxxxxxxxxxx|

This solution leads to the reverse problem, ie, what if left occupies all the space leaving nothing for right?
|xxxxxxxxxxxxxxxxxxxxxxxx|-|

I have handled this with a `flex-basis`, to ensure there is some reserved space for right. It seems to work well so far, but QA is definitely needed.

@Ferossgp The 6 dots issue and accessory truncation should be resolved now.

shivekkhurana:
| Screenshots | Are | Cool | |
|---|---|---|---|
|Simulator Screen Shot - iPhone 11 - 2020-11-06 at 21 29 55|Simulator Screen Shot - iPhone 11 - 2020-11-06 at 21 29 32| Simulator Screen Shot - iPhone 11 - 2020-11-06 at 21 29 16|Simulator Screen Shot - iPhone 11 - 2020-11-06 at 21 32 52|

flexsurfer:
@shivekkhurana could you please show how amount and network fee will look with stress test labels, thanks

shivekkhurana:
@flexsurfer Stress test results for amount list item:

| Sane Digits | |
|---|---|
|Simulator Screen Shot - iPhone 11 - 2020-11-09 at 15 26 26 | Simulator Screen Shot - iPhone 11 - 2020-11-09 at 14 43 09 |

| Insane Digits | |
|---|---|
| Simulator Screen Shot - iPhone 11 - 2020-11-09 at 15 23 25 | Simulator Screen Shot - iPhone 11 - 2020-11-09 at 15 22 38 |

Ferossgp (approved):
Overall, looks good to me 👍

shivekkhurana:
@Ferossgp The e2e tests failed because I missed an edge case where no right-side was present. 

The problem showed up when the recent "Timeline" feature was merged (I didn't see it initially because I was not rebased on develop).

My solution is to make left grow if no right side is present (which was the original behaviour), and let left occupy only the space it needs if right is present. I was able to do it without any additional props.

src/quo/components/list/item.cljs
```diff
@@ -84,10 +84,12 @@
...
 (defn title-column
   [{:keys [title text-color subtitle subtitle-max-lines
-           title-accessibility-label size text-size title-text-weight]}]
+           title-accessibility-label size text-size title-text-weight
+           right-side-present?]}]
   [rn/view {:style (merge (:tiny spacing/padding-horizontal)
-                          {:justify-content :center
-                           :flex            1})}
+                          ;; make left-side title grow if nothing is present on right-side
+                          (when-not right-side-present? {:flex            1
```
Ferossgp: I think the map could be on the new line to make lines shorter

Ferossgp (approved):

src/quo/components/list/item.cljs
shivekkhurana: Okay.

churik:
thank you for detailed instructions, appreciate that 👍

churik:
@shivekkhurana 
Tested builds, and yes currently id of the mailserver is truncated with ellipses in between (also checked in version, custom network), good work!
In #8955 also stated that expected value is the name of mailserver, but currently it is {enode-id}
Monosnap 2020-11-13 10-35-49

shivekkhurana:
@churik can you check if enode is displayed for all servers ? 

I didn't change the content, only added ellipses, so it's possible that enode id is displayed for custom mailservers. [Here's what it looks like by default](https://user-images.githubusercontent.com/1925158/98113240-6368bb80-1ec9-11eb-9eb9-f77c5361789d.png).

churik:
@shivekkhurana I meant custom mailservers for now. 
So if you add `custom_mailserver` with `enode://ee2b53b0ace9692167a410514bca3024695dbf0e1a68e1dff9716da620efb195f04a4b9e873fb9b74ac84de801106c465b8e2b6c4f0d93b8749d1578bfcaf03e:status-offline-inbox@104.197.238.144:443`
when you switched to `custom_mailserver` - you 'll see some hash value (sorry, it is not even {enode-id}, have no idea what is that) instead of name.

That was part of #8955

shivekkhurana:
@churik I have pushed a minor change that should fix the name issue. Can you please test again?

churik:
Thank you for fixes! Looks good on IOS13 and Android10.

PR #11375, opened 2020-11-05
Update number of parallel sessions to align with Sauce

Serhy:
Update number of parallel sessions to align with Sauce

### DIFF
--- a/ci/tests/Jenkinsfile.e2e-nightly
+++ b/ci/tests/Jenkinsfile.e2e-nightly
@@ -47,10 +47,12 @@ pipeline {
         ]) {
           dir('test/appium/tests') {
             sh """
-              python3 -m pytest -m testrail_id \
-                -m "not upgrade" \
-                -n24 --rerun_count=2 \
+              python3 -m pytest \
+                --numprocesses 16 \
+                --rerun_count=2 \
                 --testrail_report=True \
+                -m testrail_id \
+                -m "not upgrade" \
                 --network=${params.NETWORK} \
                 --apk=${params.APK_NAME}
             """
--- a/ci/tests/Jenkinsfile.e2e-prs
+++ b/ci/tests/Jenkinsfile.e2e-prs
@@ -83,7 +83,8 @@ pipeline {
         ]) {
           dir('test/appium/tests') {
             sh """
-              python3 -m pytest -n24 \
+              python3 -m pytest \
+                --numprocesses 16 \
                 --rerun_count=2 \
                 --testrail_report=True \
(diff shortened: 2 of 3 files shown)


churik (approved):

ci/tests/Jenkinsfile.e2e-nightly
```diff
@@ -49,7 +49,7 @@ pipeline {
             sh """
               python3 -m pytest -m testrail_id \
                 -m "not upgrade" \
-                -n24 --rerun_count=2 \
+                -n16 --rerun_count=2 \
```
jakubgs: Parameters like `-n` mean nothing. How about adding an env variable in `environment` called `PARALLEL_SESSIONS` with the number set and then using it in the command. By using a sensible variable name we can make it clear what the flag does.

Also, maybe align the flag formatting for be the same in all 3 files for easier editing/reading.

jakubgs: I just notice there isn't one here. We can add it like here:
https://github.com/status-im/status-react/blob/e7807202c93d5a1c8db28ec7846aa761e7a9a283/ci/Jenkinsfile.android#L28-L39

jakubgs: So I guess after `parameters`:
```groovy
  environment {
    PARALLEL_SESSIONS = 16
  }
```
Unless you have a better name for the variable. Maybe `TEST_PARALLEL_SESSIONS` for more clarity?
Is this parameter specific to SauceLabs or any tests?

ci/tests/Jenkinsfile.e2e-nightly
```diff
@@ -47,10 +47,12 @@ pipeline {
         ]) {
           dir('test/appium/tests') {
             sh """
-              python3 -m pytest -m testrail_id \
-                -m "not upgrade" \
-                -n24 --rerun_count=2 \
+              python3 -m pytest \
+                --numprocesses 16 \
```
jakubgs, line 51: A full flag works too, nice.

jakubgs (approved):

ci/tests/Jenkinsfile.e2e-nightly
Serhy: -n is short for --numprocesses so I updated that key to make is more meaningful. It shows number of max processes to start for combines test run. I also would keep it separate from Jenkinsfile.android because env vars in Jenkinsfile.android describe variables applied for build, while -n (--numprocesses) is more related to test framework settings. It would be better in that case to specify it in pytest.ini. I keep as it is now.

PR #11381, opened 2020-11-05
Referrals don't crash if a token is missing

Ferossgp:


Serhy:
@Ferossgp  thanks! 

Tested build https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-201105-172136-768de8-pr-universal.apk and it's good now, - on Rinkeby when 1-1 chat appears no crash and Accepting starter pack triggers contract so Starter pack received and attributions increased.

shivekkhurana (approved):
The code looks good. Is there an issue that this PR closes ?

src/status_im/ui/components/invite/chat.cljs
```diff
@@ -56,9 +56,13 @@
 (defn starter-pack []
   (let [pack        @(re-frame/subscribe [::invite/starter-pack])
         tokens      (utils/transform-tokens pack)
-        reward-text (cstr/join ", " (map (comp name :symbol first) tokens))]
+        reward-text (->> tokens
```
cammellos, line 59: We could make use of transducers here as I believe that when threading it will expand not to use transducers, if performance/memory are a concern.

cammellos (approved):

src/status_im/ui/components/invite/chat.cljs
Ferossgp: The tokens list should vary from one element to three, so we should be fine for now

PR #11399, opened 2020-11-09
[#11384] Status Updates : Open profile when tapping on username or av…

### LINKED ISSUE #11384: Status Updates : Open profile when tapping on username or avatar
# Problem

A small issue to follow up on the Status Updates / Timeline PRs
This is about linking usernames and avatars found in each status to the person who sent it.

## Implementation

Tapping on a username or avatar of the person who sent the update navigates to their profile, similar to chat messages. Tapping on your own does nothing. TouchableOpacity for the touch effect so it's visible.

## Acceptance Criteria

The behaviour specified above is implemented and confirmed to work correctly.

flexsurfer:
fixes #11384

### DIFF
--- a/src/status_im/ui/screens/status/views.cljs
+++ b/src/status_im/ui/screens/status/views.cljs
@@ -71,16 +71,18 @@
                 :flex-direction     :row
                 :background-color   (when (and timeline? outgoing) colors/blue-light)
                 :padding-horizontal 16}
-    [react/view {:padding-top 2 :padding-right 8}
-     (if outgoing
-       [photos/member-identicon (multiaccounts/displayed-photo account)]
-       [photos/member-identicon identicon])]
+    [react/touchable-highlight {:on-press #(re-frame/dispatch [:chat.ui/show-profile-without-adding-contact from])}
+     [react/view {:padding-top 2 :padding-right 8}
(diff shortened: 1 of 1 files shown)


Serhy:
Tested on Android 8, 9 and iOS 14. Looks good to me! Can open user profile like from Public chat and able to perform all action with further redirect to respective view where applied:
- Navigate to 1-1 chat 
- Add/Remove to contacts 
- Block user (no status from blocked users after that)
- Mute (iOS only for now) and no notifications for messages are shown
- Change nickname, share account

When I tap on my own username/icon of status posted then I redirected to Profile main tab (profile root view disregard the view it was before). Is it OK, @errorists ? Or to make it behave that nothing happens when user taps his own username/icon?

Ferossgp (approved):

errorists:
@Serhy you tested it, which would you prefer? :) I have no strong opinion myself, I feel like not doing anything at all is preferable but seeing we have the Profile navigation already in place here, I'm fine to merge as is ✅

Serhy:
Awesome, thanks @errorists (I just wanted to confirm it's OK since it contradicts to what described in issue as expected result for own profile)
Let's merge then, @flexsurfer, thanks!

PR #11401, opened 2020-11-09
[#10790] update Keycard SDK, remove workaround

### LINKED ISSUE #10790: Integrate new react-native-status-keycard with better error handling
Until now, the method autoOpenSecureChannel and autoPair from the SDK returned the same type of exception on connection lost and other APDU related errors.

After merging https://github.com/status-im/react-native-status-keycard/pull/29 we will get a different exception type for the two situations. The steps to take in status-react are

1. Update to the new react-native-status-keycard
2. Verify that the following scenario still behave as designed:
  a. wrong password during pairing
  b. lost connection during pairing
  c. invalidated pairing (card unpaired through different app, openSecureChannel fails)

@yenda also mentioned that a workaround was implemented because of the current lack of separation between the two error conditions. We should be able to remove this after merging.

bitgamma:
fixes #10790 

status: ready

### DIFF
--- a/nix/deps/gradle/deps.json
+++ b/nix/deps/gradle/deps.json
@@ -3542,30 +3542,30 @@
   },
 
   {
-    "path": "com/github/status-im/status-keycard-java/android/3.0.1/android-3.0.1",
+    "path": "com/github/status-im/status-keycard-java/android/3.0.4/android-3.0.4",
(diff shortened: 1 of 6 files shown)


jakubgs (approved):

jakubgs:
Why was this merged without a successful iOS build?

PR #11410, opened 2020-11-10
ci: fix iOS signing, use same keychain name

jakubgs:
Using different temporary keychains does not work if we do not set `default_keychain=true`, because `codesign` then can't find the cert:
```
error: No signing certificate "iOS Distribution" found: No "iOS Distribution" signing certificate matching team ID
```
But if we set `default_keychain=true` then we cause a race condition when the keychain is deleted by a parallel job while another is using it as its default.

For this reason we have to use a static keychain name and keep it between builds.

I tried disabling `default_keychain=true` in #11378 but it worked only because the default user keychain already had the cert.

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -1,4 +1,4 @@
-library 'status-react-jenkins@v1.2.6'
+library 'status-react-jenkins@v1.2.5'
 
 pipeline {
   agent { label 'linux' }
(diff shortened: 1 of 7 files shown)


jakubgs:
Nope, still failing with:
```
...../libswiftCore.dylib: errSecInternalComponent
```
I hate this signing stuff so much. It makes no sense.

flexsurfer (approved):

Issue #11450, opened 2020-11-19
Large messages are dropped


cammellos:
# Bug Report

## Problem

Large messages are dropped and never sent in some cases, this applies mainly to images and batched images.

#### Expected behavior

The message is sent and received.

#### Actual behavior

The message is not sent

#### Notes

Upon investigation this is due to 2 factors:

1) Slow devices
2) Slow upload speed

In the case of 1, a slow device will take a long time calculating the POW of the message, even though for images we greatly reduced it.

In both cases it results in the envelope either expiring locally, or expiring before it reaches the next hop, therefore being discarded.

A few potential solutions:

1) Reduce/remove pow (this has compatibility issues that need to be dealt with)
2) Increase ttl (but that increases pow calculation)
3) Allow expired envelopes to be propagated if they fall within the sync allowance

Another thing to investigate is how this impacts confirmations.
If a message is dropped because it is expired, is that message confirmed by the mailserver?

cammellos:
One of the reasons is addressed here: 
https://github.com/status-im/status-react/pull/11451 

There were actually some raw images being sent, though I believe the issue is still there, I can replicate with slow upload speed.

cammellos:
Looking at TTL.
Essentially the flow for sending messages is not really a good flow for bandwidth/cpu constrained devices.

How it works is:

1) The message TTL/POW is calculated.
2) The message is added to the pool of messages to be sent
3) At interval of 300 ms messages are broadcasted to each peer concurrently, batched together

This is problematic for a few reasons:

- Between 1 & 3 a lot of time be passed, especially if the PoW calculation was expensive, during that time the TTL clock is ticking
- 3 is problematic because if you are on slow upload networks (home networks for example), and you try to upload messages concurrently, you end up with none of the messages making it in time to the other end.
- Another issue is that messages generated on the device are mixed with messages received by the device. A priority queue that prioritizes messages based on the author should help.

I have changed the code so that:
1) Messages are not uploaded concurrently to all peers
2) POW calculation is not done if messages are currently being uploaded (we want to really run the POW just in time)

That seems to have helped ( I am now able to send 3 images from my home network to my mobile device), though is not the best solution as it's not a solid solution, as it would require more code changes.

The way I would normally handle this is:

1) Have a priority queue with messages ordered by Author->TTL
2) Don't calculate TTL until you are ready to send to the first peer
3) Don't send to multiple peers concurrently 

[...]

PR #11451, opened 2020-11-19
Dont send unprocessed images

cammellos:
Partially addresses: #11450

When selecting an image from camera roll we would immediately setting
the filepath with the raw image, process and later once it was processed
we would replace the file path.
On slow devices, the user might press send while the image is still
processing, causing the raw image to be sent, which might be discarded
as too large and never sent.

This commit changes the behavior so that we only set the uri once the
image is processed, it might add some delay, but the image should be
correctly sent.

status: ready

### DIFF
--- a/src/status_im/chat/models/images.cljs
+++ b/src/status_im/chat/models/images.cljs
@@ -141,12 +141,11 @@
   (let [current-chat-id (:current-chat-id db)
         images          (get-in db [:chats current-chat-id :metadata :sending-image])]
     (if (get-in db [:chats current-chat-id :timeline?])
-      {:db              (update-in db [:chats current-chat-id :metadata :sending-image] {uri {:uri uri}})
+      {:db              (assoc-in db [:chats current-chat-id :metadata :sending-image] {})
        ::image-selected uri}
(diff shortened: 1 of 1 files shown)


rasom (approved):

Serhy:
@cammellos @hesterbruikman it was tested on Android 8 (Galaxy J7) slow device, Galaxy S8 and iOS14. 

Looks good! Images with 4MB sent in chat received by recipient and all subsequent images and text messages delivered fine too. Same relates to timeline, takes some noticeable time now for image to process on slow device. 

It takes 5-7 second for 4.2MB image to appear in message input (i.e. to be processed) after it was tapped from camera roll. But it guarantees us chat is working after message is sent. 

Another downside from UX pow: while the image is processing user may nervously tap million times upon it, - in that case once image appears in chat -> user taps Send and image is sent. But. Those million taps are not forgotten. And image reappears in message box again. You close it from preview, but it reappears again. You send it and it appears again.

cammellos:
I guess the UX issues will have to be addressed, but I will merge in the meantime

flexsurfer:
we could just show a loading indicator in the image frame, and disable the send button

PR #11454, opened 2020-11-19
🧴Mentions hygiene

shivekkhurana:
# Summary
- Remove redundant call to `users-fn` and pass `mentionable-users` directly
- Rename `get-suggestions` to `get-user-suggestions`, will be helpful when the `get-channel-suggestions` function is implemented
- Extract inline function props to text input as a namespaced function. Now they won't get redefined at every key-press and are slightly easier to read and reason for

These changes were made while working on # autocomplete feature for chat text-input (which is not completed yet).

### Testing notes
- Check if autocomplete mention works fine

#### Platforms
- Android
- iOS

- All types of chat

##### Non-functional

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -79,6 +79,80 @@
       :accessibility-label :send-message-button
       :color               (styles/send-icon-color)}]]])
 
+(defn selection [cursor]
+  ;; NOTE(rasom): In case if mention is added on pressing suggestion and
+  ;; it is placed inside some text we have to specify `:selection` on
+  ;; Android to ensure that cursor is added after the mention, not after
+  ;; the last char in input. On iOS it works that way without this code
+  (when (and cursor platform/android?)
+    (clj->js {:start cursor :end cursor})))
+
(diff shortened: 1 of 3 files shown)


src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -79,6 +79,81 @@
...
+        start         (.-start ^js range)
+        end           (.-end ^js range)]
+    (re-frame/dispatch
+     [::mentions/on-text-input
+      {:new-text      text
+       :previous-text previous-text
+       :start         start
+       :end           end}])
+    ;; NOTE(rasom): on Android `on-text-input` is dispatched after
+    ;; `on-change`, that's why mention/channel suggestions are calculated
```
rasom: we can just say "suggestions" instead of "mention/channel suggestions" here and in the rest of comments

rasom (changes requested):
tests are broken

rasom (approved):

shivekkhurana:
Tests related to mentions are failing although the feature works fine when I test manually. Checking the test source to see what causes the change. It's also possible that I missed refactoring some components.

Issue #11452, opened 2020-11-19
Using "Take photo" option it is possible to attach > 1 image


churik:
# Bug Report

## Problem
Based on discussion in `release` channel, for now better solution is to disable sending images in a batch.
It is disabled when selecting from gallery, but when selecting option "Take photo", you can attach even more > 5 photos.

#### Expected behavior
1 image

#### Actual behavior
![IMAGE 2020-11-19 16:50:16](https://user-images.githubusercontent.com/4557972/99689452-4e725780-2a87-11eb-93de-0c12053c17aa.jpg)

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Start 1-1 chat
- Use "Take photo". Repeat several times

### Additional Information
- Status version: RC 1.9 (build 6)
- Operating System: Android, iOS

PR #11455, opened 2020-11-20
[Fixes: #11452] Don't allow more than x pictures if taken from camera

cammellos:
This commit fixes an issue with taking pictures from camera, where
the check for image count was not checked.

status: ready


flexsurfer (approved):

churik:
Tested on IOS and Android: on pr build it is not possible to select more than 5 photos from gallery and using "Take photo"

PR #11456, opened 2020-11-20
💸 [#11214] Add option for user to increase the gas payment for a pendin…

### LINKED ISSUE #11214: Add option for user to increase the gas payment for a pending transaction
# Feature Issue

Add option for user to increase the gas payment for a pending transaction.  The user facing changes required are:

- When the user is redirected to the History screen after making a transaction, the transaction should be displayed on the History screen in a 'pending' state (unless it's already included in a block).  The fact that the transaction is in this state should be indicated by the attached pending_state_icon.png

- A 'Increase gas' button should be added underneath the transaction entry in the history screen.

- Tapping on this button should bring up a increase gas dialogue.  

This feature should be implemented at the same time as related feature #11215

### User Story

As a User, I want to increase the gas payment for as transaction I've already made that hasn't yet been included in a block so that the transaction gets included faster.

### Description

See design below.  Note that these screens are from the forthcoming wallet design refresh.  To implement in the current wallet history screen, just use the new pending_state_icon.png (attached) to represent that a transaction is in the pending state in the history view.  

NOTE: please use current wallet design for now, new wallet design will be tracked in a separate issue.

increase gas overview

Icon (let me know what other icon format variants are required)
pending_state_icon

For the gas adjustment dialogue, just use the existing screen (shown below) for the time being.  The updated wallet design will come later.
IMG_2037

### Acceptance Criteria

Test case:
1) After making a token transfer from their wallet, the user is redirected to the History screen
2) On the History screen the user can see that their transaction is pending because the list item representing the transaction has a 'pending' icon (as opposed to the 'sent' and 'received' arrow icons).
3) User taps the 'Increase gas' button and the Increase gas dialogue appears
4) User uses the dialogue to increase the gas being paid for the transaction. After the user taps 'Update' the dialogue closes.
6) In the History screen, User taps on the transaction to go to the Transaction Details screen.  They check that the updated gas price is correctly represented in the Transaction Details screen.
7) When the transaction is included in a block, the list item representing the transaction in the history screen changes to the 'sent' state e.g. the icon changes to the 'Sent' icon and the 'Increase gas' button disappears. 

### Notes

[comment]: # (For the developer who will implement. Include planning poker estimates, guidance, etc.)

### LINKED ISSUE #11215: Add option for user to cancel a pending transaction.
# Feature Issue

In the Wallet History screen, add a button that lets a user cancel a pending transaction.  This feature should be implemented at the same time as related feature #11214 

### User Story

As a user, I want to be able to cancel a pending transaction, so that A) if I've spotted an error with the transaction before it's included in a block I can cancel it and B) if the transaction is stuck and I want to cancel it instead of increasing it's gas.

### Description

See design below.  Note that this screen is from the forthcoming wallet design refresh, the only UI change required by this issue is the addition of the 'Cancel' button underneath the pending transaction list item.  Requires #11214 to be implemented first, as this issue adds a 'pending' state to transactions in the history screen.

cancel transaction

### Acceptance Criteria

Test case
1. After making a token transfer from their wallet, the user is redirected to the History screen
2. On the History screen the user can see that their transaction is pending because the list item representing the transaction has a 'pending' icon (as opposed to the 'sent' and 'received' arrow icons).
3. User taps the 'Cancel' button.
4. The transaction is canceled and the list item representing this transaction is removed from the list of transactions in the history screen.

### Notes

- This requires signing a transaction without using the user password, if that's not feasible at the moment for technical reason (i.e we don't store the user key in memory), having a sign dialog is fine, though not desirable. UPDATE: Judging by the code, the key is not stored in memory at any point https://github.com/status-im/status-go/blob/200273aa0ead9eef16bb5054d60a9edf73192f98/account/accounts.go#L120 , but whoever picks up this task please make sure that is the case.
- For anyone picking this up, leave a comment on whether you feel you can do the keycard work as well, otherwise we'll have it out of scope.

flexsurfer:
fixes #11214
fixes #11215

### DIFF
--- a/src/status_im/signing/core.cljs
+++ b/src/status_im/signing/core.cljs
@@ -19,7 +19,8 @@
             [status-im.utils.utils :as utils]
             [status-im.wallet.prices :as prices]
             [status-im.wallet.core :as wallet]
-            [taoensso.timbre :as log]))
+            [taoensso.timbre :as log]
+            [clojure.set :as clojure.set]))
 
 (re-frame/reg-fx
  :signing/send-transaction-fx
@@ -103,24 +104,29 @@
                                          :cb       #(re-frame/dispatch [:signing/transaction-completed % tx-obj-to-send hashed-password])}})))))
 
 (fx/defn prepare-unconfirmed-transaction
-  [{:keys [db now]} hash {:keys [value gasPrice gas data to from] :as tx} symbol amount]
+  [{:keys [db now]} new-tx-hash {:keys [value gasPrice gas data to from hash]} symbol amount]
   (log/debug "[signing] prepare-unconfirmed-transaction")
   (let [token (tokens/symbol->token (:wallet/all-tokens db) symbol)
-        from  (eip55/address->checksum from)]
-    {:db (assoc-in db [:wallet :accounts from :transactions hash]
-                   {:timestamp (str now)
-                    :to        to
-                    :from      from
-                    :type      :pending
-                    :hash      hash
-                    :data      data
-                    :token     token
-                    :symbol    symbol
-                    :value     (if token
-                                 (money/unit->token amount (:decimals token))
-                                 (money/to-fixed (money/bignumber value)))
-                    :gas-price (money/to-fixed (money/bignumber gasPrice))
-                    :gas-limit (money/to-fixed (money/bignumber gas))})}))
+        from  (eip55/address->checksum from)
+        ;;if there is a hash in the tx object that means we resending transaction
+        old-tx-hash hash]
+    {:db (-> db
+             ;;remove old transaction, because we replace it with the new one
+             (update-in [:wallet :accounts from :transactions] dissoc old-tx-hash)
+             (assoc-in [:wallet :accounts from :transactions new-tx-hash]
+                       {:timestamp (str now)
+                        :to        to
+                        :from      from
+                        :type      :pending
+                        :hash      new-tx-hash
+                        :data      data
+                        :token     token
+                        :symbol    symbol
+                        :value     (if token
+                                     (money/unit->token amount (:decimals token))
+                                     (money/to-fixed (money/bignumber value)))
+                        :gas-price (money/to-fixed (money/bignumber gasPrice))
+                        :gas-limit (money/to-fixed (money/bignumber gas))}))}))
 
 (defn get-method-type [data]
   (cond
@@ -148,15 +154,17 @@
              :token    token
              :symbol   symbol}))))))
 
-(defn parse-tx-obj [db {:keys [from to value data]}]
-  (merge {:from {:address from}}
+(defn parse-tx-obj [db {:keys [from to value data cancel? hash]}]
+  (merge {:from    {:address from}
+          :cancel? cancel?
+          :hash    hash}
          (if (nil? to)
            {:contact {:name (i18n/label :t/new-contract)}}
            (let [eth-value  (when value (money/bignumber value))
                  eth-amount (when eth-value (money/to-fixed (money/wei->ether eth-value)))
                  token      (get-transfer-token db to data)]
              (cond
-               (and eth-amount (or (not (zero? eth-amount)) (nil? data)))
+               (and eth-amount (or (not (.equals ^js (money/bignumber 0)  ^js eth-amount)) (nil? data)))
                {:to      to
                 :contact (get-contact db to)
                 :symbol  :ETH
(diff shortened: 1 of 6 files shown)


hesterbruikman:
@flexsurfer do you know if there's a minimum required increase? In testing I tried:
- Send 55 SNT
- Set gas limit to 40000
- tx Pending
- Increase gas to 50000
Result: `Transaction failed. Replacement transaction underpriced`

### Other notes
#### issue 1 - Increase Gas visible on failed transaction
Current: Failed transaction shows `Increase gas` 
Expected: If increase gas effectively retries sending the same transaction, looks like it does, button label should be `Retry`. 

#### issue 2 - Increase gas on failed transaction should increase nonce
Current: Increase gas is possible on failed transaction, but doesn't increase nonce. Error shows `nonce too low`
Expected: If increase gas is possible on failed transaction, 'retry' should increase nonce

#### issue 3 - Cancel visible on failed transaction
Current: Cancel is visible on failed transaction
Expected: Cancel is not visible for failed transactions

#### issue 4 - Replacement tx too low to Cancel
Current: Default gas for replacement tx was set to 21000. Error `replacement transaction underpriced`. After underpriced replacement transaction can't try again because `nonce is too low`
Expected: Expect default gas limit to be set the same as tx the user wants to cancel

#### issue 5 - Transaction details should confirmation progress bar for failed transaction
Current: failed tx shows progress on confirmation
Expected: tx details shows progress bar in complete state

cc @johnlea-quiup

flexsurfer:
thanks for the review @hesterbruikman 
> do you know if there's a minimum required increase? In testing I tried:

you need to change the gasPrice not the gas limit

flexsurfer:
increase and cancel should be visible only for pending transactions i believe, i just forgot to add condition :(

hesterbruikman:
> increase and cancel should be visible only for pending transactions i believe, i just forgot to add condition :(

Agree. It's a separate feature to be able to Retry:

```
As a user
I want to retry a transaction after it has failed
because I want to try sending the same transaction with more gas
```

Another would be:

```
As a user
I want to send a transaction request for the amount I sent
Because I made a mistake and want to request a refund (no guarantees)

```

I'll create separate issues for these as they'd be new features and not urgent

Serhy:
@flexsurfer, I tested on IOS14 and Android9 builds https://status-im-prs.ams3.digitaloceanspaces.com/StatusIm-201120-133733-d88e4b-pr-universal.apk and https://i.diawi.com/bip2Pv 

All below scenarios were performed for ETH only: 
- Send TX and Update while it's pending PASS
- Send TX and Cancel while it's pending PASS
- Send TX, - update it again to make it Pending again, - Cancel it by third TX
- Send first TX (Max token value), send second TX (same token with Max token value) with normal gas, - Cancel first TX.  
- Send first TX (Max token value), send second TX (same token with Max token value) with normal gas, - Update first TX. 
- Can update TX for an added account
- Can update TX for an added account by seed phrase  
- Can update TX for an added account by private key
- Updating transaction but while I make the changes to gas price and original transaction was mined, - in that case there is a pop-up error 'nonce too low' is returned as expected PASS

**Issue 1:** 
Can't update properly transaction for sent token (tried SNT, MANA, AST, UNI) - it seems data and amount is not picked up for pending TX and so I end up sending 0 ETH to token contract with no data and this TX always fails.
E.g. https://etherscan.io/tx/0x81c704b33cdf4023aca71e624899e41e54016e7ac51bc7342e24305f6bf602c0
Reproduction:
1) Recover account with ETH and any token
2) Send to any address token but set Gas Price three times less than it's current recommended by network value
3) Hit 'Increase Gas' button from history view upon your 'Pending transaction'
**Actual result:** At this point it's already visible that 0 ETH is about to be sent:
2020_11_23_19_46_06_trim
4) Increase gas price
5) Tap update and confirm the transaction replaced

**Issue 2:** 
After user Updated/Canceled transaction we still show the 'Pending...' record in the transaction history but add the new one (with the same nonce as replacement/cancelation) transaction. And even after transaction was mined - we keep the `Pending...` one... with `Increase Gas` and `Cancel` buttons. Pressing buttons has no effect which is good. But is it possible @flexsurfer to hide 'Pending...' transaction from History view after replacement of this TX confirmed (Succeeded or Failed)?  

**Issue 3** 
Essentially it's separate issue but I decided to keep a record about that: 
there is no way to update pending transaction if user re-logined to account because all pending transaction are dropped from transaction history at the current app version. It's known issue (https://github.com/status-im/status-react/issues/9684 and https://github.com/status-im/status-react/issues/9976)

flexsurfer:
thanks @Serhy 1,2 should be fixed now

0kok0:
I'd need help/an explanation to understand where and how the passphrase/keys are masked in memory

flexsurfer:
@0kok0 hey, are you sure your question is related to this PR ? we don't touch keys in this PR

0kok0:
> @0kok0 hey, are you sure your question is related to this PR ? we don't touch keys in this PR

Ah okay cancel and increase-gas only redispatch sign nice, thx

Serhy:
@flexsurfer thanks! It looks good to me with all scenarios above (including token sending) except one UI issue (and the fact that after re-login there is no way to update Pending TX in Status because we don't pickup Pending TX yet after login in to account):

If I update Pending transaction of sent **token**, then on the Sending view there is a 0 ETH shown to be sent to contract token address. In fact, after sending such transaction there is correct update happens and we just increase gas price of original token TX. However on UI side it might be confusing what is user about to send. 

**To reproduce:** 
1) Recover account with ETH and any token
2) Send to any address token but set Gas Price three times less than it's current recommended by network value
3) Hit 'Increase Gas' button from history view upon your 'Pending transaction'
Actual result: 0 ETH shown as what is about to be sent

flexsurfer:
thanks @Serhy  should be fixed now

src/status_im/signing/core.cljs
```diff
@@ -107,20 +108,22 @@
...
-                    :data      data
-                    :token     token
-                    :symbol    symbol
-                    :value     (if token
-                                 (money/unit->token amount (:decimals token))
-                                 (money/to-fixed (money/bignumber value)))
-                    :gas-price (money/to-fixed (money/bignumber gasPrice))
-                    :gas-limit (money/to-fixed (money/bignumber gas))})}))
+    {:db (-> db
+             (update-in [:wallet :accounts from :transactions] dissoc (:hash tx))
```
cammellos: It's not clear what's the difference between `tx` and `hash`, we are `dissocing` first `(:hash tx)` and then `assoc` `hash` back in, I suppose that `hash` is the new hash while `(:hash tx)` is the old one? 
If so maybe we could make it clearer from the parameter names.

flexsurfer: yeah, good point, also was thinking about that

flexsurfer: @cammellos done

cammellos (approved):

Serhy:
Thanks a lot @flexsurfer! Tested with iOS 14 and Android9 latest builds and no bugs found. 
Can speed up pending token/ETH transaction from history view or Cancel it. `Sending...` views of are correctly reflect assets which are going to be sent. Works when making it from main account or added account or added by seed phrase account in Wallet.

PR #11465, opened 2020-11-23
[#11438] Add "sent" message confirmations

### LINKED ISSUE #11438: Add "sent" message confirmations
```
As a user
I want to know my message has left my device
So that I have confidence that it's on its way to the recipient
```

### Description

Next to each message sent  there should be an icon indicating the state of the message:

- The message is actually being sent (I will refer to it as the `sending` state)
- The message has been propagated to the network (I will refer to it as the `sent` state)

### Out of scope

- Transaction messages
### Designs

https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=1%3A2

### Implementation details

- A message can be marked as `sent` once a confirmation is received from a node your are connected to.
  Currently only mailservers sends back confirmations. 
  This is already implemented and as signal is already propagated to status-react. https://github.com/status-im/status-react/blob/473c5135b25e295d073ec5444ed5dcef432b5fff/src/status_im/signals/core.cljs#L54

flexsurfer:
fixes #11438


cammellos (approved):

churik:
e2e failures are fixed in https://github.com/status-im/status-react/pull/11477
Tested:
- messages are confirmed
- resending
- sending from offline
- sending from mobile network
Amazing work!

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)

PR #11480, opened 2020-11-27
nix: add a way to deal with new gradle version to docs

jakubgs:
Also adjusted the `generate.sh` script and split stuff into separate function to make it easier to run specific stages of the generation process.


flexsurfer (approved):

jakubgs:
Your failure was due to `pod repo update` not being run on `macos-03` for a while. This is normally done by this job:
https://ci.status.im/job/status-tools/job/run-pod-repo-update
But it depends on the queue to get to different hosts, and since `03` is last, it rarely gets updates.
I adjusted it to be more random:
```groovy
  agent { label params.HOST_LABEL ?: "macos-0${(currentBuild.startTimeInMillis % 3) + 1}" }
```

PR #11495, opened 2020-12-01
 Fixes: #10065 fix mailserver popup showing after canceling it

### LINKED ISSUE #10065: Popup error reappears if tap on "Cancel" in "Error connecting to mailserver" popup

cammellos:
Reopening this as it's still a valid issue, for example if the mailserver was initially online but later it went offline, the popup interferes with the user being able to change settings or ignore the warning.
 Original PR

https://github.com/status-im/status-react/pull/10342
### Summary

Fixes #10065 by setting a `dismissed` flag on Cancel, so the popup will not appear again.
Selecting a new mail server sets `dismissed` as false, so the error can appear again

#### Platforms

Tested Android simulator

### Steps to test

See issue above to see reproduction steps

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -221,6 +221,11 @@
    (mailserver/save-settings cofx current-fleet mailserver-id)))
 
 (handlers/register-handler-fx
+ :mailserver.ui/dismiss-connection-error
+ (fn [cofx [_ new-state]]
+   (mailserver/dismiss-connection-error cofx new-state)))
+
+(handlers/register-handler-fx
  :mailserver.ui/unpin-pressed
  (fn [cofx _]
    (mailserver/unpin cofx)))
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -59,6 +59,9 @@
 (defn preferred-mailserver-id [db]
   (get-in db [:multiaccount :pinned-mailservers (node/current-fleet-key db)]))
 
+(defn connection-error-dismissed [db]
+  (get-in db [:mailserver/connection-error-dismissed]))
+
 (defn mailserver-address->id [db address]
   (let [current-fleet (node/current-fleet-key db)]
     (:id (some #(when (= address (:address %))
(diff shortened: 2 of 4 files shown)


flexsurfer:
i still think flow should be redesigned from scratch, current UI flow is broken, but even if we want to solve this current case, the root of the problem is that we receive `mailserver/check-connection-timeout` from the timer, and if we'll wait some time with opened popup we'll have several popups anyway (this is not solved in this PR i believe) to solve it properly we need to stop the timer when we receive an error first time, or use timeout, also we need to clear timeout on cancel 

edit: its not a timer its async-periodic-exec

edit2: the problem is that auto selection and manuall selections are mixed, and also UI flow for custom mailservers is different, probably as a proper solution we need to stop async-periodic when we receive error first time if mode is manual

q: do we show same alert if its manual but not a custom mailserver ?

cammellos:
> q: do we show same alert if its manual but not a custom mailserver ?

Yes, we show the same alert, which I believe is the right behavior (say the IP has been banned by your ISP, you want to have the same behavior).

> the problem is that auto selection and manuall selections are mixed, and also UI flow for custom mailservers is different, probably as a proper solution we need to stop async-periodic when we receive error first time if mode is manual

We could, but that's a critical path of the code and I'd rather not touch it, before we had the timer we spend ages trying to fix issues with the mailserver not reconnecting anymore because of state management, so having a watchdog always on though is not an elegant solution, it makes me sleep better at night.

I agree with you that there might be better solutions, but currently there's little appetite to change this code and this to me feels like an ok solution, and it addresses the issue that user have reported (just a couple of days ago), so to me this is good enough, the next natural step would be to move this logic to status-go, so we can simplify the UI code and it's easier to test.

What do you think?

flexsurfer:
ok if we want to fix that lets fix
1) we need to set this flag when we receive this error , but not when cancel (because while popup is opened we accumulate popups because we still receiving errors ) 
2) we need to reset this flag when we select another mailserver (lets say you connected to custom with error, ignore error and then connect to another and receive error, we shouldn't ignore it here )

cammellos:
cool, thank you, will do that

cammellos:
@flexsurfer thanks, I have addressed the suggestions.

The last thing that would be left to do is how you suggested to reset the checks once a re-connection takes place, this should be easy to do but I will address in a different PR if that's ok.
ready to be reviewed again.

flexsurfer (approved):

Issue #10801, opened 2020-06-11
Message description should always be located below the image in chat


churik:
# Bug Report

## Problem
Leftover from https://github.com/status-im/status-react/pull/10609.
Expected that is you attach some text to image, it is always will appear after image in chat. Currently it sometimes appears below, sometimes under image.

#### Expected behavior
description is below

#### Actual behavior
<img width="183" alt="Monosnap 2020-05-27 16-26-21" src="https://user-images.githubusercontent.com/4557972/83033035-e2aeb500-a036-11ea-9a61-f87f158df70a.png">

### Reproduction

[comment]: # (Describe how we can replicate the bug step by step.)

- Open Status
- Start 1-1 chat
- Send several images with description

### Additional Information

- Status version: nightly 11/06/2020
- Operating System: Android, iOS

Ferossgp:
Actually image and text are sent as 2 different messages, so in that case the messages are not part of the image. The text which is attached to the image is `Update to latest version to see a nice image here!` and it is hidden in the UI

hesterbruikman:
@Ferossgp does this mean that it's a time stamp issue? The client doesn't know the message is related to the image and order depends on how quickly image and text message travel through the network?

Ferossgp:
Yes this is the problem of order, but the problem root I described https://discordapp.com/channels/624205794384281629/626738033701683200/740880403245826120

PR #11500, opened 2020-12-01
[Fixes: #10801] Send chat messages in order

cammellos:
Chat messages are now sent in order using a different endpoint
`sendChatMessages`.
Text should always be displayed after images.

This is not implementing a Caption field, that would require either a
protocol change or leverage the `text` in the message.
It applies for both normal chats and timelines.

Move also all inputs under `chat/inputs` so we avoid re-renders as
`chats` has changed.

### Order
The text message should always be shown below images in normal chats and above images in timeline, as order is reversed.

### Testing

- Mentions
- Sending images/text


flexsurfer (approved):

qoqobolo:
Tested:
- 1-1 and group chat
- timeline
- image + links (with and without preview)
- image + long collapsed message
- in what order the new post is displayed in a user profile view
- when going from offline to online
- on mobile network

@cammellos works great! Ready to merge.

PR #11508, opened 2020-12-02
flatlist render changes

flexsurfer:
flatlist render changes


cammellos (approved):
it'd be good to add some rationale on why the changes were made, so we don't go back to previous implementation, not sure what's the best place (maybe git commit description or comments in code)

PR #11511, opened 2020-12-03
Interact with ropsten web3 to send assets

Serhy:
Sometimes we need faucet ETH/token in testnet for tests. But if may not work correctly making e2e test false fail. For such purpose we may use separate account as a 'backup' scenario and faucet asset via web3 dirrectly. 
Two methods: donate_testnet_eth and donate_testnet_token will do the job.


churik (approved):
🎉 
🎉 
🎉 
🎉 
🎉 
🎉 
🎉  awesome work!

Issue #9851, opened 2020-01-17
'Confirmed' status in 1-1 chat pending transaction as soon as account re-logins


Serhy:
# Bug Report

## Problem

Disregard the sent transaction in 1-1 chat with 'Send' command which still in 'Pending...' state (in network and so in wallet), - after user re-opens the app the `Outgoing transaction` gets `Confirmed!` status. 
If to tap on such TX nothing happens, but it because there is no TX in history yet at all so that's expected and a part of another issue.
This issue is dedicated to align transaction `Outgoing transaction` status send in 1-1 chat to it's actual status: if TX is not in TX history - we keep showing 'Pending...' status

#### Expected behavior

Outgoing transaction keep staying with 'Pending...' status in 1-1 chat after account relogin if it's 'Pending...' in network
#### Actual behavior

Outgoing transaction gets 'Confirmed... ' status in 1-1 chat after account relogin always

Video: https://drive.google.com/file/d/1oFzAqlt-47rRII4yNrqJGqOqqamzno6E/view?usp=sharing

### Reproduction
- Open Status and recover an account with funds (UserA)
- Add UserB in contacts and open 1-1 chat with this contact
- UserA sends 'Send transaction' with valid amount of ETH/tokens
- UserB taps `Accept and share address`
- UserA signs transaction in 1-1 chat
- UserA re-logins to account and checks transaction status in 1-1 chat

### Additional Information
- Status version: develop 0.14.0 (2020011702)
- Operating System: Android and iOS

Serhy:
Not a blocker for V1 IMO but rather nice to have.

flexsurfer:
looks like a blocker for me @rachelhamlin , because it might confuse user that tx is confirmed but it isn't

flexsurfer:
@Serhy is it only outgoing or incoming as well ?

Serhy:
No, only outgoing, @flexsurfer 
Incoming kept in "Shared '[account_name]'" until transaction gets confirmations and appears in 'Wallet'

Serhy:
Only outgoing, @flexsurfer 
For incoming messages in 1-1 chat - they stay in "Shared [account_name]" until transaction is confirmed.

cammellos:
This is due to the fact that pending transaction originating in the app are not stored https://github.com/status-im/status-react/issues/9976 , one way to fix this is to at least store in the database the pending transactions that are created within the app.

cammellos:
[...]

PR #11512, opened 2020-12-04
[Fixes: #9851] 'Confirmed' status in 1-1 chat in pending outgoing transactions

cammellos:
This commit does two things:
1) Allow to set the url for verifying transactions & ens names in status-go
2) Fixes the issue mentioned above.

The problem was that if the transaction was not in the history, it would default to `confirmed`.
This commit changes the behavior so that it defaults to `pending` instead.

### Testing

Commands, verify ens in chat

I have also set the verification of transaction commands to ropsten in e2e builds


flexsurfer (approved):

churik:
check to verify 'Confirmed' state for receiver will be added just right after https://github.com/status-im/status-react/issues/11507 - blocker for transaction tests now

qoqobolo:
### ISSUE 1: ENS aren't resolved in chats

If start 1-1 chat with ENS-user or join a public chat that contains messages from users with ENS, ENS will not be displayed.
If User adds ENS to his profile, it will appear in the header of the User Profile as expected.

#### Expected behavior
ENS are displayed in chats

#### Actual behavior
ENS aren't displayed in chats, only in own User Profile

cammellos:
@qoqobolo sorry, should be fixed int the latest build

Issue #11507, opened 2020-12-02
Transactions are failed with 429 Too Many requests too often (Ropsten)


churik:
# Bug Report

## Problem
Around 50% of transaction e2e currently failed with error  429 Too Many requests on Ropsten network in e2e builds.
![image](https://user-images.githubusercontent.com/4557972/100869798-d865eb80-349d-11eb-8b84-3c0d8cd10946.png)

As @rasom said, it can be fixed by using more tokens for Infura.

#### Logs

Session: https://app.saucelabs.com/tests/0a6cb6cd0fc64ae388ea4c5db0e2b591?auth=e095fb692e88e7f3a10f7673832e56d8#120

Logcat: 
[logcat (1).log](https://github.com/status-im/status-react/files/5629060/logcat.1.log)

cammellos:
@rasom seemsl like the best way to address this is to back-off the amount of seconds and retry, (we could create multiple tokens as well, but this seems to be the first step to try), would you agree?

churik:
seems that issue somehow related to #11478  PR
Couldn't find this failure in e2e tests before this PR.

rasom:
closed by 7055c46

PR #11516, opened 2020-12-04
[#11507] Fix e2e tests

rasom:
Currently we dispatch `new-block` event on each loaded transaction on `status-go` side, and request last 20 txs on handling of this event on `status-react` side. That caused race condition (which could be reproduce only on ropsten, probably because it is faster). So now we only requests 20 txs for the last transaction in the batch. 

status: ready

### DIFF
--- a/src/status_im/ethereum/subscriptions.cljs
+++ b/src/status_im/ethereum/subscriptions.cljs
@@ -20,26 +20,32 @@
   [{:keys [db]} id handler]
   {:db (assoc-in db [:ethereum/subscriptions id] handler)})
 
+(fx/defn max-known-block
+  [{:keys [db]} block-number]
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

Serhy:
Looks good to me when using this builds on Android 9, Android 8, iOS14: on Mainnet and Ropsten I see transaction history and when pull-to-refresh is used it's updated, can send transaction and it's updated from tx history (along with tx speed up). And e2e tests (all from https://github.com/status-im/status-react/pull/11516#issuecomment-739916074) which were failing with 429 too many requests are working now (non of them failed with 429 error)
Thanks @rasom !

Serhy (approved):

Issue #11105, opened 2020-08-20
Unread message counter is not updated for the first message after changing status of command message in 1-1 chat


qoqobolo:
# Bug Report

## Problem

If User 2 declines transaction request from User 1, navigates to Home screen, and receives one more text message from User 1, counter for this new unread message won't be shown. If User 2 receives one more text message while on Home screen, the unread message counter will show "1" instead of "2". 
However, after relogin "2" unread messages will be shown on Home screen. 

#### Expected behavior

The unread message counter counts the first message after a declined transaction request.

#### Actual behavior

![ezgif com-resize](https://user-images.githubusercontent.com/67952253/90778383-d87e1900-e2fc-11ea-8d90-97e71cf28c6b.gif)

### Reproduction

User 1: 
- Open 1-1 chat with User 2
- Send Transaction request 

User 2:
- Open 1-1 chat with User 1
- Decline Transaction request 
- Open Home view

User 1:
- Send a message to User 2

User 2:
- Check the counter of unread messages (without opening the chat)
- Relogin
- Check the counter

### Additional Information

- Status version: Nightly 20/08
- Operating System: Android 10, iOS 13

churik:
Also reproducible when user 1 shares address for example (for transaction message in 1-1 chat) and then sends message.
@cammellos take a look when you've got a chance, thank you!

Issue #9895, opened 2020-01-23
Transaction state is not updated for receiver if it is received to non-default account


churik:
# Bug Report

## Problem
When sending transaction, receiver may share for transaction `Status account` or another account.
In case of another - message state will remain `shared account` forever. 

#### Expected behavior
'Confirmed'

#### Actual behavior
![Monosnap 2020-01-23 16-30-40](https://user-images.githubusercontent.com/4557972/72998201-c20eee00-3dfd-11ea-8509-0e664069deb9.png)

### Reproduction

- User A: send transaction to User B (User B shouldn't have ENS name set)
- User B: create new account in wallet
- User B: share new account
- User A: sign and send transaction
- User B: wait until transaction gets enough confirmations, try to relogin to get updated

### Additional Information

- Status version: RC 2020012313
- Operating System: Android, iOS

churik:
More visible in PR https://github.com/status-im/status-react/pull/10100 
cc @flexsurfer

churik:
Currently, the transaction state message is not updated to "Confirmed" at all if the transaction was sent between e2e builds.

It is a blocker for `test_network_mismatch_for_send_request_commands` - it will always pass until issue will be fixed

cammellos:
I believe this is just a matter of configuring the URL for verifying transactions, @churik  do you still want to enable those e2e tests? ( if so we can get this done, otherwise we can close)

churik:
Yes, I would rather enable this case.
IMO that message in "Shared account" state forever is not user-friendly (so you can't tap on it and see transaction state).
I understand it is not high-prio because not a lot of users using non-default account in 1-1 chat, but in case of usage user will be confused for sure.

cammellos:
Looking at this.

It looks like we are not checking all wallet accounts.
The assumption is that https://github.com/status-im/status-go/blob/1237cc9ac63f3b94e367a702e06acdc72ce79fb5/services/ext/service.go#L310 should return all wallet accounts, but I see that only the default account is returned.

cammellos:
checked and confirmed that the addresses are in the database and we should be querying them correctly, so it must be that at some point we are not correctly building up the map of addresses. investigating.

cammellos:
Ok, found the issue, when generating a new account is not marked as a wallet, so it's not picked up by the validator, will fix.

PR #11524, opened 2020-12-07
[Fixes: #11105][Fixes: #9895] Fix transaction unread counter and watching requests

cammellos:
This commit fixes two issues:

- Unread counter was wrongly calculated when hiding messages
- Only the default account was watched for command transactions

status: ready


flexsurfer (approved):

churik:
@cammellos 
issue #9895 is still reproducible in this PR (tried with tokens and ETH,in android and IOS - no luck)
IMAGE 2020-12-08 12:50:17
Receivieng 0.0000002 SALT to 'Account 1'
[status_logs.zip](https://github.com/status-im/status-react/files/5659066/status_logs.zip)

cammellos:
@churik thanks, I'll try to replicate as it worked locally, but I must have missed something. 
Would you mind to replicate with debug logs on please if you have time? (otherwise no worries)
Thanks!

churik:
UPD: seems the issue is gone after relogin.
So it is reproducible only for commands messages which were created after account creation (or recovering) and until 1t relogin.
All subsequent commands messages will update their states as expected.
That's the reason why I can't attach `DEBUG` logs - after that issue will not reproducible anymore.

churik:
@cammellos issue is not reproducible anymore on both mainnet and ropsten builds (tried ETH and tokens) 
Maybe it is related to new transaction history fetching methods.

Sorry for confusion.
Ready to be merged.

Issue #10721, opened 2020-05-28
Dark Mode 1.1


errorists:
# Problem
It seems that some people find the dark mode to not have enough contrast between the blue accent colour and dark backgrounds.
tl;dr this is a result of me trying to replace all blue values with a single colour. This colour couldn't be bright enough for punchy contrast because then outgoing message bubbles or notification badges would stop being readable with the white text on them.
To overcome that, we shouldn't change the accent blue on any components that have white text inside and then make the dark theme's blue brighter. 

## Implementation
Don't change the background colour on appearance change for

* [ ] outgoing messages (the dark blue ones you send)
* [ ] any Notification Badges

(anything else you can think of that has white text on blue background?, lmk)

* [ ] Update the dark theme's accent blue colour to a brighter `#758EF0`
* [ ] Update the keycard animation with values found [here](https://www.figma.com/file/9hkPcxsqDTWa9b9c49OX3X/Dark-Mode?node-id=177%3A4278)

## Acceptance Criteria
The above was implemented and I was able to test if the new blue is contrasty enough.

## Notes
Changes are already applied in [Dark Mode Figma](https://www.figma.com/file/9hkPcxsqDTWa9b9c49OX3X/Dark-Mode?node-id=5%3A6166)

Thanks for your patience while we work out the kinks of dark mode.

hesterbruikman:
As discussed on Discord, to add here: Popover and Bottom sheet need more contrast

churik:
Popups also need more contrast, as mentioned in #11789

PR #11526, opened 2020-12-08
📱➡️ 💳 Migrate on-device accounts to keycard

### LINKED ISSUE #11431: Migrate account to a regular keycard from logged out "Manage keys and storage"
Part of: https://github.com/status-im/status-react/issues/11237

As a user 
I want to migrate my regular account to keycard when I am logged out
As I might have forgotten my password and can't access it

### Description

We want to provide users a way to migrate an existing account to a keycard 

The flow starts here: https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8374%3A86 
only `Manage keys and storage` is in scope.

When clicked the user will be asked their seed phrase:  https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A7269

and go through the flow: 

https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A7230

https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A10957

Error state: https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A10131

Success: https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=5452%3A766

### Notes

For simplicity the database can be completely removed and re-initialized, eventually we will support migrating data, but out of scope for now.

shivekkhurana:
fixes #11431 

### Summary
Add an option to "Manage Keys" where a user can move the keys from a device to a Keycard. Helps bypass the cycle where a user needs to uninstall the app and recover from seed in-order to use a Keycard. Currently, the encrypted database is deleted, ie. chats will be lost. But this can be modified later to retain chats. 

### Review notes
This PR modifies some seed phrase recovery functions. The on-boarding process should be checked for mistakes.

#### Platforms

- Android
- iOS

##### Functional

- account recovery
- new account

### Steps to test
- On the multiaccounts list screen, click on a multiaccount
- Using the "Access existing keys" button, goto "Manage keys and storage"
- Follow the UI flow (Enter seed-> Select Keycard -> Confirm -> Acknowledge risk and confirm again)
- The Kecard onboarding flow should begin
- Follow the keycard flow and setup your account

In the background, the multiaccount is deleted from the device and moved to Keycard


hesterbruikman:
Some feedback on the Draft PR for your consideration @shivekkhurana as some things might still be in development:

#### 1. Entry point
On the Sign in screen, expected is a overflow [`...`]menu that replaces the current `Access existing keys` at the bottom. Bottom sheet items are absorbed into the overflow menu that will show the 3 options currently shown under `Access existing keys`. See image below 


#### 2. Disabled state checkbox
As far as I know a disabled state for the check box was added recently. Maybe @Ferossgp can point to it. It has a lightgrey background with a dark grey check

#### 3. Popup to confirm seems to miss an overlay
I'll review this in light mode, could be a dark mode thing; I'll check what other changes might be needed for the popup to be dark mode fit

#### 4. Check derivation path
Path shown was m/43'/60'/1581'/0'/0 this seems to me the path of the chat key. In these case, we would want to show the path of the full multiaccount (believe the first is thrown out for security reasons), represented by the random name generated for the chat key associated with this multiaccount. I'm no expert in the area. Good to make sure in any case. Gravityblast can probably support on the derivation path of a multiaccount

#### 5. Migration failed message
After confirming to move the keystore file, but not completing Keycard setup (Keycard already had a file stored) I'd expect the error in the image below. Can you tell me if this is possible? Would requiring checking if Keycard setup was initiated from key management, when setup is dismissed. I currently got the error: sql: no rows in result set

shivekkhurana:
1. Will make the change ✅ 
2. There is a disabled checkbox in Quo and I have used it for "Reset Database" option ✅ 
3. Missing overlay could be because of the dark mode. Did you check other popups in dark mode ⚠️ 
4. I added the path for the chat key, will replace it with master key derivation path ✅ 
5. Throwing an error is technically possible. I'll have to maintain some state about Keycard onboarding initialization and then catch the final error event. Will look into this more ⚠️

shivekkhurana:
@hesterbruikman 
Re: 1
In the current implementation, there is a dotted menu on the multi account list page with just one option to "Generate a new account". Should we merge the two menus? Total of 4 options on both list and login screen ?

hesterbruikman:
>  In the current implementation, there is a dotted menu on the multi account list page with just one option to "Generate a new account". Should we merge the two menus? Total of 4 options on both list and login screen ?

I'd keep them separate. The option to go to key management on the multiaccount list page will require changing the flow as no account has been selected to manage keys for yet. For now, I'd leave this option only available after a multiaccount has been selected either by being on the sign in screen, entering an existing seed phrase or being in the Profile/Settings of this account

gravityblast:
> Path shown was m/43'/60'/1581'/0'/0 this seems to me the path of the chat key. In these case, we would want to show the path of the full multiaccount (believe the first is thrown out for security reasons), represented by the random name generated for the chat key associated with this multiaccount. I'm no expert in the area. Good to make sure in any case. Gravityblast can probably support on the derivation path of a multiaccount

@hesterbruikman in theory the full multiaccount would be represented by the master key `m` but it wouldn't give much info. But the 3 words name is derived from the chat key, so I think that path should be good right?

hesterbruikman:
> 
> 
> > Path shown was m/43'/60'/1581'/0'/0 this seems to me the path of the chat key. In these case, we would want to show the path of the full multiaccount (believe the first is thrown out for security reasons), represented by the random name generated for the chat key associated with this multiaccount. I'm no expert in the area. Good to make sure in any case. Gravityblast can probably support on the derivation path of a multiaccount
> 
> @hesterbruikman in theory the full multiaccount would be represented by the master key `m` but it wouldn't give much info. But the 3 words name is derived from the chat key, so I think that path should be good right?

Thanks for joining @gravityblast . It's a tricky one. As we use the random name not only as a human readable form of your chat key, but also as identifier of the multiaccount; As it's used on the sign in screen. In this case, we want to indicate that you're moving the full multiaccount, which the user can distinguish from others through the random name. 

The intention is also to be future proof. Imagine an option in addition to move keystore file and reset database to select a path to add additional authentication (e.g. requires signing with Keycard while keys are still on Mobile or Desktop). In this case I think you'd want to know for what path you're adding this.

Maybe until that point, a description like `full tree` is better than showing a derivation path. Wdyt? @shivekkhurana @gravityblast

hesterbruikman:
@shivekkhurana with regards to issue 3, I think you can treat it as out of scope of this PR. I discussed with Maciej to include this in a collection of Dark mode improvements captured in #10721

shivekkhurana:
> Maybe until that point, a description like `full tree` is better than showing a derivation path. Wdyt? @shivekkhurana @gravityblast

I feel `Full tree` is more appropriate than chat key path. 

We can also indicate it as `1 root key and 4 derived keys` and maybe list paths of all derived keys.

hesterbruikman:
> We can also indicate it as `1 root key and 4 derived keys`

I think that'll work well. I propose below which is a variation of this. Referring to keypairs instead. Also, looking at the screen again I suggest using Multiaccount instead of Master account as it's more accurate and doesn't introduce a new term. (Even though we've tried to stay away from all multiaccount references, I think this is enough layers deep that the level of detail and accuracy makes sense

Key management - choose actions (details expanded)

shivekkhurana:
The label `Derivation paths` is odd. How does `Number of keys` sound ?

gravityblast:
> Maybe until that point, a description like full tree is better than showing a derivation path. Wdyt? @shivekkhurana @gravityblast

@hesterbruikman I agree, that's more intuitive than using a derivation path.
I wouldn't specify the number of keys because we are importing the master key to the card, so we will be able then to use any derivation path, not only the ones already used.

hesterbruikman:
@gravityblast I see what you mean. What do you think about if it's phrased like this?: `Full tree, includes 4 active accounts'

Key management - choose actions (details expanded)

gravityblast:
> @gravityblast I see what you mean. What do you think about if it's phrased like this?: `Full tree, includes 4 active accounts'

@hesterbruikman yeah I think something like this would make sense.

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+            [status-im.popover.core :as popover]
+            [status-im.utils.fx :as fx]
+            [status-im.utils.security :as security]
+            [status-im.utils.types :as types]))
+
+(fx/defn key-and-storage-management-pressed
+  "This event can be dispatched before login and from profile and needs to redirect accordingly"
+  {:events [::key-and-storage-management-pressed]}
+  [cofx]
+  (fx/merge cofx
```
cammellos: I don't think you need `fx/merge` in this case, only when composing `fxs`.
Here you should be able to do:
```
(navigation/naviate-to-cofx cofx :key....
```

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+(fx/defn move-keystore-checked
+  {:events [::move-keystore-checked]}
+  [{:keys [db] :as cofx} checked?]
+  {:db (assoc-in db [:multiaccounts/key-storage :move-keystore-checked?] checked?)})
+
+(fx/defn enter-seed-pressed
+  "User is logged out and probably wants to move multiaccount to Keycard. Navigate to enter seed phrase screen"
+  {:events [::enter-seed-pressed]}
+  [cofx]
+  (fx/merge cofx
```
cammellos: Similar, no need to of `fx/merge`

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+
+(fx/defn seed-phrase-input-changed
+  {:events [::seed-phrase-input-changed]}
+  [{:keys [db] :as cofx} masked-seed-phrase]
+  (let [seed-phrase (security/safe-unmask-data masked-seed-phrase)]
+    {:db (update db :multiaccounts/key-storage assoc
+                 :seed-phrase (when seed-phrase
+                                (string/lower-case seed-phrase))
+                 :seed-shape-invalid? (or (nil? seed-phrase)
+                                          (empty? seed-phrase)
```
cammellos: You can use `(empty? seed-phrase)` instead of both `(empty? nil) -> true`

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+:m.r/re-encrypt-pressed
+:i/on-key-storage-selected ([:intro-wizard :selected-storage-type] is set to :advanced)
+:m.r/select-storage-next-pressed
+
+We don't need to take the exact steps, just set the required state and redirect to correct screen
+"
+(fx/defn handle-delete-multiaccount-success
+  {:events [::delete-multiaccount-success]}
+  [{:keys [db] :as cofx} _]
+  (fx/merge cofx
```
cammellos: not needed `fx/merge`

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+                      (assoc-in [:keycard :flow] :recovery)
+                      (dissoc :multiaccounts/key-storage))}
+            (popover/hide-popover)
+            (navigation/navigate-to-cofx :intro-stack {:screen :keycard-onboarding-intro})))
+
+;; How to handle this?
+(fx/defn handle-delete-multiaccount-error
+  {:events [::delete-multiaccount-error]}
+  [_ _]
+  {})
```
cammellos: We at least want to log something, but yes, that's a good point

src/status_im/native_module/core.cljs
```diff
@@ -398,3 +398,4 @@
   (log/debug "[native-module] resetKeyboardInput")
   (when platform/android?
     (.resetKeyboardInputCursor ^js (status) input selection)))
+
```
cammellos: extra newline

src/status_im/ui/components/accordion.cljs
```diff
@@ -0,0 +1,36 @@
...
+            :color           colors/black}]]}]
+       (when @opened?
+         content)])))
+
+(defn accordion
+  "List of collapseable sections"
+  []
+  ;; TODO(shivekkhurana): Extract status-im.ui.screens.wallet.recipient.views/accordion component here
+  )
+
```
cammellos: extra newline

cammellos (approved):
Looks good to me, we don't generally use `comment` but it does help making things clearer.
Just a few style comments.

src/status_im/multiaccounts/key_storage/core.cljs
shivekkhurana: I added this nil clause while writing tests. Probably mixed some things up, but good catch. This is not needed.

shivekkhurana:
@hesterbruikman 
Re: Failed migration popup: https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A10131

Reaching a point where we are not able to migrate, doesn't necessarily mean that the multiaccount has been deleted. In fact, the user can reach this error because the app was not able to delete a multiaccount.

In the current implementation, this error is raised when the deletion of multi-account fails. Which is vague itself and may or may not mean that the account was burnt from device memory. 

Not sure how to handle this, but the popover doesn't feel okay.  It's more like "We don't know what went wrong and your account may have been deleted".

shivekkhurana:
@hesterbruikman 
> What do you think about if it's phrased like this?: `Full tree, includes 4 active accounts'

This flow is activated when the user is logged out, so we don't really know the number of active accounts (because DB is still encrypted). We can show this message when the user is logged in and tries key-storage from profile settings.

To be able to show the exact number of account while being logged in includes maintaining this number in async storage for all accounts (which is complex). 

@gravityblast suggested we can show "The master key will be moved" while logged out and the actual number of keys while logged in. What do you say?

hesterbruikman:
> @gravityblast suggested we can show "The master key will be moved" while logged out and the actual number of keys while logged in. What do you say?

In that case, I'd prefer to hide the `derivation paths` list item in the detailed account overview. As there is also the more highlevel `Type: Multiaccount`

hesterbruikman:
> Not sure how to handle this, but the popover doesn't feel okay. It's more like "We don't know what went wrong and your account may have been deleted".

I'd like to understand what we know at this point and what the user can do, so we can devise a more accurate and informative message. Can you please check and correct the statements below @shivekkhurana ?:

##### Information
- Migration was attempted
- Keys for which migration was attempted can currently not be found on device
- Keys for which migration was attempted were not on a paired device, last we checked (e.g. when you tapped Keycard)
- Your funds are safe
- Your contacts, chats, messages of >30 days are lost

##### Actions
- Use seed phrase to import again and select preferred storage location

shivekkhurana:
> I'd like to understand what we know at this point and what the user can do, so we can devise a more accurate and informative message. Can you please check and correct the statements below @shivekkhurana ?:

## Information
The cause for this error is inability to delete multiaccount. We don't have any clue about why we were not able to delete the multiaccount. Also, I don't think this error will happen, unless status-go goes rogue.

The exact state we want to convey is:
- We tried to delete a multiaccount
- But were not able to (or maybe the ma was deleted but something else broke on the way)

## Possible actions
- To go back to Home Screen and try again if you see the multi account (ie. chats, etc are not deleted) or 
- To recover from seed phrase if you don't see the ma in the list (ie chats, etc are deleted)

hesterbruikman:
Apologies for the wait. How about this @shivekkhurana 

Key management - system failed

qoqobolo:
### ISSUE 1: User cannot re-login if the migration process was interrupted 

If the migration process was interrupted after clicking the `Move and reset` button and then entering the regular password, then after the next successful migration attempt User won't be able to re-login

_**Steps:**_

1. Recover multiaccount with seed prase and storage on the device
2. Log out
3. Open the app
4. Tap `three dots` at the top right on the password screen
5. Tap `Manage keys and storage`
6. Enable `Move keystore file` checkbox
7. Enter seed phrase
8. Choose storage > Keycard
9. Tap `Confirm`
10. Tap `Move and reset`
11. Tap `Back`or kill the app
12. Try to login (Error "_sql: no rows in result set_" is shown)
13. Tap `three dots` at the top right on the password screen
14. Follow the keycard flow and setup account
15. Re-login

#### Expected result
1) IMO we should warn user that the migration process should not be interrupted 
2) Can re-login even in case of the migration process was interrupted 

#### Actual result
1) Can interrupt the migration process at any moment
2) No warning
3) Error "_sql: no rows in result set_" 
4) Cannot re-login after interrupted migration process

[geth.log](https://github.com/status-im/status-react/files/5751099/geth.log)
[Status.log](https://github.com/status-im/status-react/files/5751100/Status.log)

 [Full video with reproducing](https://drive.google.com/file/d/12Z_x5_QsjzJZwPZ3ZyKz4rT-4jS6LfeB/view?usp=sharing)

Screenshot 2020-12-29 at 15 55 54

qoqobolo:
### ISSUE 2: Tokens are not displayed in the Wallet after migration until using "Scan Tokens" 

_**Steps**_:

1. Recover multiaccount with seed phrase and choose storage on the device
2. Move the key from a device to a Keycard (Migration flow)
3. Navigate to the Wallet view

#### Expected behavior
Tokens are displayed in the wallet immediately after opening the view without using "Scan tokens"

#### Actual behavior
Need to use "Scan tokens" to see tokens


https://user-images.githubusercontent.com/67952253/103296641-c1f87480-49f6-11eb-928d-5a82ab1f232d.mp4

qoqobolo:
### ISSUE 3: Error after log-out


_**Steps:**_

1. Recover multiaccount with a seed phrase
2. Log out

#### Expected behavior
No error after log-out

#### Actual behavior

The issue is reproducible in 10-20% of cases with and without migration. 

There are two types of errors:
1. Error "Cannot read property 'toLowerCase' of null"
[geth.log](https://github.com/status-im/status-react/files/5751364/geth.log)
[Status.log](https://github.com/status-im/status-react/files/5751365/Status.log)

Screenshot 2020-12-29 at 17 30 33


2. Error "Please shake your phone to report..."
[geth.log](https://github.com/status-im/status-react/files/5751368/geth.log)
[Status.log](https://github.com/status-im/status-react/files/5751369/Status.log)

Screenshot 2020-12-29 at 17 30 02

qoqobolo:
### ISSUE 4: Cannot add an account to the wallet with a seed phrase

_**Steps:**_

1. Recover multiaccount with a seed phrase
2. Open Wallet
3. Tap "Add an account"
4. In the bottom sheet click "Enter a seed phrase"
5. Enter a password
6. Paste a seed phrase
7. Enter an account name
8. Tap "Add an account"


#### Expected result
- New account is added
- Wallet screen is opened

#### Actual result
When clicking the "Add an account" button nothing happens


https://user-images.githubusercontent.com/67952253/103345070-ab9ef700-4a90-11eb-87f5-cd00f500f636.mp4

shivekkhurana:
Re : [QA Issue 1 ⬆️ ](https://github.com/status-im/status-react/pull/11526#issuecomment-752111000) **Interrupt migration** 

There is no way to be certain that the multiaccount was deleted in case of interrupts (app killed, back button pressed, phone call, dead battery, phone dropped in a pool?).

Diana was able to reach an odd state where the encrypted db is deleted but the multiaccount entry was not deleted from open db. If this happens in real-world, the orphan account will stay there until the app is deleted.

To bypass the issue, I'm thinking about an async storage entry that marks a multiaccount for deletion and clears on success. Then in the app initialisation process, the app will check if there are any multiaccounts pending for deletion and delete them.

I'm not sure if this is the best way to go about it. Like this won't solve the issue if back button is pressed after `move and reset`. 

While typing this out, I realised that instead of an async storage flag, we can just have a background tasks queue (simple core.async channel persisted on disk), which will also serve as the backbone for anonymous metrics.

shivekkhurana:
Re: [QA Issue 4 ⬆️ ](https://github.com/status-im/status-react/pull/11526#issuecomment-752405664) **Unable to add seed** - Fixed ✅

shivekkhurana:
Re: [QA Issue 3 ⬆️](https://github.com/status-im/status-react/pull/11526#issuecomment-752147834) - **Error on logout**

I checked the logs and am certain that the issue was not introduced in this PR. It should be looked at independently 🌚

shivekkhurana:
Re: [Issue 2](https://github.com/status-im/status-react/pull/11526#issuecomment-752134046) - **Tokens not displayed until scan tokens is pressed** - Fixed ✅

shivekkhurana:
Re: Issue 1 - User not able to login on back button press

The problem was not related to database, but to React state. The account gets deleted but the React state is not persisted. 

After discussion with Andrey, we came to a solution - hide back button and disable device back handler. I have implemented this so the problem should go away.

src/status_im/events.cljs
```diff
@@ -163,8 +163,10 @@
  (fn [cofx _]
    (multiaccounts.logout/logout cofx)))
 
+
```
flexsurfer: why empty lines?

src/status_im/keycard/core.cljs
```diff
@@ -528,3 +529,17 @@
...
+(defn onboarding-intro-back-handler
+  "The back button handler is used to manage device back press.
+
+  If the handler returns false, the back button functions as usual (ie. dispatchs GO_BACK event).
+  If it returns true, the back button becomes inactive.
+
+  We want to deactivate the back button when the user comes from key-storage and migration flow."
+  []
+  (-> re-frame.db/app-db
+      deref
```
flexsurfer: just wondering why do you prefer `deref` under the `@` ?

src/status_im/multiaccounts/core.cljs
```diff
@@ -201,3 +207,7 @@
   {:events [::update-local-picture]}
   [cofx pics]
   (multiaccounts.update/optimistic cofx :images pics))
+
+(comment
+  ;; Test seed for Dim Venerated Yaffle, it's not here by mistaka, this is just a test account
```
flexsurfer: why here ?

src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -0,0 +1,182 @@
...
+
+(fx/defn goto-multiaccounts-screen
+  {:events [::hide-popover-and-goto-multiaccounts-screen]}
+  [cofx _]
+  (fx/merge cofx
+            (popover/hide-popover)
+            (navigation/navigate-to-cofx :intro-stack {:screen :multiaccounts})))
+
+(comment
+  (mnemonic/sanitize-passphrase "rocket rebel pasta kimchi kitty nani tokyo")
```
flexsurfer: why not in tests ?

src/status_im/multiaccounts/recover/core.cljs
```diff
@@ -269,3 +266,17 @@
...
+(defn login-ma-keycard-pairing
+  "Compute the keycard-pairing value of the multiaccount selected for login"
+  [db _]
+  (when-let [acc-to-login (-> db :multiaccounts/login)]
+    (-> db
+        :multiaccounts/multiaccounts
+        (get (:key-uid acc-to-login))
+        :keycard-pairing)))
+
+(re-frame/reg-sub
```
flexsurfer: all subs must be in subs.cljs

src/status_im/ui/components/accordion.cljs
```diff
@@ -0,0 +1,35 @@
...
+                              :justify-content :center}
+            :resize-mode     :center
+            :color           colors/black}]]}]
+       (when @opened?
+         content)])))
+
+(defn accordion
+  "List of collapseable sections"
+  []
+  ;; TODO(shivekkhurana): Extract status-im.ui.screens.wallet.recipient.views/accordion component here
```
flexsurfer, line 34: not in this PR ?

src/status_im/ui/screens/intro/views.cljs
```diff
@@ -88,7 +88,7 @@
                         :icon                [react/image {:source      {:uri (identicon/identicon public-key)}
                                                            :resize-mode :cover
                                                            :style       styles/multiaccount-image}]}]))]])
-
+;; TODO(shivekkhurana): Update this to use `mul`
```
flexsurfer: ?

src/status_im/ui/screens/keycard/onboarding/views.cljs
```diff
@@ -16,12 +16,15 @@
   (:require-macros [status-im.utils.views :refer [defview letsubs]]))
 
 (defview intro []
-  (letsubs [flow [:keycard-flow]]
+  (letsubs [flow [:keycard-flow]
+            {:keys [from-key-storage-and-migration?]} [:keycard]]
     [react/view styles/container
-     [topbar/topbar]
+     (when-not from-key-storage-and-migration?
+       [topbar/topbar])
```
flexsurfer, line 23: i feel like from the design we still need topbar space but without navigation

src/status_im/ui/screens/multiaccounts/key_storage/views.cljs
```diff
@@ -0,0 +1,344 @@
...
+                 :accessibility-label :learn-more
+                 :on-press #(js/alert :press)}
+     (i18n/label :learn-more)]
+    [vector-icons/icon :main-icons/tiny-external {:color  colors/blue
+                                                  :width  16
+                                                  :height 16}]]])
+
+(defn keycard-upsell-banner []
+  [react/touchable-highlight {:on-press #(.openURL ^js react/linking "https://keycard.tech/")}
+   [react/view {:background-color (if (= :dark @colors/theme) "#2C5955" "#DDF8F4")
```
flexsurfer, line 141: all colors must be in colors.cljs

src/status_im/ui/screens/multiaccounts/key_storage/views.cljs
```diff
@@ -0,0 +1,344 @@
...
+     [react/text {:style styles/popover-text}
+      (i18n/label :t/transfer-ma-unknown-error-desc-2)]]]
+   [react/view {:margin-vertical 24
+                :align-items     :center}
+    [quo/button {:on-press #(re-frame/dispatch [::multiaccounts.key-storage/hide-popover-and-goto-multiaccounts-screen])
+                 :type     :secondary}
+     (i18n/label :t/okay)]]])
+
+(comment
+  (-> re-frame.db/app-db deref keys)
```
flexsurfer: if not for test, could we move this in separate ns ?

flexsurfer (changes requested):
great job! left a few minor comments

src/status_im/events.cljs
shivekkhurana: No reason. I must have added something here and forgot to delete the empty line. Will fix it.

src/status_im/multiaccounts/core.cljs
shivekkhurana: This is inspired by a talk titled "[Running with Scissors](https://www.youtube.com/watch?v=Qx0-pViyIDU&ab_channel=StrangeLoop)" by Stuart Halloway. The idea is to capture your dev flow in git. 

While building the feature, I had to check the functioning of the `validate-mnemonic` function, so I captured it in a rich comment. If there is a bug, or we need to extend this feature, chances are that the ticket owner will need to check this function again. But the next time, it will be present in in a rich comment. 

But I also agree that it makes little sense in the this `ma.core` ns. So  I can get rid of it if there is any issue.

src/status_im/multiaccounts/recover/core.cljs
shivekkhurana: Roger that captain.

src/status_im/ui/components/accordion.cljs
shivekkhurana: No, this PR is huge already. But this is on my todo.

src/status_im/ui/screens/keycard/onboarding/views.cljs
shivekkhurana: I have added space to the container. See next few lines (look out for `margin-top`).

src/status_im/keycard/core.cljs
shivekkhurana: No particular reason, I just wrote it while I was testing it out. Have replaced `deref` with `@`. That's more crisp.

src/status_im/multiaccounts/recover/core.cljs
shivekkhurana: Fixed

src/status_im/ui/screens/intro/views.cljs
shivekkhurana: Dangling comment. Removed.

src/status_im/ui/screens/multiaccounts/key_storage/views.cljs
shivekkhurana: I have deleted db lookups. Re-frisk is much better for tasks like these.

src/status_im/ui/screens/multiaccounts/key_storage/views.cljs
shivekkhurana: I talked to Gheorghe about this and he said that it's best to put it in the component itself, because nothing else depends on it: https://discord.com/channels/624205794384281629/626738033701683200/786532503379771393

Happy to move it around.

flexsurfer (approved):

shivekkhurana:
The UI changed a bit (see point 1 on [this comment](https://github.com/status-im/status-react/pull/11526#issuecomment-744450090)). The `Access existing keys` button on login screen was removed. To get that menu, the user can press the three dots button `...` on the top right.

It could be that the e2e tests are looking for the old button.

qoqobolo:
@shivekkhurana tests 6292, 5742, 6251, 5689, 6293, and 6294 failed because the `Begin Setup` button on the "Store your keys on the Keycard" screen was shifted out of the screen. 
This issue isn't reproducible on the real device (Google Pixel 2 at least), but it's consistently reproducible on the emulator (device screen size: 768x1280).
Could you check please if any changes could affect this screen as well as fix it?

Other tests are not related to PR.

Steps:

1. Install Status
2. `Get started`
3. `Generate keys`
4. Select some three random name
5. Tap `Next`
6. `Keycard`
7. `Next`

Screenshot 2021-01-14 at 13 56 12

shivekkhurana:
Yeah, I added an 80px top margin to this screen because it was reused at a place where there is no top bar. 
Fixing it.

qoqobolo:
Tested:

- recover with seed prase > migration
- reaction a new multiaccount > migration
- login after migration
- migration > uninstall the app> install > recover with kk
- migration > sign tx
- migration > pairing > sign tx
- migration > tx in 1-1
- adding a new acc to wallet
- recover with seed phrases with different number of words (12, 24)
- creation new kk-multiaccount 
- wrong keycard 

@shivekkhurana thank you for your awesome work and patience!
PR is ready to be merged

flexsurfer:
great job @shivekkhurana . please squash commits before merge!

shivekkhurana:
This flow is missing a screen that notifies the user that migration was successful:
https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8192%3A466

PR #11529, opened 2020-12-08
[Fixes: #11520] Rotate mailservers on error conditions

### LINKED ISSUE #11520: Don't keep using the same mailservers if the requests are failing
# Bug Report

## Problem

We had a mailserver outage recently whereby 2 mailservers (central-02 & central-03) were throwing errors upon request.

The system correctly swapped mailserver when it was happening, but it would keep re-selecting them as they had the best ping time for some users.

#### Expected behavior

Mailservers that are not responding correctly are not re-selected for at least some period of time.

#### Actual behavior

Mailserver kept being selected because the fastest in ping time.

#### Notes

We should probably implement some cooldown timer where mailservers should not be re-selected if they are erroring.

cammellos:
Fixes: #11520

If a mailserver request is failing we disconnect at the 3rd attempt.

Though once we are disconnected, we connect again to the same mailserver
based on ping time.

This commit changes the behavior so that mailservers are penalized if
they are returning errors.

### Testing

Test mailservers , going offline/online etc

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -100,10 +100,11 @@
                  :method  "admin_removePeer"
                  :params  [enode]}
         payload (.stringify js/JSON (clj->js args))]
-    (status/call-private-rpc payload
-                             (handlers/response-handler
-                              #(log/info "mailserver: remove-peer success" %)
-                              #(log/error "mailserver: remove-peer error" %)))))
+    (when enode
+      (status/call-private-rpc payload
+                               (handlers/response-handler
+                                #(log/info "mailserver: remove-peer success" %)
+                                #(log/error "mailserver: remove-peer error" %))))))
 
 (re-frame/reg-fx
  :mailserver/add-peer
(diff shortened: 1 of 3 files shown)


src/status_im/mailserver/core.cljs
```diff
@@ -575,10 +601,14 @@
...
-                 (<= (- now last-connection-attempt)))
-        (fx/merge cofx
-                  (when (not= :connected (:mailserver/state db))
-                    (change-mailserver)))))))
+                 ;; We are not connected
+                 (not= :connected (:mailserver/state db))
+                 ;; We either never tried to connect to this peer
+                 (or (nil? last-connection-attempt)
+                     ;; Or 30 seconds have passed and no luck
+                     (<= (- now last-connection-attempt) (/ (* constants/connection-timeout 3) 1000))))
```
flexsurfer: now in ms ? why we don't divide it as well ?

cammellos: you are right, we should not divide it at all, nice catch!

cammellos:
@flexsurfer ready to be reviewed again, thanks!

flexsurfer (approved):

shivekkhurana (approved):
Looks good to me.

PR #11530, opened 2020-12-08
Add buy crypto banner UI

### LINKED ISSUE #10505: Link to buy crypto on exchange
Enable shortest path to buy crypto. Include a banner to a category on dap.ps

### User stories
As a user
I want to be directed to where I can buy crypto
because some functionality in the app requires holding crypto

As a user
I'm more likely to return to the application when it holds value for me

### Implementation
Design on [Figma](https://www.figma.com/file/XUehMnhyD1FGcWzvGz6SXqvh/Wallet?node-id=10626%3A0)

Banner remains until dismissed

### Future steps
- Phase 0: Include link to category on dap.ps in chat bot
- Phase 1: Allow existing users in Brazil and South Korea to purchase the amount up to the maximum that requires no or limited KYC in ETH (SNT optional).
- Phase 2: Expand to other countries and regions
- Phase X: Token swaps

Ferossgp:
Fixes #10505


src/status_im/ui/screens/wallet/events.cljs
```diff
@@ -18,3 +18,9 @@
  ::wallet-add-custom-token
  (fn [{:keys [db]}]
    {:db (dissoc db :wallet/custom-token-screen)}))
+
+(fx/defn hide-buy-crypto
+  {:events [::hide-buy-crypto]}
+  [{:keys [db]}]
+  ;; TODO: Think where to persist this preference
```
Ferossgp: @cammellos what do you think, where it is better to persist this?

cammellos: not sure, I would probably store it in `asyncStorage` as it seems very mobile centric, we could move it to status-go eventually. 
In terms of UX as a user i would expect that once I dismiss it, I won't see it anymore, even after restarting the app.
What do you think?

Ferossgp: Added `asyncStorage`

translations/en.json
```diff
@@ -775,6 +775,8 @@
     "seed-phrase-content": "A set of friendly-to-read words, randomly selected from the BIP39 standard list and used to recover or access your Ethereum account on other wallets and devices. Also referred to as a “mnemonic phrase,” “recovery phrase” or “wallet backup” across the crypto ecosystem. Most crypto apps use this same standard to generate accounts.",
     "wallet-key-title": "Account address",
     "wallet-key-content": "A 64 character hex address based on the Ethereum standard and beginning with 0x. Public-facing, your account address is shared with others when you want to receive funds. Also referred to as an “Ethereum address” or “wallet address.”",
+    "buy-crypto-title": "Looks like your wallet is empty",
+    "buy-crypt-description": "Find a dapp to buy crypto now",
```
hesterbruikman: don't know how I can edit this myself, but there's an 'o' missing in 779 which causes a translation missing error

src/status_im/async_storage/core.cljs
```diff
@@ -0,0 +1,46 @@
...
+            [status-im.async-storage.transit :refer [clj->transit transit->clj]]
+            ["@react-native-community/async-storage" :default async-storage]))
+
+(def ^:private debounce-ms 250)
+
+(def key->string str)
+
+(defn- set-item! [key value]
+  (-> ^js async-storage
+      (.setItem (key->string key)
```
flexsurfer, line 14: so these keys will be shared between multiaccounts ?

flexsurfer: if you hide banner it won't be visible in the new multiaccount right ?

Ferossgp: Yes, it is on the app level

churik:
@Ferossgp 
can toy please rebase it to have latest develop in PR? 
Can be important due to new browser

churik:
@Ferossgp good work, thank you!
tested:
- for new empty multiaccounts (keycard, non-keycard)
- for restored empty multiaccounts (keycard, non-keycard)
- dismissing banner dismiss it for all subsequent multiaccounts in this device
- link leads to Crypto Ornams
- banner is not shown for multiaccounts with ETH or SNT
- banner disappears after receiving funds to wallet

 @hesterbruikman I have one question before sending this PR to merge - should this banner will be available on POA/xdai network? 
Currently it is available and leads to empty category in dap.ps

hesterbruikman:
> @hesterbruikman I have one question before sending this PR to merge - should this banner will be available on POA/xdai network?
> Currently it is available and leads to empty category in dap.ps

Good catch @churik! Awkwardly, I would say yes. We should rather fix the issue that all categories are empty in dap.ps else we create too many dependencies. I'll keep an eye out to see if the issue persists after the refactoring is deployed https://github.com/dap-ps/discover/pull/87

cc @johnlea-quiup @guylouis

cammellos (approved):

PR #11562, opened 2020-12-18
shorter, simpler pairing passwords for Keycard

### LINKED ISSUE #9703: Keycard pairing code: evaluate if we simplify pairing code format
**Context** 
When onboarding with a brand new keycard
- status app is generating a pairing code for this card and sets it up in the card 
- the user is asked to write it down (he's not asked to type it in, because that's not necessary technically speaking)

when a user wants to use this card with another phone with status, he'll be asked to enter the pairing code 

**Problem**
The pairing code format generated by status is 20 letters or numbers, and is case sensitive. This is quite complex and can create human error writing down or typing it 

**Solution**
Should we simplify the format of the pairing code while maitaining a sufficient level of entropy (ux-security tradeoff) ? 

remark: other apps like Gnosis can initiate keycards, and are free (given keycard api) to setit up in any format. So we should keep Status entry form free format too, for the case where a status user wants to pair with a card generated on Gnosis.

bitgamma:
fixes #9703

status: ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "react-native-screens": "^2.10.1",
     "react-native-shake": "^3.3.1",
     "react-native-splash-screen": "^3.2.0",
-    "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#v2.5.27",
+    "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#v2.5.28",
     "react-native-svg": "^9.8.4",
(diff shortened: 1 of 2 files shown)


gravityblast (approved):

churik:
Thank you @bitgamma 
Tested:
- Recovering multiaccount with old format of pairing code
- frozen card flows
- creating multiaccount with KK and recovering it with pairing code of new format
- restoring multiaccount from seed with KK and recovering it with pairing code of new format
- changing PIN flows

**Devices:**

PR #11574, opened 2020-12-23
[#11379] Add a watch only account from an ENS name

### LINKED ISSUE #11379: Add a watch only account from an ENS name
As a user 
I want to add a watch-only account from an ENS name
So that I can easily watch the address of a user associated with that ENS name

### Description

I want to be able to add a watch only account from an ENS name
`Use this address` should be grayed out if the field is not a well formatted address/ens-name

rasom:
fix #11379

status: ready

### DIFF
--- a/src/status_im/wallet/accounts/core.cljs
+++ b/src/status_im/wallet/accounts/core.cljs
@@ -19,7 +19,11 @@
             [status-im.multiaccounts.recover.core :as recover]
             [status-im.ethereum.mnemonic :as mnemonic]
             [taoensso.timbre :as log]
-            [status-im.wallet.prices :as prices]))
+            [status-im.wallet.prices :as prices]
+            [status-im.utils.hex :as hex]
+            [status-im.ethereum.ens :as ens]
+            [status-im.ens.core :as ens.core]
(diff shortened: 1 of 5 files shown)


Ferossgp (approved):

src/status_im/wallet/accounts/core.cljs
```diff
@@ -19,7 +19,17 @@
             [status-im.multiaccounts.recover.core :as recover]
             [status-im.ethereum.mnemonic :as mnemonic]
             [taoensso.timbre :as log]
-            [status-im.wallet.prices :as prices]))
+            [status-im.wallet.prices :as prices]
+            [status-im.utils.hex :as hex]
+            [status-im.ethereum.ens :as ens]
+            [status-im.ethereum.resolver :as resolver]))
+
+(re-frame/reg-fx
```
Ferossgp: We have a similar to this one, an effect in `status-im.ens.core` and other places too, I think for future easier maintenance maybe it would make sense to use it only from `ens` namespaces which will get the registry and parse name.

rasom: fixed

PR #11588, opened 2020-12-28
Hide option to increase gas on keycard

cammellos:
currently is not supported on keycard,
I haven't looked much at what the issue might be and whether it requires changes on keycard, but for now I am disabling it as to be included in release, feel free to fix if it's a simple change.
status: ready

### DIFF
--- a/src/status_im/ui/screens/wallet/transactions/views.cljs
+++ b/src/status_im/ui/screens/wallet/transactions/views.cljs
@@ -38,7 +38,8 @@
 (defn render-transaction
   [{:keys [label contact address contact-accessibility-label
            address-accessibility-label currency-text amount-text
-           time-formatted on-touch-fn type hash]}]
+           time-formatted on-touch-fn type hash]}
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/wallet/transactions/views.cljs
```diff
@@ -121,7 +125,7 @@
      [list/section-list
       {:sections   transactions-history-sections
        :key-fn     :hash
-       :render-fn  render-transaction
+       :render-fn  (partial render-transaction keycard-account?)
```
Ferossgp: I think Andrey added `render-data` option to avoid functions creation on lists as it was a render performance bottleneck

Ferossgp (approved):

churik:
@cammellos 
on build with commit https://github.com/status-im/status-react/commit/fd75ae71f032b1f2037fc28f77e902b61384ba53 on keycard account options are still there
IMAGE 2020-12-28 13:03:48

Issue #11579, opened 2020-12-24
"Save password" is not saved if enable "Lock app with fingerprint" from Settings


churik:
# Bug Report

## Problem
Working as expected on 1.9.

When enabling biometric authentication from "Privacy and security" it is not saved after relogin
#### Expected behavior
"Save password" is enabled

#### Actual behavior
"Save password" is not enabled

![IMAGE 2020-12-24 11:51:18](https://user-images.githubusercontent.com/4557972/103084069-55d4d580-45de-11eb-8292-1f06953c5b66.jpg)

### Reproduction

- Open Status, create multiaccount
- Go to Profile > Privacy and security
- Switch on "Lock app with fingerprint"

### Additional Information

- Status version: 1.10
- Operating System: Android, iOS

PR #11589, opened 2020-12-28
[Fixes: #11579] Add only certain fields to multiaccount on login

cammellos:
`multiaccounts` has some fields that should not be overriden
(`save-password`), so we expliciltly `assoc` only certain fields.

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -139,13 +139,17 @@
 (handlers/register-handler-fx
  :multiaccounts.login.ui/multiaccount-selected
  (fn [{:keys [db] :as cofx} [_ key-uid]]
-   (let [multiaccount (get-in db [:multiaccounts/multiaccounts key-uid])]
+   ;; We specifically pass a bunch of fields instead of the whole multiaccount
+   ;; as we want store some fields in multiaccount that are not here
+   (let [multiaccount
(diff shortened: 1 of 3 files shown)


src/status_im/events.cljs
```diff
@@ -139,13 +139,16 @@
 (handlers/register-handler-fx
  :multiaccounts.login.ui/multiaccount-selected
  (fn [{:keys [db] :as cofx} [_ key-uid]]
-   (let [multiaccount (get-in db [:multiaccounts/multiaccounts key-uid])]
+   ;; We specifically pass a bunch of fields instead of the whole multiaccount
```
Ferossgp, line 142: I would add also in this PR `images` field similarly to the rest, so we would show the image on the login screen which is shown on multiaccounts screen.

Ferossgp (approved):

src/status_im/events.cljs
cammellos: Added, thank you

churik:
@cammellos issue is fixed, new regression issue is found.
for all users default picture on login screen is the same

BAACAgQAAxkBAAEusc9f6fIe_hQBm9kDD32auVOvtL9afAACvgYAAnbkUVNhAWFw_TROeh4E_1609167430 54

churik:
ready to be included to 1.10.
thanks, @cammellos 👍

PR #11593, opened 2020-12-29
[wallet #11535] Show PNs right after enabling

### LINKED ISSUE #11535: Push notification about incoming tx is received only after reopening the app
# Bug Report

## Problem

If User receives an incoming tx while in the foreground, he will not receive a push notification about this until the application is reopened.

#### Expected behavior

Push notification for incoming tx is received while in the foreground without reopening the app.

#### Actual behavior

Reopening is required. 

### Reproduction

Receiver:
- Enable notifications (background service and wallet txs notifications)
- Keep Status in the foreground

Sender:
- Send some ETH or tokens to Receiver from Wallet view

Receiver:
- Put the app to the background
- Reopen the app

### Additional Information

- Status version: Nightly 10/12

cc @rasom

rasom:
fix #11535 

status: ready


Serhy:
Tested with Android 9 and iOS14, - everything works as expected: recovered accounts, enabled notifications and confirmed push notification appears for eth/token transaction sent from Wallet/1-1 chat for transaction receiver (it may take up to 20 mins for PN to appears because in order to pick up the transaction it may take such time. Or pull-to-refresh in wallet account picks up History right away)

flexsurfer (approved):

PR #11598, opened 2020-12-29
hide increase gas/cancel tx, temporary for 1.10

flexsurfer:
JohnLeaToday at 11:59 AM
 I was thinking that some design thought is required regarding the best way to communicate these scenarios to the user.  I wasn't thinking of these scenarios when I did the original design.
if we push this to the next release, we give ourselves the opportunity to work out the best design solution to correctly communicate these scenarios to the user
cammellosToday at 12:02 PM
ok, let s hide and re group for next release then


cammellos (approved):

churik:
thank you for quick PR @flexsurfer 
ready to cherry-pick into 1.10

flexsurfer:
cherry-picked

PR #11604, opened 2020-12-30
Wallet/#11570 non archival node

### LINKED ISSUE #11570: Make sure wallet history and transfers watching works with a custom RPC node
# Bug Report

## Problem

Currently when a custom RPC node is set historical data can't be fetched and that makes wallet's history unusable. Also in case if a new tx is sent we don't watch for confirmation.

#### Expected behavior

[comment]: # (Describe what you expected to happen. Please include designs if possible.)
We should give a user an option to keep using our own infura RPC for historical transfers or (if that's a privacy concern for the user) to use a custom RPC only for recent history and for watching new transfers. 

#### Actual behavior

[comment]: # (Describe what actually happened. Please include screenshots if possible.)
As history can't be fetched (unless the user provides an archival node, which is unlikely) and error is shown and new transactions are not watched properly.

rasom:
fix #11570


src/status_im/ethereum/core.cljs
```diff
@@ -93,6 +93,9 @@
 (defn network->chain-keyword [network]
   (chain-id->chain-keyword (network->chain-id network)))
 
+(defn custom-rpc-node? [{:keys [id]}]
+  (= 45 (count id)))
```
rasom: Probably there is a better way to say that the node is custom. Otherwise magic value should be extracted.

src/status_im/subs.cljs
```diff
@@ -898,8 +905,8 @@
  :<- [:multiaccount]
  (fn [[contacts multiaccount] [_ id identicon]]
    (let [contact (or (get contacts id)
-                     (when (= id (:public-key multiaccount)
```
rasom, line 916: Unrelated to PR but cljfmt did not allow me to create a new commit with this code.

src/status_im/ui/screens/wallet/account/views.cljs
```diff
@@ -214,7 +214,7 @@
 
 (views/defview account []
   (views/letsubs [{:keys [name address] :as account} [:multiaccount/current-account]
-                  fetching-error [:wallet/fetching-error]]
+                  fetching-error          [:wallet/fetching-error]]
```
rasom: To be reverted

src/status_im/ui/screens/wallet/transactions/views.cljs
```diff
@@ -106,15 +106,44 @@
        {:style {:color colors/blue}}
        (i18n/label :t/check-on-etherscan)]]]))
 
+(defn custom-node []
+  [react/view
+   {:style {:flex             1
```
flexsurfer: map values alignment

flexsurfer (approved):

src/status_im/ui/screens/wallet/transactions/views.cljs
rasom: fixed

PR #11615, opened 2021-01-08
Send abridged version of group chats

### LINKED ISSUE #11602: Sluggish performance on private group chats
# Bug Report

User have reported bad performance when chatting on private group chats.

## Problem

This can be due to many factors:
1) Private group chats are an extension of 1-to-1s, so in order to send a message you will have to send n messages, one for each member.
2) Because of redundancy, the history is also transmitted with each message, increasing the cost of POW calculation.

### Potential solutions
1) Is harder to fix, a possible solution is to have a shared topic among members and collate messages there (similarly to server-side fan-out in the same way signal/whatsapp do). It can be further improved by encrypting the message with a single shared key and pairwise encrypt the encryption key.
2) Is easier, we just have to only propagate history when we think it might not have been acknowledged, and we can do that only some times and opportunistically.

cammellos:
Fixes: #11602 
When sending a message in a private group chat we send the whole history
    for redundancy and allow out-of-order processing.
    This can be very expensive in some chats, resulting in long delay when
    sending a message and calculating the POW.
    This commit improves the performance  by only forwarding the events
    necessary for the user to be able to construct a group chat and process
    correctly the message.

### Testing

Existing group chats, creating new group chats.

### Scenario to test

1) Create account A1
2) Create account B1
3) Create account C1
3) Exchange some 1-to-1 messages from A1 to B1, A1 to C1 and B1 to C1
4) Create private group on A1 (admin) including B1 & C1
4) Create account A2
5) Send message in private group chat from A1 

What do you see on A2?


flexsurfer (approved):

cammellos:
This has been merged

PR #11616, opened 2021-01-08
Cache whisper messages

cammellos:
This commit re-introduces a feature that we lost during the migration to status-go.
Messages are cached for a couple of days if processed correctly by status-go, to avoid performance issues.
It's at-leat-once processing.


flexsurfer (approved):

cammellos:
This has been merged

Issue #11197, opened 2020-09-21
IPhone is heating up quickly when sending messages in 1-1 chat


churik:
# Bug Report

## Problem
Generally, users noticed that iPhone currently heating up faster than on release.
To find out the reason, made the following test cases:

1. receiving 50 messages in public chat while IPhone is in foreground
2. sending 50 text messages to public chat
3. sending text messages or mentions to 1-1 chat
4. sending mentions in public chat
5. receiving 20 push notifications in 1-1 chat while app is on background
6. receiving 20 push notifications in public chat while the app is on background

In cases 4 - 6 there are no noticeable changes in device temperature, cases 1 - 3 are in table below

**IPhone is heating up if send messages to 1-1 chat with mentions and after stopping sending continue to heat up more (from +0,2 to +0,4)**

Didn't notice other actions that can heat up Iphone so quickly.

  | 1.6.1 (Iphone 7, IOS 13) | 1.6.1 (IPhone SE, IOS 14) | 21/09/20 (IPhone7, IOS 13) | 21/09/20(IPhone SE, IOS 14)
-- | -- | -- | -- | --
receiving 50 messages (foreground on 1.6.1, background +PNs on 21/09/20) | -0.5 | 0.5 | 0.4 | 0.2
sending 50 ordinary messages to public chat | 0.5 | 1.3 | 0.8 | 0.9
sending 20 messages to 1-1 chat (on 21/09/20 with mentions) | 0.2 | 0.6 | **1.1** | **1.3**

Let me know if more cases should be checked.
cc @cammellos

cammellos:
Ok, 
The issue seems to be in sending mentions.
What we added related to this in 1.7 is:

1) Sending PNs
2) Parsing mentions in status-go
3) Sending PNs for mentions in public chats
4) Parsing mentions in status-react (correct me if I am wrong @rasom )
5) Autocomplete for mentions in status-react

If the issue is only in public chats, it does seems to be related to sending mentions in public chats, as at least an extra message need to be sent for each mentioned user.
I'll investigate further.

churik:
**Profile > Advanced > Send PNs is ON:**
 [status_logs.zip](https://github.com/status-im/status-react/files/5255878/status_logs.zip) - last 3 mins sending 20 messages, phone heat up ~1C


**Profile > Advanced > Send PNs is OFF:** 
[status_logs.zip](https://github.com/status-im/status-react/files/5255977/status_logs.zip) - last 4
[...]

PR #11619, opened 2021-01-11
[Fixes: #11197] Fix busy loop and don't resend group chat messages

cammellos:
There was an issue with the PN system that would cause a busy loop,
likely impacting performance.

Another thing that this commit addresses is that since we fixed retry of
sending messages, we noticed that it's to eager on group chats, so for
now it's disabled for them (essentially same behavior as 1.9), while
still working for 1-to-1.


flexsurfer (approved):

churik:
@cammellos this on is included to 1.10 already?

churik:
@cammellos you are rockstar 🥇 
checked all cases defined in https://github.com/status-im/status-react/issues/11197
And in all cases Iphone 7 (which can be considered a slow device) is not heating at all (0 - 0,1 degree).

cammellos:
Closing this as it has already been included in a different PR

PR #11641, opened 2021-01-14
handle error from infura

flexsurfer:
when adding new contact with not existing ens name it shows loading indicator

at some point infura started to return execution reverted error instead of "0x" result


shivekkhurana (approved):

Serhy:
Looks good to me, Android 8, Android9, iOS14 - when searching for ENS (in 1-1 chat, wallet) there is no more endless spinner on non-existing ENS name (and positive response works in that places too).

PR #11642, opened 2021-01-14
rn config

flexsurfer:
rn config


jakubgs (approved):

churik:
Tested on IOS 13 and IOS 14, works for me

PR #11650, opened 2021-01-15
[#11643] Display profile picture only of users I have added to contacts

### LINKED ISSUE #11643: Display profile picture only of users I have added to contacts
As a user
I want to see only the profile pictures of users I have added to my contacts
So that I won't be subject to inappropriate pictures in public chats

### Description

Because of the demand for moderation in app stores to some apps, we decided to limit our exposure and protect users from inappropriate content.
Users should only be seeing profile pictures of other users if they added them as contacts.

### Designs

https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=4717%3A2107

### Implementation

This will require a change in status-go to be feature-complete.
If we want to cut the scope to a minimum we can implement the design but grey out the other two options (`Everyone`, `None`), leaving `Contacts` as the only option, which does not require any status-go changes.
Up to you, if de-scoped, please create a separate issue with the remaining work.

flexsurfer:
fixes #11643
| | |
| -- | -- |
| image | image |


cammellos (approved):

churik:
@flexsurfer thank you!
remaining issues:
### ISSUE 1: can't see own profile picture in group info 

IMAGE 2021-01-15 16:45:03
regardess of "Show profile pictures of" in settings

### ISSUE 2: after updating profile picture it is not updated for users who doesn't add him as contact

**Steps:**
1) User A: set profile picture 1, post something to public chat 
2) User A: set profile picture 2
3) User B: join public chat, set "Show profile pictures of" to  "Everyone", check public chat

**Expected result:**
picture 2 for User A
**Actual result:**
picture 1 for User A
**OS:** IOS, Android

May be it is acceptable for now, let me know @flexsurfer @cammellos

cammellos:
Issue 2 does not seem to be related to this changes, thought I can't be 100% sure, can you replicate consistently?

churik:
@cammellos yes, I can. And you are right - it is not related, as it is reproducible on nightly, so will report it separately (if it makes sense in general)
so only issue 1 is remaining

cammellos:
what abou 1.10, same issue?

churik:
@cammellos yes, profile picture will remain old until:
1) user with updated image will post smth in public chat
2) or you add him to contacts
(and maybe there are some other ways to get it updated)

cammellos:
@churik that's the expected behavior actually, we won't update pictures in public chat until you post, for privacy and bw reasons
it's a bit unexpected, but it's how we thought about it :)

flexsurfer:
fixed group chat, although it is not introduced in this PR

churik:
Ready to be merged in 1.10, thank you!

PR #11653, opened 2021-01-18
[#11117] Move PNs to status-go

### LINKED ISSUE #11117: Move notifications on Android to status-go
### Description
Notifications in java, as they are disconnected from the rest of the app, we should move them to status-go, using Java only to display them.

This issue should be resolved before addressing other issues in the epic #10737

### User Story
As a developer, I want to maintain logic of local notifications on Android in status-go so that I don't have to duplicate any code/data.

### Description

Currently android notifications are in the Java layer, they parse `messages.new` and notify on the content.
The issue with this approach is that they don't have access to a lot of information, ENS names, duplicated messages etc.

The building of notifications should be in status-go, so that the Java layer simply displays what is instructed by status-go, instead of having to replicate the logic in Java.

Local push notifications (in-app notifications) are dependent on this issue

rasom:
fixes #11117

rules:
- no PNs for a specific chat when that chat is active (the app is opened and the chat is shown)
- show PNs for all messages in 1-1 chats and private groups
- show PNs when the user is mentioned in public chats and communities
- show PNs when the user receives a reply to their message in public chats and communities **(WILL BE IMPLEMENTED SEPARATELY)**

status: ready


modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/NewMessageSignalHandler.java
```diff
@@ -122,9 +125,15 @@ public Intent getOpenAppIntent() {
     //NOTE: this method takes a chatId and returns an intent that will open the app in that chat
     //Once we support other kind of notifications we will need to adapt it. The simplest method
     //is probably to pass the universal link as param instead of the chatId.
-    public Intent getOpenAppIntent(String chatId) {
+    public Intent getOpenAppIntent(String chatId, int chatType) {
         Intent intent = getOpenAppIntent();
-        intent.setData(Uri.parse("status-im://p/" + chatId));
+        String path = "";
+        if (chatType == 1) {
```
cammellos: Probably best to put this in some constant

modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/NewMessageSignalHandler.java
```diff
@@ -122,9 +125,15 @@ public Intent getOpenAppIntent() {
...
+            path = "p/";
+        } else if (chatType == 3) {
```
cammellos: Same here

src/status_im/notifications/local.cljs
```diff
@@ -95,6 +103,66 @@
      :user-info notification
      :message   description}))
 
+(defn get-chat-type [chat]
```
cammellos: Mmh, do we need to convert to `keyword`? 
I'd use constants for everything, as our definition of `:public?/:group-chat/:one-to-one` falls short to capture all the different kinds of chats and makes refactoring hard otherwise.
You have all you need in `constants.cljs` namespace, and I think we should probably always use those

rasom: this part is unnecessary, I will just remove it

rasom: fixed

modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/NewMessageSignalHandler.java
rasom: fixed

modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/NewMessageSignalHandler.java
rasom: fixed

rasom:
@cammellos iOS crash is fixed and changes according to your comments are done

cammellos (approved):

qoqobolo:
### ISSUE 1: Can't receive PNs if the app was logged out while in the background + Can't open the app via pressing PN after such a logout (sometimes can open but the app crashes) 

If User A enables PNs and sends the app to the background for a while until auto-logout, and User B sends some messages after User A was logged out while in the background, User A won't receive PNs for messages, which were sent after logout.

If User A opens the app he will receive these messages and PNs about them.

If User A received some messages _before_ auto-logout while in the background, and some messages _after_ auto-logout, then _after_ auto-logout he won't be able to open the app via pressing PNs about that messages. 

#### Expected behavior

- No auto-logout in the background if PNs are enabled
- Opening the app by tapping on PN 

#### Actual behavior

- Logout in the background despite enabled PNs
- Can't open the app by tapping on PN (sometimes can open but the app crashes) 

**_Steps:_**

User A:
1. Enable PNs
2. Send the app to the background

User B:
1. Send a message to 1-1 chat with User A (User A receives PNs)
2. Wait ~3 minutes (until User A is logged out) 
3. Send another message to 1-1 with User A

User A:
1. Tap PNs

@rasom wdyt about this case? Does it make sense to prevent the User from logging out while in the background if PNs are enabled?

qoqobolo:
### ISSUE 2: No PNs for `Send transaction` or `Request transaction` in 1-1 chat

If User A sends address request (`Send transaction` command) or requests transaction in 1-1 chat with User B, User B doesn't receive a push notification about it.

####  Expected behavior

User receives PNs for commands `Send transaction` and `Request transaction` in 1-1

####  Actual behavior

https://user-images.githubusercontent.com/67952253/105994697-45fe5480-60a8-11eb-9a74-ce7395074929.mp4

_**Steps:**_

User B:
- Enable PNs
- Send the app to the background

User A:
- Open 1-1 chat with User B
- Open `Request transaction` dialog
- Enter some ETH and tap `Request`

User B:
- Check PNs

qoqobolo:
### ISSUE 3: App crashes when the user receives a large image in 1-1 chat in the background with PNs enabled

If User A enables PNs and sends the app to the background, and User B sends a large (3-4MB) image to User A, there will be two cases:
1. User A will receive PN, tap it and the app will crash after opening
2. User A won't receive PN, and the app will crash while in the background. Error `Status PR keeps stopping` will be shown.
Screenshot 2021-01-28 at 18 45 29

If User B sends a small image (~100KB), User A will receive PN and the app won't crash.

**_Steps:_**
User A:
- Enable PNs
- Send the app to the background

User B:
- Start 1-1 chat with User A
- Send a photo (not a screenshot) ~4MB

User A:
- Check device
- Tap PN if received 
- If chat is opened, navigate to the home screen or another tab

[crashlogs.log](https://github.com/status-im/status-react/files/5888122/crashlogs.log)

OS: Android

rasom:
issue 3 cause:

```
02-01 13:49:38.809 32045 32045 F DEBUG   : Abort message: 'java_vm_ext.cc:542] JNI DETECTED ERROR IN APPLICATION: JNI CallStaticVoidMethod called with pending exception java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 2288208 bytes'
...
```

rasom:
@qoqobolo issue1 and issue 3 should be fixed now, issue 2 will be handled in a separate pull request by @Samyoul https://github.com/status-im/status-go/pull/2130#issuecomment-771625263

qoqobolo:
@rasom thanks for the PR and fixes!

Tested on Android 14 and iOS 11:
- PNs for text, images, stickers, audio messages in 1-1 and group chats
- PNs for mentions in public chats
- foreground, background, blocked screen
- PNs when the app is closed, after logout
- PNs on the mobile network, after returning from offline 
- redirect to 1-1 on Android
- muted user on iOS
- PNs from blocked user
- PNs were enabled during onboarding/in settings
- wallet notifications
- PNs setting on iOS (from non-contacts, PN for mention and combinations)

churik:
Ready to merge, thank you

PR #11669, opened 2021-01-20
[#8284] Remove unnecessary permissions [Android]

### LINKED ISSUE #8284: Remove unnecessary permissions [Android]
# Problem

From Trail of Bits:

> There is a configuration for a debug keystore, but not for a production keystore. Ensure that you aren't running the application in a debug configuration at all times. 

> Similar to the iOS version of the application, excessive permissions are being set for the application. Does it actually require NFC access? There is even a line in the android manifest saying: 

 

    

    <uses-permission android:name="android.permission.READ_PHONE_STATE" tools:node="remove"/>

## Notes

As with[ iOS](https://github.com/status-im/status-react/issues/8282), we need to go through the Android permissions and remove anything unnecessary.

flexsurfer:
fixes #8284

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -25,10 +25,6 @@
 
     <uses-permission android:name="com.android.vending.BILLING" />
 
-    <!-- these permissions should be removed -->
-    <!-- React Native unnecessary permissions (https://github.com/facebook/react-native/issues/5886) -->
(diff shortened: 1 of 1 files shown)


shivekkhurana (approved):

Serhy:
Looks good on Android 9, Android 8 and iOS, - permissions kept on camera, storage, camera in browser as in develop. 
False failing e2e will fix separately it's not PR issue.

PR #11687, opened 2021-01-27
Avoid concurrent access to contact

cammellos:
In some instances when fetching messages for a timeline the app would
crash as a map was accessed concurrently without locking.
This fixes the issue.

status: ready


flexsurfer (approved):

rasom (approved):

Serhy:
Tested on Android 8,9,10 and iOS14, - app behaves fine (not crashes) when I fetch message history from public/1-1/group chats.

Issue #11271, opened 2020-10-05
"Won't fix" issues (to be updated)


qoqobolo:
* [ ] **ISSUE 1 - If shift the focus to the beginning of the amount on the Send transaction view and try to correct it (e.g. delete first char), new chars will be shifted off the screen.**

**Operating System:** iOS 

![ezgif com-gif-maker (4)](https://user-images.githubusercontent.com/67952253/94944068-ff716280-04d8-11eb-888c-2df07666f855.gif)

- - -
* [ ] **ISSUE 2: Animation of the "Copied" message is broken when copy the invite link to the group chat with a long press.**

_Steps:_

* Create group chat
* Open chat settings > Group info
* Tap Group invite
* Long tap Grop invite link

**Operating System:** Android, iOS 

![ezgif com-gif-maker (5)](https://user-images.githubusercontent.com/67952253/94945355-d6ea6800-04da-11eb-84ed-839c08708f54.gif)

- - -
* [ ] **ISSUE 3: If use the gesture to swipe the screen to go back during onboarding, then when the user goes to the next screens, those screens from which the User went back will be skipped.**

E.g. if the user swipes back from the password generation screen to the choosing 3-random name screen and then tap "Next", password screen will be skipped, User won't be able to enter and save it. So at the end, the User could be inside the new account, without a selected password, but it’s required anyway.

_Steps:_

* Instal Status
* Tap Get started
* Tap Generate keys
* Swipe back via the gesture 
* Tap Generate keys
* Swipe back
* Tap Generate keys

**Operating System:** iOS 

![ezgif com-gif-maker (8)](https://user-images.githubusercontent.com/67952253/95012275-a53ae380-0637-11eb-9303-0480d111a1f5.gif)

- - -
* [ ] **ISSUE 4: If a token was added only to the main account, on the Watch-only account it will be shown as "...WEENUS / ...USD" until relogin.**

**Operating System:** Android

![Screenshot 2020-10-08 at 17 18 51](https://user-images.githubusercontent.com/67952253/95478838-591ad680-098a-11eb-8802-9743d25efe77.png)

- - -
* [ ] **ISSUE 5: If block some user in public chat, and then unblock him, his name will be shown in the mention suggestions list in 1-1 and group chats with other users.**

#### Expected behavior
Mention suggestions list contains contacts and doesn't contain users from other chats, blocked, and unblocked users.

_Steps:_

* Join public chat
* Block some user and then unblock him
* Start 1-1 chat with another user
[...]

PR #11701, opened 2021-02-01
Don't rely on contacst for one-to-one mentions, and remove cursor

### LINKED ISSUE #11670: Mention disappears from the input field in chat view on Android
# Bug Report

## Problem

When typing, the mention is not displayed in the input field.
Can be sent but in the input field looks like nothing has been entered.
The issue is not reproducible at the first attempt, i.e need to re-open the chat view.

#### Expected behavior

Mention is visible in the input field

#### Actual behavior

https://user-images.githubusercontent.com/67952253/105202897-c0782300-5b42-11eb-98bd-c80e7b35b2cb.mp4

### Reproduction

- Join public chat
- Re-open the chat
- Type `@`
- Select some user name from the suggestion list
- Send message

### Additional Information

- Status version: Nightly 20/01
- Operating System: Android

### LINKED ISSUE #11671: No name of the second user in the list of suggestions for mentions in 1-1 chat
# Bug Report

## Problem

If start 1-1 chat with User A and type `@`, then his name will not appear in the list of mention suggestions, only mine.

#### Expected behavior

Both my name and the name of the second user are in the suggestions list

#### Actual behavior

Screenshot 2021-01-20 at 20 30 14

### Reproduction

- Start 1-1 chat with User A
- Type `@`

### Additional Information

- Status version: Nightly 20/01

cammellos:
[Fixes: #11670]
[Fixes: #11671]

This commit fixes 2 issues.

One issue was due to removing tribute to talk, which added in the
`contacts` map any user you open a chat with, and some code relied on
this behavior.

The second issue was likely due to a react native upgrade which might
have fixed a bug but caused a different issue, which resulted in the
text disappearing when cursor was set.

I have removed that (it was only used on android) and everything seems
to be working correctly, but not so sure about this one.

### Testing

Mentions on android/ios in one-to-one chats/group chats etc

cc @rasom I could not replicate the bug on android with cursor, but maybe I am just not looking in the right place.
I also see the text flickering, not sure if it's develop that is slow though.

### DIFF
--- a/src/status_im/contact/db.cljs
+++ b/src/status_im/contact/db.cljs
@@ -18,8 +18,8 @@
 (defn public-key->contact
   [contacts public-key]
   (when public-key
-    (get contacts public-key
-         (public-key->new-contact public-key))))
+    (or (get contacts public-key)
+        (public-key->new-contact public-key))))
 
 (defn- contact-by-address [[addr contact] address]
   (when (ethereum/address= addr address)
--- a/src/status_im/chat/models/mentions.cljs
+++ b/src/status_im/chat/models/mentions.cljs
@@ -195,6 +195,10 @@
                        :name       name}))
              {}
              group-contacts))
+          (= chat-type constants/one-to-one-chat-type)
+          (assoc users
+                 current-chat-id
+                 (contact.db/public-key->contact contacts current-chat-id))
 
           :else users)
         {:keys [name preferred-name public-key]}
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -121,7 +121,6 @@
 (reg-root-key-sub :camera-roll-photos :camera-roll-photos)
 (reg-root-key-sub :group-chat/invitations :group-chat/invitations)
 (reg-root-key-sub :chats/mention-suggestions :chats/mention-suggestions)
-(reg-root-key-sub :chats/cursor :chats/cursor)
 (reg-root-key-sub :chat/inputs-with-mentions :chat/inputs-with-mentions)
 (reg-root-key-sub :inactive-chat-id :inactive-chat-id)
 ;;browser
@@ -1065,15 +1064,20 @@
  :<- [:chats/mentionable-contacts]
  :<- [:contacts/blocked-set]
  :<- [:multiaccount]
- (fn [[{:keys [users]} contacts blocked {:keys [name preferred-name public-key]}]]
-   (apply dissoc
-          (-> users
-              (merge contacts)
-              (assoc public-key (mentions/add-searchable-phrases
-                                 {:alias      name
-                                  :name       (or preferred-name name)
-                                  :public-key public-key})))
-          blocked)))
+ (fn [[{:keys [chat-id users chat-type]} contacts blocked {:keys [name preferred-name public-key]}]]
+   (let [contacts-with-one-to-one (if (= chat-type constants/one-to-one-chat-type)
+                                    (assoc contacts chat-id (get contacts chat-id (-> chat-id
+                                                                                      contact.db/public-key->new-contact
+                                                                                      contact.db/enrich-contact)))
+                                    contacts)]
+     (apply dissoc
+            (-> users
+                (merge contacts-with-one-to-one)
+                (assoc public-key (mentions/add-searchable-phrases
+                                   {:alias      name
+                                    :name       (or preferred-name name)
+                                    :public-key public-key})))
+            blocked))))
 
 (re-frame/reg-sub
  :chat/mention-suggestions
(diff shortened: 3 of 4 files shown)


src/status_im/contact/db.cljs
```diff
@@ -18,8 +18,8 @@
 (defn public-key->contact
   [contacts public-key]
   (when public-key
-    (get contacts public-key
-         (public-key->new-contact public-key))))
+    (or (get contacts public-key)
```
flexsurfer, line 21: just wondering how its different, why `not found` didn't work ?

flexsurfer (approved):

src/status_im/contact/db.cljs
cammellos: oh no, that's fine it worked before.

I changed this for performance reasons, as if you use:

`(get x y (some-func z))` 

`(some-func z)` is always executed (in our case is expensive as it makes a call to the bridge sync.

while if you use

`(or x (some-func z))` 
`(some-func z)` is only executed if `x` is falsy (i.e it short circuit)

Also sometime it's a bit unexpected behavior `(get ..)`, as `(get {:a nil} :a "default")` returns actually `nil`, which most of the times is not what you want (it does mean though that someone didn't clean up the map correctly)

qoqobolo:
@cammellos  thank you for PR!

During testing one more issue with mentions was found in Nightly: 
**Self-mention with ENS with `.stateofus.eth` domain is resolved incorrectly or not resolved at all in the sent message.**

_Reproduction:_ 

- Set ENS name with `.stateofus.eth` domain in Settings
- Join public chat
- Send self-mention

Sometimes the issue isn't reproducible on the first try.

Should I create a separate issue or maybe you could fix it in the scope of this PR? Or maybe it's a `won't fix` bug, wdyt?
 
Screenshot 2021-02-02 at 12 23 45

Screenshot 2021-02-02 at 13 28 14

cammellos:
@qoqobolo thanks for testing,
regarding the issue probably I would address separately, maybe best to create a separate issue. 
It might be a wontfix issue, but maybe best to just have a quick look at it, as it might be a simple fix.

What do you think?

qoqobolo:
@cammellos agree. Will create a new issue.
Testing is still in progress. Thanks for waiting!

qoqobolo:
### ISSUE 1: If place the cursor in the mention after `@` and delete it, then the whole mention will disappear from the input field on Android
But if press the `Send` button, the message will be sent in plain text.

_**Steps:**_
- Open chat
- Type `@` 
- Select some mention from the suggestion list
- Place the cursor after `@` in the mention
- Delete `@`
- Tap `Send`

https://user-images.githubusercontent.com/67952253/106610716-60807400-6567-11eb-8dc1-e66b4832b0be.mp4

qoqobolo:
### ISSUE 1: The mention disappears from the input field if start to delete it (Android)

**_Steps:_**
- Join public chat
- Type `@` and select some user from the list
- Tap `Delete` on the keyboard

https://user-images.githubusercontent.com/67952253/106612285-3af46a00-6569-11eb-8afd-927eb872287a.mp4

cammellos:
@qoqobolo it should be fixed in the latest build, @rasom helped me out, thanks both!

qoqobolo:
@cammellos the issue is still reproducible :( But with a new condition: the mention disappears if there is a text before it in the input field, and you start to delete the mention.

https://user-images.githubusercontent.com/67952253/106646159-a6e8c980-658d-11eb-8197-81e085950c34.mp4

cammellos:
Thanks @qoqobolo , I can replicate but a bit sporadically, is it consistent for you?
tomorrow I will take a look at it.

rasom:
@qoqobolo @cammellos should be fixed now

qoqobolo:
@cammellos @rasom thank you, works fine now!
Last recheck on iOS is needed, so have to wait for the ios issue to be resolved.

@cammellos regarding the problem with the self-mention, which I mentioned above, it seems to make sense to consider it as a 'won't fix' bug. Because now it is reproduced sporadically and we are still figuring out the exact steps to reproduce. So I'll log it to #11271 until steps are found.

cammellos:
@qoqobolo IOS build is ready to be tested, thanks!

qoqobolo:
Tested on Android 11 and iOS 14:
- mentions in public, 1-1, group chats
- mention with a nickname (different languages and emoji)
- ENS names
- long messages with mentions
- self-mention
- dark mode
- removing mentions
- mention in reply
- mention of a blocked user
- mention in PN
- list of mention suggestions

One minor issue with nicknames in mentions in Korean, Arabic, and Chinese was found on iOS, but it'll be logged as `won't fix`.

Otherwise, PR looks great and is ready to be merged.

PR #11702, opened 2021-02-02
Webview camera permission

Serhy:
New e2e to check whether camera  works in Dapp (and access permissions)


test/appium/tests/atomic/dapps_and_browsing/test_dapps.py
```diff
@@ -36,6 +36,60 @@ def test_request_public_key_status_test_daap(self):
         if not status_test_dapp.element_by_text(user['public_key']).is_element_displayed():
             self.driver.fail('Public key is not returned')
 
+
+    @marks.testrail_id(6635)
+    @marks.medium
+    def test_webview_camera_permission(self):
+        web_view_camera_url = 'https://simpledapp.status.im/webviewtest/webviewcamera.html'
+        sign_in_view = SignInView(self.driver)
+        home_view = sign_in_view.create_user()
```
churik: as sign_in_view is not used in test later, can be used directly
`home_view =SignInView(self.driver).create_user()`

also, better to avoid _view in the end to keep code style the same (sometimes it makes calling methods in test very long, so decided to shorten it)

Serhy: Thanks, fixed.

churik (approved):

Issue #11681, opened 2021-01-26
[wallet] Reduce the number of RPC request by inactive users


rasom:
# [wallet] Reduce the number of RPC request by inactive users

Significant part of our users is more insterested in chat functionality than in wallet, and that means we don't really need to watch their addresses for incoming/outgoing transfers unless we spotted that those users have some activity related to the wallet.

## How it works currently:
- We check transfers history every 20 mins 

## How it could work:
- When account doesn't have transactions in history we do not check recent transfers unless there is some activity which might lead to new transfers (e.g. user sent transfer request, user shared address, user pulled to refresh tx history). Even if that happened we only schedule check in N minutes (20 currently) and then if we there are no new transactions we stop waching.
- If the last transaction happened 5-7 day ago we can switch to 1h interval instead of 20 minutres. If there was no transfers during past 32 days we can stop watching recent transfers at all.

That would save us shitload of requests but it is hard to estimate how many exactly :)

PR #11723, opened 2021-02-08
[wallet] RPC calls optimizations for new accounts

rasom:
fixes (partially) #11681

When a new account is created only one RPC request related to wallet is executed to get latest block number. (We had 32 requests before this PR). Also checking of new txs doesn't happen until a new user opens wallet tab.

changes:
- when a new account is created, scanning doesn't happen, we just mark `zero block - latest block` range as scanned
- we also don't start scanning unless wallet was opened
-  in case if wallet's history is empty, we only check history on login and when the user pull to refresh wallet. If a custom RPC node is used history is still checked every 4 minutes

status: ready


flexsurfer (approved):

churik:
@rasom 
when restoring multiaccount with funds on cellular network, assets are not shown until manually scan tokens even if you come back to Wi-Fi.
IMAGE 2021-03-01 10:54:07

rasom:
@churik and that's a new thing for PR, doesn't happen in develop, right?

churik:
@rasom right

rasom:
@churik I will check how this PR affects it, but I often have similar issues in release version

churik:
@rasom on your screen ETH and SNT are shown; other ... assets I suppose are from another account.
On my screen the whole balance is not shown at all.

rasom:
@churik 
One of optimisations here is that we do not request balances for all 160 tokens we have (that requires 160 `eth_call` requests) in case if account has an empty tx history (thus no need to check those balances). The problem here is that on cellular we do not scan history on account creation, and that's why balances were not checked.

I changed condition a bit and now we will execute 160 `etc_call` requests on cellular. This solves empty balances issue but might be heavy for those who have expensive/slow plan. As we don't ask user to make choice for syncing/not syncing on mobile I don't see another way to solve it at the moment.

churik:
Tested:
- creating new mutiaccount
- fetching transaction history (offline, cellular, using pull-to-refresh and "Load")
- fetching token balances for restored multiaccounts  (offline -> online, cellular)
- new transactions in tokens and ETH (ropsten,  on Wi-Fi and cellular network)

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)

PR #11725, opened 2021-02-09
Handle eth_accounts exception

cammellos:
handle eth_accounts exception

### DIFF
--- a/resources/js/provider.js
+++ b/resources/js/provider.js
@@ -84,6 +84,7 @@
                 } else if (data.isAllowed) {
                     if (data.permission == 'web3') {
                         window.statusAppcurrentAccountAddress = data.data[0];
+                        window.ethereum.emit("accountsChanged", data.data);
                     }
                     callback.resolve(data.data);
                 } else {
@@ -152,6 +153,30 @@
     EthereumProvider.prototype.status = new StatusAPI();
     EthereumProvider.prototype.isConnected = function () { return true; };
 
+    EthereumProvider.prototype._events = {};
+
+    EthereumProvider.prototype.on = function(name, listener) {
+        if (!this._events[name]) {
+          this._events[name] = [];
+        }
+        this._events[name].push(listener);
+    }
+
+    EthereumProvider.prototype.removeListener = function (name, listenerToRemove) {
+        if (!this._events[name]) {
+          return
+        }
+
+        const filterListeners = (listener) => listener !== listenerToRemove;
+        this._events[name] = this._events[name].filter(filterListeners);
+    }
+
+    EthereumProvider.prototype.emit = function (name, data) {
+        if (!this._events[name]) {
+          return
+        }
+        this._events[name].forEach(cb => cb(data));
+    }
     EthereumProvider.prototype.enable = function () {
         return sendAPIrequest('web3');
     };
@@ -159,11 +184,6 @@
     EthereumProvider.prototype.scanQRCode = function (regex) {
         return sendAPIrequest('qr-code', {regex: regex});
     };
-
-    EthereumProvider.prototype.on = function (type, handler)
-    {
-        console.log("Not supported by Status")
-    }
 
     EthereumProvider.prototype.request = function (requestArguments)
     {
--- a/src/status_im/browser/core.cljs
+++ b/src/status_im/browser/core.cljs
@@ -403,18 +403,35 @@
                                                   :error     %1
                                                   :result    %2}])]}))))
 
+(fx/defn handle-no-permissions [cofx {:keys [method id]} message-id]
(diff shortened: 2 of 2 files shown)


flexsurfer (approved):

flexsurfer:
should fix such cases as
image
image

src/status_im/browser/core.cljs
```diff
@@ -403,18 +403,34 @@
...
+                     :messageId message-id
+                     :result    {:jsonrpc "2.0"
+                                 :id      (int id)
+                                 :result  []}})
+    (send-to-bridge cofx
+                    {:type      constants/web3-send-async-callback
+                     :messageId message-id
+                     :error     {:code 4100}})))
+
+(def require-permission?
```
shivekkhurana: It is not conventional to add `?` suffix to a non-boolean value. `require-permission-enum` or `require-permission-set` is a more apt name.

shivekkhurana (changes requested):

cammellos:
I ll cast it to a boolean

cammellos:
@shivekkhurana @flexsurfer ready to be reviewed again.
I have added `.on` method, I pulled the implementation from desktop (I would instinctively used `addEventListener` myself, but I trust their code is more battle tested).

src/status_im/browser/core.cljs
```diff
@@ -403,18 +403,39 @@
...
+    "keycard_signTypedData"
+    "eth_signTypedData" "personal_sign" "personal_ecRecover"})
+
+(def require-permission?
+  (comp boolean permissioned-methods))
+
+(defn has-permissions? [{:dapps/keys [permissions]} dapp-name method]
+  (boolean
+   (and (require-permission? method)
+        (not (some #{constants/dapp-permission-web3} (get-in permissions [dapp-name :permissions]))))))
```
shivekkhurana: This can simply be:

```clojure
(and (permissioned-methods method)
     (not (some ...)))
```

You don't need to typecast manually. Is there any specific reason to add an extra `def require-permission?`

resources/js/provider.js
```diff
@@ -77,13 +77,20 @@
         var id = data.messageId;
         var callback = callbacks[id];
 
+        if(data.type === "web3-disconnect-account") {
```
flexsurfer: im not sure we need this?

resources/js/provider.js
```diff
@@ -152,6 +159,25 @@
...
+    }
+
+    EthereumProvider.prototype.removeListener = function (name, listenerToRemove) {
+        if (!this._events[name]) throw new Error(`event "${name}" does not exist`);
+        const filterListeners = (listener) => listener !== listenerToRemove;
+        this._events[name] = this._events[name].filter(filterListeners);
+    }
+
+    EthereumProvider.prototype.emit = function (name, data) {
+        if (!this._events[name]) throw new Error(`event "${name}" does not exist`);
```
flexsurfer: hm not sure we need to throw event here, what if dapp is not subscribed for accountsChanged?

resources/js/provider.js
```diff
@@ -152,6 +159,25 @@
...
+
+    EthereumProvider.prototype.on = function(name, listener) {
+        if (!this._events[name]) {
+          this._events[name] = [];
+        }
+        this._events[name].push(listener);
+    }
+
+    EthereumProvider.prototype.removeListener = function (name, listenerToRemove) {
+        if (!this._events[name]) throw new Error(`event "${name}" does not exist`);
```
flexsurfer: why do we want to throw an error here?

resources/js/provider.js
cammellos: actually you are right

resources/js/provider.js
cammellos: If you are not sure, I'll remove it :)

cammellos:
@flexsurfer addressed feedback

flexsurfer (approved):

shivekkhurana (approved):

churik:
@cammellos 
e2e failed because buttons in simpledapp.eth that should trigger transaction screen don't do anything.
Release - PR comparison:

https://user-images.githubusercontent.com/4557972/107383832-ab633400-6af1-11eb-8420-6bf0593a9b4d.mp4

Video/logs: https://saucelabs.com/jobs/e18cfc73f0474dd2939fadfad0d9349f?auth=615094ee6189f9b0883bc4f4311cf40e

churik:
Tested on:
- curve.fi
- yearn.finance
- opensea
- cryptokitties
- matcha
on IOS / Android (keycard)

Works way better than before.
thank you for finding and fixing this!

PR #11730, opened 2021-02-09
Only scale if width/height is positive

cammellos:
image width/height is returned as 0,0 from status-go, we should likely `omitempty` there, but also we should make sure they are both positive.


shivekkhurana (approved):
> yes, basically status-go ints are non-nullable so they get serialized as 0

PR #11754, opened 2021-02-16
Chats improvements

flexsurfer:
- Currently, we offload messages, so each time we open the profile we "reset" chat from which profile was opened (because we have status updates in profile), same for a timeline when we switch between timeline and chat, we don't offload messages in this PR for profiles and timeline
- we shouldn't use nested scrolls, so I've replaced the header component in the profile screen 
- Improved signal handler for messages, process 50 messages instead of 1 (it could be higher, but 50 should be enough because we show ~20 messages )

_Edited_: ok, reduced to 10 , because on slow devices it freezes 
_Edited_: actually, this is still bad on slow devices, why? 

js->clj for 20 messages is 25ms, 
preparing (adding timestamp) one message is 2-6 ms ( so if we receive 100 it will be 600ms)
rendering of 10 messages is 500ms
you can see the numbers on your phone in this build : 80c2468 | #27
just open any chat and press "..." and then "show perf log", times in ms
"enable perf" replaces item renderer on simplest one with text, for me, it reduces time from 500ms to 10-25ms

**further work in next PRs** so we need to improve our item rendered and reduce js->clj and preparing times

What's done in this PR:

1) separate subscriptions for profile and timeline streams, so we don't reset both chat and profile/timeline when we switch between them
2) sanitize messages on a signal, filter messages for a current chat, and sort them so we can process starting from the latest messages, also do js->clj as later as possible
3) process 10 messages (3 on low androids ) instead of 1
4) improve item renderer a little bit (it requires serious work to improve it significantly)
5) do not use an animated header in the profile because of nested scroll views
6) add loading indicator for messages

Why message offloading was disabled in this PR
1) it conflicts with endreached event, we offload messages and then endreached event fired and we load 20 messages again. This happens also when sending message, send message, then offload, readd all messages, then endreache, add 20 more, receiving each new messages or sending each new message cause this chain of events
2) re-adding all messages is really expensive , readding one messages is ±6ms so, if we offload 5 messages in the end of the list and we have 300 messages we just re-add 300 messages , ±1800 ms
Probably we still need to offload messages in one case, when we are in the chat for a long time and receive lots of new messages, we need to find a way how to pop messages from the arrray and do not re-add all messages in that case, but because messages will be offloaded when user switch a chat or just go back to chats list, i guess we can leave with it, and implement properly in another PR, test shows that adding messages time grows not so fast with having more messages, the only place where it grows is subscription, for 300 messages it might take up to 100-200ms on low end android device

https://github.com/status-im/status-react/issues/11846

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -1,214 +1,188 @@
 (ns status-im.chat.models.message
-  (:require [re-frame.core :as re-frame]
-            [status-im.chat.models :as chat-model]
-            [status-im.chat.models.loading :as chat-loading]
+  (:require [status-im.chat.models :as chat-model]
             [status-im.chat.models.message-list :as message-list]
             [status-im.constants :as constants]
             [status-im.data-store.messages :as data-store.messages]
             [status-im.ethereum.json-rpc :as json-rpc]
-            [status-im.multiaccounts.model :as multiaccounts.model]
             [status-im.transport.message.protocol :as protocol]
-            [status-im.ui.screens.chat.state :as view.state]
             [status-im.utils.fx :as fx]
             [taoensso.timbre :as log]
             [status-im.chat.models.mentions :as mentions]
-            [clojure.string :as string]))
+            [clojure.string :as string]
+            [status-im.contact.db :as contact.db]
+            [status-im.utils.types :as types]
+            [status-im.ui.screens.chat.state :as view.state]
+            [status-im.chat.models.loading :as chat.loading]
+            [status-im.utils.platform :as platform]))
 
-(defn- prepare-message
-  [message current-chat?]
-  (cond-> message
-    current-chat?
-    (assoc :seen true)))
+(defn- message-loaded?
+  [db chat-id message-id]
+  (get-in db [:messages chat-id message-id]))
 
+(defn- earlier-than-deleted-at?
+  [db chat-id clock-value]
+  (>= (get-in db [:chats chat-id :deleted-at-clock-value]) clock-value))
+
+(defn add-timeline-message [acc chat-id message-id message]
+  (-> acc
+      (update-in [:db :messages chat-id] assoc message-id message)
+      (update-in [:db :message-lists chat-id] message-list/add message)))
+
+;;TODO this is too expensive, probably we could mark message somehow and just hide it in the UI
 (fx/defn rebuild-message-list
   [{:keys [db]} chat-id]
   {:db (assoc-in db [:message-lists chat-id]
                  (message-list/add-many nil (vals (get-in db [:messages chat-id]))))})
 
-(fx/defn hidden-message-marked-as-seen
-  {:events [::hidden-message-marked-as-seen]}
-  [{:keys [db] :as cofx} chat-id _ hidden-message-count]
-  (when (= 1 hidden-message-count)
-    {:db (update-in db [:chats chat-id]
-                    update
-                    :unviewed-messages-count dec)}))
-(fx/defn hide-message
+(defn hide-message
   "Hide chat message, rebuild message-list"
-  [{:keys [db] :as cofx} chat-id {:keys [seen message-id]}]
-  (fx/merge cofx
-            {:db (update-in db [:messages chat-id] dissoc message-id)}
-            (data-store.messages/mark-messages-seen chat-id [message-id] #(re-frame/dispatch [::hidden-message-marked-as-seen %1 %2 %3]))
-            (rebuild-message-list chat-id)))
+  [{:keys [db]} chat-id message-id]
+  ;;TODO this is too expensive, probably we could mark message somehow and just hide it in the UI
+  (rebuild-message-list {:db (update-in db [:messages chat-id] dissoc message-id)} chat-id))
 
-(fx/defn add-message
-  [{:keys [db] :as cofx}
-   {{:keys [chat-id message-id replace timestamp from] :as message} :message
-    :keys [seen-by-user?]}]
-  (let [current-public-key (multiaccounts.model/current-public-key cofx)
-        message-to-be-removed (when replace
-                                (get-in db [:messages chat-id replace]))
-        prepared-message (prepare-message message seen-by-user?)]
-    (fx/merge cofx
-              (when message-to-be-removed
-                (hide-message chat-id message-to-be-removed))
-              (fn [{:keys [db]}]
-                {:db (cond-> (-> db
-                                 ;; We should not be always adding to the list, as it does not make sense
-                                 ;; if the chat has not been initialized, but run into
-                                 ;; some troubles disabling it, so next time
-                                 (update-in [:messages chat-id] assoc message-id prepared-message)
-                                 (update-in [:message-lists chat-id] message-list/add prepared-message))
-                       (and (not seen-by-user?)
-                            (not= from current-public-key)
-                            (not (get-in db [:chats chat-id :profile-public-key]))
-                            (not (get-in db [:chats chat-id :timeline?])))
-                       (update-in [:chats chat-id :loaded-unviewed-messages-ids]
-                                  (fnil conj #{}) message-id))}))))
+(fx/defn join-times-messages-checked
+  "The key :might-have-join-time-messages? in public chats signals that
+  the public chat is freshly (re)created and requests for messages to the
+  mailserver for the topic has not completed yet. Likewise, the key
+  :join-time-mail-request-id is associated a little bit after, to signal that
+  the request to mailserver was a success. When request is signalled complete
+  by mailserver, corresponding event :chat.ui/join-times-messages-checked
+  dissociates these two fileds via this function, thereby signalling that the
+  public chat is not fresh anymore."
+  {:events [:chat/join-times-messages-checked]}
+  [{:keys [db] :as cofx} chat-ids]
+  (reduce (fn [acc chat-id]
+            (cond-> acc
+              (:might-have-join-time-messages? (chat-model/get-chat cofx chat-id))
+              (update :db #(chat-model/dissoc-join-time-fields % chat-id))))
+          {:db db}
+          chat-ids))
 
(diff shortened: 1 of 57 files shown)


flexsurfer:
@Serhy when you have time, could you give it a try pls, thanks

Serhy:
@flexsurfer latest build: https://status-im-prs.ams3.cdn.digitaloceanspaces.com/StatusIm-210223-120731-485acb-pr-universal.apk
**Issue1**:
No message appears on chat view if user receives it in that chat. However if to re-navigate to chat from Chats - messages appear. This includes `Status` tab - no statuses (text/image) update appears there after it sent (only re-login helps).
Reproduction:
- UserA and UserB staying on the same chat view of the same public chat (OR chatting in 1-1 chat) 
- UserA sends text message
- UserB checks if message appeared

**Issue2** 
No update of the unread counter for received messages in 1-1 chat

**Issue3**
No 'Share public key' pop up appears (`Profile` -> top-right `Share` button). Well it seem something triggered and maybe appeared somewhere on background of another views because when I try to share address from `Wallet` then at first some other pop-up blinks (which I assume is the Public key pop-up). 
The next I think has the same root cause: Open Status -> Wallet -> Ethereum account -> tap 'Share' button. Then try to open Share public key. I get this:
Screenshot 2021-02-23 at 16 03 00

flexsurfer:
thanks @Serhy lets postpone it for now

Serhy:
Tested https://status-im-prs.ams3.cdn.digitaloceanspaces.com/StatusIm-210224-134838-f889c9-pr-universal.apk latest build (step 12) 

1) It's interesting but I have almost no lags when bunch of messages received: when 100 stickers is receiving in public chat - there is no delay between screen navigations. Maybe it's because no counters or something related to issues in above comment, but it's seems much better in this relation.

2) Joining chat with lots of messages has a noticeably different behavior to develop:
In develop we first load messages and then navigate happens with last 10 messages (so it lags on Chats view)
In PR when tapping on chat from Chats - we navigate straight to the empty chat view and then - latest batch of 10 messages appears. Honestly, I don't know what is better :) But it would be more responsive in latest case (so how currently in this PR) ** if**  we add some spinning wheel. But again, - this delay is visible on low-mid devices. For GalaxyS20-like devices it's better to stay with behavior we have in develop to exclude blinking (because blinks if we do a quick open of empty chat + then load batch of messages) of messages in navigation to each chat.

Difference in videos:

Develop build: 

https://user-images.githubusercontent.com/8749671/109026708-b492fa00-76c8-11eb-82d2-26b7d5a3c494.mov


PR build:


PR_build_nav_to_chat

flexsurfer:
thanks @Serhy try latest two builds, on both fast and slow devices. which you prefer more ?

Serhy:
Tested last build on Android 8 (Galaxy J7) slow device (it's also visible in Chats view `Chat - (true -27)` ) and it does not look better. It's clearly visible for instance in chat with images.
 
When navigating to 1-1 chat view with 20 images, - the time from tap on chat till images appear in chat view takes:
 
In nightly build (`2f1c644`): 2s (video - https://drive.google.com/file/d/1M7VStbOQ6yrfWB_iGpYSxHtrStPg5_8v/view?usp=sharing)
in PR build (`80c246`): 4.2s (video - https://drive.google.com/file/d/1fA-RBNvl6BXobCkwLVeZEuL1oxMg47KE/view?usp=sharing)

Similar but better situation with navigating to chat with text messages. Public chat with 100 test messages (from tap on chat till the screen fulfilled with messages):
PR build: 3s
Nightly build: 2.75

Also on low-level devices I see that we load batches of 5 images, which looks worse to me versus build when we load batch of 10 messages. I mean that blinking of first 5 text messages then next 5 message and the rest (so chat view is full of one-line text message) looks more glitched than if we keep spinning wheel but then 10-20 messages appear on chat with the rest messages loading somewhere "on background" (just not visible to user). 

I also checked the `Disabled perf mode` mode (so pure text shown, no bubbles wrapping it up). I would say if we have same performance on regular chat that would be amazing :) 
I can scroll till the top of 100 text message chat minimum 5 times faster than with PerfMode disabled (regular chat view).

So overall, based on example with images, - something in this PR makes messages loading of messages slower than in develop builds. But disabling renderer of messages (bubbles, text parsing etc) makes chat performance very nice (we could even enable 100 messages batches in that case when navigate to chat :) So clearly message rendering plays significant role in slowing chat performance

flexsurfer:
we shouldn't have to discuss this, because this screen must be opened instantly, but we have a delay and imo its better to give faster feedback, and longer spinner time because in that case app feels more responsive, because if you press and nothing is happening for 2s it's a bad feeling that app is slow, if you press and you see a reaction and then loader you just think that something loading but app is responsive, but again we should open this screen instantly in that case we don't need to choose between two evils

flexsurfer:
@Serhy ready for testing, we need to test chats , profiles and timeline

src/status_im/chat/models/message.cljs
```diff
@@ -1,214 +1,186 @@
...
+
+(defn reduce-js-messages [{:keys [db] :as acc} ^js message-js]
+  (let [chat-id (.-localChatId message-js)
+        clock-value (.-clock message-js)
+        message-id (.-id message-js)
+        current-chat-id (:current-chat-id db)
+        cursor-clock-value (get-in db [:pagination-info current-chat-id :cursor-clock-value])]
+    ;;ignore not opened chats and earlier clock
+    (if (and (get-in db [:pagination-info chat-id :messages-initialized?])
+             ;;TODO why do we need this ?
```
cammellos, line 125: mmh, I think this might actually not be needed anymore, the check is done in status-go

src/status_im/notifications/local.cljs
```diff
@@ -189,10 +190,13 @@
            local-push-android))
 
 (fx/defn process
-  [cofx evt]
-  (if platform/ios?
-    {::local-push-ios evt}
-    (local-notification-android cofx evt)))
+  [cofx evt-js]
+  (when (and (= (get-in cofx [:db :app-state]) "inactive")
+             (get-in cofx [:db :multiaccount :remote-push-notifications-enabled?]))
```
cammellos: Not sure this is correct, on android this will work and `remove-push-notifications-enabled?` is only for IOS.

What's the reason for these changes? 
If it's just performance https://github.com/status-im/status-react/issues/11806 should help, as we check on status-go if notifications are enabled and it moves all the processing for messages in status-go (not for transactions yet).

src/status_im/transport/message/core.cljs
```diff
@@ -56,76 +31,143 @@
...
+      (update-in [:db :chats chat-id :unviewed-messages-count] inc)
+
+      ;;conj incoming transaction for :watch-tx
+      (not (string/blank? tx-hash))
+      (update :transactions conj tx-hash)
+
+      :always
+      (update :chats conj chat-id))))
+
+(defn sort-js-messages!
```
cammellos, line 135: This is sorting incoming messages? (i.e the ones in `messages.new`)
if that's so we should just add it in status-go, as currently order is the receiving time, but if the client wants it ordered in canonical order, that's likely the place.

cammellos (changes requested):

src/status_im/notifications/local.cljs
flexsurfer: ok, i'll remove this condition then, thanks

src/status_im/ui/screens/status/views.cljs
```diff
@@ -182,37 +181,39 @@
...
+            (if (= :timeline (:tab @state))
+              (i18n/label :t/statuses-descr)
+              (i18n/label :t/statuses-my-status-descr))]]]
+         [list/flat-list
+          {:key-fn                    #(or (:message-id %) (:value %))
+           :render-data               {:timeline (= :timeline (:tab @state))
+                                       :account  account}
+           :render-fn                 render-message
+           :data                      messages
+           ;:on-viewable-items-changed chat.views/on-viewable-items-changed
```
cammellos: Could be removed if not used

cammellos (approved):
Seems all ok from what I can see, if @Serhy validates that there's perf. improvements and no regressions, good to merge :) 
Thanks for all the hard work @flexsurfer !

Serhy:
Thanks @flexsurfer! Regarding performance on chats I tested with short scenario of loading 20 large (4000 char each) text messages in public chat and the time it takes between tap on chat till the actual message appears on chat:

- PR build: 2.3s
- Release build: 2-2.6s (not sure from it depends but from 4 tests time very)

Anyway, the time of message appearing after tap on chat is relatively the same. I mean performance in this relation of loading messages or slightly better or kept on the same level. But with the previous build (of this PR) it's clear the message item rendering is what should definitely improve it.

However! Because we don't load messages after tap on chat (redirect to chat happens very fast) + after redirect to chat indicate with spinning wheel that something happens now. That way of opening chat looks much better, more responsive even that it has the same time of message loading.

Also, that improvements of: 
- no more jumping messages when joining to chat (they are already sorted and nicely appear ready to read)
- no reset of chat/timeline when opening user profile view (when navigating back user returns and see the latest place he was at before opening profile)
these are visually great improvement too!

I didn't cover iOS yet (tested chat's, timeline only Android 11, 9 and 6 for now).

One little bug is that message input lost fixed height when typing multiline message:

flexsurfer:
thanks @Serhy fixed

Serhy:
Now it looks great for me (iOS14, Androids 11, 8, 6) - everything as described in https://github.com/status-im/status-react/pull/11754#issuecomment-795843008 + failed e2e (was valid fails) revered issues are fixed now. 

One minor issue with navigation which is different vs develop we may fix separately (it's when blocking user from Group chat - there is a redirect to Chats main view instead back to Group Info screen.  I.e. `Chats` -> `Group Chat` -> `Group Info` -> `...`upon user -> `Block user` -> result: navigation to Chats view while expected navigation to Group Info view)

Mega cool job is done here, @flexsurfer, thanks a lot!

PR #11761, opened 2021-02-17
ci: adjust behavior of PUBLISH to be true for releases

jakubgs:
Otherwise release builds never create the GitHub draft releases.
Not unless a user explicitly stats a build with `PUBLISH: true`.


cammellos (approved):

jakubgs:
I tested this with https://ci.status.im/job/status-react/job/manual/116/ and this will only take effect from the second build. Because the first one will have `false`, while the successive builds will use the set default of `true`. Which I guess is fine for releases.

PR #11773, opened 2021-02-19
use the currently selected account for signing messages/typed data

### LINKED ISSUE #11759: Sign with Keycard doesn't use the correct account when signing messages/typed data
# Bug Report

## Problem

When signing a message or typed data from a dApp using Keycard, the sign method is called without a path, which would use the current path on card instead of the active account. When using the default account this does not give problems, but when addings accounts to the multiaccount it means the signature generated will be wrong one.

## Steps to reproduce

Using a Keycard multiaccount, create a second account. Open the browser and select that second account. Try to login to cryptokitties with that account and login will fail (because the signature will be generated by the wrong account)

## Additional Information
Test on release 1.10.1, develop branch from 17.02.2021 and PR11540. PR11540 tested on iOS as well.

bitgamma:
Fixes #11759 by searching the "from" field not only in the transaction but also in the message map. If no "from" field is specified, the default account is used as a fallback.

status: ready

### DIFF
--- a/src/status_im/keycard/sign.cljs
+++ b/src/status_im/keycard/sign.cljs
@@ -21,7 +21,7 @@
         data                 (get-in db [:keycard :data])
         typed?               (get-in db [:keycard :typed?])
         pin                  (common/vector->string (get-in db [:keycard :pin :sign]))
-        from                 (get-in db [:signing/tx :from :address])
+        from                 (or (get-in db [:signing/tx :from :address]) (get-in db [:signing/tx :message :from]) (ethereum/default-address db))
         path                 (reduce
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

bitgamma:
@churik does this need to be manually tested before merging? What I have tested so far:

1) Login to Cryptokitties using Keycard with both the default account and a second account in the same multiaccount
2) Sending transactions from each account
3) Signing typed data from each account

on OnePlus 6/Android 10 and iPhone 11/iOS 14

churik:
not really, e2e failures are not related to PR. Thank you!

PR #11791, opened 2021-02-24
fdroid: add release-fdroid target and update docs

jakubgs:
With these targets we can simplify the format of the F-Droid YAML metadata file which defines how to run `status-react` builds.

Also added download badges to the `README.md`:
image


Makefile
```diff
@@ -156,25 +156,37 @@ keystore: export KEYSTORE_PATH ?= $(HOME)/.gradle/status-im.keystore
...
+	scripts/release-android.sh
+
+release-fdroid: export BUILD_ENV ?= prod
+release-fdroid: export BUILD_TYPE = release
+release-fdroid: export ANDROID_APK_SIGNED = false
+release-fdroid: export ANDROID_ABI_SPLIT ?= false
+release-fdroid: export ANDROID_ABI_INCLUDE = armeabi-v7a;arm64-v8a;x86;x86_64
+release-fdroid: ##@build Build release for F-Droid
+	echo "GOOGLE_FREE=1" >> .env.release
 	scripts/release-android.sh
```
jakubgs, line 184: I don't think I like this `echo "GOOGLE_FREE=1" >> .env.release` in here. It seems clunky.

cammellos: I don't think we need it at the moment, we completely remove the dependency on the other branch.
According to the developer, if we upgrade to 8 rn-device info we would not need it. 
but I'd have to look more into it

cammellos (approved):

Makefile
jakubgs: What about this then:
https://github.com/status-im/status-react/blob/f71faee5c2a869485e9dfafe3de5e704838a6ca2/src/status_im/utils/config.cljs#L47
https://github.com/status-im/status-react/blob/f71faee5c2a869485e9dfafe3de5e704838a6ca2/src/status_im/acquisition/core.cljs#L111-L123
Does it still matter? Or should it be removed?

jakubgs: And I'd assume this:
https://github.com/status-im/status-react/blob/f71faee5c2a869485e9dfafe3de5e704838a6ca2/react-native.config.js#L1
And this:
https://github.com/status-im/status-react/blob/f71faee5c2a869485e9dfafe3de5e704838a6ca2/android/app/build.gradle#L129
Would also be obsolete?

jakubgs: I'm gonna get rid of the line in `status-react/android/app/build.gradle` first since it seems to do nothing.

jakubgs:
I tested 250d8a452bc52c498f309b459ee3495ccdf4844e with this config using F-Droid docker image:
```yaml
  - versionName: 1.12.0
    versionCode: 2021022512
    commit: 250d8a452bc52c498f309b459ee3495ccdf4844e
    sudo: make fdroid-nix-dir
    init: nix/scripts/setup.sh
    output: result/app-release-unsigned.apk
    scanignore: android/build.gradle
    scandelete: ios
    build: make release-fdroid BUILD_NUMBER=$$VERCODE$$
```
And it builds fine:
```
2021-02-25 11:07:53,449 INFO: Running 'build' commands in build/im.status.ethereum
2021-02-25 11:21:17,874 INFO: Successfully built version 1.12.0 of im.status.ethereum from 250d8a452bc52c498f309b459ee3495ccdf4844e
2021-02-25 11:21:18,160 INFO: success: im.status.ethereum
2021-02-25 11:21:18,161 INFO: Finished
2021-02-25 11:21:18,161 INFO: 1 build succeeded
```
Though I did cheat a bit and added `echo "1.12.0" > VERSION` in `build` to avoid changing the branch.

PR #11857, opened 2021-03-09
Add RPC usage stats screen

rasom:
status: ready


flexsurfer (approved):
so this screen is for release as well ?

rasom:
@flexsurfer why not? We can hide it but it might be useful even in release sometime.

PR #11862, opened 2021-03-10
[wallet] Store last fetched balance and nonce in db

rasom:
Now each iteration of checking for new transfers requires two less requests, stats after pulling to refresh:
before

 
after

status: ready


flexsurfer (approved):

rasom:
@churik @Serhy I tried to reproduce `test_send_eth_from_wallet_to_address`, `test_add_and_delete_watch_only_account_to_multiaccount_instance` but no luck, any ideas what's wrong there?

Serhy:
@rasom the issue with the tests here are that balance is not fetched after account recover in e2e build. 
In develop build the same steps and account works fine (ETH balance fetched fine)
**To reproduce:**
1) Install https://status-im-prs.ams3.cdn.digitaloceanspaces.com/StatusIm-210312-094134-b7d02c-pr-x86.apk build on emulator
2) Recover account with seed phrase: `warm beach mind real stay oil evil giggle core tilt paddle invest`
3) Once recovered - open Wallet and wait till balance is fetched

Note: after app restart and login in to account OR `Scan Tokens` option - balance fetches OK, but not right after account recover. The same applies to Transaction History. Pull-to-refresh doesn't help.

rasom:
That must be something architecture specific, I'm out of ideas otherwise

rasom:
@Serhy 
looks like I'm extremely unlucky 
Screenshot 2021-03-13 at 20 44 06

rasom:
@Serhy which os version and do you use on emulator?

Serhy:
@rasom I'm using Android 11 and Android 10 - reproduced on both. I tried several different accounts and indeed it's all fine loading balance. However, exactly this account `warm beach mind real stay oil evil giggle core tilt paddle invest` - fails to load balance. Maybe because there are many transactions (~5k)

rasom:
@Serhy I likely found the cause, will share details tomorrow or on Monday

Serhy:
Looks good to me now on iOS and Android 11 (balance loads under WiFi and LTE), can load transaction history under LTE too. Failed e2e are false failures. Thanks @rasom !

Issue #11387, opened 2020-11-06
Handle unpaired cards


bitgamma:
# Bug Report

## Problem

You have a Keycard multiaccount. Using another tool, you remove the pairing between the card and Status. When trying to perform any operation with the card you are not prompted to pair the card again and the operation fails. The only workaround is to reinstall the app.

#### Expected behavior

If the pairing has been broken for whatever reason, prompt the user to restore the pairing

#### Actual behavior

The pairing remains broken with no way to recover

### Reproduction

- Open Status
- Create an account with Keycard
- Log out in Status
- Pair the card with Keycard Connect
- Tap Unpair Others in Keycard Connect
- Try to login in Status

guylouis:
To be tested again when testing for keycard resumes. I thought this was managed. 
Doesn't it depend on which operation the user engages with the unpaired card (which action is not mentioned above) ?

PR #11863, opened 2021-03-10
move pairing handling to native module

### LINKED ISSUE #11389: Pairing is not saved if account recovery is interrupted
# Bug Report

## Problem

During the "Recover with Keycard" process you are prompted to pair the card and then enter the PIN. If you cancel after pairing but before completing the procedure, the pairing is not saved in Status and when you try again you will need to pair again. The problem is that on card the pairing slot is taken anyway and you have no way to recover it.

#### Expected behavior

As soon as a pairing is done, it is saved to storage. When resuming the account recovery the pairing step is skipped if the pairing already exists.

#### Actual behavior

The pairing is discarded and the pairing slot on card is wasted

### Reproduction

- Open Status
- Choose account recovery with Keycard
- Pair the card
- Cancel in any way you want
- Repeat step 2. You will prompted to pair the card again. The available pairing slots count is decreased

bitgamma:
Moves pairing handling to the native module, allowing to select the correct pairing after the card has been tapped and thus decoupling the account from pairing altogether.

Closes #11389 and #11387. Allows implementing card backup and using multiple cards with the same account

status: ready

### DIFF
--- a/src/status_im/keycard/core.cljs
+++ b/src/status_im/keycard/core.cljs
@@ -11,6 +11,7 @@
             [status-im.keycard.recovery :as recovery]
             [status-im.keycard.sign :as sign]
             [status-im.keycard.wallet :as wallet]
+            [status-im.keycard.card :as card]
             [status-im.i18n.i18n :as i18n]
             [status-im.multiaccounts.recover.core :as multiaccounts.recover]
             [status-im.multiaccounts.update.core :as multiaccounts.update]
@@ -47,6 +48,16 @@
                                                 :current             []}))}
             (common/listen-to-hardware-back-button)
             (navigation/navigate-replace :keycard-recovery-pin nil)))
+
+(fx/defn load-recovery-pin-screen-with-pairings
+  [{:keys [db] :as cofx}]
+  (let [instance-uid (get-in db [:keycard :application-info :instance-uid])
+        {:keys [pairing paired-on]} (get-in db [:keycard :pairings (keyword instance-uid)])]
+    (fx/merge cofx
+              {:db (-> db
+                       (assoc-in [:keycard :secrets :pairing] pairing)
+                       (assoc-in [:keycard :secrets :paired-on] paired-on))}
+              (load-recovery-pin-screen))))
 
 (fx/defn proceed-setup-with-initialized-card
   [{:keys [db] :as cofx} flow instance-uid]
@@ -133,8 +144,7 @@
 (fx/defn on-unblock-pin-success
   {:events [:keycard.callback/on-unblock-pin-success]}
   [{:keys [db] :as cofx}]
-  (let [pairing   (common/get-pairing db)
-        reset-pin (get-in db [:keycard :pin :reset])]
+  (let [reset-pin (get-in db [:keycard :pin :reset])]
     (fx/merge cofx
               {:db
                (-> db
@@ -156,8 +166,7 @@
 (fx/defn on-unblock-pin-error
   {:events [:keycard.callback/on-unblock-pin-error]}
   [{:keys [db] :as cofx} error]
-  (let [pairing       (common/get-pairing db)
-        tag-was-lost? (common/tag-lost? (:error error))
+  (let [tag-was-lost? (common/tag-lost? (:error error))
         puk-retries (common/pin-retries (:error error))]
     (log/debug "[keycard] unblock pin error" error)
     (when-not tag-was-lost?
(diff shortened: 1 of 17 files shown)


src/status_im/keycard/core.cljs
```diff
@@ -48,6 +49,16 @@
             (common/listen-to-hardware-back-button)
             (navigation/navigate-replace :keycard-recovery-pin nil)))
 
+(fx/defn load-recovery-pin-screen-with-pairings
+  [{:keys [db] :as cofx}]
+  (let [instance-uid (get-in db [:keycard :application-info :instance-uid])
+        {:keys [pairing paired-on]} ((get-in db [:keycard :pairings]) (keyword instance-uid))]
```
flexsurfer:
```suggestion
        {:keys [pairing paired-on]} (get-in db [:keycard :pairings (keyword instance-uid)])]
```

src/status_im/keycard/core.cljs
```diff
@@ -409,8 +416,8 @@
         instance-uid (get-in db [:keycard :application-info :instance-uid])
         multiaccount (common/find-multiaccount-by-keycard-instance-uid db instance-uid)
         paired-on    (utils.datetime/timestamp)
-        pairings     (assoc (get-in db [:keycard :pairings]) instance-uid {:pairing   pairing
-                                                                           :paired-on paired-on})
+        pairings     (assoc (dissoc (get-in db [:keycard :pairings]) (keyword instance-uid)) instance-uid {:pairing   pairing
```
flexsurfer: better to use `->` here

flexsurfer: for better readability

src/status_im/keycard/real_keycard.cljs
```diff
@@ -80,6 +80,10 @@
   [callback]
   (.addListener ^js event-emitter "keyCardOnNFCDisabled" callback))
 
+(defn set-pairings [{:keys [pairings]}]
+  (let [pairings' (or pairings {})]
+    (.. status-keycard (setPairings (clj->js pairings')))))
```
flexsurfer: no need in extra let , just use `(or pairings {})` directly

src/status_im/keycard/sign.cljs
```diff
@@ -13,10 +13,9 @@
   {:events [:keycard/sign]}
   [{:keys [db] :as cofx}]
   (let [card-connected?      (get-in db [:keycard :card-connected?])
-        pairing              (common/get-pairing db)
-        keycard-instance-uid (get-in db [:multiaccount :keycard-instance-uid])
-        instance-uid         (get-in db [:keycard :application-info :instance-uid])
-        keycard-match?       (= keycard-instance-uid instance-uid)
+        key-uid (get-in db [:multiaccount :key-uid])
```
flexsurfer: align

flexsurfer (approved):

src/status_im/keycard/core.cljs
bitgamma: Thanks for the review! All remarks have now been fixed

bitgamma:
@churik is it possible to re-trigger E2E tests on this? They all failed because I forgot to update the simulated keycard, but now it should be fixed.

churik:
sure, will do

bitgamma:
@churik I think the remaing e2e test failures are because of the changes, mostly because of two reasons

test_keycard_can_recover_keycard_account_card_pairing: fails because the card is already paired (by the test executing before), so the pairing screen is skipped. This is a new feature but it is the desired behaviour. If the point of the test is doing a new pairing then we need to introduce a new method that simulates a different card for it, wdyt?

the others fail because apparenly the tapped card is not the correct one. This happens because we now check the key uid instead of the instance uid to see if the card matches and maybe it isn't set correctly. Again, this is the desired behaviour because we want to allow using multiple cards using the same account (if those cards have the same keys on them).

by the way, I think the PR is also ready for manual testing as needed

churik:
@bitgamma 
I tested it manually and it works. In order to have it merged we need to add possibility to emulate using keycard in all failed tests (so I believe some fixes to kk framework in e2e tests are required), as it breaks a lot of e2e.

> the others fail because apparenly the tapped card is not the correct one. This happens because we now check the key uid instead of the instance uid to see if the card matches and maybe it isn't set correctly. Again, this is the desired behaviour because we want to allow using multiple cards using the same account (if those cards have the same keys on them).

so we need somehow to emulate behvior for using it in e2e builds.

bitgamma:
@churik 
Thanks for the test! I hope I can fix all the failing e2e by improving simulated_keycard to behave in a way more similar to the real card, without need for changes to the tests. The latest commit hopefully fixes 7 of those tests.

bitgamma:
@churik I have now manually tried to follow the steps done by the tests on a build with keycard-test-menu/simulated_keycard enabled and I am pretty sure I fixed the issues.

Could you retrigger them one more time? Sorry and thanks for your patience!

churik:
failures are not related to your PR, ready to be merged, thank you!

PR #11867, opened 2021-03-11
Private group chats confirmations

### LINKED ISSUE #11852: Add "delivered" message confirmations to private group chats
```
As a user
I want to know my message has been delivered to the recipient
So that I have confidence that it will eventually be read
```

### Description

Next to each private group sent chat message there should be an icon indicating the state of the message:

- The message is actually being sent (sending state)
- The message has been propagated to the network (sent state)
- The message has reached its destination (delivered state)

### Out of scope

- Public chats

### Designs
https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=1%3A2

### Implementation details

A message can be marked as delivered once is acknowledged at the datasync level https://github.com/vacp2p/mvds  by all the participants in the chat.

cammellos:
Fixes: #11852

### Testing

Messages in private group chats should be confirmed once at least one recipients has received a message (at least on one device).
If everyone is offline, the message should not be marked as confirmed.

status: ready


flexsurfer (approved):

cammellos:
@qoqobolo this is ready to be tested again, I have changed the behavior according to the UX team so that if needs to be confirmed by at least one and not everyone.
Thanks!

qoqobolo:
Checked:
- foreground / background
- logged in / logged out
- PN background service is enabled, app is in background
- switching between online and offline
- mobile network
- status nodes disabled
- after switching to another status node (including custom node)
- all message types (text, reply, link, tag, mentions, long collapsed messages, emoji, image with caption)
- changes in group settings
- groups with or without members
- sending messages before / after someone joins
- ENS
- public and 1-1 chats (no changes)

@cammellos thank you for PR!
It's ready to be merged.

PR #11886, opened 2021-03-16
e2e: push notifications

churik:
- add check for PN in group chat (test will be failed until fix in develop)
- add check for issue with unread indicator (test will be failed until fix in develop)
- add check for PN in public chats

### DIFF
--- a/test/appium/tests/atomic/chats/test_public.py
+++ b/test/appium/tests/atomic/chats/test_public.py
@@ -57,32 +57,39 @@ class TestPublicChatMultipleDevice(MultipleDeviceTestCase):
     @marks.critical
     def test_unread_messages_counter_public_chat(self):
         self.create_drivers(2)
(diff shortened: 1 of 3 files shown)


test/appium/tests/atomic/chats/test_public.py
```diff
@@ -57,31 +57,39 @@ def test_public_chat_messaging(self):
...
+        chat_2.chat_element_by_text(message_2).wait_for_element(20)
+
+        self.drivers[1].close_app()
+        self.drivers[1].launch_app()
+        SignInView(self.drivers[1]).sign_in()
+        chat_element = home_2.get_chat('#' + chat_name)
+        if chat_element.new_messages_public_chat.is_element_displayed():
+            self.errors.append('New messages counter is shown after relogin')
+
+        import time
```
Serhy: Do we need it here?

Serhy (approved):

PR #11901, opened 2021-03-18
Feature/anon metrics

Samyoul:
## What's Changed?

Added an interceptor to catch the payloads of specific events and send the filtered data to `status-go` for further processing.

## Why Make the Change?

This functionality is the foundational change to `status-react` to compliment the functionality introduced to `status-go` in this PR https://github.com/status-im/status-go/pull/2170.

Effectively the interceptor's function is to relay event data about specific user actions to `status-go` ensuring that no sensitive data is collected by or even passed to `status-go`. The `status-go` side of this functionality also has a validation mechanism that explicitly rejects incoming data that does not adhere to the strict definition of what data is allowed.

### DIFF
--- a/src/status_im/anon_metrics/interceptors.cljs
+++ b/src/status_im/anon_metrics/interceptors.cljs
@@ -1,0 +1,27 @@
+(ns status-im.anon-metrics.interceptors
+  (:require [status-im.ethereum.json-rpc :as json-rpc]
+            [taoensso.timbre :as log]
+            [re-frame.interceptor :refer [->interceptor]]
+            [status-im.utils.platform :as platform]
+            [status-im.utils.build :as build]
+            [status-im.anon-metrics.transformers :as txf]))
+
+(defn transform-and-log [context]
+  (log/info :catch-event-fn (get-in context [:coeffects :event]))
+  (when-let [transformed-payload (txf/transform context)]
+    (json-rpc/call {:method "appmetrics_saveAppMetrics"
+                    :params [[{:event (-> context :coeffects :event first)
+                               :value transformed-payload
+                               :app_version build/version
+                               :os platform/os}]]
+                    :on-failure log/error})))
+
+(defn catch-events-before [context]
+  (log/info "catch-events/interceptor fired")
+  (transform-and-log context)
+  context)
+
+(def catch-events
+  (->interceptor
+   :id     :catch-events
+   :before catch-events-before))
--- a/src/status_im/anon_metrics/transformers.cljs
+++ b/src/status_im/anon_metrics/transformers.cljs
@@ -1,0 +1,17 @@
+(ns status-im.anon-metrics.transformers
+  "`status-go` defines the shape of expected events so that we don't overcollect data (by mistake or intentionally).
+  `transformers` ns transform the event payload to match the expected shape on `status-go` side.")
+
+(defn navigate-to-txf [event]
+  {:view-id (second event)
+   :params (-> event
+               (nth 2)
+               (select-keys [:screen]))})
+
+(def transformations
+  {:navigate-to navigate-to-txf})
+
+(defn transform [ctx]
+  (let [event (-> ctx :coeffects :event)]
+    (when-let [txf (-> event first transformations)]
+      (txf event))))
--- a/src/status_im/ethereum/json_rpc.cljs
+++ b/src/status_im/ethereum/json_rpc.cljs
@@ -178,7 +178,8 @@
    "mailservers_addChatRequestRange" {}
    "mailservers_addChatRequestRanges" {}
    "mailservers_getChatRequestRanges" {}
-   "mailservers_deleteChatRequestRange" {}})
+   "mailservers_deleteChatRequestRange" {}
+   "appmetrics_saveAppMetrics" {}})
 
 (defn on-error-retry
   [call-method {:keys [method number-of-retries delay on-error] :as arg}]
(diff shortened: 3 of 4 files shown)


cammellos:
Code looks ok, nice work!

We are aware that this is only catching a subset of the navigation events as any function that calls `navigate-to-cofx` directly will not be recorded (it does not generate an event)? 

If that's the case I would leave a comment explaining it and what we will be do about it, as this potentially can't be yet used to collect data, as results will be skewed towards those views that dispatch to navigate.

https://github.com/status-im/status-react/search?q=navigate-to-cofx&type=code

src/status_im/anon_metrics/interceptors.cljs
```diff
@@ -0,0 +1,27 @@
...
+
+(defn catch-events-before [context]
+  (log/info "catch-events/interceptor fired")
+  (catch-events-fn context)
+  context)
+
+(def catch-events
+  (->interceptor
+    :id     :catch-events
+    :before catch-events))
```
shivekkhurana: Should be `catch-events-before`

src/status_im/anon_metrics/interceptors.cljs
```diff
@@ -0,0 +1,27 @@
+(ns status-im.anon-metrics.interceptors
+  (:require [status-im.ethereum.json-rpc :as json-rpc]
+            [taoensso.timbre :as log]
+            [re-frame.interceptor :refer [->interceptor get-coeffect]]
+            [status-im.utils.platform :as platform]
+            [status-im.utils.build :as build]
+            [status-im.anon-metrics.transformers :as txf]))
+
+(defn catch-events-fn [context]
```
shivekkhurana: `catch-events-fn` is a generic name, what do you think about `transform-and-log` or `transform-and-write-to-db`?

shivekkhurana (changes requested):

src/status_im/anon_metrics/interceptors.cljs
```diff
@@ -0,0 +1,27 @@
...
+  (log/info :catch-event-fn (get-in context [:coeffects :event]))
+  (when-let [transformed-payload (txf/transform context)]
+       (json-rpc/call {:method "appmetrics_validateAppMetrics"
```
shivekkhurana: We want to `appmetrics_saveAppMetrics`, not validate.

src/status_im/anon_metrics/interceptors.cljs
```diff
@@ -0,0 +1,27 @@
...
+                       :params [[{:event (get-in context [:coeffects :event :event-type])
+                                  :val transformed-payload
```
shivekkhurana: The backend expects [`:value`](https://github.com/status-im/status-go/blob/6bea21c197431c381a2ac6641a21befc12b2131c/appmetrics/database.go#L19)

shivekkhurana (approved):
@cammellos I'll be working in this today and will add the missing navigation events. So, we can do without extra comments to indicate that some navigation effects are not being captured.

src/status_im/anon_metrics/transformers.cljs
```diff
@@ -0,0 +1,14 @@
+(ns status-im.anon-metrics.transformers)
```
flexsurfer: could you please add comments , why we need this, thanks

flexsurfer (changes requested):

src/status_im/anon_metrics/transformers.cljs
shivekkhurana: @flexsurfer done

flexsurfer (approved):

src/status_im/anon_metrics/transformers.cljs
```diff
@@ -0,0 +1,16 @@
...
+  {:view-id (second context)
+   :params (-> context
+               (nth 2)
+               (select-keys [:screen]))})
+
+(def transformations
+  {:navigate-to navigate-to-txf})
+
+(defn transform [ctx]
+  (when-let [txf (-> ctx first transformations)]
```
flexsurfer: could someone please explain this line `(-> ctx first transformations)` i can't understand it , thanks

flexsurfer (changes requested):

src/status_im/anon_metrics/transformers.cljs
shivekkhurana: This is the approximate shape of the `ctx` map:
```clojure
 (def ctx
    [:navigate-to ; event-type
     :home ; view-id
     {:screen "screen-naame"
      :some :other
      :element :ts}])
```

We want to lookup a transformation against `event-type`, i.e. the first element of the vector:

* `(-> ctx first) ;; => :navigate_to`
* `(-> :navigate_to transformations) ;; => the transformation for :navigate-to event`

flexsurfer: but we call it here with re-frame context `transformed-payload (txf/transform context)` probably we should call with `(get-in context [:coeffects :event])`  then?

shivekkhurana: yes you are right. This is a bug, the shape of `ctx` is not [the vector I commented above](https://github.com/status-im/status-react/pull/11901/files#r600299010). That vector is the shape of `(-> context :coeffects :event)`. 

Fixing it. Good catch 🥇

flexsurfer (approved):

cammellos:
@shivekkhurana @Samyoul I believe there's a bug in the code that causes an error when trying to use the app, I reverted the commit on develop, would you mind having a look at it when you have time?
thanks!

PR #11907, opened 2021-03-19
Add banning of users in communities

cammellos:
This commit adds bannning of users in communities for on-request and
invitation only communities.

status: ready


flexsurfer (approved):

shivekkhurana (approved):
Looks good

PR #11920, opened 2021-03-24
Removed region from on-ramping we don't want it

Samyoul:
We **DO NOT** want to show any region on the crypto on ramping, pm me for details.

Also see https://github.com/status-im/status-go/pull/2178

Latest version of `status-go` give the Latamex on ramp option and adds updated fees for Ramp

Simulator Screen Shot - iPhone 11 - 2021-03-24 at 15 09 27


flexsurfer (approved):

ios/Podfile.lock
```diff
@@ -268,7 +268,7 @@ PODS:
     - React
   - react-native-splash-screen (3.2.0):
     - React
-  - react-native-status-keycard (2.5.32):
+  - react-native-status-keycard (2.5.33):
```
flexsurfer: how this PR related to keycard ?

flexsurfer: also why PR doesn't change go version ?

Samyoul: Hmm, this is a branch directly from develop let me have a look and see what happened

PR #11924, opened 2021-03-24
Fix/#11821 rpc finetuning

### LINKED ISSUE #11821: Finetuning RPC calls for new accounts
# Problem
#11723 optimizes the number of RPC calls based on user activity. This reduces unnecessary calls to Infura for 'chat-only' users; allowing to support more accounts without reaching daily limits of our Infura subscription.

The user activity that triggers RPC calls can be 'smarter' to ensure that in specific cases where the user expects a balance update, this appears without manual user action. Currently the user can use pull to refresh _on an account_ to trigger RPC calls. This relies on user discovery and can leave some first time users wondering why their newly purchases or received crypto is not showing

## Requirements
Trigger RPC call on the following events:
- Users taps Share
- User taps Buy crypto
- User receives tx message
- Stop checking if a transaction is not received within 30 min; But check again any succeeding time the user logs into the account 

## Implementation
tbd

## Acceptance Criteria
Balance of an empty account is updated within a range of 0-2 minutes of an incoming transaction being visible on Etherscan

## Notes
**Current behavior**
- For a newly generated account + imported account that has no transfers in history, balance is not requested periodically, but when:
     1. user sent a tx
     2. user received a tx and pulled to refresh history or re-logged into the app

- If user has a custom RPC node we just check it every 4 mins regardless of tx history

## Future Steps
- Move pull to refresh to the main account to make it more easily discoverable
- To discuss: Postpone eth_getBlockByNumber request so that no Infura calls occur (IP leaking) at all for users who do not use Wallet

rasom:
fix #11821 

Transfer history checking is resumed for 30 mins for an empty account in case if 
- Users taps Share
- User taps Buy crypto
- User receives tx message

status: ready

### DIFF
--- a/src/status_im/wallet/core.cljs
+++ b/src/status_im/wallet/core.cljs
@@ -27,7 +27,8 @@
             [status-im.utils.datetime :as datetime]
             status-im.wallet.recipient.core
             [status-im.ui.screens.wallet.signing-phrase.views :as signing-phrase]
-            [status-im.async-storage.core :as async-storage]))
+            [status-im.async-storage.core :as async-storage]
+            [status-im.popover.core :as popover.core]))
 
 (defn get-balance
   [{:keys [address on-success on-error]}]
@@ -590,7 +591,9 @@
        (ethereum/current-network db))
       ms-4-min
 
-      (and max-block (zero? max-block))
+      (and max-block
+           (zero? max-block)
+           (nil? (get db :wallet/keep-watching-until-ms)))
       (log/info "[wallet] No transactions found")
 
       :else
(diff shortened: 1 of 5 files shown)


src/status_im/wallet/core.cljs
```diff
@@ -780,15 +783,34 @@
...
+        until-ms              (get db :wallet/keep-watching-until-ms)]
+    (when-not (and until-ms (> until-ms now))
+      (fx/merge
+       cofx
+       {:db (dissoc db :wallet/keep-watching-until-ms)}
+       (if (and (not non-empty-history?)
+                (empty? transfers)
+                (not custom-node?))
+         (clear-timeouts)
+         {:db (assoc db :wallet/non-empty-tx-history? true)})))))
```
flexsurfer: this won't work without you fix for fx/merge ?

flexsurfer (approved):

src/status_im/wallet/core.cljs
rasom: nope, this just won't work :) fixing

rasom: The thing is that `fx/merge` will just pass `cofx` to that map and this will return `nil` with no effects as a result.

rasom: fixed

Serhy:
I've reviewed failed e2e tests, manually tested, good to go (two of then will be updated separately today)

PR #11927, opened 2021-03-25
Add release guide

cammellos:
Update release guide.

### DIFF
--- a/doc/RELEASE_GUIDE.md
+++ b/doc/RELEASE_GUIDE.md
@@ -9,3 +9,98 @@ You can update Play Store releae metadata using `fastlane android upload_metadat
 But that requires credentials necessary for accessing Play Store API. The simpler way is to edit files contained within [`fastlane/metadata`](metadata) and run the following CI job:
 
 https://ci.status.im/job/status-react/job/tools/job/update-playstore-metadata/
+
+# Release process
+
+## Coordination
+
+All the coordination is done in the #release channel on discord.
+
+## Using the script
+
+You can use the script to cut a release branch:
+
+`scripts/cut-release.sh`
+
+This will create a branch:
+
+`release/a.b.x`
+
+Create a branch:
+
+`chore/update-release-to-$version`
+
+And if you have the `gh` command installed, it will automatically create a PR for you.
+
+## Manual steps
+
+### Create a branch
+
+First we need to create a release branch cut from develop.
+The name of the branch should be `release/a.b.x` where `x` is actually `x`, not 
+a variable, and `a` and `b` are the `MAJOR` and `MINOR` version of the app.
+
+For example, valid branch names are:
+
+`release/1.12.x` or `release/3.12.x`
+
+
+### Update version
+
+We need to update the `VERSION` file to reflect the release we want to build.
+
+
+## Building 
+
+Jenkins will automatically build this branch and create release candidates.
+Those will go through the QA process and the QA team will let us know when
+tested successfully.
+
+## Bugfixes before app is released
+
+Because both `status-react` and `status-go` might have newer commits that we don't 
+want to include, bug fixes should be applied separately to develop and the release branch.
+The best workflow is to send a PR to develop, get it merged and the cherry-pick on the
+release branch.
+
+If that's not practical for some reason (i.e develop has diverged significantly from release) 
+other workflows are possible.
+
+In order to change status-go, a similar workflow applies.
+First checkout the tag used in `status-go-version.json`.
+
+Create a branch on status-go named: `release/vx.y.z+hotfix.w` where
+
+`x` is the `MAJOR` version
+`y` is the `MINOR` version
+`z` is the `PATCH` version
+`w` is the `HOTFIX` version
+
+For example valid names are:
+
+`release/v0.62.3+hotfix.2`
+`release/v1.3.4+hotfix.4`
+
+Once the branch is ready to use and tested successfully, tag the branch:
+
+`git tag v0.62.3+hotfix.2` 
+
+and push the tag to origin:
+
+`git push origin v0.62.3+hotfix.2`
+
+Once that's done, update the status-react branch with the new tag:
+
+`make shell`
+
+`scripts/update-status-go.sh v0.62.3+hotfix.2`
+
+And commit and push the changes
+
+# Hotfix process
+
+The hotfix process is basically identical to the workflow above:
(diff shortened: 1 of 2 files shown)


doc/RELEASE_GUIDE.md
```diff
@@ -9,3 +9,79 @@ You can update Play Store releae metadata using `fastlane android upload_metadat
...
+a variable, and `a` and `b` are the `MAJOR` and `MINOR` version of the app.
+
+For example, valid branch names are:
+
+`release/1.12.x` or `release/3.12.x`
+
+
+## Update version
+
+We need to update the `VERSION` file to reflect the release we want to build.
```
jakubgs, line 50: Shouldn't the `VERSION` in `develop` be the next version, not the last version?
Or if not the next version the the release candidate of the next version: `1.13.0-rc`
And then when you create the `release/1.13.x` branch you update the `VERSION` in it to be `1.13.0` and the one in `develop` to be `1.14.0-rc`. Or something like that. Because we notoriously forget to update `VERSION` in `develop`.

doc/RELEASE_GUIDE.md
```diff
@@ -9,3 +9,79 @@ You can update Play Store releae metadata using `fastlane android upload_metadat
...
+The best workflow is to send a PR to develop, get it merged and the cherry-pick on the
+release branch.
+
+If that's not practical for some reason (i.e develop has diverged significantly from release) 
+other workflows are possible.
+
+In order to change status-go, a similar workflow applies.
+First checkout the tag used in `status-go-version.json`.
+
+Create a branch on status-go named: `release/vx.y.z+hotfix.w` where
```
jakubgs, line 72: Is it really a "hotfix" if we are "hotfixing" an unreleased version of the app? Sounds like just a normal verison.

jakubgs: Ah, but I do recall that there was a case when `status-go` already moved on beyond the release version. I guess in that case it makes sense.

cammellos: Well, it depends, technically it might not be a hotfix, but otherwise we'd have to have yet another denomination for this, we can't use normal versions as normal versions are taken up by the `develop` branch.
So in order not to introduce yet another piece of information, for the sake of simplicity I'd stick with `hotfix` (or `patch` if we want or anything that makes sense, as long as it's the same :) )

doc/RELEASE_GUIDE.md
cammellos: What if we just create a script that cut the release branch and update version?

`scripts/cut-release-branch.sh`

and it:
1) Creates a branch `release/x.y.z`
2) Bump version to `x.y.z`
3) Creates a PR in status-react with the updated version?

That should be easy (maybe 3 requires some manual intervention, not sure)

doc/RELEASE_GUIDE.md
cammellos: sorry, missed your last comment before replying, yes, that's the main issue

doc/RELEASE_GUIDE.md
jakubgs: I like that.

jakubgs: This way we avoid a lot of manual mistakes that can happen and make sure all the three steps are always taken.

jakubgs: Do you want to make it or should I?

cammellos: I can make it!

jakubgs: Maybe to not confuse actual release candidates and nightlies/dev builds the `develop` version should use a `-dev` prefix. So when we are working on `release/1.13.x` then `develop` is at `1.14.0-dev`.

If you wanted to be really fancy the `release/1.13.x` branch should start out as `1.13.0-rc` or `1.13.0-rc.0` and then switch it to `1.13.0` proper once we consider it stable, and then release. But that kinda affects the test flow, where the final build is different from the one we tested and consider stable. But then again the only change would be the version, so it should be sable in the same way.

cammellos: I wrote the scripts and updated the docs, would you mind having a look?

scripts/cut-release.sh
```diff
@@ -0,0 +1,77 @@
...
+do
+  case $Option in
+    M ) major=true;;
+    m ) minor=true;;
+  esac
+done
+
+shift $(($OPTIND - 1))
+
+version=$(cat VERSION)
```
jakubgs, line 30: Why are you reading `VERSION` before you fetch newest changes in `develop`?

scripts/cut-release.sh
```diff
@@ -0,0 +1,77 @@
...
+
+if [ ! -z $minor ]
+then
+  ((a[1]++))
+  a[2]=0
+fi
+
+newversion=${a[0]}.${a[1]}.${a[2]}
+releasebranch="${a[0]}.${a[1]}.x"
+echo "Cutting release branch $newversion"
```
jakubgs, line 65: This `newversion` has two issues:

* Logically `develop` doesn't represent the current release version, but the development version of the next release.
* `release/A.B.x` starts with `A.B.0`, but might be bumped to `A.B.1` and further, but `develop` stays as `A.B.0`.

Although it is true that it would complicate this a bit, since you would need to do two commits.
But I don't think that's that bad, and once it's implemented in the script it should work fine.

jakubgs: Since there is not direct gain from this other than logical/syntactical correctness it's not strictly necessary, but I'd expect this or something like this to become necessary in the future. But I'm fine with it as it is right now.

scripts/cut-release.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+# Increment a version string using Semantic Versioning (SemVer) terminology.
+
+# Parse command line options.
+
+set -e
+
+git diff-index --quiet HEAD
```
jakubgs: More like:
```sh
if ! git diff-index --quiet HEAD; then
  echo "Your Git working directory is not clean!" >&2
  exit 1
fi
```

jakubgs (approved):
LGTM other than the working directory check.

jakubgs (approved):

Samyoul (approved):
I found this a very helpful guide thank you

PR #11931, opened 2021-03-25
add creation of backup keycards

### LINKED ISSUE #11394: allow a user to have a keycard back up card
as a user, I have a keycard multiaccount on my phone, and I'd like to have a second (backup) card in case I lose mine. 

opportunity: 
allow the user to have a back up card
sell card bundles 

problem:
- today the user cannot create a second card with the same mnemonic with the same phone that has already the multiaccount on it 
- if the user manages to create such a card (e.g with a second phone) he cannot use it on the first phone with the multiaccount (because of the way the UID is used)

solution: 
need to build it 

cc @gravityblast @bitgamma

bitgamma:
closes #11394 

From a Keycard account, in the Settings -> Keycard menu a "Create backup Keycard" item has been added. Tapping it asks for the seed phrase (must match the one of the account) and will start the same flow as when creating a new Keycard account. An empty (initialized or not) Keycard is required. At the end of the flow the card will have the same exact keys as the original one and can be used to login, sign, etc. This can be repeated with any number of cards.

status: ready

| |


src/status_im/keycard/backup_key.cljs
```diff
@@ -0,0 +1,36 @@
...
+  (log/debug "[keycard] start backup")
+  (fx/merge cofx
+            {:db (-> db
+                     (assoc-in [:keycard :creating-backup?] true))}
+            (navigation/navigate-to-cofx :seed-phrase nil)))
+
+(fx/defn start-keycard-backup
+  {:events [::start-keycard-backup]}
+  [{:keys [db] :as cofx}]
+  {::multiaccounts.recover/import-multiaccount {:passphrase (get-in db [:multiaccounts/key-storage :seed-phrase])
```
shivekkhurana: The `:passphrase` should be sanitised before sending to import [[example](https://github.com/status-im/status-react/blob/0c55f1208b7eb42f0cb2d77d19caa336f325bcc6/src/status_im/multiaccounts/recover/core.cljs#L165)].  This was a bug in the migration to Keycard PR as well.

shivekkhurana (changes requested):

src/status_im/keycard/backup_key.cljs
bitgamma: Oh, thanks for catching this! fixed

shivekkhurana (approved):
Nice work :)

qoqobolo:
@bitgamma thanks for awesome work!

Only one issue/question: Is it expected that the margins between these lines are different?

Screenshot 2021-04-02 at 13 29 19

_ _ _

And could you add accessibility ID for this element please?

Screenshot 2021-04-02 at 13 19 15

qoqobolo:
Tested:

- can recover multiaccount, create backup Keycard, can log in with it the same or another device, all funds are shown in balance, can send/receive txs
- generate new acc, add watch-only
- creating several 'backup Keycards' for one multiaccount
- primary or secondary KC is frozen/blocked > can log in with ‘backup Keycard’
- can log in on the same device with both cards
- can back up new and restored multiaccounts 
- can restore the same account on two devices with two Keycards and pair primary KC with secondary device (use as backup)
- delete multiaccount and reset KC
- can migrate to KC and create backup
- tx history is saved on the new KC
- change PIN
- dark mode
- mobile network
- navigation
- interrupted flows
- after logout data is saved

bitgamma:
@qoqobolo thanks for the in-depth testing! I fixed the margin issue and added the accessibility id

PR #11934, opened 2021-03-29
navigate to login when error

flexsurfer:
navigate to login when error

### DIFF
--- a/src/status_im/signals/core.cljs
+++ b/src/status_im/signals/core.cljs
@@ -16,20 +16,22 @@
   (log/debug "[signals] status-node-started"
              "error" error)
   (if error
-    {:db (-> db
-             (update :multiaccounts/login dissoc :processing)
-             (assoc-in [:multiaccounts/login :error]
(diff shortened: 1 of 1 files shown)


src/status_im/signals/core.cljs
```diff
@@ -16,20 +16,22 @@
...
-                       ;; when the user inputs the wrong password
-                       ;; if that is the error that is found
-                       ;; we show the i18n label for wrong password
-                       ;; to the user
-                       ;; in case of an unknown error we show the
-                       ;; error
-                       (if (= error "file is not a database")
-                         (i18n/label :t/wrong-password)
-                         error)))}
+    (merge {:db (-> db
```
cammellos: I would use
 ```
(cond-> {:db ...} 
   (= (:view-id db) ....)
   (assoc :dispatch [:navigate-to :login]))
```
It's probably a bit terser and should be slightly more efficient

cammellos (approved):

PR #11935, opened 2021-03-29
Move react-native-status module to a local CocoaPods dependency

briansztamfater:
### Summary

I had some problems with `react-native-status` module linking on Xcode 12 / M1 Macbook / macOS Big Sur. It compiled but the NativeModule was not recognized on the React Native side, making the app to crash as soon as it started. Moving `react-native-status` from Manual Linking to a pod dependency solved the problem, and I thought it might be better for building and compatibility to let CocoaPods manage the library linking, so pushing this PR as a proposal to move it to a pod dependency.

### Review notes

- Check if package versions modifications in `Podfile.lock` file are OK 
- Check iOS compiling, probably by checking out the code and manually trying it

### Testing notes

- A smoke testing should be fine as nothing logic-related was changed.

#### Platforms

- iOS

status: ready

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -20,6 +20,7 @@ abstract_target 'Status' do
   pod 'Permission-Camera', :path => "#{permissions_path}/Camera.podspec"
 
   pod "react-native-status-keycard", path: "../node_modules/react-native-status-keycard"
+  pod "react-native-status", path: "../modules/react-native-status"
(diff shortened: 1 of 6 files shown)


flexsurfer (approved):
nice!

shivekkhurana (approved):

flexsurfer:
merged, for some reason PR wasn't autoclosed

PR #11937, opened 2021-03-30
🔄 Pull to refresh on wallet screen

### LINKED ISSUE #11849: Move pull to refresh to the main account to make it more easily discoverable
# Problem
See also #11821 

Currently, following #11723 the account balance is not automatically updated for newly generated accounts and imported accounts with zero value. The user can pull to refresh to start fetching balance. However, pull to refresh is currently only available on individual account screens, not on the account overview screen. 

Especially for users with only one account, this can be difficult to discover, resulting in the user not seeing updates or knowing how to update their balance after sending funds to their empty account.

## Implementation
tbd

## Acceptance Criteria
Pull to refresh is available on account overview screen

## Future Steps
#11821

shivekkhurana:
fixes #11849

### Summary
In the current implementation, we have pull-to-refresh action on individual accounts page. This action restarts the wallet service.

This PR replicates the same action on the accounts overview screen.

### Testing notes
Pull-to-refresh is known to behave differently on Android on Ios. We should test both to be sure.

#### Platforms
- Android
- iOS

- Wallet

##### Functional
- wallet / transactions

### Steps to test
- Goto wallet screen, pull down to refresh, all account balances should update
- The same should happen if you are on a specific account screen and you pull-to-refresh

status: ready

### DIFF
--- a/src/quo/components/animated_header.cljs
+++ b/src/quo/components/animated_header.cljs
@@ -44,7 +44,7 @@
         offset          (reagent/atom 0)
         on-layout       (fn [evt]
                           (reset! offset (oget evt "nativeEvent" "layout" "height")))]
-    (fn [{:keys [extended-header] :as props} children]
+    (fn [{:keys [extended-header refresh-control] :as props} children]
       [animated/view {:flex           1
                       :pointer-events :box-none}
        [animated/code {:key  (str @offset)
@@ -68,7 +68,8 @@
                         (dissoc props :extended-header))]]
        (into [animated/scroll-view {:on-scroll           on-scroll
                                     :style               {:z-index 1}
-                                    :scrollEventThrottle 16}
+                                    :scrollEventThrottle 16
+                                    :refreshControl      refresh-control}
               [animated/view {:pointer-events :box-none}
                [animated/view {:pointer-events :box-none
                                :on-layout      on-layout}
--- a/src/status_im/ui/screens/multiaccounts/login/views.cljs
+++ b/src/status_im/ui/screens/multiaccounts/login/views.cljs
@@ -107,3 +107,10 @@
          {:disabled (or (not sign-in-enabled?) processing)
           :on-press #(login-multiaccount @password-text-input)}
          (i18n/label :t/sign-in)]]}]]))
+
+(comment
+  ;; fill password and login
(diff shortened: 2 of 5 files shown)


qoqobolo:
@shivekkhurana 

### ISSUE 1: Spinner loops while searching for a connection on iOS

If use `pull-to-refresh` while searching for a connection, e.g. immediately after switching from Wi-Fi to LTE (works only if the connection takes some time), or just offline (reproducible every time), the spinner loops, as shown in the video below.

#### Expected behavior

No looping, just showing spinner and 'no connection' message probably 

#### Actual behavior

https://user-images.githubusercontent.com/67952253/113135785-d4b35d00-9222-11eb-9fb2-8385372eab68.mp4

This behavior isn't reproducible on Nightly.

#### Steps:
- Open Wallet
- Turn on airplane mode
- `Pull-to-refresh`

shivekkhurana:
@qoqobolo I was able to reproduce this on Android too. But the bug seems to be incidental, and not because of the feature.

Can you please check if you see this message on any of your accounts (on the account detail page)?

qoqobolo:
@shivekkhurana nope, I don't see it. Checked on three devices (Android and iOS) with three accounts, on WiFi and cellular network.

shivekkhurana:
@qoqobolo Thanks for checking. Turns out that error was an issue on my local device due to a bad migration.

I'm not 100% sure, but the problem was that the root wallet component was subscribing to the re-frame atom directly, instead of using the `defview` macro. Hence, it updated each time the global `app-db` changed. When the component updated, the spinner was re-rendered. It would not be re-rendered if the component doesn't update unnecessarily. 

I have pushed a fix, let me know if this works.

qoqobolo:
@shivekkhurana can still be replicated on a cellular network :(

https://user-images.githubusercontent.com/67952253/113677853-504b5900-96be-11eb-83c7-16e43911ba92.mp4

flexsurfer:
@shivekkhurana any updates on this ?

shivekkhurana:
@flexsurfer Not yet. The metrics feature has consumed all my time.

flexsurfer:
hey @shivekkhurana is this still relevant?

PR #11939, opened 2021-03-30
🧽 🚿 Add missing seed phrase sanitation to key storage flow 

shivekkhurana:
### Summary

In key management section, the mnemonic was not being trimmed, which lead to issues. The issue was pointed out by @Serhy in https://github.com/status-im/status-react/issues/11735#issuecomment-782066601 .

This PR adds the missing seed sanitation.

#### Platforms
- Android
- iOS

- Key transfer to Keycard

### Steps to test
- Open Status
- Click on an on-device multi account
- Click on the context menu (three dots on the right-top)
- Chose "Manage keys and storage"
- Check "Move keystore file"
- Click "Enter seed phrase"
- Enter seed phrase of the selected account with extra spaces at the end / in between words
- You should be able to proceed as long as the words are correct

status: ready

### DIFF
--- a/src/status_im/multiaccounts/key_storage/core.cljs
+++ b/src/status_im/multiaccounts/key_storage/core.cljs
@@ -66,7 +66,8 @@
   "Check if the key-uid was generated with the given seed-phrase"
   [{:keys [import-mnemonic-fn on-success on-error]} {:keys [seed-phrase key-uid]}]
   (import-mnemonic-fn
-   seed-phrase nil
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

Serhy:
Tested with Android and iOS, - seed phrase sanitising works as expected. However, it's not fixing https://github.com/status-im/status-react/issues/11735 issue. But anyway it's improvement so let's merge. Thanks @shivekkhurana

PR #11940, opened 2021-03-30
🎶 message_sent.mp3

shivekkhurana:
### Summary
Starting point for the issue https://github.com/status-im/status-react/issues/10468. Plays a sound when a message is sent. 

This PR adds a re-frame effect to play sounds. The sounds need to be imported differently for Android and iOs [[reference](https://github.com/react-native-audio-toolkit/react-native-audio-toolkit/blob/0a2e315d168df1cb02e93d19b98fff5106976882/docs/SOURCES.md)]. As per the docs of the library we are using, mp3 format is preferred.
...

### Review notes
I got rid of the `.-canPlay` check in `start-playing` because due to an issue in the library, [`(.-canPlay player)` is always false](https://github.com/react-native-audio-toolkit/react-native-audio-toolkit/issues/102).

### Testing notes
Please check that the sound on sending messages is not too high (or low).

#### Platforms
- Android
- iOS

- 1-1 chats
- Public chats
- Group chats

##### Functional
- 1-1 chats
- public chats
- group chats

### Steps to test
- Send a message as usual
- You should here a sound

status: ready


flexsurfer:
@shivekkhurana still relevant?

shivekkhurana:
@flexsurfer This feature needs some optimization, so I kept the branch open. Happy to close PR without deleting the branch. My plan is to circle back after metrics get merged.

shivekkhurana:
TL;DR; It's hard to get the sound to play in sync with rendering.

I have tried two approaches with building chat sounds:
1. **Dispatch chat sound as a re-frame event while sending message**
This approach is clean but the sound playback is not consistent with message rendering. Most of the times, the sound plays in sync with render, but sometimes it's off by a few milliseconds. This happens more often on low-end devices.

2. **Play chat sound on message mount**
This approach is heavy in terms of resources. Messages are already complex. This chat sound in `:component-did-mount` adds another layer of complexity. The chat was noticeably slow with sounds playing in `:component-did-mount` in conjunction with `:dispatch-sync :audio/play`. The sound playback was consistent with rendering for text messages, but not for stickers. 

I'm going to close this because it's hard to fine-tune perf issues.

PR #11942, opened 2021-03-30
Improved Start New Private Chat interface

### LINKED ISSUE #11675: Improved `Start New Private Chat` interface
# Problem
The scope of this issue is to implement the redesigned changes to the Start New Chat interface. It's an improvement over the current one in clarity and use experience.

## Implementation
Frame 50

:art: [Figma can be found here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=7957%3A169)

* [x] Change the input, to the one in the design. It's full width and doesn't have a confirm button anymore, placeholder text was also changed. Typing in the input will parse the strict and show relevant suggestions below it, tapping on one of them will start a new chat.
* [x] Remove `Add contacts` button from the list
* [x] Include all of the states for the input: when no contacts match, when no non-contacts match, when it's clearly misspelled

## Acceptance Criteria
The entire redesign was implemented, tested and no issues were identified.

## Notes
This follows what desktop did with their own UI, their issue can be found [here](https://github.com/status-im/status-desktop/issues/1747)

Thanks!

briansztamfater:
fixes: #11675

### Summary

This PR solves #11675 issue which proposes UI/UX improvements when starting a new private chat.

There are new strings that would need to be translated @jinhojang6 

##### Demo

start-chat-improvments

##### Screenshots (iOS)

empty-ios not-empty-ios not-empty-search-ios

##### Screenshots (Android)

empty-android not-empty-android not-empty-search-android

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats

### Steps to test

- Open Status
- Go to Chat section
- Tap on + button
- Tap on "Start a new chat"

status: ready

### DIFF
--- a/src/quo/components/list/item.cljs
+++ b/src/quo/components/list/item.cljs
@@ -83,7 +83,7 @@
           [icons/icon icon {:color icon-color}]])])))
 
 (defn title-column
-  [{:keys [title text-color subtitle subtitle-max-lines
+  [{:keys [title text-color subtitle subtitle-max-lines subtitle-secondary
            title-accessibility-label size text-size title-text-weight
            right-side-present?]}]
   [rn/view {:style (merge (:tiny spacing/padding-horizontal)
@@ -105,14 +105,37 @@
                     :size                text-size}
          title]
         title)
-      (if (string? subtitle)
-        [text/text {:weight          :regular
-                    :color           :secondary
-                    :ellipsize-mode  :tail
-                    :number-of-lines subtitle-max-lines
-                    :size            text-size}
-         subtitle]
-        subtitle)]
+      (if (string? subtitle-secondary)
+        [rn/view {:flex-direction :row}
+         [text/text {:style {:max-width "56.5%"}
+                     :weight          :regular
+                     :color           :secondary
+                     :ellipsize-mode  :tail
+                     :number-of-lines subtitle-max-lines
+                     :size            text-size}
+          subtitle]
+         [text/text {:style {:width "7%" :text-align :center}
+                     :weight          :regular
+                     :color           :secondary
+                     :ellipsize-mode  :middle
+                     :number-of-lines subtitle-max-lines
+                     :size            text-size}
+          "•"]
+         [text/text {:style {:max-width "36.5%"}
+                     :weight          :regular
+                     :color           :secondary
+                     :ellipsize-mode  :middle
+                     :number-of-lines subtitle-max-lines
+                     :size            text-size}
+          subtitle-secondary]]
+        (if (string? subtitle)
+          [text/text {:weight          :regular
+                      :color           :secondary
+                      :ellipsize-mode  :tail
+                      :number-of-lines subtitle-max-lines
+                      :size            text-size}
+           subtitle]
+          subtitle))]
 
      title
      (if (string? title)
@@ -170,7 +193,7 @@
 (defn list-item
   [{:keys [theme accessory disabled subtitle-max-lines icon icon-container-style
            left-side-alignment
-           title subtitle active on-press on-long-press chevron size text-size
+           title subtitle subtitle-secondary active on-press on-long-press chevron size text-size
            accessory-text accessibility-label title-accessibility-label
            haptic-feedback haptic-type error animated animated-accessory? title-text-weight]
     :or   {subtitle-max-lines 1
@@ -223,6 +246,7 @@
                    :text-size                 text-size
                    :subtitle                  subtitle
                    :subtitle-max-lines        subtitle-max-lines
+                   :subtitle-secondary        subtitle-secondary
                    :right-side-present?       (or accessory chevron)}]
        [right-side {:chevron             chevron
                     :active              active
--- a/src/status_im/ui/screens/add_new/new_chat/views.cljs
+++ b/src/status_im/ui/screens/add_new/new_chat/views.cljs
@@ -13,17 +13,22 @@
             [status-im.ui.components.topbar :as topbar]
             [status-im.utils.debounce :as debounce]
             [status-im.utils.utils :as utils]
-            [reagent.core :as reagent])
+            [reagent.core :as reagent]
+            [quo.react-native :as rn]
+            [clojure.string :as string]
+            [status-im.ui.components.invite.views :as invite]
+            [status-im.ethereum.ens :as ens]
+            [quo.platform :as platform]
+            [status-im.transport.filters.core :as filters]
+            [status-im.utils.identicon :as identicon])
   (:require-macros [status-im.utils.views :as views]))
 
 (defn- render-row [row]
-  (let [[first-name second-name] (multiaccounts/contact-two-names row false)]
+  (let [first-name (first (multiaccounts/contact-two-names row false))]
     [quo/list-item
      {:title    first-name
-      :subtitle second-name
       :icon     [chat-icon/contact-icon-contacts-tab
                  (multiaccounts/displayed-photo row)]
-      :chevron  true
       :on-press #(re-frame/dispatch [:chat.ui/start-chat
                                      (:public-key row)])}]))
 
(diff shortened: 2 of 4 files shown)


translations/en.json
```diff
@@ -1330,7 +1332,7 @@
     "add-seed-account": "Add account with a seed phrase",
     "account-exists-title": "Account already exists",
     "add-private-key-account": "Add account from private key",
-    "user-not-found": "User not found",
+    "user-not-found": "Profile not found",
```
briansztamfater: I want also to change the key to `profile-not-found`, but I would also need to change the key on all the translation files. Let me know if I can proceed with that!

src/status_im/ui/screens/add_new/new_chat/views.cljs
```diff
@@ -61,9 +64,29 @@
...
+                             (string/includes? (string/lower-case (str alias)) lower-filter-text)
+                             (when nickname
+                               (string/includes? (string/lower-case (str nickname)) lower-filter-text))))]
+    (if filter-text
+      (filter filter-fn contacts)
+      contacts)))
+
+(defn is-valid-username? [username]
+  (let [is-chat-key? (and (string? username)
+                          (string/starts-with? username "0x")
```
rasom: `(string/starts-with? username "0x")` we probably already have some helper function for this

rasom (approved):

src/quo/components/list/item.cljs
```diff
@@ -105,14 +105,37 @@
...
-        [text/text {:weight          :regular
-                    :color           :secondary
-                    :ellipsize-mode  :tail
-                    :number-of-lines subtitle-max-lines
-                    :size            text-size}
-         subtitle]
-        subtitle)]
+      (if (string? subtitle-secondary)
+        [rn/view {:flex-direction :row}
+         [text/text {:style {:max-width "56.5%"}
```
flexsurfer, line 110: could you please elaborate on the width value?

flexsurfer (approved):

translations/en.json
shivekkhurana: You just need to change the `en.json` file. We use [Lokalise](https://lokalise.com/) to manage translations. 

When you update an existing translation, just inform @jinhojang6 about it.

src/status_im/ui/screens/add_new/new_chat/views.cljs
```diff
@@ -61,9 +64,29 @@
     :yourself
     (i18n/label :t/can-not-add-yourself)))
 
+(defn filter-contacts [filter-text contacts]
+  (let [lower-filter-text (string/lower-case (str filter-text))
+        filter-fn         (fn [{:keys [name alias nickname]}]
```
shivekkhurana: This function will be redefined each time the search value changes:
- You can extract `filter-fn` to be a standalone function
- Remove the `if` conditional or
- Move the `if` at the point of use : `(if @search-value (filter-contacts ...) [])`

shivekkhurana (changes requested):
Left a few comments. Also what's up with the magic numbers for width?

translations/en.json
shivekkhurana: Just rename it on `en.json` and inform @jinhojang6 about the change.

hesterbruikman:
@briansztamfater @errorists Small copy change suggestion
`0x0` to `0x04` I don't think it has major impact, but the prefix is more accurate and at least will tell avid crypto users that their chat key is not a regular wallet public key

(until we would use compresses keys that is, but in that case 0x0 would also change)

src/quo/components/list/item.cljs
briansztamfater: Sure. Those are just to comply with the designs and assigning a maximum width for displaying long texts as the GfyCat Name and the chat key in the same line, with ellipsis rules and separated by a `•` char. The separator text width is fixed, so with the left and right texts max-width added equals to the 100% of the view's width. I can make these values settable if that makes more sense. Example below:



Left text is a bit more larger so that's why there are some diffs in values.

src/status_im/ui/screens/add_new/new_chat/views.cljs
briansztamfater: Good catch! Moving the function to a standalone one.

src/status_im/ui/screens/add_new/new_chat/views.cljs
briansztamfater: Found one in `status-im.transport.filters.core` namespace, thanks!

briansztamfater:
> @briansztamfater @errorists Small copy change suggestion
> `0x0` to `0x04` I don't think it has major impact, but the prefix is more accurate and at least will tell avid crypto users that their chat key is not a regular wallet public key
> 
> (until we would use compresses keys that is, but in that case 0x0 would also change)

Good catch, I agree with the change suggestion. Just pushed the change.

src/status_im/ui/screens/add_new/new_chat/views.cljs
```diff
@@ -57,13 +61,33 @@
...
   (case value
     :invalid
-    (i18n/label :t/user-not-found)
+    (i18n/label :t/profile-not-found)
     :yourself
     (i18n/label :t/can-not-add-yourself)))
 
+(defn search-contacts [filter-text {:keys [name alias nickname]}]
+  (or
+   (string/includes? (string/lower-case (str name)) filter-text)
```
shivekkhurana, line 71: Why are you type casting manually `(str name)`? Is there any problem with `(string/lower-case name)`?

shivekkhurana (approved):

Serhy:
Thanks @briansztamfater! Few items to address: 

1) Not sure it's supposed to be implemented, but - there is no highlighting of characters which matches the username from 'Contacts' (see below in video below)

2) When search by public key - username is not visible. It opens correct account but just username never shown.

https://user-images.githubusercontent.com/8749671/113729630-db990e80-96ff-11eb-8a89-ba519563ad0c.mov

3) Disregard the searched username, - it's user_icon is always the same

Screenshot 2021-04-06 at 17 49 14

briansztamfater:
Good catch on the issues, @Serhy!

Regarding the first point you mention about highlighting characters, we would need to do a big refactor to the component and discuss the proper approach to implement it, so we talked that it would be best to tackle that later in another PR as an enhancement if needed.

BTW, I will check the other points and come back to you if I have any problems reproducing them. Thanks!

briansztamfater:
@Serhy So regarding the point 2, it was a bug regarding that ENS name is not resolved when searching by public key. Because of that technical limitation I needed to change how the row is displayed when that's the case. I ended up mirroring the same implementation as the desktop version: showing the generated name as the title and just the shortened public key as the subtitle.

The desktop version works like this:
Desktop implementation

Anyway, fixes are already pushed and ready to test.

Serhy:
Thanks @briansztamfater 
All issues fixed (tested on iOS14 and Android 11)
One regression noticed though (sorry haven't noticed it in previous build, - before issue fixes but it's there too)

When user added in accounts, - after app restart, when the 'Start 1-1 chat' screen opened - app persistently crashing with error(text snippet from logcat):
```
04-07 16:30:35.921 20009 20987 I ReactNativeJS: ***
04-07 16:30:35.921 20009 20987 I ReactNativeJS: NAME: "TypeError"
04-07 16:30:35.921 20009 20987 I ReactNativeJS: MESSAGE: Cannot read property 'toLowerCase' of null
04-07 16:30:35.921 20009 20987 I ReactNativeJS: 
04-07 16:30:35.921 20009 20987 I ReactNativeJS: TypeError: Cannot read property 'toLowerCase' of null
04-07 16:30:35.921 20009 20987 I ReactNativeJS:     at $c (/build/android/app/build/generated/assets/react/pr/index.android.bundle:7:70948)
04-07 16:30:35.921 20009 20987 I ReactNativeJS:     at NK
``` 
It's all fine when navigating to edited screen after account added in the same app session, but crash 'reveal' itself after app re-open. 
**To reproduce:**
1) Create new account
2) Add an account (via ENS name or public key)
3) Re-open Status
4) Open `(+)` -> `Start new chat` view

**Actual result:** 
Screenshot 2021-04-07 at 16 53 27

briansztamfater:
Hey @Serhy! I'm having trouble to reproduce the regression, could you expand the steps to reproduce with a video recording? Does this happen on iOS too?

briansztamfater:
@Serhy nevermind, just managed to reproduce it, so ignore my previous message. Will work on a fix and let you know. Thanks!

src/status_im/ui/screens/add_new/new_chat/views.cljs
briansztamfater: So I found out that casting serves as a null safety mechanism because `(str nil)` returns an empty string. Removing it causes a regression bug when adding contacts by public key that have not a name or alias. Will re-add it and let me know if there are any other concerns.

briansztamfater:
@Serhy regression bug should be fixed, let me know if you find something else.

Serhy:
@briansztamfater thanks for fixes! All issues fixed. Please mind my latest commit before merge.

errorists:
great job @briansztamfater one small thing, can we display the text in the input using the Inter Mono font, our version of Inter typeface with monospaced digits? it's better for addresses. it's not obvious, lmk if you need help with that

briansztamfater:
@errorists thanks, updated PR with monospaced digits for the text input :)

errorists (approved):

PR #11952, opened 2021-04-02
Pinned messages in chats

### LINKED ISSUE #11699: Pinned messages in chats
As a user
I want to be able to pin messages
So I can welcome new joiners with important information
So I can keep important messages visible

### Description

We want to allow users in a group chat to pin messages.
This should work for private group chats, one-to-ones and community chats.
Public chats are excluded.

The way it works is that pinned messages will be ordered just like normal messages, but will have a different visual 
indicator (i.e color and a label) https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=18553%3A60460

For each chat type, there will be an entry point to see the pinned messages:

For one to ones, is under the profile of a user (NOTE: profile always shows the pinned message between you and the other user in a one to one chat, regardless of where you access the profile from)

For private group chats is under group info: https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=6942%3A35457

For communities chat is under the chat details screen (not existing at the moment): https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=18729%3A63242

The way we display pinned messages once clicked is the same:

https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=6945%3A0

### Acceptance criteria

- Messages can be pinned in any relevant chat according to the rules below
- Messages can be unpinned in any relevant chat according to the rules below
- I can see the pinned messages in each subsection

#### Rules

In a private group chat, any admin can pin a message.
In a one-to-one chat any participant can pin a message.
In a community chat, any admin can pin a message.

### Implementation
Feel free to ping me if you'd like to discuss potential protocol changes.

### LINKED ISSUE #12217: If user with a long 3-random name pins a message it overflows screen edge
# Bug Report

## Problem 
Long 3-random names in pinned messages are not split into lines in the chat view and on the pinned messages screen. That's visible on a small screen, e.g iPhone SE. 

#### Expected behavior
Wrap long name to another line

#### Actual behavior
Screenshot 2021-06-10 at 11 44 37

### Reproduction

1. User A: start 1-1 chat with User B who has a long 3-random name
2.  Send a message
3. User B: pin this message

### Additional Information

- Status version: #11952 

briansztamfater:
fixes #11699 
fixes #12217

### Summary

Implementing pin message feature. User should be able to pin / unpin a message and see pinned messages on user profile (for 1-to-1 chats) and groups.

### Review notes
- Signals not implemented right now, waiting for status-go part and then will add it.
- UI may be not accurate to some designs, adapted to API (status-go does not send who pinned a message) and tap-and-hold approach for pinning

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats
- public chats
- group chats

#### Scenario 1
```
- Open Status
- Open 1-to-1 chat
- Tap and hold on message bubble
- Select Pin
- View message marked as pinned and also on pinned messages screen from profile screen
- Verify that pin status of a message is retrieved when closing and reopening the app
```
#### Scenario 2
```
- Open Status
- Open any group chat
- Tap and hold on message bubble
- Select Pin
- View message marked as pinned and also on pinned messages screen from Group Info
- Verify that pin status of a message is retrieved when closing and reopening the app

```
status: ready


src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+(fx/defn pin-messages-loaded
+  {:events [::pin-messages-loaded]}
+  [{db :db} chat-id session-id {:keys [cursor messages]}]
+  (when-not (and (get-in db [:pagination-info chat-id :pin-messages-initialized?])
+                 (not= session-id
+                       (get-in db [:pagination-info chat-id :pin-messages-initialized?])))
+    (let [pin-messages (map #(assoc-in % [:pinned?] true)
+                            messages)
+          already-loaded-pin-messages (get-in db [:pin-messages chat-id] {})
+          {:keys [all-messages new-messages]} (reduce (fn [{:keys [all-messages] :as acc}
```
shivekkhurana: What are we trying to achieve here? Can you please document it here? 
Maybe convert the inline `fn` to a standalone `fn`?

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+                                                          (update :all-messages assoc message-id message)))
+                                                      {:all-messages already-loaded-pin-messages
+                                                       :new-messages []}
+                                                      pin-messages)
+          messages-id-list (map :message-id pin-messages)
+          current-clock-value (get-in db [:pagination-info chat-id :pin-cursor-clock-value])
+          clock-value (when cursor (loading/cursor->clock-value cursor))]
+      {:db (-> db
+               ((fn [db]
+                  (reduce (fn [acc message-id]
```
shivekkhurana: This logic is complex. Do we really need to add the `pinned?` flag to each message? 

Is there anything stopping us from saving `pinned-messag-ids` in db and creating a subscription over it to manage the UI ?

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+               (assoc-in [:pagination-info chat-id :all-pin-loaded?]
+                         (empty? cursor)))})))
+
+(fx/defn receive-signal
+  [{:keys [db] :as cofx} pin-messages]
+  (let [{:keys [chat-id message_id pinned]} (first pin-messages)
+        pinned? (or pinned false)
+        pin-message (get-in db [:messages chat-id message_id])
+        message (merge pin-message {:pinned? pinned?})]
+    (println (:current-chat-id db) chat-id "DB CRREDS")
```
shivekkhurana: Rogue `println`

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+    (if pinned?
+      {:db (as-> db $
```
shivekkhurana: `as->` macro is redundant, can use `db` directly in place of `$`

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+               (-> $
+                   (assoc-in [:messages chat-id message_id :pinned?] pinned?)
+                   (update-in [:pin-message-lists chat-id] message-list/remove-message pin-message)
+                   (update-in [:pin-messages chat-id] dissoc message_id))
+               (update-in $ [:pin-messages chat-id] dissoc message_id)))})))
+
+(fx/defn load-more-pin-messages
+  {:events [:load-more-pin-messages]}
+  [{:keys [db]} chat-id first-request]
+  (when-let [session-id (get-in db [:pagination-info chat-id :pin-messages-initialized?])]
```
shivekkhurana: You can simplify the logic by bundling `when`s and `let`s:

```clojure
(fx/defn load-more-pin-messages
  {:events [:load-more-pin-messages]}
  [{:keys [db]} chat-id first-request]
  (let [session-id (get-in db [:pagination-info chat-id :pin-messages-initialized?])
        not-all-pin-loaded? (not (get-in db [:pagination-info chat-id :all-pin-loaded?]))
        not-loading-pin-messages? (not (get-in db [:pagination-info chat-id :loading-pin-messages?]))
        cursor (get-in db [:pagination-info chat-id :pin-cursor])]
    (when (and session-id
               not-all-pin-loaded?
               not-loading-pin-messages
               (or cursor first-request))
      (fx/merge
       {:db (assoc-in db [:pagination-info chat-id :loading-pin-messages?] true)}
       (data-store.messages/pinned-message-by-chat-id-rpc
        chat-id
        cursor
        constants/default-number-of-pin-messages
        #(re-frame/dispatch [::pin-messages-loaded chat-id session-id %])
        #(re-frame/dispatch [::failed-loading-pin-messages chat-id session-id %]))))))
  ```

File: src/status_im/chat/models/pin_message.cljs [hunk 6]
```diff
@@ -0,0 +1,135 @@
...
+    (when (and
+           (not (get-in db [:pagination-info chat-id :all-pin-loaded?]))
+           (not (get-in db [:pagination-info chat-id :loading-pin-messages?])))
+      (let [cursor (get-in db [:pagination-info chat-id :pin-cursor])]
+        (when (or first-request cursor)
+          (merge
```
shivekkhurana: This should be `fx/merge`

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,135 @@
...
+            constants/default-number-of-pin-messages
+            #(re-frame/dispatch [::pin-messages-loaded chat-id session-id %])
+            #(re-frame/dispatch [::failed-loading-pin-messages chat-id session-id %]))))))))
+
+(fx/defn send-pin-message
+  "Pin message, rebuild pinned messages list"
+  {:events [::send-pin-message]}
+  [{:keys [db] :as cofx} {:keys [chat-id message-id pinned?] :as pin-message}]
+  (fx/merge cofx
+            {:db (as-> db $
```
shivekkhurana: Again redundant `as->`

src/status_im/data_store/messages.cljs
```diff
@@ -10,9 +10,12 @@
...
            :sticker (:sticker content))
     :always
-    (clojure.set/rename-keys {:chat-id :chatId
+    (clojure.set/rename-keys {:chat-id :chat_id
+                              :message-id        :message_id
+                              :whisper-timestamp :whisperTimestamp
                               :community-id :communityId
-                              :clock-value :clock})))
+                              :clock-value :clock
+                              :pinned? :pinned})))
```
shivekkhurana: Can we do without renaming this key?

src/status_im/data_store/messages.cljs
```diff
@@ -103,3 +123,6 @@
 
 (fx/defn delete-messages-by-chat-id [cofx chat-id]
   (delete-messages-by-chat-id-rpc chat-id))
+
+(fx/defn send-pin-message [cofx pin-message]
```
shivekkhurana: You can move the definition inside `fx/defn`:

```clojure
(fx/defn send-pin-message [cofx pin-message]
  {::json-rpc/call [{:method (json-rpc/call-ext-method "sendPinMessage")
                     :params [(->rpc pin-message)]
                     :on-success #(log/debug "successfully pinned message" pin-message)
                     :on-failure #(log/error "failed to pin message" % pin-message)}]})

File: src/status_im/ui/screens/chat/message/message.cljs [hunk 10]
```diff
@@ -200,6 +201,20 @@
     [react/view style/not-sent-icon
      [icons/icon :main-icons/warning {:color colors/red}]]]])
 
+(defn pinned-by-indicator [outgoing display-photo?]
+  [react/view {:style (style/pin-indicator outgoing display-photo?)}
+   [react/view {:style {:flex-direction :row
+                        :align-items :center}}
+    [icons/icon :main-icons/pin {:color (if outgoing colors/white-persist colors/gray)
+                                 :height 15
```
shivekkhurana: magic numbers 🧙🏼‍♂️

src/status_im/ui/screens/chat/styles/main.cljs
```diff
@@ -6,6 +6,11 @@
    :align-items    :center
    :flex-direction :row})
 
+(def pins-name-view
```
shivekkhurana, line 9: Can align maps in columns.

src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -65,6 +65,39 @@
     {:align-self    :flex-start
      :padding-left  8}))
 
+(defn pin-indicator [outgoing display-photo?]
+  (merge
+   {:flex-direction :row
```
shivekkhurana: Can align maps in columns.

src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -133,10 +166,12 @@
      {:border-bottom-right-radius 4}
      {:border-bottom-left-radius 4})
 
-   (cond
-     (= content-type constants/content-type-system-text) nil
-     outgoing                                            {:background-color colors/blue}
-     :else                                               {:background-color colors/blue-light})
+   (if pinned?
```
shivekkhurana: You can add `pinned?` inside `cond`
```clojure
(cond
 ... ...
 pinned? {}
 :else)```

File: src/status_im/ui/screens/chat/toolbar_content.cljs [hunk 14]
```diff
@@ -53,3 +53,20 @@
 
 (defn toolbar-content-view []
   [toolbar-content-view-inner @(re-frame/subscribe [:chats/current-chat])])
+
+(defn toolbar-pin-content-view []
```
shivekkhurana: Use `defview` and `letsubs` instead of raw `re-frame/subscribe`

src/status_im/ui/screens/profile/group_chat/views.cljs
```diff
@@ -186,7 +186,8 @@
     (when current-chat
       (let [admin?                (get admins current-pk)
             allow-adding-members? (and admin? joined?
-                                       (< (count members) constants/max-group-chat-participants))]
+                                       (< (count members) constants/max-group-chat-participants))
+            pinned-messages @(re-frame/subscribe [:chats/pinned chat-id])]
```
shivekkhurana: Use `letsubs` over raw `subscribe`

shivekkhurana (changes requested):
Impressive work 👏🏼 . This PR is huge!

The overall logic around updating local db after fetching pinned messages can be simplified. Also left some suggestions around syntax and clojure api.

src/status_im/ui/screens/chat/toolbar_content.cljs
flexsurfer: honestly i don't think we have a strict rule ,  we need to discuss this, because there are also new `[status-im.utils.handlers :refer [>evt <sub]]`, i guess defview is outdated, because it was used before subscriptions is cashed, so probably we should move to `>evt <sub` the only thing i like in defview is map with component lifecycle handlers

src/status_im/chat/models/pin_message.cljs
briansztamfater: This was mostly grabbed from the messages-loaded function, basically it is a mecanism to load new pinned messages and remove duplicated messages

src/status_im/chat/models/pin_message.cljs
briansztamfater: I will explore that approach and see how it works, it impacts in so many places in the UI that decided to create this prop, but I'll try that

briansztamfater:
Moving to `E2E Tests` column to start getting feedback from QA while looking if we can optimize the `pinned?` prop logic, but the rest of the comments from code review should be fixed.

cammellos:
@briansztamfater what is the state of this PR? do you need some help rebasing/addressing feedback?

churik:
@briansztamfater 
Thank you for PR!

### ISSUE 1: can pin messages for public chats but no way to see them

**Steps:**
1) join any public chat
2) pin some messages

**Expected result:**
can see pinned messages in one place
**Actual result:**
no way to see pinned messages in one place
**OS:** IOS, Android
Maybe better to hide it for now, if no clarification is provided how it should work for public chats / communities? 
Also, I saw in community channel the we should limit amount of pinned messages.
If we plan to limit this for public channels, IMO better to do it here or to excuse messages for pin from public chats at all until it will be implemented.

cc @cammellos 

### ISSUE 2: Error when pinned message exists in unfetched (previous) history
Reproducible for public and group chats.

**Steps for public chat:**
1) User A: join any public chat with history, fetch more messages, pin some message from previous history
2) User B: join same public chat

**Steps for group chat:**
1) User A: create group chat with User B, send several messages
2) User B: join group chat
3) User A: pin messages that were sent upon joining User B

**Expected result:**
no error
**Actual result:**
IMAGE 2021-05-31 12:30:15

**OS:** IOS, Android

### ISSUE 3: Can't see pinned messages by another members until relogin 
Checked on 1-1 chat.
Issue is not reproducible after pinned messages from both users appeared on screen.

**Steps:**
1) User A: start chat with User B, pin some message 
2) User B: open 1-1 chat with User B > options > View profile > pinned messages
**Expected result:**
can see pinned messages
**Actual result:**
Monosnap 2021-05-31 12-34-28
**OS:** IOS, Android

### ISSUE 4: Not clear who pins messages in 1-1 chat
Monosnap 2021-05-31 12-45-31
@errorists maybe you have some ideas on how to improve visibility here


### ISSUE 5: After replying style of pinned message is reset until reopening chat view
It looks unpinned after sending reply until reopening chat view
**Steps:**
1) User A: start chat with User B, pin "message a"
2) User A: reply to  "message a"

https://user-images.githubusercontent.com/4557972/120182722-0b9bf400-c20f-11eb-9274-4b773cbbe566.mp4

### ISSUE 6: Not optimized for dark mode
IMAGE 2021-05-31 13:15:00
IMAGE 2021-05-31 13:15:03

### ISSUE 7: Gradient on long pinned messages
IMAGE 2021-05-31 15:16:20

### ISSUE 8: No option of pin on emojis messages
IMAGE 2021-05-31 15:22:14

churik:
### ISSUE 7: can't see transaction history on added accounts

**Steps:**
1) restore multiaccount from seed phrase
2) 
3) 
4) 

**Expected result:**

**Actual result:**

**OS:** IOS, Android

briansztamfater:
Thanks @churik, working on these issues now

src/status_im/chat/models/pin_message.cljs
briansztamfater: @shivekkhurana So I removed this logic, and simply subscribed to `chat/pinned-messages` to check if there is a pinned message for the message-id. Thanks for the suggestion!

errorists:
re 5, yes it should explicitly say who pinned them, truncated to a single line. also note that all the `Pinned` messages should share the same grey background, regardless who pinned them. 

Screenshot 2021-05-31 at 19 15 22

For the background of a pinned message, use `FFEECC` in light mode (made it slightly darker) and `34232B` in dark mode

briansztamfater:
@errorists Got it, I'll make those UI changes. I think there's a bug in status-go regarding the prop that tells who pinned the message, but I can implement the react part at the meantime -> https://github.com/status-im/status-go/issues/2234 cc @churik

briansztamfater:
@churik Issues should be addressed, let me know if anything else arises during testing.

Notice that I added option to see Pinned messages for public chats, let me know if we should keep this functionality for public chats, otherwise I can remove it.

https://user-images.githubusercontent.com/18485527/120385161-71ff4000-c2fd-11eb-9082-d1c16b20dcdb.mp4

Also, added the pinned by label, but bear in mind that won't function properly due to this issue:  status-im/status-go#2234

churik:
@briansztamfater 
based on https://github.com/status-im/status-react/issues/11699 and https://github.com/status-im/status/issues/11#issuecomment-773982069

### ISSUE 8: pin  messages shouldn't be available for public chats

> 1-1 chat
> group chat
> communities

### ISSUE 2 (NOT FIXED): Error when pin message exists in unfetched (previous) history
Reproducible.
with issue 8 makes sense 2d scenario:
**Steps for group chat:**
1) User A: create group chat with User B, send several messages
2) User B: join group chat
3) User A: pin messages that were sent upon joining User B

**Expected result:**
no error
**Actual result:**
IMAGE 2021-05-31 12:30:15

### ISSUE 9: Pin message should be available only for group chat / community admins
currently there are no limitations.

### ISSUE 10: No option to see pinned messages to community chats

**Steps:**
1) User A: create community ,invite User B
2) User B: join community
3) User A: pin message

**Expected result:**
can see pinned messages in chat info
**Actual result:**
no way to see pinned messages
**OS:** IOS, Android

Also IMO https://github.com/status-im/status-go/issues/2234 is blocker for this feature, as it confusing for users.

churik:
Agreed in communities channel to limit amount of pinned messages to 3.

**Expected behaviour:** 

1. admin in a group chat can Pin up to 3 messages
2. admin in community can Pin up to 3 messages in community channel
3. everyone in 1x1 chat can pin up to 3 messages in chat
4. when trying to pin more than 3 messages -> show **Pin limit reached** message with wording `Unpin previous message first`


Updated designs are [here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=20660%3A94234)

image
Desktop issue for reference is https://github.com/status-im/status-desktop/issues/2629

briansztamfater:
@churik Thanks for the feedback, issues should be fixed now.

I will work on the 3 pinned messages limit and integration with the `status-go` fix.

churik:
@briansztamfater thank you, let me know when PR will be ready to test

briansztamfater:
@churik Thanks for the patience, PR should be ready for testing now with all issues (hopefully) fixed, pins limit implemented and status-go fix for "pinned by" integrated.

Also @shivekkhurana could you make another pass on the review? Suggested fixes were implemented and some other things were added since last review.

churik:
@briansztamfater 
really nice work, thanks for fixes!

couple of minor stuff (don't block PR):

- [x] add accessibility id for "pinned-by" element
🙏 
Appium 2021-06-08 16-20-47
- [x] IOS only: keyboard is not closed when "Pin limit reached" is shown and focus is set in input field
IMAGE 2021-06-08 16:28:41
- [ ] [Dark mode] popup is black on black background
IMAGE 2021-06-08 16:31:01

briansztamfater:
@churik Thanks for the feedback! First two issues should be fixed now. Regarding the dark mode, which color should we use for popover background? I checked other popovers in Dark Mode and the background has the same color. Attaching example:

Captura de Pantalla 2021-06-08 a la(s) 12 21 59

Let me know if we need to change something there.

churik:
@briansztamfater you're right, not your PR - general issue.
Verifying 2 issues and good to go!

src/status_im/chat/models/loading.cljs
```diff
@@ -115,7 +115,8 @@
         (when (or first-request cursor)
           (merge
            {:db (assoc-in db [:pagination-info chat-id :loading-messages?] true)}
-           {:utils/dispatch-later [{:ms 100 :dispatch [:load-more-reactions cursor chat-id]}]}
+           {:utils/dispatch-later [{:ms 100 :dispatch [:load-more-reactions cursor chat-id]}
+                                   {:ms 100 :dispatch [:chat.ui/load-more-pin-messages-for-current-chat chat-id]}]}
```
flexsurfer: `for-current-chat` looks superfluous

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,126 @@
...
+                                                        pin-messages)]
+        {:db (-> db
+                 (assoc-in [:pin-messages chat-id] all-messages)
+                 (assoc-in [:pin-message-lists chat-id]
+                           (message-list/add-many nil (vals all-messages))))}))))
+
+(fx/defn load-more-pin-messages
+  {:events [:load-more-pin-messages]}
+  [{:keys [db]} chat-id first-request]
+  (let [session-id (get-in db [:pagination-info chat-id :pin-messages-initialized?])
```
flexsurfer: `{:keys []} (get-in db [:pagination-info chat-id])` should be shorter and more readable

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,126 @@
...
+        #(re-frame/dispatch [::failed-loading-pin-messages chat-id session-id %]))))))
+
+(fx/defn send-pin-message
+  "Pin message, rebuild pinned messages list"
+  {:events [::send-pin-message]}
+  [{:keys [db] :as cofx} {:keys [chat-id message-id pinned] :as pin-message}]
+  (let [current-public-key (get-in db [:multiaccount :public-key])
+        message (merge pin-message {:pinned-by current-public-key})]
+    (fx/merge cofx
+              {:db (as-> db $
```
flexsurfer: `cond->`

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -200,6 +205,35 @@
...
+    [quo/text {:weight :regular
+               :size   :small
+               :color  :main
+               :style  (style/pinned-by-text)}
+     (i18n/label :t/pinned-by)]
+    [quo/text {:weight :medium
+               :size   :small
+               :color  :main
+               :style  (style/pin-author-text)}
+     (pin-author-name pinned-by)]]])
```
flexsurfer: why is it a function call here?

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -331,21 +367,39 @@
...
-   (concat
-    (when (:show-input? message)
-      [{:on-press #(re-frame/dispatch [:chat.ui/reply-to-message message])
-        :label    (i18n/label :t/message-reply)}])
-    [{:on-press #(react/copy-to-clipboard
-                  (components.reply/get-quoted-text-with-mentions
-                   (get content :parsed-text)))
-      :label    (i18n/label :t/sharing-copy-to-clipboard)}])))
+(defn on-long-press-fn [on-long-press {:keys [public? pinned chat-id] :as message} content]
+  (let [{:keys [group-chat community-id admins]} @(re-frame/subscribe [:chats/current-chat-chat-view])
```
flexsurfer: im not sure about using subscribe inside function like that

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -489,21 +566,101 @@
...
+                                             :emoji-id          emoji-id
+                                             :emoji-reaction-id emoji-reaction-id}]))
+     :render          ->message}]
+   (when pinned
+     [react/view {:style (style/pin-indicator-container outgoing)}
+      [pinned-by-indicator outgoing display-photo? pinned-by]])])
+
+(def selected-unpin (reagent/atom nil))
+
+(defn render-pin-fn [{:keys [message-id outgoing] :as message}
```
flexsurfer: could we move pin functions to separate ns ?

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -54,6 +54,17 @@
                                        :padding-right 16}}
     [icons/icon :main-icons/more]]])
 
+(defn pins-topbar []
+  ;;we don't use topbar component, because we want chat view as simple (fast) as possible
```
flexsurfer: for pins screen we can use regular topbar

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -255,11 +273,35 @@
     [react/view {:style (when platform/android? {:scaleY -1})}
      [chat.group/group-chat-footer chat-id invitation-admin]]))
 
-(defn render-fn [{:keys [outgoing type] :as message}
+(defn render-fn [{:keys [outgoing type message-id] :as message}
                  idx
                  _
                  {:keys [group-chat public? current-public-key space-keeper chat-id show-input?]}]
-  [react/view {:style (when platform/android? {:scaleY -1})}
+  (let [pinned-messages @(re-frame/subscribe [:chats/pinned chat-id])
```
flexsurfer: message render function is most critical in the app, we should move this into messages stream subscription, so we should already have it in message

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -286,6 +329,12 @@
     (utils/set-timeout #(re-frame/dispatch [:chat.ui/load-more-messages-for-current-chat])
                        (if platform/low-device? 700 200))))
 
+(defn pin-list-on-end-reached []
+  (if @state/scrolling
```
flexsurfer: we have this optimization for messages list , but im not sure we need it for pinned messages list, could we simplify it and just use regular list ?

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -320,6 +369,62 @@
        :inverted                     (when platform/ios? true)
        :style                        (when platform/android? {:scaleY -1})})]))
 
+(defn pinned-messages-view [{:keys [chat bottom-space pan-responder space-keeper show-input?]}]
```
flexsurfer: could we move all pin related to separate ns ?

flexsurfer (changes requested):

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: It is a basic function which returns `You` text if public-key is the current account key, or the contact name otherwise.

src/status_im/ui/screens/chat/views.cljs
briansztamfater: Indeed I think we don't need this code anymore as specifications changed to have a limit of 3 messages, so I'll simplify it

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: Any suggestions on other possible approaches?

flexsurfer: i would move subscriptions outside, but im not sure its necessary, probably this will work ok as well, idk

qoqobolo:
@briansztamfater hey! Another related small issue was found in the latest build: https://github.com/status-im/status-react/issues/12217 
If you have time take a look, please. Maybe it would be nice to fix this within this PR.

churik:
As you've done some refactoring, it should be briefly rechecked before merge. Please ping me once approved (or Sergey, if it will be next week)

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: Ok I'd like to leave it as is if that's OK, and can refactor later in case if needed

src/status_im/chat/models/pin_message.cljs
briansztamfater: Actually I removed the code, pagination is not needed anymore as we have a limit of 3 pinned-messages now

briansztamfater:
@qoqobolo pushed the fix for long names, but there is one issue that happens only on Android that adds extra padding when there's a line break and would require further effort and writing a custom native component to fix it, so I would create a new issue (and another PR) for that one.

Here's an example of what I mean (I've used the same long name of the screenshot for the example):

**Android (extra padding):**
Captura de Pantalla 2021-06-10 a la(s) 19 46 41

**iOS (works fine):**
Captura de Pantalla 2021-06-10 a la(s) 19 45 17

churik:
thank you @briansztamfater 
anyway it looks much better than with overlapping https://github.com/status-im/status-react/issues/12217

src/status_im/contact/chat.cljs
```diff
@@ -23,4 +23,10 @@
               #(when new-contact?
                  (navigation/navigate-back %))
               #(when ens-name
-                 (contact/name-verified % public-key ens-name)))))
\ No newline at end of file
+                 (contact/name-verified % public-key ens-name)))))
+
+(fx/defn pinned-messages-pressed
+  {:events       [:contact.ui/pinned-messages-pressed]
+   :interceptors [(re-frame/inject-cofx :random-id-generator)]}
```
flexsurfer: why is :random-id-generator needed here ?

src/status_im/subs.cljs
```diff
@@ -1023,15 +1065,23 @@
 
 (def memo-timeline-messages-stream (atom nil))
 
+(re-frame/reg-sub
+ :chats/pinned-messages-stream
```
flexsurfer: it looks superfluous, why not using raw-chat-pin-messages-stream directly ?

src/status_im/subs.cljs
```diff
@@ -1023,15 +1065,23 @@
...
  :<- [:chats/message-list constants/timeline-chat-id]
  :<- [:chats/chat-messages constants/timeline-chat-id]
+ :<- [:chats/pinned constants/timeline-chat-id]
  :<- [:view-id]
- (fn [[message-list messages view-id]]
+ (fn [[message-list messages pinned-messages view-id]]
    (if (= view-id :status)
      (let [res (-> (models.message-list/->seq message-list)
-                   (hydrate-messages messages))]
+                   (hydrate-messages messages pinned-messages))]
```
flexsurfer: do we really want pinned messages for timeline ?

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -405,48 +461,69 @@
...
+                                  {:disabled      in-popover?
+                                   :on-press      (fn []
                                                     (react/dismiss-keyboard!))
                                    :on-long-press (fn []
-                                                    (on-long-press
-                                                     [{:on-press #(re-frame/dispatch [:chat.ui/reply-to-message message])
-                                                       :label    (i18n/label :t/message-reply)}
-                                                      {:on-press #(react/copy-to-clipboard (get content :text))
-                                                       :label    (i18n/label :t/sharing-copy-to-clipboard)}]))})
+                                                    (let [{:keys [chat-id group-chat community-id admins]} @(re-frame/subscribe [:chats/current-chat-chat-view])
```
flexsurfer: it seems like these all are chat related, not message related, so we must subscribe only once in chat view and pass this data to every message

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -331,21 +368,39 @@
...
-    (when (:show-input? message)
-      [{:on-press #(re-frame/dispatch [:chat.ui/reply-to-message message])
-        :label    (i18n/label :t/message-reply)}])
-    [{:on-press #(react/copy-to-clipboard
-                  (components.reply/get-quoted-text-with-mentions
-                   (get content :parsed-text)))
-      :label    (i18n/label :t/sharing-copy-to-clipboard)}])))
+(defn on-long-press-fn [on-long-press {:keys [public? pinned chat-id] :as message} content]
+  (let [{:keys [group-chat community-id admins]} @(re-frame/subscribe [:chats/current-chat-chat-view])
+        current-pk @(re-frame/subscribe [:multiaccount/public-key])
```
flexsurfer: this looks duplicated with emoji message

flexsurfer (changes requested):

flexsurfer:
chat messages are a very critical part of the app, and the performance of the render function is critical as well, we should be really careful with changes

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,99 @@
...
+                                                              (and pinned (nil? current-message))
+                                                              (update :new-messages conj current-message-pin)
+
+                                                              (nil? pinned)
+                                                              (update :all-messages dissoc message_id)
+
+                                                              (and (some? pinned) (some? current-message))
+                                                              (update :all-messages assoc message_id current-message-pin))))
+                                                        {:all-messages already-loaded-pin-messages
+                                                         :new-messages []}
```
shivekkhurana: 1. Do we need this empty `:new-messages`? 
2. Can you please document the reduce function? It's hard to understand whats happening in a glance.

src/status_im/ui/screens/chat/pinned_messages.cljs
```diff
@@ -0,0 +1,126 @@
...
+        space-keeper]))])
+
+(def list-key-fn #(or (:message-id %) (:value %)))
+(def list-ref #(reset! messages-list-ref %))
+
+(defn pinned-messages-view [{:keys [chat bottom-space space-keeper show-input?]}]
+  (let [{:keys [group-chat chat-id public?]} chat
+        pinned-messages @(re-frame/subscribe [:chats/pinned-messages-stream chat-id])
+        current-public-key @(re-frame/subscribe [:multiaccount/public-key])]
+    ;;do not use anonymous functions for handlers
```
shivekkhurana: Why?

shivekkhurana (approved):

src/status_im/ui/screens/chat/pinned_messages.cljs
flexsurfer: because each time new function will be generated, and it will be new object for react, but it should be the same function

src/status_im/subs.cljs
briansztamfater: absolutely

src/status_im/contact/chat.cljs
briansztamfater: not needed at all, removing

src/status_im/subs.cljs
briansztamfater: i think we can remove it

src/status_im/chat/models/pin_message.cljs
briansztamfater: 1. We don't need it, I refactored this part and forgot to remove it. 
2. Sure, adding it right now (and also simplifying the function)

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: Done, added as additional data on each message injected on flat list

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: Added separate functions to share this logic

src/status_im/ui/screens/chat/message/gap.cljs
```diff
@@ -4,12 +4,14 @@
...
-            [status-im.ui.screens.chat.styles.input.gap :as style]))
+            [status-im.ui.screens.chat.styles.input.gap :as style]
+            [status-im.chat.models.pin-message :as models.pin-message]))
 
 (defn on-press
   [chat-id gap-ids]
   (fn []
-    (re-frame/dispatch [:chat.ui/fill-gaps chat-id gap-ids])))
+    (re-frame/dispatch [:chat.ui/fill-gaps chat-id gap-ids])
+    (re-frame/dispatch [::models.pin-message/load-pin-messages chat-id])))
```
flexsurfer: not sure if it's necessary here? at least i don't see reactions and messages here , cc @cammellos

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -331,7 +368,28 @@
 (def message-height-px 200)
 (def max-message-height-px 150)
 
-(defn on-long-press-fn [on-long-press message content]
+(defn message-pin-enabled [{:keys [public? group-chat from-community admins]}]
+  (let [current-pk @(re-frame/subscribe [:multiaccount/public-key])
```
flexsurfer: we already have this in the message
```
[message/chat-message
        (assoc message
               :incoming-group (and group-chat (not outgoing))
               :group-chat group-chat
               :public? public?
               :current-public-key current-public-key
               :show-input? show-input?)
```

so you can use this data, or if you need something else just add it there, we don't want to subscribe in every message

flexsurfer (changes requested):

src/status_im/ui/screens/chat/message/message.cljs
flexsurfer: you could assoc `:message-pin-enabled ` for every message

src/status_im/ui/screens/chat/message/gap.cljs
briansztamfater: yeh we can remove this also, not needed anymore

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: Agree, forgot to update this file, just removed this function and moved the logic to the chat level. Also rebased to integrate with React Native Navigation changes.

src/status_im/chat/models.cljs
```diff
@@ -206,7 +207,15 @@
   "Takes chat-id and coeffects map, returns effects necessary when navigating to chat"
   {:events [:chat.ui/preload-chat-data]}
   [{:keys [db] :as cofx} chat-id]
-  (loading/load-messages cofx chat-id))
+  (fx/merge cofx
+            (loading/load-messages chat-id)
+            (pin-message/load-pin-messages chat-id)))
```
flexsurfer: we already have it in load-more-messages, so we don't need load-pin-messages here

src/status_im/chat/models.cljs
```diff
@@ -206,7 +207,15 @@
...
+
+(fx/defn preload-chat-pin-data
+  "Takes chat-id and coeffects map, returns effects necessary for preloading pinned messages"
+  {:events [:chat.ui/preload-chat-pin-data]}
+  [{:keys [db] :as cofx} chat-id]
+  (pin-message/load-pin-messages chat-id))
```
flexsurfer: why not use load-pin-messages event direclty ?

src/status_im/chat/models.cljs
```diff
@@ -232,6 +241,15 @@
     {:db (assoc-in db [:chats chat-id] chat)
      :dispatch [:chat.ui/navigate-to-chat chat-id]}))
 
+(fx/defn navigate-to-user-pinned-messages
+  "Takes coeffects map and chat-id, returns effects necessary for navigation and preloading data"
+  {:events [:chat.ui/navigate-to-pinned-messages]}
+  [{db :db :as cofx} chat-id]
+  (fx/merge cofx
+            {:db (assoc db :current-chat-id chat-id)}
+            (preload-chat-data chat-id)
```
flexsurfer: why do we need chat data here ? if its possible to open pinned messages outside the chat , it might break chat

src/status_im/subs.cljs
```diff
@@ -948,14 +962,24 @@
...
-           %)
-        message-list))
+  ([message-list messages] (hydrate-messages message-list messages {}))
+  ([message-list messages pinned-messages]
+   (keep #(if (= :message (% :type))
+            (when-let [message (messages (% :message-id))]
+              (let [pinned-message (get pinned-messages (% :message-id))
+                    pinned (if pinned-message true (some? (message :pinned-by)))
+                    pinned-by (when pinned (or (message :pinned-by) (pinned-message :pinned-by)))]
+                (merge message % {:pinned pinned :pinned-by pinned-by})))
```
flexsurfer: assoc should be faster i guess

src/status_im/ui/screens/chat/sheets.cljs
```diff
@@ -25,7 +26,9 @@
        :subtitle            (i18n/label :t/view-profile)
        :accessibility-label :view-chat-details-button
        :chevron             true
-       :on-press            #(hide-sheet-and-dispatch  [:chat.ui/show-profile chat-id])}]
+       :on-press            #(do
+                               (hide-sheet-and-dispatch  [:chat.ui/show-profile chat-id])
+                               (re-frame/dispatch [::models.pin-message/load-pin-messages chat-id]))}]
```
flexsurfer, line 31: is this for timeline ? do we have pinned messages for timeline ?

src/status_im/ui/screens/profile/contact/views.cljs
```diff
@@ -187,6 +198,7 @@
                                         :subtitle second-name})]
                                      [react/view {:height 1 :background-color colors/gray-lighter :margin-top 8}]
                                      [nickname-settings contact]
+                                     [pin-settings public-key (count pinned-messages)]
```
flexsurfer, line 202: do we really want pined messages for statuses ?

src/status_im/constants.cljs
```diff
@@ -66,6 +66,7 @@
 (def ^:const min-password-length 6)
 (def ^:const max-group-chat-participants 20)
 (def ^:const default-number-of-messages 20)
+(def ^:const default-number-of-pin-messages 50)
```
flexsurfer: why is the max number of pinned messages?

src/status_im/chat/models/pin_message.cljs
```diff
@@ -0,0 +1,94 @@
...
+(fx/defn send-pin-message
+  "Pin message, rebuild pinned messages list"
+  {:events [::send-pin-message]}
+  [{:keys [db] :as cofx} {:keys [chat-id message-id pinned] :as pin-message}]
+  (let [current-public-key (get-in db [:multiaccount :public-key])
+        message (merge pin-message {:pinned-by current-public-key})]
+    (fx/merge cofx
+              {:db (cond-> db
+                     pinned
+                     (update-in [:pin-message-lists chat-id] message-list/add message)
```
flexsurfer: will (-> work here?
```
pinned 
(->
    (update-in [:pin-message-lists chat-id] message-list/add message)
     (assoc-in [:pin-messages chat-id message-id] message)
                     ```

File: src/status_im/subs.cljs [hunk 43]
```diff
@@ -969,21 +993,39 @@
...
        (-> message-list-seq
            (chat.db/add-datemarks)
-           (hydrate-messages messages)
+           (hydrate-messages messages pin-messages)
            (chat.db/collapse-gaps chat-id synced-from (datetime/timestamp) chat-type joined loading-messages?))))))
 
+(re-frame/reg-sub
+ :chats/raw-chat-pin-messages-stream
+ (fn [[_ chat-id] _]
+   [(re-frame/subscribe [:chats/pin-message-list chat-id])
```
flexsurfer, line 1019: this is more a general question, are we duplicate messeges for pin message ? can't we use just ids ? and get messages from chat ?

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -32,39 +36,40 @@
...
-                    :delivered :tiny-icons/tiny-delivered
-                    :tiny-icons/tiny-pending)
-       {:width               16
-        :height              12
-        :color               colors/white
-        :accessibility-label (name outgoing-status)}])
-    [react/text {:style (style/message-timestamp-text outgoing)}
-     timestamp-str]]))
+  ([{:keys [timestamp-str outgoing content outgoing-status pinned in-popover?]} justify-timestamp?]
+   (when-not in-popover?
```
flexsurfer: could you please add a comment on what this in-popover? about

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -489,21 +545,101 @@
...
-                                           {:message-id (:message-id message)
-                                            :emoji-id   emoji-id}]))
-    :retract-emoji   (fn [{:keys [emoji-id emoji-reaction-id]}]
-                       (re-frame/dispatch [::models.reactions/send-emoji-reaction-retraction
-                                           {:message-id        (:message-id message)
-                                            :emoji-id          emoji-id
-                                            :emoji-reaction-id emoji-reaction-id}]))
-    :render          ->message}])
+(defn chat-message [{:keys [outgoing display-photo? pinned pinned-by] :as message} space-keeper]
+  [react/view
```
flexsurfer: better to use `[<>` here , so we don't add view for every message

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -489,21 +545,101 @@
...
+            :in-popover? true)
+     space-keeper]
+    [react/view {:style {:position    :absolute
+                         :right       18
+                         :padding-top 4}}
+     [radio/radio (= @selected-unpin message-id)]]]])
+
+(def list-key-fn #(or (:message-id %) (:value %)))
+
+(defn pinned-messages-view [chat-id]
```
flexsurfer: could we move all pin related to separate ns ?

src/status_im/ui/screens/chat/pinned_messages.cljs
```diff
@@ -0,0 +1,136 @@
...
+               :show-input? show-input?
+               :message-pin-enabled message-pin-enabled
+               :pinned true)
+        space-keeper]))])
+
+(def list-key-fn #(or (:message-id %) (:value %)))
+(def list-ref #(reset! messages-list-ref %))
+
+(defn pinned-messages-view [{:keys [chat bottom-space space-keeper show-input?]}]
+  (let [{:keys [group-chat chat-id public? community-id admins]} chat
```
flexsurfer, line 56: general question, it seems like we duplicate all from chat for pinned messages , it looks really complex, and in the future we have to consider this and make changes in two places

flexsurfer (dismissed):

src/status_im/constants.cljs
briansztamfater: We can set the max to 3 which is the limit now, 50 is no more valid, but the reason was to preferably load a larger amount of pinned messages on each page. I'll modify it.

src/status_im/subs.cljs
briansztamfater: It is a completely valid question, right now it is possible to see pinned messages without having messages from chat loaded (i.e accessing from user profile to pinned messages), so it can be possible to not have the message loaded and get the pinned messages from json-rpc API (which also provides some additional properties such as `pinned-by`, `pinned-at` and `pinned`. Also there are some diffs on params names which can conflict (message_id is one example, and on receiving signals they come with some differences too)

src/status_im/ui/screens/chat/pinned_messages.cljs
briansztamfater: Yeah nice catch, I'll see how we can avoid duplicates and share code here

src/status_im/ui/screens/profile/contact/views.cljs
briansztamfater: I'd say yes, it is the way to access pinned messages from a 1-to-1 chat. Example:

src/status_im/ui/screens/chat/sheets.cljs
briansztamfater: It is to preload pinned messages to show the pinned messages count on the profile even if we don't have them previously loaded

src/status_im/chat/models/pin_message.cljs
briansztamfater: Yes! I wasn't completely happy with this implementation so thanks, I yet have to fully incorporate threading macros

src/status_im/ui/screens/chat/message/message.cljs
briansztamfater: To keep track if message is being shown in a pin-limit-popover list and disable some props / adjust style. Added comment in the code also.

src/status_im/chat/models.cljs
briansztamfater: Wasn't using this function or event at all, removed

jakubgs:
If you're wondering why your iOS jobs are not running, it's because of #12228. You should rebase.

briansztamfater:
> If you're wondering why your iOS jobs are not running, it's because of #12228. You should rebase.

thanks, done!

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -320,7 +329,8 @@
         pan-responder (accessory/create-pan-responder position-y pan-state)
         space-keeper (get-space-keeper-ios bottom-space panel-space active-panel text-input-ref)
         set-active-panel (get-set-active-panel active-panel)
-        on-close #(set-active-panel nil)]
+        on-close #(set-active-panel nil)
+        messages (re-frame/subscribe [:chats/chat-messages-stream curr-chat-id])]
```
flexsurfer, line 333: it's better to move it inside messages-view , and use in condition, because we don't want to re-render other parts of chat view when messages are changed

flexsurfer (changes requested):

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -265,39 +266,47 @@
...
+       :data                         (if inverted-ui (reverse @messages) @messages)
+       :render-data                  {:group-chat          group-chat
+                                      :public?             public?
+                                      :current-public-key  current-public-key
+                                      :space-keeper        space-keeper
+                                      :chat-id             chat-id
+                                      :show-input?         show-input?
+                                      :message-pin-enabled message-pin-enabled}
        :render-fn                    render-fn
        :on-viewable-items-changed    on-viewable-items-changed
```
flexsurfer, line 314: so for pin messages, the max number of items is 3 ? and we don't need paging for pinned messages right ? in that case its better to do not share messages-view and just create `get-render-data` function and share it?

briansztamfater: Exactly, and makes sense as that should share most of the subscriptions. Moved `render-data` to a separate function now

src/status_im/ui/screens/chat/message/pinned_message.cljs
```diff
@@ -0,0 +1,88 @@
...
+            :in-popover? true)
+     space-keeper]
+    [react/view {:style {:position    :absolute
+                         :right       18
+                         :padding-top 4}}
+     [radio/radio (= @selected-unpin message-id)]]]])
+
+(def list-key-fn #(or (:message-id %) (:value %)))
+
+(defn pinned-messages-view [chat-id]
```
flexsurfer: just wondering, where is this used? i see we have same function in pinned_messages.cljs

src/status_im/ui/screens/chat/pinned_messages.cljs
```diff
@@ -0,0 +1,121 @@
...
+                       :justify-content :center}}
+   [react/text {:style style/intro-header-description}
+    (i18n/label :t/pinned-messages-empty)]])
+
+(defonce messages-list-ref (atom nil))
+
+(def list-key-fn #(or (:message-id %) (:value %)))
+(def list-ref #(reset! messages-list-ref %))
+
+(defn render-pin-fn [{:keys [outgoing type] :as message}
```
flexsurfer: can't we use regular message rendering? how's it different?

flexsurfer (approved):
Great Job!

src/status_im/ui/screens/chat/message/pinned_message.cljs
briansztamfater: it is used for `pin-limit-popover` popover where the user should pick a pinned message to unpin in order to pin a new one. but you are right i'll rename this as is confusing

src/status_im/ui/screens/chat/pinned_messages.cljs
briansztamfater: we are forcing pinned to true but it is probably redundant, i think it should work, i'll try it

briansztamfater: done!

briansztamfater:
Hey @churik lots of things have changed in the code since your last test of this PR. Can you do some re-test to check if I didn't break anything? Thanks!!

churik:
@briansztamfater thank you for PR!
Several issues: 
**1) On Android pinned messages appear in reverse**
IMAGE 2021-07-01 15:00:30
**2) Can tap on disabled "unpin"**
In this case message will be marked as pinned in chat, but in fact it will not be pinned

https://user-images.githubusercontent.com/4557972/124128837-71cd9e00-da7d-11eb-9f0b-712c1ef42793.mp4

briansztamfater:
@churik I'm having issues to test in my android emulator, but both issues should be fixed now.

churik:
@briansztamfater :)

this is from my Android:
IMAGE 2021-07-02 10:34:39

briansztamfater:
@churik fix pushed!

churik:
Tested pinned messages in dark and light mode, pinning/unpinning messages (online/offline), in 1-1 and group chats.
Amazing work, thank you!

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)

PR #11963, opened 2021-04-07
🗂️ Anon metrics batching

shivekkhurana:
### Summary
- Implement batching and transformers to capture all navigation events
- Implement a long running to avoid congestion on the bridge
- Fix the issue that creeped in the last merge https://github.com/status-im/status-react/pull/11941#discussion_r605585941
- Dependant on https://github.com/status-im/status-go/pull/2189

### Testing notes
The app should function as usual. No impact on the UI.

#### Platforms
- Android
- iOS

##### Non-functional

status: ready


src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,90 @@
...
+   :before catch-events-before))
+
+(re-frame/reg-fx
+ ::transform-and-log
+ (fn [context]
+   (transform-and-log context)))
+
+(fx/defn hoax-capture-event
+  "Due to usage of fx/defn with fx/merge, it might not be able to intercept some events
+  (like navigate-to-cofx). In cases like that, this hoax capture event can be used in conjunction with `fx/merge`"
```
rasom: Can we please keep it inside 80 chars limit?

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,90 @@
...
+  {:events [::hoax-capture-event]}
+  [_ {:keys [og-event]}]
+  ;; re-shape event to look like a context object
+  {::transform-and-log {:coeffects {:event og-event}}})
+
+(comment
```
rasom, line 89: to be removed?

src/status_im/anon_metrics/transformers.cljs
```diff
@@ -1,14 +1,24 @@
...
+     :params  {:screen (get-in event [1 :screen] "")}}))
 
-(def transformations
-  {:navigate-to navigate-to-txf})
+(defn screens-on-will-focus-txf [context]
+  (let [view-id (-> context :coeffects :event second)]
+    ;; we need this check because in some cases, :screens/on-will-focus is dispatched after navigate to
+    ;; this check ensures that only the events that are not handled via navigate to are captured
+    ;; hence avoiding double entry
+    (when (some #(= view-id %) [:chat :empty-tab :wallet :status :my-profile])
```
rasom: `(contains? #{:chat :empty-tab :wallet :status :my-profile} view-id)` requires less parentheses, not a big deal though

rasom (commented):
Do we want to keep all those `comment` blocks?
Also when it is a new code, can we please keep it under 80 chars limit? I switched back to 13 inches display and that's again a noticeable issue. :)

src/status_im/anon_metrics/core.cljs
shivekkhurana: No, it's a [rich comment](https://betweentwoparens.com/rich-comment-blocks#rich-comment)
> These comments are rich because they provide rich detail about the development process.

I have kept it here because when a new event is being captured, we need to check if the correct form is being stored to the database. There is no other way to access that data. The comment also captures the thought process.

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -240,7 +240,6 @@
   (let [loading-messages? @(re-frame/subscribe [:chats/loading-messages? chat-id])
         no-messages? @(re-frame/subscribe [:chats/chat-no-messages? chat-id])
         all-loaded? @(re-frame/subscribe [:chats/all-loaded? chat-id])]
-    (println "LIST FOOTER" (or loading-messages? (not chat-id) (not all-loaded?)))
```
flexsurfer: .

src/status_im/navigation.cljs
```diff
@@ -38,15 +38,16 @@
...
 (fx/defn navigate-to-cofx
-  [{:keys [db]} go-to-view-id screen-params]
+  [{:keys [db] :as cofx} go-to-view-id screen-params]
   {:db
    (-> (assoc db :view-id go-to-view-id)
        (all-screens-params go-to-view-id screen-params))
-   ::navigate-to [go-to-view-id screen-params]})
+   ::navigate-to [go-to-view-id screen-params]
+   ;; simulate a navigate-to event so it can be captured be anon-metrics
+   ::anon-metrics/transform-and-log {:coeffects {:event [:navigate-to go-to-view-id screen-params]}}})
```
flexsurfer, line 47: whats that ?

src/status_im/anon_metrics/transformers.cljs
flexsurfer: `(#{:chat :empty-tab :wallet :status :my-profile} view-id)` even better

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,90 @@
...
+
+(defn onboard-events
+  "Check if there are any events in the foyer, flush them to the backend and clear foyer on-success."
+  []
+  (let [outstanding-events @events-foyer]
+    (when (seq outstanding-events)
+      (json-rpc/call {:method     "appmetrics_saveAppMetrics"
+                      :params     [outstanding-events]
+                      ;; clear the foyer when events are stored
+                      :on-success #(reset! events-foyer [])
```
flexsurfer: it might have new events in it, why not to clear it before call ?

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,90 @@
...
+                      :on-error   (fn [err]
+                                    (log/error {:error err
+                                                :events outstanding-events})
+
+                                    (log/warn "All outstanding events will be rejected")
+                                    (reset! events-foyer []))}))))
+
+(re-frame/reg-fx
+ ::transfer-data
+ (fn [transfer?]
```
flexsurfer, line 33: no safety for a double run?

flexsurfer (dismissed):

src/status_im/anon_metrics/core.cljs
shivekkhurana: Great catch, fixed.

src/status_im/anon_metrics/core.cljs
shivekkhurana: I wanted to refactor this bit, didn't realise some events would be lost. Fixed it now. The foyer is reset before the RPC.

src/status_im/navigation.cljs
shivekkhurana: We wrote an interceptor to keep the metrics code away from app code. But some events are using `fx/merge`, not via direct dispatch. Also sometimes, we might want to capture events in an `fx/merge` chain. 

The effect `::transform-and-log` and the handler `::hoax-event-capture` help out with this situation. Both call `transform-and-log` function internally, which expects an Re-frame context (so it can parse the event and strip unnecessary information).

Since this event-handler is used with `fx/merge`, we cannot intercept it. So need to use the effect directly.

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,90 @@
...
+   (if (and transfer?
+            ;; double run safety
+            (not @periodic-tasks-chan))
+     ;; run the onboard-events fn periodically
+     (reset! periodic-tasks-chan
+             ;; interval = 4000 ms (run every `interval` milliseconds)
+             ;; timeout = 5000 ms (exit if the fn doesn't exit within `timeout` milliseconds)
+             (async-periodic-exec onboard-events 4000 5000))
+     (async-periodic-stop! @periodic-tasks-chan))))
```
flexsurfer: will periodic-tasks-chan be cleared here ?

shivekkhurana:
```clojure
(defn async-periodic-stop! [async-periodic-chan]
  (async/close! async-periodic-chan))
```

`async/close!` will will only close the channel but not clear it. We don't have anything in the channel, just a timeout. Everything is stored in the `events-foyer` atom. So we should clear that manually.

flexsurfer: yes thats why im asking, i don't see we clear it manually

shivekkhurana: I called the `onboard-events` function, that will clear the atom and also save everything to the db.

flexsurfer: mhm i don't see if it cleans periodic-tasks-chan

shivekkhurana: We don't put anything on the channel. It's never "dirty" for us to "clean".
https://github.com/status-im/status-react/blob/90ad5f2f77964eff1a7b155e475dea79f5ae4622/src/status_im/utils/async.cljs#L67

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,94 @@
...
+                                    (log/error {:error err
+                                                :events outstanding-events})
+                                    (log/warn "All outstanding events will be rejected"))}))))
+
+(re-frame/reg-fx
+ ::transfer-data
+ (fn [transfer?]
+   (if (and transfer?
+            ;; double run safety
+            (not @periodic-tasks-chan))
```
flexsurfer, line 36: i mean here, if you stop it and then try to run again it won't work, because we didn't reset it to nil

src/status_im/anon_metrics/core.cljs
```diff
@@ -0,0 +1,94 @@
...
+     ;; run the onboard-events fn periodically
+     (reset! periodic-tasks-chan
+             ;; interval = 4000 ms (run every `interval` milliseconds)
+             ;; timeout = 5000 ms (exit if the fn doesn't exit within `timeout` milliseconds)
+             (async-periodic-exec onboard-events 4000 5000))
+     (do
+       (async-periodic-stop! @periodic-tasks-chan)
+
```
flexsurfer, line 47: so here we need `(reset! periodic-tasks-chan nil)`

src/status_im/anon_metrics/core.cljs
shivekkhurana: Sorry, that's so stupid of me. I'll fix it. Thank you.

flexsurfer (approved):

PR #11969, opened 2021-04-08
Jenkins URL to rerun tests

Serhy:
Status: ready.
When there are failed tests reported in PR, - when revealing dropdown of Failed Tests - there is 'Rerun tests' which leads to Jenkins with pre-filled items to specify before starting the job (list of test ids, PR number, build. Branch is default to 'develop' for now)


churik (approved):
👍

PR #11970, opened 2021-04-08
Back / dismiss gestures on Android

### LINKED ISSUE #11913: Back gesture on Android
### User Story

As a user, I want to use a swipe-to-back gesture in Status on Android devices. 

### Description

Currently, users have to press the arrow to go back, even on Android devices that have the back gesture.

### Acceptance Criteria

Users can use the back gesture in the Status app on all Android devices that support this type of navigation.

### Notes

The feature request was originally received from the user to error-reports@status.im

briansztamfater:
fixes #11913 

### Summary

We need to enable navigation gestures on Android. In this PR i'm doing a refactor on the routing core functions where `gestureEnabled` is set to true by default on all screens on Android, which is enabling swipe gestures for navigating back and therefore that comes with animated transitions. Also we are applying different transition animations for modal presentations (default is an horizontal transition from left to right, and for modal presentation is a vertical transition from bottom to top). I set some arbitrary `gestureResponseDistance` values for enabling the swipe gesture, but can be adjusted. Finally, refactored `:presentation-ios` to `:presentation-modal` because we are enabling modal presentation for both OS.

navigation-gestures-android

### Testing notes

Be sure that no screen was broken.

#### Platforms

- Android
- iOS (changes impact on Android but doing some smoke testing on iOS is recommended)

General navigation

##### Non-functional

- Navigation transitions and gestures

### Steps to test

- Open Status
- Navigate through screens
- Swipe horizontally to go back, or swipe vertically to dismiss a modal

status: ready

### DIFF
--- a/src/status_im/ui/screens/routing/core.cljs
+++ b/src/status_im/ui/screens/routing/core.cljs
@@ -2,7 +2,7 @@
   (:require ["react" :refer (useCallback useEffect)]
             ["react-native" :refer (BackHandler)]
             ["@react-navigation/native" :refer (NavigationContainer StackActions CommonActions useFocusEffect) :as react-navigation]
-            ["@react-navigation/stack" :refer (createStackNavigator TransitionPresets)]
+            ["@react-navigation/stack" :refer (createStackNavigator TransitionPresets CardStyleInterpolators)]
             ["@react-navigation/bottom-tabs" :refer (createBottomTabNavigator)]
             [reagent.core :as reagent]
             [cljs-bean.core :refer [bean]]
@@ -43,9 +43,25 @@
 
 (def transition-presets TransitionPresets)
 
+(def card-style-interpolators CardStyleInterpolators)
+
+(def back-gesture-response-distance 70)
+
+(def dismiss-gesture-response-distance 100)
+
 (def modal-presentation-ios (merge (js->clj (.-ModalPresentationIOS ^js transition-presets))
                                    {:gestureEnabled     true
                                     :cardOverlayEnabled true}))
+
+(def enable-gestures-default-android {:gestureEnabled true
+                                      :gestureDirection :horizontal
+                                      :gestureResponseDistance {:horizontal back-gesture-response-distance}
+                                      :cardStyleInterpolator (js->clj (.-forHorizontalIOS ^js card-style-interpolators))})
+
+(def enable-gestures-modal-android {:gestureEnabled true
+                                    :gestureDirection :vertical
+                                    :gestureResponseDistance {:vertical dismiss-gesture-response-distance}
+                                    :cardStyleInterpolator (js->clj (.-forVerticalIOS ^js card-style-interpolators))})
 
 ;; TODO(Ferossgp): Unify with topbar back icon. Maybe dispatch the same event and move the all logic inside the event.
 (defn handle-on-screen-focus
@@ -97,11 +113,13 @@
                       (get style :padding-vertical))})))
 
 (defn presentation-type [{:keys [transition] :as opts}]
-  (if (and platform/ios? (= transition :presentation-ios))
+  (if (and platform/ios? (= transition :presentation-modal))
     (-> opts
         (update :options merge modal-presentation-ios)
         (assoc-in [:insets :top] false))
-    opts))
+    (if platform/android?
+      (-> opts (update :options merge (if (= transition :presentation-modal) enable-gestures-modal-android enable-gestures-default-android)))
+      opts)))
 
 (defn wrap-screen [{:keys [component] :as options}]
   (assoc options :component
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

src/status_im/ui/screens/routing/core.cljs
```diff
@@ -97,11 +113,13 @@
                       (get style :padding-vertical))})))
 
 (defn presentation-type [{:keys [transition] :as opts}]
-  (if (and platform/ios? (= transition :presentation-ios))
+  (if (and platform/ios? (= transition :presentation-modal))
```
shivekkhurana, line 116: [`cond`](https://clojuredocs.org/clojure.core/cond) is more appropriate than `if`.

shivekkhurana (changes requested):
This looks great 👏🏼

errorists:
Hey a question, is this push-pull navigation segue how Android supposed to work like these days? Because to me it easily takes the prize for the most non-Android shit I've seen yet 😅

briansztamfater:
> Hey a question, is this push-pull navigation segue how Android supposed to work like these days? Because to me it easily takes the prize for the most non-Android shit I've seen yet 😅

True, it is replicating iOS-style transitions, which may not be what an Android user is expecting. This is the Android Oreo-style transition alternative we have, what do you think?

android-oreo-transition

errorists:
Why do we want our own transition in place if the [Android 10 system back gesture utilises a floating arrow indicator that maps to your drag gesture](https://cdn-vox--cdn-com.cdn.ampproject.org/i/s/cdn.vox-cdn.com/thumbor/rf9q5lnrGcvFtJ-gasqVAfegw3Q=/1400x0/filters:no_upscale()/cdn.vox-cdn.com/uploads/chorus_asset/file/16217483/nav.0.gif). Once registered we fire up a regular back event like we would with the hardware button. No need to have our transitions, let's use the one found on Android.

briansztamfater:
Yeah, I'm not an Android user so I'm not aware of the different transitions of different Android versions, but tried enabling the gestures from the OS settings and that already works in `develop` branch version, at least on my emulator. Example below 👇 

android-back-default

Another option is to keep the default transition but adding the gesture. Example below 👇 

android-default-gesture
 
So to sum up we could:

1. Keep the `develop` version (maybe the original feature request is not needed)
2. Keep the default transition _but_ add swipe / dragging gesture to go back

Any thoughts?

errorists:
Let's keep the develop 👍 It's a system-wide behavior that we don't want to customize. Likewise, I'm not an Android person, just know that porting iOS UI is the easiest way to piss those people off, when in doubt let's try not to :)

shivekkhurana:
Swipe from left to go back is a fairly common Android technique (iOs too). I think native routing libraries implement it by default. It's something that is enabled or disabled at system level in Android.

churik:
So let's summarize.
I see several issues during onboarding when using swipe-to-back gesture on Android (null-pointer exception, possibility to create account without setting a password).
Original feature request came from user error-report https://github.com/status-im/status-react/issues/11913 and was reported by @qoqobolo 

In several chat apps (Telegram, Viber) swipe-to-back gesture works as back button, in other apps (Play Store, FB messenger) - it doesn't work.

As PR definitely contains some issues and as it comes from discussion between @errorists and @briansztamfater best solution IMO is to close this PR and original feature request.

Objections?

errorists:
to add, my only objection was to add the iOS native transitions that were introduced first in the PR. As long as we implement the swipe back gestures using native Android behaviour, it's perfectly fine and a welcome improvement

churik:
@errorists tell me what do you think.
AFAIK there are 2 types of navigation on Android - "Full screen gestures" or "Buttons" (gestures are not using on system level, nav buttons are always shown on-screen)

### so on current `develop`:
1) if system navigation is set to "Full screen gestures" (screenshot is attached) swipe-back is working inside app (when you logged in) but not working during onboarding on screens that instead of  <- contain "Cancel" 
Monosnap 2021-04-12 11-49-14Monosnap 2021-04-12 11-38-22

2)  if system navigation is set to "Buttons" swipe-back is not working inside app (I believe it is kind of additional feature and isn't related to native Android behavior and this is origin al issue https://github.com/status-im/status-react/issues/11913) 
As mentioned  in https://github.com/status-im/status-react/pull/11970#issuecomment-817637914 it is implemented selectively in some applications)

### on current PR build:
3) if system navigation is set to "Buttons" swipe-back gesture works on any screens and can cause new issues.

So we have several options here:
a) close this PR as it seems additional feature and it is not really required
b) fix all issues related to new navigation behavior (will report them if needed), as it changed a lot.

As for me better option is (a)

briansztamfater:
I think we can close this one. Feature request is not really required so spending more dev time in something that is not needed is not what we want. Thanks for your input guys!

PR #11982, opened 2021-04-12
do not render not active screen

flexsurfer:
do not render not active screen


shivekkhurana (approved):

Serhy:
Smoke tested chats, wallet, Status tab, - info updates when received while on another tab. Tested on iOS 14 and Android 11.

PR #12005, opened 2021-04-19
Load message on-component-did-mount

### LINKED ISSUE #12006: Duplicated items after navigation in chat view from Activity Center
# Bug Report

## Problem

If I have 2 or more items in Activity Center, and I navigate to chat view (Group/1-1) from Activity Center (doesn't matter which one - `Group`->`Decline` or `Group` -> `Join` or open 1-1 chat) the next time I open Activity Center - there will be still two chat items but the one which supposed to be removed from list, - it duplicates the one which is left.

(it's outstanding issue from https://github.com/status-im/status-react/pull/11962#issuecomment-821324955)

#### Expected behavior

No chat item in Activity Center after navigation in it's chat view

#### Actual behavior

https://user-images.githubusercontent.com/8749671/115212217-f2306400-a108-11eb-996c-867e1f0a3f13.mov

### Reproduction

- Create two new accounts: UserA and UserB
- UserA adds UserB in contacts
- UserA sends message in 1-1 chat to UserB
- UserA creates group chat with UserB
- UserB from Chats view opens Activity Center
- UserB taps on group chat item
- UserB navigates back to Activity Center
**Actual result:** 
Screenshot 2021-04-19 at 12 25 06

### Additional Information
- Status version: develop 1.12.2

cammellos:
Fixes: https://github.com/status-im/status-react/issues/12006

I have changed the behavior of loading notifications:

Before it would always use the same function and check for `cursor !=
""`.
This worked on most cases, but sometimes the user would jump back to the
home view without doing any clean up, so cursor would be `""` and it
would not load any notification anymore.

Now the two functions are split in initial load, and load-more.

Initial load is called on component-did-mount so we don't have to worry
about where the user is coming from. We also avoid multiple firing of
the events by checking `loading?`, so it does not result in duplicated
notifications.

### DIFF
--- a/src/status_im/notifications_center/core.cljs
+++ b/src/status_im/notifications_center/core.cljs
@@ -61,8 +61,7 @@
 (fx/defn ensure-and-open-chat
   {:events [:ensure-and-open-chat]}
   [{:keys [db]} response-js]
-  {:db (update db :activity.center/notifications dissoc :cursor)
-   :dispatch-n [[:sanitize-messages-and-process-response response-js]
+  {:dispatch-n [[:sanitize-messages-and-process-response response-js]
                 [:chat.ui/navigate-to-chat (.-id (aget (.-chats response-js) 0))]]})
 
 (fx/defn dismiss-all-activity-center-notifications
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

flexsurfer:
@cammellos lint

Serhy:
Tested on Android 11, Android 8, iOS14 - initial issue fixed, no regressions found. 
Item from Activity Center is removed from activity center when it's Rejected/Declines (for Group Chat) or 1-1 chat is opened.

PR #12032, opened 2021-04-22
[wallet] Store pending transactions

rasom:
Pending transactions are stored in the database and shown on tx history screen after re-login.

status: ready


flexsurfer (approved):

Serhy:
Tested with iOS 14, Android 11, Android 8 (with non-keycard and keycard accounts) Pending transactions are stored after relogin (for tokens, ETH, Dapp TXs) and updated into 'Confirmed' state after confirmation received from network. 
In the case Pending transaction replaced by another tx, - we keep "forever" replaced transaction in Pending status. Agreed to fix later separately.

PR #12045, opened 2021-04-27
Accept balance update when in LTE network

Serhy:
Accept balance update when in LTE network


Serhy:
Covered in the https://github.com/status-im/status-react/pull/12057. Closing this one.

PR #12046, opened 2021-04-27
small chat perf impr

flexsurfer:
small chat perf impr

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -274,7 +274,8 @@
   [{db :db :as cofx} chat-id]
   (fx/merge cofx
             (close-chat)
-            #(assoc-in % [:db :current-chat-id] chat-id)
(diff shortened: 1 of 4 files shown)


src/status_im/chat/models.cljs
```diff
@@ -274,7 +274,8 @@
   [{db :db :as cofx} chat-id]
   (fx/merge cofx
             (close-chat)
-            #(assoc-in % [:db :current-chat-id] chat-id)
+            (fn [{:keys [db]}]
```
shivekkhurana: Is this a perf improvement or a cosmetic change?

shivekkhurana (approved):
⚡

src/status_im/chat/models.cljs
flexsurfer: cosmetic change , because there was a warning , it returned cofx and re-frame couldn't find fx for cofx fields

PR #12055, opened 2021-04-28
add factory reset option on new Keycard account

bitgamma:
adds the ability to factory reset the Keycard during account creation and mnemonic import.

Factory reset reinstalls the Keycard applet. The package, NDEF applet and Cash applet are untouched. This makes resetting the card quite fast and does not require to load a cap file. 

The card can be in any state (non-initialized, initialized, loaded with keys) before the reset and will always be non-initialized after.


src/status_im/keycard/real_keycard.cljs
```diff
@@ -107,6 +107,17 @@
...
+(defn factory-reset
+  [{:keys [on-success on-failure]}]
+  (.. status-keycard
+      (factoryReset)
+      (then (fn [response]
+              (let [info (-> response
+                             (js->clj :keywordize-keys true)
+                             (update :key-uid ethereum/normalized-hex))]
+                (on-success info))))
+      (catch on-failure)))
```
shivekkhurana, line 119: Does `catch` work without a `try` block?

shivekkhurana (approved):

src/status_im/keycard/real_keycard.cljs
bitgamma: Yes the same pattern is used throughout the file and the error handlers are known to work. I would leave it like this for consistency. Thanks for the review!

qoqobolo:
@bitgamma thank you, works great!

**One minor remark:**

The checkbox `Return card to factory settings` is unchecked when pressing `Begin setup` and opening the pop-up "Are you sure you want to perform...".
Would be nice to keep enabled and uncheck if User taps `No` in the pop-up.
If you think it makes sense and is easy to fix, ofc.

https://user-images.githubusercontent.com/67952253/117129215-b3540c80-ad9e-11eb-9a21-5040a25488bf.mp4

bitgamma:
@qoqobolo thanks for catching this! it is fixed now

qoqobolo:
@bitgamma thanks for the fix!
PR is ready to be merged.

Two sporadically reproducing issues were found in develop. One of them is related to the scenario of this PR but can be reproduced in another flow as well. I will log them separately.

PR #12080, opened 2021-05-05
Allow to enter bip39 password on account recovery

rasom:
BIP39 password input will be shown only after pressing on recovery screen for 2 seconds or longer. I also can hide it behind a feature flag if necessary so that it will not be accessible in the release version.

status: ready


cammellos (approved):

flexsurfer:
@rasom just wondering why is this needed?

PR #12110, opened 2021-05-12
make target for clearing xcode's derived data dir

rasom:
`make xcode-clean`: replacement for xcode's cmd+shift+k

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,11 @@ fdroid-nix-dir: ##@prepare Create /nix directory for F-Droid Vagrant builders
 
 fdroid-build-env: fdroid-max-watches fdroid-nix-dir ##@prepare Setup build environment for F-Droud build
 
+xcode-clean: SHELL := /bin/sh
+xcode-clean: XCODE_HOME := $(HOME)/Library/Developer/Xcode
+xcode-clean: ##@prepare Clean XCode derived data and archives
+	rm -fr $(XCODE_HOME)/DerivedData/StatusIm-* $(XCODE_HOME)/Archives/*/StatusIm*
+
 #----------------
 # Release builds
 #----------------


bitgamma (approved):

Makefile
```diff
@@ -290,6 +290,10 @@ android-clean: ##@prepare Clean Gradle state
 	git clean -dxf -f ./android/app/build; \
 	[[ -d android/.gradle ]] && cd android && ./gradlew clean
 
+xcode-clean: SHELL := /bin/sh
+xcode-clean: ##@prepare Clean xcode derivation folder
```
jakubgs: Should it be in the `prepare` category? Or `other`? If `prepare` then it should be in the `General targets` section.

Truth be told both `prepare` and `other` are kinda messy. Makefile would benefit from cleaning up those two categories. But for now that's fine. Just move it to the correct section.

Makefile
```diff
@@ -290,6 +290,10 @@ android-clean: ##@prepare Clean Gradle state
...
+	rm -fr $(HOME)/Library/Developer/Xcode/DerivedData/StatusIm-*
```
jakubgs: How about also removing `$(HOME)/Library/Developer/Xcode/Archives/*/StatusIm*`?

jakubgs: You could do something like:
```Makefile
xcode-clean: SHELL := /bin/sh
xcode-clean: XCODE_HOME := $(HOME)/Library/Developer/Xcode
xcode-clean: ##@prepare Clean XCode derived data and archives
	rm -fr $(XCODE_HOME)/DerivedData/StatusIm-* $(XCODE_HOME)/Archives/*/StatusIm*
```

Makefile
rasom: > Just move it to the correct section

It is in `other` now, by "correct section" do you mean `prepare `?

jakubgs: I mean position in the file. It's in:
https://github.com/status-im/status-react/blob/f1e081310d94b9d34ef25e52f58f516707d61753/Makefile#L279-L281
And most `prepare` targets are in:
https://github.com/status-im/status-react/blob/f1e081310d94b9d34ef25e52f58f516707d61753/Makefile#L111-L113

Makefile
rasom: I can add `Archives`. The target was added rather to help with local simulator build because currently there is no option to clear derived data folder without opening Xcode.

Makefile
rasom: Yep, but do you imply it should be under `prepare`? it is already under `other` as I understand

jakubgs: It's fine under `prepare` for now. We can do a cleanup of these categories in the future.

Makefile
jakubgs: Right, but it doesn't hurt, actually helps free up space, and does what the name of the target says.

Makefile
rasom: done

Makefile
rasom: done

PR #12111, opened 2021-05-12
ci: make e2e stage depends on publish param

churik:
Changing behavior of triggering `end-to-end-tests/job/status-app-nightly/` - it will be triggered only when PUBLISH parameter is enabled (by default only for `nightly`, but not for `manual` job)


jakubgs (approved):
Tested with a manual build and appears to work:

image

https://ci.status.im/job/status-react/job/manual/147/

churik:
Thank you @jakubgs !

PR #12115, opened 2021-05-17
Keycard account recovery

bitgamma:
Allow the creation of recovery/backup Keycard from the login interface. It requires the mnemonic and either a new or frozen Keycard (or actually a Keycard in any state, since you can factory reset).

Screenshots for localization


src/status_im/keycard/recovery.cljs
```diff
@@ -197,12 +198,27 @@
                   (navigation/navigate-to-cofx (if platform/android?
                                                  :notifications-settings :welcome) nil))))))
 
+(fx/defn return-to-keycard-login
+  [{:keys [db] :as cofx}]
+  (fx/merge cofx
+            {:db (-> db
+                     (assoc-in [:keycard :pin :enter-step] :login)
```
cammellos: you can probably do as well:
```
(update-in [:keycard :pin] assoc :enter-step :login
                                                   :status nil
                                                   :login [])
```
Which is a bit neater

cammellos (approved):

src/status_im/keycard/recovery.cljs
bitgamma: thanks for the review! fixed

churik:
awesome work @bitgamma !
Tested:
- keycard recovery for fresh, recovered, or restored multicaccount
- can login with 2+ cards
- can switch between devices in case it is paired for both
- can change PIN and updated PIN is required on another device if card is paired
- "Your card is full" popup
- "Seems this card is not paired" popup

Found one edge case, but don't think it is worth fixing.
So if user has 2 devices (A and B) with same multiaccount in case of pairing card on device B at attempt to recover same multiaccount on device A (with pairing code) user will get a popup with the suggestion to UNLOCK existing multiaccount as it exists on the device, but in fact, you can't unlock as you need to pair it again.
Anyway as user can pair it to device A - it is not a big deal. 

**Devices:**
- IPhone 12 Mini (IOS 14)

Issue #12083, opened 2021-05-05
remove PUK and Pairing code from keycard onboarding


guylouis:
**Problem:** when a user onboard with a keycard he needs to remember PIN, PUK, Pairing code, and seed. That's too much information and makes the process difficult and confusing for approval

**Solution studied:** we have dismissed the idea to derive puk and pairing from the seed, mostly for security reasons. See the issue here: https://github.com/status-im/status-react/issues/9064

**Solution selected:** we will remove puk and pairing from the onboarding, and allow the user to get these informations in the keycard settings. We also need to allow a user to factory reset his card when he doesn't have his PUK and needs it, because he's blocked. Or when he needs his pairing code and he's blocked (e.g lost phone).

**Description of the solution:**
https://notes.status.im/puk-and-pairing-removal%20from%20onboarding?view

**Sequential steps:**
- [x] factory reset on keycard onboarding 1st screen 
    - PR https://github.com/status-im/status-react/pull/12055
- [x] factory reset on login screen from a keycard multiaccount 
- [x] add copy to all screens where the user is blocked when he doesn't know his PUK saying "you don't know your PUK? learn how to factory reset your card here", redirecting to the flow to factory reset
- [x] add copy to all screens where the user is blocked when he doesn't know his Pairing code saying "you don't know your Pairing code? learn how to proceed here", redirecting to the flow to factory reset
- [x] remove PUK and Pairing from onboarding

PR #12127, opened 2021-05-20
add reset with mnemonic on frozen/blocked cards

bitgamma:
as part of #12083, allow to reset a blocked or frozen card using the mnemonic. Works from both logged-in and logged-out state.

Screenshots for localization. Of these, only ``keycard-is-frozen-factory-reset``  is new. The other ones are changes to existing keys, so we need to make sure they are properly updated.


cammellos (approved):

qoqobolo:
@bitgamma 

### ISSUE 1: "Backup successful" banner appears after the card is unfrozen in both logged-in and logged-out states.
Android, iOS

**_Steps:_**
- Recover a Keycard multiaccount
- Freeze the card entering invalid passcode in any flow (login, change 6-digit passcode, sign transaction, etc)
- Reset the card with mnemonic and complete the flow
 
#### Expected behavior:
`Keycard access is reset` banner

#### Actual behavior:
`Backup successful` banner

https://user-images.githubusercontent.com/67952253/118965758-4f6a4000-b969-11eb-83b3-9899bd545a97.mp4

bitgamma:
@qoqobolo this behavior is intentional and is because this flow, like the flow from #12115 is actually the creation of a backup card. After the flow is started, with the current implementation, there is no distinction between backup and reset. Do you think the text will mislead the user?

qoqobolo:
@bitgamma It seems to me that it would be preferable if the backup banner refers only to the process of creating a backup card, and for unlocking frozen card we have another one with the corresponding text. They are still two different flows, although from a technical point of view they are the same. And for unlocking with PUK we have the "Keycard access is reset" screen.
WDYT? If you have time and it isn't too hard to fix, would be nice to do this :)

bitgamma:
@qoqobolo fixed now, thanks for the input!

qoqobolo:
Thank you @bitgamma !

Another issue was found.
### ISSUE 2: If freeze the card while changing the PIN / PUK / pairing code, start reseting it with seed phrase, open `Enter seed` screen and go back, the `Enter PUK` screen will be opened but with an input field for only 6 digits.
After that, you can reset the Keycard only with the PUK code.

**_Steps:_**
1. Open Settings > Keycard > Change 6-digit code
2. Enter invalid PIN 3 times
3. Tap `Reset with mnemonic`
4. Tap back

#### Expected behavior: `Keycard is frozen` screen is opened

#### Actual behavior:

https://user-images.githubusercontent.com/67952253/119109008-14c3de80-ba21-11eb-9a09-31421273bd33.mp4

Also, if the card was frozen in signing transaction in the wallet view, and you repeat the steps from the above flow, it will not be possible to reset the Keycard using the mnemonic after returning back from the seed phrase input screen, only using the PUK code.

**_Steps:_**
1. Open Wallet
2. Send tx
3. Enter wrong PIN 3 times while signing the transaction
4. Tap `Reset with mnemonic`
5. Tap back
6. Tap `Your Keycard is frozen. Reset card access` at the top left

https://user-images.githubusercontent.com/67952253/119109494-813edd80-ba21-11eb-8e39-0fc47a3efc7c.mp4

bitgamma:
@qoqobolo thanks for the findings!

Issue 2: fixed by just returing to the original Enter PIN display, so you don't have the strange case where you are asked for a 6-digit PUK. To get the frozen card pop-up you will have to enter the PIN one more time. Making the pop-over re-appear again automatically requires keeping track of where we came from and I think for this edge case it is not worth the effort and risk of introducing issues somewhere else (the PIN screen is used in many different places so it is a critical component).

Issue 3: actually it already was possible to reset, but the label was "lost or frozen card" (like in the login screen). I fixed it to display "reset mnemonic" instead in this case so it becomes clear you can access that functionality from this screen.

qoqobolo:
@bitgamma thanks for your work and patience! Looks great for me now.
PR can be merged.

PR #12134, opened 2021-05-20
[wallet] Show cached balance before fetching latest

rasom:
status: ready


cammellos (approved):

src/status_im/wallet/core.cljs
```diff
@@ -51,16 +51,36 @@
...
+ (fn [addresses]
+   (json-rpc/call
+    {:method     "wallet_getCachedBalances"
+     :params     [addresses]
+     :on-success #(re-frame/dispatch [::set-cached-balances addresses %])
+     :on-error   #(re-frame/dispatch [::on-get-cached-balance-fail addresses %])})))
+
+(fx/defn get-balances
+  [{:keys [db]} addresses]
+  (if (nil? (get-in db [:wallet :accounts]))
```
flexsurfer: could you elaborate on why do we check :accounts ? and not like for example that balances are nil? are we sure that accounts will be always nil here ?

rasom: It only makes sense to retrieve cached balances when we log into the app and this specific check will be true only when called from `::initialize-wallet` https://github.com/status-im/status-react/blob/wallet%2Fcached-eth-balance/src/status_im/multiaccounts/login/core.cljs#L99. In this case yes, I'm sure it will be `nil` when needed. With balances it is a bit different because those might be `nil` for a newly created accounts but it wouldn't make sense to check cache for those.

But I agree it might look weird so I will rewrite it with an explicit request for cached balances from `::initialize-wallet` instead of making it implicit inside `wallet/update-balances`.

rasom: @flexsurfer fixed

flexsurfer (approved):

PR #12158, opened 2021-05-27
Allow importing script from db

cammellos:
This is a minor change to facilitate import a script from the db, so that if no networks are saved in the db, we just default to the networks we have in status-react.


flexsurfer (approved):

Serhy:
Thanks a log @cammellos ! 
https://github.com/status-im/status-react/pull/12158#issuecomment-849769858 not this PR issue (False Failure will be fixed separately)
I can successfully generate DB with data and import in this PR.

PR #12159, opened 2021-05-27
set depreacted metamask fields

cammellos:
sets a couple of metamask deprecated fields that are still used by some dapps.


flexsurfer (approved):

churik:
@cammellos on ropsten can't open simpledapp.eth.
Error: 
image

Serhy:
Failed tests https://github.com/status-im/status-react/pull/12159#issuecomment-852914440 are false fails, I rechecked them manually and it's all working. We'll look at those tests separately.
simpledapp.eth opens fine now.

PR #12163, opened 2021-05-28
🗝️ Password reset UI

shivekkhurana:
### Summary
Implement password reset functionality:
- UI: https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8101%3A3945
- Companion PR: https://github.com/status-im/status-go/pull/2223

### Testing notes
1. Test if reset password works
   - Go to privacy > reset password
   - Enter info
   - Logout
   - You should not be able to login with old password
   - You should be able to login with new password

2. Test if delete account works with the new password
   - After resetting password, goto the delete profile screen
   - You should not be able to delete profile with old password
   - You should be able to delete profile with the new password

3. Test if password reset throws errors if wrong `current password` is supplied
4. Test if the app works fine just after password is reset :
  - Password reset doesn't logs the user out
  - But I don't know if it has any side-effects
  - Try to create chats or change settings (perform any action that writes to DB) and make sure it works 

#### Platforms

- Android
- iOS

- Login
- Profile deletion
- Keycard migration
- Anything that makes a user enter a password (except for registration)

##### Functional
- account recovery
- new account
- user profile updates

### Steps to test
- Goto profile
- Privacy settings
- Reset password form


modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -1481,5 +1481,22 @@ public void execute(NativeViewHierarchyManager nativeViewHierarchyManager) {
             }
         });
     }
+
+    @ReactMethod
+    public void changeDatabasePassword(final String keyUID, final String password, final String newPassword, final Callback callback) {
```
flexsurfer: is it only database? or we re-encrypt keys as well?

src/status_im/ui/screens/reset_password/views.cljs
```diff
@@ -0,0 +1,82 @@
...
+                                               id
+                                               (security/mask-data %)])
+     :secure-text-entry   true
+     :error               (when-let [error (get errors id)]
+                            (if (keyword? error)
+                              (i18n/label error)
+                              error))}]))
+
+(defview reset-password []
+  (letsubs [form-vals [:multiaccount/reset-password-form-vals]
```
flexsurfer: if these subs used only togather, might be its better to have only one subscription ? :multiaccount/reset-password with map

src/status_im/ui/screens/reset_password/views.cljs
```diff
@@ -0,0 +1,82 @@
...
+            errors [:multiaccount/reset-password-errors]
+            next-enabled? [:multiaccount/reset-password-next-enabled?]]
+    (let [on-submit #(re-frame/dispatch [::reset-password/reset form-vals])]
+      [react/keyboard-avoiding-view {:flex 1}
+       [topbar/topbar {:title (i18n/label :t/reset-password)}]
+       [react/view {:style {:flex               1
+                            :padding-horizontal 16
+                            :padding-vertical   16
+                            :justify-content    :space-between}}
+        [react/view
```
flexsurfer: why do you need react/view here ?

src/status_im/ui/screens/reset_password/views.cljs
```diff
@@ -0,0 +1,82 @@
...
+         [input-field :current-password errors on-submit true]
+         [input-field :new-password errors on-submit]
+         [input-field :confirm-new-password errors on-submit]]
+        [react/view
```
flexsurfer: why do you need react/view here ? we should try to use less views as possible

src/status_im/ui/screens/reset_password/views.cljs
```diff
@@ -0,0 +1,80 @@
...
+            [status-im.multiaccounts.reset-password.core :as reset-password]
+            [status-im.utils.security :as security]
+            [status-im.ui.components.toolbar :as toolbar])
+  (:require-macros [status-im.utils.views :refer [defview letsubs]]))
+
+(defn input-field
+  ([id errors on-submit] (input-field id errors on-submit false))
+  ([id errors on-submit focus?]
+   [quo/text-input
+    {:placeholder         (i18n/label (keyword :t id))
```
flexsurfer: i believe we use :t only for finding usage, so you don't need it here

src/status_im/ui/screens/popover/views.cljs
```diff
@@ -169,6 +170,9 @@
                    (= :transfer-multiaccount-unknown-error view)
                    [multiaccounts.key-storage/unknown-error-popover]
 
+                   (= :password-reset-success view)
```
flexsurfer, line 173: imo its more convenient to provide a view directly in :show-popover event

src/status_im/subs.cljs
```diff
@@ -2615,3 +2617,15 @@
        (and
         (= network-type "cellular")
         syncing-on-mobile-network?))))
+
+;; RESET PASSWORD
+(re-frame/reg-sub
+ :multiaccount/reset-password-next-enabled?
+ :<- [:multiaccount/reset-password-form-vals]
+ (fn [{:keys [current-password new-password confirm-new-password]}]
+   (and (pos? (count current-password))
```
flexsurfer: `blank?` ?

flexsurfer (approved):

flexsurfer:
great job! it's a really important feature!

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
shivekkhurana: We re-encrypt keys as well. Key re-encryption is not exposed though, it is handled by Change DB password function on go side: https://github.com/status-im/status-go/pull/2223/files#diff-3270068c140c83704dd9e117c7b423114a7f6b57307c42c1e68d1c210680fc52R467

flexsurfer: yes so its better to rename function, something like re-encrypt-keys-and-database ?

src/status_im/ui/screens/reset_password/views.cljs
flexsurfer: but it's up to you of course

src/status_im/subs.cljs
shivekkhurana: `(clojure.string/blank? "________") ;=> true` but six spaces or more is a valid password. An unsafe password, but still valid. 

(_ = single space " ")

src/status_im/ui/screens/popover/views.cljs
shivekkhurana: I didn't realise that. Definitely a better way.

src/status_im/ui/screens/reset_password/views.cljs
shivekkhurana: This view is required for flexbox positioning.

src/status_im/ui/screens/reset_password/views.cljs
shivekkhurana: same, as above. For flexbox positioning.

src/status_im/ui/screens/reset_password/views.cljs
flexsurfer: required by whom?

shivekkhurana: The UI uses `justify-content: space-between`. The positions block level content at equidistant positions like the image below, but vertically.

justify-content-space-between

Now this page has three `react/views` that we want to be justified with space between:
-  form fields
- info message text and 
- bottom toolbar

Hence we use two `react/views` plus `toolbar` which has a `react/view` internally to achieve the desired layout.

src/status_im/ui/screens/popover/views.cljs
shivekkhurana: In this case, it's easier to use an id instead of the component itself. The success popover is a result of many effects and events. Passing the real view will be more work than this implementation.

shivekkhurana:
Have hidden this behind a feature flag. The reset password option will ~only be visible in DEV mode~ not be visible until Go side is merged.

Edit:
Can't be merged before Go because build breaks down.

qoqobolo:
@shivekkhurana Hey, thanks for this feature!

If I understand correctly, PR cannot be tested yet.
Could you move it to `Contributor` for now and ping us once it's ready?
Thanks.

shivekkhurana:
@qoqobolo The Go side was merged, it can be tested now.

Serhy:
@shivekkhurana  could you rebase against latest develop please? (RNN update merged recently)

shivekkhurana:
@qoqobolo I have created this other PR #12245 to track and manage changes and issues. 

The current PR is hidden behind a feature flag and will only be visible in Development mode.

PR #12170, opened 2021-05-31
handle hash-message failure correctly

### LINKED ISSUE #11706: signing typed data with wrong chain ID fails without proper error message
# Bug Report

## Problem

When signing a typed message, the chain ID is part of the message being signed. If this differs from the currently selected network, hashing the message fails (correctly). However this failure is not caught/shown anywhere and the user is still prompted to tap their card. At this point the process stalls, with the NFC prompt showing "Processing" forever. The user can only cancel, but likely will have no idea why this happens.

If the chain id matches the current network everything works normally.

### Additional Information

Test on release 1.10.1, develop branch from 02.02.2021 and PR11540. PR11540 tested on iOS as well.

bitgamma:
Fixes #11706.

now when hash-message fails (for any reason) signing is cancelled and an error popup is shown. The current behaviour was prompting the user to enter the PIN, tap the card and enter endless processing because an empty hash was being signed.

### DIFF
--- a/src/status_im/signing/keycard.cljs
+++ b/src/status_im/signing/keycard.cljs
@@ -3,6 +3,7 @@
             [status-im.ethereum.abi-spec :as abi-spec]
             [status-im.native-module.core :as status]
             [status-im.utils.fx :as fx]
+            [status-im.i18n.i18n :as i18n]
             [status-im.utils.types :as types]
             [taoensso.timbre :as log]))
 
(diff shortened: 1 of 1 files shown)


cammellos (approved):

bitgamma:
@qoqobolo I have tested that the PR fixes the issue and I don't think it should have any side effect. Do you want to perform some additional tests or do you think it can be merged?

qoqobolo:
@bitgamma I think your testing is enough for this issue, so PR can be merged.
Thank you.

PR #12189, opened 2021-06-03
Validate community channel names

### LINKED ISSUE #12010: Naming validation for communities
As a user
I want to limit the number of characters available when creating community names/channel names/channel categories
So that the UI is not cluttered

### Description

Community names should allow anything and just limit to 30 characters

Channel names should allow anything besides:
 - spaces, they should be replaced as the user types with dashes `-`
 - Uppercase letter, they should be replaced by the lowercase counterpart as the user types
 - limit to 30 characters

If categories are also implemented, they should have the same limitations as channel names, if they are not implemented, this is out of scope for this taks.

cammellos:
Fixes: #12010

### DIFF
--- a/src/status_im/communities/core.cljs
+++ b/src/status_im/communities/core.cljs
@@ -21,6 +21,16 @@
        "/c/"
        community-id))
 
+(defn sanitize-name
+  "sanitize-name replace uppercase with lower case and
+  spaces with dashes"
+  [n]
+  (->
+   n
+   str
+   string/lower-case
+   (string/replace " " "-")))
+
 (def featured
   [{:name "Status"
     :id constants/status-community-id}])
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

flexsurfer:
still relevant ?

shivekkhurana (approved):

flexsurfer:
@cammellos  still relevant ?

du82:
It's unnecessary and a bad idea. Why limit what people can name community channels? Lets not be discord where it's all lowercase and with dashes instead of spaces.

If a community thinks having crazy unicode characters is a good idea, let them. Don't limit what everyone can do because a few people think it doesn't look good.

flexsurfer:
i guess we can close it for now

PR #12199, opened 2021-06-04
improve Keycard credential handling for e2e

### LINKED ISSUE #10827: Update framework for keycard to support reset passcode flow
# Problem

Currently there is no option to add automation tests for:
- wrong PIN error
- frozen card flow
- blocked card flow

## Acceptance Criteria
- see "Wrong passcode" error
- can finish "Reset passcode" when card is frozen
- can get "Blocked card"  screen

bitgamma:
Fixes #10827.

also adds card reset, change pin/puk/pairing to simulated keycard

### DIFF
--- a/src/status_im/keycard/simulated_keycard.cljs
+++ b/src/status_im/keycard/simulated_keycard.cljs
@@ -12,6 +12,11 @@
             [taoensso.timbre :as log]
             [status-im.multiaccounts.create.core :as multiaccounts.create]))
 
+(def kk1-password "000000")
+(def default-pin "111111")
+(def default-puk "000000000000")
+(def account-password (ethereum/sha3 "no password"))
+
 (def initial-state
   {:card-connected?  false
    :nfc-started? false
@@ -27,7 +32,7 @@
 (defn connect-selected-card []
   (swap! state assoc :application-info
          {:free-pairing-slots     5
-          :app-version            "2.2"
+          :app-version            "3.0"
           :secure-channel-pub-key "04c3071768912a515c00aeab7ceb8a5bfda91d036f4a4e60b7944cee3ca7fb67b6d118e8df1e2480b87fd636c6615253245bbbc93a6a407f155f2c58f76c96ef0e",
           :instance-uid           "1b360b10a9a68b7d494e8f059059f118"
           :paired?                true
@@ -36,6 +41,7 @@
           :pin-retry-counter      3
           :puk-retry-counter      5
           :key-uid                (get-in @re-frame.db/app-db [:multiaccounts/login :key-uid])})
+  (swap! state assoc :pin default-pin :puk default-puk :password kk1-password)
   (connect-card))
 
 (def initialization (atom false))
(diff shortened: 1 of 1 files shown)


bitgamma:
the E2E failure is likely to be caused by my changes, will fix it

churik:
@bitgamma somthing is wrong with migrating to keycard account flow in framework.
When I'm trying to reproduce it manually on emulator, I've got after relogin: 
image
Steps:
- restore multiaccount from seed phrase
- migrate multiaccount to keycard (using "connect-card" button, PIN `121212`)
- try to relogin

churik:
@bitgamma somthing is wrong with relogin after creating kk multiaccount.
When I'm trying to reproduce it manually on emulator, I've got after relogin: 
image
Steps:
- restore multiaccount from seed phrase, using keycard storage  (using "connect-card" button, trying PIN `121212`, `111111`)
- try to relogin  (using "connect-card" button,  trying PIN `121212`,`111111`)

bitgamma:
Thanks for the report, from the e2e test I hadn't fully understood the problem. Will fix as I am back from vacation 🏖️

bitgamma:
@churik fixed! BTW the default PIN when you press the "connect-selected-card" and "connect-pairing-cad" button is 111111. When using the "connect-card" button the card is empty so it doesn't have any PIN. It will accept whatever PIN you give on initialization.

churik:
@bitgamma as big RNN PR was merged, a lot can be changed. 
I suggest to wait until https://github.com/status-im/status-react/pull/12241 and https://github.com/status-im/status-react/pull/12226 merge, rebase PR and I'll test it one more time if no objections.

Thank you!

bitgamma:
@churik rebased to latest develop

churik:
@bitgamma 
after changing PIN I can't check that it is actually changed: at attempt to use it for exmple, for signing, got error:
image

Steps: 
- create new multiaccount
- change PIN
- try to sign message in Dapp to verify PIN is changed

bitgamma:
the PIN is currently used when creating the account so changing it makes it mismatch with the keys. I can probably fix it by using a fixed value instead of the PIN there.

however, is it really needed? I enabled the change pin/puk/pairing in order to be able to test these flows in e2e. Checking that it has actually been changed only tests the implementation of the simulated keycard which probably isn't what we want, wdyt?

shivekkhurana (approved):

churik:
> however, is it really needed? I enabled the change pin/puk/pairing in order to be able to test these flows in e2e. Checking that it has actually been changed only tests the implementation of the simulated keycard which probably isn't what we want, wdyt?

well depends on.
If I test it manually I always check that it is actually changed.
If it really needs to be "emulated" for working in flow, I'd rather define necessity of this by possibility of finding issues.
So basically:
1) if it is "emulated" and has a lot of differences with real keycard - no need,as we anyway should receck it manually, so it is not reducing the scope of manual tests
2) if you think that "tuning" it to use updated PIN is  more work than potential benefit - let's leave it as it is

bitgamma:
@churik I fixed it anyway. However to answer your question, the simulated keycard fully skips the native module and of course the actual hardware, so it is indeed emulated. To have a more faithful emulation we should move the simulated keycard from status-react to the native module, but that'd be quite a large task while still not having 100% accuracy. 

I think the e2e tests for Keycard with the current emulation layer are good for catching navigation, flows, ui/ux issues but they cannot really catch issue in the low-level integration layer. However, since the low-level integration (native module mainly) does not change as often the scope of manual tests can probably be reduced a lot in most cases. Maybe it is a good idea to mark with a special tag PR where the native Keycard module has been changed. Also during release testing at least a few "happy paths" should be tested with real hardware.

churik:
Works as magic. 
Thank you a lot!
A bunch of new e2e now are ready to be automated.

PR #12221, opened 2021-06-14
eip1559 changes

rasom:
what's implemented:
- `EIP1559_ENABLED` flag added to `.env`
- activation block for each network can be specified here `status-im.signing.eip1559` ns
- If activation block is mined and `EIP1559_ENABLED=1` then updated ui is shown. Later we can remove both checks. The block is checked with `eth_getBlockByNumber` request, which means if we are offline the ui will default to legacy tx signing.
- there are two new bottom sheets and on popover: 1) priority fee slider 2) custom fee settings 3) fee warnings screen
- on priority fee slider we allow user to change priority fee, currently it also updated maxFeePerGas setting `(maxFeePerGas = maxFeePerGas - oldPriorityFee + newPriorityFee)`. 
- `custom fee settings` allows to change `gasPrice`, `maxPriorityFeePerGas` and `maxFeePerGas`. If values are off there is some validation in place. Currently `maxPriorityFeePerGas` suggestions are hardcoded though, it should be changed when ethereum API for this stuff will be ready.

https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/client-integration-testnets/calaveras.md

status: wip


rasom:
Screenshot 2021-06-17 at 17 01 04
Screenshot 2021-06-17 at 17 00 56

PR #12235, opened 2021-06-18
[#12162] Allow user to onboard to a website after install

### LINKED ISSUE #12162: Allow user to onboard to a website after install
As a user
I want to be redirected to a website after I install the app
So I can have a better onboarding experience

### Description

After installing, if installed through a particular link, we should be able to redirect a user to a specific website opening the status-browser.
We had similar mechanism with chats, and this will involve changing the referral service and create a new type of links likely.

flexsurfer:
fixes #12162


shivekkhurana (approved):

Serhy:
I've tested this with 'hermez' referral (generated quick referral link with referral_ID), but nothing happens when I navigate to this PR build via link (all other referrals working fine, but not Dapp referral).

PR #12226, opened 2021-06-16
rnn leftovers

flexsurfer:
fixes: #12224
fixes: #12225
fixes: #12236

PR #12241, opened 2021-06-21
RNN e2e aligned

churik:
more e2e fixes after RN upgrade


churik:
Below will be result for failed e2e in nightly + this fix on build from #12226

Serhy (approved):

PR #11986, opened 2021-04-12
Dropping support for ARM32 iOS devices

jakubgs:
With upgrade to Go and Gomobile we now drop support for `darwin/arm32` builds.
This means all devices like iPhone 5 or older.

PR #12265, opened 2021-06-30
nix: fix Maven repositories, refactor go packages

jakubgs:
I've noticed issues with the https://repository.sonatype.org/ Maven repository we were using.
They appear to have purged their storage of a lot of packages, including our own for example:
https://repository.sonatype.org/content/groups/sonatype-public-grid/com/github/status-im/

For this reason I've release a [new version](https://github.com/status-im/go-maven-resolver/releases/tag/v1.1.1) of `go-maven-resolver` that includes two more Maven repos:
- https://plugins.gradle.org/m2
- https://jitpack.io

This also simplifies the `gomobile` package, and drops the need for `deps.nix` file.
It also slims down the #11986 Go upgrade PR for easier debugging.


flexsurfer (approved):

jakubgs:
I tested both Android and iPhone briefly and I saw no issues so I don't think there's any need for QA team to waste time on this.

PR #12272, opened 2021-07-05
properly sync command messages

### LINKED ISSUE #10994: Creating "myself" 1-1 chat in Activity center on a synchronized device after sending address request in 1-1 chat
# Bug Report

## Problem

If send ETH to any other user from Device 1, then on paired Device 2 will be created a 1-1 chat with “myself” (with the same account from which the Address request was sent). 

#### Expected behavior

"Myself" 1-1 chat on a synchronized device isn't created.

#### Actual behavior

ezgif com-resize (2)

### Reproduction

Device 1:
- Sync with Device 2
- Open 1-1 chat with any other user -> Send transaction
- Enter valid ETH amount and send Address request
- Check chats on Device 2

### Additional information

- Status version: Nightly 24/07
- Operating System: Android 8, 10

bitgamma:
Fixes #10994


bitgamma:
@qoqobolo @churik could you please help with testing? Needs to be tested if all transaction related messages are properly synchronized (request/accept/decline).

Both devices must have the fixed (this) version to work. If the device sending the original message is older, the message will be discarded (instead of creating a 1-to-1 chat to yourself like before) when syncing. If the message is sent using the new version but other synced client is the old one, the 1-to-1 chat to yourself is created (only on the old one).

The influence is only on syncing, the two parties of the transaction can have any version and it will work as before, so it doesn't break compatibility.

qoqobolo:
@bitgamma thanks for this fix, works great now!
PR can be merged after review (if required).

cammellos (approved):
Don't forget to create a new release on status go and point to that branch, nice work!

PR #12315, opened 2021-07-15
Upgrade geth and add eip 1559 code behind a toggle

cammellos:

### Upgrade geth to 1.10.4 

This is a major upgrade as a lot of things have changed, it's necessary to enable eip 1559 and be able to send transactions with the new parameters

### Disable 32 bit builds on ios

This is necessary as geth uses a newer version of some system libraries which already does not support 32 bits architectures on ios devices

### Add code for eip 1559 toggled

This is all behind a feature flag, so does not need to be tested, I have tested it myself and I am able to send transactions correctly, though there are a few issues to fix, which will be addressed in a separate PR when we will enable it.

## Testing

Unfortunately given the large extend of the changes, this will require a full test, as any part of the app can be impacted.
The most likely that will be impacted is: dapps, wallet and mailservers connectivity.


src/status_im/node/core.cljs
```diff
@@ -19,10 +19,14 @@
 (defn- add-log-level [config log-level]
   (if (empty? log-level)
     (assoc config
-           :LogLevel "ERROR"
-           :LogEnabled false)
+           :MaxPeers 5
+           :MaxPendingPeers 5
+           :LogLevel "DEBUG"
```
flexsurfer: is this only for testing ? we need to revert that when merging?

src/status_im/signing/eip1559.cljs
```diff
@@ -0,0 +1,64 @@
+(ns status-im.signing.eip1559
+  (:require [re-frame.core :as re-frame]
+            [status-im.ethereum.json-rpc :as json-rpc]
+            [status-im.utils.config :as config]
+            [status-im.utils.money :as money]))
+
+(def activation-blocks
+  {"123" (money/bignumber 0)})
```
flexsurfer: whats that ?

flexsurfer (approved):

src/status_im/node/core.cljs
cammellos: oh yes, sorry, I missed it

src/status_im/signing/eip1559.cljs
cammellos: it's the chain id and the block that london will be enabled on:
Say 
```
{"1" 200008}
```
Means that on mainnet it will be activated on block 200008 
I can add a comment to clarify

jakubgs (approved):
Nix changes look to be copied exactly right, so I see no issues.

I guess I wonder if `debug: try dropping all armv7 and armv7s architectures` is even necessary, I added it for a test, maybe it isn't. But if it is I'd squash it into the commit that does the Nix changes.

Serhy:
@cammellos we have something wrong in behavior to fetch offline messages: e.g. there is endless spinner when joining public chat (even after app restart) - each mailserver gives the same behavior - endless spinner and at some pint 'Retry to connect' dialog appears. Both iOS and Android affected. 

Attaching logs from one account on Android 
[Status.log](https://github.com/status-im/status-react/files/6830382/Status.log):
[geth.log](https://github.com/status-im/status-react/files/6830381/geth.log)

Serhy:
@cammellos we tested latest builds on iOS and Android with various scenarios (messaging in 1-1/public/group chats, offline inbox, Push Notifications, transactions with EHT/tokens in chat/wallet/Dapp). 

This one looks good for merge, thanks @qoqobolo for help with testing!

jakubgs:
@cammellos remember to squash that `debug: ...` commit into the nix one.

Edit: Or just all into one, whatever you prefer.

Issue #12336, opened 2021-07-22
Improve UX to migrate normal account to keycard account


guylouis:
**Problem:**
when a normal (his keys are stored on the phone) user wants to migrate his account to a keycard then, he is been taken through this flow
https://www.figma.com/file/Z2T0QW0TgIrJEeTjgBidAG/Untitled?node-id=0%3A1

![image](https://user-images.githubusercontent.com/11461437/126616167-853474d8-6882-4bc4-824c-426454074614.png)

Primarily, this flow has been considered
- hard to locate for users that want to migrate 
- complex (naming and copy are not obvious)
Also
- the choose storage screen is not providing any choice to the user so it could be removed or simplified to an information screen

**How to improve this?**
This flow needs design eyes to simplify the user experience 
An idea would be to add in the '...' screen an entry which would be 'move account to keycard' and which by default would keep the database, it would thus be much simpler for the user to understand and proceed. MOre advanced users could still go through 'manage keys and storage'

cc fyi @bitgamma @churik

churik:
I believe that can be closed after https://github.com/status-im/status-react/pull/12395
 @guylouis please reopne if still relevant

bitgamma:
I think it would be nice to reopen this because this was about simplyfing the flow (both from logged in and logged out state) and making it easier to find

churik:
closing due to re-design, label `recheck` added for future re-consideration

PR #12395, opened 2021-07-29
trigger migration to Keycard from logged in state

### LINKED ISSUE #11778: Enable migration to keycard from signed-in state
relates to #11431 "Migrate account to a regular keycard from logged out "Manage keys and storage" and epic #11237

when solving #11431 we sticked to the case where the user starts the flow fro a signed out state, and begins by chosing the multiaccount he wants to migrate in the list of multiaccounts 

this issue is about proposing the same migration flow from a signed-in state 

Design 
https://www.figma.com/file/dEIljL7UPbXgsZUA0Q4qlE5E/Onboarding?node-id=8218%3A8318

bitgamma:
Closes #11778. It logs you out and redirects to the key management UI since it is the safest way to actually perform the migration.

The entry point is from Profile -> Privacy and Settings -> Manage Keys and Storage.

When we do #12336 the menu label will be renamed to whatever we decide it should be called (Migrate account to Keycard I guess)

### DIFF
--- a/src/status_im/multiaccounts/key_storage/core.cljs
+++ b/src/status_im/multiaccounts/key_storage/core.cljs
@@ -5,6 +5,7 @@
             [status-im.multiaccounts.core :as multiaccounts]
             [status-im.multiaccounts.recover.core :as multiaccounts.recover]
             [status-im.multiaccounts.model :as multiaccounts.model]
+            [status-im.multiaccounts.logout.core :as multiaccounts.logout]
             [status-im.native-module.core :as native-module]
             [status-im.navigation :as navigation]
             [status-im.popover.core :as popover]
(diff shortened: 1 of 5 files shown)


src/status_im/multiaccounts/key_storage/core.cljs
```diff
@@ -240,6 +241,23 @@ We don't need to take the exact steps, just set the required state and redirect
...
+
+(fx/defn logout-and-goto-key-storage
+  {:events [::logout-and-goto-key-storage]}
+  [_]
+  {:ui/show-confirmation
+   {:title               (i18n/label :t/logout-title)
+    :content             (i18n/label :t/logout-are-you-sure)
+    :confirm-button-text (i18n/label :t/logout)
+    :on-accept           #(re-frame/dispatch [::confirm-logout-and-goto-key-storage])
+    :on-cancel           nil}})
```
shivekkhurana, line 259: `nil` is not callable, this might throw an error

flexsurfer (approved):

shivekkhurana (commented):
I don't know if `nil` in `on-cancel` will throw an error or not. Everything else seems fine.

src/status_im/multiaccounts/key_storage/core.cljs
bitgamma: thanks! Seems to work fine, pressing the cancel button hides the pop-up and no error is displayed (even in logs)

flexsurfer: it's the same as you just do not add `:on-cancel` at all in the map

churik:
Looks good to me! The only thing I could suggest to improve: maybe change the copy for "Log out" popup for this case to explicitly pointing that Log out is required for migration.
Currently, it looks like the default "Log out?" popup
IMAGE 2021-07-30 17:33:52
Appreciate your opinion here @errorists 

However, don't think it is a blocker for PR.

bitgamma:
@churik Thanks, I changed to message to make it clearer. I think it will be changed again when we do #12336 but at least for now it should be ok

PR #12403, opened 2021-07-31
Add quick way to permanently hide android background service notification

ismaeldm95:
### Summary

As a user of status on android, I want to be able to receive notifications without having to constantly see the sticky notification for the background service. I'd like an easy way to hide this notification permanently.

Since Android 8.0 users can hide notifications for specific channels. A tap on the background service notification now takes the user to the background service notification channel settings so they can hide it permanently from there. Text in the notification also reflects what's going to happen when tapping it

Screenshot 2021-07-31 at 12 54 17

Screenshot 2021-07-31 at 12 55 12

### Review notes

- Review the strings that have been modified in strings.xml

#### Platforms

- Android

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- account recovery
- new account
- user profile updates
- networks
- mailservers

### Steps to test

- Open Status on Android 8.0 or higher
- Tap on the Status Background service notification
- Disable channel notification
- Receive a message to check that notifications for that chat are still working

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/ForegroundService.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/pushnotifications/ForegroundService.java
@@ -8,6 +8,7 @@ import android.app.Notification;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
+import android.provider.Settings;
 
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.NotificationManagerCompat;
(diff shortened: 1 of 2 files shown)


flexsurfer:
thank you @ismaeldm95

churik:
Tested on Android 10 - woeks like  a charm, background notifications setting is preserved in app between different users, all type of PNs are working after hiding background notification service notification.

Thank you @ismaeldm95 - awesome job!

shivekkhurana (approved):
Looks good.

shivekkhurana:
@ismaeldm95 Thank you for the contribution 🙌🏼

PR #12407, opened 2021-08-02
[Fixes: #12393] Allow to enable transaction management

### LINKED ISSUE #12393: Re-enable edit/cancel transaction
We disabled edit/cancel transaction in order to polish the UX, but currently we are at a stage where user are not able to cancel transactions through status and they have to use other wallets to achieve the same.

We should add a toggle in `advanced` where a user can enable and display the cancel/edit transaction feature, so that at least we give them an option, while we work in improving the UX.

cammellos:
Fixes: #12393 

Adds a setting to enable management transactions, as it was before. I have briefly checked it works the cancel/increase gas, but this only makes sure that there's a setting.

The setting is per installation, not per account.

status: ready

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -50,6 +50,14 @@
         callback)))))
 
 (re-frame/reg-fx
+ ::initialize-transactions-management-enabled
+ (fn []
+   (let [callback #(re-frame/dispatch [:multiaccounts.ui/switch-transactions-management-enabled %])]
+     (async-storage/get-item
+      :transactions-management-enabled?
+      callback))))
+
+(re-frame/reg-fx
  ::login
  (fn [[key-uid account-data hashed-password]]
    (status/login key-uid account-data hashed-password)))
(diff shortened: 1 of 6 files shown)


flexsurfer (approved):

churik:
@cammellos 
several notes: 
1) missing translation:
IMAGE 2021-08-04 13:39:47
2) cancelled / increased transaction will be replaced by initial one after relogin, but after confirmation all initial transaction will be shown (which where replaced with new ones) with options "Increase gas" and "Cancel":
Monosnap 2021-08-04 13-43-03

Up to you what should be fixed.

churik:
@cammellos 
is it ready to be re-tested?

cammellos:
@churik I have changed the label to:
`Transaction management (alpha)`, I was going to check though that it fits on mobile (it fits the emulator)

churik:
IMAGE 2021-08-06 15:01:57
Fits on IPhone mini, so if we're OK with 2 issue for now, we can merge. 

Also on release we need to recheck it with EIP-1559 - not sure how it will work

cammellos:
@churik we will address them once is out of alpha, thanks for testing!

PR #12409, opened 2021-08-02
update react-native-webview for better SSL handling

### LINKED ISSUE #12073: 'SSL error: The date of the certificate is invalid' when opening CryptoKitties on Android
# Bug Report

## Problem

Screenshot 2021-04-30 at 16 04 38

### Reproduction

- Open Status
- Open Browser tab
- Enter URL cryptokitties.co or navigate to `DApps` > `Collectibles` >` CryptoKitties` > `Open`

### Additional Information

- Status version: Nightly 30/04
- Operating System: Android

bitgamma:
Fixes #12073.

update react-native-webview to release v11.3.0 + permission handling patch. This is the last release where our patch applies cleanly and seems to work fine. The improvements to permission handling introduced by newer versions of react-native-webview are not a replacement for our patch, so should we have a reason for futher upgrades additional work will be needed.

when testing please check carefully for regressions in the browser since a core component has been upgraded. Note that entering https://cryptokitties.co does not work with this version either. It doesn't work in any other browser I tested as well so something is definitely misconfigured on their side.

However, http://cryptokitties.co and https://www.cryptokitties.co both work fine. Also accessing from dap.ps is OK.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "react-native-status-keycard": "git+https://github.com/status-im/react-native-status-keycard.git#v2.5.35",
     "react-native-svg": "^9.8.4",
     "react-native-touch-id": "^4.4.1",
-    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v10.9.2",
+    "react-native-webview": "git+https://github.com/status-im/react-native-webview.git#v11.3.0-status",
     "tdigest": "^0.1.1",
     "web3-utils": "^1.2.1"
   },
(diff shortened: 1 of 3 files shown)


flexsurfer (approved):

churik:
Tested all cases related to browser, that we have in out test suite (Android and IOS): https://ethstatus.testrail.net/index.php?/suites/view/48&group_by=cases:section_id&group_order=asc&display_deleted_cases=0

Also retested browser issues - all found still can be reproduced here.
Found one issue, it also reproducible on nightly, so it is not a regression.
Thank you @bitgamma, PR can be merged

PR #12420, opened 2021-08-04
emit url event when recreating Activity with link

### LINKED ISSUE #12382: When app is closed, no redirect from PN to 1-1 chat
# Bug Report

## Problem 
Blocker for 3 e2e.
Currently if app was closed (not shown inrecent apps), no redirect to 1-1 chat on tap on push notifications.
Recent regression, likely caused by https://github.com/status-im/status-react/pull/12341 

#### Expected behavior
will be redirected to 1-1 chat

#### Actual behavior
redirected to the home screen

### Reproduction
**User A has Android device with enabled PNs**

1) User A: add User B to contacts
2) User A: close app, remove it from recent
3) User B: send message to user A
4) User A: tap on PN

### Additional Information

- Status version: nightly 27/07/21
cc @flexsurfer

bitgamma:
Fixes #12382.

the solution is a workaround to the fact that the linking module from react-native (at least in conjuction with react-native-navigation) does not emit the url event when the app is in background and the MainActivity has been destroyed (which happens, for example, when the app is removed from recent app list). The issue is also mentioned here: https://github.com/react-navigation/react-navigation/issues/9154.

Also as part of the fix, the link is stored until all events during the transition between background and active is completed. This is because otherwise it would sporadically (but quite often when recreating the Activity) fail because :chat-stack was not yet ready. This part of the fix should probably remain even if the workaround (the changes to MainActivity) becomes redundant

### DIFF
--- a/android/app/src/main/java/im/status/ethereum/MainActivity.java
+++ b/android/app/src/main/java/im/status/ethereum/MainActivity.java
@@ -23,6 +23,9 @@ import android.provider.Settings;
 import android.os.Bundle;
 import com.facebook.react.ReactActivityDelegate;
 import com.facebook.react.ReactRootView;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.WritableMap;
 import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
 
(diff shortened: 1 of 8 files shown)


cammellos (approved):

churik:
awesome work @bitgamma !
thank you. 
Ready to be merged, please mind my commit while squashing

churik (approved):

PR #12444, opened 2021-08-13
[#12161] Missing delete wallet option

### LINKED ISSUE #12161: Missing delete wallet option
# Problem
It's not possible to remove a wallet from the app.

Given the user wants to delete a wallet
When user tries to do it
Then there is no option to do it

In case the user imports other wallets, user will keep this wallet as junk on his main page.
The user should be free to add and remove wallets at any time.

### Where should it be implemented?
Desktop and Mobile

## Implementation

Include a removal button or option at the menu
A strong remind warning must also be presented to the user

## Acceptance Criteria

Considering status will create a default wallet option:
- User must be able to remove any wallet, that is not the default one, except this wallet is the last and only left.
- User must confirm, that user knows the wallet will disappear from the app

## Notes

Check other wallets as benchmark.

Given the user wants to delete a wallet
When he clicks at removal button or option
Then the app will present a modal with a warning
When the user clicks at Yes
Then the wallet is removed from the app
And the user return to the main page

Given the user wants to delete a wallet
When he clicks at removal button or option
Then the app will present a modal with a warning
When the user clicks at No
Then the wallet is not removed from the app
And the user returns to the main page

## Future Steps

On-Going operations must be considered
Test Cases must be created
Issue must be validated

flexsurfer:
fixes #12161


churik:
**ISSUE 1: if an account is shared for accepting transaction, no warning is shown**
Quite a corner case, but if you shared account for transaction in 1-1 chat, and then deleted it, you've got: 
IMAGE 2021-08-16 12:58:31
Ideally show warning (smth like "you have shared your account for transaction, are you sure you want to delete it?").

But it is not a blocker for this PR IMO.
@flexsurfer tell me what you think about it.

flexsurfer:
thanks @churik fixed, we'll show the address in that case

churik:
@flexsurfer thank you!
May we display it in short version, i.e '0xnnnn..nnnn'?
IMAGE 2021-08-16 14:40:44

flexsurfer:
thanks @churik , good point, done

churik:
Perfect now! 
Ready to be merged.

flexsurfer:
@churik i've added password for imported accounts, so now when deleting imported accounts we ask a password

churik:
Tested deleting improted, watch-only accounts, also with resetting password.
Works as expected - password is required only for imported accounts

bitgamma (approved):

PR #12447, opened 2021-08-16
preload chats

flexsurfer:
issue: - because I have all those spam chats still (can't mass delete in ui) it takes several seconds to load the chat list, in reality I rarely scroll 2 screens down, especially on first open, is it possible to load or cache the first few pages, then update, so at least the screen isn't empty with a spinner?

solution:
we want to preload 30 chats first and then load all


Serhy:
Tested last build with hundreds of chats and it's visible the difference between PR and nightly for the same account (Galaxy S8 Android 9 used):

This is the nightly build:

https://user-images.githubusercontent.com/8749671/129887541-7c9ac2de-4bf5-4361-a42a-e0076f4f0010.mov

And this is PR build:


https://user-images.githubusercontent.com/8749671/129887995-6f89301d-c6af-4599-863a-17e3a1d8317b.mov

Waiting for review and e2e test then.. @flexsurfer could you fix conflicts please?

flexsurfer:
@Serhy  done

src/status_im/ethereum/json_rpc.cljs
```diff
@@ -72,6 +72,7 @@
    "wakuext_sendContactUpdate" {}
    "wakuext_sendContactUpdates" {}
    "wakuext_chats" {}
+   "wakuext_someChats" {}
```
rasom: Why `someChats`? Is that like a set of random chats? :)

flexsurfer: any better suggestions? it's not all but some `several, a few, some, rather, something` :)

rasom: `latestChats`, `latestNChats`, that's for getting chats with more recent updates, correct?

rasom: `latestActiveChats`, idk. `someChats` sounds too random imo

rasom (approved):

src/status_im/ethereum/json_rpc.cljs
flexsurfer: latestActiveChats looks good, thanks

Serhy:
E2e tests failed above are not related to this PR (and some are False Failed).

PR #12490, opened 2021-08-25
fix latestactive chats

flexsurfer:
fix latestactive chats


Serhy:
Tested with latest build and hundreds of chats: those chats with recent activity OR those which were at the end of list but received message - appears at the top of list after login and sorted accordingly to timestamp of activity  in chat.
Thanks @flexsurfer 
Could you fix conflicts please?

shivekkhurana (approved):

flexsurfer:
@Serhy  done

Serhy:
There is a bug with Status tab:
The first post in timeline won't be visible on Status tab. You send post - but nothing appears on Status view. Second and rest posts are visible fine. However only after re-login the first post will be visible.
**To reproduce:**
1) Create new account
2) Open Status tab
3) Send message (post) 


https://user-images.githubusercontent.com/8749671/130815353-1529dfe9-df92-489f-81f4-e48cc65580d9.mp4

briansztamfater (approved):

PR #12516, opened 2021-08-30
request permissions before saving images

### LINKED ISSUE #11402: No permission requested for files and media storage on Android when saving an image from the chat view
# Bug Report

## Problem

The issue is reopened and updated as the behavior has changed.

If try to save an image from the chat view without previously granting access to "Files and media" storage (i.e. on the fresh build) the access request will not be shown **_but the image will be saved_**. 

#### Expected behavior

File and media access is requested when saving an image, if it has not been previously granted

#### Actual behavior

Image is saved although no permission was granted

### Reproduction

User A:
- Install Status

User B:
- Start 1-1 chat with User A
- Send an image

User A:
- Open the image
- Three dots -> Save
- Check the device gallery

### Additional Information

- Status version: Nightly 22/09
- Operating System: Android

bitgamma:
Fixes #11402

### DIFF
--- a/ios/StatusIm/Info.plist
+++ b/ios/StatusIm/Info.plist
@@ -78,6 +78,8 @@
 	<string>Location access is required for some DApps to function properly.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Photos access is required to give you the ability to send images.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Photos access is required to give you the ability to save images.</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Need microphone access for sending audio messages.</string>
 	<key>UIAppFonts</key>
--- a/ios/StatusImPR/Info.plist
+++ b/ios/StatusImPR/Info.plist
@@ -83,6 +83,8 @@
 	<string>Location access is required for some DApps to function properly.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>Photos access is required to give you the ability to send images.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Photos access is required to give you the ability to save images.</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Need microphone access for sending audio messages.</string>
 	<key>UIAppFonts</key>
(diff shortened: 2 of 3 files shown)


flexsurfer (approved):

Serhy:
@bitgamma , thanks a lot for this fix!
Tested on Android 11, Android 9, iOS 14. All looks good! Permissions for Storage (access to Photos on iOS) when trying to save photos is asked. 
Two moments to address:

**Issue 1**
On iOS, in a case when permissions to Photos denied, all subsequent attempts to save images just "quietly" does nothing (nothing is a saved, but no any pop-up notifications saying "In order to save image you need to grant permissions to Photos")
We can use the same pop up which appears upon trying to Send images with Denied permissions ("To grant the required photos permissions, please go to your system settings and make sure that Status -> Photos is selected")
**Reproduction**:
2) Receive an image in 1-1 / Group chat
3) Save image by long tap on it -> `Save`
4) Tap `Deny` 
5) Try to save image again
**Actual result:** Photo is not saved but no notifications about permissions denied for storage
**Expected result:** 
Screenshot 2021-09-01 at 14 26 27

**Issue 2**
On iOS the text explaining what for user is about to grant permissions a bit misleading (when tap 'Save' on image): 
Screenshot 2021-09-01 at 14 40 01

It's the same for when user wants to send images. Saying "...to give you ability **to send** images". Is it possible to add `and save` then? So the text is `Photos access is required to give you the ability to send and save images.`

Also, because `Save` images does not need this `Select Photos...` item but access to whole 'Photos' folder, - is it possible to not show 'Select Photos...' option when `Save` image requested?

bitgamma:
Thanks for testing this, @Serhy !

those issues have not been introduced by this PR, they are because of iOS14 and react-native-camera-roll using a deprecated API. https://github.com/react-native-cameraroll/react-native-cameraroll/issues/292

Since that library looks unmaintained and it will also cause problems for solving https://github.com/status-im/status-react/issues/12525 I will either find a maintained fork or make a fork myself and fix the issue there. But that would not be in the scope of this PR.

In the meantime, I tried to at least show an error message when saving fails.

Serhy:
Thanks @bitgamma, retested on iOS and Android, looks good to me.
Failed tests are false failures.

PR #12536, opened 2021-09-02
Fix show/hide password on Android

### LINKED ISSUE #11955: [Bug] show / hide password "eye" won't "hide' password
# Bug Report

#### Expected behavior

[comment]: the 👀 in the password field should hided the password. 

#### Actual behavior

[comment]: unable to hide password after revealing password. 

#### Notes

For the developer who will be investigating. Is it related to another PR? Tips on where to start? 

#### Acceptance Criteria

Rules for the future PR to be accepted.

### Reproduction

- Open Status
- log out or add an acct
- type password
- tap the show password eye
- nothing it shows the password 
- Tap again
- password still visible 

### Additional Information

- Status version: 1.12.1 (20210324160131)
[comment]: # 
- Operating System: android 10

bitgamma:
Fixes #11955.

This PR actually has three parts

- [x] Make the seed phrase entry "visibile-password" on Android, otherwise SwiftKey and similar would record the seed phrase
- [x] Make the password hide/show button work in both directions.
- [x] When the text is password/visibile-password make sure auto correct and autocomplete is disabled

As a side effect of point 2, when the password is shown, the "suggestions" from SwiftKey and probably other keyboard as activated. This is because I had to make a sort of workaround until https://github.com/facebook/react-native/issues/27946 is fixed. Point 3 seems to at least (from my limited testing) avoid that the password is stored by the keyboard.

### DIFF
--- a/src/status_im/ui/screens/multiaccounts/views.cljs
+++ b/src/status_im/ui/screens/multiaccounts/views.cljs
@@ -64,6 +64,7 @@
    [quo/text-input
     {:show-cancel         false
      :auto-correct        false
+     :keyboard-type       :visible-password
      :placeholder         (i18n/label :t/seed-phrase-placeholder)
      :monospace           true
      :multiline           true
--- a/src/quo/components/text_input.cljs
+++ b/src/quo/components/text_input.cljs
@@ -146,7 +146,7 @@
     (fn [{:keys [label multiline error style input-style keyboard-type before after
                  cancel-label on-focus on-blur show-cancel accessibility-label
                  bottom-value secure-text-entry container-style get-ref on-cancel
-                 monospace]
+                 monospace auto-complete-type auto-correct]
           :or  {cancel-label "Cancel"}
           :as  props}]
       {:pre [(check-spec ::text-input props)]}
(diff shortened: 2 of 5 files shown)


src/status_im/ui/screens/multiaccounts/views.cljs
```diff
@@ -64,6 +64,7 @@
    [quo/text-input
     {:show-cancel         false
      :auto-correct        false
+     :keyboard-type       "visible-password"
```
flexsurfer: better to use keyword

briansztamfater (approved):

flexsurfer (changes requested):
pls do not merge, i'll make review tomorrow, thanks

src/status_im/ui/screens/multiaccounts/views.cljs
bitgamma: "visibile-password" appears in the code in other places too as string so I kept it this way, but I can change it everywhere in the code

flexsurfer (approved):

Serhy:
Tested with Android 11 and Android 9 and iOS 14, - seed phrase field does not have autosuggestions (and autocomplete panel) enabled, password fields hide/show also works. 
Failed tests were false failures (can not reproduce manually), - will fix them separately. 
Thanks a lot, @bitgamma !

Serhy:
@bitgamma sorry, i've missed regression after this change, - https://github.com/status-im/status-react/issues/12560
(on iOS all one-line inputs behave like passwords now)

PR #12542, opened 2021-09-03
target Android 30, support scoped storage

### LINKED ISSUE #12063: Remove android:requestLegacyExternalStorage
# Problem

Because of some Policy changes on the play store https://developer.android.com/about/versions/11/privacy/storage 
we won't be able to use this permission anymore from the 5th of May.

One of the libraries we have is currently using it https://github.com/react-native-cameraroll/react-native-cameraroll/issues/304 and seems like it will take some time for them to upgrade to the new system.

My understanding that this will become a problem once we start targeting API 30 (we currently target API 29).

We will have to target API 30 by August (or November, still not very clear).

Here's a recap from a developer for the changes:

https://www.reddit.com/r/androiddev/comments/mwaqn1/scoped_storage_recap/

My understanding is that:

- From the 5th of May, if we target API 30, we will not be able to ask for android:requestLegacyExternalStorage and we would have instead to ask for `MANAGE_EXTERNAL_STORAGE`, but we'd have to ask google permission to allow our use case (unlikely to happen).

- If we don't target API 30 (our case), we can still use android:requestLegacyExternalStorage and publish updates to the play store, until we move to API 30. This will have to happen for August or November (to clarify).

### LINKED ISSUE #12525: Upgrade app to target API 30
There's a couple of warnings that we need to address for November 30.

```
Warning
We've detected that this app is using an old version of Google Play Billing. By November 1, 2021, all app updates must use Billing Library version 3 or newer. Update to Billing Library 3 before this date. Learn More about Billing Library 3.
```

```
Warning
Your app currently targets API level 29 and must target at least API level 30 to ensure it is built on the latest APIs optimized for security and performance.
From November 2021, app updates must target Android 11 (API level 30).
```

bitgamma:
Fixes #12525.
Fixes #12063.

- [x] update SDK declaration in gradle and nix
- [x] remove billing permission (seemed to be unused anyway) 
- [x] remove android:requestLegacyExternalStorage 
- [x] fork and fix react-native-cameraroll for Android 10 and newer with scoped storage (both save and getPhotos)
- [x] make getPhotos actually work in non-debug builds (fix was updating fresco to 2.2.0 which is the version imported by Flipper which is only enabled in debug builds)
- [x] set WRITE_EXTERNAL_STORAGE only for SDK <= 28, remove related permissions prompt on >= 29
- [x] make react-native-image-crop-picker stop requesting WRITE_EXTERNAL_STORAGE on SDK >= 29
- [x] make "Send logs" functionality work with scoped storage
- [x] check other places that might require adaptation
- [x] make react-native-cameraroll running on iOS14 ask the correct permission when saving to get clearer system prompt (unrelated but since it regards react-native-cameraroll let's do it here)

QA: Since this changes the targeted SDK, please check at least Android 9, 10 and 11. Even older versions would be nice too. In particular, saving and sending images + sending logs should be tested.  Image sending/saving should also be tested on iOS14, 13 and maybe older if you have. I only have Android 10 and iOS14 devices so could only do limited testing.

Another note, apparently there can be a difference if the APK was already installed (and thus got legacy storage access first time, which apparently is not revoked on Android 10) and when it is freshly installed. Doesn't seem to affect too much but is worth testing.


nix/pkgs/go-maven-resolver/default.nix
```diff
@@ -1,4 +1,4 @@
-{ lib, buildGo116Package, fetchFromGitHub }:
+{ lib, buildGo116Module, fetchFromGitHub }:
```
jakubgs, line 1: Oh, nice catch. I must have broken that last time when we were upgrading Go verison.

jakubgs (approved):
Nix and Gradle changes look good to me.

bitgamma:
@Serhy please use the apk that just finished building when testing. There was a mismatch between the version of fresco used in debug and pr/release/e2e build which made the small image gallery when setting status/sending image not work

Serhy:
Thanks @bitgamma 
Tested on Android 9 and Android 11 and iOS 14, - everything works as before on Androids (including Android 10 for upgrading and access to Storage)
On iOS though, after upgrade I needed to give permissions when Saving/Sending photos again, but I don't see it's a problem. 
Photos can be sent, received and saved as before, logs could be sent with the same log_level granularity (logcat and geth.log are present).

bitgamma:
@Serhy thank you for testing it! Yes the iOS behavior is expected, although I think if you give read permission first then you are not asked permission again when saving, but if you save first you are asked again when reading because reading is higher permission level. But this way we get a good prompt when saving the image.

PR #12544, opened 2021-09-04
Small UI Design Improvements

Parveshdhull:
# Small UI Design Improvements

#### Summary
Fixed distortion of Status Bar due to opening of Bottom Sheet (In Android Device)
<div align="center">
Screenshot
</div>

#### Summary
Fixed cropping of Subtitle texts

<div align="center">
Screenshot
   <div width="40"/>
Screenshot
</div>

### DIFF
--- a/src/status_im/navigation/core.cljs
+++ b/src/status_im/navigation/core.cljs
@@ -272,7 +272,7 @@
                  {:component {:name    comp
                               :id      comp
                               :options (merge (cond-> (roots/status-bar-options)
-                                                (and platform/android? (not (colors/dark?)))
+                                                (and platform/android? (colors/dark?))
                                                 (assoc-in [:statusBar :translucent] true))
                                               {:layout  {:componentBackgroundColor (if platform/android?
                                                                                      (:backdrop @quo.colors/theme)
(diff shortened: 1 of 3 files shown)


src/quo/components/list/item.cljs
```diff
@@ -198,7 +198,7 @@
            title subtitle subtitle-secondary active on-press on-long-press chevron size text-size
            accessory-text accessibility-label title-accessibility-label accessory-style
            haptic-feedback haptic-type error animated animated-accessory? title-text-weight]
-    :or   {subtitle-max-lines 1
+    :or   {subtitle-max-lines 0
```
flexsurfer: I'm not sure we want to change the default value for the component, better to pass a parameter for this case only

flexsurfer (changes requested):

Parveshdhull:
Hi @flexsurfer ,
I restored the default value and passed the parameter to all 4 instances where overflowing was happening.

src/quo/components/list/item.cljs
Parveshdhull: Hi @flexsurfer ,
I restored the default value and passed the parameter to 4 instances where overflowing was happening.

flexsurfer (approved):

qoqobolo:
@Parveshdhull thanks for these improvements, looks better now!
PR is tested and can be merged.

Issue #12486, opened 2021-08-24
Not all messages are deleted if sent and deleted several in a row


churik:
# Bug Report

## Problem 
If user A sends 6-8 messages in a row and then delete it, recipient may receive some of them (1-3 messages are not deleted for recipient)

#### Expected behavior
all messages are deleted

#### Actual behavior
several deleted messages arrived

https://user-images.githubusercontent.com/4557972/130614056-c1a500bc-bed4-4d13-bbac-6f40ebd3cd10.mp4

### Reproduction
Reproducible also for 1-1 chat.

1) User A, User B: open status, join same public chat
2) User B: close app
3) User A: send several messages (i.e. `1`, `2`, `3`, `4`, `5`, `6`, `7`)
4) User A: delete all these messages
5) User B: open app, login and check public chat

### Additional Information

- Status version: nightly 24/08/2021
- Operating System: Android, iOS

[comment]: # (Please, add logs/notes if necessary)
Logs: [Status-debug-logs.zip](https://github.com/status-im/status-react/files/7038993/Status-debug-logs.zip)

churik:
After relogin deleted messages are still there, so likely issue on go-side

PR #12561, opened 2021-09-08
[Fixes: #12486] Fix delete of messages

cammellos:
fixes deletion of messages in a row 
and https://github.com/status-im/security-reports/issues/28


flexsurfer (approved):

qoqobolo:
@cammellos thanks for the fixes!

#### ISSUE 1: A preview of one of the deleted messages is displayed on the home screen

**_Reproduction:_** 

1. User A, User B: open status, join same public chat
2. User B: close app
3. User A: send several messages (i.e. 1, 2, 3, 4, 5, 6, 7)
4. User A: delete all these messages
5. User B: open app, login and check chat preview on the home screen


https://user-images.githubusercontent.com/67952253/132712349-f2343c2f-4853-488b-a3ef-5c167d305482.mp4

cammellos:
@qoqobolo It should be fixed now, I have tested the latest build and could not replicate, so :crossed_fingers: 
Thank you

qoqobolo:
@cammellos unfortunately, the issue is still reproducible on the latest build :(
But this is no longer so noticeable because if a user deletes several messages in a row and then sends at least one more, this preview will not be visible to anyone.
So I believe we can register this leftover separately and merge this PR, if you agree.

cammellos:
@qoqobolo sounds good with me, let's do that
Thank you

shivekkhurana (approved):

PR #12575, opened 2021-09-13
enable ENS on ropsten for PR builds

bitgamma:
attempt to resolve ENS on ropsten for PR builds. The VERIFY_ENS_CONTRACT_ADDRESS does not need to be explicitly because we already have a table (also the address is the same across networks)


churik:
Tests are fixed in https://github.com/status-im/status-react/issues/12586

churik (approved):

PR #12578, opened 2021-09-13
Added swipe to reply to chat message gesture

### LINKED ISSUE #12361: Swipe-to-reply gesture
As a user
I want to be able to swipe-to-reply
So that I can easily reply to a message

### Description
A user should be able to swipe on a message, and that should populate the reply input.

Swipe-To-Reply

jo-mut:
[Fixes #12361 ] Swipe-to-reply gesture

### Summary
When a user swipes on a textual chat message, the reply input will be launched and populated with the message being replied to. The swiped message will also show a spring animation.

### Review notes
This functionality is only on textual chat message. I didn't find images on public chats to test the feature on images.

### Testing notes
- Click on any public chat
- Swipe on a text message 
- See that the reply input has been populated and the swiped message has a spring animation for visual feedback upon  swiping
#### Platforms

- Android
- iOS
- macOS
- Linux
- Windows

##### Functional

- 1-1 chats
- public chats
- group chats
- wallet / transactions
- dapps / app browsing
- account recovery
- new account
- user profile updates
- networks
- mailservers
- fleet
- bootnodes

##### Non-functional

- battery performance
- network consumption

### Steps to test

- Open Status
- ...
- Step 3, etc.

status: ready


errorists:
Leaving some design input: there's barely any perceptible feedback that you're dragging a message, the X position of the message doesn't map to your finger movements so you don't have a feeling like you're manipulating it with your gesture. It's very hit and miss, sometimes I see some animation happening, more often it goes straight to the input with no feedback.

jo-mut:
I will instead increase the X translation so that its more visible.

errorists:
I think that's not the right approach, you should try sampling the location of the finger in relation to the screen and interpolate the message position based on that, so the two are always in perfect sync

jo-mut:
Okay I will work on that then submit a new pull request.
Thank you for your guidance

PR #12580, opened 2021-09-14
update-status-go: does not need to depend on make shell

jakubgs:
The dependency on `make shell` is confusing and unnecessary since `update-status-go.sh` script only requires availability of the `nix-prefetch-url` utility to check SHA256 of the `status-go` verison.

Use of `make shell` with this script can also cause issues when running `make run-android` if already in the shell, which prevents spawning a new one and effectively stops changes to `status-go-verison.json` file taking effect.

Related docs changes: https://github.com/status-im/status.im/pull/678


bitgamma (approved):

Parveshdhull (approved):
Works Great 👍, Thanks

PR #12605, opened 2021-09-21
Change status-go version

alaibe:
Running e2e test with new status go version


alaibe:
only for testing, closing

PR #12643, opened 2021-09-23
🚀 contacts performance 🚀

flexsurfer:
🚀 contacts performance 🚀

replaced systemTags on added and blocked fields, all other changes in status-go

QA: pls ping me when testing


churik:
@flexsurfer 
can you please rebase it? 
let me know when it is ready for testing.
Thank you!

flexsurfer:
@churik  done

churik:
@flexsurfer thank you for PR!
### ISSUE 1: unblocking and removing the user from contacts is not synced between paired devices

**Steps:**
1) restore multiaccount from seed phrase on Device A and B
2) pair devices
3) device A: add user A, block user B 
4) device B: wait until it will be synced
5) device A: remove user A from contacts, unblock User B

**Expected result:**
unblocking and removing will be synced

**Actual result:**
not synced

**OS:** IOS, Android


### ISSUE 2: Users, who are not added to contacts, don't appear in mention suggestion list
Monosnap 2021-09-27 14-58-39

flexsurfer:
> ISSUE 2: Users, who are not added to contacts, don't appear in the mention suggestion list

unfortunately its a price for now, but anyway we want to remove public chats soon, for communities and group chats there should be another solution implemented separately

churik:
@cammellos @hesterbruikman 
so currently it means that mention for non-contacts will not work (public/group chats/communities), not sure what should we do here.

cammellos:
My 2p is that it would be frustrating for user if they ever have to mention someone, so I'd be inclined in fixing it, communities I'd be ok as still in alpha, but public/group chats he would be noticeable

flexsurfer:
i'll fix that then, thanks

flexsurfer:
thanks @churik  all fixed

churik:
Awesome work @flexsurfer!
No regression found.

shivekkhurana (approved):

PR #12679, opened 2021-10-04
[#12666] Allow overriding nonce when sending transaction

### LINKED ISSUE #12666: Allow overriding nonce when sending transaction
As a user
I would like to have a way to override a nonce of a transaction
So I can cancel/speed up transactions manually

### Description

We should give an option to set a custom nonce, so that advanced users can cancel/speed up a transaction.

flexsurfer:
fixes #12666

### DIFF
--- a/src/status_im/signing/core.cljs
+++ b/src/status_im/signing/core.cljs
@@ -93,7 +93,7 @@
   {:events [:signing.ui/sign-is-pressed]}
   [{{:signing/keys [sign tx] :as db} :db :as cofx}]
   (let [{:keys [in-progress? password]} sign
-        {:keys [tx-obj gas gasPrice maxPriorityFeePerGas maxFeePerGas message]} tx
+        {:keys [tx-obj gas gasPrice maxPriorityFeePerGas maxFeePerGas message nonce]} tx
         hashed-password (ethereum/sha3 (security/safe-unmask-data password))]
     (if message
       (sign-message cofx)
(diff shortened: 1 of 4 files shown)


cammellos (approved):

Serhy:
1) Sent transaction of **ETH** with low gas price.
 - Cancel tx with **Cancel** button (same nonce of TX picked up of that TX which stuck in Pending) **PASS**
 - Increase speed of tx with **Increase Gas** button (same nonce of TX picked up of that TX which stuck in Pending) **PASS**
2) Sent transaction of **token** with low gas price.
 - Cancel tx with **Cancel** button PASS (same nonce of TX picked up of that TX which stuck in Pending) **PASS**
 - Increase speed of tx with **Increase Gas** button PASS (same nonce of TX picked up of that TX which stuck in Pending) **PASS** 
 - Send transaction with custom 'nonce' value set (which is the same to that transaction in 'Pending...') -> Result: transaction replaced with the new one. **PASS**

3) Send TWO transactions with tokens: first with low gas price (so it's in pending) and second one with normal gas price. 
Cancel first transaction with 'Cancel' button. Result: both transactions sent and visible in TX history **PASS**

4) Same account recovered on two devices. On one device user sends TX with tokens with los gas price. On another device this same user sends another transaction (with high Gas Price) with the same nonce of TX sent on first device. Result: transaction 'replaced' and the one sent from second device displayed on both devices in TX history (on first device it's initial first transaction jeep staying in 'Pending...' but it's a subject of different issue) **PASS**

Serhy:
Failed e2e is not related to this PR changes.
Thanks @flexsurfer !

PR #12688, opened 2021-10-06
[#12684] Messages disappear from Timeline if join public chat and ope…

### LINKED ISSUE #12684: Messages disappear from Timeline if join public chat and open Timeline tab while public chat history is loading
# Bug Report

## Problem 

If post a message to Timeline, join a public chat, and quickly navigate to the Timeline tab while public chat history is loading, then timeline messages disappear when public chat history is loaded.
If send another message to Timeline or re-open the Timeline tab, the messages will be restored.

#### Actual behavior

https://user-images.githubusercontent.com/67952253/135993184-3f78e0c7-266b-41b0-ac55-b9c3ea9988ac.mp4

### Reproduction

1. Post smth to Timeline
2. Open the home screen and join a public chat, e.g. #status
3. Quickly navigate to the Timeline tab

### Additional Information

- Status version: nightly 5/10

[geth.log](https://github.com/status-im/status-react/files/7284408/geth.log)
[Status.log](https://github.com/status-im/status-react/files/7284410/Status.log)

flexsurfer:
fixes #12684

### DIFF
--- a/src/status_im/contact/core.cljs
+++ b/src/status_im/contact/core.cljs
@@ -39,27 +39,37 @@
 
 (fx/defn ensure-contacts
   [{:keys [db]} contacts chats]
-  {:db (update db :contacts/contacts
-               #(reduce (fn [acc {:keys [public-key] :as contact}]
-                          (-> acc
-                              (update public-key merge contact)
-                              (assoc-in [public-key :nickname] (:nickname contact))))
-                        %
-                        contacts))
-   :dispatch-n (mapcat (fn [{:keys [public-key] :as contact}]
-                         (cond-> []
-                           (:added contact)
-                           (conj [:start-profile-chat public-key])
+  (let [events
+        (reduce
+         (fn [acc {:keys [public-key] :as contact}]
+           (let [added (:added contact)
+                 was-added (contact.db/added? db public-key)
+                 blocked (:blocked contact)
+                 was-blocked (contact.db/blocked? db public-key)]
+             (cond-> acc
+               (and added (not was-added))
+               (conj [:start-profile-chat public-key])
 
-                           (not (:added contact))
-                           (conj [:offload-messages constants/timeline-chat-id])
+               (and was-added (not added))
+               (conj nil)
 
-                           (:blocked contact)
-                           (conj [::contact.block/contact-blocked contact chats])
+               (and blocked (not was-blocked))
+               (conj [::contact.block/contact-blocked contact chats])
 
-                           (contact.db/blocked? db public-key)
-                           (conj [::contact.block/contact-unblocked public-key])))
-                       contacts)})
+               (and was-blocked (not blocked))
+               (conj [::contact.block/contact-unblocked public-key]))))
+         [[:offload-messages constants/timeline-chat-id]]
+         contacts)]
+    (merge
+     {:db (update db :contacts/contacts
+                  #(reduce (fn [acc {:keys [public-key] :as contact}]
+                             (-> acc
+                                 (update public-key merge contact)
+                                 (assoc-in [public-key :nickname] (:nickname contact))))
+                           %
+                           contacts))}
+     (when (> (count events) 1)
+       {:dispatch-n events}))))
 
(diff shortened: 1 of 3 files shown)


src/status_im/contact/core.cljs
```diff
@@ -39,27 +39,35 @@
...
-                           (conj [::contact.block/contact-blocked contact chats])
-
-                           (contact.db/blocked? db public-key)
-                           (conj [::contact.block/contact-unblocked public-key])))
-                       contacts)})
+  (let [{:keys [events offload]}
+        (reduce
+         (fn [acc {:keys [public-key] :as contact}]
+           (cond-> acc
+             (and (:added contact) (not (get-in db [:contacts/contacts public-key :added])))
```
rasom: Can we extract `(:added contact)` to `added` and `(get-in db [:contacts/contacts public-key :added])` to `was-added` to make conditions a bit more readable?

rasom: same for `blocked`/`was-blocked`

src/status_im/contact/core.cljs
```diff
@@ -39,27 +39,35 @@
...
+
+             (and (:blocked contact) (not (contact.db/blocked? db public-key)))
+             (-> (update :events conj [::contact.block/contact-blocked contact chats])
+                 (assoc :offload true))
+
+             (and (contact.db/blocked? db public-key) (not (:blocked contact)))
+             (-> (update :events conj [::contact.block/contact-unblocked public-key])
+                 (assoc :offload true))))
+         {:events []
+          :offload false}
```
rasom: Can it just be replaced with `(seq events)`? Apparently if there is at least one event `:offload` is always `true`. That would also simplify code inside `reduce`.

src/status_im/contact/core.cljs
flexsurfer: sure

src/status_im/contact/core.cljs
flexsurfer: if contact was removed there is no event

flexsurfer: i also don't like this code, but didn't find a better way

rasom: > if contact was removed there is no event

but it is `true` in all causes now, no?

rasom: oh I see

flexsurfer: ok `:dispatch-n` should ignore nil., so we could add nil for contact removal case

rasom: We also could just use set instead of vector for events and then adding multiple `[:offload-messages constants/timeline-chat-id]` wouldn't be an issue.

flexsurfer: ok i found even better solution

flexsurfer: @rasom check it now

src/status_im/contact/core.cljs
```diff
@@ -39,27 +39,34 @@
...
+             (cond-> acc
+               (and added (not was-added))
+               (conj [:start-profile-chat public-key])
+
+               (and blocked (not was-blocked))
+               (conj [::contact.block/contact-blocked contact chats])
+
+               (and was-blocked (not blocked))
+               (conj [::contact.block/contact-unblocked public-key]))))
+         [[:offload-messages constants/timeline-chat-id]]
```
rasom, line 61: I still think that a set instead of vector and explicit adding of `[:offload-messages constants/timeline-chat-id]` when it is really necessary would be better

flexsurfer: but it really necessary for all cases, so it will be just duplicated for each condition, and will be noisy

rasom: ok, I see. In code you just added there is no `(and (not added) was-added)` cause, we don't need it?

flexsurfer: we need it, fixed

rasom (approved):

PR #12698, opened 2021-10-12
[#12531 pairing] Sync read messages

### LINKED ISSUE #12531: Syncing of unread/read messages across devices
As a user
I want my unread messages to be synced across devices
So that I don't have to read my messages twice when moving from one device to another

### Description
Unread/read messages should be synced across devices, so that reading a message on device 1, should also mark as read the message on device 2.

Care must be taken so that not a lot of traffic is generated (i.e not one protocol message per message read, while it should be acceptable per chat), and that ordering is taken into consideration (the "mark as read" synced message might arrive before the unread messages themselves on the other device).

rasom:
fix #12531

status: ready

### DIFF
--- a/src/status_im/transport/message/core.cljs
+++ b/src/status_im/transport/message/core.cljs
@@ -50,31 +50,10 @@
 
       (seq chats)
       (do
-        (when (seq contacts)
-          ;; Note(rasom): in case if response contains contacts entries, some
-          ;; of them might be blocked. That means we might need to recalculate
-          ;; unviewed/mentions counters in chats where blocked users sent
-          ;; messages. Currently we ignore unviewed/mentions values which come
-          ;; from go side (see code few lines below) because it (probably)
-          ;; breaks app-db state. But we still have to apply those values when
-          ;; blocked contact is received from a paired device.
-          ;; In order to do so, we store all affected chats separately so that
-          ;; process loop might continue and then we use those chats on
-          ;; processing contacts update. Currently it causes one bug: for some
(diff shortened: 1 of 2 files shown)


cammellos (approved):

churik:
@rasom 
### ISSUE 1: unread messages for group chats are not synced

**Steps:**
1) Device1, device2: restore multiaccount from seed phrase, pair devices
2) Device1: create group chat with device 3
3) Device3: send message to group chat
4) Device1: open group chat to read message

**Expected result:**
no unread indicator for group chat is shown on Device2 

**Actual result:**
unread indicator for group chat is shown on Device2 remains the same
**OS:** IOS, Android

### Q 2: what about offline?
Currently if Device1 and Device2 are paired and one of the devices is offline, unread messages won't sync if they were read until getting back paired device online.
So if Device1 got message and read it (device2 was offline), when device2 came back online, message will be unread for Device2.
Is that smth expected by now?

rasom:
@churik I believe that offline behaviour is same for all types of messages and this is not an objective of this PR.

rasom:
@churik issue 1 should be fixed

shivekkhurana (approved):

churik:
Tested:
- unread message before sync  will be shown after initial sync
- for all types of chats unread indicator is synced during several seconds if devices are online
- if 1 of devices was offline, unread messages are synced right after getting back to online 

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)
- Xiaomi Mi Note 9 Pro (Android 10) unread messages before sync will shown  after sync

PR #12701, opened 2021-10-12
Wallet Connect 2.0 integration

### LINKED ISSUE #12546: Wallet connect 2.0 integration
As a user
I want to be able to use wallet connect 2.0
So I can use dapps on desktop etc

### Description

We should support the wallet connect 2.0 protocol.
The protocol uses waku 2.0, so a connection needs to be made, likely in parallel to a waku v1 connection (currently not supported). 

### Designs

https://www.figma.com/file/MYWmd1buvc2AMvUmFP9w42t5/Browser?node-id=6030%3A15485

briansztamfater:
fixes #12546 

### Summary

Wallet Connect 2.0 integration. This first approach, as discussed with the team, includes the client library, which is a centralized and experimental approach because it depends on a relay proxy server and the wallet-connect/client package. The reason behind this is that go-waku does not support the waku2 methods that are needed to use to implement a fully decentralized approach, and also status-go does not support having a waku and waku2 connection in parallel. The plan is to remove the `@wallet-connect/client` dependency in the future and replace it with a pure waku2 implementation when feasible.

#### Platforms

- Android
- iOS

##### Functional

- wallet / transactions
- dapps / app browsing

### Steps to test

- Open Status App
- Go to Advanced settings and enable Wallet Connect 2.0
- Open desktop Dapp, as there are many few amount of dapps using WC 2.0, it can be tested with this example https://react-app.walletconnect.com/
- On desktop Dapp, select chains and click connect
- On Status app, click + on main screen and scan QR
- Wait for the session proposal bottom sheet to appear and connect a wallet
- Execute a transaction in the dapp and wait for signing sheet to appear
- Sign transaction, wait for the result in the dapp
- As a note, `eth_signTypedData` includes a chainId parameter, so wallet should be configured on that same network in order to sign that message
- We should test it with other dapps that implements WC 2.0. Need to research for more besides the testing dapp.

status: ready


flexsurfer:
just wondering, if WC2.0 uses waku why don't we have our own implementation?

flexsurfer:
i mean we should use our waku implementation in status-go, no sense to have an implementation in js with all dependencies

briansztamfater:
> i mean we should use our waku implementation in status-go, no sense to have an implementation in js with all dependencies

Yes, it make sense, I'm using the client implementation by now to implement the flows and have a basic implementation and will switch to pure wakuv2 implementation on a later iteration.

nix/scripts/node_modules.sh
```diff
@@ -111,9 +111,14 @@ function replaceNodeModules() {
   echo "Copying node_modules from Nix store:" >&2
   echo " - ${src}" >&2
   copyNodeModules ${src} ${dst}
+  nodeify
```
jakubgs: This will not work for CI builds, because the `nix/scripts/node_modules.sh` script is only used for shells, and is not used by the normal CI build process, since this is what happens in normal Nix build:
https://github.com/status-im/status-react/blob/c9c8263c9e35142693e2ebd2f640f340708296ee/nix/mobile/android/release.nix#L114-L117
A functional solution would involve changing `nix/deps/nodejs-patched/default.nix` instead.

jakubgs: But I'd like to avoid adding any more special patching steps to our build process. So if possible a much better solution could maybe involve a meta package that does the shimming/polyfill so it can be included as a normal dependency.

jakubgs: For example, here is a package that provides various polifills for `react-native`:
https://github.com/acostalima/react-native-polyfill-globals

Couldn't we do something similar for the `crypto` and `stream` package?

jakubgs:
Oh yeah, that looks much better, and no need to change Nix build process with extra patching.

briansztamfater:
> Oh yeah, that looks much better, and no need to change Nix build process with extra patching.

YES! I ended up finding a much cleaner solution that didn't involve additional patching, it is mentioned in the repo you pointed out. Thanks for your help, was really useful :)

src/status_im/utils/wallet_connect.cljs
```diff
@@ -0,0 +1,32 @@
+(ns status-im.utils.wallet-connect
+  (:require ["@walletconnect/client" :refer [CLIENT_EVENTS] :default WalletConnectClient]
+            ["@react-native-community/async-storage" :default AsyncStorage]
+            [clojure.string :as string]))
+
+(def default-relay-provider "wss://relay.walletconnect.org")
```
shivekkhurana: Config is a better place for urls and metadata.

shivekkhurana (changes requested):

src/status_im/utils/wallet_connect.cljs
briansztamfater: Moved.

src/status_im/events.cljs
```diff
@@ -60,7 +60,7 @@
             [status-im.wallet.core :as wallet]
             status-im.wallet.custom-tokens.core
             [status-im.navigation.core :as navigation.core]
-            [status-im.multiaccounts.login.core :as login.core]))
+            status-im.wallet-connect.core))
```
flexsurfer, line 65: login.core is not used ?

briansztamfater: Yes, but `status-im.multiaccounts.login.core` it is already in the require list. I made this change based on a rebase conflict, probably unrelated to this PR logic.

qoqobolo:
Hey @briansztamfater, can you ping someone from the QA team once the PR is ready for testing, please?

src/status_im/ui/screens/bottom_sheets/views.cljs
```diff
@@ -40,4 +40,4 @@
           (merge key-storage/migrate-account-password))]
     [quo/bottom-sheet opts
      (when content
-       [content])]))
+       [content (when options options)])]))
```
cammellos, line 43: This is the same as  `[content options]` I believe, as `(when false true)` would return nil (the only difference I can think is of is if you expect `options` to be literally `false`)

cammellos (approved):

churik:
@briansztamfater any apps for this to check on the mobile side (or only test app is available)?

briansztamfater:
> @briansztamfater any apps for this to check on mobile side?

I'm not sure if there's any production dapp using the 2.0 version of Wallet Connect, but theres's a simple example dapp that can be deployed to test integration. Here's an example: https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/

churik:
@briansztamfater 
thank you for PR.
Unfortunately, it is not really testable for me at that moment.

So I was able to get the session proposal bottom sheet only once on an Android device, it looked like this:

IMAGE 2022-02-09 14:07:21
I clicked on "Etereum account" and got this:
IMAGE 2022-02-09 14:08:47

From this state, I wasn't able to navigate back (clicking on the grey area or clicking on the "Back" system button didn't help), so I reopened the app.
And after that when I scanned QR nothing happens, so the session proposal bottom sheet doesn't appear anymore.
That's how I generate QR (maybe I do smth wrong):

https://user-images.githubusercontent.com/4557972/153208150-46958a53-508e-4e73-a149-8ae909e6e9a0.mp4

I switched to mainnet and enabled 'Wallet connect 2.0' in Settings in advance, scanning QR via + on home screen in app, so I'm out of ideas what can be wrong here.
Logs from Android device: 
[Status-debug-logs.zip](https://github.com/status-im/status-react/files/8032993/Status-debug-logs.zip)
On IOS device I didn't manage to get session proposal bottom sheet

briansztamfater:
@churik That's weird. I can see some UI bugs because of dark mode and maybe some Android-specific styling issues. I'll look at them.

Regarding QR scanning, the process is correct, I'll do more testing and debugging to see what could be wrong there.

briansztamfater:
@churik Made several fixes, and tested them with a Xiaomi MI 10 T, and also iOS Simulator, now it seems to be working fine. Can you try again? Let me know any issues :)

churik:
@briansztamfater 
I again repeated same steps and can't get session proposal bottom sheet.
Asked Pavel to verify - got the same.
Tried different browsers for https://react-app-git-fork-briansztamfater-v20-fi-4828be-walletconnect1.vercel.app/ (Chrome, Safari) 
The same on IOS and Android

https://user-images.githubusercontent.com/4557972/153407650-fab1ae7a-0b22-4331-86f1-956caaadaff5.mp4

Logs: [logs (3).zip](https://github.com/status-im/status-react/files/8040888/logs.3.zip)

briansztamfater:
Link on the PR description was outdated, https://react-app.walletconnect.com/ seems to work. I updated the link with this one.

churik:
New dapp url: https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/

### ISSUE 1: can't sign typed message (Mainnet)

**Steps:**
1) open status app, enable web3, switch to mainnet
2) Connect to https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/ on 'Ethereum' mainnet also
3) Accept session proposal bottom sheet
4) Check that mainnet balance will be shown
5) Use eth_signTypedData
6) sign in status

**Expected result:**
signed

**Actual result:**
IMAGE 2022-02-10 15:16:59


**OS:** IOS, Android

### ISSUE 2: can't use eth_sendTransaction (Mainnet)

**Steps:**
1) open status app, enable web3, switch to mainnet
2) Connect to https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/ on 'Ethereum' mainnet also
3) Accept session proposal bottom sheet
4) Check that mainnet balance will be shown
5) Use eth_sendTransaction

**Expected result:** network fee is set, can sign tx
**Actual result:**
IMAGE 2022-02-10 15:27:52
[logs (4).zip](https://github.com/status-im/status-react/files/8041752/logs.4.zip)


### Question 3: should tap on "Cancel" in app discard Pending JSON-RPC Request?
Sometimes tapping on some button in dapp (i.e. personal sign) may lead to several bottom sheet in Status app. Not sure we are considering that as an issue on that stage.


### ISSUE 4: if changing wallet after connecting to wallet-connect dapp, the old account will be shown

**Steps:**
1) open status app, enable web3, switch to mainnet, create second `account A` in wallet
2) Connect `account A`  to https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/ on 'Ethereum' mainnet also
3) Accept session proposal bottom sheet, don't close "Connection request" bottom sheet
4) change wallet back to `Ethereum wallet`
5) Go back to `https://react-app-git-extend-expiry-method-walletconnect1.vercel.app/`
**Expected result:** app will be reconnected to `Ethereum wallet`
**Actual result:** app is connected to `account A`

briansztamfater:
@churik Thanks for the feedback!

**Issue 1:** This issue seems like a bug on the dapp side, because it always sends chainId: 42 in the message payload. I think chain ID 42 is Kovan network, should work on that network. I can see if I can fork the dapp repo to send the current chain ID, but at the meantime you could try to configure Kovan network and see if it works (I tested before and worked with Kovan).

**Issue 2:** I need to check this. I didn't test it with value transactions because I didn't have funds, but I'll see how can I reproduce it.

**Question 3:** I would say that would not be a real life scenario so I'd dismiss it by now and we can improve later if it makes sense, but I don't have a strong opinion about it.

**Issue 4:** What happens if you refresh the browser website after changing the wallet account? If it changes, I assume it is a bug on the dapp side that is not handling session updates in real time.

churik:
> Issue 4 

Yes, you're right that's dapp behavior - after 10 sec or so after refreshing right wallet is shown, so can skip issue 4

churik:
I've managed to get some test ETH on Kovan network.
And can confirm that method eth_signTypedData is working on Kovan. For eth_sendTransaction Network fee is not set, however I can't see fee for ordinary tx also, so that may be unrelated.
IMAGE 2022-02-11 11:35:16
 We need tofind a way to check them on Mainnet, I'll check existing dapps from wallet connect website, maybe we will be lucky to find smth

churik:
Ok, tried around 20 apps from https://walletconnect.com/registry/apps and yeah, I suppose most of them are using wallet-connect 1.0 (which is understandable) 
Let me check my understanding: for now, it is in the Advanced option as wallet 2.0 is not officially released and when it will be released we enable this setting by default, right?

src/status_im/ui/components/react.cljs
```diff
@@ -39,6 +40,8 @@
 
 (def linear-gradient (reagent/adapt-react-class LinearGradient))
 
+(def blur-view (reagent/adapt-react-class (.-BlurView blur)))
```
flexsurfer, line 43: just wondering, why do we want to use it? i mean we already have bottom sheets and popups, but why for wallet connect we want to use blur-view?

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+                                                    (reset! selected-account (merge {} account))
+                                                    (when on-select (on-select)))}
+     [react/view (styles/account-container color account-selected?)
+      [quo/text {:color :inverse
+                 :weight (if account-selected? :medium :regular)}
+       name]]]))
+
+(defn account-selector [accounts selected-account on-select]
+  [react/view styles/account-selector-container
+   [react/view
```
flexsurfer: redundant [react/view

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+    [quo/text {:size :small} (i18n/label :t/select-account)]
+    [list/flat-list {:data        accounts
+                     :key-fn      :address
+                     :render-fn   render-account
+                     :render-data {:selected-account selected-account
+                                   :on-select on-select}
+                     :horizontal  true
+                     :shows-horizontal-scroll-indicator false
+                     :extraData @selected-account
```
flexsurfer: what is extraData ?

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+  (if (> (count accounts) 1)
+    [react/view {:style styles/account-selector-wrapper}
+     [account-selector accounts selected-account on-select]]
+    [react/touchable-opacity {:style styles/single-account-container}
+     [toolbar-selection {:text (:name @selected-account)
+                         :background-color (:color @selected-account)
+                         :on-press on-press}]]))
+
+(defview success-sheet-view [{:keys [topic]}]
+  (letsubs [visible-accounts @(re-frame/subscribe [:visible-accounts-without-watch-only])
```
flexsurfer: if you use `letsubs` you don't need `@(re-frame/subscribe`

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+         (vector dapps-account)
+         selected-account
+         {:on-press #(do
+                       (re-frame/dispatch [:wallet-connect/manage-app session])
+                       (reset! show-account-selector? true))}]
+        [quo/text {:weight :regular
+                   :color :secondary
+                   :style  styles/message-title}
+         (i18n/label :t/manage-connections)]
+        [react/view (styles/footer)
```
flexsurfer: why so many nested views?

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+           :theme :secondary
+           :on-press #(re-frame/dispatch [:wallet-connect/disconnect topic])}
+          (i18n/label :t/disconnect)]]
+        [account-selector
+         visible-accounts
+         selected-account
+         #(re-frame/dispatch [:wallet-connect/change-session-account topic @selected-account])]]])))
+
+(defview session-proposal-sheet [{:keys [name icons]}]
+  (let [visible-accounts @(re-frame/subscribe [:visible-accounts-without-watch-only])
```
flexsurfer: if you use defview, it makes sense to use letsubs and remove `@(re-frame/subscribe`

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,215 @@
...
+        dapps-account @(re-frame/subscribe [:dapps-account])
+        icon-uri (when (and icons (> (count icons) 0)) (first icons))
+        selected-account (reagent/atom dapps-account)]
+    [react/view (styles/acc-sheet)
```
flexsurfer: again too many nested views

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,234 @@
...
+      (.pair client (clj->js {:uri data})))
+    (merge
+     {:dispatch [:navigate-back]}
+     (when wallet-connect-enabled?
+       {:db (assoc db :wallet-connect/scanned-uri data)}))))
+
+(fx/defn wallet-connect-client-initate
+  {:events [:wallet-connect/client-init]}
+  [{:keys [db] :as cofx} client]
+  (subscribe-to-events client)
```
flexsurfer: this should be in effect

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,234 @@
...
+             (assoc :wallet-connect/sessions (js->clj (.-values (.-session client)) :keywordize-keys true))
+             (dissoc :wallet-connect/session-managed))}))
+
+(fx/defn pair-session
+  {:events [:wallet-connect/pair]}
+  [{:keys [db]} {:keys [data]}]
+  (let [client (get db :wallet-connect/client)
+        wallet-connect-enabled? (get db :wallet-connect/enabled?)]
+    (when wallet-connect-enabled?
+      (.pair client (clj->js {:uri data})))
```
flexsurfer: effect

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,234 @@
...
+        (.then #(log/debug "[wallet-connect] session proposal approved"))
+        (.catch #(log/error "[wallet-connect] session proposal approval error:" %)))
+    {:hide-wallet-connect-sheet nil}))
+
+(fx/defn reject-proposal
+  {:events [:wallet-connect/reject-proposal]}
+  [{:keys [db]} account]
+  (let [client (get db :wallet-connect/client)
+        proposal (get db :wallet-connect/proposal)]
+    (-> ^js client
```
flexsurfer: effect

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,234 @@
...
+        proposal-chain-ids (map #(last (string/split % #":")) (:chains blockchain))
+        available-chain-ids (map #(get-in % [:config :NetworkId]) (vals (get db :networks/networks)))
+        supported-chain-ids (filter (fn [chain-id] #(boolean (some #{chain-id} available-chain-ids))) proposal-chain-ids)
+        address (:address account)
+        accounts (map #(str "eip155:" % ":" (ethereum/normalized-hex address)) supported-chain-ids)
+        ;; TODO: Check for unsupported
+        metadata (get db :wallet-connect/proposal-metadata)
+        response {:state {:accounts accounts}
+                  :metadata config/default-wallet-connect-metadata}]
+    (-> ^js client
```
flexsurfer: effect

flexsurfer (changes requested):

src/status_im/ui/components/react.cljs
briansztamfater: We use it for a blur effect here

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: We tell FlatList to re-render when the extraData changes

src/status_im/ui/components/react.cljs
flexsurfer: mhm thats interesting, thanks, i guess we have other places in the app where we could use it

flexsurfer: btw, why do we need `select account` sheet? why don't select it right on the connection sheet ?

flexsurfer: i mean its a cool idea to select account with horizontal scroll so you don't need additional sheet. but for some reason we do it in additional sheet

briansztamfater: It's more of a session management sheet, at that moment the session is connected but you can update the session account or disconnect. We'll also re-use it to manage the list of active sessions.

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: Thanks, I will review the views hierarchy and see how to improve it

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: Yes, absolutely, my bad. Fixed

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: Not specific reason in this case, just refactored it

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: Absolutely. Fixed

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: Fixed

src/status_im/wallet_connect/core.cljs
briansztamfater: Moved to effect

src/status_im/wallet_connect/core.cljs
briansztamfater: Moved to effect

src/status_im/wallet_connect/core.cljs
briansztamfater: Moved to effect (and so on... :P)

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,205 @@
...
+            sessions [:wallet-connect/sessions]
+            managed-session [:wallet-connect/session-managed]]
+    (let [{:keys [peer state] :as session} (first (filter #(= (:topic %) topic) sessions))
+          {:keys [accounts]} state
+          {:keys [metadata]} peer
+          {:keys [name icons]} metadata
+          icon-uri (when (and icons (> (count icons) 0)) (first icons))
+          address (last (string/split (first accounts) #":"))
+          account (first (filter #(= (:address %) address) visible-accounts))
+          selected-account (reagent/atom account)]
```
flexsurfer: i think we still don't have any agreement on how to name atoms, but it's better to somehow show that its an atom, `selected-account-atom` would work for me

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,205 @@
...
+(defview app-management-sheet-view [{:keys [topic]}]
+  (letsubs [sessions [:wallet-connect/sessions]
+            visible-accounts [:visible-accounts-without-watch-only]]
+    (let [{:keys [peer state]} (first (filter #(= (:topic %) topic) sessions))
+          {:keys [accounts]} state
+          {:keys [metadata]} peer
+          {:keys [name icons url]} metadata
+          icon-uri (when (and icons (> (count icons) 0)) (first icons))
+          account-address (last (string/split (first accounts) #":"))
+          selected-account (reagent/atom (first (filter #(= (:address %) account-address) visible-accounts)))]
```
flexsurfer: atom must be defined  outside the render function, don't remember if its play well with letsubs , should work I guess

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
```diff
@@ -0,0 +1,205 @@
...
+        [account-selector
+         visible-accounts
+         selected-account
+         #(re-frame/dispatch [:wallet-connect/change-session-account topic @selected-account])]]])))
+
+(defview session-proposal-sheet [{:keys [name icons]}]
+  (letsubs [visible-accounts [:visible-accounts-without-watch-only]
+            dapps-account [:dapps-account]]
+    (let [icon-uri (when (and icons (> (count icons) 0)) (first icons))
+          selected-account (reagent/atom dapps-account)]
```
flexsurfer: same here

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
flexsurfer: why do you need to wrap it with the atom btw ?

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,277 @@
...
+            [status-im.utils.wallet-connect :as wallet-connect]
+            [status-im.browser.core :as browser]
+            [taoensso.timbre :as log]
+            [status-im.async-storage.core :as async-storage]
+            [status-im.utils.config :as config]))
+
+(fx/defn switch-wallet-connect-enabled
+  {:events [:multiaccounts.ui/switch-wallet-connect-enabled]}
+  [{:keys [db]} enabled?]
+  (when enabled?
```
flexsurfer: should be effect

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,277 @@
...
+   :db (cond-> db
+         (not enabled?)
+         (dissoc :wallet-connect/client)
+         :always
+         (assoc :wallet-connect/enabled? enabled?))})
+
+(fx/defn proposal-handler
+  {:events [:wallet-connect/proposal]}
+  [{:keys [db] :as cofx} request-event]
+  (let [proposal (js->clj request-event :keywordize-keys true)
```
flexsurfer: we have types/js->cljs

src/status_im/wallet_connect/core.cljs
```diff
@@ -0,0 +1,277 @@
...
+                                {:tx-obj  (-> params
+                                              first
+                                              (update :from #(or % dapps-address))
+                                              (dissoc :gasPrice))})
+                              {:on-result [:wallet-connect.dapp/transaction-on-result message-id topic]
+                               :on-error  [:wallet-connect.dapp/transaction-on-error message-id topic]}))))
+      (when (#{"eth_accounts" "eth_coinbase"} method)
+        (wallet-connect-complete-transaction cofx message-id topic (if (= method "eth_coinbase") dapps-address [dapps-address]))))))
+
+(def permissioned-method
```
flexsurfer: same methods as for dapps ? why copy them ?

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: We need to get the session account and wrap it with the atom so user can select another one and pass it to extraData props of flat list so Flat List re-renders when value of the atom changes.

src/status_im/wallet_connect/core.cljs
briansztamfater: I think is not used, will delete it

src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs
briansztamfater: I would leave this as is if possible, we need to initialize the atom with the session address value and pass it to the flat list to work properly. Any suggestions on how to do this without defining in `let`?

flexsurfer: we can leave it because its not critical here, but in proper way you have to create extra function pass params in it and in that function define atom outside render function

flexsurfer (approved):
Great job!

qoqobolo:
Hey @briansztamfater!
Unfortunately, nothing happens when scanning QR now.
Can you take a look, please? 

https://user-images.githubusercontent.com/67952253/158400810-b5c83f87-af12-4123-830d-941ea95f8608.MP4

briansztamfater:
> Hey @briansztamfater! Unfortunately, nothing happens when scanning QR now. Can you take a look, please?
> 
>  IMG_1081.MP4

May be related to wallet registry not published yet on wallet connect. Will check with @jakubgs and come back to you

jakubgs:
Still pending review:

image

briansztamfater:
@qoqobolo actually it was my fault, I left some hardcoded uri that I used to test, now you can try again. Sorry about that! cc @jakubgs

qoqobolo:
Thanks @briansztamfater, works great now.
PR can be merged.

PR #12705, opened 2021-10-13
Feature/jump to message

cammellos:


flexsurfer:
in the current state this PR works very randomly:
1) sometimes it doesn't load more messages to the beginning
2) sometimes it autoscroll down and doesn't show the target message
3) sometimes it doesn't load more messages to the end
4) sometimes it just loads all messages to the end
5) sometimes it skips messages

https://user-images.githubusercontent.com/11790366/144386223-b9a0bf66-9983-4605-a4d0-a8d94bec74e9.mov


https://user-images.githubusercontent.com/11790366/144386244-55ddf051-b213-4ac5-8bed-06f8db742176.mov

PR #12716, opened 2021-10-18
Online status indicator

### LINKED ISSUE #12547: Add online status indicator
As a user
I want to be able to be able to set my online status
So that other users can see whether I am online

### Description

We should allow a user to set their own status.
By default that should be "invisible", this needs to be changed in status-go, otherwise it's ready in status-go.

### Designs

https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=7253%3A27195

Parveshdhull:
fixes #12547

### Summary
PR implements Online Status Indicator
<div align="center">
Screenshot
Screenshot
Screenshot
</div>

PR follows these design guidelines: Designs [[1st]](https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=7253%3A27195) [[2nd]](https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=77%3A19108)
#### status-go PR:
https://github.com/status-im/status-go/pull/2408

#### Code
There were some residue code in [styles.cljs](https://github.com/status-im/status-react/pull/12706/files#diff-f54bb761ef81cb3f9f1af35f7080f4edc667f1e88fb816755a5d14d7d442992d) file from 2016 implementation of online status indicator. Currently, it is not used anywhere, so removed to avoid confusion between styles.

#### Other features:
- Status updates syncing between devices
- Offline status updates
- Sort community members & contacts based on online status and name

#### Known Issues:
- Gap between each ping is 5 minutes. So it has its own side effects for ex. the user seems online a little longer, etc.
- Offline capability works great with the single device it will broadcast its current status whenever the device becomes online, even after app restart. 
But there is one rare case where it will not work:
i) User is using multiple devices
ii) Another device is online and broadcasting different status than saved by the main device during the offline state. (with larger clock value than saved clock)
iii) Once the main device gets online it receives broadcast from another device before broadcasting saved status. 

Thanks

### DIFF
--- a/src/status_im/constants.cljs
+++ b/src/status_im/constants.cljs
@@ -162,3 +162,9 @@
 (def ^:const activity-center-notification-type-private-group-chat 2)
 (def ^:const activity-center-notification-type-mention 3)
 (def ^:const activity-center-notification-type-reply 4)
+
+(def ^:const visibility-status-unknown 0)
+(def ^:const visibility-status-automatic 1)
+(def ^:const visibility-status-dnd 2)
+(def ^:const visibility-status-always-online 3)
+(def ^:const visibility-status-inactive 4)
--- a/src/status_im/data_store/settings.cljs
+++ b/src/status_im/data_store/settings.cljs
@@ -1,7 +1,8 @@
 (ns status-im.data-store.settings
   (:require
    [status-im.utils.config :as config]
-   [status-im.ethereum.eip55 :as eip55]))
+   [status-im.ethereum.eip55 :as eip55]
+   [status-im.data-store.visibility-status-updates :as visibility-status-updates]))
 
 (defn rpc->networks [networks]
   (reduce (fn [acc {:keys [id] :as network}]
@@ -50,4 +51,5 @@
       (update :link-previews-enabled-sites set)
       (update :custom-bootnodes rpc->custom-bootnodes)
       (update :custom-bootnodes-enabled? rpc->custom-bootnodes)
-      (update :currency keyword)))
+      (update :currency keyword)
+      (visibility-status-updates/<-rpc-settings)))
--- a/src/status_im/db.cljs
+++ b/src/status_im/db.cljs
@@ -32,6 +32,7 @@
              :tooltips                           {}
              :dimensions/window                  (dimensions/window)
              :registry                           {}
+             :visibility-status-updates          {}
              :stickers/packs-owned               #{}
              :stickers/packs-pending             #{}
              :keycard                            {:nfc-enabled?   false
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -40,7 +40,8 @@
             [status-im.notifications.core :as notifications]
             [status-im.utils.currency :as currency]
             [clojure.set :as clojure.set]
-            [quo.design-system.colors :as colors]))
+            [quo.design-system.colors :as colors]
+            [status-im.ui.screens.profile.visibility-status.utils :as visibility-status-utils]))
 
 ;; TOP LEVEL ===========================================================================================================
 
@@ -86,6 +87,7 @@
 (reg-root-key-sub :app-state :app-state)
 (reg-root-key-sub :home-items-show-number :home-items-show-number)
 (reg-root-key-sub :waku/v2-peer-stats :peer-stats)
+(reg-root-key-sub :visibility-status-updates :visibility-status-updates)
 
 ;;NOTE this one is not related to ethereum network
 ;; it is about cellular network/ wifi network
@@ -200,6 +202,7 @@
 (reg-root-key-sub :intro-wizard-state :intro-wizard)
 
 (reg-root-key-sub :popover/popover :popover/popover)
+(reg-root-key-sub :visibility-status-popover/popover :visibility-status-popover/popover)
 (reg-root-key-sub :add-account :add-account)
 
 (reg-root-key-sub :keycard :keycard)
@@ -293,6 +296,36 @@
    (get-in communities [id :chats])))
 
 (re-frame/reg-sub
+ :communities/community-members
+ :<- [:communities]
+ (fn [communities [_ id]]
+   (get-in communities [id :members])))
+
+(re-frame/reg-sub
+ :communities/sorted-community-members
+ (fn [[_ community-id]]
+   (let [contacts     (re-frame/subscribe [:contacts/contacts])
+         multiaccount (re-frame/subscribe [:multiaccount])
+         members      (re-frame/subscribe
+                       [:communities/community-members community-id])]
+     [contacts multiaccount members]))
+ (fn [[contacts multiaccount members] _]
+   (let [names (reduce
+                (fn [acc identity]
+                  (let [me?     (= (:public-key multiaccount)
+                                   identity)
+                        contact (when-not me?
+                                  (multiaccounts/contact-by-identity
+                                   contacts identity))
+                        name    (first
+                                 (multiaccounts/contact-two-names-by-identity
+                                  contact multiaccount identity))]
+                    (assoc acc identity name))) {} (keys members))]
+     (->> members
+          (sort-by #(get names (get % 0)))
+          (sort-by #(visibility-status-utils/visibility-status-order (get % 0)))))))
+
+(re-frame/reg-sub
  :communities/communities
  :<- [:communities/enabled?]
  :<- [:search/home-filter]
@@ -355,6 +388,11 @@
 
 ;;GENERAL ==============================================================================================================
 
+(re-frame/reg-sub
+ :visibility-status-updates/visibility-status-update
+ :<- [:visibility-status-updates]
+ (fn [visibility-status-updates [_ public-key]]
+   (get visibility-status-updates public-key)))
 
 (re-frame/reg-sub
  :multiaccount/logged-in?
@@ -783,6 +821,12 @@
          :seed
          (string/blank? (security/safe-unmask-data seed))
          false))))
+
+(re-frame/reg-sub
+ :multiaccount/current-user-visibility-status
+ :<- [:multiaccount]
+ (fn [{:keys [current-user-visibility-status]}]
+   current-user-visibility-status))
 
 ;;CHAT ==============================================================================================================
 
@@ -2209,6 +2253,15 @@
    (contact.db/get-active-contacts contacts)))
 
 (re-frame/reg-sub
+ :contacts/sorted-contacts
+ :<- [:contacts/active]
+ (fn [active-contacts]
+   (->> active-contacts
+        (sort-by :alias)
+        (sort-by
+         #(visibility-status-utils/visibility-status-order (:public-key %))))))
+
+(re-frame/reg-sub
  :contacts/active-count
  :<- [:contacts/active]
  (fn [active-contacts]
@@ -2265,8 +2318,7 @@
  :contacts/contact-by-identity
  :<- [:contacts/contacts]
  (fn [contacts [_ identity]]
-   (or (get contacts identity)
-       (multiaccounts/contact-with-names {:public-key identity}))))
+   (multiaccounts/contact-by-identity contacts identity)))
 
 (re-frame/reg-sub
  :contacts/contact-added?
@@ -2281,11 +2333,8 @@
    [(re-frame/subscribe [:contacts/contact-by-identity identity])
     (re-frame/subscribe [:multiaccount])])
  (fn [[contact current-multiaccount] [_ identity]]
-   (let [me? (= (:public-key current-multiaccount) identity)]
-     (if me?
-       [(or (:preferred-name current-multiaccount)
-            (gfycat/generate-gfy identity))]
-       (multiaccounts/contact-two-names contact false)))))
+   (multiaccounts/contact-two-names-by-identity contact current-multiaccount
+                                                identity)))
 
 (re-frame/reg-sub
  :contacts/contact-name-by-identity
--- a/src/status_im/ui/screens/communities/members.cljs
+++ b/src/status_im/ui/screens/communities/members.cljs
@@ -51,8 +51,9 @@
      {:title               first-name
       :subtitle            second-name
       :accessibility-label :member-item
-      :icon                [chat-icon/contact-icon-contacts-tab
-                            (multiaccounts/displayed-photo member)]
+      :icon                 [chat-icon/profile-photo-plus-dot-view
+                             {:public-key public-key
+                              :photo-path (multiaccounts/displayed-photo member)}]
       :accessory           (when (not= public-key my-public-key)
                              [quo/button {:on-press
                                           #(>evt [:bottom-sheet/show-sheet
@@ -89,17 +90,17 @@
   (let [{:keys [community-id]} (<sub [:get-screen-params])]
     (fn []
       (let [my-public-key               (<sub [:multiaccount/public-key])
-            {:keys [members
-                    permissions
-                    can-manage-users?]} (<sub [:communities/community community-id])]
+            {:keys [permissions
+                    can-manage-users?]} (<sub [:communities/community community-id])
+            sorted-members              (<sub [:communities/sorted-community-members
+                                               community-id])]
         [:<>
          [topbar/topbar {:title    (i18n/label :t/community-members-title)
-
-                         :subtitle (str (count members))}]
+                         :subtitle (str (count sorted-members))}]
          [header community-id]
          (when (and can-manage-users? (= constants/community-on-request-access (:access permissions)))
            [requests-to-join community-id])
-         [rn/flat-list {:data        (keys members)
+         [rn/flat-list {:data        (keys sorted-members)
                         :render-data {:community-id community-id
                                       :my-public-key my-public-key
                                       :can-kick-users? (and can-manage-users?
--- a/src/status_im/ui/screens/contacts_list/views.cljs
+++ b/src/status_im/ui/screens/contacts_list/views.cljs
@@ -15,8 +15,9 @@
     [quo/list-item
      {:title    first-name
       :subtitle second-name
-      :icon     [chat-icon.screen/contact-icon-contacts-tab
-                 (multiaccounts/displayed-photo contact)]
+      :icon     [chat-icon.screen/profile-photo-plus-dot-view
+                 {:public-key public-key
+                  :photo-path (multiaccounts/displayed-photo contact)}]
       :chevron  true
       :on-press #(re-frame/dispatch [:chat.ui/show-profile public-key])}]))
 
@@ -30,7 +31,7 @@
 
 (defview contacts-list []
   (letsubs [blocked-contacts-count [:contacts/blocked-count]
-            contacts      [:contacts/active]]
+            sorted-contacts        [:contacts/sorted-contacts]]
     [react/scroll-view {:flex 1}
      [add-new-contact]
      (when (pos? blocked-contacts-count)
@@ -44,9 +45,9 @@
           :accessory           :text
           :accessory-text      blocked-contacts-count
           :on-press            #(re-frame/dispatch [:navigate-to :blocked-users-list])}]])
-     (if (seq contacts)
+     (if (seq sorted-contacts)
        [list.views/flat-list
-        {:data                      contacts
+        {:data                      sorted-contacts
          :key-fn                    :address
          :render-fn                 contacts-list-item}]
        [react/view {:align-items     :center
--- a/src/status_im/ui/screens/profile/visibility_status/styles.cljs
+++ b/src/status_im/ui/screens/profile/visibility_status/styles.cljs
@@ -1,0 +1,76 @@
+(ns status-im.ui.screens.profile.visibility-status.styles
+  (:require [quo.design-system.colors :as colors]))
+
+(defn visibility-status-button-container []
+  {:background-color       (:interactive-02 @colors/theme)
+   :margin-left            16
+   :border-radius          16
+   :border-top-left-radius 4
+   :align-self             :flex-start
+   :flex-direction         :row
+   :align-items            :center
+   :justify-content        :center
+   :padding                6
+   :padding-right          12})
+
+(defn visibility-status-dot [dot-color size]
+  {:background-color dot-color
+   :width            size
+   :height           size
+   :border-radius    (/ size 2)
+   :border-width     1
+   :border-color     colors/white})
+
+(defn visibility-status-profile-dot [color size border-width margin-left]
+  (merge (visibility-status-dot color size)
+         {:margin-right 6
+          :margin-left  margin-left
+          :border-width border-width}))
+
+(defn visibility-status-text []
+  {:color      (:text-01 @colors/theme)
+   :font-size  16
+   :text-align :center})
+
+(defn visibility-status-subtitle []
+  {:color        (:text-02 @colors/theme)
+   :font-size    16
+   :margin-left  22
+   :margin-right 6})
+
+(defn visibility-status-option []
+  {:flex-direction :row
+   :align-items    :center})
+
+(defn visibility-status-options [scale position]
+  {:background-color       (if (colors/dark?)
+                             (:interactive-02 @colors/theme)
+                             colors/white)
+   :border-radius          16
+   :border-top-left-radius 4
+   :justify-content        :center
+   :align-self             :flex-start
+   :left                    16
+   :top                    -76
+   :padding-bottom          6
+   :padding-top             8
+   :transform               [{:scaleY scale}
+                             {:translateY position}]})
+
+(defn visibility-status-popover-container []
+  {:position :absolute
+   :top      0
+   :bottom   0
+   :left     0
+   :right    0})
+
+(defn visibility-status-popover-ios-backdrop [alpha-value]
+  {:flex             1
+   :background-color colors/black-persist
+   :opacity          alpha-value})
+
+(defn visibility-status-popover-child-container [window-height]
+  {:position  :absolute
+   :height    window-height
+   :left      0
+   :right     0})
--- a/src/status_im/utils/datetime.cljs
+++ b/src/status_im/utils/datetime.cljs
@@ -20,6 +20,8 @@
 (def hour (* 60 minute))
 (def day (* 24 hour))
 (def week (* 7 day))
+(defn weeks [w]
+  (* w week))
 (def units [{:name :t/datetime-second-short :limit 60 :in-second 1}
             {:name :t/datetime-minute-short :limit 3600 :in-second 60}
             {:name :t/datetime-hour-short :limit 86400 :in-second 3600}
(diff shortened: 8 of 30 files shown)


src/status_im/status_updates/core.cljs
```diff
@@ -0,0 +1,103 @@
...
+                      (handle-my-status-updates prev-map my-current-status clock status-update)
+                      (handle-other-status-updates prev-map public-key clock status-update))))
+                {:status-updates      status-updates-old
+                 :current-user-status my-current-status
+                 :dispatch            nil}
+                status-updates-received)]
+    (merge {:db (-> db
+                    (update-in [:status-updates] merge status-updates)
+                    (update-in [:multiaccount :current-user-status] merge current-user-status))}
+           (if (nil? dispatch) {} {:dispatch dispatch}))))
```
rasom: could be just `(when dispatch {:dispatch dispatch})`. It is shorter although still dispatch dispatch dispatch :D

src/status_im/status_updates/core.cljs
```diff
@@ -0,0 +1,103 @@
...
+
+(fx/defn initialize-status-updates [cofx]
+  (status-updates-store/fetch-status-updates-rpc cofx #(re-frame/dispatch [::status-updates-loaded %])))
+
+(fx/defn visibility-status-option-pressed
+  {:events [:status-updates/visibility-status-option-pressed]}
+  [_ status-type]
+  (let [events-to-dispatch-later (atom [])]
+    (swap! events-to-dispatch-later conj {:ms 10 :dispatch [:status-updates/update-visibility-status status-type]})
```
rasom: When it is possible, please, keep lines' length shorter/equal 80 chars.

src/status_im/utils/datetime.cljs
```diff
@@ -167,3 +169,6 @@
     (str day (or (s (mod (- m 20) 10))
                  (s m)
                  (s 0)))))
+
+(defn to-mills [sec]
```
rasom: That's very subjective, but imo `to-ms` would look better.

src/status_im/status_updates/core.cljs
Parveshdhull: Great Tip 🙏, now I enabled Emacs auto highlight for longer lines. Thanks

src/status_im/utils/datetime.cljs
Parveshdhull: Yes, this looks much better 😅

rasom (approved):

src/status_im/constants.cljs
```diff
@@ -162,3 +162,9 @@
 (def ^:const activity-center-notification-type-private-group-chat 2)
 (def ^:const activity-center-notification-type-mention 3)
 (def ^:const activity-center-notification-type-reply 4)
+
+(def ^:const visibility-status-unknown 0)
+(def ^:const visibility-status-automatic 1)
+(def ^:const visibility-status-dnd 2)
+(def ^:const visibility-status-always-online 3)
+(def ^:const visibility-status-inactive 4)
```
flexsurfer, line 170: what about invisible ?

src/status_im/data_store/settings.cljs
```diff
@@ -50,4 +51,5 @@
       (update :link-previews-enabled-sites set)
       (update :custom-bootnodes rpc->custom-bootnodes)
       (update :custom-bootnodes-enabled? rpc->custom-bootnodes)
-      (update :currency keyword)))
+      (update :currency keyword)
+      (update :current-user-status status-updates/<-rpc)))
```
flexsurfer: we have status tab. so probably we should use something like `user-online-status` or smth

src/status_im/db.cljs
```diff
@@ -32,6 +32,7 @@
              :tooltips                           {}
              :dimensions/window                  (dimensions/window)
              :registry                           {}
+             :status-updates                     {}
```
flexsurfer: this might be confused with statuses (timeline)

src/status_im/visibility_status_popover/core.cljs
```diff
@@ -0,0 +1,21 @@
+(ns status-im.visibility-status-popover.core
```
flexsurfer, line 1: I'm not sure if it is worth it to use a separate overlay for this, probably better to just use absolute layout

src/status_im/constants.cljs
Parveshdhull: Hi @flexsurfer, [originally](https://github.com/status-im/status-react/pull/12706), I had invisible. But that created conflicts with the desktop's [designs](https://www.figma.com/file/IPpvkpDWabBKJTeo6bFop0/Kuba%E2%8E%9CDesktop?node-id=77%3A19108).

src/status_im/data_store/settings.cljs
Parveshdhull: `user-online-status` will also create some confusion because there is also a status type as online. I will use `visibility-status-updates` instead of plain `status-updates` 👍

src/status_im/visibility_status_popover/core.cljs
Parveshdhull: I tried with the absolute layout too, but I was unable to make it the same as [designs](https://www.figma.com/file/TNCyHKtR3sx5EL6YznFWUa4O/Profile?node-id=7253%3A27195). The design requires a full overlay which also closes this on touch.

flexsurfer: ok, thanks

src/status_im/constants.cljs
flexsurfer: I'm more interested in the UX, by default I would expect that my status is not propagated, is that the case ?

src/status_im/data_store/settings.cljs
du82: > we have status tab. so probably we should use something like `user-online-status` or smth

I think we should rename the Status tab to "Moments" because it's more in line with WeChat, and it's less confusing to have a Status tab in the Status app.

Parveshdhull:
New Changes:
- Added syncing functionality - [Code](https://github.com/status-im/status-react/pull/12716/files#diff-44579badd82775c385690432066be20672d359b6c15a9b305ccbc42794506dfe) - L157
- Offline Status Updates - [Code](https://github.com/status-im/status-react/pull/12716/files#diff-44579badd82775c385690432066be20672d359b6c15a9b305ccbc42794506dfe) - L138
- `status-update(s)` changed to `visibility-status-update(s)` except for some 3-4 occurrence related to status-go calls. They are currently also used by desktop
- Fixed minor design bugs

src/status_im/data_store/settings.cljs
```diff
@@ -50,4 +51,5 @@
       (update :link-previews-enabled-sites set)
       (update :custom-bootnodes rpc->custom-bootnodes)
       (update :custom-bootnodes-enabled? rpc->custom-bootnodes)
-      (update :currency keyword)))
+      (update :currency keyword)
+      (update :current-user-status visibility-status-updates/<-rpc)))
```
flexsurfer: :current-user-**visibility**-status ?

src/status_im/ui/screens/profile/visibility_status/styles.cljs
```diff
@@ -0,0 +1,70 @@
+(ns status-im.ui.screens.profile.visibility-status.styles
+  (:require [quo.design-system.colors :as colors]))
+
+(def color-error "#FA6565")
```
flexsurfer: should we move this to color ns ?

src/status_im/ui/screens/profile/visibility_status/styles.cljs
```diff
@@ -0,0 +1,70 @@
...
+(def color-online "#7CDA00")
+(def color-dnd "#FA6565")
+(def color-inactive "#939BA1")
+
+(defn visibility-status-button-container []
+  {:background-color       (:interactive-02 @colors/theme)
+   :margin-left            16
+   :border-radius          16
+   :border-top-left-radius 4
+   :align-self             "flex-start"
```
flexsurfer: better to use keywords :flex-start

src/status_im/ui/screens/profile/visibility_status/styles.cljs
```diff
@@ -0,0 +1,70 @@
...
+   :padding                6
+   :padding-right          12})
+
+(defn visibility-status-dot [dot-color size]
+  {:background-color dot-color
+   :width            size
+   :height           size
+   :border-radius    (/ size 2)
+   :border-width     1
+   :border-color     "#FFFFFF"})
```
flexsurfer: colors/white

qoqobolo:
@Parveshdhull thanks for your work!


### ISSUE 1: The screen shifts down when pressing the `Online status` button on iOS

https://user-images.githubusercontent.com/67952253/138434694-6a289b92-dbaf-4339-88f6-9c3ecb42b133.mp4

### ISSUE 2: According to the [feature request description](https://github.com/status-im/status-react/issues/12547) the default status should be `Invisible` (`Inactive`)

The current default status is `Automatic`. 
Please, let me know if this requirement has been changed.

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+  {:events [::visibility-status-updates-loaded]}
+  [{:keys [db]} visibility-status-updates-loaded]
+  (let [{:keys [visibility-status-updates]}
+        (reduce (fn [prev-map visibility-status-update-loaded]
+                  (let [{:keys [public-key] :as visibility-status-update}
+                        (visibility-status-updates-store/<-rpc
+                         visibility-status-update-loaded)]
+                    (assoc-in prev-map [:visibility-status-updates public-key]
+                              visibility-status-update)))
+                {:visibility-status-updates {}}
```
flexsurfer: why dont just have {} here ? why do you need `:visibility-status-updates` key ?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+            [status-im.utils.fx :as fx]
+            [status-im.constants :as constants]
+            [status-im.multiaccounts.update.core :as multiaccounts.update]
+            [status-im.utils.datetime :as datetime]))
+
+(fx/defn load-visibility-status-updates
+  {:events [::visibility-status-updates-loaded]}
+  [{:keys [db]} visibility-status-updates-loaded]
+  (let [{:keys [visibility-status-updates]}
+        (reduce (fn [prev-map visibility-status-update-loaded]
```
flexsurfer: usually prev-map named as `acc` - accumulator

Parveshdhull:
Hi @qoqobolo, I will check why ios behaving like this, but yes default settings is changed, you can find more details in this [conversation](https://discord.com/channels/624205794384281629/697843216048717905/898490235040710677)

qoqobolo:
@Parveshdhull 

### ISSUE 3: Error when switching to Mainnet 
**Steps:**
1. Create multiaccount
2. Advanced > Network > Mainnet > Confirm

**OS**: Android, iOS

Caught by autotests. Take a look, please.

Screenshot 2021-10-22 at 13 02 24

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+                    (assoc-in prev-map [:visibility-status-updates public-key]
+                              visibility-status-update)))
+                {:visibility-status-updates {}}
+                visibility-status-updates-loaded)]
+    {:db (assoc db :visibility-status-updates visibility-status-updates)}))
+
+(defn handle-my-visibility-status-updates
+  [prev-map my-current-status clock visibility-status-update]
+  (let [status-type (:status-type visibility-status-update)]
+    (when (or
```
flexsurfer: dont you need to return prev-map for else case ?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+                visibility-status-updates-received)]
+    (merge {:db (-> db
+                    (update-in [:visibility-status-updates]
+                               merge visibility-status-updates)
+                    (update-in [:multiaccount :current-user-status]
+                               merge current-user-status))}
+           (when dispatch {:dispatch dispatch}))))
+
+(fx/defn initialize-visibility-status-updates [cofx]
+  (visibility-status-updates-store/fetch-visibility-status-updates-rpc
```
flexsurfer: why not call this method directly ?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+   cofx #(re-frame/dispatch [::visibility-status-updates-loaded %])))
+
+(fx/defn visibility-status-option-pressed
+  {:events [:visibility-status-updates/visibility-status-option-pressed]}
+  [{:keys [db]} status-type]
+  (let [peers-count              (:peers-count db)
+        events-to-dispatch-later (atom [])]
```
flexsurfer: this doesn't look clojure idiomatic, whats the point of using atom here ?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+  [{:keys [db]} public-key clock]
+  (let [current-clock (get-in db [:visibility-status-updates public-key :clock] 0)]
+    (when (= current-clock clock)
+      {:db (update-in db [:visibility-status-updates public-key]
+                      merge {:status-type constants/visibility-status-inactive})})))
+
+(fx/defn countdown-for-online-user
+  {:events [:visibility-status-updates/countdown-for-online-user]}
+  [{:keys [db]} public-key clock ms]
+  {:dispatch-later
```
flexsurfer: what is the point of having dispach-later inside dispatch?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,168 @@
...
+  (let [send-visibility-status-updates?
+        (get-in db [:multiaccount :send-status-updates?])
+        status-type
+        (get-in db [:multiaccount :current-user-status :status-type])]
+    (when (and
+           (> peers-count 0)
+           send-visibility-status-updates?
+           (= status-type constants/visibility-status-inactive))
+      {:dispatch-later
+       [{:ms 100
```
flexsurfer: why dispatch-later and magic numbers 100 and 1000 ?

flexsurfer (changes requested):

src/status_im/ui/screens/profile/visibility_status/styles.cljs
Parveshdhull: Hi, can you please take a look at colors.cljs. It has the same value for white and white-persist and there is no pure white. That's why I hard-coded this value.

src/status_im/visibility_status_updates/core.cljs
Parveshdhull: I was trying to keep `login/core.cljs` clean and use the same pattern used earlier, for ex.  `(contact/initialize-contacts)`

src/status_im/ui/screens/profile/visibility_status/styles.cljs
flexsurfer: what do you mean pure white ?

Parveshdhull: Pure White - #FFFFFF
current white is not actually white, it is changing based on theme

src/status_im/visibility_status_updates/core.cljs
Parveshdhull: I think the first dispatch-later can be changed into dispatch but the second one is stopping broadcasts once the current status is broadcast-ed. 1000 was just a safety net, can be changed into 100 or something I guess 🤔

src/status_im/ui/screens/profile/visibility_status/styles.cljs
flexsurfer: if you want white for both dark and light use white-persist  then

Parveshdhull: Oh sorry, I was confused between theme and light theme. I will user colors/white, instead of hard coding 👍

Parveshdhull: still it is little weird that white and white-persist are exactly same

src/status_im/visibility_status_updates/core.cljs
Parveshdhull: Now as I think, accessing directly also looks great 🤔, will change this too 👍 Thanks

Parveshdhull:
Hi @qoqobolo , Thank you very much for testing PR 🙏.
I fixed the above issues and also added another functionality.
`Sort community members & contacts based on online status and name`
Please let me know if I can improve anything else. Thanks

src/status_im/data_store/settings.cljs
```diff
@@ -50,4 +51,5 @@
       (update :link-previews-enabled-sites set)
       (update :custom-bootnodes rpc->custom-bootnodes)
       (update :custom-bootnodes-enabled? rpc->custom-bootnodes)
-      (update :currency keyword)))
+      (update :currency keyword)
+      (visibility-status-updates/<-rpc-settings)))
```
flexsurfer, line 55: why is it different from others ? i mean why not `update`

src/status_im/ui/screens/contacts_list/views.cljs
```diff
@@ -28,6 +30,14 @@
     :accessibility-label :add-new-contact-button
     :on-press            #(re-frame/dispatch [:open-modal :new-contact])}])
 
+(defn visibility-status-order [contact]
+  (visibility-status-utils/visibility-status-order (:public-key contact)))
+
+(defn sorted-contacts-list [contacts]
```
flexsurfer: its better to create a new subscription, views must be dumb

src/status_im/ui/screens/communities/members.cljs
```diff
@@ -85,16 +87,27 @@
                      :title          (i18n/label :t/membership-requests)}]
      [quo/separator {:style {:margin-vertical 8}}]]))
 
+(defn member-name [member]
+  (<sub [:contacts/contact-name-by-identity (get member 0)]))
+
+(defn visibility-status-order [member]
+  (visibility-status-utils/visibility-status-order (get member 0)))
+
+(defn sorted-members-list [members]
```
flexsurfer: better to move to subscriptions

src/status_im/ui/screens/communities/members.cljs
```diff
@@ -85,16 +87,27 @@
                      :title          (i18n/label :t/membership-requests)}]
      [quo/separator {:style {:margin-vertical 8}}]]))
 
+(defn member-name [member]
+  (<sub [:contacts/contact-name-by-identity (get member 0)]))
```
flexsurfer: this looks not correct, better to have one subscription

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+(defn process-visibility-status-update
+  [acc {:keys [public-key clock] :as visibility-status-update}]
+  (let [dispatch-later (get acc :dispatch-later [])
+        {:keys [real-status-type time-left]}
+        (utils/calculate-real-status-type-and-time-left visibility-status-update)]
+    (cond-> (assoc-in acc
+                      [:visibility-status-updates public-key]
+                      visibility-status-update)
+      (= real-status-type constants/visibility-status-automatic)
+      (assoc :dispatch-later
```
flexsurfer: why not update ?

flexsurfer: and still not sure why dispatch-later is used here ?

src/status_im/visibility_status_updates/core.cljs
```diff
@@ -0,0 +1,179 @@
...
+  [{:keys [db]} peers-count]
+  (let [send-visibility-status-updates?
+        (get-in db [:multiaccount :send-status-updates?])
+        status-type
+        (get-in db [:multiaccount :current-user-visibility-status :status-type])]
+    (when (and
+           (> peers-count 0)
+           send-visibility-status-updates?
+           (= status-type constants/visibility-status-inactive))
+      {:dispatch [:visibility-status-updates/update-visibility-status status-type]
```
flexsurfer: can't we use update-visibility-status handler here instead of dispatch?

src/status_im/data_store/settings.cljs
Parveshdhull: Hi, data from go is returned as

``` 
-current-user-status
	- statusType
	- clock
```

I can use update as `(update :current-user-status visibility-status-updates/<-rpc)`, but that will only change `statusType` into `status-type`
I also want to change key `current-user-status` into `current-user-visibility-status` and `<-rpc-settings` does both those things in one call

src/status_im/visibility_status_updates/core.cljs
Parveshdhull: 1. assoc is used because dispatch-later is initially nil and update will through error
2. User is online for 5 minutes, dispatch-later is used to make him/her offline after 5 minutes

flexsurfer: 1. clojure doesn't through errors ;) it will work fine even if dispatch-later  is nil
2. hm now its even more confusing, could you please elaborate

Parveshdhull: I also tried just checking if the online status is 5 minutes old directly on view, that works great, but the view doesn't re-render after 5 minutes, So I need to use dispatch-later to expire that status.

Parveshdhull: Sure,
There are 3 status types
1. **Inactive** - User will always displayed as offline, and no ping is received - _Anonymous_
2. **Always Online** - User will always displayed as online, and online ping is received every 5 minutes (when sender is online) - _Pseudo Anonymous_
3. **Automatic** - User will only be shown online if there is a online ping in last 5 minutes, otherwise offline

So for `automatic`, to make sure the user is online or not, I have to make sure if ping received with the clock is in last 5 minutes, and I have to convert that status into offline, after whatever time is left between `now` and `received clock`

Parveshdhull: Yes it is confusing :D, you can find more details in [specs](https://github.com/status-im/status-go/pull/2408)

flexsurfer: does that mean that for every process-visibility-status-update you 'll create a new dispatch later ? but you probably need to remove old dispatches ?

Parveshdhull: Actually, another user is broadcasting at the rate of 5 minutes and this is dispatch-later for 5.05 minutes. So there will not be much issue of having many pending dispatch-laters because this one will expire whenever new will receive.

And also this dispatch checks clock of current status update before converting it into offline. If a new status is received, then it will not do anything.

So in a way there is only one dispatch-later per user, that keeps getting renewed every 5 minutes if another user is online and destroys after one run if he/she is not

Also dispatch-later will not be registered for inactive users, always online, and expired automatic. So it will only be registered for a fraction of user's, who are online at that moment.

src/status_im/subs.cljs
```diff
@@ -294,6 +295,16 @@
  (fn [communities [_ id]]
    (get-in communities [id :chats])))
 
+(re-frame/reg-sub
+ :communities/sorted-community-members
+ :<- [:communities]
+ (fn [communities [_ id]]
+   (->> (get-in communities [id :members])
+        (sort-by (fn [member]
+                   @(re-frame/subscribe
```
flexsurfer: unfortunately, this won't work properly, we have to create a new subscription :contacts/contact-names-by-identitis

src/status_im/subs.cljs
```diff
@@ -296,13 +296,22 @@
...
-                     [:contacts/contact-name-by-identity (get member 0)])))
+   (get-in communities [id :members])))
+
+(re-frame/reg-sub
+ :communities/sorted-community-members
+ (fn [[_ community-id]]
+   (let [members (re-frame/subscribe
+                  [:communities/community-members community-id])
+         names   (re-frame/subscribe
+                  [:contacts/contact-names-by-identities (keys @members)])]
```
flexsurfer: :communities/community-members should be inside contact-names-by-identities

flexsurfer: or you don't need contact-names-by-identities then you could move its code here

Parveshdhull: Yes, better to remove `contact-names-by-identities`, because this function also requires members

src/status_im/subs.cljs
```diff
@@ -292,6 +295,37 @@
...
+   (get-in communities [id :members])))
+
+(re-frame/reg-sub
+ :communities/sorted-community-members
+ (fn [[_ community-id]]
+   (let [contacts     (re-frame/subscribe [:contacts/contacts])
+         multiaccount (re-frame/subscribe [:multiaccount])
+         members      (re-frame/subscribe
+                       [:communities/community-members community-id])
+         names        (atom
```
flexsurfer: why this reduce here with atom but not in the subscription function?

flexsurfer (approved):

flexsurfer:
Great job!

qoqobolo:
@Parveshdhull Thank you for this awesome work!

There are two small design issues left

### ISSUE 4: Online indicator is duplicated when opened on Android
Can we have this drop-down menu in the way as on iOS? 
Screenshot 2021-10-29 at 12 06 48

### ISSUE 5: The background of the drop-down menu blends in with the background of the screen in dark mode on iOS
It would be better to make the drop-down menu background a dark gray if possible

UPD: skipped, since the issue is only reproducible on 1 device (Google Pixel 2) out of 5.
Screenshot 2021-10-29 at 12 07 08

qoqobolo:
@Parveshdhull another unexpected behavior found

### ISSUE 6: When a new message is received in 1-1 chat from a non-contacts user, the sender's online status remains `offline` until the recipient adds the sender to contacts or re-logs in

Sometimes, both users need to re-login in order for the sender's status to change to `Online`.


**Steps:**

UserA and UserB have default the `Automatic` status state
_UserA (sender)_:
- Start 1-1 chat with UserB
- Send a message

_UserB (receiver)_:
- Open the Activity center
- Open the 1-1 chat
- Send a message

**Expected behavior:** the online status of UserB is displayed as `Online` once the 1-1 chat is opened/some message is sent by UserB
**Actual behavior:** UserB's status remains `Offline`

https://user-images.githubusercontent.com/67952253/139449454-003e9b9f-6d4f-4631-af8a-646cf3892d07.mp4

**Logs:**
[geth.log](https://github.com/status-im/status-react/files/7442614/geth.log)
[Status.log](https://github.com/status-im/status-react/files/7442615/Status.log)

qoqobolo:
Update on ISSUE 6: the described behavior was found to be expected

Tested:
- on iOS and Android devices
- all the states
- offline status updates (with ~5 minutes delay, as expected)
- syncing between devices
- sort community members & contacts based on online status and name
- displaying status on different screens (1-1 chat, home screen, community members, contacts, user profile, activity center, blocked users)
- with PN background service enabled
- send messages with different options (while `Automatic`, `Always online`, `Inactive`)
- offline/cellular network
- dark mode

@Parveshdhull let's merge!

PR #12746, opened 2021-10-25
fix chat toolbar

flexsurfer:
fix chat toolbar


Parveshdhull (approved):
Works great 👍, Thanks

PR #12752, opened 2021-10-26
[#12750] Allow connecting ens names for imported accounts

### LINKED ISSUE #12750: Show warning when adding ENS name for wallet account
# Feature Issue

Indicate to the user that they cannot add an ENS name for a added wallet account, since that's now the account from which chat identity is derived.

### User Story

As a user I don't want to be confused by inability to configure an ENS name because it is owned by a wallet account I've added to the app rather than the main account. I want to see a warning explaining that wallet accounts can't make use of the ENS name.

### Description

Since we check ownership of ENS record against only the main account if the owner is an added wallet it just says we are not the owner, making the user confused. We should be able to easily detect that the owner isn't the main account but rather one of the added wallet accounts and explain to the user why that won't work.

### Acceptance Criteria

Want me to restate what I already wrote in like 3 different ways?

### Notes

Nope.

flexsurfer:
fixes #12750

### DIFF
--- a/src/status_im/ens/core.cljs
+++ b/src/status_im/ens/core.cljs
@@ -76,15 +76,17 @@
 
 (fx/defn set-state
   {:events [::name-resolved]}
-  [{:keys [db]} username state]
+  [{:keys [db]} username state address]
   (when (= username
            (get-in db [:ens/registration :username]))
-    {:db (assoc-in db [:ens/registration :state] state)}))
+    {:db (-> db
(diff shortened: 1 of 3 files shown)


Parveshdhull (approved):

shivekkhurana (approved):

churik:
Tested basic flows, LGTM!

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)

PR #12756, opened 2021-10-27
e2e: waku align

churik:
align e2e with https://github.com/status-im/status-react/pull/12745


flexsurfer (approved):

churik:
@cammellos works awesome!
###  Question 1: if nickname was set up without adding user to contacts, should it be backed up?
The same question for `Mute` option.

### Issue 2: if using same multiaccount on both devices, after blocking contact and performing backup of contacts app is crashing on second device
**Steps:**
- `device a, b`: restore the same multiaccount on both devices
- `device a`: add contact
- `device a`: perform contact backup, so this contact will be shown on device b
- `device a`: block this contact
- `device a`: perform contact backup
On `device b` app is crashing.
[Status-debug-logs.zip](https://github.com/status-im/status-react/files/7476633/Status-debug-logs.zip)

PR #12796, opened 2021-11-05
nix: avoid nesting of Nix shells, show errors

jakubgs:
Nesting of Nix shells does not work, since we have no way to safely do that without starting an infinite loop of nested Nix shells.

Example:

image

I'd appreciate some local testing of these changes.


Jonathansoufer (approved):
Overall looks good to me.

PR #12812, opened 2021-11-11
[#11889] Allow sending messages offline (non-datasync messages)

### LINKED ISSUE #11889: Allow sending messages offline
```
As a user
I want to be able to send messages while I am offline
So that I don't have to wait for the app to be online again to send a message
```

### Description

Currently the input is disabled if the user is offline.
We should allow the user to send messages when offline and make sure that they are sent once it goes back online.

### Technical implemnentation

For 1-to-1 & private group chat messages one way to do this is to "pause" datasync when offline, and restart it when the user is online, that should just work.

For public chat/community messages or any other message type that does not use datasync, we should re-use the emoji resending code instead, and resend (or send) once they are online.

rasom:
fixes #11889 
status: wip

### DIFF
--- a/src/status_im/transport/message/core.cljs
+++ b/src/status_im/transport/message/core.cljs
@@ -255,7 +255,9 @@
 
 (fx/defn update-envelopes-status
   [{:keys [db] :as cofx} message-id status]
-  (apply fx/merge cofx (map #(update-envelope-status % status) message-id)))
+  (when (or (not= status :not-sent)
+            (= :online (:network db)))
+    (apply fx/merge cofx (map #(update-envelope-status % status) message-id))))
 
 (fx/defn set-message-envelope-hash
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

churik:
### ISSUE 1: "Not confirmed" is shown after several minutes when sending messages from offline

**Steps:**
1) restore multi-account from seed phrase
2) go offline
3) join the public chat and send some messages 
4) wait a couple of minutes

**Expected result:**
pending messages (will be resent after coming back online)
**Actual result:**
"Not confirmed" messages (will be resent after coming back online)
**OS:** Android
IMAGE_2021-12-01_152713

rasom:
it can be reproduced on iOs too

rasom:
@churik issue 1 should be fixed now

PR #12820, opened 2021-11-16
nix: make url2json.sh parse POM content for packaging

jakubgs:
Otherwise we can download a JAR when an AAR is necessary.

Might help with: https://github.com/status-im/status-react/pull/12769


bitgamma (approved):
looks great

PR #12825, opened 2021-11-17
Added "Search with DuckDuckGo"

du82:
### Summary

This adds the option to do a search on DuckDuckGo or open a tab in the browser directly from the search box on the chat page if no other results are found. This speeds up the searching process and integrates the browser deeper into the app.

### Review notes
This is a work in progress. Help with translations and icons would be much appreciated :)

#### Platforms
- Android
- iOS

### Steps to test
- Open Status
- Type a word that isn't in any of your chats (such as "netflix")
- Check to make sure that that your query can be searched on DuckDuckGo and the URL can be opened

status: wip

https://user-images.githubusercontent.com/65834069/142092208-39973b42-9ab4-4307-81ae-c4e56206b9a4.mp4

### DIFF
--- a/src/status_im/ui/screens/home/views.cljs
+++ b/src/status_im/ui/screens/home/views.cljs
@@ -107,6 +107,9 @@
                          :on-press (fn []
                                      (debounce/dispatch-and-chill [:contact.ui/contact-code-submitted false] 3000)
                                      (re-frame/dispatch [:search/home-filter-changed nil]))}])
+       
+       ;; Open public chat
+       
        (when valid-public?
          [quo/list-item {:theme    :accent
                          :icon     :main-icons/public-chat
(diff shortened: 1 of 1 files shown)


hesterbruikman:
Thanks so much for this PR @ducheng0! I'd love to see DuckDuckGo in Status. However from a UX pov, I do insist we follow the design proposed [here](https://github.com/status-im/status-react/issues/10575#issuecomment-625380173). 

The issue with using the Chat search bar is that it raises expectations we currently can't meet. Allowing to open a url from the chat search bar makes it appear as though this search bar is global, which it's not. As it is now, the Chat search bar already creates the false impression that you can search for other users (you can't, you search your list of existing chats), I wouldn't want to add to that.

PR #12843, opened 2021-11-24
[#12818] Support tx history (in a way) when on BSC chains

### LINKED ISSUE #12818: Network balance is not updated after sending transaction when network symbol is not ETH
# Bug Report

## Problem 
Reproduced on BSC mainnet, can be tested on BSC testnetwork after merging https://github.com/status-im/status-react/pull/12810

**Expected result:**
will be updated when tx gets confirmed

**Actual result:**
no update until relogin or custom scan tokens

### Reproduction

1) create new multiaccount
2) add BSC testnet (https://data-seed-prebsc-1-s1.binance.org:8545/ , 97 from [here](https://docs.binance.org/smart-chain/developer/rpc.html)
3) get Testnet Fund: https://testnet.binance.org/faucet-smart
4) try to send it to any address and pay attention to balance update

### Additional Information

- Status version: nightly 15/11/2021

rasom:
fix #12818 

There is no way to fetch archival data on BSC mainnet with RPC access we have now, thus only txs which are sent from our app will be shown. The links to block explorer are added for viewing more data.

Although we can fetch history on BSC testnet I made it work the same way as BCS mainnet just for the sake of testing.

status: ready

### DIFF
--- a/src/status_im/ethereum/transactions/core.cljs
+++ b/src/status_im/ethereum/transactions/core.cljs
@@ -2,6 +2,7 @@
   (:require [cljs.spec.alpha :as spec]
             [re-frame.core :as re-frame]
             [status-im.ens.core :as ens]
+            [status-im.ethereum.core :as ethereum]
             [status-im.ethereum.decode :as decode]
             [status-im.ethereum.eip55 :as eip55]
             [status-im.ethereum.encode :as encode]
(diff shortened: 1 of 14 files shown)


src/status_im/ethereum/transactions/core.cljs
```diff
@@ -13,16 +13,27 @@
 
 (def confirmations-count-threshold 12)
 
-(def etherscan-supported? #{:testnet :mainnet :rinkeby})
+(def etherscan-supported? #{1 3 4})
+(def binance-mainnet-chain-id 56)
```
flexsurfer: probably better to move to constants, it's defined  in two namespaces

flexsurfer (approved):

src/status_im/ethereum/transactions/core.cljs
rasom: fixed

churik:
5381 will fix in https://github.com/status-im/status-react/pull/12836
Awesome work, thank you!

PR #12854, opened 2021-11-29
Reorder community chats and categories using drag and drop

### LINKED ISSUE #12808: Add functionality for changing the order of channels in community categories
# Feature Issue

leftover from https://github.com/status-im/status-react/issues/12292

### User Story

As a community admin, I want to change the order of channels in community categories

**Related conversation**
https://github.com/status-im/status-react/pull/12758#issuecomment-955849770

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/12808

### Summary
PR implements a feature which allows users to reorder community chats and categories using drag and drop

<div align="center">
Screenshot
</div>

#### Testing Request:
This PR also introduces a new feature, deleting community chats. 
On clicking the red button next to chat, it will show a confirmation dialog and will allow a user to delete community chat. This feature replaces the old remove from category feature.

**Known Issue:**
Auto-scroll for long lists is a little laggy in ios, it will be fixed once react-native is upgraded to 0.64

status: ready


Parveshdhull:
Data is provided to DraggableFlatList using a reagent atom. Once an item is dragged and dropped on a new position then the `onDragEnd` function is called. In this function, I am making a call to status-go for reorder that item. 
After completion of the `onDragEnd` function, the item returns to its **old position**. And once updated data reset atom, the item jumps to the **newly updated position**. This makes this weird jump animation as shown in the image.

I also tried using local atom reset, without making a call to status-go, but this jumping behavior is still visible.

This animation is not visible in the react-native js [project](https://snack.expo.io/@computerjazz/rndfl3). That project used data as `useState`. I also tried using `react/useState`, but it gave a bad hook error.

@flexsurfer please, can you take a look into this issue.
It will be great if this jump can be removed, otherwise, this feature is complete and I can create a separate issue for this behavior. Thanks

blake-hunsicker:
Hey @Parveshdhull thanks for pinging me, this is looking good!

A few issues:

- The user should be able to drag entire communities without turning on the "Collapse Chats" toggle. It would be ideal to be able to drag the `Superheroes` community and its channels up above `Party time`. [See designs here](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=22167%3A100053)
- The user should also be able to drag a non-community-owned channel, like `Train info`, outside of the `None` list without having it join a community. For example: drag `Train info` and drop it above the `Party time` community, or between the `Party time` community and the `Superheroes` community.
- About the red delete icons -- there's a note to disregard them in the designs, but maybe we can add this feature as well since you've begun to work on it? What do you think @John-44?
- The non-community chats do not need a green + icon next to their names. Please let me know if I am mistaken here and this involves something else.

Parveshdhull:
Hi @blake-hunsicker,
Thank you very much for your feedback.
Sorry but, I think you misunderstood something, these all are categories in a single community, not multiple communities.

> The user should be able to drag entire communities without turning on the "Collapse Chats" toggle. It would be ideal to be able to drag the Superheroes community and its channels up above Party time

- Currently, each row is a single item in one draggable list. For allowing child items to move with a parent, I need to use nested bundled lists.
Then each category and item will act as a single item, and it will not be possible for chats to be dragged to other categories. 
So, only one function can be implemented. Either allow chats to be dropped inside other categories, or move bundled chats along with categories.
Performance-wise and also feature-wise, being able to drag chats to other categories seems a better trade-off.
Also, it will be super hard to track, drag of 10-20 bundled channels along category, on small mobile screen.

> The user should also be able to drag a non-community-owned channel, like Train info, outside of the None list without having it join a community. 

- I think you misunderstood here too, these all channels are owned/belong to a single community. And Yes, user can drag them outside of None list, into other categories.

> About the red delete icons -- there's a note to disregard them in the designs

- Designs had red buttons. Maybe the note was added later, Sorry I didn't notice it. I just focused on the design. Still, these are very useful if user wants to delete a full category at once.

> The non-community chats do not need a green + icon next to their names

- Either we should remove both, red and green buttons, or keep both. Because they follow the same principle, one removes chat from the category, another adds. And this color difference also makes, them visibly distinguishable between chats belonging to any category, and chats that don't. Also, chats will not be aligned, if only remove green buttons .

Thanks

src/status_im/subs.cljs
```diff
@@ -903,7 +919,9 @@
  :<- [:communities]
  (fn [communities [_ id]]
    (->> (get-in communities [id :categories])
-        (sort-by #(:position (get % 1))))))
+        (map #(assoc (get % 1) :community-id id))
+        (sort-by #(:position %))
```
flexsurfer: (sort-by :position) ?

flexsurfer (approved):

flexsurfer:
> @flexsurfer please, can you take a look into this issue. It will be great if this jump can be removed, otherwise, this feature is complete and I can create a separate issue for this behavior. Thanks

did you try `(reagent/flush)` ?

churik:
Can you please rebase and we re-launch e2e?

Parveshdhull:
> did you try `(reagent/flush)` ?

@flexsurfer, Issue Fixed 😍😍😍, Thank you very much
<div>
You are life saver
</div>

blake-hunsicker:
Hi @Parveshdhull 

> Sorry but, I think you misunderstood something, these all are categories in a single community, not multiple communities.

I did, sorry about that. I meant "categories" everywhere I wrote "communities" 😅

About dragging categories (with their child chats) and dragging chats by themselves:

I see your solution is to use the toggle to collapse categories so you can drag them. Is it possible to do as shown [in the design](https://www.figma.com/file/aS1ct66VQ6V0cio7vSqS8UoG/Chat?node-id=22167%3A100053), where when you hold a category's drag icon, the category's child chats collapse under it, so you move one list item?

>Also, it will be super hard to track, drag of 10-20 bundled channels along category, on small mobile screen.

Yes, without collapsing this is not ideal. But if it's collapsed-on-drag like described above then we are fine.

> I think you misunderstood here too, these all channels are owned/belong to a single community. And Yes, user can drag them outside of None list, into other categories.

Sorry again for the wrong wording 😅

> Designs had red buttons. Maybe the note was added later, Sorry I didn't notice it. I just focused on the design. Still, these are very useful if user wants to delete a full category at once.

The note was always there, but they were left in because I think this was a rush to complete. Sorry, I should have been clearer. Can you remove the red and green buttons for now?

> Either we should remove both, red and green buttons, or keep both. Because they follow the same principle, one removes chat from the category, another adds. And this color difference also makes, them visibly distinguishable between chats belonging to any category, and chats that don't. Also, chats will not be aligned, if only remove green buttons .
Thanks

This is not the purpose of these buttons. I believe their purpose is to delete the chats or categories, so in any case the green is not applicable.

churik:
Works great! (didn't check on IOS)
The only small thing to improve: delete category `None` when rearrange categories enabled (we can't do anything with it)

https://user-images.githubusercontent.com/4557972/147354439-dda84db0-02aa-43b5-9b7e-949cc74fc5bf.mp4
But it is definitely not a blocker for feature.

PR #12873, opened 2021-12-02
[#12782] Re-open app on the last open chat

### LINKED ISSUE #12782: Re-open app on the last open chat
As a user
I want the app to re-open on the same chat I was before it was closed
So that I can pick up from where I left

### Description

When the app is closed (not on logout), if the user is on a chat, it should re-open the app on the same chat.

### Technical implementation

This is a mobile specific feature, we can consider storing this information on async storage instead of status-go.

rasom:
fixes #12782 

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -1,65 +1,64 @@
 (ns status-im.events
-  (:require [re-frame.core :as re-frame]
-            [status-im.chat.models :as chat]
-            [status-im.i18n.i18n :as i18n]
-            [status-im.mailserver.core :as mailserver]
-            [status-im.multiaccounts.core :as multiaccounts]
-            [status-im.ui.components.react :as react]
-            [status-im.utils.fx :as fx]
-            status-im.utils.logging.core
-            status-im.backup.core
-            [status-im.wallet.core :as wallet]
-            [status-im.keycard.core :as keycard]
-            [status-im.utils.dimensions :as dimensions]
-            [status-im.multiaccounts.biometric.core :as biometric]
-            [status-im.constants :as constants]
-            [status-im.native-module.core :as status]
-            [status-im.ui.components.permissions :as permissions]
-            [status-im.utils.utils :as utils]
-            [status-im.ethereum.json-rpc :as json-rpc]
-            [status-im.anon-metrics.core :as anon-metrics]
-            [status-im.utils.universal-links.core :as universal-links]
-            clojure.set
-            status-im.currency.core
-            status-im.navigation
-            status-im.utils.universal-links.core
-            status-im.wallet.custom-tokens.core
-            status-im.waku.core
-            status-im.wallet.choose-recipient.core
-            status-im.wallet.accounts.core
-            status-im.popover.core
-            status-im.visibility-status-popover.core
-            status-im.visibility-status-updates.core
-            status-im.bottom-sheet.core
+  (:require clojure.set
+            [re-frame.core :as re-frame]
             status-im.add-new.core
-            status-im.search.core
-            status-im.http.core
-            status-im.profile.core
+            [status-im.anon-metrics.core :as anon-metrics]
+            [status-im.async-storage.core :as async-storage]
+            status-im.backup.core
+            status-im.bootnodes.core
+            status-im.bottom-sheet.core
+            status-im.browser.core
+            status-im.browser.permissions
+            [status-im.chat.models :as chat]
             status-im.chat.models.images
-            status-im.ui.screens.privacy-and-security-settings.events
+            status-im.chat.models.input
+            status-im.chat.models.loading
+            status-im.chat.models.transport
+            [status-im.constants :as constants]
+            status-im.contact.block
+            status-im.contact.chat
+            status-im.contact.core
+            status-im.currency.core
+            [status-im.ethereum.json-rpc :as json-rpc]
+            status-im.ethereum.subscriptions
+            status-im.fleet.core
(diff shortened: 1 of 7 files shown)


rasom:
ok, it works on iOS simulator but doesn't work on real device

src/status_im/events.cljs
```diff
@@ -1,65 +1,64 @@
 (ns status-im.events
-  (:require [re-frame.core :as re-frame]
-            [status-im.chat.models :as chat]
-            [status-im.i18n.i18n :as i18n]
-            [status-im.mailserver.core :as mailserver]
-            [status-im.multiaccounts.core :as multiaccounts]
-            [status-im.ui.components.react :as react]
-            [status-im.utils.fx :as fx]
-            status-im.utils.logging.core
+  (:require clojure.set
```
flexsurfer, line 2: why shuffle namespaces ?

src/status_im/events.cljs
```diff
@@ -193,8 +192,19 @@
 (fx/defn on-will-focus
   {:events [:screens/on-will-focus]
    :interceptors [anon-metrics/interceptor]}
-  [cofx view-id]
+  [{:keys [db] :as cofx} view-id]
   (fx/merge cofx
+            (cond
+              (= :chat view-id)
+              {::async-storage/set! {:chat-id (get-in cofx [:db :current-chat-id])}
+               :db (assoc db :screens/was-focused-once? true)}
```
flexsurfer, line 201: could you elaborate on was-focused-once? ?

src/status_im/multiaccounts/login/core.cljs
```diff
@@ -363,6 +363,13 @@
         keychain/auth-method-biometric
         keychain/auth-method-password))))
 
+(defn check-chat-id []
+  (async-storage/get-item
```
flexsurfer: is async-storage shared between multiacccounts ? what will happen if you open chat in one multiaccount and sign in in another ?

src/status_im/utils/fx.cljs
```diff
@@ -51,11 +51,14 @@
...
+        fx-fns (if initial-fxs? rest-args args)
+        res
+        (clojure.core/reduce (fn [fxs fx-fn]
+                               (let [updated-cofx (update-db cofx fxs)]
+                                 (if fx-fn
+                                   (safe-merge fxs (fx-fn updated-cofx))
+                                   fxs)))
+                             (if initial-fxs? first-arg {:db db})
+                             fx-fns)]
+    (swap! handlers/handler-nesting-level dec)
```
flexsurfer, line 63: could you please elaborate on this change ?

src/status_im/events.cljs
rasom: I just sorted them. The more we have the harder it is to quickly check whether necessary ns is required and we already had one duplication there. https://github.com/status-im/status-react/blob/develop/src/status_im/events.cljs#L24

src/status_im/events.cljs
rasom: Yes, sure. On non development build `:screens/on-will-focus` is dispatched before we have enough time to handle `:chat-id` from async storage. Thus if we do not ignore the first `:screens/on-will-focus` `:chat-id` is set to `nil` before it can be used.

src/status_im/multiaccounts/login/core.cljs
rasom: Basically it works only if you did not logged out of your multi account, otherwise "login only" handlers won't be called. When you move to any other screen than `:chat` `:chat-id` is set to `nil`. So basically nothing new will happen "if you open chat in one multiaccount and sign in in another", to answer your question.

src/status_im/utils/fx.cljs
rasom: This is completely unrelated to PR objective (if you ignore the fact that it helped me with debugging a bit), it just fixes a tiny issue with tracing fx handlers, `nasting-level` was not properly decremented. I can file a separate PR but I'm not sure it is really necessary.

flexsurfer (approved):

churik:
@rasom 
### ISSUE 1: when app is closed on group chat view, user is redirected to "Join group chat" screen after reopening app

**Steps:**
1) User A: create a group chat
2) User A: on group chat view, close app
3) User A: reopen app (force close / relogin)

**Expected result:**
same chat view 

**Actual result:**
IMAGE 2021-12-06 14:44:00
**OS:** IOS, Android

### Question 2: in case of reopening app on community channel > tap back user is redirected to main home view instead of the community page. Expected?

 It differs from tapping back when a user is on channel view and tap back without relogin, better if it will be the same in both cases to prevent different navigation for the same scenarios.

rasom:
ISSUE 1: figuring out how to fix this

Question 2:  its hard to say, if we don't plan to have community chats in the list on Home Screen then we probably should add community screen to the views stack, but as it is now is probably fine. I will check how hard is to add another screen there

churik:
@rasom 
on the last build, https://github.com/status-im/status-react/commit/7ffbe4a2774af5d82ee0527492c9aa71b6ec36a5 feature doesn't work anymore so if you reopen the app on chat view (tried with public and 1-1 chat), home view is shown

churik:
Awesome work!
Checked with all types of chats, combinations of several multiaccounts, PNs, deep links (they have a priority)
Some e2e should be fixed, will be done separately.

**Devices:**
- IPhone 12 Mini (IOS 14)
- IPhone 7 (IOS 13)

PR #12901, opened 2021-12-10
[#12898] Fix endless loading in pubchats

### LINKED ISSUE #12898: Public chats can stuck in loading state forever
# Bug Report

## Problem 
**IOS:** several chats are stuck with an endless loading spinner (fetching history for another public chats is working) - could reproduce on IOS only.
**Android:** if start fetching history in offline mode and then switch to inline, "Fetch more messages" is available only after re-login

#### Expected behavior
no loading; fetch more messages is available after come back online

#### Actual behavior

**IOS:** 
https://user-images.githubusercontent.com/4557972/145432479-2518d99c-f0aa-4b33-ab40-7694c99b3b0c.mp4
chats are stuck with a loading spinner, re-login doesn't help 

**Android:**
IMAGE 2021-12-09 17:57:12
"Fetch more history" is not available until another relogin

### Reproduction
(Easiest way I found to reproduce behaviour) 

1) Open Status
2) Turn off wi-fi
3) Join the chat with history (i.e. #status)  and chat with no history (any random)
4) Re-login
5) Turn on Wi-Fi
6) Pay attention to both chats 

NOTE: sometimes additional re-login can fix the issue, but in some cases chat with no history stuck in loading forever

### Additional Information

- Status version: nightly 09/12/21

**IOS:** [logs (1).zip](https://github.com/status-im/status-react/files/7687040/logs.1.zip)

rasom:
fix (partially) #12898

In case if a new pubchat was created while offline, `syncedTo`/`syncedFrom` props of the chat wouldn't be updated until after the first successful request to mailserver which means that values would be updated only on the next re-login even if the device is back online. 

status: ready

### DIFF
--- a/src/status_im/mailserver/core.cljs
+++ b/src/status_im/mailserver/core.cljs
@@ -184,8 +184,9 @@
 
 (fx/defn handle-successful-request
   {:events [::request-success]}
-  [{:keys [db]} response-js]
-  {:db (dissoc db :mailserver/current-request)})
+  [{:keys [db] :as cofx} response-js]
+  {:db       (dissoc db :mailserver/current-request)
+   :dispatch [:sanitize-messages-and-process-response response-js]})
 
 (fx/defn process-next-messages-request
   {:events [::request-messages]}


src/status_im/mailserver/core.cljs
```diff
@@ -182,10 +182,33 @@
     (when (= :connected mailserver-state)
       mailserver)))
 
+(fx/defn update-nil-synced-to
```
cammellos: I think in theory you can just dispatch:
```
(re-frame/dispatch [:sanitize-messages-and-process-response response])
```

below, so you don't have to write custom logic to handle it
(it would be on line 210, so `:dispatch [:sanitize.....]`)

rasom: sure, thanks

rasom: done

cammellos (approved):

churik:
Tested on IOS (Iphone 12 mini, IOS 15.1.1):
after coming back to online public chat history still is not fetched:

https://user-images.githubusercontent.com/4557972/145590747-da14454b-576c-4caf-90a1-279295919426.mp4

History is fetched after re-login

rasom:
Ok I see. It definitely fixes the issue which I can reproduce on local iOS build (so that after adding a pubchut while offline app crashes and then I see endless loading after re-login), but it doesn't fix the case when app "survives" adding a pub chat. I will continue debugging.

rasom:
@churik it seems that you just need to wait a bit


https://user-images.githubusercontent.com/2364994/145784792-0c76e3c0-9887-46db-ae0d-8410580cbd85.mov

shivekkhurana (approved):

PR #12924, opened 2021-12-16
avoid closing/opening NFC dialog in rapid sequence

### LINKED ISSUE #12910: Can't import a multiaccount from mnemonic to keycard
with 1.18.0 (20211213183517), I can't import a mnemonic into a keycard. This has been tested only when the keycard was already used and is thus being factory reset in the process. This has not been tested when the card is brand new. 

steps: 

- access  existing keys 
- recover from seed phrase 
- enter the seed phrase manually
- chose keycard as a key storage
- chose 'return card to factory settings'
- type two times the new 6 digit passcode 
- last step is a success, bu the app gets stuck on this screen

IMG_2B4C7457CC1C-1

bitgamma:
Fixes #12910.

on iOS15 opening/reopening the NFC dialog too fast can cause an error. This error is actually handled correctly and opening is retried, which is why the bug is hard to reproduce (I still didn't manage to reproduce it) but maybe on some specific model/os combination unexpected conditions might happen.

This fix avoids closing/reopening the NFC dialog in rapid sequence, which happened when recovering from mnemonic between the pairing step and key loading step. This has the added advantage of saving about 1-2 seconds (on iOS, on Android it makes no difference).

### DIFF
--- a/src/status_im/keycard/core.cljs
+++ b/src/status_im/keycard/core.cljs
@@ -525,14 +525,14 @@
     (fx/merge cofx
               {:db (assoc-in db [:keycard :card-state] card-state)}
               (set-setup-step card-state)
-              (common/hide-connection-sheet)
 
               (when paired?
                 (load-pairing))
 
-              (when (and flow
-                         (= card-state :init))
-                (proceed-setup-with-initialized-card flow instance-uid paired?))
+              (if (and flow
+                       (= card-state :init))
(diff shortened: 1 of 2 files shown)


bitgamma:
I got confirmation from Guy-Louis that it solves his problem

rasom (approved):

qoqobolo:
@bitgamma thanks for the fix!
PR can be merged.

PR #12933, opened 2021-12-19
Increase community description character limit

du82:
This increases the community character limit to 600 characters, allowing community owners to add rules, an explanation, or other information that require long-form text. 140 characters just isn't enough to say anything useful.

### DIFF
--- a/src/status_im/ui/screens/communities/create.cljs
+++ b/src/status_im/ui/screens/communities/create.cljs
@@ -13,7 +13,7 @@
             [status-im.utils.debounce :as debounce]))
 
 (def max-name-length 30)
-(def max-description-length 140)
+(def max-description-length 600)
 
 (defn valid? [community-name community-description]
   (and (not (str/blank? community-name))


flexsurfer:
@hesterbruikman wdyt?

hesterbruikman:
Thanks for the tag @flexsurfer, @ducheng0 let me check where all the community description will be displayed. I suspect the character limit might be to prevent cut off descriptions elsewhere. 

Noted that for sure there needs to be a way to add rules, based on how I see Discord used. Another generic way to do this, after the community is created is to write a pinned message. This can be done before setting up permissions and inviting anyone. Not a super obvious flow when creating the community though.

hesterbruikman:
> Thanks for the tag @flexsurfer, @ducheng0 let me check where all the community description will be displayed. I suspect the character limit might be to prevent cut off descriptions elsewhere.
> 
> Noted that for sure there needs to be a way to add rules, based on how I see Discord used. Another generic way to do this, after the community is created is to write a pinned message. This can be done before setting up permissions and inviting anyone. Not a super obvious flow when creating the community though.

@ducheng0 I checked and we actually have more streamlined solution coming up, where community creations will be offered a dedicated long form input to describe community guidelines. Here's what it will look like on Desktop.

Because of concerns about displaying longer community descriptions in other places, I would ask not to merge this PR

Community
Compact mode

For future reference, if you have suggestions for improvement, feel free to create an issue and ping, we might be able to supply you with designs

du82:
What if I don't want rules for my community? Could this prompt be skipped?

Also, why 1400 characters? Why not 5K or 10K?

I'm not a fan of Captcha or anything proprietary in the client. Adding Captcha would be the point where I fork the project just so I can remove the Captcha and proprietary stuff.

Issue #12919, opened 2021-12-15
Emoji as last char erases value in input field if char + emoji > acceptable limit


churik:
# Bug Report

## Problem 
1 emoji = 2 chars, so in case when text + emoji > limit for field (i.e. 4096 for chat input, 600 for status), value is erased
Reproducible on different text inputs on IOS (message input, group chat invite field, status input etc)

#### Expected behavior
can't type emoji, as it 2 chars

#### Actual behavior
when typing emoji, value is erased

https://user-images.githubusercontent.com/4557972/146220608-32869769-7fa8-4cf0-8b6a-97ef8d4bf12b.mp4

### Reproduction

1) Open Status
2) Open status tab > +
3) Paste text with 599 chars
4) Type emoji

### Additional Information

- Status version: 1.18 
- Operating System:  iOS

[comment]: # (Please, add logs/notes if necessary)

rasom:
https://github.com/facebook/react-native/issues/28774

There is no way to properly fix this bug at the moment. We can implement a kind of fix mentioned in the issue above, but it will rather cause flickering UI.  

Another option is to allow user to type more than 600 chars and then disable `send` button and somehow emphasise that limit has been exceeded. That might cause worse UX, but at least it will be more predictable. :) Also this is only needed for iOS version.

rasom:
@churik it is fixed for status input 

Tested it on message input and got a bit different behaviour. Emoji is cute and `send` button is disabled, which is probably enough?

<img width=320 src=https://user-images.githubusercontent.com/2364994/149905573-2f375856-a247-43c8-8a15-6dd5a5783093.jpeg />

churik:
Yes, it is enough, thanks @rasom

PR #12949, opened 2021-12-22
[#12919] Allow adding emoji in status input even if it exceeds limit

rasom:
fixes (partially) #12919

Screenshot 2021-12-22 at 17 10 09

status: ready

### DIFF
--- a/src/status_im/ui/screens/status/new/views.cljs
+++ b/src/status_im/ui/screens/status/new/views.cljs
@@ -12,7 +12,8 @@
             [status-im.ui.components.icons.icons :as icons]
             [quo.components.animated.pressable :as pressable]
             [status-im.ui.screens.status.views :as status.views]
-            [status-im.ui.screens.status.new.styles :as styles]))
+            [status-im.ui.screens.status.new.styles :as styles]
+            [status-im.utils.platform :as platform]))
 
 (defn buttons []
   [react/view styles/buttons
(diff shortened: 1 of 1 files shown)


src/status_im/ui/screens/status/new/views.cljs
```diff
@@ -53,7 +54,8 @@
         input-text (re-frame/subscribe [:chats/timeline-chat-input-text])
         sending-image (re-frame/subscribe [:chats/timeline-sending-image])]
     (fn []
-      (let [{:keys [uri]} (first (vals @sending-image))]
+      (let [{:keys [uri]} (first (vals @sending-image))
+            text-lenght (count @input-text)]
```
Parveshdhull, line 58: lenght - typo ?

bitgamma: text-length and message-max-length are misspelled

bitgamma (approved):

Parveshdhull (approved):

src/status_im/ui/screens/status/new/views.cljs
rasom: oh jeez... I will fix it, thanks

rasom: fixed

pavloburykh:
## What has been tested: 

1. For IOS now it is possible to add emoji even if exceeds a limit and it does not result in deletion of value in input. At the same time user is unable to post message until he enters the allowed amount of 600 symbols.

2. For Android nothing changed and behaviour is correct: it is still impossible to exceed input limit and enter emoji if there is not enough symbols left available for entering. This bug  https://github.com/status-im/status-react/issues/12919  is reproducible only for IOS

PR #12951, opened 2021-12-23
e2e: updated SauceSharedMultipleDeviceTestCase with separate step for each test

### LINKED ISSUE #12931: Create a mechanizm to distinguish e2e start/end within one session
## Problem 
Currently there are `test_public_chat_messaging_emojis_timestamps` + `test_unread_messages_counter_public_chat` that share one session
We need a way to distinguish the start of new e2e within one session.
IMAGE 2021-12-17 17:35:26
As it is possible in saucelab to point link to appium command (i.e. https://app.saucelabs.com/tests/3a43a4d4035a4a4aa899ccd828e890aa?auth=2eaf5de999054c712e86c960d127e665#16 ), this would be nice to do some appium command.
Links in GH / testrail for such e2e also should be updated accordingly.

yevh-berdnyk:
fixes #12931

### DIFF
--- a/test/appium/support/base_test_report.py
+++ b/test/appium/support/base_test_report.py
@@ -28,8 +28,7 @@ class BaseTestReport:
         file_name = "%s.json" % test_name
         return os.path.join(self.TEST_REPORT_DIR, file_name)
 
-    def save_test(self, test, geth: dict):
-        file_path = self.get_test_report_file_path(test.name)
+    def save_geth(self, geth: dict):
         geth_paths = {}
         for log in geth.keys():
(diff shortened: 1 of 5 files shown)


churik:
@yevh-berdnyk thank you, awesome work!
Found issue in case when 1 test case from group is failed.
1) Can't attach grth log to failed testcase
```
 File "/Users/curikovatm/Status/status-react/test/appium/support/testrail_report.py", line 147, in add_results
    for geth in test.geth_paths.keys():
AttributeError: 'list' object has no attribute 'keys'
```
As a consequence (I suppose) other passed e2e are not marked as passed.

churik (approved):
🎉

PR #12952, opened 2021-12-23
Issue template: add instruction for release-note

churik:
Update release-umbrella template with instruction to prepare release-notes


jakubgs (approved):
The `?view` in URL is unnecessary, since that's the default for `notes.status.im`.

PR #12979, opened 2022-01-06
Backup communities

### LINKED ISSUE #12974: Backup joined communities to waku
As a user
I want to backup the communities I joined to waku
So that when re-installing my device I am able to re-join those communities

### Description

Similarly as we do for contacts, we should backup the user's communities

cammellos:
[Fixes: #12974]

status: ready

### DIFF
--- a/src/status_im/ui/screens/backup_settings/view.cljs
+++ b/src/status_im/ui/screens/backup_settings/view.cljs
@@ -47,6 +47,6 @@
         :title                  (i18n/label :t/last-backup-performed)
         :accessory              :text
         :subtitle                   (if (pos? last-backup)
-                                      (datetime/time-ago-long (datetime/to-date last-backup))
+                                      (datetime/time-ago-long (datetime/to-date (* 1000 last-backup)))
                                       (i18n/label :t/never))}]]
      [footer performing-backup?]]))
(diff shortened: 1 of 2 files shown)


briansztamfater (approved):

qoqobolo:
@cammellos thanks for your work!

### ISSUE 1: 'Last backup performed 19054 D ago' is displayed after each relogin

**Steps:**
1. Create an account
2. Tap Perform `backup`
3. Relogin and check backup settings

Screenshot 2022-03-22 at 11 23 26
 
### ISSUE 2/Q: Communities you left are not backed up
Not sure if community deletion sync is expected in this case

**Steps:**
1. Recover the same account on two devices
2. Create a community on device1 > Perform backup
3. Leave the community on device1 > Perform backup
4. Check communities on device2

**Expected result:** no community on device2
**Actual result:** community is still on the list

cammellos:
@qoqobolo Thanks for testing
Issue 1 should be fixed in the next build, issue 2 I will be looking into, would it be ok if we split this in a separate issue, so we can cut the release branch with this included, and fix issue 2 separately?

qoqobolo:
> would it be ok if we split this in a separate issue, so we can cut the release branch with this included, and fix issue 2 separately?

Absolutely. There may be another issue with pending community requests, but I'll also investigate this later and log separately.
Thank you @cammellos, PR can be merged.

cammellos:
@qoqobolo thank you!

PR #12980, opened 2022-01-06
ci: upgrade Nix from 2.3.12 to 2.6.1

### LINKED ISSUE #12832: Nix 2.4 issues with git tags
It appears that newest `2.4` release of Nix has some weird issues with Git tags:
```
 > make run-clojure 
yarn shadow-cljs watch mobile
Configuring Nix shell for target 'clojure'...
fatal: couldn't find remote ref refs/heads/v2.0.3-status-v6
error: program 'git' failed with exit code 128
(use '--show-trace' to show detailed location information)
fatal: couldn't find remote ref refs/heads/v2.0.3-status-v6
error: program 'git' failed with exit code 128

       … while fetching the input 'git+https://github.com/tbenr/react-native-audio-toolkit.git?ref=v2.0.3-status-v6&rev=7ae9055cf6169b30f5089bda7bfcfc1c40a715e5'

       … while evaluating the attribute 'buildCommand' of the derivation 'react-native-audio-toolkit.git'
```

jakubgs:
Due to changes in how Nix handles Git refs we need to specify
`refs/tags/` prefix in `package.json` to avoid the following error:
```
fatal: couldn't find remote ref refs/heads/v2.0.3-status-v6
error: program 'git' failed with exit code 128
```

I also had to rewrite some logic in `nix/scripts/source.sh` in order to take account of single-user and multi-user installations. We default to multi-user for Darwin, but not for any other OS due to discovered issues with `nix-daemon` socket on Arch and open file limits.

I also rewrote `nix/scripts/setup.sh` and `/nix/scripts/purge.sh` to support different types of installations. Both single-user and multi-user, as some operating systems have issues with multi-user installations.

Resolves: https://github.com/status-im/status-react/issues/12832
Depends on: https://github.com/status-im/status-jenkins-lib/pull/37
Issue: https://github.com/NixOS/nix/issues/5291

Related changes:
* https://github.com/status-im/infra-ci/commit/84947b9f
* https://github.com/status-im/infra-ci/commit/bb98f5f3
* https://github.com/status-im/infra-ci/commit/f75d524d
* https://github.com/status-im/infra-ci/commit/d1fc92cd
* https://github.com/status-im/infra-ci/commit/87c4091e
* https://github.com/status-im/infra-ci/commit/8d6b6b3f
* https://github.com/status-im/infra-ci/commit/c4f13285
* https://github.com/status-im/infra-ci/commit/38ac698d


jakubgs:
Looks like this will require some changes to [`packages-check-bot`](https://github.com/status-im/packages-check-bot) in order let the bot find the tags:

image

Parveshdhull (approved):

jakubgs:
I'd appreciate a manual test, especially on a MacOS machine. Will require `make nix-purge` first tho.

jakubgs:
Oooh, this might be a problem:
```
jenkins@maci7-03.ms-eu-dublin.slave.ci:~ % /tmp/nix-2.5.1-x86_64-darwin/install --no-daemon
Error: --no-daemon installs are no-longer supported on Darwin/macOS!
```
Since the single user(`--no-daemon`) way is the default for our CI hosts.

jakubgs:
I tried installing in a single user mode anyway by commenting out the check but it seems to fail:
```
jenkins@maci7-03.ms-eu-dublin.slave.ci:~ % /tmp/nix-2.5.1-x86_64-darwin/install --no-daemon                                          
performing a single-user installation of Nix...
copying Nix to /nix/store..............................................
installing 'nix-2.5.1'
building '/nix/store/4lnxyn7j0kpc4ir2qx39vd5jch28k494-user-environment.drv'...
installing 'nss-cacert-3.66'
building '/nix/store/6wndkfwgrdlc6igsnvwr22c07lbqvv2w-user-environment.drv'...
unpacking channels...
error: failed to extract archive (Cannot extract through symlink /nix/store/q18dv2s6jw2xyrmyy443msjis386ibgv-nixpkgs-22.05pre345560.b458e5133fb/nixpkgs-22.05pre345560.b458e5133fb)
error: builder for '/nix/store/cbrq0nrwrmihl5l3x1m34hwfwks411jf-nixpkgs-22.05pre345560.b458e5133fb.drv' failed with exit code 1
error: program '/nix/store/iq3ra93h9kxmnrw3zlxmqn8ng5w62dra-nix-2.5.1/bin/nix-env' failed with exit code 100
Fetching the nixpkgs channel failed. (Are you offline?)
To try again later, run "nix-channel --update nixpkgs".
modifying /Users/jenkins/.zshrc...

Installation finished!  To ensure that the necessary environment
variables are set, either log in again, or type

  . /Users/jenkins/.nix-profile/etc/profile.d/nix.sh

in your shell.
```

jakubgs:
These instructions are useful: https://gist.github.com/chriselsner/3ebe962a4c4bd1f14d39897fc5619732

jakubgs:
Another thing is that the current installer appears to not be made for non-interactive installation:
```
jenkins@maci7-03.ms-eu-dublin.slave.ci:~ % /tmp/nix-2.5.1-x86_64-darwin/install --daemon                                           
Switching to the Multi-user Installer                                             
Welcome to the Multi-User Nix Installation

This installation tool will set up your computer with the Nix package
manager. This will happen in a few stages:

1. Make sure your computer doesn't already have Nix. If it does, I
   will show you instructions on how to clean up your old install.

...

6. Start the Nix daemon.

Would you like to see a more detailed list of what I will do?
[y/n] 
```

jakubgs:
For some reason the installator is not creating the `/etc/profile.d/nix.sh` file:
```
admin@maci7-03.ms-eu-dublin.slave.ci:~ % cat /etc/profile.d/nix.sh
cat: /etc/profile.d/nix.sh: No such file or directory
```

jakubgs:
The way Nix is sources id from modified `/etc/bashrc` and `/etc/zshrc`:
```
jenkins@maci7-03.ms-eu-dublin.slave.ci:~ % tail -n6 /etc/bashrc 
# Nix
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
  . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
# End Nix
```

jakubgs:
It's worth noting that what needs to be sourced is `/nix/var/nix/profiles/default/etc/profile.d/nix.sh` and not `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`.

jakubgs:
I had to do a bunch of changes to `/nix/scripts/source.sh` in order to take into account [single-user](https://nixos.org/manual/nix/stable/installation/installing-binary.html#single-user-installation) and [multi-user](https://nixos.org/manual/nix/stable/installation/installing-binary.html#multi-user-installation) installations.

jakubgs:
Some points @briansztamfater noted after local tests:

>Here's my feedback for the nix upgrade on M1:
>
>1. Got the error `../status-react/nix/scripts/setup.sh: line 48: IN_NIX_SHELL: unbound variable`
>     - Removed that check from `setup.sh` to try the installation
>2. Then the following error `sha256sum: command not found`
>     - Solved installing `brew install coreutils`
>3. Throwed Error: --no-daemon installs are no-longer supported on Darwin/macOS!
>     - Changing the script to install Nix with `--daemon` worked
>4. `make nix-update-pods` threw an error. Attaching it in the following mesage cause it is large
>    - `Invalid configuration 'arm64-apple-darwin20.3.0': machine 'arm64-apple' not recognized`
>
>But running `cd ios && pod install` manually and then `make run-ios` worked fine.

I've addressed points 1-3, but issue number 4 is still waiting for a GNU Coreutils fix to be included in a release:
https://git.savannah.gnu.org/cgit/coreutils.git/commit/?id=e3d1ff04370b3e090e4cfeb8f6cc4a63590a516f

jakubgs:
Since we are forced to use multi-user installation on MacOS to solve the read-only root problem, this brings the question:

__Should we also use the multi-user installation on Linux, to make things more consistent?__

 In theory there should be no different in terms of builds themselves, the main difference in how the Nix profiles are source  and that builds will run under special `nixbld` users instead of your user. I'd appreciate some help from devs in testing this before I commit to making this the default though. I will need to test on both Linux and MacOS with `x86_64` architectures. Brian already tested this for me on `arm64` on MacOS.

Docs: https://nixos.org/manual/nix/stable/installation/installing-binary.html#installing-a-binary-distribution

bitgamma:
Tested on Windows 10 + WSL2 (Ubuntu). In the end everything works, but since by default systemd is not enabled (it is possible to enable but is not supported), the nix-daemon does not start automatically. Forgetting to start it does not make nix explicitly complain about that, rather it fails to do stuff requiring elevated privileges.

Fixed by starting the daemon as
`sudo /nix/var/nix/profiles/default/bin/nix-daemon`

Probably I should have started it as nixbld user but it works anyways. nix-daemon, regardless of its name is not a daemon and remains in foreground. Needs to be either started via systemd or via daemonize (not a default package). For now I simply kept it foreground and opened a separate shell.

WSL2 can be detected via `uname -r`, which in my case returns `4.19.104-microsoft-standard`. Grepping for microsoft should be pretty safe to detect it

nix/scripts/setup.sh
```diff
@@ -1,51 +1,68 @@
...
+    export NIX_INSTALLER_NO_MODIFY_PROFILE=1
+
+    # Download installer and verify SHA256
+    curl -s "${NIX_INSTALL_URL}" -o "${NIX_INSTALL_PATH}"
+    echo "${NIX_INSTALL_SHA256}  ${NIX_INSTALL_PATH}" | sha256sum -c
+    chmod +x "${NIX_INSTALL_PATH}"
+
+    # Identify installation type
+    if [[ -z "${NIX_INSTALL_OPTS}" ]]; then
+        if [[ "${UNAME}" == "Linux" ]]; then
```
bitgamma: when running the script with no parameters directly as `nix/scripts/setup.sh`, the installation options are listed. `UNAME` seems to be undefined (I don't see it defined anywhere in this script)

jakubgs: Ah, you're right, I must have lost `UNAME=$(uname -s)` during some cleanup.

jakubgs:
Looks like there's an issue with Nix multi-user installation on Linux:
```
error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Operation not permitted
```
It appears this is related to the warning the installer gives:
```
Nix won't work in active shell sessions until you restart them.
```
Based on this comment it's related to Nix daemon not starting: https://github.com/NixOS/nix/issues/3435#issuecomment-731834542

jakubgs:
Looks like the issue @Parveshdhull is seeing:
```
error: could not set permissions on '/nix/var/nix/profiles/per-user' to 755: Operation not permitted
```
Is due to his `nix-daemon` service not starting up because of:
```
Nix Daemon was skipped because of a failed condition check (ConditionPathIsReadWrite=/nix/var/nix/daemon-socket).
```
Which appears to be an issue with not fully cleaning up a previous multi-user installation:
https://www.extrema.is/blog/2022/01/05/uninstalling-multi-user-nix

jakubgs:
Since these issues appears to be present on both Arch and Arch-based systems like Manjaro I've added a special case:
https://github.com/status-im/status-react/blob/63e0d9f95f089e5af64f6facc7fa88344b80549c/nix/scripts/setup.sh#L28-L30
Thanks to @Parveshdhull for testing this.

jakubgs:
I'm seeing new type of issue with Android builds:
```
error: opening directory '/nix/store/5p6jmvh8q4bqgq773nbs0daa2ygx1y45-okhttp-3.9.1.pom': Too many open files
script returned exit code 1
```
I'm trying to raise the system-wide limit, or the user limits, but it's not helping. And based on graphs the spikes are not that big:

image

jakubgs:
And jenkins limits are clearly raised:
```
[Pipeline] sh
14:43:53  + ulimit -Ha | grep nofiles
14:43:53  nofiles              1048576
[Pipeline] sh
14:43:54  + ulimit -Sa | grep nofiles
14:43:54  nofiles              1048576
```
So that's weird. And I also raised the `nixbld` group limits:
```
admin@linux-03.he-eu-hel1.slave.ci:~ % grep nixbld /etc/security/limits.d/ansible.conf
@nixbld	soft	nofile	4096
```

jakubgs:
It appears that adding these lines to `nix-daemon.service` definition does help:
```
LimitNOFILE=1048576
LimitNOFILESoft=65536
```
But that's not a solution I'd like to use, because it would require a user intervention at install time, or extra setup steps.

According to this Systemd services ignore system-wide limits set in `/etc/security/limits.conf`:
https://fredrikaverpil.github.io/2016/04/27/systemd-and-resource-limits/

jakubgs:
I've opened and issue describing the problem I'm seeing: https://github.com/NixOS/nix/issues/6007

jakubgs:
I tried setting `max-jobs` to values lower than 1024 - like 1000, 800, and 600 - but we are still hitting the limit.

Most probably because of: https://github.com/NixOS/nix/issues/6007#issuecomment-1024198760

jakubgs:
Since I can fix the file limit settings by applying the `LimitNOFILE` change in `nix-daemon` service definition on CI hosts I've decided to go ahead with this PR while making multi-user installation default only for Darwin OS.

Infra fix: https://github.com/status-im/infra-ci/commit/d1fc92cd

jakubgs:
But I'm also seeing errors regarding passing secrets via `extra-sandbox-paths`:
```
source root is status-react-source-jsbundle
secretsPhase
/nix/store/adbsl3xmgk2nzmyjf2sb79flgvswg7aj-stdenv-linux/setup: line 1341: /tmp/nix-env-220201-102621.tmp: No such file or directory
error: builder for '/nix/store/b63h8gyl2sjy6n27j5gf9yakmix1a3qn-status-react-build-jsbundle-android.drv' failed with exit code 
```
It's as if the file is not being passed to the sandbox.

jakubgs:
But I tested this locally with some extra `set -x` and I can see it begin passed:
```
 > make release-android KEYSTORE_PATH=/tmp/keystore INFURA_TOKEN=xyz
Configuring Nix shell for target 'keytool'...
Keystore file already exists: /tmp/keystore
/tmp/keystore
scripts/release-android.sh
Missing TARGET, assuming default target. See nix/README.md for more details.
Configuring Nix shell for target 'default'...
Running: nix-build --pure --fallback --no-out-link --show-trace --attr targets.mobile.android.release --argstr secretsFile /tmp/tmp-status-react-d063549a3/tmp.ZHxQr01odc --argstr buildEnv prod --option build-use-sandbox true --option extra-sandbox-paths /tmp/keystore /tmp/tmp-status-react-d063549a3/tmp.ZHxQr01odc --arg config {status-im.android.keystore-path="/tmp/keystore";status-im.commit-hash="d063549a39881c481302d9e2731cc9bb748591fb";status-im.build-type="nightly";status-im.build-number="2021092112";status-im.android.apk-signed="true";status-im.android.abi-split="false";status-im.android.abi-include="armeabi-v7a;arm64-v8a;x86";} default.nix
these 2 derivations will be built:
  /nix/store/5dvzcx514va6yhzr2hn790yrpc9hw3z8-status-react-build-jsbundle-android.drv
  /nix/store/m4bl4c8g042gprsqh9xqd4r6ikvvnqfz-status-react-build-nightly-android.drv
building '/nix/store/5dvzcx514va6yhzr2hn790yrpc9hw3z8-status-react-build-jsbundle-android.drv'...
unpacking sources
unpacking source archive /nix/store/98lykj9yw86sv9snwn7ik8chkrg2mp8b-status-react-source-jsbundle
source root is status-react-source-jsbundle
secretsPhase
--------------------------------------------------
++ source /tmp/tmp-status-react-d063549a3/tmp.ZHxQr01odc
+++ export INFURA_TOKEN=xyz
+++ INFURA_TOKEN=xyz
++ [[ -z xyz ]]
++ set +x
--------------------------------------------------
```
So what is this about?

jakubgs:
I tried the simplest derivation I can imagine to test what's the issue:
```nix
{ pkgs ? import <nixpkgs> { } }:

pkgs.stdenv.mkDerivation rec {
  name = "nix-sandbox-test";
  builder = pkgs.writeScript "${name}-builder" ''
    source $stdenv/setup
    cp /tmp/sandbox-file $out
  '';
}
```
And it fails as well:
```
jenkins@linux-04:~/nix-sandbox$ echo TEST > /tmp/sandbox-file

jenkins@linux-04:~/nix-sandbox$ ls -l /tmp/sandbox-file
-rw-rw-r-- 1 jenkins jenkins 5 Feb 17 21:19 /tmp/sandbox-file

jenkins@linux-04:~/nix-sandbox$ nix-build --no-out-link --option extra-sandbox-paths '/tmp/sandbox-file'
this derivation will be built:
  /nix/store/q545vrhp2gq550w9vr730x0gv871bczg-nix-sandbox-test.drv
building '/nix/store/q545vrhp2gq550w9vr730x0gv871bczg-nix-sandbox-test.drv'...
cp: cannot stat '/tmp/sandbox-file': No such file or directory
error: builder for '/nix/store/q545vrhp2gq550w9vr730x0gv871bczg-nix-sandbox-test.drv' failed with exit code 1;
       last 1 log lines:
       > cp: cannot stat '/tmp/sandbox-file': No such file or directory
```
So something seems definitely wrong with sandboxing in a multi-user installation. And the `nixbld*` users can read the file:
```
admin@linux-04.he-eu-hel1.slave.ci:~ % sudo -u nixbld1 cat /tmp/sandbox-file
TEST
```

jakubgs:
Interestingly enough this works quite fine on my NixOS setup:
```
 > nix-build --no-out-link --option extra-sandbox-paths '/tmp/sandbox-file'       
this derivation will be built:
  /nix/store/qqv5gq6agqfbrp6qicsk6ksmny6b29vl-nix-sandbox-test.drv
building '/nix/store/qqv5gq6agqfbrp6qicsk6ksmny6b29vl-nix-sandbox-test.drv'...
/nix/store/20j9hq1yrww7brg90i5gxqj2ps4pny6h-nix-sandbox-test

 > cat /nix/store/20j9hq1yrww7brg90i5gxqj2ps4pny6h-nix-sandbox-test               
TEST
```

jakubgs:
If I `strace` the `nix-daemon` process I can see an `ENOENT` error:
```
admin@linux-04.he-eu-hel1.slave.ci:~ % PID=$(systemctl show --property MainPID --value nix-daemon)                   
admin@linux-04.he-eu-hel1.slave.ci:~ % sudo strace -p $PID -f -e trace=%file >/dev/null 2>&1 | grep /tmp/sandbox-file
[pid 1457879] execve("/nix/store/vizjhz04x6xl57x2vrpqa52j8q6rkjfh-coreutils-9.0/bin/cp", ["cp", "/tmp/sandbox-file", "/nix/store/984lk190lqyy1vji4y07n"...], 0x5dc010 /* 69 vars */) = 0
[pid 1457879] newfstatat(AT_FDCWD, "/tmp/sandbox-file", 0x7fffffffd3b0, 0) = -1 ENOENT (No such file or directory)
```
Which means the `nixbld*` users are not seeing the file for some reason.

jakubgs:
I've created an issue in Nix repository outlining the sandbox problem:

* https://github.com/NixOS/nix/issues/6115

jakubgs:
The issue with the sandbox is lack of defined `truested-users`, which should include `jenkins`: https://github.com/NixOS/nix/issues/6115#issuecomment-1060626260

jakubgs:
Looks like adding `jenkins` user to `trusted-users` in https://github.com/status-im/infra-ci/commit/8d6b6b3f did work, but now we get:
```
[2022-03-07T15:48:15.918Z] error: getting attributes of path ****: Permission denied
```
So looks like `nixbld*` users can't access files created by Jenkins credentials systemd.

jakubgs:
Interestingly enough, the file permissions are already quite broad, with read for all users:
```
18:24:21  + ls -l /tmp/sandbox-file.tmp
18:24:21  -rw-rw-r-- 1 jenkins jenkins 243 Mar  7 17:24 /tmp/sandbox-file.tmp
```

jakubgs:
I tried changing file permissions but it doesn't seem to work:
```
19:43:42  + ls -l ****
19:43:42  -r-------- 1 jenkins jenkins 2163 Mar  7 18:43 ****
[Pipeline] sh
19:43:43  + chgrp nixbld ****
19:43:43  + chmod 664 ****
[Pipeline] sh
19:43:44  + ls -l ****
19:43:44  -rw-rw-r-- 1 jenkins nixbld 2163 Mar  7 18:43 ****
```

jakubgs:
Ah, I see, there it is. The issue is the permissions of the folder:
```
jenkins@linux-03:~$ ls -l /home/jenkins/workspace/tatus-react_prs_android_PR-12980@tmp/
total 8
drwxrwxr-x 2 jenkins jenkins 4096 Mar  7 19:03 durable-1d67657a
drwx------ 2 jenkins jenkins 4096 Mar  7 19:02 secretFiles
```

jakubgs:
The issue on iOS with this error:
```
error: getting status of '/private/tmpxxx-tmp-status-react-cb10ffd99': No such file or directory
```
Appears to be fixed in Nix `2.6.1`, as I described in https://github.com/NixOS/nix/issues/6214.

jakubgs:
For some reason setting `750` isn't enough. I wants `775` which is far too broad.

I need to do more testing to figure out how to narrow down the permissions as much as I can.

jakubgs:
I did some stracing and found this in `nix-daemon` syscalls:
```
2291348 newfstatat(AT_FDCWD, "/tmp/sandbox-dir/test", 0x7f96160f9ac0, 0) = -1 EACCES (Permission denied)
```
And according to the docs for `newfstatat`:

>These functions return information about a file, in the buffer pointed to by statbuf. __No permissions are required on the file itself__, but—in the case of stat(), fstatat(), and lstat() - __execute (search) permission is required on all of the directories in pathname that lead to the file__.

But the folder already has execute permissions for `bixbld` group:
```
 > ls -l /tmp | grep sandbox-dir
drwxr-x--- 2 jakubgs nixbld     3 Mar  8 10:29 sandbox-dir
```
So there's something else missing. Maybe the group I'm assuming is necessary is wrong.

jakubgs:
I've managed to fix the permission issue with: https://github.com/status-im/status-jenkins-lib/commit/91a22e59

But I would really prefer if I could narrow down the permissions to just group ones instead of "other".

jakubgs:
This change is ready, but since a new release is brewing I'm going to wait with merging this until the release is out.

cammellos (approved):

PR #12987, opened 2022-01-10
update react-native-webview

### LINKED ISSUE #11565: app.tornado.cash crashes app on Android with 'java.lang.illegalArgumentException'
# Bug Report

## Problem
Initially reported by a user.
At attempt to deposit ETH via dapp `app.tornado.cash` status crashes with error:
Screenshot_20201220-105743_Status
Works with MetaMask in Chrome and with Status on IOS.

#### Expected behavior
can deposit ETH or gracefully handled error

#### Actual behavior
crash

### Reproduction

- Open Status
- Restore multiaccount with seed phrase (with at least 0.1 ETH to check the case)
- Open `app.tornado.cash`
- Connect wallet
- Try to deposit 0.1 ETH

### Additional Information

- Status version: nightly 21/12/20, release 1.9.0
- Operating System: Android only

#### Logs

[Status-debug-logs.zip](https://github.com/status-im/status-react/files/5723473/Status-debug-logs.zip)

### LINKED ISSUE #12726: ReactNativeWebView.onMessage is not a function
# Bug Report
When browsing dapp `swap.eros.fund` with android 10+, everything works fine, but with android 8.1, you encounter the problem shown in the title

### Reproduction

1) using Android Studio -> AVD Manager to create a virtual device, when creating, select "Phone" from Category, select "Pixel XL" from name list, click "next", select "Oreo" which api level is 27, ABI is x86, and most important the target should use Android 8.1(Google APIs)
2) install status app(version 1.16) into the newly created device above
3) open status login and navigate to browser tab, type `swap.eros.fund` in the address bar and go
4) you will see a screenshot like this one 

5) open chrome, type `chrome://inspect`, u will see like this

6) click inspect and navigate to `Console`
7) click the button in the red box as shown in the picture `... `

8) Connect wallet 
9) choose Allow  

10) back to DevTools

### Additional Information
a work solution is: edit package.json, and change dependency "react-native-webview" to "git+https://github.com/status-im/react-native-webview.git#js_injection_before_load", then won't have this issue on android 8.1 , i have checked `https://github.com/status-im/react-native-webview` , looks like the latest tag or branch master didn't merge branch `js_injection_before_load`, why?

- Status version: 1.16 
- Operating System: Android

bitgamma:
Fixes #11565. 
Fixes #12726. 

Upgrades react-native-webview to the latest version + our changes to permission handling. It also makes send-to-bridge more robust to dapp invoking web3 methods before the page is fully loaded

Since the react-native-webview module has changed we need to make sure no regressions.

### DIFF
--- a/nix/deps/gradle/deps.json
+++ b/nix/deps/gradle/deps.json
@@ -1148,6 +1148,20 @@
   },
 
   {
+    "path": "androidx/webkit/webkit/1.4.0/webkit-1.4.0",
+    "host": "https://dl.google.com/dl/android/maven2",
+    "type": "aar",
+    "pom": {
+      "sha1": "c67e70da5ad98b587cc030a053e3eb45b57ae2a6",
+      "sha256": "0pdmx702dmx83q6svafk1vyac9fnlbz0szxwg4ax9fnlzvc12mfb"
+    },
+    "jar": {
+      "sha1": "48e4cd9c63f7dc9fc62631d8dfbb77e890b459aa",
+      "sha256": "15wip2fimx1xpdv6mwdyn2wsd2yfh9kbfrjkv9rdwlhkmh5b448s"
+    }
+  },
+
+  {
     "path": "android/arch/core/common/1.0.0/common-1.0.0",
     "host": "https://dl.google.com/dl/android/maven2",
     "type": "jar",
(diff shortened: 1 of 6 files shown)


flexsurfer (approved):

qoqobolo:
@bitgamma thanks for your work!

Unfortunately, I couldn't fully test it as well, since the DApp was stuck on the endless loading after the wallet connection. 

The DApp behaves like this on both Android and iOS; on Nightly and PR builds. Also, we tried to open it from different locations (i.e. Ukraine and Spain).
So it looks like there is a problem on the DApp's side (please, correct me if we can fix it).

https://user-images.githubusercontent.com/67952253/149348168-a8883567-e32d-4088-a1a2-7c1600338626.MP4

Regarding https://github.com/status-im/status-react/issues/12726, I could reproduce the same error following the steps described in the issue on the PR build `8ba06eb`.

photo_2022-01-13_15-30-08

bitgamma:
@qoqobolo I will check if I find a solution for these problems, maybe they can be fixed too. I tried the Nova version of the tornado cash and it didn't have this behaviour and it did not crash (but I had no funds in the test account so I couldn't finish the setup, don't know what happens next)

qoqobolo:
@bitgamma yeah, there is no crash with the Nova version. And since the site was changed probably it will be enough for merge. 
Please, ping me once it becomes clear if these problems can be fixed.
Thanks!

bitgamma:
@qoqobolo I am almost sure that #12726 is fixed now (with my setup it required many many retries to reproduce, but now I can never reproduce it).

The refresh problem with tornado.cash is still under investigation, but I'd need to debug the dApp itself and see if we can find a workaround. Probably not the scope of this PR.

qoqobolo:
@bitgamma okay, let's merge this one then.
And thanks a lot for investigating and fixing these cases!

PR #13028, opened 2022-01-20
Implementation of remote android notifications

### LINKED ISSUE #12866: Implement remote android notifications
As a user
I want to have an option to have remote notifications instead of just local
So I don't have to run a foreground service

### Description

We should give users an option to enable remote notifications for android as well as IOS, as some users are put off by the foreground service.

The backend should be mostly implemented, but we need to set up firebase, pass a token etc.

The library we currently use for IOS is currently not maintained, so maybe it's worth looking into moving to a different library:

https://github.com/invertase/notifee

or 

https://github.com/wix/react-native-notifications

For example.

Another thing that is important is that the resulting build still has any google dependencies removed, in order to make sure it's releasable on fdroid.
cc @jakubgs that can make sure that's the case.

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/12866

PR implements android remote notifications

### Code Review Requests:
I tried to make sure google dependencies are removed in the f-droid release build. Please, let me know if I missed something. cc @jakubgs

### QA test request:
Changes for android remote notifications may have affected ios remote notifications, please let me know if anything is not working as expected.

### Known issue:
PR disables remote notifications for an account at the time of logout. But if the user logs in into another account, without logging out or disabling notifications manually (by logging in while opening the app), then the user will keep receiving notifications for another account

### Extras:
#12866 also mentions migrating ios library too. I will create a separate issue/pr for this.

### status-go PR
https://github.com/status-im/status-go/pull/2527

status: ready


du82:
This should be an opt-in option, since Google listens to notifications, which would compromise anonymity and privacy of users.

Parveshdhull:
> This should be an opt-in option, since Google listens to notifications, which would compromise anonymity and privacy of users.

Yes, it will be 👍

flexsurfer:
hey @Parveshdhull are you considering conditioning this for f-droid builds ? there must be no google services dependencies

Parveshdhull:
> hey @Parveshdhull are you considering conditioning this for f-droid builds ? there must be no google services dependencies

Hi @flexsurfer, first we trying to see if we can make remote android notifications work for android. Once it is complete, I will make sure to condition google dependencies before merging 👍

package.json
```diff
@@ -27,6 +27,7 @@
     "emojilib": "^2.4.0",
     "eth-phishing-detect": "^1.1.13",
     "functional-red-black-tree": "^1.0.1",
+    "hermes-engine": "0.5.2-rc1",
```
Parveshdhull, line 30: **TODO** - remove `hermes-engine 0.5.2-rc1` once react-native is upgraded to 0.64+ _(will create separate issue for this)_
https://github.com/wix/react-native-notifications/issues/455#issuecomment-803413032

scripts/google-free.sh
```diff
@@ -0,0 +1,16 @@
...
+#------------------------------------
+
+yarn remove react-native-notifications
+rm android/app/google-services.json
+sed -i -e '/com.wix.reactnativenotifications.RNNotificationsPackage;/d' android/app/src/main/java/im/status/ethereum/MainApplication.java
+sed -i -e '/com.google.gms:google-services/d' android/build.gradle
+sed -i -e '/react-native-notifications/,+2d' android/settings.gradle
+FIREBASE_LINE=`grep -n 'com.google.firebase.messaging.default_notification_icon' android/app/src/main/AndroidManifest.xml | cut -d : -f 1`
+sed -i -e "$((FIREBASE_LINE-1)),+5d" android/app/src/main/AndroidManifest.xml
+sed -i -e 's/\["react-native-notifications" :refer (Notifications)\]//' src/status_im/notifications/android_remote.cljs
```
jakubgs: Is this really the only way to do this? This is not very petty, I have to tell you.

jakubgs: This stiff will EASILY break in the future, it's very fragile.

Parveshdhull: Hi jakubgs, Thank you very much for you review.
- We can't condition some things like java imports, removing them at the build time seems only option 🤔
- I improved the script and removed breakable parts 👍. The only part which now can break(imo) (rare case) is 
`sed -i -e 's/\["react-native-notifications" :refer (Notifications)\]//' src/status_im/notifications/android_remote.cljs`
Sorry, I don't know how I can replace this one (looking for suggestion)

jakubgs: But why are you using Sed on `android/build.gradle`? That can be conditional.

>We can't condition some things like java imports, removing them at the build time seems only option

Why can't we use [`Class.forName()`](https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#forName(java.lang.String))? That can be conditional.

Parveshdhull: removed almost all of the sed lines 👍
(looks like, there inclusion without their counter part didn't affect resulting apk for fdroid build - checked sha1sum)
Only one more to go, any suggestions, how I remove this one?

jakubgs: I'm not a Clojure wizard, but what about the `:provided` option?
https://stackoverflow.com/a/34134158

Maybe @flexsurfer has some better ideas.

Parveshdhull: @jakubgs, all sed commands are removed, Please let me know if anything else can be improved.
Thanks, @flexsurfer for pointing me to [Conditional Reading](https://shadow-cljs.github.io/docs/UsersGuide.html#_conditional_reading), great feature.

jakubgs: That's great, thank you!

scripts/google-free.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+sed -i -e '$aGOOGLE_FREE=1' .env.release
+
+#------------------------------------
+# Remove Android Remote Notifications
```
jakubgs: This comment isn't very useful. It just describes what happens bellow.
Comments should describe __why__ something is happening, not __what__.
Same goes for the line above, it would be nice to add a comment like:
```sh
# Gradle parameter that makes some dependencies optional.
sed -i -e '$aGOOGLE_FREE=1' .env.releases
```
This makes it clear why the change is made, not what is being done, which is clear from the command itself.

jakubgs: Actually, that comment isn't fully accurate, because it is also used in Clojure.

scripts/google-free.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
```
jakubgs, line 1: One last thing, it might be sensible to add `set -e` to have it fail if any of the commands fail.

jakubgs (approved):
Thanks for getting rid of all of those disgusting `sed` commands.

package.json
flexsurfer: also please add a comment

flexsurfer (approved):

pavloburykh:
@Parveshdhull thank you for PR! 
I'm currently in process of testing the PR and I got a question about notification enabling after login: as you described here https://github.com/status-im/status-react/pull/13028#issue-1109779720 at this moment, remote notifications are disabled for an account at the time of logout. I fact, when user logs out the account both remote and background notifs are disabled. But after re-login neither remote nor  background notifls are not re enabling. Is this known issue that we accept for now or this can be considered as a bug? For me as a user, expected behaviour would be reenabling notifications after login if they where enabled before logout and user did not disabled them manually. 

https://user-images.githubusercontent.com/97245802/153401169-267edfd8-5e76-4f2f-bd6d-9ca4db869542.mp4

Parveshdhull:
Hi @pavloburykh, Thanks for testing the PR. I think you are right, they should be enabled automatically on login again. Let me fix this 👍

Parveshdhull:
should be fixed now 👍

pavloburykh:
thanx @Parveshdhull

pavloburykh:
@Parveshdhull please, take a look at this issue
## ISSUE 2 (Android)
PN about sending/receiving transactions are not received when remote notifications are enabled. 

STR: 
1. On Device A enable remote notifications and wallet transaction notifications 

photo_2022-02-11 07 57 19

2. Send transaction to Device A 
3. Send transaction from Device A
4. Pay attention if there are any PN about transactions. 
Below, I attach the video demonstrating process of sending eth to another wallet and receiving PN about sending of transaction with Background notifications enabled. Also. I attach screenshots of receive/send transactions PN from Android with Background notifications enabled.


https://user-images.githubusercontent.com/97245802/153545356-da1d0a34-2a01-49a5-95b6-7c3d56edebfb.mp4

photo_2022-02-11 08 16 37


Actual result: no PN about send/receiving of transactions.

Expected result: PN about  sending/receiving of transactions are received. 

### How to generate transactions using Ropsten testnetwork: 
1. Switch to Ropsten network in Status settings: Profile - Advanced - Network. For PR builds Ropsten is as set as default. 
2. To get some test eth, use one of Ropsten faucets for example https://faucet.ropsten.be/ or ping anyone from mobile team and we will send some test eth to you address.
3. After you receive test eth to your Status wallet - use them for generating transactions between accounts. 

Please, let me know if you need any help with replicating or generating transactions.

flexsurfer:
sorry probably this was discussed before, but settings look strange
image
we need to somehow make them similar and add a description for notification, or rename to `Local notifications` and add description `service on your device will be running`

Parveshdhull:
Hi @pavloburykh, Thanks for adding steps to reproduce 2nd issue. I think currently remote PN is only supported for messages, not wallet transactions (issue also exists in the ios - nightly version). Maybe we can work on this issue separately? wdyt

pavloburykh:
@Parveshdhull I'm receiving wallet PN on IOS nightly version (attaching video below). But anyway, I agree with you, that we can add remote wallet PN for Android by separate issue in order not to block this PR. 


https://user-images.githubusercontent.com/97245802/153583143-e12a634d-0a94-4bff-bb82-09f7f539048c.mp4

pavloburykh:
> sorry probably this was discussed before, but settings look strange image we need to somehow make them similar and add a description for notification, or rename to `Local notifications` and add description `service on your device will be running`

@Parveshdhull in terms of this comment I would like to add, that currently Notification switcher seems to have 2 functions - enables/disables notifications overall and enables background notifications in case if Remote notification switcher is disabled (attaching video below demonstrating current behaviour). It might be a little bit confusing for a user.  And better decision would be to rename Notification switcher to smth like Local notifications as was proposed by @flexsurfer but also to make this switcher independent and responsible only for enabling/disabling Background notifications. 


https://user-images.githubusercontent.com/97245802/153584521-19a0e29f-5af4-4662-8283-efdbbc9a2f4a.mp4


I think we might think to improve above issues in case if Notifications version implemented by this PR is considered to be final version. 

In case if we consider this version to be temporary one and we are planning to improve it in nearest future then we are good to go at this moment without additional UX/UI fixes.

Parveshdhull:
Hi @pavloburykh, I updated the UI as requested by you and @flexsurfer.
Also for ISSUE 2, I think we need to show a local notification for wallet transactions if the app is running(and remote notifications are enabled) as we can't send confidential information in remote notifications. Please feel free to create a separate issue for that and assign me. And please let me know if new changes are working as expected. Thank you

Parveshdhull:
> Device 1: Button by accessibility id: `notifications-settings-button` is not found on the screen

This is now replaced by` :local-notifications-settings-button`

pavloburykh:
> * Failure screenshot

Thanx @Parveshdhull we will update test.

pavloburykh:
@Parveshdhull Thanx for all fixes. PR is now ready for merge.

PR #13039, opened 2022-01-24
Use HTTPs for image loading with self-signed certs

### LINKED ISSUE #12855: Load images from an https server
For performance reasons, we should load images from an https server run locally, instead of passing them through the bridge.

The issue is to automate the creation on certificates and making sure that they are trusted.

On IOS we can follow: 

https://developer.apple.com/documentation/foundation/url_loading_system/handling_an_authentication_challenge/performing_manual_server_trust_authentication

while on Android we could use https://stackoverflow.com/a/57046889/13685841 , but it might require changing fast-image library.

The status-go PR is here https://github.com/status-im/status-go/pull/2418 , I can update it to use SSL, @flexsurfer has done already some work here https://github.com/status-im/status-react/pull/12780/files (changing in the frontend are minimal, other than accepting the certificate).

### LINKED ISSUE #13081: Spinner in 1-1 chat when navigating to the chat view when history contains some amount of images (>30 in particular case)
# Bug Report

## Problem 
Originally reported by Jarrad.
When opening 1-1 chat with a long history, which contains images there can be a delay (spinner several seconds)  in chat history loading.

#### Expected behavior
can see history instantly

#### Actual behavior
loading spinner (can be several sec, depends on images amount)

https://user-images.githubusercontent.com/4557972/152765989-e4f0551d-741a-4d08-bfbc-d652d4b10f9c.mp4

### Reproduction

1) Open Status
2) Send in 1-1 chat 30+ messages with photos (1,5-3 MB)
3) Go back, open chat  again

### Additional Information

- Status version: nightly 

[logs (12).zip](https://github.com/status-im/status-react/files/8014243/logs.12.zip)

bitgamma:
Closes #12855.
fixes: #13081


src/status_im/chat/models/images.cljs
```diff
@@ -37,31 +39,27 @@
...
-      base64-uri
-      width
-      height
-      100
-      (fn [^js resized-image]
-        (let [path (.-path resized-image)
-              path (if (string/starts-with? path "file") path (str "file://" path))]
-          (.save CameraRoll path)))
-      #(log/error "could not resize image" %)))))
+(def temp-image-url (str (fs/cache-dir) "/StatusIm_Image.jpeg"))
```
flexsurfer, line 42: could you please elaborate, why this needed ? what's this image ?

flexsurfer (approved):

src/status_im/chat/models/images.cljs
bitgamma: CameraRoll is expecting a file URL so we have to first download the image locally to cache. Before this was done by saving the base64-encoded image using image-processing/resize, but this didn't work with HTTP(s) links. I added a way to download the image as a binary and unified the code/logic used for saving to gallery and to share image and also made it the same for iOS and Android

qoqobolo:
Hi @bitgamma! Thanks for your work!
I see that some e2e have failed probably due to the outdated go version in this PR. Could you update it, please?
Thanks!

bitgamma:
@qoqobolo rebased! thanks

qoqobolo:
Thanks @bitgamma 

### ISSUE 1: No preview PNG/BMP images in chat view
Can send and open a PNG/BMP image but can't see a preview in a chat.

**Steps:**
1. Download an image in the PNG/BMP format
2. Send it in 1-1 chat

Android:

https://user-images.githubusercontent.com/67952253/152825309-950bd2c4-fdfb-4b77-9a18-af6e5bfe2ab7.mp4

### ISSUE 2: `Close`, `Save` and `Share` buttons are moved up and overlapped with the status bar in the image view mode  (iOS only)
So that they cannot be used.

Screenshot 2022-02-07 at 17 08 57

Also, noticed that the image itself seems to be shifted down, and it changes its brightness as it moves.
However, this behavior is reproducible in develop as well. Can you take a look at it in this PR? If it's not very easy to fix, I'll report it separately.
 
https://user-images.githubusercontent.com/67952253/152827119-35e4b314-795a-48f1-a88f-aab1dda9ae89.MP4

flexsurfer:
i'll fix ISSUE 2

flexsurfer:
> Also, noticed that the image itself seems to be shifted down, and it changes its brightness as it moves.

I'll take a look

flexsurfer:
ISSUE 1 probably for some reason there is no on-load event

bitgamma:
@qoqobolo thanks for testing! will let you know when it is fixed.
@flexsurfer regarding ISSUE 1 it could be the onload, yes. Actually, we could even revert to using the normal Image tag instead of fast-image and the image-get-size function since they both work now. Wdyt?

flexsurfer:
thanks @qoqobolo fixed

qoqobolo:
Thank you @flexsurfer!

### ISSUE 3: Endless loading when opening an image (any format) after screen lock (iOS)

Noticed another unexpected behavior. 
If open a chat or Timeline with a picture, lock the screen on this view, unlock it in ~1 minute, and open the picture, the endless spinner will appear instead of the picture view. 
After relogin, it works fine.
Can reproduce on iOS only.  
Tried on the latest build `ff005fe`

This one is not critical, but might be worth a look.

**Steps:**
1. Open chat with a picture
2. Lock the screen on your device
3. Wait for ~ 1 minute (to catch it before logout)
4. Unlock the screen
5. Open the image

https://user-images.githubusercontent.com/67952253/152970785-6c133254-161f-4734-a33d-ef8d5f2f4dfc.MP4

### ISSUE 4: Gray line on the left side of a horizontal picture in Timeline

The issue is in nightly as well. But if you can fix it here, it would be nice!
Screenshot 2022-02-08 at 12 05 31

flexsurfer:
ISSUE 3:  looks like critical, seems like http server is stopped

qoqobolo:
> ISSUE 3: looks like critical, seems like http server is stopped

Oh, okay. Didn't consider it as critical since could only reproduce it on iOS, relogin helps, and also you need to catch it before logout while in background.
Thanks for the clarification!

bitgamma:
Indeed ISSUE 3 seems to be reproducible by locking the screen and waiting a bit, regardless of what tab/chat is open in the Status client so it is quite major. I am trying to find out what exactly happens in that case and how we can recover from that situation

qoqobolo:
@flexsurfer Also, can you take a look at the second part of ISSUE 2, please?
> Also, noticed that the image itself seems to be shifted down, and it changes its brightness as it moves.
However, this behavior is reproducible in develop as well.

It's more visible in horizontal images. When opened, the picture is slightly darkened, but you can move it up, and it becomes lighter. Looks like it should be light and centered when opened. 

https://user-images.githubusercontent.com/67952253/152993821-41a3fbed-e0c5-4dbb-abf5-9dbbe0c8eedd.MP4

flexsurfer:
@qoqobolo unfortunately it's in the component https://github.com/jobtoday/react-native-image-viewing/issues/23 , dunno why it's not fixed, it's so essential issue

bitgamma:
@qoqobolo issue 3 is now solved. Thanks!

qoqobolo:
Thanks @bitgamma! Unfortunately, now the app crashes on iOS

### ISSUE 5: Crash when creating/recovering an account after creating password (iOS)
After the crash can reopen, log in, and use the app.

**Steps:**
1. Install the `4f7fded` build
2. Follow the account creation flow > create and confirm your password > tap `Next`

[Crashlog - StatusPR 9-2-22,11-10.log](https://github.com/status-im/status-react/files/8031701/Status.PR.9-2-22.11-10.log)


https://user-images.githubusercontent.com/67952253/153178472-0c893da5-e117-4977-ba96-e3a851af1668.MP4

bitgamma:
@qoqobolo ISSUE 5 should be fixed now. Thank you for your patience!

qoqobolo:
Thank you too, @bitgamma!

@flexsurfer @bitgamma need your opinion on one more thing (hopefully the last one here).

Noticed, that when opening a chat or scrolling it after the first opening on Android devices, image borders are loaded first.
It happens again after re-opening the chat. It is more visible on a slower device and with large images, but can also be reproduced on a fast device and on a new account with small images.
The question is, can it be fixed? WDYT?

It is not reproducible on develop, and this is an Android-specific issue. 

Here is a video with large images at normal speed

https://user-images.githubusercontent.com/67952253/153391540-db6fae5c-1551-45ad-a26b-c8ed583cd5cb.mp4

and in slow motion

https://user-images.githubusercontent.com/67952253/153391648-2ba069f9-6752-456e-a802-4760086973df.mp4

Here are small images, chat without a lot of data, normal speed

https://user-images.githubusercontent.com/67952253/153392060-5673a21a-cc75-4675-b011-36eb24921274.mp4

Slow motion

https://user-images.githubusercontent.com/67952253/153392262-3213f7ef-c105-45a9-8cdb-2c2af3cd418b.mp4

**Steps to reproduce:**
1. On an Android device create a new account
2. Start a 1-1 chat 
3. Send several images
4. Close the chat and reopen it
5. Check the receiver's device

bitgamma:
I'd say this is normal because we are downloading (from status-go) the images on demand as the view scrolls instead of downloading them all at once as we were doing before. This should actually make opening the chat a bit faster at the price of this "effect" when scrolling.

qoqobolo:
@bitgamma @flexsurfer thank you both for the great work, waiting, and patient solving issues!
Failed e2e are not PR-related.
PR is ready to be merged.

flexsurfer:
image

PR #13040, opened 2022-01-24
[#13038] Add error message to email template on crash

### LINKED ISSUE #13038: Get some info from error report in case when app crashes
# Feature Issue

### User Story

As a QA/developer/support manager, I want to get some information from error reports in cases when the user cannot fill it, e.g. if the app crashes.

### Description

We often receive error reports on email from users with <no value>, since in some cases users cannot get a screen with fields to describe the bug, such as when the application crashes.
These error reports have no value, so it would be useful to be able to get at least some information, such as the text of the error, the screen on which it appeared, etc.
So that we could get some idea of the state of the application at the time of the bug and ask clarifying questions to the user.

Screenshot 2022-01-24 at 13 43 19

### Acceptance Criteria

When receiving an error report, be able to see the text of the error/on which screen it occurred/etc, if possible.

cc @rasom

rasom:
fix #13038 

status: ready

### DIFF
--- a/src/status_im/utils/error_handler.cljs
+++ b/src/status_im/utils/error_handler.cljs
@@ -60,4 +60,4 @@
             {:title               "Error"
              :content             (.-message e)
              :confirm-button-text (i18n/label :t/send-logs)
-             :on-accept           #(re-frame/dispatch [:logging.ui/send-logs-pressed :email])})))))))
+             :on-accept           #(re-frame/dispatch [:logging/send-logs-on-error (.-message e)])})))))))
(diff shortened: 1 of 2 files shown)


flexsurfer (approved):

Parveshdhull (approved):

rasom:
@qoqobolo I added a crash on signing into account, you can check that message is added to the email. Message itself is not very readable though :)

qoqobolo:
@rasom thanks! It's okay and better than nothing :)
Checked reports on email from Android and iOS - everything is fine, the message is added.
Let's keep it this way and merge.

Issue #12783, opened 2021-11-02
Improve order of synced chats


cammellos:
As a user
I want to have chat synced in order
So that the most relevant messages are received in a shorted time interval

### Description

Currently once a user goes back online, we fetch all the topics for all the chats in one go.
We should improve the way syncing works by being selective on which chat to sync first.

In order of importance we should sync:

1) The chat the user is in when opening the chat, depends on #12782 
2) Last community channel the user was in
3) Fetch messages from all community channels the user has visited in the past 7 days
4) Fetch all 1-to-1 and ad-hoc group chats
5) Fetch messages from all community channels the user has not visited in the past 7 days
6) Fetch public unmoderated group chats (with the understanding that this 5th step will be removed as soon as we remove public unmoderated group chats from the product)

### Considerations

Some of these only make sense if the interval we are syncing is large enough (1 seems to be almost always a good choice, but after that we probably want to make sure it's worth the overhead, i.e if you are syncing 2 minutes of data, is most likely better to just sync everything in one go, as opposed if you are syncing a couple of hours).

The goal of this issue is to improve syncing, to bring the most relevant messages to the user as quickly as possible, so we should re-evaluate as we go along that it is indeed making things faster, and tests need to be performed with a device with many chats/contacts.

hesterbruikman:
@cammellos not sure how the members list works exactly. Are updates send to a dedicated topic for online status updates? If so, we'd need to discuss where it sits in the fetching priority list (cc @simonricoo )

rasom:
With ~40 pub chats (those which contain spam) after 1h offline it took 40s for 1-1 message be to received after opening chat's screen (that was the last opened chat, so it was opened right after login).

rasom:
1h, [42 4 15 18 7 12 5 7 9 15], avg=13.4
But on average it is not that bad, I will repeat tests with longer offline period.

rasom:
24h, [32 10 25 23 29 27 28 34 30 13], avg=25.1
It gets worse.

rasom:
After three days I just stopped receiving messages in that 1-1. Even if I send new ones, they are not received. Also not all outgoing messages were received on the other side, but idk whether that's because of mailservers. Basically 1-1 chat is broken.

cammellos:
@rasom could you get `geth.log` please? and maybe keep the account so we can debug

rasom:
@cammellos sure

rasom:
[...]

PR #13058, opened 2022-01-28
[#12783] Improve order of mailserver requests

rasom:
fixes (partially) #12783 

Topics are ordered by `chat.ReadMessagesAtClockValue`, which means that chats which were opened recently will have higher priority. That should make receiving updates in the last opened chats faster. The way to test it is to create an acc with 20-30 spam chats and compare how fast messages are received in the last chat (better to use a chat without spam for this) after 2-3h offline.

status: ready

### DIFF
--- a/status-go-version.json
+++ b/status-go-version.json
@@ -4,6 +4,6 @@
     "owner": "status-im",
     "repo": "status-go",
     "version": "v0.93.3",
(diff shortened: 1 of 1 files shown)


Parveshdhull (approved):

churik:
Confirmed, that for recently opened chats history fetched firstly, and then it fetched for other chats.
Tested for public / group / 1-1 chats (15 in total, 200+ messages in each).

PR #13061, opened 2022-01-28
Fastlane: update screenshots

churik:
Updated screenshots for Android.
Also got updated screenshots for IOS, but didn't find where I can update them.


jakubgs (approved):
I think iOS stuff has to be updated by hand in their WebUI. Correct @cammellos ?

jakubgs:
Actually, are those image different for different languages?

jakubgs:
Oh, they are, okay.

PR #13107, opened 2022-02-14
fix ens name registration

flexsurfer:
fix ens name registration


cammellos (approved):

jakubgs (approved):
Why are we skipping a version?

cammellos:
@jakubgs my bad, we didn't bump to 1.8.1 in the repo the version, but 1.8.0 was hotfixed and only released on the play store beta, so this is correct, it should be 1.8.2

PR #13138, opened 2022-03-01
Add changelog from status-go on upgrade

### LINKED ISSUE #13090: Add changelog from status-go on upgrade
If status-go is upgraded in status-react, we should add in the commit the changelog of the changes, this should be done automatically through hooks. Ideally that's reflected as well in the PR description so QA is aware of the changes brought in each PR.

audriu:
fixes #13090

### Summary

If status-go is upgraded in status-react, add the commit the the last PR description of status-go

### Review notes
I use Github API to get the info.

### Testing notes
Run `scripts/update-status-go.sh "tag"` and observe status-go-version.json is updated with last PR message (along of other changes)

#### Platforms
- command line scripting

### Steps to test

- cd status-react
- scripts/update-status-go.sh v0.94.9
- cat status-go-version.json
- observe last PR message added to status-go information

status: WIP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: nix-add-gcroots clean nix-clean run-metro test release _list _fix-node-perms _tmpdir-mk _tmpdir-rm
+.PHONY: nix-add-gcroots clean nix-clean run-metro test release _list _fix-node-perms _tmpdir-mk _tmpdir-rm _install-hooks
 
 help: SHELL := /bin/sh
 help: ##@other Show this help
@@ -123,6 +123,11 @@ _tmpdir-mk: ##@prepare Create a TMPDIR for temporary files
 _tmpdir-rm: SHELL := /bin/sh
 _tmpdir-rm: ##@prepare Remove TMPDIR
 	rm -fr "$(TMPDIR)"
+
+_install-hooks: SHELL := /bin/sh
+_install-hooks: ##@prepare Create prepare-commit-msg git hook symlink
+	@ln -s ../../scripts/hooks/prepare-commit-msg .git/hooks
+-include _install-hooks
 
 # Remove directories and ignored files
 clean: SHELL := /bin/sh
--- a/scripts/hooks/prepare-commit-msg
+++ b/scripts/hooks/prepare-commit-msg
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ -z $(git diff --cached --name-only -r | grep status-go-version.json) ]]; then
+  exit 0 # No status-go-version.json changes found.
+fi
+
+MERGE_BASE=$(git merge-base -a develop "$(git rev-parse --abbrev-ref HEAD)" develop)
+GO_COMMIT_MERGE_BASE=$(git show "${MERGE_BASE}":status-go-version.json | jq -r '."commit-sha1"')
+GO_COMMIT_CURRENT=$(jq -r '."commit-sha1"' status-go-version.json)
+
+GITHUB_LINK="https://github.com/status-im/status-go/compare/${GO_COMMIT_MERGE_BASE}...${GO_COMMIT_CURRENT}"
+
+COMMIT_MSG_FILE=$1
+#todo for now instead of sed I use this just to have it all working. Later will implement sed solution
+if [[ -f "${COMMIT_MSG_FILE}" ]]; then
+  if [ -f temp_file ] ; then
+          rm temp_file
+  fi
+  while IFS= read -r line; do
+    if [[ "$line" == *"https://github.com/status-im/status-go/compare"* ]]; then
+      echo "${GITHUB_LINK}" >>temp_file
+    else
+      echo "${line}" >>temp_file
+    fi
+  done <"${COMMIT_MSG_FILE}"
+
+  if ! grep -q "https://github.com/status-im/status-go/compare" temp_file; then
+    echo "${GITHUB_LINK}" >>temp_file
+  fi
+
+  cat temp_file >"${COMMIT_MSG_FILE}"
+fi


scripts/update-status-go.sh
```diff
@@ -74,6 +75,7 @@ fi
 
 # ls-remote finds only tags, branches, and pull requests, but can't find commits
 STATUS_GO_MATCHING_REFS=$(git ls-remote ${REPO_URL} ${STATUS_GO_VERSION})
+STATUS_GO_COMMIT_MESSAGE=$(curl $(curl ${REPO_API_URL}/git/ref/tags/${STATUS_GO_VERSION} | jq -r .object.url) | jq -r .message)
```
jakubgs: Why would you assume the ref provided as argument is always a tag?

jakubgs (changes requested):
What's the point of this change?

scripts/update-status-go.sh
audriu: Now I figured out what is really to be done and will fix soon...

jakubgs: Calling GitHub API is a bit ugly, but unfortunately we cannot get commit messages from `git ls-remote`.

jakubgs: Since we have to hanndle multiple types of objects - commits, branches, tags -  this code will have to be more complex than it is right now, which is a bit annoying.

jakubgs: Also, small note, if you just call `curl` without any flags it will fail silently, without an error return code.
You can provide `-f`/`--fail` to make it return an error, if that's what we actually want.
The other thing to possibly use is `-sS`, to make it silent, but show errors. And possibly `-L` for redirects.

jakubgs: One possible way to do this without calling the API is to fetch the object into the local repo(`status-react`) one:
```sh
~/work/status-react develop
 > g fetch --depth=1 status-go develop
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
From https://github.com/status-im/status-go
 * branch                develop    -> FETCH_HEAD

~/work/status-react develop
 > g log --format=format:'%s' status-go/develop
load audio from http server
```
Which is also a bit ugly, but we could clean up after that. Also, need to check how that works with tags.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED=$(git diff --cached --name-only -r | grep status-go-version.json)
+
+if [ -n "$STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED" ]
+then
+  MERGE_BASE=$(git merge-base -a develop "$(git rev-parse --abbrev-ref HEAD)" develop)
+  GO_COMMIT_USED=$(git show "$MERGE_BASE":status-go-version.json | jq -r '."commit-sha1"')
+
+  GITHUB_LINK="https://github.com/status-im/status-go/compare/$(git rev-parse HEAD)...$GO_COMMIT_USED"
```
jakubgs: This `git rev-parse HEAD` doesn't make sense. We're comparing `status-go` commits, not `status-go` commits and `status-react` commits.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
jakubgs: User `env` and `bash`:
```suggestion
#!/usr/bin/env bash
```
We're not going to be running this on some ancient AIX machines.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,17 @@
...
+STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED=$(git diff --cached --name-only -r | grep status-go-version.json)
+
+if [ -n "$STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED" ]
```
jakubgs: Why not invert the if and quit early? Less nesting is always good.
```suggestion
if [[ -z "${STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED}" ]]; then
  return # No status-go-version.json changes found.
fi
```
Also, use double square brackets for Bash.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,17 @@
...
+  MERGE_BASE=$(git merge-base -a develop "$(git rev-parse --abbrev-ref HEAD)" develop)
+  GO_COMMIT_USED=$(git show "$MERGE_BASE":status-go-version.json | jq -r '."commit-sha1"')
+
+  GITHUB_LINK="https://github.com/status-im/status-go/compare/$(git rev-parse HEAD)...$GO_COMMIT_USED"
+
+  echo "$GITHUB_LINK"
+
+  COMMIT_MSG_FILE=$1
+  if [ -f "$COMMIT_MSG_FILE" ]; then
+    echo "$GITHUB_LINK" >> "$COMMIT_MSG_FILE"
```
jakubgs: This just adds a link, but what about amending commits? We also need some logic to replace the existing link.

Probably with `sed`.

scripts/update-status-go.sh
jakubgs: This is irrelevant now.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+COMMIT_MSG_FILE=$1
+STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED=$(git diff --cached --name-only -r | grep status-go-version.json)
+
+if [[ -z "${STATUS_GO_VERSION_FILE_IS_BEING_COMMITTED}" ]]; then
+  exit 0 # No status-go-version.json changes found.
+else
```
jakubgs: You don't need the else if you exit. Just close it with `fi` and move on.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
jakubgs, line 1: I wonder if we also might want `set -euo pipefail` at the beginning.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+MERGE_BASE=$(git merge-base -a develop "$(git rev-parse --abbrev-ref HEAD)" develop)
+GO_COMMIT_MERGE_BASE=$(git show "$MERGE_BASE":status-go-version.json | jq -r '."commit-sha1"')
+GO_COMMIT_CURRENT=$(cat status-go-version.json | jq -r '."commit-sha1"')
+
+GITHUB_LINK="https://github.com/status-im/status-go/compare/$GO_COMMIT_MERGE_BASE...$GO_COMMIT_CURRENT"
+
+echo "$GITHUB_LINK"
+if [ -f "$COMMIT_MSG_FILE" ]; then
```
jakubgs: If you're using bash you should consistently use double square brackets.

jakubgs: Also, when referencing variables it's better to use the curly brackets format too.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [[ -z $(git diff --cached --name-only -r | grep status-go-version.json) ]]; then
+  exit 0 # No status-go-version.json changes found.
+fi
+
+MERGE_BASE=$(git merge-base -a develop "$(git rev-parse --abbrev-ref HEAD)" develop)
+GO_COMMIT_MERGE_BASE=$(git show "$MERGE_BASE":status-go-version.json | jq -r '."commit-sha1"')
```
jakubgs: Missing curly braces for variable.

Makefile
```diff
@@ -120,6 +120,12 @@ _tmpdir-mk: ##@prepare Create a TMPDIR for temporary files
 # Make sure TMPDIR exists every time make is called
 -include _tmpdir-mk
 
+_install-hooks: SHELL := /bin/sh
+_install-hooks: ##@prepare Create prepare-commit-msg git hook symlink
+	@ln -s ../../scripts/hooks/prepare-commit-msg .git/hooks
+# Create prepare-commit-msg git hook symlink
```
jakubgs: We don't need this duplicate comment, it's probably just some leftover above. What's important is the one above the target prefixed with `##@` which shows up in help normally.

Also, this target should probably be added to `.PHONY`.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,23 @@
...
+
+#todo remove this echo
+echo "$GITHUB_LINK"
+
+COMMIT_MSG_FILE=$1
+if [[ -f "$COMMIT_MSG_FILE" ]]; then
+  #todo edit link if exists
+  COMMIT_MSG_FILE_CONTENT=$(cat "$COMMIT_MSG_FILE")
+  echo "$COMMIT_MSG_FILE_CONTENT"
+  echo "${GITHUB_LINK}" >>"${COMMIT_MSG_FILE}"
```
jakubgs: Something like this might work:
```sh
#!/usr/bin/env bash
sed -i -n -e '/^TEST=/!p' -e '2 iTEST=text'$"\n" $1
```
Which gives me:

image

On first commit. You'll just have to adjust it to use part of github URL as prefix.

jakubgs: Also, I don't think the first argument with the message path is always there, so no need to check.

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+  while IFS= read -r line; do
+    if [[ "$line" == *"https://github.com/status-im/status-go/compare"* ]]; then
+      echo "${GITHUB_LINK}" >>temp_file
+    else
+      echo "${line}" >>temp_file
+    fi
+  done <"${COMMIT_MSG_FILE}"
+
+  if ! grep -q "https://github.com/status-im/status-go/compare" temp_file; then
+    echo "${GITHUB_LINK}" >>temp_file
```
jakubgs, line 29: This will not work, it adds the link after the commit contents, which makes it a separate commit:

image

scripts/hooks/prepare-commit-msg
```diff
@@ -0,0 +1,33 @@
...
+GO_COMMIT_MERGE_BASE=$(git show "${MERGE_BASE}":status-go-version.json | jq -r '."commit-sha1"')
+GO_COMMIT_CURRENT=$(jq -r '."commit-sha1"' status-go-version.json)
+
+GITHUB_LINK="https://github.com/status-im/status-go/compare/${GO_COMMIT_MERGE_BASE}...${GO_COMMIT_CURRENT}"
+
+COMMIT_MSG_FILE=$1
+#todo for now instead of sed I use this just to have it all working. Later will implement sed solution
+if [[ -f "${COMMIT_MSG_FILE}" ]]; then
+  if [ -f temp_file ] ; then
+          rm temp_file
```
jakubgs, line 18: This is now how you create a temporary file if you need one. Read about `mktemp`. This creates a file in the repo.

PR #13145, opened 2022-03-02
Fix ios simulator error when notifications enabled

Parveshdhull:
## Summary:
PR fixes ios simulator error when enabling notifications

### Error cause:
Remote notifications are not supported in the ios simulator and that's why when the simulator tried to register for push notifications, it fired event `:notifications/switch-error`, which called `multiaccounts.update/multiaccount-update` function with one less variable (cofx) than expected and caused the error

Thank you very much @msuess and @Samyoul for finding this issue.

status: ready

### DIFF
--- a/src/status_im/notifications/core.cljs
+++ b/src/status_im/notifications/core.cljs
@@ -143,8 +143,7 @@
   {:events [:notifications/switch-error]}
   [cofx enabled?]
   (multiaccounts.update/multiaccount-update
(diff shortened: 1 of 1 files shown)


rasom (approved):

qoqobolo:
@Parveshdhull thanks for the fix!
E2e failures are not PR-related. 
PR is smoke-tested and can be merged.

PR #13152, opened 2022-03-07
e2e: tx users

churik:


churik:
@jakubgs can you please review when have a minute? 
Thanks!

jakubgs (approved):
Not optimal, but works. Thanks!

PR #13157, opened 2022-03-07
fix screen opens twice on app reopening in ios

### LINKED ISSUE #13153: Screens open twice if put the app in the background and re-open it (iOS)
# Bug Report

## Problem 
[comment]: # (A problem description)

#### Expected behavior
If you put the app in the background and open it again, the screens will open twice and the back button will also appear on the screen.

#### Actual behavior

https://user-images.githubusercontent.com/67952253/157040214-e8d10021-d205-47c7-84a5-5176d581557d.MP4

### Reproduction

1. Open Status on iOS device
2. Put the app in the background
3. Re-open the app

### Additional Information

- Status version: nightly 7/03
- Operating System: iOS

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/13153

### Summary (Fixes):
- [Appearance module](https://reactnative.dev/docs/appearance) of react-native has a [bug](https://github.com/facebook/react-native/issues/28525#issuecomment-666244448), it fires system theme change events `(with wrong value)` when the ios app comes back from the background. So now instead of relying on the value provided by the event, we are requesting it manually.
- Back button was shown because we were calling `:navigate-to` event, even for the views on the home screen. `(for some reason, this also affected ios only)`

### QA test request:
Thanks @qoqobolo for finding this great issue. Fix might have affected system theme changes for android, Please let me know if it is not working as expected

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -38,7 +38,7 @@
             [status-im.navigation :as navigation]
             status-im.notifications-center.core
             status-im.pairing.core
-            status-im.popover.core
+            [status-im.popover.core :as popover]
             status-im.profile.core
             status-im.search.core
             status-im.signals.core
@@ -52,7 +52,7 @@
             status-im.utils.logging.core
             [status-im.utils.universal-links.core :as universal-links]
             [status-im.utils.utils :as utils]
-            status-im.visibility-status-popover.core
+            [status-im.visibility-status-popover.core :as visibility-status-popover]
             status-im.visibility-status-updates.core
             status-im.waku.core
             status-im.wallet.accounts.core
@@ -60,7 +60,8 @@
             [status-im.wallet.core :as wallet]
             status-im.wallet.custom-tokens.core
             [status-im.navigation.core :as navigation.core]
-            [status-im.multiaccounts.login.core :as login.core]))
+            [status-im.multiaccounts.login.core :as login.core]
+            [status-im.signing.core :as signing]))
 
 (re-frame/reg-fx
  :dismiss-keyboard
@@ -148,7 +149,14 @@
                  {::multiaccounts/switch-theme (if (= :dark theme) 2 1)}
                  (when (seq dispatch-later)
                    {:utils/dispatch-later dispatch-later}))
-                (bottom-sheet/hide-bottom-sheet)
+                (when (get-in db [:bottom-sheet/show?])
+                  (bottom-sheet/hide-bottom-sheet))
+                (when (get-in db [:popover/popover])
+                  (popover/hide-popover))
+                (when (get-in db [:visibility-status-popover/popover])
+                  (visibility-status-popover/hide-visibility-status-popover))
+                (when (get-in db [:signing/tx])
+                  (signing/discard))
                 (navigation/init-root root-id)
                 (when (= root-id :chat-stack)
                   (navigation/change-tab current-tab))))))
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -115,30 +115,39 @@
 (fx/defn system-theme-mode-changed
   {:events [:system-theme-mode-changed]}
   [{:keys [db] :as cofx} theme]
-  (let [cur-theme     (get-in db [:multiaccount :appearance])
-        current-tab   (get db :current-tab :chat)
-        view-id       (:view-id db)
-        screen-params (get-in db [:navigation/screen-params view-id])
-        root-id       @navigation.core/root-id]
+  (let [cur-theme      (get-in db [:multiaccount :appearance])
+        current-tab    (get db :current-tab :chat)
+        view-id        (:view-id db)
+        screen-params  (get-in db [:navigation/screen-params view-id])
+        root-id        @navigation.core/root-id
+        dispatch-later (cond-> []
+                         (= view-id :chat)
+                         (conj {:ms       1000
+                                :dispatch [:chat.ui/navigate-to-chat (:current-chat-id db)]})
+
+                         (and
+                          (= root-id :chat-stack)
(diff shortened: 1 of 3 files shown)


src/status_im/navigation/core.cljs
```diff
@@ -390,7 +390,9 @@
                          :create-community-channel
                          :community-emoji-thumbnail-picker
                          :create-community-category
-                         :community-edit-chats))
+                         :community-edit-chats
+                         :community-edit
```
flexsurfer, line 394: how is this change related to this PR ?

src/status_im/events.cljs
```diff
@@ -115,30 +115,39 @@
 (fx/defn system-theme-mode-changed
   {:events [:system-theme-mode-changed]}
```
flexsurfer, line 117: could you please elaborate how reopening the app related to system-theme-mode-changed ?

Parveshdhull: Hi @flexsurfer,
when the app reopened Appearance module fired the theme change event **twice** (even though the theme was the same), due to a [bug](https://github.com/facebook/react-native/issues/28525) in the react-native library.
So when the system theme change event fired, it called the `system-theme-mode-changed` event two times.
And in issue, we also noticed there was a back button on the home screen view when this happened. It happened because we were calling the `navigate-to` event also for view on the home screen.
We never observed this behavior before this module bug, Because in android calling `navigate-to` for view on the home screen had no effect, and in ios also it is showing the back button only after the bug is firing the theme change event twice.
So, in this change, we also make sure even if for some reason event fires twice, we should not navigate to the screens, which are already on the home screen.

src/status_im/events.cljs
```diff
@@ -115,30 +115,39 @@
...
+                         (and
+                          (= root-id :chat-stack)
+                          (not-any? #(= view-id %) '(:home :empty-tab :wallet :status :my-profile :chat)))
+                         (conj {:ms       1000
+                                :dispatch [:navigate-to view-id screen-params]})
+
+                         (some #(= view-id %) navigation.core/community-screens)
+                         (conj {:ms 800 :dispatch
+                                [:navigate-to :community
+                                 (get-in db [:navigation/screen-params :community])]})
```
Parveshdhull, line 138: When the system theme changes, we are restoring the app state. If earlier view-id is related to any community-related screen, we are making sure to first open community before opening that screen. The above [review](https://github.com/status-im/status-react/pull/13157#discussion_r821427976) screens are edge cases, (the user was editing the community when the system theme changed), but we can include these two also to cover those edge cases? (change seems too small to do in separate pr). wdyt?

src/status_im/navigation/core.cljs
Parveshdhull: Hi @flexsurfer, Thanks for reviewing PR
Please check this comment.
https://github.com/status-im/status-react/pull/13157#discussion_r821511220

src/status_im/events.cljs
flexsurfer: but i'm wondering why we should handle this event at all if the theme is not changed?

flexsurfer (approved):

qoqobolo:
@Parveshdhull oh, to be fair @churik found this, I just logged :)
Took for testing.

qoqobolo:
@Parveshdhull thanks for the fix, the main issue is no longer reproduced.
But noticed other issues that are not presented in 1.18, but they are also in nightly. Could you take a look and tell if they can be fixed in this PR, please 

### ISSUE 1: Some bottom sheets close when changing system mode
E.g. bottom sheet on the home screen when pressing the `+` button or the `Custom fee` bottom sheet in the Wallet. In the last case, some entered values may be lost.
Also for my Android, this case is relevant when the battery saver mode is activated.

Is this behaviour expected?  

**Steps:**
1. Device is in light mode
2. Open Status
3. Wallet > Send tx > Enter values > Next > Set custom fee
4. Activate dark mode on device

**OS:** iOS, Android

https://user-images.githubusercontent.com/67952253/157274318-e80a27d6-474f-4f91-8bc3-24bf0b4b6704.MP4

https://user-images.githubusercontent.com/67952253/157274259-64715818-3384-43d5-b251-75880bfc1a38.MP4

### ISSUE 2: The **Sending... ETH** header is not visible when switching mode

**OS:** iOS, Android

Screenshot 2022-03-08 at 16 58 34

### ISSUE 3: White/black lines in the` Send tx` bottom sheet when switching mode
They are visible on the screenshots above.

This looks much better than in 1.18 :) But still wanted to ask if we can fix this if it's easy and fast?

Parveshdhull:
Hi @qoqobolo, Thank you very much for testing the PR. 
This is somewhat a summary of the cause and solution status of the above issues.
- When the system or app theme changes, we have to destroy all the views and recreate for the apply theme. If we apply without doing this, we get weird color behaviors like Issue 2, Issue 3 or [12921](https://github.com/status-im/status-react/issues/12921)
- So after destroying views, the app comes back to its original state (home screen), which we get after login
- It is relatively easier to again restore screens than overlays and bottom sheets. We can try restoring popups, bottom sheets, etc. too, but it seems complicated and feels unsafe to auto-populate, especially sheets like transaction sheets, etc.

So, for now, I guess we have to hide/destroy sheets on theme change.
Currently, we are hiding all overlays and sheets on theme change. Apparently `custom theme` sheet didn't hide during theme change (thanks for finding this), because it is using `:signing-sheet` instead of `:bottom-sheet` (added now).
So hopefully, all issues are fixed now, except the sheets will be closed after the theme change.
I think battery saver is toggling dark mode and that's why it changes theme of app and closes sheet.

Please let me know if it works for you. Thanks

qoqobolo:
@Parveshdhull got it! Thanks for the fixes and the great explanation.
PR can be merged.

PR #13163, opened 2022-03-11
Tests refactoring. Created TestMessages group. Added first 6 tests

pavloburykh:
Tests refactoring. Created TestMessages group. Added first 6 tests


test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+        cls.drivers, cls.loop = create_shared_drivers(2)
+        cls.device_1, cls.device_2 = SignInView(cls.drivers[0]), SignInView(cls.drivers[1])
+        cls.home_1, cls.home_2 = cls.device_1.create_user(enable_notifications=True), cls.device_2.create_user()
+        cls.public_key_1, cls.default_username_1 = cls.home_1.get_public_key_and_username(return_username=True)
+        cls.public_key_2, cls.default_username_2 = cls.home_2.get_public_key_and_username(return_username=True)
+        profile_2 = cls.home_2.get_profile_view()
+        profile_2.switch_network()
+        [home.home_button.click() for home in (cls.home_1, cls.home_2)]
+        cls.chat_2 = cls.home_2.add_contact(cls.public_key_1)
+        cls.chat_2.chat_message_input.send_keys('test message')
```
churik: send_message. What is the reason for sending this message?

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+        cls.group_chat_2 = cls.home_2.get_chat_view()
+        cls.group_chat_2.join_chat_button.click()
+        [home.home_button.click() for home in (cls.home_1, cls.home_2)]
+        cls.public_chat_name = cls.home_1.get_random_chat_name()
+        cls.public_chat_1, cls.public_chat_2 = cls.home_1.join_public_chat(cls.public_chat_name), cls.home_2.join_public_chat(
+            cls.public_chat_name)
+        [home.home_button.click() for home in (cls.home_1, cls.home_2)]
+        cls.home_1.get_chat(cls.default_username_2).click()
+        cls.home_2.get_chat(cls.default_username_1).click()
+        cls.message_1, cls.message_2, cls.message_3, cls.message_4 = "Message1", "Message2", "Message3", "Message4"
```
churik: IMO better to divide it into blocks (creating 1-1, public, group chat)

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+            'italic text in underscore': '_',
+            'inline code': '`',
+            'code blocks': '```',
+            'quote reply (one row)': '>',
+        }
+
+        for message, symbol in markdown.items():
+            self.home_1.just_fyi('checking that "%s" is applied (%s) in 1-1 chat' % (message, symbol))
+            message_to_send = symbol + message + symbol if 'quote' not in message else symbol + message
+            self.chat_2.chat_message_input.send_keys(message_to_send)
```
churik: can be self.chat_2.send_message

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+            self.chat_2.send_message_button.click()
+            if not self.chat_2.chat_element_by_text(message).is_element_displayed():
+                self.errors.append('%s is not displayed with markdown in 1-1 chat for the sender \n' % message)
+
+            if not self.chat_1.chat_element_by_text(message).is_element_displayed():
+                self.errors.append('%s is not displayed with markdown in 1-1 chat for the recipient \n' % message)
+
+        [chat.home_button.double_click() for chat in (self.chat_1, self.chat_2)]
+        chat_name = self.home_1.get_random_chat_name()
+        [home.join_public_chat(chat_name) for home in (self.home_1, self.home_2)]
```
churik: no reason to join to public chat 2d time, as already joined in setup_method

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+            self.errors.append("Counter of reaction is not set on Emoji for message receiver!")
+        self.chat_1.set_reaction(emoji_unicode, emoji_message=True)
+        if emoji_message.emojis_below_message(own=False) == 1:
+            self.errors.append("Counter of reaction is not re-set on Emoji for message receiver!")
+
+        self.errors.verify_no_errors()
+
+    @marks.testrail_id(702067)
+    @marks.medium
+    def test_messaging_open_user_profile_long_press_on_message_1_1_chat(self):
```
churik: not sure it is worth separate case;
can be added as a check to any 1-1 chat test and removed.

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+
+        self.home_2.just_fyi("1-1 chat: receiver verifies that can open sender profile on long tap on message")
+        self.chat_2.chat_element_by_text(message_from_sender).long_press_element()
+        if self.chat_2.view_profile_by_avatar_button.is_element_displayed():
+            self.errors.append('1-1 chat: another user profile is opened on long tap on received message')
+        self.chat_2.click_system_back_button(2)
+
+    @marks.testrail_id(702068)
+    @marks.medium
+    def test_messaging_open_user_profile_long_press_on_message_in_public_chat(self):
```
churik: not sure it is worth separate case;
can be added as a check to any public chat test and removed.

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+        # chat_public_1, chat_public_2 = home_1.get_chat_view(), home_2.get_chat_view()
+        self.public_chat_2.send_message(message_from_receiver)
+        self.public_chat_2.chat_element_by_text(message_from_receiver).long_press_element()
+        if self.public_chat_2.view_profile_by_avatar_button.is_element_displayed():
+            self.errors.append('Public chat: "view profile" is shown on long tap on sent message')
+        self.public_chat_1.view_profile_long_press(message_from_receiver)
+        if not self.public_chat_1.remove_from_contacts.is_element_displayed():
+            self.errors.append('Public chat: another user profile is not opened on long tap on received message')
+
+        self.errors.verify_no_errors()
```
churik: it should be always last

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+
+        self.public_chat_1.get_back_to_home_view()
+        self.public_chat_2.click_system_back_button(2)
+
+        # [chat.home_button.double_click() for chat in [self.public_chat_1, self.public_chat_2]]
+        self.home_1.get_chat(self.default_username_2).click()
```
churik: navigation for next test must be in next test; so if you are not on the desired view when start `test_messaging_can_pin_messages_in_one_to_one_chat` - better to navigate home > navigate to chat view

test/appium/tests/atomic/chats/test_public.py
```diff
@@ -328,6 +329,304 @@ def test_public_chat_open_using_deep_link(self):
...
+        self.home_2.get_chat(self.default_username_1).click()
+
+    @marks.testrail_id(702069)
+    @marks.medium
+    def test_messaging_can_pin_messages_in_one_to_one_chat(self):
+        self.home_1.just_fyi("Check that Device1 can pin own message in 1-1 chat")
+        self.chat_1.send_message(self.message_1)
```
churik: ca add that to setup_method instead of sending 'hey'

churik (changes requested):

test/appium/tests/atomic/chats/test_public.py
pavloburykh: This piece of code was copied from original test case @marks.testrail_id(5425). The reason for sending "test message" in @marks.testrail_id(5425) was because chat_1 = home_1.get_chat(default_username_2).click() Without sending some message, chat would not appear in activity center of device 1 and therefore it would be impossible to use get_chat() method to open the chat. 

Anyway, in setup class I use chat_1 = home_1.add_contact(public_key_2) just like in case of chat_2 = chat_2 = home_2.add_contact(public_key_1) So I will remove this unnecessary piece of code with sending "test message"

test/appium/tests/atomic/chats/test_public.py
pavloburykh: fixed

test/appium/tests/atomic/chats/test_public.py
pavloburykh: fixed

test/appium/tests/atomic/chats/test_public.py
pavloburykh: @churik can't find the part of code sending 'hey' message in TestMessaging group of cases. Could you please specify what do you mean.

test/appium/tests/atomic/chats/test_public.py
pavloburykh: fixed

test/appium/tests/atomic/chats/test_public.py
pavloburykh: I have a question here: if we run test_messaging_can_pin_messages_in_one_to_one_chat separately, then it does not make sense to navigate to chat view, as we already navigated there after setup class. 

Navigating to chat view should be performed only in case if test_messaging_can_pin_messages_in_one_to_one_chat is running after test_messaging_open_user_profile_long_press_on_message_in_public_chat that's why I decided to set navigation for next test in previous one. 

Could you please explain, what is the reason of why we need to keep navigation for the next test exactly in next test?

test/appium/tests/atomic/chats/test_public.py
pavloburykh: changed to [home.get_chat(self.public_chat_name) for home in (self.home_1, self.home_2)]

test/appium/tests/atomic/chats/test_public.py
pavloburykh: to say in truth, I would prefer to leave it as separate case than moving it as check to some other irrelevant 1-1 test. IMHO, adding this test as additional check in other test case goes against our goal of making more simple and atomic cases.  But if you insist, I can move this check to @marks.testrail_id(695843) test_edit_delete_message_in_one_to_one_. What do you think?

test/appium/tests/atomic/chats/test_public.py
pavloburykh: to say in truth, I would prefer to leave it as separate case than moving it as check to some other irrelevant public test. IMHO, adding this test as additional check in other test case goes against our goal of making more simple and atomic cases.  But if you insist, I can move this check to @marks.testrail_id(700734) test_public_message_edit. What do you think?

test/appium/tests/atomic/chats/test_public.py
churik: sorry, meant "test message".
In fact instead of sending one more message you can work with already sent (if it was sent in setup method)

test/appium/tests/atomic/chats/test_public.py
churik: you can send a message, but better to work with them afterward, as I mentioned https://github.com/status-im/status-react/pull/13163/files#r825764748

test/appium/tests/atomic/chats/test_public.py
churik: to see this in logs, better to add cls.home_2.just_fyi; in that case you'll see clearly if smth go wrong

test/appium/tests/atomic/chats/test_public.py
churik: Yes, sure.
Let's say the previous test failed somewhere in the middle.
From that point, this particular test will be failed just because of the reason that previous test couldn't finish

test/appium/tests/atomic/chats/test_public.py
pavloburykh: decided to remove this check at all

test/appium/tests/atomic/chats/test_public.py
pavloburykh: decided to remove this check at all

test/appium/tests/atomic/chats/test_public.py
pavloburykh: yeah, that is exactly what I did

churik (approved):

PR #13167, opened 2022-03-13
Quo2, Switcher Navigation

Parveshdhull:
### Summary
PR implements bottom navigation bar and basic navigation for switcher implementation.
Implementation is rough draft to test how switcher can be part of navigation

> Optimized for Android only

status: wip


flexsurfer:
hey @Parveshdhull great start :) i think that switcher shouldn't be a part of quo2, because its not really a component which can be reused somehow, so I would implement it in `status-im/switcher` package, and all navigation in `status-im/navigation2` instead, wdyt ?

Parveshdhull:
> hey @Parveshdhull great start :) i think that switcher shouldn't be a part of quo2, because its not really a component which can be reused somehow, so I would implement it in `status-im/switcher` package, and all navigation in `status-im/navigation2` instead, wdyt ?

Yes, that sounds better. I will move them 👍 Thank you

Parveshdhull:
PR is still in WIP, I wanted to merge it for now to avoid complex rebase

Parveshdhull:
Most of the code consists only in Switcher and Navigation2 directory, so that it will not effect current develop

flexsurfer:
regarding icons names, im not sure how to it better, i understand we will have two versions of icons, probably instead of `-new` we could use prefix `redesign-` so we can easily find these icons they will be sorted by this prefix wdyt ?

src/quo/react_native.cljs
```diff
@@ -102,6 +106,9 @@
                                                               :property (:opacity layout-animation-properties)}
                                                :delete   #js {:type     (:ease-in-ease-out layout-animation-types)
                                                               :property (:opacity layout-animation-properties)}}})
+(defonce enable-layout-animations
```
flexsurfer, line 110: add space between lines

src/status_im/switcher/constants.cljs
```diff
@@ -0,0 +1,34 @@
+(ns status-im.switcher.constants
+  (:require [status-im.utils.handlers :refer [<sub]]
+            [status-im.utils.platform :as platform]))
+
+(def switcher-button-radius 24)
+
+(def switcher-button-size
+  (* switcher-button-radius 2))
+
+(def platform (if platform/android? :android :ios))
```
flexsurfer: why is it needed ?

flexsurfer (approved):

churik:
Please, ping me in case of any possible regression, thanks!

PR #13170, opened 2022-03-14
Only create git hook whet it is absent

audriu:
### Summary

Fix Makefile so it will not show this error "ln: failed to create symbolic link '.git/hooks/prepare-commit-msg': File exists" every time make target run.

#### Platforms
- command line build tools

##### Non-functional
- build scripts less noisy

### Steps to test
- run any make target
- see there is no error that symlink exists

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ _tmpdir-rm: ##@prepare Remove TMPDIR
 
 _install-hooks: SHELL := /bin/sh
 _install-hooks: ##@prepare Create prepare-commit-msg git hook symlink
-	@ln -s ../../scripts/hooks/prepare-commit-msg .git/hooks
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -126,7 +126,9 @@ _tmpdir-rm: ##@prepare Remove TMPDIR
 
 _install-hooks: SHELL := /bin/sh
 _install-hooks: ##@prepare Create prepare-commit-msg git hook symlink
+ifeq (,$(wildcard .git/hooks/prepare-commit-msg))
 	@ln -s ../../scripts/hooks/prepare-commit-msg .git/hooks
```
jakubgs: You can achieve the same thing with less code and less file operation by just adding `-f`.

jakubgs (approved):

PR #13174, opened 2022-03-14
Fix broken username and timestamp layout for long usernames

### LINKED ISSUE #11543: Timestamp is shifted off the screen in the Timeline on Android if user has long 3-random name or nickname
# Bug Report

## Problem

If the User has a long 3-random name that takes two rows on the Status screen or max length nickname then the timestamp will be shifted off the screen.

#### Expected behavior

Timestamps are aligned and displayed in full on screen with long 3-random and nicknames

#### Actual behavior

Timestamps are shifted off the screen

Screenshot 2020-12-14 at 12 21 50
Screenshot 2020-12-14 at 12 25 45

Screenshot 2020-12-14 at 12 41 16

### Reproduction

- Generate new multiaccount with long 3-random name (in the list of names it should be displayed in 2 rows)
- Navigate to the Status tab
- Post a new status
- Add to contacts another user who has some statuses in his timeline
- Set max length nickname
- Check Timeline

### Additional Information

- Status version: Nightly 14/12
- Operating System: Android

siddarthkay:
[comment]: Fixes #11543 

#### Platforms
- Android
- iOS

status: ready

### DIFF
--- a/src/status_im/ui/screens/status/views.cljs
+++ b/src/status_im/ui/screens/status/views.cljs
@@ -91,6 +91,7 @@
 (defn message-item [account profile]
   (fn [{:keys [content-type content from timestamp outgoing] :as message}
        {:keys [modal on-long-press close-modal]}]
+
     [react/view (merge {:padding-vertical   8
                         :flex-direction     :row
                         :background-color   (when modal colors/white)
@@ -107,15 +108,18 @@
          [photos/member-photo from])]]
      [react/view {:flex 1}
       [react/view {:flex-direction  :row
-                   :justify-content :space-between}
+                   :justify-content :space-between
+                   :width           :98%
+                   :word-break      :break-all}
        [react/touchable-highlight
         {:on-press #(do (when modal (close-modal))
                         (when profile (re-frame/dispatch [:navigate-back]))
                         (re-frame/dispatch [:chat.ui/show-profile from]))}
-        (if outgoing
-          [message/message-my-name {:profile? true :you? false}]
-          [message/message-author-name from {:profile? true}])]
-       [react/text {:style {:font-size 10 :color colors/gray}}
+        (let [message-author-width (* @(re-frame/subscribe [:dimensions/window-width]) 0.75)]
+          (if outgoing
+            [react/view {:style {:width message-author-width}} [message/message-my-name {:profile? true :you? false}]]
+            [react/view {:style {:width message-author-width}} [message/message-author-name from {:profile? true}]]))]
+       [react/text {:style {:font-size 10 :color colors/gray :margin-left :auto}}
         (datetime/time-ago (datetime/to-date timestamp))]]
       [react/view
        (if (= content-type constants/content-type-image)


src/status_im/ui/screens/status/views.cljs
```diff
@@ -107,15 +107,17 @@
          [photos/member-photo from])]]
      [react/view {:flex 1}
       [react/view {:flex-direction  :row
-                   :justify-content :space-between}
+                   :justify-content :space-between
+                   :width           :98%
```
flexsurfer: could you please elaborate on this change ? does that mean all items will be 2% shorter ?

flexsurfer: what of there will be case where we need 4% ?

siddarthkay: Hi @flexsurfer : In some android layouts as seen in https://github.com/status-im/status-react/issues/11543 The timestamps would go out of the visible screen, hence I decided to restrict the width of the parent view to 98%

siddarthkay:
This is how it looks
IMG-8981

src/status_im/ui/screens/status/views.cljs
flexsurfer: could you elaborate please why 98% ?

flexsurfer: hi :) and thank you for the contribution ;)

siddarthkay: Sure, the logic behind 98% of available width is that we leave out 2% space as a margin on the right hand side and we only occupy 98% of available space to display the username and timestamp, this ensures that an overflow does not result in data going out of the screen( as observed in https://github.com/status-im/status-react/issues/11543 ) 

If you look at the screenshot I have share in the comments below it would show how the UI now looks like.

flexsurfer: but that means that now timestamp is shifted and has a bigger margin, and I'm still not sure if all cases are covered or there still will be the case when data goes out of the screen, what i mean is that this fix is for this specific case, and is not a general solution, and also timestamp now is not on its place its shifted on 2% to the left

siddarthkay: @flexsurfer : Thanks for taking the time to let me know your thoughts on this!

setting a fixed width to the parent view combined with a word-break property should ensure that textual content within that view stays within the defined limits.
Reasoning behind setting a 98% width was to ensure that most android devices have different viewport sizes and we could account for that. 

I have highlighted the entire width of the React View element in discussion here to show you how much space it takes up on the screen.

IMG_8982

This bug  [also ](https://github.com/status-im/status-react/issues/11543) appears only in the specific case when someone has set a long username and when a user sets a max length nickname for someone in their contacts.
After re-creating the issue I tested my fix on it and only then I have raised a PR for this issue.

If you have any different approach in mind or any scenario in which this would fail please let me know of that use case and I will plan my solution around it.

flexsurfer: thank you for the detailed explanation

flexsurfer (approved):

qoqobolo:
Hi @siddarthkay, thank you for the contribution and for the fix!

### ISSUE 1: Long nickname / 3-random name overlaps with timestamp
The timestamp is no longer shifted off the screen in case of a long username, but they are overlapped now. 
Can you fix it as well, please?

The issue is only reproducible on Android. 
My device: Huawei P40 Lite

photo_2022-03-15_11-20-20
photo_2022-03-15_11-20-15

siddarthkay:
@qoqobolo : Thanks for letting me know about your device configuration, I will attempt to reproduce and fix that as well

siddarthkay:
@qoqobolo : fixed this issue and validated it by hard coding a really long username to test, now no matter how long of a username, the app would dedicate only 75% of available screen width and keep the remaining space for timestamp and user profile icon.
Please test it on Huawei P40 Lite device and let me know :)

qoqobolo:
Thanks @siddarthkay, looks great in the timeline now!
However, I also noticed a similar issue in the Activity center on Android devices (reproducible on develop as well).

Is it okay for you to fix it in the same PR? If it's out of the scope of this PR, that's totally fine, I'll create a separate issue in that case, just let me know. 

**Steps:**
1. UserA with a long username creates a new 1-1 chat with UserB
2. UserB (Android) opens Home screen > navigates to the Activity center (the bell at the top right)

Screenshot 2022-03-16 at 12 42 43

qoqobolo:
It was decided in DM that the last issue will be logged separately. 

@siddarthkay thanks again for your work!
PR can be merged.

flexsurfer:
hey @siddarthkay could you please squash commits, thanks

siddarthkay:
@flexsurfer : Squashed

flexsurfer:
it has been merged, github is broken

PR #13180, opened 2022-03-19
Fix sed command to work on mac

### LINKED ISSUE #13090: Add changelog from status-go on upgrade

audriu:
fixes #13090
### Summary
Fixes for sed command in MacOS

#### Platforms
- macOS
- Linux

### Steps to test
- Run scripts/update-status-go.sh "tag" and commit the changes. The PR message contains link to the changelog.
- Also check a commit made using this hook https://github.com/status-im/status-react/commit/f9f78e23d31db7f5eb31595b0b8fe645d6017dc7 the commit message contains link to the status-go changelog/
- Need to make sure that it works on GNU/Linux and on MacOS.

status: ready

### DIFF
--- a/scripts/hooks/prepare-commit-msg
+++ b/scripts/hooks/prepare-commit-msg
@@ -14,9 +14,9 @@ GITHUB_LINK="${GITHUB_LINK_PREFIX}${GO_COMMIT_MERGE_BASE}...${GO_COMMIT_CURRENT}
 
 COMMIT_MSG_FILE=$1
 # Check if the commit message already contains the link (rebase) and update otherwise insert into line2
-# temporary comment out because of the error on macos
-#if ! grep -qF "${GITHUB_LINK_PREFIX}" "${COMMIT_MSG_FILE}" >/dev/null; then
-#  sed -in "2i${GITHUB_LINK}\n" "${COMMIT_MSG_FILE}"
-#else
-#  sed -in "s;^${GITHUB_LINK_PREFIX}.*$;${GITHUB_LINK};" "${COMMIT_MSG_FILE}"
-#fi
+if ! grep -qF "${GITHUB_LINK_PREFIX}" "${COMMIT_MSG_FILE}" >/dev/null; then
+  sed -in'' -e "2i\\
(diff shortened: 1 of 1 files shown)


scripts/hooks/prepare-commit-msg
```diff
@@ -14,9 +14,8 @@ GITHUB_LINK="${GITHUB_LINK_PREFIX}${GO_COMMIT_MERGE_BASE}...${GO_COMMIT_CURRENT}
...
 COMMIT_MSG_FILE=$1
 # Check if the commit message already contains the link (rebase) and update otherwise insert into line2
-# temporary comment out because of the error on macos
-#if ! grep -qF "${GITHUB_LINK_PREFIX}" "${COMMIT_MSG_FILE}" >/dev/null; then
-#  sed -in "2i${GITHUB_LINK}\n" "${COMMIT_MSG_FILE}"
-#else
-#  sed -in "s;^${GITHUB_LINK_PREFIX}.*$;${GITHUB_LINK};" "${COMMIT_MSG_FILE}"
-#fi
+if ! grep -qF "${GITHUB_LINK_PREFIX}" "${COMMIT_MSG_FILE}" >/dev/null; then
+  sed -in'' -e "2i${GITHUB_LINK}\n" "${COMMIT_MSG_FILE}"
```
jakubgs: For one, the `''`argument belongs with the `-i` flag, not the `-n` flag.

audriu: but @flexsurfer has tested without the "n" at all and it did not work.

jakubgs: Second thing is, it appears the same error can trigger for two different issues, which are present here, second one being:
https://stackoverflow.com/questions/25631989/sed-insert-line-command-osx

flexsurfer:
here is the command which works on macos https://github.com/status-im/status-react/pull/13186

flexsurfer (approved):

PR #13206, opened 2022-03-23
fix community preview

cammellos:
This commit makes community previews a bit more robust and fixes an issue with the signal not being sent. 
There's still some work to do to make it more reliable, and we should add caching, but that's outside the scope of this PR.
status:ready


Parveshdhull (approved):

flexsurfer (approved):

qoqobolo:
@cammellos thanks for the fixes. 
The remaining 3 e2e failures are not PR-related.

PR #13210, opened 2022-03-24
[#13181] Cannot delete images or audio messages

### LINKED ISSUE #13181: Cannot delete images or audio messages
Users cannot delete images or audio messages. Occasionally a delete button will appear, but I believe this is a bug and not intentional.

Tested on:
~ Android 11
~ Samsung Galaxy A505U
~ Version 1.18.1 - Latest nightly

flexsurfer:
fixes #13181


Parveshdhull (approved):

qoqobolo:
@flexsurfer thanks for adding this feature to images and audio!

### ISSUE 1: Message preview and unread counter are displayed on the recipient's side if the message was sent and deleted by the sender while the recipient was offline

The number of unread messages will increase with every deleted message.

**Steps:**
1. Device1 and device2 start a 1-1 chat
2. Device2 closes the app
3. Device1 sends a message and deletes it
4. Device2 logs in

**OS:** iOS, Android

https://user-images.githubusercontent.com/67952253/160586085-8f7b6520-400b-4c04-900c-ff3541d97368.MP4

flexsurfer:
thanks @qoqobolo just wondering how it's different from regular text messages ?

qoqobolo:
@flexsurfer hm weird, I tried a couple of times with text messages and it worked fine. But now I see that there is no difference between them. So thanks for your leading question, I've updated the issue.

flexsurfer:
@qoqobolo so this case is not related to this PR and should be registered separately?

qoqobolo:
@flexsurfer well, okay. I tried to reproduce it on develop several times before answering you, and I couldn't - previews didn't display for deleted messages. But I dunno how and why, but now I see the same behavior on develop too 🤦🏻‍♀️

Sorry for this, I'll try to figure out why I couldn't reproduce it on the first try and log it separately of course.

qoqobolo:
Thanks @flexsurfer!
The issue is logged separately, PR can be merged.

PR #13211, opened 2022-03-24
status timeline: set image max width to avoid overflow of horizontally long images

### LINKED ISSUE #13149: Timeline images stretch off screen
Images in the timeline stretch off the screen.
This bug started on version 1.18 and still exists on horizontal images.

Screenshot_20220305-102717_Status

siddarthkay:
[comment]: Fixes https://github.com/status-im/status-react/issues/13149

Platforms
Android
iOS
status: ready

### DIFF
--- a/src/status_im/ui/screens/status/views.cljs
+++ b/src/status_im/ui/screens/status/views.cljs
@@ -33,7 +33,9 @@
   (let [width (reagent/atom nil)]
     (fn [uri show-close?]
       [react/view {:style               {:width         @width
+                                         :align-items   :center
                                          :height        image-max-dimension
+                                         :max-width     :100%
                                          :overflow      :hidden
                                          :opacity       (if @width 1 0)
                                          :border-radius 16


src/status_im/ui/screens/status/views.cljs
```diff
@@ -33,7 +33,9 @@
   (let [width (reagent/atom nil)]
     (fn [uri show-close?]
       [react/view {:style               {:width         @width
+                                         :align-items   :center
                                          :height        image-max-dimension
+                                         :max-width     :100%
```
flexsurfer, line 38: could you please elaborate on how this will work?

siddarthkay: The modification here is to the container view which contains the image component.
problem : In case of horizontally long images the image width would be bigger than the device width and the image would overflow outside of the view.
Solution : set the container view to have a max-width of 100% which ensures that no matter what is the width of the image (which is in the child component) in question it would never overflow because of the max-width property.

Screenshot before fix :

 
sceenshot-before-fix



Screenshot after fix : 



screenshot-after-fix

flexsurfer: oh i see, but now the image is cut on the left?

siddarthkay: Its center aligned actually, Some images can be reallly long horizontally and to accommodate them there were 2 options : 
1) Either reduce the height to match the aspect ratio ( because of long length )
OR
2) Center align them with a fixed height
The second one seemed more readable to me given our constraints of device width

siddarthkay: Also when you tap on these images it opens them in a modal view in full

flexsurfer (approved):

pavloburykh:
@siddarthkay thank you for PR. Ready for merge.

flexsurfer:
@siddarthkay please squash commits before merge

siddarthkay:
@flexsurfer : yes I will take care from next time onwards, I messed up using the merge script with this PR :(

PR #13239, opened 2022-03-31
[WIP] New communities design

jo-mut:
New communities design


src/quo/components/header_redesign.cljs
```diff
@@ -0,0 +1,175 @@
+(ns quo.components.header-redesign
```
flexsurfer, line 1: all redesign should be in quo2

jo-mut:
okay

PR #13265, opened 2022-04-12
e2e: Updated GH and TR reports

### LINKED ISSUE #13255: e2e improvments
- [x] 1) If the test group is failed at the setup stage, currently it will not be stopped.

If that happens, the desired workflow must be:
- fail all tests in a group with the result "Test setup failed" 
- mark these tests in testreail
- stop sessions

- [x] 2) Add more details in Testrail and Github report 
If e2e belongs to the group, the report should contain:
- Link + id + class (currently only link + id) for each e2e
- in case there are some untested tests, the IDs for untested test cases for a quick re-run (like `Failed tests:1234,3456`)

-  [x] 3) Find out why some passed tests from group are not marked in TestRail

Example: https://app.saucelabs.com/builds/vdc/c91b9fa6b6d93b89baa2819a6e25e38c -> class TestPublicChatMultipleDeviceMerged is passed but not marked

yevh-berdnyk:
partly fixes #13255

### DIFF
--- a/ci/tests/Jenkinsfile.e2e-prs
+++ b/ci/tests/Jenkinsfile.e2e-prs
@@ -1,3 +1,5 @@
+library 'status-jenkins-lib@v1.4.0'
+
 pipeline {
 
   agent { label 'linux' }
@@ -101,7 +103,7 @@ pipeline {
                 -m \"${params.TEST_MARKERS}\" \
                 -k \"${params.KEYWORD_EXPRESSION}\" \
                 --apk=${params.APK_NAME} \
-                --build=PR-${params.PR_ID} \
+                --build=PR-${params.PR_ID}-${utils.timestamp()} \
                 --pr_number=${params.PR_ID} \
                 ${extraPytestOpts}
             """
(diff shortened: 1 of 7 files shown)


ci/tests/Jenkinsfile.e2e-prs
```diff
@@ -101,7 +103,7 @@ pipeline {
                 -m \"${params.TEST_MARKERS}\" \
                 -k \"${params.KEYWORD_EXPRESSION}\" \
                 --apk=${params.APK_NAME} \
-                --build=PR-${params.PR_ID} \
+                --build=PR-${params.PR_ID}-${nowDate} \
```
jakubgs: The [`utils`](https://github.com/status-im/status-jenkins-lib/blob/master/vars/utils.groovy) library in [status-jenkins-lib](https://github.com/status-im/status-jenkins-lib) already has a [`timestamp()`](https://github.com/status-im/status-jenkins-lib/blob/v1.4.0/vars/utils.groovy#L15-L19) function.

You should be able to just add this at the top of the file:
```groovy
library 'status-jenkins-lib@v1.4.0'
```
And then be able to use `utils.timestamp()`.

churik (approved):

PR #13273, opened 2022-04-12
Add google-free condition for android/build.gradle and fix indents

Parveshdhull:
### Summary
PR adds google-free conditions in `android/build.gradle` and fixes indents

status: ready

### DIFF
--- a/scripts/release-android.sh
+++ b/scripts/release-android.sh
@@ -61,6 +61,11 @@ if [[ -n "${READER_FEATURES}" ]]; then
     append_env_export 'READER_FEATURES'
 fi
 
+# Used by android/build.gradle to make dependencies/plugins optional
+if [[ -n "${GOOGLE_FREE}" ]]; then
+    append_env_export 'GOOGLE_FREE'
+fi
+
 # If no secrets were passed there's no need to pass the 'secretsFile'.
 if [[ -s "${SECRETS_FILE_PATH}" ]]; then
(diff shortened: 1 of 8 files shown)


jakubgs:
Is this not rebased? I'm getting typescript errors.

Parveshdhull:
> Is this not rebased? I'm getting typescript errors.

rebased now 👍

scripts/release-android.sh
```diff
@@ -61,6 +61,11 @@ if [[ -n "${READER_FEATURES}" ]]; then
     append_env_export 'READER_FEATURES'
 fi
 
+# Used by android/build.gradle to make dependencies/plugins optional
+if [[ -n "${GOOGLE_FREE}" ]]; then
+    append_env_export 'GOOGLE_FREE'
+fi
```
jakubgs, line 67: Yeah, this is a good catch. Since we add `GOOGLE_FREE=1` to `.env.release`, but that doesn't affect Gradle in Nix.

But I really need to get around to refactoring one of these days how these settings are passed via env. This is unnecessarily complex. It grew out of just a few variables, and it needs to be simplified.

jakubgs (approved):
Indeed it works. Thanks for fixing the tabs.

I'll do a separate PR to streamline passing of env vars to the Nix build later.

jakubgs:
Hold off on merging this, since the Android notifications situation is evolving.

PR #13277, opened 2022-04-14
Show notifications for successfully saved images

### LINKED ISSUE #13133: Show notifications for successfully saved images
As a user, I want to be notified when I successfully save an image message to files/gallery on the devices

Presently, when  the user saves an image, to verify that they have successfully saved the image to the phone storage they have to quit the app and confirm that the image is present in the gallery/files. We want when the user chooses to save an image they get a notification of a successful save

qfrank:
fixes #13133

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "react-native-draggable-flatlist": "^3.0.3",
     "react-native-fast-image": "^8.5.11",
     "react-native-fetch-polyfill": "^1.1.2",
+    "react-native-flash-message": "^0.2.1",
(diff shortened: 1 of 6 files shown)


qfrank:
feature preview

flexsurfer:
hey @qfrank thank you for the contribution and sorry for late response, its better to do not use external dependencies if its possible, and implement functionality in our code base, also this issue needs a designs first, so we need an input from design team , cc @hesterbruikman

qfrank:
wish there will be a design out soon :)

PR #13296, opened 2022-04-22
nix: build unsigned Android APK, sign separately

jakubgs:
This has several benefits:

* Less abuse of `extra-sandbox-paths` Nix option
* Less inputs to the Android release build derivation
* Easier for users to sign the build themselves
* Simplification of `scripts/release-android.sh`
* Preparation for building using Nix Flakes

The only two remaining credentials passed via `extra-sandbox-paths` is the Infura and OpenSea API keys, and there is no way around that other than passing them via Nix arguments, but that would cause them to end up in `/nix/store` as part of `.drv` files.

I'm also renaming `release-fdroid` to `build-fdroid` to be consistent.

Depends on: https://github.com/status-im/status-jenkins-lib/pull/42


jakubgs:
The APK signing process works, but I encountered an issue when installing:
```
 > adb install StatusIm-Mobile-220511-155039-5c0150-pr13296-universal.apk
Performing Streamed Install
adb: failed to install StatusIm-Mobile-220511-155039-5c0150-pr13296-universal.apk:
  Failure [
    -124: Failed parse during installPackageLI:
    Targeting R+ (version 30 and above) requires the resources.arsc of installed
    APKs to be stored uncompressed and aligned on a 4-byte boundary
  ]
```
Which apparently it's because the APK(which is a ZIP) needs to be "aligned":

>__Caution__: You must use zipalign at one of two specific points in the app-building process, depending on which app-signing tool you use:
>
>* If you use [apksigner](https://developer.android.com/studio/command-line/apksigner), zipalign must only be performed before the APK file has been signed. If you sign your APK using apksigner and make further changes to the APK, its signature is invalidated.
>* If you use [jarsigner](https://docs.oracle.com/javase/tutorial/deployment/jar/signing.html), zipalign must only be performed after the APK file has been signed.

https://developer.android.com/studio/command-line/zipalign

And since in the current implementation I'm using `jarsigner` I have to either also run `zipalign`, or switch to using `apksigner`.

jakubgs:
And indeed, signing with `apksigner` does work:
```
 > adb install ~/StatusIm-Mobile-220511-172946-15aef9-pr13296-universal.apk 
Performing Streamed Install
Success
```

Parveshdhull (approved):

Parveshdhull:
Hi @jakubgs , what will be the difference between the output of `build-android` and `release-fdroid`.

Parveshdhull:
And do we need `release-fdroid`?

jakubgs:
Just look at the targets:
https://github.com/status-im/status-react/blob/56cf2ae03d460d7402b7d05ff9d0b860a6135519/Makefile#L202-L208
The first target builds an unsigned APK, the second one signs it. You cannot install an unsigned APK on an Android phone.

jakubgs:
Ah, you're asking about the difference between release and fdroid build. Here it is:
https://github.com/status-im/status-react/blob/56cf2ae03d460d7402b7d05ff9d0b860a6135519/Makefile#L197-L201
https://github.com/status-im/status-react/blob/56cf2ae03d460d7402b7d05ff9d0b860a6135519/Makefile#L209-L213
There are two main differences:

* `BUILD_TYPE` - F-Droid build uses `.env.release` while `build-android` uses `.env.nightly`(not great naming, agreed).
* `ANDROID_ABI_INCLUDE` - F-Droid build includes `x86_64` because that's what they need for their emulator.

Actually, maybe we should change the name of `release-fdroid` to `build-fdroid` to be more consistent. Since it's not signed.

cammellos (approved):

jakubgs:
Merged as: https://github.com/status-im/status-react/commit/acfa73ab43b834fcc3dcd0c9037d0d652c6f9102

PR #13376, opened 2022-05-19
Add status-go tests

### LINKED ISSUE #13335: Create node bindings of status-go that can be run in tests
As a developer
I want to have node bindings of status-go that can be run in clojure tests
So that I can run tests validating clojure & status-go code together

### Description

In order to increase the test surface and make sure status-react works with a given version of status-go, we want to have some node bindings and be able to run them in clojure tests.
A very basic test would be to pull the right version of status-go when running `make test` and hit any of the c bindings function call.

Some work has already been done in the past https://github.com/status-im/status-nodejs , so this is technically possible.

cammellos:
Fixes: #13335 
status: ready


Makefile
```diff
@@ -286,6 +290,8 @@ lint-fix: ##@test Run code style checks and fix issues
 
 test: export TARGET := clojure
 test: ##@test Run tests once in NodeJS
+	nix/scripts/build.sh targets.status-go.library
```
jakubgs: Why do this if you can just make this target depend on the library target?
```suggestion
test: status-go-library ##@test Run tests once in NodeJS
	nix/scripts/build.sh targets.status-go.library
```

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
+{ lib, stdenv, utils, callPackage, buildGo116Package
+, go, xcodeWrapper
+# object with source attributes
+, meta, source
+, goBuildFlags ? [ ]
```
jakubgs: You are receiving and passing `goBuildFlags` but not using them, is that intentional or did you remove too much?

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+  # Ensure XCode is present for iOS build, instead of failing at the end of the build
+  preConfigure = optionalString isDarwin utils.enforceXCodeAvailable;
+
+  # Build mobile libraries
+  preBuild = let
+    NIX_GOWORKDIR = "$NIX_BUILD_TOP/go-build";
+  in ''
+    mkdir ${NIX_GOWORKDIR}
+
+    export GO111MODULE=off
```
jakubgs: I don't think we need this setting here. Actually, I don't think we need this whole `preBuild` phase.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+
+  # Build the Go library
+  buildPhase = let
+    ldFlags = [ "-extldflags=-Wl,--allow-multiple-definition" ]
+      ++ lib.optionals (secretsFile != "") ["-X node.OpenseaKeyFromEnv=$OPENSEA_API_KEY"]
+      ++ goBuildLdFlags;
+    CGO_LDFLAGS = concatStringsSep " " ldFlags;
+  in ''
+    runHook preBuild
+    runHook renameImports
```
jakubgs: What is `renameImports`? I don't see it anywhere.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+    ls $out/
+
+    rm -rf $NIX_GOWORKDIR
+    rm $out/main.go
+
+    runHook postBuild
+  '';
+
+  # replace hardcoded paths to go package in /nix/store, otherwise Nix will fail the build
+  fixupPhase = ''
```
jakubgs: It doesn't seem like `fixupPhase` is necessary, as there is no references to Go path in the library:
```
post-installation fixup
find: 'remove-references-to': No such file or directory
```
We can check that manually:
```
 > strings /nix/store/d4ys954p76y5cxy7nqii4g4rifl8icd4-status-go-develop-a244d77/libstatus.a | grep '/nix/store/[a-z0-9]+-go'

```

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+  in ''
+    runHook preBuild
+    runHook renameImports
+
+    mkdir -p $out
+    cd ${source.src} && go run ${source.src}/cmd/library/*.go > $out/main.go
+
+    go build -buildmode=c-archive -o $out/libstatus.a $out/main.go
+
+    ls $out/
```
jakubgs: Why is this here?

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+
+    rm -rf $NIX_GOWORKDIR
+    rm $out/main.go
```
jakubgs: Completely unnecessary. Nix builds in a sandbox, once build is over the sandbox is gone.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+    ldFlags = [ "-extldflags=-Wl,--allow-multiple-definition" ]
+      ++ lib.optionals (secretsFile != "") ["-X node.OpenseaKeyFromEnv=$OPENSEA_API_KEY"]
+      ++ goBuildLdFlags;
+    CGO_LDFLAGS = concatStringsSep " " ldFlags;
+  in ''
+    runHook preBuild
+    runHook renameImports
+
+    mkdir -p $out
+    cd ${source.src} && go run ${source.src}/cmd/library/*.go > $out/main.go
```
jakubgs: Why are you putting `main.go` into `$out`? We don't need it there. If it's part of the build it should be under `/build`.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+    export GO111MODULE=off
+    export NIX_GOWORKDIR=${NIX_GOWORKDIR}
+  '';
+
+  # Build the Go library
+  buildPhase = let
+    ldFlags = [ "-extldflags=-Wl,--allow-multiple-definition" ]
+      ++ lib.optionals (secretsFile != "") ["-X node.OpenseaKeyFromEnv=$OPENSEA_API_KEY"]
+      ++ goBuildLdFlags;
+    CGO_LDFLAGS = concatStringsSep " " ldFlags;
```
jakubgs: This has no effect. Variables set in a `let` block of a phase(or anything really) are available in the following block to be used, but __are not an env variables in the build__. To have that take effect you have to either:

* Define `CGO_LDFLAGS` in the main argument set for `buildGo116Package`.
* Use `export CGO_LDFLAGS=${CGO_LDFLAGS}` in the `buildPhase` or some other phase.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+  version = "${source.cleanVersion}-${source.shortRev}";
+
+  inherit meta;
+  inherit (source) src goPackagePath;
+
+  nativeBuildInputs = [ ]
+    ++ optional isDarwin xcodeWrapper;
+
+  # Fixes Cgo related build failures (see https://github.com/NixOS/nixpkgs/issues/25959 )
+  hardeningDisable = [ "fortify" ];
```
jakubgs: Pretty sure this is also unnecessary.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+let
+  inherit (stdenv) isDarwin;
+  inherit (lib)
+    concatStrings concatStringsSep concatMapStrings optionalString
+    getAttr attrValues makeBinPath optional;
+
+  removeReferences = [ go ];
+  removeExpr = refs: ''remove-references-to ${concatMapStrings (ref: " -t ${ref}") refs}'';
+
+in buildGo116Package {
```
jakubgs: We should probably switch to using `buildGo117Package`.

nix/status-go/library/default.nix
```diff
@@ -0,0 +1,87 @@
...
+{ secretsFile ? "" }:
+
+let
+  inherit (stdenv) isDarwin;
+  inherit (lib)
+    concatStrings concatStringsSep concatMapStrings optionalString
+    getAttr attrValues makeBinPath optional;
+
+  removeReferences = [ go ];
+  removeExpr = refs: ''remove-references-to ${concatMapStrings (ref: " -t ${ref}") refs}'';
```
jakubgs: If we don't run `fixupPhase` we don't need this either.

jakubgs (changes requested):
Good stuff man.

nix/status-go/library/default.nix
jakubgs: A proper way would look something like this:
```nix
{
  buildPhase = let
    ldFlags = [ "-extldflags=-Wl,--allow-multiple-definition" ]
      ++ lib.optionals (secretsFile != "") ["-X node.OpenseaKeyFromEnv=$OPENSEA_API_KEY"]
      ++ goBuildLdFlags;
  in ''
    runHook preBuild
    go build \
      -buildmode=c-archive \
      -ldflags="${concatStringsSep " " ldFlags}" \
      -o libstatus.a main.go
    runHook postBuild
  '';
}
```

nix/status-go/library/default.nix
jakubgs: Athough the more proper way is to use `NIX_BUILD_TOP` variable, as in :` > $NIX_BUILD_TOP/main.go`

jakubgs: Also, this part should probably be moved to a `preBuild` hook:
```nix
{
  preBuild = ''
    pushd go/src/$goPackagePath
    go run cmd/library/*.go > $NIX_BUILD_TOP/main.go
    popd
  '';
}
```

nix/status-go/library/default.nix
cammellos: Not using them, removing

jakubgs:
I verified on a CI host that `make status-go-library` builds fine on MacOS without `nativeBuildInputs` and `preConfigure`.

jakubgs (approved):
Cool stuff.

PR #13414, opened 2022-05-27
Use LoginWithConfig method to log in

### LINKED ISSUE #13314: Use LoginWithConfig method to log in
Instead of using the method:
`Login` https://github.com/status-im/status-go/blob/5f92e84c19c62234e006bddd3ced93e3613d80c4/mobile/status.go#L240 
when logging in in status-react, we should be using:

`LoginWithConfig` https://github.com/status-im/status-go/blob/5f92e84c19c62234e006bddd3ced93e3613d80c4/mobile/status.go#L252 
which allows us to override some parameters.

This will help with migrations so that anything that is passed from the frontend will have precedence over what's stored in the db. 

This will need to be tested to make sure migrations works fine and login data is preserved.

In status react, login logic should be in:

```
src/status_im/multiaccounts/login/core.cljs
```

qfrank:
fixes #13314

### DIFF
--- a/src/status_im/multiaccounts/login/core.cljs
+++ b/src/status_im/multiaccounts/login/core.cljs
@@ -63,7 +63,7 @@
 (re-frame/reg-fx
  ::login
  (fn [[key-uid account-data hashed-password]]
-   (status/login key-uid account-data hashed-password)))
+   (status/login-with-config key-uid account-data hashed-password nil)))
 
 (re-frame/reg-fx
  ::export-db
--- a/src/status_im/native_module/core.cljs
+++ b/src/status_im/native_module/core.cljs
@@ -64,6 +64,16 @@
    key-uid
    #(.login ^js (status) account-data hashed-password)))
 
+(defn login-with-config
+  "NOTE: beware, the password has to be sha3 hashed"
+  [key-uid account-data hashed-password config]
+  (log/debug "[native-module] loginWithConfig")
+  (clear-web-data)
+  (let [config (if config (types/clj->json config) "")]
+    (init-keystore
+     key-uid
+     #(.loginWithConfig ^js (status) account-data hashed-password config))))
+
 (defn export-db
   "NOTE: beware, the password has to be sha3 hashed"
   [key-uid account-data hashed-password callback]
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -371,6 +371,18 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
     }
 
     @ReactMethod
+    public void loginWithConfig(final String accountData, final String password, final String configJSON) {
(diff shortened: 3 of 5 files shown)


src/status_im/multiaccounts/login/core.cljs
```diff
@@ -63,7 +63,7 @@
 (re-frame/reg-fx
  ::login
  (fn [[key-uid account-data hashed-password]]
-   (status/login key-uid account-data hashed-password)))
+   (status/login-with-config key-uid account-data hashed-password nil)))
```
flexsurfer, line 66: we don't use config in this PR ?

flexsurfer (approved):

src/status_im/native_module/core.cljs
```diff
@@ -64,6 +64,19 @@
    key-uid
    #(.login ^js (status) account-data hashed-password)))
 
+(defn login-with-config
+  "NOTE: beware, the password has to be sha3 hashed"
+  [key-uid account-data hashed-password config]
+  (log/debug "[native-module] loginWithConfig")
+  (clear-web-data)
+  (if config
```
flexsurfer: it would be better to prepare config in let and do not duplicate init-keystore

src/status_im/multiaccounts/login/core.cljs
qfrank: i didn't find the config we need use, any suggestion ? @flexsurfer

pavloburykh:
@qfrank thanx for your contribution

Please, take a look at following issue

### ISSUE 1 Password migration fails on IOS (error: "file does not exist" after login)

**Steps:**

1. Install any other IOS PR build (for example [this one](https://i.diawi.com/XBpxxG)) that will be subsequently upgraded to current PR build
2. Login/register multiaccount 
3. Perform upgrade by installing IOS build from this PR 
4. Try to login multiaccaunt

**Actual result:** error "file does not exist appears"

**Expected result:** successful login to account 

Sorry for not attaching logs from IOS, as this functionality is currently broken in develop branch. 

photo_2022-05-27 17 34 58

https://user-images.githubusercontent.com/97245802/170720971-ab71deed-13c4-4b15-a124-0827ca9a7831.mp4

qfrank:
hi @pavloburykh , do u know the corresponding source commit id to this [PR](https://i.diawi.com/XBpxxG) ? i want to build it on my local machine to reproduce, thanks

pavloburykh:
Hi @qfrank!  We are not sure that we used a valid upgrade check in this case. So until we figure out a valid mechanism of how to test upgrade to PR builds from external contributors you can just ignore ISSUE 1. I'll keep you updated when we deal with this and verify if bug is valid.  Thanx.

pavloburykh:
@qfrank thanx for your contribution. Failed e2e are not connected to PR. So, ready for merge.

src/status_im/multiaccounts/login/core.cljs
flexsurfer: i think we should use same config no ? @cammellos

Issue #13416, opened 2022-05-27
Put an hard cap on images before sending


cammellos:
We should put an hard-cap of 300kb (to start with, need to be tested) for images to be sent, in order to avoid getting dropped because of TTL/POW or low upload speed.
The shrinking should be done in status-go.

https://github.com/status-im/status-react/issues/11450
https://github.com/status-im/status-react/issues/11733#issuecomment-1088710229

PR #13422, opened 2022-05-27
[Fixes: #13416] resize image in status-go as well, alert user if fails

cammellos:
Cap size of images, resize them in status-go, and refuses to send images > 500KB, in case we can't achieve the desired compression.

I could not replicate the issue using the current code, I will have to push out a special build disabling compression so you can see the error message.

### Testing

Since we can't replicate the issue, but we think the reason is that some images are not compressed enough, the best bet is to try to send very large images (15M/20M). 
If that does not work I can disable client-side compression, and you should be hitting the limit (status-go compression is more forgiving, also slower, so best to keep both).

### DIFF
--- a/src/status_im/chat/models/loading.cljs
+++ b/src/status_im/chat/models/loading.cljs
@@ -49,7 +49,7 @@
   {::json-rpc/call [{:method "wakuext_chat"
                      :params [chat-id]
                      :on-success #(re-frame/dispatch [:chats-list/load-chat-success %])
-                     :on-failure #(log/error "failed to fetch chats" 0 -1 %)}]})
+                     :on-error #(log/error "failed to fetch chats" 0 -1 %)}]})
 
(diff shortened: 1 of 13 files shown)


Parveshdhull (approved):

churik:
Tested on imported [DB with data](https://drive.google.com/file/d/1-3W6-3FuRXm6AP3-COzM-tbN2kZFTAlT/view?usp=sharing) 
- 5 public chats with 20k messages each
- one 1-1 chat with 20k messages
- 200 contacts


In general,  performance looks a way better compared to nightly (video attached)

https://user-images.githubusercontent.com/4557972/171441517-7dc2a99c-9eb8-441f-aeac-d539af366f5c.mp4


Images load faster.
Can't send image 50MB, the error is shown:
IMAGE 2022-06-01 17:25:57

Can load and send image of 30 MB fine.

Thank you for awesome work @cammellos

PR #13428, opened 2022-05-30
Remove explicit join from group chats

cammellos:
## Testing

e2e will likely need changing, as now chats are not explicitly joined by the users. cc @churik 

We should test compatibility between this version and release, as behind the scene a chat is autojoined when accepted from the activity center.

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/group.cljs
+++ b/src/status_im/ui/screens/chat/group.cljs
@@ -12,13 +12,6 @@
             [status-im.utils.platform :as platform])
   (:require-macros [status-im.utils.views :refer [defview letsubs]]))
 
-(defn join-chat-button [chat-id]
-  [quo/button
-   {:type                :secondary
-    :accessibility-label :join-chat-button
-    :on-press            #(debounce/dispatch-and-chill [:group-chats.ui/join-pressed chat-id] 2000)}
-   (i18n/label :t/join-group-chat)])
-
 (defn decline-chat [chat-id]
   [quo/button
    {:type                :secondary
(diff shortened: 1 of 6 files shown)


src/status_im/ui/screens/chat/group.cljs
```diff
@@ -69,16 +62,9 @@
 
 (defview group-chat-footer
   [chat-id invitation-admin]
-  (letsubs [{:keys [joined?]} [:group-chat/inviter-info chat-id]
-            removed? [:group-chat/removed-from-current-chat?]
-            invitations [:group-chat/invitations-by-chat-id chat-id]]
+  (letsubs [invitations [:group-chat/invitations-by-chat-id chat-id]]
     (if invitation-admin
```
Parveshdhull: no `else` part, can be replaced with `when`

Parveshdhull (approved):

pavloburykh:
@cammellos thanx for PR. Please take a look at following issue

### ISSUE 1 User does not receive invitation to group chat if he has left this group chat before

1. User 1 adds User 2 to the group chat
2. User 2 joins the chat
3. User 2 leaves the chat
4. User 1 adds User 2 to the group chat again

**Actual result:** for a second indicator appears upon AC icon and then disappears. Group chat does not appear in AC. User 2 has not possibility to re-join group chat. 

**Expected result:** new group chat appears in AC and user can re-join it


https://user-images.githubusercontent.com/97245802/171884716-322cace8-34cf-42fe-81ca-3a7c4280badc.mp4

[geth.log](https://github.com/status-im/status-react/files/8833377/geth.log)

PR #13450, opened 2022-06-03
Fixes #12192 -- Disable fetching more messages when status nodes are …

### LINKED ISSUE #12192: Can fetch messages in public chat despite disabled using of status nodes
# Bug Report

## Problem 
If disable Status nodes in sync settings and join some public chat it will still be possible to fetch messages from history. Re-login doesn't help.
Fetch more should be disabled in this case.

#### Expected behavior
Can't fetch messages from history if status nodes are disabled

#### Actual behavior
https://user-images.githubusercontent.com/67952253/120655538-0fd04780-c483-11eb-9c95-8dfebd219a92.mp4

### Reproduction

1. Profile > Sync settings > Status nodes > disable
2. Open Home screen
3. Join a public chat
4. Wait until the `Fetch more messages` button becomes active and fetch the history

### Additional Information

- Status version: 1.14

ibrkhalil:
…disabled

Fixes #12192

### Summary

Added a boolean to check if status nodes are enabled or not and if they're disabled we disable the functionality of fetch more messages button

The issue didn't state to hide the fetch more messages button or to disable it
So I disabled it.

#### Platforms to test
- iOS
- macOS
- Linux
- Windows

##### Functional

- 1-1 chats
- public chats
- group chats

### Steps to test
Go to settings -> sync settings and disable status nodes and then try to fetch more messages.
It won't work

### DIFF
--- a/src/status_im/chat/models.cljs
+++ b/src/status_im/chat/models.cljs
@@ -4,6 +4,7 @@
             [status-im.multiaccounts.model :as multiaccounts.model]
             [status-im.chat.models.message-list :as message-list]
             [status-im.data-store.chats :as chats-store]
+            [status-im.mailserver.core :as mailserver]
             [status-im.data-store.contacts :as contacts-store]
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.i18n.i18n :as i18n]
@@ -465,12 +466,14 @@
 (fx/defn chat-ui-fill-gaps
   {:events [:chat.ui/fill-gaps]}
   [{:keys [db] :as cofx} chat-id gap-ids]
-  (log/info "filling gaps" chat-id gap-ids)
-  (fx/merge cofx
-            {:db (assoc db :mailserver/fetching-gaps-in-progress gap-ids)}
-            (if (= gap-ids #{:first-gap})
-              (sync-chat-from-sync-from chat-id)
-              (fill-gaps chat-id gap-ids))))
+  (let [use-status-nodes? (mailserver/fetch-use-mailservers? {:db db})]
+    (log/info "filling gaps if use-status-nodes = true" chat-id gap-ids)
+    (when use-status-nodes?
+      (fx/merge cofx
+                {:db (assoc db :mailserver/fetching-gaps-in-progress gap-ids)}
+                (if (= gap-ids #{:first-gap})
+                  (sync-chat-from-sync-from chat-id)
+                  (fill-gaps chat-id gap-ids))))))
 
 (fx/defn chat-ui-remove-chat-pressed
   {:events [:chat.ui/remove-chat-pressed]}
--- a/src/status_im/ui/screens/chat/message/gap.cljs
+++ b/src/status_im/ui/screens/chat/message/gap.cljs
@@ -22,7 +22,7 @@
     (when (or (not first-gap?) public? community?)
       [react/view {:style (style/gap-container)}
        [react/touchable-highlight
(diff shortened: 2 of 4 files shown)


cammellos:
@vampirekiddo thanks for the PR, great work!

I have checked the UI and it seems they don't show as disabled when a the user is not using status nodes, although it does not fetch history anymore (which is the desired behavior)

Basically they should show the same as when you are offline (Fetch more buttons are disabled), would you mind taking a look?

The subscription I believe is using is https://github.com/status-im/status-react/blob/366f235e31fabd6c5869e67766a09acfcd1301ac/src/status_im/subs.cljs#L2494 , so that's one place where you might want to take a look.

Thank you

ibrkhalil:
> @vampirekiddo thanks for the PR, great work!
> 
> I have checked the UI and it seems they don't show as disabled when a the user is not using status nodes, although it does not fetch history anymore (which is the desired behavior)
> 
> Basically they should show the same as when you are offline (Fetch more buttons are disabled), would you mind taking a look?
> 
> The subscription I believe is using is
> 
> https://github.com/status-im/status-react/blob/366f235e31fabd6c5869e67766a09acfcd1301ac/src/status_im/subs.cljs#L2494
> 
> , so that's one place where you might want to take a look.
> Thank you

Thanks for the explanation Andrea!
I have added the disabled button styling now.

src/status_im/ui/screens/chat/message/gap.cljs
```diff
@@ -17,18 +17,19 @@
                                 gap-ids
                                 chat-id]
                   connected?   [:mailserver/connected?]
+                  use-status-nodes? [:mailserver/use-status-nodes?]
                   first-gap?   (= gap-ids #{:first-gap})]
     (when (or (not first-gap?) public? community?)
       [react/view {:style (style/gap-container)}
        [react/touchable-highlight
-        {:on-press (when (and connected? (not in-progress?))
+        {:on-press (when (and connected? (and (not in-progress?) use-status-nodes?))
```
cammellos: I think you can simplify, as `and` is a multi-variadic function, so you can do:

```
(and
  connected?
  (not in-progress?)
  use-status-nodes?)
```

Which should be equivalent to:

```
 (and
    connected?
    (and
       (not in-progress?)
       use-status-nodes?))
```

cammellos (approved):
Looks good, just a style issue, thank you!

src/status_im/ui/screens/chat/message/gap.cljs
ibrkhalil: Thanks Andrea
Removed the redundant `and` function

src/status_im/chat/models.cljs
```diff
@@ -465,12 +466,14 @@
 (fx/defn chat-ui-fill-gaps
   {:events [:chat.ui/fill-gaps]}
   [{:keys [db] :as cofx} chat-id gap-ids]
-  (log/info "filling gaps" chat-id gap-ids)
-  (fx/merge cofx
-            {:db (assoc db :mailserver/fetching-gaps-in-progress gap-ids)}
-            (if (= gap-ids #{:first-gap})
-              (sync-chat-from-sync-from chat-id)
-              (fill-gaps chat-id gap-ids))))
+  (let [use-status-nodes? (mailserver/fetch-use-mailservers? {:db db})]
```
flexsurfer, line 469: `when-let` could be used

flexsurfer (approved):

pavloburykh:
@vampirekiddo thanx for your contribution. Ready for merge.

PR #13456, opened 2022-06-07
refactor group chat e2e

### LINKED ISSUE #13449: Remove explicit join from group chats tests
# Problem

Due to implementation of #13428 explicit join should be removed from group chat tests

pavloburykh:
refactored group chat tests due to removal of explicit join and invitations

### DIFF
--- a/test/appium/views/chat_view.py
+++ b/test/appium/views/chat_view.py
@@ -1008,16 +1008,16 @@ class ChatView(BaseView):
         return '%s has invited %s' % (admin, invited_user)
 
     @staticmethod
+    def has_added_system_message(admin, invited_user):
+        return '%s has added %s' % (admin, invited_user)
+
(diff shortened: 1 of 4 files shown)


test/appium/views/chat_view.py
```diff
@@ -1015,6 +1019,10 @@ def invited_to_join_system_message(username, chat_name):
     def join_system_message(username):
         return '%s joined the group' % username
 
+    @staticmethod
+    def you_have_joined_system_message(chat_name, admin):
```
churik: Is it used?

pavloburykh: thanx. removed unused def you_have_joined_system_message(chat_name, admin) and def create_for_admin_system_message(chat_name) methods

churik (approved):

PR #13459, opened 2022-06-08
fix message sent state for audio and emoji messages

### LINKED ISSUE #13454: Audio message gets stuck in pending state until reopening chat view
# Bug Report

## Problem 
The delivery status of audio messages is stuck in a pending state (`...`) until reopening the chat (relevant for 1-1 chats, community channels, and group with members). After reopening, it's marked as `sent` or `delivered`, as expected. 
In the case of a group chat, the issue is only reproducible for groups with joined members, i.e. if you create a group chat with no members and send an audio message, it'll be marked as `sent`. 

#### Expected behavior
The delivery status is updated without reopening the chat

#### Actual behavior

https://user-images.githubusercontent.com/67952253/172349367-6db427f6-8ca9-47a5-b315-0c495d620c63.MP4

### Reproduction

1. Start a 1-1 chat
2. Send an audio message
3. Reopen the chat

### Additional Information

- Status version: nightly 7/06

[geth.log](https://github.com/status-im/status-react/files/8852118/geth.log)
[Status.log](https://github.com/status-im/status-react/files/8852119/Status.log)

Parveshdhull:
fixes https://github.com/status-im/status-react/issues/13454

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/message/message.cljs
+++ b/src/status_im/ui/screens/chat/message/message.cljs
@@ -520,40 +520,41 @@
       [react/text-class {:style (style/status-text)}]
       (-> content :parsed-text peek :children))]]])
 
-(defmethod ->message constants/content-type-emoji
-  [{:keys [content current-public-key outgoing public? pinned in-popover? message-pin-enabled] :as message} {:keys [on-long-press modal]
-                                                                                                             :as   reaction-picker}]
-  (let [response-to (:response-to content)
-        show-timestamp? (reagent/atom false)]
-    (fn [] [message-content-wrapper message
-            [react/touchable-highlight (when-not modal
-                                         {:disabled      in-popover?
-                                          :on-press      (fn []
-                                                           (react/dismiss-keyboard!)
-                                                           (reset! show-timestamp? true))
-                                          :delay-long-press 100
-                                          :on-long-press (fn []
-                                                           (on-long-press
-                                                            (concat
-                                                             [{:on-press #(re-frame/dispatch [:chat.ui/reply-to-message message])
-                                                               :id       :reply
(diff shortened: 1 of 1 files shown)


src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -612,23 +612,23 @@
...
-                                        [])))
-                :on-press (fn []
-                            (reset! show-timestamp? true))})
-             [react/view (style/message-view-wrapper (:outgoing message))
-              [message-timestamp message show-timestamp?]
-              [react/view {:style (style/message-view message) :accessibility-label :audio-message}
-               [react/view {:style (style/message-view-content)}
-                [message.audio/message-content message] [message-status message]]]]]
-            reaction-picker])))
+    [message-content-wrapper message
```
flexsurfer: but in that case `show-timestamp?` will be `false` each time , i guess you should move params in `fn[]` instead

flexsurfer (changes requested):

src/status_im/ui/screens/chat/message/message.cljs
Parveshdhull: Hi @flexsurfer, Thank you very much for reviewing PR. I done changes as you requested. Weirdly both changes created exact same result.
1. `show-timestamp?` sets false on re-render, but we are already doing after 2 seconds, for hiding timestamp
2. Even if animation is not complete(let's say duration is 20 sec), on re-render animations finished callback is called and sets it to false.

Still, yours is right way to do 👍, Thank you

flexsurfer (approved):

qoqobolo:
@Parveshdhull LGTM, thank you!

PR #13463, opened 2022-06-08
leave group chat fix

flexsurfer:
1) A adds B to group chat
2) B accepts group chat and enters group chat
3) B leaves group chat (now B is home)
4) A re-adds B
B sees the notification flashing, but it's accepted immediately,

### DIFF
--- a/src/status_im/group_chats/core.cljs
+++ b/src/status_im/group_chats/core.cljs
@@ -21,7 +21,8 @@
   {:events [:chat-removed]}
   [cofx response]
   (fx/merge cofx
-            {:dispatch-n [[:sanitize-messages-and-process-response response]
+            {:db (dissoc (:db cofx) :current-chat-id)
+             :dispatch-n [[:sanitize-messages-and-process-response response]
                           [:pop-to-root-tab :chat-stack]]}
             (notification-center/get-activity-center-notifications-count)))
 


Parveshdhull (approved):

qoqobolo:
@flexsurfer thanks for the fix!
Found a weird behavior with a pretty tricky reproduction, but I was unable to replicate in the develop, so might be worth a look. 

### ISSUE 1: User receives a notification in the activity center for the declined group, if they removed an admin of the group from contacts, been removed from the group members by the admin, and reopens the app

**Steps:**
1. Create UserA and UserB
2. Users add to contacts each other 
3. UserA creates a group with UserB
4. UserB declines the invitation
5. UserB removes UserA from contacts
6. UserB closes the app
7. UserA: group chat info > remove UserB from the group
8. UserB reopens the app and logs in

https://user-images.githubusercontent.com/67952253/172651564-47b8ea37-35cf-48b0-bcad-96121b074bed.mp4

qoqobolo:
@flexsurfer, I still managed to reproduce Issue 1 in yesterday's develop. 
Also, since https://github.com/status-im/status-react/pull/13428 has already been merged, the bug has probably gone (the user can no longer decline an invitation to the group chat).
So please ignore my comment above. 

PR can be merged.

PR #13489, opened 2022-06-16
Stop tracking VSCode config

audriu:
Stop tracking the '.vscode/' directory because git keeps showing uncommitted changes. The existing committed config is meaningless.

#### Platforms
- macOS
- Linux
- Windows

status: ready


jakubgs (approved):
Nice.

PR #13580, opened 2022-06-25
Implement identity ring

### LINKED ISSUE #13334: Implement identity ring
As a user
I want other users to display their identity ring in chat
So that I can make sure they are not impersonating someone else

### Description

In the new designs, there's an identity ring around users pictures.
This should be served as an image in status-go, similarly how we display profile pictures.
Ideally there's an endpoint that serves the profile picture already encircled in an identity ring, as a single picture.

### Designs

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=261%3A13411

### Notes

It needs to be consistent with desktop so 

1) use the same color palette (tricky thing is that according to design it differs per theme, see https://github.com/status-im/StatusQ/blob/bb9f08155fd68a0e2f1a257ee2566099c86cecd0/src/StatusQ/Core/Theme/StatusDarkTheme.qml#L165 and https://github.com/status-im/StatusQ/blob/bb9f08155fd68a0e2f1a257ee2566099c86cecd0/src/StatusQ/Core/Theme/StatusLightTheme.qml#L163)
2) use the same initalAngleRad and the direction of colors 

https://github.com/status-im/StatusQ/blob/master/src/StatusQ/Controls/StatusIdenticonRing.qml is the desktop implementation

https://github.com/status-im/status-go/blob/d0f4a94f7502ce37adf1f08065b07262b682c4bf/protocol/identity/colorhash/colorhash.go is the status-go implementation that defines how colorshash is constructed, but no rendering

qfrank:
fixes #13334

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -528,6 +528,21 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
     }
 
     @ReactMethod
+    private void drawIdentityImageRing(final String drawRingParamsJson, final Callback callback) {
+        Log.d(TAG, "drawIdentityImageRing");
+
+        Runnable r = new Runnable() {
+            @Override
+            public void run() {
+                String result = Statusgo.drawIdentityImageRing(drawRingParamsJson);
+                callback.invoke(result);
+            }
+        };
+
+        StatusThreadPoolExecutor.getInstance().execute(r);
+    }
+
+    @ReactMethod
     public void verify(final String address, final String password, final Callback callback) {
         Log.d(TAG, "verify");
         if (!checkAvailability()) {
--- a/modules/react-native-status/ios/RCTStatus/RCTStatus.m
+++ b/modules/react-native-status/ios/RCTStatus/RCTStatus.m
@@ -594,6 +594,16 @@ RCT_EXPORT_METHOD(openAccounts:(RCTResponseSenderBlock)callback) {
     callback(@[result]);
 }
 
+//////////////////////////////////////////////////////////////////// drawIdentityImageRing
+RCT_EXPORT_METHOD(drawIdentityImageRing:(NSString *)drawRingParamsJson
+                  callback:(RCTResponseSenderBlock)callback) {
+#if DEBUG
+    NSLog(@"DrawIdentityImageRing() method called");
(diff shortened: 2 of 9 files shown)


flexsurfer:
hey @qfrank thank you for the contribution, could you please elaborate on why we call draw-identity-ring from status-react ?

qfrank:
> hey @qfrank thank you for the contribution, could you please elaborate on why we call draw-identity-ring from status-react ?

Hi @flexsurfer , there is a message i'd like to quote: 
"looks like we can only use MediaServer (https://cdn.discordapp.com/attachments/754486551827841046/991568227421409372/unknown.png) after user logined, if this is the case, how can we make identity ring for this user screen (https://cdn.discordapp.com/attachments/754486551827841046/990171157636456468/unknown.png) via http server? 
one way i think is we still keep using base64 for this screen, while others using http server ? "

flexsurfer:
i think that's fine to do not have rings for the login screen

qfrank:
> i think that's fine to do not have rings for the login screen

emm.. status-desktop have rings for the login screen, i referenced it, if u're sure it's fine, i can remove it :) 
@flexsurfer

flexsurfer:
im sure its fine, but last word for @cammellos  :) i feel like it should be implemented in http server, so we shouldn't have any dedicated methods in status api for rings

cammellos:
> im sure its fine, but last word for @cammellos :) i feel like it should be implemented in http server, so we shouldn't have any dedicated methods in status api for rings

Yes, we should move everything on the http server, but @qfrank implemented that I believe, at least for identicon, we probably just need to look into starting the http server as soon as the app starts, but that would be a separate PR.

I don't have a strong opinion on whether we should keep the status-react code or not, it would be temporary in any case, it's handy for testing to check that it works, but I would merge the PR with or without to be honest, since the important part is that the HTTP server works.
We can always decide once we are closer to merging the PR, I think as long as identity ring in http server looks exactly like desktop, we are happy :)

flexsurfer:
@qfrank what's the status of this PR ? should we merge it ?

flexsurfer:
could you pls rebase and resolve conflicts

qfrank:
> @qfrank what's the status of this PR ? should we merge it ?

Hi @flexsurfer , sorry for the delay, strange that i didn't get the notification for your comments :).
This PR can be dropped, since we have [PR 13771](https://github.com/status-im/status-mobile/pull/13771)

PR #13602, opened 2022-06-30
Move automatic status updates timeout to status-go

### LINKED ISSUE #13601: Move timing out of automatic status updates to status-go
Currently, we are handling timing out of automatic status updates after 5 minutes on the client-side.

We should move this functionality to status-go and receive similar signals for timed-out status updates as we receive for the new status updates.

Parveshdhull:
fixes: https://github.com/status-im/status-react/issues/13601

**status-go PR**: https://github.com/status-im/status-go/pull/2757

**Testing Request**:
Please check if status updates, (especially automatic status updates) are working as expected and nothing is broken.

status: waiting for status-go testing

### DIFF
--- a/src/status_im/signals/core.cljs
+++ b/src/status_im/signals/core.cljs
@@ -75,4 +75,5 @@
       "wallet" (ethereum.subscriptions/new-wallet-event cofx (js->clj event-js :keywordize-keys true))
       "local-notifications" (local-notifications/process cofx (js->clj event-js :keywordize-keys true))
       "community.found" (link.preview/cache-community-preview-data (js->clj event-js :keywordize-keys true))
+      "status.updates.timedout" (visibility-status-updates/handle-visibility-status-updates cofx (js->clj event-js :keywordize-keys true))
       (log/debug "Event " type " not handled"))))
--- a/src/status_im/ui/screens/profile/visibility_status/utils.cljs
+++ b/src/status_im/ui/screens/profile/visibility_status/utils.cljs
@@ -59,44 +59,25 @@
     :title    (i18n/label :t/status-inactive)
     :subtitle (i18n/label :t/status-inactive-subtitle)}})
 
-;; Currently, Another user is broadcasting their status updates at the rate of 5 minutes.
-;; So for status-type automatic, we need to show
-;; that user online a little longer than that time. (broadcast receiving delay)
-(defn calculate-real-status-type-and-time-left
+(defn calculate-real-status-type
   [{:keys [status-type clock]}]
-  (let [status-lifespan    (if (= status-type
-                                  constants/visibility-status-automatic)
-                             (datetime/minutes 5.05)
+  (let [status-lifespan    (if (= status-type constants/visibility-status-automatic)
+                             (datetime/minutes 5)
                              (datetime/weeks 2))
         status-expire-time (+ (datetime/to-ms clock) status-lifespan)
-        time-left          (-  status-expire-time (datetime/timestamp))
-        status-type        (if (or (nil? status-type)
-                                   (and
-                                    (not= status-type
-                                          constants/visibility-status-inactive)
-                                    (neg? time-left)))
-                             constants/visibility-status-inactive
-                             status-type)]
-    {:real-status-type status-type
-     :time-left        time-left}))
+        time-left          (- status-expire-time (datetime/timestamp))]
+    (if (or (nil? status-type)
+            (and
+             (not= status-type constants/visibility-status-inactive)
+             (neg? time-left)))
+      constants/visibility-status-inactive
+      status-type)))
 
-(defn dot-color
-  [{:keys [status-type] :as visibility-status-update} my-icon?]
-  (if my-icon?
-    (if (= status-type constants/visibility-status-inactive)
-      colors/color-inactive quo2.colors/color-online)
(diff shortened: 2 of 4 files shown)


audriu (approved):
Looks ok to me.

src/status_im/signals/core.cljs
```diff
@@ -75,4 +75,5 @@
       "wallet" (ethereum.subscriptions/new-wallet-event cofx (js->clj event-js :keywordize-keys true))
       "local-notifications" (local-notifications/process cofx (js->clj event-js :keywordize-keys true))
       "community.found" (link.preview/cache-community-preview-data (js->clj event-js :keywordize-keys true))
+      "status.updates.timedout" (visibility-status-updates/handle-visibility-status-updates cofx (js->clj event-js :keywordize-keys true))
```
flexsurfer, line 78: typo ? or intentional ? time**D**out

Parveshdhull: Hi, Thank you for reviewing PR.
yes, intentional. Status-go is signaling for timed out status updates.

flexsurfer (approved):

qoqobolo:
@Parveshdhull thank you for PR!

It works great, except for one issue that I, unfortunately, can't find the exact steps to reproduce, but which I've stumbled upon a couple of times during testing. I don't think it's a blocker for PR, so just want to let you know about it. 

The problem is this:
Sometimes, the online status of your contact doesn't update if you go offline while your contact is still online, and you go back online when your contact is already offline. In that case, you will see that your contact is still online, although more than 5 minutes have definitely passed.

Suspected steps:
1. Device1 and Device2 create multiaccounts
2. D1 and D2 add each other to contacts and start 1-1 chat
3. D1 closes the app
4. D2 closes the app
5. D2 waits for >5 minutes and opens the app

Expected result: D1 is offline
Actual result (sometimes): D1 is still online _**until relogin on D2**_

Also, I suspect that it can be related to changing the mode of online status updates, i.e. changing the mode from "automatic" to "always online" and back, and under this condition, the problem was reproduced. But still not sure as can't reproduce it in this way. 

_Summarizing the above:_ the issue is not a blocker for PR, since it's not reproducible constantly and the online status updates after relogin, so it can be merged.
But maybe you’ll have some thoughts on it for further investigation.
____
Tested:
- modes: `automatic`, `always online`, `inactive`
- after killing the app and putting it in the background
- after adding to contact and removing
- airplane mode
- always online after >12 hours of inactivity 
- displaying the status on different screens 
- iOS, Android

Parveshdhull:
Hi Diana, Thank you very much for testing the PR and reporting this really interesting issue.

I am unable to reproduce it, but I think this issue might happen due to an edge case, where status-go and client both loads a `status-update` which was just about to expire. And client loaded it first and declared it as online, but `status-go` was a little late and declared it as already offline and didn't deactivate. (not sure though)

So, I changed the initial reference clock in status-go to - 5 seconds, so it will also send deactivated status updates in the last 5 seconds. Most probably this will fix this issue.

Also, this is a very small change (more like a safety net), I don't think it will affect PR testing.
https://github.com/status-im/status-go/compare/828a7f347427f582e362ffff0381258328a6999b..59d0e8fe1e61844639cd0d035c691524134360c4

Also, there is another scenario where it will not expire that status update. If due to some error, the loop for timing out never starts. But in that case, it will not expire any status update, even new status received for another contact or for new contacts added.

I am not sure if you remember any of those cases where you were able to reproduce this issue, but will then it stop expiring only that one particular status update, or all status updates?

Sorry, for the way too long explanation, and thanks again for testing PR, this issue must have been really hard to find 🙏

*PR Status:* waiting for desktop testing

Parveshdhull:
Hi @qoqobolo, Thank you very much for testing the PR.

After several tries and analyzing several scenarios, I found the issue.

### Reproduction Steps:
1. Add Devices 1 and 2 both as contacts
2. Close app in device 2
3. Keep Device 1 open, until device 2 starts to show as offline
4. Now also close device 1.
5. Now login to device 2, and wait for 5 seconds so that it can broadcast its online status and then close device 2 again.
6. Now login to device 1, and it will show device 2 as online, **forever**.

So, the issue was if on the app start device didn't find any status update to expire, it will again check after 5 minutes and will expire new status updates according to that time.
But another device had its online status update broadcasted with a time lesser than 5 minutes, it is just that they are not yet received by our device. Once they are received, they already missed the expiration cycle. And they can only be expired, either if the device will sends a new status update or the app restarts.

The issue is fixed now.

This issue was really interesting, must have been very hard to find 🙏.
Thank you.

qoqobolo:
> Reproduction Steps:
Add Devices 1 and 2 both as contacts
Close app in device 2
Keep Device 1 open, until device 2 starts to show as offline
Now also close device 1.
Now login to device 2, and wait for 5 seconds so that it can broadcast its online status and then close device 2 again.
Now login to device 1, and it will show device 2 as online, forever.

Aaahh I should have waited for 5 seconds and close device 2 again while device 1 is closed too!
Yes, I can reproduce it! And can confirm, that it's fixed in the latest build!
Awesome @Parveshdhull! 👏 Thank you very much for the investigation, finding the missing steps, and fixing the issue!

Parveshdhull:
**pr status:** waiting for status-go testing

PR #13625, opened 2022-07-04
[13576] Alert component

ibrkhalil:
image
image
image
image

fixes #13625

### Summary

Alert component for quo2, just like the RN one.

#### Platforms
- Android

status: ready

### DIFF
--- a/src/quo/react_native.cljs
+++ b/src/quo/react_native.cljs
@@ -55,12 +55,13 @@
 
 (def keyboard (.-Keyboard ^js rn))
 
-(defn alert [alert-title alert-msg & btns]
-  (.alert (.-Alert ^js rn) alert-title alert-msg (clj->js (mapv (fn [{:keys [text on-press cancelable on-dismiss]}]
-                                                                  {:text text
-                                                                   :on-press on-press
-                                                                   :cancelable (boolean cancelable)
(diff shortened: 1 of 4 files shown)


flexsurfer (commented):
seems like Podfile changes not related to this PR and could be reverted

flexsurfer:
seems like we have alert implemented https://github.com/status-im/status-react/blob/develop/src/status_im/utils/utils.cljs

src/quo/react_native.cljs
```diff
@@ -55,6 +55,14 @@
 
 (def keyboard (.-Keyboard ^js rn))
 
+(defn alert [alert-title alert-description btns]
```
flexsurfer, line 58: if we move to quo2, probably we should move react native as well, but honestly im not sure if it should be in quo at all , we need to discuss

ibrkhalil:
> seems like we have alert implemented https://github.com/status-im/status-react/blob/develop/src/status_im/utils/utils.cljs

Oh, I have not seen that
I think then the alert component does not need anymore work, right ?

flexsurfer:
i guess we could move it in quo2

flexsurfer:
anyway this one can be closed

PR #13645, opened 2022-07-10
[13558] Lowest price component

### LINKED ISSUE #13558: UI Component - Lowest Price
This issue is to track the creation of the component: **Lowest Price**

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=2222%3A31958.

ibrkhalil:
fixes #13558

### Summary

Lowest price component

### Testing notes
If possible text on different sized

- Android
- iOS

status: ready

### DIFF
--- a/src/quo2/components/info/lowest_price.cljs
+++ b/src/quo2/components/info/lowest_price.cljs
@@ -1,0 +1,72 @@
+(ns quo2.components.info.lowest-price
+  (:require [quo.react-native :as rn]
+            [clojure.string :as string]))
+
+(def centrify {:style {:flex-direction  :row
+                       :justify-content :center
+                       :align-items     :center}})
+
+(defn border-alignment [value-bg-color]
+  {:style {:align-self         :flex-start
+           :padding-horizontal 16
+           :padding-vertical   2
+           :flex-grow          0.25
+           :border-radius      4
+           :background-color   value-bg-color}})
+
+(def line-dots-props
+  {:ellipsize-mode  :middle
+   :number-of-lines 1
+   :font-size       10
+   :line-height     2})
+
+(defn dots
+  "Returns the dots"
+  [few-of-dots?]
+  (let [dots-text (->> ". "
+                       (repeat (if few-of-dots?
+                                 16
+                                 48))
+                       string/join
+                       (str " "))]
+    dots-text))
+
+(defn dots-comp
+  "Returns the dots adding their styles"
+  [few-of-dots?]
+  [rn/view {:style {:align-items :center
+                    :bottom      5}}
+   [rn/text line-dots-props (dots few-of-dots?)]])
+
+(defn lowest-price-value-comp
+  "Component responsible for the top and bottom values"
+  [top-value top-value-bg-color top-value-text-color]
+  [rn/view (border-alignment top-value-bg-color)
+   [rn/text {:style {:color      top-value-text-color
+                     :text-align :center}} top-value]])
+
+(defn lowest-price-styles [margin-top]
+  {:style {:flex-direction  :column
+           :width           "100%"
+           :overflow        :hidden
+           :margin-top      (int margin-top)
+           :justify-content :center}})
+
+(defn lowest-price
+  "Shows the lowest price component"
+  [{:keys [top-value
+           bottom-value
+           margin-top
+           top-value-text-color
+           top-value-bg-color
+           bottom-value-bg-color
+           bottom-value-text-color]}]
+  [rn/view (lowest-price-styles margin-top)
+   [rn/view centrify
+    [dots-comp true]
+    [lowest-price-value-comp top-value top-value-bg-color top-value-text-color]
+    [dots-comp false]]
+   [rn/view centrify
+    [dots-comp false]
+    [lowest-price-value-comp bottom-value bottom-value-bg-color bottom-value-text-color]
+    [dots-comp true]]])
--- a/src/quo2/screens/info/lowest_price.cljs
+++ b/src/quo2/screens/info/lowest_price.cljs
@@ -1,0 +1,1 @@
+(ns quo2.screens.info.lowest-price)
--- a/src/quo2/screens/main.cljs
+++ b/src/quo2/screens/main.cljs
@@ -28,6 +28,7 @@
             [quo2.screens.tags.token-tag :as token-tag]
             [quo2.screens.wallet.token-overview :as token-overview]
             [quo2.screens.list-items.preview-lists :as preview-lists]
+            [quo2.screens.info.lowest-price :as lowest-price]
             [re-frame.core :as re-frame]))
 
 (def screens-categories
@@ -57,7 +58,10 @@
            :component info-message/preview-info-message}
           {:name      :information-box
            :insets    {:top false}
-           :component information-box/preview-information-box}]
+           :component information-box/preview-information-box}
+          {:name      :lowest-price
+           :insets    {:top false}
+           :component lowest-price/preview-lowest-price}]
    :markdown [{:name      :texts
                :insets    {:top false}
                :component text/preview-text}]


src/quo2/screens/lowest_price.cljs
```diff
@@ -0,0 +1,56 @@
...
+    [rn/view {:style {:display :flex
+                      :flex-direction :column
+                      :height "100%"
+                      :justify-content :center}}
+     [rn/view {:style {:overflow :hidden}}
+      [rn/view {:style {:border-width 2
+                        :border-color (:text-01 @colors/theme)
+                        :border-radius 1
+                        :margin -2
+                        :poisiton :absolute
```
siddarthkay: seems to probably be a typo here 
it would be `:position :absolute`

ibrkhalil: Thanks Sid
Yeah it was a typo
Fixed

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,55 @@
...
+           margin-top
+           top-value-text-color
+           top-value-bg-color
+           bottom-value-bg-color
+           bottom-value-text-color]}]
+  (let [{:keys [height width]} @(re-frame/subscribe [:dimensions/window])]
+    [rn/view {:style {:display :flex
+                      :flex-direction :column
+                      :width "100%"
+                      :margin-top (number-util/numerical-string margin-top)
```
flexsurfer: i believe `(int margin-top)` should work

flexsurfer:
is changes in [Podfile.lock](https://github.com/status-im/status-mobile/pull/13645/files#diff-b7d85c35701918f25c9b2c418249f9b66fd9b067676e295a9eff493b73ee1752)  necessary for this PR? if no could we revert them ?

src/quo2/components/lowest_price.cljs
ibrkhalil: I tried it
But got
```
(int "2")

=> Execution error (ClassCastException) at user/eval5 (REPL:1).
class java.lang.String cannot be cast to class java.lang.Character (java.lang.String and java.lang.Character are in module java.base of loader 'bootstrap')
```

flexsurfer: mhm that's weird , why java ?

ibrkhalil: Oh
Was running a normal REPL
but on the CLJS REPL it fails too
I'll post here what i got previously

flexsurfer: works fine for me in cljs

ibrkhalil: You're correct, I tried it in a ClojureScript REPL and it worked .. I'll change it

ibrkhalil: Apologies
I changed it, I was using the normal Java's CLJ REPL to test.
I will use the CLJS one from now on for Status-mobile

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,54 @@
...
+                      :flex-direction :column
+                      :width "100%"
+                      :margin-top (int margin-top)
+                      :justify-content :center}}
+     [rn/view {:style {:overflow :hidden}}
+      [rn/view {:style {:border-width 2
+                        :border-color (:text-01 @colors/theme)
+                        :border-radius 1
+                        :margin -2
+                        :top (* 0.017 height)
```
flexsurfer: could you elaborate, this looks like a magic, components should depend on screen dimensions for sure

ibrkhalil: I tried to optimize how it looks to get the desired look by making it's value a ratio of the height .. To make it responsive

flexsurfer: are there design requirements for this ? i don't think it should be responsive

ibrkhalil: Used flex layout
Thanks 🙇🏻

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,74 @@
...
+       :flex-grow 4
+       :overflow :hidden
+       :border-radius 1}}]
+    [rn/view {:style {:align-self :flex-start
+                      :padding-horizontal 2
+                      :padding-vertical 2
+                      :flex-grow 0.25
+                      :border-radius 3
+                      :background-color bottom-value-bg-color
+                      :border-top-color "black"}}
```
flexsurfer: use black from colors ns

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,74 @@
+(ns quo2.components.lowest-price
```
flexsurfer: two suggestion to this file, when we have lots of styles when it hard to read code , better to create styles.cljs file and move all styles there , also please format styles maps

src/quo2/components/lowest_price.cljs
ibrkhalil: Done

src/quo2/components/lowest_price.cljs
ibrkhalil: Refactored

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,49 @@
...
+  (:require [quo.design-system.colors :as colors]
```
flexsurfer: quo2 colors should be used

flexsurfer (changes requested):

src/quo2/components/lowest_price.cljs
ibrkhalil: https://github.com/status-im/status-mobile/blob/8e29231a88cc0771221a7e7c4306191d457a7239/src/quo2/screens/main.cljs#L3

It's used inside quo2 to get the theme value if it's dark/white
Should we remake a theming namespace for quo2?

flexsurfer: theming should be done in separate PR

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,49 @@
...
+            [quo.react-native :as rn]))
+
+(defn line-styles [flex-grow]
+  {:opacity 0.4
+   :border-style :dotted
+   :border-width 1
+   :border-color (:text-01 @colors/theme)
```
Parveshdhull: You should use quo2.foundations.colors
`:border-color (colors/theme-colors colors/black colors/white)`

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,49 @@
...
+   :border-radius 1})
+
+(def centrify {:flex-direction :row
+               :justify-content :center
+               :align-items :center})
+
+(defn border-alignment [color]
+  {:align-self :flex-start
+   :padding-horizontal 2
+   :padding-vertical 2
```
Parveshdhull: why don't just use `:padding 2` ?

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,49 @@
...
+    [rn/view {:style (border-alignment top-value-bg-color)}
+     [rn/text {:style {:color top-value-text-color
+                       :text-align :center}} top-value]]
+    [rn/view {:style (line-styles 4)}]]
+   [rn/view {:style centrify}
+    [rn/view {:style (line-styles 1)}]
+    [rn/view {:style (border-alignment bottom-value-bg-color)}
+     [rn/text {:style {:color bottom-value-text-color
+                       :text-align :center}} bottom-value]]
+    [rn/view {:style (line-styles 1)}]]])
```
Parveshdhull: These are lots of views.
I am not sure if we are going to use this component as it is. Better to discuss with design team but probably only one of lowest price component will be shown as per position in graph.
Screenshot
 Something like this
Screenshot

src/quo2/components/lowest_price.cljs
```diff
@@ -0,0 +1,49 @@
...
+   :border-radius 3
+   :background-color color})
+
+(defn lowest-price
+  [{:keys [top-value
+           bottom-value
+           margin-top
+           top-value-text-color
+           top-value-bg-color
+           bottom-value-bg-color
```
Parveshdhull: We should only provide inputs for dynamic values like label. Props like, colors should be calculated automatically based on theme.

src/quo2/components/lowest_price.cljs
ibrkhalil: Oh I see
Thanks for example Parvesh
Will use it and resolve

src/quo2/components/lowest_price.cljs
ibrkhalil: Correct

src/quo2/components/lowest_price.cljs
ibrkhalil: Okay so they aren't supposed to be configurable .. I see

src/quo2/components/lowest_price.cljs
ibrkhalil: Will try to refactor it to use less views

src/quo2/components/lowest_price.cljs
ibrkhalil: Please check it now

src/quo2/components/info/lowest_price.cljs
```diff
@@ -0,0 +1,68 @@
...
+(def line-dots-props
+  {:ellipsize-mode  :middle
+   :number-of-lines 1
+   :font-size       10
+   :line-height     2})
+
+(defn dots
+  "Returns the dots"
+  [lots-o-dots?]
+  (let [dots-text " . . . . . . . . . . . . . . . . "]
```
briansztamfater, line 25: Sounds like many dots :D 

Probably it should be way more cleaner using something like `(str " " (clojure.string/join (repeat 16 ". ")))`

Also this can simplify the whole function

`(str " " (clojure.string/join (repeat (if lots-o-dots? 48 16) ". ")))`

briansztamfater: You can refer the join function on the requires as `[clojure.string :refer [join]]` to make it more readable

`(str " " (join (repeat (if lots-o-dots? 48 16) ". ")))`

briansztamfater: Sounds like many dots :D

Probably it would be way cleaner using something like `(str " " (clojure.string/join (repeat 16 ". ")))`

Also this approach can simplify the whole function

`(str " " (clojure.string/join (repeat (if lots-o-dots? 16 48) ". ")))`

briansztamfater: You could refer the join function on the requires as `[clojure.string :refer [join]]` to make it more readable

`(str " " (join (repeat (if lots-o-dots? 16 48) ". ")))`

briansztamfater (changes requested):

src/quo2/components/info/lowest_price.cljs
ibrkhalil: Thanks Brian .. Done

src/quo2/components/info/lowest_price.cljs
```diff
@@ -0,0 +1,72 @@
...
+
+(def line-dots-props
+  {:ellipsize-mode  :middle
+   :number-of-lines 1
+   :font-size       10
+   :line-height     2})
+
+(defn dots
+  "Returns the dots"
+  [few-o-dots?]
```
flexsurfer: should it be few-o**F**-dots?  with `f` ?

ibrkhalil: 😅 Will rename it now.

ibrkhalil: Done

src/quo2/components/info/lowest_price.cljs
```diff
@@ -0,0 +1,72 @@
...
+                       (repeat (if few-o-dots?
+                                 16
+                                 48))
+                       string/join
+                       (str " "))]
+    dots-text))
+
+(defn dots-comp
+  "Returns the dots adding their styles"
+  [lots-o-dots?]
```
flexsurfer: here lots, but dots expects few

ibrkhalil: Correct, I messed up that logic and fixed it in the function above .. Thanks for noticing it.

ibrkhalil: Done

flexsurfer (approved):

briansztamfater (approved):

briansztamfater (approved):

PR #13733, opened 2022-07-28
fix: set max option not work when send from qrcode

### LINKED ISSUE #11460: 'Set max' option is not working if "Send transaction" screen was opened using universal scanner
# Bug Report

## Problem
When scanning QR code with wallet address using universal QR code scanner or QR code scanner in wallet, "Set max" option does nothing.
BAACAgQAAxkBAAEtg2Nft74ea1UWPnrqdV9KPzcsFfC8QQACyggAApKSwVE6GTjljgizgh4E_1605877392 96

#### Expected behavior
can set max

#### Actual behavior
nothing happens

### Reproduction

- Open Status
- Using "Universal QR scanner" scan valid wallet address
- Try to "Set max"

### Additional Information

- Status version: nightly 20/11/20 

yqrashawn:
### Summary

resolve #11460

The bug is caused by lack of `[:wallet/prepare-transaction :gasPrice]` in db
code at: https://github.com/status-im/status-mobile/blob/7a97d88983cce567b781619ad379cf96227d1ecf/src/status_im/wallet/core.cljs#L390-L395

which is caused by these two functions https://github.com/status-im/status-mobile/blob/7a97d88983cce567b781619ad379cf96227d1ecf/src/status_im/wallet/choose_recipient/core.cljs#L54-L104

`fill-prepare-transaction-details` is used to fill `:wallet/prepare-transaction` based on `details` parse from URI (parsed from qr code here)
which won't fill `:gasPrice` if it's not presented in qr code

the solution here is to return `:signing/update-gas-price` fx if there's no `[:wallet/prepare-transaction :gasPrice]` in the result of `fill-prepare-transaction-details`

#### Platforms

- Android
- iOS

logic related to [eip-681](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-681.md) or this function https://github.com/status-im/status-mobile/blob/7a97d88983cce567b781619ad379cf96227d1ecf/src/status_im/wallet/choose_recipient/core.cljs#L113

### Steps to test

- Open Status
- Using "Universal QR scanner" scan valid wallet address
- Try to "Set max"

status: ready

### DIFF
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -7,7 +7,7 @@
                 [binaryage/oops "0.7.0"]
                 [com.andrewmcveigh/cljs-time "0.5.2"]
                 [status-im/timbre "4.10.0-2-status"]
-                [com.taoensso/encore "2.105.0"]
+                [com.taoensso/encore "3.24.0"]
                 [hickory "0.7.1"]
                 [cljs-bean "1.3.0"]
                 [com.cognitect/transit-cljs "0.8.248"]
--- a/shadow-cljs.edn
+++ b/shadow-cljs.edn
@@ -7,7 +7,7 @@
                 [binaryage/oops "0.7.0"]
                 [com.andrewmcveigh/cljs-time "0.5.2"]
                 [status-im/timbre "4.10.0-2-status"]
-                [com.taoensso/encore "3.24.0"]
+                [com.taoensso/encore "2.105.0"]
                 [hickory "0.7.1"]
                 [cljs-bean "1.3.0"]
                 [com.cognitect/transit-cljs "0.8.248"]
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -75,7 +75,7 @@
      :wallet/keys   [all-tokens] :as db} :db}
    {:keys [chain-id] :as data}
    uri]
-  (let [{:keys [address] :as details}
+  (let [{:keys [address gasPrice] :as details}
         (eip681/extract-request-details data all-tokens)]
     (if address
       (if (:wallet/recipient db)
@@ -84,11 +84,10 @@
         (if (:wallet/prepare-transaction db)
           {:db (update db :wallet/prepare-transaction assoc
                        :to address :to-name (find-address-name db address))}
-          (let [current-chain-id                     (get-in networks [current-network :config :NetworkId])
-                new-db-with-transaction-from-details (fill-prepare-transaction-details db details all-tokens)]
-            (merge {:db       new-db-with-transaction-from-details
+          (let [current-chain-id (get-in networks [current-network :config :NetworkId])]
+            (merge {:db       (fill-prepare-transaction-details db details all-tokens)
                     :dispatch [:open-modal :prepare-send-transaction]}
-                   (when-not (get-in new-db-with-transaction-from-details [:wallet/prepare-transaction :gasPrice])
+                   (when-not gasPrice
                      {:signing/update-gas-price
                       {:success-callback
                        #(re-frame/dispatch
--- a/src/status_im/wallet/choose_recipient/core.cljs
+++ b/src/status_im/wallet/choose_recipient/core.cljs
@@ -13,8 +13,7 @@
             [status-im.utils.http :as http]
             [status-im.utils.money :as money]
             [status-im.utils.universal-links.utils :as links]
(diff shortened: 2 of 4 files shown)


yqrashawn:
before:

https://user-images.githubusercontent.com/15090582/181487946-674b5884-9570-4495-b964-6b22258b9140.mp4

after:

https://user-images.githubusercontent.com/15090582/181487968-24ac2e31-e45f-458c-ac6c-01b4a12b68ce.mp4

yqrashawn:
And although there are no keys like `maxPriorityFeePerGas` in eip-681
we might need to get them handled

shadow-cljs.edn
```diff
@@ -7,7 +7,7 @@
                 [binaryage/oops "0.7.0"]
                 [com.andrewmcveigh/cljs-time "0.5.2"]
                 [status-im/timbre "4.10.0-2-status"]
-                [com.taoensso/encore "2.105.0"]
+                [com.taoensso/encore "3.24.0"]
```
flexsurfer: could you pls elaborate on why this change is needed for this PR ?

yqrashawn: Oh, that's not necessary.
I saw it's not used else where. So I thought I'll just upgrade it to the latest version, which have bug fixes and improvements, and it won't break anything.

Do I need to change it back?

flexsurfer: yeah its better to rever it, thanks

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -75,22 +76,29 @@
...
-                       :to address :to-name (find-address-name db address))}
-          (let [current-chain-id (get-in networks [current-network :config :NetworkId])]
-            (merge {:db (fill-prepare-transaction-details db details all-tokens)
-                    :dispatch [:open-modal :prepare-send-transaction]}
-                   (when (and chain-id (not= current-chain-id chain-id))
-                     {:ui/show-error (i18n/label :t/wallet-invalid-chain-id
-                                                 {:data uri :chain current-chain-id})})))))
-      {:ui/show-error (i18n/label :t/wallet-invalid-address {:data uri})})))
+  (enc/cond
+    :let [{:keys [address] :as details}
```
flexsurfer: could you elaborate pls on why `enc/cond`  is used, what is the point? thanks

yqrashawn: Less nested and more readable code

yqrashawn: Hi, I'm ok to change this back and use `if` instead. What do you think?

shadow-cljs.edn
yqrashawn: The version is reverted

src/status_im/wallet/choose_recipient/core.cljs
flexsurfer: it's better to use `if`, because it's more readable if are not familiar with `enc/cond`

yqrashawn: I just removed usage of `enc/cond` in the latest commit

src/status_im/wallet/choose_recipient/core.cljs
```diff
@@ -84,9 +84,17 @@
         (if (:wallet/prepare-transaction db)
           {:db (update db :wallet/prepare-transaction assoc
                        :to address :to-name (find-address-name db address))}
-          (let [current-chain-id (get-in networks [current-network :config :NetworkId])]
-            (merge {:db (fill-prepare-transaction-details db details all-tokens)
+          (let [current-chain-id                     (get-in networks [current-network :config :NetworkId])
+                new-db-with-transaction-from-details (fill-prepare-transaction-details db details all-tokens)]
+            (merge {:db       new-db-with-transaction-from-details
                     :dispatch [:open-modal :prepare-send-transaction]}
+                   (when-not (get-in new-db-with-transaction-from-details [:wallet/prepare-transaction :gasPrice])
```
flexsurfer: i guess you could just check it in details ?

flexsurfer:
```
[{:keys [address gasPrice] :as details}
        (eip681/extract-request-details data all-tokens)]
        ```

yqrashawn: fixed

flexsurfer (approved):
thank you

pavloburykh:
@yqrashawn thanx for the fix. Tested and ready for merge.

PR #13717, opened 2022-07-24
Switcher blurred background view

Parveshdhull:
### Summary
PR implements blurred background view for switcher and also allows switcher to draw over status bar in android, same as ios.

status: ready

PR #13747, opened 2022-08-01
[#13690] Fix visibility status color

### LINKED ISSUE #13690: online visibility indicator color is different on chat screen and home
# Bug Report

## Problem 
online visibility indicator color is different on chat screen and home

https://user-images.githubusercontent.com/17097240/179629282-db95ba12-ef46-481e-837c-877ec96b3a50.mp4

#### Expected behavior
Chat screen should not use visibility icon color from redesign, as redesign is still in WIP and should not affect current UI.

#### Actual behavior
 visibility indicator dot color for chat screen is being shown from redesign

### Reproduction

1) Open Status
2) Start Chat
3) Open Chat

### Additional Information

- Status version: nightly 

Most probably regression from https://github.com/status-im/status-mobile/pull/13640
cc: @briansztamfater

ilmotta:
Fixes #13690

### Summary

This PR fixes the visibility status color of the dot icon.

### Review notes

This PR fixes the issue by introducing a new state called `:ui/old?` (true by default). Components that are _directly_ impacted by the redesign can simply subscribe to the new state and react accordingly. **Edit**: thanks to reviewers, I could replace the new state with the existing atom `utils.config.new-ui-enabled?`.

AFAIU, the current approach to redesign on top of the existing design is to duplicate functions, using the suffix `*-old`. This seems to divert from re-frame's recommendation to derive the UI from the app db and also causes an explosion of duplicated functions throughout the codebase.

The duplicate+suffix strategy also leads to unnecessary changes to the call tree. For instance, some functions are duplicated because their downstream functions return a different color, but in reality, these parent functions have the exact same behavior. A concrete example, the parent component `emoji-chat-icon-view` depends on `profile-photo-plus-dot-view`, but it doesn't care what the inner component returns, so why should it change in a breaking way (var rename)? Of course, there are cases where it might be better duplicating a function, but most of the `*-old` ones I've seen could avoid such in-place changes.

If this new approach seems reasonable to y'all, perhaps future PRs could use the new subscription, which could help reduce the number of regressions while the redesign isn't finished.

### Testing notes

For manually testing between the old/new design, you can dispatch the event below in a REPL:

```clojure
(comment
  (re-frame.core/reg-event-fx
   :ui/toggle-old
   (fn [{:keys [db]}]
     {:db (update db :ui/old? not)}))

  (re-frame.core/dispatch [:ui/toggle-old]))
```

**Edit**: Use the existing effect to toggle the new UI.

```clojure
(comment
  (re-frame.core/dispatch [:toggle-new-ui]))
```

The video below demonstrates the app already fixed, with the toggle on/off and how it affects the visibility status icon.

https://user-images.githubusercontent.com/46027/182368085-ac5e033b-4bd6-4466-924a-2eb1ef46475c.mp4

#### Platforms

- Android
- iOS

### Steps to test

To reproduce the bug:

The issue https://github.com/status-im/status-mobile/issues/13690 describes the steps to reproduce, but to summarize:

1. Open Status.
2. Open the chat with someone who's currently active and visible to you.
3. Notice the _slight_ difference between the green colors in the visibility status _dot_ icon.

For QA purposes:

1. Open Status with the fix from this PR.
2. Open the chat with someone who's currently active and visible to you.
3. Toggle between the new/old design and notice the visibility status icon is displayed with the expected green color.

Status: ready

### DIFF
--- a/src/status_im/db.cljs
+++ b/src/status_im/db.cljs
@@ -13,7 +13,7 @@
              :sync-state                         :done
              :link-previews-whitelist            []
              :app-state                          "active"
-             :wallet                              wallet.db/default-wallet
+             :wallet                             wallet.db/default-wallet
              :wallet/all-tokens                  {}
              :peers-count                        0
              :node-info                          {}
@@ -34,9 +34,9 @@
              :registry                           {}
              :visibility-status-updates          {}
              :stickers/packs-pending             #{}
-             :keycard                            {:nfc-enabled?   false
-                                                  :pin            {:original     []
-                                                                   :confirmation []
-                                                                   :current      []
-                                                                   :puk          []
-                                                                   :enter-step   :original}}})
+             :keycard                            {:nfc-enabled? false
+                                                  :pin          {:original     []
+                                                                 :confirmation []
+                                                                 :current      []
+                                                                 :puk          []
+                                                                 :enter-step   :original}}})
--- a/src/status_im/ui/screens/profile/visibility_status/styles.cljs
+++ b/src/status_im/ui/screens/profile/visibility_status/styles.cljs
@@ -15,15 +15,15 @@
    :padding-right          12})
 
 (defn visibility-status-dot
-  [{:keys [dot-color size new-ui?]}]
+  [{:keys [color size new-ui?]}]
   (if new-ui?
-    {:background-color dot-color
+    {:background-color color
      :width            size
      :height           size
      :border-radius    (/ size 2)
      :border-width     3.5
      :border-color     (quo2.colors/theme-colors quo2.colors/white quo2.colors/neutral-90)}
-    {:background-color dot-color
+    {:background-color color
      :width            size
      :height           size
      :border-radius    (/ size 2)
--- a/src/status_im/ui/screens/profile/visibility_status/styles.cljs
+++ b/src/status_im/ui/screens/profile/visibility_status/styles.cljs
@@ -14,24 +14,27 @@
    :padding                6
    :padding-right          12})
 
-(defn visibility-status-dot [dot-color size]
-  {:background-color dot-color
-   :width            size
-   :height           size
-   :border-radius    (/ size 2)
-   :border-width     3.5
-   :border-color     (quo2.colors/theme-colors quo2.colors/white quo2.colors/neutral-90)})
+(defn visibility-status-dot
+  [{:keys [dot-color size new-ui?]}]
+  (if new-ui?
+    {:background-color dot-color
+     :width            size
+     :height           size
+     :border-radius    (/ size 2)
+     :border-width     3.5
+     :border-color     (quo2.colors/theme-colors quo2.colors/white quo2.colors/neutral-90)}
+    {:background-color dot-color
+     :width            size
+     :height           size
+     :border-radius    (/ size 2)
+     :border-width     1
+     :border-color     colors/white}))
 
-(defn visibility-status-dot-old [dot-color size]
-  {:background-color dot-color
-   :width            size
-   :height           size
-   :border-radius    (/ size 2)
-   :border-width     1
-   :border-color     colors/white})
-
-(defn visibility-status-profile-dot-old [color size border-width margin-left]
-  (merge (visibility-status-dot-old color size)
+(defn visibility-status-profile-dot
+  [{:keys [color size border-width margin-left new-ui?]}]
+  (merge (visibility-status-dot {:color   color
+                                 :size    size
+                                 :new-ui? new-ui?})
          {:margin-right 6
           :margin-left  margin-left
           :border-width border-width}))
--- a/src/status_im/ui/components/chat_icon/screen.cljs
+++ b/src/status_im/ui/components/chat_icon/screen.cljs
@@ -63,25 +63,6 @@
        [react/view {:style               dot-styles
                     :accessibility-label dot-accessibility-label}])]))
 
-(defn profile-photo-plus-dot-view-old
-  [{:keys [public-key photo-container photo-path community?]}]
-  (let [photo-path             (if (nil? photo-path)
-                                 @(re-frame.core/subscribe [:chats/photo-path public-key])
-                                 photo-path)
-        photo-container        (if (nil? photo-container)
-                                 styles/container-chat-list photo-container)
-        size                    (:width photo-container)
-        identicon?              (when photo-path (profile.db/base64-png? photo-path))
-        dot-styles              (visibility-status-utils/icon-visibility-status-dot-old
-                                 public-key size identicon?)
-        dot-accessibility-label (:accessibility-label dot-styles)]
-    [react/view {:style               photo-container
-                 :accessibility-label :profile-photo}
-     [photos/photo photo-path {:size size}]
-     (when-not community?
-       [react/view {:style               dot-styles
-                    :accessibility-label dot-accessibility-label}])]))
-
 (defn emoji-chat-icon-view
   [chat-id group-chat name emoji styles]
   [react/view (:container styles)
@@ -91,16 +72,6 @@
        [emoji-chat-icon emoji styles])
      [profile-photo-plus-dot-view {:public-key      chat-id
                                    :photo-container (:default-chat-icon styles)}])])
-
-(defn emoji-chat-icon-view-old
-  [chat-id group-chat name emoji styles]
-  [react/view (:container styles)
-   (if group-chat
-     (if (string/blank? emoji)
-       [default-chat-icon name styles]
-       [emoji-chat-icon emoji styles])
-     [profile-photo-plus-dot-view-old {:public-key      chat-id
-                                       :photo-container (:default-chat-icon styles)}])])
 
 (defn chat-icon-view-toolbar
   [chat-id group-chat name color emoji]
@@ -169,7 +140,7 @@
   (if group-chat
     [default-chat-icon icon-text styles]
     (let [photo-path @(re-frame.core/subscribe [:chats/photo-path chat-id])]
-      (if-not (string/blank? photo-path)
+      (when-not (string/blank? photo-path)
         [photos/photo photo-path styles]))))
 
 (defn emoji-chat-intro-icon-view [icon-text chat-id group-chat emoji styles]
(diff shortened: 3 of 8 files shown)


src/status_im/subs.cljs
```diff
@@ -89,6 +89,7 @@
 (reg-root-key-sub :waku/v2-peer-stats :peer-stats)
 (reg-root-key-sub :visibility-status-updates :visibility-status-updates)
 (reg-root-key-sub :navigation2/navigation2-stacks :navigation2/navigation2-stacks)
+(reg-root-key-sub :ui/old? :ui/old?)
```
flexsurfer: @Parveshdhull don't we already have such subscription ?

src/status_im/ui/screens/profile/visibility_status/styles.cljs
```diff
@@ -15,23 +16,22 @@
    :padding-right          12})
 
 (defn visibility-status-dot [dot-color size]
-  {:background-color dot-color
-   :width            size
-   :height           size
-   :border-radius    (/ size 2)
-   :border-width     3.5
-   :border-color     (quo2.colors/theme-colors quo2.colors/white quo2.colors/neutral-90)})
+  (if (<sub [:ui/old?])
```
flexsurfer: it's better to pass subs value from outside this function as a parameter

flexsurfer (changes requested):

src/status_im/subs.cljs
Parveshdhull: Yes, `status-im.utils.config/new-ui-enabled?`

src/status_im/db.cljs
```diff
@@ -7,13 +7,14 @@
 (def app-db {:contacts/contacts                  {}
              :pairing/installations              {}
              :group/selected-contacts            #{}
+             :ui/old?                            true
```
Parveshdhull: hi @ilmotta, Thank you very much for your PR.
I don't see where we are changing this variable to false, old icon colors is also showing for new ui.

src/status_im/subs.cljs
ilmotta: Shortsighted of me to not find such subscription. I should definitely replace the `:ui/old?` with `new-ui-enabled?`

src/status_im/db.cljs
ilmotta: Hi @Parveshdhull. I coded under the assumption that the old (current) UI is the only active one and that the new redesign couldn't dynamically be toggled on/off. You commented there's a subscription already for this `status-im.utils.config/new-ui-enabled?`, so I believe I got this part wrong. I should remove this new state instead.

src/status_im/ui/screens/profile/visibility_status/styles.cljs
ilmotta: Could you give me a bit more context as to why you think it's better to pass it as a parameter? Is it because  `**/styles.cljs` namespaces should not use re-frame subscriptions in general? Maybe it's an architectural decision (?), which would make sense if you ask me.

flexsurfer: yes subscriptions should be used only directly in the view, and not hidden in the styles or other functions, best way is to subscribe in `let` form in the view function

src/status_im/subs.cljs
```diff
@@ -404,6 +404,11 @@
 
 ;;GENERAL ==============================================================================================================
 
+(re-frame/reg-sub
+ :ui/new-enabled?
+ (fn [_]
+   @config/new-ui-enabled?))
```
flexsurfer: i don't think its reactive, so you should use it directly without subscription

src/status_im/subs.cljs
ilmotta: Hey folks, I replaced the unnecessary state to use `config/new-ui-enabled?` and it all works (https://github.com/status-im/status-mobile/pull/13747/commits/cf0ec7c544af3a95fedfffce69da0c178bbbccdd)

I noticed `config/new-ui-enabled?` is not available as a subscription from re-frame's db. Instead, it's a Clojure atom, which shouldn't be used in subscriptions, but it is (e.g. `:home-items`). Instead, a Reagent atom is preferable, because then re-frame can keep track of derefs automatically and re-render more efficiently. You probably know about this, but I just wanted to let you know just in case.

I also found something that might be a bug for another PR, the `:home-items` subscription is not dereferencing `config/new-ui-enabled?`, which means the check always returns true. Should this be reported in an issue? See https://github.com/status-im/status-mobile/blob/develop/src/status_im/subs.cljs#L1579

src/status_im/subs.cljs
ilmotta: I did try this, but it doesn't work @flexsurfer. The subscription really needs to return a boolean, otherwise `(<sub ...`) always evaluates to true.

flexsurfer: i mean you should use `@config/new-ui-enabled?` directly in the view

ilmotta: I see @flexsurfer, I just wanted to decouple the UI from having to know where the state comes from, but if you prefer the views to consume the state directly from the atom that's fine by me. I assumed you would be okay with this implementation because the `:home-items` subscription already uses `config/new-ui-enabled?`.

You might also be interested in an apparent bug in the `:home-items` subscription, related to this very discussion about the `config/new-ui-enabled?` atom (https://github.com/status-im/status-mobile/pull/13747#discussion_r935414648).

src/status_im/ui/screens/profile/visibility_status/styles.cljs
ilmotta: https://github.com/status-im/status-mobile/pull/13747/commits/ed07d93804a7eabfba54fb4ed7f56699da2845fe

src/status_im/subs.cljs
flexsurfer: `@config/new-ui-enabled?` used only for development, it's a temporary flag which will be removed when we have new design implemented

ilmotta: https://github.com/status-im/status-mobile/pull/13747/commits/79cb56a15be5a126bb0f4ee50fd9ba60b5297d2e

src/status_im/subs.cljs
ilmotta: > I also found something that might be a bug for another PR, the :home-items subscription is not dereferencing config/new-ui-enabled?, which means the check always returns true. Should this be reported in an issue? See https://github.com/status-im/status-mobile/blob/develop/src/status_im/subs.cljs#L1579

@Parveshdhull, @flexsurfer, FYI

Parveshdhull: `(and config/new-ui-enabled? (= view-id :chat-stack)))`

Hi, Thank you very much for finding this issue, yes `config/new-ui-enabled?` should be dereferenced here.
But interestingly, it didn't affect the resulting output, because even when `config/new-ui-enabled?` always remains true, for net output to be true `(= view-id :chat-stack)` also needs to be true. And view-id is only `chat-stack` in the new UI. (Old UI has different view-id for different tabs, like :home, :wallet, ...)

Still, thank you for finding this, feel free to create a separate PR or it will be fixed once this [commit](https://github.com/status-im/status-mobile/pull/13753/files#diff-21ed1701d49fc16290b71c7c74bfae9dc355e07910b0d4507e0d29a83aacc3bf) is merged.
> Should this be reported in an issue?

Yes, the issue sounds good, easy to track the fix.
Thank you

flexsurfer:
@ilmotta , could you please rebase you branch onto develop, and resolve conflicts if any, thank you

src/status_im/subs.cljs
ilmotta: Awesome @Parveshdhull, I think I'll leave for that commit to be merged given that you pointed out it's not currently affecting users.

flexsurfer (approved):
great job, thank you

qoqobolo:
Hey @ilmotta, thank you for your contribution and for the fix!
Found an issue, take a look, please.

### ISSUE 1: Missing colors for the online status indicator in the Profile view in the dropdown menu

**Steps:**
1. Create an account
2. Open Profile

**OS:** Android, iOS

**Actual result:**

https://user-images.githubusercontent.com/67952253/182619784-82205ebe-0da5-4fc8-8719-72b55b689e3a.mp4

**Expected result:**

https://user-images.githubusercontent.com/67952253/182620262-6104bb8f-e735-4c77-a066-249887f57f3f.MP4

ilmotta:
> Missing colors for the online status indicator in the Profile view in the dropdown menu

Hey @qoqobolo, I'll most definitely try to fix this today. Thanks for finding this out.

Parveshdhull (approved):

ilmotta:
> Missing colors for the online status indicator in the Profile view in the dropdown menu

@qoqobolo, I just fixed the problem in commit https://github.com/status-im/status-mobile/pull/13747/commits/a5ac96b9049934afe4c629c7e96ca995c8a10aea

I would like to call your attention to an **existing bug in the `develop` branch** that I haven't fixed because it wasn't part of the issue's scope of this PR (see https://github.com/status-im/status-mobile/issues/13690), but if you think I should fix it no worries, I could give it a try.

Here's the problem: when the new design is enabled in the `develop` branch, the status colors still use the current (old) design.

Parveshdhull:
> Here's the problem: when the new design is enabled in the develop branch, the status colors still use the current (old) design.

Hi, thank you for the screenshot, looks like in the new UI, the visibility status drop-down is also not aligning.
As the new UI is still in WIP, and we have not yet worked on profile section, maybe we can fix both of the issues when we redesign the profile section. What do you think @qoqobolo?

**UPD:** Looks like the status updates drop-down issue is introduced in https://github.com/status-im/status-mobile/pull/13717, as we now allow the switcher to draw on the status-bar, we also need to consider the status-bar's height for calculating the dropdown's position. Created issue for that https://github.com/status-im/status-mobile/issues/13758

For the issue about dot color in profile section, I think we can leave them for now, as we didn't redesigned this screen yet.

ilmotta:
Sounds good to me @Parveshdhull. Thanks for taking a look at the problem.

qoqobolo:
> Hi, thank you for the screenshot, looks like in the new UI, the visibility status drop-down is also not aligning.
As the new UI is still in WIP, and we have not yet worked on profile section, maybe we can fix both of the issues when we redesign the profile section. What do you think @qoqobolo?

Agree. We are not yet subjecting the new UI to comprehensive testing, as many parts are still WIP.
So yes, those issues can be fixed and tested later.

qoqobolo:
> I just fixed the problem in commit https://github.com/status-im/status-mobile/commit/a5ac96b9049934afe4c629c7e96ca995c8a10aea

Thanks @ilmotta!
PR is tested and can be merged.

PR #13759, opened 2022-08-03
fix error when switching dark/light mode while splash screen is shown

### LINKED ISSUE #13740: Keycard account is displayed as regular on login screen after switching dark/light mode
# Bug Report

## Problem 
_This is an edge case, but might be worth a look. Feel free to close if you think it's a `won't fix`._ 

There are two scenarios for regular and keycard accounts.

### 1. Keycard account [still relevant]

**Steps (same as for regular account):**
1. Create a keycard account on an Android device
2. Close the app
3. Open the app and very quickly tap the overview system Android button (square one) while the splash screen is shown
4. Change the mode of device from dark to light or wise versa
5. Open the app

**Actual result**: 

Two types of behavior were found on different Android devices.
- on Huawei P40 Lite: regular account is opened instead of keycard account; cannot log in until re-opening the app

https://user-images.githubusercontent.com/67952253/182149476-0d19033a-c7a4-4005-a592-965ac267a60c.mp4

- on Pixel 2: error (until re-opening the app)

Screenshot 2022-08-01 at 14 38 46

### 2. Regular account [FIXED in #13759, kept for context]

**Steps:**
1. Create a new account on an Android device
2. Close the app
3. Open the app and very quickly tap the overview system Android button (square one) while the splash screen is shown
4. Change the mode of device from dark to light or wise versa
5. Open the app

**Actual result**: error `Cannot read property 'root' of null`; reopening helps
**Expected result**: no error

https://user-images.githubusercontent.com/67952253/182148156-941420b0-821c-4aec-a98e-93b4b3296de5.mp4

### Additional Information

- Status version: 1.19
- Operating System: Android

Parveshdhull:
partially fixes https://github.com/status-im/status-mobile/issues/13740

### Testing Note:
Not sure if the keycard issue is also fixed (if both are caused by the same bug), only tested 1st one.

status: ready

### DIFF
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -144,9 +144,8 @@
                          (conj {:ms 900 :dispatch
                                 [:navigate-to :create-community-channel
                                  (get-in db [:navigation/screen-params :create-community-channel])]}))]
-
-    (navigation.core/dismiss-all-modals)
-    (when (or (nil? cur-theme) (zero? cur-theme))
+    (when (and (some? root-id) (or (nil? cur-theme) (zero? cur-theme)))
+      (navigation.core/dismiss-all-modals)
       (fx/merge cofx
                 (merge
                  {::multiaccounts/switch-theme (if (= :dark theme) 2 1)}


flexsurfer (approved):

qoqobolo:
Thanks for the fix @Parveshdhull!
LGTM except for the Keycard issue, I still can reproduce it.
Should I edit the issue to keep only the Keycard bug or you're going to try to fix this? (both are okay since it still looks like an edge case to me)

qoqobolo:
@Parveshdhull I've checked the new builds and that's what we have:

#### 1. In `c9bae13` the Keycard issue is still reproducible. 

#### 2. In `9809099` the Keycard issue is fixed.
However, in the video below you can see that the mode is not applied immediately if switch it while the splash screen is shown. It only switches after re-opening the app / login

https://user-images.githubusercontent.com/67952253/182863120-6903f6a8-be94-4ea9-ae22-a2358e26bcc2.mp4

And how it behaves in develop for comparison (the mode of device is applied immediately)

https://user-images.githubusercontent.com/67952253/182863308-5a876219-faaf-49f0-aea4-656cfe38b826.mp4

Parveshdhull:
hi @qoqobolo, Thank you very much for testing the new builds.

> However, in the video below you can see that the mode is not applied immediately if switch it while the splash screen is shown. It only switches after re-opening the app / login

Yes, this was intentional. I was making sure if the `system-theme-mode-changed` event is the culprit. And looks like it is.
But as I currently don't have access to the keycard account and am unable to debug, I am not exactly sure where the issue is happening inside that event. So, for now, I am thinking of only merging with the fix of the first issue. wdyt?

qoqobolo:
@Parveshdhull ah sure! I got so carried away by testing that just forgot it was just for checking, sorry!
 
Yea, let's merge the fix for the first issue and I'll edit the original one to keep the Keycard bug. 
Thank you!

PR #13761, opened 2022-08-04
nix: pass OPENSEA_API_KEY via saveAccountAndLogin

jakubgs:
Here the injection of OpenSea API key was done at compile time: https://github.com/status-im/status-mobile/commit/aa72ac57

But this makes `status-go` builds impure, and also prevents them from being extracted from `status-mobile` into `status-go` repo.

Instead we pass the `OPENSEA_API_KEY` env variable to JS bundle at build time, which is then passed to `status-go` via the `Statusgo.saveAccountAndLogin` call in `saveAccountAndLogin`:
https://github.com/status-im/status-mobile/blob/51174f848899360482b7ac55e2407b7cff76ab9e/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java#L323-L327

Which sends `NodeConfig` that also contains `WalletConfig` which can include `OpenseaAPIKey`:
```go
type WalletConfig struct {
    Enabled       bool
    OpenseaAPIKey string `json:"OpenseaAPIKey"`
}
```
https://github.com/status-im/status-go/blob/0135cc15/params/config.go#L510-L514

### DIFF
--- a/src/status_im/node/core.cljs
+++ b/src/status_im/node/core.cljs
@@ -60,6 +60,14 @@
     (assoc :ListenAddr ":30304"
            :DataDir (str (:DataDir config) "_dev"))))
 
+(def login-node-config
+  {:WalletConfig (cond-> {:Enabled true}
+                   (not= config/opensea-api-key "")
+                   (assoc :OpenseaAPIKey config/opensea-api-key))})
+
+(defn- get-login-node-config [config]
+  (merge config login-node-config))
+
 (defn- pick-nodes
   "Picks `limit` different nodes randomly from the list of nodes
   if there is more than `limit` nodes in the list, otherwise return the list
@@ -99,6 +107,9 @@
       :always
       (get-base-node-config)
 
+      :always
+      (get-login-node-config)
+
       current-fleet
       (assoc :NoDiscovery   wakuv2-enabled
              :Rendezvous    (if wakuv2-enabled false (boolean (seq rendezvous-nodes)))
@@ -118,8 +129,7 @@
                              :RendezvousNodes    (if wakuv2-enabled [] rendezvous-nodes)})
 
       :always
-      (assoc :WalletConfig {:Enabled true}
-             :LocalNotificationsConfig {:Enabled true}
+      (assoc :LocalNotificationsConfig {:Enabled true}
              :BrowsersConfig {:Enabled true}
              :PermissionsConfig {:Enabled true}
              :MailserversConfig {:Enabled true}
--- a/src/status_im/utils/config.cljs
+++ b/src/status_im/utils/config.cljs
@@ -20,10 +20,12 @@
 ;; flags stay up to date and are removed once behavior introduced is stable.
 
 (goog-define INFURA_TOKEN "800c641949d64d768a5070a1b0511938")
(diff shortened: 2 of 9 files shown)


flexsurfer (approved):

jakubgs:
It looks like - as @flexsurfer pointed out - it's not enough just to use the `Statusgo.saveAccountAndLogin` call, since that would leave already created accounts without an OpenSea key.

The correct solution appears to be using the new `LoginWithConfig` call:

* https://github.com/status-im/status-mobile/issues/13314
* https://github.com/status-im/status-mobile/pull/13414

src/status_im/utils/config.cljs
```diff
@@ -20,10 +20,12 @@
 ;; flags stay up to date and are removed once behavior introduced is stable.
 
 (goog-define INFURA_TOKEN "800c641949d64d768a5070a1b0511938")
+(goog-define OPENSEA_API_KEY "")
```
rasom, line 23: I think it makes sense to define it as `nil`. We can use any random value for this purpose but we already have `nil`.

src/status_im/node/core.cljs
```diff
@@ -60,6 +60,15 @@
     (assoc :ListenAddr ":30304"
            :DataDir (str (:DataDir config) "_dev"))))
 
+(def login-node-config
+  {:WalletConfig (cond-> {:Enabled true}
+                   (not= config/opensea-api-key "")
```
rasom, line 65: and then it would be just `(not (nil? config/opensea-api-key))`

src/status_im/node/core.cljs
```diff
@@ -60,6 +60,15 @@
...
+                   (assoc :OpenseaAPIKey config/opensea-api-key))})
+
+(defn- get-login-node-config [config]
+  (merge config
+         login-node-config))
```
rasom: It would be fine to have it in a single line. Not a big deal though.

src/status_im/node/core.cljs
flexsurfer: in that case it could be just `nil` we don't need condition, but not sure how it will work in json and then in status-go

src/status_im/utils/config.cljs
jakubgs: When i try to do:
```clj
(goog-define OPENSEA_API_KEY nil)
```
I just get:
```
IllegalArgumentException: no conversion to symbol
```
Which makes sense because:

>Use instead of [def](https://cljs.github.io/api/cljs.core/def) when you want to configure its value at compile-time with the [:closure-defines](https://cljs.github.io/api/compiler-options/closure-defines) compiler option. __Must be a string, number, or boolean.__

https://cljs.github.io/api/cljs.core/goog-define

src/status_im/node/core.cljs
jakubgs: Done.

jakubgs:
@status-im/mobile-qa The manual testing should focus on features related to [OpenSea](https://opensea.io/) NFT marketplace.

Thanks!

pavloburykh:
Tested and ready for merge. Thanx for cooperation @qoqobolo

PR #13762, opened 2022-08-04
permission tag component

### LINKED ISSUE #13678: UI Component - Permission Tags
This issue is to track the creation of the component: Permission Tags

You can access the links at: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=911%3A27048

jo-mut:
fixes #13678

Summary
This PR adds permission-tag component to quo2.preview


src/quo2/screens/permission_tag.cljs
```diff
@@ -0,0 +1,173 @@
...
+                  :options [{:key   32
+                             :value "32"}
+                            {:key   24
+                             :value "24"}]}
+                 {:label "Locked:"
+                  :key   :locked
+                  :type  :boolean}])
+
+(def community-tokens
+  [{:token-groups   [{:id  1 :tokens [{:id 1 :token-icon (resources/get-image :status-logo)}]}]}
```
J-Son89: you can reduce the size of this section in many ways. Consider at least using a small helper function to create the `tokens`

e.g

```
(defn get-token-icons [length]
 (map (fn [id] 
  {:id id :token-icon (resources/get-image :status-logo)})
 (range length)))

(def community-tokens
   
  [
    {:token-groups   [{:id  1 :tokens (get-token-icons 1)}]}
   {:token-groups   [{:id  1 :tokens  (get-token-icons 2)}]}
   {:token-groups   [{:id  1 :tokens (get-token-icons 3)}]}


   {:token-groups   [{:id  1 :tokens (get-token-icons 4)}]}

   {:token-groups   [{:id  1 :tokens (get-token-icons 1)}
                     {:id  2 :tokens (get-token-icons 1)}]}

   {:token-groups   [{:id  1 :tokens (get-token-icons 1)}
                     {:id  2 :tokens (get-token-icons 2)}]}

   {:token-groups   [{:id  1 :tokens (get-token-icons 1)}
                     {:id  2 :tokens (get-token-icons 3)}]}
                     ```

File: src/quo2/components/permission_tag.cljs [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+     [react/view {:padding-left    8
+                  :padding-right   (case size 32 16 24 12)}
+      [icons/icon icon
+       {:container-style {:align-items     :center
+                          :justify-content :center}
+        :resize-mode      :center
+        :size             (case size 32 20 24 16)
+        :color            icon-color}]]
+
+     (when (= (count token-groups) 1)
```
J-Son89: Consider reducing this down to avoid code re-use. Something like:

```
 (let [left-group   ((first token-groups) :tokens)
             center-group (when  (= (count token-groups) 3) ((second token-groups ) :tokens))
             right-group (when  (>= (count token-groups) 2) ((last token-groups ) :tokens))
             ]
         [react/view {:flex-direction :row
                      :align-items    :center}
          (tag-resources left-group size)
          [tag-extra-count left-group size]
          (when (>= (count token-groups) 3)
            [tag-or-clause size]
          )
            (tag-resources center-group size)
            [tag-extra-count center-group size]
          
          (when (>= (count token-groups) 2)
            [tag-or-clause size]
          )
            (tag-resources right-group size)
            [tag-extra-count right-group size]
          
          ])]))
  ```
Albeit, perhaps there's a better alternative.

File: src/quo2/components/permission_tag.cljs [hunk 3]
```diff
@@ -0,0 +1,125 @@
...
+      ^{:key token-icon}
+      [react/view {:flex-direction :row}
+       [react/view (outer-resource-container size)
+        [react/image {:source token-icon
+                      :style  {:height        (case size 32 28 24 20)
+                               :width         (case size 32 28 24 20)
+                               :border-radius size}}]]])))
+
+(defn tag-extra-count [tokens size]
+  (let [taken-tkns (take (if (> (count tokens) 3) 2 3) tokens)
```
J-Son89: IMO dropping letters from variable names offers no value. Here it's quite obvious what's implied but I think it's better to be implicit and use full words.
e.g `taken-tkns` -> `taken-tokens` - reduces the chance of another dev being confused.

src/quo2/components/permission_tag.cljs
```diff
@@ -0,0 +1,125 @@
...
+                      colors/neutral-20
+                      colors/neutral-70)
+   :height           (case size 32 28 24 20)
+   :width            (case size 32 28 24 20)
+   :border-radius    size
+   :justify-content  :center
+   :align-items      :center})
+
+(defn tag-resources [tokens size]
+  (let [tkns (take (if (> (count tokens) 3) 2 3) tokens)]
```
J-Son89: same here, consider renaming `tkns` to `taken-tokens` as below.

J-Son89:
Looks great! :)

One thing I noticed is that when I checked out your branch the console has some warnings ->
```
Warning: Encountered two children with the same key, `39`. Keys should be unique so that components maintain their identity across updates. 
...
```
Might be worth looking into where you're using two elements with the same key to keep the console clean.

src/quo2/components/base_tag.cljs
```diff
@@ -0,0 +1,27 @@
...
+  (fn [{:keys [id size disabled border-color border-width background-color accessibility-label
+               on-press type] :or   {size 32}} children]
+    [rn/touchable-without-feedback (merge {:disabled            disabled
+                                           :accessibility-label accessibility-label}
+                                          (when on-press
+                                            {:on-press #(on-press id)}))
+     [rn/view {:style (merge (style-container size disabled border-color border-width
+                                              background-color)
+                             (when (and (= type :icon) :emoji)
+                               {:width       size}))}
```
Parveshdhull: `(and (= type :icon) :emoji)` ?
Did you meant to compare `:emoji` with something?

Parveshdhull (commented):
great job !

src/quo2/components/base_tag.cljs
jo-mut: it should be (= type :icon). I will make that change

jo-mut:
the logic for this component is not yet very decided so i just came with code to show the design. The final code will maybe be quite different but my guess is not by much

PR #13771, opened 2022-08-07
Use image server for all kind of images and integrate identity rings 

### LINKED ISSUE #13670: Use image server for all kind of images and integrate identity rings
### Description

Currently the http server in status-go https://github.com/status-im/status-go/blob/develop/server/server.go is used only for identicon and image messages.
We should also use it for profile pictures. 

There's a branch out that also handles this case, which includes work from @qfrank:
https://github.com/status-im/status-mobile/tree/feature/identity-images-http-server
https://github.com/status-im/status-go/tree/feature/13334

This branch has a few changes:

- Adds an identity ring to the image
- Adds a signal when the port of the http server changes
- Serves user profile pictures

Some of the work integrating the branch has already been done here:

The important things to note when finishing integrating the branch:

- If the port of the server changes, the URL should of the image should be updated and re-fetched
- Status-go should not serve user images if the user has their settings set to not display profile pictures, only identicons should be server
- If this setting changes, pictures should be reloaded
- Dark/light theme should be likely passed as a parameter, so that the identity ring can be correctly set
- Identity ring should be toggled, as we don't want to show it in the old UI, only on the new UI

qfrank:
fixes #13670

### DIFF
--- a/src/status_im/signals/core.cljs
+++ b/src/status_im/signals/core.cljs
@@ -70,6 +70,7 @@
       "mailserver.available" (mailserver/handle-mailserver-available cofx (.-id event-js))
       "mailserver.not.working" (mailserver/handle-mailserver-not-working cofx)
       "discovery.summary"  (summary cofx (js->clj event-js :keywordize-keys true))
+      "mediaserver.started"  {:db (assoc db :mediaserver/port (.-port event-js))}
       "wakuv2.peerstats"  (wakuv2-peer-stats cofx (js->clj event-js :keywordize-keys true))
       "messages.new" (transport.message/sanitize-messages-and-process-response cofx event-js true)
       "wallet" (ethereum.subscriptions/new-wallet-event cofx (js->clj event-js :keywordize-keys true))
--- a/src/status_im/subs.cljs
+++ b/src/status_im/subs.cljs
@@ -41,7 +41,8 @@
             [status-im.utils.money :as money]
             [status-im.utils.security :as security]
             [status-im.wallet.db :as wallet.db]
-            [status-im.wallet.utils :as wallet.utils]))
+            [status-im.wallet.utils :as wallet.utils]
+            [status-im.utils.image-server :as image-server]))
 
 ;; TOP LEVEL ===========================================================================================================
 
@@ -266,6 +267,9 @@
 (reg-root-key-sub :wallet-connect-legacy/sessions :wallet-connect-legacy/sessions)
 (reg-root-key-sub :wallet-connect/session-managed :wallet-connect/session-managed)
 (reg-root-key-sub :contact-requests/pending :contact-requests/pending)
+
+; media-server
+(reg-root-key-sub :mediaserver/port :mediaserver/port)
 
 ; Testing
 
@@ -1303,15 +1307,27 @@
 (re-frame/reg-sub
  :chats/photo-path
  :<- [:contacts/contacts]
+ :<- [:profile/multiaccount]
+ :<- [:mediaserver/port]
+ (fn [[contacts multiaccount port] [_ id]]
+   (let [contact (or (get contacts id)
+                     (when (= id (:public-key multiaccount))
+                       multiaccount))]
+     (if (nil? contact)
+       (image-server/get-identicons-uri port id)
+       (multiaccounts/displayed-photo contact)))))
+
+(re-frame/reg-sub
+ :contacts/name-and-photo
+ :<- [:contacts/contacts]
  :<- [:multiaccount]
- (fn [[contacts multiaccount] [_ id identicon]]
+ (fn [[contacts multiaccount] [_ id]]
    (let [contact (or (get contacts id)
                      (when (= id (:public-key multiaccount))
                        multiaccount)
-                     (if (not (string/blank? identicon))
-                       {:identicon identicon}
-                       (contact.db/public-key->new-contact id)))]
-     (multiaccounts/displayed-photo contact))))
+                     (contact.db/public-key->new-contact id))]
+     {:name  (multiaccounts/displayed-name contact)
+      :photo (multiaccounts/displayed-photo contact)})))
 
 (re-frame/reg-sub
  :chats/unread-messages-number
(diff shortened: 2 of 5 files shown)


src/status_im/subs.cljs
```diff
@@ -2268,13 +2289,35 @@
  (fn [multiaccount]
    (get multiaccount :profile-pictures-visibility)))
 
+(defn- replace-contact-image-uri [contact port identity]
+  (let [theme (colors/current-theme)
+        identicon (image-server/get-identicons-uri port identity theme)
+        contact-images (:images contact)
+        contact-images (reduce (fn [acc image] (let [image-name (:type image)
+                                                       ; We pass the clock so that we reload the image if the image is updated
```
J-Son89, line 2343: This is some form of hack? 🤔 if so, is there no cleaner way of reloading the image on updates?

src/status_im/signals/core.cljs
```diff
@@ -70,6 +70,7 @@
       "mailserver.available" (mailserver/handle-mailserver-available cofx (.-id event-js))
       "mailserver.not.working" (mailserver/handle-mailserver-not-working cofx)
       "discovery.summary"  (summary cofx (js->clj event-js :keywordize-keys true))
+      "mediaserver.started"  {:db (assoc db :mediaserver/port (.-port event-js))}
```
flexsurfer, line 73: why do we need a separate signal? is it possible to start it on sign-in?

src/status_im/subs.cljs
```diff
@@ -2953,6 +3002,27 @@
  (fn [multiaccount]
    (pos? (count (get multiaccount :images)))))
 
+(re-frame/reg-sub
+ :profile/multiaccount
```
J-Son89, line 3066: nit: should the variable name be `multi-account`?

src/quo/design_system/colors.cljs
```diff
@@ -191,6 +191,12 @@
 
 (def theme-type (reagent/atom :light))
 
+(defn current-theme []
```
flexsurfer: mhm, probably its only related to image server better to move this to image server ns

src/status_im/subs.cljs
qfrank: i wish name it multi-account too...but the old code called it multiaccount, so i just want keep the same :)

src/quo/design_system/colors.cljs
flexsurfer: or rename it to `current-theme-int`

src/status_im/subs.cljs
```diff
@@ -1282,15 +1285,33 @@
...
  :<- [:contacts/contacts]
+ :<- [:profile/multiaccount]
+ :<- [:mediaserver/port]
+ (fn [[contacts multiaccount port] [_ id identicon]]
+   (let [contact (or (get contacts id)
+                     (when (= id (:public-key multiaccount))
+                       multiaccount))]
+     (if-not (nil? contact)
+       (multiaccounts/displayed-photo contact)
+       (if @config/new-ui-enabled?
```
flexsurfer: why only for new design?

src/status_im/subs.cljs
qfrank: hi @J-Son89  the "cleaner" way would be generate the image link on the status-go side, but there is some special params like "current theme / new-ui-enabled?" status-go don't know , if u have better idea, i'd like to hear :)

src/status_im/subs.cljs
qfrank: oh sorry, i think i misunderstood "Identity ring should be toggled, as we don't want to show it in the old UI, only on the new UI", when new-ui not enabled, we should also use image server but no ring, thanks for point out! @flexsurfer

src/status_im/signals/core.cljs
qfrank: @cammellos 

> why do we need a separate signal? is it possible to start it on sign-in?

cammellos: @flexsurfer we do, but sometimes it gets killed and the port changes, in that case we need to notify the UI so it stops using that port and reload images which might have not loaded, we could rename it something better I suppose?

src/status_im/utils/image_server.cljs
```diff
@@ -0,0 +1,28 @@
...
+  (:require [quo.design-system.colors :as colors]
+            [status-im.utils.config :as config]))
+
+(def ^:const image-server-uri-prefix "https://localhost:")
+(def ^:const identicons-action "/messages/identicons")
+(def ^:const account-images-action "/accountImages")
+(def ^:const contact-images-action "/contactImages")
+
+(defn current-theme []
+  (case @colors/theme-type
```
flexsurfer, line 11: i would abstract image server from colors , and pass it as a parameter , but its up to you

flexsurfer (approved):
great job!

src/status_im/utils/image_server.cljs
qfrank: emm.. i thought image-server would be much easier to use without passing param "theme"

churik:
@qfrank sorry for the delay, as we're close to the 1.20 release cut and this feature shouldn't be there.
When we cut the branch, I'll ask you to rebase all for testing (your forked repos in `status-go` and `status-react` side and corresponding branches as well)

qfrank:
@churik okay :)

qfrank:
Hi @churik , this PR rebased :)

churik:
@qfrank 
### ISSUE 1: There are no border at all for all users in chats, user profile, contacts list etc. in both dark and light mode

photo_2022-09-12 12 36 34 jpeg 2022-09-12 12-37-12

**OS:** IOS, Android


### ISSUE 2: Can't see an updated image in my profile until re-login 

**Steps:**
1) restore multi-account from seed phrase
2) open Profile > Edit image > Set new image


**Expected result:**
can see updated image

**Actual result:**

https://user-images.githubusercontent.com/4557972/189636783-0953a0b6-c3c7-4589-9b73-727d3506b144.mp4




**OS:** IOS, Android

qfrank:
Hi @churik , the ring only available when new-ui enabled

qfrank:
Hi @churik , the ring only available when new-ui enabled

churik:
@qfrank yep, I got it, issues are regression for current UI
Please, ping me when it is ready to be retested.
Thank you!

flexsurfer:
@qfrank so i guess for issue1 we lost a border

qfrank:
> @qfrank yep, I got it, issues are regression for current UI Please, ping me when it is ready to be retested. Thank you!

sure

qfrank:
Hi @churik , ready to be retested :)

churik:
No regression anymore, thank you for fixes @qfrank 
Tested basic flows, LGTM!

**Devices:**
- IPhone 12 Mini (IOS 15)

PR #13772, opened 2022-08-07
fix: quo haptic impact light key

phrohdoh:
### Summary

This key did not match the medium and heavy variants.


flexsurfer:
thanks

PR #13782, opened 2022-08-09
Remove contact request notification from activity center when request is canceled

rasom:
Currently if request is canceled on sender's side and on receiver's side the activity screen is opened the notification about request would stay on the screen.

status: ready


flexsurfer (approved):

qoqobolo:
Thanks @rasom! Tested and ready for merge.

PR #13796, opened 2022-08-11
Fix disappearing of the last notification when paired

rasom:
The issue: while working on syncing contacts request decision I noticed that if the activity center screen is opened while a new notification received it disappears shortly after being shown. That happens only if there is another paired device online and it also has activity center opened. The cause of it on status-go side is `SyncActivityCenterRead` which is sent by both devices right after receiving contact request (or any other messages which would add a new notification).

This PR introduces a fix on `status-react` side while the cause of the issue should still be fixed on status-go (separately).

status: ready


bitgamma (approved):

pavloburykh:
Failed e2e are not related to PR changes.

PR #13830, opened 2022-08-21
Implement Preview lists component

### LINKED ISSUE #13815: UI Component - Preview Lists
This issue is to track the creation of the component: Preview Lists

**Design:** https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=2219%3A32014

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/13815

status: ready


src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,100 @@
...
+                                 :rounded  9}
+                 :margin-left   -4
+                 :hole-size     26
+                 :hole-x        19
+                 :hole-y        -1}
+             16 {:border-radius {:circular 8
+                                 :rounded  8}
+                 :hole-radius   {:circular 9
+                                 :rounded  9}
+                 :margin-left   -4
```
briansztamfater, line 23: Is it ok that `margin-left` has the same value for `16` and `24`?

src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,100 @@
...
+            [quo2.components.markdown.text :as quo2.text]))
+
+(def params {32 {:border-radius {:circular 16
+                                 :rounded  10}
+                 :hole-radius   {:circular 18
+                                 :rounded  12}
+                 :margin-left   -8
+                 :hole-size     36
+                 :hole-x        22
+                 :hole-y        -2}
```
briansztamfater: Maybe separate params for each case in different objects so it is easier to read

src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,100 @@
...
+                                                 colors/neutral-40)}]
+     [quo2.text/text {:size   (if (= size 32) :paragraph-2 :label)
+                      :weight :medium
+                      :style  {:color       (get-overflow-color
+                                             transparent?
+                                             colors/white-opa-70
+                                             colors/neutral-60
+                                             colors/neutral-40)
+                               :margin-left -2}}
+      (if (< label 100) (str "+" label) (i18n/label :counter-99-plus))])])
```
briansztamfater: What is 100? If it is a maximum value it would be nice to add a comment to improve code readability

briansztamfater: Or use a constant

src/quo2/components/list_items/preview_list.cljs
Parveshdhull: hi @briansztamfater, Thank you very much for reviewing PR.
Yes, both of these items use same left margin

src/quo2/components/list_items/preview_list.cljs
Parveshdhull: Improved params, so they are more readable now

src/quo2/components/list_items/preview_list.cljs
Parveshdhull: Added comment, thank you

src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,115 @@
...
+                 :hole-x        11
+                 :hole-y        -1}})
+
+;; TODO - Add avatar components for other types once implemented
+(defn avatar [item type size]
+  (case type
+    :user [user-avatar/user-avatar
+           (merge {:ring?             false
+                   :status-indicator? false
+                   :size              (case size 32 :small 24 :xs 16 :xxxs)}
```
Parveshdhull, line 34: I don't know why we are using names like this for sizes. Figma doesn't use them and they are not consistent.
For example for [icon avatar](https://github.com/status-im/status-mobile/blob/3fc5c898658be7afa23d4d43a042777c1dddd629/src/quo2/components/avatars/icon_avatar.cljs#L8) `:big` is 48, while for [user avatar](https://github.com/status-im/status-mobile/blob/3fc5c898658be7afa23d4d43a042777c1dddd629/src/quo2/components/avatars/user_avatar.cljs#L10)  `:big` is 80.
Please can we use just size literals for sizes, as I am using in [preview_list](https://github.com/status-im/status-mobile/pull/13830/files#diff-1f6e7fdc52d1ca278044e0bb84505546cc8e91b1a17c8325c820fc99db63f97d) component instead of naming them
cc: @ibrkhalil

Parveshdhull: So, I can just pass `:size 32` instead of figuring out its name.

ibrkhalil: You're right .. We can pass the value instead .. that'll make it better

src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,115 @@
...
+                  item)]))
+
+(defn list-item [idx type size item list-size margin-left
```
J-Son89: consider changing variable `idx` to `index`. IMO character saving like this does not add any value and it's better to just avoid potential confusion.

src/quo2/components/list_items/preview_list.cljs
```diff
@@ -0,0 +1,115 @@
...
+                                                 colors/neutral-40)}]
+     [quo2.text/text {:size   (if (= size 32) :paragraph-2 :label)
+                      :weight :medium
+                      :style  {:color       (get-overflow-color
+                                             transparent?
+                                             colors/white-opa-70
+                                             colors/neutral-60
+                                             colors/neutral-40)
+                               :margin-left -2}}
+      ;; If overflow label is below 100, show label as +label (ex. +30), else just show 99+
```
J-Son89, line 81: is this comment really needed?

J-Son89 (approved):
I checked out your branch and checked the designs. It looks good to me on Figma 👍

src/quo2/components/list_items/preview_list.cljs
Parveshdhull: hi @J-Son89, Thank you very much for reviewing PR, changed idx to index.

src/quo2/components/list_items/preview_list.cljs
Parveshdhull: https://github.com/status-im/status-mobile/pull/13830#discussion_r967197344

PR #13722, opened 2022-07-25
[Issue-13560] User avatar component

ibrkhalil:
fixes #13560 

status: ready

Progress so far
image

Update
image

PR #13839, opened 2022-08-22
RE: User avatar component

### LINKED ISSUE #13560: UI Component - User Avatar
This issue is to track the creation of the component: **User Avatar**

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=115%3A6787.

ibrkhalil:
fixes: #13560 
status: ready


ibrkhalil:
Same as https://github.com/status-im/status-mobile/pull/13722
But to solve conflicts

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+     children]))
+
+(defn user-avatar
+  [{:keys [ring?
+           online?
+           size
+           status-indicator?
+           profile-picture
+           full-name]
+    :or {full-name "john doe"
```
flexsurfer: better to have something like `"empty name"`

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+         status-indicator? true
+         online? true
+         size :big
+         ring? true}}]
+  (let [initials (if full-name
+                   (reduce str (map first (split full-name " ")))
```
flexsurfer, line 110: could you elaborate why `map` and `first` and `reduce`?

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+                   "")
+        first-initial-letter (if full-name
+                               (first full-name)
```
flexsurfer: `(or (first full-name) "")`

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+                               "")
+        small-sizes #{:small :xs :xxs :xxxs}
```
flexsurfer: can be moved outside function ?

flexsurfer: const

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+        outer-dimensions (get-in sizes [size :outer])
+        inner-dimensions (get-in sizes [size (if ring?
+                                               :inner
+                                               :outer)])
+        font-size (get-in sizes [size :font-size])]
+    [rn/view {:style {:width outer-dimensions
+                      :height outer-dimensions
+                      :border-radius outer-dimensions}}
+     (when (and ring? identicon?)
+       [icons/icon :main-icons/identicon-ring32 {:width outer-dimensions
```
flexsurfer: you need to use icon from quo2 and size property

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+                                                 :height outer-dimensions
+                                                 :no-color true}])
+     (if using-profile-picture?
+       [react/image {:style  (container-styling inner-dimensions outer-dimensions)
+                     :source {:uri profile-picture}}]
+       [container inner-dimensions outer-dimensions
+        [rn/text {:style {:color colors/white-opa-70
+                          :font-size font-size}}
+         (upper-case (if-not (or (blank? first-initial-letter)
```
flexsurfer: move logic outside the UI

src/quo2/components/user_avatar.cljs
```diff
@@ -0,0 +1,142 @@
...
+                      :border-radius outer-dimensions}}
+     (when (and ring? identicon?)
+       [icons/icon :main-icons/identicon-ring32 {:width outer-dimensions
+                                                 :height outer-dimensions
+                                                 :no-color true}])
+     (if using-profile-picture?
+       [react/image {:style  (container-styling inner-dimensions outer-dimensions)
+                     :source {:uri profile-picture}}]
+       [container inner-dimensions outer-dimensions
+        [rn/text {:style {:color colors/white-opa-70
```
J-Son89: for text size should you not be using sizes from "src/quo2/foundations/typography.cljs"
e.g for font-size 27 
should it not be:
```
[quo2.foundations.typography :as typography]

[rn/text {
   :size typography/heading-1 ;; you can use this from your sizes mapping
   :style {:color colors/white-opa-70}}
...
```

J-Son89: I also realised you're not using the text component from quo2. Have you tried using that, is there any reason not to use that here?

J-Son89:
i checked out your pr for a quick check of designs:
Screenshot 2022-08-25 at 17 15 41

I think you are missing the font-weight for the text here.

Also it looks like the dot indicator positioning is a bit off. 🤔

src/quo2/components/user_avatar.cljs
ibrkhalil: Splitting the full name by a space will return a collection of two elemnets and we need to iterate over them so we'd use `map` and to get the first letter we use `first` and reduce will add them to a single string by casting `str` on them

src/quo2/components/user_avatar.cljs
ibrkhalil: Done

src/quo2/components/user_avatar.cljs
ibrkhalil: Done

src/quo2/components/user_avatar.cljs
ibrkhalil: Done

src/quo2/components/user_avatar.cljs
ibrkhalil: Done

src/quo2/components/user_avatar.cljs
ibrkhalil: Done

ibrkhalil:
> i checked out your pr for a quick check of designs: Screenshot 2022-08-25 at 17 15 41
> 
> I think you are missing the font-weight for the text here.
> 
> Also it looks like the dot indicator positioning is a bit off. thinking

Hey @J-Son89 Can you please check again ?

J-Son89:
> > i checked out your pr for a quick check of designs: Screenshot 2022-08-25 at 17 15 41
> > I think you are missing the font-weight for the text here.
> > Also it looks like the dot indicator positioning is a bit off. thinking
> 
> Hey @J-Son89 Can you please check again ?

yep, I'll give it a look

src/quo2/screens/user_avatar.cljs
```diff
@@ -0,0 +1,61 @@
...
+                  :type  :boolean}
+                 {:label "Full name separated by space"
+                  :key   :full-name
+                  :type  :text}
+                 {:label "Profile Picture URL"
+                  :key   :profile-picture
+                  :type  :text}])
+
+(defn cool-preview []
+  (let [state  (reagent/atom {:full-name "empty name"
```
J-Son89: consider changing this so the default initials are A Y and comparing the designs is faster.

J-Son89:
Text size on the Big implementation looks good but the ring is still slightly off. 
Screenshot 2022-09-01 at 22 03 05

Maybe it's okay though, not sure how pixel perfect designs need to be. 🤷‍♂️

J-Son89:
Similar for Medium 
Screenshot 2022-09-01 at 22 07 15

J-Son89:
Small size with ring looks good 👍 

Also I checked the Big without the ring and the dot seems to sit nicel. Probably needs to be slightly adjusted for when the ring is being used.

ibrkhalil:
Thanks JC
I read about VRT and applied it
Big: 
image
Medium: 
image

src/quo2/screens/user_avatar.cljs
ibrkhalil: Got a comment from Andre on making it as empty name

J-Son89 (approved):

PR #13866, opened 2022-08-28
[issue-13578] Dropdown component

### LINKED ISSUE #13578: UI Component - Dropdown
This issue is to track the creation of the component: **Dropdown**

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=114%3A6778.

ibrkhalil:
fixes #13578 

status: WIP


src/quo2/components/dropdown.cljs
```diff
@@ -0,0 +1,150 @@
...
+                                                  :padding-right 6}
+                              :padding-with-no-icon {:padding-vertical 3
+                                                     :padding-horizontal 8}}}})
+(defn color-by-10
+  [color]
+  (colors/alpha color 0.6))
+
+(defn dropdown-comp [{:keys [icon open? text dd-height size disabled? dd-color border-color]}]
+  (let [dark? (colors/dark?)
+        {:keys [width height width-with-icon padding font icon-size]} (get-in sizes [size])
```
ilmotta: The `get-in` functions doesn't seem to be necessary. It's also common in Clojure to treat maps as functions when there's a guarantee they're never nil, which is normally the case for maps defined as vars, like `sizes`. So if you prefer, the refactored code could be:

```clojure
{:keys [width height width-with-icon padding font icon-size]} (sizes size)
```

src/quo2/components/dropdown.cljs
```diff
@@ -0,0 +1,150 @@
...
+                                                    :flex-direction :row
+                                                    :align-items :center
+                                                    :border-width 1
+                                                    :border-color (if @open?
+                                                                    border-color
+                                                                    (color-by-10 border-color))
+                                                    :background-color (if @open?
+                                                                        dd-color
+                                                                        (color-by-10 dd-color))})
+                                            disabled? (assoc :opacity 0.3))}
```
ilmotta: I had a bit of difficulty reading the nested `cond->` calls because it was hard to keep track of which map was being updated (i.e. the inner style map or the parent one). In case you like my like suggestion, here's the code:

```clojure
(cond-> {:on-press (fn []
                     (if (swap! open? not)
                       (apply-anim dd-height 120)
                       (apply-anim dd-height 0)))
         :style (merge
                 (if icon
                   padding-with-icon
                   padding-with-no-icon)
                 {:width (if icon
                           width-with-icon
                           width)
                  :height height
                  :border-radius (if (= :small size)
                                   8
                                   10)
                  :flex-direction :row
                  :align-items :center
                  :border-width 1
                  :border-color (if @open?
                                  border-color
                                  (color-by-10 border-color))
                  :background-color (if @open?
                                      dd-color
                                      (color-by-10 dd-color))})}
  disabled? (assoc-in [:style :opacity] 0.3)
  disabled? (assoc :disabled true))
```

src/quo2/components/dropdown.cljs
```diff
@@ -0,0 +1,150 @@
...
+                            disabled? (assoc :disabled true))
+     (when icon
+       [icons/icon icon {:no-color true
+                         :size 20
+                         :container-style {:margin-right spacing
+                                           :margin-top 1
+                                           :width icon-size
+                                           :height icon-size}}])
+     [text/text {:size font-size
```
ilmotta: I've read comments in other PRs that we prefer to align keys, so it seems some parts of this namespace are misaligned.

I don't know if we've ever discussed or tried https://github.com/kkinnear/zprint. It's an incredibly flexible tool, written in Clojure for Clojure. I understand code style can be almost personal, but if everyone agrees on the very basic styles, like map keys alignment we could easily automate this in the future.

src/quo2/components/dropdown.cljs
```diff
@@ -0,0 +1,150 @@
...
+      {:no-color true
+       :size 20
+       :container-style {:width icon-size
+                         :margin-top 1
+                         :height icon-size}}]]))
+
+(defn items-comp
+  [{:keys [items on-select]}]
+  (let [items-count (count items)]
+    [:f>
```
ilmotta: Quick question: Why do we need `:f>` and the form-2 component (with the anonymous `fn` macro)? Given that the component doesn't have local state, couldn't this component be a simple form-1?

src/quo2/components/dropdown.cljs
```diff
@@ -0,0 +1,150 @@
...
+                                                         1)
+                                  :border-color (colors/theme-colors
+                                                 colors/black
+                                                 colors/white)
+                                  :text-align :center}
+                          :on-press #(on-select item)}
+                         [text/text {:style {:text-align :center}} item]]) items))])]))
+
+(defn dropdown [{:keys [items icon text default-item on-select size disabled? border-color dd-color]}]
+  [:f>
```
ilmotta: Is the `:f` necessary here? Isn't it just the case of defining the `reagent/atom` outside the anonymous function, just like any other form-2 component?

ibrkhalil: I think reanimated uses hooks .. And hooks need functional components

src/quo2/components/dropdown.cljs
ibrkhalil: To create a functional component

src/quo2/components/dropdown.cljs
ibrkhalil: Definitely more readable .. Thanks

src/quo2/components/dropdown.cljs
ibrkhalil: Thanks .. Used it

src/quo2/components/dropdown.cljs
ibrkhalil: Can you tell me an example of an alignment you think looks better ?

src/quo2/components/dropdown.cljs
ilmotta: Yeah, the `:f>` serves that purpose as you said, I was just wondering/questioning why it can't be a simpler form-1 component, i.e. there's no need for the surrounding `:f>` and no need for the anonymous function, or so it seems to me (I might be wrong). I think the general idea is to only use form-2 and functional components if needed (hooks, local state). But maybe it doesn't work as a form-1 component (I just don't understand why in this piece of code).

src/quo2/components/dropdown.cljs
ilmotta: Sure, my comment was specifically about misaligned map keys in this namespace, and I've seen other PRs where reviewers said we prefer to align map keys. So again, this is not necessarily my preference, but it seems the rest of team prefers this style. If you're under Emacs, you probably know you can call `C-c SPC` to auto indent map keys and let bindings.

For example:

```clojure
;; Misaligned map keys.
[icons/icon icon {:no-color true
                  :size 20
                  :container-style {:margin-right spacing
                                    :margin-top 1
                                    :width icon-size
                                    :height icon-size}}]

;; Aligned map keys.
[icons/icon icon {:no-color        true
                  :size            20
                  :container-style {:margin-right spacing
                                    :margin-top   1
                                    :width        icon-size
                                    :height       icon-size}}]
```

So, on top of this idea, I mentioned about zprint, a tool that can automate checks for some of our most agreeable code styles so that we can reduce the effort to review PRs for things that a machine can do for us. This is more like an idea that I would like to raise to the team, but I thought it was worth mentioning here for you.

ibrkhalil: Used vertical alignment

src/quo2/components/dropdown.cljs
ibrkhalil: I tried without it, And the functional component didn't work .. Maybe someone from the team can help out
@flexsurfer @Parveshdhull

ilmotta: Thanks for taking a look @ibrkhalil. I tested this branch locally, but the `items-comp` function works well with/without the surrounding functional form. That is, I can select the elements in the dropdown, the animations run and the state is correctly updated. What exactly did you find that's not working?

By the way, while checking this I noticed the dropdown is not displaying the currently active element. The diff below seems to fix the issue in case it helps:

```diff
modified   src/quo2/components/dropdown.cljs
@@ -40,7 +40,7 @@
   [color]
   (colors/alpha color 0.6))
 
-(defn dropdown-comp [{:keys [icon open? dd-height size disabled? dd-color border-color]}]
+(defn dropdown-comp [{:keys [icon open? dd-height size disabled? dd-color border-color default-item]}]
   (let [dark?                                                         (colors/dark?)
         {:keys [width height width-with-icon padding font icon-size]} (size sizes)
         {:keys [padding-with-icon padding-with-no-icon]}              padding
@@ -86,7 +86,7 @@
      [text/text {:size   font-size
                  :weight :medium
                  :font   :font-medium
-                 :color  :main} "Dropdown"]
+                 :color  :main} default-item]
      [icons/icon (if @open?
                    (if dark?
                      :main-icons/pullup-dark
```

ibrkhalil: Thanks Icaro for taking a look!
Actually the design request is for the dropdown element to always show Dropdown on it's view

I'll make a video of the issue I get when removing the functional component usage

ilmotta: > Actually the design request is for the dropdown element to always show Dropdown on it's view

Oh I missed that. Thanks @ibrkhalil

ibrkhalil: https://user-images.githubusercontent.com/33176106/190894785-44c0011d-339b-4a98-ac4e-99d6712e82be.mp4

Here's the video Icaro
And also I found this for [reference](https://clojurians-log.clojureverse.org/reagent/2021-02-19/1613740755.079900)

ilmotta: Thanks for the detailed video @ibrkhalil

ilmotta (approved):

src/quo2/components/list_items/dropdown.cljs
```diff
@@ -0,0 +1,148 @@
+(ns quo2.components.list-items.dropdown
```
J-Son89, line 1: I think the namespace is wrong here.

it should be `(ns quo2.components.dropdowns.dropdown`

J-Son89:
checking the designs.

Should the border be showing when the dropdown is disabled?
I can't see this state in the designs 🤔 

Screenshot 2022-09-20 at 11 21 15

Also in the figma designs. The text and icon are white when the background is purple but in quo2 the text and icon are black when the background is purple. 
Perhaps you handled this already? 🤨 

Screenshot 2022-09-20 at 11 24 56

src/quo2/components/list_items/dropdown.cljs
```diff
@@ -0,0 +1,148 @@
...
+       [icons/icon icon {:no-color        true
+                         :size            20
+                         :container-style {:margin-right spacing
+                                           :margin-top   1
+                                           :width        icon-size
+                                           :height       icon-size}}])
+     [text/text {:size   font-size
+                 :weight :medium
+                 :font   :font-medium
+                 :color  :main} "Dropdown"]
```
J-Son89, line 87: I assume this text "Dropdown" should be dynamic.. can you make this a prop and then pass "Dropdown" from the quo2 library?

src/quo2/components/list_items/dropdown.cljs
```diff
@@ -0,0 +1,148 @@
...
+                         :height        icon-size}}]]))
+
+(defn items-comp
+  [{:keys [items on-select]}]
+  (let [items-count (count items)]
+    [rn/scroll-view {:style               {:height "100%"}
+                     :horizontal          false
+                     :nestedScrollEnabled true}
+     (doall
+      (map-indexed (fn [idx item]
```
J-Son89: consider keep variable names verbose.
i.e `idx` -> `index`

src/quo2/screens/list_items/dropdown.cljs
```diff
@@ -0,0 +1,71 @@
...
+                 {:label "Border color"
+                  :key   :border-color
+                  :type  :text}
+                 {:label "DD color"
+                  :key   :dd-color
+                  :type  :text}
+                 {:label   "Size"
+                  :key     :size
+                  :type    :select
+                  :options [{:key   :big
```
J-Son89, line 39: nothing specific to your pr. But do we have an agreed naming/mapping for the size prop.. just to help with consistency across the component library. Seems we have this in most components
If not should we define this? @flexsurfer, @Parveshdhull WDYT? 🤔

Parveshdhull: I would recommend we don't try to give names to the sizes. For few components they are only 3 options :small, :medium, :big. But for some of them, they are way too many, :xxs, :xs, etc.
Also, discussed here: https://github.com/status-im/status-mobile/pull/13830#discussion_r967881050

It will be great, if we can also fix components which are already merged with size names. Maybe we can create issue about that?

src/quo2/screens/list_items/dropdown.cljs
```diff
@@ -0,0 +1,71 @@
...
+            [quo.previews.preview :as preview]
+            [quo.react-native :as rn]
+            [quo2.components.list-items.dropdown :as quo2]
+            [quo2.foundations.colors :as colors]
+            [reagent.core :as reagent]))
+
+(def descriptor [{:label   "Icon"
+                  :key     :icon
+                  :type    :select
+                  :options [{:key   :main-icons/placeholder
```
J-Son89, line 12: can you add an option of "none" so it's possible to verify the designs here.

src/quo2/screens/list_items/dropdown.cljs
```diff
@@ -0,0 +1,71 @@
...
+                             :value "Placeholder"}
+                            {:key   :main-icons/wallet
+                             :value "Wallet"}]}
+                 {:label "Disabled"
+                  :key   :disabled?
+                  :type  :boolean}
+                 {:label "Default item"
+                  :key   :default-item
+                  :type  :text}
+                 {:label "Border color"
```
J-Son89: can you make it possible to disable the border here too? so that the designs can be verified.
Consider using a `select` option with none as an optiona nd some example colors

src/quo2/components/list_items/dropdown.cljs
```diff
@@ -0,0 +1,148 @@
...
+                                                                   width-with-icon
+                                                                   width)
+                                               :height           height
+                                               :border-radius    (case size
+                                                                   :big 12
+                                                                   :medium 10
+                                                                   :small 8)
+                                               :flex-direction   :row
+                                               :align-items      :center
+                                               :border-width     1
```
J-Son89: I think the `border-width` affects the height and width of a component as it "eats" into the height and width. 
Is it better to wrap in a separate view component with the border set there as that it will add the border on and it easier to calculate.
@Parveshdhull I recall you discussing this before. WDYT?

src/quo2/components/list_items/dropdown.cljs
```diff
@@ -0,0 +1,148 @@
...
+            :medium {:icon-size 20
+                     :font      {:font-size :paragraph-1}
+                     :height    32
+                     :padding   {:padding-with-icon    {:padding-vertical   5
+                                                        :padding-horizontal 8}
+                                 :padding-with-no-icon {:padding-vertical 5
+                                                        :padding-left     12
+                                                        :padding-right    8}}}
+            :small  {:icon-size 12
+                     :font      {:font-size :label}
```
J-Son89: I think the sizes for small are incorrect. 
As far as I can see, small uses 13/Medium (paragraph-2)

and the icons seem too small too. 🤔 
Screenshot 2022-09-20 at 12 20 51

The black text/icons are the implementation and the white is the figma

src/quo2/components/list_items/dropdown.cljs
Parveshdhull: component has inner border in figma, so its ok in this case.
image

J-Son89: ah, I didn't know about this. Thanks Parvesh 👍

src/quo2/components/list_items/dropdown.cljs
ibrkhalil: Hey JC, I chatted with Pedro on this .. It's always going to be Dropdown

src/quo2/components/list_items/dropdown.cljs
ibrkhalil: Done

src/quo2/screens/list_items/dropdown.cljs
ibrkhalil: @Parveshdhull I'd ask for a green light from @flexsurfer or @cammellos first

src/quo2/screens/list_items/dropdown.cljs
ibrkhalil: I think clicking clear on the icon selector removes it

src/quo2/screens/list_items/dropdown.cljs
ibrkhalil: Done

src/quo2/components/list_items/dropdown.cljs
ibrkhalil: Fixed, Thank you so much Jamie for taking such a close look
image

src/quo2/components/list_items/dropdown.cljs
ibrkhalil: Done

src/quo2/components/list_items/dropdown.cljs
J-Son89: ok cool, then it should be i18n 👍

J-Son89 (approved):

PR #13897, opened 2022-08-30
fix Adding/removing user from contacts duplicating their messages in …

### LINKED ISSUE #13892: Adding/removing user from contacts duplicating their messages in profile
# Bug Report

## Problem 
If adding/removing a user via "Add to contacts" button in the user profile, it duplicates messages (if any) in their status.
Messages are shown there until next rel-ogin.

#### Expected behavior
Messages are not changed

#### Actual behavior

https://user-images.githubusercontent.com/4557972/187448339-cde146c5-8687-40af-8833-d98b813dd716.mp4

### Reproduction

1) User1 and User2: adds each other to contact
2) User1: post smth to status
3) User1: open profile of User2 (chat > ...> View profile or via contacts)
 4) User2: add/remove/re-add User1 to contacts

### Additional Information

- Status version: RC1 1.20 

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/13892

status: ready

### DIFF
--- a/src/status_im/chat/models/message.cljs
+++ b/src/status_im/chat/models/message.cljs
@@ -137,10 +137,11 @@
                [{:ms 100 :dispatch [:chat/add-senders-to-chat-users (vals senders)]}]))}))
 
 (defn reduce-js-statuses [db ^js message-js]
-  (let [chat-id (.-localChatId message-js)
+  (let [chat-id             (.-localChatId message-js)
         profile-initialized (get-in db [:pagination-info chat-id :messages-initialized?])
-        timeline-message (timeline-message? db chat-id)]
-    (if (or profile-initialized timeline-message)
+        timeline-message    (timeline-message? db chat-id)
+        old-message         (get-in db [:messages chat-id (.-id message-js)])]
+    (if (and (or profile-initialized timeline-message) (nil? old-message))
       (let [{:keys [message-id] :as message} (data-store.messages/<-rpc (types/js->clj message-js))]
         (cond-> db
           profile-initialized


Parveshdhull:
Hi, this description is misleading as the visibility of this dialog is not dependent on the contact state, but on the availability of status updates. This means this dialog doesn't hide on adding contact but on receiving new messages.
So, I made changes that status updates should only be visible after contact is added and until then show this dialog. Please let me know if this is desired behavior or if I should improve or revert.

<div align="center">
Screenshot
</div>
cc: @churik 

**UPD:**
Thank you @flexsurfer for confirming desired behavior here. (We should also see messages for not added contacts)

(It's just that dialog showing, Add contact while already added as contact looked a little suspicious to me.
We can leave this for now, and maybe condition this message in next release, I guess.
And, as Status updates also shown in profile, better message might be
"Add the profile as a contact to receive updates **also** on your timeline")

src/status_im/ui/screens/profile/contact/views.cljs
```diff
@@ -229,4 +228,4 @@
          :render-data               {:chat-id current-chat-id
                                      :profile true}
          :render-fn                 status.views/render-message
-         :data                      messages}]])))
+         :data                      (when (:added? contact) messages)}]])))
```
flexsurfer: we should see messages for non added contacts

src/status_im/chat/models/message.cljs
```diff
@@ -137,10 +137,11 @@
...
 
 (defn reduce-js-statuses [db ^js message-js]
-  (let [chat-id (.-localChatId message-js)
+  (let [chat-id             (.-localChatId message-js)
         profile-initialized (get-in db [:pagination-info chat-id :messages-initialized?])
-        timeline-message (timeline-message? db chat-id)]
-    (if (or profile-initialized timeline-message)
+        timeline-message    (timeline-message? db chat-id)
+        old-message         (get-in db [:messages chat-id (.-id message-js)])]
+    (if (and (or profile-initialized timeline-message) (nil? old-message))
```
flexsurfer, line 144: could you elaborate ? im not sure how new messages will be added then ?

flexsurfer (changes requested):

src/status_im/ui/screens/profile/contact/views.cljs
Parveshdhull: Hi @flexsurfer, Thank you for reviewing the PR and confirming desired behavior. (changes reverted)

src/status_im/chat/models/message.cljs
Parveshdhull: Hi, this flag only checks if we already have that message and skips that, otherwise just normal flow.

flexsurfer: oh right, just misread it, thank you

flexsurfer (approved):

qoqobolo:
This one can be merged and cherry-picked too.
Thanks @Parveshdhull!

PR #13899, opened 2022-08-31
Implement Token Gating

chrispader:
relates to: https://github.com/status-im/status-mobile/issues/14461

### Summary

Implements the Token Gating component, as per design here:
[https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=2463%3A39587](https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=2463%3A39587)

#### Platforms

- Android
- iOS

##### Functional

- public chats
- group chats
- networks

### Steps to test

- Open Status
- Login
- Try to join community/channel


chrispader:
Current status with mock data in quo2 sandbox:

Simulator Screen Shot - iPhone 13 - 2022-09-05 at 14 29 44

chrispader:
Waiting for the channel avatar component to get merged...

chrispader:
Updated channel avatar and made button dynamic. This is how the token gating drawer looks like currently, with

insufficient tokens:

Simulator Screen Shot - iPhone 14 Pro - 2022-09-29 at 19 17 19

and with sufficient tokens:

Simulator Screen Shot - iPhone 14 Pro - 2022-09-29 at 19 17 49

.gitignore
```diff
@@ -175,3 +175,5 @@ test/appium/tests/users.py
 ##node bindings
 /bin/
 /lib/
+
+/.calva/
```
J-Son89, line 183: should this be here? 🤨

ios/Podfile.lock
```diff
@@ -238,7 +238,7 @@ PODS:
     - React
   - react-native-status (1.0.0):
     - React
-  - react-native-status-keycard (2.5.37):
+  - react-native-status-keycard (2.5.38):
```
J-Son89: also needed?

src/quo2/components/buttons/button.cljs
```diff
@@ -66,7 +66,12 @@
                               :label                {:style    {:color colors/white}}
                               :background-color     {:default  colors/neutral-95
                                                      :pressed  colors/neutral-95
-                                                     :disabled colors/neutral-95}}}
+                                                     :disabled colors/neutral-95}}
+           :purple    {:icon-color       colors/white
+                       :label            {:style {:color colors/white}}
+                       :background-color {:default  (get-in colors/customization [:purple 50])
+                                          :pressed  (get-in colors/customization [:purple 50])
```
J-Son89: I think you should use `custom-color` function here instead.
It's in the colors file.
e.g
```
(colors/custom-color :purple 50)
```

src/quo2/components/buttons/button.cljs
```diff
@@ -124,7 +129,12 @@
                                                     :disabled colors/white-opa-5}}
            :shell            {:icon-color          colors/white
                               :label               {:style   {:color colors/white}}
-                              :background-color    {:default colors/neutral-95}}}})
+                              :background-color    {:default colors/neutral-95}}
+           :purple    {:icon-color       colors/white
+                       :label            {:style {:color colors/white}}
+                       :background-color {:default  (get-in colors/customization [:purple 50])
```
J-Son89: same here

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,140 @@
...
+                                                          true
+                                                          token-requirement-list))
+
+(defn token-requirement-list [props]
+  (let [{:keys [gate]} props
+        [gate-type token-requirement-list] gate
+        is-sufficient? (is-token-requirement-met? token-requirement-list)
+        message (if
+                 (= gate-type :join)
+                  (str "Join requirements" (when-not is-sufficient? " not") " met")
```
J-Son89: I think these strings should be wrapped in i18n labels and the string to the en.json files

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,140 @@
...
+                  (if (= gate-type :read)
+                    (str "View only requirements" (when-not is-sufficient? " not") " met")
+                    (str "View and post requirements" (when-not is-sufficient? " not") " met")))]
```
J-Son89: same here

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,140 @@
...
+                      [rn/view {:style (when-not (= token-index 0) (get styles :token-tag-spacing))}
+                       [token-tag/token-tag {:token (:token token)
+                                             :value (:amount token)
+                                             :size :big
+                                             :is-required (:is-sufficient token)
+                                             :is-purchasable (:is-purchasable token)
+                                             :token-img-src
+                                             (js/require
+                                              (str
+                                               "../resources/images/tokens/mainnet/"
```
J-Son89: hmm this seems like it should not be done this way. 

Perhaps you can move this js/require into the screens file so we can use it for the examples and just pass the `token-img-src` as a prop?

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,140 @@
...
+(defn token-gating [props]
+  (let [{:keys [community channel]} props
+        type (if (some? community) :community :channel)
+        inner-props (if (= type :community) community channel)
+        name (:name inner-props)
+        emoji (:emoji inner-props)
+        emoji-background-color (:emoji-background-color inner-props)
+        community-avatar (:community-avatar inner-props)
+        gates (:gates inner-props)
+        lock-status (if
```
J-Son89, line 98: I updated channel avatar component recently and it now uses `locked?` prop instead of `lock-status`. Can you rebase and update accordingly? :D

.gitignore
chrispader: I don't know if it should be in this PR or added somewhere else, but VS Code and Calva are producing some local output files when connected to the REPL.

Should we add this to .gitignore in another PR, @cammellos / @jakubgs  ?

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,140 @@
...
+                     (when-not (= gate-index 0)
+                       [rn/view {:style (get styles :divider)}])
+                     [token-requirement-list {:gate gate}]]) gates)
+
+     (when (= type :channel)
+       [:<>
+        [button/button {:type :purple :style (get styles :enter-button) :disabled lock-status} "# Enter channel"]
+        [text/text
+         {:size :paragraph-2 :style (get styles :info-text)}
+         "Entering channel will reveal your public addresses to the node owner"]])]))
```
J-Son89: I think these strings should also be i18n labels

ios/Podfile.lock
chrispader: yes, this adds support for Xcode 14 and iOS 16. https://github.com/status-im/react-native-status-keycard/pull/46

src/quo2/components/buttons/button.cljs
chrispader: Thanks! :) Didn't see that function

src/quo2/components/community/token_gating.cljs
chrispader: Thanks!

src/quo2/components/community/token_gating.cljs
chrispader: done!

src/quo2/components/community/token_gating.cljs
chrispader: Do you think? Otherwise we would have to pass the `token` and `token-img-src` props to the gate's token requirements.

Shouldn't this be as generic as possible?

J-Son89: I might be missing context on this, but is it just for tokens that we have in the `resources` folder or is possible that tokens (more so their images) are loaded in from some endpoint?

.gitignore
jakubgs: Why another PR? Just do it in a separate commit and don't squash it.

src/quo2/components/buttons/button.cljs
```diff
@@ -124,7 +129,12 @@
                                                     :disabled colors/white-opa-5}}
            :shell            {:icon-color          colors/white
                               :label               {:style   {:color colors/white}}
-                              :background-color    {:default colors/neutral-95}}}})
+                              :background-color    {:default colors/neutral-95}}
+           :purple    {:icon-color       colors/white
+                       :label            {:style {:color colors/white}}
+                       :background-color {:default  (colors/custom-color :purple 50)
+                                          :pressed  (colors/custom-color :purple 50)
+                                          :disabled (colors/custom-color :purple 50)}}}})
```
Parveshdhull: Hi @chrispader, thank you for your work
1. For dark it should be purple 60
2. This only covers a single case, when community theme color is purple. We need to find a better way for doing this.

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,168 @@
...
+            [quo2.foundations.colors :as colors]
+            [quo.theme :as theme]
+            [status-im.i18n.i18n :as i18n]))
+
+(def styles {:container {:flex 1}
+             :header-container {:flex-direction :row
+
+                                :align-items :center}
+             :header-spacing-community {:margin-bottom 16}
+             :header-spacing-channel {:margin-bottom 24}
```
Parveshdhull: Please align map key-values ([ref](https://github.com/bbatsov/clojure-style-guide#map-keys-alignment))

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,168 @@
...
+                       :margin-horizontal -20
+                       :margin-top 12
+                       :margin-bottom 8}
+             :enter-button {:margin-top 16}
+             :info-text {:margin-top 12
+                         :text-align :center
+                         :padding-horizontal 20}})
+
+(defn is-token-requirement-met? [token-requirement-list] (reduce
+                                                          #(or (and %1 (:is-sufficient %2)) false)
```
Parveshdhull: 1. Please move reduce to next line
2. can we use `(fn [var1 var2]` instead so it would be more clear what `%1` and `%2` are.
3. #(or ... false) ? (isn't or and false are redundant?

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,168 @@
...
+                                               ".png"))}]]])
+                   token-requirement-list)]]))
+
+(defn is-community-locked? [community]
+  (let [{gates :gates} community]
+    (or
+     (when (some? gates)
+       (when (vector? (:join gates))
+         (not (is-token-requirement-met? (:join gates)))))
+     false)))
```
Parveshdhull: Same here, `(or ... false)`?

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,172 @@
...
+                      :main-icons2/unlocked)
+          {:style (get styles :header-title-lock)
+           :color           (if dark?
+                              colors/neutral-40
+                              colors/neutral-50)}])]
+
+      [button/button {:type :outline
+                      :size 32
+                      :icon :info
+                      :style (get styles :header-info-button)} "info"]]
```
Parveshdhull: Please don't hard code string, use `(i18n/label :t/info)` instead

src/quo/react_native.cljs
```diff
@@ -173,3 +173,5 @@
      :y         y
      :height    height
      :width     width}))
+
+(defn alert [props] (.alert (.-Alert  rn) props))
```
flexsurfer: we have show-popup in utils

src/quo2/components/buttons/button.cljs
chrispader: Oh ok, didn't know, that this would be "community themed". Will change the schema for token gating, so a `community-color` can be defined :)

chrispader: But the button component color would also need to be adjustable...

chrispader: Will only specific (pre-defined) "theme colors" for communities be allowed, or can any color be chosen?

src/quo2/components/community/token_gating.cljs
chrispader: Should all map values be at the same column as well?

chrispader: It appears, `make lint-fix` and Calva Formatting doesn't fix this automatically...

src/quo2/components/buttons/button.cljs
Parveshdhull: any color ([ref](https://discord.com/channels/624205794384281629/626744384565674034/1019905254134911016))

src/quo2/components/community/token_gating.cljs
chrispader: 1. ✅
2. sure!
3. If i understand correctly, `and` returns the value of what is passed to it, also if it's `nil` for example...

>  If a form returns logical false (nil or false), and returns that value and doesn't evaluate any of the other expressions, otherwise it returns the value of the last expr. (and) returns true.

I added `(or ... false)` so that `is-token-requirement-met?` function definitely always returns a boolean.

What do you think?

src/quo2/components/buttons/button.cljs
chrispader: thanks!

src/quo/react_native.cljs
chrispader: thanks! didn't know that :)

src/quo2/components/community/token_gating.cljs
Parveshdhull: I am not sure why you need boolean, nil is also a logical false. but anyway you can use something like this
```
(every?
 (fn [token]
   (get-in token [1 :is-sufficient]))
 token-requirement-list)
```

src/quo2/components/community/token_gating.cljs
Parveshdhull: > Should all map values be at the same column as well?

I am not sure what you mean. do you mean nested maps? Its up to you, but the main point is to make the code more readable, for example

image

src/quo2/components/community/token_gating.cljs
chrispader: Thanks a lot! I'm such a Clojure noob still, missing so many useful clojure functions 😅

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,172 @@
...
+                      :main-icons2/unlocked)
+          {:style (get styles :header-title-lock)
+           :color           (if dark?
+                              colors/neutral-40
+                              colors/neutral-50)}])]
+
+      [button/button {:type :outline
+                      :size 32
+                      :icon true
+                      :style (get styles :header-info-button)} :info]]
```
Parveshdhull: > `:info`

Hi, Sorry if I was not clear, but I didn't meant that we should use keyword instead of string. I meant we can put this string in `/translations/en.json` and then use it like `(i18n/label :t/info)`. So this string can be translated for other languages.

chrispader: I know, but i noticed that i was using icon buttons wrong, since i don't need to provide a label (string) at all 😃

Parveshdhull: oh I see. So second variable/label is not optional? (maybe we can make it optional?)

image

Parveshdhull: ok I see now, second variable in this case is icon instead of label. So don't we need to use it like `:main-icons2/icon`?

chrispader: if i understood it correctly, if `:icon true` is set, instead of a label the icon name is provided as the second argument

chrispader: Label would need to be provided, if there should be text inside the button

chrispader: yes probably, is adding `:main-icons2/...` to the icon name just a convention or is this technically necessary?

Parveshdhull: I am not sure if it is necessary for quo2 button (maybe @flexsurfer will know better), but in general it is not just a convention. If you use `:main-icons/info` then file/image with `:info` name will be used as icon. But if you use `:main-icons2/info` then file/image with `:info+size` name will be used as icon. So in this case `:info32@2x.png`. ([ref](https://github.com/status-im/status-mobile/blob/8dd75205f359d0f8b3cf02e5cf8e662812fae94f/src/quo2/components/icon.cljs#L10))

flexsurfer: namespace is not used, we cut it, its only for IDE, but quo2 button uses quo2 icons which adds sizes to the icon name

chrispader: Oh ok, got it. So technically adding the namespace doesn't change anything. The only difference is in using button component from either quo or quo2. Thanks!

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,168 @@
...
+                                              (str
+                                               "../resources/images/tokens/mainnet/"
+                                               (:token token)
+                                               ".png"))}]]])
+                   token-requirement-list)]]))
+
+(defn is-community-locked? [community]
+  (let [{gates :gates} community]
+    (when (some? gates)
+      (when (vector? (:join gates))
```
Parveshdhull: you can just use `(when (vector? (:join gates))` only, `(when (some? gates)` is redundant

chrispader: thanks!

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,166 @@
...
+  (let [{gates :gates} community]
+    (when (vector? (:join gates)
+                   (not (is-token-requirement-met? (:join gates)))))))
+
+(defn is-channel-locked? [channel]
+  (let [{gates :gates} channel]
+    (if (vector? (:read gates))
+      (not (is-token-requirement-met? (:read gates)))
+      (when (vector? (:write gates))
+        (not (is-token-requirement-met? (:write gates)))))))
```
Parveshdhull: we need to check every time if it is vector, while calling `is-token-requirement-met?` function, maybe better to move this check to function itself.

Parveshdhull: so, this will become
```
(defn is-channel-locked? [channel]
  (let [{gates :gates} channel]
    (not (is-token-requirement-met? (or (:read gates) (:write gates)))
```

chrispader: Checking for `vector?` is necessary, because i want to check which "gates" are set. Moving this to the enclosing function would require us to add state, since the gates from the props can change

chrispader: if no `:read` gate is given, then this function will check if there is a `:write` gate...

Parveshdhull: current implementation will return true for both `vector?` and `is-token-requirement-met?` if empty vector is passed `[]`

You can do something like
```
(defn is-token-requirement-met? [token-requirement-list]
  (and (seq token-requirement-list)
       (every?
        (fn [token]
          (get token :is-sufficient?))
        token-requirement-list)))
```
> if no :read gate is given, then this function will check if there is a :write gate...

And, for `is-channel-locked?`
```
(defn is-channel-locked? [channel]
  (let [{gates :gates} channel]
    (not (is-token-requirement-met? (or (:read gates) (:write gates)))   <------ This will make sure to pass whichever is available
```

Parveshdhull: updated the function with `and`

chrispader: thanks! 👍

src/quo2/components/buttons/button.cljs
chrispader: Waiting for a approach from the design team on that, so this is still WIP

chrispader:
This is what a very complex variant of this component could look like:

Simulator Screen Shot - iPhone 14 Pro - 2022-10-18 at 15 51 22

chrispader:
@pedro-et @Parveshdhull we would need to add the `locked` icon with size 20 to the resources, otherwise using the icon doesn't work. For testing purposes i just exported it from Figma, but i think the format isn't 100% correct

src/quo/components/bottom_sheet/style.cljs
```diff
@@ -23,12 +25,13 @@
...
 
-(defn content-container
-  [window-height]
-  {:background-color        (:ui-background @colors/theme)
-   :border-top-left-radius  border-radius
-   :border-top-right-radius border-radius
-   :height                  (* window-height 2)})
+(defn content-container  [window-height]
+  (let [dark? (theme/dark?)]
+    {:background-color (if dark? colors2/neutral-90 colors/white)
```
Parveshdhull: hi @chrispader 
1. we should not change `quo` files
2. please use colors2 only. Here [themes-colors](https://github.com/status-im/status-mobile/blob/180df967e81c9c417e71db19670d1c64b56b2cfb/src/quo2/foundations/colors.cljs#L189) function can be used.
3. Also, if possible please do implementation and integration of component in separate PRs (easy to review)

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,215 @@
...
+                    (i18n/label
+                     :t/community-channel-read-requirements {:not not-text})
+                    (i18n/label
+                     :t/community-channel-write-requirements {:not not-text})))]
+    [rn/view
+     [rn/view {:style (get styles :token-requirement-text-spacing)}
+      [text/text {:weight :medium
+                  :number-of-lines 1} message]
+      [text/text {:size :paragraph-2
+                  :number-of-lines 1} "You must hodl:"]]
```
Parveshdhull: 1. Please move this string to [translations](https://github.com/status-im/status-mobile/blob/develop/translations/en.json) file
2. Maybe designs are changed, but it is not "hodl" anymore. [design](https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=2463%3A39587)

src/quo2/components/info/information_box.cljs
```diff
@@ -56,7 +56,9 @@
   [{:keys [type closable? closed? id icon style button-label on-button-press on-close]} message]
   (let [background-color (get-color-by-type type :bg)
         border-color     (get-color-by-type type :border)
-        icon-color       (get-color-by-type type :icon)
+        icon-color       (case (name icon)
+                           "untrustworthy" :no-color
+                           :none (get-color-by-type type :icon))
```
Parveshdhull: Please can you elaborate why these changes are needed.

src/quo2/components/list_items/channel.cljs
```diff
@@ -1,26 +1,72 @@
...
+                         :on-press (when locked?
+                                     #(>evt
+                                       [:bottom-sheet/show-sheet
+                                        {:content
+                                         (constantly [token-gating/token-gating
+                                                      {:channel {:name name
+                                                                 :community-color (colors/custom-color :pink 50)
+                                                                 :emoji emoji
+                                                                 :emoji-background-color channel-color
+                                                                 :on-enter-channel (fn [] (utils/show-popup "Entered channel" "Wuhuu!! You successfully entered the channel :)"))
```
Parveshdhull: 1. > Entered channel" "Wuhuu!! You successfully entered the channel :)

testing?
2. Maybe create function for this dispatch and assign it to on-press (more readable)

src/quo2/components/info/information_box.cljs
chrispader: Because otherwise the white content of the `untrustworthy` icon (exclamation mark) would get painted over.

I decided to specifically do this check in `information-box` component, because there shouldn't be the need to manually set something like `:no-color true` on the top level component and pass it down to the icon all the way.

What do you think?

src/quo/components/bottom_sheet/style.cljs
chrispader: The problem is, that the old `bottom-sheet` component has the wrong background-color, so the content doesn't match. Where should i change this instead?

Or should we create a new `quo2/bottom-sheet` component?

src/quo2/components/info/information_box.cljs
Parveshdhull: Thank you for explanation. There might be other use cases so let's not hard-code for `untrustworthy`
Maybe we can do something like this
```
icon-color (if no-icon-color? :transparent (get-color-by-type type :icon))
```
And we can remove changes from info_message.cljs

src/quo/components/bottom_sheet/style.cljs
Parveshdhull: Yes, I think we can copy this component in quo2 (or in ui2?) and make changes.
cc: @flexsurfer

src/quo2/components/info/information_box.cljs
chrispader: Done! 👍

chrispader: `:transparent` doesnt' quite work as an icon color. It will fallback to `(colors/theme-colors colors/neutral-100 colors/white)`

[https://github.com/chrispader/status-mobile/blob/893fc2c6db5101f653c2733edaf86482c83f8a45/src/quo2/components/icon.cljs#L23](https://github.com/chrispader/status-mobile/blob/893fc2c6db5101f653c2733edaf86482c83f8a45/src/quo2/components/icon.cljs#L23)

chrispader: I'd instead just pass the `:no-icon-color?` prop to `info-message` and then the icon

chrispader: Screenshot 2022-10-20 at 15 41 10

src/quo/components/bottom_sheet/style.cljs
flexsurfer: yeah but not in this PR

ios/Podfile.lock
```diff
@@ -640,7 +640,7 @@ SPEC CHECKSUMS:
   FBLazyVector: 3bb422f41b18121b71783a905c10e58606f7dc3e
   FBReactNativeSpec: f2c97f2529dd79c083355182cc158c9f98f4bd6e
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
-  glog: 36ce0530c6d2c3a5a4326885ef4069564887a1db
+  glog: 997518ea2aa2d8cd5df9797b641b758d52ecf2bc
```
ibrkhalil: Do we need this change ?

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,260 @@
...
+            [quo2.components.buttons.button :as button]
+            [quo2.components.icon :as icon]
+            [quo2.components.info.information-box :as information-box]
+            [quo2.components.markdown.text :as text]
+            [quo2.components.tags.token-tag :as token-tag]
+            [quo2.foundations.colors :as colors]
+            [status-im.i18n.i18n :as i18n]
+            [status-im.ui.components.fast-image :as fast-image]))
+
+(def token-tag-horizontal-spacing 7)
```
ibrkhalil: Can be private
`(def token-tag-horizontal-spacing 7)`
to
`(def ^:private token-tag-horizontal-spacing 7)`

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,260 @@
...
+   (map-indexed (fn [token-index token]
+                  (let [{:keys [token-img-src token amount is-sufficient? is-purchasable?]} token]
+                    ^{:key token-index}
+                    [rn/view {:style (get styles :token-tag-spacing)}
+                     [token-tag/token-tag {:token          token
+                                           :value          amount
+                                           :size           24
+                                           :border-color   community-color
+                                           :is-required    is-sufficient?
+                                           :is-purchasable is-purchasable?
```
ibrkhalil, line 80: Can we add a question mark at the end to indicate it's a boolean ?
`:is-required` -> `:is-required?`

src/quo2/components/community/token_gating.cljs
```diff
@@ -0,0 +1,260 @@
...
+                        :t/enter-channel))]
+            [text/text
+             {:size :paragraph-2
+              :style (merge
+                      (get styles :info-text)
+                      {:color (colors/theme-colors
+                               colors/neutral-50
+                               colors/neutral-40)})}
+             (i18n/label
+              :t/community-enter-channel-info)]]))])))
```
ibrkhalil: Kindly add a new line @EOF 😄

src/quo2/screens/community/token_gating.cljs
```diff
@@ -0,0 +1,165 @@
...
+                        :right 0
+                        :top 0}}
+       [preview/customizer state descriptor]]]
+     [rn/view {:height (if (= (:type @state) :community) 280 495) :margin-top 20}
+      [rn/view {:style (merge
+                        (get styles :container-sandbox)
+                        {:background-color (colors/theme-colors
+                                            colors/white
+                                            colors/neutral-90)})}
+       [quo2/token-gating preview-props]]]]))
```
ibrkhalil: New line at end of fille bitte ..

src/status_im/ui/screens/communities/community_overview_redesign.cljs
```diff
@@ -1,21 +1,31 @@
...
+            [status-im.ui.screens.communities.request-to-join-bottom-sheet-redesign :as request-to-join]
+            [status-im.ui.screens.communities.styles :as styles]
+            [status-im.utils.handlers :refer [<sub >evt]]
+            [status-im.utils.utils :as utils]))
+
+(def knc-token-img (js/require "../resources/images/tokens/mainnet/KNC.png"))
+(def mana-token-img (js/require "../resources/images/tokens/mainnet/MANA.png"))
+(def rare-token-img (js/require "../resources/images/tokens/mainnet/RARE.png"))
+(def eth-token-img (js/require "../resources/images/tokens/mainnet/ETH.png"))
+(def dai-token-img (js/require "../resources/images/tokens/mainnet/DAI.png"))
```
ibrkhalil: Kindly fix the require list in the file, I think Calva automatically sorts requires and therefore making unnecessary changes to files

ibrkhalil (changes requested):

ios/Podfile.lock
chrispader: This gets changed automatically for me once i run `make run-ios`...

src/quo2/components/community/token_gating.cljs
chrispader: That's implemented in `token_tag` component, not created by me. So i think this belongs into another PR

ios/Podfile.lock
ibrkhalil: Please revert

chrispader: 👍

src/quo2/components/community/token_gating.cljs
ibrkhalil: Kindly read this

https://matheus.ro/2017/12/11/clean-code-boy-scout-rule/

chrispader: All good, still i think these changes should be made in a separate PR. Other extra changes i made were requested to be implemented separately too.

ibrkhalil (approved):

cammellos:
@chrispader apologies, would you mind rebasing one last time and then we can merge?
Thank you!

chrispader:
> @chrispader apologies, would you mind rebasing one last time and then we can merge?
> Thank you!

done! 👍

PR #13906, opened 2022-08-31
ci: save and show both stdout and stderr

jakubgs:
By using the `|&` operator we send both to `tee`.

### DIFF
--- a/ci/Jenkinsfile.tests
+++ b/ci/Jenkinsfile.tests
@@ -54,10 +54,10 @@ pipeline {
     stage('Checks') {
       parallel {
         stage('Lint') {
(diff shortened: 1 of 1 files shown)


cammellos (approved):

jakubgs:
Oh for fuck's sake...
```
.../script.sh: 1: Syntax error: "&" unexpected
```
```
admin@linux-01.he-eu-hel1.ci.devel:~ % sh
$ { echo 'this is stdout'; echo 'this is stderr' >&2 } |& tee test.log
sh: 1: Syntax error: "&" unexpected
```
Looks like we can't use the fancy `|&` operator.

PR #13907, opened 2022-08-31
fix icon missing regression

### LINKED ISSUE #13904: Icons missing from most menus
# Bug Report

## Problem 
Regression from https://github.com/status-im/status-mobile/pull/13607

Screenshot 2022-08-31 at 16 46 19

Screenshot 2022-08-31 at 16 50 23

Screenshot 2022-08-31 at 16 51 26

#### Expected behavior
Menu with icons 

Screenshot 2022-08-31 at 16 47 26

### Reproduction

1. Open Status 
2. Navigate to different menus in the app (Home screen, Wallet, Browser, etc.)

### Additional Information

- Status version: version with merged https://github.com/status-im/status-mobile/pull/13607

siddarthkay:
fixes #13904

### Summary

This fixes an icon render issue which was introduced after refactoring the list-item quo component.

#### Platforms

- Android
- iOS

### Steps to test

Check all broken icons

status: ready

### DIFF
--- a/src/quo/components/list/item.cljs
+++ b/src/quo/components/list/item.cljs
@@ -12,6 +12,7 @@
             [status-im.ui.components.icons.icons :as icons]
             [quo2.foundations.colors :as quo2.colors]
             [quo2.components.icon :as quo.icons]
+            [status-im.utils.config :as config]
             [quo.components.animated.pressable :as animated]))
 
 (defn themes [theme]
@@ -77,14 +78,14 @@
         children))
 
 (defn icon-column
-  [{:keys [icon icon-bg-color icon-color size icon-container-style new-ui?]}]
+  [{:keys [icon icon-bg-color icon-color size icon-container-style]}]
   (when icon
-    (let [icon-size (size->icon-size size)]
+    (let [icon-size (size->icon-size size)
+          new-ui? @config/new-ui-enabled?]
       [rn/view {:style (or icon-container-style (:tiny spacing/padding-horizontal))}
        (cond
          (vector? icon)
          icon
-
          (keyword? icon)
          (if new-ui?
            [quo.icons/icon icon {:container-style {:align-items     :center
@@ -270,8 +271,7 @@
                    :subtitle                  subtitle
                    :subtitle-max-lines        subtitle-max-lines
                    :subtitle-secondary        subtitle-secondary
-                   :right-side-present?       (or accessory chevron)
-                   :new-ui?                   :new-ui?}]
+                   :right-side-present?       (or accessory chevron)}]
        [right-side {:chevron             chevron
                     :active              active
                     :disabled            disabled


ibrkhalil (approved):

src/quo/components/list/item.cljs
```diff
@@ -86,7 +86,7 @@
          icon
 
          (keyword? icon)
-         (if new-ui?
+         (when new-ui?
```
Parveshdhull: Hi, this is not the right way to fix this issue.
`(if cond a b)`
returns a when cond is true and b when false
`(when cond a b)`
`when` always returns last value when true and else nil.
here a is component is for new-ui and b is component for old-ui.
So yes, returning old-ui component will look like it fixes the issue. But it will break for new-ui.
(We need to check why cond (new-ui?) is true in case of old-ui)

Parveshdhull (changes requested):

src/quo/components/list/item.cljs
Parveshdhull: Also, when `new-ui?` is true, why empty space is shown instead of icon?

siddarthkay: > Hi, this is not the right way to fix this issue. `(if cond a b)` returns a when cond is true and b when false `(when cond a b)` `when` always returns last value when true and else nil. here a is component is for new-ui and b is component for old-ui. So yes, returning old-ui component will look like it fixes the issue. But it will break for new-ui. (We need to check why cond (new-ui?) is true in case of old-ui)



Makes sense, I will look into why `new-ui?` is `true`

siddarthkay: After more debugging I found out that the ` new-ui? ` code block in items component which was introduced  does nothing good either for new-ui or for old ui and hence I have removed it and icons work fine for both new/old ui.

@jo-mut  can you confirm please?

jo-mut: removing the `new-ui?` would solve the regression but the same problem would not be on the redesign because icons  for the old component dont work with the redesign

flexsurfer (approved):

Parveshdhull (approved):

jo-mut (approved):

pavloburykh:
hi @siddarthkay! Could you please rebase this branch to latest develop? Thanx!

siddarthkay:
Hi @pavloburykh : I've rebased the branch on latest develop

siddarthkay:
Rebase latest develop on my branch
328n9a

pavloburykh:
@siddarthkay hey, please let me know when PR will be ready for test. cause I started testing and have run e2e but I see that new rebase commits are still coming.

siddarthkay:
> @siddarthkay hey, please let me know when PR will be ready for test. cause I started testing and have run e2e but I see that new rebase commits are still coming. 

Hi @pavloburykh : ready to test now :)

pavloburykh:
@siddarthkay  thanx for the fix. Ready for merge.

PR #13858, opened 2022-08-26
Move abi-spec to status-go

qfrank:
fixes #13432

PR #13911, opened 2022-09-01
[#11335] Use Pokt network as an infura replacement

### LINKED ISSUE #11335: [Spike] Use Pokt network as an infura replacement
As a developer
I want to use Pokt instead of Infura 
As I want to use decentralized technologies

### Description

Assess the viability of using https://www.pokt.network/ as an infura replacement

We should start from using their gateway, and making sure that the app works as expected, later, we can move to https://github.com/status-im/status-react/issues/13591

rasom:
fix #11335

status: wip

### DIFF
--- a/src/status_im/utils/config.cljs
+++ b/src/status_im/utils/config.cljs
@@ -22,9 +22,15 @@
 (goog-define INFURA_TOKEN "800c641949d64d768a5070a1b0511938")
 (goog-define OPENSEA_API_KEY "")
 
-(def mainnet-rpc-url (str "https://mainnet.infura.io/v3/" INFURA_TOKEN))
-(def testnet-rpc-url (str "https://ropsten.infura.io/v3/" INFURA_TOKEN))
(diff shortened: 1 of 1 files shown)


rasom:
@cammellos @churik as far as I can tell it just works now. I see no difference comparing to infura

churik:
@rasom 
no balance is shown after  restoring multiaccount (should be shown on the Goerli network in PR builds), this is the reason of failed tests.
Balance is displayed only after `Scan tokens`
IMAGE 2022-09-06 14:33:32

pavloburykh:
> @rasom no balance is shown after restoring multiaccount (should be shown on the Goerli network in PR builds), this is the reason of failed tests. Balance is displayed only after `Scan tokens` IMAGE 2022-09-06 14:33:32

@churik I've logged the same issue here https://github.com/status-im/status-mobile/pull/13858#issuecomment-1238315358 and it seems already to be fixed there.

Issue #13817, opened 2022-08-18
Implement request to join communities flow in the new UI


cammellos:
As a user
I want to be able to request to join a community
So that I can participate and read messages

### Description

We should implement the screens in the new UI for joining a community.
For now, ignore token gated communities.

The initial screen is:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1355%3A111104 (that should be already implemented by @jo-mut )

Then follow the manual approval step:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1356%3A115613

Ignore captcha & rules screens and you can land directly on:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A108004

Of this screen only `Leave community` and  `Share community` are implemented

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A108067

All of these are also implementable:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A115668
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1387%3A104203
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1386%3A110589

Some of the data might not be available for a community yet (Tags, number of users for example), in that case it's fine to mock the data.

jo-mut:
@cammellos @J-Son89 I will get it merged

flexsurfer:
@J-Son89 @jo-mut can we close this ?

J-Son89:
not just yet, I was waiting for page nav component which has just been merged so putting that into the design. 
There is also some advanced styling features which I need to discuss with @cammellos whether I put in another feature or put it as part of this work.

PR #13921, opened 2022-09-02
chore: refactor communities code for reusabilty

J-Son89:
I'm working on the issue -
https://github.com/status-im/status-mobile/issues/13817

To introduce the join screens for communities using the new ui.

The p.r for this work is quite large so I am going to break it up into smaller chunks.

This p.r refactors the communities card so that I can reuse the same code on the landing page of a community.


src/status_im/ui/screens/communities/styles.cljs
```diff
@@ -29,40 +29,31 @@
 
 (defn card-stats-container []
   {:flex-direction :row
```
J-Son89, line 31: general approach here was to make styling relative here and add some parameters to the functions so they can be reused

src/quo2/components/communities/community_view.cljs
```diff
@@ -0,0 +1,86 @@
+(ns quo2.components.communities.community-view
+  (:require
```
J-Son89, line 2: split the community card out into 3 parts, the core code and it's 2 use cases. 
The core code I will reuse in the following pr for the community landing page.

J-Son89:
Everything lines up as before. Text slightly adjusted in one position to better fit to designs as alignment was slightly off.


This image is old implementation (bottom layer) against new implementation
Screenshot 2022-09-05 at 13 06 12

src/status_im/ui/screens/communities/styles.cljs
```diff
@@ -80,15 +71,12 @@
...
-(defn community-title-description-container []
-  {:position  :absolute
-   :top       32
-   :left      12
-   :right     12})
+(defn community-title-description-container [{:keys [margin-top]}]
+  {:margin-top margin-top})
 
-(defn community-cover-container []
+(defn community-cover-container [{:keys [height]}]
```
flexsurfer: better to pass only height

src/status_im/ui/screens/communities/styles.cljs
```diff
@@ -29,40 +29,31 @@
...
+   :padding-left padding-horizontal
+   :padding-right padding-horizontal
    :border-radius      16
-   :padding-horizontal 12
    :background-color (colors/theme-colors
                       colors/white
                       colors/neutral-90)})
 
-(defn card-view-chat-icon []
+(defn card-view-chat-icon [{:keys [top left]}]
```
flexsurfer: same

src/status_im/ui/screens/communities/styles.cljs
```diff
@@ -29,40 +29,31 @@
...
 (defn community-tags-container  []
   {:flex-direction :row
-   :position  :absolute
-   :top       154
-   :left      12
-   :right     12})
+   :margin-top 16})
 
-(defn card-view-content-container []
+(defn card-view-content-container [{:keys [padding-horizontal]}]
```
flexsurfer: same

src/quo2/screens/community_card_view.cljs
```diff
@@ -5,15 +5,16 @@
...
             [status-im.i18n.i18n :as i18n]
             [status-im.react-native.resources :as resources]))
 
 (def community-data
   {:id             constants/status-community-id
    :name           "Status"
    :description    "Status is a secure messaging app, crypto wallet and web3 browser built with the state of the art technology"
-   :status         "gated"
+   :status         :gated
    :section        "popular"
```
flexsurfer: would be great to change section as well

flexsurfer (approved):

ilmotta:
Some of the files need to be reindented (specially maps), but otherwise LGTM.

src/status_im/ui/screens/communities/styles.cljs
J-Son89: curious, are we following some specific best practices?

imo it's better to know what value I'm passing in to the function
```
(community-cover-container 48)
vs (community-cover-container :height 48)
```
In which case I have more knowledge of what's being passed. 

So wondering what/ why we would just pass the value directly? Very happy to hear because I'm new to the Clojure ecosystem :)

flexsurfer: `imo it's better to know what value I'm passing in to the function` thats right, thats why its better to have `(container (:height obj))` instead of `(container obj)`

J-Son89: ok cool, gotcha :)

ilmotta: I still think in most cases, passing a map vs _positional args_ is a matter of **personal preference**, and no any option is objectively better than the other. In some cases positional args are more convenient for threading/piping functions, but they can also get wild in many cases. Both have their merits. For example, many people prefer `(some-function {:do-this true :do-that false})` over `(some-function true false)` because it's easier for them to read the code and know what's being passed around, as @J-Son89 mentioned.

Let's also remember that Clojure 1.11 improved support for keyword args, which blurs the line between positional vs map args, but I believe we can't yet use Clojure 1.11 (?). See https://clojure.org/news/2022/03/22/clojure-1-11-0.

PR #13896, opened 2022-08-30
[13857] Extend integration tests to recover multiaccount from seed ph…

erikseppanen:
fixes #13857 

### Testing notes
- Can skip manual QA
- Can skip e2e tests

status: ready

PR #13926, opened 2022-09-05
ci: use pipefail to actually catch test errors

jakubgs:
Because otherwise when piping optput to `tee` the exit code that affects the result of the whole `sh` call is the last command in the pipe.


flexsurfer (approved):

jakubgs:
The test failures have been introduced in:

* https://github.com/status-im/status-mobile/pull/13896

By adding a broken `recover-multiaccount-test` test:

PR #13930, opened 2022-09-05
hooks: use AWK instead of jq for better portability

### LINKED ISSUE #13322: Add `jq` to Nix configuration
# Problem

rebasing status-go changes run some git hooks that depend on `jq` package but it is not present in the current Nix configuration and therefore requires manual installation.

jakubgs:
Some systems don't have jq installed, and using something like `nix-shell` in the shebang would make this script noticeably slower.

We're not using `grep` because it lacks `-P` flag on MacOS.

Also added a fix for `n` showing up in the link on MacOS because MacOS `sed` just needs newlines to be escaped differently.

Resolves: https://github.com/status-im/status-mobile/issues/13322

### DIFF
--- a/scripts/hooks/prepare-commit-msg
+++ b/scripts/hooks/prepare-commit-msg
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -eo pipefail
 
 if [[ -z $(git diff --cached --name-only -r | grep status-go-version.json) ]]; then
   exit 0 # No status-go-version.json changes found.
 fi
+
+# MacOS sed is garbage and interprets newlines differently.
+[[ "$(uname)" -eq "Darwin" ]] && NL=$'\\\n' || NL="\n"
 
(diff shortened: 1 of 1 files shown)


jakubgs:
Oh for fuck's sake... `grep` on Macs doesn't implement the Perl regexp:
```
jenkins@maci7-03.ms-eu-dublin.ci.devel:~/status-mobile % grep -Po '"commit-sha1": "\K\w+' status-go-version.json
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```
Why does it always have to be something...

yqrashawn (approved):

audriu (approved):

yqrashawn (approved):

audriu (approved):

Issue #13793, opened 2022-08-10
Improve quo2 text component


Parveshdhull:
### 1. letter-spacing
Letter spacing currently used by the component doesn't match with letter spacing for Figma [design](https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Mobile-Foundations?node-id=617%3A208).
And looks like small changes in letter spacing are not affecting the text component.
Formula used by Figma for letter spacing: [formula](https://rsms.me/inter/dynmetrics/) ([ref](https://github.com/status-im/status-mobile/pull/13753#discussion_r941192544))
`Heading-1/Regular`
<img src="https://user-images.githubusercontent.com/17097240/184025017-e8e663ca-6032-49b1-8080-486a5dd2578d.png" alt="Screenshot" height="300"/>

### 2. Line Height:
The line height currently used by the component doesn't match with line height from the Figma design
`Paragraph-2/Regular`
<img src="https://user-images.githubusercontent.com/17097240/184026205-c64ef274-85a9-4e49-baf6-f6bb83a560b7.png" alt="Screenshot" height="300"/>

[Original image]
<img src="https://user-images.githubusercontent.com/17097240/184026699-c641aabc-19d2-4129-8beb-292343f4141d.png" alt="Screenshot" height="200"/>

yqrashawn:
> 2. Line Height:

H @Parveshdhull I compared `Paragraph-2/Regular` and can't find line height difference.

![CleanShot 2022-09-09 at 14 35 23@2x](https://user-images.githubusercontent.com/15090582/189289396-706706b9-45e2-4c37-a767-06519de3331a.png)

Parveshdhull:
hi @yqrashawn,
Thank you very much for looking into this.
Please can you try with 5-10 lines? Small line heights difference might not be visible, but they add up to a big difference with multiple lines.

yqrashawn:
<img width="1015" alt="image" src="https://user-images.githubusercontent.com/15090582/189304171-6fcff891-dfe9-432a-bbd6-2ccb75b1c514.png">

There's indeed mismatch. I'll check.

Line 1
<img width="711" alt="image" src="https://user-images.githubusercontent.com/15090582/189305268-3bab99fa-a0f2-4375-9dbf-4ad92b87c389.png">


Line 10
<img width="711" alt="image" src="https://user-images.githubusercontent.com/15090582/189305130-e9b6fefd-2e99-455d-b1a1-c860a5637449.png">

[...]

PR #13960, opened 2022-09-09
fix: typography letter-spacing

yqrashawn:
related #13793

based on https://rsms.me/inter/dynmetrics/

status: ready

### DIFF
--- a/src/quo2/foundations/typography.cljs
+++ b/src/quo2/foundations/typography.cljs
@@ -1,25 +1,40 @@
 (ns quo2.foundations.typography)
 
-;; Formulat for letter spacing from figma %: font-size*{figma-percentage}/100
+(defn- tracking-fn
+  "Check inter dynamic metric https://rsms.me/inter/dynmetrics/"
+  [font-size]
+  (let [a -0.0223
+        b 0.185
+        c -0.1745
+        e js/Math.E]
+    (->> font-size
+         (* c)
+         (js/Math.pow e)
+         (* b)
+         (+ a)
+         (* font-size))))
+
+(def tracking (memoize tracking-fn))
+
 (def heading-1 {:font-size      27
                 :line-height    32
-                :letter-spacing -0.567})
+                :letter-spacing (tracking 27)})
 
-(def heading-2 {:font-size       19
-                :line-height     25.65
-                :letter-spacing  -0.304})
+(def heading-2 {:font-size      19
+                :line-height    25.65
+                :letter-spacing (tracking 19)})
 
 (def paragraph-1 {:font-size      15
                   :line-height    21.75
-                  :letter-spacing -0.135})
+                  :letter-spacing (tracking 15)})
 
 (def paragraph-2 {:font-size      13
                   :line-height    18.2
-                  :letter-spacing -0.039})
+                  :letter-spacing (tracking 13)})
 
 (def label {:font-size      11
             :line-height    15.62
-            :letter-spacing -0.055})
+            :letter-spacing (tracking 11)})
 
 (def font-regular {:font-family "Inter-Regular"}) ; 400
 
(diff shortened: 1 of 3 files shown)


src/quo2/foundations/typography.cljs
```diff
@@ -1,25 +1,40 @@
...
+         (* b)
+         (+ a)
+         (* font-size))))
+
+(def tracking (memoize tracking-fn))
+
 (def heading-1 {:font-size      27
                 :line-height    32
-                :letter-spacing -0.567})
+                :letter-spacing (tracking 27)})
```
Parveshdhull, line 21: Hi @yqrashawn, Thank you for the PR.
Did you tried comparing resulting output with figma? Because as I mentioned in issue, text component was ignoring small changes in `letter-spacing` value.
(For comparison, please type big line. Small letter spacing difference might not be visible, but it adds up to a big difference with several characters)

yqrashawn: image

Well, the `tracking` fn generates the same `letter-spacing` compaired with https://rsms.me/inter/dynmetrics/

I do compared with figma. 
In fact, I also compared figma and screen shots from https://rsms.me/inter/dynmetrics/ 
There're minor mismatches as well (numbers get rounded).

yqrashawn: The old calculation below is definitely incorrect. This commit fix it, yet can't remove the mismatch completely (numbers get rounded)
 https://github.com/status-im/status-mobile/blob/f1f9fa105ced7f4471311f4e6a2deace2302f722/src/quo2/foundations/typography.cljs#L3

Parveshdhull: >I do compared with figma.
In fact, I also compared figma and screen shots from https://rsms.me/inter/dynmetrics/
There're minor mismatches as well (numbers get rounded).

Did you mean mismatches between website and figma or screenshots and figma.
Please can you attach those comparison screenshots. (So its easier to understand)
(If possible, please also attach comparison of old-text, new-text, and figma, aligned)

src/quo2/foundations/typography.cljs
```diff
@@ -1,25 +1,40 @@
...
+  (let [a -0.0223
+        b 0.185
+        c -0.1745
+        e js/Math.E]
+    (->> font-size
+         (* c)
+         (js/Math.pow e)
+         (* b)
+         (+ a)
+         (* font-size))))
```
Parveshdhull, line 15: why `(* font size)`? I don't see this in formula.

src/quo2/foundations/typography.cljs
yqrashawn: @Parveshdhull check the below link for comparison between various screenshots and Figma design
https://www.figma.com/file/VHRoCTBxP7J7wwIbt3Uoto/typography-letter-spacing-%2313960

src/quo2/foundations/typography.cljs
yqrashawn: The formula you get is in unit `em` which is `-0.0031585560420509633` for `13px`, rounded to `0.3%` in Figma.
So `-0.0031585560420509633em` for `13px` which is why there's a `(* font-size)`

src/quo2/foundations/typography.cljs
Parveshdhull: Hi @yqrashawn,
Thank you very much for the Figma file. But I was hoping comparison of aligned texts to see the difference between letter spacing, of pr, and develop. As shown in this Figma
https://www.figma.com/file/8zjDY1vglXCpbTfc7rEsfa/Typography?node-id=0%3A1
(don't know if you can access file, so here is a [screenshot](https://user-images.githubusercontent.com/17097240/189548576-558fc1bd-6f83-4b79-ad0b-fa4b4f5132a1.png) of comparison)

So, it looks like both formulas are giving the same output. But I guess we can merge this PR as this is the official formula. But please keep the issue open as letter spacing still needs fixing. Thank you

Parveshdhull (approved):

PR #13977, opened 2022-09-12
Disable back button when migrating to a keycard 📱➡️ 💳.

### LINKED ISSUE #13943: Crash during onboarding / migration from regular account to Keycard after password confirmation and returning back
# Bug Report

## Problem 

There are several ways to reproduce the issue. Also, the app behaves differently on iOS and Android. 

### 1. New account creation on Android

**STR:**
1. Install Status
2. `Generate a new key` > Choose device as key storage > Create a password > Tap `Next` and quickly tap `Back`while processing is in progress
3. Choose device as key storage and create a password again
4. Tap `Next`

**Actual result:** app crashes with `java.lang.NullPointerException`

https://user-images.githubusercontent.com/67952253/188643626-526b50ab-0287-47bb-a74d-bc32cc132b20.mp4

### 2. Migration from a regular account to Keycard on Android

**STR:**
1. Install Status
2. Recover/create a regular account
3. `Profile` > `Privacy and security` > `Manage keys and storage` > `Log out`
4. Put a tick `Move keystore file` > `Enter seed phrase` > `Choose a storage` > `Keycard` > `Confirm`
5. Enter password to move your data > `Confirm`
6. Tap `Back` on the `Store your keys on Keycard` screen
7. Select `Keycard` and enter the password again > `Confirm` > Complete the onboarding flow

**Actual result:** app crashes with `java.lang.NullPointerException` after Keycard passcode confirmation

### 3. Migration from a regular account to Keycard on iOS
Most problematic one

There are the same steps to reproduce as in the previous scenario but the resulting behavior is different.
You can finish the migration process, but if you log out of the app once and try to log in again, you will get `Looks like you've tapped a wrong keycard` message.
The `lost or frozen card?` flow doesn't help either - when you try to log in after recovery, the multi-account "loses" the username and changes the user's picture (see video below). Thus, login is not possible until the application is reinstalled.

https://user-images.githubusercontent.com/67952253/188650692-bae1873a-2036-47ec-a597-af27af01c9c8.MP4

### Additional Information

- Status version: nightly 1.19

audriu:
fixes #13943

### Summary
Disable the back button when migrating to keycard on android.

#### Platforms
- Android
- iOS

##### Functional
- account recovery
- new account

### Steps to test
- Open Status
- follow [these steps](https://github.com/status-im/status-mobile/issues/13943#:~:text=Keycard%20on%20Android-,STR%3A,-Install%20Status)
- see how it all works perfectly

status: ready

### DIFF
--- a/src/status_im/navigation/core.cljs
+++ b/src/status_im/navigation/core.cljs
@@ -113,7 +113,7 @@
 (defn set-view-id [view-id]
   (log/debug "set-view-id" view-id)
   (when-let [{:keys [on-focus]} (get views/screens view-id)]
-    (re-frame/dispatch [:set :view-id view-id])
+    (re-frame/dispatch [:set-view-id view-id])
     (re-frame/dispatch [:screens/on-will-focus view-id])
     (when on-focus
       (re-frame/dispatch on-focus))))
--- a/src/status_im/events.cljs
+++ b/src/status_im/events.cljs
@@ -279,6 +279,11 @@
   [{:keys [db]} k v]
   {:db (assoc db k v)})
 
+(fx/defn set-view-id
+  {:events [:set-view-id]}
+  [{:keys [db]} view-id]
+  {:db (assoc db :view-id view-id)})
+
 ;;TODO :replace by named events
 (fx/defn set-once-event
   {:events [:set-once]}
--- a/src/status_im/keycard/backup_key.cljs
+++ b/src/status_im/keycard/backup_key.cljs
@@ -4,6 +4,7 @@
             [status-im.utils.utils :as utils]
             [status-im.i18n.i18n :as i18n]
             [status-im.ethereum.mnemonic :as mnemonic]
+            [status-im.keycard.common :as common]
             [status-im.multiaccounts.recover.core :as multiaccounts.recover]
             [status-im.navigation :as navigation]
             [status-im.signing.core :as signing.core]
@@ -64,4 +65,5 @@
                       (assoc-in [:keycard :flow] :recovery)
                       (update :multiaccounts/key-storage dissoc :seed-phrase))
              :dismiss-keyboard nil}
+            (common/listen-to-hardware-back-button)
             (navigation/navigate-to-cofx :keycard-onboarding-intro nil)))
--- a/src/status_im/keycard/core.cljs
+++ b/src/status_im/keycard/core.cljs
@@ -592,19 +592,6 @@
   [{:keys [db]} listeners]
   {:db (update-in db [:keycard :listeners] merge listeners)})
 
-(defn onboarding-intro-back-handler
-  "The back button handler is used to manage device back press.
-
-  If the handler returns false, the back button functions as usual (ie. dispatchs GO_BACK event).
-  If it returns true, the back button becomes inactive.
-
-  We want to deactivate the back button when the user comes from key-storage and migration flow."
-  []
-  (-> @re-frame.db/app-db
-      :keycard
-      :from-key-storage-and-migration?
-      boolean))
-
 (fx/defn ui-recovery-phrase-cancel-pressed
   {:events [:keycard.ui/recovery-phrase-cancel-pressed]}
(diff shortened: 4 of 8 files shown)


flexsurfer:
could you elaborate on why only partially fix?

audriu:
Because it only fixes case 2 of that ticket. I can't reproduce case 1 and for case 3 I need an iPhone - which is nowhere around me. Maybe someone with an iPhone can continue...

src/status_im/navigation/core.cljs
```diff
@@ -192,7 +192,7 @@
                                               :keycard-pairing]))]
     {:init-root-fx (if keycard-account? :multiaccounts-keycard :multiaccounts)}))
 
-(defonce rset-app-launched
+(defonce reset-app-launched
```
flexsurfer: it should be `reg` or `register`, also `rset-lazy-reg` should be `set-lazy-reg`

flexsurfer (commented):
im not sure i understand how this PR fixes the issue, issue says you need to press back button, not a hardware ? could you elaborate on changes in this PR please @audriu

src/status_im/navigation/core.cljs
audriu: Maybe we can leave as it is as it functions the same.

audriu:
Back button is also considered to be Hardware back button

audriu:
The back button is also considered to be a Hardware back button in navigations documentation:
https://reactnative.dev/docs/backhandler
So I assumed that those terms are equal...

flexsurfer:
i mean the back button in topbar and hardware back button have different handlers , so i still not sure how this fix works

audriu:
This will only fix for hardware button. This only fixes for Android.

audriu:
ah yes, just realized that I could use the simulator. since this could be solved without using the keycard in this case.

flexsurfer:
so it won't solve case 2 ? or @qoqobolo replicated it only for hardware button ?

qoqobolo:
>  or @qoqobolo replicated it only for hardware button ?

No, case 2 is replicated with the app `Back` button. So PR should fix it as well.

audriu:
@flexsurfer do you think that code is ok so it could be manually tested?

qoqobolo:
Smoke tested case 2 of [the issue](https://github.com/status-im/status-mobile/issues/13943) on Android device - the error has not been completely fixed; only the hardware back button has been disabled, the issue is still reproducible when using the app back button.

@audriu please, bring this PR back for review as soon as case 2 is fully fixed.
Thanks.

cammellos:
@audriu makes sense to fix both cases, just one won't provide much benefits, I agree with @qoqobolo , thank you

audriu:
I know but this requires a physical iPhone to use with a keycard. But maybe this will fix both cases because the back button is disabled now.

audriu:
I think it now fixes both cases.

flexsurfer:
hey @audriu i still can't see how the back button in UI is fixed , no related changes in the code

qoqobolo:
@flexsurfer not sure how it should look like in the code, but in fact, the back button is disabled 

Screenshot 2022-09-20 at 10 49 22

audriu:
@flexsurfer The back button is disabled. 
Screenshot 2022-09-20 at 11 46 25
by this code https://github.com/status-im/status-mobile/blob/13943-crash-during-onboarding-migration-from-regular-account-to-keycard-after-password-confirmation-and-returning-back/src/status_im/ui/screens/screens.cljs#L751

So if it is not possible to go back - the bug does not happen.
It was done like that in the original keycard migration implementation https://github.com/status-im/status-mobile/pull/11526/files

flexsurfer (approved):

flexsurfer:
oh now i get it, thanks, so there were native topbar but we don't need to show it

qoqobolo:
Thanks @audriu, PR is tested and can be merged.
E2e failures are not related to PR.

PR #13986, opened 2022-09-12
fix frozen keycard account recovery using seed phrase

### LINKED ISSUE #13979: "Seed does not match" error in case of restoring access with seed phrase when keycard is frozen in "Recover keys" flow
# Bug Report

## Problem 
If keycard gets frozen in "Recover keys" flow, you can't restore access from the seed phrase in case it is not your first account on device.
Non-relevant error appears: `Seed doesn't match` 1st account on the device
The only way is to start the process over using `Restore with seed phrase` flow

#### Expected behavior
Can recover access via seed phrase 

#### Actual behavior
`Seed doesn't match userA` 

https://user-images.githubusercontent.com/4557972/189664459-3cc5412a-bb94-4875-bb9c-4c0ed169170d.mp4

### Reproduction

1) Device A: open Status
2) Device A: create some multiaccount (`userA`)
3) Device B: create multiaccount and store the keys on keycard, write down seed phrase (`userB`)
4) Device A: log out from `userA` account
5) Device A: `Recover with Keycard`> enter the wrong PIN 3 times > restore with seed > enter seed from `userB` account

### Additional Information

- Status version: nightly 

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/13979

### Testing Request:
- Please also test account recovery using seed phrase and storing to keycard
- Move keystore file to keycard

status: ready

### DIFF
--- a/src/status_im/keycard/recovery.cljs
+++ b/src/status_im/keycard/recovery.cljs
@@ -224,11 +224,13 @@
               (multiaccounts.model/logged-in? cofx)
               (navigation/set-stack-root :profile-stack [:my-profile :keycard-settings])
 
-              (:multiaccounts/login db)
+              (= (get-in db [:multiaccounts/login :key-uid]) (get-in db [:keycard :application-info :key-uid]))
               (return-to-keycard-login)
 
               :else
-              (navigation/set-stack-root :onboarding [:get-your-keys]))))
+              (if (seq (get db :multiaccounts/multiaccounts))
+                (navigation/pop-to-root-tab :multiaccounts)
+                (navigation/set-stack-root :onboarding [:get-your-keys])))))
 
 (re-frame/reg-fx
  ::finish-migration
--- a/src/status_im/multiaccounts/key_storage/core.cljs
+++ b/src/status_im/multiaccounts/key_storage/core.cljs
@@ -105,19 +105,22 @@
 (fx/defn seed-phrase-validated
   {:events [::seed-phrase-validated]}
   [{:keys [db] :as cofx} validation-error]
-  (let [error? (-> validation-error
-                   types/json->clj
-                   :error
-                   string/blank?
-                   not)
-        onboarding? (not (or (:multiaccounts/login db) (:multiaccount db)))]
+  (let [error?      (-> validation-error
+                        types/json->clj
+                        :error
+                        string/blank?
+                        not)
+        onboarding? (not (or (:multiaccounts/login db) (:multiaccount db)))
+        importing?  (= :import (get-in db [:keycard :flow]))]
     (if error?
       (popover/show-popover cofx {:view :custom-seed-phrase})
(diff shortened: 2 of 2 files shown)


flexsurfer (approved):

churik:
@Parveshdhull still can reproduce initial issue on IOS build https://i.diawi.com/8jf435:

https://user-images.githubusercontent.com/4557972/189868395-bc095a88-96c9-4a3e-84ea-f72e189b00e5.mp4

However, it is fixed on Android

churik:
@Parveshdhull 
You're right, I misunderstood the feature, I supposed if we have any mnemonic we can use this flow, as anyway it requires a full card reset.
Currently, I can finish flow, but basically, after recovering app is logging in to the wrong multi-account (which is 1st on device), which IMO is more dangerous.

https://user-images.githubusercontent.com/4557972/189880943-0c47de15-0037-4c7f-bb9a-61a2dd44bd9e.mp4


### Reproduction

1) Device A: open Status
2) Device A: create some multiaccount (`userA`, can login with password)
3) Device B: create multi-account and store the keys on a keycard, write down the seed phrase (`userB`)
4) Device A: log out from `userA` account
5) Device A: `Recover with Keycard`> enter the wrong PIN 3 times > restore with seed > enter seed from `userB` account
6) Recover access

#### Expected behavior
Can see that `UserB` is recovered, can log out and see 2 multiaccounts  (`UserA` and `UserB`)

#### Actual behavior
Looks like it logged in to multiaccount of `UserA`, can't login with keycard (You tapped the wrong card)
If relaunch app - can see only multi-account of `userA`, can login with password, no keycard accounts on device.

#### Additional Information

Parveshdhull:
Hi Tetiana,
Thank you very much for testing the PR and finding this issue. It was also a pre-existing issue, but we were unable to reproduce it because keycard recovery using mnemonics was not working.

### Issue Cause:
On keycard account login if we use the wrong PIN and keycards get frozen, then we can recover it using mnemonics and after recovery, it returns to that login screen.
We are using this same function to recover the keycard using mnemonics if it gets frozen while importing an account from the keycard, and that's why after the recovery process it again returned to the login screen. (even though the login screen had a different account)

### Current Solution:
As in the above case where we are recovering a frozen keycard on the login screen, we are not recreating the account because it already exists. So the current solution is using the same functionality, it is not creating an account only recovering the keycard.
If a keycard is recovered on the login screen it will return to the login screen and if a keycard is recovered in the import screen, it will return to `your keys` screen, where it will be clear that the account is not created, and user have to do that separately.

I tried quickly adding the functionality of creating an account with keycard recovery but didn't get the expected results. We can either do that separately or cover this case with redesign.

Please let me know if this works for you.

churik:
Thank you for detailed explanation!
We can just leave the issue as it is, as it is quite a rare flow; but I'd like to keep the issue open just to consider that for re-design

PR #14006, opened 2022-09-14
fix quo2 colors

Parveshdhull:
status: ready


src/quo2/foundations/colors.cljs
```diff
@@ -17,15 +17,16 @@
 ;;Solid
 (def neutral-5 "#F5F6F8")
 (def neutral-10 "#F0F2F5")
-(def neutral-20 "#ECEEF1")
-(def neutral-30 "#E7EAEE")
+(def neutral-20 "#E7EAEE")
+(def neutral-30 "#DCE0E5")
 (def neutral-40 "#A1ABBD")
```
Parveshdhull, line 23: Changed

src/quo2/foundations/colors.cljs
```diff
@@ -185,95 +186,130 @@
...
-;; Customization
-(def purple-20 (alpha "#8661C1" 0.2))
-(def purple-50 "#8661C1")
-(def purple-60 (alpha "#8661C1" 0.6))
-(def indigo-20 (alpha "#496289" 0.2))
-(def indigo-50 "#496289")
-(def indigo-60 (alpha "#496289" 0.6))
-(def turquoise-20 (alpha "#448EA2" 0.2))
-(def turquoise-50 "#448EA2")
-(def turquoise-60 (alpha "#448EA2" 0.6))
```
Parveshdhull: These values are wrong, (ex. `:purple-20` represents different color, not opacity)

src/quo2/foundations/colors.cljs
```diff
@@ -185,95 +186,130 @@
...
-(def brown-60 (alpha "#805141" 0.6))
-(def beige-20 (alpha "#CAAE93" 0.2))
-(def beige-50 "#CAAE93")
-(def beige-60 (alpha "#CAAE93" 0.6))
+;;;; Customisation
+
+(def customisation {:purple    {10 "#FAF9FD"
+                                20 "#E7DFF3"
+                                30 "#CABAE4"
+                                40 "#AD94D5"
```
Parveshdhull: Why Map?
There are several color customizations possible, user account, community, wallet etc.. And according to that selected customization color, we have to use different variations of that color for different combinations.

For. ex: (if selected color :green)
Profile: green-50
Cover: green-30-opa-20
community-title color etc.

src/quo2/foundations/colors.cljs
```diff
@@ -185,95 +186,130 @@
...
+     (custom-color suffix-dark opacity-dark)
+     (custom-color suffix-light opacity-light))))
 
 (def shadow  "rgba(9,16,28,0.04)")
 
-(def customization
-  {:dark {:purple purple-60
-          :indigo indigo-60
-          :turquoise turquoise-60
-          :blue blue-60
```
Parveshdhull: customization don't have different colors as per theme in figma. (this map only covers a use case for a component, so probably should be in that component or use avove function
`(custom-color-by-theme color 50 60`

Parveshdhull:
PR is still in WIP but can use an early review of the approach.

Should I also convert other colors into the map, for example `:primary`, because it can also be set as the theme of the account?

<details>
  <summary>Other Approach</summary>

Also tried
```
color @(resolve (symbol (str color "-" suffix)))
```
Apparently this only works with Clojure, cljs expects the argument of to be quoted. (And calling `(quote ...)` threw some more errors)
(but not sure if we even need to consider this approach)

</details>

cc: @flexsurfer @J-Son89

src/quo2/foundations/colors.cljs
```diff
@@ -185,95 +186,130 @@
...
+                                60 "#AA927C"
+                                70 "#897664"
+                                80 "#695A4C"
+                                90 "#493F35"}})
+(defn custom-color
+  ([color suffix]
+  (custom-color color suffix nil))
+  ([color suffix opacity]
+   (let [base-color get-in customisation [(keyword color) suffix]]
+     (if opacity (alpha base-color opacity) base-color))))
```
Parveshdhull: **TODO**: memoize `custom-color` function

src/quo2/foundations/colors.cljs
```diff
@@ -185,95 +186,130 @@
...
+                                50 "#CAAE93"
+                                60 "#AA927C"
+                                70 "#897664"
+                                80 "#695A4C"
+                                90 "#493F35"}})
+(defn custom-color
+  ([color suffix]
+  (custom-color color suffix nil))
+  ([color suffix opacity]
+   (let [base-color get-in customisation [(keyword color) suffix]]
```
flexsurfer: how it works without `( )` ?

Parveshdhull: oh sorry, I was playing around with different approach `@(resolve ...)` and in the end missed brackets while pushing code. will fix it. thank you

src/quo2/screens/avatars/group_avatar.cljs
```diff
@@ -20,11 +20,9 @@
                   :options
                   (map
                    (fn [c]
-                     {:key c
+                     {:key   c
                       :value c})
```
ibrkhalil, line 24: Interestingly enough no errors ..  Passing `keyword`s when quo.text expects a `string`

ibrkhalil (approved):
lgtm

J-Son89 (commented):
looks good to go 👍

Issue #13991, opened 2022-09-13
Block PR merging when there is Clojure compiler warnings


Parveshdhull:
## Summary
Currently, we can merge PRs even when there are warnings in `make run-clojure`. This even allows us to merge PR even when code is missing.

#### For ex:
In commit https://github.com/status-im/status-mobile/commit/f4bb6ec7d65cb37faeb73c2b4dccf8924375ae40 [lowest_price.cljs](https://github.com/status-im/status-mobile/commit/f4bb6ec7d65cb37faeb73c2b4dccf8924375ae40#diff-9b3de11f108b13ccb26a62b1fc65050065ead8a4355505a69f2cdd51dac5106f) file is empty and it is being imported and used in [main.cljs](https://github.com/status-im/status-mobile/commit/f4bb6ec7d65cb37faeb73c2b4dccf8924375ae40#diff-149c0f76f53044b85c50c3613bc31d8f96efb1065473e302d0acc8aae3241a7a). So we are using component, which don't even exist.
which creates warning's in `make run-clojure`
```
shadow-cljs - nREPL server started on port 7888
shadow-cljs - watching build :mobile
[:mobile] Configuring build.
[:mobile] Compiling ...
[:mobile] Build completed. (861 files, 111 compiled, 1 warnings, 5.61s)

------ WARNING #1 - :undeclared-var --------------------------------------------
 File: /home/parvesh_monu/Desktop/ArchFiles/React/status-mobile/src/quo2/screens/main.cljs:64:23
--------------------------------------------------------------------------------
  61 |            :component information-box/preview-information-box}
  62 |           {:name      :lowest-price
  63 |            :insets    {:top false}
  64 |            :component lowest-price/preview-lowest-price}]
-----------------------------^--------------------------------------------------
 Use of undeclared Var quo2.screens.info.lowest-price/preview-lowest-price
--------------------------------------------------------------------------------
  65 |    :markdown [{:name      :texts
  66 |                :insets    {:top false}
  67 |                :component text/preview-text}]
  68 |    :messages [{:name      :gap
--------------------------------------------------------------------------------
```

(missing code is fixed in different PR https://github.com/status-im/status-mobile/pull/13985)

Parveshdhull:
fixed in https://github.com/status-im/status-mobile/pull/14014 (`clj-kondo` needed to be upgraded)

PR #13987, opened 2022-09-12
[#13895] Create new Activity Center screen

ilmotta:
fixes #13895
Related to #13967  

### Summary

Create new Activity Center screen, roughly displaying read or unread notifications.

<table>
  <thead>
    <tr>
      <th>
        Screenshot
        <p><strong>Unread notifications</strong></p>
      </th>
      <th>
        Screenshot
        <p><strong>Read notifications</strong></p>
      </th>
    </tr>
  </thead>
</table>

### Review notes

This PR should not affect production code. If you find anything, please let me know.

The design and app behavior in this PR should not be considered final, quite the contrary. Its goal is to **do the basic plumbing and reduce the amount of code to be reviewed in follow-up PRs**. Examples of things that will soon be addressed:

- The Design team is still working out how to better display read/unread notifications. They will probably be displayed as clickable filters, so in this PR I implemented something ugly, but which resembles the functionality.
- The activity logs have hardcoded and unformatted data.

For now, the Activity Center issues are being tracked with the label https://github.com/status-im/status-mobile/labels/Activity%20center.

### Testing notes

I decided to keep the new screen inaccessible from the UI, given how crude and non-functional it still is. Nonetheless, if you'd like to test it, you'll need to checkout the Status Go draft branch [activity-center-read-unread](https://github.com/status-im/status-go/tree/feature/activity-center-read-unread).

Remember to run `export STATUS_GO_SRC_OVERRIDE=$HOME/<path-to-your-status-go-repo>` before running targets such as `make run-android`.

Also apply the following diff to status-mobile:

```diff
modified   src/status_im/ui/screens/home/views.cljs
@@ -270,7 +270,7 @@
                           :accessibility-label :notifications-button
                           :on-press #(do
                                        (re-frame/dispatch [:mark-all-activity-center-notifications-as-read])
-                                       (re-frame/dispatch [:navigate-to :notifications-center]))}
+                                       (re-frame/dispatch [:navigate-to :activity-center]))}
       [icons/icon :main-icons/notification2 {:color (quo2.colors/theme-colors quo2.colors/black quo2.colors/white)}]]
      (when (pos? notif-count)
        [react/view {:style (merge (styles/counter-public-container) {:top 5 :right 5})
```

#### Platforms

- Android
- iOS

status: ready

PR #14014, opened 2022-09-15
[#13964 #13962] Display contact requests

### LINKED ISSUE #13962: Allow user to see the status of their contact request
# Feature Issue

### User Story

>As a user
>I want to be able to see the status of my contact request
>So that I know whether the other end has replied

### Description

This feature is part of a larger effort to create a new Activity Center. Please, reach out to members of the UX team before starting this work in case there are important details not apparent in Figma.

### LINKED ISSUE #13964: Display contact requests
# Feature Issue

### User Story

>As a user
>I want to be able to see contact requests in the Activity Center
>So that I know someone added me to their contacts

### Description

This feature is part of a larger effort to create a new Activity Center. Please, reach out to members of the UX team before starting this work in case there are important details not apparent in Figma.

- Double-check with UX team how the loading indicator should be displayed.

ilmotta:
fixes #13962
fixes #13964

### Summary

This PR adds support for Contact Request notifications in the new Activity Center (read-only at the moment, i.e. non-interactive).

<table>
  <thead>
    <tr>
      <th>
        Screenshot
        <p><strong>Contact requests in various states</strong></p>
      </th>
    </tr>
  </thead>
</table>

### Review notes

- Upgrades clj-kondo to the latest version. The original version was pretty old, and it was the cause of incorrect static analysis that generated errors in the CI. For an example, see these [logs in Jenkins]( https://ci.status.im/blue/rest/organizations/jenkins/pipelines/status-mobile/pipelines/prs/pipelines/tests/branches/PR-14014/runs/4/nodes/35/log/?start=0).
- The user can't yet interact with notifications. See some of the remaining issues in https://github.com/status-im/status-mobile/labels/Activity%20center.
- The new `status-tag` component was improved to accept custom labels, otherwise it was only capable of displaying fixed text, like "positive", "negative", etc.
- Figma flows https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926
- The `utils.datetime` namespace was reformatted to use the widely accepted convention in Lisps (and Clojure) to separate top-level sections with 3 or 4 backticks. See (https://github.com/bbatsov/clojure-style-guide#heading-comments). On a side note, if you use Emacs, you can enable the `outline-minor-mode` to automatically get collapsible sections (as in Org mode).

<table>
  <thead>
    <tr>
      <th>
        Screenshot
        <p><strong>Lisp heading comments can be collapsed in Emacs</strong></p>
      </th>
    </tr>
  </thead>
</table>

### Testing notes

Just like the merged PR https://github.com/status-im/status-mobile/pull/13987, this one doesn't yet replace the old Notifications Center with the new Activity Center. This will only be done once the Status Go [draft branch](https://github.com/status-im/status-go/tree/feature/activity-center-read-unread) is merged, but there are still some outstanding tests to be done.

Meanwhile, you need to do the following:

- Checkout status-go branch `feature/activity-center-read-unread`.
- Always run `export STATUS_GO_SRC_OVERRIDE=$HOME/<path-to-your-status-go-repo>` before running targets such as `make run-android`.
- Apply the following patch to status-mobile:

```diff
modified   src/status_im/ui/screens/home/views.cljs
@@ -270,7 +270,7 @@
                           :accessibility-label :notifications-button
                           :on-press #(do
                                        (re-frame/dispatch [:mark-all-activity-center-notifications-as-read])
-                                       (re-frame/dispatch [:navigate-to :notifications-center]))}
+                                       (re-frame/dispatch [:navigate-to :activity-center]))}
       [icons/icon :main-icons/notification2 {:color (quo2.colors/theme-colors quo2.colors/black quo2.colors/white)}]]
      (when (pos? notif-count)
        [react/view {:style (merge (styles/counter-public-container) {:top 5 :right 5})
```

### Steps to test

You can follow the use cases below if you'd like to do some manual testing. Please, don't consider this an exaustive list of use cases. In all cases, Bob has enabled _accept chat from anyone_ in his app settings.

>Given that Alice has not added Bob as a contact,
>When Alice opens a chat with Bob by using his chat key,
>And Alice sends her first message to him,
>Then Bob sees the last message from Alice in his Activity Center (aka 1:1 notification).

>Given that Alice already sent a message to Bob,
>When Alice opens the chat with Bob,
>And when Alice clicks on Add Contact,
>Then Bob receives a notification in the Activity Center,
>And Bob can either accept or decline the contact request.

>Given that Bob accepted a contact request from Alice,
>When Bob removes Alice from his contact list,
>Then Bob doesn't see notifications related to Alice in the Activity Center.

>Given that Bob accepted a contact request from Alice,
>When Bob opens the Activity Center,
>Then he sees a notification saying he accepted the contact request from Alice.

>Given that Bob declined a contact request from Alice,
>When Bob opens the Activity Center,
>Then he sees a notification saying he declined the contact request from Alice.

>Given that Bob gave a nickname to Alice (only he sees),
>When Bob opens the Activity Center,
>Then he sees all references to Alice replaced with her nickname.

status: ready


src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -5,26 +5,88 @@
...
+(defn activity-title
+  [{:keys [type]}]
+  (case type
+    constants/activity-center-notification-type-contact-request
+    (i18n/label :t/contact-request)
+
+    constants/activity-center-notification-type-one-to-one-chat
+    "Dummy 1:1 chat title"
+
+    "Dummy default title"))
```
ilmotta, line 24: FYI: More cases for different notification types will be covered as the new Activity Center issues are played. This is true for this function and for others in this namespace.

src/quo2/screens/tags/status_tags.cljs
```diff
@@ -27,16 +27,20 @@
...
-      [rn/view {:margin-bottom 50
-                :padding       16}
-       [rn/view {:flex 1}
-        [preview/customizer state descriptor]]
-       [rn/view {:padding-vertical 60
-                 :flex-direction   :row
-                 :justify-content  :center}
-        [quo2/status-tag @state]]])))
+      (let [props (cond-> @state
+                    (= :positive (:status @state)) (assoc :status {:label "Positive" :type :positive})
```
flexsurfer, line 33: why its a duplicate value from tags-options ? also if we show this to users it must be i18n

J-Son89: I think this is okay. The `Status Tags` text was hardcoded before but needs to be dynamic. I noticed similar in some work I was doing.

And the text here is only for quo2 component library so it's probably okay to not have it i18n, right?

src/status_im/utils/datetime.cljs
```diff
@@ -10,6 +10,8 @@
             [clojure.string :as s]
             [status-im.goog.i18n :as goog.18n]))
 
+;;;; Datetime constants
```
J-Son89, line 13: nit: are these really constants? 🤔

src/status_im/utils/datetime.cljs
```diff
@@ -29,28 +31,54 @@
...
 
-;; date-time formats
+(defn- datetime-within-one-week-format
+  [^js locsym]
+  (if (is24Hour locsym)
+    "E HH:mm"
+    "E h:mm a"))
+
+(defn- medium-date-format
+  "Get medium format from current locale symbols."
```
J-Son89, line 76: what is medium format? 🤨

src/status_im/utils/datetime.cljs
```diff
@@ -97,6 +119,35 @@
...
 
+(defn timestamp->relative [ms]
+  (let [datetime       (from-long ms)
+        datetime-local (plus datetime time-zone-offset)
+        today          (minus (t/today-at-midnight) time-zone-offset)
+        yesterday      (minus today (days 1))
+        six-days-ago   (minus today (days 6))]
+    (cond
+      ;; Previous years.
+      (< (t/year datetime) (t/year today))
```
J-Son89: consider putting these into helper functions and remove comments.
```
(defn previous-years? [datetime today datetimelocal]
(< (t/year datetime) (t/year today))
      (.format ^js (date-fmt) datetime-local))
```
same for below.

src/quo2/screens/tags/status_tags.cljs
flexsurfer: oh right , its a quo screen, thanks

src/status_im/utils/datetime.cljs
ilmotta: @J-Son89, if you look at the diff, you'll see that I tried only to reformat and move the original comment to the docstring. But ideally we could do a better job at describing some of these things.

src/status_im/utils/datetime.cljs
ilmotta: You're right, most of them are constants (time units), but others are not. I would gladly take a suggestion here.

src/status_im/utils/datetime.cljs
ilmotta: Good suggestion, I'll do my best to see if this can be refactored. Just so you understand, the reason why I chose to keep things like this is because every subsequent check in the `cond` macro only works because of the previous checks.

For example, the correct way to check if a datetime is yesterday is not just `(before? datetime today)`. I hope you get it. I'll see what I can come up with, thanks.

J-Son89: yeah I get that, I mean just use named functions for the checks rather than comments. You can still keep the function `defn timestamp->relative [ms]` as is

ilmotta: It was quite more involved than a simple extraction, especially because I had to make sure the new implementation is correct. The new extracted functions are now useful on their own, and are correct, independent of the context they're being used, which was not the case of the original code. I also removed other superfluous timezone-related code that wasn't really necessary. I'm marking this thread as resolved :)

Parveshdhull:
> Upgrades clj-kondo to the latest version.

hi @ilmotta, Thank you for upgrading `clj-kondo` this also fixes https://github.com/status-im/status-mobile/issues/13991

flexsurfer (approved):

J-Son89 (approved):

src/quo2/components/tags/status_tags.cljs
```diff
@@ -50,42 +50,48 @@
...
-             :icon :verified
-             :border-color colors/success-50-opa-20
-             :text-color (if (= theme :light) colors/success-50
-                             colors/success-60)
-             :label (or label (i18n/label :positive))}])
+             :icon             :verified
+             :border-color     colors/success-50-opa-20
+             :label            (or label (i18n/label :positive))
+             :text-color       (if (= theme :light) colors/success-50
+                                   colors/success-60)}])
```
Parveshdhull, line 62: hi @ilmotta, why these changes needed?

ilmotta: @Parveshdhull, what happened here is that I had improved the status tag component to accept a label, but someone also fixed it and merged before I did. So this morning I fixed the conflicts, but these changes remained after my rebase (map keys were sorted). They're harmless though

Parveshdhull (approved):

PR #14039, opened 2022-09-19
WIP Issue 13699 Implement Shell Scan screen

### LINKED ISSUE #14011: Implement Shell Scan screen
### Instructions

Implement the Shell Scan Screen.

Figma: https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?type=design&node-id=451-18994&t=RlXEGOyFRoEJCjEz-0

You should be re-using existing components where available and create components when not.

These are the following details as specified by the design team.
```
if web or dApp address QR is scanned --> user taken to browser

if a pairing QR is scanned --> user is taken into pairing flow

if a user profile QR is scanned --> user is taken to that user's profile screen

if a community QR is scanned --> user is taken to that community

if a channel in a community QR is scanned --> user is taken to that channel (if they can access it) or to the community home if they can't access that channel

if an ETH address is scanned --> user sees a choice to either send to that address or save that address.  Note if user's client is aware that this address is shared with them via showcase, more metadata can be displayed.

if a WalletConnect QR is scanned --> user enters WalletConnect connect flow

if a user scanner a QR that is non of the above --> an error should be displayed
```

Communities scanning will work like this:
Screenshot_2023-11-29_at_00 49 21

If this feature grows too large on initial implementation it is fine to notify and only handle 2/3 differrent types of QR codes provided the underlying code easily allows for extending the other features which we can then add in follow up issues.

ibrkhalil:
status: ready


Parveshdhull:
Please also mark PR as WIP in title

src/quo2/components/shell/scan.cljs
```diff
@@ -0,0 +1,106 @@
...
+(defn- viewfinder [size border-radius flashlight-on?]
+  [:f>
+   (fn []
+     [rn/view {:style viewfinder-port}
+      [rn/view {:style {:border-width 700
+                        :border-color "rgba(0,0,0,0.7)"}}
+       [rn/view {:style {:flex-direction :row :justify-content :space-between}}
+        [corner :top-left]
+        [corner :top-right]]
+       [reanimated/view {:style (reanimated/apply-animations-to-style
```
Parveshdhull, line 54: hi, why we are using reanimated here? I don't think we are animating scanner anymore.

ibrkhalil: Alisher said we are

flexsurfer:
stale, feel free to reopen later

Issue #14016, opened 2022-09-16
Implement Channel-List component


flexsurfer:
Figma: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=411%3A18564

J-Son89:
This component is done with design approval 👍

PR #14040, opened 2022-09-19
Jc/add channel list item

J-Son89:
Implements: https://github.com/status-im/status-mobile/issues/14016

see the designs here:
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=411%3A18564

ios & android:

### Steps to test
Open quo2 library


src/quo2/components/avatars/channel_avatar.cljs
```diff
@@ -40,6 +40,6 @@
           {:color           (if dark?
                               colors/neutral-40
                               colors/neutral-50)
-           :container-style {:width  16
-                             :height 16}
+           :container-style {:width  12
```
J-Son89, line 42: this lock icon was not matching the Figma designs correctly so I adjusted for both sizes and it now seems correct.

src/quo2/foundations/colors.cljs
```diff
@@ -13,9 +13,21 @@
       (let [rgb (string/split value #",")]
         (str (string/join "," (butlast rgb)) "," opacity ")")))))
 
+(defn theme-alpha ([light-color light-opacity dark-color dark-opacity]
```
J-Son89: discussing with Pedro in ux and it seems quite common where a community will provide a color and then we have to use that color with different opacity (can be different for dark and light).

Added this help function as it will be a common utility we need.

J-Son89:
Preview of screen
Screenshot 2022-09-20 at 00 03 57

flexsurfer (approved):

ibrkhalil (approved):
lgtm

src/quo2/components/list_items/channel.cljs
```diff
@@ -0,0 +1,47 @@
...
+     {:big?                   true
+      :lock-status            lock-status
+      :emoji-background-color (colors/theme-alpha channel-color 0.1 0.1)
+      :emoji                  emoji}]
+    [quo2.text/text {:style (merge {:margin-left 12} (when  (and (not (= lock-status :locked))  muted) {:color colors/neutral-40}))
+                     :weight :medium :size :paragraph-1} (str "# " name)]]
+
+   [rn/view {:style {:height 20}}
+
+    (when (and (not (= lock-status :locked))  muted) [quo2.icons/icon :main-icons2/muted {:size            20
```
ilmotta: [Readability] For code that can span multiple lines, it's more common to break the body of the `when` macro to the subsequent line of the conditional check. I find this line a tad too difficult to parse.

[Formatting] There's extra space after the `:size` and `:no-color` keywords.

src/quo2/components/list_items/channel.cljs
```diff
@@ -0,0 +1,47 @@
...
+                                                                                          :no-color         true}])
+    (when (and (not (= lock-status :locked)) (not muted)  (pos? (int mentions-count)))
```
ilmotta: [Suggestion] Clojure has a core function named `not=`. From the docstring:

```
Same as (not (= obj1 obj2))
```

src/quo2/components/list_items/channel.cljs
```diff
@@ -0,0 +1,47 @@
...
+      [rn/view {:style {:margin-right 2
+                        :margin-top 2}}
+
+       [quo2.counter/counter {:override-bg-color channel-color} mentions-count]])
+
+    (when (and (not (= lock-status :locked)) (not muted) (not (pos? (int mentions-count))) unread-messages?)
```
ilmotta: [Readability] I find this line almost impossible to parse in my head. If you break the lines the conditions will become much more obvious.

```clojure
(and (not (= lock-status :locked))
     (not muted)
     (not (pos? (int mentions-count)))
     unread-messages?)
```

src/quo2/components/list_items/channel.cljs
```diff
@@ -0,0 +1,47 @@
...
+
+      [quo2.icons/icon :main-icons2/channel-notification {:size            20
```
ilmotta: [Formatting] Extra space after the keywords.

src/quo2/foundations/colors.cljs
ilmotta: Nothing major or that needs to be done now, but the `theme-alpha` function could be memoized (simplest strategy, just stores in memory without any cache eviction strategy). I think for lower level abstractions that can get called quite a lot, memoization can be very advantageous.

ilmotta (approved):

src/quo2/components/list_items/channel.cljs
J-Son89: thanks! :)

src/quo2/components/list_items/channel.cljs
J-Son89: will adjust! 👍

src/quo2/components/list_items/channel.cljs
```diff
@@ -0,0 +1,57 @@
...
+      [quo2.icons/icon :main-icons2/muted
+       {:size     20
+        :no-color true}])
+    (when (and (not= lock-status :locked)
+               (not muted)
+               (pos? (int mentions-count)))
+      [rn/view {:style {:margin-right 2
+                        :margin-top   2}}
+       [quo2.counter/counter {:override-bg-color channel-color} mentions-count]])
+    (when (and (not= lock-status :locked)
```
erikseppanen: If you were interested in using a boolean `locked?` instead of a map `lock-status`, maybe you could test the values like this:

```clojure
(def locked? nil) ; false | true | nil
(println "[locked?]" (true? locked?)
          "[lock exists?]" (some? locked?)
          "[unlocked or lock doesn't exist]" (or (false? locked?) (nil? locked?)))
```

I can see the simplicity of just having 3 values in a `lock-status` map, though. On the other hand, I kinda like seeing `locked?` in the code instead of `lock-status`.

src/quo2/screens/list_items/channel.cljs
```diff
@@ -0,0 +1,68 @@
+(ns quo2.screens.list-items.channel
+  (:require [quo.react-native :as rn]
+            [reagent.core :as reagent]
+            [quo.previews.preview :as preview]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.list-items.channel :as quo2-channel]))
+
+(def descriptor [{:label   "Muted:"
+                  :key     :muted
+                  :type    :boolean}
```
erikseppanen, line 10: `muted?` instead of `muted` perhaps, since it's a boolean?

src/quo2/components/list_items/channel.cljs
J-Son89: @ibrkhalil, @Parveshdhull WDYT? If I make this change will also adjust in the avatar-channel component too as I was just matching the api here.

Parveshdhull: >  "[unlocked or lock doesn't exist]" (or (false? locked?) (nil? locked?)))

This can get little confusing, as having nil value of locked can mean both, unlocked or lock doesn't exist.

I guess if want to use boolean, we can create two variables
(as we already do that later in main component [ref](https://github.com/status-im/status-mobile/blob/147958dd135aeb5abbf1c792e70f8fcfcc2c4c91/src/quo2/components/avatars/channel_avatar.cljs#L10))
1. locked?
2. lock-icon?

ibrkhalil: This will require changes in `Channel-avatar` of course we can have two values and check for `nil` on lockekd?
I second Erik's suggestion

J-Son89: Updated to include these changes into both list item channel and channel avatar component

PR #14060, opened 2022-09-22
fix: switching to light / dark mode makes the app reset to the old ui when new ui toggle is activated

briansztamfater:
### Summary

Currently, if the user activates the New UI and then changes the app appearance (dark mode to light mode or vice versa), the app changes the appearance but it resets to the old UI. This PR fixes this.

#### Platforms

- Android
- iOS

##### Functional

- settings

### Steps to test

- Open Status
- Go to Settings > Advanced
- Activate New UI toggle
- Go to Settings > Appearance
- Toggle appearance to dark mode or light mode
- Verify that the app is still on the New UI and the theme properly changes

status: ready

### DIFF
--- a/src/status_im/ui/screens/appearance/views.cljs
+++ b/src/status_im/ui/screens/appearance/views.cljs
@@ -5,13 +5,14 @@
             [status-im.react-native.resources :as resources]
             [quo.core :as quo]
             [quo.design-system.colors :as colors]
-            [status-im.i18n.i18n :as i18n]))
+            [status-im.i18n.i18n :as i18n]
+            [status-im.utils.config :as config]))
 
 (defn button [label icon theme selected?]
   [react/touchable-highlight
    {:on-press (fn []
                 (re-frame/dispatch [:multiaccounts.ui/appearance-switched theme])
-                (re-frame/dispatch [:init-root :chat-stack])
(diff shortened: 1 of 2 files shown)


src/status_im/ui/screens/appearance/views.cljs
```diff
@@ -11,7 +11,7 @@
   [react/touchable-highlight
    {:on-press (fn []
                 (re-frame/dispatch [:multiaccounts.ui/appearance-switched theme])
-                (re-frame/dispatch [:init-root :chat-stack])
```
flexsurfer, line 14: i feel like we need to move `(if @config/new-ui-enabled?` here, and keep current code for old ui?

briansztamfater: Sure, I've tested and it works, but it is probably safer to keep the code for the old UI. Will make the change.

briansztamfater: Done

Parveshdhull (approved):

flexsurfer (approved):

pavloburykh:
@briansztamfater thanx for the PR. Tested, no regression found. Ready for merge.

Issue #14015, opened 2022-09-16
Implement Selectors components


flexsurfer:
Figma: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=398%3A18229

J-Son89:
This component has just received a design review and verifying some last pieces. Will close once those remarks are checked and adjusted if needed.

PR #14078, opened 2022-09-27
feat: implement selectors for quo2

J-Son89:
Implements https://github.com/status-im/status-mobile/issues/14015

Can be tested on ios and android in quo2 library.


J-Son89:
Can use these images to check design implementation 

Light mode state 1 0f 4
simulator_screenshot_F673D2E3-6895-463D-AE38-87A04DD00186

Light mode state 2 0f 4 - 
simulator_screenshot_952DE492-AA8E-45FC-93BA-16A9A6A42E8A

Light mode state 3 0f 4
simulator_screenshot_C5AAEE68-9C47-4D4A-B903-27B9B514ADA8

Light mode state 4 0f 4
simulator_screenshot_5CDBE6D5-0B53-4D7B-9A1F-FEC0EC96E0C9

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+     (colors/alpha colors/neutral-70 (if disabled? 0.3 1)))))
+
+(defn checkbox-prefill-view [props]
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
```
ibrkhalil: Vertical align maps please

J-Son89:
Can use these images to check design implementation 

Dark mode state 1 0f 4
simulator_screenshot_B257F100-4725-4FB1-A859-2E29B85F7365

Dark mode state 2 0f 4 - 
simulator_screenshot_FFAA44C8-EC70-4196-ADE4-46BC011E6EFE



Dark mode state 3 0f 4
simulator_screenshot_C0B1DF33-FBDF-4854-BF66-A62E208C1680

Dark mode state 4 0f 4
simulator_screenshot_3E665E9E-A053-43A9-944E-94D1FF15EC38

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
+(ns quo2.components.selectors.selectors
+  (:require [reagent.core :as reagent]
+            [quo.react :as react]
+            [quo.react-native :as rn]
+            [quo2.foundations.colors :as colors]
+            [cljs-bean.core :as bean]
+            [quo2.components.icon :as icons]))
+
+(defn get-color [value disabled?]
```
ibrkhalil: I think `checked?` might be a better name for the first param

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+                    (colors/alpha colors/white (if disabled? 0.3 1)))}]])]])))
+
+(defn checkbox-view [props]
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
```
ibrkhalil: Align maps

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+            :color (colors/alpha colors/white (if disabled? 0.3 1))}]])]])))
+
+(defn radio-view [props]
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
```
ibrkhalil: vertical alignment

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+                        :border-radius 40
+                        :border-width 1
+                        :border-color (get-color value disabled?)
+                        :background-color (colors/theme-colors colors/white
+                                                               (colors/alpha colors/neutral-80 0.4))})
+        :accessibility-label (str "radio-" (if value "on" "off"))
+        :accessibility-role  :checkbox}
+
+       [rn/view {:style {:margin-left :auto
+                         :height 14
```
ibrkhalil: vertical alignment

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+                         :margin-bottom :auto}}]]])))
+
+(defn toggle-view [props]
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
```
ibrkhalil: vertical alignment

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,125 @@
...
+                       {:height 20
+                        :width 30
+                        :border-radius 40
+                        :background-color (get-color value disabled?)})
+        :accessibility-label (str "toggle-" (if value "on" "off"))
+        :accessibility-role  :checkbox}
+       [rn/view {:style {:margin-left (if value 12 2)
+                         :height 16
```
ibrkhalil: vertical alignment

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+                         :background-color (colors/alpha colors/white
+                                                         (if (theme/dark?)
+                                                           (if disabled? 0.3 1)
+                                                           1))
+                         :border-radius 40
+                         :margin-right :auto
+                         :margin-top :auto
+                         :margin-bottom :auto}}]]])))
+
+(def radio (reagent/adapt-react-class (react/memo radio-view)))
```
ilmotta: Please, correct me if I'm wrong @J-Son89, but I think `React.memo` only works reliably if we're 100% sure React can skip rendering based on shallow comparisons of the props.

Let's take the `toggle-view` component as an example, it depends on `theme/dark?`, which depends on the ratom `quo.theme/theme`.

I honestly don't understand how the component renders correctly if the outside state changes (switch dark/light modes).

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+       [rn/view {:style {:margin-left :auto
+                         :height 14
+                         :width 14
+                         :background-color (when value (get-color value disabled?))
+                         :border-radius 40
+                         :margin-right :auto
+                         :margin-top :auto
+                         :margin-bottom :auto}}]]])))
+
+(defn toggle-view [props]
```
ilmotta: [Nitpick]

If you want to explicitly tell consumers of this namespace to only use the memoized versions, then you could make these `*-view` vars private.

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+            :color (colors/alpha colors/white (if disabled? 0.3 1))}]])]])))
+
+(defn radio-view [props]
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
```
ilmotta: @J-Son89, I don't have much experience with the `bean` library, but isn't its purpose to avoid calling functions such as `js->clj` to speedup data processing? At least that's what I got from their README. If we convert `containerStyle` to a Clojure map for merging like this function does, it seems to defeat `bean`'s purpose.

Edit: typo

ilmotta:
Quick tip @J-Son89, you can avoid these huge high resolution screenshots by putting the uploaded Github URL in a `img` tag instead of using markdown. This way you can put a width or height attribute on the image. Without it, the entire screen is occupied by the screenshot.

```
<img src="<URL>" height="300" />
```

See how it looks on my 4k display:

src/quo2/components/selectors/selectors.cljs
J-Son89: good call :)

src/quo2/components/selectors/selectors.cljs
J-Son89: yep, you are right. I tested this out just to be 100%. Should we remove the react/memo so? or look for ways to optimise this? WDYT ? 🤔

src/quo2/components/selectors/selectors.cljs
J-Son89: done 👍

src/quo2/components/selectors/selectors.cljs
J-Son89: tbh I've never seen this library before, I just copied some code from the `quo`  selectors. I can look into it but all suggestions welcome! :)

src/quo2/components/selectors/selectors.cljs
ilmotta: Edit: @J-Son89 explained again to me, and it seems the memo really doesn't work.

~~I'm just generally wary of things I don't understand. In this case, I really can't understand why `React.memo` works reliably with external state affecting how colors are calculated. Maybe if it works is because the entire tree is being evicted from the cache, so in reality the memo does nothing and React is always re-rendering. If you tested and it works it looks fine by me, but I'd avoid the memo unless we profiled these components and found performance issues. It's up to you I guess `¯\_(ツ)_/¯`, unless other people have something to comment on.~~

src/quo2/components/selectors/selectors.cljs
ilmotta: Yeah me neither, it's an interesting library and I can easily see places in our codebase where it makes sense to use it, but if we're doing what it doesn't recommend us to do (call `js->clj`), I think it's just adding overhead to the component without any real benefit.

src/quo2/components/selectors/selectors.cljs
ilmotta: So I guess the best we could do now is remove the cache wrapper. If we want to optimize the app, I think the best we can do is profile and identify the bottlenecks. If we find that such components are problematic we can get always get back to the drawing board. WDYT?

J-Son89: Yeah I think for now there's no need to memoize unless it becomes an issue. 
Realistically this potential ui bug from memoizing would only happen if there is other selectors on the page where we toggle the theme. So it's pretty much safe to have memoization as long as the screen is designed accordingly.

So I will remove for now and if needs be we can improve this later if necessary.

Parveshdhull:
> Implements #14015

"Implements" do sound better as this is an implementation, not an issue fix or anything. But "fixes", "closes, "resolves" etc. are github [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) and links issues to the PR.

image

No strong opinion or anything, just a small fyi.

J-Son89:
> > Implements #14015
> 
> "Implements" do sound better as this is an implementation, not an issue fix or anything. But "fixes", "closes, "resolves" etc. are github [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) and links issues to the PR.
> 
> image
> 
> No strong opinion or anything, just a small fyi.

Thanks Parvesh, I will use these keywords from now on. I'll also send Github a strongly worded email to add in "implements"😛

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+            [quo2.foundations.colors :as colors]
+            [cljs-bean.core :as bean]
+            [quo.theme :as theme]
+            [quo2.components.icon :as icons]))
+
+(defn get-color [value disabled?]
+  (if value
+    (colors/theme-colors
+     (colors/custom-color :primary 50 (when disabled? 30))
+     (colors/custom-color :primary 60 (when disabled? 30)))
```
Parveshdhull: [custom-color-by-theme](https://github.com/status-im/status-mobile/blob/b4d95b240d88066c34c0e8fcd966a214ac0b0912/src/quo2/foundations/colors.cljs#L248) can be used

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+                       (js->clj containerStyle)
+                       {:height (if value 20 21)
+                        :width (if value 20 21)
+                        :border-radius 6
+                        :border-width (if value 0 1)
+                        :background-color (if value
+                                            (get-color value disabled?)
+                                            (colors/theme-colors
+                                             colors/white
+                                             (colors/alpha colors/neutral-80 0.4)))
```
Parveshdhull: we already have [color](https://github.com/status-im/status-mobile/blob/b4d95b240d88066c34c0e8fcd966a214ac0b0912/src/quo2/foundations/colors.cljs#L62) for this configuration.

src/quo2/components/selectors/selectors.cljs
```diff
@@ -0,0 +1,129 @@
...
+  (let [{:keys [value onChange disabled? containerStyle]} (bean/bean props)]
+    (reagent/as-element
+     [rn/touchable-without-feedback
+      {:on-press (when (and onChange (not disabled?)) onChange)}
+      [rn/view
+       {:style        (merge
+                       (js->clj containerStyle)
+                       {:height 20
+                        :width 20
+                        :border-radius 40
```
Parveshdhull: why :border-radius is more than height, width

src/quo2/components/selectors/selectors.cljs
J-Son89: yep, will adjust

src/quo2/components/selectors/selectors.cljs
J-Son89: tbh I just matched values on the design and beyond a certain value it doesn't really matter. Or does it? 🤔

Parveshdhull: Yes, it's upto you, no strong opinion.
And you are right, It doesn't affect having a large border-radius.
I guess it looks nice :D (Its upto you)

Parveshdhull: > tbh I just matched values on the design

Yes, design does have 40 radius, weird :)

ilmotta (approved):

src/quo2/components/selectors/selectors.cljs
J-Son89: changed to 20 :)

src/quo2/components/selectors/selectors.cljs
J-Son89: removed use of bean library here as I think it's unnecessary. We can adjust if needs be.

src/quo2/components/selectors/selectors.cljs
J-Son89: adjusted all of these

PR #14084, opened 2022-09-28
feat: add disclaimer component to quo 2 library

### LINKED ISSUE #14082: Implement Disclaimer component in quo2 library
Figma: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=2976%3A43217

J-Son89:
fixes https://github.com/status-im/status-mobile/issues/14082

can use these photos to test with Figma:

### DIFF
--- a/src/quo2/components/selectors/disclaimer.cljs
+++ b/src/quo2/components/selectors/disclaimer.cljs
@@ -4,26 +4,28 @@
             [quo2.components.markdown.text :as text]
             [quo2.components.selectors.selectors :as selectors]))
 
-(defn disclaimer [{:keys [checked? on-change accessibility-label]} label]
-  [rn/view {:style
-            {:display :flex
-             :flex 2
-             :flex-direction :row
-             :background-color
-             (colors/theme-colors
-              colors/neutral-5
-              colors/neutral-80-opa-40)
-             :padding 11
-             :align-self :stretch
-             :border-radius 12
-             :border-width 1
-             :border-color (colors/theme-colors
-                            colors/neutral-20
-                            colors/neutral-70)}}
-   [selectors/checkbox {:checked? checked?
-                        :on-change on-change}]
-   [text/text {:accessibility-label accessibility-label
-               :size                :paragraph-2
-               :style {:margin-left 8}}
+(defn disclaimer [{:keys [on-change accessibility-label container-style]} label]
+  [rn/view
+   {:style
+    (merge
+     container-style
+     {:flex 1
+      :flex-direction :row
+      :background-color (colors/theme-colors
+                         colors/neutral-5
+                         colors/neutral-80-opa-40)
+      :padding 11
+      :align-self :stretch
+      :border-radius 12
+      :border-width 1
+      :border-color (colors/theme-colors
+                     colors/neutral-20
+                     colors/neutral-70)})}
+   [selectors/checkbox
+    {:on-change on-change}]
+   [text/text
+    {:accessibility-label accessibility-label
+     :size                :paragraph-2
+     :style {:margin-left 8}}
     label]])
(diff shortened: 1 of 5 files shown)


src/quo2/components/selectors/disclaimer.cljs
```diff
@@ -0,0 +1,29 @@
...
+             (colors/theme-colors
+              colors/neutral-5
+              colors/neutral-80-opa-40)
+             :padding 11
+             :align-self :stretch
+             :border-radius 12
+             :border-width 1
+             :border-color (colors/theme-colors
+                            colors/neutral-20
+                            colors/neutral-70)}}
```
Parveshdhull: hi @J-Son89, Thank you for PR
1. [display](https://reactnative.dev/docs/layout-props#display) property is redundant as this is the default value.
2. why `:flex 2`?
3. Please indent map values

src/quo2/screens/selectors/disclaimer.cljs
```diff
@@ -0,0 +1,26 @@
+(ns quo2.screens.selectors.disclaimer
+  (:require [quo.react-native :as rn]
+            [reagent.core :as reagent]
+            [quo2.components.selectors.disclaimer :as quo2]
+            [quo2.foundations.colors :as colors]))
+
+(defn cool-preview []
+  (let [checked? (reagent/atom false)]
```
Parveshdhull, line 9: this is an interesting way of managing the checkbox's state.

Checkbox depends on reagent atom <- which we pass in disclaimer component <- which will be passed to disclaimer component from its parent
(and maybe more ...) that's weird

Usually, native checkbox manages their state themselves and we only pass two fields 
1. `is-checked?` for setting default value and
2. `on-change` for any callback related to the new change.

So, I guess it will be better if we move this reagent/atom in the checkbox itself. wdyt?

Parveshdhull (approved):
apart from the above small suggestions, PR looks good to me

src/quo2/screens/selectors/disclaimer.cljs
ibrkhalil: I second Parvesh on this, We should set a default state of false inside the checkbox and toggle it when we want.

Maybe, Something like this
We're gonna create a wrapper that has the state .. And whenever we need the checkbox we call that wrapper which will either use the value passed to it or default to false.

```clj
(defn checkbox-main
  [value]
  (let [checked? (reagent/atom (or value false))]
    [checkbox-comp checked?]))
```

ibrkhalil (approved):
lgtm

src/quo2/screens/selectors/disclaimer.cljs
J-Son89: cool, will adjust! thanks for the suggestions! :D

src/quo2/components/selectors/disclaimer.cljs
J-Son89: yep, they're just pieces I should have cleaned up. Will adjust and indent :)

src/quo2/screens/selectors/disclaimer.cljs
J-Son89: hmm, are we sure we don't want the value externally? Having the value passed as a prop allows for easier access to the `value` of the checkbox.
in my experience accessing internal state of a component often leads to the value being "off" by one render.

I will bring this discussion to mobile channel

ilmotta (approved):

src/quo2/screens/selectors/disclaimer.cljs
J-Son89: agreed on internal state 👍

PR #14096, opened 2022-10-02
Allow Chinese characters as ENS names

du82:
This PR lowers the minimum stateofus.eth username requirement to one character to allow usernames like @1 and adds RegEx to support Chinese characters, to allow usernames like @中国

### Why?
Lots of users in China want to use Chinese characters in their names, and over a dozen people have suggested this feature to me.
There's a huge amount of unused address space on stateofus.eth, and this PR seeks to open that up.

### Why not also emojis?
From my testing, emojis have rendering issues, but Chinese hanzi and short usernames do not.

### DIFF
--- a/src/status_im/ens/core.cljs
+++ b/src/status_im/ens/core.cljs
@@ -179,7 +179,7 @@
 (defn- state [custom-domain? username usernames]
   (cond
     (or (string/blank? username)
-        (> 4 (count username))) :too-short
+        (> 1 (count username))) :too-short
     (valid-username? custom-domain? username)
     (if (usernames (fullname custom-domain? username))
       :already-added
--- a/src/status_im/ethereum/stateofus.cljs
+++ b/src/status_im/ethereum/stateofus.cljs
@@ -55,7 +55,7 @@
 (defn valid-username? [username]
   (boolean
    (and (lower-case? username)
-        (re-find #"^[a-z0-9]+$" username))))
+        (re-find #"^[a-z0-9\u4e00-\u9eff]+$" username))))
 
 (defn ens-name-parse [contact-identity]
   (when (string? contact-identity)


du82:
I've already registered the username **@1** in the app, and I'm sure others will register short names too when I share the PR with them.

du82:
I will need to adjust the strings to accommodate the changes, but it is 01:00 and I am very tired haha

siddarthkay:
@du64 : Thanks for the PR. I am not sure why the username limit was 4 before. I've assigned other CCs for a review since they might know why.

ibrkhalil:
I guess if we still need to disable shorter names to users other than the ones in China .. Can we make a region check on signup?

du82:
Just curious, did the Status team remove my ENS names? They worked just fine last night, and now they were removed from registration.

du82:
> I guess if we still need to disable shorter names to users other than the ones in China .. Can we make a region check on signup?

That would segment the userbase because some have features, others do not. I think we should just remove the limitations for everyone.

flexsurfer:
@3esmit could you please take a look, thanks

flexsurfer:
@du64 Status team can't remove anything from the contract

flexsurfer:
> Just curious, did the Status team remove my ENS names? They worked just fine last night, and now they were removed from registration.

cc @churik

churik:
@du64 thank you for PR!
1) What exactly do you mean by removing from registration, can you please elaborate on how you see it in the app?
2) What version of the app do you use?

du82:
> @du64 thank you for PR!
> 
> 1. What exactly do you mean by removing from registration, can you please elaborate on how you see it in the app?
> 2. What version of the app do you use?

1. Currently Status only accepts a-zA-Z0-9 in the RegEx. I mean you cant register anything outside of those boundaries, although many people want to.
2. I use a fork of Status aimed at being better for Chinese users, although it works the exact same on the official client.

du82:
concept replaced by unstoppable domains for a more open namespace that supports most unicode (minus lookalike characters) https://github.com/status-im/status-mobile/pull/14150

3esmit:
The reason why Status Usernames does not support anything else than a-zA-Z0-9 is the same reason why .com domains only accept these types of characters. 

Usernames are meant to be a safe and easy way for one person to pass their contact to other user, no matter what are their native language. 

If Usernames allowed unicode, than a couple of problems would rise:
- Difficult to universally describe names to other people, e.g. People that don't speak Chinese would have a hard time typing a chinese username.
- Possible to make phishing attacks, we see this problem with unicode DNS: https://www.xudongz.com/blog/2017/idn-phishing/

Issue #14065, opened 2022-09-23
Autocomplete mentions in the input box


cammellos:
> As a user
> 

> I want to have my mentions to be autocompleted
> 

> So that I don’t have to type the full name
> 

### Description

When inputting `@` in the input box, I should see a list mentionable users, filtered by the text following `@` 

- If I have set a nickname it should have priority over everything and should be shown here https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=2740%3A258262
- If the user has a ENS name but not a user set nickname it should be shown similarly to above
- If the user has only a display name it should be shown similarly to above
- In a one to one chat only 1 user will be in the mention box
- In a private group chat all the members will be in the mention box
- In a community all members will be in the mention box

### To be defined

- If the input box is full screen, where do we display the mention box

### Designs

https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=2706%3A263716

OmarBasem:
Done with the implementation of this issue but waiting for PR [#14098](https://github.com/status-im/status-mobile/pull/14098) to be merged first.

PR #14098, opened 2022-10-03
feat: reply to message new ui

### LINKED ISSUE #14066: Reply to a message - New UI
> As a user
> 

> I want to reply to a message using text
> 

> So that I can refer to a previous post
> 

### Description

We should allow the user to reply to messages.

Only the “top name” should be displayed (The username I have set for them > their ENS name > their display name)

### Designs

Initial state https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=2839%3A278327
Component: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=3173%3A55936

Some work is already done, check within the component library.

OmarBasem:
fixes #14066

### Summary

This PR connects the reply component new UI to the chat input when the user presses "reply" on a message.

- Android
- iOS

status: ready

### DIFF
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -465,12 +465,12 @@
 (defn calculate-y [context keyboard-shown min-y max-y added-value]
   (if keyboard-shown
     (if (= (:state @context) :max)
-      (+ max-y added-value)
+      max-y
       (if (< (:y @context) max-y)
         (+ (:y @context) added-value)
         (do
           (swap! context assoc :state :max)
-          (+ max-y added-value))))
+          max-y)))
     (do
       (swap! context assoc :state :min)
       min-y)))
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -8,6 +8,7 @@
             [quo.design-system.colors :as colors]
             [status-im.ui.screens.chat.components.style :as styles]
             [status-im.utils.fx :as fx]
+            [status-im.utils.handlers :refer [<sub]]
             [status-im.ui.screens.chat.components.reply :as reply]
             [status-im.multiaccounts.core :as multiaccounts]
             [status-im.chat.constants :as chat.constants]
@@ -563,7 +564,7 @@
        (fn []
          [:f>
           (fn []
-            (let [reply @(re-frame/subscribe [:chats/reply-message])
+            (let [reply (<sub [:chats/reply-message])
                   {window-height :height} (rn/use-window-dimensions)
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
 
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -568,8 +568,8 @@
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
 
                   max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
-                  max-height (- max-y 56 (:bottom insets))  ; 56 - topbar height
-                  added-value (if reply 38 0)
+                  max-height (- max-y 56 (:bottom insets))  ; 56 - top-bar height
+                  added-value (if reply 38 0) ; the increased height of the input box needed when replying
                   min-y (+ min-y added-value)
                   y (calculate-y context keyboard-shown min-y max-y added-value)
                   translate-y (reanimated/use-shared-value 0)
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -571,7 +571,7 @@
                   max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
                   max-height (- max-y 56 (:bottom insets))  ; 56 - topbar height
                   y (calculate-y context keyboard-shown min-y max-y)
-                  y (+ y (if reply 38 0))
+                  y (+ y (when reply 38))
                   translate-y (reanimated/use-shared-value 0)
                   shared-height (reanimated/use-shared-value min-y)
                   bg-opacity (reanimated/use-shared-value 0)
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -57,17 +57,17 @@
 ;; Instead of not changing panel we can show a placeholder with no permission
 (defn- request-record-audio-permission [set-active panel]
   (re-frame/dispatch
-   [:request-permissions
-    {:permissions [:record-audio]
-     :on-allowed
-     #(set-active panel)
-     :on-denied
-     #(utils.utils/set-timeout
-       (fn []
-         (utils.utils/show-popup
-          (i18n/label :t/audio-recorder-error)
-          (i18n/label :t/audio-recorder-permissions-error)))
-       50)}]))
+    [:request-permissions
+     {:permissions [:record-audio]
+      :on-allowed
+      #(set-active panel)
+      :on-denied
+      #(utils.utils/set-timeout
+         (fn []
+           (utils.utils/show-popup
+             (i18n/label :t/audio-recorder-error)
+             (i18n/label :t/audio-recorder-permissions-error)))
+         50)}]))
 
 (defn touchable-audio-icon [{:keys [panel active set-active accessibility-label input-focus]}]
   [pressable/pressable {:type                :scale
@@ -107,13 +107,13 @@
     ;; typing. Timeout might be canceled on `on-change`.
     (when platform/ios?
       (reset!
-       timeout-id
-       (utils.utils/set-timeout
-        #(re-frame/dispatch [::mentions/on-selection-change
-                             {:start start
-                              :end   end}
-                             mentionable-users])
-        50)))
+        timeout-id
+        (utils.utils/set-timeout
+          #(re-frame/dispatch [::mentions/on-selection-change
+                               {:start start
+                                :end   end}
+                               mentionable-users])
+          50)))
     ;; NOTE(rasom): on Android we dispatch event only in case if there
     ;; was no text changes during last 50ms. `on-selection-change` is
     ;; dispatched after `on-change`, that's why there is no another way
@@ -131,11 +131,11 @@
 (defonce chat-input-key (reagent/atom 1))
 
 (re-frame/reg-fx
- :chat.ui/clear-inputs
- (fn []
-   (reset! input-texts {})
-   (reset! mentions-enabled {})
-   (reset! chat-input-key 1)))
+  :chat.ui/clear-inputs
+  (fn []
+    (reset! input-texts {})
+    (reset! mentions-enabled {})
+    (reset! chat-input-key 1)))
 
 (defn force-text-input-update!
   "force-text-input-update! forces the
@@ -203,14 +203,14 @@
       (re-frame/dispatch [::mentions/calculate-suggestions mentionable-users]))))
 
 (re-frame/reg-fx
- :set-input-text
- (fn [[chat-id text]]
+  :set-input-text
+  (fn [[chat-id text]]
     ;; We enable mentions
-   (swap! mentions-enabled assoc chat-id true)
-   (on-text-change text chat-id)
+    (swap! mentions-enabled assoc chat-id true)
+    (on-text-change text chat-id)
     ;; We update the key so that we force a refresh of the text input, as those
     ;; are not ratoms
-   (force-text-input-update!)))
+    (force-text-input-update!)))
 
 (fx/defn set-input-text
   "Set input text for current-chat. Takes db and input text and cofx
@@ -222,8 +222,8 @@
         hydrated-mentions (map (fn [[t mention :as e]]
                                  (if (= t :mention)
                                    [:mention (str "@" (multiaccounts/displayed-name
-                                                       (or (get contacts mention)
-                                                           {:public-key mention})))]
+                                                        (or (get contacts mention)
+                                                            {:public-key mention})))]
                                    e)) text-with-mentions)
         info (mentions/->info hydrated-mentions)]
     {:set-input-text [chat-id text]
@@ -244,11 +244,11 @@
       (swap! mentions-enabled assoc chat-id true))
 
     (re-frame/dispatch
-     [::mentions/on-text-input
-      {:new-text      text
-       :previous-text previous-text
-       :start         start
-       :end           end}])
+      [::mentions/on-text-input
+       {:new-text      text
+        :previous-text previous-text
+        :start         start
+        :end           end}])
     ;; NOTE(rasom): on Android `on-text-input` is dispatched after
     ;; `on-change`, that's why mention suggestions are calculated
     ;; on `on-change`
@@ -286,9 +286,9 @@
       :on-text-input            (partial on-text-input mentionable-users chat-id)}
      (if mentions-enabled
        (for [[idx [type text]] (map-indexed
-                                (fn [idx item]
-                                  [idx item])
-                                @(re-frame/subscribe [:chat/input-with-mentions]))]
+                                 (fn [idx item]
+                                   [idx item])
+                                 @(re-frame/subscribe [:chat/input-with-mentions]))]
          ^{:key (str idx "_" type "_" text)}
          [rn/text (when (= type :mention) {:style {:color "#0DA4C9"}})
           text])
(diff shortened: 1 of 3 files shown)


du82:
looks good to me, tested with no issues - OnePlus Nord N20 5G

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -587,36 +588,38 @@
                                       (reanimated/set-shared-value bg-opacity (reanimated/with-timing 0)))
                                     (reanimated/set-shared-value translate-y (reanimated/with-timing (- y)))
                                     (reanimated/set-shared-value shared-height (reanimated/with-timing (min y max-height)))))
-              [reanimated/view {:style (reanimated/apply-animations-to-style
-                                        {:height shared-height}
-                                        {})}
+              [react/view
```
briansztamfater: Maybe use fragment (`[:<>` ) instead of view

OmarBasem: Thanks @briansztamfater . Fixed it.

du82:
The send button goes below the keyboard. I also think the background color of the part above the expanded text box doesnt match the other bottom sheets

https://user-images.githubusercontent.com/65834069/193645903-1055df3f-a143-458a-af28-02dd1399d2aa.mp4

flexsurfer:
it's not a simple one, message drawer is pretty complex and might be confusing, we don't have the dynamic part implemented yet, so this reply component should be inside the drawer, and its height should be taken into account in the drawer

OmarBasem:
@flexsurfer Yes I noticed this and I am currently working on a fix

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -587,36 +587,38 @@
...
                  ;handle
-                 [rn/view {:style (styles/new-bottom-sheet-handle)}]
-                 [rn/view {:style {:height (- max-y 80)}}
-                  [text-input {:chat-id                chat-id
-                               :on-content-size-change input-content-change
-                               :sending-image          false
-                               :refs                   refs
-                               :set-active-panel       #()}]]]]
+                  [rn/view {:style (styles/new-bottom-sheet-handle)}]
+                  [rn/view {:style {:height (- max-y 80)}}
```
flexsurfer: so it should be here, between the handle and input text, and we need to subtract from input height, and also we need to account for changes in translate-y , so we need to move drawer when we add new component in it

flexsurfer (changes requested):

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: Thanks @flexsurfer. Fixed it.

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -144,7 +144,7 @@
   (when-not (= reply @had-reply)
     (reset! had-reply reply)
     (when reply
-      (js/setTimeout #(input-focus text-input-ref) 250))))
+      (js/setTimeout #(input-focus text-input-ref) 0))))
```
J-Son89, line 148: is the time out of 0 seconds necessary? 🤔

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -554,30 +555,30 @@
...
+                 :text-input-ref text-input-ref}]
        (fn []
          [:f>
           (fn []
-            (let [text-input-ref (quo.react/create-ref)
-                  send-ref (quo.react/create-ref)
-                  refs {:send-ref       send-ref
-                        :text-input-ref text-input-ref}
-
+            (let [reply @(re-frame/subscribe [:chats/reply-message])
```
J-Son89: @flexsurfer should we be using `<sub` here instead? 🤔 
i.e from `[status-im.utils.handlers :refer [<sub ]]`

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -2,7 +2,7 @@
   (:require [status-im.ui.components.icons.icons :as icons]
             [quo.react-native :as rn]
             [oops.core :refer [oget]]
-            [quo.react :as react]
+            [quo.react :as quo.react]
```
J-Son89, line 5: nit: is this needed? I think it's usually imported as `react` elsewhere in the codebase.
imo, I prefer consistency with import names etc

src/status_im/ui/screens/chat/components/reply.cljs
OmarBasem: Thanks @J-Son89 for your reviews.

Yes `setTimeout` is necessary. `setTimeout(() => statement, 0)` is not the same as `statement`. Having a `setTimeout` of 0 ensures that the statement is enqueued and will get executed ASAP. Previously it was 250ms which I think is not necessary as we want to focus on the input right away.

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: I noticed both are used in the codebase. Maybe we need to change other occurrences of importing quo.react to one or the other.

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: yes, thank you

src/status_im/ui/screens/chat/components/input.cljs
J-Son89: yep, I took a look and it seems the large  majority is `react`, only a handful are quo.react

src/status_im/ui/screens/chat/components/reply.cljs
J-Son89: cool, assumed as much but just wanted to check. Do you think we should add a comment for this? 🤔

J-Son89 (approved):

du82:
lgtm!

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -144,7 +144,8 @@
   (when-not (= reply @had-reply)
     (reset! had-reply reply)
     (when reply
-      (js/setTimeout #(input-focus text-input-ref) 250))))
+      ;; A seTimeout of 0 is necessary to ensure the statement is enqueued and will get executed ASAP.
```
flexsurfer: setTimeout lost one letter

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -597,6 +599,7 @@
                                           (styles/new-input-bottom-sheet window-height))}
                  ;handle
                  [rn/view {:style (styles/new-bottom-sheet-handle)}]
+                 [reply/reply-message-auto-focus-wrapper (:text-input-ref refs)]
```
flexsurfer: as we have reply here we can pass it to reply-message-auto-focus-wrapper now, just to do not have sub twice

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -517,7 +518,8 @@
         (reanimated/set-shared-value shared-height (reanimated/with-timing min-y))
         (reanimated/set-shared-value bg-opacity (reanimated/with-timing 0)))
       (when (not= (:state @context) :max)
-        (let [new-y (+ min-y (- (max (oget evt "nativeEvent" "contentSize" "height") 22) 22))]
+        (let [new-y (+ min-y (- (max (oget evt "nativeEvent" "contentSize" "height") 22) 22))
+              new-y (+ new-y (if reply 38 0))]
```
flexsurfer: `(if reply 38 0)` is already used in two places, and in the future, we'll have more options, so probably better to move it to the function

OmarBasem: Thanks @flexsurfer. Fixed it.

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: @flexsurfer When I tried to do that the reply component did not update (did not appear until I exited the chat screen and entered again). Also when I dismiss the reply component it did not update.

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -597,6 +599,7 @@
                                           (styles/new-input-bottom-sheet window-height))}
                  ;handle
                  [rn/view {:style (styles/new-bottom-sheet-handle)}]
+                 [reply/reply-message-auto-focus-wrapper (:text-input-ref refs)]
                  [rn/view {:style {:height (- max-y 80)}}
```
flexsurfer: you need to reduce input height as well

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: oh i see 
```
(defn reply-message-auto-focus-wrapper [text-input-ref _]
  (let [had-reply (atom nil)]
    (fn [_ reply]
    ```
you have to take reply from render function

File: src/status_im/ui/screens/chat/components/input.cljs (hunk 9 above)
Comment (OmarBasem) [reply src/status_im/ui/screens/chat/components/input.cljs]: @flexsurfer I am not sure what do you mean?

Comment (flexsurfer) [reply src/status_im/ui/screens/chat/components/input.cljs]: `[rn/view {:style {:height (- max-y 80 (if reply 38 0))}}`

File: src/status_im/ui/screens/chat/components/input.cljs [hunk 10]
```diff
@@ -554,30 +556,30 @@
...
                   {window-height :height} (rn/use-window-dimensions)
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
 
-                  max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - defaul height
+                  max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
                   max-height (- max-y 56 (:bottom insets))  ; 56 - topbar height
-                  y (calculate-y context keyboard-shown min-y max-y)
-
+                  min-y (+ min-y (when reply 38))
+                  y (calculate-y context keyboard-shown min-y max-y reply)
```
flexsurfer: probably its better to move `added-value (if reply 38 0)` here, so we could use its value in `min-y` as well and in text input height

OmarBasem: @flexsurfer yep that's right

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -570,8 +569,9 @@
 
                   max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
                   max-height (- max-y 56 (:bottom insets))  ; 56 - topbar height
-                  min-y (+ min-y (when reply 38))
-                  y (calculate-y context keyboard-shown min-y max-y reply)
+                  added-value (if reply 38 0)
+                  min-y (+ min-y (when reply added-value))
```
flexsurfer: `when reply` is redundant here

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -462,19 +462,18 @@
...
-        (if (< (:y @context) max-y)
-          (+ (:y @context) added-value)
-          (do
-            (swap! context assoc :state :max)
-            (+ max-y added-value))))
-      (do
-        (swap! context assoc :state :min)
-        min-y))))
+(defn calculate-y [context keyboard-shown min-y max-y added-value]
+  (if keyboard-shown
```
flexsurfer, line 466: added-value shouldn't be used here, instead, it should be used in `get-input-content-change`

flexsurfer: `(let [new-y (+ min-y (- (max (oget evt "nativeEvent" "contentSize" "height") 22) 22))]` i guess here we should add `added-value`

OmarBasem: @flexsurfer without `added-value` in `calculate-y` the input height will update only on content change, i.e., after the user starts typing. 
Also, `added-value` is already being reflected in `get-input-content-change` as it is added to `min-y` before being passed to the function.

du82:
The text covers the cancel operation

Screenshot_2022-10-06-11-14-47-62_65ad86f86fd2321a85272a99323da10f

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -89,7 +89,7 @@
      [rn/view {:style (styles/reply-content)}
       [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
                                          :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}]
-      [rn/view {:style {:position :absolute :left 34 :right 54 :top 3 :flex-direction :row :align-items :center}}
+      [rn/view {:style {:position :absolute :left 34 :right 54 :top 3 :flex-direction :row :align-items :center :width 150}}
```
flexsurfer: how it will work on different devices with different width?

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -89,7 +89,7 @@
      [rn/view {:style (styles/reply-content)}
       [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
                                          :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}]
-      [rn/view {:style {:position :absolute :left 34 :right 54 :top 3 :flex-direction :row :align-items :center}}
+      [rn/view {:style {:position :absolute :left 34 :right 54 :top 3 :flex-direction :row :align-items :center :width "45%"}}
```
flexsurfer: sorry, I'm not sure how it will work, we have left and right and width, why left and right is not enough? percents not a good solution better always to use margins , and right should work i guess we just need to adjust it ?

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: yeah, right, so then we could keep it in the `calculate-y` but apply only to `y` not to max

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -89,7 +89,7 @@
      [rn/view {:style (styles/reply-content)}
       [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
                                          :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}]
-      [rn/view {:style {:position :absolute :left 34 :right 54 :top 3 :flex-direction :row :align-items :center :width "45%"}}
+      [rn/view {:style {:position :absolute :left 34 :top 3 :flex-direction :row :align-items :center :width "45%"}}
```
flexsurfer, line 92: its better to use fixed margins instead of percentages, percentage are not reliable, margins will be different on different devices if we use percentages

OmarBasem: @flexsurfer using right will not work here. Actually `right` has no effect here and can be removed. Text components normally overflow their parent. So, for `right` to have an effect `:overflow :hidden` must be set. But then, the text will just be cut without ellipsizing. 
For text components to ellipsize, their parent (or themselves) need to have a width (afaik). `width 45%` will allow each of the two text components we have (user's name, and message text) to have a width of 45%, so total is 90%. And each of them can ellipsize if bigger than 45%.

flexsurfer: cc @briansztamfater

OmarBasem:
I have completed the implementation of issue [#14065](https://github.com/status-im/status-mobile/issues/14065) but waiting for this PR to be merged first as they both work on the input box.
@briansztamfater @flexsurfer

flexsurfer (approved):

pavloburykh:
@OmarBasem thank you for the PR. 
Got one question: is it ok for now that reply to image message does not contain image itself but only the word "image"? 
photo_2022-10-11 18 38 14

OmarBasem:
@pavloburykh I am not sure what you mean? it already does not contain the image itself, only the word "Image", right?

pavloburykh:
> @pavloburykh I am not sure what you mean? it already does not contain the image itself, only the word "Image", right?

correct. My question is if it is expected behaviour or users should see the image itself in reply message? Like it is implemented in old UI (see an example below) or other messengers (i.e WhatApp, Telegram etc.) I mean, its a little bit confusing for me as a user not to be able to see images itself in the reply but only the word "image". Though it might be more a question to the design team. 

photo_2022-10-11 20 33 58

OmarBasem:
@pavloburykh according to the new design, the image itself is not displayed in the reply component.

pavloburykh:
> @pavloburykh according to the new design, the image itself is not displayed in the reply component.

ok, got it. Thank you @OmarBasem

PR #14101, opened 2022-10-03
ci: bump jenkins lib to use generalized nix library

jakubgs:
Necessary so other repos can use the Nix functions.

Realted to:
* https://github.com/status-im/nwaku/pull/1216

Depends on:
* https://github.com/status-im/status-jenkins-lib/pull/50


ibrkhalil (approved):
lgtm

PR #14072, opened 2022-09-25
Feature/#14025 messages loading skeleton

smohamedjavid:
Fixes #14025

### Summary

Implement Skeleton when messages are loading in the new UI. 

Simulator Screen Recording - iPhone 11 - 2022-09-29 at 14 58 38

Simulator Screen Recording - iPhone 11 - 2022-09-29 at 14 59 26

**Design** 

([Figma Link](https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=3071%3A278296))

#### Areas impacted

- 1-1 chat
- Public chats
- Group chats

**Platforms**

- Android
- iOS

**Steps to test**

1. Open the `Status` application
2. Navigate to the `Profile` tab
3. Navigate to `Advanced` and enable the `New UI`
4. Open any chats

status: ready

PR #14105, opened 2022-10-04
[#13967] Displays filtering tabs and read/unread filter

### LINKED ISSUE #13967: Distinguish between read and unread notifications
# Feature Issue

### User Story

>As a user
>I want to be able to distinguish between read and unread notifications
>So that I know which ones require my attention

### Description

This feature is part of a larger effort to create a new Activity Center. Please, reach out to members of the UX team before starting this work in case there are important details not apparent in Figma.

ilmotta:
fixes https://github.com/status-im/status-mobile/issues/13967

### Summary

This PR displays a series of hardcoded tabs to filter notifications by their type, while also supporting scrolling, fading out the right most edge, and scrolling (with animation) to the selected tab.

### Requirements

The big challenge in this PR was to create a tabs component under the following constraints:

- Horizontally scrollable if tabs don't fit in the view area.
- Selecting a tab activates and scrolls it into view (with animation).
- The end edge (right most side) of the tabs view area should fade out if tabs don't fit in, but not the left most side.
- The fading effect should work on top of blurred backgrounds and not just on hardcoded solid colors. This excludes the usage of a simpler linear gradient on top of tabs, and instead requires the usage of `@react-native-community/masked-view` (also used in PR https://github.com/status-im/status-mobile/pull/14072).
- As the user scrolls to the last tab, the length of the fading mask should gradually reduce until it disappears when the scroll view reached its end. This is the native behavior in Android when the option [fadingEdgeLength](https://reactnative.dev/docs/scrollview#fadingedgelength-android) is used in ReactNative's `ScrollView` or `FlatList` components.
- It should work on both iOS and Android, so using `ScrollView.fadingEdgeLength` is not an option.
- Performance should be acceptable.

### Review notes

- Filtering and fetching notifications based on their type is out of scope in this PR.
- I coded the new `scrollable-tabs` component in the same namespace as `tabs` in `quo2.components.tabs.tabs` because they should always render exactly the same.
- I haven't added the new `scrollable-tabs` to the preview area. It's not a true component on Figma at the moment.
- In the future, I think the new `scrollable-tabs` component should be merged with the simpler `tabs` component in a single interface (just as it will be done in Figma).

### Demo

[demo.webm](https://user-images.githubusercontent.com/46027/193704697-99d44476-024e-4451-9e8c-a6882509e4c4.webm)

### Comparison with native Android solution

The video below shows a quick comparison with the native Android fading edge option of `ScrollView`. Animations are slower due to my setup recording capabilities. The first tab bar is Android native, the second one should work on both iOS and Android.

[Comparison with native solution on Android.webm](https://user-images.githubusercontent.com/46027/193704012-7e048bd4-49fa-4ae9-b15c-3bd598658870.webm)

#### Platforms

- Android
- iOS

### Steps to test

1. Set the flag `status-im.utils.config/new-activity-center-enabled?` to true.
2. Enable the new UI.

status: ready


src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,138 @@
...
 (defn tabs [{:keys [default-active on-change]}]
   (let [active-tab-id (reagent/atom default-active)]
-    (fn [{:keys [data size] :or {size 32}}]
+    (fn [{:keys [data size] :or {size default-tab-size}}]
       (let [active-id @active-tab-id]
         [rn/view {:flex-direction :row}
          (for [{:keys [label id]} data]
            ^{:key id}
-           [rn/view {:margin-right (if (= size 32) 12 8)}
+           [rn/view {:margin-right (if (= size default-tab-size) 12 8)}
```
ibrkhalil: Can we add this styling to a `:style` prop ?

src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,138 @@
...
+                 on-change
+                 on-scroll
+                 scroll-event-throttle
+                 scroll-on-press?
+                 size]
+          :or   {fade-end-percentage   fade-end-percentage
+                 fade-end?             false
+                 scroll-event-throttle 64
+                 scroll-on-press?      false
+                 size                  default-tab-size}
```
ibrkhalil, line 80: I think @flexsurfer recommends keeping default props to the screens namespace .. But IDK since this component doesn't have a screens namespace.
And ideas Andre?

src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,138 @@
...
+                       :horizontal                        true
+                       :scroll-event-throttle             scroll-event-throttle
+                       :shows-horizontal-scroll-indicator false
+                       :data                              data
+                       :key-fn                            :id
+                       :on-scroll                         (fn [^js e]
+                                                            (when fade-end?
+                                                              (let [offset-x       (.. e -nativeEvent -contentOffset -x)
+                                                                    content-width  (.. e -nativeEvent -contentSize -width)
+                                                                    layout-width   (.. e -nativeEvent -layoutMeasurement -width)
```
ibrkhalil: Can we improve on this using `binaryage/oops` ?

ibrkhalil (approved):
Non blocking comments

src/quo2/components/tabs/tabs.cljs
ilmotta: Interesting, but then doesn't it cause a circular dependency? E.g. `screens` requires `tabs`, but `tabs` also requires `screens` to get the default map. I still find it weird to couple different namespaces for default function maps, even more because the screens namespace is not production code, it's like test code.

ibrkhalil: You're right Icaro, It's just code for manual visual testing
But I don't think `tabs` should require `screens.tabs`

J-Son89 (approved):

src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,139 @@
...
+          :or   {fade-end-percentage   fade-end-percentage
+                 fade-end?             false
+                 scroll-event-throttle 64
+                 scroll-on-press?      false
+                 size                  default-tab-size}
+          :as   props}]
+      (let [maybe-mask-wrapper (if fade-end?
+                                 [react/masked-view
+                                  {:mask-element (reagent/as-element
+                                                  [react/linear-gradient {:colors         ["black" "transparent"]
```
briansztamfater, line 85: What do you think of having a `quo2.foundation.gradients` ns similar to what we have for `quo2.foundation.colors` so we can reuse gradients?

ilmotta: @briansztamfater, thanks for reviewing this PR. I like the idea. I could do extract the gradient in a follow-up PR. I'm thinking now of a name for this gradient because it's special in the sense that it only exists to mask what's behind.

How about `horizontal-mask`? Ideas?

briansztamfater: Sure thing, we can do it on a follow-up PR. Maybe we should also figure out naming, I was thinking of a having some more generic naming describing colors and directions, for example: `black-to-transparent-left-to-right`. But we can open that discussion on other channel.

briansztamfater (approved):

src/quo2/components/tabs/tabs.cljs
ilmotta: Cool, I took note of your comment so we can discuss this in the near future.

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -107,24 +111,64 @@
...
+         [rn/view {:flex-direction   :row
+                   :padding-vertical 12}
+          [rn/view {:flex       1
+                    :align-self :stretch}
+           [tabs/scrollable-tabs {:size                32
+                                  :style               {:padding-left screen-padding}
+                                  :fade-end-percentage 0.79
+                                  :scroll-on-press?    true
+                                  :fade-end?           true
+                                  :on-change           (partial reset! selected-activity-type)
```
flexsurfer, line 158: mhm it doesn't look like good use of `partial` its just hard to read, `#(reset! selected-activity-type %)` much readable

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -107,24 +111,64 @@
...
-       [notifications-list]])}))
+      (let [screen-padding 20]
+        [:<>
+         [button/button {:icon     true
+                         :type     :grey
+                         :size     32
+                         :style    {:margin-vertical 12
+                                    :margin-left     screen-padding}
+                         :on-press #(>evt [:navigate-back])}
+          :close]
```
flexsurfer, line 143: namespace for icons should be used always

flexsurfer: so we could distinguish them in the code

src/quo2/components/tabs/tabs.cljs
flexsurfer, line 85: colors from `quo2.foundation.colors` should be used, or at least keywords not strings

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -107,24 +111,64 @@
...
                    :render-fn      render-notification}]))
 
+(defn filter-selector-read []
+  (let [unread-filter-enabled?  (<sub [:activity-center/status-filter-unread-enabled?])]
+    ;; TODO: Replace the button by a Filter Selector component once available for use.
+    [button/button {:icon     true
+                    :type     (if unread-filter-enabled? :primary :outline)
+                    :size     32
+                    :on-press #(if unread-filter-enabled?
+                                 (>evt [:activity-center.notifications/fetch-first-page {:status-filter :read}])
```
flexsurfer, line 126: no need to duplicate events , `#(>evt [:activity-center.notifications/fetch-first-page {:status-filter (if unread-filter-enabled? :read :unread}])` reads well enough

flexsurfer: or you could move status-filter to let

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -2,17 +2,21 @@
...
             [reagent.core :as reagent]
             [status-im.constants :as constants]
             [status-im.i18n.i18n :as i18n]
             [status-im.multiaccounts.core :as multiaccounts]
-            [status-im.ui.components.topbar :as topbar]
             [status-im.utils.datetime :as datetime]
             [status-im.utils.handlers :refer [<sub >evt]]))
 
+(defonce selected-activity-type
+  (reagent/atom :activity-type/all))
```
flexsurfer, line 18: seems like not used? probably better to do not have it for now ? and implement later

src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,139 @@
...
+                                                                          :locations      [(@fading :fade-end-percentage) 1]
```
flexsurfer, line 86: better to always use `get` or first item keyword, just for safety

src/quo2/components/tabs/tabs.cljs
```diff
@@ -1,21 +1,139 @@
...
+        (conj maybe-mask-wrapper
+              [rn/flat-list
+               (merge (dissoc props
+                              :default-active
+                              :fade-end-percentage
+                              :fade-end?
+                              :on-change
+                              :scroll-on-press?
+                              :size)
+                      {:ref                               (partial reset! flat-list-ref)
```
flexsurfer, line 102: same here for partial

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: Which namespace should we use? I saw different names being used, like `:main-icons2`. Should I use `:main-icons` instead (without the number 2)?

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: Thanks for the review @flexsurfer. Even though I already merged yesterday, I'll take your comments for my next PR. I like both ways, but I'll take your suggestion if you find it more readable ;)

src/status_im/ui/screens/activity_center/views.cljs
flexsurfer: `:main-icons2` if you use quo2 component

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: You're right, and I already removed it in another branch I'm working on.

src/quo2/components/tabs/tabs.cljs
ilmotta: This is a special case, because these "colors" are used for masking what's below. It doesn't matter which color I use, the gradient mask never produces a colored output. In other words, like in Photoshop, we're manipulating shades of black and white (transparency in RN MaskedView). My reasoning for not using colors from the foundation namespace is that using them gives the wrong impression that they actually matter, where in reality they have no effect.

I'll at least convert them to keywords.

src/quo2/components/tabs/tabs.cljs
ilmotta: I'm marking this thread as resolved because I implemented this suggestion in another branch I'm currently working on.

src/quo2/components/tabs/tabs.cljs
ilmotta: I'm marking this thread as resolved because I implemented this suggestion in another branch I'm currently working on.

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: I'm marking this thread as resolved because I implemented this suggestion in another branch I'm currently working on.

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: Good suggestion. I'm marking this thread as resolved because I implemented this in another branch I'm currently working on.

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: I'm marking this thread as resolved because I implemented this in another branch I'm currently working on.

d0adb04f993a140fb46435d878d07945b5a10e2b

PR #14125, opened 2022-10-07
[#13968] Fetch and reconcile notifications by type and read status

### LINKED ISSUE #13968: Filter notifications by their type
# Feature Issue

### User Story

>As a user
>I want to be able to filter notifications by type
>So that I can search for them more effectively

### Description

This feature is part of a larger effort to create a new Activity Center.

- Given that users can accumulate lots of notifications, they should be paginated, and a proper loading indicator should be displayed when necessary. Consider talking to the UX team about this.

ilmotta:
fixes #13968

### Summary

This PR adds support for fetching notifications whenever the user switches tabs, based on the current read/unread filter and the type of notification. Reconciliation of new notifications pushed from `status-im.transport.message.core/process-response` now sort the data according to the same rules used by status-go.

### Demo

Please, ignore in the demo the incorrect toggle button icon and the non-green Accept button. These problems are being tracked already.

[PR.webm](https://user-images.githubusercontent.com/46027/194641364-125c2477-32a7-4450-bafe-ce618ba6c881.webm)

### Review notes

- Switching tabs never filter on the client-side, there's always a roundtrip to the backend.
- The app db structure changed to accommodate the fact that status-go uses a pagination cursor for every group of read/unread + type of notification.

**Extra goodies**:

- [x] Extract tabs component from the main screen component to reduce rendering cost when switching tabs.
- [x] Display empty state component when there are no notifications.
- [x] Fix screen flickering due to quickly flushing and filling the db state (a known problem in the production notification center).
- [x] Display top tab bar as a sticky header.
- [x] Namespace icon keywords.
- [x] Remove warning about children without unique key.

### For the curious

- This PR uses the [re-frame.core/path](https://day8.github.io/re-frame/api-re-frame.core/#path) standard interceptor. It's a useful way to narrow the scope of the app db, so that the event handler knows less about the "world".

- Cool fact about the test changes in this PR: although most of them changed, **their description did not**, because they communicate the overall behavior of the event handler, and that was kept mostly unchanged from previous PRs.

- When reading the tests, it's my hope that you can clearly see the 3 universal test sections `Arrange` `Act`, and `Assert`, and that they're understandable. In general, tests should be fairly easy (not just simple) to change, because they're often aggressively changed (more than the production code). If you have feedback, please do share so we can improve o/

#### Platforms

- Android
- iOS

### Steps to test

This PR relies on fixes that were pushed to a draft PR on status-go (https://github.com/status-im/status-go/pull/2831). Yes, it's getting old to mention this on every PR. It's time to get that PR reviewed! Hopefully soon, so everyone will be able to test more easily the new AC.

Therefore, to test, use the draft PR branch locally and point status-mobile to it and set `status-im.utils.config/new-activity-center-enabled?` to `true`.

status: ready

### DIFF
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -1,5 +1,6 @@
 (ns status-im.activity-center.core
   (:require [re-frame.core :as re-frame]
+            [status-im.constants :as constants]
             [status-im.data-store.activities :as data-store.activities]
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.fx :as fx]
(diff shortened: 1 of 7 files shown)


ibrkhalil (approved):
lgtm
But I wish the notification types were keywords instead of integers.

src/status_im/subs/activity_center.cljs
```diff
@@ -6,35 +6,33 @@
             [clojure.string :as string]))
 
 (re-frame/reg-sub
- :activity-center/notifications-read
+ :activity-center/notifications
```
flexsurfer, line 9: there should be root `:activity-center` subscription, and all 3 should be based on it

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -108,67 +105,108 @@
...
+                    :justify-content  :center
+                    :padding-vertical 12}}
+   [rn/view {:style {:background-color colors/neutral-80
+                     :height           120
+                     :margin-bottom    20
+                     :width            120}}]
+   [text/text {:size   :paragraph-1
+               :style  {:padding-bottom 2}
+               :weight :semi-bold}
+    "No notifications"]
```
flexsurfer, line 137: i18n

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -108,67 +105,108 @@
...
+   [text/text {:size :paragraph-2}
+    "Your notifications will be here"]])
```
flexsurfer, line 139: i18n

src/status_im/activity_center/core.cljs
```diff
@@ -8,90 +9,118 @@
...
-  {:events [:activity-center.notifications/fetch-success]}
-  [{:keys [db]} notifications-group {:keys [cursor notifications]}]
-  {:db (-> db
-           (update-in [:activity-center notifications-group] dissoc :loading?)
-           (assoc-in [:activity-center notifications-group :cursor] cursor)
-           (update-in [:activity-center notifications-group :data]
-                      concat
-                      (map data-store.activities/<-rpc notifications)))})
+  {:events       [:activity-center.notifications/fetch-success]
+   :interceptors [(re-frame/path [:activity-center :notifications])]}
```
flexsurfer: better to do not use path interceptor, it's only confusing, and might lead to mistakes, we should always work with db directly in the event ,

flexsurfer (changes requested):

flexsurfer:
```
This PR uses the [re-frame.core/path](https://day8.github.io/re-frame/api-re-frame.core/#path) standard interceptor. It's a useful way to narrow the scope of the app db, so that the event handler knows less about the "world".
```
there is nothing useful in that, i mean, the event handler still knows less even if it takes its key from db directly, but it a big confusion for devs who will be working with this event later, they could miss interceptor and think its a db

ilmotta:
> But I wish the notification types were keywords instead of integers.

I totally agree @ibrkhalil. In the past, I refrained from changing how the types of notifications are handled because I didn't want to affect the current Notification Center. I think the data store layer could be the place for such transformations, but yeah, I'll just leave this as is for the moment.

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: FYI: There's a comment on top of this function explaining why there are no translations. This is still in flux, and designers told me this will probably change.

```clojure
;; TODO(2022-10-07): The empty state is still under design analysis, so we
;; shouldn't even care about translations at this point. A placeholder box is
;; used instead of an image.
```

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: FYI: There's a comment on top of this function explaining why there are no translations. This is still in flux, and designers told me this will probably change.

```clojure
;; TODO(2022-10-07): The empty state is still under design analysis, so we
;; shouldn't even care about translations at this point. A placeholder box is
;; used instead of an image.
```

flexsurfer: oh sorry, missed

ilmotta:
> ```
> This PR uses the [re-frame.core/path](https://day8.github.io/re-frame/api-re-frame.core/#path) standard interceptor. It's a useful way to narrow the scope of the app db, so that the event handler knows less about the "world".
> ```
> 
> there is nothing useful in that, i mean, the event handler still knows less even if it takes its key from db directly, but it a big confusion for devs who will be working with this event later, they could miss interceptor and think its a db

I understand your reasoning, @flexsurfer, but I totally disagree with the argument. I think developers using re-frame are expected to know the usage of interceptors and see if event handlers have interceptors applied, always. The core path interceptor is handy for exactly the event handler shown in this PR, but I see it being useful in other places too.

Having said that, **I'll remove the usage of the `path` interceptor exclusively because you feel so strong about not using it**.

src/status_im/subs/activity_center.cljs
ilmotta: Sorry @flexsurfer, I didn't understand this one comment. This subscription looks like the following, what's your suggestion?

```clojure
(re-frame/reg-sub
 :activity-center/notifications
 (fn [db]
   (get-in db [:activity-center :notifications])))
```

src/status_im/activity_center/core.cljs
ilmotta: Please, see my comment below https://github.com/status-im/status-mobile/pull/14125#issuecomment-1273173558.

flexsurfer:
> I think developers using re-frame are expected to know the usage of interceptors and see if event handlers have interceptors applied, always. The core path interceptor is handy for exactly the event handler shown in this PR, but I see it being useful in other places too.

i mean, if we have 500 events without paths and a few with, imagine working with events in 99% you need to keep in mind something you don't really need. honestly its much more readable when you update path directly in db instead of using interceptor, it is always better and more efficient to keep things simple, we used to use reg-event-db and reg-event-fx it was a bad experience

siddarthkay (approved):

src/status_im/subs/activity_center.cljs
cammellos: I think @flexsurfer means that we should have a subscription
```
(reg-root-key-sub :activity-center :activity-center)
 ```

and then compose the 3 subscriptions:
```
(re-frame/reg-sub
:activity-center/subscriptions
 :<- [:activity-center]
 (fn [activity-center]
   (:notifications activity-center)))
```

It's tempting to do:

```
(re-frame/reg-sub
:activity-center/subscriptions
 :<- [:activity-center]
  :notifications)
```
But it's prone to error as if a second parameter is passed and `activity-center` is nil it would return that one.

Comment (flexsurfer) [reply src/status_im/subs/activity_center.cljs:9]:
```clojure
(reg-root-key-sub :activity-center :activity-center)


(re-frame/reg-sub
 :activity-center/notifications
 :<- [:activity-center]
 (fn [activity-center]
   (:notifications activity-center)))
```

flexsurfer: db should never be used in `re-frame/reg-sub`

cammellos:
> > But I wish the notification types were keywords instead of integers.
> 
> I totally agree @ibrkhalil. In the past, I refrained from changing how the types of notifications are handled because I didn't want to affect the current Notification Center. I think the data store layer could be the place for such transformations, but yeah, I'll just leave this as is for the moment.

We can convert enums to keywords in data store (status-go prefers enums & consts, rather than string consts), from previous experience in the project though that has resulted in a few bugs (typos for example) and code that's harder to refactor.

Granted, that it's probably also due to lack of tests, myself I'd lean towards const since they provide a bit more safety, the code is just as readable and they are independent from the underneath representation of the data, but I don't have a strong opinion on the matter.

src/status_im/subs/activity_center.cljs
flexsurfer: mhm for some reason i didn't have all your comments :(

ilmotta:
> > > But I wish the notification types were keywords instead of integers.
> > 
> > 
> > I totally agree @ibrkhalil. In the past, I refrained from changing how the types of notifications are handled because I didn't want to affect the current Notification Center. I think the data store layer could be the place for such transformations, but yeah, I'll just leave this as is for the moment.
> 
> We can convert enums to keywords in data store (status-go prefers enums & consts, rather than string consts), from previous experience in the project though that has resulted in a few bugs (typos for example) and code that's harder to refactor.
> 
> Granted, that it's probably also due to lack of tests, myself I'd lean towards const since they provide a bit more safety, the code is just as readable and they are independent from the underneath representation of the data, but I don't have a strong opinion on the matter.

Thanks for sharing your past experience. Yeah, it's nothing major for sure. And to be fair, it's important to be careful with data model adaptations because they might cause more headache than anything else. Some people would call our data store namespaces _anti-corruption layers_, with the goal of transforming one data-model to another.

> the code is just as readable and they are independent from the underneath representation of the data

It really is, the production code. It's the tests that I felt were harder to comprehend due to the use of numbers as indexes, but again, nothing major.

cammellos:
> > > > But I wish the notification types were keywords instead of integers.
> > > 
> > > 
> > > I totally agree @ibrkhalil. In the past, I refrained from changing how the types of notifications are handled because I didn't want to affect the current Notification Center. I think the data store layer could be the place for such transformations, but yeah, I'll just leave this as is for the moment.
> > 
> > 
> > We can convert enums to keywords in data store (status-go prefers enums & consts, rather than string consts), from previous experience in the project though that has resulted in a few bugs (typos for example) and code that's harder to refactor.
> > Granted, that it's probably also due to lack of tests, myself I'd lean towards const since they provide a bit more safety, the code is just as readable and they are independent from the underneath representation of the data, but I don't have a strong opinion on the matter.
> 
> Thanks for sharing your past experience. Yeah, it's nothing major for sure. And to be fair, it's important to be careful with data model adaptations because they might cause more headache than anything else. Some people would call our data store namespaces _anti-corruption layers_, with the goal of transforming one data-model to another.
> 
> > the code is just as readable and they are independent from the underneath representation of the data
> 
> It really is, the production code. It's the tests that I felt were harder to comprehend due to the use of numbers as indexes, but again, nothing major.

Just out of curiosity, why not use the consts in the tests as well?

src/status_im/subs/activity_center.cljs
ilmotta: Okay, thanks. Now I'm curious to see why creating the `:activity-center` root subscription is better. I'm assuming it's for performance reasons, because layer-3 subs use cached data from layer-2.

ilmotta:
> Just out of curiosity, why not use the consts in the tests as well?

It's usually standard practice in tests not to rely in production code in the assertions, otherwise we can put _banana_ in the constant and the tests will still pass. I honestly think for the tests in this PR it's okay to use the constants from prod code, the tests just lose a bit of their value, in return for less maintenance.

Edit: I'm giving this a go and changing the tests, I think it's a fair tradeoff.

src/status_im/subs/activity_center.cljs
ilmotta: Done.

```
Refactor: use more layer-2 subs
```

https://github.com/status-im/status-mobile/pull/14125/commits/81e2f4bec31466191d3ee80c23fc64efbc21cf6a

src/status_im/subs/activity_center.cljs
```diff
@@ -6,35 +6,36 @@
             [clojure.string :as string]))
 
 (re-frame/reg-sub
- :activity-center/notifications-read
- (fn [db]
-   (get-in db [:activity-center :notifications-read :data])))
+ :activity-center/notifications
```
flexsurfer, line 9: just wondering how its different from `:activity.center/notifications` why do we have both

flexsurfer (approved):

src/status_im/subs/activity_center.cljs
ilmotta: I'm not sure I understand the question correctly, but the idea is that all new activities are managed under a different root state called `:activity-center` (with a dash, not with a dot). I know it's hard to spot this, but eventually the current (old) notification center will be removed and with it, the subscription `:activity.center/notifications` (with dot).

This was a strategy to code in the new AC without ever messing around with the existing app state for the old Notification Center.

Does that answer your question?

flexsurfer: yes, thank you

Issue #14071, opened 2022-09-23
UI Component - Floating Shell Button


Parveshdhull:
This issue is to track the creation of the component: Floating Shell Button

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=8835%3A151980

PR #14127, opened 2022-10-09
Implementation of Dynamic Button Component

### LINKED ISSUE #14100: UI Component - Dynamic Button
This issue is to track the creation of the component: Dynamic Button

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=8834%3A152423

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/14100

status: ready


src/quo2/components/buttons/dynamic_button.cljs
```diff
@@ -0,0 +1,97 @@
...
+                                       :mention      2
+                                       :notification 8
+                                       :search       4
+                                       :search2      6
+                                       :bottom       6)}}])
+
+(defn dynamic-button
+  "[:dynamic-button opts]
+   opts
+   {:type        :jump-to/:mention/:notification/:search/:search2/:bottom
```
J-Son89: what's the difference between search and search2?

Parveshdhull: https://discord.com/channels/624205794384281629/626744384565674034/1028970164437393480

Parveshdhull: > Search button has the label "Back" in the dynamic button but in the floating shell button, it is just only a search icon.

J-Son89: okay, consider changing the name to something more descriptive so. e.g :search-with-label

src/quo2/components/buttons/dynamic_button.cljs
```diff
@@ -0,0 +1,103 @@
+(ns quo2.components.buttons.dynamic-button
+  (:require [quo.react-native :as rn]
+            [reagent.core :as reagent]
+            [status-im.i18n.i18n :as i18n]
+            [quo2.components.icon :as icon]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.markdown.text :as text]))
+
+(defn- in?
+  "true if coll contains elm"
```
J-Son89: in my opinion, it's better to be explicit with variable naming. I don't think there's much value add to character saving.
consider -   
```
(defn- in?
  "true if collection contains element"
  [collection element]
  (some #(= element %) collection))
```
which I assume it to be.

although perhaps this is a clojurescript practice I'm unaware of and there are some advantages to this character saving? 🤔

src/quo2/components/buttons/dynamic_button.cljs
```diff
@@ -0,0 +1,103 @@
...
+                                       :search            6
+                                       :bottom            6)}}])
+
+(defn dynamic-button
+  "[:dynamic-button opts]
+   opts
+   {:type                :jump-to/:mention/:notification-down/:notification-up/:search/:search-with-label/:bottom
+    :on-press            fn
+    :count               mentions or notifications count
+    :customization-color customize jump-to and mention button color}"
```
J-Son89, line 65: not only related to this p.r but do we have a common name for this prop??
It's common across a lot of components so maybe we should aim to standardize this name to some extent 🤔

J-Son89 (approved):
👌

src/quo2/components/buttons/dynamic_button.cljs
Parveshdhull: Hi @J-Son89, Thank you for reviewing the PR.
Yes, in some places this variable is just named `color`.
I don't have a strong opinion, but `customization-color` sounds better to me, as this is a customization color.

J-Son89: yep I agree and let's keep an eye on future pr's to try to use this name where possible 👍

src/quo2/components/buttons/dynamic_button.cljs
```diff
@@ -0,0 +1,103 @@
...
+            [quo2.foundations.colors :as colors]
+            [quo2.components.markdown.text :as text]))
+
+(defn- in?
+  "true if collection contains element"
+  [collection element]
+  (some #(= element %) collection))
+
+(defn- get-button-color [type pressed? customization-color]
+  (if (in? '(:jump-to :mention) type)
```
flexsurfer, line 15: could be just `(#{:jump-to :mention} type)`

Parveshdhull: yes it looks better, will replace in https://github.com/status-im/status-mobile/issues/14071. Thank you

PR #14141, opened 2022-10-11
clean deps and fix warnings

flexsurfer:
clean deps and fix warnings


yqrashawn (approved):

src/status_im/wallet_connect/core.cljs
```diff
@@ -105,15 +105,6 @@
        (.then #(log/debug "[wallet-connect] session topic " topic " changed to account " (first accounts)))
        (.catch #(log/error "[wallet-connect] " %)))))
 
-(re-frame/reg-fx
```
siddarthkay, line 108: I maybe wrong here, but `:wc-2-change-session` seems to be used here : https://github.com/status-im/status-mobile/blob/34bd436b16bbf6f8f41e29f0f78389a75856f49e/src/status_im/wallet_connect/core.cljs#L161

flexsurfer: it was duplicated

siddarthkay: Oh i see, Thanks for clarification :)

siddarthkay (approved):

Parveshdhull (approved):

PR #14142, opened 2022-10-11
feat: autocomplete mentions new ui

### LINKED ISSUE #14065: Autocomplete mentions in the input box

OmarBasem:
fixes #14065

### Summary

This PR implements the [new UI](https://www.figma.com/file/wA8Epdki2OWa8Vr067PCNQ/Composer-for-Mobile?node-id=2102%3A171080) for autocomplete mentions in chat. 

Since group chats new UI is not yet implemented, I am attaching a video showing how the mentions list looks when there are multiple suggestions.

#### Platforms

- Android
- iOS

https://user-images.githubusercontent.com/29354102/195049214-5df82b75-b904-4f9b-8eae-026c331ba199.mp4

status: ready

### DIFF
--- a/src/quo/design_system/colors.cljs
+++ b/src/quo/design_system/colors.cljs
@@ -179,8 +179,8 @@
                      purple
                      "#8B3131"])
 
-(def mention-incoming "#0DA4C9")
-(def mention-outgoing "#9EE8FA")
+(def mention-incoming "#4360DF")
+(def mention-outgoing "#4360DF")
 (def text black)
 (def text-gray gray)
 (def default-community-color "#773377")
--- a/src/quo/design_system/colors.cljs
+++ b/src/quo/design_system/colors.cljs
@@ -179,8 +179,8 @@
                      purple
                      "#8B3131"])
 
-(def mention-incoming "#4360DF")
-(def mention-outgoing "#4360DF")
+(def mention-incoming "#0DA4C9")
+(def mention-outgoing "#9EE8FA")
 (def text black)
 (def text-gray gray)
 (def default-community-color "#773377")
--- a/src/status_im/ui/screens/chat/components/input.cljs
+++ b/src/status_im/ui/screens/chat/components/input.cljs
@@ -331,47 +331,46 @@
                                   [idx item])
                                 @(re-frame/subscribe [:chat/input-with-mentions]))]
          ^{:key (str idx "_" type "_" text)}
-         [rn/text (when (= type :mention) {:style {:color "#0DA4C9"}})
-          text])
+         [rn/text (when (= type :mention) {:style {:color (:text-04 @colors/theme)}}) text])
        (get @input-texts chat-id))]))
 
 (defn mention-item
   [[public-key {:keys [alias name nickname] :as user}] _ _ text-input-ref]
   (let [ens-name? (not= alias name)]
-    [list-item/list-item
-     (cond-> {:icon              [photos/member-photo public-key]
-              :size              :small
-              :text-size         :small
-              :title
-              [text/text
-               {:weight          :medium
-                :ellipsize-mode  :tail
-                :number-of-lines 1
-                :size            :small}
-               (if nickname
-                 nickname
-                 name)
-               (when nickname
-                 [text/text
-                  {:weight         :regular
-                   :color          :secondary
-                   :ellipsize-mode :tail
-                   :size           :small}
-                  " "
-                  (when ens-name?
-                    "@")
-                  name])]
-              :title-text-weight :medium
-              :on-press
-              (fn []
-                (re-frame/dispatch [:chat.ui/select-mention text-input-ref user]))}
+    [rn/touchable-opacity
+     {:on-press #(re-frame/dispatch [:chat.ui/select-mention text-input-ref user])}
 
-       ens-name?
-       (assoc :subtitle alias))]))
+     [list-item/list-item
+      (cond-> {:icon              [photos/member-photo public-key]
+               :size              :small
+               :text-size         :small
+               :title
+               [text/text
+                {:weight          :medium
+                 :ellipsize-mode  :tail
+                 :number-of-lines 1
+                 :size            :small}
+                (if nickname
+                  nickname
+                  name)
+                (when nickname
+                  [text/text
+                   {:weight         :regular
+                    :color          :secondary
+                    :ellipsize-mode :tail
+                    :size           :small}
+                   " "
+                   (when ens-name?
+                     "@")
+                   name])]
+               :title-text-weight :medium}
+
+        ens-name?
+        (assoc :subtitle alias))]]))
 
 (def chat-toolbar-height (reagent/atom nil))
 
-(defn autocomplete-mentions [text-input-ref bottom]
+(defn autocomplete-mentions-old [text-input-ref bottom]
   (let [suggestions @(re-frame/subscribe [:chat/mention-suggestions])]
     (when (seq suggestions)
       (let [height (+ 16 (* 52 (min 4.5 (count suggestions))))]
@@ -388,6 +387,22 @@
             :key-fn                    first
             :render-data               text-input-ref
             :render-fn                 mention-item}]]]))))
+
+(defn autocomplete-mentions []
+  [:f>
+   (fn []
+     (let [suggestions @(re-frame/subscribe [:chat/mention-suggestions])
+           animation (reanimated/use-shared-value 0)]
+       (quo.react/effect! #(do
+                             (reanimated/set-shared-value animation (reanimated/with-timing (if (seq suggestions) 0 200)))))
+       [reanimated/view {:style (reanimated/apply-animations-to-style
+                                 {:transform [{:translateY animation}]} {:bottom 0 :position :absolute :z-index 5 :max-height 180})}
+        [list/flat-list
+         {:keyboardShouldPersistTaps :always
+          :data                      suggestions
+          :key-fn                    #(str %2)
+          :render-fn                 mention-item
+          :content-container-style    {:padding-bottom 12}}]]))])
 
 (defn on-chat-toolbar-layout [^js ev]
   (reset! chat-toolbar-height (-> ev .-nativeEvent .-layout .-height)))
@@ -474,6 +489,17 @@
     (do
       (swap! context assoc :state :min)
       min-y)))
+
+(defn calculate-y-with-mentions [y max-y max-height chat-id suggestions reply]
+  (let [input-text (:input-text (get (<sub [:chat/inputs]) chat-id))
+        num-lines (count (string/split input-text "\n"))
+        text-height (* num-lines 22)
+        mentions-height (min 132 (+ 16 (* 46 (- (count suggestions) 1))))
+        should-translate (if (< (- max-height text-height) mentions-height) true false)
+        min-value (if-not reply 132 170)
+        ; translate value when mentions list appear while at bottom of expanded input sheet
+        mentions-translate-value (if should-translate (min min-value (- mentions-height (- max-height text-height))) mentions-height)]
+    (when (or (< y max-y) should-translate) mentions-translate-value)))
 
 (defn get-bottom-sheet-gesture [context translate-y text-input-ref keyboard-shown min-y max-y shared-height max-height bg-opacity]
   (-> (gesture/gesture-pan)
@@ -564,18 +590,18 @@
          [:f>
           (fn []
             (let [reply (<sub [:chats/reply-message])
+                  suggestions (<sub [:chat/mention-suggestions])
                   {window-height :height} (rn/use-window-dimensions)
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
-
                   max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
                   max-height (- max-y 56 (:bottom insets))  ; 56 - top-bar height
-                  added-value (if reply 38 0) ; increased height of input box needed when reply
-                  min-y (+ min-y added-value)
+                  added-value (if (and (not (seq suggestions)) reply) 38 0) ; increased height of input box needed when reply
+                  min-y (+ min-y (when reply 38))
                   y (calculate-y context keyboard-shown min-y max-y added-value)
+                  y (+ y (when (seq suggestions) (calculate-y-with-mentions y max-y max-height chat-id suggestions reply)))
                   translate-y (reanimated/use-shared-value 0)
                   shared-height (reanimated/use-shared-value min-y)
                   bg-opacity (reanimated/use-shared-value 0)
-
                   input-content-change (get-input-content-change context translate-y shared-height max-height
                                                                  bg-opacity keyboard-shown min-y max-y)
                   bottom-sheet-gesture (get-bottom-sheet-gesture context translate-y (:text-input-ref refs) keyboard-shown
@@ -607,18 +633,19 @@
                                :refs                   refs
                                :set-active-panel       #()}]]]]
                ;CONTROLS
-               [rn/view {:style (styles/new-bottom-sheet-controls insets)}
-                [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/image]
-                [rn/view {:width 12}]
-                [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/reaction]
-                [rn/view {:flex 1}]
-                ;;SEND button
-                [rn/view {:ref send-ref :style (when-not (seq (get @input-texts chat-id)) {:width 0 :right -100})}
-                 [quo2.button/button {:icon     true :size 32 :accessibility-label :send-message-button
-                                      :on-press #(do (swap! context assoc :clear true)
-                                                     (clear-input chat-id refs)
-                                                     (re-frame/dispatch [:chat.ui/send-current-message]))}
-                  :main-icons2/arrow-up]]]
+               (when-not (seq suggestions)
+                 [rn/view {:style (styles/new-bottom-sheet-controls insets)}
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/image]
+                  [rn/view {:width 12}]
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/reaction]
+                  [rn/view {:flex 1}]
+                  ;;SEND button
+                  [rn/view {:ref send-ref :style (when-not (seq (get @input-texts chat-id)) {:width 0 :right -100})}
+                   [quo2.button/button {:icon     true :size 32 :accessibility-label :send-message-button
+                                        :on-press #(do (swap! context assoc :clear true)
+                                                       (clear-input chat-id refs)
+                                                       (re-frame/dispatch [:chat.ui/send-current-message]))}
+                    :main-icons2/arrow-up]]])
                ;black background
                [reanimated/view {:style (reanimated/apply-animations-to-style
                                          {:opacity bg-opacity}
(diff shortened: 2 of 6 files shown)


src/quo/design_system/colors.cljs
```diff
@@ -179,8 +179,8 @@
                      purple
                      "#8B3131"])
 
-(def mention-incoming "#0DA4C9")
-(def mention-outgoing "#9EE8FA")
+(def mention-incoming "#4360DF")
+(def mention-outgoing "#4360DF")
```
Parveshdhull: we shouldn't make changes in `quo/...`. Please use colors from `quo2.foundations.colors`

src/status_im/ui/screens/chat/views.cljs
```diff
@@ -651,7 +651,8 @@
                      :show-input?                      show-input?}]
      ;;INPUT COMPONENT
      (when show-input?
-       [components/chat-input-bottom-sheet chat-id])]))
+       [rn/view
+        [components/chat-input-bottom-sheet chat-id] [components/autocomplete-mentions]])]))
```
flexsurfer: `autocomplete-mentions` should be a part of `chat-input-bottom-sheet` so it should be inside the component

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -705,11 +708,11 @@
                                                          (into #{} (js->clj own-reactions))
                                                          #(on-emoji-press %))}]))
         on-long-press   (atom nil)]
-    [:<>
+    [react/view   {:style (merge (when mentioned {:background-color "rgba(67,96,223,0.05)" :border-radius 16 :margin-bottom 4}) {:margin-horizontal 8})}
```
flexsurfer: color from quo2 should be used

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -130,13 +130,16 @@
...
-    (conj acc [react/text-class
-               {:style    {:color (cond
-                                    (= content-type constants/content-type-system-text) colors/black
-                                    :else                                               colors/mention-incoming)}
-                :on-press (when-not (= content-type constants/content-type-system-text)
-                            #(re-frame/dispatch [:chat.ui/show-profile literal]))}
-               [mention-element literal]])
+    (conj acc
+          [react/view {:style {:background-color "rgba(67,96,223,0.1)" :border-radius 6 :padding-horizontal 3}}
+           [react/text-class
```
flexsurfer, line 139: same for color and `font-weight` should be used from quo2 foundation

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -130,13 +130,16 @@
...
+            {:style    {:color (cond
+                                 (= content-type constants/content-type-system-text) colors/black
+                                 :else                                               colors/mention-incoming)
+                        :font-weight (cond (= content-type constants/content-type-system-text) "400" :else "500")}
```
flexsurfer: why cond and not if ?

src/status_im/ui/screens/chat/components/style.cljs
```diff
@@ -143,7 +143,8 @@
    :bottom           bottom
    :background-color (colors/get-color :ui-background)
    :border-top-width 1
-   :border-top-color (colors/get-color :ui-01)})
+   :border-top-color (colors/get-color :ui-01)}
+  :z-index          400)
```
flexsurfer: something seems off here

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -389,6 +388,22 @@
             :render-data               text-input-ref
             :render-fn                 mention-item}]]]))))
 
+(defn autocomplete-mentions []
+  [:f>
+   (fn []
+     (let [suggestions @(re-frame/subscribe [:chat/mention-suggestions])
```
flexsurfer: pls use `<sub` instead of `@(re-frame/subscribe`

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -564,18 +590,18 @@
...
             (let [reply (<sub [:chats/reply-message])
+                  suggestions (<sub [:chat/mention-suggestions])
                   {window-height :height} (rn/use-window-dimensions)
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
-
                   max-y (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets)) ; 360 - default height
                   max-height (- max-y 56 (:bottom insets))  ; 56 - top-bar height
-                  added-value (if reply 38 0) ; increased height of input box needed when reply
-                  min-y (+ min-y added-value)
+                  added-value (if (and (not (seq suggestions)) reply) 38 0) ; increased height of input box needed when reply
```
flexsurfer, line 603: are we sure we want to hide reply in case of suggestions ?

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -564,18 +590,18 @@
...
+                  min-y (+ min-y (when reply 38))
                   y (calculate-y context keyboard-shown min-y max-y added-value)
```
flexsurfer: could we have all calculations with conditions in `calculate-y` ? so we have only one calculate-y function ? because we 'll have more options

src/status_im/ui/screens/chat/styles/message/message.cljs
```diff
@@ -171,21 +171,8 @@
    :flex-direction :row-reverse})
 
 (defn message-view
-  [{:keys [content-type mentioned pinned]}]
+  [{:keys [content-type]}]
   (merge
-   {:border-radius 10}
-
-   (cond
-     pinned                                             {:background-color colors/pin-background}
```
flexsurfer, line 179: how pinned related to this PR ?

src/status_im/ui/screens/chat/views.cljs
OmarBasem: @flexsurfer according to the UI design I believe it shouldn't be. If the chat input is expanded, the mentions list still needs to animate up and down from the very bottom of the screen just above the keyboard. So its position needs to be independent of the chat-input.

src/status_im/ui/screens/chat/message/message.cljs
OmarBasem: The `color` property was already using `cond` so I thought would also use `cond` for `font-weight`. Shall I change it to `if`?

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: @flexsurfer in case of suggestions we don't need the `added-value` to `y` when replying because the bottom buttons view will not be rendered, so that will make up for the difference.

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: @flexsurfer `added-value` was already inside `calculate-y` in the reply component issue, but we moved outside to use it to update the input height.

src/status_im/ui/screens/chat/styles/message/message.cljs
OmarBasem: Oh it's not related. I was working on the pins issue at the same time. These lines got removed anyways.

src/status_im/ui/screens/chat/views.cljs
flexsurfer: yeah i mean it still can be implemented as here, but we need to move `[rn/view` and `[components/autocomplete-mentions]` inside `chat-input-bottom-sheet` so from here it should be as one component, because currently they are related, but `splited` and used  from outside ns

src/status_im/ui/screens/chat/message/message.cljs
flexsurfer: yeah better to use `if`

flexsurfer: probably `cond` was used by mistake

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: i mean move `calculate-y-with-mentions` code to `calculate-y` function, so we have only one

src/status_im/ui/screens/chat/message/message.cljs
OmarBasem: Didn't find any font-weight in quo2 foundation

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: @flexsurfer there are many calculations in both of them, and having both of them in one function maybe difficult to understand, so I created a function `get-y-value` that calls both of them.

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -331,47 +331,46 @@
                                   [idx item])
                                 @(re-frame/subscribe [:chat/input-with-mentions]))]
          ^{:key (str idx "_" type "_" text)}
-         [rn/text (when (= type :mention) {:style {:color "#0DA4C9"}})
-          text])
+         [rn/text (when (= type :mention) {:style {:color (:text-04 @colors/theme)}}) text])
```
J-Son89: should this be using the new text component from quo2?

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -389,6 +388,21 @@
             :render-data               text-input-ref
             :render-fn                 mention-item}]]]))))
 
+(defn autocomplete-mentions [suggestions]
+  [:f>
```
J-Son89, line 393: this file is huge. Is there a clean way to break it up?

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -331,47 +331,46 @@
...
+                   name])]
+               :title-text-weight :medium}
+
+        ens-name?
+        (assoc :subtitle alias))]]))
 
 (def chat-toolbar-height (reagent/atom nil))
 
-(defn autocomplete-mentions [text-input-ref bottom]
+(defn autocomplete-mentions-old [text-input-ref bottom]
```
J-Son89, line 374: I guess this is where the migration strategy we discussed on Tuesday needs to start taking effect, 

If I recall correctly we want to separate old and new via the namespace, as such these additions should be in a different file.
However, I don't think we completely agreed on the how we divide these namespaces out. 
Can we make a decision so that it can be followed in this p.r (if needed)
@Parveshdhull  @flexsurfer  @ibrkhalil @ilmotta  @jo-mut

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -607,19 +635,21 @@
...
-                                                     (clear-input chat-id refs)
-                                                     (re-frame/dispatch [:chat.ui/send-current-message]))}
-                  :main-icons2/arrow-up]]]
+               (when-not (seq suggestions)
+                 [rn/view {:style (styles/new-bottom-sheet-controls insets)}
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/image]
+                  [rn/view {:width 12}]
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/reaction]
+                  [rn/view {:flex 1}]
+                  ;;SEND button
```
J-Son89, line 646: imo this old comment is redundant and can be removed

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -607,19 +635,21 @@
...
+                  [rn/view {:ref send-ref :style (when-not (seq (get @input-texts chat-id)) {:width 0 :right -100})}
+                   [quo2.button/button {:icon     true :size 32 :accessibility-label :send-message-button
+                                        :on-press #(do (swap! context assoc :clear true)
+                                                       (clear-input chat-id refs)
+                                                       (re-frame/dispatch [:chat.ui/send-current-message]))}
+                    :main-icons2/arrow-up]]])
                ;black background
```
J-Son89, line 653: imo this old comment is redundant and should be removed 😆

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -334,7 +337,7 @@
           content]
          content)
        [link-preview/link-preview-wrapper (:links (:content message)) false false]]]
-   ; delivery status
+     ; delivery status
```
J-Son89, line 341: imo this old comment is also redundant

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -607,19 +635,21 @@
...
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/image]
+                  [rn/view {:width 12}]
+                  [quo2.button/button {:icon true :type :outline :size 32} :main-icons2/reaction]
+                  [rn/view {:flex 1}]
+                  ;;SEND button
+                  [rn/view {:ref send-ref :style (when-not (seq (get @input-texts chat-id)) {:width 0 :right -100})}
+                   [quo2.button/button {:icon     true :size 32 :accessibility-label :send-message-button
+                                        :on-press #(do (swap! context assoc :clear true)
+                                                       (clear-input chat-id refs)
+                                                       (re-frame/dispatch [:chat.ui/send-current-message]))}
```
J-Son89: I believe we have agreed to use `>evt` for re/frame dispatches so can you update this here

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -131,13 +133,14 @@
...
-                :on-press (when-not (= content-type constants/content-type-system-text)
-                            #(re-frame/dispatch [:chat.ui/show-profile literal]))}
-               [mention-element literal]])
+    (conj acc
+          [react/view {:style {:background-color quo2.colors/primary-50-opa-10 :border-radius 6 :padding-horizontal 3}}
+           [react/text-class
+            {:style    (merge {:color (if (= content-type constants/content-type-system-text) colors/black (:text-04 @colors/theme))}
+                              (if (= content-type constants/content-type-system-text) typography/font-regular typography/font-medium))
+             :on-press (when-not (= content-type constants/content-type-system-text)
+                         #(re-frame/dispatch [:chat.ui/show-profile literal]))}
```
J-Son89: same here, can you refactor to >evt method instead

src/status_im/ui/screens/chat/components/style.cljs
```diff
@@ -143,7 +143,8 @@
    :bottom           bottom
    :background-color (colors/get-color :ui-background)
    :border-top-width 1
-   :border-top-color (colors/get-color :ui-01)})
+   :border-top-color (colors/get-color :ui-01)}
+  :z-index          3)
```
ibrkhalil: align map

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: Thank you for your reviews @J-Son89 
There are multiple files that I would like to break up, but it will cause circular dependencies.
I think it will get smaller when we refactor the old code.

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: Why is it redundant? 🤔

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: How is it redundant? 🤨

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: I was wondering wouldn't it better to call the alias >dispatch since it "dispatches an event", like <sub alias "subscribes to an event"

src/status_im/ui/screens/chat/components/style.cljs
OmarBasem: Thanks!

src/status_im/ui/screens/chat/components/style.cljs
```diff
@@ -155,7 +156,7 @@
           :height                  window-height
           :flex                    1
           :background-color        (quo2.colors/theme-colors quo2.colors/white quo2.colors/neutral-90)
-          :z-index                 1000}
+          :z-index                 2}
```
flexsurfer, line 159: are you sure we should have two z-indexes with value 2 ? one was 1000 before and another 2000

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: i'll file a PR soon there I'll introduce new namespace for chat , i'll help with rebasing this PR

src/status_im/ui/screens/chat/components/input.cljs
OmarBasem: > I was wondering wouldn't it better to call the alias >dispatch since it "dispatches an event", like <sub alias "subscribes to an event"

cc @flexsurfer

src/status_im/ui/screens/chat/components/input.cljs
J-Son89: because if I look at the code below I can extract this information from code:
```
[quo2.button/button {:icon     true :size 32 :accessibility-label :send-message-button
                                        :on-press #(do (swap! context assoc :clear true)
                                                       (clear-input chat-id refs)
                                                       (re-frame/dispatch [:chat.ui/send-current-message]
```
e.g from the dispatch and the accessibility-label

src/status_im/ui/screens/chat/components/input.cljs
J-Son89: We can see from the code that this codeblock is related to background - e.g `styles/new-bottom-sheet-background`. So it does not really add any information that we can not get without looking at the code directly

J-Son89 (approved):

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: yes i agree >evt is strange, but we should do refactoring and replace all >evt usage, not in this PR

src/status_im/ui/screens/chat/components/input.cljs
```diff
@@ -331,47 +332,46 @@
...
+                 :size            :small}
+                (if nickname
+                  nickname
+                  name)
+                (when nickname
+                  [text/text
+                   {:weight         :regular
+                    :color          :secondary
+                    :ellipsize-mode :tail
+                    :size           :small}
```
flexsurfer, line 362: hey @OmarBasem there is no size :small and it crashes the app for some reason

flexsurfer:
@OmarBasem please do not merge PR until it approved by dev and design teams

src/status_im/ui/screens/chat/components/input.cljs
flexsurfer: oh i see its text from quo1

src/status_im/ui/screens/chat/components/style.cljs
OmarBasem: Yes, I tested it. Why were we having 1000 and 2000 before? :D

OmarBasem:
@flexsurfer sure, where are we doing design reviews?

Issue #14024, opened 2022-09-16
Allow to unpin messages - New UI


cammellos:
As a user
I want to be able to unpin a message
So that I can remove non-relevant information

This should be implemented for 1-to-1, private group chats & communities.

Depends on: #14023 #14022 
### Designs

https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=3512%3A395567

OmarBasem:
Implementation complete. Waiting for [dependency PR](https://github.com/status-im/status-mobile/pull/14158).

Issue #14023, opened 2022-09-16
Show pinned messages - New UI


cammellos:
As a user
I want to be able to see pinned messages
So that I know the important information about a community/channel

This needs to be implemented for 1-to-1, private group chats & communities.

### Designs

https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=3425%3A413546

This includes showing the banner on top with the latest pinned message,  the number of pinned messages, and showing the pinned messages https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=3505%3A368386

Depends on: #14022 

Component dependency:
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=863%3A26323 Banner component

OmarBasem:
Implementation of this issue is complete. Waiting for [dependency PR](https://github.com/status-im/status-mobile/pull/14147) to be merged first.

PR #14147, opened 2022-10-11
feat: pinned messages new ui

### LINKED ISSUE #14022: Allow to pin messages - New UI
As a user
I want to be able to pin a message
So that is visible to other community users

This should be implemented for 1-to-1, private group chats and communities.

The behavior is that by default, only admin of the community should be able to pin messages. There's an option for users to pin messages, but currently we don't allow set that option in the UI when creating communities.

### Designs

https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=3425%3A413502

OmarBasem:
fixes #14022 

### Summary

This PR implements the [new UI ](https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=3425%3A413502) for pinned messages. It includes the new UI for a pinned message as well as the message that appears in the chat stating that a user has pinned a message. Showing a list of pinned messages, and unpinning messages are addressed in different issues (#14023 and #14024).

#### Platforms

- Android
- iOS

status: ready

### DIFF
--- a/src/status_im/chat/models/pin_message.cljs
+++ b/src/status_im/chat/models/pin_message.cljs
@@ -4,6 +4,7 @@
             [status-im.data-store.pin-messages :as data-store.pin-messages]
             [status-im.utils.fx :as fx]
             [taoensso.timbre :as log]
+            [status-im.transport.message.protocol :as protocol]
             [re-frame.core :as re-frame]))
 
 (fx/defn handle-failed-loading-pin-messages
@@ -73,7 +74,8 @@
   {:events [::send-pin-message]}
   [{:keys [db] :as cofx} {:keys [chat-id message-id pinned] :as pin-message}]
   (let [current-public-key (get-in db [:multiaccount :public-key])
-        message (merge pin-message {:pinned-by current-public-key})]
+        message (merge pin-message {:pinned-by current-public-key})
+        preferred-name (get-in db [:multiaccount :preferred-name])]
     (fx/merge cofx
               {:db (cond-> db
                      pinned
@@ -86,7 +88,13 @@
                       (update-in [:pin-messages chat-id] dissoc message-id)))}
               (data-store.pin-messages/send-pin-message {:chat-id    (pin-message :chat-id)
                                                          :message_id (pin-message :message-id)
-                                                         :pinned     (pin-message :pinned)}))))
+                                                         :pinned     (pin-message :pinned)})
+              (when pinned
+                (protocol/send-chat-messages [{:chat-id      (pin-message :chat-id)
+                                               :content-type constants/content-type-pin
+                                               :text "This is a pin!"
+                                               :response-to (pin-message :message-id)
+                                               :ens-name preferred-name}])))))
 
 (fx/defn load-pin-messages
   {:events [::load-pin-messages]}
--- a/src/status_im/chat/models/pin_message.cljs
+++ b/src/status_im/chat/models/pin_message.cljs
@@ -92,7 +92,6 @@
               (when pinned
                 (protocol/send-chat-messages [{:chat-id      (pin-message :chat-id)
                                                :content-type constants/content-type-pin
-                                               :text "This is a pin!"
                                                :response-to (pin-message :message-id)
                                                :ens-name preferred-name}])))))
 
--- a/src/status_im/ui/screens/chat/components/reply.cljs
+++ b/src/status_im/ui/screens/chat/components/reply.cljs
@@ -81,15 +81,16 @@
        [icons/icon :main-icons/close-circle {:container-style (styles/close-button)
                                              :color (:icon-02 @quo.colors/theme)}]]]]))
 
-(defn reply-message [{:keys [from identicon content-type contentType parsed-text content]} in-chat-input?]
+; This component is also used for quoted pinned message as the UI is very similar
+(defn reply-message [{:keys [from identicon content-type contentType parsed-text content]} in-chat-input? pin?]
   (let [contact-name       @(re-frame/subscribe [:contacts/contact-name-by-identity from])
         current-public-key @(re-frame/subscribe [:multiaccount/public-key])
         content-type       (or content-type contentType)]
-    [rn/view {:style {:flex-direction :row :height 24}}
-     [rn/view {:style (styles/reply-content)}
-      [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
-                                         :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}]
-      [rn/view {:style {:position :absolute :left 34 :top 3 :flex-direction :row :align-items :center :width "45%"}}
+    [rn/view {:style {:flex-direction :row :height (when-not pin? 24)}}
+     [rn/view {:style (styles/reply-content pin?)}
+      (when-not pin? [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
+                                                        :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}])
+      [rn/view {:style (styles/quoted-message pin?)}
        [photos/member-photo from identicon 16]
        [quo2.text/text {:weight          :semi-bold
                         :size            :paragraph-2
@@ -125,7 +126,7 @@
 (defn send-image [images]
   [rn/view {:style (styles/reply-container-image)}
    [rn/scroll-view {:horizontal true
-                    :style      (styles/reply-content)}
+                    :style      (styles/reply-content false)}
     (for [{:keys [uri]} (vals images)]
       ^{:key uri}
       [rn/image {:source {:uri uri}
--- a/src/status_im/ui/screens/chat/components/style.cljs
+++ b/src/status_im/ui/screens/chat/components/style.cljs
@@ -115,10 +115,10 @@
 
 (defn quoted-message [pin?]
   (merge {:flex-direction :row
-          :align-items :center}
+          :align-items :center
+          :width "45%"}
          (when-not pin? {:position :absolute
                          :left 34
-                         :right 54
                          :top 3})))
 
 (defn contact-request-content []
--- a/src/status_im/ui/screens/chat/components/style.cljs
+++ b/src/status_im/ui/screens/chat/components/style.cljs
@@ -108,10 +108,18 @@
    :padding-horizontal 10
    :flex               1})
 
-(defn reply-content []
-  {:padding-horizontal 10
+(defn reply-content [pin?]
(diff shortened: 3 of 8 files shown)


du82:
If I pin a message and then unpin it, the pinned system message doesn't go away, which allows for a system message duplication glitch

https://user-images.githubusercontent.com/65834069/195135856-dbfafd3d-5993-4821-9437-688a85102f13.mp4

OmarBasem:
@du64 I was thinking whether the system message should be deleted or not, so I checked how discord does that. The systems messages are not deleted when unpinned, however, they can be deleted by the user (yet to be implemented). Will also confirm with Pedro.

du82:
> @du64 I was thinking whether the system message should be deleted or not, so I checked how discord does that. The systems messages are not deleted when unpinned, however, they can be deleted by the user (yet to be implemented). Will also confirm with Pedro.

@OmarBasem I dont like the way Discord does system messages, because if the user pins something and thrn unpins it without removing the system message, people will easily be confused.

The same thing happens with mentions and DMs, where the original will be removed, but you'll still recieve the in-app info that it happened

OmarBasem:
@du64 checked with the design team, and they told me it's a log message so it shouldn't be deleted automatically.

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -81,15 +81,16 @@
        [icons/icon :main-icons/close-circle {:container-style (styles/close-button)
                                              :color (:icon-02 @quo.colors/theme)}]]]]))
 
-(defn reply-message [{:keys [from identicon content-type contentType parsed-text content]} in-chat-input?]
+; This component is also used for quoted pinned message as the UI is very similar
```
J-Son89, line 84: Is this really needed?

src/status_im/ui/screens/chat/components/reply.cljs
```diff
@@ -81,15 +81,16 @@
...
   (let [contact-name       @(re-frame/subscribe [:contacts/contact-name-by-identity from])
         current-public-key @(re-frame/subscribe [:multiaccount/public-key])
         content-type       (or content-type contentType)]
-    [rn/view {:style {:flex-direction :row :height 24}}
-     [rn/view {:style (styles/reply-content)}
-      [icons/icon :main-icons/connector {:color (theme-colors quo2.colors/neutral-40 quo2.colors/neutral-60)
-                                         :container-style {:position :absolute :left 10 :bottom -4 :width 16 :height 16}}]
-      [rn/view {:style {:position :absolute :left 34 :top 3 :flex-direction :row :align-items :center :width "45%"}}
+    [rn/view {:style {:flex-direction :row :height (when-not pin? 24)}}
+     [rn/view {:style (styles/reply-content pin?)}
```
J-Son89, line 90: consider using explicit variable names, character saving like this adds little value in my opinion.
.ie `pinned?` in place of `pin?`
unless of course, `pin` is more accurate here 😛

src/status_im/ui/screens/chat/components/style.cljs
```diff
@@ -108,11 +108,19 @@
    :padding-horizontal 10
    :flex               1})
 
-(defn reply-content []
-  {:padding-horizontal 10
+(defn reply-content [pin?]
```
J-Son89, line 111: same here

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -90,9 +94,9 @@
         timestamp-str]])))
 
 (defview quoted-message
-  [_ reply]
-  [react/view {:style (style/quoted-message-container)}
-   [components.reply/reply-message reply false]])
+  [_ reply pin?]
```
J-Son89, line 98: and here

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -131,13 +135,14 @@
...
-               {:style    {:color (cond
-                                    (= content-type constants/content-type-system-text) colors/black
-                                    :else                                               colors/mention-incoming)}
-                :on-press (when-not (= content-type constants/content-type-system-text)
-                            #(re-frame/dispatch [:chat.ui/show-profile literal]))}
-               [mention-element literal]])
+    (conj acc
+          [react/view {:style {:background-color quo2.colors/primary-50-opa-10 :border-radius 6 :padding-horizontal 3}}
+           [react/text-class
+            {:style    (merge {:color (if (= content-type constants/content-type-system-text) colors/black quo2.colors/primary-50)}
```
J-Son89: consider a helper 
e.g
```
(defn is-system-text? [content-type] 
(= content-type constants/content-type-system-text))
```

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -131,13 +135,14 @@
...
+                              (if (= content-type constants/content-type-system-text) typography/font-regular typography/font-medium))
+             :on-press (when-not (= content-type constants/content-type-system-text)
+                         #(re-frame/dispatch [:chat.ui/show-profile literal]))}
```
J-Son89: can you update the dispatch to >evt 🙏

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -220,28 +225,22 @@
   (let [user-contact @(re-frame/subscribe [:multiaccount/contact])
         contact-names @(re-frame/subscribe [:contacts/contact-two-names-by-identity pinned-by])]
     ;; We append empty spaces to the name as a workaround to make one-line and multi-line label components show correctly
-    (str "                   " (if (= pinned-by (user-contact :public-key)) (i18n/label :t/You) (first contact-names)))))
+    (str " " (if (= pinned-by (user-contact :public-key)) (i18n/label :t/You) (first contact-names)))))
```
J-Son89, line 232: nit: as far as I can see all translation label keys are lower case expect this one. In my opinion we should adjust for consistency.

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -296,7 +295,7 @@
   [{:keys [last-in-group?
            identicon
            from in-popover? timestamp-str
-           deleted-for-me?]
+           deleted-for-me? pinned]
```
J-Son89, line 302: if it's a boolean, consider using `pinned?`

src/status_im/ui/screens/chat/components/reply.cljs
OmarBasem: What do you mean exactly?

src/status_im/ui/screens/chat/components/reply.cljs
OmarBasem: So here `pin?` like `isPin`. I am using it as a noun rather than a verb to refer to the new system message showing that a user has pinned a message - not the pinned message itself.

src/status_im/ui/screens/chat/message/message.cljs
OmarBasem: Thanks!

src/status_im/ui/screens/chat/message/message.cljs
OmarBasem: I see the translation files have both `:t/you` and `:t/You`. So it seems intentional.

src/status_im/ui/screens/chat/message/message.cljs
OmarBasem: That's a key extracted from the message object map

src/status_im/ui/screens/chat/components/reply.cljs
J-Son89: ok, personally I would use `is-pin?` here so.

src/status_im/ui/screens/chat/components/reply.cljs
J-Son89: I personally find comments like this unnecessary. It is pretty easy to read this from the code and see that this is the case. Comments can often go out of date when code gets updated and mislead developers. Code is the source of truth. There's nothing in the code that really needs a comment to explain here.

Just my opinion 👍

J-Son89 (approved):

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -90,9 +95,12 @@
...
 
 (defview quoted-message
-  [_ reply]
-  [react/view {:style (style/quoted-message-container)}
-   [components.reply/reply-message reply false]])
+  [_ reply pin?]
+  [react/view {:style (when-not pin? (style/quoted-message-container))}
+   [components.reply/reply-message reply false pin?]])
+
+(defn is-system-text? [content-type]
```
ilmotta: I think predicates in Clojure never use the `is-`, `are-`, etc prefixes because the question mark is a strong convention, especially for function predicates.

src/status_im/chat/models/pin_message.cljs
```diff
@@ -4,6 +4,7 @@
             [status-im.data-store.pin-messages :as data-store.pin-messages]
             [status-im.utils.fx :as fx]
             [taoensso.timbre :as log]
+            [status-im.transport.message.protocol :as protocol]
             [re-frame.core :as re-frame]))
 
 (fx/defn handle-failed-loading-pin-messages
```
ilmotta, line 10: Just mentioning this for reference, but we're trying to slowly add more test coverage, particularly for things that mess with the app state and non-UI stuff. They're easier to test, and if we get them under coverage, we cover a lot of important ground.

I know this namespace already existed without tests, but in future PRs, it'd be great to see tests, at least for the event handlers that changed.

ilmotta (approved):

src/status_im/ui/screens/chat/components/reply.cljs
OmarBasem: Same, but I believe it is a Clojure convention to replace `is-` with a `?` at the end.

src/status_im/ui/screens/chat/components/reply.cljs
OmarBasem: That's right. Here the component name is `reply-message` which I also use for the new system pin message, so thought would add a comment to make it clearer.

src/status_im/chat/models/pin_message.cljs
OmarBasem: Thanks for your review @ilmotta.

Cool will add tests in future PRs. Did we do any setup for this?

ilmotta: As usual with cljs, you'll unfortunately need to run over a bit more steps to run them, especially if you want to run them in the REPL.

jo-mut (approved):

PR #14156, opened 2022-10-13
Introduce new ui2 namespace, move chat to new namespace and remove new UI toggle from UI

flexsurfer:
enable new ui by default

- `new-ui-enabled?` is not an atom anymore, its `true` by default, to show the old UI, devs should change it and reopen the app
- `new-ui-enabled?` is still used for navigation and the home screen (we should remove it for the home screen as well when home screens will be implemented)
- new `ui2` namespace is introduced


src/status_im/utils/config.cljs
```diff
@@ -158,29 +153,18 @@
...
 (def default-wallet-connect-metadata {:name "Status Wallet"
                                       :description "Status is a secure messaging app, crypto wallet, and Web3 browser built with state of the art technology."
                                       :url "#"
                                       :icons ["https://statusnetwork.com/img/press-kit-status-logo.svg"]})
 
 (def wallet-connect-project-id "87815d72a81d739d2a7ce15c2cfdefb3")
 
-(def new-ui-enabled? (atom false))
+;;TODO for development only should be removed in status 2.0
+(def new-ui-enabled? false)
```
Parveshdhull: this should be true

src/status_im/core.cljs
```diff
@@ -40,8 +40,6 @@
 
   (utils.universal-links/initialize)
 
-  (async-storage/get-item :new-ui-enabled? #(reset! config/new-ui-enabled? %))
-
```
Parveshdhull, line 44: I think we should not remove storing of this flag. We are opting for enabling new ui by default but if we toggle to the old ui for testing or something, it should not revert automatically on app restart.

src/status_im/ui/screens/chat/message/message.cljs
```diff
@@ -1,35 +1,32 @@
 (ns status-im.ui.screens.chat.message.message
-  (:require [quo.core :as quo]
```
flexsurfer, line 2: this file hasn't been changed , it was reamed, and moved, that why git shows changes , but its just different file

src/status_im/ui2/screens/chat/composer/mentions.cljs
```diff
@@ -0,0 +1 @@
+(ns status-im.ui2.screens.chat.composer.mentions)
```
flexsurfer: cc @OmarBasem

src/status_im/utils/config.cljs
flexsurfer: oh good catch, i just tested before merge :)

src/status_im/core.cljs
Parveshdhull: or are we removing toggling functionality from UI completely

flexsurfer: there is no toggle anymore

Parveshdhull: that's good :D

J-Son89: I think we had a concern that the QA team had tests using the old ui. I think we need to discuss these changes with them before merging this.

@qoqobolo, @churik

src/status_im/ui/screens/profile/visibility_status/utils.cljs
```diff
@@ -91,27 +88,17 @@
...
 
 (defn icon-dot-margin
-  [size identicon?]
-  (if @config/new-ui-enabled?
-    -2
-    (if identicon?
-      (/ size 6)
-      (/ size 7))))
+  [_ _]
+  -2)
```
Parveshdhull, line 97: So even if we opt for old ui, these params will be still used as per new ui?

ilmotta (approved):

src/status_im/ui/screens/profile/visibility_status/utils.cljs
flexsurfer: yes, because we don't want to support old ui anymore, this one was tricky because it somehow was implemented without new design, so because we won't support old ui anymore i decided to keep only new ui even if it was not properly implemented

flexsurfer: we shouldn't change old ui anymore

Parveshdhull (approved):
PR looks good to me, please request QA testing before merging to make sure that all flows are working and develop is not blocked 🙏.
Thank you for the great work

flexsurfer:
@Parveshdhull thank you. There is not much to test actually, because the new UI is default, and we don't care about old UI , so e2e should be enough, i tested manually the app

flexsurfer:
after a discussion with @churik we decided to postpone new ui by default for now, because we have still "old" PRs, and e2e tests are not ready, so devs should enable it in local code for now

J-Son89 (approved):

src/status_im/ui2/screens/chat/composer/mentions.cljs
OmarBasem: I hope this will not cause a circular dependency. 
The mentions PR is merged.

churik:
@flexsurfer is there a way to enable new UI from installed application?

flexsurfer:
@churik no, i thought that we merge this PR last week, but as we're still here , i'll enable it in the code so only new design will be available after we merge this PR

OmarBasem (approved):

PR #14157, opened 2022-10-13
feat: add drawers - action drawers component to quo2 library

### LINKED ISSUE #14143: Implement Action Drawer component
As a develop there is often the use case to create these menus in bottom sheets.

Implement the Action Drawer component in the quo2 library to the design spec

Figma https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=9535%3A164576

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14143

the designs for this were 99% approved so I was referencing this file: 
https://www.figma.com/file/apCsJHT8J58JXbJD2igtbd/Context-Drawers-for-Mobile?node-id=167%3A107697

AS the standard Figma file is yet to be updated with some changes.


J-Son89:
dark mode:


light mode

src/quo2/screens/drawers/action_drawers.cljs
```diff
@@ -0,0 +1,48 @@
...
+  (let [state         (reagent/atom {:muted?         true
+                                     :show-red-options? true})]
+    (fn []
+      [rn/touchable-without-feedback {:on-press rn/dismiss-keyboard!}
+       [rn/view {:padding-bottom 400}
+        [preview/customizer state descriptor]
+        [rn/view {:padding-vertical 60}
+         [quo2/action-drawer {:actions-with-consequence (when (:show-red-options? @state) options-with-consequences)
+                              :actions [{:icon :main-icons2/share-context
+                                         :label "View channel members and details"}
```
jo-mut, line 28: why not use `(i18n/label "some string")`

jo-mut (changes requested):

ilmotta (approved):
Apart from code formatting (which we'll soon be able to fix automatically), LGTM.

src/quo2/components/drawers/action_drawers.cljs
```diff
@@ -0,0 +1,78 @@
...
+           :size 20}]])]]))
+
+(defn action-drawer [{:keys [actions actions-with-consequence]}]
+  [react/view {:style
+               {:flex 1}}
+   [list/flat-list
+    {:key-fn (fn [_ index] index)
+     :list-key "actions-"
+     :data                              actions
+     :render-fn                         (action :black)}]
```
ilmotta: It'd be even better if we could use semantic colors (primary, secondary, etc) instead of actual color names.

Edit: same for the `:red` color option.

flexsurfer:
im not sure we should use flatlist for such cases, better to just use `[:<>` and `for` , also bottom sheet has a scroll view,

jo-mut (approved):

src/quo2/screens/drawers/action_drawers.cljs
```diff
@@ -0,0 +1,48 @@
...
+
+(defn cool-preview []
+  (let [state         (reagent/atom {:muted?         true
+                                     :show-red-options? true})]
+    (fn []
+      [rn/touchable-without-feedback {:on-press rn/dismiss-keyboard!}
+       [rn/view {:padding-bottom 400}
+        [preview/customizer state descriptor]
+        [rn/view {:padding-vertical 60}
+         [quo2/action-drawer {:actions-with-consequence (when (:show-red-options? @state) options-with-consequences)
```
flexsurfer: would be also great to have a button to open drawer in the bottom sheet

flexsurfer (changes requested):

src/quo2/screens/drawers/action_drawers.cljs
J-Son89: yeah I think it's a good suggestion 👍

src/quo2/components/drawers/action_drawers.cljs
J-Son89: I don't really know what those semantic colors are. It's just about whether it's top section or bottom. I guess I can change to :top and :bottom 🤔

ilmotta: Sounds good to me @J-Son89. Imagine when (not if) the design team changes from red to orange, then the code won't make much sense.

J-Son89: yep, I completely agree :)

J-Son89:
> im not sure we should use flatlist for such cases, better to just use `[:<>` and `for` , also bottom sheet has a scroll view,

I used map, but is there any specific reason I should use `for`?

src/quo2/components/drawers/action_drawers.cljs
```diff
@@ -0,0 +1,68 @@
...
+                     {:height 20
+                      :margin-top :auto
+                      :margin-bottom :auto
+                      :width 20}}
+         [icon/icon right-icon
+          {:color (get-icon-color section)
+           :size 20}]])]]))
+
+(defn action-drawer [{:keys [actions actions-with-consequence]}]
+  [react/view {:style
```
flexsurfer: `[:<>` could be used

flexsurfer (approved):
thank you

flexsurfer:
> I used map, but is there any specific reason I should use for?

no :)

src/quo2/components/drawers/action_drawers.cljs
J-Son89: done :)

Issue #13873, opened 2022-08-29
Implement “Delete for me” option for messages


briansztamfater:
# Feature Issue

### User Story

As a user, I want to delete messages locally **so that** I can remove message I don't want to see without removing them to the other chat participants. 

### Description

Flow TBD

### Acceptance Criteria

TBD

### Notes

Link to Message Context Drawer in Figma: https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=4388%3A631636

<img width="275" src="https://user-images.githubusercontent.com/18485527/187117579-9c0bebe4-3633-4cf5-a8ea-938e92bca472.png">

yqrashawn:
deleted for me messages ~https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments?node-id=4136%3A421265~
https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=5713%3A624211

yqrashawn:
- [x] menu UI #14007 
- [x] deleted message UI (using system message component) #14168
- [x] deleted for me notification with undo UI, waiting on toast component https://github.com/status-im/status-mobile/issues/13831
- [x] delete/undo logic #14007
- [x] status-go https://github.com/status-im/status-go/pull/2866 (version 0.111.1 already included in status-mobile)
- [x] temp UI indicating message deleted for me (deleted for me messages have a red border)

yqrashawn:
completed in https://github.com/status-im/status-mobile/pull/14618

PR #14168, opened 2022-10-17
feat: deleted for me message UI

yqrashawn:
related issue: #13873

This PR 
1. add the "Message deleted for you" system message in chat view
2. add the logic of sync deleted messages immediately after leaving chat view

simulator_screenshot_37ECF2E6-732E-40E5-93A1-C3ED9E997965

status: ready

### DIFF
--- a/src/quo2/components/messages/system_message.cljs
+++ b/src/quo2/components/messages/system_message.cljs
@@ -1,14 +1,13 @@
 (ns quo2.components.messages.system-message
-  (:require [status-im.i18n.i18n :as i18n]
-            [quo.react-native :as rn]
-            [status-im.utils.core :as utils]
+  (:require [quo.react-native :as rn]
             [quo.theme :as theme]
-            [quo2.components.buttons.button :as button]
-            [quo2.components.markdown.text :as text]
-            [quo2.reanimated :as ra]
-            [quo2.foundations.colors :as colors]
+            [quo2.components.avatars.icon-avatar :as icon-avatar]
             [quo2.components.avatars.user-avatar :as user-avatar]
-            [quo2.components.avatars.icon-avatar :as icon-avatar]))
+            [quo2.components.markdown.text :as text]
+            [quo2.foundations.colors :as colors]
+            [quo2.reanimated :as ra]
+            [status-im.i18n.i18n :as i18n]
+            [status-im.utils.core :as utils]))
 
 (def themes-landed {:pinned  colors/primary-50-opa-5
                     :added   colors/primary-50-opa-5
@@ -55,7 +54,7 @@
 
 (defmulti sm-render :type)
 
-(defmethod sm-render :deleted [{:keys [state action timestamp-str]}]
+(defmethod sm-render :deleted [{:keys [label timestamp-str]}]
   [rn/view {:align-items     :center
             :justify-content :space-between
             :flex            1
@@ -64,15 +63,12 @@
              :flex-direction :row}
     [sm-icon {:icon    :main-icons/delete16
               :color   :danger
-              :opacity (if (= state :landed) 0 5)}]
+              :opacity 5}]
     [text/text {:size  :paragraph-2
                 :style {:color        (get-color :text)
                         :margin-right 5}}
-     (i18n/label (if action :message-deleted-for-you :message-deleted))]
-    (when (nil? action) [sm-timestamp timestamp-str])]
-   (when action [button/button {:size   24
-                                :before :main-icons/timeout
-                                :type   :grey} (i18n/label :undo)])])
+     (i18n/label (or label :message-deleted))]
+    [sm-timestamp timestamp-str]]])
 
 (defmethod sm-render :added [{:keys [state mentions timestamp-str]}]
   [rn/view {:align-items    :center
@@ -140,15 +136,19 @@
                     :style {:color (get-color :time)}}
          (utils/truncate-str (:info content) 24)])]]]])
 
-(defn system-message [{:keys [type] :as message}]
+(defn system-message
+  [{:keys [type non-pressable? animate-landing?] :as message}]
   [:f>
    (fn []
-     (let [sv-color (ra/use-shared-value (get-color :bg :landed type))]
-       (ra/animate-shared-value-with-delay
-        sv-color (get-color :bg :default type) 0 :linear 1000)
+     (let [sv-color (ra/use-shared-value
+                     (get-color :bg (if animate-landing? :landed :default) type))]
+       (when animate-landing?
+         (ra/animate-shared-value-with-delay
+          sv-color (get-color :bg :default type) 0 :linear 1000))
        [ra/touchable-opacity
-        {:on-press #(ra/set-shared-value
-                     sv-color (get-color :bg :pressed type))
+        {:on-press #(when-not non-pressable?
+                      (ra/set-shared-value
+                       sv-color (get-color :bg :pressed type)))
          :style    (ra/apply-animations-to-style
                     {:background-color sv-color}
                     {:flex-direction     :row
--- a/src/quo2/components/messages/system_message.cljs
+++ b/src/quo2/components/messages/system_message.cljs
@@ -137,7 +137,7 @@
          (utils/truncate-str (:info content) 24)])]]]])
 
 (defn system-message
-  [{:keys [type non-pressable? animate-landing?] :as message}]
+  [{:keys [type style non-pressable? animate-landing?] :as message}]
   [:f>
    (fn []
      (let [sv-color (ra/use-shared-value
@@ -151,12 +151,12 @@
                        sv-color (get-color :bg :pressed type)))
          :style    (ra/apply-animations-to-style
                     {:background-color sv-color}
-                    {:flex-direction     :row
-                     :flex               1
-                     :border-radius      16
-                     :padding-vertical   9
-                     :padding-horizontal 11
-                     :width              359
-                     :height             52
-                     :background-color   sv-color})}
+                    (merge
+                     {:flex-direction     :row
+                      :flex               1
+                      :border-radius      16
+                      :padding-vertical   9
+                      :padding-horizontal 11
+                      :background-color   sv-color}
+                     style))}
         [sm-render message]]))])
--- a/src/status_im/chat/models/delete_message_for_me.cljs
+++ b/src/status_im/chat/models/delete_message_for_me.cljs
@@ -1,5 +1,6 @@
 (ns status-im.chat.models.delete-message-for-me
   (:require [re-frame.core :as re-frame]
+            [status-im.chat.models.message-list :as message-list]
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.datetime :as datetime]
             [status-im.utils.fx :as fx]
@@ -43,17 +44,22 @@
   {:events [:chat.ui/delete-message-for-me]}
   [{:keys [db]} {:keys [chat-id message-id]} undo-time-limit-ms]
   (when (get-in db [:messages chat-id message-id])
-    {:db (update-db-delete-locally db chat-id message-id undo-time-limit-ms)
+    (assoc
+     (message-list/rebuild-message-list
+      {:db (update-db-delete-locally db chat-id message-id undo-time-limit-ms)}
+      chat-id)
      :utils/dispatch-later [{:dispatch [:chat.ui/delete-message-for-me-and-sync
                                         {:chat-id    chat-id
                                          :message-id message-id}]
-                             :ms       undo-time-limit-ms}]}))
+                             :ms       undo-time-limit-ms}])))
 
 (fx/defn undo
   {:events [:chat.ui/undo-delete-message-for-me]}
   [{:keys [db]} {:keys [chat-id message-id]}]
   (when (get-in db [:messages chat-id message-id])
-    {:db (update-db-undo-locally db chat-id message-id)}))
+    (message-list/rebuild-message-list
+     {:db (update-db-undo-locally db chat-id message-id)}
+     chat-id)))
 
 (fx/defn delete-and-sync
   {:events [:chat.ui/delete-message-for-me-and-sync]}
@@ -66,3 +72,19 @@
                        :on-error    #(log/error "failed to delete message for me " %)
                        :on-success  #(re-frame/dispatch [:sanitize-messages-and-process-response
                                                          %])}]}))
+(defn- chats-reducer
+  "traverse all messages find not yet synced deleted-for-me? messages, generate dispatch vector"
+  [acc chat-id messages]
+  (reduce-kv
+   (fn [inner-acc message-id {:keys [deleted-for-me? deleted-for-me-undoable-till]}]
+     (if (and deleted-for-me? deleted-for-me-undoable-till)
+       (conj inner-acc [:chat.ui/delete-message-for-me-and-sync chat-id message-id])
+       inner-acc))
+   acc
+   messages))
+
+(fx/defn sync-all
+  "Get all deleted-for-me messages that not yet synced with status-go and sync them"
+  {:events [:chat.ui/sync-all-deleted-for-me-messages]}
+  [{:keys [db]}]
+  {:dispatch-n (reduce-kv chats-reducer [] (:messages db))})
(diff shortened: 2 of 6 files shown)


flexsurfer:
hey @yqrashawn  i've merged PR with ui2 namespace, we shouldn't create -old anymore, also src/status_im/ui/screens/chat/message/message.cljs was moved to https://github.com/status-im/status-mobile/blob/develop/src/status_im/ui2/screens/chat/messages/message.cljs, probably it will be simpler to create new branch and PR, but its up to you, please ping me if you need any help

src/quo2/components/messages/system_message.cljs
```diff
@@ -68,11 +84,8 @@
     [text/text {:size  :paragraph-2
                 :style {:color        (get-color :text)
                         :margin-right 5}}
-     (i18n/label (if action :message-deleted-for-you :message-deleted))]
-    (when (nil? action) [sm-timestamp timestamp-str])]
-   (when action [button/button {:size   24
-                                :before :main-icons/timeout
-                                :type   :grey} (i18n/label :undo)])])
```
yqrashawn, line 75: The undo button is in toast. Need for undo button here

src/quo2/components/messages/system_message.cljs
```diff
@@ -140,15 +153,17 @@
...
+           container (if non-pressable? rn/view ra/touchable-opacity)]
        (ra/animate-shared-value-with-delay
         sv-color (get-color :bg :default type) 0 :linear 1000)
-       [ra/touchable-opacity
-        {:on-press #(ra/set-shared-value
-                     sv-color (get-color :bg :pressed type))
+       [container
+        {:on-press #(when-not non-pressable?
+                      (ra/set-shared-value
+                       sv-color (get-color :bg :pressed type)))
```
yqrashawn: Not allowed to press deleted message

src/status_im/chat/models/delete_message_for_me.cljs
```diff
@@ -43,17 +44,22 @@
...
+                             :ms       undo-time-limit-ms}])))
 
 (fx/defn undo
   {:events [:chat.ui/undo-delete-message-for-me]}
   [{:keys [db]} {:keys [chat-id message-id]}]
   (when (get-in db [:messages chat-id message-id])
-    {:db (update-db-undo-locally db chat-id message-id)}))
+    (message-list/rebuild-message-list
+     {:db (update-db-undo-locally db chat-id message-id)}
+     chat-id)))
```
yqrashawn, line 62: Need to rebuild message list, cause deleted message is system message (with no avatar)

src/status_im/chat/models/delete_message_for_me_test.cljs
```diff
@@ -2,114 +2,118 @@
...
-        (is (= (update-in result [:db :messages "chat-id" "message-id"] dissoc :deleted-for-me-undoable-till)
-               expected))
-        (is (-> (get-in result [:db :messages "chat-id" "message-id" :deleted-for-me-undoable-till])
-                (- timestamp)
-                js/Math.abs
-                (< 10)))))
-    (testing "should return nil if message in db"
-      (is (= (delete-message-for-me/delete {:db {:messages []}} message 1000)
-             nil)))))
+  (with-redefs [message-list/rebuild-message-list identity]
```
yqrashawn: mock `rebuild-message-list` in test

src/status_im/chat/models/message_list.cljs
```diff
@@ -16,16 +16,20 @@
...
-       :type :message
-       :message-id message-id
-       :outgoing (boolean outgoing)}
+       :from              from
+       :one-to-one?       (= constants/message-type-one-to-one message-type)
+       :system-message?   (boolean
+                           (or
+                            (= constants/message-type-private-group-system-message
+                               message-type)
+                            deleted-for-me?))
```
yqrashawn, line 28: Count deleted-for-me message as system message in status-mobile only.
It's still user's message in status-go, since we need to since this between paired devices.

src/status_im/ui2/screens/chat/messages/message.cljs
```diff
@@ -1,6 +1,9 @@
 (ns status-im.ui2.screens.chat.messages.message
   (:require [quo.core :as quo]
             [quo.design-system.colors :as colors]
+            [quo2.components.messages.system-message :as system-message]
```
yqrashawn, line 5: Rest of the `ns` form are reorders

src/status_im/ui2/screens/chat/messages/message.cljs
```diff
@@ -295,50 +296,52 @@
...
-         content)
-       [link-preview/link-preview-wrapper (:links (:content message)) false false]]]
-     ; delivery status
-     [react/view (style/delivery-status)
-      [message-delivery-status message]]]))
+    (if deleted-for-me?
+      [react/view {:style (style/deleted-for-me-message-body)}
+       [system-message/system-message
+        {:type :deleted-for-me :timestamp-str timestamp-str :non-pressble? true}]]
+      [react/view {:style               (style/message-wrapper message)
```
yqrashawn: Add deleted-for-me message at top of `message-content-wrapper`.
This is based on design.
Should this be handled in `style/message-wrapper` or `->message`?

src/status_im/utils/config.cljs
```diff
@@ -167,4 +167,4 @@
 ;; usable enough to replace the old one **in the new UI**.
 (def new-activity-center-enabled? false)
 
-(def delete-message-for-me-undo-time-limit-ms 4000)
+(def delete-message-for-me-undo-time-limit-ms 3000)
```
yqrashawn: Based on Monday discussion

src/status_im/navigation/core.cljs
```diff
@@ -438,4 +439,6 @@
                               :community
 
                               :else
-                              :home))}))
+                              :home))
+     :dispatch-n (enc/conj-when []
+                                (and (= view-id :chat) [:chat.ui/sync-all-deleted-for-me-messages]))}))
```
yqrashawn, line 444: When leaving chat view, get all deleted-for-me messages that not yet synced with status-go and sync them

src/quo2/components/messages/system_message.cljs
```diff
@@ -140,15 +153,19 @@
...
   [:f>
    (fn []
-     (let [sv-color (ra/use-shared-value (get-color :bg :landed type))]
-       (ra/animate-shared-value-with-delay
-        sv-color (get-color :bg :default type) 0 :linear 1000)
+     (let [sv-color (ra/use-shared-value
+                     (get-color :bg (if no-landed-animation? :default :landed) type))]
+       (when-not no-landed-animation?
+         (ra/animate-shared-value-with-delay
+          sv-color (get-color :bg :default type) 0 :linear 1000))
```
yqrashawn: Only show the fading animation on delete

src/quo2/components/messages/system_message.cljs
```diff
@@ -140,15 +153,19 @@
...
        [ra/touchable-opacity
-        {:on-press #(ra/set-shared-value
-                     sv-color (get-color :bg :pressed type))
+        {:on-press #(when-not non-pressable?
+                      (ra/set-shared-value
+                       sv-color (get-color :bg :pressed type)))
```
yqrashawn, line 151: Can't have the pressed color

src/status_im/chat/models/delete_message_for_me_test.cljs
ilmotta: @yqrashawn, of course there are exceptions, but it's usually considered a better practice to not mock the code that you own and control because you end up with more brittle tests. This is especially true in Clojure, where devs often only mock functions with side effects. In this example, `rebuild-message-list` is a simple, pure function relying on what's inside the app db, so I believe you could change the test to not redefine a pure function.

There are other instances of this problem in the PR, but I'm leaving only this single comment.

src/quo2/components/messages/system_message.cljs
```diff
@@ -140,15 +153,19 @@
...
-        sv-color (get-color :bg :default type) 0 :linear 1000)
+     (let [sv-color (ra/use-shared-value
+                     (get-color :bg (if no-landed-animation? :default :landed) type))]
+       (when-not no-landed-animation?
+         (ra/animate-shared-value-with-delay
+          sv-color (get-color :bg :default type) 0 :linear 1000))
        [ra/touchable-opacity
-        {:on-press #(ra/set-shared-value
-                     sv-color (get-color :bg :pressed type))
+        {:on-press #(when-not non-pressable?
```
erikseppanen, line 149: It may be better to just name these something like `pressable?` and `animate-landing?` so it's a simpler `when something?` instead of `when-not no-something?`

src/status_im/chat/models/delete_message_for_me_test.cljs
```diff
@@ -2,114 +2,118 @@
...
+    (let [db      {:messages {cid {mid {:id mid}}}}
+          message {:message-id mid :chat-id cid}]
+      (testing "delete for me"
+        (let [expected  {:db {:messages {"chat-id" {"message-id"
```
erikseppanen: If I were writing this test, I would continue using `mid` & `cid` throughout the test (instead of "chat-id" & "message-id"), not only because it is shorter but because the values should always match.

src/status_im/chat/models/delete_message_for_me_test.cljs
```diff
@@ -2,114 +2,118 @@
...
+                           :ms       1000}]}
+              result    (delete-message-for-me/delete {:db db} message 1000)
+              timestamp (+ (datetime/timestamp) 1000)]
+          (is (= (update-in result [:db :messages "chat-id" "message-id"] dissoc :deleted-for-me-undoable-till)
+                 expected))
+          (is (-> (get-in result [:db :messages "chat-id" "message-id" :deleted-for-me-undoable-till])
+                  (- timestamp)
+                  js/Math.abs
+                  (< 10)))))
+      (testing "should return nil if message in db"
```
erikseppanen: I think you meant "should return nil if message `not` in db"

src/quo2/components/messages/system_message.cljs
```diff
@@ -55,7 +55,23 @@
...
+   [rn/view {:align-items    :center
+             :flex-direction :row}
+    [sm-icon {:icon    :main-icons/delete16
+              :color   :danger
+              :opacity 5}]
+    [text/text {:size  :paragraph-2
+                :style {:color        (get-color :text)
+                        :margin-right 5}}
+     (i18n/label :message-deleted)]
+    [sm-timestamp timestamp-str]]])
```
erikseppanen: `:deleted` is the same as `:deleted-for-me`, except for the label. Maybe they could both call a `render-deleted` fn and pass in label or keyword? It would be nice if there were a value to differentiate between "deleted" and "deleted-for-me" instead and keep using the same defmethod, but I don't see anything (earlier I was using the 'action' to differentiate between them).

src/status_im/chat/models/delete_message_for_me_test.cljs
```diff
@@ -2,114 +2,118 @@
...
+                                      {:chat-id "chat-id" :message-id "message-id"}]
+                           :ms       1000}]}
+              result    (delete-message-for-me/delete {:db db} message 1000)
+              timestamp (+ (datetime/timestamp) 1000)]
+          (is (= (update-in result [:db :messages "chat-id" "message-id"] dissoc :deleted-for-me-undoable-till)
+                 expected))
+          (is (-> (get-in result [:db :messages "chat-id" "message-id" :deleted-for-me-undoable-till])
+                  (- timestamp)
+                  js/Math.abs
+                  (< 10)))))
```
erikseppanen: If I understand this correctly, the most important part being tested here is that the key `:deleted-for-me-undoable-till` (with value timestamp + 4000) is returned. But it's being dissoc'ed in the first assert and then its value is specifically asserted in the second.

Perhaps a better way to do this would be to use `with-derefs` for `datetime/timestamp` to specify a specific timestamp for `update-db-delete-locally`, so that you can predict and test for the exact value returned.

And it might be better to use the actual `undo-time-limit-ms` value so that the test is always in sync with the code-under-test.

du82:
Does this message ever disappear?

Screenshot_2022-10-19-16-05-48-59_65ad86f86fd2321a85272a99323da10f

yqrashawn:
CleanShot 2022-10-20 at 08 43 13


 @du64 Messages won't disappear, only the toast will disappear

du82:
> CleanShot 2022-10-20 at 08 43 13
> 
> @du64 Messages won't disappear, only the toast will disappear

@yqrashawn Will all users be able to see that toast when I delete the message?

yqrashawn:
> @yqrashawn Will all users be able to see that toast when I delete the message?

@du64 No, only user who did the delete

du82:
> > @yqrashawn Will all users be able to see that toast when I delete the message?
> 
> @du64 No, only user who did the delete

Awesome! I hate how Signal tells the other person when I delete a message, they always ask what I deleted, even if its just a typo.

src/status_im/chat/models/delete_message_for_me_test.cljs
yqrashawn: Done

src/status_im/chat/models/delete_message_for_me_test.cljs
yqrashawn: Done

src/status_im/chat/models/delete_message_for_me_test.cljs
yqrashawn: Done

src/status_im/chat/models/delete_message_for_me_test.cljs
yqrashawn: Done. Thanks for your tips!

src/quo2/components/messages/system_message.cljs
yqrashawn: Done. I removed the `:deleted-for-me` type and added a `:label` option key.

src/quo2/components/messages/system_message.cljs
yqrashawn: Hi, I changed `no-landed-animation?` to `animate-landing?` and left the `non-pressable?` unchanged, since most type of system message will pressable.

ilmotta (approved):

du82:
The container overflows off the screen in dark mode

Screenshot_2022-10-20-07-39-04-36_65ad86f86fd2321a85272a99323da10f

erikseppanen (approved):

yqrashawn:
Screen Shot 2022-10-21 at 11 42 46

@du64 fixed

du82:
> Screen Shot 2022-10-21 at 11 42 46
> 
> @du64 fixed

@yqrashawn tested, it is indeed fixed. however, is the blue container supposed to disappear like the red one does in light mode?

du82:
Also, is it possible to delete the system message? If you delete a bunch of your own messages, the system ones will quickly flood your message view.

yqrashawn:
> Also, is it possible to delete the system message? If you delete a bunch of your own messages, the system ones will quickly flood your message view.

Currently not possible. Maybe we can group multiple deletes as one, which will be useful in group chat.

du82:
> > Also, is it possible to delete the system message? If you delete a bunch of your own messages, the system ones will quickly flood your message view.
> 
> Currently not possible. Maybe we can group multiple deletes as one, which will be useful in group chat.

The purpose of deleting messages is cleaning up the chat and hiding evidence that anything was ever said.
This deleted message system message directly contradicts that, I have no clue why it was even added. I'd think that "Delete for me" would remove all traces of a message.

yqrashawn:
> is the blue container supposed to disappear like the red one does in light mode?

Yes, it is. Thanks for identify that. It's the background of the chat view that needs to be changed, instead of the messages' background.

yqrashawn:
> > > Also, is it possible to delete the system message? If you delete a bunch of your own messages, the system ones will quickly flood your message view.
> > 
> > 
> > Currently not possible. Maybe we can group multiple deletes as one, which will be useful in group chat.
> 
> The purpose of deleting messages is cleaning up the chat and hiding evidence that anything was ever said. This deleted message system message directly contradicts that, I have no clue why it was even added. I'd think that "Delete for me" would remove all traces of a message.

I got what you mean. I'm also interested in the logic behind this. There's also a design for “xxx deleted this message” in group chats. Maybe it's for consistency between one-to-one chat and group chat.

yqrashawn:
I'll merge this now, https://github.com/status-im/status-mobile/pull/14168#issuecomment-1286462751 should be better implemented in another PR to change the background color including the top notch drawer.

Issue #14115, opened 2022-10-06
Implement request to leave communities flow in the new UI


J-Son89:
As a user
I want to be able to leave a community
So that I don't have to get messages from it

Description
We should implement the screens in the new UI for leaving a community.
For now, ignore token gated communities.

The initial screen is:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1355%3A111104 (that should be already implemented by @jo-mut )

Then follow the manual approval step:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1356%3A115613

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A108004

Of this screen only Leave community and Share community are implemented

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A108067

All of these are also implementable:

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401%3A115668
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1387%3A104203
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1386%3A110589

Some of the data might not be available for a community yet (Tags, number of users for example), in that case it's fine to mock the data.

PR #14169, opened 2022-10-17
feat: add leave community feature to redesign

J-Son89:
Implemented the design for leaving a community once joined:

https://github.com/status-im/status-mobile/issues/14115

Also updated the communities overview page to use the nav page component


src/quo2/components/navigation/page_nav.cljs
```diff
@@ -116,19 +129,74 @@
...
+        :icon 
+        :icon-color
+        :icon-background-color
+      }]
+   }
+  "
+
+  [{:keys [one-icon-align-left? horizontal-description? align-mid?
+           page-nav-color page-nav-background-uri
+           mid-section
```
J-Son89, line 185: I had to updated this component a bit as the icons are all needed to be clickable for navigation etc.

So I thought I would make some adjustments that I thought helped make the component easier to reason with.

Essentially bundling all mid-section, left-section and right-section props into their own map so that the prop names don't need to be prepended with mid-, left- etc and can be passed directly to their section where it's easier to read etc.

Can revert if needs be.
@flexsurfer , @ibrkhalil WDYT?

src/quo2/screens/navigation/page_nav.cljs
```diff
@@ -51,39 +51,40 @@
...
-                            :mid-section-description-color      "black"
-                            :mid-section-description-icon       :main-icons/placeholder
-                            :mid-section-description-user-icon  "https://i.picsum.photos/id/810/200/300.jpg?hmac=HgwlXd-OaLOAqhGyCiZDUb_75EgUI4u0GtS7nfgxd8s"
-                            :left-section-icon                  :main-icons/unlocked
-                            :left-section-icon-background-color (if (colors/dark?)
-                                                                  colors/neutral-80
-                                                                  colors/neutral-20)}
-        create-variation   #(merge %1 %2)
+                            :mid-section {:type                   :text-with-description
+                                          :icon                   :main-icons/placeholder
```
ilmotta: What do you think of replacing the `:main-icons/*` namespace with `:main-icons2/*` throughout this file for consistency sake?

src/status_im/ui/screens/communities/community_overview_redesign.cljs
```diff
@@ -68,17 +72,41 @@
                :render-fn                         channel/list-item}]]])
          list-of-channels)]])
 
+(def icon-color
```
ilmotta: I think the value of `icon-color` won't change if the ratom `quo.theme/theme` changes behind the scenes, so I don't see how this will work if the user changes themes. In general, things that depend on the deref'ed value of a ratom should almost always use functions.

src/quo2/components/navigation/page_nav.cljs
```diff
@@ -116,19 +129,74 @@
                     :margin-right (if push-to-the-left? 8 0))}
    [icons/icon icon (icon-props icon-color :big)]])
 
+(defn- render-right-section [right-section-buttons]
+  [rn/view {:style (assoc
+                    centrify-style
+                    :flex-direction :row
+                    :flex           1
+                    :justify-content :flex-end)}
+   (let [last-icon-index (- (count right-section-buttons) 1)]
```
ilmotta: It's more common to use `dec` for this:

```clojure
(-> right-section-buttons count dec)
```

src/quo2/components/navigation/page_nav.cljs
```diff
@@ -116,19 +129,74 @@
...
+     (map-indexed (fn [index {:keys [icon background-color icon-color on-press]}]
+                    ^{:key index}
+                    [rn/touchable-opacity {:on-press on-press}
+                     [right-section-icon
+                      {:icon             icon
+                       :background-color background-color
+                       :icon-color       icon-color
+                       :push-to-the-left? (if (= index last-icon-index) false true)}]])
```
ilmotta: It's unnecessary to explicitly return a boolean, the comparison using `=` is adequate.

src/quo2/components/navigation/page_nav.cljs
```diff
@@ -116,19 +129,74 @@
...
+                  right-section-buttons))])
+
 (defn page-nav
-  [{:keys [one-icon-align-left? horizontal-description? align-mid? page-nav-color page-nav-background-uri mid-section-type mid-section-icon mid-section-main-text mid-section-left-icon mid-section-right-icon mid-section-description mid-section-description-color mid-section-description-icon mid-section-description-user-icon mid-section-main-text-icon-color left-section-icon left-section-icon-color left-section-icon-background-color right-section-icons]}]
+  "[page-nav opts]
+   opts
+   { :one-icon-align-left? true/false
```
ilmotta: [Code style] The docstring is not properly aligned and there are map keys with line breaks in the incorrect places. I'm mentioning this because automated formatting won't detect and fix this. I think the C-style braces is also too non-idiomatic in Clojure.

[Random comment]
To be honest, I know the codebase is adding such docstrings in some components, but I'm not a fan at all.

1. They easily drift from the source code. The one in this PR is already outdated/incorrect, e.g. the `:icon-background-color` key doesn't have a a value).
2. They can't be automatically verified.
3. The types can't be reinforced.

What I think we should do is spec (with clojure.spec, Malli or another tool) and enable instrumentation during development for highly reusable components, such as the ones in quo2. Of course, this is a much bigger discussion, not in this PR's scope.

src/quo2/screens/navigation/page_nav.cljs
J-Son89: yeah sure, I can do that. I'm pretty sure we should be using main-icons2.. can't remember but I saw it some other p.r's :)

src/status_im/ui/screens/communities/community_overview_redesign.cljs
J-Son89: yes, you're right about this. I saw this bug and didn't trigger why it happened, I'll update it to a function, good spot! :)

src/quo2/components/navigation/page_nav.cljs
J-Son89: 👍

src/quo2/components/navigation/page_nav.cljs
J-Son89: yeah I guess it was done that way because the return values are backwards. I guess I can use not=

src/quo2/components/navigation/page_nav.cljs
J-Son89: Completely agree on all points here. We should bring this discussion to the team and for now I will adjust this docstring and fix the formatting on it etc.

src/status_im/ui/screens/communities/community_options_bottom_sheet.cljs
```diff
@@ -0,0 +1,90 @@
...
+                     {:icon :main-icons2/add-user
+                      :label  (i18n/label :t/invite-contacts)}
+                     {:icon :main-icons2/qr-code
+                      :label (i18n/label :t/show-qr)}
+                     {:icon :main-icons2/share-context
+                      :label (i18n/label :t/share-community)}])
+
+(defn leave-sheet [community]
+  [react/view {:style {:flex 1 :margin-left 20 :margin-right 20 :margin-bottom 20}}
+   [react/view {:style {:flex 1 :flex-direction :row :align-items :center :justify-content :space-between}}
```
OmarBasem: I think we agreed on using `rn` for consistency?

src/quo2/screens/navigation/page_nav.cljs
OmarBasem: 🙋‍♂️

Didn't know about `main-icons2` :)

src/quo2/components/navigation/page_nav.cljs
ibrkhalil: I think it's better now, I think I should've used Clojure's namespaced keywords .. But your approach is just as perfect .. Thanks Jaime 👑!

src/quo2/components/navigation/page_nav.cljs
ilmotta: Thanks @J-Son89 :)

ilmotta (approved):

src/quo2/components/navigation/page_nav.cljs
ibrkhalil: 🆗

src/status_im/ui/screens/communities/community_options_bottom_sheet.cljs
J-Son89: yep, you're right, I'll update :)

ibrkhalil (approved):
lgtm

du82:
The bottom sheet icons are smaller than they are in the messages action sheet

Screenshot_2022-10-18-21-42-05-60_65ad86f86fd2321a85272a99323da10f
Screenshot_2022-10-18-21-40-36-58_65ad86f86fd2321a85272a99323da10f

src/status_im/ui/screens/communities/community_overview_redesign.cljs
```diff
@@ -68,17 +70,41 @@
...
      :border-radius   20}}
    [react/view (styles/community-cover-container 148)
+
     [react/image
      {:source      cover
-      :style  {:position :relative
-               :flex 1}}]]
+      :style  {:position :absolute
+               :flex 1}}]
+    [react/view {:style {:margin-top 26}}
```
OmarBasem: `rn/view`

There are a few `react/view` in the updated files.

OmarBasem (approved):

src/status_im/ui/screens/communities/community_overview_redesign.cljs
J-Son89: will adjust :)

churik:
@J-Son89 is it the flow, right? 
If so, I'd test it manually first. Thanks!

PR #14170, opened 2022-10-17
Record audio button component

### LINKED ISSUE #13515: UI Component - Record Audio
This issue is to track the creation of the component: **Record Audio**

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=3289%3A63901.

briansztamfater:
fixes #13515 

### Summary

This PR implements Record audio Button component (part 1) which implements the recording button with its animations and interactions / gestures.

Will submit a follow up PR with the part 2 implementing the complete record audio flow, tracking it in #14558.

https://user-images.githubusercontent.com/18485527/208020613-b99c7f11-1f23-4741-80d9-53d03cbe233e.mp4

https://user-images.githubusercontent.com/18485527/208023942-b43f0d94-3674-4c4a-ae08-3ad759882641.mp4

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats
- public chats
- group chats

### Steps to test

- Open Status
- Go to Settings > Quo2.0 preview > record-audio
- Interact with the component

status: ready

### DIFF
--- a/src/js/record_audio_worklets.js
+++ b/src/js/record_audio_worklets.js
@@ -1,0 +1,18 @@
+import { useDerivedValue } from 'react-native-reanimated';
+
+const MAX_SCALE = 1.8;
+
+export function ringScale(scale, subtract) {
+  return useDerivedValue(
+    function () {
+      'worklet'
+      const value = scale.value;
+      const maxDelta = MAX_SCALE - 1;
+      const maxDeltaDiff = 1 - maxDelta;
+      const maxVirtualScale = MAX_SCALE + maxDelta;
+      const decimals = value - Math.floor(value);
+      const normalizedValue = value >= maxVirtualScale ? (decimals + ((parseInt(value) - 1) * maxDeltaDiff) + 1) : value;
+      return (((normalizedValue - subtract) > MAX_SCALE ? normalizedValue - maxDelta : normalizedValue) - subtract);
+    }
+  );
+}
--- a/src/quo2/components/record_audio/record_audio/style.cljs
+++ b/src/quo2/components/record_audio/record_audio/style.cljs
@@ -1,0 +1,212 @@
+(ns quo2.components.record-audio.record-audio.style
+  (:require [quo2.foundations.colors :as colors]
+            [react-native.reanimated :as reanimated]))
+
+(defn animated-circle [scale opacity color]
+  (reanimated/apply-animations-to-style
+   {:transform [{:scale scale}]
+    :opacity   opacity}
+   {:width           56
+    :height          56
+    :border-width    1
+    :border-color    color
+    :border-radius   28
+    :position        :absolute
+    :justify-content :center
+    :align-items     :center
+    :z-index         0}))
+
+(defn record-button-big-container [translate-x translate-y opacity]
+  (reanimated/apply-animations-to-style
+   {:transform [{:translateY translate-y}
+                {:translateX translate-x}]
+    :opacity    opacity}
+   {:position        :absolute
+    :bottom          0
+    :right           0
+    :width           96
+    :height          96
+    :align-items     :center
+    :justify-content :center
+    :z-index         0}))
+
+(defn record-button-big-body [button-color]
+  {:width            56
+   :height           56
+   :border-radius    28
+   :justify-content  :center
+   :align-items      :center
+   :background-color button-color
+   :overflow         :hidden})
+
+(defn record-button-big-red-overlay [red-overlay-opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity red-overlay-opacity}
+   {:position         :absolute
+    :top              0
+    :left             0
+    :right            0
+    :bottom           0
+    :background-color colors/danger-50}))
+
+(defn record-button-big-gray-overlay [gray-overlay-opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity gray-overlay-opacity}
+   {:position         :absolute
+    :top              0
+    :left             0
+    :right            0
+    :bottom           0
+    :background-color (colors/theme-colors colors/neutral-80-opa-5-opaque colors/neutral-80)}))
+
+(defn record-button-big-icon-container [icon-opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity icon-opacity}
+   {}))
+
+(def stop-icon
+  {:width            13
+   :height           13
+   :border-radius    4
+   :background-color colors/white})
+
+(defn send-button-container [opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity opacity}
+   {:justify-content :center
+    :align-items     :center
+    :position        :absolute
+    :width           56
+    :height          56
+    :top             0
+    :right           20}))
+
+(defn send-button-connector [opacity width height border-radius-first-half border-radius-second-half]
+  (reanimated/apply-animations-to-style
+   {:opacity                    opacity
+    :width                      width
+    :height                     height
+    :border-bottom-left-radius  border-radius-second-half
+    :border-top-left-radius     border-radius-first-half
+    :border-top-right-radius    border-radius-first-half
+    :border-bottom-right-radius border-radius-second-half}
+   {:justify-content  :center
+    :align-items      :center
+    :align-self       :center
+    :background-color colors/primary-50
+    :z-index          0}))
+
+(defn send-button [translate-y opacity]
+  (reanimated/apply-animations-to-style
+   {:transform [{:translateY translate-y}]
+    :opacity   opacity}
+   {:justify-content  :center
+    :align-items      :center
+    :background-color colors/primary-50
+    :width            32
+    :height           32
+    :border-radius    16
+    :position         :absolute
+    :top              0
+    :right            32
+    :z-index          10}))
+
+(def send-icon-container
+  {:z-index 10})
+
+(defn lock-button-container [opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity opacity}
+   {:transform       [{:rotate "45deg"}]
+    :justify-content :center
+    :align-items     :center
+    :position        :absolute
+    :width           56
+    :height          56
+    :top             20
+    :left            20}))
+
+(defn lock-button-connector [opacity width height border-radius-first-half border-radius-second-half]
+  (reanimated/apply-animations-to-style
+   {:opacity                    opacity
+    :width                      width
+    :height                     height
+    :border-bottom-left-radius  border-radius-first-half
+    :border-top-left-radius     border-radius-first-half
+    :border-top-right-radius    border-radius-second-half
+    :border-bottom-right-radius border-radius-second-half}
+   {:justify-content  :center
+    :align-items      :center
+    :align-self       :center
+    :background-color (colors/theme-colors colors/neutral-80-opa-5-opaque colors/neutral-80)
+    :overflow         :hidden}))
+
+(defn lock-button [translate-x-y opacity]
+  (reanimated/apply-animations-to-style
+   {:transform [{:translateX translate-x-y}
+                {:translateY translate-x-y}]
+    :opacity   opacity}
+   {:width            32
+    :height           32
+    :justify-content  :center
+    :align-items      :center
+    :background-color (colors/theme-colors colors/neutral-80-opa-5-opaque colors/neutral-80)
+    :border-radius    16
+    :position         :absolute
+    :top              24
+    :left             24
+    :overflow         :hidden
+    :z-index          12}))
+
+(defn delete-button-container [opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity opacity}
+   {:justify-content :center
+    :align-items     :center
+    :position        :absolute
+    :width           56
+    :height          56
+    :bottom          20
+    :left            0}))
+
+(defn delete-button-connector [opacity width height border-radius-first-half border-radius-second-half]
+  (reanimated/apply-animations-to-style
+   {:opacity                    opacity
+    :width                      width
+    :height                     height
+    :border-bottom-left-radius  border-radius-first-half
+    :border-top-left-radius     border-radius-first-half
+    :border-top-right-radius    border-radius-second-half
+    :border-bottom-right-radius border-radius-second-half}
+   {:justify-content  :center
+    :align-items      :center
+    :align-self       :center
+    :background-color colors/danger-50
+    :z-index          0}))
+
+(defn delete-button [translate-x opacity]
+  (reanimated/apply-animations-to-style
+   {:transform [{:translateX translate-x}]
+    :opacity   opacity}
+   {:width            32
+    :height           32
+    :justify-content  :center
+    :align-items      :center
+    :background-color colors/danger-50
+    :border-radius    16
+    :position         :absolute
+    :top              76
+    :left             0
+    :z-index          11}))
+
+(defn record-button-container [opacity]
+  (reanimated/apply-animations-to-style
+   {:opacity       opacity}
+   {:margin-bottom 32
+    :margin-right  32}))
+
+(def input-container
+  {:width           140
+   :height          140
+   :align-items     :flex-end
+   :justify-content :flex-end})
--- a/src/quo2/components/record_audio/record_audio/view.cljs
+++ b/src/quo2/components/record_audio/record_audio/view.cljs
@@ -1,0 +1,553 @@
+(ns quo2.components.record-audio.record-audio.view
+  (:require [quo.react :refer [memo effect!]]
+            [react-native.core :as rn]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.icon :as icons]
+            [quo2.components.buttons.button :as button]
+            [react-native.reanimated :as reanimated]
+            [reagent.core :as reagent]
+            [cljs-bean.core :as bean]
+            [quo2.components.record-audio.record-audio.style :as style]
+            [status-im.utils.utils :as utils]
+            [oops.core :as  oops]))
+
+(def ^:private scale-to-each 1.8)
+(def ^:private scale-to-total 2.6)
+(def ^:private scale-padding 0.16)
+(def ^:private opacity-from-lock 1)
+(def ^:private opacity-from-default 0.5)
+(def ^:private signal-anim-duration 3900)
+(def ^:private signal-anim-duration-2 1950)
+
+(def ^:private animated-ring
+  (reagent/adapt-react-class
+   (memo
+    (fn [props]
+      (let [{:keys [scale opacity color]} (bean/bean props)]
+        (reagent/as-element
+         [reanimated/view {:style (style/animated-circle scale opacity color)}]))))))
+
+(def ^:private record-button-area-big
+  {:width  56
+   :height 56
+   :x      64
+   :y      64})
+
+(def ^:private record-button-area
+  {:width  48
+   :height 48
+   :x      68
+   :y      68})
+
+(defn- delete-button-area [active?]
+  {:width  (if active? 72 82)
+   :height 56
+   :x      (if active? -16 -32)
+   :y      (if active? 64 70)})
+
+(defn- lock-button-area [active?]
+  {:width  (if active? 72 100)
+   :height (if active? 72 102)
+   :x      -32
+   :y      -32})
+
+(defn- send-button-area [active?]
+  {:width  56
+   :height (if active? 72 92)
+   :x      68
+   :y      (if active? -16 -32)})
+
+(defn- touch-inside-area? [{:keys [location-x location-y ignore-min-y? ignore-max-y? ignore-min-x? ignore-max-x?]}
+                           {:keys [width height x y]}]
+  (let [max-x (+ x width)
+        max-y (+ y height)]
+    (and
+     (and
+      (or ignore-min-x? (>= location-x x))
+      (or ignore-max-x? (<= location-x max-x)))
+     (and
+      (or ignore-min-y? (>= location-y y))
+      (or ignore-max-y? (<= location-y max-y))))))
+
+(def record-audio-worklets (js/require "../src/js/record_audio_worklets.js"))
+
+(defn- ring-scale [scale substract]
+  (.ringScale ^js record-audio-worklets
+              scale
+              substract))
+
+(defn- record-button-big [recording? ready-to-send? ready-to-lock? ready-to-delete? record-button-is-animating? record-button-at-initial-position? locked? reviewing-audio? recording-timer recording-length-ms clear-timeout touch-active?]
+  [:f>
+   (fn []
+     (let [scale                (reanimated/use-shared-value 1)
+           opacity              (reanimated/use-shared-value 0)
+           opacity-from         (if @ready-to-lock? opacity-from-lock opacity-from-default)
+           animations           (map
+                                 (fn [index]
+                                   (let [ring-scale (ring-scale scale (* scale-padding index))]
+                                     {:scale ring-scale
+                                      :opacity (reanimated/interpolate ring-scale [1 scale-to-each] [opacity-from 0])}))
+                                 (range 0 5))
+           rings-color          (cond
+                                  @ready-to-lock? (colors/theme-colors colors/neutral-80-opa-5-opaque colors/neutral-80)
+                                  @ready-to-delete? colors/danger-50
+                                  :else colors/primary-50)
+           translate-y          (reanimated/use-shared-value 0)
+           translate-x          (reanimated/use-shared-value 0)
+           button-color         colors/primary-50
+           icon-color           (if (and (not (colors/dark?)) @ready-to-lock?) colors/black colors/white)
+           icon-opacity         (reanimated/use-shared-value 1)
+           red-overlay-opacity  (reanimated/use-shared-value 0)
+           gray-overlay-opacity (reanimated/use-shared-value 0)
+           complete-animation   (fn []
+                                  (cond
+                                    (and @ready-to-lock? (not @record-button-is-animating?))
+                                    (do
+                                      (reset! locked? true)
+                                      (reset! ready-to-lock? false))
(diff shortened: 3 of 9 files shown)


src/quo2/reanimated.cljs
```diff
@@ -68,3 +73,11 @@
 (defn animate-shared-value-with-delay [anim val duration easing delay]
   (set-shared-value anim (with-delay delay (with-timing val (js-obj "duration" duration
                                                                     "easing"   (get easings easing))))))
+
+(defn animate-shared-value-with-delay-repeat [anim val duration easing delay repeat-times]
```
J-Son89: consider using full names for vars here.
e.g val -> value 
anim -> animation (?)

Also consider using an object for named params here, in my opinion for functions with more than 2 or 3 parameters it's better to use named params.
e.g
`
(defn animate-shared-value-with-delay-repeat [
{:keys :[anim val duration easing delay repeat-times]}
...`
or at least for some of the parameters

src/quo2/reanimated.cljs
```diff
@@ -68,3 +73,11 @@
...
+  (set-shared-value anim (with-delay delay (with-repeat (with-timing val (js-obj "duration" duration
+                                                                                 "easing"   (get easings easing))) repeat-times false))))
```
J-Son89: what does false do here? perhaps we can put it in a `def` so we know it's function? 🤔

src/quo2/reanimated.cljs
briansztamfater: Sure, I tried to follow same convention as other functions, but I agree with your suggestion. I will make those changes for this function but we would need to refactor the other ones in other PR so we don't make this one bigger (still more changes to come).

src/quo2/reanimated.cljs
briansztamfater: It indicates wether if the animation should reverse after finish. We can better make this parameter optional and set `false` as the default value. What do you think?

src/quo2/foundations/colors.cljs
```diff
@@ -91,6 +109,11 @@
 (def white-opa-90 (alpha white 0.9))
 (def white-opa-95 (alpha white 0.95))
 
+;;;;Black
+
+;;Solid
+(def black "#000000")
```
briansztamfater, line 117: Designs uses plain black color for the lock / unlock icon when in light mode

src/quo2/foundations/colors.cljs
```diff
@@ -1,17 +1,31 @@
...
+      (let [hex (replace value #"#" "")
+            r   (- 255 (* opacity (- 255 (js/parseInt (subs hex 0 2) 16))))
+            g   (- 255 (* opacity (- 255 (js/parseInt (subs hex 2 4) 16))))
+            b   (- 255 (* opacity (- 255 (js/parseInt (subs hex 4 6) 16))))]
+        (str "rgb(" r "," g "," b ")"))
+      (let [rgb (split value #",")
+            r   (- 255 (* opacity (- 255 (get rgb 0))))
+            g   (- 255 (* opacity (- 255 (get rgb 1))))
+            b   (- 255 (* opacity (- 255 (get rgb 2))))]
+        (str "rgb(" r "," g "," b ")")))))
```
briansztamfater: This function returns an opaque hex color with opacity as if it was shown on a white background, needed for overlapping colors with opacity on top of other views

src/js/record_audio_worklets.js
```diff
@@ -0,0 +1,20 @@
+import { useDerivedValue } from 'react-native-reanimated';
+
+// Record Audio Worklets
```
J-Son89: this comment is the name of the file so I think it probably can be removed.

src/js/record_audio_worklets.js
```diff
@@ -0,0 +1,20 @@
...
+
+const MAX_SCALE = 1.8;
+
+export function ringScale(scale, substract) {
```
J-Son89: should we create a some unit tests for this code? it could be done in a follow up

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+(def ^:private opacity-from-lock 1)
+(def ^:private opacity-from-default 0.5)
+(def ^:private signal-anim-duration 3900)
+(def ^:private signal-anim-duration-2 1950)
+
+(def ^:private animated-ring
+  (reagent/adapt-react-class
+   (memo
+    (fn [props]
+      (let [{:keys [scale opacity color]} (bean/bean props)]
```
J-Son89, line 26: do we need this bean library here? can it be done without it? Are we still intending to use the library? 🤔

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                                   (reset! ready-to-send? false)
+                                   (reset! ready-to-delete? false)
+                                   (reset! ready-to-lock? false)
+                                   (utils/clear-interval @recording-timer)
+                                   (reset! recording-length-ms 0)))
+           start-animation (fn []
+                             (reanimated/set-shared-value opacity 1)
+                             (reanimated/animate-shared-value-with-timing scale 2.6 signal-anim-duration :linear)
+                             ;; TODO: Research if we can implement this with withSequence method from Reanimated 2
+                             ;; GitHub issue [#14561]: https://github.com/status-im/status-mobile/issues/14561
```
J-Son89: something we should discuss further as a team but maybe we don't need TODO's at all and the issue is enough. Keeps the codebase cleaner. 
@ilmotta you mentioned this before

src/js/record_audio_worklets.js
ilmotta: We already have Jest available right? I think the test should come along with the PR.

J-Son89 (approved):
looks good to me, maybe you could add some tests to this as a follow up? I will provide some docs and demos on component tests soon and can help you write them if needed 👍

src/quo2/components/record_audio/record_audio/style.cljs
```diff
@@ -0,0 +1,212 @@
...
+    :border-color    color
+    :border-radius   28
+    :position        :absolute
+    :justify-content :center
+    :align-items     :center
+    :z-index         0}))
+
+(defn record-button-big-container [translate-x translate-y opacity]
+  (reanimated/apply-animations-to-style
+   {:transform [{:translateY translate-y}
```
ilmotta, line 21: Have you tried using kebab case in the `translate-*` keys? There are other camel cased keywords in this PR.

src/js/record_audio_worklets.js
J-Son89: yeah we have jest installed, it's currently only running on clojurescript files so it needs a quick config update to also run on the js files. 
I can set this up and work with Brian on it. I agree going forward the test should come with it but since it's not available just yet perhaps we should not block this p.r from going in,

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: One thing that I like about TODO's is that we can find and keep track of all the tech debt easily by searching TODO keywords in the project. But yeah we can definitely discuss it and reach consensus on how to handle them, I can adapt myself.

src/js/record_audio_worklets.js
briansztamfater: Sure thing, once we have that set up would be great to add unit tests for it, let me know and I can add it on a follow up PR this time :)

src/quo2/components/record_audio/record_audio/view.cljs
J-Son89: Okay do as you feel best for now and let's have a vote on how we want to approach this going forward. :) cc @OmarBasem @smohamedjavid As I commented something similar to you in other prs

src/quo2/foundations/colors.cljs
ilmotta: @briansztamfater, it'd be good if this function was private because it only has a single usage inside its own namespace.

But, if you want to keep it public, other devs could end up using it in places where it's expensive, so I think it should be memoized.

src/react_native/reanimated.cljs
```diff
@@ -75,4 +82,11 @@
...
   (set-shared-value anim (with-repeat (with-timing val (js-obj "duration" duration
-                                                               "easing"   (get easings easing))) number-of-repetitions reverse?)))
\ No newline at end of file
+                                                               "easing"   (get easings easing))) number-of-repetitions reverse?)))
+
+(defn animate-shared-value-with-delay-repeat
+  ([anim val duration easing delay number-of-repetitions]
+   (animate-shared-value-with-delay-repeat anim val duration easing delay number-of-repetitions false))
+  ([anim val duration easing delay number-of-repetitions reverse?]
+   (set-shared-value anim (with-delay delay (with-repeat (with-timing val (js-obj "duration" duration
```
ilmotta: Could you reformat this piece differently? I find it really hard to read. It's common for functions or macros that start with `with-` to have the second argument in the next line. If you follow this idea you end up with something more readable in this particular code.

And how about using the `#js` literal instead of `js-obj`?

```clojure
(defn animate-shared-value-with-delay-repeat
  ([anim val duration easing delay number-of-repetitions]
   (animate-shared-value-with-delay-repeat anim val duration easing delay number-of-repetitions false))
  ([anim val duration easing delay number-of-repetitions reverse?]
   (set-shared-value anim
                     (with-delay delay
                       (with-repeat
                         (with-timing val
                           #js {:duration duration
                                :easing   (get easings easing)})
                         number-of-repetitions reverse?)))))
```

src/js/record_audio_worklets.js
ilmotta: Makes sense guys, if it's not yet set up it's fine to not to test now.

src/js/record_audio_worklets.js
```diff
@@ -0,0 +1,20 @@
+import { useDerivedValue } from 'react-native-reanimated';
```
ilmotta, line 1: @briansztamfater, I couldn't find any calls to the `ringScale` worklet. Where is it being used?

src/quo2/components/record_audio/record_audio/view.cljs
ilmotta: I have mixed feelings, because what usually happens over time is that we tend to produce more TODO comments than actually fixing them. They are often written in a poor way too (not the case in this PR). Eventually, they become less and less useful because of the large amount of TODO comments.

src/quo2/foundations/colors.cljs
```diff
@@ -1,17 +1,31 @@
 (ns quo2.foundations.colors
-  (:require [clojure.string :as string]
-            [quo2.theme :as theme]))
+  (:require [clojure.string :as string :refer [join replace split starts-with?]]
+            [quo.theme :as theme]))
```
flexsurfer: could you please rollback require? better to not use refer, what's the point here ? and also why quo2 theme was replaced by quo ?

flexsurfer (changes requested):

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+         [icons/icon :i/delete {:color colors/white
+                                :size  20}]]]))])
+
+(defn- record-button []
+  [:f>
+   (fn []
+     (let [opacity (reanimated/use-shared-value 1)
+           show-animation #(reanimated/set-shared-value opacity 1)
+           hide-animation #(reanimated/set-shared-value opacity 0)]
+       (effect! #(if (or @recording? @reviewing-audio?) (hide-animation) (show-animation)) [@recording? @reviewing-audio?])
```
ilmotta: This line surely deserves some line breaks ;)

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+(defn input-view []
+  [:f>
+   (fn []
+     (fn []
+       [rn/view {:style                         style/input-container
+                 :pointer-events                :box-only
+                 :on-start-should-set-responder (fn [^js e]
+                                                  (when-not @locked?
+                                                    (let [pressed-record-button? (touch-inside-area?
+                                                                                  {:locationX (-> e .-nativeEvent.locationX)
```
ilmotta: You can use kebab case for all these keywords. There are other places in this namespace that should be changed.

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                                                        (reset! recording? pressed-record-button?)
+                                                        (when pressed-record-button?
+                                                          ;; TODO: By now we just track recording length, we need to add actual audio recording logic
+                                                          ;; GitHub issue [#14558]: https://github.com/status-im/status-mobile/issues/14558
+                                                          (reset! recording-timer (utils/set-interval #(reset! recording-length-ms (+ @recording-length-ms 500)) 500))))
+                                                      (reset! touch-active? true)))
+                                                  true)
+                 :on-responder-move             (fn [^js e]
+                                                  (when-not @locked?
+                                                    (let [locationX               (-> e .-nativeEvent.locationX)
```
ilmotta: [Guidelines] https://github.com/status-im/status-mobile/blob/develop/doc/new-guidelines.md#javascript-interop

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                                                          locationY               (-> e .-nativeEvent.locationY)
+                                                          pageX                   (-> e .-nativeEvent.pageX)
+                                                          pageY                   (-> e .-nativeEvent.pageY)
+                                                          moved-to-send-button?   (touch-inside-area?
+                                                                                   {:locationX locationX
+                                                                                    :locationY locationY}
+                                                                                   (send-button-area @ready-to-send?)
+                                                                                   true false false true)
```
ilmotta: Why the mix of positional arguments with maps?

Also, passing multiple booleans to functions makes the code hard to read. You can change the `touch-inside-area?` signature, so that the caller code becomes much more readable. Notice that I already changed the example to follow our latest guidelines:

```clojure
(touch-inside-area?
 {:location-x    (oops/oget e "nativeEvent.locationX") 
  :location-y    (oops/oget e "nativeEvent.locationY")
  :ignore-min-y? (if @reviewing-audio? record-button-area record-button-area-big)
  :ignore-max-y? false
  :ignore-min-x? false
  :ignore-min-x? false
  :ignore-max-x? false})
```

src/quo2/foundations/colors.cljs
briansztamfater: Thanks for finding this out, not sure why it quo2 theme require changed actually. Fixed it and rollbacked :)

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+            [quo2.foundations.colors :as colors]
+            [quo2.components.icon :as icons]
+            [quo2.components.buttons.button :as button]
+            [react-native.reanimated :as reanimated]
+            [reagent.core :as reagent]
+            [cljs-bean.core :as bean]
+            [quo2.components.record-audio.record-audio.style :as style]
+            [status-im.utils.utils :as utils]))
+
+(defonce recording? (reagent/atom false))
```
ilmotta: This is a lot of global state @briansztamfater, it makes reasoning about the code even harder because any function can access anything. Have you considered using local component state?

IMO we should almost never use this kind of approach to manage state. Even in pure React land we don't see this, because hooks are always used inside components and follow the normal rendering lifecycle.

Notice that by using global state we're limiting pretty much all components in this namespace to be only used once per screen. It also makes testing much more fragile (in case we want to write component tests in the future), because every test will need to reset lots of global state, otherwise we get flaky tests.

On top of that it makes using the REPL less enjoyable in my experience, because now you need to resort to different tricks to reset the state, redefine `defonce`, etc.

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: I don't have a strong reason actually, as I was first testing this PR with UX designers probably I made changes not thinking too much on code readability, but on functionaliy :)

But now is time to improve the former, so thanks for the suggestions, will make those changes.

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                             ;; GitHub issue [#14561]: https://github.com/status-im/status-mobile/issues/14561
+                             (reset! clear-timeout (utils/set-timeout #(do (reanimated/set-shared-value scale scale-to-each)
+                                                                           (reanimated/animate-shared-value-with-delay-repeat scale scale-to-total signal-anim-duration-2 :linear 0 -1))
+                                                                      signal-anim-duration)))
+           stop-animation #(do
+                             (reanimated/set-shared-value opacity 0)
+                             (reanimated/cancel-animation scale)
+                             (reanimated/set-shared-value scale 1)
+                             (when @clear-timeout (utils/clear-timeout @clear-timeout)))
+           start-y-animation #(do
```
ilmotta: The Clojure Style Guide recommends we avoid the literal `#()` for long functions. It forces the reader to read every single line to see if any argument (`%`) is used. Use `(fn [] ...)` instead.

FYI: there are many instances of this style problem in this namespace.

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+           fade-out-animation #(do
+                                 (reanimated/animate-shared-value-with-timing translate-x-y 20 200 :linear)
+                                 (reanimated/animate-shared-value-with-timing opacity 0 200 :linear)
+                                 (reanimated/set-shared-value connector-opacity 0)
+                                 (reanimated/set-shared-value width 24)
+                                 (reanimated/set-shared-value height 12)
+                                 (reanimated/set-shared-value border-radius-first-half 8)
+                                 (reanimated/set-shared-value border-radius-second-half 16))]
+       (effect! #(if @recording? (fade-in-animation) (fade-out-animation)) [@recording?])
+       (effect! #(cond @ready-to-lock? (start-x-y-animation) (and @recording? (not @locked?)) (reset-x-y-animation)) [@ready-to-lock?])
```
ilmotta: Could you please reformat this line? This is a bit beyond my comprehension.

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Agree as well

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                                                      (and @reviewing-audio? on-record-button?)
+                                                      (reset! reviewing-audio? false)
+                                                      (and @ready-to-lock? (not @record-button-is-animating?))
+                                                      (do
+                                                        (reset! locked? true)
+                                                        (reset! ready-to-lock? false))
+                                                      (and (not @reviewing-audio?) on-record-button?)
+                                                      (do
+                                                        (when (>= @recording-length-ms 500) (reset! reviewing-audio? true))
+                                                        (reset! locked? false)
```
ilmotta: I'm obviously not suggesting you to refactor the entire implementation, but it sure looks like managing all this state is fragile and difficult to get correct. It will be a nightmare for anyone having to change/evolve this namespace.

It would have been quite nice to see the separation of the state management from the actual view, something like a basic state machine could've been fruitful too. Obviously re-frame is out of question for this kind of performance requirements.

src/quo2/components/record_audio/record_audio/style.cljs
briansztamfater: Yeah, unfortunately it crashes when using `translate-*` keys for transform styling props

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,473 @@
...
+                                                                                   false false false false)
+                                                          moved-to-record-button? (and
+                                                                                   (touch-inside-area?
+                                                                                    {:locationX locationX
+                                                                                     :locationY locationY}
+                                                                                    record-button-area-big
+                                                                                    false false false false)
+                                                                                   (not= locationX pageX)
+                                                                                   (not= locationY pageY))]
+                                                      (cond
```
ilmotta: In Clojure it's highly praised to break long functions into smaller ones. This often leads to more readable code and it's especially important for using the REPL. I'm not sure if you've tried and you didn't like, but you could consider extracting many of these anonymous functions, like `on-responder-move`, `on-responder-release` and so on.

src/quo2/components/record_audio/record_audio/view.cljs
ilmotta: Totally understandable, thanks @briansztamfater

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Moved them to separate functions

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Yeah, I know, the component is internally complex. It has so many "invisible" conditions based on touch starts, touch movements, touch releases, animations completion, and also different states. I tried the idea of a basic state machine, but had some problems with useEffect that is needed to trigger some animations (currently reacting to different specific ratom updates on the different internal components). Also I think using state machine would also make useEffect execute to every state update and would make those functions even more complex, so decided to keep it as is at the time of developing.

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Thanks for pointing this out. I've tried both ways through development iteration, didn't have clear all the pros & cons of both approaches. I moved ratoms / atoms to the component itself.

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done, I separated it in two maps which I think make more sense:
- The user touch data & params
- The touch area to compare with

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done

src/react_native/reanimated.cljs
briansztamfater: Done. Probably other functions in this file need to be reformatted too, as I just copied how the other ones were implemented. Probably will raise a PR with those refactors after this one.

src/quo2/components/record_audio/record_audio/view.cljs
briansztamfater: Done

src/quo2/foundations/colors.cljs
briansztamfater: Done

src/js/record_audio_worklets.js
briansztamfater: Done

src/js/record_audio_worklets.js
```diff
@@ -0,0 +1,18 @@
...
+  return useDerivedValue(
+    function () {
+      'worklet'
+      const value = scale.value;
+      const maxDelta = MAX_SCALE - 1;
+      const maxDeltaDiff = 1 - maxDelta;
+      const maxVirtualScale = MAX_SCALE + maxDelta;
+      const decimals = value - Math.floor(value);
+      const normalizedValue = value >= maxVirtualScale ? (decimals + ((parseInt(value) - 1) * maxDeltaDiff) + 1) : value;
+      return (((normalizedValue - substract) > MAX_SCALE ? normalizedValue - maxDelta : normalizedValue) - substract);
```
ibrkhalil: Typo?
subtstract -> subtract ?

ibrkhalil (approved):
🚀

src/quo2/components/record_audio/record_audio/view.cljs
ilmotta: Understandable, it seems there's more to this problem than meets the eye.

src/quo2/components/record_audio/record_audio/view.cljs
```diff
@@ -0,0 +1,530 @@
...
+                                  (reanimated/animate-shared-value-with-timing icon-opacity 1 500 :linear)
+                                  (reanimated/animate-shared-value-with-timing gray-overlay-opacity 0 800 :linear)
+                                  (utils/set-timeout (fn [] (reset! record-button-at-initial-position? true)) 800))]
+       (effect! #(cond
+                   @recording?
+                   (start-animation)
+                   (not @ready-to-lock?)
+                   (stop-animation))
+                [@recording?])
+       (effect! #(if @ready-to-lock? (start-x-y-animation) (reset-x-y-animation)) [@ready-to-lock?])
```
ilmotta: These one-liners are hard to read, I had to add line breaks in my text editor to understand the last argument was just the dependency vector of `effect!`.

The following is longer, but easier to parse IMO:

```clojure
(effect! (fn []
           (if @ready-to-lock?
             (start-x-y-animation)
             (reset-x-y-animation)))
         [@ready-to-lock?])
```

ilmotta (approved):

flexsurfer (approved):

PR #14179, opened 2022-10-18
Use native binary for the clj-kondo linter

ilmotta:
### Summary

The command `make lint` never worked in my machines with NixOS because it failed in the step calling `yarn clj-kondo ...`. This PR simply replaces the non-native dependency with the native one, which supposedly solves the problem of _it doesn't work on my machine_.

### Review notes

The new linter version caught one or two additional warnings (which is good), and that's why you'll see some application code changes.

### Testing notes

Please, run `make lint` and see if it works for you.

status: ready

### DIFF
--- a/nix/shells.nix
+++ b/nix/shells.nix
@@ -23,7 +23,7 @@ let
 
     # for calling clojure targets in CI or Makefile
     clojure = mkShell {
-      buildInputs = with pkgs; [ clojure flock maven openjdk ];
+      buildInputs = with pkgs; [ clojure flock maven openjdk clj-kondo];
       inputsFrom = [ nodejs-sh ];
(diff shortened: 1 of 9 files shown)


nix/shells.nix
```diff
@@ -23,7 +23,7 @@ let
 
     # for calling clojure targets in CI or Makefile
     clojure = mkShell {
-      buildInputs = with pkgs; [ clojure flock maven openjdk ];
+      buildInputs = with pkgs; [ clojure flock maven openjdk clj-kondo];
```
jakubgs: Lost a space.

jakubgs (approved):
Looks correct to me. The only question would be, do we want to drop the node dependency in case someone wants to use it without using nix?

I would prefer to drop it, but it is worth asking.

PR #14182, opened 2022-10-18
Scrollable tags 

### LINKED ISSUE #14181: scrollable tags component that fades out right end
The current implementation of tags component needs to be wrapped in a scroll-view for the user to be able to scroll an horizontal list of tags. However the designs as is in the figma link show that tags components fades out which currently is not implemented. 

figma: https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=5488%3A224643

We dont want to do away with the current implementation. What we want is to address the case where tags are scrollable and have the fading effect

jo-mut:
fixes #14181

### Summary
Based on the implementation of scrollable-tabs  this PR simply add the same functionality to the tags component. This achieves two things;

1. Tags can be scrolled without having to wrap the tags component with a parent scrollView
2. Tags achieve the fading out effect as is on the figma designs

### Testing notes
Part of this PR has been scoped out. The relevance of testing for this PR is just to confirm that it does not introduce any regressions especially on activity center where scrollable-tabs have been used other should be fine.

community filter-tags have been scopped out of mvp
activity center tabs have will be part of mvp and **`scrollable-tabs`** makes use of **`scrollable-view`** in this PR


src/quo2/components/tags/tags.cljs
```diff
@@ -26,4 +32,121 @@
                                       resource)
                      :on-press      #(do (reset! active-tab-id %)
                                          (when on-change (on-change %)))})
-             tag-label]])]))))
\ No newline at end of file
+             tag-label]])]))))
+
+(defn- calculate-fade-end-percentage
```
flexsurfer: is it a copy from tabs? might be worth just moving it and use in both places instead of copying it?

src/quo2/components/tags/tag.cljs
```diff
@@ -59,22 +59,36 @@
...
-                          :border-color        (if blurred
-                                                 blurred-border-color
-                                                 border-color)
-                          :on-press            on-press
-                          :accessibility-label accessibility-label
-                          :disabled            disabled
-                          :type                type
-                          :label               label}
-       [tag-resources size type resource icon-color label text-color labelled]])))
+      [rn/view {:align-items  :center}
```
ibrkhalil: Add `:style` attribute please

src/quo2/components/tags/tag.cljs
```diff
@@ -59,22 +59,36 @@
       label])])
 
 (defn tag
+  "opts
+   {:type         :icon/:emoji/:label
+    :size         32/24
+    :on-press     fn
+    :blurred      true/false 
+    :resource     icon/image
+    :labelled     true/false
```
ibrkhalil: Can we indicate it's a boolean by adding a question mark at the end
`labelled` -> `labelled?`

src/quo2/components/tags/tag.cljs
```diff
@@ -59,22 +59,36 @@
       label])])
 
 (defn tag
+  "opts
+   {:type         :icon/:emoji/:label
+    :size         32/24
+    :on-press     fn
+    :blurred      true/false
```
ibrkhalil: Can we indicate it's a boolean by adding a question mark at the end
`blurred` -> `blurred?`

src/quo2/components/tags/tag.cljs
```diff
@@ -59,22 +59,36 @@
...
 
 (defn tag
+  "opts
+   {:type         :icon/:emoji/:label
+    :size         32/24
+    :on-press     fn
+    :blurred      true/false 
+    :resource     icon/image
+    :labelled     true/false
+    :disabled     true/false}
```
ibrkhalil: Can we indicate it's a boolean by adding a question mark at the end
`disabled` -> `disabled?`

src/quo2/components/tags/tags.cljs
```diff
@@ -1,7 +1,13 @@
...
   (:require [reagent.core :as reagent]
             [quo.react-native :as rn]
-            [quo2.components.tags.tag :as tag]))
+            [oops.core :refer [oget]]
+            [status-im.ui.components.react :as react]
+            [status-im.utils.core :as utils]
+            [quo2.components.tags.tag :as tag]
+            [quo2.components.tabs.tabs :as tabs]))
+
+(def default-tab-size 32)
```
ibrkhalil, line 10: Can we make it private to this namespace only?
`(def default-tab-size 32)` -> `(def ^:private default-tab-size 32)`

src/quo2/components/tags/tags.cljs
```diff
@@ -26,4 +32,111 @@
...
+             tag-label]])]))))
+
+(defn scrollable-tags
+  "Usage:
+  {:type            :icon/:emoji/:label
+    :size             32/24
+    :on-press         fn
+    :blurred          true/false 
+    :labelled         true/false
+    :disabled         true/false
```
ibrkhalil: Same as previous comment

src/quo2/components/tags/tags.cljs
ibrkhalil: I second Andrey on this ..

src/quo2/components/tags/tags.cljs
jo-mut: cool

src/quo2/components/tags/base_tag.cljs
```diff
@@ -2,31 +2,27 @@
...
-   {:type :icon/:emoji/:label/:permission
-    :size 32/24}
-    :labelled true"
   [_]
-  (fn [{:keys [id size disabled border-color border-width background-color on-press
-               accessibility-label label type] :or   {size 32}} children]
-    [rn/touchable-without-feedback (merge {:disabled            disabled
+  (fn [{:keys [id size disabled? border-color border-width background-color on-press
+               accessibility-label labelled? type] :or   {size 32}} children]
+    [rn/touchable-without-feedback (merge {:disabled?            disabled?
```
J-Son89: hmm, are you sure this is safe?
the :disabled key here is being passed to `rn/touchable-without-feedback` and the prop name there is `disabled` .. so does this work?
https://reactnative.dev/docs/touchablewithoutfeedback#disabled

jo-mut: weird coz it actually disables but i am going to change this back to `disabled` instead of `disabled?` just to be on the safe side

src/quo2/components/tags/tag.cljs
jo-mut: hey @ibrkhalil could just elaborate further to me why you recomend add the `:style` here

ibrkhalil: Styles should go inside the style prop to differentiate them from normal props, I don't know if there's any drawbacks of not doing that but I'd say it's cleaner

jo-mut: thanks @ibrkhalil

src/quo2/components/tags/tags.cljs
```diff
@@ -1,29 +1,142 @@
...
+                                                                       :on-press      (fn [id]
+                                                                                        (reset! active-tab-id id)
+                                                                                        (when scroll-on-press?
+                                                                                          (.scrollToIndex @flat-list-ref
+                                                                                                          #js {:animated     true
+                                                                                                               :index        index
+                                                                                                               :viewPosition 0.5}))
+                                                                                        (when on-change
+                                                                                          (on-change id)))}
+                                                              label]])})])))))
```
ibrkhalil: Can't we just do
```clj
[tabs/scrollable-tabs {:scroll-on-press?    true
                       :fade-end?           true
                       :default-active      1
                       :data                [{:id 1 :label "Music"     :resource (resources/get-image :music)}
                                             {:id 2 :label "Lifestyle" :resource (resources/get-image :lifestyle)}
                                             {:id 2 :label "Podcasts"  :resource (resources/get-image :podcasts)}
                                             {:id 2 :label "Music"     :resource (resources/get-image :music)}
                                             {:id 3 :label "Lifestyle" :resource (resources/get-image :lifestyle)}]
                       :size                32
                       :labelled?           true
                       :type                :emoji
                       :fade-end-percentage 0.4}]
```

ibrkhalil: I mean, If it's the same logic .. We must reuse it.

jo-mut: @ibrkhalil  this would mean refactoring tab and tag component as they both have different logic though slightly similar. We might end up with a very complex component which is something we are also trying to avoid with the redesign. I think its better to have both tag and tab component in separate ns. However what I think is good in this case is to place scrollable-tabs into a separate ns and refactor it so that both scrollable-tags and scrollable-tabs can share the same logic.

ibrkhalil: Got it, thanks John
If you needed any backup .. I'm happy to help

jo-mut:
@ilmotta. I noticed scrollable-tags are used in activity center. Are working working on that? This PR merges the logic for scrollable-tags & scrollable-tabs. I would like to hear your opinion on this especially if it overlaps with your work on activity center

ilmotta:
> @ilmotta. I noticed scrollable-tags are used in activity center. Are working working on that? This PR merges the logic for scrollable-tags & scrollable-tabs. I would like to hear your opinion on this especially if it overlaps with your work on activity center

Thanks for asking @jo-mut. I originally wrote the `quo2.components.tabs.tabs/scrollable-tabs` component for the new Activity Center, but since merged into `develop`, I haven't touched it again. Apart from performance considerations to update the mask length dynamically, it served its purpose so far. 

I see that the new `quo2.components.tags.scrollable-view/view` component can render either tabs or tags, and I couldn't find a breaking change in this new component in relation to the one it was copied from (`scrollable-tabs`), so I think the best in this PR would be eliminate the original `scrollable-tabs` function and just use this new one you created. If we don't do this we'll end up with unnecessary duplication, and it seems to be a _real duplication_, in the sense that if a developer changes `scrollable-view/view` they would always need to update `scrollable-tabs`.

src/quo2/components/tags/scrollable_view.cljs
```diff
@@ -0,0 +1,146 @@
...
+  [{:keys [default-active fade-end-percentage]
+    :or   {fade-end-percentage 0.8}}]
+  (let [active-tab-id (reagent/atom default-active)
+        fading        (reagent/atom {:fade-end-percentage fade-end-percentage})
+        flat-list-ref (atom nil)]
+    (fn [{:keys [data fade-end-percentage fade-end? on-change on-scroll
+                 scroll-event-throttle scroll-on-press? size type labelled?
+                 disabled? blurred? icon-color blur? component
+                 override-theme]
+          :or   {fade-end-percentage fade-end-percentage fade-end? false
```
ilmotta: [Code style] I find the line breaks here in order to save LoC, especially in the `:or` key, make the code rather difficult to read. Here's my suggestion if you like it:

```clojure
{fade-end-percentage   fade-end-percentage
 fade-end?             false
 scroll-event-throttle 64
 scroll-on-press?      false
 size                  default-tab-size}
```

src/quo2/components/tags/scrollable_view.cljs
```diff
@@ -0,0 +1,146 @@
...
+                                                                          :size          size
+                                                                          :active        (= id @active-tab-id)
+                                                                          :resource      resource
+                                                                          :blurred?      blurred?
+                                                                          :icon-color    icon-color
+                                                                          :disabled?     disabled?
+                                                                          :label         label
+                                                                          :type          type
+                                                                          :labelled?     labelled?
+                                                                          :on-press      (fn [id]
```
ilmotta: The `on-press` anonymous functions look identical. If this is true duplication, we could at least extract this function to a let binding.

jo-mut:
> > @ilmotta. I noticed scrollable-tags are used in activity center. Are working working on that? This PR merges the logic for scrollable-tags & scrollable-tabs. I would like to hear your opinion on this especially if it overlaps with your work on activity center
> 
> Thanks for asking @jo-mut. I originally wrote the `quo2.components.tabs.tabs/scrollable-tabs` component for the new Activity Center, but since merged into `develop`, I haven't touched it again. Apart from performance considerations to update the mask length dynamically, it served its purpose so far. 
> 
> I see that the new `quo2.components.tags.scrollable-view/view` component can render either tabs or tags, and I couldn't find a breaking change in this new component in relation to the one it was copied from (`scrollable-tabs`), so I think the best in this PR would be eliminate the original `scrollable-tabs` function and just use this new one you created. If we don't do this we'll end up with unnecessary duplication, and it seems to be a _real duplication_, in the sense that if a developer changes `scrollable-view/view` they would always need to update `scrollable-tabs`.

Thanks for reviewing. Your did a great job on scrollable-tabs. 

I will go ahead and refactor the code to use `scrollable-view` where ` scrollable-tabs` is used.

ilmotta (approved):

jo-mut:
@status-im/mobile-qa I would like to merge this

qoqobolo:
Hi @jo-mut, this PR was automatically moved to the `TO REBASE` column due to conflicts. Could you resolve them first, then move the PR to the `E2E Tests` column, and after that we'll test the PR manually.
Thanks

jo-mut:
this is fine @qoqobolo

flexsurfer:
hey @jo-mut whats' the state of this PR ?

jo-mut:
hi @flexsurfer. Am back from my time off. I will rebase this so that it can be merged as soonest

jo-mut:
@qoqobolo looks good now

src/status_im/ui2/screens/communities/communities_home_bottom_sheet.cljs
```diff
@@ -0,0 +1 @@
+(ns status-im.ui2.screens.communities.communities-home-bottom-sheet)
```
flexsurfer: whats that ?

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -50,7 +51,7 @@
 (defn tabs
   []
   (let [filter-type (rf/sub [:activity-center/filter-type])]
-    [quo2/scrollable-tabs {:size                32
+    [scrollable-view/view {:size                32
```
flexsurfer: we should use quo2.core and name which is used in figma design

flexsurfer: is the name of component in figma scrollable-view ?

src/quo2/components/community/community_view.cljs
```diff
@@ -18,15 +18,15 @@
...
+    count]])
 
 (defn community-stats-column [type]
   (let [icon-color (colors/theme-colors colors/neutral-50 colors/neutral-40)]
     [rn/view (if (= type :card-view)
                (style/card-stats-container)
                (style/list-stats-container))
-     [community-stats {:icon       :i/group
-                       :members-count "629.2K" ;;TODO here should be formatted value, use money/format-members from outside this component
+     [community-stats {:icon       :main-icons2/group
```
flexsurfer: `:i/group` should be used

src/quo2/components/community/community_view.cljs
```diff
@@ -18,15 +18,15 @@
                       :color           icon-color}]]
    [text/text {:weight :regular
                :size   :paragraph-1}
-    members-count]])
+    count]])
```
flexsurfer: count is bad, because there is count function in cljs core

flexsurfer (changes requested):

src/status_im/ui/screens/activity_center/views.cljs
jo-mut: scrollable-view is a namespace that provide functionality that is shared between scrollable-tags and scrollable-tabs. I didnt to avoid duplicate code. The components in figma are tabs and tags and not scrollable-view

flexsurfer: @ilmotta @J-Son89 any thoughts?

J-Son89: hmm, looking into this now. it seems the functionality of scrollable view should be encapsulated in the tabs and tags...

e.g we have `tabs` and `tags` and then also `scrollable-tabs` and `scrollable-tags` components, which are all exported from quo2 library. 
As far as I see, the current implementation still leaves a bit of work on the developer to get the "scrollable" behaviour. It should either be a separate component for both tabs and tags.. or both of these components should be configurable with a prop.. e,g `scrollable?`


Also if this scrollable view is common between tabs and tags then the namespace depth should show that.
i.e you have this in
 `quo2.components.tags.scrollable-view` however it should be in something like `quo2.components.common.scrollable-view` or  `quo2.components.scrollable-view`.

flexsurfer: makes sense, @jo-mut  wdyt?

ilmotta: > As far as I see, the current implementation still leaves a bit of work on the developer to get the "scrollable" behaviour. It should either be a separate component for both tabs and tags.. or both of these components should be configurable with a prop.. e,g scrollable?

I agree. When I first coded it I wanted something flexible given the uncertainty of its usage. Now that we see tags also using it and the solution works, I would prefer if the tabs and tags components simply had a flag to specify if it was scrollable or not as @J-Son89 said.

But, IMO this should be done in a separate PR, because this PR is working fine.

If we open [Design System > Tabs](https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=57%3A13214&t=ISb34KWjWoEyP193-0), there's no specification about scrollable appearance. I think this needs to be fixed tbh, so something to bring up to designer's attention @jo-mut.

flexsurfer (approved):

src/status_im/ui/screens/activity_center/views.cljs
jo-mut: @flexsurfer i think @J-Son89 suggestions are very good in this case. It would be better having scrollable-tags/tabs as separate components. I will work on it

src/quo2/components/tags/tag.cljs
```diff
@@ -59,22 +59,36 @@
...
+  
+   opts
+    - `blurred`  boolean: use to determine border color if the background is blurred
+    - `type`     can be icon or emoji with or without a tag label
+    - `labelled` boolean: is true if tag has label else false"
   [_ _]
-  (fn [{:keys [id on-press disabled size resource active accessibility-label
-               label type labelled blurred icon-color] :or   {size 32}}]
-    (let [state (cond disabled :disabled active :active :else :default)
+  (fn [{:keys [id on-press disabled? size resource active accessibility-label
```
J-Son89, line 85: so you have adjusted a lot of props here. Can you make sure that places where this component are used across the codebase are also updated to this api 🙏

src/status_im/ui/screens/activity_center/views.cljs
jo-mut: @ilmotta i agree. I will talk to design on this.

jo-mut: @ilmotta @J-Son89 this means the current implementation of tabs and tags components will be replaced by scrollable-tags and scrollable-tabs component with the flag being what makes the difference whether tabs/tags are scrollable or not. Going by that I don't think we would need what we currently have

J-Son89: imo:
I don't think you need to rename the tabs and tags components to `scrollable-*` if you're going to use a `scrollable?` prop

Either make tabs and tags scrollable with a prop, so don't change the name of them..

e.g
```
(defn tags [{ scrollable? .....}]
```

```
(defn tabs [{ scrollable? .....}]
```
or 

keep tabs and tags as is and then make 2 new components with the scrollable capabilities:
i.e there is 4 components, tabs and tags and scrollable-tabs and scrollable-tags

ilmotta: Sounds good @jo-mut, @J-Son89. I would just prefer if this refactoring was carried in another PR, because this one was already reviewed with the previous scope.

But of course, up to you.

ios/Podfile.lock
```diff
@@ -640,7 +640,7 @@ SPEC CHECKSUMS:
   FBLazyVector: 352a8ca9bbc8e2f097d680747a8c97ecef12d469
   FBReactNativeSpec: 7dfb84f624136a45727c813ed21d130cd3e61beb
   Folly: b73c3869541e86821df3c387eb0af5f65addfab4
-  glog: 36ce0530c6d2c3a5a4326885ef4069564887a1db
+  glog: 997518ea2aa2d8cd5df9797b641b758d52ecf2bc
```
flexsurfer: please rollback this file

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -0,0 +1,151 @@
+(ns status-im.ui.screens.activity-center.views
```
flexsurfer: rollback pls

flexsurfer (changes requested):

src/status_im/ui/screens/activity_center/views.cljs
flexsurfer: still not addressed

flexsurfer: this file has been deleted, you need to rollback it and remove

flexsurfer: cc @ilmotta

ilmotta: Right @flexsurfer, this file is long gone.

jo-mut: @flexsurfer  i missed your comment before. I have addressed that now

flexsurfer (approved):

jo-mut:
@status-im/mobile-qa  could please have a look at this PR

VolodLytvynenko:
> @status-im/mobile-qa could please have a look at this PR

Hi @jo-mut I will check it and provide feedback. Thank you!

VolodLytvynenko:
@jo-mut  Thank you for your work. PR is tested and ready to be merged. I have done regression tests and tests related to the  activity center

PR #14186, opened 2022-10-19
Helper to format number to 100000 -> 100k/ 1000000000 -> 1b/ 1000000 -> 1m

### LINKED ISSUE #14124: Helper for N thousands to be NK
# Problem

We need a helper to parse N thousands into NK format

Examples
1000 -> 1K
10000 -> 10K
10001 -> 10K
10100 -> 10.1K

ibrkhalil:
fixes #14124 

status: ready


src/status_im/utils/number.cljs
```diff
@@ -0,0 +1,25 @@
...
+    (if-not (js/isNaN numeric-value)
+      (if (<= 1000 numeric-value)
+        (str (quot numeric-value 1000) (if capitalize?
+                                         (upper-case "k")
+                                         "k"))
+        (str numeric-value))
+      "Invalid Number")))
+
+(defn number->nk
+  "Returns the thousands in a number in NK format
```
J-Son89: consider adding some examples like this:
Examples
1000 -> 1K
10000 -> 10K
10001 -> 10K
10100 -> 10.1K

I didn't know what NK was until I checked the issue description

src/status_im/utils/number.cljs
```diff
@@ -0,0 +1,25 @@
+(ns status-im.utils.number
+  "Contains helpers for numbers"
+  (:require [clojure.string :refer [upper-case]]))
+
+(defn- number-to-nk [number-string capitalize?]
+  (let [numeric-value (if (> 0 number-string)
```
J-Son89: consider using pos?
`(if (pos? number-string)`

J-Son89 (approved):

src/status_im/utils/number.cljs
```diff
@@ -0,0 +1,25 @@
...
+                        (throw (str "Invalid input " number-string))
+                        (js/parseInt number-string))]
+    (if-not (js/isNaN numeric-value)
+      (if (<= 1000 numeric-value)
+        (str (quot numeric-value 1000) (if capitalize?
+                                         (upper-case "k")
```
flexsurfer: why not to use "K" ?

src/status_im/utils/number.cljs
```diff
@@ -0,0 +1,25 @@
+(ns status-im.utils.number
+  "Contains helpers for numbers"
+  (:require [clojure.string :refer [upper-case]]))
+
+(defn- number-to-nk [number-string capitalize?]
+  (let [numeric-value (if (> 0 number-string)
+                        (throw (str "Invalid input " number-string))
+                        (js/parseInt number-string))]
+    (if-not (js/isNaN numeric-value)
+      (if (<= 1000 numeric-value)
```
flexsurfer: kinda hard to read, when number is the first argument, why not `(> numeric-value 1000)` ?

src/status_im/utils/number.cljs
flexsurfer: pos?

cammellos (commented):
@ibrkhalil this a good place to add some unit tests, since it's easily testable, would you mind adding some please?

src/status_im/utils/number.cljs
ibrkhalil: Added the case if the developer wanted it to be upper case with a param

src/status_im/utils/number.cljs
ibrkhalil: Thanks Jaime, Improved

src/status_im/utils/number.cljs
ibrkhalil: Fixed, Thanks Andrey

src/status_im/utils/number.cljs
ibrkhalil: Fixed

ibrkhalil:
> @ibrkhalil this a good place to add some unit tests, since it's easily testable, would you mind adding some please?

Done, Thanks Andrea

src/status_im/utils/number.cljs
```diff
@@ -1,4 +1,6 @@
-(ns status-im.utils.number)
+(ns status-im.utils.number
+  "Contains helpers for numbers"
```
ilmotta: [Nitpick] Although I love the idea of documenting certain namespaces, this one is very self-explanatory, so the docstring feels superfluous.

src/status_im/utils/number_test.cljs
```diff
@@ -0,0 +1,19 @@
...
+  (testing "Bad cases"
+    (testing "nil"
+      (is (thrown-with-msg? js/Error #"Invalid Number" (number/number->nk nil))))
+    (testing "negative"
+      (is (thrown-with-msg? js/Error #"Invalid Number" (number/number->nk "-1"))))
+    (testing "Empty string"
+      (is (thrown-with-msg? js/Error #"Invalid Number" (number/number->nk ""))))
+    (testing "undefined"
+      (is (thrown-with-msg? js/Error #"Invalid Number" (number/number->nk js/undefined)))))
+  (testing "Good cases"
```
ilmotta: Interestingly, the `number->nk` function docstring has more examples than its tests, so I think you could add a couple more good cases as actual tests. For example, I'm not seeing the example for uppercase.

To make things easier, with less boilerplate, you could use `cljs.test/are`. I'm a huge fan of this macro when you need to test lots of input/output cases for simple pure functions like `number->nk`. You can see one example in status-mobile https://github.com/status-im/status-mobile/blob/develop/src/status_im/router/core_test.cljs#L70

src/status_im/utils/number.cljs
```diff
@@ -13,3 +15,37 @@
   (let [scale (Math/pow 10 decimal-places)]
     (/ (Math/round (* n scale))
        scale)))
+
+(defn- caught-invalid-number []
+  (throw (js/Error "Invalid Number")))
+
+(defn- number-to-nk [number-string capitalize?]
+  (let [parsed-number (js/parseInt number-string)
```
ilmotta: It's generally recommended to add the radix when parsing numbers. In this example, it'd be `(js/parseInt number-string 10)`. There are certain strings that are parsed differently in different runtimes.

ilmotta (approved):

src/status_im/utils/number.cljs
ibrkhalil: You're right, Icaro .. It shouldn't need a docstring

src/status_im/utils/number_test.cljs
ibrkhalil: Wow, I am astonished by `are`, Thanks for letting me know about it!

src/status_im/utils/number.cljs
ibrkhalil: Okay

ibrkhalil:
So @erikseppanen suggested to check if there's an existing library for formatting strings and numbers .. What you think ?
Should we do ourselves things that other people might have already done?
Like when should we use a library?
CC: @flexsurfer @cammellos @ilmotta @J-Son89 @jakubgs

flexsurfer:
its always better to have our own code when possible, for security reasons

src/status_im/utils/number_test.cljs
```diff
@@ -0,0 +1,21 @@
+(ns status-im.utils.number-test
+  (:require [cljs.test :refer-macros [deftest testing are]]
+            [status-im.utils.number :as number]))
+
+(deftest number->nk-test
+  (testing "Bad cases"
```
churik: Thank you for tagging me and for unit test! 👍 
1) IMO naming "Positive cases" / "Negative cases" is better 
2) I'd extend positive scenarios:
 - values that should be converted to decimals (i.e. 1100 -> 1.1K)
 - decimals values that should not be converted 0.01; 0 
 3) Negative scenarios with:
  - Exponential format: 1e3
  - Decimal with unexpected delimiter,  i.e. 100,3
  
  Tell me if it doesn't make sense

ilmotta:
Important questions @ibrkhalil, thanks for bringing them up.

> So @erikseppanen suggested to check if there's an existing library for formatting strings and numbers .. What you think?

I talked to @erikseppanen yesterday and he mentioned we might be able to replace some or all of the logic to format numbers using the core function `cljs.pprint/cl-format`. Indeed, a very interesting approach. It's not an external library. Whether such function can help or should be used in this PR or future PRs is of course open for discussion. I never used any lib that does what this PR is doing, so I can't comment on it.

**I think it's fine to merge this PR as it is correct, has tests**, etc. I think it's up to you to decide if it's worth spending more time in a different solution.

> Should we do ourselves things that other people might have already done?

Not reinventing the wheel is usually good, but we have to be careful not to use external diamond shape wheels when Status needs them round, if you know what I mean ;) It's a case by case analysis really.

> Like when should we use a library?

I think this question should always be raised within a specific context (e.g. which lib, what are the problems it solves, what are the incidental complexities of introducing it, etc). I can't answer generally because there are always tradeoffs on every choice.

erikseppanen:
Thanks for chiming in @ilmotta!

> > So @erikseppanen suggested to check if there's an existing library for formatting strings and numbers .. What you think?
> 
> I talked to @erikseppanen yesterday and he mentioned we might be able to replace some or all of the logic to format numbers using the core function `cljs.pprint/cl-format`. Indeed, a very interesting approach. It's not an external library. Whether such function can help or should be used in this PR or future PRs is of course open for discussion. I never used any lib that does what this PR is doing, so I can't comment on it.

Yes, this PR looks like its intent is to 'add metric or SI-prefix formatting', and I wondered if that functionality already exists, since it seems so low level.  And I wondered if the scope would creep, like supporting other SI prefixes, locale, etc., and if so do we want to be in the business of writing that kind of library (or namespace in our case). If it was already in a core library, I figured that would be a slam-dunk... I didn't look that much, I just ran across `cljs.pprint/cl-format`, which may not fit the bill, in which case, fine, at least we looked.

> **I think it's fine to merge this PR as it is correct, has tests**, etc. I think it's up to you to decide if it's worth spending more time in a different solution.

I agree, nice job @ibrkhalil!

> > Should we do ourselves things that other people might have already done?
> 
> Not reinventing the wheel is usually good, but we have to be careful not to use external diamond shape wheels when Status needs them round, if you know what I mean ;) It's a case by case analysis really.
> 
> > Like when should we use a library?
> 
> I think this question should always be raised within a specific context (e.g. which lib, what are the problems it solves, what are the incidental complexities of introducing it, etc). I can't answer generally because there are always tradeoffs on every choice.

I agree, lots of things to consider (including security @flexsurfer) and be wary of!

src/status_im/utils/number_test.cljs
ibrkhalil: Done 😄

src/status_im/utils/number.cljs
```diff
@@ -13,3 +16,48 @@
   (let [scale (Math/pow 10 decimal-places)]
     (/ (Math/round (* n scale))
        scale)))
+
+(defn- caught-invalid-number []
+  (throw (js/Error "Invalid Number")))
+
+(defn- with-precision [precision input-number]
+  (gstring/format (str "%." precision "f") input-number))
```
ilmotta, line 23: Excellent work in this namespace @ibrkhalil!

I saw this in the documentation for [google.string.format](https://google.github.io/closure-library/api/goog.string.format.html). If you are in the mood, you could replace calls to `with-precision` to use `toFixed` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed).

> Performs sprintf-like conversion, i.e. puts the values in a template. DO NOT use it instead of built-in conversions in simple cases such as 'Cost: %.2f' as it would introduce unnecessary latency opposed to 'Cost: ' + cost.toFixed(2).

ibrkhalil: Thanks Icaro
Is there any difference between the JS function toFixed and CLJS's with-precision .. if not, let's leave it as is.

du82:
Do we account for M and B, million and billion? not sure now necessary billion would be, but I could see communities having millions of users.

src/status_im/utils/number.cljs
ilmotta: AFAIU, `with-precision` is a core function only available in Clojure JVM. At least based on the quote I shared above, `google.string` says it's safe and actually recommended to not use `format` if `toFixed` solves the problem. It's no big deal, right? I just thought you'd like to know about such recommendation.

So, in the end, it seems this PR's implementation of `with-precision` does exactly the same as `toFixed`, but I haven't tested.

ibrkhalil:
> Do we account for M and B, million and billion? not sure now necessary billion would be, but I could see communities having millions of users.

Good thought, will make it work with these cases and add tests for them and then merge

src/status_im/utils/number.cljs
ibrkhalil: Thanks for the recommendation Icaro, Of course that's helpful to know
I'll add the case for millions and billions alongside their respective tests and will try `toFixed` alongside

ibrkhalil:
> Do we account for M and B, million and billion? not sure now necessary billion would be, but I could see communities having millions of users.

Added

PR #14205, opened 2022-10-21
android: use projectDir for VERSION file path

jakubgs:
Otherwise Android Studio can't find the relative path because it uses `/` as it's working directory. Which makes no sense.


qfrank (approved):
works for me, thank you

PR #14223, opened 2022-10-25
Display contact verification requests and allow users to decline them

### LINKED ISSUE #14171: Display identity requests
# Feature Issue

Display identity requests that can be declined or accepted in the Activity Center.

img

### Acceptance criteria

>Given that Alice already accepted a contact request from Carol
>When Alice sends an identity verification to Carol
>Then Carol receives a new notification in her Activity Center
>And the notification message displays the challenge

>Given that Carol received an identity request from Alice
>Then Carol can either Accept or Decline the challenge

>Given that Carol received an identity request from Alice
>When Carol declines the challenge
>Then she sees the same notification change its status to "Declined"

### Out of scope

- There's no need to implement the actual acceptance flow, as this is more complex and can be done using a separate issue. A dummy event handler might suffice.

### Notes

At the moment (2022-10-18), it's necessary to use the unmerged status-go branch https://github.com/status-im/status-go/tree/feature/activity-center-read-unread.

### Figma

Identity request flows are described in https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926.

ilmotta:
fixes #14171 

### Summary

Displays contact (identity) verification requests in the Activity Center and allows the receiver of such requests to decline them.

- [X] Fix reconciliation when notifications are also present in the "All" tab.
- [X] Introduces a new test helper to verify log messages. See `status-im.test-helpers/using-log-test-appender`.

identity-verification-example

### Review notes

- The new test utility `status-im.test-helpers/using-log-test-appender` is used in this PR to verify a dead end (a failure event handler without user feedback) at least logs critical information in case users need to report a bug.

- As new notification types are implemented, @smohamedjavid and I agreed to start taking advantage of multimethods to completely decouple different types of notifications. In the future we'll probably find new abstractions or ways to unify them.

- Status-go is under development to add support for notifications in the existing contact verification flows. The branch https://github.com/status-im/status-go/tree/feature/activity-center-read-unread is not yet merged and it's still prone to change before the final merge, but the implementation seems stable enough for us to move forward with the mobile implementation.

#### Platforms

- Android
- iOS

#### Areas that may be impacted

It's still completely isolated from the rest of the app and behind a feature toggle that's disabled by default.

### Steps to test

- Force mobile app to use compiled version from https://github.com/status-im/status-go/tree/feature/activity-center-read-unread.
- Enable `status-im.utils.config/new-activity-center-enabled?`
- Establish mutual contact approval between two contacts.
- Send contact verification from user A to B either by using a running REPL with the example code below, or using the Desktop client.
- Users can send as many contact verification requests as desired.
- Check that the user receiving such challenges can decline them.

```clojure
(rf/reg-event-fx
   :dev/send-contact-verification-request
   (fn [{:keys [_db]} [_ {:keys [contact-id challenge]}]]
     {::json-rpc/call [{:method     "wakuext_sendContactVerificationRequest"
                        :params     [contact-id challenge]
                        :on-success #(do
                                       (tap> {:success %})
                                       nil)
                        :on-error   #(do
                                       (tap> {:error %})
                                       nil)}]}))

(def alice-id "0x04d0")
(def carol-id  "0x0445")

;; Logged in as Alice, dispatch this event:
(rf/dispatch [:dev/send-contact-verification-request
                {:challenge  "Carol, who am I really?"
                 :contact-id carol-id}])
```

status: ready

### DIFF
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -43,6 +43,32 @@
     {:db (update-in db [:activity-center :notifications]
                     update-notifications new-notifications)}))
 
+;;;; Contact verification
+
+(fx/defn contact-verification-decline
+  {:events [:activity-center.contact-verification/decline]}
+  [_ contact-verification-id]
+  {::json-rpc/call [{:method     "wakuext_declineContactVerificationRequest"
+                     :params     [contact-verification-id]
+                     :on-success #(re-frame/dispatch [:activity-center.contact-verification/decline-success %])
+                     :on-error   #(re-frame/dispatch [:activity-center.contact-verification/decline-error contact-verification-id %])}]})
+
+(fx/defn contact-verification-decline-success
+  {:events [:activity-center.contact-verification/decline-success]}
+  [cofx response]
+  (->> response
+       :activityCenterNotifications
+       (map data-store.activities/<-rpc)
+       (notifications-reconcile cofx)))
+
+(fx/defn contact-verification-decline-error
+  {:events [:activity-center.contact-verification/decline-error]}
+  [_ contact-verification-id error]
+  (log/warn "Failed to decline contact verification"
+            {:contact-verification-id contact-verification-id
+             :error                   error})
+  nil)
+
 ;;;; Notifications fetching and pagination
 
 (def defaults
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -1,6 +1,6 @@
 (ns status-im.activity-center.core
-  (:require [re-frame.core :as re-frame]
-            [status-im.constants :as constants]
+  (:require [re-frame.core :as rf]
+            [status-im.constants :as c]
             [status-im.data-store.activities :as data-store.activities]
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.fx :as fx]
@@ -29,12 +29,12 @@
               (as-> acc $
                 (update-in $ [type :read :data] remove-notification)
                 (update-in $ [type :unread :data] remove-notification)
-                (update-in $ [constants/activity-center-notification-type-no-type :read :data] remove-notification)
-                (update-in $ [constants/activity-center-notification-type-no-type :unread :data] remove-notification)
+                (update-in $ [c/activity-center-notification-type-no-type :read :data] remove-notification)
+                (update-in $ [c/activity-center-notification-type-no-type :unread :data] remove-notification)
                 (if (or (:dismissed notification) (:accepted notification))
                   $
                   (-> $ (update-in [type filter-status :data] insert-and-sort)
-                      (update-in [constants/activity-center-notification-type-no-type filter-status :data] insert-and-sort))))))
+                      (update-in [c/activity-center-notification-type-no-type filter-status :data] insert-and-sort))))))
           db-notifications
           new-notifications))
 
@@ -52,8 +52,8 @@
   [_ contact-verification-id]
   {::json-rpc/call [{:method     "wakuext_declineContactVerificationRequest"
                      :params     [contact-verification-id]
-                     :on-success #(re-frame/dispatch [:activity-center.contact-verification/decline-success %])
-                     :on-error   #(re-frame/dispatch [:activity-center.contact-verification/decline-error contact-verification-id %])}]})
+                     :on-success #(rf/dispatch [:activity-center.contact-verification/decline-success %])
+                     :on-error   #(rf/dispatch [:activity-center.contact-verification/decline-error contact-verification-id %])}]})
 
 (fx/defn contact-verification-decline-success
   {:events [:activity-center.contact-verification/decline-success]}
@@ -75,7 +75,7 @@
 
 (def defaults
   {:filter-status          :unread
-   :filter-type            constants/activity-center-notification-type-no-type
+   :filter-type            c/activity-center-notification-type-no-type
    :notifications-per-page 10})
 
 (def start-or-end-cursor
@@ -98,8 +98,8 @@
     {:db             (assoc-in db [:activity-center :notifications filter-type filter-status :loading?] true)
      ::json-rpc/call [{:method     (filter-status->rpc-method filter-status)
                        :params     [cursor (defaults :notifications-per-page) filter-type]
-                       :on-success #(re-frame/dispatch [:activity-center.notifications/fetch-success filter-type filter-status reset-data? %])
-                       :on-error   #(re-frame/dispatch [:activity-center.notifications/fetch-error filter-type filter-status %])}]}))
+                       :on-success #(rf/dispatch [:activity-center.notifications/fetch-success filter-type filter-status reset-data? %])
+                       :on-error   #(rf/dispatch [:activity-center.notifications/fetch-error filter-type filter-status %])}]}))
 
 (fx/defn notifications-fetch-first-page
   {:events [:activity-center.notifications/fetch-first-page]}
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -18,21 +18,23 @@
   ~excessively~ big, this implementation will probably need to be revisited."
   [db-notifications new-notifications]
   (reduce (fn [acc {:keys [id type read] :as notification}]
-            (let [filter-status (if read :read :unread)]
-              (cond-> (-> acc
-                          (update-in [type :read :data]
-                                     (fn [data]
-                                       (remove #(= id (:id %)) data)))
-                          (update-in [type :unread :data]
-                                     (fn [data]
-                                       (remove #(= id (:id %)) data))))
-                (not (or (:dismissed notification) (:accepted notification)))
-                (update-in [type filter-status :data]
-                           (fn [data]
-                             (->> notification
-                                  (conj data)
-                                  (sort-by (juxt :timestamp :id))
-                                  reverse))))))
+            (let [filter-status       (if read :read :unread)
+                  remove-notification (fn [data]
+                                        (remove #(= id (:id %)) data))
+                  insert-and-sort     (fn [data]
+                                        (->> notification
+                                             (conj data)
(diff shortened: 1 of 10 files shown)


src/status_im/activity_center/core_test.cljs
```diff
@@ -12,15 +12,105 @@
...
+                                                                                       :sticker     nil
+                                                                                       :text        nil}
+                                                                  :outgoing           false
+                                                                  :outgoing-status    nil
+                                                                  :quoted-message     nil}
+                                    :name                        "0x04d03f"
+                                    :read                        true
+                                    :reply-message               nil
+                                    :timestamp                   1666647286000
+                                    :type                        6}]
```
OmarBasem: Why not use a constant here for `type`?

src/status_im/activity_center/core_test.cljs
```diff
@@ -12,15 +12,105 @@
...
+                                   :id                        contact-verification-id
+                                   :message                   {}
+                                   :name                      "0x04d03f"
+                                   :read                      true
+                                   :timestamp                 1666647286000
+                                   :type                      c/activity-center-notification-type-contact-verification}]}))
+
+       (h/spy-fx spy-queue ::json-rpc/call)
+
+       (rf/dispatch [:activity-center.contact-verification/decline contact-verification-id])
```
OmarBasem, line 75: I think we agreed to use `>evt` instead of `rf/dispatch`.

src/status_im/data_store/activities_test.cljs
```diff
@@ -0,0 +1,95 @@
...
+   :replyMessage              {}})
+
+(deftest <-rpc-test
+  (testing "assocs random chat color"
+    (is (contains? (set colors/chat-colors) (:color (store/<-rpc raw-notification)))))
+
+  (testing "renames keys"
+    (is (= {:name                        "0x04c661"
+            :chat-id                     "0x04c66155"
+            :contact-verification-status constants/contact-verification-state-pending}
```
OmarBasem: Maybe it would be better to define `name` and `chat-id` in a variable since they are used multiple times.

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -15,98 +15,114 @@
...
+                   :icon      :main-icons2/add-user
+                   :timestamp (datetime/timestamp->relative (:timestamp notification))
+                   :unread?   (not (:read notification))
+                   :context   [[context-tags/user-avatar-tag
+                                {:color          :purple
+                                 :override-theme :dark
+                                 :size           :small
+                                 :style          {:background-color colors/white-opa-10}
+                                 :text-style     {:color colors/white}}
+                                (sender-name contact)
```
OmarBasem, line 55: I don't see any `colors/theme-colors` usage. Just checking if you tested the UI in dark mode.

translations/en.json
```diff
@@ -1482,6 +1482,7 @@
     "add-nickname": "Add a nickname (optional)",
     "nickname-description": "Nicknames help you identify others in Status.\nOnly you can see the nicknames you’ve added",
     "accept": "Accept",
+    "reply": "Reply",
```
OmarBasem: There is already a "Reply" translation line 869 "message-reply". Maybe delete one of them.

src/status_im/activity_center/core.cljs
```diff
@@ -43,11 +45,37 @@
...
 
+;;;; Contact verification
+
+(fx/defn contact-verification-decline
+  {:events [:activity-center.contact-verification/decline]}
+  [_ contact-verification-id]
+  {::json-rpc/call [{:method     "wakuext_declineContactVerificationRequest"
+                     :params     [contact-verification-id]
+                     :on-success #(rf/dispatch [:activity-center.contact-verification/decline-success %])
+                     :on-error   #(rf/dispatch [:activity-center.contact-verification/decline-error contact-verification-id %])}]})
```
OmarBasem, line 56: Here also `>evt`

Letting you know there is still more instances to avoid spam :)

src/status_im/activity_center/core_test.cljs
ilmotta: Right, we did (although with reluctance from some people, given how little value it provides, including me hehe). I'll make sure to change all calls in this PR, even in tests.

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: The AC is supposed to always be viewed as "if in dark mode", contrary to other parts of the app that adapt to light/dark. But @OmarBasem, there's an outstanding issue after some recent changes in `develop` that made the AC look bad in light mode, but @smohamedjavid and I prefer to fix this in a separate issue if that's okay by you.

translations/en.json
ilmotta: I created this one because `message-reply` has a different contextual application, it's just a coincidence that they refer to the same word. In any case, I can make this change.

src/status_im/activity_center/core_test.cljs
ilmotta: Good catch.

src/status_im/activity_center/core.cljs
```diff
@@ -1,6 +1,6 @@
 (ns status-im.activity-center.core
-  (:require [re-frame.core :as re-frame]
-            [status-im.constants :as constants]
+  (:require [re-frame.core :as rf]
+            [status-im.constants :as c]
```
briansztamfater: No strong opinion tbh, but I'd prefer more descriptive aliases over shorter (if cannot do both). Maybe use `const` to make it shorter as it is currently but keeping the self-descriptiveness?

ilmotta: I'm with you @briansztamfater. I like `const`, just `constants` (that's used throughout the codebase) is too verbose once you need to type it all the time.

src/status_im/activity_center/core_test.cljs
ilmotta: Hey @OmarBasem, I found a major blocker for using `>evt` both in the implementation and in tests. The `day8.re-frame.test/run-test-sync` (or the async variant) macros rely specifically on `re-frame.core/dispatch` function. If we use `>evt` in the implementation/test all tests fail.

See this comment in their docs:

> Think of it as though every dispatch in your app had been magically turned into dispatch-sync, and re-frame had lifted the restriction that says you can't call dispatch-sync from within an event handler.

I know it's a pain to go back and forth between code conventions, but given `>evt` is just an alias to `dispatch` that doesn't add any other value, I think we should just ditch this convention in favor of testable code, but this is just one guys' opinion.

Definitely something the team should be aware of, but for this PR I'm reverting the usage of `>evt` in event handlers and its tests.

ilmotta: Again, just to be clear, this affects event handlers and their tests. The UI is fine for now to use `>evt`.

src/status_im/activity_center/core.cljs
ilmotta: Done :)

translations/en.json
ilmotta: Done :)

src/status_im/activity_center/core_test.cljs
OmarBasem: @ilmotta I don't like it either, but there needs to be consistency on that call, as it is probably the most common method call in the codebase (along with `<sub` / `re-frame/subscribe`). We can discuss this with Andrey and the team later.

src/status_im/ui/screens/activity_center/views.cljs
OmarBasem: Of course it's okay by me :)

OmarBasem (approved):

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -15,98 +15,114 @@
...
-        sender-name
-        (multiaccounts/displayed-photo contact)]
-       [rn/text {:style {:color colors/white}}
-        (i18n/label :t/contact-request-sent)]])
-    nil))
+(defmethod notification-component constants/activity-center-notification-type-contact-request
+  [{:keys [id] :as notification}]
+  (let [message   (or (:message notification) (:last-message notification))
+        contact   (<sub [:contacts/contact-by-identity (:author notification)])
+        pressable (case (:contact-request-state message)
```
ibrkhalil, line 33: Can we use `if` since we have only two possible options?

ibrkhalil: After thinking about it for a bit .. I think the current approach is better, Since it's easier to extend with other values.

ibrkhalil (approved):
l(great)tm 😅🚀

src/status_im/ui/screens/activity_center/views.cljs
ilmotta: That's the idea @ibrkhalil, in theory we'll have different pressable alternatives, but maybe not. I'll keep an eye on this and simplify. Thanks for the feedback, YAGNI is a very important concept.

src/status_im/activity_center/core_test.cljs
erikseppanen: I wasn't sure what decision was made on this, the last thing I saw was Andrey's poll that seemed to imply that we'd be using `rf/sub` and `rf/dispatch`: 8 votes for, 1 vote against. Lots of preferences were voiced but that's the only thing that stood out for me.  Maybe when these discussions happen we can solidify the decision by maintaining a status-specific addendum to the clojure style guide, like this:

```clojure
;; good
rf/sub and rf/dispatch

;; bad
<sub and >evt
```

ilmotta: Thanks @erikseppanen for the clarification. I think @OmarBasem and I skipped over that conclusion in Discord. It does look like the majority of the team prefers to use the core re-frame functions, and not the aliases.

Just to confirm, by `rf/sub` I take we mean `rf/subscribe`, right?

PR #14250, opened 2022-10-27
Fix unbounded number of calls to fetch notifications

### LINKED ISSUE #14248: Infinite number of calls to fetch first page
## Problem 

Opening the new Activity Center with at least one notification causes an infinite number of calls to fetch the first page of notifications.

#### Expected behavior

Fetch the first page once.

#### Root cause

When notifications are fetched from status-go, the namespace `status-im.data-store.activities/<-rpc` augments the notifications with a random color in the set `quo.design-system.colors/chat-colors`. This is no longer needed and the visual appearance of a notification shouldn't be a responsibility of the data store in the first place.

Because notifications are given random color strings, the `quo.react/effect!` dependencies' vector equality check always gives false.

#### Solution

Remove the random color key from `status-im.data-store.activities/<-rpc`.

ilmotta:
fixes #14248 

### Summary

Please, see the issue for all details. It's a very common issue in React apps using immutable data structures, but it's hard to spot in PR reviews.

### For reviewers

Suggestion: enable _Hide whitespace_, because this PR's diff is sort of big because tests had to be surrounded by a `with-redefs`.

image

#### Platforms

- Android
- iOS

It should be a harmless fix, given that we're only removing the random notification color if the new Activity Center flag is enabled.

status: ready

### DIFF
--- a/src/status_im/activity_center/core_test.cljs
+++ b/src/status_im/activity_center/core_test.cljs
@@ -5,7 +5,8 @@
             [status-im.constants :as const]
             [status-im.ethereum.json-rpc :as json-rpc]
             status-im.events
-            [status-im.test-helpers :as h]))
+            [status-im.test-helpers :as h]
+            [status-im.utils.config :as config]))
 
 (defn setup []
   (h/register-helper-events)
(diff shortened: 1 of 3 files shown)


smohamedjavid:
Thank you for the fix @ilmotta! 
The PR looks good!

smohamedjavid (approved):

PR #14257, opened 2022-10-28
fix #14236 Sometimes profile pictures don't load

### LINKED ISSUE #14236: Sometimes profile pictures don't load
# Bug Report

## Problem 

There are several cases when user profile pictures stop displaying. 
Screenshot 2022-10-26 at 13 33 32

Sometimes the user's own profile image stops loading for the user on the profile screen, and sometimes a profile picture of a contact disappears from the chat list screen (for a 1-1 chat). 

However, the exact steps for reproduction could only be found for two cases.

### CASE 1: Images of public chat members don't load when joining a new public chat until reopening the chat view

**Steps:**
1. Install Status and create an account 
2. Join some public chat with chat history (e.g. #status)
3. Wait until the history is loaded

**Expected result:** all images are loaded
**Actual result:** some of the profile images are not displayed

**OS:** This case is only constantly reproducible on an **iOS** device, failed to replicate it on Android. 

https://user-images.githubusercontent.com/67952253/198018251-aacc4eab-54f6-430b-861f-14fc03814bba.MP4

**Logs**

[geth.log](https://github.com/status-im/status-mobile/files/9868869/geth.log)
[Status.log](https://github.com/status-im/status-mobile/files/9868870/Status.log)

### CASE 2: Own profile picture stops loading on the `Chats` screen after adding a custom profile image

**Steps:**
1. Open Profile > Press on the profile picture > Select a picture
2. Navigate to the `Chats` screen

Screenshot 2022-10-26 at 13 52 25

**OS:** Android, iOS

**Logs** 

[geth.log](https://github.com/status-im/status-mobile/files/9869139/geth.log)

### Additional Information

- Status version: nightly 26/10

qfrank:
fixes #14236

### DIFF
--- a/src/status_im/subs/chat/chats.cljs
+++ b/src/status_im/subs/chat/chats.cljs
@@ -228,10 +228,10 @@
  :<- [:contacts/contacts]
  :<- [:profile/multiaccount]
  :<- [:mediaserver/port]
- (fn [[contacts multiaccount port] [_ id]]
-   (let [contact (or (get contacts id)
-                     (when (= id (:public-key multiaccount))
-                       multiaccount))]
+ (fn [[contacts {:keys [public-key] :as multiaccount} port] [_ id]]
+   (let [contact (or (when (= id public-key)
+                       multiaccount)
+                     (get contacts id))]
      (if (nil? contact)
        (image-server/get-identicons-uri port id)
        (multiaccounts/displayed-photo contact)))))
--- a/src/status_im/multiaccounts/core.cljs
+++ b/src/status_im/multiaccounts/core.cljs
@@ -197,8 +197,6 @@
                                  :params     [key-uid (clean-path path) ax ay bx by]
                                  ;; NOTE: In case of an error we can show a toast error
                                  :on-success #(re-frame/dispatch [::update-local-picture %])}]}
-              (multiaccounts.update/optimistic :images [{:url  path
-                                                         :type (name photo-quality-large)}])
               (bottom-sheet/hide-bottom-sheet))))
 
 (fx/defn save-profile-picture-from-url
--- a/src/status_im/ui/components/fast_image.cljs
+++ b/src/status_im/ui/components/fast_image.cljs
@@ -20,7 +20,8 @@
                                 :on-load (fn [e]
                                            (when-let [on-load (:on-load props)]
                                              (on-load e))
-                                           (reset! loaded? true))})
+                                           (reset! loaded? true)
+                                           (reset! error? false))})
        (when (or @error? (not @loaded?))
          [placeholder (:style props)
           (if @error?


src/status_im/subs/chat/chats.cljs
```diff
@@ -229,9 +229,9 @@
  :<- [:profile/multiaccount]
  :<- [:mediaserver/port]
  (fn [[contacts multiaccount port] [_ id]]
-   (let [contact (or (get contacts id)
-                     (when (= id (:public-key multiaccount))
-                       multiaccount))]
+   (let [contact (or (when (= id (:public-key multiaccount))
+                       multiaccount)
+                     (get contacts id))]
```
yqrashawn: Any particular reason changes the order of these two?

src/status_im/subs/multiaccount.cljs
```diff
@@ -207,14 +207,13 @@
...
-(defn- replace-multiaccount-image-uri [multiaccount port]
+(defn- replace-multiaccount-image-uri [{:keys [key-uid] :as multiaccount} port]
   (let [public-key (:public-key multiaccount)
         identicon (image-server/get-identicons-uri port public-key)
         multiaccount (assoc multiaccount :identicon identicon)
         images (:images multiaccount)
         images (reduce (fn [acc current]
-                         (let [key-uid (:keyUid current)
-                               image-name (:type current)
+                         (let [image-name (:type current)
```
yqrashawn: What's the logic behind this? I saw there are two images (same image with different size) with same `:keyUid` in multiaccount `:images`. Is it possible that there will be images with different `:keyUid`?

src/status_im/ui/screens/chat/photos.cljs
```diff
@@ -12,7 +12,8 @@
      [react/view {:style (style/photo-container size)}
       [fast-image/fast-image {:source              {:uri photo-path}
                               :style               (style/photo size)
-                              :accessibility-label (or accessibility-label :chat-icon)}]
+                              :accessibility-label (or accessibility-label :chat-icon)
+                              :fallback true}]
```
yqrashawn: Any reason adds `:fallback` here?

yqrashawn (commented):
Got few questions.

src/status_im/subs/chat/chats.cljs
qfrank: contacts may contains the public key from multiaccount, but contact may only contains field: identicon , not include field: images, we should prefer show image than identicon

src/status_im/subs/multiaccount.cljs
qfrank: for the old one `key-uid (:keyUid current)`, key-uid may be nil in some cases, i guess there must be some issues with `status-go` that it forget to return value of key-uid which will cause CASE 2 reported from #14236 . the same image with different type(thumbnail/large) share the same :keyUid

src/status_im/ui/screens/chat/photos.cljs
qfrank: i doubt that there r some issues exists with fast-image for the implementation on the ios side which cause CASE 1 reported from #14236 , after add this option, looks like it resolved CASE 1, need feedback from test team .

src/status_im/subs/chat/chats.cljs
yqrashawn: In case 1, all users' avatar are broken, `multiaccount` only contains current user info. I don't think this is the root cause of case 1. Also most of the `id`s will be in `contacts` since most users in group is not the current user. I don't think it's a good idea to change the order here.

qfrank: this change is not relate to case 1, it's relate to case 2

src/status_im/subs/multiaccount.cljs
yqrashawn: Then we better raise this in status-go channel or repo, since this might affect somewhere else as well.

qfrank: i can raise an issue for this later, but it's another issue, can we solve this issue first? :)

src/status_im/ui/screens/chat/photos.cljs
yqrashawn: image

Seems all the broken avatar are identicon, which is the majority of avatars. We better take a closer look into this, since multiple places are using `fast-image` and we do want it to work for default identicon for performance improvements

qfrank: i'm sure fast-image is the cause of case 1, but i'm not familiar with OC, could u help to check implementation of native? currently what i found is fast-image triggered the callback `on-error`, but there is not much useful info with the callback param @yqrashawn

src/status_im/subs/multiaccount.cljs
yqrashawn: I don't think it's another issue, this makes finding the root cause harder.

> the same image with different type(thumbnail/large) share the same :keyUid

If this is the case, won't this make `reduce` generate some invalid images URLs if there are more than one set of image in multiaccount `images`?

src/status_im/ui/screens/chat/photos.cljs
qfrank: one improvement would be remove the clock param for `get-identicons-uri` (status-im.utils.image-server/get-identicons-uri) , so identicon can be cached. i will try it later.

yqrashawn: I'd be happy to look into it.

qfrank: another improvement can be `get-account-image-uri`, we need to know when the account image get updated , just like implementation of `get-contact-image-uri`

src/status_im/subs/chat/chats.cljs
yqrashawn: > this change is not relate to case 1, it's relate to case 2

I don't think this affects case 2. I can get the messages screen rendered successfully with `(re-frame/reg-sub :chats/photo-path` commented.

src/status_im/subs/multiaccount.cljs
qfrank: no, they r valid, the url of images differed by imageName.

qfrank: i mean status-go didn't return value of key-uid is another issue that we need fix

> I don't think it's another issue, this makes finding the root cause harder.

yqrashawn: > no, they r valid, the url of images differed by imageName.

If same image with different size share the same `key-uid`, won't there will be URL with invalid `&keyUid=` once there are multiple images?

https://github.com/status-im/status-mobile/blob/09675826393928d44c5a41b41a6b17727f5f703b/src/status_im/utils/image_server.cljs#L21

src/status_im/subs/chat/chats.cljs
qfrank: sorry, i made a wrong expression, i think it's an improvement, if u edited your headicon then back to chat screen, u will c the latest headicon, otherwise u will c the old identicon @yqrashawn

src/status_im/ui/screens/chat/photos.cljs
qfrank: > another improvement is when fast-image failed load image (on-error), we may make it retry again at once

okay, this shouldn't be the option, as images loaded from localhost

VolodLytvynenko:
hi, @qfrank one additional defect is still reproduced related to IOS. It's reproducible in 70% of cases


**Problem:**
**ISSUE 1** Profile pictures disappear on an IOS device if the device comes back from the background mode

**Steps to reproduce**:
1. Install App on IOS as UserA
2. Create a new account
3. Set image for profile
4. Leave the profile page as opened
5. Go to the background mode
6. Wait at least 4-5 minutes in the background
7. Turn back from the background mode

https://user-images.githubusercontent.com/52490791/199556189-e18a180b-b2f6-49f6-b7da-58d5ded59e3a.mov

**Actual result:**

The image on the profile page disappeared
image

**Expected result:**
The image is still shown
**Logs:**

**OS**
- IOS version: 15.7.1. Model: iPhone 11 pro
[Status.log](https://github.com/status-im/status-mobile/files/9922776/Status.log)


[geth.log](https://github.com/status-im/status-mobile/files/9922758/geth.log)

src/status_im/subs/chat/chats.cljs
```diff
@@ -229,9 +229,9 @@
  :<- [:profile/multiaccount]
  :<- [:mediaserver/port]
  (fn [[contacts multiaccount port] [_ id]]
```
ibrkhalil: Can we destructure `public-key` and `id` ?

src/status_im/ui/screens/chat/photos.cljs
ibrkhalil: Align map

qfrank: thought `make lint-fix` would fix this ):

qfrank:
> hi, @qfrank one additional defect is still reproduced related to IOS. It's reproducible in 70% of cases **Env:**
> 
> * IOS only
> 
> **Problem:** **ISSUE 1** Profile pictures disappear on an IOS device if the device comes back from the background mode
> 
> **Steps to reproduce**:
> 
> 1. Install App on IOS as UserA
> 2. Create a new account
> 3. Set image for profile
> 4. Leave the profile page as opened
> 5. Go to the background mode
> 6. Wait at least 4-5 minutes in the background
> 7. Turn back from the background mode
> 
>  Untitled.mov 
> **Actual result:**
> 
> The image on the profile page disappeared image
> 
> **Expected result:** The image is still shown **Logs:** [Status.log](https://github.com/status-im/status-mobile/files/9922776/Status.log)
> 
> [geth.log](https://github.com/status-im/status-mobile/files/9922758/geth.log)

Hi @VladimrLitvinenko , i am not able to reproduce this issue based on my PR build version, any further information could u provide?

VladimrLitvinenko:
> version

pavloburykh:
Hey @qfrank @VladimrLitvinenko I've tried to reproduce the bug on my IOS 14.1 iPhoneX but I didn't manage to do it. 

I have an assumption that in case of @VladimrLitvinenko it may be somehow related to IOS version or device (he has IOS 15.7.1, iPhone 11 Pro). 

I propose currently not to block this PR and move it forward as soon as it passes review, as overall it LGTM. 

@VladimrLitvinenko can perform additional investigation on the found bug and if there will be any additional details we can always re-open the issue or log a new one.

flexsurfer (approved):

VolodLytvynenko:
hi @qfrank 
The issue with the image disappearing is reproduced. In this case on Android device also. And there is no clear time of how much the device should be in the background. In my case, I left the device in the background for 7-8 hours. Is this issue visible in the attached logs?

**ISSUE 1: The image disappeared on the message page after the device came back from the background**

**Steps to reproduce:**
1. Create a new account
2. Set image for profile
3. Open the chat page
4. Go to the background mode
5. Wait several hours
6. Turn back from the background

**Actual result:**
image

**Expected result:**
Image is loan

**OS:** Adnroid version: 9. Model: huawei p20 lite 
**Logs:**
[Status.log](https://github.com/status-im/status-mobile/files/9936515/Status.log)

qfrank:
> hi @qfrank The issue with the image disappearing is reproduced. In this case on Android device also. And there is no clear time of how much the device should be in the background. In my case, I left the device in the background for 7-8 hours. Is this issue visible in the attached logs?
> 
> **ISSUE 1: The image disappeared on the message page after the device came back from the background**
> 
> **Steps to reproduce:**
> 
> 1. Create a new account
> 2. Set image for profile
> 3. Open the chat page
> 4. Go to the background mode
> 5. Wait several hours
> 6. Turn back from the background
> 
> **Actual result:** image
> 
> **Expected result:** Image is loan
> 
> **OS:** Adnroid version: 9. Model: huawei p20 lite **Logs:** [Status.log](https://github.com/status-im/status-mobile/files/9936515/Status.log)

Not much useful info, i will try my best to c if i can reproduce

pavloburykh:
Hey @qfrank I've finally managed to replicate the issue on IOS 14.1, iPhone X after returning from background. So I can confirm that the issue is still reproducible. 

photo_2022-11-04 12 16 19

[geth.log](https://github.com/status-im/status-mobile/files/9937227/geth.log)
[Status.log](https://github.com/status-im/status-mobile/files/9937228/Status.log)

qoqobolo:
Hi @qfrank ! I've also tried to replicate the issues on the latest build `45b853f` and what we have now:

**Case 1** from the issue https://github.com/status-im/status-mobile/issues/14236 is reproducible on my iOS device and _sometimes_ on Android with these steps:

1. Create a new account
2. Join a public chat with chat history
3. Put the app in the background for a couple of minutes
4. Re-open the app

You may need to try it several times.

[geth.log](https://github.com/status-im/status-mobile/files/9937218/geth.log) (iOS)


Screenshot 2022-11-04 at 13 13 15

**Case 2** from the issue is reproducible for me constantly on Android 10.
**Steps:**
1. Open Profile > Press on the profile picture > Select a picture (it is not necessary to set a picture, you can just open the gallery and close it)
2. Navigate to the `Chats` screen

[geth.log](https://github.com/status-im/status-mobile/files/9937281/geth.log) (Android)


https://user-images.githubusercontent.com/67952253/199950321-ce4d3dcd-85fa-4442-8feb-260954ef5f53.mp4

Can you check again these cases please?

qfrank:
finally, found the cause i guessss, will make a solution next week :)

qfrank:
Hi @qoqobolo @pavloburykh @VladimrLitvinenko , could u try it again with the latest push, thanks

pavloburykh:
> Hi @qoqobolo @pavloburykh @VladimrLitvinenko , could u try it again with the latest push, thanks

Hey @qfrank ! Thank you for the fix. We will re-check and provide a feedback. 

Just a quick reminder regarding the process on the board: when PR is ready for testing (either after initial work or after some bug fixes), you can also move it from CONTRIBUTOR column back to E2E or REVIEW (if PR requires Review) column. QAs are monitoring PRs from E2E column and take it into test if it is approved and has request-manual-qa label. Also, for those PRs that return after bug fixes you can remove label Tested-Issues which is set by QAs before moving PR with bugs to CONTRIBUTOR column. Thank you!

VolodLytvynenko:
Hey @qfrank . Thank you for your work! PR is tested and ready for merge.

PR #14282, opened 2022-11-03
chore: lint against re-frame dispatch/subscribe in quo components

yqrashawn:
Add a script to lint against below strings in `src/quo` and `src/quo2`.

1. `re-frame/dispatch`
1. `re-frame/subscribe`
1. `rf/dispatch`
1. `rf/subscribe`
1. `>evt`
1. `<sub`

Script added into `make lint`

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -287,6 +287,7 @@ endif
 
 lint: export TARGET := clojure
 lint: ##@test Run code style checks
+	sh scripts/lint-re-frame-in-quo-components.sh && \
 	clj-kondo --config .clj-kondo/config.edn --cache false --lint src && \
 	TARGETS=$$(git diff --diff-filter=d --cached --name-only src && echo src) && \
 	clojure -Scp "$$CLASS_PATH" -m cljfmt.main check --indents indentation.edn $$TARGETS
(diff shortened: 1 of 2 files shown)


jakubgs (approved):
It would be better if we could somehow check this using the linter we already use, but if there's no such possibility it's fine by me. But check with actual mobile devs.

flexsurfer:
it should be rf/sub

yqrashawn:
Not quite understand, what should be rf/sub
Do you mean keep `rf/...` and remove `re-frame/...`?
@flexsurfer

flexsurfer:
`rf/sub` instead of `rf/subscribe`

yqrashawn:
> `rf/sub` instead of `rf/subscribe`

I added `rf/sub` and kept `rf/subscribe`. Is that ok?

ilmotta:
This is a good implementation for now, thanks @yqrashawn. The best option in the future will be to process the output from clj-kondo's analysis because it's more reliable. The data processing could be performed by [Babashka](https://github.com/babashka/babashka) as well, so it's Clojure all the way down.

> It would be better if we could somehow check this using the linter we already use, but if there's no such possibility it's fine by me. But check with actual mobile devs.

You're correct @jakubgs, there's such a way, but it might be a bit too much work for now. Although I find such work to be quite valuable, because it could be extended to catch other conventions we're trying to establish in the codebase that are not available directly from the default linter rules.

https://cljdoc.org/d/clj-kondo/clj-kondo/2022.06.22/doc/analysis-data

PR #14293, opened 2022-11-07
fix error: package com.facebook.react.bridge does not exist while bui…

qfrank:
sometimes `make run-android` may result in following error, this pr fixed this error, reference is [here](https://stackoverflow.com/questions/74345114/error-package-com-facebook-react-bridge-does-not-exist-while-building-react-nat)
```
/Users/mac/status-react/node_modules/@react-native-async-storage/async-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java:10: error: package javax.annotation does not exist
import javax.annotation.Nullable;
                       ^
/Users/mac/status-react/node_modules/@react-native-async-storage/async-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncLocalStorageUtil.java:21: error: package com.facebook.react.bridge does not exist
import com.facebook.react.bridge.ReadableArray;
                                ^
/Users/mac/status-react/node_modules/@react-native-async-storage/async-storage/android/src/main/java/com/reactnativecommunity/asyncstorage/ReactDatabaseSupplier.java:14: error: package com.facebook.common.logging does not exist
import com.facebook.common.logging.FLog;
                                  ^
...
```
https://gist.github.com/jakubgs/d06afd176bcec79b8c410446a1715ea6

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -42,6 +42,8 @@ subprojects {
     }
 }
 
+def REACT_NATIVE_VERSION = new File(['node', '--print',"JSON.parse(require('fs').readFileSync(require.resolve('react-native/package.json'), 'utf-8')).version"].execute(null, rootDir).text.trim())
+
 allprojects {
     beforeEvaluate {
         if (System.env.STATUS_GO_ANDROID_LIBDIR == null || System.env.STATUS_GO_ANDROID_LIBDIR == "") {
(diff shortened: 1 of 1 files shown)


jakubgs (changes requested):
I don't like just patching stuff with random snippets from stackoverflow instead of addressing the actual issue. Will need to try to reproduce this myself.

Also, please put long log outputs into Gist rather than creating a massive PR post.

jakubgs:
Okay, it appears this is a known issue, and there's already a fix out(@cammellos found it):

* https://github.com/facebook/react-native/issues/35210

Which we can include by using the new [0.63.5](https://github.com/facebook/react-native/releases/tag/v0.63.5) hotfix release that came out yesterday.

Here is a proper fix:

* https://github.com/status-im/status-mobile/pull/14302

PR #14294, opened 2022-11-07
Attempt to fix too long prefix for notification types

rasom:
Just a renaming of `constant/activity-center-notification-type...` prefix which is too long.  

Please let me know if we have some agreement that we actually have to name constants like this.

status: ready

### DIFF
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -1,6 +1,6 @@
 (ns status-im.activity-center.core
   (:require [re-frame.core :as rf]
-            [status-im.constants :as const]
+            [status-im.activity-center.notification-types :as types]
             [status-im.data-store.activities :as data-store.activities]
             [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.fx :as fx]
@@ -29,12 +29,12 @@
               (as-> acc $
                 (update-in $ [type :read :data] remove-notification)
                 (update-in $ [type :unread :data] remove-notification)
-                (update-in $ [const/activity-center-notification-type-no-type :read :data] remove-notification)
-                (update-in $ [const/activity-center-notification-type-no-type :unread :data] remove-notification)
+                (update-in $ [types/no-type :read :data] remove-notification)
+                (update-in $ [types/no-type :unread :data] remove-notification)
                 (if (or (:dismissed notification) (:accepted notification))
                   $
                   (-> $ (update-in [type filter-status :data] insert-and-sort)
-                      (update-in [const/activity-center-notification-type-no-type filter-status :data] insert-and-sort))))))
+                      (update-in [types/no-type filter-status :data] insert-and-sort))))))
           db-notifications
           new-notifications))
 
@@ -75,7 +75,7 @@
 
 (def defaults
   {:filter-status          :unread
-   :filter-type            const/activity-center-notification-type-no-type
+   :filter-type            types/no-type
    :notifications-per-page 10})
 
 (def start-or-end-cursor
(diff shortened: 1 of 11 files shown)


ilmotta (commented):
Thanks for implementing such changes @rasom. I just think the name of your branch is too short.

Now seriously ;) the only thing I would change is to rename the file `activity_center/notification_types.cljs` to `activity_center/constants.cljs`. I like the name you chosen, but I would prefer if we could stick to the same naming "convention" already in place, which is to name such files as `constants.cljs`.

See for example `chat/constants.cljs` or `mailserver/constants.cljs`.

rasom:
@ilmotta 
> rename the file activity_center/notification_types.cljs to activity_center/constants.cljs

if we do this, that's virtually changes nothing, because in namespaces which still require global `constants` ns we would use something like `activity-center-constants/notification-type-`. I really see no point in the change then.

Just copying how it was done in other namespaces doesn't mean it is a good fit for this case (no guarantee it was a good decision in the first place, but I did not check). Having it ugly but same is sometimes fine, but idk what we gain in this case.

Otherwise if you just want to emphasise that it is a namespace of constants we can have something like  `status-im.activity-center.constants.notification-types` or `status-im.constants.notification-types`, which will still point to `constants` but not each time when a specific symbol is used (in Clojure it actually makes sense to emphasis that symbol is not a constant of any kind, because... you know...)

So if we do need to have another `constants.cljs` file then better just to close this pull request.

ilmotta:
> @ilmotta
> 
> > rename the file activity_center/notification_types.cljs to activity_center/constants.cljs
> 
> if we do this, that's virtually changes nothing, because in namespaces which still require global `constants` ns we would use something like `activity-center-constants/notification-type-`. I really see no point in the change then.
> 
> Just copying how it was done in other namespaces doesn't mean it is a good fit for this case (no guarantee it was a good decision in the first place, but I did not check). Having it ugly but same is sometimes fine, but idk what we gain in this case.
> 
> Otherwise if you just want to emphasise that it is a namespace of constants we can have something like `status-im.activity-center.constants.notification-types` or `status-im.constants.notification-types`, which will still point to `constants` but not each time when a specific symbol is used (in Clojure it actually makes sense to emphasis that symbol is not a constant of any kind, because... you know...)
> 
> So if we do need to have another `constants.cljs` file then better just to close this pull request.

@rasom, I think I get what you mean, thanks for the explanation. I agree with you it would become pointless :+1: 

> Just copying how it was done in other namespaces doesn't mean it is a good fit for this case (no guarantee it was a good decision in the first place, but I did not check).

I wasn't trying to judge too much the past decisions if they were good/bad, but now that you mentioned, there's something about all these const declarations split in many namespaces that I find excessively non-Clojure. It's like we're coding in Go or Java to me. I don't want to divert from the point of this PR, but one idea that I find very promising is to use macros to expand the access to global, immutable types, thus having zero runtime cost. I'm not gonna try to expand on this here because it'd be hard to explain some nuances. I might give this idea a try and share the code with you to see what you think.

But please, move on with the merge, I don't want to block this in any way.

ilmotta (approved):

rasom:
@ilmotta 
that's either separate namespaces or names with prefixes which contain entire history of the universe for sake of distinguishability

ilmotta:
> @ilmotta that's either separate namespaces or names with prefixes which contain entire history of the universe for sake of distinguishability

Not really, there are interesting ways around this with a Lisp such as Clojure. We're not treating these constants as data, we're treating them as global variables. As I said, I have an idea which could be good, but it might be a mess. I'll let you know once I see it in action in my REPL. Not for today for sure!

PR #14301, opened 2022-11-08
clean up quo2

flexsurfer:
clean up quo2

new root react-native folder
no quo usage in quo2
no react-native implementation in quo2


src/react_native/core.cljs
```diff
@@ -0,0 +1,29 @@
+(ns react-native.core
```
ilmotta, line 1: @flexsurfer, I think we had agreed we'd avoid using the name of the framework because we might actually want to add more _foundational_ things in this directory. I think @J-Son89 had his own impressions too about this.

flexsurfer: yeah we could rename it later if needed, now naming fit so well, so i would like to keep it like that, because i not sure there will be something else

ilmotta: I agree, seems good enough to me as well.

ilmotta (approved):

src/react_native/core.cljs
J-Son89: I thought we had agreed on libs 🤔  Anyway, we can always nest it later. The idea being that core components/functionality can live there, stuff which is used in both quo2 and status_im 👍

PR #14308, opened 2022-11-09
cleanup quo2 2

flexsurfer:
cleanup quo2, remove `status-im` namespace usage

create utils root namespace
move preview screens to status-im namespace


OmarBasem (approved):
All looks good to me. Just one question, why do we have a `utils.cljs` file outside of utils directory?

flexsurfer:
> All looks good to me. Just one question, why do we have a `utils.cljs` file outside of utils directory?

because it's like general js utils, and they are used in quo as well, but we don't want to use status-im in quo namespace

src/quo2/components/buttons/dynamic_button.cljs
```diff
@@ -93,8 +92,8 @@
                                                 :notification-up   0
                                                 :search-with-label 8)}}
            (case type
-             :jump-to                                       (i18n/label :t/jump-to)
-             :search-with-label                             (i18n/label :t/back)
+             :jump-to                                       label
+             :search-with-label                             label
```
Parveshdhull, line 96: Please can you elaborate, why this change required.

src/quo2/components/switcher/switcher_cards.cljs
```diff
@@ -1,27 +1,30 @@
...
             [quo2.components.tags.status-tags :as status-tags]
-            [status-im.ui.components.fast-image :as fast-image]
             [quo2.components.avatars.user-avatar :as user-avatar]
             [quo2.components.avatars.group-avatar :as group-avatar]
             [quo2.components.list-items.preview-list :as preview-list]
-            [quo2.components.avatars.channel-avatar :as channel-avatar]))
+            [quo2.components.avatars.channel-avatar :as channel-avatar]
+            [status-im.i18n.i18n :as i18n]))
+
+;;TODO move outside quo or find a way how to move i18n outside
```
Parveshdhull, line 16: moving outside quo2 will break the Figma structure we are following for components?
`quo2/componenet_directory/component`

Parveshdhull (approved):
apart from above comments, looks good

J-Son89:
> > All looks good to me. Just one question, why do we have a `utils.cljs` file outside of utils directory?
> 
> because it's like general js utils, and they are used in quo as well, but we don't want to use status-im in quo namespace

I think it's beyond the scope of this work but probably at some point we should look to split that utils file up into pieces that are more specific to each ones functionality.

J-Son89:
Great work 😎

ilmotta (approved):

ilmotta (commented):
Thanks for moving forward with these refactorings @flexsurfer

jo-mut (approved):

src/quo2/components/buttons/dynamic_button.cljs
flexsurfer: good question, i wanted quo library as pure as possible, so it shouldn't depend on status translation, so we need to pass labels as parameters to component, but there might be difficulties for sure, but at least we should try, probably there is better implementation for dynamic button, but my goal was just to move i18n outside in this PR

Parveshdhull: cool, sounds good. Thank you

Issue #14299, opened 2022-11-08
New Project Structure Document


OmarBasem:
An issue to track the document showing the new project structure we agreed on.

Issue #14298, opened 2022-11-08
Conventions Document


OmarBasem:
An issue to track the creation of the conventions document we agreed on in Istanbul.

PR #14128, opened 2022-10-10
style: format code with zprint

yqrashawn:
- [x] add config for zprint
- [ ] make sure everybody is fine with the rule
- [ ] ~format all cljx/edn file~ **(won't format all the files in this PR, relies on CI and `make lint` to gradually format changed files in each commit, the only formatted files in the end in the PR will be `shadow-cljs.edn` and `.zprintrc`)**
- [x] update `lint` `lint-fix` in `Makefile`

# Current Rule
1. based on [community rule](https://github.com/bbatsov/clojure-style-guide), [zprint doc about community style](https://github.com/kkinnear/zprint/blob/main/doc/options/community.md)
2. maximum width (column) **~80~ 105**
3. no optional comma in map
4. justify(vertically align) kvs in map
5. justify `:as`, `:refer`... in ns `:require` form
6. respect newlines, have some flexibility for formatting the code manually
7. respect blank lines
8. `clojure-lsp clean-ns` to sort `:require` and `:import` in `ns` form (alphabetically)

# Known issues
1. might require multiple times of format to get to the “end formatted state” (will report to upstream) https://github.com/kkinnear/zprint/issues/271
2. ~combine justify and respect new line in `:require` from cause syntax error (will report to upstream)~ no need to respect-nl in ns form
3. ~can't use helper fns like [`rodguide`](https://github.com/kkinnear/zprint/blob/2da480dda16a683846f57df0941f70a5ae1000d1/src/zprint/guide.cljc#L34) in `.zprintrc` when nix-shell has `TARGET=clojure`, works fine when `TARGET=default` (working on this one)~

# Editor/IDE support
- Already included in `make lint-fix`
- Has prebuilt binary for macOS and Linux, also can be used in form or uberjar or library in [REPL](https://github.com/kkinnear/zprint/blob/main/doc/using/repl.md) https://github.com/kkinnear/zprint#get-zprint
- zprint editors doc https://github.com/kkinnear/zprint/blob/main/doc/using/editor.md including Vim Emacs Sublime Text and Atom
- Make sure there's a `$HOME/.zprintrc` or `$HOME/.zprint.edn` file with `:search-config? true` or `:cwd-zprintrc?  true` or both in it. eg
```clojure
;; ~/.zprintrc
{:search-config? true}
```
## Emacs(tested)
Use package like https://github.com/radian-software/apheleia to config formatter for Clojure files

## VS Code(tested) 
[...]

Issue #13996, opened 2022-09-13
Lack of consistency on how re-frame events are named


ilmotta:
## Problem

As years went by, the Status Mobile codebase accumulated many different flavors of naming schemes for re-frame ****events****. There are events using fully qualified keywords, namespaced keywords, non-namespaced keywords, events starting with nouns, sometimes with verbs, and so on.

- PRs have more back and forth discussions about how events should be named.
- Some events have too common names, thus polluting the global registry with names that could be easily reused by mistake in other parts of the app, thus causing silent bugs. For example: `:close-chat`.
- Many events use Fully Qualified Keywords, which is known to make file renames more brittle and subject to bugs. Tools like `lsp-clojure` can help, but not everyone uses it.
- External contributors have a harder time knowing how to name events. This increases the cost for first contribution.

### Out of scope

This issue doesn't care about how all the many events are organized in the codebase. Even though this too is extremely important, it only cares about a proper naming convention that works for the Status Mobile team.

### Things we should keep in mind

1.  Some events are respecting the namespace they're part of. For example `src/status_im/multiaccounts/recover/core.cljs` defines `:multiaccounts.recover/re-encrypt-pressed`. Sometimes the keyword needs to broken down a bit further (see point 2).

2.  It's common to have event namespaces responsible for more than one *thing*, in practice, this means there's often a need to add a noun to the keyword, to make it clear what we're referring to in the same namespace. For instance, `status-im.bootnodes.core` namespace defines the events `:bootnodes.callback/qr-code-scanned` and `:bootnodes.ui/delete-pressed`, hence we can clearly see `bootnode` and `ui` are two different *things*.

3.  There's a reasonably well known article about some of the issues of fully qualifying keywords. Fully qualified keywords can also be problematic for consumers of the namespace, as we can see from Clojure's 1.11 release, which added the option `:as-alias` (similar to `:as` in `require`) to help with circular dependencies. <https://vvvvalvalval.github.io/posts/clojure-key-namespacing-convention-considered-harmful.html>

4.  Events can use different verb conjugations (e.g. past tense or present tense). For example, `:status-im.add-new.core/qr-code-scanned`, when successful, dispatches `:status-im.add-new.core/qr-code-handled`. There's always a way to convert between different conjugations though, so it's a matter of establishing a convention.

[...]

PR #14309, opened 2022-11-09
Document new guidelines

ilmotta:
Related to:

- https://github.com/status-im/status-mobile/issues/13996
- https://github.com/status-im/status-mobile/issues/14299
- https://github.com/status-im/status-mobile/issues/14298
- https://github.com/status-im/status-mobile/pull/14282
- https://github.com/status-im/status-mobile/pull/14128

### Summary

This PR simply adds a new document, soon to become the canonical documentation for guidelines in the Status Mobile repository.

Read the document rendered by Github here: https://github.com/status-im/status-mobile/blob/chore/document-conventions-and-guidelines/doc/new-guidelines.md

### Review notes

The documentation introduced in this PR is incomplete by its very open nature. I think it's better for the team to have something we agree on and we can continuously improve, than try to put too much upfront effort in making it perfect. But still, please **share any points you find are important enough to be documented** and I'll try to make them pretty :beach_umbrella: 

I've decided not to touch other documents in the repository because many of the recommendations in the new guidelines are not implemented yet, which could confuse contributors. That's why I named the new doc `doc/new-guidelines.md`, but as we continue to improve it, I think we can update/delete the old ones.

status: ready

### DIFF
--- a/doc/new-guidelines.md
+++ b/doc/new-guidelines.md
@@ -303,25 +303,25 @@ src
 ├── quo2
 │   ├── components/
 │   ├── foundations/
-│   ├── screens/
 │   └── theme.cljs
-└── status_im
-    ├── common
-    │   ├── components
-    │   │   └── bottom_sheet.cljs
-    │   ├── utils/
-    │   └── utils.cljs
-    ├── contexts/
-    ├── react_native
-    │   ├── gesture.cljs
-    │   └── platform.cljs
-    ├── setup/
-    └── subs/
+├── react_native
+│   ├── gesture.cljs
+│   └── platform.cljs
+├── status_im
+│   ├── common
+│   │   └── components
+│   │       └── bottom_sheet.cljs
+│   ├── contexts/
+│   ├── setup/
+│   └── subs/
+└── utils.cljs
 ```
 
 - `src/js`: Raw Javascript files, e.g. React Native Reanimated worklets.
 - `src/mocks`: Plumbing configuration to be able to run tests.
 - `src/quo2/`: The component library for Status Mobile.
+- `src/react_native/`: Contains only low-level constructs to help React Native
+  work in tandem with Clojure(Script).
 - `src/status_im/common/`: Directories named `common` can appear at any level of
   the directory tree. Just like directories named `utils`, their directory
   nesting level communicates their applicable limits.
@@ -330,8 +330,6 @@ src
 - `src/status_im/contexts/`: Contains [bounded contexts](#glossary), like
   `browser/`, `messaging/`, etc. As much as possible, _bounded contexts_ should
   not directly require each other's namespaces.
-- `src/status_im/react_native/`: Contains only low-level constructs to help
-  React Native work in tandem with Clojure(Script).
 - `src/status_im/setup/`: Contains namespaces that are mostly used to initialize
   the application, configure test runners, etc. In general, such namespaces
   should not be required from the outside.
--- a/doc/new-guidelines.md
+++ b/doc/new-guidelines.md
@@ -294,51 +294,68 @@ keywords and concatenating them into a single string.
 
 ## Project Structure
 
-As the mobile application grows in size and complexity, new patterns emerge to
-organize the codebase. The structure presented below is just an example of what
-the codebase could look like, so not all pieces are presented here.
+First, the bird's-eye view with some example ClojureScript files:
 
 ```
 src
-├── common
-│  └── bottom_sheet.cljs
-├── contexts
-│   ├── browser/
-│   ├── communities
-│   │   ├── components/
-│   │   ├── screens
-│   │   │   └── __test__
-│   │   └── utils
-│   │       └── example.cljs
-│   ├── messaging/
-│   ├── user_settings/
-│   └── wallet/
+├── js/
+├── mocks/
 ├── quo2
 │   ├── components/
 │   ├── foundations/
 │   ├── screens/
 │   └── theme.cljs
-├── react_native
-│   ├── gesture.cljs
-│   └── platform.cljs
-├── setup
-│   ├── core.cljs
-│   ├── events.cljs
-│   ├── navigation.cljs
-│   ├── reloader.cljs
-│   └── test_runner.cljs
-├── subs
-│   ├── ens.cljs
-│   └── pairing.cljs
-└── utils
-    ├── number.cljs
-    └── test_helpers.cljs
+└── status_im
+    ├── common
+    │   ├── components
+    │   │   └── bottom_sheet.cljs
+    │   ├── utils/
+    │   └── utils.cljs
+    ├── contexts/
+    ├── react_native
+    │   ├── gesture.cljs
+    │   └── platform.cljs
+    ├── setup/
+    └── subs/
 ```
 
-### Example structure for a quo2 component
+- `src/js`: Raw Javascript files, e.g. React Native Reanimated worklets.
+- `src/mocks`: Plumbing configuration to be able to run tests.
+- `src/quo2/`: The component library for Status Mobile.
+- `src/status_im/common/`: Directories named `common` can appear at any level of
+  the directory tree. Just like directories named `utils`, their directory
+  nesting level communicates their applicable limits.
+- `src/status_im/common/components/`: Contains reusable components that are not
+  part of the design system (quo2).
+- `src/status_im/contexts/`: Contains [bounded contexts](#glossary), like
+  `browser/`, `messaging/`, etc. As much as possible, _bounded contexts_ should
+  not directly require each other's namespaces.
+- `src/status_im/react_native/`: Contains only low-level constructs to help
+  React Native work in tandem with Clojure(Script).
+- `src/status_im/setup/`: Contains namespaces that are mostly used to initialize
+  the application, configure test runners, etc. In general, such namespaces
+  should not be required from the outside.
+- `src/status_im/subs/`: All subscriptions should live inside it.
 
-Notice the `test.cljs` file does not live inside a `__tests__/` directory, as
-this would be overkill in the quo2 component library.
+Directories named `utils/` can appear at any level of the directory tree. The
+directory nesting level precisely indicates its boundaries. For example, a
+`contexts/user_settings/utils/datetime.cljs` file communicates that it should
+only be used in the `user_settings` context.
+
+### src/quo2
+
+The `src/quo2/` directory holds all components for the new design system. As
+much as possible, its sub-directories and component names should reflect the
+same language used by designers.
+
+Even though the directory lives alongside the rest of the codebase, we should
+think of it as an external entity that abstracts away particular Status domain
+knowledge.
+
+Components inside `src/quo2/` should not rely on re-frame, i.e. they should not
+dispatch events or use subscriptions.
+
+Example structure:
 
 ```
 src
@@ -353,50 +370,10 @@ src
             └── view.cljs
 ```
 
-### Rationale
+### Re-frame events
 
-- Bringing ideas from [Domain-Driven
-  Design](https://en.wikipedia.org/wiki/Domain-driven_design), we want to
-  establish a clean separation of concerns between different [bounded
-  contexts](https://martinfowler.com/bliki/BoundedContext.html). Directories
-  inside `src/contexts/` should be named after particular Status domains.
-
-- Directories named `utils` can appear at any level of the directory tree. The
-  directory nesting level precisely indicates its boundaries. For example, a
-  `user_settings/utils/datetime.cljs` namespace communicates that the `datetime`
-  namespace should only be used in the `user_settings` context.
-
-- Directories named `common` can appear at any level of the directory tree. Just
-  like directories named `utils`, their directory nesting level communicates
-  their applicable limits.
-
-- Bounded contexts should not directly require each other's namespaces. We don't
-  want to make this a hard rule yet, but it's important to respect it as much as
-  possible. For example, namespaces inside `messaging` should not require
-  namespaces in the `browser` bounded context. In reality, different domains
-  should talk to each other using an interface of some kind.
-
-- The `src/setup/` directory contains namespaces that are mostly used to
-  initialize the application, configure test runners, etc. In general, such
-  namespaces should not be required from the outside.
-
-- The `src/react_native/` directory should contain only low-level constructs to
-  help React Native work in tandem with Clojure(Script).
-
-- The `src/quo2` directory holds all components for the new design system. As
-  much as possible, its sub-directories and component names should reflect the
-  same language used by designers. Even though the directory lives alongside the
-  rest of the codebase, we should think of it as an external entity that
-  abstracts away particular Status domain knowledge. Additionally, components
-  inside `quo2` should not rely on re-frame, i.e. they should not dispatch
-  events or use subscriptions.
-
-#### Re-frame
-
-- Subscriptions should live inside the top-level `src/subs/` directory.
-
-- Re-frame event handlers should be defined in files named `events.cljs`, and
-  they should be _close_ to other _things_, like view files, components, etc.
+Event handlers should be defined in files named `events.cljs`, and they should
+be _close_ to other _things_, like view files, components, etc.
 
 For example:
 
@@ -413,7 +390,7 @@ src
         └── view.cljs
 ```
 
-#### Tests
+### Test structure
 
 [Unit tests](#glossary) should be created alongside their respective source
 implementation. We prefer them colocated with the source and not like most
@@ -425,7 +402,7 @@ Component tests should be created in a separate directory `__tests__`. Example:
 │   ├── message.cljs
 │   └── message_test.cljs
 ├── __tests__
-│   └── component_bla_test.cljs
+│   └── input_bla_test.cljs
 ├── models.cljs
 └── models_test.cljs
 ```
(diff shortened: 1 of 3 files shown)


doc/new-guidelines.md
```diff
@@ -0,0 +1,467 @@
...
+Pay special attention to:
+
+- Align let bindings https://github.com/bbatsov/clojure-style-guide#bindings-alignment
+- Align map keys https://github.com/bbatsov/clojure-style-guide#map-keys-alignment
+
+## Dos and don'ts
+
+### Component styles
+
+Prefer to define styles in separate files named `styles.cljs`, usually colocated
```
OmarBasem: I believe we agreed on `style.cljs`

doc/new-guidelines.md
```diff
@@ -0,0 +1,467 @@
...
+   (:profile-pictures-visibility multiaccount)))
+
+;; good
+;; Notice the usage of a fake subnamespace called profile.
+(re-frame/reg-sub
+ :multiaccount.profile/pictures-visibility
+ :<- [:multiaccount]
+ (fn [multiaccount]
+   (:profile-pictures-visibility multiaccount)))
+```
```
OmarBasem: I think you meant bad -> good -> better

doc/new-guidelines.md
```diff
@@ -0,0 +1,467 @@
...
+
+```clojure
+(require '[quo2.components.icon :as icons])
+
+;; bad
+(icons/icon :main-icons2/verified)
+
+;; good
+(icons/icon :i/verified)
+```
```
OmarBasem: I think better to change it here to `quo2/icon`?

doc/new-guidelines.md
```diff
@@ -0,0 +1,467 @@
...
+;;   "biometric-auth-error": "Unable perform biometric authentication"
+(str (i18n/label :t/biometric-auth-error) "(" error-code ")")
+
+;; good
+;; Assume the translation key is:
+;;   "biometric-auth-error": "Unable perform biometric authentication ({{code}})"
+(i18n/label :t/biometric-auth-error {:code error-code})
+```
+
+## Project Structure
```
OmarBasem, line 299: We can add an example for what a single UI component/screen namespace should look like, as I showed you earlier, where we have the `component_name`, and under it we have 3 files: `view.cljs`, `style.cljs` and `test.cljs`

OmarBasem (commented):
Great work @ilmotta 🚀🚀

doc/new-guidelines.md
ilmotta: Thanks, your suggestion is clearer

doc/new-guidelines.md
ilmotta: What do you think about this example?

```
src
└── quo2
    ├── components
    │   └── dropdown
    │       ├── style.cljs
    │       ├── test.cljs
    │       └── view.cljs
    └── screens
        └── dropdown
            └── view.cljs
```

doc/new-guidelines.md
```diff
@@ -0,0 +1,468 @@
...
+- Align let bindings https://github.com/bbatsov/clojure-style-guide#bindings-alignment
+- Align map keys https://github.com/bbatsov/clojure-style-guide#map-keys-alignment
+
+## Dos and don'ts
+
+### Component styles
+
+Prefer to define styles in separate files named `style.cljs`, usually colocated
+with the source file using it. For an example, see
+[src/quo/components/controls/styles.cljs](../src/quo/components/controls/styles.cljs).
```
briansztamfater: Update `styles.cljs` to `style.cljs` here and in the usage example too

briansztamfater:
Impressive work, thanks for addressing this 🚀

doc/new-guidelines.md
ilmotta: @briansztamfater, but this is a link to the real file, and the file uses the plural form :shrug:, so unfortunately we can't use singular here otherwise the link would be broken.

briansztamfater: Actually, if it is a real example of our codebase and we decide to use `style.cljs`, we should update the example in our code to avoid confusions

ilmotta: We come back to the documentation and update it as soon as we have real examples using singular file names.

ilmotta: Actually there are style files named in the singular form. I'll update the docs to point to them.

ilmotta: Done, thanks!

doc/new-guidelines.md
```diff
@@ -0,0 +1,468 @@
...
+            :border-radius   4
+            :justify-content :center
+            :align-items     :center}}
+   [animated/view
+    (do-something)]])
+
+;; good
+(defn animated-checkbox-view [{:keys [size]}]
+  [animated/view
+   {:style (styles/animated-checkbox-style size)}
```
briansztamfater: should be `style/animated-checkbox-style` instead of `styles/...`

ilmotta: You sir are an eagle with laser focus eyes!

image

doc/new-guidelines.md
```diff
@@ -0,0 +1,468 @@
...
+;; good
+;; Use something people can more easily grep.
+;; TODO(2022-11-09,@someone): Refactor the world.
+```
+
+### Subscription names and event names
+
+Always register events and subscriptions using a meaningful namespace, but don't
+fully qualify them. We understand it's a controversial decision because there
+are both pros and cons to use FQK (fully-qualified keywords).
```
erikseppanen: I'm new to the term 'fully-qualified keywords' in clojure - I'm used to hearing them called 'namespaced keywords'.

ilmotta: It seems this is coming from my inner Clojure circles, but it's not well known outside of them. Qualifying a keyword means adding a namespace, which is a synonym for namespacing a keyword. The problem comes that sometimes we want to refer to qualifications that are valid and are expanded to the current namespace (e.g. `::whatever`).

Maybe there's no consensus on the community on how to name them, but I've seen people referring to FQKs as _namespaced keyword for the current namespace_, which is quite long...

I guess it's best if I change the text to be more explicit.

ilmotta: I've changed the text, please take a look and see if it's clear.

Pasted here for reference:

```
Always register events and subscriptions using a meaningful namespace, but don't
namespace them with `::`. We understand it's a controversial decision because
there are both pros and cons to such practice.

Whenever appropriate, it's also recommended to use _fake/invalid_ namespaces to
convey more knowledge in the keyword about which bounded context (domain) it
refers to, i.e. use dots to convey hierarchical structures.
```

doc/new-guidelines.md
```diff
@@ -0,0 +1,468 @@
...
+
+## Dos and don'ts
+
+### Component styles
+
+Prefer to define styles in separate files named `style.cljs`, usually colocated
+with the source file using it. For a real example, see
+[src/status_im/ui2/screens/chat/messages/style.cljs](../src/status_im/ui2/screens/chat/messages/style.cljs).
+Define styles as functions when you need to parameterize something, but prefer
+to define them as simple maps for a performance edge.
```
erikseppanen: I'd recommend leaving the reasoning out, like here in 'for a performance edge', mostly because it just adds more text to read... As a developer, I'm relieved when i can just see simple `good` `bad` with minimal explanation.

ilmotta: Fair enough, simplicity in the docs is quite important.

ilmotta: https://github.com/status-im/status-mobile/pull/14309/commits/b9fa37f3c021fe7798c15d8ee644f30215145df9

erikseppanen (approved):
I love this document... One step closer to world peace

OmarBasem (approved):

yqrashawn (approved):

ibrkhalil (approved):
Should we reference this files in the `README.md`

smohamedjavid (approved):

flexsurfer (approved):
if all are familiar with `contexts` I'm fine :)

flexsurfer:
great job @ilmotta

ilmotta:
> Should we reference this files in the `README.md`

We can for sure and maybe we should, I'm just afraid it could confuse other people given we still have a long way to go to implement these new guidelines.

Many files inside `doc/` are not linked in the README (including the existing `doc/codebase-structure-and-guidelines`), so I also assumed I shouldn't link this new markdown file to the README.

What do you think about we start using and evolving these new guidelines and then come back to update the existing documentation once the codebase has caught up with it?

OmarBasem:
> > Should we reference this files in the `README.md`
> 
> We can for sure and maybe we should, I'm just afraid it could confuse other people given we still have a long way to go to implement these new guidelines.
> 
> Many files inside `doc/` are not linked in the README (including the existing `doc/codebase-structure-and-guidelines`), so I also assumed I shouldn't link this new markdown file to the README.
> 
> What do you think about we start using and evolving these new guidelines and then come back to update the existing documentation once the codebase has caught up with it?

We have it mentioned that this is a work in progress: `This is a **work in progress**, and not all conventions are properly implemented in the codebase yet`. So I think it shouldn't confuse people.

ilmotta:
> if all are familiar with `contexts` I'm fine :)

It's a fair point to raise @flexsurfer. I think not everyone is familiar with Domain-Driven Design concepts such as bounded contexts. Although this is more common knowledge in backend circles, it's less used in frontend ones. Another similar word in meaning is `domains`, it's not exactly a bounded context, but it's perhaps better understood.

What do you think?

flexsurfer:
> > if all are familiar with `contexts` I'm fine :)
> 
> It's a fair point to raise @flexsurfer. I think not everyone is familiar with Domain-Driven Design concepts such as bounded contexts. Although this is more common knowledge in backend circles, it's less used in frontend ones. Another similar word in meaning is `domains`, it's not exactly a bounded context, but it's perhaps better understood.
> 
> What do you think?

in frontend usually, it's `modules` but i don't have a strong opinion

ilmotta:
> > > if all are familiar with `contexts` I'm fine :)
> > 
> > 
> > It's a fair point to raise @flexsurfer. I think not everyone is familiar with Domain-Driven Design concepts such as bounded contexts. Although this is more common knowledge in backend circles, it's less used in frontend ones. Another similar word in meaning is `domains`, it's not exactly a bounded context, but it's perhaps better understood.
> > What do you think?
> 
> in frontend usually, it's `modules` but i don't have a strong opinion

In any case, thanks for the comment @flexsurfer. I'm going to assume everyone seems to be fine with the choice because they approved the PR (as you said). In any case, it's never too late to change this if our future selves change our minds :currency_exchange:

siddarthkay (approved):
I love where this is going! Thanks for the discussions at the offsite and for capturing it on the repo  🚀

doc/new-guidelines.md
```diff
@@ -0,0 +1,454 @@
...
+- The `src/react_native/` directory should contain only low-level constructs to
+  help React Native work in tandem with Clojure(Script).
+
+- The `src/quo2` directory holds all components for the new design system. As
+  much as possible, its sub-directories and component names should reflect the
+  same language used by designers. Even though the directory lives alongside the
+  rest of the codebase, we should think of it as an external entity that
+  abstracts away particular Status domain knowledge. Additionally, components
+  inside `quo2` should no rely on re-frame, i.e. they should not dispatch events
+  or use subscriptions.
```
erikseppanen: should *not* rely on re-frame

qfrank (approved):
no words to say except **excellent** !

doc/new-guidelines.md
```diff
@@ -0,0 +1,454 @@
...
+(str (i18n/label :t/biometric-auth-error) "(" error-code ")")
+
+;; good
+;; Assume the translation key is:
+;;   "biometric-auth-error": "Unable perform biometric authentication ({{code}})"
+(i18n/label :t/biometric-auth-error {:code error-code})
+```
+
+## Project Structure
+
```
J-Son89, line 300: Can we consider giving a quick high level overview of the project structure first.
e.g src/ quo2, common, react-native status_im etc and then show that tree first. it's much smaller and easier to understand. Then we can describe what lives in each place and show the expanded trees of each if neccesary.

Something along the lines of:

As the mobile application grows in size and complexity, new patterns emerge to
organize the codebase. The structure presented below is just an example of what
the codebase could look like, so not all pieces are presented here.

At the root level the project structure is aligned into modules and applications.

```
src
├── common
├── status_im
│   
├── quo2
│   
└── utils
```

Common - where core react native dependencies and other code used across modules and applications. (IMO utils, common and react-native should all be in here. They are both code that is common to quo2 and status_im)

Quo2 - the component library for Status IM mobile

Status_IM - The code for Status IM mobile application
Separated by contexts ()

Utils - a variety of code for both testing, configuration and other aspects (probably should be in common) 


An example view of the tree should will look like the following:
```
src

├── status_im
│  └── contexts
│      ├── browser/
│      ├── communities
│      │   ├── components/
│      │   ├── screens/
│      │   └── utils
│      │       └── example.cljs
│      ├── messaging/
│      ├── user_settings/
│      └── wallet/
│  ├── setup
│      ├── core.cljs
│      ├── events.cljs
│      ├── navigation.cljs
│      ├── reloader.cljs
│      └── test_runner.cljs
│  ├── subs
│      ├── ens.cljs
│      └── pairing.cljs
├── quo2
│   ├── components/
│   ├── foundations/
│   ├── screens/
│   └── theme.cljs
├── common
│  └── bottom_sheet.cljs
├── react_native
│   ├── gesture.cljs
│   └── platform.cljs
└── utils
    ├── number.cljs
    └── test_helpers.cljs
```

doc/new-guidelines.md
```diff
@@ -0,0 +1,454 @@
...
+│   ├── ens.cljs
+│   └── pairing.cljs
+└── utils
+    ├── number.cljs
+    └── test_helpers.cljs
+```
+
+### Example structure for a quo2 component
+
+Notice the `test.cljs` file does not live inside a `__tests__/` directory, as
```
J-Son89: I found this part slightly confusing. I think we just remove `__tests__` part from this example and describe how we handle tests in a separate section

doc/new-guidelines.md
ilmotta: I think it's a nice suggestion @J-Son89, I can give that a try. Curiously, the fact that your examples don't agree with the new project structure in the doc tells me that it might not be clear to everyone, so again, just reinforcing your feedback that this part could be clearer.

doc/new-guidelines.md
ilmotta: Cool, I'm going to give it another spin @J-Son89. The hardest part is always explaining the project structure and the rationale, because the text easily grows in size and becomes confusing.

doc/new-guidelines.md
J-Son89: yeah, it was my understanding that it should be like this. imo the structure described in the doc is not correct.

most importantly we should have the `status_im` folder, which is missing at the moment in the document version

ilmotta: That's one good point, I don't think we need a status-im top level directory. In my perspective, everything that's not a vendored library is a status namespace. I understand the reasoning and the practices, but I only see it adding verbosity to namespaces for little practical value.

But having shared my opinion, I think I don't care much, if anyone feels strong about it I can totally change this right now for us.

ilmotta: This PR is only concerned with documenting our conventions, not really establishing them. I'd suggest we move the discussion to the chat. I'm going to update the docs in a separate commit and share in our channel to get votes/opinions.

J-Son89: "In my perspective, everything that's not a vendored library is a status namespace. I understand the reasoning and the practices, but I only see it adding verbosity to namespaces for little practical value."

My view on this is that we should aim to section off pieces of code to their functionality so that it is not being mixed together.

e.g if code in status_im application is being used within the quo2 library then that's an issue. The correct naming makes it easier for developers to understand what code should go where and confirm the agreements we as a team have made. Which it seems there is already some confusion on 😛 

I'm flexible either way but my opinion is that  in large projects it is better to be verbose.

ilmotta: I can live with that choice to put most things inside `src/status_im/`. Thanks for sharing your reasoning. I'm updating the docs taking your opinion in consideration and if the team agrees awesome!

ilmotta: FYI: The updated documentation is now reflecting some of your suggestions.

doc/new-guidelines.md
```diff
@@ -0,0 +1,438 @@
...
+First, the bird's-eye view with some example ClojureScript files:
+
+```
+src
+├── js/
+├── mocks/
+├── quo2
+│   ├── components/
+│   ├── foundations/
+│   ├── screens/
```
flexsurfer: screens/ moved to status-im/ui

doc/new-guidelines.md
```diff
@@ -0,0 +1,438 @@
...
+│   └── theme.cljs
+└── status_im
+    ├── common
+    │   ├── components
+    │   │   └── bottom_sheet.cljs
+    │   ├── utils/
+    │   └── utils.cljs
+    ├── contexts/
+    ├── react_native
```
flexsurfer: react_native should be on the same level with quo and status-im

J-Son89 (approved):
Great work putting this together! :D

rasom (approved):

PR #14326, opened 2022-11-10
Add display name

cammellos:
Adds display name anywhere we show the user name

this is connected to @briansztamfater work on author, so it might be superseeded by that, but some of the changes are valuable and might be kept from this PR.

### DIFF
--- a/src/status_im/multiaccounts/core_test.cljs
+++ b/src/status_im/multiaccounts/core_test.cljs
@@ -1,0 +1,85 @@
+(ns status-im.multiaccounts.core-test
+  (:require [cljs.test :refer-macros [deftest is testing]]
+            [status-im.multiaccounts.core :as ma]))
+
+(def public-key "0x0461f576da67dc0bca9888cdb4cb28c80285b756b324109da94a081585ed6f007cf00afede6b3ee5638593674fee100b590318fc7bdb0054b8dd9445acea216ad2")
+(def short-public-key "0x0461…ad2")
+(def random-name "Studious Gold Mustang")
+(def override-random-name (str "override" random-name))
+(def nickname "nickname")
+(def override-nickname (str nickname "override"))
+(def display-name "display-name")
+(def override-display-name (str display-name "override"))
+(def ens-name "jakubgs.eth")
+(def formatted-ens (str "@" ens-name))
+(def override-ens-name (str "override" ens-name))
+
+(def contact
+  {:nickname     nickname
+   :name         ens-name
+   :ens-verified true
+   :public-key   public-key
+   :display-name display-name})
+
+(deftest contact-two-names-test
+  (testing "names is nil"
+    (testing "nickname has precedence"
+      (is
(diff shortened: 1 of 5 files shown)


src/status_im/multiaccounts/core_test.cljs
```diff
@@ -0,0 +1,85 @@
+(ns status-im.multiaccounts.core-test
+  (:require [cljs.test :refer-macros [deftest is testing]]
+            [status-im.multiaccounts.core :as ma]))
+
+(def pk "0x0461f576da67dc0bca9888cdb4cb28c80285b756b324109da94a081585ed6f007cf00afede6b3ee5638593674fee100b590318fc7bdb0054b8dd9445acea216ad2")
```
ilmotta: I've seen many many people referring to PK as any of these 3 things: public key, private key or the public/private key pair. Could you rename this var to `public-key` or `pub-key` for extra clarity?

src/status_im/multiaccounts/core_test.cljs
```diff
@@ -0,0 +1,85 @@
...
+(def override-ens-name (str "override" ens-name))
+
+(def contact
+  {:nickname     nickname
+   :name         ens-name
+   :ens-verified true
+   :public-key   pk
+   :display-name display-name})
+
+(deftest contact-two-names
```
ilmotta: FYI: It's more correct to add the `-test` suffix to tests because the macro `deftest` will create a var with the name `contact-two-names` and add the macro body as metadata. Let's say, in more _busy_ namespaces, not adding the suffix can lead to `deftest` calls replacing functions previously defined in the same namespace, so that's why the general recommendation is to add the suffix. I wish the linter could catch this by default.

ilmotta (approved):

src/status_im/multiaccounts/core_test.cljs
```diff
@@ -0,0 +1,85 @@
...
+                   :display-name)
+           pk)))))
+  (testing "names is provided"
+    (let [names              {:nickname         override-nickname
+                              :display-name     override-display-name
+                              :three-words-name override-random-name
+                              :ens-name         override-ens-name}
+          contact-with-names (assoc contact :names names)]
+
+      (testing "nickname has precedence"
```
briansztamfater, line 64: I am not sure if ens name should have precedence based on the designs. We can modify this later if needed btw, just raising my doubt.

src/status_im/multiaccounts/core_test.cljs
briansztamfater: Maybe it is for another conversation but maybe it should be `chat-key` actually?
But I agree to refactor name to something more readable

briansztamfater (approved):

ibrkhalil (approved):

src/status_im/multiaccounts/core_test.cljs
cammellos: I'll double check with the design team, if that's the case I'll change it

PR #14329, opened 2022-11-11
Visual tests setup

### LINKED ISSUE #14068: Implement Visual Test mechanism and configuration setup
# Feature Issue

Reference this branch https://github.com/status-im/status-mobile/tree/jcaprani/visual-test

### User Story

As a developer I want to be able to create visual tests for the quo2 component library so that I can be confident I have not broke the ui on other components while refactoring implemented components.

### Description

Taking the work from this branch: 
https://github.com/status-im/status-mobile/tree/jcaprani/visual-test

Clean up and implement the base structure and configuration for creating visual tests.
This includes adding some dependencies to the codebase -> Detox, jest-snapshot-testing.

### Acceptance Criteria
Visual test tooling installed in codebase. 
A simple test can be run
Tool builds and runs without errors.

J-Son89:
Fixes: https://github.com/status-im/status-mobile/issues/14068

Paired with @erikseppanen in the offsite to set up the visual tests 

Adds configuration and first basic visual test to codebase.


.detoxrc.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
```
J-Son89, line 1: @jakubgs, I tried to move this out of root but it brought in some complications. I can try to move it again but would like to get this p,r in first as there are so many changes happening in the codebase.

jakubgs: Well, at least it's a hidden file.

package.json
```diff
@@ -11,6 +11,7 @@
     "app:watch": "shadow-cljs watch android",
     "app:packager": "react-native start --host 0.0.0.0 --port 8081",
     "app:android": "react-native run-android"
+
```
ibrkhalil: Unneeded change maybe ?

Makefile
```diff
@@ -313,6 +313,16 @@ test: ##@test Run tests once in NodeJS
 	yarn shadow-cljs compile test && \
 	node --require ./test-resources/override.js target/test/test.js
 
+
+run-visual-test: export TARGET := clojure
```
ilmotta: I think `run-visual-test` should be renamed to `run-visual-test-ios`, following the convention in the Makefile, because this target is not supposed to work in Linux environments.

src/quo/components/button/view.cljs
```diff
@@ -56,7 +56,7 @@
 
 (defn button [{:keys [on-press disabled type theme before after
                       haptic-feedback haptic-type on-long-press on-press-start
-                      accessibility-label loading border-radius style]
+                      accessibility-label loading border-radius style test-ID]
```
ilmotta, line 59: It's just my opinion, but it's weird to see uppercase names in Lisps, not just in Clojure. From that perspective, I think we should just follow this strong convention and always refer to test IDs as `test-id` in Clojure land. Hence, the plumbing components, like `rn/view` should rename it to interface with the Javascript land.

It's not something for this PR for sure, but a Github issue could come in handy ;)

src/quo2/__visual_tests__/button.e2e.js
```diff
@@ -0,0 +1,50 @@
...
+describe('Default Renders', () => {
+    beforeAll(async () => loginToHomePage())
+    beforeEach(async () => {
+    });
+    afterEach(async () => {
+        await element(by.id("back-button")).tap();
+        await waitToNavigate(200);
+    });
+
+
```
ilmotta: There are spurious newlines in this file. I'm already thinking we should have a Github issue to add prettier or another formatter to this project given that we are going to write more Javascript from now on. What do you think?

ilmotta (commented):
Awesome work over here! One thing I'm not seeing though is a little bit of documentation on these tests. We have `doc/TESTING.md` already that could be updated.

src/quo2/__visual_tests__/button.e2e.js
ibrkhalil: Wonderful idea

J-Son89:
> Awesome work over here! One thing I'm not seeing though is a little bit of documentation on these tests. We have `doc/TESTING.md` already that could be updated.

yep I intend to add docs, I want to get the p.r in first because then I avoid making double changes if people request updates etc or it's not working on some persons machine

src/quo2/__visual_tests__/button.e2e.js
J-Son89: yeah I would be happy to have all formatting automated :)

src/quo/components/button/view.cljs
J-Son89: yeah, this one I'm not entirely sure about because I tried to used :test-id but it doesn't work.. perhaps I need to do "test-i-d or something. In any case the discrepancy between the two is somewhat advantageous as we clearly see it is not a standard prop.
it is a bit sore on the eyes though 😆 so I would be up for changing it in follow up work

Makefile
J-Son89: yep makes sense 👍

src/quo/components/button/view.cljs
ilmotta: Yep, this is not necessarily difficult, but it's a reasonable amount of work that shouldn't be done in this PR. Ideally IMO, the plumbing layer (the functions that abstract React Native classes, like `rn/view`), should all translate Clojure idioms to RN, i.e. `:test-id` is perfectly valid and expected in Clojure, but it would be translated to `:testID` over there.

Leaking programming language idioms is usually something I'm worried about.

ilmotta (approved):

PR #14331, opened 2022-11-11
feat: add component tests using react-testing-library and jest

### LINKED ISSUE #14330: Add interactive unit tests (component-tests)
As a developer I want to be able to write unit tests that interact with the components that I have developed.

Description:
add unit that use Jest and React-Native-testing-library to the codebase so that the components in quo2 can have fast and easy tests written to help avoid the components from being broken,

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14330

run command `make component-test` to run these tests :)


J-Son89:
Screenshot 2022-11-11 at 07 02 21

jakubgs (approved):
Thanks for moving Jest config to a subfolder. But I'm only approving non-clojure changes.

src/status_im/i18n/i18n_resources.cljs
```diff
@@ -5,8 +5,11 @@
 ;;TODO (14/11/22 flexsurfer) this namespace has been moved to the status-im2 namespace, we keep this only for old (status 1.0) code,
 ;; can be removed with old code later
 
+(def get-keyword
+  (if js/global.__TEST__ str keyword))
```
flexsurfer, line 9: could you please elaborate on why this is needed? how its used

J-Son89: Yeah, perhaps it's avoidable but the test breaks without this. For some reason `keyword` function is not there when running these tests. Perhaps I need to adjust something in the shadow-cljs setup but I could not find a solution for it. 
Since these i18n pieces are kind of insignificant in these tests and mostly being mocked I figured this is not that much of a concern for now.

src/quo2/components/icons/icons.clj
```diff
@@ -4,9 +4,14 @@
 
 (def icon-path "./resources/images/icons2/")
 
+(defn combine-path [path el]
+  (if (System/getenv "__COMPONENT_TEST__")
```
J-Son89: similar change here @flexsurfer, where the resources are not found in the test setup so I had to adjust the path a bit for when using tests mode.

Makefile
```diff
@@ -313,7 +313,22 @@ test: ##@test Run tests once in NodeJS
...
+component-test: export BABEL_ENV := test
+component-test-watch: ##@ Watch tests and re-run no changes to cljs files
+	yarn install
+	nodemon --exec 'yarn shadow-cljs compile component-test && jest --config=test/jest/jest.config.js' -e cljs
+
+component-test: export TARGET := clojure
+component-test: export __COMPONENT_TEST__ := true
+component-test: export BABEL_ENV := test
+component-test: ##@test Run tests once in NodeJS
+	# Here we creates the gyp bindings for nodejs
```
siddarthkay: I am about to make a very lame comment,
brace for impact
.
.
fix grammar to :  "Here we create the gyp bindings for nodejs "

Makefile
```diff
@@ -313,7 +313,22 @@ test: ##@test Run tests once in NodeJS
 	yarn shadow-cljs compile test && \
 	node --require ./test-resources/override.js target/test/test.js
 
-#--------------
+component-test-watch: export TARGET := clojure
+component-test: export __COMPONENT_TEST__ := true
+component-test: export BABEL_ENV := test
+component-test-watch: ##@ Watch tests and re-run no changes to cljs files
```
ilmotta, line 327: I find the feedback cycle of the watcher a bit too slow (not much we can do anyway), but have you tried running the tests in a REPL?

shadow-cljs.edn
```diff
@@ -106,4 +106,13 @@
            :output-to "target/mocks/mocks.js"
            :output-dir "target/mocks"
            :compiler-options {:optimizations :simple
-                              :source-map false}}}}
+                              :source-map false}}
+          :component-test {:target :npm-module
+                           :entries [quo2.components.selectors.--tests--.selectors-component-spec]
```
ilmotta: Why do we have this namespace hardcoded here?

From the `:entries` doc:

> The namespaces that serve as the root nodes of the dependency graph for the output code of this module.

This namespace doesn't look like a root node. Perhaps we could use something like the following and create an entry point namespace. This has the added benefit of being REPL friendly instead of relying on the shadow-cljs shenanigans.

```clojure
:entries [test.components.core]
```

shadow-cljs.edn
```diff
@@ -106,4 +106,13 @@
...
+                           :ns-regexp "-component-spec$"
+                           :output-dir "component-spec"
+                           :closure-defines
+                           {quo2.components.icons.icons/__TEST__ true}
```
ilmotta: It's not obvious just glancing from this definition why it's here. Could you add a comment briefly explaining why this is required?

src/libs/test/jest/matchers.cljs
```diff
@@ -0,0 +1,36 @@
+(ns libs.test.jest.matchers)
+
+(js/expect.extend
+ #js{:toEq
+     (fn toEq [received expected]
```
ilmotta: [Small improvement] This could be a proper named function with a docstring.

src/status_im/i18n/i18n_resources.cljs
ilmotta: This is surprisingly weird given that there are so many other functions in the `cljs.core` namespace, so why would only `cljs.core/keyword` not be defined? At least I hope we don't have to do this kind of change in other namespaces in the future.

ilmotta: Let's just hope we don't have to infest other parts of the codebase with these workarounds. I'm super curious to understand why the `keyword` function is not defined, but the other `cljs.core` functions in this namespace are all properly defined.

shadow-cljs.edn
J-Son89: yep we could do that

shadow-cljs.edn
J-Son89: oh actually I think I can remove that, I was exploring options to add in environment variables

Makefile
J-Son89: I have not tried it in a REPL yet, it's something we should definitely explore

ilmotta: Cool, good to know :+1:

src/status_im/i18n/i18n_resources.cljs
J-Son89: yeah I could not figure that out either. I agree though we don't want to have a lot of parts of the codebase with code like this.

src/libs/test/jest/matchers.cljs
J-Son89: Yeah this would be nice to have. To be honest I took this code from here: https://github.com/pesterhazy/shadow-jest/blob/master/src/test/jest/matchers.cljs

I'm not sure whether we fully need it yet as it just adds some additional utility to jests expect function.
I would like to get this code merged in first and have devs writing tests and I can revisit this after wards.

I also need to add documentation on how to run/write these tests.

ilmotta (approved):

src/libs/test/jest/matchers.cljs
ilmotta: Sure, sounds good! It's just that it's a lot harder to take things away once merged, so if this matcher is not really needed we could add it only when we're certain of its value.

J-Son89: Yeah I think I prefer your approach. So I will remove it from this p.r.

PR #14332, opened 2022-11-11
Bottom sheet missing actions

### LINKED ISSUE #14328: Bottom sheet missing actions
An issue to track the creation of bottom sheet missing actions. Some actions are missing in 1-1 chats and group chats, and contact actions were not implemented.

OmarBasem:
Fixes #14328

This PR implements missing bottom sheet actions for 1-1 chats and group chats. It also implements actions for contact items. Some actions functionalities are implemented, while others require design input.

### Edit
This PR also implements the different actions for groups depending on a variety of crieteria (private or public, member or owner, home screen or chat screen)

### Designs
[View designs.](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=1972%3A460376)
[View designs.](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=1803%3A399979)

### DIFF
--- a/src/quo2/components/drawers/action_drawers.cljs
+++ b/src/quo2/components/drawers/action_drawers.cljs
@@ -1,8 +1,8 @@
 (ns quo2.components.drawers.action-drawers
-  (:require  [react-native.core :as rn]
-             [quo2.components.markdown.text :as text]
-             [quo2.components.icon :as icon]
-             [quo2.foundations.colors :as colors]))
+  (:require [react-native.core :as rn]
+            [quo2.components.markdown.text :as text]
+            [quo2.components.icon :as icon]
+            [quo2.foundations.colors :as colors]))
 
 (defn- get-icon-color [danger?]
   (if danger?
@@ -14,47 +14,48 @@
                       sub-label
                       right-icon
                       danger?
-                      on-press]}]
-  [rn/touchable-opacity {:on-press on-press}
-   [rn/view {:style
-             {:height            (if sub-label 56 47)
-              :margin-horizontal 20
-              :flex-direction    :row}}
-    [rn/view {:style
-              {:height        20
-               :margin-top    :auto
-               :margin-bottom :auto
-               :margin-right  12
-               :width         20}}
-     [icon/icon icon
-      {:color (get-icon-color danger?)
-       :size 20}]]
-    [rn/view
-     {:style
-      {:flex            1
-       :justify-content :center}}
-     [text/text
-      {:size   :paragraph-1
-       :weight :medium
-       :style  {:color
-                (when danger?
-                  (colors/theme-colors colors/danger-50 colors/danger-60))}}
-      label]
-     (when sub-label
-       [text/text
-        {:size  :paragraph-2
-         :style {:color
-                 (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
-        sub-label])]
-    (when right-icon
+                      on-press] :as item}]
+  (when item
+    [rn/touchable-opacity {:on-press on-press}
+     [rn/view {:style
+               {:height            (if sub-label 56 47)
+                :margin-horizontal 20
+                :flex-direction    :row}}
       [rn/view {:style
                 {:height        20
                  :margin-top    :auto
                  :margin-bottom :auto
-                 :width 20}}
-       [icon/icon right-icon
+                 :margin-right  12
+                 :width         20}}
+       [icon/icon icon
         {:color (get-icon-color danger?)
-         :size 20}]])]])
+         :size  20}]]
+      [rn/view
+       {:style
+        {:flex            1
+         :justify-content :center}}
+       [text/text
+        {:size   :paragraph-1
+         :weight :medium
+         :style  {:color
+                  (when danger?
+                    (colors/theme-colors colors/danger-50 colors/danger-60))}}
+        label]
+       (when sub-label
+         [text/text
+          {:size  :paragraph-2
+           :style {:color
+                   (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
+          sub-label])]
+      (when right-icon
+        [rn/view {:style
+                  {:height        20
+                   :margin-top    :auto
+                   :margin-bottom :auto
+                   :width         20}}
+         [icon/icon right-icon
+          {:color (get-icon-color danger?)
+           :size  20}]])]]))
 
 (defn divider []
   [rn/view {:style {:border-top-width 1
--- a/src/quo2/components/drawers/action_drawers.cljs
+++ b/src/quo2/components/drawers/action_drawers.cljs
@@ -14,8 +14,8 @@
                       sub-label
                       right-icon
                       danger?
-                      on-press] :as item}]
-  (when item
+                      on-press] :as action-props}]
+  (when action-props
     [rn/touchable-opacity {:on-press on-press}
      [rn/view {:style
                {:height            (if sub-label 56 47)
--- a/src/status_im/ui2/screens/chat/actions.cljs
+++ b/src/status_im/ui2/screens/chat/actions.cljs
@@ -18,7 +18,7 @@
    :on-press   on-press
    :danger?    danger?
    :sub-label  sub-label
-   :right-icon (when chevron? :main-icons2/chevron-right)})
+   :right-icon (when chevron? :i/chevron-right)})
 
 (defn hide-sheet-and-dispatch [event]
   (rf/dispatch [:bottom-sheet/hide])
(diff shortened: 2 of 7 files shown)


src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -4,45 +4,51 @@
    [status-im.chat.models.pin-message :as models.pin-message]
    [status-im.i18n.i18n :as i18n]
    [status-im.constants :as constants]
-   [status-im.utils.handlers :refer [<sub >evt]]
+   [status-im.utils.re-frame :as rf]
    [quo2.components.drawers.action-drawers :as drawer]))
 
-(defn- entry [icon label on-press danger?]
+(defn- entry [icon label on-press danger? sub-label chevron?]
```
J-Son89: personally prefer named props, particularly when a function has more than 2/3 props. 

@ilmotta, @flexsurfer, @erikseppanen  do we have any consensus on how we would like to go about this?

in this case it's basically
```
 (entry {
         :icon  :main-icons2/correct
         :label s(i18n/label :mark-as-read)
         :on-press #(mark-all-read-action chat-id)
         danger?: false
         sub-label?: nil
         chevron?: false
         }
         ))
   ```
   vs 
   ```
    (entry :main-icons2/correct
         (i18n/label :mark-as-read)
         #(mark-all-read-action chat-id)
         false
         nil
         false))
   ```
which in the 2nd version the `false`, `nil` `false` give me no idea what they're being used for. 

It also makes the function harder to re-use as you have to set `nil` as a value so you can set the next param.

File: src/quo2/components/drawers/action_drawers.cljs [hunk 2]
```diff
@@ -14,47 +14,48 @@
...
-                  (colors/theme-colors colors/danger-50 colors/danger-60))}}
-      label]
-     (when sub-label
-       [text/text
-        {:size  :paragraph-2
-         :style {:color
-                 (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
-        sub-label])]
-    (when right-icon
+                      on-press] :as item}]
```
J-Son89: is item too generic? perhaps `action-props`?

J-Son89 (approved):

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -52,66 +58,199 @@
...
+         false
+         nil
+         false))
+
+(defn notifications-entry []
+  (entry :main-icons2/notifications
+         (i18n/label :notifications)
+         #(js/alert "TODO: to be implemented, requires design input")
+         false
+         "All messages" ; placeholder
```
J-Son89: should "placeholder" be a TODO comment? 

or should "All messages" be an i18n label?

src/status_im/ui2/screens/chat/actions.cljs
jo-mut: i agree with @J-Son89. It improves code readability

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -4,45 +4,51 @@
    [status-im.chat.models.pin-message :as models.pin-message]
    [status-im.i18n.i18n :as i18n]
    [status-im.constants :as constants]
-   [status-im.utils.handlers :refer [<sub >evt]]
+   [status-im.utils.re-frame :as rf]
    [quo2.components.drawers.action-drawers :as drawer]))
 
-(defn- entry [icon label on-press danger?]
```
jo-mut, line 10: since this is an action drawer maybe you could just call it `action` instead of `entry` although not that much important its not that much straightforward what an entry is

src/status_im/ui2/screens/chat/actions.cljs
OmarBasem: cc @cammellos

OmarBasem: I agree @J-Son89. This code was initially written by Andrea, what are your thoughts @cammellos?

cammellos: I agree, maps much more readable

src/status_im/ui2/screens/chat/components/message_home_item/view.cljs
```diff
@@ -129,7 +129,7 @@
          (get-in last-message [:content :text])]
         [render-subheader (get-in last-message [:content :parsed-text])])]
      (if (> unviewed-mentions-count 0)
-       [info-count unviewed-mentions-count {:top 16}]
+       [info-count/info-count unviewed-mentions-count {:top 16}]
```
flexsurfer: now we have quo core and it should be used instead i believe, like `quo/info-count`

src/status_im/ui2/screens/chat/actions.cljs
ilmotta: Agree @J-Son89 for this case. There's a general recommendation for all programming languages that 4+ arguments **could be an indicative** that a custom type or map (in Clojure's case) should be used, but I don't think we should try to make this a rule. Some library functions I've used receive 4 arguments without a map and are a joy to use, so it all depends IMO. Function composition is also very important in some scenarios, and maps make function composition more awkward due to the need to destructure.

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -52,66 +58,199 @@
...
+  (entry :main-icons2/save
+         (i18n/label :fetch-messages)
+         #(js/alert "TODO: to be implemented, requires design input")
+         false
+         nil
+         true))
+
+(defn pinned-messages-entry []
+  (entry :main-icons2/pin
+         (i18n/label :pinned-messages)
```
ilmotta: There are a couple translations missing the `:t/` qualification.

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
```diff
@@ -1,29 +1,32 @@
...
+  (rf/dispatch [:accept-all-activity-center-notifications-from-chat chat-id]))
 
 (defn contact-item [item]
   (let [{:keys [public-key ens-verified added? images]} item
-        display-name (first (<sub [:contacts/contact-two-names-by-identity public-key]))
-        photo-path   (if-not (empty? images) (<sub [:chats/photo-path public-key]) nil)]
-    [rn/touchable-opacity (merge {:style   (style/container)
-                                  :on-press #(open-chat public-key)})
+        display-name (first (rf/sub [:contacts/contact-two-names-by-identity public-key]))
+        photo-path   (if-not (empty? images) (rf/sub [:chats/photo-path public-key]) nil)]
```
ilmotta: Since you touched this line, you can also replace `if-not` for `when-not`. `(if-not x something-true nil)` is never necessary in Clojure. Or, better yet, if `images` is a sequence:

```clojure
(when (seq images) (rf/sub [:chats/photo-path public-key]))
```

ilmotta (approved):

src/status_im/ui2/screens/chat/actions.cljs
OmarBasem: I saw they were not used in this file, so I did so. Curious about what difference it makes?

src/status_im/ui2/screens/chat/actions.cljs
OmarBasem: The word `action` is being used to define the on-press method names. So, for example, we have `edit-nickname-entry` and `edit-nickname-action`

jo-mut (approved):

src/status_im/ui2/screens/chat/actions.cljs
ilmotta: Thanks for asking @OmarBasem.

It's the same reasoning as to why we use `:i/<name-of-the-icon>`. It helps create a valid list of translation keywords in the codebase. It also helps to precisely grep for translations, which is sometimes important because some of them use very common words, like `admin` (e.g. renaming translations becomes trivial and reliable). Editors can easily autocomplete only translation keywords once you type `:t/`. In the future I bet we'll have a linter rule that if the function `i18n/label` is used, then the first argument must be qualified with `:t`.

The complete reasoning behind this convention is not in our docs, otherwise it would become too big if we explain everything in detail.

OmarBasem: Aha okay, thanks for the explanation. So the Clojure language ignores these qualifications? (Initially, I was expecting it not to work without the qualification)

ilmotta: It all depends on what we do with the namespaced keyword. For example, if the function `i18n/label` had a `:pre` condition check that verified the namespace was indeed `:t`, then you'd get runtime errors signaling this problem. But if, on the other hand, the i18n/label simply uses the core `name` function, then the namespace is discarded (which is exactly what our codebase does).

PR #14333, opened 2022-11-11
[#14245] Fix read/unread notifications filter

### LINKED ISSUE #14245: Incorrect read/unread filter behavior
## Problem 

The read/unread toggle button is currently behaving as a toggle that **only** displays either read or unread notifications.

#### Expected behavior

The toggle button should display either unread notifications or notifications in any state (both read and unread).

#### Resources

- https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=451%3A18681

rasom:
fix #14245 

status: ready

### DIFF
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -86,18 +86,23 @@
   (and (some? cursor)
        (not= cursor start-or-end-cursor)))
 
-(defn- filter-status->rpc-method
-  [filter-status]
-  (if (= filter-status :read)
-    "wakuext_readActivityCenterNotifications"
-    "wakuext_unreadActivityCenterNotifications"))
+(def ^:const status-read 1)
+(def ^:const status-unread 2)
+(def ^:const status-all 3)
+
+(defn status [filter-status]
+  (case filter-status
+    :read status-read
+    :unread status-unread
+    :all status-all
+    99))
 
 (fx/defn notifications-fetch
   [{:keys [db]} {:keys [cursor filter-type filter-status reset-data?]}]
   (when-not (get-in db [:activity-center :notifications filter-type filter-status :loading?])
     {:db             (assoc-in db [:activity-center :notifications filter-type filter-status :loading?] true)
-     ::json-rpc/call [{:method     (filter-status->rpc-method filter-status)
-                       :params     [cursor (defaults :notifications-per-page) filter-type]
+     ::json-rpc/call [{:method     "wakuext_activityCenterNotificationsBy"
+                       :params     [cursor (defaults :notifications-per-page) filter-type (status filter-status)]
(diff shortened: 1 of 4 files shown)


src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -25,6 +26,14 @@
 
 (defmulti notification-component :type)
 
+;; TODO(rasom): should be removed as soon as all notifications types covered
+(defmethod notification-component :default
+  [{:keys [type]}]
+  [rn/view {:style {:width 300, :height 100}}
+   [rn/text
+    (gstring/format
+     "I exist just to avoid crashing for no reason. I'm sorry. Type %d" type)]])
```
ilmotta, line 35: I think we don't need to tell Clojure devs that `defmulti` requires a default case, just like `case`, otherwise an exception is thrown if no match is found.

src/status_im/ui/screens/activity_center/views.cljs
```diff
@@ -25,6 +26,14 @@
 
 (defmulti notification-component :type)
 
+;; TODO(rasom): should be removed as soon as all notifications types covered
+(defmethod notification-component :default
```
ilmotta, line 30: FYI: Although I wrote the solution with a multimethod, it'll die soon. In the branch I'm working on right now I had to split different notifications into different files for two reasons: code conflicts because we're touching the same file... and second and most important, because the identity verification notification has to be rendered also in a bottom sheet.

src/status_im/activity_center/core.cljs
```diff
@@ -86,18 +86,23 @@
   (and (some? cursor)
        (not= cursor start-or-end-cursor)))
 
-(defn- filter-status->rpc-method
-  [filter-status]
-  (if (= filter-status :read)
-    "wakuext_readActivityCenterNotifications"
-    "wakuext_unreadActivityCenterNotifications"))
+(def ^:const status-read 1)
```
ilmotta, line 89: Do you have a doc reference as to why we should tell the compiler these things are constants? Does it bring any meaningful performance improvement?

ilmotta (approved):
Good call to use a different endpoint `wakuext_activityCenterNotificationsBy`.

src/status_im/activity_center/core.cljs
rasom: > Does it bring any meaningful performance improvement?

it doesn't

src/status_im/ui/screens/activity_center/views.cljs
rasom: It's just to emphasise without checking the code that this should be removed asap.

ilmotta: It was funny to read it :smile: I just try to avoid comments explaining how something in the language works, so that's why I bothered you with this.

PR #14365, opened 2022-11-14
Implement identity verification flow

### LINKED ISSUE #14172: Allow user to reply/decline an identity verification request
# Feature Issue

Allow user to reply and decline an identity verification request.

Drawer

### Acceptance criteria

>Given that Carol received an identity verification request
>When Carol presses "Accept"
>Then she can reply in a text field directly from the existing notification
>And she can write at most 280 characters

>Given that Carol received an identity verification request
>When Carol presses "Accept"
>Then she is still able to decline the challenge by pressing "Decline"

>Given that Carol already filled a valid answer
>When she presses "Send reply"
>Then the same notification changes its status to "Replied"

### Out of scope

- There's no need to implement the actual acceptance flow, as this is more complex and can be done using a separate issue. A dummy event handler might suffice.

### Notes

- It's going to be necessary to change the `activity-log` component to support the inline input text.
- At the moment (2022-10-18), it's necessary to use the unmerged status-go branch https://github.com/status-im/status-go/tree/feature/activity-center-read-unread.

### Figma

Identity request flows are described in https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926.

### LINKED ISSUE #14192: Allow user to either accept or reject an identity verification reply
# Feature Issue

>As a user
>I want to be able to mark an identity verification reply as accepted or untrustworthy
>So that I know I'm more likely communicating with whom I believe it is.

img

### Other criteria

>Given that I received a reply from an identity verification request I sent
>When I receive a notification with the reply
>And I accept it
>Then the same notification on screen updates its state to "Verified"

accepted

>Given that I received a reply from an identity verification request I sent
>When I receive a notification with the reply
>And I reject it
>Then the same notification on screen updates its state to "Rejected"

untrustworthy

### Notes

At the moment (2022-10-20), it's necessary to use the unmerged status-go branch https://github.com/status-im/status-go/tree/feature/activity-center-read-unread.

### Figma

Identity request flows are described in https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926.

ilmotta:
fixes #14172
fixes #14192

### Summary

This PR is bigger than my usual :man_shrugging: It implements the complete identity verification flow (see in [Figma](https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=3804%3A814680)).

See the section [User journey](#user-journey) below for a real example taken from this PR's implementation.

FYI: Accompanying status-go PR: https://github.com/status-im/status-go/pull/2957

### Note

- There's a regression introduced by a recent PR https://github.com/status-im/status-mobile/pull/14333, which I'm not fixing in this PR. It's related to the reconciliation of new notifications, and an issue will be soon created.
- The bottom sheet background, screen background color and blurred background will be fixed in a separate issue to match what's in Figma.

### User journey

1. Alice and Carol recently established mutual friendship, so both can chat with each other, etc.
2. On the other hand, Alice wasn't sure about Carol's identity, so she challenged her. Carol received this new super difficult math challenge!

3. Carol declines the contact verification because the question is disrespectfully easy.

4. But now Alice is pissed off, so she sends another challenge, this time it's a personal question. Carol, the challengee, then replies:

5. On the other end, Alice, the challenger, receives a new notification with Carol's reply, and she marks the challengee as trustworthy.

### Steps to test for developers

I suggest you use the REPL to dispatch contact verification requests between users.

```clojure
(rf/reg-event-fx
   :dev/send-contact-verification-request
   (fn [{:keys [_db]} [_ {:keys [contact-id challenge]}]]
     {::json-rpc/call [{:method     "wakuext_sendContactVerificationRequest"
                        :params     [contact-id challenge]
                        :on-success #(do
                                       (tap> {:success %})
                                       nil)
                        :on-error   #(do
                                       (tap> {:error %})
                                       nil)}]}))

(def carol-id "0x...")

(rf/dispatch [:dev/send-contact-verification-request
                {:challenge  "What's my favorite keyboard?"
                 :contact-id carol-id}])
```

status: ready


status-go-version.json
```diff
@@ -3,7 +3,7 @@
     "_comment": "Instead use: scripts/update-status-go.sh <rev>",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.114.3",
-    "commit-sha1": "7662cd613a9d8bdda4927e583c429b362e3d3815",
-    "src-sha256": "1ldlwhpwdwjxkx9imjinzcw3x5a6wcs70w13kypsfaf4hv17ipmb"
+    "version": "3890748c097b90efdda28f450c2161c4ce963094",
```
ilmotta: FYI: As soon as this PR is approved, this file is going to change to point to the actual release number used by the unmerged PR in status-go https://github.com/status-im/status-go/pull/2957

src/quo2/components/notifications/activity_logs.cljs
```diff
@@ -1,10 +1,50 @@
...
+(def ^:private max-reply-length
+  280)
+
+(defn- valid-reply?
+  [reply]
+  (<= (count reply) max-reply-length))
+
+(defn- activity-reply-text-input
+  [reply-input on-update-reply]
+  [rn/view {:style {:flex-direction :column}}
```
ibrkhalil: I think RN default flex-direction is `column` so this might be unneeded

src/status_im/ui/screens/activity_center/notification/contact_request/view.cljs
```diff
@@ -0,0 +1,58 @@
+(ns status-im.ui.screens.activity-center.notification.contact-request.view
```
ilmotta, line 1: FYI: Don't worry about reviewing this whole file. It has been extracted from another file after a restructure.

src/quo2/components/notifications/activity_logs.cljs
```diff
@@ -66,25 +106,24 @@
...
     [rn/view {:margin-top     12
-              :flex           1
               :flex-direction :row
               :align-items    :flex-start}
      (when button-1
        [button/button (-> button-1
                           (assoc :size size)
-                          (assoc-in [:style :margin-right] 8)
-                          (update :style merge common-style))
+                          (update :style merge common-style {:margin-right 8}))
```
ibrkhalil, line 119: common-style will be nil if `replying?` was `false` or `nil`, Did we test that ?

src/status_im/activity_center/core.cljs
```diff
@@ -6,6 +6,14 @@
             [status-im.utils.fx :as fx]
             [taoensso.timbre :as log]))
 
+;;;; Misc
+
+(fx/defn process-notification-failure
+  {:events [:activity-center/process-notification-failure]}
+  [_ notification-id action error]
+  (log/warn (str "Failed to " action)
+            {:notification-id notification-id :error error}))
```
ibrkhalil, line 15: Can this be a `js/alert` to more visible to users/designers ?

src/status_im/ui/screens/activity_center/notification/contact_request/view.cljs
```diff
@@ -0,0 +1,58 @@
...
+            [status-im.i18n.i18n :as i18n]
+            [status-im.multiaccounts.core :as multiaccounts]
+            [status-im.ui.screens.activity-center.notification.contact-request.style :as style]
+            [status-im.ui.screens.activity-center.utils :as activity-center.utils]
+            [status-im.utils.datetime :as datetime]
+            [status-im.utils.re-frame :as rf]))
+
+(defn view
+  [{:keys [id] :as notification}]
+  (let [message   (or (:message notification) (:last-message notification))
```
ibrkhalil: Can we destructure message and `last-message` from `message` on line 13 ?

src/status_im/ui/screens/activity_center/notification/contact_request/view.cljs
```diff
@@ -0,0 +1,58 @@
...
+        contact   (rf/sub [:contacts/contact-by-identity (:author notification)])
```
ibrkhalil: Destructure `author` too

src/status_im/ui/screens/activity_center/notification/contact_verification/view.cljs
```diff
@@ -0,0 +1,105 @@
...
+            [status-im.ui.screens.activity-center.notification.contact-verification.style :as style]
+            [status-im.ui.screens.activity-center.utils :as activity-center.utils]
+            [status-im.utils.datetime :as datetime]
+            [status-im.utils.re-frame :as rf]))
+
+(defn- hide-bottom-sheet-and-dispatch
+  [event]
+  (rf/dispatch [:bottom-sheet/hide])
+  (rf/dispatch [:dismiss-keyboard])
+  (rf/dispatch event))
```
ibrkhalil, line 16: Can `dispatch-n` be used here ?

src/status_im/ui/screens/activity_center/utils.cljs
```diff
@@ -0,0 +1,6 @@
+(ns status-im.ui.screens.activity-center.utils)
+
+(defn contact-name
+  [contact]
+  (or (get-in contact [:names :nickname])
+      (get-in contact [:names :three-words-name])))
```
ibrkhalil, line 6: Can [this](https://github.com/status-im/status-mobile/commit/5e814b7a035ebdfe13fa23d52dae5d8ee7b62c87) PR be helpful here ?

ibrkhalil (approved):
Non blocking comments

src/quo2/components/notifications/activity_logs.cljs
ilmotta: I did test exactly this scenario @ibrkhalil because I had the same question as you, but the UI looks alright. Clojure's `merge` function can take nil values, and they'll just be ignored/skipped. It's the whole _nil punning_ stuff. For example:

```clojure
(merge {:a 1} nil {:b 2})
; => {:a 1, :b 2}
```

src/status_im/activity_center/core.cljs
ilmotta: I have the same question @ibrkhalil. These failures should almost never happen, because the failure would mean an error from the RPC call because of an invalid RPC payload or because of a bug in status-go, both quite rare. The log is more for the end-user to be able to report a bug with more useful logs, but it's certainly miles away from a good UX.

src/status_im/ui/screens/activity_center/notification/contact_verification/view.cljs
ilmotta: Good suggestion, but I believe we can't use it, because this is only available as an effect, and not as a function. `re-frame.core/dispatch-n` doesn't exist as far as I checked.

src/status_im/ui/screens/activity_center/utils.cljs
ilmotta: I think you're onto something! This piece of code is based on the original Activity Center code, but we can clearly see it doesn't handle the names as that PR, which does something like:

```clojure
(->> [nickname ens-name display-name three-words-name short-public-key]
     (remove string/blank?)
     (take 2))
```

I'm taking note of your comment and I'll try to investigate in the future and open an issue if necessary. Are you okay with that?

src/status_im/ui/screens/activity_center/notification/contact_request/view.cljs
ilmotta: I know you're a _destructuring lover_ from your reviews :heart_decoration: I'm going to change the code ;)

src/status_im/activity_center/core.cljs
ilmotta: I'm going to bring this discussion to designers in the near future. Thanks for raising this concern.

src/status_im/ui/screens/activity_center/utils.cljs
ibrkhalil: Definitely

src/status_im/ui/screens/activity_center/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+(def header-heading
+  {:padding-horizontal screen-padding
+   :padding-vertical   12
+   :color              colors/white})
+
+(defn screen-container
+  [window-width top bottom]
+  {:flex           1
+   :width          window-width
+   :padding-top    (if (> top 0) (+ top 12) 12)
```
rasom: `pos?`

src/status_im/ui/screens/activity_center/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+   :padding-bottom bottom})
+
+(def notifications-container
+  {:flex-grow 1})
+
+(defn notification-container
+  [index]
+  {:margin-top         (if (= 0 index) 0 4)
```
rasom: `zero?`

src/status_im/activity_center/core.cljs
```diff
@@ -45,38 +53,69 @@
...
+                                                notification-id
+                                                :contact-verification/decline
+                                                %])}]})
+
+(fx/defn contact-verification-reply
+  {:events [:activity-center.contact-verification/reply]}
+  [_ notification-id reply]
+  {::json-rpc/call [{:method     "wakuext_acceptContactVerificationRequest"
+                     :params     [notification-id reply]
+                     :on-success #(rf/dispatch [:activity-center/reconcile-notifications-from-response %])
```
rasom, line 82: not really related to this PR, but I'm wondering whether having some function which will receive the event vector and will append results before actually calling `rf/dispatch` will save some `#` and `%` for us. So that we could write something like 
`(rf/dispatch-cb [:activity-center/reconcile-notifications-from-response])`
instead of
`#(rf/dispatch [:activity-center/reconcile-notifications-from-response %])`

smohamedjavid (approved):

src/status_im/activity_center/core.cljs
ilmotta: I think the preferred approach is if we actually made `:on-success` and `:on-fail` only receive data and leave the dispatch call to the fx implementation. This is how other effects are usually written in re-frame, like https://github.com/day8/re-frame-http-fx.

The idea being that event handlers should be able to represent their subsequent effects as pure, serializable data. So the fx implementation would always `conj` into the vector passed to `:on-success` and `:on-error`.

```clojure
{::json-rpc/call [{:method     "wakuext_declineContactVerificationRequest"
                   :params     [contact-verification-id]
                   :on-success [:activity-center.contact-verification/decline-success])
                   :on-error   [:activity-center.contact-verification/decline-error contact-verification-id])}]}
```

ilmotta: On top of that @rasom, describing effects as raw data helps with inspecting values (e.g. with `tap>` and with testing (it's easier to match against a vector instead of an anonymous function).

OmarBasem (commented):
Why are there several new files in the `ui` namespace?

ilmotta:
> Why are there several new files in the `ui` namespace?

I'm not sure I understand the problem @OmarBasem, could you elaborate a bit? I'm currently rebasing with what's on develop, but in general the structure of the directories and files is following whatever was already in place before this PR.

OmarBasem:
> > Why are there several new files in the `ui` namespace?
> 
> I'm not sure I understand the problem @OmarBasem, could you elaborate a bit? I'm currently rebasing with what's on develop, but in general the structure of the directories and files is following whatever was already in place before this PR.

I mean I see some new files in `ui` namespace. Shouldn't the new files be in `ui2`?

ilmotta:
> > > Why are there several new files in the `ui` namespace?
> > 
> > 
> > I'm not sure I understand the problem @OmarBasem, could you elaborate a bit? I'm currently rebasing with what's on develop, but in general the structure of the directories and files is following whatever was already in place before this PR.
> 
> I mean I see some new files in `ui` namespace. Shouldn't the new files be in `ui2`?

We should, but this PR was big enough that I avoided doing this refactoring. But I took note of your feedback @OmarBasem and I'll try to to this in a future PR.

PR #14389, opened 2022-11-17
Fix Metro Warnings ⚠️

### LINKED ISSUE #14380: ⚠️ Metro Warnings ⚠️
Currently when running `metro` bundler lots of warnings are printed which clutters metro logs, and makes it harder to find and read debugging logs.

Some of these warnings are:

* `Warning: Each child in a list should have a unique "key" prop.`
* `Warning: Invalid prop `style` supplied to `React.Fragment`.
* `Warning: Every element in a seq should have a unique :key:`
* `Warning: Reactive deref not supported in lazy seq, it should be wrapped in doall:`
* `Warning: Failed prop type: Invalid props.style key `ellipsizeMode` supplied to `Text``
* ` Warning: Failed child context type: Invalid child context `virtualizedCell.cellKey` of type `number` supplied to `CellRenderer`, expected `string`.`
* VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead. `
* `re-frame: overwriting :fx handler for:`
* `re-frame: no handler registered for effect:`

This issue is an attempt to fix some of these warnings.

OmarBasem:
fixes #14380

This PR fixes many of the warnings and errors in metro logs which were making it difficult to find and read debugging logs.
It fixes the following warnings and errors:

* Fixes unique keys for bottom sheet list 
* Fixes could not find source for identicon-ring32 icon 
* Fixes could not find source for group12 icon 
* Fixes invalid props.style key ellipsizeMode supplied to Text
* Fixes re-frame: no handler registered for effect: pop-to-root-fx and implements it
* Fixes re-frame: no handler registered for effect: change-tab (but needs implementation)
* Fixes Failed child context type: Invalid child contextvirtualizedCell.cellKey of type number supplied to CellRenderer
* Fixes Invalid prop `style` supplied to `React.Fragment`. 
* Fixes Reactive deref not supported in lazy seq, it should be wrapped in doall

This PR also removes `ignoreAllLogs` as obviously it was a bad practice, and has resulted in the project being cluttered with many errors and warnings when running metro. RN warning and error screen act as a net to catch some bugs, and removing it has resulted in some bugs slipping through. I have added some comments in `status-im2.setup.dev` regarding ignoring logs.

status: ready


src/quo2/components/avatars/user_avatar.cljs
```diff
@@ -126,10 +126,10 @@
                       :height        outer-dimensions
                       :border-radius outer-dimensions}}
      (when (and ring? identicon?)
-       [icons/icon :main-icons/identicon-ring {:width    outer-dimensions
-                                               :height   outer-dimensions
-                                               :size     outer-dimensions
-                                               :no-color true}])
+       [icons/icon :i/identicon-ring {:width    outer-dimensions
```
flexsurfer: is it like a fake right or why is it needed ?

src/quo2/components/drawers/action_drawers.cljs
```diff
@@ -65,10 +17,60 @@
...
 (defn action-drawer [sections]
-  [:<> {:style
-        {:flex 1}}
-   (interpose
-    [divider]
+  [:<>
+   (doall
     (for [actions sections]
-      (map action actions)))])
+      (doall
```
flexsurfer, line 75: are you sure its needed twice ?

src/status_im/ui2/screens/chat/home.cljs
```diff
@@ -195,7 +195,7 @@
                            {:id    :sent
                             :label (i18n/label :t/sent)}]}]
         [list/flat-list
-         {:key-fn    :first
+         {:key-fn    #(:chat-id %)
```
flexsurfer: just `:chat-id` ,keyword is a function

src/status_im2/setup/dev.cljs
```diff
@@ -4,7 +4,18 @@
             [status-im.ethereum.json-rpc :as json-rpc]
             ["react-native" :refer (DevSettings LogBox)]))
 
-(.ignoreAllLogs LogBox)
+;; Only ignore warnings/errors that cannot be fixed for the time being.
+;; When you add a warning to be ignored explain below why it is ignored and how it can be fixed.
+;; When a warning is fixed make sure to remove it from here.
+(.ignoreLogs ^js LogBox (clj->js ["undefined is not an object (evaluating 'e.message')"
```
flexsurfer, line 10: please no, it's not worth it, there are always will be annoying warnings when hot reload, etc

OmarBasem: These warnings are meant to be fixed :) Otherwise, we will be piling up warnings and errors again. And this command does not just ignore warnings but errors too, which can cause bugs. For example, someone wrote `ellipsizeMode` inside a style prop, but this is incorrect, `ellipsizeMode` is a prop itself. Another one, lists should always be using unique keys for performance reasons, and also the keys must be strings. Also, some re-frame events had no effects implemented. IgnoreAllLogs is going to ignore these things and more. So it is better to see the warnings. When you see a warning/error you can just add it to the LogBox if you do not wish to fix it right now, at least we would know there is something being ignored. :)

flexsurfer: i mean this won't work during development, because you will have warnings and errors and that's fine, but you don't want to hide them manually all the time

ilmotta: I agree with @OmarBasem, it's better not to sweep warnings under the carpet, so we can detect and fix them.

ilmotta (approved):

src/status_im2/setup/dev.cljs
flexsurfer: i also agree, we can see them in the log, but during active development i don't want to see temporary warnings  in the app

flexsurfer: ok if you really want this we can try it, but at least we need to ignore re-frame hot reload warnings

OmarBasem: @flexsurfer We shouldn't be having warnings and errors all the time. After this PR you will not find many warnings, I have fixed all the warnings that are appearing. Whenever there is a new warning/error, we either fix it right now or add it to the ignore box. Simple. And if someone really wishes they can toggle `ignoreAllLogs` while developing, tho it is better not to.

OmarBasem: There were a couple of re-frame warnings which this PR fixes, and I am not getting any other re-frame warnings.

src/quo2/components/avatars/user_avatar.cljs
OmarBasem: I did not write this code. I only changed `main-icons` to `i`

src/quo2/components/drawers/action_drawers.cljs
OmarBasem: Yes

OmarBasem: Removing either of the `doall` brings back the warning. Probably because it is a nested loop.

src/status_im2/setup/dev.cljs
OmarBasem: Btw, since RN v0.63 you do not get the warning screen in your face, you only get a small notification at the bottom of the screen which is much less intrusive

yqrashawn: I was trying to remove the in-app error overlay early today(and failed). 
So If I understand right, `.ignoreAllLogs` and `.ignoreLogs` only prevent things to display in the error overlay in the emulator, right?  We can still see them in the terminal.

OmarBasem: @yqrashawn as of now in `develop` all logs are ignored. If you are getting an error overlay, this means that it is a fatal error. These ones cannot be disabled. 
With `ignoreLogs` you can still see all logs in the terminal.

yqrashawn: OK, got that. Thanks. 

> Btw, since RN v0.63 you do not get the warning screen in your face, you only get a small notification at the bottom of the screen which is much less intrusive

(Not related to this PR) If only we can apply this to fatal error as well. It's really annoying since I hadn't found a way to minimize that without mouse.

yqrashawn (approved):

src/status_im2/setup/dev.cljs
OmarBasem: @yqrashawn they are called fatal for a reason :) I do not think there is a way to disable or minimize them in debug mode. In release mode, they would crash the app.

yqrashawn: Yes, it's just it's the same error again and again. I navigate the app by evaluating code in the editor. So, I can just navigate back and forth to check if the error fixed. The error overly is just in my way to quickly try things.

flexsurfer: @OmarBasem when you hot reload re-frame events there will be a warnings about overwriting events

OmarBasem: @flexsurfer I just tried this and I am not getting any warnings. If we start getting annoying `re-frame` warnings, then for sure will add it 👍

flexsurfer: did you try changes in the event handler? or fx ?

OmarBasem: Yes, I tried doing changes to `navigate-to-chat` fx in `status-im.chat.models` and also tried doing changes to `status-im2.setup.events` and no warnings happened

flexsurfer: hm strange, ok

src/quo2/components/avatars/user_avatar.cljs
flexsurfer: yeah i know, its not possible to add comment to the image, i see you added the image, and my question why is it needed, it looks like fake one, are we sure we want to add it ?

ibrkhalil (approved):

ibrkhalil:
Misclicked the update branch btn, Plz revert

src/quo2/components/avatars/user_avatar.cljs
OmarBasem: I am not sure what do you mean? Are you referring to the `width` and `height` on the icon component. If so, I have removed them.

flexsurfer: i mean this icon [resources/images/icons2/32x32/identicon-ring@2x.png](https://github.com/status-im/status-mobile/pull/14389/files#diff-76027346c18abfa251171b5eb3d8c02017cfd4fb52c5bb8d511bc78b6b9d47c8)

OmarBasem: This is the identicon-ring icon from figma. I may be missing something, but why are you saying it is fake?

flexsurfer: because it should be generated for every account, not static, in figma its just an example

OmarBasem: Oh okay, I did not know that. We can keep as a placeholder for now?

flexsurfer: yeah we can i guess

siddarthkay (approved):

flexsurfer (approved):

Issue #14392, opened 2022-11-17
New design implementation and tech debt


flexsurfer:
Communities

- [x] https://github.com/status-im/status-mobile/issues/14387
- [x] https://github.com/status-im/status-mobile/issues/14391
- [x] https://github.com/status-im/status-mobile/issues/14408
- [x]  https://github.com/status-im/status-mobile/issues/14527
- [x] https://github.com/status-im/status-mobile/issues/14528
- [x] https://github.com/status-im/status-mobile/issues/14597
- [x] https://github.com/status-im/status-mobile/issues/14702
- [x] https://github.com/status-im/status-mobile/issues/14529
- [x] https://github.com/status-im/status-mobile/issues/14552
- [x] https://github.com/status-im/status-mobile/issues/14584
- [x] https://github.com/status-im/status-mobile/issues/14583
- [x] https://github.com/status-im/status-mobile/issues/14585
- [x] https://github.com/status-im/status-mobile/issues/14586
- [x] https://github.com/status-im/status-mobile/issues/14587
- [x] https://github.com/status-im/status-mobile/issues/14600
- [x] https://github.com/status-im/status-mobile/issues/14615
- [x] https://github.com/status-im/status-mobile/issues/14703
- [x] https://github.com/status-im/status-mobile/issues/15513
- [x] https://github.com/status-im/status-mobile/issues/15515

churik:
@flexsurfer @ilmotta 
may be we can review epic and subtask and designs, as a lot changed since then?

Let's review the epic and close unnecessary ones.

PR #14394, opened 2022-11-17
move communities 2

flexsurfer:
move communities 2

also see https://github.com/status-im/status-mobile/issues/14392


OmarBasem (commented):
All looks good to me. A suggestion, why don't we run QA testing for migration PRs to avoid breaking anything (which may impact or slow down other PRs)? At least to test the main screens are rendering and main functionalities are working.

flexsurfer:
yeah it would be good, but more time it's open more conflicts and bugs it might cause , so better to merge it faster, and i tested it manually, i believe communities is not implemented yet properly

OmarBasem (approved):

Issue #14385, opened 2022-11-16
Chat view freezes if open/close chat multiple times (iOS)


qoqobolo:
# Bug Report

## Problem 

Sometimes, a chat view stops scrolling and the back button stops responding on iOS if you navigate to a chat and back several times (community channel, 1-1, etc.).
Such actions as focusing on the input field or swiping to the right (to go back) fix the issue.

#### Expected behavior
UI doesn't freeze

#### Actual behavior
UI partially freezes

### Reproduction

1. Open a community on an iOS device
2. Open and close a channel several times

### Additional Information

- Status version: nightly 16/11
- Operating System: iOS

rasom:
There is no need to scroll, in order to reproduce the issue on iOS simulator I only needed to open/close chat multiple times.

PR #14396, opened 2022-11-17
[#14385] Comment out background which overlaps chat screen

rasom:
Sometimes that background overlaps content of chat screen. If it is is non-functional then no point in fixing it for now.

status: ready ?

### DIFF
--- a/src/status_im/ui2/screens/chat/composer/view.cljs
+++ b/src/status_im/ui2/screens/chat/composer/view.cljs
@@ -188,7 +188,7 @@
                                                        (re-frame/dispatch [:chat.ui/send-current-message]))}
                     :i/arrow-up]]])
                ;black background
-               [reanimated/view {:style (reanimated/apply-animations-to-style
-                                         {:opacity bg-opacity}
-                                         (styles/bottom-sheet-background window-height))}]
+               #_[reanimated/view {:style (reanimated/apply-animations-to-style
+                                           {:opacity bg-opacity}
+                                           (styles/bottom-sheet-background window-height))}]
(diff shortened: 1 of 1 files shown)


src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -188,7 +188,7 @@
                                                        (re-frame/dispatch [:chat.ui/send-current-message]))}
                     :i/arrow-up]]])
                ;black background
-               [reanimated/view {:style (reanimated/apply-animations-to-style
-                                         {:opacity bg-opacity}
-                                         (styles/bottom-sheet-background window-height))}]
+               #_[reanimated/view {:style (reanimated/apply-animations-to-style
```
flexsurfer, line 191: why not fix it?

flexsurfer: just to unblock chat for now? ok then we need to create an issues so we can fix it properly

flexsurfer (approved):

src/status_im/ui2/screens/chat/composer/view.cljs
rasom: @flexsurfer I don't mind fixing it, currently I'm not sure where this background is used and how to check whether it works at all

flexsurfer: yeah it is used when you maximize the composer, you can pull it, or type lots of text

rasom: ok, I'll take a look then, thanks

rasom:
I will not have time for this now unfortunately

PR #14401, opened 2022-11-17
Fix reconciliation of notifications for new filter type :all

### LINKED ISSUE #14400: Reconciliation doesn't work after the fix to the read/unread filter
# Bug Report

Reconciliation of new notifications into the app db state stopped working after https://github.com/status-im/status-mobile/pull/14333 got merged.

#### Solution

The fix is not complicated. We need to take in consideration the difference in semantics between filtering `:all` instead of `:read` and update the private function `status-im.activity-center/update-notifications`.

#### Expected behavior

- When the user interacts with a notification and its status changes, the UI should immediately reflect the changes.
- When the user is in the Activity Center screen and new notifications arrives, the UI should automatically show them without requiring the user to refresh the screen.

ilmotta:
fixes #14400 

### Summary

See issue for details about the bug.

#### Notes

Fixed inconsistent usage of the feature flag `new-activity-center-enabled?`. Due to the duplication of certain namespaces during the ongoing project restructure, some parts of the app were pointing to the new (duplicated) flag, while others were pointing to the old one.

status: ready

### DIFF
--- a/src/status_im/activity_center/core.cljs
+++ b/src/status_im/activity_center/core.cljs
@@ -38,8 +38,7 @@
   ~excessively~ big, this implementation will probably need to be revisited."
   [db-notifications new-notifications]
   (reduce (fn [acc {:keys [id type read] :as notification}]
-            (let [filter-status       (if read :read :unread)
-                  remove-notification (fn [data]
+            (let [remove-notification (fn [data]
                                         (remove #(= id (:id %)) data))
                   insert-and-sort     (fn [data]
                                         (->> notification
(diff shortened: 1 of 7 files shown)


src/status_im/activity_center/core.cljs
```diff
@@ -137,13 +139,11 @@
   (and (some? cursor)
        (not= cursor start-or-end-cursor)))
 
-(def ^:const status-read 1)
```
ilmotta, line 140: FYI: Even though the endpoint `wakuext_activityCenterNotificationsBy` supports the `read` filter type, the UI only toggles between `all` or `unread`. Supporting the `read` filter type would incur in unnecessary computational cost to reconcile notifications, so I removed it.

J-Son89 (approved):

PR #14346, opened 2022-11-13
[WIP][issue-14175] new chat user selector redesign #14345

ibrkhalil:
status: wip
fixes: #14227, #14228, #14226

https://user-images.githubusercontent.com/33176106/201517423-d2b1bba1-177c-40bf-8617-6145b18a8d21.mp4

PR #14416, opened 2022-11-22
[14413] User list component

### LINKED ISSUE #14413: Create user list component
User list component
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=114%3A6475&t=14qN1i0geDIo8Dlo-0

ibrkhalil:
fixes #14413
Required for [this](https://github.com/status-im/status-mobile/pull/14346)

status: ready

### DIFF
--- a/src/quo2/components/list_items/user_list/style.cljs
+++ b/src/quo2/components/list_items/user_list/style.cljs
@@ -1,0 +1,40 @@
+(ns quo2.components.list-items.user-list.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(defn user-list-wrapper [can-be-invited? background-type]
+  (let [height (if can-be-invited? 56 80)]
+    {:width              359
+     :border-radius      12
+     :padding-vertical   8
+     :padding-horizontal 12
+     :height             height
+     :background-color   (case background-type
+                           :5  colors/primary-50-opa-5
+                           :10 colors/primary-50-opa-10
+                           :tranpsarent)
+     :flex-direction     :column
+     :justify-content    :center}))
+
+(def ml-8 {:margin-left 8})
+
+(def row-centered {:flex-direction :row
+                   :align-items    :center})
+
+(def icon-container-styles {:width           32
+                            :height          32
+                            :border-radius   10
+                            :border-width    1
+                            :border-color    (colors/theme-colors colors/neutral-80-opa-10 colors/white-opa-10)
+                            :justify-content :center
+                            :align-items     :center})
+
+(def container (assoc row-centered :justify-content :space-between))
+
+(def cant-be-invited {:flex-direction :row
+                      :margin-top     8
+                      :margin-left    40})
+
+(def icon-container {:margin-right 2
+                     :margin-top   2})
+
+(def cant-be-invited-text-style {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)})
--- a/src/quo2/components/list_items/user_list/style.cljs
+++ b/src/quo2/components/list_items/user_list/style.cljs
@@ -11,7 +11,7 @@
      :background-color   (case background-type
                            :5  colors/primary-50-opa-5
                            :10 colors/primary-50-opa-10
-                           :tranpsarent)
+                           :transparent :transparent)
      :flex-direction     :column
      :justify-content    :center}))
 
--- a/src/quo2/components/list_items/user_list/view.cljs
+++ b/src/quo2/components/list_items/user_list/view.cljs
@@ -1,0 +1,67 @@
+(ns quo2.components.list-items.user-list.view
+  (:require [quo2.components.avatars.user-avatar :as user-avatar]
+            [quo2.components.icon :as icons]
+            [quo2.components.list-items.user-list.style :as style]
+            [quo2.components.markdown.text :as text]
+            [quo2.foundations.colors :as colors]
+            [react-native.core :as rn]
+            [quo2.components.selectors.selectors :as selectors]
+            [status-im.multiaccounts.core :as multiaccounts]
+            [status-im.utils.utils :as utils]))
+
+(def ^:private icon-types {:options :i/options
+                           :close   :i/close})
+
+(defn- cant-be-invited-msg []
+  [rn/view {:style               style/cant-be-invited
+            :accessibility-label "user can't be invite"}
+   [icons/icon :i/info {:size            12
+                        :color           colors/neutral-40
+                        :container-style style/icon-container}]
+   [text/text {:weight :regular
+               :size   :label
+               :style  style/cant-be-invited-text-style}
+    "Can’t be invited due to their privacy settings"]])
+
+(defn- right-section [{:keys [id button-type on-press checked?]}]
+  (if (= button-type :checkbox)
+    [selectors/checkbox {:default-checked? checked?}]
+    [rn/touchable-opacity
+     {:on-press            on-press
+      :accessibility-label (str (or id "user-list") "-" (if (= button-type :options)
+                                                          "options"
+                                                          "close"))}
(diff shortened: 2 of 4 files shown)


src/quo2/components/list_items/user_list/view.cljs
```diff
@@ -0,0 +1,67 @@
+(ns quo2.components.list-items.user-list.view
+  (:require [quo2.components.avatars.user-avatar :as user-avatar]
+            [quo2.components.icon :as icons]
+            [quo2.components.list-items.user-list.style :as style]
+            [quo2.components.markdown.text :as text]
+            [quo2.foundations.colors :as colors]
+            [react-native.core :as rn]
+            [quo2.components.selectors.selectors :as selectors]
+            [status-im.multiaccounts.core :as multiaccounts]
```
flexsurfer, line 9: status-im namespace shouldn't be used in quo2

src/quo2/components/list_items/user_list/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+   [rn/view style/row-centered
+    [user-avatar/user-avatar {:size              :small
+                              :ring?             true
+                              :online?           true
+                              :profile-picture   (multiaccounts/displayed-photo contact)
+                              :status-indicator? true}]
+    [rn/view {:style style/ml-8}
+     [text/text {:weight :semi-bold
+                 :size   :paragraph-1
+                 :style  {:color (colors/theme-colors colors/neutral-100 colors/white)}} (-> contact
```
flexsurfer, line 50: this is default colors, so no need to specify them

flexsurfer (commented):
isn't this the same component https://github.com/status-im/status-mobile/blob/develop/src/status_im/ui2/screens/chat/components/contact_item/view.cljs ?

ibrkhalil:
> isn't this the same component https://github.com/status-im/status-mobile/blob/develop/src/status_im/ui2/screens/chat/components/contact_item/view.cljs ?

They are very similar .. But I think they're not the same .. It'll be better if I refactored one and used it there

src/quo2/components/list_items/user_list/style.cljs
```diff
@@ -0,0 +1,40 @@
...
+  (let [height (if can-be-invited? 56 80)]
+    {:width              359
+     :border-radius      12
+     :padding-vertical   8
+     :padding-horizontal 12
+     :height             height
+     :background-color   (case background-type
+                           :5  colors/primary-50-opa-5
+                           :10 colors/primary-50-opa-10
+                           :transparent :transparent)
```
ilmotta, line 14: I find it odd to read that `background-type` can be a keyword `:5` (as a number), it almost looks as a typo even though it might not be.

It'd be nice if we could map `:5` and `:10` to non-magical values. Is there any semantic meaning to them? That can give you a clue about their *type*.

Also note that `case` throws an exception without a default case.

src/quo2/components/list_items/user_list/style.cljs
```diff
@@ -0,0 +1,40 @@
...
+     :flex-direction     :column
+     :justify-content    :center}))
+
+(def ml-8 {:margin-left 8})
```
ilmotta, line 18: There's nothing wrong with this per se, but giving a name for a margin left seems excessive. You actually reminded me of Tailwind CSS or Bootstrap because of utility class names like `ml-8`, which I'm actually quite fond of :)

src/quo2/components/list_items/user_list/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+(defn- cant-be-invited-msg []
+  [rn/view {:style               style/cant-be-invited
+            :accessibility-label "user can't be invite"}
+   [icons/icon :i/info {:size            12
+                        :color           colors/neutral-40
+                        :container-style style/icon-container}]
+   [text/text {:weight :regular
+               :size   :label
+               :style  style/cant-be-invited-text-style}
+    "Can’t be invited due to their privacy settings"]])
```
ilmotta, line 24: Shouldn't we use a proper translation here?

src/quo2/components/list_items/user_list/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+  (if (= button-type :checkbox)
+    [selectors/checkbox {:default-checked? checked?}]
+    [rn/touchable-opacity
+     {:on-press            on-press
+      :accessibility-label (str (or id "user-list") "-" (if (= button-type :options)
+                                                          "options"
+                                                          "close"))}
+     [rn/view {:style (when (= button-type :close)
+                        style/icon-container-styles)}
+      [icons/icon (button-type icon-types) {:size            20
```
ilmotta, line 36: [Style] Excess whitespace.

src/quo2/components/list_items/user_list/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+               :size   :label
+               :style  style/cant-be-invited-text-style}
+    "Can’t be invited due to their privacy settings"]])
+
+(defn- right-section [{:keys [id button-type on-press checked?]}]
+  (if (= button-type :checkbox)
+    [selectors/checkbox {:default-checked? checked?}]
+    [rn/touchable-opacity
+     {:on-press            on-press
+      :accessibility-label (str (or id "user-list") "-" (if (= button-type :options)
```
ilmotta, line 31: I believe accessibility labels should be very human readable. Adding the ID to it doesn't seem very helpful when using a screen reader.

On top of that @ibrkhalil, not something for this PR, but for future discussions, I think our current conventions for accessibility labels might not be the best, i.e. using keywords with dashes separating words and often being very short seems to be very machine oriented. I'd love to hear more about this from people who use these labels, particularly in RN apps. I'd also be interesting to see how other accessibility properties affect screen readers and how we can better use them (e.g. https://reactnative.dev/docs/accessibility#accessibilityhint).

src/quo2/components/list_items/user_list/view.cljs
ilmotta, line 50: It seems the thread macro is unnecessary here.

src/quo2/components/list_items/user_list/style.cljs
ibrkhalil: Maybe `opacity-10` ?

ilmotta: I don't know either, naming is difficult, but when I read the word _type_ I immediately think `10` is not a type, neither `opacity-10`, these are more like values. Perhaps `background-type` is what's confusing me, because I read it like there's some meaning to it. But nevermind ;)

image

src/quo2/components/list_items/user_list/style.cljs
```diff
@@ -0,0 +1,40 @@
...
+                            :height          32
+                            :border-radius   10
+                            :border-width    1
+                            :border-color    (colors/theme-colors colors/neutral-80-opa-10 colors/white-opa-10)
+                            :justify-content :center
+                            :align-items     :center})
+
+(def container (assoc row-centered :justify-content :space-between))
+
+(def cant-be-invited {:flex-direction :row
```
ilmotta, line 33: [Nitpick] I almost missed `cant` for `can`, so maybe `cannot` would be clearer.

flexsurfer:
list is being implemented here https://github.com/status-im/status-mobile/pull/14427

src/quo2/components/list_items/user_list/view.cljs
ibrkhalil: I am interested to know how screen readers handle dashes

ilmotta: Me too, I'll try to learn more about this soon.

PR #14377, opened 2022-11-16
Group Details Screen

OmarBasem:
fixes #14351

This PR implements the group details screen (UI). Will implement the actions buttons functionalities in another PR.

[Designs.](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=2103%3A604427&t=egbHjxkwseMZlRzM-0)

status: ready

PR #14419, opened 2022-11-22
move chats home

flexsurfer:
move chats home

simplified code a lot, unified code for communities and chat home screens, moved some functions to subscriptions and made them more clojure idiomatic, replaced lots of wrong markdown usages 

also fixes: https://github.com/status-im/status-mobile/issues/14408

before 
```clojure
(defn prepare-contacts [contacts]
  (let [data (atom {})]
    (doseq [i (range (count contacts))]
      (let [first-char (get (:alias (nth contacts i)) 0)]
        (if-not (contains? @data first-char)
          (swap! data #(assoc % first-char {:title first-char :data [(nth contacts i)]}))
          (swap! data #(assoc-in % [first-char :data] (conj (:data (get @data first-char)) (nth contacts i)))))))
    (swap! data #(sort @data))
    (vals @data)))
```
after
```clojure
(re-frame/reg-sub
 :contacts/active-sections
 :<- [:contacts/active]
 (fn [contacts]
   (-> (reduce
        (fn [acc contact]
          (let [first-char (first (:alias contact))]
            (if (get acc first-char)
              (update-in acc [first-char :data] #(conj % contact))
              (assoc acc first-char {:title first-char :data [contact]}))))
        {}
        contacts)
       sort
       vals)))
```

before 
```clojure
[text/text {:style (merge typography/paragraph-1 typography/font-semi-bold
                                 {:color (colors/theme-colors colors/neutral-100 colors/white)})}
        display-name]
```
after
```clojure
[quo/text {:weight :semi-bold} display-name]
```

PR #13117, opened 2022-02-18
gap_test_improvement

pavloburykh:
Improvement of e2e https://ethstatus.testrail.net/index.php?/cases/view/700727

What added: 
1. Check there is not gap in 1-1 and group chat
2. Check gap message is not fetched before tapping on Fetch message

PR #14427, opened 2022-11-23
Group Details Screen (2)

### LINKED ISSUE #14402: Group Details Screen (2)
A second issue to track the creation of group details screen designs and functionalities, including adding members and managing members.

[Designs.](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=2105%3A522742&t=CxuYx4HdYmNSONDs-0)

Dependency - issue: https://github.com/status-im/status-mobile/issues/14351, PR: https://github.com/status-im/status-mobile/pull/14377

OmarBasem:
Fixes #14402

This PR is a continuation of [that PR](https://github.com/status-im/status-mobile/pull/14377).
This PR implements add members screen UI, and a separate `contact-list` component.
There will be one more group details screen PR (still waiting for some input from the design team).

Regarding QA, since the functionality is not yet complete, so will skip the QA for now, and will run the final PR through QA when the functionality is complete.


src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -60,14 +60,16 @@
...
   (hide-sheet-and-dispatch [:bottom-sheet/show-sheet
                             {:content (fn []
                                         (common/alert {:title       (i18n/label :t/leave-group?)
                                                        :description (i18n/label :t/leave-chat-confirmation)
                                                        :context     item
                                                        :button-text (i18n/label :t/leave-group)
-                                                       :on-press    #(hide-sheet-and-dispatch [:chat.ui/leave-chat chat-id])}))}]))
+                                                       :on-press    #(do
+                                                                       (rf/dispatch [:navigate-back])
+                                                                       (hide-sheet-and-dispatch [:group-chats.ui/leave-chat-confirmed chat-id]))}))}]))
```
ibrkhalil: Do we need to navigate back and hide the sheet?

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -117,13 +119,14 @@
...
-          :sub-label nil
-          :chevron?  false}))
+(defn leave-group-entry [item extra-data]
+  (entry {:icon         :i/log-out
+          :label        (i18n/label :t/leave-group)
+          :on-press     #(leave-group-action item (if extra-data (:chat-id extra-data) (:chat-id item)))
+          :danger?      true
+          :sub-label    nil
+          :chevron?     false
+          :add-divider? (when (not= extra-data nil) true)}))
```
ibrkhalil: Can use `some?` to check for `nil` vals

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
```diff
@@ -9,26 +9,43 @@
...
+    [rn/touchable-opacity {:on-press #(rf/dispatch [:bottom-sheet/show-sheet
+                                                    {:content (fn [] [actions/actions item extra-data])}])
+                           :style {:position :absolute
+                                   :right    20}}
+     (if (= icon :options)
+       [icons/icon :i/options {:size 20 :color (colors/theme-colors colors/neutral-50 colors/neutral-40)}]
+       [selectors/checkbox {:default-checked? member?
+                            :on-change        (fn [selected] (if selected
+                                                               (swap! added conj public-key)
+                                                               (reset! added (filter #(-> % (not= public-key)) @added))))}])]))
```
ibrkhalil: Unneeded thread first macro

src/status_im/ui2/screens/common/contact_list/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+            [quo2.core :as quo2]
+            [utils.re-frame :as rf]
+            [reagent.core :as reagent]
+            [clojure.string :as str]))
+
+(def query (reagent/atom ""))
+
+(defn prepare-contacts [contacts]
+  (let [data (atom {})]
+    (doseq [i (range (count contacts))]
```
ibrkhalil: If we passed in contacts directly, Can't `doseq` iterate over it's count ?

ibrkhalil (approved):
Non blocking comments

src/status_im/ui2/screens/chat/actions.cljs
ilmotta: The whole logic could be replaced by a simpler `(not extra-data)` because we only care about the absence of `extra-data`.

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
ilmotta: You can also use `remove` to simplify the filtering step.

```clojure
(reset! added (remove #(= % public-key) @added))
```

And you might like to use `partial` because we can't nest the `#()` reader macro, but some people find it difficult to read, so use sparingly:

```clojure
(swap! added #(remove (partial = public-key) %))
```

src/status_im/ui2/screens/common/contact_list/view.cljs
ilmotta: This function is too non-idiomatic for Clojure. In Clojure we resist the temptation of processing data in a mutable fashion. At the very least, you can use reduce with the accumulator with a map, but there could be even better ways. It would also perform better because an atom is not like an ordinary variable in Javascript, see the docstring below to get an idea:

> Atomically swaps the value of atom to be: (apply f current-value-of-atom args). Note that f may be called
multiple times, and thus should be free of side effects.  Returns the value that was swapped in.

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -141,13 +144,22 @@
           :sub-label nil
           :chevron?  false}))
 
-(defn notifications-entry []
-  (entry {:icon      :i/notifications
-          :label     (i18n/label :t/notifications)
+(defn edit-name-image-entry []
+  (entry {:icon      :i/edit
+          :label     (i18n/label :t/edit-name-and-image)
           :on-press  #(js/alert "TODO: to be implemented, requires design input")
```
ilmotta: Could you add your username to these TODOs? Adding your username can help other people to know who's the person with the original context.

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -311,23 +333,34 @@
                          (notification-actions item inside-chat?)
                          (destructive-actions item)]])
 
-(defn contact-actions [{:keys [public-key added] :as contact}]
-  [drawer/action-drawer [[(view-profile-entry public-key)
-                          (when added
+(defn contact-actions [{:keys [public-key] :as contact} {:keys [chat-id admin?] :as extra-data}]
+  (let [current-pk (rf/sub [:multiaccount/public-key])]
```
ilmotta: Could you rename this binding to `current-public-key` or `current-pub-key`? There are many usages of `pk` referring to *private key* in the literature, so this is always very confusing to me.

src/status_im/ui2/screens/chat/actions.cljs
```diff
@@ -311,23 +333,34 @@
...
     [one-to-one-actions item inside-chat?]
     constants/public-chat-type
     [public-chat-actions item inside-chat?]
     constants/private-group-chat-type
     [private-group-chat-actions item inside-chat?]
-    [contact-actions item]))
+    [contact-actions item extra-data]))
+
+(defn group-details-actions [{:keys [admins] :as group}]
+  (let [current-pk (rf/sub [:multiaccount/public-key])
```
ilmotta: Same comment as above about renaming `current-pk` to `current-public-key`.

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
```diff
@@ -9,26 +9,43 @@
             [quo2.components.markdown.text :as text]
             [status-im.ui2.screens.chat.components.message-home-item.style :as style]
             [utils.re-frame :as rf]
-            [status-im.ui2.screens.chat.actions :as actions]))
+            [status-im.ui2.screens.chat.actions :as actions]
+            [quo2.components.selectors.selectors :as selectors]))
```
ilmotta: It would be good if you could update the namespaces to use `quo2.core`.

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -17,12 +23,21 @@
...
-                :accessibility-label :options-button}
-   [quo2/icon :i/options {:color (colors/theme-colors colors/neutral-100 colors/white)}]])
+  (let [group (rf/sub [:chats/current-chat])]
+    [quo2/button {:type                :grey
+                  :size                32
+                  :width               32
+                  :style               {:margin-right 20}
+                  :accessibility-label :options-button
+                  :on-press            #(rf/dispatch [:bottom-sheet/show-sheet
+                                                      {:content (fn [] [actions/group-details-actions group])}])}
```
ilmotta, line 34: FYI: Not a suggestion, just for the curious, you can use `constantly` here instead of the anonymous function. Some people like it.

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -45,8 +60,49 @@
...
+                  :padding-horizontal 20
+                  :padding-vertical   12
+                  :padding-bottom     (+ (:bottom safe-area) 33)
+                  :width              "100%"
+                  :background-color   colors/white
+                  :flex-direction     :row
+                  :bottom             0}
+         [quo2/button {:style    {:flex 1}
+                       :on-press #(rf/dispatch [:bottom-sheet/hide])
+                       :disabled (= (count @added) 0)}
```
ilmotta: It's safe to use `zero?` instead of `(= 0`)

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: it was fixed here https://github.com/status-im/status-mobile/pull/14419

flexsurfer: actually i've moved this component, dunno why is it implemented here again ?

flexsurfer (changes requested):
please don't merge it yet, im not sure about contact-list

src/status_im/ui2/screens/common/contact_list/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+   [quo2/text {:size   :paragraph-2
+               :weight :medium
+               :style  {:color colors/neutral-50}} title]])
+
+(defn search-input []
+  [rn/text-input {:placeholder "Search..."
+                  :style       {:height             32
+                                :padding-horizontal 20
+                                :margin-vertical    12}
+                  :on-change   (fn [e] (reset! query (.-text ^js (.-nativeEvent ^js e))))}])
```
ilmotta: Are you sure you need to add all the type hints like this? I believe it's enough to only add it to the function argument. So something like this seems better:

```clojure
(fn [^js e] (reset! query (.. e -nativeEvent -text)))
```

You could also use `oops.core/oget`.

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: oh ok i see, its list not item, but still i moved this already

src/status_im/ui2/screens/common/contact_list/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+
+(defn contact-list [{:keys [search?] :as data}]
+  (let [contacts (rf/sub [:contacts/active])
+        contacts (prepare-contacts contacts)]
```
flexsurfer: we shouldn't have any logic or computation in views, so i've moved this to subscription

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: image
we have a bunch of different variants, it would be great to have it documented

flexsurfer: i like `(oget e "nativeEvent.text")`

OmarBasem:
> please don't merge it yet, im not sure about contact-list

@flexsurfer I created a `contact-list` component to be a reusable `section-list` component for contacts that includes all its logic (the list component,  prepare-contacts method, etc) as it is used in multiple places (for example, messages home contacts, add members in group details, manage members in group details)

flexsurfer:
> > please don't merge it yet, im not sure about contact-list
> 
> @flexsurfer I created a `contact-list` component to be a reusable `section-list` component for contacts that includes all its logic (the list component, prepare-contacts method, etc) as it is used in multiple places (for example, messages home contacts, add members in group details, manage members in group details)

yeah, thats ok, now we need to understand how to merge our changes :)

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: i don't see search-input in figma, why was it implemented ?

flexsurfer: oh i see its optional, and there is no search on home page , ok

OmarBasem:
@flexsurfer I will fix the few things mentioned above, then merge this PR, then you can move it to status-im2

src/status_im/ui2/screens/common/contact_list/view.cljs
ilmotta: > we have a bunch of different variants, it would be great to have it documented

Great point @flexsurfer, I totally agree. This is a common discussion in PRs.

flexsurfer:
i can't find search control for contacts in the figma in group details, and seems like lists are different, there is sections not by first letter

src/status_im/ui2/screens/chat/actions.cljs
OmarBasem: Yes! (exit the group details screen when leaving the group)

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -45,8 +60,49 @@
...
+          (i18n/label :t/save)]]]))])
+
 (defn group-details []
-  (let [{:keys [admins chat-id chat-name color public?]} (rf/sub [:chats/current-chat])
+  (let [{:keys [admins chat-id chat-name color public? muted] :as group} (rf/sub [:chats/current-chat])
         members           (rf/sub [:contacts/current-chat-contacts])
         sectioned-members (prepare-members members)
```
qfrank: as "we shouldn't have any logic or computation in views, should be moved to subscription", maybe this one need refactor too? it's not relate to your PR changes, but would u like do something? 😉

src/status_im/ui2/screens/chat/actions.cljs
OmarBasem: Actually can be replaced by just `add-divider?    extra-data`

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
OmarBasem: @ibrkhalil sorry I don't understand what you mean

OmarBasem: @ilmotta so both these statements does the same thing?

src/status_im/ui2/screens/common/contact_list/view.cljs
OmarBasem: @ibrkhalil I am not sure what do you mean? It is iterating over contacts count.

OmarBasem: @ilmotta so in Clojure it is a bad practice to update atoms in a loop?

src/status_im/ui2/screens/chat/group_details/view.cljs
OmarBasem: @ilmotta I am not sure what you mean?

src/status_im/ui2/screens/chat/group_details/view.cljs
OmarBasem: @qfrank not sure what do you mean? This is only a subscription

qfrank: i mean `prepare-members` just like `prepare-contacts` @OmarBasem

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
ibrkhalil: Hey @OmarBasem 
I mean `#(-> % (not= public-key))` is more readable as `#(not= % public-key)`

src/status_im/contact/core.cljs
```diff
@@ -151,3 +151,8 @@
    :mutual-contact-enabled?
    enabled?
    nil))
+
+(fx/defn search-query
+  {:events [:contacts/search-query]}
```
flexsurfer: better to name it like set-search-query or change-search-query or update-search-query , cc @ilmotta

src/status_im/contact/chat.cljs
```diff
@@ -30,3 +30,5 @@
   {:events       [:contact.ui/pinned-messages-pressed]}
   [cofx public-key]
   (chat/navigate-to-user-pinned-messages cofx public-key))
+
+
```
flexsurfer: too many lines :)

src/status_im/ui2/screens/chat/home.cljs
```diff
@@ -0,0 +1,373 @@
+(ns status-im.ui2.screens.chat.home
```
flexsurfer: this file doesn't exist anymore , see https://github.com/status-im/status-mobile/blob/develop/src/status_im2/contexts/chat/home/view.cljs

src/status_im/ui2/screens/common/contact_list/view.cljs
```diff
@@ -0,0 +1,27 @@
...
+  (let [contacts (rf/sub [:contacts/filtered-active-sections])]
+    [rn/section-list
+     {:key-fn                         :title
+      :sticky-section-headers-enabled false
+      :sections                       contacts
+      :render-section-header-fn       contacts-section-header
+      :content-container-style        {:padding-bottom 120}
+      :sticky-header-indices [0]
+      :render-fn                      (fn [item]
+                                        [contact-list-item/contact-list-item item data])}]))
```
flexsurfer: better to use `render-data` property

```clojure
:render-data data
:render-fn  contact-list-item/contact-list-item 
```

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -1,49 +1,71 @@
 (ns status-im2.common.contact-list-item.view
-  (:require [utils.re-frame :as rf]
-            [react-native.core :as rn]
-            [quo2.core :as quo]
+  (:require [quo2.foundations.typography :as typography]
```
flexsurfer: i know it was i tough one, but please rollback changes like 

[quo/text {:weight :semi-bold} display-name]

[quo/user-avatar

[quo/icon

also in status-im2 we use quo alias not quo2

src/status_im2/subs/chat/chats.cljs
```diff
@@ -411,3 +411,5 @@
  :<- [:chat/inputs-with-mentions]
  (fn [[chat-id cursor]]
    (get cursor chat-id)))
+
+
```
flexsurfer: too many lines

src/status_im2/subs/contact.cljs
```diff
@@ -118,21 +118,21 @@
   (let [{:keys [nickname three-words-name ens-name]} names]
     (or
      (when ens-name
-       (string/includes? (string/lower-case (str ens-name)) search-filter))
-     (string/includes? (string/lower-case three-words-name) search-filter)
+       (str/includes? (str/lower-case (str ens-name)) search-filter))
```
flexsurfer: better to use `string` alias , because there is a `str` functions in cljs.core

src/status_im/ui2/screens/chat/group_details/view.cljs
ilmotta: Hey @OmarBasem, `constantly` is a core Clojure function and some people like to use it to communicate that a function doesn't care about its arguments and the return value is always the same. It has its fair share of usages in other contexts too.

This is just FYI stuff, I don't even think it's better/worse.

```clojure
(constantly [actions/group-details-actions group])
```

src/status_im2/subs/contact.cljs
flexsurfer: probably better to document as well, cc @ilmotta

src/status_im/contact/core.cljs
ilmotta: :+1: to @flexsurfer, this event name makes other devs think it's going to fetch/search for something.

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -5,60 +5,112 @@
...
+(defn contact-requests-sheet [group]
+  [:f>
+   (fn []
+     (let [{window-height :height} (rn/use-window-dimensions)
+           safe-area (safe-area/use-safe-area)]
+       [rn/view {:style {:height (- window-height (:top safe-area))}}
+        [rn/touchable-opacity
+         {:on-press #(rf/dispatch [:bottom-sheet/hide])
+          :style
+          {:background-color (colors/theme-colors colors/neutral-10 colors/neutral-80)
```
ilmotta: Given it's new code, we could try to adhere to the new guidelines and extract styles to a separate file.

src/status_im2/common/home/actions/view.cljs
```diff
@@ -146,13 +145,24 @@
           :sub-label nil
           :chevron?  false}))
 
-(defn notifications-entry []
-  (entry {:icon      :i/notifications
-          :label     (i18n/label :t/notifications)
+;; Below TODOs are from OmarBasem
```
ilmotta: I guess we can stick to the convention in the guidelines, just write `TODO(<YOUR USERNAME>)` in each one. If you don't want you name to appear in the alert, to follow the guidelines we could start doing this:

```clojure
;; TODO(OmarBasem): Requires design input.
:on-press #(js/alert "To be implemented")
```

src/status_im2/contexts/chat/home/chat_list_item/style.cljs
```diff
@@ -1,7 +1,7 @@
 (ns status-im2.contexts.chat.home.chat-list-item.style
   (:require [quo2.foundations.colors :as colors]))
 
-(def container
+(defn container []
```
ilmotta, line 4: This function could be converted back to `def` because it's all static data.

src/status_im2/subs/contact.cljs
ilmotta: There are some tips on the internet on how automate this kind of check, which is very common in Clojure circles.

For the time being I'm accumulating notes to create a single PR to update the guidelines, just added this one @flexsurfer.

src/status_im2/subs/contact.cljs
```diff
@@ -251,4 +251,32 @@
...
-           contacts)))
\ No newline at end of file
+           contacts)))
+
+(re-frame/reg-sub
+ :contacts/filtered-active-sections
+ :<- [:contacts/active]
+ :<- [:contacts/search-query]
+ (fn [[contacts query]]
+   (let [data (atom {})]
```
ilmotta: I think we should almost never see atoms being used as temporary data for processing, because this is too non-idiomatic and performs badly. @OmarBasem, could you at least rewrite this function to use a `reduce` function?

Additionally, layer-3 subscriptions should have tests, they're critical to the correct operation of the app. I know this is not in the guidelines, but it's up to us to stop growing the tech debt of not testing what touches the app db. If it's green code I'm going to leave a comment ;)

src/status_im/ui2/screens/common/contact_list/view.cljs
ilmotta: @flexsurfer, could you expand on why :render-data is better? It's not clear to me. Is it because of what wrap-render-fn does behind the scenes?

src/status_im/ui2/screens/common/contact_list/view.cljs
ilmotta: The correct answer is it depends, as usual, but there are almost always better ways. We have to remember that atoms were conceived primarily to manage shared state, and as such, they are capable of a lot of cool stuff, but this also means they're a big bazooka for simple data processing.

Temporary state for data processing in pure functions can almost always be replaced by an *accumulator* in a reducing function (just to name one among several approaches). Using doseq to loop over indexes is quite imperative, and Clojure gives us tools to write more *functional* code.

So something along these lines:

```clojure
;; very imperative
(let [state (atom {})]
  (doseq [i (range (count contacts))
          :let [contact (nth contacts i)]]
    (swap! state ...)))

;; probably better
(reduce (fn [result contact]
          (assoc result ...))
        {}
        contacts)
```

src/status_im2/subs/contact.cljs
```diff
@@ -251,4 +251,32 @@
...
+
+(re-frame/reg-sub
+ :contacts/group-members-sections
+ :<- [:contacts/current-chat-contacts]
+ (fn [members]
+   (let [admins  (filter :admin? members)
+         online  (filter #(and (not (:admin? %)) (:online? %)) members)
+         offline (filter #(and (not (:admin? %)) (not (:online? %))) members)]
+     (vals (cond-> {}
+             (seq admins) (assoc :owner {:title "Owner" :data admins})
```
flexsurfer: i18n

src/status_im2/subs/contact.cljs
```diff
@@ -251,4 +251,32 @@
...
+     (vals @data))))
+
+(re-frame/reg-sub
+ :contacts/group-members-sections
+ :<- [:contacts/current-chat-contacts]
+ (fn [members]
+   (let [admins  (filter :admin? members)
+         online  (filter #(and (not (:admin? %)) (:online? %)) members)
+         offline (filter #(and (not (:admin? %)) (not (:online? %))) members)]
+     (vals (cond-> {}
```
flexsurfer, line 280: you could do it in one go in reduce. now you have 3 filters for the same list and also 3 conditions

src/status_im/ui2/screens/chat/components/contact_item/view.cljs
ilmotta: > @ilmotta so both these statements does the same thing?

Exactly @OmarBasem, it's up to the person to decide if they like the second way. I prefer to usually avoid `partial` and use the literal function `#()` form, but `partial` is a good tool to have at hand ;)

src/status_im/ui2/screens/common/contact_list/view.cljs
OmarBasem: @flexsurfer what is `render-data` property?

src/status_im/group_chats/core.cljs
```diff
@@ -21,89 +21,89 @@
...
 (fx/defn join-chat
   {:events [:group-chats.ui/join-pressed]}
   [_ chat-id]
-  {::json-rpc/call [{:method     "wakuext_confirmJoiningGroup"
-                     :params     [chat-id]
+  {::json-rpc/call [{:method      "wakuext_confirmJoiningGroup"
+                     :params      [chat-id]
                      :js-response true
-                     :on-success #(re-frame/dispatch [:chat-updated %])}]})
+                     :on-success  #(re-frame/dispatch [:chat-updated %])}]})
```
J-Son89, line 50: should it not be `rf/dispatch` ??

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -5,60 +5,112 @@
             [quo2.core :as quo2]
```
J-Son89, line 5: this can be updated to `:as quo`

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -5,60 +5,112 @@
...
-            (seq online) (assoc :online {:title "Online" :data online})
-            (seq offline) (assoc :offline {:title "Offline" :data offline})))))
-
 (defn contacts-section-header [{:keys [title]}]
   [rn/view {:style {:padding-horizontal 20 :border-top-width 1 :border-top-color colors/neutral-20 :padding-vertical 8 :margin-top 8}}
    [quo2/text {:size   :paragraph-2
                :weight :medium
                :style  {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)}} title]])
 
+(def added (reagent/atom ()))
```
J-Son89, line 55: this seems like a strange way to add state. I'm not quite sure how it works in clojurescript with when this gets mounted etc but it seems a bit odd that you add an atom here and not defined as a let variable in the render of the component using it. e.g contact-requests-sheet..

or is this approach correct?? 🤔

J-Son89: @ilmotta wdyt? 🤔

src/status_im/ui2/screens/chat/messages/message.cljs
```diff
@@ -41,7 +41,8 @@
             [quo2.components.avatars.user-avatar :as user-avatar]
             [quo2.components.markdown.text :as text]
             [status-im.utils.utils :as utils]
-            [status-im2.contexts.chat.home.chat-list-item.view :as home.chat-list-item])
+            [status-im2.contexts.chat.home.chat-list-item.view :as home.chat-list-item]
+            [quo2.core :as quo2])
```
J-Son89, line 45: `:as quo`

J-Son89: or maybe we just `:as quo2` for now and it will be easier to a replace all once we finally migrated over.

@flexsurfer wdyt? 
I see you started using `:as quo` for quo2 stuff in the code base. This might get confusing while we're supporting both :D

src/status_im2/common/home/actions/view.cljs
J-Son89: Ideally we should just create issues for all of these and it should not be knowledge of the dev for what the TODO is.

e.g.
`TODO: #13117 ` and the dev who creates the TODO should make sure the issue is there and create it if not.

src/status_im2/common/home/actions/view.cljs
```diff
@@ -146,13 +145,24 @@
...
-          :chevron?  true}))
+          :sub-label nil
+          :chevron?  false}))
+
+(defn notifications-entry [add-divider?]
+  (entry {:icon         :i/notifications
+          :label        (i18n/label :t/notifications)
+          :on-press     #(js/alert "TODO: to be implemented, requires design input")
+          :danger?      false
+          :sub-label    "All messages" ; TODO: placeholder
```
J-Son89, line 163: remove TODO?  or else give it more context?

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -1,49 +1,71 @@
...
-                                                    (rf/dispatch [:bottom-sheet/show-sheet
-                                                                  {:content (fn [] [actions/actions item])}]))})
-     [quo/user-avatar {:full-name         display-name
-                       :profile-picture   photo-path
-                       :status-indicator? true
-                       :online?           true
-                       :size              :small
-                       :ring?             false}]
+        photo-path   (when (seq images) (rf/sub [:chats/photo-path public-key]))
+        current-pk   (rf/sub [:multiaccount/public-key])]
```
J-Son89, line 39: should this also be `current-pub-key`. I see you are using that in another file

src/status_im2/common/contact_list_item/view.cljs
J-Son89: imo it's confusing that we're using quo (for quo2) in status_im2 and quo for quo in status_im.

We should just keep it simple :)

J-Son89 (approved):

src/status_im/ui2/screens/chat/messages/message.cljs
flexsurfer: yes, when we move stuff to `status-im2` quo alias should be used, in old namespaces there are still might be mixed usage of new and old components, so that's fine to use quo2 there

src/status_im/ui2/screens/chat/group_details/view.cljs
flexsurfer: good point, it should be a subscription, or if it really super local state then it could be in let, but it never should be defined like this with `def` , yeah might be some super special cases, but they are rare

J-Son89: Good to know! Thanks @flexsurfer :D

ilmotta: It all depends on what we want. If we want to create a reusable component, then a local state inside the function is needed, otherwise multiple instances in the same screen will compete for the same shared state and this is a recipe for disaster.

But in a namespace such as this one, having a shared state for the entire screen functions is more acceptable IMO, it's correct technically speaking.

But just like you guys, I also prefer local state inside functions in general.

src/status_im2/common/home/actions/view.cljs
ilmotta: I like this approach, I just wouldn't call it ideal. To be honest, I actually prefer to have less and less TODOs in the code and just create issues with enough context that the developer playing them knows what to do, so there's no need for contextualizing them in the code.

I'm not a big fan of littering the code with TODO comments, it almost feels like a half-a** project management tool (forgive my language please).

src/status_im2/common/contact_list_item/view.cljs
OmarBasem: Everything is confusing in this PR :/ :D

src/status_im2/contexts/chat/home/chat_list_item/style.cljs
OmarBasem: I personally prefer to always use defn for styling definitions since we may need to pass some args to it at any time later. I know we can then change it, but to be consistent. Just my opinion, what do you think?

src/status_im2/subs/contact.cljs
ilmotta: @OmarBasem, I don't know if it helps in the midst of so many comments, but this is the only outstanding thread left for me to approve your PR.

I've noticed that lots of layer-3 subscriptions don't have tests yet, and that it might not be obvious how to test them given they're registered as anonymous functions in an internal state managed by re-frame. Please, ignore my comment about testing them. I think this kind of test requires a separate PR example in order to raise awareness and share knowledge, so it would be too much effort for this PR, which is already quite busy ;)

OmarBasem: You are right @ilmotta. 
This PR has gotten 80+ reviews/comments :'(

src/status_im/group_chats/core.cljs
OmarBasem: Did not write this code, this is just formatting
We still need to update all `re-frame` occurrences to `rf`, so will open an issue for that

src/status_im/ui2/screens/chat/group_details/view.cljs
OmarBasem: I think Andrey said we are using quo2 in status-im?

src/status_im/contact/core.cljs
flexsurfer: you need to rename event as well `:contacts/set-search-query`

src/status_im/contact/core.cljs
```diff
@@ -151,3 +151,8 @@
    :mutual-contact-enabled?
    enabled?
    nil))
+
+(fx/defn set-search-query
+  {:events [:contacts/set-search-query]}
+  [{:keys [db] :as cofx} value]
+  {:db (assoc db :contacts/set-search-query value)})
```
flexsurfer: only event :( key in the db should be without set, so event `:contacts/set-search-query` sets the value for `:contacts/search-query`

src/status_im2/subs/root.cljs
```diff
@@ -140,6 +140,7 @@
 (reg-root-key-sub :contacts/new-identity :contacts/new-identity)
 (reg-root-key-sub :group/selected-contacts :group/selected-contacts)
 (reg-root-key-sub :contacts/blocked-set :contacts/blocked)
+(reg-root-key-sub :contacts/set-search-query :contacts/set-search-query)
```
flexsurfer: please rollback this to `:contacts/search-query`

src/status_im2/contexts/chat/home/chat_list_item/style.cljs
flexsurfer: in 99% of cases, it will never change, and also by using def you can already see from view if the style is dynamic or not, so you don't need to go inside to see that, so better to use def if it's not dynamic

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -5,60 +5,97 @@
...
+        [rn/text-input {:placeholder (str (i18n/label :t/search) "...")
+                        :style       {:height             32
+                                      :padding-horizontal 20
+                                      :margin-vertical    12}
+                        :on-change   (fn [e] (rf/dispatch [:contacts/set-search-query (oget e "nativeEvent.text")]))}]
+        [contact-list/contact-list {:icon  :check
+                                    :group group
+                                    :added added
+                                    :search? true}]
+        [rn/view {:style (style/bottom-container safe-area)}
```
flexsurfer: bottom-sheet should take care of that, you don't need to use here safe-area, if it looks wrong means bottom sheet is broken, but new version is being developed

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: render-fn shouldn't be anonymous because it will be recreated each time on every contacts change, and for list it will be a new function each time, so render-fn should be a defined function like `:render-fn   contact-list-item/contact-list-item` but sometimes you want to pass additional data to render function that's why we added additional property `:render-data` so you can pass data safely to render function, it will be a fourth argument in the render function

src/status_im/ui2/screens/common/contact_list/view.cljs
```diff
@@ -0,0 +1,26 @@
+(ns status-im.ui2.screens.common.contact-list.view
+  (:require [react-native.core :as rn]
+            [quo2.foundations.colors :as colors]
+            [status-im2.common.contact-list-item.view :as contact-list-item]
+            [quo2.core :as quo2]
+            [utils.re-frame :as rf]))
+
+(defn contacts-section-header [{:keys [title]}]
```
flexsurfer, line 7: please replace it by function from src/status_im2/contexts/chat/home/view.cljs

src/status_im2/contexts/chat/home/view.cljs
```diff
@@ -54,12 +54,7 @@
       [:<>
```
flexsurfer, line 51: its not possible to comment unchanged code, so pointer is wrong, but we need to move contacts-section-header from here to contact-list

src/status_im/ui2/screens/chat/group_details/style.cljs
```diff
@@ -18,3 +18,23 @@
...
+   :margin-left      20
+   :width            32
+   :height           32
+   :border-radius    10
+   :justify-content  :center
+   :align-items      :center
+   :margin-bottom    24})
+
+(defn bottom-container [safe-area]
+  {:position           :absolute
```
flexsurfer, line 33: what's the point of using :absolute ?

src/status_im/ui2/screens/common/contact_list/view.cljs
flexsurfer: it should use component from quo

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -38,12 +58,9 @@
          (when added?
            [rn/view {:style {:margin-left 5 :margin-top 4}}
             [quo/icon :i/contact {:no-color true :size 12 :color (colors/theme-colors colors/primary-50 colors/primary-60)}]]))]
-      [quo/text {:style {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
-       (utils.address/get-shortened-address public-key)]]
+      [text/text {:size  :paragraph-1
```
flexsurfer: quo/text

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -1,28 +1,48 @@
 (ns status-im2.common.contact-list-item.view
-  (:require [utils.re-frame :as rf]
-            [react-native.core :as rn]
-            [quo2.core :as quo]
-            [quo2.foundations.colors :as colors]
-            [status-im2.common.contact-list-item.style :as style]
-            [utils.address :as utils.address]
-            [status-im2.common.home.actions.view :as actions]))
+  (:require [quo2.foundations.colors :as colors]
+            [quo.react-native :as rn]
```
flexsurfer: react-native.core

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -1,28 +1,48 @@
...
+            [status-im.utils.utils :as utils]
+            [quo.platform :as platform]
```
flexsurfer: react-native.platform

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -1,28 +1,48 @@
...
-  (:require [utils.re-frame :as rf]
-            [react-native.core :as rn]
-            [quo2.core :as quo]
-            [quo2.foundations.colors :as colors]
-            [status-im2.common.contact-list-item.style :as style]
-            [utils.address :as utils.address]
-            [status-im2.common.home.actions.view :as actions]))
+  (:require [quo2.foundations.colors :as colors]
+            [quo.react-native :as rn]
+            [status-im.utils.utils :as utils]
```
flexsurfer: [utils.address :as utils.address]

src/status_im2/common/contact_list_item/view.cljs
```diff
@@ -38,12 +58,9 @@
          (when added?
            [rn/view {:style {:margin-left 5 :margin-top 4}}
             [quo/icon :i/contact {:no-color true :size 12 :color (colors/theme-colors colors/primary-50 colors/primary-60)}]]))]
-      [quo/text {:style {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
-       (utils.address/get-shortened-address public-key)]]
+      [text/text {:size  :paragraph-1
+                  :style {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)}}
+       (utils/get-shortened-address public-key)]]
```
flexsurfer: (utils.address/get-shortened-address

ilmotta (approved):

qfrank (approved):

src/status_im/ui2/screens/chat/group_details/style.cljs
OmarBasem: Because its position is absolute

flexsurfer (approved):

src/status_im/ui2/screens/chat/group_details/style.cljs
flexsurfer: yeah just wondering whats the reason, why its not just regular flex layout

src/status_im/ui2/screens/chat/group_details/view.cljs
```diff
@@ -5,60 +5,97 @@
...
 (defn contacts-section-header [{:keys [title]}]
   [rn/view {:style {:padding-horizontal 20 :border-top-width 1 :border-top-color colors/neutral-20 :padding-vertical 8 :margin-top 8}}
    [quo2/text {:size   :paragraph-2
                :weight :medium
                :style  {:color (colors/theme-colors colors/neutral-50 colors/neutral-40)}} title]])
 
+(def added (reagent/atom ()))
+
+(defn contact-requests-sheet [group]
+  [:f>
```
flexsurfer: better to move atom here 

```
(defn contact-requests-sheet [group]
  (let [added (reagent/atom ())]
    (fn []
       [:f>
          (fn []
              (let [{window-height :height}
```

OmarBasem:
@flexsurfer As mentioned above, "Regarding QA, since the functionality is not yet complete, so will skip the QA for now, and will run the final PR through QA when the functionality is complete."

src/status_im/ui2/screens/chat/group_details/view.cljs
OmarBasem: Because there is a list, and the bottom container that includes the button needs to always be floating on top of the list at the bottom of the screen.

Issue #14415, opened 2022-11-22
Mark notifications as read


ilmotta:
# Feature Issue

### User Story

>As a user,
>I want easily see what requires my attention in the Activity Center,
>So that I can promptly act on the most important notifications to me.

### Description

Nowadays, the new Activity Center is using the old Notification Center behavior for marking notifications as read. We should implement the new rules:

1. With CTA (call to action) - should stay *unread* until the user acts on it.
2. Without CTA - should loose *unread indicator* and go down the list after the user opens the AC or the same tab again.

### Expected behavior about ordering

Notifications should be ordered by timestamp descending and ID descending. This is the current behavior in status-go. Unread notifications should not be grouped at the top because the user can use the read/unread toggle button to easily find which notifications require their attention.

### Expected behavior for non-CTA notifications

Non-CTA notifications should be automatically marked as read if they’re visible on screen, according to the current active tab (which dictates which types of notifications are displayed). If the user is in the special All tab then notifications of any type should be taken in consideration. For example, if the user opens the Activity Center, unread notifications are displayed with blue dots, but once the user switches to a different tab or go to another screen, when they open the Activity Center in the same tab, non-CTA notifications that were unread won’t have the unread blue dot anymore and they’ll be marked as read in the local database.

ilmotta:
This issue is blocked, waiting for the PO to confirm the following behavior:

- Notifications should be ordered by timestamp descending and ID descending. This is the current behavior in status-go. Unread notifications should **not** be grouped at the top because the user can use the read/unread toggle button to easily find which notifications require their attention.

- Non-CTA notifications should be automatically marked as read (visible or not, all of them) according to the current active tab (which dictates which types of notifications are displayed). If the user is in the special *All tab*, then notifications of any type should be taken in consideration. For example, if the user opens the Activity Center, unread notifications are displayed with blue dots, but once the user switches to a different tab or go to another screen, when they open the Activity Center in the same tab, non-CTA notifications that were unread won’t have the unread blue dot anymore and they’ll be marked as read in the local database.

ilmotta:
Update (2023-01-19): After a new round of reviews about how should notifications be marked as read/unread, designers, PO and devs from mobile/desktop aligned the following:

- Designers will soon update Figma with the most up-to-date specs so that developers can start the implementation work.
- Developers will try to keep written specs about all things related to the AC, in order to complement the Figma specs.
[...]

PR #14439, opened 2022-11-24
Mark CTA notifications as read

ilmotta:
Partially implements https://github.com/status-im/status-mobile/issues/14415

### Summary

This PR implements the new app behavior to mark notifications with CTA (call to action) in the Activity Center as `read`. CTA notifications are only ever marked as read if the user first interacts with them, therefore, leaving/entering the Activity Center has no side-effect.

[untitled.webm](https://user-images.githubusercontent.com/46027/203836793-4d92d379-e40c-4eb0-8a0c-2cdeb9b6a3c7.webm)

### Out of scope

- Changing the structure of the files according to new guidelines.
- Marking non-CTA notifications as `read` is more complex according to new UX and will be left for another PR.
- Sorting unread notifications at the top requires changes in status-go as well, so it is also going to be implemented in a separate PR.

status: ready

### DIFF
--- a/src/status_im/activity_center/core_test.cljs
+++ b/src/status_im/activity_center/core_test.cljs
@@ -9,13 +9,105 @@
             [status-im.test-helpers :as h]
             [status-im2.setup.config :as config]))
 
+(def notification-id "0x1")
+
 (defn setup []
   (h/register-helper-events)
   (rf/dispatch [:setup/app-started]))
 
-;;;; Contact verification
+(defn test-log-on-failure
+  [{:keys [before-test notification-id event action]}]
+  (rf-test/run-test-sync
+   (setup)
+   (h/using-log-test-appender
+    (fn [logs]
+      (when before-test
+        (before-test))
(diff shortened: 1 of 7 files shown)


src/status_im/activity_center/core_test.cljs
```diff
@@ -9,13 +9,105 @@
...
+                            {:all    {:cursor "" :data [{:id   notification-id
+                                                         :read false
+                                                         :type types/one-to-one-chat}]}
+                             :unread {:cursor "" :data []}}}]
+         (rf/dispatch [:test/assoc-in [:activity-center]
+                       {:notifications notifications
+                        :filter        {:type   types/one-to-one-chat
+                                        :status :all}}])
+
+         (rf/dispatch [:activity-center.notifications/mark-as-read "0x666"])
```
J-Son89, line 54: is ""0x666" some particular value? maybe putting it in a named `def` would explain what it is.

J-Son89 (approved):

PR #14440, opened 2022-11-24
chore: fix issues with e2e tests after refactors

J-Son89:
E2e tests were broken after some recent changes. 

IT seems as though it is needed to add `accessible true` to a view component for it to be treated as an accessibility element so that Appium can find that element on the page.

### DIFF
--- a/src/quo2/components/counter/counter.cljs
+++ b/src/quo2/components/counter/counter.cljs
@@ -20,7 +20,7 @@
 (defn counter
   "type:    default, secondary, grey, outline
    value:   integer"
-  [{:keys [type override-text-color override-bg-color style]} value]
+  [{:keys [type override-text-color override-bg-color style accessibility-label]} value]
   (let [type       (or type :default)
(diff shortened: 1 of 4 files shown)


qoqobolo:
Hi @J-Son89 !
I'm also trying to figure out why the accessibility IDs you added don't work and the tests keep failing.
What I can see now is that the IDs are added in the code, but I can't see them in Appium (checking https://github.com/status-im/status-mobile/commit/25e8f534a49bb3db75413213e43309e536a3e965 build). 

Screenshot 2022-11-25 at 12 20 01
Screenshot 2022-11-25 at 12 18 47

This is what we would expect to see in Appium when an ID is actually added

Screenshot 2022-11-25 at 12 27 25

Unfortunately, I cannot give a more precise reason, but maybe it will be useful for you.
Can you check, please? 

And ping us any time you need to check in Appium or restart tests.

qoqobolo (approved):

PR #14281, opened 2022-11-02
feat: add scrolling in community

J-Son89:
Fixes https://github.com/status-im/status-mobile/issues/13605

Implemented scrolling functionality on Communities home page.

This p.r also adjusts the display of the community overview page so the top bar is no longer just white - backgrund image is behind clock and battery etc.

Also had to adjust the TopBar component to handle  both images and icons. With images we can set different sizes for images to match designs in Figma.

Designs for scrolling: https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1506%3A101672

## Steps to test
Open Status
Open communities tab
If you are a member of a community you can go to Discover page and then view that community

Scroll on the page :)

If you are not a member of a community you can click the '+' button on Communities home page.

then click "Communities (alpha)" option and select Status community.
Click "Follow" of the bottom of the screen. 

Now go back to COmmunities Home and go to Communities Discover page. You should be able to select status community. 
Click into it and start scrollling.

PR #14444, opened 2022-11-25
fix: adjust community cards on discover page

### LINKED ISSUE #14443: fix discover communities screen stretching too much
# Bug Report

Communities discover page should look like this:
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7202%3A475069&t=fc4U3iZRVexnAn93-0

After recent changes they now look like:

Screenshot 2022-11-25 at 13 38 41

Fix this

J-Son89:
Fixes https://github.com/status-im/status-mobile/issues/14443

Adjusted the discover card to use width relative to it's view size and margin handled outside. Some other refactors too.

Screenshot 2022-11-25 at 13 37 00

Screenshot 2022-11-25 at 13 37 07

### DIFF
--- a/src/quo2/components/community/community_card_view.cljs
+++ b/src/quo2/components/community/community_card_view.cljs
@@ -6,15 +6,11 @@
 
 (defn community-card-view-item
   [{:keys [name description locked
-           status tokens cover tags featured]} on-press]
+           status tokens cover tags width]} on-press]
   [rn/touchable-opacity {:on-press on-press}
-   [rn/view {:style (merge (style/community-card 20)
-                           {:margin-bottom 16}
-                           (if featured
-                             {:margin-right 12}
-                             {:flex              1
-                              :margin-horizontal 20}))}
-    [rn/view {:style    {:height        230
+   [rn/view {:style (style/community-card 20)}
+    [rn/view {:style    {:width         width
+                         :height        230
                          :border-radius 20}
               :on-press on-press}
      [rn/view
--- a/src/status_im2/contexts/communities/discover/view.cljs
+++ b/src/status_im2/contexts/communities/discover/view.cljs
@@ -11,8 +11,6 @@
             [status-im.react-native.resources :as resources]
             [status-im.ui.screens.communities.community :as community]))
 
-(def view-type (reagent/atom :card-view))
-
 (def mock-community-item-data  ;; TODO: remove once communities are loaded with this data.
   {:data {:community-color "#0052FF"
           :status          :gated
@@ -31,24 +29,23 @@
                              :tag-label (i18n/label :t/podcasts)
                              :resource  (resources/get-image :podcasts)}]}})
 
-(defn render-fn [featured? width]
-  (fn
-    [community-item]
-    (let [item (merge community-item
-                      (get mock-community-item-data :data)
-                      {:featured featured?})]
-      (if (= @view-type :card-view)
-        [quo/community-card-view-item (assoc item :width width) #(rf/dispatch [:navigate-to :community-overview item])]
-        [quo/communities-list-view-item
-         {:on-press      (fn []
-                           (rf/dispatch [:communities/load-category-states (:id item)])
-                           (rf/dispatch [:dismiss-keyboard])
-                           (rf/dispatch [:navigate-to :community {:community-id (:id item)}]))
-          :on-long-press #(rf/dispatch [:bottom-sheet/show-sheet
-                                        {:content (fn []
+(defn render-fn
+  [community-item _ _ {:keys [featured? width view-type]}]
+  (let [item (merge community-item
+                    (get mock-community-item-data :data)
+                    {:featured featured?})]
+    (if (= view-type :card-view)
+      [quo/community-card-view-item (assoc item :width width) #(rf/dispatch [:navigate-to :community-overview item])]
+      [quo/communities-list-view-item
+       {:on-press      (fn []
+                         (rf/dispatch [:communities/load-category-states (:id item)])
+                         (rf/dispatch [:dismiss-keyboard])
+                         (rf/dispatch [:navigate-to :community {:community-id (:id item)}]))
+        :on-long-press #(rf/dispatch [:bottom-sheet/show-sheet
+                                      {:content (fn []
                                                   ;; TODO implement with quo2
-                                                    [community/community-actions item])}])}
-         item]))))
+                                                  [community/community-actions item])}])}
+       item])))
 
 (defn screen-title []
   [rn/view
@@ -80,53 +77,64 @@
                                         colors/neutral-50
                                         colors/neutral-40)}]])
 
-(defn featured-list [communities view-size]
-  [rn/view {:style {:flex-direction :row
-                    :overflow :hidden
-                    :width "100%"
-                    :margin-bottom 24}
-            :on-layout #(swap! view-size (fn []
-                                           (oops/oget % "nativeEvent.layout.width")))}
-   (when-not (= @view-size 0)
-     [rn/flat-list
-      {:key-fn                            #(str "featured" (% :id))
-       :horizontal                        true
-       :keyboard-should-persist-taps      :always
-       :shows-horizontal-scroll-indicator false
-       :separator                  [rn/view {:margin-right 12}]
-       :data                              communities
-       :render-fn                         (render-fn true @view-size)}])])
+(defn featured-list [communities view-type]
+  (let [view-size (reagent/atom 0)]
+    (fn []
+      [rn/view {:style {:flex-direction :row
+                        :overflow :hidden
+                        :width "100%"
+                        :margin-bottom 24}
+                :on-layout #(swap! view-size
+                                   (fn []
+                                     (oops/oget % "nativeEvent.layout.width")))}
+       (when-not (= @view-size 0)
+         [rn/flat-list
+          {:key-fn                            :id
+           :horizontal                        true
+           :keyboard-should-persist-taps      :always
+           :shows-horizontal-scroll-indicator false
+           :separator                         [rn/view {:width 12}]
+           :data                              communities
+           :render-fn                         render-fn
+           :render-data                       {:featured? true
+                                               :width     @view-size
+                                               :view-type view-type}}])])))
 
-(defn other-communities-list [communities]
+(defn other-communities-list [communities view-type]
+  (println view-type)
   [rn/flat-list
    {:key-fn                            :id
     :keyboard-should-persist-taps      :always
     :shows-horizontal-scroll-indicator false
     :separator                 [rn/view {:margin-bottom 16}]
     :data                              communities
-    :render-fn                         (render-fn false "100%")}])
+    :render-fn                        render-fn
+    :render-data              {:featured? false
+                               :width "100%"
+                               :view-type view-type}}])
 
 (defn discover []
-  (let [communities (rf/sub [:communities/communities])
-        communities-count (reagent/atom {:value (count communities)})
-        ;;TODO move sorting to subscription
-        sorted-communities (sort-by :name communities)
-        view-size (reagent/atom 0)]
-    [safe-area/consumer
-     (fn []
-       [rn/view {:style {:margin-left     20
-                         :margin-right    20
-                         :flex             1
-                         :background-color (colors/theme-colors
-                                            colors/white
-                                            colors/neutral-90)}}
-        [quo/button {:icon     true
-                     :type     :grey
-                     :size     32
-                     :style    {:margin-vertical 12}
-                     :on-press #(rf/dispatch [:navigate-back])}
-         :i/close]
-        [screen-title]
-        (featured-communities-header (:value  @communities-count))
-        [featured-list communities view-size]
-        [other-communities-list sorted-communities]])]))
+  (let  [view-type (reagent/atom :card-view)]
+    (fn []
+      (let [communities (rf/sub [:communities/communities])
+            communities-count (count communities)
+        ; TODO move sorting to subscription
+            sorted-communities (sort-by :name communities)]
+        [safe-area/consumer
+         (fn []
+           [rn/view {:style {:margin-left     20
+                             :margin-right    20
+                             :flex             1
+                             :background-color (colors/theme-colors
+                                                colors/white
+                                                colors/neutral-90)}}
+            [quo/button {:icon     true
+                         :type     :grey
+                         :size     32
+                         :style    {:margin-vertical 12}
+                         :on-press #(rf/dispatch [:navigate-back])}
+             :i/close]
+            [screen-title]
+            [featured-communities-header communities-count]
+            [featured-list communities @view-type]
+            [other-communities-list sorted-communities @view-type]])]))))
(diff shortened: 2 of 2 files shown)


src/quo2/components/community/community_card_view.cljs
```diff
@@ -6,15 +6,11 @@
 
 (defn community-card-view-item
   [{:keys [name description locked
-           status tokens cover tags featured]} on-press]
+           status tokens cover tags width]} on-press]
   [rn/touchable-opacity {:on-press on-press}
-   [rn/view {:style (merge (style/community-card 20)
-                           {:margin-bottom 16}
```
J-Son89, line 12: cards design should care whether it is `featured` or not.
Also margin should be handled by it's container

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
-    :horizontal                        true
-    :getItemLayout                     get-item-layout-js
     :keyboard-should-persist-taps      :always
     :shows-horizontal-scroll-indicator false
+    :separator                 [rn/view {:margin-bottom 16}]
     :data                              communities
-    :render-fn                         render-featured-fn}])
+    :render-fn                         (render-fn false "100%")}])
 
-(defn featured-communities-section [communities]
```
J-Son89, line 83: these cards can use the same rendered function

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
+        communities-count (reagent/atom {:value (count communities)})
         ;;TODO move sorting to subscription
-        sorted-communities (sort-by :name communities)]
-    [rn/flat-list
-     {:key-fn                            :id
-      :getItemLayout                     get-item-layout-js
-      :keyboard-should-persist-taps      :always
-      :shows-horizontal-scroll-indicator false
-      :header                            (featured-communities-section communities)
-      :data                              sorted-communities
```
J-Son89, line 120: this header component was too advanced to be a `header` type component so I flattened the structure a bit.

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
+                                        :justify-content :center}
+                      :resize-mode     :center
+                      :size            20
+                      :color           (colors/theme-colors
+                                        colors/neutral-50
+                                        colors/neutral-40)}]])
+
+(defn featured-list [communities view-size]
+  [rn/view {:style {:flex-direction :row
+                    :overflow :hidden
```
J-Son89: actually need to remove this []

J-Son89:
@churik this is in a flow. should it get QA? It's just a fix on some design implementations

pavloburykh:
@J-Son89 thanx for the PR! Tested, looks good to me. 

Got a couple of clarifying questions. 

Should those Discover page elements be implemented or this is out of scope of current PR? Below are screenshots from Figma:

1. All, Open, Gated buttons

Communities for Mobile – Figma 2022-11-28 14-13-56

Currently these buttons are not shown 

2. Header buttons on community page (Back, Search, Options)

Communities for Mobile – Figma 2022-11-28 14-17-39

Currently they look like this:

photo_2022-11-28 14 19 58

J-Son89:
> @J-Son89 thanx for the PR! Tested, looks good to me.
> 
> Got a couple of clarifying questions.
> 
> Should those Discover page elements be implemented or this is out of scope of current PR? Below are screenshots from Figma:
> 
> 1. All, Open, Gated buttons
> 
> Communities for Mobile – Figma 2022-11-28 14-13-56
> 
> Currently these buttons are not shown
> 
> 2. Header buttons on community page (Back, Search, Options)
> 
> Communities for Mobile – Figma 2022-11-28 14-17-39
> 
> Currently they look like this:
> 
> photo_2022-11-28 14 19 58

Thanks @pavloburykh. So 1 is beyond the scope of this p.r as I was just fixing a ui issue that was brought in by some refactors.

and 2 should be fixed by another p.r I just merged today :) 
https://github.com/status-im/status-mobile/pull/14281

pavloburykh:
Thanx @J-Son89 ! Based on that, PR is ready for merge as soon as it passes the Review. I will move it to the Review column. After approval you can go ahead and merge it in case if there will be no additional commits based on the review.

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
-                                             colors/neutral-50
-                                             colors/neutral-40)}]]
-     [rn/view {:margin-top   8
-               :padding-left 20}
-      [featured-communities communities]]]))
-
-(defn other-communities []
+(defn discover []
   (let [communities (rf/sub [:communities/communities])
+        communities-count (reagent/atom {:value (count communities)})
```
flexsurfer: atom should be defined outside render function
```clojure
(let [communities-count (reagent/atom {:value (count communities)}]
   (fn []
       (let [communities (rf/sub [:communities/communities])
```

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
+                                            colors/white
+                                            colors/neutral-90)}}
+        [quo/button {:icon     true
+                     :type     :grey
+                     :size     32
+                     :style    {:margin-vertical 12}
+                     :on-press #(rf/dispatch [:navigate-back])}
+         :i/close]
+        [screen-title]
+        (featured-communities-header (:value  @communities-count))
```
flexsurfer: `(` shouldn't be used , only `[`

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
   [rn/flat-list
    {:key-fn                            :id
-    :horizontal                        true
-    :getItemLayout                     get-item-layout-js
     :keyboard-should-persist-taps      :always
     :shows-horizontal-scroll-indicator false
+    :separator                 [rn/view {:margin-bottom 16}]
     :data                              communities
-    :render-fn                         render-featured-fn}])
+    :render-fn                         (render-fn false "100%")}])
```
flexsurfer: :render-data should be used instead

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
-                   :type     :grey
-                   :size     32
-                   :style    {:margin-vertical 12
-                              :margin-left     20}
-                   :on-press #(rf/dispatch [:navigate-back])}
-       :i/close]
-      [screen-title]
-      [other-communities]])])
+        sorted-communities (sort-by :name communities)
+        view-size (reagent/atom 0)]
```
flexsurfer: same for this atom

flexsurfer: btw why do you create it here if its used only in featured-list

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
+                                           (oops/oget % "nativeEvent.layout.width")))}
+   (when-not (= @view-size 0)
+     [rn/flat-list
+      {:key-fn                            #(str "featured" (% :id))
+       :horizontal                        true
+       :keyboard-should-persist-taps      :always
+       :shows-horizontal-scroll-indicator false
+       :separator                  [rn/view {:margin-right 12}]
+       :data                              communities
+       :render-fn                         (render-fn true @view-size)}])])
```
flexsurfer: same here :render-data should be used, anon functions shouldn't  be used  for render-fn

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
+(defn featured-list [communities view-size]
+  [rn/view {:style {:flex-direction :row
+                    :overflow :hidden
+                    :width "100%"
+                    :margin-bottom 24}
+            :on-layout #(swap! view-size (fn []
+                                           (oops/oget % "nativeEvent.layout.width")))}
+   (when-not (= @view-size 0)
+     [rn/flat-list
+      {:key-fn                            #(str "featured" (% :id))
```
flexsurfer: it could be just :id , what featured gives ?

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
-                         (rf/dispatch [:navigate-to :community {:community-id (:id item)}]))
-        :on-long-press #(rf/dispatch [:bottom-sheet/show-sheet
-                                      {:content (fn []
+(defn render-fn [featured? width]
+  (fn
+    [community-item]
+    (let [item (merge community-item
+                      (get mock-community-item-data :data)
+                      {:featured featured?})]
+      (if (= @view-type :card-view)
```
flexsurfer: we shouldn't use global atoms, this should be passed as parameter

flexsurfer (changes requested):

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: because the data for both of these is the same so then it creates duplicate keys. or am I wrong about this? 🤔

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: you're referring to @view-type? . That code was already there. I figured it's a placeholder until live data is being passed in. But not sure why it was like that.

src/status_im2/contexts/communities/discover/view.cljs
flexsurfer: i mean `str "featured"` is constant, so key still will be duplicated

src/status_im2/contexts/communities/discover/view.cljs
flexsurfer: yeah i know, but we should fix it, as you're working on it anyway :)

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: hmm, 

no other communities iss using `:id`, featured is using "featured + :id"

I will revisit anyway 👍

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -4,19 +4,16 @@
...
-
+            [oops.core :as oops]
             ;; TODO move to status-im2
             [status-im.react-native.resources :as resources]
             [status-im.ui.screens.communities.community :as community]))
 
 (def view-type (reagent/atom :card-view))
 
-(def mock-community-item-data                               ;; TODO: remove once communities are loaded with this data.
+(def mock-community-item-data  ;; TODO: remove once communities are loaded with this data.
```
ilmotta, line 14: Double semicolon should use a separate line.

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
-                         (rf/dispatch [:communities/load-category-states (:id item)])
-                         (rf/dispatch [:dismiss-keyboard])
-                         (rf/dispatch [:navigate-to :community {:community-id (:id item)}]))
-        :on-long-press #(rf/dispatch [:bottom-sheet/show-sheet
-                                      {:content (fn []
+(defn render-fn [featured? width]
+  (fn
+    [community-item]
+    (let [item (merge community-item
+                      (get mock-community-item-data :data)
```
ilmotta: I wish we could code against fake data as if the implementation was *final* so that when the real data is available, nothing needs to change. In other words, the view should never know if the data is real or not because it's not its concern.

For example, view components should already correctly use a subscription and the fake data should be hidden behind the subscription; global vars to hold fake data shouldn't be used; bindings with the word "fake" in them shouldn't be used, etc.

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -34,105 +31,102 @@
...
    [quo/text {:accessibility-label :communities-screen-title
               :weight              :semi-bold
               :size                :heading-1}
     (i18n/label :t/discover-communities)]])
 
-(defn featured-communities [communities]
+(defn featured-communities-header [communities-count]
+  [rn/view {:flex-direction     :row
+            :height             30
+            :margin-bottom 8
```
ilmotta, line 62: Could you please reformat this file? There are some misaligned things, like this line.

src/status_im2/contexts/communities/discover/view.cljs
ilmotta: As @flexsurfer commented, this is going to cause an unnecessary rerender of the entire `discover` component whenever on-layout handler swaps the atom.

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: cool, I will update :)

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: yeah I agree, I'm not quite sure how long until we have access to these real data points. I will raise this discussion in the features breakdown today, as maybe it's a good idea to move all of this mock data to a centralised place similar to what you are describing.

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: sure :D

src/status_im2/contexts/communities/discover/view.cljs
ilmotta: Thanks @J-Son89. This can save us some time in the future and it makes code review a bit easier (at least for me) because then everyone is on the same page to treat the data flowing as if it was production data.

ilmotta (approved):

src/status_im2/contexts/communities/discover/view.cljs
flexsurfer: actually, it's not the problem, problem is that atom will be redefined with default value when one of subscription will change

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -51,88 +47,94 @@
...
+                        :margin-bottom 24}
+                :on-layout #(swap! view-size
+                                   (fn []
+                                     (oops/oget % "nativeEvent.layout.width")))}
+       (when-not (= @view-size 0)
+         [rn/flat-list
+          {:key-fn                            :id
+           :horizontal                        true
+           :keyboard-should-persist-taps      :always
+           :shows-horizontal-scroll-indicator false
```
flexsurfer, line 95: align map

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -51,88 +47,94 @@
...
+           :separator                  [rn/view {:margin-right 12}]
```
flexsurfer: thars weird separator

flexsurfer (approved):

src/status_im2/contexts/communities/discover/view.cljs
J-Son89: why so? it's a horizontal scroll in the designs

flexsurfer: oh, then it could be like [rn/view {:width 12}]] because an empty separator with only a margin looks strange, yeah but anyway, now i know its like a gap between items

J-Son89: okay, yeah that also makes sense.

PR #14451, opened 2022-11-28
[Feature][#14352] Added mentions in Activity Center

### LINKED ISSUE #14352: Implement Mentions in Activity Centre
### Display Mentions in Activity Centre

As a user, I would like to see the mentions from my communities and Ad-Hoc chats I'm in.

1 line

### Design

Link to [Figma](https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926)

smohamedjavid:
Fixes #14352

### Summary

Implement Mentions in Activity Center 

Mentions

**Design** 

([Figma Link](https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=639%3A21926))

#### Areas impacted

- Activity Center (Notifications Center)

**Platforms**

- Android
- iOS

**Platforms**

- 1-1 Chats
- Public Chats
- Group Chats

status: ready

### DIFF
--- a/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
+++ b/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
@@ -11,7 +11,7 @@
             [clojure.string :as str]))
 
 (defn message-body [message]
-  (let [parsed-text (get-in message [:content :parsed-text])
+  (let [parsed-text          (get-in message [:content :parsed-text])
         parsed-text-children (:children (first parsed-text))]
     [quo2/text {:number-of-lines     2
                 :style               style/tag-text
--- a/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
+++ b/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
@@ -1,0 +1,55 @@
+(ns status-im.ui.screens.activity-center.notification.mentions.view
+  (:require [quo.components.animated.pressable :as animation]
+            [quo2.core :as quo2]
+            [quo2.components.markdown.text :as text]
+            [quo2.foundations.colors :as colors]
+            [status-im.i18n.i18n :as i18n]
+            [status-im.multiaccounts.core :as multiaccounts]
+            [status-im.ui.screens.activity-center.notification.mentions.style :as style]
+            [status-im.ui.screens.activity-center.utils :as activity-center.utils]
+            [status-im.utils.datetime :as datetime]
+            [utils.re-frame :as rf]
+            [clojure.string :as str]))
+
+(defn message-body [message]
+  (let [parsed-text (get-in message [:content :parsed-text])
+        parsed-text-children (:children (first parsed-text))]
+    [text/text {:number-of-lines     2
+                :style               style/tag-text
+                :accessibility-label :activity-message-body
+                :size                :paragraph-1}
+     (for [{:keys [type literal]} parsed-text-children]
+       (case type
+         "mention" [text/text {:style style/mention-text
+                               :size  :paragraph-1}
+                    (str "@" (rf/sub [:contacts/contact-name-by-identity literal]))]
+         literal))]))
+
+(defn view
+  [{:keys [author chat-name chat-id message] :as notification}]
+  (let [contact (rf/sub [:contacts/contact-by-identity author])]
+    [animation/pressable
+     {:on-press (fn []
+                  (rf/dispatch [:hide-popover])
+                  (rf/dispatch [:chat.ui/navigate-to-chat chat-id]))}
+     [quo2/activity-log
+      {:title     (i18n/label :t/mention)
+       :icon      :main-icons2/mention
+       :timestamp (datetime/timestamp->relative (:timestamp notification))
+       :unread?   (not (:read notification))
+       :context   [[quo2/user-avatar-tag
+                    {:color          :purple
+                     :override-theme :dark
+                     :size           :small
+                     :style          style/tag
+                     :text-style     style/tag-text}
+                    (activity-center.utils/contact-name contact)
+                    (multiaccounts/displayed-photo contact)]
+                   [quo2/text {:style style/tag-text} (str/lower-case (i18n/label :t/on))]
+                  ;;  TODO: The `group-avatar-tag` component does NOT support displaying channel name along with community/chat name. Need to update the component to support it. 
+                   [quo2/group-avatar-tag chat-name {:size           :small
+                                                     :override-theme :dark
+                                                     :color          colors/primary-50
+                                                     :style          style/tag
+                                                     :text-style     style/tag-text}]]
+       :message    {:body (message-body message)}}]]))
--- a/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
+++ b/src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
@@ -1,7 +1,6 @@
 (ns status-im.ui.screens.activity-center.notification.mentions.view
   (:require [quo.components.animated.pressable :as animation]
             [quo2.core :as quo2]
-            [quo2.components.markdown.text :as text]
             [quo2.foundations.colors :as colors]
             [status-im.i18n.i18n :as i18n]
             [status-im.multiaccounts.core :as multiaccounts]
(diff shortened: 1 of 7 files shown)


src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,55 @@
...
+            [status-im.ui.screens.activity-center.notification.mentions.style :as style]
+            [status-im.ui.screens.activity-center.utils :as activity-center.utils]
+            [status-im.utils.datetime :as datetime]
+            [utils.re-frame :as rf]
+            [clojure.string :as str]))
+
+(defn message-body [message]
+  (let [parsed-text (get-in message [:content :parsed-text])
+        parsed-text-children (:children (first parsed-text))]
+    [text/text {:number-of-lines     2
```
flexsurfer: quo/text

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,55 @@
...
+(defn view
+  [{:keys [author chat-name chat-id message] :as notification}]
+  (let [contact (rf/sub [:contacts/contact-by-identity author])]
+    [animation/pressable
+     {:on-press (fn []
+                  (rf/dispatch [:hide-popover])
+                  (rf/dispatch [:chat.ui/navigate-to-chat chat-id]))}
+     [quo2/activity-log
+      {:title     (i18n/label :t/mention)
+       :icon      :main-icons2/mention
```
flexsurfer: :i/mention

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,55 @@
...
+        parsed-text-children (:children (first parsed-text))]
+    [text/text {:number-of-lines     2
+                :style               style/tag-text
+                :accessibility-label :activity-message-body
+                :size                :paragraph-1}
+     (for [{:keys [type literal]} parsed-text-children]
+       (case type
+         "mention" [text/text {:style style/mention-text
+                               :size  :paragraph-1}
+                    (str "@" (rf/sub [:contacts/contact-name-by-identity literal]))]
```
flexsurfer: wow that looks confusing at first, like deref of sub :)

flexsurfer (approved):

Parveshdhull (approved):

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,55 @@
...
+            [clojure.string :as str]))
+
+(defn message-body [message]
+  (let [parsed-text (get-in message [:content :parsed-text])
+        parsed-text-children (:children (first parsed-text))]
+    [text/text {:number-of-lines     2
+                :style               style/tag-text
+                :accessibility-label :activity-message-body
+                :size                :paragraph-1}
+     (for [{:keys [type literal]} parsed-text-children]
```
ilmotta: I think we need a unique key for each child component.

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,54 @@
...
+
+(defn message-body [message]
+  (let [parsed-text (get-in message [:content :parsed-text])
+        parsed-text-children (:children (first parsed-text))]
+    [quo2/text {:number-of-lines     2
+                :style               style/tag-text
+                :accessibility-label :activity-message-body
+                :size                :paragraph-1}
+     (for [{:keys [type literal]} parsed-text-children]
+       (case type
```
ilmotta: Do you think we need the `case` or could we just go with the simpler `if` given there's only one condition?

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
smohamedjavid: Yeah :)

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,54 @@
...
+    [animation/pressable
+     {:on-press (fn []
+                  (rf/dispatch [:hide-popover])
+                  (rf/dispatch [:chat.ui/navigate-to-chat chat-id]))}
+     [quo2/activity-log
+      {:title     (i18n/label :t/mention)
+       :icon      :i/mention
+       :timestamp (datetime/timestamp->relative (:timestamp notification))
+       :unread?   (not (:read notification))
+       :context   [[quo2/user-avatar-tag
```
ilmotta: The following is just a light suggestion ;)

Now that we are integrating the third type of notification in the Activity Center, this user-avatar-tag call is repeated three times already and they're all the same.

We could create a `activity_center/notification/common/view.cljs` file and extract the user avatar tag component into a common component and reuse them across all three different notifications. Of course we should be careful not to go over the board with the DRY principle, but this really looks like a good instance to me.

```clojure
;; Extract this component that's repeated in all notifications.
[quo2/user-avatar-tag
 {:color          :purple
  :override-theme :dark
  :size           :small
  :style          style/user-avatar-tag
  :text-style     style/user-avatar-tag-text}
 (activity-center.utils/contact-name contact)
 (multiaccounts/displayed-photo contact)]
```

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
```diff
@@ -0,0 +1,54 @@
...
+                    {:color          :purple
+                     :override-theme :dark
+                     :size           :small
+                     :style          style/tag
+                     :text-style     style/tag-text}
+                    (activity-center.utils/contact-name contact)
+                    (multiaccounts/displayed-photo contact)]
+                   [quo2/text {:style style/tag-text} (str/lower-case (i18n/label :t/on))]
+                  ;;  TODO: The `group-avatar-tag` component does NOT support displaying channel name along with community/chat name. Need to update the component to support it.
```
ilmotta: [Code style] Could you break this long comment into multiple lines? It'd be good to add your username too, like `TODO(smohamedjavid)`. Also the double semicolon is not properly aligned with the rest of the code.

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
smohamedjavid: It's a map (of text/message data) that has been iterated inside a Text Component. We return a new text component if it's a `mention` where we display the contact name (instead of the public key) with some additional text styles without affecting the parent text style and maintaining the same line as the parent. The rest of the text will be fed as it is (without any component/style).

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
smohamedjavid: I thought of the `if` for simplicity since it's a map and if we introduce any new type in future, it would be easier for us to add a case rather than refactoring the code block at the moment.

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
ilmotta: Okay, so are you sure there's no warning being generated by the for loop without a unique key? I thought `^{:key ...}` was mandatory here.

ilmotta: I just tested locally and I'm getting a warning due to the collection being returned as children of quo2/text.

```
  ERROR    Warning: Each child in a list should have a unique "key" prop.

Check the render method of `quo2.components.markdown.text.text`. See https://fb.me/react-warning-keys for more information.
    in quo2.components.markdown.text.text (created by quo2.components.markdown.text.text)
    in quo2.components.markdown.text.text (created by quo2.components.notifications.activity_log.view.activity_message)
    in RCTView (at View.js:34)
    in View (created by quo2.components.notifications.activity_log.view.activity_message)
```

smohamedjavid: My apologies - I will add the key 👍 :)

ilmotta:
@smohamedjavid, I've tested mention notifications now and they do appear in the Activity Center, but the bell icon in the home screen doesn't display a blue dot even when there's one or more unread mention notifications. This seems to be a bug. I haven't looked into why, but I think we should have an issue for this.

ilmotta (approved):

src/status_im/ui/screens/activity_center/notification/mentions/view.cljs
smohamedjavid: Moved it to `common` 👍

PR #14472, opened 2022-12-01
Introduce subscription tests

ilmotta:
### Summary

This PR brings to every contributor's reach the capability to easily test subscriptions in a way that also make sure re-frame's *signal graph* is validated. It also adds tests to more complex subscriptions to show how well the proposed solution scales to any subscription.

There's a long rationale about why this PR is doing what it's doing, and I hope the explanations that follow can help us get on the same page.

I'm assuming not everyone is familiar with some of the concepts I mention here, and because of that I prefer to more thoroughly explain them. I'm sorry if sometimes this feels pedantic to you.

### Motivation

1. Just a handful of layer-3 subscriptions have tests, and the ones that do are incomplete.

2. The team is not aligned on how to design layer-3 subscription tests and the perils lurking in the dark mysteries of re-frame. We can see this in some PR threads.

3. **Make more layers of the architecture testable**, so that we can start to actually increase the quality of the app over time. In my opinion, at the speed the mobile codebase is growing/churning and the amount of contributors coming in, we need to aggressively improve our testing story.

Quality assurance is not just trusting end-of-the-line QA processes (e.g. manual tests, PR reviews and e2e tests) and expecting engineers to write great code (whatever that means). **Quality should be embedded in everything we do.**

Subscriptions (layer-3), events, effects and interceptors should be tested and there's nothing preventing us from doing so. They are the underlying structures that make the app work, and if we don't test them we put too much burden on the last layer (view), which is the most complicated to test.

**Why should we care?** Software engineers test so that systems can **evolve with confidence from one snapshot to another**, often in highly collaborative environments. Layer-3 subs are one piece of this puzzle.

### Main problem

What's all the fuss about testing subscription handlers? After all, they're pure functions.

If you're not yet familiar with or need a refresh on **re-frame's signal graph**, I highly recommend first reading https://day8.github.io/re-frame/subscriptions/.

According to [reframe's testing doc](https://github.com/day8/re-frame/blob/master/docs/Testing.md), only layer-3 subscriptions should be tested. I completely agree with that because testing *extractors* (layer-2) is almost pointless given their simplicity (there's no real logic). Therefore, **this PR is only concerned with layer-3 subs**.

Re-frame suggests testing subscriptions by *extracting anonymous subscription handlers to named functions*. This works, but in practice, testing like that has important limitations for subscriptions because it's easy to make tests pass, when **in reality the signal graph could be broken** :boom:

### Example problem

Here's a relatively complicated layer-3 subscription:

```clojure
(re-frame/reg-sub
 :portfolio-value
 :<- [:balances]
 :<- [:prices]
 :<- [:wallet/currency]
 :<- [:wallet/token->decimals]
 (fn [[balances prices currency token->decimals]]
   (if (and balances prices)
     (let [currency-key        (-> currency :code keyword)
           balance-total-value (apply + (map #(get-balance-total-value % prices currency-key token->decimals) balances))]
       (if (pos? balance-total-value)
         (-> balance-total-value
             (money/with-precision 2)
             str
             (i18n/format-currency (:code currency)))
         "0"))
     "...")))
```

If we follow re-frame's recommendation to test this subscription, then we should extract the anonymous function, like so.

```clojure
(defn portfolio-value
  [[balances prices currency token->decimals]]
  (if (and balances prices)
    (let [currency-key        (-> currency :code keyword)
          balance-total-value (apply + (map #(get-balance-total-value % prices currency-key token->decimals) balances))]
      (if (pos? balance-total-value)
        (-> balance-total-value
            (money/with-precision 2)
            str
            (i18n/format-currency (:code currency)))
        "0"))
    "..."))

(re-frame/reg-sub
 :portfolio-value
 :<- [:balances]
 :<- [:prices]
 :<- [:wallet/currency]
 :<- [:wallet/token->decimals]
 portfolio-value)
```

At this point, the extracted public function is testable, but notice we lose the ability to actually verify the subscription is receiving the correct *inputs*.

Let's say requirements change, and `:wallet/currency` returns a different map structure. Well then, instead of breaking, tests would still pass. Now imagine the entire system's subscriptions are tested like this. In that case, **every single change to a subscription would require manually hunting its usages and checking if tests need to be updated**, otherwise **there's no guarantee a regression in the signal graph was not introduced**. This is a lot of work many contributors won't do (including myself).

### A little bit better, but with issues

The solution I found to be the most valuable from previous experiences is to not ignore the subscription inputs, i.e. subscribe to the real subscription and let re-frame do its work.

```clojure
(deftest portfolio-value-test
  (testing "returns zero when balance is not positive"
    ;; Important in REPL sessions
    (rf-subs/clear-subscription-cache!)
    (let [original-db  @rf-db/app-db
          empty-wallet {:accounts {main-account-id
                                   {:balance {:ETH money-zero
                                              :SNT money-zero}}}}]
      ;; Arrange section
      (swap! rf-db/app-db assoc
             :multiaccount/accounts accounts
             :prices prices
             :wallet empty-wallet
             :wallet/all-tokens tokens)

      ;; Act, assert and clean-up.
      (try
        (is (= "0" (rf/sub [:portfolio-value])))
        (finally
          (reset! rf-db/app-db original-db))))))
```

So as you can see, now the problem is that the test is mixed up with boring stuff to arrange and clean-up state. You could use `cljs.test/use-fixtures`, but it has a limitation that it only works for top-level tests, not for individual `cljs.test/testing` sections, so you'd end up having to define multiple `deftest`s repeating long subscription names. Or you'd need to copy & paste setup/teardown code inside `testing` macros. No fun, and again, easy to miss and end-up with flaky tests (no no no...)

### Even better solution

Clojure gives us the power of macros to abstract all that plumbing. So here's the final solution, using `test-helpers/deftest-sub`. The actual implementation of the macro is quite interesting, see the *For the Curious* section for more details.

```clojure
(deftest-sub :portfolio-value
  [sub-name]
  (testing "returns zero when balance is not positive"
    (let [empty-wallet {:accounts {main-account-id
                                   {:balance {:ETH money-zero
                                              :SNT money-zero}}}}]
      (swap! rf-db/app-db assoc
             :multiaccount/accounts accounts
             :prices prices
             :wallet empty-wallet
             :wallet/all-tokens tokens)
      (is (= "0" (rf/sub [sub-name]))))))
```

Important things:

- The name of test is derived from the actual name of the subscription. This means if the production code renames the subscription, then tests will fail.
- It's just as fast as any other simple unit test.
- The macro works as expected with clj-kondo.
- There's zero need to worry about messing up the app DB state and affecting or being affected by other tests.
- It works just fine for REPL-Drive Development. In fact, all tests in this PR were implemented using the test REPL.
- We can use multiple and nested calls to the `testing` macro, and they'll be correctly augmented by the `deftest-sub` macro to reset the subscription cache and restore the app DB state.

#### The `deftest-sub` Macro

If you'd like to learn more about macros, specifically Deep Code Walking Macros (DCWM), check out this excellent blog post https://blog.fogus.me/2013/07/17/an-introduction-to-deep-code-walking-macros-with-clojure. If, on the other hand, you have no idea how macros work, the Clojure for the Brave and True book [section on macros](https://www.braveclojure.com/writing-macros/) is a very good starting point.

Now, if you try to call the `deftest-sub` macro with invalid arguments, you get a *beatiful* error generated by [expound](https://github.com/bhb/expound). This works because I've defined a `clojure.spec.alpha/fdef` for the macro. This is what Clojure core does for many macros, and it greatly improves DX.

See, I'm not suggesting using clojure.spec everywhere, but using it for macros with instrumentation enabled is definitely a win. `fdef` has zero impact in production performance with instrumentation disabled, and besides this macro is only for testing, so no impact whatsoever.

```
------ REPL Error while processing ---------------------------------------------
(deftest-sub "some-subscription-name"
  [query-id]
  (do-something))
Syntax error macroexpanding status-im.test-helpers/deftest-sub.
Call to status-im.test-helpers/deftest-sub did not conform to spec.
-- Spec failed --------------------

("some-subscription-name" ... ...)
^^^^^^^^^^^^^^^^^^^^^^^^

should satisfy

keyword?

-------------------------
Detected 1 error
```

### What about the guidelines?

This is an introductory PR to the subject of testing subscriptions, and it's my hope that PRs from now on will be able to be opened with subscription tests whenever needed. I think we can update the guidelines once more people write subscription tests and we get practical feedback.

### More references

- https://github.com/day8/re-frame/blob/master/docs/FAQs/Why-Clear-Sub-Cache.md
- https://blog.klipse.tech/clojure/2016/10/10/defn-args.html
- https://cljdoc.org/d/expectations/clojure-test/1.2.1/doc/fixtures-focused-execution

status: ready

### DIFF
--- a/src/status_im/test_helpers.clj
+++ b/src/status_im/test_helpers.clj
@@ -12,22 +12,10 @@
        (map #(string/replace % #"\." "-"))
        (string/join "-")))
 
-(defmacro ^:private restore-app-db
-  "Saves app db, executes `body` and restores to original app-db.
-
-  Also clears the subscription cache for good measure."
-  [& body]
-  `(do (re-frame.subs/clear-subscription-cache!)
-       (let [original-db# (deref re-frame.db/app-db)]
-         (try
-           ~@body
-           (finally
-             (reset! re-frame.db/app-db original-db#))))))
-
 (defmacro ^:private testing-subscription
   [description & body]
   `(cljs.test/testing ~description
-     (restore-app-db ~@body)))
+     (restore-app-db (fn [] ~@body))))
 
 (s/fdef deftest-sub
   :args (s/cat :sub-name keyword?
@@ -60,9 +48,10 @@
   [sub-name args & body]
   `(let [sub-name# ~sub-name]
      (cljs.test/deftest ~(symbol (subscription-name->test-name sub-name))
-       (restore-app-db
-        (let [~args [sub-name#]]
-          ~@(clojure.walk/postwalk-replace
-             {'cljs.test/testing `testing-subscription
-              'testing           `testing-subscription}
-             body))))))
+       (let [~args [sub-name#]]
+         (restore-app-db
+          (fn []
+            ~@(clojure.walk/postwalk-replace
+               {'cljs.test/testing `testing-subscription
+                'testing           `testing-subscription}
+               body)))))))
--- a/src/status_im/test_helpers.clj
+++ b/src/status_im/test_helpers.clj
@@ -1,0 +1,67 @@
+(ns status-im.test-helpers
+  (:require [clojure.spec.alpha :as s]
+            [clojure.string :as string]
+            [clojure.walk :as walk]))
+
+(defn- subscription-name->test-name
+  [sub-name]
+  (->> [(namespace sub-name)
+        (name sub-name)
+        "test"]
+       (remove nil?)
+       (map #(string/replace % #"\." "-"))
+       (string/join "-")))
+
+(defmacro ^:private restore-app-db
+  "Saves app db, executes `body` and restores to original app-db.
+
+  Also clears the subscription cache for good measure."
+  [& body]
+  `(do (re-frame.subs/clear-subscription-cache!)
+       (let [original-db# (deref re-frame.db/app-db)]
+         (try
+           ~@body
+           (finally
+             (reset! re-frame.db/app-db original-db#))))))
+
+(defmacro ^:private testing-subscription
+  [description & body]
+  `(cljs.test/testing ~description
+     (restore-app-db ~@body)))
+
+(s/fdef deftest-sub
+  :args (s/cat :sub-name keyword?
+               :args (s/coll-of symbol? :count 1)
+               :body (s/* any?)))
+(defmacro deftest-sub
+  "Defines a test based on `sub-name`, executes `body` and restores the app db.
+
+  Any usage of the `cljs.test/testing` macro inside `body` will be modified to
+  also make sure the app db is restored and the subscription cache is reset.
+
+  Expressions in `body` will have access to `sub-name`, which should be used to
+  avoid needlessly repeating the subscription name.
+
+  Example:
+
+  ```clojure
+  (require '[status-im.test-helpers :as h])
+
+  (h/deftest-sub :wallet/sorted-tokens
+    [sub-name]
+    (testing \"sorts tokens by name, lowercased\"
+      ;; Arrange
+      (swap! rf-db/app-db assoc-in [<db-path>] <value>)
+
+      ;; Act and Assert
+      (is (= <expected> (rf/sub [sub-name])))))
+  ```"
+  [sub-name args & body]
+  `(let [sub-name# ~sub-name]
+     (cljs.test/deftest ~(symbol (subscription-name->test-name sub-name))
+       (restore-app-db
+        (let [~args [sub-name#]]
+          ~@(clojure.walk/postwalk-replace
+             {'cljs.test/testing `testing-subscription
+              'testing           `testing-subscription}
+             body))))))
--- a/src/status_im/test_helpers.clj
+++ b/src/status_im/test_helpers.clj
@@ -33,6 +33,7 @@
   :args (s/cat :sub-name keyword?
                :args (s/coll-of symbol? :count 1)
                :body (s/* any?)))
+
 (defmacro deftest-sub
   "Defines a test based on `sub-name`, executes `body` and restores the app db.
 
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -10,6 +10,8 @@
            status-im.test-helpers/deftest-sub clojure.core/defn
            taoensso.tufte/defnp               clojure.core/defn}
  :linters {:invalid-arity     {:skip-args [status-im.utils.fx/defn utils.re-frame/defn]}
-           ;;TODO remove number when this is fixed
-           ;;https://github.com/borkdude/clj-kondo/issues/867
-           :unresolved-symbol {:exclude [PersistentPriorityMap.EMPTY number]}}}
+           ;; TODO remove number when this is fixed
+           ;; https://github.com/borkdude/clj-kondo/issues/867
+           :unresolved-symbol {:exclude [PersistentPriorityMap.EMPTY
+                                         number
+                                         status-im.test-helpers/restore-app-db]}}}
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,13 +1,14 @@
-{:lint-as {status-im.utils.views/defview  clojure.core/defn
-           status-im.utils.views/letsubs  clojure.core/let
-           reagent.core/with-let          clojure.core/let
-           status-im.utils.fx/defn        clj-kondo.lint-as/def-catch-all
-           utils.re-frame/defn            clj-kondo.lint-as/def-catch-all
-           quo.react/with-deps-check      clojure.core/fn
-           quo.previews.preview/list-comp clojure.core/for
-           status-im.utils.styles/def     clojure.core/def
-           status-im.utils.styles/defn    clojure.core/defn
-           taoensso.tufte/defnp           clojure.core/defn}
+{:lint-as {status-im.utils.views/defview      clojure.core/defn
+           status-im.utils.views/letsubs      clojure.core/let
+           reagent.core/with-let              clojure.core/let
+           status-im.utils.fx/defn            clj-kondo.lint-as/def-catch-all
+           utils.re-frame/defn                clj-kondo.lint-as/def-catch-all
+           quo.react/with-deps-check          clojure.core/fn
+           quo.previews.preview/list-comp     clojure.core/for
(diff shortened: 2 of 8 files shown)


cammellos (approved):
Really informative and great work!

erikseppanen (commented):
While I'm in agreement (and excited) with most of this, since (and I think for good reason) the clojure community commonly advises data over functions over macros (descending order of preference), I'll do my duty here to question the necessity of the macro:

The reasoning here seems a little thin (I might be wrong):
> You could use cljs.test/use-fixtures, but it has a limitation that it only works for top-level tests, not for individual cljs.test/testing sections, so you'd end up having to define multiple deftests repeating long subscription names. Or you'd need to copy & paste setup/teardown code inside testing macros. No fun, and again, easy to miss and end-up with flaky tests (no no no...)

- Is having multiple `deftests` really such a bad thing? It's just longer test names right? Not saying it's ideal, but if we're weighing it against the overhead of a macro, something to consider...
- If we were not able to use `use-fixtures` (as you mentioned regarding with `testing` sections), would just creating a function for setup / teardown (instead of `use-fixtures`) be enough for us to avoid introducing a macro? (example at the bottom of this page: https://clojuredocs.org/clojure.test/use-fixtures)

src/status_im/test_helpers.clj
```diff
@@ -0,0 +1,67 @@
...
+
+(defmacro ^:private testing-subscription
+  [description & body]
+  `(cljs.test/testing ~description
+     (restore-app-db ~@body)))
+
+(s/fdef deftest-sub
+  :args (s/cat :sub-name keyword?
+               :args (s/coll-of symbol? :count 1)
+               :body (s/* any?)))
```
flexsurfer, line 23: new line

flexsurfer:
that's awesome, thank you, would be cool to have description of this PR somewhere in the documentation

flexsurfer (approved):

ilmotta:
> While I'm in agreement (and excited) with most of this, since (and I think for good reason) the clojure community commonly advises data over functions over macros (descending order of preference), I'll do my duty here to question the necessity of the macro:
> 
> The reasoning here seems a little thin (I might be wrong):
> 
> > You could use cljs.test/use-fixtures, but it has a limitation that it only works for top-level tests, not for individual cljs.test/testing sections, so you'd end up having to define multiple deftests repeating long subscription names. Or you'd need to copy & paste setup/teardown code inside testing macros. No fun, and again, easy to miss and end-up with flaky tests (no no no...)
> 
> * Is having multiple `deftests` really such a bad thing? It's just longer test names right? Not saying it's ideal, but if we're weighing it against the overhead of a macro, something to consider...
> * If we were not able to use `use-fixtures` (as you mentioned regarding with `testing` sections), would just creating a function for setup / teardown (instead of `use-fixtures`) be enough for us to avoid introducing a macro? (example at the bottom of this page: https://clojuredocs.org/clojure.test/use-fixtures)

Hey @erikseppanen, difficult questions to answer. As I was writing the code I could almost hear you saying: "why macros?" Before trying to give my perspective, I'd like to reiterate some of the problems we're trying to solve.

1. Contributors have varying degrees of experience with Clojure, and many are new to it.
3. Even though subscriptions are incredibly important, often containing important business rules, they didn't evolve with tests in the mobile codebase. Other parts of the codebase do have unit tests, so what happened? One reason is because people didn't know exactly how to test them correctly. In recent PRs I've seen subscriptions being changed/refactored/etc, but no tests accompanying them. **When I review and ask for tests, there's not a single good example in the code people can refer to.**

So my mindset when working in this PR was to make things as easy as possible and not as simple as possible. Which is a prioritization that I often prefer in the context of *testing*. The easier to write tests, the more tests people will write.

>Is having multiple `deftests` really such a bad thing? It's just longer test names right? Not saying it's ideal, but if we're weighing it against the overhead of a macro, something to consider...

Longer test names are not that big deal fore sure, and having multiple `deftests` is also okay, I agree with you.

>If we were not able to use `use-fixtures` (as you mentioned regarding with `testing` sections), would just creating a function for setup / teardown (instead of `use-fixtures`) be enough for us to avoid introducing a macro? (example at the bottom of this page: https://clojuredocs.org/clojure.test/use-fixtures)

Yes, a function would be enough. As usual, a macro is almost never needed, especially the ones for syntax sugar. Based in your comments, the following example could be a general solution that doesn't require `use-fixtures` (which doesn't work with the `testing` macro, so I prefer to avoid relying on it).

```clojure
(deftest communities-section-list-test
  (testing "builds sections using the first community name char (uppercased)"
    (h/setup-subs
     (fn []
       (swap! rf-db/app-db assoc :communities
              {"0x1" {:name "civilized monkeys"}
               "0x2" {:name "Civilized rats"}})
       (is (= [{:title "C"
                :data  [{:name "civilized monkeys"}
                        {:name "Civilized rats"}]}]
              (rf/sub [:communities/section-list]))))))

  (testing "sorts by section ascending"
    (h/setup-subs
     (fn []
       (swap! rf-db/app-db assoc :communities
              {"0x3" {:name "Memorable"}
               "0x1" {:name "Civilized monkeys"}})
       (is (= [{:title "C" :data [{:name "Civilized monkeys"}]}
               {:title "M" :data [{:name "Memorable"}]}]
              (rf/sub [:communities/section-list])))))))
```

I like the solution with and without the macro. By using the function approach, we lose some things in the name of simplicity, in exchange for developers to duplicate certain names and remember how/when to use the `setup-subs` function (otherwise they'd get flaky tests, as I did in my REPL sessions).

**I'd be just as happy to merge a solution with only the function if that's what's preferred. The important thing is to help all of us test layer-3 subscriptions in a coherent manner.**

ilmotta:
> that's awesome, thank you, would be cool to have description of this PR somewhere in the documentation.

Sure @flexsurfer, I could document a little bit in the `doc/TESTING.md` file. Perhaps add a link pointing from the new guidelines to `doc/TESTING`. Nothing crazy, just a simple explanation. What do you think?

vkjr (approved):

yqrashawn (approved):

siddarthkay (approved):

ibrkhalil (approved):
Thank you so much for doing this Icaro! ❤️👑

qfrank (approved):

smohamedjavid (approved):

erikseppanen (approved):

PR #14473, opened 2022-12-02
feat: add display-name in author component

yqrashawn:
### Summary

extract the display name logic into a single component

CleanShot 2022-12-02 at 09 55 37

status: ready

### DIFF
--- a/src/quo2/components/messages/author/view.cljs
+++ b/src/quo2/components/messages/author/view.cljs
@@ -9,6 +9,28 @@
 
 (def middle-dot "·")
 
+(defn display-name [{:keys [profile-name nickname ens-name text-style]}]
+  (let [ens?      (-> ens-name string/blank? not)
+        nickname? (-> nickname string/blank? not)]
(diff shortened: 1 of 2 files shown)


src/quo2/components/messages/author/view.cljs
```diff
@@ -9,6 +9,29 @@
...
+  (let [ens?      (-> ens-name string/blank? not)
+        nickname? (-> nickname string/blank? not)]
+    (if ens?
+      [text/text (merge {:weight :semi-bold
+                         :size   :paragraph-2
+                         :style  (style/ens-text)}
+                        text-style)
+       ens-name]
+      [:<>
+       (when nickname?
```
briansztamfater: Maybe instead of using both `when` and `when-not` you could just use an `if`?

yqrashawn: Fixed

flexsurfer (approved):

PR #14484, opened 2022-12-03
Add peers stats screen

rasom:
status: ready

### DIFF
--- a/src/status_im/ui/screens/peers_stats.cljs
+++ b/src/status_im/ui/screens/peers_stats.cljs
@@ -1,0 +1,14 @@
+(ns status-im.ui.screens.peers-stats
+  (:require [react-native.core :as react-native.core]
+            [status-im.i18n.i18n :as i18n]
+            [utils.re-frame :as re-frame]))
+
+(defn peers-stats []
+  (let [peers-count (re-frame/sub [:peers-count])]
+    [react-native.core/view {:flex              1
+                             :margin-horizontal 8}
+     [react-native.core/view
+      {:style {:flex-direction  :row
+               :margin-vertical 8
+               :justify-content :space-between}}
+      [react-native.core/text (str (i18n/label :t/peers-count) ": " peers-count)]]]))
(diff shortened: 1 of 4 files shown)


flexsurfer (approved):

src/status_im/ui/screens/peers_stats.cljs
```diff
@@ -0,0 +1,15 @@
+(ns status-im.ui.screens.peers-stats
+  (:require [quo.react-native :as quo.react-native]
+            [status-im.ui.components.react :as react]
+            [status-im.i18n.i18n :as i18n]
+            [re-frame.core :as re-frame.core]))
+
+(defn peers-stats []
+  (let [peers-count @(re-frame.core/subscribe [:peers-count])]
```
ilmotta: [Guidelines] We're using `utils.re-frame/sub` instead of `re-frame.core/subscribe`.

src/status_im/ui/screens/peers_stats.cljs
```diff
@@ -0,0 +1,15 @@
+(ns status-im.ui.screens.peers-stats
+  (:require [quo.react-native :as quo.react-native]
```
ilmotta: [Guidelines] Use `react-native.core` to call `view` and `text` components. You will be able to remove `[status-im.ui.components.react :as react]` as well.

ilmotta (approved):

src/status_im/ui/screens/peers_stats.cljs
flexsurfer: it looks like a temporary screen in the old code base, so I don't think it really matters there , because we don't want to refactor old code

cammellos: though if we get it right now, we can easily move it to the new screen once that's ready, or I misunderstand something?

rasom: fixed, thanks

src/status_im/ui/screens/peers_stats.cljs
rasom: fixed

PR #14487, opened 2022-12-05
Photo Selector 2

### LINKED ISSUE #14474: Photo Selector (2)
A second issue to track the creation of photo selector.

Designs: https://www.figma.com/file/wA8Epdki2OWa8Vr067PCNQ/Composer-for-Mobile?node-id=2827%3A269831&t=1NrhQoSGitjVl9dD-4

OmarBasem:
fixes #14474

This is a second PR for the implementation of the new photo selector. The PR implements reading more images from the device as the user scrolls down the selector.

Designs: https://www.figma.com/file/wA8Epdki2OWa8Vr067PCNQ/Composer-for-Mobile?node-id=2827%3A269831&t=1NrhQoSGitjVl9dD-4


Makefile
```diff
@@ -319,7 +319,7 @@ run-visual-test-ios: XCODE_DERIVED_DATA := $(HOME)/Library/Developer/Xcode/Deriv
 run-visual-test-ios: APPLICATION_NAME := StatusIm-brfnruzfrkkycpbndmdoeyrigthc
 run-visual-test-ios: export TEST_BINARY_PATH := $(XCODE_DERIVED_DATA)/$(APPLICATION_NAME)/Build/Products/Debug-iphonesimulator/StatusIm.app
 run-visual-test-ios: ##@test Run tests once in NodeJS
-	detox test --configuration ios.sim.debug
```
ibrkhalil, line 322: Is this related?

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -138,13 +162,15 @@
...
                   suggestions          (<sub [:chat/mention-suggestions])
+                  images               (get-in (rf/sub [:chat/inputs]) [chat-id :metadata :sending-image])
                   {window-height :height} (rn/use-window-dimensions)
                   {:keys [keyboard-shown keyboard-height]} (rn/use-keyboard)
                   max-y                (- window-height (if (> keyboard-height 0) keyboard-height 360) (:top insets) (:status-bar-height @navigation-const)) ; 360 - default height
                   max-height           (Math/abs (- max-y 56 (:bottom insets))) ; 56 - top-bar height
                   added-value          (if (and (not (seq suggestions)) (or edit reply)) 38 0) ; increased height of input box needed when reply
                   min-y                (+ min-y (when (or edit reply) 38))
                   y                    (get-y-value context keyboard-shown min-y max-y added-value max-height chat-id suggestions reply)
+                  y                    (+ y (when (seq images) 80))
```
ibrkhalil: duplicated var ?

src/status_im/ui2/screens/chat/photo_selector/view.cljs
```diff
@@ -78,14 +55,21 @@
    (when (some #{item} @selected)
      [info-count/info-count (+ (utils/first-index #(= item %) @selected) 1) (style/image-count)])])
 
+(defn on-end-reached [end-cursor]
+  (let [is-loading    (rf/sub [:camera-roll-loading-more])
+        has-next-page (rf/sub [:camera-roll-has-next-page])]
```
ibrkhalil: bools?
Should we add a question mark at the end ?

ibrkhalil:
Will do a thorough review later tonight

src/status_im/ui2/screens/chat/composer/style.cljs
```diff
@@ -80,3 +80,14 @@
          (when-not pin? {:position :absolute
                          :left     34
                          :top      3})))
+
+(defn remove-photo-container []
```
flexsurfer: def

src/status_im/ui2/screens/chat/photo_selector/view.cljs
```diff
@@ -35,24 +23,13 @@
...
-                         :on-press #(js/alert "Add text: to be implemented")}
-            (i18n/label :t/add-text)]
-           [quo2/button {:style    {:flex 0.48}
-                         :before   :send
-                         :on-press #(do
-                                      (rf/dispatch [:chat.ui/send-current-message])
-                                      (reset! selected [])
-                                      (rf/dispatch [:bottom-sheet/hide]))}
-            (str (i18n/label :t/send) " " (when (> (count @selected) 1) (count @selected)))]]])))])
+          [quo2/button {:style    {:width "95%"
```
flexsurfer: we shouldn't use percentage, only margins and paddings

src/status_im/ui2/screens/chat/photo_selector/view.cljs
```diff
@@ -35,24 +23,13 @@
...
+                                   :align-self :center}
+                        :on-press #(do
+                                     ;(rf/dispatch [:chat.ui/send-current-message])
```
flexsurfer: comment?

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -117,6 +120,27 @@
                  translate-y
                  (reanimated/with-timing (- max-y)))))))))))
 
+(defn small-image [item]
```
flexsurfer: better to move  to separate namespace

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -117,6 +120,27 @@
...
+  [rn/view
+   [rn/image {:source {:uri (first item)}
+              :style  {:width         56
```
flexsurfer: move to style ns

src/status_im/chat/models/images.cljs
```diff
@@ -119,13 +121,23 @@
...
+  {:db (assoc db :camera-roll-loading-more is-loading)})
 
 (fx/defn on-camera-roll-get-photos
   {:events [:on-camera-roll-get-photos]}
-  [{db :db} photos]
-  {:db (assoc db :camera-roll-photos (mapv #(get-in % [:node :image :uri]) photos))})
+  [{:keys [db] :as cofx} photos page-info end-cursor]
+  (let [photos_x (when end-cursor (:camera-roll-photos db))]
+    {:db (-> db
+             (assoc :camera-roll-photos (concat photos_x (mapv #(get-in % [:node :image :uri]) photos)))
```
ilmotta: Using `mapv` followed by `concat` is unnecessary, because `concat` will convert the vector back to a lazy seq. Just `map` suffices.

```clojure
user=> (concat [1 2 3] [4 5 6])
(1 2 3 4 5 6)
user=> (concat '(1 2 3) '(4 5 6))
(1 2 3 4 5 6)
```

src/status_im/chat/models/images.cljs
```diff
@@ -119,13 +121,23 @@
...
-  {::camera-roll-get-photos num})
+  [_ num end-cursor]
+  {::camera-roll-get-photos [num end-cursor]})
+
+(fx/defn camera-roll-loading-more
+  {:events [:chat.ui/camera-roll-loading-more]}
+  [{:keys [db]} is-loading]
+  {:db (assoc db :camera-roll-loading-more is-loading)})
 
 (fx/defn on-camera-roll-get-photos
```
ilmotta, line 140: We have ways of testing event handlers and there are examples in the codebase, I would say we should almost always test them. One of the principles of the https://deviq.com/principles/boy-scout-rule is to try to at least not increase tech debt, and if possible reduce it.

src/status_im/chat/models/images.cljs
```diff
@@ -100,13 +100,15 @@
...
  ::camera-roll-get-photos
- (fn [num]
+ (fn [[num end-cursor]]
    (permissions/request-permissions
     {:permissions [:read-external-storage]
      :on-allowed  (fn []
-                    (-> (.getPhotos CameraRoll #js {:first num :assetType "Photos" :groupTypes "All"})
-                        (.then #(re-frame/dispatch [:on-camera-roll-get-photos (:edges (types/js->clj %))]))
-                        (.catch #(log/warn "could not get cameraroll photos"))))})))
+                    (-> (if end-cursor
```
ilmotta, line 107: I usually prefer to see effects that abstract the `on-success` and `on-failure` handlers, so they become more general and are easier to test. This is a known pattern, you can see it, for instance in the http-fx effect https://github.com/day8/re-frame-http-fx/blob/master/src/day8/re_frame/http_fx.cljs#L26

But the really important feedback I have for this effect is that it lacks testing, and since you're touching it, it'd be great to have them.

src/status_im/chat/models/images.cljs
```diff
@@ -77,17 +77,17 @@
  (fn [chat-id]
    (react/show-image-picker
     (fn [^js images]
-      ;; NOTE(Ferossgp): Because we can't highlight the already selected images inside
-      ;; gallery, we just clean previous state and set all newly picked images
+        ;; NOTE(Ferossgp): Because we can't highlight the already selected images inside
```
ilmotta, line 80: There seems to be extra whitespace here.

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -117,6 +120,27 @@
...
+                       :height        56
+                       :border-radius 8
+                       :margin-bottom 20}}]
+   [rn/touchable-opacity
+    {:on-press (fn [] (rf/dispatch [:chat.ui/image-unselected (first item)]))
+     :style    (style/remove-photo-container)}
+    [quo2/icon :i/close {:color colors/white :size 12}]]])
+
+(defn images-list [images]
+  [rn/flat-list {:key-fn     (fn [item] (first item))
```
ilmotta: Just `first` suffices as the `:key-fn`.

du82:
Would you please make the figma files public so I can have a reference to look at? thanks

src/status_im/chat/models/images.cljs
OmarBasem: I need to concat the photos that were read previously with the newly read photos

ilmotta: But what I mean is that it's pointless to call `mapv` (see the `v` in the end). You don't need to convert to a vector if you're going to call `concat`, just use `map` (without the `v`). See my previous example.

OmarBasem:
> Would you please make the figma files public so I can have a reference to look at? thanks

I do not have the privilege to do so. I will ask the design team.

src/status_im/chat/models/images.cljs
OmarBasem: I could be wrong, but I do not think we can use `on-success` and `on-failure` here. This is a JS promise.

As for testing, `permissions/request-permissions` requires reading from an actual device/simulator.

du82:
The selected photos hide when opening a chat

https://user-images.githubusercontent.com/65834069/206220629-9d9758bc-0dd7-47c8-9c3f-93a8a68d4a65.mp4

du82:
Must swipe up to view the confirmation button. Tapping it requires releasing the bottom sheet while tapping repeatedly in the general area in hopes that it will recognize the button tap.

https://user-images.githubusercontent.com/65834069/206365613-a50571c8-ca1c-4c76-af27-a65589f782b3.mp4

src/status_im/chat/models/images.cljs
```diff
@@ -100,13 +100,15 @@
...
    (permissions/request-permissions
     {:permissions [:read-external-storage]
      :on-allowed  (fn []
-                    (-> (.getPhotos CameraRoll #js {:first num :assetType "Photos" :groupTypes "All"})
-                        (.then #(re-frame/dispatch [:on-camera-roll-get-photos (:edges (types/js->clj %))]))
-                        (.catch #(log/warn "could not get cameraroll photos"))))})))
+                    (-> (if end-cursor
+                          (.getPhotos CameraRoll #js {:first num :after end-cursor :assetType "Photos" :groupTypes "All"})
+                          (.getPhotos CameraRoll #js {:first num :assetType "Photos" :groupTypes "All"}))
+                        (.then #(re-frame/dispatch [:on-camera-roll-get-photos (:edges (types/js->clj %)) (:page_info (types/js->clj %)) end-cursor]))
```
ilmotta: You could extract the expensive (?) data converstion of `types/js->clj` into a binding.

src/status_im/chat/models_test.cljs
```diff
@@ -91,3 +92,8 @@
     (testing "Pagination info should be reset on navigation"
       (let [res (chat/navigate-to-chat-nav2 {:db db} chat-id false)]
         (is (nil? (get-in res [:db :pagination-info chat-id :all-loaded?])))))))
+
+(deftest camera-roll-loading-more
```
ilmotta: Just a missing `-test` suffix. `deftest` creates a var, so it's conventional to add a suffix.

src/status_im2/subs/root.cljs
```diff
@@ -100,6 +100,9 @@
 (reg-root-key-sub :chat/inputs :chat/inputs)
 (reg-root-key-sub :chat/memberships :chat/memberships)
 (reg-root-key-sub :camera-roll-photos :camera-roll-photos)
+(reg-root-key-sub :camera-roll-end-cursor :camera-roll-end-cursor)
```
ilmotta: Not suggesting anything here, but does anybody know if adding more root keys to the graph can meaningfully impact performance?

Is it generally better to have less root keys? Would it better for the performance of the app to create just one `:camera-roll` root key and have the `next-page` and `loading-more` keys inside it?

My intuition tells me that it's better to have less root keys for things that are interconnected, so that we can better leverage subscription caching.

@flexsurfer cc

src/status_im/ui2/screens/chat/photo_selector/view.cljs
```diff
@@ -78,14 +54,21 @@
    (when (some #{item} @selected)
      [info-count/info-count (+ (utils/first-index #(= item %) @selected) 1) (style/image-count)])])
 
+(defn on-end-reached [end-cursor]
```
ilmotta: @OmarBasem, I really think this kind of logic should be in event handlers. So you can define an event, and in the `:on-end-reached` of the flat list you can dispatch it. For an example, see `src/status_im/ui/screens/activity_center/views.cljs`, function `activity center`.

shadow-cljs.edn
```diff
@@ -82,6 +82,7 @@
            :output-dir       "target/test"
            :optimizations    :simple
            :target           :node-test
+           :ns-regexp        "status-im.chat.models-test$"
```
ilmotta: This is probably a left over from your local experiments.

src/status_im2/subs/root.cljs
flexsurfer: yes i just had the same feeling lately reviewing PRs,  we should group props in root keymap

flexsurfer: would be cool to add this to guidelines

src/status_im/chat/models/images.cljs
ilmotta: @OmarBasem, sorry this is a relatively long topic. Please, don't worry about re-implementing this effect in this PR, it seems to be out of scope for now.

Nevertheless, this topic is quite important, so I'm going to expand a little bit here.

What I mean by my previous comment is that this effect could abstract what to do on the success and fail conditions. This is the recommended approach for effects, i.e. the effect would still use a promise, no problem with that because that's why it exists, to abstract effectful handlers, but consumers of the effect would be able to hook into the success/fail conditions using just raw data. This is usually desirable for effects, because we can make them more flexible and easier to test.

It's important to repeat that re-frame strongly recommends that **all effects should be represented as data**, not data + anonymous functions. This is a mistake in the current implementation of effects in our codebase and it makes testing and inspecting effects harder.

> As for testing, permissions/request-permissions requires reading from an actual device/simulator.

We can always stub hard OS dependencies.

So the following is just an idea I coded now to help you understand what a better effect should look like.  Btw, I haven't tested the implementation below, and there's an edge case with `permissions/request-permissions`, because it incorrectly *sometimes returns a promise, sometimes it doesn't*.

For example:

```clojure
(re-frame/reg-fx
 :camera-roll/get-photos
 (fn [{:keys [num end-cursor on-success on-failure]}]
   (-> (permissions/request-permissions
        {:permissions [:read-external-storage]
         :on-allowed  (fn []
                        (-> (if end-cursor
                              (.getPhotos CameraRoll #js {:first      num
                                                          :after      end-cursor
                                                          :assetType  "Photos"
                                                          :groupTypes "All"})
                              (.getPhotos CameraRoll #js {:first      num
                                                          :assetType  "Photos"
                                                          :groupTypes "All"}))))})
       (.then (fn [^js js-result]
                (when on-success
                  (let [result (types/js->clj js-result)]
                    (re-frame/dispatch (conj on-success {:edges     (:edges result)
                                                         :page-info (:page_info result)}))))))
       (.catch (fn [error]
                 (when on-failure
                   (re-frame/dispatch (conj on-failure error)))) ))))
```

And here's how consumers would more or less use this effect in event handlers. Notice that `on-success` and `on-failure` are **vectors, not anonymous functions**. Everything is data ;)

```clojure
{:camera-roll/get-photos {:num num
                          :end-cursor end-cursor
                          :on-success [:on-camera-roll-get-photos end-cursor]
                          :on-failure [:camera-roll/get-photos-failure]}}
```

---

An evolution of this effect could/should get permissions using an interceptor. Effectful inputs as data as well ;)

shadow-cljs.edn
ilmotta: May I suggest you comment this out and add I quick explanation? This is quite handy for people running tests without the REPL when using `make test-watch`, so I think we could leave it here (commented) for other devs.

src/status_im/chat/models/images.cljs
OmarBasem: @ilmotta Thank you so much for this detailed explanation! ❤️ Lots of useful information.

I have one question, so here we have `.then` followed by `on-success`, like is there any other option? Can't we just drop out the `on-success` and `on-failure` since we are using `then` and `catch`?

src/status_im/ui2/screens/chat/photo_selector/view.cljs
OmarBasem: I think that's what I am doing. The difference is that here we are dispatching based on a condition first (?)

src/status_im/chat/models/images.cljs
ilmotta: @OmarBasem, we can definitely drop the `on-success` and `on-failure` options. It's up to the developer to choose how reusable the effect is. If you want to design an effect that other events can dispatch and react accordingly, then having `on-success`/`on-failure` is the recommended approach in re-frame. But if you just want other events to *fire and forget* (dispatch `::camera-roll/get-photos`) then you can drop those options.

Both alternatives are valid. Without `on-success`/`on-failure` it's also a bit harder to test, but still doable.

src/status_im/ui2/screens/chat/photo_selector/view.cljs
ilmotta: What I mean is that this function shouldn't be part of UI namespaces, does it make sense? It seems to me the function should be defined as an event handler instead.

OmarBasem: @ilmotta I think it is better to keep it that way where it checks for conditions before dispatching. Otherwise, it would be unnecessarily dispatching events (in a spamful way) every time it reaches `end`

Also, the more events are dispatched, the more the performance may be impacted.

ilmotta: I disagree @OmarBasem, I think this is real misconception in the code.

1. Subscriptions have a cost and they should ideally be left for UI components according to re-frame. `on-end-reached` is not a UI component. In fact, it's just a function that needs to read some state and dispatch two effects. This is exactly what an event handler is supposed to do.

2. The `on-end-reached` function will need to be called by re-frame every time either `camera-roll/loading-more` or `camera-roll/has-next-page` state changes and then further dispatch. Every function that subscribes to something will need to be potentially called multiple times due to rendering cycles. This is more work than only dispatching if the RN FlatList detects the onEndReachedThreshold has been reached.

3. If we follow this very logic we would then have to move lots and lots of event handlers to the view layer and rely on needless subscriptions. This would dramatically hurt the performance of the app and break re-frame's recommendations.

ilmotta (approved):

src/status_im/ui2/screens/chat/composer/composer_images/composer_images.cljs
```diff
@@ -0,0 +1,24 @@
+(ns status-im.ui2.screens.chat.composer.composer-images.composer-images
```
flexsurfer: no need to duplicate `composer` in the name, its already under `composer` namespace, and also it should be view.cljs so something like `status-im.ui2.screens.chat.composer.images.view`

flexsurfer: btw you could create it in status-im2 , so we don't need to move it

src/status_im/ui2/screens/chat/composer/composer_images/composer_images.cljs
```diff
@@ -0,0 +1,24 @@
...
+  (:require [react-native.core :as rn]
+            [status-im.ui2.screens.chat.composer.style :as style]
+            [quo2.core :as quo2]
+            [quo2.foundations.colors :as colors]
+            [re-frame.core :as rf]))
+
+(defn image [item]
+  [rn/view
```
flexsurfer: `[:<>` better to avoid using empty views

src/status_im2/subs/root.cljs
```diff
@@ -99,7 +99,10 @@
 (reg-root-key-sub :selected-participants :selected-participants)
 (reg-root-key-sub :chat/inputs :chat/inputs)
 (reg-root-key-sub :chat/memberships :chat/memberships)
-(reg-root-key-sub :camera-roll-photos :camera-roll-photos)
+(reg-root-key-sub :camera-roll/photos :camera-roll/photos)
```
flexsurfer, line 102: its better to have `:camera-roll` map instead of a bunch of root keys

shadow-cljs.edn
```diff
@@ -82,6 +82,8 @@
            :output-dir       "target/test"
            :optimizations    :simple
            :target           :node-test
+           ;; When running tests without a REPL you can uncomment the below line to `make test-watch` a specific file
```
J-Son89, line 85: can this just be passed in as a parameter to the `make test-watch` command ??

ilmotta: Yes, one way is via a environment variable, because ShadowCLJS supports this in the `edn` file.

So for example, something like this should be doable:

```bash
NS_REGEX='status-im.chat.models-test$' make test-watch
```

pavloburykh:
@OmarBasem Hi! Could you please resolve existing conflicts and rebase PR to the latest develop. After that we can take it into test.  Thanx!

shadow-cljs.edn
ilmotta: @J-Son89 ^

du82:
> Must swipe up to view the confirmation button. Tapping it requires releasing the bottom sheet while tapping repeatedly in the general area in hopes that it will recognize the button tap.
> 
>  22-12-07-23-38-49.mp4

Still having this issue

src/status_im/ui2/screens/chat/photo_selector/view.cljs
OmarBasem: Thanks @ilmotta for the explanation. Fixed it.

src/status_im/ui2/screens/chat/composer/composer_images/composer_images.cljs
OmarBasem: It's not an empty view really. It does not look right if I use `[:<>` instead of `view`. There is a button that is positioned absolutely relative to that view.

src/status_im/ui2/screens/chat/composer/composer_images/composer_images.cljs
OmarBasem: Better move the entire composer together?

OmarBasem:
> > Must swipe up to view the confirmation button. Tapping it requires releasing the bottom sheet while tapping repeatedly in the general area in hopes that it will recognize the button tap.
> > 22-12-07-23-38-49.mp4
> 
> Still having this issue

That's an issue with the bottom sheet component. Not part of this PR.

OmarBasem:
@pavloburykh ready to be tested

VolodLytvynenko:
Hi @OmarBasem thank you for your work. Here are the following found issues:
 
**ISSUE 1 The 'confirm selection' button is not shown for android devices**
**Steps to reproduce:**
1. Open 1-1 chat
2. tap image icon
3. Select any photo within device

**Actual result:**
The 'confirm selection' button is not shown
image

**Expected result:**
The 'confirm selection' button is shown
image

**Env: Samsung Galaxy S20 v10, Huawei p20 light v9**


**ENV:** Samsung Galaxy S20 v10

**ISSUE 2 The header is visible during selector is opened**
**Steps to reproduce:**
1. Open 1-1 chat
2. tap image icon

**Actual result:**
Header is shown when selector is opened
image


**Expected result:**
Header is not shown per design
image

**Env: Samsung Galaxy S20 v10, Huawei p20 light v9**

**ISSUE 3: Images are sent in the message even if [confirm selection] button is not tapped**
**Steps to reproduce:**
1. Open 1-1 chat
2. tap image icon
3. Select any images
4. Collapse the selection pop up
5. Type any message into textbox -> tap 'send' button

**Actual result:**
The non-confirmed images are sent together with text
https://user-images.githubusercontent.com/52490791/208091697-8d42d4cb-185f-4374-9314-50c1f5352fd6.mp4


**Expected result:**
The images are not sent in the message if current images are not selected

**ENV:** All devices

**ISSUE 4: The selector is blinking after its opening if a lot of photos on the device are set**
**Steps to reproduce:**
1. Set at least 50 photos into your device
2. Open 1-1 chat
3. Open the selector

**Actual result:**
The selector is blinking after its opening
https://user-images.githubusercontent.com/52490791/208098745-95db2e9e-1e60-4d93-a386-e458cac2c664.mp4

**Expected result:**
The selector is not blinking

**ENV:** Iphone 11 pro v16

**ISSUE 5 the [send] button does not appear if only images are added** 
1. Open 1-1 chat
2. Open the selector
3. Select image
4. Confirm selection
5. Try to send the image without text

**Actual result:**
Send button is not shown if only images are selected
image

**Expected result:**
The send button is shown if only images are in the textbox
image

**ISSUE 6. Photo is not shown as selected if selector is reopened**
1. Open 1-1 chat
2. Open the selector
3. Select image
4. Confirm selection
5. Tap on the 'image' icon again to open the selector

**Actual result:**
The image is shown as unselected into the opened selector

https://user-images.githubusercontent.com/52490791/208105092-cf8e45b8-fe43-4b9a-8e0a-6f629e5120d8.mp4

**Expected result:**
The image is still shown as selected if current image is shown in the textbox and selector is reopened
image

**ISSUE 7: All images are not shown into textbox after user select more than 5 photos**
**Steps to reproduce:**
1. Go to the 1-1 chat.
2.  Select at least 6-7 images
3. Confirm selection
4. Check count of selected images within textbox

**Actual result:**
only 5 photos are saved within textbox
https://user-images.githubusercontent.com/52490791/208107002-1f196e5e-1e9a-4951-a563-c759231336d1.mp4


**Expected result:**
All selected images are saved

**Questions:**
**1. Should the access error message be implemented in the scope of this PR?**

**Actual behavior:**
How it looks now if access is denied

**Expected behavior**
How design requires
image

**2. Should the long press within selector on the image be implemented in the scope of this PR?**

**Actual result:**
Nothing happens after long press is done

**Expected behavior:**
image

**3. Should the sent images be shown as per design in the scope of this PR? Currently, they are shown separately and the image is shown upper the image instead of under**

**Actual behavior:**
image

**Expected behavior**
image

src/status_im/chat/models/images.cljs
```diff
@@ -117,15 +120,32 @@
...
   {:events [:on-camera-roll-get-photos]}
-  [{db :db} photos]
-  {:db (assoc db :camera-roll-photos (mapv #(get-in % [:node :image :uri]) photos))})
+  [{:keys [db] :as cofx} photos page-info end-cursor]
+  (let [photos_x (when end-cursor (:camera-roll/photos db))]
+    {:db (-> db
+             (assoc :camera-roll/photos (concat photos_x (map #(get-in % [:node :image :uri]) photos)))
+             (assoc :camera-roll/end-cursor (:end_cursor page-info))
+             (assoc :camera-roll/has-next-page (:has_next_page page-info))
+             (assoc :camera-roll/loading-more false))}))
```
ibrkhalil, line 148: Can this be done in one `assoc`?

src/status_im/ui2/screens/chat/composer/images/view.cljs
```diff
@@ -0,0 +1,24 @@
+(ns status-im.ui2.screens.chat.composer.images.view
+  (:require [react-native.core :as rn]
+            [status-im.ui2.screens.chat.composer.style :as style]
+            [quo2.core :as quo2]
+            [quo2.foundations.colors :as colors]
+            [re-frame.core :as rf]))
+
+(defn image [item]
+  [rn/view
```
ibrkhalil, line 9: Can this be a React fragment ` [:<> ...]`?

ibrkhalil (approved):

OmarBasem:
@VladimrLitvinenko thanks for your reviews! I have fixed the issues. Please let me know if there are any more issues.

Issue 1: fixed (please note the images are not shown in the input on Android, which is not part of this PR)
Issue 2: a new bottom sheet is currently under development, not part of this PR
Issue 3: fixed
Issue 4: fixed
Issue 5: fixed
Issue 6: fixed
Issue 7: currently sending more than 5 photos is not supported, needs backend implementation
Questions 1,2,3: not part of this PR

VolodLytvynenko:
@OmarBasem thank you for PR. Issues 3,4,5,6 are fixed and PR ready to be merged
Take a look, one small priority issue still exist. I believe it can be fixed separately as follow up https://github.com/status-im/status-mobile/issues/14580

OmarBasem:
Thank you @VladimrLitvinenko for your reviews!

OmarBasem:
@flexsurfer I need to merge this one now for a dependency PR. If there is still any issue with this PR please let me know and I will fix it in the following PR.

PR #14511, opened 2022-12-08
chore: add some component tests

J-Son89:
I added some more component tests to expand on the examples we have and increase the coverage

I will add more documentation on this in the coming days about the tools used for these tests and resources to check when writing them.

### DIFF
--- a/src/quo2/components/counter/__tests__/counter_component_spec.cljs
+++ b/src/quo2/components/counter/__tests__/counter_component_spec.cljs
@@ -1,0 +1,34 @@
+(ns quo2.components.counter.--tests--.counter-component-spec
+  (:require ["@testing-library/react-native" :as rtl]
+            [quo2.components.counter.counter :as counter]
+            [reagent.core :as reagent]))
+
+(defn render-counter
+  ([]
+   (render-counter {} nil))
+  ([opts value]
+   (rtl/render (reagent/as-element  [counter/counter opts value]))))
+
+(js/global.test "default render of counter component"
+                (fn []
+                  (render-counter)
+                  (-> (js/expect  (rtl/screen.getByTestId "counter-component"))
+                      (.toBeTruthy))))
+
+(js/global.test "renders counter with a string value"
+                (fn []
+                  (render-counter {} "1")
+                  (-> (js/expect  (rtl/screen.getByText "1"))
+                      (.toBeTruthy))))
+
+(js/global.test "renders counter with an integer value"
+                (fn []
+                  (render-counter {} 1)
+                  (-> (js/expect  (rtl/screen.getByText "1"))
+                      (.toBeTruthy))))
+
+(js/global.test "renders counter with value 99+ when the value is greater than 99"
+                (fn []
+                  (render-counter {} "100")
+                  (-> (js/expect  (rtl/screen.getByText "99+"))
+                      (.toBeTruthy))))
--- a/src/quo2/components/counter/counter.cljs
+++ b/src/quo2/components/counter/counter.cljs
@@ -38,7 +38,8 @@
                      1     16
                      2     20
                      28)]
-    [rn/view {:accessible true
+    [rn/view {:test-ID :counter-component
+              :accessible true
               :accessibility-label accessibility-label
               :style (cond-> (merge
                               {:align-items     :center
--- a/src/quo2/components/drawers/action_drawers.cljs
+++ b/src/quo2/components/drawers/action_drawers.cljs
@@ -9,14 +9,16 @@
     colors/danger-50
     (colors/theme-colors colors/neutral-50 colors/neutral-40)))
 
-(defn divider []
+(def divider
   [rn/view {:style {:border-top-width 1
                     :border-top-color (colors/theme-colors
                                        colors/neutral-10 colors/neutral-90)
                     :margin-top       8
                     :margin-bottom    7
                     :align-items      :center
-                    :flex-direction   :row}}])
+                    :flex-direction   :row}
+            :accessible true
+            :accessibility-label :divider}])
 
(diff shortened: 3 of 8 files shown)


src/quo2/core_spec.cljs
```diff
@@ -1,2 +1,5 @@
 (ns quo2.core-spec
-  (:require [quo2.components.selectors.--tests--.selectors-component-spec]))
+  (:require [quo2.components.selectors.--tests--.selectors-component-spec]
```
J-Son89, line 2: new test suites get added here

src/quo2/components/counter/counter.cljs
```diff
@@ -38,7 +38,8 @@
                      1     16
                      2     20
                      28)]
-    [rn/view {:accessible true
+    [rn/view {:test-ID "counter-component"
```
ilmotta: I tried changing `:test-ID` to a keyword and it works just fine in production code. Could you please change all current test IDs to keywords? This is more idiomatic in Clojure. Not in the tests of course, because that layer doesn't like certain Clojure idioms.

src/quo2/components/counter/__tests__/counter_component_spec.cljs
```diff
@@ -0,0 +1,34 @@
+(ns quo2.components.counter.--tests--.counter-component-spec
```
ilmotta, line 1: A heuristic I use to decide *what to test* in components is to think about input -> output, especially for pure functions or even in Javascript (components without state, hooks, etc).

The `counter` component is a pure function and it receives the following arguments, but not all of them are covered:

- `type`
- `override-text-color`
- `override-bg-color`
- `style`
- `accessibility-label`
- `value`

The tests should ideally cover the impact in the output of each input (in tandem or isolated, it doesn't matter much) because the **output only changes if the input changes** (pure function).

But I understand you're just trying to increase test coverage and give more examples, which is great :100: The other heuristic I saw you used in other tests is to test the behavior of event triggers, this is also a solid approach for stateful components!

src/quo2/components/drawers/action_drawers.cljs
```diff
@@ -9,36 +9,40 @@
...
                                                :margin-horizontal 8}
                               :underlay-color (colors/theme-colors colors/neutral-5 colors/neutral-90)
                               :on-press       on-press}
       [rn/view {:style
                 {:height (if sub-label 58 50)
-
                  :margin-horizontal 12
                  :flex-direction    :row}}
-       [rn/view {:style
+       [rn/view {:test-ID (str "icon-" accessibility-label)
```
ilmotta: I would argue it's generally better to avoid dynamic test IDs. We lose the ability to search for them, which is very useful once the test suit grows in size. Of course there are exceptions, but in the `action-drawers` tests we see all of them test only a single section. But even with multiple sections, it's trivial to use something like `getAllByTestId` and match against the collection.

So unless there's no way to avoid dynamic test IDs, I think we should avoid them. Well, that's just my experience using the react testing library.

src/quo2/components/selectors/selectors.cljs
```diff
@@ -108,7 +111,9 @@
                                       (colors/theme-colors colors/white
                                                            (colors/alpha colors/neutral-80 0.4)))})
          :accessibility-label (str "radio-" (if @checked? "on" "off"))
-         :accessibility-role  :checkbox}
+         :accessibility-role  :checkbox
+          :testID "radio-component"
+         }
```
ilmotta: This needs some formatting.

ilmotta (approved):
Such great additions, thanks @J-Son89!

src/quo2/components/counter/counter.cljs
J-Son89: cool, yeah I will try it out! I'm pretty sure it should work

src/quo2/components/counter/__tests__/counter_component_spec.cljs
J-Son89: Yep, I 100% agree. ideally we have a test to cover each combination of various props.
I was trying to figure out how best to test pieces like `override-text-color` without too tightly binding to the implementation. Actually that's where I feel visual tests will work better as a testing mechanism. We can figure out an approach that best suits us though, as obviously visual tests will be much slower to run so perhaps better to do with tests such as these.

But yes, this p.r (and some follow ups) are just to create some broad coverage on components and have at the very minimum basic rendering covered.

src/quo2/components/drawers/action_drawers.cljs
J-Son89: Yeah I think you are right about this. React Native Testing library says it's best to use TestID's as a last option too. I will try your solution, seems like a good option! :)

src/quo2/components/counter/__tests__/counter_component_spec.cljs
ilmotta: Thank you so much, incredibly valuable for us to eventually have a good "test pyramid".

vkjr:
> I will add more documentation on this in the coming days about the tools used for these tests and resources to check when writing them.

@J-Son89, I would very much appreciate it if you add a couple of sentences and links regarding the used test approach to the PR description. So reviewers who are not familiar (like me) could learn a bit)

J-Son89:
> > I will add more documentation on this in the coming days about the tools used for these tests and resources to check when writing them.
> 
> @J-Son89, I would very much appreciate it if you add a couple of sentences and links regarding the used test approach to the PR description. So reviewers who are not familiar (like me) could learn a bit)

absolutely, I'm in the process of writing some documentation which will hopefully can help those less familiar. I will include some examples that we have in the codebase so it's somewhat more familiar to our devs :)

Issue #14455, opened 2022-11-29
Spike: Research captchas to use for communities join screens


J-Son89:
See the designs:
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7643%3A608085&t=bc5QBHpQ3uZT8W6b-0

There is the use of a captcha in the designs for the communities join screens. 

It is still unclear what Captcha service we want to use. Research some possible solutions, including:

https://www.hcaptcha.com/
also consider an in house solution.

Definition of done:
Research possible solutions
Document these solutions
Bring forward solutions to team and analyse the pros/cons

J-Son89:
https://github.com/hCaptcha/react-native-hcaptcha

du82:
Captchas will not solve the spam problem, because the spammers are using a forked desktop client with automation tools. All captchas do is make the app less private, and more annoying for the average user.

All captchas will do is make using Status over LAN impossible, and make it easier for censors to discover users.

If we *must* have captchas, why not make it optional and let individual communities choose if they want to enable it?

J-Son89:
There was an agreement to postpone this research until post mvp

PR #14513, opened 2022-12-09
add Non Token Gated - Manual Approval Required flow

### LINKED ISSUE #14459: Communities join screens (Non token gated) - Add Manual approval flow designs
See the steps for this flow in: https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7035%3A462899&t=4pBzp6AcBrgJRPmM-0

and ensure all screens of the flow are considered.

Sync up with design team before starting the work to ensure there is no additional details that are not clear from the designs.

Somethings to watch for:
text changing in many places from "Join" to "Request to join" etc
action buttons
page titles
request to leave
notifications for pending requests
cancel request flow

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14459

This implements the copy for the flow of Non Token Gated - Manual Approval Required

To **test** this:

Go to the community screen of a community which requires that an admin approves the request.

See that the copy is adjusted when manual approval is required.
It should match the flow of the designs here:
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7035%3A462899&t=jzuZepNc60qisfSe-0


J-Son89:
Bottom sheet copy for this flow
Screenshot 2022-12-09 at 13 34 16

Community overview copy and message adjusted
Screenshot 2022-12-09 at 13 34 10

Discover card adjusted to show community icon again
Screenshot 2022-12-09 at 13 34 01

src/quo2/components/community/icon.cljs
```diff
@@ -0,0 +1,11 @@
+(ns quo2.components.community.icon
```
J-Son89, line 1: I moved the community icon to quo2 @flexsurfer 

Removed a lot of the excess code in it, this doesn't handle the "Status" default community icon, but imo we should move away from this community and start using real data as we are adding to many branches to the code just for this community

src/status_im2/contexts/communities/discover/view.cljs
```diff
@@ -116,10 +115,9 @@
 (defn discover []
   (let  [view-type (reagent/atom :card-view)]
     (fn []
-      (let [communities (rf/sub [:communities/communities])
-            communities-count (count communities)
-        ; TODO move sorting to subscription
```
J-Son89: actually, do we really want sorting in the subs? This seems like unnecessary computation we probably don't want for composing subs.. Right @ilmotta , @cammellos , @flexsurfer ? 🤔

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -79,16 +85,17 @@
     :separator [rn/view {:margin-top 1}]
     :render-fn                         community-rule-item}])
 
-(defn request-to-join [community]
-  (let [agreed-to-rules? (reagent/atom false)]
+(defn request-to-join [{:keys [permissions name id]}]
+  (let [agreed-to-rules? (reagent/atom false)
+        is-open? (not= 3 (:access permissions))]
```
J-Son89: backend permissions for closed/open community. It would probably be nicer if they used a semantic naming for this instead. Can we provide feedback on that? @cammellos, @ilmotta 
wdyt?

src/status_im2/subs/communities.cljs
```diff
@@ -77,13 +77,19 @@
  (fn [[communities-enabled? search-filter communities]]
    (filterv
     (fn [{:keys [name featured id]}]
-      (and (or featured (= name "Status"))                  ;; TO DO: remove once featured communities exist
+      (and (or featured (= name "Status")) ;; TO DO: remove once featured communities exist
```
J-Son89: will revert this change, but really I would prefer to nuke this Status community from subs 👿

src/status_im2/contexts/communities/discover/view.cljs
flexsurfer: actually its my comment, views should be dumb, all computation should be in subscriptions, so there should be `:communities/sorted-discover-communities` subscription

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -6,8 +6,6 @@
    [utils.re-frame :as rf]
    [quo2.foundations.colors :as colors]
    [status-im2.contexts.communities.overview.style :as style]
```
flexsurfer: please keep ;; TODO move to status-im2 when reimplemented, this comment for all requirement under it

src/status_im2/subs/communities.cljs
flexsurfer: i think thats fine to nuke why not ?

src/status_im2/contexts/communities/overview/style.cljs
```diff
@@ -14,4 +14,19 @@
                                   :top (if platform/ios? 56 60)
                                   :height 34
                                   :width "100%"
-                                  :flex 1}})
\ No newline at end of file
+                                  :flex 1}})
+
+(defn scroll-view-container [border-radius] {:position :absolute
```
flexsurfer: could you please move map to the new line?

src/quo2/components/community/icon.cljs
```diff
@@ -0,0 +1,11 @@
...
+  (:require [react-native.core :as rn]))
+
+(defn community-icon [{:keys [images]} size]
+  (let [thumbnail-image (get-in images [:thumbnail :uri])]
+    [rn/image {:source {:uri thumbnail-image}
+               :style  {:border-radius 50
+                        :border-width 0
+                        :border-color :transparent
+                        :width size
+                        :height size}}]))
```
ilmotta, line 11: At some point we need to at least agree about trailing newlines at the end of files and configure our editors properly. I think the arguments are strongly in favor of a trailing newline, no exceptions. Even Github is hinting at us with a red icon ;)

Suggestion for reading:
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

src/status_im2/contexts/communities/discover/view.cljs
ilmotta: I prefer to think that sorting is not the view responsibility. The view just cares about getting the data and transforming it to Hiccup.

If we move the sorting logic out of the subscription, we actually lose the caching facilities because now sort will be computed on every render (and they happen a lot).

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,11 +157,14 @@
 
 (def channel-list-component (memoize channel-list-component-fn))
 
-(defn render-page-content [{:keys [name description locked joined id images
-                                   status tokens tags community-color]
+(def request-to-join-text (memoize (fn [is-open?] (if is-open? (i18n/label :t/join-open-community) (i18n/label :t/request-to-join-community)))))
```
ilmotta: This line is unreadable to me, could you please break it into multiple lines?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,11 +157,14 @@
...
+
+(defn render-page-content [{:keys [name description locked joined id images can-join?
+                                   status tokens tags community-color permissions]
                             :as   community}
                            channel-heights first-channel-height]
-  (let [thumbnail-image (get-in images [:thumbnail])]
+  (let [is-open? (not= 3 (:access permissions))
```
ilmotta: What is this [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) `3`? It looks like it should be refactored into a constant.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -238,19 +239,24 @@
...
-            :before :i/communities}
-           (i18n/label :t/join-open-community)])]
+        (when (and (not joined) can-join?)
+          [:<>
+           [quo/button
+            {:on-press  #(rf/dispatch [:bottom-sheet/show-sheet
+                                       {:content (constantly [requests.actions/request-to-join community])
+                                        :content-height 300}])
+             :override-background-color community-color
+             :style
```
ilmotta: This part looks confusing, the other keys have the values in the same line, so could you join the `:style` line with the next one?

```clojure
 :style {:width        "100%"
         :margin-top   20
         :margin-left  :auto
         :margin-right :auto}
```

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -33,6 +33,12 @@
                        :title "No offensive names and profile pictures"
                        :content "You will be asked to change your name or picture if the staff deems them inappropriate."}])
 
+(def request-to-join-text (memoize
```
ilmotta: I don't think we need to memoize this function. The expensive `i18n/label` calls are already memoized. It's often a mistake to memoize what's already memoized, it causes more overhead for nothing.

src/status_im2/contexts/communities/requests/actions/view.cljs
ilmotta: Great question @J-Son89. As usual, it depends, but I stand firmly in the ground that we should almost always create a well defined model of the data coming INTO the application. If the backend sends a `3`, `:three` or `1+1+1` it doesn't matter, our data model should make sure the so called `3` is transformed into something with semantic meaning as you said.

I made a comment above related to this magic number 3, I think there's another place where this number appears.

At the very least, we could extract the [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) to a constant with a proper name.

We also have the `src/data_store/` directory, where various namespaces exist to implement some kind of adapter pattern or anti-corruption layer.

cammellos: Yes, this should be definitely `const` somewhere (we have many in `constants` namespace), we use enums in status-go because there is more idiomatic, but here is best to always use a reference/variable rather than the actual value

src/status_im2/contexts/communities/overview/view.cljs
cammellos: Also, not sure memoizing is worth it here, just out of curiosity, have you tested the difference in speed? 
(a quick and dirty way, which should give you an idea on whether they are at least comparable is doing something like:
```
(time (dotimes (range 0 10000) (get {:a 1} :a))) // insert inside the operation you care about
```
They both look like map lookups I suppose, so I'd expect both to run in constant time, though I could be wrong

ilmotta: You are right, and memoizing what's memoized is no good at all. `i18n/label` expensive calls are already memoized, it's just an overhead to memoize `request-to-join-text`.

In essence, here we're memoizing an `if` expression checking a boolean, the rest is already memoized. And of course we shouldn't memoize boolean checks.

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: sure will do, I'm really waiting for the auto formatter 😢

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: I will remove the memo 👍

src/quo2/components/community/icon.cljs
J-Son89: Yeah  I agree that we need to agree, but I don't think this is something I (or any other dev) should have to even think about. `make lint-fix` should address any of these formatting issues and any errors we find in p.r's should be handled by this tool. In the case there's some formatting we don't like the tool should be updated.

Perhaps I am being too optimistic and this is not possible in our current tech stack?? 🤔 

@yqrashawn - the auto formatter you are working on can handle this case?

yqrashawn: This kind of setting is usually set in editor. There's https://editorconfig.org/ for this, but generally user need to install an editor-config plugin for that.

VS Code https://stackoverflow.com/a/54270768/5671288
IntelliJ https://stackoverflow.com/a/16761228/5671288
seems ~Emacs and~ vim does this by default
Emacs check `require-final-newline`

ilmotta: I think there are two levels of automation here.

1. The developer's text editor/IDE applies basic formatting during development. This is nearly instantaneous. Newlines at the end of files should be handled at this level.
2. The code format tool (zprint for example) checks the entire codebase and see if it's formatted appropriately. This is slow and we should rely on it only as a gate keeper for continuous integration.

Hence, relying only on `make lint` would be impractical IMO (way too slow). Even in Javascript land, users usually run Prettier or another formatter on every save of the file and only rely on `make lint` for CI. This is what zprint would bring to us that @yqrashawn is working on.

src/status_im2/contexts/communities/overview/view.cljs
ilmotta: Haha Me too, it'll come

src/quo2/components/community/icon.cljs
J-Son89: yeah, I agree. Imo we are just adding noise to our p.r's with formatting issues, when we could automate this process. if we can not automate this with `make lint-fix` then at the very least `make lint` should catch these issues and it is up to the dev to "manually" fix these issues. 
i.e the link check step on ci jobs will fail and this will notify the dev. Reviews can then focus on improving the quality of the code.

Is it just a case of we're not quite there yet, or is there something I am missing here?

J-Son89: you mean, Hence, relying only on `make lint-fix` .. ?

ilmotta: Maybe my comment wasn't clear enough. We will always rely on `make lint`, but as the last step to check. Relying on `make lint-fix` would be suboptimal during development, unless you (or any other dev) prefer that way. I never want to use `make lint-fix` in my daily development, I want my editor to automatically format the file I'm currently editing.

`make lint-fix` is slow compared to instant updates provided by IDEs/editors and you'll have to run it either on every commit or generate separate "fix lint" commits, but of course, it's up to you. Solutions with commit hooks fall in the same category, they make my interactions with git slow and I would disable them during development.

> yeah, I agree. Imo we are just adding noise to our p.r's with formatting issues

> Is it just a case of we're not quite there yet, or is there something I am missing here?

Unfortunately getting the formatter correct and capable of handling our own preferences (Status contributor preferences) is far from trivial and @yqrashawn had to actually report various issues to the maintainer of the tool. @yqrashawn just now opened a PR with new changes to introduce zprint o/

yqrashawn: It's definitely possible to check this in `make lint` and fix this in `make lint-fix`.

src/status_im2/contexts/communities/requests/actions/view.cljs
J-Son89: yep, it was already there. Found it, thanks! :)

src/status_im2/subs/communities.cljs
J-Son89: ok, nuking! =D

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: I decided to remove the code that needed these imports as it was just for mock data. So no more status_im in this file 🧹

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: resolved above

src/status_im2/contexts/communities/home/view.cljs
```diff
@@ -14,7 +14,7 @@
      {:on-press      (fn []
                        (rf/dispatch [:communities/load-category-states id])
                        (rf/dispatch [:dismiss-keyboard])
-                       (rf/dispatch [:navigate-to-nav2 :community {:community-id id}]))
+                       (rf/dispatch [:navigate-to :community-overview id]))
```
J-Son89: @flexsurfer, @jo-mut updated the navigation on the communities home page so it goes directly to the new community overview page

src/status_im2/subs/communities_test.cljs
```diff
@@ -109,3 +109,24 @@
             "0x3" {:id "0x3" :name "Civilized dolphins"}})
     (is (= {:joined ["0x1"], :pending ["0x2"], :opened ["0x3"]}
            (rf/sub [sub-name])))))
+
+(h/deftest-sub :communities/sorted-communities
```
J-Son89, line 119: @ilmotta added the sorted communities subscription and this test. It's alright? :)

src/status_im2/contexts/communities/home/view.cljs
flexsurfer: cc @Parveshdhull

Parveshdhull: Hi @J-Son89,
Please use `(re-frame/dispatch [:communities/navigate-to-community {:community-id (:id community)}])`
and change `:community` to `:community-overview` in`:communities/navigate-to-community` event.

> so it goes directly to the new community overview page

Also, is the new community overview functional? I tried PR and looks like the community screen just has mocked data and channels are not opening. Is this intentional?

J-Son89: thanks @Parveshdhull, it's not functional just yet. It is mostly using mock data. However I think it should be  it's a small adjustment to get it working.

For now, I will revert this change in this p.r and look to make this and other necessary changes in a follow up pr

src/status_im2/subs/communities_test.cljs
ilmotta: Looks like a useful test @J-Son89!

ilmotta (approved):

src/status_im2/subs/communities_test.cljs
J-Son89: thanks for checking. and for helping set it up. Had not looked at these tests before and it was easy to figure out how to use it from the other tests 👍

src/status_im2/contexts/communities/home/view.cljs
J-Son89: @vkjr - just for reference :)

pavloburykh:
Hey @J-Son89! Thanx for the PR. Please, take a look at the following issues:

### ISSUE 1 Cancel/Request join buttons are cut and slightly visible on Android
Tested on Android Galaxy A52, Android v. 11

**Steps:** 

1. Invite User A (Android) to community that requires approval (community A)
2. Open Discover community screen by User A 
3. Find community A in the list and open it 
4. Tap Request to join community button
5. Pay attention if Cancel/Request community buttons are visible

**Actual result:** Cancel/Request join buttons are cut and slightly visible on Android

photo_2022-12-15 21 14 03

pavloburykh:
### ISSUE 2 Requesting approval to join community does not result in sending join request to community owner

**Steps:**

1. Create Require approval community (community A) by User A
2. Invite User B to the community
3. By User B open Discover community screen, find community A and Request a join
4. Pay attention if User B status changed to waiting for approval request and User A received a request

**Actual result:** User B is shown active Request community button. User A has not received join community request from User B. 


https://user-images.githubusercontent.com/97245802/207948149-1a815276-1403-4028-80a7-2fdc53f0fb99.mp4

pavloburykh:
### ISSUE 3 Captcha is not displayed during requesting community join

Steps: 

1. Open Request community screen
2. See if captcha is shown

**Actual result:** captcha is not displayed.
photo_2022-12-15 21 24 52


**Expected result:** according to designs  - captcha should be shown
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7643%3A611607&t=d3tkKBjSTre4zqQ7-0

Communities for Mobile – Figma 2022-12-15 21-21-01

pavloburykh:
### QUESTION Should this PR implement logic with online/offline owner nodes?

According to [Figma](https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7035%3A462899&t=d3tkKBjSTre4zqQ7-0) user should see different info and different types of buttons depending if owner node is online or offline. 

But current seems not to follow this logic and always shows Request to join community button regardless of owner node online/offline. Also, the info below the button is different from designs. 
photo_2022-12-15 21 31 30

pavloburykh:
### ISSUE 4 Info icon on Community rules screen does not correspond to designs 

Steps: 

1. Open Community rules screen by requesting community join
2. Pay attention at info icon on Community rules screen

**Actual result:** Info icon is displayed on Community rules screen. It is not tappable and does not provide any info to User. 

photo_2022-12-15 21 37 05

Expected result: 

According to designs there is no info icon on Community rules screen. 
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7643%3A608085&t=d3tkKBjSTre4zqQ7-0

Communities for Mobile – Figma 2022-12-15 21-38-04

J-Son89:
Hi @pavloburykh , thanks very much for looking into this pr :D 

Issue 1 - (buttons not visible on join bottom sheet) will be handled in this issue: https://github.com/status-im/status-mobile/issues/14564
The reason it needs to be handled separately is because it is dependent on the new quo2 component for bottom sheet.
Could you at least scroll to that button? i.e functionality is not blocked? -> If not I will address that at least scrolling is working on android in this p.r

Issue 2 I will address as it is part of this p.r and I forgot to include this request. Good spot :) 

Issue 3 - Captcha is still being researched and it's being tracked here: 
https://github.com/status-im/status-mobile/issues/14455

Issue 4 is beyond the scope of this pr but I have created an issue to track it being implemented correctly.
https://github.com/status-im/status-mobile/issues/14565

J-Son89:
> ### QUESTION Should this PR implement logic with online/offline owner nodes?
> According to [Figma](https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7035%3A462899&t=d3tkKBjSTre4zqQ7-0) user should see different info and different types of buttons depending if owner node is online or offline.
> 
> But current seems not to follow this logic and always shows Request to join community button regardless of owner node online/offline. Also, the info below the button is different from designs. photo_2022-12-15 21 31 30

Yeah I was also wondering about this. I will check with design team :)

cc @vkjr as he is working on the info box -

pavloburykh:
> Hi @pavloburykh , thanks very much for looking into this pr :D
> 
> Issue 1 - (buttons not visible on join bottom sheet) will be handled in this issue: #14564 The reason it needs to be handled separately is because it is dependent on the new quo2 component for bottom sheet. Could you at least scroll to that button? i.e functionality is not blocked? -> If not I will address that at least scrolling is working on android in this p.r
> 
> Issue 2 I will address as it is part of this p.r and I forgot to include this request. Good spot :)
> 
> Issue 3 - Captcha is still being researched and it's being tracked here: #14455
> 
> Issue 4 is beyond the scope of this pr but I have created an issue to track it being implemented correctly. #14565

Hi @J-Son89 ! Thank you for answers. Regarding ISSUE 1 - the buttons are slightly visible and I am able to tap them, so it is not a blocker for testing. 

Ping me up when ISSUE 2 and this [Question](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1353607022) will be handled and I will perform re-test. Thank you!

pavloburykh:
ISSUE 5 List of missing accessibility ids that are required for e2e development


1. Community card element in the list of Discover community screen 
Appium 2022-12-19 16-48-02


2. Request join/Join Community button on community screen

Appium 2022-12-19 16-44-28

3. Checkbox on Community rules screen
4. Cancel and Request join buttons on Community rules screen

add Non Token Gated - Manual Approval Required flow by J-Son89 · Pull Request #14513 · status-im:status-mobile 2022-12-19 16-51-01

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -10,7 +10,7 @@
             [status-im.communities.core :as communities]
             [status-im.i18n.i18n :as i18n]
             [status-im.react-native.resources :as resources]
-            [status-im.ui.components.list.views :as list]
+            [status-im.utils.datetime :as datetime]
```
J-Son89: going to make a follow up p.r and address the rest of the imports.. didn't want to increase the size of this p.r

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -141,10 +158,20 @@
...
+                      (when-not joined
+                        (when can-join?
+                          (>evt [::communities/join id]))
+
+                        (when
+                          can-request-access?
+                          (and can-request-access?
+                               (zero? requested-to-join-at)
+                               (can-request-access-again? requested-to-join-at))
+                          (>evt [::communities/request-to-join id])))
```
J-Son89: same for events, will start moving community events to the community context

J-Son89:
@pavloburykh I addressed both issues and with regards to the question about the logic for showing the button and/or the notice that admin is offline.
I made it in this pr that once a request is made it should hide the `request to join button` so the user would never see both that button and the this admin is offline notice

I also added accessibility labels for the components you suggested. let me know if there is any issues with this.

pavloburykh:
> @pavloburykh I addressed both issues and with regards to the question about the logic for showing the button and/or the notice that admin is offline. I made it in this pr that once a request is made it should hide the `request to join button` so the user would never see both that button and the this admin is offline notice
> 
> I also added accessibility labels for the components you suggested. let me know if there is any issues with this.

Hi @J-Son89 Thanx for the fixes. 

Currently I am still reproducing [ISSUE 2](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1353591562). Regarding [QUESTION](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1354604488) seems like that behaviour remained the same as it was in previous build. 


https://user-images.githubusercontent.com/97245802/209000293-a0ef639b-6103-4b43-a460-a47f6557ef85.mp4

J-Son89:
@pavloburykh I updated this p.r with the changes needed. Should be working now 👍

pavloburykh:
> @pavloburykh I updated this p.r with the changes needed. Should be working now 👍

Hi @J-Son89 Thank you! 

Unfortunately behaviour is still the same for me. Both [ISSUE 2](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1353591562) and [QUESTION](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1354604488) remained. 

Could you please show a video how it works for you? Maybe we have some miscommunication on how it supposed to work. Thanx!

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -81,15 +88,26 @@
...
 (defn community-rules-list
   [rules]
-  [list/flat-list
+  [rn/flat-list
    {:shows-horizontal-scroll-indicator false
     :data                              rules
     :separator                         [rn/view {:margin-top 1}]
     :render-fn                         community-rule-item}])
 
+(def request-cooldown-ms (* 24 60 60 1000))
```
flexsurfer: this probably should be somewhere in constants or in setup

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -81,15 +88,26 @@
...
+
 (defn request-to-join
-  [community]
-  (let [agreed-to-rules? (reagent/atom false)]
+  [{:keys [permissions name id joined
+           can-join? can-request-access?
+           requested-to-join-at]}]
+  (let [agreed-to-rules? (reagent/atom false)
+        is-open?         (not= 3 (:access permissions))]
+    (js/console.log (not (pos? requested-to-join-at))
```
flexsurfer: leftover

src/status_im2/contexts/communities/requests/actions/view.cljs
```diff
@@ -139,12 +158,25 @@
...
         [button/button
-         {:on-press #(>evt [:bottom-sheet/hide])
-          :type     :grey
-          :style    {:flex 1 :margin-right 12}} (i18n/label :t/cancel)]
+         {:accessibility-label :cancel
+          :on-press            #(>evt [:bottom-sheet/hide])
+          :type                :grey
+          :style               {:flex         1
+                                :margin-right 12}} (i18n/label :t/cancel)]
         [button/button
```
flexsurfer: quo/button

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -56,7 +52,7 @@
                     :emoji                  emoji
                     :emoji-background-color channel-color
                     :on-enter-channel       (fn []
-                                              (utils/show-popup
+                                              (js/alert
```
flexsurfer, line 55: why ?

pavloburykh:
Had a call session with @J-Son89 
We figured out that request status is not updated to Pending right after request is sent. During the call I managed to pass a flow and get Pending and Joined statuses.

But during further testing I still faced a problem that is described in [ISSUE 2](https://github.com/status-im/status-mobile/pull/14513#issuecomment-1353591562). Status has not been updated for me neither after reopening community card nor after re-login. 

Will be waiting updates on that from @J-Son89 Thanx for your work and patience.

src/status_im2/contexts/communities/requests/actions/view.cljs
J-Son89: yeah it makes sense to have it somewhere broader in scope

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: that dependency for utils came from status_im folder and this is just mock data showing a basic case of this functionality.

J-Son89:
@pavloburykh I checked and the button not disapearring after sending a request is also happening for open communities, i.e "join Community" function. As such I think this issue was brought in from previous work. I was thinking to handle this in a follow up pr as it will be cleaner to get this work in and I have some more changes to similar parts of the codebase in other p.rs.

I can create an issue for it and get it addressed this week. WDYT?

pavloburykh:
> @pavloburykh I checked and the button not disapearring after sending a request is also happening for open communities, i.e "join Community" function. As such I think this issue was brought in from previous work. I was thinking to handle this in a follow up pr as it will be cleaner to get this work in and I have some more changes to similar parts of the codebase in other p.rs.
> 
> I can create an issue for it and get it addressed this week. WDYT?

Hey @J-Son89! SGTM. Let's merge this PR and fix issue with request in follow up PR. Thanx for your work!

J-Son89:
> > @pavloburykh I checked and the button not disapearring after sending a request is also happening for open communities, i.e "join Community" function. As such I think this issue was brought in from previous work. I was thinking to handle this in a follow up pr as it will be cleaner to get this work in and I have some more changes to similar parts of the codebase in other p.rs.
> > I can create an issue for it and get it addressed this week. WDYT?
> 
> Hey @J-Son89! SGTM. Let's merge this PR and fix issue with request in follow up PR. Thanx for your work!

Perfect, thanks Pavlo! I'll create the issue and post it here too 👍

Issue #14456, opened 2022-11-29
Communities join screens (Non token gated) - Add owner node offline information box


J-Son89:
We can't actually check when the owner is online, instead we will only show this "owner not online" box whenever there is a pending request. With the intention of optimising this approach at a later point.

Add notice to user that the user node is offline as per the designs.

https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7035%3A462899&t=ivg9IboexBv4dlSR-0

add integration tests to this.

vkjr:
@J-Son89, just to clarify, do I correctly understand that we show "Community owner node is offline" only in case when the community is free to join (no approve required) and our request is till pending)?

J-Son89:
@VladimrLitvinenko I think it's good to get this clarification in public and also helps alongside the context of the Github issue, but can you raise this question in Mobile-ui so that the design team can see it too.  They don't really use Github.

From what I understand we just want to show this message to a user at the same time their request is pending. 

So this can be in all 4 cases. i.e this can be in both token-gated and non-token gated communities and when manual approval is needed and not needed.

vkjr:
@J-Son89, I believe that in cases when a user request should be accepted by a community owner, this message would be both incorrect and redundant. Incorrect - because maybe community owner IS online, but simply don't want you in the community) Redundant - because there is "Pending" badge. Wdyt?

J-Son89:
> @J-Son89, I believe that in cases when a user request should be accepted by a community owner, this message would be both incorrect and redundant. Incorrect - because maybe community owner IS online, but simply don't want you in the community) Redundant - because there is "Pending" badge. Wdyt?

let's have this conversation on Status - mobile ui channel so design team is brought into the conversation :)

vkjr:
> @J-Son89,  
> let's have this conversation on Status - mobile ui channel so design team is brought into the conversation :)

Asked there)

du82:
this can be closed as the info box was added in #14515

vkjr:
Not yet, unfortunately)
The current implementation in #14515 adds the info box, but there is an unfinished discussion about when it should show up. 

In the Figma designs, info box shows up only for communities that are free to join and only in cases when the community owner node is offline. 
But since we don't really have information about when a community node is online/offline, the question was asked to the design team, and decided to wait for the decision of Join (product owner). Unfortunately, I don't know his github nickname to tag him here (@cammellos, maybe you know?)

[...]

PR #14515, opened 2022-12-09
Communities now correctly listed in pending/joined sections.

vkjr:
### Summary
- all communities are visible now, not only Status
- fixed tabs pending/joined/opened to correctly show communities
- added "node offline" information box from #14456
  
  
### Review notes
The behavior of added information box was not really defined (agreed to wait for John on the design meeting), so its adding doesn't really close an issue. Moreover, you won't see it in UI because currently, when you join an open community, it is treated as fully joined even if the user didn't have a network connection at the moment of joining. Not sure if this behavior is correct. @churik, maybe you know. If not, we can probably open a bug.

#### Platforms
- Android
- iOS

status: ready

### DIFF
--- a/src/status_im2/contexts/communities/home/view.cljs
+++ b/src/status_im2/contexts/communities/home/view.cljs
@@ -46,20 +46,25 @@
     :render-fn                         render-fn}])
 
 (defn segments-community-lists [selected-tab]
-  (let [communities (rf/sub [:communities/community-ids])
+  (let [ids-by-user-involvement (rf/sub [:communities/community-ids-by-user-involvement])
         tab @selected-tab]
     [rn/view {:style {:padding-left     20
                       :padding-right    8
                       :padding-vertical 12}}
      (case tab
        :joined
-       [communities-list communities]
+       [communities-list (:joined ids-by-user-involvement)]
 
        :pending
-       [communities-list communities]
+       [communities-list (:pending ids-by-user-involvement)]
 
        :opened
-       [communities-list communities])]))
+       [communities-list (:opened ids-by-user-involvement)]
+
+       [quo/information-box
+        {:type :error
+         :icon :i/info}
+        (i18n/label :t/error)])]))
 
 (defn home []
   (let [selected-tab (reagent/atom :joined)]
--- a/src/status_im2/contexts/communities/overview/style.cljs
+++ b/src/status_im2/contexts/communities/overview/style.cljs
@@ -1,17 +1,20 @@
 (ns status-im2.contexts.communities.overview.style
-  (:require [react-native.platform :as platform]
-            [quo2.foundations.colors :as colors]))
+  (:require [react-native.platform :as platform]))
 
 (def preview-user
   {:flex-direction :row
    :align-items    :center
    :margin-top     20})
 
-(def blur-channel-header {:blur-amount 32
-                          :blur-type :xlight
-                          :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
-                          :style {:position :absolute
-                                  :top (if platform/ios? 56 60)
-                                  :height 34
-                                  :width "100%"
-                                  :flex 1}})
+(def blur-channel-header
+  {:position :absolute
+   :top (if platform/ios? 56 60)
+   :height 34
+   :width "100%"
+   :flex 1})
+
+(def join-button
+  {:width "100%"
+   :margin-top 20
+   :margin-left :auto
+   :margin-right :auto})
--- a/src/status_im2/contexts/communities/overview/view.cljs
+++ b/src/status_im2/contexts/communities/overview/view.cljs
@@ -44,38 +44,38 @@
   #(rf/dispatch
     [:bottom-sheet/show-sheet
      {:content
-      (constantly [quo/token-gating
-                   {:channel {:name name
-                              :community-color (colors/custom-color :pink 50)
-                              :emoji emoji
-                              :emoji-background-color channel-color
-                              :on-enter-channel (fn [] (utils/show-popup "Entered channel" "Wuhuu!! You successfully entered the channel :)"))
-                              :gates {:read [{:token "KNC"
-                                              :token-img-src knc-token-img
-                                              :amount 200
-                                              :is-sufficient? true}
-                                             {:token "MANA"
-                                              :token-img-src mana-token-img
-                                              :amount 10
-                                              :is-sufficient? false
-                                              :is-purchasable true}
-                                             {:token "RARE"
-                                              :token-img-src rare-token-img
-                                              :amount 10
-                                              :is-sufficient? false}]
-                                      :write [{:token "KNC"
-                                               :token-img-src knc-token-img
-                                               :amount 200
-                                               :is-sufficient? true}
-                                              {:token "DAI"
-                                               :token-img-src dai-token-img
-                                               :amount 20
-                                               :is-purchasable true
-                                               :is-sufficient? false}
-                                              {:token "ETH"
-                                               :token-img-src eth-token-img
-                                               :amount 0.5
-                                               :is-sufficient? false}]}}}])
+      (fn [] [quo/token-gating
+              {:channel {:name name
+                         :community-color (colors/custom-color :pink 50)
+                         :emoji emoji
+                         :emoji-background-color channel-color
+                         :on-enter-channel (fn [] (utils/show-popup "Entered channel" "Wuhuu!! You successfully entered the channel :)"))
+                         :gates {:read [{:token "KNC"
+                                         :token-img-src knc-token-img
+                                         :amount 200
+                                         :is-sufficient? true}
+                                        {:token "MANA"
+                                         :token-img-src mana-token-img
+                                         :amount 10
+                                         :is-sufficient? false
+                                         :is-purchasable true}
+                                        {:token "RARE"
+                                         :token-img-src rare-token-img
+                                         :amount 10
+                                         :is-sufficient? false}]
+                                 :write [{:token "KNC"
+                                          :token-img-src knc-token-img
+                                          :amount 200
+                                          :is-sufficient? true}
+                                         {:token "DAI"
+                                          :token-img-src dai-token-img
+                                          :amount 20
+                                          :is-purchasable true
+                                          :is-sufficient? false}
+                                         {:token "ETH"
+                                          :token-img-src eth-token-img
+                                          :amount 0.5
+                                          :is-sufficient? false}]}}}])
       :content-height 210}]))
 
 (def mock-list-of-channels {:Welcome [{:name "welcome"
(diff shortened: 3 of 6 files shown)


J-Son89:
> ### Summary
> * all communities are visible now, not only Status
> * fixed tabs pending/joined/opened to correctly show communities
> * added "node offline" information box from [Communities join screens (Non token gated) - Add owner node offline information box #14456](https://github.com/status-im/status-mobile/issues/14456)
> 
> ### Review notes
> The behavior of added information box was not really defined (agreed to wait for John on the design meeting), so its adding doesn't really close any issue. Moreover, you won't see it in UI because currently, when you join an open community, it is treated as fully joined even if the user didn't have a network connection at the moment of joining. Not sure if this behavior is correct. @churik, maybe you know. If not, we can probably open a bug.
> 
> #### Platforms
> * Android
> * iOS
> 
> status: ready

can you add a fixes and the issue link please :)

vkjr:
@J-Son89, as I mentioned in review notes, this one doesn't really fix the issue, that it was originally expected to fix, so I'm not sure if I should add the "fixes" header, wdyt?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,8 +159,31 @@
 
 (def channel-list-component (memoize channel-list-component-fn))
 
+(defn join-community [{:keys [joined can-join? requested-to-join-at community-color] :as community}]
```
J-Son89, line 162: 👍

src/status_im2/subs/communities.cljs
```diff
@@ -105,6 +105,28 @@
  (fn [communities]
    (map :id communities)))
 
+
+;; Return communities splitted by level of user participation. Some communities user
```
J-Son89: Can this comment not go as a doc string in the function?
e.g
```
(re-frame/reg-sub
:communities/community-ids-by-user-involvement
 :<- [:communities/communities]
"doc string goes here"
...
```


also imo we only really need the last line
`Result map has form: {:joined [id1, id2] :pending [id3, id5] :opened [id4]}"`

vkjr: Unfortunately, docstring doesn't work here, because it is a function call, and looks like reg-sub doesn't support docstring (I tried).
Also, I wasn't able to remove newlines between the description and code - linter argues.

As for description, I would say that mentioning that opened communities remain in database can add a piece of information for the reader

J-Son89 (approved):

translations/en.json
```diff
@@ -1894,4 +1895,4 @@
     "remove-user-from-group": "Remove {{username}} from the group",
     "edit-name-image": "Edit name and image",
     "owner": "Owner"
-}
+}
```
flexsurfer: could you please rollback new line

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,8 +159,31 @@
...
+  (let [node-offline? (and can-join? (not joined) (pos? requested-to-join-at))]
+    [:<>
+     (when-not joined
+       [quo/button
+        {:on-press  #(rf/dispatch [:bottom-sheet/show-sheet
+                                   {:content (constantly [requests.actions/request-to-join community])
```
flexsurfer: why constantly? its hard to read here

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,8 +159,31 @@
...
+                                    :content-height 300}])
+         :override-background-color community-color
+         :style
+         {:width "100%"
```
flexsurfer: move to styles ns

src/status_im2/subs/communities.cljs
```diff
@@ -105,6 +105,28 @@
...
+(re-frame/reg-sub
+ :communities/community-ids-by-user-involvement
+ :<- [:communities/communities]
+ (fn [communities]
+   (reduce (fn [{:keys [joined pending open] :as acc} community]
+             (let [joined? (:joined community)
+                   requested? (pos? (:requested-to-join-at community))
+                   id (:id community)]
+               (cond
+                 joined?    (assoc acc :joined (conj joined id))
```
flexsurfer: update could be used

src/status_im2/contexts/communities/overview/view.cljs
vkjr: This part was extracted as-is from a bigger component)

flexsurfer: could you change it please?

vkjr: sure!

src/status_im2/contexts/communities/home/view.cljs
```diff
@@ -46,20 +46,20 @@
...
-       [communities-list communities]
+       [communities-list (:joined communities-ids)]
 
        :pending
-       [communities-list communities]
+       [communities-list (:pending communities-ids)]
 
        :opened
-       [communities-list communities])]))
+       [communities-list (:opened communities-ids)])]))
```
ilmotta: `case` without a default branch will throw an exception. If you're sure the tab is always within the valid set of branches, another alternative is to remove `case` and just use the following:

```clojure
[communities-list (get communities-ids tab)]
```

src/status_im2/contexts/communities/home/view.cljs
```diff
@@ -46,20 +46,20 @@
     :render-fn                         render-fn}])
 
 (defn segments-community-lists [selected-tab]
-  (let [communities (rf/sub [:communities/community-ids])
+  (let [communities-ids (rf/sub [:communities/community-ids-by-user-involvement])
```
ilmotta: Whenever I read `<something>-ids`, I think it's a collection (it's very conventional), but in reality it's a map indexed by the user "involvement". I would suggest renaming this binding to something like `ids-by-user-involvement`.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,8 +159,31 @@
...
+          :margin-top 20
+          :margin-left :auto
+          :margin-right :auto}
+         :before :i/communities}
+        (i18n/label :t/join-open-community)])
+     (when node-offline?
+       [quo/information-box
+        {:type :informative
+         :icon :main-icons/info
```
ilmotta: [Guidelines] Use the `:i` prefix instead of `:main-icons`.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -159,8 +159,31 @@
...
+         :style {:margin-top 12}}
+        (i18n/label :request-processed-after-node-online)])]))
```
ilmotta: [Guidelines] Use the `:t` qualification in translation keywords.

src/status_im2/subs/communities.cljs
```diff
@@ -105,6 +105,28 @@
...
 
+
+;; Return communities splitted by level of user participation. Some communities user
+;; already joined, to some of them join request sent and others were opened one day
+;; and their data remained in app-db.
+;; Result map has form: {:joined [id1, id2] :pending [id3, id5] :opened [id4]}"
+
+
+(re-frame/reg-sub
+ :communities/community-ids-by-user-involvement
```
ilmotta, line 109: How about tests for this new layer-3 subscription? It's begging for some! If it's a new layer-3 subscription such as this one, we should properly test it. Feel free to ping me if it's not clear how to test this. You can find examples in:

- subs/communities_test.cljs
- subs/wallet/wallet_test.cljs
- subs/activity_center_test.cljs

On a side note, I actually think soon we should re-evaluate our minimum requirements for test coverage in PRs.

src/status_im2/subs/communities.cljs
ilmotta: @vkjr, you're right about the docstring, we can't use it here, but could you join the comments with the `reg-sub` call? There are two empty lines between them.

src/status_im2/contexts/communities/overview/view.cljs
ilmotta: Humm, let's not generalize that `constantly` is hard to read. This is just personal preference. I find it quite clear in many contexts, and it fits like a glove in this piece of code. But of course I'm okay with switching to just `(fn [] [requests.actions/request-to-join community])`

src/status_im2/subs/communities.cljs
vkjr: @ilmotta, for some reason linter returns an error when I put the comments right before the code :-/

ilmotta: Such a weird recommendation from `cljfmt`! You're totally right, I guess if you prefer you can always add the comments at the top of the anonymous function, but not much of an improvement :man_shrugging: 

```clojure
(re-frame/reg-sub
 :communities/community-ids-by-user-involvement
 :<- [:communities/communities]
 ;; Return communities splitted by level of user participation. Some communities
 ;; user already joined, to some of them join request sent and others were opened
 ;; one day and their data remained in app-db. Result map has form:
 ;; {:joined [id1, id2] :pending [id3, id5] :opened [id4]}"
 (fn [communities]
   (reduce (fn [{:keys [joined pending open] :as acc} community]
             (let [joined? (:joined community)
                   requested? (pos? (:requested-to-join-at community))
                   id (:id community)]
               (cond
                 joined?    (assoc acc :joined (conj joined id))
                 requested? (assoc acc :pending (conj pending id))
                 :else      (assoc acc :opened (conj open id)))))
           {:joined [] :pending [] :opened []}
           communities)))
```

vkjr: Hmm, nice idea! For me, it really seems like the best option! I would go that way)

src/status_im2/contexts/communities/overview/view.cljs
flexsurfer: i mean we've never used `constantly` so im not used to it, so we need to use the same pattern for all such cases in the code, `constantly` is good in java i guess , to safely have a function with any number of arguments if you don't use them, in javascript we don't care, so `constantly` is useless

src/status_im2/contexts/communities/home/view.cljs
vkjr: I would prefer not to rely on using the same keywords for naming tabs and for naming vectors of communities, since they have different logical meanings and can change in the future. So I would add the :default section with simple error indication to visually catch this case. Wdyt?
```
:default
       [quo/information-box
        {:type :error
         :icon :i/info}
        (i18n/label :t/error)]
        ```

Comment (ilmotta) [reply src/status_im2/contexts/communities/home/view.cljs]: Makes total sense to me @vkjr. Thanks

**Review by ilmotta — APPROVED** [review]

File: src/status_im2/contexts/communities/home/view.cljs [hunk 12]
```diff
@@ -46,20 +46,26 @@
...
 
        :pending
-       [communities-list communities]
+       [communities-list (:pending ids-by-user-involvement)]
 
        :opened
-       [communities-list communities])]))
+       [communities-list (:opened ids-by-user-involvement)]
+
+       :default
```
ilmotta: Just a minor correction here @vkjr, the default branch in the `case` macro doesn't take an option, only the `cond` macro understands that.

```clojure
;; This explodes :boom: 
(let [tab :unknown]
  (case tab
    :joined  :joined
    :pending :pending
    :opened  :opened
    :default :error))

(let [tab :unknown]
  (cond 
    (= tab :joined)  :joined
    (= tab :pending) :pending
    (= tab :opened)  :opened
    :default         :error))
```

vkjr: 🤦 , thanks for noticing!

src/status_im2/contexts/communities/overview/style.cljs
```diff
@@ -7,11 +7,23 @@
...
+           :flex 1}})
+
+(def join-button
+  {:width "100%"
+   :margin-top 20
+   :margin-left :auto
+   :margin-right :auto})
+
+(def node-offline-information-box
+  {:type :informative
```
flexsurfer: we should have only styles in style ns, not properties

src/status_im2/contexts/communities/overview/style.cljs
```diff
@@ -7,11 +7,23 @@
...
-                          :blur-type :xlight
-                          :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
-                          :style {:position :absolute
-                                  :top (if platform/ios? 56 60)
-                                  :height 34
-                                  :width "100%"
-                                  :flex 1}})
\ No newline at end of file
+(def blur-channel-header
+  {:blur-amount 32
```
flexsurfer: we should have only styles in style ns, not properties

flexsurfer (approved):

PR #14520, opened 2022-12-12
feat: replace clj-fmt with zprint

yqrashawn:
There's an old PR about using zprint at https://github.com/status-im/status-mobile/pull/14128 
You can find example before/after diffs there
There are several upstream issues listed below, which are getting fixed by the zprint author, using commits in the above PR as test cases.
I created this pull request to integrate zprint and I suggest we merge this and get along with these issues until they are fixed.

The most outstanding issue is that the user needs to format a file 2/3 times to get the end formatted state. This only happens to files with specific code structure.
As a workaround, `make lint-fix` is already set to run `zprint` twice.

## To check what's the formatted code look like on your local machine
1. checkout this branch
2. `make shell TARGET=default`
3. run the commands below and check formatted file
```sh
clojure-lsp clean-ns
zprint '{:search-config? true}' -fw src/**/*.{clj,cljs,cljc}
```

# Summary
Replace clj-fmt with zprint, which does more format and output more consistent formatted code.
1. remove clj-fmt dependency
1. move clj-kondo, clojure-lsp, zprint to default nix shell
1.  use `default` nix shell instead of `clojure` for `make lint` and `make lint-fix`
1.  use zprint in lint and lint-fix, use `clojure-lsp clean-ns` in lint-fix to sort ns form
1.  `lint` and `lint-fix` only check staged files in git or `shadow-cljs.edn` which is `git diff --diff-filter=d --cached --name-only | grep -e \.clj$ -e \.cljs$ -e \.cljc$ -e \.edn$ || echo shadow-cljs.edn`

# Current Rule
1. based on [community rule](https://github.com/bbatsov/clojure-style-guide), [zprint doc about community style](https://github.com/kkinnear/zprint/blob/main/doc/options/community.md)
5. maximum width (column) **~80~ 105**
6. no optional comma in map
7. justify (vertically align) kvs in map
8. justify `:as`, `:refer`… in ns `:require` form
9. respect newlines, have some flexibility for formatting the code manually
10. respect blank lines
11. `clojure-lsp clean-ns` to sort `:require` and `:import` in `ns` form alphabetically (won't check in `lint`, only provided in `lint-fix`)

# Known issues
1. might require multiple times of format to get to the “end formatted state” (will report to upstream) https://github.com/kkinnear/zprint/issues/271
3. ~combine justify and respect new line in `:require` from cause syntax error (will report to upstream)~ no need to respect-nl in ns form
4. ~can't use helper fns like [`rodguide`](https://github.com/kkinnear/zprint/blob/2da480dda16a683846f57df0941f70a5ae1000d1/src/zprint/guide.cljc#L34) in `.zprintrc` when nix-shell has `TARGET=clojure`, works fine when `TARGET=default` (working on this one)~

# Editor/IDE support
- Already included in `make lint-fix`
- Has prebuilt binary for macOS and Linux, also can be used in form or uberjar or library in [REPL](https://github.com/kkinnear/zprint/blob/main/doc/using/repl.md) https://github.com/kkinnear/zprint#get-zprint
- zprint editors doc https://github.com/kkinnear/zprint/blob/main/doc/using/editor.md including Vim Emacs Sublime Text and Atom
- Make sure there's a `$HOME/.zprintrc` or `$HOME/.zprint.edn` file with `:search-config? true` or `:cwd-zprintrc?  true` or both in it. e.g.
```clojure
;; ~/.zprintrc
{:search-config? true}
```
- Note: zprint always read `$HOME/.zprintrc`, so rules there will affect the format result

## Emacs (tested)
Use package like https://github.com/radian-software/apheleia to config formatter for Clojure files

## VS Code (tested) 
zprint VS Code doc https://github.com/kkinnear/zprint/blob/main/doc/getting/vscode.md  
there's a zprint plugin https://github.com/rflagreca/vscode-clj-zprint
- install the plugin
- add this config in VS Code `settings.json`
```json
  "editor.formatOnSave": true,
  "[clojure]": {
    "editor.defaultFormatter": "rflagreca.vscode-clj-zprint"
  }
```

## IntelliJ IDEA (tested on macos)
Add `zprint` as external tool https://www.jetbrains.com/help/idea/configuring-third-party-tools.html
Program: path to zprint (example macos: `/opt/homebrew/bin/zprint` )
Arguments: `"{:search-config? true}" -fw $FilePathRelativeToProjectRoot$`
Working directory: `$ProjectFileDir$`
Check `Advanced Options -> Synchronize files after execution`
Check `Advanced Options -> Open console for tool output -> Make console active on message in stderr`

`Note: there is already zprint on macos, so better to install with a different name or use a homebrew path`

# Resources
1. zprint standalone binary (fast startup time) for [macOS](https://github.com/kkinnear/zprint/blob/main/doc/getting/macos.md) and [Linux](https://github.com/kkinnear/zprint/blob/main/doc/getting/linux.md)
1. [use zprint as library in REPL](https://github.com/kkinnear/zprint/blob/main/doc/using/repl.md)
1. use `zprint '{:search-config? true}' ...` to format file so that it can read the `.zprintrc` at cwd
1. add `:cwd-zprintrc? true` to your `$HOME/.zprintrc` if you have one so that it respect `.zprintrc` at cwd
1. [zprint doc](https://github.com/kkinnear/zprint/blob/main/doc/reference.md#comment)

 
status: ready


ilmotta:
@yqrashawn, I'll do something a little bit different to review this PR. I'll experiment doing code changes here and there and getting a feeling of how the formatter behaves. I'll report back once I've done enough. Thank you for moving forward with this PR, much appreciated by everyone.

ilmotta:
> justify :as, :refer… in ns :require form

@yqrashawn I found this rule to cause readability issues in a lot of namespaces. Sometimes, the aliases are padded way too much to the right, which causes lots of eye scanning from left to right (to skip the whitespace gaps).

It's also hard to find Clojure codebases following this rule, really hard. Other programming languages also avoid doing this, probably for good reasons.

IMO we should drop this rule and keep the formatting more conventional, without padding aliases to the right.

Another downside to this rule:

- It will cause more frequent diff in namespace sections because it depends on the width of the longest namespace.

Examples, but there are many others.

- `src/status_im/ui/screens/wallet_connect/session_proposal/views.cljs`
- `src/status_im/ui/screens/screens.cljs`
- `src/status_im/ui/screens/add_new/new_chat/views.cljs`
- `src/status_im/ui/screens/bootnodes_settings/edit_bootnode/views.cljs`
- `src/status_im/ui/screens/browser/bookmarks/views.cljs`
- `src/status_im/ui/screens/browser/options/views.cljs`
- `src/status_im/ui/screens/bottom_sheets/views.cljs`
- `src/status_im/ui/screens/activity_center/notification/contact_verification/view.cljs`

.zprintrc
```diff
@@ -0,0 +1,50 @@
...
+             :respect-nl
+
+             ;; respect blank line
+             ;; https://github.com/kkinnear/zprint/blob/main/doc/reference.md#respect-bl
+             :respect-bl]
+ :fn-map
+ {"reg-sub"         :arg1-pair
+  "list-comp"       :binding
+  "defview"         :arg1-body
+  "letsubs"         :binding
```
ilmotta, line 24: Could you add `"deftest-sub" :arg1-body` as well?

ilmotta (approved):
Excellent work @yqrashawn!

I tested the output of many namespaces, and the formatter output looked fairly consistent and aligned with what most contributors want.

The only thing I'm not seeing much benefit for readability (actually hurting it) is right-padding namespace aliases (please, see https://github.com/status-im/status-mobile/pull/14520#issuecomment-1346981389).

`zprint` compiled as native is also super snappy and I would recommend it instead of the jar version for anyone integrating zprint with their IDE/editor of choice.

ilmotta:
@yqrashawn, team, I think after this PR is merged we should reformat the entire codebase ASAP, otherwise everyone will have lots of difficult rebases to perform and subsequent PRs will be much bigger to review because the formatter is going to check every changed file from now on.

Instead of paying the price of the new format style for a long time and dozens and dozens of PRs, I think we should pay the price once, that is, in order:

1. Merge this PR.
2. Create another PR formatting the entire codebase, block new PRs from being merged meanwhile, and merge it as quickly as possible. This step shouldn't take more than 1h, for instance.
3. Devs rebase once and profit from then on.

For this strategy to work, everyone approving this PR should understand and accept how the code will look like, so that the next PR that actually formats everything doesn't need actual human review, so it's merged quickly.

cc @flexsurfer @cammellos @Parveshdhull

yqrashawn:
justify ns from is just a nice to have, I've removed it since it's causing readability issue

.zprintrc
yqrashawn: done

smohamedjavid (approved):

ibrkhalil (approved):

J-Son89 (approved):
working for me on mac :)

jo-mut (approved):

flexsurfer (approved):

siddarthkay (approved):

PR #14533, opened 2022-12-14
Remove old Notification Center

### LINKED ISSUE #14490: Replace Notification Center with new Activity Center
# Feature Issue

Make the new Activity Center the default, by removing the flag `setup/config/new-activity-center-enabled?` and by removing the old Notification Center code.

### Notes

- Keep in mind not all features from the old Notification Center have been or will be implemented in the Activity Center (AC). But the basic *contact request* flows do work.
- It's important to sync with the QA team, because removing the old Notification Center could break their current workflows.

### LINKED ISSUE #14495: Pending contact requests are not displayed when the new Activity Center is enabled
# Bug Report

## Problem 

When the flag `status-im2.setup.config/new-activity-center-enabled?` is enabled, the Contacts section doesn't display the *Pending requests* list item.

#### Expected behavior

Display the *Pending requests* list item when there are 1+ pending contact requests. The counter should be correct as well.

### Additional Information

- Status version: nightly 

ilmotta:
fixes #14495
fixes #14490

### Summary

In this PR:

- [x] Remove all traces of the old Notification Center.
- [x] Remove feature flag for the Activity Center.
- [x] Fix how pending contact requests are fetched given the new app db structure for the Activity Center. This fix allows the `Messages Home > Contacts tab > Pending requests` section to correctly open the bottom sheet to manage pending requests.

In an upcoming PR I'll try to move the Activity Center to the new app structure and guidelines.

- 80% of the green (+) changes come from two test files that had to be updated due to the removal of the feature flag.
- I'm opening the PR with lots of commits to help me rebase+squash before merging.

### Testing notes

Please, ignore the actual management of pending contact requests in the bottom sheet (see in [Figma](https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=2010%3A507764&t=2wtxz8fk9PxzDpNs-0)). This is out of scope for this PR, and this feature is currently buggy. Use the Activity Center instead. Future PRs will address this problem.

- Android
- iOS

### Things to look for during testing

- Go over a couple contact request flows, either decline or accept them. Observe how the `Messages home screen > Contacts tab` is updated.
- Check if the unread notification indicator is up-to-date during all tests.

status: ready


src/status_im2/contexts/chat/home/view.cljs
```diff
@@ -55,27 +56,29 @@
...
-                                      :label (i18n/label :t/groups)}
-                                     {:id                :contacts
-                                      :label             (i18n/label :t/contacts)
-                                      :notification-dot? (pos? (count contact-requests))}]}]
-         (if (= @selected-tab :contacts)
-           [contacts contact-requests]
-           [chats @selected-tab])]))))
+    [:f>
+     (fn []
+       (react/effect! #(rf/dispatch [:activity-center.notifications/fetch-unread-contact-requests]))
```
flexsurfer: could you please elaborate on why it's here as an effect? i guess we have a signal for new ? and we need to only once fetch after sign in then better to move this to event after sign in

J-Son89:
perhaps easier in a follow up p.r.. but can we not move all (or most) of these files to status_im2?

src/status_im2/contexts/chat/home/view.cljs
```diff
@@ -4,6 +4,7 @@
             [i18n.i18n :as i18n]
             [react-native.core :as rn]
             [quo2.core :as quo]
+            [quo.react :as react]
```
J-Son89: why are are serving react from quo.react.. can we not move this to `src/react` ? 🤔

J-Son89: again perhaps beyond the scope of this p.r, but it could be a follow up.

flexsurfer (commented):
honestly, I'm not sure what's the point of removing and modifying code in the old codebase.

src/status_im2/contexts/chat/home/view.cljs
J-Son89: > honestly, I'm not sure what's the point of removing and modifying code in the old codebase.

removing is a good thing because it will prevent devs from just modifying the old code. I agree about the modifying, we have to keep the migration to status_im2 a priority. imo if a follow up p.r is easier for situations like this then that's also fine

ilmotta: > honestly, I'm not sure what's the point of removing and modifying code in the old codebase.

I'm removing because there's no point in having dead code in the codebase, we have git for that. It adds burden to maintenance for no added benefit and confuses the heck of all contributors. At least that's my take on the problem @flexsurfer.

src/status_im2/contexts/chat/home/view.cljs
ilmotta: I think so @J-Son89, the react wrapper functions found in `quo.react` are quite useful, but they are in the wrong place. Of course it's beyond the scope of this PR, but it'd be nice to extract it out.

src/status_im2/contexts/chat/home/view.cljs
ilmotta: > could you please elaborate on why it's here as an effect? i guess we have a signal for new ? and we need to only once fetch after sign in then better to move this to event after sign in

I'll give it another try @flexsurfer, that was my first approach, but perhaps something else wasn't working and it confused me. I'll get back to this thread soon to let you guys know.

ilmotta:
> perhaps easier in a follow up p.r.. but can we not move all (or most) of these files to status_im2?

@J-Son89 This was lower priority to me, now that the old Notification Center is out of the way, I'm gonna move the code to the new structure as you say o/

Notice that removing the old Notification Center wasn't just about deleting code, I had to thoroughly review the old event handlers and adapt them (some were still needed).

src/status_im2/contexts/chat/home/view.cljs
ilmotta: I'm resolving this thread because I changed the code to use signals to keep the app db synchronized with pending contact requests.

src/status_im2/contexts/chat/home/view.cljs
ilmotta: I changed the code to use signals to keep the app db synchronized with pending contact requests. I have no clue why I implemented https://github.com/status-im/status-mobile/pull/14533/commits/e438f3a0bad677e1e5bae24aa3cb79d85f5c3eda in the first place.

Thanks for spotting the issue.

J-Son89 (approved):
looks good to me. Please check with Andrey before merging because I believe he might be making refactors in a similar part of the codebase so will good to work out the best approach for both of you if so :)

flexsurfer (approved):

PR #14563, opened 2022-12-16
remove comment blocks which are not useful

### LINKED ISSUE #14524: tidy up comments at modules/react-native-status/ios/RCTStatus/RCTStatus.m
# Problem

Get rid of them forward slashes at RCTStatus.m because they provide no valuable information and gives an indication to new contributors that they must also follow the same commenting pattern.

siddarthkay:
Adding a series of // characters (i.e., a "comment bar") above every method in an Objective-C file would not provide any real benefit, and would in fact be redundant, as the method list in Xcode already serves as a visual separator for methods in a class. The #pragma mark directive is a more concise and effective way to add separators and group related methods in a class. If required we should use #pragma mark to create a separator line and add a heading to describe the methods in a particular section.

fixes #14524

status: ready

### DIFF
--- a/modules/react-native-status/ios/RCTStatus/RCTStatus.m
+++ b/modules/react-native-status/ios/RCTStatus/RCTStatus.m
@@ -90,26 +90,25 @@ static RCTBridge *bridge;
 
 RCT_EXPORT_MODULE();
 
-////////////////////////////////////////////////////////////////////
 #pragma mark - shouldMoveToInternalStorage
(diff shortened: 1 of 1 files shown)


J-Son89:
🤩

J-Son89 (approved):

siddarthkay:
@pavloburykh @churik @qoqobolo : This PR basically just removes comments form an objective-C file, Can I go ahead and merge or would you need to do a smoke test here as well?

churik:
go ahead, thanks for tagging

PR #14575, opened 2022-12-19
[#14545] Force update wakuv2 default configs

### LINKED ISSUE #14545: Set waku-2 fleet by default on existing users
For existing users, we should override the waku-2 settings with the new values (the same as we have in nodes), ideally whatever the clients passes, overrides whatever is in the status-go db.

rasom:
fix #14545

status: ready

### DIFF
--- a/src/status_im/node/core.cljs
+++ b/src/status_im/node/core.cljs
@@ -64,10 +64,20 @@
     (assoc :ListenAddr ":30304"
            :DataDir    (str (:DataDir config) "_dev"))))
 
+(def wakuv2-default-config
+  {:DiscoveryLimit 20
+   :EnableDiscV5   true
+   :Host           "0.0.0.0"
+   :AutoUpdate     true
+   :PeerExchange   true
+   :Port           0
+   :UDPPort        0})
+
 (def login-node-config
   {:WalletConfig (cond-> {:Enabled true}
                    (not= config/opensea-api-key "")
-                   (assoc :OpenseaAPIKey config/opensea-api-key))})
+                   (assoc :OpenseaAPIKey config/opensea-api-key))
+   :WakuV2Config wakuv2-default-config})
 
 (defn- get-login-node-config
   [config]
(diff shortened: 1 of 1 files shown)


rasom:
@cammellos so currently it is only about `:WakuV2Config`. Let me know if something else should be added.

src/status_im/node/core.cljs
```diff
@@ -60,10 +60,20 @@
     (assoc :ListenAddr ":30304"
            :DataDir (str (:DataDir config) "_dev"))))
 
+(def wakuv2-default-config
```
J-Son89, line 67: @flexsurfer is there a designated place for this code to live outside of `status_im` folder?
can we migrate this code as part of this work?

flexsurfer: no we don't, it's better to have a separate PR

src/status_im/node/core.cljs
```diff
@@ -142,11 +152,8 @@
               :BloomFilterMode waku-bloom-filter-mode
               :LightClient     true
               :MinimumPoW      0.000001}
-             :WakuV2Config (assoc wakuv2-config
-                                  :DiscoveryLimit 20
-                                  :EnableDiscV5 true
-                                  :Enabled wakuv2-enabled
-                                  :Host "0.0.0.0")
+             :WakuV2Config (merge (assoc wakuv2-config :Enabled wakuv2-enabled)
```
ilmotta, line 165: It's very weird to read the default map argument coming second. It reads as if it's incorrect according to how merge works. Don't you mean the inverse?

```clojure
(merge wakuv2-default-config
       (assoc wakuv2-config :Enabled wakuv2-enabled))

;; Or even:
(merge wakuv2-default-config
       wakuv2-config
       {:Enabled wakuv2-enabled})
```

rasom: We don't want users to change them at the moment and we want to be sure those defaults are applied  `¯\_(ツ)_/¯`

ilmotta: All right then @rasom, good to go `¯\_(ツ)_/¯`

ilmotta (approved):

qfrank (approved):

PR #14576, opened 2022-12-19
New make target to run tests via REPL

ilmotta:
### Summary

This PR streamlines the experience of running ClojureScript unit tests via a REPL.

#### Motivation

1. Automate a chore I often have to do many times per week, i.e. run two commands manually, while waiting on the first one to complete.

2. Make testing easier and a more enjoyable experience for other contributors.

Developers new to the concept of Lisp REPLs (as in Clojure) often rely on slow file system watchers to re-run changed tests. This works, but gives a very slow feedback cycle when compared to the *instant* feedback of a REPL.

In the Status codebase, we have `make test-watch`, but it re-runs the entire test suite, which makes it almost intolerable during development, not to count the waste of machine resources on every file save. Besides, this approach will get even worse as the test suite grows.

Developers can also manually change the shadow-cljs option `:ns-regex` to control which namespaces the test runner should pick. This helps, but the developer still needs to wait for recompilation of at least a single file (tolerable, but still slow), and manually change the shadow-cljs config.

### Usage

1. `make test-watch-for-repl`.
3. Once you see the message `[repl] shadow-cljs - #3 ready!` you can connect a REPL to the `:test` target from VS Code, Emacs, etc.
4. In any test namespace, run [cljs.test/run-tests](https://cljs.github.io/api/cljs.test/#run-tests) or your preferred method to run tests in the current namespace.

### Example in Emacs

In the video below, you can see two buffers side-by-side. On the left the source implementation, on the right the REPL buffer. Whenever I press a keybinding, **tests in the current namespace instantly run**. You can achieve this exact flow in VS Code, IntelliJ, Vim, etc.

[2022-12-19 12-46.webm](https://user-images.githubusercontent.com/46027/208465927-4ad9a935-5494-45e7-85b0-8134dc32d1a1.webm)

### Example in terminal emulator

Here I'm showing a terminal-only experience using Tmux (left pane Emacs, right pane the output coming from running the make target).

[2022-12-19 13-17.webm](https://user-images.githubusercontent.com/46027/208471199-1909c446-c82d-42a0-9350-0c15ca562713.webm)

#### How does the target work?

The test REPL must run only once the `./target/test/test.js` file is created by the current execution of the shadow-cljs watcher. That's why we need `until` to periodically check the existence of this file.

The make target also uses the [concurrently](https://github.com/open-cli-tools/concurrently) NPM package, which is used to better interleave output from two concurrent processes, as well as making sure Ctrl-C (SIGINT) kills all of them.

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -305,6 +305,14 @@ test-watch: ##@ Watch tests and re-run no changes to cljs files
 	yarn install
 	nodemon --exec 'yarn shadow-cljs compile mocks && yarn shadow-cljs compile test && node --require ./test-resources/override.js target/test/test.js' -e cljs
 
+test-watch-for-repl: export TARGET := clojure
+test-watch-for-repl: ##@ Watch tests and support REPL connections
+	yarn install
+	rm -f target/test/test.js
+	concurrently --kill-others --prefix-colors 'auto' --names 'build,repl' \
+		'yarn shadow-cljs compile mocks && yarn shadow-cljs watch test --verbose' \
+		'until [ -f ./target/test/test.js ] ; do sleep 1 ; done ; node --require ./test-resources/override.js ./target/test/test.js --repl'
+
 test: export TARGET := clojure
 test: ##@test Run tests once in NodeJS
 	# Here we create the gyp bindings for nodejs
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "@testing-library/jest-native": "^5.0.0",
     "@testing-library/react-native": "^11.2.0",
     "@types/jest": "^28.1.6",
+    "concurrently": "^7.6.0",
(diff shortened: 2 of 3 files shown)


ibrkhalil (approved):
Tested it and it's EPIC
Thank you so much Icaro!

rasom (approved):

vkjr (approved):

erikseppanen (approved):

jo-mut (approved):

flexsurfer (approved):

qfrank (approved):

briansztamfater (approved):

yqrashawn (approved):

J-Son89 (approved):

PR #14594, opened 2022-12-21
fix: fx merge error on ::json-rpc/call

yqrashawn:
status: ready

### DIFF
--- a/src/status_im/utils/fx.cljs
+++ b/src/status_im/utils/fx.cljs
@@ -2,6 +2,7 @@
   (:require-macros status-im.utils.fx)
   (:require [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.handlers :as handlers]
+            [status-im2.common.json-rpc.events :as json-rpc-status-im2]
             [taoensso.timbre :as log])
   (:refer-clojure :exclude [merge reduce]))
 
@@ -21,7 +22,8 @@
     :transport/confirm-messages-processed
     :group-chats/extract-membership-signature
     :utils/dispatch-later
-    ::json-rpc/call})
+    ::json-rpc/call
+    ::json-rpc-status-im2/call})
 
 (defn- safe-merge
   [fx new-fx]
--- a/src/status_im2/setup/dev.cljs
+++ b/src/status_im2/setup/dev.cljs
@@ -2,6 +2,7 @@
   (:require ["react-native" :refer (DevSettings LogBox)]
             [react-native.platform :as platform]
             [status-im.ethereum.json-rpc :as json-rpc]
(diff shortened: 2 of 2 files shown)


flexsurfer:
this wont fix the issue, i'll looking into this

flexsurfer:
oh you added to the old one, yeah this probably will solve it temporary

yqrashawn:
Thanks @flexsurfer , I'll close this then

yqrashawn:
I can just revert related change if you want @flexsurfer

flexsurfer:
could you add it here as well `status-im2.setup.dev  (rf/set-mergeable-keys`

yqrashawn:
@flexsurfer done

flexsurfer (approved):

src/status_im/utils/fx.cljs
```diff
@@ -2,6 +2,7 @@
   (:require-macros status-im.utils.fx)
   (:require [status-im.ethereum.json-rpc :as json-rpc]
             [status-im.utils.handlers :as handlers]
+            [status-im2.common.json-rpc.events :as json-rpc-status-im2]
```
ilmotta, line 5: @yqrashawn, @flexsurfer why is this namespace duplicated from the original one? Is there any behavioral change in the implementation? It doesn't look like it to me. If not, it should've been moved (not duplicated) to the new namespace. It's very easy and automatic with global grep/replace tools, to move namespaces/functions to different directories/files, this is very low effort.

Duplicating code that's **identical in behavior** doesn't make sense to me, particularly effects, events, subscriptions and utility functions. For the UI layer (components) I agree with the approach of duplicating because that's another thing, the new UI is most definitely different than the old one.

I would say we should create a new PR and remove this unnecessary duplication, i.e properly move the effect to a new namespace. I can do it for us real quick.

To summarize my thoughts:

1. Don't duplicate effects, event handlers, interceptors, utility functions, data store functions, etc **IF** the behavior is the same, simply *move* the entire code to the new directory/file. Use grep or some other tool to globally replace references. Remember we don't need to maintain compatibility with the old app. The old app is long gone, and developers should checkout old commits to use the old app.
2. We're needlessly wasting machine resources because re-frame global registers have duplicated effects, subscriptions, etc. This would be a problem if we had real production users, a no no for sure.
3. (We all agree about this third point) Prefer to duplicate UI code for the new mobile app because even if they look the same, they will most likely be different and evolve differently.

flexsurfer: this is a quick fix, to unblock develop, we should do it properly, we should change namespaced keyword, and remove old implementation

ilmotta: Thanks @flexsurfer, it seems I was too worried about this.

PR #14604, opened 2022-12-21
migrate datetime to utils and finish i18n migration

J-Son89:
Working on my latest pr, I realise we are still using datetime from status-im.
I moved this to `src/common` where we can put common pieces, i18n, react-native, datetime etc rather than having a large flatlist at `src`.

I then updated some dependencies that were point to status_im for i18n and datetime so they're pointing to the correct place now. 
I did not do this for any code in status_im as we will be removing that soon relatively soon.


src/common/datetime.cljs
```diff
@@ -0,0 +1,258 @@
+(ns common.datetime
```
J-Son89: imo we should some for of "common" folder at the root for utilities like date time and i18n etc.

flexsurfer:
imo `datetime` should be under `utils` but i18n should stay in the root, because its not utils or common, its an external library like react-native, so they should stay root, i don't think we'll have more so this root list won't grow

src/common/datetime.cljs
```diff
@@ -0,0 +1,258 @@
...
+
+(defn timestamp-sec
+  []
+  (int (/ (timestamp) 1000)))
+
+(defn timestamp->year-month-day-date
+  [ms]
+  (t.format/unparse (:year-month-day t.format/formatters) (to-date ms)))
+
+(re-frame/reg-cofx
```
flexsurfer: this shouldn't be here for sure , somewhere in setup events

flexsurfer:
`I did not do this for any code in status_im as we will be removing that soon relatively soon.` we discussed this with @ilmotta today, and it seems like better to move and don't duplicate

J-Son89:
> imo `datetime` should be under `utils` but i18n should stay in the root, because its not utils or common, its an external library like react-native, so they should stay root, i don't think we'll have more so this root list won't grow

sorry, yeah you're right. I'll move it to utils instead. :)

flexsurfer:
> I moved this to `src/common` where we can put common pieces, i18n, react-native, datetime etc rather than having a large flatlist at `src`.

we could use `vendor` similar to go or `libs` but honestly i don't think list will be big, most will be under utils

src/common/datetime.cljs
J-Son89: cool, I can move that too

J-Son89:
> `I did not do this for any code in status_im as we will be removing that soon relatively soon.` we discussed this with @ilmotta today, and it seems like better to move and don't duplicate

ok so, I can update this p.r for that too. Or perhaps I will do them separate it as it is quite large. I'll see how it looks

ilmotta (commented):
Thanks for putting in the effort to move the datetime namespace @J-Son89, but I have some complaints and to me this PR should not be merged. I'm sorry if my message seems a bit strong, but this is a very important subject IMO.

1. The `datetime` namespace has not been moved, it has been duplicated.
2. The datetime tests weren't copied along with the source, but this shouldn't be a problem if we fix 1).
3. The guidelines are a bit outdated in relation to utility namespaces, but we've been moving everything to `src/utils/` up to this point, not `src/common/`. Why the change?

Just today I was talking to @flexsurfer about our strategy to code the new mobile app and at least both of us agreed on how things should be done from now on. I wonder what the rest of the team thinks about it.

So in essence, we should really **stop duplicating code**, especially identical code. We apparently gain velocity by doing this, but it's an illusion, it accretes tech debt and it confuses all contributors (including external ones).

Duplicating code was as a reasonable strategy in the beginning, but since we decommissioned the old app, it doesn't make much sense anymore. You see, what do we gain by maintaining two datetime namespaces? `git checkout <rev>` for anyone that wants to build/release the old app ;)

We have tools to automate the work of renaming vars/namespaces globally and to automate file renames. Moving (renaming) things is cheap and low effort.

I'm cc'ing @flexsurfer to help give visibility to this matter.

ilmotta:
> > `I did not do this for any code in status_im as we will be removing that soon relatively soon.` we discussed this with @ilmotta today, and it seems like better to move and don't duplicate
> 
> ok so, I can update this p.r for that too. Or perhaps I will do them separate it as it is quite large. I'll see how it looks

I would be okay if you update this PR @J-Son89, because I haven't fully reviewed it.

J-Son89:
> Thanks for putting in the effort to move the datetime namespace @J-Son89, but I have some complaints and to me this PR should not be merged. I'm sorry if my message seems a bit strong, but this is a very important subject IMO.
> 
> 1. The `datetime` namespace has not been moved, it has been duplicated.
> 2. The datetime tests weren't copied along with the source, but this shouldn't be a problem if we fix 1).
> 3. The guidelines are a bit outdated in relation to utility namespaces, but we've been moving everything to `src/utils/` up to this point, not `src/common/`. Why the change?
> 
> Just today I was talking to @flexsurfer about our strategy to code the new mobile app and at least both of us agreed on how things should be done from now on. I wonder what the rest of the team thinks about it.
> 
> So in essence, we should really **stop duplicating code**, especially identical code. We apparently gain velocity by doing this, but it's an illusion, it accretes tech debt and it confuses all contributors (including external ones).
> 
> Duplicating code was as a reasonable strategy in the beginning, but since we decommissioned the old app, it doesn't make much sense anymore. You see, what do we gain by maintaining two datetime namespaces? `git checkout <rev>` for anyone that wants to build/release the old app ;)
> 
> We have tools to automate the work of renaming vars/namespaces globally and to automate file renames. Moving (renaming) things is cheap and low effort.
> 
> I'm cc'ing @flexsurfer to help give visibility to this matter.

ignore the "common" folder idea, I was not aware of utils  😓  and only saw react native and thought we were starting to put things in there. `utils` is good 👍 

going to delete the old version as part of this p.r too and so I'll update that with the tests.

src/common/datetime.cljs
J-Son89: please ignore 😓  🙃

J-Son89:
@ilmotta @flexsurfer moved the files now and updated across all files. This is just for date time. 

If the p.r is too large I can separate them by the commits as one if for datetime and the second for i18n

src/common/datetime.cljs
flexsurfer: was it moved ?

src/i18n/i18n_resources.cljs
```diff
@@ -1,11 +1,7 @@
-(ns status-im.i18n.i18n-resources
+(ns i18n.i18n-resources
```
flexsurfer: it was already moved `status-im2.setup.i18n-resources`

src/i18n/i18n_test.cljs
```diff
@@ -1,14 +1,10 @@
-(ns status-im.i18n.i18n-test
+(ns i18n.i18n-test
```
flexsurfer: status-im2.setup.i18n-test

src/i18n/i18n_resources.cljs
flexsurfer: so we need just to remove this, and use new ns in the status-im

ilmotta (approved):
It's great to see old things going away, thanks @J-Son89

src/i18n/i18n_resources.cljs
J-Son89: cool, will adjust 👍

src/common/datetime.cljs
J-Son89: I moved it now to status-im2.setup.datetime

src/i18n/i18n_test.cljs
J-Son89: now using status-im2.setup.i18n-test and status-im2.setup.i18n-resources 👍

flexsurfer (approved):
thank you!:)

src/status_im2/setup/datetime.cljs
```diff
@@ -0,0 +1,9 @@
+(ns status-im2.setup.datetime
```
J-Son89, line 1: @flexsurfer should I also `require` this namespace into a file like src/status_im2/setup/core.cljs? so that it gets registered?

J-Son89: I also need to include it somewhere for the tests too.. any ideas where is best? 🤔

src/status_im/test_runner.cljs
```diff
@@ -4,7 +4,11 @@
             [clojure.string :as string]
             [shadow.test :as st]
             [shadow.test.env :as env]
-            [utils.re-frame :as rf]))
+            [utils.re-frame :as rf]
+            [i18n.i18n :as i18n]
+            [status-im2.setup.i18n-resources :as i18n-resources]))
+
+(i18n/init (name i18n-resources/default-device-language) (clj->js i18n-resources/translations-by-locale))
```
J-Son89: I had to add this here as this used to be called in status-im/i18n/i18n.cljs.. it was not getting called for tests so I included it here 👍

J-Son89:
@churik, @pavloburykh this pr is just migrating files in the codebase. Does it need QA or it's okay to merge?

churik:
App hangs on launching, so not ready to be merged @J-Son89

J-Son89:
@churik can you take a look again? It should be working now :)

J-Son89:
@churik or perhaps I will fix tests first before you check!

churik:
@J-Son89 

**Devices:**
- IPhone 12 Mini (IOS 15)

Good to go now, thank you!

churik:
We have issue in tests (may be in app) + issue in reporting (as the results added https://github.com/status-im/status-mobile/pull/14604#issuecomment-1367435231 and https://ci.status.im/job/status-mobile/job/e2e/job/status-app-prs/2971/console are different)
Based on https://github.com/status-im/status-mobile/pull/14656#issuecomment-1367377878 I can definitely say it is not your PR.



As the results are the same https://github.com/status-im/status-mobile/pull/14656#issuecomment-1367377878 it is definitely not your PR.

PR #14629, opened 2022-12-23
chore: use banner from quo library

### LINKED ISSUE #14528: [quo2 new design tech debt] refactor pinned-banner to banner component in quo2
status-im.ui2.screens.chat.pinned-banner.view should be refactor to quo2 banner component

https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=863%3A26323&t=JeFB0V7ClQQKQsuA-3

J-Son89:
fixes https://github.com/status-im/status-mobile/issues/14528

Use quo2 banner component.

add unit test to quo2 banner component. 

adjusted icon in dark mode as icon color was not changing.

To test navigate to a chat screen with a pinned message and ensure that pinned messages banner is showing correctly and that the counter only shows when there is more than 1 pinned message.


src/quo2/components/banners/banner/view.cljs
```diff
@@ -3,22 +3,29 @@
...
\ No newline at end of file
+  (when (pos? pins-count)
+    [rn/touchable-opacity
+     {:accessibility-label :pinned-banner
+      :style               style/container
+      :active-opacity      1
+      :on-press            on-press}
+     (when show-pin?
+       [icons/icon :i/pin
+        {:color (colors/theme-colors
```
ilmotta: Looks like unnecessary line breaks.

src/quo2/components/banners/banner/view.cljs
```diff
@@ -3,22 +3,29 @@
...
+                 colors/neutral-100
+                 colors/white)
+         :size  20}])
+     [text/text
+      {:number-of-lines 1
+       :size            :paragraph-2
+       :style           {:margin-left 10 :margin-right 50}}
```
ilmotta: The style can be extract to `style.cljs`.

src/quo2/components/banners/banner/view.cljs
```diff
@@ -3,22 +3,29 @@
...
+      latest-pin-text]
+     [rn/view
+      {:accessibility-label :pins-count
+       :style               style/counter}
+      (when (> pins-count 1) [counter/counter {:type :secondary} pins-count])]]))
```
ilmotta, line 30: You can use `(pos? pins-count)`.

src/status_im2/contexts/chat/messages/view.cljs
```diff
@@ -59,13 +58,19 @@
...
     [rn/keyboard-avoiding-view {:style {:flex 1}}
      [page-nav]
      ;; TODO (flexsurfer) this should be in-app notification component in quo2
      ;; https://github.com/status-im/status-mobile/issues/14527
      [pin-limit-popover/pin-limit-popover chat-id]
-     [pin.banner/banner chat-id]
+     [quo/banner
+      {:latest-pin-text latest-pin-text
+       :pins-count      pins-count
+      }]
```
ilmotta: You can merge the dangling brackets with the previous line.

src/quo2/components/banners/__tests__/banner_component_spec.cljs
```diff
@@ -15,3 +15,13 @@
                       (.toBeTruthy))
                   (-> (js/expect (rtl/screen.getByText "5"))
                       (.toBeTruthy))))
+
+(js/global.test "banner component fires an event when pressed"
```
ilmotta, line 19: Shouldn't the tests be defined inside `banners/banner/` directory? If you do it, then test descriptions like `banner component <bla>` will be unnecessary, because it would be clear the component tests are only related to the banner, and nothing else.

ilmotta (approved):

src/quo2/components/banners/banner/view.cljs
flexsurfer: pos? is more than 0, so its different, here is 1

src/status_im2/contexts/chat/messages/pin/banner/view.cljs
```diff
@@ -1,15 +0,0 @@
-(ns status-im2.contexts.chat.messages.pin.banner.view
```
flexsurfer, line 1: could you please keep this component and just replace `[pinned-banner/pinned-banner` on quo2 on here, because we still want all related subscriptions to be here

flexsurfer (commented):
could you also please add quo preview screen ?

src/quo2/components/banners/__tests__/banner_component_spec.cljs
J-Son89: yeah sure we can do that. 
I was thinking just at the top level and each component has it's own spec/test file. But I can implement this suggestion as it's clearer of what component the test relates to and will help avoid confusion 👍

src/quo2/components/banners/banner/view.cljs
J-Son89: hmm, does our auto formatter not handle stuff like this?

J-Son89: and if not can, is it possible to adjust so we don't have to think about issues like this.

src/quo2/components/banners/banner/view.cljs
J-Son89: sure 👍

src/quo2/components/banners/banner/view.cljs
J-Son89: yep it's more than 1

src/status_im2/contexts/chat/messages/view.cljs
J-Son89: imo issues like this should be created as what we want from our auto formatter, I should not have to think about how stuff like this is presented. make lint-fix should just do this. 

Do we think this is possible? 
cc: @yqrashawn

src/quo2/components/banners/banner/view.cljs
ilmotta: Precisely, sorry about that.

J-Son89: always good to double check :)

src/quo2/components/banners/banner/view.cljs
ilmotta: It handles @J-Son89, but we have configured it to explicitly respect new lines. From a previous PR opened by @yqrashawn, we have found that allowing zprint to handle all new lines caused too many questionable format changes.

ilmotta: https://github.com/kkinnear/zprint/blob/main/doc/reference.md#respect-nl

J-Son89: right but shouldn't make lint-fix adjust this correctly for me? or at least `make lint` can break and tell me that I need to adjust it.

J-Son89: ok, I see now. It's too complex to handle and we must do ourselves. Personally would prefer if we can completely automate formatting so it's not a focus of p.r reviews.

src/status_im2/contexts/chat/messages/pin/banner/view.cljs
J-Son89: yeah I can do this, it's for performance reasons?

src/status_im2/contexts/chat/messages/view.cljs
ilmotta: There's a balance @J-Son89 where the formatter causes more problems if it takes control of every aspect of the code style. We learned this based on a PR opened a long time ago by @yqrashawn.

I know where you're coming from, something like Prettier, but formatting Lisp is not that clear cut and zprint has limitations at the moment. Lisps have more room for interpretation to be formatted, because an s-expression can appear anywhere, which is not the case in other programming languages that have the concept of statements vs expressions. Please, see my other comment about respecting newlines for more info.

It would be great if you or anyone else that wants to change the format rules, first change `.zprintrc` in your local machine, fully develop using the new rules and double-check if you really like what you see. If you do than by all means suggest a change!

src/status_im2/contexts/chat/messages/pin/banner/view.cljs
flexsurfer: both performance and more clean code

src/status_im2/contexts/chat/messages/view.cljs
ilmotta: Also note that some of us are using zprint after every file save because it's really fast to format the current file, so there's no manual step to call `make lint-fix`. I would recommend you set this up if you haven't. I never use `make lint-fix`, first of all because it's painfully slow, but also because this would mean my editor is formatting using a different set of rules than what's specified in `.zprintrc`, which would be an error on my editor side.

So I use zprint just as I would use prettier in VS Code, no manual step to call `make lint-fix`.

ilmotta: And on your point, I would love if zprint could respect new lines, but never leave dangling parentheses/brackets/etc

src/quo2/components/banners/banner/view.cljs
ilmotta: It's tricky @J-Son89, even the Clojure Style Guide doesn't force certain styles. But just to be clear, I agree with you, we need to keep pushing to improve the format rules (up to a point though).

I think we need to focus in one formatting problem at a time because any rule change can have undesirable ripple effects.

I really suggest that if anyone wants to improve formatting, first change your `zprintrc` file and see if you like the formatted output, check many many files. That's because some configuration changes are good *in theory*, but when you actually reformat they become disaster.

----

Just for context, a couple years ago, there were many heated discussions about auto formatting Clojure a la Prettier.js. We need to understand Lisps have a long history and that s-expressions can be formatted in many different ways, often in very subjective ways. Expressions are not statements as we all know. For example, in Clojure you can have "variable" (binding) declarations almost anywhere, this is not the case in Javascript. The space for subjectivity in Lisps is much much higher. If we decide to format using a very strict set of rules we'll cause more trouble than value in Clojure (as @yqrashawn demonstrated in his past PR to introduce zprint), so we better do one little rule at a time. There's a reason Lisps don't usually force a specific style, because the code end up looking worse in many cases.

ilmotta: FYI: @jo-mut, we were talking about format rules just now. For context, see my previous comment in this very thread https://github.com/status-im/status-mobile/pull/14629#discussion_r1056382648

VolodLytvynenko:
hi @J-Son89  Could you please rebase the PR and resolve existing conflicts? Thanx

J-Son89:
> hi @J-Son89 Could you please rebase the PR and resolve existing conflicts? Thanx

sure, rebased and pushed now 👍

VolodLytvynenko:
> sure, rebased and pushed now 👍
Thank you!

J-Son89:
> could you also please add quo preview screen ?

I already did that in a previous p.r. This is just implementing within the application

src/status_im2/contexts/chat/messages/pin/banner/view.cljs
J-Son89: done

src/status_im2/contexts/chat/messages/view.cljs
J-Son89: cool, I'll accept that's it the best we can do for now and will involve some manual formatting. Thanks for the thorough explanation! will make sure my editor is setup correctly too

ibrkhalil (approved):

VolodLytvynenko:
Hi @J-Son89 Thank you for your work. Here is just one low-priority issue. All other looks great
 
**ISSUE 1: The count of pinned messages is shifted if short text message is shown.**
**Steps to reproduce:**
1. Go to 1-1
2. Send the message
3. Pin it
4. Send one more message with a short text
5. Pin the last message also

**Actual result:**
The count badge  of pinned messages is shifted to the left
image

**Expected result:**
The count badge  of pinned messages is always located in the right part of the banner
image

VolodLytvynenko:
Hi @J-Son89 Only one little question. In which cases the banners with 'pin' and 'without pin icon should be shown as it is disaplyed per design?
Curretly all of them are shown without 'pin' icon in the current PR
image

src/status_im2/contexts/chat/messages/view.cljs
yqrashawn: > imo issues like this should be created as what we want from our auto formatter, I should not have to think about how stuff like this is presented. make lint-fix should just do this.
> 
> Do we think this is possible? cc: @yqrashawn

I've opened a feature request issue for zprint, currently it doesn't support merge dangling parenthesis if when respect-nl is enabled.

ilmotta: Thank you @yqrashawn.

jo-mut: > Also note that some of us are using zprint after every file save because it's really fast to format the current file, so there's no manual step to call `make lint-fix`. I would recommend you set this up if you haven't. I never use `make lint-fix`, first of all because it's painfully slow, but also because this would mean my editor is formatting using a different set of rules than what's specified in `.zprintrc`, which would be an error on my editor side.
> 
> So I use zprint just as I would use prettier in VS Code, no manual step to call `make lint-fix`.

for some reason `make lint-fix` does not seem to have effect as nothing really changes before and after. Am thinking is there a step i may have missed to make the command work with zprint?

ilmotta: @jo-mut, could you share exactly the branch and file you're trying to format that's not working? I can try on my machine, just to confirm your suspicions.

src/quo2/components/banners/banner/__tests__/banner_component_spec.cljs
```diff
@@ -15,3 +15,13 @@
                       (.toBeTruthy))
                   (-> (js/expect (rtl/screen.getByText "5"))
                       (.toBeTruthy))))
+
+(js/global.test "banner component fires an event when pressed"
+                (let [mock-fn (js/jest.fn)]
```
ilmotta: The `mock-fn` binding should be declared inside the anonymous function. This is currently a mistake in other files too, which I believe you based your tests on.

In fact, in general, the entire body of the test should be put inside the anonymous function. This is quite important so that the test runner (Jest) can reliably run all setup/teardown callbacks.

J-Son89: yeah looks like you are correct, basically I need to wrap everything here in a `fn` right?

```
(js/global.test "banner component fires an event when pressed"
           (fn [] (let [mock-fn (js/jest.fn)] ...
```

ilmotta: Precisely, in Javascript we always define the test inside the anonymous function, i.e. we never define a variable *around* the anonymous function. Sure we can define it below describe and use before/after hooks, but that's another thing entirely.

```javascript
describe('some description', () => {
  it('test description', () => {
      // mocks, etc
  })
})
```

J-Son89: yep, I'm familiar with this. Probably just an oversight with me being new to converting js -> cljs. 

I'll update the test files in a separate pr 👍

src/status_im2/contexts/chat/messages/view.cljs
jo-mut: @ilmotta `make lint-fix` seems to be working now just fine now. I have no idea why wasnt before.

ilmotta: Makes sense @jo-mut. Apparently there was a change in the command to collect files before they're passed in to zprint.

jo-mut: nice thank you @ilmotta

J-Son89:
> Hi @J-Son89 Only one little question. In which cases the banners with 'pin' and 'without pin icon should be shown as it is disaplyed per design? Curretly all of them are shown without 'pin' icon in the current PR image

Hi @VladimrLitvinenko, this just depends on the screen that designers use. In all cases so far it shows the icon, but they just wanted to include the option without the pin in case it's needed for later designs.

J-Son89:
> Hi @J-Son89 Thank you for your work. Here is just one low-priority issue. All other looks great
> 
> **ISSUE 1: The count of pinned messages is shifted if short text message is shown.** **Steps to reproduce:**
> 
> 1. Go to 1-1
> 2. Send the message
> 3. Pin it
> 4. Send one more message with a short text
> 5. Pin the last message also
> 
> **Actual result:** The count badge of pinned messages is shifted to the left image
> 
> **Expected result:** The count badge of pinned messages is always located in the right part of the banner image

this should be fixed now :D

VolodLytvynenko:
Hi @J-Son89 thank you for PR. ISSUE 1 is fixed and 'pin' icon is shown. Please fix one more detected issue

**ISSUE 2 The pin banner should not be shown as focused on the expanded keyboard**

**Steps to reproduce:**
1. Go to 1-1 chat
2. Send one message
3. Pin it
4. Open input textbox
5. Start typing using 'enter' button to fulfill the whole textbox


**Actual result:**
The pinned message is focused after textbox is fulfilled
https://user-images.githubusercontent.com/52490791/210358857-80ee3166-179a-4d9f-8ee1-fdb0f9e63469.mp4

**Expected result:**
The textbox is not focused if textbox is used

J-Son89:
> Hi @J-Son89 thank you for PR. ISSUE 1 is fixed and 'pin' icon is shown. Please fix one more detected issue
> 
> **ISSUE 2 The pin banner should not be shown as focused on the expanded keyboard**
> 
> **Steps to reproduce:**
> 
> 1. Go to 1-1 chat
> 2. Send one message
> 3. Pin it
> 4. Open input textbox
> 5. Start typing using 'enter' button to fulfill the whole textbox
> 
> **Actual result:** The pinned message is focused after textbox is fulfilled https://user-images.githubusercontent.com/52490791/210358857-80ee3166-179a-4d9f-8ee1-fdb0f9e63469.mp4
> 
> **Expected result:** The textbox is not focused if textbox is used

I checked this and it should be fine now :)

src/quo2/components/banners/banner/style.cljs
```diff
@@ -3,17 +3,26 @@
...
    :justify-content :center
-   :align-items     :center})
\ No newline at end of file
+   :align-items     :center})
+
+(def icon {:margin-right 8})
+
+(defn text
+  [hide-pin?]
+  {:width (if hide-pin? 307 279)})
```
flexsurfer: please don't use width it should be flexible

VolodLytvynenko:
Hi @J-Son89 . Thanx for PR. The ISSUE 2 s fixed. Here is an additional issues, I think I missed them on the previous build
**ISSUE 3. The pinned banner is not shown within dark mode as per design**
**Actual result**:
image

**Expected result:**
image

**ENV:** IOS, Android

@J-Son89 can you fix this issue4 in the scope of this PR or better to reopen it as a separate follow up?
**ISSUE 4: The last message is not shown within the banner after the chat is reopened**

**Steps to reproduce:**
1. Send 2 messages into 1-1 chat "FIRST message" and "SECOND message"
2. Pin "FIRST message" 
3. Then pin "SECOND message"
4. Reopen the chat

**Actual result:**
 "FIRST message"  is shown as pinned within the banner instead the "SECOND message"

https://user-images.githubusercontent.com/52490791/210566587-a677962b-372e-407b-a3f6-e7a7e28faba7.mp4

**Expected result:**
The last message is shown within the banner

src/quo2/components/banners/banner/style.cljs
J-Son89: sure, will adjust!

J-Son89:
@VladimrLitvinenko 3 & 4 should be addressed now. 
I've noticed that pinned messages don't directly update in the chat screen. Should we create a follow up issue to address that?
i.e go to a chat.
send message "one"
pin message "one" -> pinned message now shows as "one"
send message "two"
pin message "two" 
expected:
  pinned message should shows as "two"
actual:
  pinned message still shows as "one"
  
  the pinned message updates after navigating away and back to the chat screen

VolodLytvynenko:
@J-Son89 thank you for the fixes and for your notification about follow-up. Sure, it is a low priority issue and can be fixed separately. Reported it https://github.com/status-im/status-mobile/issues/14743 here. 
All other issues related to the current PR are tested and can be merged. Thanx!

PR #14631, opened 2022-12-23
Group details screen 3.2

OmarBasem:
Reverting some files from that PR: https://github.com/status-im/status-mobile/pull/14494

PR #14589, opened 2022-12-20
refactor: reformat all clojure code with zprint

yqrashawn:
status: ready

PR #14635, opened 2022-12-26
Images Album

### LINKED ISSUE #14519: Image Messages Collage
An issue to track the creation of the images rendering in messages.

Designs: https://www.figma.com/file/wA8Epdki2OWa8Vr067PCNQ/Composer-for-Mobile?node-id=2827%3A270189&t=OA2AUu8KLlZZqDfw-0

OmarBasem:
fixes: https://github.com/status-im/status-mobile/issues/14519

PR 1 of 2

This PR implements grouping multiple images sent together into an album/collage.

Designs: https://www.figma.com/file/wA8Epdki2OWa8Vr067PCNQ/Composer-for-Mobile?node-id=2827%3A270189&t=zqgZeyfC57sUlzVD-0


src/status_im/chat/models/message_list.cljs
```diff
@@ -20,19 +20,21 @@
            from
            outgoing
            whisper-timestamp
-           deleted-for-me?]}]
+           deleted-for-me?
+           albumize]}]
```
J-Son89: maybe add a ? `albumize?` ?

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+   (map-indexed (fn [index item]
+                  (let [images-count    (count (:album message))
+                        images-size-key (if (< images-count 6) images-count :default)
+                        size            (get-in image-sizes [images-size-key index])]
+                    [rn/view {:key (str index)}
+                     [fast-image/fast-image
+                      {:style  {:width         size
+                                :height        size
+                                :margin-left   (when (and (not= index 0) (not= index 2)) 1)
+                                :margin-bottom (when (or (= index 0) (= index 1)) 1)}
```
J-Son89: could be `(when (< index 2) 1)` ?

src/status_im2/subs/chat/messages.cljs
```diff
@@ -96,63 +97,81 @@
...
+        (-> message-list-seq
+            (chat.db/add-datemarks)
+            (hydrate-messages messages pin-messages)
+            (chat.db/collapse-gaps chat-id
+                                   synced-from
+                                   (datetime/timestamp)
+                                   chat-type
+                                   joined
+                                   loading-messages?)
+            (find-albums))))))
```
J-Son89: does this not change the previous implementation of this sub? Should you not create a new sub for this behaviour? 🤔

status-go-version.json
```diff
@@ -3,7 +3,7 @@
     "_comment": "Instead use: scripts/update-status-go.sh <rev>",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.117.1a",
-    "commit-sha1": "b4bdfd3df6cf5fb91ab2d0e9f3b38e8d1b9703e5",
-    "src-sha256": "1s50q53p1j5ysnbj88c52zb2rh1ms21pnf6bjc7wja8gzcpih2wl"
+    "version": "6abbe98cd20b639d0c7298091f00262e876f3e17",
```
J-Son89: revert? or necessary?

J-Son89 (approved):
looks good, although there's lots of formatting/indentations that don't seem necessary? Is your editor setup correctly? or those files just didn't get picked up before?

src/status_im2/subs/chat/messages.cljs
```diff
@@ -96,63 +97,81 @@
...
+  :chats/chat-no-messages?
+  (fn [[_ chat-id] _]
+    (re-frame/subscribe [:chats/chat-messages chat-id]))
+  (fn [messages]
+    (empty? messages)))
+
+(defn find-albums [messages]
+  (let [all-messages (atom [])
+        albums       (atom {})]
+    (doseq [message messages]
```
flexsurfer: this should be done in more clojure idiomatic way, without atoms and doseq , i could help later

src/status_im2/subs/chat/messages.cljs
flexsurfer: we should be very careful when adding additional computations here, isn't this possible to do on status-go side ?

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+             1 146
+             2 72.5
+             3 72.5
+             4 72.5
+             5 72.5}})
+
+(defn album-message
+  [message]
+  [rn/view
+   {:style {:flex-direction :row
```
flexsurfer: move to style ns

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
+(ns status-im2.contexts.chat.messages.content.album.view
+  (:require [quo2.core :as quo]
+            [quo2.foundations.colors :as colors]
+            [react-native.core :as rn]
+            [react-native.fast-image :as fast-image]
+            [status-im2.contexts.chat.messages.content.album.style :as style]))
+
+(def image-sizes
```
flexsurfer: move to constants ?

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+            :border-radius  12
+            :width          293
+            :overflow       :hidden}}
+   (map-indexed (fn [index item]
+                  (let [images-count    (count (:album message))
+                        images-size-key (if (< images-count 6) images-count :default)
+                        size            (get-in image-sizes [images-size-key index])]
+                    [rn/view {:key (str index)}
+                     [fast-image/fast-image
+                      {:style  {:width         size
```
flexsurfer: move to style ns

src/status_im/chat/models/input.cljs
```diff
@@ -184,9 +184,11 @@
 
 (defn build-image-messages
   [{db :db} chat-id]
-  (let [images (get-in db [:chat/inputs chat-id :metadata :sending-image])]
+  (let [images (get-in db [:chat/inputs chat-id :metadata :sending-image])
+        album-id (str (random-uuid))]
```
flexsurfer, line 188: could you elaborate why we generate it here? and also str is not needed? its already string i believe

src/status_im/chat/models/loading.cljs
```diff
@@ -101,6 +101,12 @@
                                     :on-success #(re-frame/dispatch
                                                   [::mark-all-read-in-community-successful %])}]}))
 
+(defn mark-albums [messages]
+  (let [new-messages (atom [])]
+  (doseq [message messages]
```
flexsurfer: map should be used instead

flexsurfer: again could this be done on status-go side ?

flexsurfer (changes requested):
formatting should be solved in this PR as well with zprint

src/status_im/chat/models/loading.cljs
```diff
@@ -101,6 +101,12 @@
                                     :on-success #(re-frame/dispatch
                                                   [::mark-all-read-in-community-successful %])}]}))
 
+(defn mark-albums [messages]
```
ilmotta: @OmarBasem, this function goes completely against the grain in Clojure. All you need is to map and assoc a new key. Stateful processing is rarely needed in Clojure.

```clojure
(defn mark-albums [messages]
  (map (fn [msg]
         (if (:album-id message)
           (assoc message :albumize true)
           message))
       messages))
```

Something even better IMO is to create smaller predicate functions and reuse them in collection operations. This leads to more functional code, that composes better. Of course, there are exceptions to this recommendation.

```clojure
(defn mark-album [message]
  (if (:album-id message)
    (assoc message :albumize true)
    message))

;; And you can use it as:
new-messages (map mark-album new-messages)
```

src/status_im/chat/models/message_list.cljs
```diff
@@ -44,18 +46,18 @@
...
   [a b]
   (and
-   (not (:system-message? a))
-   (not (:system-message? b))
-   (= (:from a) (:from b))
-   (<= (js/Math.abs (- (:whisper-timestamp a) (:whisper-timestamp b))) group-ms)))
+    (not (:system-message? a))
+    (not (:system-message? b))
+    (= (:from a) (:from b))
+    (<= (js/Math.abs (- (:whisper-timestamp a) (:whisper-timestamp b))) group-ms)))
```
ilmotta: [Nitpick] For math oriented code in Lisps we should avoid one-liners in many occasions. In the example below, anyone can quickly spot that `group-ms` is the second argument of `<=`, whereas in the original code it's not clear if it's perhaps related with the call to `Math.abs`.

Also worth pointing out that as of Clojure 1.11 the core function `abs` should exist in Clojure(Script). We should migrate from 1.10 to 1.11.

```clojure
(<= (js/Math.abs (- (:whisper-timestamp a) (:whisper-timestamp b)))
    group-ms)
```

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+   {:style {:flex-direction :row
+            :flex-wrap      :wrap
+            :border-radius  12
+            :width          293
+            :overflow       :hidden}}
+   (map-indexed (fn [index item]
+                  (let [images-count    (count (:album message))
+                        images-size-key (if (< images-count 6) images-count :default)
+                        size            (get-in image-sizes [images-size-key index])]
+                    [rn/view {:key (str index)}
```
ilmotta: If I'm not mistaken, the `item` in the iteration is an album map, and it has an ID. If that's the case, please use it. We should generally avoid using the iteration index as unique keys, unless we're 100% sure the collection under iteration cannot change due to state changes, otherwise performance is affected.

src/status_im/chat/models/loading.cljs
flexsurfer: why do we even need new flag? why not just use `album-id` as flag ?

src/status_im2/subs/chat/messages.cljs
ilmotta: Yes and we should not move forward with code like this. There's absolutely no need for atoms when processing pure data in subscriptions.

@OmarBasem

It also doesn't have tests and I thought we had some general agreement that tests for layer-3 subs should be written (and this `find-albums` function is used in a layer-3 sub).

It's far from being just about *code correctness*, it's about writing code that another contributor can evolve in the future without having to manually test everything all over again. As much as having working code is great, nobody in this PR (except maybe Omar) who can actually tell if this implementation works well without careful analysis of various bits of data.

There are 3 conditions in `find-albums` and another iteration with filtering (`filterv`). Each condition requires 2 different assertions, summing up to a permutation of many test assertions with different setup data.

status-go-version.json
ilmotta: It seems like we should revert.

I thought we should never merge pointing to a revision. Revisions change after push forces and are only relatively stable if pointing to a revision in the `develop` branch.

There's a particular case where the backend and frontend need to change in lockstep, but it doesn't seem to be the case here. As Andrea once explained to me, if the frontend requires a specific implementation of status-go that has not been released, then it should go more or less like the steps below:

1. Contributor opens the frontend PR when the backend already has the non-released code, and the contributor points to a specific backend revision (not version).
2. Contributor gets the frontend PR approved and potentially tested by QAs.
3. Contributor is now positive the backend implementation works, so they open a PR in the backend, get it approved and release a new version.
4. Contributor changes the frontend PR to point to the newly released backend version (not revision).
5. Contributor rebases the frontend PR, squash and merge, potentially requesting another round of review from QAs before merge.

src/status_im/chat/models/message_list.cljs
```diff
@@ -116,11 +118,11 @@
...
-           :display-username? (and last-in-group?
-                                   (not system-message?)
-                                   (not outgoing)
-                                   (not one-to-one?))
-           :last-in-group?    last-in-group?)))
+      :display-username? (and last-in-group?
+                              (not system-message?)
+                              (not outgoing)
+                              (not one-to-one?))
+      :last-in-group? last-in-group?)))
```
briansztamfater: Seems vertical aligning got reverted here

src/status_im/constants.cljs
```diff
@@ -14,6 +14,7 @@
 (def ^:const content-type-community 9)
 (def ^:const content-type-gap 10)
 (def ^:const content-type-contact-request 11) ;; TODO: temp, will be removed
+(def ^:const content-type-album 12)
```
briansztamfater: You are adding the same constant in `src/status_im2/common/constants.cljs`, is this needed here?
I would suggest to keep just one constant

src/status_im2/contexts/chat/messages/content/album/view.cljs
briansztamfater: Agree, it's easier to read

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,50 @@
...
+                                :height        size
+                                :margin-left   (when (and (not= index 0) (not= index 2)) 1)
+                                :margin-bottom (when (or (= index 0) (= index 1)) 1)}
+                       :source {:uri (:image (:content item))}}]
+                     (when (and (> images-count 6) (= index 5))
+                       [rn/view {:style style/overlay}
+                        [quo/text
+                         {:weight :bold
+                          :size   :heading-2
+                          :style  {:color colors/white}} (str "+" (- images-count 5))]])]))
```
briansztamfater: Move style to style ns

src/status_im2/subs/chat/messages.cljs
OmarBasem: It is intended to be changed. When need to always "find albums" when loading new messages.

status-go-version.json
OmarBasem: The backend impl. on status-go is already merged. I forgot to update it.

OmarBasem:
> looks good, although there's lots of formatting/indentations that don't seem necessary? Is your editor setup correctly? or those files just didn't get picked up before?

Zprint is not working properly with IntelliJ. I am facing a similar problem in this PR also: https://github.com/status-im/status-mobile/pull/14631

src/status_im2/subs/chat/messages.cljs
OmarBasem: @ilmotta I was going to write tests for it, but wanted the function to be reviewed first :)

src/status_im2/subs/chat/messages.cljs
OmarBasem: Nope. Albums are not an actual content-type. Images that are sent together we present them to the user like a single album message, but still every image is a seperate message.

du82:
I think it would be great if a group of 6 images was shown in a 2/3 config, while groups of 9 images were shown in a 3/3 grid. 

This makes images easier to see, and is how WeChat does it

Screenshot_2022-12-26-22-14-26-43_65ad86f86fd2321a85272a99323da10f

src/status_im/chat/models/input.cljs
OmarBasem: @flexsurfer Images that are sent together we need to give them the same `album-id`.
`random-uuid` return some object. I have to use `str` to convert it to string.

src/status_im/chat/models/loading.cljs
OmarBasem: https://github.com/status-im/status-mobile/pull/14635#discussion_r1057427677

src/status_im/chat/models/loading.cljs
OmarBasem: @flexsurfer  When a user receives a group of images while inside the chat, we do not group them. We group images into albums when a user is entering the chat screen. The `messages-loaded` function is called when a user is entering the chat screen, that's why I am adding the `albumize` boolean here.

OmarBasem: Thanks @ilmotta for the snippet

OmarBasem: @ilmotta both of these snippets did not work

src/status_im/chat/models/message_list.cljs
OmarBasem: Thanks for this information. This code is not part of the PR. This is unintended formatting changes caused by Zprint

src/status_im/chat/models/message_list.cljs
OmarBasem: I am currently facing a problem using Zprint with IntelliJ, also in this PR: https://github.com/status-im/status-mobile/pull/14631

src/status_im2/contexts/chat/messages/content/album/view.cljs
OmarBasem: This particular style is a single prop so I am keeping it inline. Moved the other bigger styles.

OmarBasem:
Thanks @du64. I will share your feedback with the design team.

src/status_im2/subs/chat/messages.cljs
ilmotta: No problem @OmarBasem. Next time you open a PR with WIP pieces of code, can I ask you to add a hint in the PR for reviewers? Like a FYI comment, or a list of WIP things in the PR description. That would be very helpful!

src/status_im/chat/models/loading.cljs
ilmotta: Although I didn't test the snippets, I'm curious enough now to have coded this in a REPL, and it works as expected @OmarBasem. But perhaps I'm missing something.

```clojure
(defn mark-albums
  [messages]
  (let [new-messages (atom [])]
    (doseq [message messages]
      (swap! new-messages conj (if (:album-id message) (assoc message :albumize true) message)))
    @new-messages))

(defn mark-album
  [message]
  (if (:album-id message)
    (assoc message :albumize true)
    message))

(let [messages [{:id "0x1" :album-id 1}
                {:id "0x2"}
                {:id "0x3" :album-id 3}]
      expected [{:id "0x1" :album-id 1 :albumize true}
                {:id "0x2"}
                {:id "0x3" :album-id 3 :albumize true}]]
  ;; Original
  (assert (= expected
             (mark-albums messages)))

  ;; New implementation
  (assert (= expected
             (map mark-album messages))))

```

src/status_im/chat/models/message_list.cljs
ilmotta: Could you give us more context on the problem @OmarBasem? `zprint` is an external tool, independent of IDE, so I'm genuinely curious about the problem. I know we have other IntelliJ users in the team, so maybe it's worth asking.

src/status_im2/subs/chat/messages.cljs
OmarBasem: @ilmotta Sure!

src/status_im/chat/models/loading.cljs
OmarBasem: @ilmotta sorry my bad! It is working. I updated `albumize` to `albumize?` and did not change it in the snippet here. Thank you!

src/status_im/chat/models/message_list.cljs
OmarBasem: @ilmotta when I run `make lint-fix` or even this (https://github.com/status-im/status-mobile/pull/14589#issuecomment-1359515798) it keeps messing up the formatting of other files/code which is not part of the PR. I am not sure if this has to do with the IDE. Also there is a discussion here https://github.com/status-im/status-mobile/pull/14631/#pullrequestreview-1229075222

src/status_im/chat/models/loading.cljs
ilmotta: Thanks for giving another try @OmarBasem o/

src/status_im/chat/models/message_list.cljs
ilmotta: That's currently what `make lint-fix` is supposed to do, it fixes everything :shrug: We could have a separate make target that only formats what's changed in the branch under development or something like that, maybe that's going to be better for you.

The current `lint-fix` target collects files using:

```shell
git ls-files | grep -e '\.clj$' -e '\.cljs$' -e '\.cljc$' -e '\.edn$'
```

This returns all files. Check in your machine to be sure.

I'm not sure about the best approach for IntelliJ users, but relying on `make lint-fix` is supoptimal IMO. Perhaps you and @qfrank would have a better experience using a git hook that only formats the files you care about.

OmarBasem: It is now fixed on latest develop :)

src/status_im2/contexts/chat/messages/content/album/style.cljs
```diff
@@ -0,0 +1,24 @@
+(ns status-im2.contexts.chat.messages.content.album.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def album-container
+  {:flex-direction :row
+   :flex-wrap      :wrap
+   :border-radius  12
+   :width          293
```
flexsurfer: we should never use width, all layouts should be flexible

src/status_im2/contexts/chat/messages/content/album/view.cljs
flexsurfer: yes we move styles if there are more than 1-2 lines of props

src/status_im2/subs/chat/messages.cljs
```diff
@@ -102,6 +102,25 @@
...
+                (if (and album-id (> (count (get @albums album-id)) 3))
+                  (conj (filterv #(not= album-id (:album-id %)) acc)
+                        {:album        (get @albums album-id)
+                         :album-id     album-id
+                         :message-id   album-id
+                         :content-type constants/content-type-album})
+                  (conj acc message))))
+            []
+            messages)))
+
```
OmarBasem: @ilmotta @flexsurfer does it look better now?

flexsurfer: yeah i was looking at this, sorry for the delay, i'll take a look on your solution, for now it seems like `(conj acc message)` in the middle is a leftover

flexsurfer: could you please explain in words these two lines ?
```clojure
(if (and album-id (> (count (get albums album-id)) 3))
              (conj (filterv #(not= album-id (:album-id %)) acc)
```

flexsurfer:
```clojure
(get (reduce (fn [{:keys [messages albums]} message]
               (let [album-id (when (:albumize? message) (:album-id message))
                     albums (cond-> albums album-id (update album-id #(conj % message)))
                     messages (if (and album-id (> (count (get albums album-id)) 3))
                                (conj (filterv #(not= album-id (:album-id %)) messages)
                                      {:album        (get albums album-id)
                                       :album-id     album-id
                                       :message-id   album-id
                                       :content-type constants/content-type-album})
                                (conj messages message))]
                   {:messages messages
                    :albums albums}))
             {:messages []
              :albums {}}
             messages)
     :messages)
```
this is my suggestion, but honestly im not sure how it works, condition looks strange

flexsurfer: `:albumize?` is confusing

flexsurfer: because we don't check it in the filter

ilmotta: @OmarBasem, we really need tests for this kind of code to be merged in prod. The nice thing is that even if you change the implementation of `find-albums`, unit tests would not care if coded correctly, they would still be the same and you'd be free to refactor knowing that it all works.

ilmotta: @OmarBasem, as @flexsurfer pointed out, we shouldn't use an atom to store temporary state, just use a map as the initial reducing value. Atom deref's are much more expensive than simple maps, which is particularly important in subscriptions.

flexsurfer:
seems like `:albumize?` is not needed, it's only confusing

src/status_im2/subs/chat/messages.cljs
ilmotta: @OmarBasem, I reviewed in more detail the changed subscription `:chats/raw-chat-messages-stream`. A couple points:

1. The final call to `find-albums` seems to be a breaking change. I haven't seen this PR update all the other places where the subscription is used, and there are a handful. Are you sure everything is really working? I actually don't understand how it could work, maybe it's me being shortsighted.

Original implementation is a sequence of message entities with some additional keys. New implementation is also a vector, but when the messages have an album ID, they're grouped into a new map that's incompatible with a message entity.

Places using the affected subscription:

- `status-im.ui.screens.chat.views/messages-view`
- `status-im2.contexts.chat.messages.list.view/messages-list`
- Subscription `:chats/profile-messages-stream`.

2. I would also suggest changing the name `find-albums`. `find` is a word used for search operations, and searching doesn't generally change the shape of the data being searched.
   
3. Why do we need the `albumize?` keyword assoc'ed to messages? Wouldn't it be enough to simply check the existence of the `:album-id` key? In this PR we see the function `models.loading/mark-album` does exactly this, it always assocs `:albumize?` as true if there's a `:album-id` value, so `:albumize?` seems to be redundant.

4. A good chunk of the logic in `find-albums` is related to indexing messages by `album-id`, but this can be simplified by using `group-by`.

5. To reiterate, missing tests :man_shrugging: 

Feel free to ping me if you'd like to pair program on some of these points. I have some ideas, but writing it all takes too much time via GH threads.

src/status_im2/contexts/chat/messages/content/album/view.cljs
```diff
@@ -0,0 +1,28 @@
...
+   {:style style/album-container}
+   (map-indexed (fn [index item]
+                  (let [images-count    (count (:album message))
+                        images-size-key (if (< images-count 6) images-count :default)
+                        size            (get-in constants/album-image-sizes [images-size-key index])]
+                    [rn/view {:key (:message-id item)}
+                     [fast-image/fast-image
+                      {:style  (style/image size index)
+                       :source {:uri (:image (:content item))}}]
+                     (when (and (> images-count 6) (= index 5))
```
ilmotta: Would be better if we parameterized the magic number `6`. If, in the future, this number changes from 6 to 7 for instance, we should just have to change the constant value.

src/status_im2/subs/chat/messages.cljs
OmarBasem: > could you please explain in words these two lines ?
> 
> ```clojure
> (if (and album-id (> (count (get albums album-id)) 3))
>               (conj (filterv #(not= album-id (:album-id %)) acc)
> ```

We group images into albums that has a count of more than 3. Before adding the "album" message into the list of messages, we need to filter out the image messages of that album that was added into the list.

OmarBasem:
> seems like `:albumize?` is not needed, it's only confusing

@flexsurfer For example, when a user receives a list of 4 image messages while inside the chat screen we shouldn't group the images into albums. When the user exists the chat screen then enters the chat screen again, we now need to group the images into albums (like WhatsApp). The `albumize?` boolean is used to know whether we need to group these images into albums now or not. The `album-id` can't be used for this because it will always be there.

src/status_im2/contexts/chat/messages/content/album/view.cljs
OmarBasem: Parameterized the magic number 😄

OmarBasem:
@ilmotta  replying to your long message 😄:

1. It breaks nothing! 🧙
2. Changed it to `albumize-messages`
3. For example, when a user receives a list of 4 image messages while inside the chat screen we shouldn't group the images into albums. When the user exists the chat screen then enters the chat screen again, we now need to group the images into albums (like WhatsApp). The `albumize?` boolean is used to know whether we need to group these images into albums now or not. The `album-id` can't be used for this because it will always be there.
4. Could you elaborate please?
5. Done! 🧪

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -161,10 +161,10 @@
      (let [min-y               112
```
flexsurfer: please rollback this file

flexsurfer (approved):

ilmotta:
> @ilmotta replying to your long message smile:
> 
> 1. It breaks nothing!
> 2. Changed it to `albumize-messages`
> 3. For example, when a user receives a list of 4 image messages while inside the chat screen we shouldn't group the images into albums. When the user exists the chat screen then enters the chat screen again, we now need to group the images into albums (like WhatsApp). The `albumize?` boolean is used to know whether we need to group these images into albums now or not. The `album-id` can't be used for this because it will always be there.
> 4. Could you elaborate please?
> 5. Done!

1. Magician of software at play here ;)
3. Got it @OmarBasem. This kind of explanation is rather useful, because some reviewers (like me) got confused. It could be added as a comment in the function `models.loading/mark-album` if you want.
5. :rocket: :rocket: 

About 4) and `group-by`. No changes suggested for this PR. If you find yourself creating an empty map and `conj`ing by some ID (like you do in the subscription), then it's worth seeing if `group-by` can help you simplify the solution.

src/status_im2/subs/chat/messages.cljs
```diff
@@ -102,6 +102,25 @@
  (fn [messages]
    (empty? messages)))
 
+(defn albumize-messages
+  [messages]
+  (get (reduce (fn [{:keys [messages albums]} message]
+                 (let [album-id (when (:albumize? message) (:album-id message))
+                       albums   (cond-> albums album-id (update album-id #(conj % message)))
```
ilmotta: @OmarBasem, using `update` like this with the anonymous function is redundant, because `update` already calls the function by passing the updatable value as the first argument. This is by design, very useful.

```clojure
;; Old
(update album-id #(conj % message))

;; New
(update album-id conj message)
```

```clojure
;; Proof
(assert [1 2 3 4]
        (update {"0x1" [1 2 3]} "0x1" conj 4))
```

src/status_im2/subs/subs_test.cljs
```diff
@@ -40,3 +42,24 @@
...
+  [{:album        [{:message-id "0x444" :album-id "abc" :albumize? true}
+                   {:message-id "0x333" :album-id "abc" :albumize? true}
+                   {:message-id "0x222" :album-id "abc" :albumize? true}
+                   {:message-id "0x111" :album-id "abc" :albumize? true}]
+    :album-id     "abc"
+    :message-id   "abc"
+    :content-type constants/content-type-album}
+   {:message-id "0x555" :album-id "efg" :albumize? true}])
+
+(deftest find-albums
```
ilmotta: @OmarBasem, these tests should be added to a new test file in `src/status_im2/subs/chat/messages_test.cljs`. This top level test file will be gone soon. I might even do it this week.

So the idea is to write one separate test file for every subscription file.

src/status_im2/subs/chat/messages.cljs
OmarBasem: Thanks!

src/status_im2/subs/subs_test.cljs
OmarBasem: @ilmotta Done!

ilmotta (approved):

flexsurfer:
would be cool to make manual testing for such PRs, cc @churik

PR #14636, opened 2022-12-26
Add spacing for cross button on Android

### LINKED ISSUE #14620: The cross button is cut off for images in the input field (Android)
# Bug Report

## Problem 

Screenshot 2022-12-22 at 13 29 28

### Reproduction

1. Open a chat
2. Select an image from the gallery
3. Confirm selection

### Additional Information

- Status version: nightly 22/12
- Operating System: Android

cc @ibrkhalil could you take a look at this issue as well when you have time, please? 
As far as I understood, you are the boss of the new input field :)

ibrkhalil:
fixes #14620 
image

### Summary
status: ready


src/status_im/ui2/screens/chat/composer/images/view.cljs
```diff
@@ -3,11 +3,13 @@
...
             [react-native.core :as rn]
-            [status-im.ui2.screens.chat.composer.style :as style]))
+            [status-im.ui2.screens.chat.composer.style :as style]
+            [status-im.utils.platform :as platform]))
 
 (defn image
   [item]
-  [rn/view
+  [rn/view (when platform/android?
+             {:style {:margin-top 8}})
```
flexsurfer: this looks like magic, why we need to add margin-top and only for Android, looks not obvious

ibrkhalil: I'll find out why Android only adds the spacing/collapsing and elaborate here

ibrkhalil: But just a wild guess, I think when the images component was being built .. It's UI was mostly built on iOS and didn't cater for all cases for Android

flexsurfer: but i mean it shouldn't be done in every view, this padding should set only once somewhere in general view i guess

ibrkhalil: https://github.com/facebook/react-native/issues/25592
:)

ibrkhalil: Should I do custom overrides for each platform till we finish the RN upgrade (And add a comment to remove it once RN upgrade is done), As this issue is fixed in the newer RN version?

flexsurfer: i think its not related, something completely wrong in the implementation its not about bug

PR #14618, opened 2022-12-22
feat: undo delete with toast

yqrashawn:
this add undo toast to delete message for everyone and delete message for me

1. delete action will trigger undo toast
1. continue delete within undo time limit will update existing undo toast, `1 message deleted` -> `2 messages deleted`
1. continue delete within undo time limit will reset all undoable deletes undo time limit
1. undo when toasting `4 messages deleted` will undo 4 messages
1. swaped the dark/light theme of toast component based on this [figma design](https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=5713%3A624120&t=h8EZBmIX2jH49yLS-4)

https://user-images.githubusercontent.com/15090582/209910901-24f0724c-e812-49bf-967e-ae6f9ace60c7.mp4

image

status: ready

PR #14645, opened 2022-12-27
chore: tweak zprint conf

yqrashawn:
1. prev ns rule is not following community rule's `better` option, fix it by using builtin `:how-to-ns` rule, https://stuartsierra.com/2016/clojure-how-to-ns.html
1. remove unused rule `:rj-var`
1. `make lint-fix` will format all files tracked by git, instead of only staged files
1. [ ] try to solve https://github.com/orgs/status-im/teams/mobile-devs/discussions/1?from_comment=1 https://github.com/kkinnear/zprint/issues/282 (in seperate PR)

~**NOTE**: this requires another reformat, will change all `(ns ...)`~

status: ready

### DIFF
--- a/.zprintrc
+++ b/.zprintrc
@@ -8,6 +8,8 @@
              ;; no comma in map
              :no-comma
 
+             :how-to-ns
+
              :custom-justify
 
              ;; respect all newlines
@@ -26,9 +28,6 @@
   "deftest-sub"     :arg1-body
   "wait-for"        :arg1-body
   "with-deps-check" :arg1-body
-  "ns"              [:arg1-body
-                     {:list   {:respect-nl? false}
-                      :vector {:respect-nl? false}}]
   "->"              [:noarg1-body
                      {:list               {:constant-pair? false :force-nl? false}
                       :next-inner-restore [[:list :constant-pair?]]}]
--- a/.zprintrc
+++ b/.zprintrc
@@ -40,7 +40,6 @@
  :remove    {:fn-force-nl #{:noarg1-body}}
  :style-map
  {:no-comma       {:map {:comma? false}}
-  :rj-var         {:pair {:justify {:max-variance 1000}}}
   :custom-justify
   {:doc     "Justify everything using pre-1.1.2 approach"
    :binding {:justify? true :justify {:max-variance 1000}}
--- a/.zprintrc
+++ b/.zprintrc
@@ -8,8 +8,6 @@
              ;; no comma in map
              :no-comma
 
-             :how-to-ns
-
              :custom-justify
 
              ;; respect all newlines
--- a/Makefile
+++ b/Makefile
@@ -294,10 +294,10 @@ lint: ##@test Run code style checks
 
 lint-fix: export TARGET := default
 lint-fix: ##@test Run code style checks and fix issues
-	TARGETS=$$(git diff --diff-filter=d --cached --name-only | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn$$ || echo shadow-cljs.edn) && \
-	clojure-lsp clean-ns --filenames $$(echo $$TARGETS | xargs | sed -e 's/ /,/g') && \
-	zprint '{:search-config? true}' -fw $$TARGETS && \
-	zprint '{:search-config? true}' -fw $$TARGETS
+	ALL_CLOJURE_FILE=$$(git ls-files | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn$$) && \
+	clojure-lsp clean-ns && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE
 
 
 shadow-server: export TARGET := clojure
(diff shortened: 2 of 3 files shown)


siddarthkay (approved):

ibrkhalil (approved):

flexsurfer:
also have a question, why `make lint` says all good, but if run fix-lint on all files there will be changes?

vkjr:
Oh, thank you for changing the "only staged" approach. It tricked me once

vkjr (approved):

ilmotta:
@yqrashawn, I think the issue I found in my local tests is what @flexsurfer reported as well.

The zprint rule `:how-to-ns` is not fully following Stuart Sierra's recommendations, in particular sorting namespaces.

Steps to reproduce this issue with `zprint 1.2.4` (our version installed with Nix):

1. `make shell`
2. Open a file with unordered namespaces, like `src/status_im/transport/message/protocol.cljs`
3. Format it, `zprint '{:search-config? true}' -sw src/status_im/transport/message/protocol.cljs`
4. Run the same command in step 3, however many times, and zprint will never sort the namespaces.

The reason why `make lint-fix` sorts namespaces is because it also runs the command `clojure-lsp clean-ns`, not because of zprint.

So everybody should just be aware that we'll sometimes merge code without sorting namespaces, just as it is today. This is going to be confusing because collaborators will think that zprint is the source of truth for all things formatting and that it respects Stuart's recommendations, which it doesn't.

So we have a couple options IMO:

1. Don't use the `:how-to-ns`, it's not really following Stuart Sierra's recommendations.
2. Don't rely on `clojure-lsp clean-ns` in `make lint-fix`. Remove it from all make targets. This means zprint would be the only source of truth for formatting.
3. Keep using `zprint` and `clojure-lsp` to clean up and sort namespaces, i.e. move forward with this PR.

I'm okay with all options, but I'm raising awareness here. Maybe we should report another bug @yqrashawn.

Note: one of the major things the command `clojure-lsp clean-ns` does is removal of unused namespaces, but this is not that important in our codebase because we use clj-kondo (local and CI), which already reports unused namespaces, so no collaborators should be able to merge code with unused namespaces.

ilmotta:
@yqrashawn, I think the rule `:how-to-ns` is generally broken. I gave this branch another go, but almost all rules are not respected, so I'm not sure what's the value of using `:how-to-ns`. Could you please confirm on your end if it's just my environment (I doubt it because I'm testing inside our Nix shell). If you confirm the issues below also fail for you, I'd say we should remove `:how-to-ns` from zprint until we can actually rely on this rule.

For each one of the recommendations below, I ran `zprint '{:search-config? true}' -sw <file>` twice. None of the rules below work (from Stuart Sierra's blog):

- `Use lists (parentheses) for the top-level clauses such as (:require ...)`
- `Put :refer’d or :exclude’d symbols in a vector (square brackets).`
- `Sort :require’d namespaces lexicographically`. See my previous comment for details https://github.com/status-im/status-mobile/pull/14645#issuecomment-1365943967.
- `When a :require’d namespace has both :as and :refer, put :as before :refer.` The file `src/status_im/ui/screens/chat/components/hooks.cljs` is one good example.

yqrashawn:
> also have a question, why `make lint` says all good, but if run fix-lint on all files there will be changes?

@flexsurfer might be caused by
1. `make lnt` only check on staged files, 
2. there's `clean-ns` in `make lint-fix` but not in `make lint`

---

@ilmotta the `:how-to-ns` zprint rule only following the newline and indentation part [`Make newlines and indentation match 'how to ns'`](https://github.com/kkinnear/zprint/blob/0b5e0242240f36acbf706a96c417e38113295896/src/zprint/config.cljc#L782)
we currently rely on `clojure-lsp clean-ns` to sort namespace, zprint doesn't do ns sort

In style guide, it's `better` to have newline after `:require`, it's `good` have no newline, [`clojure-lsp clean-ns` is following the `better` one](https://clojure-lsp.io/settings/#next-line)
And there's no need for another reformat now.

`:how-to-ns` is used here to try to simulate more `clojure-lsp clean-ns` change with `zprint` and
1. check them in `make lint`
2. don't reformat too much after `clojure-lsp clean-ns`
3. it's following the `better` one

but there are still differences.

I do think `clojure-lsp clean-ns` give us more trouble than convenience. I'll remove `clojure-lsp` and `:how-to-ns`

And there's no need for another reformat now

yqrashawn:
another thing is it seems `make lint` don't work in CI, file in changed in this commit and these lines won't pass `make lint` but CI passes https://github.com/status-im/status-mobile/pull/14618/files#diff-f19099d11cab857c4ab6061a65d7054b139f4e5feebd3d01860382f554cdd7c4R5-R6

https://ci.status.im/blue/organizations/jenkins/status-mobile%2Fprs%2Ftests/detail/PR-14618/3/pipeline/35

yqrashawn:
Ok, old one is `git diff --diff-filter=d --cached --name-only src && echo src`, so it's check all files in src if any files changed in src

ilmotta:
Btw @yqrashawn, I find this documentation (https://prettier.io/docs/en/precommit.html) incredibly insightful about some patterns used to grab files and which git hooks to use.

Maybe it gives you some ideas.

PR #14661, opened 2022-12-29
Move photo selector to status-im2

### LINKED ISSUE #14655: Add access ids for objects shown into 'photo selector' overlay
**Steps to reproduce:**
1. Open 1-1 chat
2. Tap 'image icon' shown in the textbox

**Actual result:**
The following buttons are without 'access id':
1. images within selector
image

2. The [confirm selection] button
image

3. 'Clear' button
image

4. Mark inside the image
image

**Expected result:**
all elements are with acc id

OmarBasem:
fixes: #14655

Moves photo selector to status-im2 and adds access ids.


src/status_im/ui/screens/screens.cljs
```diff
@@ -206,10 +205,6 @@
     :options   {:topBar {:visible false}}
     :component pin-messages/pinned-messages}
 
-   {:name      :photo-selector
```
flexsurfer, line 210: you need to add it to the new screens then?

OmarBasem: Photo selector is a bottom sheet. This was added there by mistake.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+            [utils.re-frame :as rf]))
+
+(def selected (reagent/atom []))
+
+(defn bottom-gradient
+  [chat-id selected-images]
+  [:f>
+   (fn []
+     (let [safe-area (safe-area/use-safe-area)]
+       (when (or (pos? (count @selected)) selected-images)
```
ilmotta: In Clojure you check for existence with `(seq collection)`.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+           :start  {:x 0 :y 1}
+           :end    {:x 0 :y 0}
+           :style  (style/gradient-container safe-area)}
+          [quo2/button
+           {:style               {:align-self        :stretch
+                                  :margin-horizontal 20}
+            :on-press            #(do
+                                    (rf/dispatch [:chat.ui/clear-sending-images chat-id])
+                                    (doseq [item @selected]
+                                      (rf/dispatch [:chat.ui/camera-roll-pick item]))
```
ilmotta: Very light suggestion.

Some other places in the code have a similar need to dispatch multiple events.

One could easily add this utility to `utils.re-frame`, which is a bit more flexible.

```clojure
(defn dispatch-n
  [events]
  (doseq [event events]
    (dispatch event)))
```

And then use like this:

```clojure
(rf/dispatch-n (map (fn [item] [:chat.ui/camera-roll-pick item])
                    @selected))
```

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+                                    (reset! selected [])
+                                    (rf/dispatch [:bottom-sheet/hide]))
+            :accessibility-label :confirm-selection}
+           (i18n/label :t/confirm-selection)]])))])
+
+(defn clear-button
+  []
+  (when (pos? (count @selected))
```
ilmotta: In Clojure we use `seq` instead.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
+(ns status-im2.contexts.chat.photo-selector.view
+  (:require [i18n.i18n :as i18n]
+            [quo.components.safe-area :as safe-area]
+            [quo2.components.notifications.info-count :as info-count]
+            [quo2.core :as quo2]
```
ilmotta: Please, alias just as `quo` so that in the future only the namespaces have to change once we eliminate the old `quo`.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+       (when (or (pos? (count @selected)) selected-images)
+         [linear-gradient/linear-gradient
+          {:colors [:black :transparent]
+           :start  {:x 0 :y 1}
+           :end    {:x 0 :y 0}
+           :style  (style/gradient-container safe-area)}
+          [quo2/button
+           {:style               {:align-self        :stretch
+                                  :margin-horizontal 20}
+            :on-press            #(do
```
ilmotta: https://guide.clojure.style/#no-multiple-forms-fn-literals

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+            [quo2.core :as quo2]
+            [quo2.foundations.colors :as colors]
+            [react-native.core :as rn]
+            [react-native.linear-gradient :as linear-gradient]
+            [reagent.core :as reagent]
+            [status-im2.contexts.chat.photo-selector.style :as style]
+            [status-im.utils.core :as utils]
+            [utils.re-frame :as rf]))
+
+(def selected (reagent/atom []))
```
ilmotta, line 15: You can avoid the global state and make it local to the top-level component, then pass it as an argument to downstream functions. This is literally a global mutable var, initialized outside the React rendering cycle, so I think it's quite obvious we should avoid this.

Even though other places in the codebase do this, IMO this is not the way to go in React, in Re-frame, and even in Clojure apps in general.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+      :accessibility-label :clear}
+     [quo2/text {:weight :medium} (i18n/label :t/clear)]]))
+
+(defn image
+  [item index _ {:keys [window-width]}]
+  [rn/touchable-opacity
+   {:active-opacity      1
+    :on-press            (fn []
+                           (if (some #{item} @selected)
+                             (reset! selected (vec (remove #(= % item) @selected)))
```
ilmotta: Most of the time we can replace `reset!` by `swap!`. Clojure was really well thought out to help us write small, reusable functions.

Two advantages:

1. More flexible, reusable.
2. Should perform slightly better due to removal of the anonymous function on every render.

```clojure
(defn remove-selected
  [coll item]
  (vec (remove #(= % item) coll)))

;; Example usage
(swap! selected remove-selected item)
```

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+     [rn/view {:style (style/overlay window-width)}])
+   (when (some #{item} @selected)
+     [info-count/info-count
+      {:style               style/image-count
+       :accessibility-label (str "count-" index)}
+      (inc (utils/first-index #(= item %) @selected))])])
+
+(defn photo-selector
+  [chat-id]
+  (rf/dispatch [:chat.ui/camera-roll-get-photos 20])
```
ilmotta, line 78: Is this really correct? This `dispatch` is definitely going to be called multiple times, once for every render. The component shouldn't have side-effect calls in free form like this, they should at least exist inside a useEffect if the re-frame machinery isn't sufficient.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+  (let [selected-images (keys (get-in (rf/sub [:chat/inputs]) [chat-id :metadata :sending-image]))]
```
ilmotta: This extraction could be a separate subscription. Notice that by not having a subscription and doing the extraction in the UI, re-frame will unnecessarily re-render the entire component even if the end result is the same (the keys of the map).

We shouldn't always create subscriptions, but we should be aware that subscriptions can substantially reduce the amount of time wasted on rendering cycles, when used appropriately.

ilmotta (commented):
@OmarBasem, one thing that was very confusing to me: the PR says "move photo selector to status-im2", but where is the *move* part? I'm seeing mostly green code, so where's the old code? Are we going to remove it?

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+          [rn/view
+           {:style {:flex-direction :row
+                    :position       :absolute
+                    :align-self     :center}}
+           [quo2/text {:weight :medium} (i18n/label :t/recent)]
+           [rn/view {:style (style/chevron-container)}
+            [quo2/icon :i/chevron-down {:color (colors/theme-colors colors/neutral-100 colors/white)}]]]
+          [clear-button]
+          [rn/flat-list
+           {:key-fn                  (fn [item] item)
```
ilmotta: You can just use `identity` here.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+            :render-fn               image
+            :render-data             {:window-width window-width}
+            :data                    camera-roll-photos
+            :num-columns             3
+            :content-container-style {:width          "100%"
```
ilmotta: Could be extract to style namespace.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+             end-cursor                                  (rf/sub [:camera-roll/end-cursor])
+             loading?                                    (rf/sub [:camera-roll/loading-more])
+             has-next-page?                              (rf/sub [:camera-roll/has-next-page])]
+         [rn/view {:style {:height (- window-height (:top safe-area))}}
+          [rn/touchable-opacity
+           {:on-press #(js/alert "Camera: not implemented")
+            :style    (style/camera-button-container)}
+           [quo2/icon :i/camera {:color (colors/theme-colors colors/neutral-100 colors/white)}]]
+          [rn/view
+           {:style {:flex-direction :row
```
ilmotta: Could be extracted to style namespace.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+  [chat-id]
+  (rf/dispatch [:chat.ui/camera-roll-get-photos 20])
+  (let [selected-images (keys (get-in (rf/sub [:chat/inputs]) [chat-id :metadata :sending-image]))]
+    (when selected-images
+      (reset! selected (vec selected-images)))
+    [:f>
+     (fn []
+       (let [{window-height :height window-width :width} (rn/use-window-dimensions)
+             safe-area                                   (safe-area/use-safe-area)
+             camera-roll-photos                          (rf/sub [:camera-roll/photos])
```
ilmotta: No action on your part: FYI: I know we've discussed this topic in the past, but just a reminder, these top level subscriptions impact re-frame's performance because they are root subscriptions and need to be recomputed on *every* state change.

src/status_im/ui2/screens/chat/composer/view.cljs
```diff
@@ -15,7 +15,7 @@
             [status-im.ui2.screens.chat.composer.mentions :as mentions]
             [status-im.ui2.screens.chat.composer.reply :as reply]
             [status-im.ui2.screens.chat.composer.style :as style]
-            [status-im.ui2.screens.chat.photo-selector.view :as photo-selector]
```
J-Son89, line 18: is it possible to also delete the file "status-im.ui2.screens.chat.photo-selector.view :as photo-selector" as part of this pr?

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Hmm.. what is the benefit of this? I think `doseq` is more concise?

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Good to know

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: But that would be kinda "prop drilling". There are 4 different components in that file, all of which use the `selected` state, so I think making it global in that file is better (?)

The `selected` state is updated `on-press` on an image, not part of any lifecycle.

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Thank you!

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: This is a functional component. The part before ` [:f> (fn []` gets called once.

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Is this a function that just returns the argument as it is?

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Is `(rf/sub [:camera-roll/XYZ])` a root sub?

What I understood is that `(rf/sub [:camera-roll-photos])` would be a root sub, and changing it to `camera-roll/XYZ` turns in into a map (?)

src/status_im2/contexts/chat/photo_selector/view.cljs
ilmotta: The general idea is that `dispatch-n` is a known name in re-frame (although it can only be used in event handlers). So when the developer reads it, it should be obvious what it does.

But that's all I can say about it, and that's why I emphasized it's a "very light suggestion".

src/status_im2/contexts/chat/photo_selector/view.cljs
ilmotta: > But that would be kinda "prop drilling". There are 4 different components in that file, all of which use the selected state, so I think making it global in that file is better (?)

But that's exactly the argument developers often use about any state in a any application ever, "let me create a global state here so that I don't have to pass things around". So I don't see this as a good enough reason.

Prop drilling is a much more severe issue IMO, where components have to pass state deep down the component hierarchy, usually across lots of different files and it causes massive maintenance issues. Here it's just 4 functions in a single namespace.

> The selected state is updated on-press on an image, not part of any lifecycle.

It really is part of a lifecycle, even though it's not triggered immediately. `selected` is a reactive atom, and changing this global state triggers a whole cascade of reprocessing in Re-frame > Reagent > React.

---

So we can all disagree about this topic of how to manage local state in components, but it's a constant source of discussion in PRs. Maybe it's about time we have some definition around this topic (not in this PR of course @OmarBasem, so I suggest we leave the discussion for another moment).

src/status_im2/contexts/chat/photo_selector/view.cljs
ilmotta: The subscription `:camera-roll/XYZ` is just a single keyword. We could transform it into `:camera-roll.iphone.14/x/y/z` and it would still be a single keyword. Hence there's no relationship between the keyword name and how nested a map is. Namespaced keywords do follow a convention of using dots to convey hierarchy, and maybe that's one source of confusion. Dots in Clojure keywords mean nothing, except convention in certain cases. Nested maps are never accessed using dots in Clojure, we need `get`, `get-in`, etc.

So in the code we have:

```clojure
(reg-root-key-sub :camera-roll/photos :camera-roll/photos)
```

So in the app DB it's stored at the root, as `{:camera-roll/photos ...}`, not `{:camera-roll {:photos ...}}`.

In order to reduce the performance penalty I mentioned, we would need to redesign the app state to register a root subscription called `:camera-roll`, and then we would register layer-2 and layer-3 subscriptions.

I hope the explanation helps :)

src/status_im2/contexts/chat/photo_selector/view.cljs
ilmotta: You're correct about `:f>` :) You could also decouple the decision about when to fetch data by moving the dispatch to:

https://github.com/status-im/status-mobile/blob/8c1af01bf8cc15d839d7846e043e19fa1dc6e4cb/src/status_im/ui2/screens/chat/composer/view.cljs#L301 

But it's fine by me.

src/status_im2/subs/chat/chats.cljs
```diff
@@ -145,6 +145,15 @@
  (fn [[inputs public-key]]
    (get inputs (chat.models/profile-chat-topic public-key))))
 
+(re-frame/reg-sub
+ :chats/sending-image
+ :<- [:chats/current-chat-id]
+ :<- [:chat/inputs]
+ (fn [[chat-id inputs]]
+   (get-in inputs [chat-id :metadata :sending-image])))
+
```
ilmotta, line 154: Extra lines

ilmotta (approved):

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -18,52 +19,62 @@
...
   [rn/touchable-opacity
-   {:active-opacity 1
-    :on-press       (fn []
-                      (if (some #{item} @selected)
-                        (reset! selected (vec (remove #(= % item) @selected)))
-                        (swap! selected conj item)))}
+   {:active-opacity      1
+    :on-press            (fn []
+                           (if (some #{item} @selected)
+                             ;(reset! selected (vec (remove #(= % item) @selected)))
```
ilmotta: Looks like a left-over.

src/status_im2/contexts/chat/photo_selector/view.cljs
```diff
@@ -18,52 +19,62 @@
...
-            :on-press #(do
-                         (rf/dispatch [:chat.ui/clear-sending-images chat-id])
-                         (doseq [item @selected]
-                           (rf/dispatch [:chat.ui/camera-roll-pick item]))
-                         (reset! selected [])
-                         (rf/dispatch [:bottom-sheet/hide]))}
+          [quo/button
+           {:style               {:align-self        :stretch
+                                  :margin-horizontal 20}
+            :on-press            (fn []
```
ilmotta: No major concern, but I think the code is more understandable when we have single-purpose functions.

You can extract the `on-press` anonymous function. 

```clojure
(defn on-press-confirm-selection
  [chat-id]
  (rf/dispatch [:chat.ui/clear-sending-images chat-id])
  (doseq [item @selected]
    (rf/dispatch [:chat.ui/camera-roll-pick item]))
  (reset! selected [])
  (rf/dispatch [:bottom-sheet/hide]))

(defn bottom-gradient
  [chat-id selected-images]
  [:f>
   (fn []
     (let [safe-area (safe-area/use-safe-area)]
       (when (or (seq @selected) selected-images)
         [linear-gradient/linear-gradient
          {:colors [:black :transparent]
           :start  {:x 0 :y 1}
           :end    {:x 0 :y 0}
           :style  (style/gradient-container safe-area)}
          [quo/button
           {:style               {:align-self        :stretch
                                  :margin-horizontal 20}
            :on-press            #(on-press-confirm-selection chat-id)
            :accessibility-label :confirm-selection}
           (i18n/label :t/confirm-selection)]])))])
```

src/status_im2/contexts/chat/photo_selector/view.cljs
OmarBasem: Thanks @ilmotta for this information. Probably we need to discuss this in a retro maybe. I am going to try to play around with such an example and compare the renders count.

PR #14667, opened 2022-12-29
Add accessibility labels to mentions

### LINKED ISSUE #14648: No aceessibility label on mention suggest list
# Bug Report

## Problem 
No way to locate a particular element
Appium 2022-12-27 15-41-22
On  UI 1.20 it was the part of //*[@content-desc='suggestions-list']

## Acceptance criteria
accessibility id for suggestion list in input field

ibrkhalil:
fixes #14648 
status: ready


src/status_im/ui2/screens/chat/composer/mentions.cljs
```diff
@@ -11,32 +11,33 @@
 (defn mention-item
   [[public-key {:keys [alias name nickname] :as user}] _ _ text-input-ref]
   (let [ens-name? (not= alias name)]
-    [rn/touchable-opacity {:on-press #(rf/dispatch [:chat.ui/select-mention text-input-ref user])}
+    [rn/touchable-opacity
```
J-Son89, line 14: since you're touching this file.. maybe you could also migrate this to status_im2 while you're at it 🙏  :)

src/status_im/ui2/screens/chat/composer/mentions.cljs
```diff
@@ -11,32 +11,33 @@
...
+     {:on-press            #(rf/dispatch [:chat.ui/select-mention text-input-ref user])
+      :accessibility-label (str "Press to mention " (or nickname alias nickname) " in your message")}
```
J-Son89, line 16: why not just - `:mention-item`? 🤔

src/status_im/ui2/screens/chat/composer/mentions.cljs
flexsurfer: it's a bad practice actually, to make a change and migrate, just because you make a change., its always better to have separate PRs for refactoring, moving stuff etc , probably we should add this to guidelines cc @ilmotta

src/status_im/ui2/screens/chat/composer/mentions.cljs
flexsurfer: we use `accessibility-label` only for e2e, so it should be a keyword, like `:mention-item` cc @churik

flexsurfer (changes requested):

src/status_im/ui2/screens/chat/composer/mentions.cljs
ilmotta: Although I did my fare share of PRs that mixed features/chores with large scale refactors, I tend to avoid and I completely agree with @flexsurfer.

I think these sort of guidelines are not very technical, so it'd be nice to see our [CONTRIBUTING](https://github.com/status-im/status-mobile/blob/b8a84e91d4e1c2c4dda36d9e99465b8e0e76dd49/.github/CONTRIBUTING.md#contributing-to-status) doc evolve. Right now I think it's not really explaining how to contribute to Status. The current doc only has a bunch of links to the Status website.

There are many famous open-source projects we could get inspiration on on how to document our contributing guide. I think @flexsurfer wrote some pull-request guidelines some time ago, no? That could be a starting point.

src/status_im/ui2/screens/chat/composer/mentions.cljs
```diff
@@ -11,32 +11,33 @@
...
-             " "
-             (when ens-name?
-               "@")
-             name])]
+         :title             [text/text
+                             {:weight          :medium
+                              :ellipsize-mode  :tail
+                              :number-of-lines 1
+                              :size            :small}
+                             (if nickname
```
ilmotta: You can just `or` these two values.

ibrkhalil: Done

src/status_im/ui2/screens/chat/composer/mentions.cljs
ibrkhalil: Done

src/status_im/ui2/screens/chat/composer/mentions.cljs
ibrkhalil: I think we're going to have a lot of migrating/refactoring issues going forward, Which I am happy to work on 🕺🏻

J-Son89: Sure it can be separate but it would be nice to have instant follow ups for prs like this. The devs already have knowledge and context of the code they are touching so it should be faster for them

ibrkhalil: Cool, Let's merge this one and then do a follow up for migrating it to status_im2

flexsurfer (approved):

PR #14690, opened 2023-01-03
abstract running code in separate thread for native java modules

### LINKED ISSUE #14687: Abstract out `StatusThreadPoolExecutor` from StatusModule.java
# Problem
Several methods follow a similar pattern and duplicate code in `StatusModule.java` 

They all check if a certain condition is true, and if it is, they execute a task in a separate thread using the StatusThreadPoolExecutor and invoke a callback with the result.

Its better to abstract this functionality out and make a common function.
This will make the code easier to read and maintain.

siddarthkay:
This commit adds a method `executeRunnableStatusGoMethod ` which abstracts common logic out of several methods in  `StatusModule.java`

fixes #14687

#### Platforms
- Android

Any and everything where the app uses status-go

### Steps to test
please test everything

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -138,7 +138,7 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
     }
 
 
-    private void runFunctionInSeparateThread(Supplier<String[]> func, Callback callback) {
+    private void runStatusGoMethodInSeparateThread(Supplier<String[]> func, Callback callback) {
        if (!checkAvailability()) {
           callback.invoke(false);
           return;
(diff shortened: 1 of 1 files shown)


bitgamma:
while this surely makes the code more readable/shorter, I am not sure using reflection for this is a good idea. It both makes errors a little harder to catch (a typo would only be caught at runtime) and adds overhead/prevents JIT from optimizing calls (don't know how much of an impact this will have exactly). Maybe using lambda expressions could be a good compromise instead?

siddarthkay:
> while this surely makes the code more readable/shorter, I am not sure using reflection for this is a good idea. It both makes errors a little harder to catch (a typo would only be caught at runtime) and adds overhead/prevents JIT from optimizing calls (don't know how much of an impact this will have exactly). Maybe using lambda expressions could be a good compromise instead?

I wanted to use lambda as well, but for that we'll have to upgrade java version to 8..

bitgamma:
> > while this surely makes the code more readable/shorter, I am not sure using reflection for this is a good idea. It both makes errors a little harder to catch (a typo would only be caught at runtime) and adds overhead/prevents JIT from optimizing calls (don't know how much of an impact this will have exactly). Maybe using lambda expressions could be a good compromise instead?
> 
> I wanted to use lambda as well, but for that we'll have to upgrade java version to 8..

seems to be supported (https://developer.android.com/studio/write/java8-support) but otherwise there is also retrolambda

siddarthkay:
> > > while this surely makes the code more readable/shorter, I am not sure using reflection for this is a good idea. It both makes errors a little harder to catch (a typo would only be caught at runtime) and adds overhead/prevents JIT from optimizing calls (don't know how much of an impact this will have exactly). Maybe using lambda expressions could be a good compromise instead?
> > 
> > 
> > I wanted to use lambda as well, but for that we'll have to upgrade java version to 8..
> 
> seems to be supported (https://developer.android.com/studio/write/java8-support) but otherwise there is also retrolambda

yes but we're currently on java7.
so the upgrade to java8 is a challenge and once that is done we can visit lambdas for this PR

siddarthkay:
@bitgamma : I've made another issue here : https://github.com/status-im/status-mobile/issues/14692  to see what it'll take to bump java version to 8

flexsurfer (dismissed):

siddarthkay:
@bitgamma : I discovered an issue in our codebase and am implementing a fix for it so that native modules are compiled with java version 8 instead of 7.

Once that is merged in, I will refactor the current method to use lambda functions instead :)

ilmotta (approved):
Great refactoring @siddarthkay.

Just curious, why did you mention we were using Java 7? I see from `android/app/build.gradle` we're using Java 8 already.

siddarthkay:
> Great refactoring @siddarthkay.
> 
> Just curious, why did you mention we were using Java 7? I see from `android/app/build.gradle` we're using Java 8 already.

Thanks @ilmotta 
Our codebase was not compiling native modules android project with java 8 , it was only compiling the React Native App with Android 8.
issue : https://github.com/status-im/status-mobile/issues/14692#issuecomment-1369600962
fix : https://github.com/status-im/status-mobile/pull/14693

siddarthkay:
> > Great refactoring @siddarthkay.
> > Just curious, why did you mention we were using Java 7? I see from `android/app/build.gradle` we're using Java 8 already.
> 
> Thanks @ilmotta Our codebase was not compiling native modules android project with java 8 , it was only compiling the React Native App with Android 8. issue : [#14692 (comment)](https://github.com/status-im/status-mobile/issues/14692#issuecomment-1369600962) fix : #14693

Only after this fix I was able to use Lambdas here in this PR. Lambdas were introduced in java 8.

ilmotta:
> > > Great refactoring @siddarthkay.
> > > Just curious, why did you mention we were using Java 7? I see from `android/app/build.gradle` we're using Java 8 already.
> > 
> > 
> > Thanks @ilmotta Our codebase was not compiling native modules android project with java 8 , it was only compiling the React Native App with Android 8. issue : [#14692 (comment)](https://github.com/status-im/status-mobile/issues/14692#issuecomment-1369600962) fix : #14693
> 
> Only after this fix I was able to use Lambdas here in this PR. Lambdas were introduced in java 8.

I get it now, thanks for that @siddarthkay!

siddarthkay:
@status-im/mobile-qa : This is an android native module refactoring PR which will impact only android. 
Everything may work or nothing may work, Slight chance that something will break.

pavloburykh:
@siddarthkay thanx for the PR. Manually tested, no regression found. Failed e2e are not PR related. Ready for merge.

siddarthkay:
> @siddarthkay thanx for the PR. Manually tested, no regression found. Failed e2e are not PR related. Ready for merge.


Thanks @pavloburykh, rebased this PR on latest develop, will merge now :)

PR #14453, opened 2022-11-29
chore: refactor scroll page component to be more generic

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14320

# Steps to test
Open Status
Open communities tab
If you are a member of a community you can go to Discover page and then view that community

Scroll on the page :)

If you are not a member of a community you can click the '+' button on Communities home page.

then click "Communities (alpha)" option and select Status community.
Click "Follow" of the bottom of the screen.

Now go back to COmmunities Home and go to Communities Discover page. You should be able to select status community.
Click into it and start scrollling.

PR #14695, opened 2023-01-03
Scroll page animations

### LINKED ISSUE #14686: Scroll Page Component Animation
Add animation to scroll page component

OmarBasem:
Fixes: #14686

This PR is a continuation of that work by @J-Son89 https://github.com/status-im/status-mobile/pull/14453

This PR implements animations for the scroll page component used in communities and messaging.
It implements the following animations:

* Scaling animation for display picture
* Translate animation for blur view
* Opacity animation for top header picture and title

Below is a video showing the animations:

https://user-images.githubusercontent.com/29354102/210356859-3be82582-0045-44b7-9a8d-fbe2d93ed01f.mp4


J-Son89:
Looks great! :D

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -1,37 +1,24 @@
 (ns status-im2.common.scroll-page.view
   (:require [oops.core :as oops]
             [quo2.core :as quo]
+            [quo.react]
```
J-Son89: can we avoid new uses of `quo`. Please migrate the necessary files 👍

J-Son89 (approved):
Looks good but let's move animated pieces out of `quo` :) 

CC @briansztamfater as pretty sure you're also familiar with the animation code. 
Perhaps you have some thoughts where it should best live?

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+     (let [input-range         (if platform/ios? [-47 10] [0 150])
+           output-range        (if platform/ios? [-100 0] [-169 -45])
+           y                   (animated/value @scroll-height)
+           translate-animation (animated/interpolate y
+                                                     {:inputRange  input-range
+                                                      :outputRange output-range
+                                                      :extrapolate (:clamp animated/extrapolate)})
+           opacity-animation   (reanimated/use-shared-value 0)
+           threshold           (if platform/ios? 30 170)]
+       (quo.react/effect!
```
ilmotta: We have `react-native.core/use-effect` available.

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+                                                :outputRange [1.2 0.5]
+                                                :extrapolate (:clamp animated/extrapolate)})]
+           [:<>
+            [scroll-page-header scroll-height name page-nav cover sticky-header]
+            [rn/scroll-view
+             {:style                           (style/scroll-view-container
+                                                (diff-with-max-min @scroll-height 16 0))
+              :shows-vertical-scroll-indicator false
+              :scroll-event-throttle           4
+              :on-scroll                       #(swap! scroll-height
```
ilmotta: `swap!` is unnecessary here given you're not using any parameter in the anonymous `fn`.

```clojure
(fn [event]
  (reset! scroll-height (int (oops/oget event "nativeEvent.contentOffset.y"))))
```

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+                                               {:inputRange  input-range
+                                                :outputRange [1.2 0.5]
+                                                :extrapolate (:clamp animated/extrapolate)})]
+           [:<>
+            [scroll-page-header scroll-height name page-nav cover sticky-header]
+            [rn/scroll-view
+             {:style                           (style/scroll-view-container
+                                                (diff-with-max-min @scroll-height 16 0))
+              :shows-vertical-scroll-indicator false
+              :scroll-event-throttle           4
```
ilmotta: 16ms is generally low enough, and a reasonable value to help achieve 60 fps. Why 4ms?

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+           {:size   :paragraph-1
+            :weight :semi-bold
+            :style  {:line-height 21}}
+           name]]
+         [quo/page-nav
+          {:horizontal-description? true
+           :one-icon-align-left?    true
+           :align-mid?              false
+           :page-nav-color          :transparent
+           :page-nav-background-uri ""
```
ilmotta: Is this empty string for `page-nav-background-uri` really necessary? What happens if you don't pass the key at all?

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
-       (min max-image-size)))
+(defn scroll-page-header
+  [scroll-height name page-nav cover sticky-header]
+  [:f>
+   (fn []
+     (let [input-range         (if platform/ios? [-47 10] [0 150])
+           output-range        (if platform/ios? [-100 0] [-169 -45])
+           y                   (animated/value @scroll-height)
+           translate-animation (animated/interpolate y
+                                                     {:inputRange  input-range
```
ilmotta: FYI: See the implementation of `animated/interpolate`. You don't need camel case in keywords.

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
-           {:flex             1
-            :border-radius    (diff-with-max-min @scroll-height 16 0)
-            :background-color (colors/theme-colors
-                               colors/white
-                               colors/neutral-90)}
-           [children @scroll-height icon-top-fn icon-size-fn]])]])))
+      [:f>
+       (fn []
+         (let [y         (animated/value @scroll-height)
+               animation (animated/interpolate y
```
ilmotta: FYI: See the implementation of `animated/interpolate`. You don't need camel case in keywords.

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -1,37 +1,24 @@
...
+            [quo.react]
             [quo2.foundations.colors :as colors]
             [react-native.core :as rn]
             [react-native.platform :as platform]
             [reagent.core :as reagent]
             [status-im2.common.scroll-page.style :as style]
-            [utils.re-frame :as rf]))
+            [utils.re-frame :as rf]
+            [react-native.reanimated :as reanimated]
+            [quo.animated :as animated]))
```
ilmotta: I thought `quo.animated` shouldn't be used in the new `status-im2`, but I'm not sure tbh. Shouldn't we just use `react-native.reanimated`?

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+                                      (reanimated/with-timing (if (>= @scroll-height threshold) 1 0))))
+       [:<>
+        [animated/blur-view
+         {:blur-amount   32
+          :blur-type     :xlight
+          :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
+          :style         (merge
+                          {:transform [{:translateY translate-animation}]}
+                          style/blur-slider)}]
+        [rn/view {:style {:z-index 6 :margin-top (if platform/ios? 44 0)}}
```
ilmotta: You can move this to `style.cljs`.

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+           :page-nav-background-uri ""
+           :mid-section             {:type            :text-with-description
+                                     :main-text       nil
+                                     :description-img nil}
+           :right-section-buttons   (:right-section-buttons page-nav)
+           :left-section            {:icon                  :i/close
+                                     :icon-background-color (icon-color)
+                                     :on-press              #(rf/dispatch [:navigate-back])}}]
+         (when sticky-header [sticky-header @scroll-height])
+        ]]))])
```
ilmotta: Dangling stuff ;)

src/status_im2/common/scroll_page/style.cljs
```diff
@@ -3,32 +3,64 @@
...
-                   :top      (if platform/ios? 0 -64)
-                   ;; -64 is needed on android as the scroll doesn't
-                   ;; bounce so this slider won't disapear otherwise
-                   :position :absolute
-                   :height   height
-                   :width    "100%"
-                   :flex     1}})
+(def blur-slider
+  {:z-index  5
+   :top      (if platform/ios? 0 0)
```
Parveshdhull: typo or `:top 0`?

src/status_im2/common/scroll_page/view.cljs
Parveshdhull: yes, we should use reanimated.

Parveshdhull (commented):
Hi @OmarBasem,
Thank you for the great work, the animation looks cool.
Please don't mix reanimated v1 and v2 animations. (Only use v2 from reanimated.cljs)
Also, do we have animations available from the design team?

src/status_im2/common/scroll_page/view.cljs
OmarBasem: I tried using this one but it does not seem to work

src/status_im2/common/scroll_page/view.cljs
OmarBasem: Hmm, what about phones supporting 120fps? Should we use 8ms?

src/status_im2/common/scroll_page/view.cljs
OmarBasem: I think it is not necesssary. CC @J-Son89

src/status_im2/common/scroll_page/view.cljs
OmarBasem: I think we agreed 2 styles or less can be kept inline

OmarBasem:
Hi @Parveshdhull, I wasn't able to get the interpolate function in `react-native/reanimated` working. Can you please show an example?

OmarBasem:
> Also, do we have animations available from the design team?

Not sure if we have animations available from the design team. 
CC @J-Son89

src/status_im2/common/scroll_page/view.cljs
briansztamfater: We should use `reanimated/interpolate` and `reanimated/use-shared-value` if possible btw

src/status_im2/common/scroll_page/view.cljs
briansztamfater: Agree, we should just use `react-native.reanimated` namespace

src/status_im2/common/scroll_page/style.cljs
```diff
@@ -3,32 +3,64 @@
...
-                   ;; bounce so this slider won't disapear otherwise
-                   :position :absolute
-                   :height   height
-                   :width    "100%"
-                   :flex     1}})
+(def blur-slider
+  {:z-index  5
+   :top      0
+   ;; -64 is needed on android as the scroll doesn't
+   ;; bounce so this slider won't disapear otherwise
```
briansztamfater: Probably a left over? I don't see any `-64` value here

OmarBasem:
@briansztamfater do you know how to use interpolate from `react-native/reanimated`? I tried the following but did not work:

```
(defn scroll-page
  [cover page-nav name]
  (let [scroll-height (reagent/atom negative-scroll-position-0)
        input-range   (if platform/ios? [-67 10] [0 150])]
    (fn [sticky-header children]
      [:f>
       (fn []
         (let [y         (reanimated/use-shared-value @scroll-height)
               animation (reanimated/interpolate y input-range [1.2 0.5])]
           [:<>
            [scroll-page-header scroll-height name page-nav cover sticky-header]
            [rn/scroll-view
             {:style                           (style/scroll-view-container
                                                 (diff-with-max-min @scroll-height 16 0))
              :shows-vertical-scroll-indicator false
              :scroll-event-throttle           4
              :on-scroll                       #(swap! scroll-height
                                                       (fn []
                                                         (int (oops/oget % "nativeEvent.contentOffset.y"))))}
             [rn/view {:style {:height 151}}
              [rn/image
               {:source cover
                :style  {:overflow :visible
                         :flex     1}}]]
             (when children
               [rn/view
                {:flex             1
                 :border-radius    (diff-with-max-min @scroll-height 16 0)
                 :background-color (colors/theme-colors
                                     colors/white
                                     colors/neutral-90)}

                [reanimated/view
                 {:style (reanimated/apply-animations-to-style
                           {:transform        [{:scale animation}]}
                           style/display-picture-container)}
                 [rn/image
                  {:source cover
                   :style  style/display-picture}]]
                [children]])]]))])))
```

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,65 +28,98 @@
...
+          :blur-type     :xlight
+          :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
+          :style         (merge
+                          {:transform [{:translateY translate-animation}]}
+                          style/blur-slider)}]
+        [rn/view {:style {:z-index 6 :margin-top (if platform/ios? 44 0)}}
+         [reanimated/view
+          {:style (reanimated/apply-animations-to-style
+                   {:opacity opacity-animation}
+                   style/sticky-header-title)}
```
briansztamfater: I feel it would be nicer to read something like `{:style (style/sticky-header-title opacity)}`

briansztamfater:
> @briansztamfater do you know how to use interpolate from `react-native/reanimated`? I tried the following but did not work:
> 
> ```
> (defn scroll-page
>   [cover page-nav name]
>   (let [scroll-height (reagent/atom negative-scroll-position-0)
>         input-range   (if platform/ios? [-67 10] [0 150])]
>     (fn [sticky-header children]
>       [:f>
>        (fn []
>          (let [y         (reanimated/use-shared-value @scroll-height)
>                animation (reanimated/interpolate y input-range [1.2 0.5])]
>            [:<>
>             [scroll-page-header scroll-height name page-nav cover sticky-header]
>             [rn/scroll-view
>              {:style                           (style/scroll-view-container
>                                                  (diff-with-max-min @scroll-height 16 0))
>               :shows-vertical-scroll-indicator false
>               :scroll-event-throttle           4
>               :on-scroll                       #(swap! scroll-height
>                                                        (fn []
>                                                          (int (oops/oget % "nativeEvent.contentOffset.y"))))}
>              [rn/view {:style {:height 151}}
>               [rn/image
>                {:source cover
>                 :style  {:overflow :visible
>                          :flex     1}}]]
>              (when children
>                [rn/view
>                 {:flex             1
>                  :border-radius    (diff-with-max-min @scroll-height 16 0)
>                  :background-color (colors/theme-colors
>                                      colors/white
>                                      colors/neutral-90)}
> 
>                 [reanimated/view
>                  {:style (reanimated/apply-animations-to-style
>                            {:transform        [{:scale animation}]}
>                            style/display-picture-container)}
>                  [rn/image
>                   {:source cover
>                    :style  style/display-picture}]]
>                 [children]])]]))])))
> ```

@OmarBasem Usage of `reanimated/interpolate` looks fine at first glance, I don't know exactly what's not working with that code but one thing that I noted is that you are not using `reanimated/set-shared-value` (or any other reanimated method) to update `y` on `on-scroll` event. Maybe that's the reason.

OmarBasem:
@briansztamfater shouldn't the interpolate function be updating the value? In the current implementation there is no method to update `y`, interpolate function updates it

src/quo/animated.cljs
```diff
@@ -16,6 +17,7 @@
 (def scroll-view (reagent/adapt-react-class (.-ScrollView animated)))
 (def code (reagent/adapt-react-class (.-Code animated)))
 (def animated-flat-list (create-animated-component gh/flat-list-raw))
+(def blur-view (create-animated-component (.-BlurView blur)))
```
smohamedjavid: Just a question - the `blur-view` component is already available under the namespaces
1. `status-im.ui.components.react`
2. `react-native.core`

now, we are adding one more in `quo.animated`. 

Based on the [new guidelines](https://github.com/status-im/status-mobile/blob/develop/doc/new-guidelines.md#project-structure), I believe this should live under the `react-native` directory.

Which would be the correct namespace or structure/categorization for the React Native dependencies (BlurView, MaskedView, FastImage,...etc)? 



@flexsurfer @ilmotta

src/status_im2/common/scroll_page/view.cljs
ilmotta: Good question, I think testing different throttle settings in such phones would be ideal, and measure the app's performance.

But this throttle setting is just about triggering the scroll event handler, our app still needs to do a lot of processing to re-render after the scroll event happens, so I doubt we're getting 60fps anyway nowadays. That's also why we have to be careful when lowering the throttle setting, because it can actually make things worse.

src/quo/animated.cljs
OmarBasem: @smohamedjavid  That's an animated blur view. (But I should update it to be in reanimated)

React Native comps now live in react-native ns. We will eventually get rid of ui ns.

src/status_im2/common/scroll_page/view.cljs
ilmotta: In summary, it's the complete rendering path that should be measured. Lowering the throttle setting too much can put too much pressure on the rendering pipeline (re-frame, React, RN) and reduce the overall snappiness of the app.

src/status_im2/common/scroll_page/view.cljs
ilmotta: Could you share the details with us? If it doesn't work we should probably investigate why on a separate issue. It would be great if you could create a GH issue describing how to reproduce the problem.

src/status_im2/common/scroll_page/view.cljs
ilmotta: Cool, didn't know about that.

ilmotta:
>Also, do we have animations available from the design team?

@Parveshdhull Status recently hired an animation expert AFAIR, so I believe we'll soon start to have better/well defined animations. The design team probably has more details, better double-check.

briansztamfater:
> @briansztamfater shouldn't the interpolate function be updating the value? In the current implementation there is no method to update `y`, interpolate function updates it

@OmarBasem `interpolate` is calculated based on a shared-value value (in this case `y`, see first parameter of `reanimated/interpolate`), so if `y` is not updated, interpolated value won't as well.

src/quo/animated.cljs
ilmotta: I think so @smohamedjavid. This PR should use only `react-native.reanimated` and the `react-native.core/blur-view` as you said. We shouldn't rely on `quo.animated` anymore for the new app code.

Parveshdhull:
> > Also, do we have animations available from the design team?
> 
> @Parveshdhull Status recently hired an animation expert AFAIR, so I believe we'll soon start to have better/well defined animations. The design team probably has more details, better double-check.

In that case, we probably should focus more on work, that is ready. Otherwise, we will have to improve or in the worst case re-implement animations.

ilmotta:
> > > Also, do we have animations available from the design team?
> > 
> > 
> > @Parveshdhull Status recently hired an animation expert AFAIR, so I believe we'll soon start to have better/well defined animations. The design team probably has more details, better double-check.
> 
> In that case, we probably should focus more on work, that is ready. Otherwise, we will have to improve or in the worst case re-implement animations.

I agree with you @Parveshdhull. I am also working on animations for the Activity Center right now, and maybe some of it will change in the near future, but I'm only working on the very basics of animating insertion/removal of elements on screen, so even if some part of the animation changes, the groundwork will be there.

So what I'm saying is that IMO it depends on the kind of animation we need to do. If the animations are more open to interpretation from designers (such as the ones in this PR), then we have to be careful to avoid redoing work as you said. But if the animations are touching some fundamental aspects of the app behavior, perhaps it's fine to do something now.

In any case, I think anyone working on animations should double-check with designers to get their opinion, especially now that we have a dedicated person for animations.

Edit: AFAIR, the animation expert is not a developer.

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -41,54 +27,99 @@
...
+                                               [1.2 0.5]
+                                               {:extrapolateLeft  "clamp"
+                                                :extrapolateRight "clamp"})]
+       ;(rn/use-effect  #(reanimated/set-shared-value y @scroll-height)
+       ;               [@scroll-height])
+       (quo.react/effect! #(reanimated/set-shared-value y @scroll-height))
+       [reanimated/view
+        {:style (reanimated/apply-animations-to-style
+                 {:transform [{:scale animation}]}
+                 style/display-picture-container)}
```
OmarBasem: @ilmotta so here `quo.react/effect!` works but `rn/use-effect` crashes the app with error `destroy is not a function. (In 'destroy()', 'destroy' is -44)`
The weird thing is that `rn/use-effect` works in the `scroll-page-header` component above. I don't understand what's the problem here

src/status_im2/common/scroll_page/view.cljs
OmarBasem: Hmm.. I think we need to investigate more into this. Also I think we can have a condition to lower the throttle depending on the device's refresh rate

OmarBasem:
> > @briansztamfater shouldn't the interpolate function be updating the value? In the current implementation there is no method to update `y`, interpolate function updates it
> 
> @OmarBasem `interpolate` is calculated based on a shared-value value (in this case `y`, see first parameter of `reanimated/interpolate`), so if `y` is not updated, interpolated value won't as well.

Thanks @briansztamfater. Fixed it.

src/status_im2/common/scroll_page/view.cljs
ilmotta: Yep, the impact of this setting needs to be better understood, but for sure our RN app is not capable of keeping a 16ms average to render the entire stack top to bottom, or at least I highly doubt that.

src/status_im2/common/scroll_page/view.cljs
ilmotta: Ouch! I would definitely need to dig deeper to know why it fails like that. Thanks for sharing

OmarBasem: @ilmotta I figured it out. `useEffect` should return nothing or `undefined`, but for some reason, it was returning the value from the dependecy array which was `-44`. Fixed it by returning undefined from the hook function, but it would be better to update the implementation in `react-native` ns. Opened an issue: https://github.com/status-im/status-mobile/issues/14714

ilmotta: You're absolutely correct. Thank you!

src/status_im2/common/scroll_page/view.cljs
briansztamfater: I think we should vertically align map either way.

Particularly I like all styles to be in a separate file to keep some consistency. But if we will except styles with 2 props or less maybe we should update the styleguide docs so we don't create future confusion.

briansztamfater (approved):
Great work 🚀 

Consider requesting manual QA and/or designers feedback to test it on different devices.

ilmotta (approved):

ilmotta (approved):

pavloburykh:
@OmarBasem thanx for the PR! Please, take a look at following issues. 

### ISSUE 1 Animation seems not to be smooth enough (especially on IOS device)

It is hardly noticeable on recorded video but quite noticeable during live experience. It would be great if we could improve it.

During fast scrolling it is easily visible that animation is not smooth enough. 

But this is my subjective opinion, maybe other team members (especially design team) can take a look and tell if we are okay with this for now or need improvements. 

**Example recorded on iPhone X (IOS 14.1):**

https://user-images.githubusercontent.com/97245802/211003235-af9da4cc-db8c-4ab6-a681-81019ec24a62.mp4

**Example recorded on Samsung Galaxy A52 (Android 11):**

https://user-images.githubusercontent.com/97245802/211004227-88c5e1c4-07e8-470b-a2cb-9051af3f1599.mp4

OmarBasem:
Hey @pavloburykh, please let me know if it is better now

pavloburykh:
@OmarBasem thanx for your work and patience. 

We have re-checked scrolling animation on different IOS and Android devices (thanx a lot for assistance @qoqobolo @churik @VladimrLitvinenko). 

According to testing results we noticed that there is still noticeable scroll and animation twitches. We also compared the behaviour of PR and latest nightly and noticed that scroll and animation is more smooth in nightly build. 

Below is video record from my Samsung Galaxy A52. On the left part of the screen is nightly build and pr on the right:


https://user-images.githubusercontent.com/97245802/211044562-222c9778-01f2-4c29-9cbf-f22ed9f54cfe.mp4

Also, @qoqobolo has pointed out that opacity of header blur view element and community category blur element seems to be different in PR build compared to nightly were they possessed the same opacity.

On the video below you can see split screen left for nightly and right for PR build. 
As a result of comparing we see that in case of PR - channel element during scrolling "flashes" twice under the community category blur element but is slightly visible under header blur view element. In case of nightly we see that it "flashes" first during scroll under category blur element and then once it appears in transition from category to  header blur view element. The behaviour in nightly build seems to be more natural. 
The main point is that we have an assumption that these changes might be a result of the described above twitching of scroll/animation. 


https://user-images.githubusercontent.com/97245802/211050069-fc791e14-e3f5-43d0-b2c3-fa2c711e1823.mp4

src/status_im2/common/scroll_page/view.cljs
J-Son89: I think you can remove it too 👍

J-Son89:
> > Also, do we have animations available from the design team?
> 
> Not sure if we have animations available from the design team. CC @J-Son89

I was never sent a video but there is these screens in the designs -> https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1506%3A101672&t=dwxPuiEQ6Cwq2M9l-0

Alisher saw my initial implementation and approved it, so your work here is the same ( but better). If there's any questions then best to directly communicate with the design team. @Parveshdhull @OmarBasem

pedro-et:
@OmarBasem thank you for your work! I believe we need to provide you with an animation for this as there a lot of small details that need to be improved. We'll let you know! :)

J-Son89:
> @OmarBasem thank you for your work! I believe we need to provide you with an animation for this as there a lot of small details that need to be improved. We'll let you know! :)

@pedro-et  @OmarBasem - @jo-mut and I were discussing this. It seems the page style is similar in many places ->

Community Discover Page
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7545%3A487912&t=JHWSgMS8UsZSUldq-0

Community Overview Page

1-1 Chat page

Let's have a quick call to align on this as it would be great to have a single component that handles all of these cases. It will save a lot of effort and will mean adjusting animations will be easier in the future.

OmarBasem:
Thanks @pavloburykh and the QA team for your reviews! I will take a look and get back to you

J-Son89:
> > @OmarBasem thank you for your work! I believe we need to provide you with an animation for this as there a lot of small details that need to be improved. We'll let you know! :)
> 
> @pedro-et @OmarBasem - @jo-mut and I were discussing this. It seems the page style is similar in many places ->
> 
> Community Discover Page https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7545%3A487912&t=JHWSgMS8UsZSUldq-0
> 
> Community Overview Page
> 
> 1-1 Chat page
> 
> Let's have a quick call to align on this as it would be great to have a single component that handles all of these cases. It will save a lot of effort and will mean adjusting animations will be easier in the future.

also perhaps it's best to get this merged in when stable @OmarBasem and we can follow up with adjustments from @pedro-et once they have them ready? 🤔

OmarBasem:
> > > @OmarBasem thank you for your work! I believe we need to provide you with an animation for this as there a lot of small details that need to be improved. We'll let you know! :)
> > 
> > 
> > @pedro-et @OmarBasem - @jo-mut and I were discussing this. It seems the page style is similar in many places ->
> > Community Discover Page https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7545%3A487912&t=JHWSgMS8UsZSUldq-0
> > Community Overview Page
> > 1-1 Chat page
> > Let's have a quick call to align on this as it would be great to have a single component that handles all of these cases. It will save a lot of effort and will mean adjusting animations will be easier in the future.
> 
> also perhaps it's best to get this merged in when stable @OmarBasem and we can follow up with adjustments from @pedro-et once they have them ready? 🤔

Yes I will wait for the animations input from @pedro-et

OmarBasem:
@J-Son89 I may have misunderstood you, what do you mean by "when stable"?

J-Son89:
> @J-Son89 I may have misunderstood you, what do you mean "when stable"?

when it's ready to merge from a dev and QA perspective. 👍

ibrkhalil (approved):

J-Son89:
@pavloburykh & @qoqobolo could we consider allowing this pr to be merged more or less as is and handle the scrolling smoothness in a follow pr/issue.
This pr achieves a lot and will run into code clashes delaying more work if we keep it blocked. Other devs are reusing this component in other parts of the app and so they will be unblocked too if we have this pr merged.
e.g https://github.com/status-im/status-mobile/pull/14825

OmarBasem:
> @pavloburykh & @qoqobolo could we consider allowing this pr to be merged more or less as is and handle the scrolling smoothness in a follow pr/issue. This pr achieves a lot and will run into code clashes delaying more work if we keep it blocked. Other devs are reusing this component in other parts of the app and so they will be unblocked too if we have this pr merged. e.g #14825

@pavloburykh I am currently busy with other tasks and won't be able to do the QA reviews very soon. The animation's smoothness problem needs a deeper dig. In order to not block other people's work, I have opened an issue https://github.com/status-im/status-mobile/issues/14832, and we can get back to it.

qoqobolo:
> @pavloburykh & @qoqobolo could we consider allowing this pr to be merged more or less as is and handle the scrolling smoothness in a follow pr/issue.
This pr achieves a lot and will run into code clashes delaying more work if we keep it blocked. Other devs are reusing this component in other parts of the app and so they will be unblocked too if we have this pr merged.
e.g https://github.com/status-im/status-mobile/pull/14825

Sure, we can merge it in this case.

@OmarBasem thanks for creating a follow-up issue and for your work!

Issue #14668, opened 2022-12-29
Switcher buttons missing in chat view


du82:
The switcher buttons are missing now

![Screenshot_2022-12-29-09-54-39-13_435383e61849826ed84c38f611283930](https://user-images.githubusercontent.com/65834069/209977861-505d48cd-7d32-4454-85d7-eebe1b9057c5.jpg)

qoqobolo:
Designs:
1-1 chat: https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=332%3A78250&t=bdJR9zuV7v2na9hC-4
Community channel: https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1558%3A118695&t=BJfJ2aZ5rDjYcfH8-4

@du64 thanks for reporting this!

PR #14701, opened 2023-01-03
Persist switcher cards

Parveshdhull:
status-go PR: https://github.com/status-im/status-go/pull/3049

### Testing:
PR handles storing the card when opening chat, community, etc. And deleting the card by pressing the close button.
Removing the card when deleting chat, blocking user, etc. will be implemented later. 

status : ready

### DIFF
--- a/src/status_im/data_store/switcher_cards.cljs
+++ b/src/status_im/data_store/switcher_cards.cljs
@@ -1,0 +1,41 @@
+(ns status-im.data-store.switcher-cards
+  (:require [clojure.set :as set]
+            [clojure.walk :as walk]
+            [utils.re-frame :as rf]
+            [taoensso.timbre :as log]))
+
+(defn <-rpc
+  [switcher-cards]
+  (walk/postwalk-replace
+   {:cardId   :card-id
+    :screenId :screen-id}
+   switcher-cards))
+
+(defn rpc->
+  [switcher-card]
+  (set/rename-keys switcher-card
+                   {:card-id   :cardId
+                    :screen-id :screenId}))
+
+(rf/defn upsert-switcher-card-rpc
+  [_ switcher-card]
+  {:json-rpc/call [{:method     "wakuext_upsertSwitcherCard"
+                    :params     [(rpc-> switcher-card)]
+                    :on-success #()
+                    :on-error   #()}]})
+
+(rf/defn delete-switcher-card-rpc
+  [_ card-id]
+  {:json-rpc/call [{:method     "wakuext_deleteSwitcherCard"
+                    :params     [card-id]
+                    :on-success #()
+                    :on-error   #()}]})
+
+(rf/defn fetch-switcher-cards-rpc
+  [_]
+  {:json-rpc/call [{:method     "wakuext_switcherCards"
+                    :params     []
+                    :on-success #(rf/dispatch
+                                  [:shell/switcher-cards-loaded
+                                   (:switcherCards ^js %)])
+                    :on-error   #(log/error "Failed to fetch switcher cards" %)}]})
--- a/src/status_im2/contexts/shell/events.cljs
+++ b/src/status_im2/contexts/shell/events.cljs
@@ -1,10 +1,12 @@
 (ns status-im2.contexts.shell.events
   (:require [utils.re-frame :as rf]
             [re-frame.core :as re-frame]
+            [status-im.utils.core :as utils]
             [status-im2.common.constants :as constants]
             [status-im2.navigation.events :as navigation]
             [status-im2.contexts.shell.animation :as animation]
-            [status-im2.contexts.shell.constants :as shell.constants]))
+            [status-im2.contexts.shell.constants :as shell.constants]
+            [status-im.data-store.switcher-cards :as switcher-cards-store]))
 
 ;; Effects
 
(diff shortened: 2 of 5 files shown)


src/status_im2/contexts/shell/events.cljs
```diff
@@ -29,59 +29,78 @@
...
-                                        {:type       shell.constants/community-channel-card
-                                         :id         (:community-id chat)
-                                         :clock      now
-                                         :channel-id (:chat-id chat)})}
+        {:navigate-from :communities-stack
+         :card-id       (:community-id chat)
+         :card-params   {:type  shell.constants/community-channel-card
+                         :id    (:community-id chat)
+                         :clock now
+                         :extra (:chat-id chat)}}
```
Parveshdhull: extra field: https://github.com/status-im/status-go/pull/3049#discussion_r1062277109

src/status_im/data_store/switcher_cards.cljs
```diff
@@ -0,0 +1,12 @@
+(ns status-im.data-store.switcher-cards
+  (:require [utils.re-frame :as rf]
+            [taoensso.timbre :as log]))
+
+(rf/defn fetch-switcher-cards-rpc
+  [_]
+  {:json-rpc/call [{:method     "wakuext_switcherCards"
```
flexsurfer, line 36: just wondering why this one in a separate ns ?

src/status_im2/contexts/shell/events.cljs
ilmotta: FYI: @Parveshdhull explained some fields will be renamed in status-go, so my comment is irrelevant now.

Outdated:

As I mentioned over there in the status-go PR, maybe renaming the key from `:extra` to `:metadata` could be more clear, because metadata is a word that usually implies some structure.

I would prefer if we could persist stuff in this `extra/metadata` field as serialized JSON. That way we always know what's stored from inspecting the DB, otherwise the value is too opaque. Of course, that means you'll need to call `JSON.parse` to deserialize, but that's trivial and worth it IMO.

src/status_im2/contexts/shell/events.cljs
```diff
@@ -29,59 +29,78 @@
...
 
-(rf/defn add-switcher-card
-  {:events [:shell/add-switcher-card]}
-  [{:keys [db now] :as cofx} view-id id]
+(rf/defn switcher-cards-loaded
+  {:events [:shell/switcher-cards-loaded]}
+  [{:keys [db]} loaded-switcher-cards]
+  {:db (assoc db
+              :shell/switcher-cards
+              (reduce (fn [acc switcher-card]
```
ilmotta: You can replace the entire reduce processing with `status-im.utils.core/index-by`.

src/status_im2/contexts/shell/events.cljs
Parveshdhull: resolved in status-go. Thank you @ilmotta

src/status_im/data_store/switcher_cards.cljs
Parveshdhull: fixed, thank you

src/status_im2/contexts/shell/events.cljs
Parveshdhull: now, also renaming keys. So index-by will not be sufficient

ilmotta: @Parveshdhull, you can still use `index-by` because the collection can be transformed outside the reducing function.

```clojure
(->> loaded-switcher-cards
     (map switcher-cards-store/<-rpc)
     (utils.core/index-by :card-id))
```

src/status_im/data_store/switcher_cards.cljs
```diff
@@ -0,0 +1,39 @@
+(ns status-im.data-store.switcher-cards
+  (:require [utils.re-frame :as rf]
+            [taoensso.timbre :as log]
+            [clojure.set :as clojure.set]))
+
+(defn <-rpc
+  [switcher-card]
+  (clojure.set/rename-keys switcher-card
```
ilmotta: I believe we're following the Clojure Style Guide for certain namespace aliases. See the table here https://guide.clojure.style/#use-idiomatic-namespace-aliases

`clojure.set` should be aliased as `set`

src/status_im2/contexts/shell/events.cljs
```diff
@@ -29,59 +30,80 @@
...
+       (switcher-cards-store/upsert-switcher-card-rpc switcher-card)))))
+
 (rf/defn close-switcher-card
   {:events [:shell/close-switcher-card]}
-  [{:keys [db]} id]
-  {:db (update-in db [:shell/switcher-cards] dissoc id)})
+  [{:keys [db] :as cofx} card-id]
+  (rf/merge
+   cofx
+   {:db (update-in db [:shell/switcher-cards] dissoc card-id)}
```
ilmotta: Just `update` is sufficient instead of `update-in`.

ilmotta (approved):

siddarthkay (approved):

src/status_im2/contexts/shell/events.cljs
Parveshdhull: hi, thank you for suggestion, but only update is not working

src/status_im2/contexts/shell/events.cljs
Parveshdhull: done using `clojure.walk`, thank you

src/status_im2/contexts/shell/events.cljs
ilmotta: `update` without the square brackets around the key should work.

`(update db :shell/switcher-cards dissoc card-id)`

```clojure
(let [card-id "0x1"
      db      {:shell/switcher-cards {card-id {:card-data :anything}}}]
  (assert (= {:shell/switcher-cards {}}
             (update db :shell/switcher-cards dissoc card-id))))
```

Parveshdhull: oh, missed that, working now. Thank you

src/status_im/data_store/switcher_cards.cljs
```diff
@@ -0,0 +1,41 @@
...
+    :screenId :screen-id}
+   switcher-cards))
+
+(defn rpc->
+  [switcher-card]
+  (set/rename-keys switcher-card
+                   {:card-id   :cardId
+                    :screen-id :screenId}))
+
+(rf/defn upsert-switcher-card-rpc
```
ilmotta, line 20: Just a quick thing I didn't realize before @Parveshdhull. I had the impression the data store directory should know nothing about re-frame. This layer should only care about the domain model, often using only pure functions.

Event handlers should be added to another layer of the stack IMO, in `events.cljs` files somewhere. I think this is rather important.

ilmotta: I see there are other namespaces in `data_store/` defining event handlers, but this is a mistake IMO, especially now that we have new guidelines.

So event handlers should be added to `events.cljs` files in their corresponding `status-im2/contexts/*` directory, or to `status-im2/common` if they're really general.

qoqobolo:
Hey @Parveshdhull, thanx for PR, works great!

Two things here:

1. Q: Is it expected, that a card is not persisted on the first opening of a chat?

I.e. if you create a new group chat or a community and go back to the home screen and then press `Jump to`, this card won't be there. For the chat card to appear on the screen, you need to open that chat again from the chat screen.
It behaves this way in develop as well, I'm just wondering if we want to fix it.

https://user-images.githubusercontent.com/67952253/210816634-9d3a833a-9930-42c0-88ad-6a73b4ca242a.mp4

(asked the same question [here ](https://github.com/status-im/status-mobile/pull/14717#issuecomment-1372360608), please answer wherever you want :))

2. In this PR I can't test navigation from the chat view due to this issue https://github.com/status-im/status-mobile/issues/14668.

But it has already been fixed in #14717.
I don't want to block this PR due to 14668, so actually it can be merged.
Or we can wait until 14717 is merges, then you rebase your PR and I can test that flow.
WDYT?

Parveshdhull:
Hi @qoqobolo,
Thank you very much for testing the PR.

Actually, adding of switcher card is not handled by the jump-to button. Jump-to just closes the corresponding screen, so that we can see the shell screen and access cards.

Adding cards is handled by navigation, and currently, only the home screen is using navigation2 calls. We have to use it at several other places like communities list (as shown in the video), or new 1-1 to chat or starting a chat from a contact. I will work on that in a separate PR. 

Also, about Q2, we can skip that, as I mentioned jump-to doesn't have anything to do with cards.

Will that be ok?

qoqobolo:
@Parveshdhull perfectly clear now, thank you!

>Also, about Q2, we can skip that, as I mentioned jump-to doesn't have anything to do with cards.
Will that be ok?

Yep, PR can be merged.

flexsurfer (approved):

PR #14722, opened 2023-01-06
[WIP] adding errorCallback to native modules for errors throw on status-go side

### LINKED ISSUE #14720: Add support for error messages as a separate on-error callback in native modules
# Problem
For example lets consider a call to this native module 
`(status/compressed-key->public-key "good-key-format" #(println "key:" %))`

we get output back in the callback method which prints the key.
In the event our code has an input error, the code return such an error message via the same print statement
`{"error":"unsupported public key type 'xyz'"}`

To know whether there is an error or not we will have to parse the data returned from the callback.
it would be better error handling if native modules followed the approach of an on-success callback and an on-error callback.

The on-success callback would return data on success and on-error callback would return the error message on error.

siddarthkay:
fixes #14720

### Review notes
This is probably the only way to handle an error call back.
Status-go returns everything as 1 string.. I would not want to change that since Desktop team also relies on status-go.

#### Platforms
- Android
- iOS

Any and everything where the app uses status-go

### Steps to test
Please do a full test

status: ready

### DIFF
--- a/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
+++ b/modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
@@ -531,33 +531,37 @@ class StatusModule extends ReactContextBaseJavaModule implements LifecycleEventL
         StatusThreadPoolExecutor.getInstance().execute(r);
     }
 
-    private void executeRunnableStatusGoMethod(Supplier<String> method, Callback callback) throws JSONException {
+    private void executeRunnableStatusGoMethod(Supplier<String> method, Callback successCallback,Callback errorCallback) throws JSONException {
         if (!checkAvailability()) {
-            callback.invoke(false);
+            successCallback.invoke(false);
             return;
         }
 
         Runnable runnableTask = () -> {
             String res = method.get();
-            callback.invoke(res);
+             if (res.contains("error")) {
+                errorCallback.invoke(res);
+             } else {
+                successCallback.invoke(res);
+             }
         };
 
         StatusThreadPoolExecutor.getInstance().execute(runnableTask);
     }
 
     @ReactMethod
-    public void verify(final String address, final String password, final Callback callback) throws JSONException {
+    public void verify(final String address, final String password, final Callback successCallback, final Callback errorCallback) throws JSONException {
         Activity currentActivity = getCurrentActivity();
 
         final String absRootDirPath = this.getNoBackupDirectory();
         final String newKeystoreDir = pathCombine(absRootDirPath, "keystore");
 
-        executeRunnableStatusGoMethod(() -> Statusgo.verifyAccountPassword(newKeystoreDir, address, password), callback);
+        executeRunnableStatusGoMethod(() -> Statusgo.verifyAccountPassword(newKeystoreDir, address, password), successCallback, errorCallback);
     }
 
     @ReactMethod
-    public void verifyDatabasePassword(final String keyUID, final String password, final Callback callback) throws JSONException {
-        executeRunnableStatusGoMethod(() -> Statusgo.verifyDatabasePassword(keyUID, password), callback);
+    public void verifyDatabasePassword(final String keyUID, final String password, final Callback successCallback, final Callback errorCallback) throws JSONException {
+        executeRunnableStatusGoMethod(() -> Statusgo.verifyDatabasePassword(keyUID, password), successCallback, errorCallback);
     }
 
     public String getKeyStorePath(String keyUID) {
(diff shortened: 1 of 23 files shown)


modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -531,33 +531,37 @@ public void run() {
...
         if (!checkAvailability()) {
-            callback.invoke(false);
+            successCallback.invoke(false);
             return;
         }
 
         Runnable runnableTask = () -> {
             String res = method.get();
-            callback.invoke(res);
+             if (res.contains("error")) {
```
cammellos: This is very dangerous, what if you receive a message with `error` in the content? we need to parse it as JSON (could be optimized by scanning the string, but don't think it's worth the risk)

src/status_im/browser/core.cljs
```diff
@@ -534,7 +537,10 @@
         (do
           (log/warn :web3-response-error)
           (callback "web3-response-error" nil))
-        (callback nil (.parse js/JSON response)))))))
+        (callback nil (.parse js/JSON response))))
+    (fn [error-message]
+      (log/debug "error while status/call-rpc" error-message))
+   )))
```
cammellos: linting

src/status_im/keycard/real_keycard.cljs
```diff
@@ -321,7 +321,10 @@
      (let [{:keys [error]} (types/json->clj result)]
        (if-not (string/blank? error)
```
cammellos, line 322: if we start using the callbacks, we should not handle errors though on the on-success, or that's still an eventuality? Similarly in other parts of the code

cammellos (changes requested):
I think we need to be more careful in detecting an error, this will give false positives

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
siddarthkay: hmmm, every method in status-go uses `makeJSONResponse ` method 
which will add the error key to the response string :  reference https://github.com/status-im/status-go/blob/ec7c0e9c7d34f73d89d8fb4be0e116aa29c3a29d/mobile/status.go#L625

src/status_im/keycard/real_keycard.cljs
siddarthkay: yes this will definitely need some more refactoring :) 
But since i did not write this functionality , I was hesitant on modifying it.. I will look at this one more time :)

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
siddarthkay: > This is very dangerous, what if you receive a message with `error` in the content? we need to parse it as JSON (could be optimized by scanning the string, but don't think it's worth the risk)

Hmmm I see your point

siddarthkay: I've done this on Android native module, I will work on the IOS side as well! Thanks for the suggestion @cammellos

siddarthkay: > This is very dangerous, what if you receive a message with `error` in the content? we need to parse it as JSON (could be optimized by scanning the string, but don't think it's worth the risk)

For ios side I converted the response to a dictionary and used a similar approach here, this should ensure we look for the error key and we do not trigger false positives by "error" in message

src/status_im/keycard/simulated_keycard.cljs
```diff
@@ -402,28 +407,29 @@
...
+                   key-uid
+                   [path]
+                   hashed-password
+                   (fn [result]
+                     (let [{:keys [publicKey]} (get result (keyword path))]
+                       (on-success publicKey)))
+                   (fn [error-message]
+                     (on-failure error-message))))))
+            (fn [error-message]
+              (log/debug "error while status/multiaccount-derive-addresses" error-message)))))
```
ilmotta: @siddarthkay In this PR I see a lot of `log/debug` in *error callback* functions. They seem to swallow the error, because they don't rethrow it. Are the errors really propagating up the call tree? If not, then I think this is a more serious problem.

If they propagate correctly, then it shouldn't be necessary to add intermediary calls to `log/debug`, only a top level error handler would be enough. In Java, this would be equivalent of doing a try/catch in intermediary methods that don't actually do anything with the error (except logging). This is a known anti-pattern.

It's probably out of scope for your PR, but these nested callbacks for success/error could be dramatically improved in the future if we used the de facto standard in Javascript (promises), or some other means in Clojure land. This would give us a sane way to handle errors and create composable error handlers.

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
```diff
@@ -531,33 +531,43 @@ public void run() {
         StatusThreadPoolExecutor.getInstance().execute(r);
     }
 
-    private void executeRunnableStatusGoMethod(Supplier<String> method, Callback callback) throws JSONException {
+    private void executeRunnableStatusGoMethod(Supplier<String> method, Callback successCallback,Callback errorCallback) throws JSONException {
         if (!checkAvailability()) {
-            callback.invoke(false);
+            successCallback.invoke(false);
```
qfrank, line 536: it makes me a little confused here, if we can't execute `method`, why not use `errorCallback` instead but `successCallback`?

src/status_im/browser/core.cljs
```diff
@@ -418,7 +418,10 @@
                                                       [:browser.callback/call-rpc
                                                        {:type      constants/web3-send-async-callback
                                                         :messageId message-id
-                                                        :result    (types/json->clj %)}])}}
+                                                        :result    (types/json->clj %)}])
+                                      :on-error     (fn [error-message]
+                                                      (log/debug "error while status/recover-message"
```
qfrank, line 423: if it's an error, i prefer to use log/warn or log/error rather than log/debug

src/status_im/multiaccounts/create/core.cljs
```diff
@@ -40,7 +41,7 @@
 
 (re-frame/reg-fx
  ::store-multiaccount
- (fn [[id key-uid hashed-password callback]]
+ (fn [[id key-uid hashed-password callback on-error]]
```
qfrank, line 44: can we rename callback to on-success so that we can c the `good pair` on-success/on-error? 😅

src/status_im/signing/core.cljs
```diff
@@ -25,10 +25,11 @@
 
 (re-frame/reg-fx
  :signing/send-transaction-fx
- (fn [{:keys [tx-obj hashed-password cb]}]
+ (fn [{:keys [tx-obj hashed-password cb on-error]}]
    (status/send-transaction (types/clj->json tx-obj)
                             hashed-password
-                            cb)))
+                            cb
```
qfrank, line 31: can we also rename `cb` here ?

siddarthkay: > can we also rename `cb` here ?

Well it was always `cb` from the beginning, but sure why not, I'll improve this :)

modules/react-native-status/android/src/main/java/im/status/ethereum/module/StatusModule.java
siddarthkay: > it makes me a little confused here, if we can't execute `method`, why not use `errorCallback` instead but `successCallback`?

Well I need to do some more research on this but what this basically means is that when we only had a single callback we used to do the callback anyways with a false return whenver `checkAvailability` would return `false` .
But I could definitely make this better.

thanks for pointing it out @qfrank  :)

src/status_im/keycard/simulated_keycard.cljs
siddarthkay: > dramatically improved in the future if we used the de facto standard in Javascript (promises)

I did look into the native modules side on React Native and it does support Promises and would make the code a lot simpler but would require a lot of re-write ( same as this PR 🫣 ).

I would for now log a new issue to address this concern later.

siddarthkay: > Are the errors really propagating up the call tree? If not, then I think this is a more serious problem.

yes they do and that was the inspiration to introduce error-callbacks to avoid handling of both success and error cases in one callback.

siddarthkay:
Started working on this again

src/status_im/keycard/simulated_keycard.cljs
ilmotta: > I would for now log a new issue to address this concern later.

Thanks, better in a separate issue as you said.

src/status_im/signing/core.cljs
ilmotta: On the Clojure side, more specifically re-frame events, it's preferred to use more explicit names, like `on-error`, `on-success`, particularly because re-frame uses this convention throughout its documentation. There are also external effects, like (re-frame-http-fx) who follow this convention (although they use `on-failure` instead of `on-error`). Additionally, pretty much all effects in Status codebase use this naming convention (`:rpc/call` being the most famous).

I think we should keep the consistency and use `on-error` and `on-success` as the names of such arguments.

flexsurfer:
hey @siddarthkay still relevant?

siddarthkay:
@flexsurfer : yes it has some work in progress which i plan to merge this week

siddarthkay:
I'll convert to Draft

flexsurfer:
feel free to reopen when work in progress again

PR #14725, opened 2023-01-08
[14703] Remove using function as children warning from android text input

### LINKED ISSUE #14703: TextInput with children warning
```
[Wed Jan 04 2023 09:58:18.141]  ERROR    Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.
    in RCTMultilineTextInputView (at TextInput.js:1077)
    in TouchableWithoutFeedback (at TextInput.js:1134)
    in InternalTextInput (at TextInput.js:1159)
    in ForwardRef(TextInput) (created by status_im.ui2.screens.chat.composer.input.text_input)
    in status_im.ui2.screens.chat.composer.input.text_input (created by f)
```

in `status-im.ui2.screens.chat.composer.input` 

`[rn/text-input props children]`

where `children` is a function, which is wrong? it should be text component, but if we implement it properly , mentions works wrong, so we need to fix this

ibrkhalil:
fixes #14703
status: ready

### DIFF
--- a/src/status_im/ui2/screens/chat/composer/input.cljs
+++ b/src/status_im/ui2/screens/chat/composer/input.cljs
@@ -175,7 +175,6 @@
          :editable                 (not cooldown-enabled?)
          :blur-on-submit           false
          :auto-focus               false
-         :default-value            initial-value
          :on-focus                 #(set-active-panel nil)
          :max-length               chat.constants/max-text-size
          :placeholder-text-color   (:text-02 @quo.colors/theme)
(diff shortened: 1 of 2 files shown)


src/status_im/ui2/screens/chat/composer/input.cljs
```diff
@@ -210,7 +210,7 @@
     (if platform/android?
       [selectable-text-input chat-id props children]
       [rn/text-input props
-       children])))
+       [children]])))
```
flexsurfer: unfortunately, it's not that simple, there is differently something wrong, it doesn't play well with mentions, also we shouldn't use default-value, because we set default value as children

ibrkhalil: I am currently testing it on iOS

flexsurfer: yes, i tested on ios as well

flexsurfer: for some reason it starts to duplicate text, when you typing or when you select mention

ibrkhalil: Kindly take another look, I checked on iOS and tested the selectable text input changes on Android

ibrkhalil: And the selectable text input now has the repeating issue, Looking ..

ibrkhalil: You're right, This issue isn't as easy as I expected .. I'll close it for now and work on it later

PR #14738, opened 2023-01-09
fix default shell for -include targets

### LINKED ISSUE #14732: Make (version 4.4-1) command is not creating a temp directory
# Problem

If I checkout a new branch and run `make run-clojure` then this error is thrown:

```
[status-mobile]-> make run-clojure                                                            
yarn shadow-cljs watch mobile                                                                 
Configuring Nix shell for target 'clojure'...                                                 
error: getting status of '/tmp/tmp-status-mobile-1cdcd298b': No such file or directory        
make: *** [Makefile:260: run-clojure] Error 1
```

Apparently `_tmpdir-mk` target is not being called. We have to call it manually `make _tmpdir-mk`

(Creating issue for context, if anyone else gets this error)

Parveshdhull:
fixes: https://github.com/status-im/status-mobile/issues/14732

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: nix-add-gcroots clean nix-clean run-metro test release _list _fix-node-perms _tmpdir-mk _tmpdir-rm _install-hooks
+.PHONY: nix-add-gcroots clean nix-clean run-metro test release _list _fix-node-perms _tmpdir-rm
 
 help: SHELL := /bin/sh
 help: ##@other Show this help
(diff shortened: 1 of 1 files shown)


jakubgs (approved):
I don't know WHY the removal from `.PHONY` section fixes this, but it appears it does.

This means that `make` will now check for existence of `_tmpdir-mk` file at every run, but I guess that is a tiny cost to bear. Ultimately the way we use `-include` is some kind of hack, since the actual purpose of `include` is to load extra `make` files, but we are not doing that in this case.

I guess this is fine for now. But this use of `include` might break in the future.

Parveshdhull:
> But this use of include might break in the future.

I guess a more robust (but ugly) solution will be to create a default target and use that as a dependency for all other targets.

PR #14680, opened 2022-12-30
Reset bottom sheet on subsequent opens

ibrkhalil:
fixes #14671, #14677, #14672, #14664, #14681

### Summary
The new bottom sheet component calculates content height on launch so it needs to be completely reset on subsequent launches to re-initiate the height calculation process
And the bottom-sheet/hide subscription should be replaced with bottom-sheet/close-bottom-sheet-fn

status: ready

PR #14739, opened 2023-01-10
New chat fix

OmarBasem:
Fixes app crash when pressing new chat and bottom sheet not showing properly

### QA review notes:
1. Currently the bottom button is not positioned properly (overlapping with safe area) as we are currently using an old bottom sheet.
2. There is a 1 second delay when pressing "New chat". That's because it is trying to open a bottom sheet from inside a bottom sheet. We would have to migrate it to a screen component. Opened an issue https://github.com/status-im/status-mobile/issues/14741

### DIFF
--- a/src/status_im/ui/screens/home/sheet/views.cljs
+++ b/src/status_im/ui/screens/home/sheet/views.cljs
@@ -73,11 +73,13 @@
      :icon-color                 (colors/theme-colors colors/neutral-50 colors/neutral-40)
      :accessibility-label        :start-a-new-chat
      :icon                       :i/new-message
-     :on-press                   (fn [] ;; Note: currently the bottom button is overlapping with safe-area
+     :on-press                   (fn [] ;; Note: currently the bottom button is overlapping with
+                                        ;; safe-area
                                         ;; because it is using old bottom-sheet
                                    (rf/dispatch [:bottom-sheet/hide])
                                    ;; A one second delay is needed because there are 2 bottom-sheets here
-                                   ;; It would better to migrate contact-selection-list component to use a screen
+                                   ;; It would better to migrate contact-selection-list component to use
+                                   ;; a screen
                                    (js/setTimeout #(rf/dispatch [:bottom-sheet/show-sheet
                                                                  {:content
                                                                   new-chat-aio/contact-selection-list}])
--- a/src/status_im/ui/screens/home/sheet/views.cljs
+++ b/src/status_im/ui/screens/home/sheet/views.cljs
@@ -73,9 +73,11 @@
      :icon-color                 (colors/theme-colors colors/neutral-50 colors/neutral-40)
      :accessibility-label        :start-a-new-chat
      :icon                       :i/new-message
-     :on-press                   (fn [] ; Note: currently the bottom button is overlapping with safe-area
-                                        ; because it is using old bottom-sheet
+     :on-press                   (fn [] ;; Note: currently the bottom button is overlapping with safe-area
+                                        ;; because it is using old bottom-sheet
                                    (rf/dispatch [:bottom-sheet/hide])
+                                   ;; A one second delay is needed because there are 2 bottom-sheets here
+                                   ;; It would better to migrate contact-selection-list component to use a screen
                                    (js/setTimeout #(rf/dispatch [:bottom-sheet/show-sheet
                                                                  {:content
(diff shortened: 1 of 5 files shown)


cammellos:
@OmarBasem make sure once approved QA takes a look at this one before merging, thanks for the fix!

src/status_im/ui/screens/home/sheet/views.cljs
```diff
@@ -73,8 +73,18 @@
      :icon-color                 (colors/theme-colors colors/neutral-50 colors/neutral-40)
      :accessibility-label        :start-a-new-chat
      :icon                       :i/new-message
-     :on-press                   #(hide-sheet-and-dispatch [:bottom-sheet/show-sheet
-                                                            :start-a-new-chat])}]
+     :on-press                   (fn [] ;; Note: currently the bottom button is overlapping with
```
J-Son89, line 76: can we create an issue to properly address this? This seems like a temporary fix to me.

J-Son89 (approved):

src/status_im/ui/screens/home/sheet/views.cljs
OmarBasem: Yes it is temporary. WIll create an issue.

src/status_im2/subs/contact.cljs
```diff
@@ -93,7 +93,7 @@
        vals)))
 
 (re-frame/reg-sub
- :contacts/add-members-sections
+ :contacts/sorted-and-grouped-by-first-letter
```
ibrkhalil: Please rename this sub, As I am adding the deleted sub of `:sorted-and-grouped-by-first-letter` that you accidentally deleted again [here](https://github.com/status-im/status-mobile/pull/14680)

ibrkhalil (changes requested):

qoqobolo:
Hi @OmarBasem, thanks for the fix!

Q: There are several issues related to the `Start new chat` flow itself. For example, incorrect work of checkboxes on the `New chat` bottom sheet or the ability to start a chat with a user who hasn't yet approved your contact request.
Should I report such issues here, or is that out of the scope of this PR?

ibrkhalil:
> Hi @OmarBasem, thanks for the fix!
> 
> Q: There are several issues related to the `Start new chat` flow itself. For example, incorrect work of checkboxes on the `New chat` bottom sheet or the ability to start a chat with a user who hasn't yet approved your contact request. Should I report such issues here, or is that out of the scope of this PR?

Most of them are fixed [here](https://github.com/status-im/status-mobile/pull/14680)

qoqobolo:
> Most of them are fixed https://github.com/status-im/status-mobile/pull/14680

Ah, cool! Then it would make sense now to merge the fix only for the crash (after addressing your comment above) and fix all the remaining bugs (if any) separately.

ibrkhalil:
> > Most of them are fixed #14680
> 
> Ah, cool! Then it would make sense now to merge the fix only for the crash (after addressing your comment above) and fix all the remaining bugs (if any) separately.

Yes, But this fix is going to break my PR, That's why I suggested a new subscription name .. Because the one I created uses this same name

qoqobolo:
@ibrkhalil yeah, let's wait for @OmarBasem's answer. 
After changing the subscription name I'll re-check the PR again briefly.

src/status_im2/subs/contact.cljs
OmarBasem: Fixed

OmarBasem:
> Hi @OmarBasem, thanks for the fix!
> 
> Q: There are several issues related to the `Start new chat` flow itself. For example, incorrect work of checkboxes on the `New chat` bottom sheet or the ability to start a chat with a user who hasn't yet approved your contact request. Should I report such issues here, or is that out of the scope of this PR?

As @ibrkhalil has mentioned, they should be fixed in his PR

OmarBasem:
Hey @qoqobolo, please let me know if all is okay now.

FYI, this PR fixes the crash that is happening when clicking on "new chat", and makes the bottom sheet properly visible (previously is was squashed to the bottom)

qoqobolo:
> Hey @qoqobolo, please let me know if all is okay now.

Yep, LGTM, thank you! PR can be merged.

ibrkhalil (approved):

OmarBasem:
@qoqobolo thank you!

Issue #14752, opened 2023-01-10
Android screen flickers while navigating


Parveshdhull:
# Bug Report

Android screen flickers while navigating

https://user-images.githubusercontent.com/17097240/211683411-de884c4a-2b71-42da-a54c-786bb1b0e9ec.mp4

### Cause
Layout animations are disabled by default in current version (2.3.3), and on [enabling](https://github.com/status-im/status-mobile/blob/9ed89ac97dc5a37dfabf1c8a232d45cf730fcdd0/src/react_native/reanimated.cljs#L20) them this flicker happens ([reference](https://github.com/status-im/status-mobile/issues/13452#issuecomment-1159800179)).
And even if we don't enable them by default, and we use any layout animation from reanimated then they [will be enabled](https://github.com/software-mansion/react-native-reanimated/pull/2651#issue-1059867386) at runtime. This enabling of layout animation at runtime also breaks layout, as mentioned in [comment](https://github.com/status-im/status-mobile/pull/14618#issuecomment-1371071337).

So, for fixing this issue either we can try to upgrade reanimated
(not sure if that will fix it, but still upgrade has its own [problems](https://github.com/status-im/status-mobile/issues/13452#issuecomment-1159800641)),
or for now we can avoid using layout animations. cc @yqrashawn

### Reproduction

1) Open community chat
2) Restart app
3) Press back button in community chat
(@status-im/mobile-qa Please can you verify issue. Compare current nightly with version before https://github.com/status-im/status-mobile/pull/14618 for verification)

yqrashawn:
I'll disable the layout animation first and try upgrade reanimated later.

qoqobolo:
> https://github.com/orgs/status-im/teams/mobile-qa Please can you verify issue. Compare current nightly with version before https://github.com/status-im/status-mobile/pull/14618 for verification

Hi @Parveshdhull, thanks for reporting this. Confirming that the issue is indeed reproducible on Android devices after https://github.com/status-im/status-mobile/pull/14618 (it is more visible in the PR than in the latest develop though).

Also checked https://github.com/status-im/status-mobile/pull/14753 - looks better with the fix, thanks @yqrashawn.
This is a slow-mo video of how the animation looks like in 14753 with steps from the issue description


https://user-images.githubusercontent.com/67952253/211784709-7ef164a0-a5af-47f7-be21-b738256d3ffe.mp4

J-Son89:
[...]

PR #14753, opened 2023-01-11
fix: not using layout animation, causing flickers on android

yqrashawn:
related https://github.com/status-im/status-mobile/issues/14752

status: ready

### DIFF
--- a/src/status_im2/common/toasts/view.cljs
+++ b/src/status_im2/common/toasts/view.cljs
@@ -8,23 +8,23 @@
             [status-im2.common.toasts.style :as style]
             [utils.re-frame :as rf]))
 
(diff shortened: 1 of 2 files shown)


Parveshdhull (approved):

src/status_im2/common/toasts/view.cljs
```diff
@@ -83,7 +83,10 @@
             [reanimated/view
              {:entering slide-in-up-animation
               :exiting  slide-out-up-animation
```
Parveshdhull: we need to disable these too

yqrashawn: These two are removed.

PR #14766, opened 2023-01-12
Removed removed files and add untracked from linting

cammellos:
The linter would fail if there were removed files, as it would try to lint them but would not find them.
Similarly, untracked files would not be linted.

This commit changes the behavior so that untracked files are linted and removed files are ignored, that way we can run it before committing if there are unstaged changes that include removed/untracked files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,18 +286,24 @@ endif
 # Tests
 #--------------
 
+# Get all clojure files, including untracked, excluding removed
+define find_all_clojure_files
+$$(comm -23 <(sort <(git ls-files --cached --others --exclude-standard)) <(sort <(git ls-files --deleted)) | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn)
+endef
+
 lint: export TARGET := default
 lint: ##@test Run code style checks
-	sh scripts/lint-re-frame-in-quo-components.sh && \
+	@sh scripts/lint-re-frame-in-quo-components.sh && \
 	clj-kondo --config .clj-kondo/config.edn --cache false --lint src && \
-	ALL_CLOJURE_FILE=$$(git ls-files | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn$$) && \
-	zprint '{:search-config? true}' -sfc $$ALL_CLOJURE_FILE
+	ALL_CLOJURE_FILES=$(call find_all_clojure_files) && \
+	zprint '{:search-config? true}' -sfc $$ALL_CLOJURE_FILES
 
+# NOTE: We run the linter twice because of https://github.com/kkinnear/zprint/issues/271
 lint-fix: export TARGET := default
 lint-fix: ##@test Run code style checks and fix issues
-	ALL_CLOJURE_FILE=$$(git ls-files | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn$$) && \
-	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE && \
-	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE
+	ALL_CLOJURE_FILES=$(call find_all_clojure_files) && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILES && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILES
 
 
 shadow-server: export TARGET := clojure


yqrashawn (approved):

Makefile
```diff
@@ -282,6 +282,11 @@ else
 	npx react-native run-ios
 endif
 
+# Get all clojure files, including untracked, excluding removed
+define all_clojure_files
```
jakubgs: Can you move this below into the `Tests` section?

Makefile
```diff
@@ -290,14 +295,14 @@ lint: export TARGET := default
...
+	zprint '{:search-config? true}' -sfc $$ALL_CLOJURE_FILES
 
 lint-fix: export TARGET := default
 lint-fix: ##@test Run code style checks and fix issues
-	ALL_CLOJURE_FILE=$$(git ls-files | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn$$) && \
-	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE && \
-	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILE
+	ALL_CLOJURE_FILES=$(call all_clojure_files) && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILES && \
+	zprint '{:search-config? true}' -sw $$ALL_CLOJURE_FILES
```
jakubgs, line 306: Why is this command being called twice? Is this some sort of hack-fix?

Makefile
jakubgs: Also, this is a function, so the name should probably be `find_all_clojure_files`.

Makefile
```diff
@@ -282,6 +282,11 @@ else
 	npx react-native run-ios
 endif
 
+# Get all clojure files, including untracked, excluding removed
+define all_clojure_files
+$$(comm -23 <(sort <(git ls-files --cached --others --exclude-standard)) <(sort <(git ls-files --deleted)) | grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn)
```
jakubgs: We could make this into an actual function rather than just a definition of commands we want to run:
```suggestion
$(shell 
	comm -23 \
		<(sort <(git ls-files --cached --others --exclude-standard)) \
		<(sort <(git ls-files --deleted)) \
		| grep -e \.clj$$ -e \.cljs$$ -e \.cljc$$ -e \.edn
)
```
But then we need double quotes around the `$(call find_all_clojure_files)`.

Makefile
```diff
@@ -290,14 +295,14 @@ lint: export TARGET := default
 lint: ##@test Run code style checks
 	sh scripts/lint-re-frame-in-quo-components.sh && \
```
jakubgs: Maybe we can quiet it down a bit with a `@`?
```suggestion
	@sh scripts/lint-re-frame-in-quo-components.sh && \
```
Same for the other one.

jakubgs (approved):

Makefile
ilmotta: Yes, it's a temporary workaround while the maintainer of zprint is working on a fix. @yqrashawn already reported the bug.

jakubgs: I see, then a `TODO`/`FIXME` comment explaining that is usually in order.

ilmotta (approved):

Makefile
ilmotta: Reference:

- https://github.com/kkinnear/zprint/issues/271

Makefile
erikseppanen: Would be nice to accomplish this in one `git ls-files` command, but I wasn't able to. I hoped some combination of `--exclude=':S' would do it, but apparently not.

I haven't seen `comm` before - interesting!

erikseppanen (approved):

Makefile
erikseppanen: I have the lingering feeling that this could all be done in one `ls-files` command, including the [`grep` part](https://stackoverflow.com/questions/36753573/how-do-i-exclude-files-from-git-ls-files).

But I am also getting rubik's cube vibes (seeing this as an interesting puzzle to solve) that no one else would be interested in or in maintaining. We might be more comfortable with this mix of `comm, sort, ls-files, grep` (which makes sense) over figuring out arguments to `ls-files`. However, if we have a lot of `ls-files` stuff to do, it might pay off.

If there's interest, I might pursue it, otherwise I'll move on to other things. @jakubgs

cammellos: @jakubgs I'd keep it as it is for now, but if you feel strongly I can address in a separate PR (I really want to merge before the weekend ;) ) but I have address the rest of the points, thanks for the review!

PR #14767, opened 2023-01-12
Move bottom sheet to use RF events

ibrkhalil:


src/utils/re_frame.cljs
```diff
@@ -84,3 +84,5 @@
 (def sub (comp deref re-frame/subscribe))
 
 (def dispatch re-frame/dispatch)
+
+(def dispatch-sync re-frame/dispatch-sync)
```
flexsurfer, line 88: could you please elaborate on why sync is needed ?

src/status_im2/subs/general.cljs
```diff
@@ -198,6 +198,22 @@
  (fn [seed]
    (or seed {:step :intro})))
 
+(re-frame/reg-sub
+ :bottom-sheet/config
```
flexsurfer, line 202: two points here
1. (get-in db [:bottom-sheet/config should be a root sub
2. you could use destructuring with :or default values

src/status_im2/contexts/quo_preview/bottom_sheet/bottom_sheet.cljs
```diff
@@ -26,7 +26,7 @@
   [rn/view
    {:style {:justify-content :center
             :align-items     :center}}
-   [button/button {:on-press #(do (re-frame/dispatch [:bottom-sheet/hide]))} "Close bottom sheet"]
+   [button/button {:on-press #(do (re-frame/dispatch-sync [:dismiss-bottom-sheet]))} "Close bottom sheet"]
```
flexsurfer, line 29: extra (do

src/utils/re_frame.cljs
ibrkhalil: If two bottom sheets got opened consecutively, the new bottom sheet will have the older one's height.
So we need to dispatch the closing mechanism immediately and queue the opening of the new one so that the new one won't have any knowledge that there was one before it and will have the correct height

flexsurfer: why do we want to store shared height ?

ibrkhalil: I guess there's additional calculations that rely on it on bottom sheet, I don't know why it needs to be an atom

flexsurfer: it fine to have it in atom, but in local state, so new atom for new bottom sheet

src/status_im2/subs/general.cljs
ibrkhalil: 1- You mean having each key in the `:bottom-sheet/config` has a separate root subscription ?

src/utils/re_frame.cljs
ibrkhalil: This PR moved [here](https://github.com/status-im/status-mobile/pull/14791/files)

src/status_im2/subs/general.cljs
ibrkhalil: This PR moved [here](https://github.com/status-im/status-mobile/pull/14791/files)

PR #14791, opened 2023-01-15
[14677, 14673, 14672, 14671, 14664] Fix bottom sheet dismissal problems

ibrkhalil:
The PR makes the bottom sheet uses Re-frame's DB instead of local state, This makes it's configuration accessible from any part of the app, And fixes and issues that happens when two bottom sheets are opened after one another, The height doesn't have enough time to get calculated properly .. So we dismiss the older bottom-sheet synchronously and queue the dispatching of the new one to give it a chance to calculate it's content height.

fixes #14677, #14673, #14672, #14671, #14664
status: ready

PR #14792, opened 2023-01-15
Add contact list subscription test

ibrkhalil:
status: ready

### DIFF
--- a/src/status_im2/subs/contact.cljs
+++ b/src/status_im2/subs/contact.cljs
@@ -8,7 +8,8 @@
             [status-im.ui.screens.profile.visibility-status.utils :as visibility-status-utils]
             [status-im.utils.gfycat.core :as gfycat]
             [status-im.utils.image-server :as image-server]
-            [utils.collection]))
+            [utils.collection]
+            [status-im.constants :as constants]))
 
 (re-frame/reg-sub
  ::query-current-chat-contacts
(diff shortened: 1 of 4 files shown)


flexsurfer:
it seems like its not only test here there is new functionality as well, do we have a taks ?

src/status_im2/subs/contact_test.cljs
```diff
@@ -0,0 +1,183 @@
+(ns status-im2.subs.contact-test
+  (:require [cljs.test :refer [is testing]]
+            [re-frame.db :as rf-db]
+            [test-helpers.unit :as h]
+            status-im2.subs.contact
+            [utils.re-frame :as rf]))
+
+(def ^:private contacts-sample-data
```
ilmotta, line 8: I would say privacy is irrelevant in test files because nothing should require a test file, except for the test runner.

src/status_im2/subs/contact_test.cljs
```diff
@@ -0,0 +1,183 @@
...
+                                        :socialLinks           nil
+                                        :blocked               false
+                                        :verificationStatus    0
+                                        :lastUpdatedLocally    1672582563204
+                                        :public-key            "0xtest"
+                                        :has-added-us          true
+                                        :contact-request-state 1}}})
+
+(def ^:private expected-sorted-contacts
+  '({:title "F"
```
ilmotta, line 72: No need to use lists in test assertions. Just replace them with vectors. This is more idiomatic in Clojure tests because equality of `(is (= [1 2 3] '(1 2 3)))` is true. By using vectors everywhere in tests there's no need to *quote* collections with the `'` character.

It's rare to see tests that actually care about strict equality. In general, it doesn't matter for the test if the return value is a lazy collection returned by a `map` call or if the return value is a vector returned by `mapv`.

```clojure
[{:title "F"
  :data [{:active? true]
          :last-updated 1672582629695
          ...
  ...
```

src/status_im2/subs/contact.cljs
```diff
@@ -102,6 +103,23 @@
...
+ :<- [:selected-contacts-count]
+ (fn [[contacts selected-contacts-count]]
+   (->> contacts
+        (filter :mutual?)
+        (map #(assoc %
+                     :allow-new-users?
+                     (< selected-contacts-count
+                        (dec constants/max-group-chat-participants))))
+        (group-by (comp (fnil string/upper-case "") first :alias))
+        (sort-by (fn [[title]] title))
```
ilmotta, line 118: `(sort-by first)` is sufficient

ilmotta:
@ibrkhalil, what was the reason for closing the PR? It'd be nice if we can close PRs with a quick explanation/comment.

ibrkhalil:
> @ibrkhalil, what was the reason for closing the PR? It'd be nice if we can close PRs with a quick explanation/comment.

Sorry for not explaining Icaro
I had the tests in [this](https://github.com/status-im/status-mobile/pull/14791) PR already

PR #14793, opened 2023-01-16
make sure nodejs uses a fixed timezone

### LINKED ISSUE #14742: test group-transactions-by-date fails `status_im2/subs/subs_test.cljs:24:9` due to timezone issues
# Bug Report
When I run `make test` locally this test fails with the following output :

```clojure
FAIL in (group-transactions-by-date) (status_im2/subs/subs_test.cljs:24:9)
Check if transactions are sorted by date
expected: (= (wallet.transactions/group-transactions-by-date transactions) grouped-transactions)
actual: (not (= ({:title "20 Sep", :key :20170920, :data ({:timestamp "1505912551000"})} {:title "19 Sep", :key :20170919, :data ({:timestamp "1505764322000"})} {:title "18 Sep", :key :20170918, :data ({:timestamp "1505750000000"})}) ({:title "20 Sep", :key :20170920, :data ({:timestamp "1505912551000"})} {:title "18 Sep", :key :20170918, :data ({:timestamp "1505764322000"} {:timestamp "1505750000000"})})))
```

## Problem 
Its a timezone problem.
We need to update the test so that a change in timezone does not break the test for no good reason :D 

#### Expected behaviour
Test does not produce a failure just because of a change in timezone of the environment.

siddarthkay:
we had a _timebomb_ tests in the codebase that would break on some environments and work on other environments. Making sure nodejs specifies a timezone fixes them.

fixes #14742

status: ready

### DIFF
--- a/test-resources/override.js
+++ b/test-resources/override.js
@@ -1,6 +1,6 @@
 var Mocks = require('../target/mocks/mocks.js');
 var Module = require('module');
-process.env.TZ = 'Etc/Universal'
+process.env.TZ = 'Etc/UTC'
 
 const originalLoader = Module._load;
 
--- a/test-resources/override.js
+++ b/test-resources/override.js
@@ -1,5 +1,6 @@
 var Mocks = require('../target/mocks/mocks.js');
 var Module = require('module');
+process.env.TZ = 'Europe/Amsterdam'
 
 const originalLoader = Module._load;
 
--- a/test-resources/override.js
+++ b/test-resources/override.js
@@ -1,6 +1,6 @@
 var Mocks = require('../target/mocks/mocks.js');
 var Module = require('module');
-process.env.TZ = 'Europe/Amsterdam'
+process.env.TZ = 'Etc/Universal'
 
 const originalLoader = Module._load;
 


test-resources/override.js
```diff
@@ -1,5 +1,6 @@
 var Mocks = require('../target/mocks/mocks.js');
 var Module = require('module');
+process.env.TZ = 'Europe/Amsterdam'
```
cammellos: We can probably set to UTC, just to make sure it's not dependent on summer time etc

test-resources/override.js
```diff
@@ -1,5 +1,6 @@
 var Mocks = require('../target/mocks/mocks.js');
 var Module = require('module');
+process.env.TZ = 'Etc/Universal'
```
cammellos: `Etc/UTC` better if there's is, `Etc/Universal` is the old name

J-Son89 (approved):

test-resources/override.js
siddarthkay: it does : https://nodejs.org/api/cli.html#tz:~:text=such%20as%20%27-,Etc/UTC,-%27%2C%20%27Europe
updated

cammellos (approved):

PR #14761, opened 2023-01-11
fix community join buttons hiding after joining

J-Son89:
fixes https://github.com/status-im/status-mobile/issues/14659

PR #14799, opened 2023-01-16
Refactor mock data out of community page

### LINKED ISSUE #14462: Move mock data to central location and out of views.
As discussed with @ilmotta -https://github.com/status-im/status-mobile/pull/14444#discussion_r1033704197

It would be nice to move mock data to a higher centralised location in the application.
This means that views etc do not need to handle such data mocking and eventually when we go to remove the mock data then it is done in one place 
and the respective views do not need adjustment.

Discuss with the team on an appropriate solution.
Document this discussion and decision

Go through the codebase and find any mock data.
Apply the solution that the team has agreed upon.

Some examples:
"src/status_im2/contexts/communities/discover/view.cljs"
"src/status_im2/contexts/communities/overview/view.cljs"

discuss with Desktop team if they have any knowledge of this api should look so that we can match what will eventually be implemented.

vkjr:
Partially fixes #14462.

This PR doesn't move mock data to the `status-go` level, but separates mock data and makes the next changes on the community page easier (f.ex. the one @J-Son89 has in #14761)

**UPDATE**: After discussions it was decided to get rid of mock data and not to add any toggles for it.

https://user-images.githubusercontent.com/5786310/212721333-05153cd5-1813-47da-aacf-ab83b5eb91e3.mov

### Summary
- community pages with new design can be opened directly (no need to go into "discover" page)
- real list of community channels available
- mock data moved to a separate namespace
- mock channels and mock gating info can be toggled from the menu. Should be removed later when mocks moved to `status-go`, or when we will have new communities with implemented gating.
- community overview page was refactored, reagent components are more idiomatic now:
  - data from the inner components to the top passed via callbacks instead of atoms
  - got rid of memoizing in favor of correct reagent form-2 components. That reduces complexity and gives the possibility to use re-frame subscriptions easier. (No performance implications noticed).  

#### Platforms

- Android
- iOS

status: ready

### DIFF
--- a/src/status_im2/common/scroll_page/view.cljs
+++ b/src/status_im2/common/scroll_page/view.cljs
@@ -28,98 +28,100 @@
 
 (defn scroll-page-header
   [scroll-height name page-nav cover sticky-header]
-  [:f>
-   (fn []
-     (let [input-range         (if platform/ios? [-47 10] [0 150])
-           output-range        (if platform/ios? [-100 0] [-169 -45])
-           y                   (reanimated/use-shared-value @scroll-height)
-           translate-animation (reanimated/interpolate y
-                                                       input-range
-                                                       output-range
-                                                       {:extrapolateLeft  "clamp"
-                                                        :extrapolateRight "clamp"})
-           opacity-animation   (reanimated/use-shared-value 0)
-           threshold           (if platform/ios? 30 170)]
-       (rn/use-effect
-        #(do
-           (reanimated/set-shared-value y @scroll-height)
-           (reanimated/set-shared-value opacity-animation
-                                        (reanimated/with-timing (if (>= @scroll-height threshold) 1 0)
-                                                                (clj->js {:duration 300}))))
-        [@scroll-height])
-       [:<>
-        [reanimated/blur-view
-         {:blur-amount   32
-          :blur-type     :xlight
-          :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
-          :style         (style/blur-slider translate-animation)}]
-        [rn/view
-         {:style {:z-index    6
-                  :margin-top (if platform/ios? 44 0)}}
-         [reanimated/view
-          {:style (style/sticky-header-title opacity-animation)}
-          [rn/image
-           {:source cover
-            :style  style/sticky-header-image}]
-          [quo/text
-           {:size   :paragraph-1
-            :weight :semi-bold
-            :style  {:line-height 21}}
-           name]]
-         [quo/page-nav
-          {:horizontal-description? true
-           :one-icon-align-left?    true
-           :align-mid?              false
-           :page-nav-color          :transparent
-           :mid-section             {:type            :text-with-description
-                                     :main-text       nil
-                                     :description-img nil}
-           :right-section-buttons   (:right-section-buttons page-nav)
-           :left-section            {:icon                  :i/close
-                                     :icon-background-color (icon-color)
-                                     :on-press              #(rf/dispatch [:navigate-back])}}]
-         (when sticky-header [sticky-header @scroll-height])]]))])
+  (let [input-range         (if platform/ios? [-47 10] [0 150])
+        output-range        (if platform/ios? [-100 0] [-169 -45])
+        y                   (reanimated/use-shared-value scroll-height)
+        translate-animation (reanimated/interpolate y
+                                                    input-range
+                                                    output-range
+                                                    {:extrapolateLeft  "clamp"
+                                                     :extrapolateRight "clamp"})
+        opacity-animation   (reanimated/use-shared-value 0)
+        threshold           (if platform/ios? 30 170)]
+    (rn/use-effect
+     #(do
+        (reanimated/set-shared-value y scroll-height)
+        (reanimated/set-shared-value opacity-animation
+                                     (reanimated/with-timing (if (>= scroll-height threshold) 1 0)
+                                                             (clj->js {:duration 300}))))
+     [scroll-height])
+    [:<>
+     [reanimated/blur-view
+      {:blur-amount   32
+       :blur-type     :xlight
+       :overlay-color (if platform/ios? colors/white-opa-70 :transparent)
+       :style         (style/blur-slider translate-animation)}]
+     [rn/view
+      {:style {:z-index    6
+               :margin-top (if platform/ios? 44 0)}}
+      [reanimated/view
+       {:style (style/sticky-header-title opacity-animation)}
+       [rn/image
+        {:source cover
+         :style  style/sticky-header-image}]
+       [quo/text
+        {:size   :paragraph-1
+         :weight :semi-bold
+         :style  {:line-height 21}}
+        name]]
+      [quo/page-nav
+       {:horizontal-description? true
+        :one-icon-align-left?    true
+        :align-mid?              false
+        :page-nav-color          :transparent
+        :mid-section             {:type            :text-with-description
+                                  :main-text       nil
+                                  :description-img nil}
+        :right-section-buttons   (:right-section-buttons page-nav)
+        :left-section            {:icon                  :i/close
+                                  :icon-background-color (icon-color)
+                                  :on-press              #(rf/dispatch [:navigate-back])}}]
+      sticky-header]]))
+
 
 (defn display-picture
   [scroll-height cover]
-  [:f>
-   (fn []
-     (let [input-range (if platform/ios? [-67 10] [0 150])
-           y           (reanimated/use-shared-value @scroll-height)
-           animation   (reanimated/interpolate y
-                                               input-range
-                                               [1.2 0.5]
-                                               {:extrapolateLeft  "clamp"
-                                                :extrapolateRight "clamp"})]
-       (rn/use-effect #(do
-                         (reanimated/set-shared-value y @scroll-height)
-                         js/undefined)
-                      [@scroll-height])
-       [reanimated/view
-        {:style (style/display-picture-container animation)}
-        [rn/image
-         {:source cover
-          :style  style/display-picture}]]))])
+  (let [input-range (if platform/ios? [-67 10] [0 150])
+        y           (reanimated/use-shared-value scroll-height)
+        animation   (reanimated/interpolate y
+                                            input-range
+                                            [1.2 0.5]
+                                            {:extrapolateLeft  "clamp"
+                                             :extrapolateRight "clamp"})]
+    (rn/use-effect #(do
+                      (reanimated/set-shared-value y scroll-height)
+                      js/undefined)
+                   [scroll-height])
+    [reanimated/view
+     {:style (style/display-picture-container animation)}
+     [rn/image
+      {:source cover
+       :style  style/display-picture}]]))
 
 (defn scroll-page
-  [cover page-nav name]
+  [_ _ _]
   (let [scroll-height (reagent/atom negative-scroll-position-0)]
-    (fn [sticky-header children]
+    (fn [{:keys [cover-image page-nav-right-section-buttons name on-scroll]}
+         sticky-header
+         children]
       [:<>
-       [scroll-page-header scroll-height name page-nav cover sticky-header]
+       [:f> scroll-page-header @scroll-height name page-nav-right-section-buttons cover-image
+        sticky-header]
        [rn/scroll-view
         {:style                           (style/scroll-view-container
                                            (diff-with-max-min @scroll-height 16 0))
          :shows-vertical-scroll-indicator false
          :scroll-event-throttle           16
-         :on-scroll                       (fn [event]
+         :on-scroll                       (fn [^js event]
                                             (reset! scroll-height (int
                                                                    (oops/oget
                                                                     event
-                                                                    "nativeEvent.contentOffset.y"))))}
+                                                                    "nativeEvent.contentOffset.y")))
+                                            (when on-scroll
+                                              (on-scroll @scroll-height)))}
         [rn/view {:style {:height 151}}
          [rn/image
-          {:source cover
+          {:source cover-image
            :style  {:overflow :visible
                     :flex     1}}]]
         (when children
@@ -129,5 +131,5 @@
             :background-color (colors/theme-colors
                                colors/white
                                colors/neutral-90)}
-           [display-picture scroll-height cover]
-           [children]])]])))
+           [:f> display-picture @scroll-height cover-image]
+           children])]])))
--- a/src/status_im2/contexts/communities/home/view.cljs
+++ b/src/status_im2/contexts/communities/home/view.cljs
@@ -11,10 +11,7 @@
   [id]
   (let [community-item (rf/sub [:communities/home-item id])]
     [quo/communities-membership-list-item
-     {:on-press      (fn []
-                       (rf/dispatch [:communities/load-category-states id])
-                       (rf/dispatch [:dismiss-keyboard])
-                       (rf/dispatch [:navigate-to-nav2 :community {:community-id id}]))
+     {:on-press      #(rf/dispatch [:navigate-to :community-overview id])
       :on-long-press #(rf/dispatch
                        [:bottom-sheet/show-sheet
                         {:content       (fn []
--- a/src/status_im2/contexts/communities/overview/style.cljs
+++ b/src/status_im2/contexts/communities/overview/style.cljs
@@ -20,16 +20,15 @@
    :margin-left  :auto
    :margin-right :auto})
 
-(defn scroll-view-container
-  [border-radius]
-  {:position      :absolute
-   :top           -48
-   :overflow      :scroll
-   :border-radius border-radius
-   :height        "100%"})
-
 (def review-notice
   {:color        colors/neutral-50
    :margin-top   12
    :margin-left  :auto
    :margin-right :auto})
+
+(def community-overview-container
+  {:position :absolute
+   :top      (if platform/ios? 0 44)
+   :left     0
+   :right    0
+   :bottom   0})
--- a/src/status_im2/contexts/communities/overview/utils.cljs
+++ b/src/status_im2/contexts/communities/overview/utils.cljs
@@ -1,0 +1,21 @@
+(ns status-im2.contexts.communities.overview.utils
+  (:require [clojure.string :as string]
+            [utils.i18n :as i18n]))
+
+(defn join-existing-users-string
+  [user-list]
+  (let [users-count (count user-list)
+        first-two   (->> user-list
+                         (take 2)
+                         (map #(string/split (:full-name %) #" "))
+                         (map first))]
+    (case users-count
+      0 ""
+      1 (i18n/label :t/join-one-user {:user (first first-two)})
+      2 (i18n/label :join-two-users
+                    {:user1 (first first-two)
+                     :user2 (second first-two)})
+      (i18n/label :join-more-users
+                  {:user1      (first first-two)
+                   :user2      (second first-two)
+                   :left-count (- users-count 2)}))))
--- a/src/status_im2/contexts/communities/overview/view.cljs
+++ b/src/status_im2/contexts/communities/overview/view.cljs
@@ -12,164 +12,94 @@
             [status-im2.contexts.communities.overview.style :as style]
             [status-im2.contexts.communities.menus.community-options.view :as options]
             [status-im2.contexts.communities.menus.request-to-join.view :as join-menu]
+            [status-im2.contexts.communities.overview.utils :as utils]
             [utils.re-frame :as rf]))
 
-(def knc-token-img (js/require "../resources/images/tokens/mainnet/KNC.png"))
-(def mana-token-img (js/require "../resources/images/tokens/mainnet/MANA.png"))
-(def rare-token-img (js/require "../resources/images/tokens/mainnet/RARE.png"))
-(def eth-token-img (js/require "../resources/images/tokens/mainnet/ETH.png"))
-(def dai-token-img (js/require "../resources/images/tokens/mainnet/DAI.png"))
-
-;; Mocked list items
-(def user-list
-  [{:full-name "Alicia K"}
-   {:full-name "Marcus C"}
-   {:full-name "MNO PQR"}
-   {:full-name "STU VWX"}])
 
 (defn preview-user-list
-  []
+  [user-list]
   [rn/view style/preview-user
    [quo/preview-list
     {:type      :user
-     :user      user-list
-     :list-size 4
-     :size      24}]
+     :list-size (count user-list)
+     :size      24}
+    user-list]
    [quo/text
     {:accessibility-label :communities-screen-title
      :style               {:margin-left 8}
      :size                :label}
-    "Join Alicia, Marcus and 2 more"]]) ;; TODO remove mocked data and use from contacts list/communities members
+    (utils/join-existing-users-string user-list)]])
 
-(defn open-token-gating-mocked
-  [name emoji channel-color]
-  #(rf/dispatch
-    [:bottom-sheet/show-sheet
-     {:content
-      (fn []
-        [quo/token-gating
-         {:channel {:name                   name
-                    :community-color        (colors/custom-color :pink 50)
-                    :emoji                  emoji
-                    :emoji-background-color channel-color
-                    :on-enter-channel       (fn []
-                                              (js/alert
-                                               "Entered channel"
-                                               "Wuhuu!! You successfully entered the channel :)"))
-                    :gates                  {:read  [{:token          "KNC"
-                                                      :token-img-src  knc-token-img
-                                                      :amount         200
-                                                      :is-sufficient? true}
-                                                     {:token          "MANA"
-                                                      :token-img-src  mana-token-img
-                                                      :amount         10
-                                                      :is-sufficient? false
-                                                      :is-purchasable true}
-                                                     {:token          "RARE"
-                                                      :token-img-src  rare-token-img
-                                                      :amount         10
-                                                      :is-sufficient? false}]
-                                             :write [{:token          "KNC"
-                                                      :token-img-src  knc-token-img
-                                                      :amount         200
-                                                      :is-sufficient? true}
-                                                     {:token          "DAI"
-                                                      :token-img-src  dai-token-img
-                                                      :amount         20
-                                                      :is-purchasable true
-                                                      :is-sufficient? false}
-                                                     {:token          "ETH"
-                                                      :token-img-src  eth-token-img
-                                                      :amount         0.5
-                                                      :is-sufficient? false}]}}}])
-      :content-height 210}]))
+(defn channel-token-gating-details
+  [name token-gating emoji channel-color]
+  [rn/view {:height 350 :margin-top 20}
+   [quo/token-gating
+    {:channel {:name                   name
+               :community-color        channel-color
+               :emoji                  emoji
+               :emoji-background-color channel-color
+               :on-enter-channel       (fn []
+                                         (js/alert
+                                          "Entered channel"
+                                          "Wuhuu!! You successfully entered the channel :)"))
+               :gates                  token-gating}}]])
 
-(def mock-list-of-channels
-  {:Welcome [{:name  "welcome"
-              :emoji "🤝"}
-             {:name     "onboarding"
-              :emoji    "🍑"
-              :locked?  true
-              :on-press #((open-token-gating-mocked
-                           "onboarding"
-                           "🍑"
-                           (colors/custom-color :pink 50)))}
-             {:name     "intro"
-              :emoji    "🦄"
-              :locked?  true
-              :on-press #((open-token-gating-mocked
-                           "intro"
-                           "🦄"
-                           (colors/custom-color :pink 50)))}]
-   :General [{:name  "general"
-              :emoji "🐷"}
-             {:name     "people-ops"
-              :emoji    "🌏"
-              :locked?  true
-              :on-press #((open-token-gating-mocked
-                           "onboarding"
-                           "🌏"
-                           (colors/custom-color :blue 50)))}
-             {:name  "announcements"
-              :emoji "🎺"}]
-   :Mobile  [{:name  "mobile"
-              :emoji "👽"}
-             {:name  "mobile-ui"
-              :emoji "👽"}
-             {:name  "mobile-ui-reviews"
-              :emoji "👽"}]
-   :Desktop [{:name  "desktop"
-              :emoji "👽"}
-             {:name  "desktop-ui"
-              :emoji "👽"}
-             {:name  "desktop-ui-reviews"
-              :emoji "👽"}
-             {:name  "desktop2"
-              :emoji "👽"}
-             {:name  "desktop-ui2"
-              :emoji "👽"}
-             {:name  "desktop-ui2-reviews"
-              :emoji "👽"}]})
+(defn open-channel-token-gating-details
+  [name token-gating emoji channel-color]
+  (rf/dispatch
+   [:bottom-sheet/show-sheet
+    {:content
+     (fn []
+       [channel-token-gating-details name token-gating emoji channel-color])
+     :content-height 210}]))
 
-(defn channel-list-component-fn
-  [channel-heights first-channel-height]
-  [rn/view
-   {:on-layout #(swap! first-channel-height
-                  (fn []
-                    (+ (if platform/ios?
-                         0
-                         38)
-                       (int (Math/ceil (oops/oget % "nativeEvent.layout.y"))))))
-    :style     {:margin-top 20 :flex 1}}
-   (map-indexed (fn [index category]
-                  (let [first-category (first category)]
-                    ^{:key first-category}
-                    [rn/view
-                     {:flex      1
-                      :key       (str index first-category)
-                      :on-layout #(swap! channel-heights
-                                    (fn []
-                                      (sort-by :height
-                                               (conj @channel-heights
-                                                     {:height (int (oops/oget % "nativeEvent.layout.y"))
-                                                      :label  first-category}))))}
+(defn layout-y
+  [event]
+  (oops/oget event "nativeEvent.layout.y"))
 
-                     [quo/divider-label
-                      {:label            first-category
-                       :chevron-position :left}]
-                     [rn/view
-                      {:margin-left   8
-                       :margin-top    10
-                       :margin-bottom 8}
-                      (map-indexed (fn [inner-index channel-data]
-                                     [rn/view
-                                      {:key        (str inner-index (:name channel-data))
-                                       :margin-top 4}
-                                      [quo/channel-list-item channel-data]])
-                                   (first-category mock-list-of-channels))]]))
-                mock-list-of-channels)])
+(defn add-category-height
+  [categories-heights category height]
+  (swap! categories-heights
+    (fn []
+      (sort-by :height
+               (conj @categories-heights
+                     {:height height
+                      :label  category})))))
 
-(def channel-list-component (memoize channel-list-component-fn))
+(defn channel-list-component
+  [{:keys [on-categories-heights-changed
+           on-first-channel-height-changed]}
+   channels-list]
+  (let [categories-heights (reagent/atom [])]
+    [rn/view
+     {:on-layout #(on-first-channel-height-changed (+ (if platform/ios?
+                                                        0
+                                                        38)
+                                                      (int (Math/ceil (layout-y %)))))
+      :style     {:margin-top 20 :flex 1}}
+     (map-indexed
+      (fn [index [category channels-for-category]]
+        [rn/view
+         {:flex      1
+          :key       (str index category)
+          :on-layout #(do
+                        (add-category-height categories-heights category (int (layout-y %)))
+                        (on-categories-heights-changed @categories-heights))}
+
+         [quo/divider-label
+          {:label            category
+           :chevron-position :left}]
+         [rn/view
+          {:margin-left   8
+           :margin-top    10
+           :margin-bottom 8}
+          (map (fn [channel]
+                 [rn/view
+                  {:key        (:id channel)
+                   :margin-top 4}
+                  [quo/channel-list-item channel]])
+               channels-for-category)]])
+      channels-list)]))
 
 (defn request-to-join-text
   [is-open?]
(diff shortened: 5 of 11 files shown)


src/status_im2/common/scroll_page/view.cljs
```diff
@@ -79,19 +64,27 @@
...
-         :scroll-event-throttle 4
-         :on-scroll #(swap! scroll-height (fn [] (int (oops/oget % "nativeEvent.contentOffset.y"))))}
+         :scroll-event-throttle           4
+         :on-scroll                       #(do
+                                             (swap! scroll-height (fn []
+                                                                    (int
+                                                                     (oops/oget
+                                                                      %
+                                                                      "nativeEvent.contentOffset.y"))))
+                                             (on-scroll @scroll-height))}
```
ilmotta: ReactNative `ScrollView` has the `onScroll` prop optional, so I think we should do the same here. Check first with `(when on-scroll ...)`

src/status_im2/contexts/communities/menus/community_options/view.cljs
```diff
@@ -77,6 +77,17 @@
    :label               (i18n/label :t/share-community)
    :on-press            #(hide-sheet-and-dispatch [:communities/share-community-pressed id])})
 
+(def toggle-channel-mocks
+  {:icon                :i/edit
+   :accessibility-label :toggle-channel-mocks
+   :label               "Toggle channel mocks"
+   :on-press            #(hide-sheet-and-dispatch [:mocks/toggle-community-channels-mocking])})
```
ilmotta: Missing new line.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
+                                                        38)
+                                                      (int (Math/ceil (layout-y %)))))
+      :style     {:margin-top 20 :flex 1}}
+     (map-indexed
+      (fn [index category]
+        (let [first-category (first category)]
+          ^{:key first-category}
+          [rn/view
+           {:flex      1
+            :key       (str index first-category)
```
ilmotta: The `key` is defined twice for the `view` component, but only one is required (either the metadata map or the `:key` prop).

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
+               (conj @categories-heights
+                     {:height height
+                      :label  category})))))
+
+(defn channel-list-component
+  [{:keys [on-categories-heights-changed
+           on-first-channel-height-changed]}
+   channels-list]
+  (let [categories-heights (reagent/atom [])]
+
```
ilmotta: Extra line

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
+             :chevron-position :left}]
+           [rn/view
+            {:margin-left   8
+             :margin-top    10
+             :margin-bottom 8}
+            (map-indexed (fn [inner-index channel-data]
+                           [rn/view
+                            {:key        (str inner-index (:name channel-data))
+                             :margin-top 4}
+                            [quo/channel-list-item channel-data]])
```
ilmotta: I think the `data` suffix is redundant, because `channel` is just a map.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
+           [quo/divider-label
+            {:label            first-category
+             :chevron-position :left}]
+           [rn/view
+            {:margin-left   8
+             :margin-top    10
+             :margin-bottom 8}
+            (map-indexed (fn [inner-index channel-data]
+                           [rn/view
+                            {:key        (str inner-index (:name channel-data))
```
ilmotta: Does the `channel` have an ID? Can use use an ID instead of index + name?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
               (i18n/label :t/joined)
               (i18n/label :t/pending))}])
 
-(defn render-page-content
+
+(defn round-community-icon
+  [community scroll-height icon-top icon-size]
+  [rn/view
+   {:border-radius    40
```
ilmotta: This style map is large enough to deserve a refactor to a `style.cljs` file.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+    :top              (icon-top scroll-height)
+    :left             17
+    :padding          2
+    :background-color (colors/theme-colors
+                       colors/white
+                       colors/neutral-90)}
+   [quo/community-icon community
+    (icon-size scroll-height)]])
+
+
```
ilmotta: Extra line

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+(defn community-token-gating-details
+  [name thumbnail-image tokens]
+  [rn/view {:height 200 :margin-top 20} ;TODO: fix size
+   [quo/token-gating
+    {:community {:name                     name
+                 :community-color          colors/primary-50
+                 :community-avatar-img-src thumbnail-image
+                 :gates                    tokens}}]])
+
```
ilmotta, line 160: Extra line

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+    :background-color (colors/theme-colors
+                       colors/white
+                       colors/neutral-90)}
+   [quo/community-icon community
+    (icon-size scroll-height)]])
+
+
+(defn community-token-gating-details
+  [name thumbnail-image tokens]
+  [rn/view {:height 200 :margin-top 20} ;TODO: fix size
```
ilmotta: This TODO is not very actionable. If you think it's important enough, you can delete it and create an issue instead, which is what we've been trying to do recently.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+   [quo/token-gating
+    {:community {:name                     name
+                 :community-color          colors/primary-50
+                 :community-avatar-img-src thumbnail-image
+                 :gates                    tokens}}]])
+
+
+(defn categorize-chats
```
ilmotta: What do you think of moving this kind of logic outside the view and into a subscription? I think you can decouple the data processing part from the part that actually builds the component with all the event dispatches.

The end goal is to have a very dumb UI layer. Also, having a subscription will give you the chance to easily unit test it.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+                    (rf/dispatch [:chat.ui/navigate-to-chat-nav2 (str community-id id)])
+                    (rf/dispatch [:search/home-filter-changed nil]))}))))
+
+(defn add-on-press-handler-to-chats
+  [community-id chats]
+  (mapv (partial add-on-press-handler community-id) chats))
+
+(defn add-on-press-handler-to-categorized-chats
+  [community-id categorized-chats]
+  (into {}
```
ilmotta: It looks like you could use `reduce-kv` here instead of `into` and `map`.

Not tested:

```clojure
(defn add-on-press-handler-to-categorized-chats
  [community-id categorized-chats]
  (reduce-kv (fn [res category chats]
               (assoc res category (add-on-press-handler-to-chats community-id chats)))
             {}
             categorized-chats))
```

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -218,142 +167,241 @@
...
+(defn icon-top-fn
+  [scroll-height]
+  (if (<= scroll-height scroll-page/negative-scroll-position-0)
+    -40
+    (->> (+ scroll-page/scroll-position-0 scroll-height)
+         (* (if platform/ios? 3 1))
+         (+ -40)
+         (min 8))))
+
+(defn icon-size-fn
```
ilmotta: It's more common to just name the function without the suffix `fn`. Consumers should know if it's a function already. But I see the original code used this name, fine :man_shrugging:

ilmotta (commented):
@vkjr I understand the motivation behind this PR, as I also made some comments about mocked code in production related to issue https://github.com/status-im/status-mobile/issues/14462

My comment here is more general to the idea of solving #14462, not exclusive to your PR.

When I made the suggestions to isolate mock implementation, I meant that the UI doesn't need and shouldn't know anything about *how* and from *where* the data is fetched. There's no need for us to keep adding subscriptions with the `word` mock in their name or defining events with the word mock, etc.

As long as we have a reasonable agreement on the contract of the data structures that status-go will publish (which we should for the most part at this point in the race), we can code as if the data is coming from the real sources.

Therefore, when the time comes to switch to real data, the UI shouldn't need to change a single bit or change a lot less.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
-(def eth-token-img (js/require "../resources/images/tokens/mainnet/ETH.png"))
-(def dai-token-img (js/require "../resources/images/tokens/mainnet/DAI.png"))
 
-;; Mocked list items
-(def user-list
-  [{:full-name "Alicia K"}
-   {:full-name "Marcus C"}
-   {:full-name "MNO PQR"}
-   {:full-name "STU VWX"}])
+(defn join-existing-users-string
```
ilmotta: This is some heavy logic in the view ;)

I think this function should be either a subscription or some utility function that can live a local `utils` file or directory. This idea is documented in the Project Structure section of the guidelines (https://github.com/status-im/status-mobile/blob/develop/doc/new-guidelines.md#project-structure)

Having the subscription or localized utility function will also help us write a unit test for this function. It's begging for tests ;)

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -4,171 +4,120 @@
...
-                      (map-indexed (fn [inner-index channel-data]
-                                     [rn/view
-                                      {:key        (str inner-index (:name channel-data))
-                                       :margin-top 4}
-                                      [quo/channel-list-item channel-data]])
-                                   (first-category mock-list-of-channels))]]))
-                mock-list-of-channels)])
-
-(def channel-list-component (memoize channel-list-component-fn))
+  [rn/view {:height 350 :margin-top 20} ;TODO: fix size
```
ilmotta: Same comment I made about the TODO. If this TODO is important, better have a separate GH issue, and consider deleting it.

flexsurfer:
> This PR doesn't move mock data to the status-go level

im not sure we want this at all, actually we shouldn't have any mock data at all in the app, we have ui tests no?

src/status_im2/contexts/communities/events.cljs
```diff
@@ -0,0 +1,14 @@
+(ns status-im2.contexts.communities.events
+  (:require [utils.re-frame :as rf]))
+
+(rf/defn enable-community-channels-mocks
+  {:events [:mocks/toggle-community-channels-mocking]}
```
flexsurfer: im not sure , why we need this ? we should mock only in test, no code for mocks should be in app codebase

src/status_im2/contexts/communities/menus/community_options/view.cljs
```diff
@@ -77,6 +77,17 @@
    :label               (i18n/label :t/share-community)
    :on-press            #(hide-sheet-and-dispatch [:communities/share-community-pressed id])})
 
+(def toggle-channel-mocks
```
flexsurfer: for sure it shouldn't be here

flexsurfer (changes requested):

flexsurfer:
my suggestion would be just to remove all mocked data or move it to ui tests, i don't see any value in having it in the main codebase

ilmotta:
> my suggestion would be just to remove all mocked data or move it to ui tests, i don't see any value in having it in the main codebase

@flexsurfer, while I agree with you, I'm not sure we can remove all mocked data because many of the features implemented for communities rely on fake data at the moment, otherwise a considerable number of features won't display anything (no data).

Btw, how's the implementation going on the backend @vkjr, @J-Son89? Are we ready to start consuming from the backend and stopping using fake data?

I believe the original strategy for community screens were: *let's advance fast in the frontend without the backend*. There's nothing wrong with this strategy IMO as long as the contracts between backend and frontend are well established.

As I commented here https://github.com/status-im/status-mobile/pull/14799#pullrequestreview-1250551376, the problem to me is the **views and subscriptions have concrete knowledge of where the data comes from** (mocks). They should not really.

If we need fake data for development while something is not ready in status-go, then **we should mock only at the re-frame event layer to mimic real RPC calls**, or even lower than that to mimic the communication between back & front. After the mock data is inserted in the app db, the views, subscriptions and everything else should not be able to tell if something is fake or not.

vkjr:
@ilmotta, @flexsurfer, thank you for all the comments!
I see that having additional toggles in UI data is a highly arguable decision, so let me add a few cents)

This PR is for sure not considered as a full fix for #14462, just as a step towards it. 
When I started to get the hardcoded data out of components, I understood that the components themselves also need to be refactored and this is a main improvement in the PR, so I just separated that hardcoded data and concentrated on the components.
So far I didn't bring up the discussion as to where to store the mock data (which is definitely my mistake, I should've started it earlier!) because in my perception work on this part of the task is not really started yet). But it still would be nice to have view changes committed to not mess with rebasing. Also, it will make changes like #14761 easier.

And my rationale behind having toggles in UI at this moment was simple: 
- we already have some UI parts that rely on hardcoded data, so to see them we'd better keep that data for now
- mock data should be an option
- toggles are a quick and cheap solution to make mocks optional for now. And it can be very easily removed later

vkjr:
@ilmotta 

> Btw, how's the implementation going on the backend @vkjr, @J-Son89? Are we ready to start consuming from the backend and stopping using fake data?

Not sure, I didn't touch the backend yet.

> If we need fake data for development while something is not ready in status-go, then **we should mock only at the re-frame event layer to mimic real RPC calls**, or even lower than that to mimic the communication between back & front. After the mock data is inserted in the app db, the views, subscriptions and everything else should not be able to tell if something is fake or not.

I'm very up for this change, just wondering if we can make it a next separate step with a separate issue/PR?
Since we are in a phase where the product is not visible to clients, I believe we can make a series of smaller steps toward the goal even if at the end of each step the product is not fully polished, wdyt?

J-Son89:
> > Btw, how's the implementation going on the backend @vkjr, @J-Son89? Are we ready to start consuming from the backend and stopping using fake data?
> 
> Not sure, I didn't touch the backend yet.
> 
> > If we need fake data for development while something is not ready in status-go, then **we should mock only at the re-frame event layer to mimic real RPC calls**, or even lower than that to mimic the communication between back & front. After the mock data is inserted in the app db, the views, subscriptions and everything else should not be able to tell if something is fake or not.
> 
> I'm very up for this change, just wondering if we can make it a next separate step with a separate issue/PR? Since we are in a phase where the product is not visible to clients, I believe we can make a series of smaller steps toward the goal even if at the end of each step the product is not fully polished, wdyt?

@ilmotta, @vkjr, @flexsurfer, some communities features were not developed yet in the backend and needed mock data to build the ui out. There's also the issue that there's no featured communities yet.

As discussed before I think it would be beneficial if we had some embedded communities in the application so that we can use these "real" communities with real data and validate the ui is correct.

I will get in touch with the desktop team to align with where we are at and we can figure out where we need to go from here to get communities in a more production ready state.

Ideally we get rid of mock data completely but I agree for now it's best to move this higher up the application code so that there is less areas impacted by it.

@vkjr perhaps a better option for the toggle is to put it in the options page (beside quo2 components menu). This way we have a global toggle for all mock data.? Personally I don't like that we corrupt the "ready" ui by adding temporary dev features in there.

flexsurfer:
> @flexsurfer, while I agree with you, I'm not sure we can remove all mocked data because many of the features implemented for communities rely on fake data at the moment, otherwise a considerable number of features won't display anything (no data).

im not sure i understand, but who needs these features to be displayed? if devs, they could just opt in them in the code

vkjr:
> @vkjr perhaps a better option for the toggle is to put it in the options page (beside quo2 components menu). This way we have a global toggle for all mock data.? Personally I don't like that we corrupt the "ready" ui by adding temporary dev features in there.

@J-Son89 
Yes, totally agree.

vkjr:
> im not sure i understand, but who needs these features to be displayed? if devs, they could just opt in them in the code

@flexsurfer,
As an absolutely theoretical thought: if some UI already exists in the code, but is not available in the app, it becomes kind of a "secret knowledge". Next developer needs to know beforehand that it exists instead of just seeing it)
So for me, if something is already implemented, it better be visible. Also, as an eyesore, it will have bigger chances to be fixed quicker. Wdyt?

flexsurfer:
> > im not sure i understand, but who needs these features to be displayed? if devs, they could just opt in them in the code
> 
> @flexsurfer, As an absolutely theoretical thought: if some UI already exists in the code, but is not available in the app, it becomes kind of a "secret knowledge". Next developer needs to know beforehand that it exists instead of just seeing it) So for me, if something is already implemented, it better be visible. Also, as an eyesore, it will have bigger chances to be fixed quicker. Wdyt?

but this sounds like we are trying to find a reason why we're doing this :) so if there are no real reasons we could just move all mocked data in separate dev namespace, and if anyone needs it he could use it, for sure we shouldn't have any toggles in the UI or whatever

vkjr:
> but this sounds like we are trying to find a reason why we're doing this :) so if there are no real reasons we could just move all mocked data in a separate dev namespace, and if anyone needs it he could use it, for sure we shouldn't have any toggles in the UI or whatever

@flexsurfer,
For me believing that already implemented UI should be visible is already existing reason)

What about testing reasons? For example, if community gating is not implemented at `status-go` level, but the QA team wants to manually check the functionality of related UI components since they are already implemented (maybe even on different sets of mock data). Would it be okay to have some toggle in the app to turn on mock data?
Just checking borders of your tolerance to UI toggles 😜

flexsurfer:
> > but this sounds like we are trying to find a reason why we're doing this :) so if there are no real reasons we could just move all mocked data in a separate dev namespace, and if anyone needs it he could use it, for sure we shouldn't have any toggles in the UI or whatever
> 
> @flexsurfer, For me believing that already implemented UI should be visible is already existing reason)
> 
> What about testing reasons? For example, if community gating is not implemented at `status-go` level, but the QA team wants to manually check the functionality of related UI components since they are already implemented (maybe even on different sets of mock data). Would it be okay to have some toggle in the app to turn on mock data? Just checking borders of your tolerance to UI toggles 😜

i don't see any reasons to develop UI without real data, components sure can be done and available in quo2 preview for designers, but no reason to have it in the app with mocked data, when data will be available then QA should test it, imo

vkjr:
> i don't see any reasons to develop UI without real data, components sure can be done and available in quo2 preview for designers, but no reason to have it in the app with mocked data, when data will be available then QA should test it, imo

I believe that concurrency is the reason) We don't really need to wait for real data if we can mock it. But of course I agree that mocks should be as close as possible to real data and not on the level of the UI.

ilmotta:
@vkjr, @J-Son89, Thanks for all the explanations. I think I'm on the same page as you guys.

>i don't see any reasons to develop UI without real data, components sure can be done and available in quo2 preview for designers, but no reason to have it in the app with mocked data, when data will be available then QA should test it, imo

@flexsurfer, the fake data is a way to iterate on the UI development without being blocked by the backend implementation, and I believe this was the original intention. It greatly helped us achieve what we have now related to Communities. As far as I understand, we can't rely on the QA team to test some features, because status-go is not ready to persist and process some of the data.

Using fake data to accelerate development is a very common strategy. I've seen this strategy being used many times with great success. If we had real users, we would obviously add a feature toggle and not display anything to end-users. And we have to be careful not to create a false sense of progress/velocity, because once the backend is implemented, we usually find out that the frontend made incorrect assumptions. So as is the case of any strategy, there are trade-offs.

To me, it's just how it's done now that I think is far from ideal, but good thing we have a clear path on how to technically solve this problem.

vkjr:
@ilmotta, @flexsurfer, @J-Son89 ,

I would suggest the following:

This PR:
- all UI toggles removed, so mock data won't be visible
- in the code, in `communities/overview/views.cljs` temporary flag left to enable that mock data
- review notes fixed
merge.

Next PR:
- I will create one community, that will have special treatment. When it is fetched, before going to re-frame db it will be additionally populated with mock data to imitate missing gating features. 
- all mock data removed from UI level

This way, we won't need to have any toggles to turn on mock data, but we still will be able to see it in one special community.

Wdyt?

flexsurfer:
> the fake data is a way to iterate on the UI development without being blocked by the backend implementation, and I believe this was the original intention. It greatly helped us achieve what we have now related to Communities. As far as I understand, we can't rely on the QA team to test some features, because status-go is not ready to persist and process some of the data.

yes sure i know all of that, but I've reviewed all the new code for communities, and all we have is some mock data for components, which can be just in quo preview, so for this PR we just need this data to quo preview

later if we'll have a need to have a status-go API we could use mocked data for sure on status-go side

ilmotta:
> yes sure i know all of that, but I've reviewed all the new code for communities, and all we have is some mock data for components, which can be just in quo preview, so for this PR we just need this data to quo preview

I see, I agree with you, for those cases they should be just in quo preview.

src/status_im2/contexts/communities/overview/style.cljs
```diff
@@ -33,3 +33,9 @@
    :margin-top   12
    :margin-left  :auto
    :margin-right :auto})
+
+(def community-overview-container
+  {:position :absolute
+   :top      (if platform/ios? 0 44)
+   :width    "100%"
+   :height   "110%"})
```
flexsurfer: we should never use % for dimensions, this looks like magic

ilmotta:
> @ilmotta, @flexsurfer, @J-Son89 ,
> 
> I would suggest the following:
> 
> This PR:
> 
> * all UI toggles removed, so mock data won't be visible
> * in the code, in `communities/overview/views.cljs` temporary flag left to enable that mock data
> * review notes fixed
>   merge.
> 
> Next PR:
> 
> * I will create one community, that will have special treatment. When it is fetched, before going to re-frame db it will be additionally populated with mock data to imitate missing gating features.
> * all mock data removed from UI level
> 
> This way, we won't need to have any toggles to turn on mock data, but we still will be able to see it in one special community.
> 
> Wdyt?

I like the suggestions @vkjr, it's a great step in making the code more production ready once status-go has been sorted out. I'd just add what @flexsurfer suggested here https://github.com/status-im/status-mobile/pull/14799#issuecomment-1385309683

vkjr:
> I like the suggestions @vkjr, it's a great step in making the code more production ready once status-go has been sorted out. I'd just add what @flexsurfer suggested here [#14799 (comment)](https://github.com/status-im/status-mobile/pull/14799#issuecomment-1385309683)

Sure, then I will check if our existing quo-preview data needs anything from current mocks and will move mocks there. Or will remove them fully if nothing is needed.

src/status_im2/contexts/communities/overview/style.cljs
vkjr: This is a part of the code that I didn't write, just moved. But I'm not sure why in general we shouldn't use percentages for dimensions? `:width "100%"` looks quite self-explanatory to me.

Could you please elaborate a bit? Maybe we can put the reason into the guidelines document

J-Son89: I believe he is referring to height 110%. This is needed as the screen needs to overflow somewhat and tbh it doesn't matter how much it overflows, just more than the 100%

ilmotta: Perhaps he's referring to the fact that percentages are rarely needed in a flexbox rendering model, which is the case of RN.

flexsurfer: so here two different issues
1) we never should use percentage
  a) because 100% doesn't make sense in flexbox
   b) because we always have fixed margins or paddings in the design, so instead of useing 80% we  should use padding-horizontal 20, for example, because % will be always different on differrent devices in pixels, but we should always have same paddings in px on all devices
2) 110% and top looks like a hack, hacks are bad, better to find a proper reliable solution

flexsurfer: yes this would be cool to have in guidlines, i see more and more usage of % in new design implementations

vkjr:
@flexsurfer, @ilmotta, could you please re-review the changes? And in the meanwhile I'll try to rebase the PR, it looks like we have a significant clash with #14695

All the notes you provided are fixed.
There is only one small piece of mock data left:
```clojure
(re-frame/reg-sub
 :communities/users
 :<- [:communities]
 (fn [_ [_ _]]
   [{:full-name "Alicia K"}
    {:full-name "Marcus C"}
    {:full-name "MNO PQR"}
    {:full-name "STU VWX"}]))
```

Since we don't have a well-prepared community users list yet, I'd like to leave this mock to make the users list component visible, if you don't mind)

ilmotta:
> @flexsurfer, @ilmotta, could you please re-review the changes? And in the meanwhile I'll try to rebase the PR, it looks like we have a significant clash with #14695
> 
> All the notes you provided are fixed. There is only one small piece of mock data left:
> 
> ```clojure
> (re-frame/reg-sub
>  :communities/users
>  :<- [:communities]
>  (fn [_ [_ _]]
>    [{:full-name "Alicia K"}
>     {:full-name "Marcus C"}
>     {:full-name "MNO PQR"}
>     {:full-name "STU VWX"}]))
> ```
> 
> Since we don't have a well-prepared community users list yet, I'd like to leave this mock to make the users list component visible, if you don't mind)

Sure @vkjr, I'll be able to review later today, I mean, my local today ;)

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -80,19 +65,28 @@
...
        [rn/scroll-view
-        {:style (style/scroll-view-container (diff-with-max-min @scroll-height 16 0))
+        {:style                           (style/scroll-view-container (diff-with-max-min @scroll-height
+                                                                                          16
+                                                                                          0))
          :shows-vertical-scroll-indicator false
-         :scroll-event-throttle 4
-         :on-scroll #(swap! scroll-height (fn [] (int (oops/oget % "nativeEvent.contentOffset.y"))))}
+         :scroll-event-throttle           4
+         :on-scroll                       #(do
```
ilmotta: https://github.com/bbatsov/clojure-style-guide#no-multiple-forms-in-function-literals

From the guidelines, let's avoid multiple forms within `#()` as it hurts readability. I know some of us might disagree, but it's just easier if we follow the guide. I would love to have this reinforced by zprint, but I haven't bothered looking.

So replace the `#()` with `(fn [^js event])`. Some might say the type hint is unnecessary because we're using `oops/oget`, so it's up to you. I usually prefer to explicitly type hint native objects in function parameters.

src/status_im2/contexts/communities/overview/utils.cljs
```diff
@@ -0,0 +1,15 @@
...
+  (let [users-count (count user-list)
+        first-two   (->>
+                     user-list
+                     (take 2)
+                     (map #(first (string/split (:full-name %) #" "))))]
+    (case users-count
+      0 ""
+      1 (str "Join " (first first-two))
+      2 (str "Join " (string/join " and " first-two))
+      (str "Join " (string/join ", " first-two) " and " (- users-count 2) " more"))))
```
ilmotta: It seems we should be using translations in this function.

Sorry to mention this if you already know, but you can create translations as templates, like:

```clojure
(i18n/label :t/change-log-level {:log-level name})
```

```json
"change-log-level": "Confirm and restart the app to change log level to {{log-level}}"
```

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -219,142 +150,198 @@
...
+             categorized-chats))
+
+(defn community-header
+  [name]
+  [quo/text
+   {:accessibility-label :chat-name-text
+    :number-of-lines     1
+    :ellipsize-mode      :tail
+    :weight              :semi-bold
+    :size                :heading-1} name])
```
ilmotta: I thought zprint would format this for us, but apparently not.

From the style guide, `name` should be on a separate line.

Ref:
https://github.com/bbatsov/clojure-style-guide#function-arguments-alignment

src/status_im2/subs/communities.cljs
```diff
@@ -214,3 +214,30 @@
...
         (into []))))
+
+(re-frame/reg-sub
+ :communities/categorized-channels
+ (fn [[_ community-id]]
+   [(re-frame/subscribe [:communities/community community-id])])
+ (fn [[{:keys [joined categories chats]}] [_ _]]
+   (reduce
+    (fn [acc [_ {:keys [name categoryID id emoji can-post?]}]]
+      (let [category (keyword (get-in categories [categoryID :name] "None"))]
```
ilmotta: The `None` word should probably be translated.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -12,164 +12,95 @@
...
+      (sort-by :height
+               (conj @categories-heights
+                     {:height height
+                      :label  category})))))
+
+(defn channel-list-component
+  [{:keys [on-categories-heights-changed
+           on-first-channel-height-changed]}
+   channels-list]
+  (let [categories-heights (reagent/atom [])]
```
ilmotta, line 73: Shouldn't this local `categories-heights` state be defined using a form-2 component? Right now the state is recreated whenever `channel-list-component` renders. Even if this works right now, I think we should be extra cautious and redefine the component as a form-2 one.

Or of course maybe I got something wrong here.

src/status_im2/contexts/communities/overview/utils.cljs
```diff
@@ -0,0 +1,15 @@
+(ns status-im2.contexts.communities.overview.utils
+  (:require [clojure.string :as string]))
+
+(defn join-existing-users-string
+  [user-list]
+  (let [users-count (count user-list)
+        first-two   (->>
+                     user-list
+                     (take 2)
+                     (map #(first (string/split (:full-name %) #" "))))]
```
ilmotta: Since you're already using the thread last macro, you can *pipeline* a bit more for a little bit extra clarity :)

```clojure
(->> user-list
     (take 2)
     (map #(string/split (:full-name %) #" "))
     (map first))
```

It's also more common to see the first argument aligned with the thread macro. I actually thought zprint would format like this automatically.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -219,142 +150,198 @@
...
+                    (rf/dispatch [:chat.ui/navigate-to-chat-nav2 (str community-id id)])
+                    (rf/dispatch [:search/home-filter-changed nil]))}))))
+
+(defn add-on-press-handler-to-chats
+  [community-id chats]
+  (mapv (partial add-on-press-handler community-id) chats))
+
+(defn add-on-press-handler-to-categorized-chats
+  [community-id categorized-chats]
+  {}
```
ilmotta: This looks like a leftover.

ilmotta (approved):
@vkjr Excellent work with all the refactorings. I left a handful of comments, but I'm approving now to avoid blocking the PR in your timezone.

J-Son89 (approved):

src/status_im2/subs/communities.cljs
flexsurfer: yeah and its better to use categoryID as a key not a keyword("Name")

src/status_im2/contexts/communities/home/view.cljs
```diff
@@ -11,10 +11,7 @@
   [id]
   (let [community-item (rf/sub [:communities/home-item id])]
     [quo/communities-membership-list-item
-     {:on-press      (fn []
-                       (rf/dispatch [:communities/load-category-states id])
-                       (rf/dispatch [:dismiss-keyboard])
-                       (rf/dispatch [:navigate-to-nav2 :community {:community-id id}]))
+     {:on-press      #(rf/dispatch [:navigate-to :community-overview id])
```
flexsurfer, line 14: dismiss-keyboard removed intentionally?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -12,164 +12,95 @@
             [status-im2.contexts.communities.overview.style :as style]
             [status-im2.contexts.communities.menus.community-options.view :as options]
             [status-im2.contexts.communities.menus.request-to-join.view :as join-menu]
+            [status-im2.contexts.communities.overview.utils :as utils]
+            [cljs.core :as core]
```
flexsurfer: `[cljs.core :as core]` could you elaborate?

src/status_im2/contexts/communities/overview/view.cljs
flexsurfer: yes atoms should be outside of render function

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -12,164 +12,95 @@
...
+           on-first-channel-height-changed]}
+   channels-list]
+  (let [categories-heights (reagent/atom [])]
+    [rn/view
+     {:on-layout #(on-first-channel-height-changed (+ (if platform/ios?
+                                                        0
+                                                        38)
+                                                      (int (Math/ceil (layout-y %)))))
+      :style     {:margin-top 20 :flex 1}}
+     (map-indexed
```
flexsurfer, line 80: most likely it should be in doall form because there is a deref inside

flexsurfer: why not a flatlist btw ?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -219,142 +150,198 @@
...
-      [rn/view
-       [rn/view {:padding-horizontal 20}
+   scroll-height
+   {:keys [on-categories-heights-changed
+           on-first-channel-height-changed]}]
+  (let [pending?          (pos? requested-to-join-at)
+        thumbnail-image   (get-in images [:thumbnail])
+        chats-by-category (rf/sub [:communities/categorized-channels id])
+        users             (rf/sub [:communities/users id])]
+    [rn/view
```
flexsurfer, line 220: [:<>

flexsurfer (changes requested):

src/status_im2/common/scroll_page/view.cljs
```diff
@@ -108,7 +104,7 @@
          sticky-header
          children]
       [:<>
-       [scroll-page-header @scroll-height name page-nav-right-section-buttons cover-image sticky-header]
+       [:f> scroll-page-header @scroll-height name page-nav-right-section-buttons cover-image sticky-header]
```
flexsurfer: could you elaborate pls ?

vkjr: `:f>` it is a way to create a functional React component. Only in functional component hooks can be used. And hooks were introduced for animations here - #14695.
So in my changes, it is as a result of rebase.

flexsurfer: yeah sorry wrong place for comment, i wanted to ask about `display-picture` why f> was moved from function to outside usage of a component, what's the point

vkjr: Okay, full story :)

Original reference to `display-picture` was:
```clojure
[display-picture scroll-height cover]
```
where `scroll-height` is an atom. And inside `display-picture` there were a bunch of dereferences of that atom.

I don't like passing atoms inside components. Because react already has idiomatic ways to get info in and out of components - props and callbacks. Also passing atom in the props makes changes of atom by child components uncontrolled. 

So I removed atoms dereferencing within `display-picture` and changed its call:
```clojure
[display-picture @scroll-height cover]
```

And flickering started.
At that point `display-picture` looked like this:
```clojure
(defn display-picture
  [scroll-height cover]
  [:f>
   (fn []
     (let [input-range (if platform/ios? [-67 10] [0 150])
...
```
It is a functional react component inside one more component. When atom used as an argument, it wasn't technically changing as a prop, so outer component wasn't re-rendering content. But now, when value of `scroll-height` used and changes constantly `display-picture` started to recreate internal functional react component all the time. Flickering as a result.
And since there is no any reason in additional wrapper component, I got rid of it, leaving a function itself. That cures flickering. It rerenders, but not recreates

And I believe `[:f> display-picture @scroll-height cover-image]` is quite convenient, it makes it obvious that we are dealing with functional component here.

flexsurfer: how a dev might know that ` display-picture` can be used only with `:f>` ? anyway mixing reactive atoms and hooks not a good idea, but its not a discussion for this PR

flexsurfer: `:f>` should definitely be inside component, and we need to find a proper solution for `scroll-height` probably it needs to be reimplemented

vkjr: Good question! I think docstring can be a good solution.

ilmotta: Thanks for the detailed explanation @vkjr. I don't have much to add, but we can see in Reagent's unit tests and docs the usage of `:f>` outside the component. https://github.com/reagent-project/reagent/blob/master/doc/ReactFeatures.md#hooks

So to me this doesn't feel like a workaround in any way and I wouldn't even worry about the docstring. If we can, I agree we should use `:f>` inside components, but for maximum flexibility, as @vkjr demonstrated, using it outside can have benefits too.

vkjr: > `:f>` should definitely be inside component, and we need to find a proper solution for `scroll-height` probably it needs to be implemented

`:f>` makes a component on its own. So it is not really "outside" of it. If we wrap it in another component - we will have 2 components) and no additional value.

I don't really see any problem in using `:>f` outside of the component, to be honest. Even if some developer will use `display-picture` without it, it will get a quite clear error explaining that he is trying to use hooks outside of the functional component. And since this is not even a part of the quo2 library there aren't many chances that it will be ever reused.

flexsurfer: ok so i read again the comment, and if you want to use value only as the initial value only once you shouldn't use it in a render function
```clojure
(defn display-picture
  [scroll-height cover]
 (fn [_ _ ]
  [:f>
   (fn []
     (let [input-range (if platform/ios? [-67 10] [0 150])
```
so now display-picture returns render function but we use `scroll-height` only once when component was initialized

src/status_im2/contexts/communities/overview/view.cljs
vkjr: Here the behavior is intentional. We need `categories-heights` to be an atom, to store data after consecutive `on-layout` calls. But we need it to be recreated on rerender, because when`channels-list` property updated, all values of `categories-heights` should be discarded.

src/status_im2/contexts/communities/home/view.cljs
vkjr: yes, I believe it is a leftover from old pages where the "find" section was present and the user could have a keyboard opened when opening community. Not really needed here

src/status_im2/contexts/communities/overview/view.cljs
vkjr: leftover, thanks!

src/status_im2/contexts/communities/overview/view.cljs
vkjr: Not sure if `doall` is needed. Atom derefs in `:on-layout` callback, so it will be at the moment when item already realized. Or there is something else?

Not sure why not a FlatList.. it is an inherited piece of code, not written by me.

src/status_im2/common/scroll_page/view.cljs
vkjr: We actually do need the `scroll-height` to be updated constantly, it is not enough to initialize it once. But we don't need the component to be re-created when `scroll-height` changed, it should be only re-rendered.
And as far as I can tell, old form creates full component recreation:
```clojure
(defn display-picture
  [scroll-height cover]
  [:f>
   (fn []
     (let [input-range (if platform/ios? [-67 10] [0 150])
```

And the form you suggested won't react on `scroll-height` changes... and it is quite heavy for comprehension, I'd say.

flexsurfer: > and it is quite heavy for comprehension, I'd say.

it's a general approach in reagent, so if you're familiar with reagent it should be fine, more of that it's transparent, we clearly see that we don't want to use updated value in the render function

src/status_im2/contexts/communities/overview/view.cljs
ilmotta: Missed that, thanks @vkjr.

Upon further inspection, I can see this entire namespace is doing some rather inefficient stuff. We're doing a `swap!` on a reactive atom on every element of the `channels-list`, and then after every `swap!` another `swap!` after triggering `on-categories-heights-changed`. It looks like a lot of re-rendering work that doesn't scale well with the size of the list.

Unfortunately I don't any practical suggestion. You probably thought about this already.

src/status_im2/common/scroll_page/view.cljs
flexsurfer: honestly for me `[:f> display-picture @scroll-height cover-image]` looks like a hack, do we understand why `:f>` doesn't work inside `display-picture` ?  we need to understand that so we won't have this later in the code

flexsurfer: @ilmotta do you have any thoughts ?

ilmotta: @flexsurfer, not really unfortunately. As I said, I don't have much to offer here... I'd need to spend time in the code to see if there's anything to be done. To me, as long as we treat `:f>` outside functions with hooks as the exception, I think we're fine. I take that this is an exception and that it leads to a better design from @vkjr's perspective.

Using `:f>` outside functions is relatively common tbh (not in our codebase though).

flexsurfer: yeah we don't have much experience with :f> so i just feel we might miss something important, performance issues etc , agree that to find out we need to spend time and understand whats wrong with this code

src/status_im2/contexts/communities/overview/view.cljs
flexsurfer: there will be a warning in metro logs if you need to use doall, but i would add it anyway

flexsurfer: but it seems like flatlist should be used here, but not related to this PR

flexsurfer (approved):

src/status_im2/contexts/communities/overview/view.cljs
vkjr: There are really a lot of questionable things in this namespace from the very beginning. My main suggestion is to fix them step by step, because otherwise, I will never close this PR :)))

vkjr:
@flexsurfer, @ilmotta, I addressed more reviews and rebased one more time) 
I'd be really happy if we can merge this PR! Rebasing it becomes really problematic with these changes)

flexsurfer:
@vkjr pls take a look at lint and tests

vkjr:
> @vkjr pls take a look at lint and tests

of course)

src/status_im2/contexts/communities/overview/view.cljs
ilmotta: You're absolutely correct @vkjr. Baby steps! Or baby reps?

https://www.youtube.com/watch?v=Nnksg20hhvw

ilmotta:
> @flexsurfer, @ilmotta, I addressed more reviews and rebased one more time) I'd be really happy if we can merge this PR! Rebasing it becomes really problematic with these changes)

@vkjr: I'm totally fine with a merge. Let's improve later as you said. Thank you for the patience in the discussions.

src/status_im2/common/scroll_page/view.cljs
J-Son89: yep, and in this case it's both cleaner code and improves performance as it stops a lot of unnecessary rendering which was happen. (I did not benchmark this, but just assuming from how the screen was previously looking compared  with this pr)

pavloburykh:
@vkjr thanx for the PR. As current changes remove old community screens, it brakes current e2e which were based on those old screens. 

So we need to wait with merging of this PR until we refactor e2e according to a new flow introduced by this PR. @churik will take it into work ASAP and give an update.

vkjr:
@pavloburykh, I see, thanks for the explanation!
Okay, will merge when e2e ready)

J-Son89:
> @vkjr thanx for the PR. As current changes remove old community screens, it brakes current e2e which were based on those old screens.
> 
> So we need to wait with merging of this PR until we refactor e2e according to a new flow introduced by this PR. @churik will take it into work ASAP and give an update.

@pavloburykh, @churik  should we really block this because the tests are not ready? It seems more practical to merge this and then follow up with the e2e test adjustments since they are technically "new" tests. 
These changes are needed for the codebase and until they are in it is blocking changes to this file/pages, e.g
https://github.com/status-im/status-mobile/pull/14761

it's possible that we work on this file/pages but it will lead to a lot of tedious rebasing for @vkjr.

churik:
OK, go ahead with merging if it is required for other PRs. I'll get to fixing e2e asap in separate PR
cc @vkjr @J-Son89

vkjr:
Thank you!

PR #14803, opened 2023-01-16
Add Collectible ui component

### LINKED ISSUE #14775: Implement algorithm for tiling images for collectibles
We need to implement the component below.

The scope of this task is to implement the behavior when displaying tiles (image only in the design).
You can use the test images taken from Figma from this, ultimately they will be passed to the component as a URL.

Tiles have all the same size.

Depending on how many images the component is passed, it should display them according to the design, for example:

1 Image: The image will occupy the whole tile
2 Images: The two images should share the space within the tile: https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=2010%3A35736&t=QtTqoPpY029TXvwX-4

and so on

If there are 8 or more images, the design is https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=2010%3A31247&t=QtTqoPpY029TXvwX-4 . `+x` indicates how many images are not displayed within the collection.

Out of scope:
Text component (Doodle #...)
and interactions/animations

### Designs
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=15671%3A187484&t=5oN55S16EhsyENzY-0

ulisesmac:

Fixes #14775

### Summary

This PR implements the [Collectible tiles component](https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=2010%3A35742&t=qePMYYJz53KsjrEW-4) and creates a test screen.

### Review notes
The shadows behind the tiles (which are barely noticeable), such as the following behind the yellow square:

image

where not added, since they don't look as the figma's design on Android devices when using  the `elevation` property, and I don't have an iPhone to test the `shadow-*` props.

#### Component preview

image image image 

image image image

image image 

##### Dark theme 
image image

### Testing notes
The test screen can be visited in the route `Preview Quo2 Components -> Profile -> Collectible`.
Links from https://opensea.io/ were added in order to test the component with images as URLs (as mentioned in #14775).

#### Platforms

- Android
- iOS

##### Functional
- dapps / app browsing

### Steps to test
- Open Status
- Naviage to  `Preview Quo2 Components -> Profile -> Collectible`
- Press the select "Number of images" to make the component render the selected number of images
- Press on "Random order of images" to shuffle the images shown.

status: ready


src/quo2/components/profile/collectible/view.cljs
```diff
@@ -0,0 +1,67 @@
+(ns quo2.components.profile.collectible.view
+  (:require [quo2.components.markdown.text :as text]
+            [quo2.components.profile.collectible.style :as style]
+            [react-native.core :as rn]))
+
+(defn plus-n-tile
+  [n]
```
briansztamfater: I recommend using more descriptive parameter names, `n` is hard to understand at a first glance.

flexsurfer: imo here it's fine no ? plus-n function and we pass n

src/quo2/components/profile/collectible/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+   [rn/view {:style style/top-left}
+    [tile {:uri (first imgs) :size size}]]
+   [rn/view {:style style/bottom-right}
+    [tile {:uri (second imgs) :size size}]]])
+
+(defn three-tiles
+  [{:keys [tiles size]}]
+  (let [[img-1 img-2 img-3 & _] tiles]
+    [:<>
+     [rn/view {:style style/top-left}
```
flexsurfer: we could pass style as parameter to tile, so we avoid using containers views

flexsurfer (approved):
looks good to me, thank you

Parveshdhull (approved):
Great work!

src/status_im2/contexts/quo_preview/profile/collectible.cljs
```diff
@@ -0,0 +1,55 @@
...
+                     :value (str n " images")})
+                  (range 1 11))}
+   {:label "Random order of images"
+    :key   :shuffle-imgs
+    :type  :boolean}])
+
+(defn cool-preview
+  []
+  (let [state       (reagent/atom {:num-imgs 1 :shuffle-imgs false})
+        test-images ["https://i.seadn.io/gcs/files/ddefc5cbae2a8bc5d5aa368ca5887a23.jpg"
```
flexsurfer: oh its better to do not use external resources

J-Son89:
can you add some screenshots of the component preview to this pr?

src/quo2/components/profile/collectible/view.cljs
briansztamfater: Yeah I would rename the function name too, but just a detail to make it more understandable, not a big deal

src/quo2/components/profile/collectible/view.cljs
```diff
@@ -0,0 +1,67 @@
...
+      [tile {:uri img-1 :size size}]]
+     [rn/view {:style style/top-right}
+      [tile {:uri img-2 :size size}]]
+     [rn/view {:style style/bottom-left}
+      [tile {:uri img-3 :size size}]]]))
+
+(defn tile-container
+  [{:keys [imgs]}]
```
ilmotta: I would prefer if we didn't shorten certain names, because in this PR they affect readability for no actual gain.

- `imgs` => `images`
- `rem-imgs` => `remaining-images`
- `first-3-imgs` => `first-3-images`

src/status_im2/contexts/quo_preview/profile/collectible.cljs
```diff
@@ -0,0 +1,55 @@
...
+          {:padding-vertical 100
+           :align-items      :center}
+          [quo/collectible {:imgs imgs-to-show}]]]))))
+
+(defn preview-collectible
+  []
+  [rn/view {:style {:flex 1}}
+   [rn/flat-list
+    {:flex                      1
+     :keyboardShouldPersistTaps :always
```
ilmotta: I believe there's no need to use camel case for top level keys passed to components transformed via `reagent/adapt-react-class` (which is the case for `rn/flat-list`).

ilmotta (approved):

briansztamfater (approved):

src/quo2/components/profile/collectible/view.cljs
flexsurfer: agree

briansztamfater: Also I would prefer `first-three-images` instead of `first-3-images`

src/status_im2/contexts/quo_preview/profile/collectible.cljs
ulisesmac: Ok! What links should I use instead?

Parveshdhull: we can use same images used in designs and put them in mocks folder
https://github.com/status-im/status-mobile/tree/develop/resources/images/mock

src/quo2/components/profile/collectible/view.cljs
ulisesmac: ok! renamed to `remaining-tiles` and the parameter is `amount` in 08c4d35e0d452ef73b91482837eb8d67721542d2

src/quo2/components/profile/collectible/view.cljs
ulisesmac: Got it! it's in fact a lot better, thanks!

It has been done in `110608d493491317bedc25bfa700609323fc0164`

src/status_im2/contexts/quo_preview/profile/collectible.cljs
ulisesmac: The change has been applied in `3af349471929f98c6347b00e4ecc0bf2ffe80471`. But it  required some significant changes:

1. An URI is a string, an internally loaded resource is a path that has to be passed to `js/require` which returns an integer. This component  was only able to receive URIs, so now it's able to work with internal resources.

2. I was naming `uri` to the paramater received by the `tile` component, now it's called `resource`.

3. The files exported by figma have different sizes and a sort of  transparent paddings. e.g.
image  image  image

and, as you can see, the more images the component has, the smaller they are exported.

So I decided to only export the first one, edit it using Krita to remove the transparent padding and repeat it the requested times in the preview screen. (I've just updated the PR description, so you can see how it looks).

Also, I didn't remove the shuffle option in the preview screen, since maybe in the future we'd like to add more images to test this component.

ulisesmac:
> can you add some screenshots of the component preview to this pr?

It has been done! :)

src/quo2/components/profile/collectible/view.cljs
ulisesmac: Completely agree!

The change is in b1f93d96cc745575eea7114b68b271a28690ba83

src/status_im2/contexts/quo_preview/profile/collectible.cljs
ulisesmac: Yep, you're right.

I'm not sure how exactly is the preview module implemented, so I was copying the structure already used in other preview screens, and this is something I didn't notice. The change is in 2e25e76cc0a31eebad5d183bbb8963d3a8a02681

ulisesmac:
Thanks to everyone for your review!
I applied corrections in individual commits to make the review easier

Please, let me know when it's ready so I can squash the commits.

src/status_im2/contexts/quo_preview/profile/collectible.cljs
ilmotta: Yeah @ulisesmac, I see other places unnecessarily using camel case, but good thing you changed in your PR, thanks. In fact, as you probably noticed, all preview screens are very similar and there's a lot of copy and paste going on to set up each one. Eventually we should be able to abstract the preview screens a bit better.

src/quo2/components/profile/collectible/view.cljs
```diff
@@ -0,0 +1,99 @@
...
+            [:<>
+             [three-tiles
+              {:tiles (take 3 remaining-images)
+               :size  :xs}]
+             [remaining-tiles (- (count remaining-images) 3)]])]]))))
+
+(defn collectible
+  [{:keys [images]}]
+  [rn/view {:style style/tile-outer-container}
+   [rn/view {:style style/tile-inner-container}
```
flexsurfer, line 96: just wondering why it can't be only one container ?

src/quo2/components/profile/collectible/view.cljs
```diff
@@ -0,0 +1,99 @@
...
+  [rn/view {:style (merge style/bottom-right (style/remaining-tiles))}
+   [text/text
+    {:style  (style/remaining-tiles-text)
+     :size   :paragraph-2
+     :weight :medium}
+    (str "+" amount)]])
+
+(defn tile
+  [{:keys [style resource size]}]
+  (let [source (cond->> resource
```
flexsurfer: this pretty hard to read
```clojure 
(if (string? resource) {:uri resource} resource)
```
regural if is more readable imo

flexsurfer (approved):

src/quo2/components/profile/collectible/view.cljs
ulisesmac: That's because the outer one sets the container of the whole component (176 x 176):
image

and sets a padding of 8:
image

The inner one contains the tile or tiles, which are defined with an absolute position (styles `top-left`, `top-right`, `bottom-`...) to easliy create things like this:

top-left + bottom-right:
image

top-left + top-right + bottom-left
image

If only the outer container exist and as I'm using absolute positions, the padding previously set (8px) would be ignored.

A solution could be add those paddings to the absolute styles, for example, instead of:
```clojure
(def bottom-left
  {:position :absolute
   :bottom   0
   :left     0})
```
do:
```clojure
(def bottom-left
  {:position :absolute
   :bottom   8
   :left     8})
   ```
   
but that wouldn't allow us to reuse the same styles in the smaller tiles and look the same as the design: 
   
image
image

Hope this is clear, let me know if any additional explanation is required!

File: src/quo2/components/profile/collectible/view.cljs (hunk 7 above)
Comment (ulisesmac) [reply src/quo2/components/profile/collectible/view.cljs]: Agree, I changed it.

Also, just wanted to comment that IMO the `cond->` is more readable when it's aligned in pairs:
```clojure
(cond-> v
  a fn-1
  b fn-2
  c fn-3)
```
But the lint-fix is aligning it as follows:
```clojure
(cond-> v
        a
        fn-1
        b
        fn-2
        c
        fn-3)
```

which is a little confusing.

ulisesmac:
I've squashed everything into a single commit and rebased the branch on develop.

XxSNiPxX:
Is this component done?

flexsurfer:
yes

PR #14808, opened 2023-01-17
cleaning

flexsurfer:
cleaning (removed unused files in status-im, mostly for chat context), introduce react-native.red-black-tree and move messages list events
image


ilmotta:
Awesome stuff here, thank you! I'll review this PR later today @flexsurfer.

src/mocks/js_dependencies.cljs
```diff
@@ -1,8 +1,7 @@
 (ns mocks.js-dependencies
   (:require-macros [status-im.utils.slurp :refer [slurp]])
   (:require [status-im.fleet.default-fleet :refer (default-fleets)])
-  (:require [status-im.utils.test :as utils.test])
-  (:require [status-im.chat.default-chats :refer (default-chats)]))
+  (:require [status-im.utils.test :as utils.test]))
```
ilmotta, line 4: Would be nice to move this `utils.test` namespace somewhere else. This file has side-effects even to load it, and we can see from the only 2 usages in the codebase, that it should be required only once.

A better place for this file is in `status-im2/setup`, where we have defined in the guidelines to put some files related to raw plumbing. Another alternative would be to move it to `src/test_helpers/` directory because the integration tests use the `init!` function to define poll for signals.

flexsurfer:
> Awesome stuff here, thank you! I'll review this PR later today @flexsurfer.

there is not much to review, i just removed a lot of old code, and moved one namespace, just moved rn library in separate ns

src/react_native/red_black_tree.cljs
```diff
@@ -0,0 +1,51 @@
+(ns react-native.red-black-tree
+  (:refer-clojure :exclude [remove update find])
+  (:require ["functional-red-black-tree" :as rb-tree]))
```
ilmotta: In the spirit of our guidelines, perhaps we should use the alias as `red-black-tree`, there's no need to shorten the alias.

src/react_native/red_black_tree.cljs
```diff
@@ -0,0 +1,51 @@
...
+
+(def tree ^js rb-tree)
+
+(defn find
+  [^js tree item]
+  (.find tree item))
+
+(defn insert
```
ilmotta: Really nice to see this refactor, but IMO this namespace has nothing to do with React Native. Putting it inside `react_native/` implies there's some relationship, but there's none. How about we move it to `src/utils/red_black_tree.cljs`?

src/status_im/chat/models/message_test.cljs
```diff
@@ -1,148 +0,0 @@
-(ns status-im.chat.models.message-test
```
ilmotta: Why are we removing this test file? Shouldn't we adapt it?

src/status_im2/contexts/chat/messages/list/events.cljs
```diff
@@ -1,12 +1,12 @@
-(ns status-im.chat.models.message-list
-  (:require ["functional-red-black-tree" :as rb-tree]
-            [status-im.constants :as constants]
+(ns status-im2.contexts.chat.messages.list.events
+  (:require [react-native.red-black-tree :as rb-tree]
```
ilmotta: I think we should stick to our convention of preferring explicit namespace aliases, so in this case: `red-black-tree` instead of `rb-tree`.

ilmotta (approved):
@flexsurfer, although it's pretty much impossible to review a PR of this size, it looks good to me :rocket: 

I have left a couple comments that I would like to see addressed, but I'm okay if you need to merge this PR faster to avoid conflicts and work on the changes in a separate PR.

src/react_native/red_black_tree.cljs
flexsurfer: really good point thanks

src/mocks/js_dependencies.cljs
flexsurfer: i agree, but its kinda not related to this PR, i just removed one require line here

flexsurfer: i was trying to keep this PR only for chat context

src/status_im/chat/models/message_test.cljs
flexsurfer: mhm good question

src/mocks/js_dependencies.cljs
ilmotta: Agree, this is more of tangential point.

src/status_im/chat/models/message_test.cljs
flexsurfer: removed it by mistake, thanks. reverted

ilmotta: Thanks Andrey

PR #14821, opened 2023-01-18
move constants/config to status-im2 root and remove old constants/config

flexsurfer:
image


Parveshdhull (approved):

ilmotta (commented):
@flexsurfer:

The idea of the `setup` directory is to put stuff that nobody mostly cares and that almost no other namespace should require. For example, things that often need to run only once to set up the app.

From the guidelines:

>`src/status_im2/setup/`: Contains namespaces that are mostly used to initialize the application, configure test runners, etc. In general, such namespaces should not be required from the outside.

The `constants` namespace is actually used throughout the whole app, so I actually think it should live outside the `setup/` dir. Where? I'm not sure, but I kind of like the way it is now, inside `status-im2.common`.

The `constants` namespace is actually quite *common*, so it kind of fits well with the `status-im2.common` directory :man_shrugging:

flexsurfer:
> @flexsurfer:
> 
> The idea of the `setup` directory is to put stuff that nobody mostly cares and that almost no other namespace should require. For example, things that often need to run only once to set up the app.
> 
> From the guidelines:
> 
> > `src/status_im2/setup/`: Contains namespaces that are mostly used to initialize the application, configure test runners, etc. In general, such namespaces should not be required from the outside.
> 
> The `constants` namespace is actually used throughout the whole app, so I actually think it should live outside the `setup/` dir. Where? I'm not sure, but I kind of like the way it is now, inside `status-im2.common`.
> 
> The `constants` namespace is actually quite _common_, so it kind of fits well with the `status-im2.common` directory 🤷‍♂️

yeah i agree, but in recent time i needed to change one constant and i went to setup folder, because constants are like setup values for the app, we set them once, so for me setup is more natural for constants, but i agree, probably we need a vote? because im fine to keep it common, but dunno what other think

flexsurfer:
we have config in setup, so constants in some areas part of config, so idk, wdyt @Parveshdhull @J-Son89 @yqrashawn ?

flexsurfer:
`such namespaces should not be required from the outside.` mhm, but config required as well, so probably constants and config should go into status-im2 root ? wdyt

flexsurfer:
yeah i like `status-im2.config` and `status-im2.constants`

J-Son89:
+1 for status-im2.config and status-im2.constants

ilmotta:
> `such namespaces should not be required from the outside.` mhm, but config required as well, so probably constants and config should go into status-im2 root ? wdyt

It's not a hard rule @flexsurfer, that's why it's using the word `should`. We can have exceptions. Let's also consider that some stuff inside the `setup` dir are left overs from our on going code migrations. For example:

- The namespace `setup/datetime` shouldn't be there (it's a co-effect, should be somewhere else)
- Log namespace has mixed stuff, like re-frame effects (should be somewhere else).
- `setup/config` is suspicious.

And so on. Now that I see this `setup` dir, I think it's not the best of the examples we have, it's contradicting the guidelines as of today.

>yeah i like status-im2.config and status-im2.constants

Let's go for it!

ilmotta:
> yeah i agree, but in recent time i needed to change one constant and i went to setup folder, because constants are like setup values for the app, we set them once, so for me setup is more natural for constants, but i agree, probably we need a vote? because im fine to keep it common, but dunno what other think

@flexsurfer Whatever we decide we just need to be consistent and update the guidelines. If the guidelines are open for too much interpretation, then it's value is diminished, so I think we should try to reword to be more precise.

Another thing, I've been thinking about adding README files on top level directories to explain their purpose and any other aspects of code organization. For example, a README inside `quo2`, a README inside `status_im2`, and so on. This is a common practice in many projects. We do follow this practice in the `nix/deps/` directory for example.

The reasoning is that this would help collaborators understand the structure of the project.

flexsurfer:
> * The namespace `setup/datetime` shouldn't be there (it's a co-effect, should be somewhere else)

yeah and i've already moved it :)

flexsurfer:
ok thank you all
moved config and constants to  status-im2 root
removed old config and constants

PR #14824, opened 2023-01-18
[Fix] Admin Notification marked unread after closing and reopening AC

### LINKED ISSUE #14823: Admin Notifications are marked unread after closing and reopening Activity Center
# Bug Report

## Problem 

Admin Notifications that are marked `read` by performing any action (Accept/Decline) are marked as `unread` after the AC is closed and reopened.

#### Expected behaviour

The Admin notifications should be marked as read if the user performs any action (Accept / Decline) on the notification

#### Actual behaviour

The Admin Notifications are marked as unread if we close the Activity Centre and reopen it again.

### Reproduction

1) Received an Admin notification on Activity Center
2) Accept / Decline the notification
3) Notification will be marked as read until you close the Activity Center
4) Close and reopen the Activity Center
5) The admin notification which was Accepted/Declined would be marked unread.

### Additional Information

- Status Version: nightly 

smohamedjavid:
fixes #14823

### DIFF
--- a/src/status_im/communities/core.cljs
+++ b/src/status_im/communities/core.cljs
@@ -555,14 +555,18 @@
   [community-id]
   (re-frame/dispatch [::fetch-requests-to-join community-id]))
 
+(defn update-activity-center-notification
+  [id]
+  [{:dispatch [:activity-center.notifications/mark-as-read id]
+    :ms       100}])
+
 (rf/defn request-to-join-accepted
   {:events [::request-to-join-accepted]}
   [{:keys [db] :as cofx} community-id request-id response-js]
   (rf/merge
    cofx
    {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
-    :dispatch-later [{:dispatch [:activity-center.notifications/mark-as-read request-id]
-                      :ms       100}]}
+    :dispatch-later (update-activity-center-notification request-id)}
    (handle-response response-js)))
 
 (rf/defn request-to-join-declined
@@ -571,8 +575,7 @@
   (rf/merge
    cofx
    {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
-    :dispatch-later [{:dispatch [:activity-center.notifications/mark-as-read request-id]
-                      :ms       100}]}
+    :dispatch-later (update-activity-center-notification request-id)}
    (handle-response response-js)))
 
 (rf/defn accept-request-to-join-pressed
--- a/src/status_im/communities/core.cljs
+++ b/src/status_im/communities/core.cljs
@@ -555,28 +555,23 @@
   [community-id]
   (re-frame/dispatch [::fetch-requests-to-join community-id]))
 
-(defn update-activity-center-notification
-  [id]
-  [{:dispatch [:activity-center.notifications/mark-as-read id]
-    :ms       100}])
-
 (rf/defn request-to-join-accepted
   {:events [::request-to-join-accepted]}
   [{:keys [db] :as cofx} community-id request-id response-js]
   (rf/merge
    cofx
-   {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
-    :dispatch-later (update-activity-center-notification request-id)}
-   (handle-response response-js)))
+   {:db         (update-in db [:communities/requests-to-join community-id] dissoc request-id)
+    :dispatch-n [[:sanitize-messages-and-process-response response-js]
+                 [:activity-center.notifications/mark-as-read request-id]]}))
 
 (rf/defn request-to-join-declined
   {:events [::request-to-join-declined]}
   [{:keys [db] :as cofx} community-id request-id response-js]
   (rf/merge
    cofx
-   {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
-    :dispatch-later (update-activity-center-notification request-id)}
-   (handle-response response-js)))
+   {:db         (update-in db [:communities/requests-to-join community-id] dissoc request-id)
+    :dispatch-n [[:sanitize-messages-and-process-response response-js]
+                 [:activity-center.notifications/mark-as-read request-id]]}))
 
 (rf/defn accept-request-to-join-pressed
   {:events [:communities.ui/accept-request-to-join-pressed]}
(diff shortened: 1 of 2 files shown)


ilmotta:
@smohamedjavid, do you know if this is happening to other notifications?

Something tells me the proper solution is different, and that it would solve this *ordering* of events issue. Here's my proposal: we need to *order the events*, i.e. make sure marking as read only happens after accept/decline succeeds. You'll need to adapt the event handlers to accept a `on-success` (optional) argument.

This solution would have the added benefit of making the UX more reliable, because the app would never mark a notification as read if the previous update on status-go didn't succeed.

There should be no need for a workaround like this one. We should avoid this `js/setTimeout` trick as much as possible. I see many parts of the code resorting to this technique, but there's often a better way.

smohamedjavid:
> @smohamedjavid, do you know if this is happening to other notifications?
> 
> Something tells me the proper solution is different, and that it would solve this _ordering_ of events issue. Here's my proposal: we need to _order the events_, i.e. make sure marking as read only happens after accept/decline succeeds. You'll need to adapt the event handlers to accept a `on-success` (optional) argument.
> 
> This solution would have the added benefit of making the UX more reliable, because the app would never mark a notification as read if the previous update on status-go didn't succeed.
> 
> There should be no need for a workaround like this one. We should avoid this `js/setTimeout` trick as much as possible. I see many parts of the code resorting to this technique, but there's often a better way.

As far as I checked, the admin notification has this issue. 
The rest of the notifications don't have this issue if they get marked as read.

I agree that the `js/setTimeout` is not a good solution and could make a notification read if the status-go didn't succeed. I removed the `js/setTimeout` workaround and dispatched the `mark-as-read` event on the `on-success` functions.

Thanks for the valuable feedback @ilmotta 💯

src/status_im/communities/core.cljs
```diff
@@ -558,16 +558,22 @@
...
 (rf/defn request-to-join-declined
   {:events [::request-to-join-declined]}
   [{:keys [db] :as cofx} community-id request-id response-js]
-  (rf/merge cofx
-            {:db (update-in db [:communities/requests-to-join community-id] dissoc request-id)}
-            (handle-response response-js)))
+  (rf/merge
+   cofx
+   {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
+    :dispatch-later [{:dispatch [:activity-center.notifications/mark-as-read request-id]
```
flexsurfer: could mark-as-read be used here directly without dispatches? or at least `dispatch-later` looks redundant here

flexsurfer: still open question

smohamedjavid: `mark-as-read` could not be used directly. 

Cleaned up the redundant code in `dispatch-later` 👍.  These events (`accept-request-to-join-pressed` & `decline-request-to-join-pressed`) are used in communities (apart from Activity Center).

src/status_im/communities/core.cljs
```diff
@@ -555,19 +555,28 @@
...
 (rf/defn request-to-join-declined
   {:events [::request-to-join-declined]}
   [{:keys [db] :as cofx} community-id request-id response-js]
-  (rf/merge cofx
-            {:db (update-in db [:communities/requests-to-join community-id] dissoc request-id)}
-            (handle-response response-js)))
+  (rf/merge
+   cofx
+   {:db             (update-in db [:communities/requests-to-join community-id] dissoc request-id)
+    :dispatch-later (update-activity-center-notification request-id)}
```
flexsurfer: are you sure later is still required ? it could be just `:dispatch [:activity-center.notifications/mark-as-read id]`

jo-mut (approved):

src/status_im/communities/core.cljs
smohamedjavid: Yes, @flexsurfer - I tried those as well initially when I work on moving the event to `on-success` arguments. No luck in fixing the bug. A slight delay in dispatching the `:activity-center.notifications/mark-as-read` works.

flexsurfer: thats weird

flexsurfer: @ilmotta any thoughts?

flexsurfer: @smohamedjavid could you pls rollback then, because no need to hide dispatch in the function

flexsurfer (changes requested):
we need to find a proper solution, dispatch-later shouldn't be used

src/status_im/communities/core.cljs
ilmotta: I'm honestly not sure, but the dispatch later shouldn't be necessary as @flexsurfer said.

Well, I have a guess @smohamedjavid, but it's hard without trying some things out, and it's hard to explain. Let me try...

The event to mark a notification as *read* receives a notification ID, and then get it from app DB (https://github.com/status-im/status-mobile/blob/develop/src/status_im2/contexts/activity_center/events.cljs#L131). The fact that it gets from the app DB means that it could be getting an outdated reference to the notification map while the other event hasn't completed. When you use `dispatch-later` you're giving re-frame a chance to update the app DB first.

So it's likely something to do with the fact that these events are seeing a different state stored in the app DB, and that's why one seem to replace the other's work and thus cause the UI bug.

ilmotta: I think it would be valuable to understand why this problem happens, otherwise we're missing the chance of understanding something quite fundamental in terms of state management. I'm super curious too.

smohamedjavid:
@flexsurfer @ilmotta -  The event `:sanitize-messages-and-process-response` (community response - which updates the AC notifications for membership status) needs to be completed before we dispatch `mark-as-read` as it takes from the app DB state. Re-implemented the solution using `dispatch-n` for ordering events.

flexsurfer (approved):

ilmotta:
> @flexsurfer @ilmotta - The event `:sanitize-messages-and-process-response` (community response - which updates the AC notifications for membership status) needs to be completed before we dispatch `mark-as-read` as it takes from the app DB state. Re-implemented the solution using `dispatch-n` for ordering events.

Most excellent @smohamedjavid, thanks for trying different solutions.

ilmotta (approved):

J-Son89 (approved):

pavloburykh:
@smohamedjavid thanx fro the fixe. Tested, ready to be merged.

PR #14831, opened 2023-01-19
chore: fix component tests and permission drawer preview

J-Son89:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,6 @@ component-test: export COMPONENT_TEST := true
 component-test: export BABEL_ENV := test
 component-test: ##@test Run tests once in NodeJS
 	# Here we create the gyp bindings for nodejs
-	yarn install
 	yarn shadow-cljs compile component-test && \
 	jest --config=test/jest/jest.config.js
 
--- a/src/quo2/components/drawers/action_drawers/component_spec.cljs
+++ b/src/quo2/components/drawers/action_drawers/component_spec.cljs
@@ -1,6 +1,6 @@
-(ns quo2.components.drawers.--tests--.action-drawers-component-spec
+(ns quo2.components.drawers.action-drawers.component-spec
   (:require ["@testing-library/react-native" :as rtl]
-            [quo2.components.drawers.action-drawers :as action-drawer]
+            [quo2.components.drawers.action-drawers.view :as action-drawer]
             [reagent.core :as reagent]))
 
 (defn render-action-drawer
--- a/src/quo2/components/drawers/action_drawers/style.cljs
+++ b/src/quo2/components/drawers/action_drawers/style.cljs
@@ -1,0 +1,41 @@
+(ns quo2.components.drawers.action-drawers.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def divider
+  {:border-top-width 1
+   :border-top-color (colors/theme-colors
+                      colors/neutral-10
+                      colors/neutral-90)
+   :margin-top       8
+   :margin-bottom    7
+   :align-items      :center
+   :flex-direction   :row})
+
+(defn container
+  [sub-label]
+  {:border-radius     12
+   :height            (if sub-label 58 50)
+   :margin-horizontal 8})
+
+(defn row-container
+  [sub-label]
+  {:height            (if sub-label 58 50)
+   :margin-horizontal 12
+   :flex-direction    :row})
+
+(def left-icon
+  {:height        20
+   :margin-top    :auto
+   :margin-bottom :auto
+   :margin-right  12
+   :width         20})
+
+(def text-container
+  {:flex            1
+   :justify-content :center})
+
+(def right-icon
+  {:height        20
+   :margin-top    :auto
+   :margin-bottom :auto
+   :width         20})
--- a/src/quo2/components/drawers/permission_context/component_spec.cljs
+++ b/src/quo2/components/drawers/permission_context/component_spec.cljs
@@ -1,10 +1,10 @@
-(ns quo2.components.drawers.permission-context.--tests--.permission-context-component-spec
+(ns quo2.components.drawers.permission-context.component-spec
   (:require [quo2.components.drawers.permission-context.view :as permission-context]
             [react-native.core :as rn]
             [test-helpers.component :as h]))
 
 (h/describe "permission context"
-  (h/test
+  (h/test "it tests the default render"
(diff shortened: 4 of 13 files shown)


Makefile
```diff
@@ -351,7 +351,6 @@ component-test: export COMPONENT_TEST := true
 component-test: export BABEL_ENV := test
 component-test: ##@test Run tests once in NodeJS
 	# Here we create the gyp bindings for nodejs
-	yarn install
```
J-Son89, line 354: running yarn every time is 🤯

src/quo2/components/drawers/action_drawers/component_spec.cljs
```diff
@@ -1,6 +1,6 @@
-(ns quo2.components.drawers.--tests--.action-drawers-component-spec
+(ns quo2.components.drawers.action-drawers.component-spec
```
J-Son89, line 1: aligned with best practices

src/quo2/components/drawers/action_drawers/style.cljs
```diff
@@ -0,0 +1,41 @@
+(ns quo2.components.drawers.action-drawers.style
```
J-Son89, line 1: addressing best practices with separate file for style

src/quo2/components/drawers/permission_context/component_spec.cljs
```diff
@@ -1,10 +1,10 @@
-(ns quo2.components.drawers.permission-context.--tests--.permission-context-component-spec
+(ns quo2.components.drawers.permission-context.component-spec
   (:require [quo2.components.drawers.permission-context.view :as permission-context]
             [react-native.core :as rn]
             [test-helpers.component :as h]))
 
 (h/describe "permission context"
-  (h/test
+  (h/test "it tests the default render"
```
J-Son89, line 7: test was broke as it did not have a description string

src/quo2/components/drawers/permission_context/style.cljs
```diff
@@ -3,9 +3,13 @@
 
 (def radius 20)
 (def container
-  {:padding-top             16
+  {:flex-direction          :row
```
J-Son89, line 6: this component was missing some properties needed it for it to match designs easier

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
```diff
@@ -0,0 +1,75 @@
+(ns status-im2.contexts.quo-preview.drawers.permission-drawers
+  (:require [quo2.core :as quo]
```
J-Son89, line 2: this component was missing a preview. Discussed with @cammellos as it is unclear of what this component should be more so than a styled container. 
I included some examples so we can see how to get the designs from this at the very least. 👍

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
```diff
@@ -0,0 +1,75 @@
...
+     [rn/view {:margin-top 60}
+      [quo/permission-context [example-3] #(js/alert "drawer pressed")]]]))
+(defn preview-permission-drawers
+  []
+  [rn/view
+   {:background-color (colors/theme-colors colors/neutral-20 colors/neutral-100)
+    :flex             1}
+   [rn/flat-list
+    {:flex                      1
+     :nestedScrollEnabled       true
```
ilmotta: FYI: You can use kebab case just fine for all component props (keys in the map). This is guaranteed by `reagent/adapt-react-class`. The values of the props are a different story altogether.

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
```diff
@@ -0,0 +1,75 @@
...
+     [rn/view {:margin-top 60}
+      [quo/permission-context [example-1] #(js/alert "drawer pressed")]]
+     [rn/view {:margin-top 60}
+      [quo/permission-context [example-2] #(js/alert "drawer pressed")]]
+     [rn/view {:margin-top 60}
+      [quo/permission-context [example-3] #(js/alert "drawer pressed")]]]))
+(defn preview-permission-drawers
+  []
+  [rn/view
+   {:background-color (colors/theme-colors colors/neutral-20 colors/neutral-100)
```
ilmotta: Less important in preview files perhaps, but we're trying to put the style map inside the `:style` key.

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
```diff
@@ -0,0 +1,75 @@
...
+    :flex             1}
+   [rn/flat-list
+    {:flex                      1
+     :nestedScrollEnabled       true
+     :keyboardShouldPersistTaps :always
+     :header                    [cool-preview]
+     :key-fn                    (fn [_ index] (str "actions-drawers-" index))}]])
```
ilmotta: Curious, why do we need the prefix `actions-drawers-`?

ilmotta (approved):

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
J-Son89: that's copy pasta. Will address, it's probably in tonnes of those files. I'll check and do a follow up on all them if so

ilmotta: It is all over, I'm just trying to raise awareness little by little. Thanks @J-Son89

src/status_im2/contexts/quo_preview/drawers/permission_drawers.cljs
J-Son89: i think i was getting duplicate keys before. not sure if needed in this file though

ilmotta: Probably not, because the string is fixed, so it won't help with the uniqueness problem.

Another common approach that works well enough when the content of each element of the list is guaranteed to be different and there's no unique ID, is to use the `hash` function.

```clojure
:key-fn hash
```

OmarBasem:
Hey @J-Son89, the drawer items currently does not look right on iOS

J-Son89:
> Hey @J-Son89, the drawer items currently does not look right on iOS

can you expand what issues you are seeing with it?

PR #14839, opened 2023-01-20
add shadows to foundations

### LINKED ISSUE #14806: Quo2 - Add shadows to foundations
implement the shadows to their own foundation as shown here:

https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Foundations?node-id=624%3A750&t=JNECjV8RIJbZ2ivl-0

This will allow shadows to be used as a named variable across the codebase so they can easily be adjusted if requested by the design team

Provide a preview of these foundations in the quo2 preview screens.

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14806

from the designs: https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Foundations?node-id=624%3A752&t=QRewvOIBGZMsW0ct-0

To see in app, go to quo2 preview foundations

Android preview
Screenshot 2023-01-20 at 11 24 28


jo-mut (approved):
looks good

src/quo2/foundations/elevation.cljs
```diff
@@ -0,0 +1,70 @@
...
+
+(defn- get-inverted [inverted? number] (if inverted? (* -1 number) number))
+
+(defn- get-scales
+  [inverted?]
+  (if (theme/dark?)
+    {:elevation-1 {:shadow-color   (colors/alpha colors/neutral-100 0.5)
+                   :shadow-offset  {:width  0
+                                    :height (get-inverted inverted? 4)}
+                   :elevation      3
```
flexsurfer: could you please elaborate on why it starts from 3 , and also i believe elevation is relative, so if it will be in the container with different elevation there will be different shadow ?

J-Son89: tbh, I am not very familiar with elevation. I will read up more on it. My intention here was to get this setup and then work with a designer to configure the values directly. The values in figma are for css and it's not very clear of the 1-1 mapping between react-native shadows and css shadows. especially since ios and android are handled separately. In any case we can add the correct named shadows into the codebase and then devs can start using these named shadows as in the designs and will get the correct updates once we have approval from the design team.

flexsurfer: great, thanks

flexsurfer (approved):

src/quo2/foundations/elevation.cljs
```diff
@@ -0,0 +1,70 @@
+(ns quo2.foundations.elevation
```
flexsurfer: is it named elevation in figma? not shadow?

J-Son89: yeah, I will ask the designers to adjust this. and then update it. shadows is better!

src/quo2/foundations/shadows.cljs
```diff
@@ -0,0 +1,70 @@
+(ns quo2.foundations.shadows
+  (:require [quo2.foundations.colors :as colors]
+            [quo2.theme :as theme]))
+
+
+(defn- get-inverted [inverted? number] (if inverted? (* -1 number) number))
```
ilmotta: I think this could be easier to read with at least a line break after the square brackets. In general, single-line function definitions should be avoided.

```clojure
(defn- get-inverted [inverted? number]
  (if inverted? (* -1 number) number))
```

ilmotta (approved):

ibrkhalil (approved):

Issue #14794, opened 2023-01-16
Crash when the maximum number of pinned messages is exceeded


qoqobolo:
# Bug Report

### Reproduction

1. Open a community channel
2. Send 4 messages to the chat
3. Pin all the messages

#### Expected behavior
The message `You can pin a max of 3 messages...` is shown, no crash

https://www.figma.com/file/PPWkgOYlZZDxZv5SDGsZVV/Posts-%26-Attachments-for-Mobile?node-id=3512%3A404015&t=ovHK8GTOofB9Dxwi-4

#### Actual behavior

https://user-images.githubusercontent.com/67952253/212669092-656cec37-ef87-4ab2-8ded-bbbb84c9ce98.mp4

### Additional Information

- Status version: nightly 
- Operating System: Android, iOS

### Notes

Related issue: [[quo2 new design tech debt] refactor pin-limit-popover to in-app notification component in quo2](https://github.com/status-im/status-mobile/issues/14527)

PR #14842, opened 2023-01-20
[Fixes: #14794] Use toast instead of old pin modal

### LINKED ISSUE #14527: [quo2 new design tech debt] refactor pin-limit-popover to in-app notification component in quo2
status-im.ui2.screens.chat.pin-limit-popover.view should be refactored to in-app notification component in quo2

https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=1686%3A34385&t=wPuo9JTlBvif5Vfn-3

### LINKED ISSUE #14637: "You can't pin more 3 messages" popover is not possible to interact with and overlapping with other elements
# Bug Report

## Problem 
image

#### Expected behavior
can interact with popover, text is visible

#### Actual behavior
can't close it; text is overlapping with other UI elements

### Reproduction

1) Open Status
2) Start 1-1 chat with any user 
3) Pin 3 messages
4) Try to pin 4 message

### Additional Information

- Status version: nightly 26/12/22

cammellos:
Instead of using the old modal, we show a toast when the limit of 3 messages is reached, it removes the old modal as not used anymore.

It should [Fixes: #14527] as well, @flexsurfer can you confirm, I could not find the exact designs, but as I understand we only show a toast when the limit is reached, for sure this is good enough™ for RC.

closes #14637 

I have also remove the entry on the home screen to show pinned messages on long press, as not in the designs (anymore?).

The icon and text are slightly off on my device, though it would not be a regression caused by this PR as it does not touch that part of the code, but maybe worth opening a separate issue if we are not aware?

### Testing

1) Pin 3 messages in a one-to-one chat (tested only there for now, let me know if you'd like me to test on communities, but should work the same)
2) Pin a 4th message
3) The toast should be shown on top


src/status_im2/contexts/chat/messages/pin/events.cljs
```diff
@@ -1,6 +1,9 @@
 (ns status-im2.contexts.chat.messages.pin.events
   (:require [re-frame.core :as re-frame]
+            [utils.i18n :as i18n]
+            [quo.design-system.colors :as colors]
```
flexsurfer: quo2 colors should be used

flexsurfer (approved):

ilmotta (approved):
>I have also remove the entry on the home screen to show pinned messages on long press, as not in the designs (anymore?).

Would be better to confirm with them before merging.

>The icon and text are slightly off on my device, though it would not be a regression caused by this PR as it does not touch that part of the code, but maybe worth opening a separate issue if we are not aware?

:+1:

flexsurfer:
hey @cammellos @churik is this issues addressed in this PR https://github.com/status-im/status-mobile/issues/14637  ?

cammellos:
> hey @cammellos @churik is this issues addressed in this PR #14637 ?

I think so, it's the same/similar repro steps? but wasn't so sure since I have missed that one

churik:
It works differently now, but totally acceptable - popover just saying that you need to unpin the message. So I'd say that this PR closes #14637 - just need to adjust e2e to it

PR #14850, opened 2023-01-21
[14849] Only allow dragging inside bottom sheet when no handle is shown

### LINKED ISSUE #14781: Can't scroll photo gallery on Android
# Bug Report

## Problem 

You can only scroll through the gallery with two fingers on the screen. 

https://user-images.githubusercontent.com/67952253/212388773-cf92191a-c7d4-496a-92dc-9affe3db8af1.mp4

### Reproduction

1. Open a chat on Android
2. Open a photo gallery
3. Try to scroll it

### Additional Information

- Status version: nightly 13/01
- Operating System: Android

### LINKED ISSUE #14849: Make bottom-sheet drag gesture only on work on handle
# Bug Report

## Problem 
This issue is to prevent the bottom sheet from being draggle on all of it's view (Ideally we want it to be draggable only from it's top handle) this will fix the issue of photo selector not scrolling and it should've been this way from the start.

ibrkhalil:
fixes #14849, #14781

### Summary
The current bottom sheet implementation allows to drag the bottom sheet down to dismiss it in any part of it, If a flat-list or any scrollable component exists nested inside it, it won't be able to scroll in it .. So a solution that @flexsurfer suggested was to only drag it down from it's handle .. This PR does the proposed solution

status: ready


src/status_im2/common/bottom_sheet/view.cljs
```diff
@@ -247,9 +211,54 @@
                {:style (reanimated/apply-animations-to-style
                         {:opacity bg-opacity}
                         styles/backdrop)}]]
+             ;; VVVVVVV
+             ;; This is repeated because gesture-handler doesn't play nicely when it's conditionally
```
ilmotta: It's hard for me to accept this @ibrkhalil. There must be a way to not simply copy and paste the whole thing. We should be able to extract the common logic into a separate function and remove the duplication.

Extracting to a separate function would have the added benefit of helping the bottom sheet component *not grow* any longer. This component is crying to be broken down into smaller chunks, really hard to read IMO.

src/status_im2/common/bottom_sheet/view.cljs
```diff
@@ -247,9 +211,54 @@
...
+                    [rn/view
+                     {:style     (styles/content-style insets)
+                      :on-layout (when-not (and
+                                            (some? @content-height)
+                                            (> @content-height 0))
+                                   on-content-layout)}
+                     children]]
+
+                   [handle-comp window-width]]]]]
+               #_[gesture/gesture-detector {:gesture bottom-sheet-gesture}
```
ilmotta: Why keep the zombie code here? Note that zombie code is worse than dead code, it gently attacks developers every now and then, confuses them, and drags them into unnecessary discussions ;)

Seriously, if there's a need to keep the zombie code, please, always add a comment explaining why. But preferably, just remove them and we can rely on git for going back in history.

ibrkhalil: My bad Icaro, This is a mistake

src/status_im2/common/bottom_sheet/view.cljs
ibrkhalil: Alright, I'll investigate more on ways to refactor it and not repeat myself

src/status_im2/common/bottom_sheet/view.cljs
ilmotta: No worries :mechanical_arm:

src/status_im2/common/bottom_sheet/view.cljs
ibrkhalil: Can you please check the current approach Icaro ?

src/status_im2/common/bottom_sheet/view.cljs
ibrkhalil: Done

src/status_im2/common/bottom_sheet/view.cljs
```diff
@@ -200,28 +213,27 @@
...
+                          [rn/view {:style (styles/selected-background)}
+                           [selected-item]])
+                        [rn/view {:style (styles/background)}
+                         [rn/keyboard-avoiding-view
+                          {:behaviour (if platform/ios? :padding :height)
+                           :style     {:flex 1}}
+                          [rn/view
+                           {:style     (styles/content-style insets)
+                            :on-layout (when-not (and
+                                                  (some? @content-height)
```
cammellos, line 231: This looks like it can be simplified, I think
at least `(when-not (pos? @content-height) ...)`, you are just checking if it's `0` or `nil?` right?

src/status_im2/common/bottom_sheet/view.cljs
ilmotta: Much better!

src/status_im2/common/bottom_sheet/view.cljs
ilmotta: I usually prefer to avoid inverse logic, also `some?` is usually not needed in Clojure, because Clojure has *truthiness* (non-nil is true).

```clojure
(when (and @content-height (<= @content-height 0))
  on-content-layout)
```

Or if we are sure the content height can't be negative:

```clojure
(when (and @content-height (zero? @content-height))
  on-content-layout)
```

ilmotta (approved):

src/status_im2/common/bottom_sheet/view.cljs
```diff
@@ -83,6 +83,17 @@
...
+  [window-width]
+  [rn/view
+   {:style {:width            window-width
+            :position         :absolute
+            :background-color :transparent
+            :top              0
+            :height           20}}
+   [rn/view {:style (styles/handle)}]])
+
+
```
ilmotta: Extra line

alwx (approved):

VolodLytvynenko:
Hi @ibrkhalil Thank you for your work.  The PR is tested and ready to be merged!

PR #14855, opened 2023-01-23
fix nil value in reanimated style crashing at runtime

### LINKED ISSUE #14756: nil value in reanimated style crashing at runtime
# Problem

If there exists a nil value for any of reanimated style parameters, then it makes the app crash.
Even if nil is for direct param or nested param like, `:transform [{:style nil}]` or even when nil is the value of shared-value. (related [comment](https://github.com/status-im/status-mobile/blob/ceaa363f087ff825e5d74a9143332b9c834c9a01/src/status_im2/contexts/shell/animation.cljs#L77))

Thanks @flexsurfer for finding this

**Error:**
erro

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/14756

status: ready

### DIFF
--- a/src/react_native/reanimated.cljs
+++ b/src/react_native/reanimated.cljs
@@ -61,7 +61,7 @@
 
 (defn set-shared-value
   [anim val]
-  (when (and anim val)
+  (when (and anim (some? val))
     (set! (.-value anim) val)))
 
 ;; Worklets
--- a/src/react_native/reanimated.cljs
+++ b/src/react_native/reanimated.cljs
@@ -14,7 +14,6 @@
                              SlideInUp
                              SlideOutUp
                              LinearTransition)]
-            [camel-snake-kebab.core :as csk]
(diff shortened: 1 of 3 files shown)


src/react_native/reanimated.cljs
```diff
@@ -79,17 +78,10 @@
                       (clj->js extrapolation))))
 
 ;;;; Component Animations
-
-;; kebab-case styles are not working for worklets
```
flexsurfer, line 83: just wondering why you decided to move this to js side ?

flexsurfer (approved):

src/react_native/reanimated.cljs
Parveshdhull: The function we had to convert `kebab-case` to `camel-case` was only mapping direct keys, not nested keys of the transform param.
And we already iterating through maps in js side, so it made more sense to change the case of keys at that time.

Issue #14706, opened 2023-01-04
Red icons in Activity Centre


churik:
# Bug Report

## Problem 
The avatars are red in the activity center.
Was introduced on nightly 04/01/23

![IMAGE 2023-01-04 11:32:19](https://user-images.githubusercontent.com/4557972/210536178-f1fa8193-be0b-4d56-b94d-3bc4bfed9ffe.jpg)

### Additional Information

- Status version: nightly 
- Operating System: Android, iOS

[comment]: # (Please, add logs/notes if necessary)

churik:
If not relevant, can be closed :) thanks!

Parveshdhull:
red color fixed in https://github.com/status-im/status-mobile/pull/14856, and original issue superseded by https://github.com/status-im/status-mobile/issues/14857

PR #14626, opened 2022-12-22
Improve switcher last content

Parveshdhull:
### Summary:
Added the last content of switcher cards for the photo, sticker, gif, audio, community link, and community channel.
(photo & sticker only using the mock image for now, due to https://github.com/status-im/status-mobile/issues/14625)

### Testing Request:
Please test for any regressions while sending message of above content types.

status: ready

PR #14856, opened 2023-01-23
Implement unread badge for bottom tabs

Parveshdhull:
status: ready

### DIFF
--- a/src/quo2/components/tags/context_tags.cljs
+++ b/src/quo2/components/tags/context_tags.cljs
@@ -64,7 +64,7 @@
        [rn/image
         {:style  {:width            20
                   :border-radius    10
-                  :background-color :red
+                  :background-color :white
                   :height           20}
          :source photo}]
        [rn/view
--- a/src/status_im2/contexts/shell/bottom_tabs.cljs
+++ b/src/status_im2/contexts/shell/bottom_tabs.cljs
@@ -1,35 +1,39 @@
 (ns status-im2.contexts.shell.bottom-tabs
   (:require [quo2.components.navigation.bottom-nav-tab :as bottom-nav-tab]
             [react-native.core :as rn]
+            [utils.re-frame :as rf]
             [react-native.reanimated :as reanimated]
             [status-im2.contexts.shell.animation :as animation]
             [status-im2.contexts.shell.constants :as shell.constants]
             [status-im2.contexts.shell.style :as styles]))
 
 (defn bottom-tab
-  [icon stack-id shared-values]
+  [icon stack-id shared-values notifications-data]
   [bottom-nav-tab/bottom-nav-tab
-   {:test-ID             stack-id
-    :icon                icon
-    :icon-color-anim     (get
-                          shared-values
-                          (get shell.constants/tabs-icon-color-keywords stack-id))
-    :on-press            #(animation/bottom-tab-on-press stack-id)
-    :accessibility-label (str (name stack-id) "-tab")}])
+   (merge
+    (get notifications-data stack-id)
+    {:test-ID             stack-id
+     :icon                icon
+     :icon-color-anim     (get
+                           shared-values
+                           (get shell.constants/tabs-icon-color-keywords stack-id))
+     :on-press            #(animation/bottom-tab-on-press stack-id)
+     :accessibility-label (str (name stack-id) "-tab")})])
 
 (defn bottom-tabs
   []
   [:f>
    (fn []
-     (let [shared-values  @animation/shared-values-atom
-           original-style (styles/bottom-tabs-container @animation/pass-through?)
-           animated-style (reanimated/apply-animations-to-style
-                           {:height (:bottom-tabs-height shared-values)}
-                           original-style)]
+     (let [notifications-data (rf/sub [:shell/bottom-tabs-notifications-data])
+           shared-values      @animation/shared-values-atom
+           original-style     (styles/bottom-tabs-container @animation/pass-through?)
+           animated-style     (reanimated/apply-animations-to-style
+                               {:height (:bottom-tabs-height shared-values)}
+                               original-style)]
        (animation/load-stack @animation/selected-stack-id)
        [reanimated/view {:style animated-style}
         [rn/view {:style (styles/bottom-tabs)}
-         [bottom-tab :i/communities :communities-stack shared-values]
-         [bottom-tab :i/messages :chats-stack shared-values]
-         [bottom-tab :i/wallet :wallet-stack shared-values]
-         [bottom-tab :i/browser :browser-stack shared-values]]]))])
+         [bottom-tab :i/communities :communities-stack shared-values notifications-data]
+         [bottom-tab :i/messages :chats-stack shared-values notifications-data]
+         [bottom-tab :i/wallet :wallet-stack shared-values notifications-data]
+         [bottom-tab :i/browser :browser-stack shared-values notifications-data]]]))])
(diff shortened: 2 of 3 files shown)


src/quo2/components/tags/context_tags.cljs
```diff
@@ -64,7 +64,7 @@
        [rn/image
         {:style  {:width            20
                   :border-radius    10
-                  :background-color :red
+                  :background-color :white
```
Parveshdhull, line 67: Hi, I changed this to `red` in https://github.com/status-im/status-mobile/pull/14626 because context tags don't have a fallback design. But not sure if the red color approach works for this component, as this is quite small. So creating an [issue](https://github.com/status-im/status-mobile/issues/14857) for that and reverting it back. (fixes https://github.com/status-im/status-mobile/issues/14706)

src/status_im2/contexts/shell/bottom_tabs.cljs
```diff
@@ -1,35 +1,39 @@
...
+  [icon stack-id shared-values notifications-data]
   [bottom-nav-tab/bottom-nav-tab
-   {:test-ID             stack-id
-    :icon                icon
-    :icon-color-anim     (get
-                          shared-values
-                          (get shell.constants/tabs-icon-color-keywords stack-id))
-    :on-press            #(animation/bottom-tab-on-press stack-id)
-    :accessibility-label (str (name stack-id) "-tab")}])
+   (merge
```
cammellos: No need to merge when the second key is not dynamic, in this case you can use `assoc` as it's multivariadic and it's clear that the keys are known and limited:

```
(assoc (get-notifications-data stack-id)
  :test-ID stack-id
  :icon icon
  :icon-color-anim (get...)
  ...)
```
It's also considerably more performant (numbers are not relevant since I run it on my desktop, but percentage difference is meaningful/ish, run 10/20 times and numbers are consistent):

```
(time (doseq [i (range 0 1000000)] (assoc {} :a "a" :b "b")))
"Elapsed time: 966.267824 msecs
```

```
cljs.user=> (time (doseq [i (range 0 1000000)] (merge {} {:a "a" :b "b"})))
"Elapsed time: 1361.368269 msecs"
```

Likely since you are not allocating a map, and not iterating over the dynamic keys.

src/status_im2/subs/shell.cljs
```diff
@@ -147,3 +147,35 @@
    (let [community-id (:community-id channel)
          community    (get communities (:community-id channel))]
      (community-channel-card community community-id channel channel-id))))
+
+(re-frame/reg-sub
```
cammellos, line 151: This would benefit from some tests, since the logic is fairly complex

src/status_im2/subs/shell.cljs
```diff
@@ -147,3 +147,35 @@
...
+ (fn [chats]
+   (let [chats-seq                      (map second chats)
+         unviewed-messages-count-seq    (map #(hash-map (:chat-type %)
+                                                        (or (:unviewed-messages-count %) 0))
+                                             chats-seq)
+         unviewed-mentions-count-seq    (map #(hash-map (:chat-type %)
+                                                        (or (:unviewed-mentions-count %) 0))
+                                             chats-seq)
+         unviewed-messages-count-total  (apply merge-with + unviewed-messages-count-seq)
+         unviewed-mentions-count-total  (apply merge-with + unviewed-mentions-count-seq)
```
flexsurfer: TIL

flexsurfer: omg it's so satisfying when you find such solutions :)

flexsurfer (approved):

src/status_im2/subs/shell.cljs
```diff
@@ -147,3 +147,35 @@
    (let [community-id (:community-id channel)
          community    (get communities (:community-id channel))]
      (community-channel-card community community-id channel channel-id))))
+
+(re-frame/reg-sub
+ :shell/bottom-tabs-notifications-data
+ :<- [:chats/chats]
+ (fn [chats]
+   (let [chats-seq                      (map second chats)
+         unviewed-messages-count-seq    (map #(hash-map (:chat-type %)
```
cammellos: I think this can be done in one go with a reduce, it should be faster and a bit simpler, if I understand correctly, you want to get a counter for mentions/messages, something like that should work (pseudo-code/ish and you can extract and test separately, also you can make other decisions in terms of how you want the accumulator to look like, if you want already to group private/one to one in the reduce etc):

```
(reduce
   (fn [acc {:keys [unviewed-messages-count unviewed-mentions-count chat-type]}]
       (if (chat-type is not one we care about)
          acc
          (-> acc
             (update-in [chat-type :mentions-count] + unviewed-mentions-count) // adding nil is ok, it's counted as 0, you can coherce if you want though)
             (update-in [chat-type :unviewed-mentions-count] + unviewed-mentions-count))
            
)
    {community-chat-type {:mentions-count 0 :messages-count}
      one-to-one-chat-type {:mentions-count 0 :messages-count}}
     chats)
  ```

You can also filter before the merge if you want, so you only care about certain chat types, instead of having it to do in the reduce, in that case a transducer might be more efficient https://clojure.org/reference/transducers since otherwise you'd be building an intermediate collection, but I don't think it's worth the complexity, nor I know whether it is actually going to be more performant.

File: src/status_im2/contexts/shell/bottom_tabs.cljs (hunk 2 above)
Comment (Parveshdhull) [reply src/status_im2/contexts/shell/bottom_tabs.cljs]: done, thank you

File: src/status_im2/subs/shell.cljs (hunk 4 above)
Comment (flexsurfer) [reply src/status_im2/subs/shell.cljs]: but yeah sure performance-wise one `reduce` always win :)

**Comment by pavloburykh** [comment]
@Parveshdhull hi! thanx for the PR. Could you please rebase it to the latest develop? Thank you!

File: src/status_im2/subs/shell.cljs (hunk 5 above)
Comment (Parveshdhull) [reply src/status_im2/subs/shell.cljs]: I actually was implementing this with reduce initially. But don't know why that looked more complex (maybe because of too many ifs).

But yes with reduce time taken is decreased from 0.91 ms to 0.46 ms :tada:. Thank you very much.

Comment (Parveshdhull) [reply src/status_im2/subs/shell.cljs]: looking at second suggestion

Comment (Parveshdhull) [reply src/status_im2/subs/shell.cljs]: tried filtering input of reduce function using

```
          (filter #(#{constants/one-to-one-chat-type
                      constants/community-chat-type
                      constants/private-group-chat-type}
                    (:chat-type %)) (vals chats)))
```
But that increased time to `~0.65`

src/status_im2/subs/shell.cljs
Parveshdhull: this looked cool though :)

src/status_im2/subs/shell.cljs
Parveshdhull: sounds good, will do that later in a separate PR.

jo-mut (approved):

cammellos (approved):

pavloburykh:
@Parveshdhull thanx for the PR. 

Just to make sure that I haven't missed anything: I have been able to test unread badge for 2 tabs - Communities and Chats tabs. 
Unread badge appears when new messages are received in community channels, 1-1 chats and groups
Counters of unread messages appear when mentions and replies are received in channels, 1-1 chats and groups

Are there any other cases when unread badge/counter should appear on for bottom tabs?

Parveshdhull:
Hi @pavloburykh, Thank you for testing PR.

> Are there any other cases when unread badge/counter should appear on for bottom tabs?

No these are only the cases.

pavloburykh:
> No these are only the cases.

@Parveshdhull cool. PR is tested and ready for merge. Thanx for your work.

PR #14873, opened 2023-01-24
[#14718] Fix profile photo in several places

### LINKED ISSUE #14718: Profile photo is not updated on receiver side
# Bug Report

## Problem 
The profile photo is shown as an empty picture or placeholder to mutual contacts.

#### Expected behavior
The profile photo is updated

#### Actual behavior
photo_2023-01-05 16 57 24 jpeg 2023-01-05 16-58-46

### Reproduction

1) User A and User B are added mutual contact 
2) User A: start chat with User B
3) User A: set profile photo
4) User B: check that profile photo is updated

### Additional Information

- Status version: nightly 05/01/23

rasom:
fix #14718

status: ready

### DIFF
--- a/src/react_native/fast_image.cljs
+++ b/src/react_native/fast_image.cljs
@@ -14,11 +14,14 @@
   [_]
   (let [loaded? (reagent/atom false)
         error?  (reagent/atom false)]
-    (fn [props]
+    (fn [{:keys [source] :as props}]
       [fast-image-class
        (merge
         props
-        {:on-error (fn [e]
+        {:source   (if (string? source)
+                     {:uri source}
+                     source)
+         :on-error (fn [e]
                      (when-let [on-error (:on-error props)]
                        (on-error e))
                      (reset! error? true))


ilmotta (approved):
@rasom: I know you wanted a quick fix and it LGTM, but I think it would be preferable if callers always passed the correct map with `:uri` key. I found only 6 usages of `fast-image` in the code, and they're rather simple to fix. In other words, the bug is on the callers, not the implementation of `fast-image`.

In general I just prefer if we can avoid having an argument with multiple types. If a map is expected, then only a map is supported, and so on. When multiple types are supported, then every place that deals with the argument has to remember to check which type its dealing with.

rasom:
@ilmotta 
not only that, it is quick fix anyway (even if I fix it where this nice `fast-image` is used). The problem is that the issue is going be a recurring one unless we a) just handle it for the client as in PR b) or show some kind of warning or something (less preferable because is going to be ignored anyway).

So I see no harm in allowing clients to use string instead of map in this particular case, because in 99.99% of use cases it is going to be a string and we would just force client to wrap it via `{:uri string}` for no real reason except our desire for it to be a map everywhere `¯\_(ツ)_/¯`

ilmotta:
> @ilmotta not only that, it is quick fix anyway (even if I fix it where this nice `fast-image` is used). The problem is that the issue is going be a recurring one unless we a) just handle it for the client as in PR b) or show some kind of warning or something (less preferable because is going to be ignored anyway).
> 
> So I see no harm in allowing clients to use string instead of map in this particular case, because in 99.99% of use cases it is going to be a string and we would just force client to wrap it via `{:uri string}` for no real reason except our desire for it to be a map everywhere `¯\_(ツ)_/¯`

Okay, I get it @rasom, then the solution sounds more reasonable to me.

briansztamfater (approved):

qoqobolo:
Thanks for the fix @rasom, LGTM!

PR #14874, opened 2023-01-24
chore: fix sorting of pinned messages

### LINKED ISSUE #14743: The pinned message is not updated in the chat screen if one more message is pinned
**Steps to reproduce:**
1. Go to 1-1 chat
2.  Send message "one"
3. Pin message "one" -> pinned message now shows as "one"
4. Send message "two"
5. Pin message "two"

**Actual result:**
The pinned message still shows as "one"

**Expected result:**
The pinned message is shown as "two"

**Notes:**
The pinned message updates after navigating away and back to the chat screen. But should be updated online without chat reopening

J-Son89:
fixes https://github.com/status-im/status-mobile/issues/14743

To test: 

go to a 1-1 chat 
send a message "Test 1"
pin message "Test 1"
send a message "Test 2"
pin message "Test 2"

expected result - "Test 2" should be shown as the pinned message on screen banner. 

Can also test other variants of this to ensure the latest pinned message is showing in the banner

### DIFF
--- a/src/status_im2/contexts/chat/messages/pin/banner/view.cljs
+++ b/src/status_im2/contexts/chat/messages/pin/banner/view.cljs
@@ -4,14 +4,12 @@
 
 (defn banner
   [chat-id]
-  (let [pinned-messages (rf/sub [:chats/pinned chat-id])
-        latest-pin-id   (-> pinned-messages
-                            vals
-                            first
-                            (get :message-id))
-        latest-pin-text (get-in (rf/sub [:chats/chat-messages chat-id])
-                                [latest-pin-id :content :text])
-        pins-count      (count (seq pinned-messages))]
+  (let [pinned-messages (rf/sub [:chats/pinned-sorted-list chat-id])
+        latest-pin-text (->> pinned-messages
+                             last
+                             :content
+                             :text)
+        pins-count      (count pinned-messages)]
     [quo/banner
      {:latest-pin-text latest-pin-text
       :pins-count      pins-count
--- a/src/status_im2/subs/chat/messages.cljs
+++ b/src/status_im2/subs/chat/messages.cljs
@@ -149,14 +149,26 @@
                 {}
                 pin-messages))))
 
+;; local messages will not have a :pinned-at key until user navigates away and to
+;; chat screen. For this reason we want to retain order of local messages with :pinned-at nil
+;; as these will be a stack on top of the messages, however we do want to sort previous messages
+;; from backend that have a :pinned-at value.
+(defn sort-pinned
+  [a b]
+  (cond
+    (and a b) (- a b)
+    (or a b)  (if b false true)
+    :else     a))
+
 (re-frame/reg-sub
  :chats/pinned-sorted-list
  (fn [[_ chat-id] _]
    (re-frame/subscribe [:chats/pinned chat-id]))
  (fn [pin-messages _]
-   (->> pin-messages
-        vals
-        (sort-by :pinned-at <))))
+   (let [pin-messages-vals (vals pin-messages)]
+
+     (sort-by :pinned-at sort-pinned pin-messages-vals))))
+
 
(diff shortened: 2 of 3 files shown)


src/status_im2/contexts/chat/messages/pin/banner/view.cljs
```diff
@@ -4,14 +4,12 @@
 
 (defn banner
   [chat-id]
-  (let [pinned-messages (rf/sub [:chats/pinned chat-id])
-        latest-pin-id   (-> pinned-messages
-                            vals
-                            first
-                            (get :message-id))
-        latest-pin-text (get-in (rf/sub [:chats/chat-messages chat-id])
```
J-Son89, line 12: this second sub seems unnecessary as there is access to latest-pin-text from the first sub

src/status_im2/subs/chat/messages.cljs
```diff
@@ -156,7 +156,11 @@
  (fn [pin-messages _]
    (->> pin-messages
         vals
-        (sort-by :pinned-at <))))
+        (sort-by :pinned-at
```
J-Son89: newly pinned messages don't have `pinned-at` property until the user navigates away and back to the page so we just sort by the pinned messages with this pinned-at  value

ilmotta: I tried to understand this sorting change @J-Son89, but I couldn't :man_shrugging: The code only sorts subsequent elements in the list containing a non-nil pinned-at value. Is this really what you expected? It's weird to see sorting that doesn't sort anything *sometimes*.

```clojure
(defn pinned-at-comparator
  [a b]
  (if (and a b)
    <
    (if a b a)))

;; Nothing is sorted because there are elements with nil interspersed.
(let [pin-messages [{:id 2 :pinned-at nil}
                    {:id 5 :pinned-at 9}
                    {:id 4 :pinned-at 11}
                    {:id 3 :pinned-at nil}
                    {:id 1 :pinned-at 10}]]
  (assert (= pin-messages
             (sort-by :pinned-at
                      pinned-at-comparator
                      pin-messages))))

;; The list is sorted in ascending order because there are no elements with nil interspersed.
(assert (= [{:id 5 :pinned-at 9}
            {:id 1 :pinned-at 10}
            {:id 4 :pinned-at 11}
            {:id 2 :pinned-at nil}
            {:id 3 :pinned-at nil}]
           (sort-by :pinned-at
                    pinned-at-comparator
                    [{:id 4 :pinned-at 11}
                     {:id 1 :pinned-at 10}
                     {:id 5 :pinned-at 9}
                     {:id 2 :pinned-at nil}
                     {:id 3 :pinned-at nil}])))

```

J-Son89: yeah, I agree I should adjust. but in this particular case there will never be interspersed `nil` values because the `nil` values only get stacked on when a user is in the chat menu and pinning/un-pinning messages. After they navigate back and to the pinned-messages have a `pinned-at` value

J-Son89: so yeah, ideally there is cleaner code but this actually does exactly what I want :D

ilmotta: Cool, thanks for the info @J-Son89.

ilmotta (approved):

jo-mut (approved):

qoqobolo:
Thank you @J-Son89, looks good to me.
PR can be merged.

src/status_im2/contexts/chat/messages/pin/banner/view.cljs
ibrkhalil: The second sub was to make pinned messages react to being edited @J-Son89

J-Son89: Ah 😑

PR #14892, opened 2023-01-25
Remove test for join-existing-users-string

vkjr:
### Summary

using `(i18n/label)` within the test gives unstable results, so the test is removed for now.


flexsurfer (approved):

ilmotta:
@vkjr, I went in a deep rabbit hole now to understand what happened. I'm not gonna solve this problem, but I'll share this with you guys. It's worth understanding this if you want to bring the tests back.

Here's the thing, when `(js/require "../translations/en.json")` runs during the setup of translations, the require actually uses cached and stale data.

To understand if changing how the file is read fixes the problem, I implemented a macro in CLJ to read any file using `slurp` **during compile time**.

```clojure
(defmacro read-file
  [path]
  (slurp path))
```

In the function `i18n-resources/require-translation`:

```diff
       :de      (js/require "../translations/de.json")
       :el      (js/require "../translations/el.json")
-      :en      (js/require "../translations/en.json")
+      :en      (js/JSON.parse (read-file "../translations/en.json"))
       :es      (js/require "../translations/es.json")
       :es_419  (js/require "../translations/es_419.json")
```
 
Then I used this macro in place of the `js/require` and ran `make test`. All tests pass, always :100: 

So, I do not recommend that we fix the issue like I did. That was just me trying to prove what's the root cause of the problem. If you want to dig into this problem, I think you'll need to understand why `i18n-resources/require-translation` is reading stale data. If that's fixed, then all tests will pass as I showed above.

vkjr:
@ilmotta, wow, thats very interesting, thank you!

PR #14639, opened 2022-12-26
fix home stack height

Parveshdhull:
fixes: https://github.com/status-im/status-mobile/issues/14633

### Testing:
I don't have access to `Xiaomi Mi 9T Pro`, please can you confirm if this fixes the issue? cc @churik 
Also, updated scroll to bottom accessibility-id as mentioned in https://github.com/status-im/status-mobile/issues/14617#issuecomment-1362778722 cc @VladimrLitvinenko 

status: ready

Issue #14633, opened 2022-12-23
"Jump to" panel overlaps tabs (Xiaomi Mi 9T Pro)


churik:
# Bug Report

## Problem 
Was not able to reproduce it on other devices (Tried on Xiaomi, Huawei, and Samsung)
Any help with reproduction is highly appreciated.

#### Expected behavior
no overlapping

#### Actual behavior

https://user-images.githubusercontent.com/4557972/209359542-59a7ba6a-4455-459f-a201-6ac1e1c7232f.mp4

### Reproduction

1) Open Status
2) Tap on "Jump to"

### Additional Information

- Status Version: nightly 23/12/22
- Operating System: Android 10 (MIUI 12.0.5) 
- Device: Mi 9T Pro

Parveshdhull:
**Issue Cause:** https://github.com/facebook/react-native/issues/23693
**Current Workaround:** https://github.com/status-im/status-mobile/blob/a39e0f6fbdcf73779f4cd618b02f24691c1ea737/src/status_im2/contexts/shell/constants.cljs#L20

Will fix it, thank you for finding it.

Parveshdhull:
Another react-native Issue, reporting same bug: https://github.com/facebook/react-native/issues/27231

Parveshdhull:
So the issue is with devices with notch and dimensions/window

- With notch devices providing height only of the screen without a status bar included.
- While without notch devices providing height with status bar included.

So first we have to detect if the device has a notch, and if it has we need to manually add status-bar height.

But the problem is how to detect the notch.
1. [First way](https://github.com/facebook/react-native/issues/23693#issuecomment-662860819) - According to official android [guidelines](https://m2.material.io/design/platform-guidance/android-bars.html#status-bar), a simple status bar (without a notch) should have a height of 24, so we can detect the status bar height and compare it with this. But for some reason, several android devices don't follow this. For example `Real me 1`
2. [Another way](https://stackoverflow.com/a/56975058) is using the `react-native-device-info` library. But I am not sure, if even this library will be able to provide correct data because from some research of native java code on how to detect notch, they all are relying on 24 guidelines.
But I am currently not able to test this due to https://github.com/status-im/status-mobile/issues/14651

And even if this library will be able to detect notch correctly, we probably shouldn't use it. Because we had a problem with the library in the past while releasing fdroid releases (due to google dependency)
https://github.com/status-im/status-mobile/issues/11809 https://github.com/status-im/status-mobile/pull/13360

[...]

PR #14895, opened 2023-01-25
fix home-screen half-width

### LINKED ISSUE #14894: Home and Profile screens is sometimes displayed at half height
# Bug Report

## Problem 
The only known thing atm is that both times the issue was reproduced, there were at least three chats on the `Chat` screen: one group and two 1-1 chats, and one community on the home screen (this may have been a coincidence though). Also, the bug was reproduced after login / return from the background.

The video does not show that the `Profile` screen also began to be displayed at half height.

If you create a new multi-account, the issue persists.

Screenshot 2023-01-25 at 16 56 10

https://user-images.githubusercontent.com/67952253/214611542-7c4a135a-f534-4ad1-871e-d1abf3b74523.mp4

### Reproduction

images

### Additional Information

- Status version: nightly 25/01
- Operating System: Android

Parveshdhull:
fixes https://github.com/status-im/status-mobile/issues/14894

### Summary

For fixing https://github.com/status-im/status-mobile/issues/14633, in https://github.com/status-im/status-mobile/pull/14639 we used the height provided by the rendered layout as screen height.
I guess for some reason in a very rare case height calculated by on-layout was half of the total height.
(probably due to screen rotation) (even though we only have portrait orientation🤔)

1. So PR makes sure that height is more than width before storing
2. Also removed the `(when-not @animation/screen-height` condition, because it is small overhead, but better than the user getting stuck at one stored value(even if value is wrong)
3. Make sure height returned by on-layout is more than height returned by dimensions

### Testing
Please test https://github.com/status-im/status-mobile/issues/14633, screen height for Xiaomi devices

Thank you @qoqobolo for finding this

status: ready

### DIFF
--- a/src/status_im2/contexts/shell/view.cljs
+++ b/src/status_im2/contexts/shell/view.cljs
@@ -113,6 +113,26 @@
           :style {:margin-top (:top insets)
                   :z-index    2}}]])]))
 
+(defn on-layout
+  [evt]
+  (let [dimensions (rf/sub [:dimensions/window])
+        height     (or (oget evt "nativeEvent" "layout" "height") 0)
+        width      (or (oget evt "nativeEvent" "layout" "width") 0)]
+    ;; Layout height calculation
+    ;; 1. Make sure height is more than width, and on-layout is not fired while the
+    ;; screen is horizontal
+    ;; 2. Initialize values with 0 in case of nil
+    ;; 3. In the case of notch devices, the dimensions height will be smaller than
+    ;; on-layout,
+    ;; (without status bar height included)
+    ;; https://github.com/status-im/status-mobile/issues/14633
+    ;; 4. In the case of devices without a notch, both heights should be the same,
+    ;; but actual values differ in some pixels, so arbitrary 5 pixels is allowed
+    (when (and (> height width)
+               (>= (+ height 5) (or (:height dimensions) 0)))
+      (reset! animation/screen-height height)
+      (async-storage/set-item! :screen-height height))))
+
 (defn shell-stack
   []
   [:f>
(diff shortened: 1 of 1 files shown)


src/status_im2/contexts/shell/view.cljs
```diff
@@ -117,13 +117,26 @@
...
-     (let [shared-values (animation/calculate-shared-values)]
+     (let [shared-values (animation/calculate-shared-values)
+           dimensions    (rf/sub [:dimensions/window])]
        (animation/change-root-status-bar-style)
        [rn/view
         {:style     {:flex 1}
-         :on-layout (when-not @animation/screen-height
-                      (fn [evt]
-                        (let [height (oget evt "nativeEvent" "layout" "height")]
+         :on-layout (fn [evt]
```
J-Son89: consider moving to a helper function 👍

J-Son89 (approved):

src/status_im2/contexts/shell/view.cljs
```diff
@@ -117,13 +117,26 @@
...
+                        ;; screen is horizontal
+                        ;; 2. Initialize values with 0 in case of nil
+                        ;; 3. In the case of notch devices, the dimensions height will be smaller than
+                        ;; on-layout,
+                        ;; (without status bar height included)
+                        ;; https://github.com/status-im/status-mobile/issues/14633
+                        ;; 4. In the case of devices without a notch, both heights should be the same,
+                        ;; but actual values differ in some pixels, so arbitrary 5 pixels is allowed
+                        (when (and (> height width)
+                                   (>= (+ height 5) (or (:height dimensions) 0)))
```
flexsurfer: +5 ?

Parveshdhull:
```
;; 4. In the case of devices without a notch, both heights should be the same,
;; but actual values differ in some pixels, so arbitrary 5 pixels is allowed
```
better safe than sorry :)

src/status_im2/contexts/shell/view.cljs
Parveshdhull: done, thank you

qoqobolo:
Hi @Parveshdhull !
Something's off with iOS builds, could you check please?

churik:
@Parveshdhull tested on Xiaomi mi 9T pro, works the same way as in nightly, so height re-calculates after relogin, so LGTM.
Thank you!

qoqobolo:
Works fine on my devices too.
Let's merge, I'll let you know if the issue reappears at some point. 
@Parveshdhull thank you for finding a possible cause almost blindly and fixing it!

PR #14925, opened 2023-01-30
Animated Header Flatlist

### LINKED ISSUE #14912: Reimplement Animated Scroll View component
Currently, the animated scroll view component is using a plain RN scroll view.

It needs to be reimplemented using a Reanimated FlatList component, for two reasons:

1. The animations will not work smoothly without an animated scrolling component.
2. It will not be possible to nest a FlatList inside a ScrollView.

OmarBasem:
Fixes: https://github.com/status-im/status-mobile/issues/14912

Previous related work: https://github.com/status-im/status-mobile/pull/14695

This PR implements a reusable, smoothly animated header flatlist using a Reanimated Flatlist. It uses a Reanimated Flatlist for two reasons:

1. The animations will not work smoothly without an animated scrolling component.
2. It will not be possible to nest a FlatList inside a ScrollView.


src/status_im2/common/animated_header_list/style.cljs
```diff
@@ -0,0 +1,71 @@
...
+    :align-items      :center
+    :background-color (colors/theme-colors colors/white-opa-40 colors/neutral-80-opa-40)
+    :position         :absolute
+    :top              56
+    :z-index          3}
+   position))
+
+(defn blur-view
+  [animation]
+  (reanimated/apply-animations-to-style
```
ilmotta: I think we should not have hooks running in the functions declared in the `style` namespace, i.e. just define the map, but let the exact hook call be decided by the component consuming the `style.cljs` function.

Couple reasons:

1. Style functions should have no side-effects. `apply-animations-to-style` uses the hook `useAnimatedStyle` and it has side-effects. On a side note, the name `apply-animations-to-style` is actually misleading... In React land, there's also a convention to name hooks with the prefix `use`.
2. Style functions shouldn't use hooks. This is not REPL friendly. For instance, I often run functions declared in `style.cljs` files individually to see the returned map.

In the end, if we allow style functions to use hooks then where do we delineate the boundary between what they should do? I would prefer if could keep them pure functions just returning style maps.

src/react_native/reanimated.cljs
```diff
@@ -27,6 +28,10 @@
 
 (def view (reagent/adapt-react-class (.-View reanimated)))
 (def image (reagent/adapt-react-class (.-Image reanimated)))
+(def reanimated-flat-list (reagent/adapt-react-class (.-FlatList reanimated)))
+(defn flat-list
+  [props]
+  [reanimated-flat-list (rn-flat-list/base-list-props props)])
```
ilmotta, line 35: Wouldn't it be better to only use the reanimated flat list if necessary instead of making it the default? Reanimated is yet another layer on top of flat list, which is already problematic on its own.

src/status_im2/navigation/screens.cljs
```diff
@@ -51,6 +52,11 @@
      :options   {:topBar {:visible false}}
      :component communities.discover/discover}
 
+    ;; THIS IS A TEMPORARY SCREEN. WILL REMOVE AFTER QA TESTING.
```
ilmotta: It would be preferable if you could've avoided the temporary screen, and instead, used the new animated component in one or more real screens. This would give everyone more confidence that the new component behaves and performs well. Could you elaborate why you decided against this route?

src/status_im2/common/animated_header_list/view.cljs
```diff
@@ -0,0 +1,101 @@
...
+
+(defn animated-header-list
+  [{:keys [title-comp main-comp] :as parameters}]
+  [safe-area/consumer
+   (fn [insets]
+     (let [window-height     (:height (rn/get-window))
+           status-bar-height (rn/status-bar-height)
+           bottom-inset      (:bottom insets)
+           ;; view height calculation is different because window height is different on iOS and Android:
+           ;; https://i.stack.imgur.com/LSyW5.png
```
ilmotta: [Nitpick] Imgur is not a reliable source of information. Something like a git repository, Wikipedia or SO would be preferable.

src/status_im2/common/animated_header_list/view.cljs
```diff
@@ -0,0 +1,101 @@
...
+(def header-height 192)
+
+(def threshold (/ header-height 2))
+
+(defn scroll-handler
+  [event initial-y scroll-y opacity-value]
+  (let [current-y (- (oops/oget event "nativeEvent.contentOffset.y") initial-y)]
+    (reanimated/set-shared-value scroll-y current-y)
+    (reanimated/set-shared-value opacity-value
+                                 (reanimated/with-timing (if (> current-y threshold) 1 0)))))
```
ilmotta: @OmarBasem, I'm wondering if we can avoid the repeated calls to `set-shared-value` to set the opacity.

For instance, there will be many different `ys` that will be greater than `threshold`, but all of them will set the value to `1`. Instead, I think we can optimize to only set the shared value if it's different than the previous one. I would say it's worth a shot.

```clojure
(defn scroll-handler
  [event initial-y scroll-y opacity-value]
  (let [current-y   (- (oops/oget event "nativeEvent.contentOffset.y") initial-y)
        new-opacity (if (> current-y threshold) 1 0)]
    (reanimated/set-shared-value scroll-y current-y)
    (when (not= new-opacity opacity-value)
      (reanimated/set-shared-value opacity-value
                                   (reanimated/with-timing new-opacity)))))
```

src/status_im2/common/animated_header_list/style.cljs
OmarBasem: I agree with this. I started writing animation styles this way as suggested by Brian. What are your thoughts @briansztamfater?

src/react_native/reanimated.cljs
OmarBasem: What do you mean default? This is reanimated file. `reanimted-flat-list` defines the Reanimated Flatlist component itself. And the `flat-list` definition below it is just to be able to use kebab-case props, like the normal flat-list we have defined in RN.flat-list file

src/status_im2/navigation/screens.cljs
OmarBasem: I decided not to in order to keep this component and its API as isolated/pure as possible. The initial implementation of the component was kind of mixed with one of the community's screen. The result was that the API was not very reusable. The focus of this PR is on the smoothness of the animations regardless of what is rendered inside of it. 

Ofc following PRs that uses this component will be qa tested. (and also did not want to clutter the PR with changes not directly related to the component)

src/status_im2/common/animated_header_list/view.cljs
OmarBasem: I thought about this, but I believe reanimated is going to ignore setting the same value. But I will check and confirm.

src/react_native/reanimated.cljs
ilmotta: My bad @OmarBasem, got some things messed up. You can resolve this thread.

src/status_im2/navigation/screens.cljs
ilmotta: >Ofc following PRs that uses this component will be qa tested. (and also did not want to clutter the PR with changes not directly related to the component)

It's fine, but you cluttered the PR with fake/temp stuff anyway, right? ;) The effort to create the fake screen could've been used to integrate with a real & simple screen. Maybe a `preview` screen would've been even better for QAs and reviewers to play around with this component, and to help out with future changes because you wouldn't need to throw away the code.

But anyways @OmarBasem, thanks for the explanation.

src/status_im2/common/animated_header_list/view.cljs
ilmotta: Cool, given this function is called over 120 times per second, such optimizations can be quite helpful.

src/status_im2/common/animated_header_list/style.cljs
ilmotta: cc'ing @flexsurfer as well.

src/status_im2/navigation/screens.cljs
OmarBasem: Not the same effort really :) I tried with the messages screen, but considerable amount of changes would need to be done to the styling (lots of z-indexes which I do not know, and would need some debugging)

J-Son89: Yeah, I wonder should we not have this as a component in quo2 library? 🤔 and then it has a preview screen too. I agree Omar it's easier to keep the API abstracted when it's not being used for it's implementation on creation 👍

src/status_im2/common/animated_header_list/style.cljs
OmarBasem: @ilmotta sometimes when a component has multiple animations defined, it can start cluttering the view component. I can see your point, but maybe defining them in style.cljs would be better, especially since this hook is not affecting the view's data or causing re-renders, its only mutating the style, so why keep it in the view

src/status_im2/navigation/screens.cljs
OmarBasem: Yeah maybe I can move it to quo2

src/status_im2/common/animated_header_list/view.cljs
OmarBasem: Got rid of it altogether

src/status_im2/common/animated_header_list/style.cljs
ilmotta: It's a matter of responsibilities and consistency. The guidelines suggest the style namespaces exist to declutter the view from defining style maps. If we start to move hooks to the `style` namespace we enter a nebulous territory where stateful hooks can be called anywhere. I don't know if the analogy will help, but in React land, CSS modules have a very narrow purpose, and that's why they serve their purpose so well.

>@ilmotta sometimes when a component has multiple animations defined, it can start cluttering the view component

So then the view can create a separate function to do all the animation calls, for example.

-----

At the very least, I think we should either change the guidelines or respect it, otherwise they get outdated and become pointless.

ilmotta (approved):
Great work @OmarBasem!

Edit: I experimented with refactoring based on my own suggestion in comment https://github.com/status-im/status-mobile/pull/14925/files#r1091779515, but I didn't like what I saw :man_shrugging: So it's all good from my end.

~~My only open point is about respecting the guidelines and using the `style` namespaces only for defining style maps (ref: https://github.com/status-im/status-mobile/pull/14925/files#r1091779515).~~

src/status_im2/common/animated_header_list/style.cljs
ilmotta: Please, take my previous comments with a grain of salt ;) You can close this thread if you want @OmarBasem 

>I experimented with refactoring based on my own suggestion in comment https://github.com/status-im/status-mobile/pull/14925/files#r1091779515, but I didn't like what I saw :man_shrugging: So it's all good from my end.
>
>-- https://github.com/status-im/status-mobile/pull/14925#pullrequestreview-1277033921

ilmotta: Just opened a PR https://github.com/status-im/status-mobile/pull/14941 after reviewing all `style.cljs` files in the codebase, and motivated by this thread.

Thanks @OmarBasem for the insightful discussion.

OmarBasem: My pleasure @ilmotta, thank you for the discussion!

PR #14941, opened 2023-01-31
Update style namespaces to follow guidelines

ilmotta:
### Summary

A quick PR to update certain `style.cljs` files to more strictly follow our guidelines. Except for animations and theme colors, *style* functions should be pure.

It turns out `style` namespaces are well behaved and are almost perfectly following guidelines :rocket: 

The motivation for this PR came from this thread https://github.com/status-im/status-mobile/pull/14925#discussion_r1090485454

status: ready

### DIFF
--- a/doc/new-guidelines.md
+++ b/doc/new-guidelines.md
@@ -38,31 +38,28 @@ Pay special attention to:
 
 ### Component styles
 
-Prefer to define styles in separate files named `style.cljs`, usually colocated
-with the source file using it. For a real example, see
-[src/status_im/ui2/screens/chat/messages/style.cljs](../src/status_im/ui2/screens/chat/messages/style.cljs).
+Prefer to define styles in a separate file named `style.cljs`, colocated with
+the source file. For a real example, see
+[src/quo2/components/record_audio/record_audio/style.cljs](../src/quo2/components/record_audio/record_audio/style.cljs).
 
 ```clojure
 ;; bad
-(defn animated-checkbox-view [{:keys [size]}]
-  [animated/view
+(defn checkbox-view
+  [{:keys [size]}]
+  [rn/view
    {:style {:width           size
             :height          size
             :border-radius   4
             :justify-content :center
             :align-items     :center}}
-   [animated/view
-    (do-something)]])
+   [rn/view (do-something)]])
 
 ;; good
-(defn animated-checkbox-view [{:keys [size]}]
-  [animated/view
-   {:style (style/animated-checkbox-style size)}
-   [animated/view
-    (do-something)]])
+(defn checkbox-view
+  [{:keys [size]}]
+  [rn/view {:style (style/checkbox size)}
+   [rn/view (do-something)]])
 ```
-
-
 
 ### Don't use percents to define width/height
 
--- a/src/status_im2/contexts/add_new_contact/style.cljs
+++ b/src/status_im2/contexts/add_new_contact/style.cljs
@@ -1,6 +1,5 @@
 (ns status-im2.contexts.add-new-contact.style
   (:require [quo2.foundations.colors :as colors]
-            [status-im.react-native.resources :as resources]
             [react-native.platform :as platform]
             [quo2.foundations.typography :as typography]))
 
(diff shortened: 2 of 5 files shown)


Parveshdhull (approved):

J-Son89 (approved):

OmarBasem (approved):

erikseppanen (approved):
This is good, but I'm not 100% clear on this reasoning.

Some functions in style.cljs are already impure, for example, theme colors are you stated. Is it a case of that those impure functions are so prevalent that it's ok for them to be there?

I conceptualized style.cljs as a namespace where all the style-related responsibilities can be maintained, leaving just layout in the views.cljs, so that we don't have style-related information cluttering the application logic of views.cljs (once look-and-feel is done, it can stay segregated from the domain logic of views.cljs). That was my reasoning for giving style.cljs the responsibility of getting the image that the view was asking for (I left the keyword image name in the view because it seems useful for identifying what it is in the layout).

Because once that style-related code is stable and complete, it is no longer of value in reviewing/working on domain logic in view.cljs - it just adds to visual clutter to parse through, IMO.  In particular, seeing `{:style (style/` looks redundant to me.

But you know I'm strongly in favor of the ideals of consistency/responsibility/purity, and this PR looks good to me... Thanks for thinking about these things @ilmotta!

ilmotta:
I very much agree @erikseppanen. Thanks for the thoughtful comment. Do you have any particular point you'd like to be addressed? I'll go over some of your points to help expand the discussion.

>Some functions in style.cljs are already impure, for example, theme colors are you stated. Is it a case of that those impure functions are so prevalent that it's ok for them to be there?

I would say they're controlled exceptions. I reviewed all `style.cljs` files, and basically the only stateful things we do are checks for the current `platform` (a very static piece of information actually) and the current theme state (used to get the correct color)..

**Calls to `reanimated/apply-animations-to-style` are the new exception**. I didn't document this in the guidelines, maybe I should (?)

The function `reanimated/apply-animations-to-style` uses a hook behind the scenes, which in turn uses a worklet called `applyAnimationsToStyle`. In the PR description I mentioned the [original thread](https://github.com/status-im/status-mobile/pull/14925#discussion_r1090485454) where I questioned @OmarBasem about the practice of using hooks (stateful by nature), but upon further refactoring I realized allowing animations to build up styles still follows the purpose of decluttering **style data** out of view namespaces. It seems like a natural progression, so now I agree with @OmarBasem and @briansztamfater, who first started this practice.

>I conceptualized style.cljs as a namespace where all the style-related responsibilities can be maintained

Precisely, nothing changed about it. I think the purpose of the `style` namespaces are still kept intact, even with the animation calls, because they're so tightly coupled with raw styling of CSS properties.

>That was my reasoning for giving style.cljs the responsibility of getting the image that the view was asking for (I left the keyword image name in the view because it seems useful for identifying what it is in the layout).

It makes sense, but I thought it was violating the purpose of the style namespace, which is, in essence, to build up style maps. For this example you're sharing: the `source` information is not a style.

>Is it a case of that those impure functions are so prevalent that it's ok for them to be there?

The current theme state is the only piece of reactive state I could find in style files. Are you suggesting it would be better to only have pure functions in style namespaces? This would incur in some considerable changes in the codebase for very little return because we only consume one piece of state (the current theme). I think it's fair to say it's a harmless exception.

As the diff in the PR shows, other states should be passed already deref'ed to functions in `style` namespaces.

ilmotta:
>IMO. In particular, seeing {:style (style/ looks redundant to me.

@erikseppanen, if I may suggest, it would be easier to see the before/after in a WIP PR, because there are some tradeoffs to remove this redundancy, including the fact that it would force callers to often use the core `merge` function to build up the map of props in the view layer. So we might be trading one convenience for another inconvenience.

But again, I'd love to see code instead of words, it's just easier for this kind of stuff.

erikseppanen:
> I very much agree @erikseppanen. Thanks for the thoughtful comment. Do you have any particular point you'd like to be addressed?

Thanks for the expanding on your thinking @ilmotta! You covered the points I was curious about.

PR #14952, opened 2023-02-01
Introduce marking all notifications as read in Activity Center

### LINKED ISSUE #14902: Allow user to mark all notifications as read
# Problem

Sometimes users need to quickly mark all notifications as read, so they can start from a pristine state. Currently they need to individually mark notifications as read, which is tedious, especially for users with lots of unread notifications.

### Solution

Give users the option to mark all notifications as read. See in Figma for all the details https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=5305%3A519966&t=ANqnR83VEKQqzHRY-0

### Out of scope

- Notification settings to customize if the user should see a confirmation dialog or not before marking all as read.
- Undo the operation of marking all as read (this should be done in a separate issue).

### LINKED ISSUE #14917: Let users undo the operation of marking all notifications as read
# Feature Issue

Give users the option to undo the action of *marking all notifications as read*.

### Notes

- Check with designers for how long should the user be able to rollback the action.
- This issue should be preferably played after https://github.com/status-im/status-mobile/issues/14902.

### Reference

https://www.figma.com/file/eDfxTa9IoaCMUy5cLTp0ys/Shell-for-Mobile?node-id=5387%3A529826&t=Kz1gAAYpKi1vcJbh-0

smohamedjavid:
fixes #14902 and #14917 

### Summary

This PR introduces Activity Center's long-awaited feature, marking all notifications as read in one tap :tada:. Users can mark all the notifications as `read` and `undo` the action within the time limit.

Screenshot 2023-02-01 at 4 41 58 PM

#### Review notes

- When the user taps `Mark all notifications as read`, we don't broadcast it. Instead, we update the unread notifications in the `app-db` as `read` and set a time limit for the `undo` action. Once the time limit is expired, we broadcast it.
- `Activity Center` forces dark mode for all the components rendered in it. Hence, introduced the `override-theme` key in the components required for this feature.

#### Platforms

- Android
- iOS

#### Areas impacted

- Activity Center

#### Steps to test

- Open Status
- Generate any notification in your Activity Center (Contact Request, Replies, Mentions, Membership,..etc)
- Open the `Activity Center` > `more`
- Tap on `Mark all notifications as read`
- If need to undo, tap on the `undo` button on the toast message

status: ready

### DIFF
--- a/src/status_im2/common/bottom_sheet/view.cljs
+++ b/src/status_im2/common/bottom_sheet/view.cljs
@@ -70,14 +70,14 @@
                (reset! expanded? false))))))))
 
 (defn handle-comp
-  [window-width]
+  [window-width override-theme]
   [rn/view
    {:style {:width            window-width
             :position         :absolute
             :background-color :transparent
             :top              0
             :height           20}}
-   [rn/view {:style (styles/handle)}]])
+   [rn/view {:style (styles/handle override-theme)}]])
 
 (defn bottom-sheet
   [props children]
@@ -90,6 +90,7 @@
          bottom-safe-area-spacing? :bottom-safe-area-spacing?
          selected-item             :selected-item
          is-initially-expanded?    :expanded?
+         override-theme            :override-theme
          :or                       {show-handle?              true
                                     backdrop-dismiss?         true
                                     expandable?               false
@@ -125,7 +126,7 @@
                 window-height (if selected-item (- height 72) height)
                 {:keys [keyboard-shown]} (hooks/use-keyboard)
                 bg-height-expanded (- window-height (:top insets))
-                bg-height (max (min @content-height bg-height-expanded) 200)
+                bg-height (max (min @content-height bg-height-expanded) 109)
                 bottom-sheet-dy (reanimated/use-shared-value 0)
                 pan-y (reanimated/use-shared-value 0)
                 translate-y (.useTranslateY ^js bottom-sheet-js window-height bottom-sheet-dy pan-y)
@@ -154,7 +155,7 @@
                                       close-bottom-sheet
                                       gesture-running?)
                 handle-comp [gesture/gesture-detector {:gesture bottom-sheet-gesture}
-                             [handle-comp window-width]]]
+                             [handle-comp window-width override-theme]]]
 
             (react/effect! #(do
                               (cond
@@ -221,9 +222,9 @@
                                  :height window-height})}
                        [rn/view {:style styles/container}
                         (when selected-item
-                          [rn/view {:style (styles/selected-background)}
+                          [rn/view {:style (styles/selected-background override-theme)}
                            [selected-item]])
-                        [rn/view {:style (styles/background)}
+                        [rn/view {:style (styles/background override-theme)}
                          [rn/keyboard-avoiding-view
                           {:behaviour (if platform/ios? :padding :height)
                            :style     {:flex 1}}
--- a/src/status_im2/constants.cljs
+++ b/src/status_im2/constants.cljs
@@ -36,6 +36,8 @@
 (def ^:const activity-center-membership-status-accepted 2)
 (def ^:const activity-center-membership-status-declined 3)
 
+(def ^:const activity-center-mark-all-as-read-undo-time-limit-ms 4000)
+
 (def ^:const emoji-reaction-love 1)
 (def ^:const emoji-reaction-thumbs-up 2)
 (def ^:const emoji-reaction-thumbs-down 3)
--- a/src/status_im2/contexts/activity_center/events.cljs
+++ b/src/status_im2/contexts/activity_center/events.cljs
@@ -4,7 +4,8 @@
             [status-im2.contexts.activity-center.notification-types :as types]
             [status-im2.contexts.chat.events :as chat.events]
             [taoensso.timbre :as log]
-            [utils.re-frame :as rf]))
+            [utils.re-frame :as rf]
+            [status-im2.constants :as constants]))
 
 (def defaults
   {:filter-status          :unread
@@ -152,6 +153,66 @@
   {:events [:activity-center.notifications/mark-as-read-success]}
   [cofx notification]
   (notifications-reconcile cofx [(assoc notification :read true)]))
+
+(rf/defn mark-all-as-read
+  {:events [:activity-center.notifications/mark-all-as-read]}
+  [{:keys [db now]}]
+  (when-let [undoable-till (get-in db [:activity-center :mark-all-as-read-undoable-till])]
+    (when (>= now undoable-till)
+      {:json-rpc/call [{:method     "wakuext_markAllActivityCenterNotificationsRead"
+                        :params     []
+                        :on-success #(rf/dispatch
+                                      [:activity-center.notifications/mark-all-as-read-success])
+                        :on-error   #(rf/dispatch [:activity-center/process-notification-failure
+                                                   nil
+                                                   :notification/mark-all-as-read
+                                                   %])}]})))
+
+(rf/defn mark-all-as-read-success
+  {:events [:activity-center.notifications/mark-all-as-read-success]}
+  [{:keys [db]}]
+  {:db (-> (reduce (fn [acc notification-type]
+                     (assoc-in acc [:activity-center :unread-counts-by-type notification-type] 0))
+                   db
+                   types/all-supported)
+           (update :activity-center dissoc :mark-all-as-read-undoable-till))})
+
+(rf/defn undo-mark-all-as-read
+  {:events [:activity-center.notifications/undo-mark-all-as-read-locally]}
+  [{:keys [db]} {:keys [notifications]}]
+  {:db (-> db
+           (update-in [:activity-center :notifications]
+                      update-notifications
+                      notifications)
+           (update :activity-center dissoc :mark-all-as-read-undoable-till))})
+
+(rf/defn mark-all-as-read-locally
+  {:events [:activity-center.notifications/mark-all-as-read-locally]}
+  [{:keys [db now]} get-toast-ui-props]
+  (let [unread-notifications (get-in db [:activity-center :notifications types/no-type :unread :data])
+        undo-time-limit-ms   constants/activity-center-mark-all-as-read-undo-time-limit-ms
+        undoable-till        (+ now undo-time-limit-ms)]
+    {:db                   (-> db
+                               (update-in [:activity-center :notifications]
+                                          update-notifications
+                                          (data-store.activities/mark-notifications-as-read
+                                           unread-notifications))
+                               (assoc-in [:activity-center :mark-all-as-read-undoable-till]
+                                         undoable-till))
+     :dispatch             [:toasts/upsert
+                            (merge
+                             {:id            :activity-center-mark-all-as-read
+                              :duration      undo-time-limit-ms
+                              :undo-duration (/ undo-time-limit-ms 1000)
+                              :undo-on-press
+                              (fn []
+                                (rf/dispatch
+                                 [:activity-center.notifications/undo-mark-all-as-read-locally
+                                  {:notifications unread-notifications}])
+                                (rf/dispatch [:toasts/close :activity-center-mark-all-as-read]))}
+                             (get-toast-ui-props))]
+     :utils/dispatch-later [{:dispatch [:activity-center.notifications/mark-all-as-read]
+                             :ms       undo-time-limit-ms}]}))
 
 ;;;; Acceptance/dismissal
 
--- a/src/status_im2/contexts/activity_center/events.cljs
+++ b/src/status_im2/contexts/activity_center/events.cljs
@@ -201,8 +201,8 @@
                                          undoable-till))
      :dispatch             [:toasts/upsert
                             (merge
-                             {:id            :activity-center-mark-all-as-read
-                              :duration      undo-time-limit-ms
+                             {:id :activity-center-mark-all-as-read
+                              :duration undo-time-limit-ms
                               :undo-duration (/ undo-time-limit-ms 1000)
                               :undo-on-press
                               (fn []
--- a/src/status_im2/contexts/activity_center/view.cljs
+++ b/src/status_im2/contexts/activity_center/view.cljs
@@ -1,7 +1,7 @@
 (ns status-im2.contexts.activity-center.view
-  (:require [utils.i18n :as i18n]
-            [quo.react :as react]
+  (:require [quo.react :as react]
             [quo2.core :as quo]
+            [quo2.foundations.colors :as colors]
             [react-native.core :as rn]
             [react-native.safe-area :as safe-area]
             [status-im2.contexts.activity-center.notification-types :as types]
@@ -13,6 +13,7 @@
             [status-im2.contexts.activity-center.notification.mentions.view :as mentions]
             [status-im2.contexts.activity-center.notification.reply.view :as reply]
             [status-im2.contexts.activity-center.style :as style]
+            [utils.i18n :as i18n]
             [utils.re-frame :as rf]))
 
 (defn filter-selector-read-toggle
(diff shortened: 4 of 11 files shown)


src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,74 @@
...
+          []
+          notifications))
+
+
+;;;; Mark all notifications as read
+
+(rf/defn mark-all-as-read
+  {:events [:activity-center.notifications/mark-all-as-read]}
+  [{:keys [db] :as cofx}]
+  (when-let [undoable-till (get-in db [:activity-center :mark-all-as-read-undoable-till])]
```
J-Son89, line 160: should updating the db not be in the `on-success` method? 🤔

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,74 @@
...
 
+(defn- update-read-flag
+  [notifications read?]
+  (reduce (fn [acc notification]
+            (conj acc (assoc notification :read read?)))
+          []
+          notifications))
+
+
+;;;; Mark all notifications as read
```
J-Son89: comment is redundant imo, consider removing 👍

src/status_im2/contexts/activity_center/view.cljs
```diff
@@ -85,15 +101,25 @@
...
+      :on-press            #(rf/dispatch [:hide-popover])}
+     :i/close]
+    [quo/button
+     {:icon                true
+      :type                :blur-bg
+      :size                32
+      :accessibility-label :activity-center-open-more
+      :override-theme      :dark
+      :on-press            #(rf/dispatch [:bottom-sheet/show-sheet
+                                          {:content        render-more
```
J-Son89: I'm a bit confused what `render-more` is? 🤔 
I would have assumed it was some form of pagination function but it seems to be content in it's own bottom sheet. Is there a better way to describe what this use case is?

smohamedjavid: Yes, It's content for the bottom sheet.

I agree that the naming (`render-more`) is misleading. I will change it to something more meaningful. 👍

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Moved 👍

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Removed 👍

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+                            {:id             :activity-center-mark-all-as-read
+                             :icon           :up-to-date
+                             :icon-color     colors/success-50
+                             :text           (i18n/label :t/notifications-marked-as-read
+                                                         {:count unread-count})
+                             :override-theme :dark
+                             :duration       undo-time-limit-ms
+                             :undo-duration  (/ undo-time-limit-ms 1000)
+                             :undo-on-press
+                             #(do
```
ilmotta: There are some instances of this code style issue :)

https://guide.clojure.style/#no-multiple-forms-fn-literals

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+           (assoc-in [:activity-center :unread-count] unread-count)
+           (update-in [:activity-center] dissoc :mark-all-as-read-undoable-till))})
+
+(rf/defn mark-all-as-read-locally
+  {:events [:activity-center.notifications/mark-all-as-read-locally]}
+  [{:keys [db] :as cofx}]
+  (let [unread-count         (get-in db [:activity-center :unread-count])
+        unread-notifications (get-in db [:activity-center :notifications types/no-type :unread :data])
+        undo-time-limit-ms   constants/activity-center-mark-all-as-read-undo-time-limit-ms
+        undoable-till        (+ (datetime/timestamp) undo-time-limit-ms)]
```
ilmotta: You can get the current timestamp via the co-effect `:now`, i.e. just destructure the key `:now` alongside the `:db`. It's injected by an interceptor added automatically when we use `rf/defn`. See https://github.com/status-im/status-mobile/blob/119d13645079014b2be8b9ddb6e6dbaa35503389/src/utils/re_frame.cljs#L21

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+    {:db                   (-> db
+                               (update-in [:activity-center :notifications]
+                                          update-notifications
+                                          (update-read-flag unread-notifications true))
+                               (assoc-in [:activity-center :unread-count] 0)
+                               (assoc-in [:activity-center :mark-all-as-read-undoable-till]
+                                         undoable-till))
+     :dispatch             [:toasts/upsert
```
ilmotta, line 202: It's interesting to see the solution to display toasts is unnecessarily coupling the view layer with state management. I think it's a fair solution given other places do the same, but there's a better way, even more *React-y & reframe-y*.

I'll try to explain, but **you really don't need to implement my suggestion at all :)** I'm throwing the idea here to help get something going for us, see if you have an opinion, etc.

1. So first, `mark-all-as-read-locally` is dispatched and the app db is updated, but the effect `:toasts/upsert` is NOT dispatched. The event handler simply adds to the app state whatever information is necessary for the view component to subscribe and be able to dispatch.

2. In the view component, a `useEffect` or other means is used to listen to changes coming from a subscription and the necessary event handler is dispatched to display the toast.

This idea could lead to a simpler design overall. Hard to say without trying ;) The event handler wouldn't need to care about colors, translations, etc. This would be all the responsibility of the view. The same tactic could be used for other toasts in the app.

src/status_im2/contexts/activity_center/view.cljs
```diff
@@ -27,6 +27,22 @@
...
+  (let [unread-count (rf/sub [:activity-center/unread-count])]
+    [quo/action-drawer
+     [[(when (pos? unread-count)
+         {:icon           :i/check
+          :override-theme :dark
+          :label          (i18n/label :t/mark-all-notifications-as-read)
+          :on-press       #(do
+                             (rf/dispatch [:bottom-sheet/hide])
+                             (rf/dispatch [:activity-center.notifications/mark-all-as-read-locally]))})
+       {:icon           :i/settings
```
ilmotta: @smohamedjavid Why the urge to implement the dummy `settings` option? To me, it would be better to have a separate issue to implement support for customizing the settings and skip the whole dummy implementation.

I even think it aligns better with the product strategy because the PO constantly says we want to release only things that actually work and are functional.

Maybe I'm over extending this topic a bit too much given we don't have real users yet :man_shrugging: But soon enough we will have, and I find disconcerting to release dummy UI elements to end-users. We don't need to agree, just sharing my 2 cents Javid. This feedback is not exclusive to your PR, I've see this kind of approach a lot in others too.

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+            (conj acc (assoc notification :read read?)))
+          []
+          notifications))
+
+(rf/defn mark-all-as-read
+  {:events [:activity-center.notifications/mark-all-as-read]}
+  [{:keys [db] :as cofx}]
+  (when-let [undoable-till (get-in db [:activity-center :mark-all-as-read-undoable-till])]
+    (when (>= (datetime/timestamp) undoable-till)
+      {:db            (update-in db [:activity-center] dissoc :mark-all-as-read-undoable-till)
```
ilmotta: `update-in` can be simplified to `update` here and in other places in this namespace.

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
   [cofx notification]
   (notifications-reconcile cofx [(assoc notification :read true)]))
 
+(defn- update-read-flag
```
ilmotta: You can simplify and optimize this function to use `map`. Another more important suggestion IMO is to move such logic to the *data store* namespaces, e.g. `status-im.data-store.activities`.

```clojure
(map #(assoc % :read read?) notifications)
```

Or even better in `status-im.data-store.activities`:

```clojure
(defn mark-notifications-as-read
  [notifications]
  (map #(assoc % :read true) notifications))
```

Maybe this is not the best example, given how trivial it is, but there are many more interesting cases lurking around, just waiting for the refactor slayer... :skull: 

I see a lot of potential for such refactorings, so we can keep the state management layer simpler, allowing re-frame to do what it does best, but reducing the amount of knowledge it has about the domain. I'm just cc'ing @erikseppanen because he's always into this kind of discussion ;)

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+                                                   %])}]})))
+
+(rf/defn mark-all-as-read-success
+  {:events [:activity-center.notifications/mark-all-as-read-success]}
+  [{:keys [db]}]
+  {:db (update-in db [:activity-center] dissoc :mark-all-as-read-undoable-till)})
+
+(rf/defn undo-mark-all-as-read
+  {:events [:activity-center.notifications/undo-mark-all-as-read-locally]}
+  [{:keys [db] :as cofx} {:keys [unread-count notifications]}]
```
ilmotta: It looks like `cofx` is not used here. Notice `clj-kondo` doesn't catch this because the `rf/defn` macro is not understood by the linter. Just another downside of this whole `rf/defn` pattern created in `status-mobile`... Oh my...

src/status_im2/constants.cljs
```diff
@@ -36,6 +36,8 @@
 (def ^:const activity-center-membership-status-accepted 2)
 (def ^:const activity-center-membership-status-declined 3)
 
+(def ^:const activity-center-mark-all-as-read-undo-time-limit-ms 4000)
```
ilmotta, line 39: Nothing for you to do in this PR, but eventually it would be good to extract all activity center constants and move them into `status-im2.contexts.activity-center.constants`. That's one of the goals of these `contexts` directory, to keep related things closer to each other.

src/status_im2/contexts/activity_center/events.cljs
```diff
@@ -153,6 +157,76 @@
...
+  {:db (-> db
+           (update-in [:activity-center :notifications]
+                      update-notifications
+                      notifications)
+           (assoc-in [:activity-center :unread-count] unread-count)
```
ilmotta: Unfortunately this must change now @smohamedjavid. The `[:activity-center :unread-count]` state is no longer usable, it's now `[:activity-center :unread-counts-by-type]`. That's because every tab needs a separate count of the unread notifications. You can play around with the app and you should be able to reproduce that the unread indicator in any tab is not updated correctly after marking all notifications as read, nor the undo works as expected.

The new state is a map of *notification type* => *unread count*. See the unit test https://github.com/status-im/status-mobile/blob/develop/src/status_im2/contexts/activity_center/events_test.cljs#L709

Behind the scenes, the subscription `:activity-center/unread-count` sums all the counts to give a total to the UI. See https://github.com/status-im/status-mobile/blob/develop/src/status_im2/subs/activity_center.cljs#L30

ilmotta (commented):
@smohamedjavid, very cool to see this feature implemented along with the undo functionality!

I have a bunch of comments, but the PR is almost there!

I've noticed there's extra whitespace in the bottom sheet that differs from what's in Figma.

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Changed to `fn` literals. 👍

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Replaced the usage of `datetime/timestamp` with `now` (from `cofx`).

src/status_im2/contexts/activity_center/view.cljs
smohamedjavid: Totally agreed on it 💯. 

Removed the `Notification Settings` action from the drawer.

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Removed it 👍

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: Apologies, I missed out the `unread-count-by-type`.  During the initial implementation, it was not added into the `develop`. 

No worries, I updated the PR with the updation of unread count by each type upon marking all notifications as read.

Thanks for letting me know, Icaro!

src/status_im2/contexts/activity_center/events.cljs
smohamedjavid: For this specific feature issue, we are dispatching the actual `mark-all-as-read` event after a specific time along with toast. So, dispatching the toast from the view would put us back in this situation. 

Moved the UI related props to respective view to send those props in a callback function when dispatching `mark-all-as-read-locally`. This helps us to get those props from the callback inside the event handler and the event handlers doesn't need to worry about the UI props.

src/status_im2/common/bottom_sheet/view.cljs
```diff
@@ -123,7 +124,7 @@
                 window-height (if selected-item (- height 72) height)
                 {:keys [keyboard-shown]} (hooks/use-keyboard)
                 bg-height-expanded (- window-height (:top insets))
-                bg-height (max (min @content-height bg-height-expanded) 200)
+                bg-height (max (min @content-height bg-height-expanded) 109)
```
smohamedjavid, line 129: **PROBLEM:**

The drawer has a min height of `200`, which looks good if the bottom sheet has 3 or more `action-drawer` items. If we *reduce* the number of `action-drawer` items to 2 or below. It leaves a space at the bottom.

imgonline-com-ua-twotoone-kbxQHR9h4j copy

**SOLUTION:**

I reduced the minimum height to `109` (as per Figma) to accommodate one action-drawer item and leave unwanted spaces at the bottom.

cc: @ilmotta - https://github.com/status-im/status-mobile/pull/14952#pullrequestreview-1279538720

ilmotta: Excellent @smohamedjavid!

ilmotta (approved):

pavloburykh:
@smohamedjavid thanx for the PR! Could you please rebase PR to the latest develop and then move it to E2E column when it will be ready for testing? thanx!

pavloburykh:
@smohamedjavid please, take a look at the following issue. 

### ISSUE 1 Mentions and replies remain to be unread in chats despite they have been marked as read in AC

**Steps:**

1. Generate some unread mentions and replies notifications
2. Mark all notifications as read
3. .Check if there are any unread counters for unread mentions/replies in chats/community channels

**Actual result:** Mentions and replies remain to be unread in chats, unread counters are still displayed in chats/channels

[Status-debug-logs (1).zip](https://github.com/status-im/status-mobile/files/10756922/Status-debug-logs.1.zip)

https://user-images.githubusercontent.com/97245802/219391867-9605f853-2135-4b16-b4b5-144d3f7af07e.mp4


**Expected result:** unread counters for mentions/replies should disappear in chats after they have been marked as read in AC

pavloburykh:
@smohamedjavid hi! Could you please take a look at the above [issue](https://github.com/status-im/status-mobile/pull/14952#issuecomment-1433175472) when you have a chance? Thanx!

smohamedjavid:
@pavloburykh - This is a known issue across the platform when you mark any mention/replies notification as `read`. The syncing of the `read` state between chat and notification is handled in a separate issue https://github.com/status-im/status-desktop/issues/9349.

Please be informed that PR #15106 will have this same issue.

pavloburykh:
> @pavloburykh - This is a known issue across the platform when you mark any mention/replies notification as `read`. The syncing of the `read` state between chat and notification is handled in a separate issue [status-im/status-desktop#9349](https://github.com/status-im/status-desktop/issues/9349).

Thanx for clearing out @smohamedjavid 
Based on your answer PR is ready to be merged, as no other issues have been found.

PR #14957, opened 2023-02-01
fix compatibility with ConvertToKeycardAccount after recent changes

rasom:
status: ready

### DIFF
--- a/src/status_im2/common/bottom_sheet/events.cljs
+++ b/src/status_im2/common/bottom_sheet/events.cljs
@@ -3,7 +3,7 @@
 
 (rf/defn show-bottom-sheet
   [{:keys [db]} {:keys [view options]}]
-  {:dispatch                  [:dismiss-keyboard]
+  {:dispatch-n                [[:dismiss-keyboard]]
    :show-bottom-sheet-overlay nil
    :db                        (assoc db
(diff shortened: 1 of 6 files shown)


src/status_im2/common/bottom_sheet/events.cljs
```diff
@@ -3,7 +3,7 @@
 
 (rf/defn show-bottom-sheet
   [{:keys [db]} {:keys [view options]}]
-  {:dispatch                  [:dismiss-keyboard]
+  {:dispatch                  [[:dismiss-keyboard]]
```
siddarthkay: is this needed?

siddarthkay, line 6: is this needed?

rasom: well, yep, I think it is although it has to be changed to `:dispatch-n`.

The problem is that when you merge this with other effects some of them might already contain `:dispatch` and this will cause some problems (it did cause them with keycard code). `:dispatch-n` is safer in this case I suppose

siddarthkay: coool.

siddarthkay (approved):

Issue #14480, opened 2022-12-02
Validate ENS name when adding contact


cammellos:
As a user
I want to check whether the ENS name I have inputted resolves
So that I can correct any mistake

### Description

If the user inputs a valid ENS name format (.stateofus.eth & .eth), we should try to resolve the ENS name, if it doesn't resolve, we should show an error to the user (ENS name not found), any change in the input should clear the error.

### Designs

https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=1694%3A380190&t=D6cAP4sUcnKzVSgH-0

PR #14964, opened 2023-02-01
Support join.status.im links for adding a contact (#14814)

### LINKED ISSUE #14814: Support join.status.im links for adding a contact
As a user
I would like to add a contact by typing their join.status.im link

### Description

If the [https://join.status.im/u/{key/ens-name}](https://join.status.im/u/%7Bkey/ens-name%7D) link is valid, we should enable the View Profile button, which if clicked, will take you to the user's profile.

erikseppanen:
fixes #14814

- Add new contact

status: ready

### DIFF
--- a/src/status_im2/contexts/add_new_contact/events.cljs
+++ b/src/status_im2/contexts/add_new_contact/events.cljs
@@ -10,11 +10,6 @@
             [status-im.utils.utils :as utils]))
 
 (re-frame/reg-fx
- :contacts/resolve-public-key-from-ens-name
- (fn [{:keys [chain-id ens-name on-success on-error]}]
-   (ens/pubkey chain-id ens-name on-success on-error)))
-
-(re-frame/reg-fx
  :contacts/decompress-public-key
  (fn [{:keys [public-key on-success on-error]}]
    (status/compressed-key->public-key
@@ -23,50 +18,73 @@
       (let [{:keys [error]} (types/json->clj resp)]
         (if error
           (on-error error)
-          (on-success
-           (str "0x" (subs resp 5)))))))))
+          (on-success (str "0x" (subs resp 5)))))))))
+
+(re-frame/reg-fx
+ :contacts/resolve-public-key-from-ens-name
+ (fn [{:keys [chain-id ens-name on-success on-error]}]
+   (ens/pubkey chain-id ens-name on-success on-error)))
+
+(defn fx-callbacks
+  [input ens-name]
+  {:on-success (fn [pubkey]
+                 (rf/dispatch
+                  [:contacts/set-new-identity-success
+                   input ens-name pubkey]))
+   :on-error   (fn [err]
+                 (rf/dispatch
+                  [:contacts/set-new-identity-error err input]))})
+
+(defn identify-type
+  [input]
+  (let [regex           #"https?://join.status.im/u/(.+)"
+        id              (as-> (utils/safe-trim input) $
+                          (if-some [[_ match] (re-matches regex $)]
+                            match
+                            $)
+                          (if (empty? $) nil $))
+        public-key?     (validators/valid-public-key? id)
+        compressed-key? (validators/valid-compressed-key? id)
+        type            (cond (empty? id)     :empty
+                              public-key?     :public-key
+                              compressed-key? :compressed-key
+                              :else           :ens-name)
+        ens-name        (when (= type :ens-name)
+                          (stateofus/ens-name-parse id))]
+    {:input    input
+     :id       id
+     :type     type
+     :ens-name ens-name}))
 
 (rf/defn set-new-identity
   {:events [:contacts/set-new-identity]}
   [{:keys [db]} input]
-  (let [input           (utils/safe-trim input)
-        public-key?     (validators/valid-public-key? input)
-        compressed-key? (validators/valid-compressed-key? input)
-        ens-name        (if compressed-key? nil (stateofus/ens-name-parse input))
-        on-success      (fn [pubkey]
-                          (rf/dispatch
-                           [:contacts/set-new-identity-success
-                            input ens-name pubkey]))
-        on-error        (fn [err]
-                          (rf/dispatch
-                           [:contacts/set-new-identity-error err input]))]
-    (cond
-      (empty? input) {:db (dissoc db :contacts/new-identity)}
-      public-key?    {:db (assoc db
-                                 :contacts/new-identity
-                                 {:input      input
-                                  :public-key input
-                                  :ens-name   nil
-                                  :state      :error
-                                  :error      :uncompressed-key})}
-      :else
-      (cond->
-        {:db (assoc db
-                    :contacts/new-identity
-                    {:input      input
-                     :public-key nil
-                     :ens-name   nil
-                     :state      :searching
-                     :error      nil})}
-        compressed-key?       (assoc :contacts/decompress-public-key
-                                     {:public-key input
-                                      :on-success on-success
-                                      :on-error   on-error})
-        (not compressed-key?) (assoc :contacts/resolve-public-key-from-ens-name
-                                     {:chain-id   (ethereum/chain-id db)
-                                      :ens-name   ens-name
-                                      :on-success on-success
-                                      :on-error   on-error})))))
+  (let [{:keys [input id type ens-name]} (identify-type input)]
+    (case type
+      :empty          {:db (dissoc db :contacts/new-identity)}
+      :public-key     {:db (assoc db
+                                  :contacts/new-identity
+                                  {:input      input
+                                   :public-key id
+                                   :state      :error
+                                   :error      :uncompressed-key})}
+      :compressed-key {:db
+                       (assoc db
+                              :contacts/new-identity
+                              {:input input
+                               :state :searching})
+                       :contacts/decompress-public-key
+                       (merge {:public-key id}
+                              (fx-callbacks id ens-name))}
+      :ens-name       {:db
+                       (assoc db
+                              :contacts/new-identity
+                              {:input input
+                               :state :searching})
+                       :contacts/resolve-public-key-from-ens-name
+                       (merge {:chain-id (ethereum/chain-id db)
+                               :ens-name ens-name}
+                              (fx-callbacks id ens-name))})))
 
 (rf/defn set-new-identity-success
   {:events [:contacts/set-new-identity-success]}
@@ -76,19 +94,16 @@
               {:input      input
                :public-key pubkey
                :ens-name   ens-name
-               :state      :valid
-               :error      nil})})
+               :state      :valid})})
 
 (rf/defn set-new-identity-error
   {:events [:contacts/set-new-identity-error]}
   [{:keys [db]} error input]
   {:db (assoc db
               :contacts/new-identity
-              {:input      input
-               :public-key nil
-               :ens-name   nil
-               :state      :error
-               :error      :invalid})})
+              {:input input
+               :state :error
+               :error :invalid})})
 
 (rf/defn clear-new-identity
   {:events [:contacts/clear-new-identity :contacts/new-chat-focus]}
--- a/src/status_im2/contexts/add_new_contact/events.cljs
+++ b/src/status_im2/contexts/add_new_contact/events.cljs
@@ -37,7 +37,7 @@
 
 (defn identify-type
   [input]
-  (let [regex           #"https?://join.status.im/u/(.+)"
+  (let [regex           #"^https?://join.status.im/u/(.+)"
         id              (as-> (utils/safe-trim input) $
                           (if-some [[_ match] (re-matches regex $)]
                             match
--- a/src/status_im2/contexts/add_new_contact/events_test.cljs
+++ b/src/status_im2/contexts/add_new_contact/events_test.cljs
@@ -2,52 +2,57 @@
   (:require [cljs.test :refer-macros [deftest is]]
             [status-im2.contexts.add-new-contact.events :as core]))
 
-(def init-db
-  {:networks/current-network "mainnet_rpc"
-   :networks/networks        {"mainnet_rpc"
-                              {:id     "mainnet_rpc"
-                               :config {:NetworkId 1}}}})
+(def ukey
+  "0x045596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce694f1bf2d8773cd59f4dd0863e951f9b7f7351c5516291a0fceb73f8c392a0e88")
+(def ckey "zQ3shWj4WaBdf2zYKCkXe6PHxDxNTzZyid1i75879Ue9cX9gA")
+(def ens "esep")
 
-(defn search-db
-  [input]
-  {:contacts/new-identity {:input      input
-                           :public-key nil
-                           :ens-name   nil
-                           :state      :searching
-                           :error      nil}})
+(deftest identify-empty-string
+  (is (= (core/identify-type "")
+         {:input    ""
+          :id       nil
+          :type     :empty
+          :ens-name nil})))
 
 (deftest search-empty-string
-  (let [input    ""
-        expected {:db init-db}
-        actual   (core/set-new-identity {:db init-db} input)]
-    (is (= actual expected))))
+  (is (= (core/set-new-identity {:db {:contacts/new-identity :foo}} "")
+         {:db {}})))
 
-(deftest search-ens
-  (let [input    "esep"
-        clean    (fn [db]
-                   (-> db
-                       (assoc-in [:contacts/resolve-public-key-from-ens-name :on-success] nil)
-                       (assoc-in [:contacts/resolve-public-key-from-ens-name :on-error] nil)))
-        expected {:db                                        (merge init-db (search-db input))
-                  :contacts/resolve-public-key-from-ens-name
-                  {:chain-id   1
-                   :ens-name   (str input ".stateofus.eth")
-                   :on-success nil
-                   :on-error   nil}}
-        actual   (core/set-new-identity {:db init-db} input)]
-    (is (= (clean actual) expected))))
+(deftest identify-uncompressed-key
+  (is (= (core/identify-type ukey)
+         {:input    ukey
+          :id       ukey
+          :type     :public-key
+          :ens-name nil})))
 
-(deftest search-compressed-key
-  (let [input    "zQ3shWj4WaBdf2zYKCkXe6PHxDxNTzZyid1i75879Ue9cX9gA"
-        clean    (fn [db]
-                   (-> db
-                       (assoc-in [:contacts/decompress-public-key :on-success] nil)
-                       (assoc-in [:contacts/decompress-public-key :on-error] nil)))
-        expected {:db                             (merge init-db (search-db input))
-                  :contacts/decompress-public-key
-                  {:public-key input
-                   :on-success nil
-                   :on-error   nil}}
-        actual   (core/set-new-identity {:db init-db} input)]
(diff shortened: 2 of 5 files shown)


src/status_im2/contexts/add_new_contact/events.cljs
```diff
@@ -23,50 +18,73 @@
...
+  [input ens-name]
+  {:on-success (fn [pubkey]
+                 (rf/dispatch
+                  [:contacts/set-new-identity-success
+                   input ens-name pubkey]))
+   :on-error   (fn [err]
+                 (rf/dispatch
+                  [:contacts/set-new-identity-error err input]))})
+
+(defn identify-type
```
ilmotta, line 38: I think this function deserves to be extracted out of the events namespace. The data store directory is an interesting place to move domain-only concerns. You could also extract to an utility namespace too.

I find event namespaces already have too much going on, too many responsibilities and the complexities associated with effects, state management, asynchronicity, toasts, etc. It's also a namespace that doesn't promote reusability, because in general we require it only once to register event/effects/etc in the re-frame registries.

src/status_im2/contexts/add_new_contact/events_test.cljs
```diff
@@ -2,52 +2,57 @@
...
-                  :contacts/decompress-public-key
-                  {:public-key input
-                   :on-success nil
-                   :on-error   nil}}
-        actual   (core/set-new-identity {:db init-db} input)]
-    (is (= (clean actual) expected))))
+  (is (= (core/set-new-identity {:db {:contacts/new-identity :foo}} "")
+         {:db {}})))
+
+(deftest identify-uncompressed-key
```
ilmotta: It's generally recommended to add a suffix `-test` because `deftest` creates a var in the test namespace that can shadow whatever was required. It's not a problem most of the time though.

src/status_im2/contexts/add_new_contact/events_test.cljs
```diff
@@ -2,52 +2,57 @@
...
+
+(deftest search-uncompressed-key
+  (is (= (core/set-new-identity {:db {}} ukey)
+         {:db {:contacts/new-identity
+               {:input      ukey
+                :public-key ukey
+                :state      :error
+                :error      :uncompressed-key}}})))
+
+(deftest identify-ens
```
ilmotta: For pure functions with concise input/output (such as `identify-type`) I find the `are` macro to be much easier to read and write the tests. It forces the developer to think in terms of examples in a more exhaustive way because there's very little friction to add new examples. No need to come up with unique and long test names, just throw more and more examples for happy/sad paths.

Related to that, I see the tests for `identify-type` are too "optimistic" (happy path only). For instance, the function `identify-type` incorrectly identifies anything that's not `empty|public-key|compressed-key` as an `:ens-name`. Having an easier way to add more test cases could've helped you catch this kind of problem.

Going further, I think the `ens` validation is flawed. I think we should follow the specification https://eips.ethereum.org/EIPS/eip-137#name-syntax and never fallback to `ens`. The event handler that sets the identity should be more restrictive, otherwise the app state will be incorrect and other problems will follow.

The regex for the URL should have an anchor `^` (beginning of line) to avoid matching strings that don't start with the HTTP protocol. Node also offers a native URL parser that would allow us to extract the ID in a more future proof way (for instance if the the invite URL had query params), but I find it unnecessary for this case. Just for the sake of sharing knowledge and because I was curious to refresh my memory:

```clojure
(defn invite-url->id
  [invite-url]
  (when-let [url (and (re-find #"^https?://join.status.im/u/.+" invite-url)
                      (try (new js/URL invite-url)
                           (catch js/Error _ nil)))]
    (let [pathname (.-pathname url)
          id       (nth (string/split pathname "/") 2)]
      id)))

(comment
  (assert (= "some-id" (invite-url->id "https://join.status.im/u/some-id?a=1&b=2"))))
```

src/status_im2/contexts/add_new_contact/events.cljs
```diff
@@ -23,50 +18,73 @@
...
-                                      :on-error   on-error})
-        (not compressed-key?) (assoc :contacts/resolve-public-key-from-ens-name
-                                     {:chain-id   (ethereum/chain-id db)
-                                      :ens-name   ens-name
-                                      :on-success on-success
-                                      :on-error   on-error})))))
+  (let [{:keys [input id type ens-name]} (identify-type input)]
+    (case type
+      :empty          {:db (dissoc db :contacts/new-identity)}
+      :public-key     {:db (assoc db
```
ilmotta, line 65: I noticed the `develop` branch in `status-mobile` stopped accepting public keys when adding contacts. I also tried in this PR's branch and it doesn't work. The input box says "This is not a chatkey" and I can only establish new contact requests by using compressed keys. It's not a problem originated from this PR, but I'm mentioning here because I think this was a regression.

@churik, are you aware of this?

src/status_im2/contexts/add_new_contact/events.cljs
erikseppanen: Thanks for your review @ilmotta! Always a pleasure to hear your feedback!

I'm teetering on agreeing with this. I looked at `status-im/data-store` and `status-im2/utils/validators.cljs` but neither one seems ideal. And looking again at `identify-type`, it's somewhat coupled to the event's validation logic, with the way it extracts a `key|ens` from a url. Maybe when/if it is used by another namespace, it can be moved? ATM it seems convenient to keep it next to the `:contacts/set-new-identity` event.

The reason I extracted this as a function was to make the decision-making logic more testable and to get rid of the extra baggage of dealing with the `:on-success` and `:on-failure` callbacks (which I didn't think had as much value any more, with the new integration test). I hadn't thought of it as something other namespaces would be interested in. WDYT?

src/status_im2/contexts/add_new_contact/events_test.cljs
erikseppanen: > For pure functions with concise input/output (such as `identify-type`) I find the `are` macro to be much easier to read and write the tests. It forces the developer to think in terms of examples in a more exhaustive way because there's very little friction to add new examples. No need to come up with unique and long test names, just throw more and more examples for happy/sad paths.

Yes, i was thinking the same thing when I saw the repetitive `deftest`s, and tried consolidating them under the `are` macro. I ran into some trouble with using the input test vars in the expected output... But I didn't try that hard either. And it didn't look that much better, like it does when you have small integers or strings to compare. I'll give it another shot, sometimes `are` is so much nicer!

erikseppanen: > Related to that, I see the tests for `identify-type` are too "optimistic" (happy path only). For instance, the function `identify-type` incorrectly identifies anything that's not `empty|public-key|compressed-key` as an `:ens-name`. Having an easier way to add more test cases could've helped you catch this kind of problem.

Thanks for looking closely at that! However, I'm not sure if I agree that it's incorrect or a problem. What input do you think is problematic?

erikseppanen: > Going further, I think the `ens` validation is flawed. I think we should follow the specification https://eips.ethereum.org/EIPS/eip-137#name-syntax and never fallback to `ens`. The event handler that sets the identity should be more restrictive, otherwise the app state will be incorrect and other problems will follow.

Ah, ens... I did not put in more extensive validation because:
- This feature belongs to another PR: #14480
- ens names seem to be quite permissive (I'll take another look at it, thanks for the link!)
- status-go will attempt to resolve it anyway

erikseppanen: > The regex for the URL should have an anchor `^` (beginning of line) to avoid matching strings that don't start with the HTTP protocol. Node also offers a native URL parser that would allow us to extract the ID in a more future proof way (for instance if the the invite URL had query params), but I find it unnecessary for this case. Just for the sake of sharing knowledge and because I was curious to refresh my memory:
> 
> ```clojure
> (defn invite-url->id
>   [invite-url]
>   (when-let [url (and (re-find #"^https?://join.status.im/u/.+" invite-url)
>                       (try (new js/URL invite-url)
>                            (catch js/Error _ nil)))]
>     (let [pathname (.-pathname url)
>           id       (nth (string/split pathname "/") 2)]
>       id)))
> 
> (comment
>   (assert (= "some-id" (invite-url->id "https://join.status.im/u/some-id?a=1&b=2"))))
> ```

I had included the caret earlier, but took it out after I started trimming the string. I'll add it back in, np!

Thanks for sharing the code snippet, interesting!

src/status_im2/contexts/add_new_contact/events.cljs
erikseppanen: This would seem like a regression, but it is intended behavior. For example, the "This is not a chatkey" is in the figma, and @cammellos and I both worked specifically on removing support for uncompressed keys.

src/status_im2/contexts/add_new_contact/events.cljs
ilmotta: >The reason I extracted this as a function was to make the decision-making logic more testable and to get rid of the extra baggage of dealing with the :on-success and :on-failure

Nice approach :)

>And looking again at identify-type, it's somewhat coupled to the event's validation logic,

I see `identify-type` does generally 3 things: 1. validation as a means of matching the type of the input, 2. key/id extraction and 3. it returns a map describing the parsed input. This function really looks like some kind of parser given it transforms a raw invite URL into a data structure suitable for further processing. I don't see any coupling between this function and the event handlers, except for the fact that's in the same namespace ;)

I'm not sure if you agree with me, but `parse-invite-url` seems like a better name for it, because the function is not just identifying a type, it's parsing a specific invite URL string -> data representation.

The snippet below should be close to working, and I think it promotes a better separation of responsibilities. Notice the proposal for a new function to correctly validate an ENS name `ens.core/valid?`. I'm sure there's a lib for that too, that'd be even better than reinventing the wheel. Here I'm also suggesting the matcher for the id/key type should return `unknown` when no match is found. It could be `nil` too.

I like to sometimes code for reviewing PRs, please don't interpret this comment as me trying to tell you what you should code :relaxed: 

The original code also reminded me of a tactic for refactoring, I think first shared by Sandi Metz, called [east-oriented coding](https://github.com/btihen/East_Oriented_Code). East goes to the right in the traditional Eurocentric Mercator's map. That means reducing the amount of west-oriented (left-oriented) code. Applied to a functional language like Clojure, it could mean using less binding declarations, and more functions that compose.

```clojure
(defn id-type
  [id]
  (cond
    (validators/valid-public-key? id)     :public-key
    (validators/valid-compressed-key? id) :compressed-key
    (ens/valid? id)                       :ens-name
    :else                                 :unknown))

(defn invite-url->id
  [invite-url]
  (when-let [url (and (re-find #"^https?://join.status.im/u/.+" invite-url)
                      (try (new js/URL invite-url)
                           (catch js/Error _ nil)))]
    (let [pathname (.-pathname url)
          id       (nth (string/split pathname "/") 2)]
      id)))

(defn parse-invite-url
  [invite-url]
  (let [id   (invite-url->id invite-url)
        type (id-type id)]
    {:input    invite-url
     :id       id
     :type     type
     :ens-name (when (= type :ens-name)
                 (stateofus/ens-name-parse id))}))
```

src/status_im2/contexts/add_new_contact/events_test.cljs
ilmotta: Again sharing in case someone is not aware, the `are` macro is quite flexible and you can design the assertions to eliminate duplication because when using the `are` macro, the whole point is to easily describe input + expected.

This passes in the test suite:

```clojure
(deftest identify-type-test
  (are [args] (= (assoc (:expected args) :input (:input args))
                 (core/identify-type (:input args)))
   {:input    ukey
    :expected {:id       ukey
               :type     :public-key
               :ens-name nil}}
   (let [ens-name (str ens ".stateofus.eth")]
     {:input    ens
      :expected {:id       ens
                 :type     :ens-name
                 :ens-name ens-name}})
   (let [input (str "https://join.status.im/u/" ckey)]
     {:input    input
      :expected {:id       ckey
                 :type     :compressed-key
                 :ens-name nil}})))
```

src/status_im2/contexts/add_new_contact/events.cljs
ilmotta: I really remember someone telling me nodes will always support uncompressed and compressed keys, and so should the clients. I'm confused, when I look at the specs in https://specs.status.im/spec/2#public-key-serialization, there's nothing saying uncompressed keys should be avoided or not supported when implementing clients. I'm probably missing something.

src/status_im2/contexts/add_new_contact/events.cljs
erikseppanen: > > And looking again at identify-type, it's somewhat coupled to the event's validation logic,
> 
> I see `identify-type` does generally 3 things: 1. validation as a means of matching the type of the input, 2. key/id extraction and 3. it returns a map describing the parsed input. This function really looks like some kind of parser given it transforms a raw invite URL into a data structure suitable for further processing. I don't see any coupling between this function and the event handlers, except for the fact that's in the same namespace ;)

Yes, sorry I didn't mean coupling in the technical sense, probably should have used a different word, but I mean the function probably won't be used anywhere else (I might be wrong, in which case, I'm suggesting to move it when it is). I meant that it's accepting multiple types of input: compressed key, ens, and links that have compressed keys/ens embedded in them. Some reasons why I'm not in agreement about moving it is that:
- no other namespace needs it right now
- we'd have to give it a longer name, because right now `identify-type` piggy-backs on the `add-new-contact` namespace (we understand it based on it being in that namespace)
- I'd have to open another buffer for it (haha but true)

> I'm not sure if you agree with me, but `parse-invite-url` seems like a better name for it, because the function is not just identifying a type, it's parsing a specific invite URL string -> data representation.

It accepts more than just url's though - it accepts compressed keys and ens names (which are not url's)...

> The snippet below should be close to working, and I think it promotes a better separation of responsibilities. Notice the proposal for a new function to correctly validate an ENS name `ens.core/valid?`. I'm sure there's a lib for that too, that'd be even better than reinventing the wheel. Here I'm also suggesting the matcher for the id/key type should return `unknown` when no match is found. It could be `nil` too.

Yes, I like `ens.core/valid?`, I just didn't work that much on validation because there's another PR for it #14480

src/status_im2/contexts/add_new_contact/events_test.cljs
erikseppanen: Thanks Icaro, I got `are` working, here's what I'll probably commit. (defs at the top for the other tests to use)
``` clojure
(def ukey
  "0x045596a7ff87da36860a84b0908191ce60a504afc94aac93c1abd774f182967ce694f1bf2d8773cd59f4dd0863e951f9b7f7351c5516291a0fceb73f8c392a0e88")
(def ckey "zQ3shWj4WaBdf2zYKCkXe6PHxDxNTzZyid1i75879Ue9cX9gA")
(def ens "esep")
(def link-ckey (str "https://join.status.im/u/" ckey))

(deftest identify-type-test
  (are [input expected] (= (core/identify-type input) expected)
    "" {:input    ""
        :id       nil
        :type     :empty
        :ens-name nil}    

    ukey {:input    ukey
          :id       ukey
          :type     :public-key
          :ens-name nil}

    ens {:input    ens
         :id       ens
         :type     :ens-name
         :ens-name (str ens ".stateofus.eth")}

    ckey {:input    ckey
          :id       ckey
          :type     :compressed-key
          :ens-name nil}

    link-ckey {:input    link-ckey
               :id       ckey
               :type     :compressed-key
               :ens-name nil}))
```

src/status_im2/contexts/add_new_contact/events_test.cljs
erikseppanen: done

src/status_im2/contexts/add_new_contact/events.cljs
ilmotta: >no other namespace else needs it right now

It's a moot point, and as we can see we won't agree. That's cool, we need to disagree more ;)

I believe the code should be designed around the proper layers and abstractions, and not if something is required one or more times. For example, if a utility function is needed to transform a datetime, we put it in the utility datetime namespace, even though it might be required only once. A function that parses a contact ID clearly is not related to state management (event handler layer), even though it's used there.

>I'd have to open another buffer for it (haha but true)

haha okay Jump in and back, `M-.` and `M-,` and don't let Emacs split buffers automatically if you didn't ask. Okay, now I'm diverging too much ;)

>It accepts more than just url's though - it accepts compressed keys and ens names (that are not prefaced by url's)...

True indeed, but `identify-type` is too general. I see you referred to this "thing" as `id`. The function could be named `parse-contact-id`. Par

src/status_im2/contexts/add_new_contact/events_test.cljs
ilmotta: >Ah, ens... I did not put in more extensive validation because:

>- This feature belongs to another PR: https://github.com/status-im/status-mobile/issues/14480
>- ens names seem to be quite permissive (I'll take another look at it, thanks for the link!)
>- status-go will attempt to resolve it anyway

I didn't know about the separate issue. Thanks!

>ens names seem to be quite permissive (I'll take another look at it, thanks for the link!)

Permissive, but not that much as the current implementation does. The spec is more strict. Also I think we shouldn't fallback to type `:ens-name` if we're not sure it's an ENS name.

>status-go will attempt to resolve it anyway

Yes, but the frontend needs to efficiently give the user feedback without going over the slow process of calling the backend, so a strong client side validation is necessary too.

ilmotta (approved):

src/status_im2/contexts/add_new_contact/events.cljs
erikseppanen: OK, well like I said earlier, I was teetering on it, so thanks for explaining your reasoning. I agree with reasoning that once I moved it out of the event handler, it now belonged to the world haha. Go forth little one!

I'll move it to utils.validators where it can join its colleagues `valid-public-key?` and `valid-compressed-key?`. I hope that they can support each other there haha

A good result of this discussion is that I was toying with the idea of calling it `something-id`, so i'll rename it `parse-contact-id`. Thanks!

ilmotta: >it now belonged to the world haha. Go forth little one!

Loved the sarcasm!

Lynyrd Skynyrd - Free Bird (Official Audio)
https://www.youtube.com/watch?v=0LwcvjNJTuM

erikseppanen: To be clear, no sarcasm intended at all (no mocking or contempt on my side that is). You and I both love to discuss the intricacies of how to build software... I'm just adding the humorous metaphor because it's kind of true - I extracted this function to make it testable, but in doing so, it joined the wider code base (the world). Just funny to think of code like that, like a baby/child (bird I guess is what you imagined), making its own out in the world, away from it's cozy protected event handler.

erikseppanen: @ilmotta As discussed, attempting to move this led to a series of circular dependencies. As agreed, this refactoring can happen in a separate PR.

VladimrLitvinenko:
hi @erikseppanen. Thank you for your work. PR is tested and ready to be merged. No issues from my side. Could you merge it, please?

VolodLytvynenko:
hi @erikseppanen. Don't merge this PR yet. I need to investigate e2e tests. I let you know when the tests are passed

VolodLytvynenko:
@erikseppanen thank you for your work. No issues from my side. Could you merge it, please?

briansztamfater (approved):

PR #14969, opened 2023-02-02
 fastlane: remove saucelab upload + upload apk from url in e2e 

churik:
@yevh-berdnyk thank you for help ❤️

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -50,25 +50,6 @@ def retry_curl_upload(url, file, auth, conn_timeout = 5, timeout = 60, retries =
   end
 end
 
-# uploads `file` to sauce labs (overwrites if there is anoter file from the
-# same commit)
-def upload_to_saucelabs(file)
-  key = ENV['SAUCE_ACCESS_KEY']
-  username = ENV['SAUCE_USERNAME']
-  unique_name = ENV['SAUCE_LABS_NAME']
-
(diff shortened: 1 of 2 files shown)


yevh-berdnyk (approved):

jakubgs:
I don't think this change makes sense. There is currently no mechanism to provide the APK to the E2E tests from the `android-e2e` build that creates the `x86` APK for tests. The way it worked before this PR was merged was that the APK uploaded to SauceLabs was downloaded based on PR ID, since that was the name of the APK uploaded: `${PR_ID}.apk`
```groovy
def uploadToSauceLabs(Boolean verbose = false) {
  def changeId = utils.changeId()
  if (changeId != null) {
    env.SAUCE_LABS_NAME = "${changeId}.apk"
  } else {
```
https://github.com/status-im/status-jenkins-lib/blob/bc9ee025/vars/android.groovy#L122-L126

Now that that's gone we need another method. The simplest way I can think of is using [`copyArtifacts()`](https://www.jenkins.io/doc/pipeline/steps/copyartifact/) method.
Since we have the PR ID, and we know the path of `android-e2e` jobs on Jenkins, we can use that to copy the most recent `android-e2e` build for the tests. The issue with that is that it's always the most recent one, not the specific one picked when the e2e tests job was queued, but it appears that's how it already worked with SauceLabs upload, so that's probably not a major issue.

Ping me if you need help implementing use of `copyArtifacts()`.

Issue #14564, opened 2022-12-16
Community join screens - implement bottom sheet with correct scrolling and designs


J-Son89:
See the designs:
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=7643%3A608085&t=bc5QBHpQ3uZT8W6b-0

On the join screens the bottom sheet has some advanced styling. 

This is also dependent on the new bottom sheet being finished as that should be used here so best to wait for this work to be merged: https://github.com/status-im/status-mobile/pull/14209

alwx:
Linked to https://github.com/status-im/status-mobile/issues/14955

churik:
@alwx are you still working on this?

alwx:
@churik just removed the assignment since I probably have too many issues assigned to me

PR #14971, opened 2023-02-02
The [Cancel] and [Join] buttons are hidden within 'Join Community' overlay

### LINKED ISSUE #14955: The [Cancel] and [Join] buttons are hidden  within 'Join Community' overlay
# Bug Report

## Problem 
The [Cancel] and [Join] buttons are out of the screen for some devices on the 'Join Community' overlay

**Preconditions:**
User didn't join to the status community yet

**Steps to reproduce:**
1. Go to the communities page
2. Tap 'Opened' tab
3. Tap Status
4. Tap [Join Community] button
5. Check opened 'Join Community' overlay

#### Actual behavior
[Cancel] and [Join] buttons  are hidden
image

#### Expected behavior
[Cancel] and [Join] buttons  are shown
image

### Additional Information
- Status version: nightly 1 Feb 2023 [2612]
- Operating System: Android version 9, Huawei P20 light

alwx:
Makes 'Join Community' overlay scrollable.
Fixes https://github.com/status-im/status-mobile/issues/14955

### Summary

Because of the safe-area inset that's always added to the overlay on the bottom, scrollable overlays did look a bit ugly — so this PR also adds an option `bottom-safe-area-spacing?` (set to `true` by default) that when set to `false` removes the inset which makes it possible for scrollable components to handle insets manually (for example, inside a scroll-view component itself).

#### Platforms

- Android
- iOS

##### Functional

- bottom sheets

Status: ready


src/status_im2/contexts/communities/menus/request_to_join/view.cljs
```diff
@@ -1,6 +1,7 @@
 (ns status-im2.contexts.communities.menus.request-to-join.view
   (:require [quo.react-native :as rn]
             [quo2.core :as quo]
+            [quo.components.safe-area :as safe-area]
```
flexsurfer: quo shouldn't be used in new designs, we have safe-area in react-native namespace

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
```diff
@@ -1,6 +1,7 @@
 (ns status-im2.contexts.communities.menus.request-to-join.view
   (:require [quo.react-native :as rn]
```
flexsurfer: please replace quo here as well as you're making changes in this ns anyway

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
```diff
@@ -76,57 +77,64 @@
...
-        is-open?         (not= 3 (:access permissions))]
-    (fn []
-      [rn/scroll-view {:style style/page-container}
-       [rn/view
-        {:style style/title-container}
+  (let [agreed-to-rules? (reagent/atom false)]
+    [:f>
+     (fn []
+       (let [{window-height :height} (rn/use-window-dimensions)
+             safe-area               (safe-area/use-safe-area)
```
flexsurfer: if its a drawer component and it's opened in bottom sheet i would say that bottom sheet should take care about safe-area

J-Son89: Yes I agree, this is implemented before the new bottom sheet was ready so it's using the old bottom sheet. @alwx, I think the better solution is to implement this with the new bottom sheet and then we only fix this once.

alwx: @J-Son89 it uses `status-im2.common.bottom-sheet.view` if that's what you're talking about. Is it the new one?

alwx: @flexsurfer how do you propose to do this? The drawer component indeed cares about the safe-area and adds a spacing in the bottom of the screen but the problem is we actually don't need this spacing because our component contains a scroll-view that should all be scrollable (and unnecessary spacing makes it look weird)
In fact, I just implemented it similarly to `status-im2.contexts.chat.group-details.view` which is also a bottom sheet view.

J-Son89: @ibrkhalil can you point us where the new one is :D

alwx: @J-Son89 I do feel like `status-im2.common.bottom-sheet.view` is in fact the new one (the latest changes to it were done a week ago)

J-Son89: yep it looks like it! Sorry for delayed response. It was created by @chrispader and seems he is first on the commits for it 👍

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
alwx: Done!

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
alwx: fixed!

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
alwx: Left it as it was because we need to control `safe-area` within the component for reasons explained above.

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
```diff
@@ -73,60 +74,72 @@
...
-       [rn/view
-        {:style style/title-container}
+  (let [agreed-to-rules? (reagent/atom false)]
+    [safe-area/consumer
+     (fn [insets]
+       [:f>
+        (fn []
+          (let [{window-height :height} (rn/use-window-dimensions)
+                is-open?                (not= 3 (:access permissions))]
+            [rn/scroll-view {:style {:max-height (- window-height (:top insets))}}
```
flexsurfer, line 91: i feel like something wrong is here, why content should care about max height? bottom sheet should care about this cc @ibrkhalil

flexsurfer: also im not sure about `window-height` , bottom sheet should never go so high, probably it should be a modal screen then ?

flexsurfer:
i'll left comment in the issue

alwx:
Ok, closing this one because we decided to implement the new component instead of fixing this bug.
More details here: https://github.com/status-im/status-mobile/issues/14955

alwx:
Reopening the PR because it seems like we want to merge the hotfix before the proper component is implemented.
cc @churik

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
alwx: We will focus on developing a new component later on, now we need a quick fix (as per today's retro meeting). Here is the issue for a separate modal screen: https://github.com/status-im/status-mobile/issues/14564

flexsurfer (approved):

J-Son89 (approved):

PR #14973, opened 2023-02-02
Add ability to cancel a request to join a community

### LINKED ISSUE #14837: Communities - add cancel request feature to a joinable community
As a user I want to be able to request to join a community and then cancel that request.

The end point is available from the backend but needs to set up and then needs to be connected to the corresponding option on the communities menus.

Add any relevant tests

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/14837

TO TEST:

Go to a community that needs admin approval.

request to join that community.

on the community overview page the three dots in the top right hand corner should bring up a menu with a list of options.

The option in red should be there to "Cancel request to join".

clicking this button should cancel the request for the user and the overview page should reset to how it was beforehand.

Additionally, I updated the mechanism to see what pending requests are available. So to check this is still correct, you can go to the Communities Home page and the 'Pending' tab should have the correct number of pending community requests.


ios/Podfile.lock
```diff
@@ -210,8 +210,8 @@ PODS:
     - React
   - react-native-blob-util (0.13.18):
     - React-Core
-  - react-native-blur (3.6.1):
```
J-Son89: will revert 👍

src/status_im2/subs/root.cljs
```diff
@@ -232,6 +232,7 @@
 
 ;; communities
 
+(reg-root-key-sub :my-pending-requests-to-join :my-pending-requests-to-join)
```
J-Son89: so for this we need to get a list of users requests to join different communities from backend.

Was not sure if it is better to rename this to `:my-pending-requests-to-join-communities`.

Or maybe better
```
:my-requests {
   :communities {}
   :some-other-section
}
```

Does anyone know if we will have something similar with active requests that need approval? I guess there is this with contacts etc.

src/status_im2/subs/communities.cljs
```diff
@@ -116,15 +116,16 @@
...
+ :<- [:communities/my-pending-requests-to-join]
  ;; Return communities splitted by level of user participation. Some communities user
  ;; already joined, to some of them join request sent and others were opened one day
  ;; and their data remained in app-db.
  ;; Result map has form: {:joined [id1, id2] :pending [id3, id5] :opened [id4]}"
- (fn [communities]
+ (fn [[communities requests]]
    (reduce (fn [acc community]
              (let [joined?    (:joined community)
-                   requested? (pos? (:requested-to-join-at community))
```
J-Son89: we get active requests from the requests list instead now, rather than the `community` object. Discussed this with @borismelnik 👍

src/status_im2/contexts/communities/menus/community_options/view.cljs
```diff
@@ -82,19 +82,19 @@
...
                                         :content-height 400}])})
 
 (defn cancel-request-to-join
   [id]
-  {:icon                :i/block
-   :label               (i18n/label :t/cancel-request-to-join)
-   :accessibility-label :cancel-request-to-join
-   :danger?             true
-   :on-press            #(js/alert (str "implement action" id))})
+  (let [request-id (rf/sub [:communities/my-pending-request-to-join id])]
```
J-Son89: should this `sub` be higher up the chain?

src/status_im/communities/core.cljs
```diff
@@ -105,6 +105,16 @@
                db
                communities)})
 
+(rf/defn handle-my-pending-requests-to-join
+  {:events [::fetched-my-communities-requests-to-join]}
+  [{:keys [db]} my-requests]
+  {:db (assoc-in db
+        [:my-pending-requests-to-join]
+        (reduce (fn [acc request]
+                  (assoc acc (aget request "communityId") (aget request "id")))
```
J-Son89: these are "js" values, should I convert to keywords? or it's okay as is?

src/status_im/multiaccounts/login/core.cljs
```diff
@@ -403,6 +403,7 @@
               (data-store.chats/fetch-chats-rpc
                {:on-success
                 #(do (re-frame/dispatch [:chats-list/load-success %])
+                     (rf/dispatch [:communities/get-user-requests-to-join])
```
J-Son89, line 408: Discussed with @cammellos, currently the users requests are fetched on login. 
Should we just do this when the user first navigates to the communities home page? 

This list will also be re-fetched when a user cancels a request, or makes a new request.

@ilmotta, @flexsurfer @erikseppanen and everyone else.. wdyt? 🤔

ilmotta: @J-Son89, I think the reasoning is that we only need to fetch once the app initializes to get the initial & latest state from status-go. Later on, the backend will push signals to the mobile client. And all the mobile client has to do is reconcile the new state with the existing app db state.

This is a rather efficient process. So as long as the signals are coming and the mobile logic to process them is correct, things should just work. But are the signals for community requests flowing to the client?

J-Son89: Btw, by currently I'm by comment above. I mean, as of this pr this is how I implemented it. 
This call to get the lost of users requests was not in place before this pr

src/status_im/communities/core.cljs
ilmotta: I think we have in our guidelines that we should use the `oops/oget` macro to interop with JS. It compiles efficiently and use CLJS type hints so we reduce the surface for weird compilation behavior that I've faced already in `status-mobile` when I forgot to add `^js` in the proper place. The macro avoids some of the problems with `aget` (see https://github.com/binaryage/cljs-oops#object-operations, but they don't mention exactly which problems).

Due to resource constraints, I know that some pieces of data flow into the mobile client without being converted to cljs data structures. If you have this concern it might be okay to use oops, otherwise I usually prefer to deal with cljs data structures as early as possible to avoid surprises and polluting the code with interop macros (oops/* or native ones).

So ideally, the `my-requests` data should already come as a cljs data structure, so it's always transparent to the event handler. Doing this also helps avoid the situation of storing native js objects in the app db by mistake, which would force all consumers of the state to remember to repeatably transform them or worry about using interop macros.

J-Son89: Perfeito, yes that's what I was looking for. Will update 👍

src/status_im/multiaccounts/login/core.cljs
ilmotta: I see now. I shared more what would be the ideal scenario IMO, where the mobile client only needs to react to changes coming from the DB after the initialization, hence the client would not make ad-hoc RPC calls to get the most up-to-date state from the backend because this is bug-driven.

src/status_im/communities/core.cljs
```diff
@@ -175,11 +185,38 @@
...
+                                    (re-frame/dispatch [:communities/get-user-requests-to-join]))
                     :on-error    #(do
                                     (log/error "failed to request to join community" community-id %)
                                     (re-frame/dispatch [::failed-to-request-to-join %]))}]})
 
+
+
+(rf/defn get-user-requests-to-join
+  {:events [:communities/get-user-requests-to-join]}
+  [cofx]
```
ilmotta: `cofx` is not being used. Hey, this is a common example for why some macros can be problematic. The linter doesn't understand `cofx` is not used, nor does the Clojure language server.

src/status_im/communities/core.cljs
```diff
@@ -175,11 +185,38 @@
...
+  {:json-rpc/call [{:method      "wakuext_myPendingRequestsToJoin"
+                    :params      []
+                    :js-response true
+                    :on-success  #(do
```
ilmotta: You can remove the surrounding `do` macro.

src/status_im/communities/core.cljs
```diff
@@ -175,11 +185,38 @@
   {:json-rpc/call [{:method      "wakuext_requestToJoinCommunity"
                     :params      [{:communityId community-id}]
                     :js-response true
-                    :on-success  #(re-frame/dispatch [::requested-to-join %])
+                    :on-success  #(do
```
ilmotta: I have this one saved in my favorites, unfortunately the formatter can't automate for us.

By the Clojure Style Guide powers, I invoke `No multiple forms in function literals`! https://guide.clojure.style/#no-multiple-forms-fn-literals

src/status_im/communities/core.cljs
```diff
@@ -175,11 +185,38 @@
...
+                    :params      []
+                    :js-response true
+                    :on-success  #(do
+                                    (re-frame/dispatch [::fetched-my-communities-requests-to-join %]))
+                    :on-error    #(do
+                                    (log/error "failed to get requests to join community" %))}]})
+
+(rf/defn cancel-request-to-join
+  {:events [:communities/cancel-request-to-join]}
+  [cofx request-to-join-id]
```
ilmotta: `cofx` is also not used here.

src/status_im2/subs/communities.cljs
```diff
@@ -150,6 +151,19 @@
     (get communities identity)
     counts)))
 
+(re-frame/reg-sub
+ :communities/my-pending-requests-to-join
+ (fn [[]]
+   [(re-frame/subscribe [:my-pending-requests-to-join])])
```
ilmotta: This one is weird @J-Son89. Could it be a leftover? Can't you just use the arrow `:<-` syntax sugar? The syntax you used here is often used when there's a parameter for the subscription, but there's none.

src/status_im2/subs/communities.cljs
```diff
@@ -150,6 +151,19 @@
...
+ (fn [[requests]]
+   requests))
+
+(re-frame/reg-sub
+ :communities/my-pending-request-to-join
+ :<- [:communities/my-pending-requests-to-join]
+ (fn [requests [_ community-id]]
+   (get requests (name community-id))))
```
ilmotta: Curious, isn't the `community-id` a string already? Why the need to convert it to a string? Could it be a keyword? is that why `name` is used?

src/status_im2/contexts/communities/menus/community_options/view.cljs
ilmotta: I think the answer is yes, the subscription should be higher up, in an actual view component that needs re-rendering capabilities. Even though the code above works, I believe subscriptions should be used for building up hiccup structures/components, not normal Clojure data structures.

src/status_im2/subs/root.cljs
ilmotta: No right or wrong, but I would generally expect community state to live under a root `:communities` key (everything community related actually).

Additionally, avoiding new state at the root of the app db can help us reduce the overhead in the re-frame *signal graph* because it allows us to design the state with more layer-3 subs (which are cached). Our app could probably get a meaningful boost in performance if we started to implement this practice more aggressively.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -108,12 +107,11 @@
...
 (defn join-community
-  [{:keys [joined can-join? requested-to-join-at
+  [{:keys [joined can-join?
            community-color permissions]
-    :as   community}]
-  (let [pending?      (pos? requested-to-join-at)
-        is-open?      (not= constants/community-channel-access-on-request (:access permissions))
-        node-offline? (and can-join? (not joined) (pos? requested-to-join-at))]
+    :as   community} pending?]
+  (let [is-open?      (not= constants/community-channel-access-on-request (:access permissions))
```
cammellos: This is a bit fragile, maybe is better to positively match it? 
```(= constants/community-channel-access-no-membership (:access permissions)```
or 
if there's multiple open communities, something like:
```
(def open-communities #{type-a type-b}) ;; defined somewhere else
(let [is-open? (open-communities (:access permissions)]
...)
```

File: src/status_im2/contexts/communities/menus/community_options/view.cljs (hunk 4 above)
Comment (cammellos) [inline src/status_im2/contexts/communities/menus/community_options/view.cljs]: I'd rename `id` to `community-id`, since it's not immediately clear from the function (I believe that's `id` ->  `community-id` passed as parameter, correct?)

File: src/status_im2/contexts/communities/overview/view.cljs [hunk 14]
```diff
@@ -209,12 +207,11 @@
...
   [{:keys [name description locked joined images
-           status tokens tags requested-to-join-at id]
-    :as   community}
+           status tokens tags id]
+    :as   community} pending?
    {:keys [on-categories-heights-changed
            on-first-channel-height-changed]}]
-  (let [pending?          (pos? requested-to-join-at)
-        thumbnail-image   (get-in images [:thumbnail])
+  (let [thumbnail-image   (get-in images [:thumbnail])
```
cammellos: `(:thumbnail images)`?

src/status_im2/subs/root.cljs
cammellos: In general, I think the overall recommendation is to avoid nesting, and prefer "flat" structures rather than nesting.

The suggestion is to structure it similarly to database tables, so in this case you'd have a root key 
`:communities-requests`, and then how you store the data within the map is contextual to how you query it, since you are building up your own "index" effectively (i.e if you query always by community-id `{:communities-requests {:community-id {:request-id-a {}}}}` might be what you want, if you always want the full list, `{:communities-request {:request-id-a {}}` is probably a better option, and there's more I guess.

This is suggested by Mike Thompson https://groups.google.com/g/clojurescript/c/x5RHF0710CU/m/Th1fizgnlmkJ where he mentions that's how it conceptualizes the app db.

When it comes to performance, it's harder to judge for me, since it's dependent on how the data changes and how you nest things I guess, and as Icaro suggested having some level of nesting, could save performance since you are reducing the level-2 subscriptions, and increasing level-3 subs and therefore leveraging caching, but really hard to judge for me in this case, it'd be super interesting to try things out and see, though I have the feeling that it's very contextual to how the data is updated in the database.

What I can say though for sure, is that getting a nested structure right, is more prone to error, and historically performance issues where due mostly to coding error or not being careful with how we placed things, rather than one approach being more performant than the other.
 
For example we had `{:chats {:chat-id-1 {:messages {:message-id-1 {}}}` in the past, which can look natural, but normally will lead to bad performance since messages are updated frequently, and a new message lead to a chat re-render since we were not so careful when creating a subscription that was pulling the chat object which included messages (we should have subscribed to less state to avoid this, but it's an easy mistake to make which would be avoided if we "flatten" the collection).

So I'd err in this case on the side of caution and keep things flatter, mostly because it's less error prone and easier to reason about, although it could be potentially less performant, but don't have a strong opinion on it

src/status_im2/contexts/communities/menus/community_options/view.cljs
J-Son89: Yep, good point

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: Will adjust

src/status_im2/contexts/communities/overview/view.cljs
J-Son89: That was a lazy refactor by me before. Will do it right this time 😁

src/status_im/communities/core.cljs
```diff
@@ -105,6 +105,16 @@
                db
                communities)})
 
+(rf/defn handle-my-pending-requests-to-join
+  {:events [::fetched-my-communities-requests-to-join]}
```
flexsurfer: by guidelines we shouldn't use namespaced keywords

src/status_im/communities/core.cljs
J-Son89: will remove

src/status_im/multiaccounts/login/core.cljs
J-Son89: @cammellos do we have some form of pub sub for stuff like this?

I would not really consider it ad-hoc. They are just standard lifecycle checks imo. Basically when user logs in and whenever there is a request to the backend to adjust the `users-lists-of-requests`. 
Perhaps I should check if the cancelRequest method comes with a response (ideally the updated list) and then I can remove an unnecessary call to get the list again, as I have it now.
Same for making a request. 

@borismelnik - do you know if there is a response from cancelling/ creating a request to join a community. 

Ideally we could generate some docs for Status Go.

src/status_im2/subs/communities.cljs
J-Son89: yeah, I'll try update the nicer version 👍 . Should be fine afaik

src/status_im2/subs/communities.cljs
J-Son89: it's a string already, I kind of assumed we would want to make it a keyword. But considering it now, I should probably just leave it as a string. Or as we discussed above use oops to handle it early as possible.

src/status_im2/subs/root.cljs
ilmotta: >In general, I think the overall recommendation is to avoid nesting, and prefer "flat" structures rather than nesting.
>
>The suggestion is to structure it similarly to database tables

I don't see it that way. I see the app db for what it is, an open map, without any of the capabilities of a database. In fact, it's much more similar to the wild west of document-oriented databases like Firestore. But any analogy with a database is very frail IMO (unless we use a proper tool, see below).

>This is suggested by Mike Thompson
>https://groups.google.com/g/clojurescript/c/x5RHF0710CU/m/Th1fizgnlmkJ where he
>mentions that's how it conceptualizes the app db.

I respectfully disagree with him, and I think his comment is disconnected from the `status-mobile` reality. I also didn't suggest deep nesting in any way in my previous comment, which his comment focuses a lot. We might have a different definition of what "deep" means in data structures.

I think we should design the app state around the app's read & write patterns, plus the correct data structures. I believe that's what you're hinting when you say "how you store the data within the map is contextual to how you query it". I completely agree with that. We get that right and a lot of good things naturally follow, but I think data structure flatness is tangential to that.

Going back to the app db being a database. Many products went beyond the conceptualization and actually manage the app state with an in-memory database running on the client, with full support for indexing, query language, etc. This might sound crazy at first, but it's really not. I know of some projects that use [DataScript](https://github.com/tonsky/datascript) (and maybe [re-posh](https://github.com/denistakeda/re-posh)), one famous one is http://logseq.com/. Of course, this would require careful experimentation because every tool brings incidental complexity. Performance wise, it could actually improve, but impossible to say without measurements.

>So I'd err in this case on the side of caution and keep things flatter, mostly because it's less error prone and easier to reason about

I would actually caution against using flatness as a meaningful guide to design the app state. The state should mostly serve the purposes of helping efficiently render trees of components. There are always ways to create abstractions to simplify how the data is read/written (I'm again hinting on the idea of a proper client side database).

But good points overall, when in doubt, let's bring the Occam's razor and keep things simple as you suggest Andrea.

src/status_im/multiaccounts/login/core.cljs
borismelnik: @J-Son89 about docs - I totally agreed :)
If I understand you correct, you can find it in `MessageResponse.RequestsToJoinCommunity`. Im a little bit out of current discussion, but I did it in exactly the same way as usual request to join. So when you cancel some request to community, technically you just send it like request to join with special status. So processing the result should be in the same way as it was with requests to join

cammellos: > @J-Son89, I think the reasoning is that we only need to fetch once the app initializes to get the initial & latest state from status-go. Later on, the backend will push signals to the mobile client. And all the mobile client has to do is reconcile the new state with the existing app db state.

I believe that's the correct/desired behavior, so:

1) As the user logs in, we fetch all (or paginated if there's pagination) requests to join (this could be done on component load not to have to store everything in memory from the get go, but it's the same conceptually).
2) Status go will signal new requests through `messages.new`, this are the requests that come from the network (syncing maybe?).  This request should be reconciled with the app-db state.
3) When you create a request, you hit the RPC to create a request in status-go, and status-go will return (not through a signal, but through the callback response) the newly created request, which again will be reconciled

now, that's the theory, I'll have a quick look at the code to see if that's the case.

src/status_im/communities/core.cljs
```diff
@@ -105,6 +105,16 @@
...
+  [{:keys [db]} my-requests]
+  {:db (assoc-in db
+        [:my-pending-requests-to-join]
```
cammellos: no need to assoc-in here

src/status_im/communities/core.cljs
```diff
@@ -105,6 +119,14 @@
                db
                communities)})
 
+(rf/defn handle-my-pending-requests-to-join
+  {:events [:communities/fetched-my-communities-requests-to-join]}
+  [{:keys [db]} my-requests]
+  {:db (assoc db
+              :communities/my-pending-requests-to-join
+              (<-requests-to-join-community-rpc (js->clj my-requests :keywordize-keys true)
```
ilmotta: I would suggest we stick with the functions provided by `status-im.utils.types`. In this case, replace this with `types/js->clj`.

src/status_im/communities/core.cljs
```diff
@@ -36,9 +36,9 @@
                     :chatId      :chat-id}))
 
 (defn <-requests-to-join-community-rpc
```
ilmotta, line 39: This and the other singular version of this function would be better placed in the `data_store/*` namespaces. Something for a future PR if you want :)

Also this `<-requests-to-join-community-rpc` is doing 2 things that can be better described with specialized functions. The function is indexing by the result of `key-fn` and the value in the map is transformed by `<-request-to-join-community-rpc`.

For step 1 we should use `utils.collection/index-by`, and for step 2 we have Clojure's core function `update-vals`. Look at the implementation, it's basically what's written here, **but it performs better** because it uses transients. https://github.com/clojure/clojure/blob/2b3ba822815981e7f76907a3b75e9ecc428f7656/src/clj/clojure/core.clj#L8008

This is one of the reasons I love Clojure for data processing!

```clojure
(let [coll [{:id 1 :name "A"}
            {:id 2 :name "B"}
            {:id 3 :name "C"}]]
  (update-vals (index-by :id coll)
               #(assoc % :age (* (:id %) 10))))
;; => {1 {:id 1, :name "A", :age 10},
;;     2 {:id 2, :name "B", :age 20},
;;     3 {:id 3, :name "C", :age 30}}
```

src/status_im/communities/core.cljs
```diff
@@ -76,14 +76,28 @@
...
+  {:db (assoc-in db [:communities/requests-to-join community-id id] request)})
+
 (rf/defn handle-request-to-join
   [{:keys [db]} r]
-  (let [{:keys [id community-id] :as request} (<-request-to-join-community-rpc r)]
-    {:db (assoc-in db [:communities/requests-to-join community-id id] request)}))
+  (let [my-public-key                                          (get-in db [:multiaccount :public-key])
+        {:keys [id community-id state public-key] :as request} (<-request-to-join-community-rpc r)
+        is-my-request?                                         (= my-public-key public-key)]
+    (if is-my-request?
```
ilmotta: It's probably old news already, but the `is-` prefix is redundant most of the time when there's a question mark. `my-request?` is just a bit more idiomatic.

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -107,15 +106,26 @@
     (i18n/label :t/join-open-community)
     (i18n/label :t/request-to-join-community)))
 
+
```
ilmotta: Extra line

src/status_im2/subs/communities.cljs
```diff
@@ -150,6 +151,19 @@
     (get communities identity)
     counts)))
 
+;; (re-frame/reg-sub
```
ilmotta: This looks like a leftover.

src/status_im2/subs/root.cljs
```diff
@@ -239,6 +239,8 @@
 (reg-root-key-sub :communities/community-id-input :communities/community-id-input)
 (reg-root-key-sub :communities/enabled? :communities/enabled?)
 (reg-root-key-sub :communities/resolve-community-info :communities/resolve-community-info)
+(reg-root-key-sub :communities/my-pending-requests-to-join :communities/my-pending-requests-to-join)
+
```
ilmotta, line 249: Extra line ;)

src/status_im/communities/core.cljs
J-Son89: 🤤

J-Son89:
Thanks for the informative and very helpful feedback @ilmotta :D

src/status_im2/subs/communities.cljs
J-Son89: yep! :)

src/status_im/communities/core.cljs
ilmotta: Quick note: @J-Son89 found that `updated-vals` is not available in our CLJS version https://github.com/status-im/status-mobile/blob/develop/nix/deps/clojure/deps.json#L809, it was added in `1.11`.

Given that upgrading Clojure is generally safe, we should consider moving to `1.11.x` soon. `1.11.0` was released already almost 1 year ago https://clojure.org/news/2022/03/22/clojure-1-11-0.

ilmotta (approved):

VolodLytvynenko:
hi, @J-Son89 thanks for the PR. Here is a found issue that blocks testing the ability to cancel a request. Could you check and fix please failed integration tests?

## ISSUE 1: The community goes to the 'open' status instead of 'pending' after user sends request to join**
**Steps to reproduce:**
1. Create closed community
2. Send request to the community

**Actual result:**
- THe "request to join" button is still  shown after it is tapped
- Community is still in open status after 'request to join' button is tapped

https://user-images.githubusercontent.com/52490791/220086574-b1ec2dca-a165-4534-b50a-730a80a50087.mp4


**Expected result:**
- THe "request to join" button is disappeared after it is tapped
- Community goes to the pending status

VolodLytvynenko:
Hi, @J-Son89 one question related to this PR. Should the notif of 'request to join' disappear in AC for the Admin user if a non-admin user cancels the current request to join the community?

J-Son89:
> Hi, @J-Son89 one question related to this PR. Should the notif of 'request to join' disappear in AC for the Admin user if a non-admin user cancels the current request to join the community?

tbh I'm not sure about activity center. Perhaps @ilmotta or @smohamedjavid can answer this?

smohamedjavid:
> > Hi, @J-Son89 one question related to this PR. Should the notif of 'request to join' disappear in AC for the Admin user if a non-admin user cancels the current request to join the community?
> 
> tbh I'm not sure about activity center. Perhaps @ilmotta or @smohamedjavid can answer this?

I believe @pedro-et or @John-44 is the right person to answer.

P.S. From a UX standpoint, Either the request should be removed from AC for Community Manager or the request should be updated with **invalid** or **revoked** status.

VolodLytvynenko:
> > > Hi, @J-Son89 one question related to this PR. Should the notif of 'request to join' disappear in AC for the Admin user if a non-admin user cancels the current request to join the community?
> > 
> > 
> > tbh I'm not sure about activity center. Perhaps @ilmotta or @smohamedjavid can answer this?
> 
> I believe @pedro-et or @John-44 is the right person to answer.
> 
> P.S. From a UX standpoint, Either the request should be removed from AC for Community Manager or the request should be updated with **invalid** or **revoked** status.

@smohamedjavid Thank you. 
@J-Son89 I will ask someone from the design team and let you know

xAlisher:
> Hi, @J-Son89 one question related to this PR. Should the notif of 'request to join' disappear in AC for the Admin user if a non-admin user cancels the current request to join the community?

It should disappear because there will be nothing to accept or reject

src/quo2/components/tags/status_tags.cljs
```diff
@@ -77,14 +77,18 @@
                         colors/danger-60)}])
 
 (defn- pending
-  [size _ label]
+  [size theme label]
```
J-Son89: needed to revert these changes made to pending tag

J-Son89:
@VladimrLitvinenko, apologies there was some issues with this branch which I have now addressed. Can you take a look again whenever you have time?

J-Son89:
@VladimrLitvinenko, thanks for your patience! I have this branch in a healthy state now, all tests etc passing and ready for QA :)

VolodLytvynenko:
> @VladimrLitvinenko, thanks for your patience! I have this branch in a healthy state now, all tests etc passing and ready for QA :)

hi @J-Son89 . Sorry, a have missed your message. Could you please rebase the PR and resolve existing conflicts one more time? Thanx!

J-Son89:
Sure, I will rebase and ping you once done :)

J-Son89:
@VladimrLitvinenko this branch is rebased :)

VolodLytvynenko:
Hi, @J-Son89 Thank you for your work. Request canceling works as described in [specs](https://github.com/status-im/status-mobile/pull/14973#issue-1568467905)
 Here is the list of found issues:

## ISSUE 1: The community notification is still shown for admin in AC after the community request is canceled

#### Preconditions:
1. User_A is an admin of a closed community
2. User_A and User_B are mutual contacts

#### Steps to reproduce:
1. User_A: Shares community with User_B
2. User_B: Sends "Request to join to the community" (now notification is shown for User_A in AC)
3. User_B: Cancels request
4. User_A: Checks notification in AC

#### Actual result:
The "Request to join" Notification  is still shown for admin user  in AC 

https://user-images.githubusercontent.com/52490791/223774871-fd48aeb1-f9cd-4d8f-9da9-f9d9c2b40a5e.mp4

#### Expected result:
The "Request to join" Notification disappeared for the admin user  in AC after requester cancels the current request

## ISSUE 2: The confirmation drawer and toast are not shown after the 'cancel request' option is tapped

#### Preconditions:
1. User_A is an admin of a closed community
2. User_A and User_B are mutual contacts

#### Steps to reproduce:
1. User_A: Shares community with User_B
2. User_B: Sends "Request to join to the community" (now notification is shown for User_A in AC)
3. User_B: Cancels request

#### Actual result:
- The confirmation drawer  "cancel request?"  is not shown
- The toast 'you canceled request to join' is not shown
https://user-images.githubusercontent.com/52490791/223777274-9e5cc66f-1454-4fad-b29e-89bb9ea0c7ce.mp4


#### Expected result:
https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401-115668&t=zAUQyVNr2QhRIz4u-0
- The confirmation drawer  "cancel request?"  is shown
image

- The toast 'you canceled request to join' is not shown
image

VolodLytvynenko:
@J-Son89 one more question. As I noticed in the PR the following notification is shown:
image
As I understood it should work in the following way:
- If a member user sends a request to a community and the admin is offline, then the current notif should be shown.
Is it implemented in the scope of this PR?

For now, it works so:
### The notification is not shown when the admin is online, but it is shown if the joined user leaves the community and requests it again
#### Preconditions:
- User_A is an admin of a closed community
- User_A and User_B are mutual contacts
#### Steps to reproduce:
1. User_A: Shares community with User_B
2. User_B: Sends "Request to join the community" 
3. User_A: Accepts request
4. User_B leaves the community
5. User_B sends a request to join one more time
6. User_B checks the community overview page

#### Actual result:
- After step 2: User_B does not see the notification on the community overview page even if the admin user is offline
image

- After step 6: User_B sees the notification even if the admin user is online 
image

#### Expected result:
User_B sees the notification on the community overview page after  the admin user goes offline

Is it in the scope of this PR?

J-Son89:
> @J-Son89 one more question. As I noticed in the PR the following notification is shown: image As I understood it should work in the following way:
> 
> * If a member user sends a request to a community and the admin is offline, then the current notif should be shown.
>   Is it implemented in the scope of this PR?
> 
> For now, it works so:
> 
> ### The notification is not shown when the admin is online, but it is shown if the joined user leaves the community and requests it again
> #### Preconditions:
> * User_A is an admin of a closed community
> * User_A and User_B are mutual contacts
> 
> #### Steps to reproduce:
> 1. User_A: Shares community with User_B
> 2. User_B: Sends "Request to join the community"
> 3. User_A: Accepts request
> 4. User_B leaves the community
> 5. User_B sends a request to join one more time
> 6. User_B checks the community overview page
> 
> #### Actual result:
> * After step 2: User_B does not see the notification on the community overview page even if the admin user is offline
>   image
> * After step 6: User_B sees the notification even if the admin user is online
>   image
> 
> #### Expected result:
> User_B sees the notification on the community overview page after the admin user goes offline
> 
> Is it in the scope of this PR?

I don't believe this to be in scope of the PR. @vkjr perhaps you know if this is expected behaviour and/or if there is an issue already created to handle this? 🤔

J-Son89:
> Hi, @J-Son89 Thank you for your work. Request canceling works as described in [specs](https://github.com/status-im/status-mobile/pull/14973#issue-1568467905) Here is the list of found issues:
> 
> ## ISSUE 1: The community notification is still shown for admin in AC after the community request is canceled
> #### Preconditions:
> 1. User_A is an admin of a closed community
> 2. User_A and User_B are mutual contacts
> 
> #### Steps to reproduce:
> 1. User_A: Shares community with User_B
> 2. User_B: Sends "Request to join to the community" (now notification is shown for User_A in AC)
> 3. User_B: Cancels request
> 4. User_A: Checks notification in AC
> 
> #### Actual result:
> The "Request to join" Notification is still shown for admin user in AC
> 
> #### Expected result:
> The "Request to join" Notification disappeared for the admin user in AC after requester cancels the current request
> 
> ## ISSUE 2: The confirmation drawer and toast are not shown after the 'cancel request' option is tapped
> #### Preconditions:
> 1. User_A is an admin of a closed community
> 2. User_A and User_B are mutual contacts
> 
> #### Steps to reproduce:
> 1. User_A: Shares community with User_B
> 2. User_B: Sends "Request to join to the community" (now notification is shown for User_A in AC)
> 3. User_B: Cancels request
> 
> #### Actual result:
> * The confirmation drawer  "cancel request?"  is not shown
> * The toast 'you canceled request to join' is not shown
>   https://user-images.githubusercontent.com/52490791/223777274-9e5cc66f-1454-4fad-b29e-89bb9ea0c7ce.mp4
> 
> #### Expected result:
> https://www.figma.com/file/h9wo4GipgZURbqqr1vShFN/Communities-for-Mobile?node-id=1401-115668&t=zAUQyVNr2QhRIz4u-0
> 
> * The confirmation drawer  "cancel request?"  is shown
>   image
> * The toast 'you canceled request to join' is not shown
>   image

Thanks @VladimrLitvinenko, I will address both of these asap!

J-Son89:
@VladimrLitvinenko I have updated with the extra confirmation bottom sheet and the toast. Can you please check again when you have a chance :)

src/status_im2/contexts/communities/menus/see_rules/view.cljs
```diff
@@ -1,9 +1,12 @@
...
+            [status-im2.contexts.communities.menus.community-rules-list.view :as community-rules]
+            [utils.i18n :as i18n]))
 
 (defn view
   [id]
   [generic-menu/view
-   id
+   {:id    id
+    :title (i18n/label :t/community-rules)
+   }
```
jo-mut, line 11: this too

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
```diff
@@ -23,7 +23,9 @@
            can-join?
            can-request-access?
            requested-to-join-at]}]
-  (let [agreed-to-rules? (reagent/atom false)]
+  (let [agreed-to-rules? (reagent/atom false)
+        pending?         (rf/sub [:communities/my-pending-request-to-join id])
+       ]
```
jo-mut, line 28: something minor the bracket can be at the end of the previous line. I dont think `make lint-fix` will help in this case

jo-mut (approved):

src/status_im2/contexts/communities/menus/request_to_join/view.cljs
J-Son89: Will adjust!

VolodLytvynenko:
hi, @J-Son89 thank you for the fixes. Issue 2 is fixed. 

issue 1: is still reproducible:
## ISSUE 1: The community notification is still shown for admin in AC after the community request is canceled 
 The steps are the same as described initially in https://github.com/status-im/status-mobile/pull/14973#issuecomment-1460630220

## ISSUE 3: The accessibility id is not added to the confirm button
image

VolodLytvynenko:
@J-Son89 Some integration tests failed https://ci.status.im/job/status-mobile/job/prs/job/tests/job/PR-14973/16/consoleText . Could you check it, please?

J-Son89:
> @J-Son89 Some integration tests failed https://ci.status.im/job/status-mobile/job/prs/job/tests/job/PR-14973/16/consoleText . Could you check it, please?

Sure I will take a look at this and the other issue.

J-Son89:
@VladimrLitvinenko I have fixed integration tests issue, it was just linting. 
Also I have added the accessibility label to the button as requested.

For issue 1, handle the activity center. I would prefer to handle this in a separate issue because it is beyond the scope of this pr. Would that be okay? If so I can file an issue for that work.

VolodLytvynenko:
@J-Son89 thank you for your fixes. PR is tested and ready to be merged
Issue 1 is reported as follow up https://github.com/status-im/status-mobile/issues/15328

vkjr:
> I don't believe this to be in scope of the PR. @vkjr perhaps you know if this is expected behaviour and/or if there is an issue already created to handle this? 🤔

@J-Son89, it is not) We can add it as a comment to this issue - https://github.com/status-im/status-mobile/issues/14456

VolodLytvynenko:
hi @J-Son89 Could you please rebase the PR and resolve existing conflicts one more time? Before merging I would like to retest this PR one more time because develop branch is going ahead too fast 
 with new commits for the last 2 days :)

J-Son89:
> hi @J-Son89 Could you please rebase the PR and resolve existing conflicts one more time? Before merging I would like to retest this PR one more time because develop branch is going ahead too fast with new commits for the last 2 days :)

thanks @VladimrLitvinenko that would be great! The corresponding pr for Status go has been merged and this branch is now on the latest and rebased with develop. 

So once the checks pass then it is ready to be checked again

VolodLytvynenko:
@J-Son89 thank you for your work and patience. PR is tested and ready to be merged

J-Son89:
> @J-Son89 thank you for your work and patience. PR is tested and ready to be merged

Thanks for for your help @VladimrLitvinenko! :D

PR #14997, opened 2023-02-06
update react-native-status-keycard

bitgamma:
Fixes #14815

PR #15000, opened 2023-02-06
fix security exception in keycard

### LINKED ISSUE #14815: Keycard NFC tag remove Issue: Java Security Exception
## Problem 
Pairing the Keycard on Android 12 in Status.im (`Device 1`) causes problems.
Since some time Android 12 seems to have problem with the NFC connectivity when pairing and connecting the keycard in Status as well as signing transactions.

#### Expected behavior
When pairing a fresh keycard with the newest applet my Status App in Android 12, the app should go through the pairing process, requesting the pairing password and give access to the wallet.

#### Actual behavior
The following Java Runtime Error occurs when pairing a new card or connecting an existing card:

`AndroidRuntime: java.lang.SecurityException: Permission Denial: Tag ( ID: XX XX XX XX XX XX XX ) is out of date`

So this leads me to the conclusion that something happens during the NFC Rediscover routine that is not suppose to happen (maybe due to an uncatched exception when a Nfc reconnect happens in Android 12 when the card is not placed correctly on the sensor, which happens a lot during discovery).

### Reproduction

1) Open Status
2) Pair an existing keycard or connect an already paired keycard to the status app on Android 13. 
3.) When scanning the NFC Card, the following Java Exception Occurs:

 `AndroidRuntime: java.lang.SecurityException: Permission Denial: Tag ( ID: XX XX XX XX XX XX XX ) is out of date.` 

The app should crash after that exception.

4.) Other errors (APDU failures on `Device 2`) are also observed if connecting the card to an Android 11 after step 1 & 2.

`AndroidRuntime: java.lang.IllegalArgumentException: APDU response must be at least 2 bytes`

The pairing and connecting of the Keycard in Android 10 however still works.

### Additional Information
`Device 1:`
- Status version: 1.20
- Operating System: Android 12 (CalyxOS  4.5.0)
- Device: Fairphone 4 (PF4)

`Device 2:`
- Status version: 1.20
- Operating System: Android 11
- Device: Samsung Galaxy A50

After some investigation this seems to be a known issue at AOSP:

https://issuetracker.google.com/issues/238257723

It seems that this SecurityException was introduced in AOSP commit [113f3a4]
It is thrown when an NFC tag is removed.

The relevant Diff is given here:
https://android.googlesource.com/platform/frameworks/base/+/113f3a4a1d9171db7456aa9ae8b0a2b50545c326%5E%21/

**A history of the relevant code and the original Android 12 upstream issue at is attached below:** - NFC Tag Routines in the Android Nfc Service:

Methods: 
`void setTagUpToDate(long cookie)`
`boolean isTagUpToDate(long cookie)`

The return values from this functions should be checked as well as Catched Exceptions on tag removal.

https://android.googlesource.com/platform/packages/apps/Nfc/+/master/src/com/android/nfc/NfcService.java#2124
https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/nfc/Tag.java#380

An example fix for reference:
https://github.com/open-keychain/open-keychain/pull/2804

bitgamma:
Fixes #14815 (based on 1.20.x branch)

### DIFF
--- a/ci/Jenkinsfile.android
+++ b/ci/Jenkinsfile.android
@@ -4,7 +4,7 @@ library 'status-jenkins-lib@v1.5.5'
 def isPRBuild = utils.isPRBuild()
 
 pipeline {
-  agent { label 'linux && x86_64 && nix-2.8' }
+  agent { label 'linux && x86_64 && nix-2.11' }
 
   options {
     timestamps()
--- a/ci/Jenkinsfile.ios
+++ b/ci/Jenkinsfile.ios
@@ -4,7 +4,7 @@ library 'status-jenkins-lib@v1.5.5'
 def isPRBuild = utils.isPRBuild()
 
 pipeline {
-  agent { label 'macos && x86_64 && nix-2.8 && xcode-13.3' }
+  agent { label 'macos && x86_64 && nix-2.11 && xcode-14.2' }
 
   parameters {
     string(
--- a/ci/Jenkinsfile.tests
+++ b/ci/Jenkinsfile.tests
@@ -4,7 +4,7 @@ library 'status-jenkins-lib@v1.5.5'
 def isPRBuild = utils.isPRBuild()
 
 pipeline {
-  agent { label 'linux && x86_64 && nix-2.8' }
+  agent { label 'linux && x86_64 && nix-2.11' }
 
   options {
     timestamps()
--- a/nix/deps/gradle/deps.json
+++ b/nix/deps/gradle/deps.json
@@ -3990,30 +3990,30 @@
   },
 
   {
-    "path": "com/github/status-im/status-keycard-java/android/3.0.4/android-3.0.4",
+    "path": "com/github/status-im/status-keycard-java/android/3.1.1/android-3.1.1",
     "host": "https://jitpack.io",
     "type": "jar",
     "pom": {
-      "sha1": "e6e1933d85ac05b51953effc21ab893679f5864b",
-      "sha256": "1y9kqp3qgx3kqaf6vm6iwphdj8caqd9x6kncjsnp6f0nirp5ijzq"
+      "sha1": "ac7e6699628e329dfc028345f32bc41c5d6c2f33",
+      "sha256": "12qh5qwncj2gbv6g1whz32rliap34mv2kxlyqv2ivg5jc6yfvyhj"
     },
     "jar": {
-      "sha1": "dca6e47e087a649643bbad1c1feb4b0b5e1aaf45",
-      "sha256": "0krxjj5h5zmap4645llscwkmyszm8qyngwg8vi8ds1mf2kp0jaz1"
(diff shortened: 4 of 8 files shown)


jakubgs (approved):
Does this replace #14997?

Also, nice digits:
image

bitgamma:
> Does this replace #14997?

not really, I wanted CI to build something based off the 1.20.x branch to make testing and cherry-picking easier. Build seems stuck though.

> Also, nice digits: image

😎

jakubgs:
In general please don't create new PRs just to run builds, there are better ways.
But yeah, they are in the queue, give them a bit:

image

jakubgs:
You need to rebase dude:
```
There are no nodes with the label 'linux&&x86_64&&nix-2.8'
```
These jobs will NEVER run if you keep using old CI host labels that no longer exist.

bitgamma:
> You need to rebase dude:
> 
> ```
> There are no nodes with the label 'linux&&x86_64&&nix-2.8'
> ```
> 
> These jobs will NEVER run if you keep using old CI host labels that no longer exist.

this PR is for a hotfix to the [1.20.x](https://github.com/status-im/status-mobile/tree/release/1.20.x) branch (as you can see the target is not develop) so I cannot really rebase it. I think it will be easier to merge this PR than cherry-picking the commit from the #14997 PR (chances are the commit there won't apply cleanly to 1.20.x). @churik do you think this is needed? or is it better to keep only #14997?

jakubgs:
You can just edit the host labels alone to use the current labels:
https://github.com/status-im/status-mobile/blob/d6c09d289cb72adeaacf49983f61ae772410d8bc/ci/Jenkinsfile.ios#L7
https://github.com/status-im/status-mobile/blob/d6c09d289cb72adeaacf49983f61ae772410d8bc/ci/Jenkinsfile.android#L7
https://github.com/status-im/status-mobile/blob/d6c09d289cb72adeaacf49983f61ae772410d8bc/ci/Jenkinsfile.tests#L7

churik:
We can have only one PR (#14997) but in order to make a hotfix we'll still need to cherry-pick it for branch 1.20. 
I could properly test it only on the current PR (if it would be possible to reproduce it for me) or on hotfixed version, as in #14997 there is no option for signing transactions.
Bottom line: best to have this one if it is possible

jakubgs:
Yeah, iOS might have to run on the `x86_64` host instead of `arm64` if this is an older branch. M1 Macs might not work.

bitgamma:
@jakubgs thanks, that did the trick, at least for Android, which is the only build affected by the PR. I tried changing the host to x86_64 for iOS anyway, let's see if it builds

@churik ok, let's keep this PR too then, Android build is generated correctly now

churik:
@bitgamma tested it on Adnroid12, no regression is found (can pair/sign transactions without exceptions), so I believe it is ready to be merged and cherry-picked to 1.20.x branch for hotfix

cc @cammellos

bitgamma:
superseeded by #14997

Issue #13429, opened 2022-05-30
Migrate Wallet Connect to Rc-2


briansztamfater:
WalletConnect Beta phase is over.
Latest Release Candidate guide is present here : https://docs.walletconnect.com/2.0/javascript/guides/react-native

qfrank:
i'll look into this :) @flexsurfer

qfrank:
Hi @flexsurfer , when i update `"@walletconnect/client": "^2.0.0-beta.23"` to `"@walletconnect/sign-client": "^2.0.0-beta.100"` , i couldn't make status startup normally, error log:
```
[Fri Aug 12 2022 18:27:54.915]  ERROR    ReferenceError: Property 'TextEncoder' doesn't exist, js engine: hermes
[Fri Aug 12 2022 18:27:54.915]  ERROR    Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication), js engine: hermes
```
i'm not good at this, so i think i should leave this opportunity to other talents :)

flexsurfer:
@briansztamfater any thoughts?

siddarthkay:
I'll try this out @flexsurfer

siddarthkay:
> [Fri Aug 12 2022 18:27:54.915]  ERROR    ReferenceError: Property 'TextEncoder' doesn't exist, js engine: hermes
> [Fri Aug 12 2022 18:27:54.915]  ERROR    Invariant Violation: Module AppRegistry is not a registered callable module (calling runApplication), js engine: hermes

I was able to recreate this error.

The migration docs here : https://gist.github.com/pedrouid/1a36c6a8776e49453838578ec84715e6 mention that we have to explicitly also install the following dependency : `@react-native-async-storage/async-storage` 
we however have a different version of Async Storage in our `package.json`  : `@react-native-community/async-storage`

It is clear that these 2 dependencies can not co-exist in our codebase, I'll do more research on this to find the best way forward.

siddarthkay:
it seems like `@react-native-community/async-storage` is deprecated and we should get rid of its usage/bindings in our code and replace with `@react-native-async-storage/async-storage` instead.


![Screenshot 2022-08-19 at 2 34 30 PM](https://user-images.githubusercontent.com/64726664/185585318-4d44907f-ef0c-4054-b366-401e08af9fa4.png)

siddarthkay:
[...]

PR #15010, opened 2023-02-07
feat: remove wallet connect 1.0

### LINKED ISSUE #14978: remove WalletConnect v1.0
WalletConnect v1.0 is shutting down on March 1st

briansztamfater:
fixes #14978 

### Summary

Wallet Connect 1.0 is shutting down on March 1st, this PR removes its integration. Also, as the Wallet Connect 2.0 integration was made with the library in beta version, it has got outdated and not working anymore, so disabled session pairing until we re-enable them with #13429 implementation

### Testing

In this case smoke testing scanning wallet connect QR codes and check that don't crash and nothing happens would be fine.

status: ready


translations/en.json
```diff
@@ -1960,5 +1960,6 @@
     "my-albums": "My albums",
     "images": "images",
     "only-6-images": "You can only add 6 images to your message",
-    "delivered": "Delivered"
+    "delivered": "Delivered",
+    "wallet-connect-not-implemented": "Wallet Connect not implemented"
```
flexsurfer: is it like only for us or for the release app as well? not sure it should be translated

flexsurfer (approved):

translations/en.json
briansztamfater: Only for us I think, we need to upgrade and re-enable WC 2.0 for the MVP afaik, so this alert shouldn't see the light. I can remove the translation and use a plain string.

yqrashawn (approved):

siddarthkay (approved):
🥲

flexsurfer:
@briansztamfater could you pls resolve conflicts? and lets merge it, im not sure it need any manual testing

briansztamfater:
@flexsurfer thanks for merging, I did a quick testing locally and seems that nothing was broken with these changes

PR #15036, opened 2023-02-07
14861 improve chat avatar

### LINKED ISSUE #14861: Improve home screen chat avatar
# Problem

Home screen chat avatar needs to be improved

### Improvements 
1. Add identity ring for chat avatar
2. Avatar looks different from actual [designs](https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System?node-id=115%3A6787&t=rh3OC99AuNYNVy6d-0). Fox ex, in design we are only using two characters as initials, but in implementation it is three

ulisesmac:

fixes #14861 

### Summary

#### Problem

The chat avatar in the messages list was not showing the ring around the image profile and it was showing more than 2 initial letters in the name.

#### Solution
The solution consists of the following changes (splitted in different commits to make the review easier):

1 - a63e62245012ea45e3ba48a6c9c3747c3792c3f4 -  Modifies the `user-avatar` API  since the profile pictures already come with the ring drawn over them, and `ring?` was a parameter passed to the `user-avatar` component (please take a loot at the `user-avatar`'s docstring).

2 - 96c23a5f8ad93994245000fd61f4619bd4028f02 - Updates the preview screen, which can be found in `avatar` -> `user-avatar` in the Quo2 Components.

3 - 7f5490c88ee2343d6d1d75dbdc40a775263e6d5b - Adapts existing uses of the component when the ring **was not intented** to be displayed

4 - 3272bce9ddd497d3590a61ff3087ee4f8e8ca4b6 - Adapts existing uses of the component when the ring **was intented to be displayed**

(i.e. 3 & 4 adjust the uses of the component across the system to show or not the ring).

### Review notes

Two problems were noted while solving this issue:

#### 1. Ring width (issue created #15037)
The component withthe ring should looks as follows:
image image

Which consists of a 28x28 user image (or inital letters) and a ring of width 2.

I used the identicon image that comes with the ring:
(this without the pen icon)
image

using a size of (32x32) and putting the inital-letters profile image on top of it (28x28) and it's noticeable that the ring is thinner than expected (look at that white ring between them):

image

This issue is also happening when the user has a profile image set, so we need to make a thicker ring.

  
#### 2. User profile in nav bar

In the chat navigation bar:
image

that "EN" is a use of the avatar component, but it's being called without enough parameters: there's no name, profile image, etc. Is it correct? should we create an issue?

#### Platforms

- Android
- iOS

##### Functional

- 1-1 chats

### Steps to test

- Open Status
- Go to the chats tab, the chat's profile image always display the ring, it doesn't matter if the user has already set a picture or not.
- Open the chat with someone and the avatar at the top of the screen is showing the ring
- Go to `Settings` -> `Quo2.0 Preview`  -> `avatar` -> `user-avatar`.
- When the profile picture is set as "None", and the size is Medium or bigger, the ring can be turned on/off.

**NOTE: The ring is showed always when the profile picture is set because the image comes with the ring from the back-end, in the testing screen it's not showed because those profile images don't have the ring**

status: ready


src/quo2/components/avatars/user_avatar.cljs
```diff
@@ -84,71 +76,70 @@
...
+  (-> [rn/view
+       {:style (merge
+                {:background-color (colors/custom-color-by-theme :turquoise 50 60)
+                 :justify-content  :center
+                 :align-items      :center}
+                (container-styling inner-dimensions outer-dimensions))}]
+      (concat children)
+      (vec)))
+
+(def one-inital-letter-sizes #{:xs :xxs :xxxs})
```
erikseppanen: spelling: `initial`

src/status_im2/contexts/chat/home/chat_list_item/view.cljs
```diff
@@ -102,33 +102,39 @@
     (datetime/to-short-str timestamp)]])
 
 (defn avatar-view
-  [group-chat color display-name photo-path chat-id]
-  (if group-chat
+  [{:keys [contact chat-id full-name color]}]
+  (if contact ;; `contact` is passed when it's not a group chat
```
erikseppanen: just a single semicolon is recommended here

src/quo2/components/avatars/user_avatar.cljs
```diff
@@ -84,71 +76,70 @@
...
-                  (container-styling inner-dimensions outer-dimensions))}
-   children])
+  [{:keys [inner-dimensions outer-dimensions]} & children]
+  (-> [rn/view
+       {:style (merge
+                {:background-color (colors/custom-color-by-theme :turquoise 50 60)
+                 :justify-content  :center
+                 :align-items      :center}
+                (container-styling inner-dimensions outer-dimensions))}]
+      (concat children)
```
erikseppanen: Is this function ever going to receive anything other than the single `[text/text ...` below? If so, I'd look at simplifying this, maybe just including the code in the `initials-avatar` function that seems to be the only thing that uses it.

src/quo2/components/avatars/user_avatar.cljs
```diff
@@ -84,71 +76,70 @@
...
-(def small-sizes #{:xs :xxs :xxxs})
-(def identicon-sizes #{:big :medium :small})
+(defn initials-avatar
+  [{:keys [full-name size inner-dimensions outer-dimensions]}]
+  (let [amount-initials (if (one-inital-letter-sizes size) 1 2)
+        initials        (->> (split full-name " ")
+                             (transduce (comp (map first) (take amount-initials))
+                                        str)
+                             (upper-case))
+        font-size       (get-in sizes [size :font-size])]
```
erikseppanen, line 99: It's probably best to avoid transducers because they're not as commonly understood as map/filter/reduce:
``` clojure
(->> (split full-name " ")
     (map (comp first clojure.string/upper-case))
     (take amount-initials)
     (clojure.string/join))
```

erikseppanen: It's also recommended to use `string/upper-case` and `string/join` here instead of `upper-case` and `join`.

src/quo2/components/avatars/user_avatar.cljs
```diff
@@ -40,40 +39,33 @@
...
+                       :medium 0
+                       :small  -2
+                       0)
+        bottom       (case size
+                       :big    (if ring? -1 2)
+                       :medium (if ring? 0 -2)
+                       :small  -2
+                       0)]
+    [rn/view
+     {:style {:background-color (if online?
```
ilmotta, line 56: According to the guidelines https://github.com/status-im/status-mobile/blob/develop/doc/new-guidelines.md#component-styles, we should try to extract styles into a separate `style` namespace, co-located with the component.

Given this work would considerably change this file (and probably rename it to `components/avatars/user_avatar/view.cljs`), you might prefer to do this in a separate PR to cause less confusion for reviewers.

src/quo2/components/avatars/user_avatar.cljs
ilmotta: I agree with @erikseppanen, but for a different reason.

I think the original implementation is mixing transducers with the normal pipelining macro, but IME this often complicates things. It's similar to code that mixes thread-first and thread-last, sometimes nested. I think it's more idiomatic to pick one style for each kind of data processing.

On the other hand, I'm actually on the camp that think transducers should be used far more often, especially for processing larger collections, since they perform better and produce less intermediate data, thus putting less pressure on the GC. For small collections there's no meaningful difference. I can imagine certain screens and subscriptions could take advantage of this. There are good benchmarks out there proving there can be substantial differences, so maybe, just maybe, in a resource constrained environment such as mobile, the performance improvements could be significant.

src/quo2/components/avatars/user_avatar.cljs
ilmotta: I usually follow the pattern @ulisesmac used here when I want the function to show it doesn't care what/who are the children, if it's just a text component, etc.

Going a bit on a tangent, this new library https://github.com/pitch-io/uix tries to help developers write components in a more uniform way https://github.com/pitch-io/uix/blob/master/docs/components.md. Reagent, on the other hand, is much more open. Different approaches.

src/quo2/components/avatars/user_avatar.cljs
erikseppanen: I agree @ilmotta, for processing large collections. I had a problem exactly like that once, and looked at both transducers and reducers, and just used [reducers](https://clojure.org/reference/reducers) for that. Also agree with your other points!

src/quo2/components/avatars/user_avatar.cljs
ulisesmac: Solved in 97a0742c1a54987eeb6b3f7eac021bd9db250c70

src/status_im2/contexts/chat/home/chat_list_item/view.cljs
ulisesmac: I thought a single semicolon was used for right alined comments, but checked bbastsov style guide and I agree. It's in 609400c9123bee8d041853fa87ea5d3b4a14ca4f

src/quo2/components/avatars/user_avatar.cljs
ulisesmac: @erikseppanen  That component is not used in any other part of the code. In the previous implementation, it was used twice in this namespace. I wanted to keep general existing code structure and not to perform a huge refactor over the file when fixing the current issue, but I agree with you.

@ilmotta  Yes, sometimes it's better than just using children as is, (e.g. in the previous implementation the child text required the `^key` even though it was a single child).

Anyway, in 9b0176641bced3ce44395de303a4fdaad381c63a I removed the `container` component and renamed the `container-styling` function since both were used just once, the code is now smaller and IMO clearer.

src/quo2/components/avatars/user_avatar.cljs
ulisesmac: @erikseppanen @ilmotta 
Well, I used a transducer because it's familiar to me, but  I agree with @ilmotta about mixing thread-last and transduce sometimes makes code more complicated.

In this case it doesn't wort to use transducers, it's removed in af3d6dec6ce6082a6555029ab069f20dcfb7c8fb, I used an `as->>` since I was not confortable with that  `(split full-name " ")`  as parameter of `->>`.

For large collections, it'd be nice to use transducers :)

src/quo2/components/avatars/user_avatar.cljs
ulisesmac: @ilmotta Have you tried that library? if so, what do you think?

I've just checked it, it looks as using hiccup but replacing `[comp props]` by `($ comp props)`.
> tries to help developers write components in a more uniform way

Have you found an example you could share to us?

thanks!

src/quo2/components/avatars/user_avatar.cljs
ulisesmac: Yep, I'll refactor this namespace after this branch is merged, I've just created #15045

src/quo2/components/avatars/user_avatar.cljs
ilmotta: > in the previous implementation the child text required the ^key even though it was a single child

:+1: If we use the pattern `(into [comp-a props] children)` then React won't complain about missing unique keys.

> I've just checked it, it looks as using hiccup but replacing [comp props] by ($ comp props)

I have zero experience with that lib, but I'm interested, especially by the performance claims, but also because of some design choices. I didn't know it used hiccup behind the scenes. Apparently, the performance improvements are in great part achieved by the macro `$`, which can ahead of time compile and optimize a bunch of stuff. Reagent also has hiccup a compiler, but it's slower. Very interesting! Always useful to get outside the Reagent bubble. https://github.com/pitch-io/uix/blob/25cc16a5787d30492e17b7d3e254d351e45fc6e9/core/src/uix/core.clj#L128

>tries to help developers write components in a more uniform way

By uniformity I mean the `defui` macro is telling developers there's one specific way to write a component. With Reagent it's much more open for interpretation. In React, every component that needs arguments will receive them via a map. The ability to use `children` is also implicitly supported. In Reagent every component can be called in any way, and to use children we need to do the dance of wrapping in a `into` or `concat` call. Also in Reagent we have some weird stuff like using `reagent/current-component` and `reagent/props` to extract data from the component that's not apparent from the component/function signature. It all feels like workarounds...

So based on my experience with React, I see the API uniformity of uix as an advantage, as well as other libraries out there. Maybe I'm being too picky ;) 

Thanks for asking. I'd be interested in further discussions around these topics!

src/quo2/components/avatars/user_avatar.cljs
ilmotta: Thanks Ulisses, that'd be great :)

src/quo2/components/avatars/user_avatar.cljs
ilmotta: Just a quick addendum here, but the whole `initials` binding could become a function instead. This would help with RDD too. That implementation is a bit too dense to live in a `let` binding.

ilmotta (approved):

src/status_im2/contexts/chat/home/chat_list_item/view.cljs
ilmotta: Oh it's Lisp inheritance, Clojure is just following the convention o/

erikseppanen (approved):

J-Son89 (approved):

J-Son89:
Consider adding some component tests for this work.

Insere you have a follow issue to refactor this so we could do it as part of that too! :)

ulisesmac:
> Consider adding some component tests for this work.
> 
> Insere you have a follow issue to refactor this so we could do it as part of that too! :)

Got it! I'll add them in the refactor issue

VolodLytvynenko:
hi @ulisesmac. Let me know when PR is ready for tests

ulisesmac:
> hi @ulisesmac. Let me know when PR is ready for tests

@VladimrLitvinenko It is ready!

VolodLytvynenko:
@ulisesmac looks like some integration tests are failed https://ci.status.im/job/status-mobile/job/prs/job/tests/job/PR-15036/6/consoleText. Could you fix them, please?

ulisesmac:
> @ulisesmac looks like some integration tests are failed https://ci.status.im/job/status-mobile/job/prs/job/tests/job/PR-15036/6/consoleText. Could you fix them, please?

Hey @VladimrLitvinenko 
I've just fixed them :)

VolodLytvynenko:
Hi @ulisesmac. My apologies for the testing delays. Worked with more prioritized PR. Could you resolve existing conflicts?

ulisesmac:
> Hi @ulisesmac. My apologies for the testing delays. Worked with more prioritized PR. Could you resolve existing conflicts?

Hey @VladimrLitvinenko I've just solved the conflicts :)

VolodLytvynenko:
hi @ulisesmac . Thank you for PR. It is tested and ready to be merged. No issues from my side

VolodLytvynenko:
> that "EN" is a use of the avatar component, but it's being called without enough parameters: there's no name, profile image, etc. Is it correct? should we create an issue?

@ulisesmac I see only this problem is left still as a question. Should I create a separate issue?

ulisesmac:
> > that "EN" is a use of the avatar component, but it's being called without enough parameters: there's no name, profile image, etc. Is it correct? should we create an issue?
> 
> @ulisesmac I see only this problem is left still as a question. Should I create a separate issue?

I don't know how to exactly solve it. but we're redesigning the app, so I think it will be addressed when the redesign reaches that part.

VolodLytvynenko:
Hi @ulisesmac Sorry, one more question I missed. Should the current ring be shown inside the opened chat and maybe in some other places where it was not shown or in the scope of this PR related only to improving 1-1 chat avatar?
**Example:**
1. For now, the ring is not shown in the opened chat after users sent some messages if the profile image is not set
image

2. Group chats and community chats avatars are without the ring
image

ulisesmac:
> Hi @ulisesmac Sorry, one more question I missed. Should the current ring be shown inside the opened chat and maybe in some other places where it was not shown or in the scope of this PR related only to improving 1-1 chat avatar?
> **Example:**
> 1. For now, the ring is not shown in the opened chat after users sent some messages if the profile image is not set
> image
> 
> 2. Group chats and community chats avatars are without the ring
> image
> 


Hey @VladimrLitvinenko 

The previous chat avatar implementation had a non-working parameter called `ring` which was set to `true` or `false`  in different places across the app.

I changed the implementation to something equivalent that now is able to show the ring, so the ring is only shown in the places where they were previously marked as `true`.

Probably these questions are for the design team 🤔 we could ask them to check if the rings are displayed in the correct places and if not, open an issue, also I could take that issue.

What do you think? Do you know someone on the design team that can help us?

VolodLytvynenko:
> What do you think? Do you know someone on the design team that can help us?

@ulisesmac Got it. Thank you for the explanation.  Sure, I will ask someone from the design team @pedro-et @xAlisher. 
But a better way to ask them is on the 1-1 chat, not in GitHub

PR #15054, opened 2023-02-09
Add back send contact update behavior

cammellos:
Desktop 0.90 expects a contact update when accepting a contact request, this PR restores the behavior,
status:ready

### DIFF
--- a/status-go-version.json
+++ b/status-go-version.json
@@ -3,7 +3,7 @@
     "_comment": "Instead use: scripts/update-status-go.sh <rev>",
     "owner": "status-im",
     "repo": "status-go",
-    "version": "v0.130.0",
-    "commit-sha1": "dea21f440a9f630c4b8324fcde0b016d6c3fc02b",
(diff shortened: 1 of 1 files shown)


flexsurfer (approved):

churik:
@cammellos 
compatibility mobile <-> desktop works as expected, but there is an issue with mobile <-> mobile

### ISSUE 1: Accept of CR generates one more CR

User A and User B are using mobile apps (PR builds)

**Steps:**
1) A: send CR
2) B: accept CR

**Expected result:**
A: user B is added to contacts, no pending CR anymore 


**Actual result:**
A: user B is added to contacts, and pending CR is shown 
IMAGE 2023-02-09 18:17:07

**OS:** IOS, Android

cammellos:
@churik yes, unfortunately that's a bit expected, since we have to send 2 for compatibility.
I could address, but looks like a non-blocker (most people will be desktop 0.9), but in the meantime, I'd merge the fix if you are ok, and then we can revisit as we get closer. what do you think?

churik:
yes, good to go then, created follow up issue

cammellos:
merged in develop

PR #15062, opened 2023-02-10
feat: add drawer buttons to quo2

### LINKED ISSUE #15013: Implement Quo2 Component - Drawers/ Drawer Buttons
Implement as per designs:

https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System-for-Mobile?node-id=14759%3A209025&t=29WaMjPTYMtHqa1n-0

Include component tests
Include preview screen

J-Son89:
fixes: https://github.com/status-im/status-mobile/issues/15013

to test it out go to quo2 preview and go to drawers -> drawer buttons

Screenshot 2023-02-10 at 11 33 25

### DIFF
--- a/src/quo2/components/drawers/drawer_buttons/style.cljs
+++ b/src/quo2/components/drawers/drawer_buttons/style.cljs
@@ -1,0 +1,24 @@
+(ns quo2.components.drawers.drawer-buttons.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def outer-container {:height 216})
+
+(def top-card
+  {:height             "100%"
+   :display            :flex
+   :padding-vertical   12
+   :padding-horizontal 20
+   :border-radius      20
+   :background-color   colors/neutral-80})
+
+(def bottom-card
+  {:position           :absolute
+   :top                80
+   :left               0
+   :right              0
+   :bottom             0
+   :display            :flex
+   :padding-vertical   12
+   :padding-horizontal 20
+   :border-radius      20
+   :background-color   (colors/alpha colors/white 0.05)})
--- a/src/quo2/components/drawers/drawer_buttons/style.cljs
+++ b/src/quo2/components/drawers/drawer_buttons/style.cljs
@@ -4,8 +4,7 @@
 (def outer-container {:height 216})
 
 (def top-card
-  {:height             "100%"
-   :display            :flex
+  {:flex               1
    :padding-vertical   12
    :padding-horizontal 20
    :border-radius      20
@@ -17,8 +16,33 @@
    :left               0
    :right              0
    :bottom             0
-   :display            :flex
    :padding-vertical   12
    :padding-horizontal 20
    :border-radius      20
    :background-color   (colors/alpha colors/white 0.05)})
+
+(def bottom-container
+  {:flex-direction  :row
+   :justify-content :space-between})
+
+(def bottom-icon
+  {:border-radius   40
+   :border-width    1
+   :margin-left     24
+   :height          28
+   :width           28
+   :justify-content :center
+   :align-items     :center
+   :border-color    (colors/alpha colors/white 0.05)})
+
+(def bottom-text
+  {:flex  1
+   :color (colors/alpha colors/white 0.7)})
+
+(def top-text
+  {:color (colors/alpha colors/white 0.7)})
+
+(defn heading-text
+  [gap]
+  {:color         colors/white
+   :margin-bottom gap})
--- a/src/quo2/components/drawers/drawer_buttons/view.cljs
+++ b/src/quo2/components/drawers/drawer_buttons/view.cljs
@@ -1,0 +1,91 @@
+(ns quo2.components.drawers.drawer-buttons.view
+  (:require [react-native.core :as rn]
+            [quo2.components.icon :as icon]
+            [quo2.components.markdown.text :as text]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.drawers.drawer-buttons.style :as style]))
+
+(defn render-bottom
+  [children]
+  [rn/view
+   {:flex-direction  :row
+    :justify-content :space-between}
+   [children]
+   [rn/view
+    {:style {:border-radius   40
+             :border-width    1
+             :margin-left     24
+             :height          28
+             :width           28
+             :justify-content :center
+             :align-items     :center
+             :border-color    (colors/alpha colors/white 0.05)}}
+    [icon/icon :arrow-right
+     {:color colors/white
+
+      :size  20}]]])
+
+(defn render-children-bottom
+  [children]
+  (if (or (string? children) (keyword children))
+    [render-bottom
+     (fn []
+       [text/text
+        {:size   :paragraph-2
+         :style  {:flex  1
+                  :color (colors/alpha colors/white 0.7)}
+         :weight :semi-bold}
+        children])]
+    [render-bottom children]))
+
+(defn render-children-top
+  [children]
+  (if (or (string? children) (keyword children))
+    [text/text
+     {:size   :paragraph-2
+      :style  {:color (colors/alpha colors/white 0.7)}
+      :weight :semi-bold}
+     children]
+    [children]))
+
+(defn card
+  [{:keys [on-press style heading children gap is-top?]}]
+  [rn/touchable-highlight
+   {:on-press       on-press
+    :border-radius  20
+    :style          style
+    :underlay-color (:background-color style)}
+   [rn/view {}
+    [text/text
+     {:size   :heading-1
+      :style  {:color         colors/white
+               :margin-bottom gap}
+
+      :weight :semi-bold}
+     heading]
+    (if is-top?
+      [render-children-top children]
+      [render-children-bottom children])]])
+
+(defn view
+  "[view opts]
+   opts
+   {:container-style  style-object
+    :top-card         { :on-press event
+                        :heading  string 
+                        :children string or render-fn}
+    :bottom-card      { :on-press event
+                        :heading  string 
+                        :children string or render-fn}}"
+  [{:keys [container-style top-card bottom-card]}]
+  [rn/view
+   {:style (merge container-style style/outer-container)}
+   [card
+    (merge {:gap     4
+            :is-top? true
+            :style   style/top-card}
+           top-card)]
+   [card
+    (merge {:style style/bottom-card
+            :gap   20}
+           bottom-card)]])
--- a/src/quo2/components/drawers/drawer_buttons/view.cljs
+++ b/src/quo2/components/drawers/drawer_buttons/view.cljs
@@ -7,49 +7,39 @@
 
 (defn render-bottom
   [children]
-  [rn/view
-   {:flex-direction  :row
-    :justify-content :space-between}
-   [children]
+  [rn/view {:style style/bottom-container}
+   children
    [rn/view
-    {:style {:border-radius   40
-             :border-width    1
-             :margin-left     24
-             :height          28
-             :width           28
-             :justify-content :center
-             :align-items     :center
-             :border-color    (colors/alpha colors/white 0.05)}}
+    {:style style/bottom-icon}
     [icon/icon :arrow-right
      {:color colors/white
-
       :size  20}]]])
+
+(defn label? [el] (or (string? el) (keyword? el)))
 
 (defn render-children-bottom
   [children]
-  (if (or (string? children) (keyword children))
+  (if (label? children)
     [render-bottom
-     (fn []
-       [text/text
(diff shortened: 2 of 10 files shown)


src/quo2/components/drawers/drawer_buttons/style.cljs
```diff
@@ -0,0 +1,24 @@
+(ns quo2.components.drawers.drawer-buttons.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def outer-container {:height 216})
+
+(def top-card
+  {:height             "100%"
+   :display            :flex
```
flexsurfer: no need to specify "100%" and :flex

src/quo2/components/drawers/drawer_buttons/style.cljs
```diff
@@ -0,0 +1,24 @@
...
+   :border-radius      20
+   :background-color   colors/neutral-80})
+
+(def bottom-card
+  {:position           :absolute
+   :top                80
+   :left               0
+   :right              0
+   :bottom             0
+   :display            :flex
```
flexsurfer: redundant :flex

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+             :align-items     :center
+             :border-color    (colors/alpha colors/white 0.05)}}
+    [icon/icon :arrow-right
+     {:color colors/white
+
+      :size  20}]]])
+
+(defn render-children-bottom
+  [children]
+  (if (or (string? children) (keyword children))
```
J-Son89: render subtitle/children can be both a string or a render prop as the text examples used require inline styling on certain words in the text.

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+            [quo2.components.drawers.drawer-buttons.style :as style]))
+
+(defn render-bottom
+  [children]
+  [rn/view
+   {:flex-direction  :row
+    :justify-content :space-between}
+   [children]
+   [rn/view
+    {:style {:border-radius   40
```
flexsurfer: move to style ns

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+    [children]))
+
+(defn card
+  [{:keys [on-press style heading children gap is-top?]}]
+  [rn/touchable-highlight
+   {:on-press       on-press
+    :border-radius  20
+    :style          style
+    :underlay-color (:background-color style)}
+   [rn/view {}
```
briansztamfater, line 48: why the empty map?

J-Son89: Just a habit i guess, can remove

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+  (if (or (string? children) (keyword children))
+    [text/text
+     {:size   :paragraph-2
+      :style  {:color (colors/alpha colors/white 0.7)}
+      :weight :semi-bold}
+     children]
+    [children]))
+
+(defn card
+  [{:keys [on-press style heading children gap is-top?]}]
```
ulisesmac: Probably it's a personal preference, but I'd say that the boolean vars don't need the `is` prefix (such as Java conventions) because in Clojure it's enough if you add the `?` as suffix.

So I'd rename `is-top?` to `top?`.
Would be good to have feedback about this from someone else.

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+        {:size   :paragraph-2
+         :style  {:flex  1
+                  :color (colors/alpha colors/white 0.7)}
+         :weight :semi-bold}
+        children])]
+    [render-bottom children]))
+
+(defn render-children-top
+  [children]
+  (if (or (string? children) (keyword children))
```
ulisesmac: That `(keyword children)`, will try to convert `children` to a keyword if it's not a string, and that could arise an exception, return a logical true and in some cases return `nil`.

Probably, you wanted to use `keyword?` (with the question mark at the end)

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+             :border-width    1
+             :margin-left     24
+             :height          28
+             :width           28
+             :justify-content :center
+             :align-items     :center
+             :border-color    (colors/alpha colors/white 0.05)}}
+    [icon/icon :arrow-right
+     {:color colors/white
+
```
ulisesmac: Extra empty line

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+   {:on-press       on-press
+    :border-radius  20
+    :style          style
+    :underlay-color (:background-color style)}
+   [rn/view {}
+    [text/text
+     {:size   :heading-1
+      :style  {:color         colors/white
+               :margin-bottom gap}
+
```
ulisesmac: Extra empty line again

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+       [text/text
+        {:size   :paragraph-2
+         :style  {:flex  1
+                  :color (colors/alpha colors/white 0.7)}
+         :weight :semi-bold}
+        children])]
+    [render-bottom children]))
+
+(defn render-children-top
+  [children]
```
ulisesmac, line 32: I'd say it's more comfortable to pass children as a hiccup in ClojureScript instead of passing a render-fn. If children were passed as hiccup, you won't need to wrap children in `[ ]`, and it'd be more natural to pass children to this component.

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+(defn view
+  "[view opts]
+   opts
+   {:container-style  style-object
+    :top-card         { :on-press event
+                        :heading  string 
+                        :children string or render-fn}
+    :bottom-card      { :on-press event
+                        :heading  string 
+                        :children string or render-fn}}"
```
ulisesmac: Following my previous comment about the children prop, two comments:

1. I'd be more natural to pass children as the rest of the params, to be clear, currently this component will be called as follows:

```clojure
     [component props-map-with-children]
```

  and, IMO, it'd be more consistent with existing react components if it were called:
```clojure
     [component props-map 
       child-1
       child-2 ,,,]
```

Additionally,
2. If children were hiccups instead of render functions and combined with the first comment, you could use this component as follows:

```clojure
     [component props-map
       [child-1 props-1]
       [child-2 props-2
         [child-3 props-3]]]
```

Giving us a more familiar/comfortable API than passing render-fns.

What do you think?

src/status_im2/contexts/quo_preview/drawers/drawer_buttons.cljs
```diff
@@ -0,0 +1,75 @@
...
+  []
+  (let [state (reagent/atom {:top-heading     "Sign in "
+                             :top-sub-heading "You already use Status"
+                             :bottom-heading  "I'm new to status"})]
+    (fn []
+      [rn/touchable-without-feedback {:on-press rn/dismiss-keyboard!}
+       [rn/view {:padding-bottom 400}
+        [preview/customizer state descriptor]
+        [rn/view {:padding-vertical 60}
+         (render-drawer-buttons state)]]])))
```
ulisesmac: Why is this component being called using `()` insted of `[]`?

src/quo2/components/drawers/drawer_buttons/view.cljs
briansztamfater: Agree, I think there was a discussion of this topic before and it should be reflected in our guidelines, but didn't find it.

ilmotta: It's not there @briansztamfater, although I find this a strong default for Clojure, I know it's not unanimous and it's not in the Clojure Style Guide (only predicates are mentioned https://guide.clojure.style/#naming-predicates).

So it's worth adding to the guidelines.

src/quo2/components/drawers/drawer_buttons/view.cljs
ilmotta: Good point, consistency is key.

src/quo2/components/drawers/drawer_buttons/view.cljs
```diff
@@ -0,0 +1,91 @@
...
+  [{:keys [container-style top-card bottom-card]}]
+  [rn/view
+   {:style (merge container-style style/outer-container)}
+   [card
+    (merge {:gap     4
+            :is-top? true
```
ilmotta: To go with the flow, just `:top?` ;)

src/test_helpers/component.cljs
```diff
@@ -34,6 +36,8 @@
   [label]
   (rtl/screen.getByLabelText (name label)))
 
+
```
ilmotta: Extra line

src/quo2/components/drawers/drawer_buttons/view.cljs
ilmotta: I would appreciate if other people helped update the guidelines too, right now is a bit too concentrated on me. For the ones that already reached consensus and are quick (like this case of the question mark), they can even be done in the same PR if the contributor wants, whatever is easier to keep the docs up-to-date.

src/status_im2/contexts/quo_preview/drawers/drawer_buttons.cljs
J-Son89: it's a mistake 🙃

src/quo2/components/drawers/drawer_buttons/view.cljs
J-Son89: yep, will do! :)

src/quo2/components/drawers/drawer_buttons/view.cljs
J-Son89: I can update as part of this pr 👍

src/quo2/components/drawers/drawer_buttons/style.cljs
J-Son89: actually it's needed here, the card doesn't display otherwise 🤔

ulisesmac:
@J-Son89 It looks good enough to me! :)

ulisesmac (approved):
@J-Son89 It looks good enough to me! :)

src/quo2/components/drawers/drawer_buttons/style.cljs
```diff
@@ -0,0 +1,49 @@
+(ns quo2.components.drawers.drawer-buttons.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def outer-container {:height 216})
+
+(def top-card
+  {:display            :flex
+   :height             "100%"
```
flexsurfer: redundant :height "100%"

J-Son89: I commented about this above. So without this the card disappears, so it's not redundant. Although perhaps I should use a different approach? 🤔

flexsurfer: we should never specify display :flex and height "100%", we should just use `:flex 1` instead

ilmotta: I've fixed some of these problems by using other flex properties, like `flex` and `flex-grow`. It shouldn't be needed as @flexsurfer said.

J-Son89: cool, will adjust 👍

src/quo2/components/drawers/drawer_buttons/style.cljs
```diff
@@ -0,0 +1,49 @@
+(ns quo2.components.drawers.drawer-buttons.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def outer-container {:height 216})
+
+(def top-card
+  {:display            :flex
```
ilmotta: Redundant :)

ilmotta (approved):

src/quo2/components/drawers/drawer_buttons/style.cljs
J-Son89: 💥

PR #15063, opened 2023-02-10
Revert "communities home header flickering when user scrolls"

jo-mut:
Reverts status-im/status-mobile#15060

I reverted this pr because I realized after having merged it, there was a request-manual-qa label.


flexsurfer:
im not sure it makes sense to revert it if its already merged

PR #14442, opened 2022-11-25
[14429] Fix mentions

ibrkhalil:
fixes #14429
status: ready

PR #15069, opened 2023-02-11
Prevent #links to work except in community chats and allow redirection from channel to another when clicked

### LINKED ISSUE #15048: Handle #links in chats
#links in chat if clicked will open a public chat.
This is not necessary anymore as we won't support them in the future.

Clicking on #links should be disabled on 1-to-1 chats and private group chats. They should not appear highlighted if possible.

Clicking on #links on a community should take you to the channel with the same name, if a channel with that name does not exist, nothing should happen.

There's multiple ways to implement this, please ping me before implementing it.

ibrkhalil:
fixes: #15048
Allow users to press #links to go from one channel to another in community chats and disable #links in non-community chats and do nothing when it's current chat
status: ready


src/status_im/ui2/screens/chat/messages/message.cljs
```diff
@@ -40,7 +40,8 @@
   (rf/sub [:contacts/contact-name-by-identity from]))
 
 (defn render-inline
-  [_message-text content-type acc {:keys [type literal destination]}]
+  [_message-text content-type acc {:keys [type literal destination]} use-status-tag?
+   community-channel-names-and-ids community-id current-chat-id]
```
cammellos: I think we should avoid passing so much state to a message for now, as this might cause re-rendering.

a bit less correct, but simpler for now, is to pass "community-id" only, and just on-press fire that one with the literal, and do the work on the event handler instead.

so in this case:

```
[_message-text content-type acc {:keys [type literal destination]} use-status-tag? community-id current-chat-id]
   (when use-status-tag?
               {:style    {:color                :blue
                           :text-decoration-line :underline}
                :on-press #(rf/dispatch [:communities/status-tag-pressed community-id literal]))})

```

And in the event handler, you do the lookup, and navigate to chat if a chat is found.

This is a bit less nice in terms of UX, since  any link will show as blue, even if there's not chat with that name,  but it's going to be more performant & simpler, I think that's an acceptable compromise.

ilmotta (approved):
@ibrkhalil This namespace really deserves a good refactor. Not now for this PR ;) The existing code is complicating things a bit too much, like the usage of nested `reduce` calls, the `acc` bindings being shadowed and making things more confusing, the long `case` switches. There is unnecessary interop (`.-length literal`), code out of guidelines, among other stuff.

It would be good if, in the near future, we first refactor this namespace and only then make changes to it, otherwise we just make things worse. I'm inclined to accept almost anything in this namespace, because lots of windows are broken already https://en.wikipedia.org/wiki/Broken_windows_theory

ibrkhalil:
> @ibrkhalil This namespace really deserves a good refactor. Not now for this PR ;) The existing code is complicating things a bit too much, like the usage of nested `reduce` calls, the `acc` bindings being shadowed and making things more confusing, the long `case` switches. There is unnecessary interop (`.-length literal`), code out of guidelines, among other stuff.
> 
> It would be good if, in the near future, we first refactor this namespace and only then make changes to it, otherwise we just make things worse. I'm inclined to accept almost anything in this namespace, because lots of windows are broken already https://en.wikipedia.org/wiki/Broken_windows_theory

Will create a follow up and refactor it Icaro, Thanks for pointing it out!

src/status_im/ui2/screens/chat/messages/message.cljs
ibrkhalil: Okay Andrea, I agree performance is something we should focus on when we have tons of messages in a single view .. Will do as you said

ibrkhalil: If we sufficed with the community id and the literal we'd still have to search in community channels for the desired one .. But that only does the part with the time complexity when clicked and not always .. Understood

ibrkhalil: Done, Please take a look when you have time @cammellos

src/status_im/ui2/screens/chat/messages/message.cljs
```diff
@@ -123,11 +133,18 @@
...
 
 (defn render-parsed-text
-  [{:keys [content] :as message-data}]
-  (reduce (fn [acc e]
-            (render-block message-data acc e))
-          [:<>]
-          (:parsed-text content)))
+  [{:keys [content chat-id]
+    :as   message-data}]
+  (let [{:keys [chat-type community-id]} (rf/sub [:chats/chat chat-id])
```
cammellos: This is subscribing to the whole chat object, which means that a rerender will be triggered if any of the fields in `chat` changes.

Better to subscribe only to the state you need, you can have a subscription:

`(let [use-status-tag? (rf/sub [:chats/is-community-chat? chat-id])]`
which re-uses `chats/chat`.

This should never realistically trigger a re-render since the type of a chat never changes.

src/status_im/communities/core.cljs
```diff
@@ -198,6 +198,17 @@
                                                                %)
                                                     (re-frame/dispatch [::failed-to-leave %]))}]}))
 
+(rf/defn status-tag-pressed
+  {:events [:communities/status-tag-pressed]}
+  [{:keys [db]} community-id literal]
+  (let [current-chat-id (:current-chat-id db)
+        {:keys [id]}    (some #(when (= (:name %) literal) %)
+                              (vals (get-in db [:communities community-id :chats])))]
+    (cond-> {}
```
cammellos: is it not the same as:

```
(when (not=.....)
    {:dispatch [:chat/navigate-to-chat (str...)})
```

cammellos (approved):

Parveshdhull:
> If you think PR needs and is ready for manual testing, please add the request-manual-qa label.
> When manual testing of PR is fully completed and all issues are fixed, QA adds the Tested - OK label and drags the card to the MERGE column, after which developer merges PR into develop. 

Hi @ibrkhalil, Please follow the [pipeline guideline](https://github.com/status-im/status-mobile/blob/develop/doc/pipeline_process.md)s for PRs

https://github.com/status-im/status-mobile/pull/14442#issuecomment-1327523948
https://github.com/status-im/status-mobile/pull/14682
https://github.com/status-im/status-mobile/pull/15069
https://github.com/status-im/status-mobile/pull/14667

(all above PRs have request-manual-qa label, and all merged before testing)

ibrkhalil:
Forgot again :(
Gonna make a revert PR

PR #14783, opened 2023-01-13
Add blur overlay for bottom tabs and shell top nav

Parveshdhull:
### Summary
Add blur overlay for bottom tabs and shell top nav

**Note:** This is just initial implementation of bottom tabs blur, needs further improvement. Especially for ios.

### Testing
PR doesn't add new flow, so there isn't much to test.
Please just test for any visible regressions in the home stack, shell, bottom tabs animations, etc.

#### Known Issue:
IOS blur overlay is [blue](https://github.com/Kureev/react-native-blur/pull/513#issuecomment-1399026306)

status: ready

PR #15071, opened 2023-02-13
ci: add component test

### LINKED ISSUE #14417: Add CI/CD job for component tests
# Add CI/CD job for component tests

Following on from https://github.com/status-im/status-mobile/issues/14330

There is a need for component tests to be run in the CI/CD job on developer builds

Description:
- Add this job to develop p.rs
- Add this job to master branch (??)
- Add documentation on how this is setup/working.

yqrashawn:
fixes https://github.com/status-im/status-mobile/issues/14417

status: ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -349,8 +349,8 @@ component-test-watch: ##@ Watch tests and re-run no changes to cljs files
 component-test: export TARGET := clojure
 component-test: export COMPONENT_TEST := true
 component-test: export BABEL_ENV := test
-component-test: ##@test Run tests once in NodeJS
-	# Here we create the gyp bindings for nodejs
+component-test: ##@test Run component tests once in NodeJS
+	yarn install
 	yarn shadow-cljs compile component-test && \
 	jest --config=test/jest/jest.config.js
 
--- a/ci/Jenkinsfile.tests
+++ b/ci/Jenkinsfile.tests
@@ -68,6 +68,14 @@ pipeline {
         }
       }
     }
+    stage('Component Tests') {
+      steps {
+        sh """#!/bin/bash
+          set -eo pipefail
+          make component-test 2>&1 | tee -a ${LOG_FILE}
+        """
+      }
+    }
     stage('Upload') {
       steps {
         script {
--- a/doc/testing.md
+++ b/doc/testing.md
@@ -18,8 +18,6 @@ Also test watcher can be launched. It will re-run the entire test suite when any
 
 Developers can also manually change the shadow-cljs option `:ns-regex` to control which namespaces the test runner should pick. 
 
-
-
 ## Testing with REPL
 
 The most convenient way to develop and run test locally is using REPL:
@@ -52,3 +50,19 @@ In the video below, you can see two buffers side-by-side. On the left the source
 Here I'm showing a terminal-only experience using Tmux (left pane Emacs, right pane the output coming from running the make target).
 
 [2022-12-19 13-17.webm](https://user-images.githubusercontent.com/46027/208471199-1909c446-c82d-42a0-9350-0c15ca562713.webm)
+
+## Component tests
+
+To run tests:
+
+```
+   make component-test
+```
+
+Also test watcher can be launched. It will re-run the entire test suite when any file is modified
+
+```
+   make component-test-watch
+```
+
+Check [component tests doc](./component-tests-overview.md) for more.
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@babel/preset-typescript": "^7.17.12",
     "@react-native-async-storage/async-storage": "^1.17.9",
     "@react-native-community/audio-toolkit": "git+https://github.com/tbenr/react-native-audio-toolkit.git#refs/tags/v2.0.3-status-v6",
-    "@react-native-community/blur": "git+https://github.com/status-im/react-native-blur#refs/tags/v4.3.0-status",
+    "@react-native-community/blur": "git+https://github.com/status-im/react-native-blur#refs/tags/v4.3.1-status",
     "@react-native-community/cameraroll": "git+https://github.com/status-im/react-native-cameraroll.git#refs/tags/v4.0.4-status.0",
     "@react-native-community/clipboard": "^1.2.2",
     "@react-native-community/hooks": "^2.5.1",
--- a/test/jest/jest.config.js
+++ b/test/jest/jest.config.js
@@ -1,26 +1,23 @@
 module.exports = {
-    "preset": "react-native",
-    "setupFilesAfterEnv": ["@testing-library/jest-native/extend-expect"
-        , "../test/jest/jestSetup.js"
+    preset: 'react-native',
+    setupFilesAfterEnv: [
+        '@testing-library/jest-native/extend-expect',
+        '../test/jest/jestSetup.js',
     ],
-    "setupFiles": [
-    ],
-    "testPathIgnorePatterns": [
-    ],
-    "moduleNameMapper": {
-        '^[@./a-zA-Z0-9$_-]+\\.(png|jpg|jpeg|gif)$': '<rootDir>/../node_modules/react-native/Libraries/Image/RelativeImageStub'
-      },
-    "testTimeout": 60000,
-    "transformIgnorePatterns": [
-        "/node_modules/(?!(@react-native|react-native-haptic-feedback|react-native-redash|react-native-image-crop-picker|@react-native-community|react-native-linear-gradient|react-native-background-timer|react-native|rn-emoji-keyboard|react-native-languages|react-native-shake|react-native-reanimated|react-native-redash|react-native-permissions|@react-native-community/blur)/).*/"
-    ],
-    "globals": {
-        "__TEST__": true
+    setupFiles: [],
+    testPathIgnorePatterns: [],
+    moduleNameMapper: {
+        '^[@./a-zA-Z0-9$_-]+\\.(png|jpg|jpeg|gif)$':
+            '<rootDir>/../node_modules/react-native/Libraries/Image/RelativeImageStub',
     },
-    "testEnvironment": "node",
-    rootDir: "../../component-spec",
-    testMatch: [
-        "**/*__tests__*",
-         "**/*.component_spec.js"
-    ]
-}
+    testTimeout: 60000,
+    transformIgnorePatterns: [
+        '/node_modules/(?!(@react-native|react-native-haptic-feedback|react-native-redash|react-native-image-crop-picker|@react-native-community|react-native-linear-gradient|react-native-background-timer|react-native|rn-emoji-keyboard|react-native-languages|react-native-shake|react-native-reanimated|react-native-redash|react-native-permissions|@react-native-community/blur)/).*/',
+    ],
+    globals: {
+        __TEST__: true,
+    },
+    testEnvironment: 'node',
+    rootDir: '../../component-spec',
+    testMatch: ['**/*__tests__*', '**/*.component_spec.js'],
+};
--- a/test/jest/jestSetup.js
+++ b/test/jest/jestSetup.js
@@ -1,5 +1,5 @@
 const WebSocket = require('ws');
-const { NativeModules } = require('react-native');
+const {NativeModules} = require('react-native');
 
 require('@react-native-async-storage/async-storage/jest/async-storage-mock');
 require('react-native-gesture-handler/jestSetup');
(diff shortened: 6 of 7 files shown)


Makefile
```diff
@@ -351,6 +351,7 @@ component-test: export COMPONENT_TEST := true
 component-test: export BABEL_ENV := test
 component-test: ##@test Run tests once in NodeJS
 	# Here we create the gyp bindings for nodejs
+	yarn install && \
```
yqrashawn: @J-Son89 have to add this back
we can always skip `yarn install` manually

yqrashawn:
to draft, since ci failed

yqrashawn:
Seems caused by two `yarn install` at the same time
try run sequentially

yqrashawn:
works fine running one by one

seems this is the error causing the crash
https://ci.status.im/blue/organizations/jenkins/status-mobile%2Fprs%2Ftests/detail/PR-15071/3/pipeline/#step-39-log-155
`../modules/react-native-status/nodejs/status.cpp:1936:1: fatal error: opening dependency file ./Release/.deps/Release/obj.target/status_nodejs_addon/modules/react-native-status/nodejs/status.o.d.raw: No such file or directory`

is it possible to tell react-native-status to wait for another build?

---

also need to checkout if we need it in tests

---

~or we just run them one by one, each takes little less than 2m~

---

another error, related to `react-native-status` as well
```
[2023-02-13T08:23:09.598Z] rm: cannot remove './Release/.deps/Release/obj.target/status_nodejs_addon/modules/react-native-status/nodejs/status.o.d.raw': No such file or directory
[2023-02-13T08:23:09.598Z] make[1]: *** [status_nodejs_addon.target.mk:107: Release/obj.target/status_nodejs_addon/modules/react-native-status/nodejs/status.o] Error 1
```

Makefile
J-Son89: yeah, that's fine. I just removed while not in CI. but knew this day would come. :)

and yes we can skip manually 👍

ci/Jenkinsfile.tests
```diff
@@ -64,6 +64,10 @@ pipeline {
               set -eo pipefail
               make test 2>&1 | tee -a ${LOG_FILE}
             """
+            sh """#!/bin/bash
+              set -eo pipefail
+              make component-test 2>&1 | tee -a ${LOG_FILE}
```
jakubgs: This should probably be it's own stage. Can they run in parallel?

J-Son89: yeah, imo this would be better to have as it's own job. The two tests types are completely separate to each other so running in parallel should be no issue.

jakubgs: Separate job is overkill, I mean just using parallel functionality in a single Jenkinsfile.

J-Son89: Sounds good 👍

yqrashawn:
ready for review
use `CHILD_CONCURRENCY=1` to avoid node-gyp build error when running multiple `yarn install` in parallel

yarn doc https://classic.yarnpkg.com/en/docs/envvars#toc-child-concurrency
related discussion https://github.com/yarnpkg/yarn/issues/1874

ci/Jenkinsfile.tests
```diff
@@ -62,7 +62,16 @@ pipeline {
           steps {
             sh """#!/bin/bash
               set -eo pipefail
-              make test 2>&1 | tee -a ${LOG_FILE}
+              # set CHILD_CONCURRENCY to 1 to avoid node-gyp build error when running yarn install in parallel
+              CHILD_CONCURRENCY=1 make test 2>&1 | tee -a ${LOG_FILE}
```
jakubgs: Use `enviornment` section above:
```Groovy
      environment {
        /* Avoid node-gyp build error when running yarn install in parallel. */
        CHILD_CONCURRENCY = '1'
      }
```
Why repeat the same thing that the line after the comment already states?

yqrashawn: https://ci.status.im/blue/organizations/jenkins/status-mobile%2Fprs%2Ftests/detail/PR-15071/18/pipeline/39
failed again
`../modules/react-native-status/nodejs/status.cpp:1936:1: fatal error: opening dependency file ./Release/.deps/Release/obj.target/status_nodejs_addon/modules/react-native-status/nodejs/status.o.d.raw: No such file or directory`

jakubgs: Well, where did you get this variable from?

* https://github.com/yarnpkg/yarn/issues/6118

jakubgs: Also, not sure why we need to run `yarn install` in the target itself, but if that's absolutely necessary and it causes issues just drop the parallel part and let them run sequentially, should be fine if they don't take longer than builds.

yqrashawn: here https://github.com/status-im/status-mobile/pull/15071#issuecomment-1429621434

also tried `--ignore-scripts`, but component test relies on this to build `@react-native-community/blur`

jakubgs: And seems to me like the failure isn't due to concurrently running `yarn install` commands:
```
../modules/react-native-status/nodejs/status.cpp:1936:1: fatal error: opening dependency file ./Release/.deps/Release/obj.target/status_nodejs_addon/modules/react-native-status/nodejs/status.o.d.raw: No such file or directory
```
https://ci.status.im/blue/organizations/jenkins/status-mobile%2Fprs%2Ftests/detail/PR-15071/20/pipeline

jakubgs: It is worth noting that running them in parallel makes the LOG files hard to read.
But that could be fixed by using multiple log files. Or running serially.

yqrashawn: > Also, not sure why we need to run `yarn install` in the target itself, but if that's absolutely necessary and it causes issues just drop the parallel part and let them run sequentially, should be fine if they don't take longer than builds.

it's used to build the status-go Node.js binding
https://github.com/status-im/status-mobile/commit/e4cba0f6636b60c5d19c6a2003e7433bcb6312d0#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R299

jakubgs: Hm, worked locally for me just fine.

yqrashawn: Ok, seems we don't need `yarn install` for component tests, since it's testing UI components

jakubgs: Right, well, if we use Nix shell that should run `yarn install` anyway(which is the main reason why it annoys me that it is part of `packages.json` config) and it will prevent conflicts since it does its won locking.

yqrashawn: Failed when rerun with `yarn install` removed

can't find `@react-native-community/blur` in component tests

my guess is there is a `prepare` hook in `@react-native-community/blur` package.json, and it somehow do cleanup before rebuild, removes the `lib` folder

try to run sequentially without `yarn install` in component tests

J-Son89: @yqrashawn, please ignore the issue with @react-native-community/blur. That is separate and is happening locally. Some changes got merged recently which brought that  issue in and needs to be addressed. In fact you can mute the tests that bring in those problems as it will make it easier to verify the CI is working correctly with all tests passing effectively

yqrashawn:
Synced with @J-Son89 the other day
Better separate this into two issues.

1. some of the packages in node_modules relies on scripts in package.json hooks, e.g. `prepare`, `install`, `preinstall`, `postinstall`, to generate files that needed in component tests 
2. two parallel `yarn install` conflicts with each other when running in CI, caused by `node-gyp`, which is required to build `react-native-status`

- [ ] no idea why point 1 don't affect `run-clojure`, check `yarn2nix`, `shadow-cljs` nodejs target
- [ ] point 2 is using node-pre-gyp, figure out why there's error when running `node-pre-gyp install --fallback-to-build` instead of `yarn install`

---

https://github.com/NixOS/nixpkgs/blob/07ce7b2f5b110c28b48c9690c5be130f5b19e6e6/pkgs/development/tools/yarn2nix-moretea/yarn2nix/default.nix#L421
`--ignore-scripts` is used in `yarn2nix`
so packages relies on hooks related to `yarn install` may causing problem

yqrashawn:
add lib folder to status-im/react-native-blur https://github.com/status-im/react-native-blur/pull/1
This will fix the can't find module @react-native-community/blur error in component tests

Parveshdhull:
Hi @yqrashawn, @J-Son89,

I was looking into why component tests are failing and getting the error `Cannot find module` once we [changed](https://github.com/status-im/status-mobile/pull/14783) `@react-native-community/blur"` source to `git+https://github.com/status-im/react-native-blur#refs/tags/v4.3.0-status`.

And it turns out it was not caused due to how we add the library (`v4.3.0-status`), the error remains even if I change the source to `git+https://github.com/Kureev/react-native-blur#refs/tags/v4.3.0`

This error starts after this commit https://github.com/Kureev/react-native-blur/commit/3c3cbe3527bf79593c7b67e38f718e46e3b02528, where they migrate code to typescript.

So I think we should fix this issue in our code, instead of modifying the library, as we also faced this issue with other libraries. wdyt?

Also, I will look more into this, and let you know if I find something.

Not sure, but looks helpful
https://stackoverflow.com/questions/59892582/jest-tests-not-working-after-moving-to-typescript
https://stackoverflow.com/questions/62505728/typescript-module-not-found-error-after-compilation

Parveshdhull:
> This error starts after this commit https://github.com/Kureev/react-native-blur/commit/3c3cbe3527bf79593c7b67e38f718e46e3b02528, where they migrate code to typescript.

Another change that might be causing the issue - `Refactored everything to use https://github.com/callstack/react-native-builder-bob`

Parveshdhull:
calling `yarn add --dev ts-jest` and adding `ts-jest` to preset probably should fix the issue. ([source](https://jestjs.io/docs/getting-started#using-typescript))

But we already have `react-native` as preset. And if I pass both, it throws error.
`Option "preset" must be of type: string but instead received: array`

**Probable Solution**: Using multiple presets: https://stackoverflow.com/questions/51002460/is-it-possible-to-use-jest-with-multiple-presets-at-the-same-time

Parveshdhull:
> Probable Solution: Using multiple presets: https://stackoverflow.com/questions/51002460/is-it-possible-to-use-jest-with-multiple-presets-at-the-same-time

Adding both preset using this method still gave error `cannot find module ...`
Followed this guide https://kulshekhar.github.io/ts-jest/docs/guides/react-native/, still same problem

Parveshdhull:
Sorry @yqrashawn, for spamming PR, I think we should create an issue

yqrashawn:
Hi @Parveshdhull the `Cannot find module` error is caused by no `lib` folder in `node_modules/@react-native-community/blur/`

It's included in `files` in `package.json` https://github.com/Kureev/react-native-blur/blob/master/package.json#L12 which means it's included when publishing to npm

We are using the `git+https` scheme so that there's no `lib` folder until we run `yarn install` (`yarn2nix` runs `yarn install` with `--ignore-scripts`), which will trigger the `prepare` script in `package.json` https://github.com/Kureev/react-native-blur/blob/master/package.json#L28

We can fix it by one of below means
1. add `lib` folder and other missing files in `package.json` `files` field into git
2. publish the package to npm and use the npm version instead of `git+https`, this seems to be the long term solution if we keep facing the need to use self patched packages
3. try to make it possible to run these hook scripts in `yarn2nix`, I tried this by setting `yarnFlags`in `nix/deps/nodejs/default.nix`, but yarn still print warning about scripts disabled. Although, I'm not really understand nix.

Parveshdhull:
Hi @yqrashawn, Thank you finding this fix. I merged your PR https://github.com/status-im/react-native-blur/pull/1 and bumped version to 
[4.3.1-status](https://github.com/status-im/react-native-blur/releases/tag/v4.3.1-status)

That error got fixed, but another similar error popped up
`/home/parvesh_monu/Desktop/ArchFiles/React/status-mobile/node_modules/react-native-reanimated/lib/createAnimatedComponent.js: Cannot find module '@babel/plugin-proposal-optional-chaining'`

Parveshdhull:
> That error got fixed, but another similar error popped up

Looks like this error was caused by some leftover local changes, working great now :tada:

yqrashawn:
Got these errors from 3 fails after deps change. The 4/5th run passes.

```
Cannot find module '@babel/runtime/helpers/interopRequireDefault' from 'index.js'
Cannot find module 'object-assign' from 'factoryWithTypeCheckers.js'
Cannot find module 'slash'
Cannot find module 'object-assign' from 'factoryWithTypeCheckers.js'
```

`jest` using node.js `os.tmpdir` as default cache dir, which got cleared after each run by `make purge`

yqrashawn:
Decide to put component test after unit test and run them sequentially. 
`make component-test` relies on the `yarn install` in `make test`

jakubgs (approved):

Parveshdhull (approved):

J-Son89 (approved):
Great work @yqrashawn 🚀

Issue #13756, opened 2022-08-03
UI Component - QR Code 


siddarthkay:
This issue is to track the creation of the component: QR Code

You can access the link at https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Mobile-Components?node-id=1338%3A34510

Please have a chat with the design team before starting with the component, and don't forget to add sizing to the issue

siddarthkay:
need this component for https://github.com/status-im/status-mobile/issues/13439 
I have done most of the work in the screen

siddarthkay:
decision pending on updates made to existing QR component.

Existing problems : 
- Having a logo in the centre of the QR code requires modification of QR code and currently the only library that supports it on the client side is not compatible with React Native : https://www.npmjs.com/package/qr-code-styling 
unless we decide to explore the webview route : https://github.com/kozakdenys/qr-code-styling/issues/118
https://github.com/kozakdenys/qr-code-styling/issues/118

- Modifying the edge of the circles also requires a similar library.

siddarthkay:
As recently discussed the design team is okay with the edges not being circular but all they want is the ability to place a logo in the centre of the QR code which helps with identifying what the QR code is generated for.

This can be achieved by overlapping the desired logo on top of the generated QR.

I now need to experiment with the QR code's Error Correction Level (ref : https://github.com/soldair/node-qrcode#error-correction-level) and the QR code's capacity (ref : https://github.com/soldair/node-qrcode#qr-code-capacity) to ensure there is no information loss.

Essentially what happens when there are logos present inside QR codes it corrupts the information to some level and the above values must be adjusted to make sure no information is lost when reading the QR code :)

siddarthkay:
I see an attempt was made to do this 2 years ago : https://github.com/status-im/status-mobile/pull/10703 
and it makes sense for the core logic of QR code generation to be on the status-go side so that in the future even Desktop Team can use this.

PR #15076, opened 2023-02-13
Allow user to add contact by scanning their QR code (with a fake QR code reader)

### LINKED ISSUE #14482: Allow user to add contact by scanning their QR code
As a user
I would like to add a contact by scanning a QR code

Out of scope:
- Retrieve user info
- User's profile page

### Designs

https://www.figma.com/file/7KIYbhoqNGAIFonE0w9TDz/Messages-for-Mobile?node-id=628%3A174935&t=D6cAP4sUcnKzVSgH-0

Depends on: https://github.com/status-im/status-mobile/issues/14477

alwx:
fixes #14482
HOWEVER this doesn't add a new QR code reader and uses a fake one instead. Issue for the QR code reader itself: https://github.com/status-im/status-mobile/issues/13756

#### Platforms

- Android
- iOS

##### Functional

- add contact scren
- 
status: ready


src/status_im2/contexts/add_new_contact/views.cljs
```diff
@@ -42,7 +43,8 @@
                                    600)})]
         [quo/button
          (merge style/button-qr
-                {:on-press #(js/alert "TODO: to be implemented")})
+                {:on-press #(rf/dispatch [::qr-scanner/scan-code
```
flexsurfer, line 46: we shouldn't use namespaced keywords, and use status-im namespace in status-im2, there are two ways, simple, just add one more key to events , or replace namespaced keyword, but it think first one is simpler

ilmotta (approved):

src/status_im2/contexts/add_new_contact/views.cljs
ilmotta: I think the namespaced keyword fix is not really the scope of this PR, given the namespace `status-im.qr-scanner.core` already exists and it's not following the guidelines at all in many other event names. I would say it would be better to do this refactoring separately.

flexsurfer: yes, so we could just add new event name , its not refactoring

ilmotta: But @flexsurfer the event `::qr-scanner/scan-code` already exists and has a couple usages in the codebase. Creating a new one without changing the original would be unnecessary duplication just to get a different event keyword. if I understand correctly your suggestion I think that would be worse.

flexsurfer (approved):

PR #15078, opened 2023-02-14
Zoomable Image

### LINKED ISSUE #15038: Zoomable Image component
Implement a zoomable image component to be used in the lightbox screen

OmarBasem:
fixes: https://github.com/status-im/status-mobile/issues/15038

This PR implements a zoomable image component using Gesture Handler and Reanimated.

The component has 5 gestures:

**1. Tap gesture**: 
  - Tap to toggle surrounding content opacity
 
**2. Double tap gesture**:
 - Double tap to zoom
 - Detects the focal x and y positions and translates the image accordingly

**3. Pinch gesture**: 
 - Pinch with two fingers to zoom
 - Detects the focal x and y positions and translates the image accordingly
 - On pinching again, it can calculate the currently focused part of the image to zoom to the right spot again precisely
 - Pinch and pan simultaneously

**4. Pan-X gesture**:
 - Drag to translate the image along the x-axis
 - Snaps the image back to its edge when being dragged beyond its width
 - Applies decaying animation based on velocity after lifting the touch of the screen
 
**5. Pan-Y gesture**:
 - Drag to translate the image along the y-axis
 -  Snaps the image back to its edge when being dragged beyond its height
 - Applies decaying animation based on velocity after lifting the touch of the screen

status: ready


src/status_im2/subs/root.cljs
```diff
@@ -33,6 +33,7 @@
 (reg-root-key-sub :view-id :view-id)
 (reg-root-key-sub :screen-params :navigation/screen-params)
 (reg-root-key-sub :shared-element-id :shared-element-id)
+(reg-root-key-sub :close-lightbox :close-lightbox)
```
flexsurfer: close-lightbox naming is a little bit confusing, what's the purpose of this prop?

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+    (.Tap gesture/gesture)
+    (.onStart #(on-tap))))
+
+(defn double-tap-gesture
+  [{:keys [scale pan-x pan-x-start pan-y pan-y-start]}
+   {:keys [width height screen-height]}
+   {:keys [y-threshold-scale]}
+   rescale]
+  (->
+    (.Tap gesture/gesture)
```
flexsurfer: so the idea of having react-native.* namespaces is to hide all interactions with js objects, so it would be cool to move tap and numbersOfTaps to gesture namespace, we already have onStart, but we need to make it more general and remove pan argument

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+   rescale]
+  (->
+    (.Pinch gesture/gesture)
+    (.onStart (fn [evt]
+                ;; Read the x and y touch positions on the screen
+                (reset! focal-x (.-focalX evt))
+                (reset! focal-y (.-focalY evt))))
+    (.onUpdate (fn [evt]
+                 ;; First, find the scale ration
+                 (let [new-scale     (* (.-scale evt) (get-val saved-scale))
```
J-Son89: personally, I prefer code without such comments. You can put this into helper functions which explains what each part is doing and then you should naturally need less comments 

```
(defn find-pinch-x [width saved-scale pinch-x-start pan-x-start focal-x scale-ratio]
  (let [scaled-width  (/ width (get-val saved-scale))
        x-start       (+ (get-val pinch-x-start) (get-val pan-x-start))
        center-x      (get-current-center width scaled-width x-start)]
    (* (- center-x @focal-x) scale-ratio)))

(defn find-pinch-y [height pinch-y-start scale pan-y-start focal-y scale-ratio]
  (let [scaled-height (/ height (get-val scale))
        y-start       (+ (get-val pinch-y-start) (get-val pan-y-start))
        center-y      (get-current-center height scaled-height y-start)]
    (* (- center-y @focal-y) scale-ratio)))

(defn pinch-gesture
  [{:keys [width height]}
   {:keys [saved-scale scale pinch-x pinch-y pinch-x-start pinch-y-start pan-x-start pan-y-start]}
   {:keys [pan-x-enabled? pan-y-enabled? x-threshold-scale y-threshold-scale focal-x focal-y]}
   rescale]
  (->
   (.Pinch gesture/gesture)
   (.onStart (fn [evt]
                ;; Read the x and y touch positions on the screen
               (reset! focal-x (.-focalX evt))
               (reset! focal-y (.-focalY evt))))
   (.onUpdate (fn [evt]
                (let [new-scale     (* (.-scale evt) (get-val saved-scale))
                      scale-ratio   (get-scale-ratio new-scale (get-val saved-scale))
                      new-pinch-x   (find-pinch-x width saved-scale pinch-x-start pan-x-start focal-x scale-ratio) 
                      new-pinch-y    (find-pinch-y height pinch-y-start scale pan-y-start focal-y scale-ratio)]
                  (set-val pinch-x (+ new-pinch-x (get-val pinch-x-start)))
                  (set-val pinch-y (+ new-pinch-y (get-val pinch-y-start)))
                  (set-val scale new-scale))))))
```

J-Son89: same for `onStart` method, could be
```
(defn read-x-and-y-on-screen [focal-x focal-y]
  (fn [evt]
                ;; Read the x and y touch positions on the screen
    (reset! focal-x (.-focalX evt))
    (reset! focal-y (.-focalY evt))))

(defn pinch-gesture
  [{:keys [width height]}
   {:keys [saved-scale scale pinch-x pinch-y pinch-x-start pinch-y-start pan-x-start pan-y-start]}
   {:keys [pan-x-enabled? pan-y-enabled? x-threshold-scale y-threshold-scale focal-x focal-y]}
   rescale]
  (->
   (.Pinch gesture/gesture)
   (.onStart (read-x-and-y-on-screen focal-x focal-y))
   ```

File: src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs [hunk 4]
```diff
@@ -0,0 +1,313 @@
...
+                           (set-val scale new-scale))))
+    (gesture/on-end
+     (fn []
+       (cond
+         ;; if less than min-scale, then scale back to x1
+         (< (get-val scale) min-scale)
+         (rescale min-scale)
+         ;; if greater than max-scale, then scale back to max-scale
+         (> (get-val scale) max-scale)
+         (do
```
J-Son89, line 132: I personally would do similar here too, and just put most of these actions in named helper functions below. Anyway, not sure on your/teams preference but consider doing so :)

Should help the need for so many comments in the file

src/status_im2/subs/root.cljs
OmarBasem: On ios, when attempting to navigate back while zoomed in, the shared-element transition animation doesn't execute properly, so we need to zoom out first. This prop is used a signal for that. Renamed it to `exit-lightbox-signal`.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: @flexsurfer fixed it.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+
+(def max-scale 5)
+
+(def init-offset 0)
+
+(def velocity-factor 0.25)
+
+(def default-duration 300)
+
+;; Some aliases for reanimated methods, as they are used 10s of times in this file
```
ilmotta: Comments like these are non-idiomatic in Clojure and even Lisps dare I say. You either create sections with triple/quadruple semicolons and add normal comments below and/or you document the functions with docstrings. But double-semicolon comments at the top-level should be avoided from this stylistic perspective. It's in the Clojure Style Guide too https://guide.clojure.style/#comments.

These sections are also valuable to Emacs users of the outline minor mode, because they can be narrowed, collapsed, etc.

TL;DR: Use 3 or 4 semicolons to separate different sections of the namespace.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+(defn pinch-gesture
+  [{:keys [width height]}
+   {:keys [saved-scale scale pinch-x pinch-y pinch-x-start pinch-y-start pan-x-start pan-y-start]}
+   {:keys [pan-x-enabled? pan-y-enabled? x-threshold-scale y-threshold-scale focal-x focal-y]}
+   rescale]
+  (->
+    (gesture/gesture-pinch)
+    (gesture/on-start (fn [evt]
+                        ;; Read the x and y touch positions on the screen
+                        (reset! focal-x (.-focalX evt))
```
ilmotta: The lack of type hints will generate warnings and there's always a chance that the release build can have subtle bugs because of this. The advanced compilation mode can get confused. I've faced this issue before in status-mobile and it's a nightmare to trace back the root cause if you don't know it's because of a lack of type hint. So we better always add the `^js` type hints in the correct places when doing interop.

But forget what I just said, it's even better if you can stick to `oops/oget`, it's in the guidelines ;)

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+  [animation]
+  (reanimated/get-shared-value animation))
+
+(defn set-val
+  [animation value]
+  (reanimated/set-shared-value animation value))
+
+(defn use-val
+  [value]
+  (reanimated/use-shared-value value))
```
ilmotta, line 35: I don't understand the purpose of these wrapper functions, they just create indirection and pass along the unmodified arguments. I think we should just stick to the normal `use-shared-value`, `set-shared-value`, etc.

I guess you really wanted shorter names :wink:, but I think if that's the case we should consider renaming the original functions in a separate PR if it bothers other people too, but we should avoid needless wrapper functions.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+
+(defn timing
```
ilmotta, line 37: This is the only wrapper function that actually does something different and it's general enough that it deserves being moved to `react-native.reanimated` IMO.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+  [size screen-size scale]
+  (/ (/ (- (* size scale) screen-size) 2) scale))
+
+(defn get-scale-ratio
+  [new-scale saved-scale]
+  (/ (- (dec new-scale) (dec saved-scale)) new-scale))
+
+(defn get-current-center
+  [size scaled-size offset]
+  (- (+ (/ (- size scaled-size) 2) (/ scaled-size 2)) offset))
```
ilmotta: Please, avoid math calculation without proper line breaks. In any Lisp it gets really difficult to read.

```clojure
(- (+ (/ (- size scaled-size) 2)
      (/ scaled-size 2))
   offset)
```

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+                 upper-bound (- (Math/abs max-offset) (get-val pinch-y))]
+             (set-val pan-y-start (get-val pan-y))
+             ;; Apply decaying animation
+             (set-val-decay pan-y (* (.-velocityY evt) velocity-factor) [lower-bound upper-bound])
+             (set-val-decay pan-y-start
+                            (* (.-velocityY evt) velocity-factor)
+                            [lower-bound upper-bound]))))))))
+
+;; A helper method to rescale and reset values
+(defn rescale-image
```
J-Son89, line 232: perhaps the name should reflect it's use case.
e.g `rescale-image-and-reset-values`

src/react_native/reanimated.cljs
```diff
@@ -137,3 +139,18 @@
...
+                    (with-decay (js-obj "velocity" velocity
+                                        "clamp"    (clj->js clamp)))))
+
+(def in-out
+  (.-inOut Easing))
+
+(defn with-timing-duration
+  [val duration]
+  (with-timing val
+               (js-obj "duration" duration
```
ilmotta: `#js {...}` could be used here and in other places in this file. I find the reader macro much easier to read. The `js-obj` function is only needed if you need to pass *interleaved* keys and values as arguments according to the docstring, but this is rarely the case in Clojure.

```clojure
(js-obj "duration" 100 "easing" 200) ;; => #js {:duration 100, :easing 200}
```

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,313 @@
...
+           height               (* image-height (/ (:width (rn/get-window)) image-width))
+           screen-height        (:height (rn/get-window))
+           dimensions           {:width         width
+                                 :height        height
+                                 :screen-height screen-height}
+           animations           {:scale         (use-val min-scale)
+                                 :saved-scale   (use-val min-scale)
+                                 :pan-x-start   (use-val init-offset)
+                                 :pan-x         (use-val init-offset)
+                                 :pan-y-start   (use-val init-offset)
```
J-Son89, line 279: why not put in a let variable? it's being called 8 times :)

src/react_native/reanimated.cljs
```diff
@@ -137,3 +139,18 @@
                                  (js-obj "mass"      mass
                                          "damping"   damping
                                          "stiffness" stiffness))))
+
+(defn animate-shared-value-with-decay
+  [anim velocity clamp]
+  (set-shared-value anim
+                    (with-decay (js-obj "velocity" velocity
+                                        "clamp"    (clj->js clamp)))))
```
ilmotta: This is more complicated than it needs to be. Call `clj->js` once and remove `js-obj`.

```clojure
(with-decay (clj->js {:velocity velocity :clamp clamp}))
```

ilmotta (commented):
You really did a great work with the animations :clap:

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: Thanks @ilmotta

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: Marking with `^js` seems to be more concise. Is there an advantage to using `oops/oget` over `^js`?

OmarBasem: @ilmotta Do we mark the top `evt` in the function argument, or also every single `evt`?

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: These are some aliases for reanimated methods (for shorter names), as they are used 10s of times in this file. Using `reanimated/set-shared-value` causes long lines all over the file.

I don't know what can we rename it to. It seems to me it is more of a clojure problem due to namespacing and kebab-case. Using it in JS would be `setSharedValue` vs in clojure `reanimated/set-shared-value` (14 vs 27 chars, almost double :(  ) 

When it occurs few times in a file, it is not a problem, but for this particular file, it looked a bit unreadable and ugly. Hence I used aliases.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: Currently in the `react-native.reanimated` there are few different versions of the `with-timing` function. The file needs some refactoring, and also its occurrences in the codebase. I will do it in a seperate PR.

src/react_native/reanimated.cljs
OmarBasem: changed it to `clj->js` to be consistent with the one below also

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: @J-Son89 fixed it.

OmarBasem:
> You really did a great work with the animations 👏

Thank you Icaro!

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
```diff
@@ -0,0 +1,312 @@
...
+  (/ (- (dec new-scale) (dec saved-scale))
+     new-scale))
+
+(defn get-current-center
+  [size scaled-size offset]
+  (- (+ (/ (- size scaled-size) 2)
+        (/ scaled-size 2))
+     offset))
+
+;;;; 5 Gestures: tap, double-tap, pinch, pan-x, pan-y
```
J-Son89, line 72: nit: maybe this can just be `;;;; Gestures` 
The likelihood is that if someone adds, updates or removes one of these gestures below for whatever reason then they are likely to miss this comment when updating and then the comment becomes fake news/outdated

J-Son89 (approved):
would be cool if you could include some screen shots or even better a short video of it working 👍

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
ilmotta: Good questions :)

>Do we mark the top evt in the function argument, or also every single evt?

We need to type hint only once per function body. If a function body has interop macros, like `(.-focalX evt)`, then the function argument should add the type hint `^js evt` to make sure the advanced compilation to Javascript does its best.

But we'll see plenty of cases where the type hint should be next to the Javascript object, for example when we need to deref and call a native Javascript method. In such cases, the type hint might need to be added multiple times.

```clojure
(.scrollToIndex ^js @scroll-ref)
```

>Marking with ^js seems to be more concise. Is there an advantage to using oops/oget over ^js?

I agree, interop macros can be more concise, but the team has decided on using `oops`. oops offers some advantages, like:

- Less risk of null/undefined exceptions when doing chaining properties. `oops/oget` gives this for free, sort of similar to `m.z?.a`, a new syntax already supported in major browsers.
- When using oops, the type hint is unnecessary. I think the macros will add them for us.
- oops also helps generate valid code when advanced optimizations are enabled (which they are in status-mobile). In some cases the compiler might rename some references and the code can break.

About conciseness, sometimes it might be good to have some exceptions about namespace aliasing. If oops ends up being so prevalent in the codebase, we could consider only using `:refer [oget]`, that should make things less fatiguing.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
ilmotta: I understand where you're coming from, but I disagree @OmarBasem. These wrapper functions only add complication to an already complicated code (pretty much all heavily animated code is difficult to digest). There's a better way.

>These are some aliases for reanimated methods (for shorter names), as they are used 10s of times in this file. Using reanimated/set-shared-value causes long lines all over the file.

The team decided to use long and explicit namespace aliases. And there's nothing stopping us from renaming `set-shared-value` to `set-val` in `react-native.reanimated`. This change alone would make the code much shorter, but let's not stop there.

Much more importantly, **the problem is not at the language (Clojure) level, it's at the abstraction level**. The repeated calls to `reanimated/set-shared-value` or `set-val` are ugly and unnecessary. Here's a better idea IMO, change the function to accept setting multiple values at once, so that the name of the function becomes way less relevant.

Instead of this:

```clojure
(set-val pan-x (timing init-offset duration))
(set-val pinch-x (timing init-offset duration))
(set-val pan-x-start (timing init-offset duration))
(set-val pinch-x-start (timing init-offset duration))
(set-val pan-y (timing init-offset duration))
(set-val pinch-y (timing init-offset duration))
(set-val pinch-y-start (timing init-offset duration))
(set-val pan-y-start (timing init-offset duration))
```

Better! Notice I used the word `animated` instead of `reanimated` too. This is common practice in Javascript projects.

```clojure
(animated/set-val
 pan-x (timing init-offset duration)
 pinch-x (timing init-offset duration)
 pan-x-start (timing init-offset duration)
 pinch-x-start (timing init-offset duration)
 pan-y (timing init-offset duration)
 pinch-y (timing init-offset duration)
 pinch-y-start (timing init-offset duration)
 pan-y-start (timing init-offset duration))
```

ilmotta: For creating true aliases in Clojure, you should use the pattern:

```clojure
(def set-val reanimated/set-shared-value)
```

But as I commented, I don't like this solution at all. Much better is to rename `reanimated/set-shared-value` to `reanimated/set-val` or something and also make the function accept multiple values.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
ilmotta: I would do the same @OmarBasem, thanks.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
ilmotta: But hey @OmarBasem, I'm not suggesting you implement any of the changes above in this PR. I'm just saying we should do the right thing that moves the needle for everybody, and creating aliases just for this namespace is a dubious practice at best.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: Thanks @ilmotta for the information. Refactored it to using `oget`

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: Thanks @ilmotta. You are right, we need to abstract it in a better way. I will try to make a PR for that.

OmarBasem:
> would be cool if you could include some screen shots or even better a short video of it working 👍

QA will test the PR 👍
And PR builds are available if you are interested in checking it yourself :)

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
OmarBasem: What do you mean? These are different animations.

src/react_native/reanimated.cljs
```diff
@@ -137,3 +139,18 @@
                                  (js-obj "mass"      mass
                                          "damping"   damping
                                          "stiffness" stiffness))))
+
+(defn animate-shared-value-with-decay
+  [anim velocity clamp]
+  (set-shared-value anim
+                    (with-decay (clj->js {"velocity" velocity
```
flexsurfer: if `clj->js` is used better to have a clojure map with keywords for keys

src/status_im2/contexts/chat/lightbox/view.cljs
```diff
@@ -161,31 +151,34 @@
...
-                :get-item-layout           get-item-layout
-                :viewability-config        {:view-area-coverage-percent-threshold 50}
-                :on-viewable-items-changed (rn/use-callback (fn [e]
-                                                              (on-viewable-items-changed e
-                                                                                         scroll-index)))
-                :content-container-style   {:justify-content :center
-                                            :align-items     :center}}]
-              [bottom-view messages index scroll-index insets opacity-value]])])]))])
+          ;; We use setTimeout to enqueue `scrollToIndex` until the `data` has been updated.
+          (js/setTimeout #(.scrollToIndex ^js @flat-list-ref #js {:animated false :index index}) 0)
```
flexsurfer: shouldn't this be in effect? as it's f component

src/status_im2/subs/root.cljs
ilmotta: I think we (team) had an opinion that we should not create global keywords anymore, without a qualification. This can help logically group related keywords. For example, name it `:lightbox/exit-signal` or something. Let's take advantage of this Clojure feature of keywords! I'm not saying we should nest, that's a discussion we had recently and it's a completely different topic.

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
ilmotta: I would gladly review it, thank you @OmarBasem

ilmotta (approved):

VolodLytvynenko:
Hi @OmarBasem Thank you for PR. I didn't finish testing yet. Just a quick update. Here is one issue that causes the failing of other e2e. 

## ISSUE 1: The  'underfined is not a function' error is shown after user tap [+] button on the communities page
**Steps to reproduce:**
1. Go to community tab -> tap [+]

**Actual result:**
'underfined is not a function' error is shown

https://user-images.githubusercontent.com/52490791/220367107-003324d4-9c27-4f5e-ad83-c5b24063ba08.mp4

**Expected result:**
Community is able to be created

**Logs:**
[Status.log](https://github.com/status-im/status-mobile/files/10794216/Status.log)
[geth.log](https://github.com/status-im/status-mobile/files/10794217/geth.log)

OmarBasem:
Hi @VladimrLitvinenko, thanks for your review. The issue if fixed now.

VolodLytvynenko:
Hi @OmarBasem . Thank you for PR. Here are some found issues

## ISSUE 2: The "null is an object (evaluating ju(hoe).scrollToIndex)" error is shown after the images are scrolled
**Steps to reproduce:**
1. Open chat
2. Send 2-3 images in one album
3. Open the image
4. Single tap on the image
5. Scroll the images

**Actual result:**
The "null is an object (evaluating ju(hoe).scrollToIndex)" error is shown

https://user-images.githubusercontent.com/52490791/220650945-42dad1b0-833e-4a82-89f9-0ccfb4aaf3cc.mp4

**Expected result:**
The error is not shown after images are scrolled

**ENV:** 
- Real device: iPhone 11 PRO max, version 15;
-  Real device: Huawei p20 light, android 9

**Logs:**
[Status.log](https://github.com/status-im/status-mobile/files/10804800/Status.log)
[geth.log](https://github.com/status-im/status-mobile/files/10804801/geth.log)

## ISSUE 3: No ability to drag the along the y-axis if the image was zoomed by double tap
**Steps to reproduce:**
1. Send the horizontal image (example: https://www.dropbox.com/s/09epe4qpp0vfv3b/coffee1.jpg?dl=0)
2. Double-tap to zoom it
3. Pinch with two fingers to zoom some area
4. Try to drag along the y-axis

**Actual result:**
The drag along the y-axis is not available
https://user-images.githubusercontent.com/52490791/220658151-f2c3638e-b945-48ec-8ed9-cf1a37115d08.mp4


**Expected result:**
The drag along the y-axis is available

**ENV:** 
- Real device: iPhone 11 PRO max, version 15;
-  Real device: Huawei p20 light, android 9

## ISSUE 4: The images overlap each other if the image from an album is zoomed and scrolled
**Steps to reproduce:**
1. Send at least 2-3 images
2. Open image
3. Zoom it
4. Tap on the next image in the image scroll bar
image


**Actual result:**
The next image overlaps the previously zoomed image

https://user-images.githubusercontent.com/52490791/220660777-bdea7d37-cdf4-443a-ba59-994ead95aa02.mp4

**Expected result:**
The next image overlaps the previously zoomed image

## ISSUE 5: The images are not able to be opened by scrolling while one image is zoomed
**Steps to reproduce:**
1. Send 2-3 image
2. Open the image
3. Zoom it
4. Try to scroll to the next image

**Actual result:**
The next images are not able to be opened by scrolling while the initial image is zoomed

https://user-images.githubusercontent.com/52490791/220672351-99f9d631-d143-4145-ae25-7b4e420da284.mp4



**Expected result:**
The next image is able to be opened  by scrolling

(how it is implemented in telegram service)

https://user-images.githubusercontent.com/52490791/220672425-8404a7dc-be25-4c4d-a4b0-e0ba9b47fc4c.mp4

**ENV:** 
- Real device: iPhone 11 PRO max, version 15;
- Real device: Huawei p20 light, android 9

VolodLytvynenko:
@OmarBasem One question:
- Is it expected that it is possible to scroll to a black background or the maximum scroll should be only to the edge of the image? 

https://user-images.githubusercontent.com/52490791/220683015-aea97f6e-b4b9-440f-913e-38e97d1c1c40.mp4

VolodLytvynenko:
@OmarBasem It's a bit inconvenient that the image can't be closed when scrolling up or down. Not sure if this is in the range of this PR or should not be implemented at all. What do you think?

**Steps to reproduce:**
1. Open the image
2. Scroll up or down

**Actual result:**
image is not closed after scrolling up or scrolling down

https://user-images.githubusercontent.com/52490791/220688690-c4f7b6f1-eb48-477c-af8a-922ee502f6d8.mp4

**Expected result:**
image is closed after scrolling up or scrolling down

how it works on Telegram and WhatsApp services
https://user-images.githubusercontent.com/52490791/220688742-55d42094-9e99-4bde-a747-f8312e344aff.mp4

OmarBasem:
Hi @VladimrLitvinenko, thanks for your work and reviews!

This PR implements zooming gestures. Other gestures like closing the lightbox screen when dragging up or down is not part of this PR.

Issue 2: fixed.
Issue 3: Well spotted! fixed.
Issue 4: fixed.
Issue 5: not part of this PR

Answer to your question: Yes it is expected to drag beyond the image size and then snap back on release. However, on dragging beyond a certain threshold it would exit the screen, which would be implemented in another PR.

VolodLytvynenko:
@OmarBasem Thank you for your work and additional thanks for such detailed description of the task. 2-5 Issues are fixed. Could you merge it?

OmarBasem:
@VladimrLitvinenko Thank you!

src/status_im2/contexts/chat/lightbox/zoomable_image/view.cljs
Parveshdhull: Hi, Sorry for bumping this thread, but I just noticed that we are making abstraction changes in PR https://github.com/status-im/status-mobile/pull/15175 based on above comment.

> Better! Notice I used the word animated instead of reanimated too. This is common practice in Javascript projects.

[Animated](https://reactnative.dev/docs/animated) and [Reanimated](https://docs.swmansion.com/react-native-reanimated/) are different libraries and [Animated values](https://reactnative.dev/docs/animatedvalue) and [Shared values](https://docs.swmansion.com/react-native-reanimated/docs/fundamentals/shared-values/) are different things, abstracting and shortening names here will probably be misleading.

And as I remember, we are trying to avoid abstraction if possible to avoid confusion. Maybe @J-Son89 has some thoughts

OmarBasem: @Parveshdhull I am aware of these differences. But the current abstraction of reanimated is causing very long names, especially when used multiple times in a file. Using `reanimated/set-shared-value animation (reanimated/with-timing value)` 10s of times in a file makes it more difficult to read. 

That issue https://github.com/status-im/status-mobile/issues/15176 is an attempt to improve the abstraction. For example (just suggestions) by using animated instead of reanimated (which is what is actually used in reanimated docs), and using shorter names like `set-val` instead of `set-shared-value`.

OmarBasem: And probably we should get rid of React Native's Animated namespace/definitions/code since we shouldn't be using it and to avoid confusion.

Parveshdhull: `reanimated/set-shared-value animation (reanimated/with-timing value)`
1. Here function name is only `set-shared-value` (16 chars) others are just params. If we are also counting them in name, then we are in trouble ([quo/button](https://github.com/status-im/status-mobile/blob/69c3a7235709c108621143f01edcd3a7754d5041/src/status_im2/contexts/syncing/sheets/sync_device_notice/view.cljs#L52))
2. You can just refer to functions with names and use like
```
[react-native.reanimated :refer [use-shared-value set-shared-value]]

(set-shared-value 4) 
```
  Only 21 chars, and you will get rid of reanimated

3. We do have a helper function for animating value with timing `animate-shared-value-with-timing`
 (It also has a bigger name :))
4. Changing `set-shared-value` to `set-value` only decreases your code from 68 chars to 61, not sure if it's worth the confusion it might create.
5. Even if the code is a little lengthier, in my opinion, all the parts are required to convey the work/process that the code is doing.

If similar code keeps repeating in your file, I guess it might be ok to create localized abstraction in that file, but not sure we need do this globally (as mentioned in https://github.com/status-im/status-mobile/issues/15176)

OmarBasem: I generally agree with this @Parveshdhull. I am okay with using `refer` for this case, and also okay with creating localized abstraction when required, but do not know what others think about that. If others are okay with that I can close the issue https://github.com/status-im/status-mobile/issues/15176.

cc @ilmotta @flexsurfer what are your thoughts guys?

ilmotta: About the names:

When I read something as `reanimated/set-value` I know it's setting a shared value because of the namespace. The `shared` word is quite redundant in the context of reanimated. I also see that `reanimated` (using the prefix `re`) is redundant. In most React projects I've seen the alias is just `animated`. The vast majority of the time we're not mixing animated + reanimated code, so just `animated` as alias seems better IMO. Btw the alias name can be reinforced by the linter.

>And as I remember, we are trying to avoid abstraction if possible to avoid confusion.

@Parveshdhull I don't see any problem in re-evaluating the current abstractions and improving them as we go. It's a function just as any other function, nothing special about it being related to animations. Omar felt they're not very ergonomic, and I agree while reviewing his PRs.

>Only 21 chars, and you will get rid of reanimated

This *convenience argument* (less chars) is used for pretty much all usages of `:refer`, but lots of Clojure projects avoid them, because it creates a burden on the developer, that now needs to know where each function comes from to reason about the code. I've seen many well known Clojure devs recommending against `:refer` (I don't have the references obviously). I think our guidelines are err'ing on the side of readability, not shortness. We can have exceptions sure (we do have some in test namespaces), but each exception puts more burden on collaborators to know when to refer vs when to alias.

I'd say, let's keep avoiding `refer` and not open more exceptions. I can see a lot of repetition in other namespaces, but they're not using `:refer`, so why should we open an exception here?

ilmotta: @OmarBasem ^^

Parveshdhull: Hi @ilmotta,
Thank you very much for your valuable input.

If `use-val`, `set-val`, and `animated` are all understable then sounds awesome, I am all in for these changes. I just wanted to be sure we are not compromising the understandability for the short names.

> The shared word is quite redundant in the context of reanimated.

It will not be redundant in the case of an `animated` alias. Unless we check what we are importing.

> In most React projects I've seen the alias is just animated.

Just for curiosity, please can you refer to them?

> (omar): Using it in JS would be setSharedValue vs in clojure reanimated/set-shared-value (14 vs 27 chars, almost double :( )

I don't recommend refer too, it was a suggestion because of this statement.  That it's not a shortcoming of Clojure itself, we can achieve similar here.

@OmarBasem 
- If we are going to use these names for function, please make changes in the original file instead of creating another wrapper.
- `apply-animations-to-style` also sounds confusing I think we can rename this too into `use-animated-style` wdyt?
- Please can we also do these changes in separate PR. As https://github.com/status-im/status-mobile/pull/15175 is already big

ilmotta: > It will not be redundant in the case of an animated alias. Unless we check what we are importing.

Good point @Parveshdhull

> In most React projects I've seen the alias is just animated.

I'm mostly referring to blog posts and courses, pretty much all use `Animated` to import the default module. Also the `reanimated` docs all use the `Animated` alias, so I guess that's one strong reason why the JavaScript community went for `Animated` instead of `Reanimated`.

It's nothing major, but it'd be nice if we could use the same alias, although I wouldn't like to confuse it with the actual animated module from RN. Well, perhaps better keep it as is to avoid confusion ;)

> Please can we also do these changes in separate PR. As https://github.com/status-im/status-mobile/pull/15175 is already big

Totally agree @Parveshdhull, it should be done in a separate PR.

OmarBasem: Thanks @ilmotta for your input!

Sure @Parveshdhull in a separate PR!

PR #15094, opened 2023-02-15
[Fixes: #15093] Add support for mention @everyone

### LINKED ISSUE #15093: Support posting @everyone mentions in communities
As a user
I want to be able to @everyone mention in a community
So that I can notify everyone

### Description

When typing @ev.. in the autocomplete for mention, there should be an option to mention everyone. 

A lot of mention code should be ported in status go for compatibility with desktop, so as a part of this feature, I'd suggest creating an endpoint (if it doesn't exist already and desktop uses it), to suggest mentions based on some text in status-go.

Feel free to ping me so we can have a chat before starting on this issue.

cammellos:
This commit adds support for displaying @everyone mention.

### Testing

1) Create a community and invite user B on Desktop
2) On Desktop, post `@everyone` in a community channel
3) Open the community on mobile (this PR)

`@everyone` should be displayed as a mention, clicking on it should have no effect.

@ulisesmac @flexsurfer this is ready to be reviewed now, thank you!

### DIFF
--- a/src/status_im2/subs/chat/messages.cljs
+++ b/src/status_im2/subs/chat/messages.cljs
@@ -3,6 +3,7 @@
             [status-im.chat.models.reactions :as models.reactions]
             [status-im2.constants :as constants]
             [status-im2.contexts.chat.messages.list.events :as models.message-list]
+            [utils.i18n :as i18n]
             [utils.datetime :as datetime]))
 
 (defn intersperse-datemark
@@ -271,3 +272,13 @@
                           joined
                           loading-messages?)
            (albumize-messages))))))
+
+(re-frame/reg-sub
+ :messages/resolve-mention
+ (fn [[_ mention] _]
+   [(re-frame/subscribe [:contacts/contact-name-by-identity mention])])
+ (fn [[contact-name] [_ mention]]
+   (if (= mention constants/everyone-mention-id)
+     (i18n/label :t/everyone-mention)
+     contact-name)))
+
--- a/src/status_im/ui/screens/chat/components/contact_request.cljs
+++ b/src/status_im/ui/screens/chat/components/contact_request.cljs
@@ -42,7 +42,7 @@
              (get-quoted-text-with-mentions children)
 
              (= type "mention")
-             @(re-frame/subscribe [:contacts/contact-name-by-identity literal])
+             @(re-frame/subscribe [:messages/resolve-mention literal])
 
              (seq children)
              (get-quoted-text-with-mentions children)
--- a/src/status_im/ui/screens/chat/components/reply.cljs
+++ b/src/status_im/ui/screens/chat/components/reply.cljs
@@ -43,7 +43,7 @@
              (get-quoted-text-with-mentions children)
 
              (= type "mention")
-             @(re-frame/subscribe [:contacts/contact-name-by-identity literal])
+             @(re-frame/subscribe [:messages/resolve-mention literal])
 
              (seq children)
              (get-quoted-text-with-mentions children)
--- a/src/status_im/ui/screens/home/views/inner_item.cljs
+++ b/src/status_im/ui/screens/home/views/inner_item.cljs
@@ -49,7 +49,7 @@
 
                  "mention"
                  {:components [react/text-class
-                               @(re-frame/subscribe [:contacts/contact-name-by-identity literal])]
+                               @(re-frame/subscribe [:messages/resolve-mention literal])]
                   :length     4} ;; we can't predict name length so take the smallest possible
 
                  "status-tag"
--- a/src/status_im/ui2/screens/chat/components/reply/view.cljs
+++ b/src/status_im/ui2/screens/chat/components/reply/view.cljs
@@ -22,7 +22,7 @@
              (get-quoted-text-with-mentions children)
 
              (= type "mention")
-             (rf/sub [:contacts/contact-name-by-identity literal])
+             (rf/sub [:messages/resolve-mention literal])
 
              (seq children)
              (get-quoted-text-with-mentions children)
--- a/src/status_im/ui2/screens/chat/messages/message.cljs
+++ b/src/status_im/ui2/screens/chat/messages/message.cljs
@@ -35,7 +35,7 @@
 
 (defn mention-element
   [from]
-  (rf/sub [:contacts/contact-name-by-identity from]))
+  (rf/sub [:messages/resolve-mention from]))
 
 (defn render-inline
   [_message-text content-type acc {:keys [type literal destination]}]
--- a/src/status_im2/constants.cljs
+++ b/src/status_im2/constants.cljs
@@ -281,3 +281,5 @@
 (def ^:const local-pair-action-connect 1)
 (def ^:const local-pair-action-pairing-account 2)
 (def ^:const local-pair-action-sync-device 3)
+
+(def ^:const everyone-mention-id "0x00001")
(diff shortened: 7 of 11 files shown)


src/status_im2/subs/chat/messages.cljs
```diff
@@ -267,3 +268,13 @@
                           joined
                           loading-messages?)
            (albumize-messages))))))
+
+(re-frame/reg-sub
+ :messages/resolve-mention
+ (fn [[_ identity] _]
+   (if (= identity constants/everyone-mention-id)
+     [(atom [(i18n/label :t/everyone-mention)])]
```
cammellos: This looks a bit odd, basically, I don't think we should make a decision in the render and call a different subscription for mentions that are not public keys, so I like the fact that there's a subscription that encapsulate all the logic.
It's though a bit odd that I am shaping the data into what the computation function wants, not sure whether it should be just a signal function and remove the computation function (it's not collating multiple signals anyway), or keeping it as it is.

any opinion?  cc @flexsurfer @ilmotta @erikseppanen @Parveshdhull @ulisesmac 
once we made a decision, I can write a test for it.

ilmotta: I honestly like the solution. I agree it's self-contained and gets the job done. Clever usage of the atom to fulfill the interface of the signal. The word clever sometimes sends shivers down my spine, but not in this case ;)

src/status_im2/subs/chat/messages.cljs
```diff
@@ -267,3 +268,13 @@
                           joined
                           loading-messages?)
            (albumize-messages))))))
+
+(re-frame/reg-sub
+ :messages/resolve-mention
+ (fn [[_ identity] _]
```
ilmotta: I would just recommend not using the word `identity` because it's shadowing a core Clojure function. It confused me a little bit when reading for the first time.

src/status_im2/subs/chat/messages.cljs
ulisesmac: It looks good that the subcription handles all the logic :+1: 

I'm not sure about "faking" a node in the signal graph (`[(atom [(i18n/label :t/everyone-mention)])]`). IDK if this could break re-frame at any level now or in the future :thinking: , I'd prefer to avoid it  because this could encourage to replicate the pattern elsewhere.

I thought of two alternatives:

1. Constantly (and not dinamically) subscribe and make the decision in the computing function:
```clojure
(re-frame/reg-sub
 :messages/resolve-mention
 (fn [[_ identity] _]
   [(re-frame/subscribe [:contacts/contact-name-by-identity identity])]) ; sub also changed
 (fn [[contact-name] [_ identity]]
   (if (= identity constants/everyone-mention-id)
     (i18n/label :t/everyone-mention)
     contact-name)))
```

2. If we care about subscribing to a value that will probably not be used (such as my first alternative), we could subscribe to nodes closer to the root in the signal graph:

```clojure
(re-frame/reg-sub
 :messages/resolve-mention
 :<- [:contacts/contacts]
 :<- [:multiaccount]
 (fn [[contacts current-multiaccount] [_ identity]]
   (if (= identity constants/everyone-mention-id)
     (i18n/label :t/everyone-mention)
     (-> contacts
         (multiaccounts/contact-by-identity identity)
         (multiaccounts/contact-two-names-by-identity current-multiaccount identity)
         (first)))))
```


What do you think?

ilmotta: I like your first suggestion too @ulisesmac.

flexsurfer: why don't just add a contact for `constants/everyone-mention-id` ?

flexsurfer: we could even create it in status-go, so desktop could use it as well

cammellos: > why don't just add a contact for `constants/everyone-mention-id` ?

I think `contacts` are already enough messy without having also to add synthetic ones :) as well the fact is that eventually we might want to add mention groups (i.e `@mobile`), and sticking everything into `contacts` is likely to become a bit confusing (and as we discussed, contacts are already incredibly confusing).

I like either of @ulisesmac solutions, the only concern is that from what I gathered (tested very quickly), they are a bit more expensive, since we execute the code for `[:contacts/contact-name-by-identity]`, or `:contacts/contacts, :multiaccount`, and `contacts/contacts` is heavy (we should try not to use it).
Generally that would not be an issue, but with messages, it's a place where it's easy to die of a thousand paper cuts, since it's by far the heaviest screen, so not sure, a bit on the fences :( , maybe it doesn't matter so much, but I am conflicted :)

ilmotta: >I like either of @ulisesmac solutions, the only concern is that from what I gathered (tested very quickly), they are a bit more expensive

That's the reason I like his first suggestion, but I think the second one causes too much unnecessary recomputation, I'd avoid that.

ulisesmac: AFAIK, subscriptions are cached and the second alternative is subscribing to the same root nodes as `:contacts/contact-name-by-identity]` & `[:contacts/contact-two-names...]`.

I'm not completely sure, but I'd say that the condition to recompute both the first and second alternatives are the same (they are recomputed if either `:contacts/contacts` or `:multiaccount` change), so probably the performance is almost the same 🤔

I'd implement the solution that I think is better and if a significant performance problem arises, then I'd open an issue to optimize the char screen. Probably there are more fragments of code than this subscription that are computationally expensive in that screen.

cammellos: > AFAIK, subscriptions are cached and the second alternative is subscribing to the same root nodes as :contacts/contact-name-by-identity] & [:contacts/contact-two-names...].

Yes, in the second example `contacts/contacts` would be calculated only once, so you are probably right, I'll double check that, thanks
I think the first one since it's parameterized by `identity`, will trigger `(re-frame/subscribe [:contacts/contact-name-by-identity identity])` for the `@everyone` mention as well, while the second one as you mention will be caching `contacts/contacts`, and always do the work on `displayed-name` for contacts when any of the contact change, while it will short-circuit for mention.

Regardless, happy to go with either of the two,  I think your advice is sound, we can revisit as we go along.
Thanks all for the input!

ilmotta: I'm not 100% sure either about how the cache would behave. Only proper measurement can really tell what re-frame is actually doing for this particular sub graph, although I prefer to err on the side of performance.

But also in terms of simplicity. The second option is sort of out of question to me because it's unbundled copy & pasta from the original subscription implementation `:contacts/contact-name-by-identity`. Given that layer-3 subs should always be tested in status-mobile, I think this the second solution doesn't have enough upsides to justify the increased effort + complexity.

If we believe (ideally measure) performance is on the same ballpark, then we should pick the simplest.

cammellos: first one it is then :)

flexsurfer (approved):

Parveshdhull (approved):

ulisesmac (approved):
Looks good :)

churik:
@cammellos 
everything works as you described in https://github.com/status-im/status-mobile/pull/15094#issue-1586260563, however I see some degradation in comparison with `develop`


Before there was a notification to the activity center and it is gone in PR builds:
IMAGE 2023-02-21 11:36:00
I consider this notification valuable for the user, should we add it here or log the issue separately?
Not sure about push notification in a particular case.

churik:
Turned out that it is an issue in `develop` (notifications are still working in RC-1) 
Will report separately, thank you for fix!

PR #15102, opened 2023-02-16
[WIP] Mock gating on Status community

### LINKED ISSUE #14462: Move mock data to central location and out of views.

vkjr:
fixes #14462

### Summary
Mock gating added to default Status community

status: wip

### DIFF
--- a/src/status_im2/contexts/communities/overview/view.cljs
+++ b/src/status_im2/contexts/communities/overview/view.cljs
@@ -31,7 +31,7 @@
     (utils/join-existing-users-string user-list)]])
 
 (defn channel-token-gating-details
-  [name token-gating emoji channel-color]
+  [name gates emoji channel-color]
   [rn/view {:height 350 :margin-top 20}
    [quo/token-gating
     {:channel {:name                   name
@@ -42,15 +42,15 @@
                                          (js/alert
                                           "Entered channel"
                                           "Wuhuu!! You successfully entered the channel :)"))
-               :gates                  token-gating}}]])
+               :gates                  gates}}]])
 
 (defn open-channel-token-gating-details
-  [name token-gating emoji channel-color]
+  [name gates emoji channel-color]
   (rf/dispatch
    [:bottom-sheet/show-sheet
     {:content
      (fn []
-       [channel-token-gating-details name token-gating emoji channel-color])
+       [channel-token-gating-details name gates emoji channel-color])
      :content-height 210}]))
 
 (defn layout-y
@@ -160,13 +160,13 @@
                  :gates                    tokens}}]])
 
 (defn add-on-press-handler
-  [community-id {:keys [name emoji id locked? token-gating] :or {locked? false} :as chat}]
+  [community-id {:keys [name emoji id locked? gates] :or {locked? false} :as chat}]
   (merge
    chat
-   (if (and locked? token-gating)
+   (if (and locked? gates)
      {:on-press #(open-channel-token-gating-details
                   name
-                  token-gating
+                  gates
                   emoji
                   (colors/custom-color :pink 50))}
 
@@ -210,14 +210,16 @@
 
 (defn community-content
   [{:keys [name description locked joined images
-           status tokens tags requested-to-join-at id]
+           status tags requested-to-join-at id]
     :as   community}
    {:keys [on-categories-heights-changed
            on-first-channel-height-changed]}]
-  (let [pending?          (pos? requested-to-join-at)
-        thumbnail-image   (get-in images [:thumbnail])
-        chats-by-category (rf/sub [:communities/categorized-channels id])
-        users             (rf/sub [:communities/users id])]
+  (let [pending?              (pos? requested-to-join-at)
+        thumbnail-image       (get-in images [:thumbnail])
+        chats-by-category     (rf/sub [:communities/categorized-channels id])
+        users                 (rf/sub [:communities/users id])
+        join-gates            (rf/sub [:community/join-gates id])
+        permission-tag-tokens (rf/sub [:community/permission-tag-tokens id])]
     [rn/view
      [rn/view {:padding-horizontal 20}
       (when (and (not joined)
@@ -230,7 +232,7 @@
          [quo/permission-tag-container
           {:locked   locked
            :status   status
-           :tokens   tokens
+           :tokens   permission-tag-tokens
            :on-press #(rf/dispatch
                        [:bottom-sheet/show-sheet
                         {:content-height 210
@@ -239,7 +241,7 @@
                            [community-token-gating-details
                             name
                             thumbnail-image
-                            tokens])}])}]])
+                            join-gates])}])}]])
       (when (or pending? joined)
         [rn/view
          {:position :absolute
--- a/src/status_im2/subs/communities.cljs
+++ b/src/status_im2/subs/communities.cljs
@@ -216,6 +216,34 @@
         (sort-by :position)
         (into []))))
 
+
+(def token-images
+  {"KNC"  (js/require "../resources/images/tokens/mainnet/KNC.png")
+   "MANA" (js/require "../resources/images/tokens/mainnet/MANA.png")
+   "RARE" (js/require "../resources/images/tokens/mainnet/RARE.png")
+   "ETH"  (js/require "../resources/images/tokens/mainnet/ETH.png")
+   "DAI"  (js/require "../resources/images/tokens/mainnet/DAI.png")})
+
+(defn token-image
+  [token]
+  (get token-images token))
+
+(defn enrich-gate-for-ui
+  [{:keys [token] :as gate}]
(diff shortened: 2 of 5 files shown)


vkjr:
@flexsurfer, @ilmotta, I'd like to ask your opinion on these changes.
In a previous PR related to mock data we had a discussion and I [proposed](https://github.com/status-im/status-mobile/pull/14799#issuecomment-1385309316) to add mock data to a single community after it is fetched.
This PR does exactly this - adds gating to the default Status community and one of its channels. Also, it refines the path gating data have into UI, so a couple of useful subs added. 
Having mocks in `status-mobile` is still not super-pretty, so I currently investigate how it can be done on the `status-go`. 
Meanwhile, wdyt about having these changes in `status-mobile` code? I think it can be useful because shows how gating data can look in our re-frame db and how it gets to ui.

p.s. will add tests for subs if we decide that PR worth merging)

src/status_im2/subs/communities.cljs
```diff
@@ -216,14 +216,42 @@
         (sort-by :position)
         (into []))))
 
+
+(def token-images
```
flexsurfer, line 220: this shouldn't be here, all js/require must be in resources

src/status_im2/subs/communities.cljs
```diff
@@ -216,14 +216,42 @@
...
+
+(defn enrich-gates-vector-for-ui
+  [gates]
+  (vec (map enrich-gate-for-ui gates)))
+
+(defn enrich-gates-for-ui
+  [gates]
+  (vec (map (fn [v]
+              (if (vector? v)
+                (enrich-gates-vector-for-ui v)
```
flexsurfer, line 243: im not sure what the point of having 3 one line functions , it would be much simple and more readable if we have all code in one place

src/status_im2/subs/communities.cljs
```diff
@@ -235,6 +263,11 @@
                 #(vec (conj %1 %2))
                 {:name             name
                  :emoji            emoji
+                 :gates            (merge
+                                    (when (contains? gates :read)
```
flexsurfer, line 267: it could be just 
```clojure
(-> gate
      (update :read enrich-gates-for-ui)
      (update :write enrich-gates-for-ui))
```
but im not sure why we're doing this in subscription, can this be done when we fetch gates ?

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -210,14 +210,16 @@
...
+           status tags requested-to-join-at id]
     :as   community}
    {:keys [on-categories-heights-changed
            on-first-channel-height-changed]}]
-  (let [pending?          (pos? requested-to-join-at)
-        thumbnail-image   (get-in images [:thumbnail])
-        chats-by-category (rf/sub [:communities/categorized-channels id])
-        users             (rf/sub [:communities/users id])]
+  (let [pending?              (pos? requested-to-join-at)
+        thumbnail-image       (get-in images [:thumbnail])
```
flexsurfer, line 218: (:thumbnail images)

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -239,7 +241,7 @@
                            [community-token-gating-details
                             name
                             thumbnail-image
```
flexsurfer, line 243: actually its better here (:thumbnail images), not need in extra let

src/status_im2/contexts/communities/overview/view.cljs
```diff
@@ -31,7 +31,7 @@
     (utils/join-existing-users-string user-list)]])
 
 (defn channel-token-gating-details
-  [name token-gating emoji channel-color]
+  [name gates emoji channel-color]
```
flexsurfer, line 34: could you elaborate please on renaming ?

flexsurfer:
> Having mocks in `status-mobile` is still not super-pretty, so I currently investigate how it can be done on the `status-go`.

sorry, i still don't understand why we're doing this , i don't see any value

src/status_im2/subs/communities.cljs
vkjr: At the moment of writing the first subscription, I wasn't sure if we will need the gating data in different formats derived from the basic one. Now it looks like no, so you are right, that can be done when we fetch gates

src/status_im2/subs/communities.cljs
vkjr: That probably about personal preferences) For me, it is easier to grasp a few short-named functions than few-layer structure.

src/status_im2/contexts/communities/overview/view.cljs
vkjr: Hm.. at the moment of renaming I had some internal reasoning.. but can't remember it)) So definetely can be rolled back)

src/status_im2/subs/communities.cljs
flexsurfer: im not sure its personal, its about speed of reading and understanding the code, to read and understand the code you have to go in 3 different places, also namespace is getting noisy

ilmotta:
> @flexsurfer, @ilmotta, I'd like to ask your opinion on these changes. In a previous PR related to mock data we had a discussion and I [proposed](https://github.com/status-im/status-mobile/pull/14799#issuecomment-1385309316) to add mock data to a single community after it is fetched. This PR does exactly this - adds gating to the default Status community and one of its channels. Also, it refines the path gating data have into UI, so a couple of useful subs added. Having mocks in `status-mobile` is still not super-pretty, so I currently investigate how it can be done on the `status-go`. Meanwhile, wdyt about having these changes in `status-mobile` code? I think it can be useful because shows how gating data can look in our re-frame db and how it gets to ui.
> 
> p.s. will add tests for subs if we decide that PR worth merging)

Hey @vkjr, thanks for asking and attempting to work on this.

I understand the goal of this PR, but tbh I think the root cause of the problem is not yet addressed. IMO the mobile client should avoid implementing pure UI screens not backed by real data and real implementation on the backend. This is a strategic mistake in my experience. This gives a false sense of progress. It's also risky because the client ends up assuming how the server will behave in the future without a feedback loop between the two. The risk is having to rework on the wrong assumptions (time & $ cost). The inverse is also true, if the server evolves independently of any real client usage, the server needs to guess future client requirements.

The mock strategy is sound IMO, it's just that it should be the exception, not the norm. So I get the need for this PR and **I actually think it's fine**, but I think it's about time we address the root cause and stop actually relying on mock data and creating fake screens, even if it's just for development. I could go on and on with other points :man_shrugging: 

But going back to your points :)

>Having mocks in status-mobile is still not super-pretty, so I currently investigate how it can be done on the status-go.

I prefer if status-go doesn't implement stuff that only one client needs, especially mock data. @cammellos would be a better person to guide us here.

vkjr:
@ilmotta, @flexsurfer, I see your points and agree with them. Probably I became too engaged in all these mock data things)
So, as in fact we don't need to deal with mocks at all, I suggest closing #14462 as a surplus and just skipping this PR. And I will switch to something more beneficial for the product)

ilmotta:
> @ilmotta, @flexsurfer, I see your points and agree with them. Probably I became too engaged in all these mock data things) So, as in fact we don't need to deal with mocks at all, I suggest closing #14462 as a surplus and just skipping this PR. And I will switch to something more beneficial for the product)

Cool @vkjr, good idea on closing that other issue. Thanks for raising the PR too, it helped clarify some things for me at least.

J-Son89:
> > @ilmotta, @flexsurfer, I see your points and agree with them. Probably I became too engaged in all these mock data things) So, as in fact we don't need to deal with mocks at all, I suggest closing #14462 as a surplus and just skipping this PR. And I will switch to something more beneficial for the product)
> 
> 
> 
> Cool @vkjr, good idea on closing that other issue. Thanks for raising the PR too, it helped clarify some things for me at least.

Yeah I believe some/most of this started from my work. I was not aware we can embed real communities into the application and so going forward we can develop from real data given that desktop features (might be wrong about this) are generally going to be implemented before mobile. 
In the case of token gating it's not ready yet on status go so yeah there's no point of us investing any dev effort into it on mobile until it's good to go.
Glad to see the discussion above and cheers for getting it sorted 👍

PR #15105, opened 2023-02-16
feat: quo2 color-picker component

### LINKED ISSUE #15021: Implement quo2 component - Colors/ Color Picker
Designs:
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System-for-Mobile?node-id=2972%3A43112&t=29WaMjPTYMtHqa1n-0

Ensure list of colors matches set in designs:

Component tests 
-> include color set in tests to keep these colors valid.
Preview screens

ajayesivan:
fixes #15021

Test via the quo2 component preview: colors -> color-picker

### Preview video

https://user-images.githubusercontent.com/19279756/219462701-530fc353-e4d5-4f79-96b9-983d38965617.mov

status: ready

### DIFF
--- a/src/quo2/components/colors/color_picker/component_spec.cljs
+++ b/src/quo2/components/colors/color_picker/component_spec.cljs
@@ -1,0 +1,23 @@
+(ns quo2.components.colors.color-picker.component-spec
+  (:require [quo2.components.colors.color-picker.view :as color-picker]
+            [reagent.core :as reagent]
+            [test-helpers.component :as h]))
+
+(h/describe "color-picker"
+  (h/test "color picker rendered"
+    (h/render [color-picker/view])
+    (-> (h/expect (h/get-all-by-label-text :color-picker-item))
+        (.toHaveLength 12)))
+  (h/test "clicks on a color item"
+    (let [event (h/mock-fn)]
+      (h/render [color-picker/view {:on-change #(event)}])
+      (h/fire-event :press (get (h/get-all-by-label-text :color-picker-item) 0))
+      (-> (h/expect event)
+          (.toHaveBeenCalled))))
+  (h/test "color picker color changed"
+    (let [selected (reagent/atom nil)]
+      (h/render [color-picker/view {:on-change #(reset! selected %)}])
+      (h/fire-event :press (get (h/get-all-by-label-text :color-picker-item) 0))
+      (-> (h/expect @selected)
+          (.toStrictEqual :blue)))))
+
--- a/src/quo2/components/colors/color_picker/style.cljs
+++ b/src/quo2/components/colors/color_picker/style.cljs
@@ -1,0 +1,48 @@
+(ns quo2.components.colors.color-picker.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def color-picker-container
+  {:max-width       338
+   :flex-direction  :row
+   :flex-wrap       :wrap
+   :justify-content :space-between})
+
+(def flex-break
+  {:flex-basis "100%"
+   :height     10})
+
+(defn color-button
+  [color selected?]
+  (merge {:width         48
+          :height        48
+          :border-width  4
+          :border-radius 24
+          :transform     [{:rotate "45deg"}]
+          :border-color  :transparent}
+         (when selected?
+           {:border-top-color    (colors/alpha color 0.4)
+            :border-end-color    (colors/alpha color 0.4)
+            :border-bottom-color (colors/alpha color 0.2)
+            :border-start-color  (colors/alpha color 0.2)})))
+
+(defn color-circle
+  [color border?]
+  {:width            40
+   :height           40
+   :transform        [{:rotate "-45deg"}]
+   :background-color color
+   :justify-content  :center
+   :align-items      :center
+   :border-color     color
+   :border-width     (if border? 2 0)
+   :overflow         :hidden
+   :border-radius    20})
+
+(defn secondary-overlay
+  [secondary-color border?]
+  {:width            (if border? 18 20)
+   :height           40
+   :position         :absolute
+   :right            0
+   :background-color secondary-color})
+
--- a/src/quo2/components/colors/color_picker/view.cljs
+++ b/src/quo2/components/colors/color_picker/view.cljs
@@ -1,0 +1,71 @@
+(ns quo2.components.colors.color-picker.view
+  (:require [react-native.core :as rn]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.icon :as icon]
+            [reagent.core :as reagent]
+            [quo2.components.colors.color-picker.style :as style]))
+
+(def color-list [:blue :yellow :turquoise :copper :sky :camel :orange :army :pink :purple :magenta])
+
+(defn picker-colors
+  [blur?]
+  (concat (map (fn [color]
+                 {:name  color
+                  :color (colors/custom-color-by-theme color (if blur? 60 50) 60)})
+               color-list)
+          [{:name            :yinyang
+            :color           (colors/theme-colors (colors/custom-color :yin 50)
+                                                  (colors/custom-color :yang 50))
+            :secondary-color (colors/theme-colors (colors/custom-color :yang 50)
+                                                  (colors/custom-color :yin 50))}]))
+
+(defn- on-change-handler
+  [selected color-name on-change]
+  (reset! selected color-name)
+  (when on-change (on-change color-name)))
+
+(defn- color-item
+  [{:keys [name
+           color
+           secondary-color
+           selected?
+           on-press
+           blur?]}]
+  (let [border? (and (not blur?) (and secondary-color (not selected?)))]
+    [rn/touchable-opacity
+     {:style               (style/color-button color selected?)
+      :accessibility-label :color-picker-item
+      :on-press            #(on-press name)}
+     [rn/view
+      {:style (style/color-circle color border?)}
+      (when (and secondary-color (not selected?))
+        [rn/view
+         {:style (style/secondary-overlay secondary-color border?)}])
+      (when selected?
+        [icon/icon :i/check
+         {:size  20
+          :color (or secondary-color
+                     colors/white)}])]]))
+
+(defn view
+  "Options
+   - `default-selected?` Default selected color name.
+   - `selected` Selected color name.
+   - `on-change` Callback called when a color is selected `(fn [color-name])`.
+   - `blur?` Boolean to enable blur background support.}"
+  [{:keys [default-selected?]}]
+  (let [internal-selected (reagent/atom default-selected?)]
+    (fn [{:keys [blur? on-change selected]}]
+      (when (and (not (nil? selected)) (not= @internal-selected selected))
+        (reset! internal-selected selected))
+      [rn/view {:style style/color-picker-container}
+       (doall (map-indexed (fn [index color]
+                             [:<> {:key (color :name)}
+                              [color-item
+                               (merge color
+                                      {:selected? (= (color :name) @internal-selected)
+                                       :on-press  #(on-change-handler internal-selected % on-change)
+                                       :blur?     blur?})]
+                              (when (= index 5) [rn/view {:style style/flex-break}])])
+                           (picker-colors blur?)))])))
+
--- a/src/quo2/foundations/colors.cljs
+++ b/src/quo2/foundations/colors.cljs
@@ -170,26 +170,40 @@
 (def customization
   {:primary   {50 primary-50 ;; User can also use primary color as customisation color
                60 primary-60}
-   :purple    {50 "#8661C1"
-               60 "#5E478C"}
+   :purple    {50 "#7140FD"
+               60 "#5A33CA"}
    :indigo    {50 "#496289"
                60 "#3D5273"}
-   :turquoise {50 "#448EA2"
-               60 "#397788"}
-   :blue      {50 "#4CB4EF"
-               60 "#4097C9"}
+   :turquoise {50 "#2A799B"
+               60 "#22617C"}
+   :blue      {50 "#2A4AF5"
+               60 "#223BC4"}
    :green     {50 "#5BCC95"
                60 "#4CAB7D"}
-   :yellow    {50 "#FFCB53"
-               60 "#D6AA46"}
-   :orange    {50 "#FB8F61"
-               60 "#D37851"}
+   :yellow    {50 "#F6B03C"
+               60 "#C58D30"}
+   :orange    {50 "#FF7D46"
+               60 "#CC6438"}
    :red       {50 "#F46666"
                60 "#CD5656"}
-   :pink      {50 "#FC7BAB"
-               60 "#D46790"}
+   :pink      {50 "#F66F8F"
+               60 "#C55972"}
    :brown     {50 "#99604D"
                60 "#805141"}
+   :sky       {50 "#1992D7"
+               60 "#1475AC"}
+   :army      {50 "#216266"
+               60 "#1A4E52"}
+   :magenta   {50 "#EC266C"
+               60 "#BD1E56"}
+   :copper    {50 "#CB6256"
+               60 "#A24E45"}
+   :camel     {50 "#C78F67"
+               60 "#9F7252"}
+   :yin       {50 "#09101C"
+               60 "#1D232E"}
+   :yang      {50 "#FFFFFF"
+               60 "#EBEBEB"}
    :beige     {50 "#CAAE93"
                60 "#AA927C"}})
 
--- a/src/quo2/core.cljs
+++ b/src/quo2/core.cljs
@@ -10,6 +10,7 @@
     quo2.components.banners.banner.view
     quo2.components.buttons.button
     quo2.components.buttons.dynamic-button
+    quo2.components.colors.color-picker.view
     quo2.components.community.community-card-view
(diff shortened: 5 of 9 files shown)


J-Son89:
Great work @ajayesivan !! 💥

src/quo2/components/colors/color_picker/component_spec.cljs
```diff
@@ -0,0 +1,22 @@
...
+
+(h/describe "color-picker"
+  (h/test "color picker rendered"
+    (h/render [color-picker/view])
+    (-> (h/expect (h/get-by-test-id (str "quo2-color-picker-sky")))
+        (.toBeTruthy)))
+  (h/test "clicks on a color item"
+    (let [event (h/mock-fn)]
+      (h/render [color-picker/view {:on-change #(event)}])
+      (h/fire-event :press (h/get-by-test-id "quo2-color-picker-copper"))
```
J-Son89: nit: ideally you can use an accessibility label instead? and the method `get-by-label-text`

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,108 @@
+(ns quo2.components.colors.color-picker.view
+  (:require [react-native.core :as rn]
+            [quo2.foundations.colors :as colors]
+            [quo2.components.icon :as icon]
+            [reagent.core :as reagent]
+            [quo2.components.colors.color-picker.style :as style]))
+
+(def picker-colors
+  [{:name       "blue"
```
J-Son89: I think these names should keywords. e.g `:blue`

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+    :color      "#2A4AF5"
```
Parveshdhull: Please don't hard colors, use them from [colors file](https://github.com/status-im/status-mobile/blob/cbad7f4c87d3bfb555d2a4de3a5c51186d91ab3e/src/quo2/foundations/colors.cljs#L170) using function [custom-color-by-theme](https://github.com/status-im/status-mobile/blob/cbad7f4c87d3bfb555d2a4de3a5c51186d91ab3e/src/quo2/foundations/colors.cljs#L216).
(Looks like some colors are new, please add them)

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+      (when (and secondary-color (not selected?))
+        [rn/view
+         {:style (style/secondary-overlay theme-secondary border?)}])
+      (when selected?
+        [icon/icon :i/check
+         {:size  20
+          :color (or theme-secondary
+                     colors/white)}])]]))
+
+(defn- flex-break
```
Parveshdhull: why we need a separate function here?

Parveshdhull (commented):
Hi @ajayesivan, awesome work.

For selected color, right side should be with opacity 40 and left should be with 20

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+   {:name       "camel"
+    :color      "#C78F67"
+    :color-dark "#9F7252"}
+   {:name       "orange"
+    :color      "#FF7D46"
+    :color-dark "#CC6438"}
+   {:name       "army"
+    :color      "#216266"
+    :color-dark "#1A4E52"}
+   {:name       "flamingo"
```
Parveshdhull: pink?

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,108 @@
...
+    :color      "#F66F8F"
+    :color-dark "#C55972"}
+   {:name       "purple"
+    :color      "#7140FD"
+    :color-dark "#5A33CA"}
+   {:name       "magenta"
+    :color      "#EC266C"
+    :color-dark "#BD1E56"}
+   {:name                 "fengshui"
```
Parveshdhull, line 42: fengchui?

src/quo2/components/colors/color_picker/view.cljs
Parveshdhull: no idea why it says flamingo, but in foundation file color is named as pink
image

src/quo2/components/colors/color_picker/view.cljs
briansztamfater: Also, for the sake of consistency I would prefer naming keys as `:color-light` and `:color-dark`.

briansztamfater: Also we should add one test for each color checking the hex or rgb values, as this is a sensible component that will affect many others and we need to be sure we don't break color sets in the future.

Parveshdhull: > Also, for the sake of consistency I would prefer naming keys as `:color-light` and `:color-dark`.

Once we use custom-color-by-theme function, we don't need to store these values.
`(custom-color-by-theme :blue 50 60)`

Parveshdhull: > Also we should add one test for each color checking the hex or rgb values, as this is a sensible component that will affect many others and we need to be sure we don't break color sets in the future.

yes this sounds good

Parveshdhull: If we are just going to compare values from foundations file, we probably don't need component test for that. We can just use `cljs.test`.

briansztamfater: > > Also, for the sake of consistency I would prefer naming keys as `:color-light` and `:color-dark`.
> 
> Once we use custom-color-by-theme function, we don't need to store these values. `(custom-color-by-theme :blue 50 60)`

Yes, if we can use just one key and use that function from the `quo2.foundations.colors` namespace, would be better

src/quo2/components/colors/color_picker/component_spec.cljs
ajayesivan: Based on [new guidelines](https://github.com/status-im/status-mobile/blob/develop/doc/new-guidelines.md#accessibility-labels), it seems that accessibility labels are only used for e2e tests and dynamic labels should be avoided. Also, I've noticed that some other component tests are using test-IDs, which is why I opted to use them in this case. Should I switch to accessibility labels instead and is there any specific reason for this?

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: Initially, I had thought to keep the break component separate, but it seems unnecessary to have a separate component for this. I will change it.

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: @Parveshdhull In the new design, I noticed that color values are different compared to our existing customization colors. For example, the existing color code for blue 50 is `#4CB4EF`, while the new design uses `#2A4AF5`. Would it be acceptable to update the existing color codes with the new values?

Parveshdhull: Yes please update them (Note: please check actual color value from inspect, label are wrong sometime)
image

src/quo2/components/colors/color_picker/component_spec.cljs
J-Son89: Yes, some tests were implemented before we realised that the guidance from react-native-testing-library is to prefer accessibility labels over test I.ds where possible.

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: @Parveshdhull Could you please share the link to the design file with the color values? So that I can update all colors to the new values.

Parveshdhull: https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Foundations?node-id=619%3A10064&t=dSqqKUVodgbr84nw-0

ajayesivan:
> Hi @ajayesivan, awesome work.
> 
> For selected color, right side should be with opacity 40 and left should be with 20

Fixed

ajayesivan:
@J-Son89 @Parveshdhull @briansztamfater I have updated the PR with the requested changes.

src/quo2/components/colors/color_picker/style.cljs
```diff
@@ -0,0 +1,47 @@
...
+   :overflow         :hidden
+   :border-radius    20})
+
+(defn secondary-overlay
+  [secondary-color border?]
+  {:width            (if border? 18 20)
+   :height           40
+   :position         :absolute
+   :right            0
+   :background-color secondary-color})
```
J-Son89: can you add a newline to the end of files please. It's part of guidelines afaik :) you can adjust your editor to autosave and adjust if it helps too!

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,69 @@
...
+            [quo2.components.icon :as icon]
+            [reagent.core :as reagent]
+            [quo2.components.colors.color-picker.style :as style]))
+
+(defn picker-colors
+  [blur?]
+  (concat (map (fn [color]
+                 {:name  color
+                  :color (colors/custom-color-by-theme color (if blur? 60 50) 60)})
+               [:blue :yellow :turquoise :copper :sky :camel :orange :army :pink :purple :magenta])
```
J-Son89: nit: separate color list into it's own def

```
(def colors-list [:blue :yellow :turquoise :copper :sky :camel :orange :army :pink :purple :magenta])

(defn picker-colors
  [blur?]
  (concat (map (fn [color]
                 {:name  color
                  :color (colors/custom-color-by-theme color (if blur? 60 50) 60)})
              colors-list )
          [{:name            :fengshui
            :color           (colors/theme-colors (colors/custom-color :feng 50)
                                                  (colors/custom-color :chui 50))
            :secondary-color (colors/theme-colors (colors/custom-color :chui 50)
                                                  (colors/custom-color :feng 50))}]))
```

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,69 @@
...
+          [{:name            :fengshui
```
J-Son89: @Parveshdhull, should fengshui not just be added directly to the custom-color-by-theme function??

I thought that list is just the custom colors selected by design team so this should be alongside it 👍

J-Son89 (approved):
LGTM, a few small comments but other than that all good :)

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: From my understand, `feng` and `chui` are two separate colors(see the [customization colors](https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Foundations?node-id=619%3A9895&t=J7cjdtt9EzPQIs5C-4)). The `custom-color-by-theme` is implemented in such a way that it can get variations of one color.

Here we are using the combination of these two colors to produce a special picker element so that it will look good on both light & dark themes.

Screenshot 2023-02-17 at 19 54 29

src/quo2/components/colors/color_picker/style.cljs
```diff
@@ -0,0 +1,47 @@
+(ns quo2.components.colors.color-picker.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def color-picker-container
+  {:max-width       338
```
Parveshdhull, line 5: Please don't hard code container width
colors width, height 48, margin between colors 19, so wouldn't container width will be fixed?

src/quo2/components/colors/color_picker/view.cljs
J-Son89: Yea but as I understand it the function is designed to handle the custom Color's for users so we should adjust it to also handle two colors.
Let's get feedback from @Parveshdhull as initially implemented and knows it's use case

Parveshdhull: yup we can't add this color, because its a component

how feng+chui = fengshui? (don't understand what they mean, just curious)

src/quo2/components/colors/color_picker/style.cljs
```diff
@@ -0,0 +1,48 @@
...
+   :flex-direction  :row
+   :flex-wrap       :wrap
+   :justify-content :space-between})
+
+(def flex-break
+  {:flex-basis "100%"
```
flexsurfer, line 11: could you pls elaborate on why :flex-basis "100%" is needed?

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,72 @@
...
+  [{:keys [name
+           color
+           secondary-color
+           selected?
+           on-press
+           blur?]}]
+  (let [border? (and (not blur?) (and secondary-color (not selected?)))]
+    [rn/touchable-opacity
+     {:style               (style/color-button color selected?)
+      :accessibility-label (str "quo2-color-picker-" name)
```
flexsurfer: it should be just `:color-picker-item`

src/quo2/components/colors/color_picker/view.cljs
```diff
@@ -0,0 +1,72 @@
...
+      :on-press            #(on-press name)}
+     [rn/view
+      {:style               (style/color-circle color border?)
+       :accessibility-label (str "quo2-color-picker-item" name)}
```
flexsurfer: no need in this , only one needed for the component

flexsurfer (commented):
great job!

src/quo2/components/colors/color_picker/style.cljs
ajayesivan: I'm using fixed container width & `:justify-content :space-between` to space the components. I thought it was better than using margin in this case. Any specific reason for avoiding this approach in this scenario?

Parveshdhull: That's a really good question. I am not sure which is recommended way, let's confirm with the design team.

Parveshdhull: @pedro-et, Whole color-picker container has width 335 (40 x 6+19 x 5), where 40 is color item's size and 19 is margin between them.
But lets say we have a wide android device. For that should we just hard code these margins, or spread these color items around as per available gap `(:justify-content :space-between)`?

pedro-et: if its slightly larger just keep the side margins and space the elements evenly. If its significantly larger maybe just had an extra column. @Parveshdhull

src/quo2/components/colors/color_picker/style.cljs
ajayesivan: My initial approach was to implement rows using a dedicated row component([see this commit](https://github.com/ajayesivan/status-mobile/commit/4962d7deefdb4ba9d5f5ec6ed4399b647173d523)). But I was not satisfied with that. Then I found another approach in this article: https://tobiasahlin.com/blog/flexbox-break-to-new-row/.

Here, I'm using a full-width(flex-basis 100% for the width) component in the middle to break a single flex row into two rows.

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: Are you suggesting not to use dynamic labels or to just change the 'quo2-color-picker-' part to `:color-picker-item'?

Could you please have a look at the component tests, I'm using dynamic labels there.

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: Done

src/quo2/components/colors/color_picker/view.cljs
ajayesivan: Done

src/quo2/components/colors/color_picker/style.cljs
ajayesivan: Done

src/quo2/components/colors/color_picker/view.cljs
flexsurfer: yes we shouldn't use dynamic labels

ajayesivan: Done. Using static accessibility-label.

src/quo2/foundations/colors.cljs
```diff
@@ -170,26 +170,40 @@
...
+   :magenta   {50 "#EC266C"
+               60 "#BD1E56"}
+   :copper    {50 "#CB6256"
+               60 "#A24E45"}
+   :camel     {50 "#C78F67"
+               60 "#9F7252"}
+   :feng      {50 "#09101C"
+               60 "#1D232E"}
+   :chui      {50 "#FFFFFF"
+               60 "#EBEBEB"}
```
Parveshdhull: :feng, :chui are renamed to :yin, :yang :yin_yang: in [colors file ](https://www.figma.com/file/v98g9ZiaSHYUdKWrbFg9eM/Foundations?node-id=619%3A10064&t=Mv9KUEwGj294AFxG-0) (please update)

Parveshdhull (approved):

src/quo2/foundations/colors.cljs
ajayesivan: Done

Parveshdhull:
PR lgtm, should we merge?

PR #15122, opened 2023-02-17
Fix for empty display-name

vitvly:
Fixes #15120

Issue #15120, opened 2023-02-17
3 random name is lost after relogin (release blocker)


pavloburykh:
Probably related to https://github.com/status-im/status-mobile/pull/15083 cc @siphiuel @cammellos 

**Steps:**

1. Create/restore multiaccount
2. Open Profile tab, make sure that user's 3 random name is shown
3. Perform relogin (manually tap Sign out and then login again)
6. See if 3 random name is still displayed on profile screen 

**Expected result:** 3 random name is not lost. It is possible to mention user by 3 random name both from mobile and desktop.

**Actual result:** 3 random name is lost. After that it is impossible to mention such user from desktop.

https://user-images.githubusercontent.com/97245802/219677446-6a7d3797-ca40-4b05-9a98-12afb9f72707.mp4

 
[Status-debug-logs (2).zip](https://github.com/status-im/status-mobile/files/10767601/Status-debug-logs.2.zip)

### Additional Information

- Status version: release 1. 
- Operating System: Android, iOS

[comment]: # (Please, add logs/notes if necessary)

vitvly:
Initially i thought it might be due to https://github.com/status-im/status-go/commit/2f41034b47d67eacc9002043b451962a45dd93ba not getting included, but it's there in the `v0.131.7+hotfix.1` tag. Debugging.

cammellos:
```
t=2023-02-17T13:58:56+0000 lvl=dbug msg="Handling Backup" site=RetrieveAll hash=0x3b0dd7ecf53c3e744db25fa43489646d4865ad4ca66c71145ee56836f7b263d6 message-id=0xbf944352f7e5a5e688783dccfc0b791d643777c6dbd00627191356ac942e94e3 message="{Clock:1676042107182 Id: Contacts:[] Communities:[] ContactsDetails:total_number:1  CommunitiesDetails:total_number:4  Profile:<nil> ProfileDetails:total_number:1  Setting:type:DISPLAY_NAME clock:1675959864104 value_string:\"\"  SettingsDetails:data_number:2 total_number:5  XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}"
```
@siphiuel , looks like it might be backup issue https://github.com/status-im/status-go/blob/1d1a95091df0197199ea502aae24c823faf9b989/protocol/messenger_backup_handler.go#L77 we don't check for empty values

cammellos:
If that's the issue, this is also a fix that desktop will want to include

[...]

PR #15125, opened 2023-02-17
Set display-name during settings creation

vitvly:
This way we can avoid always invoking (although memoized) `generate-gfycat` function in `displayed-name` fn https://github.com/status-im/status-mobile/blob/900688dc159b2040f66bd58fed1f1828fbb47215/src/status_im/multiaccounts/core.cljs#L78 in cases when no custom `display-name` has been set by the user.

This makes https://github.com/status-im/status-mobile/pull/15122 unnecessary.

Related to https://github.com/status-im/status-go/issues/3205 and https://github.com/status-im/status-mobile/issues/15120

### DIFF
--- a/src/status_im/multiaccounts/create/core.cljs
+++ b/src/status_im/multiaccounts/create/core.cljs
@@ -202,6 +202,7 @@
                      constants/path-wallet-root-keyword
                      :address])
             :name name
+            :display-name name
             :identicon identicon
             ;; public key of the chat account
             :public-key public-key


cammellos:
@siphiuel I think it's best to keep it empty, so we know the user didn't set it explicitly, otherwise if we want to migrate, we need to check `name` == `3 random words name`, which is a bit more annoying

vitvly:
Agree, ok!

PR #15126, opened 2023-02-17
Add display name/ens to activity center

cammellos:
We were only checking nickname/three-random words, added support for ens-name/display-name in the correct order, probably this should be pulled by a subscription and we should rationalize the logic and keep it in a single place (maybe `contacts` namespace? )
But this is an include for RC, so best to do in a separate PR

status: ready

### DIFF
--- a/src/status_im2/contexts/activity_center/utils.cljs
+++ b/src/status_im2/contexts/activity_center/utils.cljs
@@ -2,5 +2,10 @@
 
 (defn contact-name
   [contact]
-  (or (get-in contact [:names :nickname])
-      (get-in contact [:names :three-words-name])))
+  (->> [(get-in contact [:names :nickname])
+        (get-in contact [:names :ens-name])
+        (get-in contact [:names :display-name])
+        (get-in contact [:names :three-words-name])]
+
+       (filter seq)
+       first))


src/status_im2/contexts/activity_center/utils.cljs
```diff
@@ -2,5 +2,8 @@
 
 (defn contact-name
   [contact]
-  (or (get-in contact [:names :nickname])
-      (get-in contact [:names :three-words-name])))
+  (or
+   (get-in contact [:names :ens-name])
+   (get-in contact [:names :display-name])
+   (get-in contact [:names :nickname])
```
Parveshdhull: Is display-name has high priority then nickname?

Parveshdhull (approved):

src/status_im2/contexts/activity_center/utils.cljs
ilmotta: Same question, I always thought the user defined `nickname` had the highest priority.

ilmotta: Or in other words, if I have my mom as a mutual contact, and I set her nickname to be `mom`, then I don't care if she has an ens name. She's `mom` after all!

cammellos: ouch! fixing  :dagger:

cammellos: done, thanks for the feedback!

ilmotta (approved):

qoqobolo:
Hi @cammellos, thanks for PR!

Currently, display names are displayed in notifications for contact reqeusts in AC, but 3 random and ENS names are not.

Here is a screenshot with 3 contact requests: 
- display name (from desktop)
- 3 random (from mobile)
- ENS (from mobile; although maybe ENS names are not supposed to be resolved in contact requests? let me know if that's the expected behavior)


Screenshot 2023-02-20 at 11 15 14

As I can see, ENS names work fine for mention notifications after the CR has been approved.
However, 3 random names are still not displayed even for mutual contacts:

Screenshot 2023-02-20 at 11 21 08

---
Also, [the issue with names lost after relogin in profile](https://github.com/status-im/status-mobile/issues/15120) is still reproducible in this PR. Could you rebase it please to include the fix?

cammellos:
@qoqobolo rebased, would you mind trying the new build please?

qoqobolo:
LGTM now, thanks @cammellos!

One question just to make sure: ENS names are only resolved after becoming mutual contacts **and** sending a message in any chat. This is expected at the moment, right? I'm just not sure it is possible to get ENS name until receiving a message from the ENS contact.

Also, there is no way atm to test the case of receiving a notification from the desktop user with ENS. Afaik, the Mainnet network is set on desktop by default, while in our PR builds it is Goerli. So skipping this case for now.

cammellos:
> One question just to make sure: ENS names are only resolved after becoming mutual contacts and sending a message in any chat. This is expected at the moment, right? I'm just not sure it is possible to get ENS name until receiving a message from the ENS contact.

I believe that's the behavior currently, but it doesn't have to be, could you create please an issue tracking this? 
we could display the ens name as soon as you become a mutual contact 

Thank you

qoqobolo:
> I believe that's the behavior currently, but it doesn't have to be, could you create please an issue tracking this?
we could display the ens name as soon as you become a mutual contact

Got it, thanks. Will log an issue.

PR is ready to be merged and cherry-picked.
Thanks for the fix @cammellos!

cammellos:
@qoqobolo cherry picked on release and merged in develop, thanks for testing!

PR #15150, opened 2023-02-21
Use nil instead of empty string for display-name

vitvly:
Settings json now uses omitempty for displayName field. This helps to obsolete https://github.com/status-im/status-mobile/pull/15122. Steps outlined in https://github.com/status-im/status-mobile/issues/15120 should work.

Depends on https://github.com/status-im/status-go/pull/3215


qfrank (approved):

flexsurfer:
display-name is not used anymore

PR #15157, opened 2023-02-21
quo2 password-tips component

### LINKED ISSUE #15027: Implement Quo2 component - Password/Tips
Designs:
https://www.figma.com/file/WQZcp6S0EnzxdTL4taoKDv/Design-System-for-Mobile?node-id=14927%3A187285&t=29WaMjPTYMtHqa1n-0

Component tests

Preview screens

ajayesivan:
fixes #15027 

Test component from quo2 preview -> password -> tips

#### iOS
ios-password-tips
ios-password-tips-completed

#### Android
android-password-tips
android-password-tips-completed

status: ready

### DIFF
--- a/src/quo2/components/password/tips/style.cljs
+++ b/src/quo2/components/password/tips/style.cljs
@@ -1,0 +1,18 @@
+(ns quo2.components.password.tips.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def container
+  {:align-self :flex-start})
+
+(defn tip-text
+  [completed?]
+  {:color (if completed? colors/white-opa-40 colors/white-opa-70)})
+
+(def strike-through
+  {:position         :absolute
+   :background-color colors/white
+   :top              10
+   :left             0
+   :right            0
+   :height           0.8})
+
--- a/src/quo2/components/password/tips/view.cljs
+++ b/src/quo2/components/password/tips/view.cljs
@@ -1,0 +1,24 @@
+(ns quo2.components.password.tips.view
+  (:require [react-native.core :as rn]
+            [quo2.components.password.tips.style :as style]
+            [quo2.components.markdown.text :as text]))
+
+(defn view
+  "Options
+   - `completed?` Password tip state
+   
+   `text` Password tip string
+   "
+  [{:keys [completed?]} text]
+  [rn/view
+   {:style               style/container
+    :accessibility-label :password-tips}
+   [text/text
+    {:style  (style/tip-text completed?)
+     :weight :regular
+     :size   :paragraph-2} text]
+   (when completed?
+     [rn/view
+      {:style               style/strike-through
+       :accessibility-label :password-tips-completed}])])
+
--- a/src/quo2/components/password/tips/component_spec.cljs
+++ b/src/quo2/components/password/tips/component_spec.cljs
@@ -1,0 +1,20 @@
+(ns quo2.components.password.tips.component-spec
+  (:require [quo2.components.password.tips.view :as quo]
+            [test-helpers.component :as h]))
+
+(h/describe "password tips component"
+  (h/test "render component"
+    (h/render [quo/view])
+    (-> (h/expect (h/get-by-label-text :password-tips))
+        (.toBeTruthy)))
+  (h/test "render component with proper text"
+    (h/render [quo/view {:completed? false} "Upper case"])
+    (-> (h/expect (h/get-by-text "Upper case"))
+        (.toBeTruthy)))
+  (h/test "render component with completed proper text & completed state"
+    (h/render [quo/view {:completed? true} "Numbers"])
+    (-> (h/expect (h/get-by-text "Numbers"))
+        (.toBeTruthy))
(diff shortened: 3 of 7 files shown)


ajayesivan:
@J-Son89 Since there are only dark theme designs available I'm assuming this component will only be used in dark themes.

Please review.

J-Son89:
> @J-Son89 Since there are only dark theme designs available I'm assuming this component will only be used in dark themes.
> 
> Please review.

yes that's fine, there's only dark use cases in design file so that is all that is needed for now 👍 Will check the code now

src/quo2/components/password/tips/view.cljs
```diff
@@ -0,0 +1,19 @@
+(ns quo2.components.password.tips.view
+  (:require [react-native.core :as rn]
+            [quo2.components.password.tips.style :as style]
+            [quo2.components.markdown.text :as text]))
+
+(defn view
+  "Options
+   - `text` Password tip string
+   - `completed?` Password tip state"
+  [{:keys [completed? text]}]
```
J-Son89: nit: should text be just it's own param? We use this approach in a lot of components. right @Parveshdhull

src/quo2/components/password/tips/view.cljs
```diff
@@ -0,0 +1,19 @@
...
+  [rn/view
+   {:style               style/container
+    :accessibility-label :password-tips}
+   [text/text
+    {:style  (style/tip-text completed?)
+     :weight :regular
+     :size   :paragraph-2} text]
+   (when completed? [rn/view {:style style/strike-through}])])
```
J-Son89: we could add an accessibility label (or test id) to this view component so that it can be tested to check it renders 👍

J-Son89 (approved):
nice, LGTM :)

src/quo2/components/password/tips/view.cljs
ajayesivan: Which one should I use in this case accessibility label/test ID? Any preferences?

src/quo2/components/password/tips/view.cljs
ajayesivan: Are you suggesting to do `(defn view [text completed?])`  instead of using `:keys`?

src/quo2/components/password/tips/style.cljs
```diff
@@ -0,0 +1,19 @@
+(ns quo2.components.password.tips.style
+  (:require [quo2.foundations.colors :as colors]))
+
+(def container
+  {:align-self :flex-start
+   :position   :relative})
```
flexsurfer: no need in specifying :relative

flexsurfer (approved):

Parveshdhull (approved):

src/quo2/components/password/tips/view.cljs
J-Son89: actually the other way, so the text is the second param and the first param is the options object.
```
(defn view [{:keys [completed?]} text]) 
```

src/quo2/components/password/tips/view.cljs
J-Son89: hmm, actually an accessibility label is probably good to have here. 👍. You might need to add `:accessible true` prop as well for it to work

src/quo2/components/password/tips/view.cljs
ajayesivan: Is there any reason for preferring this pattern?

ajayesivan: One issue I see with this approach is that we always have to pass the options objects,

With the previous approach we can create a component by just passing `text` - `[quo/tips {:text "Numbers"}]`. But if we want to do the same with the new approach we always have to pass an object as the first param `[quo/tips {} "Numbers"]` (this usage may not useful for this component, but I'm curious why we prefer this pattern)

ajayesivan:
PR updated with all suggested changes.

src/quo2/components/password/tips/view.cljs
J-Son89: The reason it's preferable (as far as I know) is because it is consistent across our other components. Additionally it allows us to pass through more complex text combinations and not just a string.

To avoid always passing an empty object you can do.
```
(defn view
  "Options
   - `completed?` Password tip state
   
   `text` Password tip string
   "
  ([{:keys [completed?]} text]
  [rn/view
   {:style               style/container
    :accessibility-label :password-tips}
   [text/text
    {:style  (style/tip-text completed?)
     :weight :regular
     :size   :paragraph-2} text]
   (when completed?
     [rn/view
      {:style               style/strike-through
       :accessibility-label :password-tips-completed}])])
([text] ( view {} text))
)
       
  ```

J-Son89: another example which uses mutliple arity (not sure if that's the correct name) is the `custom-color` method in colors file:
https://github.com/status-im/status-mobile/blob/develop/src/quo2/foundations/colors.cljs#L216-L228

ajayesivan: Okay, got it. Thanks for the explanation.

PR #15163, opened 2023-02-22
[#14622] Show ugly network state icons and connection bottom sheet

### LINKED ISSUE #14622: Cannot see offline state in the app and amount of peers connected, no mobile data usage flow
# Problem

On 1.20 when the user was offline/or on the cellular network it was possible and very handy to see the special state on the home view, which opens an informational popup on tap (with status node and peer-to-peer info)

Currently, there is no sign for user if they go from WI-Fi network, that can be a reason for the overuse of cellular traffic.
The task is about creating some temporary solution to get info about the app state + number of connected peers, which can be useful for debug purposes.

No design is still available for this part.

On video from 1.20:
1) User switches to the mobile network, choose the option for using the app over cellular data, tap on the connection state on the home view
2) User goes offline, can see the change in the connection state, can tap on it, and see there are no peers connected

https://user-images.githubusercontent.com/4557972/209182821-487ce856-f3fc-4f12-8ee0-5f16a8970270.mp4

## Implementation
Ugly, but understandable placeholders for described flows.

## Acceptance Criteria

For all these options as there are no designs we need to consider that  they must look ugly (from the design team) and not something that looks like a potential implementation (see https://github.com/status-im/status-mobile/pull/14616 for more context):
- can see when I'm offline
- can see when I'm switching to the cellular network (when mobile usage data is switched on or off)
- can see the number of peers I'm connected to

rasom:
fix #14622

status: ready

### DIFF
--- a/src/status_im2/common/home/view.cljs
+++ b/src/status_im2/common/home/view.cljs
@@ -58,10 +58,35 @@
              :size              :small}
             avatar)]]])
 
+(defn connectivity-sheet
+  []
+  (let [peers-count  (rf/sub [:peers-count])
+        network-type (rf/sub [:network/type])]
+    [rn/view
+     [quo/text {:accessibility-label :peers-network-type-text} (str "NETWORK TYPE: " network-type)]
+     [quo/text {:accessibility-label :peers-count-text} (str "PEERS COUNT: " peers-count)]]))
+
 (defn- right-section
   [{:keys [button-type search?]}]
-  (let [button-common-props (get-button-common-props button-type)]
+  (let [button-common-props (get-button-common-props button-type)
+        network-type        (rf/sub [:network/type])]
     [rn/view {:style style/right-section}
+     (when (= network-type "cellular")
+       [quo/button
+        (merge button-common-props
+               {:icon                false
+                :accessibility-label :on-cellular-network
+                :on-press            #(rf/dispatch [:bottom-sheet/show-sheet
+                                                    {:content connectivity-sheet}])})
+        "🦄"])
+     (when (= network-type "none")
+       [quo/button
+        (merge button-common-props
+               {:icon                false
+                :accessibility-label :no-network-connection
+                :on-press            #(rf/dispatch [:bottom-sheet/show-sheet
+                                                    {:content connectivity-sheet}])})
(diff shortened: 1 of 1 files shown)


rasom:
@churik 
I did not add "syncing on mobile" bottom shit at the moment (do we actually need it now?), added icons on the top nav when app is on cellular network or offline and added an ugly bottom sheet with peers count.

Do we need to add some access label there or it will make sense only when designs will be ready?

churik:
@rasom best to have access-labels, thank you!
and as there is no option to restrict data usage on the mobile network, that's fine not to add "syncing on mobile" bottom sheet

rasom:
@churik labels added

flexsurfer (approved):

ilmotta (approved):
I new from the title of the PR it was coming from Sir @rasom! :hankey: :skull:

VolodLytvynenko:
Hi, @rasom thank you for PR and for funny icons :)
One question. I can still see the peers in the bottom sheet  when I switch off the internet connection. I suppose when an internet connection is interrupted, the peers count should equal 0. Should it work so for now?

#### Steps to reproduce:
1. Go to home page
2. Switch off the internet connection

#### Actual result:
Count of peers still more than 0

https://user-images.githubusercontent.com/52490791/225708643-50beba9a-18c5-47a2-af51-c2f91ec45fe7.mp4
image

#### Expected result:
Count of peers = 0 if the internet connection is interrupted

#### Additional info:
Reproducible for IOS only

#### Device:
iPhone 11 Pro max; version 16

rasom:
@VladimrLitvinenko tbh either way it is irrelevant to this PR. There is some delay before that peers count gets to zero usually. But as I mentioned, that's a very distant and completely different part of the app: here we merely display what that part of the app told as to display :)

VolodLytvynenko:
> @VladimrLitvinenko tbh either way it is irrelevant to this PR. There is some delay before that peers count gets to zero usually. But as I mentioned, that's a very distant and completely different part of the app: here we merely display what that part of the app told as to display :)

 @rasom  Got it. Thank you for the clarification. No issues from my side. PR is tested and ready to be merged

